PR #49938, opened 2023-04-05
[Security] Improve DX when invalid custom authenticators

alamirault:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #... 
| License       | MIT

Before this PR error was

```yaml
security:
    firewalls:
        main:
            lazy: true
            provider: users_in_memory
            custom_authenticators:
                - App\Security\FooProvider # Mistake is here, must be an AuthenticatorInterface
```

image

### DIFF
--- a/src/Symfony/Component/Security/Http/Authentication/AuthenticatorManager.php
+++ b/src/Symfony/Component/Security/Http/Authentication/AuthenticatorManager.php
@@ -104,6 +104,10 @@ class AuthenticatorManager implements AuthenticatorManagerInterface, UserAuthent
         foreach ($this->authenticators as $authenticator) {
             $this->logger?->debug('Checking support on authenticator.', ['firewall_name' => $this->firewallName, 'authenticator' => $authenticator::class]);
 
+            if (!$authenticator instanceof AuthenticatorInterface) {
+                throw new \InvalidArgumentException(sprintf('Authenticator "%s" must implement "%s".', get_debug_type($authenticator), AuthenticatorInterface::class));
+            }
+
             if (false !== $supports = $authenticator->supports($request)) {
                 $authenticators[] = $authenticator;
                 $lazy = $lazy && null === $supports;
--- a/src/Symfony/Component/Security/Http/Tests/Authentication/AuthenticatorManagerTest.php
+++ b/src/Symfony/Component/Security/Http/Tests/Authentication/AuthenticatorManagerTest.php
@@ -77,6 +77,17 @@ class AuthenticatorManagerTest extends TestCase
         yield [[], false];
     }
 
+    public function testSupportsInvalidAuthenticator()
+    {
+        $manager = $this->createManager([new \stdClass()]);
+
+        $this->expectExceptionObject(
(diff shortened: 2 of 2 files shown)


src/Symfony/Component/Security/Http/Authentication/AuthenticatorManager.php
```diff
@@ -60,6 +60,12 @@ class AuthenticatorManager implements AuthenticatorManagerInterface, UserAuthent
      */
     public function __construct(iterable $authenticators, TokenStorageInterface $tokenStorage, EventDispatcherInterface $eventDispatcher, string $firewallName, LoggerInterface $logger = null, bool $eraseCredentials = true, bool $hideUserNotFoundExceptions = true, array $requiredBadges = [])
```
alamirault: We must probably also update validation of config in SecurityBundle, but I didn't find an example which check interface from FQCN . Should we you ReflectionClass ?https://github.com/symfony/symfony/blob/3040d433ecd90c8250b4e5df6dedb84b4897fe00/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/CustomAuthenticatorFactory.php#L49-L56

stof: Given that what you configure here is a service id and not a FQCN, we cannot validate the interface in the configuration validation (as it does not have access to the service definition)

src/Symfony/Component/Security/Http/Authentication/AuthenticatorManager.php
```diff
@@ -60,6 +60,12 @@ class AuthenticatorManager implements AuthenticatorManagerInterface, UserAuthent
...
     {
+        foreach ($authenticators as $authenticator) {
```
chalasr: Traversing `$authenticators` breaks laziness. The check should be done as late as possible, just before calling `supports()` on the authenticator.

chalasr (changes requested):

src/Symfony/Component/Security/Http/Authentication/AuthenticatorManager.php
lyrixx: But we can do it in the extension. This is help new comers more than in src/Symfony/Component/Security/Http/Authentication/AuthenticatorManager.php
(but we can keep both checks)

stof: Actually, this should be checked in a compiler pass, not in the extension (the DI extension runs with a dedicated container builder and so does not have access to the services defined elsewhere)

src/Symfony/Component/Security/Http/Authentication/AuthenticatorManager.php
alamirault: Thanks, updated

nicolas-grekas (approved):

nicolas-grekas:
Thank you @alamirault.

PR #49138, opened 2023-01-27
[HttpKernel] Create Attributes `#[MapRequestPayload]` and `#[MapQueryString]` to map Request input to typed objects

Koc:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | #36037, #36093, #45628, #47425,  #49002, #49134
| License       | MIT
| Doc PR        | TBD

Yet another variation of how we can map raw Request data to typed objects with validation. We can even build OpenApi Specification based on this DTO classes using [NelmioApiDocBundle](https://github.com/nelmio/NelmioApiDocBundle).

### `#[MapRequestPayload]`

```php
class PostProductReviewPayload
{
    public function __construct(
        #[Assert\NotBlank]
        #[Assert\Length(min: 10, max: 500)]
        public readonly string $comment,
        #[Assert\GreaterThanOrEqual(1)]
        #[Assert\LessThanOrEqual(5)]
        public readonly int $rating,
    ) {
    }
}

class PostJsonApiController
{
    public function __invoke(
        #[MapRequestPayload] PostProductReviewPayload $payload,
    ): Response {
        // $payload is validated and fully typed representation of the request body $request->getContent() 
        //  or $request->request->all()
    }
}
```

### `#[MapQueryString]`

```php
class GetOrdersQuery
{
    public function __construct(
        #[Assert\Valid]
        public readonly ?GetOrdersFilter $filter,
        #[Assert\LessThanOrEqual(500)]
        public readonly int $limit = 25,
        #[Assert\LessThanOrEqual(10_000)]
        public readonly int $offset = 0,
    ) {
    }
}

class GetOrdersFilter
{
    public function __construct(
        #[Assert\Choice(['placed', 'shipped', 'delivered'])]
        public readonly ?string $status,
        public readonly ?float $total,
    ) {
    }
}

class GetApiController
{
    public function __invoke(
        #[MapQueryString] GetOrdersQuery $query,
    ): Response {
        // $query is validated and fully typed representation of the query string $request->query->all()
    }
}
```

### Exception handling :boom: 
- Validation errors will result in an HTTP 422 response, accompanied by a serialized `ConstraintViolationList`.
- Malformed data will be responded to with an HTTP 400 error.
- Unsupported deserialization formats will be responded to with an HTTP 415 error.

## Comparison to another implementations :bookmark_tabs: 

Differences to #49002:
- separate Attributes for explicit definition of the used source
[...]

PR #49944, opened 2023-04-05
[Serializer] Make `ProblemNormalizer` give details about `ValidationFailedException` and `PartialDenormalizationException`

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

This PR provides some infrastructure that I think is missing and would be nice using in #49138

With this patch, if an `HttpException` is thrown and if it wraps either a `ValidationFailedException` or a `PartialDenormalizationException`, Symfony will return a JSON containing the details of the validation failures / missing properties when the request comes with `Accept: application/json`.

I also fixed a few minor things that I found while digging the code on this topic.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/serializer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/serializer.php
@@ -103,7 +103,7 @@ return static function (ContainerConfigurator $container) {
             ->tag('serializer.normalizer', ['priority' => -950])
 
         ->set('serializer.normalizer.problem', ProblemNormalizer::class)
-            ->args([param('kernel.debug')])
+            ->args([param('kernel.debug'), '$translator' => service('translator')->nullOnInvalid()])
             ->tag('serializer.normalizer', ['priority' => -890])
 
         ->set('serializer.denormalizer.unwrapping', UnwrappingDenormalizer::class)
--- a/src/Symfony/Component/Serializer/Normalizer/MimeMessageNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/MimeMessageNormalizer.php
@@ -17,6 +17,7 @@ use Symfony\Component\Mime\Header\Headers;
 use Symfony\Component\Mime\Header\UnstructuredHeader;
 use Symfony\Component\Mime\Message;
 use Symfony\Component\Mime\Part\AbstractPart;
+use Symfony\Component\Serializer\Exception\LogicException;
 use Symfony\Component\Serializer\SerializerAwareInterface;
 use Symfony\Component\Serializer\SerializerInterface;
 
@@ -30,10 +31,10 @@ use Symfony\Component\Serializer\SerializerInterface;
  */
 final class MimeMessageNormalizer implements NormalizerInterface, DenormalizerInterface, SerializerAwareInterface, CacheableSupportsMethodInterface
 {
-    private $serializer;
-    private $normalizer;
-    private $headerClassMap;
-    private $headersProperty;
+    private NormalizerInterface&DenormalizerInterface $serializer;
+    private PropertyNormalizer $normalizer;
+    private array $headerClassMap;
+    private \ReflectionProperty $headersProperty;
 
     public function __construct(PropertyNormalizer $normalizer)
     {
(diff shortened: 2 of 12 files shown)


src/Symfony/Component/Serializer/Normalizer/ProblemNormalizer.php
```diff
@@ -53,15 +60,33 @@ public function normalize(mixed $object, string $format = null, array $context =
...
 
+        $data = [];
+        $exception = $context['exception'] ?? null;
+        if ($exception instanceof HttpExceptionInterface) {
+            $exception = $exception->getPrevious();
+
+            if ($exception instanceof PartialDenormalizationException) {
+                $data = [
+                    self::TYPE => 'https://symfony.com/errors/denormalization',
+                    'errors' => array_map(fn ($e) => ['title' => $e->getMessage()], $exception->getErrors()),
```
lyrixx: I would be nice to add some hint if possible!

```
if ($exception->canUseMessageForUser()) {
    $parameters['hint'] = $exception->getMessage();
}
```

nicolas-grekas: updated

src/Symfony/Component/Serializer/Normalizer/MimeMessageNormalizer.php
```diff
@@ -57,6 +58,9 @@ public function getSupportedTypes(?string $format): array
 
     public function setSerializer(SerializerInterface $serializer): void
     {
+        if (!$serializer instanceof NormalizerInterface || !$serializer instanceof DenormalizerInterface) {
+            throw new LogicException(sprintf('The passed serializer should implement both NormalizerInterface and DenormalizerInterface, "%s" given.', get_debug_type($serializer)));
```
WebMamba, line 62: This message sounds a bit weird to me. You test with || so if serializer implement NomalizerInterface OR DenormalizerInterface. But in your message you use 'and' like you are using && operator.

nicolas-grekas: I also use `!`, that makes it an `&&`, as in `!(a && b)`

lyrixx (approved):
LGTM 👍🏼

src/Symfony/Bundle/FrameworkBundle/Resources/config/serializer.php
```diff
@@ -103,7 +103,7 @@
             ->tag('serializer.normalizer', ['priority' => -950])
 
         ->set('serializer.normalizer.problem', ProblemNormalizer::class)
-            ->args([param('kernel.debug')])
+            ->args([param('kernel.debug'), '$translator' => service('translator')->nullOnInvalid()])
```
stof, line 106: should we really rely on the named parameter here ? We don't include parameter names in our BC policy except for attribute constructors.

src/Symfony/Component/Serializer/Normalizer/ProblemNormalizer.php
```diff
@@ -53,15 +63,47 @@ public function normalize(mixed $object, string $format = null, array $context =
...
+                ];
+            } elseif ($exception instanceof ValidationFailedException
+                && $this->serializer instanceof NormalizerInterface
+                && $this->serializer->supportsNormalization($exception->getViolations(), $format, $context)
+            ) {
+                $data = $this->serializer->normalize($exception->getViolations(), $format, $context);
+            }
+        }
+
+        $context += $this->defaultContext;
```
stof: moving this after the usage of `$context['debug'] ?? true` looks wrong to me.

src/Symfony/Bundle/FrameworkBundle/Resources/config/serializer.php
nicolas-grekas: we have to because of the way SerializerPass works:
https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Serializer/DependencyInjection/SerializerPass.php#L79

src/Symfony/Component/Serializer/Normalizer/ProblemNormalizer.php
nicolas-grekas: should be good now

stof (approved):

yceruto:
Great! 👍

PR #49947, opened 2023-04-05
Extract the `EmojiTransliterator` to a dedicated package

### LINKED ISSUE #49943: Split intl packages in smaller packages
### Description

Right now Symfony Intl package is 43MB, which makes all Applications really big 
image

The problems for me are:

- Packaging for Bref can be a problem as AWS has a limit of 250 MB (unzipped)
- Waste of Traffic and Disk Space

### Example

My ideas so far:

- Create separate packages for the actual data emoji, locales, timezones and optin people for each feature (symfony/intl-data-emoji)
- Or complete symfony/intl-emoji like split the repository on Intl level again

stof:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | Fix #49943
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/18172

Due to the size of the data needed to transliterate emojis, using a dedicated package allows to include the package in the project only when needed.
In Symfony 6.x, the symfony/intl component depends on this component for BC reasons. Root projects can implement a replace hack in their composer.json to skip this package if they don't need it. This dependency will be removed in Symfony 7.0.


src/Symfony/Component/String/composer.json
```diff
@@ -24,12 +24,13 @@
...
     "require-dev": {
         "symfony/error-handler": "^5.4|^6.0",
-        "symfony/intl": "^6.2",
+        "symfony/emoji-transliterator": "^6.3",
         "symfony/http-client": "^5.4|^6.0",
         "symfony/translation-contracts": "^2.5|^3.0",
         "symfony/var-exporter": "^5.4|^6.0"
     },
     "conflict": {
+        "symfony/intl": "6.2.*",
```
stof, line 33: This conflict is part of the BC layer. An existing project using `$slugger->withEmoji(true)` will have `symfony/string: ^6.2, symfony/intl: ^6.2` as requirement. If we were upgrading to symfony/string 6.3 while keeping symfony/intl 6.2.x, we would not be able to use EmojiTransliterator as symfony/string 6.3 uses the new class from symfony/emoji-transliterator. If symfony/intl is also updated to 6.3, the BC requirement on symfony/emoji-transliterator in symfony/intl will make sure that the code keeps working (for 7.0, you will need to depend on `symfony/emoji-transliterator` though)

src/Symfony/Component/Intl/Tests/Transliterator/EmojiTransliteratorTest.php
```diff
@@ -82,30 +84,6 @@ public static function provideTransliterateTests(): iterable
...
-     * @dataProvider provideLocaleTest
-     */
-    public function testAllTransliterator(string $locale)
-    {
-        $tr = EmojiTransliterator::create($locale);
-
-        $this->assertNotEmpty($tr->transliterate('😀'));
-    }
-
-    public static function provideLocaleTest(): iterable
```
stof, line 95: I intentionally kept most tests as legacy tests, to ensure that the BC layer works fine (and I added the assertion about actually instantiating the requested class as my first version of the BC layer was broken in that regard, which could break type checks for existing code using the old class as type). I only removed this test because it relied on the data to find all supported locales, and the data is gone from this component. This is still tested in the new component.

lyrixx:
Thanks 👍🏼

stof:
@fabpot fabbot needs to be updated to exclude the `EmojiTransliterator/Resources/data` folder, similar to how it excludes `Intl/Resources/data`

Regarding other CI failures:
- intl-data is fixed in #49948. My PR is not the cause of this failure.
- Psalm is confused by the conditional class declaration, which is something we are fine about here.
- the unit test failure on 8.2 is one of those random failures we have these days, with PHPUnit saying OK and PHP crashing after that.

src/Symfony/Component/EmojiTransliterator/README.md
```diff
@@ -0,0 +1,14 @@
+EmojiTransliterator Component
+=============================
+
+The EmojiTransliterator component provides an ext-intl Transliterator that can
+transliterate emojis to ASCII.
+
+Resources
+---------
+
+ * [Documentation](https://symfony.com/doc/current/components/emoji_transliterator.html)
```
OskarStark, line 10: This page needs to be created in `symfony/symfony-docs` afterwards

OskarStark: I created an issue https://github.com/symfony/symfony-docs/issues/18171

src/Symfony/Component/EmojiTransliterator/Tests/EmojiTransliteratorTest.php
```diff
@@ -0,0 +1,181 @@
...
+ */
+
+namespace Symfony\Component\EmojiTransliterator\Tests;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Finder\Finder;
+use Symfony\Component\EmojiTransliterator\EmojiTransliterator;
+
+/**
+ * @requires extension intl
```
OskarStark, line 19: Do we need to add `ext-intl` as dependency to composer.json?

stof: There is a TODO in the code to do it in 7.0.

In 6.3, we have `symfony/intl` depending on the new package for BC reasons, and we don't want to add a ext-intl dependency there.

src/Symfony/Component/EmojiTransliterator/README.md
stof: I created a PR to create that page

src/Symfony/Component/String/composer.json
```diff
@@ -24,12 +24,13 @@
     },
     "require-dev": {
         "symfony/error-handler": "^5.4|^6.0",
-        "symfony/intl": "^6.2",
+        "symfony/emoji-transliterator": "^6.3",
```
kaznovac, line 27: maybe put the `emoji-transliterator` in composer's `suggest` section and explain its use

src/Symfony/Component/Intl/composer.json
```diff
@@ -24,11 +24,12 @@
         }
     ],
     "require": {
-        "php": ">=8.1"
+        "php": ">=8.1",
+        "symfony/deprecation-contracts": "^2.5|^3",
+        "symfony/emoji-transliterator": "^6.3"
```
kaznovac, line 29: can this be optional?
i.e. when i need the `intl` component without the emoji support

wouterj: Not in 6.x, see the pull request description.

src/Symfony/Component/String/composer.json
stof: We stopped using the suggest section IIRC

fabpot: Indeed

nicolas-grekas:
Closing in favor of #49970, as discussed there. Thanks for the PR!

PR #49291, opened 2023-02-07
[Serializer] Add methods `getSupportedTypes` to allow better performance

tucksaun:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes (new method in the interfaces, one interface deprecated)
| License       | MIT
| Doc PR        | to be written

The PRs allows normalizers or denormalizers to expose their supported types and the associated cacheability to the Serializer.
With this info,  even if the `supports*` call is not cacheable, the Serializer can skip a ton of method calls to `supports*` improving performance substaintially in some cases:
Screenshot 2023-02-02 at 15 46 49

<details>
 <summary>I found this design while working on a customer project performance (a big app built around API Platform): we reached the point where the slowest part of main application endpoint was `Symfony\Component\Serializer\Serializer::getNormalizer`.</summary>

After some digging, we found out we were experiencing the conjunction of two phenomenons:
- the application is quite complex and returns deep nested and repeating structures, exposing the underlying bottleneck;
- and a lot of custom non-cacheable normalizers.
Because most of the normalizers are not cacheable, the Serializer has to call every normalizer over and over again leading `getNormalizer` to account for 20% of the total wall time:
Screenshot 2023-02-07 at 16 56 02

We first tried to improve cacheability based on context without much success, then an approach similar to #45779 with some success but still feeling this could be faster.
We then thought that even if the `supportsNormalization` could not be cached (because of the context), maybe we could avoid the calls at the origin by letting the `Normalizers` expose the types they support and came to this PR with pretty good results.
</details>

The perfornance improvement was only measured by adapting Symfony's normalizers as well as the project ones, proper third party normalizers updates should improve performance even more.

This should effectively replaces the `CacheableSupportsMethodInterface` as the cacheability can now be returned by `getSupportedTypes`.

PR #49950, opened 2023-04-05
[Serializer] Marking some Normalizer classes as final

tucksaun:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Tickets       | https://github.com/symfony/symfony/pull/49291#pullrequestreview-1306165711
| License       | MIT
| Doc PR        | n/a

As mentioned in https://github.com/symfony/symfony/pull/49291#pullrequestreview-1306165711.

Making those classes final will allow to get rid of all the `__CLASS__ === static::class` for cacheability checks in 7.0

### DIFF
--- a/src/Symfony/Component/Serializer/CHANGELOG.md
+++ b/src/Symfony/Component/Serializer/CHANGELOG.md
@@ -10,7 +10,17 @@ CHANGELOG
  * Make `ProblemNormalizer` give details about `ValidationFailedException` and `PartialDenormalizationException`
  * Deprecate `MissingConstructorArgumentsException` in favor of `MissingConstructorArgumentException`
  * Deprecate `CacheableSupportsMethodInterface` in favor of the new `getSupportedTypes(?string $format)` methods
- * The following Normalizer classes will become final in 7.0: `ObjectNormalizer`
+ * The following Normalizer classes will become final in 7.0:
+   * `ConstraintViolationListNormalizer`
+   * `CustomNormalizer`
+   * `DataUriNormalizer`
+   * `DateIntervalNormalizer`
+   * `DateTimeNormalizer`
+   * `DateTimeZoneNormalizer`
+   * `GetSetMethodNormalizer`
+   * `JsonSerializableNormalizer`
+   * `ObjectNormalizer`
+   * `PropertyNormalizer`
 
 6.2
 ---
--- a/UPGRADE-6.3.md
+++ b/UPGRADE-6.3.md
@@ -117,4 +117,14 @@ Serializer
 
  * Deprecate `MissingConstructorArgumentsException` in favor of `MissingConstructorArgumentException`
  * Deprecate `CacheableSupportsMethodInterface` in favor of the new `getSupportedTypes(?string $format)` methods
- * The following Normalizer classes will become final in 7.0: `ObjectNormalizer`
+ * The following Normalizer classes will become final in 7.0:
+   * `ConstraintViolationListNormalizer`
+   * `CustomNormalizer`
+   * `DataUriNormalizer`
+   * `DateIntervalNormalizer`
+   * `DateTimeNormalizer`
+   * `DateTimeZoneNormalizer`
+   * `GetSetMethodNormalizer`
+   * `JsonSerializableNormalizer`
+   * `ObjectNormalizer`
+   * `PropertyNormalizer`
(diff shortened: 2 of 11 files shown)


src/Symfony/Component/Serializer/Normalizer/ObjectNormalizer.php
```diff
@@ -25,6 +25,8 @@
  * Converts between objects and arrays using the PropertyAccess component.
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
+ *
+ * @final since Symfony 6.3
```
tucksaun: @nicolas-grekas I'm puzzled about this one as it fits into the category of the one we wanted to make final but it is actually extending in the tests.
Also IIRC ApiPlatform also extends it

nicolas-grekas: when it's "since Symfony 6.x" it means we need to keep bc with child classes

tucksaun: yes.
what I mean is that I don't see how we can achieve what is done in 7.0 once this class is final.
so I'm wondering if this makes sense to make it final in the first place?

stof: however, if ApiPlatform extends it, it might make it hard for them to add support for Symfony 7.0. So the reason why it is extended should probably be investigated.

tucksaun: what do you think of removing the change for this specific class and opening another PR for this one where we can discuss about the test cases and ApiPlatform with @dunglas ?

nicolas-grekas: let's do that yes

src/Symfony/Component/Serializer/CHANGELOG.md
```diff
@@ -9,6 +9,17 @@ CHANGELOG
...
+   * `ConstraintViolationListNormalizer`
+   * `CustomNormalizer`
+   * `DataUriNormalizer`
+   * `DateIntervalNormalizer`
+   * `DateTimeNormalizer`
+   * `DateTimeZoneNormalizer`
+   * `GetSetMethodNormalizer`
+   * `JsonSerializableNormalizer`
+   * `ObjectNormalizer`
+   * `PropertyNormalizer`
```
nicolas-grekas, line 23: we need a similar message in UPGRADE-6.3.md

src/Symfony/Component/Serializer/Normalizer/ObjectNormalizer.php
tucksaun: done, see #49953

src/Symfony/Component/Serializer/CHANGELOG.md
tucksaun: added

fabpot (approved):

fabpot:
Thank you @tucksaun.

ninsuo:
Hello peeps,

Since this class is now final, wouldn't it be possible to put `extractAttributes`/`getAttributeValue`/`setAttributeValue`  methods in a trait, to make them reusable?

To give you more context, I've a normalizer that extends `ObjectNormalizer` and supports classes that implement `HasSensitivePropertiesInterface`; it first checks for `#[NullIfNotGranted(role)]` attributes in the object and set unauthorized properties to null, and then calls the parent ObjectNormalizer. 

The whole `PropertyAcces` logic is quite useful and it seems that there are no built-in replacements for them, what's your insights?

Thank you!

Cheers,

ninsuo:
Sorry, wrong PR, going to leave this message into the right ObjectNormalizer's one.

larowlan:
I've got a question about this one, in the docs we say

> The custom name converter can be used by passing it as second parameter of any class extending [AbstractNormalizer](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php), including [GetSetMethodNormalizer](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Serializer/Normalizer/GetSetMethodNormalizer.php) and [PropertyNormalizer](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Serializer/Normalizer/PropertyNormalizer.php):

But now we've marked GetSetMethodNormalizer as final, I assume this will become `final class` in symfony 7.

So what is the new recommended method on that front? And should we update the docs?

tucksaun:
I believe the "including" wording from the sentence is not here to meant "any classes extending GetSetMethodNormalizer or PropertyNormalizer" but more to mention that this statement is true for any classes extending `AbstractNormalizer` and that `GetSetMethodNormalizer` and `PropertyNormalizer` happen to extends `AbstractNormalizer` and so you can pass the name converter to those specific classes.

PR #49952, opened 2023-04-05
[FrameworkBundle] Make private helper methods protected in test traits

loevgaard:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

Making these methods protected instead of private will allow users to use these methods in their own test files (e.g. traits and extended test classes). 

An example of my own application. I wanted to make an assertion that tested if a flash message was shown. It would have been very easy to just copy the existing assertions and just change one line, but because I can't use the `self::getCrawler()` and `self::getClient()` it was much more tedious.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Test/BrowserKitAssertionsTrait.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Test/BrowserKitAssertionsTrait.php
@@ -156,7 +156,7 @@ trait BrowserKitAssertionsTrait
         self::assertThat(self::getClient(), $constraint, $message);
     }
 
-    private static function getClient(AbstractBrowser $newClient = null): ?AbstractBrowser
+    protected static function getClient(AbstractBrowser $newClient = null): ?AbstractBrowser
     {
         static $client;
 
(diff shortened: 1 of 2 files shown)


lyrixx:
You can import theses trait directly in your test. Does that solves your issue?

lyrixx:
You can import theses trait directly in your test. Does that solves your issue?

stof:
The benefit of keeping them private is that using the trait in your testcase does not automatically expose those methods to your child classes (in which case they become part of your own API to be covered by BC for child classes). And you can still choose to make those methods protected when using the trait if you want to expose such protected APIs.

Note that private methods of a trait **are** usable by the class using the trait. They are private to the user of the trait, not to the trait itself (traits are just compiler-assisted copy/paste)

stof:
For now, I'm -1 on that change. I don't see a good reason to do it but I see some drawbacks (see my previous comment).
To me, the actual reason for opening this PR is a misunderstanding of what private visibility actually means in a trait.

loevgaard:
> You can import theses trait directly in your test. Does that solves your issue?

This was in fact my first action, but my IDE complained. At least I thought it did because I just did it again based on your comments, and it worked.

Sorry for wasting your time, guys!

Have a great evening!

PR #49953, opened 2023-04-05
[Serializer] Mark `ObjectNormalizer` as final for 7.0

tucksaun:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Tickets       | https://github.com/symfony/symfony/pull/49291#pullrequestreview-1306165711
| License       | MIT
| Doc PR        | none

Extracted from #49950 and related to https://github.com/symfony/symfony/pull/49291#pullrequestreview-1306165711

### DIFF
--- a/UPGRADE-6.3.md
+++ b/UPGRADE-6.3.md
@@ -112,3 +112,4 @@ Serializer
 
  * Deprecate `MissingConstructorArgumentsException` in favor of `MissingConstructorArgumentException`
  * Deprecate `CacheableSupportsMethodInterface` in favor of the new `getSupportedTypes(?string $format)` methods
+ * The following Normalizer classes will become final in 7.0: `ObjectNormalizer`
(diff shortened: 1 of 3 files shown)


tucksaun:
Friendly ping @dunglas @soyuka as this small change would actually affect ApiPlatform's [DocumentNormalizer](https://github.com/api-platform/core/blob/ccef472cade42b189fa8cb7793e27fad1349e289/src/Elasticsearch/Serializer/DocumentNormalizer.php#L24)
Looks to me like ApiPlatform could go with decoration instead?

soyuka:
Indeed IMO this is a mistake from us, we should've used the AbstractObjectNormalizer instead. :+1: on making this class final!

derrabus (approved):

fabpot (approved):

fabpot:
Thank you @tucksaun.

ninsuo:
Hello peeps,

Since this class is now final, wouldn't it be possible to put extractAttributes/getAttributeValue/setAttributeValue methods in a trait, to make them reusable?

To give you more context, I've a normalizer that extends ObjectNormalizer and supports classes that implement HasSensitivePropertiesInterface; it first checks for #[NullIfNotGranted(role)] attributes in the object and set unauthorized properties to null, and then calls the parent ObjectNormalizer.

The whole PropertyAcces logic is quite useful and it seems that there are no built-in replacements for them, what's your insights?

Thank you!

Cheers,

---

Okey, the solution is to inject it instead of extending it.
Injecting the ObjectNormalizer into my normalizer, and calling it instead does the job.
I've the head in the handlebars and can't see such easy solutions 😅

Issue #49804, opened 2023-03-25
[Scheduler] handling failures


kbond:
### Symfony version(s) affected

6.3

### Description

When running `messenger:consume` for your schedule and a scheduled message handler throws an exception, there is a hard failure and the process exits. [This is the exception](https://github.com/symfony/symfony/blob/aadd302a0a2d2505979b9eb34e6767e98533b3ef/src/Symfony/Component/Scheduler/Messenger/SchedulerTransport.php#L43) that's thrown.

Is this expected? If not, what should we do?

### How to reproduce

Run `bin/console messenger:consume default_schedule` and have a handler throw an exception.

### Possible Solution

I'm thinking by default we should swallow the exception and move on? Of course dispatching relative events (#49803).

Perhaps this behaviour should be configurable?
- Retry
- Send to failures transport

tucksaun:
I wanted to share my experience on this topic because I recently experienced the retry mechanism (because I configured the transport myself the retry mechanism became enabled automatically by Messenger)
In that case, one encounters another kind of failure: "cannot send messages" because this is not possible with the current SchedulerTransport. This raises the question of how retry would be possible.

But anyway, because `SchedulerTransport::reject` throws an exception, Messenger can not go on. If we remove the exception thrown in `reject`, Messenger is at least be able to report the initial error correctly and should be able to send the message to the failure transport if one is configured.

Regarding retries, I see three options here:
 - We could state that if retries are actually important for someone, then they should instead use a `RedispatchMessage` to dispatch the message to a transport that actually supports retries.
- We could let the user configure a retry transport where to send retries, and only forward messages with a scheduled and redeliveryStamp
- We could embed an in-memory transport to perform retries

Option 1 has the advantage of relying on all the usual Messenger infrastructure and keeping the Scheduler code simple.

Note that the last option is not safe as it does not persist errors across consumer reboots but it can be a fallback of option 2 and it looks like it works "okay" from a quick test I just did:
```
16:38:49 WARNING   [messenger] Error thrown while handling message App\Messenger\TestMessage. Sending for retry #1 using 1000 ms delay. Error: "Handling "App\Messenger\TestMessage" failed: " ["class" => "App\Messenger\TestMessage","retryCount" => 1,"delay" => 1000,"error" => "Handling "App\Messenger\TestMessage" failed: ","exception" => Symfony\Component\Messenger\Exception\HandlerFailedException^ { …}]
[...]

PR #49964, opened 2023-04-06
[Scheduler] Allow messages from the SchedulerTransport to be rejected

tucksaun:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no 
| Tickets       | Related to #49804
| License       | MIT

Currently, when the processing of a scheduled message fails, Messenger will not be able to go on because the SchedulerTransport will throw an exception. More importantly, the original error will be swallowed by this one and not sent to the failure transport if one is configured. By allowing message rejection, Messenger will continue working as expected.

This is the first step to more changes regarding failure handling in ScheduledMessage processing.

/cc @fabpot @kbond

### DIFF
--- a/src/Symfony/Component/Scheduler/Messenger/SchedulerTransport.php
+++ b/src/Symfony/Component/Scheduler/Messenger/SchedulerTransport.php
@@ -40,7 +40,7 @@ class SchedulerTransport implements TransportInterface
 
     public function reject(Envelope $envelope): void
     {
-        throw new LogicException(sprintf('Messages from "%s" must not be rejected.', __CLASS__));
+        // ignore
     }
 
     public function send(Envelope $envelope): Envelope
--- a/src/Symfony/Component/Scheduler/Tests/Messenger/SchedulerTransportTest.php
+++ b/src/Symfony/Component/Scheduler/Tests/Messenger/SchedulerTransportTest.php
@@ -52,7 +52,7 @@ class SchedulerTransportTest extends TestCase
     {
         $transport = new SchedulerTransport($this->createMock(MessageGeneratorInterface::class));
 
-        $this->expectException(LogicException::class);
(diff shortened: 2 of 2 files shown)


kbond:
To clarify the _on failure_ scenarios:
1. no `failure_transport` configured: swallows the error and appropriate events are dispatched
2. `failure_transport` configured: swallows the error, appropriate events are dispatched and sent to the failure transport (no retries)
3. `retry_strategy` configured for your scheduler transport: _ignored_

kbond (approved):
Perfect! I think we can mark this as a fix for https://github.com/symfony/symfony/issues/49804. Additional features could be added but it's no longer a _bug_.

tucksaun:
Actually:
3. `retry_strategy` configured for your scheduler transport: a "cannot send messages" exception is thrown

kbond:
> Actually:
> 3. retry_strategy configured for your scheduler transport: a "cannot send messages" exception is thrown

That's better I think (doesn't give you the impression that they're being retried).

fabpot:
As the Scheduler component is all about **recurring** messages, I think they should never be retried.

fabpot:
If you want the message to be retried, you should use the `RedispatchMessage` feature to send the message to another transport that can be retried.

kbond:
> As the Scheduler component is all about recurring messages, I think they should never be retried.

I'm fine with this stance for https://github.com/symfony/symfony/pull/49965 but this PR is required imo.

tucksaun:
I agree (and therefore closed #49965).
But this PR is required if we don't want workers to crash when the handling of a recurring message fails (because Messenger Worker will call `reject` when a failure happens leading to an unhandled exception)

fabpot (approved):

fabpot:
Thank you @tucksaun.

PR #49965, opened 2023-04-06
Scheduler/retries and failures

tucksaun:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | subjective
| New feature?  | yes
| Deprecations? | no
| Tickets       | Related to #49804
| License       | MIT
| Doc PR        | Can be directly integrated into https://github.com/symfony/symfony-docs/issues/18067

Right now the handling of failures happening during Scheduled Messages handling is not great (see #49804).
#49964 allows one to be notified of errors and save failed messages if a failure transport is configured.

But one might want to have messages retried automatically which is not possible at the moment.
There are two approaches to this:
- either we do nothing and one is encouraged to use `RedispatchMessage` to ensure retry happens when necessary on a transport supporting retries
- or we allow a user to specify which transport should be used for retries, leveraging in memory for simple use cases.

This PR is an attempt at the second approach.
It allows one to define the scheduler transport with a retry option to the DSN:
```yaml
framework:
    messenger:
        transports:
             scheduler_default: 'schedule://default?retry=async'
```
In case of failure when Messenger tries to send the message to ScheduleTransport it will then actually forward the envelope to the specified transport which will then be picked up by regular consumers.
For simple use cases, one can use the `'schedule://default?retry` that will use the in-memory transport and the SchedulerTransport will then pick them up once available after the retry delay.

/cc @kbond @fabpot

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/scheduler.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/scheduler.php
@@ -18,6 +18,7 @@ return static function (ContainerConfigurator $container) {
         ->set('scheduler.messenger_transport_factory', SchedulerTransportFactory::class)
             ->args([
                 tagged_locator('scheduler.schedule_provider', 'name'),
+                tagged_locator('messenger.receiver', 'alias'),
                 service('clock'),
             ])
             ->tag('messenger.transport_factory')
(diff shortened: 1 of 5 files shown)


tucksaun:
Closing per https://github.com/symfony/symfony/pull/49964#issuecomment-1500381545 and https://github.com/symfony/symfony/pull/49964#issuecomment-1500383698

PR #49966, opened 2023-04-06
[Security] Create Controller Argument Resolver for Security Token

Koc:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | TBD

Sometimes Authentication Token contains more data than we have persisted in the User. For this cases we can resolve Controller Arguments typed with `TokenInterface` or any sub-class/interface.

## Usage
```php
final class GetCurrentAuthenticatedUserController
{
    public function __invoke(TokenInterface $token)
    {
        // deal with $token
    }
}
```

## TODO:
- [x] Add changelog entry
- [x] Cover with tests

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/Resources/config/security.php
+++ b/src/Symfony/Bundle/SecurityBundle/Resources/config/security.php
@@ -42,6 +42,7 @@ use Symfony\Component\Security\Core\User\InMemoryUserProvider;
 use Symfony\Component\Security\Core\User\MissingUserProvider;
 use Symfony\Component\Security\Core\Validator\Constraints\UserPasswordValidator;
 use Symfony\Component\Security\Http\Authentication\AuthenticationUtils;
+use Symfony\Component\Security\Http\Controller\SecurityTokenValueResolver;
 use Symfony\Component\Security\Http\Controller\UserValueResolver;
 use Symfony\Component\Security\Http\EventListener\IsGrantedAttributeListener;
 use Symfony\Component\Security\Http\Firewall;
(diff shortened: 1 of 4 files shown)


src/Symfony/Component/Security/Http/Controller/SecurityTokenValueResolver.php
```diff
@@ -0,0 +1,50 @@
...
+        if (!($type = $argument->getType()) || !is_subclass_of($type, TokenInterface::class)) {
+            return [];
+        }
+
+        if (null === $token = $this->tokenStorage->getToken()) {
+            if ($argument->isNullable()) {
+                return [];
+            }
+
+            throw new HttpException(Response::HTTP_UNAUTHORIZED, 'The security token required but not provided.');
```
ro0NL: UserValueResolver uses 403

Koc: Yeah, I know. But semantically 401 is more logical, isn't?

Token not provided - 401
Token provided but without required role - 403

Koc:
Test failures looks unrelated, so PR more or less ready for review

src/Symfony/Bundle/SecurityBundle/Resources/config/security.php
```diff
@@ -102,6 +103,12 @@
             ])
             ->tag('controller.argument_value_resolver', ['priority' => 120, 'name' => UserValueResolver::class])
 
+        ->set('security.security_token_value_resolver', SecurityTokenValueResolver::class)
+            ->args([
+                service('security.token_storage'),
+            ])
+            ->tag('controller.argument_value_resolver', ['priority' => 120])
```
nicolas-grekas: this is missing the name attribute

Koc: such Argument Resolver has no any attribute. We are injecting any sub-class of `TokenInterface` interface. Behavior is simpler comparing to `#[CurrentUser]`

nicolas-grekas: See 6 lines above, we still give the corresponding resolver a name.
This allows short-circuiting the resolver chain by using `#[MapResolver(UserValueResolver::class)]`

Koc: @nicolas-grekas done

nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @Koc.

PR #48525, opened 2022-12-06
[HttpFoundation] Add `ParameterBag::getString()` and deprecate accepting invalid values

GromNaN:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | Fix #44787 #48219
| License       | MIT
| Doc PR        | symfony/symfony-docs#...

There were a lot of discussions on #44787 regarding the implementation. The main debate is to determine the right behavior when the value is invalid: try to convert the value, throw an exception, return the default value or the falsy?

I added plenty of test cases for the methods `getAlpha`, `getAlnum`, `getDigits`, `getInt`, ~`getInteger`~, `getString` so that we can discuss the expected result for each input value.

My goals:
- Provide safe methods to get values in the expected type. Example: The parameters being generally of type `string`, the `getString` method is useful to be sure we don't get an array.
- Reduce deprecations on methods that exist since Symfony 2.0, in one of the most popular Symfony component.

How are these getter methods used?
- Most of the time, parameter values are `string` (from routing, query string, request payload). `get` is used but does not validate the return type.
- `getInt` is used for [pagination (UX Autocomplete)](https://github.com/symfony/ux-autocomplete/blob/697f1cb4914480b811d978efe031a6f4a0dc3814/src/Controller/EntityAutocompleteController.php#L41) or [getting an index (EasyAdmin)](https://github.com/EasyCorp/EasyAdminBundle/blob/f210bd1e494b699dec54d2ef29302db1211267b5/src/Context/AdminContext.php#L157-L158)
- I think there was an unidentified breaking change when we introduced return types [#42507](https://github.com/symfony/symfony/pull/42507/files#diff-04f3b8ea71029f48853b804129631aeb9bf3dad7a7a398feb9568bf5397d0e69L117). Methods `getAlpha`, `getAlnum` and `getDigits` could return an array, but their return type have been modified to `string`, resulting in a `TypeError`. [example of use](https://github.com/asaaa1997/Licencjat/blob/8c10abaf87120c904557977ae14284c7d443c9a7/src/Controller/QuizController.php#L164)

PR #49972, opened 2023-04-07
[HttpFoundation] Adds various return-typed getters to ParameterBag

### LINKED ISSUE #45775: Add various return-typed getters to ParameterBag
### Description

right now, `ParameterBag` has only a few getter options for fetching values:

```
get(): mixed
getAlpha(): string
getAlnum(): string
getDigits(): string
getInt(): int
getBoolean(): bool
```

For those of us using static analyzers like PHPStan, we often find ourselves having to add a bunch of manual type checking code to ensure that the parameter we are fetching is of a specified type.  I see two main problems/annoyances here:

1. In terms of fetching configuration parameters, this often means implementing type checks to ensure that the returned value is indeed the type that we already know it to be.
2. In terms of fetching Request and Query parameters, we have to contend with the fact that forms cast everything to strings, which means fetching a mixed result from `get()`, type checking it as a string, then casting it to the actual type it should be. 
 
The two exclusions here are `getInt()` and `getBoolean()`. Furthermore, `getInt()` is potentially dangerous with it's naked cast, as you may be fetching `5.5` or `"5.5"` and it will return `5` instead of throwing an exception.

### Example

Would the community find it valuable to have many more typed getters along the lines of `getInt()` and `getBoolean()` that return *validated* parameters returned as their proper type? I'm thinking things along the lines of:

```
getArray(): array
getArrayOrNull(): ?array
getBoolean(): bool
getBooleanOrNull(): ?bool
getInt(): int
getIntOrNull(): ?int
getFloat(): float
getFloatOrNull(): ?float
getString(): string
getStringOrNull(): ?string
```

here are some quick examples, including an improvement of `getInt()`:

```php
protected function getInt(string $key, int $default = 0): int
{
    if (is_int($value = $this->get($key, $default))) {
        return $value;
    }
    throw $this->getException("int", $value);
}

protected function getIntOrNull(string $key, int $default = 0): ?int
{
    $value = $this->get($key, $default)
    if (is_int($value) || is_null($value)) {
        return $value;
    }
    throw $this->getException("int|null", $value);
}

protected function getString(string $key, string $default = ""): string
{
    if (is_string($value = $this->get($key, $default))) {
        return $value;
    }
    throw $this->getException("string", $value);
}

protected function getStringOrNull(string $key, string $default = ""): ?string
{
    $value = $this->get($key, $default)
    if (is_string($value) || is_null($value)) {
        return $value;
    }
    throw $this->getException("string|null", $value);
}

/** @param mixed $value */
private function getException(string $expected, $value): TypeException
{
    return new TypeException("expected parameter of type $expected but found " . gettype($value));
}
```

If the Symfony developers think this is a good idea and/or community members want it, I can work on a PR.  I just didn't want to do the work only to find that this discussion has already been had, or there's something I'm missing, or such a feature wouldn't be accepted by Symfony.

And perhaps this logic shouldn't live in `ParameterBag`.  Maybe it warrants a new `Typer` object that can be recycled all over the place.  I currently have this type of logic implemented on a Trait that I include as needed.

mleczakm:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #45775 
| License       | MIT

### DIFF
--- a/src/Symfony/Component/HttpFoundation/ParameterBag.php
+++ b/src/Symfony/Component/HttpFoundation/ParameterBag.php
@@ -149,7 +149,7 @@ class ParameterBag implements \IteratorAggregate, \Countable
     public function getStringOrNull(string $key): ?string
     {
         return $this->has($key)
-            ? $this->filter($key, null, \FILTER_DEFAULT, \FILTER_NULL_ON_FAILURE)
+            ? $this->getString($key)
             : null;
     }
 
@@ -165,7 +165,7 @@ class ParameterBag implements \IteratorAggregate, \Countable
     public function getIntOrNull(string $key): ?int
     {
         return $this->has($key)
-            ? $this->filter($key, null, \FILTER_VALIDATE_INT, \FILTER_NULL_ON_FAILURE)
+            ? $this->getInt($key)
             : null;
     }
 
@@ -180,7 +180,7 @@ class ParameterBag implements \IteratorAggregate, \Countable
     public function getBooleanOrNull(string $key): ?bool
     {
         return $this->has($key)
-            ? $this->filter($key, null, \FILTER_VALIDATE_BOOL, \FILTER_NULL_ON_FAILURE)
+            ? $this->getBoolean($key)
             : null;
     }
 
--- a/src/Symfony/Component/HttpFoundation/ParameterBag.php
+++ b/src/Symfony/Component/HttpFoundation/ParameterBag.php
@@ -146,6 +146,11 @@ class ParameterBag implements \IteratorAggregate, \Countable
         return (string) $value;
     }
 
+    public function getStringOrNull(string $key, string $default = null): ?string
+    {
+        return $this->filter($key, $default, \FILTER_SANITIZE_STRING, \FILTER_NULL_ON_FAILURE);
+    }
+
     /**
      * Returns the parameter value converted to integer.
      */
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/HttpFoundation/Tests/ParameterBagTest.php
```diff
@@ -378,6 +378,30 @@ public function testGetEnumThrowsExceptionWithInvalidValueType()
...
+    {
+        $bag = new ParameterBag(['valid-value' => 'foo']);
+
+        $this->assertSame('foo', $bag->getStringOrNull('valid-value'));
+        $this->assertNull($bag->getStringOrNull('invalid-key'));
+    }
+
+    public function testGetIntOrNull(): void
+    {
+        $bag = new ParameterBag(['valid-value' => 1]);
```
GromNaN, line 392: An assertion with an invalid value is necessary, i.e. with a value that cannot be converted to int. Same for Boolean.

src/Symfony/Component/HttpFoundation/Tests/ParameterBagTest.php
```diff
@@ -378,6 +378,30 @@ public function testGetEnumThrowsExceptionWithInvalidValueType()
...
+
+        $this->assertSame(1, $bag->getIntOrNull('valid-value'));
+        $this->assertNull($bag->getIntOrNull('invalid-key'));
```
GromNaN: I would name it `missing-key` for clarity.

src/Symfony/Component/HttpFoundation/ParameterBag.php
```diff
@@ -155,6 +162,13 @@ public function getInt(string $key, int $default = 0): int
         return $this->filter($key, $default, \FILTER_VALIDATE_INT, ['flags' => \FILTER_REQUIRE_SCALAR]) ?: 0;
     }
 
+    public function getIntOrNull(string $key): ?int
+    {
+        return $this->has($key)
```
GromNaN, line 167: Checking if the value is set is not necessary:
`filter_var(null, \FILTER_VALIDATE_INT, \FILTER_NULL_ON_FAILURE)` returns `null`.

GromNaN (changes requested):
Returning `null` as default or in case of invalid value are 2 distinct features. In 7.0, an exception will be thrown when the value is incompatible with the expected return type, one may prefer this behavior over returning `null` indistinctly.

Please check issues reported by fabbot.

src/Symfony/Component/HttpFoundation/Tests/ParameterBagTest.php
mleczakm: I've changed it to using native `getInt` calls etc. so behaviour will be same in both cases.

src/Symfony/Component/HttpFoundation/Tests/ParameterBagTest.php
mleczakm: Renamed

src/Symfony/Component/HttpFoundation/ParameterBag.php
mleczakm: I've changed it to use `getInt` implementation instead.

nicolas-grekas:
I'm sorry but I don't think adding these methods is beneficial. If one wants to check if an entry exists before, using `has()` works well.

Other issues described in #45775 are solved by #48525 so we can close it IMHO.

PR #49982, opened 2023-04-08
[BrowserKit] Add proper exception hierarchy

fabpot:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Tickets       | 
| License       | MIT
| Doc PR        | n/a

Back in the day, we didn't always create exception class hierarchies for "small" component. But with time, the BrowserKit defined two of them, without the same convention as the other components. Let's fix that.

### DIFF
--- a/src/Symfony/Component/BrowserKit/Tests/AbstractBrowserTest.php
+++ b/src/Symfony/Component/BrowserKit/Tests/AbstractBrowserTest.php
@@ -14,6 +14,7 @@ namespace Symfony\Component\BrowserKit\Tests;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\BrowserKit\CookieJar;
 use Symfony\Component\BrowserKit\Exception\BadMethodCallException;
+use Symfony\Component\BrowserKit\Exception\InvalidArgumentException;
 use Symfony\Component\BrowserKit\History;
 use Symfony\Component\BrowserKit\Request;
 use Symfony\Component\BrowserKit\Response;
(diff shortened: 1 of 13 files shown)


nicolas-grekas (approved):

stof (approved):

src/Symfony/Component/BrowserKit/Tests/AbstractBrowserTest.php
```diff
@@ -294,12 +295,8 @@ public function testClickLinkNotFound()
         $client->setNextResponse(new Response('<html><a href="/foo">foobar</a></html>'));
         $client->request('GET', 'http://www.example.com/foo/foobar');
 
-        try {
-            $client->clickLink('foo');
-            $this->fail('->clickLink() throws a \InvalidArgumentException if the link could not be found');
-        } catch (\Exception $e) {
-            $this->assertInstanceOf(\InvalidArgumentException::class, $e, '->clickLink() throws a \InvalidArgumentException if the link could not be found');
-        }
+        $this->expectException(\InvalidArgumentException::class);
```
OskarStark, line 298:
```suggestion
        $this->expectException(InvalidArgumentException::class);
```
?

fabpot: No, this exception comes from the client, not browserkit.

stof: to be exact, it comes from the crawler, not from the client.

PR #41113, opened 2021-05-05
[Console] Fix Windows code page support

orkan:
| Q             | A
| ------------- | ---
| Branch?       | 5.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #37385, Fix #35842, Fix #36324, Fix #37495, Fix #37278
| License       | MIT

Corrects previous fixes that dealt with the mojibake problem on Windows where an OEM code page was applied to an input string and then messed with PHP.internal_encoding setting used by the script. This caused strings with different encodings to be displayed on the console output.

PR #49987, opened 2023-04-10
[Console] Restoring the ability to output unicode text to the Win10 console

aleksandr-shevchenko:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Restoring the ability to output unicode text to the Win10 console after corrupting the console on line 224 in symfony/console/Terminal.php

```

require_once __DIR__ . '/../vendor/autoload.php';

use App\Commands\ClearcacheCommand;
use Symfony\Component\Console\Application;

echo "abc йфяЙФЯёЁ üÜiİöÖğĞıIəƏçÇşŞ" . PHP_EOL;

$app = new Application();
$app->setName("abc йфяЙФЯёЁ üÜiİöÖğĞıIəƏçÇşŞ");
$app->setVersion("0.0.001");

$app->run();

```

before
ss_2023_04_10__13_58_30
after
ss_2023_04_10__14_00_31

### DIFF
--- a/src/Symfony/Component/Console/Terminal.php
+++ b/src/Symfony/Component/Console/Terminal.php
@@ -157,6 +157,8 @@ class Terminal
             2 => ['pipe', 'w'],
         ];
 
+        $cp = \function_exists('sapi_windows_cp_set') ? sapi_windows_cp_get() : 0;
+
         $process = proc_open($command, $descriptorspec, $pipes, null, null, ['suppress_errors' => true]);
         if (!\is_resource($process)) {
             return null;
@@ -166,6 +168,10 @@ class Terminal
         fclose($pipes[1]);
         fclose($pipes[2]);
         proc_close($process);
+
+        if ($cp) {
+            sapi_windows_cp_set($cp);
+        }
 
         return $info;
     }


nicolas-grekas:
Why do you say "restore"? Does the call change some state? If yes, can't we restore the previous cp instead? Aka use `sapi_windows_cp_get()` before and restore after, instead of hardcoding 65001?

aleksandr-shevchenko:
> Why do you say "restore"? Does the call change some state? If yes, can't we restore the previous cp instead? Aka use `sapi_windows_cp_get()` before and restore after, instead of hardcoding 65001?

```
    private static function readFromProcess(string|array $command): ?string
    {
        if (!\function_exists('proc_open')) {
            return null;
        }

        $descriptorspec = [
            1 => ['pipe', 'w'],
            2 => ['pipe', 'w'],
        ];

        $process = proc_open($command, $descriptorspec, $pipes, null, null, ['suppress_errors' => true]);
        if (!\is_resource($process)) {
            return null;
        }

        $info = stream_get_contents($pipes[1]);
        fclose($pipes[1]);
        fclose($pipes[2]);
        proc_close($process);

        if (\function_exists('sapi_windows_cp_set')) {
            sapi_windows_cp_set(65001);
        }

        return $info;
    }

```
In this function, before line `$info = stream_get_contents($pipes[1]);` echo can print unicode text to Win10 console, but after that line - not.
I don't understand why the console gets corrupted, but my solution works. What's wrong with hardcoding UTF-8(65001)?


```
        if (\function_exists('sapi_windows_cp_get')) {
            $cp1 = sapi_windows_cp_get();
            echo $cp1 . " abc йфяЙФЯёЁ üÜiİöÖğĞıIəƏçÇşŞ" . PHP_EOL;
        }

        $info = stream_get_contents($pipes[1]);

        if (\function_exists('sapi_windows_cp_get')) {
            $cp2 = sapi_windows_cp_get();
            echo $cp2 . " abc йфяЙФЯёЁ üÜiİöÖğĞıIəƏçÇşŞ" . PHP_EOL;
        }
        echo $info;

```
ss_2023_04_11__08_42_41

nicolas-grekas:
Strange :)

> What's wrong with hardcoding UTF-8(65001)?

hardcoding is always an issue, especially when there's a way not to hardcode. Here, this change would break consoles that rely on another CP. Wouldn't my previous suggestion work?

aleksandr-shevchenko:
> Wouldn't my previous suggestion work?

Also works well (and without hardcoding 65001).

```
        if (\function_exists('sapi_windows_cp_get')) {
            $cp = sapi_windows_cp_get();
            echo $cp . " abc йфяЙФЯёЁ üÜiİöÖğĞıIəƏçÇşŞ" . PHP_EOL;
        }

        $info = stream_get_contents($pipes[1]);

        if (\function_exists('sapi_windows_cp_get') && \function_exists('sapi_windows_cp_set')) {
            sapi_windows_cp_set($cp);
            echo $cp . " abc йфяЙФЯёЁ üÜiİöÖğĞıIəƏçÇşŞ" . PHP_EOL;
        }
        echo $info;

```
ss_2023_04_11__14_23_25

chalasr:
We should probably do the same as https://github.com/symfony/symfony/pull/41113 (or even reuse if possible). 
Also this qualifies as a bugfix, isn't it? Please fill in the PR template you deleted when opening (you can find it in other PRs).

nicolas-grekas:
Thank you @aleksandr-shevchenko.

PR #49992, opened 2023-04-10
[Mailer] [Mailjet] Use body MessageID instead of X-MJ-Request-GUID

### LINKED ISSUE #48550: [Mailer][Mailjet] MailjetApiTransport SentMessage MessageId is Request ID

5.4

### Description

The ID assigned using $sentMessage->setMessageId is an ID generated by MailjetAPI for the request but not for the message. If you want to identify messages on webhooks (e.g. bounce or block see https://dev.mailjet.com/email/guides/webhooks/ ) it would not work, you need the MessageUUID

I got this feedback from their support:

> Yes, correct, these values are of the same format, but are assigned to different things.
The MessageUUID is the ID of the message. 
The x-mj-request-guid is the ID of the API request. This is solely saved for us, for logging purposes. 

Problem is in case there is more than one Recipient more than one MessageUUID is generated and SentMessage can only store a single Id for a message.

https://dev.mailjet.com/email/guides/send-api-v31/#send-a-basic-email 

### How to reproduce

Send a message using the MailjetApiTransport

### Possible Solution

If you need to identify email messages send them as custom headers 
`$email->getHeaders()->addHeader('MessageID', $this->messageId);`
and rely on the value coming back on webhook calls

### Additional Context

I am not sure if bug is the right label for this, maybe this is just something to add to the documentation.

Starfox64:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #48550
| License       | MIT
| Doc PR        | N/A

As per #48550, the message ID of the `SentMessage` returned by the `MailjetApiTransport` is the `X-MJ-Request-GUID`. The issue is that this ID is only there for debug purposes by the MailJet engineers and cannot be consumed by the MailJet API to query the message metadata.

This change instead uses the `MessageID` attribute returned in the JSON body of the response which can be used in further API requests.

As the existing message ID was not usable for anything besides opening support tickets (which the new ID can still be used for), I've decided to submit this PR as a bug fix.

A slight limitation however is that MailJet will return multiple messages when sending to multiple recipients, each having their own `MessageID`. This implementation therefore only returns the first `MessageID`, there is an ongoing discussion on #48550 about that behavior but I am submitting this PR as is since it's still a net improvement over the previous ID.

### DIFF
--- a/src/Symfony/Component/Mailer/Bridge/Mailjet/Tests/Transport/MailjetApiTransportTest.php
+++ b/src/Symfony/Component/Mailer/Bridge/Mailjet/Tests/Transport/MailjetApiTransportTest.php
@@ -130,7 +130,7 @@ class MailjetApiTransportTest extends TestCase
 
         $sentMessage = $transport->send($email);
         $this->assertInstanceOf(SentMessage::class, $sentMessage);
-        $this->assertSame('576460756513665524', $sentMessage->getMessageId());
+        $this->assertSame('576460756513665525', $sentMessage->getMessageId());
     }
 
     public function testSendWithDecodingException()
--- a/src/Symfony/Component/Mailer/Bridge/Mailjet/Tests/Transport/MailjetApiTransportTest.php
+++ b/src/Symfony/Component/Mailer/Bridge/Mailjet/Tests/Transport/MailjetApiTransportTest.php
@@ -108,7 +108,7 @@ class MailjetApiTransportTest extends TestCase
                         [
                             'Email' => 'passenger1@mailjet.com',
                             'MessageUUID' => '7c5f9f29-42ba-4959-b19c-dcd8b2f327ca',
-                            'MessageID' => 576460756513665525,
+                            'MessageID' => '576460756513665525',
                             'MessageHref' => 'https://api.mailjet.com/v3/message/576460756513665525',
                         ],
                     ],
--- a/src/Symfony/Component/Mailer/Bridge/Mailjet/Tests/Transport/MailjetApiTransportTest.php
+++ b/src/Symfony/Component/Mailer/Bridge/Mailjet/Tests/Transport/MailjetApiTransportTest.php
@@ -109,10 +109,10 @@ class MailjetApiTransportTest extends TestCase
                             'Email' => 'passenger1@mailjet.com',
                             'MessageUUID' => '7c5f9f29-42ba-4959-b19c-dcd8b2f327ca',
                             'MessageID' => 576460756513665525,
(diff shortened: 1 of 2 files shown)


Starfox64:
Will submit updated tests tomorrow, seems the dummy response's `Messages` field is missing a real message's data.

Starfox64:
AppVeyor is failing because it does not seem to support 64 bit integers. Is this a requirement for 5.4 or any of the later Symfony versions?

src/Symfony/Component/Mailer/Bridge/Mailjet/Tests/Transport/MailjetApiTransportTest.php
```diff
@@ -102,15 +102,21 @@ public function testSendSuccess()
...
         $json = json_encode([
             'Messages' => [
-                'foo' => 'bar',
+                [
+                    'Status' => 'success',
+                    'To' => [
+                        [
+                            'Email' => 'passenger1@mailjet.com',
+                            'MessageUUID' => '7c5f9f29-42ba-4959-b19c-dcd8b2f327ca',
+                            'MessageID' => 576460756513665525,
```
nicolas-grekas: yes, this test should be made to work on 32b. We decode with the  JSON_BIGINT_AS_STRING flag so I guess this should be turned into a string here

Starfox64:
AppVeyor appears to be failing for unrelated reasons now (Form component).

If this is all good let me know so I can squash this branch since I've forked into our org.

src/Symfony/Component/Mailer/Bridge/Mailjet/Transport/MailjetApiTransport.php
```diff
@@ -85,7 +85,7 @@ protected function doSendApi(SentMessage $sentMessage, Email $email, Envelope $e
             throw new HttpTransportException(sprintf('Unable to send an email: "%s" malformed api response.', $response->getContent(false)), $response);
         }
 
-        $sentMessage->setMessageId($response->getHeaders(false)['x-mj-request-guid'][0]);
+        $sentMessage->setMessageId((string) $result['Messages'][0]['To'][0]['MessageID'] ?? '');
```
nicolas-grekas: I don't think the cast to string is correct as it applies before the `??`
Also why is it needed at all?

Starfox64: That was a remnant from earlier, since HttpClient uses JSON_BIGINT_AS_STRING, it's no longer needed.

nicolas-grekas:
Thank you @Starfox64.

PR #48044, opened 2022-10-29
[SecurityBundle] Set request stateless when firewall is stateless

alamirault:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Tickets       | Fix #40372 
| License       | MIT
| Doc PR        | symfony/symfony-docs#... 

Automatically add `_stateless` attribute to the request when firewall is stateless

PR #49997, opened 2023-04-11
[SecurityBundle] Set request stateless only if the attribute is not defined

tucksaun:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes-ish
| New feature?  | no
| Deprecations? | no
| Tickets       | https://github.com/symfony/symfony/pull/48044#issuecomment-1504267293
| License       | MIT
| Doc PR        | n/a

The current implementation makes sense for most cases but not for every case as one can have a stateless authentication but still requires sessions.
This PR allows setting the request as non-stateless while having a stateless firewall but keeping the new behavior by default.

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/CHANGELOG.md
+++ b/src/Symfony/Bundle/SecurityBundle/CHANGELOG.md
@@ -5,7 +5,7 @@ CHANGELOG
 ---
 
  * Deprecate enabling bundle and not configuring it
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
Thank you @tucksaun.

alamirault:
Thanks @tucksaun, I missed this case !

PR #50002, opened 2023-04-12
[VarDumper] Add missing return types

alexandre-lass:
| Q             | A
| ------------- | ---
| Branch?       | 6.3 
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no 
| Tickets       | Contribution for #47551 
| License       | MIT
| Doc PR        |

Contribution on the issue #47551 concerning the **VarDumper** component.
The type is added on PHPDoc when BC risk is present, else native type is used (private methods / tests).

### DIFF
--- a/src/Symfony/Component/VarDumper/Caster/ClassStub.php
+++ b/src/Symfony/Component/VarDumper/Caster/ClassStub.php
@@ -85,6 +85,9 @@ class ClassStub extends ConstStub
         }
     }
 
+    /**
+     * @return mixed
+     */
     public static function wrapCallable(mixed $callable)
     {
         if (\is_object($callable) || !\is_callable($callable)) {
--- a/src/Symfony/Component/VarDumper/Cloner/Data.php
+++ b/src/Symfony/Component/VarDumper/Cloner/Data.php
@@ -414,7 +414,7 @@ class Data implements \ArrayAccess, \Countable, \IteratorAggregate
         return $hashCut;
     }
 
-    private function getStub(mixed $item)
+    private function getStub(mixed $item): mixed
     {
         if (!$item || !\is_array($item)) {
             return $item;
--- a/src/Symfony/Component/VarDumper/Dumper/DataDumperInterface.php
+++ b/src/Symfony/Component/VarDumper/Dumper/DataDumperInterface.php
@@ -20,5 +20,9 @@ use Symfony\Component\VarDumper\Cloner\Data;
  */
 interface DataDumperInterface
 {
+
+    /**
+     * @return string|null
+     */
     public function dump(Data $data);
 }
--- a/.github/expected-missing-return-types.diff
+++ b/.github/expected-missing-return-types.diff
@@ -648,10 +648,10 @@ index 6e7669a710..27517d34ae 100644
      {
          if (!$container->hasDefinition('test.private_services_locator')) {
 diff --git a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/UnusedTagsPass.php b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/UnusedTagsPass.php
-index 7fa0fb2890..49bcc6dfd9 100644
+index 9da5b91bb3..d7acf8040b 100644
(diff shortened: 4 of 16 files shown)


src/Symfony/Component/VarDumper/Caster/ClassStub.php
```diff
@@ -85,6 +85,9 @@ public function __construct(string $identifier, callable|array|string $callable
         }
     }
 
+    /**
+     * @return callable|mixed|static
```
stof: This is actually `@return mixed` as other types are subtypes of mixed

src/Symfony/Component/VarDumper/Cloner/Data.php
```diff
@@ -414,12 +414,14 @@ private function dumpChildren(DumperInterface $dumper, Cursor $parentCursor, arr
         return $hashCut;
     }
 
-    private function getStub(mixed $item)
+    /**
+     * @return mixed|Stub
```
stof: this is still meaning `mixed`, so there is no benefit to this phpdoc

src/Symfony/Component/VarDumper/Cloner/Data.php
```diff
@@ -414,12 +414,14 @@ private function dumpChildren(DumperInterface $dumper, Cursor $parentCursor, arr
...
+     */
+    private function getStub(mixed $item): mixed
     {
         if (!$item || !\is_array($item)) {
             return $item;
         }
-
```
stof: please don't remove this empty line

src/Symfony/Component/VarDumper/Dumper/DataDumperInterface.php
```diff
@@ -20,5 +20,9 @@
  */
 interface DataDumperInterface
 {
+
+    /**
+     * @return void|null|string
```
stof: `void` does not make sense in a union type (that would be forbidden when migrating to native types)

alexandre-lass:
@stof i updated the branch with your feedback, thanks

src/Symfony/Component/VarDumper/Dumper/DataDumperInterface.php
alexandre-lass: We got a small problem on this type, all implementation return void, except the `\Symfony\Component\VarDumper\Dumper\AbstractDumper::dump` who return `null|string` (it return the output giving a bool parameter)
If we remove the `void` from PHPDoc return type, we get an error for all dumpers returning nothing (eg: https://github.com/symfony/symfony/actions/runs/4679151738/jobs/8288795339?pr=50002)

I think that we need to make the AbstractDumper::dump method return type as same of other (void), and maybe add a method to get the output (but this add BC changes)

What do you think @stof ?

stof: To me, we should return `string|null`.

For our 2 dumpers that don't extend AbstractDumper:
- ContextualizedDumper should return the return value of its wrapped dumper (not doing so is actually a bug IMO)
- ServerDumper should always return null (which won't be a BC break as a caller trying to use the returned value would get `null` in both cases.

stof:
What are those `phpunit.stdout` files that you committed ?

src/Symfony/Component/VarDumper/Dumper/DataDumperInterface.php
alexandre-lass: Ok i get it, i pushed the change

stof (approved):

alexandre-lass:
> What are those `phpunit.stdout` files that you committed ?

I pushed it by error, it's now clear

alexandre-lass:
@stof I added a commit for updating the `expected-missing-return-types.diff` file

alexandre-lass:
@stof i see an error on `src/Symfony/Component/HttpKernel/DataCollector/DumpDataCollector::dump` who is returning void and implementing `DataDumperInterface`.
Do you think i can safely update his return type ?

nicolas-grekas:
Thank you @alexandre-lass.

PR #50004, opened 2023-04-12
[HttpClient] fix proxied redirects in curl client

matthi4s:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

When using the `CurlHttpClient` with a proxy defined using the options array, the proxy isn't used after the first request on redirects. The `CURLOPT_PROXY` option is overwritten in the redirect resolver, but the `$options` variable isn't present in that case which causes it to be overwritten with environment variables or nothing, which disables the proxy.

I don't know why the proxy option gets rewritten with every redirect, since the same curl handle is used, it should not be necessary to set it again. It also wasn't set before the original commit which introduced this bug (https://github.com/symfony/symfony/commit/9e5305e68aaf022ba24a7da10318716335205a46). I've removed that part, but I can also change it to pass on the reference to the `$options` variable, which also fixes the issue. In that case, I would suggest moving the validation of the environment variable to a separate function to avoid duplicating that part.

I've also added a test case that fails for the curl client without the fix but succeeds for the other clients and the curl client with both possible fixes.

### DIFF
--- a/src/Symfony/Component/HttpClient/CurlHttpClient.php
+++ b/src/Symfony/Component/HttpClient/CurlHttpClient.php
@@ -95,10 +95,7 @@ final class CurlHttpClient implements HttpClientInterface, LoggerAwareInterface,
         $scheme = $url['scheme'];
         $authority = $url['authority'];
         $host = parse_url($authority, \PHP_URL_HOST);
-        $proxy = $options['proxy']
-            ?? ('https:' === $url['scheme'] ? $_SERVER['https_proxy'] ?? $_SERVER['HTTPS_PROXY'] ?? null : null)
-            // Ignore HTTP_PROXY except on the CLI to work around httpoxy set of vulnerabilities
-            ?? $_SERVER['http_proxy'] ?? (\in_array(\PHP_SAPI, ['cli', 'phpdbg'], true) ? $_SERVER['HTTP_PROXY'] ?? null : null) ?? $_SERVER['all_proxy'] ?? $_SERVER['ALL_PROXY'] ?? null;
+        $proxy = self::getProxyUrl($options['proxy'], $url);
         $url = implode('', $url);
 
         if (!isset($options['normalized_headers']['user-agent'])) {
@@ -411,7 +408,7 @@ final class CurlHttpClient implements HttpClientInterface, LoggerAwareInterface,
             }
         }
 
-        return static function ($ch, string $location, bool $noContent) use (&$redirectHeaders) {
+        return static function ($ch, string $location, bool $noContent) use (&$redirectHeaders, $options) {
             try {
                 $location = self::parseUrl($location);
             } catch (InvalidArgumentException $e) {
@@ -436,11 +433,7 @@ final class CurlHttpClient implements HttpClientInterface, LoggerAwareInterface,
             $url = self::parseUrl(curl_getinfo($ch, \CURLINFO_EFFECTIVE_URL));
             $url = self::resolveUrl($location, $url);
 
-            curl_setopt($ch, \CURLOPT_PROXY, $options['proxy']
-                ?? ('https:' === $url['scheme'] ? $_SERVER['https_proxy'] ?? $_SERVER['HTTPS_PROXY'] ?? null : null)
-                // Ignore HTTP_PROXY except on the CLI to work around httpoxy set of vulnerabilities
-                ?? $_SERVER['http_proxy'] ?? (\in_array(\PHP_SAPI, ['cli', 'phpdbg'], true) ? $_SERVER['HTTP_PROXY'] ?? null : null) ?? $_SERVER['all_proxy'] ?? $_SERVER['ALL_PROXY'] ?? null
-            );
+            curl_setopt($ch, \CURLOPT_PROXY, self::getProxyUrl($options['proxy'], $url));
 
             return implode('', $url);
         };
--- a/src/Symfony/Component/HttpClient/HttpClientTrait.php
+++ b/src/Symfony/Component/HttpClient/HttpClientTrait.php
@@ -655,16 +655,7 @@ trait HttpClientTrait
      */
     private static function getProxy(?string $proxy, array $url, ?string $noProxy): ?array
     {
-        if (null === $proxy) {
-            // Ignore HTTP_PROXY except on the CLI to work around httpoxy set of vulnerabilities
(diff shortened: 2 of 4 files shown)


src/Symfony/Component/HttpClient/CurlHttpClient.php
```diff
@@ -436,12 +436,6 @@ private static function createRedirectResolver(array $options, string $host): \C
             $url = self::parseUrl(curl_getinfo($ch, \CURLINFO_EFFECTIVE_URL));
             $url = self::resolveUrl($location, $url);
 
-            curl_setopt($ch, \CURLOPT_PROXY, $options['proxy']
-                ?? ('https:' === $url['scheme'] ? $_SERVER['https_proxy'] ?? $_SERVER['HTTPS_PROXY'] ?? null : null)
```
nicolas-grekas, line 440: the reason why we have this code is to make the proxy URL-sensitive, so we should keep it and instead ensure the option is passed up to this point

matthi4s: That makes sense, I've just seen that `HttpClientTrait` has a `getProxy()` function. Is there any reason not to use that function in the `CurlHttpClient`? I would change both usages and make sure that the $options are passed on correctly.

nicolas-grekas: I'm not sure, I'll let you check if there is any diff in behavior :)

src/Symfony/Component/HttpClient/CurlHttpClient.php
```diff
@@ -95,10 +95,7 @@ public function request(string $method, string $url, array $options = []): Respo
         $scheme = $url['scheme'];
         $authority = $url['authority'];
         $host = parse_url($authority, \PHP_URL_HOST);
-        $proxy = $options['proxy']
-            ?? ('https:' === $url['scheme'] ? $_SERVER['https_proxy'] ?? $_SERVER['HTTPS_PROXY'] ?? null : null)
-            // Ignore HTTP_PROXY except on the CLI to work around httpoxy set of vulnerabilities
-            ?? $_SERVER['http_proxy'] ?? (\in_array(\PHP_SAPI, ['cli', 'phpdbg'], true) ? $_SERVER['HTTP_PROXY'] ?? null : null) ?? $_SERVER['all_proxy'] ?? $_SERVER['ALL_PROXY'] ?? null;
+        $proxy = self::getProxyURL($options['proxy'], $url);
```
nicolas-grekas:
```suggestion
        $proxy = self::getProxyUrl($options['proxy'], $url);
```

src/Symfony/Component/HttpClient/CurlHttpClient.php
```diff
@@ -411,7 +408,7 @@ private static function createRedirectResolver(array $options, string $host): \C
             }
         }
 
-        return static function ($ch, string $location, bool $noContent) use (&$redirectHeaders) {
+        return static function ($ch, string $location, bool $noContent) use (&$redirectHeaders, &$options) {
```
nicolas-grekas:
```suggestion
        return static function ($ch, string $location, bool $noContent) use (&$redirectHeaders, $options) {
```

src/Symfony/Component/HttpClient/CurlHttpClient.php
```diff
@@ -436,11 +433,7 @@ private static function createRedirectResolver(array $options, string $host): \C
             $url = self::parseUrl(curl_getinfo($ch, \CURLINFO_EFFECTIVE_URL));
             $url = self::resolveUrl($location, $url);
 
-            curl_setopt($ch, \CURLOPT_PROXY, $options['proxy']
-                ?? ('https:' === $url['scheme'] ? $_SERVER['https_proxy'] ?? $_SERVER['HTTPS_PROXY'] ?? null : null)
-                // Ignore HTTP_PROXY except on the CLI to work around httpoxy set of vulnerabilities
-                ?? $_SERVER['http_proxy'] ?? (\in_array(\PHP_SAPI, ['cli', 'phpdbg'], true) ? $_SERVER['HTTP_PROXY'] ?? null : null) ?? $_SERVER['all_proxy'] ?? $_SERVER['ALL_PROXY'] ?? null
-            );
+            curl_setopt($ch, \CURLOPT_PROXY, self::getProxyURL($options['proxy'], $url));
```
nicolas-grekas:
```suggestion
            curl_setopt($ch, \CURLOPT_PROXY, self::getProxyUrl($options['proxy'], $url));
```

src/Symfony/Component/HttpClient/HttpClientTrait.php
```diff
@@ -655,14 +655,7 @@ private static function mergeQueryString(?string $queryString, array $queryArray
...
-            // Ignore HTTP_PROXY except on the CLI to work around httpoxy set of vulnerabilities
-            $proxy = $_SERVER['http_proxy'] ?? (\in_array(\PHP_SAPI, ['cli', 'phpdbg'], true) ? $_SERVER['HTTP_PROXY'] ?? null : null) ?? $_SERVER['all_proxy'] ?? $_SERVER['ALL_PROXY'] ?? null;
-
-            if ('https:' === $url['scheme']) {
-                $proxy = $_SERVER['https_proxy'] ?? $_SERVER['HTTPS_PROXY'] ?? $proxy;
-            }
-        }
+        $proxy = self::getProxyURL($proxy, $url);
 
         if (null === $proxy) {
```
nicolas-grekas:
```suggestion
        if (null === $proxy = self::getProxyURL($proxy, $url)) {
```

src/Symfony/Component/HttpClient/HttpClientTrait.php
```diff
@@ -692,6 +685,22 @@ private static function getProxy(?string $proxy, array $url, ?string $noProxy):
         ];
     }
 
+    private static function getProxyURL(?string $proxy, array $url): ?string
```
nicolas-grekas:
```suggestion
    private static function getProxyUrl(?string $proxy, array $url): ?string
```

nicolas-grekas (approved):
LGTM thanks. Here are some nitpicking. Failures are false positives.

fabpot (approved):

fabpot:
Thank you @matthi4s.

PR #50005, opened 2023-04-12
[Messenger] Amazon SQS transport can handle its own retry/DLQ

### LINKED ISSUE #45104: Symfony SQS Transport not compatible with SQS-DLQs

5.4.*

### Description + Reproduction

Given the following system:
 * 1: A SQS queue that will be configured as a Dead-Letter-Queue
 * 2: A SQS queue that a symfony consumer should consume from, configured with 1) as DLQ 
 * 3: A symfony transport consumer configured to read from the queue from 2. It's AWS permissions are the minimal permissions of a queue consumer:

```
{
            "Effect": "Allow",
            "Action": [
                "sqs:ReceiveMessage",
                "sqs:ChangeMessageVisibility",
                "sqs:GetQueueUrl",
                "sqs:DeleteMessage",
                "sqs:GetQueueAttributes"
            ],
            "Resource": "arn:aws:sqs:eu-central-1:1111111111:foobarqueue"
        }
   
```

If I have a message that causes an exception to be thrown on the consumer and can't be processed, according to AWS docs, the consumer should not call `DeleteMessage`, but instead simply leave the message in the queue. After a configured VisibilityTimeout (https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html) the message will be made visible again and the consumer will try to process it again. This will happen for as long until the configured maximum amount of receives is reached and then the message is moved to the DLQ.

Symfony's consumer behaviour breaks with this assumption by orchestrating the retry logic on the consumer side, which isn't great in multiple ways:
 * A: The consumer also needs to be able to send messages into the queue.
 * B: it breaks with the idea of SQS DLQs as the receives of the message in SQS won't increase.
 * C: If the consumer can't send the message for any reason (permissions, errors, networking issues, process crashes etc.) the message might be lost.
 
I am wondering if the transport implementation can be configured in a way that makes it compatible again with what the expected usage of SQS is, so that DLQs will work properly again. If not, how does the implementation need to change to make the symfony consumer behaviour compatible with SQS DLQs again?
Could there be an Exception type that would cause the message to simply not be deleted and re-added to the queue so that we'll fall back to SQS' model of retries?

### Possible Solution

_No response_

### Additional Context

_No response_

alex-dev:
| Q | A
|---|---
| Branch? | 6.3
| Bug fix? | no
| New feature? | yes
| Deprecations? | no
| BC breaks?  | no
| Tickets | Fix #45104
| License | MIT
| Doc PR | symfony/symfony-docs#NA

- [ ] Update CHANGELOG.md
- [x] Update documentation judged relevant

## Problem
Amazon SQS could handle its own failures. It may be app responsability to not configure `failure_transport`. But if Symfony and SQS merely remove every rejected messages from SQS, it is useless.

## Solution
`AmazonSqsReceiver::reject()` should only increase SQS visibility timeout to a flat n seconds. This behavior is toggleable by configuration/DSN `delete_on_rejection=false`.

## Changes
- Split `Connection` tests in multiple classes for ease of maintaining
- Added tests for `Connection::send`
- Split `Connection::delete` into `Connection::ack` and `Connection::reject`
- `Connection::reject` can either delete or not the message depending on `delete_on_rejection`

## Alternatives
Solving this problem at a higher level would have required a larger refactor in how message retry are handled by Symfony. Current solution allows to merely disable failure transports and use the correct DSN option to delegate retry and DLQ back to SQS.


alex-dev:
Considering there is not much DSN documentation for messenger, I didn't provide documentation.

alex-dev:
`continuous-integration/appveyor/pr` seems to fail in untouched code... In actual integration tests with Redis....

alex-dev:
@fabpot fix and close #45104

alexbaileyuk:
Something that I've noticed locally here (still running 5.x here atm).

Let's say you have a transport that looks like this:

```
            defaults: &messenger-defaults
                dsn: 'https://sqs.eu-west-1.amazonaws.com/12345678910/%environment_name%-default'
                retry_strategy:
                    max_retries: 3
                    delay: 30000
                options:
                    auto_setup: false
                    poll_timeout: 10
                    delete_on_rejection: false
```

What seems to happen is:
- Message pulled from queue
- Exception during handling message
- Message is sent back to the queue (`SendFailedMessageForRetryListener::onMessageFailed` -> `TransportInterface::send`)
- Message is rejected (`TransportInterface::reject`)

So message gets duplicated to queue and then the original messages visibility timeout is set to `$this->configuration['visibility_timeout'] ?? 30` so the original also goes back to the queue.

The more I dig into this the more I find will not work with messenger functionality. For example, the delay strategy and retry strategies from Symfony will not work with this change. Is this going to be explicit in the documentation or blocked somehow (maybe if someone specifies `max_retries` or `delay` then an error should be thrown?

Edit: this would actually cause more issues. For example, there would be no new message stamps like the `AmazonSqsReceivedStamp` etc added because the message is not modified :(

alex-dev:
The intent is to give SQS full control over retry and DLQ... Yes it is incompatible with using Messenger built in retry features. That's the whole goal! Currently, it is impossible to opt out of them because the transport will still delete the message on failures.
Symfony pattern to handle errors doesn't work with SQS at all:
- FIFO queue are broken (failure causes out of order execution)
- Any monitoring on queues is faulty since Symfony can write to a queue it only read.
- Symfony stamps on errors may be so large they exceed SQS limits, which causes Symfony to silently delete the message.

TLDR: while Messenger whole retry flow is good enough for some transport, it is invalid for others. A full rework of it would be an undertaking that would be much more involved and difficult than a simple opt out on Symfony deleting messages.

alexbaileyuk:
Would utilizing multiple failure transports make more sense here than essentially blocking 90% of the components functionality? 

I totally get what you mean. SQS isn't really all that compatible with Symfony Messenger. I wonder though what will happen if documentation doesn't describe all these limitations. For example, lots of middleware etc might just not work.

I feel like at the very least documentation needs to be explicit about these shortcomings.

alex-dev:
What would make more sense would be rewriting transport layer to handle failure inside a the same transport as failure if possible, if not, rely on a second transport. Then you can take full control....

alexbaileyuk:
:thinking: I guess so. For fifo it's probably the only way. For standard queues, perhaps it would be fine if Symfony was just aware of the DLQ. Then I guess metrics in AWS would be incorrect for messages sent to queue etc.

It'd make more sense to me here to have a setting like "delegate_retry_to_transport" or something that totally disables a bunch of options. I'd even ask if it's a totally different transport.

I don't know the best option here. They all kinda suck for integrating messenger to SQS in the current form of both services. Unless AWS allows you to update a message (unlikely) or messenger is re-engineered to allow this kind of delegation of retry functionality (accepting everything you lose) then nothing is great here.

Given all the drawbacks, caveats etc with the integration, should Symfony event provide an offocial integration?

alexbaileyuk:
@alex-dev I think I found another problem with this implementation today with `poll_timeout` options.

- Symfony Messenger begins consuming on a queue with 4 messages in it and a 10 second poll time with `--limit 10` to restrict processing to 10 messages.
- After pulling the first 4 messages, 10 are added to the queue (while the 4 are being processed).
- The four finish processing and Symfony pulls another 10.
- Symfony processes 6 and then drops the rest (`AmazonSqsReceiver` does not implement `ResetInterface` so it just drops them.

If that happens to the same message _n_ times where _n_ is the `maximum_receives` value set on the DLQ policy then they will end up in the SQS DLQ without ever being processed.

I think before the same thing may have happened? Perhaps that's a separate issue.

alex-dev:
@alexbaileyuk  I don't see how this would happen. Unless by drop you mean the worker exits....

In this case, that's already an issue. Symfony, on close, with 100 workers, on a shutdown, will consume 100s of messages just to throw them away.

alexbaileyuk:
Yes I mean say worker has --limit 5 option but pulls 10 SQS messages in a batch. It could process 5 and then exit.

I'm going to leave this to someone else but I think that allowing SQS to manage the queue would cause problems and we should accept that Symfony is taking over a bunch of SQS functionality in this integration. Otherwise it's going to need lists of caveats.

alex-dev:
Current worker batching for SQS is already worrying....
- In a large deployment (100+ workers), if you shutdown all workers, they will quickly pull 100s of message to not process them.
- Batching is private to the `Connection` while Symfony is designed to expose batch through `TransportInterface`.

Symfony doesn't integrate with SQS well currently. I dislike my solution.... But a significant refactor to Messenger worker as whole is not something I can do on company time without guarantee thee will be interest to and support for it.

PR #50010, opened 2023-04-13
[Serializer] Add missing return types

alexandre-lass:
| Q             | A
| ------------- | ---
| Branch?       | 6.3 
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no 
| Tickets       | Contribution for #47551 
| License       | MIT
| Doc PR        |

Contribution on the issue #47551 concerning the **Serializer** component.
The type is added on PHPDoc when BC risk is present, else native type is used (private methods / tests).

### DIFF
--- a/src/Symfony/Component/Serializer/Mapping/AttributeMetadataInterface.php
+++ b/src/Symfony/Component/Serializer/Mapping/AttributeMetadataInterface.php
@@ -32,7 +32,7 @@ interface AttributeMetadataInterface
     /**
      * Adds this attribute to the given group.
      */
-    public function addGroup(string $group);
+    public function addGroup(string $group): void;
 
     /**
      * Gets groups of this attribute.
@@ -44,7 +44,7 @@ interface AttributeMetadataInterface
     /**
      * Sets the serialization max depth for this attribute.
      */
-    public function setMaxDepth(?int $maxDepth);
+    public function setMaxDepth(?int $maxDepth): void;
 
     /**
      * Gets the serialization max depth for this attribute.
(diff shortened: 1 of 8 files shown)


nicolas-grekas (commented):
We should revert the changes on fixtures unless they are required by a base interface. The first reason is that it might very well be possible that a test case is written to cover the behavior when no return type is specified. The second reason is that adding return types to fixtures doesn't improve anything :)

src/Symfony/Component/Serializer/Mapping/AttributeMetadataInterface.php
```diff
@@ -31,6 +31,8 @@ public function getName(): string;
 
     /**
      * Adds this attribute to the given group.
+     *
+     * @return void
      */
     public function addGroup(string $group);
```
wouterj: This interface is internal. @nicolas-grekas do you also not want to use the real void type for safe methods?

wouterj (approved):

src/Symfony/Component/Serializer/Mapping/AttributeMetadataInterface.php
nicolas-grekas: if it's internal, we definitely can add the real return type now

alexandre-lass: I updated the code to use native void return types on internal interfaces

nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @alexandre-lass.

PR #50013, opened 2023-04-13
[HttpKernel] Don't use eval() to render ESI/SSI

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Yes, this would not be needed if people were able to use a real `kernel.secret`. But apparently, many fail.
And it looks like we can do something about it, so let's makes this impossible to exploit :)

### DIFF
--- a/src/Symfony/Component/HttpKernel/HttpCache/Esi.php
+++ b/src/Symfony/Component/HttpKernel/HttpCache/Esi.php
@@ -74,8 +74,10 @@ class Esi extends AbstractSurrogate
         $content = preg_replace('#<esi\:remove>.*?</esi\:remove>#s', '', $content);
         $content = preg_replace('#<esi\:comment[^>]+>#s', '', $content);
 
+        static $cookie;
+        $cookie = hash('xxh128', $cookie ??= random_bytes(16), true);
+        $boundary = base64_encode($cookie);
         $chunks = preg_split('#<esi\:include\s+(.*?)\s*(?:/|</esi\:include)>#', $content, -1, \PREG_SPLIT_DELIM_CAPTURE);
-        $chunks[0] = str_replace($this->phpEscapeMap[0], $this->phpEscapeMap[1], $chunks[0]);
 
         $i = 1;
         while (isset($chunks[$i])) {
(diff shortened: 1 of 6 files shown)


src/Symfony/Component/HttpKernel/HttpCache/Esi.php
```diff
@@ -89,16 +88,12 @@ public function process(Request $request, Response $response): Response
...
-                isset($options['onerror']) && 'continue' === $options['onerror'] ? 'true' : 'false'
-            );
-            ++$i;
-            $chunks[$i] = str_replace($this->phpEscapeMap[0], $this->phpEscapeMap[1], $chunks[$i]);
-            ++$i;
+            $chunks[$i] = $options['src']."\n".($options['alt'] ?? '')."\n".('continue' === ($options['onerror'] ?? ''));
+            $i += 2;
         }
-        $content = implode('', $chunks);
+        static $boundary;
```
stof: shouldn't this use an instance variable instead, so that separate instances can have separate boundaries ? Or maybe a normal local variable so that each processed response has a separate boundary.

And shouldn't we escape the boundary itself in case it appears in the content ?

nicolas-grekas: the boundary is random enough to be reused and to ignore escaping
multipart mime messages use the same strategy and ignore collisions,
we can safely make the same assumptions IMHO

stof: if you reuse it (especially in a long-running app server), you will lose the benefit of the randomness, potentially allowing to craft an attack by injecting the boundary in the output (as such setup would reuse the boundary long enough to try to guess it)

nicolas-grekas: alright, updated to rotate it

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

fritzmg:
> Yes, this would not be needed if people were able to use a real kernel.secret. But apparently, many fail.

imho the issue is that it is not actually mandatory at all. Even the `UriSigner` allows an empty secret and in the `Configuration` tree of the `FrameworkBundle`, `framework.secret` is allowed to be anything. Currently only `symfony/flex` will automatically generate an `APP_SECRET` for you when you first set up the project. But that does not help if you either do not setup your project via `symfony/flex` or when you deploy the website to the live environment.

In any case, will this change also be back ported to Symfony 5.4?

alexisLefebvre:
@fritzmg I can't find the relevant page in the doc but backports are not applied in the Symfony ecosystem. This new way to render is a feature, it can't land on a previous release.

fritzmg:
But it's a security issue that arbitrary code can be executed via the `_fragment` controller.

alexisLefebvre:
Fair enough, does it require a CVE issue? :grimacing:

nicolas-grekas:
There is no security issue when one sets the secret as documented.

fabpot:
Let's merge the fix in 5.4 and 6.2 as well.

nicolas-grekas:
See #50238

PR #50017, opened 2023-04-14
[Validator] Fix support of Enum to `ConstraintValidator::formatValue`

### LINKED ISSUE #50000: Make ConstraintValidator->formatValue aware of enums
### Description

The `Symfony\Component\Validator\ConstraintValidator->formatValue()` method transforms values of a form into strings so they can be used in the error message. Now, if it is an object it checks if it's `Stringable` to convert it correctly, and if it isn't, it will use the value `"Object"`.

When the value is an enum, it will consider it as an object and check if it implements Stringable, the problem is that enums can't implement the magic method `__toString()` so it will be always "Object" which isn't too useful.

I propose to make it enum aware by adding a check to see if it's an enum and in that case use the name as value (or a concatenation of Enum-{name} to represent it.

### Example

For example, with the next code for validation of a form model:

```lang=php
// EnumNiceClass.php
enum EnumNiceClass: int {
    case DoSomething = 2;
    case NothingToDo = 3;
}

// FormModel.php
class FormModel {

Assert\NotIdenticalTo(
        value: EnumNiceClass::DoSomething,
)]
public EnumNiceClass $whatWillDo;

}
```

You will get the next message if it is invalid (with the default message `This value should not be identical to {{ compared_value_type }} {{ compared_value }}.`:

```
This value should not be identical to EnumNiceClass Object.
```

But changing the code to something like this:

```lang=php
    protected function formatValue(mixed $value, int $format = 0): string
    {

// ...

        if (\is_object($value)) {
            if (($format & self::OBJECT_TO_STRING) && $value instanceof \Stringable) {
                return $value->__toString();
            }
            
            if ($value instanceof \UnitEnum) {
                return $value->name;
            }

            return 'object';
        }
// ...
```

It would be:

```
This value should not be identical to EnumNiceClass NothingToDo.
```

That I think is more useful for debugging (and users in general).

PhoneixS:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50000
| License       | MIT

I have added a simple check in the `formatValue` to see if the object is an enum and in that case, return the name of the enum instead of the "object" string.

### DIFF
--- a/src/Symfony/Component/Validator/ConstraintValidator.php
+++ b/src/Symfony/Component/Validator/ConstraintValidator.php
@@ -99,6 +99,10 @@ abstract class ConstraintValidator implements ConstraintValidatorInterface
             return $value->format('Y-m-d H:i:s');
         }
 
+        if ($value instanceof \UnitEnum) {
+            return $value->name;
+        }
+
         if (\is_object($value)) {
             if (($format & self::OBJECT_TO_STRING) && method_exists($value, '__toString')) {
                 return $value->__toString();
--- a/src/Symfony/Component/Validator/Tests/ConstraintValidatorTest.php
+++ b/src/Symfony/Component/Validator/Tests/ConstraintValidatorTest.php
@@ -41,6 +41,7 @@ class ConstraintValidatorTest extends TestCase
             ['array', []],
             ['object', $toString = new TestToStringObject()],
             ['ccc', $toString, ConstraintValidator::OBJECT_TO_STRING],
+            ['FirstCase', $toString = TestEnum::FirstCase],
             ['object', $dateTime = new \DateTimeImmutable('1971-02-02T08:00:00UTC')],
             [class_exists(\IntlDateFormatter::class) ? 'Oct 4, 2019, 11:02 AM' : '2019-10-04 11:02:03', new \DateTimeImmutable('2019-10-04T11:02:03+09:00'), ConstraintValidator::PRETTY_DATE],
             [class_exists(\IntlDateFormatter::class) ? 'Feb 2, 1971, 8:00 AM' : '1971-02-02 08:00:00', $dateTime, ConstraintValidator::PRETTY_DATE],
@@ -72,4 +73,10 @@ final class TestToStringObject
     {
         return 'ccc';
     }
+}
+
+enum TestEnum
+{
+    case FirstCase;
+    case SecondCase;
 }


PhoneixS:
> Hey!
> 
> Thanks for your PR. You are targeting branch "6.3" but it seems your PR description refers to branch "6.2". Could you update the PR description or change target branch? This helps core maintainers a lot.
> 
> Cheers!
> 
> Carsonbot

Sorry, first time making a PR. I think it's changed now.

stof:
I'm wondering whether this should be done only when the `OBJECT_AS_STRING` flag is passed

src/Symfony/Component/Validator/ConstraintValidator.php
```diff
@@ -93,6 +93,10 @@ protected function formatValue(mixed $value, int $format = 0): string
                 return $value->__toString();
             }
 
+            if ($value instanceof \UnitEnum) {
+                return $value->name;
```
nicolas-grekas: I'd suggest going with `$value::class.'::'.$value->name`.
can you please add a test case about it?

PhoneixS:
I have changed so enums are treated more like objects, if there isn't an OBJECT_TO_STRING flag they are displayed as "enum"; if the flag is present it returns the enum name.

I have also added both test cases (with or without the flag).

src/Symfony/Component/Validator/ConstraintValidator.php
PhoneixS: I don't see why it is better to check it with your code `$value::class.'::'.$value->name`. I think it is more readable with `instanceof`. Is there a case in which it can fail?

PhoneixS:
Tests passed but failed because of unrelated  `Error: KO src/Symfony/Component/HttpClient`.

src/Symfony/Component/Validator/ConstraintValidator.php
nicolas-grekas: I'm not commenting on the check but on the return value.
We should return the name of the class + the name of case, not only the former.
`$value::class.'::'.$value->name`

src/Symfony/Component/Validator/ConstraintValidator.php
```diff
@@ -88,6 +88,14 @@ protected function formatValue(mixed $value, int $format = 0): string
             return $value->format('Y-m-d H:i:s');
         }
 
+        if (\is_object($value) && $value instanceof \UnitEnum) {
```
nicolas-grekas: is_object check is not needed, instanceof in enough

src/Symfony/Component/Validator/ConstraintValidator.php
PhoneixS: ohhh, I see it now. I will change it.

src/Symfony/Component/Validator/ConstraintValidator.php
PhoneixS: Correct, I'll remove it.

src/Symfony/Component/Validator/ConstraintValidator.php
PhoneixS: Mmm, with that change it's a bit too verbose as it returns the full qualified class name. Although as this is more of a developer thing than something which would be seen by clients it could be ok.

src/Symfony/Component/Validator/ConstraintValidator.php
```diff
@@ -88,6 +88,14 @@ protected function formatValue(mixed $value, int $format = 0): string
             return $value->format('Y-m-d H:i:s');
         }
 
+        if ($value instanceof \UnitEnum) {
+            if ($format & self::OBJECT_TO_STRING) {
+                return $value::class.'::'.$value->name;
```
nicolas-grekas: On a closer look, my advice was wrong I think, since the message already contains the class of the enum.
Let's make this be `return 'case '.$value->name;` so that messages looks like this?
"This value should not be identical to Symfony\Component\Validator\Tests\TestEnum case FirstCase"

PhoneixS: In my opinion, we shouldn't add the 'case ' text, as it would be better added in the message itself.

nicolas-grekas: but how can the message know about this? or maybe we should change the type to make it report "enum" and keep this as is?

nicolas-grekas: Or maybe formatTypeOf should be changed instead to return `$value::class.'::'.$value->name`?

PhoneixS: > Or maybe formatTypeOf should be changed instead to return `$value::class.'::'.$value->name`?

I see your point, however, I think that the function should only return the type (the class is the type and is enough for debugging) the same way the formatValue should only return the value (and the name is the textual representation of the enum value). Making each method return one thing whatever is called for. This is why I initially only return the name and probably we could just return to my initial implementation of returning the name of the enum.

nicolas-grekas: OK, let's do this.

src/Symfony/Component/Validator/ConstraintValidator.php
```diff
@@ -88,6 +88,14 @@ protected function formatValue(mixed $value, int $format = 0): string
             return $value->format('Y-m-d H:i:s');
         }
 
+        if ($value instanceof \UnitEnum) {
+            if ($format & self::OBJECT_TO_STRING) {
```
nicolas-grekas: I would remove this check personally, I don't see its value.

PhoneixS: The first it's to check if it's an enum instead of a generic object.  The second is to maintain the same behavior as with objects, if the flag is set you return a value if not set, you return the string of what it is, an "enum" in this case and "object" in the object case.

I think we should keep them to maintain a similar functionality as with objects.

nicolas-grekas: I don't see the link with OBJECT_TO_STRING flag here. This is unrelated to calling `__toString`. This method is creating a representation of the value. "enum" is not accurate enough, independently from any flag.

PhoneixS: It's true that it could be better to have another flag like `self::NAME_OF_ENUM` specific for the enums (the same way objects have `self::OBJECT_TO_STRING` and `self::PRETTY_DATE` for dates. As this is my first PR I wanted to change the less possible but I can add it if it fits.

nicolas-grekas: No need for any flag IMHO.

PhoneixS: So you say to always return the name? In my opinion is better to let the developer decide what they want. Maybe someone wants to hide the actual value because it's shown to the user but others prefer to see it for debugging proposed. Isn't it the same reasoning as for dates and objects?

nicolas-grekas: We display strings and numbers, so why not enums? I see each flag as a complication. The class of the enum is already exposed, not saying which case is concerned just worsen DX to me, this is not something that devs want to opt-in for, nor disable.

PhoneixS: Ok, I will remove the flag and return always the name

nicolas-grekas: Thanks, and sorry for the confusion :)

nicolas-grekas:
Thank you @PhoneixS.

PhoneixS:
You’re welcome. It's always nice to return a little bit to a great project.

PR #50022, opened 2023-04-14
[DependencyInjection] Fallback to default value when autowiring undefined parameters for optional arguments

### LINKED ISSUE #50020: Dependency Injection Autowiring throws exception for nullable parameters that don't exist

all?

### Description

I get a ParameterNotFoundException `"You have requested a non-existent parameter "myclass.optional_parameter"` with the below code.

`ExistentClass $requiredToExist` appropriately throws an exception when ExistentClass is not found in the container.
`?NonExistentClass $optionalNotExist` appropriately is set to null when the NonExistentClass is not found in the container.

I have not tested on any version other than 6.2, so I don't know if this bug exists in other versions.

### How to reproduce

```
namespace App\Service;

class MyServiceClass
{
    public function __construct(
        private ExistentClass $requiredToExist,
        private ?NonExistentClass $optionalNotExist = null,
        #[Autowire('%myclass.optional_parameter%')]
        private ?string $optionalParameter = null,
    ) {}

    public function doStuff()
    {
        if ('value1' == $this->optionalParameter) {
            // do something
        } elseif ('value2' == $this->optionalParameter) {
            // do something else
        }
    }
}
```

Register App\Service\MyServiceClass as a service, and request it as a parameter to a controller method.

### Possible Solution

Naive update to the `$attribute instanceof Autowire` case of the [processAttribute method in Symfony/Component/DependencyInjection/Compiler/AutowirePass.php](https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php#L165):

```
try {
    $value = $this->container->getParameterBag()->resolveValue($attribute->value);
    return $value instanceof Reference && $isOptional ? new Reference($value, ContainerInterface::NULL_ON_INVALID_REFERENCE) : $value;
} catch (ParameterNotFoundException $ex) {
    return $isOptional ? null : throw $ex;
}
```

### Additional Context

_No response_

radar3301:
| Q             | A
| ------------- | ---
| Branch?       | 6.2 / 6.3
| Bug fix?      | yes
| New feature?  | yes 
| Deprecations? | no 
| Tickets       | Fix #50020 
| License       | MIT
| Related       | PR #50023

Allows autowiring of undefined nullable parameters, similar to existing behavior for non-parameter values. See related issue for more detail.

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
+++ b/src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
@@ -23,6 +23,7 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Exception\AutowiringFailedException;
+use Symfony\Component\DependencyInjection\Exception\ParameterNotFoundException;
 use Symfony\Component\DependencyInjection\Exception\RuntimeException;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\DependencyInjection\TypedReference;
@@ -279,9 +280,16 @@ class AutowirePass extends AbstractRecursivePass
             if ($checkAttributes) {
                 foreach ($parameter->getAttributes() as $attribute) {
                     if (\in_array($attribute->getName(), [TaggedIterator::class, TaggedLocator::class, Autowire::class, MapDecorated::class], true)) {
-                        $arguments[$index] = $this->processAttribute($attribute->newInstance(), $parameter->allowsNull());
-
-                        continue 2;
+                        try {
+                            $arguments[$index] = $this->processAttribute($attribute->newInstance(), $parameter->allowsNull());
+                            continue 2;
+                        } catch (ParameterNotFoundException $e) {
+                            if (!$parameter->isDefaultValueAvailable()) {
+                                throw new AutowiringFailedException($this->currentId, $e->getMessage(), 0, $e);
+                            }
+                            $arguments[$index] = clone $this->defaultArgument;
+                            $arguments[$index]->value = $parameter->getDefaultValue();
+                        }
                     }
                 }
             }
--- a/src/Symfony/Component/DependencyInjection/Tests/Compiler/AutowirePassTest.php
+++ b/src/Symfony/Component/DependencyInjection/Tests/Compiler/AutowirePassTest.php
@@ -1217,6 +1217,38 @@ class AutowirePassTest extends TestCase
         $this->assertNull($service->invalid);
     }
 
+    public function testAutowireAttributeNullFallbackTestRequired()
+    {
+        $container = new ContainerBuilder();
+
+        $container->register('foo', AutowireAttributeNullFallback::class)
+            ->setAutowired(true)
+            ->setPublic(true)
+        ;
+
+        $this->expectException(AutowiringFailedException::class);
+        $this->expectExceptionMessage('You have requested a non-existent parameter "required.parameter".');
+        (new AutowirePass())->process($container);
+    }
+
+    public function testAutowireAttributeNullFallbackTestOptional()
+    {
+        $container = new ContainerBuilder();
+
+        $container->register('foo', AutowireAttributeNullFallback::class)
+            ->setAutowired(true)
+            ->setPublic(true)
+        ;
+
+        $container->setParameter('required.parameter', 'foo');
+
+        (new AutowirePass())->process($container);
+
+        $definition = $container->getDefinition('foo');
+
+        $this->assertSame(['foo'], $definition->getArguments());
+    }
+
     public function testAsDecoratorAttribute()
     {
         $container = new ContainerBuilder();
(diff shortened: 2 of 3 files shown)


radar3301:
Grr... Tried renaming local patch to https://github.com/radar3301/symfony/tree/autowire-undefined-nullable-6.2

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
```diff
@@ -166,9 +167,12 @@ private function processAttribute(object $attribute, bool $isOptional = false):
     {
         switch (true) {
             case $attribute instanceof Autowire:
-                $value = $this->container->getParameterBag()->resolveValue($attribute->value);
-
-                return $value instanceof Reference && $isOptional ? new Reference($value, ContainerInterface::NULL_ON_INVALID_REFERENCE) : $value;
+                try {
+                    $value = $this->container->getParameterBag()->resolveValue($attribute->value);
+                    return $value instanceof Reference && $isOptional ? new Reference($value, ContainerInterface::NULL_ON_INVALID_REFERENCE) : $value;
```
nicolas-grekas: should be outside the try/catch, isn't it?

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
```diff
@@ -166,9 +167,12 @@ private function processAttribute(object $attribute, bool $isOptional = false):
...
+                } catch (ParameterNotFoundException $ex) {
```
nicolas-grekas: $ex should be renamed to $e

nicolas-grekas (commented):
Can you add a test case please?

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
radar3301: Sorry, which statement should be outside the try/catch?

radar3301: Resolved, after thinking about it more.

radar3301:
> Can you add a test case please?

I don't understand any of the test suite stuff...

I'm not even sure that the existing `AutowirePassTest::testAutowireAttribute()` even passes.
I don't see an `AutowireAttribute` class defined in `Symfony/Component/DependencyInjection/Tests/Compiler`, nor does a `use` alias exist for it.
I'm not even sure how `$this->assertCount(9, $definition->getArguments());` on a non-existent class passes...

radar3301:
> non-existent AutowireAttribute class

Nevermind, I found it...

Test cases submitted.

src/Symfony/Component/DependencyInjection/Tests/Compiler/AutowirePassTest.php
```diff
@@ -1216,6 +1216,57 @@ public function testAutowireAttribute()
...
+            ->setPublic(true)
+        ;
+
+        (new AutowirePass())->process($container);
+
+        $definition = $container->getDefinition(AutowireAttributeNullFallback::class);
+
+        $this->assertCount(2, $definition->getArguments());
+        $this->assertEquals(new Reference('required.parameter', ContainerInterface::EXCEPTION_ON_INVALID_REFERENCE), $definition->getArgument(0));
+        $this->assertEquals(new Reference('optional.parameter', ContainerInterface::NULL_ON_INVALID_REFERENCE), $definition->getArgument(1));
```
radar3301: Might need to be `$this->assertEquals(null, $definition->getArgument(1));` ?

src/Symfony/Component/DependencyInjection/Tests/Compiler/AutowirePassTest.php
```diff
@@ -1216,6 +1216,57 @@ public function testAutowireAttribute()
...
+        
+        $container->setParameter('required.parameter', 'foo');
+        
+        (new AutowirePass())->process($container);
+
+        $definition = $container->getDefinition(AutowireAttributeNullFallback::class);
+
+        $this->assertCount(2, $definition->getArguments());
+        $this->assertEquals(new Reference('required.parameter', ContainerInterface::EXCEPTION_ON_INVALID_REFERENCE), $definition->getArgument(0));
+        $this->assertEquals(new Reference('optional.parameter', ContainerInterface::NULL_ON_INVALID_REFERENCE), $definition->getArgument(1));
```
radar3301: Same as above, might need to be `$this->assertEquals(null, $definition->getArgument(1));` ?

src/Symfony/Component/DependencyInjection/Tests/Compiler/AutowirePassTest.php
nicolas-grekas: Likely so, but for now there is this:
Cannot use null as default value for parameter $required of type string in /home/runner/work/symfony/symfony/src/Symfony/Component/DependencyInjection/Tests/Fixtures/includes/autowiring_classes_80.php on line 61

nicolas-grekas: should be assertNull BTW !

radar3301: Fixed.

nicolas-grekas (approved):
LGTM (I made some tweaks to the patch)

nicolas-grekas:
Thank you @radar3301.

PR #50023, opened 2023-04-14
[DependencyInjection] [Autowiring] Allow undefined nullable parameters (6.3)

### LINKED ISSUE #50020: Dependency Injection Autowiring throws exception for nullable parameters that don't exist

radar3301:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | yes 
| Deprecations? | no 
| Tickets       | Fix #50020 
| License       | MIT
| Related       | PR #50022

Allows autowiring of undefined nullable parameters, similar to existing behavior for non-parameter values. See related issue for more detail.

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
+++ b/src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
@@ -21,6 +21,7 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Exception\AutowiringFailedException;
+use Symfony\Component\DependencyInjection\Exception\ParameterNotFoundException;
 use Symfony\Component\DependencyInjection\Exception\RuntimeException;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\DependencyInjection\TypedReference;
--- a/src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
+++ b/src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
@@ -287,8 +287,14 @@ class AutowirePass extends AbstractRecursivePass
             if ($checkAttributes) {
                 foreach ($parameter->getAttributes(Autowire::class, \ReflectionAttribute::IS_INSTANCEOF) as $attribute) {
                     $attribute = $attribute->newInstance();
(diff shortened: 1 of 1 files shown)


radar3301:
Grr... Tried renaming local patch to https://github.com/radar3301/symfony/tree/autowire-undefined-nullable-6.3

nicolas-grekas:
Thanks for the PRs. Usually we ask for the patch on the lowest affected branch, and we merge up periodically. When the merge generates conflicts, we fix the conflict and we sometimes ask the author for their help if it's non trivial. I'm therefor closing this PR but it'll definitely help when merging #50022, if accepted!

radar3301:
Thanks for the feedback Nicolas.

---

The Autowire Attribute was added in 6.1, and handling of the Autowire attribute was added directly to `autowireMethod`.
6.2 added a `processAttribute` method to do the actual processing, and rewrote `autowireMethod` to handle the attribute.
Then 6.3 removed `processAttribute` and moved the attribute handling back into `autowireMethod`.

It's all a big mess, hence the 3 different patches I created.

Hope this brief overview helps when then time comes!

PR #50024, opened 2023-04-15
[Serializer] Fix denormalization of object with typed constructor arg (not castable) and with COLLECT_DENORMALIZATION_ERRORS

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

There is already a test for constructor instantiation but it worked "by chance".
The test is `testCollectDenormalizationErrorsWithConstructor`.
It worked because `$reflectionClass->newInstanceArgs($params)` works if
one give a `string` value that should fit in a `bool` type. PHP casts
automatically the value for us. (see: https://3v4l.org/GYr0T)

Now we ensure to not throw an exception if the instantiation cannot be
done. The mismatch is already collected few lines above, so there
nothing to do more

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
@@ -417,7 +417,15 @@ abstract class AbstractNormalizer implements NormalizerInterface, DenormalizerIn
             }
 
             if ($constructor->isConstructor()) {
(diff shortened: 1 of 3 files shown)


chalasr (approved):

nicolas-grekas:
Thank you @lyrixx.

PR #48272, opened 2022-11-21
[Security] Add OidcUserInfoTokenHandler and OidcUser

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/17463

Hi,

This PR aims to complete [the previous one](https://github.com/symfony/symfony/pull/46428) from @Spomky with an AccessTokenHandler ready-to-use with an OIDC server (Keycloak, Auth0).

## TODO

- [x] Rebase from 6.3
- [x] Rebase from #48285
- [x] Rebase from #48594
- [x] Write doc (https://github.com/symfony/symfony-docs/pull/17463)
- [x] Add TokenHandlerFactory
- [x] Add ServiceTokenHandlerFactory for BC layer
- [x] Add OidcUserInfoTokenHandlerFactory
- [x] Add OidcTokenHandlerFactory (using web-token/jwt-*)
- [x] Implement OidcUser to keep user claims from OIDC server
- [x] Update doc PR about claims usage in a custom UserProvider
- [x] ~Update doc PR about OidcUserProvider usage~ (abandonned)

## Usage

```yaml
# usage with a custom client
security:
    firewalls:
        main:
            pattern: ^/
            access_token:
                token_handler:
                    oidc_user_info:
                        client: oidc.client
```
```yaml
# usage with generic HttpClient
security:
    firewalls:
        main:
            pattern: ^/
            access_token:
                token_handler:
                    oidc_user_info:
                        claim: email
                        client:
                            base_uri: https://www.example.com/realms/demo/protocol/openid-connect/userinfo
```
```yaml
# usage with token decode (no call to OIDC server)
security:
    firewalls:
        main:
            pattern: ^/
            access_token:
                token_handler:
                    oidc:
                        signature:
                            # Algorithm used to sign the JWS
                            algorithm: 'HS256'
                            # A JSON-encoded JWK
                            key: '{"kty":"...","k":"..."}'
```
```php
# usage with a custom UserProvider
class CustomUserProvider implements UserProviderInterface
{
    public function loadUserByIdentifier(string $identifier, array $claims = []): UserInterface
    {
        // do some magic
    }
}
```

Issue #30914, opened 2019-04-06
[RFC] Symfony Security rework tracking issue


curry684:
After discussions at EU FOSSA Hackathon we have some ideas on how to rework the Security component. I'm writing this "tracker issue" to gather up the info and choices made thus far.

The biggest functional issues have been discussed frequently over the years (https://github.com/symfony/symfony/issues/10316 and others) that the authentication makes some shortcut assumptions no longer valid in 2019:
- The object of security considerations is not always a user, it may be a device, a computer, an API key, or any number of other abstract things. In the security world the concept of "something that can participate in security questions" is called a Principal.
- There should be a functional separation between the Principal performing an action and the Context he is in while doing so. Authorization decisions should be based on the presented Context. This solves many practical issues like impersonation and "act as company".
- Principals may be authenticated by various means, and be oblivious themselves on the subject. Principals should therefore not expose functionality to get passwords or roles as they may not have any. Tokens handle authentication.

The Symfony Security component was originally based on the Java Spring Security framework, but has deviated over the years. Today Spring Security has both fixed some structural issues that Symfony has (the notion of principals versus users and no central assumption that passwords and roles exist) and has some features that are currently lacking or hacky at best (structural support for OpenID/OAuth, support for 2FA). Conclusion for now is that most of the current Security component, especially authorization, is mostly fine and does not need a lot of work, and it wouldn't be all that hard to fix the authentication parts, and more strongly decouple authentication and authorization, by taking some current inspiration from Spring again.

Once these changes are propagated a lot of other features related to authentication should be easier to implement. Wishlists have been assembled:

Authentication
=
- [x] User impersonation (including multiple impersonation without having to log out) *not affected by Security changes*
- [x] <s>Social login (login in the app using Google, Facebook, GitHub, Twitter, etc.) and generic OAuth support</s> (https://github.com/symfony/symfony/pull/31952#issuecomment-698424823)
- [x] Login throttling (limit the number of failed login attempts over a period of time)
- [ ] Simultaneous session limiting (e.g. each user can login only from one device at the same time)
- [x] <s>Two-factor (or multi-factor) authentication</s> (https://github.com/symfony/symfony/issues/28868#issuecomment-774069638)
- [ ] “Sudo mode” (trigger a new authentication before sensitive operations, like GitHub) (#33955)
- [ ] Native JWT implementation if some package is available (firebase/php-jwt ?)
- [ ] CSRF on router level 
[...]

PR #50027, opened 2023-04-15
[Security] OAuth2 Introspection Endpoint (RFC7662)

Spomky:
| Q             | A
| ------------- | ---
| Branch?       | 7.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | none
| License       | MIT
| Doc PR        | symfony/symfony-docs#[**TODO**]

In addition to the excellent work of @vincentchalamon #48272, this PR allows getting the data from the OAuth2 Introspection Endpoint. This endpoint is defined in the [RFC7662](https://datatracker.ietf.org/doc/html/rfc7662). It returns the following information that is used to retrieve the user:

* If the access token is active
* A set of claims that are similar to the OIDC one, including the `sub` or the `username`.

Example of configuration:

```yaml
framework:
    http_client:
        scoped_clients:
            oauth2.client:
                base_uri: 'https://authorization-server.example.com/introspection'
                scope: 'https://authorization-server\.example\.com'
                headers:
                    Authorization: 'Basic Y2xpZW50OnBhc3N3b3Jk' # Introspection Endpoint usually requires client authentication

security:
    firewalls:
        main:
            pattern: ^/
            access_token:
                token_handler:
                    oauth2: ~
                token_extractors: 'header' 
                realm: 'My API'
```

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/CHANGELOG.md
+++ b/src/Symfony/Bundle/SecurityBundle/CHANGELOG.md
@@ -9,6 +9,7 @@ CHANGELOG
  * Add `expose_security_errors` config option to display `AccountStatusException`
  * Deprecate the `security.hide_user_not_found` config option in favor of `security.expose_security_errors`
  * Add ability to fetch LDAP roles
+ * Add `OAuth2TokenHandlerFactory` for `AccessTokenFactory`
 
 7.2
 ---
--- a/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OAuth2TokenHandlerFactory.php
+++ b/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OAuth2TokenHandlerFactory.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\SecurityBundle\DependencyInjection\Security\AccessToken;
+
+use Symfony\Component\Config\Definition\Builder\NodeBuilder;
+use Symfony\Component\DependencyInjection\ChildDefinition;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * Configures a token handler for an OAuth2 Token Introspection endpoint.
+ *
+ * @internal
+ */
+class OAuth2TokenHandlerFactory implements TokenHandlerFactoryInterface
+{
+    public function create(ContainerBuilder $container, string $id, array|string $config): void
+    {
+        $container->setDefinition($id, new ChildDefinition('security.access_token_handler.oauth2'));
+    }
+
+    public function getKey(): string
+    {
+        return 'oauth2';
+    }
+
+    public function addConfiguration(NodeBuilder $node): void
+    {
+        $node->scalarNode($this->getKey())->end();
+    }
+}
--- a/src/Symfony/Bundle/SecurityBundle/Resources/config/security_authenticator_access_token.php
+++ b/src/Symfony/Bundle/SecurityBundle/Resources/config/security_authenticator_access_token.php
@@ -36,6 +36,7 @@ use Jose\Component\Signature\Algorithm\RS512;
 use Symfony\Component\Security\Http\AccessToken\ChainAccessTokenExtractor;
 use Symfony\Component\Security\Http\AccessToken\FormEncodedBodyExtractor;
 use Symfony\Component\Security\Http\AccessToken\HeaderAccessTokenExtractor;
+use Symfony\Component\Security\Http\AccessToken\OAuth2\Oauth2TokenHandler;
 use Symfony\Component\Security\Http\AccessToken\Oidc\OidcTokenHandler;
 use Symfony\Component\Security\Http\AccessToken\Oidc\OidcUserInfoTokenHandler;
 use Symfony\Component\Security\Http\AccessToken\QueryAccessTokenExtractor;
@@ -186,5 +187,13 @@ return static function (ContainerConfigurator $container) {
 
         ->set('security.access_token_handler.oidc.encryption.A256GCM', A256GCM::class)
             ->tag('security.access_token_handler.oidc.encryption_algorithm')
+
+        // OAuth2 Introspection (RFC 7662)
+        ->set('security.access_token_handler.oauth2', Oauth2TokenHandler::class)
+            ->abstract()
+            ->args([
+                service('http_client'),
+                service('logger')->nullOnInvalid(),
+            ])
     ;
 };
--- a/src/Symfony/Component/Security/Core/User/OAuth2User.php
+++ b/src/Symfony/Component/Security/Core/User/OAuth2User.php
@@ -1,0 +1,70 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Security\Core\User;
+
+/**
+ * UserInterface implementation used by the access-token security workflow with an OIDC server.
+ */
+class OAuth2User implements UserInterface
+{
+    public readonly array $additionalClaims;
+
+    public function __construct(
+        private array $roles = ['ROLE_USER'],
+        // Standard Claims (https://datatracker.ietf.org/doc/html/rfc7662#section-2.2)
+        public readonly ?string $scope = null,
+        public readonly ?string $clientId = null,
+        public readonly ?string $username = null,
+        public readonly ?string $tokenType = null,
+        public readonly ?int $exp = null,
+        public readonly ?int $iat = null,
+        public readonly ?int $nbf = null,
+        public readonly ?string $sub = null,
+        public readonly ?string $aud = null,
+        public readonly ?string $iss = null,
+        public readonly ?string $jti = null,
+
+        // Additional Claims ("
+        //    Specific implementations MAY extend this structure with
+        //    their own service-specific response names as top-level members
+        //    of this JSON object.
+        // ")
+        ...$additionalClaims,
+    ) {
+        if ((null === $sub || '' === $sub) && (null === $username || '' === $username)) {
+            throw new \InvalidArgumentException('The claim "sub" or "username" must be provided.');
+        }
+
+        $this->additionalClaims = $additionalClaims['additionalClaims'] ?? $additionalClaims;
+    }
+
+    /**
+     * OIDC or OAuth specs don't have any "role" notion.
+     *
+     * If you want to implement "roles" from your OIDC server,
+     * send a "roles" constructor argument to this object
+     * (e.g.: using a custom UserProvider).
+     */
+    public function getRoles(): array
+    {
+        return $this->roles;
+    }
+
+    public function getUserIdentifier(): string
+    {
+        return (string) ($this->sub ?? $this->username);
+    }
+
+    public function eraseCredentials(): void
+    {
+    }
+}
--- a/src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
+++ b/src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
@@ -1,0 +1,100 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Security\Http\AccessToken\OAuth2;
+
+use Psr\Log\LoggerInterface;
+use Symfony\Component\Security\Core\Exception\AuthenticationException;
+use Symfony\Component\Security\Core\Exception\BadCredentialsException;
+use Symfony\Component\Security\Core\User\OAuth2User;
+use Symfony\Component\Security\Http\AccessToken\AccessTokenHandlerInterface;
+use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+use function Symfony\Component\String\u;
+
+/**
+ * The token handler validates the token on the authorization server and the Introspection Endpoint.
+ *
+ * @see https://tools.ietf.org/html/rfc7662
+ *
+ * @internal
+ */
+final class Oauth2TokenHandler implements AccessTokenHandlerInterface
+{
+    public function __construct(
+        private readonly HttpClientInterface $client,
+        private readonly ?LoggerInterface $logger = null,
+    ) {
+    }
+
+    public function getUserBadgeFrom(string $accessToken): UserBadge
+    {
+        try {
+            // Call the Authorization server to retrieve the resource owner details
+            // If the token is invalid or expired, the Authorization server will return an error
+            $claims = $this->client->request('POST', '', [
+                'body' => [
+                    'token' => $accessToken,
+                    'token_type_hint' => 'access_token',
+                ],
+            ])->toArray();
+
+            $sub = $claims['sub'] ?? null;
+            $username = $claims['username'] ?? null;
+            if (!$sub && !$username) {
+                throw new BadCredentialsException('"sub" and "username" claims not found on the authorization server response. At least one is required.');
+            }
+            $active = $claims['active'] ?? false;
+            if (!$active) {
+                throw new BadCredentialsException('The claim "active" was not found on the authorization server response or is set to false.');
+            }
+
+            return new UserBadge($sub ?? $username, fn () => $this->createUser($claims), $claims);
+        } catch (AuthenticationException $e) {
+            $this->logger?->error('An error occurred on the authorization server.', [
+                'error' => $e->getMessage(),
+                'trace' => $e->getTraceAsString(),
+            ]);
+
+            throw new BadCredentialsException('Invalid credentials.', $e->getCode(), $e);
+        }
+    }
+
+    private function createUser(array $claims): OAuth2User
+    {
+        if (!\function_exists(\Symfony\Component\String\u::class)) {
+            throw new \LogicException('You cannot use the "OAuth2TokenHandler" since the String component is not installed. Try running "composer require symfony/string".');
+        }
+
+        foreach ($claims as $claim => $value) {
+            unset($claims[$claim]);
+            if ('' === $value || null === $value) {
+                continue;
+            }
+            $claims[u($claim)->camel()->toString()] = $value;
+        }
+
+        if ('' !== ($claims['updatedAt'] ?? '')) {
+            $claims['updatedAt'] = (new \DateTimeImmutable())->setTimestamp($claims['updatedAt']);
+        }
+
+        if ('' !== ($claims['emailVerified'] ?? '')) {
+            $claims['emailVerified'] = (bool) $claims['emailVerified'];
+        }
+
+        if ('' !== ($claims['phoneNumberVerified'] ?? '')) {
+            $claims['phoneNumberVerified'] = (bool) $claims['phoneNumberVerified'];
+        }
+
+        return new OAuth2User(...$claims);
+    }
+}
--- a/src/Symfony/Component/Security/Http/Tests/AccessToken/OAuth2/OAuth2TokenHandlerTest.php
+++ b/src/Symfony/Component/Security/Http/Tests/AccessToken/OAuth2/OAuth2TokenHandlerTest.php
@@ -1,0 +1,52 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Security\Http\Tests\AccessToken\OAuth2;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\HttpClient\MockHttpClient;
+use Symfony\Component\HttpClient\Response\MockResponse;
+use Symfony\Component\Security\Core\User\OAuth2User;
+use Symfony\Component\Security\Http\AccessToken\OAuth2\Oauth2TokenHandler;
+use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
+
+class OAuth2TokenHandlerTest extends TestCase
+{
+    public static function testGetsUserIdentifierFromOAuth2ServerResponse(): void
+    {
+        $accessToken = 'a-secret-token';
+        $claims = [
+            'active' => true,
+            'client_id' => 'l238j323ds-23ij4',
+            'username' => 'jdoe',
+            'scope' => 'read write dolphin',
+            'sub' => 'Z5O3upPC88QrAjx00dis',
+            'aud' => 'https://protected.example.net/resource',
+            'iss' => 'https://server.example.com/',
+            'exp' => 1419356238,
+            'iat' => 1419350238,
+            'extension_field' => 'twenty-seven',
+        ];
+        $expectedUser = new OAuth2User(...$claims);
+
+        $client = new MockHttpClient([
+            new MockResponse(json_encode($claims, \JSON_THROW_ON_ERROR)),
+        ]);
+
+        $userBadge = (new Oauth2TokenHandler($client))->getUserBadgeFrom($accessToken);
+        $actualUser = $userBadge->getUserLoader()();
+
+        self::assertEquals(new UserBadge('Z5O3upPC88QrAjx00dis', fn () => $expectedUser, $claims), $userBadge);
+        self::assertInstanceOf(OAuth2User::class, $actualUser);
+        self::assertSame($claims, $userBadge->getAttributes());
+        self::assertSame($claims['sub'], $actualUser->getUserIdentifier());
+    }
+}
--- a/src/Symfony/Bundle/SecurityBundle/SecurityBundle.php
+++ b/src/Symfony/Bundle/SecurityBundle/SecurityBundle.php
@@ -24,6 +24,7 @@ use Symfony\Bundle\SecurityBundle\DependencyInjection\Compiler\RegisterTokenUsag
 use Symfony\Bundle\SecurityBundle\DependencyInjection\Compiler\ReplaceDecoratedRememberMeHandlerPass;
 use Symfony\Bundle\SecurityBundle\DependencyInjection\Compiler\SortFirewallListenersPass;
 use Symfony\Bundle\SecurityBundle\DependencyInjection\Security\AccessToken\CasTokenHandlerFactory;
+use Symfony\Bundle\SecurityBundle\DependencyInjection\Security\AccessToken\OAuth2TokenHandlerFactory;
 use Symfony\Bundle\SecurityBundle\DependencyInjection\Security\AccessToken\OidcTokenHandlerFactory;
 use Symfony\Bundle\SecurityBundle\DependencyInjection\Security\AccessToken\OidcUserInfoTokenHandlerFactory;
 use Symfony\Bundle\SecurityBundle\DependencyInjection\Security\AccessToken\ServiceTokenHandlerFactory;
@@ -80,6 +81,7 @@ class SecurityBundle extends Bundle
             new OidcUserInfoTokenHandlerFactory(),
             new OidcTokenHandlerFactory(),
             new CasTokenHandlerFactory(),
+            new OAuth2TokenHandlerFactory(),
         ]));
 
         $extension->addUserProviderFactory(new InMemoryFactory());
--- a/src/Symfony/Bundle/SecurityBundle/Tests/DependencyInjection/Security/Factory/AccessTokenFactoryTest.php
+++ b/src/Symfony/Bundle/SecurityBundle/Tests/DependencyInjection/Security/Factory/AccessTokenFactoryTest.php
@@ -13,6 +13,7 @@ namespace Symfony\Bundle\SecurityBundle\Tests\DependencyInjection\Security\Facto
 
 use PHPUnit\Framework\TestCase;
 use Symfony\Bundle\SecurityBundle\DependencyInjection\Security\AccessToken\CasTokenHandlerFactory;
+use Symfony\Bundle\SecurityBundle\DependencyInjection\Security\AccessToken\OAuth2TokenHandlerFactory;
 use Symfony\Bundle\SecurityBundle\DependencyInjection\Security\AccessToken\OidcTokenHandlerFactory;
 use Symfony\Bundle\SecurityBundle\DependencyInjection\Security\AccessToken\OidcUserInfoTokenHandlerFactory;
 use Symfony\Bundle\SecurityBundle\DependencyInjection\Security\AccessToken\ServiceTokenHandlerFactory;
@@ -423,6 +424,22 @@ class AccessTokenFactoryTest extends TestCase
         $this->processConfig($config, $factory);
     }
 
+    public function testOAuth2TokenHandlerConfiguration()
+    {
+        $container = new ContainerBuilder();
+        $config = [
+            'token_handler' => ['oauth2' => true],
+        ];
+
+        $factory = new AccessTokenFactory($this->createTokenHandlerFactories());
+        $finalizedConfig = $this->processConfig($config, $factory);
+
+        $factory->createAuthenticator($container, 'firewall1', $finalizedConfig, 'userprovider');
+
+        $this->assertTrue($container->hasDefinition('security.authenticator.access_token.firewall1'));
+        $this->assertTrue($container->hasDefinition('security.access_token_handler.firewall1'));
+    }
+
     public function testNoTokenHandlerSet()
     {
         $this->expectException(InvalidConfigurationException::class);
@@ -482,6 +499,7 @@ class AccessTokenFactoryTest extends TestCase
             new OidcUserInfoTokenHandlerFactory(),
             new OidcTokenHandlerFactory(),
             new CasTokenHandlerFactory(),
+            new OAuth2TokenHandlerFactory(),
         ];
     }
 }
--- a/src/Symfony/Bundle/SecurityBundle/Tests/Functional/app/AccessToken/config_oauth2.yml
+++ b/src/Symfony/Bundle/SecurityBundle/Tests/Functional/app/AccessToken/config_oauth2.yml
@@ -1,0 +1,34 @@
+imports:
+    - { resource: ./../config/framework.yml }
+
+framework:
+    http_method_override: false
+    serializer: ~
+    http_client:
+        scoped_clients:
+            oauth2.client:
+                scope: 'https://authorization-server\.example\.com'
+                headers:
+                    Authorization: 'Basic Y2xpZW50OnBhc3N3b3Jk'
+
+security:
+    password_hashers:
+        Symfony\Component\Security\Core\User\InMemoryUser: plaintext
+
+    providers:
+        in_memory:
+            memory:
+                users:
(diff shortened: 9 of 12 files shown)


wouterj:
Thanks for yet another security feature proposal!
First a practical thing: We're entering feature freeze very shortly, meaning that we don't merge newly opened feature PRs for 6.3 anymore (unless really needed or very small). So we have some more time to polish this PR for 6.4 :)

I want to discuss a general remark that I don't think I've shared publicly yet. _(disclaimer: I know literally nothing about OAuth 2.0 and OIDC, except that it consists of a ton of different standards, so I can't determine if this applies to this PR)_

My vision for Symfony Security is to provide a minimal implementation following industry best practices based on standards and let community libraries/bundle add all "bonus features". We don't have the man power nor knowledge to maintain let's say all OAuth, SAML, Webauthn and MFA standards.

When adding new features, I think we must always test it to those 3 conditions:
* Is the feature based on a standard? (which is clearly yes for this one)
* Is this feature required to have a working authentication/authorization system, within reason (i.e. if 90% of the apps use it when using OIDC)? (this is what I'm a bit unsure about for this specific PR)
* Is this feature following industry best practices? (e.g. this is why I pushed a bit to add a JWT implementation in the OIDC PR, given that the user info handler is generally not recommended for production)

cc @chalasr

chalasr:
I agree with Wouter.
Implementing all OAuth2/OIDC features in core is a non-goal for now. Better improve community packages that are already supporting them or are meant to do so.

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public function getUserBadgeFrom(string $accessToken): UserBadge
+    {
+        try {
+            // Call the Introspection Endpoint to retrieve the user info
+            // If the token is invalid or expired, the Introspection Endpoint will return 'active' => false
+            $claims = $this->client->request('POST', '',[
+                'headers' => [
+                    'Content-Type' => 'application/x-www-form-urlencoded',
```
nicolas-grekas: not needed, that's already the default

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
```diff
@@ -0,0 +1,76 @@
...
+        private string $claim = 'sub'
+    ) {
+    }
+
+    public function getUserBadgeFrom(string $accessToken): UserBadge
+    {
+        try {
+            // Call the Introspection Endpoint to retrieve the user info
+            // If the token is invalid or expired, the Introspection Endpoint will return 'active' => false
+            $claims = $this->client->request('POST', '',[
```
nicolas-grekas: missing space before `[`

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
```diff
@@ -0,0 +1,76 @@
...
+                'headers' => [
+                    'Content-Type' => 'application/x-www-form-urlencoded',
+                ],
+                'body' => [
+                    'token' => $accessToken,
+                    'token_type_hint' => 'access_token',
+                ],
+            ])->toArray();
+
+            if (empty($claims['active'])) {
```
nicolas-grekas: if (!($claims['active'] ?? false)) {

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
```diff
@@ -0,0 +1,76 @@
...
+                throw new MissingClaimException(sprintf('"%s" claim not found on the authorization server response.', 'active'));
+            }
+            if ($claims['active'] !== true) {
```
nicolas-grekas: `if ($claims['active']) {`

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
```diff
@@ -0,0 +1,76 @@
...
+                throw new InactiveAccessTokenException('The access token is not active.');
+            }
+            if (empty($claims[$this->claim])) {
```
nicolas-grekas: We try to avoid using `empty()` and using comparisons to boolean, unless a type check is required.
`if (!($claims[$this->claim] ?? false)) {`

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
```diff
@@ -0,0 +1,76 @@
...
+                throw new MissingClaimException(sprintf('"%s" claim not found on the authorization server response.', 'active'));
+            }
+
+            return new UserBadge($claims[$this->claim], fn () => $this->createUser($claims), $claims);
+        } catch (\Throwable $e) {
```
nicolas-grekas: can we make this more specific somehow? catching any throwable is also something we try to avoid unless there are reasons

src/Symfony/Component/Security/Http/Tests/AccessToken/OAuth2/OAuth2TokenHandlerTest.php
```diff
@@ -0,0 +1,74 @@
...
+            ])
+            ->willReturn($responseMock);
+
+        $userBadge = (new OAuth2TokenHandler($clientMock))->getUserBadgeFrom($accessToken);
+        $actualUser = $userBadge->getUserLoader()();
+
+        $this->assertEquals(new UserBadge($expected, fn () => $expectedUser, $claims), $userBadge);
+        $this->assertInstanceOf(OAuth2User::class, $actualUser);
+        $this->assertEquals($claims, $userBadge->getAttributes());
+        $this->assertEquals($claims['sub'], $actualUser->getUserIdentifier());
```
nicolas-grekas: please use assertSame everywhere possible

src/Symfony/Component/Security/Http/Tests/AccessToken/OAuth2/OAuth2TokenHandlerTest.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public function getClaims(): iterable
```
nicolas-grekas: phpunit providers should be all static

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
```diff
@@ -0,0 +1,76 @@
...
+use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+/**
+ * The token handler validates the token on the authorization server.
+ * It uses the Introspection Endpoint as per the RFC7662
+ *
+ * @see https://tools.ietf.org/html/rfc7662
+ *
+ * @experimental
```
nicolas-grekas: we usually don't add experimental features to existing component: either a new component is introduced and it can be exprimental, or it's a new non-experimental feat of an existing component

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
```diff
@@ -0,0 +1,76 @@
...
+use Symfony\Component\Security\Core\Exception\BadCredentialsException;
+use Symfony\Component\Security\Http\AccessToken\AccessTokenHandlerInterface;
+use Symfony\Component\Security\Http\AccessToken\OAuth2\Exception\InactiveAccessTokenException;
+use Symfony\Component\Security\Http\AccessToken\OAuth2\Exception\MissingClaimException;
+use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+/**
+ * The token handler validates the token on the authorization server.
+ * It uses the Introspection Endpoint as per the RFC7662
```
nicolas-grekas: Missing blank line before: docblock should be formatted as one line for the title, a blank line, then longer description.

src/Symfony/Component/Security/Http/AccessToken/OAuth2/OAuth2Trait.php
```diff
@@ -0,0 +1,53 @@
...
+
+        foreach ($claims as $claim => $value) {
+            unset($claims[$claim]);
+            if ('' === $value || null === $value) {
+                continue;
+            }
+            $claims[u($claim)->camel()->toString()] = $value;
+        }
+
+        if (isset($claims['updatedAt']) && '' !== $claims['updatedAt']) {
```
nicolas-grekas: `if ('' !== ($claims['updatedAt'] ?? '')) {`

src/Symfony/Component/Security/Http/AccessToken/OAuth2/OAuth2Trait.php
```diff
@@ -0,0 +1,53 @@
...
+        if (!\function_exists(\Symfony\Component\String\u::class)) {
+            throw new \LogicException('You cannot use the "OAuth2TokenHandler" since the String component is not installed. Try running "composer require symfony/string".');
+        }
+
+        foreach ($claims as $claim => $value) {
+            unset($claims[$claim]);
+            if ('' === $value || null === $value) {
+                continue;
+            }
+            $claims[u($claim)->camel()->toString()] = $value;
```
nicolas-grekas: not sure this deserves the added dependency to the String component

src/Symfony/Component/Security/Http/AccessToken/OAuth2/OAuth2Trait.php
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        if (isset($claims['updatedAt']) && '' !== $claims['updatedAt']) {
+            $claims['updatedAt'] = (new \DateTimeImmutable())->setTimestamp($claims['updatedAt']);
+        }
+
+        if (\array_key_exists('emailVerified', $claims) && null !== $claims['emailVerified'] && '' !== $claims['emailVerified']) {
```
nicolas-grekas: `if ('' !== ($claims['emailVerified'] ?? '')) {`, same below

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * @see https://tools.ietf.org/html/rfc7662
+ *
+ * @experimental
+ */
+final class Oauth2TokenHandler implements AccessTokenHandlerInterface
+{
+    use OAuth2Trait;
```
nicolas-grekas: this trait is used only once, so let's inline?

src/Symfony/Component/Security/Core/User/OAuth2User.php
```diff
@@ -0,0 +1,132 @@
...
+        private ?string $clientId = null,
+        private ?string $username = null,
+        private ?string $tokenType = null,
+        private ?int $exp = null,
+        private ?int $iat = null,
+        private ?int $nbf = null,
+        private ?string $sub = null,
+        private ?string $aud = null,
+        private ?string $iss = null,
+        private ?string $jti = null,
```
nicolas-grekas: what about making all props public (readonly?) and removing all getters?

Spomky:
Hi @wouterj and @chalasr,

Many thanks for your feedback and advices. Hereafter my thoughts on this

> First a practical thing: We're entering feature freeze very shortly, meaning that we don't merge newly opened feature PRs for 6.3 anymore (unless really needed or very small). So we have some more time to polish this PR for 6.4 :)

No problem. I based it on the branch 6.3, but it does not matter to rebase to another branch.

> We don't have the man power nor knowledge to maintain let's say all OAuth, SAML, Webauthn and MFA standards.

Even if the community is large enough, I understand there is a risk to include specific authentication means that could be a problem in the future.

> When adding new features, I think we must always test it to those 3 conditions

I was not aware of those, but seem to be fair enough.

> Implementing all OAuth2/OIDC features in core is a non-goal for now. Better improve community packages that are already supporting them or are meant to do so.

Noted. Thanks. In this case, there is no need to go further here. I will continue that work in a side project. 👍

src/Symfony/Component/Security/Core/User/OAuth2User.php
```diff
@@ -0,0 +1,132 @@
...
+
+        // Additional Claims ("
+        //    Specific implementations MAY extend this structure with
+        //    their own service-specific response names as top-level members
+        //    of this JSON object.
+        // ")
+        ...$additionalClaims
+    ) {
+        if ((null === $sub || '' === $sub) && (null === $username || '' === $username)) {
+            throw new \InvalidArgumentException('The "sub" and "username" claims cannot be empty.');
```
nicolas-grekas: `You must provide one of "sub" or "username" claims.`

src/Symfony/Component/Security/Core/User/OAuth2User.php
```diff
@@ -0,0 +1,132 @@
...
+        private ?string $aud = null,
+        private ?string $iss = null,
+        private ?string $jti = null,
+
+        // Additional Claims ("
+        //    Specific implementations MAY extend this structure with
+        //    their own service-specific response names as top-level members
+        //    of this JSON object.
+        // ")
+        ...$additionalClaims
```
nicolas-grekas: could this work as a regular array argument?

src/Symfony/Bundle/SecurityBundle/Resources/config/security_authenticator_access_token.php
```diff
@@ -61,5 +62,14 @@
...
             ])
+
+        // OAuth2 Introspection (RFC 7662)
+        ->set('security.access_token_handler.oauth2', Oauth2TokenHandler::class)
+        ->abstract()
+        ->args([
+            abstract_arg('http client'),
+            service('logger')->nullOnInvalid(),
+            'sub',
+        ])
```
nicolas-grekas: missing indentation after "set" line

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OAuth2TokenHandlerFactory.php
```diff
@@ -0,0 +1,77 @@
...
+use Symfony\Component\Config\Definition\Builder\NodeBuilder;
+use Symfony\Component\DependencyInjection\ChildDefinition;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\HttpClient\HttpClient;
+
+/**
+ * Configures a token handler for an Introspection Endpoint.
+ *
+ * @experimental
```
nicolas-grekas: `@internal` instead

Spomky:
Hi,

I revised my mind and I would like to reopen this PR. I think this should be challenged a bit more.
Hereafter a (simplified) diagram showing where this new feature will be useful.

```mermaid
sequenceDiagram
    participant Client
    participant Authorization Server
    participant Resource Server
    Authorization Server->>Client: IDToken+Access Token (after consent+authentication)
    loop PR 48272 acts here
        Client->>Client: IDToken
        Client->>Authorization Server: Access Token (UserInfo Endpoint)
    end
    Client->>Resource Server: Access Token
    loop This PR acts here
        Resource Server->>Resource Server: Access Token (e.g. JWT, PASETO...)
        Resource Server->>Authorization Server: Access Token (Introspection Endpoint)
    end
```

As you can see, the features introduced with #48272 allow a Client application to authenticate end-users (by end-user, the OAuth spec means a person). The Client app can either use the IDToken (JWT) or the Access Token+UserInfo Endpoint to retreive the end-user details.
```mermaid
sequenceDiagram
    participant Client
    participant Authorization Server
    loop PR 48272 acts here
        Client->>Client: IDToken
        Client->>Authorization Server: Access Token (UserInfo Endpoint)
    end
```

This PR acts in a very similar way. The main differences are as follows:
* It is s not meant to be used by the Client, but the Resource Server
* The Resource Server is not supposed to receive the ID Token (unless Client + Resource Server = the same application)
* The Resource Server can directly use the Access Token if it can be verified (e.g. JWT)
* The "user" (consumer of the Resource Server) can be a end-user or an application (e.g. API to API communication)

```mermaid
sequenceDiagram
    participant Resource Server
    participant Authorization Server
    loop This PR acts here
        Resource Server->>Resource Server: Access Token (e.g. JWT, PASETO...)
        Resource Server->>Authorization Server: Access Token (Introspection Endpoint)
    end
```

The reason why I think this PR deserves more attention is that is cover the missing piece for covering most of the OAuth2 use cases.

I understand the point of view you say we could rely community on community works. And that's true that several solutions exist. But here, no third party library with exotic dependencies is needed<sup>1</sup>. We can directly use it with few configuration lines on e.g. ApiPlatform-based resource servers.

Also, I understand that Symfony does not aim at implementing all industry/web standards (even if all the Notifier channels make me think the opposite). However, OAuth2 is so widely used by the industry that it would be a shame not to support it, in particular after the works on the previous PRs. It is not a trendy technology, but a standard widely used for more than 10 years (and in continuous evolution).

At last but not least, the item Support for third-party authentication services (e.g. Auth0: https://auth0.com/) in [[RFC] Symfony Security rework tracking issue](https://github.com/symfony/symfony/issues/30914) proposed by @curry684 could be marked as checked as this feature will allow communication with [auth0/Okta](https://developer.okta.com/docs/reference/api/oidc/#introspect), [OneLogin](https://developers.onelogin.com/openid-connect/api/validate-token) and many more.

---

<sup>1</sup>: this does not mean community works are useless. They all have great features, but having too much external dependencies is something we all try to avoid.

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OAuth2TokenHandlerFactory.php
```diff
@@ -0,0 +1,72 @@
...
+        return 'oauth2';
+    }
+
+    public function addConfiguration(NodeBuilder $node): void
+    {
+        $node
+            ->arrayNode($this->getKey())
+                ->fixXmlConfig($this->getKey())
+                ->children()
+                    ->arrayNode('client')
```
vincentchalamon: [suggestion] Allow to create the client directly here, like in [OidcUserInfoTokenHandlerFactory](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcUserInfoTokenHandlerFactory.php). It would also add consistency with `oidc_user_info` usage.

Spomky: HI @vincentchalamon,

I am not sure to clearly understand. The behavior looks identical. What have I missed?

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OAuth2TokenHandlerFactory.php
Spomky: OK. Done

src/Symfony/Bundle/SecurityBundle/Resources/config/security_authenticator_access_token.php
Spomky: Fixed

src/Symfony/Component/Security/Core/User/OAuth2User.php
Spomky: To be confirmed. As the object is called using the spread operator, this seems to be the safest way.
As it is inspired by the OIDC, the signature of the OIDCUser should also be changed.

src/Symfony/Component/Security/Core/User/OAuth2User.php
Spomky: OK. Done

src/Symfony/Component/Security/Core/User/OAuth2User.php
Spomky: Correct. I changed the design. Looks good to me.

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
Spomky: Fixed

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
Spomky: 👌🏼

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
Spomky: 👌🏼

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
Spomky: 👍🏽

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
Spomky: 👍🏽

src/Symfony/Component/Security/Http/Tests/AccessToken/OAuth2/OAuth2TokenHandlerTest.php
Spomky: Fixed

src/Symfony/Component/Security/Http/Tests/AccessToken/OAuth2/OAuth2TokenHandlerTest.php
Spomky: 👍🏽

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
Spomky: 👍🏽

src/Symfony/Component/Security/Http/AccessToken/OAuth2/OAuth2Trait.php
Spomky: The OIDC handler also uses it. If we can get rid of it, I will change that line. TBD with @vincentchalamon

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OAuth2TokenHandlerFactory.php
vincentchalamon: It's fixed, maybe I read it wrong or you already fixed it. Thanks :-)

Spomky: 👍🏽 Thanks!

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OAuth2TokenHandlerFactory.php
```diff
@@ -0,0 +1,72 @@
...
+namespace Symfony\Bundle\SecurityBundle\DependencyInjection\Security\AccessToken;
+
+use Symfony\Component\Config\Definition\Builder\NodeBuilder;
+use Symfony\Component\DependencyInjection\ChildDefinition;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\HttpClient\HttpClient;
+
+/**
+ * Configures a token handler for an OIDC server.
```
OskarStark: Is this comment correct?

src/Symfony/Component/Security/Http/Tests/AccessToken/OAuth2/OAuth2TokenHandlerTest.php
```diff
@@ -0,0 +1,71 @@
...
+
+        $this->assertEquals(new UserBadge($expected, fn () => $expectedUser, $claims), $userBadge);
+        $this->assertInstanceOf(OAuth2User::class, $actualUser);
+        $this->assertSame($claims, $userBadge->getAttributes());
+        $this->assertSame($claims['sub'], $actualUser->getUserIdentifier());
+    }
+
+    public static function getClaims(): iterable
+    {
+        yield ['Z5O3upPC88QrAjx00dis'];
```
OskarStark: Why a data provider just for one?

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OAuth2TokenHandlerFactory.php
Spomky: Indeed. I updated this comment.

src/Symfony/Component/Security/Http/Tests/AccessToken/OAuth2/OAuth2TokenHandlerTest.php
Spomky: The first iteration of this PR had several use cases and I simplified it to that unique one.
But yes it is not relevant now and I removed it.

Spomky:
Rebased for targeting Symfony 7.1

src/Symfony/Bundle/SecurityBundle/CHANGELOG.md
```diff
@@ -24,12 +24,14 @@ CHANGELOG
...
 ---
 
+ * Add `OAuth2TokenHandlerFactory` for `AccessTokenFactory`
  * Deprecate `Security::ACCESS_DENIED_ERROR`, `AUTHENTICATION_ERROR` and `LAST_USERNAME` constants, use the ones on `SecurityRequestAttributes` instead
  * Allow an array of `pattern` in firewall configuration
  * Add `$badges` argument to `Security::login`
  * Deprecate the `require_previous_session` config option. Setting it has no effect anymore
  * Add `LogoutRouteLoader`
 
+
```
chalasr: should be reverted

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Exception/MissingClaimException.php
```diff
@@ -0,0 +1,27 @@
...
+namespace Symfony\Component\Security\Http\AccessToken\OAuth2\Exception;
+
+use Symfony\Component\Security\Core\Exception\AuthenticationException;
+
+/**
+ * This exception is thrown when a claim is missing or invalid.
+ *
+ * @experimental
+ */
+class MissingClaimException extends AuthenticationException
```
chalasr: I'm not sure we need this very specific exception given it's immediately caught where it's throw and wrapped in a `BadCredentialsException`. I suggest to throw `BadCrendentialsException`s with precise messages where needed instead (see https://github.com/symfony/symfony/pull/50695 - I'll finish it at some point so better not continue this way)

chalasr (approved):
I changed my mind, given we have OAuth/OIDC support let's do it right + as complete as possible. 👍 once the small comments addressed & PR rebased

Spomky:
That's good news!
The PR is now rebased and I addressed the remarks raised earlier.

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OAuth2TokenHandlerFactory.php
```diff
@@ -0,0 +1,72 @@
...
+        $node
+            ->arrayNode($this->getKey())
+                ->fixXmlConfig($this->getKey())
+                ->children()
+                    ->arrayNode('client')
+                        ->info('HttpClient to call the Introspection Endpoint.')
+                          ->isRequired()
+                        ->beforeNormalization()
+                            ->ifString()
+                           ->then(static function ($v): array { return ['id' => $v]; })
```
nicolas-grekas:
```suggestion
                            ->then(static fn ($v) => ['id' => $v])
```

src/Symfony/Bundle/SecurityBundle/Resources/config/security_authenticator_access_token.php
```diff
@@ -135,5 +136,13 @@
...
         ->set('security.access_token_handler.oidc.signature.PS512', PS512::class)
             ->tag('security.access_token_handler.oidc.signature_algorithm')
+
+    // OAuth2 Introspection (RFC 7662)
+    ->set('security.access_token_handler.oauth2', Oauth2TokenHandler::class)
+        ->abstract()
+        ->args([
+            abstract_arg('http client'),
+            service('logger')->nullOnInvalid(),
+        ])
```
nicolas-grekas: indentation is off by one

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OAuth2TokenHandlerFactory.php
```diff
@@ -0,0 +1,72 @@
...
+
+        // Create the client service
+        if (!isset($config['client']['id'])) {
+            $clientDefinitionId = 'http_client.security.access_token_handler.oauth2';
+            if (!ContainerBuilder::willBeAvailable('symfony/http-client', HttpClient::class, ['symfony/security-bundle'])) {
+                $container->register($clientDefinitionId, 'stdClass')
+                    ->addError('You cannot use the "oauth2" token handler since the HttpClient component is not installed. Try running "composer require symfony/http-client".');
+            } else {
+                $container->register($clientDefinitionId, HttpClient::class)
+                    ->setFactory([HttpClient::class, 'create'])
```
nicolas-grekas: we should use the existing http_client service instead, so that we get integration with the profiler for free
or is there a reason why this is a separate service?

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OAuth2TokenHandlerFactory.php
```diff
@@ -0,0 +1,72 @@
...
+        $tokenHandlerDefinition = $container->setDefinition($id, new ChildDefinition('security.access_token_handler.oauth2'));
+
+        // Create the client service
+        if (!isset($config['client']['id'])) {
+            $clientDefinitionId = 'http_client.security.access_token_handler.oauth2';
+            if (!ContainerBuilder::willBeAvailable('symfony/http-client', HttpClient::class, ['symfony/security-bundle'])) {
+                $container->register($clientDefinitionId, 'stdClass')
+                    ->addError('You cannot use the "oauth2" token handler since the HttpClient component is not installed. Try running "composer require symfony/http-client".');
+            } else {
+                $container->register($clientDefinitionId, HttpClient::class)
```
nicolas-grekas: if we need that service, it should be defined in a service-declaration file

src/Symfony/Component/Security/Core/User/OAuth2User.php
```diff
@@ -0,0 +1,72 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Security\Core\User;
+
+/**
+ * UserInterface implementation used by the access-token security workflow with an OIDC server.
+ *
+ * @experimental
```
nicolas-grekas: not sure this is useful, the API is standard

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
```diff
@@ -0,0 +1,97 @@
...
+            // Call the Authorization server to retrieve the resource owner details
+            // If the token is invalid or expired, the Authorization server will return an error
+            $claims = $this->client->request('POST', '', [
+                'body' => [
+                    'token' => $accessToken,
+                    'token_type_hint' => 'access_token',
+                ],
+            ])->toArray();
+
+            if (empty($claims['sub']) && empty($claims['username'])) {
```
nicolas-grekas: this and adjust below to use the new vars:
```suggestion
            $sub = \is_array($claims['sub']) ? $claims['sub'] : null;
            $username = \is_array($claims['username']) ? $claims['username'] : null;
            if (!$sub && !$username)) {
```

src/Symfony/Component/Security/Http/AccessToken/Oidc/OidcTokenHandler.php
```diff
@@ -85,19 +84,16 @@ public function getUserBadgeFrom(string $accessToken): UserBadge
...
                 new Checker\ExpirationTimeChecker(clock: $this->clock, allowedTimeDrift: 0, protectedHeaderOnly: false),
                 new Checker\AudienceChecker($this->audience),
                 new Checker\IssuerChecker($this->issuers),
-            ];
-            $claimCheckerManager = new ClaimCheckerManager($checkers);
-            // if this check fails, an InvalidClaimException is thrown
-            $claimCheckerManager->check($claims);
+            ]))->check($claims);
 
             if (empty($claims[$this->claim])) {
```
nicolas-grekas: empty is dangerous and we try to avoid using it

nicolas-grekas (commented):
Here are some CS suggestions, please rebase also.

src/Symfony/Component/Security/Http/Tests/AccessToken/OAuth2/OAuth2TokenHandlerTest.php
```diff
@@ -0,0 +1,63 @@
...
+
+        $clientMock = $this->createMock(HttpClientInterface::class);
+        $clientMock->expects($this->once())
+            ->method('request')->with('POST', '', [
+                'body' => [
+                    'token' => $accessToken,
+                    'token_type_hint' => 'access_token',
+                ],
+            ])
+            ->willReturn($responseMock);
```
welcoMattic: Why not using `MockHttpClient` here? https://symfony.com/doc/current/http_client.html#http-client-and-responses

Spomky: Indeed. MockHttpClient is now used in the last revision of this PR.
Thank you.

src/Symfony/Bundle/SecurityBundle/Resources/config/security_authenticator_access_token.php
Spomky: Indentation should be fine now.

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OAuth2TokenHandlerFactory.php
Spomky: Indeed. This is now gone.

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OAuth2TokenHandlerFactory.php
Spomky: No need of it after all.

src/Symfony/Component/Security/Core/User/OAuth2User.php
Spomky: Indeed. Gone.

src/Symfony/Component/Security/Http/AccessToken/OAuth2/Oauth2TokenHandler.php
Spomky: Thank you for the review.
This part is now up to date.

src/Symfony/Component/Security/Http/AccessToken/Oidc/OidcTokenHandler.php
Spomky: Should be fine in the last revision of this PR.
Thanks.

fabpot (approved):

fabpot:
Thank you @Spomky.

PR #50030, opened 2023-04-16
Add new twig bridge function to generate impersonation path

PhilETaylor:
| Q             | A
| ------------- | ---
| Branch?       | 6.3 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| License       | MIT

# Before this PR
So we already have impersonation features in Symfony (https://symfony.com/doc/current/security/impersonating_user.html) and we have two twig helper functions `impersonation_exit_url` and `impersonation_exit_path ` which both work with the configuration parameter for the switch user. 

If the developer changes the switch parameter (`_switch_user`), then these helper functions will dynamically update the `_switch_user=_exit` type urls/paths.

However, to switch TO a user, hand crafted urls with `?_switch_user=MYIDENTIFIER` like `http://example.com/somewhere?_switch_user=thomas` need to be hand crafted currently. 

# The problem

if we now go and change `_switch_user` to be something else, like `_want_to_be_this_user ` in the Symfony configuration (Because the boss told us to do that), then all our exit path/urls will dynamically update, but our hard coded ?_switch_user=MYIDENTIFIER` will stop working.

# The solution this PR provides

The solution this PR provides is to provide a new Twig Helper function for the impersonation path only, taking into account the configured value in Symfony config of the parameter (default is still `_switch_user ` but can be anything like `_want_to_be_this_user` as per the docs) 

This new twig function can be used as such:

```twig
<a href="{{ impersonation_path('mike') }}">Impersonate Mike</a>
```

This would output `?_want_to_be_this_user=mike` or if the default parameter still used would be `?_switch_user=mike`

The PR repurposes the existing code to generate the paths and is backward compatible.

### DIFF
--- a/src/Symfony/Bridge/Twig/Extension/SecurityExtension.php
+++ b/src/Symfony/Bridge/Twig/Extension/SecurityExtension.php
@@ -69,12 +69,22 @@ final class SecurityExtension extends AbstractExtension
         return $this->impersonateUrlGenerator->generateExitPath($exitTo);
     }
 
+    public function getImpersonatePath(string $identifier = null): string
+    {
+        if (null === $this->impersonateUrlGenerator) {
+            return '';
+        }
+
+        return $this->impersonateUrlGenerator->generateImpersonationPath($identifier);
+    }
+
     public function getFunctions(): array
     {
         return [
             new TwigFunction('is_granted', $this->isGranted(...)),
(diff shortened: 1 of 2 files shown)


fabpot (approved):

fabpot:
Thank you @PhilETaylor.

PR #50031, opened 2023-04-16
[Routing] Convert BackedEnums passed as controller action parameters to their value

### LINKED ISSUE #49203: [Router] Can't put a default Enum value in a controller action

6.2, possibly 6.1

### Description

When defining routes with enums, the only way to make a default parameter work is by defining it in the `#[Route]` attribute. You can't define it in PHP:

```php
#[Route('/nullable/{postEnum}')]
public function enumRouteNullable(?PostEnum $postEnum = null): Response
{
    // works without parameter, but no default value
    return $this->render('home/enumRoute.html.twig', ['enum' => $postEnum]);
}

#[Route('/default/{postEnum}')]
public function enumRouteDefault(PostEnum $postEnum = PostEnum::VLOG): Response
{
    // doesn't compile, `Symfony\Component\Routing\Route cannot contain objects.`
    return $this->render('home/enumRoute.html.twig', ['enum' => $postEnum]);
}

#[Route('/default/{postEnum}', defaults: ['postEnum' => PostEnum::VLOG->value])]
public function enumRouteDefaultString(PostEnum $postEnum): Response
{
    // works, with proper default value
    return $this->render('home/enumRoute.html.twig', ['enum' => $postEnum]);
}
```

The second solution is the most convenient one, but is not allowed, I get this error: `Symfony\Component\Routing\Route cannot contain objects.` (screenshot below)

### How to reproduce

I made a test project that crashes accordingly. One controller with the use case, and another controller checking that it's valid with a pure string parameter.

https://github.com/florentdestremau/enum-router-bug

### Possible Solution

It goes back to the fact that Backed Enum classes have no __toString() method and this makes the router compiling crash. There is some treatment for route generation that should detect enums and call the `->value` method.

### Additional Context

image

nalabdou:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #49203 
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Routing/Loader/AnnotationClassLoader.php
+++ b/src/Symfony/Component/Routing/Loader/AnnotationClassLoader.php
@@ -208,7 +208,11 @@ abstract class AnnotationClassLoader implements LoaderInterface
             }
             foreach ($paths as $locale => $path) {
                 if (preg_match(sprintf('/\{%s(?:<.*?>)?\}/', preg_quote($param->name)), $path)) {
-                    $defaults[$param->name] = $param->getDefaultValue();
+                    if (\is_scalar($defaultValue = $param->getDefaultValue()) || null === $defaultValue) {
+                        $defaults[$param->name] = $defaultValue;
+                    } elseif ($defaultValue instanceof \BackedEnum) {
+                        $defaults[$param->name] = $defaultValue->value;
+                    }
                     break;
                 }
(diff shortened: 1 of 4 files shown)


nicolas-grekas:
Thank you @nalabdou.

PR #50040, opened 2023-04-17
[Translation] Crowdin Bridge: Fix locale vs LanguageId

pculka:
| Q             | A
| ------------- | ---
| Branch?       | 6.2 
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #... 
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Translation/Bridge/Crowdin/CrowdinProvider.php
+++ b/src/Symfony/Component/Translation/Bridge/Crowdin/CrowdinProvider.php
@@ -109,6 +109,8 @@ final class CrowdinProvider implements ProviderInterface
         $translatorBag = new TranslatorBag();
         $responses = [];
 
+        $localeLanguageMap = $this->mapLocalesToLanguageId($locales);
+
         foreach ($domains as $domain) {
             $fileId = $this->getFileIdByDomain($fileList, $domain);
 
@@ -118,7 +120,7 @@ final class CrowdinProvider implements ProviderInterface
 
             foreach ($locales as $locale) {
                 if ($locale !== $this->defaultLocale) {
-                    $response = $this->exportProjectTranslations($locale, $fileId);
+                    $response = $this->exportProjectTranslations($localeLanguageMap[$locale], $fileId);
                 } else {
                     $response = $this->downloadSourceFile($fileId);
                 }
@@ -402,5 +404,39 @@ final class CrowdinProvider implements ProviderInterface
         }
 
         return $result;
+    }
+
+    private function mapLocalesToLanguageId(array $locales): array
+    {
+        /**
+         * We cannot query by locales, we need to fetch all and filter out the relevant ones.
+         *
+         * @see https://developer.crowdin.com/api/v2/#operation/api.languages.getMany (Crowdin API)
+         * @see https://developer.crowdin.com/enterprise/api/v2/#operation/api.languages.getMany (Crowdin Enterprise API)
+         */
+        $response = $this->client->request('GET', '../../languages?limit=500');
+
+        if (200 !== $response->getStatusCode()) {
+            throw new ProviderException('Unable to list set languages.', $response);
+        }
+
+        $localeLanguageMap = [];
+        foreach ($response->toArray()['data'] as $language) {
+            foreach (['locale', 'osxLocale', 'id'] as $key) {
+                if (\in_array($language['data'][$key], $locales)) {
+                    $localeLanguageMap[$language['data'][$key]] = $language['data']['id'];
+                }
+            }
+        }
+
+        if (\count($localeLanguageMap) !== \count($locales)) {
+            $message = implode('", "', array_diff($locales, array_keys($localeLanguageMap)));
+            $message = sprintf('Unable to find all requested locales: "%s" not found.', $message);
+            $this->logger->error($message);
+
+            throw new ProviderException($message, $response);
+        }
+
+        return $localeLanguageMap;
     }
 }
--- a/src/Symfony/Component/Translation/Bridge/Crowdin/Tests/CrowdinProviderTest.php
+++ b/src/Symfony/Component/Translation/Bridge/Crowdin/Tests/CrowdinProviderTest.php
@@ -632,6 +632,30 @@ XLIFF
                     ],
                 ]));
             },
+            'listLanguages' => function (string $method, string $url, array $options = []): ResponseInterface {
+                $this->assertSame('GET', $method);
+                $this->assertSame('https://api.crowdin.com/api/v2/languages?limit=500', $url);
+                $this->assertSame('Authorization: Bearer API_TOKEN', $options['normalized_headers']['authorization'][0]);
+
+                return new MockResponse(json_encode([
+                    'data' => [
+                        [
+                            'data' => [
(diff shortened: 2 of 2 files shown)


src/Symfony/Component/Translation/Bridge/Crowdin/CrowdinProvider.php
```diff
@@ -109,6 +109,25 @@ public function read(array $domains, array $locales): TranslatorBag
         $translatorBag = new TranslatorBag();
         $responses = [];
 
+        $localeLanguageMap = [];
+        foreach ($this->listLanguages($locales) as $language) {
```
andrii-bodnar: this method doesn't accept any arguments

src/Symfony/Component/Translation/Bridge/Crowdin/CrowdinProvider.php
```diff
@@ -118,7 +137,7 @@ public function read(array $domains, array $locales): TranslatorBag
 
             foreach ($locales as $locale) {
                 if ($locale !== $this->defaultLocale) {
-                    $response = $this->exportProjectTranslations($locale, $fileId);
+                    $response = $this->exportProjectTranslations($localeLanguageMap[$locale], $fileId);
```
andrii-bodnar, line 123: The same issue probably is also relevant for the `uploadTranslations` method

src/Symfony/Component/Translation/Bridge/Crowdin/CrowdinProvider.php
andrii-bodnar: Probably it would be better to extract this to a separate method, e.g. `mapLocaleToLanguageId` or something like this

andrii-bodnar:
Hi @pculka, could you please provide locale examples than currently can't be synced with Crowdin?

andrii-bodnar:
The failed Unit tests do not expect the GET languages call, you should include the mock for this call in the tests

pculka:
Hi @andrii-bodnar - In our project we use full locale codes, not language codes. For example for Germany the full locale is de-DE. This is in languages endpoint of Crowdin is stated as locale attribute:

```
"data": {
  "id": "de",
  "name": "German",
  "editorCode": "de",
  "twoLettersCode": "de",
  "threeLettersCode": "deu",
  "locale": "de-DE",
  "androidCode": "de-rDE",
  "osxCode": "de.lproj",
  "osxLocale": "de",
  "pluralCategoryNames": [
      "one",
      "other"
  ],
  "pluralRules": "(n != 1)",
  "pluralExamples": [
      "1",
      "0, 2-999; 1.2, 2.07..."
  ],
  "textDirection": "ltr",
  "dialectOf": null
}
```
When downloading translations in your code - you also properly annotated that the requested parameter has to be LanguageID, but you pass in locale and that fails. If I pass locale, I have to find the correct language ID by the locale and then the languages will be downloaded.

pculka:
I'll try to add GET for the mock

andrii-bodnar:
@pculka got it, the current approach might be unreliable since a user can use a different language code format and in this case it's hard to guess the right language ID. I would suggest encouraging users to use [Language Mapping](https://support.crowdin.com/project-settings/#:~:text=Adding%20Custom%20Language%20Codes) in this case. 

**The flow:**

- a user has some locale that is different from the default Crowdin language ID
- a user set up the Language mapping for this locale in the Crowdin project
- then, on the Provider side you can use the [Get Project](https://developer.crowdin.com/api/v2/#operation/api.projects.get) to get the `languageMapping` property and find the language ID for the mapped locale.
- use this language ID for requests.

This will not break any existing integrations and will allow customizing the used language codes.

By the way, something similar is suggested for the other translation provider here - https://symfony.com/doc/current/translation.html#installing-and-configuring-a-third-party-provider

pculka:
Well I think the core problem lies in misunderstanding and misdocumentation of how to use locales should be used within symfony itself. It tends to be easy to confuse code vs locale vs osxlocale, etc. That's why I ended up here, as I was using proper i18n locale string, but in communication with CrowdIn the expected parameters are Language IDs.
Hence in your suggestion you suggest MIX UP between locale and language id once again. I mean - it is doable, but not sure if that's the right thing to do.

Anyway I've added the mapping, in get project info I now have:
```
        "languageMapping": {
            "de": {
                "locale": "de-DE"
            },
            "hu": {
                "locale": "hu-HU"
            }
        },
```

You suggest I change this code I provided to check for the project languageMapping setting and just fix it that way, right?

andrii-bodnar:
@pculka does Symfony has some language codes list that is recommended to use for translation catalogs?

pculka:
I doubt it. both 'de' and 'de-DE' are valid locale strings, both referencing german language, but -DE specifies also the region which is optional. I mean. Meaning de-DE should try automatically a fallback maybe? Not sure how to tackle this situation properly.

andrii-bodnar:
Well, the Language Mapping approach might be more complicated. Probably we can keep the suggested approach in this case and not overcomplicate this for now

pculka:
added one more fallback for underscore locale (you had that in one of your tests), and tests are passing.
Hopefully someone merges this

src/Symfony/Component/Translation/Bridge/Crowdin/CrowdinProvider.php
pculka: Actually from my observation the mapping is done on Crowdin side and works (i.e. de-DE source gets properly mapped to de)

pculka:
and refactored - extracted the mapping function out

pculka:
@nicolas-grekas please can you check this PR? I also noticed a failed test in appveyor, 1) Symfony\Component\Messenger\Bridge\Redis\Tests\Transport\RedisExtIntegrationTest::testGetAfterReject

something in terms of redis. Maybe there is a problem with redis.dll for windows, which is not compiled for 8.2 - If it is needed, I compiled a working redis.dll for 8.2.5 for x64 both ts and nts versions: https://github.com/pculka/phpredis/releases/tag/5.3.7

But if it is possible to merge without passing appveyor, it would be great to get this done (and if @andrii-bodnar agrees)

src/Symfony/Component/Translation/Bridge/Crowdin/CrowdinProvider.php
```diff
@@ -330,6 +332,26 @@ private function downloadSourceFile(int $fileId): ResponseInterface
...
     }
 
+    private function listLanguages(): array
+    {
+        /**
+         * @see https://developer.crowdin.com/api/v2/#operation/api.languages.getMany (Crowdin API)
+         * @see https://developer.crowdin.com/enterprise/api/v2/#operation/api.languages.getMany (Crowdin Enterprise API)
+         */
+        // cannot query by locales, we need to fetch all and filter out the relevant ones
+        $response = $this->client->request('GET', '../../languages', [
```
nicolas-grekas: let's make this simpler and inline the query string: `request('GET', '../../languages?limit=500');`

nicolas-grekas (approved):
I'm going to trust you on this one :)

nicolas-grekas:
Thank you @pculka.

Issue #49808, opened 2023-03-25
[Serializer] Support `readonly` classes or classes that use readonly for constructor property promotion


ruudk:
### Description

The Serializer is perfectly capable of deserializing the following object without configuration:
```php
final class CreateUserCommand
{
    private function __construct(
        public string $firstName,
        public string $lastName,
    ) {}
}
```

But as soon as you make the class `readonly` (PHP 8.2 feature) or make the constructor properties `readonly` (PHP 8.1) you'll get an error:

```
Error : Cannot initialize readonly property CreateUserCommand::$firstName from scope Symfony\Component\PropertyAccess\PropertyAccessor
 /Volumes/CS/www/cosmos/vendor/symfony/property-access/PropertyAccessor.php:530
```

```php
final readonly class CreateUserCommand
{
    private function __construct(
        public string $firstName,
        public string $lastName,
    ) {}
}
```

```php
final class CreateUserCommand
{
    private function __construct(
        public readonly string $firstName,
        public readonly string $lastName,
    ) {}
}
```

ghost:
I'm on PHP 8.1 and I write commands that are (de)serialized with `public readonly` promoted constructor properties and it's working fine. Are you sure you somehow didn't mess with the serializer configuration?

ruudk:
Thanks for your reply 😊
I'm not using any Serializer configuration at all. Do you? If so? What are you using? Do you maybe pass `$context`?

ghost:
I've just created a symfony/skeleton:^5.4 project, added symfony/serializer and symfony/property-access and successfully (de)serialized class with readonly properties.

<details>
  <summary>See code</summary>

```php
<?php

namespace App;

use Symfony\Component\Console\Attribute\AsCommand;
use Symfony\Component\Console\Command\Command;
use Symfony\Component\Console\Input\InputInterface;
use Symfony\Component\Console\Output\OutputInterface;
use Symfony\Component\Serializer\SerializerInterface;

#[AsCommand(name: 'my-command')]
class MyCommand extends Command
{
    function __construct(
        private SerializerInterface $serializer
    ){
        parent::__construct();
    }

    protected function execute(InputInterface $input, OutputInterface $output)
    {
        $a = new A('a', 1);
        $serialized = $this->serializer->serialize($a, 'json');
        dump($serialized);
        $deserialized = $this->serializer->deserialize($serialized, A::class, 'json');
        dump($deserialized);

        return 1;
    }
}

class A {
    function __construct(
[...]

PR #50042, opened 2023-04-17
[Serializer] Convert names using denormalize during denormalization

wouterj:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Might be related to #49808 
| License       | MIT
| Doc PR        | -

This code is called during denormalization, but it was wrongly calling the normalize function of the name converter. This bug exists since the start of name converters (e14854fe22858648053685f2e5df4493f148eb14). I believe we never caught this bug as the `ObjectNormalizer` would simply set the property using the name converter in the correct direction after initialization.

However, with the recent `readonly` feature this does not work as the constructor sets it to a value and thus the fallback of `ObjectNormalizer` errors with "Cannot modify readonly property".

It might be a bit risky to change this direction in an LTS release, but that is only because people were relying on buggy behavior.

### DIFF
--- a/src/Symfony/Component/Serializer/Tests/Fixtures/Php80ReadonlyDummy.php
+++ b/src/Symfony/Component/Serializer/Tests/Fixtures/Php80ReadonlyDummy.php
@@ -1,0 +1,20 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Serializer\Tests\Fixtures;
+
+class Php80ReadonlyDummy
+{
+    public function __construct(
+        public readonly string $someValue
+    ) {
+    }
+}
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Any way to test this?

src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
```diff
@@ -351,7 +351,7 @@ protected function instantiateObject(array &$data, string $class, array &$contex
             $params = [];
             foreach ($constructorParameters as $constructorParameter) {
                 $paramName = $constructorParameter->name;
-                $key = $this->nameConverter ? $this->nameConverter->normalize($paramName, $class, $format, $context) : $paramName;
+                $key = $this->nameConverter ? $this->nameConverter->denormalize($paramName, $class, $format, $context) : $paramName;
```
ogizanagi: Not sure about your issue, but this bugs me out:
I think the code was correct ; for each of the constructor properties, it attempts to get the normalized name, to use as key in the `$data` array.

Calling `denormalize` would have been done the other way around: starting from `$data` keys, call `denormalize` on each to get the property name.

wouterj: You were right, the test proves that this code is correct (if the input is not already converted).

wouterj:
I've added a test (which is more like an integration test than a unit test, but that seems "normal" in the `AbstractNormalizerTest`).

This proves that I was wrong and the logic is correct. So this is no longer a bug, but we do not support inputs that already have the correct name format as input to normalization.

wouterj:
I'll close for now, as I have no time to investigate how to support this case further. If someone has, you can take my commit as a starting point.

PR #39622, opened 2020-12-23
[Messenger] Be able to get raw data when a message in not decodable by the PHP Serializer

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | yes
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

---

It's easier to review the pr with [`?w=1`](https://github.com/symfony/symfony/pull/39622/files?w=1)

---

## Why?

This PR aimed to handle properly messaged that are already failed and also **not decodable** anymore.

This use case occurs when a Message class is renamed / moved to another namespace

## Reproducer

1. Publish many message of type `App\Message\Foobar`
1. Consume them all, but throw an Exception exception in the handler
1. Renamed `Foobar` to `Foo`

##### Without an ID

```
…abs/symfony/messenger-class-not-found(new) bin/console  messenger:failed:show
There are 94 messages pending in the failure transport.

In PhpSerializer.php line 84:

  Message class "App\Message\Foobar" not found during decoding.


messenger:failed:show [--max MAX] [--transport [TRANSPORT]] [--stats] [--class-filter CLASS-FILTER] [--] [<id>]
```

And this **message is lost**

##### With an ID

```
>…abs/symfony/messenger-class-not-found(new) bin/console  messenger:failed:show 68
There are 93 messages pending in the failure transport.

In PhpSerializer.php line 84:

  Message class "App\Message\Foobar" not found during decoding.


messenger:failed:show [--max MAX] [--transport [TRANSPORT]] [--stats] [--class-filter CLASS-FILTER] [--] [<id>]
```

**Message lost too**

##### Without an ID

```
>…goire/dev/labs/symfony/symfony-5.2(remaned) bin/console  messenger:failed:show
There are 94 messages pending in the failure transport.
 ----- ------------------------ --------------------- ---------------------------------------
  Id    Class                    Failed at             Error
 ----- ------------------------ --------------------- ---------------------------------------
  62    __PHP_Incomplete_Class   2022-08-12 16:26:06   FoobarHandler is not implemented yet.
  63    __PHP_Incomplete_Class   2022-08-12 16:26:06   FoobarHandler is not implemented yet.
  64    __PHP_Incomplete_Class   2022-08-12 16:26:06   FoobarHandler is not implemented yet.

 // Showing first 50 messages.

 // Run messenger:failed:show {id} --transport=failed -vv to see message details.

```

**No messages are lost**

##### With an ID

```
[...]

PR #50043, opened 2023-04-17
[Messenger] Proper message decoding error handling

### LINKED ISSUE #44117: [Messenger] MessageDecodingFailedException should not delete message from queue

All

### Description

`MessageDecodingFailedException` should not delete a message from the queue.

```
try {
    $envelope = $this->serializer->decode([
        'body' => $sqsEnvelope['body'],
        'headers' => $sqsEnvelope['headers'],
    ]);
} catch (MessageDecodingFailedException $exception) {
    $this->connection->delete($sqsEnvelope['id']);

    throw $exception;
}
```

### Issues it causes

- Since decoding is not exactly reverse of encoding with arrays (int vs string keys), legitimate messages end up being dropped.
- Any other kind of dev ops error can mess up the decoding temporarily.

alex-dev:
| Q | A
|---|---
| Branch? | 6.3
| Bug fix? | yes
| New feature? | no
| Deprecations? | no
| BC breaks?  | yes
| Tickets | Fix #44117
| License | MIT
| Doc PR | N/A

- [ ] Update CHANGELOG.md
- [x] Update whatever is needed for BC breaks
- [x] Ensure serialization works across a full failing message (failed decoding into encoding)

## Problem
Handling deserialization in transports prevents actual error handling from running... And depending on transports, can even drop messages. While envelope and stamps deserialization failures may be actual error warranting undefined behavior, message denormalization failures is merely an application concern. Symfony should never drop messages that are broken by the application.

## Solution
Reusing some part of #39622:
- Ensure both `PhpSerializer` and `Serializer` do not throw error in message deserialization
  - Stamps and envelope are still allowed to throw.
  - Remove toggling incomplete classes from `PhpSerializer`
- Handle `MessageDecodingFailedStamp` in `Worker`
  - Trigger standard message rejection (event and retry into full rejection)
- Ensure compatibility with `AbstractFailedMessagesCommand`

## BC breaks
- `Messenger\Transport\Serialization\Serializer::__construct()` first arguments goes from `?SerializerInterface $serializer = null` to `DecoderInterface&DenormalizerInterface&SerializerInterface $serializer`
  - To support distinguishing between encoding errors and normalization errors, we must call each separately. This requires an intersection of interfaces
  - Framework always pass a constructed serializer, so no issue
  - No other instantiations are done... So only BC breaks will be in standalone uses passing null as first argument
  - `Messenger\Transport\Serialization\Serializer::create()` still return a default constructed instance


src/Symfony/Component/Messenger/Transport/Serialization/Serializer.php
```diff
@@ -69,20 +70,24 @@ public function decode(array $encodedEnvelope): Envelope
...
         }
 
         try {
-            $message = $this->serializer->deserialize($encodedEnvelope['body'], $encodedEnvelope['headers']['type'], $this->format, $context);
-        } catch (ExceptionInterface $e) {
-            throw new MessageDecodingFailedException('Could not decode message: '.$e->getMessage(), $e->getCode(), $e);
+            $message = $this->serializer->denormalize($body, $type, $this->format, $context);
+        } catch (\Throwable $e) {
+            $stamps[] = new MessageDecodingFailedStamp($e->getMessage());
+            $message = $this->serializer->denormalize($body, \stdClass::class, $this->format, $context);
```
nikophil, line 90: not sure it is very common, but what happens if a handler consumes messages of class `\stdClass::class`? or if `allow_no_handlers` is activated?

FMI, why do we need to separate the two behaviors, when decoding fails and when denormalization fails?

alex-dev: - Anything related to handlers does not matter here. We handle the denormalization failure using stamps long before handlers.
- We split decoding and denormalization to match `PhpSerializer` behavior when it allows incomplete classes: an `Envelope` must be generated, message must be a valid object.

alex-dev:
@lyrixx Here is a fix for #44117

ismail1432:
Hello @alex-dev,
I think this fix is important, why the PR was closed?

alex-dev:
I could not justify spending more time on this to update it. Seeing Symfony had  little interest in this fix, my company chose to implement a solution inside the transport layer. It is hacked, but the right behavior is guaranteed.
If thee is clear interest in this, I may reopen in January.

B-Galati:
@alex-dev any chance you could share your solution?

alex-dev:
We just reimplemented the transport and made rejection a no-op. There are a few annoying caveats. 
- Retries and failures must be handled by the queue itself (SQS does).
- You lose a lot of debugging utilities Symfony gives you (you must unconfigure failure handling from Symfony)

Other alternative would be to reimplement the worker and associated commands (what this PR more or less does).

PR #50044, opened 2023-04-17
[HttpClient] Add `JsonMockResponse`

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

This tiny DX feature eases the creation of mocked JSON responses which is a common and repetitive use case. 

Before:
```php
new MockResponse(json_encode([
    'foo' => 'bar'
]), [
    'response_headers' => [
        'content-type' => 'application/json',
    ],
]);
```

After:
```php
new JsonMockResponse([
    'foo' => 'bar'
]);
```

### DIFF
--- a/src/Symfony/Component/HttpClient/Response/JsonMockResponse.php
+++ b/src/Symfony/Component/HttpClient/Response/JsonMockResponse.php
@@ -1,0 +1,30 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\HttpClient\Response;
+
+use Symfony\Component\HttpClient\Exception\InvalidArgumentException;
+
+class JsonMockResponse extends MockResponse
+{
+    public function __construct(mixed $body = [], array $info = [])
+    {
+        try {
+            $json = json_encode($body, \JSON_THROW_ON_ERROR);
+        } catch (\JsonException $e) {
+            throw new InvalidArgumentException('JSON encoding failed: '.$e->getMessage(), $e->getCode());
+        }
+
+        $info['response_headers']['content-type'] ??= 'application/json';
+
+        parent::__construct($json, $info);
+    }
+}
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/HttpClient/Response/JsonMockResponse.php
```diff
@@ -0,0 +1,31 @@
...
+    public function __construct(mixed $body = [], array $info = [])
+    {
+        try {
+            $json = json_encode($body, \JSON_THROW_ON_ERROR);
+        } catch (\JsonException $e) {
+            throw new InvalidArgumentException('JSON encoding failed: '.$e->getMessage(), $e->getCode());
+        }
+
+        $info['response_headers'] ??= [];
+        $info['response_headers'] += ['content_type' => 'application/json'];
```
nicolas-grekas: shouldn't this be `content-type` and not `content_type`?

src/Symfony/Component/HttpClient/Response/JsonMockResponse.php
```diff
@@ -0,0 +1,31 @@
...
+    public function __construct(mixed $body = [], array $info = [])
+    {
+        try {
+            $json = json_encode($body, \JSON_THROW_ON_ERROR);
+        } catch (\JsonException $e) {
+            throw new InvalidArgumentException('JSON encoding failed: '.$e->getMessage(), $e->getCode());
+        }
+
+        $info['response_headers'] ??= [];
+        $info['response_headers'] += ['content-type' => 'application/json'];
```
derrabus: This array merge operation looks unnecessarily complicated.

```suggestion
        $info['response_headers']['content-type'] = 'application/json';
```

fancyweb: The idea is we don't override the content-type if it's already set.

derrabus: Oh right, sorry. 😓 

```suggestion
        $info['response_headers']['content-type'] ??= 'application/json';
```

nicolas-grekas:
Thank you @fancyweb.

src/Symfony/Component/HttpClient/Response/JsonMockResponse.php
```diff
@@ -0,0 +1,30 @@
...
+use Symfony\Component\HttpClient\Exception\InvalidArgumentException;
+
+class JsonMockResponse extends MockResponse
+{
+    public function __construct(mixed $body = [], array $info = [])
+    {
+        try {
+            $json = json_encode($body, \JSON_THROW_ON_ERROR);
+        } catch (\JsonException $e) {
+            throw new InvalidArgumentException('JSON encoding failed: '.$e->getMessage(), $e->getCode());
```
stof, line 23: We should attach the original exception as previous exception IMO

PR #47419, opened 2022-08-29
[VarDumper] [bugfix] Make possible to create sync tcp-connection

nikserg:
| Q             | A
| ------------- | ---
| Branch?       |  5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #46145
| License       | MIT
| Doc PR        | 

On some systems creating async TCP connection results in "failed to open stream: Resource temporarily unavailable" exception. To solve this, added `VAR_DUMPER_ASYNC` variable, which is true by default, so won't affect default behavior.

If one faces "failed to open stream: Resource temporarily unavailable" error, he should add `VAR_DUMPER_ASYNC=0` to `.env` file, and problem would be solved.

PR #50047, opened 2023-04-18
[VarDumper] Make the server TCP connection sync

### LINKED ISSUE #46145: [VarDumper] Server socket problem on OS X

6.0

### Description

I have a small code project (cf reproducer) that works fine on Dell XPS intel with ubuntu and breaks on last M1 mac.

Looks like I have a `stream_socket_sendto(): Socket is not connected` error when I tried to figure it out. What is weird is that the script can work "sometimes" which means that IMHO it can't be a configuration error.

This bug may be present on older versions as well, but I didn't test it on older versions.

### How to reproduce

I created a reproducer 

https://github.com/FabienPapet/test-php with all installation instructions

### Possible Solution

_No response_

### Additional Context

_No response_

ogizanagi:
| Q             | A
| ------------- | ---
| Branch?       | 5.4 
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #46145 
| License       | MIT
| Doc PR        | N/A

Alternative to https://github.com/symfony/symfony/pull/47419, since the async mode is not consistently working and I agree with Nicolas [about not adding a new option](https://github.com/symfony/symfony/pull/47419#issuecomment-1349051704) to control this. Let's always go non-async.

### DIFF
--- a/src/Symfony/Component/VarDumper/Server/Connection.php
+++ b/src/Symfony/Component/VarDumper/Server/Connection.php
@@ -91,7 +91,7 @@ class Connection
     {
         set_error_handler([self::class, 'nullErrorHandler']);
         try {
-            return stream_socket_client($this->host, $errno, $errstr, 3, \STREAM_CLIENT_CONNECT | \STREAM_CLIENT_ASYNC_CONNECT);
+            return stream_socket_client($this->host, $errno, $errstr, 3);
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Thank you @ogizanagi.

Issue #15544, opened 2015-08-13
[Bug] [Form] 2.7 - Form collection prototype required option


jeremywaguet:
When using the 'collection' form type with the 'required' option set to false, The rendered HTML is different between the form_widget() macro and the prototype. The macro does not output the required field contrary to the prototype field.

Form definition :
```
$formBuilder->add('emails', 'collection', array(
            'type' => 'email',
            'allow_add' => true,
            'required' => false
        ));
```
Twig Template : 
```
{% for emailField in form.emails %}
    {{ form_widget(emailField) }}
{% endfor %}
```
HTML output :
```
<input name="form[emails][1]" id="form_emails_1" type="email">
<input name="form[emails][2]" id="form_emails_2" type="email">
<input name="form[emails][3]" id="form_emails_2" type="email">
```
data-prototype:
```
<input type="email" id="form_emails___name__" 
name="form[emails][__name__]" required="required" class="" />
```
Is that, the expected result ? I would expect the prototype field not to contain 'required' as the form_widget() macro doesn't output the 'required' field

I know the 'option' parameter is supposed to do the job but as the 'required' parameter is supported, the prototype output should match the macro output.
```
'options' => array(
       'required' => false
  )
```

stof:
I think this is because there is some special logic for the ``required`` option so that a child of a non-required field is also not required. but the prototype is not a child of the collection. It lives separately because it is special. So we would have to handle things specially for it.
We probably have the same issue for the inheritance of the translation domain btw.

@webmozart do you have an idea to fix it ?

HeahDude:
@stof in which case can "translation_domain" inheritance of prototype become an issue ?

HeahDude:
Inheritance of options for the prototype is not preserved, every option except 'label' is set to default by a call to `array_replace`:
```php
// Symfony/Component/Form/Extension/Core/Type/CollectionType.php

public function buildForm(FormBuilderInterface $builder, array $options)
{
    if ($options['allow_add'] && $options['prototype']) {
        $prototype = $builder->create($options['prototype_name'], $options['type'], array_replace(array(
                 'label' => $options['prototype_name'].'label__',
             ), $options['options']));
             $builder->setAttribute('prototype', $prototype->getForm());
     }
    ...
```

PR #50056, opened 2023-04-18
[Form] Fix "prototype_data" option creating duplicates instead of new lines

### LINKED ISSUE #49961: [Form] CollectionType: prototype_data option creates duplicates instead of new lines

6.2.8

### Description

I'm using CollectionType to embed a collection of forms and when I'm using `prototype_data` option and there is a bug when I add new lines in my form. It creates duplicates instead of create x new lines.

FYI, i'm using DTO's to populate my entities, idk if it changes something but it was working before the package upgrade.
I don't have the issue when I removed the `prototype_data` option but i need it to initialize data in new fields

### How to reproduce

**My parent form** 

```
$builder
    ->add('name', TextType::class)
    ->add('steps', CollectionType::class, [
        'entry_type' => StepForm::class,
        'allow_add' => true,
        'allow_delete' => true,
        'by_reference' => false,
        'prototype' => true,
        'prototype_name' => '__steps__',
        'prototype_data' => new StepData()
    ]);
```

**My child form** 

```
$builder
    ->add('name', TextType::class)
    ->add('probability', ChoiceType::class, [
        'choice_loader' => new CallbackChoiceLoader(function () {
            $probabilities = [];

            for ($i = 0; $i <= 10; $i++) {
                $value = $i * 10;
                $probabilities["{$value}%"] = $value;
            }

            return $probabilities;
        }),
    ])
    ->add('position', HiddenType::class, [
        'attr' => ['class' => 'position']
    ]);
```

**My form in front** _only the last 3 lines are added and new_
Capture d’écran 2023-04-06 à 16 20 10

**After form submission** _you can see that the last 3 lines are the same with the same id (2112)_
Capture d’écran 2023-04-06 à 16 20 31

 

### Possible Solution

I have downgraded the package `symfony/form` to the version `6.2.7` to solve this issue 

### Additional Context

I created a Symfony project to reproduce the bug. You can solve it to downgrade `symfony/form` package
https://github.com/jkgroupe/symfony-bug-form

Thorry84:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #49961 
| License       | MIT

Fix for issue 49961.

Changed the prototype options passed to the ResizeFormListener to prevent prototype_data from being passed to new fields.

### DIFF
--- a/src/Symfony/Component/Form/Extension/Core/Type/CollectionType.php
+++ b/src/Symfony/Component/Form/Extension/Core/Type/CollectionType.php
@@ -23,12 +23,13 @@ class CollectionType extends AbstractType
 {
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
-        $prototypeOptions = null;
+        $resizePrototypeOptions = null;
         if ($options['allow_add'] && $options['prototype']) {
+            $resizePrototypeOptions = array_replace($options['entry_options'], $options['prototype_options']);
             $prototypeOptions = array_replace([
                 'required' => $options['required'],
                 'label' => $options['prototype_name'].'label__',
-            ], array_replace($options['entry_options'], $options['prototype_options']));
+            ], $resizePrototypeOptions);
 
             if (null !== $options['prototype_data']) {
                 $prototypeOptions['data'] = $options['prototype_data'];
@@ -44,7 +45,7 @@ class CollectionType extends AbstractType
             $options['allow_add'],
             $options['allow_delete'],
             $options['delete_empty'],
-            $prototypeOptions
+            $resizePrototypeOptions
         );
 
         $builder->addEventSubscriber($resizeListener);


src/Symfony/Component/Form/Extension/Core/Type/CollectionType.php
```diff
@@ -36,6 +36,8 @@ public function buildForm(FormBuilderInterface $builder, array $options)
 
             $prototype = $builder->create($options['prototype_name'], $options['entry_type'], $prototypeOptions);
             $builder->setAttribute('prototype', $prototype->getForm());
+
+            $resizeFormListenerPrototypeOptions = array_replace($options['entry_options'], $options['prototype_options']);
```
nicolas-grekas: What about the options set on L29-30, do we want them in this array?
Sorry I'm not much into this domain so this might be a dumb question.

Thorry84: Very good question, I was wondering about that as well.

Those provide some default values in case they aren't provided via the `entry_options` and `prototype_options` array. Since required defaults to true, a collection with required set to false would have the entries set to required. (See issue #15544)

However since the `entry_options` and later the `prototype_options` have been introduced, this behavior is kind of redundant.

Same for the label, this is because otherwise the label isn't rendered as a replaceable string when using `prototype_name` as a label. Since the prototype is normally used to add new fields to a form using JS, it's important to have them be replaceable (see issue #3738). For the new fields rendered server-side, the `prototype_name` isn't used, so the label for the prototype shouldn't be used.
I'm not sure of the use case for falling back to the `prototype_name` as a label, but it's being kept the same for BC reasons.

So I think we don't want those values in the array and only use `entry_options` and `prototype_options` for the `ResizeFormListener`.

nicolas-grekas:
Thank you @Thorry84.

PR #49158, opened 2023-01-30
[Form] Deprecate not honoring with_seconds => false for HTML5 datetime inputs in DateTimeType

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | https://github.com/symfony/symfony/issues/48076
| License       | MIT
| Doc PR        | -

The `with_seconds` option defaults to false. It is honored when the form doesn't have data (and when the browser doesn't display the seconds). But it's not honored when the form has data because `DateTimeToHtml5LocalDateTimeTransformer` format the `\DateTimeInterface` to `Y-m-d\\TH:i:s` unconditionally.

In this PR, we add a new argument to `DateTimeToHtml5LocalDateTimeTransformer` to allow formatting to a datetime string without seconds (`Y-m-d\\TH:i`).

Moreover, we deprecate not setting `with_seconds => true` on all forms using HTML5 datetime inputs since they rely on that behavior (and it's the default value of the `with_seconds` option).

In Symfony 7.0, we can honor the option (by forwarding its value to the transformer) because every project should be using `with_seconds => true` before upgrading.

I don't really know if forcing everyone to add/set the option is the best strategy but I did not find another one.

PR #50059, opened 2023-04-19
[Form] Don't render seconds for HTML5 date pickers unless "with_seconds" is explicitly set

### LINKED ISSUE #48076: [Form] DateTimeType remove seconds with HTML5
### Description

When passing a DateTime object to DateTimeType we can not remove the seconds from the HTML5 datepicker.

There is currently no official way to disable seconds in a HTML5 dateTime input, but by not passing the seconds the seconds picker is omitted.

> <input type="datetime-local" value="2020-01-01T11:00" />

However we cannot set the time format in DateTimeToHtml5LocalDateTimeTransformer.
Also no `format` option can be used together with html5.

> ->add('date', null, ['label' => 'date', 'widget' => 'single_text', 'format' => 'Y-m-d H:i'])

results in 

> Cannot use the "format" option of "Symfony\Component\Form\Extension\Core\Type\DateTimeType" when the "html5" option is enabled.

So i would suggest to either allow `format` option or add a dedicated `without_seconds` option for HTML5 and `single_text`. 

### Example

_No response_

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #48076
| License       | MIT
| Doc PR        | -

Alternative to #49158 (commit credits due to @fancyweb)

I don't think we need a deprecation for this change. To me, it's a bugfix.
But merging it on 6.3 might be safer so let's do it as a new feature.


stof (approved):

nicolas-grekas:
Thank you @fancyweb.

fancyweb:
Thank you @nicolas-grekas :smile:

PR #50061, opened 2023-04-19
[DependencyInjection] Fix support for inner collections when using `<bind>`

### LINKED ISSUE #50046: The <bind> tag does not support inner collections

6.2.0

### Description

The inner collection in the following code snippet is ignored:
```xml
<bind key="$foo" type="collection">
    <bind key="inner-collection" type="collection">
       <argument>item-1</argument>
       <argument>item-2</argument>
    </bind>
</bind>
```

### How to reproduce

Bind a variable `$foo` with the following code snippet in a container configuration file, and try to dump this variable in a controller, the inner collection will be empty:

```xml
<bind key="$foo" type="collection">
    <bind key="inner-collection" type="collection">
       <argument>item-1</argument>
       <argument>item-2</argument>
    </bind>
</bind>
```

### Possible Solution

Change `Symfony\Component\DependencyInjection\Loader\XmlFileLoader` as follows:

```diff
                 case 'collection': 
-                    $arguments[$key] = $this->getArgumentsAsPhp($arg, $name, $file);
+                    $argumentName = 'bind' === $arg->parentNode->nodeName ? 'argument' : $name;
+                    $arguments[$key] = $this->getArgumentsAsPhp($arg, $argumentName, $file);
```

### Additional Context

_No response_

zerustech:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50046
| License       | MIT

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Loader/schema/dic/services/services-1.0.xsd
+++ b/src/Symfony/Component/DependencyInjection/Loader/schema/dic/services/services-1.0.xsd
@@ -275,7 +275,7 @@
 
   <xsd:complexType name="bind" mixed="true">
     <xsd:choice maxOccurs="unbounded">
-      <xsd:element name="bind" type="argument" minOccurs="0" maxOccurs="unbounded" />
+      <xsd:element name="bind" type="bind_argument" minOccurs="0" maxOccurs="unbounded" />
       <xsd:element name="service" type="service" />
     </xsd:choice>
     <xsd:attribute name="type" type="argument_type" />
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/DependencyInjection/Loader/schema/dic/services/services-1.0.xsd
```diff
@@ -276,6 +276,7 @@
   <xsd:complexType name="bind" mixed="true">
     <xsd:choice maxOccurs="unbounded">
       <xsd:element name="bind" type="argument" minOccurs="0" maxOccurs="unbounded" />
+      <xsd:element name="argument" type="argument" minOccurs="0" maxOccurs="unbounded" />
```
nicolas-grekas: On closer review, I feel like I mislead you: maybe we can just set the type to "bind" on the line above and remove this one.
Yes, that would forbid XML files that use `<argument>` nested inside `<bind>`. But these were ignored, isn't it? In that case, I would definitely be fine "breaking" those files, proved they are really misleading, thus broken also at the moment.

zerustech: This will cause an XML validation error: since attribute `key` was optional for `argument` elements, but is required for `bind` elements now, any XML file that contains an inner `<bind>` element without a `key` attribute will not be parsed successfully.

nicolas-grekas: Got it thanks! I updated the PR to add a new "bind_argument" complex type.

nicolas-grekas:
Thank you @zerustech.

src/Symfony/Component/DependencyInjection/Loader/schema/dic/services/services-1.0.xsd
zerustech: Thanks, that works perfectly.

zerustech:
> 

No problem.

PR #23098, opened 2017-06-07
Cache ipCheck (2.7)

gonzalovilaseca:
In our app we use trusted proxies. Using Blackfire we found `IpUtils::checkIp` was being called 454 times taking 3.15ms. 
Caching the result saves those 3ms.

| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        |

PR #50064, opened 2023-04-19
[HttpFoundation] Clear IpUtils cache to prevent memory leaks

### LINKED ISSUE #50032: [HttpFoundation] IpUtils::$checkedIps is never cleared

5.4, 6.2

### Description

Foreword: I don't really know how I can make this bug report more concise while still presenting my entire thought process.
Sorry for the long read!

To avoid duplicate work, the IpUtils utility class caches its IP validation results but **never** clears its cache.

Because Symfony is also supporting "infinite" runtimes like swoole or FrankenPHP, the memory usage of this array may become pretty sizeable if either an application checks many IP Adresses per request or the runner is running for a long time.

I might be missing something because I would be kind of surprised if no one encountered this memory usage issue before. So there might be a way already built in by the runtimes that garbage collect static arrays like this.

The best comparison I can think of is the [ArrayAdapter](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Cache/Adapter/ArrayAdapter.php) from the cache component which would have the same problem if it wouldnt implement the [ResettableInterface](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Cache/ResettableInterface.php) which inherits from [ResetInterface](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Contracts/Service/ResetInterface.php). 

### How to reproduce

I had many issues setting up runtimes like swoole or FrankenPHP back when I tried it, so I kind of try to avoid it right now,
but I am pretty sure that the fine people contributing to this project can get the gist of the potential issue I described above. 
Otherwise Im sorry that I couldnt express my thoughts more clearly.

### Possible Solution

Now, we could register the IpUtils utility as a service in the FrameworkBundle and tag it with "kernel.reset", but I think it would be a huge overkill because all methods of IpUtils are currently static and therefor the extra cost of adding it to the container for literally **every single Symfony application in the world** is just not worth it in my opinion.

Instead I would propose the following two (very similar) options:

1. Add something along the lines of `IpUtils::clearCache` and call it in Kernel::boot (before L113 or after L115)
https://github.com/symfony/symfony/blob/f84279720ae4de0942f8d1b41d6161ee321d1ade/src/Symfony/Component/HttpKernel/Kernel.php#L109-L116
2. Add something along the lines of `IpUtils::clearCache` and call it in the [ServicesResetter](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HttpKernel/DependencyInjection/ServicesResetter.php) directly. This wouldnt be ideal because the service may be overwritten or not registered at all. Both is really rare (I think) but worth the consideration.

Everyone who uses a runtime such as swoole (notably not FrankenPHP) is basically forced to use a reverse proxy and [Request::isFromTrustedProxy](https://github.com/symfony/symfony/blob/f84279720ae4de0942f8d1b41d6161ee321d1ade/src/Symfony/Component/HttpFoundation/Request.php#L1964-L1967) is therefor called on every request (even if they use REMOTE_ADDR as trusted proxy value). 

Somehow not invalidating the cache on every request could be quite benefitial because of this. 
We could implement something like a probabilistic cache invalidation or just clear the cache when a fixed size is reached (we know that only boolean values with string keys are cached). 

Registering IpUtils as a service would definitly be the more "Symfony" way of handling this issue but I think the `Kernel::boot` hack is worth it and **NOT** totally unprecedented since [Kernel::preBoot](https://github.com/symfony/symfony/blob/f84279720ae4de0942f8d1b41d6161ee321d1ade/src/Symfony/Component/HttpKernel/Kernel.php#L745-L770) actually sets the trusted proxies / hosts.

### Additional Context

_No response_

danielburger1337:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50032 
| License       | MIT
| Doc PR        | 

Fixes #50032 

IpUtils cache now uses LRU principal

### DIFF
--- a/src/Symfony/Component/HttpFoundation/IpUtils.php
+++ b/src/Symfony/Component/HttpFoundation/IpUtils.php
@@ -75,23 +75,23 @@ class IpUtils
     public static function checkIp4(string $requestIp, string $ip): bool
     {
         $cacheKey = $requestIp.'-'.$ip.'-v4';
-        if (isset(self::$checkedIps[$cacheKey])) {
-            return self::$checkedIps[$cacheKey];
+        if (null !== $cacheValue = self::getCacheResult($cacheKey)) {
+            return $cacheValue;
         }
 
         if (!filter_var($requestIp, \FILTER_VALIDATE_IP, \FILTER_FLAG_IPV4)) {
-            return self::$checkedIps[$cacheKey] = false;
+            return self::setCacheResult($cacheKey, false);
         }
 
         if (str_contains($ip, '/')) {
             [$address, $netmask] = explode('/', $ip, 2);
 
             if ('0' === $netmask) {
-                return self::$checkedIps[$cacheKey] = false !== filter_var($address, \FILTER_VALIDATE_IP, \FILTER_FLAG_IPV4);
+                return self::setCacheResult($cacheKey, false !== filter_var($address, \FILTER_VALIDATE_IP, \FILTER_FLAG_IPV4));
             }
 
             if ($netmask < 0 || $netmask > 32) {
-                return self::$checkedIps[$cacheKey] = false;
+                return self::setCacheResult($cacheKey, false);
             }
         } else {
             $address = $ip;
@@ -99,10 +99,10 @@ class IpUtils
         }
 
         if (false === ip2long($address)) {
-            return self::$checkedIps[$cacheKey] = false;
+            return self::setCacheResult($cacheKey, false);
         }
 
-        return self::$checkedIps[$cacheKey] = 0 === substr_compare(sprintf('%032b', ip2long($requestIp)), sprintf('%032b', ip2long($address)), 0, $netmask);
+        return self::setCacheResult($cacheKey, 0 === substr_compare(sprintf('%032b', ip2long($requestIp)), sprintf('%032b', ip2long($address)), 0, $netmask));
     }
 
     /**
@@ -120,8 +120,8 @@ class IpUtils
     public static function checkIp6(string $requestIp, string $ip): bool
     {
         $cacheKey = $requestIp.'-'.$ip.'-v6';
-        if (isset(self::$checkedIps[$cacheKey])) {
-            return self::$checkedIps[$cacheKey];
+        if (null !== $cacheValue = self::getCacheResult($cacheKey)) {
+            return $cacheValue;
         }
 
         if (!((\extension_loaded('sockets') && \defined('AF_INET6')) || @inet_pton('::1'))) {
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/HttpFoundation/IpUtils.php
```diff
@@ -120,8 +120,8 @@ public static function checkIp4(string $requestIp, string $ip): bool
     public static function checkIp6(string $requestIp, string $ip): bool
     {
         $cacheKey = $requestIp.'-'.$ip.'-v6';
-        if (isset(self::$checkedIps[$cacheKey])) {
-            return self::$checkedIps[$cacheKey];
+        if (null !== ($cacheValue = self::getCacheResult($cacheKey))) {
```
nicolas-grekas: extra brackets should be removed
but more importantly,this line need to by duplicated on L78 ALSO

nicolas-grekas:
Can you run some bench comparing with/without LRU, also maybe without cache at all?

danielburger1337:
> Can you run some bench comparing with/without LRU, also maybe without cache at all?

Just a a quick reminder why the cache was implemented: IPv6 checking is **extremely** slow (in comparison to IPv4).

PHP 8.2.4

Average after 3 runs each and 1 million IP addresses / cache entries: 

- With LRU: 3.778 seconds, 2.097152 mb
- Max Items Cache: 3.615 seconds, 2.097152 mb
- Current Cache: 2.833 seconds, 100.663296 mb 

Other Tests:

- Keeping count of the items stored and not using `count`: 3.593 seconds, 2.097152 mb
- Keeping count of the items stored and not using `count` (LRU): 3.653 seconds, 2.097152 mb

- Without any Cache (IPv4): 2.484 seconds, 2.097152 mb
- Without any Cache (IPv6): 5.741 seconds, 2.097152 mb

An option to improve a tiny bit would be to only cache IPv6 results since IPv4 checking seems to be a little bit faster than **ANY** caching.

---
```php
<?php

require './vendor/autoload.php';

use Symfony\Component\HttpFoundation\IpUtils;

$start = microtime(true);

for ($i = 0; $i < 1_000_000; ++$i) {
    IpUtils::checkIp4(random_int(1, 254).random_int(0, 254).random_int(0, 254).random_int(0, 255), '192.168.1.1');
    // IpUtils::checkIp6('::' . $i, 'fe80::/65');
}

$end = microtime(true);

var_dump($end - $start, get_peak_memory_usage());
```
---

When only checking about 100 IP addresses the difference becomes basically indistinguishable:

- With LRU: 0.00061893463134766 seconds, 2.097152 mb
- Max Items Cache: 0.00063395500183105 seconds, 2.097152 mb
- Current Cache: 0.00051212310791016 seconds, 2.097152 mb
- Without any Cache (IPv4): 0.00046300888061523 seconds, 2.097152 mb
- Without any Cache (IPv6): 0.00075387954711914 seconds, 2.097152 mb

nicolas-grekas:
> An option to improve a tiny bit would be to only cache IPv6 results since IPv4 checking seems to be a little bit faster than ANY caching.

let's do this yes! (with an inline comment reminding us about this)
let's drop the -v6 suffix while doing so

nicolas-grekas:
> With LRU: 3.778 seconds, 2.097152 mb
> Without any Cache (IPv6): 5.741 seconds, 2.097152 mb

The difference is not that big actually.

nicolas-grekas:
The cache has been introduced in https://github.com/symfony/symfony/pull/23098
What about dropping it altogether? The diff doesn't look much worth the added complexity to me.

stof:
the benchmark being done seems to be testing the performance in case of cache misses though.

the cache was added to improve performance for the common case of checking the trusted ips many times, which will do cache hits for most checks.

nicolas-grekas (approved):
@stof is right, let's keep the cache :)

nicolas-grekas:
Thank you @danielburger1337.

PR #50066, opened 2023-04-20
[Dumper] Trim leading newlines when checking if value begins with a space

### LINKED ISSUE #50065: [Yaml] [Dumper] Dumper does not use block indentation indicator when multiline value begins with newline followed by spaces.

\>=3.4.5

### Description

`\Symfony\Component\Yaml\Dumper::dump` only checks for leading spaces when deciding whether to add a block indentation indicator. If the value begins with a newline followed by spaces then the indicator isn't added.

https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Yaml/Dumper.php#L76

### How to reproduce

```php
// Create dumper with 2-space indentation.
$dumper = new \Symfony\Component\Yaml\Dumper(2);

// Prepare value beginning with newline followed by four spaces.
$data = ['key' => "\n    multiline value\nwith leading newline\nfollowed by spaces"];

// Dump the value to YAML.
$dumper->dump($data2, 1, 0, \Symfony\Component\Yaml\Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK);
```

Result:
```yaml
key: |-

      multiline value
  with leading newline
  followed by spaces
```

Expected result:
```yaml
key: |2-

      multiline value
  with leading newline
  followed by spaces
```

### Possible Solution

Remove leading newlines when determining the value of ` $blockIndentationIndicator`.

### Additional Context

_No response_

bradtreloar:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50065
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Yaml/Dumper.php
+++ b/src/Symfony/Component/Yaml/Dumper.php
@@ -69,9 +69,7 @@ class Dumper
                 }
 
                 if (Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK & $flags && \is_string($value) && false !== strpos($value, "\n") && false === strpos($value, "\r")) {
-                    // If the first line starts with a space character, the spec requires a blockIndicationIndicator
-                    // http://www.yaml.org/spec/1.2/spec.html#id2793979
-                    $blockIndentationIndicator = (' ' === substr($value, 0, 1)) ? (string) $this->indentation : '';
+                    $blockIndentationIndicator = $this->getBlockIndentationIndicator($value);
 
                     if (isset($value[-2]) && "\n" === $value[-2] && "\n" === $value[-1]) {
                         $blockChompingIndicator = '+';
@@ -98,9 +96,7 @@ class Dumper
                     $output .= sprintf('%s%s !%s', $prefix, $dumpAsMap ? Inline::dump($key, $flags).':' : '-', $value->getTag());
 
                     if (Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK & $flags && \is_string($value->getValue()) && false !== strpos($value->getValue(), "\n") && false === strpos($value->getValue(), "\r\n")) {
-                        // If the first line starts with a space character, the spec requires a blockIndicationIndicator
-                        // http://www.yaml.org/spec/1.2/spec.html#id2793979
-                        $blockIndentationIndicator = (' ' === substr($value->getValue(), 0, 1)) ? (string) $this->indentation : '';
+                        $blockIndentationIndicator = $this->getBlockIndentationIndicator($value->getValue());
                         $output .= sprintf(' |%s', $blockIndentationIndicator);
 
                         foreach (explode("\n", $value->getValue()) as $row) {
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/Yaml/Dumper.php
```diff
@@ -69,9 +69,9 @@ public function dump($input, int $inline = 0, int $indent = 0, int $flags = 0):
                 }
 
                 if (Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK & $flags && \is_string($value) && false !== strpos($value, "\n") && false === strpos($value, "\r")) {
-                    // If the first line starts with a space character, the spec requires a blockIndicationIndicator
+                    // If the first non-empty line starts with a space character, the spec requires a blockIndicationIndicator
```
xabbuh: Is that true? Or what happens when we change the fixture from your test to something like this?

```php
$data = [
    'data' => [
        'multi_line' => "\n    \n    the first line has leading newline and spaces\nThe second line does not.",
    ],
];
```

src/Symfony/Component/Yaml/Dumper.php
```diff
@@ -163,4 +159,16 @@ private function dumpTaggedValue(TaggedValue $value, int $inline, int $indent, i
...
     }
+
+    private function getBlockIndentationIndicator(string $value) {
+        $lines = explode("\n", $value);
+
+        // If the first non-empty line starts with a space character, the spec requires a blockIndicationIndicator
+        // http://www.yaml.org/spec/1.2/spec.html#id2793979
+        foreach ($lines as $line) {
+            if (trim($line, ' ') !== '') {
+                return (' ' === substr(ltrim($line, "\n"), 0, 1)) ? (string) $this->indentation : '';
```
xabbuh: the `ltrim()` is now not needed anymore

src/Symfony/Component/Yaml/Dumper.php
bradtreloar: The parser treats any line containing only spaces as an empty line, even if the number of space exceeds the indentation.

So it isn't enough to trim off the first newline before looking for a leading space, the dumper needs to find the first line that is neither empty nor contains only spaces.

I've added an additional test to check that the indicator is not added if the value begins with a line containing _only_ spaces.

The indentation indicator check is now more than a single-line operation, so I've moved it into a private function.

xabbuh:
The fix itself looks good to me now. Can you take a look at the things fabbot reported though please?

src/Symfony/Component/Yaml/Dumper.php
```diff
@@ -163,4 +157,20 @@ private function dumpTaggedValue(TaggedValue $value, int $inline, int $indent, i
 
         return $output."\n".$this->dump($value->getValue(), $inline - 1, $indent, $flags);
     }
+
+    private function getBlockIndentationIndicator(string $value): string
+    {
+        $lines = explode("\n", $value);
+
+        // If the first line (that is neither empty nor contains only spaces)
+        // starts with a space character, the spec requires a blockIndicationIndicator
```
xabbuh:
```suggestion
        // starts with a space character, the spec requires a block indentation indicator
```

xabbuh (approved):

xabbuh:
👍 just one last minor comment

fabpot:
Thank you @bradtreloar.

PR #50075, opened 2023-04-20
[Messenger] Add EventListener to handle delayed Messages on `kernel.terminate` event

Koc:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | TBD

This PR adds possibility to register special "delayed" in-memory transport to delay messages handling on [`kernel.terminate`](https://symfony.com/doc/current/components/http_kernel.html#8-the-kernel-terminate-event) event. It can be useful for non-critical but time-consuming operations like sending some analytics, metrics, various experiments like Parallel running, etc.

## :wrench: Usage:
```yaml
# config/messenger.yaml
framework:
    messenger:
        delayed_transport: delayed
        transports:
            delayed: 'in-memory://'
        routing:
            'App\Presentation\Message\TrackSearchAnalyticsMessage': delayed
```

```php
final class SearchProductsController
{
    public function __construct(MessageBusInterface $commandBus)
    {
    }

    #[Route('/api/public/v1/products', name: 'search_products', methods: 'GET')]
    public function __invoke(#[MapQueryParameter] string $query): ProductsCollectionResponse
    {
        // do search ...

        // message will be handled on "kernel.terminate" instead of immediate handling
        $this->commandBus->dispatch(new TrackSearchAnalyticsMessage($query));

        return $response;
    }
}
```

## :spiral_notepad: Todo:
- [x] Changelog
- [ ] tests


src/Symfony/Component/Messenger/EventListener/HandleDelayedMessagesOnKernelTerminateListener.php
```diff
@@ -0,0 +1,53 @@
...
+            $this->eventDispatcher,
+            $this->logger,
+            [$this->receiverName => $this->rateLimiterFactory],
+        );
+        $worker->run(['sleep' => 1]);
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [TerminateEvent::class => 'onKernelTerminate'];
```
94noni: following your logic of delaying dispatching, we can imagine that project already have listener/subscriber to such event, shouldnt the priority of this one be defined as -256 to be at the very end of the event propagation?

magikid:
If you have a time consuming task, why are you running the handler during the request instead of in a separate process?

Koc:
@magikid 

>  It can be useful for non-critical but time-consuming operations like sending some analytics, metrics, various experiments like Parallel running, etc.

Adding separate consumers for some simple tasks can increase complexity of the infrastructure.

ro0NL:
IIUC this makes `in-memory://` an implicit sync transport?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1519,6 +1519,10 @@ function ($a) {
                             ->defaultNull()
                             ->info('Transport name to send failed messages to (after all retries have failed).')
                         ->end()
+                        ->scalarNode('delayed_transport')
```
artyuum, line 1522: I find the naming confusing. To me, it sounds like my message could be delayed no matter the transport used ([like this](https://symfony.com/doc/current/messenger.html#envelopes-stamps)) while this option actually brings a different strategy of handling the message: using the `kernel.terminate` event.

Maybe `handle_after_request` would be more suitable?

valtzu:
How about using stamp+middleware, similar to `DispatchAfterCurrentBusStamp`? Instead of running a Worker on `kernel.terminate`, we would just finish the middleware stack execution 🤔  This would support any transport.

Also this would make it easy to ignore the "delaying" behavior when running on CLI.

chalasr:
Closing as there is no more activity here. Please feel free to reopen

Issue #50054, opened 2023-04-18
[VarDumper] Tell us more about "uninitialized" properties


lyrixx:
### Description

TIL `var_dump()` displays more information than the VarDumper component about uninitialized properties.

```
object(Subscription)#2 (1) {
  ["id"]=>
  string(3) "123"
  ["canceled"]=>
  uninitialized(bool)
}
```

[see it on 3v4l.org](https://3v4l.org/uNhBV#v8.2.5)

But the var dumper don't:

```
^ Subscription^ {#22
  +id: "123"
}
```

It would be useful to add this information in the dump, isn't?

### ~~Example~~ Proposal

```
^ Subscription^ {#22
  +id: "123"
  +canceled: ⚠ uninitialized(bool)
}

PR #50076, opened 2023-04-20
[VarDumper] show uninitialized properties

guillaumeVDP:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no
| Tickets       | Fix #https://github.com/symfony/symfony/issues/50054
| License       | MIT
| Doc PR        | -

First commit for the feature. Please review this pull request and help me doing this in a good way :pray: 

I am not sure if it's good to store uninitialized properties to the attr of the stub. Any better ideas ?

Missing:
- tests
- updating CHANGELOG.md

### DIFF
--- a/src/Symfony/Component/VarDumper/Cloner/Data.php
+++ b/src/Symfony/Component/VarDumper/Cloner/Data.php
@@ -361,7 +361,7 @@ class Data implements \ArrayAccess, \Countable, \IteratorAggregate
                             $withChildren = false;
                             $cut = -1;
                         } else {
-                            $cut = $this->dumpChildren($dumper, $cursor, $refs, $children, $cut, $item->type, null !== $item->class);
+                            $cut = $this->dumpChildren($dumper, $cursor, $refs, $children, $cut, $item->type, $item->attr, null !== $item->class);
                         }
                     } elseif ($children && 0 <= $cut) {
                         $cut += \count($children);
@@ -392,7 +392,7 @@ class Data implements \ArrayAccess, \Countable, \IteratorAggregate
      *
      * @return int The final number of removed items
      */
-    private function dumpChildren(DumperInterface $dumper, Cursor $parentCursor, array &$refs, array $children, int $hashCut, int $hashType, bool $dumpKeys): int
+    private function dumpChildren(DumperInterface $dumper, Cursor $parentCursor, array &$refs, array $children, int $hashCut, int $hashType, array $hashAttributes, bool $dumpKeys): int
     {
         $cursor = clone $parentCursor;
         ++$cursor->depth;
@@ -409,6 +409,13 @@ class Data implements \ArrayAccess, \Countable, \IteratorAggregate
 
                 return $hashCut >= 0 ? $hashCut + $cursor->hashLength - $cursor->hashIndex : $hashCut;
             }
+        }
+
+        // Add uninitialized properties
+        foreach ($hashAttributes["uninitialized"] as $propName => $propType) {
+            $string = "⚠ uninitialized($propType)";
+            $cursor->hashKey = $dumpKeys ? $propName : null;
+            $this->dumpItem($dumper, $cursor, $refs, $string);
         }
 
         return $hashCut;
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/VarDumper/Cloner/VarCloner.php
```diff
@@ -128,6 +128,15 @@ protected function doClone(mixed $var): array
                             $stub->value = $v;
                             $stub->handle = $h;
                             $a = $this->castObject($stub, 0 < $i);
+                            if ($stub->value !== '') {
+                                $uninitializedProps = array_diff_key(get_class_vars($stub->value::class), $a);
+                                $stubClass = $stub->value::class;
+                                $stub->attr['uninitialized'] = [];
+                                foreach ($uninitializedProps as $key => $value) {
+                                    $rp = new \ReflectionProperty($stubClass, $key);
+                                    $stub->attr['uninitialized'][] = [$key => $rp->getType()->getName()];
```
stof: this is broken for any union or intersection types. Not all ReflectionType are ReflectionNamedType.

stof: why using a list of single-item arrays with an unknown key for this `$stub->attr['uninitialized']` ? why not using `$stub->attr['uninitialized'][$key] = $type` here ?

src/Symfony/Component/VarDumper/Cloner/VarCloner.php
```diff
@@ -128,6 +128,15 @@ protected function doClone(mixed $var): array
                             $stub->value = $v;
                             $stub->handle = $h;
                             $a = $this->castObject($stub, 0 < $i);
+                            if ($stub->value !== '') {
+                                $uninitializedProps = array_diff_key(get_class_vars($stub->value::class), $a);
+                                $stubClass = $stub->value::class;
+                                $stub->attr['uninitialized'] = [];
+                                foreach ($uninitializedProps as $key => $value) {
+                                    $rp = new \ReflectionProperty($stubClass, $key);
```
lyrixx, line 136: Can you add a local cache for this?

lyrixx (commented):
wahoo, super good start!

src/Symfony/Component/VarDumper/Cloner/Data.php
```diff
@@ -367,6 +367,14 @@ private function dumpItem(DumperInterface $dumper, Cursor $cursor, array &$refs,
                         $cut += \count($children);
                     }
                     $cursor->skipChildren = false;
+                    // Add uninitialized properties
+                    foreach ($item->attr["uninitialized"] as $uninitializedProperty) {
```
stof: shouldn't this be done inside `dumpChildren` instead ? Otherwise, things will be weird when the cursor is meant to dump without children.

src/Symfony/Component/VarDumper/Cloner/VarCloner.php
```diff
@@ -128,6 +128,15 @@ protected function doClone(mixed $var): array
                             $stub->value = $v;
                             $stub->handle = $h;
                             $a = $this->castObject($stub, 0 < $i);
+                            if ($stub->value !== '') {
```
stof, line 131: Does this work for nested objects or should we implement that logic in `castObject` ?

And I'm wondering whether this should be implemented in `Caster::castObject` by adding those properties as a special `UninitializedStub`, so that this is done _before_ any custom caster runs. This way, if a caster decides to cut a property, this would also cut it if it is uninitialized.

nicolas-grekas: I agree with this. I would add this to AbstractCloner::castObject, where there is already `$this->classInfo` that we can use to store those `UninitializedStub`.

guillaumeVDP: Ok I'll try implementing this solution. Thanks

nicolas-grekas:
I went with another approach in #51130
Thanks for pushing this forward!

PR #50078, opened 2023-04-20
Add "composer require..." in all exception messages when needed

### LINKED ISSUE #50077: [DX] Add "composer require..." in all exception messages when needed

6.3

### Description

I noticed that exception messages lacked normalization, sometimes they have a good DX and allow the dev. to simply copy/paste a part of the message to quickly install the required package, and sometimes they just point out a missing package and nothing more.

This is my attempt to improve that.

### How to reproduce

Look into the code for all occurrences of `LogicException` and see by yourself. But here is an example:
https://github.com/symfony/routing/blob/69062e2823f03b82265d73a966999660f0e1e404/Matcher/UrlMatcher.php#L256

### Possible Solution

_No response_

### Additional Context

_No response_

artyuum:
| Q             | A
| ------------- | ---
| Branch?       | 6.3 
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #50077 
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Symfony/Bridge/Doctrine/composer.json
+++ b/src/Symfony/Bridge/Doctrine/composer.json
@@ -68,14 +68,6 @@
         "symfony/security-core": "<6.0",
         "symfony/validator": "<5.4"
     },
-    "suggest": {
-        "symfony/form": "",
-        "symfony/validator": "",
-        "symfony/property-info": "",
-        "doctrine/data-fixtures": "",
-        "doctrine/dbal": "",
-        "doctrine/orm": ""
-    },
     "autoload": {
         "psr-4": { "Symfony\\Bridge\\Doctrine\\": "" },
         "exclude-from-classmap": [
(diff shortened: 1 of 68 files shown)


fabpot:
I like it. Maybe it's time to remove all `suggest` sections in our `composer.json` files. @symfony/mergers

artyuum:
@fabpot should it be done in this PR or in a separate?

fabpot:
> @fabpot should it be done in this PR or in a separate?

I think it makes sense to do it here as well.

artyuum:
@fabpot done.

fabpot (approved):

nicolas-grekas (approved):

lyrixx (approved):

fabpot:
Thank you @artyuum.

chalasr:
What about adding some domain exception to reduce boilerplate and future inconcistencies on such errors?

artyuum:
@chalasr I like the idea. While changing these lines of code by hand, I was actually wondering the same. To make sure I understood your suggestion, are you suggesting something like this?

```php
class MissingRequiredPackageException extends DomainException {
    public function __construct(string $package)
    {
        parent::__construct(sprintf('The package "%s" is required. Try running "composer require %s".', $package, $package));
    }
}

...

throw new MissingRequiredPackageException('symfony/security-core');
```

nicolas-grekas:
Most messages also tell which feature cannot be used because of the missing dep.
I don't think domain exceptions would help as they are for humans, not for any app-related logic.

Issue #49981, opened 2023-04-08
[Routing] Allow using invokable controller's FQCN (i.e. `::class`) to generate URLs (instead of route name)


ThomasLandauer:
### Description

With the advent of invokable controllers, it would be possible to move away from using names (i.e. strings) as route idenfifiers, and use the controller's FQCN instead:
```php
#[Route(path: '/some')]
final class SomeController
{
    public function __invoke(): Response
    {
        // ...
    }
}
```
Then in any service:
```php
$this->router->generate(SomeController::class);
```
Twig:
```twig
{{ path('App\\Controller\\SomeController') }}
```

Advantage: No more strings! :-)

I got the idea from https://tomasvotruba.com/blog/to-route-or-to-action-thats-the-question which says that this is  possible in Laravel:
> It allows using the controller directly instead of a middleman string:
> +`<form action="{{ action(\App\Controllers\ProcessPhpFormController::class) }}" method="post">`

derrabus:
I like the idea. However, we currently generate a name for routes already, if the name property is omitted. Silently changing the strategy how that name is generated would be a BC break.

ro0NL:
https://symfony.com/blog/new-in-symfony-5-4-route-aliasing#deprecating-routes :eyes:

nicolas-grekas:
Not sure I would deprecate the currently generated name, but why not create an alias indeed.
This would work only for invokable controllers.

Issue #49369, opened 2023-02-14
[RFC] Deprecate most of or all annotations?


alexislefebvre:
### Description

There is ongoing work about deprecating the usage of annotations:

- #49358
- #48990 

I didn't see a general discussion about this subject, so I created this issue to raise the question:

Should some or all annotations be deprecated in 6.3 and removed in 7.0?

Attributes are natively supported with Symfony 6.3+, because it requires PHP ≥ 8.1 and attributes were added in PHP 8.0. While annotations require to install `doctrine/annotations`.

On a side note, PHPUnit 10 allow to use attributes instead of annotations: https://github.com/sebastianbergmann/phpunit/issues/4502

### Example

```diff
-    /**
-     * @required
-     */
+    #[\Symfony\Contracts\Service\Attribute\Required]
     public function setFoo(Foo $foo)
     {
         // should be called
     }
```

derrabus:
Yes. Let's kill Doctrine Annotations.

stof:
Knowing that the Rector implementation of migrating annotations to attributes is currently broken (because their parsing of Doctrine annotations is broken, ignoring the fact that string literals are a thing in annotations and quotes are not just a parenthesis-like delimiter), there is no easy *working* migration path for now.

yguedidi:
@stof there is a work in progress to fix that: https://github.com/rectorphp/rector-src/pull/3645

derrabus:
Fixed by:

* #50888
* #49358
* #50982
* #50983
* #51050
* #51082
* #51085
* #51080

alexisLefebvre:
@derrabus Thanks for the update! Is there still some annotations that are not (yet) deprecated?

derrabus:
No, it's done.

PR #50084, opened 2023-04-20
[Routing] Add FQCN and FQCN::method aliases when applicable

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | https://github.com/symfony/symfony/issues/49981
| License       | MIT
| Doc PR        | -

See https://github.com/symfony/symfony/issues/49981, I think it's a great idea :smiley:

* We add an FQCN alias only if the target class has an `__invoke` method that adds a route AND if the target class added 1 route exactly.
* We add a FQCN::method alias for every method that defines only one route.

### DIFF
--- a/src/Symfony/Component/Routing/CHANGELOG.md
+++ b/src/Symfony/Component/Routing/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.3
+---
+
+ * Add FQCN and FQCN::method aliases for routes loaded from attributes/annotations when applicable
+
 6.2
 ---
 
--- a/src/Symfony/Component/Routing/Loader/AnnotationClassLoader.php
+++ b/src/Symfony/Component/Routing/Loader/AnnotationClassLoader.php
@@ -125,10 +125,20 @@ abstract class AnnotationClassLoader implements LoaderInterface
             return $collection;
         }
 
+        $fqcnAlias = false;
         foreach ($class->getMethods() as $method) {
             $this->defaultRouteIndex = 0;
+            $routeNamesBefore = array_keys($collection->all());
             foreach ($this->getAnnotations($method) as $annot) {
                 $this->addRoute($collection, $annot, $globals, $class, $method);
+                if ('__invoke' === $method->name) {
+                    $fqcnAlias = true;
+                }
+            }
+
+            if (1 === $collection->count() - \count($routeNamesBefore)) {
+                $newRouteName = current(array_diff(array_keys($collection->all()), $routeNamesBefore));
+                $collection->addAlias(sprintf('%s::%s', $class->name, $method->name), $newRouteName);
             }
         }
 
@@ -136,7 +146,13 @@ abstract class AnnotationClassLoader implements LoaderInterface
             $globals = $this->resetGlobals();
             foreach ($this->getAnnotations($class) as $annot) {
                 $this->addRoute($collection, $annot, $globals, $class, $class->getMethod('__invoke'));
+                $fqcnAlias = true;
             }
+        }
+
+        if ($fqcnAlias && 1 === $collection->count()) {
+            $collection->addAlias($class->name, $invokeRouteName = key($collection->all()));
+            $collection->addAlias(sprintf('%s::__invoke', $class->name), $invokeRouteName);
         }
 
         return $collection;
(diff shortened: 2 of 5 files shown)


src/Symfony/Component/Routing/Tests/Fixtures/AnnotationFixtures/InvokableMethodController.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Symfony\Component\Routing\Tests\Fixtures\AnnotationFixtures;
+
+use Symfony\Component\Routing\Annotation\Route;
+
+class InvokableMethodController
+{
+    /**
+     * @Route("/here", name="lol", methods={"GET", "POST"}, schemes={"https"})
```
alexisLefebvre, line 10: Suggestion: remove this annotation and provide only an attribute. See:

- #49369

src/Symfony/Component/Routing/Tests/Fixtures/AttributeFixtures/InvokableMethodController.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Symfony\Component\Routing\Tests\Fixtures\AttributeFixtures;
+
+use Symfony\Component\Routing\Annotation\Route;
+
+class InvokableMethodController
+{
+    #[Route(path: '/here', name: 'lol', methods: ["GET", "POST"], schemes: ['https'])]
+
```
nicolas-grekas: blank line should be removed

src/Symfony/Component/Routing/Tests/Fixtures/AnnotationFixtures/InvokableMethodController.php
nicolas-grekas: I would keep this test case, no need to degrade the DX, even if we plan to deprecate annotations.

nicolas-grekas (commented):
Nice :)

What about deprecating the name generated by getDefaultRouteName (when it exists) and always adding an alias named `FQCN::method` (or just FQCN when method is `__invoke`)

src/Symfony/Component/Routing/Tests/Fixtures/AnnotationFixtures/InvokableMethodController.php
stof: anyway, we cannot remove this annotation just for this new case, as this is the _existing_ annotation.

alexisLefebvre: Yes I misread this as a new annotation, sorry.

src/Symfony/Component/Routing/Loader/AnnotationClassLoader.php
```diff
@@ -125,20 +125,36 @@ public function load(mixed $class, string $type = null): RouteCollection
...
             $globals = $this->resetGlobals();
             foreach ($this->getAnnotations($class) as $annot) {
                 $this->addRoute($collection, $annot, $globals, $class, $class->getMethod('__invoke'));
+                $fqcnAlias = true;
             }
         }
 
+        if ($fqcnAlias && 1 === $collection->count()) {
+            $collection->addAlias($class->name, $invokeRouteName = key($collection->all()));
+            $collection->addAlias(sprintf('%s::__invoke', $class->name), $invokeRouteName);
```
fancyweb, line 155: That means we add an "FQCN::__invoke" alias for controllers like this:
```php
#[Route(path: '/foo')]
final class FooController
{
    public function __invoke():
    {
    }
}
````

**Review by nicolas-grekas — APPROVED** [review]

**Comment by stof** [comment]
This does not seem to deprecate the default name though AFAICT

**Comment by fancyweb** [comment]
@stof Indeed, I don't think it's worth it to deprecate the default name and Nicolas agreed after discussing it with him on Slack.

File: src/Symfony/Component/Routing/CHANGELOG.md [hunk 4]
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.3
```
nicolas-grekas, line 4: 6.4 now

nicolas-grekas (approved):
Please update the PR description for 6.4 and give some example use cases

fabpot (approved):

fabpot:
Thank you @fancyweb.

PR #36586, opened 2020-04-26
[DI] allow loading and dumping tags with an attribute named "name"

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

This is a minor feature added for consistency: using PHP, we can already define tags with an attribute named `"name"`. But then, we cannot dump such definitions in YAML nor XML since we don't have a syntax to declare such tags in these formats.

I spotted this while looking at a dumped container: we already use an attribute named `"name"` on two tags: `cache.pool` and `workflow.definition`. Currently, the dumped XML is wrong because of this.

This PR enables the following new syntaxes (the current style still works as usual):
- in YAML, consistently with the new syntax for method calls:
  ```yaml
  tags:
    - cache.pool: { name: my_cache_pool }
  ```
- in XML:
  ```xml
  <tag name="my_cache_pool">cache.pool</tag>
  ```

PR #50088, opened 2023-04-21
[DependencyInjection] Do not ignore tags `name` attribute when it does not define their name

### LINKED ISSUE #50081: [6.3] non-existent resolver Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver

6.3.x-dev

### Description

I had a fully working 6.2 app and I upgraded to 6.3.x-dev as a test.

My controller has some new (and working in 6.2) `MapEntity` attributes like

```php
 #[MapEntity(mapping: [
            'location_id' => 'id',
        ])] Locations $location,
```

Since `[HttpKernel] Introduce pinnable value resolvers with #[ValueResolver] and #[AsPinnedValueResolver]` #48992 was merged I now get this error 

ScreenShot-2023-04-20-18 06 50

Sorry way beyond me... 

Ping @MatTheCat 

### How to reproduce

Here is a reproducer - https://github.com/PhilETaylor/mapentity_reproducer

To create this reproducer I did 

- composer create-project symfony/skeleton bug_app
- change composer.json to allow 6.3.x and dev stability
- composer require orm 
- composer require symfony/maker-bundle --dev
- use maker bundle to create a controller - edit the controller to add the MapEntity
- use maker bundle to create an entity
- symfony serve
- go to http://127.0.0.1:8000/bug - you can see the error there

### Possible Solution

Pay @MatTheCat some sponsorship... incoming.... Edit: No sponsor button :-( 

### Additional Context

I can "fix" this for my own purposes with this in services.php so maybe its a tagging issue?

```php
$services->set(\Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver::class)
        ->tag('doctrine.orm.entity_value_resolver');
```

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50081
| License       | MIT
| Doc PR        | N/A

Tags `name` attribute is ignored using XML if the tag name is its node content. That means `<tag name="name_attribute">tag_name</tag>` will return a `tag_name` tag without any attribute.

This seems to be a regression from #36586.

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Loader/XmlFileLoader.php
+++ b/src/Symfony/Component/DependencyInjection/Loader/XmlFileLoader.php
@@ -336,13 +336,14 @@ class XmlFileLoader extends FileLoader
         $tags = $this->getChildren($service, 'tag');
 
         foreach ($tags as $tag) {
-            if ('' === $tagName = $tag->childElementCount || '' === $tag->nodeValue ? $tag->getAttribute('name') : $tag->nodeValue) {
+            $tagNameComesFromAttribute = $tag->childElementCount || '' === $tag->nodeValue;
+            if ('' === $tagName = $tagNameComesFromAttribute ? $tag->getAttribute('name') : $tag->nodeValue) {
                 throw new InvalidArgumentException(sprintf('The tag name for service "%s" in "%s" must be a non-empty string.', (string) $service->getAttribute('id'), $file));
             }
 
             $parameters = $this->getTagAttributes($tag, sprintf('The attribute name of tag "%s" for service "%s" in %s must be a non-empty string.', $tagName, (string) $service->getAttribute('id'), $file));
             foreach ($tag->attributes as $name => $node) {
-                if ('name' === $name) {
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
Thanks; Can you add a test case please?

nicolas-grekas (approved):
nvm, there are test cases already :)

PhilETaylor:
I have taken these proposed changes to `src/Symfony/Component/DependencyInjection/Loader/XmlFileLoader.php` and applied them, manually, in my live project that identified this issue originally, 

Then, after a cache:clear, this problem has been resolved. So I confirm this PR fixes my original issue, in the original app that was complaining. 

Thanks @MatTheCat - please consider signing up to GitHub Sponsors :)

MatTheCat:
Thanks for testing on a real project @PhilETaylor! Will look at this sponsorship thing :eyes:

nicolas-grekas:
Thank you @MatTheCat.

PR #50089, opened 2023-04-21
[DependencyInjection] More predictable EnvVarProcessor CSV empty string parsing

andreyserdjuk:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| License       | MIT

Small improvement, that solves problem of getting non-empty array: `[0 => null]`, when parse empty string by csv processor. After this fix '%env(csv:EMPTY_STRING)%' will return `[]`, but not `[0 => null]`.

### DIFF
--- a/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
+++ b/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
@@ -313,7 +313,7 @@ class EnvVarProcessor implements EnvVarProcessorInterface
         }
 
         if ('csv' === $prefix) {
-            return str_getcsv($env, ',', '"', '');
+            return '' === $env ? [] : str_getcsv($env, ',', '"', '');
         }
 
         if ('trim' === $prefix) {
--- a/src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php
+++ b/src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php
@@ -714,7 +714,7 @@ class EnvVarProcessorTest extends TestCase
 CSV;
 
         return [
-            ['', [null]],
+            ['', []],
             [',', ['', '']],
             ['1', ['1']],
             ['1,2," 3 "', ['1', '2', ' 3 ']],


nicolas-grekas:
Thanks, that makes sense. Can you add a test case please?

src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
```diff
@@ -77,7 +77,7 @@ public function getEnv(string $prefix, string $name, \Closure $getEnv): mixed
             }
 
             if (!isset($array[$key]) && !\array_key_exists($key, $array)) {
-                throw new EnvNotFoundException(sprintf('Key "%s" not found in %s (resolved from "%s").', $key, json_encode($array), $next));
+                throw new EnvNotFoundException(sprintf('Key "%s" not found in "%s" (resolved from "%s").', $key, json_encode($array), $next));
```
nicolas-grekas: this one should be reverted, it's a false-positive from fabbot

src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
```diff
@@ -313,6 +313,10 @@ public function getEnv(string $prefix, string $name, \Closure $getEnv): mixed
         }
 
         if ('csv' === $prefix) {
+            if ('' === $env) {
+                return [];
+            }
+
             return str_getcsv($env, ',', '"', '');
```
nicolas-grekas:
```suggestion
            return '' === $env ? [] : str_getcsv($env, ',', '"', '');
```

src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
andreyserdjuk: done

nicolas-grekas:
Thank you @andreyserdjuk.

andreyserdjuk:
> Thanks, that makes sense. Can you add a test case please?

Please look at the updated `src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php` which asserted before that csv parsing empty string returns `[null]`. If you mean to add one case, please specify it.

PR #50091, opened 2023-04-21
Fix the semaphore tests

stof:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

The save method of the store is not meant to return a boolean. And with the new `void` type in the phpdoc, phpunit complains about that.

### DIFF
--- a/src/Symfony/Component/Semaphore/Tests/SemaphoreFactoryTest.php
+++ b/src/Symfony/Component/Semaphore/Tests/SemaphoreFactoryTest.php
@@ -35,8 +35,7 @@ class SemaphoreFactoryTest extends TestCase
                 $keys[] = $key;
 
                 return true;
-            }))
-            ->willReturn(true);
+            }));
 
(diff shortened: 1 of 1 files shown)


src/Symfony/Component/Semaphore/Tests/SemaphoreFactoryTest.php
```diff
@@ -62,8 +60,6 @@ public function testCreateSemaphoreFromKey()
             ->method('save')
             ->with($this->callback(function ($key) use (&$keys) {
                 $keys[] = $key;
-
-                return true;
             }))
             ->willReturn(true);
```
nicolas-grekas: this line should be removed (same above)

nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @stof.

PR #50087, opened 2023-04-21
[Config] Improve GlobResource performance

staabm:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #... 
| License       | MIT
| Doc PR        | symfony/symfony-docs#... 

Re-order conditions in GlobResource to prevent unnecessary file IO, which improves performance of the following test-script.

put this script into the symfony/symfony project root and run it
```php
<?php

error_reporting(E_ALL);

require __DIR__.'/vendor/autoload.php';

$glob = new \Symfony\Component\Config\Resource\GlobResource(__DIR__.'/src/', '/**/*.php', true);
foreach($glob as $file) {
    echo $file->getRealPath()."\n";
}
```

I did a few measures and it consistently reports a 15-18% improvement

grafik

https://blackfire.io/profiles/compare/647af42c-c4e5-495e-8f8d-1b77f61631dd/graph

PR #50095, opened 2023-04-21
[Cache] Prevent unnecessary file IO

staabm:
| Q             | A
| ------------- | ---
| Branch?       | 6.3 
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #... 
| License       | MIT

Similar to https://github.com/symfony/symfony/pull/50087 this prevents file IO by re-ordering conditions.

----

There were even more similar code spots, but I wondered because they have different semantics
(different interpretation of success because of the placement of parantheses)

https://github.com/symfony/symfony/blob/51666290232ee13165c98fa9e40d3475a3cabfd8/src/Symfony/Component/Cache/Traits/FilesystemCommonTrait.php#L62
https://github.com/symfony/symfony/blob/51666290232ee13165c98fa9e40d3475a3cabfd8/src/Symfony/Component/Cache/Traits/FilesystemCommonTrait.php#L74

### DIFF
--- a/src/Symfony/Component/Cache/Adapter/PhpFilesAdapter.php
+++ b/src/Symfony/Component/Cache/Adapter/PhpFilesAdapter.php
@@ -82,7 +82,7 @@ class PhpFilesAdapter extends AbstractAdapter implements PruneableInterface
                 }
 
                 if ($time >= $expiresAt) {
-                    $pruned = $this->doUnlink($file) && !file_exists($file) && $pruned;
+                    $pruned = $this->doUnlink($file) && $pruned && !file_exists($file);
                 }
             }
         } finally {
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
I don't think this is correct. These are guard checks to me, the order matters.

stof:
Well, switching the IO check and the boolean check is fine to me.

However, the ones linked in the description indeed can be changed as we cannot change the order for _write_ IO operation (skipping a IO write is not an optimization opportunity)

nicolas-grekas:
BUT, the check looks wrong to me!

it should be "unlink || !file_exists"!
Can you send a PR on 5.4?

PR #49864, opened 2023-03-29
[Scheduler] add schedule name to `ScheduledStamp`

kbond:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | todo

As discussed in #49838, it would be nice to have the schedule name available on the stamp.

PR #49865, opened 2023-03-29
[Scheduler] allow `MessageGenerator`'s to return `RecurringMessage[]`

kbond:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | todo

- Ensures `SchedulerTransport` returns `RecurringMessage[]`
- Adds `RecurringMessage` to `ScheduledStamp`.

As discussed in #49838 and a companion to #49864.

PR #49838, opened 2023-03-27
[Scheduler] add `RecurringMessage::getId()` and prevent duplicates

kbond:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | todo

This could allow running specific scheduled tasks manually (via CLI or GUI).

PR #49792, opened 2023-03-24
[Scheduler] add "hashed" cron expression support

kbond:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | todo

This is an interesting feature from zenstruck/schedule-bundle that I borrowed from Jenkins. It helps with the problem touched upon in [Fabien's talk at SFLive Paris 2023](https://speakerdeck.com/fabpot/s?slide=33).

From the [zenstruck/schedule-bundle docs](https://github.com/zenstruck/schedule-bundle/blob/1.x/doc/define-tasks.md#hashed-cron-expression) _(some of the issues discussed below don't apply to symfony/scheduler the same way)_:

> If you have many tasks scheduled at midnight (0 0 * * *) this could create a very long running schedule right at this time. Tasks scheduled at the same time are run synchronously. This may cause an issue if a task has a memory leak.
>
> This bundle extends the standard Cron expression syntax by adding a # (for hash) symbol. # is replaced with a random value for the field. The value is deterministic based on the task's description. This means that while the value is random, it is predictable and consistent. A task with the description my task and a defined frequency of # # * * * will have a calculated frequency of 56 20 * * * (every day at 8:56pm). Changing the task's description will change it's calculated frequency. If the task from the previous example's description is changed to another task, it's calculated frequency would change to 24 12 * * * (every day at 12:24pm).
>
> A hash range #(x-y) can also be used. For example, # #(0-7) * * * means daily, some time between midnight and 7am. Using the # without a range creates a range of any valid value for the field. # # # # # is short for #(0-59) #(0-23) #(1-28) #(1-12) #(0-6). Note the day of month range is 1-28, this is to account for February which has a minimum of 28 days.

### Usage

```php
// MyMessage MUST be stringable to use "hashed cron"
$schedule->add(RecurringMessage::cron('#midnight', new MyMessage()));
```

PR #50096, opened 2023-04-21
[Scheduler] add dependency on symfony/messenger

### LINKED ISSUE #50085: [Scheduler] Expose the run date?
### Description

I'm currently assessing the possibility for one of my clients to switch their scheduled task to the new Scheduler component once 6.3 is released.
Everything is going great so far but I found one thing limiting them: the fact that it is not possible to access the run datetime. And using the current datetime does not really seems appropriate to me as it could be not in sync if the Scheduler is catching up on some missed schedules.

This looks to me like a quite common use case with scheduled tasks (when one wants to run some kind of report generation for example) where you need the current month or week to know which timespan to use. And so supporting it looks appealing.

Any thoughts? Does supporting this use case would make sense? 
/cc @fabpot @kbond (as it seems to me you have been the most active on this component so far :) )

I did some quick fiddling and it seems to be possible to yield the $time from the MessageGenerator, add it time to the ScheduledStamp and then the user is free to do what they want with a Middleware or a listener.
I have also tried to think about it could be done with events (#49803 ) but introducing a dependency to the Event Dispatcher only for this seems a bit overkill.
So far, I have not been able to find another more satisfactory way of supporting this use case.

Maybe someone has a better idea?

### Example

Here is the listener that I wrote in the app to support this use case, as you can see it becomes relatively easy to get the date and transmit it (or do some logging or whatever you want):

```
#[AsEventListener(WorkerMessageReceivedEvent::class)]
class ScheduledStampListener
{
    public function __invoke(WorkerMessageReceivedEvent $event): void
    {
        if(!$event->getEnvelope()->last(ScheduledStamp::class)) {
            return;
        }

        if (!$event->getEnvelope()->getMessage() instanceof ManagementInformationReportCommand) {
            return;
        }

        $event->addStamps(new HandlerArgumentsStamp([$event->getEnvelope()->last(ScheduledStamp::class)->at]));
    }
}
```

kbond:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #50085, Fix #49864, Helps with #49838, #49865
| License       | MIT
| Doc PR        | n/a

I think we should create a _hard_ dependency on `symfony/messenger`. This makes adding messenger-related features much easier. When using standalone, I've made the fact that messenger is used under the hood _completely_ transparent.

- 2d05655f18f88b939792cb43ab285560e6368141: switches the `Scheduler` to use messenger's `Worker` under the hood
- 8a9e1af95bb679bf5a4261cabde159e20039f354: Adds useful context to the `ScheduleStamp` (schedule name, trigger, current run time, next run time). This could be moved to a followup PR but wanted to demonstrate how much easier it is to add this context with the hard dep.

/cc @upyx, any objections here?

### DIFF
--- a/src/Symfony/Component/Scheduler/Messenger/ScheduledStamp.php
+++ b/src/Symfony/Component/Scheduler/Messenger/ScheduledStamp.php
@@ -12,10 +12,18 @@
 namespace Symfony\Component\Scheduler\Messenger;
 
 use Symfony\Component\Messenger\Stamp\NonSendableStampInterface;
+use Symfony\Component\Scheduler\Trigger\TriggerInterface;
 
 /**
  * @experimental
  */
 final class ScheduledStamp implements NonSendableStampInterface
 {
+    public function __construct(
+        public readonly ?string $name,
+        public readonly TriggerInterface $trigger,
+        public readonly \DateTimeImmutable $runAt,
+        public readonly \DateTimeImmutable $nextRun,
+    ) {
+    }
 }
--- a/src/Symfony/Component/Scheduler/Generator/ChainMessageGenerator.php
+++ b/src/Symfony/Component/Scheduler/Generator/ChainMessageGenerator.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace Symfony\Component\Scheduler\Generator;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ */
+final class ChainMessageGenerator implements MessageGeneratorInterface
+{
+    /**
+     * @param MessageGeneratorInterface[] $generators
+     */
+    public function __construct(private iterable $generators)
+    {
+    }
+
+    public function getMessages(): iterable
+    {
+        foreach ($this->generators as $generator) {
+            yield from $generator->getMessages();
+        }
+    }
+}
--- a/src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
+++ b/src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
@@ -13,6 +13,8 @@ namespace Symfony\Component\Scheduler\Generator;
 
 use Psr\Clock\ClockInterface;
 use Symfony\Component\Clock\Clock;
+use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Scheduler\Messenger\ScheduledStamp;
 use Symfony\Component\Scheduler\Schedule;
 use Symfony\Component\Scheduler\Trigger\TriggerInterface;
 
@@ -29,6 +31,7 @@ final class MessageGenerator implements MessageGeneratorInterface
         private readonly Schedule $schedule,
         string|CheckpointInterface $checkpoint,
         private readonly ClockInterface $clock = new Clock(),
+        private readonly ?string $name = null,
     ) {
         $this->waitUntil = new \DateTimeImmutable('@0');
         if (\is_string($checkpoint)) {
@@ -70,7 +73,7 @@ final class MessageGenerator implements MessageGeneratorInterface
             }
 
             if ($yield) {
-                yield $message;
+                yield Envelope::wrap($message, [new ScheduledStamp($this->name, $trigger, $time, $nextTime)]);
                 $this->checkpoint->save($time, $index);
             }
         }
--- a/src/Symfony/Component/Scheduler/Messenger/SchedulerTransport.php
+++ b/src/Symfony/Component/Scheduler/Messenger/SchedulerTransport.php
@@ -29,7 +29,7 @@ class SchedulerTransport implements TransportInterface
     public function get(): iterable
     {
         foreach ($this->messageGenerator->getMessages() as $message) {
-            yield Envelope::wrap($message, [new ScheduledStamp()]);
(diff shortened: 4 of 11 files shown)


src/Symfony/Component/Scheduler/Messenger/ScheduledStamp.php
```diff
@@ -12,10 +12,18 @@
...
 
 /**
  * @experimental
  */
 final class ScheduledStamp implements NonSendableStampInterface
 {
+    public function __construct(
+        public readonly ?string $name,
+        public readonly TriggerInterface $trigger,
+        public readonly \DateTimeImmutable $runAt,
```
kbond, line 25: Maybe we need a better name? This might not be the true run date (if the scheduler is catching up). Maybe `$expectedRunAt`?

/cc @tucksaun

tucksaun: `expectedRunAt` sounds indeed better than `runAt`.

Other suggestions:
- `scheduledAt`
- `triggeredAt` (to keep the idea of the trigger)

upyx:
@kbond On the one hand, it simplifies adding features. On the other hand, I'm afraid of lacking incapsulation and spreading internal details, which complicate future extension.

kbond:
Ok, I understand your concern, what about the following ideas:

1. Add events: before the `MessageGenerator` yeilds a message, it dispatches an event that has the context and listeners can manipulate the message (turn it into an envelope with stamps). I'm thinking this doesn't fully address your concern though.
2. Extract most of the logic from `MessageGenerator` into an abstract `BaseMessageGenerator` (or another object entirely). We then can create a `MessengerMessageGenerator` that takes this data (object, time, next time, trigger) and wraps in an envelope. In this case we choose to make this internal data available in a stamp. This would allow removing the hard-dependency on messenger.

I'm very open to other ideas!

tucksaun:
I might have something by mixing your approach and the one in #50085: we could imagine an intermediary class, not bound to Messenger, yield from the generator.
Then we attach it to the stamp in the SchedulerTransport. Direct users of the MessageGenerator can decide whether to use or not the provided context, or extend it, and as such we don't introduce any hard dependencies on Messenger.

See a draft here: https://github.com/symfony/symfony/compare/6.3...tucksaun:symfony:scheduler/generator-context?expand=1

kbond:
Nice (simple) approach! You chose not to add the next run timestamp? (I don't have a strong opinion on this as it can be retrieved from the trigger).

What do you think about including the schedule name? This is important context for me. Should it be added to the `MessageContext`, done like in #49864, or something else?

tucksaun:
> You chose not to add the next run timestamp.

I didn't think about it tbh, I just wanted to draft quickly if this could be a way to do things or not.
But I like it, so I could easily add it if we say that I should open a PR with this approach.

> What do you think about including the schedule name

In fact, I started to pick your change but then I was confused: should it go into the schedule or the generator? 🤔
So I went back to the same "draft" approach and left it aside for now.
Also, I was wondering, isn't it kind of available within the `ReceivedStamp` (in the context of Messenger only though)?

kbond:
> I didn't think about it tbh, I just wanted to draft quickly if this could be a way to do things or not.

I think it can't hurt to add. This will provide the expected next run time which may or may not match `TriggerInterface::getNextRunDate()`.

> But I like it, so I could easily add it if we say that I should open a PR with this approach.

I say open a PR - I like your approach better than this one. @upyx, what do you think?

> In fact, I started to pick your change but then I was confused: should it go into the schedule or the generator?

I've struggled with this myself. The Schedule object itself doesn't have access to it's name. This is configured with the `AsSchedule` service attribute so it's only currently available in `SchedulerTransportFactory` as a service alias. I think it makes sense as an optional argument for `MessageGenerator` - then could be added to your `MessageContext` object.

Since `MessageGenerator::__construct()` takes the `$checkpoint` as a `string|CheckpointInterface`, if it's a string and no `$name` is passed, we could use that.

kbond:
> Also, I was wondering, isn't it kind of available within the ReceivedStamp (in the context of Messenger only though)?

It is, and marked with the schedule transport name (`schedule_<name>`). This is how I'm accessing it now (by removing `schedule_`). So, if we don't add, I'll continue to use this.

---

For some context, I'm working on a generic messenger-monitor bundle that has scheduler support. This is why I want all this context. In this bundle, a "monitored message" has `queued_at`, `received_at` and `handled_at` timestamps. This allows for some nice reporting. Currently, for messages that are from a scheduler transport, the `queued_at` and `received_at` are identical (because scheduled messages aren't async). My plan, once this `triggeredAt` value is available, is to set this as the `queued_at` value for these monitored messages. This will allow reporting on how far off the expected run time and the actual run time are. I _think_ this is what you were looking to do in #50085.

tucksaun:
> I think this is what you were looking to do in https://github.com/symfony/symfony/issues/50085.

This is not the same reason but this requires the same data indeed 😃 

> I say open a PR - I like your approach better than this one

Done. See #50130

fabpot:
Close in favor of #49792

PR #50101, opened 2023-04-21
[DependencyInjection] Fix support for empty env vars

### LINKED ISSUE #50094: [6.3] blank env vars - Environment variable not found

6.3.x-dev

### Description

as per conversation with @nicolas-grekas here https://github.com/symfony/symfony/pull/48705#issuecomment-1516785234 and possibly related to #48705

Using the reproducer, which has a .env with `MY_EXAMPLE_ENV=` when clearing cache (clearing cache in enough in the reproducer to reproduce the issue, but in a live app loading any webpage also crashes with same message) 

Note the reproducer is[ creating a parameter](https://symfony.com/doc/current/best_practices.html#use-parameters-for-application-configuration) from the env var and then twig is consuming that parameter to [create global variables](https://symfony.com/doc/current/templates.html#global-variables). 

IT IS THIS CONSUMPTION by twig that generates the issue, without /config/twig.php trying to set a global from the parameter, the problem doesn't happen

ScreenShot-2023-04-21-12 59 14

### How to reproduce

Here is a reproducer 

 - Check out https://github.com/PhilETaylor/bug_app_env.git

This was created with 

- composer create-project symfony/skeleton bug_app_env
- cd bug_app_env
- change composer.json to be 6.3
- composer req twig
- add config/example.php to create a parameter
- add config/twig.php to consume that parameter

to replicate the bug then run `php bin/console cache:clear`

### Possible Solution

_No response_

### Additional Context

to be clear, `debug:dotenv` correctly loads and gives 

ScreenShot-2023-04-21-13 04 21

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50094
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
+++ b/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
@@ -149,7 +149,7 @@ class EnvVarProcessor implements EnvVarProcessorInterface
             || false === ($env = $env ?? getenv($name) ?? false) // null is a possible value because of thread safety issues
         ) {
             foreach ($this->loadedVars as $vars) {
-                if (false !== ($env = ($vars[$name] ?? false)) && '' !== $env) {
+                if (false !== ($env = ($vars[$name] ?? $env)) && '' !== $env) {
                     break;
                 }
             }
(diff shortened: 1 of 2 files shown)


PhilETaylor:
Works perfectly - thanks @nicolas-grekas

PR #50102, opened 2023-04-21
[ErrorHandler] Avoid compile crash while trying to find candidate when a class is not found

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

---

## Before

image

## After

image

### DIFF
--- a/src/Symfony/Component/ErrorHandler/ErrorEnhancer/ClassNotFoundErrorEnhancer.php
+++ b/src/Symfony/Component/ErrorHandler/ErrorEnhancer/ClassNotFoundErrorEnhancer.php
@@ -152,6 +152,14 @@ class ClassNotFoundErrorEnhancer implements ErrorEnhancerInterface
             }
         }
 
+        // Symfony may ship some polyfills, like "Normalizer". But if the Intl
+        // extension is already installed, the next require_once will fail with
+        // a compile error because the class is already defined. And this one
+        // does not throw a Throwable. So it's better to skip it here.
+        if (str_contains($file, 'Resources/stubs')) {
+            return null;
+        }
+
         try {
             require_once $file;
         } catch (\Throwable) {


src/Symfony/Component/ErrorHandler/ErrorEnhancer/ClassNotFoundErrorEnhancer.php
```diff
@@ -155,6 +155,14 @@ private function convertFileToClass(string $path, string $file, string $prefix):
             }
         }
 
+        // Symfony may ship some polyfills, like "Normalizer". But if the Intl
+        // extension is already installed, the next require_once will fail with
+        // a compile error because the class is already defined. And this one
+        // does not throw a Throwable. So it's better to skip it here.
+        if (str_contains($file, 'Resources/stubs')) {
```
stof, line 159: To me, this looks like the root cause of your issue is that the candidate is not properly identified to skip this on line 153

stof: Or maybe that it tries to perform class discovery based on the configured PSR4 prefixes of Composer, but the Composer class discovery feature (used to optimize the autoloader) has additional inputs (excluded folders)

lyrixx: I [pushed](https://github.com/lyrixx/test/commit/5d79e77714cc45b65a3068f497ab0a576f7abe3e) a reproducer.

Let me explain some bits:

* It's a legacy application 4.4
* It has a custom exception listener that does not catch this error
* The one in symfony seems to not catch it either
* We have some class that are defined in the `services.yaml` (ie: no autowire)
* **I** made a typo in the `services.yaml`: When I copied a FQCN, I put only the namespace, and I forgot the class. So I copied `App\Normalizer` instead of `App\Normalizer\FoobarNormalizer`

---

I understand this patch seems a bit odd, but this is pragmatic. And I already hit this error and it's very confusing. The `Normalizer` namespace is quite common, because of the Serializer component.

---

So I don't think the root cause is the one I description in the PR description.

lyrixx:
Side note : I noticed `Symfony\Component\ErrorHandler\ErrorHandler::getErrorEnhancers()` feature is not used with the "new" system with the ErrorRendered. 

Should I open a new issue to wire it?

nicolas-grekas (approved):
LGTM

>  I noticed Symfony\Component\ErrorHandler\ErrorHandler::getErrorEnhancers() feature is not used with the "new" system with the ErrorRendered.

Issue or PR yes. Maybe @yceruto has an idea about this?

lyrixx:
I'll look at it for the enhancers. Which branch should I target?

nicolas-grekas:
6.4 I suppose as that's more an improvement than a bugfix.

fabpot (approved):

fabpot:
Thank you @lyrixx.

PR #50053, opened 2023-04-18
[BrowserKit][HttpClient] Update the value of some user agents to comply with the RFC 9110 specification

javiereguiluz:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | fixes #50052
| License       | MIT
| Doc PR        | -

PR #50106, opened 2023-04-21
[BrowserKit][HttpClient] [BrowserKit, HttpClient] Minor updates in the default user agents

javiereguiluz:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

After merging #50053 I realized that some of my changes were not really needed.

According to the spec, the user agent _"consists of one or more product identifiers, each followed by zero or more comments"_

For example this user agent:

```
Mozilla/5.0 (iPad; U; CPU OS 3_2_1 like Mac OS X; en-us) AppleWebKit/531.21.10 (KHTML, like Gecko) Mobile/7B405
```

`Mozilla/5.0` is the product and version
`AppleWebKit/531.21.10` is a subproduct and its own version
In both cases the version is optional.

So, our user agent `Symfony BrowserKit` is correct:

* `Symfony` is the product, without version (completely legal)
* `BrowserKit` is the subproduct, and omits its version (legal too)

The real issue was with user agents like `Symfony HttpClient/Amp`, because `/Amp` would be interpreted as the version of the `HttpClient`, which is wrong (the version is `6.2`, `6.3`, etc.). The `Amp` text is a detail about a product, so it's common to wrap it in parenthesis.

So, sorry for the noise and I think this time this is fully correct. Thanks.

### DIFF
--- a/UPGRADE-6.3.md
+++ b/UPGRADE-6.3.md
@@ -1,12 +1,6 @@
 UPGRADE FROM 6.2 to 6.3
 =======================
 
-BrowserKit
-----------
-
- * The default user agent has been renamed from `Symfony BrowserKit` to `SymfonyBrowserKit`
-   to comply with the RFC 9110 specification
-
 Console
 -------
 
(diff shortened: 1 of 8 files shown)


nicolas-grekas:
Thank you @javiereguiluz.

stof (approved):

PR #50109, opened 2023-04-21
[FrameworkBundle] Add --show-aliases option to debug:router command

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Add the `--show-aliases` option to display all the route aliases, for example:

Screenshot from 2023-04-21 18-55-11

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/TextDescriptor.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/TextDescriptor.php
@@ -53,6 +53,10 @@ class TextDescriptor extends Descriptor
             $tableHeaders[] = 'Controller';
         }
 
+        if ($showAliases = $options['show_aliases'] ?? false) {
+            $tableHeaders[] = 'Aliases';
+        }
+
         $tableRows = [];
         foreach ($routes->all() as $name => $route) {
             $controller = $route->getDefault('_controller');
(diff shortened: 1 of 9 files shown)


94noni (approved):
is the option of deprecated route can be also added?

src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/TextDescriptor.php
```diff
@@ -47,12 +47,21 @@ public function __construct(FileLinkFormatter $fileLinkFormatter = null)
     protected function describeRouteCollection(RouteCollection $routes, array $options = []): void
     {
         $showControllers = isset($options['show_controllers']) && $options['show_controllers'];
+        $showAliases = $options['show_aliases'] ?? false;
```
javiereguiluz: Shouldn't we update all the other descriptors (JSON, Markdown, etc.) too?

fancyweb: Good point! `show_controllers` is only leveraged by the text descriptor. `show_aliases` should behave the same IMO.

But it's a good idea to add aliases info to other descriptors. I'll do it in this PR since it's related.

fabpot (approved):

fabpot:
Thank you @fancyweb.

PR #50110, opened 2023-04-21
[Console] Fix the usage of the zsh completion through the fpath discovery

stof:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Zsh completion files registered through the fpath discovery mecanism have 2 requirements:

- the file name must have a `_` prefix
- the file must start with a `#compdef` comment describing the registration for that file.

The file can then either execute the completion logic directly or call `compdef` to register a function for completion. Keeping the second approach allows to use the same generated script for the global use case than for a sourced configuration script.


chalasr (approved):

fabpot (approved):

fabpot:
Thank you @stof.

PR #50111, opened 2023-04-21
Fix the list of supported shells for completions in a phar

stof:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Using glob inside a phar does not work (it does not find anything). Using a DirectoryIterator on the other hand works with phars. This allows this command to compute the list of supported shells properly when used inside a phar.

Easy way to reproduce the bug: run `composer completion unsupported` (or more realistically `composer completion` in a zsh shell as zsh is not a supported shell in symfony/console 5.4) with composer 2.5.5 and see that you get this error with an empty list of supported shells:

```
Detected shell "unsupported", which is not supported by Symfony shell completion (supported shells: "").
```

### DIFF
--- a/src/Symfony/Component/Console/Command/DumpCompletionCommand.php
+++ b/src/Symfony/Component/Console/Command/DumpCompletionCommand.php
@@ -132,8 +132,14 @@ EOH
      */
     private function getSupportedShells(): array
     {
-        return array_map(function ($f) {
-            return pathinfo($f, \PATHINFO_EXTENSION);
-        }, glob(__DIR__.'/../Resources/completion.*'));
+        $shells = [];
+
+        foreach (new \DirectoryIterator(__DIR__.'/../Resources/') as $file) {
+            if (str_starts_with($file->getBasename(), 'completion.') && $file->isFile()) {
+                $shells[] = $file->getExtension();
+            }
+        }
+
+        return $shells;
     }
 }


fabpot (approved):

fabpot:
Thank you @stof.

guvra:
FYI this introduced a regression on our side because the folder "Resources" is now mandatory (which was not the case before), so our compiled phar file didn't work anymore because we only included php files in it.

Easily fixable on our side by forcing the phar compiler to include these completion files though.

stof:
Well, I would say that working with modified source code of the component is not really covered by our BC policy.
Also, the whole command would break if you don't include these completion files (executing the command will read those files to generate things).

guvra:
> Well, I would say that working with modified source code of the component is not really covered by our BC policy.

Yes I understand that, it's why i said that it's easily fixable on our side :+1: 

>  Also, the whole command would break if you don't include these completion files (executing the command will read those files to generate things).

Actually, when the Resources folder does not exist, the new code logic breaks any command that uses the symfony/console package, not just the command "DumpCompletionCommand" where the code logic was changed.

It's because the logic is contained in the method `configure`, which is called for all commands defined in the application.

PR #48371, opened 2022-11-28
 [ImportMap] Manage JavaScript dependencies without a JS toolchain 

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Tickets       | Fix https://github.com/symfony/symfony/issues/48349
| License       | MIT
| Doc PR        | todo

 Yarn, NPM, pnpm, Babel, SWC, Webpack, Rollup, Parcel... Is it really necessary to introduce so much complexity to create a beautiful and interactive website?

The Symfony UX initiative greatly simplified how to build frontend applications with Symfony by going back to the roots: server-side generated HTML (goodbye JSX, our good old Twig is back), and minimalist JavaScript thanks to Hotwire. However, to use Symfony UX, you still need to install, set up, and maintain a full JS toolchain... which is no picnic.

Is it still really necessary?! Browers recently gained features making it possible to do without most of them. Let's get rid of the complexity by using the web platform thanks to this new component!

The ImportMaps component is very similar to [importmap-rails](https://github.com/rails/importmap-rails/), but has command names similar to Composer.

# Usage

Add or update packages:

```
bin/console importmap:require '@hotwired/stimulus'
```

Remove packages:

```
bin/console importmap:remove '@hotwired/stimulus'
```

Update all packages to their latest versions:

```
bin/console importmap:update
```

Make the packages available in your app:

```twig
<!doctype html>
<head>
    {{ importmap() }}
    <script type="module">
    import { Controller } from "@hotwired/stimulus"
    // ...
    </script>
</head>
{# ... #}
```

## TODO

* [ ] Add tests
* [x] Do not hardcode the URL of the polyfill
* [x] Support downloading the JS files locally instead of using the CDN
* [ ] Update Symfony UX to support Import Maps in addition to Webpack Encore

PR #50112, opened 2023-04-21
[Asset] [AssetMapper] New AssetMapper component: Map assets to publicly available, versioned paths

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Partner of #48371 
| License       | MIT
| Doc PR        | TODO

Hi!

This will partners with and includes the importmaps PR #48371 (so that will no longer be needed). The goal is to allow you to write modern JavaScript & CSS, without needing a build system. This idea comes from Rails: https://github.com/rails/propshaft - and that heavily inspires this PR.

Example app using this: https://github.com/weaverryan/testing-asset-pipeline

Here's how it works:

A) You activate the asset mapper:

```yml
framework:
    asset_mapper:
        paths: ['assets/']
```

B) You put some files into your `assets/` directory (which sits at the root of your project - exactly like now with Encore). For example, you might create an `assets/app.js`, `assets/styles/app.css` and `assets/images/skeletor.jpg`.

C) Refer to those assets with the normal `asset()` function

```twig
<link rel="stylesheet" href="{{ asset('styles/app.css') }}">
<script src="{{ asset('app.js') }}" defer></script>

<img src="{{ asset('images/skeletor.jpg') }}">
```

That's it! The final paths will look like this:

```html
<link rel="stylesheet" href="/assets/styles/app-b93e5de06d9459ec9c39f10d8f9ce5b2.css">
<script src="/assets/app-1fcc5be55ce4e002a3016a5f6e1d0174.js" defer type="module"></script>
<img src="/assets/images/skeletor-3f24cba25ce4e114a3116b5f6f1d2159.jpg">
```

How does that work? 

* In the `dev` environment, a controller (technically a listener) intercepts the requests starting with `/assets/`, finds the file in the source `/assets/` directory, and returns it.
* In the `prod` environment, you run a `assets:mapper:compile` command, which copies all of the assets into `public/assets/` so that the real files are returned. It also dumps a `public/assets/manifest.json` so that the source paths (eg. `styles/app.css`) can be exchanged for their final paths super quickly.

### Extras Asset Compilers

There is also an "asset" compiler system to do some minor transformations in the source files. There are 3 built-in compilers:

A) `CssAssetUrlCompiler` - finds `url()` inside of CSS files and replaces with the real, final path - e.g. `url(../images/skeletor.jpg')` becomes `url(/assets/images/skeletor-3f24cba25ce4e114a3116b5f6f1d2159.jpg)` - logic taken from Rails
B) `SourceMappingUrlsCompiler` - also taken from Rails - if the CSS file already contains a sourcemap URL, this updates it in the same way as above (Note: actually ADDING sourcemaps is not currently supported)
C) `JavaScriptImportPathCompiler` - experimental (I wrote the logic): replaces relative imports in JavaScript files `import('./other.js')` with their final path - e.g. `import('/assets/other.123456abcdef.js')`.

### Importmaps

This PR also includes an "importmaps" functionality. You can read more about that in #48371. In short, in your code you can code normally - importing "vendor" modules and your own modules:

```
// assets/app.js

import { Application } from '@hotwired/stimulus';
import CoolStuff from './cool_stuff.js';
```

Out-of-the-box, your browser won't know where to load `@hotwired/stimulus` from. To help it, run:

```
./bin/console importmap:require '@hotwired/stimulus';
```

This will updated/add an `importmap.php` file at the root of your project:

```php
return [
    'app' => [
        'path' => 'app.js',
        'preload' => true,
    ],
    '@hotwired/stimulus' => [
        'url' => 'https://ga.jspm.io/npm:@hotwired/stimulus@3.2.1/dist/stimulus.js',
    ],
];
```

In your `base.html.twig`, you add: `{{ importmap() }}` inside your `head` tag. The result is something like this:

```
        
<script type="importmap">{"imports": {
    "app": "/assets/app-cf9cfe84e945a554b2f1f64342d542bc.js",
    "cool_stuff.js": "/assets/cool_stuff-10b27bd6986c75a1e69c8658294bf22c.js",
    "@hotwired/stimulus": "https://ga.jspm.io/npm:@hotwired/stimulus@3.2.1/dist/stimulus.js",
}}</script>
</script>
<link rel="modulepreload" href="/assets/app-cf9cfe84e945a554b2f1f64342d542bc.js">
<link rel="modulepreload" href="/assets/cool_stuff-10b27bd6986c75a1e69c8658294bf22c.js">
<script type="module">import 'app';</script>
```

A few important things:

~~A) In the final `assets/app.js`, the `import CoolStuff from './cool_stuff';` will change to `import CoolStuff from './cool_stuff.js';` (the `.js` is added)~~
B) When your browser parses the final `app.js` (i.e. `/assets/app-cf9cfe84e945a554b2f1f64342d542bc.js`), when it sees the import for `./cool_stuff.js` it will then use the `importmap` above to find the real path and download it. It does the same thing when it sees the import for `@hotwired/stimulus`.
C) Because `app.js` has `preload: true` inside `importmap.php`, it (and anything it or its dependencies import) will also be preloaded - i.e. the `link rel="modulepreload"` will happen. This will tell your browser to download `app.js` and `cool_stuff.js` immediately. The is helpful for `cool_stuff.js` because we don't want to wait for the browser to download `app.js` and THEN realize it needs to download `cool_stuff.js`. 

There is also an option to `--download` CDN paths to your local machine.

### Path "Namespaces" and Bundle Assets

You can also give each "path: in the mapper a "namespace" - e.g. an alternative syntax to the config is:

```yml
framework:
    asset_mapper:
        paths:
            assets: ''
            other_assets: 'other_namespace'
```

In this case, if there is an `other_assets/foo.css` file, then you can use `{{ asset('other_namespace/foo.css') }}` to get a path to it.

In practice, users won't do this. However, this DOES automatically add the `Resources/public/` or `public/` directory of every bundle as a "namespaced" path. For example, in EasyAdminBundle, the following code could be used:

```
<script src="{{ asset('bundles/easyadmin/login.js') }}"></script>
```

(Note: EA has some fancy versioning, but on a high-level, this is all correct). This would already work today thanks to `assets:install`. But as soon as this code is used in an app where the mapper is activated, the mapper would take over and would output a versioned filename - e.g.

```
<script src="/assets/bundles/easyadmin/login12345abcde.js"></script>
```

**OPEN QUESTIONS / NOTES**

* Only the "default" asset package uses the mapper. Extend to all?

TODO: 

* [x] Twig importmap() extension needs a test
* [x] Need a way to allow a bundle to hook into `importmap()` - e.g. to add `data-turbo-track` on the `script` tags.
* [x] Make the AssetMapper have lazier dependencies

There are also a number of smaller things that we probably need at some point:

A)  a way to "exclude" paths from the asset map
~~B) a cache warmer (or something) to generate the `importmap` on production~~
C) perhaps a smart caching and invalidation system for the contents of assets - this would be for dev only - e.g. on every page load, we shouldn't need to calculate the contents of EVERY file in order to get its public path. If only cool_stuff.js was updated, we should only need to update its contents to get its path.
D) `debug:pipeline` command to show paths
E) Perhaps an `{{ asset_preload('styles/app.css') }}` Twig tag to add `<link rel="modulepreload">` for non-JS assets. This would also add `modulepreload` links for any CSS dependencies (e.g. if `styles/app.css` `@import`s another CSS file, that would also be preloaded).

Cheers!


src/Symfony/Bundle/FrameworkBundle/Command/AssetsPipelinePublishCommand.php
```diff
@@ -0,0 +1,94 @@
...
+        private readonly bool $isDebug,
+    ) {
+        parent::__construct();
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setHelp(<<<'EOT'
+The <info>%command.name%</info> command compiles and moves all the assets in
```
stof: I would say `compiles and dumps` rather than `compile and moves`. Moving would imply removing the source

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -732,6 +732,18 @@ private function addAssetsSection(ArrayNodeDefinition $rootNode, callable $enabl
                         ->scalarNode('version_format')->defaultValue('%%s?%%s')->end()
                         ->scalarNode('json_manifest_path')->defaultNull()->end()
                         ->scalarNode('base_path')->defaultValue('')->end()
+                        ->arrayNode('pipeline')
+                            ->canBeEnabled()
+                            ->children()
+                                // add array node called "paths" that will be an array of strings
+                                ->arrayNode('paths')
```
stof: missing `->fixXmlConfig('path')` in the parent

src/Symfony/Component/Asset/composer.json
```diff
@@ -16,7 +16,9 @@
         }
     ],
     "require": {
-        "php": ">=8.1"
+        "php": ">=8.1",
+        "symfony/filesystem": "^5.4|^6.0",
+        "symfony/mime": "^5.4|^6.0"
```
stof: those should be optional dependencies, as only the asset pipeline needs them, which is an optional feature of the component (projects using webpack won't use the asset pipeline).

Alternatively, the AssetPipeline might be big enough to become its own component (depending on symfony/asset), with the additional benefit that FrameworkBundle could then have a smart default to enable it when the package is installed. Projects will generally need either the AssetPipeline or WebpackEncoreBundle but not both

stof (commented):
this misses tests for the whole Pipeline system.

weaverryan:
> this misses tests for the whole Pipeline system

It's one of my TODO items above :)

src/Symfony/Component/Asset/composer.json
weaverryan: Yup - I'd be interested in what others think about this. If we keep this in `Asset`, but make these optional (which is definitely more correct), then users might get met with extra `Run "composer require symfony/mime" to use the pipeline` type of messages. Unless we *also* add an "asset pipeline pack" that contains all of it... but that doesn't make me super excited either.

If I had to lean towards something, it would probably be to split this into a new `AssetPipeline` component.

src/Symfony/Bundle/FrameworkBundle/Resources/config/assets_pipeline.php
```diff
@@ -0,0 +1,78 @@
...
+                param('kernel.project_dir'),
+            ])
+        ->set('assets.pipeline.asset_package', PipelineAwareAssetPackage::class)
+            ->decorate('assets._default_package')
+            ->args([
+                service('.inner'),
+                service('assets.pipeline'),
+            ])
+
+        ->set('assets.pipeline.dev_server_subscriber', AssetPipelineDevServerSubscriber::class)
```
94noni: you describe in PR desc `In the dev environment, a controller (technically a listener)`
but I fail to see where this is conditionally registered in dev

src/Symfony/Component/Asset/Pipeline/AssetPipelineDevServerSubscriber.php
```diff
@@ -0,0 +1,82 @@
...
+            ->setEtag($this->assetPipeline->getDigest($asset))
+        ;
+
+        $event->setResponse($response);
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            // priority higher than RouterListener
```
94noni: this seems touchy, how we can assert that this comment stay true overtime ?

src/Symfony/Component/Asset/Pipeline/Compiler/CssAssetUrlCompiler.php
```diff
@@ -0,0 +1,62 @@
...
+
+use Symfony\Component\Asset\Pipeline\AssetCompilerInterface;
+use Symfony\Component\Asset\Pipeline\AssetPipeline;
+
+/**
+ * Resolves url() paths in CSS files.
+ *
+ * Originally sourced from https://github.com/rails/propshaft/blob/main/lib/propshaft/compilers/css_asset_urls.rb
+ */
+class CssAssetUrlCompiler implements AssetCompilerInterface
```
94noni: nitpick, what about StyleSheetAsset as you named JavaScript and not JS in the class below :)

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2994,4 +3014,19 @@ private function writeConfigEnabled(string $path, bool $value, array &$config):
         $this->configsEnabled[$path] = $value;
         $config['enabled'] = $value;
     }
+
+    private function getPublicDirectoryName(string $projectDir): string
+    {
+        $defaultPublicDir = 'public';
```
94noni: what about making this variable? still have some projects using /web as "public" directory 
can't we define in the kernel a public dir like there is a cache dir etc?

src/Symfony/Bundle/FrameworkBundle/Command/AssetsPipelinePublishCommand.php
```diff
@@ -0,0 +1,94 @@
...
+ * Publishes the assets in the pipeline to the final output directory.
+ *
+ * This command is intended to be used during deployment.
+ *
+ * @author Ryan Weaver <ryan@symfonycasts.com>
+ *
+ * @final
+ */
+#[AsCommand(name: 'assets:pipeline:publish', description: 'Publishes all pipeline assets to the final public output directory.')]
+class AssetsPipelinePublishCommand extends Command
```
94noni: what about 
```suggestion
class AssetsPipelineCompileCommand extends Command
```

to me, "publish" sound like "deploy" (I'm not native English speaker tho)
here it is just compiling/dumping the content

94noni (commented):
having a look on this :)
nice to get inspired by other ecosystems to improve frontend DX development for "backend" devs ^^

src/Symfony/Component/Asset/composer.json
fabpot: I would vote for a new component as well.

nicolas-grekas: if the split is just motivated by the extra deps, I think mime and filesystem are too niche in the new pipeline system to be a strong enough reason:

here is the only place where mime is used:
`return self::EXTENSIONS_MAP[$extension] ?? $this->mimeTypes->guessMimeType($filePath);` 

Given the list in `EXTENSIONS_MAP`, this means mime is almost never going to be used in practice. Instead of adding the dep to it, I'd rather suggest to throw an exception when an extenstion is unknown, and add a way to configure the mapping.

For filesystem, we're only using `isAbsolutePath`, so either we borrow some logic from there, our we're fine with the dep. Borrowing might be my preference.

I'd prefer having all this in the Asset component personally.

javiereguiluz:
@weaverryan how would be the workflow for folks using e.g. SCSS? Before we just used Webpack Encore. Should we keep using that now to compile it into CSS? Then, how do we connect the compiled assets from `public/build/` with this new pipeline feature? Thanks.

src/Symfony/Component/Asset/composer.json
stof: It is not only about the dependencies. It would also allow FrameworkBundle to enable the pipeline automatically based on the installed packages.

src/Symfony/Bundle/FrameworkBundle/Resources/config/assets_pipeline.php
stof: See the FrameworkExtension which conditionally removes that definition depending on whether the server is enabled or no (which defaults to `kernel.debug`)

src/Symfony/Component/Asset/Pipeline/AssetPipelineDevServerSubscriber.php
stof: changing the priority of RouterListener would be a BC break for the ecosystem anyway, as many projects may have listeners that must run either before or after the router.

stof:
@javiereguiluz if the assets are already in the `public` folder, you don't need the asset pipeline at all. The pipeline does not replace the existing way (which still makes sense when using the webpack stack)

javiereguiluz:
So, if you use anything that it's not pure CSS or JS, you can't use this, right?

weaverryan:
> @weaverryan how would be the workflow for folks using e.g. SCSS? Before we just used Webpack Encore. Should we keep using that now to compile it into CSS? Then, how do we connect the compiled assets from public/build/ with this new pipeline feature? Thanks.
> So, if you use anything that it's not pure CSS or JS, you can't use this, right?

This will be the NUMBER 1 question & thing to document. If you need to compile scss -> css or use something like Tailwind, the answer is that you can/should now use *their* standard tools. For example, for Sass, you would install the `sass` binary - https://sass-lang.com/install. You'd probably have an `assets/styles/*.scss` type of setup like you have now. When you run `sass`, you would output these to perhaps `assets/styles/*.css` or perhaps another directory like `assets/built/*.css`. 

Anyways, once you do that, you'll ultimately end up with plain, normal CSS files. And THOSE are what you'd reference in your app:

```twig
<link rel="stylesheet" href="{{ asset('styles/app.css') }}">
```

Same goes for Tailwind. We'll definitely need to give people specific instructions / recommendations on how to set this all up.

yguedidi:
@weaverryan being a "pipeline", what about providing extra asset compilers for those external tools?

weaverryan:
That’s definitely an option :). Those would use the underlying binaries (e.g. sass), which I’m pretty sure is what you’re thinking too. But yes, certainly an option! Rails does *something* like this - iirc for tailwind for example - but I haven’t dug into the specifics.

src/Symfony/Bundle/FrameworkBundle/Command/AssetsPipelinePublishCommand.php
weaverryan: It *is* an action you take during deploy though? Rails does call this command “precompile”, however…

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
weaverryan: For whatever reason, this IS configurable, but via composer.json. I stole this logic from the assets:install command, which seems to be the only other place where symfony is aware of your public directory.

src/Symfony/Component/Asset/Pipeline/AssetPipelineDevServerSubscriber.php
weaverryan: Yup, so actually not nearly as touchy as it would seem at first!

src/Symfony/Component/Asset/composer.json
weaverryan: And we could ship a recipe with a config file that adds your assets/ as a pipeline path. That recipe flexibility might be another reason - though I still need to think about how we’ll accomplish the mixture of the pipeline recipe, importmaps and stimulus recipes… as each system can technically be used without the others 😀

weaverryan:
I DO think this should be split into its own component if we want to make the recipe situation simpler. Basically, you could still install symfony/asset without opting into the pipeline - which is what Encore users would do, for example.  It if you installed symfony/asset-pipeline, then we could ship a config file that activates the pipeline and likely ships with a starting assets/ directory. 

Other than splitting and some windows tests (the code loves /‘s), this is ready for review.

Neirda24:
This seems neat ! However I'm a bit confused too by the `pipeline` naming. I was expecting a multi step configurable process where each output is injected as input of the next one. To me it looks a lot like what `assets.packages` behave at the moment I don't understand the `pipeline` here.

src/Symfony/Bundle/FrameworkBundle/Resources/config/assets_pipeline.php
94noni: thx @stof I get it better now

src/Symfony/Component/Asset/Pipeline/AssetPipelineDevServerSubscriber.php
94noni: copy that

Kocal:
> > @weaverryan how would be the workflow for folks using e.g. SCSS? Before we just used Webpack Encore. Should we keep using that now to compile it into CSS? Then, how do we connect the compiled assets from public/build/ with this new pipeline feature? Thanks.
> > So, if you use anything that it's not pure CSS or JS, you can't use this, right?
> 
> This will be the NUMBER 1 question & thing to document. If you need to compile scss -> css or use something like Tailwind, the answer is that you can/should now use _their_ standard tools. For example, for Sass, you would install the `sass` binary - [sass-lang.com/install](https://sass-lang.com/install). You'd probably have an `assets/styles/*.scss` type of setup like you have now. When you run `sass`, you would output these to perhaps `assets/styles/*.css` or perhaps another directory like `assets/built/*.css`.
> 
> Anyways, once you do that, you'll ultimately end up with plain, normal CSS files. And THOSE are what you'd reference in your app:
> 
> ```
> <link rel="stylesheet" href="{{ asset('styles/app.css') }}">
> ```
> 
> Same goes for Tailwind. We'll definitely need to give people specific instructions / recommendations on how to set this all up.

How will it works with HMR?

sanderdlm:
> @weaverryan being a "pipeline", what about providing extra asset compilers for those external tools?

If you go down this route you're just building a new Assetic. It seems more fitting to keep the scope of this component small and focused on vanilla CSS/JS, matching the related PR about importmaps (https://github.com/symfony/symfony/pull/48371). Webpack can still be the default option for people who need a compilation step.

weaverryan:
> This seems neat ! However I'm a bit confused too by the pipeline naming

I borrowed this name from rails "Propshaft is an asset pipeline library for Rails" - https://github.com/rails/propshaft - but i'm open to other names (though pipeline is pretty catchy). There IS a "compiler" system in there where you can make changes to the source code, but at the moment (and this should perhaps remain this way) that is meant to be lightweight - it's used, e.g., for updating import paths in JS/CSS files.

> ... Same goes for Tailwind. We'll definitely need to give people specific instructions / recommendations on how to set this all up.
> How will it works with HMR?

AFAIK, HMR doesn't work without a built system like Webpack. If someone wanted to get crazy, an HMR system could be built using Mercure - e.g. run a command that watches CSS files for changes, use Mercure to send that change to the frontend, and on the frontend, swap the old CSS file out for the new one. It would be a neat extra for someone to build in an external bundle, I think. You might even be able to do this without Mercure (since it's just a dev thing) by having an endpoint that stays alive forever, watches the file changes in a loop, and return the server events as they happen.

> If you go down this route you're just building a new Assetic. It seems more fitting to keep the scope of this component small and focused on vanilla CSS/JS, matching the related PR about importmaps (https://github.com/symfony/symfony/pull/48371). Webpack can still be the default option for people who need a compilation step.

This is fair. We shouldn't rush into adding asset compilers for transforming Sass or Tailwind. Let's document using those tools on their own the "normal" way and see how things go.

Neirda24:
```
I borrowed this name from rails "Propshaft is an asset pipeline library for Rails" - https://github.com/rails/propshaft - but i'm open to other names (though pipeline is pretty catchy). There IS a "compiler" system in there where you can make changes to the source code, but at the moment (and this should perhaps remain this way) that is meant to be lightweight - it's used, e.g., for updating import paths in JS/CSS files.
```
⚠️ I would emit a warning about the "Catchy" because I think one of the strong assets of Symfony is the fact that we can understand the purpose of each component by their name. "Pipeline" is often used in CI/CD context and can be very misleading. I don't really see the benfit of having a dedicated component where it seems to integrate very nicely with the Asset component. Shouldn't it be directly in here ?

weaverryan:
>  I would emit a warning about the "Catchy" because I think one of the strong assets of Symfony is the fact that we can understand the purpose of each component by their name.

Can you suggest some alternative names? I like descriptive names too - what kinds of words would be helpful? If not pipeline, then...

* AssetResolver?
* AssetCompiler?
* AssetExposer?
* AssetMapper?
* AssetPublisher?

> I don't really see the benfit of having a dedicated component where it seems to integrate very nicely with the Asset component. Shouldn't it be directly in here ?

From a practical perspective, if it lives in the same component as `Asset`, then we probably can't ship a recipe that enables the pipeline by default... because often people `composer require symfony/asset` to *just* get the `{{ asset() }}` function. And so we wouldn't want to activate a new system for them. By splitting, if you install `symfony/asset`, you get exactly what you have right now. THEN you can install `symfony/asset-pipeline` (or whatever we name it) to activate that new system in your app.

> it seems to integrate very nicely with the Asset component. Shouldn't it be directly in here 

It does integrate nicely :). At the same time, there is only 1 small class that that interacts with the Asset component! 95% of the code in the new component is standalone. I just split it into its own component, and it felt very natural.

src/Symfony/Bundle/FrameworkBundle/Command/AssetsPipelineCompileCommand.php
```diff
@@ -0,0 +1,94 @@
...
+            $manifest[$asset->getLogicalPath()] = $asset->getPublicPath();
+        }
+
+        $manifestPath = $publicDir.$this->pipeline->getPublicPrefix().AssetPipeline::MANIFEST_FILE_NAME;
+        $this->filesystem->dumpFile($manifestPath, json_encode($manifest, \JSON_PRETTY_PRINT));
+        $io->comment(sprintf('Manifest written to <info>%s</info>', $manifestPath));
+
+        if ($this->isDebug) {
+            $io->warning(sprintf(
+                'You are compiilng assets in development. Symfony will not serve any changed assets until you delete %s.',
```
jaapio: Compiling

MatTheCat:
`AssetExposer` seems fitting to me. From [Wiktionary](https://en.m.wiktionary.org/wiki/expose#Verb) “to expose” means

> To make available to [...] other programs.

IIUC the goal of this component is to give assets URLs, to expose them to user agents.

weaverryan:
I've just renamed to AssetMapper to try that out (after a few different people suggested this). The job of the component is to  map assets to a publicly available path.

Neirda24:
Cool @weaverryan  ! Sounds... more precise :) I still wonder : is it not almost the same as https://symfony.com/doc/current/components/asset.html#named-packages ? How would you recommend one over the other ? If we think your component further, wouldn't it have the same kind of configuration ? Being able to map from multiple places, on a different host with a customised public prefix, etc etc ?

stof:
Regarding the JavaScriptImportPathCompiler, I'm wondering whether we need it. Wouldn't it be better to integrate with import maps instead, which would avoid the need to rewrite JS files when their dependency change just to change the imported hash (and so busting them out of the cache as well) ?
Thus, doing such rewrite of import means that we need complex machinery to compute the file hashes in case of circular dependencies (you cannot compute the hash before rewriting imports as this would break cache busting)

stof:
@Neirda24 the asset component won't expose private files publicly, which is what AssetMapper does. But there is indeed a relation to those packages from `symfony/asset`: AssetMapper provides a special implementation of the asset package of `symfony/asset`. It extends that existing system with more features.

weaverryan:
> Regarding the JavaScriptImportPathCompiler, I'm wondering whether we need it. Wouldn't it be better to integrate with import maps instead, which would avoid the need to rewrite JS files when their dependency change just to change the imported hash

Very good job noticing this @stof - it’s maybe the most important detail in the PR. For others, this feature allows you to say things like “import ‘./other’” and it’ll convert the “other” to the final, versioned filename.

This is the ONE thing that I added that Rails doesn’t have. Rails wants you to “import ‘other.js’” and then register “other.js” in an importmap to point to it.

That really bothers me. The point of importmaps and this new philosophy is to “write JavaScript like you already are, but without the build system”. With Node or Webpack, you use “import ./other” and I feel strongly that this should still work. And we should leave importmaps to be used for external packages (again, so that your code looks exactly like it used to).

If we DID want to solve this cache problem you’re talking about, in theory, I’d rather do it by allowing the user to use ./other, then, in conjunction with importmaps, rewriting that to something like “assets/other.js” and automatically adding it to the importmap for the user. That might be cool - but beyond this PR’s scope :).

Neirda24:
I understand the same thing @stof . Just pointing out that I do not see the value of it being a dedicated component with a dedicated config. IMO it should be an improvement of the actual package system and included in the current Asset component. But if everyone says otherwise maybe I'm missing something...

stof:
@weaverryan browsers will also check the import map for relative imports in case the resolved URL matches the import map. So the import map can be used to handle hashed paths for imports too: https://github.com/WICG/import-maps#mapping-away-hashes-in-script-filenames

stof:
@Neirda24 please read the discussion at https://github.com/symfony/symfony/pull/50112#discussion_r1174102524

weaverryan:
@stof you’re the best. Thank you :). This didn’t work when I tested, because “other.js” wasn’t in my importmap. So, indeed, it seems like we should combine asset mapper and import map together. Then:

- A) when parsing app.js, we see an import for ./other. We rewrite this path simply to /assets/other.js, but we take note that we saw this file as a dependency (or can we just leave it as ./other?)
- B) we then automatically add “/assets/other.js” to the import map pointing to its versioned file.

By doing (B), we don’t require the user to need to think “hey, I have this relative import for other.js, I need to add it to my importmap”… which is how I missed this originally :).

WDYT?

stof:
AssetMapper should indeed automatically generate the importmap for any JS file for which it renames the file.

src/Symfony/Component/AssetMapper/AssetMapperDevServerSubscriber.php
```diff
@@ -0,0 +1,89 @@
...
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Event\RequestEvent;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * Functions like a controller that returns assets from the asset mapper.
+ */
+class AssetMapperDevServerSubscriber implements EventSubscriberInterface
```
onEXHovia: Maybe use `EventListener` namespace by analogy other components?

leroy0211:
How would AssetMapper handle importing of other assets into your js files, or importing npm modules? 

Sometimes you wish to import a css file into a js file, which when using webpack will be compiled into the dom as <style> tags. 
The same applies for icons or images, which can be loaded as base64. 

Would AssetMapper handle these use-cases too?

stof:
@leroy0211 asset mapper does not perform any bundling. Importing a CSS file in a JS file is not a supported feature of *browsers* (for now), so this won't work.

weaverryan:
This is now ready for review again!

* I've marked the component as experimental. That's a no-brainer - we can get 6 months to learn from real-world edge cases and optimize.
* I've integrated **importmaps** into this PR - replacing #48371. AssetMapper and importmaps play really well together.

Reminder: there is a super-easy-to-setup demo you can try https://github.com/weaverryan/testing-asset-pipeline

weaverryan:
Crossed off some final TODOs. This is ready to go. Failures are unrelated :)

src/Symfony/Component/AssetMapper/AssetDependency.php
```diff
@@ -0,0 +1,29 @@
...
+ * Represents a dependency that a MappedAsset has.
+ *
+ * @experimental
+ */
+class AssetDependency
+{
+    public function __construct(
+        public readonly MappedAsset $asset,
+        /**
+         * @var bool whether this dependency is immediately needed
```
nicolas-grekas: should be an `@param` on the constructor instead

src/Symfony/Component/AssetMapper/AssetMapper.php
```diff
@@ -0,0 +1,235 @@
...
+        }
+
+        $extension = MappedAsset::extractExtension($logicalPath);
+
+        if (isset(self::EXTENSIONS_MAP[$extension])) {
+            return self::EXTENSIONS_MAP[$extension];
+        }
+
+        if (null === $this->mimeTypes) {
+            throw new \LogicException(sprintf('The file extension "%s" from "%s" does not correspond to any known types so the mime type can\'t be determined. Run "composer require symfony/mime-type" to guess the mime type.', $extension, $logicalPath));
```
nicolas-grekas: instead, what about allowing one to configure the map of extensions (via constructor I suppose)?
I don't think using content-based mime-type detection is really useful in the context of assets.
Alternatively, we could skip adding this extension mechanism and wait for users to contribute one (or send PRs to add to the const)

src/Symfony/Component/AssetMapper/MappedAsset.php
```diff
@@ -0,0 +1,178 @@
...
+        // remove last part of publicPath and replace with last part of logicalPath
+        $publicPathParts = explode('/', $this->publicPath);
+        $logicalPathParts = explode('/', $this->logicalPath);
+        array_pop($publicPathParts);
+        $publicPathParts[] = array_pop($logicalPathParts);
+
+        return implode('/', $publicPathParts);
+    }
+
+    public static function extractExtension(string $logicalPath): string
```
nicolas-grekas: can't we drop this method and use the native call instead? less maintenance for us

src/Symfony/Component/AssetMapper/MappedAsset.php
```diff
@@ -0,0 +1,178 @@
...
+    private string $publicPath;
+    private string $sourcePath;
+    private string $content;
+    private string $digest;
+    private bool $isPredigested;
+    private ?string $mimeType;
+    /** @var AssetDependency[]  */
+    private array $dependencies = [];
+
+    public function __construct(private string $logicalPath)
```
nicolas-grekas: I'm wondering if we couldn't make all the properties of this class `public readonly`, and remove/most setters+getters?

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
```diff
@@ -0,0 +1,76 @@
...
+
+            $output .= <<<HTML
+
+                <!-- ES Module Shims: Import maps polyfill for modules browsers without import maps support -->
+                <script async src="$url"$attributeString></script>
+                HTML;
+        }
+
+        foreach ($this->importMapManager->getModulesToPreload() as $url) {
+            $url = htmlspecialchars($url, \ENT_COMPAT | \ENT_SUBSTITUTE, $this->charset);
```
nicolas-grekas: escapeAttributeValue

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
```diff
@@ -0,0 +1,76 @@
...
+
+            $output .= "\n<link rel=\"modulepreload\" href=\"{$url}\">";
+        }
+
+        if (null !== $entryPoint) {
+            $output .= "\n<script type=\"module\">import '$entryPoint';</script>";
```
nicolas-grekas: do we need some escaping of `$entryPoint`?

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
```diff
@@ -0,0 +1,76 @@
...
+        private readonly array $scriptAttributes = [],
+    ) {
+    }
+
+    public function render(?string $entryPoint = null): string
+    {
+        $attributeString = '';
+        foreach ($this->scriptAttributes as $name => $value) {
+            $attributeString .= ' ';
+            if ($value === true) {
```
nicolas-grekas: yoda rulez :)

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -0,0 +1,399 @@
...
+
+    private array $importMapEntries;
+    private array $modulesToPreload;
+    private string $json;
+
+    public function __construct(
+        private readonly AssetMapper $assetMapper,
+        private readonly string      $importMapConfigPath,
+        private readonly string      $vendorDir,
+        private readonly string      $provider = self::PROVIDER_JSPM,
```
nicolas-grekas: we don't align `$`, that'd require a php-cs-fixer rule if we start doing so
for now, we should keep consistency I suppose

src/Symfony/Component/AssetMapper/README.md
```diff
@@ -0,0 +1,21 @@
+AssetMapper Component
+=======================
```
nicolas-grekas: extra `==`

src/Symfony/Component/AssetMapper/Tests/Compiler/JavaScriptImportPathCompilerTest.php
```diff
@@ -0,0 +1,204 @@
...
+        $compiler = new JavaScriptImportPathCompiler();
+        $this->assertSame($expectedOutput, $compiler->compile($input, $asset, $assetMapper));
+        $actualDependencies = [];
+        foreach ($asset->getDependencies() as $dependency) {
+            $actualDependencies[$dependency->asset->getLogicalPath()] = $dependency->isLazy;
+        }
+        $this->assertEquals($expectedDependencies, $actualDependencies);
+    }
+
+    public function provideCompileTests(): iterable
```
nicolas-grekas: all providers should be static

src/Symfony/Component/AssetMapper/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+CHANGELOG
+=========
+
+6.3
+---
+
+ * added the component
```
nicolas-grekas: Add

src/Symfony/Component/AssetMapper/AssetMapperRepository.php
```diff
@@ -0,0 +1,134 @@
...
+            }
+        }
+
+        return $paths;
+    }
+
+    private function getDirectories(): array
+    {
+        $filesystem = new Filesystem();
+        if (null === $this->absolutePaths) {
```
nicolas-grekas: if (null !== ) + unindent everything :)

src/Symfony/Component/AssetMapper/AssetMapperCompiler.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public function compile(string $content, MappedAsset $mappedAsset, AssetMapper $assetMapper): string
+    {
+        foreach ($this->assetCompilers as $compiler) {
+            if (!$compiler->supports($mappedAsset)) {
+                continue;
+            }
+
+            $content = $compiler->compile($content, $mappedAsset, $assetMapper);
```
nicolas-grekas: so there can be many on one content string?

nicolas-grekas (commented):
Here are some random comments :)

src/Symfony/Component/AssetMapper/MappedAsset.php
stof: Without accessors, it is a lot harder to write a BC layer for properties than for methods though.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
stof: We definitely do need escaping in a JS context

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
stof: We explicit rejected aligning `=>` in multiline arrays years ago, to avoid diffs re-indenting the whole array (and causing git conflicts) any time a longer key is added. I'm very strongly in favor of not aligning the `$` either for promoted properties for the same reason.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -810,6 +813,66 @@ private function addAssetsSection(ArrayNodeDefinition $rootNode, callable $enabl
...
+                    ->children()
+                        // add array node called "paths" that will be an array of strings
+                        ->arrayNode('paths')
+                            ->info('Directories that hold assets that should be in the mapper')
+                            ->example(['assets/'])
+                            ->beforeNormalization()
+                                ->castToArray()
+                                ->ifTrue(function ($v) { return isset($v[0]); })
+                                // change indexed array to one where each value
+                                // becomes the key and all values are ''
```
stof: If you want paths as keys rather than values, you must use `->useAttributeAsKey()` on the node. Otherwise, when merging multiple files that configure this section, they will be merged as a list, not as a map, and all keys will be lost.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -810,6 +813,66 @@ private function addAssetsSection(ArrayNodeDefinition $rootNode, callable $enabl
         ;
     }
 
+    private function addAssetMapperSection(ArrayNodeDefinition $rootNode, callable $enableIfStandalone): void
+    {
+        $rootNode
+            ->children()
+                ->arrayNode('asset_mapper')
```
stof: The XSD file must be updated to include the config of this component

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2994,4 +3050,19 @@ private function writeConfigEnabled(string $path, bool $value, array &$config):
         $this->configsEnabled[$path] = $value;
         $config['enabled'] = $value;
     }
+
+    private function getPublicDirectoryName(string $projectDir): string
```
stof: This method needs to register the root composer.json as a resource, so that editing the config there rebuilds the container.

src/Symfony/Component/AssetMapper/AssetMapperRepository.php
```diff
@@ -0,0 +1,134 @@
...
+     *
+     * Key is the logical path, value is the absolute path.
+     *
+     * @return string[]
+     */
+    public function all(): array
+    {
+        $paths = [];
+        foreach ($this->getDirectories() as $path => $namespace) {
+            /** @var RecursiveDirectoryIterator $iterator */
```
stof: this `@var` is totally lying about the type of the variable

src/Symfony/Component/AssetMapper/AssetMapperDevServerSubscriber.php
```diff
@@ -0,0 +1,92 @@
...
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * Functions like a controller that returns assets from the asset mapper.
+ *
+ * @experimental
+ *
+ * @author Ryan Weaver <ryan@symfonycasts.com>
+ */
+class AssetMapperDevServerSubscriber implements EventSubscriberInterface
```
stof: I suggest making this event listener final, as there is no valid reason to extend it IMO

src/Symfony/Component/AssetMapper/AssetMapper.php
```diff
@@ -0,0 +1,235 @@
...
+namespace Symfony\Component\AssetMapper;
+
+use Symfony\Component\Mime\MimeTypesInterface;
+
+/**
+ * Finds and returns assets in the pipeline.
+ *
+ * @experimental
+ */
+class AssetMapper
```
stof: I suggest making classes `@final` as much as possible, to reduce the inheritance-based BC surface. All the work done to add types for instance was much easier when inheritance was not allowed.

src/Symfony/Component/AssetMapper/Compiler/AssetCompilerInterface.php
```diff
@@ -0,0 +1,32 @@
...
+ * @author Ryan Weaver <ryan@symfonycasts.com>
+ */
+interface AssetCompilerInterface
+{
+    public function supports(MappedAsset $asset): bool;
+
+    /**
+     * Applies any changes to the contents of the asset.
+     */
+    public function compile(string $content, MappedAsset $asset, AssetMapper $assetMapper): string;
```
stof: I think this interface should use a `AssetMapperInterface` as its argument type here when receiving a service as argument instead of requiring a particular implementation (we might want to decorate things in the future)

src/Symfony/Component/AssetMapper/AssetMapper.php
weaverryan: That mime type dep really bothered me too. I've replaced it with a configurable "extensions" array where you can pass in extra extensions. Then we'll see if there are any real-world cases where this isn't enough. It is a fairly minor feature anyways: it impacts only (A) the content-type returned by the dev controller (if you have a wacky file type, you probably aren't relying on it being downloaded and understood by your browser anyways) and (B) activating the asset compiles (again, if you have something wacky, it's unlikely you'd want any existing compilers to operate on it anyways).

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
```diff
@@ -0,0 +1,74 @@
...
+            $resolvedPath = $this->resolvePath(\dirname($asset->getLogicalPath()), $matches[1]);
+
+            $finalResolvedPath = $resolvedPath;
+            if (!str_ends_with($finalResolvedPath, '.js')) {
+                $finalResolvedPath .= '.js';
+            }
+            $dependentAsset = $assetMapper->getAsset($finalResolvedPath);
+
+            if (!$dependentAsset) {
+                // maybe this is a directory with an index.js file
```
stof: If the goal is to support writing JS in the standard way, this should not be supported. Browsers don't automatically load an index.js if you import something corresponding to a directory.

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
```diff
@@ -0,0 +1,74 @@
...
+    // https://regex101.com/r/VFdR4H/1
+    private const IMPORT_PATTERN = '/(?:import\s+(?:(?:\*\s+as\s+\w+|[\w\s{},*]+)\s+from\s+)?|\bimport\()\s*[\'"`](\.\/[^\'"`]+|(\.\.\/)+[^\'"`]+)[\'"`]\s*[;\)]?/m';
+
+    public function compile(string $content, MappedAsset $asset, AssetMapper $assetMapper): string
+    {
+        return preg_replace_callback(self::IMPORT_PATTERN, function ($matches) use ($asset, $assetMapper) {
+            $resolvedPath = $this->resolvePath(\dirname($asset->getLogicalPath()), $matches[1]);
+
+            $finalResolvedPath = $resolvedPath;
+            if (!str_ends_with($finalResolvedPath, '.js')) {
```
stof: Native imports don't allow omitting the `.js` extension. I would say that this addition of a missing extension goes against the spirit of making you write JS in the modern way (plus, it is broken if you file uses the `.mjs` extension for instance)

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
```diff
@@ -0,0 +1,74 @@
...
+use Symfony\Component\AssetMapper\MappedAsset;
+
+/**
+ * Resolves import paths in JS files.
+ *
+ * @experimental
+ *
+ * @author Ryan Weaver <ryan@symfonycasts.com>
+ */
+class JavaScriptImportPathCompiler implements AssetCompilerInterface
```
stof: Do we still need this ? Can't we rely on the import map for project files too, by having all the available local assets in the import map (when digests are enabled and so changing the file name is needed) ?

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -0,0 +1,399 @@
...
+     *
+     * Returns an array of the entries that were added.
+     *
+     * @param PackageRequireOptions[] $packagesToRequire
+     * @param ImportMapEntry[] $importMapEntries
+     */
+    private function requirePackages(array $packagesToRequire, array &$importMapEntries): array
+    {
+        if (!$packagesToRequire) {
+            return $importMapEntries;
```
stof: should this really return the current entries here ? They have not been added AFAICT.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -0,0 +1,399 @@
...
+        return $newEntries;
+    }
+
+    /**
+     * Gets information about (and optionally downloads) the packages & updates the entries.
+     *
+     * Returns an array of the entries that were added.
+     *
+     * @param PackageRequireOptions[] $packagesToRequire
+     * @param ImportMapEntry[] $importMapEntries
```
stof: looking at the usages, this seems to be a `array<string, ImportMapEntry>`

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -0,0 +1,399 @@
...
+
+        $assetPath = $this->assetMapper->getFilesystemPath($entry->path);
+
+        if (is_file($assetPath)) {
+            @unlink($assetPath);
+        }
+    }
+
+    /**
+     * @return ImportMapEntry[]
```
stof: This returns `array<string, ImportMapEntry>`

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
```diff
@@ -0,0 +1,76 @@
...
+                $attributeString .= $name;
+
+                continue;
+            }
+            $attributeString .= sprintf('%s="%s"', $name, $this->escapeAttributeValue($value));
```
stof: shouldn't we also escape names ? We have no idea whether they come from a trusted source of no (and this requires a stricter escaping than values as they are not inside quotes. See Twig's `html_attr` escaping for the logic)

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
```diff
@@ -0,0 +1,76 @@
...
+        private readonly string $charset = 'UTF-8',
+        private readonly string|false $polyfillUrl = ImportMapManager::POLYFILL_URL,
+        private readonly array $scriptAttributes = [],
+    ) {
+    }
+
+    public function render(?string $entryPoint = null): string
+    {
+        $attributeString = '';
+        foreach ($this->scriptAttributes as $name => $value) {
```
stof: the `type` name should probably be forbidden. And for safety, we should probably also forbid `src`

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
```diff
@@ -0,0 +1,76 @@
...
+            </script>
+            HTML;
+
+        if ($this->polyfillUrl) {
+            $url = $this->escapeAttributeValue($this->polyfillUrl);
+
+            $output .= <<<HTML
+
+                <!-- ES Module Shims: Import maps polyfill for modules browsers without import maps support -->
+                <script async src="$url"$attributeString></script>
```
stof: should we really put the attributes on both the importmap script and the polyfill script ?

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
weaverryan: @stof can you mention a problematic input? Obviously if `$entrypoint` had a `'` in it, that'd be a problem. But that's not allowed. What are the cases we need to escape? thx :)

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
weaverryan: he sure does!

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
weaverryan: np - something in my toolset did this for me :) 

Edit: I think that "toolset" was Kévin Dunglas when I stole his code 😆

src/Symfony/Component/AssetMapper/AssetMapperCompiler.php
weaverryan: There's no use-case specifically for that at the moment, but yea - and I think that's a safe thing to allow.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
weaverryan: thank you - this stuff is tough! I also added `->normalizeKeys(false)` - pretty sure I missed that too in this spot.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
stof: The class does not make any validation about the content of `$entrypoint`. So anything is allowed.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -810,6 +813,66 @@ private function addAssetsSection(ArrayNodeDefinition $rootNode, callable $enabl
...
+                    ->info('Asset Mapper configuration')
+                    ->{$enableIfStandalone('symfony/asset-mapper', AssetMapper::class)}()
+                    ->fixXmlConfig('path')
+                    ->fixXmlConfig('importmap_script_attribute')
+                    ->children()
+                        // add array node called "paths" that will be an array of strings
+                        ->arrayNode('paths')
+                            ->info('Directories that hold assets that should be in the mapper')
```
stof: This info is incomplete IMO if this is indeed a map and not a list. What are the values ?

src/Symfony/Component/AssetMapper/AssetMapperRepository.php
weaverryan: you're right - forgot about that. I've fixed it - but it's still a bit ugly.

src/Symfony/Component/AssetMapper/AssetMapper.php
weaverryan: Good catch - I've added this to everything (a real `final` when possible other wise `@final` when we're mocking it)

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
weaverryan: Sigh. Yea... we need to find the right balance here. I'm trying to make coding with importmaps identical to coding with Webpack. But... bending standards/rules is shaky business. But it's painful to have my editor autocomplete `import './foo'` and my editor is even aware what that means... then it would simply break when I refresh the page with a 404 to `/assets/foo`. @dunglas?

> plus, it is broken if you file uses the .mjs extension for instance

This could be fixed by first looking to see if the file exists without changing it. Then checking for `.js` and `.mjs`.

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
weaverryan: ~~Hmm. It absolutely IS needed so that we can automatically discover that `app.js` is importing `./other.js`, and so automatically add `/assets/other.js` to the importmap (and also preload it if `app.js` is marked as preloaded).~~

~~What is less certain and relates to your [previous question](https://github.com/symfony/symfony/pull/50112#discussion_r1180681205) is whether or not this class should *adjust* the paths - e.g. `import './other'` to `import './other.js'`.~~

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
weaverryan: To me, these ARE trusted values. They're passed in through the constructor. And in practice, they are configured in a YAML file. Do we really think this is a possible security vector where these are controlled by an untrusted source?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
weaverryan: This is a tricky bit of config. The description should be better now.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
weaverryan: I saw a YouTube video where DHH was demo'ing Rails and it had this ;). I kind of agree with you - but I was playing it safe.

weaverryan:
I've got a "false" failure from psalm:

> Error: src/Symfony/Component/AssetMapper/MappedAsset.php:58:9: InaccessibleProperty: Symfony\Component\AssetMapper\MappedAsset::$publicPath is marked readonly (see https://psalm.dev/054)

Referring to https://github.com/symfony/symfony/blob/26d0d028d417d36726a671d72d4476f699243e83/src/Symfony/Component/AssetMapper/MappedAsset.php#L58

But in fact we ARE allowed to set that value... once. Really, I should remove the `if (isset($this->publicPath)) {` above that - it's not strictly needed. But that wouldn't fix the error. Ideas?

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
weaverryan: I would consider removing this functionality (adding the `.js` on behalf of the user) if we added a "strict" type (enabled by default) to asset mapper where if we don't recognize a path, we throw an error. Then, instead of things just "not working" + getting a 404 for `/asset/foo` when you refresh, we could give a hard error:

> The import for `./foo` in `assets/app.js` could not be found. Did you mean `./foo.js`?

dunglas: I agree with @stof. The goal here is not to replace Webpack but to "use the platform" and promote using standard modern JS as much as possible. Standard editors and browser developers decided to not support this feature, so we shouldn't.
This philosophy will also allows
 to keep this component as small and as simple as possible.

To be honest I would even go as far as **not** supporting the `./localfile.js` syntax, as Rails.

src/Symfony/Component/AssetMapper/AssetMapper.php
```diff
@@ -0,0 +1,228 @@
...
+ *
+ * @experimental
+ *
+ * @final
+ */
+class AssetMapper implements AssetMapperInterface
+{
+    public const MANIFEST_FILE_NAME = 'manifest.json';
+    private const EXTENSIONS_MAP = [
+        'js' => 'application/javascript',
```
dunglas:
```suggestion
        'js' => 'application/javascript',
        'mjs' => 'application/javascript',

```

src/Symfony/Component/AssetMapper/AssetMapper.php
```diff
@@ -0,0 +1,228 @@
...
+ * Finds and returns assets in the pipeline.
+ *
+ * @experimental
+ *
+ * @final
+ */
+class AssetMapper implements AssetMapperInterface
+{
+    public const MANIFEST_FILE_NAME = 'manifest.json';
+    private const EXTENSIONS_MAP = [
```
dunglas: Maybe should we import this database (not sure if it is worth it): https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
```diff
@@ -0,0 +1,75 @@
...
+            $pureOriginalImportName = ltrim($matches[1], './');
+
+            // tweak the import in case it needs the .js or /index.js added
+            return str_replace($pureOriginalImportName, $finalResolvedPath, $matches[0]);
+        }, $content);
+    }
+
+    public function supports(MappedAsset $asset): bool
+    {
+        return 'application/javascript' === $asset->mimeType;
```
dunglas:
```suggestion
        return 'application/javascript' === $asset->mimeType || 'text/javascript' === $asset->mimeType;
```

Just in case.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -0,0 +1,399 @@
...
+            }
+            $installData[] = $constraint;
+            $packageRequiresByName[$requireOptions->packageName] = $requireOptions;
+        }
+
+        $json = [
+            'install' => $installData,
+            'flattenScope' => true,
+            // always grab production-ready assets
+            'env' => ['browser', 'module', 'production'],
```
dunglas: Shouldn't we do this only in production mode to ease debugging?

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
weaverryan: Ok, I will remove this - and add a strict mode.

> To be honest I would even go as far as not supporting the ./localfile.js syntax, as Rails.

Using `import './localfile.js` IS valid JS though, no? Like, if we were building everything directly inside of `public/` (with no moving around of files), using `./localfile.js` would work out-of-the-box. The only reason it doesn't work out-of-the-box for us is because we're digesting the files. And *making* this work is quite easy: we "notice" the imported file and make sure it's in the importmap.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
weaverryan: I THINK what I have is correct, but tell me if you disagree. Try switching to `dev` mode and adding `react` in the builder - https://generator.jspm.io/ - you'll get `https://ga.jspm.io/npm:react@18.2.0/dev.index.js`. So even though we are adding `react` while coding in the `dev` environment, if we pass `dev` here, we'll get a CDN to the dev package. And then when the user deploys, they will still have the `dev` version.

So, if we wanted to be nicer here, I think we'd need to maintain both a `dev` and `prod` version of the URL inside of `importmap.php`. Not sure that's worth it.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
weaverryan: > Obviously if $entrypoint had a ' in it, that'd be a problem. But that's not allowed

Actually, I see that I was wrong about this - `'` would be possible to allow. So we should escape that. Anything else? I don't think we need to escape anything else because this argument is coming from the developer. Also, if we escape `'`, then... would it even be theoretically possible for someone to close the string and inject XSS? I don't think so...

Update: Escaping added for `'`

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
dunglas: Sure but the developper isn't working in the public directory and I'm not fond of editing the JS files on the fly. Import Maps allow to have something simpler. But I'll not block this feature!

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
dunglas: Makes sense! However having the ability to install the dev builds is a must have for debugging.

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
weaverryan: > Sure but the developper isn't working in the public directory and I'm not fond of editing the JS files on the fly. Import Maps allow to have something simpler. But I'll not block this feature!

If we remove the rewriting of `./foo` to `./foo.js` (which i AM about to remove), then there is no "editing JS files" I can happily say - and that's thanks to importmaps. If we see `import './foo.js'`, we won't change it. What we WILL do is add an importmap entry for that `foo.js` file automatically so that the browser can find the digested filename. So, I think we're both happy? :)

dunglas: Indeed, perfect!

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
weaverryan: I'll save this for a future PR - I think it'll involve making 2x requests to jspm - one for dev and one for prod - then maintain 2 URLs (when necessary) in `importmap.php`. I'm not aware of any packages that have a different dev vs prod that I would use, but I'm sure there are some (`react` is an example, but using react with importmaps seems edge-case).

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
weaverryan: Let me answer your question better :p. Yes, we could decide to put EVERY JavaScript file in the "asset mapper" system into the `importmap` instead of actually looking to see which are needed (i.e. which are being imported). That just felt wasteful to me. But looking for the imports does add complexity.

However, BECAUSE we look for the imports, we can give really nice errors - e.g. if I `import './other'` and forgot my .js:

> Unable to find asset “other” imported from “/path/to/project/assets/app.js”. Try adding “.js” to the end of the import - i.e. “other.js”.

weaverryan:
This is ready to go! Tests are green (failures unrelated), some controversial parts have been ironed out.

fabpot:
Thank you @weaverryan.

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
stof: Note that for the part where your IDE autocomplete `./foo`, this is configurable (I know it is in PHPStorm, and I'm almost sure VS Code has an equivalent config option)

PR #50125, opened 2023-04-23
[HttpKernel] Fix handling of `MapRequest*` attributes

### LINKED ISSUE #50120: [HttpKernel] `#[MapRequestPayload]` is "handled" before `#[IsGranted]`

6.3

### Description

After playing a bit with [this new feature](https://symfony.com/blog/new-in-symfony-6-3-mapping-request-data-to-typed-objects), I noticed that the payload is getting mapped/validated **before** ensuring that the user is actually allowed to access the resource.

This seems odd to me and I'd like to know your opinion about that. 

### How to reproduce

1. create a route and its payload.
2. use the `#[MapRequestPayload]` attribute on the payload argument (in the controller)
3. secure the route using the `#[IsGranted]` attribute
4. access the route by posting some data (voluntarily malformed data) but without being logged-in
5. see the error message related to the malformed data (and not because the user is not logged-in)

### Possible Solution

Few years ago, I created a bundle to achieve the same thing that this feature is offering, and I encountered the same problem. 
I played around with the "priority" option but without luck. So my only solution was to create a listener on the [ControllerArgumentsEvent](https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/HttpKernel/Event/ControllerArgumentsEvent.php) instead of using an ArgumentResolver. That way, the `#[IsGranted]` attribute would always be fired before my listener. No clue if I did things right at the time, but that's what worked for me.

### Additional Context

_No response_

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50120
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/web.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/web.php
@@ -70,6 +70,8 @@ return static function (ContainerConfigurator $container) {
                 service('translator')->nullOnInvalid(),
             ])
             ->tag('controller.targeted_value_resolver', ['name' => RequestPayloadValueResolver::class])
+            ->tag('kernel.event_subscriber')
+            ->lazy()
 
         ->set('argument_resolver.request_attribute', RequestAttributeValueResolver::class)
             ->tag('controller.argument_value_resolver', ['priority' => 100, 'name' => RequestAttributeValueResolver::class])
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
@@ -381,7 +381,8 @@ class FrameworkExtension extends Extension
                 );
 
             $container->getDefinition('argument_resolver.request_payload')
-                ->replaceArgument(0, new Reference('.argument_resolver.request_payload.no_serializer', ContainerInterface::RUNTIME_EXCEPTION_ON_INVALID_REFERENCE));
+                ->replaceArgument(0, new Reference('.argument_resolver.request_payload.no_serializer', ContainerInterface::RUNTIME_EXCEPTION_ON_INVALID_REFERENCE))
+                ->clearTag('kernel.event_subscriber');
 
             $container->removeDefinition('console.command.serializer_debug');
         }
--- a/src/Symfony/Component/HttpKernel/Attribute/MapQueryString.php
+++ b/src/Symfony/Component/HttpKernel/Attribute/MapQueryString.php
@@ -12,6 +12,7 @@
 namespace Symfony\Component\HttpKernel\Attribute;
 
 use Symfony\Component\HttpKernel\Controller\ArgumentResolver\RequestPayloadValueResolver;
+use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
 use Symfony\Component\Validator\Constraints\GroupSequence;
 
 /**
@@ -22,6 +23,8 @@ use Symfony\Component\Validator\Constraints\GroupSequence;
 #[\Attribute(\Attribute::TARGET_PARAMETER)]
 class MapQueryString extends ValueResolver
 {
+    public ArgumentMetadata $metadata;
+
     public function __construct(
         public readonly array $serializationContext = [],
         public readonly string|GroupSequence|array|null $validationGroups = null,
--- a/src/Symfony/Component/HttpKernel/Attribute/MapRequestPayload.php
+++ b/src/Symfony/Component/HttpKernel/Attribute/MapRequestPayload.php
@@ -12,6 +12,7 @@
 namespace Symfony\Component\HttpKernel\Attribute;
 
 use Symfony\Component\HttpKernel\Controller\ArgumentResolver\RequestPayloadValueResolver;
+use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
 use Symfony\Component\Validator\Constraints\GroupSequence;
 
 /**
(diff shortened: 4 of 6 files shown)


artyuum:
@nicolas-grekas I cloned your repo, checked out your branch and used the `link` script to copy that specific version of Symfony into a working Sf project and I'm having the following error (even after nuking the cache):
> Attempted to call an undefined method named "onKernelControllerArguments" of class "Symfony\Component\HttpKernel\Controller\ArgumentResolver\TraceableValueResolver".

<details><summary>Stacktrace</summary>

```text
Symfony\Component\ErrorHandler\Error\UndefinedMethodError:
Attempted to call an undefined method named "onKernelControllerArguments" of class "Symfony\Component\HttpKernel\Controller\ArgumentResolver\TraceableValueResolver".

  at /var/www/html/test/vendor/symfony/event-dispatcher/Debug/WrappedListener.php:116
  at Symfony\Component\EventDispatcher\Debug\WrappedListener->__invoke()
     (/var/www/html/test/vendor/symfony/event-dispatcher/EventDispatcher.php:220)
  at Symfony\Component\EventDispatcher\EventDispatcher->callListeners()
     (/var/www/html/test/vendor/symfony/event-dispatcher/EventDispatcher.php:56)
  at Symfony\Component\EventDispatcher\EventDispatcher->dispatch()
     (/var/www/html/test/vendor/symfony/event-dispatcher/Debug/TraceableEventDispatcher.php:139)
  at Symfony\Component\EventDispatcher\Debug\TraceableEventDispatcher->dispatch()
     (/var/www/html/test/vendor/symfony/http-kernel/HttpKernel.php:161)
  at Symfony\Component\HttpKernel\HttpKernel->handleRaw()
     (/var/www/html/test/vendor/symfony/http-kernel/HttpKernel.php:74)
  at Symfony\Component\HttpKernel\HttpKernel->handle()
     (/var/www/html/test/vendor/symfony/http-kernel/EventListener/ErrorListener.php:108)
  at Symfony\Component\HttpKernel\EventListener\ErrorListener->onKernelException()
     (/var/www/html/test/vendor/symfony/event-dispatcher/Debug/WrappedListener.php:116)
  at Symfony\Component\EventDispatcher\Debug\WrappedListener->__invoke()
     (/var/www/html/test/vendor/symfony/event-dispatcher/EventDispatcher.php:220)
  at Symfony\Component\EventDispatcher\EventDispatcher->callListeners()
     (/var/www/html/test/vendor/symfony/event-dispatcher/EventDispatcher.php:56)
  at Symfony\Component\EventDispatcher\EventDispatcher->dispatch()
     (/var/www/html/test/vendor/symfony/event-dispatcher/Debug/TraceableEventDispatcher.php:139)
  at Symfony\Component\EventDispatcher\Debug\TraceableEventDispatcher->dispatch()
     (/var/www/html/test/vendor/symfony/http-kernel/HttpKernel.php:224)
  at Symfony\Component\HttpKernel\HttpKernel->handleThrowable()
     (/var/www/html/test/vendor/symfony/http-kernel/HttpKernel.php:117)
  at Symfony\Component\HttpKernel\HttpKernel->terminateWithException()
     (/var/www/html/test/vendor/symfony/http-kernel/EventListener/DebugHandlersListener.php:74)
  at Symfony\Component\HttpKernel\EventListener\DebugHandlersListener::Symfony\Component\HttpKernel\EventListener\{closure}()
     (/var/www/html/test/vendor/symfony/error-handler/ErrorHandler.php:537)
  at Symfony\Component\ErrorHandler\ErrorHandler->handleException()
```

</details> 

Do you have the same error?

nicolas-grekas:
Yes, I do. This should now be fixed. Can you please try again?

artyuum:
@nicolas-grekas it works. I commented on the issue.

Koc:
We should be careful with usage of the payload as Subject from Controller Arguments in [`#[IsGranted`](https://symfony.com/doc/current/security/expressions.html) Attribute

src/Symfony/Bundle/FrameworkBundle/Resources/config/web.php
```diff
@@ -63,14 +63,24 @@
...
             ->tag('controller.argument_value_resolver', ['priority' => 100, 'name' => DateTimeValueResolver::class])
 
-        ->set('argument_resolver.request_payload', RequestPayloadValueResolver::class)
+        ->set('request_payload_value_resolver', RequestPayloadValueResolver::class)
             ->args([
                 service('serializer'),
                 service('validator')->nullOnInvalid(),
                 service('translator')->nullOnInvalid(),
             ])
+            ->lazy()
```
nicolas-grekas: I made this lazy to prevent instantiating the serialize+validator+translator tuple on controllers that don't use `MapRequest*`.

nicolas-grekas (commented):
> We should be careful with usage of the payload as Subject from Controller Arguments in [#[IsGranted](https://symfony.com/doc/current/security/expressions.html) Attribute

Accessing unserialized payloads in `#[IsGranted]` directly contradicts the expectations described in #50120: the expectation there is that a payload shouldn't be parsed if we can rule out access to some route before said parsing happens, which makes sense to me.

One way to solve this would be to allow registering an IsGranted attribute for late evaluation, aka after other controller argument listeners. There might be others.

Let's merge this PR and solve late the case of accessing the parsed subject from IsGranted?

src/Symfony/Bundle/FrameworkBundle/Resources/config/web.php
Koc: but why do we need this workaround with `current` and array?

nicolas-grekas: > we need this workaround with current

I had to do this because I needed to register the service as both a listener and a resolver, which is something that was not allowed by the way traceable resolvers were wired. I fixed this so I could remove the special definitions.

nicolas-grekas:
I updated the way traceable resolvers and serializers are wired so that we don't rely on service decoration, but use standard decoration instead. This preserves the definition of the original decorated service, so that we can e.g. use several tags on them without conflicts.

nicolas-grekas:
PR rebased on top of #50158 to split the concern of decoration apart.

nicolas-grekas:
PR rebased and ready.

derrabus (approved):

chalasr (approved):

chalasr:
Thanks @nicolas-grekas.

PR #50130, opened 2023-04-23
[Scheduler] Make `MessageGenerator` yield some scheduling context

### LINKED ISSUE #50085: [Scheduler] Expose the run date?

tucksaun:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #50085, "replaces" #50096?, could help with #49864
| License       | MIT
| Doc PR        | no doc yet for Scheduler

This PR introduces a new `MessageContext` that `MessageGenerator` should use as key when yielding messages to provide more context about the scheduled message.
The consumer of the scheduled messages can decide to use it or not.
The PR also adds the wiring to add this context to the `ScheduledStamp` when using Scheduler within Messenger.
This allows to add more context but without introducing a hard dependency on Messenger.

/cc @kbond 

Note: this PR is not adding the scheduler name (cf #49864) because I don't know exactly where to wire it yet but I don't see anything preventing it to be added to the context once we know where it should belong.

### DIFF
--- a/src/Symfony/Component/Scheduler/Generator/MessageContext.php
+++ b/src/Symfony/Component/Scheduler/Generator/MessageContext.php
@@ -1,0 +1,29 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Scheduler\Generator;
+
+use Symfony\Component\Scheduler\Trigger\TriggerInterface;
+
+/**
+ * @author Tugdual Saunier <tugdual@saunier.tech>
+ *
+ * @experimental
+ */
+final class MessageContext
+{
+    public function __construct(
+        public readonly TriggerInterface $trigger,
+        public readonly \DateTimeImmutable $triggeredAt,
+        public readonly \DateTimeImmutable|null $nextTriggerAt = null,
+    ) {
+    }
+}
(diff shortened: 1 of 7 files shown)


kbond (approved):
I like this much better than #50096!

fabpot (approved):

fabpot:
Thank you @tucksaun.

PR #49789, opened 2023-03-23
[Validator] New `PasswordStrength` constraint

Spomky:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Tickets       | none
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/18124

This PR adds a new constraint `PasswordStrength`. This constraint is able to determine if the password strength (or any other string) fulfils with the minimum requirement.
It leverages on [`bjeavons/zxcvbn-php`](https://github.com/bjeavons/zxcvbn-php) which is required when this constraint is used.

Example:

```php
<?php

declare(strict_types=1);

namespace App\Form;

use Symfony\Component\Form\AbstractType;
use Symfony\Component\Form\Extension\Core\Type\PasswordType;
use Symfony\Component\Form\Extension\Core\Type\RepeatedType;
use Symfony\Component\Form\FormBuilderInterface;
use Symfony\Component\OptionsResolver\OptionsResolver;
use Symfony\Component\Validator\Constraints\NotBlank;
use Symfony\Component\Validator\Constraints\PasswordStrength;

final class ChangePasswordFormType extends AbstractType
{
    public function buildForm(FormBuilderInterface $builder, array $options): void
    {
        $restrictedData = $options['restrictedData'] ?? [];

        $builder
            ->add('plainPassword', RepeatedType::class, [
                'type' => PasswordType::class,
                'options' => [
                    'attr' => [
                        'autocomplete' => 'new-password',
                    ],
                ],
                'first_options' => [
                    'constraints' => [
                        new NotBlank(),
                        new PasswordStrength(['restrictedData' => $restrictedData])
                    ],
                    'label' => 'New password',
                ],
                'second_options' => [
                    'label' => 'Repeat the new password',
                ],
                'mapped' => false,
            ])
        ;
    }

    public function configureOptions(OptionsResolver $resolver): void
    {
        $resolver->setDefaults([
            'restrictedData' => [],
        ])
            ->setAllowedTypes('restrictedData', 'string[]')
        ;
    }
}
```

Then from e.g. a controller

```php
$form = $this->createForm(ChangePasswordFormType::class, null, [
    'restrictedData' => [
        $user->getUsername(),
        $user->getEmail(),
        $user->getGivenName(),
        $user->getFamilyName(),
        'ApplicationName', // Arbitrary data
    ],
]);
```

It can be added as a property attribute:

```php
<?php

[...]

PR #50133, opened 2023-04-24
[Validator] Update the name of a password strength level

javiereguiluz:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

After checking the docs of the new PasswordStrength constraint, I propose to rename the `STRENGTH_REASONABLE` level name.

"Reasonable" is subjective. In a banking app, reasonable can be level 4 (very strong) and in a one-time password that expires in 60 seconds, reasonable can be level 0 (very weak).

I propose to rename it as a more neutral `STRENGTH_MEDIUM` name. Other common names for this are "Moderate" and "Fair".

### DIFF
--- a/src/Symfony/Component/Validator/Constraints/PasswordStrength.php
+++ b/src/Symfony/Component/Validator/Constraints/PasswordStrength.php
@@ -26,7 +26,7 @@ final class PasswordStrength extends Constraint
 {
     public const STRENGTH_VERY_WEAK = 0;
     public const STRENGTH_WEAK = 1;
-    public const STRENGTH_REASONABLE = 2;
+    public const STRENGTH_MEDIUM = 2;
     public const STRENGTH_STRONG = 3;
     public const STRENGTH_VERY_STRONG = 4;
 
@@ -42,7 +42,7 @@ final class PasswordStrength extends Constraint
 
     public function __construct(array $options = null, int $minScore = null, array $groups = null, mixed $payload = null)
     {
-        $options['minScore'] ??= self::STRENGTH_REASONABLE;
+        $options['minScore'] ??= self::STRENGTH_MEDIUM;
 
         parent::__construct($options, $groups, $payload);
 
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
This has been introduced by @Spomky in https://github.com/symfony/symfony/pull/49789. I'd like to have his opinion on this. Does "reasonable" come from any recommended source?

nicolas-grekas:
My personal opinion: "reasonable" is appropriate wording.
We take responsibility for defining what's a good tradeoff between strength and being annoying to end users according to external sources. Medium doesn't convey that's it's both good enough and annoying enough and people might over-zealously go for strong just because it feels more ... strong, without realizing the effect on UX.

javiereguiluz:
In the revision 3 of the  "NIST Special Publication 800-63B" (https://pages.nist.gov/800-63-3/sp800-63b.html) (last updates from 2020) they link to this paper when talking about password strength meters:

[Meters] de Carné de Carnavalet, Xavier and Mohammad Mannan. “From Very Weak to Very Strong: Analyzing Password-Strength Meters.” In Proceedings of the Network and Distributed System Security Symposium (NDSS), 2014. Available at: http://www.internetsociety.org/sites/default/files/06_3_1.pdf

In that paper we can see the strength scale used by some popular tech companies:

password-strength-levels

Nobody seems to be using "Reasonable" and instead they use "Medium", "Moderate" or "Fair".

Spomky:
Hi,

To be honest, when I came up with the word `reasonable`, I had in mind that it's hard to say how it might evolve in the future. It is also voluntary to use a vague term since in reality it is difficult to really define the strength of a password.
Nevertheless I agree with you and the term should indeed align with common practice in the field and a word like `Moderate` or `Medium` would be more appropriate.

javiereguiluz:
Thanks for your comment @Spomky.

What really made me think about this was the `minScore` option of the constraint, which can be described as _"Which is the reasonable password strength for your app?"_

If this "reasonable level" can be configured, then none of the predefined levels can be the reasonable one. I think all depends on the app using it. Thanks!

nicolas-grekas:
What about fair or good?

Spomky:
We already have very weak, weak, strong, very strong.
In the table, FedEx used the same plus Medium and this is the one proposed in this PR. It looks good to me

nicolas-grekas (approved):
Let's go, thanks for the contributions to the discussion.

nicolas-grekas:
Thank you @javiereguiluz.

PR #50136, opened 2023-04-24
[Notifier] [SpotHit] Support `smslong` and `smslongnbr` API parameters

camillebaronnet:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? |no 
| License       | MIT
| Doc PR        | mandatory for components ?

By default, SMS via Spot Hit are strictly limited to 160 characters maximum. This PR allows to forward the corresponding arguments from the DSN to the API request to unlock this capability

### DIFF
--- a/src/Symfony/Component/Notifier/Bridge/SpotHit/CHANGELOG.md
+++ b/src/Symfony/Component/Notifier/Bridge/SpotHit/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+ * Support `SMSLong` and `SMSLongNBr` API parameters
+
 6.2
 ---
 
--- a/src/Symfony/Component/Notifier/Bridge/SpotHit/README.md
+++ b/src/Symfony/Component/Notifier/Bridge/SpotHit/README.md
@@ -6,12 +6,14 @@ Provides [Spot-Hit](https://www.spot-hit.fr/) integration for Symfony Notifier.
 #### DSN example
 
 ```
-SPOTHIT_DSN=spothit://TOKEN@default?from=FROM
+SPOTHIT_DSN=spothit://TOKEN@default?from=FROM&smslong=SMSLONG&smslongnbr=SMSLONGNBR
 ```
 
 where:
  - `TOKEN` is your Spot-Hit API key
  - `FROM` is the custom sender (3-11 letters, default is a 5 digits phone number)
+ - `SMSLONG` (optional) 0 or 1 : allows SMS messages longer than 160 characters
+ - `SMSLONGNBR` (optional) integer : allows to check the size of the long SMS sent. You must send the number of concatenated SMS as a value. If our counter indicates a different number, your message will be rejected.
 
 Resources
 ---------
--- a/src/Symfony/Component/Notifier/Bridge/SpotHit/SpotHitTransport.php
+++ b/src/Symfony/Component/Notifier/Bridge/SpotHit/SpotHitTransport.php
@@ -35,6 +35,8 @@ final class SpotHitTransport extends AbstractTransport
 
     private string $token;
     private ?string $from;
+    private ?bool $smsLong = null;
+    private ?int $smsLongNBr = null;
 
     public function __construct(#[\SensitiveParameter] string $token, string $from = null, HttpClientInterface $client = null, EventDispatcherInterface $dispatcher = null)
     {
@@ -46,7 +48,27 @@ final class SpotHitTransport extends AbstractTransport
 
     public function __toString(): string
     {
-        return sprintf('spothit://%s%s', $this->getEndpoint(), null !== $this->from ? '?from='.$this->from : '');
+        $query = http_build_query([
+            'from' => $this->from,
+            'smslong' => $this->smsLong,
+            'smslongnbr' => $this->smsLongNBr,
+        ]);
+
+        return sprintf('spothit://%s', $this->getEndpoint()).('' !== $query ? '?'.$query : '');
+    }
+
+    public function setSmsLong(?bool $smsLong): self
+    {
+        $this->smsLong = $smsLong;
+
+        return $this;
+    }
+
+    public function setLongNBr(?int $smsLongNBr): self
+    {
+        $this->smsLongNBr = $smsLongNBr;
+
+        return $this;
     }
 
     public function supports(MessageInterface $message): bool
@@ -76,6 +98,8 @@ final class SpotHitTransport extends AbstractTransport
                 'type' => 'premium',
                 'message' => $message->getSubject(),
                 'expediteur' => $message->getFrom() ?: $this->from,
+                'smslong' => $this->smsLong,
+                'smslongnbr' => $this->smsLongNBr,
             ],
         ]);
 
--- a/src/Symfony/Component/Notifier/Bridge/SpotHit/SpotHitTransportFactory.php
+++ b/src/Symfony/Component/Notifier/Bridge/SpotHit/SpotHitTransportFactory.php
@@ -30,10 +30,12 @@ final class SpotHitTransportFactory extends AbstractTransportFactory
 
         $token = $this->getUser($dsn);
         $from = $dsn->getOption('from');
+        $smsLong = $dsn->getOption('smslong');
+        $smsLongNBr = $dsn->getOption('smslongnbr');
         $host = 'default' === $dsn->getHost() ? null : $dsn->getHost();
         $port = $dsn->getPort();
 
-        return (new SpotHitTransport($token, $from, $this->client, $this->dispatcher))->setHost($host)->setPort($port);
+        return (new SpotHitTransport($token, $from, $this->client, $this->dispatcher))->setHost($host)->setPort($port)->setSmsLong($smsLong)->setLongNBr($smsLongNBr);
     }
 
     protected function getSupportedSchemes(): array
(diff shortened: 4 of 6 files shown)


src/Symfony/Component/Notifier/Bridge/SpotHit/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.3
+---
+
+ * Support SMSLong and SMSLongNBr API parameters
```
OskarStark:
```suggestion
 * Support `SMSLong` and `SMSLongNBr` API parameters
```

src/Symfony/Component/Notifier/Bridge/SpotHit/SpotHitTransport.php
```diff
@@ -46,11 +48,25 @@ public function __construct(#[\SensitiveParameter] string $token, string $from =
...
-        return sprintf('spothit://%s?from=%s', $this->getEndpoint(), $this->from);
+    public function setSmsLong(?bool $smsLong): self
+    {
+        $this->smsLong = $smsLong;
+        return $this;
+    }
+
+    public function setLongNBr(?int $smsLongNBr): self
+    {
+        $this->smsLongNBr = $smsLongNBr;
```
OskarStark:
```suggestion
        $this->smsLong = $smsLong;
        
        return $this;
    }

    public function setLongNBr(?int $smsLongNBr): self
    {
        $this->smsLongNBr = $smsLongNBr;
        
```

src/Symfony/Component/Notifier/Bridge/SpotHit/SpotHitTransportFactory.php
```diff
@@ -30,10 +30,17 @@ public function create(Dsn $dsn): SpotHitTransport
...
+        $smsLongNBr = $dsn->getOption('smslongnbr');
         $host = 'default' === $dsn->getHost() ? null : $dsn->getHost();
         $port = $dsn->getPort();
 
-        return (new SpotHitTransport($token, $from, $this->client, $this->dispatcher))->setHost($host)->setPort($port);
+        return (new SpotHitTransport($token, $from, $this->client, $this->dispatcher))
+            ->setHost($host)
+            ->setPort($port)
+            ->setSmsLong($smsLong)
+            ->setLongNBr($smsLongNBr)
```
OskarStark: Lets keep it on one line

src/Symfony/Component/Notifier/Bridge/SpotHit/README.md
```diff
@@ -6,12 +6,14 @@ Provides [Spot-Hit](https://www.spot-hit.fr/) integration for Symfony Notifier.
...
 ```
-SPOTHIT_DSN=spothit://TOKEN@default?from=FROM
+SPOTHIT_DSN=spothit://TOKEN@default?from=FROM&smslong=SMSLONG&smslongnbr=SMSLONGNBR
 ```
 
 where:
  - `TOKEN` is your Spot-Hit API key
  - `FROM` is the custom sender (3-11 letters, default is a 5 digits phone number)
+ - `SMSLONG` (optional) 0 or 1 : allows SMS messages longer than 160 characters
+ - `SMSLONGNBR` (optional) Allows to check the size of the long SMS sent. You must send the number of concatenated SMS as a value. If our counter indicates a different number, your message will be rejected.
```
OskarStark: So this parameter defines if I allow to check? So this should be a bool, like the option before, right? And if I am right, how can I "send the number of concatenated SMS as a value"?

src/Symfony/Component/Notifier/Bridge/SpotHit/SpotHitTransport.php
```diff
@@ -82,6 +98,8 @@ protected function doSend(MessageInterface $message): SentMessage
                 'type' => 'premium',
                 'message' => $message->getSubject(),
                 'expediteur' => $from,
+                'smslong' => $this->smsLong,
+                'smslongnbr' => $this->smsLongNBr,
```
OskarStark, line 102: Is it ok to send null here?

src/Symfony/Component/Notifier/Bridge/SpotHit/Tests/SpotHitTransportTest.php
```diff
@@ -42,4 +47,61 @@ public static function unsupportedMessagesProvider(): iterable
...
+
+    public function argumentsProvider(): Generator
+    {
+        yield [
+            function(SpotHitTransport $transport) { $transport->setSmsLong(true); },
+            function(array $bodyArguments) { $this->assertSame('1', $bodyArguments['smslong']); },
+        ];
+
+        yield [
+            function(SpotHitTransport $transport) { $transport->setLongNBr(3); },
```
OskarStark: Please add a TestCase where we send both parameters

src/Symfony/Component/Notifier/Bridge/SpotHit/SpotHitTransport.php
camillebaronnet: Here "body" is an RFC1738 query like http_build_query() does. So null values are removed. This case is validated by the tests I added on the basic operation.

https://github.com/symfony/symfony/blob/0773491f79cf2d1555ac83fb5b90139c5d1db9fc/src/Symfony/Component/Notifier/Bridge/SpotHit/Tests/SpotHitTransportTest.php#L57

src/Symfony/Component/Notifier/Bridge/SpotHit/README.md
camillebaronnet: No it is directly an integer. "3" would mean that the SMS cannot exceed 3 SMS, otherwise the API throws an error.

The text is translated directly from the official documentation. I just added the expected type for more clarity

src/Symfony/Component/Notifier/Bridge/SpotHit/Tests/SpotHitTransportTest.php
camillebaronnet: Done, below

src/Symfony/Component/Notifier/Bridge/SpotHit/README.md
OskarStark: Ok, understand, my question is, if this should be configured at project level? I mean would you as a developer limit it to 3 messages all the time, or should this be set programmatically via the options for example on runtime?

SMSLONG is ok to define on project level IMHO, because you can allow or forbid this (regarding the costs e.g.)

camillebaronnet:
wow sorry, the rebase notified everyone

OskarStark:
> wow sorry, the rebase notified everyone

don't worry 😃

src/Symfony/Component/Notifier/Bridge/SpotHit/README.md
camillebaronnet: Ok that's a good point. Personally I use it at project level, as a guard. This option being very specific to this provider, I imagine that a need at runtime would not be done via this option but directly calculated in the code to be compatible whatever the provider

camillebaronnet: @OskarStark, you may not have seen my last answer :)

OskarStark: Maybe we should rename the options then?

SMSLONG -> LONG
SMSLONGNBR -> MAX 

it is more speakable, WDYT?

camillebaronnet: I agree, the name is not clear at all. I hesitated. I thought it was best to use the terms as requested and documented by SpotHit. Do you think it would be better to change this?

OskarStark: @fabpot WDYT?

src/Symfony/Component/Notifier/Bridge/SpotHit/SpotHitTransport.php
```diff
@@ -46,11 +48,27 @@ public function __construct(#[\SensitiveParameter] string $token, string $from =
...
-        if (!$this->from) {
-            return sprintf('spothit://%s', $this->getEndpoint());
-        }
+        $query = http_build_query([
+            'from' => $this->from,
+            'smslong' => $this->smsLong,
+            'smslongnbr' => $this->smsLongNBr,
+        ]);
+
+        return sprintf('spothit://%s', $this->getEndpoint()).(!empty($query) ? '?'.$query : '');
```
fabpot: We tend to never use `empty` in Symfony.

src/Symfony/Component/Notifier/Bridge/SpotHit/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.3
```
fabpot:
```suggestion
6.4
```

fabpot (changes requested):

src/Symfony/Component/Notifier/Bridge/SpotHit/README.md
fabpot: Let's keep the name used by the provider.

OskarStark (approved):

fabpot (approved):

fabpot:
Thank you @camillebaronnet.

PR #50137, opened 2023-04-24
[BrowserKit][Cache][Config][Console] Add missing return types

wouterj:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | Ref https://github.com/symfony/symfony/issues/47551
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Symfony/Component/Config/Definition/PrototypedArrayNode.php
+++ b/src/Symfony/Component/Config/Definition/PrototypedArrayNode.php
@@ -156,6 +156,8 @@ class PrototypedArrayNode extends ArrayNode
     /**
      * Disable adding concrete children for prototyped nodes.
      *
+     * @return never
+     *
      * @throws Exception
      */
     public function addChild(NodeInterface $node)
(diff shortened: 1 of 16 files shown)


nicolas-grekas:
(failures are related)

wouterj:
Should be fixed now. The remaining failures are unrelated

stof (approved):

src/Symfony/Component/Config/Definition/PrototypedArrayNode.php
```diff
@@ -156,6 +156,8 @@ public function getPrototype(): PrototypeNodeInterface
     /**
      * Disable adding concrete children for prototyped nodes.
      *
+     * @return never
```
nicolas-grekas, line 159: I think we should use `@return void` here, and maybe review where we added those and revert them
the reason is that if the class is truely extensible, then we have no reason to force child classes to also throw here

wouterj: We can surely challenge whether this should not be an `@final` class, but I think also without it we should force new implementations to not have children. It does not make sense for a prototyped node to contain concrete children.

nicolas-grekas: ok, works for me then :)

nicolas-grekas:
Thank you @wouterj.

src/Symfony/Component/Config/Definition/PrototypedArrayNode.php
stof: This should indeed be `never` for that one (it overrides the parent method in a similar way we do in FrozenParameterBag overriding ParameterBag methods)

PR #50138, opened 2023-04-24
[DependencyInjection][DomCrawler][ExpressionLanguage][Filesystem][Form] Add missing return types

wouterj:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | Ref https://github.com/symfony/symfony/issues/47551
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Symfony/Component/DependencyInjection/ParameterBag/FrozenParameterBag.php
+++ b/src/Symfony/Component/DependencyInjection/ParameterBag/FrozenParameterBag.php
@@ -34,26 +34,41 @@ class FrozenParameterBag extends ParameterBag
         $this->resolved = true;
     }
 
+    /**
+     * @return never
+     */
     public function clear()
     {
         throw new LogicException('Impossible to call clear() on a frozen ParameterBag.');
     }
 
+    /**
+     * @return never
+     */
     public function add(array $parameters)
     {
         throw new LogicException('Impossible to call add() on a frozen ParameterBag.');
     }
 
+    /**
+     * @return never
+     */
     public function set(string $name, array|bool|string|int|float|\UnitEnum|null $value)
     {
         throw new LogicException('Impossible to call set() on a frozen ParameterBag.');
     }
 
+    /**
+     * @return never
+     */
     public function deprecate(string $name, string $package, string $version, string $message = 'The parameter "%s" is deprecated.')
     {
         throw new LogicException('Impossible to call deprecate() on a frozen ParameterBag.');
     }
 
+    /**
+     * @return never
+     */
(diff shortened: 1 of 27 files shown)


nicolas-grekas:
(failures are related)

wouterj:
Related failures should be fixed now.

(not sure why Fabbot decided to start a war on me about 3 of the tens of added return PHPdocs)

stof (approved):

src/Symfony/Component/DependencyInjection/ParameterBag/FrozenParameterBag.php
```diff
@@ -34,26 +34,41 @@ public function __construct(
...
+    /**
+     * @return never
+     */
     public function deprecate(string $name, string $package, string $version, string $message = 'The parameter "%s" is deprecated.')
     {
         throw new LogicException('Impossible to call deprecate() on a frozen ParameterBag.');
     }
 
+    /**
+     * @return never
```
nicolas-grekas, line 70: here, I agree this is part of the contract of the FrozenParameterBag

src/Symfony/Component/Form/ButtonBuilder.php
```diff
@@ -324,6 +324,8 @@ public function setDataLocked(bool $locked): static
     /**
      * Unsupported method.
      *
+     * @return never
```
nicolas-grekas, line 367: we should also consider fixing the other methods that throw in this file

src/Symfony/Component/Form/Extension/Core/Type/MoneyType.php
```diff
@@ -104,7 +104,7 @@ public function getBlockPrefix(): string
      * The pattern contains the placeholder "{{ widget }}" where the HTML tag should
      * be inserted
      */
-    protected static function getPattern(?string $currency)
+    protected static function getPattern(?string $currency): string
```
nicolas-grekas: why not an annotation for this one?

wouterj: My bad, should be indeed.

src/Symfony/Component/Form/ButtonBuilder.php
wouterj: To be honest, I don't get why any of the methods throwing an exception in this file exist. This class extends nothing that requires them to be in this class.

nicolas-grekas:
Thank you @wouterj.

PR #50139, opened 2023-04-24
Check if trace.curlCommand is defined in profiler

Toflar:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? |no
| Tickets       | 
| License       | MIT
| Doc PR        | 

We'll get a `Key "curlCommand" for array with keys "method, url, options, content, http_code, info" does not exist.` `RuntimeError` otherwise, if this key does not exist :)


nicolas-grekas:
Thank you @Toflar.

PR #50143, opened 2023-04-24
[Console] trim(): Argument #1 () must be of type string, bool given

danepowell:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  |no 
| Deprecations? |no 
| Tickets       | Fix #... 
| License       | MIT

If you run a Symfony Console application in Git Bash, ANSICON is not set out of the box and results in this (suppressed) error:
> trim(): Argument ($string) must be of type string, bool given

AFAICT, this method is only called from Application::run() and errors are suppressed. So users will never actually see that error. Still, it could have side effects and breaks downstream projects: https://github.com/box-project/box/pull/990

### DIFF
--- a/src/Symfony/Component/Console/Terminal.php
+++ b/src/Symfony/Component/Console/Terminal.php
@@ -77,7 +77,8 @@ class Terminal
     private static function initDimensions()
     {
         if ('\\' === \DIRECTORY_SEPARATOR) {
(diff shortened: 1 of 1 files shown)


derrabus (approved):

stof (approved):

chalasr (approved):

chalasr:
Thank you @danepowell.

PR #48599, opened 2022-12-11
[VarDumper] Update types for dd()

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

Our beloved `dd()` function gets the PHP 8.1 treatment it deserves. 🚀

PR #50146, opened 2023-04-25
[VarDumper] Use native `never` return type for `dd` function

ruudk:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      |no
| New feature?  |no 
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

The native never return type was introduced in PHP 8.1 so we can use it.

https://www.php.net/manual/en/language.types.never.php


derrabus:
Duplicate of #48599

PR #50155, opened 2023-04-25
[Scheduler] add schedule name to `MessageContext`

kbond:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Alternative to #49864.

### DIFF
--- a/src/Symfony/Component/Scheduler/Generator/MessageContext.php
+++ b/src/Symfony/Component/Scheduler/Generator/MessageContext.php
@@ -21,9 +21,10 @@ use Symfony\Component\Scheduler\Trigger\TriggerInterface;
 final class MessageContext
 {
     public function __construct(
+        public readonly string $name,
         public readonly TriggerInterface $trigger,
         public readonly \DateTimeImmutable $triggeredAt,
-        public readonly \DateTimeImmutable|null $nextTriggerAt = null,
+        public readonly ?\DateTimeImmutable $nextTriggerAt = null,
     ) {
     }
 }
--- a/src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
+++ b/src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
@@ -27,14 +27,12 @@ final class MessageGenerator implements MessageGeneratorInterface
 
     public function __construct(
         private readonly Schedule $schedule,
-        string|CheckpointInterface $checkpoint,
+        private readonly string $name,
         private readonly ClockInterface $clock = new Clock(),
+        CheckpointInterface $checkpoint = null,
     ) {
         $this->waitUntil = new \DateTimeImmutable('@0');
-        if (\is_string($checkpoint)) {
-            $checkpoint = new Checkpoint('scheduler_checkpoint_'.$checkpoint, $this->schedule->getLock(), $this->schedule->getState());
-        }
-        $this->checkpoint = $checkpoint;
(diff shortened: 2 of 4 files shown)


src/Symfony/Component/Scheduler/Generator/MessageContext.php
```diff
@@ -21,6 +21,7 @@
 final class MessageContext
 {
     public function __construct(
+        public readonly ?string $name,
         public readonly TriggerInterface $trigger,
         public readonly \DateTimeImmutable $triggeredAt,
         public readonly \DateTimeImmutable|null $nextTriggerAt = null,
```
tucksaun: oops I let this slip (I can do a separate PR but that seem overkill):
```suggestion
        public readonly ?\DateTimeImmutable $nextTriggerAt = null,
```

src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
```diff
@@ -24,13 +24,16 @@ final class MessageGenerator implements MessageGeneratorInterface
...
         private readonly Schedule $schedule,
         string|CheckpointInterface $checkpoint,
         private readonly ClockInterface $clock = new Clock(),
+        string $name = null,
     ) {
         $this->waitUntil = new \DateTimeImmutable('@0');
+        $this->name = $name ?? (\is_string($checkpoint) ? $checkpoint : null);
         if (\is_string($checkpoint)) {
             $checkpoint = new Checkpoint('scheduler_checkpoint_'.$checkpoint, $this->schedule->getLock(), $this->schedule->getState());
         }
```
tucksaun: I'm wondering if it would make more sense to require a name and make the checkpoint optional?
If this is something we want, as the component is not yet released and is experimental we can change the signature easily. For example:

```suggestion
    public function __construct(
        private readonly Schedule $schedule,
        private readonly string $name,
        private readonly ClockInterface $clock = new Clock(),
        ?CheckpointInterface $checkpoint = null,
    )
    {
        $this->waitUntil = new \DateTimeImmutable('@0');
        $this->checkpoint = $checkpoint ?? new Checkpoint('scheduler_checkpoint_' . $this->name, $this->schedule->getLock(), $this->schedule->getState());
    }
```

tucksaun: (If so, it means the name will always be here and so we can require it for `MessageContext`)

kbond: I like this idea!

kbond:
Based on feedback from @tucksaun, I've re-arranged the `MessageGenerator::__construct()` parameters slightly to require a name and make the checkpoint optional.

tucksaun (approved):

fabpot (approved):

fabpot:
Thank you @kbond.

src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
```diff
@@ -27,14 +27,12 @@ final class MessageGenerator implements MessageGeneratorInterface
 
     public function __construct(
         private readonly Schedule $schedule,
-        string|CheckpointInterface $checkpoint,
+        private readonly string $name,
         private readonly ClockInterface $clock = new Clock(),
+        CheckpointInterface $checkpoint = null,
```
OskarStark, line 32:
```diff
- CheckpointInterface $checkpoint = null,
+ ?CheckpointInterface $checkpoint = null,
```

PR #50158, opened 2023-04-26
Fix registering traceable voters, argument resolvers and normalizers

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

By using service decoration to trace voters, resolvers and normalizers, we break the possibility to register them under several tags. We also make it harder to decorate them for end users. But adding observability to those doesn't require adding side effect to service definitions. Instead, we can and should use regular decoration.

### DIFF
--- a/src/Symfony/Component/HttpKernel/DependencyInjection/ControllerArgumentValueResolverPass.php
+++ b/src/Symfony/Component/HttpKernel/DependencyInjection/ControllerArgumentValueResolverPass.php
@@ -43,30 +43,30 @@ class ControllerArgumentValueResolverPass implements CompilerPassInterface
         $namedResolvers = $this->findAndSortTaggedServices(new TaggedIteratorArgument('controller.targeted_value_resolver', 'name', needsIndexes: true), $container);
         $resolvers = $this->findAndSortTaggedServices(new TaggedIteratorArgument('controller.argument_value_resolver', 'name', needsIndexes: true), $container);
 
-        foreach ($resolvers as $name => $resolverReference) {
-            $id = (string) $resolverReference;
-
-            if ($definitions[$id]->hasTag('controller.targeted_value_resolver')) {
+        foreach ($resolvers as $name => $resolver) {
+            if ($definitions[(string) $resolver]->hasTag('controller.targeted_value_resolver')) {
                 unset($resolvers[$name]);
             } else {
-                $namedResolvers[$name] ??= clone $resolverReference;
+                $namedResolvers[$name] ??= clone $resolver;
             }
         }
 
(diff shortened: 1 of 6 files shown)


src/Symfony/Component/HttpKernel/DependencyInjection/ControllerArgumentValueResolverPass.php
```diff
@@ -56,11 +56,20 @@ public function process(ContainerBuilder $container)
...
-            foreach ($resolvers + $namedResolvers as $resolverReference) {
+            $debugResolvers = [];
+            foreach ($resolvers as $resolverReference) {
                 $id = (string) $resolverReference;
-                $container->register("debug.$id", TraceableValueResolver::class)
-                    ->setDecoratedService($id)
-                    ->setArguments([new Reference("debug.$id.inner"), new Reference('debug.stopwatch')]);
+                $debugResolvers[$id] = $container->register('.debug.value_resolver.'.$id, TraceableValueResolver::class)
+                    ->setArguments([new Reference($id), new Reference('debug.stopwatch')])
+                    ->addTag('controller.argument_value_resolver');
```
stof: shouldn't we copy the tag attributes ?

stof: Wouldn't this break support for the `AsTaggedItem` attribute when resolving the TaggedIteratorArgument for those resolvers as it would look for the attribute on the TraceableValueResolver rather than on the actual resolver ?

stof: same in other compiler passes btw

stof (changes requested):

nicolas-grekas:
Thanks for the review @stof. You made me realize that we don't need to move the tags to the decorator. PR updated.

chalasr:
Could you double check that this does not break data collectors? Pretty sure it's ok for voters due to the way they're registered in debug mode but I don't know for normalizers and arg resolvers (thinking to some tagged locator referencing the original services while expecting to end up with the debug wrappers injected)

nicolas-grekas:
@chalasr I checked the code: the only way they're registered is via this pass. The updated tests should give you the hint to be confident this doesn't break the profiler: the injected refs are the debug ones, which have the stopwatch in them.

chalasr:
Thanks for confirming

chalasr (approved):

stof (approved):

PR #50161, opened 2023-04-26
Improve type annotation of ContainerInterface::get()

mathroc:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

Hi! When using Psalm we could improve the type annotations on `ContainerInterface::get() ` to make them understand better the return type, especially:

* can it return `null`?
* Do we know if it will return a specific type or if it could be an `object`

This shows how psalm understands the version of `ContainerInterface::get()` : https://psalm.dev/r/c9cb6f3655

I'm currently introducing this in a project in a Psalm stub, but thought it could be interesting to upstream this change. Is this something that could be integrated in Symfony? I'm not sure what's the policy is about Psalm/PHPStan here

note: I know that there are no guarantee that `$container->get(\DateTimeImmutable::class)` will return `DateTimeImmutable` and that a different service could have been registered with that name. If that's a blocker, we could at least implement the part saying if the method will return `null` or not depending on the second argument.

---

note about PHPStan: I couldn't make it work (yet), here is the playground https://phpstan.org/r/eb1465d2-79cf-4dd1-beb3-5929f565b4a8 I'll try again / ask for help if there's interest in this merge request

### DIFF
--- a/src/Symfony/Component/DependencyInjection/ContainerInterface.php
+++ b/src/Symfony/Component/DependencyInjection/ContainerInterface.php
@@ -36,6 +36,12 @@ interface ContainerInterface extends PsrContainerInterface
     public function set(string $id, ?object $service);
 
     /**
+     * @template B of self::*_REFERENCE
+     *
+     * @param B $invalidBehavior
+     *
+     * @psalm-return (B is self::EXCEPTION_ON_INVALID_REFERENCE|self::RUNTIME_EXCEPTION_ON_INVALID_REFERENCE ? object : object|null)
+     *
      * @throws ServiceCircularReferenceException When a circular reference is detected
      * @throws ServiceNotFoundException          When the service is not defined
      *


derrabus:
I believe that this should be handled by https://github.com/psalm/psalm-plugin-symfony instead.

stof:
there is strictly no guarantee that passing a class name to `ContainerInterface::get()` will return you an instance of that class. It depends entirely of the configuration of the container.

So this phpdoc is a big -1 as it lies to the static analyzer, making things worse than the existing type (it would make the SA tool consider some valid code as invalid and some invalid code as valid).

The plugin can instead inspect the actual configuration in your project to return more detailed info (I know that the phpstan symfony plugin does that, but I don't know what is currently implemented in the psalm one)

mathroc:
> there is strictly no guarantee that passing a class name to ContainerInterface::get() will return you an instance of that class. It depends entirely of the configuration of the container.

@stof yeah I mentioned that, that's exactly why I asked. and it's absolutely reasonable to not accept that, would a version that only take care of the nullable part be of interest?

Something like that:

```    
     * @template B of self::*
     * @param B $invalidBehavior
     * @psalm-return (
     *     B is self::EXCEPTION_ON_INVALID_REFERENCE|self::RUNTIME_EXCEPTION_ON_INVALID_REFERENCE
     *         ? object
     *         : object|null
     * )
```

About the plugin, I am unfortunately not in a position to enable it (because of https://github.com/vimeo/psalm/issues/2905)

wouterj:
I'm fine with adding the generic null conditional (this also is accepted per [our standards](https://symfony.com/doc/current/contributing/code/standards.html#documentation)).

I agree with the others about the FQCN resolving. Note that this also does not have to be transferred to the Symfony plugin, it can read the dumped XML file and use actual information about the return type of any service ID (including non-FQCNs).

derrabus:
> About the plugin, I am unfortunately not in a position to enable it (because of [vimeo/psalm#2905](https://github.com/vimeo/psalm/issues/2905))

Okay, but that should not influence the decision whether we merge a change or not. The plugin is the proper place for this logic and if it cannot be installed (for whatever reason), that's the issue that needs to be resolved.

> I'm fine with adding the generic null conditional

Me too.

mathroc:
> Okay, but that should not influence the decision whether we merge a change or not

absolutely :+1: that's just to explain why I was suggesting this

I'll update the merge request

mathroc:
done

ro0NL:
do you actually use ContainerInterface::get()? or is this a theoratical concern?

having this in PSR container sounds more useful, but it was rejected somehow: https://github.com/php-fig/container/pull/35

mathroc:
> do you actually use ContainerInterface::get()? or is this a theoratical concern?

yeah discovered about 400 usages in a project by making Psalm stricter :see_no_evil: 

> having this in PSR container sounds more useful, but it was rejected somehow: https://github.com/php-fig/container/pull/35

just looked at it but it was very different then both my first proposal and this last version

ro0NL:
symfony's container is a subset of PSR container

dealing with $invalidBehavior is a symfony specific concern, i agree

at this point im more concerned about 400 container usages in a project, regardless of phpdoc :sweat_smile:

src/Symfony/Component/DependencyInjection/ContainerInterface.php
```diff
@@ -41,12 +41,22 @@ public function set(string $id, ?object $service);
      * @param string $id              The service identifier
      * @param int    $invalidBehavior The behavior when the service does not exist
      *
+     * @see Reference
+     *
+     * @template B of self::*
```
nicolas-grekas:
```suggestion
     * @template B of self::*_REFERENCE
```

src/Symfony/Component/DependencyInjection/ContainerInterface.php
```diff
@@ -41,12 +41,22 @@ public function set(string $id, ?object $service);
...
+     *
+     * @param B $invalidBehavior
+     *
      * @return object|null
      *
      * @throws ServiceCircularReferenceException When a circular reference is detected
      * @throws ServiceNotFoundException          When the service is not defined
      *
-     * @see Reference
+     * @psalm-return (
```
nicolas-grekas: please put everything on one line

nicolas-grekas (approved):
Works for me after minor changes.

nicolas-grekas:
Thank you @mathroc.

PR #50163, opened 2023-04-26
[HttpClient] Add PSR-18 to HttplugClient since HTTPlug is deprecated

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Relates to https://github.com/php-http/httplug/pull/171

### DIFF
--- a/src/Symfony/Component/HttpClient/HttplugClient.php
+++ b/src/Symfony/Component/HttpClient/HttplugClient.php
@@ -26,6 +26,7 @@ use Http\Message\UriFactory;
 use Nyholm\Psr7\Factory\Psr17Factory;
 use Nyholm\Psr7\Request;
 use Nyholm\Psr7\Uri;
+use Psr\Http\Client\ClientInterface;
 use Psr\Http\Message\RequestFactoryInterface;
 use Psr\Http\Message\RequestInterface;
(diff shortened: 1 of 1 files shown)


derrabus:
Does this change do anything? We should already implement through `Http\Client\HttpClient`.

nicolas-grekas:
You should be right, but I still prefer adding this code now since it's going to stay in v7

stof (approved):

derrabus:
So the plan is to not implement the HTTPlug interface anymore in v7? Does the name `HttplugClient` still make sense then or should we rather add a `PsrClient` as a replacement?

stof:
@derrabus not the *sync* interface, as it is deprecated in favor of the PSR-18 one (the sync Httplug interface is already an empty one extending the PSR-18 one). the client will still implement the async interface (and the PSR-18 one).

stof:
and we *already* have a Psr18Client that implements only the PSR interfaces

derrabus:
Ah, got it, thanks.

derrabus (approved):

fabpot:
Thank you @nicolas-grekas.

PR #50176, opened 2023-04-27
[Mailer] Mark suppress bounces as DROPPED

### LINKED ISSUE #50164: [Webhook] Mailgun Webhook Converter - Special Mailgun code
### Description

Hello,

We rely on Mailgun to send some of our emails, and as such have developed a similar feature as the new Webhook component to handle their webhook calls.

In doing so, we found helpful to handle some Mailgun-specific error codes (6xx) for permanent fails.

Some might be relevant to handle in the default Symfony integration:
- 605 => Not delivering to a previously bounced address
- 606 => Not delivering to an unsubscribed address
- 607 => Not delivering to a user that already marked you as spam

These three error codes are linked to the Mailgun suppression listst, where they keep track of the users that already generated a hard bounce, unsubscribed or reported as spam.

In these cases, Mailgun blocks the email sending before it even reaches the email server.

In our opinion, these permanent fails might be better parsed to a `EmailDeliveryEvent::DROPPED` event than the `EmailDeliveryEvent::BOUNCE` that comes from simply parsing the event name & severity.

Might a PR on this topic be accepted?

### Example

_No response_

TristanPouliquen:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50164 
| License       | MIT
| Doc PR        | 

Mailgun sends out *permanent failures* webhooks in some cases when it does not even send the email to the recipient's server, eg when the email is in a suppression list.

This PR maps these codes to the `MailerDeliveryEvent::DROPPED` which seems more semantically correct as Mailgun has in fact dropped this email before sending it.

### DIFF
--- a/src/Symfony/Component/Mailer/Bridge/Mailgun/RemoteEvent/MailgunPayloadConverter.php
+++ b/src/Symfony/Component/Mailer/Bridge/Mailgun/RemoteEvent/MailgunPayloadConverter.php
@@ -19,6 +19,13 @@ use Symfony\Component\RemoteEvent\PayloadConverterInterface;
 
 final class MailgunPayloadConverter implements PayloadConverterInterface
 {
+    private const MAILGUN_SPECIFIC_DROPPED_CODES = [
+        605, // Not delivering to previously bounced address
+        606, // Not delivering to unsubscribed address
+        607, // Not delivering to a user who marked your messages as spam
+        625, // Poor mailing list quality
+    ];
+
     public function convert(array $payload): AbstractMailerEvent
     {
         if (\in_array($payload['event'], ['accepted', 'rejected', 'delivered', 'failed', 'blocked'], true)) {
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/Mailer/Bridge/Mailgun/RemoteEvent/MailgunPayloadConverter.php
```diff
@@ -19,6 +19,13 @@
 
 final class MailgunPayloadConverter implements PayloadConverterInterface
 {
+    public const MAILGUN_SPECIFIC_DROPPED_CODES = [
```
stof: this should be a private constant, not a public one

stof (changes requested):

src/Symfony/Component/Mailer/Bridge/Mailgun/RemoteEvent/MailgunPayloadConverter.php
TristanPouliquen: You are right, habit from our code where we reference it in the test class!

nicolas-grekas (approved):

nicolas-grekas:
Thank you @TristanPouliquen.

PR #50178, opened 2023-04-27
[Mailer] Catch Postmark inactive email exceptions

### LINKED ISSUE #50165: [Mailer] Sending to a blocked email address through Postmark
### Description

Hello,

When integrating Postmark to our stack to send emails, we noticed that if we send an email to an address that has been suppressed by Postmark (previous hard bounce, unsubscribe, spam...), the Postmark API throws an exception that is not catched by the Mailer component and comes from the PostmarkApiTransport through to the SendEmailHandler.

As of today, we decorated the SendEmailHandler to add a try/catch & dispatch a message as if we had received a webhook for a `EmailDeliveryEvent::DROPPED` event, but this seems hacky.

Is it an issue that would be interesting to solve in the Symfony code? If yes, do you have an idea on how to go at it in the best way?

### Example

_No response_

TristanPouliquen:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #50165 
| License       | MIT
| Doc PR        | 

When sending an email through Postmark to an email address that has been added to the suppression list, the API call returns an exception.

This PR wants to catch this exception and dispatch a message notifying of a `MailerDeliveryEvent::DROPPED` to simulate a webhook behaviour as for other providers such as Mailgun.

### DIFF
--- a/src/Symfony/Component/Mailer/Bridge/Postmark/Transport/PostmarkApiTransport.php
+++ b/src/Symfony/Component/Mailer/Bridge/Postmark/Transport/PostmarkApiTransport.php
@@ -20,7 +20,10 @@ use Symfony\Component\Mailer\Header\MetadataHeader;
 use Symfony\Component\Mailer\Header\TagHeader;
 use Symfony\Component\Mailer\SentMessage;
 use Symfony\Component\Mailer\Transport\AbstractApiTransport;
+use Symfony\Component\Messenger\MessageBusInterface;
 use Symfony\Component\Mime\Email;
+use Symfony\Component\RemoteEvent\Event\Mailer\MailerDeliveryEvent;
+use Symfony\Component\RemoteEvent\Messenger\ConsumeRemoteEventMessage;
 use Symfony\Contracts\HttpClient\Exception\DecodingExceptionInterface;
 use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
@@ -37,11 +40,11 @@ class PostmarkApiTransport extends AbstractApiTransport
 
     private $messageStream;
 
-    public function __construct(string $key, HttpClientInterface $client = null, EventDispatcherInterface $dispatcher = null, LoggerInterface $logger = null)
+    public function __construct(string $key, HttpClientInterface $client = null, EventDispatcherInterface $dispatcher = null, LoggerInterface $logger = null, MessageBusInterface $messageBus = null)
     {
         $this->key = $key;
 
-        parent::__construct($client, $dispatcher, $logger);
+        parent::__construct($client, $dispatcher, $logger, $messageBus);
     }
 
     public function __toString(): string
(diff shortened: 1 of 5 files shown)


src/Symfony/Component/Mailer/Bridge/Postmark/Transport/PostmarkApiTransport.php
```diff
@@ -51,13 +54,26 @@ public function __toString(): string
...
+                'json' => $this->getPayload($email, $envelope),
+            ]);
+        } catch (HttpTransportException $e) {
+            if (
+                str_contains($e->getMessage(), 'Unable to send an email: You tried to send to recipient(s) that have been marked as inactive.')
+                && null !== $this->messageBus
+                && class_exists(ConsumeRemoteEventMessage::class)
+            ) {
+                // Postmark throws an exception when sending an email to an inactive email address
+                $this->messageBus->dispatch(new ConsumeRemoteEventMessage('default', new MailerDeliveryEvent(MailerDeliveryEvent::DROPPED, 'MessageID', [])));
```
TristanPouliquen, line 72: I don't know how to correctly instantiate this object.

When looking at `WebhookController`, the type is parsed from the config file, so probably hard to access the right one here.

I can't either get the MessageID from the response, as my IDE warns me that it is potentially undefined, and I don't have a similar payload as I would have in a webhook call.

🤔

TristanPouliquen:
Hello, 

This is a first attempt at trying to catch the exception at the PostmarkApiTransport level & turning it into the corresponding `MailerDeliveryEvent` but I can't get my head around how to make it work properly.

Any input would be appreciated, or any other direction as this might not be the cleanest one.

src/Symfony/Component/Mailer/Transport/AbstractTransport.php
```diff
@@ -31,13 +32,15 @@
 abstract class AbstractTransport implements TransportInterface
 {
     private ?EventDispatcherInterface $dispatcher;
+    protected ?MessageBusInterface $messageBus;
```
stof, line 35: I suggest making it protected readonly, or to make it private with a protected getter like the logger. Child classes should not need to be able to modify it.

src/Symfony/Component/Mailer/Bridge/Postmark/Transport/PostmarkApiTransport.php
```diff
@@ -51,13 +54,26 @@ public function __toString(): string
...
-        ]);
+        try {
+            $response = $this->client->request('POST', 'https://'.$this->getEndpoint().'/email', [
+                'headers' => [
+                    'Accept' => 'application/json',
+                    'X-Postmark-Server-Token' => $this->key,
+                ],
+                'json' => $this->getPayload($email, $envelope),
+            ]);
+        } catch (HttpTransportException $e) {
```
stof, line 65: This catch won't ever happen, as HttpTransportException is thrown later in this method, in the check `if (200 !== $statusCode) {`

src/Symfony/Component/Mailer/Bridge/Postmark/Transport/PostmarkApiTransport.php
```diff
@@ -51,13 +54,26 @@ public function __toString(): string
...
+            if (
+                str_contains($e->getMessage(), 'Unable to send an email: You tried to send to recipient(s) that have been marked as inactive.')
```
stof, line 67: this should be detected based on the error code provided by the postmark response instead of detecting that based on the exact wording of the human-readable error message. See https://postmarkapp.com/developer/api/overview#error-codes

src/Symfony/Component/Mailer/Bridge/Postmark/Transport/PostmarkApiTransport.php
stof: if postmark refuses to send a message due to marking a recipient as inactive, there is probably no message id at all (as they haven't generated a message yet)

stof (changes requested):

fabpot:
@TristanPouliquen Do you have time to take comments into account?

fabpot:
Closing as there is no more activity here. Feel free to reopen when you have time.

PR #50184, opened 2023-04-28
[Dotenv] Add comment to `parse()` method

Elikill58:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| Deprecations? | no
| License       | MIT

Add a way to get parsed values for the given `.env` file. Help to parse env files without needing to change `$_ENV` variables.

### DIFF
--- a/src/Symfony/Component/Dotenv/Dotenv.php
+++ b/src/Symfony/Component/Dotenv/Dotenv.php
@@ -222,6 +222,8 @@ final class Dotenv
      * @param string $data The data to be parsed
      * @param string $path The original file name where data where stored (used for more meaningful error messages)
      *
+     * @return the array with all parsed data from env file
+     *
      * @throws FormatException when a file has a syntax error
      */
     public function parse(string $data, string $path = '.env'): array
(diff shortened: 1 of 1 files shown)


sstok:
FYI: When you use public methods to read the file (with the exception of `parse()`) the ENV variables are already set and overwritten, if you only need the values without setting `$_ENV` it's possible to use `parse()`.

Elikill58:
> FYI: When you use public methods to read the file (with the exception of `parse()`) the ENV variables are already set and overwritten, if you only need the values without setting `$_ENV` it's possible to use `parse()`.

Oh, when checking for this, the doc wasn't saying what were returned. After seeing this, my PR becomes quitely useless and should more be about return content.

sstok:
I checked the source code as that's the only source (no pun 😄 ) of truth.

Elikill58:
So, should I edit doc, close the PR or it's even a good PR?

GromNaN:
You can update the PR to modify the comment on `parse`, to something like "Parses the contents of a .env file and returns the variables it contains as a map."

Your branch can be reset on 6.4

nicolas-grekas (commented):
I'd suggest changing L220 instead of adding an annotation.
Or doing nothing :)

Elikill58:
> I'd suggest changing L220 instead of adding an annotation. Or doing nothing :)

`@return` annotation are not made for this?
Also yes, actually it were faster you add a line than I made my PR and ask for reviews xd I prefer to nothing than add a line into a comment while there is an annotation made for this

nicolas-grekas:
Adding the annotation would duplicate the information that's already in the title of the phpdoc. Duplication leads to increased maintenance.

Let's close then.

PR #50185, opened 2023-04-28
[HttpClient][HttpFoundation][HttpKernel][Ldap][Mailer][Mime][Process][Routing][Security][Serializer][Translation][Validator] Add missing return types

wouterj:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Tickets       | Ref #47551
| License       | MIT
| Doc PR        | -

Last one, for the components. Only some bundle methods are missing now.

### DIFF
--- a/src/Symfony/Component/Translation/Extractor/PhpExtractor.php
+++ b/src/Symfony/Component/Translation/Extractor/PhpExtractor.php
@@ -168,7 +168,7 @@ class PhpExtractor extends AbstractFileExtractor implements ExtractorInterface
     /**
      * Seeks to a non-whitespace token.
      */
-    private function seekToNextRelevantToken(\Iterator $tokenIterator)
+    private function seekToNextRelevantToken(\Iterator $tokenIterator): void
     {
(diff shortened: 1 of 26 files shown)


src/Symfony/Component/Translation/Extractor/PhpExtractor.php
```diff
@@ -262,7 +262,7 @@ private function getValue(\Iterator $tokenIterator): string
     /**
      * Extracts trans message from PHP tokens.
      */
-    protected function parseTokens(array $tokens, MessageCatalogue $catalog, string $filename)
+    protected function parseTokens(array $tokens, MessageCatalogue $catalog, string $filename): void
```
nicolas-grekas: BC break?

wouterj: Good spot, should be PHPdoc.

(guess I was put on a wrong foot by all the private methods in the file before this one)

nicolas-grekas:
Thank you @wouterj.

PR #49850, opened 2023-03-28
[FrameworkBundle] Fix auto-discovering validator constraints

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | #49397
| License       | MIT
| Doc PR        | -

Alternative to #49785

Note that this kind of container-assisted auto-discovery of non-service classes looks like an interesting pattern for many other subsystems. /cc @symfony/mergers @mtarld

PR #50186, opened 2023-04-28
[Translation] throw again if non-excluded abstract services are tagged

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix https://github.com/symfony/symfony/pull/49850#issuecomment-1488742207
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Symfony/Component/Translation/DependencyInjection/TranslatorPass.php
+++ b/src/Symfony/Component/Translation/DependencyInjection/TranslatorPass.php
@@ -56,12 +56,10 @@ class TranslatorPass implements CompilerPassInterface
             $constraintVisitorDefinition = $container->getDefinition('translation.extractor.visitor.constraint');
             $constraintClassNames = [];
 
-            foreach ($container->getDefinitions() as $definition) {
-                if (!$definition->hasTag('validator.constraint_validator')) {
-                    continue;
-                }
+            foreach ($container->findTaggedServiceIds('validator.constraint_validator', true) as $definition) {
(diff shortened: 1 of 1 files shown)


src/Symfony/Component/Translation/DependencyInjection/TranslatorPass.php
```diff
@@ -56,12 +56,10 @@ public function process(ContainerBuilder $container)
             $constraintVisitorDefinition = $container->getDefinition('translation.extractor.visitor.constraint');
             $constraintClassNames = [];
 
-            foreach ($container->getDefinitions() as $definition) {
-                if (!$definition->hasTag('validator.constraint_validator')) {
-                    continue;
-                }
+            foreach ($container->findTaggedServiceIds('validator.constraint_validator', true) as $definition) {
```
nicolas-grekas, line 59: `as $id`

nicolas-grekas: ` as $id => $attributes` actually

stof:
This is breaking the previous change, by not processing excluded services at all anymore, while they still needed to be processed for the discovery of available constraints.

This compiler was instentionally changed to process excluded service definitions (which are not returned anymore by `findTaggedServiceIds` in response to the comment you linked)

nicolas-grekas:
@stof is correct.

PR #47128, opened 2022-07-30
[Serializer] Throw InvalidArgumentException if the data needed in the constructor doesn't belong to a backedEnum 

alli83:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? |no
| Tickets       | Fix #46977 [Serializer] Collect deserialization errors on enum cause TypeError 
| License       | MIT
| Doc PR        | 

Targeted case: Deserialization with "collect_denormalization_errors" set but the data to be serialized does not correspond to the scalar value of the targeted enum instance in the object's constructor.
updated in last commit : Keep "from "(in the previous commits I had switched to "tryFrom") to map the data submitted to an enum instance targeted.

Imho, even if the "collect_denormalization_errors" is specified, this type of error seems to have to be caught upstream. By just trying to collect the error we do not prevent the object from being instantiated. But this action is bound to fail because we already know that the data is not valid (checked by the BackedEnumNormalizer). This instantiation 'failure' leads to a confusing/misleading error (i.e. _Argument # 1 ($test) must be of type TestEnum, string given_) which is not the original error.

PR #50192, opened 2023-04-28
[Serializer] backed enum throw notNormalizableValueException outside construct method

alli83:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | https://github.com/symfony/symfony/pull/47128#issuecomment-1527776830
| License       | MIT
| Doc PR        | 

PR https://github.com/symfony/symfony/pull/47128#issuecomment-1527776830 was aiming  to throw  an InvalidArgumentException if a constructor's argument is not a valid backedEnum( knowing an error will be thrown in all case e.g when it will be instantiated).

This PR aims to throw a NotNormalizableValueException  (like it was the case before) if an error occurs on an BackedEnum outside the case of the first PR

So we have two different types of exceptions depending on whether it's an error at the constructor level or not.

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/BackedEnumNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/BackedEnumNormalizer.php
@@ -64,7 +64,11 @@ final class BackedEnumNormalizer implements NormalizerInterface, DenormalizerInt
         try {
             return $type::from($data);
         } catch (\ValueError $e) {
-            throw new InvalidArgumentException('The data must belong to a backed enumeration of type '.$type);
+            if (isset($context['has_constructor'])) {
+                throw new InvalidArgumentException('The data must belong to a backed enumeration of type '.$type);
+            }
+
+            throw NotNormalizableValueException::createForUnexpectedDataType('The data must belong to a backed enumeration of type '.$type, $data, [$type], $context['deserialization_path'] ?? null, true, 0, $e);
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/Symfony/Component/Serializer/Normalizer/BackedEnumNormalizer.php
```diff
@@ -64,7 +64,11 @@ public function denormalize($data, string $type, string $format = null, array $c
         try {
             return $type::from($data);
         } catch (\ValueError $e) {
-            throw new InvalidArgumentException('The data must belong to a backed enumeration of type '.$type);
+            if (isset($context['has_constructor'])) {
+                throw new InvalidArgumentException('The data must belong to a backed enumeration of type '.$type);
+            }
+
+            throw new NotNormalizableValueException('The data must belong to a backed enumeration of type '.$type);
```
TerjeBr: Please add all the data that belongs in a NotNormalizableValueException

TerjeBr (changes requested):
Looks good, but missing all the data in the NotNormalizableValueException that makes it useful.

src/Symfony/Component/Serializer/Tests/SerializerTest.php
```diff
@@ -1230,7 +1231,47 @@ public function testCollectDenormalizationErrorsWithEnumConstructor()
...
+                new ObjectNormalizer($classMetadataFactory, null, null, $propertyInfoExtractor),
+            ],
+            ['json' => new JsonEncoder()]
+        );
+
+        try {
+            $serializer->deserialize('{"get": "POST"}', DummyObjectWithEnumProperty::class, 'json', [
+                 DenormalizerInterface::COLLECT_DENORMALIZATION_ERRORS => true,
+             ]);
+        } catch (\Throwable $th) {
```
nicolas-grekas:
```suggestion
        } catch (\Throwable $e) {
```

src/Symfony/Component/Serializer/Normalizer/BackedEnumNormalizer.php
TerjeBr:
```suggestion
            throw NotNormalizableValueException::createForUnexpectedDataType('The data must belong to a backed enumeration of type '.$type, $data, [$type], $context['deserialization_path'] ?? null, true, 0, $e);
```

TerjeBr (approved):
Looks good to me

chalasr (approved):

fabpot (approved):

fabpot:
Thank you @alli83.

PR #50193, opened 2023-04-29
[Serializer] Fix `SerializedPath` not working with constructor arguments

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Currently the `#[SerializedPath]` attribute doesn't work with constructor arguments:

```php
class Test
{
    public function __construct(
        #[SerializedPath('[foo][bar]')] public string $bar,
    ) {
    }
}

$serializer->denormalize(['foo' => ['bar' => 'something']], Test::class);
```
```
In AbstractNormalizer.php line 384:
                                                                                                                                    
  Cannot create an instance of "Test" from serialized data because its constructor requires parameter "bar" to be present.  
```

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
@@ -128,8 +128,8 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
 
         $this->propertyTypeExtractor = $propertyTypeExtractor;
 
-        if (null === $classDiscriminatorResolver && null !== $classMetadataFactory) {
-            $classDiscriminatorResolver = new ClassDiscriminatorFromClassMetadata($classMetadataFactory);
+        if ($classMetadataFactory) {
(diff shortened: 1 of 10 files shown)


nicolas-grekas:
Thank you @HypeMC.

src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
```diff
@@ -319,11 +308,9 @@ public function denormalize(mixed $data, string $type, string $format = null, ar
         $normalizedData = $this->prepareForDenormalization($data);
         $extraAttributes = [];
 
-        $reflectionClass = new \ReflectionClass($type);
-        $object = $this->instantiateObject($normalizedData, $type, $context, $reflectionClass, $allowedAttributes, $format);
-        $resolvedClass = $this->objectClassResolver ? ($this->objectClassResolver)($object) : $object::class;
+        $mappedClass = $this->getMappedClass($normalizedData, $type, $context);
```
soyuka, line 311: warning this breaks code where the doctrine discriminator has no property (ie exists only in the database)

PR #50197, opened 2023-04-29
Add disposition form-data to HeaderUtils

JoshuaBehrens:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT
| Doc PR        | symfony/symfony-docs#... will do later

I want to use the symfony request as PSR request. The PSR bridge is already helpful but it lacks to fill the server request body due to php://input being empty on multipart form data requests. So now I want to re-create the content of the request and build my own multipart message. I know the HeaderUtils being super useful for content disposition headers and I need disposition headers in the multipart message. So I tried it to make use of the HeaderUtils but they do not allow form-data as input ([although allowed looking at MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#as_a_header_for_a_multipart_body)). Therefore I changed the constraint to also allow form-data as disposition type.

Now as I wanted to write a test I also saw that you only make use of these utils in a file download, not in a file upload. So I am not even sure if you like having this support in this component. So shall I work further on this? Any specific test you expect as the context extends from only file download to also contain file upload?

I will add docs and changelog changes depending on the reaction of my suggestion.

TODO:
- Always add tests and ensure they pass.
- For new features, provide some code snippets to help understand usage.
- Changelog entry should follow https://symfony.com/doc/current/contributing/code/conventions.html#writing-a-changelog-entry

### DIFF
--- a/src/Symfony/Component/HttpFoundation/HeaderUtils.php
+++ b/src/Symfony/Component/HttpFoundation/HeaderUtils.php
@@ -20,6 +20,7 @@ class HeaderUtils
 {
     public const DISPOSITION_ATTACHMENT = 'attachment';
     public const DISPOSITION_INLINE = 'inline';
+    public const DISPOSITION_FORM_DATA = 'form-data';
 
     /**
      * This class should not be instantiated.
@@ -148,7 +149,7 @@ class HeaderUtils
     /**
      * Generates an HTTP Content-Disposition field-value.
      *
-     * @param string $disposition      One of "inline" or "attachment"
+     * @param string $disposition      One of "inline", "attachment" or "form-data"
      * @param string $filename         A unicode string
      * @param string $filenameFallback A string containing only ASCII characters that
      *                                 is semantically equivalent to $filename. If the filename is already ASCII,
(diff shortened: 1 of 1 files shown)


ro0NL:
> the header must also include a name parameter to identify the relevant field

but we can't specify any extra $params :thinking:

JoshuaBehrens:
Good point. So for this we have to add an additional nullable string parameter. I think this gets more bloated. But the header utils have checks and fallback behaviour one does not want to copy as well ... mhmm

nicolas-grekas:
Did you try using the Mime component for this part? It has tools to generate multipart mime messages.

JoshuaBehrens:
Thank you @nicolas-grekas for the hint. I just had a quick look into it. I admit not liking, that I can only use TextPart with FormDataPart. I currently building it using a temporary stream to reduce application memory. When I now have a file part I have to load it completely into RAM. But I admit I did not see that component. I will give it a try.

stof:
Generating a multipart body is indeed something we implement already in the Mime component, which seems a better fit for your use case than changing the signature of those HeaderUtils.

@JoshuaBehrens you can use a `DataPart` that does not load the content of the file (or use a TextPart directly with something else than a string for the body, but DataPart provides a nicer DX for the case of reading a file from the filesystem)

fabpot:
It looks like we can close this one, right?

PR #48821, opened 2022-12-29
[Serializer] add a context to allow invalid values in BackedEnumNormalizer

nikophil:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT
| Doc PR        | todo if PR gets merged

sometimes it is needed to allow deserialization of an enum to end up in a `null` value, this PR allows this behavior by passing a context

Issue #46977, opened 2022-07-18
[Serializer] Collect deserialization errors on enum cause TypeError


elliotbruneel:
### Symfony version(s) affected

6.1

### Description

When i want to collect denomalization errors i got and a TypeError instead

### How to reproduce

```php
class SerializerIssueReport extends TestCase
{
    public function testCollectErrorsOnEnum()
    {
        $serializer = new Serializer(
            [
                new BackedEnumNormalizer(),
                new ObjectNormalizer(),
            ],
            ['json' => new JsonEncoder()]
        );

        $serializer->deserialize('{"test": "invalid"}', DummyObject::class, 'json', [
            DenormalizerInterface::COLLECT_DENORMALIZATION_ERRORS => true
        ]);
        
        $this->expectException(PartialDenormalizationException::class);
    }
}


class DummyObject {
    public function __construct(
        public TestEnum $test,
    )
    {
    }
}

enum TestEnum: String
{
    case Test = "test";
}
```

Result: `TypeError: DummyObject::__construct(): Argument #1 ($test) must be of type TestEnum, string given`

chalasr:
Should be fixed by #47128. Confirmation welcome @elliotbruneel

elliotbruneel:
@chalasr That looks good to me. Thanks for the PR.

PR #43047, opened 2021-09-16
[Validator] Support enums in Choice constraints

bpolaszek:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      |no
| New feature?  | yes
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        | _TODO_

This PR allows using PHP 8.1 (backed) enums in `Choice` constraints.

Usage:

```php
enum Suit: string {
  case Hearts = 'H';
  case Diamonds = 'D';
  case Clubs = 'C';
  case Spades = 'S';
}
```

```php
class Card {
    public Suit $suit;
}
```

```php
class CardDTO {

    #[Choice(choices: Suit::class)]
    public string $suit;
}
```

PR #42502, opened 2021-08-12
[Serializer] Add support for collecting type error during denormalization

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #27824, Fix #42236, Fix #38472, Fix #37419 Fix #38968
| License       | MIT
| Doc PR        |

---

There is something that I don't like about the (de)Serializer. It's about the way it deals with typed properties.
As soon as you add a type to a property, the API can return 500.

Let's consider the following code:
```php
class MyDto
{
    public string $string;
    public int $int;
    public float $float;
    public bool $bool;
    public \DateTime $dateTime;
    public \DateTimeImmutable $dateTimeImmutable;
    public \DateTimeZone $dateTimeZone;
    public \SplFileInfo $splFileInfo;
    public Uuid $uuid;
    public array $array;
    /** @var MyDto[] */
    public array $collection;
}
```

and the following JSON:

```json
{
	"string": null,
	"int": null,
	"float": null,
	"bool": null,
	"dateTime": null,
	"dateTimeImmutable": null,
	"dateTimeZone": null,
	"splFileInfo": null,
	"uuid": null,
	"array": null,
	"collection": [
		{
			"string": "string"
		},
		{
			"string": null
		}
	]
}
```

**By default**, I got a 500:

<details>

image

</details>

It's the same with the prod environment. This is far from perfect when you try to make a public API :/ 
ATM, the only solution, is to remove all typehints and add assertions (validator component). With that, the public API is nice, but the internal PHP is not so good (PHP 7.4+ FTW!)

In APIP, they have support for transforming to [something](https://github.com/api-platform/core/blob/53837eee3ebdea861ffc1c9c7f052eecca114757/src/Core/Serializer/AbstractItemNormalizer.php#L233-L237) they can handle gracefully. But the deserialization stop on the first error (so the end user must fix the error, try again, fix the second error, try again etc.). And the raw exception message is leaked to the end user. So the API can return something like  `The type of the "string" attribute for class "App\Dto\MyDto" must be one of "string" ("null" given).`. Really not cool :/ 

So ATM, building a nice public API is not cool.

That's why I propose this PR that address all issues reported
* be able to collect all error
* with their property path associated
* don't leak anymore internal 

In order to not break the BC, I had to use some fancy code to make it work 🐒

[...]

PR #50201, opened 2023-05-01
[Serializer] Improve denormalization of backed enums

rodion-k:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | --
| License       | MIT
| Doc PR        | --

The goal is to improve error handling when denormalizing an enum.

Currently, if a string value is given when an integer is expected for the BackedEnum then such an error is thrown:
`TypeError: TestEnum::from(): Argument #1 ($value) must be of type int, string given`

This PR adds validation for such cases.

Also in case of an invalid enum case, the NotNormalizableValueException is thrown.

### DIFF
--- a/src/Symfony/Component/Serializer/CHANGELOG.md
+++ b/src/Symfony/Component/Serializer/CHANGELOG.md
@@ -21,6 +21,8 @@ CHANGELOG
    * `JsonSerializableNormalizer`
    * `ObjectNormalizer`
    * `PropertyNormalizer`
+ * Make `BackedEnumNormalizer` throw `NotNormalizableValueException` when the backed type doesn't match
+ * Make `BackedEnumNormalizer` throw `NotNormalizableValueException` when the value does not belong to a backed enumeration
 
 6.2
 ---
--- a/src/Symfony/Component/Serializer/Normalizer/BackedEnumNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/BackedEnumNormalizer.php
@@ -11,7 +11,6 @@
 
 namespace Symfony\Component\Serializer\Normalizer;
 
-use Symfony\Component\PropertyInfo\Type;
 use Symfony\Component\Serializer\Exception\InvalidArgumentException;
 use Symfony\Component\Serializer\Exception\NotNormalizableValueException;
 
@@ -69,14 +68,17 @@ final class BackedEnumNormalizer implements NormalizerInterface, DenormalizerInt
             }
         }
 
-        if (!\is_int($data) && !\is_string($data)) {
-            throw NotNormalizableValueException::createForUnexpectedDataType('The data is neither an integer nor a string, you should pass an integer or a string that can be parsed as an enumeration case of type '.$type.'.', $data, [Type::BUILTIN_TYPE_INT, Type::BUILTIN_TYPE_STRING], $context['deserialization_path'] ?? null, true);
+        $backingType = (new \ReflectionEnum($type))->getBackingType()->getName();
+        $givenType = \get_debug_type($data);
+
+        if ($givenType !== $backingType) {
+            throw NotNormalizableValueException::createForUnexpectedDataType(sprintf('Data expected to be "%s", "%s" given. You should pass a value that can be parsed as an enumeration case of type %s.', $backingType, $givenType, $type), $data, [$backingType], $context['deserialization_path'] ?? null, true);
         }
 
         try {
             return $type::from($data);
         } catch (\ValueError $e) {
-            throw new InvalidArgumentException('The data must belong to a backed enumeration of type '.$type);
+            throw NotNormalizableValueException::createForUnexpectedDataType('The data must belong to a backed enumeration of type '.$type, $data, [$backingType], $context['deserialization_path'] ?? null, true);
         }
     }
 
--- a/src/Symfony/Component/Serializer/Tests/Normalizer/BackedEnumNormalizerTest.php
+++ b/src/Symfony/Component/Serializer/Tests/Normalizer/BackedEnumNormalizerTest.php
@@ -86,9 +86,16 @@ class BackedEnumNormalizerTest extends TestCase
         $this->normalizer->denormalize(new \stdClass(), StringBackedEnumDummy::class);
     }
 
+    public function testDenormalizeBadBackingTypeThrowsException()
+    {
+        $this->expectException(NotNormalizableValueException::class);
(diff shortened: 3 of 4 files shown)


src/Symfony/Component/Serializer/Normalizer/BackedEnumNormalizer.php
```diff
@@ -69,14 +69,17 @@ public function denormalize(mixed $data, string $type, string $format = null, ar
             }
         }
 
-        if (!\is_int($data) && !\is_string($data)) {
-            throw NotNormalizableValueException::createForUnexpectedDataType('The data is neither an integer nor a string, you should pass an integer or a string that can be parsed as an enumeration case of type '.$type.'.', $data, [Type::BUILTIN_TYPE_INT, Type::BUILTIN_TYPE_STRING], $context['deserialization_path'] ?? null, true);
+        $backingType = (new ReflectionEnum($type))->getBackingType()->getName();
+        $givenType = get_debug_type($data);
```
nicolas-grekas:
```suggestion
        $givenType = \get_type($data);
```

src/Symfony/Component/Serializer/Normalizer/BackedEnumNormalizer.php
```diff
@@ -69,14 +69,17 @@ public function denormalize(mixed $data, string $type, string $format = null, ar
...
+
+        if ($givenType !== $backingType) {
+            throw NotNormalizableValueException::createForUnexpectedDataType(sprintf('Data expected to be "%s", "%s" given. You should pass a value that can be parsed as an enumeration case of type %s.', $backingType, $givenType, $type), $data, [$backingType], $context['deserialization_path'] ?? null, true);
         }
 
         try {
             return $type::from($data);
         } catch (\ValueError $e) {
-            throw new InvalidArgumentException('The data must belong to a backed enumeration of type '.$type);
+            throw NotNormalizableValueException::createForUnexpectedDataType('The data must belong to a backed enumeration of type '.$type, $data, [$backingType], $context['deserialization_path'] ?? null, true);
```
nicolas-grekas, line 81: this change basically reverts #47128
see #50192 also

src/Symfony/Component/Serializer/Normalizer/BackedEnumNormalizer.php
```diff
@@ -69,14 +69,17 @@ public function denormalize(mixed $data, string $type, string $format = null, ar
             }
         }
 
-        if (!\is_int($data) && !\is_string($data)) {
-            throw NotNormalizableValueException::createForUnexpectedDataType('The data is neither an integer nor a string, you should pass an integer or a string that can be parsed as an enumeration case of type '.$type.'.', $data, [Type::BUILTIN_TYPE_INT, Type::BUILTIN_TYPE_STRING], $context['deserialization_path'] ?? null, true);
+        $backingType = (new ReflectionEnum($type))->getBackingType()->getName();
```
nicolas-grekas:
```suggestion
        $backingType = (new \ReflectionEnum($type))->getBackingType()->getName();
```

and please remove the `use` on L14

src/Symfony/Component/Serializer/Normalizer/BackedEnumNormalizer.php
rodion-k: if you mean [gettype](https://www.php.net/manual/en/function.gettype.php) function, it returns "integer" and ReflectionNamedType::getName returns "int', so get_debug_type fits better

src/Symfony/Component/Serializer/Normalizer/BackedEnumNormalizer.php
rodion-k: https://github.com/symfony/symfony/pull/47128 was resolving https://github.com/symfony/symfony/issues/46977 issue which is covered by the `\Symfony\Component\Serializer\Tests\SerializerTest::testCollectDenormalizationErrorsWithWrongEnum` test and it works as expected. It doesn't throw TypeError but gives PartialDenormalizationException as expected.

mtarld: Agreed with @nicolas-grekas here, it seems to revert the previous work.
If you take a look at https://github.com/symfony/symfony/issues/46977 (for instance https://github.com/symfony/symfony/pull/47128#issuecomment-1433651496), and https://github.com/symfony/symfony/pull/48821, you'll see that the `InvalidArgumentException` is intentional.

TerjeBr: Throwing `InvalidArgumentException` instead of `NotNormalizableValueException` undoes the feature done by https://github.com/symfony/symfony/pull/42502

I first viewed https://github.com/symfony/symfony/pull/43047#issuecomment-921077295
Then I was happy that may be https://github.com/symfony/symfony/pull/40830 + https://github.com/symfony/symfony/pull/42502 would be a solution to my problem. But then https://github.com/symfony/symfony/pull/47128 undid that functionality again. It was partially reverted back to an `NotNormalizableValueException` in https://github.com/symfony/symfony/pull/50192

The whole point of having it as an `NotNormalizableValueException` is that you should be able to serialize DTOs with enums and then collect all the errors at the end in an `PartialDenormalizationException`. You should take care not to ruin that functionality.

src/Symfony/Component/Serializer/CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ CHANGELOG
    * `JsonSerializableNormalizer`
    * `ObjectNormalizer`
    * `PropertyNormalizer`
+ * Make `BackedEnumNormalizer` throw `NotNormalizableValueException` when the backed type doesn't match
+ * Make `BackedEnumNormalizer` throw `NotNormalizableValueException` when the value does not belong to a backed enumeration
```
nicolas-grekas, line 25:
```suggestion
 * Make `BackedEnumNormalizer` throw `NotNormalizableValueException` when the backed type doesn't match or when the value does not belong to a backed enumeration
```

BuechelFra:
@rodion-k are you planning to resolve the mentioned issues? Otherwise I'd also be happy to have a look 
(ping @nicolas-grekas)

BuechelFra:
From my point of view this issue is covered by [PR 62321](https://github.com/symfony/symfony/pull/62321)

PR #50203, opened 2023-05-02
[Messenger] Fix registering message handlers

### LINKED ISSUE #50202: [Messenger] `AsMessageHandler` attribute on methods do not work anymore

6.3.0-BETA1

### Description

Hey :wave: 

My friend @welcoMattic told me Symfony 6.3.0-BETA1 got out and I had to test if there is some regression issues and found one with Messenger component.

When you make a Handler, if you do:
```php
#[AsMessageHandler]
class MessageHandler
{
    public function __invoke(SimpleMessage $message): void
    {
        // do stuff
    }
}
```
You can use it freely and it will work like a charm.

But if you do:
```php
class MessageHandler
{
    #[AsMessageHandler]
    public function batch(BatchMessage $batchMessage): void
    {
        // batch
    }

    #[AsMessageHandler]
    public function simple(SimpleMessage $message): void
    {
        // simple
    }
}
```
During cache warmup you will have the following error:
image

### How to reproduce

See https://github.com/Korbeil/reproducer-63-messenger-handler/ for a simple reproducer.
And more specifically https://github.com/Korbeil/reproducer-63-messenger-handler/blob/main/src/Messenger/MessageHandler.php since it's where the issue is.

### Possible Solution

_No response_

### Additional Context

_No response_

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50202
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Symfony/Component/Messenger/DependencyInjection/MessengerPass.php
+++ b/src/Symfony/Component/Messenger/DependencyInjection/MessengerPass.php
@@ -143,7 +143,7 @@ class MessengerPass implements CompilerPassInterface
                     }
 
                     if ('__invoke' !== $method) {
-                        $wrapperDefinition = (new Definition('callable'))->addArgument([new Reference($serviceId), $method])->setFactory('Closure::fromCallable');
+                        $wrapperDefinition = (new Definition('Closure'))->addArgument([new Reference($serviceId), $method])->setFactory('Closure::fromCallable');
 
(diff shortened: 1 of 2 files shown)


chalasr (approved):
The exception message in LazyClosure needs a fix too for the `$id = null` case

PR #50206, opened 2023-05-02
[AssetMapper] Fix import map package parsing with an @ namespace

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | Not needed

This method is used with the `importmap:require` command - e.g.

```
./bin/console importmap:require @hotwired/stimulus
```

It parses the string into the package name and potential version (and something called a "registry" which is specific to jspm). The original regex didn't account for namespaced packages.

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
@@ -111,8 +111,8 @@ class ImportMapManager
      */
     public static function parsePackageName(string $packageName): ?array
     {
-        // https://regex101.com/r/58bl9L/1
-        $regex = '/(?:(?P<registry>[^:\n]+):)?(?P<package>[^@\n]+)(?:@(?P<version>[^\s\n]+))?/';
+        // https://regex101.com/r/d99BEc/1
+        $regex = '/(?:(?P<registry>[^:\n]+):)?((?P<package>@?[^@\n]+))(?:@(?P<version>[^\s\n]+))?/';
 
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Thank you @weaverryan.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -111,8 +111,8 @@ public function update(): array
      */
     public static function parsePackageName(string $packageName): ?array
     {
-        // https://regex101.com/r/58bl9L/1
-        $regex = '/(?:(?P<registry>[^:\n]+):)?(?P<package>[^@\n]+)(?:@(?P<version>[^\s\n]+))?/';
+        // https://regex101.com/r/d99BEc/1
```
stof, line 114: I would remove this comment entirely. Someone wanting to get the regex explanation on regex101.com can copy-paste the regex instead.

weaverryan: I got this idea from @dunglas - the nice thing about it is that the regex includes a bunch of examples. So if there is a problem in the future, you can quickly open this to see what the passing test cases are and then either validate that (A) the situation you are having trouble with SHOULD be working or (B) that your test case was never included.

dunglas: Exactly, the main benefit it that we can add tests for the regex.

PR #50212, opened 2023-05-02
[FrameworkBundle][Serializer] Add TranslatableNormalizer

Jean-Beru:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | maybe
| Tickets       | 
| License       | MIT
| Doc PR        | WIP

In 5.2, Symfony introduced translatable objects. This PR adds a dedicated normalizer to translate them. It can be useful for an API to return translated properties.

Also I'm wondering if it could be considered as a BC break since translatable objects were handled by the ObjectNormalizer which currently returns a plain object.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
@@ -75,6 +75,7 @@ use Symfony\Component\Serializer\Normalizer\DateTimeNormalizer;
 use Symfony\Component\Serializer\Normalizer\FormErrorNormalizer;
 use Symfony\Component\Serializer\Normalizer\JsonSerializableNormalizer;
 use Symfony\Component\Serializer\Normalizer\ObjectNormalizer;
+use Symfony\Component\Serializer\Normalizer\TranslatableNormalizer;
 use Symfony\Component\Serializer\Serializer;
 use Symfony\Component\Translation\DependencyInjection\TranslatorPass;
 use Symfony\Component\Translation\LocaleSwitcher;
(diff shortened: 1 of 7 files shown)


src/Symfony/Component/Serializer/Normalizer/TranslatableNormalizer.php
```diff
@@ -0,0 +1,54 @@
...
+{
+    public const NORMALIZATION_LOCALE_KEY = 'translatable_normalization_locale';
+
+    private array $defaultContext = [
+        self::NORMALIZATION_LOCALE_KEY => null,
+    ];
+
+    public function __construct(
+        array $defaultContext = [],
+        private readonly ?TranslatorInterface $translator = null,
```
dunglas: Shouldn't we force this property to be non-null and remove the related to this class if the translator isn't available?

src/Symfony/Component/Serializer/Normalizer/TranslatableNormalizer.php
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public function supportsNormalization(mixed $data, string $format = null, array $context = []): bool
+    {
+        return null !== $this->translator && $data instanceof TranslatableInterface;
+    }
+
+    public function getSupportedTypes(?string $format): array
+    {
+        return [TranslatableInterface::class => true];
```
dunglas, line 52: With the current implementation, this method should return an empty array of no translator is injected. But see my other comment.

dunglas (changes requested):

src/Symfony/Component/Serializer/Normalizer/TranslatableNormalizer.php
Jean-Beru: Good catch, especially since this normalizer is useless without the Translator

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
```diff
@@ -1586,6 +1587,18 @@ public function testConstraintViolationListNormalizerRegistered()
...
     }
 
+    public function testTranslatableNormalizerRegistered()
+    {
+        $container = $this->createContainerFromFile('full');
+
+        $definition = $container->getDefinition('serializer.normalizer.translatable');
+        $tag = $definition->getTag('serializer.normalizer');
+
+        $this->assertEquals(TranslatableNormalizer::class, $definition->getClass());
```
dunglas: `assertSame` ?

dunglas (approved):

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
Jean-Beru: :+1:

fabpot (approved):

fabpot:
Thank you @Jean-Beru.

PR #48037, opened 2022-10-28
[WebProfilerBundle] Add missing Javascript function to prevent undefined function

PhilETaylor:
| Q             | A
| ------------- | ---
| Branch?       | 5.4+
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

So.... To replicate create a demo app with Symfony 5.4, 6.1, 6.2....

```
composer create-project symfony/symfony-demo mySymfony54 1.8.0
cd mySymfony54
symfony serve
# Visit http://127.0.0.1:8000/en/blog/ to confirm its working
# Edit BlogController.php and throw an \Exception('test') in the index function
```

The result is a Javascript error in the console log:

```
TypeError: Sfjs.createFilters is not a function. (In 'Sfjs.createFilters()', 'Sfjs.createFilters' is undefined)
```

Screenshot 2022-10-28 at 23 32 20

Now, Javascript in the profiler is loaded from two places (and each of these files has a header alerting you to that fact and that changes in one should be considered for the other too) 

```
Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/base_js.html.twig (has no Sfjs.createFilters function)
Symfony/Component/ErrorHandler/Resources/assets/js/exception.js (Has a Sfjs.createFilters function) 
````

Both files will only run their definition of the `Sfjs` if the `typeof Sfjs === 'undefined'` is true... so if one loads before the other then the latter will not redefine the `Sfjs` 

However, what is happening is that the `base_js.html.twig` is defining the `Sfjs` object before `exception.js` gets a chance - and because the `Sfjs.createFilters` function is not in `base_js.html.twig` its not defined and so the error message tells us that.

This PR duplicates the method into the `base_js.html.twig` version of the`Sfjs` object so that it can be called from the code (at the bottom of) `exception.js` and thus the console.log error message is now gone. 

The code provided in this PR is not my code, its direct copy of the code from `exception.js`

PR #50213, opened 2023-05-02
[ErrorHandler] Prevent conflicts with WebProfilerBundle’s JavaScript

### LINKED ISSUE #50108: [ErrorHandler][WebProfilerBundle] Logs filter no longer appears on exception pages

≥ 5.4

### Description

Since #42195 the UI of logs changed between the profiler and the exception page so `createFilters` was removed from `base_js.html.twig`’s `Sfjs`.

`exception.js` also defines its own `Sfjs` with a `createFilters` method it calls on `DOMContentLoaded`.

Problem is, `base_js.html.twig`’s `Sfjs` overrides `exception.js`’ one because the latter does not define `loadToolbar`.

As a result, when `DOMContentLoaded` kicks in, `Sfjs.createFilters` is called while it does not exist anymore so an error is logged and filters don’t work.

> Uncaught TypeError: Sfjs.createFilters is not a function

Else, the UI would look like this on exception pages:

### How to reproduce

Display the `ErrorHandler`’s exception page with some logs and the profiler’s toolbar enabled.

### Possible Solution

Not sure about this. The fact the web profiler and error handler can override each other’s `Sfjs` seems weird.

---

Also, remnants of `Sfjs.createFilters` are still present in `translation.html.twig`.

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50108
| License       | MIT
| Doc PR        | N/A

ErrorHandler’s `exception.js` and WebProfilerBundle’s `base_js.html.twig` both expose a global `Sfjs` variable, which create conflicts (see linked issue e.g.).

As the exception page
- does not make use of `Sfjs`
- is not extensible

I updated `exception.js` to not create any global scope variable. I also removed `DOMContentLoaded` listeners as the script is run inline at the end of the document.

This PR stems from #48037.
Hiding whitespaces will produce a much more readable diff.

### DIFF
--- a/src/Symfony/Component/ErrorHandler/Resources/assets/js/exception.js
+++ b/src/Symfony/Component/ErrorHandler/Resources/assets/js/exception.js
@@ -1,297 +1,285 @@
 /* This file is based on WebProfilerBundle/Resources/views/Profiler/base_js.html.twig.
    If you make any change in this file, verify the same change is needed in the other file. */
 /*<![CDATA[*/
-if (typeof Sfjs === 'undefined') {
-    Sfjs = (function() {
-        "use strict";
+(function() {
+    "use strict";
 
-        if ('classList' in document.documentElement) {
-            var hasClass = function (el, cssClass) { return el.classList.contains(cssClass); };
-            var removeClass = function(el, cssClass) { el.classList.remove(cssClass); };
-            var addClass = function(el, cssClass) { el.classList.add(cssClass); };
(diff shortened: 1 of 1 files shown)


yceruto (approved):
LGMT, thanks!

fabpot:
Thank you @MatTheCat.

PR #42195, opened 2021-07-19
[WebProfilerBundle] Redesigned the log section

javiereguiluz:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #42155
| License       | MIT
| Doc PR        | -

This PR is not fully polished, but it's ready for review, both code/design and UX/usability. The main missing feature is dark mode styles.

### Before

The logs were divided in sections (that's the main issue reported in #42155). If the app had errors, we displayed errors section first:

before-error

If there are no errors but there are deprecations, we display that section:

before-deprecation

Abut filtering data, it depends on each section. For example, some allow to filter by log channel:

before-filter-channel

And others allow to filter by priority:

before-filter-priority

### After

This PR proposes to display a single table with all the logs (and separate the "container compilation" logs because they are a bit different than the others):

after-logs

**Question**: log timestamps now display milliseconds. Do you like this or is it enough with seconds?

Deprecations and errors are now more highlighted on each row (this needs a bit more polishing):

after-deprecation-error

Filters are now super dynamic and always available:

after-filters

Container logs are still available at the bottom of the page:

after-container

So, what do you think?

PR #50214, opened 2023-05-02
[WebProfilerBundle] Remove legacy filters remnants

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

#42195 introduced new filters for the WebProfilerBundle but it still contains some remnants, for example in the translation panel:

> Uncaught TypeError: Sfjs.createFilters is not a function

This PR removes them all.

### DIFF
--- a/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/translation.html.twig
+++ b/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/translation.html.twig
@@ -155,22 +155,20 @@
             </div>
         </div>
 
-        <script>Sfjs.createFilters();</script>
-
         {% endblock messages %}
     {% endif %}
 
 {% endblock %}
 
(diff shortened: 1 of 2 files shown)


javiereguiluz (approved):
👍 thanks Mathieu!

The ideal solution would be to recreate these filters in the new system instead of just removing the old filters which no longer work.

I've added this to my TODO list to see if I can do it for Symfony 6.4/7.0.

Thanks!

nicolas-grekas:
Thank you @MatTheCat.

MatTheCat:
@javiereguiluz yes having the new filters everywhere (on the exception page as well; see #50213) would be nice!

nicolas-grekas:
I merged this up to 6.3

PR #50198, opened 2023-04-30
[HttpKernel] Forward `_stateless` attribute to inline rendered fragments

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Prerequisite for #50037
| License       | MIT
| Doc PR        | N/A

For now inline rendered fragments have no mean of checking if [the `_stateless` attribute](https://symfony.com/blog/new-in-symfony-5-1-routing-improvements#added-stateless-route-attribute) has been set so they cannot adapt accordingly. This PR would allow them not to use the session if `$request->attributes->getBoolean('_stateless')` is true.

This could be used in `ProfilerController`’s `searchBarAction` eg.

PR #50218, opened 2023-05-02
Profiler respect stateless attribute

### LINKED ISSUE #50037: [WebProfilerBundle] Profiler is not stateless

`5.4.21` `6.2.8`

### Description

I have a SPA and I'm importing all routes in stateless mode

I'm having the error `Session was used while the request was declared stateless` beacause Profier is not stateless

### How to reproduce

Edit Kernel.php to be stateless
```php
        $routes->import($confDir.'/{routes}/*'.self::CONFIG_EXTS, 'glob')->prefix('/api/')->stateless();
        $routes->import($confDir.'/{routes}/'.$this->environment.'/**/*'.self::CONFIG_EXTS, 'glob')->prefix('/api/')->stateless();
        $routes->import($confDir.'/{routes}'.self::CONFIG_EXTS, 'glob')->prefix('/api/')->stateless();
```

### Possible Solution

_No response_

### Additional Context

_No response_

alamirault:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #50037
| License       | MIT

Now that https://github.com/symfony/symfony/pull/50198 forward `_stateless` attribute in fragments, this PR check this attribute and make profiler stateless when is required.

I don't know which tests I have to add because session is not checked. (I checked on reproducer and it's ok)

### DIFF
--- a/src/Symfony/Bundle/WebProfilerBundle/Controller/ProfilerController.php
+++ b/src/Symfony/Bundle/WebProfilerBundle/Controller/ProfilerController.php
@@ -128,7 +128,9 @@ class ProfilerController
             throw new NotFoundHttpException('The profiler must be enabled.');
         }
 
-        if ($request->hasSession() && ($session = $request->getSession())->isStarted() && $session->getFlashBag() instanceof AutoExpireFlashBag) {
+        if (!$request->attributes->getBoolean('_stateless') && $request->hasSession()
+            && ($session = $request->getSession())->isStarted() && $session->getFlashBag() instanceof AutoExpireFlashBag
+        ) {
             // keep current flashes for one more request if using AutoExpireFlashBag
             $session->getFlashBag()->setAll($session->getFlashBag()->peekAll());
         }
@@ -172,7 +174,11 @@ class ProfilerController
 
         $this->cspHandler?->disableCsp();
 
-        $session = $request->hasSession() ? $request->getSession() : null;
+
+        $session = null;
+        if ($request->attributes->getBoolean('_stateless') && $request->hasSession()) {
+            $session = $request->getSession();
+        }
 
         return new Response(
             $this->twig->render('@WebProfiler/Profiler/search.html.twig', [
@@ -247,7 +253,7 @@ class ProfilerController
         $limit = $request->query->get('limit');
         $token = $request->query->get('token');
 
-        if ($request->hasSession()) {
+        if (!$request->attributes->getBoolean('_stateless') && $request->hasSession()) {
             $session = $request->getSession();
 
             $session->set('_profiler_search_ip', $ip);


MatTheCat:
#50198 was a new feature so it is only available on the 6.3 branch :sweat_smile: 

I looked for a way to test this and it seems we have to update `WebProfilerBundleKernel` to allow configuring the debug mode.

src/Symfony/Bundle/WebProfilerBundle/Controller/ProfilerController.php
```diff
@@ -176,7 +176,7 @@ public function searchBarAction(Request $request): Response
             $this->cspHandler->disableCsp();
         }
 
-        if (!$request->hasSession()) {
+        if (true === $request->attributes->getBoolean('_stateless') || !$request->hasSession()) {
```
stof: `true === ` should be removed. There is no need to turn a boolean into another (same) boolean

stof: Actually, this code looks broken to me. When we don't have a session, we should still support reading the query string.

src/Symfony/Bundle/WebProfilerBundle/Controller/ProfilerController.php
```diff
@@ -272,7 +272,7 @@ public function searchAction(Request $request): Response
         $limit = $request->query->get('limit');
         $token = $request->query->get('token');
 
-        if ($request->hasSession()) {
+        if (false === $request->attributes->getBoolean('_stateless') && $request->hasSession()) {
```
stof:
```suggestion
        if (!$request->attributes->getBoolean('_stateless') && $request->hasSession()) {
```

stof (commented):
This should be done in 6.3 IMO as it relies on the previous PR that is merged in 6.3

stof (approved):

fabpot (approved):

fabpot:
Thank you @alamirault.

bigfoot90:
Didn't work if firewall is stateless.

**security.yaml**
```yml
security:
  ---
  firewalls:
    dev:
      pattern: ^/(_(profiler|wdt))/
      security: false

    api:
      pattern:    ^/api
#      stateless:  true
      http_basic: ~
```

src/Symfony/Bundle/WebProfilerBundle/Controller/ProfilerController.php
```diff
@@ -172,7 +174,11 @@ public function searchBarAction(Request $request): Response
 
         $this->cspHandler?->disableCsp();
 
-        $session = $request->hasSession() ? $request->getSession() : null;
+
+        $session = null;
+        if ($request->attributes->getBoolean('_stateless') && $request->hasSession()) {
```
themasch, line 179: I know I am late, but I thought posting it here was less interruptive then creating a new issue or discussion for this question: 

The PR touches 3 placed where it gets the session and adds checks for the `_stateless` request attribute. 
Two checks check for `!stateless && hasSession` while one of them (in `searchBarAction`)  checks for `stateless && hasSession` (without the `!`). 

I found this while debugging for what starts `Session was used while the request was declared stateless.` when calling the profiler. As far as I can tell, the only place using the session there is the this spot. 

This line looks incorret to me, I think its missing the `!` in front.  But since I'm probably wrong (since this is like this for a year now, with no one noticing) I'd like to learn why I am wrong!

MatTheCat: Just tested and you’re not wrong. Maybe you really are the first to notice :sweat_smile:

themasch: If I understand the implications of this correct, then that means that for users that have `stateless: false`, restoring the search bar state from session should not work, as it would not use the session in those cases. It would only use the session if `stateless: true`  (e.g. when there should be no session). Correct?

MatTheCat: > for users that have `stateless: false`, restoring the search bar state from session should not work

Yes, that also includes users who did not set `_stateless`. Are you up for a PR? Looks like you already know the fix!

themasch: I can give it a try: https://github.com/symfony/symfony/pull/57679 
Hope thats correct so far.

PR #50219, opened 2023-05-02
[AssetMapper] Adding debug:assetmap command + normalize paths

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | Docs are TODO already for the component

For UX, those packages will add their own namespaced paths to the "asset mapper" to make their Stimulus controllers available. So, between your own paths and automatically-added paths, having a way to visualize everything seems helpful.

Screenshot 2023-05-02 at 4 15 49 PM

**ALSO**

This PR normalizes the paths in asset mapper in 2 ways:

A) Use `/` always for logical path - e.g. you would use `asset('images/foo.png')` everywhere... you wouldn't say `asset('images\\foo.png')` on Windows
B) All absolute paths are converted with `realpath` internally. We do some "directory comparisons" (e.g. is `/var/foo` inside of `/var/foo/bar`) so having real paths consistently everywhere keeps this working.

Tests added for all of these changes

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/AssetMapperRepository.php
+++ b/src/Symfony/Component/AssetMapper/AssetMapperRepository.php
@@ -55,7 +55,7 @@ class AssetMapperRepository
 
             $file = rtrim($path, '/').'/'.$localLogicalPath;
             if (file_exists($file)) {
-                return $file;
+                return realpath($file);
             }
         }
 
@@ -64,17 +64,24 @@ class AssetMapperRepository
 
     public function findLogicalPath(string $filesystemPath): ?string
     {
+        if (!is_file($filesystemPath)) {
+            return null;
+        }
+
+        $filesystemPath = realpath($filesystemPath);
+
         foreach ($this->getDirectories() as $path => $namespace) {
             if (!str_starts_with($filesystemPath, $path)) {
                 continue;
             }
 
             $logicalPath = substr($filesystemPath, \strlen($path));
+
             if ('' !== $namespace) {
-                $logicalPath = $namespace.'/'.$logicalPath;
+                $logicalPath = $namespace.'/'.ltrim($logicalPath, '/\\');
             }
 
-            return ltrim($logicalPath, '/');
+            return $this->normalizeLogicalPath($logicalPath);
         }
 
         return null;
@@ -100,11 +107,20 @@ class AssetMapperRepository
                 /** @var RecursiveDirectoryIterator $innerIterator */
                 $innerIterator = $iterator->getInnerIterator();
                 $logicalPath = ($namespace ? rtrim($namespace, '/').'/' : '').$innerIterator->getSubPathName();
+                $logicalPath = $this->normalizeLogicalPath($logicalPath);
                 $paths[$logicalPath] = $file->getPathname();
             }
         }
 
         return $paths;
+    }
+
+    /**
+     * @internal
+     */
+    public function allDirectories(): array
+    {
+        return $this->getDirectories();
     }
 
     private function getDirectories(): array
@@ -120,13 +136,13 @@ class AssetMapperRepository
                 if (!file_exists($path)) {
                     throw new \InvalidArgumentException(sprintf('The asset mapper directory "%s" does not exist.', $path));
                 }
-                $this->absolutePaths[$path] = $namespace;
+                $this->absolutePaths[realpath($path)] = $namespace;
 
                 continue;
             }
 
             if (file_exists($this->projectRootDir.'/'.$path)) {
-                $this->absolutePaths[$this->projectRootDir.'/'.$path] = $namespace;
+                $this->absolutePaths[realpath($this->projectRootDir.'/'.$path)] = $namespace;
 
                 continue;
             }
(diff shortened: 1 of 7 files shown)


src/Symfony/Component/AssetMapper/Command/DebugAssetMapperCommand.php
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+
+        $allAssets = $this->assetMapper->allAssets();
+
+        $pathRows = [];
+        foreach ($this->assetMapperRepository->allDirectories() as $path => $namespace) {
```
chalasr, line 60:
```suggestion
        foreach ($this->assetMapperRepository->getDirectories() as $path => $namespace) {
```

src/Symfony/Component/AssetMapper/Command/DebugAssetMapperCommand.php
```diff
@@ -0,0 +1,114 @@
...
+            $path = $this->relativizePath($path);
```
chalasr, line 61: We could make `AssetMapperRepository::getRepositories()` able to return the relative paths with some bool argument or even a new method. No big deal though, fine as is

weaverryan: That's true - we already have the projectRootDir there. I don't feel too strongly either way :)

src/Symfony/Component/AssetMapper/AssetMapperRepository.php
```diff
@@ -55,7 +55,7 @@ public function find(string $logicalPath): ?string
 
             $file = rtrim($path, '/').'/'.$localLogicalPath;
             if (file_exists($file)) {
-                return $file;
+                return realpath($file);
```
weaverryan, line 58: Using `realpath` everywhere now in this file so that all paths use the "real" path. That's just "nice" because you'll get back full paths - instead of potentially getting back file paths like `/var/something/else/../other-path/app.js` - and is important especially for `findLogicalPath()` where we pass in a `$filesystemPath` and then do some matching by the filesystem path.

weaverryan:
.. checking on one other edge case bug fix for the repository...

weaverryan:
Ok, this is ready - test failures are unrelated.

stof (approved):

src/Symfony/Component/AssetMapper/Command/DebugAssetMapperCommand.php
```diff
@@ -0,0 +1,114 @@
...
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+/**
+ * Outputs all the assets in the asset mapper.
+ *
+ * @experimental
+ *
+ * @author Ryan Weaver <ryan@symfonycasts.com>
+ */
+#[AsCommand(name: 'debug:assetmap', description: 'Outputs all mapped assets.')]
```
fabpot: Not sure, but I would expect `asset-map` instead of `assetmap`:

```suggestion
#[AsCommand(name: 'debug:asset-map', description: 'Outputs all mapped assets.')]
```

src/Symfony/Component/AssetMapper/Command/DebugAssetMapperCommand.php
```diff
@@ -0,0 +1,114 @@
...
+        parent::__construct();
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->addOption('full', null, null, 'Whether to show the full paths')
+            ->setHelp(<<<'EOT'
+The <info>%command.name%</info> command outputs all of the assets in
+asset mapper for debugging purposes!
```
fabpot:
```suggestion
asset mapper for debugging purposes.
```

src/Symfony/Component/AssetMapper/Command/DebugAssetMapperCommand.php
```diff
@@ -0,0 +1,114 @@
...
+            $rows[] = [
+                $logicalPath,
+                $sourcePath,
+            ];
+        }
+        $io->section('Mapped Assets');
+        $io->table(['Logical Path', 'Filesystem Path'], $rows);
+
+        if ($this->didShortenPaths) {
+            $io->note('To see the full paths, re-run with the --full.');
```
fabpot:
```suggestion
            $io->note('To see the full paths, re-run with --full.');
```

Or

```suggestion
            $io->note('To see the full paths, re-run with the --full option.');
```
?

src/Symfony/Component/AssetMapper/Command/DebugAssetMapperCommand.php
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        return self::SUCCESS;
```
fabpot: We use `return 0;` mostly everywhere in the built-in commands.

fabpot (approved):
Some small suggestions.

src/Symfony/Component/AssetMapper/Command/DebugAssetMapperCommand.php
stof: if only the debug command needs that, I would keep it in the bug command for now

weaverryan:
This is ready for merge - 2 test failures are unrelated. Thanks for the feedback!

fabpot:
Thank you @weaverryan.

PR #50222, opened 2023-05-03
[Mime] Add PGP feature

### LINKED ISSUE #50160: PGPEncrypter for Symfony Mailer
### Description

Because I needed PGP mail encryption in one of my projects and didn't found a compatible solution for the Symfony Mailer, I implemented one myself. It's not finished yet, but maybe a good starting point, if you consider a PGPEncrypter besides the SMimeEncypter.
Currently I have still issues in some mail clients with the sign only part of a message due to the line ending encoding. Maybe you guys can help me out here. I would be happy to help with a co-authored pull-request, if you like what you're seeing.

Let me know, what you think 🙃

### Example

Can be found here:

https://github.com/the-pulli/pgp-mailer/blob/main/example.php

The repository lies here:

https://github.com/the-pulli/pgp-mailer

the-pulli:
As discussed in #50160 here is the corresponding PR.


composer.json
```diff
@@ -40,6 +40,7 @@
         "doctrine/event-manager": "^1.2|^2",
         "doctrine/persistence": "^2|^3",
         "twig/twig": "^2.13|^3.0.4",
+        "pear/crypt_gpg": "^1.6",
```
stof: Adding that as a required dependency is a no-go to me. This must stay an optional feature.

src/Symfony/Component/Mime/Tests/Crypto/PGPEncryptorTest.php
```diff
@@ -0,0 +1,80 @@
...
+
+namespace Symfony\Component\Mime\Tests\Crypto;
+
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Mime\Address;
+use Symfony\Component\Mime\Crypto\PGPEncrypter;
+use Symfony\Component\Mime\Email;
+
+#[CoversClass(PGPEncrypter::class)]
```
stof: We don't use those attributes in Symfony.

composer.json
the-pulli: Okay, thought so too. But without it the unit test breaks. Is there another solution to make them work?

stof: that's what dev requirements are for.

src/Symfony/Component/Mime/composer.json
```diff
@@ -17,6 +17,7 @@
     ],
     "require": {
         "php": ">=8.1",
+        "pear/crypt_gpg": "^1.6",
```
GromNaN, line 20: This needs to be moved to `require-dev` as said in [previous comment](https://github.com/symfony/symfony/pull/50222#discussion_r1184761607).

GromNaN (commented):
This is a great new feature. IMO it should be a new component: `symfony/mime` is independent of these new classes and it would make it easier to resolve dependencies with `pear/crypt_gpg`.

src/Symfony/Component/Mime/composer.json
the-pulli: Haven't I moved it to require-dev in d5b9214a7832b152cbdb27b9c7b68bdfc07cc0df?
I'm glad you like the idea of that feature. A new component would be essential a standalone package under a certain namespace?

GromNaN: > Haven't I moved it to require-dev in [d5b9214](https://github.com/symfony/symfony/commit/d5b9214a7832b152cbdb27b9c7b68bdfc07cc0df)? 

The commit sorts packages alphabetically, but does not move the package to `require-dev`.

the-pulli: Yeah, I did that too. But it is actually under `require-dev` now:

https://github.com/symfony/symfony/blob/d5b9214a7832b152cbdb27b9c7b68bdfc07cc0df/composer.json#L123C1-L162

HypeMC: https://github.com/symfony/symfony/commit/d5b9214a7832b152cbdb27b9c7b68bdfc07cc0df fixes the `composer.json` in the root dir, this is `src/Symfony/Component/Mime/composer.json`.

the-pulli: I see. If it should't live in `symfony/mime` then it should be entirely removed from there and instead moved into a new component (for example `symfony/pgp`) where the dependency is handled?

the-pulli: Hey,

the last months I tested my implementation in a Symfony app and discovered several flaws. For example it doesn't work with templated emails and sending via mail provider via api is also not possible due to the nature of a PGP structured mail.
I still would like to see that feature in Symfony and I'll work on that in the upcoming weeks.

fabpot:
@the-pulli Are you still willing to work on this? I agree with the previous comment that it should be part of a new component that would be well integrated with Mime.

the-pulli:
Sure @fabpot , I can give it another shot. My mentioned package from https://github.com/symfony/symfony/issues/50160 still working well for me in a Laravel app. But I discovered that most of the transactional mail provider doesn't support PGP encryption. The ones working for me were Mailtrap.io and ZohoMail. I keep you posted with updates and it's nice to know that there's still a demand for this feature.

Spomky:
Is it really necessary to use `pear/crypt_gpg`?
Wouldn't it be possible to integrate only what we need to encrypt/sign emails directly into Symfony? I notice that these are mostly command line calls (https://github.com/pear/Crypt_GPG/blob/master/Crypt/GPG.php#L1621-L1666)

* this would avoid throwing third-party exceptions (eg. https://github.com/symfony/symfony/pull/50222/files#diff-776e4312eeeec2cf57f709f82f94b4e6638f0d9df0b3be3a532cbed8866c2ccdR85)
* this would avoid depending on a command line call system while symfony/console is available (https://github.com/pear/Crypt_GPG/blob/master/composer.json#L30-L31)

I am also interested in this feature. I can make myself available to help integrate it

stof:
Closing in favor of https://github.com/symfony/symfony/pull/59372

PR #50225, opened 2023-05-03
[HttpClient] Favor php-http/discovery instead of nyholm/psr7

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

This PR fine-tunes the DX related to using `HttplugClient` and `Psr18Client`.

By favoring `php-http/discovery`, we build on [its latest capabilities](https://github.com/php-http/discovery/releases) to use whatever implementation might be installed and preferred by the app.

Of course, we keep `nyholm/psr7` as a privileged PSR-7 implementation so that when it's installed but `php-http/discovery` is not, things keep working as usual.

The deprecated interfaces from `php-http/message-factory` are also moved to a new conditional and internal `LegacyHttplugInterface`, so that we can make this dependency optional.

### DIFF
--- a/src/Symfony/Component/ErrorHandler/DebugClassLoader.php
+++ b/src/Symfony/Component/ErrorHandler/DebugClassLoader.php
@@ -21,7 +21,6 @@ use PHPUnit\Framework\MockObject\MockObject;
 use Prophecy\Prophecy\ProphecySubjectInterface;
 use ProxyManager\Proxy\ProxyInterface;
 use Symfony\Component\ErrorHandler\Internal\TentativeTypes;
-use Symfony\Component\HttpClient\HttplugClient;
 use Symfony\Component\VarExporter\LazyObjectInterface;
 
 /**
@@ -422,9 +421,7 @@ class DebugClassLoader
             if (!isset(self::$checkedClasses[$use])) {
                 $this->checkClass($use);
             }
-            if (isset(self::$deprecated[$use]) && strncmp($vendor, str_replace('_', '\\', $use), $vendorLen) && !isset(self::$deprecated[$class])
-                && !(HttplugClient::class === $class && \in_array($use, [\Http\Client\HttpClient::class, \Http\Message\RequestFactory::class, \Http\Message\StreamFactory::class, \Http\Message\UriFactory::class], true))
-            ) {
+            if (isset(self::$deprecated[$use]) && strncmp($vendor, str_replace('_', '\\', $use), $vendorLen) && !isset(self::$deprecated[$class])) {
                 $type = class_exists($class, false) ? 'class' : (interface_exists($class, false) ? 'interface' : 'trait');
                 $verb = class_exists($use, false) || interface_exists($class, false) ? 'extends' : (interface_exists($use, false) ? 'implements' : 'uses');
 
--- a/src/Symfony/Component/HttpClient/HttplugClient.php
+++ b/src/Symfony/Component/HttpClient/HttplugClient.php
@@ -17,13 +17,9 @@ use GuzzleHttp\Promise\Utils;
 use Http\Client\Exception\NetworkException;
 use Http\Client\Exception\RequestException;
 use Http\Client\HttpAsyncClient;
-use Http\Client\HttpClient as HttplugInterface;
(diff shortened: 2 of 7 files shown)


src/Symfony/Component/ErrorHandler/DebugClassLoader.php
```diff
@@ -422,9 +421,7 @@ public function checkAnnotations(\ReflectionClass $refl, string $class): array
             if (!isset(self::$checkedClasses[$use])) {
                 $this->checkClass($use);
             }
-            if (isset(self::$deprecated[$use]) && strncmp($vendor, str_replace('_', '\\', $use), $vendorLen) && !isset(self::$deprecated[$class])
-                && !(HttplugClient::class === $class && \in_array($use, [\Http\Client\HttpClient::class, \Http\Message\RequestFactory::class, \Http\Message\StreamFactory::class, \Http\Message\UriFactory::class], true))
```
stof, line 426: Great that we don't need to do special cases here anymore

src/Symfony/Component/HttpClient/Internal/LegacyHttplugInterface.php
```diff
@@ -0,0 +1,35 @@
...
+     * @internal
+     *
+     * @deprecated since Symfony 6.3
+     */
+    interface LegacyHttplugInterface extends HttpClient, RequestFactory, StreamFactory, UriFactory
+    {
+    }
+} else {
+    /**
+     * @internal
```
stof, line 30: I would also mark it as deprecated here, to avoid having conditional definitions that is conditionally deprecated

nicolas-grekas: updated

src/Symfony/Component/HttpClient/Internal/LegacyHttplugInterface.php
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+namespace Symfony\Component\HttpClient\Internal;
+
+use Http\Client\HttpClient;
+use Http\Message\RequestFactory;
+use Http\Message\StreamFactory;
+use Http\Message\UriFactory;
+
+if (interface_exists(RequestFactory::class)) {
```
stof, line 19: What about the existence of `HttpClient` ?

nicolas-grekas: It's always here, because we check for HttpAsyncClient, which covers HttpClient
I realized I forgot to list HttpClient on the "else" path, updated.

stof (approved):

src/Symfony/Component/HttpClient/HttplugClient.php
```diff
@@ -36,37 +32,32 @@
...
 use Symfony\Contracts\Service\ResetInterface;
 
-if (!interface_exists(HttplugInterface::class)) {
-    throw new \LogicException('You cannot use "Symfony\Component\HttpClient\HttplugClient" as the "php-http/httplug" package is not installed. Try running "composer require php-http/httplug".');
-}
-
-if (!interface_exists(RequestFactory::class)) {
-    throw new \LogicException('You cannot use "Symfony\Component\HttpClient\HttplugClient" as the "php-http/message-factory" package is not installed. Try running "composer require php-http/message-factory".');
+if (!interface_exists(HttpAsyncClient::class)) {
+    throw new \LogicException('You cannot use "Symfony\Component\HttpClient\HttplugClient" as the "php-http/httplug" package is not installed. Try running "php-http/discovery php-http/async-client-implementation:*".');
```
andersonamuller, line 43: @nicolas-grekas 
```suggestion
throw new \LogicException('You cannot use "Symfony\Component\HttpClient\HttplugClient" as the "php-http/httplug" package is not installed. Try running "composer require php-http/discovery php-http/async-client-implementation:*".');
```

nicolas-grekas: Fixed thanks!

PR #50231, opened 2023-05-03
[AssetMapper] Fixing 2 bugs related to the compile command and importmaps

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | none
| Tickets       | none
| License       | MIT
| Doc PR        | not needed

Fixes 2 bugs:

A) Running `assetmap:compile` gives you a `public/assets/importmap.json` for fast importmap dumping on production. But, running `assetmap:compile` again later would re-use the `importmap.json` info to create the new `importmap.json`... meaning it would never update after the first compile.

B) The importmap process generates 2 pieces of information: (A) the importmap and (B) which files from the importmap should be preloaded. When we use `assetmap:compile`, we need to dump both pieces of info. So, we now also dump an `importmap.preload.json` file. These are TWO files (and not just one) because we avoid parsing `importmap.json` at runtime: we read the file and dump it straight out. We DO need to parse the "preload" file. So, I've kept them separate.

Blocked by #50219, which has some changes that will fix the tests here.

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
@@ -49,6 +49,7 @@ class ImportMapManager
      */
     private const PACKAGE_PATTERN = '/^(?:https?:\/\/[\w\.-]+\/)?(?:(?<registry>\w+):)?(?<package>(?:@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*)(?:@(?<version>[\w\._-]+))?(?:(?<subpath>\/.*))?$/';
     public const IMPORT_MAP_FILE_NAME = 'importmap.json';
+    public const IMPORT_MAP_PRELOAD_FILE_NAME = 'importmap.preload.json';
 
     private array $importMapEntries;
     private array $modulesToPreload;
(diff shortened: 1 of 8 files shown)


stof (approved):

fabpot:
@weaverryan Can you rebase this PR now that #50219 is merged?

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -125,9 +126,11 @@ private function buildImportMapJson(): void
             return;
         }
 
-        $dumpedPath = $this->assetMapper->getPublicAssetsFilesystemPath().'/'.self::IMPORT_MAP_FILE_NAME;
-        if (file_exists($dumpedPath)) {
-            $this->json = file_get_contents($dumpedPath);
+        $dumpedImportMapPath = $this->assetMapper->getPublicAssetsFilesystemPath().'/'.self::IMPORT_MAP_FILE_NAME;
+        $dumpedModulePreloadPath = $this->assetMapper->getPublicAssetsFilesystemPath().'/'.self::IMPORT_MAP_PRELOAD_FILE_NAME;
+        if (file_exists($dumpedImportMapPath) && file_exists($dumpedModulePreloadPath)) {
```
nicolas-grekas: we prefer `is_file` over `file_exists` when possible because it's been proved faster (even if we don't really care here, but that's a CS thing now)

nicolas-grekas:
(rebase needed)

weaverryan:
Good to go now - thanks for the merges!

nicolas-grekas:
Thank you @weaverryan.

PR #50232, opened 2023-05-03
[AssetMapper] Better public without digest

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | Not needed

We have a `MappedAsset::getPublicPathWithoutDigest()` method, which would return something like `/assets/file2.js`. This, sort of odd method, is super useful, for example, when generating the importmap, where we want to communicate to the browser that when it searches for `/assets/file2.js` (e.g. because inside `assets/file1.js` there is an `import './file1.js'`), here it should use the digested path - e.g.

```
'/assets/file2.js' => '/assets/file2.1234abcd.js',
```

Anyway, the way that I calculated this originally was a little sketchy. In truth, we have this info very easily in `AssetMapper`, so now we pass it in. This also allows an "asset compiler" to figure out which "directory" the final asset will be in, which isn't possible with "public path" as it's still being generated. That little nice detail will be needed in StimulusBundle :).

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/AssetMapper.php
+++ b/src/Symfony/Component/AssetMapper/AssetMapper.php
@@ -138,6 +138,7 @@ class AssetMapper implements AssetMapperInterface
             $asset->setSourcePath($filePath);
 
             $asset->setMimeType($this->getMimeType($logicalPath));
+            $asset->setPublicPathWithoutDigest($this->getPublicPathWithoutDigest($logicalPath));
             $publicPath = $this->getPublicPath($logicalPath);
             $asset->setPublicPath($publicPath);
             [$digest, $isPredigested] = $this->getDigest($asset);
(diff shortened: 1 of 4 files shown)


src/Symfony/Component/AssetMapper/AssetMapper.php
```diff
@@ -239,11 +245,7 @@ private function getMimeType(string $logicalPath): ?string
 
         $extension = pathinfo($logicalPath, \PATHINFO_EXTENSION);
 
-        if (!isset($this->extensionsMap[$extension])) {
-            throw new \LogicException(sprintf('The file extension "%s" from "%s" does not correspond to any known types in the asset mapper. To support this extension, configure framework.asset_mapper.extensions.', $extension, $logicalPath));
-        }
-
-        return $this->extensionsMap[$extension];
+        return $this->extensionsMap[$extension] ?? null;
```
weaverryan, line 248: This is unrelated. I had a file that mentioned a sourcemap - so `foo.js.map` - and the system exploded because we didn't know the mime type of `.map`. The mime type is only important for (A) the dev server to return that asset with the correct`Content-Type` and (B) for asset compilers, if they rely on the mime type. So I think this is a reasonable thing to do.

stof (approved):

nicolas-grekas (approved):

nicolas-grekas:
Thank you @weaverryan.

PR #48300, opened 2022-11-23
[Mailer] Fix: do not clear context of message after rendering

silverbackdan:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #47993
| License       | MIT
| Doc PR        | n/a

When running tests in 6.1 I am checking the message context has been set correctly. The email that is collected in the logger is updated after rendering. This line doesn't seem necessary, but if it is, perhaps we should be looking here and to why the message, when cloned still has the context modified in the profiler.
https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Mailer/Mailer.php#L54

Thoughts?

//cc @fabpot as this change was a commit by him. Happy to discuss.

Thank you

PR #47075, opened 2022-07-27
[Mime] Change the way we avoid rendering an email twice

fabpot:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes-ish
| New feature?  | yes-ish 
| Deprecations? | no 
| Tickets       | n/a 
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/17065

Trying to kill several birds with one stone here.

I was not very happy with the way we ensured that emails were not renderer more than once.
This solution seems easier and allows for people to render an email **before** sending it to Messenger (important when you have Doctrine entities in the context for instance) as documented in the related doc PR.

PR #50233, opened 2023-05-03
[Mime] Keep context after rendering emails

Rainrider:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| License       | MIT

This PR keeps the context after rendering emails. This behavior was present up to 6.1 and was changed with cc3b87189dbd99afd82a77f4986eeacabbacc276.

The context is useful when testing generated values passed to the template. I.e. we generate a salutation for emails depending on user gender and pass it to the template through the context. A test would then look like that:

```php
final class EmailTest extends KernelTestCase
{
    public function testContext(): void
    {
        $this->createUser(); // trigger welcome email

        $this->assertQueuedEmailCount(1);

        $message = $this->getMailerMessage();
        $salutation = $message->getContext()['salutation'];

        $this->assertEmailHtmlBodyContains($message, $salutation);
    }
}
```

This way we can change the salutation generator implementation without having to adapt the functional tests for email delivery and content.

### DIFF
--- a/src/Symfony/Bridge/Twig/Mime/TemplatedEmail.php
+++ b/src/Symfony/Bridge/Twig/Mime/TemplatedEmail.php
@@ -76,7 +76,6 @@ class TemplatedEmail extends Email
     {
         $this->textTemplate = null;
         $this->htmlTemplate = null;
-        $this->context = [];
     }
 
     /**
--- a/src/Symfony/Bridge/Twig/Tests/Mime/BodyRendererTest.php
+++ b/src/Symfony/Bridge/Twig/Tests/Mime/BodyRendererTest.php
@@ -133,6 +133,24 @@ HTML;
         $this->assertEquals('Text', $email->getTextBody());
     }
 
+    public function testRendererKeepsContext()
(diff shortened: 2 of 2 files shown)


fabpot:
See #48300 for a discussion about why this cannot be done this way.

Rainrider:
@fabpot https://github.com/symfony/symfony/pull/47075#discussion_r930788561 states this is for serialization. Could we just remove the context from `__serialize` if this is the only problem with this PR?

Rainrider:
@fabpot I updated my branch based on [this comment](https://github.com/symfony/symfony/pull/48300/files#r1031704015). In case you might consider this a solution, see https://github.com/symfony/symfony/compare/6.2...Rainrider:symfony:mime/keep-email-context (the PR here does not reflect this because you closed it)

Rainrider:
I just switched to using the DomCrawler component to extract the values I need from the rendered email, so I don't have a valid use case for this anymore.

PR #50235, opened 2023-05-04
[HttpClient] Fix getting through proxies via CONNECT

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Spotted while trying to hit google via `symfony proxy:start`:

```sh
https_proxy=http://127.0.0.1:7080 php test.php
```

where test.php contains:

```php
$client = new CurlHttpClient();

$r = $client->request('GET', 'https://google.com/');

dump($r->getStatusCode());
dump($r->getHeaders());
dump($r->getInfo('debug'));
```

### DIFF
--- a/src/Symfony/Component/HttpClient/Response/AmpResponse.php
+++ b/src/Symfony/Component/HttpClient/Response/AmpResponse.php
@@ -47,7 +47,6 @@ final class AmpResponse implements ResponseInterface, StreamableInterface
 
     private $multi;
     private $options;
-    private $canceller;
     private $onProgress;
 
     private static $delay;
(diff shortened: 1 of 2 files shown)


stof:
is there any way to have a test covering this to prevent regressions ?

nicolas-grekas:
I won't be able to add a test covering that because it requires an HTTPS endpoint. Of course we could add some network services, but that's going to be too time consuming to setup properly...

nicolas-grekas:
Of course, I would gladly accept a PR that adds the test if anyone reading is up to give it a try!

PR #50238, opened 2023-05-04
[HttpKernel] Don't use eval() to render ESI/SSI

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Because this might be an important security hardening, this PR is a backport of #50013 for 5.4.

### DIFF
--- a/src/Symfony/Component/HttpKernel/HttpCache/Esi.php
+++ b/src/Symfony/Component/HttpKernel/HttpCache/Esi.php
@@ -80,8 +80,10 @@ class Esi extends AbstractSurrogate
         $content = preg_replace('#<esi\:remove>.*?</esi\:remove>#s', '', $content);
         $content = preg_replace('#<esi\:comment[^>]+>#s', '', $content);
 
+        static $cookie;
+        $cookie = hash('md5', $cookie ?? $cookie = random_bytes(16), true);
+        $boundary = base64_encode($cookie);
         $chunks = preg_split('#<esi\:include\s+(.*?)\s*(?:/|</esi\:include)>#', $content, -1, \PREG_SPLIT_DELIM_CAPTURE);
-        $chunks[0] = str_replace($this->phpEscapeMap[0], $this->phpEscapeMap[1], $chunks[0]);
 
         $i = 1;
         while (isset($chunks[$i])) {
@@ -95,16 +97,10 @@ class Esi extends AbstractSurrogate
                 throw new \RuntimeException('Unable to process an ESI tag without a "src" attribute.');
             }
 
-            $chunks[$i] = sprintf('<?php echo $this->surrogate->handle($this, %s, %s, %s) ?>'."\n",
-                var_export($options['src'], true),
-                var_export($options['alt'] ?? '', true),
-                isset($options['onerror']) && 'continue' === $options['onerror'] ? 'true' : 'false'
-            );
-            ++$i;
-            $chunks[$i] = str_replace($this->phpEscapeMap[0], $this->phpEscapeMap[1], $chunks[$i]);
-            ++$i;
+            $chunks[$i] = $boundary.$options['src']."\n".($options['alt'] ?? '')."\n".('continue' === ($options['onerror'] ?? ''))."\n";
+            $i += 2;
         }
-        $content = implode('', $chunks);
+        $content = $boundary.implode('', $chunks).$boundary;
 
         $response->setContent($content);
         $response->headers->set('X-Body-Eval', 'ESI');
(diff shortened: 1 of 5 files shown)


src/Symfony/Component/HttpKernel/HttpCache/HttpCache.php
```diff
@@ -636,7 +636,21 @@ private function restoreResponseBody(Request $request, Response $response)
             if ($response->headers->has('X-Body-File')) {
                 include $response->headers->get('X-Body-File');
```
stof, line 637: why still have a case relying on including a file here. Is this something we should also try to change ?

stof: or maybe combining `X-Body-File` and `X-Body-Eval` is an impossible case ? And if we stop using PHP code, wouldn't we break this case that expect to include the file to evaluate it ?

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

src/Symfony/Component/HttpKernel/HttpCache/Esi.php
```diff
@@ -95,16 +97,10 @@ public function process(Request $request, Response $response)
...
-                isset($options['onerror']) && 'continue' === $options['onerror'] ? 'true' : 'false'
-            );
-            ++$i;
-            $chunks[$i] = str_replace($this->phpEscapeMap[0], $this->phpEscapeMap[1], $chunks[$i]);
-            ++$i;
+            $chunks[$i] = $boundary.$options['src']."\n".($options['alt'] ?? '')."\n".('continue' === ($options['onerror'] ?? ''))."\n";
+            $i += 2;
         }
-        $content = implode('', $chunks);
+        $content = $boundary.implode('', $chunks).$boundary;
```
stof, line 103: instead of putting the boundaries around the content of the returned response, which force the caller to be aware of it to remove it (even if the content has no ESI tag), we might put the boundary in a header `X-Body-Boundary` (that the caller can still remove), which might make the removal easier

src/Symfony/Component/HttpKernel/Tests/HttpCache/EsiTest.php
```diff
@@ -102,7 +102,7 @@ public function testMultilineEsiRemoveTagsAreRemoved()
         $response = new Response('<esi:remove> <a href="http://www.example.com">www.example.com</a> </esi:remove> Keep this'."<esi:remove>\n <a>www.example.com</a> </esi:remove> And this");
         $this->assertSame($response, $esi->process($request, $response));
 
-        $this->assertEquals(' Keep this And this', $response->getContent());
+        $this->assertEquals(' Keep this And this', substr($response->getContent(), 24, -24));
```
stof, line 105: This looks like a BC break not suitable in a patch release as projects might use the Esi class directly.

src/Symfony/Component/HttpKernel/HttpCache/HttpCache.php
nicolas-grekas: There's no need to change this: the required file cannot come from any user input, and we do validate the name of the file, so that arbitrary file inclusion is not possible.

src/Symfony/Component/HttpKernel/HttpCache/Esi.php
nicolas-grekas: The caller has to be aware of the content in any cases.
Putting it before+after allows a quick check to ensure it's correct in HttpCache:
`if (substr($content, -24) === $boundary = substr($content, 0, 24)) {`

src/Symfony/Component/HttpKernel/Tests/HttpCache/EsiTest.php
nicolas-grekas: This is a security-related fix, I'd better break existing implems that do that so that they can adjust. (I also doubt this will hit anyone in practice :crossed_fingers: )

src/Symfony/Component/HttpKernel/HttpCache/HttpCache.php
stof: Well, from my understanding of re-reading this code, this `X-Body-File` corresponds to a file written in the cache store. If we stop using PHP to implement the ESI logic, we might need to process the boundaries there instead of evaluating the file.

Do we have functional tests covering the case of a processing of ESI tags alongside a cached response of HttpCache ? Because I suspect that this is the case that is now broken (we would write the cache with boundaries instead of PHP code but read it as if it was PHP code).

src/Symfony/Component/HttpKernel/Tests/HttpCache/EsiTest.php
stof: then the boundary length should at least be a public constant, so that they don't have to hardcode this `24` everywhere

PR #50240, opened 2023-05-04
[HttpClient] Add `max_retries` option to `RetryableHttpClient`

danielburger1337:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Tickets       | N/A
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/18288

Added a `max_retries` option to RetryableHttpClient that allows the user to configure the RetryableHttpClient on a per request level, which is especially useful when `retry_failed` was configured for the global `http_client` service.

```php
<?php

use Symfony\Component\HttpClient\HttpClient;
use Symfony\Component\HttpClient\RetryableHttpClient;

$client = HttpClient::create();
$client = new RetryableHttpClient($client, null, 3);

$client->request('GET', '/foo-bar', [
    'max_retries' => 1 // 0 disables retrying
]);

// $client = $client->withOptions(['max_retries' => 1]);
```

### DIFF
--- a/src/Symfony/Component/HttpClient/RetryableHttpClient.php
+++ b/src/Symfony/Component/HttpClient/RetryableHttpClient.php
@@ -60,6 +60,9 @@ class RetryableHttpClient implements HttpClientInterface, ResetInterface
         }
 
         $clone = clone $this;
+        $clone->maxRetries = (int) ($options['max_retries'] ?? $this->maxRetries);
+        unset($options['max_retries']);
+
         $clone->client = $this->client->withOptions($options);
 
         return $clone;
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
What about adding `max_retries` instead? this would be more flexible.
Note that this doesn't need to be an "extra" option - it shouldn't actually.
This will need to override the `withOptions` method also.
You might want to take some inspiration from UriTemplateHttpClient, which also manages its own option.

danielburger1337:
> What about adding `max_retries` instead? this would be more flexible. Not that this doesn't need to be an "extra" option - it shouldn't actually. This will need to override the `withOptions` method also. You might want to take some inspiration from UriTemplateHttpClient, which also manages its own option.

Yes, this would be much more flexible approach and I like it!

To be extra clear, would this option overide the  `maxRetries` property of the RetryableHttpClient when cloned? 
Because otherwise I'm not quite sure what you mean.

nicolas-grekas:
Yes, it would override it indeed.

OskarStark:
Can you please update the pr header? You still mention no_retry. Thanks

danielburger1337:
> Can you please update the pr header? You still mention no_retry. Thanks

What do you mean by PR header? Do you mean my original PR description?

Sorry, I never heard that term before.

OskarStark:
Yes the PR description and the code example. Thanks

danielburger1337:
> Yes the PR description and the code example. Thanks

Done

src/Symfony/Component/HttpClient/RetryableHttpClient.php
```diff
@@ -71,11 +74,17 @@ public function request(string $method, string $url, array $options = []): Respo
         $baseUris = \is_array($baseUris) ? $baseUris : [];
         $options = self::shiftBaseUri($options, $baseUris);
 
-        if ($this->maxRetries <= 0) {
+        $maxRetries = $this->maxRetries;
```
nicolas-grekas:
```suggestion
        $maxRetries = (int) ($options['max_retries'] ?? $this->maxRetries);
        unset($options['max_retries']);
```

nicolas-grekas (approved):
Look good to me thanks. I just have a minor comment.

fancyweb (approved):

nicolas-grekas:
Thank you @danielburger1337.

PR #50241, opened 2023-05-04
[HttpKernel] Prevent initialising lazy services during services reset

tucksaun:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes-ish
| New feature?  | no 
| Deprecations? | no 
| Tickets       | none
| License       | MIT
| Doc PR        | 

If one happens to have a service that is configured to be lazy AND also implements `ResetInterface`, when Symfony resets resettableService, it will call `reset` on the service which will trigger the initialization even though the service has not been actually used by the code, which defeats the purpose of having the service lazy.
To fix this I propose to skip resetting a service that is marked as `LazyObjectInterface` and has not been initialized (ie not used yet)

### DIFF
--- a/src/Symfony/Component/HttpKernel/DependencyInjection/ServicesResetter.php
+++ b/src/Symfony/Component/HttpKernel/DependencyInjection/ServicesResetter.php
@@ -11,6 +11,8 @@
 
 namespace Symfony\Component\HttpKernel\DependencyInjection;
 
+use ProxyManager\Proxy\LazyLoadingInterface;
+use Symfony\Component\VarExporter\LazyObjectInterface;
 use Symfony\Contracts\Service\ResetInterface;
 
 /**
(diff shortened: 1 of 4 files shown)


tucksaun:
If we are ok with going forward with this, shall we support `ProxyManager` proxies given that the bridge is deprecated in 6.3?

nicolas-grekas (approved):

stof:
Do we also have the issue for lazy ghosts or no ?

tucksaun:
> Do we also have the issue for lazy ghosts or no ?

Good question. I would say it should be. But let me check that later on today

nicolas-grekas:
Thank you @tucksaun.

tucksaun:
> > Do we also have the issue for lazy ghosts or no ?
> 
> Good question. I would say it should be. But let me check that later on today

Confirmed, the object was a lazy ghost when I encountered this issue

PR #50242, opened 2023-05-05
[ErrorHandler] Fix the design of the exception page tabs

javiereguiluz:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

In Symfony 6.3 we've updated the tabs of the Profiler to make them accessible. In turn, this also updates the tabs of the exception page from ErrorHandler. However, the design of that exception page hasn't been updated yet, so it looks like this:

exception-page-tabs

Now the tabs look like the profiler. It's not perfect, but it's better than the current broken situation. The proper fix is to update a bit the design of the exception page. Let's do that for Symfony 6.4/7.0. Thanks!

### DIFF
--- a/src/Symfony/Component/ErrorHandler/Resources/assets/css/exception.css
+++ b/src/Symfony/Component/ErrorHandler/Resources/assets/css/exception.css
@@ -9,12 +9,23 @@
     --color-warning: #a46a1f;
     --color-error: #b0413e;
     --color-muted: #999;
-    --tab-background: #fff;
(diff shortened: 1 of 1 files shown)


stof (approved):

fabpot (approved):

fabpot:
Thank you @javiereguiluz.

PR #50251, opened 2023-05-06
[Serializer] Handle datetime deserialization in U format

### LINKED ISSUE #50236: [Serializer] DateTimeNormalizer throws NotNormalizableValueException if using datetime_format U

6.2.10

### Description

Since the changes for the DateTimeNormallizer in [this PR](https://github.com/symfony/symfony/pull/49537) I am getting a NotNormalizableValueException when using datetime_format U (Seconds since the Unix Epoch) for denormalization - our timestamp value comes via API. The error message we now get is "The data is either not an string, an empty string, or null; you should pass a string that can be parsed with the passed format or a valid DateTime string".

The value used for creating DateTime object from timestamp is integer, and now the validation is set to explicitly accept only string values.

### How to reproduce

```
<?php

namespace App\Controller;

use Symfony\Component\HttpFoundation\JsonResponse;
use Symfony\Component\Serializer\Normalizer\DateTimeNormalizer;
use Symfony\Component\Serializer\SerializerInterface;

class DateTimeNormalizerErrorController extends AbstractController
{
    #[Route('/normalizer-bug-example', name: 'normalizer_bug_example')]
    public function __invoke(SerializerInterface $serializer): JsonResponse
    {
        $data = '{"exampleDate":1685716333}';  // Use timestamp to create DateTime
        // $data = '{"exampleDate":"13-05-2023"}';  // When using string value - everything works as expected

        // Throws NotNormalizableValueException
        $triggerErrorObject = $serializer->deserialize($data, ExampleEntity::class, 'json', [
            DateTimeNormalizer::FORMAT_KEY => 'U',  // When using string value - change the format to 'd-m-Y'
        ]);

        // dd($triggerErrorObject->getExampleDate());  // When using string value - this dumps a normal DateTime object

        return new JsonResponse($data, 200, [], true);
    }
}

class ExampleEntity
{
    private \DateTime $exampleDate;

    public function getExampleDate(): \DateTimeInterface
    {
        return $this->exampleDate;
    }

    public function setExampleDate(\DateTime $exampleDate): void
    {
        $this->exampleDate = $exampleDate;
    }
}
```

### Possible Solution

Maybe revert the changes in `src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php:94`:

```
-       if (null === $data || (\is_string($data) && '' === trim($data))) {
+       if (null === $data || !\is_string($data) || '' === trim($data)) {
```

Or cover the case for timestamps in the validation.

### Additional Context

_No response_

tugmaks:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50236 
| License       | MIT
| Doc PR        | -

Fixes https://github.com/symfony/symfony/issues/50236

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
@@ -91,7 +91,14 @@ class DateTimeNormalizer implements NormalizerInterface, DenormalizerInterface, 
         $dateTimeFormat = $context[self::FORMAT_KEY] ?? null;
         $timezone = $this->getTimezone($context);
 
-        if (null === $data || !\is_string($data) || '' === trim($data)) {
+        if (\is_int($data) || \is_float($data)) {
+            switch ($dateTimeFormat) {
+                case 'U': $data = sprintf('%d', $data); break;
+                case 'U.u': $data = sprintf('%.6F', $data); break;
+            }
+        }
+
+        if (!\is_string($data) || '' === trim($data)) {
             throw NotNormalizableValueException::createForUnexpectedDataType('The data is either not an string, an empty string, or null; you should pass a string that can be parsed with the passed format or a valid DateTime string.', $data, [Type::BUILTIN_TYPE_STRING], $context['deserialization_path'] ?? null, true);
         }
 
--- a/src/Symfony/Component/Serializer/Tests/Normalizer/DateTimeNormalizerTest.php
+++ b/src/Symfony/Component/Serializer/Tests/Normalizer/DateTimeNormalizerTest.php
@@ -178,6 +178,8 @@ class DateTimeNormalizerTest extends TestCase
         $this->assertEquals(new \DateTimeImmutable('2016/01/01', new \DateTimeZone('UTC')), $this->normalizer->denormalize('2016-01-01T00:00:00+00:00', \DateTimeImmutable::class));
         $this->assertEquals(new \DateTime('2016/01/01', new \DateTimeZone('UTC')), $this->normalizer->denormalize('2016-01-01T00:00:00+00:00', \DateTime::class));
         $this->assertEquals(new \DateTime('2016/01/01', new \DateTimeZone('UTC')), $this->normalizer->denormalize('  2016-01-01T00:00:00+00:00  ', \DateTime::class));
+        $this->assertEquals(new \DateTimeImmutable('2023-05-06T17:35:34.000000+0000', new \DateTimeZone('UTC')), $this->normalizer->denormalize(1683394534, \DateTimeImmutable::class, null, [DateTimeNormalizer::FORMAT_KEY => 'U']));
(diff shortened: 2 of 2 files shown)


src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
```diff
@@ -83,7 +83,13 @@ public function denormalize(mixed $data, string $type, string $format = null, ar
         $dateTimeFormat = $context[self::FORMAT_KEY] ?? null;
         $timezone = $this->getTimezone($context);
 
-        if (null === $data || !\is_string($data) || '' === trim($data)) {
+        $data = match ($dateTimeFormat){
```
nicolas-grekas:
```suggestion
        if (\is_int($data) || \is_float($data)) {
            $data = match ($dateTimeFormat) {
                'U' => (int) $data,
                'U.u' => sprintf('%.6F', $data),
                default => $data,
            };
        }
```

I'd rather not be extra strict with float vs int, let's accept both in both cases, that might event help with x86 archs.

tugmaks: @nicolas-grekas thank you, hmm, it should be 'U' => (string) $data, right ?

nicolas-grekas: should be `(string) (int) $data`!

nicolas-grekas: or better: `sprintf('%d', $data)`

tugmaks: Fixed

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
```diff
@@ -83,7 +83,15 @@ public function denormalize(mixed $data, string $type, string $format = null, ar
         $dateTimeFormat = $context[self::FORMAT_KEY] ?? null;
         $timezone = $this->getTimezone($context);
 
-        if (null === $data || !\is_string($data) || '' === trim($data)) {
+        if (\is_int($data) || \is_float($data)) {
+            $data = match ($dateTimeFormat){
```
nicolas-grekas: missing space after `)`

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
```diff
@@ -83,7 +83,15 @@ public function denormalize(mixed $data, string $type, string $format = null, ar
...
+                'U' =>  sprintf('%d', $data),
+                'U.u' => sprintf('%.6F', $data),
+                default => $data
```
nicolas-grekas: missing trailing `,`

nicolas-grekas (approved):

nicolas-grekas:
Thank you @tugmaks.

tugmaks:
> Thank you @tugmaks.

Thank you too!

PR #50252, opened 2023-05-06
[Translation] Improve handling of non-string messages in `MessageCatalogue`

### LINKED ISSUE #49894: [Translation] `MessageCatalogue::add()` accepts non-string messages

5.4.22

### Description

`MessageCatalogue::set()` expects message to be string, but `MessageCatalogue::add()` (and the same applies for constructor) does not have any validation for messages array and in practice message could be anything, for example `null`. This creates inconsistencies, because `MessageCatalogue::all()` may return these keys with `null` as value even if `MessageCatalogue::has()` will return `false` for them.

In my case loading https://github.com/rob006-software/flarum-ext-last-post-avatar/blob/1ab3d5a5a7b8959f77481e389a1aba08bf97d37f/locale/en.yml resulted `rob006-last-post-avatar.forum` key with `null` as value, which resulted deprecation warning in Flarum, because they do not expected that `MessageCatalogue::all()` may return non-string messages (see https://github.com/rob006-software/flarum-ext-last-post-avatar/pull/2).

### How to reproduce

```php
$catalogue = new MessageCatalogue('en');
$catalogue->add(['test' => null]);
assert($catalogue->all() === [] || $catalogue->all(['test' => ''])); 
```

### Possible Solution

I think that all messages should be cast to string before adding to `MessageCatalogue::$messages`. I case of loading YAML like https://github.com/rob006-software/flarum-ext-last-post-avatar/blob/1ab3d5a5a7b8959f77481e389a1aba08bf97d37f/locale/en.yml I would expect that these null-keys will be completely skipped, but I'm not sure if skipping them in `MessageCatalogue::add()` would be a good place to do this.

### Additional Context

_No response_

rob006:
| Q             | A
| ------------- | ---
| Branch?       |  5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #49894
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Symfony/Component/Translation/MessageCatalogue.php
+++ b/src/Symfony/Component/Translation/MessageCatalogue.php
@@ -32,6 +32,15 @@ class MessageCatalogue implements MessageCatalogueInterface, MetadataAwareInterf
     public function __construct(string $locale, array $messages = [])
     {
         $this->locale = $locale;
+        foreach ($messages as $domain => $domainMessages) {
+            foreach ($domainMessages as $key => $message) {
+                if (null === $message) {
+                    unset($messages[$domain][$key]);
+                } else {
+                    $messages[$domain][$key] = (string) $message;
+                }
+            }
+        }
         $this->messages = $messages;
     }
 
@@ -157,8 +166,12 @@ class MessageCatalogue implements MessageCatalogueInterface, MetadataAwareInterf
     {
         $altDomain = str_ends_with($domain, self::INTL_DOMAIN_SUFFIX) ? substr($domain, 0, -\strlen(self::INTL_DOMAIN_SUFFIX)) : $domain.self::INTL_DOMAIN_SUFFIX;
         foreach ($messages as $id => $message) {
-            unset($this->messages[$altDomain][$id]);
-            $this->messages[$domain][$id] = $message;
+            if (null === $message) {
+                unset($this->messages[$domain][$id]);
+            } else {
+                unset($this->messages[$altDomain][$id]);
+                $this->messages[$domain][$id] = (string) $message;
+            }
         }
 
         if ([] === ($this->messages[$altDomain] ?? null)) {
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/Translation/MessageCatalogue.php
```diff
@@ -157,8 +166,12 @@ public function add(array $messages, string $domain = 'messages')
     {
         $altDomain = str_ends_with($domain, self::INTL_DOMAIN_SUFFIX) ? substr($domain, 0, -\strlen(self::INTL_DOMAIN_SUFFIX)) : $domain.self::INTL_DOMAIN_SUFFIX;
         foreach ($messages as $id => $message) {
-            unset($this->messages[$altDomain][$id]);
-            $this->messages[$domain][$id] = $message;
+            if (null === $message) {
+                unset($this->messages[$domain][$id]);
+            } else {
+                unset($this->messages[$altDomain][$id]);
+                $this->messages[$domain][$id] = (string) $message;
```
javiereguiluz, line 173: But the `(string) $message` cast will throw an exception if for objects/resources that can't be converted into strings. Maybe it's what we want (to fail loudly so you can catch these errors) but I just wanted to mention it. Thanks.

stof:
Adding the validation in the MessageCatalogue constructor will have a performance impact in prod for 2 reasons:

- this logic will have to run as part of instantiating the cached catalogue (so time spent)
- the array mutations being performed mean that the array will have to be copied out of the OPCache shared memory (so increased memory usage)

To me, it would be much better to leave MessageCatalogue as is (except adding more precise phpdoc describing the accepted array as being `array<string, array<string, string>>`) and adding such validation of invalid values in loaders instead.

rob006:
How about removing this validation from constructor and leaving it only in `add()`?

I could move validation to loaders, but it will be duplicated at least 3 times, and `add()` will be less bulletproof.

stof:
Having the validation in loaders might allow providing better error messages (as they would be able to tell you which resources is being loaded that contains an error). And not all loaders will have the same validation (for instance, Xliff will not allow setting null values as all values are strings in XML)

Side note, I don't think the low-level MessageCatalogue object should silently skip values.

rob006:
Alternative PR: https://github.com/symfony/symfony/pull/50315

nicolas-grekas:
Closing in favor of https://github.com/symfony/symfony/pull/50315

PR #50253, opened 2023-05-06
[FrameworkBundle] Generate caches consistently on successive run of `cache:clear` command

### LINKED ISSUE #46324: Build dir is different on subsequent runs of composer install

6.0.8

### Description

I have a custom build dir:

```php
public function getBuildDir(): string
{
    return $this->getProjectDir().'/var/build/'.$this->environment;
}
```

I've noticed that sometimes `debug:container --deprecations` fails with `The deprecation file does not exist, please try warming the cache first.` due to different build dir. I managed to reproduce this only when running composer install.

### How to reproduce

```
s new --dir symfony_build_dir_problem
cd ./symfony_build_dir_problem
echo '<?php

namespace App;

{
    use MicroKernelTrait;

    public function getBuildDir(): string
    {
        return $this->getProjectDir()."/var/build/".$this->environment;
    }
}' > ./src/Kernel.php
composer install -q
grep 'kernel.build_dir' var/build/dev/App_KernelDevDebugContainer.xml
composer install -q
grep 'kernel.build_dir' var/build/dev/App_KernelDevDebugContainer.xml
```

The result will be:
<parameter key="kernel.build_dir">/Users/vudaltsov/PhpProjects/happy-inc/server/.stuff/test/symfony_build_dir_problem/var/**build**/dev</parameter>
<parameter key="kernel.build_dir">/Users/vudaltsov/PhpProjects/happy-inc/server/.stuff/test/symfony_build_dir_problem/var/**cache**/dev</parameter>

### Possible Solution

_No response_

### Additional Context

_No response_

okhoshi:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #46324 
| License       | MIT
| Doc PR        | N/A

When the Kernel is configured to use different directories for `build_dir` and `cache_dir`, successive runs of the `cache:clear` produce different outputs. See #46324 for the reproduction steps and the differences observed.

This is due to the fact that if the cached kernel needs to be regenerated when invoking `cache:clear` command, the cache warmers are run in the same directory where the kernel was regenerated (`$warmupDir`, which will become the `build_dir`). When the cache is fresh, only cache warmers are run, and in `cache_dir` this time.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Command/CacheClearCommand.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Command/CacheClearCommand.php
@@ -137,14 +137,7 @@ EOF
                 if ($output->isVerbose()) {
                     $io->comment('Warming up optional cache...');
                 }
-                $warmer = $kernel->getContainer()->get('cache_warmer');
-                // non optional warmers already ran during container compilation
-                $warmer->enableOnlyOptionalWarmers();
-                $preload = (array) $warmer->warmUp($realCacheDir);
-
-                if ($preload && file_exists($preloadFile = $realCacheDir.'/'.$kernel->getContainer()->getParameter('kernel.container_class').'.preload.php')) {
-                    Preloader::append($preloadFile, $preload);
-                }
+                $this->warmupOptionals($realCacheDir);
             }
         } else {
             $fs->mkdir($warmupDir);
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Thank you @Okhoshi.

vudaltsov:
Thank you!

PR #50264, opened 2023-05-08
[AssetMapper] Flexible public paths + relative path imports + possibility of "building" assets

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | Still a TODO in general

Hi!

Working hard to see where AssetMapper adjustments need to be made. 3 changes proposed here:

1) AssetMapper's "public output path" awareness was moved to a new `PublicAssetsPathResolver` service
2) The asset compilers were updated to use relative paths when adjusting paths in files.
3) Drop the "mime type" stuff entirely - except for the "dev server" so it can set the content-type. Asset compiler's can use `$mappedAsset->getPublicExtension()` instead.

### Why these changes?

1) **Allow public/assets/ structure to flatten/change**

The new `PublicAssetsPathResolver` + the "adjustment" of import/url paths in the compilers allows the "output" directory structure to be decoupled from the source structure. Currently the output structure matches your "source" structure, so you get things like `public/assets/styles/app.css` or `public/assets/bundles/easyadmin/foo.css`. We could add an option in the future (or the user could decorate `PublicAssetsPathResolver` immediately) to "flatten" everything into just `public/assets/`.

2) **Allow for building of certain files, like `.scss` or `.jsx`**

These changes allow users to create... more controversial functionality that could build certain files automatically. For example, you could have:

```twig
<link rel="stylesheet" href="{{ asset('styles/app.scss') }}">
```

Locally, I have a custom compiler that executes the `dart-sass` binary to convert this to CSS. Then, by decorating `PublicAssetsPathResolver`, I'm changing the public path for `styles/app.scss` so that it ends with `.css`. This is about 50 lines of code that results in the sass-compiled CSS file being properly rendered onto the page.

I have a similar compiler + path decorator locally that compiles `.jsx` => `.js` using the babel binary. Usage:

```javascript
// assets/app.js
import HelloWorldComponent from './HelloReact.jsx';

// ... render like normal
```

I realize usage like this is more controversial... however it's also super pragmatic if you DO use Sass or have some React components. The alternative would be to run the `sass` binary manually, import the built `.css` file and probably ignore that file from `git`.

Thanks for your consideration :).

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/AssetMapper.php
+++ b/src/Symfony/Component/AssetMapper/AssetMapper.php
@@ -11,6 +11,8 @@
 
 namespace Symfony\Component\AssetMapper;
 
+use Symfony\Component\AssetMapper\Path\PublicAssetsPathResolverInterface;
+
 /**
  * Finds and returns assets in the pipeline.
  *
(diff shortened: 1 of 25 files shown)


src/Symfony/Component/AssetMapper/AssetMapperDevServerSubscriber.php
```diff
@@ -26,9 +26,87 @@
  */
 final class AssetMapperDevServerSubscriber implements EventSubscriberInterface
 {
+    // source: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
+    private const EXTENSIONS_MAP = [
+        'aac' => 'audio/aac',
+         'abw' => 'application/x-abiword',
```
javiereguiluz: This and the next lines seem to be indented by 1 extra white space.

src/Symfony/Component/AssetMapper/AssetMapperDevServerSubscriber.php
```diff
@@ -74,19 +154,10 @@ public static function getSubscribedEvents(): array
         ];
     }
 
-    private function extractAssetPathAndDigest(string $fullPath): array
+    private function getMimeType(string $path): ?string
```
javiereguiluz: MIME types is the former name of "media types": https://en.wikipedia.org/wiki/Media_type

I guess it's fine to keep using MimeType because most people know them, but I just wanted to mention this just in case.

src/Symfony/Component/AssetMapper/Compiler/AssetCompilerPathResolverTrait.php
```diff
@@ -47,4 +56,35 @@ private function resolvePath(string $directory, string $filename): string
 
         return implode('/', $output);
     }
+
+    private function createRelativePath(string $fromPath, string $toPath): string
+    {
+        $fromPath = rtrim($fromPath, '/');
+        $toPath = rtrim($toPath, '/');
+
+        $fromParts = explode('/', $fromPath);
```
javiereguiluz: Quick question: are directory separators normalized before calling this function? Should we use `DIRECTORY_SEPARATOR` constant? Does PHP handle all this automatically so this is not important?

weaverryan: Good question. I've been relying on a strong test suite and CI to catch Windows problems. For the most part, paths inside of AssetMapper use `/` always and we normalize for that. In this case, the paths passed to this are the "public paths", which will be the "asset prefix" (e.g. `/assets/`, which is configurable, but it's meant to be the URL so `/` should be used) + the "logical path name" of the asset, which is also normalized to `/`). We may have some blind spot here for Windows... but I'm not aware of it yet. And we can fix those when they come up.

javiereguiluz (approved):
After thinking about this, I changed my initial thoughts and now I'm 👍 about this.

It's true that it's controversial and the _"you are recreating Assetic"_ argument is half-true .... but at the same time:

* Without any utility to build SCSS into CSS, I can't use AssetMapper. I'd stick to Webpack, Encore, node.js, etc.
* My current usage of SCSS is minimal _(and from what I've seen, for many other Symfony developers is the same)_
* In fact, my only current reason for using SCSS is selector nesting. This is already a native browser feature (https://caniuse.com/css-nesting) so, as soon as there's more support for it, I'll remove SCSS and just use plain CSS.

So, this "controversial" feature would be useful for me during the 1 or 2 year transition from today until I can stop using SCSS and move to CSS.

Thanks Ryan!

src/Symfony/Component/AssetMapper/AssetMapper.php
```diff
@@ -195,29 +113,26 @@ public function getPublicPath(string $logicalPath): ?string
...
 
-        return $this->publicPrefix.preg_replace_callback('/\.(\w+)$/', function ($matches) use ($digest) {
+        $digestedPath = preg_replace_callback('/\.(\w+)$/', function ($matches) use ($digest) {
             return "-{$digest}{$matches[0]}";
         }, $logicalPath);
+
+        return $this->assetsPathResolver->resolvePublicPath($digestedPath);
     }
 
     private function getPublicPathWithoutDigest(string $logicalPath): string
```
stof: do we even need to keep this private method ?

weaverryan: yup, that can be simplified 👍

stof (approved):

fabpot:
Thank you @weaverryan.

PR #49806, opened 2023-03-25
[Scheduler] randomize a cron schedule

vtsykun:
| Q             | A
| ------------- | ---
| Branch?       | 6.3 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

This PR resolve problem of running things at midnight or once an hour at XX:00, to avoid loading peaks every hour/minutes.
Example usage 

```php
return (new Schedule())
    ->add(RecurringMessage::cron('*/10 * * * *', new \stdClass(), 120))
```

Selection_1375

PR #50270, opened 2023-05-08
[Scheduler] add `JitterTrigger`

kbond:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | #49806
| License       | MIT
| Doc PR        | n/a

I think this is what @vtsykun was trying to accomplish with #49806. FreeBSD has a `-jitter` option that does the same thing so I think there's precedent. From the (https://man.freebsd.org/cgi/man.cgi?cron):

> -j	jitter
>	     Enable time jitter.  Prior	to executing commands, cron will sleep
	     a random number of	seconds	in the range from 0 to jitter.	~This
	     will not affect superuser jobs (see -J).~  A value for jitter must
	     be	between	0 and 60 inclusive.  Default is	0, which effectively
	     disables time jitter.
>	     This option can help to smooth down system	load spikes during mo-
	     ments when	a lot of jobs are likely to start at once, e.g., at
	     the beginning of the first	minute of each hour.

Usage:

```php
// defaults to 0-60 seconds
$schedule->add(RecurringMessage::cron('@daily', $message1)->withJitter());

// customize the max seconds (0-30)
$schedule->add(RecurringMessage::cron('@daily', $message1)->withJitter(30));
```

### DIFF
--- a/src/Symfony/Component/Scheduler/RecurringMessage.php
+++ b/src/Symfony/Component/Scheduler/RecurringMessage.php
@@ -14,6 +14,7 @@ namespace Symfony\Component\Scheduler;
 use Symfony\Component\Scheduler\Exception\InvalidArgumentException;
 use Symfony\Component\Scheduler\Trigger\CronExpressionTrigger;
 use Symfony\Component\Scheduler\Trigger\DateIntervalTrigger;
+use Symfony\Component\Scheduler\Trigger\JitterTrigger;
 use Symfony\Component\Scheduler\Trigger\TriggerInterface;
 
 /**
@@ -57,6 +58,11 @@ final class RecurringMessage
     public static function trigger(TriggerInterface $trigger, object $message): self
     {
         return new self($trigger, $message);
+    }
+
+    public function withJitter(int $maxSeconds = 60): self
+    {
+        return new self(new JitterTrigger($this->trigger, $maxSeconds), $this->message);
     }
 
     public function getMessage(): object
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/Scheduler/Trigger/CronExpressionTrigger.php
```diff
@@ -73,9 +76,26 @@ public static function fromSpec(string $expression = '* * * * *', string $contex
         return new self(new CronExpression(self::parseHashed($expression, $context)));
     }
 
+    public function jitter(int $value = 60): self
```
kbond: I kept this as a _builder_ method but can add as a constructor argument if desired.

fabpot:
I'm wondering if we should do this at a higher level. Having a jitter is probably a common need for most triggers.

kbond:
Yes, I was thinking that also.

src/Symfony/Component/Scheduler/Trigger/CronExpressionTrigger.php
javiereguiluz: Just wondering if we could rename the method argument from `$value` to `$seconds` or `$numOfSeconds` or something that makes it self-explanatory. Thanks.

kbond:
> I'm wondering if we should do this at a higher level. Having a jitter is probably a common need for most triggers.

I've updated this PR to add a `JitterTrigger` that can add a _jitter_ to any trigger. Also, added `RecurringMessage::withJitter()` to help with DX (description updated to show usage).

src/Symfony/Component/Scheduler/Trigger/CronExpressionTrigger.php
kbond: Good idea. I've renamed to `$maxSeconds`.

fabpot (approved):

fabpot:
Thank you @kbond.

PR #50272, opened 2023-05-09
cast integer to string in case date time format is „U“

### LINKED ISSUE #50236: [Serializer] DateTimeNormalizer throws NotNormalizableValueException if using datetime_format U

PercyDwn:
In case the date time format is „U“, the data comes as a integer (timestamp in seconds). To keep the „data is string“-condition, the integer is casted to a string.

| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #50236 
| License       | MIT
| Doc PR        |  

Fixes DateTimeNormalizer::denormalize throwing a NotNormalizableValueException when DateTimeFormat is "U" and data is passed as an integer.
The Exception should not be thrown, as data being an integer is valid for a timestamp.
My solution is to check if $dateTImeFormat is "U" and $data is an integer. In that case the integer is casted to a string. 
This way, the condition that $data is a string does not need to be changed.

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
@@ -96,6 +96,8 @@ class DateTimeNormalizer implements NormalizerInterface, DenormalizerInterface, 
         $dateTimeFormat = $context[self::FORMAT_KEY] ?? null;
         $timezone = $this->getTimezone($context);
 
+        if ( "U" === $dateTimeFormat &&  "integer" === gettype($data)) $data = (string) $data;
+
         if (null === $data || !\is_string($data) || '' === trim($data)) {
             throw NotNormalizableValueException::createForUnexpectedDataType('The data is either not an string, an empty string, or null; you should pass a string that can be parsed with the passed format or a valid DateTime string.', $data, [Type::BUILTIN_TYPE_STRING], $context['deserialization_path'] ?? null, true);
         }
--- a/src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
@@ -80,9 +80,6 @@ class DateTimeNormalizer implements NormalizerInterface, DenormalizerInterface, 
         return $object->format($dateTimeFormat);
     }
 
-    /**
-     * @param array $context
-     */
     public function supportsNormalization(mixed $data, string $format = null /* , array $context = [] */): bool
     {
         return $data instanceof \DateTimeInterface;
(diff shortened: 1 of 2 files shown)


stof:
I'd rather change the type check to also allow integers when using `U` instead of casting the value before the type check.

stof:
And this needs to have tests covering it to prevent regressions

nicolas-grekas:
We might want to account for `U.u` also.
Please use is_int instead of gettype while updating :)

PercyDwn:
Thank you for the feedback, @stof and @nicolas-grekas.

Is it the better solution to allow integers and floats for `U` and `U.u`? 
In the PHP documentation the datetime for `createFromFormat` is typed as a string:
```
public static DateTimeImmutable::createFromFormat(string $format, string $datetime, ?[DateTimeZone] $timezone = null): [DateTimeImmutable]|false
```
With PHP 8.1 following code works just fine (does not return false)
```
$object = DateTimeImmutable::createFromFormat("U", 1683635480);
$object = DateTimeImmutable::createFromFormat("U.u", 1683635480.111111);
```
but could this behavior change in the future?

stof:
@PercyDwn My guess is that this code would not work fine with strict types.

stof:
> Is it the better solution to allow integers and floats for `U` and `U.u`?

IMO, yes

PercyDwn:
Okay so I updated the type check to allow integers when dateTimeFormat is "U". 
But I ignored "U.u" as writing a test for it was a total mess. I tried allowing floats with "U.u", but i have the feeling that in such a case the data would come as a string, not as a float. I was not able to write a working test where the time (including milliseconds) came as a float and was identical with the resulting denormalized date.

Some of the github tests are failing, but I am not quite sure why.

nicolas-grekas:
Fixed by #50251

PR #50274, opened 2023-05-09
[HttpClient] Add option `crypto_method` to set the minimum TLS version and make it default to v1.2

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Idea borrowed from https://github.com/async-aws/aws/pull/1402 by @GrahamCampbell 

Note that Firefox/Chrome disabled support for TLS < 1.2 in 2020 and TLSv1.2 is available since 2008.


src/Symfony/Component/HttpClient/HttpClientTrait.php
```diff
@@ -116,6 +116,15 @@ private static function prepareRequest(?string $method, ?string $url, array $opt
             $options['peer_fingerprint'] = self::normalizePeerFingerprint($options['peer_fingerprint']);
         }
 
+        if (isset($options['crypto_method']) && !\in_array($options['crypto_method'], [
+            \STREAM_CRYPTO_METHOD_TLSv1_0_CLIENT,
```
GrahamCampbell, line 121: Shouldn't we define our own constants here, rather than leaking constants from the PHP streams API?

GrahamCampbell (changes requested):

src/Symfony/Component/HttpClient/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+ * Add option `crypto_method` to set the minimum SSL version and make it default to TLSv1.2
```
GrahamCampbell: Defaulting to 1.2 is a breaking change. I think we need to default to 1.0, then change this to 1.2 in v7.

GrahamCampbell (changes requested):

src/Symfony/Component/HttpClient/CHANGELOG.md
GrahamCampbell: Unless we treat this as a security/bug fix... which means we need to target an earlier branch. ;)

src/Symfony/Component/HttpClient/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
```
GrahamCampbell: Target branch is 6.3?

GrahamCampbell (changes requested):

src/Symfony/Component/HttpClient/CHANGELOG.md
nicolas-grekas: I think we should consider this as an improvement, that's why I target the feature-branch

src/Symfony/Component/HttpClient/CHANGELOG.md
nicolas-grekas: branch 6.4 will be opened after 6.3 reaches it's first RC, for now I have no other branch to target

src/Symfony/Component/HttpClient/HttpClientTrait.php
nicolas-grekas: I'd rather not, those existing consts are just fine.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1849,6 +1849,9 @@ private function addHttpClientSection(ArrayNodeDefinition $rootNode, callable $e
                                         ->variableNode('md5')->end()
                                     ->end()
                                 ->end()
+                                ->scalarNode('crypto_method')
+                                    ->info('The minimum version of SSL to accept; must be one of STREAM_CRYPTO_METHOD_TLSv*_CLIENT constants.')
```
javiereguiluz: Minor suggestion:

`The minimum version of SSL to accept [...]`
->
`The minimum version of TLS/SSL to accept [...]`

TLS is the successor protocol to SSL, so I think we should mention it and do that in the first place. Thanks.

stof:
I don't see where you actually configure the min TLS version of the amp/socket TLS context in this PR. Is this actually implemented in this PR ?
Side note, the default in amp/socket is TLS 1.2, so the current behavior for what is supported in symfony/http-client depends on the client being used.

src/Symfony/Component/HttpClient/Internal/AmpClientState.php
```diff
@@ -141,6 +142,7 @@ private function getClient(array $options): array
         $options['local_cert'] && $context = $context->withCertificate(new Certificate($options['local_cert'], $options['local_pk']));
         $options['ciphers'] && $context = $context->withCiphers($options['ciphers']);
         $options['capture_peer_cert_chain'] && $context = $context->withPeerCapturing();
+        $options['crypto_method'] && $context = $context->withMinimumVersion($options['crypto_method']);
```
nicolas-grekas, line 145: @stof this is the line you're looking for

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
nicolas-grekas: updated, thanks for the suggestion

nicolas-grekas:
PR updated to target 6.3 in case we'd like to merge it now.

Status: needs review

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
stof: I would even mention only TLS, as we don't support any SSL version for this setting.

nicolas-grekas: PR updated to tell only about "TLS"

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

GrahamCampbell:
FYI, this has now similarly been implemented in Guzzle 7.6.0, though without the change to the default, not just because I feel it is arguably breaking, but also because Guzzle tends to have people using much older technology including versions of curl that may not even have the needed constant defined.

PR #50279, opened 2023-05-09
[Mailer] [Sendinblue] refacto sendinblue to brevo

PEtanguy:
WIP: I will make a new bridge altogether.

| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | TODO

Hello,

This PR is aimed at updating the config for the sendinblue mailer.
As you might have seen, Sendinblue has rebranded to [Brevo](https://developers.brevo.com/) and also rewrote their apis.  
This change ensure compatibility with the new endpoints and removes any reference to Sendinblue.

I do not think that this change qualifies as a feature but I also do not believe that it is a bug fix.

Please let me know if that does not fit in 6.3


OskarStark:
From a code perspective we should introduce this as a new bridge and deprecate the sendinblue bridge. AFAIK we had a bridge which was "renamed" in the past, not sure yet how we did in this case.

@fabpot please leave your 2 cents here, thanks

fabpot:
> From a code perspective we should introduce this as a new bridge and deprecate the sendinblue bridge. AFAIK we had a bridge which was "renamed" in the past, not sure yet how we did in this case.
> 
> @fabpot please leave your 2 cents here, thanks

That's correct.

PEtanguy:
@fabpot @OskarStark that's what I was wondering.

I will have a go at it, thanks.

PEtanguy:
I opened 2 new PRs to add a new notifier and a new mailer

https://github.com/symfony/symfony/pull/50296

PR #50280, opened 2023-05-09
[PropertyAccess] Fix nullsafe operator on array index

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Currently the nullsafe operator doesn't work with array index paths. However, this is not obvious at first since `THROW_ON_INVALID_PROPERTY_PATH` is not enabled by default:

```php
use Symfony\Component\PropertyAccess\PropertyAccessor;

var_dump((new PropertyAccessor())->getValue([], '[foo]')); // NULL
var_dump((new PropertyAccessor())->getValue([], '[foo?]')); // NULL

var_dump((new PropertyAccessor(throw: PropertyAccessor::THROW_ON_INVALID_INDEX))->getValue([], '[foo]'));
// Cannot read index "foo" while trying to traverse path "[foo]".

var_dump((new PropertyAccessor(throw: PropertyAccessor::THROW_ON_INVALID_INDEX))->getValue([], '[foo?]'));
// Cannot read index "foo" while trying to traverse path "[foo]". (THIS IS WRONG, SHOULD BE NULL)
```


nicolas-grekas:
Thank you @HypeMC.

PR #50281, opened 2023-05-10
[FlashBag] Add `peekMultiple` method

### LINKED ISSUE #49659: FlashBag peekMultiple() method
### Description

The FlashBag allows us to `peek()` for a SINGLE type of message, or for ALL messages, but it would be useful to peek for multiple specific types.

My use case is to check whether or not there are any messages before outputting a wrapper HTML element.

I considered extending the existing `peekAll()` method to take an optional argument, but a dedicated method feels "more correct".

If this sounds like a useful feature I can supply a PR.

### Example

Before:
```twig
{% if app.session.flashbag.peek('error') is not empty or app.session.flashbag.peek('warning') is not empty or app.session.flashbag.peek('notice') is not empty or app.session.flashbag.peek('foo') is not empty or app.session.flashbag.peek('bar') is not empty or app.session.flashbag.peek('baz') is not empty %}
	<div class="messages-wrapper">
		{# show the messages #}
	</div>
{% endif %}
```

After:
```
{% if app.session.flashbag.peekMultiple(['error','warning','notice','foo','bar','baz']) is not empty %}
	<div class="messages-wrapper">
		{# show the messages #}
	</div>
{% endif %}
```
Note that I would prefer to pass a variable number of string arguments to peekMultiple() rather than an array, but I can't see how that would work in conjunction with allowing a final `$default` argument, as the current `peek()` method does.

alexmerlin:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #49659
| License       | MIT
| Doc PR        | -

This PR adds the `peekMultiple` method to `FlashBag` and `AutoExpireFlashBag`, as requested in #49659.


alexmerlin:
The PHP8.2 unit test fails, though my modifications are not related to Redis.
Can someone please take a look?

PR #50286, opened 2023-05-10
[AssetMapper] Add cached asset factory

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | Still TODO

Hi!

This is built on top of #50264. I CAN separate them if needed - sorry for the noise.

tl;dr When using asset mapper, to load the page, we need to build the importmap & resolve a few paths, like `{{ asset('styles/app.css') }}`. Because asset mapper loads the contents and performs some regex on CSS and JS files, this can slow down the page in dev (in prod, `asset-map:compile` removes all overhead).

This PR fixes that (I can see a big improvement in my test app). We "cache" the `MappedAsset` objects via the config cache system so that if an underlying file changes (e.g. `assets/styles/app.css`), we only need to rebuild that ONE `MappedAsset` when the page is loading. 

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/AssetDependency.php
+++ b/src/Symfony/Component/AssetMapper/AssetDependency.php
@@ -19,11 +19,20 @@ namespace Symfony\Component\AssetMapper;
 final class AssetDependency
 {
     /**
-     * @param bool $isLazy whether this dependency is immediately needed
+     * @param bool $isLazy              Whether the dependent asset will need to be loaded eagerly
+     *                                  by the parent asset (e.g. a CSS file that imports another
+     *                                  CSS file) or if it will be loaded lazily (e.g. an async
+     *                                  JavaScript import).
+     * @param bool $isContentDependency Whether the parent asset's content depends
+     *                                  on the child asset's content - e.g. if a CSS
+     *                                  file imports another CSS file, then the parent's
+     *                                  content depends on the child CSS asset, because
+     *                                  the child's digested filename will be included.
      */
     public function __construct(
         public readonly MappedAsset $asset,
-        public readonly bool $isLazy,
+        public readonly bool $isLazy = false,
+        public readonly bool $isContentDependency = true,
     ) {
     }
 }
--- a/src/Symfony/Component/AssetMapper/AssetMapper.php
+++ b/src/Symfony/Component/AssetMapper/AssetMapper.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Component\AssetMapper;
 
+use Symfony\Component\AssetMapper\Factory\MappedAssetFactoryInterface;
 use Symfony\Component\AssetMapper\Path\PublicAssetsPathResolverInterface;
 
 /**
@@ -23,50 +24,24 @@ use Symfony\Component\AssetMapper\Path\PublicAssetsPathResolverInterface;
 class AssetMapper implements AssetMapperInterface
 {
     public const MANIFEST_FILE_NAME = 'manifest.json';
-    private const PREDIGESTED_REGEX = '/-([0-9a-zA-Z]{7,128}\.digested)/';
 
     private ?array $manifestData = null;
-    private array $fileContentsCache = [];
-    private array $assetsBeingCreated = [];
-
-    private array $assetsCache = [];
 
     public function __construct(
         private readonly AssetMapperRepository $mapperRepository,
-        private readonly AssetMapperCompiler $compiler,
+        private readonly MappedAssetFactoryInterface $mappedAssetFactory,
         private readonly PublicAssetsPathResolverInterface $assetsPathResolver,
     ) {
     }
 
     public function getAsset(string $logicalPath): ?MappedAsset
     {
-        if (\in_array($logicalPath, $this->assetsBeingCreated, true)) {
-            throw new \RuntimeException(sprintf('Circular reference detected while creating asset for "%s": "%s".', $logicalPath, implode(' -> ', $this->assetsBeingCreated).' -> '.$logicalPath));
+        $filePath = $this->mapperRepository->find($logicalPath);
+        if (null === $filePath) {
+            return null;
         }
 
-        if (!isset($this->assetsCache[$logicalPath])) {
-            $this->assetsBeingCreated[] = $logicalPath;
-
-            $filePath = $this->mapperRepository->find($logicalPath);
-            if (null === $filePath) {
-                return null;
-            }
-
-            $asset = new MappedAsset($logicalPath);
-            $this->assetsCache[$logicalPath] = $asset;
-            $asset->setSourcePath($filePath);
-
-            $asset->setPublicPathWithoutDigest($this->assetsPathResolver->resolvePublicPath($logicalPath));
-            $publicPath = $this->getPublicPath($logicalPath);
-            $asset->setPublicPath($publicPath);
-            [$digest, $isPredigested] = $this->getDigest($asset);
-            $asset->setDigest($digest, $isPredigested);
-            $asset->setContent($this->calculateContent($asset));
-
-            array_pop($this->assetsBeingCreated);
-        }
-
-        return $this->assetsCache[$logicalPath];
+        return $this->mappedAssetFactory->createMappedAsset($logicalPath, $filePath);
     }
 
     /**
@@ -103,56 +78,9 @@ class AssetMapper implements AssetMapperInterface
             return $manifestData[$logicalPath];
         }
 
-        $filePath = $this->mapperRepository->find($logicalPath);
-        if (null === $filePath) {
-            return null;
-        }
+        $asset = $this->getAsset($logicalPath);
 
-        // grab the Asset - first look in the cache, as it may only be partially created
-        $asset = $this->assetsCache[$logicalPath] ?? $this->getAsset($logicalPath);
-        [$digest, $isPredigested] = $this->getDigest($asset);
-
-        if ($isPredigested) {
-            return $this->assetsPathResolver->resolvePublicPath($logicalPath);
-        }
-
-        $digestedPath = preg_replace_callback('/\.(\w+)$/', function ($matches) use ($digest) {
-            return "-{$digest}{$matches[0]}";
-        }, $logicalPath);
-
-        return $this->assetsPathResolver->resolvePublicPath($digestedPath);
-    }
-
-    /**
-     * Returns an array of "string digest" and "bool predigested".
-     *
-     * @return array{0: string, 1: bool}
-     */
-    private function getDigest(MappedAsset $asset): array
-    {
-        // check for a pre-digested file
-        if (1 === preg_match(self::PREDIGESTED_REGEX, $asset->getLogicalPath(), $matches)) {
-            return [$matches[1], true];
-        }
-
-        return [
-            hash('xxh128', $this->calculateContent($asset)),
-            false,
-        ];
-    }
-
-    private function calculateContent(MappedAsset $asset): string
-    {
-        if (isset($this->fileContentsCache[$asset->getLogicalPath()])) {
-            return $this->fileContentsCache[$asset->getLogicalPath()];
-        }
-
-        $content = file_get_contents($asset->getSourcePath());
-        $content = $this->compiler->compile($content, $asset, $this);
-
-        $this->fileContentsCache[$asset->getLogicalPath()] = $content;
-
-        return $content;
+        return $asset?->getPublicPath();
     }
 
     private function loadManifest(): array
--- a/src/Symfony/Component/AssetMapper/AssetMapperCompiler.php
+++ b/src/Symfony/Component/AssetMapper/AssetMapperCompiler.php
@@ -22,21 +22,24 @@ use Symfony\Component\AssetMapper\Compiler\AssetCompilerInterface;
  */
 class AssetMapperCompiler
 {
+    private AssetMapperInterface $assetMapper;
+
     /**
      * @param iterable<AssetCompilerInterface> $assetCompilers
+     * @param \Closure(): AssetMapperInterface $assetMapperFactory
      */
-    public function __construct(private iterable $assetCompilers)
+    public function __construct(private readonly iterable $assetCompilers, private readonly \Closure $assetMapperFactory)
     {
     }
 
-    public function compile(string $content, MappedAsset $mappedAsset, AssetMapperInterface $assetMapper): string
+    public function compile(string $content, MappedAsset $mappedAsset): string
     {
         foreach ($this->assetCompilers as $compiler) {
             if (!$compiler->supports($mappedAsset)) {
                 continue;
             }
 
-            $content = $compiler->compile($content, $mappedAsset, $assetMapper);
+            $content = $compiler->compile($content, $mappedAsset, $this->assetMapper ??= ($this->assetMapperFactory)());
         }
 
         return $content;
--- a/src/Symfony/Component/AssetMapper/Compiler/CssAssetUrlCompiler.php
+++ b/src/Symfony/Component/AssetMapper/Compiler/CssAssetUrlCompiler.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Component\AssetMapper\Compiler;
 
+use Symfony\Component\AssetMapper\AssetDependency;
 use Symfony\Component\AssetMapper\AssetMapperInterface;
 use Symfony\Component\AssetMapper\MappedAsset;
 
@@ -47,7 +48,7 @@ final class CssAssetUrlCompiler implements AssetCompilerInterface
                 return $matches[0];
             }
 
-            $asset->addDependency($dependentAsset);
+            $asset->addDependency(new AssetDependency($dependentAsset));
             $relativePath = $this->createRelativePath($asset->getPublicPathWithoutDigest(), $dependentAsset->getPublicPath());
 
             return 'url("'.$relativePath.'")';
--- a/src/Symfony/Component/AssetMapper/Factory/CachedMappedAssetFactory.php
+++ b/src/Symfony/Component/AssetMapper/Factory/CachedMappedAssetFactory.php
@@ -1,0 +1,74 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\AssetMapper\Factory;
+
+use Symfony\Component\AssetMapper\MappedAsset;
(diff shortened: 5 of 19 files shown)


src/Symfony/Component/AssetMapper/AssetMapperCompiler.php
```diff
@@ -25,20 +26,25 @@ class AssetMapperCompiler
     /**
      * @param iterable<AssetCompilerInterface> $assetCompilers
      */
-    public function __construct(private iterable $assetCompilers)
+    public function __construct(private iterable $assetCompilers, private ContainerInterface $container)
```
weaverryan: This is the ugliest part of this PR, and it's actually a small fix so that decorating the AssetMapper will work properly.

* It's super convenient for compilers to have access to the AssetMapper
* But, we get a circular reference: AssetMapper -> MappedAssetFactory -> AssetMapperCompiler -> AssetMapper

Is there a more elegant way to make *just* this one dependency lazy?

src/Symfony/Component/AssetMapper/Factory/CachedMappedAssetFactory.php
```diff
@@ -0,0 +1,50 @@
...
+
+use Symfony\Component\AssetMapper\MappedAsset;
+use Symfony\Component\Config\ConfigCache;
+use Symfony\Component\Config\Resource\FileResource;
+
+class CachedMappedAssetFactory implements MappedAssetFactoryInterface
+{
+    public function __construct(
+        private MappedAssetFactoryInterface $innerFactory,
+        private string $cacheDir,
```
stof: should they be readonly, like in other classes in this component ?

src/Symfony/Component/AssetMapper/Factory/CachedMappedAssetFactory.php
```diff
@@ -0,0 +1,50 @@
...
+            return null;
+        }
+
+        $resources = [new FileResource($mappedAsset->getSourcePath())];
+        foreach ($mappedAsset->getDependencies() as $dependency) {
+            if (!$dependency->contentDependency) {
+                continue;
+            }
+
+            $resources[] = new FileResource($dependency->asset->getSourcePath());
```
stof: what about recursive dependencies ? Are they handled properly ?

src/Symfony/Component/AssetMapper/Factory/CachedMappedAssetFactory.php
```diff
@@ -0,0 +1,50 @@
...
+        }
+        $configCache->write(serialize($mappedAsset), $resources);
+
+        return $mappedAsset;
+    }
+
+    private function getCacheFilePath(string $logicalPath, string $sourcePath): string
+    {
+        return $this->cacheDir . '/mapped_assets_cache_' . md5($logicalPath.':'.$sourcePath) . '.php';
```
stof: why using md5 here and not xxh128 which is faster ?

stof: Instead of a `mapped_assets_cache_` prefix on files being directly in the cache folder, I think this cache should be in a dedicated subfolder, as done for most caches in FrameworkBundle (which would then be the responsibility of the bundle when configuring the cache directory)

src/Symfony/Component/AssetMapper/Factory/CachedMappedAssetFactory.php
```diff
@@ -0,0 +1,50 @@
...
+        private MappedAssetFactoryInterface $innerFactory,
+        private string $cacheDir,
+    )
+    {
+    }
+
+    public function createMappedAsset(string $logicalPath, string $sourcePath): ?MappedAsset
+    {
+        $cachePath = $this->getCacheFilePath($logicalPath, $sourcePath);
+        $configCache = new ConfigCache($cachePath, true);
```
stof: should we always enable debug mode ?

src/Symfony/Component/AssetMapper/AssetDependency.php
```diff
@@ -19,11 +19,20 @@
 final class AssetDependency
 {
     /**
-     * @param bool $isLazy whether this dependency is immediately needed
+     * @param bool $isLazy Whether the dependent asset will need to be loaded eagerly
+     *                     by the parent asset (e.g. a CSS file that imports another
+     *                     CSS file) or if it will be loaded lazily (e.g. an async
+     *                     JavaScript import).
+     * @param bool $contentDependency Whether the parent asset's content depends
```
stof: I don't see any place where you configure a dependency as a content dependency. Any asset compiler retrieving the public path of a dependency needs to do it.

src/Symfony/Component/AssetMapper/MappedAsset.php
```diff
@@ -120,9 +120,9 @@ public function setContent(string $content): void
         $this->content = $content;
     }
 
-    public function addDependency(self $asset, bool $isLazy = false): void
+    public function addDependency(self $asset, bool $isLazy = false, bool $contentDependency = false): void
```
stof: this method needs to have phpdoc describing those 2 arguments, as they are the ones that implementors of asset compilers will see in their IDE when using the code, not the AssetDependency constructor (alternatively, we might change this to `->addDependency(AssetDependency $dependency)` to avoid having to duplicate the signature)

stof: shouldn't `$contentDependency` default to `true` to be safer ? Marking dependencies as affecting the content when they are not would lead to useless cache invalidation (which only makes thing slower) while the opposite leads to returning the wrong result.

src/Symfony/Component/AssetMapper/Factory/CachedMappedAssetFactory.php
```diff
@@ -0,0 +1,58 @@
...
+
+        $resources = [new FileResource($mappedAsset->getSourcePath())];
+        foreach ($mappedAsset->getDependencies() as $dependency) {
+            if (!$dependency->contentDependency) {
+                continue;
+            }
+
+            $resources[] = new FileResource($dependency->asset->getSourcePath());
+        }
+        $configCache->write(serialize($mappedAsset), $resources);
```
stof: Given that the MappedAsset object contains AssetDependency objects that contain other MappedAsset objects, what happens when you serialize and unserialize those ?

Also, what is the impact of unserializing a MappedAsset with dependencies that will have MappedAsset objects not known by the AssetMapper internal cache (while potentially referring to the same assets) and also different from MappedAsset created for dependencies of other cached MappedAsset ?

src/Symfony/Component/AssetMapper/Factory/CachedMappedAssetFactory.php
weaverryan: they're not! Thx - fixing now.

src/Symfony/Component/AssetMapper/Factory/CachedMappedAssetFactory.php
weaverryan: I'm using `param('kernel.cache_dir').'/asset_mapper',` in the `asset_mapper.php` service config. It's not configurable like other systems... but does it need to be?

You're right about this `mapped_assets_cache_` prefix - that's redundant - removing that.

src/Symfony/Component/AssetMapper/Factory/CachedMappedAssetFactory.php
weaverryan: It's a good question. In practice, this system will never be used in production, because you'll have dumped your `manifest.json` and `importmap.json` files, so only those will be used. 

But, might as well. I've updated it to use `%kernel.debug%`.

src/Symfony/Component/AssetMapper/AssetDependency.php
weaverryan: ty - great catch

src/Symfony/Component/AssetMapper/Factory/CachedMappedAssetFactory.php
stof: it does not need to be configurable (others are not either). When reviewing this code, I did not check the service config to see whether the prefix was just redundant or if it was there to put files directly in the main cache folder.

src/Symfony/Component/AssetMapper/AssetMapperCompiler.php
stof: You could inject a lazy proxy here, using the same kind of trick than used when using `#[Autowire(lazy: AssetMapperInterface::class)]` for autowiring. But this would make the component much harder to use standalone if the circular dependency is between constructors (as the _only_ way to instantiate the object graph would be through a lazy proxy)

src/Symfony/Component/AssetMapper/MappedAsset.php
weaverryan: Good idea just to use `->addDependency(AssetDependency $dependency)`

src/Symfony/Component/AssetMapper/Factory/CachedMappedAssetFactory.php
weaverryan: > Given that the MappedAsset object contains AssetDependency objects that contain other MappedAsset objects, what happens when you serialize and unserialize those ?

It seems to work fine, in practice. Both `MappedAsset` and `AssetDependency` are objects that hold scalars. For example, here is a dump of one situation (dumping the result of the `unserialize` in the cached factory) in a test project:

Screenshot 2023-05-10 at 1 32 25 PM

What concerns do you have?

> Also, what is the impact of unserializing a MappedAsset with dependencies that will have MappedAsset objects not known by the AssetMapper internal cache (while potentially referring to the same assets) and also different from MappedAsset created for dependencies of other cached MappedAsset ?

I think I understand what you're referring to. So, for example:

* `asset1.css` has a dependency on `asset2.css`. When something asks for the `asset1.css` asset, its cache is fresh, so we get back `MappedAsset` instances from the cache for both `asset1.css` and `asset2.css`.
* Meanwhile, `asset3.css` also has a dependency on `asset2.css`. When something asks for the `asset3.css` asset, its cache is NOT fresh, so we get back FRESH `MappedAsset `instances for both `asset3.css` and `asset2.css`.

So, at this moment, there are two distinct `asset2.css` `MappedAsset` objects floating around the system. Their data is equal, but the objects are not identical. Other than being wasteful, I don't think it's a problem. `MappedAsset` objects are "read-only" informational objects. Well, except for `MappedAsset::addDependency()`. But, in practice, once a `MappedAsset` is returned from `AssetMapper`, it is intended to be read-only: its data is entirely set *during* the creation process (we could even "enforce" this by adding some flag to "freeze" the `MappedAsset` and thus reject calling `addDependency()`, but that might be overkill).

However, if we're concerned about this, we *could*, "refresh" the dependencies after fetching a cached `MappedAsset` - e.g. loop over `$mappedAsset->getDependencies()`... grab the `->logicalName` + `->getSourcePath()` of each, call `$this->createMappedAsset($logicalName, $sourcePath)` on each, then "replace" the `MappedAsset` on each dependency.

tl;dr I don't see any real problem, but I could add extra code to be absolutely sure.

src/Symfony/Component/AssetMapper/AssetMapperCompiler.php
weaverryan: > But this would make the component much harder to use standalone

Thank you - that makes me want to keep this as it is :)

fabpot:
Should be rebased :)

weaverryan:
Rebased, reviewed again (by me) and tested again in my local app. Looks good 👍

src/Symfony/Component/AssetMapper/AssetMapperCompiler.php
nicolas-grekas: what about using a closure here, that'd return the mapper? that's a service_closure inthe DI world

src/Symfony/Component/AssetMapper/Factory/MappedAssetFactory.php
```diff
@@ -0,0 +1,111 @@
...
+    private function getPublicPath(MappedAsset $asset): ?string
+    {
+        [$digest, $isPredigested] = $this->getDigest($asset);
+
+        if ($isPredigested) {
+            return $this->assetsPathResolver->resolvePublicPath($asset->logicalPath);
+        }
+
+        $digestedPath = preg_replace_callback('/\.(\w+)$/', function ($matches) use ($digest) {
+            return "-{$digest}{$matches[0]}";
```
nicolas-grekas: short closure FTW

src/Symfony/Component/AssetMapper/Tests/Compiler/JavaScriptImportPathCompilerTest.php
```diff
@@ -34,6 +34,9 @@ public function testCompile(string $sourceLogicalName, string $input, array $exp
             $actualDependencies[$dependency->asset->logicalPath] = $dependency->isLazy;
         }
         $this->assertEquals($expectedDependencies, $actualDependencies);
+        if (\count($expectedDependencies) > 0) {
```
nicolas-grekas: we never "count" on array: 
```suggestion
        if ($expectedDependencies) {
```

src/Symfony/Component/AssetMapper/AssetMapper.php
```diff
@@ -23,50 +24,24 @@
...
-            $asset->setContent($this->calculateContent($asset));
-
-            array_pop($this->assetsBeingCreated);
+        $filePath = $this->mapperRepository->find($logicalPath);
+        if (null === $filePath) {
+            return null;
         }
 
-        return $this->assetsCache[$logicalPath];
+        return $this->mappedAssetFactory->createMappedAsset($logicalPath, $filePath);
```
nicolas-grekas: can't this get `$this` as argument to remove the need for the container/service-closure on the compiler?

stof: passing `$this` will make this incompatible with decorating the AssetMapper. as it would not pass the decorator.

Not sure it is needed though. This feature relies on decorating the asset factory, not the AssetMapper.

weaverryan: Stof is right. If someone decorated the main `AssetMapper`, then the compilers would get the core, non-decorated one. It's... a bit unfortunate, I admit.

src/Symfony/Component/AssetMapper/AssetMapperCompiler.php
fabpot: I don't think we can keep this as is TBH.

weaverryan: I'll fix this up shortly!

weaverryan:
This should be happy now - `service_closure` added - much, much nicer.

src/Symfony/Component/AssetMapper/AssetMapperCompiler.php
```diff
@@ -22,23 +22,34 @@
...
+            $content = $compiler->compile($content, $mappedAsset, $this->getAssetMapper());
         }
 
         return $content;
     }
+
+    private function getAssetMapper(): AssetMapperInterface
+    {
+        if (!isset($this->assetMapper)) {
+            $this->assetMapper = $this->assetMapperFactory->__invoke();
```
fabpot: Would that work?

```suggestion
            $this->assetMapper = ($this->assetMapperFactory)();
```

src/Symfony/Component/AssetMapper/AssetMapperCompiler.php
```diff
@@ -22,23 +22,34 @@
  */
 class AssetMapperCompiler
 {
+    private AssetMapperInterface $assetMapper;
+
     /**
      * @param iterable<AssetCompilerInterface> $assetCompilers
```
stof: I suggest adding `@param \Closure(): AssetMapperInterface $assetMapperFactory` to document the expected signature of the closure

src/Symfony/Component/AssetMapper/Compiler/CssAssetUrlCompiler.php
```diff
@@ -47,7 +48,7 @@ public function compile(string $content, MappedAsset $asset, AssetMapperInterfac
                 return $matches[0];
             }
 
-            $asset->addDependency($dependentAsset);
+            $asset->addDependency(new AssetDependency($dependentAsset, isContentDependency: true));
```
stof: This is the default for `isContentDependency` here.

src/Symfony/Component/AssetMapper/AssetMapperCompiler.php
```diff
@@ -22,23 +22,35 @@
...
-    public function compile(string $content, MappedAsset $mappedAsset, AssetMapperInterface $assetMapper): string
+    public function compile(string $content, MappedAsset $mappedAsset): string
     {
         foreach ($this->assetCompilers as $compiler) {
             if (!$compiler->supports($mappedAsset)) {
                 continue;
             }
 
-            $content = $compiler->compile($content, $mappedAsset, $assetMapper);
+            $content = $compiler->compile($content, $mappedAsset, $this->getAssetMapper());
```
nicolas-grekas:
```suggestion
            $content = $compiler->compile($content, $mappedAsset, $this->assetMapper ??= ($this->assetMapperFactory)());
```

src/Symfony/Component/AssetMapper/Factory/MappedAssetFactory.php
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Returns an array of "string digest" and "bool predigested".
+     *
+     * @return array{0: string, 1: bool}
+     */
+    private function getDigest(MappedAsset $asset): array
+    {
+        // check for a pre-digested file
+        if (1 === preg_match(self::PREDIGESTED_REGEX, $asset->getLogicalPath(), $matches)) {
```
nicolas-grekas:
```suggestion
        if (self::PREDIGESTED_REGEX, $asset->getLogicalPath(), $matches)) {
```

nicolas-grekas (approved):

weaverryan:
Updated - thanks for the suggestions :)

src/Symfony/Component/AssetMapper/Factory/MappedAssetFactory.php
stof: the suggestion removes too many things

weaverryan: you're correct - but I got it right when I applied it:

```php
if (preg_match(self::PREDIGESTED_REGEX, $asset->getLogicalPath(), $matches))
```

fabpot:
Thank you @weaverryan.

PR #50287, opened 2023-05-10
[Messenger] Store dates in UTC when using Doctrine

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Spotted by running the receiver in an UTC container and the sender in a non-UTC process.

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
+++ b/src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
@@ -125,7 +125,7 @@ class Connection implements ResetInterface
      */
     public function send(string $body, array $headers, int $delay = 0): string
     {
-        $now = new \DateTimeImmutable();
+        $now = new \DateTimeImmutable('UTC');
         $availableAt = $now->modify(sprintf('+%d seconds', $delay / 1000));
 
         $queryBuilder = $this->driverConnection->createQueryBuilder()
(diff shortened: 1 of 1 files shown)


fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

alex-dev:
`new \DateTimeImmutable('UTC')` does not work. `new \DateTimeImmutable('now', new \DateTimeZone('UTC'))` is the proper syntax.

nicolas-grekas:
Any proof? As far as I tested, it works, but I may have missed something of course!

DesLynx:
Hi @nicolas-grekas !
Do you think this change should be reflected in the messenger documentation? There's an alert for the use of the doctrine transport when writing in the same database from different server potentially using a different timezone. But I came across an issue due to this change.
Although both my servers use the same timezone they don't run the same symfony version (one is 5.4 the other 6.3) and so when the server running on 5.4 sends a message to the one on 6.3 it write the message as available in its timezone but the server on 6.3 look for message available in the UTC timezone...
Bonus idea: I would have appreciate to be allowed to specified the timezone to use for a transport via an option in the transport config (with a default value set to UTC to stay compliant with this change). This would also allow for an easy fix for servers in different timezone.

nicolas-grekas:
@DesLynx it could make sense to document this yes, even if it's mostly internal state. PR welcome. About making this configurable, what would be the purpose? I'm not sure it would enable any valid use case. Better sync settle everything on UTC.

Issue #47580, opened 2022-09-15
Ban `DateTime` from the codebase


nicolas-grekas:
By being mutable, `DateTime` is flawed. We should figure out a way to remove it from the codebase and have some CI check that ensures we don't add it back by mistake. I suppose we could configure psalm to do that check for us.

Help welcome!

stof:
We will probably not be able to ban it entirely. We still need to have support for working with DateTime in several places (for instance, the form and serializer components need to support them, otherwise we're not banning them from the core but from the ecosystem)

nicolas-grekas:
True. We need to keep supporting them but ideally we would not generate any anymore in v7.
We could also try to actively move ppl out of DateTime by deprecating handling them in some critical places, but that's another step.

ro0NL:
i'll assume we generally don't care about tests?

nicolas-grekas:
We do, the ban should be on all the codebase.

ro0NL:
What's your thought on `DateTimeInterface`? Do we prefer `DateTime|DateTimeImmutable` where needed?

nicolas-grekas:
We should use DateTimeImmutable when possible, and consider BC/FC otherwise to use DateTimeInterface or not.
This might mean in practice: use `DateTimeInterface` on arguments and `DateTimeImmutable` on return types.
Ideally, the engine should consider the union as equivalent to the interface. That's not the case right now: https://3v4l.org/vBTvD

stof:
To me, when we accept an input for a date we will format, we should use DateTimeInterface.
For return types (and for argument types where it makes sense, for instance in private methods that might care about that), we should use `DateTimeImmutable` rather than the union type)

ghost:
This is HUGE! Mutable DateTime should have been reverted from the PHP Core already in 5.5 when DateTimeImmutable was introduced to address DateTime's mutability issues. Unfortunately this did not happen and we still have it around today.
I think if Symfony makes such a big step in their codebase then this will speed up the process to remove it from the core what I think has already started with Derick's tweet in July.
I would like to help you with the migration.

ro0NL:
> DateTime is flawed.

what's PHP internals POV on the matter? I feel like preserving DateTime with new immutable methods is the fastest route, while deprecating anything mutable.

if DateTime is flawed, so is DateTimeInterface IMHO

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

[...]

PR #50289, opened 2023-05-10
[Messenger] Use immutable dates in the storage when using Doctrine

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| License       | MIT
| Doc PR        | -

We already started using `DateTimeImmutable` in that class but we missed these updates.


stof (approved):
This won't matter much in practice (when converting from PHP to the database, both types accept any DateTimeInterface, the difference is only when converting in the other direction, which is not used in this class). But this is indeed good for consistency.

PR #50290, opened 2023-05-10
[Security] Make `PersistentToken` immutable and tell `TokenProviderInterface::updateToken()` implementations should accept `DateTimeInterface`

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Symfony/Component/Security/Core/Authentication/RememberMe/PersistentToken.php
+++ b/src/Symfony/Component/Security/Core/Authentication/RememberMe/PersistentToken.php
@@ -22,9 +22,9 @@ final class PersistentToken implements PersistentTokenInterface
     private string $userIdentifier;
     private string $series;
     private string $tokenValue;
-    private \DateTime $lastUsed;
+    private \DateTimeImmutable $lastUsed;
 
-    public function __construct(string $class, string $userIdentifier, string $series, #[\SensitiveParameter] string $tokenValue, \DateTime $lastUsed)
+    public function __construct(string $class, string $userIdentifier, string $series, #[\SensitiveParameter] string $tokenValue, \DateTimeInterface $lastUsed)
     {
         if (empty($class)) {
             throw new \InvalidArgumentException('$class must not be empty.');
@@ -43,7 +43,7 @@ final class PersistentToken implements PersistentTokenInterface
         $this->userIdentifier = $userIdentifier;
         $this->series = $series;
         $this->tokenValue = $tokenValue;
-        $this->lastUsed = $lastUsed;
+        $this->lastUsed = \DateTimeImmutable::createFromInterface($lastUsed);
     }
 
     public function getClass(): string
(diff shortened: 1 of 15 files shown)


src/Symfony/Component/Security/Core/Authentication/RememberMe/PersistentToken.php
```diff
@@ -38,6 +38,9 @@ public function __construct(string $class, string $userIdentifier, string $serie
         if (empty($tokenValue)) {
             throw new \InvalidArgumentException('$tokenValue must not be empty.');
         }
+        if ($lastUsed instanceof \DateTime) {
```
ro0NL: this is not really "banning DateTime from the codebase" :angel: 

we should probably restore one test then ;)

nicolas-grekas: test case added

stof: I suggest using `$lastUsed = \DateTimeImmutable::createFromInterface($lastUsed);` unconditionnally, which will avoid those discussions about not really banning DateTime from the codebase when the argument accepts the interface.

nicolas-grekas: good idea, updated

src/Symfony/Component/Scheduler/Trigger/CronExpressionTrigger.php
```diff
@@ -75,7 +75,7 @@ public static function fromSpec(string $expression = '* * * * *', string $contex
 
     public function getNextRunDate(\DateTimeImmutable $run): ?\DateTimeImmutable
     {
-        return \DateTimeImmutable::createFromMutable($this->expression->getNextRunDate($run));
+        return \DateTimeImmutable::createFromInterface($this->expression->getNextRunDate($run));
```
danielburger1337: This is a good change because it provides compatability if the dragonmantank/cron-expression package ever changes its return type. 

But why is a change to the Scheduler component included here?

nicolas-grekas: I'm moving this to a separate PR.

nicolas-grekas:
PR rebased and ready for merge in 6.3 if desired.

nicolas-grekas:
PR ready for 6.3.

fabpot:
@nicolas-grekas This PR should be adapted for 6.4 instead of 6.3.

nicolas-grekas:
PR updated, ready for 6.4

xabbuh:
The low deps failures are related, aren't they?

nicolas-grekas:
> The low deps failures are related, aren't they?

Right, fixed!

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

PR #50291, opened 2023-05-10
[AssetMapper] Adding "excluded_patterns" option

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | Still TODO

Hi!

Invariably, some people will use AssetMapper with certain files that will need to be "built". Examples might be:

A) An `assets/app.css` that uses Tailwind, and the user builds this manually to `assets/build/app.css` and includes *that* file in their pipeline. The source `assets/app.css` should not be published to `public/assets`.
B) Similar with SCSS - e.g. all `*.scss` might be ignored because you are building them yourself.

The new `excluded_patterns` is a glob, for simplicity, and must match the entire file path - hence examples like `*/assets/*.scss`

Cheers!


src/Symfony/Component/AssetMapper/AssetMapperRepository.php
```diff
@@ -160,4 +169,21 @@ private function normalizeLogicalPath(string $logicalPath): string
...
     }
+
+    private function isExcluded(string $filesystemPath): bool
+    {
+        // normalize Windows slashes and remove trailing slashes
+        $filesystemPath = rtrim(str_replace('\\', '/', $filesystemPath), '/');
+
+        foreach ($this->excludedPathPatterns as $pattern) {
+            if (str_ends_with($filesystemPath, 'ignored.css')) {
+                dump($filesystemPath, $pattern, preg_match($pattern, $filesystemPath));
```
fabpot: Is it needed? :joy:

weaverryan: That's a very useful function! :D 

Removed now - thank you

fabpot (approved):

fabpot:
Thank you @weaverryan.

PR #47776, opened 2022-10-03
[PropertyAccess] Implement DateTime type conversion in PropertyAccess

curry684:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| License       | MIT

In short
===
This PR enables implicit conversion between `\DateTime` and `\DateTimeImmutable` objects in the PropertyAccess component, allowing developers to store values of one type where the other was typed. 

Longer story
===

Since PHP introduced the `\DateTimeImmutable` type, and related to that the `\DateTimeInterface`, and afterwards added strict typing, programmers have been banging their heads over converting between 2 datetime types which essentially implement the same thing but are incompatible without helper methods. https://github.com/nelmio/alice/issues/1096 is a notable example of the frustrations this can cause, it makes no sense for a helper library like Faker to implement tons of extra methods just to support that PHP has 2 different ways of storing dates. Similar issues occur with Symfony Forms and Doctrine in certain integrations and combinations.

Which got me thinking why the PropertyAccess component, at the core of a lot of these integrations, doesn't simply support automatic conversion between the types. Conversion from a DateTime to a DateTimeImmutable is by definition harmless - the copy in the target is locked. The other way around it may not be entirely safe, as a programmer could expect a modification of the target to be reflected in the source property, however this is not a logical expectation since in that case the source was Immutable. Hence I implemented the implicit conversion both ways.

Coercion was set up to be more extensible for future similar cases, and will only be attempted for method-based and public access. Coercing values for adder/remover makes no sense as it would (nearly) always break the remover functionality. Likewise coercion is specifically not attempted for union/intersection types, only for explicitly named types.

PR #50295, opened 2023-05-10
[PropertyAccess] Auto-cast from/to DateTime/Immutable when appropriate

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Best reviewed [ignoring whitespaces](https://github.com/symfony/symfony/pull/50295/files?w=1).

My starting point is a form with a date defined as such:

```php
class MyFormType extends AbstractType {
    public function buildForm(FormBuilderInterface $builder, array $options): void
    {
        $builder
            ->add('some_date', DateTimeType::class)
        ;
    }
// ...
}

class MyEntity
{
    private ?\DateTimeImmutable $someDate = null;
//...
}
```

Then doing something like:
```php
$myEntity = new MyEntity();
$form = $this->createForm(MyFormType::class, $myEntity);
```

If I submit this form, I get:
```
Expected argument of type "DateTimeImmutable", "DateTime" given at property path "some_date".
```

The current solution is to set option `input` to `datetime_immutable` in `buildForm()`.

But this is boring. Too much boilerplate.

I've tried looking at the Form component to define the `input` option based on the entity provided when building the form, but this entity is not passed to sub-forms, so that we have no way to do this.

The only way I found is the one attached: if we cannot set a `DateTime`, we try setting a `DateTimeImmutable`.

I thought about changing the default value of option `input` to `datetime_immutable`, but this would need a nasty deprecation that'd force everybody to write even more boilerplate before v7, while what we want is less.

So here we are.

### DIFF
--- a/src/Symfony/Component/PropertyAccess/PropertyAccessor.php
+++ b/src/Symfony/Component/PropertyAccess/PropertyAccessor.php
@@ -514,7 +514,7 @@ class PropertyAccessor implements PropertyAccessorInterface
      *
      * @throws NoSuchPropertyException if the property does not exist or is not public
      */
-    private function writeProperty(array $zval, string $property, mixed $value): void
+    private function writeProperty(array $zval, string $property, mixed $value, bool $recursive = false): void
     {
         if (!\is_object($zval[self::VALUE])) {
             throw new NoSuchPropertyException(sprintf('Cannot write property "%s" to an array. Maybe you should write the property path as "[%1$s]" instead?', $property));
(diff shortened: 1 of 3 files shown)


ro0NL:
would the try/catch be possible on the calling side?

stof:
If we go this way, I suggest supporting the auto-cast in both directions. This would make it easier to change the default input mode in the Form component in a future version.
In 7.0, the Form component should probably be changed to use DateTimeImmutable is its norm data and converting to DateTime when the input needs it rather than working with mutable DateTime for all intermediate layers. But this would make sense if we don't convert twice for projects using DateTimeImmutable and the default input format.

nicolas-grekas:
> would the try/catch be possible on the calling side?
> supporting the auto-cast in both directions

Done for both.

nicolas-grekas:
Now with tests. PR ready for review.

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

curry684:
Good that the feature is in, but you're now relying on exceptions for normal program flow. The approach I followed in #47776 does not, and may therefore be more performant I believe?

nicolas-grekas:
Could you run some bench? The reflection logic might be quite involving also, especially considering using types/etc

curry684:
Also I made some specific exceptions:

> Coercion was set up to be more extensible for future similar cases, and will only be attempted for method-based and public access. Coercing values for adder/remover makes no sense as it would (nearly) always break the remover functionality. Likewise coercion is specifically not attempted for union/intersection types, only for explicitly named types.

Your implementation is not excluding adder/remover (which may cause collections to end up with a rag tag mix of all types) nor union/intersection types.

PR #50297, opened 2023-05-11
Remove unnecessary usages of DateTime

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| License       | MIT
| Doc PR        | -

Together with #50290, this PR removes `DateTime` everywhere possible.

What remains is:
- test cases that test both DateTime and DateTimeImmutable - legit to keep
- date/time form-types and transformers in the Form component - we'd need a separate effort for them, related to #50295
- `PersistentTokenInterface::getLastUsed(): \DateTime` - separate effort also

### DIFF
--- a/src/Symfony/Bridge/Doctrine/Security/RememberMe/DoctrineTokenProvider.php
+++ b/src/Symfony/Bridge/Doctrine/Security/RememberMe/DoctrineTokenProvider.php
@@ -89,12 +89,12 @@ class DoctrineTokenProvider implements TokenProviderInterface, TokenVerifierInte
         $sql = 'UPDATE rememberme_token SET value=:value, lastUsed=:lastUsed WHERE series=:series';
         $paramValues = [
             'value' => $tokenValue,
-            'lastUsed' => $lastUsed,
+            'lastUsed' => \DateTimeImmutable::createFromInterface($lastUsed),
             'series' => $series,
         ];
         $paramTypes = [
             'value' => ParameterType::STRING,
-            'lastUsed' => Types::DATETIME_MUTABLE,
+            'lastUsed' => Types::DATETIME_IMMUTABLE,
             'series' => ParameterType::STRING,
         ];
         if (method_exists($this->conn, 'executeStatement')) {
@@ -118,14 +118,14 @@ class DoctrineTokenProvider implements TokenProviderInterface, TokenVerifierInte
             'username' => $token->getUserIdentifier(),
             'series' => $token->getSeries(),
             'value' => $token->getTokenValue(),
-            'lastUsed' => $token->getLastUsed(),
+            'lastUsed' => \DateTimeImmutable::createFromInterface($token->getLastUsed()),
         ];
         $paramTypes = [
             'class' => ParameterType::STRING,
             'username' => ParameterType::STRING,
             'series' => ParameterType::STRING,
             'value' => ParameterType::STRING,
-            'lastUsed' => Types::DATETIME_MUTABLE,
+            'lastUsed' => Types::DATETIME_IMMUTABLE,
         ];
         if (method_exists($this->conn, 'executeStatement')) {
             $this->conn->executeStatement($sql, $paramValues, $paramTypes);
@@ -186,6 +186,7 @@ class DoctrineTokenProvider implements TokenProviderInterface, TokenVerifierInte
         $this->conn->beginTransaction();
         try {
             $this->deleteTokenBySeries($tmpSeries);
+            $lastUsed = \DateTime::createFromInterface($lastUsed);
             $this->createNewToken(new PersistentToken($token->getClass(), $token->getUserIdentifier(), $tmpSeries, $token->getTokenValue(), $lastUsed));
 
             $this->conn->commit();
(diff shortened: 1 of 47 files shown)


src/Symfony/Bridge/Doctrine/Security/RememberMe/DoctrineTokenProvider.php
```diff
@@ -89,12 +89,12 @@ public function updateToken(string $series, #[\SensitiveParameter] string $token
         $sql = 'UPDATE rememberme_token SET value=:value, lastUsed=:lastUsed WHERE series=:series';
         $paramValues = [
             'value' => $tokenValue,
-            'lastUsed' => $lastUsed,
+            'lastUsed' => \DateTimeImmutable::createFromInterface($lastUsed),
```
stof, line 92: this can even be skipped. When writing data to the database, both `Types::DATETIME_MUTABLE` and `Types::DATETIME_IMMUTABLE` accept the interface.

src/Symfony/Component/HttpFoundation/HeaderBag.php
```diff
@@ -193,15 +193,17 @@ public function remove(string $key)
...
      *
+     * @return \DateTimeImmutable|null
+     *
      * @throws \RuntimeException When the HTTP header is not parseable
      */
-    public function getDate(string $key, \DateTime $default = null): ?\DateTimeInterface
+    public function getDate(string $key, \DateTimeInterface $default = null): ?\DateTimeInterface
     {
         if (null === $value = $this->get($key)) {
             return $default;
```
stof: If we want to enforce returning a DateTimeImmutable, we need to convert the default

src/Symfony/Bridge/Doctrine/Security/RememberMe/DoctrineTokenProvider.php
nicolas-grekas: right, updted

nicolas-grekas: actually tests fail if I remove this line

stof (approved):

yceruto (approved):

src/Symfony/Component/Form/Tests/Extension/Validator/ValidatorTypeGuesserTest.php
```diff
@@ -93,6 +93,7 @@ public static function guessTypeProvider()
             [new Type('long'), new TypeGuess(IntegerType::class, [], Guess::MEDIUM_CONFIDENCE)],
             [new Type('string'), new TypeGuess(TextType::class, [], Guess::LOW_CONFIDENCE)],
             [new Type(\DateTime::class), new TypeGuess(DateType::class, [], Guess::MEDIUM_CONFIDENCE)],
+            [new Type(\DateTimeImmutable::class), new TypeGuess(DateType::class, ['input' => 'datetime_immutable'], Guess::MEDIUM_CONFIDENCE)],
```
HeahDude, line 96: Missing a case for `\DateTimeInterface::class`?

HeahDude (approved):

fabpot (approved):

fabpot:
@nicolas-grekas  Can you have a look at the tests, they seem broken by your changes.

nicolas-grekas:
PR ready, failures are now false-positives.

wouterj:
This is for 6.4 right, given we are close to RC phase?

src/Symfony/Component/Validator/Constraints/AbstractComparisonValidator.php
```diff
@@ -66,13 +66,10 @@ public function validate(mixed $value, Constraint $constraint)
         // the DateTime constructor:
         // https://php.net/datetime.formats
         if (\is_string($comparedValue) && $value instanceof \DateTimeInterface) {
-            // If $value is immutable, convert the compared value to a DateTimeImmutable too, otherwise use DateTime
-            $dateTimeClass = $value instanceof \DateTimeImmutable ? \DateTimeImmutable::class : \DateTime::class;
```
chalasr, line 70: references to `DateTime` in the comment above should probably be updated

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
```diff
@@ -108,29 +108,29 @@ public function denormalize(mixed $data, string $type, string $format = null, ar
         }
 
         try {
-            if (null !== $dateTimeFormat) {
-                $object = \DateTime::class === $type ? \DateTime::createFromFormat($dateTimeFormat, $data, $timezone) : \DateTimeImmutable::createFromFormat($dateTimeFormat, $data, $timezone);
+            if (\DateTimeInterface::class === $type) {
+                $type = \DateTimeImmutable::class;
```
chalasr, line 112: using  `\DatetimeImmutable` instead of `$type` everywhere would read better imho

chalasr (commented):
While i'm tempted to vote 6.3 given it is the continuation a long-running work started a while ago, I think every public method signature changed in this PR deserves to be mentioned in the relevant changelog. It is likely going to impact a lot of final applications as they have to deal with that flaw and they often do it the wrong way (mostly on/because legacy code). So a whole set of beta releases wouldn't be too much either imho... we can't blame people to not test beta releases if we merge breaking changes after them, especially ones that don't bring much value from userland. I'd prefer postponing too

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
nicolas-grekas: this would break when $type is "DateTime" or a child class of "DateTime(Immutable)"

src/Symfony/Component/Validator/Constraints/AbstractComparisonValidator.php
nicolas-grekas: updated thanks

nicolas-grekas:
@chalasr UPGRADE/CHANGELOG files updated.

chalasr (approved):

fabpot:
Thank you @nicolas-grekas.

PR #50298, opened 2023-05-11
[DependencyInjection] Drop IO in FileLoader

staabm:
| Q             | A
| ------------- | ---
| Branch?       | 6.3 
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #... 
| License       | MIT

Defer IO as long as possible.

I am not sure whether its a important change, but at least I could see `$info->isReadable()` showing up in our blackfire profiles in a hot path

grafik

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Loader/FileLoader.php
+++ b/src/Symfony/Component/DependencyInjection/Loader/FileLoader.php
@@ -262,7 +262,7 @@ abstract class FileLoader extends BaseFileLoader
                 continue;
             }
 
-            if (!str_ends_with($path, '.php') || !$info->isReadable()) {
+            if (!str_ends_with($path, '.php')) {
                 continue;
             }
             $class = $namespace.ltrim(str_replace('/', '\\', substr($path, $prefixLen, -4)), '\\');


src/Symfony/Component/DependencyInjection/Loader/FileLoader.php
```diff
@@ -94,7 +94,6 @@ public function import(mixed $resource, string $type = null, bool|string $ignore
      * @param string               $namespace The namespace prefix of classes in the scanned directory
      * @param string               $resource  The directory to look for classes, glob-patterns allowed
      * @param string|string[]|null $exclude   A globbed path of files to exclude or an array of globbed paths of files to exclude
-     * @param string|null          $source    The path to the file that defines the auto-discovery rule
```
staabm: fabbot suggested this change

nicolas-grekas: but it should be ignored :)

staabm: :)

src/Symfony/Component/DependencyInjection/Loader/FileLoader.php
```diff
@@ -271,6 +270,10 @@ private function findClasses(string $namespace, string $pattern, array $excludeP
                 continue;
             }
 
+            if (!$info->isReadable()) {
```
nicolas-grekas: what about removing the call altogether? I don't think it's needed in practice

staabm: great suggestion. I tried it locally and it worked for me. 

grafik

staabm:
From the CI output I don't see whether the build error is related to the changes or not

nicolas-grekas:
Thank you @staabm.

PR #50296, opened 2023-05-10
[Notifier] Add Brevo bridge (formerly Sendinblue)

PEtanguy:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | TODO

Hello,

This PR is aimed at updating the config for the sendinblue notifier.
As you might have seen, Sendinblue has rebranded to [Brevo](https://developers.brevo.com/) and also rewrote their apis.  
This change ensure compatibility with the new endpoints and removes any reference to Sendinblue.

This is the mailer PR: https://github.com/symfony/symfony/pull/50302

PR #50302, opened 2023-05-11
[Mailer] New Brevo mailer bridge (formerly Sendinblue)

PEtanguy:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | TODO

Hello,

This PR is aimed at updating the config for the sendinblue mailer.
As you might have seen, Sendinblue has rebranded to [Brevo](https://developers.brevo.com/) and also rewrote their apis.  
This change ensure compatibility with the new endpoints and removes any reference to Sendinblue.

This is the notifier PR: https://github.com/symfony/symfony/pull/50296

### DIFF
--- a/src/Symfony/Component/Mailer/Bridge/Brevo/CHANGELOG.md
+++ b/src/Symfony/Component/Mailer/Bridge/Brevo/CHANGELOG.md
@@ -1,0 +1,7 @@
+CHANGELOG
+=========
+
+6.4
+---
+
+* Added the bridge as a replacement of the deprecated Sendinblue one.
--- a/src/Symfony/Component/Mailer/Bridge/Brevo/README.md
+++ b/src/Symfony/Component/Mailer/Bridge/Brevo/README.md
@@ -1,0 +1,56 @@
+Brevo Bridge
+============
+
+Provides Brevo integration for Symfony Mailer.
+This was added uppon Sendinblue's rebranding to Brevo.
+
+Configuration example:
+
+```env
+# SMTP
+MAILER_DSN=brevo+smtp://USERNAME:PASSWORD@default
+
+# API
+MAILER_DSN=brevo+api://KEY@default
+```
+
+where:
+- `KEY` is your Brevo API Key
+
+With API, you can use custom headers.
+
+```php
+$params = ['param1' => 'foo', 'param2' => 'bar'];
+$json = json_encode(['"custom_header_1' => 'custom_value_1']);
+
+$email = new Email();
+$email
+    ->getHeaders()
+    ->add(new MetadataHeader('custom', $json))
+    ->add(new TagHeader('TagInHeaders1'))
+    ->add(new TagHeader('TagInHeaders2'))
+    ->addTextHeader('sender.ip', '1.2.3.4')
+    ->addTextHeader('templateId', 1)
+    ->addParameterizedHeader('params', 'params', $params)
+    ->addTextHeader('foo', 'bar')
+;
+```
+
+This example allow you to set :
+
+* templateId
+* params
+* tags
+* headers
+    * sender.ip
+    * X-Mailin-Custom
+
+For more informations, you can refer to [Brevo API documentation](https://developers.brevo.com/reference/sendtransacemail).
+
+Resources
+---------
+
+* [Contributing](https://symfony.com/doc/current/contributing/index.html)
+* [Report issues](https://github.com/symfony/symfony/issues) and
+  [send Pull Requests](https://github.com/symfony/symfony/pulls)
+  in the [main Symfony repository](https://github.com/symfony/symfony)
--- a/src/Symfony/Component/Mailer/Bridge/Brevo/Transport/BrevoApiTransport.php
+++ b/src/Symfony/Component/Mailer/Bridge/Brevo/Transport/BrevoApiTransport.php
@@ -9,7 +9,7 @@
  * file that was distributed with this source code.
  */
 
-namespace Symfony\Component\Mailer\Bridge\Sendinblue\Transport;
+namespace Symfony\Component\Mailer\Bridge\Brevo\Transport;
 
 use Psr\EventDispatcher\EventDispatcherInterface;
 use Psr\Log\LoggerInterface;
@@ -28,11 +28,9 @@ use Symfony\Contracts\HttpClient\HttpClientInterface;
 use Symfony\Contracts\HttpClient\ResponseInterface;
 
 /**
- * @author Yann LUCAS
- *
- * @deprecated since Symfony 6.3, use BrevoApiTransport instead
+ * @author Pierre TANGUY
  */
-final class SendinblueApiTransport extends AbstractApiTransport
+final class BrevoApiTransport extends AbstractApiTransport
 {
     private string $key;
 
@@ -45,7 +43,7 @@ final class SendinblueApiTransport extends AbstractApiTransport
 
     public function __toString(): string
     {
-        return sprintf('sendinblue+api://%s', $this->getEndpoint());
+        return sprintf('brevo+api://%s', $this->getEndpoint());
     }
 
     protected function doSendApi(SentMessage $sentMessage, Email $email, Envelope $envelope): ResponseInterface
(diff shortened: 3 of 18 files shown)


src/Symfony/Component/Mailer/Bridge/Brevo/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+CHANGELOG
+=========
+
+6.3
+---
+
+* Added the bridge as a replacement of the deprecated Sendiblue one.
```
stof:
```suggestion
* Added the bridge as a replacement of the deprecated Sendinblue one.
```

src/Symfony/Component/Mailer/Bridge/Brevo/README.md
```diff
@@ -0,0 +1,55 @@
+Brevo Bridge
+=================
```
stof:
```suggestion
============
```

src/Symfony/Component/Mailer/Bridge/Brevo/Transport/BrevoApiTransport.php
```diff
@@ -0,0 +1,188 @@
...
+
+                continue;
+            }
+            $headersAndTags['headers'][$header->getName()] = $header->getBodyAsString();
+        }
+
+        return $headersAndTags;
+    }
+
+    private function stringifyAddress(Address $address): array
```
stof: the name `stringifyAddress` is weird as it does not turn them into a string

src/Symfony/Component/Mailer/Bridge/Brevo/Transport/BrevoApiTransport.php
```diff
@@ -0,0 +1,188 @@
...
+            throw new HttpTransportException('Unable to send an email: '.$result['message'].sprintf(' (code %d).', $statusCode), $response);
+        }
+
+        $sentMessage->setMessageId($result['messageId']);
+
+        return $response;
+    }
+
+    /**
+     * @return array<int, string>
```
stof: the return type is actually `list<array{email: string, name?: string}>`

src/Symfony/Component/Mailer/Bridge/Brevo/Transport/BrevoApiTransport.php
```diff
@@ -0,0 +1,188 @@
...
+     */
+    protected function stringifyAddresses(array $addresses): array
```
stof: However the protected method of the AbstractTransport with a different return type is a bad idea. This should use a different method name for the needs of this class

src/Symfony/Component/Mailer/Bridge/Brevo/composer.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "symfony/brevo-mailer",
```
stof: Our composer.json files are using 4 spaces for indentation

src/Symfony/Component/Mailer/Bridge/Brevo/composer.json
```diff
@@ -0,0 +1,35 @@
...
+  "type": "symfony-mailer-bridge",
+  "description": "Symfony Brevo Mailer Bridge - formerly Sendinblue",
```
stof: I don't think we need `formerly Sendinblue` in the package description

src/Symfony/Component/Mailer/Bridge/Sendinblue/composer.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "symfony/sendinblue-mailer",
     "type": "symfony-mailer-bridge",
-    "description": "Symfony Sendinblue Mailer Bridge",
+    "description": "Symfony Sendinblue Mailer Bridge - deprecated",
```
stof: this should be reverted. We don't use the package description for that

stof (changes requested):

src/Symfony/Component/Mailer/Bridge/Brevo/Transport/BrevoApiTransport.php
PEtanguy: formatAddresses() ?

stof: this looks like a good name

stof: @PEtanguy you haven't replaced this method

src/Symfony/Component/Mailer/Bridge/Brevo/Transport/BrevoApiTransport.php
```diff
@@ -76,7 +76,7 @@ protected function doSendApi(SentMessage $sentMessage, Email $email, Envelope $e
     /**
      * @return list<array{email: string, name?: string}>
      */
-    protected function stringifyAddresses(array $addresses): array
+    protected function formatAddresses(array $addresses): array
```
stof: should be a private method as it does not try to overwrite the parent method anymore.

PEtanguy: Good point

PEtanguy:
I will make the documentation once this is merged. I also noticed an error in the doc of another transport.

src/Symfony/Component/Mailer/Bridge/Brevo/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+CHANGELOG
+=========
+
+6.3
+---
+
+* Added the bridge as a replacement of the deprecated Sendinblue one.
```
OskarStark, line 7:
```suggestion
* Add the bridge
```

src/Symfony/Component/Mailer/Bridge/Brevo/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+CHANGELOG
+=========
+
+6.3
```
OskarStark:
```suggestion
6.4
```

OskarStark:
I would not touch the other bridge in this PR; let's just add this as a new bridge. No need to deprecate the code IMHO, we can abandon the whole package via packagist.

WDYT @fabpot ?

stof:
Well, triggering a deprecation when someone configures a mailer DSN using the `sendinblue` schemes is needed to help them migrate.

If we only abandon the package, projects removing the old bridge without updating their env variables in prod would get a broken project due to using an unsupported scheme.
Side note: I would even wait a bit before marking the package as abandoned as FrameworkBundle 5.4 and 6.3 won't support the Brevo bridge (it does not register it as a supported factory) while the SendinblueBridge technically still works thanks to the BC layer of the API itself (where the old domain still works). We cannot mark a package as abandoned only for some versions and not for others.

PEtanguy:
I agree with @stof , we can abandon Sendinblue later. I will keep the deprecations on this PR and on the one of the notifier (https://github.com/symfony/symfony/pull/50296)

stof (approved):

src/Symfony/Component/Mailer/Bridge/Brevo/CHANGELOG.md
stof: @PEtanguy you forgot this change when fixing requested things

fabpot:
Thank you @PEtanguy.

thomas2411:
Hello, I have received information from Brevo that they will move to Brevo domain on 31 of August. Symfony 6.4 is planned for 2024 so is there a chance to get it within a month from now?

> 
> Update your SMTP relay domain from smtp-relay.sendinblue.com to smtp-relay.brevo.com. 
>  Please do so before August 30, 2023 to ensure your sending is not impacted.

thomas2411:
> Side note: I would even wait a bit before marking the package as abandoned as FrameworkBundle 5.4 and 6.3 won't support the Brevo bridge (it does not register it as a supported factory) while the SendinblueBridge technically still works thanks to the BC layer of the API itself (where the old domain still works). We cannot mark a package as abandoned only for some versions and not for others.

@stof Can you please see https://github.com/symfony/symfony/pull/50302#issuecomment-1645227961 ? I think we need to add this within two weeks.

xabbuh:
#51295 should be enough, right?

thomas2411:
> #51295 should be enough, right?

Yes, thank you :)

PR #50306, opened 2023-05-12
[DomCrawler][FrameworkBundle] Add `assertAnySelectorText*`

### LINKED ISSUE #48213: [FrameworkBundle] Test `assertAnySelectorText*` functions
### Description

In a functional test, it is useful to check that a specific text is present on an HTML element. For example, at the end of a controller action test, an assertion could be made to check that an entity property has been modified.
The current `assertSelectorTextContains` and `assertSelectorTextSame` only check the first occurence of a given selector. It is a problem when the element is in an unknown position.

I would like to know if others already had the same issue, and if this 2 functions would be useful for the community (`assertAnySelectorTextContains` and `assertAnySelectorTextSame`).

### Example

A simple example with `ul li` selector

```html
<ul>
    <li>test1</li>
    <li>test2</li>
    <li>test3</li>
</ul>
```

```php
// If the node is at first position
public function testIndexAction(): void
{
    $client = self::createClient();
    $client->request('GET', '/');

    self::assertResponseIsSuccessful();
    self::assertSelectorTextContains('ul li', 'test1');
}

// If the node position is known
public function testIndexAction2(): void
{
    $client = self::createClient();
    $client->request('GET', '/');

    self::assertResponseIsSuccessful();
    self::assertSelectorTextContains('ul li:nth-child(2)', 'test2');
}

// If the node position is unknown
public function testIndexAction3(): void
{
    $client = self::createClient();
    $crawler = $client->request('GET', '/');

    $this->assertResponseIsSuccessful();
    $nodes = $crawler->filter('ul li')->each(fn(Crawler $node) => $node->text());

    self::assertResponseIsSuccessful();
    self::assertContains('test3', $nodes);
}
```

SVillette:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #48213
| License       | MIT
| Doc PR        | -

As mentioned in #48213, it can be useful to check that any selector contains/equals a string. But the current `assertSelectorTextSame` and `assertSelectorTextContains` only assert with the first selector.

Before this PR, the solution was 
```html
<ul>
    <li>test1</li>
    <li>test2</li>
    <li>test3</li>
</ul>
```
```php
public function testIndexAction(): void
{
    $client = self::createClient();
    $crawler = $client->request('GET', '/');

    $nodes = $crawler->filter('ul li')->each(fn(Crawler $node) => $node->text());
    self::assertContains('test3', $nodes);
}
```

And after this changes

```php
public function testIndexAction(): void
{
    $client = self::createClient();
    $crawler = $client->request('GET', '/');

    self::assertAnySelectorTextSame('ul li', 'test3');
    self::assertAnySelectorTextContains('ul li', 'tes');
    self::assertAnySelectorTextNotContains('ul li', 'foo');
}
```

A doc PR will be made if the changes are accepted

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
+++ b/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
@@ -7,6 +7,9 @@ CHANGELOG
  * Add native return type to `Translator` and to `Application::reset()`
  * Deprecate the integration of Doctrine annotations, either uninstall the `doctrine/annotations` package or disable the integration by setting `framework.annotations` to `false`
  * Enable `json_decode_detailed_errors` context for Serializer by default if `kernel.debug` is true and the `seld/jsonlint` package is installed
+ * Add `DomCrawlerAssertionsTrait::assertAnySelectorTextContains(string $selector, string $text)`
+ * Add `DomCrawlerAssertionsTrait::assertAnySelectorTextSame(string $selector, string $text)`
+ * Add `DomCrawlerAssertionsTrait::assertAnySelectorTextNotContains(string $selector, string $text)`
 
 6.3
 ---
--- a/src/Symfony/Bundle/FrameworkBundle/Test/DomCrawlerAssertionsTrait.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Test/DomCrawlerAssertionsTrait.php
@@ -47,6 +47,14 @@ trait DomCrawlerAssertionsTrait
         ), $message);
     }
 
+    public static function assertAnySelectorTextContains(string $selector, string $text, string $message = ''): void
+    {
+        self::assertThat(self::getCrawler(), LogicalAnd::fromConstraints(
+            new DomCrawlerConstraint\CrawlerSelectorExists($selector),
+            new DomCrawlerConstraint\CrawlerAnySelectorTextContains($selector, $text)
+        ), $message);
+    }
+
     public static function assertSelectorTextSame(string $selector, string $text, string $message = ''): void
     {
         self::assertThat(self::getCrawler(), LogicalAnd::fromConstraints(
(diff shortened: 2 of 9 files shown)


SVillette:
The failing tests seems unrelated to this PR (related to the Notifier component).
However, Psalm fails with the 2 same errors:
- The first one is because phpunit defines method `\PHPUnit\Framework\Constraint\Constraint::matches($other)` and the signature of the first argument is changed (renamed to `$crawler`). Should I fix this error as all other constraints have the same naming ?
- The second error happens because of the type of `$crawler` argument. As the previous error, all other constraints have the same signature on `matches()` method. 

What is the correct fix here ? Should this errors be added to the Psalm baseline ? 
Thanks

stof:
Instead of `@param Crawler`, the argument should have the same type than in the parent class, with a first check inside the class about unsupported type for the asserted value.

SVillette:
> Instead of `@param Crawler`, the argument should have the same type than in the parent class, with a first check inside the class about unsupported type for the asserted value.

Sure, I get what you meant here.
My question was about consistency with the other `Constraint` classes. For example, the class [CrawlerSelectorCount](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/DomCrawler/Test/Constraint/CrawlerSelectorCount.php) has been added "recently" (5 months ago) and all methods of this class have the same signature than the methods of the constraints I added in this PR. I think other constraints throw Psalm errors too but this errors have been added to the baseline.

stof:
Our CI generates a baseline automatically for each run, so that it complains only about new issues.

SVillette:
Ok, so I guess you want it to be fixed for this PR and another PR can fix it for the other constraints ?

stof:
indeed

SVillette:
@stof Psalm errors are fixed now, let me now if you want me to make other changes.

src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
```diff
@@ -22,6 +22,9 @@ CHANGELOG
  * Deprecate the `Http\Client\HttpClient` service, use `Psr\Http\Client\ClientInterface` instead
  * Add `stop_worker_on_signals` configuration option to `messenger` to define signals which would stop a worker
  * Add support for `--all` option to clear all cache pools with `cache:pool:clear` command
+ * Add `DomCrawlerAssertionsTrait::assertAnySelectorTextContains(string $selector, string $text)`
+ * Add `DomCrawlerAssertionsTrait::assertAnySelectorTextSame(string $selector, string $text)`
+ * Add `DomCrawlerAssertionsTrait::assertAnySelectorTextNotContains(string $selector, string $text)`
```
stof: Those should go in a new `6.4` section. The feature won't make it into 6.3 as the feature freeze was beginning of March (and the stable release is due for the end of the month, i.e. in one week from now)

SVillette: Sure, will do the update and change the base branch when the branch  `6.4` will be created

fabpot (approved):

fabpot:
Thank you @SVillette.

PR #50308, opened 2023-05-12
[Notifier] Bring consistency to bridges

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -


src/Symfony/Component/Notifier/Bridge/ContactEveryone/ContactEveryoneOptions.php
```diff
@@ -25,61 +25,33 @@ public function __construct(array $options = [])
...
     {
-        return $this->options['recipient_id'] ?? null;
+        return null;
     }
 
-    public function setDiffusionName(int $diffusionName): self
+    /**
+     * @return $this
+     */
+    public function diffusionName(int $diffusionName): static
```
nicolas-grekas, line 36: it's a bit weird to me that an option named "diffusionName" only accepts integers
can you please confirm this is correct @gnito-org?

nicolas-grekas:
The TL;DR of this PR is that options shouldn't have getters when they're not required, and their setters should not start with `set*`.
The options should also not provide yet another way to define what the messages can already convey (from, recipient, etc.), unless there is a compelling reason to. (There might be existing option classes that violate this rule, but that shouldn't be a reason to add more.)

The rest is more CS-related.

/cc @OskarStark @alamirault since you contribute to this component. :pray:

src/Symfony/Component/Notifier/Bridge/Bandwidth/BandwidthTransport.php
```diff
@@ -60,31 +60,26 @@ protected function doSend(MessageInterface $message): SentMessage
...
-
-        if (!isset($options['priority']) && $this->priority) {
-            $options['priority'] = $this->priority;
-        }
-
-        if (!preg_match('/^\+[1-9]\d{1,14}$/', $this->from)) {
-            throw new InvalidArgumentException(sprintf('The "From" number "%s" is not a valid phone number. The number must be in E.164 format.', $this->from));
+        $options['from'] = $message->getFrom() ?: $this->from;
+        $options['to'] = [$message->getPhone()];
+        $options['accountId'] ??= $this->accountId;
```
OskarStark, line 67: Before we used account_id kn the request, now accountId, is it correct or did I overlooked sth?

src/Symfony/Component/Notifier/Bridge/Clickatell/ClickatellOptions.php
```diff
@@ -1,61 +0,0 @@
...
- */
-
-namespace Symfony\Component\Notifier\Bridge\Clickatell;
-
-use Symfony\Component\Notifier\Message\MessageOptionsInterface;
-
-/**
- * @author gnito-org <https://github.com/gnito-org>
- */
-final class ClickatellOptions implements MessageOptionsInterface
```
OskarStark, line 19: Why removing?

OskarStark: Ah because it does not really have an options.

We can remove it because it was merged and not released right now, correct?

fabpot: Correct

src/Symfony/Component/Notifier/Bridge/Bandwidth/BandwidthTransport.php
nicolas-grekas: The option array should be aligned with the API. The API uses `accountId`, that's why.

nicolas-grekas:
I'm merging to fix the CI and because this is ready to me, but feel free to comment after the merge if needed!

PR #50309, opened 2023-05-12
[HttpFoundation] UrlHelper is now aware of RequestContext changes

giosh94mhz:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | yes
| Tickets       | 
| License       | MIT
| Doc PR        |

`RequestContext` in routing can change at runtime after the `UrlHelper` has been created, so using a `RequestContextAwareInterface` instance (i.e. the router) will workaround URL generation issues.

In my application I have a CLI command that dynamically create the RequestContext, based on database queries and other user-driven configuration. With this information at hand, I retrieve the router and set the new context that should then be used on any other URL generation.

This works for the Router, but then I render some twig templates (to send formatted email) and found that the `UrlHelper` is using the old context, and not helping at all :) Also, the helper is final and without interface, so I have no way of decorating or replacing it.

Long story short: in my opinion, the `RequestContext` should be treated just like the `Request`, and so we should use `RequestContextAwareInterface` just like we use the `RequestStack`.

....or maybe the context should be carried out with the `RequestStack`? Any other ideas?

~~no longer relevant: this PR introduce a BC on the constructor of `UrlHelper`, but since it's final maybe is not required to keep backward compatility. If not, I may update the class to support both `RequestContextAwareInterface` and `RequestContext` as the second parameter.~~

### DIFF
--- a/src/Symfony/Component/HttpFoundation/UrlHelper.php
+++ b/src/Symfony/Component/HttpFoundation/UrlHelper.php
@@ -12,6 +12,7 @@
 namespace Symfony\Component\HttpFoundation;
 
 use Symfony\Component\Routing\RequestContext;
+use Symfony\Component\Routing\RequestContextAwareInterface;
 
 /**
  * A helper service for manipulating URLs within and outside the request scope.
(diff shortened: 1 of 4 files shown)


src/Symfony/Component/HttpFoundation/UrlHelper.php
```diff
@@ -21,12 +21,12 @@
 final class UrlHelper
 {
     private $requestStack;
-    private $requestContext;
+    private $contextAware;
 
-    public function __construct(RequestStack $requestStack, RequestContext $requestContext = null)
+    public function __construct(RequestStack $requestStack, RequestContextAwareInterface $contextAware = null)
```
stof: For BC reasons, it is mandatory to keep support for passing a RequestContext. Otherwise, this cannot be accepted

stof:
NOTE: this PR introduce a BC on the constructor of UrlHelper, but since it's final maybe is not required to keep backward compatility.

A final constructor is not an internal one. You can still write code than instantiate the class, and would be impacted by the BC break. Constructors are not only called as `parent::__construct()`

giosh94mhz:
Thank you for the feedback.

> A final constructor is not an internal one. You can still write code than instantiate the class, and would be impacted by the BC break. Constructors are not only called as `parent::__construct()`

Yes, I'm very well aware, I meant that I didn't know if you care of BC of a final helper class. Since you do, I've implemented the compatibility with the deprecation and all.

stof:
https://symfony.com/bc has a good overview of the Symfony BC policy, including what kind of changes you are allowed to do as a contributor.

nicolas-grekas (approved):
(I updated the patch to make it suitable for 5.4)

nicolas-grekas:
Thank you @giosh94mhz.

giosh94mhz:
Thank to you guys, for all your hard work!

Issue #44211, opened 2021-11-23
Support Enum as workflow places


stof:
### Discussed in https://github.com/symfony/symfony/discussions/44203

<div type='discussions-op-text'>

<sup>Originally posted by **stephanvierkant** November 22, 2021</sup>
I've tried the new Enum in PHP 8.1 and I love that Symfony supports this new feature in the Form component.

I'd love to have this support in the Workflow component as well. 

I've changed my status attribute from a string to an Enum field and now this doesn't work anymore:

`$workflow->can($entity, $transition->getName()`

I'd be great is Symfony would support Enum in the Workflow Component as well! </div>

alexandre-daubois:
Looking into it. :+1: 
Started to work on 6.0 branch while waiting 6.1 branch (as mentioned in the original issue, this feature is most likely to be included in 6.1). It's coming up together, trying to bring support of both `UnitEnum` and `BackedEnum`. Adding final tests right now.

sxsws:
Generally, this is the problem of another component other than the workflow component.

I have a working example of this for yaml. It's relatively simple to monkey-patch.

At the moment, I have this example working:

My enum:

```php
enum Suit: string
{
    case Clubs = 'c';
    case Hearts = 'h';
    case Spades = 's';
    case Diamonds = 'd';
}
```

and part of my workflow.

```yaml
places:
        !php/enum App\Enum\Suit::Hearts.name:
            metadata:
              - description: !php/const App\Enum\Suit::Spades

...
transitions:
    to_review:
        from: !php/enum App\Enum\Suit::Hearts.name
        to:   reviewed
```

The place name is the name string of the enum `Hearts`. If I change it to `.value`, it'd give, from above, `h`.
The value for description gives the instance of Suit::Spades just for fun.

Of course, this is only for yaml. It will of course work outside of the workflow component

stof:
@sxsws what you are suggesting here is **not** enum support for the workflow component. In your proposal, place names are still arbitrary strings, meaning that there is no benefit of using an enum for that at all.
and there is no need to spend time on implementing your `!php/enum` proposal for the Yaml component (for which the current ticket is the wrong to discuss it). I would immediately give it a -1 vote as it totally defeats the points of PHP enums.

sxsws:
and what then pray tell is the difference then between an enum and a class of constants for workflow?

stof:
[...]

PR #50311, opened 2023-05-12
Supported "Marking" enum php

Fuck4ik:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

I would like to be able to use the getters and setters of my entity with the enum type without custom implements `MarkingStoreInterface`.

Because php does not support enum in associative arrays, and `\WeakMap` is not suitable for this case, it was decided to make a method `getEnumPlaces`

### DIFF
--- a/src/Symfony/Component/Workflow/Marking.php
+++ b/src/Symfony/Component/Workflow/Marking.php
@@ -22,6 +22,9 @@ class Marking
      * @var array<string, int>
      */
     private array $places = [];
+    /**
+     * @var array<string, \UnitEnum>
+     */
     private array $enumPlaces = [];
     private ?array $context = null;
 
@@ -37,7 +40,7 @@ class Marking
 
     public function mark(string|\UnitEnum $place): void
     {
-        $key = $this->placeToKey($place);
+        $key = $this->enumOrStringToKey($place);
         $this->places[$key] = 1;
         if ($place instanceof \UnitEnum) {
             $this->enumPlaces[$key] = $place;
@@ -46,7 +49,7 @@ class Marking
 
     public function unmark(string|\UnitEnum $place): void
     {
-        $key = $this->placeToKey($place);
+        $key = $this->enumOrStringToKey($place);
         unset(
             $this->places[$key],
             $this->enumPlaces[$key]
@@ -55,7 +58,7 @@ class Marking
 
     public function has(string|\UnitEnum $place): bool
     {
-        return isset($this->places[$this->placeToKey($place)]);
+        return isset($this->places[$this->enumOrStringToKey($place)]);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


stof:
Places are strings in the workflow component. See https://github.com/symfony/symfony/issues/44211 for the discussion about the support for enums in that component.

src/Symfony/Component/Workflow/Marking.php
```diff
@@ -31,36 +35,43 @@ public function __construct(array $representation = [])
         }
     }
 
-    /**
-     * @return void
-     */
-    public function mark(string $place)
+    public function mark(string|\UnitEnum $place): void
```
derrabus, line 41: Note that this method is not final and could be overridden downstream. Widening the parameter type and narrowing the native return type are both breaking changes.

derrabus (changes requested):
Thank you for your attempt to improve Symfony. Please implement all changes in a backwards-compatible manner. Here's an overview of what you can and cannot change: https://symfony.com/doc/current/contributing/code/bc.html#working-on-symfony-code

Fuck4ik:
> Places are strings in the workflow component. See #44211 for the discussion about the support for enums in that component.

Thank you, I have read. Nevertheless, the desire to use enum is still present, and this decision turned out to be not so cumbersome and critical.

src/Symfony/Component/Workflow/Marking.php
Fuck4ik: I propose to make Marking.php the final class

wouterj: We cannot do that now, as making a class final is also a breaking change. So we can at most flag it as final using PHPdoc and make it final in 7.0. But that would also mean that we probably need to introduce a `MarkingInterface` and update the rest of the component to use that typehint, as otherwise this is fully closed for extension.

stof:
"Desire" to use a PHP feature in your own code does not exempt your contribution from respecting required rules of all contributions submitted to Symfony (as we cannot accept it if it breaks BC)

src/Symfony/Component/Workflow/Marking.php
Fuck4ik: Yes, you're right, it's better to leave it like that than to make many changes in favor of MarkingInterface

Fuck4ik:
> "Desire" to use a PHP feature in your own code does not exempt your contribution from respecting required rules of all contributions submitted to Symfony (as we cannot accept it if it breaks BC)

I understand you and completely agree. Nevertheless, this feature will be useful to many users.
In modern development, few people use enums as strings anymore. And this problem has to be solved with crutches like two properties "status" and "statusAsString"

fabpot:
Closing as we don't merge PRs with BC breaks.

PR #50312, opened 2023-05-13
[Security] Skip clearing CSRF Token on stateless logout

### LINKED ISSUE #50310: `CsrfTokenClearingLogoutListener` trying to clear storage on logout from stateless firewall

6.3

### Description

We have a functional test that covers checking login and logout through our internal API firewall to make sure tokens generated with [`LexikJWTAuthenticationBundle`](https://github.com/lexik/LexikJWTAuthenticationBundle) and [`JWTRefreshTokenBundle`](https://github.com/markitosgv/JWTRefreshTokenBundle) are properly invalidated.  As this API only uses token authentication, the full firewall is configured as stateless.  Rather basic test scenario:

```php
<?php

use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;

final class AuthTest extends WebTestCase
{
    public function testLogoutTokenInvalidation(): void
    {
        $client = self::createClient();

        // Load database fixtures

        $client->jsonRequest(
            'POST',
            '/api/internal/login',
            ['username' => 'user1@app.local', 'password' => 'password'],
        );

        $this->assertResponseIsSuccessful('Could not authenticate to the internal API.');

        $response = $client->getResponse();

        if ($response->getContent() === false) {
            $this->fail('The API did not send a proper response.');
        }

        $data = json_decode($response->getContent(), true, flags: \JSON_THROW_ON_ERROR);

        if (!isset($data['refresh_token'])) {
            $this->fail('The refresh token was not included in the authentication response.');
        }

        $client->jsonRequest(
            'POST',
            '/api/internal/logout',
            ['refresh_token' => $data['refresh_token']],
            ['HTTP_Authorization' => sprintf('Bearer %s', $data['token'])],
        );

        $this->assertResponseIsSuccessful();

        // Check token validity
    }
}
```

This test case passes on 6.2 but fails with 6.3 beta

```sh
Failed asserting that the Response is successful.
HTTP/1.1 500 Internal Server Error
Cache-Control: max-age=0, must-revalidate, private
Content-Type:  application/problem+json
Date:          Fri, 12 May 2023 19:57:13 GMT
Expires:       Fri, 12 May 2023 19:57:13 GMT
Vary:          Accept
X-Robots-Tag:  noindex

{"type":"https:\/\/tools.ietf.org\/html\/rfc2616#section-10","title":"An error occurred","status":500,"detail":"Session was used while the request was declared stateless."}
 /app/vendor/symfony/framework-bundle/Test/BrowserKitAssertionsTrait.php:142
 /app/vendor/symfony/framework-bundle/Test/BrowserKitAssertionsTrait.php:33
 /app/tests/functional/AuthTest.php:175
```

After debugging, the session use is being logged from a call inside the `Symfony\Component\Security\Http\EventListener\CsrfTokenClearingLogoutListener::onLogout()` event listener.

### How to reproduce

Logout from a stateless firewall (haven't taken the time to get down to a smaller reproducing scenario)

### Possible Solution

_No response_

### Additional Context

_No response_

chalasr:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50310 
| License       | MIT
| Doc PR        | -

Not targeting 5.4 LTS as the bug is only breaking on 6.3 although it does exist on prior versions.

### DIFF
--- a/src/Symfony/Component/Security/Http/EventListener/CsrfTokenClearingLogoutListener.php
+++ b/src/Symfony/Component/Security/Http/EventListener/CsrfTokenClearingLogoutListener.php
@@ -13,6 +13,7 @@ namespace Symfony\Component\Security\Http\EventListener;
 
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\Security\Csrf\TokenStorage\ClearableTokenStorageInterface;
+use Symfony\Component\Security\Csrf\TokenStorage\SessionTokenStorage;
 use Symfony\Component\Security\Http\Event\LogoutEvent;
 
 /**
@@ -31,6 +32,10 @@ class CsrfTokenClearingLogoutListener implements EventSubscriberInterface
 
     public function onLogout(LogoutEvent $event): void
     {
+        if ($this->csrfTokenStorage instanceof SessionTokenStorage && !$event->getRequest()->hasPreviousSession()) {
+            return;
+        }
+
         $this->csrfTokenStorage->clear();
     }
 


src/Symfony/Component/Security/Http/EventListener/CsrfTokenClearingLogoutListener.php
```diff
@@ -31,6 +32,10 @@ public function __construct(ClearableTokenStorageInterface $csrfTokenStorage)
 
     public function onLogout(LogoutEvent $event): void
     {
+        if ($this->csrfTokenStorage instanceof SessionTokenStorage && !$event->getRequest()->hasPreviousSession()) {
```
fabpot, line 35: What about fixing it in `SessionTokenStorage` instead?

chalasr: Using `SessionTokenStorage` without a session has been deprecated in 5.x: https://github.com/symfony/symfony/blob/8c637a5bbfea8b322ec93eb51ac47c0db8948e2a/src/Symfony/Component/Security/Csrf/TokenStorage/SessionTokenStorage.php#L95-L108

Ideally this listener shouldn't be registered for stateless firewalls, problem is that it's not a per-firewall listener but a global one. We should probably change that in another (feature) PR.

nicolas-grekas:
Any way to test this?

nicolas-grekas:
Thank you @chalasr.

chalasr:
> Any way to test this?

Sure, at least something preventing regressions. I'll do!

PR #50315, opened 2023-05-13
[Translation] Fix handling of null messages in `ArrayLoader`

### LINKED ISSUE #49894: [Translation] `MessageCatalogue::add()` accepts non-string messages

rob006:
| Q             | A
| ------------- | ---
| Branch?       |  5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #49894
| License       | MIT
| Doc PR        | -

Replaces https://github.com/symfony/symfony/pull/50252

### DIFF
--- a/src/Symfony/Component/Translation/Loader/ArrayLoader.php
+++ b/src/Symfony/Component/Translation/Loader/ArrayLoader.php
@@ -46,9 +46,11 @@ class ArrayLoader implements LoaderInterface
         foreach ($messages as $key => $value) {
             if (\is_array($value)) {
                 foreach ($this->flatten($value) as $k => $v) {
-                    $result[$key.'.'.$k] = $v;
+                    if (null !== $v) {
+                        $result[$key.'.'.$k] = $v;
+                    }
                 }
-            } else {
+            } elseif (null !== $value) {
                 $result[$key] = $value;
             }
         }
--- a/src/Symfony/Component/Translation/Tests/Loader/YamlFileLoaderTest.php
+++ b/src/Symfony/Component/Translation/Tests/Loader/YamlFileLoaderTest.php
@@ -30,6 +30,15 @@ class YamlFileLoaderTest extends TestCase
         $this->assertEquals([new FileResource($resource)], $catalogue->getResources());
     }
 
+    public function testLoadNonStringMessages()
+    {
+        $loader = new YamlFileLoader();
+        $resource = __DIR__.'/../fixtures/non-string.yml';
+        $catalogue = $loader->load($resource, 'en', 'domain1');
+
+        $this->assertSame(['root.foo2' => '', 'root.bar' => 'bar'], $catalogue->all('domain1'));
+    }
+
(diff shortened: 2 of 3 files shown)


src/Symfony/Component/Translation/Loader/ArrayLoader.php
```diff
@@ -46,10 +46,12 @@ private function flatten(array $messages): array
...
                 foreach ($this->flatten($value) as $k => $v) {
-                    $result[$key.'.'.$k] = $v;
+                    if (null !== $v) {
+                        $result[$key.'.'.$k] = (string) $v;
+                    }
                 }
-            } else {
-                $result[$key] = $value;
+            } elseif (null !== $value) {
+                $result[$key] = (string) $value;
```
nicolas-grekas: I think we should remove the cast to string, so that stringable objects aren't resolved eagerly, if that makes sense

rob006: I don't think that `MessageCatalogue` expects non-string input - in 6.x line there are already types that would force cast to string in `get()` and `set()`. If performance is concern, then cache should be used, and stringable objects here would only be a problem in that case.

nicolas-grekas: Still: your PR is motivated by nulls, not by non-strings. Since this is 5.4, we'd better not change the behavior unless we have a compelling reason to.

nicolas-grekas:
Thank you @rob006.

PR #50317, opened 2023-05-14
[Console] Remove redundant method getSaturation()

ivanpepelko:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | none
| License       | MIT
| Doc PR        | none

1. in `getSaturation()`, `$diff` is compared against `int` zero. This will only yield `true` when RGB values contain min/max color value (`0` or `255`). It will be always `false` for floats.
2. in `degradeHexColorToAnsi4()` value returned from `getSaturation()` is `round`ed, which returns float, thus making comparison to `int` zero useless.
3. in the same method, bitwise operations (final `return` statement) will convert number to `int` anyway, so the explicit cast was removed

In conclusion, `getSaturation` is redundant. Tests are expanded to show that even in min/max cases everything works fine.

### DIFF
--- a/src/Symfony/Component/Console/Output/AnsiColorMode.php
+++ b/src/Symfony/Component/Console/Output/AnsiColorMode.php
@@ -78,25 +78,7 @@ enum AnsiColorMode
 
     private function degradeHexColorToAnsi4(int $r, int $g, int $b): int
     {
-        if (0 === round($this->getSaturation($r, $g, $b) / 50)) {
-            return 0;
-        }
-
-        return (int) ((round($b / 255) << 2) | (round($g / 255) << 1) | round($r / 255));
-    }
-
-    private function getSaturation(int $r, int $g, int $b): int
-    {
-        $r = $r / 255;
-        $g = $g / 255;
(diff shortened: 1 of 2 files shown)


nicolas-grekas (approved):

chalasr (approved):

maxbeckers (approved):
LGTM

fabpot (approved):

fabpot:
Merging in 6.3 as this is not a bug fix, but mainly an improvement.

fabpot:
Thank you @ivanpepelko.

downhiller:
I think the main issue was a bug on line 167:

    return (int) $diff * 100 / $v;

should have been

    return (int) ($diff * 100 / $v);

$diff was always less between 0.0 and 1.0, so would always (int) to 0, or 1.

My contribution feels a bit redundant given that they've moved on and refactored a while back\, but might help someone wondering why getSaturation doesn't return anything useful.

PR #49887, opened 2023-03-31
[WebProfilerBundle] Add clickable entry view to debug toolbar

PhilETaylor:
| Q             | A
| ------------- | ---
| Branch?       | 6.3 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT
| Doc PR        | symfony/symfony-docs#... 

Just a proposed idea. 

One of the great parts of the tool bar is the Clickable IDE integration to quickly jump to the Controller method that the page is rendered from.

example:

ScreenShot-2023-03-31-21 01 07

One thing I find myself doing often is needing to get to the rendered view (a twig returned by the controller method with `return $this->render('template.twig.html,[]);` etc.  To do that I need to Hover the toolbar left corner, move up and click the controller, wait for phpStorm to load the controller file, then scroll down to find the rendered view, then go and find it manually in the tree (or with the [phpStorm Symfony plugin (amazing plugin)](https://plugins.jetbrains.com/plugin/7219-symfony-support) use the gutter icon to select the view).

After this PR the first template used in the stack will be rendered into to the twig debug toolbar (much like the controller in the request toolbar) and will be clickable if the [IDE integration is enabled](https://symfony.com/doc/current/reference/configuration/framework.html#ide) meaning less clicks to get from the webpage to the twig template to make changes in HTML quickly. 

Just a thought, would make less clicks...thoughts?

ScreenShot-2023-03-31-20 46 24

PR #50325, opened 2023-05-15
[WebProfilerBundle] Tweak the HTML code of the Twig entry view

javiereguiluz:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Stretched links are used to make entire sections, like a card, clickable. See https://getbootstrap.com/docs/5.3/helpers/stretched-link/

Maybe @PhilETaylor can answer us about this (he made the original PR to add this in #49887) but I think it wasn't intended to make entire Twig panel clickable to open the template. So, in this PR I propose to remove this behavior.

Here you can see how in the Security panel, the link is not stretched but in the Twig panel it is:

profiler-panel

### DIFF
--- a/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/twig.html.twig
+++ b/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/twig.html.twig
@@ -53,7 +53,7 @@
             <b>Entry View</b>
             <span>
                  {% if link %}
-                     <a href="{{ link }}" title="{{ file }}" class="stretched-link">
+                     <a href="{{ link }}" title="{{ file }}">
                          {{ template }}
                      </a>
                  {% else %}


PhilETaylor:
I just copied and pasted code from elsewhere in the toolbar, sorry.

I never noticed this problem, but I can confirm it's not meant to be like this and needs fixing, thanks for correcting my bad code. 

Since adding this feature though, it's literally been the thing I click most in the toolbar - its a real timesaver.

javiereguiluz:
Phil, thanks for confirming this. So, then this is OK for review and merge. Thanks.

stof (approved):

nicolas-grekas:
Thank you @javiereguiluz.

PR #48045, opened 2022-10-29
[DependencyInjection] Allow enum as service parameter in php config files

alexndlm:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        |

PR #50330, opened 2023-05-15
[DependencyInjection] Allow using enums in container configurator

### LINKED ISSUE #50275: Cannot set an enum as argument in services.php

5.4.23

### Description

Although Symfony 5.4 allows set an enum as an argument in DI Definition, it is not possible set the enum as the argument in Configurator.

Works:
```php
protected function build(ContainerBuilder $container)
{
    $container->register(MyService::class)
        ->setPublic(true)
        ->setArgument('$myEnum', MyEnum::First)
    ;
}
```

Not works:
```php
return static function (ContainerConfigurator $configurator): void {
    $configurator->services()->set(MyService::class)
        ->public()
        ->arg('$myEnum', MyEnum::First)
    ;
};
```

Throws: *Cannot use values of type "App\Service\MyEnum" in service configuration files.*

### How to reproduce

config/services.php

```php
$configurator->services()->set(MyService::class)
    ->arg('$myEnum', MyEnum::First)
;
```

src/Service/MyEnum.php

```php
<?php

namespace App\Service;

enum MyEnum: string
{
    case First = 'first';
    case Second = 'second';
}
```

src/Service/MyService.php

```php
<?php

namespace App\Service;

class MyService
{
    public function __construct(MyEnum $myEnum)
    {
    }
}
```

### Possible Solution

Symfony\Component\DependencyInjection\Loader\Configurator\AbstractConfigurator::processValue

Line 102

Replace

```php
case null === $value:
case \is_scalar($value):
    return $value;
```

by

```php
case null === $value:
case \is_scalar($value):
case $value instanceof \UnitEnum:
    return $value;
```

### Additional Context

_No response_

ludekbenedik:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50275
| License       | MIT
| Doc PR        | 

[DependencyInjection] Allow using enums in container configurator

```php
return static function (ContainerConfigurator $configurator): void {
    $configurator->services()->set(MyService::class)
        ->arg('$myEnum', MyEnum::First)
    ;
};
```

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Loader/Configurator/AbstractConfigurator.php
+++ b/src/Symfony/Component/DependencyInjection/Loader/Configurator/AbstractConfigurator.php
@@ -101,6 +101,7 @@ abstract class AbstractConfigurator
         switch (true) {
             case null === $value:
             case \is_scalar($value):
+            case $value instanceof \UnitEnum:
                 return $value;
 
             case $value instanceof ArgumentInterface:
--- a/src/Symfony/Component/DependencyInjection/Tests/Fixtures/config/services_with_enumeration.php
+++ b/src/Symfony/Component/DependencyInjection/Tests/Fixtures/config/services_with_enumeration.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\Tests\Fixtures\FooClassWithEnumAttribute;
+use Symfony\Component\DependencyInjection\Tests\Fixtures\FooUnitEnum;
+
+return function (ContainerConfigurator $container) {
+    $container->parameters()
+        ->set('unit_enum', FooUnitEnum::BAR)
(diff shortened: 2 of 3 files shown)


nicolas-grekas:
Thanks but this has already been merged as a new feature in #48045
You need to upgrade to 6.2+

ludbene:
Hi,

I know but versions 6.2 and 6.3 are not LTS. A new LTS will be minimal six months later. And then not all project will be updated immediately.

1. The Symfony 5.4 LTS will be maintained to 2025
2. It is allowed `Definition::setArgument(0, MyEnum::Foo)` but not `ServiceConfigurator::arg(0, MyEnum::Foo)`
3. Enums can be injected via `!php/const ...` in YAML but cannot be injected via PHP File Configurator. PHP File Configurator is preferred and all Symfony configurations are in PHP files
4. If a programmer wants to use PHP File Configurator and services uses enums then a part of configuration must be defined in `Kernel::configureContainer` or `Extension::load`

Why this simple fix/feature cannot be merged to Symfony 5.4 LTS?

stof:
Well, if you want the latest feature, you should not stay on the LTS. the LTS version only receives bug fixes (as it only receives semver patch releases, not new minor versions)

PR #47709, opened 2022-09-27
[HttpFoundation] Add `StreamedJsonResponse` for efficient JSON streaming

alexander-schranz:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Tickets       | Fix #... 
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/17301

When big data are streamed via JSON API it can sometimes be difficult to keep the resources usages low. For this I experimented with a different way of streaming data for JSON responses. It uses combination of `structured array` and `generics` which did result in a lot better result.

More can be read about here: [https://github.com/alexander-schranz/efficient-json-streaming-with-symfony-doctrine](https://github.com/alexander-schranz/efficient-json-streaming-with-symfony-doctrine).

I thought it maybe can be a great addition to Symfony itself to make this kind of responses easier and that APIs can be made more performant.

## Usage

<details><summary>First Version (replaced)</summary>

```php
class ArticleListAction {
    public function __invoke(EntityManagerInterface  $entityManager): Response
    {
        $articles = $this->findArticles($entityManager);

        return new StreamedJsonResponse(
            // json structure with replacers identifiers
            [
                '_embedded' => [
                    'articles' => '__articles__',
                ],
            ],
            // array of generator replacer identifier used as key
            [
                '__articles__' => $this->findArticles('Article'),
            ]
        );
    }

    private function findArticles(EntityManagerInterface  $entityManager): \Generator
    {
        $queryBuilder = $entityManager->createQueryBuilder();
        $queryBuilder->from(Article::class, 'article');
        $queryBuilder->select('article.id')
            ->addSelect('article.title')
            ->addSelect('article.description');

        return $queryBuilder->getQuery()->toIterable();
    }
}
```

</details>

Update Version (thx to @ro0NL for the idea):

```php
class ArticleListAction {
    public function __invoke(EntityManagerInterface  $entityManager): Response
    {
        $articles = $this->findArticles($entityManager);

        return new StreamedJsonResponse(
            // json structure with generators in it which are streamed
            [
                '_embedded' => [
                    'articles' => $this->findArticles('Article'), // returns a generator which is streamed
                ],
            ],
        );
    }

    private function findArticles(EntityManagerInterface  $entityManager): \Generator
    {
        $queryBuilder = $entityManager->createQueryBuilder();
        $queryBuilder->from(Article::class, 'article');
[...]

PR #50331, opened 2023-05-15
[HttpFoundation] Fix problem with empty generator in StreamedJsonResponse

alexander-schranz:
| Q             | A
| ------------- | ---
| Branch?       | 6.3 (Feature `StreamedJsonResponse`: https://github.com/symfony/symfony/pull/47709)
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix - was reported to me on Slack by @norkunas
| License       | MIT

Currently when the Generator is empty the return is invalid JSON which should not happen. So adding a testcase and a fix to the problem with the empty generator.

### DIFF
--- a/src/Symfony/Component/HttpFoundation/StreamedJsonResponse.php
+++ b/src/Symfony/Component/HttpFoundation/StreamedJsonResponse.php
@@ -130,6 +130,10 @@ class StreamedJsonResponse extends StreamedResponse
                 echo json_encode($item, $jsonEncodingOptions);
             }
 
+            if ($isFirstItem) { // indicates that the generator was empty
+                echo '[';
+            }
+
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Thank you @alexander-schranz.

src/Symfony/Component/HttpFoundation/StreamedJsonResponse.php
```diff
@@ -130,6 +130,10 @@ private function stream(): void
                 echo json_encode($item, $jsonEncodingOptions);
             }
 
+            if (true === $isFirstItem) { // indicates that the generator was empty
```
stof:
```suggestion
            if ($isFirstItem) { // indicates that the generator was empty
```

PR #30335, opened 2019-02-21
[PropertyInfo] ConstructorExtractor which has higher priority than PhpDocExtractor and ReflectionExtractor

karser:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | hopefully no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #30053
| License       | MIT

Supersedes #30056 #30128

In short, when using PhpDocExtractor, it ignores the constructor argument type, although `argument types from the constructor are the only types that are valid for the class instantiation`.

This PR adds a separate extractor - `ConstructorExtractor` which is called first (-999) and it attempts to extract the type from constructor only, either from PhpDoc or using reflection.
I added `getTypesFromConstructor` to `PhpDocExtractor` and `ReflectionExtractor` - they implement `ConstructorArgumentTypeExtractorInterface` interface. `ConstructorExtractor` aggregates those extractors using compiler pass.

So the flow of control looks like this:
```
PropertyInfoExtractor::getTypes:
    - ConstructorExtractor::getTypes
        - PhpDocExtractor::getTypesFromConstructor
        - ReflectionExtractor::getTypesFromConstructor
    - PhpDocExtractor::getTypes
    - ReflectionExtractor::getTypes
```

PR #30128, opened 2019-02-09
[PropertyInfo] ConstructorExtractor which has higher priority than PhpDocExtractor and ReflectionExtractor

karser:
| Q             | A
| ------------- | ---
| Branch?       | 4.2
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | hopefully no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #30053
| License       | MIT

The discussion is in previous PR #30056
In short, when using PhpDocExtractor, it ignores the constructor argument type, although `argument types from the constructor are the only types that are valid for the class instantiation`.

This PR adds a separate extractor - `ConstructorExtractor` which is called first (-999) and it attempts to extract the type from constructor only, either from PhpDoc or using reflection.
I added `getTypesFromConstructor` to `PhpDocExtractor` and `ReflectionExtractor` - they implement `ConstructorArgumentTypeExtractorInterface` interface. `ConstructorExtractor` aggregates those extractors using compiler pass.

So the flow of control looks like this:
```
PropertyInfoExtractor::getTypes:
    - ConstructorExtractor::getTypes
        - PhpDocExtractor::getTypesFromConstructor
        - ReflectionExtractor::getTypesFromConstructor
    - PhpDocExtractor::getTypes
    - ReflectionExtractor::getTypes
```

PR #50334, opened 2023-05-16
[FrameworkBundle][PropertyInfo] Wire the `ConstructorExtractor` class

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 7.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

A sort of followup to #30335.

The original PR removed the wiring to prevent creating a BC break, see https://github.com/symfony/symfony/pull/30128#issuecomment-462169119 & https://github.com/symfony/symfony/pull/30335#issuecomment-480799499.

This PR proposes adding a new `framework.property_info.with_constructor_extractor` configuration to allow optionally enabling the framework integration. The configuration defaults to `false` to prevent creating a BC break. Only when it's set to `true` will the `ConstructorExtractor` be registered in the container.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
+++ b/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
  * Add support for assets pre-compression
  * Rename `TranslationUpdateCommand` to `TranslationExtractCommand`
  * Add JsonEncoder services and configuration
+ * Add new `framework.property_info.with_constructor_extractor` option to allow enabling or disabling the constructor extractor integration
 
 7.2
 ---
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
@@ -1226,7 +1226,22 @@ class Configuration implements ConfigurationInterface
                 ->arrayNode('property_info')
                     ->info('Property info configuration')
                     ->{$enableIfStandalone('symfony/property-info', PropertyInfoExtractorInterface::class)}()
+                    ->children()
+                        ->booleanNode('with_constructor_extractor')
+                            ->info('Registers the constructor extractor.')
+                        ->end()
+                    ->end()
                 ->end()
+            ->end()
+            ->validate()
+                ->ifTrue(fn ($v) => $v['property_info']['enabled'] && !isset($v['property_info']['with_constructor_extractor']))
+                ->then(function ($v) {
+                    $v['property_info']['with_constructor_extractor'] = false;
+
+                    trigger_deprecation('symfony/property-info', '7.3', 'Not setting the "with_constructor_extractor" option explicitly is deprecated because its default value will change in version 8.0.');
+
+                    return $v;
+                })
             ->end()
         ;
     }
--- a/src/Symfony/Component/PropertyInfo/Extractor/ConstructorArgumentTypeExtractorInterface.php
+++ b/src/Symfony/Component/PropertyInfo/Extractor/ConstructorArgumentTypeExtractorInterface.php
@@ -18,8 +18,6 @@ use Symfony\Component\TypeInfo\Type;
  * Infers the constructor argument type.
  *
  * @author Dmitrii Poddubnyi <dpoddubny@gmail.com>
- *
- * @internal
  */
 interface ConstructorArgumentTypeExtractorInterface
 {
@@ -27,8 +25,6 @@ interface ConstructorArgumentTypeExtractorInterface
      * Gets types of an argument from constructor.
      *
      * @return LegacyType[]|null
-     *
-     * @internal
      */
     public function getTypesFromConstructor(string $class, string $property): ?array;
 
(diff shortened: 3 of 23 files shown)


michaljusiega:
What is the status of this PR ? It would be nice to see this in 7.1.

src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ CHANGELOG
  * Make the `config/` directory optional in `MicroKernelTrait`, add support for service arguments in the
    invokable Kernel class, and register `FrameworkBundle` by default when the `bundles.php` file is missing
  * [BC BREAK] The `secrets:decrypt-to-local` command terminates with a non-zero exit code when a secret could not be read
+ * Wire the `Symfony\Component\PropertyInfo\Extractor\ConstructorExtractor` class
```
stof: this line should be removed IMO (the next line describes the relevant user-visible change)

src/Symfony/Component/PropertyInfo/Extractor/ConstructorArgumentTypeExtractorInterface.php
```diff
@@ -18,8 +18,6 @@
  * Infers the constructor argument type.
  *
  * @author Dmitrii Poddubnyi <dpoddubny@gmail.com>
- *
- * @internal
```
stof, line 22: is it expected to make the interface not internal anymore but to keep its methods internal ?

OskarStark:
What about directly deprecating this option, so that it needs to be set on 8.0? Or defaulting it to true in 8.0?

src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
HypeMC: Done

src/Symfony/Component/PropertyInfo/Extractor/ConstructorArgumentTypeExtractorInterface.php
HypeMC: No, I must've not noticed those, removed.

HypeMC:
> What about directly deprecating this option, so that it needs to be set on 8.0? Or defaulting it to true in 8.0?

I'm ok with whatever is decided.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1199,7 +1199,14 @@ private function addPropertyInfoSection(ArrayNodeDefinition $rootNode, callable
             ->children()
                 ->arrayNode('property_info')
                     ->info('Property info configuration')
+                    ->addDefaultsIfNotSet()
                     ->{$enableIfStandalone('symfony/property-info', PropertyInfoExtractorInterface::class)}()
+                    ->children()
+                        ->booleanNode('with_constructor_extractor')
+                            ->info('Registers the constructor extractor.')
+                            ->defaultFalse()
```
mtarld: I wouldn't have added any defaults here. Indeed, in that way, we could be able to raise a deprecation message in `FrameworkExtension` telling that `with_constructor_extractor` must be explicitly set, as its default value will change in `8.0`.

If we're doing so, we must update the recipe so that it sets the `with_constructor_extractor` value to `true` for new `7.3` projects.

HypeMC: @mtarld Something like this https://github.com/symfony/symfony/pull/50334/commits/f50586a65173d6607a2f42f3abbae31b7414c7eb ?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1225,9 +1225,25 @@ private function addPropertyInfoSection(ArrayNodeDefinition $rootNode, callable
             ->children()
                 ->arrayNode('property_info')
                     ->info('Property info configuration')
+                    ->addDefaultsIfNotSet()
```
mtarld: Is it required?

src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ CHANGELOG
  * Add support for assets pre-compression
  * Rename `TranslationUpdateCommand` to `TranslationExtractCommand`
  * Add JsonEncoder services and configuration
+ * Add new `framework.property_info.with_constructor_extractor` option to allow enabling or disabling the constructor extractor integration
```
mtarld, line 10: I think we must update the `UPGRADE-7.3.md` file as well, as we trigger a deprecation

HypeMC: Good catch, done

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
HypeMC: No, removed.

mtarld (approved):

fabpot:
Thank you @HypeMC.

PR #50335, opened 2023-05-16
[HttpKernel] Add optional `$className` param to `ControllerEvent::getAttributes()`

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

A minor DX improvement since most of the time only one type of attribute is needed, eg:

```php
public function onKernelControllerArguments(ControllerArgumentsEvent $event)
{
    $attributes = $event->getAttributes(SomeAttribute::class);

    // ...
}
```

### DIFF
--- a/src/Symfony/Component/HttpKernel/CHANGELOG.md
+++ b/src/Symfony/Component/HttpKernel/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+ * Add optional `$className` parameter to `ControllerEvent::getAttributes()`
+
 6.3
 ---
 
--- a/src/Symfony/Component/HttpKernel/Event/ControllerArgumentsEvent.php
+++ b/src/Symfony/Component/HttpKernel/Event/ControllerArgumentsEvent.php
@@ -94,10 +94,16 @@ final class ControllerArgumentsEvent extends KernelEvent
     }
 
     /**
-     * @return array<class-string, list<object>>
+     * @template T of class-string|null
+     *
+     * @param T $className
+     *
+     * @return array<class-string, list<object>>|list<object>
+     *
+     * @psalm-return (T is null ? array<class-string, list<object>> : list<object>)
      */
-    public function getAttributes(): array
+    public function getAttributes(string $className = null): array
     {
-        return $this->controllerEvent->getAttributes();
+        return $this->controllerEvent->getAttributes($className);
     }
 }
--- a/src/Symfony/Component/HttpKernel/Event/ControllerEvent.php
+++ b/src/Symfony/Component/HttpKernel/Event/ControllerEvent.php
@@ -79,12 +79,18 @@ final class ControllerEvent extends KernelEvent
     }
 
     /**
-     * @return array<class-string, list<object>>
+     * @template T of class-string|null
+     *
+     * @param T $className
+     *
+     * @return array<class-string, list<object>>|list<object>
+     *
+     * @psalm-return (T is null ? array<class-string, list<object>> : list<object>)
      */
-    public function getAttributes(): array
+    public function getAttributes(string $className = null): array
     {
         if (isset($this->attributes)) {
-            return $this->attributes;
+            return null === $className ? $this->attributes : $this->attributes[$className] ?? [];
         }
 
         if (\is_array($this->controller) && method_exists(...$this->controller)) {
@@ -102,6 +108,6 @@ final class ControllerEvent extends KernelEvent
             }
         }
 
-        return $this->attributes;
+        return null === $className ? $this->attributes : $this->attributes[$className] ?? [];
     }
 }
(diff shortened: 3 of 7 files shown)


src/Symfony/Component/HttpKernel/Event/ControllerArgumentsEvent.php
```diff
@@ -94,10 +94,16 @@ public function getNamedArguments(): array
...
+     * @template T of class-string|null
+     *
+     * @param T $className
+     *
+     * @return array<class-string, list<object>>|list<object>
+     *
+     * @psalm-return (T is null ? array<class-string, list<object>> : list<object>)
      */
-    public function getAttributes(): array
+    public function getAttributes(string $className = null): array
```
JohJohan, line 105: Is it not better to make a `getAttribute()` instead?

src/Symfony/Component/HttpKernel/Event/ControllerArgumentsEvent.php
```diff
@@ -94,10 +94,16 @@ public function getNamedArguments(): array
...
 
     /**
-     * @return array<class-string, list<object>>
+     * @template T of class-string|null
+     *
+     * @param T $className
+     *
+     * @return array<class-string, list<object>>|list<object>
+     *
+     * @psalm-return (T is null ? array<class-string, list<object>> : list<object>)
```
stof, line 103: Instead of a conditional return type, I'm wondering whether a separate method would be better.

src/Symfony/Component/HttpKernel/Event/ControllerArgumentsEvent.php
stof: it still returns an array of attributes, as you can have multiple attributes of the same class

src/Symfony/Component/HttpKernel/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ CHANGELOG
  * Add `#[MapQueryString]` to map and validate request query string from `Request::$query->all()` to typed objects
  * Add `#[MapQueryParameter]` to map and validate individual query parameters to controller arguments
  * Collect data from every event dispatcher
+ * Add optional `$className` parameter to `ControllerEvent::getAttributes()`
```
stof: This won't make it in 6.3 as we are 1.5 month after the feature freeze.

src/Symfony/Component/HttpKernel/Event/ControllerArgumentsEvent.php
nicolas-grekas: We already have similar APIs and they work great IMHO.

nicolas-grekas:
Is this really worth it?

```diff
-$attributes = $event->getAttributes(SomeAttribute::class);
+$attributes = $event->getAttributes()[SomeAttribute::class] ?? null;
```

stof:
@nicolas-grekas the difference is that the new argument has a generic signature, telling SA tools that all returned objects are an instance of the provided class name.
For the case of returning all attributes grouped by a class name, there is no interoperable way to indicate that in phpdoc (Psalm has a `class-string-map` type, but no other tool support it)

nicolas-grekas:
Wouldn't this work?
`@return array<class-string<T>, list<T>>`

stof:
@nicolas-grekas no. This won't work as it would force using the same `T` (and inferring it from somewhere).

That's precisely why Psalm introduced a `class-string-map` type, which would be exactly what we need here if it was interoperable.

src/Symfony/Component/HttpKernel/Event/ControllerEvent.php
```diff
@@ -79,29 +79,33 @@ public function setController(callable $controller, array $attributes = null): v
...
+     *
+     * @return array<class-string, list<object>>|list<object>
+     *
+     * @psalm-return (T is null ? array<class-string, list<object>> : list<object>)
      */
-    public function getAttributes(): array
+    public function getAttributes(string $className = null): array
     {
-        if (isset($this->attributes)) {
-            return $this->attributes;
```
nicolas-grekas, line 87: I would prefer keeping the early return
`return null === $className ? $this->attributes : $this->attributes[$className] ?? [];`

HypeMC: Done

nicolas-grekas (approved):

src/Symfony/Component/HttpKernel/CHANGELOG.md
HypeMC: Fixed

fabpot (approved):

fabpot:
Thank you @HypeMC.

PR #50338, opened 2023-05-16
[Console] Remove ``exec`` and replace it by ``shell_exec``

### LINKED ISSUE #49646: [Console] Terminal::hasSttyAvailable() checks availability of exec but then all calls are made via shell_exec

6.2.7

### Description

Error message:
- `Error: Call to undefined function Symfony\Component\Console\shell_exec() in /{path_to_project}/vendor/symfony/console/Application.php:989 `
- https://github.com/symfony/console/blob/cbad09eb8925b6ad4fb721c7a179344dc4a19d45/Application.php#L989

`Terminal:hasSttyAvailable()` checks for availability of `exec`
- https://github.com/symfony/console/blob/cbad09eb8925b6ad4fb721c7a179344dc4a19d45/Terminal.php#L120-L134

However, all calls inside the condition are then run using `shell_exec` and that fails when it is disabled.
- https://github.com/symfony/console/blob/cbad09eb8925b6ad4fb721c7a179344dc4a19d45/Application.php#L988-L996

One might argue that it does not make much sense to have one function enabled and the other disabled, but apparently it can happen and then it fails.

Or am I missing something? :)

### How to reproduce

Have a system, that has:
- exec enabled
- shell_exec disabled
- symfony/console installed
- symfony/events installed (otherwise checks for dispatcher fail and a different route in the code is taken that will not manifest the problem)

Then try to run the symfony console from command line. 

### Possible Solution

Either check availability of `shell_exec` in `Terminal::hasSttyAvailable()` or change all calls to use `exec`.

### Additional Context

The check has been introduced here:
- https://github.com/symfony/symfony/pull/37466

maxbeckers:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #49646
| License       | MIT
| Doc PR        | n/a

It could happen an error like `Error: Call to undefined function Symfony\Component\Console\shell_exec() in /{path_to_project}/vendor/symfony/console/Application.php:989 ` when `exec` is available but `shell_exec` not. There is very often used the check `Terminal::hasSttyAvailable()` in the symfony codebase and then `shell_exec` used.

More Details see in the issue #49646.

### DIFF
--- a/src/Symfony/Component/Console/Helper/QuestionHelper.php
+++ b/src/Symfony/Component/Console/Helper/QuestionHelper.php
@@ -500,13 +500,11 @@ class QuestionHelper extends Helper
             return self::$stdinIsInteractive = @posix_isatty(fopen('php://stdin', 'r'));
         }
 
-        if (!\function_exists('exec')) {
+        if (!\function_exists('shell_exec')) {
             return self::$stdinIsInteractive = true;
         }
 
-        exec('stty 2> /dev/null', $output, $status);
-
-        return self::$stdinIsInteractive = 1 !== $status;
+        return self::$stdinIsInteractive = (bool) shell_exec('stty 2> '.('\\' === \DIRECTORY_SEPARATOR ? 'NUL' : '/dev/null'));
     }
 
     /**
--- a/src/Symfony/Component/Console/Terminal.php
+++ b/src/Symfony/Component/Console/Terminal.php
@@ -64,14 +64,12 @@ class Terminal
             return self::$stty;
         }
 
-        // skip check if exec function is disabled
-        if (!\function_exists('exec')) {
+        // skip check if shell_exec function is disabled
+        if (!\function_exists('shell_exec')) {
             return false;
         }
 
-        exec('stty 2>&1', $output, $exitcode);
-
-        return self::$stty = 0 === $exitcode;
+        return self::$stty = (bool) shell_exec('stty 2> '.('\\' === \DIRECTORY_SEPARATOR ? 'NUL' : '/dev/null'));
     }
 
     private static function initDimensions()
--- a/src/Symfony/Component/Console/Tests/Helper/QuestionHelperTest.php
+++ b/src/Symfony/Component/Console/Tests/Helper/QuestionHelperTest.php
@@ -430,7 +430,7 @@ class QuestionHelperTest extends AbstractQuestionHelperTestCase
         $this->assertEquals('8AM', $dialog->ask($this->createStreamableInputInterfaceMock($this->getInputStream("8AM\n")), $this->createOutputInterface(), $question));
     }
 
-    public function testAskHiddenResponseTrimmed()
+    public function testAskHiddenResponseNotTrimmed()
     {
         if ('\\' === \DIRECTORY_SEPARATOR) {
             $this->markTestSkipped('This test is not supported on Windows');
@@ -442,7 +442,7 @@ class QuestionHelperTest extends AbstractQuestionHelperTestCase
         $question->setHidden(true);
         $question->setTrimmable(false);
 
-        $this->assertEquals(' 8AM', $dialog->ask($this->createStreamableInputInterfaceMock($this->getInputStream(' 8AM')), $this->createOutputInterface(), $question));
+        $this->assertEquals(' 8AM'.\PHP_EOL, $dialog->ask($this->createStreamableInputInterfaceMock($this->getInputStream(' 8AM'.\PHP_EOL)), $this->createOutputInterface(), $question));
     }
 
     public function testAskMultilineResponseWithEOF()
--- a/src/Symfony/Component/Console/Tests/TerminalTest.php
+++ b/src/Symfony/Component/Console/Tests/TerminalTest.php
@@ -77,8 +77,8 @@ class TerminalTest extends TestCase
             $this->markTestSkipped('Must be on windows');
         }
 
-        $sttyString = exec('(stty -a | grep columns) 2>&1', $output, $exitcode);
-        if (0 !== $exitcode) {
+        $sttyString = shell_exec('(stty -a | grep columns) 2> NUL');
+        if (!$sttyString) {
             $this->markTestSkipped('Must have stty support');
         }
 


nicolas-grekas:
Can't we use `exec` everywhere instead?

maxbeckers:
I could try this ... give me a bit of time for that, could become a bigger PR :)

src/Symfony/Component/Console/Helper/QuestionHelper.php
```diff
@@ -417,8 +417,8 @@ private function getHiddenResponse(OutputInterface $output, $inputStream, bool $
                 $exe = $tmpExe;
             }
 
-            $sExec = shell_exec('"'.$exe.'"');
-            $value = $trimmable ? rtrim($sExec) : $sExec;
+            $sExec = exec('"'.$exe.'"');
+            $value = $trimmable ? rtrim($sExec) : $sExec . PHP_EOL; // add PHP_EOL to keep the same result as shell_exec
```
maxbeckers: Do we need this or can this be ignored @nicolas-grekas ?

chalasr: I don't think so: `rtrim()` would have removed it before this patch

src/Symfony/Component/Console/Tests/Helper/QuestionHelperTest.php
```diff
@@ -430,7 +430,7 @@ public function testAskHiddenResponse()
         $this->assertEquals('8AM', $dialog->ask($this->createStreamableInputInterfaceMock($this->getInputStream("8AM\n")), $this->createOutputInterface(), $question));
     }
 
-    public function testAskHiddenResponseTrimmed()
+    public function testAskHiddenResponseNotTrimmed()
```
chalasr, line 433: this test was valid before the patch, it should keep passing as-is with it

src/Symfony/Component/Console/Helper/QuestionHelper.php
nicolas-grekas: I played a bit with exec vs shell_exec and what the php doc says is correct: lines are trimmed when using exec.
That means using it would break answers with trailing spaces.
The only way is using shell_exec indeed.
What about doing the reverse and always using shell_exec everywhere, including in the function_exists check?

chalasr: 👍 exec would break multiline answers as well since it only returns the last line

src/Symfony/Component/Console/Tests/Helper/QuestionHelperTest.php
maxbeckers: @chalasr yes it did pass. I just optimized the test, because it was called `testAskHiddenResponseTrimmed` and `$question->setTrimmable(false);` was set so it's more a not trimmed. And I added the EOL to see that the end is also not trimmed.

But I'll make it parameterized with a DataProvider to keep the old test also.

chalasr: 🤦 thanks

src/Symfony/Component/Console/Helper/QuestionHelper.php
maxbeckers: @nicolas-grekas in `QuestionHelper::isInteractiveInput` for example we use `exec` to also get the result code.

Multiline answers would be possibe using the $output array, I indeed forgot to check this.

nicolas-grekas: I think we can use $output instead: it's going to be empty if stty fails:
`'' === trim(shell_exec(...))`

src/Symfony/Component/Console/Helper/QuestionHelper.php
```diff
@@ -500,13 +500,11 @@ private function isInteractiveInput($inputStream): bool
...
 
-        if (!\function_exists('exec')) {
+        if (!\function_exists('shell_exec')) {
             return self::$stdinIsInteractive = true;
         }
 
-        exec('stty 2> /dev/null', $output, $status);
-
-        return self::$stdinIsInteractive = 1 !== $status;
+        return self::$stdinIsInteractive = is_string(shell_exec('stty 2> /dev/null'));
```
nicolas-grekas:
```suggestion
        return self::$stdinIsInteractive = (bool) shell_exec('stty 2> /dev/null');
```

src/Symfony/Component/Console/Terminal.php
```diff
@@ -64,14 +64,12 @@ public static function hasSttyAvailable(): bool
...
-        if (!\function_exists('exec')) {
+        // skip check if shell_exec function is disabled
+        if (!\function_exists('shell_exec')) {
             return false;
         }
 
-        exec('stty 2>&1', $output, $exitcode);
-
-        return self::$stty = 0 === $exitcode;
+        return self::$stty = (bool) shell_exec('stty 2>&1');
```
nicolas-grekas: should be `stty 2> /dev/null` I think now

src/Symfony/Component/Console/Tests/TerminalTest.php
```diff
@@ -77,8 +77,8 @@ public function testSttyOnWindows()
             $this->markTestSkipped('Must be on windows');
         }
 
-        $sttyString = exec('(stty -a | grep columns) 2>&1', $output, $exitcode);
-        if (0 !== $exitcode) {
+        $sttyString = shell_exec('(stty -a | grep columns) 2>&1');
```
nicolas-grekas:
```suggestion
        $sttyString = shell_exec('(stty -a | grep columns) 2> /dev/null');
```

src/Symfony/Component/Console/Terminal.php
maxbeckers: This would change a bit the behavior, at least on windows (using GitBash):

Example code:
```
        exec('stty 2>&1', $output, $exitcode);
        var_dump(0 === $exitcode);
        var_dump((bool) shell_exec('stty 2> /dev/null'));
```

Result:
```
bool(true)
The System Cannot Find The Path Specified.
bool(false)
```

nicolas-grekas: what a mess!
Can you try this instead?
`command -v stty >/dev/null 2>&1 && stty 2> /dev/null`

src/Symfony/Component/Console/Tests/TerminalTest.php
maxbeckers: This is only for windows so i changes to `NUL`

src/Symfony/Component/Console/Terminal.php
maxbeckers: I found a way, working on windows. Please have a look if it's fine for you.

nicolas-grekas (approved):

nicolas-grekas:
Thank you @maxbeckers.

PR #50340, opened 2023-05-16
[HttpKernel] Make `QueryParameterValueResolver` provide a value if possible when a parameter is not found

### LINKED ISSUE #50339: [HttpKernel] #[MapQueryParameter] Problem with nullable parameters

6.3.0-BETA3

### Description

Hi,

In documentation is an example with optional nullable string parameter (?string $theme = null).

> Symfony has allowed to map route parameters to controller arguments since its first version. In Symfony 6.3 you will also be able to map query string parameters to specific controller arguments. To do so, use the new #[MapQueryParameter] attribute:
> 
```#[Route(path: '/', name: 'admin_dashboard')]
public function indexAction(
    #[MapQueryParameter] array $ids,
    #[MapQueryParameter] string $firstName,
    #[MapQueryParameter] bool $required,
    #[MapQueryParameter] int $age,
    #[MapQueryParameter] string $category = '',
    #[MapQueryParameter] ?string $theme = null,
)
```
> If the query string of the incoming request is /?ids[]=1&ids[]=2&firstName=Ruud&required=3&age=123, this code will make the controller arguments to have these values:

```
$ids = ['1', '2']
$firstName = "Ruud"
$required = false
$age = 123
$category = ''
$theme = null
```

 

I test similar example on v6.3.0-BETA3 on my application but nullable query string parameter doesn't work.

```
    #[Route(path: '/')]
    public function indexAction(
        #[MapQueryParameter] ?string $theme = null
    ): Response {
        return new Response('test');
    }
```

1) `GET /?theme` works

2) `GET /` leads to RuntimeException (Controller "App\Controller\IndexController::indexAction" requires that you provide a value for the "$theme" argument. Either the argument is nullable and no null value has been provided, no default value has been provided or because there is a non optional argument after this one.)

### How to reproduce

Example: https://github.com/szepczynski/symfony-6.3-map-query-parameter-bug-reproduction/

### Possible Solution

_No response_

### Additional Context

_No response_

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50339
| License | MIT
| Doc PR        | N/A

As `QueryParameterValueResolver` is “targeted” it must provide a value itself if possible.

When a parameter is not found in the query string, we want to return the argument’s default value if it has one, or `null` if it is nullable. Else the behavior do not change: we throw a `NotFoundHttpException`.

### DIFF
--- a/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php
+++ b/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php
@@ -31,8 +31,12 @@ final class QueryParameterValueResolver implements ValueResolverInterface
 
         $name = $attribute->name ?? $argument->getName();
         if (!$request->query->has($name)) {
-            if ($argument->isNullable() || $argument->hasDefaultValue()) {
-                return [];
+            if ($argument->hasDefaultValue()) {
+                return [$argument->getDefaultValue()];
+            }
+
+            if ($argument->isNullable()) {
+                return [null];
             }
 
             throw new NotFoundHttpException(sprintf('Missing query parameter "%s".', $name));
--- a/src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/QueryParameterValueResolverTest.php
+++ b/src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/QueryParameterValueResolverTest.php
@@ -179,14 +179,14 @@ class QueryParameterValueResolverTest extends TestCase
         yield 'parameter not found but nullable' => [
             Request::create('/', 'GET'),
             new ArgumentMetadata('firstName', 'string', false, false, false, true, [new MapQueryParameter()]),
-            [],
+            [null],
             null,
         ];
 
         yield 'parameter not found but optional' => [
             Request::create('/', 'GET'),
             new ArgumentMetadata('firstName', 'string', false, true, false, attributes: [new MapQueryParameter()]),
-            [],
+            [false],
             null,
         ];
 


src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php
```diff
@@ -31,8 +31,12 @@ public function resolve(Request $request, ArgumentMetadata $argument): array
 
         $name = $attribute->name ?? $argument->getName();
         if (!$request->query->has($name)) {
-            if ($argument->isNullable() || $argument->hasDefaultValue()) {
-                return [];
+            if ($argument->hasDefaultValue()) {
+                return [$argument->getDefaultValue()];
```
nicolas-grekas, line 35: ~Can't we remove this "if" and~ let PHP do that part?

nicolas-grekas:
```suggestion
                return [];
```

stof: I don't think this will work fine if that argument is not the last one of the controller signature.

nicolas-grekas: You mean to account for optional parameter declared before required parameters. This is deprecated but OK.

nicolas-grekas (approved):

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php
MatTheCat: Wouldn’t `ArgumentResolver` throw if we don’t return the default value here anyways?

stof: No, I mean  to account for optional parameters declared before another optional parameter for which an argument resolver would provide a value.

nicolas-grekas:
Thank you @MatTheCat.

PR #50347, opened 2023-05-16
[DebugBundle][VarDumper] Fix dump labels compatibility

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

This PR fixes the handling of dumps labels when the DebugBundle is enabled and stays BC with older versions.

@javiereguiluz Unfortunately, the CSS of the web profiler toolbar is not perfect for labels and invisible characters:
Screenshot from 2023-05-16 18-31-01
Could you take care of this part plz? :pray:

### DIFF
--- a/src/Symfony/Component/VarDumper/Cloner/Data.php
+++ b/src/Symfony/Component/VarDumper/Cloner/Data.php
@@ -269,17 +269,10 @@ class Data implements \ArrayAccess, \Countable, \IteratorAggregate
     {
         $refs = [0];
         $cursor = new Cursor();
-        $label = $this->context['label'] ?? '';
-
-        if ($cursor->attr = $this->context[SourceContextProvider::class] ?? []) {
-            $cursor->attr['if_links'] = true;
-            $cursor->hashType = -1;
-            $dumper->dumpScalar($cursor, 'default', $label.'^');
-            $cursor->attr = ['if_links' => true];
-            $dumper->dumpScalar($cursor, 'default', ' ');
-            $cursor->hashType = 0;
-        }
-
+        $cursor->hashType = -1;
+        $cursor->attr = $this->context[SourceContextProvider::class] ?? [];
+        $dumper->dumpScalar($cursor, 'label', $this->context['label'] ?? '');
+        $cursor->hashType = 0;
         $this->dumpItem($dumper, $cursor, $refs, $this->data[$this->position][$this->key]);
     }
 
--- a/src/Symfony/Component/VarDumper/Resources/functions/dump.php
+++ b/src/Symfony/Component/VarDumper/Resources/functions/dump.php
@@ -25,7 +25,7 @@ if (!function_exists('dump')) {
             return null;
         }
 
-        if (isset($vars[0]) && 1 === count($vars)) {
+        if (array_key_exists(0, $vars) && 1 === count($vars)) {
             VarDumper::dump($vars[0]);
             $k = 0;
         } else {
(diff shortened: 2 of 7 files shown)


nicolas-grekas:
can you please confirm that the label is a link to the source?

fancyweb:
DebugBundle behavior is to have links on the file (FooAction) not on the label. But wait I've found one more problem and I'm crafting a better patch.

src/Symfony/Component/VarDumper/Resources/functions/dump.php
```diff
@@ -25,7 +25,7 @@ function dump(mixed ...$vars): mixed
             return null;
         }
 
-        if (isset($vars[0]) && 1 === count($vars)) {
+        if (array_key_exists(0, $vars) && 1 === count($vars)) {
```
fancyweb, line 28: Fix dump(null) showing 1^ null

src/Symfony/Component/VarDumper/Cloner/Data.php
```diff
@@ -271,11 +271,19 @@ public function dump(DumperInterface $dumper)
         $cursor = new Cursor();
         $label = $this->context['label'] ?? '';
 
-        if ($cursor->attr = $this->context[SourceContextProvider::class] ?? []) {
-            $cursor->attr['if_links'] = true;
+        if (($cursor->attr = $this->context[SourceContextProvider::class] ?? []) || '' !== $label) {
+            $label .= '^';
+            if ($cursor->attr) {
+                $cursor->attr['if_links'] = '^' === $label ? true : $label;
```
stof: Passing a string to `if_links` to force it to be displayed does not make sense IMO. If you want a node to be always rendered, use `if_links => false` (or omit it entirely as that's the default behavior).

The whole point of `if_links` is to render the node only if links are supported.

fancyweb: It's complicated :smiling_face_with_tear: The problem is that we want the label to be part of the link but at this point, we don't know if the link will be displayed. If links are not supported we still want to render the label.

stof: but then, it means you always want to render the node (with a href if links are supported).

fancyweb: No, otherwise we'd change the behavior when there is no label and no link (we would display `^ foo` instead of `foo` atm).
Maybe the simple move is just to always render `^` in CLI whether there's a link or not as we do in HTML dumps... So we can remove the if_links logic and make the behaviors coherent.

nicolas-grekas: I fixed the issue by adding a new "label" style

nicolas-grekas:
Thank you @fancyweb.

bobthecow:
@fancyweb @nicolas-grekas This change introduces a backwards compatibility break. [See PsySH dependency test failures at dev stability](https://github.com/bobthecow/psysh/actions/workflows/deps.yml) starting with this change.

bobthecow:
Specifically, [PsySH extends VarDumper and uses its `styles` map](https://github.com/bobthecow/psysh/blob/main/src/VarDumper/Dumper.php#L93). It looks like this change doesn't add `label` to the map, but expects callers to use `default` instead, e.g.:

```php
$this->styles['label' === $style ? 'default' : $style]
```

PR #48685, opened 2022-12-16
[DependencyInjection] Exclude referencing service (self) in `TaggedIteratorArgument`

chalasr:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Suggested by @OskarStark in https://twitter.com/OskarStark/status/1594641457226436608?s=20&t=Wbqw_Mu2tMYQ0iouaG8yig.
This PR avoids the need to explicitly indicate to exclude the referencing service when using `#[TaggedIterator]` and variants. 

Before:
```php
final class DelegatingErrorTracker implements ErrorTracker
{
    public function __construct(
        #TaggedIterator(ErrorTracker::class, exclude: self::class)]
        private iterable $trackers
    ) {}
}
```

After:
```php
final class DelegatingErrorTracker implements ErrorTracker
{
    public function __construct(
        #TaggedIterator(ErrorTracker::class]
        private iterable $trackers
    ) {}
}
```

I went with no deprecation as I think the breakage potential is close to zero, but happy to reconsider.

PR #50351, opened 2023-05-16
[DependencyInjection] Add `excludeSelf` option to dumpers

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Followup to #48685.

It seems that the `excludeSelf` option was not added to the dumpers or configuration functions. I don't think this was done intentionally, but was an oversight.

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Dumper/XmlDumper.php
+++ b/src/Symfony/Component/DependencyInjection/Dumper/XmlDumper.php
@@ -324,6 +324,9 @@ class XmlDumper extends Dumper
                         }
                     }
                 }
+                if (!$tag->excludeSelf()) {
+                    $element->setAttribute('exclude-self', 'false');
+                }
             } elseif ($value instanceof IteratorArgument) {
                 $element->setAttribute('type', 'iterator');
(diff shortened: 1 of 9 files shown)


stof (approved):

src/Symfony/Component/DependencyInjection/Dumper/XmlDumper.php
```diff
@@ -324,6 +324,9 @@ private function convertParameters(array $parameters, string $type, \DOMElement
                         }
                     }
                 }
+                if (!$excludeSelf = $tag->excludeSelf()) {
```
nicolas-grekas: no need for the $excludeSelf variable (same in the yaml dumper)

HypeMC: Good point, done https://github.com/symfony/symfony/compare/698361a81af67199510ec429e630ac16641d41f4..09c758142d57c6fddffe9813b18aaaa52e054387

nicolas-grekas:
Thank you @HypeMC.

PR #50354, opened 2023-05-17
[Process] Stop the process correctly even if underlying input stream is not closed

joelwurtz:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       |  /
| License       | MIT
| Doc PR        | /

While checking a process to end, on posix system, process component only checks if pipes are still open, this fix ensure that if the process is terminated it correctly return, even if the underlying pipe is not closed.

It can be useful when using \STDIN as a input stream as it will always be open

### DIFF
--- a/src/Symfony/Component/Process/Process.php
+++ b/src/Symfony/Component/Process/Process.php
@@ -428,7 +428,7 @@ class Process implements \IteratorAggregate
 
         do {
             $this->checkTimeout();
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Thank you @joelwurtz.

PR #50355, opened 2023-05-17
[AssetMapper] Bug fix for paths that start with the same string

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | Still TODO

Hi!

This should be minor:

1) Fixed a bug where if 1 asset path directory is a substring of another (e.g. `assets/dir` and `assets/dir2`), when you called `AssetMapperRepository::findLogicalPath()` passing a filesystem path pointed at `assets/dir2`, it would incorrectly match something in `assets/dir`.

~~2) Allowed `importmap.php` paths to be absolute. The use-case is for UX libraries. UX libraries WILL automatically add a "namespaced path" to the asset mapper. For example, the StimulusBundle I'm working on will likely expose a `symfony/stimulus-bundle` asset mapper namespace that points to the `assets/` directory of that bundle. But, for users, I don't want this to be "another thing you need to know". So, the result of this PR is that, in the user's `importmap.php`:~~

```diff
    '@symfony/stimulus-bundle' => [
-        // "symfony/stimulus-bundle" is an asset path that points into that bundle
-        'path' => 'symfony/stimulus-bundle/loader.js',
+        'path' => __DIR__.'/../../assets/loader.js',
    ],
```

~~The user points to the absolute file, WE figure out which "asset path" that's in, and everything works nicely.~~

UPDATE: I've removed this.

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/AssetMapperRepository.php
+++ b/src/Symfony/Component/AssetMapper/AssetMapperRepository.php
@@ -76,7 +76,7 @@ class AssetMapperRepository
         }
 
         foreach ($this->getDirectories() as $path => $namespace) {
-            if (!str_starts_with($filesystemPath, $path)) {
+            if (!str_starts_with($filesystemPath, $path.\DIRECTORY_SEPARATOR)) {
                 continue;
             }
 
--- a/src/Symfony/Component/AssetMapper/Tests/AssetMapperRepositoryTest.php
+++ b/src/Symfony/Component/AssetMapper/Tests/AssetMapperRepositoryTest.php
@@ -70,6 +70,7 @@ class AssetMapperRepositoryTest extends TestCase
     public function testFindLogicalPath()
     {
         $repository = new AssetMapperRepository([
+            'dir' => '',
             'dir1' => 'some_namespace',
             'dir2' => '',
         ], __DIR__.'/fixtures');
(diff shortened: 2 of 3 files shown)


src/Symfony/Component/AssetMapper/AssetMapperRepository.php
```diff
@@ -76,7 +76,7 @@ public function findLogicalPath(string $filesystemPath): ?string
         }
 
         foreach ($this->getDirectories() as $path => $namespace) {
-            if (!str_starts_with($filesystemPath, $path)) {
+            if (!str_starts_with($filesystemPath, $path.'/')) {
```
stof: Be careful. Import map paths might already have a trailing `/`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap#mapping_path_prefixes

weaverryan: I think we're ok. These are all comparing filesystem paths - not the actual paths that are exposed to the user / importmap.

weaverryan:
I've just simplified this - it's now just a simple bug fix. The "absolute path" thing might be cool. But in practice, during a UX bundle install, we'll use `importmap:require` to update the importmap.php using #50363 and there was no graceful way, anyways, for that feature to output a string like `__DIR__.'/...'`, which is what I wanted. So, let's just stick to using asset logical paths everywhere.

nicolas-grekas:
Thank you @weaverryan.

PR #50356, opened 2023-05-17
[AssetMapper] Fix bug where dependencies were preloaded even if the parent was not

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | Still TODO

In `importmap.php`, if you have `preload => false` for an assets... but that asset imports another assets, we would always list that imported asset as "preload => true". Basically, when looking at a dependency to determine if it should be preloaded, we need to first look at the parent to see if IT was preloaded.

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
@@ -417,11 +417,11 @@ class ImportMapManager
                 $this->modulesToPreload[] = $path;
             }
 
-            $dependencyImportMapEntries = array_map(function (AssetDependency $dependency) {
+            $dependencyImportMapEntries = array_map(function (AssetDependency $dependency) use ($entryOptions) {
                 return new ImportMapEntry(
                     $dependency->asset->getPublicPathWithoutDigest(),
                     $dependency->asset->getLogicalPath(),
-                    preload: !$dependency->isLazy,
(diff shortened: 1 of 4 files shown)


src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -140,7 +140,7 @@ private function buildImportMapJson(): void
         $entries = $this->loadImportMapEntries();
         $this->modulesToPreload = [];
 
-        $imports = $this->convertEntriesToImports($entries);
+        $imports = $this->convertEntriesToImports(array_values($entries));
```
nicolas-grekas: this change is not needed, isn't it?

nicolas-grekas (approved):

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
weaverryan: No - I realized that, in this one case, I was passing in a hash to the method and in other cases I was passing in an array. I normalized it. But, yea, it's unnecessary. I've just reverted it.

nicolas-grekas (approved):

nicolas-grekas:
Thank you @weaverryan.

PR #50358, opened 2023-05-17
[AssetMapper] Not allowing installation alongside FwBundle less than 6.3

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | Not needed

Hi!

AssetMapper shouldn't be installed alongside FrameworkBundle 6.2 or lower. I believe adding this conflict is pretty standard.

Thanks!

### DIFF
--- a/src/Symfony/Component/AssetMapper/composer.json
+++ b/src/Symfony/Component/AssetMapper/composer.json
@@ -29,6 +29,9 @@
         "symfony/http-foundation": "^5.4|^6.0",
         "symfony/http-kernel": "^5.4|^6.0"
     },
+    "conflict": {
+        "symfony/framework-bundle": "<6.3"
+    },
     "autoload": {
         "psr-4": { "Symfony\\Component\\AssetMapper\\": "" },
         "exclude-from-classmap": [


nicolas-grekas:
Are you sure we need this? We don't reference bundles in components usually. Can't this be dealt with at the recipe level if that's what bothers you?

weaverryan:
> We don't reference bundles in components usually

Hmm. I see that. This isn't strictly needed. Have we just ignored this in the past and it's been ok? Like, when something new like Notifier is added, I guess we allowed it to be installed with a FWBundle that didn't support its config? Not a huge deal if we can't think of a good way to fix it. Actually, I think AssetMapper will TwigBundle 6.3: if 6.2 is installed, the `{{ importmap() }}` won't exist and installing AssetMapper's recipe will result in an error. Maybe best to ignore this problem.

nicolas-grekas:
Recipes can conflict, so that's how we should deal with that, isn't it?

weaverryan:
I forgot about that - thank you!

PR #50362, opened 2023-05-18
[FrameworkBundle] Fix Workflow without a marking store definition uses marking store definition of previously defined workflow

### LINKED ISSUE #48099: [FrameworkBundle] Workflow without a marking store definition uses marking store definition of previously defined workflow

4.4.44

### Description

Today we noticed some weird behaviour with one of our workflows: Suddenly it started complaining about the marking store method not existing, which was true, since we use the `getStatus` method for that workflow and it was complaining that the `getState` method did not exist. We just introduced a new workflow that actually does use the `getState` method, so that probably had something to do with it.

After some digging it seems that a workflow without a marking store definition will use a previously defined marking store, because the `$markingStoreDefinition` variable is used inside a loop in the `FrameworkExtension` in the FrameworkBundle and is never unset.

I haven't tested this with any other versions besides 4.4.44, but from looking at the latest version of the `FrameworkExtension`, it looks like this is still an issue.

### How to reproduce

1. Create two empty classes `App\Entity\ObjectA` and `App\Entity\ObjectB`

2. Define the following two workflows:
```
framework:
    workflows:
        workflow_a:
            type: 'state_machine'
            marking_store:
                type: 'method'
                property: 'status'
            supports:
                - App\Entity\ObjectA
            initial_marking: start
            places:
                - start
                - finish
            transitions:
                finish:
                    from: start
                    to: finish
        workflow_b:
            type: 'state_machine'
            supports:
                - App\Entity\ObjectB
            initial_marking: start
            places:
                - start
                - finish
            transitions:
                finish:
                    from: start
                    to: finish
```

3. Request the marking store for both of them
4. You will notice that workflow_b will use the same marking store as workflow_a (not same instance, but same definition)

### Possible Solution

Either set `$markingStoreDefinition` to `null` initially or unset it after it won't be used anymore.

The piece of code in question:
```
            // Create MarkingStore
            if (isset($workflow['marking_store']['type'])) {
                $markingStoreDefinition = new ChildDefinition('workflow.marking_store.'.$workflow['marking_store']['type']);
                if ('method' === $workflow['marking_store']['type']) {
                    $markingStoreDefinition->setArguments([
                        'state_machine' === $type, // single state
                        $workflow['marking_store']['property'] ?? 'marking',
                    ]);
                } else {
                    foreach ($workflow['marking_store']['arguments'] as $argument) {
                        $markingStoreDefinition->addArgument($argument);
                    }
                }
            } elseif (isset($workflow['marking_store']['service'])) {
                $markingStoreDefinition = new Reference($workflow['marking_store']['service']);
            }

            // Create Workflow
            $workflowDefinition = new ChildDefinition(sprintf('%s.abstract', $type));
            $workflowDefinition->replaceArgument(0, new Reference(sprintf('%s.definition', $workflowId)));
            if (isset($markingStoreDefinition)) {
                $workflowDefinition->replaceArgument(1, $markingStoreDefinition);
            }
```

### Additional Context

First dump is workflow_a, second dump is workflow_b

Screenshot 2022-11-03 at 16 33 15

krciga22:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #48099
| License       | MIT

Fixed a bug in the `Framework Bundle` with the `registerWorkflowConfiguration()` method not initializing `$workflowMarkingStore` to null when processing each workflow. This results in subsequent workflows reusing the previously set `$workflowMarkingStore` value if one has not been explicitly configured.

In the example below, `workflow_b` has no marking_store configured, so it will utilize the marking_store configuration for `workflow_a` instead of the default `MethodMarkingStore` due to this bug.

```yaml
framework:
    workflows:
        workflow_a:
            type: 'state_machine'
            marking_store:
                type: 'method'
                property: 'status'
        workflow_b:
            type: 'state_machine'
```

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
@@ -920,6 +920,7 @@ class FrameworkExtension extends Extension
             $workflows[$workflowId] = $definitionDefinition;
 
             // Create MarkingStore
+            $markingStoreDefinition = null;
             if (isset($workflow['marking_store']['type'])) {
                 $markingStoreDefinition = new ChildDefinition('workflow.marking_store.method');
                 $markingStoreDefinition->setArguments([
@@ -933,7 +934,7 @@ class FrameworkExtension extends Extension
             // Create Workflow
             $workflowDefinition = new ChildDefinition(sprintf('%s.abstract', $type));
             $workflowDefinition->replaceArgument(0, new Reference(sprintf('%s.definition', $workflowId)));
-            $workflowDefinition->replaceArgument(1, $markingStoreDefinition ?? null);
+            $workflowDefinition->replaceArgument(1, $markingStoreDefinition);
             $workflowDefinition->replaceArgument(3, $name);
             $workflowDefinition->replaceArgument(4, $workflow['events_to_dispatch']);
             $workflowDefinition->addTag('container.private', [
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Thank you @krciga22.

PR #50363, opened 2023-05-19
[AssetMapper] Adding "path" option to importmap:require

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | Still TODO

The `path` option was added by @dunglas - it was always intended to be there. I forgot about it and (mostly) failed to add it. Now I've found a use-case for it when installing UX bundles (i.e. call the command to update the user's importmap with a pointer to an asset inside of the bundle).

Thanks!


dunglas (approved):

nicolas-grekas:
Thank you @weaverryan.

PR #50371, opened 2023-05-19
[PhpUnitBridge] Deprecation bridge for PHPUnit 10.2

### LINKED ISSUE #49069: [PHPUnit bridge] Support for PHPUnit 11.1+
### Description

PHPUnit 10 is around the corner (possibly released next month), and it contains a complete overhaul of its internals, particularly the new event system: https://github.com/sebastianbergmann/phpunit/issues/4676 

The new system also provides an easy way too hook into those events: https://github.com/sebastianbergmann/phpunit/blob/2a89c3c62c4d94fa92b72cb3693ca4fa057e4517/src/Event/Facade.php#L40
Among the events, there are even multiple dedicated to deprecations:
 * triggered by `E_USER_DEPRECATED`: https://github.com/sebastianbergmann/phpunit/blob/main/src/Event/Events/Test/Issue/DeprecationTriggered.php
 * triggered by `E_DEPRECATED`: https://github.com/sebastianbergmann/phpunit/blob/main/src/Event/Events/Test/Issue/PhpDeprecationTriggered.php
 * triggered manually by PHPUnit itself: https://github.com/sebastianbergmann/phpunit/blob/main/src/Event/Events/Test/Issue/PhpunitDeprecationTriggered.php

This maybe warrants a discussion here on how to tackle this new release, and how to leverage all this new features.

### Example

_No response_

mondrake:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | tbd 
| Tickets       | Fix #49069  
| License       | MIT

This is a first _attempt_ at implementing the deprecation bridge for PHPUnit 10, re https://github.com/symfony/symfony/issues/49069. Given the *massive* changes in PHPUnit internals, the deprecation bridge will have, one way or another, to change accordingly.

So ATM this is more to share my findings to enable a discussion on _what_ the bridge should look like.

The PR is functional, but only partially, and totally untested.

Key points:
1. Listeners and hooks are gone in PHPUnit 10. In their place, an internal event system is introduced. The error handler also dramatically changed, and now only converts `E_*` errors into events that are then channeled through the new system, i.e. dispatched to subscribers. So `SymfonyTestsListener` and `DeprecationErrorHandler` implementations cannot be just reused. ==> My proposal here is to replace these, for the deprecation bridge, with an event collector and event subscribers.
2. The test objects have been armoured, it's no longer possible to change their state and make a passed test fail or error. At least, I could not find a way. This means probably that reporting of any fail/errors found by the bridge should be actually deferred to the end of the test run. No more 'E' or 'F' on the PHPUnit cli. ==> ~~For the time being, I made so that where errors or fails should be triggered, we have a `dump()` instead. But that needs to change.~~ Errors and failures due to the bridge are reported at report-out and no longer as errors/failures in the cli.
3. PHPUnit's event system does not seem to have a way to allow subscribers to stop event propagation. This means that no matter what the deprecation bridge will capture, PHPUnit itself will perform its own processing and reporting. ATM PHPUnit does not do anything with silenced deprecations (i.e. when the `trigger_error()` call is prefixed with the `@` symbol). But unsilenced deprecations would be reported by both the bridge (according to its configuration) and by PHPUnit itself (unconditionally).
4. ~~I do not know yet (have not investigated) whether it's possible to influence PHPUnit's cli exit code.~~
5. PHPUnit version 10.2 ~~(unreleased yet)~~ is the minimum version for this PR. That's because earlier than that, silenced deprecations were not even processed by the error handler, and that was only fixed in https://github.com/sebastianbergmann/phpunit/issues/5328

### DIFF
--- a/src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler.php
+++ b/src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler.php
@@ -284,7 +284,7 @@ class DeprecationErrorHandler
      *
      * @return string
      */
-    private static function colorize($str, $red)
+    public static function colorize($str, $red)
     {
         if (!self::hasColorSupport()) {
             return $str;
--- a/src/Symfony/Bridge/PhpUnit/ExpectDeprecationTrait.php
+++ b/src/Symfony/Bridge/PhpUnit/ExpectDeprecationTrait.php
@@ -12,19 +12,28 @@
 namespace Symfony\Bridge\PhpUnit;
 
 use Symfony\Bridge\PhpUnit\Legacy\ExpectDeprecationTraitBeforeV8_4;
+use Symfony\Bridge\PhpUnit\Legacy\ExpectDeprecationTraitForV10_2;
 use Symfony\Bridge\PhpUnit\Legacy\ExpectDeprecationTraitForV8_4;
 
-if (version_compare(\PHPUnit\Runner\Version::id(), '8.4.0', '<')) {
+if (version_compare(\PHPUnit\Runner\Version::id(), '10.2.0', '>=')) {
+    /**
+     * @method void expectDeprecation(string $message)
+     */
     trait ExpectDeprecationTrait
     {
-        use ExpectDeprecationTraitBeforeV8_4;
+        use ExpectDeprecationTraitForV10_2;
     }
-} else {
+} elseif (version_compare(\PHPUnit\Runner\Version::id(), '8.4.0', '>=')) {
     /**
      * @method void expectDeprecation(string $message)
      */
     trait ExpectDeprecationTrait
     {
         use ExpectDeprecationTraitForV8_4;
+    }
+} else {
+    trait ExpectDeprecationTrait
+    {
+        use ExpectDeprecationTraitBeforeV8_4;
     }
 }
--- a/src/Symfony/Bridge/PhpUnit/Legacy/EventSubscribers/ApplicationFinishedSubscriber.php
+++ b/src/Symfony/Bridge/PhpUnit/Legacy/EventSubscribers/ApplicationFinishedSubscriber.php
@@ -1,0 +1,17 @@
+<?php declare(strict_types=1);
+
+namespace Symfony\Bridge\PhpUnit\Legacy\EventSubscribers;
+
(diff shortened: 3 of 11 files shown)


mondrake:
Made changes to report error/failures at the report-out time, and advocated exiting early if bridge's status is higher than phpunit's exit code.

stof:
The deprecations being expected by legacy tests should still be checked during the test run. This should be implemented by having a post condition method (i.e. a protected method with the `PostCondition` attribute of PHPUnit or the `@postCondition` tag in phpdoc) in the trait of the bridge that triggers the assertions on the expected deprecations for the current test. this would allow running those assertions before the test status is frozen.
this is the same extension point than used to verify mock objects expectations.

alexisLefebvre:
This PR needs to be rebased, it currently has 152 commits and 3762 files changed.

OskarStark:
Related
* https://github.com/sebastianbergmann/phpunit/issues/5532
* https://github.com/sebastianbergmann/phpunit-documentation-english/issues/333

rvanlaak:
> Right now one can filter the deprecations and get them as grouped output via [restrictDeprecations](https://docs.phpunit.de/en/10.3/configuration.html#the-restrictdeprecations-attribute) option 👍

@OskarStark this reference to your above ticket on the phpunit repo is really worth mentioning over here! 🙌  Is that feature introduced in PHPUnit for all PHP versions that are supported by the Symfony PhpUnitBridge?

OskarStark:
>  Is that feature introduced in PHPUnit for all PHP versions that are supported by the Symfony PhpUnitBridge?

I am not sure and didn't test the feature yes, I just proposed the idea and @sebastianbergmann implemented it

mondrake:
> Is that feature introduced in PHPUnit for all PHP versions that are supported by the Symfony PhpUnitBridge?

It will be part of PHPUnit 10.5 (release due Dec 1 2023), that requires PHP >=8.1. Do not expect backports.

mondrake:
For the record, in Drupal we are working on a possible alternative without the bridge, implementing an error handler shim on top of PHPUnit's one, for every test. Sharing in case something there might be reused here.

See https://www.drupal.org/project/drupal/issues/3417066.

PR #50380, opened 2023-05-21
[HttpFoundation] Add `peekMultiple` method to FlashBag

### LINKED ISSUE #49659: FlashBag peekMultiple() method

alexmerlin:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #49659
| License       | MIT
| Doc PR        | -

This PR adds the `peekMultiple` method to `FlashBag` and `AutoExpireFlashBag`, as requested in #49659.


alexmerlin:
Hey @derrabus, thank you for looking into this PR.
Do you have any insight on what should I fix in order to go green with the tests?
From what I see, the tests did not fail, it's just an error that the `Process completed with exit code 1`.

andreybolonin:
@alexmerlin As I understand 6.2 is feature freeze stage and no adding new features
You should change 6.3 to 6.4 after 31 may

derrabus:
> Do you have any insight on what should I fix in order to go green with the tests?

I don't think that the failure is related to your changes. You can ignore the red unit tests.

nicolas-grekas:
Closing this one. Thanks for the work, and thanks for your patience with how long it sat without a review.

`peekAll()` already returns every flash without clearing them, so selecting a few of them is `array_intersect_key($bag->peekAll(), array_flip($types))`, and the templates most people use this from already accept a list: `app.flashes(['error', 'warning'])`. That is not enough to justify new surface on `FlashBagInterface`, which is the other half of the problem here: adding a method to a public interface breaks every third-party implementation, so it would need the `@method` treatment plus an entry in `UPGRADE-9.0.md` rather than a plain addition.

PR #50381, opened 2023-05-21
[HttpKernel] fix exception wording

### LINKED ISSUE #50375: ArgumentResolver exception has incorrect wording

6.2.x and 6.3.x

### Description

The exception in [ArgumentResolver](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver.php#L106) reads like the following:

>Controller "%s" requires that you provide a value for the "$%s" argument. Either the argument is nullable and no null value has been provided, no default value has been provided or because there is a non optional argument after this one.

The "because" is clearly superfluous, and "non optional" should be written as "non-optional".

### How to reproduce

Not so relevant, since the error is inside the code.
Anyway, I stumbled upon this problem when I tried to use the new `#MapQueryParameter` feature in Symfony 6.3-beta3 (there was a bug, fixed in the current dev but still not released).

### Possible Solution

Replace with:

>Controller "%s" requires that you provide a value for the "$%s" argument. Either the argument is nullable and no null value has been provided, no default value has been provided or there is a non-optional argument after this one.

### Additional Context

_No response_

garak:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50375
| License       | MIT
| Doc PR        | not needed

No description needed, it should be quite clear from diff and/or related issue

### DIFF
--- a/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver.php
+++ b/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver.php
@@ -78,7 +78,7 @@ final class ArgumentResolver implements ArgumentResolverInterface
                 $representative = get_debug_type($representative);
             }
 
-            throw new \RuntimeException(sprintf('Controller "%s" requires that you provide a value for the "$%s" argument. Either the argument is nullable and no null value has been provided, no default value has been provided or because there is a non optional argument after this one.', $representative, $metadata->getName()));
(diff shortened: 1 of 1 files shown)


derrabus (approved):

wouterj:
Thanks @garak!

PR #29117, opened 2018-11-07
[Serializer] Add CompiledClassMetadataFactory

fbourigault:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | ?
| License       | MIT
| Doc PR        | todo (issue: https://github.com/symfony/symfony-docs/issues/10706)

This introduce a dumped `ClassMetadataFactoryInterface` implementation to speed up the serializer by leveraging [PHP7 immutable array](https://blog.blackfire.io/php-7-performance-improvements-immutable-arrays.html).

Like for https://github.com/symfony/symfony/pull/28865, if the user have the opcache extension enabled, the compilation time will be skipped. The user will also have a performance boost when not using opcache as we are no longer fetching `ClassMetadata` from the PSR-6 cache. 

This allow to speed up the normalization (without opcache) by 9-12% depending on how many objects are involved in the graph:

- [SymfonyObjectNormalizerBenchmark, 100 iterations with complexity of 1](https://blackfire.io/profiles/compare/d937a9cc-eebf-47eb-be90-c8e65cdf12b3/graph)
- [SymfonyObjectNormalizerBenchmark, 3 iterations with complexity of 60](https://blackfire.io/profiles/compare/d490542c-9a79-48a0-b7bc-1ed3ca6a9148/graph)

On the `FrameworkBundle` side, I suggest to add a `CacheWarmer` to dump the metadata array from configured class list. The list could have a _good_ default which will load the classes found in `src/Entity`.

## Dumping the `ClassMetadata`
```php
$classMetadatas = [];

foreach([Category::class, Comment::class, Forum::class, Thread::class, User::class] as $class) {
    $classMetadatas[] = $this->classMetadataFactory->getMetadataFor($class);
}

file_put_contents('dumped.php', $this->classMetadataFactoryCompiler->compile($classMetadatas));
```

## Using the dumped `ClassMetadata`
```php
$classMetadataFactory = new CompiledClassMetadataFactory(
    'dumped.php',
    new CacheClassMetadataFactory(
        new ClassMetadataFactory(new AnnotationLoader(new AnnotationReader())),
        new ApcuAdapter('SymfonyMetadata')
    )
);
```
# To do

- [x] Tests.
- [x] Cache warmer.
- [x] Documentation.
- [x] Changelog entry.

PR #50391, opened 2023-05-22
[FrameworkBundle][HttpKernel] Introduce `$buildDir` argument to `WarmableInterface::warmup` to warm read-only artefacts in `build_dir`

### LINKED ISSUE #50357: [FrameworkBundle] CacheWarmers are never executed in `kernel.build_dir`
### Description

While working on #50253, I realised that CacheWarmers are never executed in the `kernel.build_dir` (except in the case of the bug that I fixed in #50253).

For now, in a brand new Symfony project, only the compiled container is dumped in the `build_dir`. Looking at the documentation for `kernel.build_dir` ([Build directory](https://symfony.com/doc/current/reference/configuration/kernel.html#build-directory)), it is described as 
> This directory can be used to separate read-only cache (i.e. the compiled container) from read-write cache (i.e. [cache pools](https://symfony.com/doc/current/cache.html)).

However, there's plenty of other ressources that could be considered as read-only IMHO: the router cache, Doctrine proxy classes, Symfony Config classes, translations, and probably many more.

Additionally, non-optional CacheWarmers are run during container init, in the `cache_dir`, which is erased if you run `cache:clear` command. Since the non-optional CacheWarmers are not executed in `cache:clear` command, because they run during container init, all the artefacts they produce are actually lost if you run `cache:clear` without an already existing container, which is happening on every composer install (some Doctrine CacheWarmers are concerned by this at least).

Besides, the mechanism of CacheWarmer seems a bit weird to me in some implementations of the interface. More specifically, the `Symfony\Bundle\FrameworkBundle\Routing\Router` that implements the `WarmableInterface`.

For reference, here's its implementation of the `warmup` function
```php
    /**
     * @return string[] A list of classes to preload on PHP 7.4+
     */
    public function warmUp(string $cacheDir): array
    {
        $currentDir = $this->getOption('cache_dir');

        // force cache generation
        $this->setOption('cache_dir', $cacheDir);
        $this->getMatcher();
        $this->getGenerator();

        $this->setOption('cache_dir', $currentDir);

        return [
            $this->getOption('generator_class'),
            $this->getOption('matcher_class'),
        ];
    }
```

It seems to me that it's really error prone to have such back and forth with the `cache_dir` option. If someone changes the `cache_dir` setting of the Router, it won't be properly warmed up anymore, without any notice to the user.

So, to sum up a little bit, I can see three different issues, that may or may not be solve together.
- `build_dir` is under utilised and many read-only warm ups could make use of it
- Non-optional warmer artefacts are lost on initial `cache:clear`
- `WarmableInterface` does not protect from misconfiguration. 

I'm willing to contribute and propose a change to solve at least the first issue but would like to get some opinions on the best way to solve it beforehand :)

### Example

```shell
s new --dir symfony_build_dir_problem
cd ./symfony_build_dir_problem
echo '<?php

namespace App;

{
    use MicroKernelTrait;

    public function getBuildDir(): string
    {
        return $this->getProjectDir()."/var/build/".$this->environment;
    }
}' > ./src/Kernel.php
composer install -q
```

Then see the content of `var/build/dev` and `var/cache/dev`.

okhoshi:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #50357 
| License       | MIT

See #50357 for the details and the reproduction steps. In this PR, I also moved `ConfigBuilderCacheWarmer` to use the `buildDir` argument.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ConfigBuilderCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ConfigBuilderCacheWarmer.php
@@ -38,11 +38,17 @@ class ConfigBuilderCacheWarmer implements CacheWarmerInterface
     }
 
     /**
-     * @return string[]
+     * @param string|null $buildDir
      */
-    public function warmUp(string $cacheDir): array
+    public function warmUp(string $cacheDir /* , string $buildDir = null */): array
     {
-        $generator = new ConfigBuilderGenerator($this->kernel->getBuildDir());
+        $buildDir = 1 < \func_num_args() ? func_get_arg(1) : null;
+
+        if (!$buildDir) {
+            return [];
+        }
+
+        $generator = new ConfigBuilderGenerator($buildDir);
 
         foreach ($this->kernel->getBundles() as $bundle) {
             $extension = $bundle->getContainerExtension();
--- a/src/Symfony/Bundle/FrameworkBundle/Command/CacheWarmupCommand.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Command/CacheWarmupCommand.php
@@ -74,7 +74,8 @@ EOF
 
         $preload = $this->cacheWarmer->warmUp($cacheDir);
 
-        if ($preload && file_exists($preloadFile = $cacheDir.'/'.$kernel->getContainer()->getParameter('kernel.container_class').'.preload.php')) {
+        $buildDir = $kernel->getContainer()->getParameter('kernel.build_dir');
+        if ($preload && $cacheDir === $buildDir && file_exists($preloadFile = $buildDir.'/'.$kernel->getContainer()->getParameter('kernel.container_class').'.preload.php')) {
             Preloader::append($preloadFile, $preload);
         }
 
--- a/src/Symfony/Component/HttpKernel/CacheWarmer/CacheWarmerAggregate.php
+++ b/src/Symfony/Component/HttpKernel/CacheWarmer/CacheWarmerAggregate.php
@@ -48,8 +48,17 @@ class CacheWarmerAggregate implements CacheWarmerInterface
         $this->onlyOptionalsEnabled = $this->optionalsEnabled = true;
     }
 
-    public function warmUp(string $cacheDir, SymfonyStyle $io = null): array
+    /**
+     * @param string|null $buildDir
+     */
+    public function warmUp(string $cacheDir, string|SymfonyStyle $buildDir = null, SymfonyStyle $io = null): array
     {
+        if ($buildDir instanceof SymfonyStyle) {
+            trigger_deprecation('symfony/http-kernel', '6.4', 'Passing a "%s" as second argument of "%s()" is deprecated, pass it as third argument instead, after the build directory.', SymfonyStyle::class, __METHOD__);
+            $io = $buildDir;
+            $buildDir = null;
+        }
+
         if ($collectDeprecations = $this->debug && !\defined('PHPUNIT_COMPOSER_INSTALL')) {
             $collectedLogs = [];
             $previousHandler = set_error_handler(function ($type, $message, $file, $line) use (&$collectedLogs, &$previousHandler) {
@@ -96,8 +105,8 @@ class CacheWarmerAggregate implements CacheWarmerInterface
                 }
 
                 $start = microtime(true);
-                foreach ((array) $warmer->warmUp($cacheDir) as $item) {
-                    if (is_dir($item) || (str_starts_with($item, \dirname($cacheDir)) && !is_file($item))) {
+                foreach ((array) $warmer->warmUp($cacheDir, $buildDir) as $item) {
+                    if (is_dir($item) || (str_starts_with($item, \dirname($cacheDir)) && !is_file($item)) || ($buildDir && str_starts_with($item, \dirname($buildDir)) && !is_file($item))) {
                         throw new \LogicException(sprintf('"%s::warmUp()" should return a list of files or classes but "%s" is none of them.', $warmer::class, $item));
                     }
                     $preload[] = $item;
--- a/.github/expected-missing-return-types.diff
+++ b/.github/expected-missing-return-types.diff
@@ -8109,11 +8109,11 @@ diff --git a/src/Symfony/Component/HttpKernel/CacheWarmer/CacheWarmerInterface.p
 diff --git a/src/Symfony/Component/HttpKernel/CacheWarmer/WarmableInterface.php b/src/Symfony/Component/HttpKernel/CacheWarmer/WarmableInterface.php
 --- a/src/Symfony/Component/HttpKernel/CacheWarmer/WarmableInterface.php
 +++ b/src/Symfony/Component/HttpKernel/CacheWarmer/WarmableInterface.php
-@@ -24,4 +24,4 @@ interface WarmableInterface
+@@ -27,4 +27,4 @@ interface WarmableInterface
       * @return string[] A list of classes or files to preload on PHP 7.4+
       */
--    public function warmUp(string $cacheDir);
-+    public function warmUp(string $cacheDir): array;
+-    public function warmUp(string $cacheDir /* , string $buildDir = null */);
++    public function warmUp(string $cacheDir /* , string $buildDir = null */): array;
  }
 diff --git a/src/Symfony/Component/HttpKernel/DataCollector/DataCollector.php b/src/Symfony/Component/HttpKernel/DataCollector/DataCollector.php
 --- a/src/Symfony/Component/HttpKernel/DataCollector/DataCollector.php
--- a/src/Symfony/Bridge/Doctrine/CacheWarmer/ProxyCacheWarmer.php
+++ b/src/Symfony/Bridge/Doctrine/CacheWarmer/ProxyCacheWarmer.php
@@ -38,9 +38,9 @@ class ProxyCacheWarmer implements CacheWarmerInterface
     }
 
     /**
-     * @return string[] A list of files to preload on PHP 7.4+
+     * @param string|null $buildDir
      */
-    public function warmUp(string $cacheDir): array
+    public function warmUp(string $cacheDir /* , string $buildDir = null */): array
     {
         $files = [];
         foreach ($this->registry->getManagers() as $em) {
--- a/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
+++ b/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
@@ -32,6 +32,7 @@ CHANGELOG
  * Add `--exclude` option to the `cache:pool:clear` command
  * Add parameters deprecations to the output of `debug:container` command
  * Change `framework.asset_mapper.importmap_polyfill` from a URL to the name of an item in the importmap
+ * Provide `$buildDir` when running `CacheWarmer` to build read-only resources
 
 6.3
 ---
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/AbstractPhpFileCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/AbstractPhpFileCacheWarmer.php
@@ -35,15 +35,16 @@ abstract class AbstractPhpFileCacheWarmer implements CacheWarmerInterface
     }
 
     /**
-     * @return string[] A list of classes to preload on PHP 7.4+
+     * @param string|null $buildDir
      */
-    public function warmUp(string $cacheDir): array
+    public function warmUp(string $cacheDir /* , string $buildDir = null */): array
     {
+        $buildDir = 1 < \func_num_args() ? func_get_arg(1) : null;
         $arrayAdapter = new ArrayAdapter();
 
         spl_autoload_register([ClassExistenceResource::class, 'throwOnRequiredClass']);
         try {
-            if (!$this->doWarmUp($cacheDir, $arrayAdapter)) {
+            if (!$this->doWarmUp($cacheDir, $arrayAdapter, $buildDir)) {
                 return [];
             }
         } finally {
@@ -78,7 +79,9 @@ abstract class AbstractPhpFileCacheWarmer implements CacheWarmerInterface
     }
 
     /**
+     * @param string|null $buildDir
+     *
      * @return bool false if there is nothing to warm-up
      */
-    abstract protected function doWarmUp(string $cacheDir, ArrayAdapter $arrayAdapter): bool;
+    abstract protected function doWarmUp(string $cacheDir, ArrayAdapter $arrayAdapter /* , string $buildDir = null */): bool;
 }
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/AnnotationsCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/AnnotationsCacheWarmer.php
@@ -44,7 +44,10 @@ class AnnotationsCacheWarmer extends AbstractPhpFileCacheWarmer
         parent::__construct($phpArrayFile);
     }
 
-    protected function doWarmUp(string $cacheDir, ArrayAdapter $arrayAdapter): bool
+    /**
+     * @param string|null $buildDir
+     */
+    protected function doWarmUp(string $cacheDir, ArrayAdapter $arrayAdapter /* , string $buildDir = null */): bool
     {
         $annotatedClassPatterns = $cacheDir.'/annotations.map';
 
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/CachePoolClearerCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/CachePoolClearerCacheWarmer.php
@@ -37,10 +37,7 @@ final class CachePoolClearerCacheWarmer implements CacheWarmerInterface
         $this->pools = $pools;
     }
 
-    /**
-     * @return string[]
-     */
-    public function warmUp(string $cacheDirectory): array
+    public function warmUp(string $cacheDir, string $buildDir = null): array
     {
         foreach ($this->pools as $pool) {
             if ($this->poolClearer->hasPool($pool)) {
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/RouterCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/RouterCacheWarmer.php
@@ -34,12 +34,12 @@ class RouterCacheWarmer implements CacheWarmerInterface, ServiceSubscriberInterf
         $this->container = $container;
     }
 
-    public function warmUp(string $cacheDir): array
+    public function warmUp(string $cacheDir, string $buildDir = null): array
     {
         $router = $this->container->get('router');
 
         if ($router instanceof WarmableInterface) {
-            return (array) $router->warmUp($cacheDir);
+            return (array) $router->warmUp($cacheDir, $buildDir);
         }
 
         throw new \LogicException(sprintf('The router "%s" cannot be warmed up because it does not implement "%s".', get_debug_type($router), WarmableInterface::class));
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/SerializerCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/SerializerCacheWarmer.php
@@ -39,7 +39,10 @@ class SerializerCacheWarmer extends AbstractPhpFileCacheWarmer
         $this->loaders = $loaders;
     }
 
-    protected function doWarmUp(string $cacheDir, ArrayAdapter $arrayAdapter): bool
+    /**
+     * @param string|null $buildDir
+     */
+    protected function doWarmUp(string $cacheDir, ArrayAdapter $arrayAdapter /* , string $buildDir = null */): bool
     {
         if (!$this->loaders) {
             return true;
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/TranslationsCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/TranslationsCacheWarmer.php
@@ -34,14 +34,16 @@ class TranslationsCacheWarmer implements CacheWarmerInterface, ServiceSubscriber
     }
 
     /**
-     * @return string[]
+     * @param string|null $buildDir
      */
-    public function warmUp(string $cacheDir): array
+    public function warmUp(string $cacheDir /* , string $buildDir = null */): array
     {
         $this->translator ??= $this->container->get('translator');
 
         if ($this->translator instanceof WarmableInterface) {
-            return (array) $this->translator->warmUp($cacheDir);
+            $buildDir = 1 < \func_num_args() ? func_get_arg(1) : null;
+
+            return (array) $this->translator->warmUp($cacheDir, $buildDir);
         }
 
         return [];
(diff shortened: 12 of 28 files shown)


nicolas-grekas (commented):
As discussed on Slack, I'd suggest adding a `string $buildDir = null` argument to the existing `warmUp` method (using a commented argument for BC). We'd pass `null` when only warming up, and the build-dir when building.

src/Symfony/Component/HttpKernel/CacheWarmer/CacheWarmerAggregate.php
```diff
@@ -96,7 +96,7 @@ public function warmUp(string $cacheDir, SymfonyStyle $io = null): array
                 }
 
                 $start = microtime(true);
-                foreach ((array) $warmer->warmUp($cacheDir) as $item) {
+                foreach ((array) $warmer->warmUp($cacheDir, null === $buildDir || $warmer->isOptional() ? null : $buildDir) as $item) {
```
nicolas-grekas: What's the rationale for tightening optional+build-dir concepts together? I'd prefer removing this extra check, unless reason.

nicolas-grekas (commented):
Here are some comment, and my review as a patch :)

Now that we have this, I'm wondering about each warmers: are they doing something for build-time or compilte-time? what's their purpose, etc; E.g `CompiledClassMetadataFactory` added in #29117 by @fbourigault is never wired? We should likely investigate this in a follow up PR this and the others too.

<details>

```diff
diff --git a/src/Symfony/Bridge/Doctrine/CacheWarmer/ProxyCacheWarmer.php b/src/Symfony/Bridge/Doctrine/CacheWarmer/ProxyCacheWarmer.php
index b148b9347b..baedf9c033 100644
--- a/src/Symfony/Bridge/Doctrine/CacheWarmer/ProxyCacheWarmer.php
+++ b/src/Symfony/Bridge/Doctrine/CacheWarmer/ProxyCacheWarmer.php
@@ -40,9 +40,7 @@ class ProxyCacheWarmer implements CacheWarmerInterface
     }
 
     /**
-     * @param string|null $buildDir directory to put generated source code that can be marked as read-only at runtime.
-     *                              The directory is not provided when only cache_dir should be warmed up.
-     * @return string[] A list of files to preload on PHP 7.4+
+     * @param string|null $buildDir
      */
     public function warmUp(string $cacheDir /* , string $buildDir = null */): array
     {
diff --git a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/AbstractPhpFileCacheWarmer.php b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/AbstractPhpFileCacheWarmer.php
index 49fd7f035d..98c281276b 100644
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/AbstractPhpFileCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/AbstractPhpFileCacheWarmer.php
@@ -35,17 +35,16 @@ abstract class AbstractPhpFileCacheWarmer implements CacheWarmerInterface
     }
 
     /**
-     * @param string|null $buildDir directory to put generated source code that can be marked as read-only at runtime.
-     *                              The directory is not provided when only cache_dir should be warmed up.
-     * @return string[] A list of classes to preload on PHP 7.4+
+     * @param string|null $buildDir
      */
     public function warmUp(string $cacheDir /* , string $buildDir = null */): array
     {
+        $buildDir = 1 < \func_num_args() ? func_get_arg(1) : null;
         $arrayAdapter = new ArrayAdapter();
 
         spl_autoload_register([ClassExistenceResource::class, 'throwOnRequiredClass']);
         try {
-            if (!$this->doWarmUp($cacheDir, $arrayAdapter)) {
+            if (!$this->doWarmUp($cacheDir, $arrayAdapter, $buildDir)) {
                 return [];
             }
         } finally {
@@ -80,7 +79,9 @@ abstract class AbstractPhpFileCacheWarmer implements CacheWarmerInterface
     }
 
     /**
+     * @param string|null $buildDir
+     *
      * @return bool false if there is nothing to warm-up
      */
-    abstract protected function doWarmUp(string $cacheDir, ArrayAdapter $arrayAdapter): bool;
+    abstract protected function doWarmUp(string $cacheDir, ArrayAdapter $arrayAdapter /* , string $buildDir = null */): bool;
 }
diff --git a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/AnnotationsCacheWarmer.php b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/AnnotationsCacheWarmer.php
index 279dc4ec9e..20533bb60e 100644
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/AnnotationsCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/AnnotationsCacheWarmer.php
@@ -44,7 +44,10 @@ class AnnotationsCacheWarmer extends AbstractPhpFileCacheWarmer
         parent::__construct($phpArrayFile);
     }
 
-    protected function doWarmUp(string $cacheDir, ArrayAdapter $arrayAdapter): bool
+    /**
+     * @param string|null $buildDir
+     */
+    protected function doWarmUp(string $cacheDir, ArrayAdapter $arrayAdapter /* , string $buildDir = null */): bool
     {
         $annotatedClassPatterns = $cacheDir.'/annotations.map';
 
diff --git a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/CachePoolClearerCacheWarmer.php b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/CachePoolClearerCacheWarmer.php
index 7cd02d1cc5..7498a82d1f 100644
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/CachePoolClearerCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/CachePoolClearerCacheWarmer.php
@@ -37,10 +37,7 @@ final class CachePoolClearerCacheWarmer implements CacheWarmerInterface
         $this->pools = $pools;
     }
 
-    /**
-     * @return string[]
-     */
-    public function warmUp(string $cacheDirectory, string $buildDir = null): array
+    public function warmUp(string $cacheDir, string $buildDir = null): array
     {
         foreach ($this->pools as $pool) {
             if ($this->poolClearer->hasPool($pool)) {
diff --git a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ConfigBuilderCacheWarmer.php b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ConfigBuilderCacheWarmer.php
index 12e24bd77c..225fea11b7 100644
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ConfigBuilderCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ConfigBuilderCacheWarmer.php
@@ -38,11 +38,13 @@ class ConfigBuilderCacheWarmer implements CacheWarmerInterface
     }
 
     /**
-     * @return string[]
+     * @param string|null $buildDir
      */
-    public function warmUp(string $cacheDir, string $buildDir = null): array
+    public function warmUp(string $cacheDir /* , string $buildDir = null */): array
     {
-        if (null === $buildDir) {
+        $buildDir = 1 < \func_num_args() ? func_get_arg(1) : null;
+
+        if (!$buildDir) {
             return [];
         }
 
diff --git a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/RouterCacheWarmer.php b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/RouterCacheWarmer.php
index 82a16c01d3..2af9a2fe80 100644
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/RouterCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/RouterCacheWarmer.php
@@ -34,16 +34,12 @@ class RouterCacheWarmer implements CacheWarmerInterface, ServiceSubscriberInterf
         $this->container = $container;
     }
 
-    /**
-     * @param string|null $buildDir directory to put generated source code that can be marked as read-only at runtime.
-     *                              The directory is not provided when only cache_dir should be warmed up.
-     */
-    public function warmUp(string $cacheDir /* , string $buildDir = null */): array
+    public function warmUp(string $cacheDir, string $buildDir = null): array
     {
         $router = $this->container->get('router');
 
         if ($router instanceof WarmableInterface) {
-            return (array) $router->warmUp($cacheDir);
+            return (array) $router->warmUp($cacheDir, $buildDir);
         }
 
         throw new \LogicException(sprintf('The router "%s" cannot be warmed up because it does not implement "%s".', get_debug_type($router), WarmableInterface::class));
diff --git a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/SerializerCacheWarmer.php b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/SerializerCacheWarmer.php
index 35595474ef..b47a48ce69 100644
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/SerializerCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/SerializerCacheWarmer.php
@@ -39,7 +39,10 @@ class SerializerCacheWarmer extends AbstractPhpFileCacheWarmer
         $this->loaders = $loaders;
     }
 
-    protected function doWarmUp(string $cacheDir, ArrayAdapter $arrayAdapter): bool
+    /**
+     * @param string|null $buildDir
+     */
+    protected function doWarmUp(string $cacheDir, ArrayAdapter $arrayAdapter /* , string $buildDir = null */): bool
     {
         if (!$this->loaders) {
             return true;
diff --git a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/TranslationsCacheWarmer.php b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/TranslationsCacheWarmer.php
index d3abbaca5b..39b1444b0e 100644
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/TranslationsCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/TranslationsCacheWarmer.php
@@ -34,16 +34,16 @@ class TranslationsCacheWarmer implements CacheWarmerInterface, ServiceSubscriber
     }
 
     /**
-     * @param string|null $buildDir directory to put generated source code that can be marked as read-only at runtime.
-     *                              The directory is not provided when only cache_dir should be warmed up.
-     * @return string[]
+     * @param string|null $buildDir
      */
     public function warmUp(string $cacheDir /* , string $buildDir = null */): array
     {
         $this->translator ??= $this->container->get('translator');
 
         if ($this->translator instanceof WarmableInterface) {
-            return (array) $this->translator->warmUp($cacheDir);
+            $buildDir = 1 < \func_num_args() ? func_get_arg(1) : null;
+
+            return (array) $this->translator->warmUp($cacheDir, $buildDir);
         }
 
         return [];
diff --git a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ValidatorCacheWarmer.php b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ValidatorCacheWarmer.php
index 3119c9942a..224e90985e 100644
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ValidatorCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ValidatorCacheWarmer.php
@@ -39,7 +39,10 @@ class ValidatorCacheWarmer extends AbstractPhpFileCacheWarmer
         $this->validatorBuilder = $validatorBuilder;
     }
 
-    protected function doWarmUp(string $cacheDir, ArrayAdapter $arrayAdapter): bool
+    /**
+     * @param string|null $buildDir
+     */
+    protected function doWarmUp(string $cacheDir, ArrayAdapter $arrayAdapter /* , string $buildDir = null */): bool
     {
         $loaders = $this->validatorBuilder->getLoaders();
         $metadataFactory = new LazyLoadingMetadataFactory(new LoaderChain($loaders), $arrayAdapter);
diff --git a/src/Symfony/Bundle/FrameworkBundle/Routing/Router.php b/src/Symfony/Bundle/FrameworkBundle/Routing/Router.php
index fe89fef6d6..3367ecec2b 100644
--- a/src/Symfony/Bundle/FrameworkBundle/Routing/Router.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Routing/Router.php
@@ -81,9 +81,7 @@ class Router extends BaseRouter implements WarmableInterface, ServiceSubscriberI
     }
 
     /**
-     * @param string|null $buildDir directory to put generated source code that can be marked as read-only at runtime.
-     *                              The directory is not provided when only cache_dir should be warmed up.
-     * @return string[] A list of classes to preload on PHP 7.4+
+     * @param string|null $buildDir
      */
     public function warmUp(string $cacheDir /* , string $buildDir = null */): array
     {
diff --git a/src/Symfony/Bundle/FrameworkBundle/Translation/Translator.php b/src/Symfony/Bundle/FrameworkBundle/Translation/Translator.php
index 70f5874307..04b56308f3 100644
--- a/src/Symfony/Bundle/FrameworkBundle/Translation/Translator.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Translation/Translator.php
@@ -95,9 +95,7 @@ class Translator extends BaseTranslator implements WarmableInterface
     }
 
     /**
-     * @param string|null $buildDir directory to put generated source code that can be marked as read-only at runtime.
-     *                              The directory is not provided when only cache_dir should be warmed up.
-     * @return string[]
+     * @param string|null $buildDir
      */
     public function warmUp(string $cacheDir /* , string $buildDir = null */): array
     {
diff --git a/src/Symfony/Bundle/SecurityBundle/CacheWarmer/ExpressionCacheWarmer.php b/src/Symfony/Bundle/SecurityBundle/CacheWarmer/ExpressionCacheWarmer.php
index c63c973e41..1cbb681c2d 100644
--- a/src/Symfony/Bundle/SecurityBundle/CacheWarmer/ExpressionCacheWarmer.php
+++ b/src/Symfony/Bundle/SecurityBundle/CacheWarmer/ExpressionCacheWarmer.php
@@ -35,9 +35,7 @@ class ExpressionCacheWarmer implements CacheWarmerInterface
     }
 
     /**
-     * @param string|null $buildDir directory to put generated source code that can be marked as read-only at runtime.
-     *                              The directory is not provided when only cache_dir should be warmed up.
-     * @return string[]
+     * @param string|null $buildDir
      */
     public function warmUp(string $cacheDir /* , string $buildDir = null */): array
     {
diff --git a/src/Symfony/Bundle/TwigBundle/CacheWarmer/TemplateCacheWarmer.php b/src/Symfony/Bundle/TwigBundle/CacheWarmer/TemplateCacheWarmer.php
index 1adabe6636..2ab801130b 100644
--- a/src/Symfony/Bundle/TwigBundle/CacheWarmer/TemplateCacheWarmer.php
+++ b/src/Symfony/Bundle/TwigBundle/CacheWarmer/TemplateCacheWarmer.php
@@ -36,9 +36,7 @@ class TemplateCacheWarmer implements CacheWarmerInterface, ServiceSubscriberInte
     }
 
     /**
-     * @param string|null $buildDir directory to put generated source code that can be marked as read-only at runtime.
-     *                              The directory is not provided when only cache_dir should be warmed up.
-     * @return string[] A list of template files to preload on PHP 7.4+
+     * @param string|null $buildDir
      */
     public function warmUp(string $cacheDir /* , string $buildDir = null */): array
     {
diff --git a/src/Symfony/Component/HttpKernel/CacheWarmer/CacheWarmerAggregate.php b/src/Symfony/Component/HttpKernel/CacheWarmer/CacheWarmerAggregate.php
index 638d9ad2c4..a672956e0f 100644
--- a/src/Symfony/Component/HttpKernel/CacheWarmer/CacheWarmerAggregate.php
+++ b/src/Symfony/Component/HttpKernel/CacheWarmer/CacheWarmerAggregate.php
@@ -48,8 +48,17 @@ class CacheWarmerAggregate implements CacheWarmerInterface
         $this->onlyOptionalsEnabled = $this->optionalsEnabled = true;
     }
 
-    public function warmUp(string $cacheDir, string $buildDir = null, SymfonyStyle $io = null): array
+    /**
+     * @param string|null $buildDir
+     */
+    public function warmUp(string $cacheDir, string|SymfonyStyle $buildDir = null, SymfonyStyle $io = null): array
     {
+        if ($buildDir instanceof SymfonyStyle) {
+            trigger_deprecation('symfony/http-kernel', '6.4', 'Passing a "%s" as second argument of "%s()" is deprecated, pass it as third argument instead, after the build directory.', SymfonyStyle::class, __METHOD__);
+            $io = $buildDir;
+            $buildDir = null;
+        }
+
         if ($collectDeprecations = $this->debug && !\defined('PHPUNIT_COMPOSER_INSTALL')) {
             $collectedLogs = [];
             $previousHandler = set_error_handler(function ($type, $message, $file, $line) use (&$collectedLogs, &$previousHandler) {
@@ -96,8 +105,8 @@ class CacheWarmerAggregate implements CacheWarmerInterface
                 }
 
                 $start = microtime(true);
-                foreach ((array) $warmer->warmUp($cacheDir, null === $buildDir || $warmer->isOptional() ? null : $buildDir) as $item) {
-                    if (is_dir($item) || (str_starts_with($item, \dirname($cacheDir)) && !is_file($item))) {
+                foreach ((array) $warmer->warmUp($cacheDir, $buildDir) as $item) {
+                    if (is_dir($item) || (str_starts_with($item, \dirname($cacheDir)) && !is_file($item)) || ($buildDir && str_starts_with($item, \dirname($buildDir)) && !is_file($item))) {
                         throw new \LogicException(sprintf('"%s::warmUp()" should return a list of files or classes but "%s" is none of them.', $warmer::class, $item));
                     }
                     $preload[] = $item;
diff --git a/src/Symfony/Component/HttpKernel/CacheWarmer/WarmableInterface.php b/src/Symfony/Component/HttpKernel/CacheWarmer/WarmableInterface.php
index 799705e42d..cd051b1add 100644
--- a/src/Symfony/Component/HttpKernel/CacheWarmer/WarmableInterface.php
+++ b/src/Symfony/Component/HttpKernel/CacheWarmer/WarmableInterface.php
@@ -21,8 +21,8 @@ interface WarmableInterface
     /**
      * Warms up the cache.
      *
-     * @param string|null $buildDir directory to put generated source code that can be marked as read-only at runtime.
-     *                              The directory is not provided when only cache_dir should be warmed up.
+     * @param string      $cacheDir Where warm-up artifacts should be stored
+     * @param string|null $buildDir Where read-only artifacts should go; null when called after compile-time
      *
      * @return string[] A list of classes or files to preload on PHP 7.4+
      */
diff --git a/src/Symfony/Component/Translation/DataCollectorTranslator.php b/src/Symfony/Component/Translation/DataCollectorTranslator.php
index bdd0e6ce4d..785e5b89ed 100644
--- a/src/Symfony/Component/Translation/DataCollectorTranslator.php
+++ b/src/Symfony/Component/Translation/DataCollectorTranslator.php
@@ -72,14 +72,14 @@ class DataCollectorTranslator implements TranslatorInterface, TranslatorBagInter
     }
 
     /**
-     * @param string|null $buildDir directory to put generated source code that can be marked as read-only at runtime.
-     *                              The directory is not provided when only cache_dir should be warmed up.
-     * @return string[]
+     * @param string|null $buildDir
      */
     public function warmUp(string $cacheDir /* , string $buildDir = null */): array
     {
+        $buildDir = 1 < \func_num_args() ? func_get_arg(1) : null;
+
         if ($this->translator instanceof WarmableInterface) {
-            return (array) $this->translator->warmUp($cacheDir);
+            return (array) $this->translator->warmUp($cacheDir, $buildDir);
         }
 
         return [];
```

</details>

src/Symfony/Component/HttpKernel/CacheWarmer/CacheWarmerAggregate.php
okhoshi: The check in itself is normally useless, because non-optional and optional warmers are never run together from the same place.
Non-optional warmers are always run at the end of `Symfony\Component\HttpKernel\Kernel::initializeContainer`.
Optional warmers (and only them) are run from the `CacheClearCommand` and `CacheWarmupCommand`.

The rationale of coupling optional to build-dir concepts is more about coupling the build-dir to the moment you run the warmers and the potential write state of the build-dir. It seemed more natural to me to consider the build-dir only writeable when the Kernel files are being compiled. The easiest way to know that the Kernel is compiling is to restrict the build-dir to non-optional warmers. Do you have any objection to that rationale ?

Writing this, I realise we could also create a third state with optional warmer with build-dir access, that would be enabled only when running `CacheClearCommand`, but I'm wondering if it is possible to end up in a situation where `CacheClearCommand` is executed with a read-only build-dir ?

fbourigault:
> Here are some comment, and my review as a patch :)
> 
> Now that we have this, I'm wondering about each warmers: are they doing something for build-time or compilte-time? what's their purpose, etc; E.g `CompiledClassMetadataFactory` added in #29117 by @fbourigault is never wired? We should likely investigate this in a follow up PR this and the others too.

Sadly, I never finished the framework integration of `CompiledClassMetadataFactory` :( Sadly I no longer have time to work on this topic anytime soon. It would be awesome if someone want to contribute the integration.

src/Symfony/Component/HttpKernel/CacheWarmer/CacheWarmerAggregate.php
okhoshi: After some reflexion time on my side, and playing a bit with possibilities, I removed completely the coupling between the optional and build-dir concepts, as you were suggesting 👍

src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ConfigBuilderCacheWarmer.php
```diff
@@ -79,6 +85,6 @@ private function dumpExtension(ExtensionInterface $extension, ConfigBuilderGener
 
     public function isOptional(): bool
     {
-        return true;
+        return false;
```
okhoshi:
```suggestion
        return true;
```

src/Symfony/Bundle/FrameworkBundle/Command/CacheWarmupCommand.php
```diff
@@ -66,9 +66,9 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             $this->cacheWarmer->enableOptionalWarmers();
         }
 
-        $preload = $this->cacheWarmer->warmUp($cacheDir = $kernel->getContainer()->getParameter('kernel.cache_dir'));
+        $preload = $this->cacheWarmer->warmUp($kernel->getContainer()->getParameter('kernel.cache_dir'));
 
-        if ($preload && file_exists($preloadFile = $cacheDir.'/'.$kernel->getContainer()->getParameter('kernel.container_class').'.preload.php')) {
+        if ($preload && file_exists($preloadFile = $kernel->getContainer()->getParameter('kernel.build_dir').'/'.$kernel->getContainer()->getParameter('kernel.container_class').'.preload.php')) {
```
nicolas-grekas: I guess this is broken: build_dir might be read-only at this stage.
Instead, I suppose we need to skip this part unless build_dir === $cacheDir

nicolas-grekas (commented):
please rebase

src/Symfony/Bundle/FrameworkBundle/Command/CacheWarmupCommand.php
nicolas-grekas: why did you resolve this comment? I see no changes in the code :)

nicolas-grekas:
```suggestion
        if ($kernel->getContainer()->getParameter('kernel.build_dir') !== $cacheDir) {
            return 0;
        }
        if ($preload && file_exists($preloadFile = $cacheDir.'/'.$kernel->getContainer()->getParameter('kernel.container_class').'.preload.php')) {
```

okhoshi: I did a change for that, in the 2nd commit, and somehow managed to get it reverted in the third 🤔 Fixing that

nicolas-grekas (approved):

nicolas-grekas:
Thank you @Okhoshi.

PR #46602, opened 2022-06-06
[SecurityBundle] Add service alias for legacy Security helper

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | yes
| Tickets       | -
| License       | MIT
| Doc PR        | -

The service alias for `Symfony\Component\Security\Core\Security` was removed in #46094 which causes existing code to break when upgrading to 6.2, eg:

```php
use Symfony\Component\Security\Core\Security;

class HomeController extends AbstractController
{
    public function __construct(private Security $security)
    {
    }
}
```
results in:
image

PR #50392, opened 2023-05-22
Move UriSigner from HttpKernel to HttpFoundation package

### LINKED ISSUE #50384: [HttpKernel] Move UriSigner to Http Foundation
### Description

Based on [Symfony Docs](https://github.com/symfony/symfony-docs/issues/18330) discussion. I want to open a discussion here to maybe move the `UriSigner` class to the `http-foundation`. Think the `HttpKernel` is a little bit unexpected place for the Signer. Would have expected it to be part of the routing component or http foundation component. Atleast I would have searched there for its documentation.

I did stumble over the [UrlHelper](https://symfony.com/doc/current/components/http_foundation.html#generating-relative-and-absolute-urls) and would think the **UriSigner** could be side by side to it.

What do you think?

### Example

```diff
-Symfony\Component\HttpKernel\UriSigner;
+Symfony\Component\HttpFoundation\UriSigner;
```

Reference: [PR Introduction of UriSigner](https://github.com/symfony/symfony/pull/6459)

alexander-schranz:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | Fix #50384
| License       | MIT
| Doc PR        | TODO

Move UriSigner from HttpKernel to HttpFoundation package as discussed in https://github.com/symfony/symfony/issues/50384.


src/Symfony/Component/HttpFoundation/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ CHANGELOG
 6.3
 ---
 
+ * Add `UriSigner` from HttpKernel
```
nicolas-grekas:
```suggestion
 * Add `UriSigner`
```

src/Symfony/Component/HttpKernel/UriSigner.php
```diff
@@ -12,11 +12,14 @@
...
 
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\UriSigner as HttpFoundationUriSigner;
 
 /**
  * Signs URIs.
  *
  * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @deprecated since Symfony 6.3, use {@link HttpFoundationUriSigner} instead
```
nicolas-grekas: 6.3

src/Symfony/Component/HttpFoundation/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ CHANGELOG
 6.3
```
nicolas-grekas, line 11: should be moved to a new 6.4 section

src/Symfony/Bridge/Twig/Tests/Extension/HttpKernelExtensionTest.php
```diff
@@ -18,10 +18,10 @@
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\UriSigner;
```
nicolas-grekas, line 21: this requires a version bump in composer.json

src/Symfony/Bundle/FrameworkBundle/Resources/config/services.php
```diff
@@ -35,6 +35,7 @@
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
+use Symfony\Component\HttpFoundation\UriSigner;
```
nicolas-grekas, line 38: this requires a version bump in composer.json

src/Symfony/Component/HttpFoundation/Tests/UriSignerTest.php
```diff
@@ -9,11 +9,11 @@
...
 
-namespace Symfony\Component\HttpKernel\Tests;
+namespace Symfony\Component\HttpFoundation\Tests;
 
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\HttpFoundation\Request;
-use Symfony\Component\HttpKernel\UriSigner;
+use Symfony\Component\HttpFoundation\UriSigner;
 
 class UriSignerTest extends TestCase
```
nicolas-grekas, line 18: should be duplicated, not moved
the legacy one should be marked `@group legacy`

src/Symfony/Component/HttpFoundation/UriSigner.php
```diff
@@ -0,0 +1,31 @@
...
+namespace Symfony\Component\HttpFoundation;
+
+use Symfony\Component\HttpKernel\UriSigner as HttpKernelUriSigner;
+
+/**
+ * Signs URIs.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+class UriSigner extends HttpKernelUriSigner
```
nicolas-grekas: we should rather duplicate the code IMHO, we can't make http-kernel a dep of http-foundation

src/Symfony/Component/HttpKernel/Tests/EventListener/FragmentListenerTest.php
```diff
@@ -13,11 +13,11 @@
 
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\UriSigner;
```
nicolas-grekas, line 16: requires a version bump

src/Symfony/Component/HttpKernel/UriSigner.php.orig
```diff
@@ -0,0 +1,116 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
```
nicolas-grekas: bad push :)

src/Symfony/Component/HttpFoundation/UriSigner.php
alexander-schranz: Would work to extend the HttpKernel UriSigner from the HttpFoundation UriSigner.

I first did have that but now I don't remember why I did switch it 🙈

alexander-schranz: I remember now the only problem then would be that the `uri_signer` service need to link to the old http kernel class else it would be a bc break. So only the `Symfony\Component\HttpFoundation\UriSigner` could be wired to the none deprecated instance of UriSigner.

The good one is if we extend the `Symfony\Component\HttpKernel\UriSigner` could still be given to service already switched to the new `Symfony\Component\HttpFoundation\UriSigner` (ESI, ... could then already be switched without bc break).

src/Symfony/Bundle/FrameworkBundle/Resources/config/services.php
```diff
@@ -47,7 +48,6 @@
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 use Symfony\Component\HttpKernel\KernelEvents;
 use Symfony\Component\HttpKernel\KernelInterface;
-use Symfony\Component\HttpKernel\UriSigner;
```
HypeMC, line 50: This removes the current service alias which is a BC break, it should be deprecated instead, see #46602

src/Symfony/Component/HttpFoundation/UriSigner.php
wouterj: As the class is moved without any logic change, you can use `class_alias`.

As an example, see the Messenger transport classes that were moved from [`Transport`](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Messenger/Transport/Doctrine/DoctrineTransport.php) to [Bridges](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/DoctrineTransport.php#L133-L135).

Also note that we should not trigger a deprecation when it's instantiated, but when it's loaded (this is implemented by the above example).

alexander-schranz: @wouterj that is a really great solution 👍

src/Symfony/Component/HttpKernel/UriSigner.php
fabpot: 6.4 :)

fabpot:
@alexander-schranz Do you have time to finish this PR?

src/Symfony/Bundle/FrameworkBundle/Resources/config/services.php
alexander-schranz: do we still require the alias when `Symfony\Component\HttpKernel\UriSigner` is an class alias to `Symfony\Component\HttpFoundation\UriSigner`?

alexander-schranz:
The fabbot seems to return a false positive currently:

Bildschirmfoto 2023-08-04 um 12 52 13

src/Symfony/Bundle/FrameworkBundle/Resources/config/services.php
alexander-schranz: I readded the alias as ReflectionParameter would return from as `->getType()->__toString()` still the old class and I'm not sure if that kind of reflection is used for autowiring.

alexander-schranz:
If I see it correctly all failing tests are related to current 6.4 branch and not the pull request. So I hope I tackled all parts.

src/Symfony/Bundle/FrameworkBundle/Resources/config/services.php
HypeMC: The alias should be deprecated though.

alexander-schranz: I dont see a way to deprecate here as it returns the parent again?

src/Symfony/Bundle/FrameworkBundle/Resources/config/services.php
```diff
@@ -157,6 +158,7 @@ class_exists(WorkflowEvents::class) ? WorkflowEvents::ALIASES : []
                 param('kernel.secret'),
             ])
         ->alias(UriSigner::class, 'uri_signer')
+        ->alias(HttpKernelUriSigner::class, 'uri_signer')
```
wouterj, line 161:
```suggestion
        ->alias(HttpKernelUriSigner::class, 'uri_signer')
            ->deprecate('symfony/framework-bundle', '6.4', 'The "%alias_id%" alias is deprecated, use "'.UriSigner::class.'" instead.')
```

src/Symfony/Bundle/FrameworkBundle/Resources/config/services.php
wouterj: Seems like you forgot to update FrameworkBundle's `composer.json`

src/Symfony/Component/HttpFoundation/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ CHANGELOG
 6.4
 ---
 
+ * Add `UriSigner`
```
wouterj:
```suggestion
 * Add `UriSigner` from the HttpKernel component
```

src/Symfony/Component/HttpFoundation/Tests/UriSignerTest.php
wouterj: I see this is now marked as legacy, but it still is moved. The file should exist in both HttpKernel and HttpFoundation, with the one in HttpKernel having `@group legacy`

wouterj (changes requested):

src/Symfony/Component/HttpFoundation/CHANGELOG.md
wouterj: We also need an entry to HttpKernel's CHANGELOG file (saying it was deprecated, use the one in HttpFoundation instead) and in the `UPGRADE-6.4.md` file in the root of the project (for both components)

src/Symfony/Bundle/FrameworkBundle/Resources/config/services.php
alexander-schranz: was already on 6.4: https://github.com/symfony/symfony/blob/614165670d40b1281a747723677ae2655cb6962c/src/Symfony/Bundle/FrameworkBundle/composer.json#L28

alexander-schranz:
@wouterj thx for the review, I hope I didn't miss something but it should be all tackled.

UPGRADE-6.4.md
```diff
@@ -89,8 +89,11 @@ HttpFoundation
 HttpKernel
 ----------
 
+ * Deprecate `UriSigner`, use `HttpFoundation` `UriSigner` instead
  * [BC break] `BundleInterface` no longer extends `ContainerAwareInterface`
  * [BC break] Add native return types to `TraceableEventDispatcher` and to `MergeExtensionConfigurationPass`
+ * `BundleInterface` no longer extends `ContainerAwareInterface`
+ * Add native return types to `TraceableEventDispatcher` and to `MergeExtensionConfigurationPass`
```
wouterj: Hmm, these 2 seems unrelated. I think maybe something went wrong resolving the merge conflict?

src/Symfony/Component/HttpKernel/Tests/UriSignerTest.php
```diff
@@ -9,12 +9,15 @@
  * file that was distributed with this source code.
  */
 
-namespace Symfony\Component\HttpKernel\Tests;
+namespace Symfony\Component\HttpFoundation\Tests;
```
wouterj:
```suggestion
namespace Symfony\Component\HttpKernel\Tests;
```

src/Symfony/Component/HttpKernel/Tests/UriSignerTest.php
```diff
@@ -9,12 +9,15 @@
...
 
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\HttpFoundation\Request;
-use Symfony\Component\HttpKernel\UriSigner;
+use Symfony\Component\HttpFoundation\UriSigner;
```
wouterj:
```suggestion
use Symfony\Component\HttpKernel\UriSigner;
```

wouterj (approved):
Thanks, just a few things were missed. But other than that, this looks good to go from my side!

fabpot:
@alexander-schranz Do you have time to finish this PR?

wouterj:
If not, please tell us and we (I) will force push the final tweaks to this PR so we can get it in 6.4/7.0 🙂

alexander-schranz:
@wouterj currently on vacation feel free to force push. Else I can finish it also in two weeks after my vacation.

wouterj:
Enjoy your holiday!

This PR should be ready now.

fabpot (approved):

fabpot:
Thank you @alexander-schranz.

PR #50393, opened 2023-05-23
[AssetMapper] Fixing incorrect exception & adding allowing more realistic error mode

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes-ish
| New feature?  | no-ish
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | Still WIP

Apologies for the late PR:

A) We were accidentally using an exception class from the `Asset` component. Fixed.

B) We are parsing .js files to find imports. That, by its very nature, won't be perfect, which is fine - if there is an import we don't recognize we can just leave it alone and (ideally) notify the user. Due to the imperfect matching, the `strict` mode is too strict to be used as a default - it'll cause bug reports. I've fixed this by adding a "warning" mode and using that. It's a balance between an exception (too much) vs not reporting to the user at all (too little).

Thanks!


src/Symfony/Component/AssetMapper/Compiler/CssAssetUrlCompiler.php
```diff
@@ -59,4 +72,19 @@ public function supports(MappedAsset $asset): bool
     {
         return 'css' === $asset->getPublicExtension();
     }
+
+    private function handleMissingImport(string $message, \Throwable $e = null): void
+    {
+        switch ($this->missingImportMode) {
```
chalasr: (nitpick) `match()` is preferred by now

chalasr (approved):

nicolas-grekas:
Thank you @weaverryan.

PR #50394, opened 2023-05-23
[AssetMapper] Avoid loading potentially ALL assets in dev server

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | Still TODO

Hi!

One other rough edge found when using on a real project. The dev server needs to quickly go from "public path" -> "logical path", so it can then look up the `MappedAsset`. Previously, for every served asset, it would loop over ALL assets until it found a match. We have a `CachedMappedAssetFactory`, which makes this happen quickly, but it still loads many assets into memory, including their contents. I was seeing a 13mb jump in the memory on those requests in a modest-sized app. If someone was serving a lot of images, it could get huge. No reason to do that work multiple times.

Btw, none of this happens on production - caching is here just for dev experience.

Thanks!

### DIFF
--- a/src/Symfony/Component/AssetMapper/AssetMapperDevServerSubscriber.php
+++ b/src/Symfony/Component/AssetMapper/AssetMapperDevServerSubscriber.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Component\AssetMapper;
 
+use Psr\Cache\CacheItemPoolInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Event\RequestEvent;
@@ -104,6 +105,7 @@ final class AssetMapperDevServerSubscriber implements EventSubscriberInterface
         private readonly AssetMapperInterface $assetMapper,
         string $publicPrefix = '/assets/',
         array $extensionsMap = [],
+        private readonly ?CacheItemPoolInterface $cacheMapCache = null,
     ) {
         $this->publicPrefix = rtrim($publicPrefix, '/').'/';
         $this->extensionsMap = array_merge(self::EXTENSIONS_MAP, $extensionsMap);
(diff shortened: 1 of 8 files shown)


chalasr (approved):
👍

src/Symfony/Component/AssetMapper/AssetMapperDevServerSubscriber.php
```diff
@@ -160,4 +156,32 @@ private function getMediaType(string $path): ?string
...
+        $cachedAsset = $this->cacheMapCache->getItem(hash('xxh128', $pathInfo));
+        $asset = $cachedAsset->isHit() ? $this->assetMapper->getAsset($cachedAsset->get()) : null;
+
+        if (null !== $asset && $asset->getPublicPath() === $pathInfo) {
+            return $asset;
+        }
+
+        // we did not find a match
+        $asset = null;
+        foreach ($this->assetMapper->allAssets() as $assetCandidate) {
```
nicolas-grekas, line 174: would there be a way to list only the public paths, and skip loading the full assets in memory?

nicolas-grekas: if not, can't we make allAssets return an iterator instead of an array, so that we don't have to load everything in memory?

weaverryan: > would there be a way to list only the public paths, and skip loading the full assets in memory?

MAYBE :p. To generate the public paths, something needs to needs to load all the `MappedAsset[]` (as public path depends on contents for the digest). `AssetMapper` *could* have some sort of `->allPublicPaths()` or even `findByPublicPath()` where we create that "public path map" and cache it (I "think" we would not need to worry about invalidation thanks to the digest: i.e. public path changes when file changes). It's basically what I'm doing here, which I realize is kind of an odd spot, but it's also only needed here.

> if not, can't we make allAssets return an iterator instead of an array, so that we don't have to load everything in memory?

Definitely! Though... if we have 200 assets and the one we want is number 190, it's still not ideal. It would be simpler, but each request would, on average, I guess still load 50% of the assets?

weaverryan:
Just made `AssetMapperInterface::allAssets()` iterable - a safer bet and easy win.

nicolas-grekas:
Tests fail :)

weaverryan:
> Tests fail :)

Only because my tests were so good 😛. It's better now - sorry about that. One failure is unrelated.

nicolas-grekas:
Thank you @weaverryan.

Issue #50377, opened 2023-05-20
[WebProfilerBundle] Select more than one HTTP method to filter or exclude a HTTP method


artyuum:
### Description

Currently, we can only select one HTTP method to filter:
![image](https://github.com/symfony/symfony/assets/17199757/f27cc19a-4aa9-4c03-893b-9a82d692c93c)

When working on an application that receives requests from the browser, it's common to have many preflight requests. When debugging, it would help to be able to filter out these requests by only selecting the needed HTTP methods or by excluding the  `OPTIONS` method for example.

### Example

As the first option, we can add the `multiple` attribute on the <select> input. It will do the job but it would break the design IMHO (cc @javiereguiluz), because the input would always display as expanded. I don't know if you would agree to add some kind of multi-select library but that can be a solution.

##

A second option is to add another input called "Exclude method".

MatTheCat:
Wouldn’t logs filters “buttons” be appropriate?

![](https://github.com/symfony/symfony/assets/1898254/a6b03563-e167-430b-bfc3-bb1f77ce14f4)

artyuum:
I like it.

javiereguiluz:
I like the idea of using a dropdown similar to the filters used in other panels. Thanks!

MatTheCat:
FTR I tried giving this a go but many CSS rules got in the way being too specific, and I didn’t want to end up rewriting so much :sweat_smile:

javiereguiluz:
If you want, continue working on your PR to add this feature as a form field that allows selecting multiple choices, and later I can submit another PR to try to update the UI.

MatTheCat:
Another issue was what to display when the dropdown is closed. When filtering logs the label is inside and the number of selected items is displayed next to it. In this case the label would be outside so I found it weird to just display the number of selected methods 🤔 

Would displaying as many selected items as possible be better? Eg if you selected GET and OPTIONS you could have

> GET, OPT… (2)

WDYT?

carsonbot:
Thank you for this suggestion.
There has not been a lot of activity here for a while. Would you still like to see this feature?

artyuum:
Yes

carsonbot:
Thank you for this suggestion.
There has not been a lot of activity here for a while. Would you still like to see this feature?

artyuum:
Yes.

carsonbot:
Thank you for this suggestion.
There has not been a lot of activity here for a while. Would you still like to see this feature?
[...]

PR #50395, opened 2023-05-23
[HttpKernel][WebProfilerBundle] Allow filtering profiles by many request methods

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Part of #50377 
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/search.html.twig
+++ b/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/search.html.twig
@@ -11,7 +11,7 @@
                 <select name="method" id="method">
                     <option value="">Any</option>
                     {% for m in ['DELETE', 'GET', 'HEAD', 'PATCH', 'POST', 'PUT'] %}
-                        <option {{ m == method ? 'selected="selected"' }}>{{ m }}</option>
+                        <option {{ m in method ? 'selected="selected"' }}>{{ m }}</option>
                     {% endfor %}
                 </select>
             </div>
--- a/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/search.html.twig
+++ b/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/search.html.twig
@@ -11,7 +11,7 @@
                 <select name="method" id="method">
                     <option value="">Any</option>
                     {% for m in ['DELETE', 'GET', 'HEAD', 'PATCH', 'POST', 'PUT'] %}
-                        <option {{ m in method ? 'selected="selected"' }}>{{ m }}</option>
+                        <option {{ m == method or (method is iterable and m in method) ? 'selected' }}>{{ m }}</option>
                     {% endfor %}
                 </select>
             </div>
--- a/src/Symfony/Component/HttpKernel/Profiler/FileProfilerStorage.php
+++ b/src/Symfony/Component/HttpKernel/Profiler/FileProfilerStorage.php
@@ -42,7 +42,7 @@ class FileProfilerStorage implements ProfilerStorageInterface
         }
     }
 
-    public function find(?string $ip, ?string $url, ?int $limit, ?string $method, int $start = null, int $end = null, string $statusCode = null): array
+    public function find(?string $ip, ?string $url, ?int $limit, string|array|null $method, int $start = null, int $end = null, string $statusCode = null): array
     {
         $file = $this->getIndexFilename();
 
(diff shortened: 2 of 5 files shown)


src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/search.html.twig
```diff
@@ -11,7 +11,7 @@
                 <select name="method" id="method">
                     <option value="">Any</option>
                     {% for m in ['DELETE', 'GET', 'HEAD', 'PATCH', 'POST', 'PUT'] %}
-                        <option {{ m == method ? 'selected="selected"' }}>{{ m }}</option>
+                        <option {{ m in method ? 'selected="selected"' }}>{{ m }}</option>
```
stof: will this work fine with a string or null ?

MatTheCat: Yes, `in` works with strings and returns `false` if the value is neither a string nor an iterable, cf. [`twig_in_filter`](https://github.com/twigphp/Twig/blob/833e51ea358e152d41ad46f30a7a4cbb38c44aed/src/Extension/CoreExtension.php#L916).

stof: yes, but it does not do an equality check. It uses `str_contains`, which is not what we want

MatTheCat: It works in this case and is simpler than

```twig
m == method or method is iterable and m in method
```

Should I change it this way?

src/Symfony/Component/HttpKernel/Profiler/FileProfilerStorage.php
```diff
@@ -42,7 +42,7 @@ public function __construct(string $dsn)
         }
     }
 
-    public function find(?string $ip, ?string $url, ?int $limit, ?string $method, int $start = null, int $end = null, string $statusCode = null): array
+    public function find(?string $ip, ?string $url, ?int $limit, string|array|null $method, int $start = null, int $end = null, string $statusCode = null): array
```
fabpot, line 45: Looks like a BC break as this class is not `@internal`.

src/Symfony/Component/HttpKernel/Profiler/Profiler.php
```diff
@@ -127,7 +127,7 @@ public function purge()
      *
      * @see https://php.net/datetime.formats for the supported date/time formats
      */
-    public function find(?string $ip, ?string $url, ?int $limit, ?string $method, ?string $start, ?string $end, string $statusCode = null): array
+    public function find(?string $ip, ?string $url, ?int $limit, string|array|null $method, ?string $start, ?string $end, string $statusCode = null): array
```
fabpot, line 130: BC break as well here?

src/Symfony/Component/HttpKernel/Profiler/ProfilerStorageInterface.php
```diff
@@ -33,7 +33,7 @@ interface ProfilerStorageInterface
      * @param int|null $start The start date to search from
      * @param int|null $end   The end date to search to
      */
-    public function find(?string $ip, ?string $url, ?int $limit, ?string $method, int $start = null, int $end = null): array;
+    public function find(?string $ip, ?string $url, ?int $limit, string|array|null $method, int $start = null, int $end = null): array;
```
fabpot, line 36: BC break

fabpot (changes requested):

MatTheCat:
Not sure what to do 🤔 

`ProfilerStorageInterface` is `@internal` since v4 and says

> the file storage is more than enough and no other implementations will ever be supported

Is there a way to keep BC here or should I target 7.0?

src/Symfony/Component/HttpKernel/Profiler/ProfilerStorageInterface.php
94noni: the interface is marked internal since +4years, and the profiler is generally a dev component
also https://symfony.com/doc/current/contributing/code/bc.html#using-our-interfaces

**of course we never know how devs use components in the wild**

but challenging a little bit here: perhaps interface and implementation class can be changed with an upgrade notice?
otherwise, defined a new method that takes a string with a specific separator that explode it, calls the old one if only one item or the new one as array

but this feature looks great and useful :)

94noni:
Friendly ping @MatTheCat and reviewers 
What is the state of PR also regarding the potential bc layer ?

MatTheCat:
I still have no clue how to add a BC layer, and since there are many conflicts I will close this PR.

PR #50396, opened 2023-05-23
[Validator] Allow single integer for the `versions` option of the `Uuid` constraint

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | _NA_
| License       | MIT
| Doc PR        | Todo

Improve DX by allowing to pass a single integer for the `versions` option of the `Uuid` constraint

### DIFF
--- a/src/Symfony/Component/Validator/Constraints/Uuid.php
+++ b/src/Symfony/Component/Validator/Constraints/Uuid.php
@@ -100,12 +100,12 @@ class Uuid extends Constraint
     public $normalizer;
 
     /**
-     * @param int[]|null $versions
+     * @param int[]|int|null $versions
      */
     public function __construct(
         array $options = null,
         string $message = null,
-        array $versions = null,
+        array|int $versions = null,
         bool $strict = null,
         callable $normalizer = null,
         array $groups = null,
@@ -114,7 +114,7 @@ class Uuid extends Constraint
         parent::__construct($options, $groups, $payload);
 
         $this->message = $message ?? $this->message;
-        $this->versions = $versions ?? $this->versions;
+        $this->versions = (array) ($versions ?? $this->versions);
         $this->strict = $strict ?? $this->strict;
         $this->normalizer = $normalizer ?? $this->normalizer;
 
--- a/src/Symfony/Component/Validator/CHANGELOG.md
+++ b/src/Symfony/Component/Validator/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+ * Allow single integer for the `versions` option of the `Uuid` constraint
+
 6.3
 ---
 
--- a/src/Symfony/Component/Validator/Tests/Constraints/UuidValidatorTest.php
+++ b/src/Symfony/Component/Validator/Tests/Constraints/UuidValidatorTest.php
@@ -296,4 +296,13 @@ class UuidValidatorTest extends ConstraintValidatorTestCase
         yield Uuid::V7_MONOTONIC => ['0184c292-b133-7e10-a3b4-d49c1ab49b2a', true];
         yield Uuid::V8_CUSTOM => ['00112233-4455-8677-8899-aabbccddeeff', false];
     }
+
+    public function testAcceptsSingleIntegerAsVersion()
+    {
+        $constraint = new Uuid(versions: 7);
+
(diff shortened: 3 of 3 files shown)


UPGRADE-7.0.md
```diff
@@ -6,3 +6,8 @@ release process, both versions will have the same features, but Symfony 7.0 won'
 To upgrade, make sure to resolve all deprecation notices.
 
 This file will be updated on the branch 7.0 for each deprecated feature that is removed.
+
+Validator
+---------
+
+ * Changed `$versions` parameter type of the `Uuid::__construct()` method from `array` to `array|int`.
```
xabbuh: I don't think we need an entry in the upgrade file.

xabbuh (approved):

xabbuh:
@alexandre-daubois Can you rebase on the `6.4` branch?

UPGRADE-7.0.md
alexandre-daubois: Are you sure? Because this is actually a BC break for someone who extends this constraint, that's why I targeted 6.4 at first and changed to 7.0 afterward :thinking:

nicolas-grekas: This cannot not target 7.0. We never merge unannounced BC breaks.
But what's the BC break like?

src/Symfony/Component/Validator/Constraints/UuidValidator.php
```diff
@@ -237,7 +237,7 @@ private function validateStrict(string $value, Uuid $constraint): void
         }
 
         // Check version
-        if (!\in_array($value[self::STRICT_VERSION_POSITION], $constraint->versions)) {
+        if (!\in_array($value[self::STRICT_VERSION_POSITION], (array) $constraint->versions)) {
```
nicolas-grekas: this should be done inside the constructor of `Uuid`

src/Symfony/Component/Validator/Constraints/Uuid.php
```diff
@@ -89,11 +89,11 @@ class Uuid extends Constraint
     public $strict = true;
 
     /**
-     * Array of allowed versions (see version constants above).
+     * Specific version or array of allowed versions (see version constants above).
      *
      * All UUID versions are allowed by default
      *
-     * @var int[]
+     * @var int[]|int
```
nicolas-grekas: should be reverted, there's no need to change the property

src/Symfony/Component/Validator/Constraints/Uuid.php
```diff
@@ -105,7 +105,7 @@ class Uuid extends Constraint
     public function __construct(
         array $options = null,
         string $message = null,
-        array $versions = null,
+        array|int $versions = null,
```
nicolas-grekas, line 108: the `@param` needs to be updated to mention the change

UPGRADE-7.0.md
alexandre-daubois: I thought about it again, you are right. I updated the PR accordingly :+1:

fabpot:
I'm not sure I like this change. The argument is named `versions`, so I would not expect a single integer to be supported.

nicolas-grekas:
> The argument is named versions, so I would not expect a single integer to be supported.

There is a least one precedent on the topic, Route's schemes + methods accept a single value:
https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Routing/Annotation/Route.php#L44-L45

alexandre-daubois:
I actually found some other examples. Many other constraints accept `array|string`:

- CardScheme with `$schemes` https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Validator/Constraints/CardScheme.php#L57
- Cascade with `$exclude` https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Validator/Constraints/Cascade.php#L28
- File with `$mimeTypes` and `$extensions` https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Validator/Constraints/File.php#L55

This PR (and [this one](https://github.com/symfony/symfony/pull/50425)) could bring some consistency here in my opnion :slightly_smiling_face:

nicolas-grekas (approved):

yceruto (approved):

xabbuh (approved):

nicolas-grekas:
Thank you @alexandre-daubois.

PR #50397, opened 2023-05-23
[HttpKernel][VarDumper] Fix dumping with labels

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Should fix https://github.com/symfony/symfony/pull/50347#issuecomment-1558458542 /cc @bobthecow 

Also improves the display in the WDT:

image

image

### DIFF
--- a/src/Symfony/Bundle/DebugBundle/Resources/views/Profiler/dump.html.twig
+++ b/src/Symfony/Bundle/DebugBundle/Resources/views/Profiler/dump.html.twig
@@ -11,6 +11,9 @@
             {% for dump in collector.getDumps('html') %}
                 <div class="sf-toolbar-info-piece">
                     <span>
+                    {% if dump.label is defined and '' != dump.label %}
+                        <span class="sf-toolbar-file-line"><strong>{{ dump.label }}</strong> in </span>
+                    {% endif %}
                     {% if dump.file %}
                         {% set link = dump.file|file_link(dump.line) %}
                         {% if link %}
@@ -45,7 +48,12 @@
 
     {% for dump in collector.getDumps('html') %}
         <div class="sf-dump sf-reset">
-            <span class="metadata">In
+            <span class="metadata">
+                {% if dump.label is defined and '' != dump.label %}
+                    <strong>{{ dump.label }}</strong> in
+                {% else %}
+                    In
+                {% endif %}
                 {% if dump.line %}
                     {% set link = dump.file|file_link(dump.line) %}
                     {% if link %}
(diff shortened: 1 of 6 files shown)


src/Symfony/Bundle/DebugBundle/Resources/views/Profiler/dump.html.twig
```diff
@@ -45,7 +48,12 @@
 
     {% for dump in collector.getDumps('html') %}
         <div class="sf-dump sf-reset">
-            <span class="metadata">In
+            <span class="metadata">
+                {% if dump.label is defined and '' != dump.label %}
+                    <strong>{{ dump.label }}</strong> in
+                {% else %}
+                    In
```
fancyweb, line 55: When there's no label we display `In FooController.php line 14` now?

src/Symfony/Component/HttpKernel/DataCollector/DumpDataCollector.php
```diff
@@ -125,7 +128,7 @@ public function collect(Request $request, Response $response, \Throwable $except
             }
 
             foreach ($this->data as $dump) {
-                $this->doDump($dumper, $dump['data'], $dump['name'], $dump['file'], $dump['line']);
+                $this->doDump($dumper, $dump['data'], $dump['name'], $dump['file'], $dump['line'], $dump['label'] ?? '');
```
fancyweb, line 131: Doesn't the label key always exist (same below)?

nicolas-grekas: I thought about existing profiles, which we might load from a previous version during migrations.

src/Symfony/Bundle/DebugBundle/Resources/views/Profiler/dump.html.twig
nicolas-grekas: as was before, no? (see L48 before)

stof (approved):

fancyweb (approved):

bobthecow:
Thanks for jumping on this, @nicolas-grekas! 

This fix works, in that it no longer breaks functionality. But it does look like it introduces a change in formatted output that puts empty spans in front of a lot of things:

Screenshot 2023-05-23 at 10 19 29 AM

Either that or this was introduced somewhere else but masked by the BC break in my tests.

nicolas-grekas:
@bobthecow fixed!

bobthecow:
@nicolas-grekas perfect, thanks so much!

PR #48750, opened 2022-12-21
[Security] Add generics to PasswordUpgraderInterface

VincentLanglet:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes (but phpdoc only)
| Deprecations? | no
| Tickets       | Fix #... 
| License       | MIT
| Doc PR        | symfony/symfony-docs#... 

Related to the discussion https://github.com/symfony/symfony/issues/48743

A PasswordUpgraderInterface might not support every existing user implementing `PasswordAuthenticatedUserInterface` because of specific needs like a `setId` method. I would even say that most of the time it's only supporting one class of user and the method `upgradePassword` might be implemented in the doctrine `UserRepository`.

This would allow to write
```
/**
 * @extends EntityRepository<User>
 * @implements PasswordUpgraderInterface<User>
 */
class UserRepository extends EntityRepository implements UserLoaderInterface, PasswordUpgraderInterface
{
     public function upgradePassword(PasswordAuthenticatedUserInterface $user, string $newHashedPassword): void
     {
         // static analysis tool will know $user is supposed to be a User.
     }
}
```
and also having a static analysis error when using
```
$userRepository->($anotherUserClass, $password);
```

PR #50399, opened 2023-05-23
[Security] Add generic to UserProviderInterface

VincentLanglet:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes (but phpdoc only)
| Deprecations? | no
| Tickets       | Fix #... 
| License       | MIT

Similar to https://github.com/symfony/symfony/pull/48750 @nicolas-grekas

### DIFF
--- a/src/Symfony/Bridge/Doctrine/Security/User/EntityUserProvider.php
+++ b/src/Symfony/Bridge/Doctrine/Security/User/EntityUserProvider.php
@@ -29,6 +29,10 @@ use Symfony\Component\Security\Core\User\UserProviderInterface;
  *
  * @author Fabien Potencier <fabien@symfony.com>
  * @author Johannes M. Schmitt <schmittjoh@gmail.com>
+ *
+ * @template TUser of UserInterface
+ *
+ * @template-implements UserProviderInterface<TUser>
  */
 class EntityUserProvider implements UserProviderInterface, PasswordUpgraderInterface
 {
--- a/src/Symfony/Component/Ldap/Security/LdapUserProvider.php
+++ b/src/Symfony/Component/Ldap/Security/LdapUserProvider.php
@@ -30,6 +30,8 @@ use Symfony\Component\Security\Core\User\UserProviderInterface;
  * @author Grégoire Pineau <lyrixx@lyrixx.info>
  * @author Charles Sarrazin <charles@sarraz.in>
  * @author Robin Chalas <robin.chalas@gmail.com>
+ *
+ * @template-implements UserProviderInterface<LdapUser>
  */
 class LdapUserProvider implements UserProviderInterface, PasswordUpgraderInterface
 {
--- a/src/Symfony/Component/Security/Core/User/AttributesBasedUserProviderInterface.php
+++ b/src/Symfony/Component/Security/Core/User/AttributesBasedUserProviderInterface.php
@@ -18,6 +18,10 @@ use Symfony\Component\Security\Core\Exception\UserNotFoundException;
  * This is particularly useful with self-contained access tokens.
  *
  * @experimental
+ *
+ * @template-covariant TUser of UserInterface
+ *
+ * @template-extends UserProviderInterface<TUser>
  */
 interface AttributesBasedUserProviderInterface extends UserProviderInterface
 {
@@ -25,6 +29,8 @@ interface AttributesBasedUserProviderInterface extends UserProviderInterface
      * Loads the user for the given user identifier (e.g. username or email) and attributes.
      *
      * This method must throw UserNotFoundException if the user is not found.
+     *
+     * @return TUser
      *
      * @throws UserNotFoundException
      */
--- a/src/Symfony/Component/Security/Core/User/ChainUserProvider.php
+++ b/src/Symfony/Component/Security/Core/User/ChainUserProvider.php
@@ -21,6 +21,8 @@ use Symfony\Component\Security\Core\Exception\UserNotFoundException;
  * handle the request.
  *
  * @author Johannes M. Schmitt <schmittjoh@gmail.com>
+ *
+ * @template-implements UserProviderInterface<UserInterface>
  */
 class ChainUserProvider implements UserProviderInterface, PasswordUpgraderInterface
 {
--- a/src/Symfony/Component/Security/Core/User/InMemoryUserProvider.php
+++ b/src/Symfony/Component/Security/Core/User/InMemoryUserProvider.php
@@ -21,6 +21,8 @@ use Symfony\Component\Security\Core\Exception\UserNotFoundException;
  * (a backend with a unique admin for instance)
  *
  * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @template-implements UserProviderInterface<InMemoryUser>
  */
 class InMemoryUserProvider implements UserProviderInterface
 {
--- a/src/Symfony/Component/Security/Core/User/MissingUserProvider.php
+++ b/src/Symfony/Component/Security/Core/User/MissingUserProvider.php
@@ -18,6 +18,8 @@ use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
  * when a firewall requires a user provider but none was defined.
  *
  * @internal
+ *
+ * @template-implements UserProviderInterface<never>
  */
 class MissingUserProvider implements UserProviderInterface
 {
--- a/src/Symfony/Component/Security/Core/User/UserProviderInterface.php
+++ b/src/Symfony/Component/Security/Core/User/UserProviderInterface.php
@@ -28,6 +28,8 @@ use Symfony\Component\Security\Core\Exception\UserNotFoundException;
  * information is submitted or what the UserInterface object looks like.
  *
  * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @template-covariant TUser of UserInterface
  */
 interface UserProviderInterface
 {
@@ -40,6 +42,8 @@ interface UserProviderInterface
      * map.
      *
      * @return UserInterface
+     *
+     * @psalm-return TUser
      *
      * @throws UnsupportedUserException if the user is not supported
      * @throws UserNotFoundException    if the user is not found
@@ -57,6 +61,8 @@ interface UserProviderInterface
      * Loads the user for the given user identifier (e.g. username or email).
      *
      * This method must throw UserNotFoundException if the user is not found.
+     *
+     * @return TUser
      *
      * @throws UserNotFoundException
      */


src/Symfony/Component/Security/Core/User/AttributesBasedUserProviderInterface.php
```diff
@@ -18,6 +18,9 @@
  * This is particularly useful with self-contained access tokens.
  *
  * @experimental
+ *
+ * @template TUser of UserInterface
```
stof: this should be `@template-covariant` to be more flexible as the template type is only used in covariant positions

src/Symfony/Component/Security/Core/User/UserProviderInterface.php
```diff
@@ -49,6 +52,8 @@ public function refreshUser(UserInterface $user);
     /**
      * Whether this provider supports the given user class.
      *
+     * @param class-string<TUser> $class
```
stof: this is wrong. `supportsClass` will be called with other kind of class strings as well (if should return false for classes that are not TUser, but the way you wrote this would mean that it is forbidden to pass other class strings)

src/Symfony/Component/Security/Core/User/UserProviderInterface.php
```diff
@@ -28,6 +28,8 @@
  * information is submitted or what the UserInterface object looks like.
  *
  * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @template TUser of UserInterface
```
stof: this should be `@template-covariant` to be more flexible as the template type is only used in covariant positions (once the usage in `supportsClass` is removed)

stof (changes requested):
You should add the proper `@template-implements` on implementations.

stof:
This will be for 6.4, not for 6.3 (we are 2 months after the feature freeze)

src/Symfony/Component/Security/Core/User/UserProviderInterface.php
VincentLanglet: Indeed

VincentLanglet:
```
@psalm-assert-if-true class-string<TUser> $class
```
seems better then

src/Symfony/Component/Security/Core/User/UserProviderInterface.php
VincentLanglet: What about adding
```
@param TUser $user
```
to the refreshUser method ?

Because we're supposed to call
```
$provider->supportsClass($userClass)
```
before `refreshUser`.

src/Symfony/Component/Security/Core/User/UserProviderInterface.php
stof: I don't think this assertion would provide much value.

src/Symfony/Component/Security/Core/User/UserProviderInterface.php
stof: This would forbid making the generic type covariant, which would make it harder to properly write types for code using UserProviderInterface instances.

VincentLanglet: I'm not sure about the big win of the covariant template:

Every UserProvider like ChainUserProvider or EntityUserProvider has
```
 * @template TUser of PasswordAuthenticatedUserInterface&UserInterface
 *
 * @template-implements UserProviderInterface<TUser>
 * @template-implements PasswordUpgraderInterface<TUser>
```
cannot be covariant because PasswordUpgraderInterface has a not covariant template.

src/Symfony/Component/Security/Core/User/AttributesBasedUserProviderInterface.php
```diff
@@ -18,6 +18,10 @@
...
+ *
+ * @template-implements UserProviderInterface<TUser>
```
stof: this should be `@template-extends`, not `@template-implements` (the rule is simple: if the PHP code uses the `extends` keyword, the phpdoc tag is also the `extends` one)

src/Symfony/Component/Ldap/Security/LdapUserProvider.php
```diff
@@ -30,6 +30,11 @@
  * @author Grégoire Pineau <lyrixx@lyrixx.info>
  * @author Charles Sarrazin <charles@sarraz.in>
  * @author Robin Chalas <robin.chalas@gmail.com>
+ *
+ * @template TUser of PasswordAuthenticatedUserInterface&UserInterface
```
stof: This is wrong. the type for this provider is `LdapUser`

src/Symfony/Component/Security/Core/User/ChainUserProvider.php
```diff
@@ -21,6 +21,11 @@
  * handle the request.
  *
  * @author Johannes M. Schmitt <schmittjoh@gmail.com>
+ *
+ * @template TUser of PasswordAuthenticatedUserInterface&UserInterface
+ *
+ * @template-implements UserProviderInterface<TUser>
```
stof: This should use `@template-implements UserProviderInterface<UserInterface>` instead, as the different chained providers won't have the same type

src/Symfony/Component/Security/Core/User/MissingUserProvider.php
```diff
@@ -18,6 +18,10 @@
  * when a firewall requires a user provider but none was defined.
  *
  * @internal
+ *
+ * @template TUser of UserInterface
+ *
+ * @template-implements UserProviderInterface<TUser>
```
stof: This class should not be generic, and use `UserProviderInterface<UserInterface>`

src/Symfony/Component/Security/Core/User/UserProviderInterface.php
stof: having `UserProviderInterface` being covariant is a big win (it would be the only way to actually fully type ChainUserProvider with static analysis tools at their strict levels).

The template type of EntityUserProvider can be a non-generic one.

src/Symfony/Component/Security/Core/User/UserProviderInterface.php
```diff
@@ -39,8 +41,12 @@ interface UserProviderInterface
      * object can just be merged into some internal array of users / identity
      * map.
      *
+     * @param TUser $user
```
stof: technically, passing a different user is valid in the interface, as it defines that the implementation should throw a UnsupportedUserException in such case.

src/Symfony/Component/Security/Core/User/InMemoryUserProvider.php
```diff
@@ -21,6 +21,10 @@
  * (a backend with a unique admin for instance)
  *
  * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @template TUser of UserInterface
```
stof: This should not be a generic class. This class implements a user provider for `InMemoryUser`

stof (changes requested):

src/Symfony/Component/Ldap/Security/LdapUserProvider.php
VincentLanglet: Let's not make this generic then.

src/Symfony/Component/Security/Core/User/ChainUserProvider.php
VincentLanglet: Ok, let's not make this generic then.

src/Symfony/Component/Security/Core/User/MissingUserProvider.php
VincentLanglet: Done

src/Symfony/Component/Security/Core/User/UserProviderInterface.php
VincentLanglet: I made it covariant then and removed some generics from specific UserProvider.

VincentLanglet:
Hi @stof, is it ok for you now ?

src/Symfony/Bridge/Doctrine/Security/User/EntityUserProvider.php
```diff
@@ -29,6 +29,9 @@
  *
  * @author Fabien Potencier <fabien@symfony.com>
  * @author Johannes M. Schmitt <schmittjoh@gmail.com>
+ *
+ * @template-implements UserProviderInterface<UserInterface>
+ * @template-implements PasswordUpgraderInterface<PasswordAuthenticatedUserInterface>
  */
 class EntityUserProvider implements UserProviderInterface, PasswordUpgraderInterface
```
derrabus, line 37: This class can be generic too. `EntityUserProvider` is configured for an entity class and will only load instances of that class.

src/Symfony/Component/Security/Core/User/MissingUserProvider.php
```diff
@@ -18,6 +18,8 @@
  * when a firewall requires a user provider but none was defined.
  *
  * @internal
+ *
+ * @template-implements UserProviderInterface<UserInterface>
```
derrabus: Does Psalm allow us to do this?

```suggestion
 * @template-implements UserProviderInterface<never>
```

VincentLanglet: Seems like it does ; I made the change

src/Symfony/Bridge/Doctrine/Security/User/EntityUserProvider.php
VincentLanglet: Is 
```
 * @template TUser of UserInterface&PasswordAuthenticatedUserInterface
 *
 * @template-implements UserProviderInterface<TUser>
 * @template-implements PasswordUpgraderInterface<TUser>
 ```
 better ?

Comment (derrabus) [reply src/Symfony/Bridge/Doctrine/Security/User/EntityUserProvider.php:37]: Yes. 🙂

**Review by nicolas-grekas — APPROVED** [review]

File: src/Symfony/Bridge/Doctrine/Security/User/EntityUserProvider.php [hunk 12]
```diff
@@ -29,6 +29,11 @@
  *
  * @author Fabien Potencier <fabien@symfony.com>
  * @author Johannes M. Schmitt <schmittjoh@gmail.com>
+ *
+ * @template TUser of UserInterface&PasswordAuthenticatedUserInterface
```
wouterj:
```suggestion
 * @template TUser of UserInterface
```

User entities don't have to be password authenticated as far as I know.

VincentLanglet: The class implements PassworUpgraderInterface, with the method:
```
public function upgradePassword(PasswordAuthenticatedUserInterface $user, string $newHashedPassword): void
    {
        $class = $this->getClass();
        if (!$user instanceof $class) {
            throw new UnsupportedUserException(sprintf('Instances of "%s" are not supported.', get_debug_type($user)));
        }

        $repository = $this->getRepository();
        if ($repository instanceof PasswordUpgraderInterface) {
            $repository->upgradePassword($user, $newHashedPassword);
        }
    }
```

So technically it should be password authenticated.

Another way to see it, it has the annotation
```
@template-implements PasswordUpgraderInterface<TUser>
```
and `PasswordUpgraderInterface` has the annotation `@template TUser of PasswordAuthenticatedUserInterface`

wouterj: Hmm, this let's me question whether we should actually want to make any of these interfaces generic.

Implementing `PasswordUpgraderInterface` does not require the user to be password authenticated. As this method implementation demonstrates, we can conditionally upgrade the password (in this case, by decorating the repository).

So defining that it must be password authenticated is a false assumption. That makes me wonder what advantages such generics gives us.

stof: I don't think PasswordUpgraderInterface should be generic, given that the generic type is only used to refine the argument type and this is actually wrong (an upgrader is expected to check supported user classes and throw UnsupportedUserException as there is no `supports` method).

VincentLanglet: > I don't think PasswordUpgraderInterface should be generic, given that the generic type is only used to refine the argument type and this is actually wrong (an upgrader is expected to check supported user classes and throw UnsupportedUserException as there is no `supports` method).

That's exactly the purpose of a generic.

If I have a `FooUgrader` which `implements PasswordUpgraderInterface` and only supports `FooUser`
This won't prevent to check for the argument type and throw an UnsupportedUserException, but this will allow to PHPStan to throw an error If I do:
```
$fooUpgrader->upgradePassord($notFooUser);
```
and statically prevent a runtime exception.

> Hmm, this let's me question whether we should actually want to make any of these interfaces generic. [...] That makes me wonder what advantages such generics gives us.

In the same way, if I have multiples `UserProviderInterface<Foo>`, I can assume that the return type of `loadUserByIdentifier` will be `Foo`. Like values of `Foo[]` are `Foo`...

> As this method implementation demonstrates, we can conditionally upgrade the password (in this case, by decorating the repository).

The repository does work with decoration:
```
if ($repository instanceof PasswordUpgraderInterface) {
        $repository->upgradePassword($user, $newHashedPassword);
}
```
But the param still require a `PasswordAuthenticatedUserInterface $user`

But I understand the conditionally point because you rely on the fact the user do
```
if ($user instance PasswordAuthenticatedUserInterface) {
    $provider->upgradePassword(...);
}
```
Not sure it's classic for PHPStan.

Unless ondrejmirtes has a perfect solution, what can I do is this:
https://github.com/symfony/symfony/pull/50399/commits/b0cdf12e2196d04b3f3f85a0ba57ed0c10315425

stof: @VincentLanglet but Symfony will definitely do `$fooUpgrader->upgradePassord($notFooUser);` as Symfony has no way to know whether your upgrader supports that user (see ChainUserProvider for instance).

So this generic type is a lie. your code **must** handle any PasswordAuthenticatedUserInterface as input, throwing an UnsupportedUserException for the ones it does not support. It cannot rely on the caller passing only supported users.

VincentLanglet: > @VincentLanglet but Symfony will definitely do `$fooUpgrader->upgradePassord($notFooUser);` as Symfony has no way to know whether your upgrader supports that user (see ChainUserProvider for instance).
> 
> So this generic type is a lie. your code **must** handle any PasswordAuthenticatedUserInterface as input, throwing an UnsupportedUserException for the ones it does not support. It cannot rely on the caller passing only supported users.

A generic never said you **mustn't** handle any PasswordAuthenticatedUserInterface and throw an exception.
That the same way that prior to php 7, you could write
```
/**
 * @param int $int
 */
function foo($int) {
    if (!is_int($int)) {
        throw new InvalidArgument();
    }
    
    // do things
}
```
The phpdoc was here to **help** the user.

This is an example without generic https://phpstan.org/r/4a125adf-b166-4cdb-ac5f-084d4deb27f4
and with https://phpstan.org/r/03eaf45f-f76b-408e-83aa-a9a0212c77e0
I can still add an instanceof check inside the method, but only the example with the generic can statically report the error (instead of waiting for runtime error).

But anyways, I think we can disagree here ; that shouldn't be the point here. 
This PR doesn't add the template to PasswordAuthenticatedUserInterface, and I changed, as requested to.
```
/**
 * @template TUser of UserInterface
 *
 * @template-implements UserProviderInterface<TUser>
 */
class EntityUserProvider implements UserProviderInterface, PasswordUpgraderInterface
```

stof: But there is **no** error calling it with another user. That's an expected usage of the interface.

On the other hand, without the generic, phpstan will detect broken implementations of the upgrader that forget to do the check, because it will know that the check is necessary.

VincentLanglet: Can it be discuss somewhere else ? My PR is about the UserProviderInterface, which does have a supportClass method. There is no usage of `PasswordUpgraderInterface` in this PR. So your request changes are done @stof

src/Symfony/Component/Security/Core/User/UserProviderInterface.php
```diff
@@ -41,6 +43,8 @@ interface UserProviderInterface
      *
      * @return UserInterface
      *
+     * @psalm-return TUser
```
nicolas-grekas, line 46: we can use `@return` for this, we do already in some places IIRC

VincentLanglet: It's indeed used in some place, because the @return was not necessary (the native type was added).
For this method, 
```
public function refreshUser(UserInterface $user);
```
there is no native return type.

`@return TUser` could loose some IDE or developer which doesn't know generics.
that's why I thought about adding the both annotations.

Do you prefer I change the `@return UserInterface` to `@return TUser` ?

nicolas-grekas: > Do you prefer I change the @return UserInterface to @return TUser ?
 
I do yes

VincentLanglet: Done in https://github.com/symfony/symfony/pull/50399/commits/10cd4cc307447162c70776ccd24cec7cebd23708

stof: When we don't have the native return type yet, we need to keep `@return` with the future return type. Otherwise we break things for the DebugClassLoader with its deprecation warning about the future return type and its type patching.

VincentLanglet: Reverted then

nicolas-grekas:
Thank you @VincentLanglet.

PR #50400, opened 2023-05-23
[AssetMapper] Sometimes asset contents are built from non-asset files

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes-ish
| New feature?  | yes-ish
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | Still TODO

Apologies again for the late thing - I'm *really* trying to push the component to find critical bugs or missing items.

In StimulusBundle, we dynamically build the contents of a `controllers.js` mapped asset from the `assets/controllers.json` file. This PR allows us to "bust" that asset's cache when `assets/controllers.json` is modified. Right now, it would be a bug in StimulusBundle that we can't really work around.

Other use-cases: someone decides to write an asset compiler that runs Tailwind automatically on a CSS file, and they want that dynamic asset to "vary" on the `tailwind.config.js` file.

Thanks!

### DIFF
--- a/src/Symfony/Component/AssetMapper/Factory/CachedMappedAssetFactory.php
+++ b/src/Symfony/Component/AssetMapper/Factory/CachedMappedAssetFactory.php
@@ -59,7 +59,8 @@ class CachedMappedAssetFactory implements MappedAssetFactoryInterface
      */
     private function collectResourcesFromAsset(MappedAsset $mappedAsset): array
     {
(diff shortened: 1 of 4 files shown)


stof (approved):

nicolas-grekas:
Thank you @weaverryan.

PR #50408, opened 2023-05-24
[AssetMapper] Change default importmap "provider" to JsDelivr+esm

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | yes
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        |Still TODO

We currently use the https://jspm.org/ API in `importmap:require` to find a CDN URL for an npm package - just like Rails. Unfortunately, this is NOT as robust as we had thought. For me, it's broken. 3 big issues:

A) **Not Combined**
Some packages are not packed/combined. Example: [chart.js/auto](https://ga.jspm.io/npm:chart.js@4.3.0/auto/auto.js) imports other packages and results in 3 requests instead of 1. Not TERRIBLE... so here IS a terrible example: [@popperjs/core](https://ga.jspm.io/npm:@popperjs/core@2.11.7/lib/index.js) (needed by `bootstrap`) results in nearly 50 requests ❗

B) **Downloading Broken**
For some packages, downloading simply doesn't work - https://github.com/rails/importmap-rails/issues/65. `@popperjs/core` is another good example. Many of its imports have the form `import"./utils/getOppositeVariationPlacement.js`. If we download the main file, it looks locally for that `utils/` file, which won't be there. [@chart.js/auto](https://ga.jspm.io/npm:chart.js@4.3.0/auto/auto.js) is another example.

C) **process.env.NODE_ENV included** 
Some packages (yes, again `@popperjs/core` is a great example!) contain `process.env.NODE_ENV` inside - https://github.com/rails/importmap-rails/issues/65#issuecomment-999569497

I believe that some package advertise an "esm" package... but just don't do a good job of creating it... or create it without the browser context in mind (or at least in a way that's inconvenient for downloading).

### jsDelivr to the rescue

THANKFULLY, jsDelivr seems to have a fantastic API/hosting that is *almost* exactly what we want: https://www.jsdelivr.com/?docs=esm

They deliver fully "packaged" modules, where the only import is for peer dependencies - e.g. https://cdn.jsdelivr.net/npm/bootstrap@5.2.3/+esm

There IS still an issue when downloading. "Peer imports" are relative -e.g. `import*as t from"/npm/@popperjs/core@2.11.7/+esm";` However, these imports follow a VERY strict pattern. So, when `--download` is passed, we parse these, download the peer dependency and update the import contents to `@popperjs/core`, which works with the importmap. It's not ideal that we need to do that, but it's straightforward and works great.

Sorry again for this late PR - I had assumed that jspm was robust because Rails is using it. It turns out it's robust... unless you hit a "bad" package, then it's terrible. And they're not that rare: on ux.symfony.com, I have hit several.

Thanks!


weaverryan:
The "high-deps" failure is because it's trying to pull in the latest commit from `symfony/framework-bundle`, but that is missing the commit from this PR which handles the wiring. But I'm not sure how to fix that.

stof:
The failing high-deps is because your Pr does not target the highest version (as the 6.4 branch has been created). It is fine to ignore that failure as it will be resolved once your merged PR is merged up from 6.3 to 6.4.

src/Symfony/Component/AssetMapper/ImportMap/Providers/JsDelivrEsmImportMapProvider.php
```diff
@@ -0,0 +1,150 @@
...
+            $version = $this->findVersion($packageName, $requireOptions->versionConstraint);
+
+            $url = sprintf('https://esm.run/%s@%s', $packageName, $version);
+            if ($filePath) {
+                $url .= '/'.$filePath;
+            }
+
+            $response = $this->httpClient->request('GET', $url);
+            if (200 !== $response->getStatusCode()) {
+                throw new \RuntimeException(sprintf('Error requiring JavaScript package from jsDelivr using URL "%s". Check your package name. Response: ', $url, $response->getContent(false)));
```
stof: instead of only putting the response content in the error message, we should also provide the status code (knowing that it failed with a 404 or a 503 or something else can help understand the issue)

src/Symfony/Component/AssetMapper/Tests/ImportMap/ImportMapManagerTest.php
```diff
@@ -20,17 +21,17 @@
...
 use Symfony\Component\Filesystem\Filesystem;
-use Symfony\Component\HttpClient\MockHttpClient;
-use Symfony\Component\HttpClient\Response\MockResponse;
 
 class ImportMapManagerTest extends TestCase
 {
-    private MockHttpClient $httpClient;
     private Filesystem $filesystem;
     private AssetMapperInterface $assetMapper;
+    private ImportMapPackageProviderInterface|MockObject $packageProvider;
```
stof: should be an intersection type, not a union type.

stof (changes requested):

src/Symfony/Component/AssetMapper/ImportMap/Providers/JsDelivrEsmImportMapProvider.php
```diff
@@ -0,0 +1,150 @@
...
+ */
+
+namespace Symfony\Component\AssetMapper\ImportMap\Providers;
+
+use Symfony\Component\AssetMapper\ImportMap\PackageRequireOptions;
+use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+/**
+ * @experimental
```
dunglas:
```suggestion
 * @experimental
 * @final
```

src/Symfony/Component/AssetMapper/ImportMap/Providers/JsDelivrEsmImportMapProvider.php
dunglas: For the best debuggability, we should even throw the exception created by HttpClient by using a logic similar to https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php#L267C9-L276

src/Symfony/Component/AssetMapper/ImportMap/Providers/JspmImportMapProvider.php
```diff
@@ -0,0 +1,95 @@
...
+
+namespace Symfony\Component\AssetMapper\ImportMap\Providers;
+
+use Symfony\Component\AssetMapper\ImportMap\ImportMapManager;
+use Symfony\Component\AssetMapper\ImportMap\PackageRequireOptions;
+use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+/**
+ * @experimental
```
dunglas:
```suggestion
 * @experimental
 * @final
```

src/Symfony/Component/AssetMapper/ImportMap/Providers/ResolvedImportMapPackage.php
```diff
@@ -0,0 +1,27 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\AssetMapper\ImportMap\Providers;
+
+use Symfony\Component\AssetMapper\ImportMap\PackageRequireOptions;
+
+/**
+ * @experimental
```
dunglas:
```suggestion
 * @experimental
 * @final
```

dunglas (changes requested):
Awesome!!

src/Symfony/Component/AssetMapper/ImportMap/Providers/JsDelivrEsmImportMapProvider.php
stof: We could even use an actual `final` keyword (any mocking can be done on the interface anyway)

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -1314,14 +1315,21 @@ private function registerAssetMapperConfiguration(array $config, ContainerBuilde
             ->getDefinition('asset_mapper.importmap.manager')
             ->replaceArgument(2, $config['importmap_path'])
             ->replaceArgument(3, $config['vendor_dir'])
-            ->replaceArgument(4, $config['provider'])
+        ;
+
+        $container
+            ->getDefinition('asset_mapper.importmap.resolver')
+            ->replaceArgument(0, $config['provider'])
```
nicolas-grekas, line 1322: not processing the value at compile-time is important to make this configurable by env var

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -1314,14 +1315,21 @@ private function registerAssetMapperConfiguration(array $config, ContainerBuilde
...
         ;
 
         $container
             ->getDefinition('asset_mapper.importmap.renderer')
             ->replaceArgument(2, $config['importmap_polyfill'] ?? ImportMapManager::POLYFILL_URL)
             ->replaceArgument(3, $config['importmap_script_attributes'])
         ;
+
+        $container->registerForAutoconfiguration(PackageResolverInterface::class)
+            ->addTag('asset_mapper.importmap.resolver');
```
nicolas-grekas, line 1332: this new tag allows defining custom package resolvers in case one wants to host their own

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -0,0 +1,145 @@
...
+{
+    public const URL_PATTERN_VERSION = 'https://data.jsdelivr.com/v1/packages/npm/%s/resolved?specifier=%s';
+    public const URL_PATTERN_DIST = 'https://cdn.jsdelivr.net/npm/%s@%s%s/+esm';
+
+    private HttpClientInterface $httpClient;
+
+    public function __construct(
+        HttpClientInterface $httpClient = null,
+        private readonly string $versionUrlPattern = self::URL_PATTERN_VERSION,
+        private readonly string $distUrlPattern = self::URL_PATTERN_DIST,
```
nicolas-grekas, line 33: I made those URLs configurable to make it possible to reuse the code while self-hosting somewhere else.

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -0,0 +1,145 @@
...
+            $i = strpos($packageName, '/');
+
+            if ($i && (!str_starts_with($packageName, '@') || $i = strpos($packageName, '/', $i + 1))) {
+                // @vendor/package/filepath or package/filepath
+                $filePath = substr($packageName, $i);
+                $packageName = substr($packageName, 0, $i);
+            }
+
+            $response = $this->httpClient->request('GET', sprintf($this->versionUrlPattern, $packageName, urlencode($constraint)));
+            $requiredPackages[] = [$options, $response, $packageName, $filePath, $options];
```
nicolas-grekas, line 64: by stacking HTTP requests, we can fetch versions concurrently

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -0,0 +1,145 @@
...
+
+        $errors = [];
+        foreach ($requiredPackages as $i => [$options, $response, $packageName, $filePath]) {
+            if (200 !== $response->getStatusCode()) {
+                $errors[] = [$options->packageName, $response];
+                continue;
+            }
+
+            $version = $response->toArray()['version'];
+            $requiredPackages[$i][1] = $this->httpClient->request('GET', sprintf($this->distUrlPattern, $packageName, $version, $filePath));
```
nicolas-grekas, line 75: and we can also concurrently fetch package dists

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JspmResolver.php
```diff
@@ -0,0 +1,102 @@
...
+            $response->getHeaders();
+        }
+
+        // if we're requiring just one package, in case it has any peer deps, match the preload
+        $defaultOptions = $packagesToRequire[0];
+
+        $resolvedPackages = [];
+        foreach ($response->toArray()['map']['imports'] as $packageName => $url) {
+            $options = $packageRequiresByName[$packageName] ?? new PackageRequireOptions($packageName, null, $defaultOptions->download, $defaultOptions->preload);
+            $resolvedPackages[] = [$options, $url, $options->download ? $this->httpClient->request('GET', $url, ['base_uri' => $this->baseUri]) : null];
```
nicolas-grekas, line 89: fetch contents from jspm concurrently also

src/Symfony/Component/AssetMapper/MappedAsset.php
```diff
@@ -20,119 +20,72 @@
...
-    /** @var AssetDependency[] */
     private array $dependencies = [];
-    /** @var string[] */
-    private array $fileDependencies = [];
-
-    public function __construct(private readonly string $logicalPath)
-    {
-    }
-
-    public function getLogicalPath(): string
```
nicolas-grekas, line 41: replacing most getters+setters by readonly props (except for "dependencies" arrays, which need mutability)

nicolas-grekas (approved):
I pushed a 2nd commit that addresses @stof's and @dunglas' comments + does some CS changes but more importantly fixes wiring of resolvers, sends requests in parallel and rewrites the MappedAsset class to use readonly properties.

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -0,0 +1,145 @@
...
+
+            throw new RuntimeException(sprintf('Error %d requiring packages from jsDelivr for "%s". Check your package names. Response: ', $response->getStatusCode(), $packages).$response->getContent(false), 0, $e);
+        }
+
+        if ($packagesToRequire) {
+            goto resolve_packages;
+        }
+
+        return array_values($resolvedPackages);
+    }
```
weaverryan, line 120: 😳

weaverryan (commented):
Looks great - I checked the changes, I don't see any issues. It's more polished - you accomplished the immutability with `MappedAsset`, which I couldn't quite pull off. Thank you!

nicolas-grekas:
Thank you @weaverryan.

PR #22690, opened 2017-05-11
[Console] Fix errors not rethrown even if not handled by console.error listeners

chalasr:
| Q             | A
| ------------- | ---
| Branch?       | 2.7
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | https://github.com/symfony/symfony/issues/22678
| License       | MIT
| Doc PR        | n/a

https://github.com/symfony/symfony/pull/22261 has been squashed while revisiting error handling, this fixes it again while keeping latest changes intact.

__code__
```php
public function execute(InputInterface $input, OutputInterface $output) {
    $this->barr();
}

public function bar() { }
```

__before__
before

__after__
after

PR #50420, opened 2023-05-24
[Console] add support for catching `\Throwable` errors

### LINKED ISSUE #46596: No catch Throwable errors

6.0

### Description

```       try {
            $exitCode = $this->doRun($input, $output);
        } catch (\Exception $e) {
            if (!$this->catchExceptions) {
                throw $e;
            }
```

Current code: https://github.com/symfony/console/blob/c9646197ef43b0e2ff44af61e7f0571526fd4170/Application.php#L172

Command exception handler skips type errors Throwable

### How to reproduce

Raise an exception that throws an exception not of type Exception

### Possible Solution

Change type to Throwable 

### Additional Context

_No response_

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #46596
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Symfony/Component/Console/Application.php
+++ b/src/Symfony/Component/Console/Application.php
@@ -79,6 +79,7 @@ class Application implements ResetInterface
     private string $version;
     private ?CommandLoaderInterface $commandLoader = null;
     private bool $catchExceptions = true;
+    private bool $catchErrors = false;
     private bool $autoExit = true;
     private InputDefinition $definition;
     private HelperSet $helperSet;
@@ -172,8 +173,11 @@ class Application implements ResetInterface
 
         try {
             $exitCode = $this->doRun($input, $output);
-        } catch (\Exception $e) {
-            if (!$this->catchExceptions) {
+        } catch (\Throwable $e) {
+            if ($e instanceof \Exception && !$this->catchExceptions) {
+                throw $e;
+            }
+            if (!$e instanceof \Exception && !$this->catchErrors) {
                 throw $e;
             }
 
@@ -425,6 +429,14 @@ class Application implements ResetInterface
     public function setCatchExceptions(bool $boolean)
     {
         $this->catchExceptions = $boolean;
+    }
+
+    /**
+     * Sets whether to catch errors or not during commands execution.
+     */
+    public function setCatchErrors(bool $catchErrors = true): void
+    {
+        $this->catchErrors = $catchErrors;
     }
 
     /**
--- a/src/Symfony/Component/Console/CHANGELOG.md
+++ b/src/Symfony/Component/Console/CHANGELOG.md
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Add `SignalMap` to map signal value to its name
  * Multi-line text in vertical tables is aligned properly
+ * The application can also catch errors with `Application::setCatchErrors(true)`
 
 6.3
 ---
--- a/src/Symfony/Component/Console/Tests/ApplicationTest.php
+++ b/src/Symfony/Component/Console/Tests/ApplicationTest.php
@@ -771,10 +771,15 @@ class ApplicationTest extends TestCase
         $this->assertInstanceOf(\FooCommand::class, $application->find('foo:'));
     }
 
-    public function testSetCatchExceptions()
+    /**
+     * @testWith [true]
+     *           [false]
+     */
+    public function testSetCatchExceptions(bool $catchErrors)
     {
         $application = new Application();
         $application->setAutoExit(false);
+        $application->setCatchErrors($catchErrors);
         putenv('COLUMNS=120');
         $tester = new ApplicationTester($application);
 
@@ -796,6 +801,33 @@ class ApplicationTest extends TestCase
             $this->assertInstanceOf(\Exception::class, $e, '->setCatchExceptions() sets the catch exception flag');
             $this->assertEquals('Command "foo" is not defined.', $e->getMessage(), '->setCatchExceptions() sets the catch exception flag');
         }
+    }
+
+    /**
+     * @testWith [true]
+     *           [false]
+     */
+    public function testSetCatchErrors(bool $catchExceptions)
+    {
+        $application = new Application();
+        $application->setAutoExit(false);
+        $application->setCatchExceptions($catchExceptions);
+        $application->add((new Command('boom'))->setCode(fn () => throw new \Error('This is an error.')));
+
+        putenv('COLUMNS=120');
+        $tester = new ApplicationTester($application);
+
+        try {
+            $tester->run(['command' => 'boom']);
+            $this->fail('The exception is not catched.');
+        } catch (\Throwable $e) {
+            $this->assertInstanceOf(\Error::class, $e);
+            $this->assertSame('This is an error.', $e->getMessage());
+        }
+
+        $application->setCatchErrors(true);
+        $tester->run(['command' => 'boom']);
+        $this->assertStringContainsString('  This is an error.', $tester->getDisplay(true));
     }
 
     public function testAutoExitSetting()


ghost:
Historical question: were all the `try..catch` blocks in the Symfony code base modified from `\Exception` to `\Throwable` after the latter had been introduced? I can still see blocks that catch \Exception and I wonder if this is just an oversight or there is a reason behind this.

maxbeckers (approved):
LGTM

chalasr (changes requested):
The broken test covers the behavior for which we cannot just do this unfortunately. Check https://github.com/symfony/symfony/pull/22690 for background (tldr  errors must keep reaching custom error handlers).

lyrixx:
Hmm, When using the component directly without the error handler, a command can crash badly :/ 

What about adding a flag to choose if we should handle throwable or not?

chalasr:
Sounds like a good idea 👍

nicolas-grekas:
The very lines before the try/catch setup a global error/exception handler, why does this crash badly as you write, while we have this in place?

lyrixx:
@nicolas-grekas let's say you are building a project that run commands, when someone makes an error in the command, it will produces theses outputs:


### Without this patch:

```
>/tmp/foobar foobar  hello
PHP Fatal error:  Uncaught Error: Call to undefined function hello2() in /tmp/foobar/foobar.php:9
Stack trace:
#0 [internal function]: hello()
#1 /home/gregoire/dev/github.com/jolicode/foobar/src/Console/Command/TaskCommand.php(152): ReflectionFunction->invoke()
#2 /home/gregoire/dev/github.com/jolicode/foobar/vendor/symfony/console/Command/Command.php(326): Foobar\Console\Command\TaskCommand->execute()
#3 /home/gregoire/dev/github.com/jolicode/foobar/vendor/symfony/console/Application.php(1063): Symfony\Component\Console\Command\Command->run()
#4 /home/gregoire/dev/github.com/jolicode/foobar/src/Console/Application.php(57): Symfony\Component\Console\Application->doRunCommand()
#5 /home/gregoire/dev/github.com/jolicode/foobar/vendor/symfony/console/Application.php(320): Foobar\Console\Application->doRunCommand()
#6 /home/gregoire/dev/github.com/jolicode/foobar/src/Console/Application.php(45): Symfony\Component\Console\Application->doRun()
#7 /home/gregoire/dev/github.com/jolicode/foobar/vendor/symfony/console/Application.php(174): Foobar\Console\Application->doRun()
#8 /home/gregoire/dev/github.com/jolicode/foobar/src/Console/ApplicationFactory.php(28): Symfony\Component\Console\Application->run()
#9 /home/gregoire/dev/github.com/jolicode/foobar/bin/foobar(14): Foobar\Console\ApplicationFactory::run()
#10 {main}
  thrown in /tmp/foobar/foobar.php on line 9

Fatal error: Uncaught Error: Call to undefined function hello2() in /tmp/foobar/foobar.php:9
Stack trace:
#0 [internal function]: hello()
#1 /home/gregoire/dev/github.com/jolicode/foobar/src/Console/Command/TaskCommand.php(152): ReflectionFunction->invoke()
#2 /home/gregoire/dev/github.com/jolicode/foobar/vendor/symfony/console/Command/Command.php(326): Foobar\Console\Command\TaskCommand->execute()
#3 /home/gregoire/dev/github.com/jolicode/foobar/vendor/symfony/console/Application.php(1063): Symfony\Component\Console\Command\Command->run()
#4 /home/gregoire/dev/github.com/jolicode/foobar/src/Console/Application.php(57): Symfony\Component\Console\Application->doRunCommand()
#5 /home/gregoire/dev/github.com/jolicode/foobar/vendor/symfony/console/Application.php(320): Foobar\Console\Application->doRunCommand()
#6 /home/gregoire/dev/github.com/jolicode/foobar/src/Console/Application.php(45): Symfony\Component\Console\Application->doRun()
#7 /home/gregoire/dev/github.com/jolicode/foobar/vendor/symfony/console/Application.php(174): Foobar\Console\Application->doRun()
#8 /home/gregoire/dev/github.com/jolicode/foobar/src/Console/ApplicationFactory.php(28): Symfony\Component\Console\Application->run()
#9 /home/gregoire/dev/github.com/jolicode/foobar/bin/foobar(14): Foobar\Console\ApplicationFactory::run()
#10 {main}
  thrown in /tmp/foobar/foobar.php on line 9
```

### With this patch (there is color, and Symfony styling):
```
>/tmp/foobar foobar  hello

In foobar.php line 9:
                                       
  Call to undefined function hello2()  
                                       

hello

```

And if the developer wants more output, they can run the same command with `-v` flags

lyrixx:
Very simple reproducer : 

take only symfony/console, then run:

```php
<?php

require __DIR__.'/vendor/autoload.php';

use Symfony\Component\Console\SingleCommandApplication;

(new SingleCommandApplication())
    ->setCode(function () {
        echo "AAA";
        foobar();
    })
    ->run()
;
```

lyrixx:
I have rebased the PR, restored BC, added a new flag to handle errors, and added some tests.

I think it's ready ✅

src/Symfony/Component/Console/Application.php
```diff
@@ -172,10 +173,13 @@ public function run(InputInterface $input = null, OutputInterface $output = null
 
         try {
             $exitCode = $this->doRun($input, $output);
-        } catch (\Exception $e) {
+        } catch (\Throwable $e) {
             if (!$this->catchExceptions) {
                 throw $e;
             }
+            if (!$e instanceof \Exception && !$this->catchErrors) {
```
GromNaN, line 180: Instead of the `instanceof`, why not having 2 `catch`?

```php
try {
} catch (\Exception $e) {
    if (!$this->catchExceptions) {
        throw $e;
    }
} catch (\Throwable $e) {
    if (!$this->catchErrors) {
        throw $e;
    }
}
```

Someone may want to catch errors but not exceptions, who knows.

src/Symfony/Component/Console/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Add `SignalMap` to map signal value to its name
  * Multi-line text in vertical tables is aligned properly
+ * The application can also catch `\Throwable` exceptions
```
GromNaN:
```suggestion
 * The application can also catch errors with `Application::setCatchErrors(true)`
```

src/Symfony/Component/Console/Application.php
lyrixx: But we want the exception handling in the `} catch (\Exception $e) {` part. With your code we loose this feature.

src/Symfony/Component/Console/CHANGELOG.md
lyrixx: Thanks

src/Symfony/Component/Console/Application.php
GromNaN: Indeed, I didn't pay much attention to that. You could factor the error handling into a function, but that wouldn't be any better. Let's keep it that way.

GromNaN (approved):

src/Symfony/Component/Console/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Add `SignalMap` to map signal value to its name
  * Multi-line text in vertical tables is aligned properly
+ * The application can also catch errors with `Application::setCatchErrors(true)`
```
chalasr, line 9: It would be nice to describe a bit in which case it is useful, here and on the setter docblock

derrabus (approved):

src/Symfony/Component/Console/CHANGELOG.md
lyrixx: For the PHP Docs, I copied the one from `setCatchExceptions()`. And usually, we don't put description about usages in CHANGELOG, does'nt?

src/Symfony/Component/Console/Application.php
```diff
@@ -172,10 +173,13 @@ public function run(InputInterface $input = null, OutputInterface $output = null
 
         try {
             $exitCode = $this->doRun($input, $output);
-        } catch (\Exception $e) {
+        } catch (\Throwable $e) {
             if (!$this->catchExceptions) {
```
chalasr: what if `$e` is an `\Error` and `$this->catchErrors` is `true`?

chalasr (approved):

chalasr:
Thank you @lyrixx.

chalasr:
@Seldaek FYI

Seldaek:
@chalasr thanks for the ping, see https://github.com/composer/composer/commit/f4738d97b (I guess that's why you pinged me :)

PR #50421, opened 2023-05-25
[Mime] Remote email feature

mmouih:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | NO
| New feature?  | YES
| Deprecations? | NO
| Tickets       | -
| License       | MIT
| Doc PR        | TODO

Hello,

To send emails via sendingblue, I used Mime Email class with sendingBlue bridge,
I noticed an Exception while trying to send the email to a remote template hosted on sending blue; 
Sendingblue/brevo expects to receive a `templateId` header to select the template to be used for this email, that is supposed to be enough; 
I encountered a LogicException because an Email *should always have body or an attachment*,
The source of this exception comes from `Email::ensureBodyValid()` private method.

A simple solution for this issue, is to define an arbitrary body (I usually send the templateId as body text), so no big deal.
I do not consider this to be a bug, not a serious one at least.

That gave me the idea to make a pull request, to define a *special type* for *Remote email*, with I called `RemoteEmail` in the Mime component for emails hosted on the cloud, which allows the email to have an empty body, instead it will define two things, a **templateId**,  the **header name** to be sent the email api, 

I had to change the visibility of the class `Email::ensureBodyValid()` to protected instead of private, to be able to customize it's behaviour.

I included tests for `RemoteEmail` class in the Mime component, and Tests to integrate it with **SendingBlueBridge**, but I did not integration bridge tests to avoid unit tests issues in the Mailer component since the RemoteEmail class would not be recognised,

Sending Blue Test file in another branch: 
https://github.com/mmouih/symfony/blob/be08c7ca6bdf44ac148f09761cdd0b9e8c755190/src/Symfony/Component/Mailer/Bridge/Sendinblue/Tests/Transport/SendinblueApiTransportRemoteEmailTest.php

Example of a failure test: 
https://github.com/symfony/symfony/blob/75157ca610da69186b80c6dfafb61541121c53c4/src/Symfony/Component/Mime/Tests/RemoteEmailTest.php#L30-L42

A success example below:
https://github.com/symfony/symfony/blob/75157ca610da69186b80c6dfafb61541121c53c4/src/Symfony/Component/Mime/Tests/RemoteEmailTest.php#L44-L57

This RemoteEmail class is a subtitute to Email on most usages, however the body is always Empty.

### DIFF
--- a/src/Symfony/Component/Mime/CHANGELOG.md
+++ b/src/Symfony/Component/Mime/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+ * Change the visibility of Email::ensureBodyValid() to "protected" to override its behavior in child classes.  
+ * Add RemoteEmail to support remote emails with template defined on the cloud (empty body)
+
 6.3
 ---
 
--- a/src/Symfony/Component/Mime/Email.php
+++ b/src/Symfony/Component/Mime/Email.php
@@ -410,7 +410,7 @@ class Email extends Message
         parent::ensureValidity();
     }
 
-    private function ensureBodyValid(): void
+    protected function ensureBodyValid(): void
     {
         if (null === $this->text && null === $this->html && !$this->attachments) {
             throw new LogicException('A message must have a text or an HTML part or attachments.');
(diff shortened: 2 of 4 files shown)


fabpot:
We don't support email templates from mailer providers (as we wouldn't be able to abstract that for all providers).
Let's not add this feature as there is apparently a workaround that works great for you.
Thank you for the pull request and sorry it cannot make it.

PR #50423, opened 2023-05-25
[Cache] Redis garbage collector

### LINKED ISSUE #47835: Implement PrunableInterface on RedisTagAwareAdapter
### Description

See https://github.com/symfony/symfony/pull/47509#issuecomment-1245141595 for background.

Tagged items that expire can leave never-removed tags.

/cc @Spea

### Example

_No response_

zoidbergx:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? |no
| Tickets       | Fix #47835
| License       | MIT
| Doc PR        

added a garbage collector for Redis to remove Tags which have no references, see https://github.com/symfony/symfony/issues/47835 for further details

### DIFF
--- a/src/Symfony/Component/Cache/Adapter/AbstractTagAwareAdapter.php
+++ b/src/Symfony/Component/Cache/Adapter/AbstractTagAwareAdapter.php
@@ -35,7 +35,7 @@ abstract class AbstractTagAwareAdapter implements TagAwareAdapterInterface, TagA
     use AbstractAdapterTrait;
     use ContractsTrait;
 
-    protected const TAGS_PREFIX = "\1tags\1";
+    final protected const TAGS_PREFIX = "\1tags\1";
 
     protected function __construct(string $namespace = '', int $defaultLifetime = 0)
     {
--- a/src/Symfony/Component/Cache/Adapter/AbstractTagAwareAdapter.php
+++ b/src/Symfony/Component/Cache/Adapter/AbstractTagAwareAdapter.php
@@ -35,7 +35,7 @@ abstract class AbstractTagAwareAdapter implements TagAwareAdapterInterface, TagA
     use AbstractAdapterTrait;
     use ContractsTrait;
 
-    private const TAGS_PREFIX = "\1tags\1";
+    protected const TAGS_PREFIX = "\1tags\1";
 
     protected function __construct(string $namespace = '', int $defaultLifetime = 0)
     {
--- a/src/Symfony/Component/Cache/Adapter/RedisTagAwareAdapter.php
+++ b/src/Symfony/Component/Cache/Adapter/RedisTagAwareAdapter.php
@@ -23,6 +23,7 @@ use Symfony\Component\Cache\Exception\LogicException;
 use Symfony\Component\Cache\Marshaller\DeflateMarshaller;
 use Symfony\Component\Cache\Marshaller\MarshallerInterface;
 use Symfony\Component\Cache\Marshaller\TagAwareMarshaller;
+use Symfony\Component\Cache\PruneableInterface;
 use Symfony\Component\Cache\Traits\RedisTrait;
 
 /**
@@ -44,7 +45,7 @@ use Symfony\Component\Cache\Traits\RedisTrait;
  * @author Nicolas Grekas <p@tchwork.com>
  * @author André Rømcke <andre.romcke+symfony@gmail.com>
  */
-class RedisTagAwareAdapter extends AbstractTagAwareAdapter
+class RedisTagAwareAdapter extends AbstractTagAwareAdapter implements PruneableInterface
 {
     use RedisTrait;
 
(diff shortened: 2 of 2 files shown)


stof:
@zoidbergx it looks like you merged your non-rebased remote branch instead of forcing the push after the rebase was done, which means your PR contains duplicate commits in its history and a merge commit.

src/Symfony/Component/Cache/Adapter/AbstractTagAwareAdapter.php
```diff
@@ -35,7 +35,7 @@ abstract class AbstractTagAwareAdapter implements TagAwareAdapterInterface, TagA
     use AbstractAdapterTrait;
     use ContractsTrait;
 
-    private const TAGS_PREFIX = "\1tags\1";
+    protected const TAGS_PREFIX = "\0tags\0";
```
stof: Why is this changing the value of the constant ? IIRC, we stopped using `\0` because NUL bytes were causing issues with some storages.

src/Symfony/Component/Cache/Traits/RedisTrait.php
```diff
@@ -49,7 +49,7 @@ trait RedisTrait
         'failover' => 'none',
         'ssl' => null, // see https://php.net/context.ssl
     ];
-    private \Redis|Relay|\RedisArray|\RedisCluster|\Predis\ClientInterface $redis;
+    protected \Redis|Relay|\RedisArray|\RedisCluster|\Predis\ClientInterface $redis;
```
stof: This is a no-go. It would make the property protected in the RedisTagAwareAdapter.

The RedisTagAwareAdapter can already access the private property of the trait as it is the class using the trait.

src/Symfony/Component/Cache/Adapter/RedisTagAwareAdapter.php
```diff
@@ -305,4 +306,193 @@ private function getRedisEvictionPolicy(): string
...
+    }
+
+    /**
+     *
+     * @TODO Verify the LUA scripts are redis-cluster safe.
+     *
+     * @param bool $compressMode
+     * @return bool
+     */
+    protected function pruneOrphanedTags(bool $compressMode = false): bool
```
stof: why is this a protected method ? I don't see why we would introduced an inheritance-based extension point there, that we would have to maintain

src/Symfony/Component/Cache/Adapter/RedisTagAwareAdapter.php
```diff
@@ -305,4 +306,193 @@ private function getRedisEvictionPolicy(): string
...
+        return $tagKeys;
+    }
+
+
+    /**
+     * Checks all tags in the cache for orphaned items and creates a "report" array.
+     *
+     * By default, only completely orphaned tag keys are reported. If
+     * compressMode is enabled the report will include all orphaned cache item
+     * reference for the
```
stof: missing end of the sentence.

src/Symfony/Component/Cache/Adapter/RedisTagAwareAdapter.php
```diff
@@ -305,4 +306,193 @@ private function getRedisEvictionPolicy(): string
...
+     * @TODO Is there anything that can be done to reduce memory footprint?
+     *
+     * @param bool $compressMode
+     * @return array{tagKeys: string[], orphanedTagKeys: string[], orphanedTagReferenceKeys?: array<string, string[]>}
+     *      tagKeys: List of all tags in the cache.
+     *      orphanedTagKeys: List of tags that only reference orphaned cache items.
+     *      orphanedTagReferenceKeys: List of all orphaned cache item references per tag.
+     *                           Keyed by tag, value is the list of orphaned cache item keys.
+     */
+    protected function getOrphanedTagsStats(bool $compressMode = false): array
```
stof: why is this protected ?

src/Symfony/Component/Cache/Adapter/RedisTagAwareAdapter.php
```diff
@@ -305,4 +306,193 @@ private function getRedisEvictionPolicy(): string
 
         return $this->redisEvictionPolicy = '';
     }
+
+
+    protected function getPrefix(): string
```
stof: why is this protected ?

stof:
`bug fix: yes` in the description looks wrong to me (in general, your change cannot be both a bug fix and a new feature)

src/Symfony/Component/Cache/Adapter/AbstractTagAwareAdapter.php
zoidbergx: agreed was wrongly merged from 5.4

src/Symfony/Component/Cache/Traits/RedisTrait.php
zoidbergx: agreed, carless mistake from my side

src/Symfony/Component/Cache/Adapter/RedisTagAwareAdapter.php
zoidbergx: agreed, completed sentence

src/Symfony/Component/Cache/Adapter/RedisTagAwareAdapter.php
zoidbergx: agreed, changed visiblity

src/Symfony/Component/Cache/Adapter/RedisTagAwareAdapter.php
zoidbergx: agreed, changed visibilty

src/Symfony/Component/Cache/Adapter/RedisTagAwareAdapter.php
zoidbergx: this was thought to have the possibility to read out the cache stats e.g. to make it visible in a console / UI.
but I changed the visibility as we can live without for now

zoidbergx:
@stof > `bug fix: yes` in the description looks wrong to me (in general, your change cannot be both a bug fix and a new feature)

yeah, for me it was a little bit of both as the current implementation could lead into a buggy cache but it may be more like a feature to have a garbage collector for Redis

src/Symfony/Component/Cache/Adapter/RedisTagAwareAdapter.php
stof: protected does not allow that. A Console or UI would need to have a public method.

src/Symfony/Component/Cache/Adapter/AbstractTagAwareAdapter.php
```diff
@@ -35,7 +35,7 @@ abstract class AbstractTagAwareAdapter implements TagAwareAdapterInterface, TagA
     use AbstractAdapterTrait;
     use ContractsTrait;
 
-    private const TAGS_PREFIX = "\1tags\1";
+    protected const TAGS_PREFIX = "\1tags\1";
```
derrabus:
```suggestion
    final protected const TAGS_PREFIX = "\1tags\1";
```

src/Symfony/Component/Cache/Adapter/RedisTagAwareAdapter.php
```diff
@@ -305,4 +306,185 @@ private function getRedisEvictionPolicy(): string
...
+
+        // get all SET entries which are tagged
+        $getTagsLua = <<<'EOLUA'
+            redis.replicate_commands()
+            local cursor = ARGV[1]
+            local prefix = ARGV[2]
+            local tagPrefix = string.gsub(KEYS[1], prefix, "")
+            return redis.call('SCAN', cursor, 'COUNT', 5000, 'MATCH', '*' .. tagPrefix .. '*', 'TYPE', 'set')
+        EOLUA;
+        $cursor = null;
```
das-peter, line 341: This should be 0, null leads to something like: `ERR invalid cursor script: 5880028e0d564320af1c1aabb595c9985f5a1a04, on @user_script:5.`
This is used in SCAN which expects 0 to start and returns 0 when the iteration is finished: https://redis.io/docs/latest/commands/scan/ 
Unfortunately I can't find older docs and have to assume this is backward compatible.

src/Symfony/Component/Cache/Adapter/RedisTagAwareAdapter.php
```diff
@@ -305,4 +306,185 @@ private function getRedisEvictionPolicy(): string
...
+            return redis.call('SSCAN', KEYS[1], cursor, 'COUNT', 5000)
+        EOLUA;
+
+        $orphanedTagReferenceKeys = [];
+        $orphanedTagKeys = [];
+        // Iterate over each tag and check if its entries reference orphaned
+        // cache items.
+        foreach ($tagKeys as $tagKey) {
+            $tagKey = substr($tagKey, \strlen($prefix));
+            $cursor = null;
```
das-peter, line 390: This should be 0, null leads to something like: `ERR invalid cursor script: 5880028e0d564320af1c1aabb595c9985f5a1a04, on @user_script:5.`
This is used in SCAN which expects 0 to start and returns 0 when the iteration is finished: https://redis.io/docs/latest/commands/scan/ 
Unfortunately I can't find older docs and have to assume this is backward compatible.

nicolas-grekas:
Closing due to inactivity: there has been no update since June 2023, the runtime error das-peter reported with the SCAN cursor was never fixed, and the Redis tag key layout has since changed (namespace-based invalidation, `Relay\Cluster` support), so this would need to be rewritten rather than rebased. The underlying request in #47835 stays open and a fresh PR is very welcome. Thanks for the work you put into this.

nicolas-grekas:
Superseded by #65353, which rewrites this for the current tag layout.

PR #50425, opened 2023-05-25
[Validator] Allow single constraint to be passed to the `constraints` option of the `When` constraint

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | _NA_
| License       | MIT
| Doc PR        | Todo

It would be nice for the DX to allow passing a single constraint to `When`, allowing to write the following:

```php
final class MyClass
{
    public string $type;

    #[Assert\When(
        expression: 'this.type === ...',
        constraints: new Assert\NotNull()
    )]
    public ?int $thickness = null;
}
```

### DIFF
--- a/src/Symfony/Component/Validator/CHANGELOG.md
+++ b/src/Symfony/Component/Validator/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * Allow single integer for the `versions` option of the `Uuid` constraint
+ * Allow single constraint to be passed to the `constraints` option of the `When` constraint
 
 6.3
 ---
--- a/src/Symfony/Component/Validator/Constraints/When.php
+++ b/src/Symfony/Component/Validator/Constraints/When.php
@@ -13,10 +13,12 @@ namespace Symfony\Component\Validator\Constraints;
 
 use Symfony\Component\ExpressionLanguage\Expression;
 use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
+use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\Exception\LogicException;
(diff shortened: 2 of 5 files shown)


chalasr:
Symfony doesn't do backward-incompatible changes without deprecating first, this should target 6.4 and provide BC layers if/where needed.

alexandre-daubois:
@chalasr Nicolas and Christian are giving me instructions for the same issue [here](https://github.com/symfony/symfony/pull/50396). I'm having a look, thanks!

chalasr:
AFAICT there is no BC break here, this should be fine as-is on 6.4.

OskarStark:
> AFAICT there is no BC break here, this should be fine as-is on 6.4.

I agree, as it is widening the type

alexandre-daubois:
I thought again about it, you're right, it shouldn't cause any issue as it is in the constructor. Thank you both :pray:

nicolas-grekas:
This PR should be put on hold until we settle on https://github.com/symfony/symfony/pull/50396#issuecomment-1575013763

nicolas-grekas (approved):

src/Symfony/Component/Validator/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.0
```
xabbuh:
```suggestion
6.4
```

xabbuh (changes requested):

nicolas-grekas:
Thank you @alexandre-daubois.

PR #50436, opened 2023-05-25
[AssetMapper] Adding documentation to importmap.php to guide the user

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | Still TODO

I know we don't love having documentation in the core, but this ends up being documentation for the user. Normally, we would do this via a recipe, but as this file is generated in its entirety here, it's the only place.

If we wanted to change this to a documentation link later, we could. But the docs don't (quite) exist yet. I think this comment is very helpful.

Thanks!


src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -337,7 +337,22 @@ private function writeImportMapConfig(array $entries): void
         }
 
         $map = class_exists(VarExporter::class) ? VarExporter::export($importMapConfig) : var_export($importMapConfig, true);
-        file_put_contents($this->importMapConfigPath, "<?php\n\nreturn {$map};\n");
+        file_put_contents($this->importMapConfigPath, <<<EOF
+        <?php
```
fabpot, line 341: Can you add a blank line below?

fabpot: While changing this file, what about marking it as being generated?

DI does it:

```
* @internal This class has been auto-generated by the Symfony Dependency Injection Component.
```

And the router does it as well:

```
* This file has been auto-generated by the Symfony Routing Component.
```

fabpot (changes requested):

fabpot (approved):

fabpot:
Thank you @weaverryan.

PR #50437, opened 2023-05-25
[Filesystem] Follow symlinks when dumping files

ausi:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Currently, if `Filesystem::dumpFile()` is called on a symlink, the link gets replaced with a new file.

This is unexpected I think because both `Filesystem::appendToFile()` as well as `file_put_contents()` follow the symlink instead.

This PR changes `dumpFile()` to behave the same way.

### DIFF
--- a/src/Symfony/Component/Filesystem/Filesystem.php
+++ b/src/Symfony/Component/Filesystem/Filesystem.php
@@ -669,6 +669,12 @@ class Filesystem
 
         $dir = \dirname($filename);
 
+        if (is_link($filename) && $linkTarget = $this->readlink($filename)) {
+            $this->dumpFile(Path::makeAbsolute($linkTarget, $dir), $content);
+
+            return;
+        }
+
         if (!is_dir($dir)) {
             $this->mkdir($dir);
         }
--- a/src/Symfony/Component/Filesystem/Tests/FilesystemTest.php
+++ b/src/Symfony/Component/Filesystem/Tests/FilesystemTest.php
@@ -1091,14 +1091,16 @@ class FilesystemTest extends FilesystemTestCase
     {
         $this->markAsSkippedIfSymlinkIsMissing();
 
-        if ('\\' === \DIRECTORY_SEPARATOR) {
-            $this->markTestSkipped('Windows does not support creating "broken" symlinks');
+        if ('\\' === \DIRECTORY_SEPARATOR && \PHP_VERSION_ID < 70400) {
+            $this->markTestSkipped('Windows does not support reading "broken" symlinks in PHP < 7.4.0');
         }
 
         $file = $this->workspace.'/file';
         $link = $this->workspace.'/link';
 
+        touch($file);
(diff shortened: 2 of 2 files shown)


src/Symfony/Component/Filesystem/Filesystem.php
```diff
@@ -667,6 +667,10 @@ public function dumpFile(string $filename, $content)
             throw new \TypeError(sprintf('Argument 2 passed to "%s()" must be string or resource, array given.', __METHOD__));
         }
 
+        if (is_link($filename) && $linkTarget = $this->readlink($filename, true)) {
```
nicolas-grekas: This doesn't work if the symlink exists but is broken (but file_put_contents works with them)
Not sure how we should fix this. Any idea?

fritzmg: In this case `readlink` will return `false` and `dumpFile` will thus overwrite the invalid symlink, will it not?

ausi: I think it should behave the same as `file_put_contents()`
Fixed in 0479d87791f24fc2222c45469d37f66c63bbe1b1

nicolas-grekas (commented):
LGTM. Can you please consider this minor patch and figure out the failures on appveyor?

```diff
--- a/src/Symfony/Component/Filesystem/Filesystem.php
+++ b/src/Symfony/Component/Filesystem/Filesystem.php
@@ -667,14 +667,12 @@ class Filesystem
             throw new \TypeError(sprintf('Argument 2 passed to "%s()" must be string or resource, array given.', __METHOD__));
         }
 
-        if (is_link($filename) && $linkTarget = $this->readlink($filename)) {
-            $this->dumpFile(Path::makeAbsolute($linkTarget, \dirname($filename)), $content);
+        $dir = \dirname($filename);
 
-            return;
+        if (null !== $linkTarget = $this->readlink($filename)) {
+            $filename = Path::makeAbsolute($linkTarget, $dir);
         }
 
-        $dir = \dirname($filename);
-
         if (!is_dir($dir)) {
             $this->mkdir($dir);
```

ausi:
> Can you please consider this minor patch

I applied the `\dirname($filename)` changes in c2ffd538d57463d4d50963eaada41fa4cb4fa65a
Your other change would break recursive symlinks, I adjusted the tests to cover that case in 48a66c1c7f7d1aee325e65112a79190c2507040c

ausi:
> …and figure out the failures on appveyor?

Done in d729da4cceb8be2b44580898cfa07d558f6ad858
In Windows below PHP version 7.4 it seems impossible to read a broken symlink.

nicolas-grekas (approved):

nicolas-grekas:
Thank you @ausi.

PR #13761, opened 2015-02-22
Automatically process extensions when they implement CompilerPassInterface

wouterj:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Tests pass? | yes |
| Fixed tickets | - |
| License | MIT |
| Doc PR | todo (if people are in favor of this PR) |

Compiler passes are very powerfull, but also quite strange to work with. Especially when you just need a very simple compiler pass (like https://github.com/symfony-cmf/RoutingBundle/blob/master/DependencyInjection/Compiler/SetRouterPass.php). For 3 lines of code, you need to tweak your bundle class and create a new class. 

When using the DI component standalone, compiler passes are even harder to work with, as DI extensions can't register them. I believe that's why libraries like Behat make their extensions compiler passes by default.

I think it would be very easy to just implement an interface and have a `compile` method for the simple compiler pass stuff. If a bundle needs multiple compiler passes or need compiler passes to be executed at other times in the compile process, a bundle can use the normal compiler passes. But if it's just one simple thing, like replacing a definition or getting services with a specific tag, I think this method will be very usefull.

PR #50444, opened 2023-05-26
[DependencyInjection] Allow extensions to register compiler passes

wouterj:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | -
| License       | MIT
| Doc PR        | -

Without the bundle system, the DI extension is sometimes used for plugins/add-ons in projects. I've recently introduced this to the [phpdocumentor/guides](https://github.com/phpDocumentor/guides/blob/main/packages/guides-cli/src/DependencyInjection/ContainerFactory.php#L84-L94) package (which will probably be the new RST parser that we're going to use at symfony.com some time in the future).

To make the DI extension fully workable as a plugin class, it would be great if it could also register compiler passes. This PR introduces a new method for this purpose: `ExtensionInterface::build()` Most users will probably not get a deprecation, as `Extension` has a default implementation.

Build precedence is: `KernelInterface::build()` > `BundleInterface::build()` > `ExtensionInterface::build()`

### DIFF
--- a/src/Symfony/Component/DependencyInjection/CHANGELOG.md
+++ b/src/Symfony/Component/DependencyInjection/CHANGELOG.md
@@ -8,6 +8,7 @@ CHANGELOG
  * Deprecate `ContainerAwareInterface` and `ContainerAwareTrait`, use dependency injection instead
  * Add `defined` env var processor that returns `true` for defined and neither null nor empty env vars
  * Add `#[AutowireLocator]` attribute
+ * Add `ExtensionInterface::build(ContainerBuilder $container): void` to register compiler passes from an extension
 
 6.3
 ---
--- a/src/Symfony/Component/DependencyInjection/ContainerBuilder.php
+++ b/src/Symfony/Component/DependencyInjection/ContainerBuilder.php
@@ -209,6 +209,11 @@ class ContainerBuilder extends Container implements TaggedContainerInterface
         if (false !== $extension->getNamespace()) {
             $this->extensionsByNs[$extension->getNamespace()] = $extension;
         }
+
+        // @deprecated remove if statement in Symfony 7
+        if (method_exists($extension, 'build') && \is_callable([$extension, 'build'])) {
+            $extension->build($this);
+        }
     }
 
     /**
--- a/src/Symfony/Component/DependencyInjection/Extension/ExtensionInterface.php
+++ b/src/Symfony/Component/DependencyInjection/Extension/ExtensionInterface.php
@@ -16,6 +16,8 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
 /**
  * ExtensionInterface is the interface implemented by container extension classes.
  *
+ * @method void build(ContainerBuilder $container) Called just before compilation, use this to register compiler passes
+ *
  * @author Fabien Potencier <fabien@symfony.com>
  */
 interface ExtensionInterface
--- a/UPGRADE-6.4.md
+++ b/UPGRADE-6.4.md
@@ -15,6 +15,7 @@ DependencyInjection
 -------------------
 
  * Deprecate `ContainerAwareInterface` and `ContainerAwareTrait`, use dependency injection instead
+ * Add `ExtensionInterface::build(ContainerBuilder $container): void` to register compiler passes from an extension
 
    *Before*
    ```php
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/Functional/Bundle/ExtensionWithoutConfigTestBundle/DependencyInjection/ExtensionWithoutConfigTestExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/Functional/Bundle/ExtensionWithoutConfigTestBundle/DependencyInjection/ExtensionWithoutConfigTestExtension.php
@@ -16,6 +16,10 @@ use Symfony\Component\DependencyInjection\Extension\ExtensionInterface;
 
 class ExtensionWithoutConfigTestExtension implements ExtensionInterface
 {
+    public function build(ContainerBuilder $container): void
+    {
+    }
+
     public function load(array $configs, ContainerBuilder $container): void
     {
     }
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/Functional/app/AppKernel.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/Functional/app/AppKernel.php
@@ -78,7 +78,7 @@ class AppKernel extends Kernel implements ExtensionInterface, ConfigurationInter
         $loader->load($this->rootConfig);
     }
 
-    protected function build(ContainerBuilder $container): void
+    public function build(ContainerBuilder $container): void
     {
         $container->register('logger', NullLogger::class);
         $container->registerExtension(new TestDumpExtension());
--- a/src/Symfony/Component/DependencyInjection/Extension/Extension.php
+++ b/src/Symfony/Component/DependencyInjection/Extension/Extension.php
@@ -28,6 +28,10 @@ abstract class Extension implements ExtensionInterface, ConfigurationExtensionIn
 {
     private array $processedConfigs = [];
 
+    public function build(ContainerBuilder $container): void
+    {
+    }
+
     /**
      * @return string|false
      */
--- a/src/Symfony/Component/DependencyInjection/Tests/ContainerBuilderTest.php
+++ b/src/Symfony/Component/DependencyInjection/Tests/ContainerBuilderTest.php
@@ -39,6 +39,7 @@ use Symfony\Component\DependencyInjection\Exception\ParameterNotFoundException;
 use Symfony\Component\DependencyInjection\Exception\RuntimeException;
 use Symfony\Component\DependencyInjection\Exception\ServiceCircularReferenceException;
 use Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException;
+use Symfony\Component\DependencyInjection\Extension\Extension;
 use Symfony\Component\DependencyInjection\Extension\ExtensionInterface;
 use Symfony\Component\DependencyInjection\LazyProxy\Instantiator\RealServiceInstantiator;
 use Symfony\Component\DependencyInjection\Loader\ClosureLoader;
@@ -1326,6 +1327,17 @@ class ContainerBuilderTest extends TestCase
         $this->assertEquals([$second, $first], $configs);
     }
 
+    public function testExtensionBuild()
+    {
+        $container = new ContainerBuilder();
+
+        $container->registerExtension(new BuildExtension());
+
+        $container->compile();
+        $testpass = array_filter($container->getCompilerPassConfig()->getPasses(), fn ($pass) => $pass instanceof TestPass);
+        $this->assertCount(1, $testpass, 'The extension did not register a compiler pass.');
+    }
+
     public function testAbstractAlias()
     {
         $container = new ContainerBuilder();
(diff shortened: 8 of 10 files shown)


src/Symfony/Component/DependencyInjection/Extension/CompilerPassProvidingExtensionInterface.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Symfony\Component\DependencyInjection\Extension;
+
+interface CompilerPassProvidingExtensionInterface
+{
+    /**
+     * @return array<CompilerPassInterface>|array<array{0: CompilerPassInterface, 1: string, 2?: int}>
```
stof: The right type is actually `array<CompilerPassInterface|array{0: CompilerPassInterface, 1: string, 2?: int}>` as you allow mixing both types of items in the returned array.

And you are missing the use statement for CompilerPassInterface

stof (changes requested):

src/Symfony/Component/DependencyInjection/Extension/CompilerPassProvidingExtensionInterface.php
stof: I would also document the index 1 as `PassConfig::TYPE_*` instead of `string`, as we don't allow any string but only one of those constants (don't forget to add the usage statement)

src/Symfony/Component/DependencyInjection/Extension/AbstractExtension.php
```diff
@@ -22,7 +23,7 @@
  *
  * @author Yonel Ceruto <yonelceruto@gmail.com>
  */
-abstract class AbstractExtension extends Extension implements ConfigurableExtensionInterface, PrependExtensionInterface
+abstract class AbstractExtension extends Extension implements ConfigurableExtensionInterface, PrependExtensionInterface, CompilerPassProvidingExtensionInterface, CompilerPassInterface
```
derrabus: I wouldn't have the `AbstractExtension` implement `CompilerPassInterface`, tbh. If I override `getCompilerPasses()` and forget adding `$this` to the returned array, my extension has a `process()` method that is ignored. That looks like a trap.

src/Symfony/Component/DependencyInjection/Extension/AbstractExtension.php
```diff
@@ -62,4 +67,9 @@ final public function load(array $configs, ContainerBuilder $container): void
 
         $this->executeConfiguratorCallback($container, $callback, $this);
     }
+
+    public function getCompilerPasses(): array
```
derrabus: Would it make sense to use the same signature as we use on the `BundleInterface`?

https://github.com/symfony/symfony/blob/4813d669ff27cc0d1c58a69af9c781b68d25c052/src/Symfony/Component/HttpKernel/Bundle/BundleInterface.php#L39-L46

That allows us to call `$container->addCompilerPass()` from the extension which I believe is a bit easier to understand than the array structure that we're supposed to return here.

derrabus (changes requested):
Blocking the merge because there are no tests. 😉

src/Symfony/Component/DependencyInjection/Extension/AbstractExtension.php
ro0NL: i assume extensions implementing CompilerPassInterface are already accounted for in symfony? they should not be ignored by this PR

but i'd prefer users letting implement CompilerPassInterface as needed rather than knowing they can override default process() from AbstractExtension

edit: OH it's glued by `return [$this];` below, i agree with @derrabus :)

src/Symfony/Component/DependencyInjection/Extension/AbstractExtension.php
ro0NL: what about merging compiler passes in ContainerBuilder::merge? then one can register passes from `load()` directly

src/Symfony/Component/DependencyInjection/Extension/AbstractExtension.php
yceruto: Indeed, `ExtensionCompilerPass` is intended to support extensions implementing the `CompilerPassInterface`. However, this is a separate topic from the one suggested by the title of the PR 😕.

src/Symfony/Component/DependencyInjection/Extension/AbstractExtension.php
yceruto: I second @derrabus's idea. A new interface featuring the `public function build(ContainerBuilder $container)` method might be sufficient for adding your custom compiler passes via the extension.

src/Symfony/Component/DependencyInjection/ContainerBuilder.php
```diff
@@ -211,6 +212,18 @@ public function registerExtension(ExtensionInterface $extension)
         if (false !== $extension->getNamespace()) {
             $this->extensionsByNs[$extension->getNamespace()] = $extension;
         }
+
+        if ($extension instanceof CompilerPassProvidingExtensionInterface) {
```
yceruto: Following the previous suggestions, I would do this in `Kernel::prepareContainer()` instead, where the `build()` calls are happening. Likely inside the last extensions loop.

yceruto (commented):
I've the same need in some projects. Thanks for working on it!

src/Symfony/Component/DependencyInjection/Extension/AbstractExtension.php
ro0NL: what's wrong with ContainerBuilder::merge? build+load seems confusing, or could be avoided maybe?

edit: right, MergeExtension is a pass itself :facepalm: that complicates it a bit

src/Symfony/Component/DependencyInjection/Extension/AbstractExtension.php
wouterj: Ah, I somehow forgot that I contributed that `ExtensionCompilerPass` feature years ago :smile: 

I'll revert this change.

src/Symfony/Component/DependencyInjection/ContainerBuilder.php
wouterj: I've instead moved this to the start of `ContainerBuilder::compile()`. The idea of this feature is to do this within the DI component, as it's useful for standalone use-cases of the DI component. When using it with HttpKernel, I would always recommend to use `Kernel::build()` or `BundleInterface::build()` instead.

This is as close to bundle/kernel `build()` as we can get in the DI component.

wouterj:
Thanks for the reviews. PR is updated by only introducing an `ExtensionInterface::build()` method. Somehow, I forgot that implementing `CompilerPassInterface` from a DI extension already works.

There is one decision left: Should we move the `build()` method to a specific interface (like we do for configuration), to remove the deprecation notice?
I think it's safe like this, most of the users will extend `Extension` or `AbstractExtension`, which already implement this method for them. Having yet another interface that you can implement in an extension seems a bit overkill.

src/Symfony/Component/DependencyInjection/ContainerBuilder.php
```diff
@@ -751,6 +751,13 @@ public function deprecateParameter(string $name, string $package, string $versio
      */
     public function compile(bool $resolveEnvPlaceholders = false)
     {
+        foreach ($this->extensions as $extension) {
+            // @deprecated remove if statement in Symfony 7
+            if (method_exists($extension, 'build')) {
+                $extension->build($this);
+            }
```
derrabus: … else trigger a deprecation? The method will be enforced by `ExtensionInterface` in v7, right?

wouterj: The DebugClassLoader automatically triggers a deprecation when implementing `ExtensionInterface` without this method (thanks to the `@method build` annotation).

(I've used this deprecation to find out which fixtures I had to update)

wouterj:
Please keep merge blocked: One test failure is related (need a BC layer for `KernelInterface` classes implementing `ExtensionInterface`).

wouterj:
Status: Needs review

The tests failures that are remaining are unrelated.

derrabus (approved):

derrabus:
Would it make sense to deprecate `BundleInterface::build()`?

stof:
> Would it make sense to deprecate `BundleInterface::build()`?

Clearly not yet. Having the same version adding a new way to register compiler passes and deprecating the existing way would be a nightmare for the ecosystem.
If we want to deprecate it, it should be done after the EOM of Symfony 5.4 so that bundles don't need to drop support of the 5.4 LTS as soon as 6.4 is released.

Side note, the comment in https://github.com/symfony/symfony/pull/50444#discussion_r1225282599 makes me think that we might not want to deprecate it at all.

yceruto (approved):

src/Symfony/Component/DependencyInjection/Extension/ExtensionInterface.php
```diff
@@ -16,6 +16,8 @@
 /**
  * ExtensionInterface is the interface implemented by container extension classes.
  *
+ * @method void build(ContainerBuilder $container) Called just before compilation, use this to register compiler passes
```
ro0NL, line 19: > use this to register compiler passes

or anything really ... load() v build() is confusing at this point IMHO

src/Symfony/Component/DependencyInjection/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+ * Add `ExtensionInterface::build(ContainerBuilder $container): void` to register compiler passes from an extension.
```
nicolas-grekas:
```suggestion
 * Add `ExtensionInterface::build(ContainerBuilder $container): void` to register compiler passes from an extension
```

src/Symfony/Component/DependencyInjection/ContainerBuilder.php
```diff
@@ -751,6 +751,13 @@ public function deprecateParameter(string $name, string $package, string $versio
      */
     public function compile(bool $resolveEnvPlaceholders = false)
     {
+        foreach ($this->extensions as $extension) {
+            // @deprecated remove if statement in Symfony 7
+            if (method_exists($extension, 'build') && \is_callable([$extension, 'build'])) {
+                $extension->build($this);
```
nicolas-grekas: Shouldn't this happen in Kernel::prepareContainer instead?
This would give the highest precedence to the App\Kernel, which looks desired to me.

wouterj: This feature is meant to be used when using the DI component standalone (like in `phpdocumentor/guides`), so that doesn't work.

If the precedence is important, we can maybe call this from `Container::registerExtension()`? In that case, the precedence would be Kernel, Bundle, Extension ?

nicolas-grekas: I think that'd be better yes.

wouterj:
Precedence fixed and updated the PR description to match current state. Ready to merge as far as I'm concerned.

wouterj:
The failing test revealed a conflict when you have a class `Kernel&ExtensionInterface`. In this case, the `build()` method of that class is called twice (once for both interfaces).

As the primary use-case of this PR is already supported (#13761), I'll close here. If someone else is interested in this feature, feelf free to continue with what's available in this PR :)

PR #50445, opened 2023-05-26
[AssetMapper] Add "=alias" syntax to importmap:require

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | yes-ish
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | Still TODO

I'm doing final testing against all of the UX packages. There is one edge-case where we need to require a specific package path - `vue/dist/vue.esm-bundler.js` - but alias it (i.e. make its key) set to something else - `vue` - in `importmap.php`. We need the command to support this so that https://github.com/symfony/flex/pull/975 can use it when installing UX packages.

Hopefully the last thing I come across - I've done a LOT of testing at this point.

Thanks!


fabpot (approved):

fabpot:
Thank you @weaverryan.

PR #50447, opened 2023-05-26
[DependencyInjection] Add support for excluding services with declared custom attribute

angelov:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       |  /
| License       | MIT
| Doc PR        | TODO

Currently, if we want to exclude classes from registering as services in the container, we can add their namespace(s) in the app configuration, or use the `#[Exclude]`/`#[When('never')]` attributes on each of the classes.

When it comes to certain types of classes, eg. entities, they can be placed all around the apps in many different folders, which can result in many entries under the `exclude` configuration. It is also usual for us to have many entity classes, so adding the attribute to each of them can be error-prone.

However, classes of a same type may already have a common attribute declared on them - such as `Doctrine\ORM\Mapping\Entity`.

Because of that, I'm proposing a new way for excluding classes by making it possible to register any attribute for exclusion. 

Underneath, this is just extending the functionality for handling the classes with `#[Exclude]` attribute by additionally checking the list of attributes marked for exclusion.

**Usage:**

To exclude all classes with the `#[Entity]` attribute:

```php
use Doctrine\ORM\Mapping\Entity;

{
    use MicroKernelTrait;

    protected function build(ContainerBuilder $container)
    {
        $container->registerAttributeForExclusion(Entity::class);
    }
}
```

### DIFF
--- a/src/Symfony/Component/DependencyInjection/CHANGELOG.md
+++ b/src/Symfony/Component/DependencyInjection/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+* Add support for excluding services with declared custom attribute (`ContainerBuilder::registerAttributeForExclusion`)
+
 7.0
 ---
 
--- a/src/Symfony/Component/DependencyInjection/ContainerBuilder.php
+++ b/src/Symfony/Component/DependencyInjection/ContainerBuilder.php
@@ -132,6 +132,11 @@ class ContainerBuilder extends Container implements TaggedContainerInterface
     private array $autoconfiguredAttributes = [];
 
     /**
+     * @var array<class-string>
+     */
+    private array $exclusionAttributes = [];
+
+    /**
      * @var array<string, bool>
      */
     private array $removedIds = [];
@@ -1342,6 +1347,20 @@ class ContainerBuilder extends Container implements TaggedContainerInterface
     }
 
     /**
+     * Registers an attribute that will be used for excluding all services from classes it is declared on.
+     *
+     * @param class-string $attributeClass
+     */
+    public function registerAttributeForExclusion(string $attributeClass): void
+    {
+        if (\in_array($attributeClass, $this->exclusionAttributes)) {
+            return;
+        }
+
+        $this->exclusionAttributes[] = $attributeClass;
+    }
+
+    /**
      * Registers an autowiring alias that only binds to a specific argument name.
      *
      * The argument name is derived from $name if provided (from $id otherwise)
@@ -1384,6 +1403,14 @@ class ContainerBuilder extends Container implements TaggedContainerInterface
     public function getAutoconfiguredAttributes(): array
     {
         return $this->autoconfiguredAttributes;
+    }
+
+    /**
+     * @return array<class-string>
+     */
+    public function getExclusionAttributes(): array
+    {
+        return $this->exclusionAttributes;
     }
 
     /**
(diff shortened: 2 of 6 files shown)


angelov:
The failures in the checks seem unrelated.

src/Symfony/Component/DependencyInjection/Tests/Fixtures/Utils/ToBeExcluded.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Symfony\Component\DependencyInjection\Tests\Fixtures\Utils;
+
+#[\Attribute(\Attribute::TARGET_CLASS)]
+class ToBeExcluded
```
nicolas-grekas, line 6: It'd be great to allow this feature to work even if the attribute class doesn't not exist.
Can you remove this one and make the feature still work?

src/Symfony/Component/DependencyInjection/Loader/FileLoader.php
```diff
@@ -125,10 +125,15 @@ public function registerClasses(Definition $prototype, string $namespace, string
         foreach ($classes as $class => $errorMessage) {
             if (null === $errorMessage && $autoconfigureAttributes) {
                 $r = $this->container->getReflectionClass($class);
-                if ($r->getAttributes(Exclude::class)[0] ?? null) {
-                    $this->addContainerExcludedTag($class, $source);
-                    continue;
+                $exclusionAttributes = $this->container->getExclusionAttributes();
```
nicolas-grekas, line 143: IIRC, file loaders are run on separate container instances, right? This would mean that a bundle cannot expect registerAttributeForExclusion to have any effect on classes in `src/`. Am I correct? If yes, don't we want to make these app-wide instead?

src/Symfony/Component/DependencyInjection/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+* Add support for excluding services with declared custom attribute (`ContainerBuilder::registerAttributeForExclusion`)
```
nicolas-grekas, line 7:
```suggestion
 * Add support for excluding services by attribute with `ContainerBuilder::registerAttributeForExclusion()`
```

src/Symfony/Component/DependencyInjection/ContainerBuilder.php
```diff
@@ -131,6 +132,11 @@ class ContainerBuilder extends Container implements TaggedContainerInterface
      */
     private array $autoconfiguredAttributes = [];
 
+    /**
+     * @var array<class-string>
```
nicolas-grekas, line 136: ?
```suggestion
     * @var array<class-string, true>
```

src/Symfony/Component/DependencyInjection/Loader/FileLoader.php
nicolas-grekas: what about this? did you give it a try? configuring an exclusion within a bundle and seeing it work on the app?

nicolas-grekas (commented):
Small rebase needed, but more importantly: did you try using registerForAutoconfiguration to define the container.exclude tag?

nicolas-grekas:
Let me close as I think this has been resolved in 7.3 with the introduction of resource definitions
https://symfony.com/blog/new-in-symfony-7-3-dependency-injection-improvements#resource-tags

PR #46279, opened 2022-05-06
[DependencyInjection] Optimize autowiring logic by telling it about excluded symbols

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Replaces https://github.com/symfony/symfony/pull/46275

While inspecting the dumped container of the demo app, I noticed that it contains so-called "errored" services. They come e.g from entities that are used as type-hints on controllers. To compute those services, the autowiring logic currently loads all services looking for possible type-compatible candidates.

But here, we should know that those won't be found: they're excluded from being loaded.

Instead of completely ignoring excluded classes, this PR still registers them, but as abstract and with a special `container.excluded` tag.

This allows the rest of the compilation logic to skip+remove them, while allowing `AutowiringPass` to compute a better error message and way faster.

Here is the relevant part of a https://blackfire.io comparison:

image

PR #50452, opened 2023-05-27
Hide definitions bearing the `container.excluded` tag

### LINKED ISSUE #50417: Service Container no working "exclude"

6.2.8

### Description

In version `6.2.8` I get a Service Container error. Files and folders are not excluded. Instead, folders are added to the container as `Service ID`: [folder/filename] and `Class name`: [folder/filename]. At the same time, it works on version `6.1.*`

Example: https://drive.google.com/file/d/1AfmbLkxpFGtixKnATdKToJUa5x_inl8v/view?usp=share_link
For example, `Command`, `Query`, `Entity` are folders, but they are specified as services in the container. 

### How to reproduce

Configuration:

**src/Kernel.php**
```php
<?php

declare(strict_types=1);

namespace App;

use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;

final class Kernel extends BaseKernel
{
    use MicroKernelTrait;

    private function configureContainer(ContainerConfigurator $container): void
    {
        $configDir = $this->getConfigDir();
        $container->import($configDir . '/{packages}/*.{php,yaml}');
        $container->import($configDir . '/{services}/*.{php,yaml}');
        $container->import('./**/config.{php,yaml}');// import config from module
    }
}
```

**src/Auth/config.php**
```php
<?php

declare(strict_types=1);

namespace App\Auth;

use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;

use function Symfony\Component\DependencyInjection\Loader\Configurator\service;

return static function (ContainerConfigurator $configurator): void {
    $services = $configurator
        ->services()
        ->defaults()
        ->autowire()
        ->autoconfigure();

    $services->load('App\\Auth\\', '.')->exclude([
        './config.php',
        './Command',
        './Entity',
        './Query',
        './Test',
        './Validator/Constraint',
    ]);

    $services->load('App\\Auth\\Command\\', './Command/**/Handler.php');
    $services->load('App\\Auth\\Entity\\', './Entity/**/*Repository.php');
    $services->load('App\\Auth\\Query\\', './Query/**/Handler.php');
};
```
**config**
Global file `services.yaml` not used.

Example App: https://github.com/Myks92/symfony-service-container/tree/master

### Possible Solution

_No response_

### Additional Context

_No response_

Myks92:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| | Depreciation? | no
| Tickets       | Fix #50417
| License       | MIT
| Doc PR        | -

Normally I test whether the exclusion rules from the service container of my bundles work correctly by just using the `debug:container` command and looking whether my excluded services occur in the last.
However, due to the latest changes in #46279 all folders and subfulders (excluded or not) are always in this list.

You need to open the definition to see if it was excluded:
Service Container

For example, `Command`, `Query`, `Entity` are folders, but they are specified as services in the container. 

This PR fixes this problem to ignore definitions bearing the container.excluded tag if those are to be hidden for Descriptors: `JsonDescriptor`,  `MarkdownDescriptor`, `TextDescriptor` and `XmlDescriptor`. 

After fixes problem:
Снимок экрана 2023-05-27 в 17 03 19

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/JsonDescriptor.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/JsonDescriptor.php
@@ -106,6 +106,9 @@ class JsonDescriptor extends Descriptor
             if ($service instanceof Alias) {
                 $data['aliases'][$serviceId] = $this->getContainerAliasData($service);
             } elseif ($service instanceof Definition) {
+                if ($service->hasTag('container.excluded')) {
+                    continue;
+                }
                 $data['definitions'][$serviceId] = $this->getContainerDefinitionData($service, $omitTags, $showArguments, $builder, $serviceId);
             } else {
                 $data['services'][$serviceId] = $service::class;
--- a/src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/TextDescriptor.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/TextDescriptor.php
@@ -198,6 +198,10 @@ class TextDescriptor extends Descriptor
             }
 
             if ($definition instanceof Definition) {
+                if ($definition->hasTag('container.excluded')) {
+                    unset($serviceIds[$key]);
+                    continue;
+                }
                 if ($showTag) {
                     $tags = $definition->getTag($showTag);
                     foreach ($tags as $tag) {
--- a/src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/MarkdownDescriptor.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/MarkdownDescriptor.php
@@ -162,6 +162,9 @@ class MarkdownDescriptor extends Descriptor
             if ($service instanceof Alias) {
                 $services['aliases'][$serviceId] = $service;
             } elseif ($service instanceof Definition) {
+                if ($service->hasTag('container.excluded')) {
+                    continue;
+                }
                 $services['definitions'][$serviceId] = $service;
             } else {
                 $services['services'][$serviceId] = $service;
--- a/src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/XmlDescriptor.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/XmlDescriptor.php
@@ -294,6 +294,12 @@ class XmlDescriptor extends Descriptor
                 continue;
             }
 
+            if ($service instanceof Definition) {
+                if ($service->hasTag('container.excluded')) {
+                    continue;
+                }
+            }
+
             $serviceXML = $this->getContainerServiceDocument($service, $serviceId, null, $showArguments);
             $containerXML->appendChild($containerXML->ownerDocument->importNode($serviceXML->childNodes->item(0), true));
         }
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/Fixtures/ContainerExcluded.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/Fixtures/ContainerExcluded.php
@@ -1,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Symfony\Bundle\FrameworkBundle\Tests\Fixtures;
+
+/**
+ * @author Maksim Vorozhtsov <myks1992@mail.ru>
+ */
+class ContainerExcluded
+{
+}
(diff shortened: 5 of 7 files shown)


nicolas-grekas:
Can you please check if this issue happens on 6.3? We might have fixed this as an improvement already.

src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/JsonDescriptor.php
```diff
@@ -103,6 +103,10 @@ protected function describeContainerServices(ContainerBuilder $builder, array $o
                 continue;
             }
 
+            if (!$showHidden && $builder->hasDefinition($serviceId) && $builder->getDefinition($serviceId)->hasTag('container.excluded')) {
```
chalasr: This won’t work if the service is an alias (see hasDefinition() vs has())

MatTheCat:
@nicolas-grekas I can confirm this still happens with Symfony 6.3.0 RC2

src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/JsonDescriptor.php
MatTheCat: Use the existing `$service` variable instead of calling `$builder->getDefinition($serviceId)` (this is the case for all descriptors).

Myks92: Thank you for your help! Fixed it.

Myks92: Thanks. I’m fixed with help use existing variable `$service`. Do you have any other comments?

MatTheCat (approved):
LGTM, maybe just make line breaks consistent 👍

src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/JsonDescriptor.php
```diff
@@ -106,6 +106,9 @@ protected function describeContainerServices(ContainerBuilder $builder, array $o
             if ($service instanceof Alias) {
                 $data['aliases'][$serviceId] = $this->getContainerAliasData($service);
             } elseif ($service instanceof Definition) {
+                if (!$showHidden && $service->hasTag('container.excluded')) {
```
chalasr: Hidden services are true services. Mixing them with excluded services would give a false impression that those are available for use. Let's just skip them in all cases

chalasr (changes requested):

src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/JsonDescriptor.php
Myks92: I agree. Rule for hidden services already exists in [102](https://github.com/symfony/symfony/blob/48a333caf283e005fd49a7824a90533f430121e3/src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/JsonDescriptor.php#L102) line, so I deleted this. Is that better or did I misunderstand you?

chalasr: 👍

chalasr:
Looks good, thank you for contributing.
Can you have a look ContainerDebugCommandTest and see if you can add a test to prevent future regressions?

Myks92:
@chalasr, I added tests for `Descriptor` and `ContainerDebugCommand`. Is that what you need?

chalasr (approved):

chalasr:
Thank you @Myks92.

src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/TextDescriptor.php
```diff
@@ -198,6 +198,10 @@ protected function describeContainerServices(ContainerBuilder $builder, array $o
             }
 
             if ($definition instanceof Definition) {
+                if ($definition->hasTag('container.excluded')) {
+                    unset($serviceIds[$key]);
```
xabbuh, line 202: Don't we have to do this in the other descriptors as well?

MatTheCat: `TextDescriptor` does this because it loops twice on `$serviceIds` but this is not the other descriptors’ case, so I’d say this is not needed :thinking:

PR #50432, opened 2023-05-25
[Security] Validate `aud` and `iss` claims on OidcTokenHandler

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

According to the [OIDC Specification](https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation), the `aud` and `iss` claims MUST be validated by the server.

PR #50453, opened 2023-05-27
[SecurityBundle] add missing xsd definition for OIDC

aegypius:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Follow-up of symfony/symfony#50432

I hope I didn't broke anything this PR is far out of my comfort zone !

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/Resources/config/schema/security-1.0.xsd
+++ b/src/Symfony/Bundle/SecurityBundle/Resources/config/schema/security-1.0.xsd
@@ -309,12 +309,44 @@
     <xsd:complexType name="access_token">
         <xsd:choice minOccurs="0" maxOccurs="unbounded">
             <xsd:element name="token-extractor" type="xsd:string" />
+            <xsd:element name="token-handler" type="oidc_token_handler" />
         </xsd:choice>
         <xsd:attribute name="token-handler" type="xsd:string" />
         <xsd:attribute name="realm" type="xsd:string" />
         <xsd:attribute name="success-handler" type="xsd:string" />
         <xsd:attribute name="failure-handler" type="xsd:string" />
         <xsd:attribute name="provider" type="xsd:string" />
+    </xsd:complexType>
+
+    <xsd:complexType name="oidc_token_handler">
+        <xsd:sequence>
+            <xsd:choice minOccurs="0" maxOccurs="1">
+                <xsd:element name="oidc-user-info" type="oidc_user_info"></xsd:element>
+                <xsd:element name="oidc" type="oidc"></xsd:element>
+            </xsd:choice>
+        </xsd:sequence>
+        <xsd:attribute name="oidc-user-info" type="xsd:anyURI"></xsd:attribute>
+    </xsd:complexType>
+
+    <xsd:complexType name="oidc_user_info">
+        <xsd:attribute name="base-uri" type="xsd:anyURI" use="required" />
+        <xsd:attribute name="claim" type="xsd:string" />
+        <xsd:attribute name="client" type="xsd:string" />
+    </xsd:complexType>
+
+    <xsd:complexType name="oidc">
+        <xsd:sequence>
+            <xsd:choice minOccurs="1" maxOccurs="1">
+                <xsd:element name="signature" type="oidc_signature" />
+            </xsd:choice>
+        </xsd:sequence>
+        <xsd:attribute name="claim" type="xsd:string" />
+        <xsd:attribute name="audience" type="xsd:string" />
+    </xsd:complexType>
+
+    <xsd:complexType name="oidc_signature">
+        <xsd:attribute name="algorithm" type="xsd:string" use="required" />
+        <xsd:attribute name="key" type="xsd:string" use="required" />
     </xsd:complexType>
 
     <xsd:complexType name="login_throttling">


src/Symfony/Bundle/SecurityBundle/Resources/config/schema/security-1.0.xsd
```diff
@@ -317,6 +317,74 @@
...
+        <xsd:sequence>
+            <xsd:element name="oidc-user-info" type="oidc_user_info"></xsd:element>
+        </xsd:sequence>
+    </xsd:complexType>
+
+    <xsd:complexType name="oidc_user_info" abstract="true">
+        <xsd:attribute name="claim" type="xsd:string" />
+    </xsd:complexType>
+
+    <xsd:complexType name="oidc_user_info_generic" abstract="true">
```
nicolas-grekas: abstract really?

src/Symfony/Bundle/SecurityBundle/Resources/config/schema/security-1.0.xsd
```diff
@@ -317,6 +317,74 @@
...
+        <xsd:complexContent>
+            <xsd:extension base="access_token">
```
nicolas-grekas: oidc_user_info? (same below)

src/Symfony/Bundle/SecurityBundle/Resources/config/schema/security-1.0.xsd
```diff
@@ -317,6 +317,74 @@
...
+                </xsd:choice>
+            </xsd:extension>
+        </xsd:complexContent>
+    </xsd:complexType>
+
+    <xsd:complexType name="oidc_client">
+        <xsd:attribute name="base-uri" type="xsd:anyURI" />
+    </xsd:complexType>
+
+    <xsd:complexType name="oidc">
```
nicolas-grekas, line 337: where is this type used?

aegypius:
Thanks @nicolas-grekas for the feedback, I hope the PR is in a better shape now !

nicolas-grekas (dismissed):

aegypius:
Sorry @nicolas-grekas, access token was not usable as-is, I merged defintions to be a little more loose but to make sure every case can be achieved

src/Symfony/Bundle/SecurityBundle/Resources/config/schema/security-1.0.xsd
```diff
@@ -309,12 +309,49 @@
     <xsd:complexType name="access_token">
         <xsd:choice minOccurs="0" maxOccurs="unbounded">
             <xsd:element name="token-extractor" type="xsd:string" />
+            <xsd:element name="token-handler" type="oidc_token_handler" />
         </xsd:choice>
-        <xsd:attribute name="token-handler" type="xsd:string" />
```
nicolas-grekas, line 313: I missed this: this breaks BC with previous xml configs, isn't it?

aegypius: I added it back on line 318 but I can move it up to avoid confusion

nicolas-grekas: yes please, let the diff be minimal

src/Symfony/Bundle/SecurityBundle/Resources/config/schema/security-1.0.xsd
```diff
@@ -317,6 +318,42 @@
...
+    <xsd:complexType name="oidc_user_info">
+        <xsd:choice minOccurs="0" maxOccurs="1">
+            <xsd:element name="client" type="oidc_client"></xsd:element>
+        </xsd:choice>
+        <xsd:attribute name="claim" type="xsd:string" />
+        <xsd:attribute name="client" type="xsd:string" />
+    </xsd:complexType>
+
+    <xsd:complexType name="oidc_client">
+        <xsd:attribute name="base-uri" type="xsd:anyURI" />
```
nicolas-grekas: Looking at the code, this should accept any options allowed by HttpClient (see `http_client_scope_options` in `symfony-1.0.xd`)

BUT

I'm wondering why we allow creating an HttpClient outside of the existing `framework.http_client` config tree.
The code in `OidcUserInfoTokenHandlerFactory` doesn't reuse the `http_client` service, which means this cannot benefit from shared DNS cache for example, nor from retries, nor scoped clients, etc.

Also, why is the `client` option required in the config? Can't we remove all this code to just accept a string here: the name of one of the client configured in `framework.http_client`, and default to the `http_client` service?

This question is for @vincentchalamon I suppose.

vincentchalamon: The `client` option can accept 2 value types:
- a client configured in `framework.http_client`
- or HTTPClient arguments to dynamically create it without passing by `framework.http_client`

Cf. https://github.com/symfony/symfony-docs/pull/17463/files

> The code in OidcUserInfoTokenHandlerFactory doesn't reuse the http_client service, which means this cannot benefit from shared DNS cache for example, nor from retries, nor scoped clients, etc.

Indeed, it should be better to remove the option allowing to dynamically create an HTTPClient. I can do it ASAP in 6.3 before it's stable, if you're ok with it. Doc update will follow just after.

nicolas-grekas: yes please, let's remove this, it's too clumsy IMHO :)

vincentchalamon: @nicolas-grekas https://github.com/symfony/symfony/pull/50470

nicolas-grekas: @aegypius can you please update the PR to account for #50470?

aegypius: @nicolas-grekas @vincentchalamon PR updated !

nicolas-grekas:
Thank you @aegypius.

PR #48992, opened 2023-01-16
[HttpKernel] Introduce pinnable value resolvers with `#[ValueResolver]` and `#[AsPinnedValueResolver]`

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #48927
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/17763

Introducing a new `ValueResolver` attribute, which allows to
- “pin” a value resolver to an argument, meaning only said resolver will be called
- prevent a resolver to be called for an argument

Every existing resolver-related attribute (`MapEntity`, `CurrentUser`…) now extends `ValueResolver`.

Each `controller.argument_value_resolver` tag is added a `name` attribute, which is the resolver’s FQCN. This is the first argument `ValueResolver` expects.

A new  `AsPinnedValueResolver` attribute is added for autoconfiguration, adding the `controller.pinned_value_resolver` tag. Such resolvers can only be “pinned”, meaning they won’t ever be called for an argument missing the `ValueResolver` attribute.

PR #50458, opened 2023-05-28
[HttpKernel] Fix default value ignored with pinned resolvers

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Since #48992 the default value is ignored when, for example, `#[MapEntity]` is used:

```php
#[Route('/')]
#[Route('/{someId}')]
public function index(#[MapEntity(id: 'someId')] ?Post $post): Response
{
    // ...
}
```

Before, `$post` would be `null` when making a request to `/`, now an exception is thrown:

```
Controller "App\Controller\TestController::index" requires that you provide a value for the "$post" argument.
Either the argument is nullable and no null value has been provided,
no default value has been provided or there is a non-optional argument after this one.
```

Since I can't think of a valid case when one would want to ignore the default value, I'd suggest always adding the `DefaultValueResolver` to the list when a pinned resolver is used.

### DIFF
--- a/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver.php
+++ b/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver.php
@@ -71,7 +71,10 @@ final class ArgumentResolver implements ArgumentResolverInterface
                         throw new ResolverNotFoundException($resolverName, $this->namedResolvers instanceof ServiceProviderInterface ? array_keys($this->namedResolvers->getProvidedServices()) : []);
                     }
 
-                    $argumentValueResolvers = [$this->namedResolvers->get($resolverName)];
+                    $argumentValueResolvers = [
+                        $this->namedResolvers->get($resolverName),
+                        new DefaultValueResolver(),
+                    ];
                 }
             }
 
--- a/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php
+++ b/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php
@@ -31,12 +31,8 @@ final class QueryParameterValueResolver implements ValueResolverInterface
 
         $name = $attribute->name ?? $argument->getName();
         if (!$request->query->has($name)) {
-            if ($argument->hasDefaultValue()) {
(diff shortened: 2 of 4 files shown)


nicolas-grekas:
Can you please revert #50340 in the patch also please? Your approach makes it unneeded.

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver.php
```diff
@@ -72,6 +72,10 @@ public function getArguments(Request $request, callable $controller, \Reflection
                     }
 
                     $argumentValueResolvers = [$this->namedResolvers->get($resolverName)];
+
+                    if ($this->namedResolvers->has(DefaultValueResolver::class) && !$argumentValueResolvers[0] instanceof DefaultValueResolver) {
```
nicolas-grekas: what about this instead?
```suggestion
                    $argumentValueResolvers = [
                        $this->namedResolvers->get($resolverName),
                        new DefaultValueResolver(),
                    ];
```

HypeMC: I think you would still need to check that `$this->namedResolvers->get($resolverName)` isn't an instance of `DefaultValueResolver`, but other then that I can update it.

HypeMC:
> Can you please revert #50340 in the patch also please? Your approach makes it unneeded.

Done

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver.php
nicolas-grekas: no need to check for that: if it's the default resolver, then the first will do it's job

HypeMC: You're right, if some pinned the `DefaultValueResolver` then it's unlikely that there's no default value. Done.

nicolas-grekas:
Thank you @HypeMC.

MatTheCat:
There is no default value in your example though 🤔

PR #50459, opened 2023-05-28
[Form] Add constraints_from_* options

tasiot:
Apply constraints to field from an entity not mapped by the form data_class.

| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | no
| License       | MIT

**Use-cases:**
- DTO form like user registration with many entities concerned (user / address)

**Current usage:**
You set constraints in your entities then create a DTO but you must duplicate your constraints in your DTO object to valide your fields.

**New usage:**
You set constraints in your entities then create a DTO and link your DTO form fields to your entities properties.

**Usage:**

Add `constraints_from_entity` option on any form field (child of `FormType::class`).
If the entity related property name is different that the field name, you must provide `constraints_from_property` to declare the property name.

Works on (un)mapped field and (un)set `data_class` form.

```php
<?php
// App/Form/UserRegistrationType.php

namespace App\Form;

use App\Entity\User;
use App\Entity\Address;
use App\Form\Model\UserRegistration;
use Symfony\Component\Form\AbstractType;
use Symfony\Component\Form\Extension\Core\Type\TextType;
use Symfony\Component\Form\FormBuilderInterface;
use Symfony\Component\OptionsResolver\OptionsResolver;

class UserRegistrationType extends AbstractType
{
    public function buildForm(FormBuilderInterface $builder, array $options): void
    {
        $builder
            ->add('username', TextType::class, [
                'label' => 'Your username',
                'constraints_from_entity' => User::class,
                // Not required if property name is same as the field name
                // 'constraints_from_property' => 'username' // Constraints from User::username
            ])
            ->add('zipCode', TextType::class, [
                'label' => 'Your zip/postal code',
                'constraints_from_entity' => Address::class,
                'constraints_from_property' => 'postalCode' // Constraints from Address::postalCode
            ]);
    }

    public function configureOptions(OptionsResolver $resolver): void
    {
        $resolver->setDefaults([
            'data_class' => UserRegistration::class
        ]);
    }
}
```

**Notes:**
It's my first pull request, I hope I done all required things ;)

### DIFF
--- a/src/Symfony/Component/Form/CHANGELOG.md
+++ b/src/Symfony/Component/Form/CHANGELOG.md
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+6.4
+---
+ * Add `constraints_from_entity` and `constraints_from_property` option to `FormType`
+
 6.3
 ---
 
--- a/src/Symfony/Component/Form/Extension/Validator/EventListener/ConstraintsFromListener.php
+++ b/src/Symfony/Component/Form/Extension/Validator/EventListener/ConstraintsFromListener.php
@@ -1,0 +1,58 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Form\Extension\Validator\EventListener;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Form\FormError;
+use Symfony\Component\Form\FormEvent;
+use Symfony\Component\Form\FormEvents;
+use Symfony\Component\Validator\ConstraintViolationInterface;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
+
+class ConstraintsFromListener implements EventSubscriberInterface
+{
+    private ValidatorInterface $validator;
+
+    public static function getSubscribedEvents(): array
+    {
+        return [FormEvents::POST_SUBMIT => 'validateConstraints'];
+    }
+
+    public function __construct(ValidatorInterface $validator)
+    {
+        $this->validator = $validator;
+    }
+
+    public function validateConstraints(FormEvent $event): void
+    {
+        $form = $event->getForm();
+
+        $entity = $form->getConfig()->getOption('constraints_from_entity');
+
+        if (null !== $entity) {
+            $property = $form->getConfig()->getOption('constraints_from_property') ?? $form->getName();
+
+            $violations = $this->validator->validatePropertyValue($entity, $property, $event->getData());
+            /** @var ConstraintViolationInterface $violation */
+            foreach ($violations as $violation) {
+                $form->addError(
+                    new FormError(
+                        $violation->getMessage(),
+                        $violation->getMessageTemplate(),
+                        $violation->getParameters(),
+                        $violation->getPlural()
+                    )
+                );
+            }
+        }
+    }
+}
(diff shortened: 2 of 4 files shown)


tasiot:
I think the target branch is 6.4 but she's behind the 6.3 so I didn't know which branch I must to target

tasiot:
I think the target branch is 6.4 but she's behind the 6.3 so I didn't know which branch I must to target

tasiot:
I don't understand why the high-deps and low-deps unit tests fail…

Can anyone help me?

nicolas-grekas:
6.4 is the right target, we merge 6.3 into 6.4 regularly so it will catch up eventually.

src/Symfony/Component/Form/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ CHANGELOG
  * Don't render seconds for HTML5 date pickers unless "with_seconds" is explicitly set
  * Add a `placeholder_attr` option to `ChoiceType`
  * Deprecate not configuring the "widget" option of date/time form types, it will default to "single_text" in v7
+ * Add `constraints_from_entity` and `constraints_from_property` option to `FormType`
```
nicolas-grekas: for a new 6.4 section

tasiot:
The job low-deps fails because it cannot find the trait `Symfony\Component\Form\Tests\VersionAwareTest` in `Symfony\Component\Form\Test\FormPerformanceTestCase`.
Can anyone help me? I don't understand how to pass this test because I don't think I have any relation with these files…

ro0NL:
why does your entity need constraint still if the DTO is in valid state already?

tasiot:
> 

Maybe the term "DTO" is not right, a form model will be more appropriate? An object who contains some properties to create some entities…

Sometimes, you can have many forms… A form with `'data_class' => YourEntity::class` and another one with `'data_class' => YourDTO::class`.

When you make a form to create/edit an entity, it's recommanded to set constraints in this entity instead of each form field, isn't it?
Why duplicate these constraints in your second form (with DTO object)?

With the option `constraints_from_entity` another form (DTO) or an unmapped field can also apply these constraints without needs to duplicate the entity field constraints in each form field.

Consider these entities (simplified):
| Topic | 
| --- |
| name _(string)_ |
| category _(TopicCategory)_ |

| TopicCategory |
| --- |
| name _(string)_ |

| TopicMessage |
| --- |
| topic _(Topic)_ |
| message _(text)_ |

You can easily imagine a form to create the first message which asks all theses entity (a category, a name and a message) (=a DTO with these properties) and another form to create the next messages which only asks a message (=TopicMessage Entity).
Instead of duplicate your entities constraints to your DTO form, you only need to add option constraints_from_entity as below:

```php
// With a DTO (data_class => App\Form\Model\FirstTopicMessage::class)
$builder
->add('category', EntityType::class, [
    'class' => TopicCategory::class
])
->add('name', TextType::class, [
    'constraints_from_entity' => Topic::class
])
-> add('message', TextareaType::class, [
    'constraints_from_entity' => TopicMessage::class
])
```

_OR_

```php
// With unmapped fields (data_class => App\Entity\TopicMessage::class)
$builder
->add('category', EntityType::class, [
    'class' => TopicCategory::class,
    'mapped' => false,
])
->add('name', TextType::class, [
    'constraints_from_entity' => Topic::class,
    'mapped' => false,
])
-> add('message', TextareaType::class)
```

ro0NL:
my personal opionion is constraints should be but on the form type (`constraints` options), regardless of the underlying model

adding 1 entity which affects N forms sounds like a footgun to me :)

nicolas-grekas:
Closing this one. Thanks for the patch, and thanks for your patience with how long it took to get back to you.

On the shape, the listener adds the violations with `$form->addError()`, so it bypasses the ViolationMapper: `error_mapping` does not apply and the resulting `FormError` carries no cause, no constraint and no invalid value. There is also no way to pass validation groups, so it always validates against `Default` while the constraints on the same field may not, and the subscriber is registered in `FormTypeValidatorExtension::buildForm()`, so every field of every form allocates one for an option that is almost always null.

The need behind it is real, but pointing a form option at another class's metadata puts it in the wrong layer. Declaring the constraints on the form model, or sharing them through a trait between the model and the entity, covers the case without new options on `FormType`.

PR #50467, opened 2023-05-30
[Command] Add default name

94noni:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes (kind of)
| New feature?  | yes (TODO)
| Deprecations? | yes (TODO)
| Tickets       | Fix https://github.com/symfony/symfony-docs/pull/17782
| License       | MIT
| Doc PR        | TODO

Regarding the sf doc PR, I propose to default to a name here before going full name on sfv7
Other option is, like proposed in the PR doc, adding more context in the exception message but to me it is not the best option

If accepted, will document doc and various upgrade/changelog files

### DIFF
--- a/src/Symfony/Component/Console/Command/Command.php
+++ b/src/Symfony/Component/Console/Command/Command.php
@@ -128,6 +128,9 @@ class Command
 
         if (null !== $name) {
             $this->setName($name);
+        } else {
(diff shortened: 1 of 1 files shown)


chalasr:
Thanks for proposing, but I don't think commands need an automatically guessed default name a that would be misleading. Even if we did, it wouldn't be the command FQCN as it's way too painful to type. Throwing is the most appropriate thing to do in such situation.

94noni:
Oki @chalasr i will report on the issue if we can improve documentation or the message exception
Thx

PR #50470, opened 2023-05-30
[SecurityBundle] Fix configuring OIDC user info token handler client

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/18354

Cf. https://github.com/symfony/symfony/pull/50453#discussion_r1210028968

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcUserInfoTokenHandlerFactory.php
+++ b/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcUserInfoTokenHandlerFactory.php
@@ -15,7 +15,7 @@ use Symfony\Component\Config\Definition\Builder\NodeBuilder;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
-use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
 
 /**
  * Configures a token handler for an OIDC server.
@@ -26,24 +26,20 @@ class OidcUserInfoTokenHandlerFactory implements TokenHandlerFactoryInterface
 {
     public function create(ContainerBuilder $container, string $id, array|string $config): void
     {
-        $tokenHandlerDefinition = $container->setDefinition($id, new ChildDefinition('security.access_token_handler.oidc_user_info'));
-        $tokenHandlerDefinition->replaceArgument(2, $config['claim']);
+        $clientDefinition = (new ChildDefinition('security.access_token_handler.oidc_user_info.http_client'))
+            ->replaceArgument(0, ['base_uri' => $config['base_uri']]);
 
-        // Create the client service
-        if (!isset($config['client']['id'])) {
-            $clientDefinitionId = 'http_client.security.access_token_handler.oidc_user_info';
-            if (!ContainerBuilder::willBeAvailable('symfony/http-client', HttpClient::class, ['symfony/security-bundle'])) {
-                $container->register($clientDefinitionId, 'stdClass')
-                    ->addError('You cannot use the "oidc_user_info" token handler since the HttpClient component is not installed. Try running "composer require symfony/http-client".');
-            } else {
-                $container->register($clientDefinitionId, HttpClient::class)
-                    ->setFactory([HttpClient::class, 'create'])
-                    ->setArguments([$config['client']])
-                    ->addTag('http_client.client');
-            }
+        if (isset($config['client'])) {
+            $clientDefinition->setFactory([new Reference($config['client']), 'withOptions']);
+        } elseif (!ContainerBuilder::willBeAvailable('symfony/http-client', HttpClientInterface::class, ['symfony/security-bundle'])) {
+            $clientDefinition
+                ->setFactory(null)
+                ->addError('You cannot use the "oidc_user_info" token handler since the HttpClient component is not installed. Try running "composer require symfony/http-client".');
         }
 
-        $tokenHandlerDefinition->replaceArgument(0, new Reference($config['client']['id'] ?? $clientDefinitionId));
+        $container->setDefinition($id, new ChildDefinition('security.access_token_handler.oidc_user_info'))
+            ->replaceArgument(0, $clientDefinition)
+            ->replaceArgument(2, $config['claim']);
     }
 
     public function getKey(): string
@@ -56,19 +52,23 @@ class OidcUserInfoTokenHandlerFactory implements TokenHandlerFactoryInterface
         $node
             ->arrayNode($this->getKey())
                 ->fixXmlConfig($this->getKey())
+                ->beforeNormalization()
+                    ->ifString()
+                    ->then(static fn ($v) => ['claim' => 'sub', 'base_uri' => $v])
+                ->end()
                 ->children()
-                    ->scalarNode('claim')
-                        ->info('Claim which contains the user identifier (e.g.: sub, email..).')
-                        ->defaultValue('sub')
-                    ->end()
-                    ->arrayNode('client')
-                        ->info('HttpClient to call the OIDC server.')
+                    ->scalarNode('base_uri')
+                        ->info('Base URI of the userinfo endpoint on the OIDC server.')
                         ->isRequired()
-                        ->beforeNormalization()
-                            ->ifString()
-                            ->then(static function ($v): array { return ['id' => $v]; })
-                        ->end()
-                        ->prototype('scalar')->end()
+                        ->cannotBeEmpty()
+                    ->end()
+                    ->scalarNode('claim')
+                        ->info('Claim which contains the user identifier (e.g. sub, email, etc.).')
+                        ->defaultValue('sub')
+                        ->cannotBeEmpty()
+                    ->end()
+                    ->scalarNode('client')
+                        ->info('HttpClient service id to use to call the OIDC server.')
                     ->end()
                 ->end()
             ->end()
--- a/src/Symfony/Bundle/SecurityBundle/Tests/DependencyInjection/Security/Factory/AccessTokenFactoryTest.php
+++ b/src/Symfony/Bundle/SecurityBundle/Tests/DependencyInjection/Security/Factory/AccessTokenFactoryTest.php
@@ -18,7 +18,9 @@ use Symfony\Bundle\SecurityBundle\DependencyInjection\Security\AccessToken\Servi
 use Symfony\Bundle\SecurityBundle\DependencyInjection\Security\Factory\AccessTokenFactory;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
+use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
 
 class AccessTokenFactoryTest extends TestCase
 {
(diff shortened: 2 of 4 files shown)


src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcUserInfoTokenHandlerFactory.php
```diff
@@ -61,14 +45,9 @@ public function addConfiguration(NodeBuilder $node): void
                         ->info('Claim which contains the user identifier (e.g.: sub, email..).')
                         ->defaultValue('sub')
                     ->end()
-                    ->arrayNode('client')
+                    ->scalarNode('client')
                         ->info('HttpClient to call the OIDC server.')
                         ->isRequired()
```
nicolas-grekas: why do we keep this? can't we default to `http_client`?

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcUserInfoTokenHandlerFactory.php
```diff
@@ -27,23 +26,8 @@ class OidcUserInfoTokenHandlerFactory implements TokenHandlerFactoryInterface
...
         $tokenHandlerDefinition = $container->setDefinition($id, new ChildDefinition('security.access_token_handler.oidc_user_info'));
+        $tokenHandlerDefinition->replaceArgument(0, new Reference($config['client']));
         $tokenHandlerDefinition->replaceArgument(2, $config['claim']);
-
-        // Create the client service
-        if (!isset($config['client']['id'])) {
-            $clientDefinitionId = 'http_client.security.access_token_handler.oidc_user_info';
-            if (!ContainerBuilder::willBeAvailable('symfony/http-client', HttpClient::class, ['symfony/security-bundle'])) {
-                $container->register($clientDefinitionId, 'stdClass')
-                    ->addError('You cannot use the "oidc_user_info" token handler since the HttpClient component is not installed. Try running "composer require symfony/http-client".');
```
nicolas-grekas, line 37: we might still need to register this special def when the client is not installed, isn't it?

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcUserInfoTokenHandlerFactory.php
vincentchalamon: Cause we need the base_uri of the OIDC server:
```yaml
# config/packages/framework.yaml
framework:
    http_client:
        scoped_clients:
            oidc.client:
                base_uri: 'https://www.example.com/realms/demo/protocol/openid-connect/userinfo'

# config/packages/security.yaml
security:
    firewalls:
        main:
            access_token:
                token_handler:
                    oidc_user_info:
                        client: oidc.client
```

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcUserInfoTokenHandlerFactory.php
vincentchalamon: `client` option receives an http_client service id, so I guess the user installed symfony/http-client to declare it. Do we need to double-check it here?

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcUserInfoTokenHandlerFactory.php
nicolas-grekas: are there any other options that might be useful to configure?
because if not, we might just need `oidc_user_info: <base-uri-as-string>`

To do so, I'd suggest defining an abstract service, let's name it `security.access_token_handler.oidc_user_info.http_client`, defined as such:

```php
->set('security.access_token_handler.oidc_user_info.http_client', HttpClientInterface::class)
	->abstract()
	->factory([service('http_client'), 'withOptions'])
```

and in the OidcUserInfoTokenHandlerFactory, we'd create a ChildDefinition of this service and set it's argument to `['base_uri' => $config['base_uri']]`.

Does this make sense to you?

vincentchalamon: `oidc_user_info` can also accept a `claim` optional parameter:
```yaml
oidc_user_info:
  claim: email # default: sub
  client: oidc.client # required
```

I can allow `oidc_user_info: <base-uri>` and normalize it, but I find this syntax more confusing than setting the `oidc_user_info.client` parameter. But this is just a question of point of view. WDYT?

Anyway, if a user wants to change the `claim`, we still need a parameter to set the base_uri (or a custom client).

nicolas-grekas: I think the DX will be much better with the changes I suggested.
We can accept both:
```yaml
oidc_user_info: 'https://...'
# same as:
oidc_user_info:
  claim: sub
  base_uri: 'https://...'
```

vincentchalamon: We should keep the `client` option in case the dev wants to manually configure its client externally (e.g.: `framework.http_client`, or any service manually created).

In your approach, the `base_uri` option would be used to dynamically create the client (and set the `client` option), and the `client` option would receive a HttpClient service id:
```yaml
# HttpClient dynamically created with "base_uri" only
oidc_user_info: https://...

# HttpClient dynamically created with "base_uri" only
oidc_user_info:
  claim: sub
  base_uri: https://... # "base_uri" is not compatible with "client" option

# HttpClient manually created with custom options
oidc_user_info:
  claim: sub
  client: oidc.client
```

But I think it's making this configuration and DX more complex... IMHO, I would prefer to set the HttpClient options in the `client` option like it was earlier (but this time, using the abstract definition you suggested):
```yaml
# HttpClient manually created with custom options
oidc_user_info:
  claim: sub
  client:
    base_uri: https://...
    # other HttpClient options

# HttpClient dynamically created with "base_uri" only
oidc_user_info: https://...
```

nicolas-grekas: I'm fine adding another option to set the service id if that's something that'd be useful, but I don't think allowing other http_client options is desired (as explained in https://github.com/symfony/symfony/pull/50453#discussion_r1210028968)
But being able to set only the base_uri looks the best DX to me. I can do the code if you're not comfortable with it of course. Let me know!

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcUserInfoTokenHandlerFactory.php
vincentchalamon: Solved with the addition of `base_uri` option

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcUserInfoTokenHandlerFactory.php
vincentchalamon: Fixed in my last commit

src/Symfony/Bundle/SecurityBundle/Tests/DependencyInjection/Security/Factory/AccessTokenFactoryTest.php
```diff
@@ -103,7 +120,19 @@ public function testOidcUserInfoTokenHandlerConfigurationWithClientCreation()
...
+
+        $expected = [
+            'index_0' => (new ChildDefinition('security.access_token_handler.oidc_user_info.http_client'))
+                ->replaceArgument(0, ['base_uri' => 'https://www.example.com/realms/demo/protocol/openid-connect/userinfo']),
+            'index_2' => 'sub',
+        ];
+        $this->assertEquals($expected, $container->getDefinition('security.access_token_handler.firewall1')->getArguments());
+    }
+
+    public function getOidcUserInfoConfiguration(): iterable
```
mtarld: Must be static in order to be compatible with the next PHPUnit version

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcUserInfoTokenHandlerFactory.php
```diff
@@ -56,19 +52,21 @@ public function addConfiguration(NodeBuilder $node): void
         $node
             ->arrayNode($this->getKey())
                 ->fixXmlConfig($this->getKey())
+                ->beforeNormalization()
+                    ->ifString()
+                    ->then(static fn ($v) => ['claim' => 'sub', 'base_uri' => $v])
+                ->end()
                 ->children()
+                    ->scalarNode('base_uri')
+                        ->info('Base URI of the OIDC server.')
```
vincentchalamon:
```suggestion
                        ->info('Base URI of the userinfo endpoint on the OIDC server.')
```

nicolas-grekas (approved):
I tweaked the implementation a bit, see
https://github.com/symfony/symfony/compare/8e7452011403b4bbe67a4d2abd12dde3df95f0c3..8b7fdc108237bd8e88c59aa0282aa6d36d9ae7ad

TL;DR, the `base_uri` option is required, and the `client` option can be used optionally to configure any other options.

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcUserInfoTokenHandlerFactory.php
```diff
@@ -56,19 +52,21 @@ public function addConfiguration(NodeBuilder $node): void
...
+                        ->isRequired()
+                    ->end()
                     ->scalarNode('claim')
-                        ->info('Claim which contains the user identifier (e.g.: sub, email..).')
+                        ->info('Claim which contains the user identifier (e.g. sub, email, etc.).')
                         ->defaultValue('sub')
```
mtarld, line 67: You can add `isRequired` and `canntBeEmpty` as well IMHO

nicolas-grekas: cannotBeEmpty why not, can you take care of @mtarld's comments and squash @vincentchalamon please?

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcUserInfoTokenHandlerFactory.php
```diff
@@ -60,10 +60,13 @@ public function addConfiguration(NodeBuilder $node): void
                     ->scalarNode('base_uri')
                         ->info('Base URI of the userinfo endpoint on the OIDC server.')
                         ->isRequired()
+                        ->cannotBeEmpty()
                     ->end()
                     ->scalarNode('claim')
                         ->info('Claim which contains the user identifier (e.g. sub, email, etc.).')
                         ->defaultValue('sub')
+                        ->isRequired()
```
nicolas-grekas: does this one needs to be required if it has a default value?

vincentchalamon: Nop, removed in a squashed commit. Tests revealed the bug too.
Thanks for noticing :wink:

nicolas-grekas:
Thank you @vincentchalamon.

PR #50472, opened 2023-05-30
[Serializer] test: discriminator without matching property

soyuka:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| License       | MIT

https://github.com/symfony/symfony/pull/50193 breaks a test on API Platform w/ Symfony  6.2.11 where the discriminator has no matching property. 

Fixture: https://github.com/api-platform/core/blob/main/tests/Fixtures/TestBundle/Entity/AbstractDummy.php
Feature: https://github.com/api-platform/core/blob/main/features/main/crud_abstract.feature#L108-L132

This adds a failing test case, but I'm unsure whether our test case misuses the discriminator feature (does it require a property matching the discriminator ?) or if the #50193 fix breaks.

ping @HypeMC wdyt can you help me on that?

### DIFF
--- a/src/Symfony/Component/Serializer/Tests/Fixtures/AbstractWithDiscriminator.php
+++ b/src/Symfony/Component/Serializer/Tests/Fixtures/AbstractWithDiscriminator.php
@@ -1,0 +1,28 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Serializer\Tests\Fixtures;
+
+use Symfony\Component\Serializer\Annotation\DiscriminatorMap;
+
+#[DiscriminatorMap(typeProperty: 'discr', mapping: ['concrete' => ConcreteWithDiscriminator::class])]
+abstract class AbstractWithDiscriminator
+{
+    /**
+     * @var int The id
+     */
+    public ?int $id = null;
+
+    /**
+     * @var string The dummy name
+     */
+    public string $name;
+}
--- a/src/Symfony/Component/Serializer/Tests/Fixtures/ConcreteWithDiscriminator.php
+++ b/src/Symfony/Component/Serializer/Tests/Fixtures/ConcreteWithDiscriminator.php
@@ -1,0 +1,20 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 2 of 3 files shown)


HypeMC:
Hi, here's what I was able to figure out: unfortunately your test case doesn't really illustrate the real issue since `$loaderMock` doesn't return the right data, it should look something like this:

```php
        $loaderMock = new class() implements ClassMetadataFactoryInterface {
            public function getMetadataFor($value): ClassMetadataInterface
            {
                if (ConcreteWithDiscriminator::class === $value) {
                    return new ClassMetadata(ConcreteWithDiscriminator::class);
                }

                throw new InvalidArgumentException();
            }

            public function hasMetadataFor($value): bool
            {
                return ConcreteWithDiscriminator::class === $value;
            }
        };
```

I know this because I've tested the following cases on a clean Symfony installation:

```php
$serializer->denormalize(['id' => 0, 'instance' => 'test', 'name' => 'test'], ConcreteWithDiscriminator::class, 'any');
$serializer->denormalize(['id' => 0, 'instance' => 'test', 'name' => 'test', 'discr' => 'concrete'], AbstractWithDiscriminator::class, 'any');

$serializer->denormalize(['id' => 0, 'instance' => 'test', 'name' => 'test'], AbstractWithDiscriminator::class, 'any');
```

Both before and after #50193 the first two cases work, while the third one throws an exception:
```
  Type property "discr" not found for the abstract object "App\DTO\AbstractWithDiscriminator".
```

With the API Platform test case things are a little different because the [`ItemNormalizer` adds `AbstractNormalizer::OBJECT_TO_POPULATE` to the context](https://github.com/api-platform/core/blob/039ba865ef634e79a757e49c2ff1ecacad2b7aa0/src/JsonLd/Serializer/ItemNormalizer.php#L147), so your test case would look like this:

```php
$this->serializer->denormalize(
    ['id' => 0, 'instance' => 'test', 'name' => 'test'],
    AbstractWithDiscriminator::class,
    'any',
    [AbstractNormalizer::OBJECT_TO_POPULATE => $object = new ConcreteWithDiscriminator()],
);
```

Again, for this case, both before and after #50193 an exception is thrown.

The only reason your test passed before is because the `AbstractItemNormalizer` [overrides the `instantiateObject()` method](https://github.com/api-platform/core/blob/039ba865ef634e79a757e49c2ff1ecacad2b7aa0/src/Serializer/AbstractItemNormalizer.php#L264-L268), but since `getMappedClass()` is now called before `instantiateObject()`, your test breaks.

That being said, I do think this is a bug in Symfony since it would make sense for the last example to pass, at least in my opinion, and as far as I can tell the fix is really simple, so I'll try to open a PR.

nicolas-grekas:
Thanks for the report and the test case to illustrate the issue!
Closing in favor of #50480, please have a look there :pray:

PR #50475, opened 2023-05-30
[FrameworkBundle] Prevent `cache:clear` to lose files on subsequent runs

### LINKED ISSUE #50474: Subsequent calls to `cache:clear` delete warmed up files

5.4.24,6.2.11

### Description

#50253 forces the `cache:clear` to always run in the `cache_dir`, even when both `build_dir` and `cache_dir` are the same directory.

This leads to some files being deleted if the container cache is not fresh (which is always the case if the `cache_dir` directory was already existing and warmed up).

Since the optional CacheWarmers are run in the `cache_dir`, their artefacts are lost once the `cache_dir` is replaced with  the warmup directory.

### How to reproduce

```shell
s new --dir symfony_cache_dir_problem
cd ./symfony_cache_dir_problem
composer install -q
# See that router cache is warmed up
composer install -q
# See that router cache is missing
```

### Possible Solution

_No response_

### Additional Context

_No response_

okhoshi:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50474 
| License       | MIT
| Doc PR        | N/A

See #50474 for the details and reproduction information.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Command/CacheClearCommand.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Command/CacheClearCommand.php
@@ -137,7 +137,7 @@ EOF
                 if ($output->isVerbose()) {
                     $io->comment('Warming up optional cache...');
                 }
-                $this->warmupOptionals($realCacheDir);
+                $this->warmupOptionals($realCacheDir, $realBuildDir);
             }
         } else {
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Thank you @Okhoshi.

PR #50476, opened 2023-05-30
[FrameworkBundle] remove support for preloading ESM using headers

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Sorry to be that late on this one.
It looks like preloading ESM using HTTP headers (and so using 103 Early Hints) doesn't work, at least for now: https://github.com/WICG/import-maps/issues/273#issuecomment-1097276232

ImportMap must be defined before downloading any ESM, and by definition HTTP headers are sent before importmap definition (it's currently not possible to define importmap using headers).

I don't know why I didn't caught this during my previous testing session, but now I get this error in the console when using this feature (which, according to the previously mentioned issue, is expected):

```
An import map is added after module script load was triggered.
Uncaught TypeError: Failed to resolve module specifier "app". Relative references must start with either "/", "./", or "../".
```

This patch simply removes this feature, which is entirely non-functional and misleading for the end user.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php
@@ -14,7 +14,6 @@ namespace Symfony\Bundle\FrameworkBundle\Controller;
 use Psr\Container\ContainerInterface;
 use Psr\Link\EvolvableLinkInterface;
 use Psr\Link\LinkInterface;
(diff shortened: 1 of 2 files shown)


weaverryan (approved):

nicolas-grekas:
Thank you @dunglas.

Issue #50441, opened 2023-05-26
[Security] Support JWE on OidcTokenHandler


vincentchalamon:
### Description

The [OIDC Core Specification](https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation) recommends to decrypt the ID Token if it's encrypted:
> If the ID Token is encrypted, decrypt it using the keys and algorithms that the Client specified during Registration that the OP was to use to encrypt the ID Token. If encryption was negotiated with the OP at Registration time and the ID Token is not encrypted, the RP SHOULD reject it.

AFAIK, JWE are not often used. But as Symfony supports OIDC feature (experimentally), it could be interesting to support JWE. It concerns only OidcTokenHandler.

Question: should it be a new feature or a bugfix (as it's part of the Core Specification)?

### Example

To decrypt a JWE, a public JWK is required. Hopefully, it is available on `/certs` endpoint on OIDC server, so we can import it dynamically (cf. https://github.com/symfony/symfony/issues/50434).

```yaml
security:
  firewalls:
    main:
      oidc:
        # add new "encryption" option
        encryption:
          # proposal 1: same approach than "signature.key" option, the JWK is set here as JSON string
          key: '{"kid": "...","kty": "RSA","alg": "RS256","use": "sig","n": "3G..."}'
          # proposal 2: detect value is an URI, call it to import the JWK
          key: 'https://www.example.com/protocol/openid-connect/certs'
          # proposal 3: add "certs" option non combinable with "key" to import and set "key" option dynamically
          certs: 'https://www.example.com/protocol/openid-connect/certs'
```

chalasr:
I'd say it's a new feature given solving it implies to extend the configuration.  

> It concerns only OidcTokenHandler.

Note that pretty much everything may be encrypted in OIDC interactions, even [UserInfo](https://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse) success responses used by `OidcUserInfoTokenHandler`. Use cases are probably even less common though.

carsonbot:
Thank you for this suggestion.
There has not been a lot of activity here for a while. Would you still like to see this feature?

vincentchalamon:
Hey @carsonbot, let's keep this one open, I just didn't have time to focus on it yet.

carsonbot:
Thank you for this suggestion.
There has not been a lot of activity here for a while. Would you still like to see this feature?

vincentchalamon:
[...]

PR #50477, opened 2023-05-30
[Security] Add clock dependency to OidcTokenHandler

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

From "web-token/jwt-checker":
The parameter "$clock" will become mandatory in 4.0.0. Please set a valid PSR Clock implementation instead of "null".

Also fixing some other issues found meanwhile.

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
+++ b/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
@@ -13,7 +13,6 @@ namespace Symfony\Bundle\SecurityBundle\DependencyInjection\Security\AccessToken
 
 use Jose\Component\Core\Algorithm;
 use Jose\Component\Core\JWK;
-use Symfony\Bundle\SecurityBundle\DependencyInjection\Security\Factory\SignatureAlgorithmFactory;
 use Symfony\Component\Config\Definition\Builder\NodeBuilder;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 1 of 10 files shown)


src/Symfony/Component/Security/Http/AccessToken/Oidc/OidcTokenHandler.php
```diff
@@ -20,8 +20,9 @@
 use Jose\Component\Signature\JWSVerifier;
 use Jose\Component\Signature\Serializer\CompactSerializer;
 use Jose\Component\Signature\Serializer\JWSSerializerManager;
+use Psr\Clock\ClockInterface;
 use Psr\Log\LoggerInterface;
-use Symfony\Component\Clock\NativeClock;
```
nicolas-grekas, line 24: NativeClock should never be used directly since 6.3

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
```diff
@@ -28,26 +29,32 @@ class OidcTokenHandlerFactory implements TokenHandlerFactoryInterface
...
+        }
+        if (\in_array($config['algorithm'], ['ES256', 'ES384', 'ES512'], true)) {
+            $tokenHandlerDefinition->replaceArgument(0, new Reference('security.access_token_handler.oidc.signature.'.$config['algorithm']));
         } else {
-            $container->register('security.access_token_handler.oidc.signature', Algorithm::class)
-                ->setFactory([SignatureAlgorithmFactory::class, 'create'])
-                ->setArguments([$config['signature']['algorithm']]);
-            $container->register('security.access_token_handler.oidc.jwk', JWK::class)
-                ->setFactory([JWK::class, 'createFromJson'])
-                ->setArguments([$config['signature']['key']]);
```
nicolas-grekas, line 47: This fixes the wiring. Before this change, all firewalls would share a single jwk+algo definition, since varying key/algo were registered under the same id

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
```diff
@@ -69,18 +76,13 @@ public function addConfiguration(NodeBuilder $node): void
...
                     ->end()
-                    ->arrayNode('signature')
+                    ->scalarNode('algorithm')
+                        ->info('Algorithm used to sign the token.')
+                        ->isRequired()
+                    ->end()
+                    ->scalarNode('key')
+                        ->info('JSON-encoded JWK used to sign the token (must contain a "kty" key).')
                         ->isRequired()
-                        ->children()
```
nicolas-grekas, line 74: removing one unneeded nesting level in the config /cc @vincentchalamon for the doc

src/Symfony/Component/Security/Http/AccessToken/Oidc/OidcUserInfoTokenHandler.php
```diff
@@ -49,7 +49,7 @@ public function getUserBadgeFrom(string $accessToken): UserBadge
 
             // UserLoader argument can be overridden by a UserProvider on AccessTokenAuthenticator::authenticate
             return new UserBadge($claims[$this->claim], fn () => $this->createUser($claims), $claims);
-        } catch (\Throwable $e) {
+        } catch (\Exception $e) {
```
nicolas-grekas, line 52: we should never catch `Throwable` unless reasons which should be explicit/obvious

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
vincentchalamon: The `signature` level was useful in anticipation of https://github.com/symfony/symfony/issues/50441. But removing it is more MVP. Let's talk about it later on https://github.com/symfony/symfony/issues/50441

PR #50492, opened 2023-05-31
[HttpFoundation] add @throws to getPayload

### LINKED ISSUE #50490: [HttpFoundation] getPayload throws JsonException if request body is empty.

6.3.0

### Description

If I call `$request->getPayload()` it throws a `JsonException`.

What is the intended behavior?
In my opinion throwing an exception is fine, but `@throws Symfony\Component\HttpFoundation\Exception\JsonException` on the method would be nice.

### How to reproduce

Call getPayload on request object in a controller and make POST request with empty body.

### Possible Solution

Add `@throws Symfony\Component\HttpFoundation\Exception\JsonException`

### Additional Context

_No response_

wiseguy1394:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no
| Tickets       | Fix #50490
| License       | MIT

Fixes #50490

### DIFF
--- a/src/Symfony/Component/HttpFoundation/Request.php
+++ b/src/Symfony/Component/HttpFoundation/Request.php
@@ -1508,6 +1508,8 @@ class Request
 
     /**
      * Gets the decoded form or json request body.
+     * 
+     * @throws JsonException When the body cannot be decoded to an array
      */
     public function getPayload(): InputBag
     {


src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1508,6 +1508,8 @@ public function getContent(bool $asResource = false)
 
     /**
      * Gets the decoded form or json request body.
+     * 
+     * @throws JsonException
```
```suggestion
     * @throws JsonException When the body cannot be decoded to an array
```

nicolas-grekas (approved):

nicolas-grekas:
Thank you @wiseguy1394.

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1508,6 +1508,8 @@ public function getContent(bool $asResource = false)
 
     /**
      * Gets the decoded form or json request body.
+     * 
+     * @throws JsonException When the body cannot be decoded to an array
```
ro0NL, line 1512: should be `RequestExceptionInterface` IMHO

wiseguy1394: Why not exact JsonException?

ro0NL: @wiseguy1394 catching the general case seems more future proof

but no strong opionion, i dont believe in `@throws` bookkeeping anyway :)

nicolas-grekas:
Behavior changed in #50730 for empty contents.

PR #50493, opened 2023-05-31
[VarDumper] Use documentElement instead of body for JS flag

### LINKED ISSUE #50487: using var-dumper as standalone, expand or collapse not working

n/a

### Description

after upgrading var-dumper to 6.3, when using `dump()` or `dd()`, expand or collapse for object/array/string suddenly not working.

image

https://github.com/symfony/symfony/assets/6737021/240856c0-3a7e-40c5-89c3-505115154cea

### How to reproduce

- `composer init` then install `symfony/var-dumper` v6.3
- open `index.php` in browser
- expand or collapse not working

[composer.json.txt](https://github.com/symfony/symfony/files/11607410/composer.json.txt)
[index.php.txt](https://github.com/symfony/symfony/files/11607411/index.php.txt)

### Possible Solution

_No response_

### Additional Context

- browser: Chrome
- platform: Windows
- user agent: `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/113.0.0.0 Safari/537.36`

ohader:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50487
| License       | MIT
| Doc PR        | ---

Fixes: #50487

Using `document.documentElement` allows to use the dumper outside of a `<html><body>...</body></html>` scenario.

### DIFF
--- a/src/Symfony/Component/VarDumper/Dumper/HtmlDumper.php
+++ b/src/Symfony/Component/VarDumper/Dumper/HtmlDumper.php
@@ -163,9 +163,7 @@ class HtmlDumper extends CliDumper
 <script>
 Sfdump = window.Sfdump || (function (doc) {
 
-if (doc.body instanceof HTMLElement) {
-    doc.body.classList.add('sf-js-enabled');
-}
+doc.documentElement.classList.add('sf-js-enabled');
 
 var rxEsc = /([.*+?^${}()|\[\]\/\\])/g,
     idRx = /\bsf-dump-\d+-ref[012]\w+\b/,


src/Symfony/Component/VarDumper/Dumper/HtmlDumper.php
```diff
@@ -163,8 +163,8 @@ protected function getDumpHeader()
 <script>
 Sfdump = window.Sfdump || (function (doc) {
 
-if (doc.body instanceof HTMLElement) {
-    doc.body.classList.add('sf-js-enabled');
+if (doc.documentElement) {
```
MatTheCat: Can this value ever be falsy?

ohader: Browsers usually "fix" it, however I don't know about all browser tweaks - that's why I added the check here.

ohader: https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement?retiredLocale=de#notes

> For any non-empty HTML document, documentElement will always be an [<html>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/html) element. For any non-empty XML document, documentElement will always be whatever element is the root element of the document.

ohader: ... and I forgot to mention Selenium, WebDriver & Co

MatTheCat: What does it have to do with Selenium?

Anyways if documentElement is falsy I don't think the following script will serve any purpose since the dump won't appear to work. Maybe it would then easier to debug if an error is logged in the console 🤔

(Also the code would crash if documentElement is truthy but not an HTMLElement.)

ohader: I've remove the check. @MatTheCat are you happy with this?

MatTheCat: LGTM, thanks 👍

MatTheCat (approved):

src/Symfony/Component/VarDumper/Dumper/HtmlDumper.php
ohader: Merci! 👍

nicolas-grekas:
Thank you @ohader.

PR #50503, opened 2023-05-31
[SecurityBundle] Fix error message when using OIDC and web-token/jwt-core is not installed

### LINKED ISSUE #50491: [Security] OIDC signature services declaration

6.3

### Description

Reported here: https://github.com/symfony/symfony-docs/pull/18354

image

### How to reproduce

Use `algorithm: ES256` on OIDC security feature, this should throw the error.

### Possible Solution

`ECDSA` classes do not receive any arguments. Removing the arguments from the services declaration should do the trick.

### Additional Context

_No response_

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50491
| License       | MIT
| Doc PR        | -

Before:
> Service "security.access_token_handler.oidc.signature.ES256": Cannot replace arguments for class "Jose\Component\Signature\Algorithm\ES256" if none have been configured yet.  

After:
> You cannot use the "oidc" token handler since "web-token/jwt-core" is not installed. Try running "composer require web-token/jwt-core".

/cc @vincentchalamon


chalasr:
nice catch

PR #50515, opened 2023-05-31
[Mailer] [MailPace] Fix undefined array key in errors response

flofloflo:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | none
| License       | MIT
| Doc PR        | -

This PR fixes a bug which occurs when MailPace returns an `errors` response instead of `error` (see https://docs.mailpace.com/reference/responses/ for possible responses).

Without this fix, the `errors` response leads to the following exception: `ErrorException: Undefined array key "error"`

I added two tests to verify that:
- responses with multiple errors are returned as readable error message
- empty responses generate a readable error message

To fix the bug I added a check for the existence of the `error`/`errors` key in the `$result` variable and a fallback just in case that both are not present.


nicolas-grekas:
Thank you @flofloflo.

Issue #50415, opened 2023-05-24
Environment variable typecast processors return null when used in conjunction with default:


melkamar:
### Symfony version(s) affected

5.4.22

### Description

When typecast env-var processors (e.g. `bool:`) are used together with `default`, the expression may evaluate to `null`. 

When `OPTIONAL_ENV_VAR` is undefined in the following example, the `default:` processor will assign `null` to the result, and the `bool:` processor will pass it through, even though [the documentation](https://symfony.com/doc/current/configuration/env_var_processors.html#built-in-environment-variable-processors) states the return type is `bool`. This is a problem when the value is passed as an non-nullable argument to a strictly-typed class.

```yaml
\EnvironmentService:
  arguments:
    $boolFlag: '%env(bool:default::OPTIONAL_ENV_VAR)%'
```

This is not indented and the processor should be strictly non-nullable, as per the initial discussion here https://github.com/symfony/symfony-docs/pull/18283 .

### How to reproduce

Pass the following as an argument to a Symfony service. Do not set the value of `OPTIONAL_ENV_VAR`.

```php
final class EnvironmentService
{
    public function __construct(
        private bool $boolFlag,
    ) {}
}
```

```yaml
# services.yaml
\EnvironmentService:
  arguments:
    $boolFlag: '%env(bool:default::OPTIONAL_ENV_VAR)%'
```

Symfony container build will fail due to `null` being passed as the `$boolFlag` which has to be a `bool`.

fancyweb:
Looking at the code, it looks more intended than a bug.

fancyweb:
I'd like to close this one. `null` is a "special" value that is not cast by casting env var processors. WDYT @nicolas-grekas?

You should use an explicit default fallback param instead of trying to rely on casting `null`:
```yaml
parameters:
    bool_flag_default_value: false

services:
    \EnvironmentService:
        arguments:
            $boolFlag: '%env(bool:default:bool_flag_default_value:OPTIONAL_ENV_VAR)%'
```
Plz close the issue if it works (I didn't test it :sweat_smile:) and if you think it's good enough :smile:

ro0NL:
alternatively, dont use 'optional envs' :)

melkamar:
> You should use an explicit default fallback param instead of trying to rely on casting null

I know that is a possibility, but it's pretty verbose to be honest. And that was the reason why I wanted to try if it could be made simpler and shorter. Something like this would be ideal and explicit.

```php
services:
[...]

PR #50517, opened 2023-05-31
[DependencyInjection] Fix casting scalar env vars from null

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | -
| Tickets       | https://github.com/symfony/symfony/issues/50415
| License       | MIT
| Doc PR        | -

Using a casting env var processor on `null` returns `null` (it doesn't cast) which is not what one could expect and not what is documented. We could allow it.

### DIFF
--- a/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
+++ b/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
@@ -181,10 +181,12 @@ class EnvVarProcessor implements EnvVarProcessorInterface
                 throw new RuntimeException(sprintf('Unsupported env var prefix "%s".', $prefix));
             }
 
-            return null;
+            if (!\in_array($prefix, ['string', 'bool', 'not', 'int', 'float'], true)) {
+                return null;
+            }
         }
 
-        if (!\is_scalar($env)) {
+        if (null !== $env && !\is_scalar($env)) {
             throw new RuntimeException(sprintf('Non-scalar env var "%s" cannot be cast to "%s".', $name, $prefix));
         }
 
@@ -199,7 +201,7 @@ class EnvVarProcessor implements EnvVarProcessorInterface
         }
 
         if ('int' === $prefix) {
-            if (false === $env = filter_var($env, \FILTER_VALIDATE_INT) ?: filter_var($env, \FILTER_VALIDATE_FLOAT)) {
+            if (null !== $env && false === $env = filter_var($env, \FILTER_VALIDATE_INT) ?: filter_var($env, \FILTER_VALIDATE_FLOAT)) {
                 throw new RuntimeException(sprintf('Non-numeric env var "%s" cannot be cast to int.', $name));
             }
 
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
```diff
@@ -196,6 +196,10 @@ public function getEnv(string $prefix, string $name, \Closure $getEnv): mixed
                 throw new RuntimeException(sprintf('Unsupported env var prefix "%s".', $prefix));
             }
 
+            if (\in_array($prefix, ['string', 'bool', 'not', 'int', 'float'], true)) {
+                goto cast;
```
fancyweb: Alternative to goto :sweat_smile: 
```diff
diff --git a/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php b/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
index f8c56d23f8..5959d88920 100644
--- a/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
+++ b/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
@@ -196,7 +196,9 @@ class EnvVarProcessor implements EnvVarProcessorInterface
                 throw new RuntimeException(sprintf('Unsupported env var prefix "%s".', $prefix));
             }
 
-            return null;
+            if (!\in_array($prefix, ['string', 'bool', 'not', 'int', 'float'], true)) {
+                return null;
+            }
         }
 
         if ('shuffle' === $prefix) {
@@ -205,7 +207,7 @@ class EnvVarProcessor implements EnvVarProcessorInterface
             return $env;
         }
 
-        if (!\is_scalar($env)) {
+        if (null !== $env && !\is_scalar($env)) {
             throw new RuntimeException(sprintf('Non-scalar env var "%s" cannot be cast to "%s".', $name, $prefix));
         }
 
@@ -220,7 +222,7 @@ class EnvVarProcessor implements EnvVarProcessorInterface
         }
 
         if ('int' === $prefix) {
-            if (false === $env = filter_var($env, \FILTER_VALIDATE_INT) ?: filter_var($env, \FILTER_VALIDATE_FLOAT)) {
+            if (null !== $env && false === $env = filter_var($env, \FILTER_VALIDATE_INT) ?: filter_var($env, \FILTER_VALIDATE_FLOAT)) {
                 throw new RuntimeException(sprintf('Non-numeric env var "%s" cannot be cast to int.', $name));
             }
 
@@ -228,7 +230,7 @@ class EnvVarProcessor implements EnvVarProcessorInterface
         }
 
         if ('float' === $prefix) {
-            if (false === $env = filter_var($env, \FILTER_VALIDATE_FLOAT)) {
+            if (null !== $env && false === $env = filter_var($env, \FILTER_VALIDATE_FLOAT)) {
                 throw new RuntimeException(sprintf('Non-numeric env var "%s" cannot be cast to float.', $name));
             }

```

nicolas-grekas: let's go for the version without goto :)

nicolas-grekas (commented):
I'm fine fixing this on 5.4 !

nicolas-grekas (approved):

nicolas-grekas:
Thank you @fancyweb.

bendavies:
is this a not a BC break, or introducion of a regression?
or neither because it's not what is documented?

previously a env var with a default of null:
```yaml
parameters:
  env(SOME_VAR): ~
```

6.3.0:
`'%env(SOME_VAR)%'` will return `null`

6.3.1:
`'%env(SOME_VAR)%'` will return `''` (empty string)

fancyweb:
@bendavies Can you open a new issue please with a reproducer and I'll look at it, thanks.

PR #50524, opened 2023-06-01
Fix Doctrine deprecations

### LINKED ISSUE #50481: Fix Doctrine deprecations

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50481 
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
+++ b/src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
@@ -144,9 +144,9 @@ class Connection implements ResetInterface
             $now,
             $availableAt,
         ], [
-            null,
-            null,
-            null,
+            Types::STRING,
+            Types::STRING,
+            Types::STRING,
             Types::DATETIME_MUTABLE,
             Types::DATETIME_MUTABLE,
         ]);
@@ -273,7 +273,7 @@ class Connection implements ResetInterface
     {
         $configuration = $this->driverConnection->getConfiguration();
         $assetFilter = $configuration->getSchemaAssetsFilter();
-        $configuration->setSchemaAssetsFilter(null);
+        $configuration->setSchemaAssetsFilter(static function () { return true; });
         $this->updateSchema();
         $configuration->setSchemaAssetsFilter($assetFilter);
         $this->autoSetup = false;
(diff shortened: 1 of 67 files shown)


src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
```diff
@@ -480,9 +480,11 @@ private function updateSchema(): void
 
         $schemaManager = $this->createSchemaManager();
         $comparator = $this->createComparator($schemaManager);
-        $schemaDiff = $this->compareSchemas($comparator, $schemaManager->createSchema(), $this->getSchema());
+        $schemaDiff = $this->compareSchemas($comparator, method_exists($schemaManager, 'introspectSchema') ? $schemaManager->introspectSchema() : $schemaManager->createSchema(), $this->getSchema());
```
brusch, line 483: @nicolas-grekas actually this change deletes all existing tables in the database 😬 
Reverting this code-block fixes the issue.

xabbuh: Can you please open a new issue if you think that this change is wrong (best with a failing example application)? Comments on closed PRs are likely to get lost.

xabbuh: By the way, reading the code I fail to see how this change can change the actual behaviour (the new `instrospectSchema()` method calls `createSchema()` internally).

brusch: @xabbuh yeah, I also thought so, but when I just reverted this code block (not the entire file) it continued to work as expected. But yes, I'll create a follow-up ticket with a reproducible case. Thanks so far 👍😊

PR #48705, opened 2022-12-19
[DependencyInjection] Stop considering empty env vars as populated

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no (DX improvement)
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

When a `.env` file declares an empty var (eg `FOO=`), env-var loaders are skipped.
This is a foot gun that has hit me many times when using secrets and that I consider a DX bug.
This PR fixes the issue by ensuring that env vars defined as empty strings don't prevent loaders from being called.

PR #50530, opened 2023-06-01
[DependencyInjection] Fix support for `false` boolean env vars

okhoshi:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

After upgrading `symfony/dependency-injection` package to version 6.3, some of our env vars couldn't be found anymore.

For some scripts, we are providing an adhoc default value to env vars by setting `$_SERVER` directly. However, since version 6.3 of the DependencyInjection component, setting an env var to `false` (the boolean value, like in the snippet below) doesn't work anymore, and Symfony reports that the variable cannot be found.

```php
$_SERVER['FOO'] = false;
```

It seems to be a side effect of the changes made in #48705.

### DIFF
--- a/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
+++ b/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
@@ -146,7 +146,7 @@ class EnvVarProcessor implements EnvVarProcessorInterface
         if (false !== $i || 'string' !== $prefix) {
             $env = $getEnv($name);
         } elseif ('' === ($env = $_ENV[$name] ?? (str_starts_with($name, 'HTTP_') ? null : ($_SERVER[$name] ?? null)))
-            || false === ($env = $env ?? getenv($name) ?? false) // null is a possible value because of thread safety issues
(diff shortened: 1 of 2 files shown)


stof:
environment variables are always strings though. What you are covering in tests is not something compatible with actual environment variables (or with the DotEnv component).

okhoshi:
Because you can set any scalar value from PHP code, and they won't be converted to strings nor anything will prevent you to set them in the first place. FWIW, we've never had any troubles so far, be it with Symfony or DotEnv component.

I'm covering with tests because it was a supported behaviour that has been broken with 6.3 release. If you think it shouldn't have been supported from the start, and thus that this is not a bug, I'm also fine with that, feel free to let me know and close this PR.
But it's not really consistent then since any other scalar value will work.

stof:
It was not a _supported_ behavior. It was an _undefined_ behavior that happened to work.

nicolas-grekas (approved):
Works for me, I'm fine keeping the behavior from 6.2.

nicolas-grekas:
Thank you @Okhoshi.

PR #50537, opened 2023-06-02
[Console] Add placeholders to ProgressBar for exact times

### LINKED ISSUE #49927: [Console] Improve the helpfulness of %remaining% in Progress Bar
### Description

At present the remaining time on the progress bar is not that helpful. If the remaining time is between 24 and 47 hours it will just read "1 days remaining".

Similarly if the time remaining is 1 hour 55 minutes, the text will actually display "1 hour remaining".

Obviously this isn't that helpful when planning around a process finishing.

In all my projects I now add this:

```php
ProgressBar::setPlaceholderFormatterDefinition('remaining', function($bar) {
            $remaining = $bar->getRemaining();
            $hours = floor($remaining / 3600);
            $minutes = floor(($remaining / 60) % 60);
            $seconds = $remaining % 60;
            return sprintf('%02d:%02d:%02d', $hours, $minutes, $seconds);
        });
```

With this, when using the format: `%current%/%max% [%bar%] %percent:3s%% %elapsed:6s% taken %remaining:-6s% remaining`, I will get the following:

This: `5/169200 [>---------------------------]   0% 5 secs taken 46:59:55 remaining`

Instead of this: `5/169200 [>---------------------------]   0% 5 secs taken 1 day  remaining`

Or in the case of a 1 hour 48 minute process:

This: `5/6500 [>---------------------------] 0% 5 secs taken 01:48:15 remaining`

Instead of this: `5/6500 [>---------------------------] 0% 5 secs taken 1 hr remaining`

This could be implemented by updating `Symfony\Component\Console\Helper`

https://github.com/symfony/console/blob/3582d68a64a86ec25240aaa521ec8bc2342b369b/Helper/Helper.php#L91

To use the same logic as above. Although this would have the side effect of changing all dates. So it would in fact look like this:

`5/6500 [>---------------------------] 0% 00:00:05 taken 01:48:15 remaining`

I don't have an issue with that, but perhaps it's not desired?

I can create a PR if this is okay, but i've never contributed to Symfony before so figured I would start by creating an issue.

### Example

_No response_

maxbeckers:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #49927
| License       | MIT
| Doc PR        | symfony/symfony-docs#... TBD

This is an idea for exact times using ProgressBar based on the idea of @GromNaN in the issue  #49927.

Open to discuss this first way of implementing it.

I'll create the docs PR when the feature is agreed and there won't come up any bigger changes.

The Idea is to show the exact time in seconds, the ProgressBar will run / is running.

### DIFF
--- a/src/Symfony/Component/Console/Helper/ProgressBar.php
+++ b/src/Symfony/Component/Console/Helper/ProgressBar.php
@@ -540,20 +540,20 @@ final class ProgressBar
 
                 return $display;
             },
-            'elapsed' => fn (self $bar) => Helper::formatTime(time() - $bar->getStartTime()),
+            'elapsed' => fn (self $bar) => Helper::formatTime(time() - $bar->getStartTime(), 2),
             'remaining' => function (self $bar) {
                 if (!$bar->getMaxSteps()) {
                     throw new LogicException('Unable to display the remaining time if the maximum number of steps is not set.');
                 }
 
-                return Helper::formatTime($bar->getRemaining());
+                return Helper::formatTime($bar->getRemaining(), 2);
             },
             'estimated' => function (self $bar) {
                 if (!$bar->getMaxSteps()) {
                     throw new LogicException('Unable to display the estimated time if the maximum number of steps is not set.');
                 }
 
-                return Helper::formatTime($bar->getEstimated());
+                return Helper::formatTime($bar->getEstimated(), 2);
             },
             'memory' => fn (self $bar) => Helper::formatMemory(memory_get_usage(true)),
             'current' => fn (self $bar) => str_pad($bar->getProgress(), $bar->getStepWidth(), ' ', \STR_PAD_LEFT),
--- a/src/Symfony/Component/Console/Tests/Helper/HelperTest.php
+++ b/src/Symfony/Component/Console/Tests/Helper/HelperTest.php
@@ -20,26 +20,31 @@ class HelperTest extends TestCase
     public static function formatTimeProvider()
     {
         return [
-            [0,      '< 1 sec'],
-            [1,      '1 sec'],
-            [2,      '2 secs'],
-            [59,     '59 secs'],
-            [60,     '1 min'],
-            [61,     '1 min'],
-            [119,    '1 min'],
-            [120,    '2 mins'],
-            [121,    '2 mins'],
-            [3599,   '59 mins'],
-            [3600,   '1 hr'],
-            [7199,   '1 hr'],
-            [7200,   '2 hrs'],
-            [7201,   '2 hrs'],
-            [86399,  '23 hrs'],
-            [86400,  '1 day'],
-            [86401,  '1 day'],
-            [172799, '1 day'],
-            [172800, '2 days'],
-            [172801, '2 days'],
+            [0,      '< 1 sec', 1],
+            [0.95,   '< 1 sec', 1],
+            [1,      '1 sec', 1],
+            [2,      '2 secs', 2],
+            [59,     '59 secs', 1],
+            [59.21,  '59 secs', 1],
+            [60,     '1 min', 2],
+            [61,     '1 min, 1 sec', 2],
+            [119,    '1 min, 59 secs', 2],
+            [120,    '2 mins', 2],
+            [121,    '2 mins, 1 sec', 2],
+            [3599,   '59 mins, 59 secs', 2],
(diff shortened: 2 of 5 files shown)


src/Symfony/Component/Console/Helper/ProgressBar.php
```diff
@@ -535,20 +535,35 @@ private static function initPlaceholderFormatters(): array
                 return $display;
             },
             'elapsed' => fn (self $bar) => Helper::formatTime(time() - $bar->getStartTime()),
```
OskarStark: Why not refine the existing data points and make them more exact instead of doubling?

maxbeckers: My idea was to keep the existing api and just add new parameters if it's needed that the times are a bit more finetuned.

wouterj:
I think we can simplify this to just replace the existing `elapsed` parameter with the new format. We do not promise backwards compatible CLI output, and I would argue that interactive elements like the progress bar should never be interpreted by any program anyways.

maxbeckers:
Thanks @wouterj and @OskarStark for your input. I now changed it.

wouterj (approved):
Looks good to me!

src/Symfony/Component/Console/Tests/Helper/HelperTest.php
```diff
@@ -21,25 +21,27 @@ public static function formatTimeProvider()
...
             [7200,   '2 hrs'],
-            [7201,   '2 hrs'],
-            [86399,  '23 hrs'],
+            [7201,   '2 hrs, 1 sec'],
+            [86399,  '23 hrs, 59 mins, 59 secs'],
             [86400,  '1 day'],
-            [86401,  '1 day'],
-            [172799, '1 day'],
+            [86401,  '1 day, 1 sec'],
+            [172799, '1 day, 23 hrs, 59 mins, 59 secs'],
```
GromNaN, line 42: Do we need precision to the second when the duration is in days? 2 units should be enough.

```suggestion
            [172799, '1 day, 23 hrs'],
```

GromNaN: Do we need precision to the second when the duration is in days? 2 units should be enough.

```suggestion
            [86401,  '1 day'],
            [172799, '1 day, 23 hrs'],
```

maxbeckers: Good idea, I'll add it.

GromNaN: If we add a `$precision = 1` to `Helper::formatTime()`, it keeps the function backward compatible and allows. To fine-tune the display by updating the placeholder function.

GromNaN: @maxbeckers thanks. I see we thought of the same solution. The default value of `$precision` could be `1` for backward compatibility. And the argument added in the method call.
https://github.com/symfony/symfony/blob/3147082e35faadd873a801a4011bc8cbb79807e0/src/Symfony/Component/Console/Helper/ProgressIndicator.php#L231

maxbeckers: Yes you're right that we that keep the current behavior, but then we don't have the benefit for the functions here like this one:
https://github.com/symfony/symfony/blob/3147082e35faadd873a801a4011bc8cbb79807e0/src/Symfony/Component/Console/Helper/ProgressBar.php#L551-L556
Because to make this a bit more exact was the issue. So i could make the default to `$precision = 1` and maybe change it here to `2` or i keep it everywhere with 2. 

Fine with both ways, what do you think?

GromNaN: I prefer that you add the argument where the function is called. @wouterj @OskarStark What do you think?

maxbeckers: Changet it to default 1 and added the 2 to the function calls.

OskarStark (approved):

GromNaN (approved):

fabpot (approved):

fabpot:
Thank you @maxbeckers.

PR #50545, opened 2023-06-03
[Process] Fix standard streams not available in PhpProcess scripts

### LINKED ISSUE #50063: [Process] STDIN/STDOUT/STDERR with PhpProcess

6.3.x-dev

### Description

`PhpProcess` is using STDIN to feed the script, which makes STDIN/STDOUT/STDERR are not availble (see php/php-src#11104)

Note: it's tested with `6.3.x-dev` but it should be reproducible with old versions like `5.4.x` as well

### How to reproduce

```php
<?php

include './vendor/autoload.php';
use Symfony\Component\Process\PhpProcess;

$process = new PhpProcess(<<<PHP
<?php
var_dump(php_sapi_name());
var_dump(STDOUT);

PHP
);
```

### Possible Solution

Create a temporary script before running, just following what are done for phpdbg (https://github.com/symfony/process/blob/6.2/PhpProcess.php#L43-L47), 

### Additional Context

_No response_

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50063
| License       | MIT
| Doc PR        | _NA_

Although[ it's been fixed in PHP lately](https://github.com/php/php-src/commit/f6c0c60ef63c1e528dd3bd945c8f22270bbe3837), older versions are suffering this bug, including minimal version of PHP of every supported Symfony version.

### DIFF
--- a/src/Symfony/Component/Process/PhpProcess.php
+++ b/src/Symfony/Component/Process/PhpProcess.php
@@ -39,15 +39,17 @@ class PhpProcess extends Process
             $php = $executableFinder->find(false);
             $php = false === $php ? null : array_merge([$php], $executableFinder->findArguments());
         }
-        if ('phpdbg' === \PHP_SAPI) {
-            $file = tempnam(sys_get_temp_dir(), 'dbg');
-            file_put_contents($file, $script);
-            register_shutdown_function('unlink', $file);
-            $php[] = $file;
-            $script = null;
+
+        $file = tempnam(sys_get_temp_dir(), 'phpdbg' === \PHP_SAPI ? 'dbg' : 'cli');
+        if (false === $file) {
+            throw new RuntimeException('The system temp directory is not writable.');
         }
 
-        parent::__construct($php, $cwd, $env, $script, $timeout);
+        file_put_contents($file, $script);
+        register_shutdown_function('unlink', $file);
+        $php[] = $file;
+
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/Process/PhpProcess.php
```diff
@@ -39,15 +39,17 @@ public function __construct(string $script, string $cwd = null, array $env = nul
...
-        if ('phpdbg' === \PHP_SAPI) {
-            $file = tempnam(sys_get_temp_dir(), 'dbg');
-            file_put_contents($file, $script);
-            register_shutdown_function('unlink', $file);
-            $php[] = $file;
-            $script = null;
+
+        $file = tempnam(sys_get_temp_dir(), 'phpdbg' === \PHP_SAPI ? 'dbg' : 'cli');
+        if (false === $file) {
+            throw new RuntimeException('The system temp directory could not be written.');
```
javiereguiluz: `... could not be written` -> `... is not writable`?

alexandre-daubois: That's better, thanks! :slightly_smiling_face:

nicolas-grekas:
Closing as won't fix. This can be worked around by using `fopen('php://input', 'r')` and the workaround would require write access, which is a significant behavior change to me.

PR #50546, opened 2023-06-04
[FrameworkBundle] Fix `debug:config` & `config:dump` in debug mode

### LINKED ISSUE #47623: [FrameworkBundle] ConfigDumpReferenceCommand is not recognizing container extensions loaded via kernel
### Description

Symfony\Bundle\FrameworkBundle\Command\ConfigDumpReferenceCommand do not use the existing container generated by the kernel to find container extensions. This leads to an issue dumping config references for container extensions not loaded via an (old) bundle but via the kernel.

### Example

I've added a container extension on my Kernel->build() function via... 
```
protected function build(ContainerBuilder $container): void
{
    $container->registerExtension(new MytestExtension());
}
```
Now i want to dump the configuration references via...
```
bin/console debug:config mytest
and
bin/console config:dump-reference mytest
```
but i get the error output
```
No extensions with configuration available for "mytest".
```

After some debugging i found out, that Symfony\Bundle\FrameworkBundle\Command\ConfigDumpReferenceCommand uses a findExtension() function from Symfony\Bundle\FrameworkBundle\Command\AbstractConfigCommand which loads available container extensions just for bundles in the [initializeBundles()](https://github.com/symfony/framework-bundle/blob/6.1/Command/AbstractConfigCommand.php#L134) function.

Some steps back i found out, that[ i have to implement Symfony\Component\DependencyInjection\Extension\ExtensionInterface and Symfony\Component\DependencyInjection\Extension\ConfigurationExtensionInterface](https://github.com/symfony/framework-bundle/blob/6.1/Command/AbstractConfigCommand.php#L63) on my Kernel so my extension is loaded directly via my Kernel->getConfiguration() function.

**Questions**
Wouldn't it be possible to easily use the existing container [(which includes my extension) build by the kernel](https://github.com/symfony/http-kernel/blob/6.1/Kernel.php#L663) inside of AbstractConfigCommand so my kernel don't have to implement those interfaces ?

...or at least add the kernel->build() call to the initializeBundles() function ?

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #47623
| License       | MIT
| Doc PR        | -

Displaying configs for extensions without a bundle using the `debug:config` & `config:dump-reference` commands works depending on whether debug is `true` or `false`. The reason behind this is the following code:

https://github.com/symfony/symfony/blob/b4128fdefea4ff95b27861bf65ba789d24868df1/src/Symfony/Bundle/FrameworkBundle/Command/BuildDebugContainerTrait.php#L42-L60

When debug is `true` the extensions are never loaded in the container, so the commands don't work.

There are even tests for these cases but they are all executed with debug `false`.

This PR aims to make the commands work with both debug `true` & `false`. Another problem is that these extensions are not visible in the list of available extensions and are not offered by the completion feature, but since those seem more like new features I've created a separate PR for 6.4: #50548

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/Functional/ConfigDebugCommandTest.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/Functional/ConfigDebugCommandTest.php
@@ -23,36 +23,53 @@ use Symfony\Component\Console\Tester\CommandTester;
  */
 class ConfigDebugCommandTest extends AbstractWebTestCase
 {
-    private $application;
-
-    protected function setUp(): void
+    /**
+     * @testWith [true]
+     *           [false]
+     */
+    public function testDumpKernelExtension(bool $debug)
     {
-        $kernel = static::createKernel(['test_case' => 'ConfigDump', 'root_config' => 'config.yml']);
-        $this->application = new Application($kernel);
-        $this->application->doRun(new ArrayInput([]), new NullOutput());
+        $tester = $this->createCommandTester($debug);
+        $ret = $tester->execute(['name' => 'foo']);
+
+        $this->assertSame(0, $ret, 'Returns 0 in case of success');
+        $this->assertStringContainsString('foo:', $tester->getDisplay());
+        $this->assertStringContainsString('    foo: bar', $tester->getDisplay());
     }
 
-    public function testDumpBundleName()
+    /**
+     * @testWith [true]
+     *           [false]
+     */
+    public function testDumpBundleName(bool $debug)
     {
-        $tester = $this->createCommandTester();
+        $tester = $this->createCommandTester($debug);
         $ret = $tester->execute(['name' => 'TestBundle']);
 
         $this->assertSame(0, $ret, 'Returns 0 in case of success');
         $this->assertStringContainsString('custom: foo', $tester->getDisplay());
     }
 
-    public function testDumpBundleOption()
+    /**
+     * @testWith [true]
+     *           [false]
+     */
+    public function testDumpBundleOption(bool $debug)
     {
-        $tester = $this->createCommandTester();
+        $tester = $this->createCommandTester($debug);
         $ret = $tester->execute(['name' => 'TestBundle', 'path' => 'custom']);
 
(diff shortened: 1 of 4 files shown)


chalasr (approved):

src/Symfony/Bundle/FrameworkBundle/Tests/Functional/ConfigDebugCommandTest.php
```diff
@@ -23,114 +23,154 @@
...
 
-        $tester = $this->createCommandTester();
+        $tester = $this->createCommandTester($debug);
         $tester->execute(['name' => 'ExtensionWithoutConfigTestBundle']);
     }
 
+    public static function provideDebugValues(): iterable
+    {
+        yield [true];
+        yield [false];
```
nicolas-grekas: let's replace this provider with `@testWith` annotations

src/Symfony/Bundle/FrameworkBundle/Tests/Functional/ConfigDumpReferenceCommandTest.php
```diff
@@ -92,27 +100,49 @@ public function testDumpAtPathXml()
         $this->assertStringContainsString('[ERROR] The "path" option is only available for the "yaml" format.', $tester->getDisplay());
     }
 
+    public static function provideDebugValues(): iterable
+    {
+        yield [true];
+        yield [false];
```
nicolas-grekas: `@testWith` here also

src/Symfony/Bundle/FrameworkBundle/Tests/Functional/ConfigDebugCommandTest.php
HypeMC: Done https://github.com/symfony/symfony/compare/73871a75d718a38ee9204ed4918579ea14162b3b..2824fc5d69e874a625e91bad56cbbf429469dc89

nicolas-grekas (approved):

nicolas-grekas:
Thank you @HypeMC.

PR #50552, opened 2023-06-04
[Security] Allow custom scheme to be used as redirection URIs

### LINKED ISSUE #50500: [Security] `DefaultAuthenticationSuccessHandler` crashes if the target URL is wrongly considered as a route name

≥ 5.4

### Description

Because `HttpUtils::generateUri` considers `$path` a route name if it does not start with `http` or `/`,  `DefaultAuthenticationSuccessHandler` will crash for any target URL whose scheme does not start with `http`.

### How to reproduce

Simplest way to reproduce is to create a login form configured with `always_use_default_target_path` and `default_target_path` as a URL whose scheme does not start with `http`, like `android-app://com.google.android.gm/` (see #50385).

### Possible Solution

I think `HttpUtils::generateUri` should consider any URL as one, whatever its scheme.

Then, maybe `DefaultAuthenticationSuccessHandler` could expose a mechanism to confirm whether to redirect to the referrer?

### Additional Context

_No response_

Spomky:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #50500
| License       | MIT
| Doc PR        | not needed

ping @sdespont and @MatTheCat

This PR aims at fixing the redirection issue where only URLs starting with `http` are allowed.
With the modified behavior, it is now allowed to use any URL scheme. It will be possible to redirect to `android-app://com.google.android.gm/`.

~In addition, it prevents the redirection to the following URLs:~

* ~With path traversal e.g. `https://example.com/foo/../../.htpasswd`~
* ~With protocol-relative e.g. `//malicious.app/foo/bar`~

### DIFF
--- a/src/Symfony/Component/Security/Http/HttpUtils.php
+++ b/src/Symfony/Component/Security/Http/HttpUtils.php
@@ -148,7 +148,9 @@ class HttpUtils
      */
     public function generateUri(Request $request, string $path)
     {
-        if (str_starts_with($path, 'http') || !$path) {
+        $url = parse_url($path);
+
+        if ('' === $path || isset($url['scheme'], $url['host'])) {
             return $path;
         }
 
--- a/src/Symfony/Component/Security/Http/Tests/HttpUtilsTest.php
+++ b/src/Symfony/Component/Security/Http/Tests/HttpUtilsTest.php
@@ -59,6 +59,54 @@ class HttpUtilsTest extends TestCase
     }
 
     /**
+     * @dataProvider validRequestDomainUrls
+     */
+    public function testCreateRedirectResponse(?string $domainRegexp, string $path, string $expectedRedirectUri)
+    {
+        $utils = new HttpUtils($this->getUrlGenerator(), null, $domainRegexp);
+        $response = $utils->createRedirectResponse($this->getRequest(), $path);
+
+        $this->assertTrue($response->isRedirect($expectedRedirectUri));
+        $this->assertEquals(302, $response->getStatusCode());
+    }
+
+    public static function validRequestDomainUrls()
+    {
+        return [
+            '/foobar' => [
+                null,
+                '/foobar',
+                'http://localhost/foobar',
+            ],
+            'http://symfony.com/ without domain regex' => [
+                null,
+                'http://symfony.com/',
+                'http://symfony.com/',
+            ],
+            'http://localhost/blog with #^https?://symfony\.com$#i' => [
+                '#^https?://symfony\.com$#i',
+                'http://symfony.com/blog',
+                'http://symfony.com/blog',
+            ],
+            'http://localhost/blog with #^https?://%s$#i' => [
+                '#^https?://%s$#i',
+                'http://localhost/blog',
+                'http://localhost/blog',
+            ],
+            'custom scheme' => [
+                null,
+                'android-app://com.google.android.gm/',
+                'android-app://com.google.android.gm/',
+            ],
+            'custom scheme with all URL components' => [
+                null,
+                'android-app://foo:bar@www.example.com:8080/software/index.html?lite=true#section1',
+                'android-app://foo:bar@www.example.com:8080/software/index.html?lite=true#section1',
+            ],
+        ];
+    }
+
+    /**
      * @dataProvider badRequestDomainUrls
      */
     public function testCreateRedirectResponseWithBadRequestsDomain($url)
@@ -77,6 +125,7 @@ class HttpUtilsTest extends TestCase
             ['http:/\\pirate.net/foo'],
             ['http:\\/pirate.net/foo'],
             ['http://////pirate.net/foo'],
+            ['http:///foo'],
         ];
     }
 


Spomky:
The support for URNs could be removed. It looks like it is not part of the current best practices (see [RFC8252](https://datatracker.ietf.org/doc/html/rfc8252)); custom scheme like `android-app` is now prefered.

src/Symfony/Component/Security/Http/HttpUtils.php
```diff
@@ -148,7 +148,11 @@ public function checkRequestPath(Request $request, string $path)
      */
     public function generateUri(Request $request, string $path)
     {
-        if (str_starts_with($path, 'http') || !$path) {
+        $parsedUrl = parse_url($path);
+        $uriScheme = $parsedUrl['scheme'] ?? null;
+        $uriHost = $parsedUrl['host'] ?? null;
+        $uriPath = $parsedUrl['path'] ?? '';
+        if (('urn' === $uriScheme && null === $uriHost && null !== $uriPath) || (null !== $uriScheme && null !== $uriHost)) {
```
nicolas-grekas:
```suggestion
        if (isset($url['scheme']) && (isset($url['host']) || ('urn' === $url['scheme'] && isset($url['path'])))) {
```

Spomky: Many thanks!

src/Symfony/Component/Security/Http/HttpUtils.php
```diff
@@ -148,7 +148,11 @@ public function checkRequestPath(Request $request, string $path)
      */
     public function generateUri(Request $request, string $path)
     {
-        if (str_starts_with($path, 'http') || !$path) {
+        $parsedUrl = parse_url($path);
```
nicolas-grekas:
```suggestion
        $url = parse_url($path);
```

+ remove the net lines :)

src/Symfony/Component/Security/Http/HttpUtils.php
```diff
@@ -148,7 +148,11 @@ public function checkRequestPath(Request $request, string $path)
      */
     public function generateUri(Request $request, string $path)
     {
-        if (str_starts_with($path, 'http') || !$path) {
+        $url = parse_url($path);
+        $uriScheme = $parsedUrl['scheme'] ?? null;
+        $uriHost = $parsedUrl['host'] ?? null;
+        $uriPath = $parsedUrl['path'] ?? '';
```
nicolas-grekas:
```suggestion
```

nicolas-grekas (approved):
(don't miss syncing the PR description with latest changes)

src/Symfony/Component/Security/Http/HttpUtils.php
```diff
@@ -148,7 +148,20 @@ public function checkRequestPath(Request $request, string $path)
...
+        $url = parse_url($path);
+
+        // Prevent path traversal (ex: https://example.com/../../file)
+        if (preg_match('#/\.\.?(/|$)#', $url['path'])) {
+            return $request->getUriForPath('/');
+        }
+
+        // Prevent protocol-relative redirection (ex: //example.com/file)
+        if (!isset($url['scheme']) && isset($url['host'], $url['path'])) {
+            return $request->getUriForPath('/');
```
stof: this is not the same behavior than before your PR

chalasr (changes requested):
See stof's review. Also keeping existing tests untouched makes the patch easier to review, which makes me much more confident to merge on security-related topics especially. Please avoid any refactoring if possible :)

Spomky:
Hi @chalasr,

Many thanks for your comment. I restored the previous tests and keep the one I created. Let me know if you agree with the modifications.

Regarding the behavior of paths starting with `//`, the change is intentional. I took the opportunity of this PR to fix the comment `Protocol-relative redirection should not be supported for security reasons`. There is now reason for a developer to redirect to such path and could be considered as an open redirect in some situation.
I can revert the change and open another PR if you prefer. At some point, it should be addressed and it seems to be the perfect moment for me.

nicolas-grekas (approved):
I removed the part about supposedly security hardening, which are unproven to me and change the behavior.

Spomky:
Agreed. Let's keep it simple and without any BC.

src/Symfony/Component/Security/Http/HttpUtils.php
```diff
@@ -148,7 +148,10 @@ public function checkRequestPath(Request $request, string $path)
      */
     public function generateUri(Request $request, string $path)
     {
-        if (str_starts_with($path, 'http') || !$path) {
+        $url = parse_url($path);
+
+        // absolute URL
+        if ('' === $path || isset($url['scheme'], $url['host'])) {
```
stof, line 153: `'' === $path` is not really an absolute URL. The comment is misleading.

nicolas-grekas: comment removed

nicolas-grekas:
Thank you @Spomky.

PR #50561, opened 2023-06-05
Fix AppVeyor

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | -
| Deprecations? | -
| Tickets       | -
| License       | MIT
| Doc PR        | -

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -5,7 +5,7 @@ clone_folder: c:\projects\symfony
 init:
     - SET PATH=c:\php;%PATH%
     - SET COMPOSER_NO_INTERACTION=1
-    - SET SYMFONY_DEPRECATIONS_HELPER=strict
+    - SET "SYMFONY_DEPRECATIONS_HELPER=max[total]=0&baselineFile=c:\projects\symfony\.github\deprecations-baseline.json"
     - SET ANSICON=121x90 (121x90)
     - SET SYMFONY_PHPUNIT_DISABLE_RESULT_CACHE=1
     - REG ADD "HKEY_CURRENT_USER\Software\Microsoft\Command Processor" /v DelayedExpansion /t REG_DWORD /d 1 /f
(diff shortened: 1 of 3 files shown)


.appveyor.yml
```diff
@@ -5,7 +5,7 @@ clone_folder: c:\projects\symfony
 init:
     - SET PATH=c:\php;%PATH%
     - SET COMPOSER_NO_INTERACTION=1
-    - SET SYMFONY_DEPRECATIONS_HELPER=strict
+    - SET "SYMFONY_DEPRECATIONS_HELPER=max[total]=0&baselineFile=c:\projects\symfony\.github\deprecations-baseline.json"
```
fancyweb: strict === max[total]=0
It looks like we cannot use a predefined mode such as strict or weak AND another option

src/Symfony/Component/Process/Process.php
```diff
@@ -428,7 +428,7 @@ public function wait(callable $callback = null)
 
         do {
             $this->checkTimeout();
-            $running = $this->isRunning() && ('\\' === \DIRECTORY_SEPARATOR || $this->processPipes->areOpen());
+            $running = '\\' === \DIRECTORY_SEPARATOR ? $this->isRunning() : $this->processPipes->areOpen();
```
stof, line 431: this would stop using `isRunning` on Unix systems. Is this expected ?

src/Symfony/Component/Process/Tests/ProcessTest.php
```diff
@@ -1538,16 +1538,6 @@ public function testEnvCaseInsensitiveOnWindows()
         }
     }
 
-    public function testNotTerminableInputPipe()
```
stof, line 1541: Why are you removing this test ?

fancyweb:
Sorry, I’m just trying to find a guilty commit that might cause the
constant AppVeyor 20min timeout and it could be this one so I reverted it
just to see.

PR #49610, opened 2023-03-06
[DoctrineBridge] deprecate doctrine schema subscribers in favor of listeners

alli83:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | yes
| Tickets       | Fix #49387 
| License       | MIT
| Doc PR        | 

As @nicolas-grekas explained in https://github.com/symfony/symfony/issues/49387#issuecomment-1442179095 : 

" Because PdoSessionHandlerSchemaSubscriber is registered as a subscriber, ContainerAwareEventManager must instantiate it to know which events should be listened for. This triggers the instantiation of the session.handler service, which is a NativeFileSessionHandler, and this notice when this is done after headers have been sent."

It was therefore decided to change all the doctrine schema subscribers in favor of listeners in order to avoid this kind of problem. 

Being already present before Symfony 6.3, 
`MessengerTransportDoctrineSchemaSubscriber`, `DoctrineDbalCacheAdapterSchemaSubscriber` and `RememberMeTokenProviderDoctrineSchemaSubscriber`have been deprecated. 

TODO:
- [x] : update DoctrineBundle => in progress

PR #50571, opened 2023-06-06
[DoctrineBridge] Kill DBAL 2 support

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

DBAL 2 is dead and buried. Let's remove code that we needed to maintain DBAL 2.

TODO:

- [x] Deprecation layer for removed classes
- [x] Check changes for possible backports to 5.4/6.4
- [x] #50576
- [x] #50579
- [x] #50575

PR #50575, opened 2023-06-06
[DoctrineBridge] Remove deprecated classes and `ContainerAwareEventManager::getListeners()` deprecation

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Removes deprecated code introduced in https://github.com/symfony/symfony/pull/49610


src/Symfony/Bridge/Doctrine/Tests/ContainerAwareEventManagerTest.php
```diff
@@ -245,16 +245,14 @@ public function testGetListenersForEventWhenSubscribersArePresent()
...
     {
         $this->container->set('lazy', $listener1 = new MyListener());
         $this->evm->addEventListener('foo', 'lazy');
         $this->evm->addEventListener('foo', $listener2 = new MyListener());
 
-        $this->expectDeprecation('Since symfony/doctrine-bridge 6.2: Calling "Symfony\Bridge\Doctrine\ContainerAwareEventManager::getListeners()" without an event name is deprecated. Call "getAllListeners()" instead.');
+        $this->expectExceptionMessage('"Symfony\Bridge\Doctrine\ContainerAwareEventManager::getListeners()" must be called with an event name.');
+        $this->expectException(\InvalidArgumentException::class);
 
         $this->assertSame([$listener1, $listener2], array_values($this->evm->getListeners()['foo']));
```
derrabus: We expect the call to throw, so performing assertions on the return value does not make sense anymore.

derrabus (changes requested):

src/Symfony/Bridge/Doctrine/Tests/ContainerAwareEventManagerTest.php
alexandre-daubois: Oh that's right. Fixed :slightly_smiling_face:

src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
```diff
@@ -69,10 +69,9 @@ public function dispatchEvent($eventName, EventArgs $eventArgs = null): void
     public function getListeners($event = null): array
     {
         if (null === $event) {
-            trigger_deprecation('symfony/doctrine-bridge', '6.2', 'Calling "%s()" without an event name is deprecated. Call "getAllListeners()" instead.', __METHOD__);
-
-            return $this->getAllListeners();
+            throw new \InvalidArgumentException(sprintf('"%s()" must be called with an event name.', __METHOD__));
```
derrabus: The reason for deprecating `null` here was doctrine/event-manager#50. Simply throwing is not enough because as long as we declare compatibility with EventManager 1, we have to accept `null` as argument because that's the contract of EventManager 1.

What we need to do here is to declare the DoctrineBridge incompatible with `doctrine/event-manager` < 2. And once we did that, we can change the signature to…

```php
public function getListeners(string $event): array
```

UPGRADE-7.0.md
```diff
@@ -5,6 +5,14 @@ Symfony 6.4 and Symfony 7.0 will be released simultaneously at the end of Novemb
 release process, both versions will have the same features, but Symfony 7.0 won't include any deprecated features.
 To upgrade, make sure to resolve all deprecation notices.
 
+Doctrine
```
stof: the title should be `Doctrine Bridge` IMO

alexandre-daubois: Yes!

src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
alexandre-daubois: Updated dependencies as well as tests with the last deprecation present in `ContainerAwareEventManager` :+1: Thanks for the hints!

composer.json
```diff
@@ -164,6 +164,7 @@
         "async-aws/core": "<1.5",
         "doctrine/annotations": "<1.13.1",
         "doctrine/dbal": "<2.13.1",
+        "doctrine/event-manager": "<2",
```
derrabus: I wasn't aware that `doctrine/event-manager` is a mandatory dependency. Your adjustments to the `require` section are enough. No need for this conflict rule then.

derrabus (changes requested):

composer.json
alexandre-daubois: Oh right, updated both composer.json

src/Symfony/Bridge/Doctrine/Tests/ContainerAwareEventManagerTest.php
```diff
@@ -81,62 +73,22 @@ public function testDispatchEvent()
...
-        $this->assertSame(1, $listener2->calledByEventNameCount);
-        $this->assertSame(0, $subscriber1->calledByInvokeCount);
-        $this->assertSame(1, $subscriber1->calledByEventNameCount);
-        $this->assertSame(1, $subscriber2->calledByInvokeCount);
-        $this->assertSame(0, $subscriber2->calledByEventNameCount);
-    }
-
     public function testAddEventListenerAfterDispatchEvent()
     {
-        $this->container->set('lazy1', $listener1 = new MyListener());
```
stof, line 120: Why is this changing the test ?

src/Symfony/Bridge/Doctrine/Tests/ContainerAwareEventManagerTest.php
```diff
@@ -40,29 +40,21 @@ public function testDispatchEventRespectOrder()
...
-        $this->expectDeprecation('Since symfony/doctrine-bridge 6.3: Using Doctrine subscribers as services is deprecated, declare listeners instead');
-        $this->assertSame([$subscriber1,  $subscriber2], array_values($this->evm->getListeners('foo')));
+        $this->expectException(\InvalidArgumentException::class);
+        $this->expectExceptionMessage('Using Doctrine subscriber "Symfony\Bridge\Doctrine\Tests\MySubscriber" is not allowed, declare it as a listener instead.');
+        $this->evm->getListeners('foo');
     }
 
     public function testDispatchEvent()
     {
-        $this->container->set('lazy1', $listener1 = new MyListener());
```
stof: Why are you changing all those tests ? When removing deprecated features, non-legacy tests should not change at all.

stof (changes requested):

src/Symfony/Bridge/Doctrine/Tests/ContainerAwareEventManagerTest.php
alexandre-daubois: `EventManager::addEventListener()` now only accepts an object as the second argument. Because of this, there's no need to set services in the container and we must pass the listener object directly to the method instead of the service alias

stof: But this would mean we would have to deprecate passing a string in 6.4 (and then, is this still lazy-loading the listeners, which is what this whole class is about ?). We cannot remove a non-deprecated API.

stof: Side note: it is still valid to widen a parameter type in a child class. So we are not forced to use `object` as type there. We could use `object|string` if we need it for our lazy-loading API.

alexandre-daubois: Alright, I missed the lazy-loading part. Thank you, I updated and reverted accordingly :+1:

stof: That's exactly why it is very important to not do any change in non-legacy tests in this kind of PR removing deprecated code, so that the testsuite keeps telling you about the code that is not allowed to regress.

src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
```diff
@@ -66,13 +66,8 @@ public function dispatchEvent($eventName, EventArgs $eventArgs = null): void
     /**
      * @return object[][]
```
stof: I suggest removing this phpdoc, so that we inherit the more precise one defined in the parent class (should probably be done in 6.4 btw), especially given that this is a wrong phpdoc as this method does not return nested arrays when an event name is passed.

alexandre-daubois: Alright thanks!

UPGRADE-7.0.md
```diff
@@ -5,6 +5,15 @@ Symfony 6.4 and Symfony 7.0 will be released simultaneously at the end of Novemb
 release process, both versions will have the same features, but Symfony 7.0 won't include any deprecated features.
 To upgrade, make sure to resolve all deprecation notices.
 
+DoctrineBridge
+--------------
+
+ * Removed `DoctrineDbalCacheAdapterSchemaSubscriber`, use `DoctrineDbalCacheAdapterSchemaListener` instead
+ * Removed `MessengerTransportDoctrineSchemaSubscriber`, use `MessengerTransportDoctrineSchemaListener` instead
+ * Removed `RememberMeTokenProviderDoctrineSchemaSubscriber`, use `RememberMeTokenProviderDoctrineSchemaListener` instead
```
derrabus: We use present tense in these documents: "Remove" instead of "Removed".

src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
```diff
@@ -107,7 +99,7 @@ public function hasListeners($event): bool
         return isset($this->listeners[$event]) && $this->listeners[$event];
     }
 
-    public function addEventListener($events, $listener): void
+    public function addEventListener($events, object|string $listener): void
```
derrabus: Now that you've dropeed EventManager 1, you can match the parameter types from upstream.

```suggestion
    public function addEventListener(string|array $events, object|string $listener): void
```

Please check the other overridden methods too.

derrabus (changes requested):

src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
alexandre-daubois: That was a question I had in mind: should I add type to everything in this class?

Well I have my answer :smile:

UPGRADE-7.0.md
alexandre-daubois: Updated :+1:

derrabus (approved):

nicolas-grekas:
(rebase needed)

alexandre-daubois:
Rebased and updated with `DbalLogger` (and other) deprecations introduced in 6.4

derrabus:
FYI: DbalLogger is removed by #50571 already. But I'm fine either way, let's keep the change here and finish your PR first.

src/Symfony/Bridge/Doctrine/composer.json
```diff
@@ -17,7 +17,7 @@
     ],
     "require": {
         "php": ">=8.2",
-        "doctrine/event-manager": "^1.2|^2",
+        "doctrine/event-manager": "^2",
         "doctrine/persistence": "^2|^3",
         "symfony/deprecation-contracts": "^2.5|^3",
```
derrabus: I think we can remove deprecation-contracts from the bridge's dependencies now.

derrabus:
You need to adjust the expected-missing-return-types.diff file. You can borrow the necessary change from #50571.

src/Symfony/Bridge/Doctrine/CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
...
 =========
 
+7.0
+---
+
+ * Remove `DoctrineDbalCacheAdapterSchemaSubscriber`, `MessengerTransportDoctrineSchemaSubscriber` and `RememberMeTokenProviderDoctrineSchemaSubscriber`
+ * `ContainerAwareEventManager::getListeners()` must be called with an event name
+ * DoctrineBridge now requires `doctrine/event-manager:^2`
+ * Remove `DbalLogger`
+ * Remove `DoctrineDataCollector::addLogger()`, use a `DebugDataHolder` instead
```
derrabus: These removals should also be documented in UPGRADE-7.0.md

nicolas-grekas:
Thank you @alexandre-daubois.

stof:
@derrabus to avoid having to regenerate the expected-missing-return-types.diff file all the time in those code removal PRs, maybe we should first add those return types in 7.0 (by applying the patch defined in this diff file)

derrabus:
That's probably a good idea.

PR #50577, opened 2023-06-06
[HttpClient] Remove final keyword on `AsyncResponse`

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | kind of
| New feature?  | no
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

In symfony < 6.3, it was possible to override HTTP response headers. See this [blog post](https://jolicode.com/blog/aggressive-caching-with-symfony-http-client) for a use case

With Symfony 6.3, this is not possible anymore. I created a reproducer here: https://github.com/lyrixx/test/tree/http-cache-pp

And @nicolas-grekas came with another solution: https://github.com/lyrixx/test/issues/11

So here we go!

### DIFF
--- a/src/Symfony/Component/HttpClient/Response/AsyncResponse.php
+++ b/src/Symfony/Component/HttpClient/Response/AsyncResponse.php
@@ -27,7 +27,7 @@ use Symfony\Contracts\HttpClient\ResponseInterface;
  *
  * @author Nicolas Grekas <p@tchwork.com>
  */
-final class AsyncResponse implements ResponseInterface, StreamableInterface
+class AsyncResponse implements ResponseInterface, StreamableInterface
 {
     use CommonResponseTrait;
 


nicolas-grekas (approved):

stof:
Doesn't this risk breaking concurrency features of support for streaming chunks ? The override of headers is probably the only use case for which extending AsyncResponse is a _working_ solution.

So if we don't make AsyncResponse final anymore, maybe a bunch of its methods should become final instance, to prevent overriding them by thinking that it will work while it would break half of the ways the client can be used. Any of the methods dealing with the content cannot be used to implement additional logic without breaking the streamed usage.

nicolas-grekas:
Let's make getContent ~, toArray~ and toStream final? The rest should be fine being extended.

lyrixx:
@nicolas-grekas done ✅

src/Symfony/Component/HttpClient/Response/CommonResponseTrait.php
```diff
@@ -35,7 +35,7 @@ trait CommonResponseTrait
     private int $offset = 0;
     private ?array $jsonData = null;
 
-    public function getContent(bool $throw = true): string
+    final public function getContent(bool $throw = true): string
```
nicolas-grekas: this break tests :)
but honestly I'm not convinced this is usefull.
I'd vote for removing the keyword here and from toStream.

lyrixx: Done ✅

nicolas-grekas:
Thank you @lyrixx.

PR #50579, opened 2023-06-06
[DoctrineBridge] Deprecate using the old DBAL logger system

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | yes
| Tickets       | Backport from #50571
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/src/Symfony/Bridge/Doctrine/DataCollector/DoctrineDataCollector.php
+++ b/src/Symfony/Bridge/Doctrine/DataCollector/DoctrineDataCollector.php
@@ -43,15 +43,23 @@ class DoctrineDataCollector extends DataCollector
     ) {
         $this->connections = $registry->getConnectionNames();
         $this->managers = $registry->getManagerNames();
+
+        if (null === $debugDataHolder) {
+            trigger_deprecation('symfony/doctrine-bridge', '6.4', 'Not passing an instance of "%s" as "$debugDataHolder" to "%s()" is deprecated.', DebugDataHolder::class, __METHOD__);
+        }
     }
 
     /**
      * Adds the stack logger for a connection.
      *
      * @return void
+     *
+     * @deprecated since Symfony 6.4, use a DebugDataHolder instead.
      */
     public function addLogger(string $name, DebugStack $logger)
     {
+        trigger_deprecation('symfony/doctrine-bridge', '6.4', '"%s()" is deprecated. Pass an instance of "%s" to the constructor instead.', __METHOD__, DebugDataHolder::class);
+
         $this->loggers[$name] = $logger;
     }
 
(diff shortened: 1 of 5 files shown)


src/Symfony/Bridge/Doctrine/Logger/DbalLogger.php
```diff
@@ -28,6 +30,8 @@ class DbalLogger implements SQLLogger
 
     public function __construct(LoggerInterface $logger = null, Stopwatch $stopwatch = null)
     {
+        trigger_deprecation('symfony/doctrine-bridge', '6.4', '"%s" is deprecated, use a middleware instead.', __CLASS__);
```
nicolas-grekas: would it make sense to trigger this before the class declaration?

src/Symfony/Bridge/Doctrine/DataCollector/DoctrineDataCollector.php
```diff
@@ -43,15 +43,23 @@ public function __construct(
...
     /**
      * Adds the stack logger for a connection.
      *
      * @return void
+     *
+     * @deprecated since Symfony 6.4, use a DebugDataHolder instead.
      */
     public function addLogger(string $name, DebugStack $logger)
     {
+        trigger_deprecation('symfony/doctrine-bridge', '6.4', '"%s()" is deprecated. Pass an instance of "%s" to the constructor instead.', __METHOD__, DebugDataHolder::class);
```
nicolas-grekas, line 61: there's a call to this method in DoctrineDataCollectorWithDebugStackTest::createCollector
DoctrineDataCollectorWithDebugStackTest is `@group legacy` but is this legit?

stof: DoctrineDataCollectorWithDebugStackTest is covering the legacy case (DebugStack is the legacy logger-based implementation)

src/Symfony/Bridge/Doctrine/Logger/DbalLogger.php
derrabus: Yeah, probably.

src/Symfony/Bridge/Doctrine/DataCollector/DoctrineDataCollector.php
derrabus: Exactly, `DoctrineDataCollectorWithDebugStackTest` covers the workflow I'm deprecating here: The constructor is called without the second parameter, and instead `addLogger()` is called to add the deprecated `DebugStack` logger.

In #50571, I'm deleting the whole test.

src/Symfony/Bridge/Doctrine/Logger/DbalLogger.php
derrabus: Done.

stof (approved):

src/Symfony/Bridge/Doctrine/DataCollector/DoctrineDataCollector.php
derrabus: I've added expectations for the deprecation messages to the test you mentioned.

nicolas-grekas:
Thank you @derrabus.

PR #50582, opened 2023-06-06
[Security/Http] Fix false-string handling in `RememberMeAuthenticator`

ossinkine:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #... 
| License       | MIT

I found some errors "No remember-me cookie is found." in my logs. I didn't find another way to reproduce it other than set "false-string" to the cookie value, for example `0`.
This PR fixes this.


nicolas-grekas:
That's a bit strange as such values are unexpected in these cookies.
Can you confirm that after this patch the log lines are gone?

nicolas-grekas:
Thank you @ossinkine.

PR #50596, opened 2023-06-07
[DependencyInjection] Introduce "list" env var processor

amenophis:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | symfony/symfony-docs#18402

This MR allows to take an arbitrary list of values seperated by comma and parse it as an array with each item typed.
The implementation is just a hack that wrap the list with hooks and parse it as json

```
parameters:
    env(VALUES): 1,2.0,3.1,1e1
    values: '%env(list:VALUES)%' # array(4) { [0]=> int(1) [1]=> float(2) [2]=> float(3.1) [3]=> float(10) }
```

### DIFF
--- a/src/Symfony/Component/DependencyInjection/CHANGELOG.md
+++ b/src/Symfony/Component/DependencyInjection/CHANGELOG.md
@@ -22,6 +22,7 @@ CHANGELOG
  * Deprecate `#[MapDecorated]`, use `#[AutowireDecorated]` instead
  * Deprecate the `@required` annotation, use the `Symfony\Contracts\Service\Attribute\Required` attribute instead
  * Add `constructor` option to services declaration and to `#[Autoconfigure]`
+ * Add `list` env var processor
 
 6.2
 ---
--- a/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
+++ b/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
@@ -45,6 +45,7 @@ class EnvVarProcessor implements EnvVarProcessorInterface
             'file' => 'string',
             'float' => 'float',
             'int' => 'int',
+            'list' => 'array',
             'json' => 'array',
             'key' => 'bool|int|float|string|array',
             'url' => 'array',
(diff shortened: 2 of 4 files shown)


ro0NL:
IIUC this creates incompatible types:

https://github.com/symfony/symfony/blob/e1cc5f294c1029106824a0ad763fe3e2e6020d3c/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php#L46-L47

of which im not sure what the side effects are :sweat_smile:

nicolas-grekas:
I agree with @ro0NL, this would break expectations. Maybe with a new `int_array`? Or just filter on your side? You have to anyway since php arrays aren't typed...

ro0NL:
> since php arrays aren't typed

with psalm they are :)

this is why i prefer `json:`, to control the types by value, eg. `[1, "1", true]`

maybe for `csv:` auto casting some string values to php types (like `XmlUtils` does) is reasonable?

ro0NL:
another idea is to deprecate `csv:` in favor of new `list:` which could forward to json wrapped in brackets :)

amenophis:
@ro0NL Thanks for the suggestion, this way is more clean.
@nicolas-grekas I can deprecate the csv prefix, WDYT ?

nicolas-grekas:
Please don't deprecate the CSV processor, it fits a use case.

ro0NL:
i forgot to mention, we could detect unquoted strings in list and quote them before passing to json

but let's keep them separate :+1:

amenophis:
PR Doc added, both can be merged if OK for you @ro0NL and @nicolas-grekas :)

nicolas-grekas:
Looking at the code, I'm sorry but I don't see the point of adding this processor. Just use json if you have the choice of the format you put into your env var.

amenophis:
@nicolas-grekas thanks for the feedback, closing the PR

welcoMattic:
IIUC, this new `list` processor will wrap the list of values between `[]` to use json_decode. What if in the value, I have something like `foo[bar]`? Should I refer to the JSON syntax and adding `"` around this value? If yes, I agree with @nicolas-grekas, we could use `json` processor. I think it will be difficult (impossible?) to avoid to refer to the JSON syntax for values with special characters.

To assert that, could you add some tests with values containing `[` or `{` with and without wrapping around `"`?

ro0NL:
@welcoMattic `list: 1,2,3` and `json: [1,2,3]` would be equivalent, thus follow same error handling

my main drive here was to call a list a list :sweat_smile:, and compatibility with existing envs, while bringing json capabilities

TLDR: just use json :) but having `list: 1,"2",true,string,"also string"` would be noble :angel:

PR #50602, opened 2023-06-08
[Mailer] Make Response optional for AbstractApiTransport::doSendApi

VincentLanglet:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #... 
| License       | MIT

When sending an http/api request you might
- Sometimes want to not send any request for X reasons
- Use a custom httpClient to send the request

In both of these situations, there is no response to return.

Since the response is just used for debug purpose, returning the response should not be a requirement and could be optional.

### DIFF
--- a/src/Symfony/Component/Mailer/Transport/AbstractHttpTransport.php
+++ b/src/Symfony/Component/Mailer/Transport/AbstractHttpTransport.php
@@ -62,13 +62,13 @@ abstract class AbstractHttpTransport extends AbstractTransport
         return $this;
     }
 
-    abstract protected function doSendHttp(SentMessage $message): ResponseInterface;
+    abstract protected function doSendHttp(SentMessage $message): ?ResponseInterface;
 
     protected function doSend(SentMessage $message): void
     {
         try {
             $response = $this->doSendHttp($message);
-            $message->appendDebug($response->getInfo('debug') ?? '');
+            $message->appendDebug($response?->getInfo('debug') ?? '');
         } catch (HttpTransportException $e) {
             $e->appendDebug($e->getResponse()->getInfo('debug') ?? '');
 
--- a/src/Symfony/Component/Mailer/Transport/AbstractApiTransport.php
+++ b/src/Symfony/Component/Mailer/Transport/AbstractApiTransport.php
@@ -24,9 +24,9 @@ use Symfony\Contracts\HttpClient\ResponseInterface;
  */
 abstract class AbstractApiTransport extends AbstractHttpTransport
 {
-    abstract protected function doSendApi(SentMessage $sentMessage, Email $email, Envelope $envelope): ResponseInterface;
+    abstract protected function doSendApi(SentMessage $sentMessage, Email $email, Envelope $envelope): ?ResponseInterface;
 
-    protected function doSendHttp(SentMessage $message): ResponseInterface
+    protected function doSendHttp(SentMessage $message): ?ResponseInterface
     {
         try {
             $email = MessageConverter::toEmail($message->getOriginalMessage());


nicolas-grekas:
This is a potential BC break since extending classes would now need to handle a case that couldn't happen before.
I'm also not super convinced by the given rationale sorry.

VincentLanglet:
> This is a potential BC break since extending classes would now need to handle a case that couldn't happen before.

I was expecting to avoid any BC break, by the fact that
```
class A
{
     public function foo(): ?Response;
}
class B extends
{
    public function foo(): Response;
}
```
is valid.

> I'm also not super convinced by the given rationale sorry.

I was afraid of such reaction. I understand this can be considered as "edge case", on the other side, I find this could be too much constraint to ask to return a whole ResponseInterface when the only usage is
```
$message->appendDebug($response->getInfo('debug') ?? '');
```

Currently, I have to write 
```
return new MockResponse();
```
for my use cases, which kinda feel weird because of the name of the class. I may not be familiar with ReponseInterface implementation but I would have expect to be able to write something like `new Response()`.

src/Symfony/Component/Mailer/Transport/AbstractHttpTransport.php
```diff
@@ -62,13 +62,13 @@ public function setPort(?int $port): static
...
 
-    abstract protected function doSendHttp(SentMessage $message): ResponseInterface;
+    abstract protected function doSendHttp(SentMessage $message): ?ResponseInterface;
 
     protected function doSend(SentMessage $message): void
     {
         try {
             $response = $this->doSendHttp($message);
-            $message->appendDebug($response->getInfo('debug') ?? '');
+            $message->appendDebug($response?->getInfo('debug') ?? '');
```
nicolas-grekas, line 71: here is the BC break in practice; if you don't do this change: :boom: - any child class can be affected in the same way

VincentLanglet: Yes ; but only if Symfony implementations start to really returning `null` instead of a Response, which is not the case currently.
And the day someone start to use an implementation which returns null (by installing this library) they will have to support null, but this won't be considered as a BC break since they install a new lib.

stof: Once the return type of AbstractApiTransport say it is `ResponseInterface|null`, existing implementation would not see a BC break if they change a returned value while respecting this contract, as this would not change the contract.
The BC break happens when you change the contract (the fact that the runtime failure does not happen immediately in all case is a different matter).

VincentLanglet: Should I close this then or is something else possible ?

stof: To me, this should be closed. Not doing an HTTP request in the method `doSendHttp` does not make sense to me.

VincentLanglet: > To me, this should be closed. Not doing an HTTP request in the method `doSendHttp` does not make sense to me.

First, you can do an HttpRequest with something else than an HttpClientInterface which doesn't return an ResponseInterface (and there is no easy way to re-create something implementing the response interface)

Secondly, my issue was more about the `AbstractApiTransport`.
This abstract class provide useful things like
- an email param
- a getRecipients method

And again you can do API calls with custom SDK, which does not implements HttpClientInterface and therefore doesn't return a ResponseInterface.

stof: Well, `AbstractHttpTransport` depends on `symfony/http-client-contracts`. That's why it expects the ResponseInterface of that package. If you want to build things on your own, do it without the base class IMO.

> This abstract class provide useful things like
> 
>     * an email param
> 
>     * a getRecipients method

If you want this Email object without the parent `AbstractHttpTransport`, you could reuse `\Symfony\Component\Mime\MessageConverter::toEmail` directly (which is how this `$email` is obtained).

For `getRecipients`, this is just a helper filtering the Envelope recipients to remove the recipients coming from Cc and Bcc (as APIs will take those separately generally).

PR #50607, opened 2023-06-08
Run unit tests with PHP 8.3

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

It's that time of the year… PHP 8.3.0-alpha1 has been released, so we might start to test against that new version. Right now, it looks like PHP 8.3 will be a pretty boring release. The good kind of boring, of course!

This PR bumps the PHP 8.2 job to 8.3 and the high/low deps jobs from PHP 8.1 to 8.2. All dependency bumps in the component's composer.json files were done to fix test failures on the low-deps run.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -28,11 +28,11 @@ jobs:
         include:
           - php: '7.2'
           - php: '7.4'
-          - php: '8.1'
-            mode: high-deps
-          - php: '8.1'
-            mode: low-deps
           - php: '8.2'
+            mode: high-deps
+          - php: '8.2'
+            mode: low-deps
+          - php: '8.3'
             #mode: experimental
       fail-fast: false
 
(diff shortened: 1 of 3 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -28,12 +28,12 @@ jobs:
...
-          - php: '8.1'
+          - php: '8.2'
             mode: high-deps
-          - php: '8.1'
-            mode: low-deps
           - php: '8.2'
-            #mode: experimental
+            mode: low-deps
+          - php: '8.3'
+            mode: experimental
```
nicolas-grekas: Let's keep this commented out, it's green.

nicolas-grekas:
Thank you @derrabus.

.github/workflows/unit-tests.yml
derrabus: Done. But we will need to enable it again on 6.2 and above because of Laminas Code.

.github/workflows/unit-tests.yml
```diff
@@ -28,11 +28,11 @@ jobs:
         include:
           - php: '7.2'
           - php: '7.4'
-          - php: '8.1'
```
stof, line 31: Should we loose entirely the testing on 8.1 ?

derrabus: Yes, like we don't test with 7.3 and 8.0 either.

PR #50611, opened 2023-06-08
[Clock] Fix MockClock::modify() on PHP 8.3

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Symfony/Component/Clock/MockClock.php
+++ b/src/Symfony/Component/Clock/MockClock.php
@@ -49,7 +49,12 @@ final class MockClock implements ClockInterface
 
     public function modify(string $modifier): void
     {
-        if (false === $modifiedNow = @$this->now->modify($modifier)) {
+        try {
+            $modifiedNow = @$this->now->modify($modifier);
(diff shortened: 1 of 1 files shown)


src/Symfony/Component/Clock/MockClock.php
```diff
@@ -49,7 +49,12 @@ public function sleep(float|int $seconds): void
 
     public function modify(string $modifier): void
     {
-        if (false === $modifiedNow = @$this->now->modify($modifier)) {
+        try {
+            $modifiedNow = @$this->now->modify($modifier);
+        } catch (\DateMalformedStringException) {
```
stof, line 54: should we attach it as previous exception of the exception we throw ?

nicolas-grekas: I'm not sure: the exception doesn't contain anything useful

PR #46901, opened 2022-07-11
[Console] Be explicit about the completion API version

wouterj:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #44733 
| License       | MIT
| Doc PR        | n/a

Using the application version is wrong, especially for standalone apps like PHPStan. The completion API is not bound to change on a PHPStan version upgrade, but instead only when something in symfony/console changes.

I suggest having an explicit completion API version, that we can increase once we change something in the API (and thus require an update of the shell scripts).

PR #50613, opened 2023-06-09
[Console] Remove deprecations across the component

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

In short: preparing the component for 7.0 :slightly_smiling_face:

### DIFF
--- a/src/Symfony/Component/Console/Command/CompleteCommand.php
+++ b/src/Symfony/Component/Console/Command/CompleteCommand.php
@@ -34,16 +34,6 @@ final class CompleteCommand extends Command
 {
     public const COMPLETION_API_VERSION = '1';
 
-    /**
-     * @deprecated since Symfony 6.1
-     */
-    protected static $defaultName = '|_complete';
-
-    /**
-     * @deprecated since Symfony 6.1
-     */
-    protected static $defaultDescription = 'Internal command to provide shell completion suggestions';
-
     private $completionOutputs;
 
     private $isDebug = false;
--- a/src/Symfony/Component/Console/Tests/Command/CommandTest.php
+++ b/src/Symfony/Component/Console/Tests/Command/CommandTest.php
@@ -442,37 +442,6 @@ class CommandTest extends TestCase
         $this->assertSame(['f'], $command->getAliases());
     }
 
-    /**
-     * @group legacy
-     */
-    public function testDefaultNameProperty()
-    {
-        $this->expectDeprecation('Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "Symfony\Component\Console\Tests\Command\MyCommand" class instead.');
-
-        $this->assertSame('my:command', MyCommand::getDefaultName());
-    }
-
-    /**
-     * @group legacy
-     */
-    public function testDefaultDescriptionProperty()
-    {
-        $this->expectDeprecation('Since symfony/console 6.1: Relying on the static property "$defaultDescription" for setting a command description is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "Symfony\Component\Console\Tests\Command\MyCommand" class instead.');
-
-        $this->assertSame('This is a command I wrote all by myself', MyCommand::getDefaultDescription());
-    }
-
-    /**
-     * @group legacy
-     */
-    public function testStaticDefaultProperties()
-    {
-        $command = new MyCommand();
-
-        $this->assertSame('my:command', $command->getName());
-        $this->assertSame('This is a command I wrote all by myself', $command->getDescription());
-    }
-
     public function testAttributeOverridesProperty()
     {
         $this->assertSame('my:command', MyAnnotatedCommand::getDefaultName());
@@ -518,29 +487,10 @@ class Php8Command2 extends Command
 {
 }
 
-class MyCommand extends Command
-{
-    /**
-     * @deprecated since Symfony 6.1
-     */
-    protected static $defaultName = 'my:command';
-
-    /**
-     * @deprecated since Symfony 6.1
-     */
-    protected static $defaultDescription = 'This is a command I wrote all by myself';
-}
-
 #[AsCommand(name: 'my:command', description: 'This is a command I wrote all by myself')]
 class MyAnnotatedCommand extends Command
 {
-    /**
-     * @deprecated since Symfony 6.1
-     */
     protected static $defaultName = 'i-shall-be-ignored';
 
-    /**
-     * @deprecated since Symfony 6.1
-     */
     protected static $defaultDescription = 'This description should be ignored.';
 }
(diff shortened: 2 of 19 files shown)


src/Symfony/Bridge/Monolog/Command/ServerLogCommand.php
```diff
@@ -50,10 +50,7 @@ public function isEnabled(): bool
         return parent::isEnabled();
     }
 
-    /**
-     * @return void
-     */
-    protected function configure()
+    protected function configure(): void
```
nicolas-grekas: please revert changes to return types, we're going to deal with them separately

nicolas-grekas (commented):
let's finish existing PRs before opening new ones ;)

src/Symfony/Component/Runtime/composer.json
```diff
@@ -20,9 +20,10 @@
         "composer-plugin-api": "^1.0|^2.0"
     },
     "require-dev": {
-        "composer/composer": "^1.0.2|^2.0",
+        "composer/composer": "^2.6",
         "symfony/console": "^6.4|^7.0",
         "symfony/dotenv": "^6.4|^7.0",
+        "symfony/finder": "^6.4",
```
alexandre-daubois: This is necessary because [composer/class-map-generator](https://packagist.org/packages/composer/class-map-generator#dev-main) doesn't support `Symfony ^7.0` yet. The highest dep for the Finder must be set to `^6.4` because of this for now.

src/Symfony/Component/Runtime/composer.json
```diff
@@ -20,9 +20,10 @@
         "composer-plugin-api": "^1.0|^2.0"
     },
     "require-dev": {
-        "composer/composer": "^1.0.2|^2.0",
+        "composer/composer": "^2.6",
```
alexandre-daubois, line 23: [Version 2.6](https://packagist.org/packages/composer/composer#dev-main) brings support of Symfony 7.0

src/Symfony/Component/Runtime/composer.json
nicolas-grekas: please send a PR to class-map-generator instead

src/Symfony/Component/Console/Command/CompleteCommand.php
```diff
@@ -82,8 +71,7 @@ protected function initialize(InputInterface $input, OutputInterface $output): v
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         try {
-            // "symfony" must be kept for compat with the shell scripts generated by Symfony Console 5.4 - 6.1
-            $version = $input->getOption('symfony') ? '1' : $input->getOption('api-version');
+            $version = $input->getOption('api-version');
```
nicolas-grekas: What about this one @wouterj? should we keep support for the deprecated flag to provide a smooth transition?
Also, we're missing a deprecation notice if this should be removed in 7.0

src/Symfony/Component/Console/Tests/Command/CommandTest.php
```diff
@@ -442,48 +442,6 @@ public function testCommandAttribute()
...
-     */
-    public function testStaticDefaultProperties()
-    {
-        $command = new MyCommand();
-
-        $this->assertSame('my:command', $command->getName());
-        $this->assertSame('This is a command I wrote all by myself', $command->getDescription());
-    }
-
-    public function testAttributeOverridesProperty()
```
nicolas-grekas, line 476: this one is not legacy :thinking:

alexandre-daubois: That's right, but now that `$defaultName` and `$defaultDescription` properties are removed, it feels irrelevant to me to have this test. I guess this test should have been marked as legacy as well when the properties were deprecated, or maybe am I missing something?

alexandre-daubois: That's right, but now that `$defaultName` and `$defaultDescription` are removed, it feels irrelevant to me to have this test. I guess this should have been marked as legacy as well when the properties were deprecated? Or maybe am I missing something?

However, I may update this test with a command without the attribute and checking that the `getDefaultName()` and `getDefaultDescription()` are null

nicolas-grekas: We should still keep it to ensure no regressions on the topic. (Removing depreciation annotations on the fixture)

alexandre-daubois: That's reverted and updated :+1:

src/Symfony/Component/Runtime/composer.json
alexandre-daubois: It has been dealt by @stloyd [here](https://github.com/composer/class-map-generator/pull/4), thanks!

I'll revert the change once it's merged.

src/Symfony/Component/Console/Command/CompleteCommand.php
GromNaN: This was modified by #46901.

We need to keep the `symfony` option if we want developers to be able use to bash completion for different projects with different Symfony versions on the same computer. The completion script in `symfony/console` from 5.4.x to 6.1.x sets the option `-S` but not `-a`. If the option is removed, the `_complete` command will fail.
https://github.com/symfony/symfony/blob/a6e5b75ecdf1266ab019fbfd7042a5d6efab0c04/src/Symfony/Component/Console/Resources/completion.bash#L28

We can fallback to `api-version=1` when not specified.
```suggestion
            $version = $input->getOption('api-version', 1);
```

It would be better to make this change in a dedicated PR.

alexandre-daubois: If I understand correctly, here is what you suggest:
- Create a PR targeting 6.4 as a new feature, making `api-version` fallback to `1`
- Update `symfony` option deprecation to be removed in 8.0, when 5.4 is EOLed.

I'll take care of this if it is what you meant

wouterj: I think we should revert this change completely and leave it as-is. Notice the comment says "kept for compat" without deprecating the old one, which is intentional.

Shell completion scripts are global and there is no reason to make life extremely difficult for developers that work on several Symfony versions, some of which are older than 7.0, for the gain of removing these 2 LOC :)

alexandre-daubois: This suits me fine. As you said, these are just 2 lines of code :smile: I reverted this change

src/Symfony/Component/Runtime/composer.json
nicolas-grekas: Unlocked!

alexandre-daubois: I was going to update when I saw you updated the file. Thank you! :slightly_smiling_face:

nicolas-grekas:
Thank you @alexandre-daubois.

src/Symfony/Component/Console/Command/CompleteCommand.php
GWilson69: Do I have something to do with this?

PR #50578, opened 2023-06-06
[DependencyInjection] Remove deprecations across the component

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

 * Remove `#[MapDecorated]`, use `#[AutowireDecorated]` instead
 * Remove `ProxyHelper`, use `Symfony\Component\VarExporter\ProxyHelper` instead
 * Remove `ReferenceSetArgumentTrait`
 * Remove support of `@required` annotation, use the `Symfony\Contracts\Service\Attribute\Required` attribute instead
 * Passing `null` to `ContainerAwareTrait::setContainer()` must be done explicitly
 * Remove `PhpDumper` options `inline_factories_parameter` and `inline_class_loader_parameter`, use options `inline_factories` and `inline_class_loader` instead
 * Parameter names of `ParameterBag` cannot be numerics
 * Remove `ContainerAwareInterface` and `ContainerAwareTrait`, use dependency injection instead
 * Remove `ContainerAwareLoader`, use dependency injection in your fixtures instead

PR #50615, opened 2023-06-09
[DependencyInjection] Deprecate `ContainerAwareInterface`, `ContainerAwareTrait` and `ContainerAwareLoader`

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Tickets       | -
| License       | MIT
| Doc PR        | -

As discussed in https://github.com/symfony/symfony/pull/50578#discussion_r1224004179

### DIFF
--- a/UPGRADE-6.4.md
+++ b/UPGRADE-6.4.md
@@ -1,12 +1,18 @@
 UPGRADE FROM 6.3 to 6.4
 =======================
 
+DependencyInjection
+-------------------
+
+ * Deprecate `ContainerAwareInterface` and `ContainerAwareTrait`, use dependency injection instead
+
 DoctrineBridge
 --------------
 
  * Deprecate `DbalLogger`, use a middleware instead
  * Deprecate not constructing `DoctrineDataCollector` with an instance of `DebugDataHolder`
  * Deprecate `DoctrineDataCollector::addLogger()`, use a `DebugDataHolder` instead
+ * Deprecate `ContainerAwareLoader`, use dependency injection in your fixtures instead
 
 HttpFoundation
 --------------
--- a/src/Symfony/Bridge/Doctrine/Tests/DataFixtures/ContainerAwareLoaderTest.php
+++ b/src/Symfony/Bridge/Doctrine/Tests/DataFixtures/ContainerAwareLoaderTest.php
@@ -16,6 +16,9 @@ use Symfony\Bridge\Doctrine\DataFixtures\ContainerAwareLoader;
 use Symfony\Bridge\Doctrine\Tests\Fixtures\ContainerAwareFixture;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
+/**
+ * @group legacy
+ */
 class ContainerAwareLoaderTest extends TestCase
 {
     public function testShouldSetContainerOnContainerAwareFixture()
--- a/src/Symfony/Bundle/FrameworkBundle/Console/Application.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Console/Application.php
@@ -115,6 +115,7 @@ class Application extends BaseApplication
         $command = parent::get($name);
 
         if ($command instanceof ContainerAwareInterface) {
+            trigger_deprecation('symfony/dependency-injection', '6.4', 'Relying on "%s" to get the container in "%s" is deprecated, register the command as a service and use dependency injection instead.', ContainerAwareInterface::class, get_debug_type($command));
             $command->setContainer($this->kernel->getContainer());
         }
 
--- a/src/Symfony/Bundle/FrameworkBundle/Controller/ControllerResolver.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Controller/ControllerResolver.php
@@ -26,6 +26,7 @@ class ControllerResolver extends ContainerControllerResolver
         $controller = parent::instantiateController($class);
 
         if ($controller instanceof ContainerAwareInterface) {
+            trigger_deprecation('symfony/dependency-injection', '6.4', 'Relying on "%s" to get the container in "%s" is deprecated, register the controller as a service and use dependency injection instead.', ContainerAwareInterface::class, get_debug_type($controller));
             $controller->setContainer($this->container);
         }
         if ($controller instanceof AbstractController) {
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/Functional/app/ControllerServiceResolution/config.yml
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/Functional/app/ControllerServiceResolution/config.yml
@@ -5,6 +5,7 @@ services:
     Symfony\Bundle\FrameworkBundle\Tests\Functional\Bundle\TestBundle\Controller\SubRequestServiceResolutionController:
         public: true
         tags: [controller.service_arguments]
+        arguments: ['@service_container']
 
     logger: { class: Psr\Log\NullLogger }
     Psr\Log\LoggerInterface: '@logger'
--- a/src/Symfony/Component/DependencyInjection/ContainerAwareInterface.php
+++ b/src/Symfony/Component/DependencyInjection/ContainerAwareInterface.php
@@ -15,6 +15,8 @@ namespace Symfony\Component\DependencyInjection;
  * ContainerAwareInterface should be implemented by classes that depends on a Container.
  *
  * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @deprecated since Symfony 6.4, use dependency injection instead
  */
 interface ContainerAwareInterface
 {
--- a/src/Symfony/Component/DependencyInjection/ContainerAwareTrait.php
+++ b/src/Symfony/Component/DependencyInjection/ContainerAwareTrait.php
@@ -11,10 +11,14 @@
 
 namespace Symfony\Component\DependencyInjection;
 
+trigger_deprecation('symfony/dependency-injection', '6.4', '"%s" is deprecated, use dependency injection instead.', ContainerAwareTrait::class);
+
 /**
  * ContainerAware trait.
  *
  * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @deprecated since Symfony 6.4, use dependency injection instead
  */
 trait ContainerAwareTrait
 {
--- a/src/Symfony/Component/DependencyInjection/Tests/ContainerAwareTraitTest.php
+++ b/src/Symfony/Component/DependencyInjection/Tests/ContainerAwareTraitTest.php
@@ -13,17 +13,16 @@ namespace Symfony\Component\DependencyInjection\Tests;
 
 use PHPUnit\Framework\TestCase;
 use Symfony\Bridge\PhpUnit\ExpectDeprecationTrait;
-use Symfony\Component\DependencyInjection\ContainerAwareInterface;
-use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\Tests\Fixtures\ContainerAwareDummy;
 
+/**
+ * @group legacy
+ */
 class ContainerAwareTraitTest extends TestCase
 {
     use ExpectDeprecationTrait;
 
-    /**
-     * @group legacy
-     */
     public function testSetContainerLegacy()
     {
         $container = $this->createMock(ContainerInterface::class);
@@ -33,7 +32,7 @@ class ContainerAwareTraitTest extends TestCase
 
         self::assertSame($container, $dummy->getContainer());
 
-        $this->expectDeprecation('Since symfony/dependency-injection 6.2: Calling "Symfony\Component\DependencyInjection\Tests\ContainerAwareDummy::setContainer()" without any arguments is deprecated, pass null explicitly instead.');
+        $this->expectDeprecation('Since symfony/dependency-injection 6.2: Calling "Symfony\Component\DependencyInjection\Tests\Fixtures\ContainerAwareDummy::setContainer()" without any arguments is deprecated, pass null explicitly instead.');
 
         $dummy->setContainer();
         self::assertNull($dummy->getContainer());
@@ -50,15 +49,5 @@ class ContainerAwareTraitTest extends TestCase
 
         $dummy->setContainer(null);
         self::assertNull($dummy->getContainer());
-    }
-}
-
-class ContainerAwareDummy implements ContainerAwareInterface
-{
-    use ContainerAwareTrait;
-
-    public function getContainer(): ?ContainerInterface
-    {
-        return $this->container;
     }
 }
--- a/src/Symfony/Component/DependencyInjection/Tests/Fixtures/ContainerAwareDummy.php
+++ b/src/Symfony/Component/DependencyInjection/Tests/Fixtures/ContainerAwareDummy.php
@@ -1,0 +1,30 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\DependencyInjection\Tests\Fixtures;
+
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * @deprecated since Symfony 6.4, to be removed in 7.0
+ */
+class ContainerAwareDummy implements ContainerAwareInterface
+{
+    use ContainerAwareTrait;
+
+    public function getContainer(): ?ContainerInterface
+    {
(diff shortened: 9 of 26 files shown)


nicolas-grekas:
Thanks. We also need to scan the code base for all usages of both and adjust as needed.

alexandre-daubois:
Should this be done in this PR? I thought this would be done in the 7.0 one. But I can have a look in this one if needed :slightly_smiling_face:

nicolas-grekas:
That's always been our policy yes: ensure we don't trigger self-deprecation and that the code path that are going to be removed do trigger a notice before being removed.

src/Symfony/Component/DependencyInjection/ContainerAwareInterface.php
```diff
@@ -15,6 +15,8 @@
  * ContainerAwareInterface should be implemented by classes that depends on a Container.
  *
  * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @deprecated since Symfony 6.4, use dependency injection instead
```
nicolas-grekas, line 19: let's trigger a deprecation notice here also and adjust the code base as needed

src/Symfony/Component/DependencyInjection/ContainerAwareTrait.php
```diff
@@ -15,6 +15,8 @@
  * ContainerAware trait.
  *
  * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @deprecated since Symfony 6.4, use dependency injection instead
```
nicolas-grekas, line 21: we should throw a deprecation notice in addition to this

src/Symfony/Component/DependencyInjection/ContainerAwareInterface.php
nicolas-grekas: ok, we won't be able to trigger at runtime because BundleInterface extends ContainerAwareInterface
let's still see where this is used and adjust

alexandre-daubois: `BundleInterface` is the last place where the interface is implemented/extended in the codebase (removing it seems to keep tests green, good point for 7.0). I added a few lines to `Bundle` to get rid of `ContainerAwareTrait`

src/Symfony/Component/DependencyInjection/Tests/Fixtures/ContainerAwareDummy.php
```diff
@@ -0,0 +1,30 @@
...
+namespace Symfony\Component\DependencyInjection\Tests\Fixtures;
+
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * @deprecated since Symfony 6.4, to be removed in 7.0
+ */
+class ContainerAwareDummy implements ContainerAwareInterface
```
alexandre-daubois, line 21: Moved to a new file in order to get rid of the self-deprecation raised by the autoloader.

src/Symfony/Component/HttpKernel/Tests/DependencyInjection/RegisterControllerArgumentLocatorsPassTest.php
```diff
@@ -545,9 +543,14 @@ protected function barAction(ControllerDummy $bar)
     }
 }
 
-class ContainerAwareRegisterTestController implements ContainerAwareInterface
+class ContainerAwareRegisterTestController
 {
-    use ContainerAwareTrait;
+    protected ?ContainerInterface $container;
+
+    public function setContainer(ContainerInterface $container = null): void
```
chalasr, line 550: is it still called?

src/Symfony/Component/HttpKernel/Bundle/Bundle.php
```diff
@@ -149,4 +152,9 @@ private function parseClassName(): void
         $this->namespace = false === $pos ? '' : substr(static::class, 0, $pos);
         $this->name ??= false === $pos ? static::class : substr(static::class, $pos + 1);
     }
+
+    public function setContainer(?ContainerInterface $container): void
```
chalasr, line 156: should it be deprecated or is it meant to stay? same for the property?

src/Symfony/Bridge/Doctrine/Tests/DataFixtures/ContainerAwareLoaderTest.php
```diff
@@ -16,6 +16,11 @@
 use Symfony\Bridge\Doctrine\Tests\Fixtures\ContainerAwareFixture;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
+/**
+ * @group legacy
+ *
+ * @deprecated since Symfony 6.4, to be removed in 7.0
```
chalasr: can be removed as it is easy to spot once the deprecated thing it covers gets removed. Same for all fixtures/tests

chalasr (changes requested):

src/Symfony/Component/HttpKernel/Tests/DependencyInjection/RegisterControllerArgumentLocatorsPassTest.php
alexandre-daubois: It is actually checked in the pass:

https://github.com/symfony/symfony/blob/52a92926f7fed15cdff399c6921100a10e0d6f61/src/Symfony/Component/HttpKernel/DependencyInjection/RegisterControllerArgumentLocatorsPass.php#L80

I guess this should be kept regardless of deprecations introduced in this PR, as it was added when some checks on `ContainerAwareInterface` were removed in this commit https://github.com/symfony/symfony/commit/739789f384cca3b06a3b5f35793e116fff337473#diff-12c7c16828a1595f567c3a7e70a5678ee892171843c304b98a7760f6af1b2903

src/Symfony/Component/HttpKernel/Bundle/Bundle.php
alexandre-daubois: I would say it should stay. Some bundles are using this (like `FrameworkBundle`) in order to get container parameters or check service existence in the `boot()` method:

https://github.com/symfony/symfony/blob/52a92926f7fed15cdff399c6921100a10e0d6f61/src/Symfony/Bundle/FrameworkBundle/FrameworkBundle.php#L105

src/Symfony/Bridge/Doctrine/Tests/DataFixtures/ContainerAwareLoaderTest.php
alexandre-daubois: Alright, removed! 🙂

UPGRADE-6.4.md
```diff
@@ -1,12 +1,49 @@
...
+
+       // ...
+   }
+   ```
+
+   After:
+   ```php
+   class MyService
+   {
+       private ContainerInterface $container;
```
nicolas-grekas: let's use CPP

src/Symfony/Bundle/FrameworkBundle/Console/Application.php
```diff
@@ -115,6 +115,7 @@ public function get(string $name): Command
         $command = parent::get($name);
 
         if ($command instanceof ContainerAwareInterface) {
+            trigger_deprecation('symfony/dependency-injection', '6.4', sprintf('Command "%s" must not implement "%s", use dependency injection instead.', $command::class, ContainerAwareInterface::class));
```
nicolas-grekas:
```suggestion
            trigger_deprecation('symfony/dependency-injection', '6.4', sprintf('Relying on "%s" to get the container in "%s" is deprecated, register the command as a service and use dependency injection instead.', ContainerAwareInterface::class, get_debug_type($command)));
```

src/Symfony/Bundle/FrameworkBundle/Controller/ControllerResolver.php
```diff
@@ -26,6 +26,7 @@ protected function instantiateController(string $class): object
         $controller = parent::instantiateController($class);
 
         if ($controller instanceof ContainerAwareInterface) {
+            trigger_deprecation('symfony/dependency-injection', '6.4', sprintf('Implementing "%s" in controller "%s" is deprecated, use dependency injection instead.', ContainerAwareInterface::class, $controller::class));
```
nicolas-grekas:
```suggestion
            trigger_deprecation('symfony/dependency-injection', '6.4', sprintf('Relying on "%s" to get the container in "%s" is deprecated, register the controller as a service and use dependency injection instead.', ContainerAwareInterface::class, get_debug_type($controller)));
```

src/Symfony/Bundle/FrameworkBundle/Tests/Functional/app/ControllerServiceResolution/config.yml
```diff
@@ -5,6 +5,8 @@ services:
     Symfony\Bundle\FrameworkBundle\Tests\Functional\Bundle\TestBundle\Controller\SubRequestServiceResolutionController:
         public: true
         tags: [controller.service_arguments]
+        arguments:
+            $container: '@service_container'
```
nicolas-grekas:
```suggestion
        arguments: ['@service_container']
```

src/Symfony/Component/DependencyInjection/Tests/ContainerAwareTraitTest.php
```diff
@@ -13,17 +13,18 @@
...
 use Symfony\Bridge\PhpUnit\ExpectDeprecationTrait;
-use Symfony\Component\DependencyInjection\ContainerAwareInterface;
-use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\Tests\Fixtures\ContainerAwareDummy;
 
+/**
+ * @group legacy
+ *
+ * @deprecated since Symfony 6.4, to be removed in Symfony 7.0
```
nicolas-grekas:
```suggestion
 * @group legacy
```

src/Symfony/Component/DependencyInjection/Tests/ContainerAwareTraitTest.php
```diff
@@ -33,12 +34,15 @@ public function testSetContainerLegacy()
...
-        $this->expectDeprecation('Since symfony/dependency-injection 6.2: Calling "Symfony\Component\DependencyInjection\Tests\ContainerAwareDummy::setContainer()" without any arguments is deprecated, pass null explicitly instead.');
+        $this->expectDeprecation('Since symfony/dependency-injection 6.2: Calling "Symfony\Component\DependencyInjection\Tests\Fixtures\ContainerAwareDummy::setContainer()" without any arguments is deprecated, pass null explicitly instead.');
 
         $dummy->setContainer();
         self::assertNull($dummy->getContainer());
     }
 
+    /**
+     * @group legacy
+     */
```
nicolas-grekas:
```suggestion
```

src/Symfony/Component/HttpKernel/Bundle/Bundle.php
nicolas-grekas: We should keep it indeed. Which means we should also add the method to BundleInterface, via an `@method` for now.

alexandre-daubois: The `@method` annotation has been added to `BundleInterface`

src/Symfony/Bundle/FrameworkBundle/Console/Application.php
alexandre-daubois: Updated! What are the pros on using `get_debug_type` over `::class`?

nicolas-grekas: Better output for anonymous classes.
Which means get_debug_type should *always* be used when dealing with messages for humans.

alexandre-daubois: Got it, thanks!

UPGRADE-6.4.md
```diff
@@ -1,12 +1,47 @@
...
+   ```
+
+   After:
+   ```php
+   class MyService
+   {
+       // Inject the container through the constructor...
+       public function __construct(private ContainerInterface $container)
+       {
+           $this->container = $container;
```
nicolas-grekas: cpp rulez

nicolas-grekas (approved):

UPGRADE-6.4.md
alexandre-daubois: Thinking of it again, isn't having CCP strange with the below example of the `#[Required]` attribute?

nicolas-grekas: why? Those are decorrelated

alexandre-daubois: It's true that it's decorrelated, it just made me feel like we were using 2 types of injection at the same time to write this example this way. But I'm aware it's a very subjective feeling :slightly_smiling_face:

src/Symfony/Component/HttpKernel/Bundle/Bundle.php
stof: Actually, as BundleInterface extends the ContainerAwareInterface currently, the method is already required for implementations. So we can add it natively rather than through `@method`

alexandre-daubois: I agree, but isn't it more comfortable to have `@method` with `Not implementing it is deprecated since Symfony 6.4.` next to it for the migration path?

But we could as well add this mention on top of the "native" method. I'll take care of removing it in https://github.com/symfony/symfony/pull/50578 (I couldn't find any example of this in the current codebase, only `@method` examples).

stof: it is **impossible** not to implement it in 6.4. As BundleInterface extends ContainerAwareInterface, the method is **already** part of the BundleInterface (as an inherited method).

that's not a case of adding a new method in BundleInterface but of moving it down

alexandre-daubois: Understood, I added the method prototype to `BundleInterface`

src/Symfony/Component/HttpKernel/Bundle/BundleInterface.php
```diff
@@ -66,4 +67,11 @@ public function getNamespace(): string;
      * The path should always be returned as a Unix path (with /).
      */
     public function getPath(): string;
+
+    /**
+     * Sets the container.
+     *
+     * @return void
```
nicolas-grekas:
```suggestion
     * @return void
```

nicolas-grekas (approved):
(I pushed the fixed .github/expected-missing-return-types.diff for you)

Kocal:
What is the upgrade path for apps/bundles using this interface/trait? eg: the DoctrineBundle for container-aware migrations.

ro0NL:
@Kocal we've implemented `Doctrine\Migrations\Version\MigrationFactory`

UPGRADE-6.4.md
```diff
@@ -1,12 +1,46 @@
...
+       }
+
+       // ... or by using the #[Required] attribute
+       #[Required]
+       public function setContainer(ContainerInterface $container): void
+       {
+            $this->container = $container;
+       }
+   }
+   ```
```
chalasr: Might be worth mentioning that injecting the DIC itself is discouraged

chalasr (approved):
🎉

UPGRADE-6.4.md
nicolas-grekas: maybe we should drop those examples? I don't think it's common to use them

alexandre-daubois: I'd be fine dropping the examples. Injecting the container itself _seems_ to mainly occur in some legacy projects, I didn't see myself places where the container is directly injected in "recent" codebases I faced.

About discouraging injecting the container, maybe it would be better to mention it somewhere in the documentation?

chalasr: Dropping the example sounds good enough. It has been said a lot already :)

alexandre-daubois: Dropped :+1:

src/Symfony/Bundle/FrameworkBundle/Console/Application.php
```diff
@@ -115,6 +115,7 @@ public function get(string $name): Command
         $command = parent::get($name);
 
         if ($command instanceof ContainerAwareInterface) {
+            trigger_deprecation('symfony/dependency-injection', '6.4', sprintf('Relying on "%s" to get the container in "%s" is deprecated, register the command as a service and use dependency injection instead.', ContainerAwareInterface::class, get_debug_type($command)));
```
ro0NL: https://github.com/symfony/symfony/blob/5bc63ada903433f5eaad469c42dfb019e4e9197c/src/Symfony/Contracts/Deprecation/function.php#L16-L19

;)

alexandre-daubois: Good catch! Thanks :slightly_smiling_face:

nicolas-grekas:
Thank you @alexandre-daubois.

PR #47071, opened 2022-07-26
AbstractSessionHandler implements SessionIdInterface

d3n972:
| Q             | A
| ------------- | ---
| Branch?       | 6.2  
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Tickets       | Fix #45186 
| License       | MIT
| Doc PR        | none 

Add `SessionIdInterface` as implemented interface to `AbstractSessionHandler` to create the possibility of creating/setting the correct session id if it is provided in an alternative manner. For example, you can implement a custom `create_sid()` method in a SessionHandler class which uses the  `AbstractSessionHandler`.

PR #24523, opened 2017-10-11
[HttpFoundation] Make sessions secure and lazy

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 3.4
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | not yet
| Fixed tickets | #6388, #6036, #12375, #12325
| License       | MIT
| Doc PR        | -

The `SessionUpdateTimestampHandlerInterface` (new to PHP 7.0) is mostly undocumented, and just not implemented anywhere. Yet, it's required to implement session fixation preventions and lazy write in userland session handlers (there is https://wiki.php.net/rfc/session-read_only-lazy_write which describes the behavior.)

By implementing it, we would make Symfony session handling much better and stronger. Meanwhile, doing some cookie headers management, this also gives the opportunity to fix the "don't start if session is only read issue".

So, here we are for the general idea. Now needs more (and green) tests, and review of course.

PR #24338, opened 2017-09-26
[HttpFoundation] Removed compatibility layer for PHP <5.4 sessions

afurculita:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?  | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

This is a follow-up of https://github.com/symfony/symfony/pull/24239. This PR removes the compatibility layer added for sessions for PHP <5.4.

PR #24239, opened 2017-09-17
[HttpFoundation] Deprecate compatibility with PHP <5.4 sessions

afurculita:
| Q             | A
| ------------- | ---
| Branch?       | 3.4
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

This PR removes functionality added in Symfony 2.1 as a compatibility layer with sessions from PHP <5.4.

- [x] Fix tests

PR #50626, opened 2023-06-10
[HttpFoundation] Make `SessionHandlerProxy` implement `SessionIdInterface`

### LINKED ISSUE #45186: Symfony Session Handlers should implement SessionIdInterface interface
### Description

AbstractSessionHandler should implement the interface SessionIdInterface.

Use case:

I have a small AP and I need to maintain a session without using cookies. In my case, the client passes the session ID in a custom HTTP header.

Currently I use an event subscriber to change the id of the session to the desired header value:

```php
public function onKernelRequest(RequestEvent $event): bool
{
        $request = $event->getRequest();
        $session = $event->getRequest()->getSession();
        $session->setId($request->headers->get('my-custom-header'));

        return true;
}
```

This more or less works in most of the cases and fails if the session has already been started.

I think that is what the PHP interface SessionIdInterface (https://www.php.net/manual/en/class.sessionidinterface.php) is for:

```php
interface SessionIdInterface {
    /* Methods */
    public create_sid(): string;
}
```

With this, I could implement my own create_sid() method, thus setting the correct session id from the begining.

### Example

_No response_

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #45186 and #50583
| License       | MIT
| Doc PR        | Not yet

Supersedes #47071

It is currently not easy to provide its own session ids because even if your handler implements `SessionIdInterface`, it will be wrapped in a `SessionHandlerProxy` which does not, so your `create_sid` method won’t be called.

### DIFF
--- a/src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Proxy/SessionHandlerProxyTest.php
+++ b/src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Proxy/SessionHandlerProxyTest.php
@@ -168,8 +168,26 @@ class SessionHandlerProxyTest extends TestCase
             [new SessionHandlerProxy(new StrictSessionHandler(new \SessionHandler()))],
         ];
     }
+
+    public function testCreateSid()
+    {
+        $mock = $this->createMock(SessionIdSessionHandler::class);
+        $mock->expects($this->once())
+            ->method('create_sid')
+            ->willReturn('a-valid-session-identifier');
+
+        $proxy = new SessionHandlerProxy($mock);
+        $this->assertSame('a-valid-session-identifier', $proxy->create_sid());
+
+        $this->proxy->create_sid();
+        $this->addToAssertionCount(1);
+    }
 }
 
 abstract class TestSessionHandler implements \SessionHandlerInterface, \SessionUpdateTimestampHandlerInterface
+{
+}
+
+abstract class SessionIdSessionHandler implements \SessionHandlerInterface, \SessionIdInterface
 {
 }
--- a/src/Symfony/Component/HttpFoundation/CHANGELOG.md
+++ b/src/Symfony/Component/HttpFoundation/CHANGELOG.md
@@ -9,7 +9,8 @@ CHANGELOG
  * Add `UriSigner` from the HttpKernel component
  * Add `partitioned` flag to `Cookie` (CHIPS Cookie)
  * Add argument `bool $flush = true` to `Response::send()`
-* Make `MongoDbSessionHandler` instantiable with the mongodb extension directly
+ * Make `MongoDbSessionHandler` instantiable with the mongodb extension directly
+ * Make `SessionHandlerProxy` implement `SessionIdInterface`
 
 6.3
 ---
--- a/src/Symfony/Component/HttpFoundation/Exception/SessionIdCreationException.php
+++ b/src/Symfony/Component/HttpFoundation/Exception/SessionIdCreationException.php
@@ -1,0 +1,20 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\HttpFoundation\Exception;
+
+class SessionIdCreationException extends \RuntimeException
+{
+    public function __construct(string $message = 'Could not create a session id.', int $code = 0, \Throwable $previous = null)
+    {
+        parent::__construct($message, $code, $previous);
+    }
+}
--- a/src/Symfony/Component/HttpFoundation/Session/Storage/Proxy/SessionHandlerProxy.php
+++ b/src/Symfony/Component/HttpFoundation/Session/Storage/Proxy/SessionHandlerProxy.php
@@ -11,12 +11,13 @@
 
 namespace Symfony\Component\HttpFoundation\Session\Storage\Proxy;
 
+use Symfony\Component\HttpFoundation\Exception\SessionIdCreationException;
 use Symfony\Component\HttpFoundation\Session\Storage\Handler\StrictSessionHandler;
 
 /**
  * @author Drak <drak@zikula.org>
  */
-class SessionHandlerProxy extends AbstractProxy implements \SessionHandlerInterface, \SessionUpdateTimestampHandlerInterface
+class SessionHandlerProxy extends AbstractProxy implements \SessionHandlerInterface, \SessionUpdateTimestampHandlerInterface, \SessionIdInterface
 {
     protected $handler;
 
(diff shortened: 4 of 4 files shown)


src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Proxy/SessionHandlerProxyTest.php
```diff
@@ -179,8 +179,25 @@ public static function provideNativeSessionStorageHandler()
...
+    {
+        $mock = $this->createMock(SessionIdSessionHandler::class);
+        $mock->expects($this->once())
+            ->method('create_sid');
+
+        $proxy = new SessionHandlerProxy($mock);
+        $proxy->create_sid();
+
+        $this->proxy->create_sid();
+        $this->addToAssertionCount(1);
```
nicolas-grekas, line 183: let's assert that both return the same value?

MatTheCat: `$proxy->create_sid()` would return the mock’s `create_sid` and `$this->proxy->create_sid()` would return the result of `session_create_id()` so they’re not supposed to return the same value?

MatTheCat:
There are other decorating handlers but I don’t know if it makes sense to make them all implement `SessionIdInterface` :thinking: 

- `MarshallingSessionHandler`
- `MigratingSessionHandler`
- `StrictSessionHandler`

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Proxy/SessionHandlerProxyTest.php
stof: as the mock implements the interface, the proxy is expected to return the sid created by the mock. Your test does not enforce that.

MatTheCat: Oh okay I added an assertion :+1:

deepsidhu1313:
Does this suppose to help setting the a custom session id using `setId($some_id)` function of `Symfony\Component\HttpFoundation\Session\Session` ??

nicolas-grekas:
I think all decorators should implement this, isn't it?

nicolas-grekas (dismissed):

MatTheCat:
Woops missed your comment.

> I think all decorators should implement this, isn't it?

I guess they could :thinking: else you’re required to put your `SessionIdInterface` on top of the stack. I guess this isn’t an issue since you approved this PR?

MatTheCat:
Eeer after reading https://symfony.com/doc/current/session.html#session-proxies I wonder if this PR makes sense as you’re supposed to extend `SessionHandlerProxy` rather than decorate the inner handler.

MatTheCat:
Okay, just tried to do that and I got a `ServiceCircularReferenceException`… Does the issue come from Symfony or the documentation?

nicolas-grekas:
I'm a bit lost between the handler, the proxy, etc. I'll let you investigate if you don't mind :)

MatTheCat:
I’m still lost but I have some leads: session proxies purpose was originally to ease the migration from PHP 5.3 where `SessionHandlerInterface` did not exist. They were [deprecated](https://github.com/symfony/symfony/pull/24239) then [removed](https://github.com/symfony/symfony/pull/24338), but `AbstractProxy` and `SessionHandlerProxy` escaped death thanks to #24523 because the latter would [turns any handler into an instance of `AbstractSessionHandler`](https://github.com/symfony/symfony/pull/24523#issuecomment-336170029). As it doesn’t seem to be the case maybe I’m misunderstanding something :thinking: 

The `SessionHandlerProxy` always has been documented as an extension point (like in https://symfony.com/doc/current/session.html#session-proxies). I don’t know if it used to work at some point, but now doing as documented results in a `ServiceCircularReferenceException` because the `framework.session.handler_id` service is aliased by `session.handler`, which is aliased by `SessionHandlerInterface` (you can fix this issue by wiring your wrapped handler yourself though). Also by doing this you couldn’t use decorators like the `MarshallingSessionHandler`.

That’s why I thought you were supposed to decorate `session.handler` instead, which requires less configuration but more boilerplate code to proxy all of `SessionHandlerInterface`, `SessionUpdateTimestampHandlerInterface` (and `SessionIdInterface`?) methods.

MatTheCat:
Closing, since there already is a way to configure a `SessionIdInterface` handler.

jdevinemt:
I have been struggling for most of the day trying to figure out why the create_sid() method was never being called from my custom session handler. I finally found it was because even though I am specifying a class that implements `\SessionHandlerInterface, \SessionIdInterface, \SessionUpdateTimestampHandlerInterface` as the `framework.session.handler_id` value, it is being replaced/proxied by the `\Symfony\Component\HttpFoundation\Session\Storage\Proxy\SessionHandlerProxy` class. Which, as mentioned in this PR does not implement the `\SessionIdInterface`.

I think it would be beneficial to reconsider implementing this.

@MatTheCat Is right that you can decorate the `SessionHandlerProxy` to implement `\SessionIdInterface`. But I think it is more straightforward to have the `\SessionIdInterface` implemented in `SessionHandlerProxy` as initially submitted in this PR. Then we have an implementation of `SessionHandlerProxy` that handles all possible PHP session interfaces, which could help prevent the assumptions that having a handler with `\SessionIdInterface` being wrong.

If it is decided that it's best not to implement this, we may want to consider deprecating passing a service that does not extend `SessionHandlerProxy` as a value to `framework.session.handler_id` , or at least update the documentation to discourage using a handler that isn't an instance of that class.

jdevinemt:
> The SessionHandlerProxy always has been documented as an extension point (like in https://symfony.com/doc/current/session.html#session-proxies). I don’t know if it used to work at some point, but now doing as documented results in a ServiceCircularReferenceException because the framework.session.handler_id service is aliased by session.handler, which is aliased by SessionHandlerInterface (you can fix this issue by wiring your wrapped handler yourself though). Also by doing this you couldn’t use decorators like the MarshallingSessionHandler.

This is another big reason that this PR should be reconsidered. It's awkward to work around the current implementation due to the circular dependency @MatTheCat mentioned above. You have to either have two classes, one for the majority of the custom handler implementation that implements `\SessionHandlerInterface` and `\SessionUpdateTimestampHandlerInterface`. And another class that extends `SessionHandlerProxy` and implements `\SessionIdInterface`. Then you have to explicitly wire the main handler to the proxied handler doing something like this:

```yaml
framework:
  session:
    enabled: true
    storage_factory_id: session.storage.factory.native
    handler_id: App\SessionHandlerProxy

services:
  App\SessionHandlerProxy:
    arguments:
      $handler: App\SessionHandler
```

You could implement all the handler functionality to the `App\SessionHandlerProxy`, but it still needs a `\SessionHandlerInterface` in it's constructor. You could pass something like the native handler, but then if you forget to override on of the session handler methods, it would defer to that handler, which just adds confusion and may cause unexpected behavior.

MatTheCat:
Hey @jdevinemt 👋

Glad to see you concerned, but closed PRs don't usually generate traction; opening an issue may be better.

jdevinemt:
Thanks @MatTheCat. I had a feeling that might be the case. I just finished my implementation to work around this restriction in my current project. I'm working through writing tests to cover my handler and handler proxy services. Once I'm done with that I'll open a new issue and create a PR if I can figure out the best way to implement this interface.

PR #49726, opened 2023-03-18
[HttpFoundation] Add IpUtils::isPrivateIp

danielburger1337:
| Q             | A
| ------------- | ---
| Branch?       | 6.3 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Tickets       | N/A
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/18102

This is only my second PR for this project, so I hope I followed all the guidelines correctly.
Recently I had more and more use cases where I had to make exceptions (mostly rate limiting) for private IP ranges.

Symfony currently does not provide an easy way to check if an IP is private or public but implements such logic internally (private constant) for the [NoPrivateNetworkHttpClient](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HttpClient/NoPrivateNetworkHttpClient.php).

This PR intents to make the private subnet list reusable by adding it to [IpUtils](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HttpFoundation/IpUtils.php).

In the original PR of the NoPrivateNetworkHttpClient it was also briefly mentioned that this constant may have value when made public. https://github.com/symfony/symfony/pull/35566#discussion_r374111850

I think symfony should and always should have exposed this constant.

PR #50633, opened 2023-06-12
[HttpClient] Postpone symfony/http-foundation < 6.3 conflict

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

I'd like to postpone the conflict introduced in https://github.com/symfony/symfony/pull/49726 until 7.0.

The reason is that it was added for a very tiny particular case but it prevents bumping symfony/http-client to 6.3 for all Laravel users.
The current Laravel major version requires symfony/http-foundation:^6.2 and it won't change until February 2024 (it should be ^7.0) by then.

### DIFF
--- a/src/Symfony/Component/HttpClient/NoPrivateNetworkHttpClient.php
+++ b/src/Symfony/Component/HttpClient/NoPrivateNetworkHttpClient.php
@@ -30,6 +30,21 @@ final class NoPrivateNetworkHttpClient implements HttpClientInterface, LoggerAwa
 {
     use HttpClientTrait;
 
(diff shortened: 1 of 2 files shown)


fancyweb:
Or maybe we can just maintain the duplicated list because when is it gonna change?

stof:
`^6.2` means `>=6.2 <7.0`. So it allows installing version 6.3.

fancyweb:
Indeed sorry :man_facepalming:

PR #48457, opened 2022-12-03
[FrameworkBundle] Improve UX ConfigDebugCommand has not yaml component

alamirault:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Tickets       | Fix #48429
| License       | MIT
| Doc PR        | symfony/symfony-docs#... 

Throw exception in order to improve UX when Yaml component is missing 

Message is inspired by https://github.com/symfony/symfony/blob/043257f6fc1fceb28bc0c37c59f2bea4d9c8d79d/src/Symfony/Component/Serializer/Encoder/YamlEncoder.php#L56

PR #50637, opened 2023-06-12
[FrameworkBundle] Fixed parsing new JSON output of debug:config not possible

Toflar:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

See my comment to the original PR which introduced `debug:config --format=json`: https://github.com/symfony/symfony/pull/48457#issuecomment-1534780396

Currently, it's not possible to parse the JSON output (neither is the YAML) which makes `--format=json` kind of pointless. The argument against this - I guess - is that the JSON format was introduced to make the command independent from the YAML component but still, why is `--format` an option then?

So I guess by delivering an additional option, we can fix this issue and make it parsable.

(Maybe instead of having `--no-tilte` we should call it `parsable` in the first place so that if in the future there's something on top of the title, the option would still be valid?)

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Command/ConfigDebugCommand.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Command/ConfigDebugCommand.php
@@ -49,7 +49,7 @@ class ConfigDebugCommand extends AbstractConfigCommand
                 new InputArgument('name', InputArgument::OPTIONAL, 'The bundle name or the extension alias'),
                 new InputArgument('path', InputArgument::OPTIONAL, 'The configuration option path'),
                 new InputOption('resolve-env', null, InputOption::VALUE_NONE, 'Display resolved environment variable values instead of placeholders'),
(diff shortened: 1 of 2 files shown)


stof:
The title should be only for the text format IMO

Toflar:
There is no text format. The default is `yaml` and it hasn't been parsable ever since. Now in 6.3, there's a new fallback to `json` in case the Yaml component is not installed. But there is also a new `--format` option.

stof:
then maybe the title should be removed or be an info written to stderr instead of stdout

Toflar:
I don't really see any value in the title either but I don't feel like I'm in charge to decide on this 😄

ro0NL:
maybe show title if --format is not given

or put different, the default format can be new `text` which shows title+yaml, but --format=yaml shows no title

Toflar:
I like that suggestions so I've quickly implemented a first version - wdyt?

nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @Toflar.

PR #50640, opened 2023-06-12
[FrameworkBundle][Routing] Throw error on debug:router when requirement matching no variable in path regex

### LINKED ISSUE #50461: Check for unused routing requirements
### Description

It would be great if we can detect that parameter requirement does not match path parameters.

```yaml
foo_url:
  path:      /foo-url/{clientId}
  methods:   [ GET ]
  defaults:  { _controller: App\Controller\FooController }
  requirements:
    client: '\d+' // <------------- client does not exist, good name is clientId
```

In this case `client` parameter does not exist, so a deprecation or a warning will avoid validation error. 

WDYT ?

### Example

_No response_

mbuliard:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #50461 
| License       | MIT
| Doc PR        | -

Hello,
In the current state of Routing, a requirement matching no variable in the path is simply ignored.
This may lead to misconfiguration : if a typo is made in the requirement name, it will just be ignored, without any warning.

I am not sure where an error should appear. On compilation, it will break projects having a misconfiguration. To do lesser damage, I think of throwing it when debugging an only route, but I fear it may not be really useful then.
I you have any review/advice, I am looking for it ! :-)

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Command/RouterDebugCommand.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Command/RouterDebugCommand.php
@@ -23,6 +23,7 @@ use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\HttpKernel\Debug\FileLinkFormatter;
+use Symfony\Component\Routing\Route;
 use Symfony\Component\Routing\RouteCollection;
 use Symfony\Component\Routing\RouterInterface;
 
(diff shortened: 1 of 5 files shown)


danielburger1337:
As i mentioned here https://github.com/symfony/symfony/issues/50461#issuecomment-1585712749):

This breaks with controller level annotations.

mbuliard:
> As i mentioned here [#50461 (comment)](https://github.com/symfony/symfony/issues/50461#issuecomment-1585712749)):
> 
> This breaks with controller level annotations.

Thank you for your reply. Sorry, I missed your comment on the issue.
I'm afraid you are right. If not doing it when compiling the route, I see no easy way to check the method-level requirements while ignoring the class-level ones.

PR #50641, opened 2023-06-12
[HttpClient] Add "no_private_network" http client configuration option

danielburger1337:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Tickets       | n/a
| License       | MIT
| Doc PR        | In Progress

Added ability to configure the `NoPrivateNetworkHttpClient` globally.
I did not add this feature to the `scoped_clients` sections because the destination networks of them SHOULD always be known.

I have an application where it sends requests to mostly user supplied URLs and was not satisified with the dx of how I had to create the decorator in the services.yaml file.

Doc PR is WIP because a) I'm not sure how likely this feature is to get merged and b) I'm not yet sure how to formulate the documenation.

For the reviewers: Please pay extra attention to the XML-Config. All tests pass locally and it "looks right", but I usually dont work with XSD.


nicolas-grekas:
Why don't you hardcode the decorator in your service instead?

```php
function __construct(HttpClientInterface $client)
{
	$this->client = new NoPrivateNetworkHttpClient($client, etc);
}
```

danielburger1337:
That's how I handle it "normally". 

But this app has 6 different Services which request user supplied URLs. 

That's why I configured it globally.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1864,6 +1864,21 @@ private function addHttpClientSection(ArrayNodeDefinition $rootNode, callable $e
...
                                 ->end()
+                                ->arrayNode('no_private_network')
+                                    ->fixXmlConfig('subnet')
+                                    ->canBeEnabled()
+                                    ->addDefaultsIfNotSet()
+                                    ->children()
+                                        ->arrayNode('subnets')
+                                            ->scalarPrototype()->end()
+                                            ->beforeNormalization()
+                                                ->ifString()
```
javiereguiluz:
```php
->ifString()
->then(fn (string $n): array => [$n])
```

maybe can be replaced by the following?

```php
->castToArray()
```

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2454,6 +2457,20 @@ private function registerHttpClientConfiguration(array $config, ContainerBuilder
             $this->registerRetryableHttpClient($retryOptions, 'http_client', $container);
         }
 
+        if ($this->readConfigEnabled('http_client.no_private_network', $container, $noPrivateNetworksOptions)) {
+            $subnets = $noPrivateNetworksOptions['subnets'];
+            if (0 === \count($subnets)) {
```
nicolas-grekas, line 2462:
```suggestion
            if (!$subnets = $noPrivateNetworksOptions['subnets']) {
```

nicolas-grekas (commented):
I'm not sure the implementation makes sense to me: it currently allows configuring all clients with the filter, which means once this is enabled, the app cannot hit private servers anymore.

Didn't you want to enable this only for a few specific scoped clients instead?

I feel like we'd better not merge this. The repetition of a one liner look OK to me, compared to adding complexity in the config. (a trait on your side could maybe help, or an abstract class if it makes sense, etc).

danielburger1337:
Totally get your point. 

But how many apps actually access private http servers? I don't think too many actually do.

Maybe my use case is too specific (that my app will never request a private server).

Anyway, I'm ok if this doesn't get merged.

nicolas-grekas:
Let's close, thanks for proposing.

PR #50644, opened 2023-06-13
[VarDumper] Dumping DateTime throws error if getTimezone is false

bogdanmoza:
| Q             | A
| ------------- | ---
| Branch? | 5.4
| Bug fix? | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        |  
----
**TL;DR:**
In case you have a DateTime object that returns false when calling `getTimezone()`, an error is thrown in DateCaster.

**Description**
So I have encountered this case when I have tried to dump a mocked DateTime object. The mock will not have a timezone making the `getTimezone()` method return false. I know that this is not really a bug, as a DateTime object not having a DateTimeZone should not be happening, but it is possible in case someone extends the DateTime object and overrides the `getTimezone()` method.
The error thrown is 'The DatetimeInterface object has not been correctly initialized by its constructor'

**How to reproduce**
```
class A extends \DateTime {
    public function getTimezone(): \DateTimeZone|false
    {
        return false;
    }
}
```
Dump the A class will cause an error.

**Final notes**
Even though this a special case, I still think that you should be able to dump a mocked DateTime.
I hope the description and everything is ok as this is my first pr here.

Cheers!

### DIFF
--- a/src/Symfony/Component/VarDumper/Caster/DateCaster.php
+++ b/src/Symfony/Component/VarDumper/Caster/DateCaster.php
@@ -27,7 +27,7 @@ class DateCaster
     public static function castDateTime(\DateTimeInterface $d, array $a, Stub $stub, bool $isNested, int $filter)
     {
         $prefix = Caster::PREFIX_VIRTUAL;
-        $location = $d->getTimezone()->getLocation();
+        $location = $d->getTimezone() ? $d->getTimezone()->getLocation() : null;
         $fromNow = (new \DateTime())->diff($d);
 
         $title = $d->format('l, F j, Y')
--- a/src/Symfony/Component/VarDumper/Tests/Caster/DateCasterTest.php
+++ b/src/Symfony/Component/VarDumper/Tests/Caster/DateCasterTest.php
@@ -90,6 +90,52 @@ EODUMP;
         ];
     }
 
+    /**
+     * @dataProvider provideNoTimezoneDateTimes
+     */
+    public function testCastDateTimeNoTimezone($time, $xDate, $xInfos)
+    {
+        $stub = new Stub();
+        $date = new NoTimezoneDate($time);
+        $cast = DateCaster::castDateTime($date, Caster::castObject($date, \DateTime::class), $stub, false, 0);
+
+        $xDump = <<<EODUMP
+array:1 [
+  "\\x00~\\x00date" => $xDate
+]
+EODUMP;
+
+        $this->assertDumpEquals($xDump, $cast);
+
+        $xDump = <<<EODUMP
+Symfony\Component\VarDumper\Caster\ConstStub {
+  +type: 1
+  +class: "$xDate"
+  +value: "%A$xInfos%A"
+  +cut: 0
+  +handle: 0
+  +refCount: 0
+  +position: 0
+  +attr: []
+}
+EODUMP;
+
+        $this->assertDumpMatchesFormat($xDump, $cast["\0~\0date"]);
+    }
+
+    public static function provideNoTimezoneDateTimes()
+    {
+        return [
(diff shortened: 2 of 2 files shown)


src/Symfony/Component/VarDumper/Caster/DateCaster.php
```diff
@@ -27,8 +27,8 @@ class DateCaster
     public static function castDateTime(\DateTimeInterface $d, array $a, Stub $stub, bool $isNested, int $filter)
     {
         $prefix = Caster::PREFIX_VIRTUAL;
-        $location = $d->getTimezone()->getLocation();
-        $fromNow = (new \DateTime())->diff($d);
+        $location = $d->getTimezone() ? $d->getTimezone()->getLocation() : null;
+        $fromNow = $d->getTimezone() ? (new \DateTime())->diff($d) : new \DateInterval('PT0S');
```
derrabus: This is wrong. Having no timezone does not imply an offset of zero.

derrabus (changes requested):
Please add a test that covers your change.

derrabus:
Thank you for your PR. I disagree with the motivation of your change for two reasons:

* I don't want to implement workarounds for broken mocks. Your mock probably does not behave like the `DateTime` objects your code is usually dealing with. That is a clear sign that you should fix your mock.
* Mocking `DateTime` is in most cases a very pointless thing to do. One exception might be testing this very edge case. 🙈 

However [as per PHP's documentation](https://www.php.net/manual/en/datetime.gettimezone), `false` is a possible return value of `getTimezone()`, so VarDumper should be able to handle it.

I'm fine with the change, once my review has been addressed.

src/Symfony/Component/VarDumper/Caster/DateCaster.php
bogdanmoza: You are right, I have changed it.

bogdanmoza:
Thanks for the quick reply, you are right about the motivation, what I was thinking is that since PHP enables you to return false, then of course that allows for some special cases like dumping a Mock, and it is possible without a Mock also.

src/Symfony/Component/VarDumper/Tests/Caster/DateCasterTest.php
```diff
@@ -407,3 +453,15 @@ private function createInterval($intervalSpec, $ms, $invert)
         return $interval;
     }
 }
+
+class NoTimezoneDate extends \DateTime
+{
+    /**
+     * @return \DateTimeZone|false
+     */
+    #[\ReturnTypeWillChange]
```
stof, line 462: use an actual return type instead.

bogdanmoza: Hi thanks for the review, unfortunately I can't because union types were only introduced in php 8, so the unit tests will fail for the 7.* versions

stof: ah, I missed the fact that this PR targets 5.4

src/Symfony/Component/VarDumper/Caster/DateCaster.php
```diff
@@ -27,11 +27,11 @@ class DateCaster
     public static function castDateTime(\DateTimeInterface $d, array $a, Stub $stub, bool $isNested, int $filter)
     {
         $prefix = Caster::PREFIX_VIRTUAL;
-        $location = $d->getTimezone()->getLocation();
-        $fromNow = (new \DateTime())->diff($d);
+        $location = $d->getTimezone() ? $d->getTimezone()->getLocation() : null;
+        $fromNow = $d->getTimezone() ? (new \DateTime())->diff($d) : null;
```
nicolas-grekas: I reverted this change

nicolas-grekas:
Thank you @bogdanmoza.

PR #50654, opened 2023-06-13
[Validator] Add the `message` option to the `PasswordStrength` constraint

### LINKED ISSUE #50651: Message parameter not exists in PasswordStrength constraint attribute

6.3.0

### Description

I just upgraded to Symfony 6.3 to use the new `PasswordStrength` constraint. In the docs I can see how to change the message that is issued when the password does not reach the minimum required score. But the parameter message is not existent for the attribute.

Link to documentation: https://symfony.com/doc/current/reference/constraints/PasswordStrength.html#message

### How to reproduce

Using the attribute as documented, e.g.

```php

use Symfony\Component\Validator\Constraints\PasswordStrength;

...

    #[PasswordStrength(
        minScore: PasswordStrength::STRENGTH_STRONG,
        message: 'A message for the user if the password is not strong enough.'
    )]
    private string $plainPassword;
```

### Possible Solution

Adding the parameter to the attribute.

### Additional Context

Just another point: the documentation about the requirements of each score (1-4) is a bit unclear in the current version of the documentation.

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50651
| License       | MIT
| Doc PR        | No need

### DIFF
--- a/src/Symfony/Component/Validator/Constraints/PasswordStrength.php
+++ b/src/Symfony/Component/Validator/Constraints/PasswordStrength.php
@@ -40,13 +40,14 @@ final class PasswordStrength extends Constraint
 
     public int $minScore;
 
-    public function __construct(array $options = null, int $minScore = null, array $groups = null, mixed $payload = null)
+    public function __construct(array $options = null, int $minScore = null, array $groups = null, mixed $payload = null, string $message = null)
     {
         $options['minScore'] ??= self::STRENGTH_MEDIUM;
 
         parent::__construct($options, $groups, $payload);
 
         $this->minScore = $minScore ?? $this->minScore;
+        $this->message = $message ?? $this->message;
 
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
Let's target 6.3, that's clearly something we forgot about.

alexandre-daubois:
That's rebased on 6.3

Should the parameter be placed in second or third position while at it, to keep consistency between constraints?

nicolas-grekas (approved):
> Should the parameter be placed in second or third position while at it, to keep consistency between constraints?

That'd be a BC break, + it's not needed with named args.

nicolas-grekas:
Thank you @alexandre-daubois.

PR #50660, opened 2023-06-14
Make some classes final

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Symfony/Component/Mailer/Transport/Smtp/SmtpTransport.php
+++ b/src/Symfony/Component/Mailer/Transport/Smtp/SmtpTransport.php
@@ -244,12 +244,7 @@ class SmtpTransport extends AbstractTransport
         }
     }
 
-    /**
-     * @internal since version 6.1, to be made private in 7.0
-     *
-     * @final since version 6.1, to be made private in 7.0
-     */
-    protected function doHeloCommand(): void
+    private function doHeloCommand(): void
     {
         $this->executeCommand(sprintf("HELO %s\r\n", $this->domain), [250]);
     }
--- a/src/Symfony/Component/Serializer/Normalizer/ObjectNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/ObjectNormalizer.php
@@ -25,10 +25,8 @@ use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
  * Converts between objects and arrays using the PropertyAccess component.
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
- *
- * @final since Symfony 6.3
  */
-class ObjectNormalizer extends AbstractObjectNormalizer
+final class ObjectNormalizer extends AbstractObjectNormalizer
 {
     protected $propertyAccessor;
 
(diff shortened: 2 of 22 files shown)


src/Symfony/Component/Mailer/Transport/Smtp/SmtpTransport.php
```diff
@@ -29,7 +29,7 @@
  * @author Fabien Potencier <fabien@symfony.com>
  * @author Chris Corbyn
  */
-class SmtpTransport extends AbstractTransport
+final class SmtpTransport extends AbstractTransport
```
derrabus: This class has to remain `@final` because we actually extend it. See also the error Psalm produced because of that.

nicolas-grekas: not `@final` then if we extend it

derrabus: > not `@final` then if we extend it

Right. I totally missed that the class itself wasn't actually flagged as `@final`, just one protected method.

lyrixx: Oups, fixed

src/Symfony/Component/Serializer/Normalizer/ObjectNormalizer.php
```diff
@@ -25,10 +25,8 @@
  * Converts between objects and arrays using the PropertyAccess component.
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
- *
- * @final since Symfony 6.3
  */
-class ObjectNormalizer extends AbstractObjectNormalizer
+final class ObjectNormalizer extends AbstractObjectNormalizer
```
lyrixx, line 29: This breaks some tests:

```
PHP Fatal error:  Class Symfony\Component\Serializer\Normalizer\ObjectNormalizer@anonymous cannot extend final class Symfony\Component\Serializer\Normalizer\ObjectNormalizer in /home/runner/work/symfony/symfony/src/Symfony/Component/Serializer/Tests/Normalizer/ObjectNormalizerTest.php on line 685```

Comment (stof) [reply src/Symfony/Component/Serializer/Normalizer/ObjectNormalizer.php:29]: to me, this looks like a test hacking a way to look at the default context. This test should probably be rewritten (on lower branch) to avoid relying on the fact that `isCircularReference` will be called, allowing their child class to override this method to perform some assertion on the protected state of the class.

File: src/Symfony/Component/ErrorHandler/Tests/DebugClassLoaderTest.php [hunk 3]
```diff
@@ -15,7 +15,7 @@
 use Symfony\Bridge\ErrorHandler\Tests\Fixtures\ExtendsDeprecatedParent;
 use Symfony\Component\ErrorHandler\DebugClassLoader;
 
-class DebugClassLoaderTest extends TestCase
+final class DebugClassLoaderTest extends TestCase
```
stof: why changing test classes ?

src/Symfony/Component/Serializer/Normalizer/ObjectNormalizer.php
lyrixx: see https://github.com/symfony/symfony/pull/50735

nicolas-grekas:
(rebase needed)

lyrixx:
I need to finish https://github.com/symfony/symfony/pull/50736 first

nicolas-grekas:
Thank you @lyrixx.

lyrixx:
This is weird, the PR has not been closed?!

nicolas-grekas:
:shrug:

PR #50662, opened 2023-06-14
[FrameworkBundle] Add `HttpClientAssertionsTrait` which provide shortcuts to assert HTTP calls was triggered

welcoMattic:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/18970

This PR brings a new collection (to enhance) of assertions to help users to ensure some HTTP calls has been triggered during a Request processing.

## Example

An example is better than thousands words:

HttpClientAssertionsTest.php
```php
public function testHttpClientAssertions()
{
    $client = static::createClient();
    $client->enableProfiler();

    $client->request('GET', '/foo');

    $this->assertHttpClientRequest($client, 'symfony.http_client', 'https://symfony.com/');
    $this->assertHttpClientRequest($client, 'symfony.http_client', 'https://symfony.com/', 'POST', 'foo');
    $this->assertHttpClientRequest($client, 'symfony.http_client', 'https://symfony.com/', 'POST', ['foo' => 'bar']);
    $this->assertHttpClientRequest($client, 'symfony.http_client', 'https://symfony.com/', 'POST', ['foo' => 'bar']);
    $this->assertNotHttpClientRequest($client, 'symfony.http_client', 'https://laravel.com');


    $this->assertHttpClientRequestCount($client, 'symfony.http_client', 5);
}
```

FooController.php
```php
public function index(HttpClientInterface $symfonyHttpClient)
{
    $symfonyHttpClient->request('GET', '/');
    $symfonyHttpClient->request('POST', '/', ['body' => 'foo']);
    $symfonyHttpClient->request('POST', '/', ['body' => ['foo' => 'bar']]);
    $symfonyHttpClient->request('POST', '/', ['json' => ['foo' => 'bar']]);
    $symfonyHttpClient->request('GET', '/doc/current/index.html');

    return new Response();
}
```

http_client.yaml
```yaml
framework:
    http_client:
        scoped_clients:
            symfony.http_client:
                base_uri: 'https://symfony.com'
```

This is a first try to improve DX in test, to allow users to assert more complex things (here it's HTTP calls, but I think also about adding assertions on logs for example).

## Todo

- [x] Write doc

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
@@ -1,0 +1,134 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\FrameworkBundle\Test;
+
+use Symfony\Bundle\FrameworkBundle\KernelBrowser;
+use Symfony\Component\HttpClient\DataCollector\HttpClientDataCollector;
+
+/*
+ * @author Mathieu Santostefano <msantostefano@protonmail.com>
+ */
+
+trait HttpClientAssertionsTrait
+{
+    public static function assertHttpClientRequest(string $expectedUrl, string $expectedMethod = 'GET', string|array $expectedBody = null, array $expectedHeaders = [], string $httpClientId = 'http_client'): void
+    {
+        /** @var KernelBrowser $client */
+        $client = static::getClient();
+
+        if (!($profile = $client->getProfile())) {
+            static::fail('The Profiler must be enabled for the current request. Please ensure to call "$client->enableProfiler()" before making the request.');
+        }
+
+        /** @var HttpClientDataCollector $httpClientDataCollector */
+        $httpClientDataCollector = $profile->getCollector('http_client');
+        $expectedRequestHasBeenFound = false;
+
+        if (!\array_key_exists($httpClientId, $httpClientDataCollector->getClients())) {
+            static::fail(sprintf('HttpClient "%s" is not registered.', $httpClientId));
+        }
+
+        foreach ($httpClientDataCollector->getClients()[$httpClientId]['traces'] as $trace) {
+            if (($expectedUrl !== $trace['info']['url'] && $expectedUrl !== $trace['url'])
+                || $expectedMethod !== $trace['method']
+            ) {
+                continue;
+            }
+
+            if (null !== $expectedBody) {
+                $actualBody = null;
+
+                if (null !== $trace['options']['body'] && null === $trace['options']['json']) {
+                    $actualBody = \is_string($trace['options']['body']) ? $trace['options']['body'] : $trace['options']['body']->getValue(true);
+                }
+
+                if (null === $trace['options']['body'] && null !== $trace['options']['json']) {
+                    $actualBody = $trace['options']['json']->getValue(true);
+                }
+
+                if (!$actualBody) {
+                    continue;
+                }
+
+                if ($expectedBody === $actualBody) {
+                    $expectedRequestHasBeenFound = true;
+
+                    if (!$expectedHeaders) {
+                        break;
+                    }
+                }
+            }
+
+            if ($expectedHeaders) {
+                $actualHeaders = $trace['options']['headers'] ?? [];
+
+                foreach ($actualHeaders as $headerKey => $actualHeader) {
+                    if (\array_key_exists($headerKey, $expectedHeaders)
+                        && $expectedHeaders[$headerKey] === $actualHeader->getValue(true)
+                    ) {
+                        $expectedRequestHasBeenFound = true;
+                        break 2;
+                    }
+                }
+            }
+
+            $expectedRequestHasBeenFound = true;
+            break;
+        }
+
+        self::assertTrue($expectedRequestHasBeenFound, 'The expected request has not been called: "'.$expectedMethod.'" - "'.$expectedUrl.'"');
+    }
+
+    public function assertNotHttpClientRequest(string $unexpectedUrl, string $expectedMethod = 'GET', string $httpClientId = 'http_client'): void
+    {
+        /** @var KernelBrowser $client */
+        $client = static::getClient();
+
+        if (!$profile = $client->getProfile()) {
+            static::fail('The Profiler must be enabled for the current request. Please ensure to call "$client->enableProfiler()" before making the request.');
+        }
+
+        /** @var HttpClientDataCollector $httpClientDataCollector */
+        $httpClientDataCollector = $profile->getCollector('http_client');
+        $unexpectedUrlHasBeenFound = false;
+
+        if (!\array_key_exists($httpClientId, $httpClientDataCollector->getClients())) {
+            static::fail(sprintf('HttpClient "%s" is not registered.', $httpClientId));
+        }
+
+        foreach ($httpClientDataCollector->getClients()[$httpClientId]['traces'] as $trace) {
+            if (($unexpectedUrl === $trace['info']['url'] || $unexpectedUrl === $trace['url'])
+                && $expectedMethod === $trace['method']
+            ) {
+                $unexpectedUrlHasBeenFound = true;
+                break;
+            }
+        }
+
+        self::assertFalse($unexpectedUrlHasBeenFound, sprintf('Unexpected URL called: "%s" - "%s"', $expectedMethod, $unexpectedUrl));
+    }
+
+    public static function assertHttpClientRequestCount(int $count, string $httpClientId = 'http_client'): void
+    {
+        /** @var KernelBrowser $client */
+        $client = static::getClient();
+
+        if (!($profile = $client->getProfile())) {
+            static::fail('The Profiler must be enabled for the current request. Please ensure to call "$client->enableProfiler()" before making the request.');
+        }
+
+        /** @var HttpClientDataCollector $httpClientDataCollector */
+        $httpClientDataCollector = $profile->getCollector('http_client');
+
+        self::assertCount($count, $httpClientDataCollector->getClients()[$httpClientId]['traces']);
+    }
+}
--- a/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
+++ b/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 6.4
 ---
 
+ * Add `HttpClientAssertionsTrait`
  * Add `AbstractController::renderBlock()` and `renderBlockView()`
  * Add native return type to `Translator` and to `Application::reset()`
  * Deprecate the integration of Doctrine annotations, either uninstall the `doctrine/annotations` package or disable the integration by setting `framework.annotations` to `false`
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/Functional/Bundle/TestBundle/Controller/HttpClientController.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/Functional/Bundle/TestBundle/Controller/HttpClientController.php
@@ -1,0 +1,35 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\FrameworkBundle\Tests\Functional\Bundle\TestBundle\Controller;
+
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+class HttpClientController
+{
+    public function index(HttpClientInterface $httpClient, HttpClientInterface $symfonyHttpClient): Response
+    {
+        $httpClient->request('GET', 'https://symfony.com/');
+
+        $symfonyHttpClient->request('GET', '/');
+        $symfonyHttpClient->request('POST', '/', ['body' => 'foo']);
+        $symfonyHttpClient->request('POST', '/', ['body' => ['foo' => 'bar']]);
+        $symfonyHttpClient->request('POST', '/', ['json' => ['foo' => 'bar']]);
+        $symfonyHttpClient->request('POST', '/', [
+            'headers' => ['X-Test-Header' => 'foo'],
+            'json' => ['foo' => 'bar'],
+        ]);
+        $symfonyHttpClient->request('GET', '/doc/current/index.html');
+
+        return new Response();
+    }
+}
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/Functional/Bundle/TestBundle/Resources/config/routing.yml
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/Functional/Bundle/TestBundle/Resources/config/routing.yml
@@ -61,6 +61,10 @@ send_email:
     path:     /send_email
     defaults: { _controller: Symfony\Bundle\FrameworkBundle\Tests\Functional\Bundle\TestBundle\Controller\EmailController::indexAction }
 
+http_client_call:
+    path:     /http_client_call
+    defaults: { _controller: Symfony\Bundle\FrameworkBundle\Tests\Functional\Bundle\TestBundle\Controller\HttpClientController::index }
+
 uid:
     resource: "../../Controller/UidController.php"
     type: "annotation"
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/Functional/Bundle/TestBundle/Tests/MockClientCallback.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/Functional/Bundle/TestBundle/Tests/MockClientCallback.php
@@ -1,0 +1,23 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\FrameworkBundle\Tests\Functional\Bundle\TestBundle\Tests;
+
+use Symfony\Component\HttpClient\Response\MockResponse;
+use Symfony\Contracts\HttpClient\ResponseInterface;
+
+class MockClientCallback
+{
+    public function __invoke(string $method, string $url, array $options = []): ResponseInterface
+    {
+        return new MockResponse('foo');
+    }
+}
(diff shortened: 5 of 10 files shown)


src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        if (($profile = $client->getProfile()) === false) {
+            static::fail('The Profiler must be enabled for the current request. Please ensure you have called $client->enableProfiler() before making the request.');
+        }
+
+        /** @var HttpClientDataCollector $httpClientDataCollector */
+        $httpClientDataCollector = $profile->getCollector('http_client');
+
+        foreach ($urls as $i => $url) {
+            self::assertSame($url, $httpClientDataCollector->getClients()[$httpClientId]['traces'][$i]['info']['url']);
```
stof: should we enforce that urls in the tests and registered traces are in the same order ?

stof: This will trigger a notice if the trace `$i` does not exist, which is a bad way to fail the test if the expected call was not done.

welcoMattic: I think it's useful to ensure URLs are called in a certain order. Do you think we should do it in an another way?

welcoMattic: Yes, sure. I will add a check on the sizes of both arrays, and use array_values to ensure keys are reset.

stof: Well, I'm wondering whether we should _enforce_ that part of the check (after all, the client itself might run requests concurrently depending on how you triggered them)

And this currently forbids having additional requests being triggered before or in-between expected ones, but not after them.

stof: Also, this expectation taking a list of URLs only does not give any chance to perform an assertion on which method was used for them for instance. So the API seems flawed to me.

welcoMattic: I have 2 ideas to add a check on the method:

```php
$this->assertHttpClientHasBeenCalled($client, 'symfony.http_client', [
    'https://symfony.com/' => [
        'method' => 'GET',
    ],
    'https://symfony.com/doc/current/index.html' => [
        'method' => 'GET',
    ],
]);
```

or

```php
$this->assertHttpClientHasBeenCalled($client, 'symfony.http_client', [
    'GET' => [
        'https://symfony.com/',
        'https://symfony.com/doc/current/index.html'
    ],
]);
```

I prefer the first, because we could add `body` to add a check on the body for POST/PATCH/PUT methods.

welcoMattic: I've abandoned the idea to respect the order. It could lead to lots of false positive, as you said "after all, the client itself might run requests concurrently depending on how you triggered them".

stof: I would not index by the URL either, because multiple requests might be done for the same URL with different methods (or different body in the case of requests with a body).

I would suggest passing a list of request info, with both the method and the url inside the request info instead.

Or maybe the assertion should care about a single request instead (and if you want to assert that multiple ones have been done, you would run separate assertions). This way, the various checks could be handled as optional arguments of the method instead.

welcoMattic:
I've update the examples in PR description. Now we allow assertion on a single URL (with method and optional body). I've added the opposition assertion `assertHttpClientHasNotBeenCalledForUrl` and `assertHttpClientCountRequests`.

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,85 @@
...
+namespace Symfony\Bundle\FrameworkBundle\Test;
+
+/*
+ * @author Mathieu Santostefano <msantostefano@protonmail.com>
+ */
+
+use Symfony\Bundle\FrameworkBundle\KernelBrowser;
+use Symfony\Component\HttpClient\DataCollector\HttpClientDataCollector;
+
+trait HttpClientAssertionsTrait
```
OskarStark:
```suggestion

use Symfony\Bundle\FrameworkBundle\KernelBrowser;
use Symfony\Component\HttpClient\DataCollector\HttpClientDataCollector;

/*
 * @author Mathieu Santostefano <msantostefano@protonmail.com>
 */
trait HttpClientAssertionsTrait
```

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,87 @@
...
+use Symfony\Bundle\FrameworkBundle\KernelBrowser;
+use Symfony\Component\HttpClient\DataCollector\HttpClientDataCollector;
+
+/*
+ * @author Mathieu Santostefano <msantostefano@protonmail.com>
+ */
+
+trait HttpClientAssertionsTrait
+{
+    public static function assertHttpClientHasBeenCalledForUrl(KernelBrowser $client, string $httpClientId, string $expectedUrl, string $expectedMethod = 'GET', string|array $expectedBody = null): void
```
ro0NL, line 23: assertHttpClientRequestsUrl v assertHttpClientRequestsNotUrl?

what about headers?

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,87 @@
...
+    {
+        if (!($profile = $client->getProfile())) {
```
ro0NL, line 25: im curious if we should collect requests in test env by other means

nicolas-grekas, line 28:
```suggestion
        if (!$profile = $client->getProfile()) {
```

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,87 @@
...
+            static::fail('The Profiler must be enabled for the current request. Please ensure you have called $client->enableProfiler() before making the request.');
+        }
+
+        /** @var HttpClientDataCollector $httpClientDataCollector */
+        $httpClientDataCollector = $profile->getCollector('http_client');
+        $expectedUrlHasBeenFound = false;
+
+        foreach ($httpClientDataCollector->getClients()[$httpClientId]['traces'] as $trace) {
```
nicolas-grekas, line 40: we should check that `$httpClientDataCollector->getClients()[$httpClientId]` exists first

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,87 @@
...
+            if (($expectedUrl === $trace['info']['url'] || $expectedUrl === $trace['url'])
+                && $expectedMethod === $trace['method']) {
```
nicolas-grekas:
```suggestion
            if (($expectedUrl !== $trace['info']['url'] && $expectedUrl !== $trace['url'])
                || $expectedMethod !== $trace['method']
            ) {
               continue;
```

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,87 @@
...
+                break;
+            }
+        }
+
+        self::assertFalse($unexpectedUrlHasBeenFound, 'The unexpected URL has been called: "'.$expectedMethod.'" - "'.$unexpectedUrl.'"');
+    }
+
+    public static function assertHttpClientCountRequests(KernelBrowser $client, string $httpClientId, int $count): void
+    {
+        if (!($profile = $client->getProfile())) {
```
nicolas-grekas, line 125:
```suggestion
        if (!$profile = $client->getProfile()) {
```

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,87 @@
...
+            static::fail('The Profiler must be enabled for the current request. Please ensure you have called $client->enableProfiler() before making the request.');
+        }
+
+        /** @var HttpClientDataCollector $httpClientDataCollector */
+        $httpClientDataCollector = $profile->getCollector('http_client');
+        $unexpectedUrlHasBeenFound = false;
+
+        foreach ($httpClientDataCollector->getClients()[$httpClientId]['traces'] as $trace) {
+            if (($unexpectedUrl === $trace['info']['url'] || $unexpectedUrl === $trace['url'])
+                && $expectedMethod === $trace['method']) {
```
nicolas-grekas:
```suggestion
                && $expectedMethod === $trace['method']
            ) {
```

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,87 @@
...
+    {
+        if (!($profile = $client->getProfile())) {
+            static::fail('The Profiler must be enabled for the current request. Please ensure you have called $client->enableProfiler() before making the request.');
+        }
+
+        /** @var HttpClientDataCollector $httpClientDataCollector */
+        $httpClientDataCollector = $profile->getCollector('http_client');
+        $unexpectedUrlHasBeenFound = false;
+
+        foreach ($httpClientDataCollector->getClients()[$httpClientId]['traces'] as $trace) {
```
nicolas-grekas, line 108: missing isset check

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
nicolas-grekas: `assertHttpClientRequest` / `assertNotHttpClientRequest`?

(no strong opinion either)

nicolas-grekas (commented):
Random thoughts:

No need for checking headers using these?
Can't we remove the $client argument somehow? (don't we have it in the context somewhere?)
WHat about defaulting $httpClientId default to `http_client`?

welcoMattic:
@nicolas-grekas 

> Random thoughts:
> 
> No need for checking headers using these? 

Sure, I can add header checks :+1: 

> Can't we remove the $client argument somehow? (don't we have it in the context somewhere?) 

~As `BrowserKitAssertionTrait::getClient` is private, we don't have any other way to get the client here. Should we dig deep to find a way to do?~

We have access to this trait through `WebTestAssertionsTrait` in `WebTestCase`, so we can call `static::getClient()`

> WHat about defaulting $httpClientId default to `http_client`?

Nice idea

weaverryan:
Same thoughts as Nicolas: if we can remove the need for the client arg and service name (when using the default), I think I’ll really dig it!

welcoMattic:
@weaverryan it's done :) I've removed it https://github.com/symfony/symfony/pull/50662/files#diff-be0ff1fa989a87452ee0ecc8a932f09d50d59d3ecb4c6bce9565202aa254bd37R23. 

Through traits we have access to `getClient` from `BrowserKitAssertionsTrait`, et tada!

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
fabpot: Not resolved AFAICS

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,134 @@
...
+
+trait HttpClientAssertionsTrait
+{
+    public static function assertHttpClientRequest(string $expectedUrl, string $expectedMethod = 'GET', string|array $expectedBody = null, array $expectedHeaders = [], string $httpClientId = 'http_client'): void
+    {
+        /** @var KernelBrowser $client */
+        $client = static::getClient();
+
+        if (!($profile = $client->getProfile())) {
+            static::fail('The Profiler must be enabled for the current request. Please ensure you have called $client->enableProfiler() before making the request.');
```
fabpot:
```suggestion
            static::fail('The Profiler must be enabled for the current request. Please ensure to call "$client->enableProfiler()" before making the request.');
```

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        /** @var HttpClientDataCollector $httpClientDataCollector */
+        $httpClientDataCollector = $profile->getCollector('http_client');
+        $expectedRequestHasBeenFound = false;
+
+        if (!\array_key_exists($httpClientId, $httpClientDataCollector->getClients())) {
+            static::fail(sprintf('The HttpClient "%s" is not registered.', $httpClientId));
```
fabpot:
```suggestion
            static::fail(sprintf('HttpClient "%s" is not registered.', $httpClientId));
```

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,134 @@
...
+                if ($expectedBody === $actualBody) {
+                    $expectedRequestHasBeenFound = true;
+
+                    if ([] === $expectedHeaders) {
+                        break;
+                    }
+                }
+            }
+
+            if ([] !== $expectedHeaders) {
```
fabpot:
```suggestion
            if ($expectedHeaders) {
```

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,134 @@
...
+
+        self::assertTrue($expectedRequestHasBeenFound, 'The expected request has not been called: "'.$expectedMethod.'" - "'.$expectedUrl.'"');
+    }
+
+    public function assertNotHttpClientRequest(string $unexpectedUrl, string $expectedMethod = 'GET', string $httpClientId = 'http_client'): void
+    {
+        /** @var KernelBrowser $client */
+        $client = static::getClient();
+
+        if (!($profile = $client->getProfile())) {
```
fabpot:
```suggestion
        if (!$profile = $client->getProfile()) {
```

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,134 @@
...
+            static::fail('The Profiler must be enabled for the current request. Please ensure you have called $client->enableProfiler() before making the request.');
```
fabpot: Same as above

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        /** @var HttpClientDataCollector $httpClientDataCollector */
+        $httpClientDataCollector = $profile->getCollector('http_client');
+        $unexpectedUrlHasBeenFound = false;
+
+        if (!\array_key_exists($httpClientId, $httpClientDataCollector->getClients())) {
+            static::fail(sprintf('The HttpClient "%s" is not registered.', $httpClientId));
```
fabpot:
```suggestion
            static::fail(sprintf('HttpClient "%s" is not registered.', $httpClientId));
```

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,134 @@
...
+        foreach ($httpClientDataCollector->getClients()[$httpClientId]['traces'] as $trace) {
+            if (($unexpectedUrl === $trace['info']['url'] || $unexpectedUrl === $trace['url'])
+                && $expectedMethod === $trace['method']
+            ) {
+                $unexpectedUrlHasBeenFound = true;
+                break;
+            }
+        }
+
+        self::assertFalse($unexpectedUrlHasBeenFound, 'The unexpected URL has been called: "'.$expectedMethod.'" - "'.$unexpectedUrl.'"');
```
fabpot:
```suggestion
        self::assertFalse($unexpectedUrlHasBeenFound, sprintf('Unexpected URL called: "%s" - "%s"", $expectedMethod, $unexpectedUrl');
```

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
fabpot: Please take that into account

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    public static function assertHttpClientRequestCount(int $count, string $httpClientId): void
+    {
+        /** @var KernelBrowser $client */
+        $client = static::getClient();
+
+        if (!($profile = $client->getProfile())) {
+            static::fail('The Profiler must be enabled for the current request. Please ensure you have called $client->enableProfiler() before making the request.');
```
fabpot: Same as above

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,134 @@
...
+                        break 2;
+                    }
+                }
+            }
+
+            $expectedRequestHasBeenFound = true;
+            break;
+        }
+
+        self::assertTrue($expectedRequestHasBeenFound, 'The expected request has not been called: "'.$expectedMethod.'" - "'.$expectedUrl.'"');
```
fabpot, line 88:
```suggestion
        self::assertTrue($expectedRequestHasBeenFound, 'Expected request not called: "'.$expectedMethod.'" - "'.$expectedUrl.'"');
```

fabpot (changes requested):

src/Symfony/Bundle/FrameworkBundle/Test/HttpClientAssertionsTrait.php
```diff
@@ -0,0 +1,134 @@
...
+                }
+
+                if (!$actualBody) {
+                    continue;
+                }
+
+                if ($expectedBody === $actualBody) {
+                    $expectedRequestHasBeenFound = true;
+
+                    if ([] === $expectedHeaders) {
```
fabpot:
```suggestion
                    if (!$expectedHeaders) {
```

fabpot (changes requested):

fabpot (approved):

fabpot:
Thank you @welcoMattic.

PR #50663, opened 2023-06-14
[Console] Add `SignalMap` to map signal value to its name

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

I'm migrating some code to use new SF features, but I have a very little regression. 

Before :

```php
    private function setUpSignals(): void
    {
        pcntl_signal(\SIGTERM, function () {
            $this->stop('SIGTERM');
        });
        pcntl_signal(\SIGINT, function () {
            $this->stop('SIGINT');
        });

        pcntl_async_signals(true);
    }
```

Now:

```php
    public function handleSignal(int $signal): void
    {
        // I need the following line
        $signal = SignalMap::getSignalName($signal);
        $this->stop("Signal {$signal} received.");
    }
```

As you can see, now in my log I have an int (`2`, `15`), and not anymore `SIGTERM`, `SIGINT`.

### DIFF
--- a/src/Symfony/Component/Console/SignalRegistry/SignalMap.php
+++ b/src/Symfony/Component/Console/SignalRegistry/SignalMap.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Console\SignalRegistry;
+
+/**
+ * @author Grégoire Pineau <lyrixx@lyrixx.info>
+ */
+class SignalMap
+{
+    private static array $map;
+
+    public static function getSignalName(int $signal): ?string
+    {
+        if (!\extension_loaded('pcntl')) {
+            return null;
+        }
+
+        if (!isset(self::$map)) {
+            $r = new \ReflectionExtension('pcntl');
+            $c = $r->getConstants();
+            $map = array_filter($c, fn ($k) => str_starts_with($k, 'SIG') && !str_starts_with($k, 'SIG_'), \ARRAY_FILTER_USE_KEY);
+            self::$map = array_flip($map);
+        }
+
+        return self::$map[$signal] ?? null;
+    }
+}
--- a/src/Symfony/Component/Console/CHANGELOG.md
+++ b/src/Symfony/Component/Console/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+ * Add `SignalMap` to map signal value to its name
+
 6.3
 ---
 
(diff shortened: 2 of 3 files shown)


src/Symfony/Component/Console/SignalRegistry/SignalMap.php
```diff
@@ -0,0 +1,32 @@
...
+ * @author Grégoire Pineau <lyrixx@lyrixx.info>
+ */
+class SignalMap
+{
+    private static array $map;
+
+    public static function getSignalName(int $signal): ?string
+    {
+        if (!isset(self::$map)) {
+            $r = new \ReflectionExtension('pcntl');
```
stof, line 28: what if `pcntl` is not loaded (for instance when running on windows) ? Should we initialize the map to an empty array in that case ?

lyrixx: I didn't know if I should throw an exception or return null.
WDYT?

stof: given that it already returns null for unknown signals, I would return null.

stof: Btw, the test about the unknown signal could be extracted into a separate test (instead of being in the same dataprovider), and the `@requires extension pcntl` moved to the test covering known signals only.

lyrixx: Good idea 👍🏼 
I updated the PR

stof (approved):

src/Symfony/Component/Console/SignalRegistry/SignalMap.php
```diff
@@ -0,0 +1,36 @@
...
+    public static function getSignalName(int $signal): ?string
+    {
+        if (!\extension_loaded('pcntl')) {
+            return null;
+        }
+
+        if (!isset(self::$map)) {
+            $r = new \ReflectionExtension('pcntl');
+            $c = $r->getConstants();
+            $map = array_filter($c, fn ($k) => str_starts_with($k, 'SIG') && !str_starts_with($k, 'SIG_'), \ARRAY_FILTER_USE_KEY);
```
nicolas-grekas, line 30: what about harcoding the list instead? the list of signals and their name+value is known and standard, isn't it?

lyrixx: not really :( Look at the comments: https://www.php.net/manual/en/pcntl.constants.php

stof: But then, don't we have an issue in tests where values are hardcoded instead of using the constants ?

lyrixx: I used very stable ones

chalasr (approved):

fabpot (approved):

fabpot:
Thank you @lyrixx.

PR #50664, opened 2023-06-14
[HttpFoundation] Default accessing structured request body to empty array

### LINKED ISSUE #50647: [HttpFoundation] getPayload throws JsonException if request body is empty (again)

6.3.0

### Description

Calling `$request->request->get("param")` returns `null` if the body is empty

Replacing it with `$request->getPayload()->get("param")` throws a `JsonException` under the same condition.

This is a big BC just for a name change.

The issue was reported in https://github.com/symfony/symfony/issues/50490 but it got no answer except documenting that `getPayload()` throws a `JsonException`.

### How to reproduce

To reproduce, simply call `$request->getPayload()->get("param")` or `$request->getPayload()->all()` or even `$request->getPayload()->has("param")` to check if a parameter was passed in the body when none was in fact passed.

### Possible Solution

Unless we are supposed to keep using `$request->request` and only infrequently use `$request->getPayload()` for cases where we need to know the client failed to provide a body, I suggest to return an empty array.

```php
    public function toArray(): array
    {
        if ('' === $content = $this->getContent()) {
            return [];
        }
```

An exception should be used for exceptional events. An empty body is nothing but exceptional.

### Additional Context

_No response_

ro0NL:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes-ish
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50647
| License       | MIT

Similar like in PHP `$_POST` is always type array, and in Symfony `$request->request` is always array-able

The state can be checked by other means (headers, method, etc.), if nessecary

### DIFF
--- a/src/Symfony/Component/HttpFoundation/Request.php
+++ b/src/Symfony/Component/HttpFoundation/Request.php
@@ -1526,7 +1526,7 @@ class Request
     public function toArray(): array
     {
         if ('' === $content = $this->getContent()) {
-            throw new JsonException('Request body is empty.');
+            return [];
         }
 
         try {
--- a/src/Symfony/Component/HttpFoundation/Tests/RequestTest.php
+++ b/src/Symfony/Component/HttpFoundation/Tests/RequestTest.php
@@ -1279,9 +1279,7 @@ class RequestTest extends TestCase
     public function testToArrayEmpty()
     {
         $req = new Request();
-        $this->expectException(JsonException::class);
-        $this->expectExceptionMessage('Request body is empty.');
-        $req->toArray();
+        $this->assertSame([], $req->toArray());
     }
 
     public function testToArrayNonJson()


stof:
I suggest changing that only for `getPayload`.

`toArray` is only doing json-decoding, and an empty body is not valid json.

ro0NL:
i consider toArray / getPayload equivalent public API :)

ro0NL:
i mean, from the original issue, a throwing `$r->getPayload()->get('key')` would be equally annoying as a throwing `$r->toArray()['key']` :thinking: 

perhaps both POVs are equally valid, let me know if we prefer a getPayload specific solution :+1: 

> toArray is only doing json-decoding

it's kinda an implementation detail ;)

stof:
@ro0NL if you don't know that the body is valid json, `toArray` might still throw for something else than an empty body. So I don't see how this changes much things.

note that HttpKernel will automatically convert those exceptions to a 4xx response.

ro0NL:
it enables doing `if ($r->toArray()['key'])` over `if ($r->getContent && $r->toArray()['key'])`

if the request content is actually corrupt, let it be bad JSON, then we should throw (which hasnt changed here)

my point here is, toArray/getPayload should not be used for state checking IMHO, but a route/controller condition should

stof:
@ro0NL if your endpoint expects JSON, sending an empty body is actually a corrupt request.

ro0NL:
im still not sure toArray conveys your endpoint expects JSON, at least not by its name :roll_eyes: 

to me, and if i recall it correctly, it was designed to be an implementation detail

btw, see also https://github.com/symfony/symfony/pull/50492#discussion_r1212088960 :sweat_smile:

ro0NL:
TLDR: toArray to me is a poor-man getPayload solution, and should have been deprecated

stof:
@ro0NL we explicitly decided not to deprecate it immediately to make things easier for the ecosystem needing to support multiple version of Symfony.

I would expect us to deprecate it in 7.1.

ro0NL:
> Gets the request body decoded as array, typically from a JSON payload

to me proves JSON is an implementation detail

in that sense the method is already bugged for not returning $_POST

im gonna wait for toArray deprecation, so we can avoid this debate

ro0NL:
@stof 
> I would expect us to deprecate it in 7.1.

we should get rid of it

PR #50667, opened 2023-06-15
Improve return type of `getProvidedServices` in `ServiceProviderInterface`

andersonamuller:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | maybe
| New feature?  | no
| Deprecations? | no
| License       | MIT

This describes better the shape of the array it returns.

### DIFF
--- a/src/Symfony/Contracts/Service/ServiceProviderInterface.php
+++ b/src/Symfony/Contracts/Service/ServiceProviderInterface.php
@@ -39,7 +39,7 @@ interface ServiceProviderInterface extends ContainerInterface
      *  * ['foo' => '?'] means the container provides service name "foo" of unspecified type
      *  * ['bar' => '?Bar\Baz'] means the container provides a service "bar" of type Bar\Baz|null
      *
-     * @return string[] The provided service types, keyed by service names
+     * @return array<string, string> The provided service types, keyed by service names
      */
     public function getProvidedServices(): array;
 }


src/Symfony/Contracts/Service/ServiceProviderInterface.php
```diff
@@ -39,7 +39,7 @@ public function has(string $id): bool;
      *  * ['foo' => '?'] means the container provides service name "foo" of unspecified type
      *  * ['bar' => '?Bar\Baz'] means the container provides a service "bar" of type Bar\Baz|null
      *
-     * @return string[] The provided service types, keyed by service names
+     * @return array<string, string> The provided service types, keyed by service names
```
nicolas-grekas, line 42: the key can be omitted which signals that the name should be the same as the type
maybe this is worth some more words in the description also?
```suggestion
     * @return array<string|int, string> The provided service types, keyed by service names
```

stof: `array<string|int, string>` is really the same than `string[]` (as array keys cannot be something else than `string|int`, so this would be the same than the current code.

However, the case of using an integer key to use the type as name is not documented in the phpdoc. This should be added if it is indeed supported.

andersonamuller: Indeed, in the examples showed in the PHPDoc, integer keys are not documented, and I don't think they should, neither the service provider should support it. What would be the use case? 
If you need a list of services you can use the `TaggedIterator`.

nicolas-grekas: The use case is omitting keys for brevity:

`$services[] = Foo::class` is the interpreted as `$services[Foo::class] = Foo::class`

andersonamuller: But this is just for the set of the internal array, when returning it it actually fills the array keys with a string in the `ServiceLocatorTrait`. Also, if you look at the consumers of this method on the Symfony codebase, they all expect a array key of string. 

The `get` and `has` methods of the interface also pass a string `$id`

stof: @nicolas-grekas aren't you confusing ServiceProviderInterface with ServiceSubscriberInterface ?

stof:
To solve the issue reported by psalm for ServiceLocatorTrait, you should also improve the type of its constructor argument to say that its array of callables should be indexed by strings.

src/Symfony/Contracts/Service/ServiceProviderInterface.php
nicolas-grekas: Yes indeed I am! Sorry for the confusion!

nicolas-grekas (approved):

andersonamuller:
> To solve the issue reported by psalm for ServiceLocatorTrait, you should also improve the type of its constructor argument to say that its array of callables should be indexed by strings.

Yes, indeed, I first wanted to confirm that is a valid change. I will fix the rest when I have more time.

fabpot (approved):

fabpot:
@andersonamuller Can you rebase your PR to get rid of the merge commit? Thank you.

nicolas-grekas:
Thank you @andersonamuller.

PR #49299, opened 2023-02-08
[HttpClient] Fix over-encoding of URL parts to match browser's behavior

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #48315
| License       | MIT
| Doc PR        | -

Yes, the RFC says these chars should be url-encoded:
```
[ ] \ ^ ` { | }
```

But in practice, browser's don't encode them, and some servers don't expect them to be encoded either.
See https://stackoverflow.com/questions/2366260/whats-valid-and-whats-not-in-a-uri-query for some pointers.

PR #50671, opened 2023-06-15
[HttpClient] Fix encoding some characters in query strings

### LINKED ISSUE #50670: HttpClient parseUrl is broken and not RFC compilant

5.4.24

### Description

When I use HttpClient with url which contain this data a part of path segment:
```
custom|2010-01-01 00:00:00|2023-06-15 05:50:35
```

I have got error from calling part because of wrongly url encoded data.

Issue is caused by this pull https://github.com/symfony/symfony/pull/49299

### How to reproduce

Calling  `HttpClient::parseUrl('http://some.url/custom|2010-01-01 00:00:00|2023-06-15 05:50:35?a=b#c)'`  should return result array:
```
Array
(
    ...
    [path] => /custom%7C2010-01-01%2000:00:00%7C2023-06-15%2005:50:35
    ...
)
```

### Possible Solution

_No response_

### Additional Context

_No response_

Kozzi11:
| Q             | A
| ------------- | ---
| Branch?       | 5.4 
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #50670 
| License       | MIT

### DIFF
--- a/src/Symfony/Component/HttpClient/HttpClientTrait.php
+++ b/src/Symfony/Component/HttpClient/HttpClientTrait.php
@@ -547,7 +547,7 @@ trait HttpClientTrait
             }
 
             // https://tools.ietf.org/html/rfc3986#section-3.3
-            $parts[$part] = preg_replace_callback("#[^-A-Za-z0-9._~!$&/'()[\]*+,;=:@\\\\^`{|}%]++#", function ($m) { return rawurlencode($m[0]); }, $parts[$part]);
+            $parts[$part] = preg_replace_callback("#[^-A-Za-z0-9._~!$&/'()[\]*+,;=:@{}%]++#", function ($m) { return rawurlencode($m[0]); }, $parts[$part]);
         }
 
         return [
@@ -634,11 +634,7 @@ trait HttpClientTrait
                     '%3B' => ';',
                     '%40' => '@',
                     '%5B' => '[',
-                    '%5C' => '\\',
                     '%5D' => ']',
-                    '%5E' => '^',
-                    '%60' => '`',
-                    '%7C' => '|',
                 ]);
             }
 
--- a/src/Symfony/Component/HttpClient/Tests/HttpClientTraitTest.php
+++ b/src/Symfony/Component/HttpClient/Tests/HttpClientTraitTest.php
@@ -155,12 +155,13 @@ class HttpClientTraitTest extends TestCase
         yield [['http:', '//example.com', null, null, null], 'http://Example.coM:80'];
         yield [['https:', '//xn--dj-kia8a.example.com:8000', '/', null, null], 'https://DÉjà.Example.com:8000/'];
         yield [[null, null, '/f%20o.o', '?a=b', '#c'], '/f o%2Eo?a=b#c'];
+        yield [[null, null, '/custom%7C2010-01-01%2000:00:00%7C2023-06-15%2005:50:35', '?a=b', '#c'], '/custom|2010-01-01 00:00:00|2023-06-15 05:50:35?a=b#c'];
         yield [[null, '//a:b@foo', '/bar', null, null], '//a:b@foo/bar'];
         yield [[null, '//a:b@foo', '/b{}', null, null], '//a:b@foo/b{}'];
         yield [['http:', null, null, null, null], 'http:'];
         yield [['http:', null, 'bar', null, null], 'http:bar'];
         yield [[null, null, 'bar', '?a=1&c=c', null], 'bar?a=a&b=b', ['b' => null, 'c' => 'c', 'a' => 1]];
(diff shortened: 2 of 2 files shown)


src/Symfony/Component/HttpClient/HttpClientTrait.php
```diff
@@ -625,20 +625,18 @@ private static function mergeQueryString(?string $queryString, array $queryArray
                 // https://tools.ietf.org/html/rfc3986#section-2.3 + some chars not encoded by browsers
                 $queryString = strtr($queryString, [
                     '%21' => '!',
+                    '%23' => '#',
```
stof: this is wrong. You cannot revert the encoding of `#` in the query string as the `#` would then mark the end of the query string (starting the fragment part)

src/Symfony/Component/HttpClient/HttpClientTrait.php
```diff
@@ -625,20 +625,18 @@ private static function mergeQueryString(?string $queryString, array $queryArray
...
                     '%24' => '$',
                     '%28' => '(',
                     '%29' => ')',
                     '%2A' => '*',
                     '%2F' => '/',
                     '%3A' => ':',
                     '%3B' => ';',
+                    '%3F' => '?',
```
stof: I'm quite sure browsers keep `?` escaped in the query string to avoid risk of bad parsing.

stof (changes requested):

src/Symfony/Component/HttpClient/HttpClientTrait.php
Kozzi11: good catch make sense I have remove this change

src/Symfony/Component/HttpClient/HttpClientTrait.php
Kozzi11: Yes you are right I have removed that

src/Symfony/Component/HttpClient/HttpClientTrait.php
```diff
@@ -547,7 +547,7 @@ private static function parseUrl(string $url, array $query = [], array $allowedS
             }
 
             // https://tools.ietf.org/html/rfc3986#section-3.3
-            $parts[$part] = preg_replace_callback("#[^-A-Za-z0-9._~!$&/'()[\]*+,;=:@\\\\^`{|}%]++#", function ($m) { return rawurlencode($m[0]); }, $parts[$part]);
+            $parts[$part] = preg_replace_callback("#[^-A-Za-z0-9._~!$&/'()[\]*+,;=:@?\#%]++#", function ($m) { return rawurlencode($m[0]); }, $parts[$part]);
```
nicolas-grekas: This force-encodes characters that are not in the list.
Which means with this change, `?` and `#` won't be encoded. This looks incorrect to me (even if I don't expect these to be found in parts of valid URLs).
Also *when the original URL contain these unencoded*, there is no need to force-encode the chars removed in the diff.
Which means I would keep the current regexp.

src/Symfony/Component/HttpClient/Tests/HttpClientTraitTest.php
```diff
@@ -70,8 +70,8 @@ public static function provideResolveUrl(): array
             [self::RFC3986_BASE, '/g',            'http://a/g'],
             [self::RFC3986_BASE, '//g',           'http://g/'],
             [self::RFC3986_BASE, '?y',            'http://a/b/c/d;p?y'],
-            [self::RFC3986_BASE, '?y={"f":1}',    'http://a/b/c/d;p?y={%22f%22:1}'],
-            [self::RFC3986_BASE, 'g{oof}y',       'http://a/b/c/g{oof}y'],
+            [self::RFC3986_BASE, '?y={"f":1}',    'http://a/b/c/d;p?y=%7B%22f%22:1%7D'],
+            [self::RFC3986_BASE, 'g{oof}y',       'http://a/b/c/g%7Boof%7Dy'],
```
nicolas-grekas: these changes should be reverted, per my comment above

nicolas-grekas (commented):
Unfortunately, the RFC is very loose on the topic, and implementations varies.

src/Symfony/Component/HttpClient/HttpClientTrait.php
Kozzi11: Ok can we remove `|` char from original regexp, this one is the problematic one, which cause issues for us.

Kozzi11: And I still do not understand why the original PR https://github.com/symfony/symfony/pull/49299 has been merged to 5.4 branch? Maybe rever it from 5.4 branch would be ok too?

nicolas-grekas:
Thank you @Kozzi11.

PR #50678, opened 2023-06-15
[Contracts/Deprecation] Allow screaming

ro0NL:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #... 
| License       | MIT

This allows screaming deprecations, collected by symfony, using env var.

~Then in test env we can disable `framework.php_error.log` to handle them natively.~ This seems irrelevant, i think i misunderstood _"Use the application logger instead of the PHP logger for logging PHP errors."_.

Then in phpunit we can benefit from `displayDetailsOnTestsThatTriggerDeprecations`

Removing the `@` made it work.

But overriding the trigger_deprecation during test bootstrap did not work:

```
$ phpunit 
[15-Jun-2023 17:17:15 UTC] PHP Fatal error:  Cannot redeclare trigger_deprecation() (previously declared in /usr/local/etc/tools/vendor/symfony/deprecation-contracts/function.php:23) in /app/tests/bootstrap.php on line 5
```

### DIFF
--- a/src/Symfony/Contracts/Deprecation/function.php
+++ b/src/Symfony/Contracts/Deprecation/function.php
@@ -11,7 +11,7 @@
 
 if (!function_exists('trigger_deprecation')) {
     /**
-     * Triggers a silenced deprecation notice.
+     * Triggers a (silenced) deprecation notice.
      *
      * @param string $package The name of the Composer package that is triggering the deprecation
      * @param string $version The version of the package that introduced the deprecation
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
:-1: 

This should be improved in phpunit / the part that gets deprecations.

Also, to override the function you need to require yours before this one to make it work.

ro0NL:
> Also, to override the function you need to require yours before this one to make it work.

phpunit is globally installed, i guess it's not supported

ro0NL:
also i dont see any problem in making it configurable for whatever purpose, but OK

ro0NL:
> This should be improved in phpunit / the part that gets deprecations.

i believe it's supported in phpunit 10.2 https://github.com/sebastianbergmann/phpunit/issues/5328 :tada:

PR #50679, opened 2023-06-15
[Validator] Validate all groups when special group name `*` is specified in `validate()` method

### LINKED ISSUE #49932: Include option in validator to validate all groups
### Description

https://symfony.com/doc/current/validation/groups.html

Validation groups allows us to define groups on individual properties of objects as per the following documented example:

```
class User implements UserInterface
{
    #[Assert\Email(groups: ['registration'])]
    private $email;

    // $city belongs to special group "Default"
    #[Assert\Length(min: 2)]
    private $city;
```

We can then validate the registration group specifically by calling ` $validator->validate($author, null, ['registration']);`, or we can validate any properties which do not have an explicit group configured by using the group name `Default` or not specifying a group name, e.g. ` $validator->validate($author, null, ['Default']);`, or just ` $validator->validate($author);`

But at the moment, if we want to validate both groups, we'd have to use `$validator->validate($author, null, ['Default', 'registration']);`

This is fine when you only have maybe two or three groups to work with, including `Default`, but gets messy when you have an arbitrarily larger number of groups.

At the moment, if you want to easily validate _all_ groups, there is no built-in option for doing this except to either list every group in the call to `validate()`, or to add a group name such as `all` to every single property.

Would be nice if we could do `$validator->validate($author, null, ['All']);` or similar as a special group name which automatically applied validation to all properties, regardless of groups.

### Example

```
class User
{
    #[Assert\Email(groups: ['registration'])]
    private $email;

    #[Assert\Length(min: 2, groups:['personal'])]
    private $firstName;

    #[Assert\Length(min: 2)]
    private $city;
}

$errors = $validator->validate($user, null, ['All']);
```

dwgebler:
| Q             | A
| ------------- | ---
| Branch?       | 7.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #49932
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/18420

Adds a special group name `*` which can be used with the validator to validate all groups on the object being validated. This is comparable to how the group name `*` can be used on `Serializer` component, ref. https://symfony.com/doc/current/components/serializer.html#attributes-groups

### DIFF
--- a/src/Symfony/Component/Validator/Mapping/GenericMetadata.php
+++ b/src/Symfony/Component/Validator/Mapping/GenericMetadata.php
@@ -209,8 +209,10 @@ class GenericMetadata implements MetadataInterface
             foreach ($this->constraintsByGroup as $groupConstraints) {
                 $constraints = array_merge($constraints, $groupConstraints);
             }
+
             return $constraints;
         }
+
         return $this->constraintsByGroup[$group] ?? [];
     }
 
--- a/src/Symfony/Component/Validator/Mapping/GenericMetadata.php
+++ b/src/Symfony/Component/Validator/Mapping/GenericMetadata.php
@@ -204,6 +204,13 @@ class GenericMetadata implements MetadataInterface
      */
     public function findConstraints(string $group): array
     {
+        if ('*' === $group) {
+            $constraints = [];
+            foreach ($this->constraintsByGroup as $groupConstraints) {
+                $constraints = array_merge($constraints, $groupConstraints);
+            }
+            return $constraints;
+        }
         return $this->constraintsByGroup[$group] ?? [];
     }
 
--- a/src/Symfony/Component/Validator/CHANGELOG.md
+++ b/src/Symfony/Component/Validator/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+ * Add the special group name `*` which can be passed to `ValidatorInterface::validate()` to validate all constraints, regardless of their group.
+
 6.3
 ---
 
--- a/src/Symfony/Component/Validator/Tests/Validator/RecursiveValidatorTest.php
+++ b/src/Symfony/Component/Validator/Tests/Validator/RecursiveValidatorTest.php
@@ -1114,6 +1114,29 @@ class RecursiveValidatorTest extends TestCase
         $this->assertCount(2, $violations);
     }
 
+    public function testValidateAllGroups()
+    {
+        $entity = new Entity();
+
+        $callback = function ($value, ExecutionContextInterface $context) {
+            $context->addViolation('Message');
+        };
+
+        $this->metadata->addConstraint(new Callback([
+            'callback' => $callback,
+            'groups' => 'Group 1',
+        ]));
+        $this->metadata->addConstraint(new Callback([
+            'callback' => $callback,
+            'groups' => 'Group 2',
+        ]));
+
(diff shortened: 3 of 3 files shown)


xabbuh:
I am afraid that doing this change would be a BC break in case someone actually uses this "special" group.

ro0NL:
From the issue, im curious about your usecase

or put different, IMHO context should tell you with groups to validate, isnt it

there can also be contractions between groups IIUC

dwgebler:
> From the issue, im curious about your usecase
> 
> or put different, IMHO context should tell you with groups to validate, isnt it
> 
> there can also be contradictions between groups IIUC

I had a case of a large object graph (so object to be validated had several other objects with `Valid` constraint on them, which all also needed to be validated) and there were about 25 groups in total. So to validate the entire object every possible group on the object and every object it contained had to be specified.

It very much seems like there should be an easy way to validate an entire object regardless of group metadata; it doesn't strike me as a particularly niche case.

dwgebler:
> I am afraid that doing this change would be a BC break in case someone actually uses this "special" group.

I'm happy to change the approach if necessary (open to discussion on that), or target 7.0.

ro0NL:
i guess it could work in some cases, till it doesnt, eg.:

```php
#[Blank(groups: ['a'])]
#[NotBlank(groups: ['b'])]
public $a;
```

but maybe that's more of a user concern :+1: 

> @param string|GroupSequence|array<string|GroupSequence>|null $groups

maybe support `true` here?

dwgebler:
> i guess it could work in some cases, till it doesnt, eg.:
> 
> ```
> #[Blank(groups: ['a'])]
> #[NotBlank(groups: ['b'])]
> public $a;
> ```
> 
> but maybe that's more of a user concern 👍

Yeah I mean you'll get a violation there, but that's expected behaviour because the constraints can't both be valid. If you had contradicting constraints, it'd be a solid indicator validating everything at once isn't your use case.

> 
> > @param string|GroupSequence|array<string|GroupSequence>|null $groups
> 
> maybe support `true` here?

Possibly. I'd be interested to get more feedback on this and other options from other maintainers before I update the PR with any new approach.

dwgebler:
@nicolas-grekas @fabpot had this PR open for about 18 months. I still believe it's a valuable feature to provide a way to validate all constraint groups on an object. If there are concerns still about the validation group name `*` causing a BC-break, can we either target next major version or re-open a conversation about how else this could be solved?

src/Symfony/Component/Validator/Mapping/GenericMetadata.php
```diff
@@ -204,6 +204,15 @@ public function hasConstraints(): bool
      */
     public function findConstraints(string $group): array
     {
+        if ('*' === $group) {
```
GromNaN, line 207: To work with BC break, you can deprecate the use of `*` as a group name in 7.4 and throw an exception in this case in 8.0. You can then use this special group name in 8.1 (in 1 year).

nicolas-grekas:
Closing as stale. Taking '*' as a special group name is a BC break for anyone already using it as an ordinary group, as xabbuh pointed out in 2023, and the deprecation path GromNaN proposed in 2025 (deprecate the name first, give it its new meaning a major later) was never started, so the 7.4 and 8.0 window it relied on has passed. The feature itself was never rejected. If you still want it, please open a fresh PR that starts with the deprecation on the current branch and we will pick the discussion back up. Thanks for your patience.

PR #48554, opened 2022-12-08
[Security] Fix invalid deprecation messages in Security constants

IonBazan:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

This PR changes invalid deprecation message, suggesting to use `\Symfony\Bundle\SecurityBundle\Security::ACCESS_DENIED_ERROR` instead of `\Symfony\Component\Security\Core\Security::ACCESS_DENIED_ERROR`, while `\Symfony\Component\Security\Http\SecurityRequestAttributes::ACCESS_DENIED_ERROR` should be used.

For reference: #47760

PR #50680, opened 2023-06-16
Mark class-constants as non-deprecated

heiglandreas:
| Q             | A
| ------------- | ---
| Branch?       | 6.2 
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | none

Static analysis tools like PHPStan treat the `@deprecated` flag as inherited. So when a class constant is marked as deprecated, then the same class constant in an extending class is also marked deprecated unless it gets explicitly marked as `@non-deprecated`.

Without this flag PHPStan currenlty will return the message 

```
Fetching deprecated class constant LAST_USERNAME of class Symfony\Bundle\SecurityBundle\Security:                          
since Symfony 6.2, use \Symfony\Bundle\SecurityBundle\Security::LAST_USERNAME instead 
```
which is not really helpful as that is exactly what one did at that point.

By marking these class-constants as `@non-deprecated` PHPStan and other static analysis tools can recognize these class constants as being the rightly used ones.

For more information have a look at the discussion at https://github.com/phpstan/phpstan-deprecation-rules/issues/100

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/Security.php
+++ b/src/Symfony/Bundle/SecurityBundle/Security.php
@@ -38,8 +38,13 @@ use Symfony\Contracts\Service\ServiceProviderInterface;
  */
 class Security extends LegacySecurity
 {
+    /** @non-deprecated */
     public const ACCESS_DENIED_ERROR = SecurityRequestAttributes::ACCESS_DENIED_ERROR;
+
+    /** @non-deprecated */
     public const AUTHENTICATION_ERROR = SecurityRequestAttributes::AUTHENTICATION_ERROR;
+
+    /** @non-deprecated */
     public const LAST_USERNAME = SecurityRequestAttributes::LAST_USERNAME;
 
     public function __construct(private readonly ContainerInterface $container, private readonly array $authenticators = [])


derrabus:
Maybe we should actually deprecate them as well, given that the constants are duplicated on the `SecurityRequestAttributes` class and we somewhat create two sources of truth for the same thing.

I probably didn't fully understand the rationale behind #48554 though. 🙈 /cc @IonBazan @chalasr

ro0NL:
`@non-deprecated` looks weird, is it a standard? IMHO redeclaring implicitly undeprecates it

also https://psalm.dev/r/f642da203a https://phpstan.org/r/89339c04-ac08-4f6a-b0ab-241f60670c3f

2 tools, 1 lang does not work :)

wouterj:
I'm -1 on adding these annotations to Symfony, as they violate our [phpdoc standards](https://symfony.com/doc/current/contributing/code/standards.html#documentation). Given only one tool (with an extra ruleset) uses this annotation as far as I know, this better belongs in the PHPstan stubs (as also suggested by Ondrej).

But I'm +1 on deprecating these constants as well in favor of `SecurityRequestAttributes`. It does not make sense to have them in both places imho. Would love to hear @chalasr's opinion here, given the previous PR undeprecated these.

nicolas-grekas:
I'm closing as discussed, but I'm also looking forward to the PR that will deprecate these consts!

PR #50687, opened 2023-06-16
[EventDispatcher] add a way to call a listener before or after another one (WIP)

nikophil:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT
| Doc PR        | will be done if feature gets accepted

When hooking into the event system, it's a common need to define the priority compared to another existing listener:
```php
// should occur BEFORE "router_listener"
#[AsEventListener(event: KernelEvents::REQUEST, priority: 33)]

// OR

// should occur AFTER "router_listener"
#[AsEventListener(event: KernelEvents::REQUEST, priority: 31)]
```
This is not optimal: without the comment it is not clear why the listener has the given priority. Moreover the order is not enforced by code and the priority of the other listener can change.

I'd like to propose another approach, and introduce a new way to handle priorities between listeners:
```php
// the listener will occur before "router_listener::onKernelRequest()"
#[AsEventListener(event: KernelEvents::REQUEST, before: RouterListener::class)] 

 // the listener will occur after "router_listener::onKernelRequest()"
#[AsEventListener(event: KernelEvents::REQUEST, after: RouterListener::class)]
```

`before` and `after` priority should follow these rules:
- it can be declared as a simple string:
    - another listener's serviceId, if it only have one method listening to the same event
    - a listener's class name if a service is declared for this class AND if it has only one method listening to the same event
- it can be declared as an array:
    - {0: 'some_service_id', 1: 'someMethodName'}
    - {0: SomeListenerClass::class, 1: 'someMethodName'} a class name is accepted if a service is declared with this class

The idea is to rely on the priority system and to guess the priority of the listener based on its "before" or "after" attributes (the priority computation would be recursive if the given before/after listener should also occur before/after another one). 

Here are some rules:
- if a listener uses before/after, it must listen the same event than the listener which is referenced
- a listener cannot have a priority configured and use before/after
- a listener cannot declare at the same time "before" and "after"
- we should detect circular references

Here is a first draft PR of the implementation, it's missing a lot of stuff:
- tests
- ability to define before/after in php, xml, yaml
- doing the same thing with subscribers
- certainly some edge cases are not handled correctly here

at this point, I'd like to have early feedback about the idea and if it has some chances to be accepted. and if the given implementation is the way to go? 

I think if this feature is accepted, it would be nice to commit it to all tags using priorities, and then add it to `PriorityTaggedServiceTrait` 

thanks for your feedbacks!


kbond (commented):
I like this. Didn't look to deep into the code but it would fail during compile if the before/after service id changed?

Maybe before/after could be configured as a _quasi_ callable (attributes can't accept real callables) to help with IDE click throughs and refactoring?

```php
/** @var string|class-string|callable|null */
public string|array|null $before = null,

before: [MyListener::class, 'myMethod']

before: MyListener::class // uses __invoke
```

nikophil:
> it would fail during compile if the before/after service id changed?

Yes, you'll have a `\Symfony\Component\DependencyInjection\Exception\InvalidArgumentException` with the following message: `Given "before" "some_service::some_method" for listener "your_listener::your_method" does not listen the same event.`

But maybe I should fail before, and directly say `Given "before" "some_service::some_method" for listener "your_listener::your_method" does not exists.`

> Maybe before/after could be configured as a quasi callable (attributes can't accept real callables) to help with IDE click throughs and refactoring?

this is a bit tricky since we're working with service ids and not classes...

kbond:
> this is a bit tricky since we're working with service ids and not classes...

For internal event listeners, yes. I'm thinking about ones in your app that you probably have autowired as the class name.

nikophil:
yes ok, I think I definitely need to be more flexible how we can name the listeners:
- "quasi callable" as you said
- maybe "first class callable": `MyListener::myMethod(...)` (not sure the syntax is allowed in the attributes :thinking: )
- also, if the class only have one public method, we may allow to only pass its service id
- maybe also allow to pass the classes FQCN instead of the service id? I mean: `before: RequestListener::class. .'::onKernelRequest()'` but I'm not sure if I really can rely on this?

kbond:
> maybe "first class callable": MyListener::myMethod(...) (not sure the syntax is allowed in the attributes thinking )

I'd prefer this but yeah, don't know if it's possible.

> maybe also allow to pass the classes FQCN instead of the service id? I mean: before: RequestListener::class. .'::onKernelRequest()' but I'm not sure if I really can rely on this?

This would be nice but we'd need to throw an ambiguous exception if there are multiple service ids for this class.

nikophil:
> I'd prefer this but yeah, don't know if it's possible.

not possible: https://3v4l.org/9HYHJ#v8.2.7

> This would be nice but we'd need to throw an ambiguous exception if there are multiple service ids for this class.

ok, I'm gonna do that!

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -182,6 +167,130 @@ private function getEventFromTypeDeclaration(ContainerBuilder $container, string
...
+        foreach ($container->findTaggedServiceIds('kernel.event_listener', true) as $id => $events) {
+            foreach ($events as $event) {
+                if (!isset($event['event']) && $container->getDefinition($id)->hasTag('kernel.event_subscriber')) {
+                    continue;
+                }
+
+                $this->guessEventAndMethod($event, $container, $id, $aliases);
+
+                if (isset($event['before']) || isset($event['after'])) {
+                    if (isset($event['before']) && isset($event['after'])) {
```
mtarld: You can leverage `xor` I guess, to avoid `&&` and `||` combination

src/Symfony/Component/EventDispatcher/Attribute/AsEventListener.php
```diff
@@ -24,6 +24,8 @@ public function __construct(
         public ?string $method = null,
         public int $priority = 0,
         public ?string $dispatcher = null,
+        public ?string $before = null,
```
mtarld: Could it at some point be an array shaped like: `array{service: string, method=: string}` (or even a DTO) to get rid of the string convention?

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -182,6 +167,130 @@ private function getEventFromTypeDeclaration(ContainerBuilder $container, string
...
+                ], fn ($matches) => strtoupper($matches[0]), $eventDefinition['event']);
+            $eventDefinition['method'] = preg_replace('/[^a-z0-9]/i', '', $eventDefinition['method']);
+
+            if (null !== ($class = $container->getDefinition($serviceId)->getClass()) && ($r = $container->getReflectionClass($class, false)) && !$r->hasMethod($eventDefinition['method']) && $r->hasMethod('__invoke')) {
+                $eventDefinition['method'] = '__invoke';
+            }
+        }
+    }
+
+    private function normalizeBeforeAfter(string|null $beforeAfter, array $aliases): string|null
```
mtarld:
```suggestion
    private function normalizeBeforeAfter(string $beforeAfter, array $aliases): string|null
```
And avoid to call that method if `beforeAfter` is null?

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -182,6 +167,130 @@ private function getEventFromTypeDeclaration(ContainerBuilder $container, string
...
+        if (null === $beforeAfter) {
+            return null;
+        }
+
+        $beforeAfter = explode('::', $beforeAfter);
+
+        return sprintf(
+            '%s::%s',
+            $aliases[$beforeAfter[0]] ?? $beforeAfter[0],
+            trim($beforeAfter[1] ?? '__invoke', '()')
```
mtarld:
```suggestion
            rtrim($beforeAfter[1] ?? '__invoke', '()')
```

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -182,6 +167,130 @@ private function getEventFromTypeDeclaration(ContainerBuilder $container, string
...
+        );
+    }
+
+    private function computePriority(array $listeners, string $listenerName, array $alreadyVisited = []): int
+    {
+        if ($alreadyVisited[$listenerName] ?? false) {
```
mtarld:
```suggestion
        if (isset($alreadyVisited[$listenerName]) {
```
Might be faster

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -182,6 +167,130 @@ private function getEventFromTypeDeclaration(ContainerBuilder $container, string
...
+
+        foreach ($events as &$event) {
+            if ($event['event'] === $eventName) {
+                $event['priority'] = $priority;
+                break;
+            }
+        }
+
+        $listenerDefinition->clearTag('kernel.event_listener');
+        $listenerDefinition->addTag('kernel.event_listener', $events);
```
mtarld: By doing that, every `kernel.event_listener` tag will have the same priority, no matter the event, isn't that an issue?

mtarld (changes requested):
Nice addition!

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
nikophil: either I don't understand your point, or I think you're wrong: 
see previous loop, and notice the condition:
```php
        foreach ($events as &$event) {
            if ($event['event'] === $eventName) {
                $event['priority'] = $priority;
                break;
            }
        }
```

AFAIK there is no way to update only ONE event/listener definition for a given service

mtarld: Indeed, I was wrong 😅

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
nikophil: ok but it is such a micro optim! :sweat_smile: 
https://3v4l.org/sY5ZR#v8.2.7

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
nikophil: I'll actually change it to `trim($beforeAfter[1] ?? '__invoke', ' ()')` (with the space)

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
nikophil: since the method is called twice, moreover in an already complex method, I'd rather to keep this check here

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
nikophil: you're right! but this code does no exist anymore (not pushed yet)

src/Symfony/Component/EventDispatcher/Attribute/AsEventListener.php
nikophil: indeed, I was wondering about introducing a value object
but it is actually incompatible with @kbond's [comment](https://github.com/symfony/symfony/pull/50687#pullrequestreview-1483529449) about using "quasi callable" which would be a nice DX addition (PHPStorm would rename the class/method in case of refactoring). I wish first class callable would be allowed in attributes definition :disappointed: 

(EDIT: I just played around with this and even PHPStorm would not understand the method's name, so there is no problem with your suggestion)

Since `before` and `after` are incompatible, I'm wondering if I should not merge them, ie:
```php
#[AsEventListener(
    event: KernelEvents::REQUEST,
    priorityHook: new PriorityHookOrder(
        // I'd like to allow to add classes here, such as `RouteListener::class` 
		// as soon as only one listener is declared for the given event on this class
        serviceIdOrClass: 'route_listener',  
        method: 'onKernelRequest',
        // an enum would be best here, not sure if it will fit into the whole DI stuff, with service declaration in xml, yaml...
        order: PriorityHookOrder::BEFORE 
    )
)]
```
not sure about the naming, but I cannot find a better name... 

WDYT?

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
nikophil: I think I'll get rid of the name convention and rely on the pair `[serviceId, ?method]`

I'm closing this

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -182,6 +167,130 @@ private function getEventFromTypeDeclaration(ContainerBuilder $container, string
...
+                    );
+                }
+            }
+        }
+    }
+
+    private function guessEventAndMethod(array &$eventDefinition, ContainerBuilder $container, string $serviceId, array $aliases): void
+    {
+        if (!isset($eventDefinition['event'])) {
+            $eventDefinition['method'] ??= '__invoke';
```
GromNaN: Usually, classes have only 1 method that listen to a given event. Instead of assuming no-method is `__invoke`, the behavior should be to find the method of the class that have a listener for this event. Or throw an exception if there is not exactly 1.
This would solve en "pseudo-closure" issue by just setting the class name in the `AsEventListener` attribute.

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
chalasr: Yet more explicit IMHO

src/Symfony/Component/EventDispatcher/Attribute/AsEventListener.php
GromNaN: > Could it at some point be an array shaped like: `array{service: string, method=: string}` (or even a DTO) to get rid of the string convention?

It should be `class-string|array{class?: class-string, service?: string, method?: string}`

---

This `priorityHook` is more verbose. There is an exception at compile-time if both `before` and `after` are used, so the current proposition is fine.

stof: even if first class callable was available in attributes, this would not be usable here: to get a callable for the RoutingListener `onKernelRequest` method, you need the *instance* of the RoutingListener as it is not a static method. `RoutingListener::onKernelRequest(...)` is a PHP error saying that the method cannot be called statically.

nikophil: yes, indeed you're right!

GromNaN: Only specifying the class name or service name should cover most of the use-cases. The method name can be guessed when a class have only 1 method listening to an event. https://github.com/symfony/symfony/pull/50687#discussion_r1239639330

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
nikophil: Yes, I'm perfectly fine with this, but this code was [extracted from the original compiler pass](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php#L74-L81), 
So changing this would change the main behavior of how the listener method is resolved.

By the way, I've just noticed what I thing is a bug while playing around with this:
```php
#[AsEventListener(event: KernelEvents::REQUEST, priority: 5)]
class MyListenerNotInvokable
{
    public function someMethod(RequestEvent $event): {}
}
```
It registers a	listener on `MyListenerNotInvokable::onKernelRequest()` and then an error is dispatched at runtime: `Error: Call to undefined method App\EventListener\MyListenerNotInvokable::onKernelRequest()`. Whereas the correct behavior would be to resolve the listener method to the only public method in the class (or throw an exception if several public methods exist)

I think I'll fix this in another PR before finishing the work here

GromNaN: I understand that you've extracted the existing code, but these are 2 different things.
- the existing code finds the method to call when declaring a listener by analyzing the class (or with a fallback to a default method name). I think it's dangerous to use the public method if it's unique. There's a risk that the code will break if you have to add a new public method.
- the new code must be based on existing listeners only. It's then no longer a question of detecting public methods, but only of searching among the methods attached to a listener on the event concerned.

nikophil: yes I totally agree but don't you think that registering a listener on method `MyListenerNotInvokable::onKernelRequest()` (see example above) which does not exist is a mistake? which by the way throws a runtime exception as soon as the event is dispatched, although we could have detected the problem at compile-time. See [this code](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php#L85-L95) which seems buggy: if the computed method name does not exist, and if `__invoke` does not exist, then we still keep the wrong method name.

What I have in mind is to remove the automatic fallback on `__invoke` [here](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php#L79) and [here](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php#L93) and replace it with a more fine-grained method resolving (would be done in another PR)

> Or throw an exception if there is not exactly 1.

this would be done exclusively for listeners declaring before/after, when resolving their priority: if not enough information is provided (ie: only class/serviceId is provided and the given class contains multiple methods which listen to the same event) then we'd throw an exception because before/after declaration is ambiguous

maybe these are two different things, but the latter will benefit from the former because I've reused all the code in `RegisterListenersPass::process()` to iterate on listeners definitions

WDYT?

alexisLefebvre:
If 2 listeners subscribe to the same “after X”, which one will be called first? Could we define the priority of each listener?

kbond:
> If 2 listeners subscribe to the same “after X”, which one will be called first? Could we define the priority of each listener?

I would say the order shouldn't matter here. If it does, you'd need to set one to be called after the other.

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
GromNaN: > yes I totally agree but don't you think that registering a listener on method `MyListenerNotInvokable::onKernelRequest()` (see example above) which does not exist is a mistake? which by the way throws a runtime exception as soon as the event is dispatched, although we could have detected the problem at compile-time. See [this code](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php#L85-L95) which seems buggy: if the computed method name does not exist, and if `__invoke` does not exist, then we still keep the wrong method name.

This is indeed a bug. Would you mind opening a PR or an issue?

> maybe these are two different things, but the latter will benefit from the former because I've reused all the code in RegisterListenersPass::process() to iterate on listeners definitions

I'm still not convinced that it can be the same mechanism for both. For the before/after feature, we know the actual listeners of the given event, which restrict the list of possible method names. I would not modify in this PR the way the listener method are guessed when they are register.

src/Symfony/Component/EventDispatcher/Attribute/AsEventListener.php
nikophil: I'm closing this one, discussion goes further in https://github.com/symfony/symfony/pull/50687#discussion_r1239639330

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
nikophil: > This is indeed a bug. Would you mind opening a PR or an issue?

yep I'll open a PR soon, I'll ping you there

> I'm still not convinced that it can be the same mechanism for both. For the before/after feature, we know the actual listeners of the given event, which restrict the list of possible method names. I would not modify in this PR the way the listener method are guessed when they are register.

I've almost finished the work on the compiler pass with the definition you suggested:
```php
class AsEventListener
{
    public function __construct(
        public ?string $event = null,
        public ?string $method = null,
        public int $priority = 0,
        public ?string $dispatcher = null,
        /** @var string|array{class?: class-string, service?: string, method?: string} */
        public string|array|null $before = null,
        /** @var string|array{class?: class-string, service?: string, method?: string} */
        public string|array|null $after = null,
    ) {
    }
}
```
I'll push the code soon, the discussion will be easier then :)

nikophil: just a word about this:

> I would not modify in this PR the way the listener method are guessed when they are register.

even if I've refactored the compiler pass, I've touched nothing about the logic and the way listener methods are guessed.
All I do is iterating on the listener definitions, and recomputing the priority of the listeners which have before/after properties

alexisLefebvre:
> > If 2 listeners subscribe to the same “after X”, which one will be called first? Could we define the priority of each listener?
> 
> I would say the order shouldn't matter here. If it does, you'd need to set one to be called after the other.

Like chaining them? Is it possible with the changes from this PR? Or is it a special case that require to define priorities and is out of scope for this PR?

nikophil:
Hi @alexislefebvre 

yes there is no problem with the current implementation to do something like 
```php
#[AsEventListener(event: SomeEvent::class, priority: 10)]
class SomeListener{}

// will have priority 11
#[AsEventListener(event: SomeEvent::class, before: SomeListener::class)]
class SomeOtherListener{}

// will have priority 12
#[AsEventListener(event: SomeEvent::class, before: SomeOtherListener::class)]
class AnotherListener{}

// will have priority 10 (notice "after")
#[AsEventListener(event: SomeEvent::class, after: SomeOtherListener::class)]
class YetAnotherListener{}
```

src/Symfony/Component/EventDispatcher/Attribute/AsEventListener.php
```diff
@@ -24,6 +24,10 @@ public function __construct(
         public ?string $method = null,
         public int $priority = 0,
         public ?string $dispatcher = null,
+        /** @var string|array{class?: class-string, service?: string, method?: string} */
```
nikophil: not `class-string|array{class?: class-string, service?: string, method?: string}` [as suggested](https://github.com/symfony/symfony/pull/50687#discussion_r1239646002) because we accept service id as well as class names

nikophil:
I've updated the PR.

before/after can be defined as: a class name, a service name or a pair of [class, method] or [service, method].
If the service or the class is defined alone, then we can resolve it if and only if it defines one listener for the given event.

we have to create 2 "maps" of the listener definitions: 
- one which indexes all the listeners/subscribers definitions 
- one based on the classes FQCN where we can easily find listener/subscriber definitions based on a class name

Thus, we need to loop over all listeners/subscribers (indistinctly: a listener can be defined before/after a subscriber and the other way round) before actually registering them in the event dispatcher. I needed to create a normalized representation of all listeners/subscribers definitions, with their method and event resolved, so I extracted the logic from `RegisterListenersPass::process()` which now leverage this new representation. I actually hesitated to introduce a value object for `ListenerDefinition`, but for now I've used a psalm pseudo-type.

Because before/after logic was getting bigger and bigger, I decided to split the compiler pass, hence the trait `ListenersPassTrait` because we need to share the logic of the listeners definitions extraction.

I tried to limit the "array hell" effect, but it's kinda hard since we need a hash map which is indexed by event, service id and method.

I hope that the implementation is OK and that I did not over-engineered all the things :sweat_smile: 
I'd love to have your feedbacks @GromNaN @kbond @chalasr 

Once we've found a satisfying solution, I'll provide the missing parts of the PR (tests, xml/yaml/php configuration, implement before/after for subscribers)

kbond:
> - it can be declared as an array:
>     - ['service': 'some_service_id', 'method': 'someMethodName']

Are the array key's required? Would be nice if `['some_service', 'someMethod]` and `[SomeClass::class, 'someMethod']` just worked.

> I think if this feature is accepted, it would be nice to commit it to all tags using priorities, and then add it to PriorityTaggedServiceTrait

oo good call, that would be nice. Controller value resolvers come to mind.

> - it can be declared as a simple string:
>     - another listener's serviceId, if it only have one method listening to the same event
a listener's class name if only one service is declared for this class AND if it has only one method listening to the same event

To clarify, it would be fine if the service/class listened to other events?

nikophil:
> Are the array key's required? Would be nice if ['some_service', 'someMethod] and [SomeClass::class, 'someMethod'] just worked.

they're not required, but I'd say they are useful for autodiscovery
EDIT: I'm actually tempted to remove the keys, and force to provide the method. Currently it is possible to do `before: ['service': 'some_service']` which is equivalent to `before: 'some_service'`

> >I think if this feature is accepted, it would be nice to commit it to all tags using priorities, and then add it to PriorityTaggedServiceTrait
> oo good call, that would be nice. Controller value resolvers come to mind.

Yes, and normalizers :smile: 
But it would be a totally different PR. I think both functionalities won't share any code

> To clarify, it would be fine if the service/class listened to other events?

Yes, since before/after MUST listen to the same event, we can restrain the list to the methods listening the same event.
For instance, following code perfectly works, although `RouteListener` has multiple subscribers, but only one applies to `kernel.request`
```php
#[AsEventListener(event: KernelEvents::REQUEST, before: RouteListener::class)]
```

kbond:
> EDIT: I'm actually tempted to remove the keys, and force to provide the method. Currently it is possible to do before: ['service': 'some_service'] which is equivalent to before: 'some_service'

So the typehint would be: `string|class-string|array{0:string|class-string,1:string}`?

nikophil:
since `string` encompasses `class-string`, it should be `string|array{0:string,1:string}`
I actually think static analysis tools yell when you do things like `string|class-string` or `object|SomeClass`

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
nikophil: this is outdated

nikophil:
I've updated the implementation:
- kept only one compiler pass where we resolve "before/after" priorities before registering the listeners: the former needs listeners and subscribers to be handled the same way, so they are normalized as a psalm pseudo type `ListenerDefinition` (see `collectListeners()` and `collectSubscribers()` methods) Then, in order to not collect and resolve twice listeners and subscribers, I've reworked how the listeners are registered in the event dispatcher (see `registerListeners()`).
- changed how the before/after can be defined: a service id or a class if they only have one listener for the given event OR a tuple [serviceId|class, method]
- added tests for all cases (happy paths and compile errors) for `RegisterListenersPass`

TemirkhanN:
Looks good if it is for the symfony internal usage.   
Arranging the priority on the application level might bring some complications since one listener can't know how another listener is implemented and thus unable to guarantee that there is no async process initiated underneath (which basically nullifies the idea behind the priority).

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -22,12 +22,30 @@
 
 /**
  * Compiler pass to register tagged services for an event dispatcher.
+ *
+ * @psalm-type BeforeAfterDefinition = string|array{0: string, 1: string}
+ * @psalm-type ListenerDefinition = array{serviceId: string, event: string, method: string, before?: BeforeAfterDefinition, after?: BeforeAfterDefinition, priority?: int, dispatchers: list<string>}
```
stof: We don't use psalm type aliases in symfony

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -22,12 +22,30 @@
...
  */
 class RegisterListenersPass implements CompilerPassInterface
 {
     private array $hotPathEvents = [];
     private array $noPreloadEvents = [];
 
+    /**
+     * $allListenersMap['event_name']['listener_service_id']['method'] => $listenerDefinition.
+     *
+     * @var array<string, array<string, array<string, ListenerDefinition>>>
```
stof: using a psalm type alias in `@var` rather than `@psalm-var` is a very bad idea. another tool that does not understand the psalm-specific annotation will read this type as using a ListenerDefinition class, being totally wrong.

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -22,12 +22,30 @@
...
+     */
+    private array $allListenersMap = [];
+
+    /**
+     * $listenerClassesMap['listener_FQCN']['event_name'][] => array{serviceId: string, method:string}.
+     *
+     * @var array<string, array<string, list<array{serviceId: string, method:string}>>>
+     */
+    private array $listenerClassesMap = [];
```
stof: state stored during the processing **must** be released at the end of the processing.

Or even better, avoid storing state in properties but pass those maps as argument to the methods needing them.

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -84,38 +313,36 @@ public function process(ContainerBuilder $container)
 
                 if (!isset($event['method'])) {
                     $event['method'] = 'on'.preg_replace_callback([
-                        '/(?<=\b|_)[a-z]/i',
-                        '/[^a-z0-9]/i',
-                    ], fn ($matches) => strtoupper($matches[0]), $event['event']);
+                            '/(?<=\b|_)[a-z]/i',
+                            '/[^a-z0-9]/i',
+                        ], fn ($matches) => strtoupper($matches[0]), $event['event']);
```
stof: this change is wrong

src/Symfony/Component/EventDispatcher/Attribute/AsEventListener.php
```diff
@@ -24,6 +24,10 @@ public function __construct(
         public ?string $method = null,
         public int $priority = 0,
         public ?string $dispatcher = null,
+        /** @var string|array{0: string, 1: string}|null */
```
stof: IIRC, we use `@param` instead to describe types for constructor promoted properties needing them.

src/Symfony/Component/EventDispatcher/DependencyInjection/InvalidBeforeAfterListenerDefinitionException.php
```diff
@@ -0,0 +1,81 @@
...
+ * @psalm-import-type BeforeAfterDefinition from RegisterListenersPass
+ */
+final class InvalidBeforeAfterListenerDefinitionException extends InvalidArgumentException
+{
+    private function __construct(string $errorServiceId, string $message)
+    {
+        parent::__construct(sprintf('Invalid before/after definition for service "%s": %s', $errorServiceId, $message));
+    }
+
+    public static function beforeAndAfterAtSameTime(string $errorServiceId): self
```
stof: We generally don't use named constructor for exceptions in Symfony. Build the exception message in the compiler pass in the place triggering the error.

I'm not even sure we need a dedicated exception class. `Symfony\Component\DependencyInjection\Exception\InvalidArgumentException` could be used directly.

stof (changes requested):

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
nikophil: would you recommend that I use array shapes in PHPDoc everywhere instead?
I was also wondering if I should create an actual class for this, WDYT?

stof: I would use the array shape directly.

Anyway, if you follow my other comments, it will remove many usages.

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
nikophil: done, I've removed the state, but now have much more parameters (`computeBeforeAfterPriorities()` have 5 params)

src/Symfony/Component/EventDispatcher/DependencyInjection/InvalidBeforeAfterListenerDefinitionException.php
nikophil: done, but I feel like it is a bad move and makes the code less readable

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
nikophil: done, but now we have a lot of cryptic phpdoc really less understandable by humans :sweat_smile:

nicolas-grekas:
I'm seeing a lot of activity here but I'm also seeing a lot code, and that scares me a bit.
Priorities are so much easier to manage from a maintainer pov.

I think we can simplify a lot the logic by turning before/after attributes into pure verification steps: priorities work as usual and they're still the only way to specify the order, but then we throw an exception if the before/after attributes don't work.

WDYT?

nikophil:
Hi @nicolas-grekas

I understand your concerns about the big amount of code in the PR.

if we keep before/after definition like we imagined (ie: could be a class, a service id or an array [class|serviceId, method]) I'm not really sure we could save that much code: we'll still need to compute both `$listenersMap` and `$listenerClassesMap` and we'll still need the big method `RegisterListenersPass::normalizeBeforeAfter()`

> Priorities are so much easier to manage from a maintainer pov.

Not sure to understand what you mean here :thinking:

nicolas-grekas:
I mean the current code is fine in terms of complexity.

> if we keep before/after definition like we imagined

I think we should take the simplest option: we handle only  `[serviceId, method]` and we check if the required order is respected. If not, we throw. An array_search can do it (to get the resolved order after priorities have been used).

nicolas-grekas:
the current code => the one in Symfony (not this PR :) )

nikophil:
Hi,

I've tried a rewrite of the PR with an object oriented approach. I think the code is really cleaner and simpler, any thought before I change  before/after attributes into pure verification steps?

fabpot:
I agree with @nicolas-grekas that it adds a lot of complexity to the code for a problem that was never reported before.
Priorities for built-in listeners have not changed for years, and using numbers works well enough.
Maybe we can try to implement Nicolas' suggestion?

diimpp:
> for a problem that was never reported before.

For me it's mainly useful if vendor code like `sylius/sylius` defines multiple listeners with non-zero priority and it's important to execute my code before or after. As result App listener definitions becomes ridden with "magic numbers" like `priority: 47`, which can be solved with "before" or "after".

Though it's pretty much edge case.

fabpot:
Closing as I still don't think this is something we need to support like implemented here. Thank you for proposing.

PR #50691, opened 2023-06-17
[Console] Aligned multiline text in vertical table

### LINKED ISSUE #50632: Multiline fields in vertical console tables
### Description

Currently, when displaying a console table in vertical orientation, multiline data is rendered incorrectly.

image (8)

It would make sense to split the cell on newlines, and pad them so they all stay in the same column:

image (7)

Would this be a good addition (pr is easy enough), or are we breaking BC or other reasons Symfony doesn't do this?

### Example

_No response_

jaytaph:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Tickets       | Fix #50632 
| License       | MIT

When using a vertical table helper in the console, any multi-line (\n) text starts on a new line underneath the header. This fixes so the extra lines are aligned with the first line in the table.

### DIFF
--- a/src/Symfony/Component/Console/CHANGELOG.md
+++ b/src/Symfony/Component/Console/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * Add `SignalMap` to map signal value to its name
+ * Multi-line text in vertical tables is aligned properly
 
 6.3
 ---
--- a/src/Symfony/Component/Console/Helper/Table.php
+++ b/src/Symfony/Component/Console/Helper/Table.php
@@ -364,14 +364,26 @@ class Table
                 $maxRows = max(\count($headers), \count($row));
                 for ($i = 0; $i < $maxRows; ++$i) {
                     $cell = (string) ($row[$i] ?? '');
-                    if ($headers && !$containsColspan) {
-                        $rows[] = [sprintf(
-                            '<comment>%s</>: %s',
-                            str_pad($headers[$i] ?? '', $maxHeaderLength, ' ', \STR_PAD_LEFT),
-                            $cell
-                        )];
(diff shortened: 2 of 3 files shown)


jaytaph:
Seems that fabbot wants to strip away the trailing spaces in the table helper unit-tests, but in these cases they need to be present

chalasr (approved):

maxbeckers (approved):

src/Symfony/Component/Console/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+* Multi-line text in vertical tables is aligned properly
```
chalasr:
```suggestion
 * Multi-line text in vertical tables is aligned properly
```

GromNaN (approved):

src/Symfony/Component/Console/Helper/Table.php
```diff
@@ -364,14 +364,26 @@ public function render()
...
-                            str_pad($headers[$i] ?? '', $maxHeaderLength, ' ', \STR_PAD_LEFT),
-                            $cell
-                        )];
-                    } elseif ('' !== $cell) {
-                        $rows[] = [$cell];
+
+                    $parts = explode("\n", $cell);
+                    foreach ($parts as $idx => $part) {
+                        if ($headers && !$containsColspan) {
+                            if (0 == $idx) {
```
stloyd: Should be strict probably:
```suggestion
                            if (0 === $idx) {
```

fabpot (approved):

fabpot:
Thank you @jaytaph.

PR #50694, opened 2023-06-18
[SecurityBundle] Remove last usages of tag `security.remember_me_aware`

alamirault:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #... 
| License       | MIT

`security.remember_me_aware` was deprecated in 5.4 an tag was removed in 6.0 branch.

This PR remove last references still remaining in source code.

Original issue on [symfony-docs](https://github.com/symfony/symfony-docs/issues/18242)

Friendly cc @wouterj

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/RememberMeFactory.php
+++ b/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/RememberMeFactory.php
@@ -111,15 +111,6 @@ class RememberMeFactory implements AuthenticatorFactoryInterface, PrependExtensi
             ->replaceArgument(3, $config['name'] ?? $this->options['name'])
         ;
 
-        foreach ($container->findTaggedServiceIds('security.remember_me_aware') as $serviceId => $attributes) {
-            // register ContextListener
-            if (str_starts_with($serviceId, 'security.context_listener')) {
-                continue;
-            }
-
-            throw new \LogicException(sprintf('Symfony Authenticator Security dropped support for the "security.remember_me_aware" tag, service "%s" will no longer work as expected.', $serviceId));
-        }
-
         return $authenticatorId;
     }
 
--- a/src/Symfony/Bundle/SecurityBundle/DependencyInjection/SecurityExtension.php
+++ b/src/Symfony/Bundle/SecurityBundle/DependencyInjection/SecurityExtension.php
@@ -417,11 +417,10 @@ class SecurityExtension extends Extension implements PrependExtensionInterface
         $listeners[] = new Reference('security.channel_listener');
 
         $contextKey = null;
-        $contextListenerId = null;
         // Context serializer listener
         if (false === $firewall['stateless']) {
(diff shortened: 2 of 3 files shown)


src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/RememberMeFactory.php
```diff
@@ -111,15 +111,6 @@ public function createAuthenticator(ContainerBuilder $container, string $firewal
             ->replaceArgument(3, $config['name'] ?? $this->options['name'])
         ;
 
-        foreach ($container->findTaggedServiceIds('security.remember_me_aware') as $serviceId => $attributes) {
-            // register ContextListener
-            if (str_starts_with($serviceId, 'security.context_listener')) {
-                continue;
-            }
-
-            throw new \LogicException(sprintf('Symfony Authenticator Security dropped support for the "security.remember_me_aware" tag, service "%s" will no longer work as expected.', $serviceId));
```
chalasr, line 120: @wouterj Do you have any idea why we didn't just remove that code in 6.0? It feels like it was on purpose

chalasr (approved):

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/RememberMeFactory.php
wouterj: I think we've had a badly resolved merge conflict in this part. But this tag appears to be so underused, that we never received an issue about it :)

This code contained the correct logic in 5.1 (when the security rework was done) and 5.2: https://github.com/symfony/symfony/blob/5.1/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/RememberMeFactory.php#L114-L126

The legacy logic in 5.3 and 5.4: https://github.com/symfony/symfony/blob/5.3/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/RememberMeFactory.php#L64-L84

And finally, the current state without any real logic in 6.0, where it should just be completely removed (As it doesn't do anything anymore).

src/Symfony/Bundle/SecurityBundle/DependencyInjection/SecurityExtension.php
```diff
@@ -647,10 +647,6 @@ private function getUserProvider(ContainerBuilder $container, string $id, array
             return $providerIds[$normalizedName];
         }
 
-        if ('remember_me' === $factoryKey && $contextListenerId) {
```
maxbeckers, line 650: I guess we then also can remove the `$contextListenerId` from the method parameters

alamirault: You're right. It's updated !

nicolas-grekas:
Thank you @alamirault.

PR #50695, opened 2023-06-18
[Security] Improve error handling in OIDC access token handlers

chalasr:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | not needed

This PR fine-tunes the way errors are handled in OIDC token handlers by catching more specific exceptions from the underlying JWT library and moving error-free assignments out of try/catch blocks, which makes 2 OIDC-specific exception classes (experimental) useless.


src/Symfony/Component/Security/Http/AccessToken/Oidc/OidcTokenHandler.php
```diff
@@ -55,52 +51,49 @@ public function getUserBadgeFrom(string $accessToken): UserBadge
...
             $jws = $serializerManager->unserialize($accessToken);
-            $claims = json_decode($jws->getPayload(), true);
+            $claims = json_decode($jws->getPayload(), true, 512, JSON_THROW_ON_ERROR);
+        } catch (\InvalidArgumentException|\JsonException $e) {
+            throw new BadCredentialsException('Unable to parse the token.', 0, $e);
+        }
 
-            // Verify the signature
-            if (!$jwsVerifier->verifyWithKey($jws, $this->jwk, 0)) {
-                throw new InvalidSignatureException();
```
chalasr, line 66: throwing specific exceptions to catch them immediately is pointless, the BadCredentialsException's message (not obfuscated messageKey) is enough. If we want to provide extension points to allow e.g. fixing the token on the fly, we can do it by dispatching specific events.

src/Symfony/Component/Security/Http/AccessToken/Oidc/OidcTokenHandler.php
```diff
@@ -55,52 +51,49 @@ public function getUserBadgeFrom(string $accessToken): UserBadge
...
+            new Checker\NotBeforeChecker(0, false, $this->clock),
+            new Checker\ExpirationTimeChecker(0, false, $this->clock),
+            new Checker\AudienceChecker($this->audience),
+            new Checker\IssuerChecker($this->issuers),
+        ]);
 
-            // UserLoader argument can be overridden by a UserProvider on AccessTokenAuthenticator::authenticate
-            return new UserBadge($claims[$this->claim], fn () => $this->createUser($claims), $claims);
-        } catch (\Exception $e) {
-            $this->logger?->error('An error while decoding and validating the token.', [
```
chalasr: missing word -> #50696

src/Symfony/Component/Security/Http/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+ * [BC BREAK] Removed `InvalidSignatureException` and `MissingClaimException` classes from the `AccessToken\Oidc\Exception` namespace
```
maxbeckers: I like this PR to cleanup a bit. But having a BC break in 6.4 sounds to be not our plan of having BC breaks only in Major versions. And then I tought about the change ... why is it a BC break? Two `@experimental` Classes are deleted or the Output changes? I guess removing experimental classes in not a bc break by our definition (and as you already said, the Exceptions have been thrown and directly catched, so they never ver available out of the changed code).

chalasr: Changing the public API in a backward incompatible way is a BC break. The fact it is part of an experimental feature only means BC breaks are tolerated as long as we are confident it is improving the feature. See https://symfony.com/doc/current/contributing/code/experimental.html :)

maxbeckers: Thanks for clearification ;)

maxbeckers (approved):

maxbeckers:
@chalasr did you see the test failure? might be related to the change:
>1) Symfony\Bundle\SecurityBundle\Tests\Functional\AccessTokenTest::testOidcSuccess
Symfony\Component\DependencyInjection\Exception\InvalidParameterTypeException: Invalid definition for service "security.access_token_handler.main": argument 6 of "Symfony\Component\Security\Http\AccessToken\Oidc\OidcTokenHandler::__construct()" accepts "Psr\Clock\ClockInterface", "Psr\Log\NullLogger" passed.

src/Symfony/Component/Security/Http/CHANGELOG.md
nicolas-grekas:
```suggestion
 * [BC BREAK] Remove experimental `InvalidSignatureException` and `MissingClaimException` classes from the `AccessToken\Oidc\Exception` namespace
```

src/Symfony/Component/Security/Http/AccessToken/Oidc/OidcTokenHandler.php
```diff
@@ -44,7 +41,6 @@ public function __construct(
         private string $audience,
         private array $issuers,
         private string $claim = 'sub',
-        private ?LoggerInterface $logger = null,
         private ClockInterface $clock = new Clock()
```
nicolas-grekas: Let's add a small BC layer because this will break the bundle v6.3 otherwise:
```suggestion
        private ClockInterface|LoggerInterface|null $clock = new Clock()
```
then trigger a deprecation 
and add a line in the changelog

chalasr: done

chalasr:
Comments addressed. The failing test should turn back to green once merged up to 7.0.

nicolas-grekas:
Can you please rebase? Can you also pleas update the description of the PR so that we can get what this is about? :)

chalasr:
@nicolas-grekas done

src/Symfony/Component/Security/Http/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ CHANGELOG
 
  * `UserValueResolver` no longer implements `ArgumentValueResolverInterface`
  * Deprecate calling the constructor of `DefaultLoginRateLimiter` with an empty secret
+ * [BC BREAK] Remove experimental `InvalidSignatureException` and `MissingClaimException` classes from the `AccessToken\Oidc\Exception` namespace
```
stof, line 9: This BC break is not worth it IMO. you could deprecate this 2 exceptions instead and remove them in 7.0

src/Symfony/Component/Security/Http/AccessToken/Oidc/OidcTokenHandler.php
```diff
@@ -43,9 +41,12 @@ public function __construct(
         private string $audience,
         private array $issuers,
         private string $claim = 'sub',
-        private ?LoggerInterface $logger = null,
-        private ClockInterface $clock = new Clock()
+        private ClockInterface|LoggerInterface|null $clock = new Clock()
     ) {
+        if (!$clock || $clock instanceof LoggerInterface) {
+            $this->clock = new Clock();
```
stof, line 47: You also need to account for the case where the clock is pass explicitly in the old position instead of ignoring it and assuming that the default clock is used.

stof (changes requested):

chalasr:
That logic is getting updated in some more recent PRs. I may resume it later if needed.

PR #50705, opened 2023-06-19
[Mailer][Webhook] Add Sendgrid webhook support

### LINKED ISSUE #50704: [Webhook][RemoteEvent] Add Sendgrid
### Description

Allow webhooks from [SendGrid](https://docs.sendgrid.com/for-developers/tracking-events/getting-started-event-webhook)

### Example

_No response_

WoutervanderLoopNL:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Tickets       | Fix #50704
| License       | MIT

Support webhooks from [SendGrid](https://docs.sendgrid.com/for-developers/tracking-events/getting-started-event-webhook)

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
@@ -2615,6 +2615,7 @@ class FrameworkExtension extends Extension
             $webhookRequestParsers = [
                 MailerBridge\Mailgun\Webhook\MailgunRequestParser::class => 'mailer.webhook.request_parser.mailgun',
                 MailerBridge\Postmark\Webhook\PostmarkRequestParser::class => 'mailer.webhook.request_parser.postmark',
+                MailerBridge\Sendgrid\Webhook\SendgridRequestParser::class => 'mailer.webhook.request_parser.sendgrid',
             ];
 
             foreach ($webhookRequestParsers as $class => $service) {
--- a/src/Symfony/Component/Mailer/Bridge/Sendgrid/README.md
+++ b/src/Symfony/Component/Mailer/Bridge/Sendgrid/README.md
@@ -16,6 +16,30 @@ MAILER_DSN=sendgrid+api://KEY@default
 where:
  - `KEY` is your Sendgrid API Key
 
+
+Webhook:
+--------
+Create route:
+```yaml
+framework:
+    webhook:
+        routing:
+            sendgrid:
+                service: mailer.webhook.request_parser.sendgrid
+                secret: '!SENDGRID_VALIDATION_SECRET!' #Leave blank if you dont want to use the signature validation
+```
+Create consumer:
+```php
+#[\Symfony\Component\RemoteEvent\Attribute\AsRemoteEventConsumer(name: 'sendgrid')]
+class SendGridConsumer implements ConsumerInterface
+{
+    public function consume(RemoteEvent|MailerDeliveryEvent $event): void
+    {
+        //your code
+    }
+}
+```
+
 Resources
 ---------
 
--- a/src/Symfony/Component/Mailer/Bridge/Sendgrid/Webhook/SendgridRequestParser.php
+++ b/src/Symfony/Component/Mailer/Bridge/Sendgrid/Webhook/SendgridRequestParser.php
@@ -1,0 +1,108 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Mailer\Bridge\Sendgrid\Webhook;
+
+use Symfony\Component\HttpFoundation\ChainRequestMatcher;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestMatcher\IsJsonRequestMatcher;
+use Symfony\Component\HttpFoundation\RequestMatcher\MethodRequestMatcher;
+use Symfony\Component\HttpFoundation\RequestMatcherInterface;
+use Symfony\Component\Mailer\Bridge\Sendgrid\RemoteEvent\SendgridPayloadConverter;
+use Symfony\Component\RemoteEvent\Event\Mailer\AbstractMailerEvent;
+use Symfony\Component\RemoteEvent\Exception\ParseException;
+use Symfony\Component\Webhook\Client\AbstractRequestParser;
+use Symfony\Component\Webhook\Exception\RejectWebhookException;
+
+/**
+ * @author WoutervanderLoop.nl <info@woutervanderloop.nl>
+ */
+final class SendgridRequestParser extends AbstractRequestParser
+{
+    public function __construct(
+        private readonly SendgridPayloadConverter $converter,
+    ) {
+    }
+
+    protected function getRequestMatcher(): RequestMatcherInterface
+    {
+        return new ChainRequestMatcher([
+            new MethodRequestMatcher('POST'),
+            new IsJsonRequestMatcher(),
+        ]);
+    }
+
+    protected function doParse(Request $request, string $secret): ?AbstractMailerEvent
+    {
+        $content = $request->toArray();
+        if (
+            !isset($content[0]['email'])
+            || !isset($content[0]['timestamp'])
+            || !isset($content[0]['event'])
+            || !isset($content[0]['sg_message_id'])
+        ) {
+            throw new RejectWebhookException(406, 'Payload is malformed.');
+        }
+
+        if ($secret) {
+            if (!$request->headers->get('X-Twilio-Email-Event-Webhook-Signature')
+                || !$request->headers->get('X-Twilio-Email-Event-Webhook-Timestamp')
+            ) {
+                throw new RejectWebhookException(406, 'Signature is required.');
+            }
+
+            $this->validateSignature(
+                $request->headers->get('X-Twilio-Email-Event-Webhook-Signature'),
+                $request->headers->get('X-Twilio-Email-Event-Webhook-Timestamp'),
+                $request->getContent(),
+                $secret,
+            );
+        }
+
+        try {
+            return $this->converter->convert($content[0]);
+        } catch (ParseException $e) {
+            throw new RejectWebhookException(406, $e->getMessage(), $e);
+        }
+    }
+
+    /**
+     * Verify signed event webhook requests.
+     *
+     * @param string $signature value obtained from the
+     *                          'X-Twilio-Email-Event-Webhook-Signature' header
+     * @param string $timestamp value obtained from the
+     *                          'X-Twilio-Email-Event-Webhook-Timestamp' header
+     * @param string $payload   event payload in the request body
+     * @param string $secret    base64-encoded DER public key
+     *
+     * @see https://docs.sendgrid.com/for-developers/tracking-events/getting-started-event-webhook-security-features
+     */
+    private function validateSignature(
+        string $signature,
+        string $timestamp,
+        string $payload,
+        string $secret,
+    ): void {
+        $timestampedPayload = $timestamp.$payload;
+
+        // Sendgrid provides the verification key as base64-encoded DER data. Openssl wants a PEM format, which is a multiline version of the base64 data.
+        $pemKey = "-----BEGIN PUBLIC KEY-----\n".chunk_split($secret, 64, "\n")."-----END PUBLIC KEY-----\n";
+
+        if (!$publicKey = openssl_pkey_get_public($pemKey)) {
+            throw new RejectWebhookException(406, 'Public key is wrong.');
+        }
+
+        if (1 !== openssl_verify($timestampedPayload, base64_decode($signature), $publicKey, \OPENSSL_ALGO_SHA256)) {
+            throw new RejectWebhookException(406, 'Signature is wrong.');
+        }
+    }
+}
--- a/src/Symfony/Component/Mailer/Bridge/Sendgrid/composer.json
+++ b/src/Symfony/Component/Mailer/Bridge/Sendgrid/composer.json
@@ -20,10 +20,12 @@
         "symfony/mailer": "^5.4.21|^6.2.7|^7.0"
     },
     "require-dev": {
-        "symfony/http-client": "^5.4|^6.0|^7.0"
+        "symfony/http-client": "^5.4|^6.0|^7.0",
+        "symfony/webhook": "^6.3|^7.0"
     },
     "conflict": {
-        "symfony/mime": "<6.2"
+        "symfony/mime": "<6.2",
+        "symfony/http-foundation": "<6.2"
     },
     "autoload": {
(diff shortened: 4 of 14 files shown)


src/Symfony/Component/Mailer/Bridge/Sendgrid/Webhook/SendgridRequestParser.php
```diff
@@ -0,0 +1,102 @@
...
+     *
+     * @see https://docs.sendgrid.com/for-developers/tracking-events/getting-started-event-webhook-security-features
+     */
+    private function validateSignature(
+        string $signature,
+        string $timestamp,
+        string $payload,
+        PublicKey $publicKey,
+    ): void {
+        // see
```
OskarStark: ?

src/Symfony/Component/Mailer/Bridge/Sendgrid/composer.json
```diff
@@ -17,13 +17,17 @@
     ],
     "require": {
         "php": ">=8.1",
+        "psr/event-dispatcher": "^1",
+        "starkbank/ecdsa": "^0.*",
```
OskarStark: A 0.* version is not stable from a semver POV

fabpot: I see a version 2.x on Packagist, maybe we can use it?
Also, it should not be part of the require section as this is optional, so you can move it to request-dev and throw an exception if not available in the code.

src/Symfony/Component/Mailer/Bridge/Sendgrid/Webhook/SendgridRequestParser.php
```diff
@@ -0,0 +1,107 @@
...
+            || !isset($content[0]['timestamp'])
+            || !isset($content[0]['event'])
+            || !isset($content[0]['sg_message_id'])
+        ) {
+            throw new RejectWebhookException(406, 'Payload is malformed.');
+        }
+
+        if ($request->headers->get('X-Twilio-Email-Event-Webhook-Signature')
+            && $request->headers->get('X-Twilio-Email-Event-Webhook-Timestamp')
+        ) {
```
SherinBloemendaal: Shouldn't we add a configurable option to enforce signature expectation? The current implementation exposes a security vulnerability because it allows the client to make a malformed request without the headers, causing the message to be processed even though a signature was expected, provided the signed event functionality was also enabled in SendGrid.

WoutervanderLoopNL: Agreed, will work on that

WoutervanderLoopNL: @SherinBloemendaal  Done

SherinBloemendaal (approved):

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -2139,6 +2139,17 @@ private function addMailerSection(ArrayNodeDefinition $rootNode, callable $enabl
...
                         ->end()
+                        ->arrayNode('webhook')
+                            ->info('Mailer Webhook configuration')
+                            ->normalizeKeys(false)
+                            ->useAttributeAsKey('provider')
+                            ->prototype('array')
+                                ->normalizeKeys(false)
+                                ->useAttributeAsKey('name')
+                                ->requiresAtLeastOneElement()
+                                ->prototype('scalar')->end()
```
stof: as the DI extension expects dedicated keys, this should not be a scalar prototype but a structured array prototype instead, making it a lot easier to preserve BC on the configuration structure.

src/Symfony/Component/Mailer/Bridge/Sendgrid/README.md
```diff
@@ -16,6 +16,39 @@ MAILER_DSN=sendgrid+api://KEY@default
...
+}
+```
+
+Default sendgrid enforces to validate the signature, to disable the signature validation:
+```yaml
+framework:
+    mailer:
+        webhook:
+            sendgrid:
+                validate_signature: false
```
stof: Why configuring this as part of `mailer` in a new `webhook` configuration and not in the webhook configuration ?

src/Symfony/Component/Mailer/Bridge/Sendgrid/composer.json
```diff
@@ -17,13 +17,17 @@
     ],
     "require": {
         "php": ">=8.1",
+        "psr/event-dispatcher": "^1",
```
stof: why is this adding a new dependency ?

src/Symfony/Component/Mailer/Bridge/Sendgrid/Webhook/SendgridRequestParser.php
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        if ($this->validateSignature) {
+            if (!$request->headers->get('X-Twilio-Email-Event-Webhook-Signature')
+                || !$request->headers->get('X-Twilio-Email-Event-Webhook-Timestamp')
+            ) {
+                throw new RejectWebhookException(406, 'Signature is required.');
+            }
+
+            if (!class_exists(Ecdsa::class)) {
```
stof: this library also requires `ext-gmp` to work, which is neither a composer requirement nor a runtime check for now.

src/Symfony/Component/Mailer/Bridge/Sendgrid/Webhook/SendgridRequestParser.php
```diff
@@ -0,0 +1,111 @@
...
+                throw new \LogicException('Package "starkbank/ecdsa" is required to use the "event-webhook-security" feature. Try running "composer require starkbank/ecdsa".');
```
stof: Why using this external library instead of relying on `openssl_verify` to do the job ?

src/Symfony/Component/Mailer/Bridge/Sendgrid/README.md
WoutervanderLoopNL: because the component is acually a mailer component. who is able to handle the webhooks. and the mailer bridge has the logic of whether or not verify the signature

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
WoutervanderLoopNL: agreed, but I wanted to make the configuration not directly depended on the sendgrid bridge. so I didn't exactly know how to do it the right way, this was just a way so that the sendgrid bridge (and more in the feature) have the ability to define a config value. 

any correct suggestions?

src/Symfony/Component/Mailer/Bridge/Sendgrid/composer.json
WoutervanderLoopNL: whoops, i did that wrong

src/Symfony/Component/Mailer/Bridge/Sendgrid/Webhook/SendgridRequestParser.php
WoutervanderLoopNL: based on how sendgrid validates there requests in there own lib: https://github.com/sendgrid/sendgrid-php/blob/main/lib/eventwebhook/EventWebhook.php

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
stof: If the processing of the configuration in the extension is specific, making the Configuration class generic is actually worse as it means that both classes have a different understanding of the structure of the configuration.

src/Symfony/Component/Mailer/Bridge/Sendgrid/Webhook/SendgridRequestParser.php
stof: The sendgrid SDK is free to choose their dependencies as they want. But in Symfony, we are careful about them (and especially when it comes to a library that re-implements a crypto algorithm on its own with very few tests)

smnandre:
I closed my https://github.com/symfony/symfony/pull/51305#event-10126833608 and i present you again my sincere apologies. 

Feel free to take anything you want from my PR (maybe the tests + fixtures ?)

src/Symfony/Component/Mailer/Bridge/Sendgrid/Webhook/SendgridRequestParser.php
WoutervanderLoopNL: @stof  I really cant find a way to make it work with default PHP openssl_verify. 

I'm not that experianced with crypto algorithms, but my guess is, that if a big platform a sendgrid choose a algorithm and library as this, it should be secure? :shrug: 

they use it in all there SDKs, (https://docs.sendgrid.com/for-developers/tracking-events/getting-started-event-webhook-security-features) 

Even [ecdsa-php](https://github.com/starkbank/ecdsa-php) has more than [11 million](https://packagist.org/packages/starkbank/ecdsa) installs. 



> (and especially when it comes to a library that re-implements a crypto algorithm on its own with very few tests).

but i completly understand your reason from your point of view. but I also think at the end of the story, the enduser/developer is responsible for making the descission wheter or not implement this feature, at the end, the install of the ecdsa lib is still optional

src/Symfony/Component/Mailer/Bridge/Sendgrid/Webhook/SendgridRequestParser.php
WoutervanderLoopNL: it is a composer requirement of the [starkbank/ecdsa-php](https://github.com/starkbank/ecdsa-php/blob/master/composer.json) library

src/Symfony/Component/Mailer/Bridge/Sendgrid/Webhook/SendgridRequestParser.php
stof: I'm not saying that the ECDSA algorithm is bad (it is not). I'm saying that the ecdsa-php library implementing that algorithm in pure PHP is not properly tested (its tests are only covering 1 input for the crypto algorithm).
I trust a lot more the ECDSA implementation of openssl than of that library.

Having 11 million installs is probably due only to the sendgrid SDK which depends on it (this SDK has 24 million installs).
So this only says that the sendgrid team decided to use that implementation.

stof: The openssl implementation would look like that (my code snippet misses the error handling for openssl failures):

```php
// Sendgrid provides the verification key as base64-encoded DER data. Openssl wants a PEM format, which is a multiline version of the base64 data.
$pemKey = "-----BEGIN PUBLIC KEY-----\n".chunk_split($secret, 64, "\n")."-----END PUBLIC KEY-----\n";

$publicKey = openssl_get_publickey($pemKey);

return openssl_verify($timestampedPayload, base64_decode($signature), $publicKey, OPENSSL_ALGO_SHA256);
```

composer.json
```diff
@@ -36,6 +36,7 @@
         "php": ">=8.1",
         "composer-runtime-api": ">=2.1",
         "ext-xml": "*",
+        "ext-gmp": "*",
```
stof: this should not become a requirement of the whole framework

src/Symfony/Component/Mailer/Bridge/Sendgrid/Webhook/SendgridRequestParser.php
WoutervanderLoopNL: Done

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2624,6 +2625,10 @@ private function registerMailerConfiguration(array $config, ContainerBuilder $co
                     $container->removeDefinition($service);
                 }
             }
+
+            if (ContainerBuilder::willBeAvailable('symfony/sendgrid-mailer', MailerBridge\Sendgrid\Webhook\SendgridRequestParser::class, ['symfony/framework-bundle', 'symfony/mailer'])) {
+                $container->setParameter('mailer.webhook.sendgrid.validate_signature', true);
```
stof: this parameter should rather be defined in the config file

src/Symfony/Component/Mailer/Bridge/Sendgrid/Webhook/SendgridRequestParser.php
```diff
@@ -0,0 +1,107 @@
...
+        string $timestamp,
+        string $payload,
+        string $secret,
+    ): void {
+        $timestampedPayload = $timestamp.$payload;
+
+        // Sendgrid provides the verification key as base64-encoded DER data. Openssl wants a PEM format, which is a multiline version of the base64 data.
+        $pemKey = "-----BEGIN PUBLIC KEY-----\n".chunk_split($secret, 64, "\n")."-----END PUBLIC KEY-----\n";
+
+        $publicKey = openssl_get_publickey($pemKey);
```
stof: you should handle the case where this function returns `false` because the provided key is invalid

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
WoutervanderLoopNL: where do you mean? i tried to do this trick with `ContainerBuilder::willBeAvailable` in the `Configuration.php` as a config value, but that didnt work

WoutervanderLoopNL: @stof  check again

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2624,6 +2625,10 @@ private function registerMailerConfiguration(array $config, ContainerBuilder $co
                     $container->removeDefinition($service);
                 }
             }
+
+            if (ContainerBuilder::willBeAvailable('symfony/sendgrid-mailer', MailerBridge\Sendgrid\Webhook\SendgridRequestParser::class, ['symfony/framework-bundle', 'symfony/mailer'])) {
+                $container->setParameter('mailer.webhook.sendgrid.validate_signature', $config['webhook']['sendgrid']['validate_signature']);
```
stof: As you handle that through the semantic configuration now, I would rather configure the service argument directly instead of defining a parameter

src/Symfony/Component/Mailer/Bridge/Sendgrid/README.md
stof: I'm wondering whether we actually need this boolean parameter at all. Couldn't we automatically enable the signature validation when a non-empty secret is configured for the webhook ?

WoutervanderLoopNL: good point, will look into it

fabpot (approved):

fabpot:
Thank you @WoutervanderLoopNL.

PR #50706, opened 2023-06-19
Update return-type in Command.php

sebastianstucke87:
Calling `Command::getHelper()` will either return an instance of `HelperInterface` or throw an `Exception`.

| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

I was wondering by my IDE was acting up:

Which was fixed by this docblock:

```php
 /** @var HelperInterface $helper */
$helper = $this->getHelper('question');
```

Looking further, I saw that `Command::getHelper()` will either return an instance of `\Symfony\Component\Console\Helper\HelperInterface` or throw an `\Symfony\Component\Console\Exception\InvalidArgumentException` when a question-helper was not found, thus making `mixed` obsolete.

### DIFF
--- a/src/Symfony/Component/Console/Command/Command.php
+++ b/src/Symfony/Component/Console/Command/Command.php
@@ -700,7 +700,7 @@ class Command
      * @throws LogicException           if no HelperSet is defined
      * @throws InvalidArgumentException if the helper is not defined
      */
-    public function getHelper(string $name): mixed
+    public function getHelper(string $name): HelperInterface
     {
         if (null === $this->helperSet) {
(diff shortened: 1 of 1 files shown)


ghost:
The question is why there was a `mixed` TH when DocBlock says `HelperInterface`.

stof:
Changing a return type to a more restrictive one is a BC break for child classes. So this change can only be done in 7.0 (which is why we added a phpdoc to document the more precise type until we're able to update it)

sebastianstucke87:
oh sorry, I totally forgot about inheritance..

PR #50709, opened 2023-06-19
[DoctrineBridge] Improve subscriber deprecation message

### LINKED ISSUE #50701: [DoctrineBridge] Improve deprecation message for Doctrine subscribers
### Description

Is it possible to add more context to this deprecation message? It is hard to find them, at which location this is used.
Context and Trace is not helpful in the profiler.
Context like Class, method, etc. would be useful if this info is available at this point.

`Since symfony/doctrine-bridge 6.3: Using Doctrine subscribers as services is deprecated, declare listeners instead`

### Example

_No response_

alamirault:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | yes/no 
| Tickets       | Fix #50701
| License       | MIT

Imrprove deprecation message in order to know which subscriber should be updated

### DIFF
--- a/src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
+++ b/src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
@@ -189,7 +189,7 @@ class ContainerAwareEventManager extends EventManager
                 $listener = $this->container->get($listener);
             }
             // throw new \InvalidArgumentException(sprintf('Using Doctrine subscriber "%s" is not allowed, declare it as a listener instead.', \is_object($listener) ? $listener::class : $listener));
-            trigger_deprecation('symfony/doctrine-bridge', '6.3', 'Using Doctrine subscribers as services is deprecated, declare listeners instead');
+            trigger_deprecation('symfony/doctrine-bridge', '6.3', 'Registering "%s" as a Doctrine subscriber is deprecated. Register it as a listener instead, using e.g. the #[AsDoctrineListener] attribute.', \is_object($listener) ? get_debug_type($listener) : $listener);
             parent::addEventSubscriber($listener);
         }
     }
--- a/src/Symfony/Bridge/Doctrine/DependencyInjection/CompilerPass/RegisterEventListenersAndSubscribersPass.php
+++ b/src/Symfony/Bridge/Doctrine/DependencyInjection/CompilerPass/RegisterEventListenersAndSubscribersPass.php
@@ -106,7 +106,7 @@ class RegisterEventListenersAndSubscribersPass implements CompilerPassInterface
                     $refs = $managerDef->getArguments()[1] ?? [];
                     $listenerRefs[$con][$id] = new Reference($id);
                     if ($subscriberTag === $tagName) {
-                        trigger_deprecation('symfony/doctrine-bridge', '6.3', 'Using Doctrine subscribers as services is deprecated, declare listeners instead');
+                        trigger_deprecation('symfony/doctrine-bridge', '6.3', 'Registering "%s" as a Doctrine subscriber is deprecated. Register it as a listener instead, using e.g. the #[AsDoctrineListener] attribute.', $id);
                         $refs[] = $id;
                     } else {
                         $refs[] = [[$tag['event']], $id];
--- a/src/Symfony/Bridge/Doctrine/Tests/ContainerAwareEventManagerTest.php
+++ b/src/Symfony/Bridge/Doctrine/Tests/ContainerAwareEventManagerTest.php
@@ -50,7 +50,7 @@ class ContainerAwareEventManagerTest extends TestCase
         $this->container->set('sub1', $subscriber1 = new MySubscriber(['foo']));
         $this->container->set('sub2', $subscriber2 = new MySubscriber(['foo']));
 
-        $this->expectDeprecation('Since symfony/doctrine-bridge 6.3: Using Doctrine subscribers as services is deprecated, declare listeners instead');
+        $this->expectDeprecation('Since symfony/doctrine-bridge 6.3: Registering "Symfony\Bridge\Doctrine\Tests\MySubscriber" as a Doctrine subscriber is deprecated. Register it as a listener instead, using e.g. the #[AsDoctrineListener] attribute.');
         $this->assertSame([$subscriber1,  $subscriber2], array_values($this->evm->getListeners('foo')));
     }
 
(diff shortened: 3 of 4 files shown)


src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
```diff
@@ -189,7 +189,7 @@ private function initializeSubscribers(): void
                 $listener = $this->container->get($listener);
             }
             // throw new \InvalidArgumentException(sprintf('Using Doctrine subscriber "%s" is not allowed, declare it as a listener instead.', \is_object($listener) ? $listener::class : $listener));
-            trigger_deprecation('symfony/doctrine-bridge', '6.3', 'Using Doctrine subscribers as services is deprecated, declare listeners instead');
+            trigger_deprecation('symfony/doctrine-bridge', '6.3', 'Using Doctrine subscriber "%s" as service is deprecated, declare it as a listener instead', \is_object($listener) ? $listener::class : $listener);
```
chalasr: As pretty much everything is a service in Symfony, the message still feels confusing. 
What do you think about:
```suggestion
            trigger_deprecation('symfony/doctrine-bridge', '6.3', 'Implementing "%s" in "%s" is deprecated, declare it as a listener instead.', EventSubscriber::class, \is_object($listener) ? $listener::class : $listener);
```

chalasr (commented):
Thanks for caring.

src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
alamirault: You're right, PR is now updated

nicolas-grekas: get_debug_type() instead of ::class

xabbuh:
There is also a compiler pass that we should update as well.

alamirault:
@xabbuh do you think this deprecation is better and valid ? I don't know if use `$id` is correct

```php
trigger_deprecation('symfony/doctrine-bridge', '6.3', 'Implementing "%s" in "%s" is deprecated, declare it as a listener instead.', EventSubscriber::class, $id);                        
```

https://github.com/symfony/symfony/blob/c08780e0242b9c3f1768d5212a63e1ff831abcb3/src/Symfony/Bridge/Doctrine/DependencyInjection/CompilerPass/RegisterEventListenersAndSubscribersPass.php#L109

xabbuh:
@alamirault looks good to me

src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
```diff
@@ -189,7 +189,7 @@ private function initializeSubscribers(): void
                 $listener = $this->container->get($listener);
             }
             // throw new \InvalidArgumentException(sprintf('Using Doctrine subscriber "%s" is not allowed, declare it as a listener instead.', \is_object($listener) ? $listener::class : $listener));
-            trigger_deprecation('symfony/doctrine-bridge', '6.3', 'Using Doctrine subscribers as services is deprecated, declare listeners instead');
+            trigger_deprecation('symfony/doctrine-bridge', '6.3', 'Implementing "%s" in "%s" is deprecated, declare it as a listener instead.', EventSubscriber::class, \is_object($listener) ? get_debug_type($listener) : $listener);
```
nicolas-grekas: we didn't deprecate implementing the interface, but we deprecated relying on it to register listeners
```suggestion
            trigger_deprecation('symfony/doctrine-bridge', '6.3', 'Registering "%s" as a Doctrine subscriber is deprecated. Register it as a listener with the #[AsDoctrineListener] attribute instead.', EventSubscriber::class, \is_object($listener) ? get_debug_type($listener) : $listener);
```

Chris53897:
Thanks for working on this

src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
alamirault: Applied: I removed `EventSubscriber::class` as there is only one placholder in your suggestion

nicolas-grekas (approved):

stof:
I'm not sure we should mention the attribute in this deprecation message. For services not relying on autoconfiguration, the attribute is not the way to migrate the configuration.

And for services relying on autoconfiguration to register it as a subscriber, it means that they implement the EventSubscriberInterface of DoctrineBundle (the base EventSubscriber of doctrine/event-manager is not autoconfigured as we would not know in which event manager the listener should be registered when you have multiple doctrine packages in your project), which is already deprecated in DoctrineBundle 2.10 in factor of the the attribute.

Thus, depending on the Doctrine project you are using, the DoctrineBundle would not be relevant for the migration.

nicolas-grekas:
For apps the attribute is the best suggestion we can make, isn't it? For bundles I agree but those need more care.
Any proposal?

stof:
Even for apps, the attribute to use will be different depending on whether they use the ORM or the MongoDB ODM or another Doctrine project. And some apps might have decided to not use autoconfiguration but manual configuration.

Well, I think this deprecation should keep saying `register them as listener` for the ContainerAwareEventManager which has no way to know for which project it is used. for the compiler pass, it could mention the tag name to use (as it knows which tag name it processes).
For apps using the deprecated EventSubscriberInterface of DoctrineBundle for autoconfiguration, they will already get a deprecation warning hinting at the attribute (because in that case, it is indeed the proper migration path)

src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
```diff
@@ -189,7 +189,7 @@ private function initializeSubscribers(): void
                 $listener = $this->container->get($listener);
             }
             // throw new \InvalidArgumentException(sprintf('Using Doctrine subscriber "%s" is not allowed, declare it as a listener instead.', \is_object($listener) ? $listener::class : $listener));
-            trigger_deprecation('symfony/doctrine-bridge', '6.3', 'Using Doctrine subscribers as services is deprecated, declare listeners instead');
+            trigger_deprecation('symfony/doctrine-bridge', '6.3', 'Registering "%s" as a Doctrine subscriber is deprecated. Register it as a listener with the #[AsDoctrineListener] attribute instead.', \is_object($listener) ? get_debug_type($listener) : $listener);
```
nicolas-grekas: @stof would that work for you?
```suggestion
            trigger_deprecation('symfony/doctrine-bridge', '6.3', 'Registering "%s" as a Doctrine subscriber is deprecated. Register it as a listener instead, using e.g. the #[AsDoctrineListener] attribute.', \is_object($listener) ? get_debug_type($listener) : $listener);
```

stof: this is a bit better. But it still provides a wrong migration path for anything else than the ORM listeners or for non-autowired projects.

nicolas-grekas: @alamirault can you take this message? I don't have a better idea, and while @stof is right, I think it's fine to optimize for the end users (bundle authors need more skills anyway, and this can server as a hint for them)

alamirault: @nicolas-grekas I updated PR

nicolas-grekas:
Thank you @alamirault.

PR #50710, opened 2023-06-20
[FrameworkBundle] Fix setting decorated services during tests

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

When a service is decorated, setting this service with the test container should set the inner service, not the outer one.

Spotted while trying to set a custom scoped http-client in a test case, which was failing because this replaced the traceable+retryable+etc decoration chain we put on top of custom clients.


stof:
Test failures are related to the PR.

nicolas-grekas:
Not anymore :)

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

PR #50716, opened 2023-06-20
[Messenger Doctrine] Fixed regression by #50524 causing data loss

brusch:
Regression caused by #50524 

When using `$platform->getAlterSchemaSQL($schemaDiff)` instead of `$schemaDiff->toSaveSql($platform)` causes to call `\Doctrine\DBAL\Schema\SchemaDiff::_toSql($platform, false)` instead of `\Doctrine\DBAL\Schema\SchemaDiff::_toSql($platform, true)`.  When $saveMode=false the diff from the schema is getting remove, so actually all other tables ... in the DB are getting deleted

| Q             | A
| ------------- | ---
| Branch?       |6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no >
| License       | MIT


src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
```diff
@@ -480,7 +480,7 @@ private function updateSchema(): void
         $comparator = $this->createComparator($schemaManager);
         $schemaDiff = $this->compareSchemas($comparator, method_exists($schemaManager, 'introspectSchema') ? $schemaManager->introspectSchema() : $schemaManager->createSchema(), $this->getSchema());
         $platform = $this->driverConnection->getDatabasePlatform();
-        $queries = method_exists($platform, 'getAlterSchemaSQL') ? $platform->getAlterSchemaSQL($schemaDiff) : $schemaDiff->toSaveSql($platform);
+        $queries = $schemaDiff->toSaveSql($platform);
```
nicolas-grekas: But toSaveSql is deprecated, isn't it? so we need to find the future-proof version of this.

blankse:
@brusch https://github.com/symfony/symfony/pull/50524 was merged in branch `5.4`. So I think we should fix it also in branch `5.4`, right?

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
brusch: Yeah, that's true, but I haven't found an adequate replacement yet 🥲 Therefore I'd keep the current implementation until there's a solution offered by Doctrine. 
Probably `doctrine/dbal` needs to add a parameter to `getAlterSchemaSQL()` or offers an additional method for that 🤔

stof:
@brusch if you want to change the target branch to 5.4, you need to rebase your branch so that it is based on 5.4 too. If the branch is still based on 6.3, that would merge 6.3 into 5.4 which is a no-go.

brusch:
@stof yep, on it 😉

PR #50717, opened 2023-06-20
[Messenger Doctrine] Fixed regression by #50524 causing data loss

brusch:
Regression caused by #50524 

When using `$platform->getAlterSchemaSQL($schemaDiff)` instead of `$schemaDiff->toSaveSql($platform)` causes to call `\Doctrine\DBAL\Schema\SchemaDiff::_toSql($platform, false)` instead of `\Doctrine\DBAL\Schema\SchemaDiff::_toSql($platform, true)`.  When $saveMode=false the diff from the schema is getting remove, so actually all other tables ... in the DB are getting deleted

| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no >
| License       | MIT

Follow up to #50716 which was closed automatically by mistake 😊


brusch:
Ok, according to https://github.com/doctrine/dbal/pull/5766 there's no replacement for `SchemaDiff::toSaveSql()`.
But still, in Symfony we can't change to `getAlterSchemaSQL()` in a minor version as it would be a breaking change.

nicolas-grekas:
Replaced by #50742, thanks for raising this.

PR #48707, opened 2022-12-19
[DependencyInjection] Target Attribute must fail if the target does not exist

rodmen:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #48555 
| License       | MIT
| Doc PR        | -

This change checks if an attribute is a Target to not auto bind with a new default object if the target name does not exists.
The existent behavior should be in charge of suggest possible options for a typo on the Target name.

I can't mark `allow edit by maintainers` I think because I'm open the PR from a organization repo.

PR #50718, opened 2023-06-20
[DependencyInjection] Improve reporting named autowiring aliases

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

This PR started as a fix of #48707 to improve the error message reported when `#[Target]` is used with an invalid target, and ended up with many other related improvements:

- Allow using `#[Target]` with no arguments. This has the effect of throwing an exception if the name of the argument that has the attribute doesn't match a named autowiring alias.

- Improve the error message when a target doesn't match:

image

- Improve `debug:autowiring` to display the target name to use with `#[Target]`:

image

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Attribute/Target.php
+++ b/src/Symfony/Component/DependencyInjection/Attribute/Target.php
@@ -12,6 +12,7 @@
 namespace Symfony\Component\DependencyInjection\Attribute;
 
 use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
+use Symfony\Component\DependencyInjection\Exception\LogicException;
 
 /**
  * An attribute to tell how a dependency is used and hint named autowiring aliases.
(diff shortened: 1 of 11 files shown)


src/Symfony/Component/DependencyInjection/Attribute/Target.php
```diff
@@ -21,11 +22,18 @@
...
 final class Target
 {
-    public string $name;
+    public function __construct(
+        public ?string $name = null,
+    ) {
+    }
 
-    public function __construct(string $name)
+    public function getParsedName(): string
```
stof, line 30: should we even have it as a public method if calling it directly would be a mistake ?

nicolas-grekas: I think yes. Calling it directly works, but only if you gave a name to the constructor.

lyrixx:
i like the feature a lot. Thanks

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

PR #50720, opened 2023-06-20
[Form] [MoneyType] Add "input" option

### LINKED ISSUE #50707: Add new option to MoneyType
### Description

Currently, the MoneyType in the Form component is forcing the value to a float, using a data transformer.
This cannot be very pleasant when using that type with some external library, like the most famous moneyphp, which requires the value to be an integer.
My proposal is to add a new option (a possible name could be "format"), so the user is not forced to re-cast every value before using it with the Money class. Using "float" as the default of this new option can ensure a full BC.

### Example

_No response_

BASAKSemih:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Tickets       | Fix #50707 
| License       | MIT
| Doc PR        | - WIP

### DIFF
--- a/src/Symfony/Component/Form/Extension/Core/Type/MoneyType.php
+++ b/src/Symfony/Component/Form/Extension/Core/Type/MoneyType.php
@@ -37,7 +37,9 @@ class MoneyType extends AbstractType
                 $options['grouping'],
                 $options['rounding_mode'],
                 $options['divisor'],
-                $options['html5'] ? 'en' : null
+                $options['html5'] ? 'en' : null,
+                $options['format']
+
             ))
         ;
     }
@@ -50,7 +52,7 @@ class MoneyType extends AbstractType
         $view->vars['money_pattern'] = self::getPattern($options['currency']);
 
         if ($options['html5']) {
-            $view->vars['type'] = 'number';
+            $view->vars['type'] = $options['format'] === 'integer' ? 'integer' : 'number';
         }
     }
 
@@ -60,14 +62,15 @@ class MoneyType extends AbstractType
     public function configureOptions(OptionsResolver $resolver)
     {
         $resolver->setDefaults([
-            'scale' => 2,
-            'grouping' => false,
-            'rounding_mode' => \NumberFormatter::ROUND_HALFUP,
-            'divisor' => 1,
-            'currency' => 'EUR',
-            'compound' => false,
-            'html5' => false,
+            'scale'           => 2,
+            'grouping'        => false,
+            'rounding_mode'   => \NumberFormatter::ROUND_HALFUP,
+            'divisor'         => 1,
+            'currency'        => 'EUR',
+            'compound'        => false,
+            'html5'           => false,
             'invalid_message' => 'Please enter a valid money amount.',
+            'format'          => 'float'
         ]);
 
         $resolver->setAllowedValues('rounding_mode', [
@@ -80,6 +83,7 @@ class MoneyType extends AbstractType
             \NumberFormatter::ROUND_CEILING,
         ]);
 
+        $resolver->setAllowedValues('format', ['float', 'integer']);
         $resolver->setAllowedTypes('scale', 'int');
 
         $resolver->setAllowedTypes('html5', 'bool');
--- a/src/Symfony/Component/Form/Extension/Core/Type/MoneyType.php
+++ b/src/Symfony/Component/Form/Extension/Core/Type/MoneyType.php
@@ -38,7 +38,7 @@ class MoneyType extends AbstractType
                 $options['rounding_mode'],
                 $options['divisor'],
                 $options['html5'] ? 'en' : null,
-                $options['format']
+                $options['input']
 
             ))
         ;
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/Form/Extension/Core/Type/MoneyType.php
```diff
@@ -50,7 +52,7 @@ public function buildView(FormView $view, FormInterface $form, array $options)
         $view->vars['money_pattern'] = self::getPattern($options['currency']);
 
         if ($options['html5']) {
-            $view->vars['type'] = 'number';
+            $view->vars['type'] = $options['format'] === 'integer' ? 'integer' : 'number';
```
xabbuh: If the `divisor` option has a value greater than 1, we still need a `number` input here, right?

src/Symfony/Component/Form/Extension/Core/Type/MoneyType.php
```diff
@@ -80,6 +83,7 @@ public function configureOptions(OptionsResolver $resolver)
             \NumberFormatter::ROUND_CEILING,
         ]);
 
+        $resolver->setAllowedValues('format', ['float', 'integer']);
```
xabbuh: In the context of our date types we use the term `input`.

BASAKSemih: Maybe `input_type` ?

src/Symfony/Component/Form/Extension/Core/Type/MoneyType.php
```diff
@@ -60,14 +62,15 @@ public function buildView(FormView $view, FormInterface $form, array $options)
...
     {
         $resolver->setDefaults([
-            'scale' => 2,
-            'grouping' => false,
-            'rounding_mode' => \NumberFormatter::ROUND_HALFUP,
-            'divisor' => 1,
-            'currency' => 'EUR',
-            'compound' => false,
-            'html5' => false,
+            'scale'           => 2,
```
xabbuh, line 65: the codestyle changes have to be reverted

src/Symfony/Component/Form/Tests/Extension/Core/Type/MoneyTypeTest.php
```diff
@@ -123,4 +123,23 @@ public function testHtml5EnablesSpecificFormatting()
...
+    }
+
+    public function testValueToIntegerWithSpecificOptionInputToInteger()
+    {
+        $form = $this->factory->create(static::TESTED_TYPE, null, ['input' => 'integer']);
+        $form->submit('12345.6');
+        $this->assertSame(12345, $form->getData());
+    }
+
+    public function testInputTypeIntegerAndDivisorNotEgalToOne()
```
xabbuh, line 140:
```suggestion
    public function testInputTypeIntegerAndDivisorNotEqualToOne()
```

src/Symfony/Component/Form/Tests/Extension/Core/Type/MoneyTypeTest.php
```diff
@@ -123,4 +123,23 @@ public function testHtml5EnablesSpecificFormatting()
...
+    public function testHtml5EnablesSpecificFormattingWithIntegerFormat()
+    {
+        $form = $this->factory->create(static::TESTED_TYPE, null, ['html5' => true, 'scale' => 2, 'input' => 'integer']);
+        $this->assertSame('integer', $form->createView()->vars['type']);
+    }
+
+    public function testValueToIntegerWithSpecificOptionInputToInteger()
+    {
+        $form = $this->factory->create(static::TESTED_TYPE, null, ['input' => 'integer']);
+        $form->submit('12345.6');
```
xabbuh, line 136: IMO this case should trigger an exception in during `reverseTransform()`

src/Symfony/Component/Form/Tests/Extension/Core/Type/MoneyTypeTest.php
```diff
@@ -123,4 +123,23 @@ public function testHtml5EnablesSpecificFormatting()
         $this->assertSame('12345.60', $form->createView()->vars['value']);
         $this->assertSame('number', $form->createView()->vars['type']);
     }
+
+    public function testHtml5EnablesSpecificFormattingWithIntegerFormat()
+    {
+        $form = $this->factory->create(static::TESTED_TYPE, null, ['html5' => true, 'scale' => 2, 'input' => 'integer']);
```
xabbuh, line 129: I am not sure that a scale option with a value other than `0` plays well with how the `input` option is implemented currently.

But, this makes me wonder if the way this option is currently implemented actually makes sense. So maybe we should indeed allow the `division` to be greater than 1 and take it into account to decide whether we want to render a `number` or an `integer` widget.

javiereguiluz (commented):
I like this feature, but I wonder about the new option name: `input` sounds too generic.

Just wondering: since we have other options called `model_*` (e.g. `model_timezone` in nhttps://symfony.com/doc/current/reference/forms/types/datetime.html#model-timezone). Should we rename this option to follow that pattern? Some ideas:

```
model_formatting
model_format
model_data_type
...
```

fabpot:
@BASAKSemih Do you have time to take comments into account?

BASAKSemih:
> @BASAKSemih Do you have time to take comments into account?

No sorry i don't have time

fabpot:
Thank you for your answer. Let’s close for now then. Maybe someone will take over.

PR #50722, opened 2023-06-20
[Mime][Address] Add Custom Validator Address

### LINKED ISSUE #50592: [Mime] Allow not RFC-compliant addresses
### Description

`Symfony\Component\Mime\Address` only accepts RFC-compliant addresses. Otherwise the constructor throws `RfcComplianceException`.

This works for most users, but I want to change this behavior.
Until several years ago, some cell-phone carriers in Japan used to allow users to create an RFC-incompliant email address, e.g. `foo..bar@example.com` (with two consecutive dots).
I would like to send an email to such addresses with mailer component.

One possible solution is to make `EmailValidation` configurable.
`Address` class uses `Egulias\EmailValidator\Validation\MessageIDValidation` or `Egulias\EmailValidator\Validation\RFCValidation`. These classes implements the rule of validation.
https://github.com/symfony/mime/blob/6.3/Address.php#L53
Introducing `Address::setEmailValidation(EmailValidation $emailValidation)` will allow users to employ their own validation rule.

### Example

```php
<?php
use Symfony\Component\Mime\Address;
use Egulias\EmailValidator\Validation\EmailValidation;

$customValidation = new class implements EmailValidation {
    public function isValid(string $email, EmailLexer $emailLexer): bool
    {
        // implement custom validation rule
    }

    // implement getErrors and getWarnings too
};
Address::setEmailValidation($customValidation);

Address::create('foo..bar@example.com'); // Validated by the custom rule
(new Email())->to('foo..bar@example.com'); // Validated by the custom rule too
```

BASAKSemih:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #50592 
| License       | MIT
| Doc PR        | (Still an idea if that makes sense. The documentation will be updated before merging.)

I think this is the simplest way, not the best. I'm waiting for your opinions.

### DIFF
--- a/src/Symfony/Component/Mime/Address.php
+++ b/src/Symfony/Component/Mime/Address.php
@@ -33,20 +33,18 @@ final class Address
      */
     private const FROM_STRING_PATTERN = '~(?<displayName>[^<]*)<(?<addrSpec>.*)>[^>]*~';
 
-    private static EmailValidator $validator;
+    private static EmailValidator|EmailValidatorInterface $validator;
     private static IdnAddressEncoder $encoder;
 
     private string $address;
     private string $name;
 
-    public function __construct(string $address, string $name = '')
+    public function __construct(string $address, string $name = '', EmailValidatorInterface $customValidator = null)
     {
-        if (!class_exists(EmailValidator::class)) {
-            throw new LogicException(sprintf('The "%s" class cannot be used as it needs "%s". Try running "composer require egulias/email-validator".', __CLASS__, EmailValidator::class));
+        if (($customValidator === null) && !class_exists(EmailValidator::class)) {
+            throw new LogicException(sprintf('The "%s" class cannot be used as it needs "%s". Try running "composer require egulias/email-validator" or create a custom validator implementing "%s"', __CLASS__, EmailValidator::class, EmailValidatorInterface::class));
         }
-
-        self::$validator ??= new EmailValidator();
-
+        self::$validator = ($customValidator) ?: new EmailValidator();
         $this->address = trim($address);
         $this->name = trim(str_replace(["\n", "\r"], '', $name));
 
--- a/src/Symfony/Component/Mime/EmailValidatorInterface.php
+++ b/src/Symfony/Component/Mime/EmailValidatorInterface.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace Symfony\Component\Mime;
+
+interface EmailValidatorInterface
+{
+    public function isValid(string $email, mixed $emailValidation): bool;
+}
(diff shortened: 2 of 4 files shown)


src/Symfony/Component/Mime/Address.php
```diff
@@ -33,20 +33,18 @@ final class Address
...
     {
-        if (!class_exists(EmailValidator::class)) {
-            throw new LogicException(sprintf('The "%s" class cannot be used as it needs "%s". Try running "composer require egulias/email-validator".', __CLASS__, EmailValidator::class));
+        if (($customValidator === null) && !class_exists(EmailValidator::class)) {
+            throw new LogicException(sprintf('The "%s" class cannot be used as it needs "%s". Try running "composer require egulias/email-validator" or create a custom validator implementing "%s"', __CLASS__, EmailValidator::class, EmailValidatorInterface::class));
         }
-
-        self::$validator ??= new EmailValidator();
-
+        self::$validator = ($customValidator) ?: new EmailValidator();
```
maxbeckers, line 47: Maybe better this way?
```suggestion
        self::$validator ??= $customValidator ?? new EmailValidator();
```

src/Symfony/Component/Mime/EmailValidatorInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Symfony\Component\Mime;
+
+interface EmailValidatorInterface
+{
+    public function isValid(string $email, mixed $emailValidation): bool;
```
maxbeckers, line 7: Why is this `mixed` and not `EmailValidation`, MessageIDValidation and RFCValidation are extending from? And do we really need a new Validator or would it make sense to have a way to add an own Validation?

maxbeckers (commented):
I dont really understand the reason to have an own email validator getting MessageIDValidation or RFCValidation vor the `isValid` or we need a way to have a custom validation to inject it into the EmailValidator to solve the problem?

src/Symfony/Component/Mime/Address.php
```diff
@@ -33,20 +33,18 @@ final class Address
...
 
-    private static EmailValidator $validator;
+    private static EmailValidator|EmailValidatorInterface $validator;
     private static IdnAddressEncoder $encoder;
 
     private string $address;
     private string $name;
 
-    public function __construct(string $address, string $name = '')
+    public function __construct(string $address, string $name = '', EmailValidatorInterface $customValidator = null)
```
ro0NL, line 42: should be a https://github.com/egulias/EmailValidator/blob/4.x/src/Validation/EmailValidation.php IMHO

fabpot:
Closing as there is no activity and for the reasons explained in the linked ticket.

PR #50723, opened 2023-06-20
[DependencyInjection] Allow using $schema in yaml config files

### LINKED ISSUE #50721: Ignore $schema property in yaml config files
### Description

I want to use proper JSON schema in my config files to have autocomplete for that I want to use `$schema` property. But that field get's interpreted as Bundle and then throws this missing bundle error:

```
There is no extension able to load the configuration for "$schema"
```

see also https://github.com/yaml/yaml-spec/issues/33

### Example

```yaml
$schema: https://raw.githubusercontent.com/FriendsOfShopware/FroshTools/system-config-over-sf-config/frosh-tools-schema.json
frosh_tools:
```

image

shyim:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #50721
| License       | MIT
| Doc PR        | 

Allows to specify `$schema` in config files to have autocomplete in config files.

Example:

```yaml
$schema: 'https://raw.githubusercontent.com/shopware/platform/trunk/config-schema.json'
shopware:
  <Autocomplete Key>
```

### DIFF
--- a/src/Symfony/Component/DependencyInjection/CHANGELOG.md
+++ b/src/Symfony/Component/DependencyInjection/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * Deprecate `ContainerAwareInterface` and `ContainerAwareTrait`, use dependency injection instead
+ * Added support for `$schema` in Yaml config files
 
 6.3
 ---
--- a/src/Symfony/Component/DependencyInjection/Loader/YamlFileLoader.php
+++ b/src/Symfony/Component/DependencyInjection/Loader/YamlFileLoader.php
@@ -799,7 +799,7 @@ class YamlFileLoader extends FileLoader
         }
 
         foreach ($content as $namespace => $data) {
-            if (\in_array($namespace, ['imports', 'parameters', 'services']) || str_starts_with($namespace, 'when@')) {
+            if (\in_array($namespace, ['imports', 'parameters', 'services', '$schema']) || str_starts_with($namespace, 'when@')) {
                 continue;
             }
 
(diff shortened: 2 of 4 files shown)


nicolas-grekas:
Are there any normative document on the subject?
Also, which tool do support it?
And which major projects do publish this?

shyim:
It's supported in VSCode using the [Yaml extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) and also in Jetbrains products. 
Normally JSON schemas are mapped in the IDE by [schemastore](https://www.schemastore.org/json/), but with that attribute you can override it by file in the IDE. (Schemastore defines a file pattern which JSON schema should be enabled)

Right now I am only aware that [Shopware](https://github.com/SchemaStore/schemastore/commit/854fbdb9aa53b4a9a65329626086a215a261451d) uses it, but I hope other bundles or Symfony will use that too. It really can improve the developer experience, as your IDE says which configurations are existing with documentation, you can deprecate options in proper and so on without looking into the docs.

Example from our schema in PhpStorm: 
image

stof:
To me, this is wrong. Adding a `$schema` key in the data of the file is not the proper way to use a schema for that file (the schema may even define that key as being invalid in this file). In JSON-schema, `$schema` is used for files defining a meta-schema.

Using it for any kind of file is an abuse done by IDE plugins.

shyim:
Yea in some way it's an abuse. I am also open to close it. Maybe Jetbrains adds in future support for:
```
# yaml-language-server: $schema=SCHEMA
```

shyim:
I will annoy Jetbrains 😅  to support that comment. But maybe it makes sense for Symfony to have an JSON schema for their configs to have autocomplete without looking into docs

stof:
Well, given that a config file can contain configuration for any of the bundles, we cannot have a single json schema for them. We would have to generate one dynamically. This is quite complex.

shyim:
For the future world :) https://youtrack.jetbrains.com/issue/IDEA-323117/Support-for-YAML-Schema-using-yaml-language-server-comment

PR #49388, opened 2023-02-15
[CssSelector] Add `:has()` support

franckranaivo:
| Q             | A
| ------------- | ---
| Branch       | 8.1
| Bug fix       | no
| New feature  | yes
| Tickets          | Fix https://github.com/symfony/symfony/issues/40179
| Deprecations | no
| License       | MIT
| Doc PR        |  https://github.com/symfony/symfony-docs/pull/17994

Add support for :has() function.

PR #50726, opened 2023-06-21
[CssSelector] Tests on XPath Translator will always pass

franckranaivo:
While working on another PR (https://github.com/symfony/symfony/pull/49388), I noticed that the assertion below will always pass as it is testing the count of the same array on both side.

| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | 
| License       | MIT
| Doc PR        | 

Instead of
`$this->assertCount(\count($elementsId), $elementsId);` 
It should be 
`$this->assertCount(\count($elementsId), $elements);`

Careful though, This PR will fix the assertion but some tests will not pass after applying this. I am working on it on the other PR above. What should I do ?  Fixing it here or fix this on the other PR?


stof:
to me, this should be fixed in 5.4 to make tests actually test something. and then, numbers should be updated when they are wrong, or the actual bug should be fixed in the XPath translator.

franckranaivo:
Yes, my bad, I forgot to change the base branch to 5.4. Okay it's better to bring necessary modifications here, then.

franckranaivo:
I will close this as my fork isn't based on 5.4 and open another one

PR #50727, opened 2023-06-21
[DependencyInjection] Make better use of memory and CPU during auto-discovery

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

While profiling an app with ~2600 auto-discovered services, I figured out that using `unserialize(serialize())` for deep cloning was a significant memory and CPU hog.

This patch skips deep-cloning when no definitions are nested in the prototype definition.

From 328 MB → 239 MB (-27%) as measured by Blackfire.io
From to 5.0s → 4.2s (-16%) (without Blackfire)

:rocket:


fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

PR #50729, opened 2023-06-21
[CssSelector] Tests on Xpath translator will always pass

franckranaivo:
While working on another PR (https://github.com/symfony/symfony/pull/49388), I noticed that the assertion below will always pass as it is testing the count of the same array on both side.

| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | 
| License       | MIT
| Doc PR        | 

Instead of
`$this->assertCount(\count($elementsId), $elementsId);`
It should be
`$this->assertCount(\count($elementsId), $elements);`

This PR should fix also numbers and/or Xpath translator issues from this fixed assertion. 

P.S.
This is the right PR for this issue (https://github.com/symfony/symfony/pull/50726)

### DIFF
--- a/src/Symfony/Component/CssSelector/Tests/XPath/Fixtures/ids.html
+++ b/src/Symfony/Component/CssSelector/Tests/XPath/Fixtures/ids.html
@@ -45,4 +45,8 @@ c"></li>
 </div>
 <div id="foobar-div" foobar="ab bc
 cde"><span id="foobar-span"></span></div>
-</body></html>
+<section>
+    <span id="no-siblings-of-any-type"></span>
+</section>
+</body>
+</html>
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
And the test now fails :sweat_smile: 
Can you have a look?

franckranaivo:
> And the test now fails 😅 Can you have a look?

Sure, this is why @stof suggested me [here](https://github.com/symfony/symfony/pull/50726#issuecomment-1600916901) to fix also failing tests from this PR.

franckranaivo:
Fixed the failing tests. Xpath translator in itself is not to blame, it is just some wrong assertion or misunderstanding of rules of the selector for some tests.

nicolas-grekas:
Good catch, thanks @franckranaivo.

PR #50734, opened 2023-06-21
[ErrorHandler] Improve fileLinkFormat handling

### LINKED ISSUE #50619: Improve fileLinkFormat / SYMFONY_IDE configuration outside Symfony
### Description

Hello,

I mostly use Symfony components in a standalone way (outside Symfony Framework) and sometimes find quite hard to handle the `fileLinkFormat` globally. 

## error-handler

https://github.com/symfony/symfony/blob/42d43b7ba563b0e9841faaa545289b423917de6d/src/Symfony/Component/ErrorHandler/ErrorRenderer/HtmlErrorRenderer.php#L55-L58

Although it has been easier to setup since `SYMFONY_IDE` env var was introduced, this works perfectly if you don't use a container (Docker, etc.). But the [link format described here](https://symfony.com/doc/current/reference/configuration/framework.html#ide:~:text=When%20running%20your%20app%20in%20a%20container%20or%20in%20a%20virtual%20machine) to handle path mapping won't work with the only `SYMFONY_IDE` env var set. In such case, you will have to set a custom exception handler overriding the default one here:

https://github.com/symfony/symfony/blob/42d43b7ba563b0e9841faaa545289b423917de6d/src/Symfony/Component/ErrorHandler/ErrorHandler.php#L647-L662

Just to provide a default [`new FileLinkFormatter()`](https://github.com/symfony/symfony/blob/42d43b7ba563b0e9841faaa545289b423917de6d/src/Symfony/Component/ErrorHandler/ErrorRenderer/HtmlErrorRenderer.php#L40) object to the `HtmlErrorRenderer` renderer.

```php
$handler->setExceptionHandler(function renderException(\Throwable $exception) use($debug): void
{
    $renderer = \in_array(\PHP_SAPI, ['cli', 'phpdbg'], true) ? new CliErrorRenderer() : new HtmlErrorRenderer(
        $debug, 
        null, 
        new FileLinkFormatter() // <-- HERE
    );

    $exception = $renderer->render($exception);

    if (!headers_sent()) {
        http_response_code($exception->getStatusCode());

        foreach ($exception->getHeaders() as $name => $value) {
            header($name.': '.$value, false);
        }
    }

    echo $exception->getAsString();
});
```

And this will only work for Html errors since the CliErrorRenderer will use the `var-dumper` component.

## var-dumper

In a non Symfony app, the `var-dumper` will almost always fall into the default case here: https://github.com/symfony/symfony/blob/42d43b7ba563b0e9841faaa545289b423917de6d/src/Symfony/Component/VarDumper/VarDumper.php#L91-L92

If you  want your links to be fully functional (e.g. path mapping replacements working in a docker context), you'll have to override the handler:

```php
$cloner = new VarCloner();
$cloner->addCasters(ReflectionCaster::UNSET_CLOSURE_FILE_INFO);
$dumper = in_array(PHP_SAPI, array('cli', 'phpdbg'), true) ? new CliDumper() : new HtmlDumper();
$dumper->setDisplayOptions(['fileLinkFormat' => new FileLinkFormatter()]);

// Configure VarDumper
VarDumper::setHandler(function ($var, ?string $label = null) use ($cloner, $dumper): void {
    $var = $cloner->cloneVar($var);

    if (null !== $label) {
        $var = $var->withContext(['label' => $label]);
    }

    $dumper->dump($var);
});
``` 

It would nice to make fileLinkFormat more easily consistent/configurable when used outside Symfony.

### Example

_No response_

nlemoine:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | Fix #50619
| License       | MIT
| Doc PR        | 

- Avoid repeating file link format guessing (logic is already in FileLinkFormatter class)
- Always set a fileLinkFormat to a FileLinkFormatter object to handle path mappings properly


src/Symfony/Component/ErrorHandler/ErrorRenderer/HtmlErrorRenderer.php
```diff
@@ -52,10 +52,10 @@ public function __construct(bool|callable $debug = false, string $charset = null
...
         $this->debug = \is_bool($debug) ? $debug : $debug(...);
         $this->charset = $charset ?: (\ini_get('default_charset') ?: 'UTF-8');
-        $fileLinkFormat ??= $_ENV['SYMFONY_IDE'] ?? $_SERVER['SYMFONY_IDE'] ?? null;
-        $this->fileLinkFormat = \is_string($fileLinkFormat)
-            ? (ErrorRendererInterface::IDE_LINK_FORMATS[$fileLinkFormat] ?? $fileLinkFormat ?: false)
-            : ($fileLinkFormat ?: \ini_get('xdebug.file_link_format') ?: get_cfg_var('xdebug.file_link_format') ?: false);
+        $this->fileLinkFormat = \is_string($fileLinkFormat) ? new FileLinkFormatter($fileLinkFormat) : ($fileLinkFormat ?: false);
+        if (false === $this->fileLinkFormat) {
+            $this->fileLinkFormat = new FileLinkFormatter();
+        }
```
GromNaN: The type of `$fileLinkFormat` is `string|FileLinkFormatter|null`. And the `FileLinkFormatter` constructor accepts `string|array|null`. The assignment can be one-lined.

```suggestion
        $this->fileLinkFormat = $fileLinkFormat instanceof FileLinkFormatter ? $fileLinkFormat : new FileLinkFormatter($fileLinkFormat);
```

---

But, note the class `FileLinkFormatter` can be missing as the `symfony/error-handler` package doesn't require `symfony/http-kernel`.

GromNaN (changes requested):
Using the `Symfony\Component\HttpKernel\Debug\FileLinkFormatter` in `Symfony\Component\ErrorHandler\ErrorRenderer\HtmlErrorRenderer` is a good idea, but the 2 classes are not in the same composer packages.
I think the `FileLinkFormatter` should be moved to `symfony/error-handler`: this package is already required by `symfony/http-kernel`, but not in the other direction.

This can be done in a backward-compatible way by replacing the class with:
```php
namespace Symfony\Component\HttpKernel\Debug;

use Symfony\Component\ErrorHandler\ErrorRenderer\FileLinkFormatter as NewFileLinkFormatter;

/** @deprecated since 6.4, use Symfony\Component\ErrorHandler\ErrorRenderer\FileLinkFormatter */
class FileLinkFormatter extends NewFileLinkFormatter
{
    // Inherit all the methods from the parent
}
```

I'll just ask @symfony/mergers opinion before you can start editing if you wish.

src/Symfony/Component/ErrorHandler/ErrorRenderer/HtmlErrorRenderer.php
nlemoine: Thanks for you review! I must say these one liners gives me headaches 🙂  Sure they are concise but [takes minutes to process](https://github.com/zakirullin/cognitive-load) for my (modest/slow) developer brain 🤯

nlemoine:
> Using the Symfony\Component\HttpKernel\Debug\FileLinkFormatter in Symfony\Component\ErrorHandler\ErrorRenderer\HtmlErrorRenderer is a good idea, but the 2 classes are not in the same composer packages.

That would be great! My original issue about was actually aiming to consolidate file link formatting and make it more consistent across components. Some business logic, that's already contained in `FileLinkFormatter` class, is repeated in various places:
- https://github.com/symfony/symfony/blob/81c102dd54372d0b353f8a80044c825205a33c1a/src/Symfony/Bridge/Twig/Extension/CodeExtension.php#L31
- https://github.com/symfony/symfony/blob/81c102dd54372d0b353f8a80044c825205a33c1a/src/Symfony/Component/VarDumper/Dumper/CliDumper.php#L84
- https://github.com/symfony/symfony/blob/81c102dd54372d0b353f8a80044c825205a33c1a/src/Symfony/Component/HttpKernel/DataCollector/DumpDataCollector.php#L48
- https://github.com/symfony/symfony/blob/81c102dd54372d0b353f8a80044c825205a33c1a/src/Symfony/Component/VarDumper/Dumper/HtmlDumper.php#L81

Probably all those statements could be replaced with something like (long version):
```php
if ($fileLinkFormat instanceof FileLinkFormatter) {
    $this->fileLinkFormat = $fileLinkFormat;
} else {
    $this->fileLinkFormat = class_exists(FileLinkFormatter::class) ? new FileLinkFormatter() : $fileLinkFormat;
}
```

What do you think @GromNaN?

> I'll just ask @symfony/mergers opinion before you can start editing if you wish.

That would be welcome!

src/Symfony/Component/ErrorHandler/ErrorRenderer/HtmlErrorRenderer.php
GromNaN: Sorry for your brain. That's actually less logic and operations.

The type of `$fileLinkFormat;` can be modified to `private FileLinkFormatter $fileLinkFormat;`

nlemoine: Done!

nicolas-grekas (approved):
I updated all references to the deprecated namespace.
This should be good to do (despite tests)

fabpot (approved):

fabpot:
Thank you @nlemoine.

nlemoine:
Great, thanks!

PR #50742, opened 2023-06-22
[Messenger] Preserve existing Doctrine schema

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no (fixing a not released bug)
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50717
| License       | MIT
| Doc PR        | -

[Inlining](https://github.com/symfony/symfony/pull/50742/files#diff-2a38c9deb498eec50126c297bb3d094c8efb29364934e902d6481761ace6f20a) since Doctrine deprecated "toSaveSql" without providing an alternative.

Cleaning up a few unneeded calls to `ORMSetup::createConfiguration()` also.

### DIFF
--- a/src/Symfony/Component/Cache/Adapter/DoctrineDbalAdapter.php
+++ b/src/Symfony/Component/Cache/Adapter/DoctrineDbalAdapter.php
@@ -22,7 +22,6 @@ use Doctrine\DBAL\ParameterType;
 use Doctrine\DBAL\Schema\DefaultSchemaManagerFactory;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Tools\DsnParser;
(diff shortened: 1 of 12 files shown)


derrabus:
IIRC, the alternative to `toSaveSql` is to introspect only the schema assets that you actually manage before comparing it to the target schema. The rationale is that if you introspect a database with lots of tables/columns/indexes/etc although you're interested in one table only, you keep the database unnecessarily busy.

nicolas-grekas:
Can you help with this? I'm not fluent enough in Doctrine to turn that into code :see_no_evil: 
Or is the attached patch good enough?

derrabus (approved):
I think, the change will work, given that you simply inlined the method. We could iterate on this for 6.4.

PR #50743, opened 2023-06-22
[DependencyInjection] Skip scanning scalar values in compiler passes when not needed

bastnic:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

While profiling an app with ~2600 auto-discovered services, I figured out that we have a way bigger calls to string process that logic would expect.

image

image

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Compiler/AbstractRecursivePass.php
+++ b/src/Symfony/Component/DependencyInjection/Compiler/AbstractRecursivePass.php
@@ -31,6 +31,7 @@ abstract class AbstractRecursivePass implements CompilerPassInterface
      */
     protected $container;
     protected $currentId;
+    protected bool $skipScalars = false;
 
     private bool $processExpressions = false;
     private ExpressionLanguage $expressionLanguage;
@@ -77,6 +78,9 @@ abstract class AbstractRecursivePass implements CompilerPassInterface
     {
         if (\is_array($value)) {
             foreach ($value as $k => $v) {
+                if ((!$v || \is_scalar($v)) && $this->skipScalars) {
+                    continue;
+                }
                 if ($isRoot) {
                     if ($v->hasTag('container.excluded')) {
                         continue;
--- a/src/Symfony/Component/DependencyInjection/Compiler/ResolveParameterPlaceHoldersPass.php
+++ b/src/Symfony/Component/DependencyInjection/Compiler/ResolveParameterPlaceHoldersPass.php
@@ -23,6 +23,8 @@ use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
  */
 class ResolveParameterPlaceHoldersPass extends AbstractRecursivePass
 {
+    protected bool $skipScalars = false;
+
     private ParameterBagInterface $bag;
 
     public function __construct(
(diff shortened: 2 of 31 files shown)


src/Symfony/Component/DependencyInjection/ParameterBag/ParameterBag.php
```diff
@@ -198,7 +198,7 @@ public function resolveValue(mixed $value, array $resolving = []): mixed
             return $args;
         }
 
-        if (!\is_string($value) || 2 > \strlen($value)) {
+        if (!\is_string($value) || 2 > \strlen($value) || !str_contains($value, '%')) {
```
nicolas-grekas: let's optimize for longer strings, which should be the most common:
```suggestion
        if (!\is_string($value) || '' === $value || !str_contains($value, '%')) {
```

src/Symfony/Component/DependencyInjection/Compiler/AbstractRecursivePass.php
```diff
@@ -36,6 +36,11 @@ abstract class AbstractRecursivePass implements CompilerPassInterface
     private ExpressionLanguage $expressionLanguage;
     private bool $inExpression = false;
 
+    public function __construct(
+        private bool $skipScalars = true,
```
nicolas-grekas: this cannot use CPP because child classes are not required to call this constructor (and the property would remain undefined as is)

in order to preserve BC, the property should default to false, and all passes that don't care about scalars should explicitly call this with `true`

src/Symfony/Component/DependencyInjection/Compiler/ResolveParameterPlaceHoldersPass.php
```diff
@@ -29,6 +29,7 @@ public function __construct(
         private bool $resolveArrays = true,
         private bool $throwOnResolveException = true,
     ) {
+        parent::__construct(false);
```
nicolas-grekas: can be removed: it's all the other passes that need to opt-in instead :)

src/Symfony/Component/DependencyInjection/Compiler/AbstractRecursivePass.php
nicolas-grekas: could be a protected property that is redefined in child classes BTW, to save the boilerplate of calling the parent constructor

src/Symfony/Component/DependencyInjection/Compiler/AbstractRecursivePass.php
```diff
@@ -77,6 +78,9 @@ protected function processValue(mixed $value, bool $isRoot = false)
     {
         if (\is_array($value)) {
             foreach ($value as $k => $v) {
+                if (\is_scalar($v ?? true) && $this->skipScalars) {
```
bastnic: the point is that only `ResolveParameterPlaceHoldersPass` should NOT skip that

nicolas-grekas (approved):

src/Symfony/Component/DependencyInjection/Compiler/AbstractRecursivePass.php
nicolas-grekas: And `ResolveEnvPlaceholdersPass` also :)

ghost:
image

image


Are you profiling the same app guys :smiley: ?

bastnic:
@javaDeveloperKid guilty. I used the famous github function "ask a core contrib". Costs me a lot of coffee and beers.

maxbeckers (approved):

nicolas-grekas:
Thank you @bastnic.

PR #50745, opened 2023-06-22
[DependencyInjection] Add `CheckAliasValidityPass` to check interface compatibility

n-valverde:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | /
| License       | MIT
| Doc PR        | TODO

Hello, I would like to propose adding a pass to the compiler, to make sure that when defining an alias which happens to be an interface, the resolved service actually implements that interface. See example below.

```xml
<services>

    <service id="a_vendor.service" class="Foo\Bar" />
    <service id="Foo\BarInterface" alias="a_vendor.service" />

</services>
```

```php
namespace Foo;

interface BarInterface {}
interface BazInterface {}

class Bar implements BazInterface
{
}
```

```php
public function __construct(private BarInterface $bar) // Type error, $bar must be of type BarInterface
```

Currently the above situation is allowed, and will ultimately lead to a type error at runtime when trying to inject `BarInterface`, because `Bar` is not compatible. I think this situation should be raised as a configuration error, as I can't think of any valid use case for this.

I might be missing some use cases, and this might have been already discussed in some way, so any input appreciated, even if you think this is a bad idea for some reason 🙃 
Thanks!

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
+++ b/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 7.1
 ---
 
+ * Add `CheckAliasValidityPass` to `lint:container` command
  * Add `private_ranges` as a shortcut for private IP address ranges to the `trusted_proxies` option
  * Mark classes `ConfigBuilderCacheWarmer`, `Router`, `SerializerCacheWarmer`, `TranslationsCacheWarmer`, `Translator` and `ValidatorCacheWarmer` as `final`
  * Move the Router `cache_dir` to `kernel.build_dir`
--- a/src/Symfony/Bundle/FrameworkBundle/Command/ContainerLintCommand.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Command/ContainerLintCommand.php
@@ -19,6 +19,7 @@ use Symfony\Component\Console\Exception\RuntimeException;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\DependencyInjection\Compiler\CheckAliasValidityPass;
 use Symfony\Component\DependencyInjection\Compiler\CheckTypeDeclarationsPass;
 use Symfony\Component\DependencyInjection\Compiler\PassConfig;
 use Symfony\Component\DependencyInjection\Compiler\ResolveFactoryClassPass;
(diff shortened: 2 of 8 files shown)


n-valverde:
Few more thoughts for people coming here:
* There is the `lint:container` command which could be useful, but I think does not match exactly this use case. As far as I understand it, it only checks type compatibility of arguments, so that leaves the possibility of defining a bad service from a lib developer side, and the issue will only be spotted by end user when trying to inject the badly configured service (Which is why I came here: api-platform/core:^2.7 is severely affected by this, and running `lint:container` does not spot the issue)
* I am also thinking that adding a flag like `--all` to the `lint:container` command, to make the `CheckTypeDeclarationPass` run before removals could be a good addition, any thought on this?
* I understand that this proposal could be a dead end, with autoload, performance concern, or whatever, so maybe that could just be added to the `lint:container` checks
* Failing tests are due to reflection going through the `trigger_deprecation` calls, and same happens if autoload triggers, so I'm not sure there is any way to deal with that

Friendly ping @nicolas-grekas since you've worked on the initial `lint:container` command 😇

nicolas-grekas:
For sure this shouldn't be part of the always-on passes but if adding this to `lint:container` helps, why not.

fabpot:
@n-valverde Can you work on the change suggested by @nicolas-grekas

n-valverde:
Hello, sorry for the delay, I rebased on 7.1 and made the change so the pass is only applied to `lint:container` but I have additional questions/concerns:

* I have to register this pass before removals, that makes sense for me because of the situation described above, but I'm not sure if this could trigger false positives
* Was about to update the changelog, but I am wondering if this pass should remain in the DependencyInjection component since it is only used by the `lint:container` from FrameworkBundle (but it is the same for `CheckTypeDeclarationsPass`)

fabpot:
> Hello, sorry for the delay, I rebased on 7.1 and made the change so the pass is only applied to `lint:container` but I have additional questions/concerns:
> 
>     * I have to register this pass before removals, that makes sense for me because of the situation described above, but I'm not sure if this could trigger false positives
> 
>     * Was about to update the changelog, but I am wondering if this pass should remain in the DependencyInjection component since it is only used by the `lint:container` from FrameworkBundle (but it is the same for `CheckTypeDeclarationsPass`)

That makes sense to me as it allows developers not using the full stack framework to still use the pass.

fabpot (approved):

fabpot (approved):

fabpot:
Thank you @n-valverde.

PR #50749, opened 2023-06-22
[AssetMapper] Allow DirectoryResource for cache

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | (needed for a bug fix)
| New feature?  | yes/no
| Deprecations? | no
| Tickets       | [None](https://github.com/symfony/ux/issues/961)
| License       | MIT
| Doc PR        | Not needed

Hi!

In StimulusBundle, one file is dynamically generated to contain the list of custom Stimulus controllers. This means the file's contents need to be regenerated in the `dev` environment whenever a new file is added to that directory. This PR allows the `fileDependencies` to also be a directory so that we can use a `DirectoryResource`.

This may not *quite* fit technically as a bug fix. However, as the component is experimental, this affects the `dev` environment only and it will allow for a pretty critical bug fix in StimulusBundle, I've characterized it as a bug fix for 6.3.

Thanks!

### DIFF
--- a/src/Symfony/Component/AssetMapper/Tests/Factory/CachedMappedAssetFactoryTest.php
+++ b/src/Symfony/Component/AssetMapper/Tests/Factory/CachedMappedAssetFactoryTest.php
@@ -17,6 +17,7 @@ use Symfony\Component\AssetMapper\Factory\CachedMappedAssetFactory;
 use Symfony\Component\AssetMapper\Factory\MappedAssetFactoryInterface;
 use Symfony\Component\AssetMapper\MappedAsset;
 use Symfony\Component\Config\ConfigCache;
+use Symfony\Component\Config\Resource\DirectoryResource;
 use Symfony\Component\Config\Resource\FileResource;
 use Symfony\Component\Filesystem\Filesystem;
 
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/AssetMapper/Tests/Factory/CachedMappedAssetFactoryTest.php
```diff
@@ -103,6 +104,7 @@ public function testAssetConfigCacheResourceContainsDependencies()
 
         // just adding any file as an example
         $mappedAsset->addFileDependency(__DIR__.'/../fixtures/importmap.php');
+        $mappedAsset->addFileDependency(__DIR__.'/../fixtures/dir3');
```
stof, line 107: Should we have a `addDirectoryDependency` instead to make the intent more clear ?

weaverryan: Perhaps - though I'm also trying to keep the patch on the smaller side. 

We could merge this as-is, then think about a more generic method name like `addPathDependency()` for 6.4.

nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @weaverryan.

PR #48961, opened 2023-01-12
[WebProfilerBundle] right blocks: fix display

jmsche:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes?
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

Hi, since Symfony 6.2 there's a design issue that bothers me with the toolbar: the hover effect seems "wrong" when hovering one of the right blocks.

This PR fixes the following issues on hover:

* removes green background color
* removes the extra box shadow
* moves the remaining box shadow to the left
* moves the non-radius border from left to right
* uses CSS variables for -green color
* fixes non visible Symfony CLI icon on small screen
* uses same style for green status (bottom colored line instead of full background)

| Before | After |
| --- | --- |
| cli-before | cli-after |
| cli-hover-before | cli-hover-after |
| toolbar-before | toolbar-after |

PR #50752, opened 2023-06-23
[WebProfilerBundle] Fix the accessibility of some background color

javiereguiluz:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

#48961 is a great PR that fixes some small issues in the toolbar. However, I'm worried about one of its changes. Instead of using the custom green color defined for the toolbar, it now uses one of the green colors used in the profiler. The result is that now the profiler has a too low contrast ratio:

before

This PR restores the original green color, which looks like this:

after after-server


fabpot (approved):

fabpot:
Thank you @javiereguiluz.

PR #23834, opened 2017-08-08
[DI] Add "PHP fluent format" for configuring the container

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 3.4
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #22407
| License       | MIT
| Doc PR        | -

This PR allows one to write DI configuration using just PHP, with full IDE auto-completion.
Example:
```php

namespace Symfony\Component\DependencyInjection\Loader\Configurator;

use Symfony\Component\DependencyInjection\Tests\Fixtures\Prototype\Foo;

return function (ContainerConfigurator $c) {

    $c->import('basic.php');

    $s = $c->services()->defaults()
        ->public()
        ->private()
        ->autoconfigure()
        ->autowire()
        ->tag('t', array('a' => 'b'))
        ->bind(Foo::class, ref('bar'))
        ->private();

    $s->set(Foo::class)->args([ref('bar')])->public();
    $s->set('bar', Foo::class)->call('setFoo')->autoconfigure(false);

};
```

PR #50754, opened 2023-06-23
[HttpKernel] when configuring the container add services_{env} with php extension 

helyakin:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | none
| License       | MIT

Hello the community

This is my first PR attempt.

So after reading this [documentation](https://symfony.com/doc/current/service_container.html#remove-services) and unsuccessfully trying to load my `service_test.php`, I've noticed that the `configureContainer(..)` function in the `MicroKernelTrait` file was not configured to automatically load this file.

So either we should fix the documentation, either we should fix the configuration.

Since this the [framework-bundle](https://github.com/symfony/framework-bundle) is backed by [Alximy](https://alximy.io) I figured it would be cool 😎 to try and fix 🐛 the configuration.

Anyway share me your thoughts about what should be done !

And I wanted to say that php service configuration is 🔥 so shoutout to the one who did this (I think it's you @nicolas-grekas with this [PR](https://github.com/symfony/symfony/pull/23834) right ? 💪🏻)

Also big thanks to @jeremyFreeAgent for debugging this with me and @HeahDude for showing me the php service configuration PR.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Kernel/MicroKernelTrait.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Kernel/MicroKernelTrait.php
@@ -58,6 +58,7 @@ trait MicroKernelTrait
             $container->import($configDir.'/{services}_'.$this->environment.'.yaml');
         } else {
             $container->import($configDir.'/{services}.php');
+            $container->import($configDir.'/{services}_'.$this->environment.'.php');
         }
     }
 


chalasr (approved):
While it's a bit late and despite `when()` is the new best practice, I think this doesn't hurt. Not having these files loaded when moving config from yaml to php can be surprising, formats are used to behave the same as each other.

xabbuh (approved):

nicolas-grekas:
We didn't add this because this can be done in PHP directly: `if ('dev' === $env) etc`.
Adding this doesn't hurt of course (but it's not a bugfix)

javiereguiluz (approved):
A nice first contribution to Symfony and with a perfectly explained reasoning behind it. Thanks Marvin!

helyakin:
Thank you everyone for your feedbacks, and thank you @javiereguiluz for those very nice words 😊

I'm really happy to have finally made my first contribution to Symfony 👯 !

chalasr:
Thank you @helyakin.

chalasr:
I didn't open a documentation issue because I think there's no impact there, but I may be wrong. 
@helyakin if you could have a look and create an issue or PR if you find something that needs to be updated, that'd be great.

PR #50761, opened 2023-06-23
[DoctrineBridge] Ignore invalid stores in `LockStoreSchemaListener` raised by `StoreFactory`

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | _NA_
| License       | MIT
| Doc PR        | _NA_

In our test environment, we're setting our `LOCK_DSN` env var at `null`. This works well in 6.2 but not in 6.3 anymore.

The version 6.3 of DoctrineBridge is bringing this BC break. Indeed, lock stores are being injected in `Symfony\Bridge\Doctrine\SchemaListener\LockStoreSchemaListener`. At runtime, the `StoreFactory` is used to instantiate stores when they are needed in `postGenerateSchema`.

Unfortunately, our `LOCK_DSN=null` doesn't match any case in `StoreFactory`, which has the effect to throw `Symfony\Component\Lock\Exception\InvalidArgumentException`, which is not catched in `LockStoreSchemaListener`.

This PR takes care of this: at this time, `LockStoreSchemaListener` only deals with `DoctrineDbalStore`. I think it is safe to ignore any instantiation problem that may occur in the `StoreFactory`.

### DIFF
--- a/src/Symfony/Bridge/Doctrine/SchemaListener/LockStoreSchemaListener.php
+++ b/src/Symfony/Bridge/Doctrine/SchemaListener/LockStoreSchemaListener.php
@@ -12,6 +12,7 @@
 namespace Symfony\Bridge\Doctrine\SchemaListener;
 
 use Doctrine\ORM\Tools\Event\GenerateSchemaEventArgs;
+use Symfony\Component\Lock\Exception\InvalidArgumentException;
 use Symfony\Component\Lock\PersistingStoreInterface;
 use Symfony\Component\Lock\Store\DoctrineDbalStore;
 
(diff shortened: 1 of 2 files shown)


src/Symfony/Bridge/Doctrine/SchemaListener/LockStoreSchemaListener.php
```diff
@@ -28,12 +29,20 @@ public function postGenerateSchema(GenerateSchemaEventArgs $event): void
     {
         $connection = $event->getEntityManager()->getConnection();
 
-        foreach ($this->stores as $store) {
-            if (!$store instanceof DoctrineDbalStore) {
-                continue;
+        $storesIterator = new \ArrayIterator($this->stores);
```
nicolas-grekas, line 32: should be IteratorIterator ?

nicolas-grekas: hum, no, because arrays

nicolas-grekas:
Thank you @alexandre-daubois.

tourze:
This PR make some error while using DoctrineDbalStore, in \Symfony\Bridge\Doctrine\SchemaListener\LockStoreSchemaListener.

See these two picture, iterator_count is greater than zero, but `dump(111);` is not working as expect.

WX20231019-182853

WX20231019-182944

stof:
`iterator_count` will iterate the iterator. So by adding your dump, you modified the behavior of the method here (as you don't rewind the iterator after counting it)

tourze:
After downgrade `symfony/doctrine-bridge` to `6.3.4`, works fine.
WX20231019-184251

tourze:
> `iterator_count` will iterate the iterator. So by adding your dump, you modified the behavior of the method here (as you don't rewind the iterator after counting it)

remove `iterator_count `, still error.

WX20231019-184712

PR #50763, opened 2023-06-23
[DependencyInjection] Skip errored definitions deep-referenced as runtime exceptions

### LINKED ISSUE #50708: #[MapRequestPayload] does not work if a DTO contains **only** nested DTOs

6.3.0+

### Description

If you use `#MapRequestPayload` attribute on a DTO whose only field is a nested DTO, Symfony crashes on the container compilation stage.

### How to reproduce

(Also see https://github.com/someniatko/symfony-50708-reproducer)

Consider a following situation:

```php
class ParentDto
{
    public function __construct(public NestedDto $nested) {}
}

class NestedDto
{
    public function __construct(public string $value) {}
}

class Controller extends AbstractController
{
    #[Route(path: [ '/path' ], methods: 'POST')]
    public function index(#[MapRequestPayload] ParentDto $dto): Response
    {
        return $this->json($dto);
    }
}
```

Then execute `php bin/console cache:clear` for any environment.

**Expected:** Cache clears successfully

**Actual:**
```
In DefinitionErrorExceptionPass.php line 51:
                                                                                                                                                                
  Cannot autowire service "App\Dto\NestedDto": argument "$value" of method "__construct()" is type-hinted "string", you should configure its value explicitly.
```

It took me almost four hours to pinpoint the issue, because I thought maybe some of the bundles I use are the culprit, or my config is incorrect, and also because, and it's the most "funny" part, this will work fine:

```php
class ParentDto
{
    public function __construct(public string $additionalValue, public NestedDto $nested) {}
}

class NestedDto
{
    public function __construct(public string $value) {}
}
```

### Possible Solution

_No response_

### Additional Context

_No response_

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50708
| License       | MIT
| Doc PR        | -

This solves the linked issue by not throwing a compile-time error when the service graph looks like this (when ServiceC has errors):

```
ServiceA ---(runtime-exception-on-invalid-ref)---> ServiceB ---(exception-on-invalid-ref)---> ServiceC
```

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Compiler/DefinitionErrorExceptionPass.php
+++ b/src/Symfony/Component/DependencyInjection/Compiler/DefinitionErrorExceptionPass.php
@@ -11,6 +11,8 @@
 
 namespace Symfony\Component\DependencyInjection\Compiler;
 
+use Symfony\Component\DependencyInjection\Argument\ArgumentInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Exception\RuntimeException;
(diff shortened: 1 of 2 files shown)


NicolJamie:
Understand this has been merged/closed. Not seen any reports of this as of yet, however, I am getting this repeated `OutOfMemoryError` message.

```
Symfony\Component\ErrorHandler\Error\OutOfMemoryError:
Error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 81920 bytes)

  at /application/vendor/symfony/dependency-injection/Compiler/DefinitionErrorExceptionPass.php:69 
```

X-Coder264:
Same here. The other day I had an errored service definition and I also got the `OutOfMemoryError`. Calling `bin/console ca:cl` with `memory_limit=-1` fixed the problem, though it took a very long time for it to finish. After setting a breakpoint in this compiler pass and fixing that service definition `bin/console ca:cl` went back to only taking a few seconds to finish again. When I replaced the code for this compiler pass with the one prior to this change and then tried running `bin/console ca:cl` with the errored service definition it also went back to taking just a few seconds. So this PR is definitively the culprit, it looks like some loop there is not exactly infinite but takes a very long time to finish.

nicolas-grekas:
Please open a dedicated issue and provide a reproducer (even a private one you'd send me on Slack). Commenting on closed PRs doesn't work since we can't track those.

nicolas-grekas:
Please also let me know if the following patch has any effect on your side, both in terms of memory and time:

```patch
--- a/DefinitionErrorExceptionPass.php
+++ b/DefinitionErrorExceptionPass.php
@@ -71,7 +71,7 @@ class DefinitionErrorExceptionPass extends AbstractRecursivePass
                     }
                 }
 
-                unset($this->sourceReferences[$id]);
+                unset($this->sourceReferences[$id], $this->targetReferences[$id]);
             }
```

PR #49134, opened 2023-01-27
[HttpKernel] Add `#[MapQueryParameter]` to map and validate individual query parameters to controller arguments

ruudk:
| Q             | A
| ------------- | ---
| Branch?       | 6.3 
| Bug fix?      |no
| New feature?  | yes
| Deprecations? |no 
| Tickets       | 
| License       | MIT
| Doc PR        | 

We increased the PHPStan level from 8 to 9. This lead to some problems when working with query or request parameters.

For example:
```php
$firstName = $request->get('firstName');
```

Because Symfony types `Request::get()` as `mixed` there is no type safety and you have to assert everything manually.

We then learned that we shouldn't use `Request::get()` but use the explicit parameter bag like:
```php
$request->query->get('firstName');
```

This `ParameterBag` is used for `request` and `query`. It contains interesting methods like:
```php
$request->query->getAlpha('firstName') : string;
$request->query->getInt('age') : int;
```

This has the benefit that now the returned value is of a correct type.

Why aren't we being explicit by requiring the parameters and their types in the controller action instead?

Luckily Symfony has a concept called [ValueResolver](https://symfony.com/doc/current/controller/value_resolver.html).

It allows you to do dynamically alter what is injected into a controller action.

So in this PR, we introduces a new attribute: `#[MapQueryParameter]` that can be used in controller arguments.

It allows you to define which parameters your controller is using and which type they should be. For example:

```php
#[Route(path: '/', name: 'admin_dashboard')]
public function indexAction(
  #[MapQueryParameter]
  array $ids,
  #[MapQueryParameter]
  string $firstName,
  #[MapQueryParameter]
  bool $required,
  #[MapQueryParameter]
  int $age,
  #[MapQueryParameter]
  string $category = '',
  #[MapQueryParameter]
  ?string $theme = null,
)
```

When requesting `/?ids[]=1&ids[]=2&firstName=Ruud&required=3&age=123` you'll get:
```
$ids = ['1', '2']
$firstName = "Ruud"
$required = false
$age = 123
$category = ''
$theme = null
```

It even supports variadic arguments like this:
```php
#[Route(path: '/', name: 'admin_dashboard')]
public function indexAction(
  #[MapQueryParameter]
  string ...$ids,
)
```

When requesting `/?ids[]=111&ids[]=222` the `$ids` argument will have an array with values ['111','222'].

Unit testing the controller now also becomes a bit easier, as you only have to pass the required parameters instead of constructing the `Request` object.

[...]

PR #50767, opened 2023-06-25
[HttpKernel] RequestPayloadValueResolver Add support for custom http status code

zim32:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | no
| License       | MIT
| Doc PR        | Will do if accepted

Reading this discussion https://github.com/symfony/symfony/pull/49134#issuecomment-1476251293 it looks like a good idea to have ability to customize http status code in case of a validation error. 

This MR does not create any BC's.

**MapQueryString** and **MapRequestPayload** attributes now have additional parameter `validationFailedStatusCode`, which allows to specify which http status code will be used if validation fails.

### DIFF
--- a/src/Symfony/Component/HttpKernel/Attribute/MapQueryString.php
+++ b/src/Symfony/Component/HttpKernel/Attribute/MapQueryString.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Component\HttpKernel\Attribute;
 
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Controller\ArgumentResolver\RequestPayloadValueResolver;
 use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
 use Symfony\Component\Validator\Constraints\GroupSequence;
@@ -29,6 +30,7 @@ class MapQueryString extends ValueResolver
         public readonly array $serializationContext = [],
         public readonly string|GroupSequence|array|null $validationGroups = null,
         string $resolver = RequestPayloadValueResolver::class,
+        public readonly int $validationFailedStatusCode = Response::HTTP_NOT_FOUND,
     ) {
         parent::__construct($resolver);
     }
--- a/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php
+++ b/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php
@@ -88,10 +88,10 @@ class RequestPayloadValueResolver implements ValueResolverInterface, EventSubscr
         foreach ($arguments as $i => $argument) {
             if ($argument instanceof MapQueryString) {
                 $payloadMapper = 'mapQueryString';
-                $validationFailedCode = Response::HTTP_NOT_FOUND;
+                $validationFailedCode = $argument->validationFailedStatusCode;
             } elseif ($argument instanceof MapRequestPayload) {
                 $payloadMapper = 'mapRequestPayload';
-                $validationFailedCode = Response::HTTP_UNPROCESSABLE_ENTITY;
+                $validationFailedCode = $argument->validationFailedStatusCode;
             } else {
                 continue;
             }
--- a/src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/RequestPayloadValueResolverTest.php
+++ b/src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/RequestPayloadValueResolverTest.php
@@ -124,7 +124,7 @@ class RequestPayloadValueResolverTest extends TestCase
 
         $resolver->onKernelControllerArguments($event);
 
-        $this->assertEquals([null], $event->getArguments());
+        $this->assertSame([null], $event->getArguments());
     }
 
     public function testQueryNullableValueArgument()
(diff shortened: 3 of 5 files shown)


src/Symfony/Component/HttpKernel/Attribute/MapQueryString.php
```diff
@@ -29,6 +30,7 @@ public function __construct(
         public readonly array $serializationContext = [],
         public readonly string|GroupSequence|array|null $validationGroups = null,
         string $resolver = RequestPayloadValueResolver::class,
+        public readonly int $errorHttpStatusCode = Response::HTTP_NOT_FOUND,
```
mdeboer: Personally I would rename this to `validationFailedCode` I think or `validationFailedStatusCode`? Same ofc. for the `MapRequestPayload` attribute.

zim32: validationFailedStatusCode for me is nicer because it is more clear that this is something related to http status code

mdeboer: Agreed 😄

ghost:
Please modify tests to check it's a status code from the attribute that is thrown.

zim32:
> 

Sorry but I can not find Tests folder in my symfony app http-kernel bundle. What is a workflow to fix and run tests?

src/Symfony/Component/HttpKernel/Attribute/MapQueryString.php
zim32: Changed param name to validationFailedStatusCode

zim32:
> Please modify tests to check it's a status code from the attribute that is thrown.

Figured out. Sorry )

zim32:
> Please modify tests to check it's a status code from the attribute that is thrown.

Two tests added, for query and request payload.

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php
```diff
@@ -88,10 +88,10 @@ public function onKernelControllerArguments(ControllerArgumentsEvent $event): vo
         foreach ($arguments as $i => $argument) {
             if ($argument instanceof MapQueryString) {
                 $payloadMapper = 'mapQueryString';
-                $validationFailedCode = Response::HTTP_NOT_FOUND;
+                $validationFailedCode = $argument->validationFailedStatusCode;
```
BafS, line 91: What about "validationQueryFailedStatusCode"and "validationPayloadFailedStatusCode"? Like that we can differentiate both (and not introduce a BC).

src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/RequestPayloadValueResolverTest.php
```diff
@@ -251,6 +251,7 @@ public function testValidationNotPassed()
             $this->fail(sprintf('Expected "%s" to be thrown.', HttpException::class));
         } catch (HttpException $e) {
             $validationFailedException = $e->getPrevious();
+            $this->assertEquals(404, $e->getStatusCode());
```
BafS:
```suggestion
            $this->assertSame(404, $e->getStatusCode());
```

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php
```diff
@@ -88,10 +88,10 @@ public function onKernelControllerArguments(ControllerArgumentsEvent $event): vo
...
             } elseif ($argument instanceof MapRequestPayload) {
                 $payloadMapper = 'mapRequestPayload';
-                $validationFailedCode = Response::HTTP_UNPROCESSABLE_ENTITY;
+                $validationFailedCode = $argument->validationFailedStatusCode;
```
BafS, line 94: I think it would be great to be able to define another error by default (from the constructor of this class). Otherwise it means that for every attribute we would have to do `validationFailedStatusCode: Response::HTTP_SOMETHING`.

zim32: Agree. But maybe it will be better to do this by another MR?

mdeboer: Don't see why that can't be added to this PR? Seems easy enough to add two constructor arguments and rely on those for the default value. This means in the attribute those arguments should be nullable and null by default. This way we can override it with the argument on the attribute but don't rely on the attribute for the default value of those arguments.

If you want, I can help with that?

zim32: Well this is another feature amd it is always better to separate features into several MR otherwise MR can become quite messy

nicolas-grekas: I'm not sure this is a good idea. Descriptive info (attributes here) should be self-descriptive. Making them rely on some external configuration blurs the line and readers can't expect a stable outcome anymore.

The solution to your concern is to create and use a child class that sets the code to what you want.

src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/RequestPayloadValueResolverTest.php
```diff
@@ -612,3 +680,9 @@ public function __construct(public readonly float $price)
     {
     }
 }
+
+class QueryPayload {
```
nicolas-grekas:
```suggestion
class QueryPayload
{
```

nicolas-grekas (commented):
LGTM. Please add a line to the changelog of the component.

nicolas-grekas (approved):
I fixed the remaining comments.

nicolas-grekas:
Thank you @zim32.

PR #50769, opened 2023-06-25
[HttpClient] declaring dynamic attribut in HttpClientTrait

helyakin:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | no
| New feature?  | yes/no 
| Deprecations? | yes/no 

So as it is php8.2 does not like the HttpClientTrait as it considers that the defaultOptions should not be dynamic.
So this PR is here to declare it officially in the trait


nicolas-grekas:
The property is not dynamic: it should be declared by classes that use the trait (and it is in all classes that do so.)
The trait can be used while the property remains unused (happens when using only a subset of the trait.)
For these reasons (wrong motivation + forces a sometimes dead property), I'm not in favor of this change.

PR #50775, opened 2023-06-26
[EventDispatcher] Throw exception when listener method cannot be resolved

nikophil:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT
| Doc PR        | not needed

while working on #50687 I've discovered bugs on how the method for listener is resolved:

Given this code:
```php
#[AsEventListener(event: KernelEvents::REQUEST, priority: 5)]
class MyListenerNotInvokable
{
    public function someMethod(RequestEvent $event): void {}
}
```
A listener on `MyListenerNotInvokable::onKernelRequest()` is registered, and then an error is dispatched at runtime when the event is dispatched: `Error: Call to undefined method App\EventListener\MyListenerNotInvokable::onKernelRequest()`

The problem comes from [this code](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php#L85-L95): since the `method` is omitted in tag definition, we "camelize" the event's name. We then try to fallback on `__invoke` but because it does not exist, the wrong method name is kept.

this PR fixes this behavior by throwing an exception at compile time if neither the camlized method exist, not the `__invoke` method.

ping @GromNaN

### DIFF
--- a/src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
+++ b/src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
@@ -115,7 +115,11 @@ class RegisterListenersPass implements CompilerPassInterface
                     ], function ($matches) { return strtoupper($matches[0]); }, $event['event']);
                     $event['method'] = preg_replace('/[^a-z0-9]/i', '', $event['method']);
 
-                    if (null !== ($class = $container->getDefinition($id)->getClass()) && ($r = $container->getReflectionClass($class, false)) && !$r->hasMethod($event['method']) && $r->hasMethod('__invoke')) {
+                    if (null !== ($class = $container->getDefinition($id)->getClass()) && ($r = $container->getReflectionClass($class, false)) && !$r->hasMethod($event['method'])) {
+                        if (!$r->hasMethod('__invoke')) {
+                            throw new InvalidArgumentException(sprintf('None of the "%s" or "__invoke" methods exist for the service "foo". Please define the "method" attribute on "kernel.event_listener" tags.', $event['method'], $id, $this->listenerTag));
+                        }
+                        
                         $event['method'] = '__invoke';
                     }
                 }
--- a/src/Symfony/Component/EventDispatcher/Tests/DependencyInjection/RegisterListenersPassTest.php
+++ b/src/Symfony/Component/EventDispatcher/Tests/DependencyInjection/RegisterListenersPassTest.php
@@ -200,10 +200,20 @@ class RegisterListenersPassTest extends TestCase
     public function testInvokableEventListener()
     {
         $container = new ContainerBuilder();
-        $container->register('foo', \stdClass::class)->addTag('kernel.event_listener', ['event' => 'foo.bar']);
+        $container->setParameter('event_dispatcher.event_aliases', [AliasedEvent::class => 'aliased_event']);
+
+        $container->register('foo', \get_class(new class() {
+            public function onFooBar()
+            {
+            }
+        }))->addTag('kernel.event_listener', ['event' => 'foo.bar']);
         $container->register('bar', InvokableListenerService::class)->addTag('kernel.event_listener', ['event' => 'foo.bar']);
         $container->register('baz', InvokableListenerService::class)->addTag('kernel.event_listener', ['event' => 'event']);
-        $container->register('zar', \stdClass::class)->addTag('kernel.event_listener', ['event' => 'foo.bar_zar']);
+        $container->register('zar', \get_class(new class() {
+            public function onFooBarZar()
+            {
+            }
+        }))->addTag('kernel.event_listener', ['event' => 'foo.bar_zar']);
         $container->register('event_dispatcher', \stdClass::class);
 
         $registerListenersPass = new RegisterListenersPass();
@@ -245,6 +255,20 @@ class RegisterListenersPassTest extends TestCase
             ],
         ];
         $this->assertEquals($expectedCalls, $definition->getMethodCalls());
+    }
+
+    public function testItThrowsAnExceptionIfTagIsMissingMethodAndClassHasNoValidMethod()
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('None of the "onFooBar" or "__invoke" methods exist for the service "foo". Please define the "method" attribute on "kernel.event_listener" tags.');
+
+        $container = new ContainerBuilder();
+
+        $container->register('foo', \stdClass::class)->addTag('kernel.event_listener', ['event' => 'foo.bar']);
+        $container->register('event_dispatcher', \stdClass::class);
+
+        $registerListenersPass = new RegisterListenersPass();
+        $registerListenersPass->process($container);
     }
 
     /**


nicolas-grekas:
This should target 5.4 to me since it's a bugfix.

nikophil:
@nicolas-grekas I've rebased the PR

nicolas-grekas (approved):

GromNaN:
Thank you for your findings on this bug. 

> If it does not, we evaluate if one unique public method exists which would take the event as unique parameter (ie: the listener definition above would resolve someMethod() as the listener's method).

This looks like a new feature. Even if there is currently a runtime exception for this situation, I don't think we should support it as a bugfix.
The bugfix should be restricted to throwing an exception on build time when the method (`__invoke` or `onEventName`) doesn't exist.

I see situations where the "only public method" will fail and the developers will have to rework their configuration while it would be working previously:
- a public getter or setter is added for DI or testing
- a second event listener is added

For brevity, developers should set the `AsEventListener` attribute on the method instead of the class.

nikophil:
> This looks like a new feature. Even if there is currently a runtime exception for this situation, I don't think we should support it as a bugfix.
> The bugfix should be restricted to throwing an exception on build time when the method (__invoke or onEventName) doesn't exist.

ok let's do this: I'll split the PR and in the bugfix I'll only throw an exception if the resolved method does not exist

> I see situations where the "only public method" will fail and the developers will have to rework their configuration while it would be working previously:
>
>    a public getter or setter is added for DI or testing
>    a second event listener is added
>
> For brevity, developers should set the AsEventListener attribut on the method instead of the class.

indeed, there will be a failure in these cases, but for both cases, we don't have any way to resolve the method name, then their listener declaration becomes invalid. And since the error will occur in compile time, I think this is acceptable

GromNaN:
> ok let's do this: I'll split the PR and in the bugfix I'll only throw an exception if the resolved method does not exist

👍🏻 Sounds like a plan.

nikophil:
> For brevity, developers should set the AsEventListener attribut on the method instead of the class.

This is just an idea, but since listeners are callables and not actual services, maybe in 7.0 we should allow `#[AsEventListener]` to be set on the class only if the class in invokable? This would simplify a lot all this method-resolving stuff, and remove some un-needed magic

GromNaN:
I wouldn't bother developers with such a restriction. We're not going to deprecate [documented](https://symfony.com/doc/current/event_dispatcher.html#defining-event-listeners-with-php-attributes) way of using this attribute.

nicolas-grekas:
>  I'll split the PR and in the bugfix I'll only throw an exception if the resolved method does not exist

Totally. I missed this was a new feature.

> This is just an idea, but since listeners are callables and not actual services, maybe in 7.0 we should allow #[AsEventListener] to be set on the class only if the class in invokable? This would simplify a lot all this method-resolving stuff, and remove some un-needed magic

I would advise against this. It'd be needlessly rigid.

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -115,8 +115,12 @@ public function process(ContainerBuilder $container)
                     ], function ($matches) { return strtoupper($matches[0]); }, $event['event']);
                     $event['method'] = preg_replace('/[^a-z0-9]/i', '', $event['method']);
 
-                    if (null !== ($class = $container->getDefinition($id)->getClass()) && ($r = $container->getReflectionClass($class, false)) && !$r->hasMethod($event['method']) && $r->hasMethod('__invoke')) {
-                        $event['method'] = '__invoke';
+                    if (null !== ($class = $container->getDefinition($id)->getClass()) && ($r = $container->getReflectionClass($class, false)) && !$r->hasMethod($event['method'])) {
+                        if ($r->hasMethod('__invoke')) {
+                            $event['method'] = '__invoke';
+                        } elseif (!$r->hasMethod($event['method'])) {
+                            throw new InvalidArgumentException(sprintf('Service "%s" must define the "method" attribute on "%s" tags.', $id, $this->listenerTag));
```
GromNaN: You can add that none of `$event['method']` and `__invoke` are defined. That would help to discover the default naming conventions and eventually track a typo in the method name.

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -208,6 +212,53 @@ private function getEventFromTypeDeclaration(ContainerBuilder $container, string
...
         return $name;
     }
+
+    /**
+     * Resolve the listener method:
+     *   - look for a method named after the event (camelized, with "on" prefix)
+     *   - if such a method does not exist, check if only one public method exist which expect the event as parameter
+     *   - otherwise, throw an exception because we need more information.
+     */
+    private function resolveListenerMethodFromTypeDeclaration(ContainerBuilder $container, string $id, string $eventName, array $aliases): string
```
GromNaN: This method is no longer used.

GromNaN (changes requested):
LGTM, after few changes.

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
nikophil: whopsy

nikophil:
CI failures seems unrelated

src/Symfony/Component/EventDispatcher/Tests/DependencyInjection/RegisterListenersPassTest.php
```diff
@@ -247,6 +257,20 @@ public function testInvokableEventListener()
         $this->assertEquals($expectedCalls, $definition->getMethodCalls());
     }
 
+    public function testItThrowsAnExceptionIfTagIsMissingMethodAndClassHasNoValidMethod()
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('Methods "onFooBar" and "__invoke" do not exist for service "foo". Please define the "method" attribute on "kernel.event_listener" tags.');
```
GromNaN: Maybe better like this.
```suggestion
        $this->expectExceptionMessage('None of the "onFooBar" or "__invoke" methods exist for the service "foo". Please define the "method" attribute on "kernel.event_listener" tags.');
```

GromNaN (approved):

src/Symfony/Component/EventDispatcher/Tests/DependencyInjection/RegisterListenersPassTest.php
nikophil: :pray:

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -115,8 +115,12 @@ public function process(ContainerBuilder $container)
...
                     $event['method'] = preg_replace('/[^a-z0-9]/i', '', $event['method']);
 
-                    if (null !== ($class = $container->getDefinition($id)->getClass()) && ($r = $container->getReflectionClass($class, false)) && !$r->hasMethod($event['method']) && $r->hasMethod('__invoke')) {
-                        $event['method'] = '__invoke';
+                    if (null !== ($class = $container->getDefinition($id)->getClass()) && ($r = $container->getReflectionClass($class, false)) && !$r->hasMethod($event['method'])) {
+                        if ($r->hasMethod('__invoke')) {
+                            $event['method'] = '__invoke';
+                        } elseif (!$r->hasMethod($event['method'])) {
+                            throw new InvalidArgumentException(sprintf('None of the "%s" or "__invoke" methods exist for the service "foo". Please define the "method" attribute on "kernel.event_listener" tags.', $event['method'], $id, $this->listenerTag));
+                        }
```
nicolas-grekas:
```suggestion
                        if (!$r->hasMethod('__invoke')) {
                            throw new InvalidArgumentException(sprintf('None of the "%s" or "__invoke" methods exist for the service "foo". Please define the "method" attribute on "kernel.event_listener" tags.', $event['method'], $id, $this->listenerTag));
                        }
                        $event['method'] = '__invoke';
```

nicolas-grekas (commented):
One minor thing. Can you please also update the commit+PR messages to make them tell what this does?

nikophil:
@nicolas-grekas done!

nicolas-grekas (approved):

nicolas-grekas:
Thank you @nikophil.

PR #50776, opened 2023-06-26
[Serializer] Fix type error not be accessed before initialization

shyim:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | https://github.com/shopware/platform/issues/3190
| License       | MIT
| Doc PR        | 

Make the property like in Symfony 6.2, with 6.3 it got the type hint but the default value was missing

### DIFF
--- a/src/Symfony/Component/Serializer/Mapping/ClassMetadata.php
+++ b/src/Symfony/Component/Serializer/Mapping/ClassMetadata.php
@@ -39,7 +39,7 @@ class ClassMetadata implements ClassMetadataInterface
      *           class' serialized representation. Do not access it. Use
      *           {@link getClassDiscriminatorMapping()} instead.
      */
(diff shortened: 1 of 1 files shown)


xabbuh:
I fail to see how the error can actually happen as this property is initialised inside the constructor.

nicolas-grekas:
I guess it can happen when the constructor is not called by a child class. Which child class are we talking about? Isn't this an issue with the child class?

shyim:
I think the problem comes from old cache items, without null on the typed property it gets to uninitialized. Always hard to add types to serialised objects 😅 . We have no class in Shopware which extends that class

https://3v4l.org/ajENt

nicolas-grekas (approved):
That's a risk we need to remember when adding types to properties!

nicolas-grekas:
Thank you @shyim.

PR #50777, opened 2023-06-26
[EventDispatcher] Improve method resolving when event is omitted in tag

nikophil:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| License       | MIT
| Doc PR        | don't know if it is needed

if both event and method names are omitted in the listener's tag, we automatically fallback on `__invoke` 
maybe we can check if a unique public method exists, and then fallback on it?

(not sure if this should be considered as a bugfix or new feature...)

ping @GromNaN

### DIFF
--- a/src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
+++ b/src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
@@ -73,7 +73,7 @@ class RegisterListenersPass implements CompilerPassInterface
                         continue;
                     }
 
-                    $event['method'] ??= '__invoke';
+                    $event['method'] ??= $this->resolveListenerMethodFromPublicMethod($container, $id);
                     $event['event'] = $this->getEventFromTypeDeclaration($container, $id, $event['method']);
                 }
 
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -182,6 +182,23 @@ private function getEventFromTypeDeclaration(ContainerBuilder $container, string
...
+     * If such method exists, return its name, if multiple public methods exist, fallback on `__invoke`.
+     */
+    private function resolveListenerMethodFromPublicMethod(ContainerBuilder $container, string $id): string
+    {
+        if (
+            null === ($class = $container->getDefinition($id)->getClass())
+            || !($r = $container->getReflectionClass($class, false))
+            || 1 !== count($publicMethods = $r->getMethods(\ReflectionMethod::IS_PUBLIC))
+        ) {
+            return '__invoke';
```
nikophil, line 198: I kept this for BC, but I'm wondering if an exception should not be thrown here :thinking:

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -182,6 +182,23 @@ private function getEventFromTypeDeclaration(ContainerBuilder $container, string
...
+    /**
+     * Resolve the listener method by checking if the class has only one public method.
+     * If such method exists, return its name, if multiple public methods exist, fallback on `__invoke`.
+     */
+    private function resolveListenerMethodFromPublicMethod(ContainerBuilder $container, string $id): string
+    {
+        if (
+            null === ($class = $container->getDefinition($id)->getClass())
+            || !($r = $container->getReflectionClass($class, false))
+            || 1 !== \count($publicMethods = $r->getMethods(\ReflectionMethod::IS_PUBLIC))
```
nicolas-grekas, line 196: Shouldn't we filter static methods also?

nicolas-grekas (commented):
LGTM, please rebase

nikophil:
@nicolas-grekas done

failures in the CI are unrelated

derrabus:
Sorry for being late to the party, but what exactly do we gain by adding more guesswork to the method resolution? What's wrong with defaulting to `__invoke()`?

fabpot:
I agree with @derrabus. I would add that this feels a bit magical and arbitrary too.

PR #50783, opened 2023-06-26
[EventDispatcher] Improve method resolving when it is omitted in tag

nikophil:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT
| Doc PR        | will be done if bugfix gets accepted

(this PR is a split of https://github.com/symfony/symfony/pull/50775)

this PR proposes to improve the listener's method resolution by:
1. checking if a method with the "camelized" name exist. 
2. If it does not, we evaluate if one unique public method exists which would take the event as unique parameter (ie: the listener definition above would resolve `someMethod()` as the listener's method). 
3. if an `__invoke` method exist we'll use this one without checking its parameters (this is also useful for BC)
4. Otherwise, method name needs to be provided and an exception is thrown

ping @GromNaN 

(CI failures seems unrelated to this PR :thinking: )

### DIFF
--- a/src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
+++ b/src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
@@ -80,19 +80,7 @@ class RegisterListenersPass implements CompilerPassInterface
                 $event['event'] = $aliases[$event['event']] ?? $event['event'];
 
                 if (!isset($event['method'])) {
-                    $event['method'] = 'on'.preg_replace_callback([
-                        '/(?<=\b|_)[a-z]/i',
-                        '/[^a-z0-9]/i',
-                    ], fn ($matches) => strtoupper($matches[0]), $event['event']);
-                    $event['method'] = preg_replace('/[^a-z0-9]/i', '', $event['method']);
-
-                    if (null !== ($class = $container->getDefinition($id)->getClass()) && ($r = $container->getReflectionClass($class, false)) && !$r->hasMethod($event['method'])) {
-                        if (!$r->hasMethod('__invoke')) {
-                            throw new InvalidArgumentException(\sprintf('None of the "%s" or "__invoke" methods exist for the service "%s". Please define the "method" attribute on "kernel.event_listener" tags.', $event['method'], $id));
-                        }
-
-                        $event['method'] = '__invoke';
-                    }
+                    $event['method'] = $this->resolveListenerMethodFromTypeDeclaration($container, $id, $event['event'], $aliases);
                 }
 
                 $dispatcherDefinition = $globalDispatcherDefinition;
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -182,6 +174,53 @@ private function getEventFromTypeDeclaration(ContainerBuilder $container, string
...
+            return $method;
+        }
+
+        $publicMethods = $r->getMethods(\ReflectionMethod::IS_PUBLIC);
+        $eventName = array_flip($aliases)[$eventName] ?? $eventName;
+
+        $candidateMethods = [];
+        foreach ($publicMethods as $publicMethod) {
+            if (
+                1 === $publicMethod->getNumberOfParameters()
```
nicolas-grekas:
```suggestion
                1 === $publicMethod->getNumberOfRequiredParameters()
```

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -182,6 +174,53 @@ private function getEventFromTypeDeclaration(ContainerBuilder $container, string
...
+                && ($type = $publicMethod->getParameters()[0]->getType()) instanceof \ReflectionNamedType
+                && $type->getName() === $eventName
```
nicolas-grekas: what about using an `is_a` check?

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -182,6 +174,53 @@ private function getEventFromTypeDeclaration(ContainerBuilder $container, string
...
+
+        if (1 === \count($candidateMethods)) {
+            return $candidateMethods[0];
+        }
+
+        if ($r->hasMethod('__invoke')) {
+            return '__invoke';
+        }
+
+        throw new InvalidArgumentException(sprintf('Service "%s" must define the "method" attribute on "kernel.event_listener" tags.', $id));
```
nicolas-grekas:
```suggestion
        throw new InvalidArgumentException(sprintf('Service "%s" is missing a "method" attribute on "kernel.event_listener" tags.', $id));
```

nicolas-grekas (commented):
LGTM, Here are minor comments. Please rebase while addressing them.

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
nicolas-grekas: Ah, shouldn't we filter static methods also?

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
nikophil: done, I added a test case for this

src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php
```diff
@@ -183,6 +171,54 @@ private function getEventFromTypeDeclaration(ContainerBuilder $container, string
...
+        }
+
+        $publicMethods = $r->getMethods(\ReflectionMethod::IS_PUBLIC);
+        $eventName = array_flip($aliases)[$eventName] ?? $eventName;
+
+        $candidateMethods = [];
+        foreach ($publicMethods as $publicMethod) {
+            if (
+                !$publicMethod->isStatic()
+                && 1 === $publicMethod->getNumberOfRequiredParameters()
```
stof, line 204: This check is bad given that the EventDispatcher component actually provides 3 parameters to listeners:
- the event object
- the event name
- the EventDispatcher instance

chalasr:
@nikophil any chance you handle the remaining comments?

nicolas-grekas:
Closing for inactivity. stof pointed out in 2024 that filtering on a single required parameter is wrong, since the dispatcher passes the event, the event name and the dispatcher itself to listeners, and RegisterListenersPass has been reworked since then so this no longer applies cleanly. Please reopen or submit a new PR against the current branch if you want to pick this up again, the idea itself is still welcome.

PR #49539, opened 2023-02-26
[Messenger] make StopWorkerOnSignalsListener listen by default on SIGTERM and SIGINT

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | Fix #49156
| License       | MIT
| Doc PR        | no need I guess

PR #50787, opened 2023-06-27
[Messenger] Fix exiting `messenger:failed:retry` command

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

#49539 introduced a bug where it's impossible to exit the `messenger:failed:retry` command:

Screenshot

`Ctrl+C` doesn't work because the `StopWorkerOnSignalsListener` handles the signal but doesn't actually exit the command, so the only way to currently exit the command is to kill it by force.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
@@ -2098,6 +2098,16 @@ class FrameworkExtension extends Extension
             $container->getDefinition('messenger.transport.beanstalkd.factory')->addTag('messenger.transport_factory');
         }
 
+        if ($config['stop_worker_on_signals'] && $this->hasConsole()) {
+            $container->getDefinition('console.command.messenger_consume_messages')
+                ->replaceArgument(8, $config['stop_worker_on_signals']);
+            $container->getDefinition('console.command.messenger_failed_messages_retry')
+                ->replaceArgument(6, $config['stop_worker_on_signals']);
+        }
+
+        if ($this->hasConsole() && $container->hasDefinition('messenger.listener.stop_worker_signals_listener')) {
+            $container->getDefinition('messenger.listener.stop_worker_signals_listener')->clearTag('kernel.event_subscriber');
+        }
         if ($config['stop_worker_on_signals']) {
             $container->getDefinition('messenger.listener.stop_worker_signals_listener')->replaceArgument(0, $config['stop_worker_on_signals']);
         }
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/console.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/console.php
@@ -163,6 +163,7 @@ return static function (ContainerConfigurator $container) {
                 service('messenger.listener.reset_services')->nullOnInvalid(),
                 [], // Bus names
                 service('messenger.rate_limiter_locator')->nullOnInvalid(),
+                null,
             ])
             ->tag('console.command')
             ->tag('monolog.logger', ['channel' => 'messenger'])
@@ -194,6 +195,7 @@ return static function (ContainerConfigurator $container) {
                 service('event_dispatcher'),
                 service('logger')->nullOnInvalid(),
                 service('messenger.transport.native_php_serializer')->nullOnInvalid(),
+                null,
             ])
             ->tag('console.command')
             ->tag('monolog.logger', ['channel' => 'messenger'])
--- a/src/Symfony/Component/Messenger/Command/ConsumeMessagesCommand.php
+++ b/src/Symfony/Component/Messenger/Command/ConsumeMessagesCommand.php
@@ -15,6 +15,7 @@ use Psr\Container\ContainerInterface;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Command\SignalableCommandInterface;
 use Symfony\Component\Console\Completion\CompletionInput;
 use Symfony\Component\Console\Completion\CompletionSuggestions;
 use Symfony\Component\Console\Exception\InvalidOptionException;
@@ -39,7 +40,7 @@ use Symfony\Component\Messenger\Worker;
  * @author Samuel Roze <samuel.roze@gmail.com>
  */
 #[AsCommand(name: 'messenger:consume', description: 'Consume messages')]
-class ConsumeMessagesCommand extends Command
+class ConsumeMessagesCommand extends Command implements SignalableCommandInterface
 {
     private RoutableMessageBus $routableBus;
     private ContainerInterface $receiverLocator;
@@ -49,8 +50,10 @@ class ConsumeMessagesCommand extends Command
     private ?ResetServicesListener $resetServicesListener;
     private array $busIds;
     private ?ContainerInterface $rateLimiterLocator;
+    private ?array $signals;
+    private ?Worker $worker = null;
 
-    public function __construct(RoutableMessageBus $routableBus, ContainerInterface $receiverLocator, EventDispatcherInterface $eventDispatcher, LoggerInterface $logger = null, array $receiverNames = [], ResetServicesListener $resetServicesListener = null, array $busIds = [], ContainerInterface $rateLimiterLocator = null)
+    public function __construct(RoutableMessageBus $routableBus, ContainerInterface $receiverLocator, EventDispatcherInterface $eventDispatcher, LoggerInterface $logger = null, array $receiverNames = [], ResetServicesListener $resetServicesListener = null, array $busIds = [], ContainerInterface $rateLimiterLocator = null, array $signals = null)
     {
         $this->routableBus = $routableBus;
         $this->receiverLocator = $receiverLocator;
@@ -60,6 +63,7 @@ class ConsumeMessagesCommand extends Command
         $this->resetServicesListener = $resetServicesListener;
         $this->busIds = $busIds;
         $this->rateLimiterLocator = $rateLimiterLocator;
+        $this->signals = $signals;
 
         parent::__construct();
     }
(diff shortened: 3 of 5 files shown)


src/Symfony/Component/Messenger/Command/FailedMessagesRetryCommand.php
```diff
@@ -165,6 +167,16 @@ private function runInteractive(string $failureTransportName, SymfonyStyle $io,
 
     private function runWorker(string $failureTransportName, ReceiverInterface $receiver, SymfonyStyle $io, bool $shouldForce): int
     {
```
lyrixx: We could instead rely on the Signalable Interface?

HypeMC: @lyrixx Unfortunately no, the signal handlers from the `SignalableCommandInterface` are registered before the `WorkerStartedEvent` is triggered. Since the `StopWorkerOnSignalsListener` overrides the signal handlers, it wouldn't work.

lyrixx: I mean, we could rely entirely on the console component, and not implement things here.

HypeMC: So if I understood you correctly, you would replace the `StopWorkerOnSignalsListener` with the `SignalableCommandInterface`? This would then also need to be implemented on the `ConsumeMessagesCommand` as well.

lyrixx: yes 👍🏼

HypeMC: @lyrixx Done, the idea is that if the console component is installed, the listener doesn't get registered.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2098,6 +2098,17 @@ private function registerMessengerConfiguration(array $config, ContainerBuilder
             $container->getDefinition('messenger.transport.beanstalkd.factory')->addTag('messenger.transport_factory');
         }
 
+        if ($this->hasConsole()) {
+            $container->getDefinition('messenger.listener.stop_worker_signals_listener')->clearTag('kernel.event_subscriber');
```
HypeMC, line 2109: It'd probably be better to remove the definition itself, but that'd be a BC break.

src/Symfony/Bundle/FrameworkBundle/Resources/config/console.php
```diff
@@ -192,10 +193,12 @@
                 abstract_arg('Receivers'),
                 service('messenger.routable_message_bus'),
                 service('event_dispatcher'),
-                service('logger'),
+                service('logger')->nullOnInvalid(),
                 service('messenger.transport.native_php_serializer')->nullOnInvalid(),
+                null,
             ])
             ->tag('console.command')
+            ->tag('monolog.logger', ['channel' => 'messenger'])
```
HypeMC, line 201: See #50944

src/Symfony/Component/Messenger/composer.json
```diff
@@ -22,7 +22,7 @@
     },
     "require-dev": {
         "psr/cache": "^1.0|^2.0|^3.0",
-        "symfony/console": "^5.4|^6.0",
+        "symfony/console": "^6.3",
```
HypeMC, line 25: The `SignalableCommandInterface::handleSignal()` methods signature changed in 6.3, it's either this or a conditional trait I guess.

lyrixx (commented):
I like it.

1. Could you deprecate the listener
2. And the service associated with
3. Add a note in the changelog
4. Add a note in the upgrade guide?

Thanks

HypeMC:
> I like it.
> 
> 1. Could you deprecate the listener
> 2. And the service associated with
> 3. Add a note in the changelog
> 4. Add a note in the upgrade guide?
> 
> Thanks

@lyrixx Done

src/Symfony/Component/Messenger/EventListener/StopWorkerOnSignalsListener.php
```diff
@@ -18,6 +18,8 @@
 /**
  * @author Tobias Schultze <http://tobion.de>
  * @author Grégoire Pineau <lyrixx@lyrixx.info>
+ *
+ * @deprecated since Symfony 6.3, use the {@see \Symfony\Component\Console\Command\SignalableCommandInterface} instead
```
nicolas-grekas: deprecations are only for the feature branch (6.4 today)
should we split the PR in two parts, one for 6.3 the other for 6.4?

HypeMC: Makes sense to me, see #50997

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2098,6 +2098,16 @@ private function registerMessengerConfiguration(array $config, ContainerBuilder
...
         }
 
+        if ($config['stop_worker_on_signals'] && $this->hasConsole()) {
+            $container->getDefinition('console.command.messenger_consume_messages')
+                ->replaceArgument(8, $config['stop_worker_on_signals']);
+            $container->getDefinition('console.command.messenger_failed_messages_retry')
+                ->replaceArgument(6, $config['stop_worker_on_signals']);
+        }
+
+        if ($this->hasConsole()) {
```
xabbuh: Why is this `if` necessary?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
xabbuh: The `messenger.listener.stop_worker_signals_listener` service may not exist in case the installed version of the Messenger component is >= 7.0.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
HypeMC: ~~This ensures that the messenger component works even when the console component is not installed (see #50945).~~
No, I'm wrong, the idea was that if the console component is not installed, the listener can be registered since it wouldn't clash with the console.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
HypeMC: Yes, I can update #51093 once this is merged.

fabpot: #51093 has been merged now.

HypeMC: I've added a `hasDefinition()` check, should be good now.

HypeMC:
@lyrixx @xabbuh Hi guys, is there anything else I can do here, this bug is really becoming a nuisance so it'd be great if it got fixed :smile:

src/Symfony/Component/Messenger/Command/ConsumeMessagesCommand.php
```diff
@@ -49,8 +50,10 @@ class ConsumeMessagesCommand extends Command
     private ?ResetServicesListener $resetServicesListener;
     private array $busIds;
     private ?ContainerInterface $rateLimiterLocator;
+    private ?array $signals;
+    private Worker $worker;
```
nicolas-grekas:
```suggestion
    private ?Worker $worker = null;
```

src/Symfony/Component/Messenger/Command/ConsumeMessagesCommand.php
```diff
@@ -222,14 +226,14 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
-        $worker = new Worker($receivers, $bus, $this->eventDispatcher, $this->logger, $rateLimiters);
+        $this->worker = new Worker($receivers, $bus, $this->eventDispatcher, $this->logger, $rateLimiters);
         $options = [
             'sleep' => $input->getOption('sleep') * 1000000,
         ];
         if ($queues = $input->getOption('queues')) {
             $options['queues'] = $queues;
         }
-        $worker->run($options);
+        $this->worker->run($options);
```
nicolas-grekas:
```suggestion
        try {
            $this->worker->run($options);
        } finally {
            $this->worker = null;
        }
```

src/Symfony/Component/Messenger/Command/FailedMessagesRetryCommand.php
```diff
@@ -123,6 +127,24 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
     }
 
+    public function getSubscribedSignals(): array
+    {
+        return $this->signals ?? [\SIGTERM, \SIGINT];
+    }
+
+    public function handleSignal(int $signal, int|false $previousExitCode = 0): int|false
+    {
+        if (!isset($this->worker)) {
```
nicolas-grekas:
```suggestion
        if (!$this->worker) {
```

nicolas-grekas (approved):
LGMT, just minor things on my side.

src/Symfony/Component/Messenger/Command/ConsumeMessagesCommand.php
```diff
@@ -247,6 +251,24 @@ public function complete(CompletionInput $input, CompletionSuggestions $suggesti
...
     }
 
+    public function getSubscribedSignals(): array
+    {
+        return $this->signals ?? [\SIGTERM, \SIGINT];
+    }
+
+    public function handleSignal(int $signal, int|false $previousExitCode = 0): int|false
+    {
+        if (!isset($this->worker)) {
```
nicolas-grekas:
```suggestion
        if (!$this->worker) {
```

src/Symfony/Component/Messenger/Command/ConsumeMessagesCommand.php
HypeMC: Done

src/Symfony/Component/Messenger/Command/ConsumeMessagesCommand.php
HypeMC: Done

src/Symfony/Component/Messenger/Command/FailedMessagesRetryCommand.php
HypeMC: Done

src/Symfony/Component/Messenger/Command/ConsumeMessagesCommand.php
HypeMC: Done

nicolas-grekas:
Thank you @HypeMC.

PR #50788, opened 2023-06-27
[Validator] Fix regression with class metadatada on parent classes

### LINKED ISSUE #50780: Regression in validator

5.4.25, 6.2.12, 6.3.1

### Description

Upgrading symfony/validator from 5.4.34 to 5.4.35 breaks validation in a project of mine.
Some properties that have a NotBlank constraint applied are not detected as invalid when submitted without a value.
It seems that the regression was introduced in this change https://github.com/symfony/symfony/commit/6a3a4d7fb7b5b5358b46edbdfe89a80814e49a00#diff-50c23a0a9ba92c8fb19fa501720ad061da508e3ccfd98fbcdfff2b7c8c6e6499 but unfortunately I cannot understand how.

### How to reproduce

I'm sorry but it seems not so easy to provide a reproducer. In my project, I use XML validation and the problem appears on a property which is not directly defined in the validated object (it's defined in a parent object)

### Possible Solution

Sorry, no clue about this

### Additional Context

I isolated the problem during an upgrade, by requiring in my composer.json `"symfony/validator": "^5.4,<5.4.25"`. With this requirement, all my tests pass.
If I replace it with the simple `"symfony/validator": "^5.4"`, the tests break

rmikalkenas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50780 
| License       | MIT

### DIFF
--- a/src/Symfony/Bridge/Doctrine/Validator/DoctrineLoader.php
+++ b/src/Symfony/Bridge/Doctrine/Validator/DoctrineLoader.php
@@ -108,7 +108,7 @@ final class DoctrineLoader implements LoaderInterface
                 if (isset($mapping['originalClass']) && !str_contains($mapping['declaredField'], '.')) {
                     $metadata->addPropertyConstraint($mapping['declaredField'], new Valid());
                     $loaded = true;
-                } elseif (property_exists($className, $mapping['fieldName'])) {
+                } elseif (property_exists($className, $mapping['fieldName']) && (!$doctrineMetadata->isMappedSuperclass || $metadata->getReflectionClass()->getProperty($mapping['fieldName'])->isPrivate())) {
                     $metadata->addPropertyConstraint($mapping['fieldName'], new Length(['max' => $mapping['length']]));
                     $loaded = true;
                 }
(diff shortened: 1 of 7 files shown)


rmikalkenas:
`Unit Tests / Unit Tests (8.2, high-deps)` - looks like failing `TimezoneValidatorTest::testValidTimezones with data set #9 ('America/Montreal')` is unrelated with the changes.

`Unit Tests / Unit Tests (8.2, low-deps)` - `DoctrineLoaderTest::testLoadClassMetadata` failing, because [doctrineMetadata->fieldMappings](https://github.com/symfony/symfony/pull/50788/files#diff-6104ea4469946639bfc73e0b040e9a5ff857e4a9452d2ddb605140fc955b786cR75) does not contain DoctrineLoaderParentEntity::publicParentMaxLength field. Could it be due to some doctrine bug in lower versions..? Need some help with this one  

cc @nicolas-grekas

nicolas-grekas (commented):
For deps=high, the failure is a false positive.
For deps=low, it's something we need to figure out.

I force-pushed a minor change, see https://github.com/symfony/symfony/compare/60bd66422334099482c110206619eb54cd187c0a..2ba0f57c655ab9921cbaeeb2c60bdb7de781f23c

You can reproduce the issue locally by requiring orm `~2.15.0`:
```sh
COMPOSER_ROOT_VERSION=5.4.x-dev composer require --dev doctrine/orm:~2.15.0
./phpunit src/Symfony/Bridge/Doctrine/ --filter testLoadClassMetadata
```

Does this help?

nicolas-grekas (approved):
I figured out the proper fix on DoctrineLoader.

nicolas-grekas:
Thank you @rmikalkenas.

PR #50790, opened 2023-06-27
[WebProfilerBundle] Split and refactor all the JavaScript code

javiereguiluz:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

As discussed in the past with @stof, the current JavaScript code is less than ideal because everything is coded into the same `Sfjs` object. That's why e.g. the toolbar loads code to handle tabs and togglers (Which don't exist in the database), the logger profiler panel, the mailer profiler panel, etc.

This PR splits the code into different files:

* `base_js.html.twig` keeps the code that it's only used in most/all profiler panels
* `toolbar_js.html.twig` now stores the code that it's only used in the toolbar (mostly about Ajax requests)
* Logger, Mailer, Form, etc. panels now store the code that only applies to those panels

~This feature is ready to review but:~

* ~In `toolbar_js.html.twig` **there's a bug that I can't fix**. **I need your help.** It says "Illegal invocation", which is related to the usage of `this` in the call: `oldXHROpen.apply(this, Array.prototype.slice.call(arguments));` because `this` it's not the XHR object but the new SymfonyWebDebugToolbar class.~
* ~In `toolbar_js.html.twig` I want to refactor more code to use modern JS features after the previous bug is fixed.~

Thanks!


src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/toolbar_js.html.twig
```diff
@@ -9,13 +9,624 @@
...
+                        setTimeout(() => {
+                            copyButton.textContent = oldContent;
+                            copyButton.classList.remove('status-success');
+                        }, 7000);
+                    }
+                });
+            });
+        }
+
+        #request(url, onSuccess, onError, payload, options, tries) {
```
stof: For the toolbar, I strongly suggest sticking to older JS features. Otherwise, we forbid projects to test their own pages in older browsers as the Symfony toolbar would break the JS code.

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/toolbar_js.html.twig
```diff
@@ -9,13 +9,624 @@
...
+                    }
+
+                    return promise;
+                };
+            }
+
+            if (window.XMLHttpRequest && XMLHttpRequest.prototype.addEventListener) {
+                const oldXHROpen = window.XMLHttpRequest.prototype.open;
+
+                window.XMLHttpRequest.prototype.open = (method, url, async, user, pass) => {
```
stof: the fix for your issue about illegal invocation is to _not_ use an arrow function here, as you want to redefine the method and so have access to the `this` of that method invocation.

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/toolbar_js.html.twig
```diff
@@ -9,13 +9,624 @@
...
+                    const self = oldXHROpen;
```
stof: this is broken. In the existing code, we use `self = this`, not `self = proxied`

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/toolbar_js.html.twig
```diff
@@ -9,13 +9,624 @@
...
 </style>
-<script{% if csp_script_nonce %} nonce="{{ csp_script_nonce }}"{% endif %}>/*<![CDATA[*/
-    (function () {
-        Sfjs.loadToolbar('{{ token }}');
-    })();
-/*]]>*/</script>
+
+<script>
+    window.addEventListener('DOMContentLoaded', () => {
+        new SymfonyWebDebugToolbar('{{ token }}');
```
stof: How does this behaves with Turbo ?

Previously, we would have a `loadToolbar` call executed for each Turbo visits (as it would execute the new script tag). Now, we have it only on `DOMContentLoaded` whcih won't be triggered again during a Turbo visit.

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/toolbar_js.html.twig
```diff
@@ -9,13 +9,624 @@
...
+            if (window.fetch && window.fetch.polyfill === undefined) {
+                const oldFetch = window.fetch;
+                window.fetch = () => {
+                    const promise = oldFetch.apply(this, arguments);
+                    const url = arguments[0];
+                    const params = arguments[1];
+                    const paramType = Object.prototype.toString.call(arguments[0]);
+
+                    if ('[object Request]' === paramType) {
+                        url = arguments[0].url;
```
stof, line 331: this is broken if you define your variables as `const` above

stof (changes requested):

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/toolbar_js.html.twig
javiereguiluz: I disagree about this. In my opinion, there are no reasons to have to use an old browser when developing a Symfony app locally. My guess is that 100% of Symfony developers use browsers that support all these features (which are not that advanced, probably only the prehistoric Edge browser doesn't support them).

javiereguiluz:
@stof your review has made me change my mind about some part of this PR. Modernizing the code of the toolbar means that many things will break. Let's keep the old code that works perfectly.

This PR now only splits the toolbar code and modernizes the profiler code.

Thanks!

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/toolbar_js.html.twig
stof: Well, not as your daily browser. But being able to reproduce an issue in another browser to work on a fix is important, and it would be a pain to have to disable the WDT to be able to work on such a fix.

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/toolbar_js.html.twig
```diff
@@ -9,13 +9,672 @@
...
+                    }
+
+                    return this;
+                },
+            };
+        })();
+    }
+
+    window.addEventListener('DOMContentLoaded', () => {
+        Sfjs.loadToolbar('{{ token }}')
```
stof: this is not the way this was done in the old code. Using a DOMContentLoaded listener will likely break things with Turbo as it won't load the new toolbar.

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/base_js.html.twig
```diff
@@ -1,958 +1,217 @@
...
-            var addClass = function(el, cssClass) { if (!hasClass(el, cssClass)) { el.className += " " + cssClass; } };
-            var toggleClass = function(el, cssClass) { hasClass(el, cssClass) ? removeClass(el, cssClass) : addClass(el, cssClass); };
-        }
-
-        var noop = function() {};
-
-        var profilerStorageKey = 'symfony/profiler/';
 
-        var addEventListener;
+{# CAUTION: the contents of this file are processed by Twig before loading
```
stof, line 4: We probably need the same comment in toolbar_js.html.twig

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/base_js.html.twig
```diff
@@ -1,958 +1,217 @@
...
-
-                    if (toggles[i].hasAttribute('data-toggle-initial') && toggles[i].getAttribute('data-toggle-initial') == 'display') {
-                        addClass(toggles[i], 'sf-toggle-on');
-                        addClass(element, 'sf-toggle-visible');
-                    } else {
-                        addClass(toggles[i], 'sf-toggle-off');
-                        addClass(element, 'sf-toggle-hidden');
+                    /* needed because when the toggle contains HTML contents, user can click */
+                    /* on any of those elements instead of their parent '.sf-toggle' element */
+                    while (!toggle.classList.contains('sf-toggle')) {
```
stof: you could use `toggle.closest('.sf-toggle')` instead of a manual implementation of it.

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/base_js.html.twig
```diff
@@ -1,958 +1,217 @@
...
+                        return;
                     }
 
-                    tabGroups[i].setAttribute('data-processed', 'true');
-                }
-            },
-
-            createToggles: function() {
-                var toggles = document.querySelectorAll('.sf-toggle:not([data-processed=true])');
+                    let toggle = e.target || e.srcElement;
```
stof: `e.srcElement` is for old IE, so this can be removed from the code targetting only modern browsers supporting private class fields.

stof (changes requested):

javiereguiluz:
Christophe, thanks for the new review. I've made all the changes that you proposed. Thanks!

stof (approved):

fabpot:
Thank you @javiereguiluz.

PR #50791, opened 2023-06-27
[DependencyInjection] Add `defined` prefix for env var processor

GaryPEGEOT:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | TODO

Allow to add a `defined` prefix:

```yaml
parameters:
  is_google_recaptcha_enabled: '%env(defined:GOOGLE_RECAPTCHA_SITE_KEY)%'
```

Returns `false` if the env var doesn't exist or if it's null or the empty string.
Returns `true` otherwise.

### DIFF
--- a/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
+++ b/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
@@ -56,6 +56,7 @@ class EnvVarProcessor implements EnvVarProcessorInterface
             'require' => 'bool|int|float|string|array',
             'enum' => \BackedEnum::class,
             'shuffle' => 'array',
+            'defined' => 'bool',
         ];
     }
 
@@ -101,6 +102,14 @@ class EnvVarProcessor implements EnvVarProcessorInterface
             }
 
             return $backedEnumClassName::tryFrom($backedEnumValue) ?? throw new RuntimeException(sprintf('Enum value "%s" is not backed by "%s".', $backedEnumValue, $backedEnumClassName));
+        }
+
+        if ('defined' === $prefix) {
+            try {
+                return '' !== ($getEnv($name) ?? '');
+            } catch (EnvNotFoundException) {
+                return false;
+            }
         }
 
         if ('default' === $prefix) {
--- a/src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php
+++ b/src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php
@@ -925,4 +925,21 @@ CSV;
             });
         }));
     }
+
+    /**
+     * @dataProvider provideGetEnvDefined
+     */
+    public function testGetEnvDefined(bool $expected, callable $callback)
+    {
+        $this->assertSame($expected, (new EnvVarProcessor(new Container()))->getEnv('defined', 'NO_SOMETHING', $callback));
+    }
+
+    public static function provideGetEnvDefined(): iterable
+    {
+        yield 'Defined' => [true, fn () => 'foo'];
+        yield 'Falsy but defined' => [true, fn () => '0'];
+        yield 'Empty string' => [false, fn () => ''];
+        yield 'Null' => [false, fn () => null];
+        yield 'Env var not defined' => [false, fn () => throw new EnvNotFoundException()];
+    }
 }
--- a/src/Symfony/Component/DependencyInjection/CHANGELOG.md
+++ b/src/Symfony/Component/DependencyInjection/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * Deprecate `ContainerAwareInterface` and `ContainerAwareTrait`, use dependency injection instead
+ * Add `defined` env var processor that returns `true` for defined and neither null nor empty env vars
 
 6.3
 ---
--- a/src/Symfony/Component/DependencyInjection/Tests/Compiler/RegisterEnvVarProcessorsPassTest.php
+++ b/src/Symfony/Component/DependencyInjection/Tests/Compiler/RegisterEnvVarProcessorsPassTest.php
@@ -50,6 +50,7 @@ class RegisterEnvVarProcessorsPassTest extends TestCase
             'require' => ['bool', 'int', 'float', 'string', 'array'],
             'enum' => [\BackedEnum::class],
             'shuffle' => ['array'],
+            'defined' => ['bool'],
         ];
 
         $this->assertSame($expected, $container->getParameterBag()->getProvidedTypes());


nicolas-grekas:
`bool:default::GOOGLE_RECAPTCHA_SITE_KEY`?

GaryPEGEOT:
> `bool:default::GOOGLE_RECAPTCHA_SITE_KEY`?

I tried with bool first, but any other value than '1' or 'true' get interpreted as false:

```
 // Displaying detailed environment variable usage matching GOOGLE_RECAPTCHA_SITE_KEY                                   

%env(bool:GOOGLE_RECAPTCHA_SITE_KEY)%
-------------------------------------

 ----------------- -------------------------------------------- 
  Default value     n/a                                         
  Real value        "some_value"  
  Processed value   false                                       
 ----------------- -------------------------------------------- 
```

src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php
```diff
@@ -925,4 +925,21 @@ public function testGetEnvCastsNull($expected, string $prefix)
...
+        $this->assertSame(
+            $expected,
+            (new EnvVarProcessor(new Container()))->getEnv('defined', '', fn () => $resolved),
+        );
+    }
+
+    public static function provideGetEnvDefined(): iterable
+    {
+        yield 'Defined' => [true, 'foo'];
+        yield 'Empty' => [false, ''];
```
GromNaN: Given `empty('0') === true`, you should add this case: `[true, '0']` and rework the condition.

GromNaN (changes requested):
Sounds like a useful feature.

nicolas-grekas:
> I tried with bool first, but any other value than '1' or 'true' get interpreted as false:

That's unexpected. We use FILTER_VALIDATE_BOOL, which handles way more cases. The test suite is also proving that.
Not convinced personally sorry. Please provide more insights.

ro0NL:
> Returns true for "1", "true", "on" and "yes". Returns false otherwise.

;)

personally im not to fond of turning `SOME=false` into `bool(true)` and `SOME=0` into `bool(false)`, but yes, this processor would allow to derive a bool value as such. IMHO an empty string check should be used instead of `empty()`.

we do this in runtime for some env: `$enabled = '' !== $value`

however, in hindsight i'd split it into 2 envs, rather than 1 conditional env, thus:

```
$enabled = bool:SOME_ENABLED
$value = string:SOME_VALUE
```

nicolas-grekas (commented):
On second thought, we already do have the `NO_COLOR` env vars which works like that, and this is indeed not the same as `bool`.
On the behavior side, this should not throw when the env var is not defined. It does current if I'm not wrong.
The check should be moved below the code handling "default" and should account for not found vars. I would also not use empty but use a comparison to the empty string only.

GaryPEGEOT:
> On second thought, we already do have the `NO_COLOR` env vars which works like that, and this is indeed not the same as `bool`. On the behavior side, this should not throw when the env var is not defined. It does current if I'm not wrong. The check should be moved below the code handling "default" and should account for not found vars. I would also not use empty but use a comparison to the empty string only.

@nicolas-grekas maybe check for `null` as well? I wouldn't consider null as defined

nicolas-grekas:
`null`and `false` also yes

src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php
GaryPEGEOT: Fixed!

GaryPEGEOT:
Fixed the checks and added a few test cases

src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php
```diff
@@ -925,4 +925,26 @@ public function testGetEnvCastsNull($expected, string $prefix)
             });
         }));
     }
+
+    /**
+     * @dataProvider provideGetEnvDefined
+     */
+    public function testGetEnvDefined(bool $expected, callable $callback)
+    {
+        $this->assertSame(
```
nicolas-grekas: should be on one line

src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php
```diff
@@ -925,4 +925,26 @@ public function testGetEnvCastsNull($expected, string $prefix)
...
+            $expected,
+            (new EnvVarProcessor(new Container()))->getEnv('defined', ':NO_SOMETHING', $callback),
```
nicolas-grekas:
```suggestion
            (new EnvVarProcessor(new Container()))->getEnv('defined', 'NO_SOMETHING', $callback),
```

src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
```diff
@@ -103,6 +104,17 @@ public function getEnv(string $prefix, string $name, \Closure $getEnv): mixed
             return $backedEnumClassName::tryFrom($backedEnumValue) ?? throw new RuntimeException(sprintf('Enum value "%s" is not backed by "%s".', $backedEnumValue, $backedEnumClassName));
         }
 
+        if ('defined' === $prefix) {
+            try {
+                $next = substr($name, $i + 1);
+                $env = $getEnv($next);
```
nicolas-grekas:
```suggestion
                return '' !== ($getEnv($name) ?? '');
```

src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php
```diff
@@ -925,4 +925,26 @@ public function testGetEnvCastsNull($expected, string $prefix)
...
+        );
+    }
+
+    public static function provideGetEnvDefined(): iterable
+    {
+        $returnValue = fn (mixed $value) => fn () => $value;
```
nicolas-grekas:
```suggestion
        $returnValue = fn ($value) => fn () => $value;
```

src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
```diff
@@ -103,6 +104,16 @@ public function getEnv(string $prefix, string $name, \Closure $getEnv): mixed
             return $backedEnumClassName::tryFrom($backedEnumValue) ?? throw new RuntimeException(sprintf('Enum value "%s" is not backed by "%s".', $backedEnumValue, $backedEnumClassName));
         }
 
+        if ('defined' === $prefix) {
+            try {
+                return '' !== ($getEnv($name) ?? '');
+
+                return '' !== $env && null !== $env;
```
nicolas-grekas:
```suggestion
```

GaryPEGEOT: Fixed

nicolas-grekas (approved):

nicolas-grekas (approved):

GromNaN (approved):

src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php
```diff
@@ -925,4 +925,23 @@ public function testGetEnvCastsNull($expected, string $prefix)
...
+    public function testGetEnvDefined(bool $expected, callable $callback)
+    {
+        $this->assertSame($expected, (new EnvVarProcessor(new Container()))->getEnv('defined', 'NO_SOMETHING', $callback));
+    }
+
+    public static function provideGetEnvDefined(): iterable
+    {
+        $returnValue = fn ($value) => fn () => $value;
+
+        yield 'Defined' => [true, $returnValue('foo')];
```
GromNaN: I don't think `$returnValue` is necessary. The other way is less verbose.
```suggestion
        yield 'Defined' => [true, fn () => 'foo'];
```

stof:
@nicolas-grekas but as we don't use `FILTER_NULL_ON_FAILURE`, anything that does not parse as either true or false in `FILTER_VALIDATE_BOOLEAN` gets returned as `false`. This is what happens for this key having a value `some_value`

src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php
GaryPEGEOT: Fair enough

nicolas-grekas:
Thank you @GaryPEGEOT.

PR #50793, opened 2023-06-27
[DependencyInjection] Fix resource tracking for lazy services

### LINKED ISSUE #50792: [VarExporter] Ghost fails when calling internal property when not initialized yet

6.3.1

### Description

I got an exception when I try to access a "static" properties.
```
Typed property App\Lazy\Lazy::$a must not be accessed before initialization
```

### How to reproduce

With this code

```php
<?php

namespace App\Lazy;

class Lazy
{
    private int $a;

    public function __construct(
        private readonly Deps $deps,
    ) {
        dump(__METHOD__);
        $this->a = 1;
    }

    public function do(): int
    {
        // $this->deps->do();
        return $this->doPrivate();

    }

    private function doPrivate(): int
    {
        return $this->a;
    }
}
```

And the configuration
```yaml
service:
    App\Lazy\Lazy:
        lazy: true
```

It throw an exception.

see this [commit](https://github.com/lyrixx/test/commit/27d7a335426af95cb2b850f4549680ee14b01b22) for a better reproducer

**WARNING**

It does work if I clear the cache, but If I add a new property, it does not work anymore.

So to reproduce
1. run the app
1. apply this patch (without cleanring the cache)

    ```diff
    diff --git a/src/Lazy/Lazy.php b/src/Lazy/Lazy.php
    index f40f60a..b2ed79c 100644
    --- a/src/Lazy/Lazy.php
    +++ b/src/Lazy/Lazy.php
    @@ -5,16 +5,19 @@ namespace App\Lazy;
     class Lazy
     {
         private int $a;
    +    private int $b;
     
         public function __construct(
             private readonly Deps $deps,
         ) {
             dump(__METHOD__);
             $this->a = 1;
    +        $this->b = 1;
         }
     
         public function do(): int
         {
    +        $this->b;
             // $this->deps->do();
             return $this->doPrivate();
    ```

1. refresh the page

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50792
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Dumper/PhpDumper.php
+++ b/src/Symfony/Component/DependencyInjection/Dumper/PhpDumper.php
@@ -12,6 +12,7 @@
 namespace Symfony\Component\DependencyInjection\Dumper;
 
 use Composer\Autoload\ClassLoader;
+use Symfony\Component\Config\Resource\FileResource;
 use Symfony\Component\DependencyInjection\Argument\AbstractArgument;
 use Symfony\Component\DependencyInjection\Argument\ArgumentInterface;
 use Symfony\Component\DependencyInjection\Argument\IteratorArgument;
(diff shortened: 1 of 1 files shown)


lyrixx (approved):
👍🏼

src/Symfony/Component/DependencyInjection/Dumper/PhpDumper.php
```diff
@@ -555,8 +556,12 @@ private function generateProxyClasses(): array
                 continue;
             }
             $alreadyGenerated[$asGhostObject][$class] = true;
-            // register class' reflector for resource tracking
-            $this->container->getReflectionClass($class);
+
+            $r = $this->container->getReflectionClass($class);
+            do {
+                $this->container->addResource(new FileResource($r->getFileName()));
```
stof: this should account for cases where `$r->getFilename()` does not return a file path (when it returns `false` for classes defined in C or when it returns a weird string for classes defined through `eval`)

nicolas-grekas: right, updated

PR #46491, opened 2022-05-29
Introduce FlashBagAwareSessionInterface

VincentLanglet:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #41765
| License       | MIT
| Doc PR        | symfony/symfony-docs#... Not needed

This solve https://github.com/symfony/symfony/issues/41765

People will be able to write
```
$session = $request->getSession();
if ($session instanceof FlashBagAwareSessionInterface) {
     $session->addFlashBag(...);
}
```

This is better than a `instanceof Session` check (which is not working with decoration).

PR #50794, opened 2023-06-27
[TwigBridge] Change return type of Symfony\Bridge\Twig\AppVariable::getSession()

### LINKED ISSUE #50789: [TwigBridge] twig-bridge/AppVariable::getSession() has return type Session instead of SessionInterface

6.3.0

### Description

The getSession function in the AppVariable class in TwigBridge has a `Session` return type. This prevents creating a custom session factory / custom session. Currently the following type error is thrown:
`Symfony\Bridge\Twig\AppVariable::getSession(): Return value must be of type ?Symfony\Component\HttpFoundation\Session\Session, App\Framework\CustomSession returned`.

Changing the return type to  `Symfony\Component\HttpFoundation\Session\SessionInterface` will resolve the error.

### How to reproduce

Decorate the 'session.factory' service with a custom session factory in services.yaml:

```
App\Framework\CustomSessionFactory:
    decorates: session.factory
    arguments:
      $requestStack: '@request_stack'
      $storageFactory: '@session.storage.factory'
```

Create a custom session factory:

```
<?php

declare(strict_types=1);

namespace App\Framework;

use Symfony\Component\HttpFoundation\Session\Session;
use Symfony\Component\HttpFoundation\Session\SessionFactoryInterface;
use Symfony\Component\HttpFoundation\Session\SessionInterface;
use Symfony\Component\HttpFoundation\Session\Storage\SessionStorageFactoryInterface;

class CustomSessionFactory implements SessionFactoryInterface {

	public function __construct(private RequestStack $requestStack, private SessionStorageFactoryInterface $storageFactory) {
	}

	public function createSession(): SessionInterface {
		return new CustomSession(
			new Session(
				$this->storageFactory->createStorage($this->requestStack->getMainRequest())
			)
		);
	}
}
```

Create a custom session class: 

```
<?php

declare(strict_types=1);

namespace App\Framework;

use Symfony\Component\HttpFoundation\Session\SessionInterface;

class CustomSession implements SessionInterface {
	public function __construct(private SessionInterface $session) {
	}
        // ... ///

}
```

Render a twig template and the error will be thrown

### Possible Solution

Change return type of `Symfony\Bridge\Twig\AppVariable::getSession()` from `Session` to `SessionInterface`

### Additional Context

_No response_

Dirkhuethorst:
Change the return type of AppVariable.php getSession from 'Session' to 'SessionInterface', to be in line with the return type of getSession in http-foundation/Request.php

| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #50789
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Symfony/Bridge/Twig/AppVariable.php
+++ b/src/Symfony/Bridge/Twig/AppVariable.php
@@ -13,7 +13,7 @@ namespace Symfony\Bridge\Twig;
 
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
-use Symfony\Component\HttpFoundation\Session\Session;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 use Symfony\Component\Security\Core\User\UserInterface;
@@ -93,7 +93,7 @@ class AppVariable
     /**
      * Returns the current session.
      */
-    public function getSession(): ?Session
+    public function getSession(): ?SessionInterface
     {
         if (!isset($this->requestStack)) {
             throw new \RuntimeException('The "app.session" variable is not available.');
--- a/src/Symfony/Bridge/Twig/AppVariable.php
+++ b/src/Symfony/Bridge/Twig/AppVariable.php
@@ -13,7 +13,7 @@ namespace Symfony\Bridge\Twig;
 
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
-use Symfony\Component\HttpFoundation\Session\SessionInterface;
+use Symfony\Component\HttpFoundation\Session\FlashBagAwareSessionInterface;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 use Symfony\Component\Security\Core\User\UserInterface;
@@ -93,7 +93,7 @@ class AppVariable
     /**
      * Returns the current session.
      */
-    public function getSession(): ?SessionInterface
+    public function getSession(): ?FlashBagAwareSessionInterface
     {
         if (!isset($this->requestStack)) {
             throw new \RuntimeException('The "app.session" variable is not available.');
--- a/src/Symfony/Bridge/Twig/AppVariable.php
+++ b/src/Symfony/Bridge/Twig/AppVariable.php
@@ -143,6 +143,10 @@ class AppVariable
             return [];
         }
 
+        if (!method_exists($session, 'getFlashBag')) {
+            return [];
+        }
+
         if (null === $types || '' === $types || [] === $types) {
             return $session->getFlashBag()->all();
         }
--- a/src/Symfony/Bridge/Twig/AppVariable.php
+++ b/src/Symfony/Bridge/Twig/AppVariable.php
@@ -13,7 +13,7 @@ namespace Symfony\Bridge\Twig;
 
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
-use Symfony\Component\HttpFoundation\Session\FlashBagAwareSessionInterface;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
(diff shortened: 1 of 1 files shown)


derrabus:
The failing tests on the low-deps run look related.

src/Symfony/Bridge/Twig/AppVariable.php
```diff
@@ -93,7 +93,7 @@ public function getRequest(): ?Request
     /**
      * Returns the current session.
      */
-    public function getSession(): ?Session
+    public function getSession(): ?FlashBagAwareSessionInterface
```
GromNaN: The return type could be the intersection of all the interfaces implemented by `Session` in `symfony/http-foundation: 5.4`.

```suggestion
    public function getSession(): ?SessionInterface&\Traversable&\Countable
```

Nyholm:
Thank you for this PR. 

I see you replaced the return type from class `Session` to interface `FlashBagAwareSessionInterface`. This is not always correct. It could be a Session that do not implement FlashBag feature. 

I suggest to change it to `SessionInterface`

GromNaN (changes requested):
The `FlashBagAwareSessionInterface` was introduced in `symfony/http-foundation: 6.2` by #46491, and `symfony/twig-bridge: 6.2` allows `symfony/http-foundation: 5.4` in which the interface doesn't exit.
We usually don't update the dependencies constraints in maintenance branches except for security reasons. I don't think we should accept this change as a bugfix, it looks more like a new feature that would target the next version (6.4 ATM).

Also, `Symfony\Component\HttpFoundation\Session\Session` implements `\IteratorAggregate` and `\Countable` on top of `FlashBagAwareSessionInterface` or `SessionInterface`. Changing the return type to this single interface would reduce the functions that could be used in the templates.

Nyholm:
I was hesitating if this was a bugfix or not. But with @GromNaN arguments, I agree. This is a feature.

src/Symfony/Bridge/Twig/AppVariable.php
derrabus: @GromNaN Native intersection types are not nullable. And unions of a intersection type with the `null` type require PHP >= 8.2. That aside, your proposal generates the same problem as the current type: The HttpFoundation `Request` cannot guarantee that the session it returns is `Traversable` and/or `Countable`.

derrabus:
I think, if we change the return type to `?SessionInterface`, we can file it as a bugfix. Using a custom implementation of the `SessionInterface` should've worked in Symfony 5 and it stopped working because of a too narrow native return type in Symfony 6. I believe this breaking change was unintended.

src/Symfony/Bridge/Twig/AppVariable.php
GromNaN: Well, I've forgotten nullable in my proposal. Clearly, the return type cannot be changed in a totally backwards-compatible way.

Noticed [your proposition](https://github.com/symfony/symfony/pull/50794#issuecomment-1611409156) looks more correct.

GromNaN (approved):

Dirkhuethorst:
I've changed the return type to SessionInterface, but now the Psalm check fails because in the AppVariable class the Flashbag of the session is accessed. 

`src/Symfony/Bridge/Twig/AppVariable.php:147:30: UndefinedInterfaceMethod: Method Symfony\Component\HttpFoundation\Session\SessionInterface::getFlashBag does not exist (see https://psalm.dev/181)`

I am not sure how to proceed with this issue

GromNaN:
You can `return []` if `$session` is not an instance of `FlashBagAwareSessionInterface` or `!method_exists($session, 'getFlashBag')`.

Dirkhuethorst:
> You can `return []` if `$session` is not an instance of `FlashBagAwareSessionInterface`.

Won't that break backwards compatibility  with http-foundation 5.4 where the FlashBagAwareSessionInterface does not exist?

src/Symfony/Bridge/Twig/AppVariable.php
```diff
@@ -143,6 +143,10 @@ public function getFlashes(string|array $types = null): array
             return [];
         }
 
+        if (!method_exists($session, 'getFlashBag')) {
```
GromNaN, line 148: You can add a comment that this can be updated in 7.0 to check `FlashBagAwareSessionInterface`, when `symfony/http-foundation: 6.4` is required.

fabpot (approved):

fabpot:
Thank you @Dirkhuethorst.

fabpot:
Merged, but Github didn't get the memo.

PR #50796, opened 2023-06-27
[HttpClient] Fix JsonMockResponse annotation for "body" parameter

neufman:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Allow explicitly mixed type to allow not string types imposed by the MockResponse annotation (e.g.: int, boolean, null, ...) and avoid static code analysis errors.

### DIFF
--- a/src/Symfony/Component/HttpClient/Response/JsonMockResponse.php
+++ b/src/Symfony/Component/HttpClient/Response/JsonMockResponse.php
@@ -15,6 +15,9 @@ use Symfony\Component\HttpClient\Exception\InvalidArgumentException;
 
 class JsonMockResponse extends MockResponse
 {
+    /**
+     * @param mixed $body The json response body to encode
+     */
     public function __construct(mixed $body = [], array $info = [])
     {
         try {


src/Symfony/Component/HttpClient/Response/JsonMockResponse.php
```diff
@@ -15,6 +15,9 @@
 
 class JsonMockResponse extends MockResponse
 {
+    /**
+     * @param mixed $body The json response body to encode
```
fancyweb, line 19: It looks like redundant PHPDoc. Do you mean the real arg type does not have priority over the inherited PHPDoc?
Which SCA tool reports that plz?

stof:
The native type being wider than the parent native type should already be enough for that. I suggest reporting that to your IDE instead.

AFAIK, phpstan already handles that case properly (I haven't checked the behavior of Psalm)

neufman:
I get this error with PHPStan (lvl 7) :

"Parameter #1 $body of class Symfony\Component\HttpClient\Response\JsonMockResponse constructor expects iterable<string|Throwable>|string, array<string, array<int, array<string, int|string>>|bool|string> given."

```php
        $content = [
            'idQueue' => '',
            'hasErrors' => false,
            'hasWorkflowErrors' => false,
            'partiallyImport' => false,
            'errors' => [],
            'messages' => [],
            'messageDetailed' => [],
        ];

        if (isset(self::MAP_URI_MOCK_SUCCESS_RESPONSE[$uri])) {
            $content = self::MAP_URI_MOCK_SUCCESS_RESPONSE[$uri];
        }

        return new JsonMockResponse($content);
```

nicolas-grekas:
That's a bug in the SA tool.

PR #50804, opened 2023-06-28
[Serializer] Fix Normalizer not utilizing converted name for index variadic param

DidierLmn:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

The AbstractNormalizer uses the wrong array index when setting variadic constructor parameters. It checks if the `$key` index (which is the normalized name) exists, but then tries to access the `$paramName` index (which is non-normalized name).

This becomes an issue when using a custom NameConverter, for example one that switches between PascalCase and camelCase.


nicolas-grekas:
Thank you @DidierLmn.

PR #50807, opened 2023-06-28
[HttpClient] Add an HAR response factory for testing

GaryPEGEOT:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Tickets       | 
| License       | MIT
| Doc PR        | TODO

From the network tab of most modern browsers, you can save one or more request as an HAR archive file. This pull request adds a way to use those files in test to replay responses previously saved.

Example usage:

```php
// Inside a KernelTestCase

// ...
$factory = new HarFileResponseFactory('path/to/some_file.har');
static::getContainer()->get('http_client')->setResponseFactory($factory);


// Will find the suitable response(s) inside some_file.har
$myServiceWithHttpCalls->doSomething();
```

### DIFF
--- a/src/Symfony/Component/HttpClient/Test/HarFileResponseFactory.php
+++ b/src/Symfony/Component/HttpClient/Test/HarFileResponseFactory.php
@@ -1,0 +1,97 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\HttpClient\Test;
+
+use Symfony\Component\HttpClient\Exception\TransportException;
+use Symfony\Component\HttpClient\Response\MockResponse;
+use Symfony\Contracts\HttpClient\ResponseInterface;
+
+/**
+ * See: https://w3c.github.io/web-performance/specs/HAR/Overview.html.
+ *
+ * @author Gary PEGEOT <garypegeot@gmail.com>
+ */
+class HarFileResponseFactory
+{
+    public function __construct(private string $archiveFile)
+    {
+    }
+
+    public function setArchiveFile(string $archiveFile): void
+    {
+        $this->archiveFile = $archiveFile;
+    }
+
+    public function __invoke(string $method, string $url, array $options): ResponseInterface
+    {
+        if (!is_file($this->archiveFile)) {
+            throw new \InvalidArgumentException(sprintf('Invalid file path provided: "%s".', $this->archiveFile));
+        }
+
+        $json = json_decode(json: file_get_contents($this->archiveFile), associative: true, flags: \JSON_THROW_ON_ERROR);
+
+        foreach ($json['log']['entries'] as $entry) {
+            /**
+             * @var array{status: int, headers: array, content: array}  $response
+             * @var array{method: string, url: string, postData: array} $request
+             */
+            ['response' => $response, 'request' => $request, 'startedDateTime' => $startedDateTime] = $entry;
+
+            $body = $this->getContent($response['content']);
+            $entryMethod = $request['method'];
+            $entryUrl = $request['url'];
+            $requestBody = $options['body'] ?? null;
+
+            if ($method !== $entryMethod || $url !== $entryUrl) {
+                continue;
+            }
+
+            if (null !== $requestBody && $requestBody !== $this->getContent($request['postData'] ?? [])) {
+                continue;
+            }
+
+            $info = [
+                'http_code' => $response['status'],
+                'http_method' => $entryMethod,
+                'response_headers' => [],
+                'start_time' => strtotime($startedDateTime),
+                'url' => $entryUrl,
+            ];
+
+            /** @var array{name: string, value: string} $header */
+            foreach ($response['headers'] as $header) {
+                ['name' => $name, 'value' => $value] = $header;
+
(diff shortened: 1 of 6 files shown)


src/Symfony/Component/HttpClient/Test/HarFileResponseFactory.php
```diff
@@ -0,0 +1,101 @@
...
+            throw new \InvalidArgumentException(sprintf('Invalid file path provided: "%s".', $this->archiveFile));
+        }
+
+        $json = json_decode(
+            json: file_get_contents($this->archiveFile),
+            associative: true,
+            flags: \JSON_THROW_ON_ERROR
+        );
+
+        foreach ($json['log']['entries'] as $entry) {
```
Nyholm, line 42: We always just look at the first entry, right? 

Or should we return a generator and yield Responses?

Nyholm (commented):
Thank you. 


If the HAR file can contain multiple request/responses. Then I suggest `yielding`. That means we can do: 

```php
$factory = new HarFileResponseFactory($myHarFile);
$client = new MockHttpClient($factory);
```

src/Symfony/Component/HttpClient/Test/HarFileResponseFactory.php
GaryPEGEOT: In the same order they came in the file?

I though about it, but you most likely want to map a request to a specific entry of that file (one with same method / uri / request body)

Nyholm: You are probably correct. The user will probably match uri with request body in most cases. But can we still make it return a Generator? Ie, if a user does include multiple request/responses in the HAR file, we wont silently ignore it. 


-----

Btw, since the HAR file includes method and request URL... would it be possible to match the request with the response.. or is that a separate PR?

stof: the matching is precisely what this loop implements (see the checks below to continue to the next entry.

And the current logic supports multiple requests. When passing a callable (which is the case of this class), it gets invoked for each request to provide a response.

Nyholm: That is awesome. 

Consider the following code: 

```php
$factory = new HarFileResponseFactory($myHarFile);
$client = new MockHttpClient($factory);

$client->request(/** A */);
$client->request(/** B */);
```

And then our MockHttpClient: 

https://github.com/symfony/symfony/blob/d5423dbabd68370190dec217cd7b833027308a0c/src/Symfony/Component/HttpClient/MockHttpClient.php#L62-L85

I thought the callable `$this->responseFactory` was reset after each call to `MockHttpClient::request()`. Since it isn't, it makes no sense to return a generator. 

Sorry for being slow and thank you for the explanations. Im super happy with this PR.

Nyholm (approved):

src/Symfony/Component/HttpClient/Test/HarFileResponseFactory.php
```diff
@@ -0,0 +1,101 @@
...
+    public function __invoke(string $method, string $url, array $options): ResponseInterface
+    {
+        if (!is_file($this->archiveFile)) {
+            throw new \InvalidArgumentException(sprintf('Invalid file path provided: "%s".', $this->archiveFile));
+        }
+
+        $json = json_decode(
+            json: file_get_contents($this->archiveFile),
+            associative: true,
+            flags: \JSON_THROW_ON_ERROR
```
fabpot:
```suggestion
            flags: \JSON_THROW_ON_ERROR,
```

fabpot: Let's be consistent with the code base and use the "short" version here on one line.

fabpot (approved):

fabpot:
Can you rebase on 7.0 and add a note in the CHANGELOG?

src/Symfony/Component/HttpClient/Test/HarFileResponseFactory.php
GaryPEGEOT: Fixed!

src/Symfony/Component/HttpClient/Test/HarFileResponseFactory.php
```diff
@@ -0,0 +1,97 @@
...
+        $this->archiveFile = $archiveFile;
+    }
+
+    public function __invoke(string $method, string $url, array $options): ResponseInterface
+    {
+        if (!is_file($this->archiveFile)) {
+            throw new \InvalidArgumentException(sprintf('Invalid file path provided: "%s".', $this->archiveFile));
+        }
+
+        $json = json_decode(json: file_get_contents($this->archiveFile), associative: true, flags: \JSON_THROW_ON_ERROR);
```
fabpot, line 40:
```suggestion
        $json = json_decode(file_get_contents($this->archiveFile), true, \JSON_THROW_ON_ERROR);
```

GaryPEGEOT: I think third argument is `depth`, that's why is used named arguments. You want me to use default value here?

fabpot: Ok, let's keep it as is

fabpot (approved):

fabpot:
Thank you @GaryPEGEOT.

PR #50809, opened 2023-06-28
Issue 50808 security validator dependency

### LINKED ISSUE #50808: symfony/security-bundle depends on symfony/validator

6.3

### Description

The package `symfony/security-bundle` does not require `symfony/validator` in its composer dependencies, but the configuration registers the `UserPasswordValidator`, which depends on `symfony/validator`.

### How to reproduce

Install `symfony/security-bundle` without `symfony/validator` and let the container build.

### Possible Solution

1. Only register `UserPasswordValidator` if `symfony/validator` is installed
2. Add `symfony/validator` to the requirements in the composer.json

### Additional Context

_No response_

uncaught:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50808
| License       | MIT
| Doc PR        | -

Removes dependency of `symfony/security-bundle` to `symfony/validator`.

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/Resources/config/security.php
+++ b/src/Symfony/Bundle/SecurityBundle/Resources/config/security.php
@@ -58,7 +58,8 @@ return static function (ContainerConfigurator $container) {
         ->set('security.role_hierarchy.roles', [])
     ;
 
-    $container->services()
+    /** @var AbstractServiceConfigurator $configurator */
+    $configurator = $container->services()
         ->set('security.authorization_checker', AuthorizationChecker::class)
             ->args([
                 service('security.token_storage'),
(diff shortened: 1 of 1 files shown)


derrabus (changes requested):
I don't understand why this change is necessary. As long as this service is not wired, it should not cause a problem even if the class does not exist.

uncaught:
@derrabus you get an exception:

image

wouterj:
The stacktrace shows `src/HContent/CoreBundle/DependencyInjection/Compiler/FixAbstractPass.php`, which is a class inside your application. It probably iterates services too eagerly, to cause all classes to be loaded instead of only available classes.

This bug does not occur in a Symfony application without this custom compiler pass. I think this is not a bug that needs fixing in the Symfony code.

uncaught:
Ah darn, yes, thanks, that's definitely the cause on why the class is being loaded. My bad! I'll close this.

PR #50812, opened 2023-06-28
[Messenger] Add auto stamp attribute and middleware

### LINKED ISSUE #50800: [Messenger] Stamps as Attributes on message class
### Description

Some of my messages always have the same stamps in the dispatch, I'd like to reduce this code by adding one or more attributes to my message. 

The envelope constructor could read the stamps assigned to the message

### Example

_No response_

KerianMM:
| Q             | A
| ------------- | ---
| Branch?       | 7.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #50800 
| License       | MIT
| Doc PR        | [symfony/symfony-docs#18461](https://github.com/symfony/symfony-docs/pull/18461)

Add `AutoStamp` attribute on message and configure messenger to enable `AutoStampMiddleware` to automatically add stamps on message.
No needs to pass them when dispatch message.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/messenger.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/messenger.php
@@ -26,6 +26,7 @@ use Symfony\Component\Messenger\EventListener\StopWorkerOnRestartSignalListener;
 use Symfony\Component\Messenger\EventListener\StopWorkerOnSignalsListener;
 use Symfony\Component\Messenger\Handler\RedispatchMessageHandler;
 use Symfony\Component\Messenger\Middleware\AddBusNameStampMiddleware;
+use Symfony\Component\Messenger\Middleware\AutoStampMiddleware;
 use Symfony\Component\Messenger\Middleware\DispatchAfterCurrentBusMiddleware;
 use Symfony\Component\Messenger\Middleware\FailedMessageProcessingMiddleware;
 use Symfony\Component\Messenger\Middleware\HandleMessageMiddleware;
@@ -111,6 +112,8 @@ return static function (ContainerConfigurator $container) {
             ->args([
                 service('router'),
             ])
+
+        ->set('messenger.middleware.auto_stamp_middleware', AutoStampMiddleware::class)
 
         // Discovery
         ->set('messenger.receiver_locator', ServiceLocator::class)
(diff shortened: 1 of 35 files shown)


derrabus:
Maybe it's a stupid idea, but are there arguments against using stamps as attributes directly?

```php
#[AutoStamp(new DelayStamp(123))]
#[AutoStamp([new ValidationStamp(['Default', 'Extra'])])]
class AutoStampedMessage
{
}
```

Vs.

```php
#[DelayStamp(123)]
#[ValidationStamp(['Default', 'Extra'])]
class AutoStampedMessage
{
}
```

KerianMM:
It's done, i'm not sure about second commit
Should we really tag all stamps as class attributes ? Are they really all repeatable ?

kbond:
> Maybe it's a stupid idea, but are there arguments against using stamps as attributes directly?

I think I'm in favor of @derrabus' idea. This allows 3rd party stamps that aren't attributes to be used.

Nyholm:
Interesting. Thank you for this PR.
 
> Maybe it's a stupid idea, but are there arguments against using stamps as attributes directly?

It is the dispatcher that is smart, it knows how long a message should be delayed. It should not be the receiver nor the message itself. 

If we add this kind of configuration on the message, how do a dispatcher opt out? Or how can a dispatcher change the config, Say delay for 10s instead of 5s? (think retry middleware). 

I think we open a can of worms.. What is the reason for adding auto-stamps or adding stamps to the message directly? 
If it is only syntax sugar, then I would prefer keep it explicit as we do now.

derrabus:
> It is the dispatcher that is smart, it knows how long a message should be delayed.

No, the dispatcher should stay dumb. Adding stamps based on attributes (however those might look like) can and should be done by a middleware.

> Or how can a dispatcher change the config, Say delay for 10s instead of 5s? (think retry middleware).

The sttributes on the message should be a default. The retry middleware needs be able to override those defaults.

> What is the reason for adding auto-stamps or adding stamps to the message directly? If it is only syntax sugar, then I would prefer keep it explicit as we do now.

Yes, it is syntactic sugar for wiring a piece of middleware that always adds the same stamps to a specific message class.

KerianMM:
I see it like defaults too.
Maybe the middleware should be smarter and read other stamps to apply attributes stamps on the envelope ?

ro0NL:
>  it is syntactic sugar for wiring a piece of middleware

i tend to believe apps dealing with their default stamps at their middleware layer is the better practice (let it read attributes as you like)

VincentLanglet:
With the current implementation, if I write
```
$bus->dispatch(new MyMessage(), [new DelayStamp(5)]);
```
and the `MyMessage` has a DelayStamp attribute ;
wouldn't the `new DelayStamp(5)` overridden by the attribute of the message in the middleware ?

I would expect the opposite, the attribute should be a "default" but can be overriden when dispatching the message.

I just thought/wanted a similar feature and tried to introduce a SelfStampableInterface in 
https://github.com/symfony/symfony/pull/54366 dunno if it's a better/simpler solution ?

OskarStark:
Thank you @KerianMM for your contribution and the effort you put into this PR\! We're closing this due to lack of recent activity. If this feature is still needed, please feel free to open a new PR against the current branch.

PR #50813, opened 2023-06-28
[DoctrineBridge] Load refreshed user proxy

### LINKED ISSUE #50758: [IMPERSONATOR] user with prop-type user causes the impersonating to fail

6.2.9

### Description

I just came across a weird bug or behaviour.

The issue is caused by the entity `User` with a property of type `User`.
Let's assume you have multiple security roles, so `user1` has role `X` and `user2` has role `Y`.

You are logged in as `user1` and you start to impersonate `user2`, everything still works, but if you stop the impersonation an exception will be thrown.

### How to reproduce

A few instructions to reproduce the issue.

Or reproduce the bug with a [sample project](https://github.com/ludescher/symfony-impersonator-issue).

## 1. install symfony

```sh
$ symfony new --webapp my_project
$ cd my_project
$ composer require symfony/security-bundle
```

## 2. create a user via

Follow the instructions [here](https://symfony.com/doc/current/security.html#the-user).

```sh
$ php bin/console make:migration
$ php bin/console doctrine:migrations:migrate
```

## 3. create basic login page

Follow the instructions [here](https://symfony.com/doc/current/security.html#form-login)

## 4. configure `security.yaml` to match

```yaml
security:
    role_hierarchy:
        ROLE_SYSADMIN:        [ROLE_USER, ROLE_ALLOWED_TO_SWITCH]
        ROLE_SALE:             ROLE_USER

    password_hashers:

    providers:
        app_user_provider:
            entity:
                class: App\Entity\User
                property: email
    firewalls:
        dev:
            security: false
        main:
            lazy: true
            provider: app_user_provider
            form_login:
                login_path: app_login
                check_path: app_login
            switch_user: true
    access_control:
        - { path: ^/login,                   roles: PUBLIC_ACCESS }
        - { path: ^/,                        roles: ROLE_USER }
```

## 5. create two Users

`test1@symfony.com` with ROLE `ROLE_SYSADMIN`

and

`test2@symfony.com` with ROLE `ROLE_SALE`

## 6. finally create a new Controller `HomeController`

```php
namespace App\Controller;

class HomeController extends AbstractController {

    #[Route('/', name: 'app_home')]
    public function home(): Response {
        return $this->render('home.html.twig');
    }
}

```

```twig
{# templates/home.html.twig #}
{% extends 'base.html.twig' %}

{# ... #}

{% block body %}
    <h1>
        {{ app.user.email }}
    </h1>

    <div style="margin: 25px">
        {% if is_granted('IS_IMPERSONATOR') %}
            <a href="{{ path('app_home', {'_switch_user': '_exit'}) }}">
                Stop impersonating
            </a>
        {% endif %}
    </div>

    <div style="margin: 25px">
        {% if app.user.email != 'test2@symfony.com' %}
            <a class="btn" href="{{ path('app_home', { '_switch_user': 'test2@symfony.com' }) }}">
                Start impersonating
            </a>
        {% endif %}
    </div>
{% endblock %}
```
## 7. testing

Now we have everything ready to start testing.

Login as `test1@symfony.com` and start impersonating.

Until now everything should work just fine. But if we modify the `User` entity and add a new property `created_by` with type `User`, this causes the impersonating to fail.

```php
+    #[ORM\ManyToOne(targetEntity: self::class)]
+    private ?self $created_by = null;
```

Right after we stop the impersonating of a user we get an error.
Oddly, when we refresh the page, the problem goes away. Therefore, the problem definitely lies in the user transition.

`App\Entity\User::getPassword(): Return value must be of type string, null returned`

### Possible Solution

When I looked at the callstack, I initially assumed that the error must come from the security package.
But I've already spoken to the Symfony security team and they assured me that it's not a security issue and it might be in the User class that returns an invalid value.

At this point I could not find any indication of where the problem really comes from.

## callstack:

App/Entity/User.php&line=###

```php
    /**
     * @see PasswordAuthenticatedUserInterface
     */
    public function getPassword(): string {
        return $this->password; // <<<
    }
```

> in vendor/symfony/security-http/Firewall/ContextListener.php -> getPassword (line 290)
>
> in vendor/symfony/security-http/Firewall/ContextListener.php :: hasUserChanged (line 212)
>
> in vendor/symfony/security-http/Firewall/ContextListener.php -> refreshUser (line 127)
>
> in vendor/symfony/security-bundle/Debug/WrappedLazyListener.php -> authenticate (line 46)
>
> in vendor/symfony/security-bundle/Security/LazyFirewallContext.php -> authenticate (line 60)
>
> in vendor/symfony/security-bundle/Debug/TraceableFirewallListener.php -> __invoke (line 70)
>
> in vendor/symfony/security-http/Firewall.php -> callListeners (line 92)
>
> in vendor/symfony/event-dispatcher/Debug/WrappedListener.php -> onKernelRequest (line 116)
>
> in vendor/symfony/event-dispatcher/EventDispatcher.php -> __invoke (line 206)
>
> in vendor/symfony/event-dispatcher/EventDispatcher.php -> callListeners (line 56)
>
> in vendor/symfony/event-dispatcher/Debug/TraceableEventDispatcher.php -> dispatch (line 127)
>
> in vendor/symfony/http-kernel/HttpKernel.php -> dispatch (line 139)
>
> in vendor/symfony/http-kernel/HttpKernel.php -> handleRaw (line 74)
>
> in vendor/symfony/http-kernel/Kernel.php -> handle (line 184)
>
> in vendor/symfony/runtime/Runner/Symfony/HttpKernelRunner.php -> handle (line 35)
>
> in vendor/autoload_runtime.php -> run (line 29)
>
> in public/index.php (line 5)

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50758
| License       | MIT
| Doc PR        | N/A

When [exiting impersonation](https://github.com/symfony/symfony/blob/529973b88db84aee4e962b2ba8bb1838bbbe6a38/src/Symfony/Component/Security/Http/Firewall/SwitchUserListener.php#L210), the impersonator get refreshed in the original token. Problem is, the refreshed user can be a proxy if it is referenced by the impersonated. In that case, all its properties (except identifiers) will be `null` when it is unserialised by the `ContextListener`.

The refreshed user data will be needed anyways, so loading proxies in `EntityUserProvider::refreshUser` should not impact performance.

### DIFF
--- a/src/Symfony/Bridge/Doctrine/Security/User/EntityUserProvider.php
+++ b/src/Symfony/Bridge/Doctrine/Security/User/EntityUserProvider.php
@@ -15,6 +15,7 @@ use Doctrine\Persistence\ManagerRegistry;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\ObjectManager;
 use Doctrine\Persistence\ObjectRepository;
+use Doctrine\Persistence\Proxy;
 use Symfony\Component\Security\Core\Exception\UnsupportedUserException;
 use Symfony\Component\Security\Core\Exception\UserNotFoundException;
 use Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface;
@@ -115,6 +116,10 @@ class EntityUserProvider implements UserProviderInterface, PasswordUpgraderInter
 
                 throw $e;
             }
+        }
+
+        if ($refreshedUser instanceof Proxy && !$refreshedUser->__isInitialized()) {
+            $refreshedUser->__load();
         }
 
         return $refreshedUser;
--- a/src/Symfony/Bridge/Doctrine/Tests/Security/User/EntityUserProviderTest.php
+++ b/src/Symfony/Bridge/Doctrine/Tests/Security/User/EntityUserProviderTest.php
@@ -16,6 +16,7 @@ use Doctrine\ORM\Tools\SchemaTool;
 use Doctrine\Persistence\ManagerRegistry;
 use Doctrine\Persistence\ObjectManager;
 use Doctrine\Persistence\ObjectRepository;
+use Doctrine\Persistence\Proxy;
 use PHPUnit\Framework\TestCase;
 use Symfony\Bridge\Doctrine\Security\User\EntityUserProvider;
 use Symfony\Bridge\Doctrine\Security\User\UserLoaderInterface;
@@ -195,6 +196,27 @@ class EntityUserProviderTest extends TestCase
         );
 
         $provider->upgradePassword($user, 'foobar');
+    }
+
+    public function testRefreshedUserProxyIsLoaded()
+    {
+        $em = DoctrineTestHelper::createTestEntityManager();
+        $this->createSchema($em);
+
+        $user = new User(1, 1, 'user1');
+
+        $em->persist($user);
+        $em->flush();
+        $em->clear();
+
+        // store a proxy in the identity map
+        $em->getReference(User::class, ['id1' => 1, 'id2' => 1]);
+
+        $provider = new EntityUserProvider($this->getManager($em), User::class);
+        $refreshedUser = $provider->refreshUser($user);
+
+        $this->assertInstanceOf(Proxy::class, $refreshedUser);
+        $this->assertTrue($refreshedUser->__isInitialized());
     }
 
     private function getManager($em, $name = null)


nicolas-grekas (approved):

nicolas-grekas:
Thank you @MatTheCat.

PR #50819, opened 2023-06-29
[SecurityBundle] Do not translate `Bearer` header’s `error_description`

### LINKED ISSUE #50811: Non ASCII characters in response headers

6.3 (previous versions should be affected too)

### Description

When access token authentication fails, a response is sent with a WWW-Authenticate headers which sometimes contains non ASCII characters, which are invalid characters for a header value.

This is especially problematic when hosting our website on Azure appservices because the webserver on Azure side will fail to forward the response and return an empty 500 error response instead of a 401 for example. This is not a bug from Azure however, webservers should not support those characters, some do but not everyone of them.

Those non ASCII characters only appear for certain translations, for example in spanish, the WWW-Authenticate header will have this value 
`Bearer error="invalid_token",error_description="Credenciales no vÃ¡lidas."`
Containing the character `Ã` or `0x00C3`

I believe the error_description should not be translated and only return a translation key, or another way to return the translated value should be used (response body for example).

### How to reproduce

Throw an exception that will trigger this response (it may be required to throw this error inside a token_handler for an access_token firewall)

For example
```throw new BadCredentialsException('Invalid credentials.')```

You should see a WWW-Authenticate header in the given response, on some locale, for example es_ES, the header contains non ASCII characters.

### Possible Solution

Inside the AccessTokenAuthenticator file, the method `onAuthenticationFailure` should not attempt to translate the messageKey.

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by MatTheCat**
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50811
| License       | MIT
| Doc PR        | N/A

From [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750#section-3):

> the "error_description" attribute [provides] a human-readable explanation that is not meant to be displayed to end-users.

Not translating it avoids non-ASCII characters in `Bearer`’s value, which in turn avoid messing with [RFC 8187](https://www.rfc-editor.org/rfc/rfc8187.html) (fun read; thanks @derrabus!).

Not sure if and how this should be tested :thinking:

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/Resources/config/security_authenticator_access_token.php
+++ b/src/Symfony/Bundle/SecurityBundle/Resources/config/security_authenticator_access_token.php
@@ -33,7 +33,6 @@ return static function (ContainerConfigurator $container) {
                 null,
                 null,
             ])
-            ->call('setTranslator', [service('translator')->ignoreOnInvalid()])
 
         ->set('security.authenticator.access_token.chain_extractor', ChainAccessTokenExtractor::class)
             ->abstract()

### CODE REVIEW

File: src/Symfony/Component/Security/Http/Authenticator/AccessTokenAuthenticator.php [hunk 1]
```diff
@@ -97,7 +91,7 @@ public function onAuthenticationFailure(Request $request, AuthenticationExceptio
 
     public function setTranslator(?TranslatorInterface $translator)
     {
-        $this->translator = $translator;
+        trigger_deprecation('symfony/security-http', '6.2', '"%s()" has no effect and will be removed in 7.0.', __METHOD__);
```
derrabus: I would postpone the runtime deprecation to 6.4. We usually don't deprecate stuff in bugfix releases.

MatTheCat: So, no-op for 6.2 and adding `trigger_deprecation` in 6.4?

Should I tag the method `@deprecated` in this PR? :thinking:

nicolas-grekas: WDYT of just not wiring this on 6.2 (aka only patching the bundle) and no-op + deprec go on 6.4?

nicolas-grekas: or better: we keep the translator in place, no need to change the behavior in 6.*, unwiring is enough to fix the issue

nicolas-grekas (approved):

derrabus (approved):

nicolas-grekas:
Thank you @MatTheCat.

src/Symfony/Component/Security/Http/Authenticator/AccessTokenAuthenticator.php
stof: I still suggest to deprecate the method in 6.4 to be able to remove it in 7.0

PR #50821, opened 2023-06-29
Add missing return types and enforce return types on all methods

wouterj:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Tickets       | -
| License       | MIT
| Doc PR        | -

In 6.3, we enforced return types on all interface methods. We're missing just a handful return types on non-interface methods (all fixed in this PR). As you can see, the added return types are either data collectors or methods introduced by 6.3 features.

In my opinion, we should start to enforce return types on all methods from 6.4 (either as PHPdoc - for `resource` and to avoid BC breaks - or PHP native declaration).

### DIFF
--- a/src/Symfony/Bridge/Doctrine/Form/DoctrineOrmTypeGuesser.php
+++ b/src/Symfony/Bridge/Doctrine/Form/DoctrineOrmTypeGuesser.php
@@ -12,6 +12,7 @@
 namespace Symfony\Bridge\Doctrine\Form;
 
 use Doctrine\DBAL\Types\Types;
+use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\Mapping\MappingException as LegacyMappingException;
 use Doctrine\Persistence\ManagerRegistry;
@@ -151,6 +152,13 @@ class DoctrineOrmTypeGuesser implements FormTypeGuesserInterface
         return null;
     }
 
+    /**
+     * @template T of object
+     *
+     * @param class-string<T> $class
+     *
+     * @return array{0:ClassMetadata<T>, 1:string}|null
+     */
     protected function getMetadata(string $class)
     {
         // normalize class name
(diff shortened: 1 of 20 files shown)


wouterj:
@nicolas-grekas I'm afraid I need your help with the data collector return types. It sometimes seem to return a `Data` object and other times a plain value. I can't really reason when a method can return `Data`. I've fixed the types that caused the tests to fail, but I'm not sure about the untested methods. Can you have a close look at them? :pray:

nicolas-grekas:
What about going with mixed? We changed some of them to return Data instead of the original data because from the pov of a twig template, we can make a Data object behave as the original one. mixed and `@return Data|mixed` if that makes sense?

stof:
`Data|mixed` is still mixed though.

src/Symfony/Component/Form/Extension/DataCollector/FormDataCollector.php
```diff
@@ -242,6 +242,9 @@ protected function getCasters(): array
         ];
     }
 
+    /**
+     * @return array
```
stof: As this is a private method, why not using a native one ?

wouterj:
Alright, I did some further testing and it seems like scalar types are kept as-is, while other types become `Data` when calling `cloneVar()`. For the data collectors that do `$this->cloneVar($this->data)`, I've updated the non-scalar return types to be `Data|array`. Seems like that does the trick (at least, to make all tests green).

src/Symfony/Bridge/Doctrine/Form/DoctrineOrmTypeGuesser.php
```diff
@@ -151,6 +152,15 @@ public function guessPattern(string $class, string $property): ?ValueGuess
...
     }
 
+    /**
+     * @template T of object
+     *
+     * @psalm-param class-string<T> $className
+     *
+     * @return array|null
+     *
+     * @psalm-return array{0:ClassMetadata<T>, 1:string}|null
```
nicolas-grekas:
```suggestion
     * @param class-string<T> $className
     *
     * @return array{0:ClassMetadata<T>, 1:string}|null
```

nicolas-grekas:
Thank you @wouterj.

Issue #50815, opened 2023-06-29
Null environment variable cast to empty string


fmarchalemisys:
### Symfony version(s) affected

6.3.1

### Description

The following code used to work with symfony 6.3.0 but not with 6.3.1.

In `services.yaml`:

```yaml
parameters:
    env(HTTP_UNIQUEID): ~
```

In a service:

```php
    public function __construct(
        #[Autowire('%env(string:HTTP_UNIQUEID)%')] ?string $xRequestId
    ) {
        // symfony 6.3.0: $xRequestId is null
        // symfony 6.3.1: $xRequestId is ""
        $this->xRequestId = $xRequestId ?? $this->generateSafeUnpredictableId();
    }
```

With symfony 6.3.1, the `$xRequestId` is the empty string. It defeats the call to `generateSafeUnpredictableId()`.

This issues was introduced by this change: https://github.com/symfony/dependency-injection/commit/6c3c271943ec6ea1a0233f14dd5f1586c9fe77f4#diff-0ebfb024d710d87ccb36098690a9aa8ec4cea8bae635a2c5f95813e3ceba3dabR184

This is a big BC break for a minor version change.

### How to reproduce

In `services.yaml`:

```yaml
parameters:
    env(HTTP_UNIQUEID): ~
```

In a service:

```php
    public function __construct(
        #[Autowire('%env(string:HTTP_UNIQUEID)%')] ?string $xRequestId
    ) {
        // symfony 6.3.0: $xRequestId is null
        // symfony 6.3.1: $xRequestId is ""
        $this->xRequestId = $xRequestId ?? $this->generateSafeUnpredictableId();
        // $this->xRequestId is the very predictable empty string!
    }
```

### Possible Solution

Replace 

```php
if (!\in_array($prefix, ['string', 'bool', 'not', 'int', 'float'], true)) {
    return null;
}
```

with

```
return null;
```

when `$env` is null in `EnvVarProcessor.php`.

### Additional Context

Apache with the unique_id module generates a unique id for each request it processes. The unique id is stored in `HTTP_UNIQUEID`.

When the unique_id module is not enabled or caddy is used instead of apache, there are no `HTTP_UNIQUEID`. It must be generated on the fly by the service.

As a quick patch, I simply check if the env var is the empty string but this is a big BC break that's not easy to spot.

fancyweb:
Thanks for reporting the issue with detailed info.

[...]

PR #50837, opened 2023-06-30
[DependencyInjection] Fix autocasting null env values to empty string

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | -
| Deprecations? | -
| Tickets       | https://github.com/symfony/symfony/issues/50815#issuecomment-1612586382
| License       | MIT
| Doc PR        | -

This PR fixes autocasting `null` env values to `''`.
We continue to autocast all other values to string when there's no prefix.

It doesn't revert https://github.com/symfony/symfony/pull/50517, it just fixes an unintended side effect.

It doesn't fix https://github.com/symfony/symfony/issues/50815 request, it confirms the behavior: the solution for this issue would be to remove the `string:` prefix.

There's still no solution to have the "keep null as null" and "cast if not null" behavior.

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Container.php
+++ b/src/Symfony/Component/DependencyInjection/Container.php
@@ -390,7 +390,15 @@ class Container implements ContainerInterface, ResetInterface
             $prefix = 'string';
             $localName = $name;
         }
-        $processor = $processors->has($prefix) ? $processors->get($prefix) : new EnvVarProcessor($this);
+
+        if ($processors->has($prefix)) {
+            $processor = $processors->get($prefix);
+        } else {
+            $processor = new EnvVarProcessor($this);
+            if (false === $i) {
+                $prefix = '';
+            }
+        }
 
         $this->resolving[$envName] = true;
         try {
--- a/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
+++ b/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
@@ -126,6 +126,12 @@ class EnvVarProcessor implements EnvVarProcessorInterface
             }
         }
 
+        $returnNull = false;
+        if ('' === $prefix) {
+            $returnNull = true;
+            $prefix = 'string';
+        }
+
         if (false !== $i || 'string' !== $prefix) {
             $env = $getEnv($name);
         } elseif (isset($_ENV[$name])) {
@@ -177,6 +183,10 @@ class EnvVarProcessor implements EnvVarProcessorInterface
         }
 
         if (null === $env) {
+            if ($returnNull) {
+                return null;
+            }
+
             if (!isset($this->getProvidedTypes()[$prefix])) {
                 throw new RuntimeException(sprintf('Unsupported env var prefix "%s".', $prefix));
             }
--- a/src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php
+++ b/src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php
@@ -763,12 +763,13 @@ CSV;
 
     /**
      * @testWith    ["", "string"]
+     *              [null, ""]
      *              [false, "bool"]
      *              [true, "not"]
      *              [0, "int"]
      *              [0.0, "float"]
      */
-    public function testGetEnvCastsNull($expected, string $prefix)
+    public function testGetEnvCastsNullBehavior($expected, string $prefix)
     {
         $processor = new EnvVarProcessor(new Container());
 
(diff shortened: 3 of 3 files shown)


src/Symfony/Component/DependencyInjection/Container.php
```diff
@@ -387,7 +387,7 @@ protected function getEnv(string $name)
             $prefix = substr($name, 0, $i);
             $localName = substr($name, 1 + $i);
         } else {
-            $prefix = 'string';
+            $prefix = '';
```
fancyweb: Damn, what if people override the default string prefix :thinking: This is probably not a good solution.

src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
```diff
@@ -177,6 +177,10 @@ public function getEnv(string $prefix, string $name, \Closure $getEnv)
         }
 
         if (null === $env) {
+            if ('' === $prefix) {
```
fancyweb, line 180: `''` becomes a magic value that acts as the `string` prefix except for null where we return null

src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
```diff
@@ -126,6 +126,12 @@ public function getEnv(string $prefix, string $name, \Closure $getEnv)
             }
         }
 
+        $returnNull = false;
+        if ('' === $prefix) {
```
fancyweb, line 130: `''` becomes a magic value that acts as the `string` prefix except for null where we return null

nicolas-grekas:
Thank you @fancyweb.

melkamar:
I may have shot myself in the foot with https://github.com/symfony/symfony/issues/50415 😁  I have now upgraded to a version including this fix and am having issues with Swiftmailer configuration:

```yaml
swiftmailer:
    encryption: '%env(string:default::MAILER_ENCRYPTION)%'
```

With this and `MAILER_ENCRYPTION` unset, the parameter now becomes an empty string ``. However, the Swiftmailer library does not like that as it only handles `null` in a special way, not an empty string, and throws 

```
  [InvalidArgumentException]
  The  encryption is not supported
```

When I change the definition of the parameter to 

```yaml
swiftmailer:
    encryption: '%env(default::MAILER_ENCRYPTION)%'
```

which, as I understand this PR, should do what I expect it to (pass `null` as `null`), the container build crashes on type mismatch:

```
  [Symfony\Component\Config\Definition\Exception\InvalidTypeException]
  Invalid type for path "swiftmailer.mailers.default.encryption". Expected one of "bool", "int", "float", "string", but got one of "bool", "int", "float", "string", "array".
```

This is because the SwiftMailer [configuration](https://github.com/symfony/swiftmailer-bundle/blob/main/DependencyInjection/Configuration.php#L133C11-L133C11) expects a scalar node but the result of the env-var-processing may be an array (though not in my particular case, I think).

This seems to be the case that you wrote about in the PR description

> There's still no solution to have the "keep null as null" and "cast if not null" behavior.

It feels like there is a disconnect between what's possible with the parameter/yaml based configuration and what's possible to do with environment variables. One solution to the problem above would be to have a "nullable" set of prefixes, but I don't know if that's making things even more complicated.

-----

Edit: this seems to do the trick in our code, at least:

```yaml
swiftmailer:
    encryption: '%env(nullablestring:default::MAILER_ENCRYPTION)%'
```

```php
<?php

use Symfony\Component\DependencyInjection\EnvVarProcessorInterface;
use Symfony\Component\DependencyInjection\Exception\RuntimeException;

final class NullableStringEnvProcessor implements EnvVarProcessorInterface
{
/**
     * {@inheritDoc}
     */
    public function getEnv(string $prefix, string $name, \Closure $getEnv)
    {
        if ($prefix === 'nullablestring') {
            $val = $getEnv($name);

            if ($val === null) {
                return null;
            }

            return (string) $val;
        }

        throw new RuntimeException(\sprintf('Unsupported env var prefix "%s" for env name "%s".', $prefix, $name));
    }

    /**
     * {@inheritDoc}
     */
    public static function getProvidedTypes()
    {
        return [
            'nullablestring' => 'string',
        ];
    }
}
```

fancyweb:
Could you open a new issue with a reproducer and a full stack trace plz?

src/Symfony/Component/DependencyInjection/Container.php
```diff
@@ -390,7 +390,15 @@ protected function getEnv(string $name)
             $prefix = 'string';
             $localName = $name;
         }
-        $processor = $processors->has($prefix) ? $processors->get($prefix) : new EnvVarProcessor($this);
+
+        if ($processors->has($prefix)) {
```
bendavies, line 394: this doesn't work  as `$prefix` is always set to `string` on 390.

bendavies:
confirmed that this fix doesn't work unfortunately, empty stings instead of nulls are still passed to services.
problem noted above.

fancyweb:
I tested it and indeed it doesn't work in the full framework context. I'll fix it again, for the last time I hope :sweat:

PR #50842, opened 2023-07-01
Add missing return types to magic methods

wouterj:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Tickets       | -
| License       | MIT
| Doc PR        | -

I promise, these 72 methods really are the last methods that did not have a return type already :upside_down_face:

### DIFF
--- a/src/Symfony/Bridge/Monolog/Handler/ElasticsearchLogstashHandler.php
+++ b/src/Symfony/Bridge/Monolog/Handler/ElasticsearchLogstashHandler.php
@@ -154,6 +154,9 @@ class ElasticsearchLogstashHandler extends AbstractHandler
         throw new \BadMethodCallException('Cannot serialize '.__CLASS__);
     }
 
+    /**
+     * @return void
+     */
     public function __wakeup()
     {
         throw new \BadMethodCallException('Cannot unserialize '.__CLASS__);
--- a/.github/expected-missing-return-types.diff
+++ b/.github/expected-missing-return-types.diff
@@ -352,6 +352,17 @@ index 57a4c1c2b7..2fb70d7774 100644
 +    public function onTerminate(ConsoleTerminateEvent $event): void
      {
          $this->close();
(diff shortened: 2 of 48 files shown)


stof (approved):

src/Symfony/Bridge/Monolog/Handler/ElasticsearchLogstashHandler.php
```diff
@@ -149,11 +149,17 @@ private function sendToElasticsearch(array $records): void
         $this->wait(false);
     }
 
+    /**
+     * @return never
```
nicolas-grekas: I'm not a fan of the added `never` types to sleep+wakeup. To me, this is motivated by the internal implementation, but the implementation shouldn't drive the contract. Since these methods are magic, they're not supposed to be called by any userland code directly. This more tailored type doesn't serve any purpose. Actually, it only prevents potential child classes from providing their own implementation if they'd like to. I don't have a use case in mind, but that's not a reason to close the API here to me.

wouterj: I'm also a bit hesitant to add `never` as a return type, as it blocks the user from implementing any other behavior.

But in this case, I added them because all these sleep+wakeup calls are `never` to prevent insecure deserialization. I thought it would be a good idea to actually prevent user-code from loosing this requirement, as that could unknowingly make them vulnerable for this type of attacks.

If you still disagree, I'm open to reverting these to `array` and `void` again.

nicolas-grekas: Sure and I understand that. That's on the edge of the open/closed principle, just a bit too much on the closed side to me :)

nicolas-grekas: I realized that the security argument doesn't stand: implementing `__serialize` still allows working around `never`; this adds to "open" side.

nicolas-grekas:
Thank you @wouterj.

PR #50844, opened 2023-07-01
Convert to UTF-8

DavidPrevot:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT


nicolas-grekas:
This non-utf8 class is on purpose. It increases the chance this can't collide. Also, changing this would break existing pools. 👎 Thus.

PR #50846, opened 2023-07-02
[Bridges][Bundles] Convert to native return types

wouterj:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

The first batch of many. This is I think the easiest one: bundles and bridges contain classes that almost never get extended by third party packages. As such, I've converted all types (including void) to native ones.

For the components, we should probably split up non-void types (easy merges) and void types (need close reviews).


wouterj:
For reference, commands to automate 90% of these changes:

```bash
# convert types
sed -i 's/ *"\*\*\/Tests\/"//' composer.json
composer u -o
SYMFONY_PATCH_TYPE_DECLARATIONS='force=2&php=8.2' php .github/patch-types.php
git checkout composer.json

# cleanup
git diff-tree --no-commit-id --name-only HEAD -r | xargs -L1 ./tools/php-cs-fixer/vendor/bin/php-cs-fixer fix --rules phpdoc_trim_consecutive_blank_line_separation,no_superfluous_phpdoc_tags,ordered_imports,no_empty_phpdoc,no_whitespace_in_blank_line,no_extra_blank_lines,phpdoc_trim_consecutive_blank_line_separation
git diff-tree --no-commit-id --name-only HEAD -r | xargs -L1 sed -i ':a;N;$!ba;s/{\n\n/{\n/'
```

nicolas-grekas:
Thank you @wouterj.

PR #50849, opened 2023-07-02
[WebProfilerBundle] Fix minify test after JS refactor

wouterj:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

This test makes sure the JS files are valid after we "minify" them by removing newlines. After the refactor of #50790, it was not updated correctly causing a red CI.

This fixes the test, and makes it more future proof.

### DIFF
--- a/src/Symfony/Bundle/WebProfilerBundle/Tests/Resources/MinifyTest.php
+++ b/src/Symfony/Bundle/WebProfilerBundle/Tests/Resources/MinifyTest.php
@@ -23,8 +23,12 @@ class MinifyTest extends TestCase
     public function testNoSingleLineComments()
     {
         $dir = \dirname(__DIR__, 2).'/Resources/views/Profiler';
-        $message = 'There cannot be any single line comment in this file. Consider using multiple line comment. ';
-        $this->assertTrue(2 === substr_count(file_get_contents($dir.'/base_js.html.twig'), '//'), $message);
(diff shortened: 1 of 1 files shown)


src/Symfony/Bundle/WebProfilerBundle/Tests/Resources/MinifyTest.php
```diff
@@ -23,8 +23,12 @@ class MinifyTest extends TestCase
     public function testNoSingleLineComments()
     {
         $dir = \dirname(__DIR__, 2).'/Resources/views/Profiler';
-        $message = 'There cannot be any single line comment in this file. Consider using multiple line comment. ';
-        $this->assertTrue(2 === substr_count(file_get_contents($dir.'/base_js.html.twig'), '//'), $message);
-        $this->assertTrue(0 === substr_count(file_get_contents($dir.'/toolbar.css.twig'), '//'), $message);
```
stof, line 28: what about this CSS files ?

wouterj: I don't think this applies, `//` is not a valid CSS comment and CSS only supports the `/* ... */` comment syntax.

nicolas-grekas:
Thank you @wouterj.

PR #38478, opened 2020-10-08
[Serializer] Serialization versioning

ArnaudTarroux:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #30848
| License       | MIT
| Doc PR        | n/a

Add Serialization versioning.

- Added `@Since` annotation
- Added `@Until` annotation
- Added support of `since` and `until` to the `xml` and `yaml` mapping
- Added `version` context option
- Added the possibility to configure the `serializer.default_version` in the framework config

PR #35617, opened 2020-02-06
[Serializer] Serialization versioning

ArnaudTarroux:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #30848
| License       | MIT
| Doc PR        | n/a

Add Serialization versioning.

- Added `@Since` annotation
- Added `@Until` annotation
- Added support of `since` and `until` to the `xml` and `yaml` mapping
- Added `version` context option
- Added the possibility to configure the `serializer.default_version` in the framework config

PR #50850, opened 2023-07-02
[Serializer] Add versionning

### LINKED ISSUE #30848: [Serializer] Serialization versioning
**Description**  
Sometimes we need to have a version control on the properties which are serialized.
Currently the Symfony Serializer don't provide a simple and scalable way to do it.

**Example**  
As JmsSerializer do, it would be great to provide two annotations `@Since` and `@Until` who require the version as value (cf. example below):

    class User
    {
	    /**
	     * @Since("1.1.0")
	     */
	     private $email;
	     
	     /**
	     * @Until("1.0.x")
	     */
	     private $username;
     }

Of course xml and yaml mappings need to support it too.
The version to compare to might be passed via the serializer context
`$serializer->serialize($myObject, 'json', ['version' => '1.1.0']);`

The version numbers must be in the standardized PHP version format to make it compatible with the `version_compare` function ([https://www.php.net/manual/en/function.version-compare.php](https://www.php.net/manual/en/function.version-compare.php))

Example of results:

| Version  | Since  | Until  | Serialized  |
|---|---|---|---|
| 1.0.0  | 0.9  | ~  | ✓  |
| 1.9  | 2.0  | ~  |  ✗ |
|  2.3 | 1.3  | 2.1  | ✗  |
|  2.3 | ~ | 2.1  | ✗  |
| 2.4  | 1.1  | 3.x  | ✓  |

If this feature make sense, I can make a PR 🙂.

micoli:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #30848
| License       | MIT

- Add `version` annotation/attribute to be set on the property holding the version of the normalized representation (only one property can be set as `Version`)
- Add `VersionConstraint` annotation/attribute. That attribute control restriction on normalisation/denormalization depending on the value of the property that hold `Version` attribute. Inner properties are `since` and `until`. At least one of them has to be set.
- Either `Version.version`, `VersionConstraint.since` and `VersionConstraint.until` are expressed as semver string, and must be compliant with [version_compare](https://www.php.net/manual/fr/function.version-compare.php) function. A property restricted by a constraint is defined as : `notDefined` < `since|null` <= `defined` <= `until|null` < `notDefined`

Behavior and naming are inspired by https://jmsyst.com/libs/serializer/master/reference/annotations#since and https://jmsyst.com/libs/serializer/master/reference/annotations#until

Logic is as follow:
 - Normalization
    A restricted property by a `VersionConstraint` is not added to the normalized representation if the constraint is not satisfied 
 - Denormalization
    A restricted property by a `VersionConstraint` is set to null if the constraint is not satisfied or to the value if the constraint is satisfied

### Example of usage
```php
class VersionAwareObject
{
    private $foo;

    /**
     * @Version
     */
    public string $version;

    /**
     * @VersionConstraint(
     *     since="1.1",
     *     until="1.5"
     * )
     */
    public ?string $name;

    /**
     * @VersionConstraint(
     *     since="1.6"
     * )
     */
    public ?string $lastName;
}
```

I don't see any Breaking Changes up to now

### DIFF
--- a/src/Symfony/Component/Serializer/Annotation/Version.php
+++ b/src/Symfony/Component/Serializer/Annotation/Version.php
@@ -1,0 +1,31 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Serializer\Annotation;
+
+/**
+ * Annotation class for @Version().
+ *
+ * @Annotation
+ *
+ * @NamedArgumentConstructor
+ *
+ * @Target({"PROPERTY", "METHOD"})
+ *
+ * @author Olivier Michaud <olivier@micoli.org>
+ */
+#[\Attribute(\Attribute::TARGET_METHOD | \Attribute::TARGET_PROPERTY)]
+final class Version
+{
+    public function __construct()
+    {
+    }
+}
--- a/src/Symfony/Component/Serializer/Annotation/VersionConstraint.php
+++ b/src/Symfony/Component/Serializer/Annotation/VersionConstraint.php
@@ -1,0 +1,58 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Serializer\Annotation;
+
+use Symfony\Component\Serializer\Exception\InvalidArgumentException;
+
+/**
+ * Annotation class for @VersionConstraint().
+ *
+ * @Annotation
+ *
+ * @NamedArgumentConstructor
+ *
+ * @Target({"PROPERTY", "METHOD"})
+ *
+ * @author Olivier Michaud <olivier@micoli.org>
+ */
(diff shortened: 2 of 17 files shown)


antonkomarev:
How to deal with cases when in the future version the field should become not nullable or it's type should be changed?

asispts:
Similar PR https://github.com/symfony/symfony/pull/38478

micoli:
> Similar PR #38478

I'v only seen https://github.com/symfony/symfony/pull/35617, not this one. Hence this PR and mine have the same purpose, I can close mine if it's desired

micoli:
> How to deal with cases when in the future version the field should become not nullable or it's type should be changed?

Indeed, to have versioned field, due to the fact it can once be null, then by design those versionConstrained fields should be defined as nullable. I can add a check on that when reading field metadatas

micoli:
> > How to deal with cases when in the future version the field should become not nullable or it's type should be changed?

I've added a check to enforce the fact the field should be, by design, either without typehint either with a nullable type.

Regarding the type change of the field, i've no clue how to enforce it. Do you have an idea?

94noni:
hello (just passing by)
is it relevant to add annotation as the PR target v6 with [php v8.1](https://github.com/symfony/symfony/blob/61786647f3a22e78577a7e2193400a92314a0e2e/composer.json#L36) as minimal requirement?
perhaps only native attribute is enough

nicolas-grekas:
Thanks for this proposal. It is built on the Serializer's `Annotation` namespace and `AnnotationLoader`, both removed in Symfony 7.0, so the patch no longer applies to any maintained branch. The very similar #38478 was also closed last year without the feature being retained. Closing for both reasons.

PR #50852, opened 2023-07-02
[Components] Convert to native return types

wouterj:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -


nicolas-grekas:
rebase unlocked + some fixes needed I guess

nicolas-grekas (approved):
I finished the PR.

I order to help third party bundles to be compatible with Symfony 7, I didn't add the `void` return type to the following interfaces (all listed in `.github/expected-missing-return-types.diff`):
- src/Symfony/Component/DependencyInjection/Compiler/CompilerPassInterface.php
- src/Symfony/Component/DependencyInjection/Extension/PrependExtensionInterface.php
- src/Symfony/Component/HttpKernel/Bundle/BundleInterface.php

Please open an issue or send a PR if you think we should add more interfaces to this list.

Note that src/Symfony/Contracts/Service/ResetInterface.php also doesn't have a native type, and that's on purpose, to keep compatibility with implementations that do have a return value.

I also patched DebugClassLoader so that it doesn't skip `void` return types.

nicolas-grekas:
Thank you @wouterj.

PR #50853, opened 2023-07-03
[Security] `random_bytes` length should be an int greater than 7

asispts:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | 
| License       | MIT
| Doc PR        |

`random_bytes` can only accept an `int<1, max>` value.
- See: https://3v4l.org/tHf9P
- See: https://3v4l.org/u57Ol

This PR ensures that the value passed to `random_bytes` is an integer greater than 0.

### DIFF
--- a/src/Symfony/Component/Security/Csrf/Tests/TokenGenerator/UriSafeTokenGeneratorTest.php
+++ b/src/Symfony/Component/Security/Csrf/Tests/TokenGenerator/UriSafeTokenGeneratorTest.php
@@ -57,4 +57,28 @@ class UriSafeTokenGeneratorTest extends TestCase
         $this->assertStringNotMatchesFormat('%S/%S', $token, 'is URI safe');
         $this->assertStringNotMatchesFormat('%S=%S', $token, 'is URI safe');
     }
+
+    /**
+     * @dataProvider validDataProvider
+     */
+    public function testValidLength(int $entropy, int $length)
+    {
+        $generator = new UriSafeTokenGenerator($entropy);
+        $token = $generator->generateToken();
+        $this->assertSame($length, \strlen($token));
+    }
+
+    public static function validDataProvider(): \Iterator
+    {
+        yield [24, 4];
+        yield 'Float length' => [20, 3];
+    }
+
+    public function testInvalidLength()
+    {
+        $this->expectException(\InvalidArgumentException::class);
+        $this->expectExceptionMessage('Entropy should be greater than 7.');
+
+        new UriSafeTokenGenerator(7);
+    }
 }
--- a/src/Symfony/Component/Security/Csrf/TokenGenerator/UriSafeTokenGenerator.php
+++ b/src/Symfony/Component/Security/Csrf/TokenGenerator/UriSafeTokenGenerator.php
@@ -27,6 +27,10 @@ class UriSafeTokenGenerator implements TokenGeneratorInterface
      */
     public function __construct(int $entropy = 256)
     {
+        if ($entropy <= 7) {
+            throw new \InvalidArgumentException('Entropy should be greater than 7.');
+        }
+
         $this->entropy = $entropy;
     }
 
@@ -38,7 +42,7 @@ class UriSafeTokenGenerator implements TokenGeneratorInterface
         // Generate an URI safe base64 encoded string that does not contain "+",
         // "/" or "=" which need to be URL encoded and make URLs unnecessarily
         // longer.
-        $bytes = random_bytes($this->entropy / 8);
+        $bytes = random_bytes(intdiv($this->entropy, 8));
 
         return rtrim(strtr(base64_encode($bytes), '+/', '-_'), '=');
     }


nicolas-grekas:
What about throwing an InvalidArgumentException instead? I don't think we need to support negative number, they're invalid here.

src/Symfony/Component/Security/Csrf/TokenGenerator/UriSafeTokenGenerator.php
```diff
@@ -38,7 +38,8 @@ public function generateToken()
         // Generate an URI safe base64 encoded string that does not contain "+",
         // "/" or "=" which need to be URL encoded and make URLs unnecessarily
         // longer.
-        $bytes = random_bytes($this->entropy / 8);
+        $length = max(1, abs((int) ($this->entropy / 8)));
```
maxbeckers: can entropy really be negative? I guess by definition it's non-negative. And is it correct then to use the absolute value? Or was your idea to make it positive when any code accidently adds a negative value? For now you would get an php error and now the code would run and use the absolute value, is this correct/expected?
Or maybe validate in the constructor for a positive or non-negative value?

asispts: Agreed. We should throw an exception here. The idea is just to ensure that any code that uses this class is testable, either by throwing an exception or converting the invalid input.

asispts:
Done. What should we do with the float `$length`? Do we need to cast it to an int?

src/Symfony/Component/Security/Csrf/TokenGenerator/UriSafeTokenGenerator.php
```diff
@@ -38,7 +38,16 @@ public function generateToken()
...
         // "/" or "=" which need to be URL encoded and make URLs unnecessarily
         // longer.
-        $bytes = random_bytes($this->entropy / 8);
+        $length = $this->entropy / 8;
+        if (!\is_int($length)) {
+            throw new \InvalidArgumentException(sprintf('Length should be an int, "%s" given.', \gettype($length)));
+        }
+
+        if ($length <= 0) {
+            throw new \InvalidArgumentException('Length should be greater than 0.');
```
nicolas-grekas: length is not something the end user knows about. Entropy is. This should be checked in the constructor also.

src/Symfony/Component/Security/Csrf/TokenGenerator/UriSafeTokenGenerator.php
```diff
@@ -38,7 +38,16 @@ public function generateToken()
...
+        }
+
+        $bytes = random_bytes($length);
```
nicolas-grekas:
```suggestion
        $bytes = random_bytes(intdiv($entropy, 8));
```

src/Symfony/Component/Security/Csrf/Tests/TokenGenerator/UriSafeTokenGeneratorTest.php
```diff
@@ -19,42 +19,46 @@
...
+    public function testInvalidLength()
     {
-        $token = $this->generator->generateToken();
+        $this->expectException(\InvalidArgumentException::class);
+        $this->expectExceptionMessage('Entropy should be greater than 0.');
 
-        $this->assertTrue(ctype_print($token), 'is printable');
-        $this->assertStringNotMatchesFormat('%S+%S', $token, 'is URI safe');
-        $this->assertStringNotMatchesFormat('%S/%S', $token, 'is URI safe');
-        $this->assertStringNotMatchesFormat('%S=%S', $token, 'is URI safe');
```
nicolas-grekas, line 58: why removing these test cases? can you please minimize the diff on this file to ease the review, or explain the changes otherwise?

src/Symfony/Component/Security/Csrf/TokenGenerator/UriSafeTokenGenerator.php
```diff
@@ -27,6 +27,10 @@ class UriSafeTokenGenerator implements TokenGeneratorInterface
      */
     public function __construct(int $entropy = 256)
     {
+        if ($entropy <= 0) {
+            throw new \InvalidArgumentException('Entropy should be greater than 0.');
```
derrabus: It should be greater than 7, actually. ☝🏻🤓 

https://3v4l.org/KLgeS

src/Symfony/Component/Security/Csrf/Tests/TokenGenerator/UriSafeTokenGeneratorTest.php
asispts: Actually, I didn't delete the test; I just modified it. The code that I removed was the code below. `$bytes` is not used, and I thought the remaining code was unnecessary because it's just a simple test. The diff should be fine now.

```php
private const ENTROPY = 1000;

/**
 * A non alpha-numeric byte string.
 *
 * @var string
 */
private static $bytes;

/**
 * @var UriSafeTokenGenerator
 */
private $generator;

public static function setUpBeforeClass(): void
{
    self::$bytes = base64_decode('aMf+Tct/RLn2WQ==');
}

{
    $this->generator = new UriSafeTokenGenerator(self::ENTROPY);
}

protected function tearDown(): void
{
    $this->generator = null;
}
```

src/Symfony/Component/Security/Csrf/TokenGenerator/UriSafeTokenGenerator.php
asispts: Done :smile:

stof (approved):

nicolas-grekas:
Thank you @asispts.

PR #50861, opened 2023-07-03
[Mailer] Open up visibility of `MailjetApiTransport::getPayload`

ostrolucky:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | 
| License       | MIT
| Doc PR        | 

Normally we use Mailer interface, but on few places we are using Mailjet API's bulk functionality. The way it works is pretty much to send `array{messages: list<array>}` via HTTP Client, so list of whatever output `getPayload` function here returns. However, since this is a private method, we had to copy almost whole transport. I think it would be good for users if they can reuse this transport for other use cases.

### DIFF
--- a/src/Symfony/Component/Mailer/Bridge/Mailjet/Transport/MailjetApiTransport.php
+++ b/src/Symfony/Component/Mailer/Bridge/Mailjet/Transport/MailjetApiTransport.php
@@ -104,7 +104,7 @@ class MailjetApiTransport extends AbstractApiTransport
         return $response;
     }
 
-    private function getPayload(Email $email, Envelope $envelope): array
+    public function getPayload(Email $email, Envelope $envelope): array
     {
         $html = $email->getHtmlBody();
         if (null !== $html && \is_resource($html)) {


stof:
I'm -1 on that. This would prevent us to perform any refactoring in that class. This class is not meant to be a generic API client for the Mailjet API.

ostrolucky:
What refactoring? doSendApi is protected and has pretty much same signature. If you ever need to change signature, you would have to change both methods.

src/Symfony/Component/Mailer/Bridge/Mailjet/Transport/MailjetApiTransport.php
```diff
@@ -104,7 +104,7 @@ protected function doSendApi(SentMessage $sentMessage, Email $email, Envelope $e
         return $response;
     }
 
-    private function getPayload(Email $email, Envelope $envelope): array
+    public function getPayload(Email $email, Envelope $envelope): array
     {
         $html = $email->getHtmlBody();
         if (null !== $html && \is_resource($html)) {
```
malekbig, line 110: Does a public method come after a protected method? Could you move it before the doSendApi method?

ostrolucky: Fabpot is not complaining. I'll keep it as it is for easier diff until maintainers ask this to be moved.

fabpot:
I tend to be :-1: on this PR.
The `getPayload()` method is an implementation detail, not part of the public interface we want to expose.
If we go down this route, we should add an interface and ensure we expose the payload for all other transports (most of them have such a payload method). But then again, we don't want to support this internal method (and we cannot guarantee that upstream won't change the format anyway).
So, if we want to support your use case, we should need to think about how to support it properly.

ostrolucky:
I wanted to give you guys _less_ stuff to maintain by opening up only one method in one transport. Would you really prefer to expose this info to public in all the API transports (and having to maintain BC in all of them)? I can do that, just confirm once more please before I do that.

fabpot:
To be clear, my preference would be to reject this PR/new feature.

PR #50877, opened 2023-07-04
[Finder] Add early directory prunning filter support

mvorisek:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | symfony/symfony-docs#18499

Filtering terminology first:

- "exclude" - whole filesystem subtree is traversed, directories (and files) are excluded after, slow
- "prune" - pruned directories are not traversed, fast

Currently there is early directory prunning support possible, but with string patterns only, not even full regex is supported as the input is always quoted - https://github.com/symfony/symfony/blob/v6.2.12/src/Symfony/Component/Finder/Iterator/ExcludeDirectoryFilterIterator.php#L45.

This PR adds early directory prunning with full callback support.

Tested with VFS to assert the pruned directories are really not traversed and the count of IO syscalls is minimal.

### DIFF
--- a/src/Symfony/Component/Finder/Finder.php
+++ b/src/Symfony/Component/Finder/Finder.php
@@ -50,6 +50,7 @@ class Finder implements \IteratorAggregate, \Countable
     private array $notNames = [];
     private array $exclude = [];
     private array $filters = [];
+    private array $pruneFilters = [];
     private array $depths = [];
     private array $sizes = [];
     private bool $followLinks = false;
@@ -580,13 +581,21 @@ class Finder implements \IteratorAggregate, \Countable
      * The anonymous function receives a \SplFileInfo and must return false
      * to remove files.
      *
+     * @param \Closure(SplFileInfo): bool $closure
+     * @param bool                        $prune   Whether to skip traversing directories further
+     *
      * @return $this
      *
      * @see CustomFilterIterator
      */
-    public function filter(\Closure $closure): static
+    public function filter(\Closure $closure /* , bool $prune = false */): static
     {
+        $prune = 1 < \func_num_args() ? func_get_arg(1) : false;
         $this->filters[] = $closure;
+
+        if ($prune) {
+            $this->pruneFilters[] = $closure;
+        }
 
         return $this;
     }
@@ -740,6 +749,10 @@ class Finder implements \IteratorAggregate, \Countable
     {
         $exclude = $this->exclude;
         $notPaths = $this->notPaths;
+
+        if ($this->pruneFilters) {
+            $exclude = array_merge($exclude, $this->pruneFilters);
+        }
 
         if (static::IGNORE_VCS_FILES === (static::IGNORE_VCS_FILES & $this->ignore)) {
             $exclude = array_merge($exclude, self::$vcsPatterns);
--- a/src/Symfony/Component/Finder/Iterator/ExcludeDirectoryFilterIterator.php
+++ b/src/Symfony/Component/Finder/Iterator/ExcludeDirectoryFilterIterator.php
@@ -27,12 +27,15 @@ class ExcludeDirectoryFilterIterator extends \FilterIterator implements \Recursi
     /** @var \Iterator<string, SplFileInfo> */
     private \Iterator $iterator;
     private bool $isRecursive;
+    /** @var array<string, true> */
     private array $excludedDirs = [];
     private ?string $excludedPattern = null;
+    /** @var list<callable(SplFileInfo):bool> */
+    private array $pruneFilters = [];
 
     /**
-     * @param \Iterator<string, SplFileInfo> $iterator    The Iterator to filter
-     * @param string[]                       $directories An array of directories to exclude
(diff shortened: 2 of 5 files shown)


src/Symfony/Component/Finder/Iterator/ExcludeDirectoryFilterIterator.php
```diff
@@ -70,6 +83,14 @@ public function accept(): bool
             return !preg_match($this->excludedPattern, $path);
         }
 
+        if ($this->isRecursive && $this->isDir()) {
```
mvorisek, line 86: psalm is wong here, see `$this->isDir()` used a few lines above, the code is correct

src/Symfony/Component/Finder/Iterator/ExcludeDirectoryFilterIterator.php
```diff
@@ -59,8 +72,13 @@ public function __construct(\Iterator $iterator, array $directories)
...
     public function accept(): bool
     {
-        if (isset($this->excludedDirs[$this->getFilename()]) && $this->hasChildren()) {
-            return false;
+        $hasChildren = null;
+        if (isset($this->excludedDirs[$this->getFilename()])) {
+            $hasChildren = $this->hasChildren();
+
+            if ($hasChildren) {
+                return false;
```
nicolas-grekas:
```suggestion
        if (isset($this->excludedDirs[$this->getFilename()]) && $hasChildren = $this->hasChildren()) {
            return false;
```

src/Symfony/Component/Finder/Iterator/ExcludeDirectoryFilterIterator.php
```diff
@@ -70,6 +88,20 @@ public function accept(): bool
...
+                $hasChildren = $this->hasChildren();
+            }
+
+            if ($hasChildren) {
+                foreach ($this->pruneFilters as $pruneFilter) {
+                    if (!$pruneFilter($this->current())) {
+                        return false;
+                    }
+                }
+            }
```
nicolas-grekas:
```suggestion
        if ($this->pruneFilters && $hasChildren ??= $this->hasChildren()) {
            foreach ($this->pruneFilters as $pruneFilter) {
                if (!$pruneFilter($this->current())) {
                    return false;
                }
            }
```

src/Symfony/Component/Finder/Finder.php
```diff
@@ -580,14 +581,22 @@ public function sortByModifiedTime(): static
      * The anonymous function receives a \SplFileInfo and must return false
      * to remove files.
      *
+     * When prune is enabled, directories are not traversed further.
+     *
+     * @param \Closure(SplFileInfo): bool $closure
```
nicolas-grekas:
```suggestion
     * @param \Closure(SplFileInfo): bool $closure
     * @param bool $prune Whether to traverse directories further
``````

File: src/Symfony/Component/Finder/Finder.php [hunk 5]
```diff
@@ -580,14 +581,22 @@ public function sortByModifiedTime(): static
...
+     *
      * @return $this
      *
      * @see CustomFilterIterator
      */
-    public function filter(\Closure $closure): static
+    public function filter(\Closure $closure, bool $prune = false): static
     {
```
nicolas-grekas:
```suggestion
    public function filter(\Closure $closure/* , bool $prune = false */): static
    {
        $prune = \func_num_args >= 2 ? \func_get_arg(1) : false;
```

nicolas-grekas (commented):
Can you expand a bit on the PR description to say what pruning is about in the context of the Finder component? It would help ppl get what this is about more quickly I think.

src/Symfony/Component/Finder/Finder.php
```diff
@@ -580,14 +581,23 @@ public function sortByModifiedTime(): static
      * The anonymous function receives a \SplFileInfo and must return false
      * to remove files.
      *
+     * @param \Closure(SplFileInfo): bool $closure
+     *
+     * @Symfony7param bool $prune Whether to traverse directories further
```
nicolas-grekas:
```suggestion
     * @param bool $prune Whether to traverse directories further
```

nicolas-grekas (approved):
LGTM, please just use a regular `@param` annotation for the added argument.

src/Symfony/Component/Finder/Finder.php
mvorisek: CS Fixer will force it to remove as not declared

nicolas-grekas: we accept false positives, like this one is

mvorisek:
@nicolas-grekas this PR is here for a while, can it be merged?

fabpot:
@mvorisek Can you rebase to resolve the conflicts?

nicolas-grekas:
Thank you @mvorisek.

PR #50879, opened 2023-07-04
[Notifier] support local development for sns by adding sslmode option

Ferror:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Tickets       | none
| License       | MIT

When developing locally with AWS -> localstack we have to communicate via HTTP similarly to the SQS Transport for Messenger component.

https://github.com/symfony/amazon-sqs-messenger/blob/6.3/Transport/Connection.php#L50C14-L50C14

### DIFF
--- a/src/Symfony/Component/Notifier/Bridge/AmazonSns/AmazonSnsTransportFactory.php
+++ b/src/Symfony/Component/Notifier/Bridge/AmazonSns/AmazonSnsTransportFactory.php
@@ -31,8 +31,10 @@ final class AmazonSnsTransportFactory extends AbstractTransportFactory
 
         $host = 'default' === $dsn->getHost() ? null : $dsn->getHost();
         $port = $dsn->getPort();
+        $sslmode = $dsn->getOption('sslmode', 'enable');
+        $protocol = $sslmode === 'disable' ? 'http' : 'https';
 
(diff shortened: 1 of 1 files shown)


src/Symfony/Component/Notifier/Bridge/AmazonSns/AmazonSnsTransportFactory.php
```diff
@@ -31,8 +31,10 @@ public function create(Dsn $dsn): AmazonSnsTransport
 
         $host = 'default' === $dsn->getHost() ? null : $dsn->getHost();
         $port = $dsn->getPort();
+        $sslmode = $dsn->getOption('sslmode', true);
```
fabpot: Can we be consistent with SQS? It seems that we are using `disable` there to disable SSL.

fabpot (changes requested):

src/Symfony/Component/Notifier/Bridge/AmazonSns/AmazonSnsTransportFactory.php
Ferror: I noticed it has only disable/enable values which seems like a great usage of a simple boolean. Yea, let's keep it consistent.

fabpot (approved):

fabpot:
Thank you @Ferror.

Issue #50871, opened 2023-07-04
UrlType and Url constraint changes


javiereguiluz:
Recently I used `UrlType` and `Url` constraint and faced some issues:

-----

**(1)** The default value of the `default_protocol` option is `http`

This is problematic because:

**(1.1)** `http` is too legacy; it should be `https`

**(1.2)** If this option is different from `null`, the form field is rendered as `<input type="text" inputmode="url">` instead of `<input type="url">` So, the default behavior is not what you expect. Should we change the default value to `null`?

-----

**(2)** Minor but, first I used `https://` as the value of `default_protocol` (instead of `https`, which is the right value). This means that if you input `foo`, the value is updated as `https://://foo` instead of just `https://foo`

If this happened to others too, maybe we could validate the values passed to this option?

-----

**(3)** The `Url` constraint considers something like `aaa` as a valid URL. I guess this is correct according to some RFC, but it's not what most people would like to accept as a valid URL.

Should we add a config option to require the presence of at least one TLD? Note that I'm not thinking about validating that the TLD is correct, because there are too many TLDs and change often.

Thanks!

stof:
I suggest changing the default value of `default_protocol` to `null` in 7.0 (with a configurable opt-in in 6.4).

Validating that the default protocol does not contain `://` is probably also a good idea (or maybe supporting both modes)

Chris53897:
I guess 3) needs to be this way to allow values like `localhost`. I would assert that this should be allowed

javiereguiluz:
Some of the issues reported here were solved. But the last issue remains: something like `aaa` or `foobar` is considered a valid URL.

As pointed by @Chris53897 maybe it's enough to only support `localhost` and then require other values to contain at least one `.` to ensure that looks like a URL?

wouterj:
Any word is a valid url. You can see this often in local networks, e.g. in a container network you can reference other containers by their name (e.g. with Docker Compose).

I think we can introduce a `requireTld` option, defaulting to `false`, to make it easy for people to validate a string contains a URL with a TLD.

xabbuh:
@wouterj's suggestion makes sense to me. We already have such an option in the `Hostname` constraint.

wouterj:
Oh didn't know we already had such option with the same name on another constraint :smiley: The only issue I see is that an option with the same name has a different default on 2 related constraints (given it's opt-out for `Hostname`).

[...]

PR #50882, opened 2023-07-05
[Form] Deprecate not configuring the `default_protocol` option of the `UrlType`

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Tickets       | Part of #50871 
| License       | MIT
| Doc PR        | N/A

You would expect an `<input type="url">` from the `UrlType`, but it is not possible as long as `default_protocol` has a value, because then you have to accept inputs that are not URLs (and you get an `<input type="text" inputmode="url">`).

In order to change `default_protocol` from `'http'` to `null` in 7.0, we must first deprecate not configuring it.

Don’t know how I can update tests though; all of `UrlTypeTest`’s parent test cases will trigger the deprecation :thinking:

### DIFF
--- a/src/Symfony/Component/VarExporter/LazyGhostTrait.php
+++ b/src/Symfony/Component/VarExporter/LazyGhostTrait.php
@@ -332,6 +332,10 @@ trait LazyGhostTrait
         }
 
         if ((Registry::$parentMethods[self::class] ??= Registry::getParentMethods(self::class))['clone']) {
+            if ($this->lazyObjectState->skippedProperties) {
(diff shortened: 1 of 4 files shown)


yceruto:
Hi @MatTheCat, can you check your commit? it looks unrelated to this PR's proposal.

MatTheCat:
Damn seems I’ve been mixing up my branches :grimacing: Opened #50922

PR #50885, opened 2023-07-05
[Finder] Fix children condition in ExcludeDirectoryFilterIterator

mvorisek:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | no
| License       | MIT
| Doc PR        | no

see https://github.com/php/php-src/blob/6091603b5350c1ca0e792a6da604e25e2a0121c8/ext/spl/spl_directory.h#L113

Original `isDir` follow symlinks, so this PR has no effect, but it avoids expensive `isDir` call and makes the condition more correct.


nicolas-grekas:
Thank you @mvorisek.

PR #49358, opened 2023-02-13
[Routing] Deprecate annotations in favor of attributes

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | Follows #50888
| License       | MIT
| Doc PR        | symfony/symfony-docs#18589

This PR deprecates the integration of Doctrine Annotations for the Routing component.

Attributes are the way to go now if we want to annotate controllers with routing information. Existing applications can be migrated easily using [Rector](https://getrector.com/blog/how-to-upgrade-annotations-to-attributes).

Thus I believe that keeping support for oldschool Doctrine Annotations is not necessary anymore.

If this PR is accepted, I would work on a follow-up that renames all `Annotation*Loader` classes because the names of those classes are probably misleading as soon as they support attributes only.

PR #50888, opened 2023-07-05
[FrameworkBundle] Deprecate doctrine/annotations integration

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | N/A
| License       | MIT
| Doc PR        | TODO

This PR deprecates the integration of the doctrine/annotations package in FrameworkBundle. Currently, the integration is enabled by default if `doctrine/annotations` is installed. This PR adds a runtime deprecation if the app does not explicitly disable it by setting:

```php
framework:
    annotations: false
```

The plan is to make this the default and only valid option in Symfony 7 and remove that setting entirely in Symfony 8. This change unlocks #49358 and similar PRs that remove support for Doctrine Annotations from components.

### DIFF
--- a/UPGRADE-6.4.md
+++ b/UPGRADE-6.4.md
@@ -34,6 +34,8 @@ FrameworkBundle
 ---------------
 
  * Add native return type to `Translator` and to `Application::reset()`
+ * Deprecate the integration of Doctrine annotations, either uninstall the `doctrine/annotations` package or disable
+   the integration by setting `framework.annotations` to `false`
 
 HttpFoundation
 --------------
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
@@ -1716,6 +1716,8 @@ class FrameworkExtension extends Extension
             throw new LogicException('Annotations cannot be enabled as the Doctrine Annotation library is not installed. Try running "composer require doctrine/annotations".');
         }
 
+        trigger_deprecation('symfony/framework-bundle', '6.4', 'Enabling the integration of Doctrine annotations is deprecated. Set the "framework.annotations.enabled" config option to false.');
+
         $loader->load('annotations.php');
 
         if ('none' === $config['cache']) {
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/php/default_config.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/php/default_config.php
@@ -1,4 +1,6 @@
 <?php
 
 $container->loadFromExtension('framework', [
-    'http_method_override' => false,]);
+    'annotations' => false,
+    'http_method_override' => false,
+]);
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/php/serializer_mapping.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/php/serializer_mapping.php
@@ -2,7 +2,7 @@
 
 $container->loadFromExtension('framework', [
     'http_method_override' => false,
-    'annotations' => ['enabled' => true],
+    'annotations' => false,
     'serializer' => [
         'enable_annotations' => true,
         'mapping' => [
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/xml/workflow_with_no_events_to_dispatch.xml
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/xml/workflow_with_no_events_to_dispatch.xml
@@ -7,6 +7,7 @@
                         http://symfony.com/schema/dic/symfony http://symfony.com/schema/dic/symfony/symfony-1.0.xsd">
 
     <framework:config http-method-override="false">
+        <framework:annotations enabled="false" />
         <framework:workflow name="my_workflow" type="state_machine">
             <framework:initial-marking>one</framework:initial-marking>
             <framework:marking-store type="method" property="state" />
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/yml/full.yml
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/yml/full.yml
@@ -45,10 +45,7 @@ framework:
         paths: ['%kernel.project_dir%/Fixtures/translations']
     validation:
         enabled: true
-    annotations:
-        cache:   file
-        debug:   true
-        file_cache_dir: '%kernel.cache_dir%/annotations'
+    annotations: false
     serializer:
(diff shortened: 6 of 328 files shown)


src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -1716,6 +1716,8 @@ private function registerAnnotationsConfiguration(array $config, ContainerBuilde
             throw new LogicException('Annotations cannot be enabled as the Doctrine Annotation library is not installed. Try running "composer require doctrine/annotations".');
         }
 
+        trigger_deprecation('symfony/framework-bundle', '6.4', 'Enabling the integration of Doctrine annotations is deprecated. Set the "framework.annotations" config option to false.');
```
OskarStark:
```suggestion
        trigger_deprecation('symfony/framework-bundle', '6.4', 'Enabling the integration of Doctrine annotations is deprecated. Set the "framework.annotations.enabled" config option to false.');
```

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/php/default_config.php
```diff
@@ -1,4 +1,5 @@
 <?php
 
 $container->loadFromExtension('framework', [
+    'annotations' => false,
     'http_method_override' => false,]);
```
OskarStark:
```suggestion
    'http_method_override' => false,
]);
```

UPGRADE-6.4.md
```diff
@@ -30,6 +30,12 @@ Form
  * Deprecate using `DateTime` or `DateTimeImmutable` model data with a different timezone than configured with the
    `model_timezone` option in `DateType`, `DateTimeType`, and `TimeType`
 
+FrameworkBundle
+---------------
+
+ * Deprecate the integration of Doctrine annotations, either uninstall the `doctrine/annotations` package or disable
```
94noni, line 37: isnt also good to document here that the way to go is native attributes?
it may seem obvious but just asking :)

derrabus: That's certainly what we should do when we deprecate Doctrine annotations for components (Routing and friends).

GromNaN:
I don't have the whole thing in my head. Is it necessary to add the `framework.annotations.enabled: false` config when the `doctrine/annotations` package is not detected?

Self-reply: it's documented in FrameworkBundle's changelog.
Thanks.

derrabus:
> Is it necessary to add the `framework.annotations.enabled: false` config when the `doctrine/annotations` package is not detected?

No.

Long answer: Unfortunately, we currently enable the annotations integration by default if the `doctrine/annotations` package is installed. However if the package is not installed, FrameworkBundle's default behavior is fine, so you don't have to opt out of anything.

stof:
@derrabus regarding the "unlocking" part, shouldn't we actually first deprecate annotations in components and then deprecate the FrameworkBundle integration configuring a reader only once we don't rely on it anymore ?

derrabus:
> shouldn't we actually first deprecate annotations in components and then deprecate the FrameworkBundle integration configuring a reader only once we don't rely on it anymore ?

That was my first approach, but I think it's easier this way. This PR is self-contained: If we merge it and remove the integration in 7.0, the components might still support Doctrine annotations, but FrameworkBundle won't have a way to enable them. Removing a feature that FrameworkBundle doesn't use should be easy. 🙂

stof:
@derrabus but the DX would be worse as devs would not get deprecation warnings telling them where they still use annotations, which they should solve *before* disabling them.
To me, we should first have a component-level deprecation triggered when annotations are actually found (so in the case of the routing component, triggered when you actually have a route defined with annotations).

derrabus:
> @derrabus but the DX would be worse as devs would not get deprecation warnings telling them where they still use annotations, which they should solve _before_ disabling them.

#49358, #50982 and #50983 trigger the deprecations you want.

> To me, we should first have a component-level deprecation triggered when annotations are actually found (so in the case of the routing component, triggered when you actually have a route defined with annotations).

This is kinda a chicken-egg problem. I'm working on removing annotations everywhere and I believe we both agree on how the end result should look like. Unless we want to discuss a big all-or-nothing PR, I think we can live with this DX flaw, knowing we will resolve it before 6.4.

UPGRADE-6.4.md
derrabus: Will be fixed in #49358, #50982, #50983.

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/php/serializer_mapping.php
```diff
@@ -2,7 +2,7 @@
 
 $container->loadFromExtension('framework', [
     'http_method_override' => false,
-    'annotations' => ['enabled' => true],
+    'annotations' => ['enabled' => false],
```
nicolas-grekas:
```suggestion
    'annotations' => false,
```

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/xml/workflow_with_no_events_to_dispatch.xml
```diff
@@ -7,11 +7,12 @@
                         http://symfony.com/schema/dic/symfony http://symfony.com/schema/dic/symfony/symfony-1.0.xsd">
 
     <framework:config http-method-override="false">
+        <framework:annotations enabled="false" />
         <framework:workflow name="my_workflow" type="state_machine">
             <framework:initial-marking>one</framework:initial-marking>
             <framework:marking-store type="method" property="state" />
             <framework:support>Symfony\Bundle\FrameworkBundle\Tests\DependencyInjection\FrameworkExtensionTestCase</framework:support>
-            <framework:event-to-dispatch></framework:event-to-dispatch>
+            <framework:event-to-dispatch />
```
nicolas-grekas: needless change

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/yml/full.yml
```diff
@@ -46,9 +46,7 @@ framework:
     validation:
         enabled: true
     annotations:
-        cache:   file
-        debug:   true
-        file_cache_dir: '%kernel.cache_dir%/annotations'
+        enabled: false
```
nicolas-grekas: annotations: false

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/yml/serializer_mapping.yml
```diff
@@ -1,7 +1,7 @@
 framework:
     http_method_override: false
     annotations:
```
nicolas-grekas: (trying to bring consistency to these)
    annotations: false

nicolas-grekas (approved):

nicolas-grekas:
Thank you @derrabus.

PR #50890, opened 2023-07-05
Add native return types to fix compat with v7

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

### DIFF
--- a/.github/expected-missing-return-types.diff
+++ b/.github/expected-missing-return-types.diff
@@ -375,17 +375,10 @@ index 718be59c13..091f24a8f8 100644
      {
          $this->mailer->send($this->buildMessage($content, $records));
 diff --git a/src/Symfony/Bridge/Monolog/Logger.php b/src/Symfony/Bridge/Monolog/Logger.php
-index 367b3351ff..fb603e74ba 100644
+index 4d4cc8eaf6..2a0a330181 100644
 --- a/src/Symfony/Bridge/Monolog/Logger.php
 +++ b/src/Symfony/Bridge/Monolog/Logger.php
-@@ -44,5 +44,5 @@ class Logger extends BaseLogger implements DebugLoggerInterface, ResetInterface
-      * @return void
-      */
--    public function clear()
-+    public function clear(): void
-     {
-         if ($logger = $this->getDebugLogger()) {
-@@ -63,5 +63,5 @@ class Logger extends BaseLogger implements DebugLoggerInterface, ResetInterface
+@@ -60,5 +60,5 @@ class Logger extends BaseLogger implements DebugLoggerInterface, ResetInterface
       * @return void
       */
 -    public function removeDebugLogger()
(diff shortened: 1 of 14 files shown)


src/Symfony/Component/Form/Extension/Validator/Constraints/FormValidator.php
```diff
@@ -29,10 +29,7 @@ class FormValidator extends ConstraintValidator
      */
     private \SplObjectStorage $resolvedGroups;
 
-    /**
-     * @return void
-     */
-    public function validate(mixed $form, Constraint $formConstraint)
+    public function validate(mixed $form, Constraint $formConstraint): void
```
stof: isn't this a BC break as this class is not final ?

nicolas-grekas: yep, they all are. I'm iterating on the topic to figure out the native types we should revert on v7 (like this one, see 96bec7c640) or the ones we can change on v6 because nobody can reasonably override the patched methods in a child class.

src/Symfony/Bundle/FrameworkBundle/Tests/Kernel/ConcreteMicroKernel.php
```diff
@@ -68,7 +68,7 @@ public function __sleep(): array
         throw new \BadMethodCallException('Cannot serialize '.__CLASS__);
     }
 
-    public function __wakeup()
+    public function __wakeup(): void
```
derrabus:
```suggestion
    public function __wakeup(): never
```

src/Symfony/Bundle/FrameworkBundle/Tests/Kernel/flex-style/src/FlexStyleMicroKernel.php
```diff
@@ -69,7 +69,7 @@ public function __sleep(): array
         throw new \BadMethodCallException('Cannot serialize '.__CLASS__);
     }
 
-    public function __wakeup()
+    public function __wakeup(): void
```
derrabus:
```suggestion
    public function __wakeup(): never
```

nicolas-grekas: not needed, wouldn't provide anything, better be boring.

nicolas-grekas (commented):
All green, PR is ready.

What this does is adding a few return types on some selected classes on 6.4.
Technically, this is a BC break, but I think the patched methods have a very low chance of being overridden in a child class.

Note that while preparing this PR, I reverted some native return types on 7.0 as well, see full list in:
https://github.com/symfony/symfony/blob/7.0/.github/expected-missing-return-types.diff

Alternatively to this PR, we could revert more native return types on 7.0, and we could also declare v6 and v7 as being incompatible for the related packages. This is a matter of finding the sweet spot between those approaches, and this PR is my proposal :)

I'm merging it now to unlock the CI, but feel free to discuss the topic even after merge, we still have a few months to adjust this before the release.

PR #50899, opened 2023-07-06
[VarExporter] Initialize ghosts before calling their `__clone` method if they have skipped properties

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50874
| License       | MIT
| Doc PR        | N/A

When cloning an uninitialized ghost, there is an issue if the parent class’ `__clone` method changes data used by the initializer before it is initialized. This is illustrated in #50874 where unsetting a proxy’s identifier later causes the entity to be hydrated with wrong data.

There is no efficient way to detect such a case, but it can only happen if the parent implements `__clone` and its ghost has skipped properties (entity identifiers e.g.). Initializing the ghost before calling `__clone` then ensure it will be done properly.

### DIFF
--- a/src/Symfony/Component/VarExporter/LazyGhostTrait.php
+++ b/src/Symfony/Component/VarExporter/LazyGhostTrait.php
@@ -332,6 +332,10 @@ trait LazyGhostTrait
         }
 
         if ((Registry::$parentMethods[self::class] ??= Registry::getParentMethods(self::class))['clone']) {
+            if ($this->lazyObjectState->skippedProperties) {
+                $this->initializeLazyObject();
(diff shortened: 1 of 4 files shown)


nicolas-grekas:
I'm closing here because this is something that should be dealt with in the initializer, not in the generic implementation.

Here is the fix I'm proposing upstream: https://github.com/doctrine/orm/pull/10819

I wouldn't have been so quick to figure it out without your help so thanks a lot for investigating this issue! :pray:

PR #50907, opened 2023-07-06
[Validator] Update `Type` constraint, add `number`, `finite-float` and `finite-number` validations

### LINKED ISSUE #50782: [Validator] `Type` constraint considers `NAN` and `INF` as floats

6.3

### Description

If a number is outside of `2.2250738585072014E-308` to `1.7976931348623158E+308` range, PHP converts it to `-INF` or `INF`.

Values `INF`, `-INF` and `NAN` are considered `float` in PHP as for the `is_float` function.

The `Type` constraint relies in this function to check for a valid float.

Now, are this valid float numbers? IMHO no.

FTR: this doesn't happen with integers, as PHP converts them to float once they go outside the `PHP_INT_MIN` and `PHP_INT_MAX` range...

### How to reproduce

```php
var_dump(is_float(INF));
var_dump(is_float(NAN));
```

### Possible Solution

Add an extra check for `is_finite`, when dealing with floats?

### Additional Context

_No response_

guillaume-a:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #50782
| License       | MIT
| Doc PR        | symfony/symfony-docs#18527

Contraint `Type(['float'])` can produce positives with INF or NAN.
This new types car narrow validation and limit validation to finite numbers.

```
#[Assert\Type(['int', 'finite-float'])]
private $value;
```

Thank you for helping me with this one.


src/Symfony/Component/Validator/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Allow single integer for the `versions` option of the `Uuid` constraint
  * Allow single constraint to be passed to the `constraints` option of the `When` constraint
+ * Add `Finite` contraint
```
nicolas-grekas:
```suggestion
 * Add `Finite` constraint
```

guillaume-a: Oh. Thank you for that :)

src/Symfony/Component/Validator/Constraints/Finite.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    public const NOT_FINITE_ERROR = '5f809eb0-78b9-492d-ad37-5a5188390415';
+
+    protected const ERROR_NAMES = [
+        self::NOT_FINITE_ERROR => 'NOT_FINITE_ERROR',
+    ];
+
+    public $message = 'This value should be finite.';
+
+    public function __construct(string $message = null, array $groups = null, mixed $payload = null, array $options = [])
```
stof: to be compatible with the XML and Yaml loaders, the `options` argument must be the first one.

stof (changes requested):
This is missing tests

src/Symfony/Component/Validator/Constraints/Finite.php
guillaume-a: Made the changes, and also found where ValidatorTests were.
Please tell me if I miss anything.

stof (approved):

fabpot:
@guillaume-a Can you fix fabbot errors?

derrabus:
> Contraint `Type(['float'])` can produce false positives.

No, actually that constraint works as advertised because NaN and infinity are valid `float` values.

We should change the PR description.

src/Symfony/Component/Validator/Tests/Constraints/FiniteValidatorTest.php
```diff
@@ -0,0 +1,81 @@
...
+
+        $this->validator->validate(null, $constraint);
+
+        $this->assertNoViolation();
+    }
+
+    /**
+     * @dataProvider getValidValues
+     */
+    public function testValidValues($value)
```
derrabus:
```suggestion
    public function testValidValues(mixed $value)
```

src/Symfony/Component/Validator/Tests/Constraints/FiniteValidatorTest.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        $constraint = new Finite();
+
+        $this->validator->validate($value, $constraint);
+
+        $this->assertNoViolation();
+    }
+
+    public static function getValidValues()
```
derrabus:
```suggestion
    public static function getValidValues(): array
```

src/Symfony/Component/Validator/Tests/Constraints/FiniteValidatorTest.php
```diff
@@ -0,0 +1,81 @@
...
+            [1.5],
+            ['1.5'],
+            ['12345'],
+        ];
+    }
+
+    /**
+     * @dataProvider getInvalidValues
+     */
+    public function testInvalidValues($value, $valueAsString)
```
derrabus:
```suggestion
    public function testInvalidValues(mixed $value, string $valueAsString)
```

src/Symfony/Component/Validator/Tests/Constraints/FiniteValidatorTest.php
```diff
@@ -0,0 +1,81 @@
...
+
+        $this->validator->validate($value, $constraint);
+
+        $this->buildViolation('myMessage')
+            ->setParameter('{{ value }}', $valueAsString)
+            ->setCode(Finite::NOT_FINITE_ERROR)
+            ->assertRaised();
+    }
+
+    public static function getInvalidValues()
```
derrabus:
```suggestion
    public static function getInvalidValues(): array
```

derrabus (commented):
What happens if that validator encounters…
* … a non-numeric string?
* … `null` or an empty string?
* … something completely different lika an array or object?

Can we cover those scenarios with tests?

guillaume-a:
Fixed fabbot errors.
Check against `is_numeric()` in the validator (in case of string, array, object `$value`)
Added tests scenarios for all thoses last checks.

`null` was allready checked (considered as valid)
`empty` is now checked and considered as invalid

src/Symfony/Component/Validator/Constraints/Finite.php
```diff
@@ -0,0 +1,40 @@
...
+#[\Attribute(\Attribute::TARGET_PROPERTY | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
+class Finite extends Constraint
+{
+    public const NOT_FINITE_ERROR = '5f809eb0-78b9-492d-ad37-5a5188390415';
+
+    protected const ERROR_NAMES = [
+        self::NOT_FINITE_ERROR => 'NOT_FINITE_ERROR',
+    ];
+
+    public $message = 'This value should be finite.';
```
javiereguiluz: Just asking if we should make this error message a bit more precise.

`This value should be finite.` -> `This value should be a finite number.`

Maybe we should **not** do this, because the error messages of equivalent constraints are also like this:

`This value should be positive.` (instead of `This value should be a positive number.`)

javiereguiluz (approved):

nicolas-grekas:
What about adding new types to the Type constraint instead?
```php
#[Assert\Type(['finite-float'])]
#[Assert\Type(['number'])] # this would be (is_int || is_float) && is_finite
```

guillaume-a:
Renamed `Finite` to `IsFinite` to match existing validators `IsTrue` `IsNull` ... 

@nicolas-grekas originally I looked that way, but IMO it will add lots of complexity to `Type` validator since `$types` are considered in a global 'OR' condition. And we should then check againts `is_finite()` only in some cases only which may be hard to deterime.

But i'm open to discussion.

nicolas-grekas:
It looks like this would be enough to support my proposal. Not too much complexity IMHO.

```patch
--- a/src/Symfony/Component/Validator/Constraints/TypeValidator.php
+++ b/src/Symfony/Component/Validator/Constraints/TypeValidator.php
@@ -26,9 +26,11 @@ class TypeValidator extends ConstraintValidator
         'int' => 'is_int',
         'integer' => 'is_int',
         'long' => 'is_int',
+        'finite-float' => 'is_float && is_finite',
         'float' => 'is_float',
         'double' => 'is_float',
         'real' => 'is_float',
+        'number' => 'is_int || is_float && is_finite',
         'numeric' => 'is_numeric',
         'string' => 'is_string',
         'scalar' => 'is_scalar',
@@ -69,7 +71,13 @@ class TypeValidator extends ConstraintValidator
 
         foreach ($types as $type) {
             $type = strtolower($type);
-            if (isset(self::VALIDATION_FUNCTIONS[$type]) && self::VALIDATION_FUNCTIONS[$type]($value)) {
+            if (isset(self::VALIDATION_FUNCTIONS[$type]) && match (self::VALIDATION_FUNCTIONS[$type]) {
+                'finite-float' => \is_float($value) && \is_finite($value),
+                'number' => \is_int($value) || \is_float($value) && \is_finite($value),
+                default => self::VALIDATION_FUNCTIONS[$type]($value),
+            }) {
                 return;
             }
```

guillaume-a:
I agree. And I love your code, (always forget about `match` and love to re-discover it again and again)

But do we define `finite` as a type, or as an additional condition like `null`, `true` or `false` ?

(Real question, I'm not trying to push my proposal)

nicolas-grekas:
I'm not sure I'd add `finite`. `finite-float` and `number` look just fine to me. We might want to split `number` in two possibly: `finite-number` (=int|float&finite) and `number` (=int|float).

guillaume-a:
Ok, sounds good to me.
Could we add 

`finite-int` and `finite-float` : to match existing types checking
And a new `number` check with its finite twin `finite-number` ?

nicolas-grekas:
`finite-int` doesn't make sense since an int is always finite. But good to me for the others.

src/Symfony/Component/Validator/Constraints/TypeValidator.php
```diff
@@ -21,35 +21,38 @@
...
-        'xdigit' => 'ctype_xdigit',
+      'bool' => 'is_bool',
+      'boolean' => 'is_bool',
+      'int' => 'is_int',
+      'integer' => 'is_int',
+      'long' => 'is_int',
+      'float' => 'is_float',
+      'double' => 'is_float',
+      'real' => 'is_float',
+      'number' => 'is_int || is_float',
```
nicolas-grekas: please mind the indentation
now I'm wondering: should number be defined as is_int || is_float && !is_nan?

src/Symfony/Component/Validator/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Allow single integer for the `versions` option of the `Uuid` constraint
  * Allow single constraint to be passed to the `constraints` option of the `When` constraint
+ * Add `Finite` constraint
```
nicolas-grekas: missing update

src/Symfony/Component/Validator/Constraints/TypeValidator.php
```diff
@@ -29,6 +29,9 @@ class TypeValidator extends ConstraintValidator
         'float' => 'is_float',
         'double' => 'is_float',
         'real' => 'is_float',
+        'number' => 'is_int || is_float',
```
nicolas-grekas:
```suggestion
        'number' => 'is_int || is_float && !is_nan',
```

src/Symfony/Component/Validator/Constraints/TypeValidator.php
```diff
@@ -69,7 +72,12 @@ public function validate(mixed $value, Constraint $constraint)
 
         foreach ($types as $type) {
             $type = strtolower($type);
-            if (isset(self::VALIDATION_FUNCTIONS[$type]) && self::VALIDATION_FUNCTIONS[$type]($value)) {
+            if (isset(self::VALIDATION_FUNCTIONS[$type]) && match ($type) {
+                'finite-float' => \is_float($value) && is_finite($value),
+                'finite-number' => \is_int($value) || \is_float($value) && is_finite($value),
+                'number' => \is_int($value) || \is_float($value),
```
nicolas-grekas:
```suggestion
                'number' => \is_int($value) || \is_float($value) && !is_nan($value),
```

src/Symfony/Component/Validator/Constraints/TypeValidator.php
guillaume-a: Sorry, my IDE was not configured properly.
It can make sens for `number` to check againt `is_nan` yes.
And since `is_finite` allready checks that, we can do it only for `number` and not `finite-number`

I'll do it tomorrow.
And i'll update the docs PR too.

src/Symfony/Component/Validator/Constraints/TypeValidator.php
ro0NL: for what it's worth, psalm calls it `numeric` and it's a super type of int or float and `numeric-string`

nicolas-grekas: numeric is already a thing for validator, we don't have to follow psalm (we don't already)

number != is-not-a-number

any other opinion on this?

ro0NL: excusi, i missed existing `numeric` below

having both numeric + number is fine :+1: but i'd prefer consistent `finite-*` for all cases

guillaume-a:
Can someone explain me why we must write `\is_float` but we must write `is_nan` or `is_finite` without ?
What are the exact rules ?

nicolas-grekas:
@guillaume-a check https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3048

nicolas-grekas (approved):

nicolas-grekas:
(can you have a look at the test failure?)

src/Symfony/Component/Validator/Constraints/TypeValidator.php
```diff
@@ -29,6 +29,9 @@ class TypeValidator extends ConstraintValidator
         'float' => 'is_float',
         'double' => 'is_float',
         'real' => 'is_float',
+        'number' => 'is_int || is_float && !is_nan',
+        'finite-float' => 'is_float && is_finite',
+        'finite-number' => 'is_int || is_float && is_finite',
         'numeric' => 'is_numeric',
```
ro0NL, line 35: \+ `finite-numeric`?

guillaume-a:
> (can you have a look at the test failure?)

Done. (And rebased too.)

nicolas-grekas (approved):

src/Symfony/Component/Validator/Constraints/TypeValidator.php
ro0NL: omfg, ofcourse, how could i miss that.

fabpot (approved):

fabpot:
Thank you @guillaume-a.

PR #50911, opened 2023-07-07
[HttpKernel] Enhance exception if possible

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | restore an old feature
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

I don't know if it's the best place to do that.
I think in the kernel, it would be better, WDYT?

Before 

image

After : 

image

### DIFF
--- a/src/Symfony/Component/ErrorHandler/ErrorHandler.php
+++ b/src/Symfony/Component/ErrorHandler/ErrorHandler.php
@@ -516,14 +516,7 @@ class ErrorHandler
             }
         }
 
-        if (!$exception instanceof OutOfMemoryError) {
-            foreach ($this->getErrorEnhancers() as $errorEnhancer) {
-                if ($e = $errorEnhancer->enhance($exception)) {
-                    $exception = $e;
-                    break;
-                }
-            }
-        }
+        $exception = $this->enhanceError($exception);
 
         $exceptionHandler = $this->exceptionHandler;
         $this->exceptionHandler = [$this, 'renderException'];
@@ -659,6 +652,21 @@ class ErrorHandler
         }
 
         echo $exception->getAsString();
+    }
+
+    public function enhanceError(\Throwable $exception): \Throwable
+    {
+        if ($exception instanceof OutOfMemoryError) {
+            return $exception;
+        }
+
+        foreach ($this->getErrorEnhancers() as $errorEnhancer) {
+            if ($e = $errorEnhancer->enhance($exception)) {
+                return $e;
+            }
+        }
+
+        return $exception;
     }
 
     /**
--- a/src/Symfony/Component/ErrorHandler/Tests/ErrorHandlerTest.php
+++ b/src/Symfony/Component/ErrorHandler/Tests/ErrorHandlerTest.php
@@ -363,13 +363,14 @@ class ErrorHandlerTest extends TestCase
     /**
      * @dataProvider handleExceptionProvider
      */
-    public function testHandleException(string $expectedMessage, \Throwable $exception)
+    public function testHandleException(string $expectedMessage, \Throwable $exception, string $enhancedMessage = null)
     {
         try {
             $logger = $this->createMock(LoggerInterface::class);
             $handler = ErrorHandler::register();
 
             $logArgCheck = function ($level, $message, $context) use ($expectedMessage, $exception) {
+                $this->assertSame('critical', $level);
                 $this->assertSame($expectedMessage, $message);
                 $this->assertArrayHasKey('exception', $context);
                 $this->assertInstanceOf($exception::class, $context['exception']);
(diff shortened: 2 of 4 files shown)


stof (approved):

src/Symfony/Component/ErrorHandler/ErrorHandler.php
```diff
@@ -664,9 +664,11 @@ private function renderException(\Throwable $exception): void
     /**
      * Override this method if you want to define more error enhancers.
      *
+     * @internal
```
GromNaN: Are there other places where an `@internal` function is used in another package?

yceruto (approved):

src/Symfony/Component/HttpKernel/Controller/ErrorController.php
```diff
@@ -38,6 +39,13 @@ public function __construct(HttpKernelInterface $kernel, string|object|array|nul
 
     public function __invoke(\Throwable $exception): Response
     {
+        foreach (ErrorHandler::getErrorEnhancers() as $errorEnhancer) {
+            if ($e = $errorEnhancer->enhance($exception)) {
+                $exception = $e;
+                break;
+            }
+        }
+
```
yceruto: Before this PR, the error enhancement depends on the `framework.handle_all_throwables` config, if `true` the `Error` instance is rendered through the error-renderer system directly (skipping the error enhancement), otherwise the `ErrorHandler::handleException()` will address it (the error enhancement happens here) and then it will go through the error-renderer system. That's why it's happening. If you change the config to `false` you will see the enhanced error.

Now I'm wondering if it's better to add a new static method inside the ErrorHandler to do this job, thus avoiding duplicated logic (I don't remember the reason but it's currently skipping instances of `OutOfMemoryError`).

Maybe something like this:
```php
// ErrorHandler class
public static enhanceError($error: \Throwable): \Throwable 
{
  if ($exception instanceof OutOfMemoryError) {
     return $error;
  }

  foreach (self::getErrorEnhancers() as $errorEnhancer) {
    if ($error = $errorEnhancer->enhance($exception)) {
      return $error;
    }
  }

  return $error;
}
```
and reuse it in both sides.

yceruto (changes requested):

src/Symfony/Component/HttpKernel/Controller/ErrorController.php
stof: in case of an OutOfMemoryError, enhancing it is likely to break due to needing more memory to compute the enhancements. That's why it is skipped.

lyrixx: I was thinking the same things. I'll move the logic

lyrixx:
I really wonder if the controller is the best place to do that 

* If one use a custom controller, it won't work
* I should also add support for the console components, right?

src/Symfony/Component/HttpKernel/Controller/ErrorController.php
lyrixx: Why is framework.handle_all_throwables false by default?

lyrixx:
I updated the PR

* Introduced `ErrorHandler::enchanceError()`
* Move the call from the ErrorController to the ErrorListener (this is more global like this)

lyrixx:
```
There were 15 errors:

1) Symfony\Component\HttpKernel\Tests\EventListener\ErrorListenerTest::testHandleWithoutLogger with data set #0 (Symfony\Component\HttpKernel\Event\ExceptionEvent Object (...), Symfony\Component\HttpKernel\Event\ExceptionEvent Object (...))
Error: Call to undefined method Symfony\Component\ErrorHandler\ErrorHandler::enhanceError()
```

tests are KO because the new code does not exist yet in 7.0?

yceruto (approved):

yceruto:
> tests are KO because the new code does not exist yet in 7.0?

I think after synchronizing with 7.0 it should be fixed

src/Symfony/Component/ErrorHandler/ErrorHandler.php
```diff
@@ -661,12 +654,27 @@ private function renderException(\Throwable $exception): void
...
+        return $exception;
+    }
+
     /**
      * Override this method if you want to define more error enhancers.
      *
      * @return ErrorEnhancerInterface[]
      */
-    protected function getErrorEnhancers(): iterable
+    protected static function getErrorEnhancers(): iterable
```
nicolas-grekas: BC break ahead, isn't it?
Can't we get the ErrorHandler instance instead in ErrorListener? It must be somewhere for us to use.

stof: this is indeed a BC break.

lyrixx: Fixed

src/Symfony/Component/ErrorHandler/ErrorHandler.php
```diff
@@ -516,14 +516,7 @@ public function handleException(\Throwable $exception): void
...
 
-        if (!$exception instanceof OutOfMemoryError) {
-            foreach ($this->getErrorEnhancers() as $errorEnhancer) {
-                if ($e = $errorEnhancer->enhance($exception)) {
-                    $exception = $e;
-                    break;
-                }
-            }
-        }
+        $exception = self::enhanceError($exception);
```
yceruto: this should be `$this` instead of `self`, right?

so if a sub-class is overriding the `enhanceError()` method, it will be called for sure.

lyrixx: of course, I forgot to revert this change! Thanks (fixed)

yceruto (approved):

fancyweb (approved):

fabpot (approved):

fabpot:
Thank you @lyrixx.

PR #50913, opened 2023-07-07
[HttpKernel][WebProfilerBundle] Fix search feature

Pelagoss:
| Q             | A
| ------------- | ---
| Branch?       |  5.4 
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

This fix research feature in the profiler. 

Sometimes lines in index.csv are broken with something like this : 

,,200

the try and catch allow the profiler to skip these lines

### DIFF
--- a/src/Symfony/Component/HttpKernel/Profiler/FileProfilerStorage.php
+++ b/src/Symfony/Component/HttpKernel/Profiler/FileProfilerStorage.php
@@ -61,6 +61,12 @@ class FileProfilerStorage implements ProfilerStorageInterface
         $result = [];
         while (\count($result) < $limit && $line = $this->readLineFromFile($file)) {
             $values = str_getcsv($line);
+
+            if (7 !== \count($values)) {
+                // skip invalid lines
+                continue;
+            }
+
             [$csvToken, $csvIp, $csvMethod, $csvUrl, $csvTime, $csvParent, $csvStatusCode] = $values;
             $csvTime = (int) $csvTime;
 


GromNaN:
> Sometimes lines in index.csv are broken

It looks like you're trying to hide a problem in `FileProfilerStorage`. Do you know why there is invalid rows in the file?

Pelagoss:
I think it only occurs when the file is very big. Or maybe when multiple client requests the backend at same time

MatTheCat:
The root issue seems to be the same as #50816.

Pelagoss:
I will try with 6.3 and let you know

Pelagoss:
image
image
In this exemple (6.2.12), I dump everytime a line is writed in the .csv and look the line with only a "0", it corresponds with neither dump

src/Symfony/Component/HttpKernel/Profiler/FileProfilerStorage.php
```diff
@@ -61,7 +61,13 @@ public function find(?string $ip, ?string $url, ?int $limit, ?string $method, in
         $result = [];
         while (\count($result) < $limit && $line = $this->readLineFromFile($file)) {
             $values = str_getcsv($line);
-            [$csvToken, $csvIp, $csvMethod, $csvUrl, $csvTime, $csvParent, $csvStatusCode] = $values;
+            
+            try {
+                [$csvToken, $csvIp, $csvMethod, $csvUrl, $csvTime, $csvParent, $csvStatusCode] = $values;
+            } catch (\ErrorException $e) {
```
derrabus: This catch requires a specific error handler and we cannot assume that such an error handler is present at this point.

Don't catch `ErrorException`s. Prevent them from happening instead.

derrabus (changes requested):

src/Symfony/Component/HttpKernel/Profiler/FileProfilerStorage.php
Pelagoss: Sure, I Can check if $values has 7 values and if not, continue to the next iteration

nicolas-grekas:
Thank you @Pelagoss.

PR #50921, opened 2023-07-09
[Finder] Fix phpdoc about glob syntax

### LINKED ISSUE #42675: Finder name glob does not behave as documented
**Symfony version(s) affected**: 5.3.6

**Description**
According to the [`Finder::name()` docblock](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Finder/Finder.php#L167-L168), the following two calling conventions are equivalent:

https://github.com/symfony/symfony/blob/bf8ecc417fb900db91978cf6ec933319f7070083/src/Symfony/Component/Finder/Finder.php#L167-L168

However, this is demonstrably false for dot-files.

**How to reproduce**

Given the file `.foo.php` and `$finder` is called with `ignoreDotFiles(false)`:
* `$finder->name('*.php')` will not match.
* `$finder->name('/\.php$/')` will match.

This is because `*.php` is actually translated to the following regular expression: `#^(?=[^\.])[^/]*\.php$#`. The assertion prevents the file starting with a dot (`.`). (Aside: that assertion should probably have been written as `(?!\.)` for clarity).

It is unclear what is the intended behaviour here. By default, shells like Bash will not match dotfiles with a glob like this, although they can be set to match dot-files (with `shopt -s dotglob`). In any case, Finder's own documentation claims that the glob *should* match because the glob is equivalent to a suffix-only match, failing to mention that dot-files are implicitly excluded from such globs.

**Possible Solution**
Either:

* Allow globs to match dot-files.
* Change the documentation to stop pretending the glob and simple suffix-only regex are equivalent.

My view is that, from the perspective of Finder, the only correct solution can be to allow globs to match dot-files, because Finder has its own way of excluding dot-files (with `ignoreDotFiles()`). But then we are stuck with the problem that fixing the issue is a breaking change within Finder.

VincentLanglet:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes ? no ? just phpdoc
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #42675
| License       | MIT

Closes #42675 cc @nicolas-grekas

### DIFF
--- a/src/Symfony/Component/Finder/Finder.php
+++ b/src/Symfony/Component/Finder/Finder.php
@@ -164,8 +164,8 @@ class Finder implements \IteratorAggregate, \Countable
      *
      * You can use patterns (delimited with / sign), globs or simple strings.
      *
-     *     $finder->name('*.php')
-     *     $finder->name('/\.php$/') // same as above
+     *     $finder->name('/\.php$/')
+     *     $finder->name('*.php') // same as above, without dot files
      *     $finder->name('test.php')
      *     $finder->name(['test.py', 'test.php'])
      *


nicolas-grekas:
Thank you @VincentLanglet.

PR #50922, opened 2023-07-09
[Form] Deprecate not configuring the `default_protocol` option of the `UrlType`

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Tickets       | Part of #50871 
| License       | MIT
| Doc PR        | N/A

You would expect an `<input type="url">` from the `UrlType`, but it is not possible as long as `default_protocol` has a value, because then you have to accept inputs that are not URLs (and you get an `<input type="text" inputmode="url">`).

In order to change `default_protocol` from `'http'` to `null` in 8.0, we must first deprecate not configuring it.

### DIFF
--- a/src/Symfony/Component/Form/Extension/Core/Type/UrlType.php
+++ b/src/Symfony/Component/Form/Extension/Core/Type/UrlType.php
@@ -16,6 +16,7 @@ use Symfony\Component\Form\Extension\Core\EventListener\FixUrlProtocolListener;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\Form\FormView;
+use Symfony\Component\OptionsResolver\Options;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
 class UrlType extends AbstractType
@@ -38,7 +39,11 @@ class UrlType extends AbstractType
     public function configureOptions(OptionsResolver $resolver): void
     {
         $resolver->setDefaults([
-            'default_protocol' => 'http',
+            'default_protocol' => static function (Options $options) {
+                trigger_deprecation('symfony/form', '7.1', 'Not configuring the "default_protocol" option when using the UrlType is deprecated. It will default to "null" in 8.0.');
+
+                return 'http';
+            },
             'invalid_message' => 'Please enter a valid URL.',
         ]);
 
--- a/src/Symfony/Component/Form/Tests/Extension/Core/Type/UrlTypeTest.php
+++ b/src/Symfony/Component/Form/Tests/Extension/Core/Type/UrlTypeTest.php
@@ -11,14 +11,21 @@
 
 namespace Symfony\Component\Form\Tests\Extension\Core\Type;
 
+use Symfony\Bridge\PhpUnit\ExpectDeprecationTrait;
 use Symfony\Component\OptionsResolver\Exception\InvalidOptionsException;
 
 class UrlTypeTest extends TextTypeTest
 {
+    use ExpectDeprecationTrait;
+
     public const TESTED_TYPE = 'Symfony\Component\Form\Extension\Core\Type\UrlType';
 
+    /**
+     * @group legacy
+     */
     public function testSubmitAddsDefaultProtocolIfNoneIsIncluded()
     {
+        $this->expectDeprecation('Since symfony/form 7.1: Not configuring the "default_protocol" option when using the UrlType is deprecated. It will default to "null" in 8.0.');
         $form = $this->factory->create(static::TESTED_TYPE, 'name');
 
         $form->submit('www.domain.com');
@@ -86,6 +93,7 @@ class UrlTypeTest extends TextTypeTest
     public function testSubmitNullUsesDefaultEmptyData($emptyData = 'empty', $expectedData = 'http://empty')
     {
         $form = $this->factory->create(static::TESTED_TYPE, null, [
+            'default_protocol' => 'http',
             'empty_data' => $emptyData,
         ]);
         $form->submit(null);
@@ -94,5 +102,10 @@ class UrlTypeTest extends TextTypeTest
         $this->assertSame($expectedData, $form->getViewData());
         $this->assertSame($expectedData, $form->getNormData());
         $this->assertSame($expectedData, $form->getData());
+    }
+
+    protected function getTestOptions(): array
+    {
+        return ['default_protocol' => 'http'];
     }
 }
--- a/src/Symfony/Component/Form/Tests/Extension/Validator/Type/UrlTypeValidatorExtensionTest.php
+++ b/src/Symfony/Component/Form/Tests/Extension/Validator/Type/UrlTypeValidatorExtensionTest.php
@@ -20,7 +20,7 @@ class UrlTypeValidatorExtensionTest extends BaseValidatorExtensionTestCase
 
     protected function createForm(array $options = [])
     {
-        return $this->factory->create(UrlType::class, null, $options);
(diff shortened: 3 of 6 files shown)


xabbuh:
IIRC we overrode the tests in other cases. But we should add a comment that the overridden methods can be removed in the `7.0` branch.

MatTheCat:
Thanks @xabbuh, hope I did it correctly

src/Symfony/Component/Form/Tests/Extension/Core/Type/UrlTypeTest.php
```diff
@@ -83,9 +90,13 @@ public function testThrowExceptionIfDefaultProtocolIsInvalid()
         ]);
     }
 
+    /**
+     * @todo remove in 7.0
```
xabbuh: Can this test really be removed in 7.0?

MatTheCat: Nope, just line 99 :grimacing:

xabbuh: I would add the comment above the option then instead

MatTheCat: Damn I mixed up `testSubmitNullUsesDefaultEmptyData` and `testSubmitNullReturnsNullWithEmptyDataAsString`. Should be good now.

MatTheCat:
Is it still time to merge this in 6.4 or will I have to rebase on 7.1 when the branch will be created?

stof:
we should have a test ensuring that passing `null` explicitly (to not have any default protocol) works without triggering a deprecation. Otherwise it will make it impossible to have that behavior (which is the wanted one when you want an `<input type="url">`) without triggering a deprecation.

src/Symfony/Component/Form/Tests/Extension/Core/Type/UrlTypeTest.php
```diff
@@ -95,4 +103,47 @@ public function testSubmitNullUsesDefaultEmptyData($emptyData = 'empty', $expect
         $this->assertSame($expectedData, $form->getNormData());
         $this->assertSame($expectedData, $form->getData());
     }
+
+    /**
+     * @todo remove in 7.0
```
stof: As we won't remove the `default_protocol` feature in 7.0, there is no reason to remove this test if it covers a behavior that is not covered by other tests

src/Symfony/Component/Form/Tests/Extension/Core/Type/UrlTypeTest.php
```diff
@@ -95,4 +103,47 @@ public function testSubmitNullUsesDefaultEmptyData($emptyData = 'empty', $expect
...
+        $form->submit(null);
+        $this->assertSame('', $form->getData());
+        $this->assertSame('', $form->getNormData());
+        $this->assertSame('', $form->getViewData());
+    }
+
+    /**
+     * @dataProvider provideZeros
+     *
+     * @todo remove in 7.0
```
stof: same here

stof (changes requested):

src/Symfony/Component/Form/Tests/Extension/Core/Type/UrlTypeTest.php
MatTheCat: This test was inherited from `TextTypeTest` but it would have triggered a deprecation since it does not configure `default_protocol`. Is there a better way to handle this?

MatTheCat:
@stof thanks for the review :pray: 

Do you know how I can test whether a given code does not trigger a deprecation?

stof:
@MatTheCat Don't put the `@group legacy` on the test, and the phpunit bridge will already make the test fail if it triggers a deprecation.
But looking at the test file in full (not visible in the diff of the PR), we already have some tests passing `null` explicitly so this will be fine.

nicolas-grekas:
You should add a test that runs into the situation, and if it triggers a deprecation, the CI will fail.

src/Symfony/Component/Form/Tests/Extension/Core/Type/UrlTypeTest.php
stof: To me, the right fix would be that TextTypeTest should use the `getTestedType` and `getTestOptions` methods, as done in tests defined in BaseTypeTestCase, so that they respect the options defined by child test classes

src/Symfony/Component/Form/Tests/Extension/Core/Type/UrlTypeTest.php
```diff
@@ -86,6 +93,7 @@ public function testThrowExceptionIfDefaultProtocolIsInvalid()
     public function testSubmitNullUsesDefaultEmptyData($emptyData = 'empty', $expectedData = 'http://empty')
     {
         $form = $this->factory->create(static::TESTED_TYPE, null, [
+            'default_protocol' => 'http', // TODO: remove in 7.0
```
nicolas-grekas: that's fine if it stays as is so better remove all similar todo notices. Less merge conflicts ahead also.
```suggestion
            'default_protocol' => 'http',
```

src/Symfony/Component/Form/Tests/Extension/Validator/Type/UrlTypeValidatorExtensionTest.php
```diff
@@ -20,7 +20,8 @@ class UrlTypeValidatorExtensionTest extends BaseValidatorExtensionTestCase
 
     protected function createForm(array $options = [])
     {
-        return $this->factory->create(UrlType::class, null, $options);
+        // TODO: remove default_protocol option in 7.0
```
nicolas-grekas: not needed

src/Symfony/Component/Form/Tests/Extension/Core/Type/UrlTypeTest.php
MatTheCat: Indeed; done in https://github.com/symfony/symfony/pull/50922/commits/7956552402c76d490f21f1a0a15159f82b26ea9c

src/Symfony/Component/Form/Tests/Extension/Core/Type/UrlTypeTest.php
MatTheCat: Fixed in https://github.com/symfony/symfony/pull/50922/commits/7956552402c76d490f21f1a0a15159f82b26ea9c

stof (approved):

src/Symfony/Component/Form/Extension/Core/Type/UrlType.php
```diff
@@ -47,7 +48,11 @@ public function buildView(FormView $view, FormInterface $form, array $options)
     public function configureOptions(OptionsResolver $resolver)
     {
         $resolver->setDefaults([
-            'default_protocol' => 'http',
+            'default_protocol' => static function (Options $options) {
+                trigger_deprecation('symfony/form', '6.4', 'Not configuring the "default_protocol" option is deprecated. It will default to "null" in 7.0.');
```
nicolas-grekas:
```suggestion
                trigger_deprecation('symfony/form', '6.4', 'Not configuring the "default_protocol" option when using the UrlType is deprecated. It will default to "null" in 7.0.');
```

nicolas-grekas (approved):
I think this can wait for 7.1

fabpot (approved):

fabpot:
Thank you @MatTheCat.

PR #50931, opened 2023-07-11
[Form] Support Translatable Enum

### LINKED ISSUE #50919: Support translatable Enum in Forms
### Description

I found [this StackOverflow post](https://stackoverflow.com/a/74642943/1489264) showing how to make an Enum translatable using `TranslatableInterface`:

```php
enum NotifyType: int implements TranslatableInterface
{
    case EMAIL = 1;
    case WEBHOOK = 2;
    case PUNCH_OUT = 4;

    public function trans(TranslatorInterface $translator, string $locale = null): string
    {
        return match ($this) {
            self::EMAIL     => $translator->trans('supplier.notify.email', locale: $locale),
            self::WEBHOOK   => $translator->trans('supplier.notify.webhook', locale: $locale),
            self::PUNCH_OUT => $translator->trans('supplier.notify.punch_out', locale: $locale),
        };
    }
}
```

```
PHP usage
NotifyType::EMAIL->trans($this->translator)

Twig usage
{{ supplier.notifyType | trans }}
```

I would love to see this also supported by the Form component to automatically show the translated label when using an [`EnumType`](https://symfony.com/doc/current/reference/forms/types/enum.html) with an Enum implementing the `TranslatableInterface`.

This can be a good solution for https://github.com/symfony/symfony/issues/44596

### Example

_No response_

Seb33300:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix https://github.com/symfony/symfony/issues/50919
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/18599

This PR introduce support for Enum implementing `TranslatableInterface` in `EnumType`.

Example of use:

```php
$builder->add('textAlign', EnumType::class, [
    'class' => TextAlign::class,
])
```

```php

use Symfony\Contracts\Translation\TranslatableInterface;
use Symfony\Contracts\Translation\TranslatorInterface;

enum TextAlign: int implements TranslatableInterface
{
    case Left = 1;
    case Center = 2;
    case Right = 3;

    public function trans(TranslatorInterface $translator, string $locale = null): string
    {
        // Translate enum from name (Left, Center or Right)
        return $translator->trans($this->name, locale: $locale);

        // Translate enum from custom labels
        return match ($this) {
            self::Left   => $translator->trans('Left aligned', locale: $locale),
            self::Center => $translator->trans('Centered', locale: $locale),
            self::Right  => $translator->trans('Right aligned', locale: $locale),
        };
    }
}
```

### DIFF
--- a/src/Symfony/Component/Form/Tests/Extension/Core/Type/EnumTypeTest.php
+++ b/src/Symfony/Component/Form/Tests/Extension/Core/Type/EnumTypeTest.php
@@ -16,8 +16,11 @@ use Symfony\Component\Form\Tests\Extension\Core\Type\BaseTypeTestCase;
 use Symfony\Component\Form\Tests\Fixtures\Answer;
 use Symfony\Component\Form\Tests\Fixtures\Number;
 use Symfony\Component\Form\Tests\Fixtures\Suit;
+use Symfony\Component\Form\Tests\Fixtures\TranslatableTextAlign;
 use Symfony\Component\OptionsResolver\Exception\InvalidOptionsException;
 use Symfony\Component\OptionsResolver\Exception\MissingOptionsException;
+use Symfony\Component\Translation\IdentityTranslator;
+use Symfony\Contracts\Translation\TranslatableInterface;
 
 class EnumTypeTest extends BaseTypeTestCase
 {
@@ -255,6 +258,20 @@ class EnumTypeTest extends BaseTypeTestCase
         $view = $form->createView();
 
         $this->assertSame('Yes', $view->children[0]->vars['label']);
+    }
+
+    public function testChoiceLabelTranslatable()
+    {
+        $form = $this->factory->create($this->getTestedType(), null, [
+            'multiple' => false,
+            'expanded' => true,
+            'class' => TranslatableTextAlign::class,
+        ]);
+
+        $view = $form->createView();
+
+        $this->assertInstanceOf(TranslatableInterface::class, $view->children[0]->vars['label']);
+        $this->assertEquals('Left', $view->children[0]->vars['label']->trans(new IdentityTranslator()));
     }
 
     protected function getTestOptions(): array
--- a/src/Symfony/Component/Form/Extension/Core/Type/EnumType.php
+++ b/src/Symfony/Component/Form/Extension/Core/Type/EnumType.php
@@ -14,6 +14,7 @@ namespace Symfony\Component\Form\Extension\Core\Type;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\OptionsResolver\Options;
 use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Contracts\Translation\TranslatableInterface;
 
 /**
  * A choice type for native PHP enums.
@@ -29,7 +30,7 @@ final class EnumType extends AbstractType
             ->setAllowedTypes('class', 'string')
             ->setAllowedValues('class', enum_exists(...))
             ->setDefault('choices', static fn (Options $options): array => $options['class']::cases())
-            ->setDefault('choice_label', static fn (\UnitEnum $choice): string => $choice->name)
+            ->setDefault('choice_label', static fn (\UnitEnum $choice) => $choice instanceof TranslatableInterface ? $choice : $choice->name)
             ->setDefault('choice_value', static function (Options $options): ?\Closure {
                 if (!is_a($options['class'], \BackedEnum::class, true)) {
                     return null;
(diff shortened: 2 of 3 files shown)


yceruto (approved):

src/Symfony/Component/Form/Tests/Extension/Core/Type/EnumTypeTest.php
```diff
@@ -257,6 +259,20 @@ public function testChoiceLabel()
...
+        $form = $this->factory->create($this->getTestedType(), null, [
+            'multiple' => false,
+            'expanded' => true,
+            'class' => TranslatableTextAlign::class,
+        ]);
+
+        $view = $form->createView();
+
+        $this->assertInstanceOf(\UnitEnum::class, $view->children[0]->vars['label']);
+        $this->assertInstanceOf(TranslatableInterface::class, $view->children[0]->vars['label']);
```
derrabus: What are we actually testing here?

Seb33300: We check that the choice item is a "translatable enum" just before the translation happens.
(so the `choice_label` correctly return the a "translatable enum" and not just a string as before)

derrabus: Shouldn't we test that the translation actually happens?

Seb33300: I tried but I did no find an easy way to test it.
I suspect the translation to be handled by twig somehow when the field is rendered.
And unfortunately I did not have much more time to investigate on this.

If someone can help to improve the test, he will be very welcome.

ro0NL: it's roughly covered at https://github.com/symfony/symfony/blob/43066ff98d158946f6b5f8473802c9301f0abf49/src/Symfony/Bridge/Twig/Tests/Extension/AbstractLayoutTestCase.php#L568 already

Seb33300:
Just created the documentation PR: https://github.com/symfony/symfony-docs/pull/18599

src/Symfony/Component/Form/Tests/Extension/Core/Type/EnumTypeTest.php
derrabus: Very well, but what about this line:
```php
$this->assertInstanceOf(\UnitEnum::class, $view->children[0]->vars['label']);
```

Why would we assert that we get an arbitrary `UnitEnum` here? I'd remove that line.

```php
$this->assertInstanceOf(TranslatableInterface::class, $view->children[0]->vars['label']);
```

This assertion tells me that the label is actually translatable which is good. Maybe we could additionally call the `trans()` method that the label object exposes. That would tell us that we do get the correct translation.

Seb33300: I updated the test following your recommandations.

src/Symfony/Component/Form/Tests/Extension/Core/Type/EnumTypeTest.php
```diff
@@ -257,6 +260,24 @@ public function testChoiceLabel()
...
+            'expanded' => true,
+            'class' => TranslatableTextAlign::class,
+        ]);
+
+        $view = $form->createView();
+
+        $this->assertInstanceOf(TranslatableInterface::class, $view->children[0]->vars['label']);
+
+        $translator = $this->createMock(TranslatorInterface::class);
+        $translator->expects($this->once())->method('trans');
```
derrabus: You can use an `IdentityTranslator` instead.

src/Symfony/Component/Form/Tests/Extension/Core/Type/EnumTypeTest.php
```diff
@@ -257,6 +260,24 @@ public function testChoiceLabel()
...
+
+        $view->children[0]->vars['label']->trans($translator);
```
derrabus: Let's check the returned value shall we?

src/Symfony/Component/Form/Tests/Extension/Core/Type/EnumTypeTest.php
Seb33300: Done

src/Symfony/Component/Form/Tests/Extension/Core/Type/EnumTypeTest.php
Seb33300: Done

derrabus (approved):

derrabus:
Thank you @Seb33300.

PR #49557, opened 2023-02-28
[PropertyInfo] Fix phpDocExtractor nullable array value type

MudrakIvan:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #49555
| License       | MIT
| Doc PR        |

Bug fix for nullable array value types + for multiple array key/value types.
Two existing tests changed:
- property 'i' - change ?string|int comment to string|int|null (in newer versions (5.4+) already fixed)
- property 'iteratorCollection' - change key types from null to string/int (default key values from phpDoc)

Added new tests:
- property 'nonNullableCollectionOfNullableElements' - array<null|int> (nullable array value type)
- property 'nullableCollectionOfMultipleNonNullableElementTypes' - null|array<int|string> (nullable array of int and string value types)

PR #50933, opened 2023-07-11
[Serializer] Fix deserializing nested arrays of objects with mixed keys

### LINKED ISSUE #50675: Version v5.4.24 of symfony/property-info broke ArrayDenormalizer

v5.4.24

### Description

I have an array with numeric keys to denormalize, and the property the data will be assigned to is a [Doctrine ArrayCollection](https://github.com/doctrine/collections/blob/2.1.x/src/ArrayCollection.php). Previous to this release denormalization worked fine, but now this denormalization exception is collected:

```
Symfony\Component\Serializer\Exception\NotNormalizableValueException object
 currentType => string "int" (3)
 expectedTypes => array(1)
   integer 0 => string "string" (6)
 path => string "[0].answerOptions[0]" (20)
 useMessageForUser => true
 message => string "The type of the key "0" must be "string" ("int" given)." (55)
 code => integer 0
 file => string "path/to/symfony/serializer/Exception/NotNormalizableValueException.php" (98)
 line => integer 31
```

The exception is thrown in the ArrayDenormalizer class in line 58.

### How to reproduce

I can see that the [ArrayDenormalizerTest](https://github.com/symfony/symfony/blob/v5.4.24/src/Symfony/Component/Serializer/Tests/Normalizer/ArrayDenormalizerTest.php) class doesn't test arrays with numeric keys or no specified keys at all, so that might be the place to start.

The PhpDocExtractor does seem to extract both int and string correctly from the property but because they are passed individually to the ArrayDenormalizer, it fails when $context['key_type'] is 'string'.

### Possible Solution

The right solution is probably to update the type check in the ArrayDenormalizer class to check all possible types rather than just the key_type value.

### Additional Context

_No response_

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50675
| License       | MIT
| Doc PR        | -

Currently an error is thrown when trying to deserialize the following nested array of objects:

```php
use Symfony\Component\PropertyInfo\Extractor\PhpDocExtractor;
use Symfony\Component\Serializer\Normalizer\ArrayDenormalizer;
use Symfony\Component\Serializer\Normalizer\ObjectNormalizer;

$serializer = new Serializer([
    new ObjectNormalizer(null, null, null, new PhpDocExtractor()),
    new ArrayDenormalizer(),
], ['json' => new JsonEncoder()]);

class Outer
{
    /**
     * @var array<int|string, Inner>
     */
    public array $inners;
}

class Inner
{
    public string $name;
}

$serializer->deserialize('{"inners": {"1": {"name": "One"}, "two": {"name": "Two"}}}', Outer::class, 'json');
```

```
Fatal error: Uncaught Symfony\Component\Serializer\Exception\NotNormalizableValueException:
The type of the key "two" must be "int" ("string" given).
```

The same happens when using generics, e.g. `ArrayCollection<Inner>`.

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
@@ -532,7 +532,7 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
                     $class = $collectionValueType->getClassName().'[]';
 
                     if (\count($collectionKeyType = $type->getCollectionKeyTypes()) > 0) {
-                        [$context['key_type']] = $collectionKeyType;
+                        $context['key_type'] = \count($collectionKeyType) > 1 ? $collectionKeyType : $collectionKeyType[0];
                     }
 
                     $context['value_type'] = $collectionValueType;
--- a/src/Symfony/Component/Serializer/composer.json
+++ b/src/Symfony/Component/Serializer/composer.json
@@ -34,7 +34,7 @@
         "symfony/http-kernel": "^4.4|^5.0|^6.0",
         "symfony/mime": "^4.4|^5.0|^6.0",
         "symfony/property-access": "^5.4|^6.0",
-        "symfony/property-info": "^5.3.13|^6.0",
+        "symfony/property-info": "^5.4.24|^6.2.11",
         "symfony/uid": "^5.3|^6.0",
         "symfony/validator": "^4.4|^5.0|^6.0",
         "symfony/var-dumper": "^4.4|^5.0|^6.0",
@@ -47,7 +47,7 @@
         "phpdocumentor/type-resolver": "<1.4.0",
         "symfony/dependency-injection": "<4.4",
         "symfony/property-access": "<5.4",
-        "symfony/property-info": "<5.3.13",
+        "symfony/property-info": "<5.4.24|>=6,<6.2.11",
         "symfony/uid": "<5.3",
         "symfony/yaml": "<4.4"
     },
--- a/src/Symfony/Component/Serializer/Normalizer/ArrayDenormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/ArrayDenormalizer.php
@@ -49,14 +49,15 @@ class ArrayDenormalizer implements ContextAwareDenormalizerInterface, Denormaliz
 
         $type = substr($type, 0, -2);
 
-        $builtinType = isset($context['key_type']) ? $context['key_type']->getBuiltinType() : null;
+        $builtinTypes = array_map(static function (Type $keyType) {
+            return $keyType->getBuiltinType();
+        }, \is_array($keyType = $context['key_type'] ?? []) ? $keyType : [$keyType]);
+
         foreach ($data as $key => $value) {
             $subContext = $context;
             $subContext['deserialization_path'] = ($context['deserialization_path'] ?? false) ? sprintf('%s[%s]', $context['deserialization_path'], $key) : "[$key]";
 
-            if (null !== $builtinType && !('is_'.$builtinType)($key)) {
-                throw NotNormalizableValueException::createForUnexpectedDataType(sprintf('The type of the key "%s" must be "%s" ("%s" given).', $key, $builtinType, get_debug_type($key)), $key, [$builtinType], $subContext['deserialization_path'] ?? null, true);
-            }
+            $this->validateKeyType($builtinTypes, $key, $subContext['deserialization_path']);
 
             $data[$key] = $this->denormalizer->denormalize($value, $type, $format, $subContext);
         }
@@ -101,5 +102,23 @@ class ArrayDenormalizer implements ContextAwareDenormalizerInterface, Denormaliz
     public function hasCacheableSupportsMethod(): bool
     {
         return $this->denormalizer instanceof CacheableSupportsMethodInterface && $this->denormalizer->hasCacheableSupportsMethod();
+    }
+
+    /**
+     * @param mixed $key
+     */
+    private function validateKeyType(array $builtinTypes, $key, string $path): void
+    {
+        if (!$builtinTypes) {
+            return;
+        }
+
+        foreach ($builtinTypes as $builtinType) {
+            if (('is_'.$builtinType)($key)) {
+                return;
+            }
+        }
+
+        throw NotNormalizableValueException::createForUnexpectedDataType(sprintf('The type of the key "%s" must be "%s" ("%s" given).', $key, implode('", "', $builtinTypes), get_debug_type($key)), $key, $builtinTypes, $path, true);
     }
 }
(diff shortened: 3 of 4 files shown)


src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
```diff
@@ -533,6 +533,7 @@ private function validateAndDenormalize(array $types, string $currentClass, stri
 
                     if (\count($collectionKeyType = $type->getCollectionKeyTypes()) > 0) {
                         [$context['key_type']] = $collectionKeyType;
```
HypeMC: Not sure what to do about the `key_type` option, should it be deprecated in 6.4?

nicolas-grekas: Can't we reuse key_type? When there's only one, we keep as is, and when there are more, we go with an array?

HypeMC: Yes, we could, but wouldn't that be BC break since we would be changing the expected type? Or does that not count for internal parameters like this one?

nicolas-grekas: from what I understand, this was broken already for union types so :shrug:

HypeMC: Done

nicolas-grekas (commented):
Can you check the follow patch please?

<details>

```diff
diff --git a/src/Symfony/Component/Serializer/Normalizer/ArrayDenormalizer.php b/src/Symfony/Component/Serializer/Normalizer/ArrayDenormalizer.php
index 05d8b79d1a..dfd023a437 100644
--- a/src/Symfony/Component/Serializer/Normalizer/ArrayDenormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/ArrayDenormalizer.php
@@ -49,9 +49,7 @@ class ArrayDenormalizer implements ContextAwareDenormalizerInterface, Denormaliz
 
         $type = substr($type, 0, -2);
 
-        $builtinTypes = array_map(static function (Type $keyType) {
-            return $keyType->getBuiltinType();
-        }, \is_array($keyType = $context['key_type'] ?? []) ? $keyType : [$keyType]);
+        $builtinTypes = array_map(static function ($keyType) { return $keyType->getBuiltinType(); }, (array) ($context['key_type'] ?? []));
 
         foreach ($data as $key => $value) {
             $subContext = $context;
@@ -109,20 +107,12 @@ class ArrayDenormalizer implements ContextAwareDenormalizerInterface, Denormaliz
      */
     private function validateKeyType(array $builtinTypes, $key, string $path): void
     {
         if (!$builtinTypes) {
             return;
         }
 
-        $unexpectedDataType = true;
         foreach ($builtinTypes as $builtinType) {
             if (('is_'.$builtinType)($key)) {
-                $unexpectedDataType = false;
-                break;
+                return;
             }
         }
 
-        if ($unexpectedDataType) {
-            throw NotNormalizableValueException::createForUnexpectedDataType(sprintf('The type of the key "%s" must be "%s" ("%s" given).', $key, implode('", "', $builtinTypes), get_debug_type($key)), $key, $builtinTypes, $path, true);
-        }
+        throw NotNormalizableValueException::createForUnexpectedDataType(sprintf('The type of the key "%s" must be "%s" ("%s" given).', $key, implode('", "', $builtinTypes), get_debug_type($key)), $key, $builtinTypes, $path, true);
     }
 }
diff --git a/src/Symfony/Component/Serializer/composer.json b/src/Symfony/Component/Serializer/composer.json
index 1b9537f8f8..11e67a8a7e 100644
--- a/src/Symfony/Component/Serializer/composer.json
+++ b/src/Symfony/Component/Serializer/composer.json
@@ -34,7 +34,7 @@
         "symfony/http-kernel": "^4.4|^5.0|^6.0",
         "symfony/mime": "^4.4|^5.0|^6.0",
         "symfony/property-access": "^5.4|^6.0",
-        "symfony/property-info": "^5.4.24|^6.0",
+        "symfony/property-info": "^5.4.24|^6.2.11",
         "symfony/uid": "^5.3|^6.0",
         "symfony/validator": "^4.4|^5.0|^6.0",
         "symfony/var-dumper": "^4.4|^5.0|^6.0",
@@ -47,7 +47,7 @@
         "phpdocumentor/type-resolver": "<1.4.0",
         "symfony/dependency-injection": "<4.4",
         "symfony/property-access": "<5.4",
-        "symfony/property-info": "<5.4.24",
+        "symfony/property-info": "<5.4.24|>=6,<6.2.11",
         "symfony/uid": "<5.3",
         "symfony/yaml": "<4.4"
     },
```

</details>

src/Symfony/Component/Serializer/composer.json
```diff
@@ -34,7 +34,7 @@
         "symfony/http-kernel": "^4.4|^5.0|^6.0",
         "symfony/mime": "^4.4|^5.0|^6.0",
         "symfony/property-access": "^5.4|^6.0",
-        "symfony/property-info": "^5.3.13|^6.0",
+        "symfony/property-info": "^5.4.24|^6.0",
```
nicolas-grekas: "^5.4.24|^6.2.11" I guess, but can you confirm that 6.3.0 contains the fix we're needing here?

HypeMC: @nicolas-grekas Did you mean "^5.4.26|^6.2.13"? Also, 6.3.0 does not have this fix, so I'm guessing it should be "^5.4.26|^6.2.13|^6.3.2".

nicolas-grekas: I'm wondering why this change is needed at all: this PR patches only Serializer, not PropertyInfo.
Do you remember?

HypeMC:
> Can you check the follow patch please?

@nicolas-grekas The following part won't work:

```diff
-        $builtinTypes = array_map(static function (Type $keyType) {
-            return $keyType->getBuiltinType();
-        }, \is_array($keyType = $context['key_type'] ?? []) ? $keyType : [$keyType]);
+        $builtinTypes = array_map(static function ($keyType) { return $keyType->getBuiltinType(); }, (array) ($context['key_type'] ?? []));
 ```

`$context['key_type']` is not a scalar but an instance of `\Symfony\Component\PropertyInfo\Type`.

src/Symfony/Component/Serializer/composer.json
HypeMC: Yes, this fix requires #49557 to work. Without it the tests fail with lowest deps.

HypeMC: But yeah, you were right, it's "^5.4.24|^6.2.11", got a little confused for a second.

HypeMC: Done

nicolas-grekas:
Thank you @HypeMC.

PR #48342, opened 2022-11-26
[Form] Call `getChoicesForValues()` once, to prevent several SQL queries

elementaire:
| Q             | A
| ------------- | ---
| Branch?       | 6.3 
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | -
| License       | MIT
| Doc PR        | -

The feature "Keeping valid submitted choices when additional choices are submitted" adds some overload to list the valid choices.

With options set to `multiple=true`, `expanded=false` and using an `EntityType`, it performs several queries, one for each value.

PR #48208, opened 2022-11-13
[Form] Call `getChoicesForValues()` once, to prevent several SQL queries

elementaire:
| Q             | A
| ------------- | ---
| Branch?       | 4.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

The feature "Keeping valid submitted choices when additional choices are submitted" adds some overload to list the valid choices.

With options set to `multiple=true`, `expanded=false` and using an `EntityType`, it performs several queries, one for each value.

PR #50937, opened 2023-07-11
[Form] fetch all known ChoiceType values at once

### LINKED ISSUE #50898: [Form] EntityType in forms generates too many queries

5.4.24

### Description

The EntityType field in forms generates too many queries in the POST action.

If field is multiple and user select about 15/20 options POST method add additional queries in each options.

I know that the EntityType want to check if selected option is in our database.
But why does it work in foreach and send query through each selected options.

In **ORMQueryBuilderLoader.php** there is a method called **getEntitiesByIds**.
But when we see how it works we can see that it works in foreach of our selected options.

When we add additional inserts and selections the database queries are too large.

### How to reproduce

`$builder->add('test', EntityType::class, [
            'label' => 'Label test',
            'class' => YourClass::class,
            'multiple' => true,
            'required' => false,
            'choice_label' => 'name',
            'choice_value' => 'id',
        ]);`

### Possible Solution

Maybe there is the chance to change this foreach to one request with all selected options with IN array.

### Additional Context

_No response_

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50898
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Symfony/Component/Form/Extension/Core/Type/ChoiceType.php
+++ b/src/Symfony/Component/Form/Extension/Core/Type/ChoiceType.php
@@ -165,11 +165,9 @@ class ChoiceType extends AbstractType
                         }
                     }
                 } else {
-                    foreach ($data as $value) {
-                        if ($choiceList->getChoicesForValues([$value])) {
-                            $knownValues[] = $value;
-                            unset($unknownValues[$value]);
-                        }
+                    foreach ($choiceList->getChoicesForValues($data) as $key => $choice) {
+                        $knownValues[] = $data[$key];
+                        unset($unknownValues[$data[$key]]);
                     }
                 }
 


stof:
~shouldn't we also keep fetching them at once for expanded choice fields as well ?~

Forget that. The loop above for expanded fields does not call the choice list in that loop.

stof (approved):

stof:
Is there a way to write a test ensuring that a submission of a multiple choice type calls `getChoicesForValues` only once during the processing, to prevent similar regressions ?

fabpot (approved):

fabpot:
Thank you @xabbuh.

elementaire:
For record, denied as bug fix for 4.4 #48208, accepted as feature and already merged in 6.3 #48342

PR #50939, opened 2023-07-11
[SecurityBundle] Add `$badges` argument to `Security::login`

### LINKED ISSUE #50909: Programatically login and RememberMe
### Description

Before Symfony 6.2, I could programmatically login with RememberMe enabled like this:
```php
    #[Route('/register', name: 'app_register', methods: ['GET', 'POST'])]
    public function register(
        UserRepository             $userRepository,
        Request                    $request,
        LoginFormAuthenticator     $authenticator,
        UserAuthenticatorInterface $userAuthenticator,
    ): Response
    {

        //...
        return $userAuthenticator->authenticateUser($user, $authenticator, $request, [(new RememberMeBadge())->enable()]);
    }
```
Symfony 6.2 added a cleaner way to programmatically login, but it doesn't support badges.
```php
$security->login($user, LoginFormAuthenticator::class, 'main');
```

I'd like to suggest adding an optional parameter to `$security->login()` to support badges.

### Example

```php

    // symfony/security-bundle/Security.php

    public function login(UserInterface $user, string $authenticatorName = null, string $firewallName = null, array $badges = []): ?Response
    {
        // ...
        return $this->container->get('security.authenticator.managers_locator')->get($firewallName)->authenticateUser($user, $authenticator, $request, $badges);
    }

```

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #50909
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/18700

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/Security.php
+++ b/src/Symfony/Bundle/SecurityBundle/Security.php
@@ -24,6 +24,7 @@ use Symfony\Component\Security\Core\Security as LegacySecurity;
 use Symfony\Component\Security\Core\User\UserInterface;
 use Symfony\Component\Security\Csrf\CsrfToken;
 use Symfony\Component\Security\Http\Authenticator\AuthenticatorInterface;
+use Symfony\Component\Security\Http\Authenticator\Passport\Badge\BadgeInterface;
 use Symfony\Component\Security\Http\Event\LogoutEvent;
 use Symfony\Component\Security\Http\ParameterBagUtils;
 use Symfony\Component\Security\Http\SecurityRequestAttributes;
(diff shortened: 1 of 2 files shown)


src/Symfony/Bundle/SecurityBundle/Security.php
```diff
@@ -101,13 +102,14 @@ public function getFirewallConfig(Request $request): ?FirewallConfig
...
-     * @param string|null   $firewallName      The firewall name - required only if multiple firewalls are configured
+     * @param UserInterface    $user              The user to authenticate
+     * @param string|null      $authenticatorName The authenticator name (e.g. "form_login") or service id (e.g. SomeApiKeyAuthenticator::class) - required only if multiple authenticators are configured
+     * @param string|null      $firewallName      The firewall name - required only if multiple firewalls are configured
+     * @param BadgeInterface[] $badges            Badges to add to the user's passport
      *
      * @return Response|null The authenticator success response if any
      */
-    public function login(UserInterface $user, string $authenticatorName = null, string $firewallName = null): ?Response
+    public function login(UserInterface $user, string $authenticatorName = null, string $firewallName = null /* array $badges = [] */): ?Response
```
stof: As this class is `@final`, we should be able to add the argument natively without such BC layer

MatTheCat: Oh okay

fabpot (approved):

fabpot:
Thank you @MatTheCat.

PR #50945, opened 2023-07-11
[DebugBundle][FrameworkBundle] Fix using the framework without the Console component

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

While working on #50787 I've noticed that it's impossible to use the framework without the Console component, event though it's an optional dependency.
This PR aims to fix that. I've tested these changes on 5.4 & 6.3, however, I only made sure the container can compile. I'm not sure if there are any other gotchas.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
@@ -2154,12 +2154,14 @@ class FrameworkExtension extends Extension
             ->replaceArgument(0, $transportRetryReferences);
 
         if (\count($failureTransports) > 0) {
-            $container->getDefinition('console.command.messenger_failed_messages_retry')
-                ->replaceArgument(0, $config['failure_transport']);
-            $container->getDefinition('console.command.messenger_failed_messages_show')
-                ->replaceArgument(0, $config['failure_transport']);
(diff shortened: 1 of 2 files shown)


src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2154,12 +2154,14 @@ private function registerMessengerConfiguration(array $config, ContainerBuilder
             ->replaceArgument(0, $transportRetryReferences);
 
         if (\count($failureTransports) > 0) {
-            $container->getDefinition('console.command.messenger_failed_messages_retry')
-                ->replaceArgument(0, $config['failure_transport']);
-            $container->getDefinition('console.command.messenger_failed_messages_show')
-                ->replaceArgument(0, $config['failure_transport']);
-            $container->getDefinition('console.command.messenger_failed_messages_remove')
-                ->replaceArgument(0, $config['failure_transport']);
+            if ($this->hasConsole()) {
```
HypeMC, line 2157: Maybe it'd be better to call `hasDefinition()` for every command?

nicolas-grekas:
Thank you @HypeMC.

Issue #37786, opened 2020-08-09
[Routing]RouteSubscriberInterface


ahundiak:
Like the event subscriber, a route subscriber would be able to define their own routes.  No annotations needed and developers would have the power of the route configurator.
```
interface RouteSubscriberInterface
{
    public static function getRoutes(RoutingConfigurator $routes);
}
class HomeIndexAction implements RouteSubscriberInterface
{
    public static function getRoutes(RoutingConfigurator $routes)
    {
        $routes->add('home_index', "/")->controller(self::class);
    }

# config/routes.php
return function (RoutingConfigurator $routes) {

    HomeIndexAction::getRoutes($routes);
```
A file loader like the annotation loader could provide some automation.  Maybe a compiler pass as well.

derrabus:
That sounds like you‘re suggesting yet another way to load routes. Can you elaborate why the methods described in https://symfony.com/doc/current/routing/custom_route_loader.html don‘t work for you?

ahundiak:
To the best of my knowledge, annotations are the only approach for letting controllers define their own routes.  Annotations are of course not code and don't always provide useful feedback when writing.  Annotations appear to be very popular so the desire to define routes within the same file as the controller seems relevant.  As opposed to using external routing files.

My approach relies on the RoutingConfigurator which is basically a route builder and avoids the need for dealing with array keys and such.  Basically it replaces annotations with code while staying inside of the same file.  It is of course what the framework itself uses for PHP based routing configuration.

Annotations are also a bit awkward to use outside of the framework.  You need to load the doctrine annotation package which always seemed a bit strange to me.

I'm just proposing a standard component level interface that could be used to move code from config/routes.php to individual controller classes with a minimum of effort.  And could then potentially be used to automate the process of finding the route subscriber classes within the framework when building the route collection.

I think this approach is also consistent with current efforts to move to more PHP based configuration.

Guikingone:
> I think this approach is also consistent with current efforts to move to more PHP based configuration.

Actually, you can define routes in a `.php` file: https://symfony.com/doc/current/routing.html#creating-routes-in-yaml-xml-or-php-files.

That's probably the closest that you can get to php based configuration IMO 🤔

jvasseur:
> Annotations are also a bit awkward to use outside of the framework. You need to load the doctrine annotation package which always seemed a bit strange to me.

This requirement will be removed once https://github.com/symfony/symfony/pull/37474 is merged.

[...]

PR #50946, opened 2023-07-11
[Routing][SecurityBundle] Add `LogoutRouteLoader`

### LINKED ISSUE #50920: [Security] Automatically create logout route if it does not exists
### Description

Ref https://github.com/symfony/symfony/issues/45827#issuecomment-1620148846:

When the value of `logout` is set to `true`(which is equivalent to `logout: { path: '/logout' }`) or is set to a path e.g.`logout: {  path: '/some-logout-path' }` and no such route exists, we could define it automatically from a compiler pass or any another way (ideally with no runtime overhead). 
That would remove the error-prone and odd need to declare that route explicitly. 
We'd need to make the docs super clear about when one actually needs to define that route explicitly with or without attached controller. Bonus if we can make that route publicly accessible by default, if it doesn't make it too much confusing of course.

### Example

_No response_

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #50920
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/19000

#50920 is about avoiding for users to create logout routes. Given [we don’t want to allow bundles registering routes](https://github.com/symfony/symfony/issues/37786), I added a `LogoutRouteLoader` service bearing the `routing.route_loader` tag to be imported by the user. Such import could be added to the SecurityBundle recipe:

```yaml
# config/routes/security.yaml
logout:
    resource: security.route_loader.logout
    type: service
```

To invalidate routes when logout paths change, I stored them in a parameter so that the `ContainerParametersResourceChecker` can check the collection. Not sure if it’s okay or if a better way exists.

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/DependencyInjection/SecurityExtension.php
+++ b/src/Symfony/Bundle/SecurityBundle/DependencyInjection/SecurityExtension.php
@@ -49,6 +49,7 @@ use Symfony\Component\PasswordHasher\Hasher\NativePasswordHasher;
 use Symfony\Component\PasswordHasher\Hasher\Pbkdf2PasswordHasher;
 use Symfony\Component\PasswordHasher\Hasher\PlaintextPasswordHasher;
 use Symfony\Component\PasswordHasher\Hasher\SodiumPasswordHasher;
+use Symfony\Component\Routing\Loader\ContainerLoader;
 use Symfony\Component\Security\Core\Authorization\Strategy\AffirmativeStrategy;
 use Symfony\Component\Security\Core\Authorization\Strategy\ConsensusStrategy;
 use Symfony\Component\Security\Core\Authorization\Strategy\PriorityStrategy;
@@ -170,6 +171,13 @@ class SecurityExtension extends Extension implements PrependExtensionInterface
         }
 
         $this->createFirewalls($config, $container);
+
+        if ($container::willBeAvailable('symfony/routing', ContainerLoader::class, ['symfony/security-bundle'])) {
+            $this->createLogoutUrisParameter($config['firewalls'] ?? [], $container);
+        } else {
+            $container->removeDefinition('security.route_loader.logout');
+        }
+
         $this->createAuthorization($config, $container);
         $this->createRoleHierarchy($config, $container);
 
@@ -1094,5 +1102,21 @@ class SecurityExtension extends Extension implements PrependExtensionInterface
         }
 
         return $this->sortedFactories;
+    }
+
+    private function createLogoutUrisParameter(array $firewallsConfig, ContainerBuilder $container): void
+    {
+        $logoutUris = [];
+        foreach ($firewallsConfig as $name => $config) {
+            if (!$logoutPath = $config['logout']['path'] ?? null) {
+                continue;
+            }
+
+            if ('/' === $logoutPath[0]) {
+                $logoutUris[$name] = $logoutPath;
+            }
+        }
+
+        $container->setParameter('security.logout_uris', $logoutUris);
     }
 }
--- a/src/Symfony/Bundle/SecurityBundle/Routing/LogoutRouteLoader.php
+++ b/src/Symfony/Bundle/SecurityBundle/Routing/LogoutRouteLoader.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\SecurityBundle\Routing;
+
+use Symfony\Component\DependencyInjection\Config\ContainerParametersResource;
+use Symfony\Component\Routing\Route;
+use Symfony\Component\Routing\RouteCollection;
+
+final class LogoutRouteLoader
+{
+    /**
+     * @param array<string, string> $logoutUris    Logout URIs indexed by the corresponding firewall name
+     * @param string                $parameterName Name of the container parameter containing {@see $logoutUris}' value
+     */
+    public function __construct(
+        private readonly array $logoutUris,
+        private readonly string $parameterName,
+    ) {
+    }
+
+    public function __invoke(): RouteCollection
+    {
+        $collection = new RouteCollection();
+        $collection->addResource(new ContainerParametersResource([$this->parameterName => $this->logoutUris]));
+
+        $routeNames = [];
+        foreach ($this->logoutUris as $firewallName => $logoutPath) {
+            $routeName = '_logout_'.$firewallName;
+
+            if (isset($routeNames[$logoutPath])) {
+                $collection->addAlias($routeName, $routeNames[$logoutPath]);
+            } else {
+                $routeNames[$logoutPath] = $routeName;
+                $collection->add($routeName, new Route($logoutPath));
+            }
+        }
+
+        return $collection;
+    }
+}
--- a/src/Symfony/Bundle/SecurityBundle/CHANGELOG.md
+++ b/src/Symfony/Bundle/SecurityBundle/CHANGELOG.md
@@ -8,6 +8,7 @@ CHANGELOG
  * Allow an array of `pattern` in firewall configuration
  * Add `$badges` argument to `Security::login`
  * Deprecate the `require_previous_session` config option. Setting it has no effect anymore
+ * Add `LogoutRouteLoader`
 
 6.3
 ---
--- a/src/Symfony/Bundle/SecurityBundle/Resources/config/security.php
+++ b/src/Symfony/Bundle/SecurityBundle/Resources/config/security.php
@@ -13,6 +13,7 @@ namespace Symfony\Component\DependencyInjection\Loader\Configurator;
 
 use Symfony\Bundle\SecurityBundle\CacheWarmer\ExpressionCacheWarmer;
 use Symfony\Bundle\SecurityBundle\EventListener\FirewallListener;
+use Symfony\Bundle\SecurityBundle\Routing\LogoutRouteLoader;
 use Symfony\Bundle\SecurityBundle\Security;
 use Symfony\Bundle\SecurityBundle\Security\FirewallConfig;
 use Symfony\Bundle\SecurityBundle\Security\FirewallContext;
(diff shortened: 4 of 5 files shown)


wouterj (commented):
Looks very cool to me. Thanks for creating this PR :)

I feel like the `ContainerParameterResource` is probably the only way to do this indeed. Maybe someone else has a better suggestion?

MatTheCat:
Gave it a shot. Still open to suggestions!

src/Symfony/Bundle/SecurityBundle/DependencyInjection/SecurityExtension.php
```diff
@@ -170,6 +171,13 @@ public function load(array $configs, ContainerBuilder $container)
         }
 
         $this->createFirewalls($config, $container);
+
+        if ($container::willBeAvailable('symfony/routing', ContainerLoader::class, ['symfony/security/bundle'])) {
```
chalasr:
```suggestion
        if ($container::willBeAvailable('symfony/routing', ContainerLoader::class, ['symfony/security-bundle'])) {
```

chalasr (changes requested):

src/Symfony/Bundle/SecurityBundle/DependencyInjection/SecurityExtension.php
MatTheCat: Woops :see_no_evil:

chalasr (approved):

src/Symfony/Bundle/SecurityBundle/Routing/LogoutRouteLoader.php
```diff
@@ -0,0 +1,50 @@
...
+namespace Symfony\Bundle\SecurityBundle\Routing;
+
+use Symfony\Component\DependencyInjection\Config\ContainerParametersResource;
+use Symfony\Component\Routing\Route;
+use Symfony\Component\Routing\RouteCollection;
+
+class LogoutRouteLoader
+{
+    /**
+     * @param iterable<string, string> $logoutPaths   Logout paths indexed by the corresponding firewall name
```
stof: looking at the usage in ContainerParametersResource, does this really support any iterables or only arrays ? Container parameters don't allow objects.

src/Symfony/Bundle/SecurityBundle/Routing/LogoutRouteLoader.php
```diff
@@ -0,0 +1,50 @@
...
+        foreach ($this->logoutPaths as $firewallName => $logoutPath) {
+            $routeName = '_logout_'.$firewallName;
+
+            if (isset($routeNames[$logoutPath])) {
+                $collection->addAlias($routeNames[$logoutPath], $routeName);
+            } else {
+                $routeNames[$logoutPath] = $routeName;
+            }
+
+            $collection->add($routeName, new Route($logoutPath));
```
stof: shouldn't this be in the `else` block instead of adding a route when we already add an alias ?

src/Symfony/Bundle/SecurityBundle/Routing/LogoutRouteLoader.php
```diff
@@ -0,0 +1,50 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\SecurityBundle\Routing;
+
+use Symfony\Component\DependencyInjection\Config\ContainerParametersResource;
+use Symfony\Component\Routing\Route;
+use Symfony\Component\Routing\RouteCollection;
+
+class LogoutRouteLoader
```
stof: should probably be final (or at least `@final`) to

src/Symfony/Bundle/SecurityBundle/DependencyInjection/SecurityExtension.php
```diff
@@ -1091,4 +1099,16 @@ private function getSortedFactories(): array
 
         return $this->sortedFactories;
     }
+
+    private function createLogoutPathsParameter(array $firewallsConfig, ContainerBuilder $container): void
+    {
+        $logoutPaths = [];
+        foreach ($firewallsConfig as $name => $config) {
+            if ($logoutPath = $config['logout']['path'] ?? null) {
+                $logoutPaths[$name] = $logoutPath;
```
stof: Be careful. The SecurityBundle configuration supports both route names and actual paths as part of this config (see `Symfony\Component\Security\Http\HttpUtils`).

For configurations that reference a route name, we must not registering them in this parameter

stof (changes requested):

src/Symfony/Bundle/SecurityBundle/Routing/LogoutRouteLoader.php
MatTheCat: I forgot to change the type when trying the `ContainerParametersResourceChecker` :see_no_evil: there is no need to hint `$logoutPaths` as iterable

src/Symfony/Bundle/SecurityBundle/Routing/LogoutRouteLoader.php
MatTheCat: I was surprised about that too, but `RouteCollection::addAlias` will remove the original route, so I have to add it anyways: https://github.com/symfony/symfony/blob/33fedcee2814947c59df2570952f8c800dfdb23e/src/Symfony/Component/Routing/RouteCollection.php#L372-L381

So the test would fail because the collection will be empty.

src/Symfony/Bundle/SecurityBundle/DependencyInjection/SecurityExtension.php
MatTheCat: I checked paths not starting with `/` like the `LogoutUrlGenerator` does.

maxhelias (approved):

src/Symfony/Bundle/SecurityBundle/Routing/LogoutRouteLoader.php
stof: well, if you define a name as being an alias, it cannot be a route as well (each name is either an alias of another route or a route itself).
adding the route will remove the alias.

MatTheCat: Not sure to understand; is there something left to do here? :sweat_smile:

stof: if you always add the route at the end of the logic, all the handling of adding aliases is dead code as you always override the alias.

stof: ah no. this is even more weird.

Your current logic actually replace the **previous** route with an alias to the newly added route (and you might end up with a chain of aliases if this gets done multiple times)

MatTheCat: Yes, that way we have one route and duplicates are aliases. Is there something wrong with this approach?

src/Symfony/Bundle/SecurityBundle/Routing/LogoutRouteLoader.php
```diff
@@ -0,0 +1,50 @@
...
+        foreach ($this->logoutUris as $firewallName => $logoutPath) {
+            $routeName = '_logout_'.$firewallName;
+
+            if (isset($routeNames[$logoutPath])) {
+                $collection->addAlias($routeNames[$logoutPath], $routeName);
+            } else {
+                $routeNames[$logoutPath] = $routeName;
+            }
+
+            $collection->add($routeName, new Route($logoutPath));
```
maxhelias: I think we can do something here, to prevent this issue : https://github.com/symfony/symfony/issues/51313

src/Symfony/Bundle/SecurityBundle/Routing/LogoutRouteLoader.php
stof: @MatTheCat you could also achieve that by defining an alias for the first occurrence of the logout path and then defining next routes as aliases of the first one. this would be much simpler than defining the last firewall as a route and previous ones as aliases.

currently, you are defining this in the end:

```
_logout_firewall1:
  alias: _logout_firewall2

_logout_firewall2:
  alias: _logout_firewall3

_logout_firewall3:
  alias: _logout_firewall4

_logout_firewall4:
  path: /logout
```

the more firewall you have with the same logout path, the longer the alias chain becomes to generate the logout URL for the first one.

stof: And re-reading the code, I'm not even sure that this is what happens here as I think you will actually end up with the first firewall being defining as an alias to the last route and all other firewalls as actual routes (as `$routeNames[$logoutPath]` will stay the name of the route of the first firewall, and this is the alias being redefined constantly).

My guess is that you confused the meaning of the 2 parameters of `addAlias`

MatTheCat: Spot-on analysis! Should be better with https://github.com/symfony/symfony/pull/50946/commits/f9dba15a020219be32fd2bbd4a223f6b40dbb7fc

src/Symfony/Bundle/SecurityBundle/Routing/LogoutRouteLoader.php
MatTheCat: Do you mean we should restrict these routes to the POST method?

maxhelias: After reading @Spomky's comment https://github.com/symfony/symfony/issues/51313#issuecomment-1685219937, I would tend to agree with him

MatTheCat: There is more to this than just restricting the routesʼ method. I don't think this PR is a good place to introduce such changes.

maxhelias: You're right, it probably has an impact on the LogoutListener, LogoutUrlGenerator or LogoutUrlExtension.
It would be better to do a general PR on the topic

wouterj (approved):

wouterj:
I like the nice little DX improvement here! Thanks for working on this Mathieu! Let's get this in 6.4/7.0

PR #50960, opened 2023-07-12
[VarDumper] Fix dumping `ArrayObject` with `DumpDataCollector`

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50940
| License       | MIT
| Doc PR        | -

The main problem here is that the `ArrayObjectstorage` key in [the array returned by `__debugInfo()`](https://github.com/symfony/symfony/blob/de7ab4d85dcc94032b77f1a15d7a030714a1f734/src/Symfony/Component/VarDumper/Caster/SplCaster.php#L229-L229) seems to be a reference: https://3v4l.org/8rSFn (a bug in PHP perhaps?)

Since the `DumpDataCollector` does the actual dumping in the `__destructor()` the `ArrayObjectstorage` is modified by then which messes with the object created by the `VarCloner`.


src/Symfony/Component/VarDumper/Dumper/ContextProvider/SourceContextProvider.php
```diff
@@ -44,7 +44,7 @@ public function getContext(): ?array
 
         $file = $trace[1]['file'];
         $line = $trace[1]['line'];
-        $name = false;
+        $name = 'Standard input code' === $file ? $file : false;
```
nicolas-grekas:
```suggestion
        $name = '-' === $file || 'Standard input code' === $file ? 'Standard input code' : false;
```

nicolas-grekas (approved):
nice thanks!

src/Symfony/Component/VarDumper/Dumper/ContextProvider/SourceContextProvider.php
HypeMC: Done

lyrixx (approved):
Yeah! Thanks

nicolas-grekas:
Thank you @HypeMC.

PR #50972, opened 2023-07-13
[Doctrine][Messenger] Make the messenger doctrine transport scalable

### LINKED ISSUE #47828: [Messenger] Bad performances with the Doctrine/MySQL transport

5.4.13 (but I guess it happens with 6.* as well as the code is the same)

### Description

We have the following setup
* MySQL 8.0.15 with an InnoDb engine and `READ-COMMITTED` as transaction isolation level
* 4 Ubuntu-based workers running supervisor which runs 2 processes
* Command to consume: `php bin/console messenger:consume async_email --limit=100 --time-limit=300 --memory-limit=480M`

We experienced a lag of up to ~80k messages in the queue because workers spend most of their time waiting to lock a message to process.
A colleague tried to increase the number of processes from 8 to 80 but it made things worst.
Our processing tops at 3 or 4 consumed messages per seconds on this specific queue

Here are my findings so far:

The MySQL command `SHOW FULL PROCESSLIST` shows a lot of the `SELECT ... FOR UPDATE` requests are in the `Sending data` state.

The `SELECT ... FOR UPDATE` uses the index to find and lock a message:
```
EXPLAIN SELECT m.id FROM messenger_email m WHERE (m.delivered_at is null OR m.delivered_at < '2022-10-10 20:16:29') AND (m.available_at <= '2022-10-10 21:16:29') AND (m.queue_name = 'email') ORDER BY available_at ASC LIMIT 1 FOR UPDATE
```

id | select_type | table | partitions | type | possible_keys | key | key_len | ref | rows | filtered | Extra
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | SIMPLE | m | NULL | range | IDX_DC4C2A4016BA31DB,IDX_DC4C2A40E3BD61CE,IDX_DC4C2A40FB7336F0 | IDX_DC4C2A40E3BD61CE | 5 | NULL | 37451 | 25.00 | Using index condition; Using where

_IDX_DC4C2A40E3BD61CE is on the `available_at` column_

Because it uses an index and we are using `READ-COMMITTED` as the transaction isolation level, then the following happens:
> If the WHERE condition includes an indexed column, and InnoDB uses the index, only the indexed column is considered when taking and retaining record locks. In the following example, the first [UPDATE](https://dev.mysql.com/doc/refman/8.0/en/update.html) takes and retains an x-lock on each row where b = 2. The second [UPDATE](https://dev.mysql.com/doc/refman/8.0/en/update.html) blocks when it tries to acquire x-locks on the same records, as it also uses the index defined on column b.
```
CREATE TABLE t (a INT NOT NULL, b INT, c INT, INDEX (b)) ENGINE = InnoDB;
INSERT INTO t VALUES (1,2,3),(2,2,4);
COMMIT;

# Session A
START TRANSACTION;
UPDATE t SET b = 3 WHERE b = 2 AND c = 3;

# Session B
UPDATE t SET b = 4 WHERE b = 2 AND c = 4;
```
Source: https://dev.mysql.com/doc/refman/8.0/en/innodb-transaction-isolation-levels.html

So the bottleneck in my case happens when workers search for a message to lock because the query is blocking.
Once locked, the message is very very quick to handle.

### How to reproduce

I guess it's quite difficult to reproduce as I think there is a balance between:
* The number of processes consuming the queue
* The time spend to consume one message
* The load on the MySQL server

### Possible Solution

# Solution 1 - Remove the indexes
As suggested by https://github.com/symfony/symfony/issues/42868 removing the indexes actually removes the blockings issue as per MySQL documentation because without indexes:
> If READ COMMITTED is used instead, the first [UPDATE](https://dev.mysql.com/doc/refman/8.0/en/update.html) acquires an x-lock on each row that it reads and releases those for rows that it does not modify:

```
x-lock(1,2); unlock(1,2)
x-lock(2,3); update(2,3) to (2,5); retain x-lock
x-lock(3,2); unlock(3,2)
x-lock(4,3); update(4,3) to (4,5); retain x-lock
x-lock(5,2); unlock(5,2)
```
> For the second UPDATE, InnoDB does a “semi-consistent” read, returning the latest committed version of each row that it reads to MySQL so that MySQL can determine whether the row matches the WHERE condition of the [UPDATE](https://dev.mysql.com/doc/refman/8.0/en/update.html):

```
x-lock(1,2); update(1,2) to (1,4); retain x-lock
x-lock(2,3); unlock(2,3)
x-lock(3,2); update(3,2) to (3,4); retain x-lock
x-lock(4,3); unlock(4,3)
x-lock(5,2); update(5,2) to (5,4); retain x-lock
```

However, when there are a lot of messages, it leads to slow queries because MySQL uses a "filesort" because of the `ORDER BY available_at` clause.
I think this clause could be removed if losing the FIFO-like strategy is acceptable

# Solution 2 - Use SKIP LOCKED
MySQL has a `SKIP LOCKED` clause that could be used here: https://dev.mysql.com/doc/refman/8.0/en/innodb-locking-reads.html
> A locking read that uses SKIP LOCKED never waits to acquire a row lock. The query executes immediately, removing locked rows from the result set.

I measured these performances for the SELECT query:
* Without `FOR UPDATE` & `SKIP LOCKED`: about 0.002 seconds to return a row
* With `FOR UPDATE` & without `SKIP LOCKED`: about 2 seconds to return a row
* With `FOR UPDATE`& with `SKIP LOCKED`: about 0.2 seconds to return a row

=> x10 faster with the `SKIP LOCKED` option

As far as I know, `doctrine/dbal` doesn't support the `SKIP LOCKED` option.
An issue has been around for years about it with `doctrine/orm`: https://github.com/doctrine/orm/issues/7746

I suggest here that the Doctrine transport "manually" adds the `SKIP LOCKED` option when the database platform is MySQL 8.0 or later (maybe others that support it)

This is my preferred solution

### Additional Context

_No response_

hgraca:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Tickets       | Fix [#47828](https://github.com/symfony/symfony/issues/47828)
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/18566

### Relevant changes
This PR has a few commits with small improvements and two commits with relevant changes.

Here are the most significant changes:

#### Replacement and removal of the DoctrineReceivedStamp
It seems to be redundant because there is already another stamp there being used for the same purpose. Thus, in a commit I replace its usage and in the other commit I remove it. This might be considered a breaking change because the DoctrineReceivedStamp is not marked final, Im not sure, you tell me, these commits can be easily dropped without impact to the rest :)

#### Add "SKIP LOCKED" to the Mysql8 and Postgres query that retrieves messages

The current `SELECT ... FOR UPDATE` retrieves rows, using the index to find
and lock the retrieved rows.
This means that when we have more than one consumer, while
one consumer is locking those rows, the whole index is locked thus
other queries (consumers) will be put on hold.
While with a small table this might not be noticeable, as the table
grows the meddling with the index becomes slower and the other
consumers have to wait more time, eventually making the MQ inoperable.

The `SKIP LOCKED` addition will allow other consumers to query the
table and get messages immediately, ignoring the rows that other
consumers are locking.

#### Add batch delivery

Currently, the doctrine transport only delivers one message at a time,
resulting in low performance and a very high amount of hits to the DB.
(one hit per message)

With batch delivery, the transport will retrieve several messages in a single query.

### Testing

I adjusted the tests, and added a few more cases.

I also tested this in a playground project, sending and receiving 400000 asynchronous messages, which were handled by 2 worker containers, each with 2 consumers.
All messages were processed and there were no duplicates.
These tests were done with mysql, postgres and mssql.

Maria DB was not updated with the `SKIP LOCKED` because the engine version that supports it, is not yet supported by Doctrine.

Oracle was not updated with `SKIP LOCKED` either because I could not even connect to the damn thing using my db client.


derrabus:
Thank you very much for this PR.

> This PR has a few commits with small improvements and two commits with relevant changes.

I think, this should've been three PRs. Because you've changed multiple aspects of the Doctrine transport, the review might take a while. Smaller isolated changes are usually easier to review.

> #### Replacement and removal of the DoctrineReceivedStamp
> 
> It seems to be redundant because there is already another stamp there being used for the same purpose. Thus, in a commit I replace its usage and in the other commit I remove it. This might be considered a breaking change because the DoctrineReceivedStamp is not marked final, Im not sure, you tell me, these commits can be easily dropped without impact to the rest :)

Changing the stamps that the transport attaches to the message is a breaking change. If you can split this into a separate PR, please do that and we discuss the change there.

> #### Add "SKIP LOCKED" to the Mysql8 and Postgres query that retrieves messages

Very nice. I wonder if it makes sense to contribute support for `SKIP LOCKED` to DBAL. This kind of query manipulation feels like it belongs there.

> #### Add batch delivery

Do we have other transports that support batches? Would it makes sense to generalize this feature?

> Maria DB was not updated with the `SKIP LOCKED` because the engine version that supports it, is not yet supported by Doctrine.

And what version would that be? DBAL supports all MariaDB versions since 10.0, up to 11.0 currently.

> Oracle was not updated with `SKIP LOCKED` either because I could not even connect to the damn thing using my db client.

😄

hgraca:
Hi @derrabus tkx for the review.

> I think, this should've been three PRs.

How would you separate the commits per PR?
Would this
```
1. 
Add DoctrineTransportFactory DSN_PREFIX constant 
Add type hint for $driverConnection in DoctrineTransportFactory
Ensure messages flagged for deletion have clear flag 

2.
Replace usage of DoctrineReceivedStamp
Remove the unused DoctrineReceivedStamp

3.
Add "SKIP LOCKED" to the query that retrieves messages

4.
Add batch delivery
```
make sense?

> Changing the stamps that the transport attaches to the message is a breaking change. If you can split this into a separate PR, please do that and we discuss the change there.

Will do

> I wonder if it makes sense to contribute support for SKIP LOCKED to DBAL.

Yes, I believe it makes sense, I was eager to get this functionality in the messenger though.
Do you want me to do that update in DBAL first?

> Do we have other transports that support batches? Would it makes sense to generalize this feature?

The transport was already prepared for this, cozz the `ReceiverInterface::get(): iterable;` returns an iterable (batch of messages), and it was in fact returning `[Envelope]` and then consumer is handling all envelopes in the array, so I guess it is already the case that it is generalised, just the doctrine transport was not making use of it.

> And what version would that be? DBAL supports all MariaDB versions since 10.0, up to 11.0 currently.

The SKIP LOCKED clause was introduced in MariaDB 10.6.0, but DBAL supports from `10.2.7 GA` and there's no way to choose a higher version (AFAIK) so I assumed we needed to support the lowest version.
Let me know if that is not the case and I can add support for MariaDB.

derrabus:
> > I think, this should've been three PRs.
> 
> How would you separate the commits per PR? Would this

We usually squash PRs before merging. Since you deliver multiple features with this PR, multiple PRs would make it easier to discuss the implementation of one feature without blocking the merge of another feature.

> > I wonder if it makes sense to contribute support for SKIP LOCKED to DBAL.
> 
> Yes, I believe it makes sense, I was eager to get this functionality in the messenger though. Do you want me to do that update in DBAL first?

Since we're targeting 6.4 which is due in November, we're not in a hurry. If you have an idea how to contribute this to DBAL, please go ahead.

> I guess it is already the case that it is generalised, just the doctrine transport was not making use of it.

I know, but unless I've missed something, this PR would be the first that makes the batch size configurable. But that was just an idea/observation from my side. We can go with your approach and think about generalizing that functionality later.

> DBAL supports from `10.2.7 GA` and there's no way to choose a higher version (AFAIK)

DBAL does not introduce new platform classes unless they're needed. DBAL 3.7 will ship new ones for MariaDB, but still not the one you need. You can query the database version from the connection though. However, the fact that you'll need a new versioned platform class probably indicates that adding that feature to DBAL is a good idea.

hgraca:
@derrabus 
Cool, I will break this up in separate PRs, open a PR in DBAL and when that's merged I will open the PR for the `SKIP LOCKED` commit here.

hgraca:
Opened up this PR
https://github.com/symfony/symfony/pull/51034

The others will follow up after the DBAL PR is merged

PR #50974, opened 2023-07-14
[Workflow] Add support for storing the marking in a property

### LINKED ISSUE #49778: symfony/workflow: custom marking store with access to property or arguments
### Description

We would find it helpful if a custom implementation of [`MarkingStoreInterface`](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Workflow/MarkingStore/MarkingStoreInterface.php) had access to the configured `property` in workflow configuration. Currently, it does not seem possible.

Default [`MethodMarkingStore`](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Workflow/MarkingStore/MethodMarkingStore.php) takes into account the configured `property` name:

```yaml
framework:
    workflows:
        order:
            type: "state_machine"
            marking_store:
                type: "method"
                property: "status"
```

With this settings, the `MethodMarkingStore` operates with property `status` instead of the property `marking` which is set as default value in constructor:

```php
final class MethodMarkingStore implements MarkingStoreInterface
{
    public function __construct(bool $singleState = false, string $property = 'marking')
    {
        $this->property = $property;
        // ...
    }
```

We would expect similar behaviour when using custom marking store, but the `property` configuration does not seem to have any effect:

```yaml
framework:
    workflows:
        order:
            type: "state_machine"
            marking_store:
                service: app.marking_store.custom
                property: "status"
```

```php
namespace App\Workflow\MarkingStore;

use Symfony\Component\Workflow\MarkingStore\MarkingStoreInterface;

class CustomMarkingStore implements MarkingStoreInterface
{
    protected string $property;

    public function __construct(string $property = 'marking')
    {
        $this->property = $property;
    }
}
```

In this case, the `CustomMarkingStore` always operates with property `marking` (which is set as default in constructor), but it should operate with property `status` instead.

Alternatively, having an option to define service arguments in workflow configuration would solve the issue:

```yaml
framework:
    workflows:
        order:
            type: "state_machine"
            marking_store:
                service: app.marking_store.custom
                arguments:
                    $property: "status"
```

This seems to have been [possible in version Symfony 4.4](https://github.com/symfony/symfony/blob/4.4/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php), but is no longer possible in Symfony 6.2.

### Example

Currently, the only way to achieve the expected behaviour is to register a pre-configured service for each workflow definition like this:

```yaml
services:
    app.marking_store.oder:
        class: App\Workflow\MarkingStore\CustomMarkingStore
        arguments:
            $property: 'status'
    app.marking_store.item:
        class: App\Workflow\MarkingStore\CustomMarkingStore
        arguments:
            $property: 'condition'
```

```yaml
framework:
    workflows:
        order:
            type: "state_machine"
            marking_store:
                service: app.marking_store.order
        item:
            type: "state_machine"
            marking_store:
                service: app.marking_store.item
```

This feels unfortunate in case we have many workflow definitions that use the same custom marking store, but differ in property name only.

If using `property` or `arguments` as described above is not an option, would it be possible to at least introduce a new interface, for example `PropertyAwareInterface` that could be implemented together with `MarkingStoreInterface` like this?

```php
namespace App\Workflow\MarkingStore;

use Symfony\Component\Workflow\MarkingStore\MarkingStoreInterface;
use Symfony\Component\Workflow\MarkingStore\PropertyAwareInterface;

class CustomMarkingStore implements MarkingStoreInterface, PropertyAwareInterface
{
    protected string $property;

    /**
      * {@inheritdoc}
      */
    public function setProperty(string $property): void
    {
        $this->property = $property;
    }
}
```

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | fixes #49778
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Symfony/Component/Workflow/CHANGELOG.md
+++ b/src/Symfony/Component/Workflow/CHANGELOG.md
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Add `with-metadata` option to the `workflow:dump` command to include places,
    transitions and workflow's metadata into dumped graph
+ * Add support for storing marking in a property
 
 6.2
 ---
--- a/src/Symfony/Component/Workflow/MarkingStore/MethodMarkingStore.php
+++ b/src/Symfony/Component/Workflow/MarkingStore/MethodMarkingStore.php
@@ -15,45 +15,43 @@ use Symfony\Component\Workflow\Exception\LogicException;
 use Symfony\Component\Workflow\Marking;
 
 /**
- * MethodMarkingStore stores the marking with a subject's method.
+ * MethodMarkingStore stores the marking with a subject's public method
+ * or public property.
  *
- * This store deals with a "single state" or "multiple state" Marking.
+ * This store deals with a "single state" or "multiple state" marking.
  *
- * "single state" Marking means a subject can be in one and only one state at
- * the same time. Use it with state machine.
+ * "single state" marking means a subject can be in one and only one state at
+ * the same time. Use it with state machine. It uses a string to store the
+ * marking.
  *
- * "multiple state" Marking means a subject can be in many states at the same
- * time. Use it with workflow.
+ * "multiple state" marking means a subject can be in many states at the same
+ * time. Use it with workflow. It uses an array of strings to store the marking.
  *
  * @author Grégoire Pineau <lyrixx@lyrixx.info>
  */
 final class MethodMarkingStore implements MarkingStoreInterface
 {
-    private bool $singleState;
-    private string $property;
+    /** @var array<class-string, MarkingStoreMethod> */
+    private array $getters = [];
+    /** @var array<class-string, MarkingStoreMethod> */
+    private array $setters = [];
 
     /**
-     * @param string $property Used to determine methods to call
-     *                         The `getMarking` method will use `$subject->getProperty()`
-     *                         The `setMarking` method will use `$subject->setProperty(string|array $places, array $context = array())`
+     * @param string $property Used to determine methods or property to call
+     *                         The `getMarking` method will use `$subject->getProperty()` or `$subject->property`
+     *                         The `setMarking` method will use `$subject->setProperty(string|array $places, array $context = [])` or `$subject->property = string|array $places`
      */
-    public function __construct(bool $singleState = false, string $property = 'marking')
-    {
-        $this->singleState = $singleState;
-        $this->property = $property;
+    public function __construct(
+        private bool $singleState = false,
+        private string $property = 'marking',
+    ) {
     }
 
     public function getMarking(object $subject): Marking
     {
-        $method = 'get'.ucfirst($this->property);
-
-        if (!method_exists($subject, $method)) {
-            throw new LogicException(sprintf('The method "%s::%s()" does not exist.', get_debug_type($subject), $method));
-        }
-
         $marking = null;
         try {
-            $marking = $subject->{$method}();
+            $marking = ($this->getGetter($subject))();
         } catch (\Error $e) {
             $unInitializedPropertyMessage = sprintf('Typed property %s::$%s must not be accessed before initialization', get_debug_type($subject), $this->property);
             if ($e->getMessage() !== $unInitializedPropertyMessage) {
@@ -68,7 +66,7 @@ final class MethodMarkingStore implements MarkingStoreInterface
         if ($this->singleState) {
             $marking = [(string) $marking => 1];
         } elseif (!\is_array($marking)) {
-            throw new LogicException(sprintf('The method "%s::%s()" did not return an array and the Workflow\'s Marking store is instantiated with $singleState=false.', get_debug_type($subject), $method));
+            throw new LogicException(sprintf('The marking stored in "%s::$%s" is not an array and the Workflow\'s Marking store is instantiated with $singleState=false.', get_debug_type($subject), $this->property));
         }
 
         return new Marking($marking);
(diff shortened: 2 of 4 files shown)


src/Symfony/Bundle/FrameworkBundle/Resources/config/schema/symfony-1.0.xsd
```diff
@@ -443,13 +443,13 @@
         <xsd:attribute name="type" type="marking_store_type" />
         <xsd:attribute name="service" type="xsd:string" />
         <xsd:attribute name="property" type="xsd:string" />
+        <xsd:attribute name="context_property" type="xsd:string" />
     </xsd:complexType>
 
     <xsd:simpleType name="marking_store_type">
         <xsd:restriction base="xsd:string">
-            <xsd:enumeration value="multiple_state" />
-            <xsd:enumeration value="single_state" />
```
lyrixx: Dead code for ages

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/xml/workflows.xml
```diff
@@ -10,7 +10,6 @@
         <framework:workflow name="article" type="workflow">
             <framework:audit-trail enabled="true"/>
             <framework:initial-marking>draft</framework:initial-marking>
-            <framework:marking-store type="method" property="state" />
```
lyrixx: this is already the default value. I removed it to be iso with PHP, and YAML

src/Symfony/Component/Workflow/Tests/MarkingStore/MethodMarkingStoreTest.php
```diff
@@ -26,14 +25,14 @@ public function testGetSetMarkingWithMultipleState()
 
         $marking = $markingStore->getMarking($subject);
 
-        $this->assertInstanceOf(Marking::class, $marking);
```
lyrixx: This is useless, because enforced by the interface typing

nicolas-grekas:
It'd be great to not have to configure anything to store using either a property or a method. We're able to do so in the form component. Maybe add an ObjectMarkingStore that relies on property-access? Or if we really want to have standalone logic, implement all this in one single class that can do both?

lyrixx:
It cannot work since we have two things : the marking and a context. The property access component can only work with one value 

So we need to recode everything. Add some cache etc. I'm not sure it really worth it. And anyway, we cannot guess the context property so we Will need to have some convention or configuration anyway

src/Symfony/Component/Workflow/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ CHANGELOG
 
  * Add `with-metadata` option to the `workflow:dump` command to include places,
    transitions and workflow's metadata into dumped graph
+ * Add `PropertiesMarkingStore` to store marking directly to the suject's
```
pbowyer:
```suggestion
 * Add `PropertiesMarkingStore` to store marking directly to the subject's
```

lyrixx: Thanks, fixed.

nicolas-grekas:
What about something like that (no need for the "type" option when property is set):

```yaml
marking_store:
    property: foo
```

Then:
- we use `getFoo()` if it exists, and we use `$obj->foo` otherwise (one of them must be avail).
- we use `setFoo()` if it exists, and we use `$obj->foo = $state` otherwise (one of them must be avail).

If people want to get the context, we offer only one option: implementing the setter. This enforces an API that legitimately correlates state+context.

lyrixx:
It make sens to me 👍🏼 we can check everything during the compilation

lyrixx:
I pushed a whole new implementation. There is no change in the bundle configuration.

The current MethodMarkingStore has been updated to support methods or property

src/Symfony/Component/Workflow/MarkingStore/MethodMarkingStore.php
```diff
@@ -15,45 +15,42 @@
...
 /**
- * MethodMarkingStore stores the marking with a subject's method.
+ * MethodMarkingStore stores the marking with a subject's public method if exist,
+ * then to a property.
  *
  * This store deals with a "single state" or "multiple state" Marking.
  *
  * "single state" Marking means a subject can be in one and only one state at
- * the same time. Use it with state machine.
+ * the same time. Use it with state machine. It uses a string to store the marking
```
nicolas-grekas:
```suggestion
 * the same time. Use it with state machine. It uses a string to store the marking.
```

src/Symfony/Component/Workflow/MarkingStore/MethodMarkingStore.php
```diff
@@ -15,45 +15,42 @@
...
  *
  * "multiple state" Marking means a subject can be in many states at the same
- * time. Use it with workflow.
+ * time. Use it with workflow. It uses an array to store the marking
```
nicolas-grekas:
```suggestion
 * time. Use it with workflow. It uses an array of strings to store the marking.
```

src/Symfony/Component/Workflow/MarkingStore/MethodMarkingStore.php
```diff
@@ -15,45 +15,42 @@
...
  *
  * @author Grégoire Pineau <lyrixx@lyrixx.info>
  */
 final class MethodMarkingStore implements MarkingStoreInterface
 {
-    private bool $singleState;
-    private string $property;
+    private readonly \SplObjectStorage $getters;
+    private readonly \SplObjectStorage $setters;
```
nicolas-grekas: we should use an array indexed by class instead, so that we don't keep subjects in memory
the closures you want should take the subject as argument to make this work

src/Symfony/Component/Workflow/MarkingStore/MethodMarkingStore.php
```diff
@@ -82,12 +79,64 @@ public function setMarking(object $subject, Marking $marking, array $context = [
...
+            }
+
+            if (!$r->isPublic()) {
+                throw new LogicException(sprintf('The public method "%1$s::%3$s()" must be declared, or the property "%1$s::%2$s" must be public.', get_debug_type($subject), $propertyName, $method));
+            }
+
+            return MarkingStoreMethod::PROPERTY;
+        }
+
+        if (!$r->isPublic()) {
```
nicolas-grekas: if the method is not public but the property is, maybe we should use the property instead of throwing?

src/Symfony/Component/Workflow/MarkingStore/MethodMarkingStore.php
lyrixx: I don't store the callable anymore, so the change is really simple!

lyrixx:
Thanks for your review. I addressed your comments

nicolas-grekas (approved):

src/Symfony/Component/Workflow/MarkingStore/MethodMarkingStore.php
```diff
@@ -82,12 +80,50 @@ public function setMarking(object $subject, Marking $marking, array $context = [
...
+            }
+        } catch (\ReflectionException) {
         }
 
-        $subject->{$method}($marking, $context);
+        throw new LogicException(sprintf('Cannot store marking: class "%s" should have either a public method named "%s()" or a public property named "$%s"; none found.', get_debug_type($subject), $method, $property));
     }
 }
+
+enum MarkingStoreMethod
```
stof, line 128: should be tagged as `@internal` (even if not autoloadable through PSR-4, IDEs will still discover this enum and suggest it in completion otherwise)

stof (changes requested):

src/Symfony/Component/Workflow/MarkingStore/MethodMarkingStore.php
lyrixx: Fixed, thanks

src/Symfony/Component/Workflow/MarkingStore/MethodMarkingStore.php
```diff
@@ -15,45 +15,43 @@
...
-    private string $property;
+    /** @var array<class-string, MarkingStoreMethod> */
+    private array $getters = [];
+    /** @var array<class-string, MarkingStoreMethod> */
+    private array $setters = [];
 
     /**
      * @param string $property Used to determine methods to call
      *                         The `getMarking` method will use `$subject->getProperty()`
      *                         The `setMarking` method will use `$subject->setProperty(string|array $places, array $context = array())`
```
stof: this should probably be updated

lyrixx: Good catch. Fixed!

fabpot (approved):

fabpot:
Thank you @lyrixx.

PR #50980, opened 2023-07-14
[Messenger] Make `HandleMessageMiddleware::callHandler` protected

ruudk:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      |no
| New feature?  | yes
| Deprecations? | no 
| Tickets       | 
| License       | MIT
| Doc PR        | 

I made the following changes:
- make `callHandler` protected
- pass the HandlerDescriptor (can be used to read options from the handler)
- pass the Envelope (can be used to read more stamps)

This allows people to decorate the `HandleMessageMiddleware` and intercept the moment the handler is called.

I currently have the following use cases for this in my application:

1) Every handler can decide if they want to wrap their handling in a transaction. This is done by
   a custom `#[WrapInTransaction]` attribute. When the attribute is there, we wrap it in a database
   transaction. I know the `DoctrineTransactionMiddleware` exists in the Symfony Doctrine bridge,
   but that runs as middleware and cannot be toggled individually. It would be possible to modify it
   so that it only runs when the handler has it enabled using an option, but that only works for
   buses with only 1 handler. For an event bus, with multiple handlers, that won't be a solution.

2) Every handler can decide if they want to ignore certain exceptions that are thrown. This is done
   by a custom `#[IgnoreException]` attribute. For example the handler uses
   `#[IgnoreException(EventNotFoundException::class)]`. Whenever that exception is thrown, the
   exception is caught and ignored. The middleware thinks it's executed without issues.

### DIFF
--- a/src/Symfony/Component/Messenger/CHANGELOG.md
+++ b/src/Symfony/Component/Messenger/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+ * Make `HandleMessageMiddleware::callHandler` protected
+
 6.3
 ---
 
--- a/src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
+++ b/src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
@@ -79,7 +79,7 @@ class HandleMessageMiddleware implements MiddlewareInterface
                         $ackStamp->ack($envelope, $e);
                     });
 
-                    $result = $this->callHandler($handler, $message, $ack, $envelope->last(HandlerArgumentsStamp::class));
+                    $result = $this->callHandler($handlerDescriptor, $envelope, $ack);
 
                     if (!\is_int($result) || 0 > $result) {
                         throw new LogicException(sprintf('A handler implementing BatchHandlerInterface must return the size of the current batch as a positive integer, "%s" returned from "%s".', \is_int($result) ? $result : get_debug_type($result), get_debug_type($batchHandler)));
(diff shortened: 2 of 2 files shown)


ro0NL:
1. can be solved with a custom middleware isnt it?
2. why not try/catch where needed?

ruudk:
Sure, this can be solved by a custom middleware. But that means that I will be replacing a big part of Messenger, namely the way it handles messages: [HandleMessageMiddleware.php](https://github.com/symfony/symfony/pull/50980#diff-5f6379354b24ad7db8292cc3899c3190b71d87abeab9f57f9fe4a9ee5a7e6eea).

That requires me to keep my "fork" middleware, sync with every release.

What I want is an extension point, where it passes a message to more than 1 handlers.

My use case: I want to add extra functionality for event subscribers. I want to know which subscriber failed, and which didn't. Or I want to wrap every subscriber inside its own database transaction. 

Currently, not possible as we can only wrap ALL handlers.

ruudk:
A try/catch inside the subscribers will work, but that requires putting this in place in 1000's of subscribers. Or it requires me to create an abstract subscriber that all subscribers need to extend from.

I want to handle this in a higher level. Abstracted away from what developers will see in their day to day.

ro0NL:
oh i get it, envelope has many handlers.

perhaps the better approach is to dispatch [an event](https://symfony.com/doc/current/messenger.html#messenger-events)

ruudk:
@ro0NL Interesting. I checked it but I don't see how that would work. Or how that could help my use case. 

I really think there should be an extension point for the execution of the handler. 

Or, in an ideal world, there is a dedicated middleware stacks per handler. That would allow wrapping handlers isolated, instead of the current situation where multiple handlers _can_ be called.

ro0NL:
> enabling wrapping the handler callable

so, `$event->setHandler(fn ()=> $prevHandler(...func_get_args()))`

ro0NL:
see #52425 :)

jerowork:
Bringing back to the attention 😎 .

If we would make the `callHandler` protected, it allows to extend on the `HandleMessageMiddleware`, which avoids maintaining own versions of the full middleware. It makes the Messenger more flexible for extension :).
 
At the same time, the risk/impact of such a change (private to protected) is limited, fairly safe to do so.

Is this still an option to agree on and merge? 🙏

PR #50985, opened 2023-07-14
[DependencyInjection] Fix fetching lazy non-shared services multiple times

### LINKED ISSUE #50975: Fetching of lazy non-shared services sometimes broken

6.3.*

### Description

This is a close-to-absurd edge case, but still it occurred in my project so I report it.

When fetching a service from the service container that is public, non-shared, and lazy, twice (!), the following exception is thrown:

```
Symfony\Component\DependencyInjection\Container::get(): Return value must be of type ?object, string returned
```

On Symfony 6.2 or earlier versions, this exception was not thrown. It also doesn't seem to occur if the service is injected in a constructor instead of being pulled from the container.

Thank you for a look!

### How to reproduce

Define a service as follows:

```yaml
App\Service\SomeService:
    public: true
    shared: false
    lazy: true
```

Then fetch it from the service container twice:

```php
$this->container->get(SomeService::class);
$this->container->get(SomeService::class);
```

### Possible Solution

_No response_

### Additional Context

The problem occurs in the container's factory method for `SomeService`. On the second call, `$lazyLoad` contains the service ID instead of the service instance.

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50975
| License       | MIT
| Doc PR        | -

An attempt to fix #50975.


nicolas-grekas:
Thanks for having a look. I checked why the id could come as first argument and I think the fix we need is this one:

```diff
--- a/src/Symfony/Component/DependencyInjection/Dumper/PhpDumper.php
+++ b/src/Symfony/Component/DependencyInjection/Dumper/PhpDumper.php
@@ -900,6 +900,8 @@ EOF;
 
                     if ($asFile) {
                         $code .= "fn () => self::do(\$container);\n\n";
+                    } elseif ($definition->isPublic()) {
+                        $code .= sprintf("fn() => \$this->%s();\n\n", $methodName);
                     } else {
                         $code .= sprintf("\$this->%s(...);\n\n", $methodName);
                     }
```

WDYT?

nicolas-grekas:
Thank you @HypeMC.

HypeMC:
@nicolas-grekas There's one more case this doesn't cover, see #50996 .

PR #49665, opened 2023-03-10
[DependencyInjection] Add `constructor` option to `#[Autoconfigure]`

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | _NA_
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/18070

Following [this discussion](https://twitter.com/nicolasgrekas/status/1634259754104025088?s=20) on Twitter, here is my try to use service factories with attributes. This PR adds the `constructor` option to the `Autoconfigure` attribute, and more generally, the `constructor` option on service definitions.

This allows to write services using a factory this way:

```php
#[Autoconfigure(constructor: 'create')]
class FactoryAttributeService
{
    public function __construct(private readonly string $bar)
    {
    }

    public function getBar(): string
    {
        return $this->bar;
    }

    public static function create(string $foo = 'foo'): static
    {
        return new self($foo);
    }
}
```

PR #20943, opened 2016-12-15
[DependencyInjection] Use current class as default class for factory declarations

ogizanagi:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #20888
| License       | MIT
| Doc PR        | Should update the notice about the "class" attribute on http://symfony.com/doc/current/service_container/factories.html

#20888 makes sense to me, considering the following sample extracted from the documentation:

```xml
<service id="app.newsletter_manager" class="AppBundle\Email\NewsletterManager">
    <factory class="AppBundle\Email\NewsletterManager" method="create" />
</service>
```

The class is used as a factory to create itself, thus it can be simplified to:

```xml
<service id="app.newsletter_manager" class="AppBundle\Email\NewsletterManager">
    <factory method="create" />
</service>
```

However, it's not possible to provide the same feature for the YAML format, because it doesn't allow to distinct a function from a method call if the class is not provided explicitly under the `factory` key, whereas the xml format use a dedicated `function` attribute.
Would this inconsistency between those two formats be a no-go for this feature?

The doc notices:
> When using a factory to create services, the value chosen for the class option has no effect on the resulting service. The actual class name only depends on the object that is returned by the factory. However, the configured class name may be used by compiler passes and therefore should be set to a sensible value.

If this is merged, it should be updated wisely in order to not confuse everyone with this feature when using the xml format.

UPDATE: The yaml format is now supported when the class is not provided in the factory array:

```yml
services:
    my_factory:
        class: Bar\Baz
        factory: [~, 'create']
```

PR #50988, opened 2023-07-15
[FrameworkBundle] Run the `ResolveFactoryClassPass` when `lint:container` builds the container from a dump

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50622
| License       | MIT
| Doc PR        | N/A

#49665 replaced the `factory` node by a `constructor` attribute in the XML and YAML dumper when the factory’s class is the same as the definition’s. The corresponding loader then creates a definition where the factory class is `null`.

As the `ResolveFactoryClassPass` will not run when the `lint:container` command builds the container from an XML dump, such factories would make `ContainerBuilder::createService` crash. This PR adds this compiler pass to the list.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Command/ContainerLintCommand.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Command/ContainerLintCommand.php
@@ -21,6 +21,7 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\DependencyInjection\Compiler\CheckTypeDeclarationsPass;
 use Symfony\Component\DependencyInjection\Compiler\PassConfig;
+use Symfony\Component\DependencyInjection\Compiler\ResolveFactoryClassPass;
 use Symfony\Component\DependencyInjection\Container;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
@@ -99,7 +100,7 @@ final class ContainerLintCommand extends Command
             $refl->setValue($parameterBag, true);
 
             $container->getCompilerPassConfig()->setBeforeOptimizationPasses([]);
-            $container->getCompilerPassConfig()->setOptimizationPasses([]);
+            $container->getCompilerPassConfig()->setOptimizationPasses([new ResolveFactoryClassPass()]);
             $container->getCompilerPassConfig()->setBeforeRemovingPasses([]);
         }
 


lugosium:
I confirm your PR fix the issue 👍 🙏 

Thanks a lot.

nicolas-grekas:
Did you consider enabling the pass on the command? That'd make more sense to me because if we do this, then we should do the same in PHP dumper, and the pass becomes useless. Also, this misses the check for $class that the pass has. WDYT ?

MatTheCat:
TBH I didn’t consider updating the command because it looks like it must work without any compiler pass.

> then we should do the same in PHP dumper

I may have missed something because I don’t understand why :thinking:

nicolas-grekas:
> it looks like it must work without any compiler pass.

It never did before we introduced ResolveFactoryClassPass so I don't think so.

MatTheCat:
Is this what you’re asking for?

```diff
diff --git a/src/Symfony/Bundle/FrameworkBundle/Command/ContainerLintCommand.php b/src/Symfony/Bundle/FrameworkBundle/Command/ContainerLintCommand.php
--- a/src/Symfony/Bundle/FrameworkBundle/Command/ContainerLintCommand.php	(revision 6f2e603c176724873f92942ea7461150d0e32040)
+++ b/src/Symfony/Bundle/FrameworkBundle/Command/ContainerLintCommand.php	(date 1689424425782)
@@ -20,6 +20,7 @@
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\DependencyInjection\Compiler\CheckTypeDeclarationsPass;
 use Symfony\Component\DependencyInjection\Compiler\PassConfig;
+use Symfony\Component\DependencyInjection\Compiler\ResolveFactoryClassPass;
 use Symfony\Component\DependencyInjection\Container;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
@@ -120,7 +121,7 @@
             }
 
             $container->getCompilerPassConfig()->setBeforeOptimizationPasses([]);
-            $container->getCompilerPassConfig()->setOptimizationPasses([]);
+            $container->getCompilerPassConfig()->setOptimizationPasses([new ResolveFactoryClassPass()]);
             $container->getCompilerPassConfig()->setBeforeRemovingPasses([]);
         }
 
```

It feels weird needing this particular pass to build the container from a dump.

MatTheCat:
Changed the target branch to 5.4 because the issue actually comes from #20943; #49665 just make it easier to trigger.

nicolas-grekas:
Yes that's what I mean. Dealing with th dumped XML requires special care, this makes sense to me. At least more than making the pass half-needeed. I prefer keeping the symmetry between the dumped container and a compiled container builder.

MatTheCat:
Okay PR updated, thanks for the feedback :pray:

nicolas-grekas (approved):

nicolas-grekas:
Thank you @MatTheCat.

PR #50993, opened 2023-07-16
[HttpKernel] Customise error code on validation error occurred in #[MapRequestPayload]

### LINKED ISSUE #50992: [HttpKernel] Allow customising HTTP status code on validation error of #[MapRequestPayload] objects
### Description

`#[MapRequestPayload]` attribute, introduced in Symfony 6.3, is an awesome feature for APIs that use JSON for requests. The attribute works with Serializer and Validator components, and returns `422 Unprocessable Entity` status code on Validator returned any violation(s). However, in API development, sometimes we want `400 Bad Request` instead of 422. So it would be great if we can customise the status code that MapRequestPayload returns on validation error.

### Option 1. Use attribute to customise for each endpoint

```php
#[Route('/pets', methods: ['POST'])]
public function createPet(
    #[MapRequestPayload(statusOnError: Response::HTTP_BAD_REQUEST) PetDto $pet,
): Response;
```

### Option 2. Use config to customise globally

```yaml
framework:
    map_request_payload:
        default_status_code_on_error: 400
```

siketyan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #50992 
| License       | MIT
| Doc PR        | TBD

This pull request proposes to allow customising HTTP status code that will be returned from `#[MapRequestPayload]` value resolver.

Users can customise by two methods:

1. **Use the argument in the attribute**
   ```php
   #[MapRequestPayload(statusCodeOnError: Response::HTTP_BAD_REQUEST)]
   ```

2. **Use framework configuration for global**
   ```yaml
   framework:
       map_request_payload:
           status_code_on_error: 400
   ```

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
@@ -139,6 +139,15 @@ class Configuration implements ConfigurationInterface
                     ->defaultValue('error_controller')
                 ->end()
                 ->booleanNode('handle_all_throwables')->info('HttpKernel will handle all kinds of \Throwable')->end()
+                ->arrayNode('map_request_payload')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->scalarNode('status_code_on_error')
+                            ->info('HTTP status code to be returned on error occurred while parsing the request payload.')
+                            ->defaultNull()
+                        ->end()
+                    ->end()
+                ->end()
             ->end()
         ;
 
--- a/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
+++ b/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * Add native return type to `Translator` and to `Application::reset()`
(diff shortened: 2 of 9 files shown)


src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -139,6 +139,15 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->defaultValue('error_controller')
                 ->end()
                 ->booleanNode('handle_all_throwables')->info('HttpKernel will handle all kinds of \Throwable')->end()
+                ->arrayNode('map_request_payload')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->scalarNode('status_code_on_error')
```
ro0NL, line 145: is there a usecase that requires a different default?

siketyan: As described in #50992, sometimes we want use 400 instead of 422 in application-global. So this configuration node supports to change the HTTP status code on error globally.

siketyan: BTW, this is actually dupe of #50767, so I'll close this pull request now.

PR #48469, opened 2022-12-04
[DependencyInjection] Use WeakReference to break circular references in the container

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

While working on #48461, I realized that we could break many circular loops involving the container by using a `WeakReference` to access it. This happens when we generate closures.

Conceptually, we generate this at the moment:
```php
$internalClosure = function () { return $this->get('foo'); };
```

And we'd generate this with this PR:
```php
$containerRef = \WeakReference::create($this);
$internalClosure = static function () use ($containerRef) { return $containerRef->get()->get('foo'); };
```

This should reduce the pressure on the garbage collector.

PR #50996, opened 2023-07-16
[DependencyInjection] Fix fetching lazy non-shared services multiple times with as files `true`

### LINKED ISSUE #50975: Fetching of lazy non-shared services sometimes broken

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50975
| License       | MIT
| Doc PR        | -

Followup to #50985, fixes case when `as_files` is `true`.

This particular problem exists only on v6.3 and was introduced in #48469.


src/Symfony/Component/DependencyInjection/Tests/Dumper/PhpDumperTest.php
```diff
@@ -772,7 +754,7 @@ public function testNonSharedLazy()
         $container = new ContainerBuilder();
 
         $container
-            ->register('foo', Foo::class)
+            ->register('foo', \Bar\FooLazyClass::class)
```
HypeMC, line 757: A minor mistake I made in my original PR, doesn't really change anything, but it makes the test a bit more consistent.

src/Symfony/Component/DependencyInjection/Tests/Dumper/PhpDumperTest.php
```diff
@@ -803,6 +785,48 @@ public function testNonSharedLazy()
...
+        $stringDump = print_r($dumps, true);
+        $this->assertStringMatchesFormatFile(
+            self::$fixturesPath.'/php/services_non_shared_lazy_as_files.txt',
+            '\\' === \DIRECTORY_SEPARATOR ? str_replace("'.\\DIRECTORY_SEPARATOR.'", '/', $stringDump) : $stringDump
+        );
+
+        $lastDump = array_pop($dumps);
+        foreach (array_reverse($dumps) as $dump) {
+            eval('?>'.$dump);
+        }
```
HypeMC, line 817: Is there a better way to do this?

nicolas-grekas: Nah, that's fine

nicolas-grekas:
Thank you @HypeMC.

PR #50998, opened 2023-07-17
[Messenger] Introduce `HandlerDescriptorInterface`

ruudk:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Tickets       | 
| License       | MIT
| Doc PR        | 

The `HandlerDescriptor` class is final and therefore cannot be extended. To make this a bit more flexible we introduce the `HandlerDescriptorInterface` that is used everywhere.

It also introduces a `.messenger.handler_descriptor_class` parameter that's used in the MessengerPass. With this parameter it's possible to use a custom implementation of the `HandlersLocator`.

This is an alternative approach to https://github.com/symfony/symfony/pull/50980. That PR also describes reasonings for this.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/messenger.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/messenger.php
@@ -24,6 +24,8 @@ use Symfony\Component\Messenger\EventListener\SendFailedMessageToFailureTranspor
 use Symfony\Component\Messenger\EventListener\StopWorkerOnCustomStopExceptionListener;
 use Symfony\Component\Messenger\EventListener\StopWorkerOnRestartSignalListener;
 use Symfony\Component\Messenger\EventListener\StopWorkerOnSignalsListener;
+use Symfony\Component\Messenger\Handler\HandlerDescriptor;
+use Symfony\Component\Messenger\Handler\HandlersLocator;
 use Symfony\Component\Messenger\Handler\RedispatchMessageHandler;
 use Symfony\Component\Messenger\Middleware\AddBusNameStampMiddleware;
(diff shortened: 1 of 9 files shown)


ruudk:
Before I fix the failing tests, I'd like to know if this would be considered for merge or not.

ruudk:
@OskarStark what do you think about this?

derrabus:
> It also introduces a `.messenger.handler_descriptor_class` parameter that's used in the MessengerPass.

We've removed this kind of parameter from our codebase and I don't think we want to start this again. A big issue with that class parameter is that it only works if the custom class uses the exact same constructor signature as the default implementation.

> With this parameter it's possible to use a custom implementation of the `HandlersLocator`.

You could also just replace or decorate the service.

ruudk:
Ahh, didn't think of decorating! Will update the PR to remove the parameter.

ruudk:
@derrabus But how to decorate something like this:
https://github.com/symfony/symfony/blob/bb7a7dd333dba1a88bd7992e1c5e4bd1d28000b1/src/Symfony/Component/Messenger/DependencyInjection/MessengerPass.php#L166-L169

The only way I know, is to walk through the container definitions and find anything that starts with `.messenger.handler_descriptor`?

ruudk:
@nicolas-grekas What do you think of this?

OskarStark:
friendly ping @derrabus

ro0NL:
im not convinced HandlerDescription should be extensible

HandlersLocatorInterface::getHandlers will give you the descriptors, that should be sufficient for #50980 isnt it?

PR #51000, opened 2023-07-17
[WebProfilerBundle] Fix error in case of 'Content-Type' set null in dev environment with no debug

### LINKED ISSUE #50977: [WebProfiler] Error "Deprecated: str_contains(): Passing null to parameter #1 ($haystack) of type string ..." in dev environment with no debug

6.*

### Description

PHP version: >=8.1
Symfony version: (probably) any with dependes on PHP>=8.1

Sent HTTP request to my service with no header `Content-type`. Testing in **dev** (or **test**) mode with **debug** disabled. 

And have error `Deprecated: str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated in /var/www/vendor/symfony/web-profiler-bundle/EventListener/WebDebugToolbarListener.php on line 124`

### How to reproduce

- Set environments:
```
APP_ENV=dev
APP_DEBUG=false
 ```
- `WebProfilerBundle` enabled and collects data 
- Send HTTP request to any endpoint with no `Content-type` header

### Possible Solution

Its working for me

```
--- a/vendor/symfony/web-profiler-bundle/EventListener/WebDebugToolbarListener.php	
+++ b/vendor/symfony/web-profiler-bundle/EventListener/WebDebugToolbarListener.php
@@ -121,7 +121,7 @@
         if (self::DISABLED === $this->mode
             || !$response->headers->has('X-Debug-Token')
             || $response->isRedirection()
-            || ($response->headers->has('Content-Type') && !str_contains($response->headers->get('Content-Type'), 'html'))
+            || ($response->headers->has('Content-Type') && !str_contains($response->headers->get('Content-Type') ?? '', 'html'))
             || 'html' !== $request->getRequestFormat()
             || false !== stripos($response->headers->get('Content-Disposition', ''), 'attachment;')
         ) {
```

But unfortunly not working
```
            || ($response->headers->has('Content-Type') && !str_contains($response->headers->get('Content-Type', ''), 'html'))
```

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by alexbuyanow**
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50977
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Symfony/Bundle/WebProfilerBundle/EventListener/WebDebugToolbarListener.php
+++ b/src/Symfony/Bundle/WebProfilerBundle/EventListener/WebDebugToolbarListener.php
@@ -121,7 +121,7 @@ class WebDebugToolbarListener implements EventSubscriberInterface
         if (self::DISABLED === $this->mode
             || !$response->headers->has('X-Debug-Token')
             || $response->isRedirection()
-            || ($response->headers->has('Content-Type') && !str_contains($response->headers->get('Content-Type'), 'html'))
+            || ($response->headers->has('Content-Type') && !str_contains($response->headers->get('Content-Type') ?? '', 'html'))
             || 'html' !== $request->getRequestFormat()
             || false !== stripos($response->headers->get('Content-Disposition', ''), 'attachment;')
         ) {
--- a/src/Symfony/Bundle/WebProfilerBundle/Tests/EventListener/WebDebugToolbarListenerTest.php
+++ b/src/Symfony/Bundle/WebProfilerBundle/Tests/EventListener/WebDebugToolbarListenerTest.php
@@ -344,6 +344,20 @@ class WebDebugToolbarListenerTest extends TestCase
         $this->assertEquals("<html><head></head><body>\nWDT\n</body></html>", $response->getContent());
     }
 
+    public function testNullContentTypeWithNoDebugEnv()
+    {
+        $response = new Response('<html><head></head><body></body></html>');
+        $response->headers->set('Content-Type', null);
+        $response->headers->set('X-Debug-Token', 'xxxxxxxx');
+
+        $event = new ResponseEvent($this->createMock(Kernel::class), new Request(), HttpKernelInterface::MAIN_REQUEST, $response);
+
+        $listener = new WebDebugToolbarListener($this->getTwigMock(), false, WebDebugToolbarListener::ENABLED, null);
+        $listener->onKernelResponse($event);
+
+        $this->expectNotToPerformAssertions();
+    }
+
     protected function getTwigMock($render = 'WDT')
     {
         $templating = $this->createMock(Environment::class);

### CODE REVIEW

File: src/Symfony/Bundle/WebProfilerBundle/EventListener/WebDebugToolbarListener.php [hunk 1]
```diff
@@ -121,7 +121,7 @@ public function onKernelResponse(ResponseEvent $event)
         if (self::DISABLED === $this->mode
             || !$response->headers->has('X-Debug-Token')
             || $response->isRedirection()
-            || ($response->headers->has('Content-Type') && !str_contains($response->headers->get('Content-Type'), 'html'))
+            || ($response->headers->has('Content-Type') && !str_contains($response->headers->get('Content-Type') ?? '', 'html'))
```
fancyweb, line 124:
```suggestion
            || !str_contains($response->headers->get('Content-Type') ?? '', 'html')
```
We could remove the call to has I guess?

alexbuyanow: I guess not.  `has` checks header presence. If it absent, next condition (`get`) is meaningless

If header absent, will check `has` (`array_key_exists`) only

fancyweb: Yes but `get` returns `null` if the header is not present :man_shrugging:

alexbuyanow: just fewer operators

- with `has`
-- header absent => `has`
-- header present => `has` + `get` + coalesce + `str_contains` 
- with no `has`
-- header absent => `has` + `get` + coalesce + `str_contains` 
-- header present => `has` + `get` + coalesce + `str_contains` 

programmaticly it should be same

if not important we could remove

doing it?

src/Symfony/Bundle/WebProfilerBundle/Tests/EventListener/WebDebugToolbarListenerTest.php
```diff
@@ -344,6 +344,20 @@ public function testCspIsKeptEnabledIfDumperWasNotUsed()
...
+        $response = new Response('<html><head></head><body></body></html>');
+        $response->headers->set('Content-Type', null);
+        $response->headers->set('X-Debug-Token', 'xxxxxxxx');
+
+        $event = new ResponseEvent($this->createMock(Kernel::class), new Request(), HttpKernelInterface::MAIN_REQUEST, $response);
+
+        $listener = new WebDebugToolbarListener($this->getTwigMock(), false, WebDebugToolbarListener::ENABLED, null);
+        $listener->onKernelResponse($event);
+
+        $this->assertNull($response->headers->get('Content-Type'));
```
fancyweb:
```suggestion
        $this->assertSame('<html><head></head><body></body></html>', $response->getContent());
```
Better test than the toolbar is not injected?

src/Symfony/Bundle/WebProfilerBundle/EventListener/WebDebugToolbarListener.php
fancyweb: Ok it doesn't really matter we can keep it like this.

src/Symfony/Bundle/WebProfilerBundle/Tests/EventListener/WebDebugToolbarListenerTest.php
alexbuyanow: I think every unit test should be atomic. This one checks header behavior only

Сould make one another test for toolbar check if its really necessary

Something like this
```php
    public function testToolbarWithNullContentTypeWithNoDebugEnv()
    {
        $response = new Response('<html><head></head><body></body></html>');
        $response->headers->set('Content-Type', null);
        $response->headers->set('X-Debug-Token', 'xxxxxxxx');

        $event = new ResponseEvent($this->createMock(Kernel::class), new Request(), HttpKernelInterface::MAIN_REQUEST, $response);

        $listener = new WebDebugToolbarListener($this->getTwigMock(), false, WebDebugToolbarListener::ENABLED, null);
        $listener->onKernelResponse($event);

        $this->assertSame('<html><head></head><body></body></html>', $response->getContent());
    }
```

fancyweb: The point is that `$this->assertNull($response->headers->get('Content-Type'));` can never fail since the test set it itself and the code has no intent of modifying it.
Anyway, the point of the test is to ensure there's no warning or errors, so maybe just replace the assertion with `$this->expectNotToPerformAssertions();`?

fancyweb (approved):

nicolas-grekas:
Thank you @alexbuyanow.

PR #51003, opened 2023-07-17
[Serializer] Handle type error constructing input

### LINKED ISSUE #50904: MapQueryString & MapRequestPayload empty when type missmatch

6.3.1

### Description

When there is a type missmatch on an objet with the attribute `MapQueryString` or `MapRequestPayload` an empty instance is created.

For instance if I use `MapQueryString` expecting a name (string) and age (int) but the user send a "John" and "foo", I end up with an empty object.

### How to reproduce

1. Create a simple DTO (I deliberately omitted NotBlank)

```php
final readonly class PayloadDTO
{

    public function __construct(
        #[Assert\Length(min: 3)]
        public string $name,

        #[Assert\Positive]
        public int    $age
    )
    {
    }

}
``` 

2. Map this DTO in a controller

```php
class HomeController extends AbstractController
{
    #[Route('/', name: 'home')]
    public function index(
        #[MapQueryString] PayloadDTO $payload,
    ): Response
    {
        dd($payload); // I expect the dto to be valid (with an age / name property)
    }

}
```

3. Try to access the page using an "age" that is not a number. `http://localhost:8000/?name=John&age=foo`, you'll end up with an empty `PayloadDTO` (which will cause bugs further down the application).

### Possible Solution

I tracked down the issue to the `AbstractNormalizer` that ends up instantiating the object without using the constructor (using the reflectionClass)  since there is a `not_normalizable_value_exceptions` in the context. https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php#L424-L428

A solution would be to check the context for "collect_denormalization_errors" before skipping the exception or to collect the exception when it happens.

```php
            try {
                return $reflectionClass->newInstanceArgs($params);
            } catch (\TypeError $e) {
                if (!isset($context['not_normalizable_value_exceptions'])) {
                    throw $e;
                }

                $context['not_normalizable_value_exceptions'][] = NotNormalizableValueException::createForUnexpectedDataType(/* Find which parameters to give */);

                return $reflectionClass->newInstanceWithoutConstructor();
            }
```

kylekatarnls:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues? | Fix #50904
| License       | MIT

When mapping a input to an object, for instance with:
```php
public function myApiRoute(#[MapQueryString] MyInput $input)
{
}
```

And `MyInput` has an hard-type:
```php
class MyInput
{
    public function __construct(
        public readonly ?string $id = null,
    ) {
    }
}
```

Then you try to call this route with invalid type: `/api/endpoint?id[]=abc`

The construction of `MyInput` will fail with a `TypeError` `MyInput::__construct(): Argument #1 ($id) must be of type ?string, array given`.

But this is caught in `AbstractNormalizer` then you get a 500 error (unhandled) `Typed property MyInput::id must not be accessed before initialization`.

With this PR, I propose to handle it instead as a 400 error  (bad request) `The type of the "id" parameter for class "MyInput" must be of type "?string" ("array" given).`

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
@@ -425,6 +425,28 @@ abstract class AbstractNormalizer implements NormalizerInterface, DenormalizerIn
                     throw $e;
                 }
 
+                $message = $e->getMessage();
+
+                if (preg_match('/#\d+ \(\$([^)]+)\) must be of type ([^,]+), ([^,]+) given/', $message, $match)) {
+                    $message = sprintf(
+                        'The type of the "%s" parameter for class "%s" must be of type "%s" ("%s" given).',
+                        $match[1],
+                        $class,
+                        $match[2],
+                        $match[3],
+                    );
+                }
+
+                $exception = NotNormalizableValueException::createForUnexpectedDataType(
+                    $message,
+                    $data,
+                    [$class],
+                    $context['deserialization_path'] ?? null,
+                    true,
+                    previous: $e,
+                );
+                $context['not_normalizable_value_exceptions'][] = $exception;
+
                 return $reflectionClass->newInstanceWithoutConstructor();
             }
         }
--- a/src/Symfony/Component/Serializer/Tests/SerializerTest.php
+++ b/src/Symfony/Component/Serializer/Tests/SerializerTest.php
@@ -1021,6 +1021,15 @@ class SerializerTest extends TestCase
                 'useMessageForUser' => false,
                 'message' => 'The type of the "something" attribute for class "Symfony\Component\Serializer\Tests\Fixtures\Php74FullWithTypedConstructor" must be one of "float" ("string" given).',
             ],
+            [
+                'currentType' => 'array',
+                'expectedTypes' => [
+                    'Symfony\Component\Serializer\Tests\Fixtures\Php74FullWithTypedConstructor',
+                ],
+                'path' => 'php74FullWithTypedConstructor',
+                'useMessageForUser' => true,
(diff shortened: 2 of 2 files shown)


src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
```diff
@@ -425,6 +425,28 @@ protected function instantiateObject(array &$data, string $class, array &$contex
...
+                }
+
+                $exception = NotNormalizableValueException::createForUnexpectedDataType(
+                    $message,
+                    $data,
+                    [$class],
+                    $context['deserialization_path'] ?? null,
+                    true,
+                    previous: $e,
+                );
```
Victor-Truhanovich, line 447: Perhaps it would be better this way
```suggestion
                $type = $class;
                
                if (preg_match('/#\d+ \(\$([^)]+)\) must be of type ([^,]+), ([^,]+) given/', $message, $match)) {
                    $message = sprintf(
                        'The type of the "%s" parameter for class "%s" must be of type "%s" ("%s" given).',
                        $match[1],
                        $class,
                        $match[2],
                        $match[3],
                    );
                    
                    $type = $match[2];
                }

                $exception = NotNormalizableValueException::createForUnexpectedDataType(
                    $message,
                    $data,
                    [$type],
                    $context['deserialization_path'] ?? null,
                    true,
                    previous: $e,
                );
```

nicolas-grekas:
Thanks, and sorry for the long wait. The underlying problem from #50904 was fixed on 8.2 by #63979: `AbstractNormalizer` now records a `NotNormalizableValueException` when the constructor throws a `TypeError` while denormalization errors are being collected, so the case turns into a 400 instead of a 500. Closing as superseded.

PR #51004, opened 2023-07-17
[HttpKernel] Support backed enums in `#[MapQueryParameter]`

### LINKED ISSUE #50910: Support backed enums in `#[MapQueryParameter]`

6.3.1

### Description

If you use #[MapQueryParameter] and type the property with Enum, the QueryParameterValueResolver only accepts scalars.

If I change the resolver to BackedEnumValueResolver, it only looks at the attributes and not the query.
The enum is never resolved and the value is null.

### How to reproduce

Controller:

````php
#[Route('/', name: 'test_param_enum']
public function testParameterAction(#[MapQueryParameter] Status $status = null) 
{
        
}
````

OR

````php
#[Route('/', name: 'test_param_enum']
public function testParameterAction(#[MapQueryParameter(resolver: BackedEnumValueResolver::class)] Status $status = null) 
{
        
}
````

Enum:

````php
enum Status: string
{
    case New = 'new';
    case Published = 'published';
}
````

Query:

````
/?status=new
````

### Possible Solution

1.  Allow enum in QueryParameterValueResolver
2. Check for query parameters in BackedEnumValueResolver
3. Add a new resolver QueryParameterBackedEnumValueResolver
4. ???

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by andersmateusz**
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #50910
| License       | MIT
| Doc PR        | I think documentation about MapQueryParameter is missing

I had two options to introduce this feature. Extending `QueryParameterValueResolver` or extending `BackedEnumValueResolver`. Both options have it's prons and cons. Extending `QueryParameterValueResolver` is not consistent with DRY principle, but on the other hand does not mislead users about usage of `ValueResolverInterface` (`#[MapQueryParameter]`takes `resolver` argument but only `QueryParameterValueResolver` resolves value unless creating your own implementation of the interface).  I have chosen to extend  `QueryParameterValueResolver`. I think in the future separation of concerns should be introduced (resolving query params, resolving attributes etc.). For example by typing resolvers with new interface and using intersection types.

### DIFF
--- a/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php
+++ b/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php
@@ -18,8 +18,11 @@ use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
 /**
+ * Resolve arguments of type: array, string, int, float, bool, \BackedEnum from query parameters.
+ *
  * @author Ruud Kamphuis <ruud@ticketswap.com>
  * @author Nicolas Grekas <p@tchwork.com>
+ * @author Mateusz Anders <anders_mateusz@outlook.com>
  */
 final class QueryParameterValueResolver implements ValueResolverInterface
 {
@@ -39,8 +42,9 @@ final class QueryParameterValueResolver implements ValueResolverInterface
         }
 
         $value = $request->query->all()[$name];
+        $type = $argument->getType();
 
-        if (null === $attribute->filter && 'array' === $argument->getType()) {
+        if (null === $attribute->filter && 'array' === $type) {
             if (!$argument->isVariadic()) {
                 return [(array) $value];
             }
@@ -59,23 +63,44 @@ final class QueryParameterValueResolver implements ValueResolverInterface
             'options' => $attribute->options,
         ];
 
-        if ('array' === $argument->getType() || $argument->isVariadic()) {
+        if ('array' === $type || $argument->isVariadic()) {
             $value = (array) $value;
             $options['flags'] |= \FILTER_REQUIRE_ARRAY;
         } else {
             $options['flags'] |= \FILTER_REQUIRE_SCALAR;
         }
 
-        $filter = match ($argument->getType()) {
+        $enumType = null;
+        $filter = match ($type) {
             'array' => \FILTER_DEFAULT,
             'string' => \FILTER_DEFAULT,
             'int' => \FILTER_VALIDATE_INT,
             'float' => \FILTER_VALIDATE_FLOAT,
             'bool' => \FILTER_VALIDATE_BOOL,
-            default => throw new \LogicException(sprintf('#[MapQueryParameter] cannot be used on controller argument "%s$%s" of type "%s"; one of array, string, int, float or bool should be used.', $argument->isVariadic() ? '...' : '', $argument->getName(), $argument->getType() ?? 'mixed'))
+            default => match ($enumType = is_subclass_of($type, \BackedEnum::class) ? (new \ReflectionEnum($type))->getBackingType()->getName() : null) {
+                'int' => \FILTER_VALIDATE_INT,
+                'string' => \FILTER_DEFAULT,
+                default => throw new \LogicException(sprintf('#[MapQueryParameter] cannot be used on controller argument "%s$%s" of type "%s"; one of array, string, int, float, bool or \BackedEnum should be used.', $argument->isVariadic() ? '...' : '', $argument->getName(), $type ?? 'mixed')),
+            }
         };
 
         $value = filter_var($value, $attribute->filter ?? $filter, $options);
+
+        if (null !== $enumType && null !== $value) {
+            $enumFrom = static function ($value) use ($type) {
+                if (!\is_string($value) && !\is_int($value)) {
+                    return null;
+                }
+
+                try {
+                    return $type::from($value);
+                } catch (\ValueError) {
+                    return null;
+                }
+            };
+
+            $value = \is_array($value) ? array_map($enumFrom, $value) : $enumFrom($value);
+        }
 
         if (null === $value && !($attribute->flags & \FILTER_NULL_ON_FAILURE)) {
             throw new NotFoundHttpException(sprintf('Invalid query parameter "%s".', $name));
--- a/src/Symfony/Component/HttpKernel/CHANGELOG.md
+++ b/src/Symfony/Component/HttpKernel/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 6.4
 ---
 
+ * Support backed enums in #[MapQueryParameter]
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by nicolas-grekas — APPROVED** [review]
I pushed a few CS changes. LGTM. Thanks.

**Review by maxbeckers — APPROVED** [review]

File: src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php [hunk 1]
```diff
@@ -59,20 +63,25 @@ public function resolve(Request $request, ArgumentMetadata $argument): array
...
-        $filter = match ($argument->getType()) {
+        $enumType = null;
+        $filter = match ($type) {
             'array' => \FILTER_DEFAULT,
             'string' => \FILTER_DEFAULT,
             'int' => \FILTER_VALIDATE_INT,
             'float' => \FILTER_VALIDATE_FLOAT,
             'bool' => \FILTER_VALIDATE_BOOL,
-            default => throw new \LogicException(sprintf('#[MapQueryParameter] cannot be used on controller argument "%s$%s" of type "%s"; one of array, string, int, float or bool should be used.', $argument->isVariadic() ? '...' : '', $argument->getName(), $argument->getType() ?? 'mixed'))
+            default => match ($enumType = is_subclass_of($type, \BackedEnum::class) ? (new \ReflectionEnum($type))->getBackingType()->getName() : null) {
```
andersmateusz, line 80: @nicolas-grekas `::getName` only allowed since php 8.2.0: https://www.php.net/manual/en/reflectionenum.getbackingtype

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php
```diff
@@ -81,6 +90,22 @@ public function resolve(Request $request, ArgumentMetadata $argument): array
             throw new NotFoundHttpException(sprintf('Invalid query parameter "%s".', $name));
         }
 
+        if (null !== $enumType && null !== $value) {
+            $enumFrom = static function ($value) use ($type, $name) {
+                if (null !== $value && !\is_string($value) && !\is_int($value)) {
```
andersmateusz: @nicolas-grekas null check on line 93 already

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php
```diff
@@ -81,6 +90,22 @@ public function resolve(Request $request, ArgumentMetadata $argument): array
...
+                    throw new NotFoundHttpException(sprintf('Invalid query parameter "%s": expecting an int or string, got "%s".', $name, get_debug_type($value)));
+                }
+
+                try {
+                    return null === $value ? null : $type::from($value);
```
andersmateusz: @nicolas-grekas null check on line 93 already. Is it on purpose?

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php
nicolas-grekas: Not sure what you mean: https://www.php.net/manual/en/reflectionnamedtype.getname.php

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php
nicolas-grekas: It covers null nested in arrays.

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php
```diff
@@ -81,6 +90,22 @@ public function resolve(Request $request, ArgumentMetadata $argument): array
...
+                } catch (\ValueError $e) {
+                    throw new NotFoundHttpException(sprintf('Invalid query parameter "%s": '.$e->getMessage().'.', $name), $e);
```
nicolas-grekas: I realized that we're missing a check like this one (and a corresponding test case): 
```suggestion
                    if ($attribute->flags & \FILTER_NULL_ON_FAILURE) {
                        return null;
                    }
                    throw new NotFoundHttpException(sprintf('Invalid query parameter "%s": '.$e->getMessage().'.', $name), $e);
```

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php
andersmateusz: Now I see that `ReflectionType` is abstract and you get `ReflectionNamedType` always. I was convinced that you get `ReflectionNamedType` since 8.2.0 only, that is why I've used `::_toString`. Just return type has changed since 8.2.0 from abstract to concrete type. :)

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php
nicolas-grekas: Up to finish this aspect @andersmateusz?

andersmateusz: Ok

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php
```diff
@@ -59,24 +63,45 @@ public function resolve(Request $request, ArgumentMetadata $argument): array
...
         $value = filter_var($value, $attribute->filter ?? $filter, $options);
 
+        if (null !== $enumType && null !== $value) {
+            $enumFrom = static function ($value) use ($type) {
+                if (!\is_string($value) && !\is_int($value)) {
+                    return null;
+                }
+
+                try {
+                    return $type::from($value);
```
kells64000, line 96: Why not use [`BackedEnum::tryFrom`](https://www.php.net/manual/en/backedenum.tryfrom.php)?

andersmateusz: I was going to use it but this is why I've not done this:
``` php
#! /usr/bin/env php
<?php
enum StatusName: string {case DONE = 'sth';}
echo var_export(StatusName::tryFrom(1)); // NULL
enum Status: int {case DONE = 0;}
Status::tryFrom('sth'); // Fatal error: Uncaught TypeError: Status::tryFrom(): Argument #1 ($value) must be of type int, string given
```

kells64000: OK, I didn't know, thanks.

nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @andersmateusz.

PR #45721, opened 2022-03-11
[Serializer] Add types to private and internal properties

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 6.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

This PR adds types to private and `@internal` properties of the serializer component. This time, I've also applied CPP and `readonly` flags where possible. If this makes the review too hard or there are other reasons not to do that, please tell me and I'll skip it in future PRs.

PR #51005, opened 2023-07-17
[Routing] adding typehint to some functions and parameters

helyakin:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no

So i was walking around the Routing component for diverse reasons, and I came across some not type hinted:
- Function
- Parameter

I've also taken the freedom to add the property accessibility into the constructor since we are using php81

I have absolutely no idea if this PR is welcome, as this might be something you would not want to add to this framework for now.

Feel free to share your thoughts and opinions about this PR, I'll gladly learn whatever you might say about it !

Thanks again for your time

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Routing/AnnotatedRouteControllerLoader.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Routing/AnnotatedRouteControllerLoader.php
@@ -27,7 +27,7 @@ class AnnotatedRouteControllerLoader extends AnnotationClassLoader
      *
      * @return void
      */
-    protected function configureRoute(Route $route, \ReflectionClass $class, \ReflectionMethod $method, object $annot)
+    protected function configureRoute(Route $route, \ReflectionClass $class, \ReflectionMethod $method, object $annot): void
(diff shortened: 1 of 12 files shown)


wouterj:
Hi!

Adding a return type is a backwards compatibility break for child classes overriding this method. In 6.4, all methods have either a PHP return type or an `@return` PHPdoc (#47551, #50821). This results in a deprecation notices when you override a method without adding the correct PHP return type. In 7.0, we've added all missing return types (#50852, #50846) - except from some that we want to extend for another 2 years to avoid a big burden on the community.

For the same reasons, we've also limited property types to private or internal properties, or properties on final or internal classes (see #41924 for a bit more context).

Finally, I'm not sure what our policy is with CPP. I'm not too fond of it, as it's merely syntax sugar and can cause edge-case BC breaks as well: https://github.com/symfony/symfony/pull/45721#discussion_r1265948565

nicolas-grekas:
Closing as explained by @wouterj, thanks for proposing.

Issue #51010, opened 2023-07-18
[VarExporter] foreach() argument must be of type array|object, string given, ClassNotFoundException


vudaltsov:
### Symfony version(s) affected

6.3.0

### Description

I get these errors when using Symfony VarExporter on PHP 8.2.8.

```
PHP Warning:  foreach() argument must be of type array|object, string given in /vendor/symfony/var-exporter/VarExporter.php on line 100
PHP Stack trace:
PHP   1. {main}() /index.php:0
PHP   2. Symfony\Component\VarExporter\VarExporter::export($value = class A { public string $x = 'a'; public array $y = [0 => class A { ... }] }, $isStaticValue = *uninitialized*, $foundClasses = *uninitialized*) /index.php:25

Warning: foreach() argument must be of type array|object, string given in /vendor/symfony/var-exporter/VarExporter.php on line 100

Call Stack:
    0.0009     354480   1. {main}() /index.php:0
    0.2091     785312   2. Symfony\Component\VarExporter\VarExporter::export($value = class A { public string $x = 'a'; public array $y = [0 => class A { ... }] }, $isStaticValue = ???, $foundClasses = ???) /index.php:25

PHP Warning:  foreach() argument must be of type array|object, string given in /vendor/symfony/var-exporter/VarExporter.php on line 100
PHP Stack trace:
PHP   1. {main}() /index.php:0
PHP   2. Symfony\Component\VarExporter\VarExporter::export($value = class A { public string $x = 'a'; public array $y = [0 => class A { ... }] }, $isStaticValue = *uninitialized*, $foundClasses = *uninitialized*) /index.php:25

Warning: foreach() argument must be of type array|object, string given in /vendor/symfony/var-exporter/VarExporter.php on line 100

Call Stack:
    0.0009     354480   1. {main}() /index.php:0
    0.2091     785312   2. Symfony\Component\VarExporter\VarExporter::export($value = class A { public string $x = 'a'; public array $y = [0 => class A { ... }] }, $isStaticValue = ???, $foundClasses = ???) /index.php:25

PHP Fatal error:  Uncaught Symfony\Component\VarExporter\Exception\ClassNotFoundException: Class "y" not found. in /vendor/symfony/var-exporter/Internal/Hydrator.php:99
Stack trace:
#0 /vendor/symfony/var-exporter/Internal/Hydrator.php(45): Symfony\Component\VarExporter\Internal\Hydrator::getHydrator('y')
[...]

PR #51017, opened 2023-07-18
[VarExporter] Fix exporting classes with __serialize() but not __unserialize()

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | -
| Tickets       | https://github.com/symfony/symfony/issues/51010
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Symfony/Component/VarExporter/Internal/Exporter.php
+++ b/src/Symfony/Component/VarExporter/Internal/Exporter.php
@@ -73,20 +73,29 @@ class Exporter
 
             $class = \get_class($value);
             $reflector = Registry::$reflectors[$class] ?? Registry::getClassReflector($class);
+            $properties = [];
 
             if ($reflector->hasMethod('__serialize')) {
                 if (!$reflector->getMethod('__serialize')->isPublic()) {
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/VarExporter/Internal/Exporter.php
```diff
@@ -83,6 +83,15 @@ public static function prepare($values, $objectsPool, &$refsPool, &$objectsCount
                     throw new \TypeError($class.'::__serialize() must return an array');
                 }
 
+                if (!$reflector->hasMethod('__unserialize')) {
+                    $serializeProperties = $properties;
```
fancyweb: @nicolas-grekas Maybe this could all be simplified to `$properties = [$class => $properties];` for this case? What's the benefit of the `stdClass` stuff strategy btw, is it for perf?

nicolas-grekas: It's for compactness but I guess this is not needed. Let's fix the issue by always using $class indeed.

nicolas-grekas: on closer review, let's keep as is

nicolas-grekas:
Thank you @fancyweb.

PR #51021, opened 2023-07-19
[Uid] Mark Uuid::v5, Uuid::fromString, Uuid::toRfc4122 as pure

### LINKED ISSUE #51001: Uuid::v5, Uuid::fromString, Uuid::toRfc4122  are not marked as pure

6.3.x

### Description

The whole selling point of UUID v5 is their immutability, every time you call it with the same input, you get the same output.

Same applies to `Uuid::fromString` and `Uuid::toRfc4122`.

### How to reproduce

Try this

```php
    /**
     * @psalm-pure
     */
    private static function serialize(string $payload): string
    {
        return Uuid::v5(Uuid::fromString(self::NAMESPACE), $payload)->toRfc4122();
    }
```
    
Psalm produces errors

```
ERROR: ImpureMethodCall - src/Document/Identifier/IdentifierTrait.php:42:22 - Cannot call an impure method from a pure context (see https://psalm.dev/203)
        return Uuid::v5(Uuid::fromString(self::NAMESPACE), $payload)->toRfc4122();


ERROR: ImpureMethodCall - src/Document/Identifier/IdentifierTrait.php:42:31 - Cannot call an impure method from a pure context (see https://psalm.dev/203)
        return Uuid::v5(Uuid::fromString(self::NAMESPACE), $payload)->toRfc4122();


ERROR: ImpureMethodCall - src/Document/Identifier/IdentifierTrait.php:42:71 - Cannot call a non-mutation-free method Symfony\Component\Uid\UuidV5::toRfc4122 from a pure context (see https://psalm.dev/203)
        return Uuid::v5(Uuid::fromString(self::NAMESPACE), $payload)->toRfc4122();

```

### Possible Solution

Mark the method as pure.

### Additional Context

Important for Psalm users.

maxbeckers:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #51001
| License       | MIT
| Doc PR        | n/a

The Problem is described in #51001.

Make it compatible for Psalm users.

### DIFF
--- a/src/Symfony/Component/Uid/Uuid.php
+++ b/src/Symfony/Component/Uid/Uuid.php
@@ -42,6 +42,9 @@ class Uuid extends AbstractUid
         }
     }
 
+    /**
+     * @psalm-pure
+     */
     public static function fromString(string $uuid): static
     {
         if (22 === \strlen($uuid) && 22 === strspn($uuid, BinaryUtil::BASE58[''])) {
@@ -107,6 +110,9 @@ class Uuid extends AbstractUid
         return new UuidV4();
     }
 
+    /**
+     * @psalm-pure
+     */
     final public static function v5(self $namespace, string $name): UuidV5
     {
         // don't use uuid_generate_sha1(), some versions are buggy
@@ -152,6 +158,9 @@ class Uuid extends AbstractUid
         return uuid_parse($this->uid);
     }
 
+    /**
+     * @psalm-pure
+     */
     public function toRfc4122(): string
     {
         return $this->uid;


derrabus:
Looking at the Psalm failures, it seems there are a lot of other methods that we need to flag as pure. I'm not sure if we want to introduce this kind of annotation to this codebase or if that's something that belongs into https://github.com/psalm/psalm-plugin-symfony.

nicolas-grekas:
Definitely for the plugin. Thanks for proposing.

PR #51027, opened 2023-07-19
[Validator] Fix constraints merge for protected and public properties

kontsevoye:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

There was backward compatibility break in https://github.com/symfony/validator/commit/81772c16452f67357c1a7afdbb6cc4f3985ae67b.

When you have a parent class with a public or protected property that has a constraint, and a child class that also has a constraint for the same property, you need to merge them rather than override them.


xabbuh:

kontsevoye:
Yes, you're right. So, I'm going to close this PR

PR #51028, opened 2023-07-19
[Messenger] Fix presenting "[OK] No failed messages were found."

brzuchal:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | yes
| New feature?  | no
| License       | MIT

This improves DX - prevents from showing a message saying `[OK] No failed messages were found.` if they are but `--max=0`. Passing `--max=0` can be used to see the count of failed messages without deserializing and printing them.

no-failed-messages-were-found


fabpot:
It feels like a hack. But anyway, I think what you're looking for is the `messenger:stats` command which does exactly this, right?

brzuchal:
@fabpot I was not aware of the command, thanks it feels like a better way to check the counters. Regardless of that, I think the message improvement is still valid.

fabpot:
I think we should make 0 for max invalid and throw an exception.

PR #51033, opened 2023-07-19
Add "SKIP LOCKED" to the query that retrieves messages

hgraca:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | [#47828](https://github.com/symfony/symfony/issues/47828)
| License       | MIT
| Doc PR        | nothing to document, its purely internal
<!--

The current `SELECT ... FOR UPDATE` retrieves rows, using the index to find and lock the retrieved rows.
This means that when we have more than one consumer, while one consumer is locking those rows, the whole index is locked thus other queries (consumers) will be put on hold.
While with a small table this might not be noticeable, as the table grows the meddling with the index becomes slower and the other consumers have to wait more time, eventually making the MQ inoperable.

The `SKIP LOCKED` addition will allow other consumers to query the table and get messages immediately, ignoring the rows that other consumers are locking.


hgraca:
Gonna close this as we actually decided to open an MR in DBAL first

PR #51034, opened 2023-07-19
[Messenger] Batch delivery

### LINKED ISSUE #47828: [Messenger] Bad performances with the Doctrine/MySQL transport

hgraca:
| Q             | A
| ------------- | ---
| Branch?       | 7.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix [#47828](https://github.com/symfony/symfony/issues/47828)
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/18566

Currently, the doctrine transport only delivers one message at a time,
resulting in low performance and a very high amount of hits to the DB.
(one hit per message)

With batch delivery, the transport will retrieve several messages in a single query.

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
+++ b/src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Component\Messenger\Bridge\Doctrine\Transport;
 
+use Doctrine\DBAL\ArrayParameterType;
 use Doctrine\DBAL\Connection as DBALConnection;
 use Doctrine\DBAL\Driver\Exception as DriverException;
 use Doctrine\DBAL\Exception as DBALException;
@@ -43,6 +44,7 @@ class Connection implements ResetInterface
     protected const DEFAULT_OPTIONS = [
         'table_name' => 'messenger_messages',
         'queue_name' => 'default',
+        'batch_size' => 1,
         'redeliver_timeout' => 3600,
         'auto_setup' => true,
     ];
@@ -63,6 +65,8 @@ class Connection implements ResetInterface
     protected ?float $queueEmptiedAt = null;
 
     private bool $autoSetup;
+    /** @var int[] */
+    private array $lastDeliveredButNotYetHandledEnvelopesIds = [];
 
     public function __construct(array $configuration, DBALConnection $driverConnection)
     {
@@ -151,7 +155,7 @@ class Connection implements ResetInterface
         return $this->driverConnection->lastInsertId();
     }
 
-    public function get(): ?array
+    public function get(): array
     {
         if ($this->driverConnection->getDatabasePlatform() instanceof AbstractMySQLPlatform) {
             try {
@@ -166,7 +170,7 @@ class Connection implements ResetInterface
         try {
             $query = $this->createAvailableMessagesQueryBuilder()
                 ->orderBy('available_at', 'ASC')
-                ->setMaxResults(1);
+                ->setMaxResults($this->configuration['batch_size']);
 
             if ($this->driverConnection->getDatabasePlatform() instanceof OraclePlatform) {
                 $query->select('m.id');
@@ -200,40 +204,45 @@ class Connection implements ResetInterface
                 $sql .= ' '.$this->driverConnection->getDatabasePlatform()->getWriteLockSQL();
             }
 
-            $doctrineEnvelope = $this->executeQuery(
+            $doctrineEnvelopeList = $this->executeQuery(
                 $sql,
                 $query->getParameters(),
                 $query->getParameterTypes()
-            )->fetchAssociative();
+            )->fetchAllAssociative();
 
-            if (false === $doctrineEnvelope) {
+            if ([] === $doctrineEnvelopeList) {
                 $this->driverConnection->commit();
                 $this->queueEmptiedAt = microtime(true) * 1000;
 
-                return null;
+                return [];
             }
             // Postgres can "group" notifications having the same channel and payload
             // We need to be sure to empty the queue before blocking again
             $this->queueEmptiedAt = null;
 
-            $doctrineEnvelope = $this->decodeEnvelopeHeaders($doctrineEnvelope);
+            $this->lastDeliveredButNotYetHandledEnvelopesIds = $this->extractDoctrineEnvelopeListIds($doctrineEnvelopeList);
 
             $queryBuilder = $this->driverConnection->createQueryBuilder()
                 ->update($this->configuration['table_name'])
                 ->set('delivered_at', '?')
-                ->where('id = ?');
-            $now = new \DateTimeImmutable('UTC');
-            $this->executeStatement($queryBuilder->getSQL(), [
-                $now,
-                $doctrineEnvelope['id'],
-            ], [
-                Types::DATETIME_IMMUTABLE,
-            ]);
+                ->where($this->driverConnection->createQueryBuilder()->expr()->in('id', '?'));
+            $this->executeStatement(
+                $queryBuilder->getSQL(),
+                [
+                    new \DateTimeImmutable('UTC'),
+                    $this->lastDeliveredButNotYetHandledEnvelopesIds,
+                ],
+                [
+                    Types::DATETIME_IMMUTABLE,
+                    class_exists(ArrayParameterType::class) ? ArrayParameterType::INTEGER : Types::SIMPLE_ARRAY,
+                ]
+            );
 
             $this->driverConnection->commit();
 
-            return $doctrineEnvelope;
+            return $this->decodeDoctrineEnvelopeListHeaders($doctrineEnvelopeList);
         } catch (\Throwable $e) {
+            $this->lastDeliveredButNotYetHandledEnvelopesIds = [];
             $this->driverConnection->rollBack();
 
             if ($this->autoSetup && $e instanceof TableNotFoundException) {
@@ -247,6 +256,7 @@ class Connection implements ResetInterface
 
     public function ack(string $id): bool
     {
+        $this->removeHandledFromDeliveredList($id);
         try {
             if ($this->driverConnection->getDatabasePlatform() instanceof AbstractMySQLPlatform) {
                 return $this->driverConnection->update($this->configuration['table_name'], ['delivered_at' => '9999-12-31 23:59:59'], ['id' => $id]) > 0;
@@ -336,6 +346,27 @@ class Connection implements ResetInterface
     public function getExtraSetupSqlForTable(Table $createdTable): array
     {
         return [];
+    }
+
+    public function undeliverNotHandled(): void
+    {
+        if (0 === \count($this->lastDeliveredButNotYetHandledEnvelopesIds)) {
+            return;
+        }
+
+        $queryBuilder = $this->driverConnection->createQueryBuilder()
+            ->update($this->configuration['table_name'])
+            ->set('delivered_at', 'NULL')
+            ->where($this->driverConnection->createQueryBuilder()->expr()->in('id', '?'));
+        $this->executeStatement(
+            $queryBuilder->getSQL(),
+            [
+                $this->lastDeliveredButNotYetHandledEnvelopesIds,
+            ],
+            [
+                ArrayParameterType::INTEGER,
+            ]
+        );
     }
 
     private function createAvailableMessagesQueryBuilder(): QueryBuilder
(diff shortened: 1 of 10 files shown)


src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
```diff
@@ -46,10 +48,13 @@ class Connection implements ResetInterface
     protected const DEFAULT_OPTIONS = [
         'table_name' => 'messenger_messages',
         'queue_name' => 'default',
+        'batch_size' => 1,
         'redeliver_timeout' => 3600,
         'auto_setup' => true,
     ];
 
+    public const FLAGGED_FOR_DELETION = '9999-12-31 23:59:59';
```
fabpot: To be reverted as not related to this PR.

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
```diff
@@ -66,6 +71,8 @@ class Connection implements ResetInterface
     protected $queueEmptiedAt;
     private ?SchemaSynchronizer $schemaSynchronizer;
     private bool $autoSetup;
+    /** @var array<int> */
```
fabpot:
```suggestion
    /** @var int[] */
```

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
fabpot: Or at least, make it private.

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/DoctrineTransportFactory.php
```diff
@@ -20,9 +21,12 @@
 
 /**
  * @author Vincent Touzet <vincent.touzet@gmail.com>
+ * @author Herberto Graca <herberto.graca@gmail.com>
  */
 class DoctrineTransportFactory implements TransportFactoryInterface
 {
+    public const DSN_PREFIX = 'doctrine://';
```
hgraca: After reading [this comment](https://github.com/symfony/symfony/pull/51034#pullrequestreview-1554567967), I'm doubting if this public constant ok, or if I should drop this commit? (only has this change)

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
hgraca: I made it private. 
What about [this one](https://github.com/symfony/symfony/pull/51034/files#r1281056793)?
Is it ok to have that one public, or should I drop that commit? (only ahs that change)

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
hgraca: changed

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/DoctrineTransportFactory.php
fabpot: I would drop the constant altogether.

hgraca: Done, and tkx for the CR btw :)

hgraca:
I've rebased this PR on branch 6.4, and removed all the commits that were merely small improvements (I will open PRs for them later).

I see that the pipeline is failing but it seems to be unrelated to this PR.

I am, however, afraid that this PR might be getting stale since it is open for a few weeks now.

Is this normal or is there something else that I should do? (I'm new to contributing to Symfony so I really don't know)

btw, there are checks failing but I believe they are unrelated to this PR.

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
hgraca: I removed this and other small fixes from the PR as to remove noise in the PR

hgraca:
Rebased again and put the changelog in the correct file

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/DoctrineTransport.php
```diff
@@ -37,6 +39,18 @@ public function __construct(Connection $connection, SerializerInterface $seriali
         $this->serializer = $serializer;
     }
 
+    public function __destruct()
+    {
+        // The worker using this transport might have pulled 50 msgs out of the mq, marking them as "in process",
+        // but because of its options (ie --limit, --failure-limit, --memory-limit, --time-limit) it might terminate
```
ro0NL, line 45: shouldnt you read the options instead :thinking:

hgraca, line 45: Then I think we would increase complexity and it wouldn't scale.

For `--limit` we could do some calculations... would be doable despite the increase in complexity..
But for `-failure-limit` we would need to somehow keep track of how many failures happened, probably replicating logic that is already built somewhere. For `--memory-limit` and `--time-limit` the same.

Then, if/when in the future we would add more options, someone would need to also adjust this code, to be able to handle the new option.

With the implementation in this PR however, we don't need to do anything special to adjust to a particular option, neither now nor in the future if/when we add new options.

ro0NL: got ya, i was thinking we might correlate limit and batch_size somehow, but i agree it increases complexity for little gain, and the conceptual issue remains either way

ro0NL: im not sure if `__destruct` is the best extension point compared to eg. events, but if it works .. :shrug:

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
```diff
@@ -510,4 +541,53 @@ private function fallBackToForUpdate(QueryBuilder $query, string $sql): string
             return $sql;
         }
     }
+
+    /**
+     * @return int[]
+     */
+    public function extractDoctrineEnvelopeListIds(array $doctrineEnvelopeList): array
+    {
+        return array_map(
```
stof, line 550: `return array_column($doctrineEnvelopeList, 'id')` will do the same and might be a bit faster by keeping all the logic in the C layer.

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
```diff
@@ -151,7 +155,7 @@ public function send(string $body, array $headers, int $delay = 0): string
         return $this->driverConnection->lastInsertId();
     }
 
-    public function get(): ?array
+    public function get(): array
```
stof, line 158: I suggest document the type of the returned array.

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
```diff
@@ -200,40 +204,45 @@ public function get(): ?array
...
+                ->where($this->driverConnection->createQueryBuilder()->expr()->in('id', '?'));
+            $this->executeStatement(
+                $queryBuilder->getSQL(),
+                [
+                    new \DateTimeImmutable('UTC'),
+                    $this->lastDeliveredButNotYetHandledEnvelopesIds,
+                ],
+                [
+                    Types::DATETIME_IMMUTABLE,
+                    class_exists(ArrayParameterType::class) ? ArrayParameterType::INTEGER : Types::SIMPLE_ARRAY,
```
stof, line 237: This is wrong. `Types::SIMPLE_ARRAY` is something totally different. The fallback for older DBAL versions is `Connection::PARAM_INT_ARRAY`.

And the code below does not check for older versions, so it is inconsistent.

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/DoctrineReceiver.php
```diff
@@ -58,11 +60,16 @@ public function get(): iterable
             throw new TransportException($exception->getMessage(), 0, $exception);
         }
 
-        if (null === $doctrineEnvelope) {
+        if (null === $doctrineEnvelopeList) {
```
stof, line 63: This won't ever be true. `Connection::get` cannot return `null` anymore.

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/DoctrineReceiver.php
```diff
@@ -83,6 +90,11 @@ public function reject(Envelope $envelope): void
         }
     }
 
+    public function undeliverNotHandled(): void
```
stof, line 93: should it be marked as `@internal` ?

OskarStark:
Closing due to lack of feedback. Feel free to reopen. Thanks for proposing.

Issue #50690, opened 2023-06-16
[HttpKernel] #[MapQueryString] with an empty query string always returns null


DrLuke:
### Symfony version(s) affected

6.3.0

### Description

`RequestPayloadValueResolver` always [returns null](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php#L152-L154) when mapping query strings and query string is empty.

This is problematic when using a DTO where all arguments have default values, e.g.

```php
class MyDto {
    public function __construct(public string $foo = "bar")
    {}
}
```

### How to reproduce

Controller:

```php
class MyDto {
    public function __construct(public string $foo = "bar")
    {}
}

class FooController extends AbstractController
{
    #[Route("/foo", name: "foo", methods: ["GET"])]
    public function getPaginated(
        #[MapQueryString] MyDto $myDto,
    ): JsonResponse
    {
        return $this->json($myDto);
    }
}
```

`GET /foo` -> `404 HttpException`
`GET /foo?foo=baz` -> `200 {"foo": "baz"}`

### Possible Solution

Remove [this](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php#L152-L154) early return and let the serializer decide what to do with an empty query string.

DrLuke:
Workaround for the time being:

```php
class MyDto {
    public function __construct(public string $foo = "bar")
    {}
}

class FooController extends AbstractController
{
    #[Route("/foo", name: "foo", methods: ["GET"])]
    public function getPaginated(
        #[MapQueryString] ?MyDto $myDto,
    ): JsonResponse
    {
        return $this->json($myDto ?? new MyDto());
    }
}
```

mdeboer:
I don't think removing the early return is the right way to do it imho.

What if the argument is nullable and you want it to be null when no query parameters are present?

So instead of removing it, it should add a condition that checks if the parameter is nullable. If it is, it should return null. If it isn't it should just let the serialiser handle it.

DrLuke:
@mdeboer I would let the serializer handle it first. If it can not deserialize, return `null` if nullable.

mdeboer:
Oh no. You would never be able to handle other errors like you can now.

I don't see what good that would do. Can you elaborate?

DrLuke:
[...]

PR #51036, opened 2023-07-20
[HttpKernel] Fix issue when no query parameter with mapQueryString

AurelienPillevesse:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

```php
#[Route('/hello', methods: ['GET'])]
public function getHello(
    #[MapQueryString] HelloGetRequest $helloGetRequest
): Response {
    var_dump($helloGetRequest);
    die;
}

final class HelloGetRequest
{
    #[Assert\NotBlank(allowNull: true)]
    #[Assert\Date]
    public ?string $dateDebut = null;

    #[Assert\NotBlank(allowNull: true)]
    #[Assert\Date]
    public ?string $dateFin = null;
}
```

If I call the URL ```http://localhost:8080/hello```, I will receive a 404 Not Found error because there is nothing in the query parameters.
The idea is to have the object ```HelloGetRequest``` initialized to access it and if I try to get ```$helloGetRequest->dateDebut```, ```null``` will be displayed.

### DIFF
--- a/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php
+++ b/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php
@@ -157,11 +157,13 @@ class RequestPayloadValueResolver implements ValueResolverInterface, EventSubscr
 
     private function mapQueryString(Request $request, string $type, MapQueryString $attribute): ?object
     {
-        if (!$data = $request->query->all()) {
-            return null;
+        $data = $request->query->all();
+
+        if (class_exists($type) || $data) {
+            return $this->serializer->denormalize($data, $type, null, self::CONTEXT_DENORMALIZE + $attribute->serializationContext);
         }
 
-        return $this->serializer->denormalize($data, $type, null, self::CONTEXT_DENORMALIZE + $attribute->serializationContext);
+        return null;
     }
 
     private function mapRequestPayload(Request $request, string $type, MapRequestPayload $attribute): ?object


nicolas-grekas:
You should write the argument this way instead:
```php
 #[MapQueryString] HelloGetRequest $helloGetRequest = new HelloGetRequest()
```

AurelienPillevesse:
Thanks for the reply @nicolas-grekas, it might be interesting to add this information somewhere in the documentation to share the knowledge.

nicolas-grekas:
You're right. Can you please consider a PR to the doc?

kubk:
> You should write the argument this way instead:
> 
> ```
>  #[MapQueryString] HelloGetRequest $helloGetRequest = new HelloGetRequest()
> ```

@nicolas-grekas While I understand that it can be solved on the user's end, I can't imagine why anyone would want the route with query params to give a 404 error when there are no query parameters. Query parameters are always optional. If the issue isn't going to be fixed, then developers should provide a default value for all the routes using #[MapQueryString]. Which is easy to forget (but can be solved by new static analysis rules)

nicolas-grekas:
See also https://github.com/symfony/symfony/issues/50690#issuecomment-1659890007

But there are too many comments on this topic scattered on closed issues.
Please open a dedicated one if you want any follow up on the topic.

On my side, I see no issues with `#[MapQueryString] HelloGetRequest $helloGetRequest = null`, and I'm sure there are URLs that mandate a query string so the use case should be covered.

PR #51038, opened 2023-07-20
Fix for: Regression in validator #50780

### LINKED ISSUE #50780: Regression in validator

michaelehrlein:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50780
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Validator/Mapping/ClassMetadata.php
+++ b/src/Symfony/Component/Validator/Mapping/ClassMetadata.php
@@ -458,8 +458,7 @@ class ClassMetadata extends GenericMetadata implements ClassMetadataInterface
     private function addPropertyMetadata(PropertyMetadataInterface $metadata): void
     {
         $property = $metadata->getPropertyName();
-
-        $this->members[$property][] = $metadata;
+        $this->members[$property][spl_object_id($metadata)] = $metadata;
     }
 
     private function checkConstraint(Constraint $constraint): void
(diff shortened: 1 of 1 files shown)


xabbuh:

michaelehrlein:
> This looks like a duplicate of #50788 to me.

You're right, thanks, hadn't seen it 🙈

PR #51041, opened 2023-07-20
[Form] Use `form.post_set_data` in `ResizeFormListener`

HeahDude:
| Q             | A
| ------------- | ---
| Branch?       | 7.2
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | ~
| License       | MIT
| Doc PR        | TODO

The `form.pre_set_data` event allows to change the initial data, modifying the form based on it should be done early during `form.post_set_data` instead.

Ref https://github.com/symfony/symfony-docs/pull/18587.

Also, the collect of form config should be done as late as possible to account for modification from other listeners.

### DIFF
--- a/src/Symfony/Component/Form/CHANGELOG.md
+++ b/src/Symfony/Component/Form/CHANGELOG.md
@@ -7,6 +7,8 @@ CHANGELOG
  * Deprecate the `VersionAwareTest` trait, use feature detection instead
  * Add support for the `calendar` option in `DateType`
  * Add `LazyChoiceLoader` and `choice_lazy` option in `ChoiceType` for loading and rendering choices on demand
+ * Use `form.post_set_data` instead of `form.pre_set_data` in `ResizeFormListener`
+ * Change the priority of `DataCollectorListener` from 255 to -255
 
 7.1
 ---
--- a/src/Symfony/Component/Form/Extension/Core/EventListener/ResizeFormListener.php
+++ b/src/Symfony/Component/Form/Extension/Core/EventListener/ResizeFormListener.php
@@ -12,6 +12,7 @@
 namespace Symfony\Component\Form\Extension\Core\EventListener;
 
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Form\Event\PostSetDataEvent;
 use Symfony\Component\Form\Exception\UnexpectedTypeException;
 use Symfony\Component\Form\FormEvent;
 use Symfony\Component\Form\FormEvents;
@@ -27,6 +28,9 @@ class ResizeFormListener implements EventSubscriberInterface
     protected array $prototypeOptions;
 
     private \Closure|bool $deleteEmpty;
+    // BC, to be removed in 8.0
+    private bool $overridden = true;
+    private bool $usePreSetData = false;
 
     public function __construct(
         private string $type,
@@ -44,15 +48,57 @@ class ResizeFormListener implements EventSubscriberInterface
     public static function getSubscribedEvents(): array
     {
         return [
-            FormEvents::PRE_SET_DATA => 'preSetData',
+            FormEvents::PRE_SET_DATA => 'preSetData', // deprecated
+            FormEvents::POST_SET_DATA => ['postSetData', 255], // as early as possible
             FormEvents::PRE_SUBMIT => 'preSubmit',
             // (MergeCollectionListener, MergeDoctrineCollectionListener)
             FormEvents::SUBMIT => ['onSubmit', 50],
         ];
     }
 
+    /**
+     * @deprecated Since Symfony 7.2, use {@see postSetData()} instead.
+     */
     public function preSetData(FormEvent $event): void
     {
+        if (__CLASS__ === static::class
+            || __CLASS__ === (new \ReflectionClass($this))->getMethod('preSetData')->getDeclaringClass()->name
+        ) {
+            // not a child class, or child class does not overload PRE_SET_DATA
+            return;
+        }
+
+        trigger_deprecation('symfony/form', '7.2', 'Calling "%s()" is deprecated, use "%s::postSetData()" instead.', __METHOD__, __CLASS__);
+        // parent::preSetData() has been called
+        $this->overridden = false;
+        try {
+            $this->postSetData($event);
+        } finally {
+            $this->usePreSetData = true;
+        }
+    }
+
+    /**
+     * Remove FormEvent type hint in 8.0.
+     *
+     * @final since Symfony 7.2
+     */
+    public function postSetData(FormEvent|PostSetDataEvent $event): void
+    {
+        if (__CLASS__ !== static::class) {
+            if ($this->overridden) {
+                trigger_deprecation('symfony/form', '7.2', 'Calling "%s::preSetData()" is deprecated, use "%s::postSetData()" instead.', static::class, __CLASS__);
+                // parent::preSetData() has not been called, noop
+
+                return;
+            }
+
+            if ($this->usePreSetData) {
+                // nothing else to do
+                return;
+            }
+        }
+
         $form = $event->getForm();
         $data = $event->getData() ?? [];
 
--- a/src/Symfony/Component/Form/Extension/DataCollector/EventListener/DataCollectorListener.php
+++ b/src/Symfony/Component/Form/Extension/DataCollector/EventListener/DataCollectorListener.php
@@ -32,8 +32,8 @@ class DataCollectorListener implements EventSubscriberInterface
     public static function getSubscribedEvents(): array
     {
         return [
-            // High priority in order to be called as soon as possible
-            FormEvents::POST_SET_DATA => ['postSetData', 255],
+            // Low priority in order to be called as late as possible
+            FormEvents::POST_SET_DATA => ['postSetData', -255],
             // Low priority in order to be called as late as possible
             FormEvents::POST_SUBMIT => ['postSubmit', -255],
         ];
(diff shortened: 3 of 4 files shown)


src/Symfony/Component/Form/Extension/DataCollector/EventListener/DataCollectorListener.php
```diff
@@ -34,8 +34,8 @@ public function __construct(FormDataCollectorInterface $dataCollector)
     public static function getSubscribedEvents(): array
     {
         return [
-            // High priority in order to be called as soon as possible
-            FormEvents::POST_SET_DATA => ['postSetData', 255],
+            // Low priority in order to be called as late as possible
```
xabbuh, line 35: Do you know why the priority was really high before? Could it be that we will forbid some use case with these changes?

HeahDude: Not really, this is there since the origin (#9082).
There are no core usages of post set data event (except for a recent deprecation concerning timezone :)).
But the changes in this PR tells me that if any listener out there modifies the form structure, we would collect a tree out of sync and may display irrelevant information in the profiler.

xabbuh (approved):

yceruto (approved):

src/Symfony/Component/Form/Extension/Core/EventListener/ResizeFormListener.php
```diff
@@ -77,6 +80,11 @@ public function preSetData(FormEvent $event)
         }
     }
 
+    public function postSetData(PostSetDataEvent $event): void
+    {
+        $this->preSetData($event);
```
nicolas-grekas: this should be done the other way around: the deprecated method calls the non-deprecated one

src/Symfony/Component/Form/Extension/Core/EventListener/ResizeFormListener.php
```diff
@@ -45,14 +46,16 @@ public function __construct(string $type, array $options = [], bool $allowAdd =
...
             FormEvents::SUBMIT => ['onSubmit', 50],
         ];
     }
 
     /**
+     * @deprecated Since 6.4, use {@see postSetData()} instead.
+     *
      * @return void
      */
     public function preSetData(FormEvent $event)
```
nicolas-grekas: missing runtime deprecation in the body of the method

src/Symfony/Component/Form/Extension/Core/EventListener/ResizeFormListener.php
HeahDude: By doing the other way around we might break a child class by not calling `preSetData` anymore.
The last time I did such deprecation we handle it that way https://github.com/symfony/symfony/commit/ad08b951d798a5419f57bb092d636ff2c86dfece, https://github.com/symfony/symfony/commit/8a6cf9d93eda5d4c7685fd794a3733f787368055, should we do something similar instead?

nicolas-grekas: I guess so yes. The runtime notice is a must.

HeahDude: See https://github.com/symfony/symfony/pull/51041/commits/ba36dd37091dd6df8b7953185c4bfc0f1f1bf920.

src/Symfony/Component/Form/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ CHANGELOG
    `model_timezone` option in `DateType`, `DateTimeType`, and `TimeType`
  * Deprecate `PostSetDataEvent::setData()`, use `PreSetDataEvent::setData()` instead
  * Deprecate `PostSubmitEvent::setData()`, use `PreSubmitDataEvent::setData()` or `SubmitDataEvent::setData()` instead
+ * Use `form.post_set_data` instead of `form.pre_set_data` in `ResizeFormListener`
+ * Change the priority of `DataCollectorListener` from 255 to -255
```
nicolas-grekas: Are we sure about this? From experience this can lead to unwanted behaviors, eg what if a listener stop propagation? (is that possible with Doctrine?)
What's the motivation for this change of priority? Is that just modern-us thinking ancient-us made a mistake? Given the comment, this looks on purpose to me.

src/Symfony/Component/Form/Extension/Core/EventListener/ResizeFormListener.php
```diff
@@ -45,18 +50,51 @@ public function __construct(string $type, array $options = [], bool $allowAdd =
...
+            || __CLASS__ === (new \ReflectionClass($this))->getMethod('preSetData')->getDeclaringClass()->name
+        ) {
+            // not a child class, or child class does not overload PRE_SET_DATA
+        } else {
+            trigger_deprecation('symfony/form', '6.4', 'Calling "%s()" is deprecated, use "%s::postSetData()" instead.', __METHOD__, __CLASS__);
+            // parent::preSetData() has been called
+            $this->overridden = false;
+            $this->postSetData($event);
+            $this->usePreSetData = true;
+        }
```
nicolas-grekas:
```suggestion
            // not a child class, or child class does not overload PRE_SET_DATA
            return;
        }
        trigger_deprecation('symfony/form', '6.4', 'Calling "%s()" is deprecated, use "%s::postSetData()" instead.', __METHOD__, __CLASS__);
        // parent::preSetData() has been called
        $this->overridden = false;
        $this->postSetData($event);
        $this->usePreSetData = true;
```

src/Symfony/Component/Form/CHANGELOG.md
HeahDude: It raises the same concern as @xabbuh in https://github.com/symfony/symfony/pull/51041#discussion_r1270524066.
Yes, I do think that was a mistake.
Given than we don't currently use post set data event in core, I cannot think of a legit use case where propagation is stopped.
To ensure that this listener is called, we could use a flag and `__destruct()` to warn or deprecate such behavior.

nicolas-grekas: OK (no need for the destruct check)

src/Symfony/Component/Form/Extension/Core/EventListener/ResizeFormListener.php
HeahDude: Done.

src/Symfony/Component/Form/Extension/Core/EventListener/ResizeFormListener.php
```diff
@@ -45,18 +50,52 @@ public function __construct(string $type, array $options = [], bool $allowAdd =
     public static function getSubscribedEvents(): array
     {
         return [
-            FormEvents::PRE_SET_DATA => 'preSetData',
+            FormEvents::PRE_SET_DATA => 'preSetData', // deprecated
+            FormEvents::POST_SET_DATA => ['postSetData', 255], // as early as possible
```
stof, line 52: is it actually allowed to modify the structure of the form on postSetData ? To me, this is too late as those child fields won't be populated.

HeahDude: It is, the form is not cloned and can be modified by reference https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Form/Form.php#L327.
The only event that forbids it is `POST_SUBMIT` thanks to this check https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Form/Form.php#L260, in such case `$this->defaultDataSet` is true, but not enough, only data cannot further be changed.

HeahDude: >To me, this is too late as those child fields won't be populated.

They will be: https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Form/Form.php#L776.

HeahDude: I've refactored the tests to ensure data is mapped consistently: https://github.com/symfony/symfony/pull/51041/commits/6588362b4f6a72b0b956b18aa46a8e52a39a5dce.

OskarStark:
Needs a rebase, thanks

nicolas-grekas:
Friendly up @HeahDude

xabbuh:
I have rebased the PR

src/Symfony/Component/Form/Extension/Core/EventListener/ResizeFormListener.php
```diff
@@ -44,15 +48,50 @@ public function __construct(
...
+        ) {
+            // not a child class, or child class does not overload PRE_SET_DATA
+            return;
+        }
+
+        trigger_deprecation('symfony/form', '7.2', 'Calling "%s()" is deprecated, use "%s::postSetData()" instead.', __METHOD__, __CLASS__);
+        // parent::preSetData() has been called
+        $this->overridden = false;
+        $this->postSetData($event);
+        $this->usePreSetData = true;
```
nicolas-grekas:
```suggestion
        try {
            $this->postSetData($event);
        } finally {
            $this->usePreSetData = true;
        }
```

src/Symfony/Component/Form/Extension/Core/EventListener/ResizeFormListener.php
```diff
@@ -44,15 +48,50 @@ public function __construct(
...
+    }
+
+    // Remove FormEvent type hint in 8.0
+    final public function postSetData(FormEvent|PostSetDataEvent $event): void
```
nicolas-grekas: `@final since Symfony 7.3` insead
```suggestion
    /**
     * @param PostSetDataEvent $event
     */
    public function postSetData(FormEvent|PostSetDataEvent $event): void
```

nicolas-grekas (approved):
LGTM after minor comments.

xabbuh:
Thank you Jules.

HeahDude:
Thank you Christian :).

Issue #34846, opened 2019-12-06
Mailer Transports to SMTP Cannot Establish Secure Connections


brandonsavage:
**Symfony version(s) affected**: 5.0.1

**Description**  
When using the Transport out of the box, the Transport is unable to establish a secure connection. This issue looks similar to https://github.com/symfony/symfony/issues/34064

I am using OpenSSL 1.1.1d and PHP 7.4. I've reproduced the issue on PHP 7.3 as well.

The exact error message I receive is:

```Connection could not be established with host "ssl://smtp.postmarkapp.com:587": stream_socket_client(): SSL operation failed with code 1. OpenSSL Error messages: error:1408F10B:SSL routines:ssl3_get_record:wrong version number```

**How to reproduce**  
```php
$transport = new PostmarkSmtpTransport($key);
$mailer = new Mailer($transport);
$mail = (new Email)->to('test@example.com')
->from('test@example.com')
->subject('Test Email')
->html('Test Email');
$mailer->send($mail);
```

fabpot:
maybe fixed by #34760?

brandonsavage:
I applied the patch in #34760 without success in resolving this bug.

kbsanders:
I am experiencing this exact same issue using the following SMTP transport:

```php
$transport = new EsmtpTransport('smtp.office365.com',587,true);
```

The error received is:

`
Connection could not be established with host "ssl://smtp.office365.com:587": stream_socket_client(): SSL operation failed with code 1. OpenSSL Error messages:
error:1408F10B:SSL routines:ssl3_get_record:wrong version number.
`

I am using Symfony version 5.0.1 and PHP version 7.3.11.
I am also using the localhost Symfony web server for development.

The patch mentioned above did not help to correct the issue.

kbsanders:
**Update:**

I ran my code again, but this time with the TLS option set to `false` and the patch above applied and the transport actually succeeded.

```php
$transport = new EsmtpTransport('smtp.office365.com',587,false);
```

However, when the patch is not applied, and TLS = `false`, I receive the following error:

`Warning: fgets(): SSL: An existing connection was forcibly closed by the remote host. in C:\arwa\site-alruralwater\vendor\symfony\mailer\Transport\Smtp\Stream\AbstractStream.php (line 77) `

Does this working with TLS = `false` perhaps have something to do with TLS vs STARTTLS?

toooni:
[...]

PR #51042, opened 2023-07-20
[Mailer] [Mailgun] Disable tls for mailgun as it should use STARTTLS

joelwurtz:
| Q             | A
| ------------- | ---
| Branch?       | 6.3 
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no 
| Tickets       | 
| License       | MIT
| Doc PR        | 

Since symfony 6.3 smtp mailgun use 587 as default port with tls true, but it does not work : 

image

Setting tls to false, force the smtp transport to try to use STARTTLS if available like others smtp providers (mailchimp in example)


joelwurtz:
Ref https://github.com/symfony/symfony/issues/34846#issuecomment-580190302 for more explanation

xabbuh (approved):

nicolas-grekas:
Thank you @joelwurtz

PR #51043, opened 2023-07-20
[Form] Deprecate `FormEvent::setData()` for events that do not allow it

HeahDude:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Tickets       | ~
| License       | MIT
| Doc PR        | TODO

Now that specific classes exist for each event, we can improve DX by forbidding to call `FormEvent::setData()` where it does not make sense.

### DIFF
--- a/UPGRADE-6.4.md
+++ b/UPGRADE-6.4.md
@@ -29,6 +29,8 @@ Form
 
  * Deprecate using `DateTime` or `DateTimeImmutable` model data with a different timezone than configured with the
    `model_timezone` option in `DateType`, `DateTimeType`, and `TimeType`
+ * Deprecate `PostSetDataEvent::setData()`, use `PreSetDataEvent::setData()` instead
+ * Deprecate `PostSubmitEvent::setData()`, use `PreSubmitDataEvent::setData()` or `SubmitDataEvent::setData()` instead
 
 FrameworkBundle
 ---------------
--- a/src/Symfony/Component/Form/CHANGELOG.md
+++ b/src/Symfony/Component/Form/CHANGELOG.md
@@ -6,6 +6,8 @@ CHANGELOG
 
  * Deprecate using `DateTime` or `DateTimeImmutable` model data with a different timezone than configured with the
    `model_timezone` option in `DateType`, `DateTimeType`, and `TimeType`
+ * Deprecate `PostSetDataEvent::setData()`, use `PreSetDataEvent::setData()` instead
+ * Deprecate `PostSubmitEvent::setData()`, use `PreSubmitDataEvent::setData()` or `SubmitDataEvent::setData()` instead
 
 6.3
 ---
--- a/src/Symfony/Component/Form/Event/PostSetDataEvent.php
+++ b/src/Symfony/Component/Form/Event/PostSetDataEvent.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Component\Form\Event;
 
+use Symfony\Component\Form\Exception\BadMethodCallException;
 use Symfony\Component\Form\FormEvent;
 
 /**
@@ -20,4 +21,12 @@ use Symfony\Component\Form\FormEvent;
  */
 final class PostSetDataEvent extends FormEvent
 {
+    /**
+     * @deprecated since Symfony 6.4, it will throw an exception in 7.0.
+     */
+    public function setData(mixed $data): void
+    {
+        trigger_deprecation('symfony/form', '6.4', 'Calling "%s()" will throw an exception as of 7.0, listen to "form.pre_set_data" instead.', __METHOD__);
+        // throw new BadMethodCallException('Form data cannot be changed during "form.post_set_data", you should use "form.pre_set_data" instead.');
+    }
 }
--- a/src/Symfony/Component/Form/Event/PostSubmitEvent.php
+++ b/src/Symfony/Component/Form/Event/PostSubmitEvent.php
@@ -21,4 +21,12 @@ use Symfony\Component\Form\FormEvent;
  */
 final class PostSubmitEvent extends FormEvent
 {
+    /**
+     * @deprecated since Symfony 6.4, it will throw an exception in 7.0.
+     */
+    public function setData(mixed $data): void
+    {
+        trigger_deprecation('symfony/form', '6.4', 'Calling "%s()" will throw an exception as of 7.0, listen to "form.pre_submit" or "form.submit" instead.', __METHOD__);
+        // throw new BadMethodCallException('Form data cannot be changed during "form.post_submit", you should use "form.pre_submit" or "form.submit" instead.');
+    }
 }


src/Symfony/Component/Form/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Deprecate using `DateTime` or `DateTimeImmutable` model data with a different timezone than configured with the
    `model_timezone` option in `DateType`, `DateTimeType`, and `TimeType`
+ * Deprecate `FormEvent::setData()` for events that do not allow it
```
xabbuh: I think we can list both events here. And we then also need to add this to the upgrade file.

HeahDude: Done, thanks!

UPGRADE-6.4.md
```diff
@@ -29,6 +29,7 @@ Form
 
  * Deprecate using `DateTime` or `DateTimeImmutable` model data with a different timezone than configured with the
    `model_timezone` option in `DateType`, `DateTimeType`, and `TimeType`
+ * Deprecate `PostSetDataEvent::setData()` and `PostSubmitEvent::setData()`, methods will throw an exception
```
xabbuh:
```suggestion
 * Deprecate `PostSetDataEvent::setData()` and `PostSubmitEvent::setData()`, these methods will throw an exception starting with 7.0
```

src/Symfony/Component/Form/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Deprecate using `DateTime` or `DateTimeImmutable` model data with a different timezone than configured with the
    `model_timezone` option in `DateType`, `DateTimeType`, and `TimeType`
+ * Deprecate `PostSetDataEvent::setData()` and `PostSubmitEvent::setData()`
```
xabbuh: * Deprecate `PostSetDataEvent::setData()` and `PostSubmitEvent::setData()`, these methods will throw an exception starting with 7.0

UPGRADE-6.4.md
HeahDude: Done.

yceruto (approved):

xabbuh (approved):

src/Symfony/Component/Form/Event/PostSubmitEvent.php
```diff
@@ -21,4 +21,11 @@
  */
 final class PostSubmitEvent extends FormEvent
 {
+    /**
+     * @deprecated since Symfony 6.4, it will throw an exception in 7.0.
+     */
+    public function setData(mixed $data)
```
xabbuh:
```suggestion
    public function setData(mixed $data): void
```

src/Symfony/Component/Form/Event/PostSetDataEvent.php
```diff
@@ -20,4 +21,11 @@
  */
 final class PostSetDataEvent extends FormEvent
 {
+    /**
+     * @deprecated since Symfony 6.4, it will throw an exception in 7.0.
+     */
+    public function setData(mixed $data)
```
xabbuh:
```suggestion
    public function setData(mixed $data): void
```

xabbuh (changes requested):

src/Symfony/Component/Form/Event/PostSubmitEvent.php
HeahDude: Done, I did not notice these were final 🙈, thanks!
Should it be changed to `never` in 7.0?

xabbuh (approved):

derrabus (approved):

src/Symfony/Component/Form/Event/PostSetDataEvent.php
```diff
@@ -20,4 +21,11 @@
  */
 final class PostSetDataEvent extends FormEvent
 {
+    /**
+     * @deprecated since Symfony 6.4, it will throw an exception in 7.0.
+     */
+    public function setData(mixed $data): void
+    {
+        // throw new BadMethodCallException('Form data cannot be changed during "form.post_set_data", you should use "form.pre_set_data" instead.');
```
nicolas-grekas, line 30: missing runtime deprecation, isn't it? same in the other class

UPGRADE-6.4.md
```diff
@@ -29,6 +29,7 @@ Form
 
  * Deprecate using `DateTime` or `DateTimeImmutable` model data with a different timezone than configured with the
    `model_timezone` option in `DateType`, `DateTimeType`, and `TimeType`
+ * Deprecate `PostSetDataEvent::setData()` and `PostSubmitEvent::setData()`, these methods will throw an exception starting with 7.0
```
nicolas-grekas: let's give the solution instead of the new (useless) behavior; something like
```suggestion
 * Deprecate `PostSetDataEvent::setData()` and `PostSubmitEvent::setData()`, use `form.pre_set_data" instead` instead
```

src/Symfony/Component/Form/Event/PostSetDataEvent.php
HeahDude: Thanks, done.

UPGRADE-6.4.md
HeahDude: Done.

src/Symfony/Component/Form/Event/PostSubmitEvent.php
```diff
@@ -21,4 +21,12 @@
  */
 final class PostSubmitEvent extends FormEvent
 {
+    /**
+     * @deprecated since Symfony 6.4, it will throw an exception in 7.0.
+     */
+    public function setData(mixed $data): void
+    {
+        trigger_deprecation('symfony/form', '6.4', 'Calling "%s" will throw an exception as of 7.0, listen to "form.pre_submit" or "form.submit" instead.', __METHOD__);
```
xabbuh:
```suggestion
        trigger_deprecation('symfony/form', '6.4', 'Calling "%s()" will throw an exception as of 7.0, listen to "form.pre_submit" or "form.submit" instead.', __METHOD__);
```

src/Symfony/Component/Form/Event/PostSetDataEvent.php
```diff
@@ -20,4 +21,12 @@
  */
 final class PostSetDataEvent extends FormEvent
 {
+    /**
+     * @deprecated since Symfony 6.4, it will throw an exception in 7.0.
+     */
+    public function setData(mixed $data): void
+    {
+        trigger_deprecation('symfony/form', '6.4', 'Calling "%s" will throw an exception as of 7.0, listen to "form.pre_set_data" instead.', __METHOD__);
```
xabbuh:
```suggestion
        trigger_deprecation('symfony/form', '6.4', 'Calling "%s()" will throw an exception as of 7.0, listen to "form.pre_set_data" instead.', __METHOD__);
```

src/Symfony/Component/Form/Event/PostSubmitEvent.php
HeahDude: Done, thanks.

nicolas-grekas:
Thank you @HeahDude.

anyt:
Hi guys, 
As I've spent hours to get into the fact that [the method](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Form/Event/PostSubmitEvent.php#L27) now does literally nothing, while the business logic expects it to update data as it was in 5.4, I think it worth sharing.

I expect this to trigger exception in 6.4 as well, or trigger a deprecation and work as before. Otherwise it's a logical backward compatibility break that is hard to find because it only triggers a deprecation. 

From my perspective the deprecation is about using an obsolete way to do the thing, while here the method just does nothing and I have no choice except using the new approach to make it work back.

fancyweb:
Hello, please open a new issue. At first glance, I think we forgot to call `parent::setData()` after triggering the deprecation?

anyt:
Created, https://github.com/symfony/symfony/issues/53364. Thank you.

PR #51045, opened 2023-07-20
[Form] Use logical ordering for type and extension methods

HeahDude:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | ~
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/18585

Related docs PR has already been merged.

IMHO having methods ordered the same way that the component uses them gives a better understanding of how it works (same as having a "supports()" method defined first in a class like `Voter`).

It would be nice to reorder them in core implementations as well, by accepting such changes in 5.4, it would not cause conflicts when merging branches, WDYT?

### DIFF
--- a/.github/expected-missing-return-types.diff
+++ b/.github/expected-missing-return-types.diff
@@ -694,7 +694,7 @@ diff --git a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExt
 +    public function load(array $configs, ContainerBuilder $container): void
      {
          $loader = new PhpFileLoader($container, new FileLocator(\dirname(__DIR__).'/Resources/config'));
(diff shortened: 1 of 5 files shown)


maxbeckers (approved):

yceruto (approved):

nicolas-grekas:
Since this is not a bug fix, this should go on 6.4.
Conflicts don't happen if we follow this policy :)

javiereguiluz (approved):
And Nicolas is right about merging this in 6.4 because, strictly speaking, is not a bug fix. Thanks!

HeahDude:
Rebased on 6.4.

nicolas-grekas:
Thank you @HeahDude.

PR #51058, opened 2023-07-21
[FrameworkBundle] Add `--exclude` option to the `cache:pool:clear` command

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #51023
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/18595

For now this PR just ignores excluded pools/clearers when they don’t exist or wouldn’t be cleared/run anyways. Not sure what the best DX would be :thinking:

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
+++ b/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
@@ -29,6 +29,7 @@ CHANGELOG
  * Add support for relative URLs in BrowserKit's redirect assertion
  * Change BrowserKitAssertionsTrait::getClient() to be protected
  * Deprecate the `framework.asset_mapper.provider` config option
+ * Add `--exclude` option to the `cache:pool:clear` command
 
 6.3
 ---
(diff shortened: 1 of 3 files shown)


maxbeckers (approved):
Could also be done by not adding it to `$pools` and `$clearers` in the loop before instead of this `unset` solution, but fine for me this way.

Needs a related doc PR to explain the feature in docs.

nicolas-grekas:
Please always rebase, we don't accept PRs with merge commits in their history.

MatTheCat:
Sorry, tried to rebase using GitHub UI but it ended in a merge commit :sweat_smile: 
Should be good now.

chalasr:
We could deprecate `--all` and just clear all pools when the `pools` argument is not specified.

OskarStark:
> We could deprecate --all and just clear all pools when the pools argument is not specified.

To me this could be done in a separate PR, while this has some implicit drawbacks IMHO, because you could clear more than expected 👎 

I would go with this PR as is

OskarStark:
One last proposal from a DX perspective could be to give autocompletion for the possible values?

fabpot (approved):

fabpot:
Thank you @MatTheCat.

PR #51061, opened 2023-07-21
[DoctrineBridge] Bugfix - Allow to remove LazyLoaded listeners by object

### LINKED ISSUE #51057: [DoctrineBridge] Cannot remove Lazy-loaded listeners by object

6.3.x, but maybe lower

### Description

Let's say I the following code 
```
foreach ($this->em->getEventManager()->getAllListeners() as $event => $listeners) {
     foreach ($listeners as $listener) {
         $this->em->getEventManager()->removeEventListener($event, $listener);
    }
}
```

To me this is suppose to remove all the existing event listener.
Seems like with the `ContainerAwareEventManager` it does nothing.

If I dump the `listeners`, I get:
```
[
    'postPersist' => [
        '_service_App\EventListener\MyListener' => the object(App\EventListener\MyListener),
    ],
]
```

Which means that, when I call `removeEventListener($event, $listener)`, I pass the event and the object but in the remove method, the hash computed is the spl_object_hash and not the one with `_service` prefix.
https://github.com/symfony/symfony/blob/bd04cdaf6fc479a6f5602b862c190940b38a633e/src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php#L128-L146

I would say something is wrong with the `initializeListeners` method
https://github.com/symfony/symfony/blob/bd04cdaf6fc479a6f5602b862c190940b38a633e/src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php#L166-L176
because it transform the `$listener` as string to the object instance, but without changing the hash.

the $lister property is modified from 
```
[
    'postPersist' => [
        '_service_App\EventListener\MyListener' => "App\EventListener\MyListener",
    ],
]
```
to
```
[
    'postPersist' => [
        '_service_App\EventListener\MyListener' => the object(App\EventListener\MyListener),
    ],
]
```
when it should be
```
[
    'postPersist' => [
        'someHash' => the object(App\EventListener\MyListener),
    ],
]
```

### Possible Solution

First I thought of something like
Maybe the method should be
```
    private function initializeListeners(string $eventName): void
    {
        $this->initialized[$eventName] = true;
        foreach ($this->listeners[$eventName] as $hash => $listener) {
            if (\is_string($listener)) {
                $newHash = $this->getHash($listener);
                unset($this->listeners[$eventName][$hash]);
                
                $this->listeners[$eventName][$newHash] = $listener;
                $this->methods[$eventName][$newHash] = $this->getMethod($listener, $eventName);
            }
        }
    }
```

But it require listeners to be initialized then.
Which means that:
```
$listener = new MyListener();
$this->container->set('lazy', $listener);
$this->evm->addEventListener('foo', 'lazy');
$this->evm->removeEventListener('foo', $listener);
```
won't do nothing when 
```
$listener = new MyListener();
$this->container->set('lazy', $listener);
$this->evm->addEventListener('foo', 'lazy');
$this->evm->getAllListeners(); // Just to initialized
$this->evm->removeEventListener('foo', $listener);
```
will remove the lazy (now loaded) listener. Which would be weird.

### Reproducer

This is a reproducer in the Symfony Tests:
```
public function testRemoveAllEventListener()
    {
        $this->container->set('lazy', new MyListener());
        $this->evm->addEventListener('foo', 'lazy');
        $this->evm->addEventListener('foo', new MyListener());

        foreach ($this->evm->getAllListeners() as $event => $listeners) {
            foreach ($listeners as $listener) {
                $this->evm->removeEventListener($event, $listener);
            }
        }

        $this->assertSame([], $this->evm->getListeners('foo'));
    }
```

### Additional context

This seems cause by the PR https://github.com/symfony/symfony/pull/31335 and reported here https://github.com/symfony/symfony/pull/31335#issuecomment-562576326
@Koc @nicolas-grekas @dmaicher

VincentLanglet:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #51057
| License       | MIT

cc @dmaicher

### DIFF
--- a/src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
+++ b/src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
@@ -32,6 +32,7 @@ class ContainerAwareEventManager extends EventManager
     private $subscribers;
     private $initialized = [];
     private $initializedSubscribers = false;
+    private $initializedHashMapping = [];
     private $methods = [];
     private $container;
 
@@ -138,6 +139,7 @@ class ContainerAwareEventManager extends EventManager
 
             if (\is_string($listener)) {
                 unset($this->initialized[$event]);
+                unset($this->initializedHashMapping[$event][$hash]);
             } else {
                 $this->methods[$event][$hash] = $this->getMethod($listener, $event);
             }
@@ -158,6 +160,11 @@ class ContainerAwareEventManager extends EventManager
         $hash = $this->getHash($listener);
 
         foreach ((array) $events as $event) {
+            if (isset($this->initializedHashMapping[$event][$hash])) {
+                $hash = $this->initializedHashMapping[$event][$hash];
+                unset($this->initializedHashMapping[$event][$hash]);
+            }
+
             // Check if we actually have this listener associated
             if (isset($this->listeners[$event][$hash])) {
                 unset($this->listeners[$event][$hash]);
@@ -190,13 +197,25 @@ class ContainerAwareEventManager extends EventManager
     private function initializeListeners(string $eventName)
     {
         $this->initialized[$eventName] = true;
+
+        // We'll refill the whole array in order to keep the same order
+        $listeners = [];
         foreach ($this->listeners[$eventName] as $hash => $listener) {
             if (\is_string($listener)) {
-                $this->listeners[$eventName][$hash] = $listener = $this->container->get($listener);
+                $listener = $this->container->get($listener);
+                $newHash = $this->getHash($listener);
 
-                $this->methods[$eventName][$hash] = $this->getMethod($listener, $eventName);
+                $this->initializedHashMapping[$eventName][$hash] = $newHash;
+
+                $listeners[$newHash] = $listener;
+
+                $this->methods[$eventName][$newHash] = $this->getMethod($listener, $eventName);
+            } else {
+                $listeners[$hash] = $listener;
             }
         }
+
+        $this->listeners[$eventName] = $listeners;
     }
 
     private function initializeSubscribers()
--- a/src/Symfony/Bridge/Doctrine/Tests/ContainerAwareEventManagerTest.php
+++ b/src/Symfony/Bridge/Doctrine/Tests/ContainerAwareEventManagerTest.php
@@ -202,6 +202,21 @@ class ContainerAwareEventManagerTest extends TestCase
         $this->assertSame([], $this->evm->getListeners('foo'));
     }
 
+    public function testRemoveAllEventListener()
+    {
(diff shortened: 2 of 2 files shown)


src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
```diff
@@ -190,13 +201,25 @@ public function removeEventSubscriber(EventSubscriber $subscriber): void
     private function initializeListeners(string $eventName)
     {
         $this->initialized[$eventName] = true;
+
+        // We'll refill the whole array in order to keep the same order
+        $listeners = [];
```
VincentLanglet, line 202: If I just do
```
$this->listeners[$eventName][$newHash] = $listener;
```
the order of the listeners change.

I'm not sure if there is a better solution to keep the order

src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
```diff
@@ -158,6 +160,15 @@ public function removeEventListener($events, $listener)
         $hash = $this->getHash($listener);
 
         foreach ((array) $events as $event) {
+            if (!isset($this->initialized[$event])) {
+                $this->initializeListeners($event);
```
VincentLanglet: Only solution found so far for the `testRemoveLazyListenerInitialized`

src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
dmaicher: seems ok to me to just re-create it :man_shrugging: 

But maybe it would be good to add a test that makes sure we keep the order?

VincentLanglet: The test already exist. If I don't keep the same order, multiple tests are failing because of the `assertSame` checks

dmaicher (approved):

src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
nicolas-grekas: I wouldn't support this use case personally. Let's remove this part.

VincentLanglet: I removed it then @nicolas-grekas

nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @VincentLanglet.

PR #51062, opened 2023-07-21
Fix bug MapQueryString when request has only optional parameters

### LINKED ISSUE #50690: [HttpKernel] #[MapQueryString] with an empty query string always returns null

dddenisov:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50690
| License       | MIT
| Doc PR        | 

Based on the discussion here https://github.com/symfony/symfony/issues/50690 I propose my solution

### DIFF
--- a/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php
+++ b/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php
@@ -135,11 +135,16 @@ class RequestPayloadValueResolver implements ValueResolverInterface, EventSubscr
             }
 
             if (null === $payload) {
-                $payload = match (true) {
-                    $argument->metadata->hasDefaultValue() => $argument->metadata->getDefaultValue(),
-                    $argument->metadata->isNullable() => null,
-                    default => throw new HttpException($validationFailedCode)
-                };
+                try {
+                    $payload = match (true) {
+                        $argument->metadata->hasDefaultValue() => $argument->metadata->getDefaultValue(),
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Closing as this can't be the solution we're looking for (and because we don't even agree on the problem :) )
Thanks for proposing.

PR #51065, opened 2023-07-21
[AssetMapper] Fixing import parsing from jsdelivr

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | None - thanks to conversation with @bendavies on Slack!
| License       | MIT
| Doc PR        | n/a

Found with: https://cdn.jsdelivr.net/npm/bootstrap@4.6.2/+esm

With the `importmap:require --download` option, we parse those `import` statements to discover which dependencies we also need to download. If a package required 2 imports and both do *not* have an `@` in their name (i.e. a namespaced package), the regex matched both package strings as a single package - so it would see the package name as `jquery@3.7.0/+esm";import e from"/npm/popper.js@1.16.1/+esm`.

Cheers!


bendavies:
nice!

nicolas-grekas:
Thank you @weaverryan.

PR #51068, opened 2023-07-21
Add types to private and internal properties

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -


xabbuh:
The failures look related.

src/Symfony/Component/BrowserKit/AbstractBrowser.php
```diff
@@ -204,7 +204,7 @@ public function getCookieJar(): CookieJar
      */
     public function getCrawler(): Crawler
     {
-        if (null === $this->crawler) {
+        if (!isset($this->crawler)) {
```
stloyd: Why this change here? That variable has a `protected` status in this class.

src/Symfony/Component/Console/Completion/CompletionInput.php
```diff
@@ -31,11 +31,11 @@ final class CompletionInput extends ArgvInput
     public const TYPE_OPTION_NAME = 'option_name';
     public const TYPE_NONE = 'none';
 
-    private $tokens;
-    private $currentIndex;
-    private $completionType;
-    private $completionName;
-    private $completionValue = '';
+    private ?array $tokens = null;
```
stloyd: Both construct methods pass an array, not null:
```suggestion
    private array $tokens;
```

src/Symfony/Component/Process/Process.php
```diff
@@ -51,37 +50,35 @@ class Process implements \IteratorAggregate
     public const ITER_SKIP_OUT = 4;     // Use this flag to skip STDOUT while iterating
     public const ITER_SKIP_ERR = 8;     // Use this flag to skip STDERR while iterating
 
-    private $callback;
-    private $hasCallback = false;
-    private $commandline;
-    private $cwd;
-    private $env = [];
+    private ?\Closure $callback = null;
+    private array $commandline;
```
stloyd: This can be also `string` or `array` (of strings)

src/Symfony/Component/BrowserKit/AbstractBrowser.php
nicolas-grekas: This change anticipates a corresponding one on 7.0 when we'll add native types to public and protected properties.

derrabus: You could use `??` with a throw expression here.

```php
public function getCrawler(): Crawler
{
    return $this->crawler
        ?? throw new BadMethodCallException(sprintf('The "request()" method must be called before "%s()".', __METHOD__));
}
```

nicolas-grekas:
PR green, comments addressed.

PhilETaylor:
I did a composer update and now I get white screen of death (the actual error below) AMA

> Fatal error: Uncaught Error: Typed property Symfony\Component\Runtime\SymfonyRuntime::$input must not be accessed before initialization in /var/www/current/vendor/symfony/runtime/SymfonyRuntime.php:109 Stack trace: #0 /var/www/current/vendor/autoload_runtime.php(16): Symfony\Component\Runtime\SymfonyRuntime->__construct() #1 /var/www/current/public/index.php(13): require_once('...') #2 {main} thrown in /var/www/current/vendor/symfony/runtime/SymfonyRuntime.php on line 109


installed version 
`   symfony/runtime: '6.4.x-dev#58f074a10fed1094ca7a2bb753c36983790fdb29'
`

PhilETaylor:
Moving the `$this->getInput();` out of the `elseif` statement in the constructor of `SymfonyRuntime` seems to fix the issue for me but Im unsure of any knock on affects. 

from this:
```php
if (isset($options['env'])) {
            $_SERVER[$envKey] = $options['env'];
        } elseif (isset($_SERVER['argv']) && class_exists(ArgvInput::class)) {
            $this->options = $options;
            $this->getInput();
        }

```

to this

```php
if (isset($options['env'])) {
            $_SERVER[$envKey] = $options['env'];
        } elseif (isset($_SERVER['argv']) && class_exists(ArgvInput::class)) {
            $this->options = $options;
        }
        $this->getInput();
```

PR #51067, opened 2023-07-21
Use typed properties in tests as much as possible

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

PR #45360, opened 2022-02-08
[ErrorHandler] trigger deprecations for `@final` properties

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | Fix #43600
| License       | MIT
| Doc PR        | -

This PR makes DebugClassLoader trigger a deprecation when a property in a child class is redefined while the same property is `@final` on the parent class (the deprecation is silenced when both classes live in the exact same namespace or when the child property is deprecated.)

It also makes all properties in the `Symfony\`  namespace implicitly `@final`, unless they are typed. The goal here is to be able to add types to all properties in 7.0, thus fixing #43600.

PR #51069, opened 2023-07-21
Add types to public and protected properties

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

:hot_face:  

Allowed by https://github.com/symfony/symfony/pull/45360

Follows #51068 and #51067

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php
@@ -55,10 +55,7 @@ use Twig\Environment;
  */
 abstract class AbstractController implements ServiceSubscriberInterface
 {
-    /**
-     * @var ContainerInterface
-     */
-    protected $container;
+    protected ContainerInterface $container;
 
     #[Required]
     public function setContainer(ContainerInterface $container): ?ContainerInterface
--- a/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php
@@ -27,14 +27,9 @@ abstract class KernelTestCase extends TestCase
     use MailerAssertionsTrait;
     use NotificationAssertionsTrait;
 
-    protected static $class;
-
-    /**
-     * @var KernelInterface
-     */
-    protected static $kernel;
-
-    protected static $booted = false;
+    protected static ?string $class = null;
+    protected static ?KernelInterface $kernel = null;
+    protected static bool $booted = false;
 
     protected function tearDown(): void
     {
--- a/src/Symfony/Bundle/FrameworkBundle/Translation/Translator.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Translation/Translator.php
@@ -24,10 +24,9 @@ use Symfony\Component\Translation\Translator as BaseTranslator;
  */
 class Translator extends BaseTranslator implements WarmableInterface
 {
-    protected $container;
-    protected $loaderIds;
-
-    protected $options = [
+    protected ContainerInterface $container;
+    protected array $loaderIds;
+    protected array $options = [
         'cache_dir' => null,
         'debug' => false,
         'resource_files' => [],
--- a/src/Symfony/Component/BrowserKit/Cookie.php
+++ b/src/Symfony/Component/BrowserKit/Cookie.php
@@ -35,14 +35,14 @@ class Cookie
         'D M d H:i:s Y T',
     ];
 
-    protected $name;
-    protected $value;
-    protected $expires;
-    protected $path;
-    protected $domain;
-    protected $secure;
-    protected $httponly;
-    protected $rawValue;
+    protected string $name;
+    protected string $value;
+    protected ?string $expires = null;
+    protected string $path;
+    protected string $domain;
+    protected bool $secure;
+    protected bool $httponly;
+    protected string $rawValue;
     private ?string $samesite;
 
     /**
--- a/src/Symfony/Component/Config/Definition/BaseNode.php
+++ b/src/Symfony/Component/Config/Definition/BaseNode.php
@@ -29,17 +29,17 @@ abstract class BaseNode implements NodeInterface
     private static array $placeholderUniquePrefixes = [];
     private static array $placeholders = [];
 
-    protected $name;
-    protected $parent;
-    protected $normalizationClosures = [];
-    protected $normalizedTypes = [];
-    protected $finalValidationClosures = [];
-    protected $allowOverwrite = true;
-    protected $required = false;
-    protected $deprecation = [];
-    protected $equivalentValues = [];
-    protected $attributes = [];
-    protected $pathSeparator;
+    protected string $name;
+    protected ?NodeInterface $parent;
+    protected array $normalizationClosures = [];
+    protected array $normalizedTypes = [];
+    protected array $finalValidationClosures = [];
+    protected bool $allowOverwrite = true;
+    protected bool $required = false;
+    protected array $deprecation = [];
+    protected array $equivalentValues = [];
+    protected array $attributes = [];
+    protected string $pathSeparator;
 
     private mixed $handlingPlaceholder = null;
 
--- a/src/Symfony/Component/Config/Definition/Builder/NodeBuilder.php
+++ b/src/Symfony/Component/Config/Definition/Builder/NodeBuilder.php
@@ -18,8 +18,8 @@ namespace Symfony\Component\Config\Definition\Builder;
  */
 class NodeBuilder implements NodeParentInterface
 {
-    protected $parent;
-    protected $nodeMapping;
+    protected (NodeDefinition&ParentNodeDefinitionInterface)|null $parent = null;
+    protected array $nodeMapping;
 
     public function __construct()
     {
@@ -39,7 +39,7 @@ class NodeBuilder implements NodeParentInterface
      *
      * @return $this
      */
-    public function setParent(?ParentNodeDefinitionInterface $parent): static
+    public function setParent((NodeDefinition&ParentNodeDefinitionInterface)|null $parent): static
     {
         $this->parent = $parent;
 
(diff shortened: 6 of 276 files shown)


src/Symfony/Bundle/FrameworkBundle/Translation/Translator.php
```diff
@@ -24,16 +24,9 @@
  */
 class Translator extends BaseTranslator implements WarmableInterface
 {
-    protected $container;
-    protected $loaderIds;
-
-    protected $options = [
-        'cache_dir' => null,
```
94noni, line 31: passing by, are those removal desired or code swap typo?

src/Symfony/Component/Config/Definition/BaseNode.php
```diff
@@ -29,17 +29,17 @@ abstract class BaseNode implements NodeInterface
...
-    protected $normalizationClosures = [];
-    protected $normalizedTypes = [];
-    protected $finalValidationClosures = [];
-    protected $allowOverwrite = true;
-    protected $required = false;
-    protected $deprecation = [];
-    protected $equivalentValues = [];
-    protected $attributes = [];
-    protected $pathSeparator;
+    protected ?string $name;
```
derrabus:
```suggestion
    protected string $name;
```

The constructor makes sure that `$name` is not `null`.

https://github.com/symfony/symfony/blob/f10050724a1d2b5a9d57a30aa730be487d620283/src/Symfony/Component/Config/Definition/BaseNode.php#L51

src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php
```diff
@@ -56,10 +55,7 @@
  */
 abstract class AbstractController implements ServiceSubscriberInterface
 {
-    /**
-     * @var ContainerInterface
-     */
-    protected $container;
+    protected ContainerInterface $container;
```
derrabus, line 58: `setContainer()` will read this property _before_ writing it. Thus, we have to declare it nullable and initialize it with `null`.

Alternatively, we could adjust the access in `setContainer()`:

```php
$previous = $this->container ?? null;
```

src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php
```diff
@@ -27,14 +27,11 @@ abstract class KernelTestCase extends TestCase
...
     use NotificationAssertionsTrait;
 
-    protected static $class;
+    protected static ?string $class;
 
-    /**
-     * @var KernelInterface
-     */
-    protected static $kernel;
+    protected static ?KernelInterface $kernel;
```
derrabus:
```suggestion
    protected static ?KernelInterface $kernel = null;
```

Otherwise `ensureKernelShutdown()` will trip an uninitialized property error if the kernel hasn't been created.

src/Symfony/Component/BrowserKit/Cookie.php
```diff
@@ -35,14 +35,14 @@ class Cookie
...
-    protected $httponly;
-    protected $rawValue;
+    protected string $name;
+    protected string $value;
+    protected ?string $expires = null;
+    protected string $path;
+    protected string $domain;
+    protected bool $secure;
+    protected bool $httponly;
+    protected ?string $rawValue;
```
derrabus: I don't think this property should be nullable.

src/Symfony/Component/Cache/Adapter/DoctrineDbalAdapter.php
```diff
@@ -29,7 +29,7 @@
 
 class DoctrineDbalAdapter extends AbstractAdapter implements PruneableInterface
 {
-    protected $maxIdLength = 255;
+    protected int $maxIdLength = 255;
```
derrabus: Psalm has a point here: In `AbstractAdapterTrait` this property is declared nullable. Maybe we should stop overriding this property and move the assignment to the constructor?

src/Symfony/Component/Cache/Adapter/MemcachedAdapter.php
```diff
@@ -30,7 +30,7 @@ class MemcachedAdapter extends AbstractAdapter
     private const RESERVED_MEMCACHED = " \n\r\t\v\f\0";
     private const RESERVED_PSR6 = '@()\{}/';
 
-    protected $maxIdLength = 250;
+    protected int $maxIdLength = 250;
```
derrabus: same

src/Symfony/Component/Cache/Adapter/PdoAdapter.php
```diff
@@ -19,7 +19,7 @@
 
 class PdoAdapter extends AbstractAdapter implements PruneableInterface
 {
-    protected $maxIdLength = 255;
+    protected int $maxIdLength = 255;
```
derrabus: same

src/Symfony/Component/HttpFoundation/StreamedResponse.php
```diff
@@ -26,9 +26,10 @@
  */
 class StreamedResponse extends Response
 {
-    protected $callback;
-    protected $streamed;
-    private bool $headersSent;
+    protected \Closure $callback;
```
derrabus, line 29: `setCallback()` accepts any callable. You need to make sure we always assign a closure:

```php
public function setCallback(callable $callback): static
{
    $this->callback = $callback(...);
    return $this;
}
```

In addition to this, in `sendContent()` we check if `$this->callback` is null.

https://github.com/symfony/symfony/blob/f10050724a1d2b5a9d57a30aa730be487d620283/src/Symfony/Component/HttpFoundation/StreamedResponse.php#L90-L92

We might want to change this to an `isset()` check, I suppose.

derrabus (changes requested):
Reviewing this PR is exhausting. I've reviewed up to the HttpFoundation component and will take a break now. 😓

src/Symfony/Bundle/FrameworkBundle/Translation/Translator.php
nicolas-grekas: fixed

src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php
nicolas-grekas: alternative applied

src/Symfony/Bundle/FrameworkBundle/Routing/Router.php
```diff
@@ -61,7 +61,7 @@ public function __construct(ContainerInterface $container, mixed $resource, arra
 
     public function getRouteCollection(): RouteCollection
     {
-        if (null === $this->collection) {
+        if (!isset($this->collection)) {
```
derrabus: For 6.4? Otherwise FrameworkBundle 6.4 won't be compatible with Routing 7 once this PR is merged.

src/Symfony/Component/Config/Definition/Builder/NodeBuilder.php
```diff
@@ -18,8 +18,8 @@
  */
 class NodeBuilder implements NodeParentInterface
 {
-    protected $parent;
-    protected $nodeMapping;
+    protected ?ParentNodeDefinitionInterface $parent = null;
```
derrabus:
```suggestion
    protected (NodeDefinition&ParentNodeDefinitionInterface)|null $parent = null;
```

… in order to match the return type of `end()`. We could also think about adjusting the parameter type of `setParent()` accordingly.

src/Symfony/Component/Config/Definition/Builder/NodeDefinition.php
```diff
@@ -22,21 +22,21 @@
...
+    protected bool $default = false;
+    protected bool $required = false;
+    protected array $deprecation = [];
+    protected MergeBuilder $merge;
+    protected bool $allowEmptyValue = true;
+    protected mixed $nullEquivalent = null;
+    protected mixed $trueEquivalent = true;
+    protected mixed $falseEquivalent = false;
+    protected string $pathSeparator = BaseNode::DEFAULT_PATH_SEPARATOR;
+    protected NodeParentInterface|NodeInterface|null $parent;
```
derrabus, line 38: The return type of `end()` is wider. My gut feeling is that they should match.

nicolas-grekas:
PR is ready for review. Needs #51121 to be :green_apple:

src/Symfony/Bundle/FrameworkBundle/Routing/Router.php
derrabus: Okay, that's what #51121 is for. Nevermind.

src/Symfony/Component/Config/Definition/Builder/NodeDefinition.php
nicolas-grekas: All those classes in the return type of `end()` implement `NodeParentInterface` so we could reduce this to only it. I suppose the list is here to hint the IDE for autocompletion. Maybe we should keep only `NodeParentInterface` as native type and use `@return` for the rest?

The other strange thing is about the need to add `NodeInterface` to make this code work:
https://github.com/symfony/symfony/blob/1935af612953715faec8b69f2be70b20c7427bc5/src/Symfony/Component/Config/Definition/Builder/ArrayNodeDefinition.php#L352-L361

`end()` can't return an `ArrayNode` because of its return type. Either the return type is wrong, or what?
Up to investigate and fix this on a lower branch?

derrabus: > All those classes in the return type of `end()` implement `NodeParentInterface` so we could reduce this to only it. I suppose the list is here to hint the IDE for autocompletion. Maybe we should keep only `NodeParentInterface` as native type and use `@return` for the rest?

Would work for me to unblock the PR.

> Up to investigate and fix this on a lower branch?

Maybe, but not right away. 🙈

nicolas-grekas: > use @return for the rest

PR updated

derrabus (approved):

src/Symfony/Component/Config/Definition/Builder/NodeDefinition.php
stof: The return type was not wrong. ArrayNode is not a NodeDefinition.

PR #39712, opened 2021-01-04
[Messenger] Kafka Transport Bridge

KonstantinCodes:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Tickets       | Fix #35549 
| License       | MIT
| Doc PR        | symfony/symfony-docs#15884

This PR integrates a simple Kafka Transport for Messenger.

PR #51070, opened 2023-07-21
[Messenger] Add Kafka bridge

andythorne:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | ~
| License       | MIT
| Doc PR        | symfony/symfony-docs (working on atm)

Adds support for Kafka as a messenger transport.

> I know there's [another PR open for a kafka messenger implementation](https://github.com/symfony/symfony/pull/39712), but it appears to be abandoned. This PR is based on https://github.com/symfony-examples/messenger-kafka, which I believe to be a better implementation.


.github/workflows/integration-tests.yml
```diff
@@ -135,7 +135,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           coverage: "none"
-          extensions: "json,couchbase-3.2.2,memcached,mongodb-1.12.0,redis,rdkafka,xsl,ldap,relay"
+          extensions: "json,couchbase-3.2.2,memcached,mongodb-1.12.0,redis,rdkafka,xsl,ldap,relay,rdkafka"
```
AurelienPillevesse: ```rdkafka``` was already present in this string, it is now duplicated

AurelienPillevesse (changes requested):

.github/workflows/integration-tests.yml
andythorne: Whoops, fixed!

src/Symfony/Component/Messenger/Bridge/Kafka/Transport/KafkaReceiver.php
```diff
@@ -0,0 +1,74 @@
...
+        /** @var KafkaReceivedMessageStamp $transportStamp */
+        $transportStamp = $envelope->last(KafkaReceivedMessageStamp::class);
+
+        $this->connection->ack($transportStamp->message);
+    }
+
+    /** @SuppressWarnings(PHPMD.UnusedFormalParameter) */
+    public function reject(Envelope $envelope): void
+    {
+        // no reject method for kafka transport
```
valtzu: From my experience, when the consumer is configured with

```yaml
enable.auto.offset.store: false
enable.auto.commit: false
```

the offset is only committed when explicitly doing `ack` call. This means in case of an error, the message is never acknowledged / the offset is never committed, hence the consumer will start lagging and won't be able to consume new messages. In some/many cases this may be intended, especially when the ordering of messages matter – but it's also possible that one does not want to block the consumer when there's a single failing message. Then if you have retry configured on the same transport, you start producing retry messages which never get consumed because the consumer is stuck on the initial message (given you only have 1 consumer or 1 consumer group).

I'd like to see a configuration variable to make ack call on reject, a bit similar to Redis transports' `delete_after_reject`, say `commit_offset_after_reject` or `ack_after_reject`.

---

Anyway, would be nice to see this Kafka transport in the core :+1:

src/Symfony/Component/Messenger/Bridge/Kafka/Transport/Connection.php
```diff
@@ -0,0 +1,311 @@
...
+
+    private bool $consumerIsSubscribed = false;
+    private ?KafkaConsumer $consumer = null;
+    private ?Producer $producer = null;
+
+    private KafkaFactory $kafkaFactory;
+
+    /**
+     * @psalm-param array{topics: list<string>, consume_timeout_ms: int, commit_async: bool, conf_options: array<string, string>} $consumerConfig
+     * @psalm-param array{topic: string, poll_timeout_ms: int, flush_timeout_ms: int, conf_options: array<string, string>} $producerConfig
```
stof: should be `@param`. The array shape config is interoperable between tools (at least as long as none of the keys are a non-lowercase variant of a phpdoc keyword as defined by the psalm parser because of https://github.com/vimeo/psalm/issues/10008)

src/Symfony/Component/Messenger/Bridge/Kafka/Transport/Connection.php
```diff
@@ -0,0 +1,311 @@
...
+
+        return new self(
+            self::setupConsumerOptions($parsedUrl['host'], $options['consumer'] ?? []),
+            self::setupProducerOptions($parsedUrl['host'], $options['producer'] ?? []),
+            $logger,
+            $kafkaFactory,
+        );
+    }
+
+    /** @psalm-param array<string, bool|float|int|string|array<string>> $configOptions */
```
stof: should be `@param`

src/Symfony/Component/Messenger/Bridge/Kafka/Transport/Connection.php
```diff
@@ -0,0 +1,311 @@
...
+    /** @psalm-param array<string, string> $headers */
+    public function publish(int $partition, int $messageFlags, string $body, string $key = null, array $headers = []): void
+    {
+        if (!$this->producerConfig['topic']) {
+            throw new LogicException('No topic configured for the producer.');
+        }
+
+        $producer = $this->getProducer();
+
+        /** @psalm-var \RdKafka\ProducerTopic $topic */
```
stof: should be `@var` (but is this actually needed ? Can't psalm infer it thanks to the return type of `newTopic` ?)

src/Symfony/Component/Messenger/Bridge/Kafka/Transport/KafkaTransportFactory.php
```diff
@@ -0,0 +1,49 @@
...
+
+class KafkaTransportFactory implements TransportFactoryInterface
+{
+    private KafkaFactory $kafkaFactory;
+
+    public function __construct(
+        private LoggerInterface $logger = new NullLogger(),
+        KafkaFactory $kafkaFactory = null,
+    ) {
+        if (!$kafkaFactory instanceof KafkaFactory) {
```
stof: why are you ignoring it when passing a KafkaFactory instead of assigning it ?

To me, the code should be like this:

```php
    public function __construct(
        private LoggerInterface $logger = new NullLogger(),
        KafkaFactory $kafkaFactory = null,
    ) {
            $this->kafkaFactory = $kafkaFactory ?? new KafkaFactory(
                new LoggingLogCallback($logger),
                new LoggingErrorCallback($logger),
                new LoggingRebalanceCallback($logger),
            );
    }
```

src/Symfony/Component/Messenger/Bridge/Kafka/Transport/KafkaTransport.php
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public function send(Envelope $envelope): Envelope
+    {
+        return $this->getSender()->send($envelope);
+    }
+
+    private function getReceiver(): KafkaReceiver
+    {
+        return $this->receiver ??= new KafkaReceiver($this->connection, $this->serializer);
```
stof, line 52: is there an actual benefit about lazy-loading those 2 objects ? Their constructor only assigns the properties

stof: And what is the benefit of splitting the logic into those separate objects if they are not available as extension points by allowing to inject them in the constructor, as done in some other bridges ?

src/Symfony/Component/Messenger/Bridge/Kafka/Transport/KafkaOption.php
```diff
@@ -0,0 +1,211 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Messenger\Bridge\Kafka\Transport;
+
+/**
+ * @SuppressWarnings(PHPMD.ExcessiveMethodLength)
```
stof: should be removed. Symfony does not use PHPMD at all.

src/Symfony/Component/Messenger/Bridge/Kafka/Transport/KafkaOption.php
```diff
@@ -0,0 +1,211 @@
...
+ *
+ * @see https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md
+ */
+final class KafkaOption
+{
+    /** @psalm-return array<string, string> */
```
stof: should be `@var`

src/Symfony/Component/Messenger/Bridge/Kafka/Transport/KafkaReceiver.php
```diff
@@ -0,0 +1,74 @@
...
+    ) {
+    }
+
+    /** @psalm-return \Traversable<Envelope> */
+    public function get(): iterable
+    {
+        yield from $this->getEnvelope();
+    }
+
+    /** @SuppressWarnings(PHPMD.UnusedFormalParameter) */
```
stof: should be removed

src/Symfony/Component/Messenger/Bridge/Kafka/Transport/KafkaReceiver.php
```diff
@@ -0,0 +1,74 @@
...
+    public function ack(Envelope $envelope): void
+    {
+        /** @var KafkaReceivedMessageStamp $transportStamp */
```
stof: should be removed for 2 reasons:

- this `@var` is a lie: the method can return null
- the generic types on `Envelope::last` already allows static analysis tool to infer that the returned object is a KafkaReceivedMessageStamp (when it is not null)

src/Symfony/Component/Messenger/Bridge/Kafka/Transport/KafkaReceiver.php
```diff
@@ -0,0 +1,74 @@
...
+        $this->connection->ack($transportStamp->message);
+    }
+
+    /** @SuppressWarnings(PHPMD.UnusedFormalParameter) */
+    public function reject(Envelope $envelope): void
+    {
+        // no reject method for kafka transport
+    }
+
+    /** @psalm-return iterable<Envelope> */
```
stof: should be `@return`

src/Symfony/Component/Messenger/Bridge/Kafka/Transport/KafkaReceiver.php
```diff
@@ -0,0 +1,74 @@
...
+            $kafkaMessage = $this->connection->get();
+        } catch (\RdKafka\Exception $exception) {
+            throw new TransportException($exception->getMessage(), 0, $exception);
+        }
+
+        if (\RD_KAFKA_RESP_ERR_NO_ERROR !== $kafkaMessage->err) {
+            switch ($kafkaMessage->err) {
+                case \RD_KAFKA_RESP_ERR__PARTITION_EOF: // No more messages
+                case \RD_KAFKA_RESP_ERR__TIMED_OUT: // Attempt to connect again
+                    return [];
```
stof: this should be `return;`, not `return []` as the method is a Generator

src/Symfony/Component/Messenger/Bridge/Kafka/Transport/KafkaReceiver.php
```diff
@@ -0,0 +1,74 @@
...
+
+class KafkaReceiver implements ReceiverInterface
+{
+    public function __construct(
+        private Connection $connection,
+        private SerializerInterface $serializer = new PhpSerializer(),
+    ) {
+    }
+
+    /** @psalm-return \Traversable<Envelope> */
```
stof: should be removed. The type is defined by the interface

src/Symfony/Component/Messenger/Bridge/Kafka/Transport/KafkaReceiver.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /** @SuppressWarnings(PHPMD.UnusedFormalParameter) */
+    public function reject(Envelope $envelope): void
+    {
+        // no reject method for kafka transport
+    }
+
+    /** @psalm-return iterable<Envelope> */
+    private function getEnvelope(): iterable
```
stof: why making a private method that returns an iterable when we can only have 0 or 1 element ? Returning `Envelope|null` would make the code easier to understand.

src/Symfony/Component/Messenger/Bridge/Kafka/Transport/KafkaFactory.php
```diff
@@ -0,0 +1,92 @@
...
+class KafkaFactory
+{
+    public function __construct(
+        private readonly mixed $logCb = null,
+        private readonly mixed $errorCb = null,
+        private readonly mixed $rebalanceCb = null,
+        private readonly mixed $deliveryReportMessageCb = null,
+        private readonly mixed $offsetCommitCb = null,
+        private readonly mixed $statsCb = null,
+        private readonly mixed $consumeCb = null,
```
stof: the constructor should expect `callable|null`, not `mixed`

src/Symfony/Component/Messenger/Bridge/Kafka/Stamp/KafkaMessageStamp.php
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Messenger\Bridge\Kafka\Stamp;
+
+use Symfony\Component\Messenger\Stamp\NonSendableStampInterface;
+
+final class KafkaMessageStamp implements NonSendableStampInterface
```
stof, line 16: should those stamps be `@internal` or no ?

src/Symfony/Component/Messenger/Bridge/Kafka/Callback/LoggingRebalanceCallback.php
```diff
@@ -0,0 +1,107 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Messenger\Bridge\Kafka\Callback;
+
+use Psr\Log\LoggerInterface;
+use RdKafka\KafkaConsumer;
+use RdKafka\TopicPartition;
+
+final class LoggingRebalanceCallback
```
stof: should those be `@internal` or no ?

src/Symfony/Component/Messenger/Bridge/Kafka/Callback/LoggingRebalanceCallback.php
```diff
@@ -0,0 +1,107 @@
...
+{
+    public function __construct(
+        private readonly LoggerInterface $logger,
+    ) {
+    }
+
+    /**
+     * @param list<TopicPartition>|null $topicPartitions
+     */
+    public function __invoke(KafkaConsumer $kafka, ?int $err, array $topicPartitions = null): void
```
stof: why making the second argument nullable and the third argument nullable and optional ? this does not match the documented signature of the rebalance callback where those are always passed.

Making them nullable means that the code inside them has to deal with `null` values which this may not be necessary.

src/Symfony/Component/Messenger/Bridge/Kafka/Callback/LoggingLogCallback.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    public function __construct(
+        private readonly LoggerInterface $logger,
+    ) {
+    }
+
+    public function __invoke(object $kafka, int $level, string $facility, string $message): void
+    {
+        $this->logger->log(
+            $level,
```
stof: this is broken. you need to convert the kafka level into a `\Psr\Log\LogLevel` value. Otherwise, you have no guarantee that things will work fine as the only mandatory supported values for the `log` method in PSR-3 are those string constants. Any other supported value for the level is implementation-defined and so you cannot rely on it when targeting only the PSR-3 interface without knowing the implementation.

src/Symfony/Component/Messenger/Bridge/Kafka/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+CHANGELOG
+=========
+
+6.4
+---
+
+ * Introduced the Kafka bridge.
```
stof: Please follow the contributing rules for changelog entries: https://symfony.com/doc/current/contributing/code/conventions.html#writing-a-changelog-entry

```suggestion
 * Introduce the Kafka bridge.
```

stof (commented):
I haven't reviewed the kafka-specific code to validate it given I have no knowledge of Kafka. So my review does not guarantee that this is working code.

src/Symfony/Component/Messenger/Bridge/Kafka/Transport/KafkaTransport.php
andythorne: This seems to be inline with the other messenger bridge transports. I can update it to be injected if you prefer?

chalasr: Indeed, only the [AmazonSQSTransport](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Messenger/Bridge/AmazonSqs/Transport/AmazonSqsTransport.php) properly exposes these dependencies through DI, which makes sense as it's the most recent one. I agree we should do the same here

src/Symfony/Component/Messenger/Bridge/Kafka/Transport/KafkaTransport.php
```diff
@@ -0,0 +1,59 @@
...
+
+use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\Transport\Serialization\PhpSerializer;
+use Symfony\Component\Messenger\Transport\Serialization\SerializerInterface;
+use Symfony\Component\Messenger\Transport\TransportInterface;
+
+class KafkaTransport implements TransportInterface
+{
+    private KafkaReceiver $receiver;
+    private KafkaSender $sender;
```
chalasr, line 22: these should also use the proper interfaces respectively instead of the concrete implementations, to allow for decoration.

AlexOstrovsky:
Is there something I can help with to move this forward? we are currently using this package: https://packagist.org/packages/koco/messenger-kafka but it would be nice to have kafka support native in symfony. 

Thank you

Dodenis:
I agree with AlexOstrovsky. Is there anything we can do to help this PR? It would be really great to have native KAFKA support in Messenger. 

Thanks!

AlexOstrovsky:
@chalasr @andythorne ping to know what we can do to help to move this futher? 👏

Thanks!

.github/workflows/psalm.yml
```diff
@@ -22,10 +22,21 @@ jobs:
...
+
+          echo "::group::install tools & libraries"
+          sudo apt-get install librdkafka-dev
+          echo "::endgroup::"
+
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ env.php-version }}
+          extensions: "rdkafka"
```
GromNaN, line 39: psalm has stubs for this extension, it should not be required to install it: https://github.com/vimeo/psalm/blob/6.x/stubs/extensions/rdkafka.phpstub

src/Symfony/Component/Messenger/Bridge/Kafka/Callback/CallbackProcessorInterface.php
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Symfony\Component\Messenger\Bridge\Kafka\Callback;
+
+use RdKafka\KafkaConsumer;
+use RdKafka\Message;
+use RdKafka\Producer;
+use RdKafka\TopicPartition;
+
+/**
+ * @see https://arnaud.le-blanc.net/php-rdkafka-doc/phpdoc/class.rdkafka-conf.html for more information on callback parameters.
```
GromNaN, line 20: This link is not working.

It would be wise to document the methods rather than linking to a page that may change over time.

src/Symfony/Component/Messenger/Bridge/Kafka/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+CHANGELOG
+=========
+
+6.4
```
GromNaN, line 4: This needs to be updated
```suggestion
7.4
```

AlexOstrovsky:
@chalasr any updaton this please? anythig how I can help?

PR #51073, opened 2023-07-22
[Intl] Add support for ISO 3166-1 numeric codes

### LINKED ISSUE #50860: Add ISO-3166-1 Numeric codes to Intl component
### Description

The Intl component is supporting access to 2 and 3 letter ISO codes for countries. 

However, there is a third, numeric type which is not currently supported. 
https://www.iso.org/obp/ui/#search/code/

Currently my requirement is to convert a 2 letter country code (FR) into a numeric equivalent (250) and it looks like I'm going to either have to install another package (https://github.com/thephpleague/iso3166) or hard-code a static list in my codebase.

Would it be acceptable to add these numeric codes to the Intl component?

Thanks!

### Example

_No response_

benr77:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #50860
| License       | MIT

Add support for ISO-3166-1 numeric codes.

As well as the existing Alpha2 and Alpha3 codes, this PR adds support for the numeric country codes as per https://www.iso.org/obp/ui/#search/code/.


src/Symfony/Component/Intl/Resources/data/regions/meta.php
```diff
@@ -754,4 +754,588 @@
...
+        'XY' => '997',
+        'XZ' => '998',
+        'YE' => '887',
+        'YT' => '175',
+        'ZA' => '710',
+        'ZM' => '894',
+        'ZW' => '716',
+        'ZZ' => '999',
+    ],
+    'NumericToAlpha2' => [
```
benr77, line 1053: Not sure the best way to handle these numeric strings with leading zeros as array keys - we have a mix of int and string keys.

src/Symfony/Component/Intl/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+5.4
```
94noni: you should target 6.4 as new feature :)

src/Symfony/Component/Intl/Resources/data/regions/meta.php
94noni: perhaps sanitize all to string?

src/Symfony/Component/Intl/Resources/data/regions/meta.php
```diff
@@ -754,4 +754,588 @@
...
+        600 => 'PY',
+        634 => 'QA',
+        959 => 'QM',
+        960 => 'QN',
+        962 => 'QP',
+        963 => 'QQ',
+        964 => 'QR',
+        965 => 'QS',
+        966 => 'QT',
+        'QV',
```
94noni: is this normal?

src/Symfony/Component/Intl/Tests/CountriesTest.php
```diff
@@ -665,4 +958,45 @@ public function testGetAlpha3Names($displayLocale)
...
+
+    /**
+     * @group intl-numeric-code-data
+     */
+    public function testNumericCodeExists()
+    {
+        $this->assertTrue(Countries::numericCodeExists('250'));
+        $this->assertTrue(Countries::numericCodeExists('982'));
+        $this->assertTrue(Countries::numericCodeExists('716'));
+        $this->assertTrue(Countries::numericCodeExists('036'));
```
94noni, line 974: test one false?

src/Symfony/Component/Intl/CHANGELOG.md
94noni: I mean mostly the target branch of this PR :)

src/Symfony/Component/Intl/Resources/data/regions/meta.php
benr77: I've forced all the array keys to be numeric strings, prefixed with an underscore, to stop PHP trying to convert string numbers such as '077' to (int) 77.

src/Symfony/Component/Intl/CHANGELOG.md
benr77: Surely it can be merged into older versions as there are no BC breaks? Is it a new feature or a bug fix if it's completing the missing ISO 3166-1 support for numeric codes? I don't mind either way just wanted to clarify...

derrabus: A PR title that starts with "Add support for…" hardly passes as a bugfix. 😉 Let's target 6.4.

benr77:
Unit test failures appear to be unrelated to the PR.

src/Symfony/Component/Intl/Resources/data/regions/meta.php
```diff
@@ -754,4 +763,588 @@
...
+        'XW' => '995',
+        'XX' => '996',
+        'XY' => '997',
+        'XZ' => '998',
+        'YE' => '887',
+        'YT' => '175',
+        'ZA' => '710',
+        'ZM' => '894',
+        'ZW' => '716',
+        'ZZ' => '999',
```
ro0NL: perhaps filter DENYLIST too

ro0NL (approved):

nicolas-grekas (approved):

src/Symfony/Component/Intl/Countries.php
```diff
@@ -52,6 +52,20 @@ public static function getAlpha3Codes(): array
...
     }
 
+    /**
+     * Returns all available numeric country codes (3 digits).
+     *
+     * Countries are returned as ISO 3166 numeric three-digit country codes.
+     *
+     * This list only contains "officially assigned ISO 3166-1 numeric" country codes.
+     *
+     * @return string[]
```
stof: this should be documented as `@return array<string, string>` to be more precise about the type of keys being returned, as this returns a mapping with alpha2 codes as keys (this should probably also document what the keys are)

src/Symfony/Component/Intl/Tests/CountriesTest.php
```diff
@@ -665,4 +953,58 @@ public function testGetAlpha3Names($displayLocale)
...
+     */
+    public function testGetNumericCode()
+    {
+        foreach (self::COUNTRIES as $country) {
+            $this->assertSame(self::ALPHA2_TO_NUMERIC[$country], Countries::getNumericCode($country));
+        }
+    }
+
+    /**
+     * @group intl-numeric-code-data
```
stof: this group should be removed from those tests

stof (changes requested):

fabpot (approved):

fabpot:
Thank you @benr77.

PR #51075, opened 2023-07-23
[Serializer] More intuitive Logic Exception message

StaffNowa:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| License       | MIT

More intuitive Logic Exception message


derrabus (approved):
Thank you. The change looks good, but I don't know if it passes as a bugfix. If we see it as an improvement, 6.4 should be targeted instead. Maybe @nicolas-grekas can advise?

chalasr (approved):
For 6.4

chalasr:
Thank you @StaffNowa.

PR #50983, opened 2023-07-14
[Serializer] Deprecate annotations in favor of attributes

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | Follows #50888
| License       | MIT
| Doc PR        | symfony/symfony-docs#18589

This PR deprecates using Doctrine annotations to configure serialization. Attributes shall be used instead.

Existing applications can be migrated easily using [Rector](https://getrector.com/blog/how-to-upgrade-annotations-to-attributes).

Deprecation errors triggered by the bundles' functional tests will be resolved once https://github.com/symfony/symfony/pull/50888 is merged.

PR #51080, opened 2023-07-24
[Serializer] Remove Doctrine annotations support

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Follows #50983 
| License       | MIT
| Doc PR        | symfony/symfony-docs#18589


src/Symfony/Component/Serializer/Mapping/Loader/AnnotationLoader.php
```diff
@@ -43,14 +42,6 @@ class AnnotationLoader implements LoaderInterface
         Context::class,
     ];
 
-    public function __construct(
-        private readonly ?Reader $reader = null,
-    ) {
-        if ($reader) {
-            trigger_deprecation('symfony/validator', '6.4', 'Passing a "%s" instance as argument 1 to "%s()" is deprecated, pass null or omit the parameter instead.', get_debug_type($reader), __METHOD__);
-        }
-    }
```
derrabus, line 52: Question: should I keep an empty constructor so we won't break overrides that call `parent::__construct()`?

nicolas-grekas: no need to IMHO

nicolas-grekas:
Thank you @derrabus.

PR #51081, opened 2023-07-24
[HttpFoundation] Serialize floats into json response

spolischook:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Tickets       | -
| License       | MIT
| Doc PR        | -

There is an unexpected behavior with float numbers in JsonResponse which is became an integer in case of decimal part is null, e.g. 1.0. So in some cases it will be int in others float, which is not ideal

### DIFF
--- a/src/Symfony/Component/HttpFoundation/JsonResponse.php
+++ b/src/Symfony/Component/HttpFoundation/JsonResponse.php
@@ -27,9 +27,9 @@ class JsonResponse extends Response
     protected $data;
     protected $callback;
 
-    // Encode <, >, ', &, and " characters in the JSON, making it also safe to be embedded into HTML.
-    // 15 === JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_AMP | JSON_HEX_QUOT
-    public const DEFAULT_ENCODING_OPTIONS = 15;
+    // Encode <, >, ', &, and " characters in the JSON, making it also safe to be embedded into HTML, represents floats.
+    // 1039 === JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_AMP | JSON_HEX_QUOT | JSON_PRESERVE_ZERO_FRACTION
+    public const DEFAULT_ENCODING_OPTIONS = 1039;
 
     protected $encodingOptions = self::DEFAULT_ENCODING_OPTIONS;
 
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/HttpFoundation/JsonResponse.php
```diff
@@ -27,9 +27,9 @@ class JsonResponse extends Response
     protected $data;
     protected $callback;
 
-    // Encode <, >, ', &, and " characters in the JSON, making it also safe to be embedded into HTML.
-    // 15 === JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_AMP | JSON_HEX_QUOT
-    public const DEFAULT_ENCODING_OPTIONS = 15;
+    // Encode <, >, ', &, and " characters in the JSON, making it also safe to be embedded into HTML, represents floats.
+    // 1039 === JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_AMP | JSON_HEX_QUOT | JSON_PRESERVE_ZERO_FRACTION
+    public const DEFAULT_ENCODING_OPTIONS = 1039;
```
stloyd, line 32: This change is a hard BC break, why should we change that? There is a way to change the encoding options to allow your application to have different ones than the default one.

spolischook: do you mean `setEncodingOptions` method?

stloyd: Yes, exactly that one, I think this will work for this case:
```
$response->setEncodingOptions(JsonResponse::DEFAULT_ENCODING_OPTIONS | \JSON_PRESERVE_ZERO_FRACTION);
```

spolischook: yes, it is how it works now and I thought it would be the right choice for default options, but it seems it is too late to change this 😔

derrabus:
> There is an unexpected behavior with float numbers in JsonResponse which is became an integer

And this is where you are wrong.

One of the fun parts of the JSON spec is that it does not define "float" and "integer" as data types. There's only one type called "number". That means that `42.0` and `42` are actually two equal representations of the same value. The behavior of the JSON encoder is correct. But since it's not the only correct way, there's a flag to fine-tune the behavior in that regard.

The thing about defaults is that you won't find one that makes everybody happy. If we did merge your change, we'll probably soon get a PR by someone else who wants to change it back because they liked the old defaults more.

I'd like to close this PR because arguing about defaults is a waste of everybody's time and because there is a way for you to get the result you want.

However, I thank you for your proposal.

PR #51082, opened 2023-07-24
[Routing] Remove Doctrine annotations support

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Follows #49358
| License       | MIT
| Doc PR        | symfony/symfony-docs#18589

PR #51084, opened 2023-07-24
[Routing] Fix `testMissingPrefixLocale` and `testMissingRouteLocale`

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | https://github.com/symfony/symfony/pull/51082#discussion_r1272195743
| License       | MIT
| Doc PR        | not needed

Those two fixtures only exist as annotation versions. Their attribute counterparts were missing.

The corresponding tests only expected a `LogicException` which is what `AnnotationClassLoader` would raise for undefined classes as well.

This PR adds new assertions for the exception messages which would have discovered the missing fixtures. Finally, the two missing fixtures are added.


nicolas-grekas:
Thank you @derrabus.

PR #51091, opened 2023-07-24
[MonologBridge] Remove support for monolog < 3.0

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

cc @seldaek


composer.json
```diff
@@ -136,7 +136,7 @@
         "guzzlehttp/promises": "^1.4",
         "league/html-to-markdown": "^5.0",
         "masterminds/html5": "^2.7.2",
-        "monolog/monolog": "^1.25.1|^2",
+        "monolog/monolog": "^3",
```
nicolas-grekas: we don't allow v3 on branch 6.4? that's surprising. can't we fix that?

src/Symfony/Bridge/Monolog/Command/ServerLogCommand.php
```diff
@@ -145,7 +146,7 @@ private function getLogs($socket): iterable
         }
     }
 
-    private function displayLog(OutputInterface $output, int $clientId, array $record): void
+    private function displayLog(OutputInterface $output, int $clientId, LogRecord $record): void
     {
         if (isset($record['log_id'])) {
```
nicolas-grekas: WDYT of psalm's findings?

src/Symfony/Bridge/Monolog/Handler/MailerHandler.php
```diff
@@ -43,31 +40,22 @@ public function handleBatch(array $records): void
...
-                }
-                $messages[] = $this->processRecord($record);
-            }
-        } else {
-            foreach ($records as $record) {
-                if ($record['level'] < $this->level) {
-                    continue;
-                }
-                $messages[] = $this->processRecord($record);
+        /** @var LogRecord $record */
```
nicolas-grekas: better move this to an `@param` annotation

nicolas-grekas (commented):
thanks for taking care of this, here are some minor comments

composer.json
lyrixx: Indeed, I guess it's because people don't use symfony/symfony anymore.

Anyway, I'll open a PR on 6.4 to add some deprecation notice

derrabus: If we do that, the bridge won't be tested with Monolog 2 anymore, will it?

Currently:
* same-deps run: Monolog 2
* high-deps run: Monolog 3
* low-deps run: Monolog 1

src/Symfony/Bridge/Monolog/Formatter/ConsoleFormatter.php
```diff
@@ -98,11 +96,9 @@ public function formatBatch(array $records): mixed
         return $records;
     }
 
-    private function doFormat(array|LogRecord $record): mixed
+    public function format(LogRecord $record): mixed
     {
-        if ($record instanceof LogRecord) {
-            $record = $record->toArray();
-        }
+        $record = $record->toArray();
```
Seldaek: Strictly speaking here you don't need to do this anymore I think, you could simply work with $record values below, this was done for monolog2 compat.

src/Symfony/Bridge/Monolog/Handler/NotifierHandler.php
```diff
@@ -57,13 +55,13 @@ public function handleBatch(array $records): void
...
-            $notification = Notification::fromThrowable($record['context']['exception']);
+        if (($record->context['exception'] ?? null) instanceof \Throwable) {
+            $notification = Notification::fromThrowable($record->context['exception']);
         } else {
-            $notification = new Notification($record['message']);
+            $notification = new Notification($record->message);
         }
 
-        $notification->importanceFromLogLevelName(Logger::getLevelName($record['level']));
+        $notification->importanceFromLogLevelName($record->level->getName());
```
Seldaek, line 64: This looks to me like a psalm bug, if it cannot determine that a const array of strings is of type `string[]`.

src/Symfony/Bridge/Monolog/Processor/DebugProcessor.php
```diff
@@ -31,37 +29,29 @@ public function __construct(RequestStack $requestStack = null)
...
-            case Logger::CRITICAL:
-            case Logger::ALERT:
-            case Logger::EMERGENCY:
+        switch ($record->level) {
+            case Level::Error:
+            case Level::Critical:
+            case Level::Alert:
+            case Level::Emergency:
                 ++$this->errorCount[$key];
         }
```
Seldaek: Could also be like this AFAICT:

```suggestion
        if ($record->level->isHigherThan(Level::Warning)) {
            ++$this->errorCount[$key];
        }
```

src/Symfony/Bridge/Monolog/Command/ServerLogCommand.php
lyrixx: Fixed

src/Symfony/Bridge/Monolog/Formatter/ConsoleFormatter.php
lyrixx: Fixed

src/Symfony/Bridge/Monolog/Command/ServerLogCommand.php
```diff
@@ -153,6 +155,18 @@ private function displayLog(OutputInterface $output, int $clientId, array $recor
...
         $output->write($logBlock);
 
+        $record = new LogRecord(
+            $record['datetime'],
+            $record['channel'],
+            Level::fromValue($record['level']),
+            $record['message'],
+            // We wrap context and extra, because they have been already dumped.
+            // So they are instance of Symfony\Component\VarDumper\Cloner\Data
+            // But LogRecord expects array
```
nicolas-grekas, line 164: iterator_to_array to the rescue?

nicolas-grekas: or better: `->getValue()`

composer.json
```diff
@@ -135,7 +135,7 @@
...
```
nicolas-grekas:
```suggestion
        "monolog/monolog": "^3.0",
```

src/Symfony/Bridge/Monolog/Command/ServerLogCommand.php
lyrixx: It cannot works (seen with @nicolas-grekas on slack)

TL;DR:
`$record` is now typed, with `array`. Before we used to put an instance of `Data` so it used to work. Now, we wrap it un an array, and we unwrap it in `ConsoleFormatter::format()` so `ConsoleFormatter::dumpData()` can work as expected

But I will update it a bit, to avoid passing this wild option `from_server_log`

src/Symfony/Bridge/Monolog/Formatter/ConsoleFormatter.php
```diff
@@ -195,6 +188,9 @@ private function dumpData(mixed $data, bool $colors = null): string
             $this->dumper->setColors($colors);
         }
 
+        if (($data['data'] ?? null) instanceof Data) {
+            $data = $data['data'];
+        }
         if (!$data instanceof Data) {
```
nicolas-grekas:
```suggestion
        } elseif (!$data instanceof Data) {
```

nicolas-grekas:
Thank you @lyrixx.

PR #51096, opened 2023-07-25
Use more "First class callable syntax" + Normalize set_error_handler/set_exception_handler calls

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Symfony/Bridge/Monolog/Handler/ServerLogHandler.php
+++ b/src/Symfony/Bridge/Monolog/Handler/ServerLogHandler.php
@@ -88,7 +88,7 @@ trait ServerLogHandlerTrait
             return false;
         }
 
-        set_error_handler(self::class.'::nullErrorHandler');
+        set_error_handler(static fn () => null);
 
         try {
             if (!$this->socket = $this->socket ?: $this->createSocket()) {
@@ -105,7 +105,7 @@ trait ServerLogHandlerTrait
     {
         $recordFormatted = $this->formatRecord($record);
 
-        set_error_handler(self::class.'::nullErrorHandler');
+        set_error_handler(static fn () => null);
 
         try {
             if (-1 === stream_socket_sendto($this->socket, $recordFormatted)) {
@@ -124,10 +124,6 @@ trait ServerLogHandlerTrait
     protected function getDefaultFormatter(): FormatterInterface
     {
         return new VarDumperFormatter();
-    }
-
-    private static function nullErrorHandler(): void
-    {
     }
 
     /**
--- a/src/Symfony/Bridge/Twig/Extension/HttpKernelExtension.php
+++ b/src/Symfony/Bridge/Twig/Extension/HttpKernelExtension.php
@@ -28,7 +28,7 @@ final class HttpKernelExtension extends AbstractExtension
             new TwigFunction('render', [HttpKernelRuntime::class, 'renderFragment'], ['is_safe' => ['html']]),
             new TwigFunction('render_*', [HttpKernelRuntime::class, 'renderFragmentStrategy'], ['is_safe' => ['html']]),
             new TwigFunction('fragment_uri', [HttpKernelRuntime::class, 'generateFragmentUri']),
-            new TwigFunction('controller', static::class.'::controller'),
+            new TwigFunction('controller', [self::class, 'controller']),
         ];
     }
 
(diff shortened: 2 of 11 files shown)


fancyweb (approved):

maxbeckers (approved):

GromNaN (approved):

src/Symfony/Bridge/Twig/Extension/HttpKernelExtension.php
```diff
@@ -28,7 +28,7 @@ public function getFunctions(): array
             new TwigFunction('render', [HttpKernelRuntime::class, 'renderFragment'], ['is_safe' => ['html']]),
             new TwigFunction('render_*', [HttpKernelRuntime::class, 'renderFragmentStrategy'], ['is_safe' => ['html']]),
             new TwigFunction('fragment_uri', [HttpKernelRuntime::class, 'generateFragmentUri']),
```
GromNaN, line 30: I suppose you didn't use the first class callable syntax for this helpers to avoid loading the class if it's not used?

lyrixx: Wrong guess 😅, I search only for this pattern `"self::class.'::"` and `"static::class.'::"`

lyrixx: Should I update this too?

derrabus (approved):

src/Symfony/Bridge/Twig/Extension/HttpKernelExtension.php
nicolas-grekas: nope, because that would break autoload laziness

src/Symfony/Bridge/Monolog/Handler/ServerLogHandler.php
```diff
@@ -88,7 +88,7 @@ private function doHandle(array|LogRecord $record): bool
             return false;
         }
 
-        set_error_handler(self::class.'::nullErrorHandler');
+        set_error_handler(self::nullErrorHandler(...));
```
nicolas-grekas:
```suggestion
        set_error_handler(fn () => null);
```

src/Symfony/Bridge/Twig/Extension/HttpKernelExtension.php
derrabus: No. Those methods are not static and thus those arrays are not valid callables. And that's on purpose. 😉

lyrixx: This cannot work actually, because this is not a real callable. It's a custom reference for this DIC loader

src/Symfony/Bridge/Monolog/Handler/ServerLogHandler.php
lyrixx: Fixed

src/Symfony/Bridge/Twig/Extension/HttpKernelExtension.php
```diff
@@ -28,7 +28,7 @@ public function getFunctions(): array
...
-            new TwigFunction('controller', static::class.'::controller'),
+            new TwigFunction('controller', static::controller(...)),
```
nicolas-grekas: class is final, it'd make sense to keep the style of previous lines actually (it makes twig take an easier path when optimizing the call):
```suggestion
            new TwigFunction('controller', [self::class, 'controller']),
```

lyrixx: okay 👍🏼 fixed, thanks

chalasr:
This no longer add any first-class callable syntax 😅

src/Symfony/Bridge/Monolog/Handler/ServerLogHandler.php
```diff
@@ -105,7 +105,7 @@ private function doWrite(array|LogRecord $record): void
     {
         $recordFormatted = $this->formatRecord($record);
 
-        set_error_handler(self::class.'::nullErrorHandler');
+        set_error_handler(fn () => null);
```
nicolas-grekas: maybe we have more of these in the codebase? can you have a look?

lyrixx: Done. I have update some, and left others. We have some [check](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/HttpKernel/EventListener/DebugHandlersListener.php#L89) in the code base with the array syntax. So it's not trivial to update everything. 

I updated only the simple cases

nicolas-grekas:
Thank you @lyrixx.

PR #51102, opened 2023-07-25
[Webhook] Allow slash in webhook type

### LINKED ISSUE #50973: [Webhook] WebhookController doesn't accept `/` character in the webhook path

6.3.0

### Description

Webhook routing with a `/` character in the path doesn't work. This is due to the Symfony router by default doesn't  allow a `/` character in a route parameter
```
// config/webhook.php
<?php
declare(strict_types=1);

namespace Symfony\Component\DependencyInjection\Loader\Configurator;

use Symfony\Config\FrameworkConfig;

return static function (FrameworkConfig $framework): void {
    $webhookConfig = $framework
        ->webhook();

    $webhookConfig->routing('some-service/orders') // <-- notice `/` in the path
        ->service(SomeServiceOrderWebhookRequestParser::class)
        ->secret('some-secret');
};
```

### How to reproduce

Install a new `symfony/webhook` component
Add a new webhook config with a `/` character in the route
The route won't match and return 404

### Possible Solution

Apply [the solution from the routing documentation](https://symfony.com/doc/current/routing.html#slash-characters-in-route-parameters) when configuring `WebhookController` class

### Additional Context

_No response_

alamirault:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #50973 
| License       | MIT

Allow `/` in webhook type.

This allow to have route like`/webhook/foo/bar`

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/routing/webhook.xml
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/routing/webhook.xml
@@ -6,5 +6,6 @@
 
     <route id="_webhook_controller" path="/{type}">
         <default key="_controller">webhook.controller::handle</default>
+        <requirement key="type">.+</requirement>
     </route>
 </routes>


nicolas-grekas (approved):

nicolas-grekas:
Thank you @alamirault.

PR #51104, opened 2023-07-25
[Security] Fix loading user from UserBadge

### LINKED ISSUE #50511: UserLoader Not Called

6.3

### Description

Always using the default userLoader.

AccessTokenHandler:

```
readonly class AccessTokenHandler implements AccessTokenHandlerInterface
{
    public function __construct(private UserRepository $userRepo, private JWT $jwt)
    {
    }

    public function getUserBadgeFrom(#[\SensitiveParameter] string $accessToken): UserBadge
    {
        $userId = $this->decodeJWT($accessToken)['id'];
        $user = $this->userRepo->find($userId);

        return new UserBadge($userId, fn () => $user);
    }
    ....
}
```

### How to reproduce

```
class UserRepository extends ApiServiceEntityRepository implements PasswordUpgraderInterface, UserLoaderInterface
{
    public function loadUserByIdentifier(string|int $identifier): ?User
    {
        $q = $this->createQueryBuilder('q');

        if (is_numeric($identifier)) {
            $q->where('q.phone = :identity')->setParameter('identity', (int) $identifier);
        } else {
            $q->where('q.email = :identity')->setParameter('identity', $identifier);
        }

        return $q->getQuery()->getOneOrNullResult();
    }
    ....
}

```

### Possible Solution

Before:
`
if (null === $userBadge->getUserLoader() && $this->userProvider) {
`

New Code:
https://github.com/vincentchalamon/symfony/blob/99a35f0fc32a7b5250aab5530129bae318c95209/src/Symfony/Component/Security/Http/Authenticator/AccessTokenAuthenticator.php#LL62C35-L62C35

Feature:
https://github.com/symfony/symfony/pull/48272

### Additional Context

_No response_

guillaumesmo:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #50511
| License       | MIT
| Doc PR        | none

Fixed a breaking change from https://github.com/symfony/symfony/pull/48272/files#diff-de9707bb338188f62878f2ebd42e7a7bf9547f6d0bf07a4fcd9c386c263c601b

### DIFF
--- a/src/Symfony/Component/Security/Http/AccessToken/Oidc/OidcTokenHandler.php
+++ b/src/Symfony/Component/Security/Http/AccessToken/Oidc/OidcTokenHandler.php
@@ -27,6 +27,7 @@ use Symfony\Component\Security\Core\Exception\BadCredentialsException;
 use Symfony\Component\Security\Http\AccessToken\AccessTokenHandlerInterface;
 use Symfony\Component\Security\Http\AccessToken\Oidc\Exception\InvalidSignatureException;
 use Symfony\Component\Security\Http\AccessToken\Oidc\Exception\MissingClaimException;
+use Symfony\Component\Security\Http\Authenticator\FallbackUserLoader;
 use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
 
 /**
@@ -93,7 +94,7 @@ final class OidcTokenHandler implements AccessTokenHandlerInterface
             }
 
             // UserLoader argument can be overridden by a UserProvider on AccessTokenAuthenticator::authenticate
-            return new UserBadge($claims[$this->claim], fn () => $this->createUser($claims), $claims);
+            return new UserBadge($claims[$this->claim], new FallbackUserLoader(fn () => $this->createUser($claims)), $claims);
         } catch (\Exception $e) {
             $this->logger?->error('An error occurred while decoding and validating the token.', [
                 'error' => $e->getMessage(),
--- a/src/Symfony/Component/Security/Http/Authenticator/FallbackUserLoader.php
+++ b/src/Symfony/Component/Security/Http/Authenticator/FallbackUserLoader.php
@@ -1,0 +1,32 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Security\Http\Authenticator;
+
+use Symfony\Component\Security\Core\User\UserInterface;
+
+/**
+ * This wrapper serves as a marker interface to indicate badge user loaders that should not be overridden by the
+ * default user provider.
+ *
+ * @internal
+ */
+final class FallbackUserLoader
+{
+    public function __construct(private $inner)
+    {
+    }
+
+    public function __invoke(mixed ...$args): ?UserInterface
+    {
+        return ($this->inner)(...$args);
+    }
+}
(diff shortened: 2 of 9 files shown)


guillaumesmo:
The fix is less clean than expected because the new OIDC feature relies on this bug.

The Badge user loader used to have priority over the firewall user provider. This was changed so that the OIDC Badge user loader only kicks in when no user provider is set (not sure how, however).

So the priorities are now as follows:
1. Provided user loader in the new Badge constructor
2. Firewall user provider
3. OIDC user loader

Alternative solutions:
- add a configuration to oidc to enable the default user loader
- add a property to UserBadge to indicate the loader must not override the firewall user loader, instead of the callable wrapper

nicolas-grekas:
Can you add a test case please?

@vincentchalamon can you have a look please?

src/Symfony/Component/Security/Http/AccessToken/Oidc/OidcTokenHandler.php
```diff
@@ -93,7 +94,7 @@ public function getUserBadgeFrom(string $accessToken): UserBadge
             }
 
             // UserLoader argument can be overridden by a UserProvider on AccessTokenAuthenticator::authenticate
-            return new UserBadge($claims[$this->claim], fn () => $this->createUser($claims), $claims);
+            return new UserBadge($claims[$this->claim], new FallbackUserLoader(fn () => $this->createUser($claims)), $claims);
```
vincentchalamon, line 97: Might need the same fix on [OidcUserInfoTokenHandler](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Security/Http/AccessToken/Oidc/OidcUserInfoTokenHandler.php)

src/Symfony/Component/Security/Http/Authenticator/FallbackUserLoader.php
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * This wrapper serves as a marker interface to indicate badge user loaders that should not be overridden by the
+ * default user provider.
+ *
+ * @internal
+ */
+final class FallbackUserLoader
+{
+    public function __construct(private $inner)
```
vincentchalamon, line 24:
```suggestion
    public function __construct(private callable $inner)
```

src/Symfony/Component/Security/Http/Authenticator/FallbackUserLoader.php
```diff
@@ -0,0 +1,30 @@
...
+    {
+    }
+
+    public function __invoke()
```
vincentchalamon:
```suggestion
    public function __invoke(): ?UserInterface
```

vincentchalamon (changes requested):
Nice catch! (and sorry for the bug introduced in my PR)

In addition to my review, can you also add a non-regression functional test on [AccessTokenTest](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Bundle/SecurityBundle/Tests/Functional/AccessTokenTest.php) please?

src/Symfony/Component/Security/Http/Authenticator/FallbackUserLoader.php
derrabus: That does not work because `callable` must not be used as a property type. `Closure` would work though.

src/Symfony/Component/Security/Http/Authenticator/FallbackUserLoader.php
```diff
@@ -0,0 +1,30 @@
...
+    {
+        return ($this->inner)(...\func_get_args());
+    }
```
derrabus:
```suggestion
    public function __invoke(mixed ...$args): mixed
    {
        return ($this->inner)(...$args);
    }
```

guillaumesmo:
Hi all, what can I do to move this PR forward?

nicolas-grekas:
"BC break" is not a proper description when fixing a regression. It doesn't tell which case is solved.
Can you please figure out something a bit more descriptive?

guillaumesmo:
> "BC break" is not a proper description when fixing a regression. It doesn't tell which case is solved. Can you please figure out something a bit more descriptive?

Sure, done

wouterj (approved):
Looks good to me.

fabpot:
Thank you @guillaumesmo.

PR #51119, opened 2023-07-26
[Mailer] update the Brevo SMTP host

### LINKED ISSUE #51118: Changes in SMTP relay address
### Description

We use Sendinblue as a mail provider for our Laravel application. Recently, we received an email to change the SMTP Server address:
https://developers.brevo.com/docs/changes-in-smtp-relay-address

As far as I understand we need to use the https://github.com/symfony/brevo-mailer package. However, as far as I can tell, this package is still using the old host address `smtp-relay.sendinblue.com`:
https://github.com/symfony/brevo-mailer/blob/6.4/Transport/BrevoSmtpTransport.php#L26

Do you have plans to update it? If yes, when can we expect this change to be released?

Thank you!

### Example

_No response_

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #51118
| License       | MIT
| Doc PR        | 

see https://developers.brevo.com/docs/changes-in-smtp-relay-address

### DIFF
--- a/src/Symfony/Component/Mailer/Bridge/Brevo/Transport/BrevoSmtpTransport.php
+++ b/src/Symfony/Component/Mailer/Bridge/Brevo/Transport/BrevoSmtpTransport.php
@@ -22,8 +22,7 @@ final class BrevoSmtpTransport extends EsmtpTransport
 {
     public function __construct(string $username, #[\SensitiveParameter] string $password, EventDispatcherInterface $dispatcher = null, LoggerInterface $logger = null)
     {
-        // This is not a typo: For now the smtp relay is still under sendinblue.com unlike the api.
-        parent::__construct('smtp-relay.sendinblue.com', 465, true, $dispatcher, $logger);
+        parent::__construct('smtp-relay.brevo.com', 465, true, $dispatcher, $logger);
 
         $this->setUsername($username);
         $this->setPassword($password);


xabbuh:
ping @PEtanguy

nicolas-grekas:
Thank you @xabbuh.

bastien-wink:
@xabbuh @nicolas-grekas Could this be also part of lower versions of symfony ? According to the email bellow the current SMTP will be terminated this month.

image

I will get in touch with Brevo to know exactly how long they will support the old hostname.

fabpot:
My answer would be that if a commercial product doesn’t care about making their user lives easier, I don’t see why we should. Anyway, I suppose we would accept a PR doing the needed changes in 6.3.

bastien-wink:
Indeed they don't care about their customers, I've done my best to have a clear answer, all we have is "that might impact the good fonction of your transactional sending".

I created a PR for 6.3 https://github.com/symfony/symfony/pull/51295

For the record :
image

stof:
Given that the new Brevo bridge won't be released until November (as it was merged in 6.4 and not in 6.3), I suggest merging that as a bugfix (and maybe even in the 5.4 LTS instead of having a non-working bridge as LTS)

fabpot:
I've merged the PR in 5.4.

stof:
@fabpot given that the previous host will stop working on August 30th, it would be great also to do a patch release (for 5.4 and 6.3) soon enough to give projects a bit of time to update this bridge before the deadline is reached so that it keeps working.

kgonella:
Hello, do we have a chance that a patch release will be available before August 30th ?

fabpot:
We make releases at the end of each month. So, the next release cycle will happen at the end of August.
But then again, I don't see why we should rush things out on our side because of a commercial company that doesn't care about their customers.

derrabus:
Until we tag a new release, use a dev snapshot:

```JSON
"require": {
    "symfony/sendinblue-mailer": "~6.3.4@dev"
}
```

or switch to the new package:

```JSON
"require": {
    "symfony/brevo-mailer": "^6.4@dev"
}
```

stof:
@derrabus switching to brevo-mailer also requires switching to framework-bundle 6.4.x-dev, otherwise the wiring won't be done.

PR #51133, opened 2023-07-27
[Cache] Fix compat with ext-redis v6

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

WIP, looking to fix the failures on appveyor.

Redis::scan() deprecations need https://github.com/phpredis/phpredis/pull/2363

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -18,8 +18,8 @@ install:
     - cd ext
     - appveyor DownloadFile https://github.com/symfony/binary-utils/releases/download/v0.1/php_apcu-5.1.22-8.2-ts-vs16-x86.zip
     - 7z x php_apcu-5.1.22-8.2-ts-vs16-x86.zip -y >nul
-    - appveyor DownloadFile https://github.com/symfony/binary-utils/releases/download/v0.1/php_redis-5.3.7-8.2-ts-vs16-x86.zip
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
There is still work to do to support ext-redis 6. Many failures on appveyor are related to this incomplete support. #51136 fixes the CI at least.
Closing as I won't work on this in the short term. Anyone, please take over if you're up to :) :pray:

faizanakram99:
We stumbled upon this today, the server is set to auto-update minor versions and the newest version has redis 6 installed and deployments have been failing since, couldn't figure it out unless i checked the prod logs to see signature mismatch between the proxy class and the actual \Redis class.

Can this be specified in composer.json somehow, it is causing sf to error out with redis 6 (with lazy services which isn't uncommon to avoid creating connection with Redis server)

ruudk:
@nicolas-grekas Just came across this PR after opening this issue: 
- https://github.com/symfony/symfony/issues/51770

Would it be possible to give this another shot?

There might be more people experiencing this as Redis 6 is now the default when you do `pecl install redis`

PR #51138, opened 2023-07-28
[Scheduler] Postpone schedule creation

### LINKED ISSUE #51112: [Scheduler] Trying to access the DB in a ScheduleProvider breaks migrations / tests / ...

6.3

### Description

If your ScheduleProvider tries to access the database, you can't run tests that recreate the DB or run your migrations from scratch anymore, as the getSchedule() gets called first and fails as it tries to get its configuration from the DB.

### How to reproduce

Create a ScheduleProvider that does something similar to this:
```php

    {
        $frequency = $this->parameterRepository->findOneBy([
            'key' => Parameter::REMINDER_FREQUENCY,
        ]);

        $frequency = $frequency->getValue();

        return (new Schedule())->add(
            RecurringMessage::every(frequency: $frequency,
                                    message  : new SomeMessage(),
                                    from     : $this->now())
        );
    }

```

Then just try to run  `bin/console doctrine:migrations:migrate` on an empty DB. You'll get `SQLSTATE[42P01]: Undefined table: 7 ERROR:  relation "parameter" does not exist` as it tries to run getSchedule() first. Same things with my behat tests that recreate the DB, everything just fails.

### Possible Solution

No clue right now. I may have missed something, as this is still pretty new?

### Additional Context

My current transport for messenger is doctrine://default, that might be linked?

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #51112
| License       | MIT
| Doc PR        | -

The issue is caused by the `MessengerTransportDoctrineSchemaListener` which is triggered when `doctrine:schema:create` is run. It [instantiates all transports](https://github.com/symfony/symfony/blob/4c1d9533355d66380e3f0fed774ebf38c601aab1/src/Symfony/Bridge/Doctrine/SchemaListener/MessengerTransportDoctrineSchemaListener.php#L37) which causes the `ScheduleProviderInterface::getSchedule()` method to be called, but since the schema is not yet created, everything breaks.

This PR changes the `MessageGenerator` class to accept the `ScheduleProviderInterface` instead of only `Schedule`. That way the call to `getSchedule()` can be postponed to when it's really needed, and since the `Shedule` class implements the `ScheduleProviderInterface`, there's no breaking change.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -133,7 +133,7 @@
         "doctrine/data-fixtures": "^1.1",
         "doctrine/dbal": "^2.13.1|^3.0",
         "doctrine/orm": "^2.12",
-        "dragonmantank/cron-expression": "^3",
+        "dragonmantank/cron-expression": "^3.1",
         "egulias/email-validator": "^2.1.10|^3.1|^4",
         "guzzlehttp/promises": "^1.4",
         "league/html-to-markdown": "^5.0",
--- a/src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
+++ b/src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
@@ -15,37 +15,39 @@ use Psr\Clock\ClockInterface;
 use Symfony\Component\Clock\Clock;
 use Symfony\Component\Scheduler\RecurringMessage;
 use Symfony\Component\Scheduler\Schedule;
+use Symfony\Component\Scheduler\ScheduleProviderInterface;
 
 /**
  * @experimental
  */
 final class MessageGenerator implements MessageGeneratorInterface
 {
+    private Schedule $schedule;
     private TriggerHeap $triggerHeap;
     private ?\DateTimeImmutable $waitUntil;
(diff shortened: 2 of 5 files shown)


composer.json
```diff
@@ -133,7 +133,7 @@
         "doctrine/data-fixtures": "^1.1",
         "doctrine/dbal": "^2.13.1|^3.0",
         "doctrine/orm": "^2.12",
-        "dragonmantank/cron-expression": "^3",
+        "dragonmantank/cron-expression": "^3.1",
```
HypeMC, line 136: Sync up with `src/Symfony/Component/Scheduler/composer.json`, lowest deps fail otherwise.

kbond (approved):
Thanks @HypeMC, this makes sense to me.

fabpot (approved):

fabpot:
Thank you @HypeMC.

PR #45476, opened 2022-02-18
[HttpKernel] Deprecate StreamedResponseListener, it serves no purpose anymore

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Tickets       | -
| License       | MIT
| Doc PR        | -

`StreamedResponseListener` has been introduced at the same time as `StreamedResponse` in #2935.

Its purpose was to make catching exceptions easier by wrapping the call to `$response->send()` in the main try/catch of `HttpKernel`.

Since #12998, we have `HttpKernel::terminateWithException()`, and we don't need that anymore, so we can just remove the listener.

This will help [integrate Symfony into e.g. Swoole](https://github.com/php-runtime/runtime/issues/115) /cc @alexander-schranz.

PR #51139, opened 2023-07-28
[HttpKernel] Fix missing `Request` in `RequestStack` for `StreamedResponse`

### LINKED ISSUE #46743: [HttpKernel] "Deprecation" of StreamedResponseListener causes RequestStack to be empty within StreamedResponse callback

6.1.*

### Description

The change made within https://github.com/symfony/symfony/pull/45476 causes the `RequestStack` to be empty when the `StreamedResponse` Callback is being called.

In detail:
When using an `StreamedResponse`, the Callback passed to `StreamedResponse` is executed AFTER the `\Symfony\Component\HttpKernel\Kernel::handle` method (which is doing a `pop` on the `RequestStack` in `\Symfony\Component\HttpKernel\HttpKernel::finishRequest`).

Instead it is executed within the `\Symfony\Component\HttpFoundation\Response::send` method causing the `\Symfony\Component\HttpFoundation\RequestStack` to be empty and not having any `Request` object inside it.

We're upgrading our legacy application which is already wrapped with symfony.
This is why we're sadly have to use a construct where we're required to inject the `\Symfony\Component\HttpFoundation\RequestStack` to get the `Request` which is not available anymore at that state within a `StreamedResponse`.

Is this really the expected / wanted behaviour for this change? That when using a `StreamedResponse` the `Request` object becomes unavailable or is required to be reinitialized manually? If this is desired it should be announced as a possible BC break.

### How to reproduce

`./src/Service/TestService.php`
```php
<?php

namespace App\Service;

use Symfony\Component\DependencyInjection\Attribute\Autoconfigure;

#[Autoconfigure(public: true)]
class TestService
{
    public function __construct(private readonly RequestStack $requestStack)
    {
    }

    public function getRequest(): Request
    {
        return $this->requestStack->getMainRequest();
    }
}

```

`./src/Controller/TestController.php`
```php
<?php

namespace App\Controller;

use App\Kernel;
use App\Service\TestService;
use Symfony\Component\HttpFoundation\StreamedResponse;
use Symfony\Component\HttpKernel\Attribute\AsController;

#[AsController]
{
    #[Route('/')]
    public function test(Kernel $kernel): StreamedResponse
    {
        return new StreamedResponse(function () use ($kernel) {
            /** @var TestService $testService */
            $testService = $kernel->getContainer()->get(TestService::class);

            $mainRequest = $testService->getRequest();

            echo $mainRequest->server->get('HTTP_HOST');
        });
    }
}

```

When calling the `/test` url an exception is thrown:
```
Return value must be of type Symfony\Component\HttpFoundation\Request, null returned
```

### Possible Solution

_No response_

### Additional Context

Behaviour was changed in pull request: https://github.com/symfony/symfony/pull/45476

DaDeather:
| Q             | A
| ------------- | ---
| Branch       | 6.3
| Bug fix      | yes
| New feature  | no
| Deprecations | no
| Tickets       | Fix #46743
| License       | MIT

Fix removing the `\Symfony\Component\HttpFoundation\Request` from `\Symfony\Component\HttpFoundation\RequestStack` when using `\Symfony\Component\HttpFoundation\StreamedResponse`.

### DIFF
--- a/src/Symfony/Component/HttpFoundation/StreamedResponse.php
+++ b/src/Symfony/Component/HttpFoundation/StreamedResponse.php
@@ -57,6 +57,14 @@ class StreamedResponse extends Response
     }
 
     /**
+     * Decorates the PHP callback associated with this Response.
+     */
+    public function getCallback(): callable
+    {
+        return $this->callback;
+    }
+
+    /**
      * This method only sends the headers once.
      *
      * @param null|positive-int $statusCode The status code to use, override the statusCode property if set and not null
--- a/src/Symfony/Component/HttpKernel/HttpKernel.php
+++ b/src/Symfony/Component/HttpKernel/HttpKernel.php
@@ -15,6 +15,7 @@ use Symfony\Component\HttpFoundation\Exception\RequestExceptionInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\StreamedResponse;
 use Symfony\Component\HttpKernel\Controller\ArgumentResolver;
 use Symfony\Component\HttpKernel\Controller\ArgumentResolverInterface;
 use Symfony\Component\HttpKernel\Controller\ControllerResolverInterface;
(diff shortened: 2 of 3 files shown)


DaDeather:
Any news on this happening? Or are any adjustments necessary?

derrabus:
Sorry that nobody has picked up your PR yet.

What I dislike about the proposed change is that the kernel is made aware of a specific response class and adds edge-case behavior to the request stack based on that.

From my PoV, it's totally reasonable that the request stack is empty at this point. If we've caused trouble for projects upgrading from Symfony 6.1, we need to find a solution, but I don't think we should pull this kind of logic into Symfony 7. 😕 

@nicolas-grekas Can you take a look please?

nicolas-grekas:
Thanks for the PR and the details in the issue. I think we completely missed this consequence. It'd be fine reverting #45476, and adding also a test case that covers this situation so that we won't forget about the use case again.

@DaDeather up to submit a PR doing so?

I'm closing this one meanwhile as I agree with @derrabus, that's not the approach we should merge.

DaDeather:
Here we go: https://github.com/symfony/symfony/pull/51391

nicolas-grekas:
That's the way forward, see https://github.com/symfony/symfony/pull/51391#issuecomment-1680094802

PR #51142, opened 2023-07-28
[Mailer] Precise AbstractApiTransport::getRecipients signature

VincentLanglet:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #... 
| License       | MIT

Not sure about the branch I should target for this PHPdoc.
I started with 6.4 cause I assume the PR is not finished...

`AbstractApiTransport::getRecipients` is filtering the `Address[]` so return an array of Address too,
adding the `@return` phpdoc will help with autocompletion from PHPStorm and for static analysis.

Currently I have a some issue with adding the phpdoc, 
`MandrillApiTransport` is overriding the getRecipients without keeping the same structure ; 
I don't think it's a good practice since we can't be sure about the array elements then...

Maybe `MandrillApiTransport` should have his own `getPayloadRecipients` instead of overriding `getRecipients` ?

### DIFF
--- a/src/Symfony/Component/Mailer/Transport/AbstractApiTransport.php
+++ b/src/Symfony/Component/Mailer/Transport/AbstractApiTransport.php
@@ -37,6 +37,9 @@ abstract class AbstractApiTransport extends AbstractHttpTransport
         return $this->doSendApi($message, $email, $message->getEnvelope());
     }
 
+    /**
+     * @return Address[]
+     */
     protected function getRecipients(Email $email, Envelope $envelope): array
     {
         return array_filter($envelope->getRecipients(), fn (Address $address) => false === \in_array($address, array_merge($email->getCc(), $email->getBcc()), true));


src/Symfony/Component/Mailer/Transport/AbstractApiTransport.php
```diff
@@ -37,6 +37,9 @@ protected function doSendHttp(SentMessage $message): ResponseInterface
         return $this->doSendApi($message, $email, $message->getEnvelope());
     }
 
+    /**
+     * @return array<Address>
```
fabpot: Let's keep it simple and consistent with the phpdoc in Envelope. 

```suggestion
     * @return Address[]
```

fabpot (changes requested):

fabpot (approved):

fabpot:
Good catch, thanks @VincentLanglet.

VincentLanglet:
> Good catch, thanks @VincentLanglet.

You're welcome.

As explained in the first message @fabpot, the MandrillApiTransport::getRecipient override the original `getRecipient` without respecting the original signature ; I think it's not a good practice. Should something be done about it ?
- Creating an issue ?
- Creating a `getPayloadRecipient` method in `MandrillApiTransport` instead of overriding `getRecipient` ?

Cf: https://github.com/symfony/symfony/pull/51142

fabpot:
The second option sounds the easiest one.

stof:
The second option is what we did in the Brevo transport

PR #32267, opened 2019-06-28
[FrameworkBundle] deprecate the framework.templating option

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 4.4
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | #32120
| License       | MIT
| Doc PR        | 

The config node has already been removed in the `master` branch in #31800. For DX it would have been better to have this deprecation in 4.3 (see e.g. #32120), but it's probably too late to ship this as a bugfix.

PR #51144, opened 2023-07-28
[Templating] deprecate the component

kbond:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

I think it's time to deprecate/remove this component.

### DIFF
--- a/src/Symfony/Component/Templating/CHANGELOG.md
+++ b/src/Symfony/Component/Templating/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+* Deprecate the component, use Twig instead
+
 2.5.0
 -----
 
--- a/src/Symfony/Component/Templating/DelegatingEngine.php
+++ b/src/Symfony/Component/Templating/DelegatingEngine.php
@@ -11,10 +11,14 @@
 
 namespace Symfony\Component\Templating;
 
+trigger_deprecation('symfony/templating', '6.4', '"%s" is deprecated since version 6.4 and will be removed in 7.0. Use Twig instead.', DelegatingEngine::class);
+
 /**
  * DelegatingEngine selects an engine for a given template.
  *
  * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @deprecated since Symfony 6.4, use Twig instead
  */
 class DelegatingEngine implements EngineInterface, StreamingEngineInterface
 {
--- a/src/Symfony/Component/Templating/EngineInterface.php
+++ b/src/Symfony/Component/Templating/EngineInterface.php
@@ -11,6 +11,8 @@
 
 namespace Symfony\Component\Templating;
 
+trigger_deprecation('symfony/templating', '6.4', '"%s" is deprecated since version 6.4 and will be removed in 7.0. Use Twig instead.', EngineInterface::class);
+
 /**
  * EngineInterface is the interface each engine must implement.
  *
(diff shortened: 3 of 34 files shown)


kbond:
I'm not sure how to suppress the deprecations. I've added `@group legacy` to the test classes but I guess the deprecations are triggered earlier?

GromNaN:
The component is [still downloaded 30K/day](https://packagist.org/packages/symfony/templating/stats). Not on legacy projects since there is 6.4K daily installs for the v6.
image

kbond:
I think a good chunk of the downloads are transient. For instance, [knplabs/knp-time-bundle](https://packagist.org/packages/knplabs/knp-time-bundle) requires it and has ~4k daily downloads.

src/Symfony/Component/Templating/DelegatingEngine.php
```diff
@@ -11,6 +11,8 @@
 
 namespace Symfony\Component\Templating;
 
+trigger_deprecation('symfony/templating', '6.4', '"%s" is deprecated since version 6.4 and will be removed in 7.0. Use Twig instead.', DelegatingEngine::class);
```
derrabus, line 14: Please also flag all classes as `@deprecated` so IDEs and static analyzers may pick up the deprecation.

src/Symfony/Component/Templating/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4.0
+-----
+
+* deprecated the component: use Twig instead
```
derrabus:
```suggestion
6.4
---

 * Deprecate the component, use Twig instead
```

Our code-style has changed since the last time we did change something in that component. 🙈

nicolas-grekas:
:+1: for the proposal. Deprecating it will still mean maintaining it for the duration of the LTS, so 3 more years.

src/Symfony/Component/Templating/EngineInterface.php
```diff
@@ -11,6 +11,8 @@
 
 namespace Symfony\Component\Templating;
 
+trigger_deprecation('symfony/templating', '6.4', '"%s" is deprecated since version 6.4 and will be removed in 7.0. Use Twig instead.', EngineInterface::class);
```
nicolas-grekas, line 14: missing dep in composer.json to have this function available

nicolas-grekas:
Any corresponding config tree in frameworkbundle that should also be deprecated?

derrabus:
> Any corresponding config tree in frameworkbundle that should also be deprecated?

Didn't this happen in #32267 already?

src/Symfony/Component/Templating/DelegatingEngine.php
```diff
@@ -17,6 +17,8 @@
  * DelegatingEngine selects an engine for a given template.
  *
  * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @deprecated since Symfony 6.5, use Twig instead
```
derrabus:
```suggestion
 * @deprecated since Symfony 6.4, use Twig instead
```

We're not planning to release a 6.5. 😉

kbond: oops, fixed.

fabpot (approved):

GromNaN (approved):

derrabus (approved):

fabpot:
Thank you @kbond.

PR #51150, opened 2023-07-28
Use Stringable interface as much as possible

VincentLanglet:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #... 
| License       | MIT

Instead of requiring `__toString()` in interfaces, wouldn't it be better to extend `\Stringable` ?
(I kept the method when comment was added)

### DIFF
--- a/src/Symfony/Component/Config/Resource/ResourceInterface.php
+++ b/src/Symfony/Component/Config/Resource/ResourceInterface.php
@@ -16,7 +16,7 @@ namespace Symfony\Component\Config\Resource;
  *
  * @author Fabien Potencier <fabien@symfony.com>
  */
-interface ResourceInterface
+interface ResourceInterface extends \Stringable
 {
     /**
      * Returns a string representation of the Resource.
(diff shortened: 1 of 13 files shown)


nicolas-grekas (approved):
It's strictly equivalent, but why not :)

derrabus (approved):

nicolas-grekas:
Missing : FormErrorIterator, HeaderBag, InMemoryUser, AbstractUid and Data.

VincentLanglet:
> It's strictly equivalent, but why not :)

Sure, but according to
https://www.php.net/manual/en/class.stringable.php
Unlike most interfaces, Stringable is implicitly present on any class that has the magic [__toString()](https://www.php.net/manual/en/language.oop5.magic.php#object.tostring) method defined, __although it can and should be declared explicitly.__

> Missing : FormErrorIterator, HeaderBag, InMemoryUser, AbstractUid and Data.

I focused on interface but added these class too then.

OskarStark (approved):

fabpot (approved):

fabpot:
Thank you @VincentLanglet.

PR #51151, opened 2023-07-28
[5.4] Fix symfony/deprecation-contracts require

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Require `symfony/deprecation-contracts` in package that use `trigger_deprecation` without requiring the dependency :see_no_evil: .

### DIFF
--- a/src/Symfony/Bundle/TwigBundle/composer.json
+++ b/src/Symfony/Bundle/TwigBundle/composer.json
@@ -18,6 +18,7 @@
     "require": {
         "php": ">=7.2.5",
         "symfony/config": "^4.4|^5.0|^6.0",
+        "symfony/deprecation-contracts": "^2.1|^3",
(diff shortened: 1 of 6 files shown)


fancyweb:
All branches are impacted so I'll fix them one by one after each one is ok and merged up?

derrabus:
I think we should drop this change when merging up. All deprecations have been resolved in 6.0, so we need to re-evaluate this topic on 6.2 anyway.

fancyweb:
I agree, I already prepared all PRs and it's easier to just drop the changes when merging up since each maintained branch is a little different :sweat_smile:

nicolas-grekas:
Thank you @fancyweb.

PR #51168, opened 2023-07-30
[Messenger] Add forward compatibily with php-amqp v2

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

see https://github.com/php-amqp/php-amqp/issues/424#issuecomment-1656692661

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Amqp/Transport/Connection.php
+++ b/src/Symfony/Component/Messenger/Bridge/Amqp/Transport/Connection.php
@@ -456,7 +456,7 @@ class Connection
 
     public function ack(\AMQPEnvelope $message, string $queueName): bool
     {
-        return $this->queue($queueName)->ack($message->getDeliveryTag());
+        return $this->queue($queueName)->ack($message->getDeliveryTag()) ?? true;
     }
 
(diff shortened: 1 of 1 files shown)


nicolas-grekas (approved):
:monkey:

maxbeckers (approved):

fabpot:
I'm not sure this qualifies as a bug fix.
In any case, I suppose we would need to run tests for both v1 and v2?

fabpot:
Thank you @lyrixx.

Danielss89:
This actually need to be done to the nack() method too, not only ack().
Is there a reason why this is only in 5.x branch and not 6.x branch? If not, i can do a PR for 6.x
It currently breaks with version 2 of php-amqp, so would be good to get patched rather quick.

Sorry, i see it actually is in 6.4!

maxbeckers:
@Danielss89 i guess the nack() was forgotten in this fix, you can provide a pr for that. And this will be merged to all the newer versions at the latest with the next release.

Danielss89:
Ok great! Which branch should i target?

stof:
@Danielss89 please target 5.4, as done for this PR.

PR #51173, opened 2023-07-31
Add some PHPDoc

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Helps IDEs with autocompletion/code navigation.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/ProfilerPass.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/ProfilerPass.php
@@ -42,8 +42,7 @@ class ProfilerPass implements CompilerPassInterface
             $template = null;
 
             $collectorClass = $container->findDefinition($id)->getClass();
-            $isTemplateAware = is_subclass_of($collectorClass, TemplateAwareDataCollectorInterface::class);
-            if (isset($attributes[0]['template']) || $isTemplateAware) {
+            if (isset($attributes[0]['template']) || is_subclass_of($collectorClass, TemplateAwareDataCollectorInterface::class)) {
                 $idForTemplate = $attributes[0]['id'] ?? $collectorClass;
                 if (!$idForTemplate) {
                     throw new InvalidArgumentException(sprintf('Data collector service "%s" must have an id attribute in order to specify a template.', $id));
--- a/src/Symfony/Component/Form/Extension/DependencyInjection/DependencyInjectionExtension.php
+++ b/src/Symfony/Component/Form/Extension/DependencyInjection/DependencyInjectionExtension.php
@@ -14,6 +14,7 @@ namespace Symfony\Component\Form\Extension\DependencyInjection;
 use Psr\Container\ContainerInterface;
 use Symfony\Component\Form\Exception\InvalidArgumentException;
 use Symfony\Component\Form\FormExtensionInterface;
+use Symfony\Component\Form\FormTypeExtensionInterface;
 use Symfony\Component\Form\FormTypeGuesserChain;
 use Symfony\Component\Form\FormTypeGuesserInterface;
 use Symfony\Component\Form\FormTypeInterface;
@@ -27,7 +28,7 @@ class DependencyInjectionExtension implements FormExtensionInterface
     private iterable $guesserServices;
 
     /**
-     * @param iterable[] $typeExtensionServices
+     * @param array<string, iterable<FormTypeExtensionInterface>> $typeExtensionServices
      */
     public function __construct(ContainerInterface $typeContainer, array $typeExtensionServices, iterable $guesserServices)
     {
--- a/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/UidValueResolver.php
+++ b/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/UidValueResolver.php
@@ -38,12 +38,11 @@ final class UidValueResolver implements ArgumentValueResolverInterface, ValueRes
         if ($argument->isVariadic()
             || !\is_string($value = $request->attributes->get($argument->getName()))
             || null === ($uidClass = $argument->getType())
-            || !is_subclass_of($argument->getType(), AbstractUid::class, true)
+            || !is_subclass_of($uidClass, AbstractUid::class, true)
         ) {
             return [];
         }
 
-        /* @var class-string<AbstractUid> $uidClass */
         try {
             return [$uidClass::fromString($value)];
         } catch (\InvalidArgumentException $e) {
--- a/src/Symfony/Component/Uid/Factory/TimeBasedUuidFactory.php
+++ b/src/Symfony/Component/Uid/Factory/TimeBasedUuidFactory.php
@@ -13,9 +13,15 @@ namespace Symfony\Component\Uid\Factory;
 
 use Symfony\Component\Uid\TimeBasedUidInterface;
 use Symfony\Component\Uid\Uuid;
+use Symfony\Component\Uid\UuidV1;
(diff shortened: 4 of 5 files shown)


src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/ProfilerPass.php
```diff
@@ -41,6 +41,7 @@ public function process(ContainerBuilder $container)
             $priority = $attributes[0]['priority'] ?? 0;
             $template = null;
 
+            /** @var class-string|class-string<TemplateAwareDataCollectorInterface> $collectorClass */
```
derrabus: This annotation does not make sense. It's basically the same as…

```suggestion
            /** @var class-string $collectorClass */
```

In addition to this, I don't really understand which problem all those inline doc blocks solve. Please elaborate.

src/Symfony/Component/Uid/Factory/TimeBasedUuidFactory.php
```diff
@@ -27,6 +30,7 @@ public function __construct(string $class, Uuid $node = null)
 
     public function create(\DateTimeInterface $time = null): Uuid&TimeBasedUidInterface
     {
+        /** @var class-string<UuidV1|UuidV6|UuidV7> $class */
```
derrabus: Shouldn't we annotate this type to the `$class` property instead?

src/Symfony/Component/Form/Extension/DependencyInjection/DependencyInjectionExtension.php
```diff
@@ -27,7 +28,7 @@ class DependencyInjectionExtension implements FormExtensionInterface
     private iterable $guesserServices;
 
     /**
-     * @param iterable[] $typeExtensionServices
+     * @param array<string, iterable<int, FormTypeExtensionInterface>> $typeExtensionServices
```
derrabus:
```suggestion
     * @param array<string, iterable<FormTypeExtensionInterface>> $typeExtensionServices
```

The class does not really care how the inner iterable is keyed.

derrabus (changes requested):

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/ProfilerPass.php
wouterj: If anything, we should add `@return class-string` to `Definition::getClass()` and revert all inline PHPdocs imho

src/Symfony/Component/Console/DependencyInjection/AddConsoleCommandPass.php
```diff
@@ -42,6 +42,7 @@ public function process(ContainerBuilder $container)
         foreach ($commandServices as $id => $tags) {
             $definition = $container->getDefinition($id);
             $definition->addTag('container.no_preload');
+            /** @var class-string<Command> $class */
```
wouterj: This is a "lie", which is [not allowed per our phpdoc standards](https://symfony.com/doc/current/contributing/code/standards.html#documentation).

Nothing in the contract guarantees that `$class` is a class string of `Command`. Adding this PHPdoc means we'll miss the edge-cases where this is not the case.

The same thing applies to the other inline PHPdocs of this PR.

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/UidValueResolver.php
```diff
@@ -35,6 +35,7 @@ public function supports(Request $request, ArgumentMetadata $argument): bool
 
     public function resolve(Request $request, ArgumentMetadata $argument): array
     {
+        /** @var class-string<AbstractUid> $uidClass */
```
wouterj: If we change the if statement like this, static analysis should be able to derive this without inline PHPdoc:

```diff
         if ($argument->isVariadic()
              || !\is_string($value = $request->attributes->get($argument->getName()))
              || null === ($uidClass = $argument->getType())
-             || !is_subclass_of($argument->getType(), AbstractUid::class, true)
+             || !is_subclass_of($uidClass, AbstractUid::class, true)
          ) {
```

wouterj (changes requested):

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/ProfilerPass.php
HypeMC: @derrabus Yeah, the reason was that my IDE had trouble understanding that `$collectorClass` could be a subclass of `TemplateAwareDataCollectorInterface`, but that's really an IDE issue I guess. In any case, inlining the `is_subclass_of` part seems to fix the issue, so hopefully that's ok.
@wouterj Done

src/Symfony/Component/Uid/Factory/TimeBasedUuidFactory.php
HypeMC: Done

src/Symfony/Component/Form/Extension/DependencyInjection/DependencyInjectionExtension.php
HypeMC: Done

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/UidValueResolver.php
HypeMC: Yes, you're right, done

src/Symfony/Component/Console/DependencyInjection/AddConsoleCommandPass.php
HypeMC: I've removed the inline PHPdocs.

wouterj (approved):
Please fix the CS error reported by Fabbot, but other than that this looks good to me now :+1:

src/Symfony/Component/Form/DependencyInjection/FormPass.php
```diff
@@ -19,6 +19,7 @@
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
 use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\Form\FormTypeExtensionInterface;
```
Seb33300: Mistake?

HypeMC: Yep, removed

HypeMC:
> Please fix the CS error reported by Fabbot, but other than that this looks good to me now +1

@wouterj Done

derrabus (approved):

nicolas-grekas:
Thank you @HypeMC.

src/Symfony/Component/Uid/Factory/TimeBasedUuidFactory.php
```diff
@@ -13,9 +13,15 @@
...
 use Symfony\Component\Uid\TimeBasedUidInterface;
 use Symfony\Component\Uid\Uuid;
+use Symfony\Component\Uid\UuidV1;
+use Symfony\Component\Uid\UuidV6;
+use Symfony\Component\Uid\UuidV7;
 
 class TimeBasedUuidFactory
 {
+    /**
+     * @var class-string<UuidV1|UuidV6|UuidV7>
```
stof, line 23: this should actually be added on the constructor parameter as well. Otherwise it won't help IDEs and static analysis tools to highlight bad usages.

PR #51191, opened 2023-07-31
[Mime] Update mimetypes

fabpot:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a


src/Symfony/Component/Mime/MimeTypes.php
```diff
@@ -1090,7 +1099,7 @@ public function guessMimeType(string $path): ?string
         'application/x-sh' => ['sh'],
         'application/x-shar' => ['shar'],
         'application/x-shared-library-la' => ['la'],
-        'application/x-sharedlib' => ['so'],
+        'application/x-sharedlib' => ['so', 'so.[0-9]*'],
```
stof: I don't think the class supports patterns in the way it uses this array.

stof (changes requested):

src/Symfony/Component/Mime/MimeTypes.php
fabpot: Done

stof: This should be solved by excluding extensions containing a range pattern (can probably be detected based on the presence of `[`) in the update script, not by manual cleanup. Otherwise, we will have the same issue in the next update (and maybe I won't be the one reviewing it to notice it)

fabpot: > This should be solved by excluding extensions containing a range pattern (can probably be detected based on the presence of `[`) in the update script, not by manual cleanup. Otherwise, we will have the same issue in the next update (and maybe I won't be the one reviewing it to notice it)

As we update this file every two years, let's keep it simple.

stof: to me, having to remember every 2 years that the script requires a manual post-processing is more risky. The check should be easy to add.

stof: The automatic check is probably just adding a check for the presence of `[` in the pattern in https://github.com/symfony/symfony/blob/139419d2b540fdee671b4012dc4fe318f314721f/src/Symfony/Component/Mime/Resources/bin/update_mime_types.php#L31

src/Symfony/Component/Mime/MimeTypes.php
```diff
@@ -1769,8 +1793,9 @@ public function guessMimeType(string $path): ?string
         'video/vnd.rn-realvideo' => ['rv', 'rvx'],
         'video/vnd.uvvu.mp4' => ['uvu', 'uvvu'],
         'video/vnd.vivo' => ['viv', 'vivo'],
+        'video/vnd.youtube.yt' => ['yt'],
         'video/webm' => ['webm'],
-        'video/x-anim' => ['anim[1-9j]'],
+        'video/x-anim' => ['animj', 'animjj', 'animjjj', 'animjjjj', 'animjjjjj', 'animjjjjjj', 'animjjjjjjj', 'animjjjjjjjjj'],
```
Seb33300: Not sure if you tried to manually translate the regex here.
But `[1-9j]` means 1 figure or 1 letter j (not j repeated 1 to 9 times)

fabpot: :joy:

src/Symfony/Component/Mime/MimeTypes.php
```diff
@@ -1769,8 +1793,9 @@ public function guessMimeType(string $path): ?string
         'video/vnd.rn-realvideo' => ['rv', 'rvx'],
         'video/vnd.uvvu.mp4' => ['uvu', 'uvvu'],
         'video/vnd.vivo' => ['viv', 'vivo'],
+        'video/vnd.youtube.yt' => ['yt'],
         'video/webm' => ['webm'],
-        'video/x-anim' => ['anim[1-9j]'],
+        'video/x-anim' => ['anim1j', 'anim2j', 'anim3j', 'anim4j', 'anim5j', 'anim6j', 'anim7j', 'anim8j', 'anim9j'],
```
nicolas-grekas:
```suggestion
        'video/x-anim' => ['anim1', 'anim2', 'anim3', 'anim4', 'anim5', 'anim6', 'anim7', 'anim8', 'anim9', 'animj'],
```

src/Symfony/Component/Mime/MimeTypes.php
```diff
@@ -1879,7 +1907,15 @@ public function guessMimeType(string $path): ?string
...
-        'anim[1-9j]' => ['video/x-anim'],
+        'anim1j' => ['video/x-anim'],
+        'anim2j' => ['video/x-anim'],
+        'anim3j' => ['video/x-anim'],
+        'anim4j' => ['video/x-anim'],
+        'anim5j' => ['video/x-anim'],
+        'anim6j' => ['video/x-anim'],
+        'anim7j' => ['video/x-anim'],
+        'anim8j' => ['video/x-anim'],
+        'anim9j' => ['video/x-anim'],
```
nicolas-grekas:
```suggestion
        'anim1' => ['video/x-anim'],
        'anim2' => ['video/x-anim'],
        'anim3' => ['video/x-anim'],
        'anim4' => ['video/x-anim'],
        'anim5' => ['video/x-anim'],
        'anim6' => ['video/x-anim'],
        'anim7' => ['video/x-anim'],
        'anim8' => ['video/x-anim'],
        'anim9' => ['video/x-anim'],
        'animj' => ['video/x-anim'],
```

src/Symfony/Component/Mime/MimeTypes.php
fabpot: But what about `anim[1-9j]` also in this file that we can handle?

stof: then look for `*` to exclude repeated symbols and implement the expansion of single-symbol character classes.

fabpot: What about just doing it manually when needed then?

stof: "when needed" is "each time the script is run", given that the script will override previous manual changes.

nicolas-grekas (approved):
Let's keep things simple, we review such changes anyway.

PR #51197, opened 2023-07-31
[PsrHttpMessageBridge] Support `php-http/discovery` for auto-detecting PSR-17 factories

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | TODO

This PR follows the example of `Symfony\Component\HttpClient\Psr18Client` and makes all dependencies of `PsrHttpFactory` optional.

If `php-http/discovery` is not installed, I'm falling back to @Nyholm's implementation if it's available. This is consistent with `Psr18Client`'s behavior. However, I wonder if this is the right course of action, given that `php-http/discovery` already handles the problem of discovering an implementation well.


nicolas-grekas (approved):

fabpot:
Thank you @derrabus.

PR #51198, opened 2023-07-31
[DependencyInjection] Fix autocasting `null` env values to empty string with `container.env_var_processors_locator`

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | https://github.com/symfony/symfony/pull/50837#discussion_r1279230875
| License       | MIT
| Doc PR        | no

The previous fix doesn't work when `$processors` comes from `container.env_var_processors_locator`.

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Container.php
+++ b/src/Symfony/Component/DependencyInjection/Container.php
@@ -391,13 +391,9 @@ class Container implements ContainerInterface, ResetInterface
             $localName = $name;
         }
 
-        if ($processors->has($prefix)) {
-            $processor = $processors->get($prefix);
-        } else {
-            $processor = new EnvVarProcessor($this);
-            if (false === $i) {
-                $prefix = '';
-            }
+        $processor = $processors->has($prefix) ? $processors->get($prefix) : new EnvVarProcessor($this);
+        if (false === $i) {
+            $prefix = '';
         }
 
         $this->resolving[$envName] = true;
--- a/src/Symfony/Component/DependencyInjection/EnvVarProcessorInterface.php
+++ b/src/Symfony/Component/DependencyInjection/EnvVarProcessorInterface.php
@@ -23,7 +23,7 @@ interface EnvVarProcessorInterface
     /**
      * Returns the value of the given variable as managed by the current instance.
      *
-     * @param string   $prefix The namespace of the variable
+     * @param string   $prefix The namespace of the variable; when the empty string is passed, null values should be kept as is
      * @param string   $name   The name of the variable within the namespace
      * @param \Closure $getEnv A closure that allows fetching more env vars
      *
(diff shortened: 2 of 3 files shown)


src/Symfony/Component/DependencyInjection/Container.php
```diff
@@ -391,13 +391,9 @@ protected function getEnv(string $name)
...
 
-        if ($processors->has($prefix)) {
-            $processor = $processors->get($prefix);
-        } else {
-            $processor = new EnvVarProcessor($this);
-            if (false === $i) {
-                $prefix = '';
-            }
+        $processor = $processors->has($prefix) ? $processors->get($prefix) : new EnvVarProcessor($this);
+        if (false === $i && EnvVarProcessor::class === \get_class($processor)) {
```
fancyweb: or instanceof I don't know?

fancyweb:
Ping @bendavies can you try this patch plz?

src/Symfony/Component/DependencyInjection/Container.php
nicolas-grekas: Why is this extra condition needed at all? I removed it and tests still pass.

nicolas-grekas (commented):
Friendly ping @bendavies can you please try this patch and report back?

bendavies:
> Friendly ping @bendavies can you please try this patch and report back?

@nicolas-grekas i'm afraid i'm afk on holiday for 3 weeks. i can check when i'm back if you can wait.

src/Symfony/Component/DependencyInjection/Container.php
fancyweb: Because we implemented the empty string behavior only in our EnvVarProcessor class.

bendavies:
yes, this seems to resolve the issue.

src/Symfony/Component/DependencyInjection/Container.php
fancyweb: Looking at this again, I think it's safer to add the condition :thinking:

nicolas-grekas: I'm not fan of hardcoding a behavior for EnvVarProcessor only. Can't we add some phpdoc in the interface to explain how a processor should behave instead?

fancyweb: We could, but my point is that we added this special behavior on 5.4 as a bug fix :sweat_smile: Isn't it too late to change the behavior for every consumer?

nicolas-grekas: I think it's too much an edge case, nobody replaced the "string" processor :)

fancyweb: I added a comment but I'm not convinced it's needed/understandable. I let you takeover :smile:

src/Symfony/Component/DependencyInjection/EnvVarProcessorInterface.php
```diff
@@ -23,7 +23,7 @@ interface EnvVarProcessorInterface
     /**
      * Returns the value of the given variable as managed by the current instance.
      *
-     * @param string   $prefix The namespace of the variable
+     * @param string   $prefix The namespace of the variable. When the prefix is '' (empty string), the returned value must be null when the resolved env value is null (it should not be cast).
```
nicolas-grekas:
```suggestion
     * @param string   $prefix The namespace of the variable; when the empty string is passed, null values should be kept as is```

nicolas-grekas:
Thank you @fancyweb.

PR #51199, opened 2023-08-01
[Console] `ChoiceQuestion` - choice to get value or key

### LINKED ISSUE #40439: SymfonyStyle ChoiceQuestion - choice to send value or key
**Description**
Currently, when creating a new Symfony command and using the choice question, like this :

```php
return $io->choice($question, $choices, $default);
```

It returns the value. But I want the key ! It could be great to give the choice to send key or value with a booleen parameter, like this : 

```php

return $io->choice($question, $choices, $default, false);
```

**Example**

```php
$choices = [
            0 => 'red',
            1 => 'blue',
            2 => 'orange'
        ];

        return $io->choice('Choose a color', $choices, 0); // Returns the value (red, blur or orange)
        return $io->choice('Choose a color', $choices, 0, false); // Returns the key (0, 1 or 2)
```

maxbeckers:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #40439
| License       | MIT
| Doc PR        | symfony/symfony-docs#... TBD

Give the developer the Choice to get either the value(s) of a ChoiceQuestion (default behavior) or the key(s) of the selected values.

### Example
```php
$choices = ['value_0','value_1', 'value_2'];

$io = new SymfonyStyle($input, $output);

$value = $io->choice('Choose something', $choices, '1'); // $value would be 'value_1' with default choice
$key = $io->choice('Choose something', $choices, '1', true); // $key would be '1' with default choice
```

### DIFF
--- a/src/Symfony/Component/Console/Helper/QuestionHelper.php
+++ b/src/Symfony/Component/Console/Helper/QuestionHelper.php
@@ -174,14 +174,18 @@ class QuestionHelper extends Helper
         } elseif ($question instanceof ChoiceQuestion) {
             $choices = $question->getChoices();
 
-            if (!$question->isMultiselect()) {
-                return $choices[$default] ?? $default;
-            }
+            if ($question->isChoiceKeys()) {
+                return $question->isMultiselect() ? array_map(fn ($v) => trim($v), explode(',', $default)) : $default;
+            } else {
+                if (!$question->isMultiselect()) {
+                    return $choices[$default] ?? $default;
+                }
 
-            $default = explode(',', $default);
-            foreach ($default as $k => $v) {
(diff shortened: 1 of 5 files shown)


nicolas-grekas:
Closing due to inactivity: this has had no update and no review since August 2023. Note that ChoiceQuestion already returns the key when the choices are given as an associative array with string keys, and adding a fifth positional boolean to SymfonyStyle::choice() is not an API we would want to introduce. If you still need this for list-style choices, feel free to reopen with a proposal that avoids the extra boolean argument. Thanks.

PR #51202, opened 2023-08-01
[Uid] Add more PHP doc to "export" functions

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

I added theses PHPDoc because it's really convenient to get this information without testing.
And I guess not everyone know about the different RFC.

I also added the length because it's very useful when adding a new field in the DB

### DIFF
--- a/src/Symfony/Component/Uid/AbstractUid.php
+++ b/src/Symfony/Component/Uid/AbstractUid.php
@@ -88,6 +88,8 @@ abstract class AbstractUid implements \JsonSerializable, \Stringable
 
     /**
      * Returns the identifier as a base58 case sensitive string.
+     *
+     * @example 2AifFTC3zXgZzK5fPrrprL (len=22)
      */
     public function toBase58(): string
     {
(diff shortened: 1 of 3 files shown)


wouterj (approved):

nicolas-grekas (approved):

maxbeckers (approved):

wouterj:
More docs is always better, thanks Grégoire! :)

src/Symfony/Component/Uid/AbstractUid.php
```diff
@@ -129,6 +139,8 @@ public function toRfc4122(): string
 
     /**
      * Returns the identifier as a prefixed hexadecimal case insensitive string.
+     *
+     * @example 09748193048a4bfbb8258528cf74fdc1 (len=34)
```
andersonamuller, line 143: Isn't this actually 32 chars?

src/Symfony/Component/Uid/Ulid.php
```diff
@@ -114,6 +114,13 @@ public function toBinary(): string
         return hex2bin($ulid);
     }
 
+    /**
+     * Returns the identifier as a base32 case insensitive string.
+     *
+     * @see https://tools.ietf.org/html/rfc4648#section-6
+     *
+     * @example 09EJ0S614A9FXVG9C5537Q9ZE1 (len=26)
```
stof, line 122: this is not needed as it should inherit from the parent class.

lyrixx: I adeded them because it looks like vscode does not do that by default :( 

image

Issue #50976, opened 2023-07-14
Extend workflow component graph /debugger by event listener actions


kswzr:
### Description

When using the symfony workflow component, I have the issue, that I don't really can easily see, what happens before, on, after different transitions. 

It is only possible to see, what states can be reached and which transition can be applied.

I think it would be very helpful, to visualize in some way the different EventListeners which subscribes to the various transition events. Maybe in the Symfony Debugger, otherwise in the state machine graph. We could use here the EventListeners name (+ Method Name, if it does not have only an invoke method).

### Example

Imagine, we have a payment state machine with a transition "capture", from "processing" to "captured". 
When the state machine dispatches the "workflow.payment.completed.capture" event, I have an EventListener called "BookAccountingItemsListener::invoke", "PaymentNotificationListener::sendCapturedCustomerEmail".

Seeing that information in some way, could be helpful to understand and see the full process and business logic behind that.

lyrixx:
I totally agree. I'm working on a new project where we'll use massively the Workflow component in complex situation (Same entity, different workflow depending on the context). I'm planning to add new features to it!

lyrixx:
Small preview #teasing :D 
![image](https://github.com/symfony/symfony/assets/408368/75d751c8-efca-4b79-9bb2-82c93d747c95)

kswzr:
@lyrixx thanks for that amazing work and progress! Looks really good and as I have expected it :)

Issue #39114, opened 2020-11-18
[RFC][WebProfiler] Show services


jderusse:
**Description**  
I worked a POC to display services, but before spending too much time on this, I wonder if it worth it.

**Example**  
http://www.youtube.com/watch?v=d9tL17wvFjQ

![Screenshot from 2020-11-18 23-59-15](https://user-images.githubusercontent.com/578547/99598859-87b0b600-29fa-11eb-80f5-3f70fc0d2821.png)

yceruto:
Very nice feature! Could we also search into the container parameters?

fabpot:
The WDT/profiler is about storing/displaying data about a given request. I know that we do have some configuration displayed as well (but I've always been against doing it).

Perhaps, it's time to have a specific non-request related section (or sections) and identify them clearly. For this new section(s), the data would NOT be coming from the profile, but from the current configuration (to avoid storing too much data).

jderusse:
@fabpot agreed, this is one of the things that have to be "fixed" if I open a PR for this new feature.

I think the "config" panel is precisely the right place for such "non-request" related data.
I mean, what are the odds, the version of PHP or extensions or Symfony or list of bundles changed between 2 requests and the user cares about comparing the 2 requests?

stof:
Storing the list of services of the container would probably require storing a huge amount of non-request-specific data in the profile if we keep the existing architecture. So I would be against it.

If we make that a UI showing the current state, (similar to what `debug:container` does in the CLI), it will be better. But then, we need to figure out how it fits in the UI (putting everything in the `config` panel does not look like a solution to me, as that would make that panel huge, and potentially slow to load if it has to compute all these data).

stof:
Also, we already have the special `phpinfo` page in the profiler, which is about the existing config (but that one is not integrated in our UI at all, and the only way to exit it is the back button)

lyrixx:
@stof we don't need to store anything here. We could "simply" interact with the **current** container

jderusse:
> If we make that a UI showing the current state, (similar to what `debug:container` does in the CLI), it will be better. But then, we need to figure out how it fits in the UI (putting everything in the `config` panel does not look like a solution to me, as that would make that panel huge, and potentially slow to load if it has to compute all these data).

If you are talking about "putting every "non-request" related configuration in the same page", including this "feature" and probably future features, then yes probably, the page will growing and being huge one day or another, but we could think about that later when we'll face the issue when adding new features.

[...]

PR #51204, opened 2023-08-01
[Workflow] Add a profiler

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

---

With [this project lyrixx/SFLive-Paris2016-Workflow](https://github.com/lyrixx/SFLive-Paris2016-Workflow)

image

[workflow.webm](https://github.com/symfony/symfony/assets/408368/e86aad22-812d-461d-950f-68d50ac31390)


src/Symfony/Bundle/WebProfilerBundle/Resources/views/Icon/workflow.svg
lyrixx, line 1: I'm totally a bad UI designer, so I found this on the web, and I adapted it a bit.
Feel free to replace it with something better

cc @jonathanlevaillant

src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
```diff
@@ -0,0 +1,58 @@
...
+    public function lateCollect()
+    {
+        foreach ($this->workflows as $workflow) {
+            $type = $workflow instanceof StateMachine ? MermaidDumper::TRANSITION_TYPE_STATEMACHINE : MermaidDumper::TRANSITION_TYPE_WORKFLOW;
+            $dumper = new MermaidDumper($type);
+            $this->data['dumps'][$workflow->getName()] = $dumper->dump($workflow->getDefinition());
+        }
+    }
+
+    public function getName()
```
nicolas-grekas:
```suggestion
    public function getName(): string
```

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/workflow.html.twig
```diff
@@ -0,0 +1,61 @@
...
+            // "good size" if they are not visible
+            document.addEventListener('DOMContentLoaded', () => {
+                document.querySelectorAll('.tab').forEach((el) => {
+                    const observer = new MutationObserver(() => {
+                        if (!el.classList.contains('block')) {
+                            return;
+                        }
+                        const mEl = el.querySelector('.sf-mermaid');
+                        if (mEl.dataset.processed) {
+                            console.log('A');
```
nicolas-grekas: ?

src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
lyrixx: done, thx (not pushed yet).

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/workflow.html.twig
lyrixx: fixed

nicolas-grekas (approved):

stof:
This panel does not collect any info related to the request. It only collect info about the configuration of the project, which will be the same for all requests. In the past, we removed some panels like that.

lyrixx:
I thought about doing everything at runtime (I mean without all the collector stuff). But I changed my mind because the workflow can evolve between 2 requests. And we collect already some information about listener, doctrine entities. API-P does the same with the resources, and so on

And this panel is **really** useful. I'm working on a new projet, with a huge workflow, and it helped me a lot to build the configuration.

And I plan to add more information to it, like the list of all event listeners. This features has been requested in #50976 

So I would **be very happy** to get this merged, it'll be a game changer for everyone using the workflow component

fabpot:
We already one such panel: "configuration"

lyrixx:
Note: In future version, this panal will also display more things

* The current marking of the subject (if a call has been made to the workflow obliviously)
* Current content of `TransitionBlockerList`
* Perf metrics about all listeners

---

So it could not fit in the configuration panel

It will be something like the security panel : some configuration + some request related content

94noni:
i also find it useful even if not linked to request/response flow
reminds me https://github.com/symfony/symfony/issues/39114

fabpot (approved):

fabpot:
Thank you @lyrixx.

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Icon/workflow.svg
94noni, line 1: perhaps use https://tabler-icons.io/i/vector @lyrixx 
new design profiler take icons from this lib :)

stof: I would say than https://tabler-icons.io/i/git-fork better represents the idea of workflows than this square that you suggested.

94noni: indeed! just passing by ;)

lyrixx: good idea, I'll do that in another PR

andreybolonin:
@lyrixx I used your WorkflowExtension to dump Workflow builded on the fly (in Controller), but I see some blocks is rendered on grey background

Screenshot 2023-10-24 at 13 03 53

How you fixed this?

I used the same code on frontend (mermaid.js and the same ver of bootstrap.min.css)

lyrixx:
I'm not sure to understand your issue. Could could open a github discussion to talk about that? Thanks

PR #51207, opened 2023-08-01
[Process] Fix test case

### LINKED ISSUE #48486: ProcessTest::testWaitStoppedDeadProcess() passing even though it should actually fail

*

### Description

The `ProcessTest::testWaitStoppedDeadProcess()` uses the `ErrorProcessInitiator.php` which itself includes the Composer autoload.php. This file, however, does not exist at all in `symfony/symfony`.

### How to reproduce

```
    public function testWaitStoppedDeadProcess()
    {
        $process = $this->getProcess(self::$phpBin.' '.__DIR__.'/ErrorProcessInitiator.php -e '.self::$phpBin);
        $process->start();
        $process->setTimeout(2);
        $process->wait();
        $this->assertFalse($process->isRunning());

        var_dump($process->getOutput());
    }
```

Then run `./phpunit src/Symfony/Component/Process --filter ProcessTest::testWaitStoppedDeadProcess`

It will contain issues saying the `autoload.php` could not be found.

### Possible Solution

Not entirely sure because I did not write that test. Maybe it would be enough to include the `autoload.php` of the mono repository.

Also, I think we should output some `dummy` content and assert that output in order to have the test fail if suddenly something doesn't work anymore. At the moment, this is kind of a false-positive.

### Additional Context

_No response_

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #48486
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Symfony/Component/Process/Tests/ErrorProcessInitiator.php
+++ b/src/Symfony/Component/Process/Tests/ErrorProcessInitiator.php
@@ -14,12 +14,12 @@ namespace Symfony\Component\Process\Tests;
 use Symfony\Component\Process\Exception\ProcessTimedOutException;
 use Symfony\Component\Process\Process;
 
-require \dirname(__DIR__).'/vendor/autoload.php';
+require is_file(\dirname(__DIR__).'/vendor/autoload.php') ? \dirname(__DIR__).'/vendor/autoload.php' : \dirname(__DIR__, 5).'/vendor/autoload.php';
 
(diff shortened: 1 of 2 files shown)


stof (approved):

Toflar:
Thanks, Nicolas ❤️

PR #51211, opened 2023-08-01
[Workflow] List place and transition listeners in profiler

### LINKED ISSUE #50976: Extend workflow component graph /debugger by event listener actions

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | fixes #50976
| License       | MIT
| Doc PR        |

---

image

[workflow.webm](https://github.com/symfony/symfony/assets/408368/06821a54-77a3-4c7e-8a74-3a64092aab98)

### DIFF
--- a/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/workflow.html.twig
+++ b/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/workflow.html.twig
@@ -1,5 +1,102 @@
 {% extends '@WebProfiler/Profiler/layout.html.twig' %}
 
+{% block stylesheets %}
+    {{ parent() }}
+    <style>
+        dialog {
+            border: none;
+            border-radius: 6px;
+            box-shadow: var(--settings-modal-shadow);
+            max-width: 94%;
+            width: 1200px;
+        }
+
+        dialog::backdrop {
+            background: linear-gradient(
+                45deg,
+                rgb(18, 18, 20, 0.4),
+                rgb(17, 17, 20, 0.8)
+            );
+        }
+
+        dialog[open] {
+            animation: scale 0.3s ease normal;
+        }
+
+        dialog[open]::backdrop {
+            animation: backdrop 0.3s ease normal;
+        }
+
+        dialog.hide {
+            animation-direction: reverse;
+        }
+
+        dialog h2 {
+            margin-top: 0.2em
+        }
+
+        dialog i.cancel {
+            cursor: pointer;
+            padding: 0 5px;
+            float: right;
+        }
+
+        dialog table {
+            border: 1px solid #ccc;
+            border-collapse: collapse;
+            margin: 0 0 1em 0;
+            margin-bottom: 1em;
+            padding: 0;
+            table-layout: fixed;
+        }
+
+        dialog table tr {
+            background-color: #f8f8f8;
+            border: 1px solid #ddd;
+            padding: .35em;
+        }
+
+        dialog table th,
+        dialog table td {
+            padding: .625em;
+            text-align: center;
+        }
+
+        dialog table th {
+            font-size: .85em;
+            letter-spacing: .1em;
+            text-transform: uppercase;
+        }
+
+        dialog menu {
+            padding: 0;
+            margin: 0;
+            display: flex;
+            align-items: center;
+            flex-direction: row;
+            vertical-align: middle;
+            justify-content: center;
+        }
+
+        dialog menu small {
+            margin-right: auto;
+        }
+        dialog menu small i {
+            margin-right: 3px;
+        }
+
+        @keyframes scale {
+            from { transform: scale(0); }
+            to { transform: scale(1); }
+        }
+
+        @keyframes backdrop {
+            from { opacity: 0; }
+            to { opacity: 1; }
+        }
+    </style>
+{% endblock %}
+
 {% block toolbar %}
     {% if collector.callsCount > 0 %}
         {% set icon %}
@@ -40,6 +137,93 @@
                 flowchart: { useMaxWidth: false },
                 securityLevel: 'loose',
             });
+
+            {% for name, data in collector.workflows %}
+                window.showNodeDetails{{ collector.hash(name) }} = function (node) {
+                    const map = {{ data.listeners|json_encode|raw }};
+                    showNodeDetails(node, map);
+                };
+            {% endfor %}
+
+            const showNodeDetails = function (node, map) {
+                const dialog = document.getElementById('detailsDialog');
+
+                dialog.querySelector('tbody').innerHTML = '';
+                for (const [eventName, listeners] of Object.entries(map[node])) {
+                    listeners.forEach(listener => {
+                        const row = document.createElement('tr');
+
+                        const eventNameCode = document.createElement('code');
+                        eventNameCode.textContent = eventName;
+
+                        const eventNameCell = document.createElement('td');
+                        eventNameCell.appendChild(eventNameCode);
+                        row.appendChild(eventNameCell);
+
+                        const listenerDetailsCell = document.createElement('td');
+                        row.appendChild(listenerDetailsCell);
+
+                        let listenerDetails;
+                        const listenerDetailsCode = document.createElement('code');
+                        listenerDetailsCode.textContent = listener.title;
+                        if (listener.file) {
+                            const link = document.createElement('a');
+                            link.href = listener.file;
+                            link.appendChild(listenerDetailsCode);
+                            listenerDetails = link;
+                        } else {
+                            listenerDetails = listenerDetailsCode;
+                        }
+                        listenerDetailsCell.appendChild(listenerDetails);
+
+                        if (typeof listener.guardExpressions === 'object') {
+                            listenerDetailsCell.appendChild(document.createElement('br'));
+
+                            const guardExpressionsWrapper = document.createElement('span');
+                            guardExpressionsWrapper.appendChild(document.createTextNode('guard expressions: '));
+
+                            listener.guardExpressions.forEach((expression, index) => {
+                                if (index > 0) {
+                                    guardExpressionsWrapper.appendChild(document.createTextNode(', '));
+                                }
+
+                                const expressionCode = document.createElement('code');
+                                expressionCode.textContent = expression;
+                                guardExpressionsWrapper.appendChild(expressionCode);
+                            });
+
+                            listenerDetailsCell.appendChild(guardExpressionsWrapper);
+                        }
+
+                        dialog.querySelector('tbody').appendChild(row);
+                    });
+                };
+
+                if (dialog.dataset.processed) {
+                    dialog.showModal();
+                    return;
+                }
+
+                dialog.addEventListener('click', (e) => {
+                    const rect = dialog.getBoundingClientRect();
+
+                    const inDialog =
+                        rect.top <= e.clientY &&
+                        e.clientY <= rect.top + rect.height &&
+                        rect.left <= e.clientX &&
+                        e.clientX <= rect.left + rect.width;
+
+                    !inDialog && dialog.close();
+                });
+
+                dialog.querySelectorAll('.cancel').forEach(elt => {
+                    elt.addEventListener('click', () => dialog.close());
+                });
+
+                dialog.showModal();
+
+                dialog.dataset.processed = true;
+            };
             // We do not load all mermaid diagrams at once, but only when the tab is opened
             // This is because mermaid diagrams are in a tab, and cannot be renderer with a
             // "good size" if they are not visible
@@ -71,6 +255,9 @@
                         <h3>Definition</h3>
                         <pre class="sf-mermaid">
                             {{ data.dump|raw }}
+                            {% for nodeId, events in data.listeners %}
+                                click {{ nodeId }} showNodeDetails{{ collector.hash(name) }}
+                            {% endfor %}
                         </pre>
 
                         <h3>Calls</h3>
@@ -128,4 +315,26 @@
             {% endfor %}
         </div>
     {% endif %}
+
+    <dialog id="detailsDialog">
+        <h2>
+            Event listeners
+            <i class="cancel">×</i>
+        </h2>
+
+        <table>
+            <thead>
+                <tr>
+                    <th>event</th>
+                    <th>listener</th>
+                </tr>
+            </thead>
+            <tbody>
+            </tbody>
+        </table>
+        <menu>
+            <small><i>⌨</i> <kbd>esc</kbd></small>
+            <button class="btn btn-sm cancel">Close</button>
+        </menu>
+    </dialog>
 {% endblock %}
--- a/src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
+++ b/src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
@@ -11,6 +11,8 @@
 
 namespace Symfony\Component\Workflow\DataCollector;
 
+use Symfony\Component\ErrorHandler\ErrorRenderer\FileLinkFormatter;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\DataCollector\DataCollector;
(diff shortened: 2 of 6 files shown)


src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
```diff
@@ -57,4 +66,115 @@ public function getWorkflows(): array
     {
         return $this->data['workflows'] ?? [];
     }
+
+    public function hash(string $string): string
+    {
+        return hash('sha256', $string);
```
nicolas-grekas:
```suggestion
        return hash('xxh128', $string);
```

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/workflow.html.twig
```diff
@@ -23,6 +145,54 @@
...
+                dialog.querySelector('tbody').innerHTML = '';
+                for (const [eventName, listeners] of Object.entries(map[node])) {
+                    listeners.forEach(listener => {
+                        const row = document.createElement('tr');
+                        let listenerDetails = `<code>${listener.title}</code>`;
+                        if (listener.file) {
+                            listenerDetails = `<a href="${listener.file}">${listenerDetails}</a>`;
+                        }
+                        if (typeof listener.guardExpressions === 'object') {
+                            console.log('XXX"');
```
stloyd: :)

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/workflow.html.twig
```diff
@@ -1,5 +1,127 @@
...
+        :root {
+            --bg: rgb(26, 27, 30);
+            --color: #ffffff;
+            --button-bg: rgb(50, 52, 61);
+            --button-color: #ffffff;
+            --warn-bg: rgb(198, 48, 48);
+            --modal-color: rgb(155, 155, 160);
+        }
+
+        button {
```
stof: applying this on all buttons on the page is a no-go. It can break other parts of the design.

src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
```diff
@@ -57,4 +66,115 @@ public function getWorkflows(): array
...
+            'title' => $title,
+            'file' => $file,
+            ...$extra,
+        ];
+    }
+
+    private function extractGuardExpressions(GuardListener $listener, string $eventName, Transition $transition): array
+    {
+        $r = new \ReflectionProperty(GuardListener::class, 'configuration');
+        $r->setAccessible(true);
```
stloyd: No needed: https://wiki.php.net/rfc/make-reflection-setaccessible-no-op

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/workflow.html.twig
```diff
@@ -1,5 +1,127 @@
...
+{% block stylesheets %}
+    {{ parent() }}
+    <style>
+        :root {
+            --bg: rgb(26, 27, 30);
+            --color: #ffffff;
+            --button-bg: rgb(50, 52, 61);
+            --button-color: #ffffff;
+            --warn-bg: rgb(198, 48, 48);
+            --modal-color: rgb(155, 155, 160);
```
stof: why do we have dedicated colors there instead of relying on the color variables of the main profiler styles (which would then have support for the dark and light themes)

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/workflow.html.twig
```diff
@@ -23,6 +145,54 @@
...
+                        if (listener.file) {
+                            listenerDetails = `<a href="${listener.file}">${listenerDetails}</a>`;
+                        }
+                        if (typeof listener.guardExpressions === 'object') {
+                            console.log('XXX"');
+                            listenerDetails = `${listenerDetails} (guard expressions: <code>${listener.guardExpressions.join('</code>, <code>')})</code>`;
+                        }
+                        row.innerHTML = `
+                            <td><code>${eventName}</code></td>
+                            <td>${listenerDetails}</td>
```
stof: this misses HTML escaping. This is a no-go.

stof: if you don't want to care about escaping manually, the solution is to create DOM nodes instead of creating HTML (and then, you can assign `textContent` or create text nodes)

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/workflow.html.twig
```diff
@@ -23,6 +145,54 @@
...
+                        `;
+                        dialog.querySelector('tbody').appendChild(row);
+                    });
+                };
+
+                dialog.addEventListener('click', (e) => {
```
stof, line 207: This listener should not be added again each time the modal is shown, otherwise we will have multiple ones when we use the dialog multiple times, which might do weird things and will definitely leak

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/workflow.html.twig
stof: maybe a simpler way would be to create the whole dialog in Twig, with different dialogs per workflow.

stof (changes requested):

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/workflow.html.twig
lyrixx: I "backported" something I had in another app, outside the profiler. I'll try to reduce this style

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/workflow.html.twig
stof: if you cannot add a class on the button itself due to being in mermaid, use `.sf-mermaid button`

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/workflow.html.twig
lyrixx: I did that initially, but the HTML will be SO huge! 
N workflow * M transition * O place === the number of modal :(

src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
```diff
@@ -57,4 +66,115 @@ public function getWorkflows(): array
...
+        $extra = [];
+
+        if ($callable instanceof \Closure || \is_string($callable)) {
+            $r = new \ReflectionFunction($callable);
+            $title = (string) $r;
+        } else  if (\is_object($callable) && method_exists($callable, '__invoke')) {
+            $r = new \ReflectionMethod($callable, '__invoke');
+            $title = \get_class($callable) . '::__invoke()';
+        } elseif (\is_array($callable)) {
+            if ($callable[0] instanceof GuardListener) {
```
stof, line 188: Will this work fine if the dispatcher is traceable ?

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/workflow.html.twig
stof: ah indeed. then use proper DOM building.

src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
lyrixx: I don't know how, but yes.
I already get the traceable dispatcher 

image

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/workflow.html.twig
lyrixx: I removed many lines of CSS!

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/workflow.html.twig
lyrixx: Fixed ✅

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/workflow.html.twig
```diff
@@ -23,6 +120,87 @@
...
+                        listenerDetailsCell.appendChild(listenerDetails);
+
+                        if (typeof listener.guardExpressions === 'object') {
+                            listenerDetailsCell.appendChild(document.createElement('br'));
+
+                            const guardExpressionsWrapper = document.createElement('span');
+                            guardExpressionsWrapper.appendChild(document.createTextNode('guard expressions: '));
+
+                            const guardExpressionsCode = document.createElement('code');
+                            guardExpressionsCode.textContent = listener.guardExpressions.join('</code>, <code>');
```
stof: this is broken. this will create a single code element with `foo</code>, <code>bar` as textual content. you need to create one `code` element per guard expression and add `, ` text nodes between them.

lyrixx: Thanks, fixed

lyrixx:
This PR is ready

lyrixx:
This PR is ready for 2 month. I'm gonna merge it by the end of the week if there are not more review

src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
```diff
@@ -24,6 +30,8 @@ final class WorkflowDataCollector extends DataCollector implements LateDataColle
 {
     public function __construct(
         private readonly iterable $workflows,
+        private readonly EventDispatcherInterface $eventDispatcher,
+        private FileLinkFormatter $fileLinkFormatter,
```
stof: those should be optional arguments to preserve BC

src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
```diff
@@ -57,4 +66,113 @@ public function getWorkflows(): array
...
+                'enter',
+                'entered',
+            ];
+            foreach ($subEventNames as $subEventName) {
+                $eventNames[] = sprintf('workflow.%s', $subEventName);
+                $eventNames[] = sprintf('workflow.%s.%s', $workflow->getName(), $subEventName);
+                $eventNames[] = sprintf('workflow.%s.%s.%s', $workflow->getName(), $subEventName, $place);
+            }
+            foreach ($eventNames as $eventName) {
+                // @phpstan-ignore-next-line
```
stof: should be removed (we don't run phpstan anyway

stof:
@lyrixx any chance to fix my latest comments

src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
lyrixx: This class has never been released :)

lyrixx:
Hello @stof. I addressed your comment. Thanks

src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
```diff
@@ -57,4 +66,111 @@ public function getWorkflows(): array
...
+        }
+
+        return $listeners;
+    }
+
+    private function summarizeListener(callable $callable, string $eventName = null, Transition $transition = null): array
+    {
+        $extra = [];
+
+        if ($callable instanceof \Closure || \is_string($callable)) {
```
nicolas-grekas: I think we can improve this method by reusing (or borrowing from) what we do in src/Symfony/Component/EventDispatcher/Debug/WrappedListener.php

nicolas-grekas (commented):
Just one comment.
What about adding some tests?

src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
lyrixx: This is what I did, and I adapted it :) (or it was from another piece of Symfony, I don't remember)

lyrixx:
@nicolas-grekas I added some tests 👍🏼

nicolas-grekas (approved):

src/Symfony/Component/Workflow/Tests/DataCollector/WorkflowDataCollectorTest.php
```diff
@@ -0,0 +1,92 @@
...
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Serializer\Tests\DataCollector;
```
fabpot:
```suggestion
namespace Symfony\Component\Workflow\Tests\DataCollector;
```

src/Symfony/Component/Workflow/composer.json
```diff
@@ -26,6 +26,7 @@
     "require-dev": {
         "psr/log": "^1|^2|^3",
         "symfony/dependency-injection": "^5.4|^6.0|^7.0",
+        "symfony/error-handler": "^6.4|^7.0",
```
fabpot, line 29: You should also add `http-kernel` right?

fabpot (changes requested):
fabbot has a patch for you :)

src/Symfony/Component/Workflow/Tests/DataCollector/WorkflowDataCollectorTest.php
```diff
@@ -0,0 +1,92 @@
...
+        $this->assertArrayHasKey('dump', $data['workflow2']);
+        $this->assertStringStartsWith("graph LR\n", $data['workflow1']['dump']);
+        $this->assertArrayHasKey('listeners', $data['workflow2']);
+        $listeners = $data['workflow2']['listeners'];
+        $this->assertArrayHasKey('place0', $listeners);
+        $this->assertArrayHasKey('workflow.workflow2.leave.a', $listeners['place0']);
+        $descriptions = $listeners['place0']['workflow.workflow2.leave.a'];
+        $this->assertCount(5, $descriptions);
+        $this->assertStringContainsString('Closure', $descriptions[0]['title']);
+        $this->assertSame('Symfony\Component\Serializer\Tests\DataCollector\WorkflowDataCollectorTest::noop()', $descriptions[1]['title']);
```
fabpot:
```suggestion
        $this->assertSame('Symfony\Component\Workflow\Tests\DataCollector\WorkflowDataCollectorTest::noop()', $descriptions[1]['title']);
```

fabpot: same for the two below

fabpot (changes requested):

src/Symfony/Component/Workflow/Tests/DataCollector/WorkflowDataCollectorTest.php
lyrixx: Rah! Thanks

fabpot (approved):

fabpot:
Thank you @lyrixx.

PR #51214, opened 2023-08-01
[Workflow] Use TRANSITION_TYPE_WORKFLOW for rendering workflow in profiler

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

---

This one is a bit special.

Usually a state machine render like this:

image

As you can see, **transitions** (ex: `start_process`) are not real nodes. They are arrows.

And it's not possible to attach events on arrows, only on nodes.

That's why I add this hybrid mode, where state machine are rendered like a workflow. It means transitions are renderered as node:

image

Now, we can attach event on transition nodes.

### DIFF
--- a/src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
+++ b/src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
@@ -16,7 +16,6 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\DataCollector\DataCollector;
 use Symfony\Component\HttpKernel\DataCollector\LateDataCollectorInterface;
 use Symfony\Component\Workflow\Dumper\MermaidDumper;
-use Symfony\Component\Workflow\StateMachine;
 
 /**
  * @author Grégoire Pineau <lyrixx@lyrixx.info>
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
```diff
@@ -35,8 +34,7 @@ public function collect(Request $request, Response $response, \Throwable $except
     public function lateCollect(): void
     {
         foreach ($this->workflows as $workflow) {
-            $type = $workflow instanceof StateMachine ? MermaidDumper::TRANSITION_TYPE_STATEMACHINE : MermaidDumper::TRANSITION_TYPE_WORKFLOW;
-            $dumper = new MermaidDumper($type);
+            $dumper = new MermaidDumper(MermaidDumper::TRANSITION_TYPE_HYBRID);
```
stof: couldn't you use `TRANSITION_TYPE_WORKFLOW` all the time here ? AFAICT, the hybrid mode is actually the same than the workflow mode.

lyrixx: You are right, I simplified the PR.

stof (approved):

fabpot (approved):

fabpot:
Thank you @lyrixx.

PR #51215, opened 2023-08-01
[FrameworkBundle] Enable `json_decode_detailed_errors` in dev by default

ostrolucky:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix https://github.com/symfony/symfony/pull/51172#issuecomment-1659846641
| License       | MIT
| Doc PR        | 

Follows
* https://github.com/symfony/symfony/pull/51172


src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Add native return type to `Translator` and to `Application::reset()`
  * Deprecate the integration of Doctrine annotations, either uninstall the `doctrine/annotations` package or disable the integration by setting `framework.annotations` to `false`
+ * Enable `json_decode_detailed_errors` context for Serializer by default if `kernel.debug` is true and `seld/jsonlint` is installed
```
OskarStark:
```suggestion
 * Enable `json_decode_detailed_errors` context for Serializer by default if `kernel.debug` is true and the `seld/jsonlint` package is installed
```

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1123,7 +1125,10 @@ private function addSerializerSection(ArrayNodeDefinition $rootNode, callable $e
                         ->arrayNode('default_context')
                             ->normalizeKeys(false)
                             ->useAttributeAsKey('name')
-                            ->defaultValue([])
+                            ->defaultValue($this->debug && class_exists(JsonParser::class)
+                                ? [JsonDecode::DETAILED_ERROR_MESSAGES => true]
+                                : [],
+                            )
```
```suggestion
                            ->defaultValue($this->debug && class_exists(JsonParser::class) ? [JsonDecode::DETAILED_ERROR_MESSAGES => true] : []))
```

nicolas-grekas: wouldn't it be better to do this in the extension, so that people that configure a default context can still get the new default?

ostrolucky: This works too in such case. Thanks to using hashmap instead of list and the way how symfony/config merge things, similarly as this works
```yaml
framework:
  foo:
when@dev:
  framework:
    bar:
```
would in `@dev` result into
```yaml
framework:
  foo:
  bar:
```

derrabus: Can we add functional tests that makes sure the default is properly merged into the user-provided context?

ostrolucky: Added. And indeed, can't use defaultValue() for this.

nicolas-grekas (approved):
This works because performNoDeepMerging is not set on the node, OK.

fabpot (approved):

fabpot:
Thank you @ostrolucky.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1123,6 +1125,10 @@ private function addSerializerSection(ArrayNodeDefinition $rootNode, callable $e
                         ->arrayNode('default_context')
                             ->normalizeKeys(false)
                             ->useAttributeAsKey('name')
+                            ->beforeNormalization()
```
stof, line 1128: Adding that default in `beforeNormalization` makes it very hard to re-disable the feature because each source will re-enable it in this normalization (well, each source configuring the `default_context`).
This should be done in the `validate` phase instead.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1123,6 +1125,10 @@ private function addSerializerSection(ArrayNodeDefinition $rootNode, callable $e
...
+                                ->ifTrue(fn () => $this->debug && class_exists(JsonParser::class))
+                                ->then(fn (array $v) => $v + [JsonDecode::DETAILED_ERROR_MESSAGES => true])
+                            ->end()
                             ->defaultValue([])
```
stof, line 1132: the default value should probably also include `JsonDecode::DETAILED_ERROR_MESSAGES => true` (when appropriate) so that this configuration also applies when no default_context is configured at all.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
xabbuh: see #57803

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
xabbuh: see #57804

PR #51220, opened 2023-08-01
[Workflow] Add a `TraceableWorkflow`

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

image

image

---

Initially, I wanted to display the current marking on the graph, but it's not possible. "current" does not mean anything. What if I use the workflow for 2 entities? What if I call twice `apply()`. Which marking should I choose? => this is not possible.

But, DoctrineBundle use ajax, like for explaining SQL request. It would be nice to be able to click on a marking, and it'll re-render the graph with this marking. WDYT?

### DIFF
--- a/src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
+++ b/src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
@@ -15,7 +15,12 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\DataCollector\DataCollector;
 use Symfony\Component\HttpKernel\DataCollector\LateDataCollectorInterface;
+use Symfony\Component\VarDumper\Caster\Caster;
+use Symfony\Component\VarDumper\Cloner\Stub;
+use Symfony\Component\Workflow\Debug\TraceableWorkflow;
 use Symfony\Component\Workflow\Dumper\MermaidDumper;
+use Symfony\Component\Workflow\Marking;
+use Symfony\Component\Workflow\TransitionBlocker;
 
 /**
  * @author Grégoire Pineau <lyrixx@lyrixx.info>
@@ -34,11 +39,17 @@ final class WorkflowDataCollector extends DataCollector implements LateDataColle
     public function lateCollect(): void
     {
         foreach ($this->workflows as $workflow) {
+            $calls = [];
+            if ($workflow instanceof TraceableWorkflow) {
+                $calls = $this->cloneVar($workflow->getCalls());
+            }
+
             // We always use a workflow type because we want to mermaid to
             // create a node for transitions
             $dumper = new MermaidDumper(MermaidDumper::TRANSITION_TYPE_WORKFLOW);
             $this->data['workflows'][$workflow->getName()] = [
                 'dump' => $dumper->dump($workflow->getDefinition()),
+                'calls' => $calls,
             ];
         }
     }
@@ -56,5 +67,39 @@ final class WorkflowDataCollector extends DataCollector implements LateDataColle
     public function getWorkflows(): array
     {
         return $this->data['workflows'] ?? [];
+    }
+
+    public function getCallsCount(): int
+    {
+        $i = 0;
+        foreach ($this->getWorkflows() as $workflow) {
+            $i += \count($workflow['calls']);
+        }
+
+        return $i;
+    }
+
+    protected function getCasters(): array
+    {
+        $casters = [
+            ...parent::getCasters(),
+            TransitionBlocker::class => function ($v, array $a, Stub $s, $isNested) {
+                unset(
+                    $a[sprintf(Caster::PATTERN_PRIVATE, $v::class, 'code')],
+                    $a[sprintf(Caster::PATTERN_PRIVATE, $v::class, 'parameters')],
+                );
+
+                $s->cut += 2;
+
+                return $a;
+            },
+            Marking::class => function ($v, array $a, Stub $s, $isNested) {
+                $a[Caster::PREFIX_VIRTUAL.'.places'] = array_keys($v->getPlaces());
+
+                return $a;
+            },
+        ];
+
+        return $casters;
     }
 }
--- a/src/Symfony/Component/Workflow/Debug/TraceableWorkflow.php
+++ b/src/Symfony/Component/Workflow/Debug/TraceableWorkflow.php
@@ -1,0 +1,122 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Workflow\Debug;
+
+use Symfony\Component\Stopwatch\Stopwatch;
+use Symfony\Component\Workflow\Definition;
+use Symfony\Component\Workflow\Marking;
+use Symfony\Component\Workflow\MarkingStore\MarkingStoreInterface;
+use Symfony\Component\Workflow\Metadata\MetadataStoreInterface;
+use Symfony\Component\Workflow\TransitionBlockerList;
+use Symfony\Component\Workflow\WorkflowInterface;
+
+/**
+ * @author Grégoire Pineau <lyrixx@lyrixx.info>
+ */
+class TraceableWorkflow implements WorkflowInterface
+{
+    private array $calls = [];
+
+    public function __construct(
+        private readonly WorkflowInterface $workflow,
+        private readonly Stopwatch $stopwatch,
+    ) {
+    }
+
+    public function getMarking(object $subject, array $context = []): Marking
+    {
+        return $this->callInner(__FUNCTION__, \func_get_args());
+    }
+
+    public function can(object $subject, string $transitionName): bool
+    {
+        return $this->callInner(__FUNCTION__, \func_get_args());
+    }
+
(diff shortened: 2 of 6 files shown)


src/Symfony/Component/Workflow/TraceableWorkflow.php
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    public function getCalls(): array
+    {
+        return $this->calls;
+    }
+
+    private function callInner(string $method, array $args): mixed
+    {
+        $startAt = microtime(true);
```
94noni: why not using a debug.stopwatch? to have the performance tab "for free"

src/Symfony/Component/Workflow/TraceableWorkflow.php
```diff
@@ -0,0 +1,112 @@
...
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Workflow;
```
94noni:
```suggestion
namespace Symfony\Component\Workflow\Debug;
```

i am used to seeing some debug classes inside a debug subnamespace

src/Symfony/Component/Workflow/Debug/TraceableWorkflow.php
```diff
@@ -0,0 +1,120 @@
...
+        try {
+            return $return = $this->workflow->{$method}(...$args);
+        } catch (\Throwable $exception) {
+            throw $exception;
+        } finally {
+            $periods = $this->stopwatch->stop($sMethod)->getPeriods();
+            $period = end($periods);
+            $call = [
+                'method' => $method,
+                'duration' => $period->getDuration(),
```
94noni: thx for having swap to stopwatch, what bout adding here:

```suggestion
                'duration' => $period->getDuration(),
                'memory' => $period->getMemory(),
```

would be awesome, as per internal usage we can plug some ±heavy listener/subscriber on workflow event, and having both duration and memory can help fast check on what cant be improved :)

lyrixx: I'm not sure it's really useful, since here it's not per event, but for the whole method (`can`, `apply`, etc). So you won't know what use memory. 

I'm working on adding profiler for each listener - but to be honest, it's super hard :(

94noni: copy that, lets see how it goes :)

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/WorkflowDebugPass.php
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Symfony\Bundle\FrameworkBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\Workflow\Debug\TraceableWorkflow;
+
+/**
+ * Adds all configured security voters to the access decision manager.
```
94noni: comment seems wrong

94noni (approved):

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/WorkflowDebugPass.php
lyrixx: thanks, fixed

src/Symfony/Component/Workflow/Debug/TraceableWorkflow.php
stof: Profiling memort is also quite hard. As soon as you have nested periods, things become wonky because the memory used by the inner traceable object now counts as well for the outside span.
Memory profiling is better done with an actual profiler tool than this traceable instrumentation. It will be less skewed.

HeahDude: >I'm working on adding profiler for each listener - but to be honest, it's super hard :(

@lyrixx It should be done by default if the workflow uses the event dispatcher service, it is traceable in debug and events should be listed in the Events profiler panel since always.

lyrixx: @HeahDude indeed, the profiler shows some data, but it's not as good as what you can see in the Security profiler

### Workflow

image

### Security 

image


### TL;DR:

I would like something with

* all listener called
  * timing information
* for the guard
  * the "response"

lyrixx:
This PR is ready for 2 month. I'm gonna merge it by the end of the week if there are not more review

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/WorkflowDebugPass.php
```diff
@@ -0,0 +1,37 @@
+<?php
```
fabpot: This class should probably be part of the component.

fabpot:
@lyrixx Last call for 6.4 :)

src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
```diff
@@ -57,4 +68,36 @@ public function getWorkflows(): array
...
+            TransitionBlocker::class => function ($v, array $a, Stub $s, $isNested) {
+                unset(
+                    $a[sprintf(Caster::PATTERN_PRIVATE, $v::class, 'code')],
+                    $a[sprintf(Caster::PATTERN_PRIVATE, $v::class, 'parameters')],
+                );
+
+                return $a;
+            },
+            Marking::class => function ($v, array $a, Stub $s, $isNested) {
+                $a[sprintf(Caster::PATTERN_PRIVATE, $v::class, 'places')] = array_keys($v->getPlaces());
```
stof: shouldn't this use the virtual prefix instead ?

src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
```diff
@@ -57,4 +68,36 @@ public function getWorkflows(): array
...
+    }
+
+    protected function getCasters()
+    {
+        $casters = [
+            ...parent::getCasters(),
+            TransitionBlocker::class => function ($v, array $a, Stub $s, $isNested) {
+                unset(
+                    $a[sprintf(Caster::PATTERN_PRIVATE, $v::class, 'code')],
+                    $a[sprintf(Caster::PATTERN_PRIVATE, $v::class, 'parameters')],
```
stof, line 89: Shouldn't this mark the stub as cut ?

src/Symfony/Component/Workflow/Debug/TraceableWorkflow.php
```diff
@@ -0,0 +1,120 @@
...
+            try {
+                $previousMarking = $this->workflow->getMarking($args[0]);
+            } catch (\Throwable) {
+            }
+        }
+
+        try {
+            return $return = $this->workflow->{$method}(...$args);
+        } catch (\Throwable $exception) {
+            throw $exception;
```
stof: This catch block should be removed entirely. It makes no sense.

stof: OK, actually, it is not useless because it assigns the variable for the finally clause.
Maybe the code should be refactored without using finally but separate registration of the failed and successful calls.

src/Symfony/Component/Workflow/Debug/TraceableWorkflow.php
stof: Shouldn't this use the event duration instead of only the last period ?

src/Symfony/Component/Workflow/Registry.php
```diff
@@ -43,7 +43,10 @@ public function has(object $subject, string $workflowName = null): bool
         return false;
     }
 
-    public function get(object $subject, string $workflowName = null): Workflow
+    /**
+     * @return WorkflowInterface
+     */
+    public function get(object $subject, string $workflowName = null)
```
stof: If there was already a native return type, why widening it to `mixed` to restrict it again in the next major version instead of widening it to WorkflowInterface ?

lyrixx:
Hello @fabpot, @stof and others reviewer. 

I addressed your comments, I think the PR is ready 👍🏼

src/Symfony/Component/Workflow/DataCollector/WorkflowDataCollector.php
```diff
@@ -57,4 +68,38 @@ public function getWorkflows(): array
...
+                );
+
+                $s->cut += 2;
+
+                return $a;
+            },
+            Marking::class => function ($v, array $a, Stub $s, $isNested) {
+                $a[Caster::PREFIX_VIRTUAL] = array_keys($v->getPlaces());
```
stof: you should still name your property after the prefix.

```suggestion
                $a[Caster::PREFIX_VIRTUAL.'places'] = array_keys($v->getPlaces());
```

lyrixx: rah, 🤦🏼 thanks !

fabpot (approved):

fabpot:
Thank you @lyrixx.

PR #51223, opened 2023-08-02
[Console] Fix linewraps in `OutputFormatter`

### LINKED ISSUE #42034: Umlauts are incorrectly separated by the OutputFormatter when creating a table for the console
**Symfony version(s) affected**: 5.3

**Description**
I noticed that umlauts lead to errors when creating a table for the Symfony console. If a row breaks into several lines and the break occurs at an umlaut, it will be destroyed. The table is then too short by a few characters.

**How to reproduce**
```
$table = new \Symfony\Component\Console\Helper\Table($output);
$table
    ->setColumnMaxWidth(1, 10)
    ->setHeaders(['ISBN', 'Title'])
    ->setRows([
        ['99921-58-10-7', 'A really long title that could need multiple lines'],
        new \Symfony\Component\Console\Helper\TableSeparator(),
        ['99921-58-10-7', 'Â rèälly löng tîtlè thät cöüld nèêd múltîplê línès']
    ])
;
$table->render();
```

A sample code with a command class and multiple tests can be found here: https://gist.github.com/Moskito89/7774c29805248a2d2568a01997ad1bfb 

**Possible Solution**
The problem is caused in the `Symfony\Component\Console\Formatter\OutputFormatter` class. In the method `applyCurrentStyle` to be precise. The problem is that `preg_replace` on line 259 doesn't work with UTF8. (https://github.com/symfony/console/blob/d927f5564049730e2589d4d53c403ede528d6967/Formatter/OutputFormatter.php#L259)
This can be solved by decoding `$text` and `$prefix` beforehand with `utf8_decode` and later encoding them again with `utf8_encode`.

Please tell me if you can confirm that behaviour. Also please let me know if you want me to change that and create a PR to solve it. Thanks! 👍

maxbeckers:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #42034
| License       | MIT
| Doc PR        | n/a

Fix output for tables with linebreaks and special chars;
```php
$table = new \Symfony\Component\Console\Helper\Table($output);
$table
    ->setColumnMaxWidth(1, 10)
    ->setHeaders(['ISBN', 'Title'])
    ->setRows([
        ['99921-58-10-7', 'A really long title that could need multiple lines'],
        new \Symfony\Component\Console\Helper\TableSeparator(),
        ['99921-58-10-7', 'Â rèälly löng tîtlè thät cöüld nèêd múltîplê línès']
    ])
;
$table->render();
```

**Before**
```
+---------------+------------+
| ISBN          | Title      |
+---------------+------------+
| 99921-58-10-7 | A really l |
|               | ong title  |
|               | that could |
|               | need multi |
|               | ple lines  |
+---------------+------------+
| 99921-58-10-7 | Â rèäll    |
|               | y löng t |
|               | tlè thä |
|               | t cöüld    |
|               | nèêd mú    |
|               | ltîplê l   |
|               | ínès       |
+---------------+------------+
```

**After**
```
+---------------+------------+
| ISBN          | Title      |
+---------------+------------+
| 99921-58-10-7 | A really   |
|               | long title |
|               | that could |
|               | need       |
|               | multiple   |
|               | lines      |
+---------------+------------+
| 99921-58-10-7 | Â rèälly   |
|               | löng tîtlè |
|               | thät cöüld |
|               | nèêd       |
|               | múltîplê   |
|               | línès      |
+---------------+------------+

```

### DIFF
--- a/src/Symfony/Component/Console/Formatter/OutputFormatter.php
+++ b/src/Symfony/Component/Console/Formatter/OutputFormatter.php
@@ -13,6 +13,8 @@ namespace Symfony\Component\Console\Formatter;
 
 use Symfony\Component\Console\Exception\InvalidArgumentException;
 
+use function Symfony\Component\String\b;
+
 /**
  * Formatter class for console output.
  *
@@ -258,7 +260,7 @@ class OutputFormatter implements WrappableOutputFormatterInterface
         }
 
         preg_match('~(\\n)$~', $text, $matches);
-        $text = $prefix.preg_replace('~([^\\n]{'.$width.'})\\ *~', "\$1\n", $text);
+        $text = $prefix.$this->addLineBreaks($text, $width);
         $text = rtrim($text, "\n").($matches[1] ?? '');
 
         if (!$currentLineLength && '' !== $current && "\n" !== substr($current, -1)) {
@@ -281,5 +283,12 @@ class OutputFormatter implements WrappableOutputFormatterInterface
         }
 
         return implode("\n", $lines);
+    }
+
+    private function addLineBreaks(string $text, int $width): string
+    {
+        $encoding = mb_detect_encoding($text, null, true) ?: 'UTF-8';
+
+        return b($text)->toCodePointString($encoding)->wordwrap($width, "\n", true)->toByteString($encoding);
     }
 }
--- a/src/Symfony/Component/Console/Tests/Formatter/OutputFormatterTest.php
+++ b/src/Symfony/Component/Console/Tests/Formatter/OutputFormatterTest.php
@@ -367,10 +367,10 @@ EOF
         $formatter = new OutputFormatter(true);
 
         $this->assertSame("fo\no\e[37;41mb\e[39;49m\n\e[37;41mar\e[39;49m\nba\nz", $formatter->formatAndWrap('foo<error>bar</error> baz', 2));
-        $this->assertSame("pr\ne \e[37;41m\e[39;49m\n\e[37;41mfo\e[39;49m\n\e[37;41mo \e[39;49m\n\e[37;41mba\e[39;49m\n\e[37;41mr \e[39;49m\n\e[37;41mba\e[39;49m\n\e[37;41mz\e[39;49m \npo\nst", $formatter->formatAndWrap('pre <error>foo bar baz</error> post', 2));
+        $this->assertSame("pr\ne \e[37;41m\e[39;49m\n\e[37;41mfo\e[39;49m\n\e[37;41mo\e[39;49m\n\e[37;41mba\e[39;49m\n\e[37;41mr\e[39;49m\n\e[37;41mba\e[39;49m\n\e[37;41mz\e[39;49m \npo\nst", $formatter->formatAndWrap('pre <error>foo bar baz</error> post', 2));
         $this->assertSame("pre\e[37;41m\e[39;49m\n\e[37;41mfoo\e[39;49m\n\e[37;41mbar\e[39;49m\n\e[37;41mbaz\e[39;49m\npos\nt", $formatter->formatAndWrap('pre <error>foo bar baz</error> post', 3));
-        $this->assertSame("pre \e[37;41m\e[39;49m\n\e[37;41mfoo \e[39;49m\n\e[37;41mbar \e[39;49m\n\e[37;41mbaz\e[39;49m \npost", $formatter->formatAndWrap('pre <error>foo bar baz</error> post', 4));
-        $this->assertSame("pre \e[37;41mf\e[39;49m\n\e[37;41moo ba\e[39;49m\n\e[37;41mr baz\e[39;49m\npost", $formatter->formatAndWrap('pre <error>foo bar baz</error> post', 5));
+        $this->assertSame("pre \e[37;41m\e[39;49m\n\e[37;41mfoo\e[39;49m\n\e[37;41mbar\e[39;49m\n\e[37;41mbaz\e[39;49m \npost", $formatter->formatAndWrap('pre <error>foo bar baz</error> post', 4));
+        $this->assertSame("pre \e[37;41mf\e[39;49m\n\e[37;41moo\e[39;49m\n\e[37;41mbar\e[39;49m\n\e[37;41mbaz\e[39;49m p\nost", $formatter->formatAndWrap('pre <error>foo bar baz</error> post', 5));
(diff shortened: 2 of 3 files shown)


src/Symfony/Component/Console/Formatter/OutputFormatter.php
```diff
@@ -282,4 +282,27 @@ private function applyCurrentStyle(string $text, string $current, int $width, in
 
         return implode("\n", $lines);
     }
+
+    private function addLineBreaks($string, $charsPerLine)
+    {
+        $result = '';
+
+        if (\function_exists('mb_detect_encoding') && false !== $encoding = mb_detect_encoding($string, null, true)) {
```
derrabus: Because the Console component requires the mbstring Polyfill, we can be certain that `mb_detect_encoding()` is always defined.

src/Symfony/Component/Console/Formatter/OutputFormatter.php
```diff
@@ -282,4 +282,27 @@ private function applyCurrentStyle(string $text, string $current, int $width, in
 
         return implode("\n", $lines);
     }
+
+    private function addLineBreaks($string, $charsPerLine)
```
OskarStark: Can you add parameter types and a return type?

derrabus:
```suggestion
    private function addLineBreaks(string $string, int $charsPerLine): string
```

src/Symfony/Component/Console/Formatter/OutputFormatter.php
```diff
@@ -282,4 +282,29 @@ private function applyCurrentStyle(string $text, string $current, int $width, in
 
         return implode("\n", $lines);
     }
+
+    private function addLineBreaks(string $text, int $width): string
```
nicolas-grekas, line 288: just wondering: can't we use wordwrap method provided by the String component?

maxbeckers: Yes we could, but this would change the general behavior of the output a bit. Wordwrap tries to not break words and makes the linebreak before the word. Only if the word is longer than the length of the line there is the option to break the word.

```
u('test me in new line')->wordwrap(12, "\n", true); // result: "test me in \nnew line"
addLineBreaks('test me in new line', 12) // behavior as before ... result: "test me in n\new line"
u('thisIsLong')->wordwrap(6, "\n", true); // only this wouls wrap ... result: "thisIs\nLong"
```
I guess to keep the change / fix small we should not change the general behavior, What do you think @nicolas-grekas ?

nicolas-grekas: Would work for me.
We just need to detect the encoding with mb_detect_encoding and do 
`b($s)->toCodePointString($encoding)->wordwrap()->toByteString($encoding)`

maxbeckers: I'll try that and see where all the console tests will fail :)

src/Symfony/Component/Console/Tests/Helper/TableTest.php
```diff
@@ -118,30 +118,30 @@ public static function renderProvider()
...
                 'compact',
-<<<'TABLE'
-ISBN          Title                    Author           
-99921-58-10-7 Divine Comedy            Dante Alighieri  
-9971-5-0210-0 A Tale of Two Cities     Charles Dickens  
-960-425-059-0 The Lord of the Rings    J. R. R. Tolkien 
-80-902734-1-6 And Then There Were None Agatha Christie  
-
-TABLE
+                implode("\n", [
```
maxbeckers, line 121: This change is only because of the line "trim_trailing_whitespace = true" in .editorconfig.

src/Symfony/Component/Console/Tests/Helper/TableTest.php
```diff
@@ -118,30 +118,30 @@ public static function renderProvider()
...
-  ISBN            Title                      Author            
- =============== ========================== ================== 
-  99921-58-10-7   Divine Comedy              Dante Alighieri   
-  9971-5-0210-0   A Tale of Two Cities       Charles Dickens   
-  960-425-059-0   The Lord of the Rings      J. R. R. Tolkien  
-  80-902734-1-6   And Then There Were None   Agatha Christie   
- =============== ========================== ================== 
-
-TABLE
+                implode("\n", [
```
maxbeckers, line 134: This change is only because of the line "trim_trailing_whitespace = true" in .editorconfig.

src/Symfony/Component/Console/Tests/Formatter/OutputFormatterTest.php
```diff
@@ -367,21 +367,23 @@ public function testFormatAndWrap()
...
+        $this->assertSame("pr\ne \nfo\no\nba\nr\nba\nz \npo\nst", $formatter->formatAndWrap('pre <error>foo bar baz</error> post', 2));
         $this->assertSame("pre\nfoo\nbar\nbaz\npos\nt", $formatter->formatAndWrap('pre <error>foo bar baz</error> post', 3));
-        $this->assertSame("pre \nfoo \nbar \nbaz \npost", $formatter->formatAndWrap('pre <error>foo bar baz</error> post', 4));
-        $this->assertSame("pre f\noo ba\nr baz\npost", $formatter->formatAndWrap('pre <error>foo bar baz</error> post', 5));
+        $this->assertSame("pre \nfoo\nbar\nbaz \npost", $formatter->formatAndWrap('pre <error>foo bar baz</error> post', 4));
+        $this->assertSame("pre f\noo\nbar\nbaz p\nost", $formatter->formatAndWrap('pre <error>foo bar baz</error> post', 5));
+        $this->assertSame("Â rèälly\nlöng tîtlè\nthät cöüld\nnèêd\nmúltîplê\nlínès", $formatter->formatAndWrap('Â rèälly löng tîtlè thät cöüld nèêd múltîplê línès', 10));
+        $this->assertSame("Â rèälly\nlöng tîtlè\nthät cöüld\nnèêd\nmúltîplê\n línès", $formatter->formatAndWrap("Â rèälly löng tîtlè thät cöüld nèêd múltîplê\n línès", 10));
```
chalasr, line 386: Are all these changes on purpose and behave as expected?

maxbeckers: Yes, @nicolas-grekas (see [comment](https://github.com/symfony/symfony/pull/51223#discussion_r1282019479)) had the idea to use wordwrap from the string component for that. Wordwrap works a bit different with the line breaks than the regex before. now it breaks the line not in a word (or only if a word is longer than the full allowed length).
Example:
"Too long li\nne" => "Too long\nline"
IMHO this makes the output more readable.

nicolas-grekas (approved):

nicolas-grekas:
Thank you @maxbeckers.

PR #51227, opened 2023-08-02
[FrameworkBundle][Workflow] Attach the workflow's configuration to the `workflow` tag

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

Since the registry is deprecated, users have to build their own registry when a service locator is not enough.
However, some information can be missing, like the `supports` configuration option.

In this PR, I add the whole configuration to the tag, so everyone can build exactly what they need.

The config is added only to the `workflow` tag, not sub tags. To get it:
```php
class Kernel extends BaseKernel implements CompilerPassInterface
{
    use MicroKernelTrait;

    public function process(ContainerBuilder $container)
    {
        foreach ($container->findTaggedServiceIds('workflow.workflow') as $id => $attributes) {
            $config = $container->getDefinition($id)->getTag('workflow')[0];
            dd($config);
        }
    }
}
```

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
+++ b/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
@@ -12,6 +12,7 @@ CHANGELOG
  * Add `rate_limiter` tags to rate limiter services
  * Add `secrets:reveal` command
  * Add `rate_limiter` option to `http_client.default_options` and `http_client.scoped_clients`
+ * Attach the workflow's configuration to the `workflow` tag
 
 7.0
 ---
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
@@ -1029,7 +1029,7 @@ class FrameworkExtension extends Extension
             $workflowDefinition->replaceArgument(3, $name);
             $workflowDefinition->replaceArgument(4, $workflow['events_to_dispatch']);
 
-            $workflowDefinition->addTag('workflow', ['name' => $name]);
+            $workflowDefinition->addTag('workflow', ['name' => $name, 'metadata' => $workflow['metadata']]);
             if ('workflow' === $type) {
                 $workflowDefinition->addTag('workflow.workflow', ['name' => $name]);
             } elseif ('state_machine' === $type) {
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
@@ -84,7 +84,6 @@ use Symfony\Component\Validator\Validation;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
 use Symfony\Component\Webhook\Client\RequestParser;
 use Symfony\Component\Webhook\Controller\WebhookController;
-use Symfony\Component\Workflow;
 use Symfony\Component\Workflow\Exception\InvalidDefinitionException;
 use Symfony\Component\Workflow\Metadata\InMemoryMetadataStore;
 use Symfony\Component\Workflow\WorkflowEvents;
@@ -302,7 +301,15 @@ abstract class FrameworkExtensionTestCase extends TestCase
         $this->assertArrayHasKey('index_4', $args);
         $this->assertNull($args['index_4'], 'Workflows has eventsToDispatch=null');
 
-        $this->assertSame(['workflow' => [['name' => 'article']], 'workflow.workflow' => [['name' => 'article']]], $container->getDefinition('workflow.article')->getTags());
+        $tags = $container->getDefinition('workflow.article')->getTags();
+        $this->assertArrayHasKey('workflow', $tags);
+        $this->assertArrayHasKey('workflow.workflow', $tags);
+        $this->assertSame([['name' => 'article']], $tags['workflow.workflow']);
+        $this->assertSame('article', $tags['workflow'][0]['name'] ?? null);
+        $this->assertSame([
+            'title' => 'article workflow',
+            'description' => 'workflow for articles',
+        ], $tags['workflow'][0]['metadata'] ?? null);
 
         $this->assertTrue($container->hasDefinition('workflow.article.definition'), 'Workflow definition is registered as a service');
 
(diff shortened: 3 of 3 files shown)


nicolas-grekas:
That's a bit strange to me, I'm not sure we ever did that. Usually tag attributes are used in a compiler pass. Shouldn't we do the same here? What's the more extended use case for this?

lyrixx:
> That's a bit strange to me, I'm not sure we ever did that. Usually tag attributes are used in a compiler pass. Shouldn't we do the same here? What's the more extended use case for this?

It looks like you missed the PR description :D The example show a usage of tag attribute in a compiler pass 

And the use case is described in the PR description too. But to go a bit further

```php
class Kernel extends BaseKernel implements CompilerPassInterface
{
    use MicroKernelTrait;

    public function process(ContainerBuilder $container)
    {
        $classesToWorkflow = [];
        foreach ($container->findTaggedServiceIds('workflow.workflow') as $id => $attributes) {
            $config = $container->getDefinition($id)->getTag('workflow')[0];
            foreach ($config['supports'] ?? [] as $class) {
                $classesToWorkflow[$class] = new Reference($id);
            }
        }

        $container->register(MyRegistry::class, MyRegistry::class)
            ->setArguments([$classesToWorkflow])
        ;
    }
}
```

> **Note**
> 1. I did not tested this very example
> 2. Instead of an array of service, it would be better to use a service locator, but that's not the point here

nicolas-grekas:
Doesn't your registry duplicate the code we have in the `Registry` class? Shouldn't we revert making it internal instead?
I'm challenging this because it feel strange to me to use tags for what you do here (I'm not sure we have any similar example).

lyrixx:
I have a big issue with the registry: People use it to get one workflow. In the documentation we used to tell to do : 
```php
$workflowRegistry->get($subject)->apply($subject, 'foobar');
````

And this is wrong: people must inject the right Workflow directly (law of demeter, performance, etc)

That's why I decided to make it internal initially.

The documentation has been updated since (I just checked, and there is still [one occurrence](https://github.com/symfony/symfony-docs/blob/d476a49a670f19d683433e51489a125b62d60ff3/workflow.rst#L1086), cc @alexandre-daubois) 

But there is still one issue. If you have generic controller / service where you cannot know in advance which workflow to use, you have to use a ServiceLocator. There are all tagged, by name, so it should be easy...

BUT, sometime, (cf [this discussion](https://github.com/symfony/symfony/discussions/49199)) the name is not enough. 

With this PR, people will have full power to build their own registry. It'll be faster, cleaner, and build exactly for each use case.

Even with the current Registry and its [`WorkflowSupportStrategyInterface`](https://github.com/symfony/workflow/blob/6.3/SupportStrategy/WorkflowSupportStrategyInterface.php) you can't build something totally dynamic, since you miss some information about the configuration

**Review by nicolas-grekas — APPROVED** [review]
OK, thanks for the explanations. I guess I would prefer wiring the registry (ours) in a compiler pass instead of doing it in FWB, but that can be done separately.

**Comment by fabpot** [comment]
Mouais, I'm not convinced.
The docs tell people the best way.
The Registry is the entry point for more powerful features.
If people want to shoot themselves in the foot, that's life.
So, I don't see why we need to have yet another way. I understand that one is userland and the other is internal, but still. If we're talking about edge cases, just using the current Registry should be enough, right?

**Comment by lyrixx** [comment]
I disagree. Registry needs instantiated workflow to work. It means no lazy loading. 

Then you can't access `supports` configuration option.

That's two pain points that show the current way to work doesn't work well.

**Comment by lyrixx** [comment]
Rebased, and conflict fixed

**Comment by ro0NL** [comment]
Maybe revert deprecation the registry? I see no easy alternative for getting a workflow from dynamic object, aka `Registry::get()`

**Comment by ro0NL** [comment]
what's the status of workflow registry for 6.4?

**Comment by lyrixx** [comment]
Hello, This PR is ready and I have rebased it. 

@fabpot I understand your concern. But I have a real use case where I need to fetch a workflow dynamically based on some entity properties. We'll have a LOT of workflows - basically one by client (B to B to C), and every `Client` lays on the same entity. I could use the registry, but there is no lazy loading, so it will degrade performance.

Finally, the diff is very minimal, and does not really impact the framework. Everything is thrown after the compilation. I don't really how it could hurt 🤓

**Comment by stof** [comment]
The issue is that if we start exposing this config on the attribute with the intent of allowing projects to read it, we must provide BC on the structure of this config (with no way at all to ever deprecate anything in it as it is not accessed through a dedicated API in which we can trigger deprecations)

**Comment by lyrixx** [comment]
Hmm, Got it 👍🏼 We could document that the shape is not BC?

**Comment by stof** [comment]
@lyrixx but then, how would you use it ?

**Comment by lyrixx** [comment]
Basically with something like this:

```yaml
# config/packages/workflow.yaml
framework:
    workflows:
        my_workflow_name:
            metadata:
                features: [repair, wash, dry]
```

```php
// src/Kernel.php
class Kernel extends BaseKernel implements CompilerPassInterface
{
    use MicroKernelTrait;

    public function process(ContainerBuilder $container)
    {
        $configurations = [];
        foreach ($container->findTaggedServiceIds('workflow') as $id => $config) {
            $configurations[$id] = $config[0]['config'];
        }

        $container
            ->getDefinition(Selector::class)
            ->setArgument('$configurations', $configurations)
        ;
    }
}
```

```php
<?php
// src/Workflow/Selector.php
namespace App\Workflow;

use Psr\Container\ContainerInterface;
use Symfony\Component\DependencyInjection\Attribute\TaggedLocator;

class Selector
{
    public function __construct(
        #[TaggedLocator('workflow')]
        private readonly ContainerInterface $container,
        private readonly array $configurations,
    ) {
    }

    public function getWorkflow(/** Article $article */)
    {
        foreach ($this->configurations as $id => $config) {
            // $features = $article->getWorkflowFeatures();
            $featuresRequired = ['repair', 'wash', 'dry'];
            sort($featuresRequired);
            $featuresAvailable = $config['metadata']['features'] ?? [];
            sort($featuresAvailable);
            if ($featuresRequired === $featuresAvailable) {
                return $this->container->get($id);
            }
        }

        throw new \RuntimeException('No workflow found');
    }
}
```

and finally


```php
    public function index(Selector $selector, Article $article): Response
    {
        $w = $selector->getWorkflow($article);
        dd($w);
```

lyrixx:
Would you accept this PR if I inject in the tag **only** the workflow metadata? It'll be enough.

lyrixx:
> Would you accept this PR if I inject in the tag **only** the workflow metadata? It'll be enough.

cc @fabpot

lyrixx:
PR rebased, and updated

PR #51229, opened 2023-08-02
[MonologBridge] Deprecate `Logger` class in favor of `Symfony\Bridge\Monolog\Monolog`

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | kind of
| Deprecations? | yes
| Tickets       |
| License       | MIT
| Doc PR        |

---

For the record, I imported the new Monolog class in Monolog, and I ran the test suite and it was okay 👍🏼 all green except one failure, but it's a [monolog issue](https://github.com/Seldaek/monolog/pull/1825) I guess

### DIFF
--- a/src/Symfony/Bridge/Monolog/CHANGELOG.md
+++ b/src/Symfony/Bridge/Monolog/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * Add native return type to `Logger::clear()` and to `DebugProcessor::clear()`
+ * Deprecate `Logger` class in favor of `Symfony\Bridge\Monolog\Monolog`
 
 6.1
 ---
--- a/src/Symfony/Bridge/Monolog/Monolog.php
+++ b/src/Symfony/Bridge/Monolog/Monolog.php
@@ -1,0 +1,403 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bridge\Monolog;
+
+use Monolog\DateTimeImmutable;
+use Monolog\Handler\HandlerInterface;
+use Monolog\Level;
+use Monolog\LogRecord;
+use Monolog\Logger as BaseLogger;
+use Monolog\Logger;
+use Monolog\Processor\ProcessorInterface;
+use Monolog\ResettableInterface;
+use Psr\Log\LoggerInterface;
+use Psr\Log\LoggerTrait;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Log\DebugLoggerInterface;
+use Symfony\Contracts\Service\ResetInterface;
+
+/**
+ * Logger implements a PSR-3 logger with psr/log and forwards logs to a Monolog
+ * instance.
+ *
+ * I has the same features a Monolog, and can be used as a drop in replacement.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ * @author Nicolas Grekas <p@tchwork.com>
+ * @author Grégoire Pineau <lyrixx@lyrixx.info>
+ */
+final class Monolog implements LoggerInterface, DebugLoggerInterface, ResettableInterface, ResetInterface
+{
+    use LoggerTrait;
+
+    /**
+     * Mapping between levels numbers defined in RFC 5424 and Monolog ones.
+     *
+     * @phpstan-var array<int, Level> $rfc_5424_levels
+     */
+    private const RFC_5424_LEVELS = [
+        7 => Level::Debug,
+        6 => Level::Info,
+        5 => Level::Notice,
+        4 => Level::Warning,
+        3 => Level::Error,
+        2 => Level::Critical,
+        1 => Level::Alert,
+        0 => Level::Emergency,
+    ];
+
+    private BaseLogger $monolog;
+
+    /**
+     * @param string             $name       The logging channel, a simple descriptive name that is attached to all log records
+     * @param HandlerInterface[] $handlers   optional stack of handlers, the first one in the array is called first, etc
+     * @param callable[]         $processors Optional array of processors
+     * @param \DateTimeZone|null $timezone   Optional timezone, if not provided date_default_timezone_get() will be used
+     *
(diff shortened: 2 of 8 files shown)


src/Symfony/Bridge/Monolog/Monolog.php
```diff
@@ -0,0 +1,500 @@
...
+     * member of new records.
+     *
+     * As of PHP7.1 microseconds are always included by the engine, so
+     * there is no performance penalty and Monolog 2 enabled microseconds
+     * by default. This function lets you disable them though in case you want
+     * to suppress microseconds from the output.
+     *
+     * @param bool $micro True to use microtime() to create timestamps
+     */
+    public function useMicrosecondTimestamps(bool $micro): BaseLogger
```
nicolas-grekas: let's return $this; instead everywhere we return BaseLogger currently, there's no need to keep returning BaseLogger if we have a new class.

src/Symfony/Bridge/Monolog/Monolog.php
```diff
@@ -0,0 +1,500 @@
...
+    public function getTimezone(): \DateTimeZone
+    {
+        return $this->monolog->getTimezone();
+    }
+
+    /**
+     * Delegates exception management to the custom exception handler,
+     * or throws the exception if no custom handler is set.
+     */
+    protected function handleException(\Throwable $e, LogRecord $record): void
```
nicolas-grekas: let's remove this one

src/Symfony/Bridge/Monolog/Monolog.php
```diff
@@ -0,0 +1,500 @@
...
+        return $this->monolog->getExceptionHandler();
+    }
+
+    /**
+     * Adds a log record at an arbitrary level.
+     *
+     * This method allows for compatibility with common interfaces.
+     *
+     * @param mixed             $level   The log level (a Monolog, PSR-3 or RFC 5424 level)
+     * @param string|Stringable $message The log message
```
nicolas-grekas:
```suggestion
     * @param string|\Stringable $message The log message
```

src/Symfony/Bridge/Monolog/Monolog.php
```diff
@@ -0,0 +1,500 @@
...
+     *
+     * This still returns a string instead of a Level for BC, but new code should not rely on this method.
+     *
+     * @throws \Psr\Log\InvalidArgumentException If level is not defined
+     *
+     * @phpstan-param  value-of<Level::VALUES>|Level $level
+     *
+     * @phpstan-return value-of<Level::NAMES>
+     *
+     * @deprecated Since 3.0, use {@see toMonologLevel} or {@see \Monolog\Level->getName()} instead
```
nicolas-grekas: let's remove this method?

src/Symfony/Bridge/Monolog/Monolog.php
```diff
@@ -0,0 +1,500 @@
...
+    /**
+     * Converts PSR-3 levels to Monolog ones if necessary.
+     *
+     * @param int|string|Level|LogLevel::* $level Level number (monolog) or name (PSR-3)
+     *
+     * @throws \Psr\Log\InvalidArgumentException If level is not defined
+     *
+     * @phpstan-param value-of<Level::VALUES>|value-of<Level::NAMES>|Level|LogLevel::* $level
+     */
+    public static function toMonologLevel(string|int|Level $level): Level
```
nicolas-grekas: drop?

src/Symfony/Bridge/Monolog/Monolog.php
```diff
@@ -0,0 +1,500 @@
...
+        if (!$level instanceof Level) {
+            if (!\is_string($level) && !\is_int($level)) {
+                throw new \InvalidArgumentException('$level is expected to be a string, int or '.Level::class.' instance');
+            }
+
+            if (isset(self::RFC_5424_LEVELS[$level])) {
+                $level = self::RFC_5424_LEVELS[$level];
+            }
+
+            $level = static::toMonologLevel($level);
```
nicolas-grekas:
```suggestion
            $level = BaseLogger::toMonologLevel($level);
```

src/Symfony/Bridge/Monolog/Monolog.php
```diff
@@ -0,0 +1,500 @@
...
+     */
+    public const EMERGENCY = 600;
+
+    /**
+     * Monolog API version.
+     *
+     * This is only bumped when API breaks are done and should
+     * follow the major version of the library
+     */
+    public const API = 3;
```
nicolas-grekas: let's remove this const and all the ones before?

lyrixx:
Thanks for the review, I pushed a new version

src/Symfony/Bridge/Monolog/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * Add native return type to `Logger::clear()` and to `DebugProcessor::clear()`
+ * Deprecate `Logger` class in favor of `Symfony\Bridge\Monolog\Monolog`
```
stof, line 8: to me, this new class should not be introduced either.

the removal of the debug loggers should either be implemented using a service configurator or we should make our debug processor itself able to behave like a no-op for environments where we don't want to collect those debug logs.

And the DebugLoggerInterface should be solved by injecting the DebugHandler in the collector instead of the main logger instance (which would be perfectly supported but is not wired this way in MonologBundle).

nicolas-grekas: Let's first discuss about removing the `@final` in https://github.com/Seldaek/monolog/pull/1827

Anyone up to do the PR implementing what @stof describes?

Seldaek: Sorry for being blunt but this looks like a mess to me. You'd get a Monolog class which you cannot even pass to something expecting a Logger instance (which you would right now do if you want to call `reset()` or smth on it.

Why not inject a DebugHandler in the Logger at runtime when in development for example?

Or worst case if you must extend Logger then go ahead and extend it, I just marked it final because people were extending it and doing crazy shit which then broke with new features. If you know what you are doing it is IMO safe to extend, and it's why it is just a `@final` and not a proper `final` keyword. I don't want to prevent people, but I want to make it clear that it voids the warranty if you do extend.

Maybe it'd be worth layout out what exactly you are trying to solve here, because I am not 100% sure I am up to speed. I'm happy to try and help find a solution, but removing `@final` is a no go for me right now.

nicolas-grekas:
Closing in favor of #51284, thanks for pushing this forward.

Issue #41892, opened 2021-06-29
`json_login_ldap` in chain with `json_login` and new authenticator


oliverreese:
**Symfony version(s) affected**: 5.3.2

**Description**
a combination of json_login_ldap and json_login will not work when the new authentication manager is enabled. They both
on their own work properly but not in combination. When new authentication manager is disabled they work properly together
in a chain.

**How to reproduce**
see attached repository with 4 different configuration cases to reproduce: 
https://github.com/oliverreese/symfony-security-ldap-login

**Possible Solution**

**Additional context**

OskarStark:
cc @wouterj

raphaelriehl:
Hi, maybe it's another issue, but I encounter same thing with `form_login` and `form_login_ldap`.
When I disable the new authentication manager it's works.
Same problem describe [here](https://stackoverflow.com/questions/68131217/how-to-use-the-ldap-form-login-and-form-login-on-symfony-5-3-with-the-new-authen).

diversantvlz:
[here](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/LdapFactoryTrait.php#L42) when the parent method is called, the base authenticator is replaced another with a different config. Parent method have hardcoded service id

B4rb4ross4:
I can reproduce this issue aswell. I have the following firewall config:

```
main:
            lazy: true
            provider: app_user_provider
            user_checker: Some\Namespace\Infrastructure\Security\UnconfirmedUserChecker
            entry_point: 'form_login_ldap'
            form_login:
                login_path: login
                check_path: login
            form_login_ldap:
                service: Symfony\Component\Ldap\Ldap
                dn_string: 'MyOrganization\{username}'
                login_path: login
                check_path: ldap_login_check
            logout:
                path: logout
```

Then you can see that the wrong options are injected into the "normal" form login authenticator.

```
array (
  'username_parameter' => '_username',
  'password_parameter' => '_password',
  'check_path' => 'ldap_login_check',
  'post_only' => true,
  'form_only' => false,
  'enable_csrf' => false,
  'csrf_parameter' => '_csrf_token',
  'csrf_token_id' => 'authenticate',
  'login_path' => 'login',
  'use_forward' => false,
  'require_previous_session' => false,
)
```

franzwilding:
[...]

PR #51231, opened 2023-08-02
[Ldap] Allow to use ldap in a chain provider

l-vo:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | #41892
| License       | MIT
| Doc PR        | 

This PR allows to fix that the `LdapUserProvider` that not work properly when used in the `ChainProvider`.

This fix relies on two ideas:
- The `LdapBadge` is not a badge that must be resolved, it's only used to carry Ldap data
- Allow to prevent `CheckLdapCredentialsListener` to check against the ldap directory the password for a non-ldap user using a new `#[WithLdapPassword]` attribute

### DIFF
--- a/src/Symfony/Component/Ldap/Security/CheckLdapCredentialsListener.php
+++ b/src/Symfony/Component/Ldap/Security/CheckLdapCredentialsListener.php
@@ -13,11 +13,13 @@ namespace Symfony\Component\Ldap\Security;
 
 use Psr\Container\ContainerInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Ldap\Attribute\WithLdapPassword;
 use Symfony\Component\Ldap\Exception\InvalidCredentialsException;
 use Symfony\Component\Ldap\Exception\InvalidSearchCredentialsException;
 use Symfony\Component\Ldap\LdapInterface;
 use Symfony\Component\Security\Core\Exception\BadCredentialsException;
 use Symfony\Component\Security\Core\Exception\LogicException;
+use Symfony\Component\Security\Core\User\UserInterface;
 use Symfony\Component\Security\Http\Authenticator\Passport\Credentials\PasswordCredentials;
 use Symfony\Component\Security\Http\Event\CheckPassportEvent;
 
@@ -48,9 +50,6 @@ class CheckLdapCredentialsListener implements EventSubscriberInterface
 
         /** @var LdapBadge $ldapBadge */
         $ldapBadge = $passport->getBadge(LdapBadge::class);
-        if ($ldapBadge->isResolved()) {
-            return;
-        }
 
         if (!$passport->hasBadge(PasswordCredentials::class)) {
             throw new \LogicException(sprintf('LDAP authentication requires a passport containing password credentials, authenticator "%s" does not fulfill these requirements.', $event->getAuthenticator()::class));
@@ -72,6 +71,16 @@ class CheckLdapCredentialsListener implements EventSubscriberInterface
         }
 
         $user = $passport->getUser();
+        $nonLdapUserWithoutLdapPasswordAttribute = false;
+        if (!$user instanceof LdapUser) {
+            $reflectionClass = new \ReflectionClass($user);
+            $attr = $reflectionClass->getAttributes(WithLdapPassword::class);
+
+            $nonLdapUserWithoutLdapPasswordAttribute = count($attr) === 0;
+            if (!$nonLdapUserWithoutLdapPasswordAttribute && !$attr[0]->newInstance()->enabled) {
+                return;
+            }
+        }
 
         /** @var LdapInterface $ldap */
         $ldap = $this->ldapLocator->get($ldapBadge->getLdapServiceId());
@@ -104,8 +113,11 @@ class CheckLdapCredentialsListener implements EventSubscriberInterface
             throw new BadCredentialsException('The presented password is invalid.');
         }
 
+        if ($nonLdapUserWithoutLdapPasswordAttribute) {
+            trigger_deprecation('symfony/ldap', '6.4', 'Authenticate a user that is not an instance of %s is deprecated and won\'t be the default behavior anymore in 7.0. Use the %s attribute to keep this behavior.', LdapUser::class, WithLdapPassword::class);
+        }
+
         $passwordCredentials->markResolved();
-        $ldapBadge->markResolved();
     }
 
     public static function getSubscribedEvents(): array
--- a/src/Symfony/Component/Ldap/Security/LdapBadge.php
+++ b/src/Symfony/Component/Ldap/Security/LdapBadge.php
@@ -24,14 +24,14 @@ use Symfony\Component\Security\Http\Authenticator\Passport\Badge\BadgeInterface;
  */
 class LdapBadge implements BadgeInterface
 {
-    private bool $resolved = false;
+    private bool $resolved = true;
     private string $ldapServiceId;
     private string $dnString;
     private string $searchDn;
     private string $searchPassword;
     private ?string $queryString;
 
-    public function __construct(string $ldapServiceId, string $dnString = '{user_identifier}', string $searchDn = '', string $searchPassword = '', string $queryString = null)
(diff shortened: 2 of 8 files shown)


src/Symfony/Component/HttpFoundation/IpUtils.php
```diff
@@ -21,6 +21,8 @@ class IpUtils
     private static $checkedIps = [];
 
     /**
+     * @psalm-suppress UnusedConstructor
```
nicolas-grekas: please remove this, it's unrelated

src/Symfony/Component/Ldap/Security/LdapBadge.php
```diff
@@ -65,8 +69,13 @@ public function getQueryString(): ?string
         return $this->queryString;
     }
 
+    /**
+     * @psalm-suppress PossiblyUnusedMethod
```
nicolas-grekas: drop, we don't use such annotations

src/Symfony/Component/Ldap/Security/LdapBadge.php
```diff
@@ -24,20 +24,24 @@
...
+    public function __construct(string $ldapServiceId, string $dnString = '{user_identifier}', string $searchDn = '', string $searchPassword = '', string $queryString = null, $resolved = false)
     {
         $this->ldapServiceId = $ldapServiceId;
         $this->dnString = $dnString;
         $this->searchDn = $searchDn;
         $this->searchPassword = $searchPassword;
         $this->queryString = $queryString;
+        $this->resolved = $resolved;
+        if (false === $this->resolved) {
+            trigger_deprecation('symfony/ldap', '5.4', 'Passing false as resolved initial value is deprecated, use true instead.');
```
nicolas-grekas: these deprecations make me think this PR should target 6.4 instead - it's a new feature / an improvement

src/Symfony/Component/Ldap/Security/LdapBadge.php
```diff
@@ -46,6 +46,10 @@ public function __construct(string $ldapServiceId, string $dnString = '{user_ide
             trigger_deprecation('symfony/ldap', '6.2', 'Using "{username}" parameter in LDAP configuration is deprecated, consider using "{user_identifier}" instead.');
         }
         $this->queryString = $queryString;
+        $this->resolved = $resolved;
+        if (false === $this->resolved) {
+            trigger_deprecation('symfony/ldap', '6.4', 'Passing false as resolved initial value is deprecated, use true instead.');
```
OskarStark:
```suggestion
            trigger_deprecation('symfony/ldap', '6.4', 'Passing "false" as resolved initial value is deprecated, use "true" instead.');
```

src/Symfony/Component/Ldap/Security/LdapBadge.php
```diff
@@ -75,6 +79,8 @@ public function getQueryString(): ?string
 
     public function markResolved(): void
     {
+        trigger_deprecation('symfony/ldap', '6.4', 'Calling %s is deprecated.', __METHOD__);
```
OskarStark: Should we add more explanation, than just "... is deprecated." Maybe sth. like:
"... is deprecated and not used for the LdapBadge anymore because of foo?"

src/Symfony/Component/Ldap/Security/CheckLdapCredentialsListener.php
```diff
@@ -72,6 +69,9 @@ public function onCheckPassport(CheckPassportEvent $event)
         }
 
         $user = $passport->getUser();
+        if (!$user instanceof LdapUser) {
+            return;
```
vtsykun: > The CheckLdapCredentialsListener does nothing when the user is not an instance of LdapUser.

This is major BC in the logic. The user can be loaded from the database, but at the same time the password is checked in LDAP.

in my opinion, it's more logical to check LDAP in the first, and if the check is failed fallback to default credentials checker.
For example my [CheckLdapCredentialsListener](https://github.com/vtsykun/packeton/blob/2.3/src/Security/CheckLdapCredentialsListener.php) implementation.

src/Symfony/Component/Ldap/Security/LdapBadge.php
l-vo: Updated, thank you

src/Symfony/Component/Ldap/Security/CheckLdapCredentialsListener.php
l-vo: I don't know if the  BC is "major" or not, but checking LDAP first and fallback on another credential system increases the attack surface IMHO. But you're right about the BC, I added a BC layer.

chalasr:
This looks like a feature

l-vo:
@chalasr I don't know if it's really a feature, it worked with the Guard system and the feature has seamlessly disappeared with the replacement of Guard by Authenticators (and there was no announcement that the feature was not supported with Authenticators).

src/Symfony/Component/Ldap/Tests/Security/CheckLdapCredentialsListenerTest.php
```diff
@@ -171,6 +196,9 @@ public static function queryForDnProvider(): iterable
         yield ['{user_identifier}', '{user_identifier}_test'];
     }
 
+    /**
+     * @group legacy
```
stof, line 200: why is this becoming a legacy tests ? Why would we remove this test when removing deprecated features related to this PR ?

stof (commented):
I'm not sure the configuration described in the linked issue (using the same check_path for `form_login` and `form_login_ldap`) actually makes sense (and so whether this solution is valid as solution for that use case anyway)

nicolas-grekas:
Moving to milestone 8.1, rebase needed if we want to make this happen.
That's a new feature indeed.
I'm not deep enough into the topic to have an opinion on the PR itself sorry.

nicolas-grekas:
Thank you for this @l-vo, and sorry it sat for so long. The problem is real and your analysis of it is right: `CheckLdapCredentialsListener` binds for every passport carrying an `LdapBadge`, so a user that exists only in the database side of a chain provider gets its local password sent to the directory and can never log in.

I have opened #64976, which reaches the same goal through an `ldap_users_only` option on the `*_ldap` firewall factories rather than an attribute on the user class. Closing here in its favour, and the reasoning is worth writing down since it is not a matter of taste.

Three things pushed me away from this shape:

* dropping `LdapBadge::markResolved()` removes the fail-closed mechanism. `AuthenticatorManager` refuses a login whose badges are not all resolved, and that is what currently makes a missing `CheckLdapCredentialsListener` reject the request. Without it, the same misconfiguration authenticates the user from `LdapUser::getPassword()`, which is the value copied out of `password_attribute`, with no bind at all. #64976 keeps marking the badge resolved on the skipped path.
* the `$resolved` constructor argument on `LdapBadge` breaks authenticators that build their own badge, which the class docblock actively recommends. With this branch applied, `new LdapBadge("app.ldap")` yields an unresolved badge that nothing resolves any more, so the bind succeeds and the login is still refused.
* the 9.0 default flip is the one I could not sign off. `form_login_ldap` with a Doctrine entity provider never involves `LdapUser`, and it is a common setup. After the flip those applications silently stop checking the directory and start trusting whatever `getPassword()` returns: a stale local hash keeps working, or an empty one breaks every login. Both outcomes are silent. The new option defaults to preserving today's behaviour, so there is nothing to flip and nothing to deprecate.

Two smaller things for the record. `ChainProviderWithLdapTest` never ran: it uses an `@dataProvider` annotation with a non-static provider, which errors under the PHPUnit this repo now uses before reaching a single assertion, so its green CI meant nothing. And the test belongs in `src/Symfony/Component/Ldap/Tests/`, which would have made the `symfony/security-core` composer.json change unnecessary, since the Ldap component already has both packages in require-dev.

@stof was also right that #41892 is largely about something else: `LdapFactoryTrait` letting the parent factory register the decorated authenticator under the id its non-LDAP variant uses, so `form_login` and `form_login_ldap` cannot coexist on one firewall. I confirmed it compiles to a single service carrying whichever check_path was written last, and that is #64975, targeting 6.4.

PR #51233, opened 2023-08-02
[HttpKernel] Add validation groups resolver option to `RequestPayloadValueResolver`

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #51012
| License       | MIT
| Doc PR        | -

This PR introduces a way to configure validation groups for the `RequestPayloadValueResolver` dynamically, similar to how it can be done in the [form component](https://symfony.com/doc/current/form/validation_group_service_resolver.html):

```php
class ValidationGroupsResolver
{
    public function __invoke(mixed $payload, Request $request)
    {
        // ...
        return $groups;
    }
}

public function index(
    #[MapRequestPayload(validationGroupsResolver: new ValidationGroupsResolver())] DTO $dto,
): Response {
    // ...
}
```

The resolver can also be a service, it just needs to be tagged with `controller.request_payload.validation_groups_resolver`:

```php
#[AutoconfigureTag('controller.request_payload.validation_groups_resolver')]
class ValidationGroupsResolver
{
    public function __construct(SomeDep $someDep)
    {
    }

    public function __invoke(mixed $payload, Request $request)
    {
        // ...
        return $groups;
    }
}

public function index(
    #[MapRequestPayload(validationGroupsResolver: ValidationGroupsResolver::class)] DTO $dto,
): Response {
    // ...
}
```

### DIFF
--- a/src/Symfony/Component/HttpKernel/Attribute/MapQueryString.php
+++ b/src/Symfony/Component/HttpKernel/Attribute/MapQueryString.php
@@ -25,13 +25,17 @@ use Symfony\Component\Validator\Constraints\GroupSequence;
 class MapQueryString extends ValueResolver
 {
     public ArgumentMetadata $metadata;
+    public readonly string|\Closure|null $validationGroupsResolver;
 
     public function __construct(
         public readonly array $serializationContext = [],
         public readonly string|GroupSequence|array|null $validationGroups = null,
         string $resolver = RequestPayloadValueResolver::class,
         public readonly int $validationFailedStatusCode = Response::HTTP_NOT_FOUND,
+        string|callable $validationGroupsResolver = null,
     ) {
         parent::__construct($resolver);
+
+        $this->validationGroupsResolver = \is_callable($validationGroupsResolver) ? $validationGroupsResolver(...) : $validationGroupsResolver;
     }
 }
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/UnusedTagsPass.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/UnusedTagsPass.php
@@ -47,6 +47,7 @@ class UnusedTagsPass implements CompilerPassInterface
         'container.service_subscriber',
         'container.stack',
         'controller.argument_value_resolver',
+        'controller.request_payload.validation_groups_resolver',
         'controller.service_arguments',
         'controller.targeted_value_resolver',
         'data_collector',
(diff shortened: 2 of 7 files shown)


src/Symfony/Component/HttpKernel/Attribute/MapQueryString.php
```diff
@@ -25,13 +25,17 @@
...
 {
     public ArgumentMetadata $metadata;
+    public readonly string|\Closure|null $validationGroupsResolver;
 
     public function __construct(
         public readonly array $serializationContext = [],
         public readonly string|GroupSequence|array|null $validationGroups = null,
         string $resolver = RequestPayloadValueResolver::class,
         public readonly int $validationFailedStatusCode = Response::HTTP_NOT_FOUND,
+        string|callable $validationGroupsResolver = null,
```
nicolas-grekas, line 35: What about supporting only service ids?

nicolas-grekas (commented):
Would you have a real-life example use case? @MilanKaranovicQuantox  maybe?

src/Symfony/Component/HttpKernel/Attribute/MapQueryString.php
HeahDude: What about an array of validation groups to use instead of a resolver (less boilerplate for simple cases)?
We could also have both options.
Another way could be to have a specific interface to implement on the value object directly.

yceruto: Determining the validation groups dynamically based on the state of the underlying object is already supported by 
`#[Assert\GroupSequenceProvider]` https://symfony.com/doc/current/validation/sequence_provider.html#group-sequence-providers

I guess the use case the author wants to cover here is determining the groups based on external services through DI (which is not so common but still)

milanqtx:
> Would you have a real-life example use case? @MilanKaranovicQuantox maybe?

Hi @nicolas-grekas! I tend to replace Forms with `MapRequestPayload` on the API endpoints. I want to use Forms only on FE, for actual forms.

I managed to solve my issue as it's suggested in [the discussion](https://github.com/symfony/symfony/discussions/51012#discussioncomment-6628307) but I think it would be good to be able to use the service as a group resolver if needed, like Forms are able to.

When I posted the discussion, I had this in mind as potential solution https://api-platform.com/docs/core/validation/#dynamic-validation-groups

nicolas-grekas:
@yceruto WDYT? Is this worth the extra bit of complexity? should we improve the doc instead? or is this desired to you?

yceruto:
This example from https://api-platform.com/docs/core/validation/#dynamic-validation-groups shows a valid use case, in my opinion. However, I would prefer for this to be implemented as a validator feature instead, exactly as it's documented here: https://api-platform.com/docs/core/validation/#sequential-validation-groups. This way, it won't be specific to the ValueResolver.

yceruto:
Here we go with another proposal, https://github.com/symfony/symfony/pull/51348, but this time from the Validator component instead, covering other advanced scenarios.

fabpot:
Closing in favor of #51348

PR #51240, opened 2023-08-03
[MonologBridge][SecurityBundle] Remove calls to `TokenInterface::isAuthenticated()`

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

`isAuthenticated()` has been removed from `TokenInterface` in 6.0. Both packages I've changed here (MonologBridge and SecurityBundle) announce that they're incompatible with SecurityCore 5.4. This is why I believe that the compatibility code I'm removing is obsolete.

Nevertheless, a token might still implement that method and we'd call it currently despite its removal from the interface (duck typing). I have targeted 6.4, but if we consider the removal of these calls a BC break, we could merge this PR to 7.0 instead.

### DIFF
--- a/src/Symfony/Bridge/Monolog/Processor/AbstractTokenProcessor.php
+++ b/src/Symfony/Bridge/Monolog/Processor/AbstractTokenProcessor.php
@@ -47,7 +47,7 @@ abstract class AbstractTokenProcessor
 
         if (null !== $token = $this->getToken()) {
             $record['extra'][$this->getKey()] = [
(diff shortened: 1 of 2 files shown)


nicolas-grekas (approved):

nicolas-grekas:
Thank you @derrabus.

PR #51241, opened 2023-08-03
[FrameworkBundle] Remove obsolete feature detection from cache warmers

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

The features that those cache warmer access have existed since Symfony 4 or earlier. Let's remove the detection logic.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/SerializerCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/SerializerCacheWarmer.php
@@ -41,8 +41,8 @@ class SerializerCacheWarmer extends AbstractPhpFileCacheWarmer
 
     protected function doWarmUp(string $cacheDir, ArrayAdapter $arrayAdapter): bool
     {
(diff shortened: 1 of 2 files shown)


src/Symfony/Bundle/FrameworkBundle/CacheWarmer/SerializerCacheWarmer.php
```diff
@@ -41,8 +41,8 @@ public function __construct(array $loaders, string $phpArrayFile)
 
     protected function doWarmUp(string $cacheDir, ArrayAdapter $arrayAdapter): bool
     {
-        if (!class_exists(CacheClassMetadataFactory::class) || !method_exists(XmlFileLoader::class, 'getMappedClasses') || !method_exists(YamlFileLoader::class, 'getMappedClasses')) {
-            return false;
+        if (!$this->loaders) {
```
derrabus, line 44: In theory, you could've created an instance of this cache warmer with an empty loaders array without the serializer being installed and the feature detection would've prevented it from blowing up.

Since this warmer doesn't do anything with an empty loaders array anyway, I use this check for an early exit instead.

nicolas-grekas:
Thank you @derrabus.

PR #49639, opened 2023-03-08
[DependencyInjection] Add support for generating lazy closures

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

On my mission to improve support for closures in DI, this allows generating lazy closures.

What is a lazy closure?

Right now, when we generate a closure, we generate something like:
```php
return ($container->privates['foo'] ?? $container->getFooService())->someMethod(...);
```

As you can see, this requires instantiating the "foo" service, while the closure may never be called.

After this PR, a closure registered as a service can be declared as lazy, which will end up generating this instead (conceptually):
```php
return fn (...$arguments) => ($container->privates['foo'] ?? $container->getFooService())->someMethod(...$arguments);
```

With this code, "foo" is created only if the closure is actually called.

This is opt-in since it requires the "lazy" flag on the service definition.

PR #51251, opened 2023-08-03
[DependencyInjection] Do not add `return` in `LazyClosure` when return type of closure is `void`

ruudk:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | 
| License       | MIT
| Doc PR        | 

The LazyClosure was introduced in #49639 and generates PHP code using reflection.

It works, but when the callable has a `void` return type, it would produce code like this:
```php
return $container->services['closure2'] = (new class(fn() => new \Symfony\Component\DependencyInjection\Tests\Compiler\FooVoid()) extends \Symfony\Component\DependencyInjection\Argument\LazyClosure {
    public function __invoke(string $name) : void
    {
        // the `return` below causes the error
        return $this->service->__invoke(...\func_get_args());
    }
})->__invoke(...);
```

That `return` statement before calling the `$this->service` is not allowed and causes an error:
```
Compile Error: A void function must not return a value
```

/cc @nicolas-grekas

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Argument/LazyClosure.php
+++ b/src/Symfony/Component/DependencyInjection/Argument/LazyClosure.php
@@ -78,7 +78,8 @@ class LazyClosure
             throw new RuntimeException("Cannot create lazy closure{$id} because its corresponding callable is invalid.");
         }
 
-        $code = ProxyHelper::exportSignature($r->getMethod($method), true, $args);
+        $methodReflector = $r->getMethod($method);
+        $code = ProxyHelper::exportSignature($methodReflector, true, $args);
 
         if ($asClosure) {
             $code = ' { '.preg_replace('/: static$/', ': \\'.$r->name, $code);
@@ -87,7 +88,7 @@ class LazyClosure
         }
 
         $code = 'new class('.$initializer.') extends \\'.self::class
-            .$code.' { return $this->service->'.$callable[1].'('.$args.'); } '
+            .$code.' { '.($methodReflector->hasReturnType() && 'void' === (string) $methodReflector->getReturnType() ? '' : 'return ').'$this->service->'.$callable[1].'('.$args.'); } '
             .'}';
 
         return $asClosure ? '('.$code.')->'.$method.'(...)' : $code;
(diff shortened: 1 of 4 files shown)


src/Symfony/Component/DependencyInjection/Tests/Fixtures/php/lazy_closure.php
```diff
@@ -49,12 +51,22 @@ protected function createProxy($class, \Closure $factory)
...
+    }
+
+    /**
+     * Gets the public 'closure2' shared service.
+     *
+     * @return \Closure
+     */
+    protected static function getClosure2Service($container, $lazyLoad = true)
+    {
+        return $container->services['closure2'] = (new class(fn () => new \Symfony\Component\DependencyInjection\Tests\Compiler\FooVoid()) extends \Symfony\Component\DependencyInjection\Argument\LazyClosure { public function __invoke(string $name): void { $this->service->__invoke(...\func_get_args()); } })->__invoke(...);
```
ruudk, line 70: I think it can be further optimized for situations that use `__invoke`:
```php
new class(fn () => new \Symfony\Component\DependencyInjection\Tests\Compiler\FooVoid()) extends \Symfony\Component\DependencyInjection\Argument\LazyClosure {
    public function __invoke(string $name): void {
        ($this->service)(...\func_get_args()); 
    }
};
```

src/Symfony/Component/DependencyInjection/Argument/LazyClosure.php
```diff
@@ -87,7 +88,7 @@ public static function getCode(string $initializer, array $callable, Definition
         }
 
         $code = 'new class('.$initializer.') extends \\'.self::class
-            .$code.' { return $this->service->'.$callable[1].'('.$args.'); } '
+            .$code.' { '.($methodReflector->getReturnType() instanceof \ReflectionNamedType && 'void' === $methodReflector->getReturnType()->getName() ? '' : 'return ').'$this->service->'.$callable[1].'('.$args.'); } '
```
nicolas-grekas:
```suggestion
            .$code.' { '.($methodReflector->hasReturnType() && 'void' === (string) $methodReflector->getReturnType() ? '' : 'return ').'$this->service->'.$callable[1].'('.$args.'); } '
```

src/Symfony/Component/DependencyInjection/Tests/Fixtures/php/lazy_closure.php
nicolas-grekas: No need to, this wouldn't optimize anything in practice.

nicolas-grekas (approved):

nicolas-grekas:
Thank you @ruudk.

PR #51270, opened 2023-08-04
[OptionsResolver] Improve invalid type message on nested option

flohw:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | not required I think

OptionsResolver exception message is very explicit on which option has an invalid value with the full path of the option on lazy option. But the message only showed the latest option name when the type is not valid. This PR fixes it and display the full option path of the lazy option's invalid type.

### DIFF
--- a/src/Symfony/Component/OptionsResolver/CHANGELOG.md
+++ b/src/Symfony/Component/OptionsResolver/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+* Improve message with full path on invalid type in nested option
+
 6.3
 ---
 
--- a/src/Symfony/Component/OptionsResolver/Tests/OptionsResolverTest.php
+++ b/src/Symfony/Component/OptionsResolver/Tests/OptionsResolverTest.php
@@ -1009,6 +1009,32 @@ class OptionsResolverTest extends TestCase
         $this->resolver->resolve();
     }
 
+    public function testResolveFailsIfInvalidValueFromNestedOption()
+    {
+        $this->expectException(InvalidOptionsException::class);
+        $this->expectExceptionMessage('The option "foo[bar]" with value "invalid value" is invalid. Accepted values are: "valid value".');
+        $this->resolver->setDefault('foo', function (OptionsResolver $resolver) {
+            $resolver
+                ->setDefined('bar')
+                ->setAllowedValues('bar', 'valid value');
+        });
+
+        $this->resolver->resolve(['foo' => ['bar' => 'invalid value']]);
+    }
+
+    public function testResolveFailsIfInvalidTypeFromNestedOption()
+    {
+        $this->expectException(InvalidOptionsException::class);
+        $this->expectExceptionMessage('The option "foo[bar]" with value 1 is expected to be of type "string", but is of type "int".');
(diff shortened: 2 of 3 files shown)


flohw:
Some unrelated tests fail on local environment and here. :thinking: 

[edit]: I removed the latest commit as it make a lot of test fail. But fabbot is not happy anymore... Let me know if I need to update them.

src/Symfony/Component/OptionsResolver/Tests/OptionsResolverTest.php
```diff
@@ -1019,6 +1019,32 @@ public function testFailIfSetAllowedValuesFromLazyOption()
         $this->resolver->resolve();
     }
 
+    public function testResolveFailsIfInvalidValueFromLazyOption()
```
yceruto: I think you meant `...FromNestedOption` https://symfony.com/doc/current/components/options_resolver.html#nested-options

Lazy options have a different definition https://symfony.com/doc/current/components/options_resolver.html#default-values-that-depend-on-another-option

src/Symfony/Component/OptionsResolver/Tests/OptionsResolverTest.php
```diff
@@ -1019,6 +1019,32 @@ public function testFailIfSetAllowedValuesFromLazyOption()
...
+        $this->resolver->setDefault('foo', function (OptionsResolver $resolver) {
+            $resolver
+                ->setDefined('bar')
+                ->setAllowedValues('bar', 'valid value');
+        });
+
+        $this->resolver->resolve(['foo' => ['bar' => 'invalid value']]);
+    }
+
+    public function testResolveFailsIfInvalidTypeFromLazyOption()
```
yceruto: same here `Nested` instead of `Lazy`

src/Symfony/Component/OptionsResolver/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+5.4
+---
+
+ * Improve message with full path on invalid type in lazy option
```
yceruto: `s/lazy/nested`

yceruto (changes requested):
Improvements should go against 6.4 branch. Please rebase your commit.

I left a minor change request, 
Thanks!

yceruto:
> Some unrelated tests fail on local environment and here.

Nothing to do in your side, unless you want to fix them in a separate PR ;-)

> I removed the latest commit as it make a lot of test fail. But fabbot is not happy anymore... Let me know if I need to update them.

They seem unrelated to your changes here and should be ignored, in my opinion.

flohw:
Thanks for your review @yceruto :-)
And the difference between nested and lazy options.

yceruto:
It seems something is wrong with the tests. Could you please squash your commits so that the tests are triggered again?

flohw:
I have squashed my two last commits.

I won't be available the next two weeks because of holidays. But I will be able to make any required changes when I come back. :-)

yceruto (approved):

nicolas-grekas:
Thank you @flohw.

PR #44166, opened 2021-11-20
[Config] Use better typehint in PHP Configuration

jderusse:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

This PR now followup #46328 to replace the generic `mixed` type-hint by type inferred from normalizations (ie, if a config use `ifString`, when now that the normalization works only for string, these `mixed` can safely be replaced by `string`)

PR #51273, opened 2023-08-04
[Config] Add `ArrayNodeDefinition::acceptAndWrap()` to list alternative types that should be accepted and wrapped in an array

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

With #58450, we lost some type info when generating config builders, by preventing #44166 to work correctly.

This PR improves the description of config trees by adding an `ArrayNodeDefinition::acceptAndWrap()` method that allows one to declare which alternative types should be accepted at array-node levels. This declaration then triggers the wrapping of such cases into arrays, and also hints the config builder generator so that it can generate more accurate types for configurator methods.


maxbeckers:
@nicolas-grekas the scheduler test is already fixed with #51271 / #51274 and not related to your change!

src/Symfony/Component/Config/Definition/Builder/ExprBuilder.php
```diff
@@ -77,7 +77,7 @@ public function ifTrue(\Closure $closure = null): static
     public function ifString(): static
     {
         $this->ifPart = \is_string(...);
-        $this->allowedTypes = self::TYPE_STRING;
+        $this->allowedTypes = self::TYPE_STRING; // XXX why is this used by ConfigBuilderGenerator instead of the prototype?
```
maxbeckers: By default it's type `TYPE_ANY` what results in `mixed`. For me it seems that this is a workaround to interpret this as a string after the check.

src/Symfony/Component/Config/Definition/Builder/ArrayNodeDefinition.php
```diff
@@ -498,7 +517,7 @@ protected function createNode(): NodeInterface
 
         if (isset($this->normalization)) {
             $node->setNormalizationClosures($this->normalization->before);
-            $node->setNormalizedTypes($this->normalization->declaredTypes);
+            $node->setNormalizedTypes($this->allowedTypes ?? $this->normalization->declaredTypes);
```
stof, line 531: Why is `$this->normalization->declaredTypes` filled normally ? and why do we need to add this new `$this->allowedTypes` property instead of using the existing feature (which gets ignored entirely) ?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -112,20 +114,22 @@ public function getConfigTreeBuilder(): TreeBuilder
...
+                    ->beforeNormalization()->ifArray()->then(static fn ($v) => [''] === $v ? [] : $v)->end()
                     ->prototype('scalar')->end()
                     ->defaultValue(['%env(default::SYMFONY_TRUSTED_HOSTS)%'])
                 ->end()
                 ->variableNode('trusted_proxies')
                     ->beforeNormalization()
-                        ->ifTrue(fn ($v) => 'private_ranges' === $v || 'PRIVATE_SUBNETS' === $v)
-                        ->then(fn () => IpUtils::PRIVATE_SUBNETS)
+                        ->ifString()
+                        ->then(static fn ($v) => 'private_ranges' === $v || 'PRIVATE_SUBNETS' === $v ? IpUtils::PRIVATE_SUBNETS : $v)
```
stof: Having to add extra conditions in the `then` part makes the code less readable IMO, by making it harder to figure out when things are actually changed.

If we need to keep an explicit string check to tell the code generator that the normalization supports only string input, we should probably add an `if*` method allowing to provide a closure taking a string as input, where the final `if` condition would be `is_string($v) && $$callback($v)` (and still recording `string` as the allowed type)

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -112,20 +114,22 @@ public function getConfigTreeBuilder(): TreeBuilder
...
                     ->end()
                     ->defaultValue(['%env(default::SYMFONY_TRUSTED_PROXIES)%'])
                 ->end()
                 ->arrayNode('trusted_headers', 'trusted_header')
                     ->performNoDeepMerging()
-                    ->beforeNormalization()->ifString()->then(static fn ($v) => $v ? [$v] : [])->end()
+                    ->castToArray(['string'])
+                    ->beforeNormalization()->ifArray()->then(static fn ($v) => [''] === $v ? [] : $v)->end()
```
stof: How is this better than the previous code matching `ifString` ?

src/Symfony/Component/Config/Definition/Builder/ArrayNodeDefinition.php
```diff
@@ -79,6 +84,20 @@ public function setBuilder(NodeBuilder $builder): void
...
+     *
+     * @param list<ExprBuilder::TYPE_*> $allowedTypes
+     *
+     * @return $this
+     */
+    public function castToArray(array $allowedTypes, ?string $key = null): static
+    {
+        $this->allowedTypes = $allowedTypes;
+
+        return $this->beforeNormalization()->castToArray($key)->end();
```
stof: this non-conditional cast to array is _not_ a replacement for the existing normalization in most places where you started using it.

stof: OK, looking at the implementation of `castToArray`, it actually is a replacement, as it overrides any existing `ifPart` (meaning we have dead code).

src/Symfony/Component/Config/Definition/Builder/ArrayNodeDefinition.php
nicolas-grekas: I was wondering the same until I (re)discovered #44166, which is based on that: sniffing for allowed types in what I'd call an indirect approach. I'm here proposing a more explicit approach, which wins over the implicit one when used.

stof: overriding entirely the types inferred from the normalization looks wrong to me (unless we want to deprecate that behavior).
And the suggested API does not play well with nodes supporting multiple normalization rules.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2291,10 +2291,6 @@ private function registerLockConfiguration(array $config, ContainerBuilder $cont
             // Generate stores
             $storeDefinitions = [];
             foreach ($resourceStores as $resourceStore) {
-                if (null === $resourceStore) {
-                    $resourceStore = 'null';
```
nicolas-grekas, line 2295: this concern is now handled in the configuration [L1532](https://github.com/symfony/symfony/pull/51273/files#diff-041b166c698c254a3f61cfc6390785d4f5097794b77a4b2f3f888996729cbce4R1532)

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/PhpFrameworkExtensionTest.php
```diff
@@ -72,36 +72,6 @@ public function testAssetPackageCannotHavePathAndUrl()
         });
     }
 
-    public function testWorkflowValidationPlacesIsArray()
```
nicolas-grekas, line 75: These 2 test cases used to cover a custom implementation of validation logic that is now replaced by built-in validation done by the config component; which is already tested separately.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
nicolas-grekas: I'm reverting this change: on a VariableNode, this makes no difference.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -536,11 +539,8 @@ private function addWorkflowSection(ArrayNodeDefinition $rootNode): void
                                     ->end()
                                     ->arrayNode('transitions', 'transition')
                                         ->beforeNormalization()
-                                            ->always(static function ($transitions) {
-                                                if (!\is_array($transitions)) {
-                                                    throw new InvalidConfigurationException('The "transitions" option must be an array in workflow configuration.');
```
nicolas-grekas, line 541: this is replaced by `ifArray()` just below

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1524,7 +1529,8 @@ private function addLockSection(ArrayNodeDefinition $rootNode, callable $enableI
                             ->end()
                             ->prototype('array')
                                 ->performNoDeepMerging()
-                                ->beforeNormalization()->castToArray()->end()
+                                ->beforeNormalization()->ifNull()->then(static fn () => ['null'])->end()
```
nicolas-grekas: Here, we have a type (`null`) that shouldn't be allowed when generating config builders, but where the value (still `null`) have to be dealt with because of the XML loader, where the a `<foo>null</foo>` tag is loaded as a `null` when one meant the `"null"` string.
This is a canonical example where a call to `->acceptAndWrap()` above is decoupled from the actual types managed by normalizers.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -2638,11 +2628,13 @@ private function addHtmlSanitizerSection(ArrayNodeDefinition $rootNode, callable
                                     ->end()
                                     ->arrayNode('allowed_link_schemes', 'allowed_link_scheme')
                                         ->info('Allows only a given list of schemes to be used in links href attributes.')
+                                        ->acceptAndWrap(['string'])
```
nicolas-grekas, line 2632: These added `acceptAndWrap()` with no corresponding `castToArray()` make explicit that strings are also allowed. As a reminder, config builders already do this wrapping implicitly.

src/Symfony/Component/Config/Definition/Builder/ArrayNodeDefinition.php
nicolas-grekas: It has to be this way, see [L1532](https://github.com/symfony/symfony/pull/51273/files#diff-041b166c698c254a3f61cfc6390785d4f5097794b77a4b2f3f888996729cbce4R1532) in the `Configuration` class for an example: implicit types are loose and too many times we have to fallback to `mixed` (`->always()`, `->ifTrue($fn)`, etc.). Being explicit need to override the implicit collection of possible types - as not all types managed by normalizers have to be accepted at the config builder level (because of the XML loader mainly).
Which means this has to be the way to go.

I renamed the new method `->acceptAndWrap()` instead of `->castToArray()`. This is both more explicit and more accurate.

nicolas-grekas (commented):
PR ready (failures are false-positives).

The meat of the PR is the added `->acceptAndWrap()`, which enables declaring which alternative types should be accepted (and wrapped) for array-nodes.

src/Symfony/Component/Config/Definition/Builder/ArrayNodeDefinition.php
```diff
@@ -79,6 +84,30 @@ public function setBuilder(NodeBuilder $builder): void
...
+    {
+        $this->allowedTypes = $allowedTypes;
+
+        foreach ($allowedTypes as $type) {
+            $this->beforeNormalization()->ifTrue(match ($type) {
+                ExprBuilder::TYPE_INT => is_int(...),
+                ExprBuilder::TYPE_STRING => is_string(...),
+                ExprBuilder::TYPE_BOOL => is_bool(...),
+                ExprBuilder::TYPE_NULL => is_null(...),
+                ExprBuilder::TYPE_BACKED_ENUM => static fn ($v) => $v instanceof \BackedEnum,
```
stof, line 105: This does not handle `ExprBuilder::TYPE_ANY`, while this is a valid value in the documented type `list<ExprBuilder::TYPE_*>`

stof: same for `TYPE_ARRAY` btw

nicolas-grekas: Yes, on purpose, but this means I should list only the allowed subset, right?

nicolas-grekas: updated

src/Symfony/Bundle/TwigBundle/DependencyInjection/Configuration.php
```diff
@@ -34,17 +33,6 @@ public function getConfigTreeBuilder(): TreeBuilder
 
         $rootNode
             ->docUrl('https://symfony.com/doc/{version:major}.{version:minor}/reference/configuration/twig.html', 'symfony/twig-bundle')
-            ->beforeNormalization()
-            ->ifTrue(fn ($v) => \is_array($v) && \array_key_exists('exception_controller', $v))
```
nicolas-grekas, line 38: this is legacy BC/FC layer dating from Symfony 4/5

src/Symfony/Component/Config/Definition/Builder/ArrayNodeDefinition.php
stof: if the method intentionally accepts only a subset, the parameter type should indeed list only that subset (so that other constants are not documented as valid inputs)

src/Symfony/Component/Config/Definition/Builder/ArrayNodeDefinition.php
```diff
@@ -79,6 +84,30 @@ public function setBuilder(NodeBuilder $builder): void
         $this->nodeBuilder = $builder;
     }
 
+    /**
+     * Allows alternative types and wraps them into arrays.
+     *
+     * @param list<ExprBuilder::TYPE_INT|ExprBuilder::TYPE_STRING|ExprBuilder::TYPE_BOOL|ExprBuilder::TYPE_NULL|ExprBuilder::TYPE_BACKED_ENUM> $allowedTypes
+     *
```
GromNaN, line 92: Could you also document the `$key` parameter? It was not obvious to me before reading the code.

src/Symfony/Component/Config/Definition/Builder/ArrayNodeDefinition.php
```diff
@@ -79,6 +84,30 @@ public function setBuilder(NodeBuilder $builder): void
...
+    {
+        $this->allowedTypes = $allowedTypes;
+
+        foreach ($allowedTypes as $type) {
+            $this->beforeNormalization()->ifTrue(match ($type) {
+                ExprBuilder::TYPE_INT => is_int(...),
+                ExprBuilder::TYPE_STRING => is_string(...),
+                ExprBuilder::TYPE_BOOL => is_bool(...),
+                ExprBuilder::TYPE_NULL => is_null(...),
+                ExprBuilder::TYPE_BACKED_ENUM => static fn ($v) => $v instanceof \BackedEnum,
```
GromNaN, line 105: Why is it restricted to `BackedEnum`? I don't see what would be the issue with regular `Enum`as they are supported all config formats.

src/Symfony/Component/Config/Definition/Builder/ArrayNodeDefinition.php
```diff
@@ -79,6 +84,30 @@ public function setBuilder(NodeBuilder $builder): void
...
+     *
+     * @return $this
+     */
+    public function acceptAndWrap(array $allowedTypes, ?string $key = null): static
+    {
+        $this->allowedTypes = $allowedTypes;
+
+        foreach ($allowedTypes as $type) {
+            $this->beforeNormalization()->ifTrue(match ($type) {
+                ExprBuilder::TYPE_INT => is_int(...),
```
GromNaN, line 101: I'm not sure if there's a use case, but `float` should be added for completeness.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -738,11 +743,11 @@ private function addRequestSection(ArrayNodeDefinition $rootNode): void
                         ->arrayNode('formats', 'format')
                             ->useAttributeAsKey('name')
                             ->prototype('array')
+                                ->acceptAndWrap(['string'])
                                 ->beforeNormalization()
-                                    ->ifTrue(fn ($v) => \is_array($v) && isset($v['mime_type']))
-                                    ->then(fn ($v) => $v['mime_type'])
+                                    ->ifArray()
+                                    ->then(static fn ($v) => (array) ($v['mime_type'] ?? $v))
```
GromNaN, line 749: One more reason to remove the XML config format, this is only necessary to represent an array in XML (https://github.com/symfony/symfony/pull/60568, [doc](https://symfony.com/doc/7.3/reference/configuration/framework.html#formats)).

The conversion of the `mime_type` field cannot be exposed by the config tree. Or else we need another method `acceptValueFromNestedProperty(property: 'mime_type', acceptedTypes: 'string')`.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
GromNaN: A comment in the code about this would be helpful.

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/AccessTokenFactory.php
```diff
@@ -47,39 +47,28 @@ public function addConfiguration(NodeDefinition $node): void
...
                 ->end()
 
                 // "isRequired" must be set otherwise the following custom validation is not called
                 ->isRequired()
                 ->beforeNormalization()
-                    ->ifTrue(fn ($v) => \is_array($v) && !$v)
-                    ->then(fn () => throw new InvalidConfigurationException('You must set a token handler.'))
+                    ->ifTrue(static fn ($v) => \is_array($v) && !$v)
+                    ->then(static fn () => throw new InvalidConfigurationException('You must set a token handler.'))
                 ->end()
```
GromNaN: Not directly related to this PR, this seems to be validation rules, not normalization.

It can be replace by this:
```php
                ->validate()
                    ->ifTrue(static fn ($v) => 1 < \count($v))
                    ->thenInvalid('You cannot configure multiple token handlers.')
                ->end()

                // "isRequired" must be set otherwise the following custom validation is not called
                ->isRequired()
                ->validate()
                    ->ifEmpty()
                    ->thenInvalid('You must set a token handler.')
                ->end()
```

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1769,13 +1766,14 @@ function ($a) {
...
-                                                'enabled' => 'allow_no_handlers' === $v ? true : $v,
+                                            ->ifString()
+                                            ->then(static fn ($v) => [
+                                                'enabled' => 'allow_no_handlers' === $v,
                                                 'allow_no_handlers' => 'allow_no_handlers' === $v,
-                                                'allow_no_senders' => true,
                                             ])
                                         ->end()
+                                        ->beforeNormalization()->ifTrue()->then(static fn ($v) => ['enabled' => true])->end()
+                                        ->beforeNormalization()->ifFalse()->then(static fn ($v) => ['enabled' => false])->end()
```
GromNaN: The argument is not used.
```suggestion
                                        ->beforeNormalization()->ifTrue()->then(static fn () => ['enabled' => true])->end()
                                        ->beforeNormalization()->ifFalse()->then(static fn () => ['enabled' => false])->end()
```

But you could also add `ExprBuilder::isBool`.

```php
->beforeNormalization()->ifBool()->then(static fn ($v) => ['enabled' => $v])->end()

```

GromNaN (commented):
I tried using the new method in https://github.com/doctrine/DoctrineMongoDBBundle/pull/921, and its works very well 🎉

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
nicolas-grekas: I'd prefer not adding too much stuff in this PR (same for floats and unit-enums)

src/Symfony/Component/Config/Definition/Builder/ArrayNodeDefinition.php
nicolas-grekas: We don't have a usecase for `UnitEnum` yet. Ideally, we'd want to support which kind of enum class might be accepted, but that's another story. Also: enums don't fit for env vars, so not sure adding more type tidiness to config is going to actually improve things.

nicolas-grekas:
Thanks for the review @GromNaN - PR updated.

src/Symfony/Component/Config/Definition/Builder/ArrayNodeDefinition.php
GromNaN: That's right, let's keep that for a future improvement.

GromNaN (approved):

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

PR #51284, opened 2023-08-05
[FrameworkBundle][HttpKernel][MonologBridge] Revisit wiring of debug loggers

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | -
| License       | MIT
| Doc PR        | -

Replaces #51229

Implements the suggestions from @stof and @Seldaek in https://github.com/symfony/symfony/pull/51229#discussion_r1282056615

It turns our we had almost everything available. This is mostly a matter of wiring.

* Deprecate class `Symfony\Bridge\Monolog\Logger`
* Deprecate `AddDebugLogProcessorPass::configureLogger()`
* Add `DebugLoggerConfigurator` to HttpKernel, which wires `DebugProcessor` when desired

Bonus:
* Add argument `$debug` to HttpKernel's `Logger`
* Deprecate `EnableLoggerDebugModePass` in favor of the previous

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/EnableLoggerDebugModePass.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/EnableLoggerDebugModePass.php
@@ -11,10 +11,15 @@
 
 namespace Symfony\Bundle\FrameworkBundle\DependencyInjection\Compiler;
 
+trigger_deprecation('symfony/framework-bundle', '6.4', 'The "%s" class is deprecated, use argument $debug of HttpKernel\'s Logger instead.', EnableLoggerDebugModePass::class);
+
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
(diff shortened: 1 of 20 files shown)


src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/EnableLoggerDebugModePass.php
```diff
@@ -32,7 +37,7 @@ public function process(ContainerBuilder $container): void
 
     public static function configureLogger(Logger $logger): void
     {
-        if (!\in_array(\PHP_SAPI, ['cli', 'phpdbg'], true) && method_exists($logger, 'enableDebug')) {
```
stof, line 35: why removing this check ?

nicolas-grekas: because the method is always available per composer constraints nowadays

PR #47422, opened 2022-08-29
[Process] Support using `Process::findExecutable()` independently of `open_basedir`

BlackbitDevs:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #41006
| License       | MIT

With this PR `Console::findExecutable()` will find executables also if the their path is not allowed in `open_basedir` config similar to  https://github.com/symfony/symfony/blob/ddaedd28bd2e4fc22b64567753cf934fe9d68c4c/src/Symfony/Component/Process/PhpExecutableFinder.php#L36-L41
`Console::findExecutable()`'s responsibility is to find an executable which can be called with a Symfony Process or by PHP's functions like `exec`, `system` etc.

The goal of PHP's `open_basedir` config is to restrict reading / writing files within PHP processes. Imho this is completely independent of finding an executable.

If PHP's intention was to restrict executing applications which are not present in `open_basedir`'s paths, it would have been implemented there.

PR #51297, opened 2023-08-07
[Process] fix tests

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

- `testFindProcessInOpenBasedir` is a duplicate of `testFindWithOpenBaseDir`
- `testFindWithOpenBaseDir` currently expects that we search open_basedir instead of PATH when the setting is set, but this doesn't really make sense, and #47422 removed this behavior
- `PhpSubprocessTest::testSubprocess` expects a php that defaults to memory_limit=-1, which is not the case currently for the sigchild-enabled binary

### DIFF
--- a/src/Symfony/Component/Process/Tests/ExecutableFinderTest.php
+++ b/src/Symfony/Component/Process/Tests/ExecutableFinderTest.php
@@ -98,30 +98,11 @@ class ExecutableFinderTest extends TestCase
             $this->markTestSkipped('Cannot test when open_basedir is set');
         }
 
+        putenv('PATH='.\dirname(\PHP_BINARY));
         $this->iniSet('open_basedir', \dirname(\PHP_BINARY).\PATH_SEPARATOR.'/');
 
         $finder = new ExecutableFinder();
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/Process/Tests/ExecutableFinderTest.php
```diff
@@ -98,6 +98,7 @@ public function testFindWithOpenBaseDir()
             $this->markTestSkipped('Cannot test when open_basedir is set');
         }
 
+        putenv('PATH='.\dirname(\PHP_BINARY));
```
stloyd, line 101: Shouldn't be restored to previous value at end of the test?

nicolas-grekas: Already the case, see https://github.com/symfony/symfony/pull/51297/files#diff-da8254fae04a586ecebf906fa89d9c726042004da48af989f9b68aeb98ad6776L24

PR #51301, opened 2023-08-07
[FrameworkBundle] add missing default-doctrine-dbal-provider cache pool attribute to XSD

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/schema/symfony-1.0.xsd
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/schema/symfony-1.0.xsd
@@ -290,6 +290,7 @@
             <xsd:element name="default-psr6-provider" type="xsd:string" minOccurs="0" maxOccurs="1" />
             <xsd:element name="default-redis-provider" type="xsd:string" minOccurs="0" maxOccurs="1" />
             <xsd:element name="default-memcached-provider" type="xsd:string" minOccurs="0" maxOccurs="1" />
+            <xsd:element name="default-doctrine-dbal-provider" type="xsd:string" minOccurs="0" maxOccurs="1" />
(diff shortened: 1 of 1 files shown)


nicolas-grekas (approved):

src/Symfony/Bundle/FrameworkBundle/Resources/config/schema/symfony-1.0.xsd
```diff
@@ -290,6 +290,7 @@
             <xsd:element name="default-psr6-provider" type="xsd:string" minOccurs="0" maxOccurs="1" />
             <xsd:element name="default-redis-provider" type="xsd:string" minOccurs="0" maxOccurs="1" />
             <xsd:element name="default-memcached-provider" type="xsd:string" minOccurs="0" maxOccurs="1" />
+            <xsd:element name="default-doctrine-dbal-provider" type="xsd:string" minOccurs="0" maxOccurs="1" />
```
stof, line 293: this should be an attribute, not an element, to match our XML conventions. The other `xsd:string` elements that cannot occur multiple times should have been attributes too, and we already added them as attributes when documenting the XML config and finding this out. So I suggest that new settings are added only as attribute to avoid continuing the non-conventional format.

And what is the difference between `default-doctrine-dbal-provider` and `default-doctrine-provider` ?

nicolas-grekas: I favor consistency on this one.

> And what is the difference between default-doctrine-dbal-provider and default-doctrine-provider ?

`default-doctrine-provider` should have been named `default-doctrine-cache-provider`, it's using an adapter to doctrine/cache

nicolas-grekas:
Thank you @xabbuh.

PR #51320, opened 2023-08-09
[Security] Prevent creating session in stateless firewalls

### LINKED ISSUE #51319: Session created by default handlers on stateless firewalls

6.3

### Description

Tests are broken on [that project](https://github.com/hwi/HWIOAuthBundle/pull/1938) since Symfony 6.0 and after more investigation I found that Symfony is starting the session in `DefaultAuthenticationSuccessHandler::onAuthenticationSuccess()` when `determineTargetUrl()` read the session to get the target URL, even if the firewall is marked as stateless.

This result in the error:
> Session was used while the request was declared stateless.

### How to reproduce

Setup a stateless firewall using `DefaultAuthenticationSuccessHandler`.

See failing test here: https://github.com/hwi/HWIOAuthBundle/pull/1938

### Possible Solution

_No response_

### Additional Context

_No response_

Seb33300:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix https://github.com/symfony/symfony/issues/51319
| License       | MIT
| Doc PR        | 

Please check related issue for details.

### DIFF
--- a/src/Symfony/Component/Security/Http/Authentication/DefaultAuthenticationFailureHandler.php
+++ b/src/Symfony/Component/Security/Http/Authentication/DefaultAuthenticationFailureHandler.php
@@ -91,7 +91,9 @@ class DefaultAuthenticationFailureHandler implements AuthenticationFailureHandle
 
         $this->logger?->debug('Authentication failure, redirect triggered.', ['failure_path' => $options['failure_path']]);
 
-        $request->getSession()->set(SecurityRequestAttributes::AUTHENTICATION_ERROR, $exception);
+        if ($request->hasPreviousSession()) {
+            $request->getSession()->set(SecurityRequestAttributes::AUTHENTICATION_ERROR, $exception);
+        }
 
         return $this->httpUtils->createRedirectResponse($request, $options['failure_path']);
     }
--- a/src/Symfony/Component/Security/Http/Authentication/DefaultAuthenticationSuccessHandler.php
+++ b/src/Symfony/Component/Security/Http/Authentication/DefaultAuthenticationSuccessHandler.php
@@ -103,7 +103,7 @@ class DefaultAuthenticationSuccessHandler implements AuthenticationSuccessHandle
         }
 
         $firewallName = $this->getFirewallName();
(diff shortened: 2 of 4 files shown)


src/Symfony/Component/Security/Http/Authentication/DefaultAuthenticationFailureHandler.php
```diff
@@ -91,7 +91,9 @@ public function onAuthenticationFailure(Request $request, AuthenticationExceptio
 
         $this->logger?->debug('Authentication failure, redirect triggered.', ['failure_path' => $options['failure_path']]);
 
-        $request->getSession()->set(SecurityRequestAttributes::AUTHENTICATION_ERROR, $exception);
+        if ($request->hasPreviousSession()) {
```
Seb33300, line 94: Better if we could check if `!$options['stateless']` here, but that option is not passed

Seb33300: We cannot check on `$request->hasPreviousSession()` this is the reason why the test is failing...

stof: hasPreviousSession is absolutely not the same than stateless.

Seb33300: So we need a way to access the `stateless` option from the firewall.
If someone has an idea how to do this, please help on that. I will close that PR for now.

chalasr: @Seb33300 You can check if the `_stateless` attribute is set to `true` instead

Seb33300: Good point! I modified my PR and reopened it.

Seb33300: Oups cannot reopen because I force pushed...
Here is the new PR: https://github.com/symfony/symfony/pull/51350

PR #51324, opened 2023-08-09
[HttpFoundation] Add `QueryParameterRequestMatcher`

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | _NA_
| License       | MIT
| Doc PR        | Todo

We know we need a `authorizationCode` query parameter to support a request in our authenticator. We would love to use only the `ChainRequestMatcher` to do so. I'd like to add this `QueryParameterRequestMatcher` in order to do the following:

```php
class OurAuthenticator extends AbstractAuthenticator
{
    public function supports(Request $request): ?bool
    {
        return (new ChainRequestMatcher([
            // ...
            new PathRequestMatcher('/sso/provider'),
            new QueryParameterRequestMatcher('authorizationCode')
        ]))->matches($request);
    }

    // ...
}
```

Which would match the following: `/sso/provider?authorizationCode=...`

### DIFF
--- a/src/Symfony/Component/HttpFoundation/CHANGELOG.md
+++ b/src/Symfony/Component/HttpFoundation/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * Add `UploadedFile::getClientOriginalPath()`
+ * Add `QueryParameterRequestMatcher`
 
 7.0
 ---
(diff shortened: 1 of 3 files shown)


norkunas:
While writing github webhook request parser I encountered that they send signature as a header, would it make sense to also add `HeaderRequestMatcher` to core?

alexandre-daubois:
That would be a big yes for me, definitely! Gave a try here: https://github.com/symfony/symfony/pull/51343

Thank you for the idea, it would be even more powerful with those features!

alexandre-daubois:
Rebased and conflicts fixed

OskarStark (approved):

fabpot (approved):

fabpot:
Thank you @alexandre-daubois.

PR #51327, opened 2023-08-09
[FrameworkBundle] Add `AbstractController::renderBlock()` and `renderBlockView()`

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

This would be especially useful when generating turbo-stream responses.
Right now, [the doc](https://symfony.com/bundles/ux-turbo/current/index.html#coming-alive-with-turbo-streams) recommends creating new partial templates, but this increases the complexity and scatters HTML fragments. Instead, we could encourage using twig blocks.

Adding this could help remove some boilerplate.

before (using blocks):
```php
return new Response($this->container->get('twig')->load('foo.html.twig')->renderBlock('the_block', $context));
```

after:
```php
return $this->renderBlock('foo.html.twig', 'the_block', $context);
```

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php
@@ -229,17 +229,17 @@ abstract class AbstractController implements ServiceSubscriberInterface
      */
     protected function renderView(string $view, array $parameters = []): string
     {
-        if (!$this->container->has('twig')) {
-            throw new \LogicException('You cannot use the "renderView" method if the Twig Bundle is not available. Try running "composer require symfony/twig-bundle".');
-        }
+        return $this->doRenderView($view, null, $parameters, __FUNCTION__);
+    }
 
-        foreach ($parameters as $k => $v) {
-            if ($v instanceof FormInterface) {
-                $parameters[$k] = $v->createView();
-            }
-        }
-
-        return $this->container->get('twig')->render($view, $parameters);
+    /**
+     * Returns a rendered block from a view.
+     *
+     * Forms found in parameters are auto-cast to form views.
+     */
+    protected function renderBlockView(string $view, string $block, array $parameters = []): string
+    {
+        return $this->doRenderView($view, $block, $parameters, __FUNCTION__);
     }
 
     /**
@@ -250,21 +250,18 @@ abstract class AbstractController implements ServiceSubscriberInterface
      */
     protected function render(string $view, array $parameters = [], Response $response = null): Response
     {
-        $content = $this->renderView($view, $parameters);
-        $response ??= new Response();
+        return $this->doRender($view, null, $parameters, $response, __FUNCTION__);
+    }
 
-        if (200 === $response->getStatusCode()) {
-            foreach ($parameters as $v) {
-                if ($v instanceof FormInterface && $v->isSubmitted() && !$v->isValid()) {
-                    $response->setStatusCode(422);
-                    break;
-                }
-            }
-        }
-
-        $response->setContent($content);
-
-        return $response;
+    /**
+     * Renders a block in a view.
+     *
+     * If an invalid form is found in the list of parameters, a 422 status code is returned.
+     * Forms found in parameters are auto-cast to form views.
+     */
+    protected function renderBlock(string $view, string $block, array $parameters = [], Response $response = null): Response
+    {
+        return $this->doRender($view, $block, $parameters, $response, __FUNCTION__);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


stof:
I think it might be more readable with a separate `renderBlock` method instead of an optional argument, allowing to easily distinguish whether we only render a block of the template and not the full block (especially when the list of variables become long and is defined inline rather than using a `$context` variable).

nicolas-grekas:
In order to prevent code duplication, we could add renderBlock but keep the new argument on renderView. Otherwise we'll need a new renderBlockView and it should duplicate most of the code in renderView/render. I'm not keen on adding renderBlockView.

nicolas-grekas:
Mmm, even adding renderBlock requires duplicating from render.

Another idea might be to use the parameters array to add options, for example `['@block' => 'the_block']`.
Can we reference a variable named `@block` from inside a twig template? If not, then this convention is safe from collisions.
But it's more difficult to discover.

With all this in mind, I think I prefer the additional argument.

yceruto:
I prefer the separate methods approach too, as it is more aligned with SRP and also results in a better argument order.
```php
return $this->renderBlock('foo.html.twig', 'the_block', $context);
```

<details><summary>An alternative could be moving the duplicated code to new private methods</summary>
<p>

```php
private function normalizeViewParameters(array $parameters): array
{
    foreach ($parameters as $k => $v) {
        if ($v instanceof FormInterface) {
            $parameters[$k] = $v->createView();
        }
    }

    return $parameters;
}

private function prepareResponse(Response $response, array $parameters, string $content): Response
{
    if (200 === $response->getStatusCode()) {
        foreach ($parameters as $v) {
            if ($v instanceof FormInterface && $v->isSubmitted() && !$v->isValid()) {
                $response->setStatusCode(422);
                break;
            }
        }
    }

    $response->setContent($content);

    return $response;
}
```

</p>
</details>

nicolas-grekas:
OK, thanks for the feedback :)

PR updated to add method `AbstractController::renderBlock()` and argument `$block` to `renderView()`

yceruto:
by the way, I would do the same for `renderBlockView()`, and use `normalizeViewParameters()` (defined in my previous comment) to avoid duplication. The LogicException is not included as it's a different method, so the message changes.

yceruto:
Full suggestion https://gist.github.com/yceruto/fc5594963df6a642c34111eea1a82b17

nicolas-grekas:
I'm not sold on renderViewBlock. There I still like the additional argument for renderView...

nicolas-grekas:
Friendly ping @symfony/mergers

src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php
```diff
@@ -226,9 +226,13 @@ protected function denyAccessUnlessGranted(mixed $attribute, mixed $subject = nu
      * Returns a rendered view.
      *
      * Forms found in parameters are auto-cast to form views.
+     *
+     * @param string|null $block
      */
-    protected function renderView(string $view, array $parameters = []): string
+    protected function renderView(string $view, array $parameters = []/* , string $block = null */): string
```
yceruto: Sorry I dont see why adding this new argument like this will be better than adding a new method `renderViewBlock()`, just like you did for `renderBlock()`, can you please elaborate a bit more about it?

yceruto: IMHO, they've the same reason as `renderBlock`:
* no deprecation warning for subclasses
* imho better argument order `renderViewBlock('foo.html.twig', 'block_name', [...])` aligned with other render methods

nicolas-grekas: That's because I have some reluctance to adding more and more methods that are slight variations of each others in AbstractController, which is already a long collection of helpers.

`$this->render('foo.html.twig', block: 'bar', parameters: $context)` would read equally well to me compared to `renderBlock()` but many have expressed their preference so I abide by them :) 

For `renderBlockView`, my reluctance is still there, because yet another method, while this method won't be much used (we didn't miss it for now). Also `$this->renderView('foo.html.twig', $context, block: 'bar')` reads nicely so keeping AbstractController shorter in terms of number of methods has my preference.

stof: keep in mind that the feature you add does not support `$this->renderView('foo.html.twig', $context, block: 'bar')` as named arguments don't work with those forward-compatibility layers.
It will only work in 7.0, not in 6.4.

yceruto: I understand your concern as a maintainer, and it makes sense to me. However, from the DX perspective, it will create some kind of inconsistency in how `renderBlock` and `renderView` are used (while your goal is rendering blocks). I think adding one more method doesn't hurt too much if it results in a more intuitive API

stof: Thus, our BC policy current states that parameter names are not covered by BC except for attributes, which means that such code should not be encouraged without first changing the policy (which would then require figuring out the proper migration path for named arguments, both for backward and forward compatibility)

nicolas-grekas: OK, I updated the PR!

yceruto (approved):

kbond (approved):

dunglas (approved):
Cool!

lyrixx (approved):
I like it !

ogizanagi (approved):

PR #51331, opened 2023-08-09
[Messenger] add handler description as array key to `HandlerFailedException::getWrappedExceptions()`

kbond:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Currently, when looking at `HandlerFailedException` to see what exceptions were thrown for a message, you can't see what handler caused the exception.

### DIFF
--- a/src/Symfony/Component/Messenger/Exception/HandlerFailedException.php
+++ b/src/Symfony/Component/Messenger/Exception/HandlerFailedException.php
@@ -20,7 +20,7 @@ class HandlerFailedException extends RuntimeException implements WrappedExceptio
     private Envelope $envelope;
 
     /**
-     * @param \Throwable[] $exceptions
+     * @param \Throwable[] $exceptions The name of the handler should be given as key
      */
     public function __construct(Envelope $envelope, array $exceptions)
     {
@@ -55,7 +55,7 @@ class HandlerFailedException extends RuntimeException implements WrappedExceptio
     {
         trigger_deprecation('symfony/messenger', '6.4', 'The "%s()" method is deprecated, use "%s::getWrappedExceptions()" instead.', __METHOD__, self::class);
 
-        return $this->exceptions;
+        return array_values($this->exceptions);
     }
 
     /**
--- a/src/Symfony/Component/Messenger/Tests/Middleware/HandleMessageMiddlewareTest.php
+++ b/src/Symfony/Component/Messenger/Tests/Middleware/HandleMessageMiddlewareTest.php
@@ -47,6 +47,36 @@ class HandleMessageMiddlewareTest extends MiddlewareTestCase
         $middleware->handle($envelope, $this->getStackMock());
     }
 
+    public function testItKeysTheHandlerFailedNestedExceptionsByHandlerDescription()
+    {
+        $message = new DummyMessage('Hey');
+        $envelope = new Envelope($message);
+        $handler = new class() {
+            public function __invoke()
+            {
+                throw new \Exception('failed');
+            }
+        };
+
+        $middleware = new HandleMessageMiddleware(new HandlersLocator([
+            DummyMessage::class => [$handler],
+        ]));
+
+        try {
+            $middleware->handle($envelope, $this->getStackMock(false));
+        } catch (HandlerFailedException $e) {
+            $key = (new HandlerDescriptor($handler))->getName();
+
+            $this->assertCount(1, $e->getWrappedExceptions());
+            $this->assertArrayHasKey($key, $e->getWrappedExceptions());
+            $this->assertSame('failed', $e->getWrappedExceptions()[$key]->getMessage());
+
+            return;
+        }
+
+        $this->fail('Exception not thrown.');
+    }
+
     /**
      * @dataProvider itAddsHandledStampsProvider
      */
--- a/src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
+++ b/src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
@@ -66,7 +66,7 @@ class HandleMessageMiddleware implements MiddlewareInterface
                 if ($batchHandler && $ackStamp = $envelope->last(AckStamp::class)) {
                     $ack = new Acknowledger(get_debug_type($batchHandler), static function (\Throwable $e = null, $result = null) use ($envelope, $ackStamp, $handlerDescriptor) {
                         if (null !== $e) {
-                            $e = new HandlerFailedException($envelope, [$e]);
+                            $e = new HandlerFailedException($envelope, [$handlerDescriptor->getName() => $e]);
                         } else {
                             $envelope = $envelope->with(HandledStamp::fromDescriptor($handlerDescriptor, $result));
                         }
@@ -95,7 +95,7 @@ class HandleMessageMiddleware implements MiddlewareInterface
                 $envelope = $envelope->with($handledStamp);
                 $this->logger?->info('Message {class} handled by {handler}', $context + ['handler' => $handledStamp->getHandlerName()]);
             } catch (\Throwable $e) {
-                $exceptions[] = $e;
+                $exceptions[$handlerDescriptor->getName()] = $e;
             }
         }
 
@@ -107,7 +107,7 @@ class HandleMessageMiddleware implements MiddlewareInterface
                     $handler = $stamp->getHandlerDescriptor()->getBatchHandler();
                     $handler->flush($flushStamp->force());
                 } catch (\Throwable $e) {
-                    $exceptions[] = $e;
+                    $exceptions[$stamp->getHandlerDescriptor()->getName()] = $e;
                 }
             }
         }


src/Symfony/Component/Messenger/Exception/HandlerFailedException.php
```diff
@@ -19,7 +19,7 @@ class HandlerFailedException extends RuntimeException
     private Envelope $envelope;
 
     /**
-     * @param \Throwable[] $exceptions
+     * @param array<string,\Throwable> $exceptions
```
nicolas-grekas:
```suggestion
     * @param array<string, \Throwable> $exceptions The name of the handler should be given as key
```

src/Symfony/Component/Messenger/Exception/HandlerFailedException.php
```diff
@@ -46,7 +46,7 @@ public function getEnvelope(): Envelope
     }
 
     /**
-     * @return \Throwable[]
+     * @return array<string,\Throwable>
```
nicolas-grekas: similar

nicolas-grekas (commented):
can this be tested somehow?

kbond:
> can this be tested somehow?

Test added.

src/Symfony/Component/Messenger/Tests/Middleware/HandleMessageMiddlewareTest.php
```diff
@@ -335,6 +360,14 @@ public function __invoke()
     }
 }
 
+class FailingHandleMessageMiddlewareTestCallable
```
kbond: Using an anonymous class in the test causes the array key to be a binary string - is this a problem?

fabpot:
Should be rebased (see #51653)

kbond:
Rebased.

fabpot (approved):

kbond:
Psalm errors fixed.

src/Symfony/Component/Messenger/Tests/Exception/HandlerFailedExceptionTest.php
```diff
@@ -97,7 +97,7 @@ public function testThatWrappedExceptionsByClassRecursive()
         $exception2 = new MyOwnException('second');
         $exception3 = new MyOwnException('third');
 
-        $handlerException = new HandlerFailedException($envelope, [$exception1, $exception2, new DelayedMessageHandlingException([$exception3])]);
+        $handlerException = new HandlerFailedException($envelope, [$exception1, $exception2, new DelayedMessageHandlingException([$exception3], $envelope)]);
```
chalasr: done in #51809. Needs another rebase I guess

chalasr:
What about reverting the changes made to the now deprecated `getNestedExceptions()` to update the newly introduced API only?

kbond:
> What about reverting the changes made to the now deprecated getNestedExceptions() to update the newly introduced API only?

You mean throw an `array_values()` into `getNestedExceptions()` before returning?

---

Another thing, it would be nice if `DelayedMessageHandlingException` (which implements this new interface) could provide handlers names as array keys as well but.. I don't think this is possible?

chalasr:
> You mean throw an array_values() into getNestedExceptions() before returning?

It's probably not a big deal here but yes, we usually avoid changing deprecated code paths.

> Another thing, it would be nice if DelayedMessageHandlingException (which implements this new interface) could provide handlers names as array keys as well but.. I don't think this is possible?

I think it would be possible to use `exception->getTrace()[0]['class']` as keys in `DispatchAfterCurrentBusMiddleware`. 
I may be missing some edge cases though e.g. if the exception wasn't thrown by the handler or if the handler callable has no class (if that's even possible?)

kbond:
> It's probably not a big deal here but yes, we usually avoid changing deprecated code paths.

@chalasr, I've made this change.

> I think it would be possible to use exception->getTrace()[0]['class'] as keys in DispatchAfterCurrentBusMiddleware.
I may be missing some edge cases though e.g. if the exception wasn't thrown by the handler or if the handler callable has no class (if that's even possible?)

I'm going to leave this off of this PR because of the possible edge cases you describe. This can be revisited later if desired.

src/Symfony/Component/Messenger/Exception/WrappedExceptionsInterface.php
```diff
@@ -19,7 +19,7 @@
 interface WrappedExceptionsInterface
 {
     /**
-     * @return \Throwable[]
+     * @return array<array-key, \Throwable>
```
stof: this is exactly what `\Throwable[]` means. And `array-key` is not a PHP type so it might not be understood by tools that don't support phpstan/psalm special pseudo types (`array-key` is actually an alias for `int|string` in those tools)

src/Symfony/Component/Messenger/Exception/HandlerFailedException.php
```diff
@@ -20,7 +20,7 @@ class HandlerFailedException extends RuntimeException implements WrappedExceptio
     private Envelope $envelope;
 
     /**
-     * @param \Throwable[] $exceptions
+     * @param array<string, \Throwable> $exceptions The name of the handler should be given as key
```
nicolas-grekas: `WrappedExceptionsInterface` needs a similar update I guess, and `DelayedMessageHandlingException` then needs to also set the handler as key.

nicolas-grekas: Oh, and if it's a "should", then it means the type should be `array<\Throwable>`

src/Symfony/Component/Messenger/Tests/Middleware/HandleMessageMiddlewareTest.php
nicolas-grekas: it shouldn't, does it?

src/Symfony/Component/Messenger/Exception/HandlerFailedException.php
kbond: Should I apply @chalasr's [solution](https://github.com/symfony/symfony/pull/51331#issuecomment-1743650831) to `DelayedMessageHandlingException` then?

src/Symfony/Component/Messenger/Tests/Middleware/HandleMessageMiddlewareTest.php
kbond: I thought I had an exception related to this but just tried now and all's good.

src/Symfony/Component/Messenger/Exception/HandlerFailedException.php
nicolas-grekas: WDYM? getWrappedExceptions is new to 6.4

kbond: There isn't an easy way to get the handler name to use as the key for `DelayedMessageHandlingException`. The handlers aren't available when creating this exception.

nicolas-grekas: ok, thanks for checking

the type should then definitely be changed:
```suggestion
     * @param array<\Throwable> $exceptions The name of the handler should be given as key
```

src/Symfony/Component/Messenger/Exception/HandlerFailedException.php
```diff
@@ -20,7 +20,7 @@ class HandlerFailedException extends RuntimeException implements WrappedExceptio
     private Envelope $envelope;
 
     /**
-     * @param \Throwable[] $exceptions
+     * @param array<\Throwable> $exceptions The name of the handler should be given as key
```
stof: you should keep using `\Throwable[]` as that's the notation we use in Symfony (both notations are the same type)

kbond: I agree, @nicolas-grekas?

nicolas-grekas: I always wonder if `list<foo>` or `array<foo>` is the equivalent of `foo[]`

kbond: > I always wonder if list<foo> or array<foo> is the equivalent of foo[]

Me too but `foo[]` could also be `\Traversable<foo>` I think.

stof: in all SA tools out there, `foo[]` means `array<foo>` (if we forget the weird hack that exist in phpstan where a union `A||foo[]` gets treated as `A<foo>` but only for the `[]` syntax for legacy reasons to account for a weird convention from before the generic syntax was introduced)

nicolas-grekas: let's revert to `[]` @kbond

kbond: Done.

nicolas-grekas:
Thank you @kbond.

PR #51336, opened 2023-08-10
[Notifier] [Pushover] Fix invalid method call + improve exception message

ahmedghanem00:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

Current code is calling `getContent()` method on an array instead of the real response object


nicolas-grekas:
Thank you @ahmedghanem00.

PR #51343, opened 2023-08-10
[HttpFoundation] Add `HeaderRequestMatcher`

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | _NA_
| License       | MIT
| Doc PR        | Todo

This a follow up to:
- https://github.com/symfony/symfony/pull/51324

After @norkunas' [comment](https://github.com/symfony/symfony/pull/51324#issuecomment-1673145964)

### DIFF
--- a/src/Symfony/Component/HttpFoundation/RequestMatcher/HeaderRequestMatcher.php
+++ b/src/Symfony/Component/HttpFoundation/RequestMatcher/HeaderRequestMatcher.php
@@ -1,0 +1,52 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\HttpFoundation\RequestMatcher;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestMatcherInterface;
+
+/**
+ * Checks the presence of HTTP headers in a Request.
+ *
+ * @author Alexandre Daubois <alex.daubois@gmail.com>
+ */
+class HeaderRequestMatcher implements RequestMatcherInterface
+{
+    /**
+     * @var string[]
+     */
+    private array $headers;
+
+    /**
+     * @param string[]|string $headers A header or a list of headers
+     *                                 Strings can contain a comma-delimited list of headers
+     */
+    public function __construct(array|string $headers)
+    {
+        $this->headers = array_reduce((array) $headers, static fn (array $headers, string $header) => array_merge($headers, preg_split('/\s*,\s*/', $header)), []);
+    }
+
+    public function matches(Request $request): bool
+    {
+        if (!$this->headers) {
+            return true;
+        }
+
+        foreach ($this->headers as $header) {
+            if (!$request->headers->has($header)) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+}
--- a/src/Symfony/Component/HttpFoundation/CHANGELOG.md
+++ b/src/Symfony/Component/HttpFoundation/CHANGELOG.md
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Add `UploadedFile::getClientOriginalPath()`
  * Add `QueryParameterRequestMatcher`
+ * Add `HeaderRequestMatcher`
 
 7.0
 ---
--- a/src/Symfony/Component/HttpFoundation/Tests/RequestMatcher/HeaderRequestMatcherTest.php
+++ b/src/Symfony/Component/HttpFoundation/Tests/RequestMatcher/HeaderRequestMatcherTest.php
@@ -1,0 +1,84 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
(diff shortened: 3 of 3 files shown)


src/Symfony/Component/HttpFoundation/RequestMatcher/HeaderRequestMatcher.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public function matches(Request $request): bool
+    {
+        if (!$this->headers) {
+            return true;
+        }
+
+        foreach ($this->headers as $header) {
+            if (!\in_array($header, $request->headers->keys(), true)) {
```
norkunas: Maybe:

```suggestion
            if (!$request->headers->has($header)) {
```
? 🙂

alexandre-daubois: Much better!

src/Symfony/Component/HttpFoundation/RequestMatcher/HeaderRequestMatcher.php
```diff
@@ -0,0 +1,52 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\HttpFoundation\RequestMatcher;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestMatcherInterface;
+
+/**
+ * Checks the HTTP headers of a Request.
```
derrabus: Can we be more precise here? After reading the code, I understood that the class matches request where a given header is simply present. We should point out that we cannot match on the value of headers with this class.

src/Symfony/Component/HttpFoundation/RequestMatcher/HeaderRequestMatcher.php
```diff
@@ -0,0 +1,52 @@
...
+class HeaderRequestMatcher implements RequestMatcherInterface
+{
+    /**
+     * @var string[]
+     */
+    private array $headers;
+
+    /**
+     * @param string[]|string $headers A header or a list of headers
+     *                                 Strings can contain a comma-delimited list of headers
```
derrabus, line 31: Should we really go down that rabbit hole of splitting strings? What would be the use-case?

derrabus (changes requested):

src/Symfony/Component/HttpFoundation/RequestMatcher/HeaderRequestMatcher.php
alexandre-daubois: That's right, I updated the description :slightly_smiling_face:

src/Symfony/Component/HttpFoundation/RequestMatcher/HeaderRequestMatcher.php
alexandre-daubois: Actually, it is to keep consistency with other request matchers, as many other ones accept a comma delimited list

src/Symfony/Component/HttpFoundation/RequestMatcher/HeaderRequestMatcher.php
```diff
@@ -0,0 +1,52 @@
...
+     */
+    public function __construct(array|string $headers)
+    {
+        $this->headers = array_reduce(array_map(strtolower(...), (array) $headers), static fn (array $headers, string $header) => array_merge($headers, preg_split('/\s*,\s*/', $header)), []);
```
norkunas: `strtolower` is not needed I think as request header checking matches case insensitive

alexandre-daubois: You're right! Removed :slightly_smiling_face:

alexandre-daubois:
Status: Needs Review

src/Symfony/Component/HttpFoundation/RequestMatcher/HeaderRequestMatcher.php
GromNaN: The tests on splitting strings are missing.

GromNaN (changes requested):

alexandre-daubois:
That's right! I added the test here and also in https://github.com/symfony/symfony/pull/51324

src/Symfony/Component/HttpFoundation/RequestMatcher/HeaderRequestMatcher.php
alexandre-daubois: That's right! I added the test here and also in https://github.com/symfony/symfony/pull/51324

alexandre-daubois:
Rebased and conflicts fixed

stof (approved):

94noni (approved):

GromNaN (approved):

alexandre-daubois:
Rebased and conflicts fixed :slightly_smiling_face: (as well as [this other related PR](https://github.com/symfony/symfony/pull/51324))

OskarStark (approved):

fabpot (approved):

fabpot:
Thank you @alexandre-daubois.

PR #51345, opened 2023-08-10
[AssetMapper] Fixing bug where a circular exception could be thrown while making error message

### LINKED ISSUE #51291: [AssetMapper] Circular reference error because IMPORT_PATTERN regexp matches comments as well

6.3.2

### Description

I'm trying to integrate AssetMapper within one of my projects, and I'm stuck because I get a circular reference error.

```
In MappedAssetFactory.php line 39:
                                                                                                 
  Circular reference detected while creating asset for "js/htmx.js": "js/htmx.js -> js/htmx.js"
```

### How to reproduce

The issue shows with htmx.js library, as it has the following style of comments within the code:

```js
/**
 *
 * @param {HTMLElement} elt
 * @param {string} swapInfoOverride
 * @returns {import("./htmx").HtmxSwapSpecification}
 */
function getSwapSpecification(elt, swapInfoOverride) {
```

The import syntax within the return annotations trips up the import pattern regexp causing a false positive when checking for circular references.

### Possible Solution

Short of writing/integratng a proper javascript parser, the only quicker workaround  I can think of right now would be to use the `PREG_OFFSET_CAPTURE` flag for the replace regexp in JavaScriptImportPathCompiler.php, and use that offset information to slice a context around the position in order to deduce if the the match is within a comment or string (?) context (requiering the implementation of these two type of parsers).

### Additional Context

As a workaround, if it is useful for other users landing on this issue, I have written a small js script to strip out all the comments from a JavaScript file https://gist.github.com/mhitza/ddc8326e1c94094bdab762e6171e459a

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | none
| Tickets       | Fix #51291
| License       | MIT
| Doc PR        | Not needed

AssetMapper's `JavaScriptImportPathCompiler` parses import statements. That process is imperfect and will over-match in some cases (e.g. matching `import()` that is commented-out). but that's not a huge issue: any matches are simply added to the importmap and matches for not-found-files are ignored.

However, in #51291, we hit a spot where, while trying to improve the log message (`Try adding ".js" to the end of the import - i.e. "%s.js"`), we triggered a circular exception. This PR suppresses that.

We may need to improve the parsing logic later to handle more edge-cases, but we'll handle those if/when they come up.

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/Tests/Compiler/JavaScriptImportPathCompilerTest.php
+++ b/src/Symfony/Component/AssetMapper/Tests/Compiler/JavaScriptImportPathCompilerTest.php
@@ -16,6 +16,7 @@ use Psr\Log\LoggerInterface;
 use Symfony\Component\AssetMapper\AssetMapperInterface;
 use Symfony\Component\AssetMapper\Compiler\AssetCompilerInterface;
 use Symfony\Component\AssetMapper\Compiler\JavaScriptImportPathCompiler;
+use Symfony\Component\AssetMapper\Exception\CircularAssetsException;
 use Symfony\Component\AssetMapper\Exception\RuntimeException;
 use Symfony\Component\AssetMapper\MappedAsset;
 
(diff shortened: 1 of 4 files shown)


src/Symfony/Component/AssetMapper/Tests/Compiler/JavaScriptImportPathCompilerTest.php
```diff
@@ -277,6 +278,31 @@ public static function provideMissingImportModeTests(): iterable
...
+                if ('htmx' === $logicalPath) {
+                    return null;
+                }
+
+                if ('htmx.js' === $logicalPath) {
+                    throw new CircularAssetsException();
+                }
+            });
+
+        $asset = new MappedAsset('html.js', '/path/to/app.js');
```
mhitza: I think there's a typo here, shouldn't it be `'htmx.js'`?

weaverryan: Yup - it doesn't matter for the test, but definitely wasn't my intention. Fixing it - thanks :)

stof: Well, with `html.js`, the test would not trigger a circular error, and so would not prevent regressions. So it definitely matters.

weaverryan: No - the circular reference is triggered a bit manually on the mock. So you're correct in a real-world example, but in this case, this part doesn't matter (creating a test that resulted in a natural, real-world circular reference was going to be a lot more work as we need a real AssetMapper, MappedAssetFactory, etc.

nicolas-grekas:
Thank you @weaverryan.

PR #51346, opened 2023-08-10
Prevent running multiple times

NKoonen:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

This prevents the command from running twice.


stof:
I'm actually -1 for that. Being able to start multiple consumer is a totally valid use case.

NKoonen:
> I'm actually -1 for that. Being able to start multiple consumer is a totally valid use case.

Fair. But it's also a valid use case not to start the consumer multiple times. Currently that's not a possible, would it be a better solution to add this as an option to the command?

GromNaN:
Even if an option is added, it will have to handle the options `receivers`/`bus`/`queues` to apply the lock on the specified consumers. A semaphore would be more appropriate to number instead of limit of 1.

If the concurrency cannot be managed by your orchestrator, a last solution is to create a new command in your application that adds this lock/semaphore with your own logic.

PR #51348, opened 2023-08-10
[FrameworkBundle][Validator] Allow implementing validation groups provider outside DTOs

yceruto:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | https://github.com/symfony/symfony-docs/pull/18744

Alternative to https://github.com/symfony/symfony/pull/51233
Inspiration: https://github.com/symfony/symfony/discussions/51012

Currently, you can determine the sequence of groups to apply dynamically based on the state of your DTO by implementing the `GroupSequenceProviderInterface` in your DTO class. https://symfony.com/doc/current/validation/sequence_provider.html#group-sequence-providers
```php
use Symfony\Component\Validator\GroupSequenceProviderInterface;

#[Assert\GroupSequenceProvider]
class UserDto implements GroupSequenceProviderInterface
{
    // ...

    public function getGroupSequence(): array|GroupSequence
    {
        if ($this->isCompanyType()) {
            return ['User', 'Company'];
        }

        return ['User'];
    }
}
```
It covers most of the common scenarios, but for more advanced ones, it may not be sufficient. Suppose now you need to provide the sequence of groups from an external configuration (or service) which can change its value dynamically:
```php
#[Assert\GroupSequenceProvider]
class UserDto implements GroupSequenceProviderInterface
{
    // ...

    public __constructor(private readonly ConfigService $config) 
    {
    }

    public function getGroupSequence(): array|GroupSequence
    {
        if ($this->config->isEnabled()) {
            return ['User', $this->config->getGroup()];
        }

        return ['User'];
    }
}
```
This issue cannot be resolved at present without managing the DTO initialization and manually setting its dependencies. On the other hand, since the state of the DTO is not used at all, the implementation of the `GroupSequenceProviderInterface` becomes less fitting to the DTO responsibility. Further, stricter programming may raise a complaint about a violation of SOLID principles here.

So, the proposal of this PR is to allow configuring the validation groups provider outside of the DTO, while simultaneously enabling the registration of this provider as a service if necessary.

To achieve this, you'll need to implement a new `GroupProviderInterface` in a separate class, and configure it using the new `provider` option within the `GroupSequenceProvider` attribute:
```php
#[Assert\GroupSequenceProvider(provider: UserGroupProvider::class)]
class UserDto
{
    // ...
}

class UserGroupProvider implements GroupProviderInterface
{
    public __constructor(private readonly ConfigService $config) 
    {
    }

    public function getGroups(object $object): array|GroupSequence
    {
        if ($this->config->isEnabled()) {
            return ['User', $this->config->getGroup()];
        }

        return ['User'];
    }
}
```
That's all you'll need to do if autowiring is enabled under your custom provider. Otherwise, you can manually tag your service with `validator.group_provider` to collect it and utilize it as a provider service during the validation process.

In conclusion, no more messing with the DTO structure, just use the new `class` option for more advanced use cases.

---

TODO:

- [x] Add tests
- [x] Create doc PR

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
@@ -177,6 +177,7 @@ use Symfony\Component\Uid\Factory\UuidFactory;
 use Symfony\Component\Uid\UuidV4;
 use Symfony\Component\Validator\Constraints\ExpressionLanguageProvider;
 use Symfony\Component\Validator\ConstraintValidatorInterface;
+use Symfony\Component\Validator\GroupProviderInterface;
 use Symfony\Component\Validator\Mapping\Loader\PropertyInfoLoader;
 use Symfony\Component\Validator\ObjectInitializerInterface;
 use Symfony\Component\Validator\Validation;
@@ -657,6 +658,8 @@ class FrameworkExtension extends Extension
             ->addTag('serializer.normalizer');
         $container->registerForAutoconfiguration(ConstraintValidatorInterface::class)
             ->addTag('validator.constraint_validator');
+        $container->registerForAutoconfiguration(GroupProviderInterface::class)
+            ->addTag('validator.group_provider');
         $container->registerForAutoconfiguration(ObjectInitializerInterface::class)
             ->addTag('validator.initializer');
         $container->registerForAutoconfiguration(MessageHandlerInterface::class)
--- a/src/Symfony/Component/Validator/Constraints/GroupSequenceProvider.php
+++ b/src/Symfony/Component/Validator/Constraints/GroupSequenceProvider.php
@@ -11,10 +11,15 @@
 
 namespace Symfony\Component\Validator\Constraints;
 
+use Doctrine\Common\Annotations\Annotation\NamedArgumentConstructor;
+use Symfony\Component\Validator\Attribute\HasNamedArguments;
+
 /**
  * Attribute to define a group sequence provider.
  *
  * @Annotation
+ *
+ * @NamedArgumentConstructor
  *
  * @Target({"CLASS", "ANNOTATION"})
  *
@@ -23,4 +28,8 @@ namespace Symfony\Component\Validator\Constraints;
 #[\Attribute(\Attribute::TARGET_CLASS)]
 class GroupSequenceProvider
 {
+    #[HasNamedArguments]
+    public function __construct(public ?string $provider = null)
+    {
+    }
 }
--- a/src/Symfony/Component/Validator/Validator/RecursiveContextualValidator.php
+++ b/src/Symfony/Component/Validator/Validator/RecursiveContextualValidator.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Component\Validator\Validator;
 
+use Psr\Container\ContainerInterface;
 use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\Constraints\Composite;
 use Symfony\Component\Validator\Constraints\Existence;
@@ -50,13 +51,14 @@ class RecursiveContextualValidator implements ContextualValidatorInterface
     private MetadataFactoryInterface $metadataFactory;
     private ConstraintValidatorFactoryInterface $validatorFactory;
     private array $objectInitializers;
+    private ?ContainerInterface $groupProviderLocator;
 
     /**
      * Creates a validator for the given context.
      *
      * @param ObjectInitializerInterface[] $objectInitializers The object initializers
      */
-    public function __construct(ExecutionContextInterface $context, MetadataFactoryInterface $metadataFactory, ConstraintValidatorFactoryInterface $validatorFactory, array $objectInitializers = [])
+    public function __construct(ExecutionContextInterface $context, MetadataFactoryInterface $metadataFactory, ConstraintValidatorFactoryInterface $validatorFactory, array $objectInitializers = [], ContainerInterface $groupProviderLocator = null)
     {
         $this->context = $context;
         $this->defaultPropertyPath = $context->getPropertyPath();
@@ -64,6 +66,7 @@ class RecursiveContextualValidator implements ContextualValidatorInterface
         $this->metadataFactory = $metadataFactory;
         $this->validatorFactory = $validatorFactory;
         $this->objectInitializers = $objectInitializers;
+        $this->groupProviderLocator = $groupProviderLocator;
     }
 
     public function atPath(string $path): static
@@ -436,10 +439,18 @@ class RecursiveContextualValidator implements ContextualValidatorInterface
                     $group = $metadata->getGroupSequence();
                     $defaultOverridden = true;
                 } elseif ($metadata->isGroupSequenceProvider()) {
-                    // The group sequence is dynamically obtained from the validated
-                    // object
-                    /* @var \Symfony\Component\Validator\GroupSequenceProviderInterface $object */
-                    $group = $object->getGroupSequence();
+                    if (null !== $provider = $metadata->getGroupProvider()) {
+                        if (null === $this->groupProviderLocator) {
+                            throw new \LogicException('A group provider locator is required when using group provider.');
+                        }
+
+                        $group = $this->groupProviderLocator->get($provider)->getGroups($object);
+                    } else {
+                        // The group sequence is dynamically obtained from the validated
+                        // object
+                        /* @var \Symfony\Component\Validator\GroupSequenceProviderInterface $object */
+                        $group = $object->getGroupSequence();
+                    }
                     $defaultOverridden = true;
 
                     if (!$group instanceof GroupSequence) {
--- a/src/Symfony/Component/Validator/ValidatorBuilder.php
+++ b/src/Symfony/Component/Validator/ValidatorBuilder.php
@@ -15,6 +15,7 @@ use Doctrine\Common\Annotations\AnnotationReader;
 use Doctrine\Common\Annotations\PsrCachedReader;
 use Doctrine\Common\Annotations\Reader;
 use Psr\Cache\CacheItemPoolInterface;
+use Psr\Container\ContainerInterface;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
 use Symfony\Component\Validator\Context\ExecutionContextFactory;
 use Symfony\Component\Validator\Exception\LogicException;
@@ -53,6 +54,7 @@ class ValidatorBuilder
     private bool $enableAttributeMapping = false;
     private ?MetadataFactoryInterface $metadataFactory = null;
     private ConstraintValidatorFactoryInterface $validatorFactory;
+    private ?ContainerInterface $groupProviderLocator = null;
     private ?CacheItemPoolInterface $mappingCache = null;
     private ?TranslatorInterface $translator = null;
     private ?string $translationDomain = null;
@@ -313,6 +315,16 @@ class ValidatorBuilder
     }
 
     /**
+     * @return $this
+     */
+    public function setGroupProviderLocator(ContainerInterface $groupProviderLocator): static
+    {
+        $this->groupProviderLocator = $groupProviderLocator;
+
+        return $this;
+    }
+
+    /**
      * Sets the translator used for translating violation messages.
      *
      * @return $this
@@ -414,7 +426,7 @@ class ValidatorBuilder
 
         $contextFactory = new ExecutionContextFactory($translator, $this->translationDomain);
 
-        return new RecursiveValidator($contextFactory, $metadataFactory, $validatorFactory, $this->initializers);
+        return new RecursiveValidator($contextFactory, $metadataFactory, $validatorFactory, $this->initializers, $this->groupProviderLocator);
     }
 
     private function createAnnotationReader(): Reader
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/UnusedTagsPass.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/UnusedTagsPass.php
@@ -101,6 +101,7 @@ class UnusedTagsPass implements CompilerPassInterface
         'twig.runtime',
         'validator.auto_mapper',
         'validator.constraint_validator',
+        'validator.group_provider',
         'validator.initializer',
         'workflow',
     ];
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/validator.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/validator.php
@@ -42,6 +42,9 @@ return static function (ContainerConfigurator $container) {
             ->call('setConstraintValidatorFactory', [
                 service('validator.validator_factory'),
             ])
+            ->call('setGroupProviderLocator', [
+                tagged_locator('validator.group_provider'),
+            ])
             ->call('setTranslator', [
                 service('translator')->ignoreOnInvalid(),
             ])
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
@@ -1230,16 +1230,18 @@ abstract class FrameworkExtensionTestCase extends TestCase
 
         $annotations = !class_exists(FullStack::class);
 
-        $this->assertCount($annotations ? 7 : 6, $calls);
+        $this->assertCount($annotations ? 8 : 7, $calls);
         $this->assertSame('setConstraintValidatorFactory', $calls[0][0]);
         $this->assertEquals([new Reference('validator.validator_factory')], $calls[0][1]);
-        $this->assertSame('setTranslator', $calls[1][0]);
-        $this->assertEquals([new Reference('translator', ContainerBuilder::IGNORE_ON_INVALID_REFERENCE)], $calls[1][1]);
-        $this->assertSame('setTranslationDomain', $calls[2][0]);
-        $this->assertSame(['%validator.translation_domain%'], $calls[2][1]);
-        $this->assertSame('addXmlMappings', $calls[3][0]);
-        $this->assertSame([$xmlMappings], $calls[3][1]);
-        $i = 3;
+        $this->assertSame('setGroupProviderLocator', $calls[1][0]);
+        $this->assertInstanceOf(ServiceLocatorArgument::class, $calls[1][1][0]);
+        $this->assertSame('setTranslator', $calls[2][0]);
+        $this->assertEquals([new Reference('translator', ContainerBuilder::IGNORE_ON_INVALID_REFERENCE)], $calls[2][1]);
+        $this->assertSame('setTranslationDomain', $calls[3][0]);
+        $this->assertSame(['%validator.translation_domain%'], $calls[3][1]);
+        $this->assertSame('addXmlMappings', $calls[4][0]);
+        $this->assertSame([$xmlMappings], $calls[4][1]);
+        $i = 4;
         if ($annotations) {
             $this->assertSame('enableAttributeMapping', $calls[++$i][0]);
         }
@@ -1288,12 +1290,12 @@ abstract class FrameworkExtensionTestCase extends TestCase
 
         $calls = $container->getDefinition('validator.builder')->getMethodCalls();
 
-        $this->assertCount(7, $calls);
-        $this->assertSame('enableAttributeMapping', $calls[4][0]);
-        $this->assertSame('addMethodMapping', $calls[5][0]);
-        $this->assertSame(['loadValidatorMetadata'], $calls[5][1]);
-        $this->assertSame('setMappingCache', $calls[6][0]);
-        $this->assertEquals([new Reference('validator.mapping.cache.adapter')], $calls[6][1]);
+        $this->assertCount(8, $calls);
+        $this->assertSame('enableAttributeMapping', $calls[5][0]);
+        $this->assertSame('addMethodMapping', $calls[6][0]);
+        $this->assertSame(['loadValidatorMetadata'], $calls[6][1]);
+        $this->assertSame('setMappingCache', $calls[7][0]);
+        $this->assertEquals([new Reference('validator.mapping.cache.adapter')], $calls[7][1]);
         // no cache this time
     }
 
@@ -1308,14 +1310,14 @@ abstract class FrameworkExtensionTestCase extends TestCase
 
         $calls = $container->getDefinition('validator.builder')->getMethodCalls();
 
-        $this->assertCount(8, $calls);
-        $this->assertSame('enableAttributeMapping', $calls[4][0]);
+        $this->assertCount(9, $calls);
+        $this->assertSame('enableAttributeMapping', $calls[5][0]);
         if (method_exists(ValidatorBuilder::class, 'setDoctrineAnnotationReader')) {
-            $this->assertSame('setDoctrineAnnotationReader', $calls[5][0]);
-            $this->assertEquals([new Reference('annotation_reader')], $calls[5][1]);
-            $i = 6;
+            $this->assertSame('setDoctrineAnnotationReader', $calls[6][0]);
+            $this->assertEquals([new Reference('annotation_reader')], $calls[6][1]);
+            $i = 7;
         } else {
-            $i = 5;
+            $i = 6;
         }
         $this->assertSame('addMethodMapping', $calls[$i][0]);
         $this->assertSame(['loadValidatorMetadata'], $calls[$i][1]);
@@ -1335,16 +1337,16 @@ abstract class FrameworkExtensionTestCase extends TestCase
 
         $calls = $container->getDefinition('validator.builder')->getMethodCalls();
 
-        $this->assertCount(8, $calls);
-        $this->assertSame('addXmlMappings', $calls[3][0]);
-        $this->assertSame('addYamlMappings', $calls[4][0]);
-        $this->assertSame('enableAttributeMapping', $calls[5][0]);
-        $this->assertSame('addMethodMapping', $calls[6][0]);
-        $this->assertSame(['loadValidatorMetadata'], $calls[6][1]);
-        $this->assertSame('setMappingCache', $calls[7][0]);
-        $this->assertEquals([new Reference('validator.mapping.cache.adapter')], $calls[7][1]);
+        $this->assertCount(9, $calls);
+        $this->assertSame('addXmlMappings', $calls[4][0]);
+        $this->assertSame('addYamlMappings', $calls[5][0]);
+        $this->assertSame('enableAttributeMapping', $calls[6][0]);
+        $this->assertSame('addMethodMapping', $calls[7][0]);
+        $this->assertSame(['loadValidatorMetadata'], $calls[7][1]);
+        $this->assertSame('setMappingCache', $calls[8][0]);
+        $this->assertEquals([new Reference('validator.mapping.cache.adapter')], $calls[8][1]);
 
-        $xmlMappings = $calls[3][1][0];
+        $xmlMappings = $calls[4][1][0];
         $this->assertCount(3, $xmlMappings);
         try {
             // Testing symfony/symfony
@@ -1355,7 +1357,7 @@ abstract class FrameworkExtensionTestCase extends TestCase
         }
         $this->assertStringEndsWith('TestBundle/Resources/config/validation.xml', $xmlMappings[1]);
 
-        $yamlMappings = $calls[4][1][0];
+        $yamlMappings = $calls[5][1][0];
         $this->assertCount(1, $yamlMappings);
         $this->assertStringEndsWith('TestBundle/Resources/config/validation.yml', $yamlMappings[0]);
     }
@@ -1370,7 +1372,7 @@ abstract class FrameworkExtensionTestCase extends TestCase
         ]);
 
         $calls = $container->getDefinition('validator.builder')->getMethodCalls();
-        $xmlMappings = $calls[3][1][0];
+        $xmlMappings = $calls[4][1][0];
         $this->assertCount(3, $xmlMappings);
 
         try {
@@ -1382,7 +1384,7 @@ abstract class FrameworkExtensionTestCase extends TestCase
         }
         $this->assertStringEndsWith('CustomPathBundle/Resources/config/validation.xml', $xmlMappings[1]);
 
-        $yamlMappings = $calls[4][1][0];
+        $yamlMappings = $calls[5][1][0];
         $this->assertCount(1, $yamlMappings);
         $this->assertStringEndsWith('CustomPathBundle/Resources/config/validation.yml', $yamlMappings[0]);
     }
@@ -1395,9 +1397,9 @@ abstract class FrameworkExtensionTestCase extends TestCase
 
         $annotations = !class_exists(FullStack::class);
 
-        $this->assertCount($annotations ? 6 : 5, $calls);
-        $this->assertSame('addXmlMappings', $calls[3][0]);
-        $i = 3;
+        $this->assertCount($annotations ? 7 : 6, $calls);
+        $this->assertSame('addXmlMappings', $calls[4][0]);
+        $i = 4;
         if ($annotations) {
             $this->assertSame('enableAttributeMapping', $calls[++$i][0]);
         }
@@ -1426,14 +1428,14 @@ abstract class FrameworkExtensionTestCase extends TestCase
 
         $calls = $container->getDefinition('validator.builder')->getMethodCalls();
 
-        $this->assertSame('addXmlMappings', $calls[3][0]);
-        $this->assertCount(3, $calls[3][1][0]);
-
-        $this->assertSame('addYamlMappings', $calls[4][0]);
+        $this->assertSame('addXmlMappings', $calls[4][0]);
         $this->assertCount(3, $calls[4][1][0]);
-        $this->assertStringContainsString('foo.yml', $calls[4][1][0][0]);
-        $this->assertStringContainsString('validation.yml', $calls[4][1][0][1]);
-        $this->assertStringContainsString('validation.yaml', $calls[4][1][0][2]);
+
+        $this->assertSame('addYamlMappings', $calls[5][0]);
+        $this->assertCount(3, $calls[5][1][0]);
+        $this->assertStringContainsString('foo.yml', $calls[5][1][0][0]);
+        $this->assertStringContainsString('validation.yml', $calls[5][1][0][1]);
+        $this->assertStringContainsString('validation.yaml', $calls[5][1][0][2]);
     }
 
     public function testValidationAutoMapping()
--- a/src/Symfony/Component/Validator/CHANGELOG.md
+++ b/src/Symfony/Component/Validator/CHANGELOG.md
@@ -15,6 +15,7 @@ CHANGELOG
  * Deprecate `ValidatorBuilder::enableAnnotationMapping()`, use `ValidatorBuilder::enableAttributeMapping()` instead
  * Deprecate `ValidatorBuilder::disableAnnotationMapping()`, use `ValidatorBuilder::disableAttributeMapping()` instead
  * Deprecate `AnnotationLoader`, use `AttributeLoader` instead
+ * Add `GroupProviderInterface` to implement validation group providers outside the underlying class
 
 6.3
 ---
--- a/src/Symfony/Component/Validator/GroupProviderInterface.php
+++ b/src/Symfony/Component/Validator/GroupProviderInterface.php
@@ -1,0 +1,30 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Validator;
+
+use Symfony\Component\Validator\Constraints\GroupSequence;
+
+/**
+ * Defines the interface for a validation group provider.
+ *
+ * @author Yonel Ceruto <yonelceruto@gmail.com>
+ */
+interface GroupProviderInterface
+{
+    /**
+     * Returns which validation groups should be used for a certain state
+     * of the object.
+     *
+     * @return string[]|string[][]|GroupSequence
+     */
+    public function getGroups(object $object): array|GroupSequence;
+}
--- a/src/Symfony/Component/Validator/Mapping/ClassMetadata.php
+++ b/src/Symfony/Component/Validator/Mapping/ClassMetadata.php
@@ -87,6 +87,13 @@ class ClassMetadata extends GenericMetadata implements ClassMetadataInterface
     public bool $groupSequenceProvider = false;
 
     /**
+     * @internal This property is public in order to reduce the size of the
+     *           class' serialized representation. Do not access it. Use
+     *           {@link getGroupProvider()} instead.
+     */
+    public ?string $groupProvider = null;
+
+    /**
      * The strategy for traversing traversable objects.
      *
      * By default, only instances of {@link \Traversable} are traversed.
@@ -123,6 +130,7 @@ class ClassMetadata extends GenericMetadata implements ClassMetadataInterface
             'getters',
             'groupSequence',
             'groupSequenceProvider',
+            'groupProvider',
             'members',
             'name',
             'properties',
@@ -319,6 +327,7 @@ class ClassMetadata extends GenericMetadata implements ClassMetadataInterface
     public function mergeConstraints(self $source)
     {
         if ($source->isGroupSequenceProvider()) {
+            $this->setGroupProvider($source->getGroupProvider());
             $this->setGroupSequenceProvider(true);
         }
 
@@ -432,7 +441,7 @@ class ClassMetadata extends GenericMetadata implements ClassMetadataInterface
             throw new GroupDefinitionException('Defining a group sequence provider is not allowed with a static group sequence.');
         }
 
-        if (!$this->getReflectionClass()->implementsInterface(GroupSequenceProviderInterface::class)) {
+        if (null === $this->groupProvider && !$this->getReflectionClass()->implementsInterface(GroupSequenceProviderInterface::class)) {
             throw new GroupDefinitionException(sprintf('Class "%s" must implement GroupSequenceProviderInterface.', $this->name));
         }
 
@@ -442,6 +451,16 @@ class ClassMetadata extends GenericMetadata implements ClassMetadataInterface
     public function isGroupSequenceProvider(): bool
     {
         return $this->groupSequenceProvider;
+    }
+
+    public function setGroupProvider(?string $provider): void
+    {
+        $this->groupProvider = $provider;
+    }
+
+    public function getGroupProvider(): ?string
+    {
+        return $this->groupProvider;
     }
 
     public function getCascadingStrategy(): int
--- a/src/Symfony/Component/Validator/Mapping/ClassMetadataInterface.php
+++ b/src/Symfony/Component/Validator/Mapping/ClassMetadataInterface.php
@@ -30,6 +30,8 @@ use Symfony\Component\Validator\GroupSequenceProviderInterface;
  * @see GroupSequence
  * @see GroupSequenceProviderInterface
  * @see TraversalStrategy
+ *
+ * @method string|null getGroupProvider()
  */
 interface ClassMetadataInterface extends MetadataInterface
 {
--- a/src/Symfony/Component/Validator/Mapping/Loader/AnnotationLoader.php
+++ b/src/Symfony/Component/Validator/Mapping/Loader/AnnotationLoader.php
@@ -51,6 +51,7 @@ class AnnotationLoader implements LoaderInterface
             if ($constraint instanceof GroupSequence) {
                 $metadata->setGroupSequence($constraint->groups);
             } elseif ($constraint instanceof GroupSequenceProvider) {
+                $metadata->setGroupProvider($constraint->provider);
                 $metadata->setGroupSequenceProvider(true);
             } elseif ($constraint instanceof Constraint) {
                 $metadata->addConstraint($constraint);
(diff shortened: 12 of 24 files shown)


yceruto:
Update:
* 8.2, high-deps failures are normal until the PR gets merged into 7.0
* appveyor failure is unrelated
* Psalm warning are okay, nothing to do there.

This is ready for review

src/Symfony/Component/Validator/Validator/RecursiveContextualValidator.php
```diff
@@ -436,10 +440,14 @@ private function validateClassNode(object $object, ?string $cacheKey, ClassMetad
                     $group = $metadata->getGroupSequence();
                     $defaultOverridden = true;
                 } elseif ($metadata->isGroupSequenceProvider()) {
-                    // The group sequence is dynamically obtained from the validated
-                    // object
-                    /* @var \Symfony\Component\Validator\GroupSequenceProviderInterface $object */
-                    $group = $object->getGroupSequence();
+                    if (null !== $providerClass = $metadata->getGroupSequenceProviderClass()) {
+                        $group = $this->groupSequenceProviderFactory->getInstance($providerClass)->getGroupSequence();
```
ro0NL: what about passing `$object` to `getGroupSequence`, for even more dynamic group computation.

this is what eg. API platform currently does already: https://api-platform.com/docs/core/validation/#dynamic-validation-groups

yceruto: Good idea! I will update accordingly, thanks!

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -669,6 +670,8 @@ public function load(array $configs, ContainerBuilder $container)
             ->addTag('serializer.normalizer');
         $container->registerForAutoconfiguration(ConstraintValidatorInterface::class)
             ->addTag('validator.constraint_validator');
+        $container->registerForAutoconfiguration(GroupSequenceProviderInterface::class)
```
stof: Using the same interface for external provider and self-provided implementation has a big drawback: all your DTOs implementing the interface for self usage will be collected with this tag and kept in the container, forcing you to manually exclude them if you don't want to face issues where the method would be called on a dummy instance instead of the DTO itself.

yceruto: Let me know if I'm missing something, but without specifying the new `class` option, it's not possible to call an external provider or any other than the one defined in the `class` option. The validator should always call the underlying DTO method in that cases.

In the other hand, there shouldn't be too many cases to be concerned about regarding the number of useless providers in this container, so I would keep them in favor of better DX.

yceruto: Let me know if still a new interface is better. I understand it's not 100% optimized.

stof: The issue with useless DTOs in the container is that any non-autowirable parameter in the constructor of those DTOs become a compilation error for the container (a removed service does not trigger an autowiring error, but preserved services must be valid).

stof: Anyway, if you want to pass the object to the method as discussed in https://github.com/symfony/symfony/pull/51348#discussion_r1297090934, you will need a different interface.

yceruto: Agreed, let's add a new interface then. Thanks for the feedback!

yceruto:
Introduced `GroupProviderInterface` for DI optimization and adding the underlying object as argument for more dynamic possibilities. PR description updated with more details.

ro0NL:
the next question is if we want to either deprecate `GroupSequenceProviderInterface` in favor of new `GroupProviderInterface`, or want to rename it to eg. `SelfProvidingGroupProviderInterface`, or keep 2 similar concepts

my pref would be deprecating the current interface, in favor of the new more flexible one

yceruto:
> the next question is if we want to either deprecate GroupSequenceProviderInterface in favor of new GroupProviderInterface, or want to rename it to eg. SelfProvidingGroupProviderInterface, or keep 2 similar concepts

@ro0NL I was wondering the same thing, and I still see value in keeping the `GroupSequenceProviderInterface` for the self-provider for simple use-cases. About renaming it, I understand that their names are really close and it's not easy to remember which one should be used in each case without reading the documentation, so I'd be in favor of doing that in another PR if desired. Not sure about the impact though.

src/Symfony/Component/Validator/ValidatorBuilder.php
```diff
@@ -287,6 +288,13 @@ public function setConstraintValidatorFactory(ConstraintValidatorFactoryInterfac
         return $this;
     }
 
+    public function setGroupProviderFactory(GroupProviderFactoryInterface $groupProviderFactory): static
```
nicolas-grekas:
```suggestion
    /**
     * @return $this
     */
    public function setGroupProviderFactory(GroupProviderFactoryInterface $groupProviderFactory): static
```

src/Symfony/Component/Validator/GroupProviderFactory.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+class GroupProviderFactory implements GroupProviderFactoryInterface
+{
+    public function __construct(private array $providers = [])
+    {
+    }
+
+    public function getInstance(string $class): GroupProviderInterface
+    {
+        return $this->providers[$class] ??= new $class();
```
nicolas-grekas: Can't we get rid of this class + interface and rely on `ContainerInterface` instead?
Also on this line, I'd prefer not doing any `new` and instead always require a service.
This would remove the requirement for the key to be a class name, which would be better IMHO (allows targetting a different service of the same class if needed.)

yceruto: >Can't we get rid of this class + interface and rely on ContainerInterface instead?

I'm not sure to understand your suggestion 😕 here 

>Also on this line, I'd prefer not doing any new and instead always require a service.
This would remove the requirement for the key to be a class name, which would be better IMHO (allows targetting a different service of the same class if needed.)

I believe that requirement exists only when the component is used alone. I copied that strategy from [`ConstraintValidatorFactory`](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Validator/ConstraintValidatorFactory.php), which uses the same approach for class instantiation.

On the other hand, when the component is used in conjunction with the Framework, it's the `ContainerGroupProviderFactory` that will be used by default, allowing targeting of a different service of the same class if needed. Same approach as `ContainerConstraintValidatorFactory`.

src/Symfony/Component/Validator/ValidatorBuilder.php
yceruto: added, thanks!

src/Symfony/Component/Validator/GroupProviderFactory.php
nicolas-grekas: I mean: `GroupProviderInterface` defines a service locator, so let's use `ContainerInterface` instead?
For standalone usage, it's still quite easy to build a locator.

yceruto: I'm delving into it again, but I don't see how we can reference `ContainerInterface` when the component is used outside the framework stack, especially since it's installed (transitively) as a dev requirement, so I'm confuse, could you please elaborate a bit more or provide an example?

Btw, as it stands, this class allows targeting a different service of the same class if desired:
```php
$factory = new GroupProviderFactory([
    'key1' => new MyGroupProvidor($groups1),
    'key2' => new MyGroupProvidor($groups2),
])
```

nicolas-grekas:
Hi @yceruto :wave: 

About my proposal to use a container directly, here is a draft to do so for inspiration. I didn't go up to the end of it but you should be able to grasp the idea. As you'll see, this would remove a lot of code.

<details>

```diff
diff --git a/src/Symfony/Bundle/FrameworkBundle/Resources/config/validator.php b/src/Symfony/Bundle/FrameworkBundle/Resources/config/validator.php
index b188ad7718..adde2de238 100644
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/validator.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/validator.php
@@ -21,7 +21,6 @@ use Symfony\Component\Validator\Constraints\NoSuspiciousCharactersValidator;
 use Symfony\Component\Validator\Constraints\NotCompromisedPasswordValidator;
 use Symfony\Component\Validator\Constraints\WhenValidator;
 use Symfony\Component\Validator\ContainerConstraintValidatorFactory;
-use Symfony\Component\Validator\ContainerGroupProviderFactory;
 use Symfony\Component\Validator\Mapping\Loader\PropertyInfoLoader;
 use Symfony\Component\Validator\Validation;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
@@ -43,8 +42,8 @@ return static function (ContainerConfigurator $container) {
             ->call('setConstraintValidatorFactory', [
                 service('validator.validator_factory'),
             ])
-            ->call('setGroupProviderFactory', [
-                service('validator.group_provider_factory'),
+            ->call('setGroupProviderLocator', [
+                tagged_locator('validator.group_provider'),
             ])
             ->call('setTranslator', [
                 service('translator')->ignoreOnInvalid(),
@@ -73,11 +72,6 @@ return static function (ContainerConfigurator $container) {
                 abstract_arg('Constraint validators locator'),
             ])
 
-        ->set('validator.group_provider_factory', ContainerGroupProviderFactory::class)
-            ->args([
-                abstract_arg('Group provider locator'),
-            ])
-
         ->load('Symfony\Component\Validator\Constraints\\', $validatorsDir.'/*Validator.php')
             ->exclude($validatorsDir.'/ExpressionLanguageSyntaxValidator.php')
             ->abstract()
diff --git a/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php b/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
index 3552b236c4..0bd0af0109 100644
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
@@ -80,7 +80,6 @@ use Symfony\Component\Serializer\Serializer;
 use Symfony\Component\Translation\DependencyInjection\TranslatorPass;
 use Symfony\Component\Translation\LocaleSwitcher;
 use Symfony\Component\Validator\DependencyInjection\AddConstraintValidatorsPass;
-use Symfony\Component\Validator\DependencyInjection\AddGroupProvidersPass;
 use Symfony\Component\Validator\Validation;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
 use Symfony\Component\Validator\ValidatorBuilder;
@@ -2391,7 +2390,7 @@ abstract class FrameworkExtensionTestCase extends TestCase
             $container->getCompilerPassConfig()->setAfterRemovingPasses([]);
         }
         $container->getCompilerPassConfig()->setBeforeOptimizationPasses([new LoggerPass()]);
-        $container->getCompilerPassConfig()->setBeforeRemovingPasses([new AddConstraintValidatorsPass(), new AddGroupProvidersPass(), new TranslatorPass()]);
+        $container->getCompilerPassConfig()->setBeforeRemovingPasses([new AddConstraintValidatorsPass(), new TranslatorPass()]);
         $container->getCompilerPassConfig()->setAfterRemovingPasses([new AddAnnotationsCachedReaderPass()]);
 
         if (!$compile) {
diff --git a/src/Symfony/Component/Validator/Constraints/GroupSequenceProvider.php b/src/Symfony/Component/Validator/Constraints/GroupSequenceProvider.php
index 16d19acee2..f5ae335595 100644
--- a/src/Symfony/Component/Validator/Constraints/GroupSequenceProvider.php
+++ b/src/Symfony/Component/Validator/Constraints/GroupSequenceProvider.php
@@ -23,10 +23,10 @@ namespace Symfony\Component\Validator\Constraints;
 #[\Attribute(\Attribute::TARGET_CLASS)]
 class GroupSequenceProvider
 {
-    public ?string $class = null;
+    public ?string $provider = null;
 
-    public function __construct(array $options = [], string $class = null)
+    public function __construct(array $options = [], string $provider = null)
     {
-        $this->class = $options['class'] ?? $class;
+        $this->provider = $options['provider'] ?? $provider;
     }
 }
diff --git a/src/Symfony/Component/Validator/ContainerGroupProviderFactory.php b/src/Symfony/Component/Validator/ContainerGroupProviderFactory.php
deleted file mode 100644
index ca36e1420e..0000000000
--- a/src/Symfony/Component/Validator/ContainerGroupProviderFactory.php
+++ /dev/null
@@ -1,52 +0,0 @@
-<?php
-
-/*
- * This file is part of the Symfony package.
- *
- * (c) Fabien Potencier <fabien@symfony.com>
- *
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- */
-
-namespace Symfony\Component\Validator;
-
-use Psr\Container\ContainerInterface;
-use Symfony\Component\Validator\Constraints\GroupSequenceProvider;
-use Symfony\Component\Validator\Exception\GroupDefinitionException;
-use Symfony\Component\Validator\Exception\UnexpectedTypeException;
-
-/**
- * Factory class leveraging a PSR-11 container to create or retrieve instances of validation group providers.
- *
- * @author Yonel Ceruto <yonelceruto@gmail.com>
- */
-class ContainerGroupProviderFactory implements GroupProviderFactoryInterface
-{
-    private array $providers = [];
-
-    public function __construct(private readonly ContainerInterface $container)
-    {
-    }
-
-    public function getInstance(string $class): GroupProviderInterface
-    {
-        if (!isset($this->providers[$class])) {
-            if ($this->container->has($class)) {
-                $this->providers[$class] = $this->container->get($class);
-            } else {
-                if (!class_exists($class)) {
-                    throw new GroupDefinitionException(sprintf('Group provider "%s" does not exist or is not enabled as service. Check the "class" option in your constraint class "%s".', $class, GroupSequenceProvider::class));
-                }
-
-                $this->providers[$class] = new $class();
-            }
-        }
-
-        if (!$this->providers[$class] instanceof GroupProviderInterface) {
-            throw new UnexpectedTypeException($this->providers[$class], GroupProviderInterface::class);
-        }
-
-        return $this->providers[$class];
-    }
-}
diff --git a/src/Symfony/Component/Validator/DependencyInjection/AddGroupProvidersPass.php b/src/Symfony/Component/Validator/DependencyInjection/AddGroupProvidersPass.php
deleted file mode 100644
index 899347a333..0000000000
--- a/src/Symfony/Component/Validator/DependencyInjection/AddGroupProvidersPass.php
+++ /dev/null
@@ -1,46 +0,0 @@
-<?php
-
-/*
- * This file is part of the Symfony package.
- *
- * (c) Fabien Potencier <fabien@symfony.com>
- *
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- */
-
-namespace Symfony\Component\Validator\DependencyInjection;
-
-use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
-use Symfony\Component\DependencyInjection\Compiler\ServiceLocatorTagPass;
-use Symfony\Component\DependencyInjection\ContainerBuilder;
-use Symfony\Component\DependencyInjection\Reference;
-
-/**
- * @author Yonel Ceruto <yonelceruto@gmail.com>
- */
-class AddGroupProvidersPass implements CompilerPassInterface
-{
-    public function process(ContainerBuilder $container): void
-    {
-        if (!$container->hasDefinition('validator.group_provider_factory')) {
-            return;
-        }
-
-        $providers = [];
-        foreach ($container->findTaggedServiceIds('validator.group_provider', true) as $id => $attributes) {
-            $definition = $container->getDefinition($id);
-
-            if (isset($attributes[0]['alias'])) {
-                $providers[$attributes[0]['alias']] = new Reference($id);
-            }
-
-            $providers[$definition->getClass()] = new Reference($id);
-        }
-
-        $container
-            ->getDefinition('validator.group_provider_factory')
-            ->replaceArgument(0, ServiceLocatorTagPass::register($container, $providers))
-        ;
-    }
-}
diff --git a/src/Symfony/Component/Validator/GroupProviderFactory.php b/src/Symfony/Component/Validator/GroupProviderFactory.php
deleted file mode 100644
index f908104230..0000000000
--- a/src/Symfony/Component/Validator/GroupProviderFactory.php
+++ /dev/null
@@ -1,29 +0,0 @@
-<?php
-
-/*
- * This file is part of the Symfony package.
- *
- * (c) Fabien Potencier <fabien@symfony.com>
- *
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- */
-
-namespace Symfony\Component\Validator;
-
-/**
- * Factory class for creating or retrieving instances of validation group providers.
- *
- * @author Yonel Ceruto <yonelceruto@gmail.com>
- */
-class GroupProviderFactory implements GroupProviderFactoryInterface
-{
-    public function __construct(private array $providers = [])
-    {
-    }
-
-    public function getInstance(string $class): GroupProviderInterface
-    {
-        return $this->providers[$class] ??= new $class();
-    }
-}
diff --git a/src/Symfony/Component/Validator/GroupProviderFactoryInterface.php b/src/Symfony/Component/Validator/GroupProviderFactoryInterface.php
deleted file mode 100644
index 003547f434..0000000000
--- a/src/Symfony/Component/Validator/GroupProviderFactoryInterface.php
+++ /dev/null
@@ -1,22 +0,0 @@
-<?php
-
-/*
- * This file is part of the Symfony package.
- *
- * (c) Fabien Potencier <fabien@symfony.com>
- *
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- */
-
-namespace Symfony\Component\Validator;
-
-/**
- * Handles the creation of validation group providers.
- *
- * @author Yonel Ceruto <yonelceruto@gmail.com>
- */
-interface GroupProviderFactoryInterface
-{
-    public function getInstance(string $class): GroupProviderInterface;
-}
diff --git a/src/Symfony/Component/Validator/Mapping/ClassMetadata.php b/src/Symfony/Component/Validator/Mapping/ClassMetadata.php
index b5609ec92d..981035ff54 100644
--- a/src/Symfony/Component/Validator/Mapping/ClassMetadata.php
+++ b/src/Symfony/Component/Validator/Mapping/ClassMetadata.php
@@ -89,9 +89,9 @@ class ClassMetadata extends GenericMetadata implements ClassMetadataInterface
     /**
      * @internal This property is public in order to reduce the size of the
      *           class' serialized representation. Do not access it. Use
-     *           {@link getGroupProviderClass()} instead.
+     *           {@link getGroupProvider()} instead.
      */
-    public ?string $groupProviderClass = null;
+    public ?string $groupProvider = null;
 
     /**
      * The strategy for traversing traversable objects.
@@ -130,7 +130,7 @@ class ClassMetadata extends GenericMetadata implements ClassMetadataInterface
             'getters',
             'groupSequence',
             'groupSequenceProvider',
-            'groupProviderClass',
+            'groupProvider',
             'members',
             'name',
             'properties',
@@ -327,7 +327,7 @@ class ClassMetadata extends GenericMetadata implements ClassMetadataInterface
     public function mergeConstraints(self $source)
     {
         if ($source->isGroupSequenceProvider()) {
-            $this->setGroupProviderClass($source->getGroupProviderClass());
+            $this->setGroupProvider($source->getGroupProvider());
             $this->setGroupSequenceProvider(true);
         }
 
@@ -441,10 +441,6 @@ class ClassMetadata extends GenericMetadata implements ClassMetadataInterface
             throw new GroupDefinitionException('Defining a group sequence provider is not allowed with a static group sequence.');
         }
 
-        if (null === $this->groupProviderClass && !$this->getReflectionClass()->implementsInterface(GroupSequenceProviderInterface::class)) {
-            throw new GroupDefinitionException(sprintf('Class "%s" must implement GroupSequenceProviderInterface.', $this->name));
-        }
-
         $this->groupSequenceProvider = $active;
     }
 
@@ -453,14 +449,14 @@ class ClassMetadata extends GenericMetadata implements ClassMetadataInterface
         return $this->groupSequenceProvider;
     }
 
-    public function setGroupProviderClass(?string $class): void
+    public function setGroupProvider(?string $class): void
     {
-        $this->groupProviderClass = $class;
+        $this->groupProvider = $class;
     }
 
-    public function getGroupProviderClass(): ?string
+    public function getGroupProvider(): ?string
     {
-        return $this->groupProviderClass;
+        return $this->groupProvider;
     }
 
     public function getCascadingStrategy(): int
diff --git a/src/Symfony/Component/Validator/Mapping/ClassMetadataInterface.php b/src/Symfony/Component/Validator/Mapping/ClassMetadataInterface.php
index 624b55cc8d..6625e37e60 100644
--- a/src/Symfony/Component/Validator/Mapping/ClassMetadataInterface.php
+++ b/src/Symfony/Component/Validator/Mapping/ClassMetadataInterface.php
@@ -31,7 +31,7 @@ use Symfony\Component\Validator\GroupSequenceProviderInterface;
  * @see GroupSequenceProviderInterface
  * @see TraversalStrategy
  *
- * @method string|null getGroupProviderClass()
+ * @method string|null getGroupProvider()
  */
 interface ClassMetadataInterface extends MetadataInterface
 {
diff --git a/src/Symfony/Component/Validator/Mapping/Loader/AnnotationLoader.php b/src/Symfony/Component/Validator/Mapping/Loader/AnnotationLoader.php
index 41b0662d35..ebc1795f2c 100644
--- a/src/Symfony/Component/Validator/Mapping/Loader/AnnotationLoader.php
+++ b/src/Symfony/Component/Validator/Mapping/Loader/AnnotationLoader.php
@@ -51,7 +51,7 @@ class AnnotationLoader implements LoaderInterface
             if ($constraint instanceof GroupSequence) {
                 $metadata->setGroupSequence($constraint->groups);
             } elseif ($constraint instanceof GroupSequenceProvider) {
-                $metadata->setGroupProviderClass($constraint->class);
+                $metadata->setGroupProvider($constraint->provider);
                 $metadata->setGroupSequenceProvider(true);
             } elseif ($constraint instanceof Constraint) {
                 $metadata->addConstraint($constraint);
diff --git a/src/Symfony/Component/Validator/Mapping/Loader/XmlFileLoader.php b/src/Symfony/Component/Validator/Mapping/Loader/XmlFileLoader.php
index a68f37009b..94d3f071e5 100644
--- a/src/Symfony/Component/Validator/Mapping/Loader/XmlFileLoader.php
+++ b/src/Symfony/Component/Validator/Mapping/Loader/XmlFileLoader.php
@@ -201,7 +201,7 @@ class XmlFileLoader extends FileLoader
     private function loadClassMetadataFromXml(ClassMetadata $metadata, \SimpleXMLElement $classDescription): void
     {
         if (\count($classDescription->{'group-sequence-provider'}) > 0) {
-            $metadata->setGroupProviderClass($classDescription->{'group-sequence-provider'}[0]->value ?: null);
+            $metadata->setGroupProvider($classDescription->{'group-sequence-provider'}[0]->value ?: null);
             $metadata->setGroupSequenceProvider(true);
         }
 
diff --git a/src/Symfony/Component/Validator/Mapping/Loader/YamlFileLoader.php b/src/Symfony/Component/Validator/Mapping/Loader/YamlFileLoader.php
index 131e6799a6..e610b45427 100644
--- a/src/Symfony/Component/Validator/Mapping/Loader/YamlFileLoader.php
+++ b/src/Symfony/Component/Validator/Mapping/Loader/YamlFileLoader.php
@@ -151,7 +151,7 @@ class YamlFileLoader extends FileLoader
     {
         if (isset($classDescription['group_sequence_provider'])) {
             if (\is_string($classDescription['group_sequence_provider'])) {
-                $metadata->setGroupProviderClass($classDescription['group_sequence_provider']);
+                $metadata->setGroupProvider($classDescription['group_sequence_provider']);
             }
             $metadata->setGroupSequenceProvider(
                 (bool) $classDescription['group_sequence_provider']
diff --git a/src/Symfony/Component/Validator/Tests/Constraints/GroupSequenceProviderTest.php b/src/Symfony/Component/Validator/Tests/Constraints/GroupSequenceProviderTest.php
index 164e39fc96..e28399cd2c 100644
--- a/src/Symfony/Component/Validator/Tests/Constraints/GroupSequenceProviderTest.php
+++ b/src/Symfony/Component/Validator/Tests/Constraints/GroupSequenceProviderTest.php
@@ -21,13 +21,13 @@ class GroupSequenceProviderTest extends TestCase
     {
         $sequence = new GroupSequenceProvider(['class' => DummyGroupProvider::class]);
 
-        $this->assertSame(DummyGroupProvider::class, $sequence->class);
+        $this->assertSame(DummyGroupProvider::class, $sequence->provider);
     }
 
     public function testCreateAttributeStyle()
     {
         $sequence = new GroupSequenceProvider(class: DummyGroupProvider::class);
 
-        $this->assertSame(DummyGroupProvider::class, $sequence->class);
+        $this->assertSame(DummyGroupProvider::class, $sequence->provider);
     }
 }
diff --git a/src/Symfony/Component/Validator/Tests/ContainerGroupProviderFactoryTest.php b/src/Symfony/Component/Validator/Tests/ContainerGroupProviderFactoryTest.php
deleted file mode 100644
index 2b0d4acf6f..0000000000
--- a/src/Symfony/Component/Validator/Tests/ContainerGroupProviderFactoryTest.php
+++ /dev/null
@@ -1,61 +0,0 @@
-<?php
-
-/*
- * This file is part of the Symfony package.
- *
- * (c) Fabien Potencier <fabien@symfony.com>
- *
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- */
-
-use PHPUnit\Framework\TestCase;
-use Symfony\Component\DependencyInjection\Container;
-use Symfony\Component\Validator\ContainerGroupProviderFactory;
-use Symfony\Component\Validator\Exception\GroupDefinitionException;
-use Symfony\Component\Validator\Exception\UnexpectedTypeException;
-use Symfony\Component\Validator\Tests\Dummy\DummyGroupProvider;
-
-class ContainerGroupProviderFactoryTest extends TestCase
-{
-    public function testGetInstanceCreatesProvider()
-    {
-        $factory = new ContainerGroupProviderFactory(new Container());
-        $this->assertInstanceOf(DummyGroupProvider::class, $factory->getInstance(DummyGroupProvider::class));
-    }
-
-    public function testGetInstanceReturnsExistingProvider()
-    {
-        $factory = new ContainerGroupProviderFactory(new Container());
-        $v1 = $factory->getInstance(DummyGroupProvider::class);
-        $v2 = $factory->getInstance(DummyGroupProvider::class);
-        $this->assertSame($v1, $v2);
-    }
-
-    public function testGetInstanceReturnsService()
-    {
-        $provider = new DummyGroupProvider();
-        $container = new Container();
-        $container->set(DummyGroupProvider::class, $provider);
-
-        $factory = new ContainerGroupProviderFactory($container);
-
-        $this->assertSame($provider, $factory->getInstance(DummyGroupProvider::class));
-    }
-
-    public function testGetInstanceProviderNotFound()
-    {
-        $this->expectException(GroupDefinitionException::class);
-
-        $factory = new ContainerGroupProviderFactory(new Container());
-        $factory->getInstance('Unknown\\Class');
-    }
-
-    public function testGetInstanceProviderDoNotImplementInterface()
-    {
-        $this->expectException(UnexpectedTypeException::class);
-
-        $factory = new ContainerGroupProviderFactory(new Container());
-        $factory->getInstance(stdClass::class);
-    }
-}
diff --git a/src/Symfony/Component/Validator/Tests/DependencyInjection/AddGroupProvidersPassTest.php b/src/Symfony/Component/Validator/Tests/DependencyInjection/AddGroupProvidersPassTest.php
deleted file mode 100644
index 73f9f59e61..0000000000
--- a/src/Symfony/Component/Validator/Tests/DependencyInjection/AddGroupProvidersPassTest.php
+++ /dev/null
@@ -1,78 +0,0 @@
-<?php
-
-/*
- * This file is part of the Symfony package.
- *
- * (c) Fabien Potencier <fabien@symfony.com>
- *
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- */
-
-namespace Symfony\Component\Validator\Tests\DependencyInjection;
-
-use PHPUnit\Framework\TestCase;
-use Symfony\Component\DependencyInjection\Argument\ServiceClosureArgument;
-use Symfony\Component\DependencyInjection\ContainerBuilder;
-use Symfony\Component\DependencyInjection\Definition;
-use Symfony\Component\DependencyInjection\Reference;
-use Symfony\Component\DependencyInjection\ServiceLocator;
-use Symfony\Component\Validator\DependencyInjection\AddGroupProvidersPass;
-
-class AddGroupProvidersPassTest extends TestCase
-{
-    public function testThatGroupProviderServicesAreProcessed()
-    {
-        $container = new ContainerBuilder();
-        $groupProviderFactory = $container->register('validator.group_provider_factory')
-            ->addArgument([]);
-
-        $container->register('my_group_provider_service1', Provider1::class)
-            ->addTag('validator.group_provider', ['alias' => 'my_group_provider_alias1']);
-        $container->register('my_group_provider_service2', Provider2::class)
-            ->addTag('validator.group_provider');
-
-        $addGroupSequenceProvidersPass = new AddGroupProvidersPass();
-        $addGroupSequenceProvidersPass->process($container);
-
-        $locator = $container->getDefinition((string) $groupProviderFactory->getArgument(0));
-        $this->assertTrue(!$locator->isPublic() || $locator->isPrivate());
-        $expected = (new Definition(ServiceLocator::class, [[
-            Provider1::class => new ServiceClosureArgument(new Reference('my_group_provider_service1')),
-            'my_group_provider_alias1' => new ServiceClosureArgument(new Reference('my_group_provider_service1')),
-            Provider2::class => new ServiceClosureArgument(new Reference('my_group_provider_service2')),
-        ]]))->addTag('container.service_locator')->setPublic(false);
-        $this->assertEquals($expected, $locator->setPublic(false));
-    }
-
-    public function testAbstractGroupProvider()
-    {
-        $this->expectException(\InvalidArgumentException::class);
-        $this->expectExceptionMessage('The service "my_abstract_group_provider" tagged "validator.group_provider" must not be abstract.');
-        $container = new ContainerBuilder();
-        $container->register('validator.group_provider_factory')
-            ->addArgument([]);
-
-        $container->register('my_abstract_group_provider')
-            ->setAbstract(true)
-            ->addTag('validator.group_provider');
-
-        $addGroupSequenceProvidersPass = new AddGroupProvidersPass();
-        $addGroupSequenceProvidersPass->process($container);
-    }
-
-    public function testThatCompilerPassIsIgnoredIfThereIsNoGroupProviderFactoryDefinition()
-    {
-        $container = new ContainerBuilder();
-
-        $definitionsBefore = \count($container->getDefinitions());
-        $aliasesBefore = \count($container->getAliases());
-
-        $addGroupSequenceProvidersPass = new AddGroupProvidersPass();
-        $addGroupSequenceProvidersPass->process($container);
-
-        // the container is untouched (i.e. no new definitions or aliases)
-        $this->assertCount($definitionsBefore, $container->getDefinitions());
-        $this->assertCount($aliasesBefore, $container->getAliases());
-    }
-}
diff --git a/src/Symfony/Component/Validator/Tests/GroupProviderFactoryTest.php b/src/Symfony/Component/Validator/Tests/GroupProviderFactoryTest.php
deleted file mode 100644
index efd632784e..0000000000
--- a/src/Symfony/Component/Validator/Tests/GroupProviderFactoryTest.php
+++ /dev/null
@@ -1,30 +0,0 @@
-<?php
-
-/*
- * This file is part of the Symfony package.
- *
- * (c) Fabien Potencier <fabien@symfony.com>
- *
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- */
-
-use PHPUnit\Framework\TestCase;
-use Symfony\Component\Validator\GroupProviderFactory;
-use Symfony\Component\Validator\Tests\Dummy\DummyGroupProvider;
-
-class GroupProviderFactoryTest extends TestCase
-{
-    public function testGetInstance()
-    {
-        $factory = new GroupProviderFactory();
-        $this->assertInstanceOf(DummyGroupProvider::class, $factory->getInstance(DummyGroupProvider::class));
-    }
-
-    public function testPredefinedGetInstance()
-    {
-        $provider = new DummyGroupProvider();
-        $factory = new GroupProviderFactory([DummyGroupProvider::class => $provider]);
-        $this->assertSame($provider, $factory->getInstance(DummyGroupProvider::class));
-    }
-}
diff --git a/src/Symfony/Component/Validator/Tests/Mapping/Loader/XmlFileLoaderTest.php b/src/Symfony/Component/Validator/Tests/Mapping/Loader/XmlFileLoaderTest.php
index 8189a76b8e..5ba519ab19 100644
--- a/src/Symfony/Component/Validator/Tests/Mapping/Loader/XmlFileLoaderTest.php
+++ b/src/Symfony/Component/Validator/Tests/Mapping/Loader/XmlFileLoaderTest.php
@@ -128,7 +128,7 @@ class XmlFileLoaderTest extends TestCase
         $this->assertEquals($expected, $metadata);
     }
 
-    public function testLoadGroupProviderClass()
+    public function testLoadGroupProvider()
     {
         $loader = new XmlFileLoader(__DIR__.'/constraint-mapping.xml');
         $metadata = new ClassMetadata(GroupProviderDto::class);
@@ -136,7 +136,7 @@ class XmlFileLoaderTest extends TestCase
         $loader->loadClassMetadata($metadata);
 
         $expected = new ClassMetadata(GroupProviderDto::class);
-        $expected->setGroupProviderClass(DummyGroupProvider::class);
+        $expected->setGroupProvider(DummyGroupProvider::class);
         $expected->setGroupSequenceProvider(true);
 
         $this->assertEquals($expected, $metadata);
diff --git a/src/Symfony/Component/Validator/Tests/Mapping/Loader/YamlFileLoaderTest.php b/src/Symfony/Component/Validator/Tests/Mapping/Loader/YamlFileLoaderTest.php
index fd31675f23..a5c983939b 100644
--- a/src/Symfony/Component/Validator/Tests/Mapping/Loader/YamlFileLoaderTest.php
+++ b/src/Symfony/Component/Validator/Tests/Mapping/Loader/YamlFileLoaderTest.php
@@ -168,7 +168,7 @@ class YamlFileLoaderTest extends TestCase
         $this->assertEquals($expected, $metadata);
     }
 
-    public function testLoadGroupProviderClass()
+    public function testLoadGroupProvider()
     {
         $loader = new YamlFileLoader(__DIR__.'/constraint-mapping.yml');
         $metadata = new ClassMetadata(GroupProviderDto::class);
@@ -176,7 +176,7 @@ class YamlFileLoaderTest extends TestCase
         $loader->loadClassMetadata($metadata);
 
         $expected = new ClassMetadata(GroupProviderDto::class);
-        $expected->setGroupProviderClass(DummyGroupProvider::class);
+        $expected->setGroupProvider(DummyGroupProvider::class);
         $expected->setGroupSequenceProvider(true);
 
         $this->assertEquals($expected, $metadata);
diff --git a/src/Symfony/Component/Validator/Tests/Validator/RecursiveValidatorTest.php b/src/Symfony/Component/Validator/Tests/Validator/RecursiveValidatorTest.php
index 4be5f8b3f2..7c0b22f9a1 100644
--- a/src/Symfony/Component/Validator/Tests/Validator/RecursiveValidatorTest.php
+++ b/src/Symfony/Component/Validator/Tests/Validator/RecursiveValidatorTest.php
@@ -1264,14 +1264,14 @@ class RecursiveValidatorTest extends TestCase
         }
     }
 
-    public function testGroupProviderClass()
+    public function testGroupProvider()
     {
         $dto = new GroupProviderDto();
 
         $metadata = new ClassMetadata($dto::class);
         $metadata->addPropertyConstraint('firstName', new NotBlank(groups: ['foo']));
         $metadata->addPropertyConstraint('lastName', new NotBlank(groups: ['foo']));
-        $metadata->setGroupProviderClass(DummyGroupProvider::class);
+        $metadata->setGroupProvider(DummyGroupProvider::class);
         $metadata->setGroupSequenceProvider(true);
         $this->metadataFactory->addMetadata($metadata);
 
diff --git a/src/Symfony/Component/Validator/Validator/RecursiveContextualValidator.php b/src/Symfony/Component/Validator/Validator/RecursiveContextualValidator.php
index 1b9096468f..7ead3ad9a9 100644
--- a/src/Symfony/Component/Validator/Validator/RecursiveContextualValidator.php
+++ b/src/Symfony/Component/Validator/Validator/RecursiveContextualValidator.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Component\Validator\Validator;
 
+use Psr\Container\ContainerInterface;
 use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\Constraints\Composite;
 use Symfony\Component\Validator\Constraints\Existence;
@@ -26,8 +27,6 @@ use Symfony\Component\Validator\Exception\RuntimeException;
 use Symfony\Component\Validator\Exception\UnexpectedValueException;
 use Symfony\Component\Validator\Exception\UnsupportedMetadataException;
 use Symfony\Component\Validator\Exception\ValidatorException;
-use Symfony\Component\Validator\GroupProviderFactory;
-use Symfony\Component\Validator\GroupProviderFactoryInterface;
 use Symfony\Component\Validator\Mapping\CascadingStrategy;
 use Symfony\Component\Validator\Mapping\ClassMetadataInterface;
 use Symfony\Component\Validator\Mapping\Factory\MetadataFactoryInterface;
@@ -52,14 +51,14 @@ class RecursiveContextualValidator implements ContextualValidatorInterface
     private MetadataFactoryInterface $metadataFactory;
     private ConstraintValidatorFactoryInterface $validatorFactory;
     private array $objectInitializers;
-    private GroupProviderFactoryInterface $groupProviderFactory;
+    private ContainerInterface $groupProviderLocator;
 
     /**
      * Creates a validator for the given context.
      *
      * @param ObjectInitializerInterface[] $objectInitializers The object initializers
      */
-    public function __construct(ExecutionContextInterface $context, MetadataFactoryInterface $metadataFactory, ConstraintValidatorFactoryInterface $validatorFactory, array $objectInitializers = [], GroupProviderFactoryInterface $groupProviderFactory = new GroupProviderFactory())
+    public function __construct(ExecutionContextInterface $context, MetadataFactoryInterface $metadataFactory, ConstraintValidatorFactoryInterface $validatorFactory, array $objectInitializers = [], ContainerInterface $groupProviderLocator = null)
     {
         $this->context = $context;
         $this->defaultPropertyPath = $context->getPropertyPath();
@@ -67,7 +66,7 @@ class RecursiveContextualValidator implements ContextualValidatorInterface
         $this->metadataFactory = $metadataFactory;
         $this->validatorFactory = $validatorFactory;
         $this->objectInitializers = $objectInitializers;
-        $this->groupProviderFactory = $groupProviderFactory;
+        $this->groupProviderLocator = $groupProviderLocator;
     }
 
     public function atPath(string $path): static
@@ -440,8 +439,8 @@ class RecursiveContextualValidator implements ContextualValidatorInterface
                     $group = $metadata->getGroupSequence();
                     $defaultOverridden = true;
                 } elseif ($metadata->isGroupSequenceProvider()) {
-                    if (null !== $groupProviderClass = $metadata->getGroupProviderClass()) {
-                        $group = $this->groupProviderFactory->getInstance($groupProviderClass)->getGroups($object);
+                    if (null !== $groupProvider = $metadata->getGroupProvider()) {
+                        $group = $this->groupProviderLocator?->get($groupProvider)->getGroups($object);
                     } else {
                         // The group sequence is dynamically obtained from the validated
                         // object
diff --git a/src/Symfony/Component/Validator/Validator/RecursiveValidator.php b/src/Symfony/Component/Validator/Validator/RecursiveValidator.php
index 2513621679..b24448ed0f 100644
--- a/src/Symfony/Component/Validator/Validator/RecursiveValidator.php
+++ b/src/Symfony/Component/Validator/Validator/RecursiveValidator.php
@@ -11,14 +11,13 @@
 
 namespace Symfony\Component\Validator\Validator;
 
+use Psr\Container\ContainerInterface;
 use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\Constraints\GroupSequence;
 use Symfony\Component\Validator\ConstraintValidatorFactoryInterface;
 use Symfony\Component\Validator\ConstraintViolationListInterface;
 use Symfony\Component\Validator\Context\ExecutionContextFactoryInterface;
 use Symfony\Component\Validator\Context\ExecutionContextInterface;
-use Symfony\Component\Validator\GroupProviderFactory;
-use Symfony\Component\Validator\GroupProviderFactoryInterface;
 use Symfony\Component\Validator\Mapping\Factory\MetadataFactoryInterface;
 use Symfony\Component\Validator\Mapping\MetadataInterface;
 use Symfony\Component\Validator\ObjectInitializerInterface;
@@ -34,20 +33,20 @@ class RecursiveValidator implements ValidatorInterface
     protected $metadataFactory;
     protected $validatorFactory;
     protected $objectInitializers;
-    protected GroupProviderFactoryInterface $groupProviderFactory;
+    protected ContainerInterface $groupProviderLocator;
 
     /**
      * Creates a new validator.
      *
      * @param ObjectInitializerInterface[] $objectInitializers The object initializers
      */
-    public function __construct(ExecutionContextFactoryInterface $contextFactory, MetadataFactoryInterface $metadataFactory, ConstraintValidatorFactoryInterface $validatorFactory, array $objectInitializers = [], GroupProviderFactoryInterface $groupProviderFactory = new GroupProviderFactory())
+    public function __construct(ExecutionContextFactoryInterface $contextFactory, MetadataFactoryInterface $metadataFactory, ConstraintValidatorFactoryInterface $validatorFactory, array $objectInitializers = [], ContainerInterface $groupProviderLocator = null)
     {
         $this->contextFactory = $contextFactory;
         $this->metadataFactory = $metadataFactory;
         $this->validatorFactory = $validatorFactory;
         $this->objectInitializers = $objectInitializers;
-        $this->groupProviderFactory = $groupProviderFactory;
+        $this->groupProviderLocator = $groupProviderLocator;
     }
 
     public function startContext(mixed $root = null): ContextualValidatorInterface
@@ -57,7 +56,7 @@ class RecursiveValidator implements ValidatorInterface
             $this->metadataFactory,
             $this->validatorFactory,
             $this->objectInitializers,
-            $this->groupProviderFactory,
+            $this->groupProviderLocator,
         );
     }
 
@@ -68,7 +67,7 @@ class RecursiveValidator implements ValidatorInterface
             $this->metadataFactory,
             $this->validatorFactory,
             $this->objectInitializers,
-            $this->groupProviderFactory,
+            $this->groupProviderLocator,
         );
     }
 
diff --git a/src/Symfony/Component/Validator/ValidatorBuilder.php b/src/Symfony/Component/Validator/ValidatorBuilder.php
index bab9f49fac..90f0dc1d74 100644
--- a/src/Symfony/Component/Validator/ValidatorBuilder.php
+++ b/src/Symfony/Component/Validator/ValidatorBuilder.php
@@ -15,6 +15,7 @@ use Doctrine\Common\Annotations\AnnotationReader;
 use Doctrine\Common\Annotations\PsrCachedReader;
 use Doctrine\Common\Annotations\Reader;
 use Psr\Cache\CacheItemPoolInterface;
+use Psr\Container\ContainerInterface;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
 use Symfony\Component\Validator\Context\ExecutionContextFactory;
 use Symfony\Component\Validator\Exception\LogicException;
@@ -53,7 +54,7 @@ class ValidatorBuilder
     private bool $enableAttributeMapping = false;
     private ?MetadataFactoryInterface $metadataFactory = null;
     private ConstraintValidatorFactoryInterface $validatorFactory;
-    private GroupProviderFactoryInterface $groupProviderFactory;
+    private ?ContainerInterface $groupProviderLocator;
     private ?CacheItemPoolInterface $mappingCache = null;
     private ?TranslatorInterface $translator = null;
     private ?string $translationDomain = null;
@@ -316,9 +317,9 @@ class ValidatorBuilder
     /**
      * @return $this
      */
-    public function setGroupProviderFactory(GroupProviderFactoryInterface $groupProviderFactory): static
+    public function setGroupProviderLocator(ContainerInterface $groupProviderLocator): static
     {
-        $this->groupProviderFactory = $groupProviderFactory;
+        $this->groupProviderLocator = $groupProviderLocator;
 
         return $this;
     }
@@ -410,7 +411,6 @@ class ValidatorBuilder
         }
 
         $validatorFactory = $this->validatorFactory ?? new ConstraintValidatorFactory();
-        $groupProviderFactory = $this->groupProviderFactory ?? new GroupProviderFactory();
         $translator = $this->translator;
 
         if (null === $translator) {
@@ -426,7 +426,7 @@ class ValidatorBuilder
 
         $contextFactory = new ExecutionContextFactory($translator, $this->translationDomain);
 
-        return new RecursiveValidator($contextFactory, $metadataFactory, $validatorFactory, $this->initializers, $groupProviderFactory);
+        return new RecursiveValidator($contextFactory, $metadataFactory, $validatorFactory, $this->initializers, $this->groupProviderLocator);
     }
 
     private function createAnnotationReader(): Reader
```

</details>

yceruto:
Hi @nicolas-grekas! Thanks for this nice code suggestion. I see your idea does indeed simplify the proposal a lot, and I'm happy to ship it as is. However, I still have concerns about the standalone usage of the component. After your change proposal, the new feature will not work without a proper implementation (_in userland_) of the `ContainerInterface` for `GroupProviderLocator`, which looks a bit weird to me as we used to ship all artifacts for simple usage. Do you think this one is an exception?

nicolas-grekas:
I'm confident it's easy enough to create a container and that standalone usage is not an issue. We already have a few use cases that require this.

Done:
```php

$container = new class($factories) implements ContainerInterface {
    use ServiceLocatorTrait;
}
```

yceruto:
Simplification done, thanks @nicolas-grekas !

src/Symfony/Component/Validator/Constraints/GroupSequenceProvider.php
```diff
@@ -23,4 +23,10 @@
 #[\Attribute(\Attribute::TARGET_CLASS)]
 class GroupSequenceProvider
 {
+    public ?string $provider = null;
+
+    public function __construct(array $options = [], string $provider = null)
```
nicolas-grekas:
```suggestion
    public function __construct(array|string $options = [], string $provider = null)
```

then if $option is a string, make it the provider? Would allow #[GroupSequenceProvider('foo')]

nicolas-grekas (approved):
Nice and easy :)

src/Symfony/Component/Validator/Constraints/GroupSequenceProvider.php
stof: As there was no constructor before, there is no BC need to support passing an array.

this could use `public function __construct(string $provider = null)` as constructor, using the `@NamedArgumentConstructor` annotation of doctrine/annotations to support annotations and `HasNamedArguments` attribute of the validator component to support the config file loaders.

yceruto: Updated! Thanks to both of you.

fabpot (approved):

fabpot:
Thank you @yceruto.

PR #51350, opened 2023-08-11
[Security] Prevent creating session in stateless firewalls

### LINKED ISSUE #51319: Session created by default handlers on stateless firewalls

Seb33300:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix https://github.com/symfony/symfony/issues/51319
| License       | MIT
| Doc PR        | 

Please check related issue for details.

Same as https://github.com/symfony/symfony/pull/51320 with @chalasr suggestion: https://github.com/symfony/symfony/pull/51320#discussion_r1290519241

### DIFF
--- a/src/Symfony/Component/Security/Http/Tests/Authentication/DefaultAuthenticationFailureHandlerTest.php
+++ b/src/Symfony/Component/Security/Http/Tests/Authentication/DefaultAuthenticationFailureHandlerTest.php
@@ -46,6 +46,7 @@ class DefaultAuthenticationFailureHandlerTest extends TestCase
 
         $this->session = $this->createMock(SessionInterface::class);
         $this->request = $this->createMock(Request::class);
+        $this->request->attributes = new ParameterBag(['_stateless' => false]);
         $this->request->expects($this->any())->method('getSession')->willReturn($this->session);
         $this->exception = $this->getMockBuilder(AuthenticationException::class)->onlyMethods(['getMessage'])->getMock();
     }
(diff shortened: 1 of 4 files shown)


chalasr (changes requested):
Please add test cases with `_stateless => true` for both success and failure handlers to prevent regressions

Seb33300:
> Please add test cases with `_stateless => true` for both success and failure handlers to prevent regressions

@chalasr tests added

stloyd (approved):

src/Symfony/Component/Security/Http/Tests/Authentication/DefaultAuthenticationFailureHandlerTest.php
```diff
@@ -46,6 +46,7 @@ protected function setUp(): void
 
         $this->session = $this->createMock(SessionInterface::class);
         $this->request = $this->createMock(Request::class);
+        $this->request->attributes = new ParameterBag(['_stateless' => false]);
```
mtarld, line 49: I don't think this is needed, as the `getBoolean` will default to `false` :slightly_smiling_face:

stloyd:
@wouterj @chalasr Could you have a second look at this?

I can confirm the fix on HWIOAuthBundle. Before:
```
PHPUnit 9.6.11 by Sebastian Bergmann and contributors.

Testing /Users/stloyd/Documents/HWIOAuthBundle/tests/Functional

<!-- Session was used while the request was declared stateless. (500 Internal Server Error) -->
// error page printed
```

After:
```
PHPUnit 9.6.11 by Sebastian Bergmann and contributors.

Testing /Users/stloyd/Documents/HWIOAuthBundle/tests/Functional

Time: 00:00.090, Memory: 18.00 MB

OK (1 test, 4 assertions)
```

mtarld (approved):

src/Symfony/Component/Security/Http/Tests/Authentication/DefaultAuthenticationFailureHandlerTest.php
Seb33300: `attributes` property is not defined in the request mock so it needs to be defined at least with a `new ParameterBag()`.

And I prefer to keep the `['_stateless' => false]` explicit so we can better understand why we need `attributes` to be defined.

stloyd:
@fabpot / @nicolas-grekas or anyone else can check this fix? I would love to have it in near 6.3 release

Seb33300:
August is usually a period of holidays in europe so @chalasr and @wouterj may not be available this week.
Let's see next week :)

stof (approved):

chalasr (approved):

chalasr:
Thank you @Seb33300.

PR #51352, opened 2023-08-11
[DependencyInjection] Allow passing an `inline_service` to a `service_locator`

ruudk:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  |no
| Deprecations? |no 
| Tickets       | 
| License       | MIT
| Doc PR        | 

Something, you want to include an inline service to a service locator.

This works fine. Except that the PHPDoc doesn't allow it causing PHPStan to fail.

Example:
```php
 $services->set(OtherService::class)
        ->args([
            '$services' => service_locator([
                'my_service' => inline_service(SomeClass::class)->args([]),
            ])
        ]);          
```

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Loader/Configurator/ContainerConfigurator.php
+++ b/src/Symfony/Component/DependencyInjection/Loader/Configurator/ContainerConfigurator.php
@@ -119,7 +119,7 @@ function inline_service(string $class = null): InlineServiceConfigurator
 /**
  * Creates a service locator.
  *
- * @param ReferenceConfigurator[] $values
+ * @param array<ReferenceConfigurator|InlineServiceConfigurator> $values
  */
 function service_locator(array $values): ServiceLocatorArgument
 {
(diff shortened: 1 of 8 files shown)


nicolas-grekas:
It'd be great adding an integration test case that runs both a compiled container and a ContainerBuilder on this situation. Historically this wasn't supported due to technical reasons. Maybe these limitations have been removed but a test case would ensure that's true, and that it remains so (maybe there's already a test case that covers this already?)
Can you please have a look?

ruudk:
@nicolas-grekas I agree. I was also surprised that there is no integration test for the `service_locator` or `inline_service` at all 😅

ruudk:
Do you maybe have an example of an existing test, that covers both ContainerBuilder and Compiler that I can use an example?

ruudk:
@nicolas-grekas Added the extra tests. Changed the base branch to 6.3 because this is only supported there, not on 5.4.

ruudk:
The failing test is not related to this PR.

nicolas-grekas (approved):

nicolas-grekas:
Thank you @ruudk.

PR #51357, opened 2023-08-11
[FrameworkBundle] Deprecate not setting some options (uid, validation)

### LINKED ISSUE #51097: [RFC] Shrink recipes for v7
Over time, recipes grow in size. For example, the recipe for 6.2 added the `handle_all_throwables` option to turn it on by default.

It'd be great to shrink recipes by making v7 defaults to the values we have currently in the recipes.

Here is the current config file for framework-bundle v7:
https://github.com/symfony/recipes/blob/main/symfony/framework-bundle/7.0/config/packages/framework.yaml

Let's deprecate not setting all/most of those options and shrink the recipe.
Let's review the other recipes also as part of the same effort.

Feedback and Help Wanted for the implementations.

Jean-Beru:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes 
| Tickets       | Fix #51097
| License       | MIT
| Doc PR        | 

This PR deprecates not setting 4 options in framework configuration to match the v7.0 recipe:
* `uid.default_uuid_version` which is `6` by default and is set to `7` by recipe
* `uid.time_based_uuid_version`  which is `6` by default and is set to `7` by recipe
* `validation.email_validation_mode` which is `null` by default and is set to `html5` by recipe

See #51097.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
@@ -1018,6 +1018,15 @@ class Configuration implements ConfigurationInterface
     private function addValidationSection(ArrayNodeDefinition $rootNode, callable $enableIfStandalone): void
     {
         $rootNode
+            ->validate()
+                ->always(function ($v) {
+                    if ($v['validation']['enabled'] && !\array_key_exists('email_validation_mode', $v['validation'])) {
+                        trigger_deprecation('symfony/framework-bundle', '6.4', 'Not setting the "framework.validation.email_validation_mode" config option is deprecated. It will default to "html5" in 7.0.');
+                    }
+
+                    return $v;
+                })
+            ->end()
             ->children()
                 ->arrayNode('validation')
                     ->info('validation configuration')
@@ -2350,6 +2359,23 @@ class Configuration implements ConfigurationInterface
     private function addUidSection(ArrayNodeDefinition $rootNode, callable $enableIfStandalone): void
     {
         $rootNode
+            ->validate()
+                ->always(function ($v) {
+                    if ($v['uid']['enabled']) {
+                        if (!\array_key_exists('default_uuid_version', $v['uid'])) {
+                            trigger_deprecation('symfony/framework-bundle', '6.4', 'Not setting the "framework.uid.default_uuid_version" config option is deprecated. It will default to "7" in 7.0.');
+                        }
+
+                        if (!\array_key_exists('time_based_uuid_version', $v['uid'])) {
+                            trigger_deprecation('symfony/framework-bundle', '6.4', 'Not setting the "framework.uid.time_based_uuid_version" config option is deprecated. It will default to "7" in 7.0.');
+                        }
+                    }
+
+                    $v['uid'] += ['default_uuid_version' => 6, 'time_based_uuid_version' => 6];
+
+                    return $v;
+                })
+            ->end()
             ->children()
                 ->arrayNode('uid')
                     ->info('Uid configuration')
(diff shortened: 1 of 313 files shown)


src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -2316,14 +2325,31 @@ private function addRateLimiterSection(ArrayNodeDefinition $rootNode, callable $
...
         $rootNode
+            ->validate()
+                ->always(function ($v) {
+                    if ($v['uid']['enabled']) {
+                        if (!array_key_exists('default_uuid_version', $v['uid'])) {
+                            trigger_deprecation('symfony/framework-bundle', '6.1', 'Not setting the "framework.uid.default_uuid_version" config option is deprecated. It will default to "7" in 7.0.');
+                        }
+
+                        if (!array_key_exists('time_based_uuid_version', $v['uid'])) {
+                            trigger_deprecation('symfony/framework-bundle', '6.1', 'Not setting the "framework.uid.time_based_uuid_version" config option is deprecated. It will default to "7" in 7.0.');
```
stof: this cannot be a deprecation introduced in FrameworkBundle 6.1 as that version is already released

src/Symfony/Bundle/SecurityBundle/DependencyInjection/MainConfiguration.php
```diff
@@ -418,6 +418,15 @@ private function addProvidersSection(ArrayNodeDefinition $rootNode): void
     private function addPasswordHashersSection(ArrayNodeDefinition $rootNode): void
     {
         $rootNode
+            ->validate()
+                ->always(function (array $v): array {
+                    if (empty($v['password_hashers'])) {
+                        trigger_deprecation('symfony/framework-bundle', '6.4', 'Not setting the "security.password_hashers" config option is deprecated. It will default to "Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface: auto" in 7.0.');
```
stof: This is wrong. This deprecation does not happen in the èsymfony/framework-bundle` package.

stof: I'm not sure about this deprecation. Default values for a prototyped array node are not merged with provided config. So as soon as any config file (or config prepended by a bundle) adds some other config for password hashers, the config for PasswordAuthenticatedUserInterface would be lost.

I'd suggest keeping that one in the recipe.

stof (changes requested):

src/Symfony/Bundle/SecurityBundle/DependencyInjection/MainConfiguration.php
nicolas-grekas: They would be merged if we didn't call `performNoDeepMerging`, isn't it?
Why don't we perform deep merging BTW?

nicolas-grekas: Anyway, let's drop this, it's not needed to deprecate not setting this

wouterj: :shrug: a5cfc2207c09f6182333b3b812673202701c33f3 "some bug fixes in DIC config classes"

nicolas-grekas (dismissed):
I addressed @stof's review.

nicolas-grekas:
Thank you @Jean-Beru.

src/Symfony/Bundle/SecurityBundle/DependencyInjection/MainConfiguration.php
stof: @nicolas-grekas no. deep merging is about the various configs received as input, not the defaults

PR #51359, opened 2023-08-11
[Security] Fix error with lock_factory in login_throttling

### LINKED ISSUE #51347: [Security] Error with lock_factory in login_throttling

 '>= 6.2'

### Description

I just upgraded my project from 5.4 to 6.3 and this error popped up :

**Rate limiter "_login_local_main" requires the Lock component to be configured.**

After some tests, I found out that this line (in my `security.yaml`) is the cause : 

```yaml
lock_factory: lock.default.factory
```

When I looked in 

```php
Symfony\Bundle\SecurityBundle\DependencyInjection\Security\Factory\LoginThrottlingFactory
```

from the Security Bundle, I noticed that this if is the problem :

```php
### LoginThrottlingFactory line 101
if (!$container->hasDefinition('lock.factory.abstract')) {
     throw new LogicException(sprintf('Rate limiter "%s" requires the Lock component to be configured.', $name));
}
```

If I dump `$container->getDefinitions()` before this if I don't see any lock in the `definitions` array, which seems to be the issue here.

This check was added in this PR : https://github.com/symfony/symfony/pull/45569

I don't see the usefulness of this check here.
The **'lock.factory.abstract'** is created in the Framework Bundle and the Security Bundle can't see it at this stage.

From my understanding we can either discard this change and let an error pop later saying that this service does not exist
or move this check in a CompilerPass. (I personally would go for the first solution)

### How to reproduce

Using PHP 8.2.7

```bash
symfony new --webapp demo
```
```bash
composer require symfony/rate-limiter

```

```bash
composer require symfony/lock
```

```php
### src/Controller/DemoController.php
<?php

namespace App\Controller;

class DemoController
{
    #[Route('/demo')]
    public function index(): Response
    {
        return new Response();
    }
}
```

```yaml
### security.yaml
security:
    password_hashers:
    providers:
        users_in_memory: { memory: null }
    firewalls:
        dev:
            security: false
        main:
            lazy: true
            provider: users_in_memory
            login_throttling:
                max_attempts: 5
                interval: '15 minutes'
                lock_factory: lock.default.factory

    access_control:
        - { path: ^/demo, roles: ROLE_ADMIN }

```

```yaml
### lock.yaml
framework:
    lock: '%env(LOCK_DSN)%'
```

### Possible Solution

_No response_

### Additional Context

_No response_

BaptisteContreras:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no
| Tickets       | Fix #51347
| License       | MIT
| Doc PR        | 

Remove incorrect check for Lock configuration in **Security Bundle** which leads to an exception when using lock_factory of the **login_throttling** (as explained in #51347)

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/LoginThrottlingFactory.php
+++ b/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/LoginThrottlingFactory.php
@@ -98,9 +98,6 @@ class LoginThrottlingFactory implements AuthenticatorFactoryInterface
             if (!interface_exists(LockInterface::class)) {
                 throw new LogicException(sprintf('Rate limiter "%s" requires the Lock component to be installed. Try running "composer require symfony/lock".', $name));
             }
-            if (!$container->hasDefinition('lock.factory.abstract')) {
-                throw new LogicException(sprintf('Rate limiter "%s" requires the Lock component to be configured.', $name));
-            }
 
             $limiter->replaceArgument(2, new Reference($limiterConfig['lock_factory']));
         }


BaptisteContreras:
I don't really understand why these tests failed. It seems unrelated to my changes...
If someone can enlighten me !

wouterj (approved):

wouterj:
> I don't really understand why these tests failed. It seems unrelated to my changes...
> If someone can enlighten me !

Indeed. Tests on the branches are often broken (due to transient tests or changes in the ecosystem). These are unrelated to the changes you've made, so we can ignore them in this PR.

fabpot (approved):

fabpot:
Thank you @BaptisteContreras.

PR #51365, opened 2023-08-12
[Clock] Add $modifier argument to the now() helper

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

With this change, one can replace all `new \DateTimeImmutable($foo)` by `now($foo)`.

### DIFF
--- a/src/Symfony/Component/Clock/Resources/now.php
+++ b/src/Symfony/Component/Clock/Resources/now.php
@@ -13,13 +13,29 @@ namespace Symfony\Component\Clock;
 
 if (!\function_exists(now::class)) {
     /**
-     * Returns the current time as a DateTimeImmutable.
-     *
-     * Note that you should prefer injecting a ClockInterface or using
-     * ClockAwareTrait when possible instead of using this function.
+     * @throws \DateMalformedStringException When the modifier is invalid
      */
-    function now(): \DateTimeImmutable
+    function now(string $modifier = null): \DateTimeImmutable
     {
-        return Clock::get()->now();
+        if (null === $modifier || 'now' === $modifier) {
+            return Clock::get()->now();
+        }
+
+        $now = Clock::get()->now();
+
+        if (\PHP_VERSION_ID < 80300) {
+            try {
+                $tz = (new \DateTimeImmutable($modifier, $now->getTimezone()))->getTimezone();
+            } catch (\Exception $e) {
+                throw new \DateMalformedStringException($e->getMessage(), $e->getCode(), $e);
+            }
+            $now = $now->setTimezone($tz);
+
+            return @$now->modify($modifier) ?: throw new \DateMalformedStringException(error_get_last()['message'] ?? sprintf('Invalid date modifier "%s".', $modifier));
+        }
+
+        $tz = (new \DateTimeImmutable($modifier, $now->getTimezone()))->getTimezone();
+
+        return $now->setTimezone($tz)->modify($modifier);
     }
 }
--- a/src/Symfony/Component/Clock/CHANGELOG.md
+++ b/src/Symfony/Component/Clock/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * Throw `DateMalformedStringException`/`DateInvalidTimeZoneException` when appropriate
+ * Add `$modifier` argument to the `now()` helper
 
 6.3
 ---
--- a/src/Symfony/Component/Clock/Tests/ClockTest.php
+++ b/src/Symfony/Component/Clock/Tests/ClockTest.php
@@ -39,6 +39,19 @@ class ClockTest extends TestCase
         $this->assertInstanceOf(NativeClock::class, Clock::get());
     }
 
+    public function testNowModifier()
+    {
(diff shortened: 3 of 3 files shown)


OskarStark (approved):
Should we add a testcase?

nicolas-grekas:
Test case added, thanks for the review.

OskarStark (approved):

xabbuh:
I am not convinced that a function named `now()` should return another date than the current date (according to the underlying clock). Shouldn't this rather be a different function?

src/Symfony/Component/Clock/Resources/now.php
```diff
@@ -18,8 +18,12 @@
      * Note that you should prefer injecting a ClockInterface or using
      * ClockAwareTrait when possible instead of using this function.
      */
-    function now(): \DateTimeImmutable
+    function now(string $modifier = null): \DateTimeImmutable
     {
+        if (null !== $modifier) {
+            return Clock::get()->now()->modify($modifier);
```
xabbuh: I think we should check the return value of `modify()` first and throw an exception if `false` is returned (and we should probably also silence the call).

chalasr:
I agree with @xabbuh

wouterj:
It reads nicely if we allow only relative modifiers: `now('+10 minutes')`
But I agree with @xabbuh and @chalasr for absolute modifiers `now('2023-01-01')` does feel wrong.

chalasr:
`now()->modify('+10minutes')` reads even better imho. Let's just keep `now()` return now?

nicolas-grekas:
PR updated: it now leverages https://github.com/symfony/polyfill/pull/440 so that we throw a `DateMalformedStringException` when the provided modifier is invalid. I also updated `MockClock` to throw this exception / DateInvalidTimeZoneException when appropriate.

> now('+10 minutes')

Yes, this reads perfectly fine

> now('2023-01-01') does feel wrong.

I don't see any issue with that personally. This is trivial to understand to me.
Introducing another function would feel worse, especially considering we cannot differentiate between relative and absolute modifiers.

> now()->modify('+10minutes') reads even better imho.

It doesn't to me, especially considering the fact this misses proper error handling. Also considering this is supposed to replace `new \DateTimeImmutable('+10minutes')`: if the number of keystrokes is not clearly reduced, this fails the target to me.

Please reconsider :)

OskarStark:
I still like the modifier and the reading, while I understand the concerns, I would go with this PR 👍🏻

OskarStark (approved):

stof:
I'm wondering whether the change to MockClock should really be bundled in the same PR (and so in the same commit in the final history). This would hide entirely the behavior change in MockClock even for people reading the full git history and not just the changelog (which filters out minor changes).

src/Symfony/Component/Clock/Resources/now.php
```diff
@@ -17,9 +17,19 @@
      *
      * Note that you should prefer injecting a ClockInterface or using
```
chalasr: https://github.com/symfony/symfony/blob/92ff2b20d4360b529afa53f179bcf00e0af7d678/src/Symfony/Component/Clock/Resources/now.php#L16 is not true anymore

chalasr:
I won't block but I think it's not worth it

antonkomarev:
I think that explicitly calling a modify method is a much more obvious way to get a modified date from now.

src/Symfony/Component/Clock/Resources/now.php
nicolas-grekas: I removed the sentence altogether.

src/Symfony/Component/Clock/Resources/now.php
```diff
@@ -13,13 +13,20 @@
...
     {
-        return Clock::get()->now();
+        if (null === $modifier || 'now' === $modifier) {
+            return Clock::get()->now();
+        }
+
+        if (\PHP_VERSION_ID >= 80300) {
+            $now = Clock::get()->now()->modify($modifier);
+        } else {
+            $now = @Clock::get()->now()->modify($modifier) ?: throw new \DateMalformedStringException(error_get_last()['message'] ?? sprintf('Invalid date modifier "%s".', $modifier));
```
GromNaN: When is there or isn't an error message?

nicolas-grekas:
PR updated with error handling moved to #51412

> explicitly calling a modify method

there are two major differences with `now()->modify($foo)`:
- error handling (could be alleviated by #51415)
- timezone handling: `$date->modify('Europe/Paris')` does nothing, while `now('Europe/Paris')` would work as we'd expect.

GromNaN (approved):
This is a nice and concise syntax.
Developers can still call `now()->modify('+10 minutes)` it they prefer.

I like that we can pass the timezone as modifier.

src/Symfony/Component/Clock/Resources/now.php
nicolas-grekas: there's always one, unless a broken custom error handler is used :shrug:

PR #51368, opened 2023-08-12
[DomCrawler] Added argument `$default` to method `Crawler::attr()`

Rastishka:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | no
| License       | MIT

Added argument `$default` to method `->attr()` like `->text()` and `->html()` have.

### DIFF
--- a/src/Symfony/Component/DomCrawler/Crawler.php
+++ b/src/Symfony/Component/DomCrawler/Crawler.php
@@ -58,7 +58,6 @@ class Crawler implements \Countable, \IteratorAggregate
      */
     private bool $isHtml = true;
 
-
     private ?HTML5 $html5Parser = null;
 
     /**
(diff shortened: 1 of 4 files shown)


nicolas-grekas:
Do we really need this? My feeling is that we now have the `??` operator and it's nicer to use.

antonkomarev:
I'd prefer to use native `??`. It feels more obvious about the result.

Rastishka:
> Do we really need this? My feeling is that we now have the `??` operator and it's nicer to use.

Yes we need, because setting default value also prevents throwing exception when node does not exist (methods text() and html() act similar).

So this code:
```
try {
    $value = $dom->filter('a.someclass')->attr('href');
} catch (\InvalidArgumentException $e) {
    $value = '';
}
```
can be simplifies to
` $value = $dom->filter('a.someclass')->attr('href', '');`

src/Symfony/Component/DomCrawler/Crawler.php
```diff
@@ -522,17 +522,23 @@ public function children(string $selector = null): static
...
      * Returns the attribute value of the first node of the list.
      *
+     * @param string|null $default When not null: the value to return when the node or attribute is empty
+     *
      * @throws \InvalidArgumentException When current node is empty
      */
-    public function attr(string $attribute): ?string
+    public function attr(string $attribute, string $default = null): ?string
     {
         if (!$this->nodes) {
```
nicolas-grekas: To keep BC:
```suggestion
    public function attr(string $attribute/* , string $default = null */): ?string
    {
        $default = \func_num_args() > 1 ? func_get_arg(1) : null;
        if (!$this->nodes) {
```

nicolas-grekas (commented):
Thanks for the explanation. Works for me.

nicolas-grekas:
Please also update the UPGRADE-6.4.md file, and please rebase to get rid of merge commits (PRs shouldn't contain any).

nicolas-grekas:
Thank you @Rastishka.

PR #51369, opened 2023-08-12
[Serializer] Fix deserializing object collection properties

### LINKED ISSUE #51261: [Serializer] Handling array properties during denormalization

6.3

### Description

Updating from Serializer 5.4 to 6.3 breaks our testsuite as the denormalizer which is supposed to denormalize an array property does not get executed anymore as it does implement the `getSupportedTypes` method and in the serializer code path that handles the `getSupportedTypes` logic it no longer calls the `supportsDenormalization` method on that denormalizer (the concrete example can be seen under the steps to reproduce). I'm not entirely familiar with the Serializer component so maybe I'm just doing something wrong.

### How to reproduce

The interfaces/classes needed to reproduce this:

```php
interface FooBaseDummy
{
}
```

```php
class FooDummy implements FooBaseDummy
{
    /**
     * @var string
     */
    public $name;
}

```

```php
class ArrayPropertyDummy
{
    /**
     * @var FooDummy[]
     */
    public $foo;

    public function getFoo(): array
    {
        return $this->foo;
    }
}

```

```php
final class FooBaseDummyDenormalizer implements DenormalizerInterface
{
    public function denormalize(mixed $data, string $type, string $format = null, array $context = [])
    {
        $result = [];
        foreach ($data as $foo) {
            $fooDummy = new FooDummy();
            $fooDummy->name = $foo['name'];
            $result[] = $fooDummy;
        }

        return $result;
    }

    public function supportsDenormalization(mixed $data, string $type, string $format = null)
    {
        if (mb_substr($type, -2) === '[]') {
            $className = mb_substr($type, 0, -2);
            $classImplements = class_implements($className);
            \assert(\is_array($classImplements));

            return class_exists($className) && \in_array(FooBaseDummy::class, $classImplements, true);
        }

        return false;
    }

    /**
     * @return array<string, bool>
     */
    public function getSupportedTypes(?string $format): array
    {
        return [FooBaseDummy::class.'[]' => false];
    }
}

```

The test that I expect to pass but currently does not with Serializer 6.3:

```php
    public function testDeserializeOnObjectWithArrayProperty()
    {
        $serializer = new Serializer([new FooBaseDummyDenormalizer(), new ObjectNormalizer(null, null, null, new PhpDocExtractor())], [new JsonEncoder()]);

        $obj = $serializer->deserialize('{"foo":[{"name":"bar"}]}', ArrayPropertyDummy::class, 'json');
        $this->assertInstanceOf(ArrayPropertyDummy::class, $obj);
        $this->assertInstanceOf(FooDummy::class, $obj->getFoo()[0]);
    }
```

The test passes when using Serializer 5.4

Removing the `getSupportedTypes` method from the `FooBaseDummyDenormalizer` when using Serializer 6.3 makes the test pass again (but with a deprecation). The `FooBaseDummyDenormalizer` is a simplified equivalent of a class from a bundle we use in our app.
When we use Serializer 5.4 the serializer calls `$normalizer->supportsDenormalization` which returns true and everything works fine -> https://github.com/symfony/symfony/blob/b39fcefea797caa48dd84f8b0b1dd1149b457283/src/Symfony/Component/Serializer/Serializer.php#L315
When we use Serializer 6.3 it skips that code path as the denormalizer has the `getSupportedTypes` method so it enters this `if` statement which just does a `continue` -> https://github.com/symfony/symfony/blob/bbb5f5cfec94f4f744c9a9db7632418665417662/src/Symfony/Component/Serializer/Serializer.php#L363-L367
`$supportedType` here is `Symfony\Component\Serializer\Tests\Fixtures\FooBaseDummy[]`, `$class` is `Symfony\Component\Serializer\Tests\Fixtures\FooDummy[]` and `$genericType` is `*`

### Possible Solution

_No response_

### Additional Context

_No response_

X-Coder264:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #51261
| License       | MIT
| Doc PR        | -

On Symfony <= 6.2 serializer versions the `supportsDenormalization` was always called in `\Symfony\Component\Serializer\Serializer::getDenormalizer`.

When the `getSupportedTypes` logic was introduced for Symfony 6.3 serializer some `if` statement conditions were introduced and if they are `true` the `supportsDenormalization` logic is not called at all anymore. Those conditions currently prevent a legit use-case that worked prior to 6.3 from calling the `supportsDenormalization` method - the case being when the denormalizer is supposed to denormalize a collection/array of objects. This use-case still works even on the 6.3 serializer as long as the denormalizer does not implement the new `getSupportedTypes` method, but that is triggering a deprecation.

This PR aims to fix that so that denormalizing an array of objects still works even when using the new `getSupportedTypes` method in a denormalizer.

cc @tucksaun @nicolas-grekas

### DIFF
--- a/src/Symfony/Component/Serializer/Serializer.php
+++ b/src/Symfony/Component/Serializer/Serializer.php
@@ -359,9 +359,12 @@ class Serializer implements SerializerInterface, ContextAwareNormalizerInterface
 
                 $supportedTypes = $normalizer->getSupportedTypes($format);
 
+                $doesClassRepresentCollection = str_ends_with($class, '[]');
+
                 foreach ($supportedTypes as $supportedType => $isCacheable) {
                     if (\in_array($supportedType, ['*', 'object'], true)
                         || $class !== $supportedType && ('object' !== $genericType || !is_subclass_of($class, $supportedType))
+                        && !($doesClassRepresentCollection && str_ends_with($supportedType, '[]') && is_subclass_of(strstr($class, '[]', true), strstr($supportedType, '[]', true)))
                     ) {
                         continue;
                     }
--- a/src/Symfony/Component/Serializer/Tests/Fixtures/FooInterfaceDummyDenormalizer.php
+++ b/src/Symfony/Component/Serializer/Tests/Fixtures/FooInterfaceDummyDenormalizer.php
@@ -1,0 +1,50 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Serializer\Tests\Fixtures;
+
+use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
+
+final class FooInterfaceDummyDenormalizer implements DenormalizerInterface
+{
+    public function denormalize(mixed $data, string $type, string $format = null, array $context = []): array
+    {
+        $result = [];
+        foreach ($data as $foo) {
+            $fooDummy = new FooImplementationDummy();
+            $fooDummy->name = $foo['name'];
+            $result[] = $fooDummy;
+        }
+
+        return $result;
+    }
+
+    public function supportsDenormalization(mixed $data, string $type, string $format = null, array $context = []): bool
+    {
+        if (str_ends_with($type, '[]')) {
+            $className = substr($type, 0, -2);
+            $classImplements = class_implements($className);
+            \assert(\is_array($classImplements));
+
+            return class_exists($className) && \in_array(FooDummyInterface::class, $classImplements, true);
+        }
+
+        return false;
+    }
+
+    /**
+     * @return array<string, bool>
+     */
+    public function getSupportedTypes(?string $format): array
+    {
+        return [FooDummyInterface::class.'[]' => false];
+    }
+}
--- a/src/Symfony/Component/Serializer/Tests/Fixtures/FooDummyInterface.php
+++ b/src/Symfony/Component/Serializer/Tests/Fixtures/FooDummyInterface.php
@@ -1,0 +1,16 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Serializer\Tests\Fixtures;
+
+interface FooDummyInterface
+{
+}
--- a/src/Symfony/Component/Serializer/Tests/Fixtures/FooImplementationDummy.php
+++ b/src/Symfony/Component/Serializer/Tests/Fixtures/FooImplementationDummy.php
@@ -1,0 +1,20 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Serializer\Tests\Fixtures;
+
+final class FooImplementationDummy implements FooDummyInterface
+{
+    /**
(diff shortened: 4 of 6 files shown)


src/Symfony/Component/Serializer/Tests/Fixtures/FooInterfaceDummyDenormalizer.php
```diff
@@ -0,0 +1,50 @@
...
+
+        return $result;
+    }
+
+    public function supportsDenormalization(mixed $data, string $type, string $format = null, array $context = []): bool
+    {
+        if (str_ends_with($type, '[]')) {
+            $className = substr($type, 0, -2);
+            $classImplements = class_implements($className);
+            \assert(\is_array($classImplements));
```
andersmateusz, line 35: I am curious. What this assertion is aimed to do?  What is the point of using this here?

src/Symfony/Component/Serializer/Serializer.php
```diff
@@ -359,9 +359,14 @@ private function getDenormalizer(mixed $data, string $class, ?string $format, ar
 
                 $supportedTypes = $normalizer->getSupportedTypes($format);
 
+                $doesClassRepresentCollection = str_ends_with($class, '[]');
+
                 foreach ($supportedTypes as $supportedType => $isCacheable) {
+                    $doesSupportedTypeRepresentCollection = str_ends_with($supportedType, '[]');
```
andersmateusz: Maybe I would avoid using so long vairable names and in the matter of fact I think that this variable is not needed here as is used only once.

andersmateusz (approved):
Tested this on my local environment. Seems legit.

src/Symfony/Component/Serializer/Tests/Fixtures/FooInterfaceDummyDenormalizer.php
mtarld: It is to be sure that `$classImplements` is an array, to avoid any `TypeError` during the `in_array` call. Because `class_implements` might return `false` on failure.

X-Coder264: yep, what @mtarld said, essentially it's to keep static analysis tools happy :smile:

src/Symfony/Component/Serializer/Serializer.php
X-Coder264: I'm fine with changing this or leaving the variable as-is just due to readability - I'm happy to go either way the SF maintainers want.

src/Symfony/Component/Serializer/Tests/Fixtures/FooInterfaceDummyDenormalizer.php
andersmateusz: But calling `\class_exists($className)` before `\in_array..` on line 37 you will never get `TypeError`.  PHP will return `false` when encountering first `false` value in `and` construct without checking second condition. According to documentation `class_implements` returns `An array on success, or false when the given class doesn't exist. `

src/Symfony/Component/Serializer/Serializer.php
andersmateusz: Sure, just my personal opinion here, not fault at all. :)

src/Symfony/Component/Serializer/Tests/Fixtures/FooInterfaceDummyDenormalizer.php
X-Coder264: @andersmateusz This was taken out from the bundle that we use which reproduced this issue. I bet that PHPStan (which that bundle uses) was throwing an error here because `class_implements` can return `false` and PHPStan is not smart enough to know that the second condition will not be executed in that case.

src/Symfony/Component/Serializer/Serializer.php
mtarld: I think that `$doesSupportedTypeRepresentCollection` can be inlined because it's in the loop and used once.

X-Coder264: inlined

mtarld (approved):

nicolas-grekas:
Thank you @X-Coder264.

PR #51374, opened 2023-08-13
[Messenger] Fix PHPDoc template for `TransportFactoryInterface`

Ferror:
Improved a little bit phpdocs when dealing with Transport & Transport Factory Interface.

### DIFF
--- a/src/Symfony/Component/Messenger/Transport/TransportFactoryInterface.php
+++ b/src/Symfony/Component/Messenger/Transport/TransportFactoryInterface.php
@@ -17,9 +17,14 @@ use Symfony\Component\Messenger\Transport\Serialization\SerializerInterface;
  * Creates a Messenger transport.
  *
  * @author Samuel Roze <samuel.roze@gmail.com>
+ *
+ * @template-covariant TTransport of TransportInterface
  */
 interface TransportFactoryInterface
 {
+    /**
+     * @return TTransport
+     */
     public function createTransport(#[\SensitiveParameter] string $dsn, array $options, SerializerInterface $serializer): TransportInterface;
 
     public function supports(#[\SensitiveParameter] string $dsn, array $options): bool;
--- a/src/Symfony/Component/Messenger/Bridge/AmazonSqs/Transport/AmazonSqsTransportFactory.php
+++ b/src/Symfony/Component/Messenger/Bridge/AmazonSqs/Transport/AmazonSqsTransportFactory.php
@@ -18,6 +18,8 @@ use Symfony\Component\Messenger\Transport\TransportInterface;
 
 /**
  * @author Jérémy Derussé <jeremy@derusse.com>
+ *
+ * @implements TransportFactoryInterface<AmazonSqsTransport>
  */
 class AmazonSqsTransportFactory implements TransportFactoryInterface
 {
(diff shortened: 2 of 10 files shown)


src/Symfony/Component/Messenger/Transport/TransportFactoryInterface.php
```diff
@@ -17,9 +17,14 @@
  * Creates a Messenger transport.
  *
  * @author Samuel Roze <samuel.roze@gmail.com>
+ *
+ * @template T
```
derrabus: Right now, `T` would be of type mixed, which is _a little_ too wide. Moreover, we should use a more expressive template variable name and allow covariance.

```suggestion
 * @template-covariant TTransport of TransportInterface
```

src/Symfony/Component/Messenger/Bridge/AmazonSqs/Transport/AmazonSqsTransport.php
```diff
@@ -78,10 +81,7 @@ public function setup(): void
         }
     }
 
-    /**
-     * @return void
-     */
-    public function reset()
+    public function reset(): void
```
derrabus: That's a BC break. We can't do that in a minor release.

src/Symfony/Component/Messenger/Bridge/AmazonSqs/Transport/AmazonSqsTransport.php
```diff
@@ -44,6 +44,9 @@ public function __construct(Connection $connection, SerializerInterface $seriali
         $this->sender = $sender;
     }
 
+    /**
+     * @inheritDoc
+     */
```
derrabus: This block serves no purpose. Let's remove it?

src/Symfony/Component/Messenger/Bridge/AmazonSqs/Transport/AmazonSqsTransport.php
```diff
@@ -90,10 +90,7 @@ public function reset()
         }
     }
 
-    /**
-     * @return MessageCountAwareInterface&ReceiverInterface
-     */
-    private function getReceiver(): ReceiverInterface
+    private function getReceiver(): MessageCountAwareInterface&ReceiverInterface
```
derrabus: Let's postpone this change to 7.0 where we can use this type for the constructor parameter and the property as well?

derrabus (changes requested):
If you make the `TransportFactoryInterface` generic, you need to update the PHPDoc on every implementation accordingly.

Ferror:
> If you make the `TransportFactoryInterface` generic, you need to update the PHPDoc on every implementation accordingly.

Did as requested :)

Ferror:
@derrabus WDYT?

derrabus (approved):

kbond (approved):

fabpot (approved):

fabpot:
Thank you @Ferror.

ro0NL:
wasnt updating the returntypes on `createTransport` sufficient? eg. `SchedulerTransportFactory`

Ferror:
> wasnt updating the returntypes on `createTransport` sufficient? eg. `SchedulerTransportFactory`

Unfortunately not. SchedulerTransportFactory shares Transport Interface therefore by pipeline config it was also required to change to add that.

ro0NL:
i mean why prefer templating a factory when the return type is already sufficient?

https://github.com/symfony/symfony/blob/2bac801f8a4cba02e8ecc840fa7bf512034f860a/src/Symfony/Component/Scheduler/Messenger/SchedulerTransportFactory.php#L34

or what does this PR solve? :thinking:

stof:
If we never benefit from the generic types in places in which we inject a TransportFactory but only in places using a concrete factory explicitly, we could indeed avoid making it generic but using a refined return type...

ro0NL:
i guess we do :rocket:

Issue #50382, opened 2023-05-21
[Console][Monolog-Bridge] Logs gain extra empty line when used with console sections


kiler129:
### Symfony version(s) affected

6.2.10

### Description

*First of all, apologies if this isn't a bug but some misconfiguration. I tried my best to isolate it however.*

## Problem
While using multiple progress bars, [as shown in the docs](https://symfony.com/doc/current/components/console/helpers/progressbar.html#displaying-multiple-progress-bars), one has to use sections. This produces an expected result. However, when using logging alongside some logs *(`ref (b)` in code)*  will be swallowed:

https://github.com/symfony/symfony/assets/1227834/23e3780e-5b80-43ed-9389-c80e836d7ccd
```
19:05:08 WARNING   [app] Hello 30
  100/1000 [==>-------------------------]  10%
19:05:08 WARNING   [app] Hello 50
19:05:08 WARNING   [app] Hello 60
19:05:08 WARNING   [app] Hello 70
  100/1000 [==>-------------------------]  10%
  200/1000 [=====>----------------------]  20%
19:05:08 WARNING   [app] Hello 100
```

## Fixing the logs breaks sections spacing
The solution I tracked down is to ensure logs (more specifically `\Symfony\Bridge\Monolog\Handler\ConsoleHandler`) uses a section above progress bars sections. This entails something I'm not 100% comfortable with - injecting `monolog.handler.console` and doing a manual `->setOutput()` on it *(`ref (a)` in code)*.

This produces almost correct result with no output being swallowed, but with extra spacing between log lines:

https://github.com/symfony/symfony/assets/1227834/e534bac2-0a0d-410f-b9c2-2d5cf9f302ef
```
19:09:07 WARNING   [app] Hello 270

19:09:07 WARNING   [app] Hello 280

19:09:08 WARNING   [app] Hello 290

  200/1000 [=====>----------------------]  20%
  500/1000 [==============>-------------]  50%
```

## Issue with `monolog-bridge` + section?
I ran two more tests: normal logs (no `->setOutput($section)` override) and logs in section but without bars.

https://github.com/symfony/symfony/assets/1227834/ca777300-c12c-4e83-b693-8618bfe102bd
```
19:13:00 WARNING   [app] Hello 10

19:13:01 WARNING   [app] Hello 20

19:13:01 WARNING   [app] Hello 30
```

---

https://github.com/symfony/symfony/assets/1227834/fe139906-6ff6-4ac7-af9d-130251e8d6f9
```
[...]

PR #51378, opened 2023-08-14
[Console] avoid multiple new line when message already ends with a new line in section output

joelwurtz:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no 
| Tickets       | https://github.com/symfony/symfony/issues/50382
| License       | MIT
| Doc PR        | 

When we write to the section with a new line in the message it ends up adding a double new line, it can be easily seen when we set a section as a console handler for the logs, each line of log will have a double new line ending adding a lot of noise.

I began to fix the implementation to correctly handle this case, but with all the logic of the max height and other stuff it appears to be way more complicated. 

Simulating this case by removing the new line at the end of the message and recalling the write function with the newline parameter to true was way easier and avoid too many headcache when looking at the code.

Should fix https://github.com/symfony/symfony/issues/50382

### DIFF
--- a/src/Symfony/Component/Console/Output/ConsoleSectionOutput.php
+++ b/src/Symfony/Component/Console/Output/ConsoleSectionOutput.php
@@ -119,8 +119,7 @@ class ConsoleSectionOutput extends StreamOutput
             // re-add the line break (that has been removed in the above `explode()` for
             // - every line that is not the last line
             // - if $newline is required, also add it to the last line
-            // - if it's not new line, but input ending with `\PHP_EOL`
-            if ($i < $count || $newline || str_ends_with($input, \PHP_EOL)) {
+            if ($i < $count || $newline) {
                 $lineContent .= \PHP_EOL;
             }
 
@@ -168,6 +167,12 @@ class ConsoleSectionOutput extends StreamOutput
      */
     protected function doWrite(string $message, bool $newline)
     {
+        // Simulate newline behavior for consistent output formatting, avoiding extra logic
+        if (!$newline && str_ends_with($message, \PHP_EOL)) {
+            $message = substr($message, 0, -\strlen(\PHP_EOL));
+            $newline = true;
+        }
+
         if (!$this->isDecorated()) {
             parent::doWrite($message, $newline);
 
--- a/src/Symfony/Component/Console/Tests/Output/ConsoleSectionOutputTest.php
+++ b/src/Symfony/Component/Console/Tests/Output/ConsoleSectionOutputTest.php
@@ -158,7 +158,7 @@ class ConsoleSectionOutputTest extends TestCase
         $expected .= 'Two'.\PHP_EOL.'Three'.\PHP_EOL.'Four'.\PHP_EOL;
 
         // cause overflow of first section (redraw whole section, without first line)
-        $firstSection->writeln("Four\nFive\nSix");
+        $firstSection->writeln('Four'.\PHP_EOL.'Five'.\PHP_EOL.'Six');
         $expected .= "\x1b[6A\x1b[0J";
         $expected .= 'Four'.\PHP_EOL.'Five'.\PHP_EOL.'Six'.\PHP_EOL;
         $expected .= 'Two'.\PHP_EOL.'Three'.\PHP_EOL.'Four'.\PHP_EOL;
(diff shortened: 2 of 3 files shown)


src/Symfony/Component/Console/Output/ConsoleSectionOutput.php
```diff
@@ -168,6 +168,13 @@ public function addNewLineOfInputSubmit(): void
      */
     protected function doWrite(string $message, bool $newline)
     {
+        // if there is a new line at the end of message, but we do not add a new line, then simulate if without new line and add new line
+        // a lot of code relies on the newline bool to correctly format the output, it avoid a lot of new logic to handle both cases which are roughly the same
+        if (!$newline && str_ends_with($message, \PHP_EOL)) {
```
maxbeckers, line 171: This should avoid the last part of the following if statement correctly? Can we fix that there instead of this or are there more related parts?
https://github.com/symfony/symfony/blob/16a4681b5b7a20f60c12f0ab1e2d22dbdf4ba50b/src/Symfony/Component/Console/Output/ConsoleSectionOutput.php#L123

src/Symfony/Component/Console/Tests/Output/ConsoleSectionOutputTest.php
```diff
@@ -158,7 +158,7 @@ public function testMaxHeightMultipleSections()
         $expected .= 'Two'.\PHP_EOL.'Three'.\PHP_EOL.'Four'.\PHP_EOL;
 
         // cause overflow of first section (redraw whole section, without first line)
-        $firstSection->writeln("Four\nFive\nSix");
+        $firstSection->writeln('Four'.\PHP_EOL.'Five'.\PHP_EOL.'Six');
```
maxbeckers, line 161: Great to fix the test for windows :)

src/Symfony/Component/Console/Output/ConsoleSectionOutput.php
joelwurtz: Certainly, didn't want to update too much of other logics, but having this can certainly clean up some code afterwards

src/Symfony/Component/Console/Tests/Output/ConsoleSectionOutputTest.php
joelwurtz: Yeah, my mistake on a recent pr ^^, always hard to tell if windows fails due to a transient test or not xD

src/Symfony/Component/Console/Output/ConsoleSectionOutput.php
```diff
@@ -168,6 +168,13 @@ public function addNewLineOfInputSubmit(): void
      */
     protected function doWrite(string $message, bool $newline)
     {
+        // if there is a new line at the end of message, but we do not add a new line, then simulate if without new line and add new line
+        // a lot of code relies on the newline bool to correctly format the output, it avoid a lot of new logic to handle both cases which are roughly the same
```
nicolas-grekas: what about just this?
`// Simulate newline behavior for consistent output formatting, avoiding extra logic`

nicolas-grekas: or maybe the commit title?
`avoid multiple new line when message already ends with a new line`

src/Symfony/Component/Console/Output/ConsoleSectionOutput.php
joelwurtz: I removed the last part of this if, as it's not needed anymore, thanks for the insight

src/Symfony/Component/Console/Output/ConsoleSectionOutput.php
joelwurtz: I used the first suggestion, thanks

lyrixx (approved):

welcoMattic (approved):

nicolas-grekas:
Thank you @joelwurtz.

danepowell:
😢 This broke our project, which uses progress bars to build a checklist-style output. It results in too many lines being cleared when the progress bar advances: https://github.com/acquia/cli/issues/1620

I'd appreciate your input, as it's not obvious to me why it broke.

chalasr:
@danepowell Please consider submitting a new issue on this repo so that it's properly tracked

danepowell:
I opened https://github.com/symfony/symfony/issues/52446

Issue #51342, opened 2023-08-10
MapRequestHeader


qdequippe:
### Description

Attribute to map request header to specific controller argument.

Similar to https://symfony.com/doc/current/controller.html#controller_map-request

### Example

```php
#[Route(path: '/', name: 'admin_dashboard')]
public function index(
    #[MapRequestHeader] string $authorization,
    #[MapRequestHeader] array $accepts,
    #[MapRequestHeader] string $xCustomName,
)
```

What do you think about it ☺️?

carsonbot:
Thank you for this suggestion.
There has not been a lot of activity here for a while. Would you still like to see this feature?

qdequippe:
@carsonbot yes there is a PR for this

carsonbot:
Thank you for this suggestion.
There has not been a lot of activity here for a while. Would you still like to see this feature?

alexandre-daubois:
Not stalled, the PR is still under review

chomik-pawel:
I would love to see this feature as well :) Is there any chance to pull it off, been a while :)

carsonbot:
Thank you for this suggestion.
There has not been a lot of activity here for a while. Would you still like to see this feature?
[Every feature is developed by the community.](https://symfony.com/doc/current/contributing/community/releases.html#development)
Perhaps someone would like to try?
You can [read how to contribute](https://symfony.com/doc/current/contributing/code/pull_requests.html) to get started.

chomik-pawel:
Yes, I would love to see this feature. I wonder why it takes so long to merge it! :)

carsonbot:
Thank you for this suggestion.
There has not been a lot of activity here for a while. Would you still like to see this feature?
[Every feature is developed by the community.](https://symfony.com/doc/current/contributing/community/releases.html#development)
Perhaps someone would like to try?
You can [read how to contribute](https://symfony.com/doc/current/contributing/code/pull_requests.html) to get started.

alexandre-daubois:
Marking as `Keep Open` as PR exists.

PR #20541, opened 2016-11-16
Fix phpDoc for UserInterface getRoles

seferov:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | yes?
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

Otherwise it is not possible to navigate to Role, thus imported `Symfony\Component\Security\Core\Role\Role` is marked as not used by PhpStorm IDE.

PR #51379, opened 2023-08-14
[HttpKernel] Adding new `#[MapRequestHeader]` attribute and resolver

StevenRenaux:
| Q             | A
| ------------- | ---
| Branch?       | 8.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | #51342 
| License       | MIT
| Doc PR        | #20541

This adds a `#[MapRequestHeader]` attribute to map HTTP request headers to controller arguments, similar to how `#[MapQueryParameter]` works for query parameters.

```php
#[Route('/')]
public function index(
    #[MapRequestHeader] string $accept,
    #[MapRequestHeader] array $acceptLanguage,
    #[MapRequestHeader('user-agent')] string $userAgent,
    #[MapRequestHeader] ?string $xCustom,
    #[MapRequestHeader] AcceptHeader $acceptEncoding,
): Response {
    // ...
}
```

Supported types are string, array and AcceptHeader.

When the header name is not explicitly provided, the argument name is converted from camelCase to kebab-case (e.g. `$acceptLanguage` for `accept-language`).

When typed as array, the arguments received all corresponding headers.

Missing required headers throw an HttpException with a configurable status code (defaults to 400).

### DIFF
--- a/src/Symfony/Component/HttpKernel/Attribute/MapRequestHeader.php
+++ b/src/Symfony/Component/HttpKernel/Attribute/MapRequestHeader.php
@@ -1,0 +1,32 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\HttpKernel\Attribute;
+
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Controller\ArgumentResolver\RequestHeaderValueResolver;
+
+#[\Attribute(\Attribute::TARGET_PARAMETER)]
+final class MapRequestHeader extends ValueResolver
+{
+    /**
+     * @param string|null  $name                       The name of the header parameter; if null, the name of the argument in the controller will be used
+     * @param class-string $resolver                   The class name of the resolver to use
+     * @param int          $validationFailedStatusCode The HTTP code to return if the validation fails
+     */
+    public function __construct(
+        public readonly ?string $name = null,
+        string $resolver = RequestHeaderValueResolver::class,
+        public readonly int $validationFailedStatusCode = Response::HTTP_BAD_REQUEST,
+    ) {
+        parent::__construct($resolver);
+    }
+}
--- a/src/Symfony/Component/HttpKernel/CHANGELOG.md
+++ b/src/Symfony/Component/HttpKernel/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 8.1
 ---
 
+ * Add `#[MapRequestHeader]` to map a header from `Request` to a controller argument
  * Add `hasErrors()` method to `Profile` to track profiles with errors (exceptions or error-level logs)
  * Validate typed route parameters before calling controllers and return an HTTP error when an invalid value is provided
  * Add `ControllerAttributeEvent` et al. to dispatch events named after controller attributes
--- a/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php
+++ b/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php
@@ -1,0 +1,64 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\HttpKernel\Controller\ArgumentResolver;
+
+use Symfony\Component\HttpFoundation\AcceptHeader;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Attribute\MapRequestHeader;
+use Symfony\Component\HttpKernel\Controller\ValueResolverInterface;
+use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
+use Symfony\Component\HttpKernel\Exception\HttpException;
+
+final class RequestHeaderValueResolver implements ValueResolverInterface
+{
+    public function resolve(Request $request, ArgumentMetadata $argument): array
+    {
+        if (!$attribute = $argument->getAttributesOfType(MapRequestHeader::class)[0] ?? null) {
+            return [];
+        }
+
+        $type = $argument->getType();
+
+        if (!\in_array($type, ['string', 'array', AcceptHeader::class])) {
+            throw new \LogicException(\sprintf('Could not resolve the argument typed "%s". Valid types are "array", "string" or "%s".', $type, AcceptHeader::class));
+        }
+
+        $name = $attribute->name ?? strtolower(preg_replace('/[a-z]\K[A-Z]/', '-$0', $argument->getName()));
+        $value = null;
+
+        if ($request->headers->has($name)) {
+            $value = match ($type) {
+                'string' => $request->headers->get($name),
+                'array' => match (strtolower($name)) {
+                    'accept' => $request->getAcceptableContentTypes(),
+                    'accept-charset' => $request->getCharsets(),
+                    'accept-language' => $request->getLanguages(),
+                    'accept-encoding' => $request->getEncodings(),
+                    default => $request->headers->all($name),
+                },
+                AcceptHeader::class => AcceptHeader::fromString($request->headers->get($name)),
+            };
+        } elseif ($argument->hasDefaultValue()) {
+            $value = $argument->getDefaultValue();
+        }
+
+        if (null === $value && 'array' === $type) {
+            $value = [];
+        }
+
+        if (null === $value && !$argument->isNullable()) {
+            throw HttpException::fromStatusCode($attribute->validationFailedStatusCode, \sprintf('Missing header "%s".', $name));
+        }
+
+        return [$value];
+    }
+}
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
@@ -105,6 +105,7 @@ use Symfony\Component\HttpKernel\Attribute\AsController;
 use Symfony\Component\HttpKernel\Attribute\AsTargetedValueResolver;
 use Symfony\Component\HttpKernel\CacheClearer\CacheClearerInterface;
 use Symfony\Component\HttpKernel\CacheWarmer\CacheWarmerInterface;
+use Symfony\Component\HttpKernel\Controller\ArgumentResolver\RequestHeaderValueResolver;
 use Symfony\Component\HttpKernel\Controller\ValueResolverInterface;
 use Symfony\Component\HttpKernel\DataCollector\DataCollectorInterface;
 use Symfony\Component\HttpKernel\EventListener\ControllerAttributesListener;
@@ -261,6 +262,9 @@ class FrameworkExtension extends Extension
         $loader->load('fragment_renderer.php');
         $loader->load('error_renderer.php');
 
+        if (!class_exists(RequestHeaderValueResolver::class)) {
+            $container->removeDefinition('argument_resolver.header_value_resolver');
+        }
         if (!class_exists(ControllerAttributesListener::class)) {
             $container->removeDefinition('kernel.controller_attributes_listener');
             $container->removeDefinition('serialize_controller_result_listener');
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/web.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/web.php
@@ -22,6 +22,7 @@ use Symfony\Component\HttpKernel\Controller\ArgumentResolver\DateTimeValueResolv
 use Symfony\Component\HttpKernel\Controller\ArgumentResolver\DefaultValueResolver;
 use Symfony\Component\HttpKernel\Controller\ArgumentResolver\QueryParameterValueResolver;
 use Symfony\Component\HttpKernel\Controller\ArgumentResolver\RequestAttributeValueResolver;
+use Symfony\Component\HttpKernel\Controller\ArgumentResolver\RequestHeaderValueResolver;
 use Symfony\Component\HttpKernel\Controller\ArgumentResolver\RequestPayloadValueResolver;
 use Symfony\Component\HttpKernel\Controller\ArgumentResolver\RequestValueResolver;
 use Symfony\Component\HttpKernel\Controller\ArgumentResolver\ServiceValueResolver;
@@ -106,6 +107,9 @@ return static function (ContainerConfigurator $container) {
 
         ->set('argument_resolver.query_parameter_value_resolver', QueryParameterValueResolver::class)
             ->tag('controller.targeted_value_resolver', ['name' => QueryParameterValueResolver::class])
+
+        ->set('argument_resolver.header_value_resolver', RequestHeaderValueResolver::class)
+            ->tag('controller.targeted_value_resolver', ['name' => RequestHeaderValueResolver::class])
 
         ->set('response_listener', ResponseListener::class)
             ->args([
--- a/src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/RequestHeaderValueResolverTest.php
+++ b/src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/RequestHeaderValueResolverTest.php
@@ -1,0 +1,223 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\HttpKernel\Tests\Controller\ArgumentResolver;
+
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\HttpFoundation\AcceptHeader;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Attribute\MapRequestHeader;
+use Symfony\Component\HttpKernel\Controller\ArgumentResolver\RequestHeaderValueResolver;
+use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
+use Symfony\Component\HttpKernel\Exception\HttpException;
+
+class RequestHeaderValueResolverTest extends TestCase
+{
+    public static function provideHeaderValueWithStringType(): iterable
+    {
+        yield 'with accept' => ['accept', 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8'];
+        yield 'with accept-language' => ['accept-language', 'en-us,en;q=0.5'];
+        yield 'with host' => ['host', 'localhost'];
+        yield 'with user-agent' => ['user-agent', 'Symfony'];
+    }
+
+    public static function provideHeaderValueWithArrayType(): iterable
+    {
+        yield 'with accept' => [
+            'accept',
+            'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8',
+            [
+                [
+                    'text/html',
+                    'application/xhtml+xml',
+                    'application/xml',
+                    '*/*',
+                ],
+            ],
+        ];
+        yield 'with accept-language' => [
+            'accept-language',
+            'en-us,en;q=0.5',
+            [
+                [
+                    'en_US',
+                    'en',
+                ],
+            ],
+        ];
+        yield 'with host' => [
+            'host',
+            'localhost',
+            [
+                ['localhost'],
+            ],
+        ];
+        yield 'with user-agent' => [
+            'user-agent',
+            'Symfony',
+            [
+                ['Symfony'],
+            ],
+        ];
+    }
+
+    public static function provideHeaderValueWithAcceptHeaderType(): iterable
+    {
+        yield 'with accept' => [
+            'accept',
+            'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8',
+            [AcceptHeader::fromString('text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8')],
+        ];
+        yield 'with accept-language' => [
+            'accept-language',
+            'en-us,en;q=0.5',
+            [AcceptHeader::fromString('en-us,en;q=0.5')],
+        ];
+        yield 'with host' => [
+            'host',
+            'localhost',
+            [AcceptHeader::fromString('localhost')],
+        ];
+        yield 'with user-agent' => [
+            'user-agent',
+            'Symfony',
+            [AcceptHeader::fromString('Symfony')],
+        ];
+    }
+
+    public static function provideHeaderValueWithDefaultAndNull(): iterable
+    {
+        yield 'with hasDefaultValue' => [true, 'foo', false, 'foo'];
+        yield 'with no isNullable' => [false, null, true, null];
+    }
+
+    public function testWrongType()
+    {
+        $this->expectException(\LogicException::class);
+
+        $metadata = new ArgumentMetadata('accept', 'int', false, false, null, false, [
+            new MapRequestHeader(),
+        ]);
+
+        $request = Request::create('/');
+
+        $resolver = new RequestHeaderValueResolver();
+        $resolver->resolve($request, $metadata);
+    }
+
+    #[DataProvider('provideHeaderValueWithStringType')]
+    public function testWithStringType(string $parameter, string $value)
+    {
+        $resolver = new RequestHeaderValueResolver();
+
+        $metadata = new ArgumentMetadata('variableName', 'string', false, false, null, false, [
+            new MapRequestHeader($parameter),
+        ]);
+
+        $request = Request::create('/');
+        $request->headers->set($parameter, $value);
+
+        $arguments = $resolver->resolve($request, $metadata);
+
+        self::assertEquals([$value], $arguments);
+    }
+
+    #[DataProvider('provideHeaderValueWithArrayType')]
+    public function testWithArrayType(string $parameter, string $value, array $expected)
+    {
+        $resolver = new RequestHeaderValueResolver();
+
+        $metadata = new ArgumentMetadata('variableName', 'array', false, false, null, false, [
+            new MapRequestHeader($parameter),
+        ]);
+
+        $request = Request::create('/');
+        $request->headers->set($parameter, $value);
+
+        $arguments = $resolver->resolve($request, $metadata);
+
+        self::assertEquals($expected, $arguments);
+    }
+
+    #[DataProvider('provideHeaderValueWithAcceptHeaderType')]
+    public function testWithAcceptHeaderType(string $parameter, string $value, array $expected)
+    {
+        $resolver = new RequestHeaderValueResolver();
+
+        $metadata = new ArgumentMetadata('variableName', AcceptHeader::class, false, false, null, false, [
+            new MapRequestHeader($parameter),
+        ]);
+
+        $request = Request::create('/');
+        $request->headers->set($parameter, $value);
+
+        $arguments = $resolver->resolve($request, $metadata);
+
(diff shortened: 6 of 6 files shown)


AurelienPillevesse:
Maybe I'm wrong but seems that something related to ```mapQueryString``` and ```mapRequestPayload``` is in this file and I see nothing from yours : 
https://github.com/symfony/symfony/blob/16a4681b5b7a20f60c12f0ab1e2d22dbdf4ba50b/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php#L84

Why did not add in the same file?

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php
```diff
@@ -0,0 +1,42 @@
...
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Attribute\MapRequestHeader;
+use Symfony\Component\HttpKernel\Controller\ValueResolverInterface;
+use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+final class RequestHeaderValueResolver implements ValueResolverInterface
+{
+    public function resolve(Request $request, ArgumentMetadata $argument): iterable
```
WebMamba: Why do you return an iterable? I think it's way more intuitive to get the value string

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        $name = $attribute->name ?? $argument->getName();
+
+        if (!$header->has($name)) {
+            if ($argument->isNullable() || $argument->hasDefaultValue()) {
+                return [];
+            }
+
+            throw new NotFoundHttpException(sprintf('Missing header parameter "%s".', $name));
```
WebMamba: I think we shouldn't throw a not found Exception here but return null instead

robert-sykes (commented):
Would this not be best to also allow it to map to an object, like how the `MapRequestPayload` currently works, as I can pass an `Object` as a reference the serializer will deserialize to this object? That way you can have a `Header` class that the headers deserialize into.

You can merge it with your current concept and have a bit of best of both? Map individual Headers to an string, or an object or even just an array if required? 

Something potentially like this?
```php
    private function mapRequestHeader(Request $request, string $type, MapRequestHeader $attribute): ?object
    {
        $headers = [];

        $requestHeaders = $request->headers->all();

        array_walk($requestHeaders, function ($value, $key) use (&$headers) {
            $key = $this->snakeCaseToCamelCase($key);
            $headers[$key] = $value[0];
        });

        $headers = json_encode($headers);

        $format = 'json';

        try {
            return $this->serializer->deserialize($headers, $type, $format, []);
        } catch (UnsupportedFormatException $e) {
            throw new HttpException(Response::HTTP_UNSUPPORTED_MEDIA_TYPE, sprintf('Unsupported format: "%s".', $format), $e);
        } catch (NotEncodableValueException $e) {
            throw new HttpException(Response::HTTP_BAD_REQUEST, sprintf('Request payload contains invalid "%s" data.', $format), $e);
        }
    }
```

and you could have a class like this:

```php
class HeaderRequest
{
    public function __construct(
        public readonly string $userAgent,
        public readonly string $accept,
        public readonly string $contentType,
        public readonly string $acceptEncoding,
    ) {
    }
}
```

An advantage is you can use the Validator too and enforce any header types if needed or create a HeaderRequest class per request as needed. 

You can add this as a new method to the existing `RequestPayloadValueResolver` as already mentioned by @AurelienPillevesse so it will tie nicely into the existing process.

This is a rough example but just food for thought.

Very useful PR this and something we need, so looking forward to seeing how it progresses :)

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php
StevenRenaux: Because `RequestHeaderValueResolver` implements [ValueResolverInterface](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/HttpKernel/Controller/ValueResolverInterface.php) with the `resolve` method who has a return type iterable.

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php
StevenRenaux: I was inspired by certain classes as [RequestPayloadValueResolver](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php) or [QueryParameterValueResolver](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/QueryParameterValueResolver.php) where the throw an exception when the value is incorrect.

WebMamba: Hum yes, but I think the need is different here. Let's say you looking for a custom header X-user-token. This header is made to authenticated the user, if this header is not present on the request you want to return a 401 not a 404. Or a 415 if you expect a specific media-type. And I think we can find many different ways where the user doesn't want to return 404.

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php
WebMamba: Ha yes my bad!

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php
```diff
@@ -194,4 +200,24 @@ private function mapRequestPayload(Request $request, string $type, MapRequestPay
...
+
+        array_walk($requestHeaders, function ($value, $key) use (&$headers) {
+            $key = $this->kebabCaseToCamelCase($key);
+            $headers[$key] = $value[0];
+        });
+
+        return $this->serializer->denormalize($headers, $type, null, self::CONTEXT_DESERIALIZE + $attribute->serializationContext);
+    }
+
+    private function kebabCaseToCamelCase($string): string
```
robert-sykes: I am not sure this is the best place to actually put this, you might be able to do it inline or maybe there is a helper function somewhere in the source code you could use?

robert-sykes (changes requested):

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php
```diff
@@ -194,4 +200,17 @@ private function mapRequestPayload(Request $request, string $type, MapRequestPay
...
         }
     }
+
+    private function mapRequestHeader(Request $request, string $type, MapRequestHeader $attribute): ?object
+    {
+        $headers = [];
+        $requestHeaders = $request->headers->all();
+
+        array_walk($requestHeaders, function ($value, $key) use (&$headers) {
+            $key = lcfirst(str_replace('-', '', ucwords($key, '-')));
```
robert-sykes: Actually thinking further on this - I think assuming that the headers will always be kebab case might be a bit of a risk (I think headers usually always are but there doesn't seem to be a set rule on this that I can find) - IT might be better to just convert to camel from any format. 

I know symfony has this ability through the `String` component, but I am not sure if we can pull that into this resolver. You would need to check on that, maybe ask on the discussions or support channel? 

There example I found that does this with the package however was this:

`u($key)->camel()->toString();`

:)

StevenRenaux: @robert-sykes 

The header will always be, if I'm not wrong, in kebab case because in the Resolver it's already formatted like this from the [HeaderBag](https://github.com/symfony/symfony/blob/9a0f178c5324c6ee7a3d99fe4fb7ef31b454925c/src/Symfony/Component/HttpFoundation/HeaderBag.php#L136) class when it set every key in the construct to return an array with keys in kebab case.

The only problem is,  if someone add a key in camelCase or PascalCase in his header it will be returned in lowercase without the hyphen of separation.
But for me it's more about a problem of conception if someone do that I guess.

robert-sykes: @StevenRenaux right you are. That makes sense! Looks good though - this will be really handy and great to map headers to a class and validate. Thanks for putting in the work :)

robert-sykes (approved):

ruudk:
Too bad the PR was changed after @robert-sykes suggestion. I think these are 2 distinct ideas. The original PR description mentions a single attribute per header, same as how MapQueryParameter works. There are valid cases for that, where you only want and need a single header scalar.

The updated code now requires you to map a header to a header object. 

Would be great to also keep supporting the original one for just a single header.

See this comment where a similar situation happened: https://github.com/symfony/symfony/pull/49134#issuecomment-1508323229

robert-sykes:
> Too bad the PR was changed after @robert-sykes suggestion. I think these are 2 distinct ideas. The original PR description mentions a single attribute per header, same as how MapQueryParameter works. There are valid cases for that, where you only want and need a single header scalar.
> 
> The updated code now requires you to map a header to a header object.
> 
> Would be great to also keep supporting the original one for just a single header.
> 
> See this comment where a similar situation happened: [#49134 (comment)](https://github.com/symfony/symfony/pull/49134#issuecomment-1508323229)

I'm all for this, to be fair.  I did imply this in my original comment where I said  `You can merge it with your current concept and have a bit of best of both? Map individual Headers to an string, or an object or even just an array if required?` That said - I didn't push it any further in my commentary, apologies.

@StevenRenaux You could check the type and if it's a string map to that similar to your original commit, otherwise if it's object map to those fields instead. We get best of both worlds this way. What do you think?

StevenRenaux:
Hey, I will try to do it and I didn't understand well for sure but we are there to share and go further 😉

robert-sykes:
> Hey, I will try to do it and I didn't understand well for sure but we are there to share and go further 😉

No problem at all. If you need a hand let me know, more than happy to pair on this with you. :)

StevenRenaux:
To get back to the original Issue I suggest this

```php
namespace App\Controller;

{
    #[Route(path: '/', name: 'test')]
    public function test(
        #[MapRequestHeader] HeaderRequestDto $headerRequestDto,
        #[MapRequestHeader(name: 'user-agent')] string $agent,
        #[MapRequestHeader] array $accept
    ): Response
    {
        dd(
            $headerRequestDto,
//      App\Entity\HeaderRequestDto {
//        accept: "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8"
//        agent: "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/116.0"
//        acceptEncoding: "gzip, deflate, br"
//        host: "localhost:8000"
//      }
           
            $agent,
//      "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/116.0"

            $accept
//      array:6 [
//          0 => "text/html"
//          1 => "application/xhtml+xml"
//          2 => "application/xml;q=0.9"
//          3 => "image/avif"
//          4 => "image/webp"
//          5 => "*/*;q=0.8"
//      ]
        );
    }
}
```
And  a Dto  could look like this :

```php
namespace App\Entity;

use Symfony\Component\Serializer\Annotation\SerializedName;
use Symfony\Component\Validator\Constraints as Assert;

class HeaderRequestDto
{
    public function __construct(
        public readonly string $accept,
        #[Assert\EqualTo('test', groups: ['test'])]
        #[SerializedName('user-agent')] public readonly string $agent = 'agent',
        #[SerializedName('accept-encoding')] public readonly string $acceptEncoding,
        #[Assert\EqualTo('localhost', groups: ['valid'])]
        public readonly string $host
    ) {
    }
}

```

We only use kebab-case (as returned by HeaderBag) now to get some property like 'user-agent'.
And if we want an array rather than a string for the property as `accept` it will return an array after an explode with a coma delimiter.

robert-sykes (approved):

StevenRenaux:
@reviewers Sorry everyone if you were notified about this PR, the rebase didn't worked well the first time

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php
```diff
@@ -0,0 +1,95 @@
...
+    public function resolve(Request $request, ArgumentMetadata $argument): iterable
+    {
+        if (!$attribute = $argument->getAttributesOfType(MapRequestHeader::class)[0] ?? null) {
+            return [];
+        }
+
+        $headers = [];
+        $requestHeaders = $request->headers->all();
+
+        array_walk($requestHeaders, function ($value, $key) use (&$headers) {
```
mathroc, line 50: this could be done after the `if` line 56 to avoid joining all the headers if we need only one

mathroc:
I also implemented something at work, but made some different design choices, so just sharing.

I've handled string vs array of a single header differently: I made 2 attributes `#[FromFirstHeader]` (that can resolve to a non array scalar or a ValueObject (e.g. `\DateTimeImmutable`) & `#[FromHeader]` that resolves to `array<T>`

And when creating a DTO it's another attribute `#[FromHeaders]`. In my case, it resolves to DTO that have array properties

---

also, when resolving a single header to a int or float, I have some handmade rules to cast them, don't know if that would be controversial:

```php
[$value, $validType] = match ($type) {
    'int' => [(int) $value, (string) $value === (string) (int) $value],
    'float' => [(float) $value, (string) $value === (string) (float) $value],
    'string' => [(string) $value, true],
    default => throw new LogicException("FromHeader: Invalid type $type"),
};

if (!$validType) {
    throw new SomeException(
        "Header `$name` is not a valid `$type`.",
    );
}
````

--- 

and last point, I also have a `$constraints` on the attributes that are checked. can be useful for scalar since they can't have validation attributes on them

---

none of them are blockers I think, so I'll let you decide if there's something to pick from this. the only one I would be a bit disappointed if it was not included is the ability to deserialize a single header to a ValueObject

**Comment by mindaugasvcs** [comment]
How to also validate multiple values from a single header, like 'bytes 0-10/16'? Using RegEx I can extract 0, 10, 16.

File: src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php [hunk 6]
```diff
@@ -0,0 +1,95 @@
...
+            $headers[$key] = implode(',', $value);
+        });
+
+        $type = $argument->getType();
+
+        if ('string' === $type || 'array' === $type) {
+            $name = $attribute->name ?? $argument->getName();
```
nicolas-grekas: I had to read the code to get the relation between the type of the parameter and the name of the attribute.
It's not obvious to me that typehinting for an object means "name" is ignored. We should at least throw when it's set to tell it shouldn't be here.
But it could be preferable to split this attribute in two: MapRequestHeader and MapRequestHeaders?

robert-sykes: @nicolas-grekas I think that's a sound idea! @StevenRenaux Are you still interested in working on this?

StevenRenaux: @robert-sykes I will try to work on it next week ;)

alexandre-daubois:
Hi @StevenRenaux! What's the current state of this PR? :slightly_smiling_face:

StevenRenaux:
> Hi @StevenRenaux! What's the current state of this PR? 🙂

@alexandre-daubois Waiting for review, since the last requested update.

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php
```diff
@@ -0,0 +1,114 @@
...
+
+        if (!$attribute) {
+            return [];
+        }
+
+        $headers = [];
+        $requestHeaders = $request->headers->all();
+
+        array_walk($requestHeaders, function ($value, $key) use (&$headers) {
+            $headers[$key] = implode(',', $value);
```
mathroc: this should be done only on the requested header and I'm not sure it should be done at all when mapping to an object

robert-sykes (approved):

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        return $this->mapHeaderValuesToObject($headers, $argument, $attribute);
+    }
+
+    private function mapHeaderValues(array $headers, ArgumentMetadata $argument, MapRequestHeader $attribute): array
+    {
+        $type = $argument->getType();
+
+        if ('string' === $type || 'array' === $type) {
```
mathroc: what happens if `MapRequestHeader` is on an argument of a different type? it looks like it will return an array anyway. if only `string` and `array` are supported I guess we should throw a `LogicException` if it's something different

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php
mathroc: is this condition still needed?

src/Symfony/Component/HttpKernel/CHANGELOG.md
```diff
@@ -46,6 +46,8 @@ CHANGELOG
  * Add argument `$buildDir` to `WarmableInterface`
  * Add argument `$filter` to `Profiler::find()` and `FileProfilerStorage::find()`
  * Add `ControllerResolver::allowControllers()` to define which callables are legit controllers when the `_check_controller_is_allowed` request attribute is set
+ * Add `#[MapRequestHeader]` and `#[MapRequestHeaders]` to map and validate request header from `Request::$headers`
+
```
alexandre-daubois: This is a superfluous newline

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php
```diff
@@ -0,0 +1,114 @@
...
+            $name = $attribute->name ?? $argument->getName();
+
+            if (!isset($headers[$name])) {
+                return [null];
```
mathroc: if type = array I would expect to get an empty array if the header is missing

mathroc: and if the argument is a non-nullable string, I would expect a response with a 4xx status

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php
StevenRenaux: To get an array from
image
To
image

If not modified all of the properties in your DTO object need to be as array type or denormalization will failed.

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php
StevenRenaux: Yes still needed, because if you write something as

```php
    #[Route(path: '/', name: 'test')]
    public function test(
        #[MapRequestHeader] string $accept,
    ): Response
    {
        // rest of your code
    }
```

name will be formed like this

```php

            $attribute->name, // null
            $argument->getName() // "accept" depending of your var name
```

And if you put a name

```php
    #[Route(path: '/', name: 'test')]
    public function test(
        #[MapRequestHeader('user-agent')] string $accept,
    ): Response
    {
        // rest of your code
    }
```

```php

            $attribute->name, // "user-agent"
            $argument->getName() // "accept" depending of your var name
```

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php
mathroc: I see :thinking: 

there was 2 parts in my comment, "this should be done only on the requested header" was about avoiding doing this on all headers when we only need one

and the second part is the part you answered.

it's true that for DTO objects, it would be annoying to have arrays everywhere. But for the odd case where you want an array it's a bit annoying to have to split the string just for that. especially since headers can have commas in their value, it's not just using `explode()`.

I don't have a very good solution for this but I do have a suggestion, we could add arguments to `MapRequestHeader(s)`, that could look like this:

* `#[MapRequestHeaders(toStringDefault: "concat")` the default: concat each header values with commas (`""` when absent?)
* `#[MapRequestHeaders(toStringDefault: "first")` take the first value of header values, `null` when absent
* `#[MapRequestHeaders(toString: ["Date" => first])` concat the values for all headers except for Date: take the first value and ignore the rest
* `#[MapRequestHeaders(toString: ["Date" => "first"])` concat the values for all headers except for Date: take the first value and ignore the rest if any
* `#[MapRequestHeaders(toString: ["Date" => false])` concat the values for all headers except for Date: keep it an array

* `#[MapRequestHeader(toString: "concat")` the default: concat each header values with commas (`""` when absent?)
* `#[MapRequestHeader(toString: "first")` take the first value of header values, `null` when absent

This should make the default behavior good enough in most cases but gives the option to opt-out of the concat shortcut when needed

wdyt?

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php
mathroc: I was thinking about the `if ()` just above. sorry that it wasn't clear. anyway, it doesn't apply anymore with your latest changes :ok_hand:

StevenRenaux:
Back to the basics of PR, using only the MapRequestHeader attribute. Mapping to a DTO does not seem like a relevant or common use case or if it is really necessary could be done in a future PR.

The allowed types are therefore:

- `string`
- `array`
- `AcceptHeader`

When the header parameter cannot be determined by the variable name, we can specify the name argument in the attribute.

````php
public function test(
    #[MapRequestHeader('user-agent')] ?string $bar,
): Response
{
  dump($bar);
  //  "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36"
}
````

If we are looking for a header of type Accept-* with an array type, we will use the dedicated methods to retrieve this data.

https://symfony.com/doc/current/components/http_foundation.html#accessing-accept-headers-data

````php
public function test(
    #[MapRequestHeader] array $accept,
): Response
{
  dump($accept);
  //        array:7 [▼
  //          0 => "text/html"
  //          1 => "application/xhtml+xml"
  //          2 => "image/avif"
  //          3 => "image/webp"
  //          4 => "image/apng"
  //          5 => "application/xml"
  //          6 => "*/*"
  //        ]
}
````

And with AcceptHeader type

````php
public function test(
    #[MapRequestHeader(name: 'accept-encoding')] AcceptHeader $encoding,
): Response
{
  dump($encoding);
  //        Symfony\Component\HttpFoundation\AcceptHeader {#4727 ▼
  //            -items: array:4 [▼
  //                "gzip" =>
  //                    Symfony\Component\HttpFoundation\AcceptHeaderItem {#2437 ▼
  //                        -value: "gzip"
  //                        -quality: 1.0
  //                        -index: 0
  //                        -attributes: []
  //                }
  //                "deflate" =>
  //                    Symfony\Component\HttpFoundation\AcceptHeaderItem {#4882 ▼
  //                        -value: "deflate"
  //                        -quality: 1.0
  //                        -index: 1
  //                        -attributes: []
  //                }
  //                "br" =>
  //                    Symfony\Component\HttpFoundation\AcceptHeaderItem {#4549 ▼
  //                        -value: "br"
  //                        -quality: 1.0
  //                        -index: 2
  //                        -attributes: []
  //                }
  //                "zstd" =>
  //                    Symfony\Component\HttpFoundation\AcceptHeaderItem {#4773 ▼
  //                        -value: "zstd"
  //                        -quality: 1.0
  //                        -index: 3
  //                        -attributes: []
  //                }
  //            ]
  //              -sorted: false
  //        }
}
````

File: src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php (hunk 7 above)
Comment (StevenRenaux) [reply src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php]: I think it depends on everyone's needs. The return must be exaustive I think.

**Comment by MatTheCat** [comment]
It bugs me the `array` type has two semantics whether it applies to an `Accept-*` header or not. Plus

- in the first case there already is the `AcceptHeader` type.
- in the second case only the first value is returned in an array, but
  - there can be many
  - the first value can actually be many values combined by the SAPI

**Comment by ruudk** [comment]
@StevenRenaux What's needed to get this to a mergeable state? I would love this to be included in the new Symfony version.

**Comment by StevenRenaux** [comment]
> @StevenRenaux What's needed to get this to a mergeable state? I would love this to be included in the new Symfony version.

@ruudk Waiting reviews from core-contributors, but it's seems to late for merging into the next 7.3 version.
Feel free also to review, approve...
I think at the moment it's fine to get a simple implementation as it is because it's the kind of stuff you can't statifies every problematics.

**Comment by ruudk** [comment]
Could you rebase it, so that the tests go green?

**Comment by ruudk** [comment]
@StevenRenaux Thanks! 🙏 It needs a few tweaks in https://github.com/symfony/symfony/actions/runs/15253101552/job/42894283925?pr=51379

File: src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php [hunk 11]
```diff
@@ -0,0 +1,62 @@
...
+
+        if ($request->headers->has($name)) {
+            $value = match ($type) {
+                'string' => $request->headers->get($name),
+                'array' => match (strtolower($name)) {
+                    'accept' => $request->getAcceptableContentTypes(),
+                    'accept-charset' => $request->getCharsets(),
+                    'accept-language' => $request->getLanguages(),
+                    'accept-encoding' => $request->getEncodings(),
+                    default => [$request->headers->get($name)],
```
mathroc:
```suggestion
                    default => $request->headers->all($name),
```

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestHeaderValueResolver.php
```diff
@@ -0,0 +1,62 @@
...
+                default => AcceptHeader::fromString($request->headers->get($name)),
+            };
+        }
+
+        if (null === $value && $argument->hasDefaultValue()) {
+            $value = $argument->getDefaultValue();
+        }
+
+        if (null === $value && !$argument->isNullable()) {
+            throw new HttpException($attribute->validationFailedStatusCode, \sprintf('Missing header "%s".', $name));
```
mathroc: I think if type is array we could return an empty array

nicolas-grekas (approved):
I rebased the PR on 8.1 and added support for kebab-case conversion so that eg `$acceptEncoding` works.

mathroc (approved):

src/Symfony/Component/HttpKernel/Attribute/MapRequestHeader.php
```diff
@@ -0,0 +1,32 @@
...
+
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Controller\ArgumentResolver\RequestHeaderValueResolver;
+
+#[\Attribute(\Attribute::TARGET_PARAMETER)]
+final class MapRequestHeader extends ValueResolver
+{
+    /**
+     * @param string|null $name                       The name of the header parameter; if null, the name of the argument in the controller will be used
+     * @param string      $resolver                   The class name of the resolver to use
```
alexandre-daubois:
```suggestion
     * @param class-string      $resolver                   The class name of the resolver to use
```

alexandre-daubois (approved):

nicolas-grekas:
Thank you @StevenRenaux.

PR #51380, opened 2023-08-14
[SecurityBundle] Auto-hash `PasswordAuthenticatedUserInterface` instances

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Will allow removing these lines from the default recipe:
```yaml
    password_hashers:
```

`auto` is the best option anyway so there is little need to configure this map nowadays.

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/CHANGELOG.md
+++ b/src/Symfony/Bundle/SecurityBundle/CHANGELOG.md
@@ -8,6 +8,7 @@ CHANGELOG
  * Allow an array of `pattern` in firewall configuration
  * Add `$badges` argument to `Security::login`
  * Deprecate the `require_previous_session` config option. Setting it has no effect anymore
+ * Auto-hash `PasswordAuthenticatedUserInterface` instances
 
 6.3
 ---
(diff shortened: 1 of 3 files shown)


chalasr (approved):
👍

chalasr:
I'm all for adding good defaults and this is definitely one. Should we keep some words about it in the recipe though? Even if just commented out? I feel like emptying the recipes might hurt discovery through hiding complex concepts entirely.

derrabus:
The problem I see here is that I cannot opt out of this configuration anymore. In my apps, I usually replace that recipe line with the user object(s) that my app should actually be able to handle. I kinda liked the idea that the hasher would jump into my face if I tried to feed it with arbitrary objects that just happen to implement Symfony's interfaces. But this might just be me being a control freak. 😉

nicolas-grekas:
Alright. On my side I don't see this as useful default because I don't see why anyone would change that when building new apps (recipes are for new apps to me.) but :shrug:

PR #51391, opened 2023-08-15
[HttpKernel] Revert #45476 to fix missing `Request` in `RequestStack` for `StreamedResponse`

### LINKED ISSUE #46743: [HttpKernel] "Deprecation" of StreamedResponseListener causes RequestStack to be empty within StreamedResponse callback

DaDeather:
| Q             | A
| ------------- | ---
| Branch       | 6.3
| Bug fix      | yes
| New feature  | no
| Deprecations | no
| Tickets       | Fix #46743
| License       | MIT

Revert #45476 and add test to ensure `Request` availability while using `StreamedResponse` callback.

### DIFF
--- a/UPGRADE-6.1.md
+++ b/UPGRADE-6.1.md
@@ -27,11 +27,6 @@ FrameworkBundle
    To prevent services resetting after each message the "--no-reset" option in "messenger:consume" command can be set
  * Deprecate not setting the `http_method_override` config option. The default value will change to `false` in 7.0.
 
-HttpKernel
-----------
-
- * Deprecate StreamedResponseListener, it's not needed anymore
-
 Routing
 -------
 
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/web.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/web.php
@@ -31,6 +31,7 @@ use Symfony\Component\HttpKernel\EventListener\DisallowRobotsIndexingListener;
 use Symfony\Component\HttpKernel\EventListener\ErrorListener;
 use Symfony\Component\HttpKernel\EventListener\LocaleListener;
 use Symfony\Component\HttpKernel\EventListener\ResponseListener;
+use Symfony\Component\HttpKernel\EventListener\StreamedResponseListener;
 use Symfony\Component\HttpKernel\EventListener\ValidateRequestListener;
 
 return static function (ContainerConfigurator $container) {
@@ -102,6 +103,9 @@ return static function (ContainerConfigurator $container) {
                 param('kernel.charset'),
                 abstract_arg('The "set_content_language_from_locale" config value'),
             ])
+            ->tag('kernel.event_subscriber')
+
+        ->set('streamed_response_listener', StreamedResponseListener::class)
             ->tag('kernel.event_subscriber')
 
         ->set('locale_listener', LocaleListener::class)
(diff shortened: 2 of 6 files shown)


alexander-schranz:
Reverting #45476 will break streamedresponses for other Runtimes like Roadrunner, Swoole (https://github.com/php-runtime/runtime/issues/115), is there any possibility todo this only when specific `Runtime` is used? /cc @nicolas-grekas.

Addition: The problem also is that it would behave differently between Roadrunner, Swoole, now as they could then still run into #46743 where this listener is required to be disabled. I understand @derrabus concern about #51139 but to have full support for such runtimes I don't see another way then make HttpKernel aware of a Response class specially it already has dependency to HttpFoundation.

nicolas-grekas:
Thanks for the reminder @alexander-schranz

derrabus:
> I understand @derrabus concern about #51139 but to have full support for such runtimes I don't see another way then make HttpKernel aware of a Response class specially it already has dependency to HttpFoundation.

The other way would be to document that the request stack is empty when `StreamedResponse`'s callback is executed. This is fine imho.

Note that the breaking change was to un-wire the deprecated listener which has happened without prior deprecation in a minor release. We could wire it again, but make it opt-out by introducing a config flag into FrameworkBundle. Yeah, I know how we don't like those.

nicolas-grekas:
> document that the request stack is empty when StreamedResponse's callback is executed

This ship has sailed IMHO. Also it's very hard to deal with, because any services can be used in that callback. I think #51396 is pragmatic and makes sense also.

derrabus:
> I think #51396 is pragmatic and makes sense also.

LGTM.

PR #51392, opened 2023-08-15
[DependencyInjection] add `#[AutowireLocator]` attribute

kbond:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | todo

The `AutowireLocator` attribute allows configuring service locators inline:

```php
use Symfony\Component\DependencyInjection\Attribute\AutowireLocator;

public function someAction(
    #[AutowireLocator(RouterInterface::class, SluggerInterface::class)]
    ContainerInterface $container,
): Response {
    $container->get(RouterInterface::class);
    $container->get(SluggerInterface::class);
}
```

You can customize the key and have optional services:

```php
use Symfony\Component\DependencyInjection\Attribute\AutowireLocator;

public function someAction(
    #[AutowireLocator(
        router: RouterInterface::class,
        slugger: '?'.SluggerInterface::class,
    )]
    ContainerInterface $container,
): Response {
    $container->get('router');

    if ($container->has('slugger')) {
        $container->get('slugger');
    }
}
```


src/Symfony/Component/DependencyInjection/Attribute/Locator.php
```diff
@@ -0,0 +1,41 @@
...
+            $nullable = str_starts_with($serviceId, '?');
+            $serviceId = ltrim($serviceId, '?');
+
+            if (is_numeric($key)) {
+                $key = $serviceId;
+            }
+
+            $values[$key] = new Reference(
+                $serviceId,
+                $nullable ? ContainerInterface::IGNORE_ON_INVALID_REFERENCE : ContainerInterface::EXCEPTION_ON_INVALID_REFERENCE,
```
kbond: `IGNORE_ON_INVALID_REFERENCE` is what we want here, correct?

src/Symfony/Component/DependencyInjection/Attribute/Locator.php
```diff
@@ -0,0 +1,41 @@
...
+#[\Attribute(\Attribute::TARGET_PARAMETER)]
+final class Locator extends Autowire
+{
+    public function __construct(string ...$serviceIds)
+    {
+        $values = [];
+
+        foreach ($serviceIds as $key => $serviceId) {
+            $nullable = str_starts_with($serviceId, '?');
+            $serviceId = ltrim($serviceId, '?');
```
nicolas-grekas:
```suggestion
            if ($nullable = str_starts_with($serviceId, '?')) {
                $serviceId = substr($serviceId, 1);
            }
```

src/Symfony/Component/DependencyInjection/Attribute/Locator.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+namespace Symfony\Component\DependencyInjection\Attribute;
+
+use Symfony\Component\DependencyInjection\Argument\ServiceLocatorArgument;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\Reference;
+
+#[\Attribute(\Attribute::TARGET_PARAMETER)]
+final class Locator extends Autowire
```
nicolas-grekas:
```suggestion
class AutowireLocator extends Autowire
```

src/Symfony/Component/DependencyInjection/Attribute/Locator.php
nicolas-grekas: I guess so, this is tested, isn't it?

nicolas-grekas (commented):
Love it, I didn't expect the implementation to be so simple :)

derrabus:
In order to address the failing Psalm job, I've opened #51395.

src/Symfony/Component/DependencyInjection/Attribute/Locator.php
kbond: Yes, I wanted to confirm we want `$container->get('nullable.service')` to _not_ return `null` (which would be the case if `NULL_ON_INVALID_REFERENCE` is used).

nicolas-grekas (approved):

nicolas-grekas:
I'm wondering how we could make these work:
```php
#[AutowireLocator([
	'cache' => new SubscribedService(type: CacheInterface::class: attributes: [new Target('my.cache')]),
	'blobs' => new TaggedLocator('blob'),
)]
ContainerInterface $container,
```

Any idea?

kbond:
I thought that too! I was going to take a look at this possibility after this was merged.

BTW, your example can't be an array, it needs to be:

```php
#[AutowireLocator(
    cache: new SubscribedService(type: CacheInterface::class: attributes: [new Target('my.cache')]),
    blobs: new TaggedLocator('blob'),
)]
ContainerInterface $container,
```

(just making sure you're ok with this syntax)

nicolas-grekas:
I just noticed the failures on high+low deps, can you please have a look?

kbond:
I believe the test suite will be fixed once merged because of the http-kernel's requirement on this `AutowireLocator`. Or should I add a conditional in the test?

src/Symfony/Component/HttpKernel/composer.json
```diff
@@ -30,7 +30,7 @@
         "symfony/config": "^6.1|^7.0",
         "symfony/console": "^5.4|^6.0|^7.0",
         "symfony/css-selector": "^5.4|^6.0|^7.0",
-        "symfony/dependency-injection": "^6.3.4|^7.0",
+        "symfony/dependency-injection": "^6.4|^7.0",
```
nicolas-grekas, line 33: the conflict line should also be updated :)

nicolas-grekas:
Thank you @kbond.

PR #51403, opened 2023-08-16
[RateLimiter] Use RateLimiterFactory with custom limit

### LINKED ISSUE #51401: [RateLimiter] Use RateLimiterFactory with custom limit
### Description

Currently the RateLimitFactory fetches the configuration and determines the limit based of that. We'd like to be more flexible with that, allowing certain users a higher limit than others. Adding an optional parameter to the `create` function which will default to the limit set in the config. 

### Example

```
// src/Controller/ApiController.php
namespace App\Controller;

use Symfony\Component\RateLimiter\RateLimiterFactory;

class ApiController extends AbstractController
{
    public function registerUser(Request $request, RateLimiterFactory $authenticatedApiLimiter): Response
    {
        $user = $request->getUser();
        $limiter = $authenticatedApiLimiter->create($user->getApiKey(), $user->getRateLimit());

        // ...
    }

    // ...
}
```

I am happy to make this PR

IlonavD:
Add an optional custom limit to the RateLimiterFactory to overwrite configured values.

| Q             | A
| ------------- | ---
| Branch?       | 6.4  
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Tickets       | Fix #51401 
| License       | MIT

The RateLimiterFactory by default uses the limit set in Config. However in certain situations you might want to change this limit per user. This allows for for that option. When no limit is passed to the Factory, the default configured value will be used.

### DIFF
--- a/src/Symfony/Component/RateLimiter/RateLimiterFactory.php
+++ b/src/Symfony/Component/RateLimiter/RateLimiterFactory.php
@@ -41,15 +41,19 @@ final class RateLimiterFactory
         $this->config = $options->resolve($config);
     }
 
-    public function create(string $key = null): LimiterInterface
+    /**
+     * Parsing a custom limit to this function will overwrite the limit in your configuration for this request.
+     */
+    public function create(string $key = null, ?int $limit = null): LimiterInterface
     {
         $id = $this->config['id'].'-'.$key;
         $lock = $this->lockFactory?->createLock($id);
+        $limit = $limit ?? $this->config['limit'];
 
         return match ($this->config['policy']) {
-            'token_bucket' => new TokenBucketLimiter($id, $this->config['limit'], $this->config['rate'], $this->storage, $lock),
-            'fixed_window' => new FixedWindowLimiter($id, $this->config['limit'], $this->config['interval'], $this->storage, $lock),
-            'sliding_window' => new SlidingWindowLimiter($id, $this->config['limit'], $this->config['interval'], $this->storage, $lock),
+            'token_bucket' => new TokenBucketLimiter($id, $limit, $this->config['rate'], $this->storage, $lock),
+            'fixed_window' => new FixedWindowLimiter($id, $limit, $this->config['interval'], $this->storage, $lock),
+            'sliding_window' => new SlidingWindowLimiter($id, $limit, $this->config['interval'], $this->storage, $lock),
             'no_limit' => new NoLimiter(),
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/RateLimiter/RateLimiterFactory.php
```diff
@@ -41,15 +41,19 @@ public function __construct(array $config, StorageInterface $storage, LockFactor
         $this->config = $options->resolve($config);
     }
 
-    public function create(string $key = null): LimiterInterface
+    /**
+     * Parsing a custom limit to this function will overwrite the limit in your configuration for this request.
+     * When using a NoLimiter the limit will be ignored.
+     */
+    public function create(string $key = null, int $limit = null): LimiterInterface
```
mtarld, line 48: What if you also want at some point to configure the rate dynamically? A solution could be to provide the whole config instead (and resolve it), WDYT? 

```suggestion
    public function create(string $key = null, array $config = null): LimiterInterface
```

stof: this would then force to validate the provided config each time this method is called.

To me, if you want to do things dynamically, it would be better to introduce a RateLimiterFactoryInterface allowing to do a custom implementation of the factory than adding a configuration override argument.
Thus, if you are creating a custom caller for the factory, you could already instantiate the limiter yourselves. And if you don't call it yourselves (relying on a usage in Symfony), you would not be able to pass a custom config.

mtarld: Indeed, I that if we add an argument to the factory method to customize it, there is no point to not adding others (even if the limit is the only common parameter).

That's why, in my opinion too, it is way better to indeed introduce a `RateLimiterFactoryInterface` so there is no limit to the customization.

IlonavD: Alright :) I will create the interface instead later this week. Thanks guys!

kbond:
I put some ideas on this feature here: https://github.com/symfony/symfony/issues/51401#issuecomment-2754958638

nicolas-grekas:
Thank you for this proposal, and sorry for the very long delay. In the review here, @stof and @mtarld asked for a `RateLimiterFactoryInterface` rather than a config-override argument on `RateLimiterFactory::create()`. That interface was added in Symfony 7.3, so per-caller limits can now be implemented with a custom factory. #60084 also explores a builder API for one-off limiters. Closing for these reasons.

PR #51408, opened 2023-08-16
[Runtime] Add environment variable APP_RUNTIME_MODE

### LINKED ISSUE #51340: Function to check if the kernel is loaded via the console or a web request.
### Description

I had the problem that I had to check in a service if the function is called via CLI or web.

Depending on that, I had to make different calculations. That's where my idea comes from, that you could put a function `isCLI()` or `isWeb()` into the kernel.

### Example

```php
class MyService {
  public function __constructor(
    private readonly KernelInterface $kernel
  ){}
  
  public function myFunction(): void {
    if(!$this->kernel->isCLI())
    {
      return;
    }
    // Do things in CLI without Request()
  }
}

### PR DESCRIPTION
**Pull request by BaptisteContreras**
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #51340
| License       | MIT
| Doc PR        | TODO

Based on https://github.com/symfony/symfony/issues/51340#issuecomment-1672895225  I added the envvar **APP_RUNTIME_MODE** in the `autoload_runtime.php`

Its purpose is to indicate in which mode the application is running (`cli` or `web`).

I had a hard time figuring out how to test this. I finally decided to only test the priority to set the value ($_ENV > $_SERVER > default guessing).

During my test, I noticed that if I define a value for `APP_RUNTIME_MODE` in `$_ENV`, and another value in `$_SERVER`, the output of `$context['APP_RUNTIME_MODE']` in my application callable will be the value defined in `$_SERVER`, and the value `$_ENV['APP_RUNTIME_MODE']` is untouched (which is normal). This is due to the way DotEnv loads and override envvars.

You can check this behavior in `src/Symfony/Component/Runtime/Tests/phpt/runtime_mode_from_env_with_server_set.php`.

For me it seems OK because I can't see a way of having `$_SERVER['APP_RUNTIME_MODE']` different from `$_ENV['APP_RUNTIME_MODE']`

I did not do the Doc PR, I first want to check if my implementation if OK for you

### DIFF
--- a/src/Symfony/Component/Runtime/Internal/autoload_runtime.template
+++ b/src/Symfony/Component/Runtime/Internal/autoload_runtime.template
@@ -12,6 +12,16 @@ if (!is_object($app)) {
     throw new TypeError(sprintf('Invalid return value: callable object expected, "%s" returned from "%s".', get_debug_type($app), $_SERVER['SCRIPT_FILENAME']));
 }
 
+if (null === ($_ENV['APP_RUNTIME_MODE'] ??= $_SERVER['APP_RUNTIME_MODE'] ?? null)) {
+    if ($_ENV['FRANKENPHP_WORKER'] ?? $_SERVER['FRANKENPHP_WORKER'] ?? false) {
+        $_ENV['APP_RUNTIME_MODE'] = 'worker';
+    } elseif (\PHP_SAPI === 'cli' || \PHP_SAPI === 'phpdbg') {
+        $_ENV['APP_RUNTIME_MODE'] = 'cli';
+    } else {
+        $_ENV['APP_RUNTIME_MODE'] = 'web';
+    }
+}
+
 $runtime = $_SERVER['APP_RUNTIME'] ?? $_ENV['APP_RUNTIME'] ?? %runtime_class%;
 $runtime = new $runtime(($_SERVER['APP_RUNTIME_OPTIONS'] ?? $_ENV['APP_RUNTIME_OPTIONS'] ?? []) + %runtime_options%);
 
--- a/src/Symfony/Component/Runtime/CHANGELOG.md
+++ b/src/Symfony/Component/Runtime/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+* Add the environnement variable `APP_RUNTIME_MODE`
+
 5.4
 ---
 
--- a/src/Symfony/Component/Runtime/Tests/phpt/runtime_mode_from_default.php
+++ b/src/Symfony/Component/Runtime/Tests/phpt/runtime_mode_from_default.php
@@ -1,0 +1,16 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+require __DIR__.'/autoload.php';
+
+return function (array $context): void {
+    echo 'From context ', $context['APP_RUNTIME_MODE'], ', from $_ENV ', $_ENV['APP_RUNTIME_MODE'];
+};
--- a/src/Symfony/Component/Runtime/Tests/phpt/runtime_mode_from_default.php.phpt
+++ b/src/Symfony/Component/Runtime/Tests/phpt/runtime_mode_from_default.php.phpt
@@ -1,0 +1,12 @@
+--TEST--
+Test set ENV variable APP_RUNTIME_MODE by guessing it from PHP_SAPI
+--INI--
+display_errors=1
+--FILE--
+<?php
+
+require $_SERVER['SCRIPT_FILENAME'] = __DIR__.'/runtime_mode_from_default.php';
+
+?>
+--EXPECTF--
+From context cli, from $_ENV cli
--- a/src/Symfony/Component/Runtime/Tests/phpt/runtime_mode_from_env.php
+++ b/src/Symfony/Component/Runtime/Tests/phpt/runtime_mode_from_env.php
@@ -1,0 +1,18 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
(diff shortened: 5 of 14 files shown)

### CODE REVIEW

**Comment by ro0NL** [comment]
For consideration `APP_RUNTIME_TYPE`?

**Comment by BaptisteContreras** [comment]
`APP_RUNTIME_TYPE` could also be a good fit. I'd rather keep `APP_RUNTIME_MODE ` but I don't mind to change it if you, or others, want :)

File: src/Symfony/Component/Runtime/Internal/autoload_runtime.template [hunk 1]
```diff
@@ -12,6 +12,8 @@ if (!is_object($app)) {
     throw new TypeError(sprintf('Invalid return value: callable object expected, "%s" returned from "%s".', get_debug_type($app), $_SERVER['SCRIPT_FILENAME']));
 }
 
+$_ENV['APP_RUNTIME_MODE'] = $_ENV['APP_RUNTIME_MODE'] ?? $_SERVER['APP_RUNTIME_MODE'] ?? (in_array(\PHP_SAPI, ['cli', 'phpdbg'], true) ? 'cli' : 'web');
```
dunglas:
```suggestion
$_ENV['APP_RUNTIME_MODE'] ??= $_SERVER['APP_RUNTIME_MODE'] ?? (in_array(\PHP_SAPI, ['cli', 'phpdbg'], true) ? 'cli' : 'web');
```

dunglas (approved):
This would be nice to have in 6.4.

dunglas:
For #51661, we can add worker mode detection like that:

```php
if (null === ($_ENV['APP_RUNTIME_MODE'] ??= $_ENV['APP_RUNTIME_MODE'] ?? $_SERVER['APP_RUNTIME_MODE'] ?? null)) {
    if ($_ENV['FRANKENPHP_WORKER'] ?? $_SERVER['FRANKENPHP_WORKER']) {
        $_ENV['APP_RUNTIME_MODE'] = 'worker';
    } elseif (\PHP_SAPI === 'cli' || \PHP_SAPI === 'phpdbg') {
        $_ENV['APP_RUNTIME_MODE'] = 'cli';
    } else {
        $_ENV['APP_RUNTIME_MODE'] = 'web';
    }
}
```

Support for other workers engine like RoadRunner will have to be done in the runtime themselves, because AFAIK they don't provide a way to detect if they are used at this point.

BaptisteContreras:
Hi @dunglas ! Thanks for the review. I updated the PR with your suggestions.

I slightly modified :  

`if ($_ENV['FRANKENPHP_WORKER'] ?? $_SERVER['FRANKENPHP_WORKER']) {`

to

`if ($_ENV['FRANKENPHP_WORKER'] ?? $_SERVER['FRANKENPHP_WORKER'] ?? false) {`

For the case when `$_SERVER['FRANKENPHP_WORKER']` does not exists.

src/Symfony/Component/Runtime/Internal/autoload_runtime.template
```diff
@@ -12,6 +12,16 @@ if (!is_object($app)) {
     throw new TypeError(sprintf('Invalid return value: callable object expected, "%s" returned from "%s".', get_debug_type($app), $_SERVER['SCRIPT_FILENAME']));
 }
 
+if (null === ($_ENV['APP_RUNTIME_MODE'] ??= $_SERVER['APP_RUNTIME_MODE'] ?? null)) {
+    if ($_ENV['FRANKENPHP_WORKER'] ?? $_SERVER['FRANKENPHP_WORKER'] ?? false) {
```
GromNaN, line 16: Do you really need to have this specific check? Wouldn't that be something to put in FrankenPHP instead?

chalasr: Or in https://github.com/php-runtime/frankenphp-symfony?

dunglas: We could do that also for CLI and SAPI. The main benefit of doing it here is to not have to duplicate this code in every runtime supporting FrankenPHP worker mode.

dunglas (approved):

src/Symfony/Component/Runtime/Internal/autoload_runtime.template
GromNaN: Setting `APP_RUNTIME_MODE` in the bridge would be a better demonstration of how this have to be configured. Otherwise we'll receive PR to add Bref or React support in this file.

nicolas-grekas:
While reviewing this PR, I thought another approach might be preferable, see #52079.

BaptisteContreras:
> While reviewing this PR, I thought another approach might be preferable, see #52079.

Your implementation seems indeed more flexible !

nicolas-grekas:
Thanks for pushing this forward @BaptisteContreras!

rustatian:
> For #51661, we can add worker mode detection like that:
> 
> ```
> if (null === ($_ENV['APP_RUNTIME_MODE'] ??= $_ENV['APP_RUNTIME_MODE'] ?? $_SERVER['APP_RUNTIME_MODE'] ?? null)) {
>     if ($_ENV['FRANKENPHP_WORKER'] ?? $_SERVER['FRANKENPHP_WORKER']) {
>         $_ENV['APP_RUNTIME_MODE'] = 'worker';
>     } elseif (\PHP_SAPI === 'cli' || \PHP_SAPI === 'phpdbg') {
>         $_ENV['APP_RUNTIME_MODE'] = 'cli';
>     } else {
>         $_ENV['APP_RUNTIME_MODE'] = 'web';
>     }
> }
> ```
> 
> Support for other workers engine like RoadRunner will have to be done in the runtime themselves, because AFAIK they don't provide a way to detect if they are used at this point.

Hey 👋🏻 
RR provides a way to detect (inside a worker) if it is used, per-plugin. For instance, if HTTP plugin is used `->` `RR_MODE` env variable would be set to `HTTP`: https://roadrunner.dev/docs/php-environment/current#default-env-values-in-php-workers

dunglas:
Hi @rustatian, thanks for the hint!

We decided to move this part in the dedicated adapters provided by php-runtime. If you mind son't hesitate to open a PR for RoadRunner! https://github.com/php-runtime/runtime

rustatian:
With pleasure, @dunglas 😄
I'll ask our PHP team to do that, since I'm not a PHP dev 😢

PR #51415, opened 2023-08-17
[Clock] Add `DatePoint`: an immutable DateTime implementation with stricter error handling and return types

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

The native `DateTime/Immutable` classes have some legacy that we can get rid of by creating a child class.

Here comes `Symfony\Component\Clock\DatePoint`, which extends the native `DateTimeImmutable` and provides the following advantages:
- usable in defaults expressions
- shorter (short) class name
- throws exceptions instead of returning false
- and last but not least, uses the global static clock to define "now" (aka `new DatePoint` is compatible with `Clock::set()`)

~This also opens the possibility to add more methods to this `DatePoint` class in the future if we want to.~ <- we don't want that :)

### DIFF
--- a/src/Symfony/Component/Clock/ClockAwareTrait.php
+++ b/src/Symfony/Component/Clock/ClockAwareTrait.php
@@ -29,8 +29,13 @@ trait ClockAwareTrait
         $this->clock = $clock;
     }
 
+    /**
+     * @return DatePoint
+     */
     protected function now(): \DateTimeImmutable
     {
-        return ($this->clock ??= new Clock())->now();
+        $now = ($this->clock ??= new Clock())->now();
+
+        return $now instanceof DatePoint ? $now : DatePoint::createFromInterface($now);
     }
 }
--- a/src/Symfony/Component/Clock/MockClock.php
+++ b/src/Symfony/Component/Clock/MockClock.php
@@ -20,7 +20,7 @@ namespace Symfony\Component\Clock;
  */
 final class MockClock implements ClockInterface
 {
-    private \DateTimeImmutable $now;
+    private DatePoint $now;
 
     /**
      * @throws \DateMalformedStringException When $now is invalid
@@ -38,20 +38,16 @@ final class MockClock implements ClockInterface
             }
         }
 
-        if (\PHP_VERSION_ID >= 80300 && \is_string($now)) {
-            $now = new \DateTimeImmutable($now, $timezone ?? new \DateTimeZone('UTC'));
-        } elseif (\is_string($now)) {
-            try {
-                $now = new \DateTimeImmutable($now, $timezone ?? new \DateTimeZone('UTC'));
-            } catch (\Exception $e) {
-                throw new \DateMalformedStringException($e->getMessage(), $e->getCode(), $e);
-            }
+        if (\is_string($now)) {
+            $now = new DatePoint($now, $timezone ?? new \DateTimeZone('UTC'));
+        } elseif (!$now instanceof DatePoint) {
+            $now = DatePoint::createFromInterface($now);
         }
 
         $this->now = null !== $timezone ? $now->setTimezone($timezone) : $now;
     }
 
-    public function now(): \DateTimeImmutable
+    public function now(): DatePoint
     {
         return clone $this->now;
     }
@@ -62,7 +58,7 @@ final class MockClock implements ClockInterface
         $now = substr_replace(sprintf('@%07.0F', $now), '.', -6, 0);
         $timezone = $this->now->getTimezone();
 
-        $this->now = (new \DateTimeImmutable($now, $timezone))->setTimezone($timezone);
+        $this->now = DatePoint::createFromInterface(new \DateTimeImmutable($now, $timezone))->setTimezone($timezone);
     }
 
     /**
--- a/src/Symfony/Component/Clock/Resources/now.php
+++ b/src/Symfony/Component/Clock/Resources/now.php
@@ -15,27 +15,14 @@ if (!\function_exists(now::class)) {
     /**
      * @throws \DateMalformedStringException When the modifier is invalid
      */
-    function now(string $modifier = null): \DateTimeImmutable
+    function now(string $modifier = 'now'): DatePoint
     {
-        if (null === $modifier || 'now' === $modifier) {
-            return Clock::get()->now();
+        if ('now' !== $modifier) {
+            return new DatePoint($modifier);
         }
 
         $now = Clock::get()->now();
 
-        if (\PHP_VERSION_ID < 80300) {
-            try {
-                $tz = (new \DateTimeImmutable($modifier, $now->getTimezone()))->getTimezone();
-            } catch (\Exception $e) {
-                throw new \DateMalformedStringException($e->getMessage(), $e->getCode(), $e);
-            }
-            $now = $now->setTimezone($tz);
-
-            return @$now->modify($modifier) ?: throw new \DateMalformedStringException(error_get_last()['message'] ?? sprintf('Invalid date modifier "%s".', $modifier));
-        }
-
-        $tz = (new \DateTimeImmutable($modifier, $now->getTimezone()))->getTimezone();
-
-        return $now->setTimezone($tz)->modify($modifier);
+        return $now instanceof DatePoint ? $now : DatePoint::createFromInterface($now);
     }
 }
--- a/.github/expected-missing-return-types.diff
+++ b/.github/expected-missing-return-types.diff
@@ -1512,6 +1512,16 @@ diff --git a/src/Symfony/Component/Cache/Traits/FilesystemCommonTrait.php b/src/
 +    public function __wakeup(): void
      {
          throw new \BadMethodCallException('Cannot unserialize '.__CLASS__);
+diff --git a/src/Symfony/Component/Clock/ClockAwareTrait.php b/src/Symfony/Component/Clock/ClockAwareTrait.php
+--- a/src/Symfony/Component/Clock/ClockAwareTrait.php
++++ b/src/Symfony/Component/Clock/ClockAwareTrait.php
+@@ -33,5 +33,5 @@ trait ClockAwareTrait
+      * @return DatePoint
+      */
+-    protected function now(): \DateTimeImmutable
++    protected function now(): DatePoint
+     {
+         $now = ($this->clock ??= new Clock())->now();
 diff --git a/src/Symfony/Component/Config/ConfigCacheInterface.php b/src/Symfony/Component/Config/ConfigCacheInterface.php
 --- a/src/Symfony/Component/Config/ConfigCacheInterface.php
 +++ b/src/Symfony/Component/Config/ConfigCacheInterface.php
--- a/src/Symfony/Component/Clock/CHANGELOG.md
+++ b/src/Symfony/Component/Clock/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 6.4
 ---
 
+ * Add `DatePoint`: an immutable DateTime implementation with stricter error handling and return types
  * Throw `DateMalformedStringException`/`DateInvalidTimeZoneException` when appropriate
  * Add `$modifier` argument to the `now()` helper
 
--- a/src/Symfony/Component/Clock/Clock.php
+++ b/src/Symfony/Component/Clock/Clock.php
@@ -44,9 +44,13 @@ final class Clock implements ClockInterface
         self::$globalClock = $clock instanceof ClockInterface ? $clock : new self($clock);
     }
 
-    public function now(): \DateTimeImmutable
+    public function now(): DatePoint
     {
         $now = ($this->clock ?? self::get())->now();
+
+        if (!$now instanceof DatePoint) {
+            $now = DatePoint::createFromInterface($now);
+        }
 
         return isset($this->timezone) ? $now->setTimezone($this->timezone) : $now;
     }
--- a/src/Symfony/Component/Clock/DatePoint.php
+++ b/src/Symfony/Component/Clock/DatePoint.php
@@ -1,0 +1,130 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Clock;
+
+/**
+ * An immmutable DateTime with stricter error handling and return types than the native one.
+ *
+ * @author Nicolas Grekas <p@tchwork.com>
+ */
+final class DatePoint extends \DateTimeImmutable
+{
+    /**
+     * @throws \DateMalformedStringException When $datetime is invalid
+     */
+    public function __construct(string $datetime = 'now', \DateTimeZone $timezone = null, parent $reference = null)
+    {
+        $now = $reference ?? Clock::get()->now();
+
+        if ('now' !== $datetime) {
+            if (!$now instanceof static) {
+                $now = static::createFromInterface($now);
+            }
+
+            if (\PHP_VERSION_ID < 80300) {
+                try {
+                    $timezone = (new parent($datetime, $timezone ?? $now->getTimezone()))->getTimezone();
+                } catch (\Exception $e) {
+                    throw new \DateMalformedStringException($e->getMessage(), $e->getCode(), $e);
+                }
+            } else {
+                $timezone = (new parent($datetime, $timezone ?? $now->getTimezone()))->getTimezone();
+            }
+
+            $now = $now->setTimezone($timezone)->modify($datetime);
+        } elseif (null !== $timezone) {
+            $now = $now->setTimezone($timezone);
+        }
+
+        if (\PHP_VERSION_ID < 80200) {
+            $now = (array) $now;
+            $this->date = $now['date'];
+            $this->timezone_type = $now['timezone_type'];
+            $this->timezone = $now['timezone'];
+            $this->__wakeup();
+
+            return;
+        }
+
+        $this->__unserialize((array) $now);
+    }
+
+    /**
+     * @throws \DateMalformedStringException When $format or $datetime are invalid
+     */
+    public static function createFromFormat(string $format, string $datetime, \DateTimeZone $timezone = null): static
+    {
+        return parent::createFromFormat($format, $datetime, $timezone) ?: throw new \DateMalformedStringException(static::getLastErrors()['errors'][0] ?? 'Invalid date string or format.');
+    }
+
+    public static function createFromInterface(\DateTimeInterface $object): static
+    {
+        return parent::createFromInterface($object);
+    }
+
+    public static function createFromMutable(\DateTime $object): static
+    {
+        return parent::createFromMutable($object);
+    }
+
+    public function add(\DateInterval $interval): static
+    {
+        return parent::add($interval);
+    }
+
+    public function sub(\DateInterval $interval): static
+    {
+        return parent::sub($interval);
+    }
+
+    /**
+     * @throws \DateMalformedStringException When $modifier is invalid
+     */
+    public function modify(string $modifier): static
+    {
+        if (\PHP_VERSION_ID < 80300) {
+            return @parent::modify($modifier) ?: throw new \DateMalformedStringException(error_get_last()['message'] ?? sprintf('Invalid modifier: "%s".', $modifier));
+        }
+
+        return parent::modify($modifier);
+    }
+
+    public function setTimestamp(int $value): static
+    {
+        return parent::setTimestamp($value);
+    }
+
+    public function setDate(int $year, int $month, int $day): static
+    {
+        return parent::setDate($year, $month, $day);
+    }
+
+    public function setISODate(int $year, int $week, int $day = 1): static
+    {
+        return parent::setISODate($year, $week, $day);
+    }
+
+    public function setTime(int $hour, int $minute, int $second = 0, int $microsecond = 0): static
+    {
+        return parent::setTime($hour, $minute, $second, $microsecond);
+    }
+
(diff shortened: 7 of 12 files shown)


derrabus:
> The native `DateTime/Immutable` classes have some legacy that we can get rid of by creating a child class.

Nice. 🙂 

> This also opens the possibility to add more methods to this `DateTime` class in the future if we want to.

That's my biggest fear about adding such a class, tbh.

If by "more methods" we mean polyfills for features that will be added in future PHP versions: yes, totally. But please, let's not open this class up for all kinds of utility methods. 🙈 

Oh, and please add tests for that new class. 🙏🏻

GromNaN:
> > This also opens the possibility to add more methods to this `DateTime` class in the future if we want to.
> 
> That's my biggest fear about adding such a class, tbh.
> 
> If by "more methods" we mean polyfills for features that will be added in future PHP versions: yes, totally. But please, let's not open this class up for all kinds of utility methods. 🙈

We don't want to create a new [Carbon](https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Carbon.php) class.

antonkomarev:
Have you looked at the Brick DateTime library? Its API looks very good and has many types for many cases. Much more stricter and more powerful than Carbon.

src/Symfony/Component/Clock/DateTime.php
```diff
@@ -0,0 +1,123 @@
...
+    public function __construct(string $datetime = 'now', \DateTimeZone $timezone = null)
+    {
+        $now = Clock::get()->now();
+
+        if (null !== $timezone) {
+            $now = $now->setTimezone($timezone);
+        }
+
+        if ('now' !== $datetime) {
+            if (!$now instanceof static) {
```
HeahDude: Maybe all these occurrences of `static` can be replaced by `self` (including return types) since this class is final.

nicolas-grekas: I prefer keeping `static`, it's the way to advertise wither APIs.

HeahDude: Whiters should differentiate returning `$this` instead of `self` or static`, which is not yet possible natively.
If the class is final, there are no benefits, and a drawback since the resolution is made at runtime instead of compilation.

nicolas-grekas: It's the job of the engine to optimize `static` into `self` for final classes, not ours ;)
About withers, we do have a way to differentiate: `@return $this`.

derrabus:
> We don't want to create a new [Carbon](https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Carbon.php) class.

My point exactly.

nicolas-grekas:
PR ready, with tests !

> We don't want to create a new Carbon

Same here, I stroke the line in the PR description. Guarding this is why we have a core-team :)

> Have you looked at the Brick DateTime library

I didn't, but we just ruled such approaches out so :shrug:

nicolas-grekas:
Next steps to consider:
- doctrine types (https://github.com/doctrine/DoctrineBundle/issues/1696 would be nice to help with that, help wanted)
- form types

But before, votes pending @symfony/mergers ;)

wouterj:
I think it's not a good idea to name this class similar to one from PHP's stdlib that has different behavior. You'll have to scroll up to the use statements to know if `new DateTime` means the PHP or Symfony one in any project.

The only case where this is desired is our polyfill packages, which provide polyfills for stdlib features. Eventhough all behavior changes in this class are probably the desired behavior for `DateTime` of all PHP internal devs, there is no official agreement/RFC that this is truly the behavior of the next PHP version as far as I know. We should make sure not to pretend to make a polyfill.

kbond:
> I think it's not a good idea to name this class similar to one from PHP's stdlib that has different behavior.

`Timestamp`?

Kocal:
To me, PHP should have only offered a `DateTime` class (no `DateTimeImmutable`) which would be immutable, and I understand that you want to improve things. 

But I don't think this is the job of a framework / component to do that, as the class name is identical but for a totally different and breaking behavior. Sure you explicity import it, but people will need to check twice if this `DateTime` is imported from the Symfony Clock or is the one from PHP core.

nicolas-grekas:
`DateTime` has my preference. It's how it should have been from the beginning. Unless they go with a very very bad BC break, PHP can't fix this. The ship has sailed on the topic :shrug: .

Figuring out the namespace requires a basic IDE and hovering the symbol, no need to scroll up. For ppl that don't use `use` for root-namespaced classes (following Symfony's recommended CS), there's nothing to do.

> there is no official agreement/RFC that this is truly the behavior of the next PHP version

I'm not sure to get what you mean here. The proposed class is not a polyfill. It's a stricter DateTime. We don't need any agreement from php-core to be future proof, thanks to LSP.

Still, if we want another name, I'd propose `Now`.
This came to my mind when I realized that the implementation of the `now($modifier)` helper can be reduced to the following (if you look at the attached patch, the implementation has a few more lines, but that's just for optimizing performance.):

```php
function now(string $modifier = 'now')
{
    return new Now($modifier);
}
```

kbond:
Not a fan of `Now`:

```php
public function getLastUpdated(): Now
```

GromNaN:
`ClockDateTime` would be descriptive. Even if that repeats the `Symfony\Component\Clock` namespace.

ro0NL:
`BetterDateTime`? because why not :sweat_smile:

nicolas-grekas:
DateMark
DatePoint
DateTick
?

wouterj:
To be honest, since typing my previous message I'm a bit on the fence on whether we should want this feature at all. I don't feel comfortable with a framework that "fixes"/"patches" the standard library.
In my eye, Symfony has always provided low and high level APIs *on top of* the standard library. This is strong: you're not writing some custom language based on PHP, you're just writing PHP. I like how we as Symfony developers aren't "writing Symfony" like everyone used to say in the jQuery days, but most people in the community are aware that they are writing PHP.

Of course, this is only one class, hence I'm not 100% against it, but if we accept this, would we also accept e.g. adding new namespaced functions that introduce strict error handling to stdlib functions that currently return false on error? (i.e. https://github.com/azjezz/psl) And if we don't (I hope we don't!), why would we accept this?

---

This is a language issue that should be solved on language level imho. Create a new date time object (you can directly properly namespace it) and make it behave like it should with our current experience. In a major version or two, deprecate the current date time objects and we're done. PHP has managed much bigger API BC breaks like this (e.g. `mysql_*` to MySQLi/PDO).

nicolas-grekas:
The core motivation for adding this class on my side is to be able to create date-time instances that leverage the global static clock. Aka to make `new DateTime` mockable. That's a unique feature that can't be provided by any future PHP version. The rest is a nice bonus to me and I'd agree with you if strictness was the only argument.

One could say that the `now()` helper is enough, but it can't be used as a default argument value. Adding the class would:
```php
function foo(DateTime $date = new DateTime) {...}
```

wouterj:
Thanks for the explanation. While using Cronos, I haven't found the need to `new DateTime` as a default value, but I can see the added value of it for the Clock component. Now, your `Now` name suggestion also makes lots of sense :)

I'm now learning towards reducing this class to exactly this use-case: being the `now()` equivalent for cases where you need a constant expression, without all the "API improvements". Something along the lines of this:

```php
class Now extends \DateTimeImmutable
{
    public function __construct()
    {
        $now = Clock::get()->now();

        parent::__construct($now->format('c'), $now->getTimezone());
    }
}
```

This does not broaden the scope of the Clock component and keeps us far away from inventing our own language :)

nicolas-grekas:
I've renamed the class to `Moment`, so that we can keep the constructor argument and make it possible to use this class to build not only "now". Keeping only `Now` without constructor arguments still allows doing `$now->modify($foo)`. The limitation is thus going to be mostly annoying without much benefits.

I think the extra bits are also desired: polyfilling the exceptions from 8.3 and making return types stricters. We're not inventing our language here. Instead, we're providing a consistent behavior across versions of PHP. Something PHP itself cannot do without a time travel machine or bad BC breaks.

src/Symfony/Component/Clock/ClockAwareTrait.php
```diff
@@ -29,8 +29,13 @@ public function setClock(ClockInterface $clock): void
...
 
+    /**
+     * @return Moment
+     */
     protected function now(): \DateTimeImmutable
     {
-        return ($this->clock ??= new Clock())->now();
+        $now = ($this->clock ??= new Clock())->now();
+
+        return $now instanceof Moment ? $now : Moment::createFromInterface($now);
```
derrabus: Would it make sense to move this ternary logic into `Moment::createFromInterface()`?

src/Symfony/Component/Clock/MockClock.php
```diff
@@ -62,7 +58,7 @@ public function sleep(float|int $seconds): void
         $now = substr_replace(sprintf('@%07.0F', $now), '.', -6, 0);
         $timezone = $this->now->getTimezone();
 
-        $this->now = (new \DateTimeImmutable($now, $timezone))->setTimezone($timezone);
+        $this->now = Moment::createFromInterface(new \DateTimeImmutable($now, $timezone))->setTimezone($timezone);
```
derrabus: What's the reason you're avoiding to call the constructor of `Moment` directly?

src/Symfony/Component/Clock/ClockAwareTrait.php
nicolas-grekas: Not sure, I'd prefer saving the method call as early as possible (the internal implementation should already implement this logic so adding it on our side wouldn't be needed)

src/Symfony/Component/Clock/MockClock.php
nicolas-grekas: Infinite loop: new Moment() gets the current clock, which can be a MockClock.

OskarStark:
I like the `Moment` wording 😍

javiereguiluz:
I'm not a big fan of the `Moment` name. The dictionary description is _"moment = a very short period of time"_, so this can be confusing.

You may know it but: JavaScript has the same problem as PHP with a badly-designed Date object ... so they are introducing `Temporal` as its replacement. See https://tc39.es/proposal-temporal/docs/

stof:
@javiereguiluz as the class still has to extend DateTimeImmutable (to work with the clock interface), I don't think Temporal can be used as inpiration here.

However, if someone is motivated, this Temporal API of EcmaScript could become the inspiration for a PHP RFC...

OskarStark:
It is also a point in time:
CleanShot 2023-09-12 at 15 00 02@2x

stof:
For that matter, the Temporal proposal of Javascript uses `Instant` as the class name for the object representing an exact time (but note that DateTimeImmutable is not really the equivalent of that object)

kbond:
I still prefer `Timestamp` but `Moment` works too.

nicolas-grekas:
~Let's go with `Instant` then? PHP cannot be exactly like JS anyway :)~
For `Timestamp`, I usually use this noun for UNIX timestamps. I guess that's because I'm not a native speaker.

nicolas-grekas:
~PR updated to use the word `Instant`. Votes pending :)~

aaa2000:
Maybe `DatePoint` https://period.thephpleague.com/5.0/datepoint/

src/Symfony/Component/Clock/Instant.php
```diff
@@ -0,0 +1,130 @@
...
+    {
+        return parent::setISODate($year, $week, $day);
+    }
+
+    public function setTime(int $hour, int $minute, int $second = 0, int $microsecond = 0): static
+    {
+        return parent::setTime($hour, $minute, $second, $microsecond);
+    }
+
+    public function setTimeZone(\DateTimeZone $timezone): static
```
ro0NL: `setTimezone` ;)

ro0NL (commented):
im having 2nd thought about having both `now()` and `new Instant` :sweat_smile: 

edit: the now() ship has sailed :)

src/Symfony/Component/Clock/Instant.php
javiereguiluz: but the PHP class is `DateTimeZone` instead of `DateTimezone` ... so, "timezone" is considered as two separate words: "time + zone"

ro0NL: we're in a subclass of https://www.php.net/manual/en/datetimeimmutable.settimezone.php

nicolas-grekas: good catch, fixed

src/Symfony/Component/Clock/Instant.php
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    public function setTimezone(\DateTimeZone $timezone): static
+    {
+        return parent::setTimezone($timezone);
+    }
+
+    public function getTimezone(): \DateTimeZone
+    {
+        return parent::getTimezone() ?: throw new \DateInvalidTimeZoneException('The Instant object has no timezone.');
```
GromNaN: A DateTime without timezone is a source of issues in applications. Would be nice to enforce UTC as default timezone.

nicolas-grekas: How do you create such an object?

remi-vasco:
My 2 cents: Instant indicate clearly the notion of time. DatePoint, DateMark, ... make me imagine that I am working with a date object (which is not existing AFAIK), not a date time.

OskarStark:
I also prefer `DatePoint` or `Timepoint` (`PointInTime`)

bendavies:
regarding the naming of things, [brick/date-time](https://github.com/brick/date-time) contains an `Instant` class, and many others - `brick/date-time` broadly follows [JSR 310](https://jcp.org/aboutJava/communityprocess/pfd/jsr310/JSR-310-guide.html).

`Instant` may be a poor name as in brick/jsr 310 it is defined a `A single instantaneous point on the time line.` - i.e. it is not zoned, unlike `ZonedDateTime`.

src/Symfony/Component/Clock/Instant.php
GromNaN: Good question, I couldn't find how.

How do you get this exception?

craigh:
`ImmutableDateTime extends \DateTimeImmutable`

nicolas-grekas:
Naming things :facepalm: 

`DatePoint` would work for me. `Moment` also. ~I've reverted to `Moment`~ as I think @OskarStark's answer to @javiereguiluz's makes sense.

> Instant may be a poor name as in brick/jsr 310 it is defined a A single instantaneous point on the time line. - i.e. it is not zoned, unlike ZonedDateTime.

same in JS, that's two implementations that use the word for the same thing - not zoned. That's a good argument against `Instant`.

> ImmutableDateTime extends \DateTimeImmutable

Hell no, "immutable" is just a noisy artifact from the past

src/Symfony/Component/Clock/Instant.php
nicolas-grekas: Dunno, it's just part of the API right now :)

smnandre:
I love `Timepoint`  <3 

I like DatePoint too, but TimePoint has the big advantage to separate the "nature"/"essence" of the object from its representation.

If someone works with "planet lifespans" perdiods ... Using TimePoint feels more abstract and universal.

nicolas-grekas:
~I like it too! PR updated, `TimePoint` FTW!~

src/Symfony/Component/Clock/TimePoint.php
```diff
@@ -0,0 +1,130 @@
...
+ */
+
+namespace Symfony\Component\Clock;
+
+/**
+ * An immmutable DateTime with stricter error handling and return types than the native one.
+ *
+ * @author Nicolas Grekas <p@tchwork.com>
+ */
+final class TimePoint extends \DateTimeImmutable
```
Alpheus: Why extend \DateTimeImmutable?

It seems implementing \DateTimeInterface and using this as a factory to return a \DateTimeImmutable should be sufficient to wrap this constructor.

src/Symfony/Component/Clock/TimePoint.php
```diff
@@ -0,0 +1,130 @@
...
+
+    public function sub(\DateInterval $interval): static
+    {
+        return parent::sub($interval);
+    }
+
+    /**
+     * @throws \DateMalformedStringException When $modifier is invalid
+     */
+    public function modify(string $modifier): static
```
Alpheus: Seems like you're trying to force a bugfix retroactively into the core behavior of \DateTimeImmutable::modify. Why does this have to be constraint to just this class, extending its responsibility?

Alpheus (changes requested):

src/Symfony/Component/Clock/TimePoint.php
stof: A PSR clock must return a DateTimeImmutable.

And PHP does not let implement DateTimeInterface in userland. It is a special interface (like Traversable for instance)

nicolas-grekas: Because the new class must pass `DateTimeImmutable` and `DateTimeInterface` type hints.

src/Symfony/Component/Clock/TimePoint.php
nicolas-grekas: This doesn't change the responsibility of the class. It's just providing a consistent behavior across versions of PHP.

Alpheus: I must be missing something, because the PR overall seems to have good intentions.

This looks like an anti-corruption layer that encapsulates bugs and breaking changes that \DateTimeImmutable has and wraps it into something that extends \DateTimeImmutable. I'm not against the indent, just the logistics. We have 20 years of evidence in the codebase that fixing bugs by forcing subclasses into user space is a terrible idea.

Alpheus (approved):

src/Symfony/Component/Clock/TimePoint.php
Alpheus: @stof thank you mentioning the need to honor PSR20—I wasn't aware of that.

src/Symfony/Component/Clock/Resources/now.php
```diff
@@ -15,27 +15,14 @@
     /**
      * @throws \DateMalformedStringException When the modifier is invalid
      */
-    function now(string $modifier = null): \DateTimeImmutable
+    function now(string $modifier = 'now'): TimePoint
```
Alpheus: Now that I understand this is honoring PSR20, why are we narrowing the **return type** from \DateTimeImmutable to TimePoint? Aren't they polymorphically equivalent?

This seems to create an incentive that carries ($x instanceof TimePoint) into a lazy cast for new methods.

Would it be inappropriate to honor the original interface and keep the type hint as **\DateTimeImmutable** (but returning a TimePoint instance).

This would also have an impact for the class-check unit test below.

src/Symfony/Component/Clock/TimePoint.php
stof: Well, that's the main responsibility of this Clock component: implementing the ClockInterface.

OskarStark (approved):

src/Symfony/Component/Clock/Resources/now.php
stof: Well, here we provide an _extra_ guarantee that our own clock always returns this TimePoint and not other implementations.

Alpheus: Is the extra guarantee _necessary_?

kbond (approved):

fancyweb (approved):

src/Symfony/Component/Clock/Resources/now.php
nicolas-grekas: nothing is necessary, but this still moves the needle forward by allowing eg static analyzers to know that some situations can be handled in a simpler way (eg not checking for `false`, catching exceptions, etc.)

Alpheus: That's very pragmatic and it makes perfect sense for a framework to do so.
The only question I have in that case: are we still honoring PSR20 in spirit?

stof: yes of course. Code written against the PSR20 ClockInterface will not benefit from the extra guarantee (because it works against the PSR20 interface that does not offer it). But we satisfy the PSR20 signature.

Alpheus: So a Mock on a TimePoint would expect callers to be using Symfony\Clock or Symfony\MockClock. Makes perfect sense in my book.

That works well for the test.
However, in a project when the "timepoint for now" is being injected as a service or lazy proxy, should developers be instructed to adhere to Psr\Clock and \DateTimeImmutable or Symfony\...\Clock and ...\TimePoint?

That's the only part I'm worried about that seems to rattle the cage on encapsulation and abstraction creep.

Alpheus: To keep it practical: should the language on the Clock docs be adjusted with this implementation detail?

https://github.com/symfony/symfony-docs/blob/6.4/components/clock.rst

nicolas-grekas: We should definitely tell about the new class in the documentation, but what is written is still accurate.

weaverryan:
`TimePoint` is fine for me. I prefer `DatePoint` as "Date" makes a closer association with `DateTime` than the word "Time".

But we can also call it `RyansDate` - that's probably my favorite 😇

nicolas-grekas:
Updated to `DatePoint` 🤘

I like that typing `Date` when using autocompletion might help discover the new class.

OskarStark:
That makes sense 👍🏻😃

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

PR #51325, opened 2023-08-09
[FrameworkBundle] Deprecate not setting some options

Jean-Beru:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes 
| Tickets       | Fix #51097
| License       | MIT
| Doc PR        | 

This PR deprecates not setting 5 options in framework configuration to match the v7.0 recipe:
* `handle_all_throwables` which is `false` by default and is set to `true` by recipe
* `php_errors.log` which use `$debug` value by default and is set to `true` by recipe
* `session.cookie_secure` (if session is enabled) which is `null` by default and is set to `auto` by recipe
* `session.cookie_samesite` (if session is enabled) which is `null` by default and is set to `lax` by recipe
* `session.handler_id` (if session is enabled) which is `session.handler.native_file` by default and is set to `null` by recipe (could be `session.handler.native_file` if `framework.session.save_path` is set)

See #51097.

PR #51428, opened 2023-08-19
[FrameworkBundle] Remove config deprecations

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | -
| License       | MIT
| Doc PR        | Todo

Follow-up of:
- https://github.com/symfony/symfony/pull/51325
- https://github.com/symfony/symfony/pull/51357

### DIFF
--- a/UPGRADE-7.0.md
+++ b/UPGRADE-7.0.md
@@ -202,6 +202,14 @@ FrameworkBundle
  * Remove the integration of Doctrine annotations, use native attributes instead
  * Remove `EnableLoggerDebugModePass`, use argument `$debug` of HttpKernel's `Logger` instead
  * Remove `AddDebugLogProcessorPass::configureLogger()`, use HttpKernel's `DebugLoggerConfigurator` instead
+ * Make the `framework.handle_all_throwables` config option default to `true`
+ * Make the `framework.php_errors.log` config option default to `true`
+ * Make the `framework.session.cookie_secure` config option default to `auto`
+ * Make the `framework.session.cookie_samesite` config option default to `lax`
+ * Make the `framework.session.handler_id` default to null if `save_path` is not set and to `session.handler.native_file` otherwise
+ * Make the `framework.uid.default_uuid_version` config option default to `7`
+ * Make the `framework.uid.time_based_uuid_version` config option default to `7`
+ * Make the `framework.validation.email_validation_mode` config option default to `html5`
 
 HttpFoundation
 --------------
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
@@ -82,15 +82,6 @@ class Configuration implements ConfigurationInterface
                     return $v;
                 })
             ->end()
-            ->validate()
-                ->always(function ($v) {
-                    if (!isset($v['handle_all_throwables'])) {
-                        trigger_deprecation('symfony/framework-bundle', '6.4', 'Not setting the "framework.handle_all_throwables" config option is deprecated. It will default to "true" in 7.0.');
-                    }
-
-                    return $v;
-                })
-            ->end()
             ->fixXmlConfig('enabled_locale')
             ->children()
                 ->scalarNode('secret')->end()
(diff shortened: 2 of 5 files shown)


src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -713,6 +683,15 @@ private function addSessionSection(ArrayNodeDefinition $rootNode): void
                     ->end()
                 ->end()
             ->end()
+            ->validate()
+                ->always(function (array $v): array {
+                    if (!\array_key_exists('handler_id', $v['session'])) {
+                        $v['session']['handler_id'] = isset($v['session']['save_path']) ? 'session.handler.native_file' : null;
```
nicolas-grekas: it looks like `save_path` has a default value, so that by default, we'll always have `handler_id` be not null.
we might need to fix the deprecation on 6.4 so that we require save_path to be explicitly set when handler_id is not, so that we can make save_path default to null in 7.0. Can you please have a look?

alexandre-daubois: Created the PR for 6.4 here: https://github.com/symfony/symfony/pull/51438

nicolas-grekas: Thanks! Time to rebase this PR :)

UPGRADE-7.0.md
```diff
@@ -202,6 +202,14 @@ FrameworkBundle
  * Remove the integration of Doctrine annotations, use native attributes instead
  * Remove `EnableLoggerDebugModePass`, use argument `$debug` of HttpKernel's `Logger` instead
  * Remove `AddDebugLogProcessorPass::configureLogger()`, use HttpKernel's `DebugLoggerConfigurator` instead
+ * Set `framework.handle_all_throwables` config option default value to `true`
```
GromNaN: This should be upgrade instructions. Let's recommend unsetting this values from the project when the value is the new default.

alexandre-daubois: Does something like `"framework.handle_all_throwables" config option default value is now "true" and should be unset if set manually to this value` would suit?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
alexandre-daubois: Done :slightly_smiling_face:

UPGRADE-7.0.md
nicolas-grekas: The current message is roughly how we documented this previously
To be more consistent we could use
`Make the foo config option default to bar`

alexandre-daubois: Updated with Nicolas' format :+1:

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -649,35 +640,6 @@ private function addRouterSection(ArrayNodeDefinition $rootNode): void
...
-                            } else {
-                                trigger_deprecation('symfony/framework-bundle', '6.4', 'Not setting the "framework.session.handler_id" config option is deprecated. It will default to "session.handler.native_file" when "framework.session.save_path" is set or "null" otherwise in 7.0.');
-                            }
-                        }
-                    }
-
-                    $v['session'] += [
-                        'cookie_samesite' => null,
-                        'handler_id' => 'session.handler.native_file',
-                        'save_path' => '%kernel.cache_dir%/sessions',
```
nicolas-grekas, line 671: when `handler_id` is non-null, save_path should still default to `%kernel.cache_dir%/sessions`, how can we deal with that?

alexandre-daubois: I updated the condition:

```php
if (!\array_key_exists('handler_id', $v['session'])) {
    $v['session']['handler_id'] = $v['session']['save_path'] ? 'session.handler.native_file' : null;
} elseif (null !== $v['session']['handler_id']) {
    $v['session']['save_path'] ??= '%kernel.cache_dir%/sessions';
}
```

Also updated changelog to `Make the framework.session.save_path config option default to %kernel.cache_dir%/sessions if framework.session.handler_id is set or null otherwise`

alexandre-daubois: I also added a call to `info()` to document this behavior

alexandre-daubois:
I'd say [this PR](https://github.com/symfony/symfony/pull/51455) should be merged before this one to fix the errors raised by `AttributeLoader`.

I fixed the `Configuration.php` file, thank you for the edits on it about default values

nicolas-grekas:
Thank you @alexandre-daubois.

PR #51431, opened 2023-08-19
[HttpFoundation] Add `ParameterBag::match` method

### LINKED ISSUE #51224: [HttpFoundation] Introducing a ParameterBag::match method
### Description

Hi 👋🏻 

This issue is created following a use case I'm facing where I need to test that a route parameter match a specific regex, I currently use something like this: 

```php
$parameter = $request->attributes->get(key: 'foo');

if (!preg_match(REGEX, $parameter,)) {
    return;
}
```

I'm wondering if `HttpFoundation` couldn't handle this check using a method like `match`: 

```php
public function match(string $key, string $regexp): bool
{
    return preg_match($regexp, $this->get($key)) > 0;
}
```

This way, we could refactor the code to: 

```php
if (!$request->atttributes->match(key: 'foo', regexp: 'REGEX') {
    return;
}
```

This is just sugar for developers but IMHO it can improve code comprehension and reduce the cases where a variable is created just to check its value, any thoughts?

Thanks 🙂 

### Example

_No response_

henrikac:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #51224 
| License       | MIT
| Doc PR        | symfony/symfony-docs#18768

This PR tries to implement a feature requested in the issue mentioned in the table above.

`ParameterBag::match` allows developers to check whether a parameter matches a regular expression like this

```php
if ($request->atttributes->match(key: 'foo', regexp: '/REGEX/') {
    // do something
}
```

### DIFF
--- a/src/Symfony/Component/HttpFoundation/ParameterBag.php
+++ b/src/Symfony/Component/HttpFoundation/ParameterBag.php
@@ -124,14 +124,10 @@ class ParameterBag implements \IteratorAggregate, \Countable
             return false;
         }
 
-        if (\strlen($regexp) < 2 || ('/' !== $regexp[0] || '/' !== $regexp[\strlen($regexp) - 1])) {
-            throw new \InvalidArgumentException('Parameter "regexp" is not a valid regular expression.');
-        }
-
-        $match = preg_match($regexp, $this->getString($key));
+        $match = @preg_match($regexp, $this->getString($key));
 
         if (false === $match) {
-            throw new \InvalidArgumentException('Parameter "regexp" is not a valid regular expression.');
+            throw new \InvalidArgumentException(sprintf('Parameter "regexp" is not a valid regular expression: "%s".', preg_last_error_msg()));
         }
 
         return 1 === $match;
--- a/src/Symfony/Component/HttpFoundation/ParameterBag.php
+++ b/src/Symfony/Component/HttpFoundation/ParameterBag.php
@@ -113,12 +113,10 @@ class ParameterBag implements \IteratorAggregate, \Countable
     /**
      * Checks whether the given parameter matches the given regular expression.
      *
-     * @param string $key
-     * @param string $regexp
-     * @return boolean Returns true if the paramter matches the regular expression.
-     *                 However, false is returned if the parameter is either not defined or if the parameter did not match the regular expression.
-     * 
-     * @throws \InvalidArgumentException InvalidArgumentException is thrown if regexp is not a valid regular expression.
+     * @return bool Returns true if the paramter matches the regular expression.
+     *              However, false is returned if the parameter is either not defined or if the parameter did not match the regular expression.
+     *
+     * @throws \InvalidArgumentException if regexp is not a valid regular expression
      */
     public function match(string $key, string $regexp): bool
     {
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/HttpFoundation/ParameterBag.php
```diff
@@ -110,6 +110,33 @@ public function remove(string $key)
...
+        }
+
+        if (\strlen($regexp) < 2 || ('/' !== $regexp[0] || '/' !== $regexp[\strlen($regexp) - 1])) {
+            throw new \InvalidArgumentException('Parameter "regexp" is not a valid regular expression.');
+        }
+
+        $match = preg_match($regexp, $this->getString($key));
+
+        if (false === $match) {
+            throw new \InvalidArgumentException('Parameter "regexp" is not a valid regular expression.');
```
Kocal: It could be nice to show the message returned by https://www.php.net/manual/fr/function.preg-last-error-msg.php

henrikac: Do you mean instead of `'Parameter "regexp" is not a valid regular expression.'` or a combination of both messages?

src/Symfony/Component/HttpFoundation/ParameterBag.php
```diff
@@ -110,6 +110,33 @@ public function remove(string $key)
...
+     */
+    public function match(string $key, string $regexp): bool
+    {
+        if (!$this->has($key)) {
+            return false;
+        }
+
+        if (\strlen($regexp) < 2 || ('/' !== $regexp[0] || '/' !== $regexp[\strlen($regexp) - 1])) {
+            throw new \InvalidArgumentException('Parameter "regexp" is not a valid regular expression.');
+        }
```
derrabus: This check feels a bit arbitrary. Why can't I use `#` or `~` as delimiter?

The `preg_match()` function validates the expression already. I think we can simply drop this block.

henrikac: I agree and I have removed this check

nicolas-grekas:
Thanks for giving this a try. Someone needs to ask, so let me do it :)

I see drawbacks to adding this method: increases API surface, thus maintenance burden, and something new to learn/teach and document.
Does this deserve adding the method? (it's an open question on my side)

```php
// before
preg_match(REGEX, $request->attributes->get(key: 'foo'));

// after
$request->attributes->match(key: 'foo', REGEX);
```

henrikac:
Great question :)

My main reason implementing this method and trying to fix this issue was to learn about how to contribute to this amazing framework - which was more complicated than first assumed. I picked an easy issue that was at least two weeks old that had only positive reactions.

This reason is of course not enough to add this "sugar method" to the framework but reading the issue and the reactions (4 so far) it looks like this method could bring value to some developers. The author of the issue states that

> This is just sugar for developers but IMHO it can improve code comprehension

but as you show in your code snippet the difference is minimal and I understand if you decide that this method does not bring enough value to be added :)

ro0NL:
better value could be provided with `getMatchingString('foo', REGEX): string`, considering the error handling in InputBag (e

94noni:
Perhaps leverage the match method from the string component? 
https://symfony.com/doc/current/components/string.html#methods-to-search-and-replace

OskarStark:
So let's close for the given reasons. Thank you for providing a PR 👍

PR #51434, opened 2023-08-20
[Security] [Throttling] Hide username and client ip in logs

### LINKED ISSUE #46362: [Security][Throttling] Hide username and client ip in logs
### Description

When using the login throttling feature, username and IP may appear in the debug logs (e.g. when an error occurs in prod with fingers_crossed logger):

```
Successfully acquired the "username_ip_login-random.username-1.2.3.4" lock.
Expiration defined for "username_ip_login-random.username-1.2.3.4" lock for "300" seconds.
```

What do you think about (optionally) masking these information? For example by adding a `MaskingLoginRateLimiter` in addition to the `DefaultLoginRateLimiter`. A straight forward approach would be to use a hash of username and client ip as the key for the limiter(s).

This would not only mask the log messages but would impact all appearances of the keys, e.g. in storage. This might be an advantage or a disadvantage. Although, at least the PdoStore hashes the key anyway. The Lock class, where the logs originate, seems to be the wrong place for tweaking, as it's (and should not be) aware of the content of the key.

Another approach, probably on application level, would be a monolog processor, which replaces the username and client ip, but this seems less efficient and not that robust.

### Example

```php
protected function getLimiters(Request $request): array
{
    $username = $request->attributes->get(Security::LAST_USERNAME, '');

    $globalKey = hash('sha256', $request->getClientIp());
    $localKey = hash('sha256', $username.'-'.$request->getClientIp());

    return [
        $this->globalFactory->create($globalKey),
        $this->localFactory->create($localKey),
    ];
}
```

Spomky:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | yes
| Deprecations? | no 
| Tickets       | Fix #46362
| License       | MIT
| Doc PR        | symfony/symfony-docs#... **TODO**

This PR is a proposal for fixing #46362. It appears the username and IP address may be both available in the log or the caching system.
The proposed feature uses the already existing kernel secret to hash the data.

### DIFF
--- a/src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
+++ b/src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
@@ -28,11 +28,20 @@ final class DefaultLoginRateLimiter extends AbstractRequestRateLimiter
 {
     private RateLimiterFactory $globalFactory;
     private RateLimiterFactory $localFactory;
+    private string $secret;
 
-    public function __construct(RateLimiterFactory $globalFactory, RateLimiterFactory $localFactory)
+    /**
+     * @param non-empty-string $secret A secret to use for hashing the IP address and username
+     */
+    public function __construct(RateLimiterFactory $globalFactory, RateLimiterFactory $localFactory, #[\SensitiveParameter] string $secret = '')
     {
+        if ('' === $secret) {
+            trigger_deprecation('symfony/security-http', '6.4', 'Calling "%s()" with an empty secret is deprecated. A non-empty secret will be mandatory in version 7.0.', __METHOD__);
+            // throw new \Symfony\Component\Security\Core\Exception\InvalidArgumentException('A non-empty secret is required.');
+        }
         $this->globalFactory = $globalFactory;
         $this->localFactory = $localFactory;
+        $this->secret = $secret;
     }
 
     protected function getLimiters(Request $request): array
@@ -41,8 +50,13 @@ final class DefaultLoginRateLimiter extends AbstractRequestRateLimiter
         $username = preg_match('//u', $username) ? mb_strtolower($username, 'UTF-8') : strtolower($username);
 
         return [
-            $this->globalFactory->create($request->getClientIp()),
-            $this->localFactory->create($username.'-'.$request->getClientIp()),
+            $this->globalFactory->create($this->hash($request->getClientIp())),
+            $this->localFactory->create($this->hash($username.'-'.$request->getClientIp())),
         ];
+    }
+
+    private function hash(string $data): string
+    {
+        return strtr(substr(base64_encode(hash_hmac('sha256', $data, $this->secret, true)), 0, 8), '/+', '._');
     }
 }
--- a/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/LoginThrottlingFactory.php
+++ b/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/LoginThrottlingFactory.php
@@ -76,6 +76,7 @@ class LoginThrottlingFactory implements AuthenticatorFactoryInterface
             $container->register($config['limiter'] = 'security.login_throttling.'.$firewallName.'.limiter', DefaultLoginRateLimiter::class)
                 ->addArgument(new Reference('limiter.'.$globalId))
                 ->addArgument(new Reference('limiter.'.$localId))
+                ->addArgument('%kernel.secret%')
             ;
         }
 
--- a/src/Symfony/Component/Security/Http/Tests/EventListener/LoginThrottlingListenerTest.php
+++ b/src/Symfony/Component/Security/Http/Tests/EventListener/LoginThrottlingListenerTest.php
@@ -47,7 +47,7 @@ class LoginThrottlingListenerTest extends TestCase
             'limit' => 6,
             'interval' => '1 minute',
         ], new InMemoryStorage());
-        $limiter = new DefaultLoginRateLimiter($globalLimiter, $localLimiter);
+        $limiter = new DefaultLoginRateLimiter($globalLimiter, $localLimiter, '$3cre7');
 
         $this->listener = new LoginThrottlingListener($this->requestStack, $limiter);
     }


src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
```diff
@@ -28,21 +28,25 @@ final class DefaultLoginRateLimiter extends AbstractRequestRateLimiter
...
     private RateLimiterFactory $globalFactory;
     private RateLimiterFactory $localFactory;
+    private ?string $secret;
 
-    public function __construct(RateLimiterFactory $globalFactory, RateLimiterFactory $localFactory)
+    public function __construct(RateLimiterFactory $globalFactory, RateLimiterFactory $localFactory, #[\SensitiveParameter] string $secret = null)
     {
         $this->globalFactory = $globalFactory;
         $this->localFactory = $localFactory;
+        $this->secret = $secret;
```
OskarStark, line 44: What about deprecate not providing a secret?

OskarStark (approved):

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
Spomky: Indeed. The constructor now triggers a deprecation if needed.

OskarStark: If not already a thing, we need to require deprecation-contracts package 📦

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
```diff
@@ -28,21 +28,28 @@ final class DefaultLoginRateLimiter extends AbstractRequestRateLimiter
...
     }
 
     protected function getLimiters(Request $request): array
     {
         $username = $request->attributes->get(SecurityRequestAttributes::LAST_USERNAME, '');
         $username = preg_match('//u', $username) ? mb_strtolower($username, 'UTF-8') : strtolower($username);
+        $ip = $this->secret ? hash_hmac('sha256', $request->getClientIp(), $this->secret) : $request->getClientIp();
+        $usernameAndIp = $this->secret ? hash_hmac('sha256', $username.'-'.$request->getClientIp(), $this->secret) : $username.'-'.$request->getClientIp();
```
nicolas-grekas: let's add a private method to split the hashing concern?
let's also base64-encode and truncate the hash:

```
$hash = strtr(substr(base64_encode(hash_hmac('sha256', $value, $this->secret, true)), 0, 8), '/+', '._');
```

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
```diff
@@ -28,21 +28,28 @@ final class DefaultLoginRateLimiter extends AbstractRequestRateLimiter
 {
     private RateLimiterFactory $globalFactory;
     private RateLimiterFactory $localFactory;
+    private ?string $secret;
 
-    public function __construct(RateLimiterFactory $globalFactory, RateLimiterFactory $localFactory)
+    public function __construct(RateLimiterFactory $globalFactory, RateLimiterFactory $localFactory, #[\SensitiveParameter] string $secret = null)
     {
+        if (!$secret) {
+            trigger_deprecation('symfony/security-http', '6.4', 'Calling "%s()" with an empty secret is deprecated, pass a non-empty string instead.', __METHOD__);
```
nicolas-grekas: please add the line that we should have in 7.0, commented, so that it's easy to remove the deprecation on 7.0

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
Spomky: Good point. It looks like [it is already included](https://github.com/symfony/symfony/blob/c36a7668e1445a6fcb369e6294bb6f380523c2db/src/Symfony/Component/Security/Http/composer.json#L20).

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
```diff
@@ -28,21 +28,36 @@ final class DefaultLoginRateLimiter extends AbstractRequestRateLimiter
 {
     private RateLimiterFactory $globalFactory;
     private RateLimiterFactory $localFactory;
+    private ?string $secret;
 
-    public function __construct(RateLimiterFactory $globalFactory, RateLimiterFactory $localFactory)
+    /**
+     * @param non-empty-string|null $secret A secret to use for hashing the IP address and username
```
nicolas-grekas: no need to document the deprecated way:
```suggestion
     * @param non-empty-string $secret A secret to use for hashing the IP address and username
```

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
```diff
@@ -28,21 +28,36 @@ final class DefaultLoginRateLimiter extends AbstractRequestRateLimiter
...
+     */
+    public function __construct(RateLimiterFactory $globalFactory, RateLimiterFactory $localFactory, #[\SensitiveParameter] string $secret = null)
```
nicolas-grekas:
```suggestion
    public function __construct(RateLimiterFactory $globalFactory, RateLimiterFactory $localFactory, #[\SensitiveParameter] string $secret = '')
```

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
Spomky: I added a private method as suggested.
Base64 truncated hash is a good idea.
One question though: why do you replace with `'._'` and not `'-_'` like base64 url-safe? In any case, it is url-safe as well

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
Spomky: OK. Comment added

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
```diff
@@ -28,21 +28,36 @@ final class DefaultLoginRateLimiter extends AbstractRequestRateLimiter
...
         $username = $request->attributes->get(SecurityRequestAttributes::LAST_USERNAME, '');
         $username = preg_match('//u', $username) ? mb_strtolower($username, 'UTF-8') : strtolower($username);
+        $ip = $this->secret ? $this->hash($request->getClientIp()) : $request->getClientIp();
+        $usernameAndIp = $this->secret ? $this->hash($username.'-'.$request->getClientIp()) : $username.'-'.$request->getClientIp();
 
         return [
-            $this->globalFactory->create($request->getClientIp()),
-            $this->localFactory->create($username.'-'.$request->getClientIp()),
+            $this->globalFactory->create($ip),
+            $this->localFactory->create($usernameAndIp),
```
nicolas-grekas:
```suggestion
            $this->globalFactory->create($this->hash($ip)),
            $this->localFactory->create($this->hash($usernameAndIp)),
```

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
```diff
@@ -28,21 +28,36 @@ final class DefaultLoginRateLimiter extends AbstractRequestRateLimiter
...
         $this->localFactory = $localFactory;
+        $this->secret = $secret;
     }
 
     protected function getLimiters(Request $request): array
     {
         $username = $request->attributes->get(SecurityRequestAttributes::LAST_USERNAME, '');
         $username = preg_match('//u', $username) ? mb_strtolower($username, 'UTF-8') : strtolower($username);
+        $ip = $this->secret ? $this->hash($request->getClientIp()) : $request->getClientIp();
+        $usernameAndIp = $this->secret ? $this->hash($username.'-'.$request->getClientIp()) : $username.'-'.$request->getClientIp();
```
nicolas-grekas:
```suggestion
```

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
nicolas-grekas: no specific reasons, I borrowed this from `ContainerBuilder::hash()` but I'd be fine with `-` also

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
```diff
@@ -28,11 +28,19 @@ final class DefaultLoginRateLimiter extends AbstractRequestRateLimiter
 {
     private RateLimiterFactory $globalFactory;
     private RateLimiterFactory $localFactory;
+    private ?string $secret;
```
nicolas-grekas: sorry I missed this one:
```suggestion
    private string $secret;
```

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
```diff
@@ -28,11 +28,19 @@ final class DefaultLoginRateLimiter extends AbstractRequestRateLimiter
...
+    private string $secret;
 
-    public function __construct(RateLimiterFactory $globalFactory, RateLimiterFactory $localFactory)
+    /**
+     * @param non-empty-string $secret A secret to use for hashing the IP address and username
+     */
+    public function __construct(RateLimiterFactory $globalFactory, RateLimiterFactory $localFactory, #[\SensitiveParameter] string $secret = '')
     {
+        if (!$secret) {// @deprecated remove this catch block in 7.0 and make $secret mandatory.
+            trigger_deprecation('symfony/security-http', '6.4', 'Calling "%s()" with an empty secret is deprecated. A non-empty secret will be mandatory in version 7.0.', __METHOD__);
```
nicolas-grekas:
```suggestion
        if (!$secret) {
            trigger_deprecation('symfony/security-http', '6.4', 'Calling "%s()" with an empty secret is deprecated. A non-empty secret will be mandatory in version 7.0.', __METHOD__);
            // throw new \InvalidArgumentException(blabla);
```

Spomky: 😄OK I got it!

nicolas-grekas (approved):

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
```diff
@@ -28,11 +28,20 @@ final class DefaultLoginRateLimiter extends AbstractRequestRateLimiter
...
 
-    public function __construct(RateLimiterFactory $globalFactory, RateLimiterFactory $localFactory)
+    /**
+     * @param non-empty-string $secret A secret to use for hashing the IP address and username
+     */
+    public function __construct(RateLimiterFactory $globalFactory, RateLimiterFactory $localFactory, #[\SensitiveParameter] string $secret = '')
     {
+        if (!$secret) {
+            trigger_deprecation('symfony/security-http', '6.4', 'Calling "%s()" with an empty secret is deprecated. A non-empty secret will be mandatory in version 7.0.', __METHOD__);
+            // throw new \InvalidArgumentException('A non-empty secret is required.');
```
mtarld: This will be for 7.0, but maybe it'll be better to use `Symfony\Component\Security\Core\Exception\InvalidArgumentException` instead?

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
```diff
@@ -28,11 +28,20 @@ final class DefaultLoginRateLimiter extends AbstractRequestRateLimiter
...
     private RateLimiterFactory $localFactory;
+    private string $secret;
 
-    public function __construct(RateLimiterFactory $globalFactory, RateLimiterFactory $localFactory)
+    /**
+     * @param non-empty-string $secret A secret to use for hashing the IP address and username
+     */
+    public function __construct(RateLimiterFactory $globalFactory, RateLimiterFactory $localFactory, #[\SensitiveParameter] string $secret = '')
     {
+        if (!$secret) {
```
mtarld:
```suggestion
        if ('' === $secret) {
```
Unless `'0'` string is forbidden as well

mtarld (changes requested):

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
Spomky: Correct. I changed the line

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
Spomky: Strictly speaking, the secret must be at least 256 bits because the `sha256` function is used.
But here I only wanted to hide the IP address and username-IP. The use of a secret is to avoid the simple `sha256` hashing function and prevent recovery e.g. using a rainbow table. So all non-empty strings are fine I think, including `0`.
From the perspective of `SecurityBundle`, `0` is very unlikely because the secret of the application is injected.

mtarld (approved):

nicolas-grekas (approved):

nicolas-grekas:
Thank you @Spomky.

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
```diff
@@ -41,8 +50,13 @@ protected function getLimiters(Request $request): array
...
-            $this->globalFactory->create($request->getClientIp()),
-            $this->localFactory->create($username.'-'.$request->getClientIp()),
+            $this->globalFactory->create($this->hash($request->getClientIp())),
+            $this->localFactory->create($this->hash($username.'-'.$request->getClientIp())),
         ];
     }
+
+    private function hash(string $data): string
+    {
+        return strtr(substr(base64_encode(hash_hmac('sha256', $data, $this->secret, true)), 0, 8), '/+', '._');
```
Seldaek, line 60: I came here to check as I was worried I'd have to complain about a massive cache size increase, but I'm happy to see that it was taken into account and this will actually probably even reduce the cache storage requirements. Thanks!

Spomky: Honestly, I fell into the trap 😅🤦‍♂️. But there are people here who are very quick-witted and who paid attention to that 😊. #bestcommunityever 👌

Issue #51441, opened 2023-08-21
[Security] FormLoginAuthenticator fails when password is an array


dmaicher:
### Symfony version(s) affected

5.4+

### Description

```
Symfony\\Component\\Security\\Http\\Authenticator\\Passport\\Credentials\\PasswordCredentials::__construct(): Argument #1 ($password) must be of type string, array given, called in /var/www/app/vendor/symfony/security-http/Authenticator/FormLoginAuthenticator.php on line 85
```

### How to reproduce

Post array data for the password to a form login 

```
curl -X POST --data '_username=foo&_password[]=bar' http://app.dev/login_check
``` 

### Possible Solution

I saw that for the username there is a check in place:

https://github.com/symfony/symfony/blob/6126cafdbaf98810766ece18e9ae620e98743235/src/Symfony/Component/Security/Http/Authenticator/FormLoginAuthenticator.php#L126

Maybe we can do the same for the password? Happy to contribute this if it makes sense

### Additional Context

Actually in dev mode there is another issue with the data collector:

```
TypeError:
rawurlencode(): Argument #1 ($string) must be of type string, array given

  at /var/www/app/vendor/symfony/http-kernel/DataCollector/RequestDataCollector.php:127

derrabus:
Please send a PR.

PR #51445, opened 2023-08-21
[Security] FormLoginAuthenticator: fail for non-string password

dmaicher:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix https://github.com/symfony/symfony/issues/51441
| License       | MIT
| Doc PR        | -

Fixes https://github.com/symfony/symfony/issues/51441 by handling it similar to the username and throwing a `BadRequestHttpException`.

### DIFF
--- a/src/Symfony/Component/Security/Http/Authenticator/FormLoginAuthenticator.php
+++ b/src/Symfony/Component/Security/Http/Authenticator/FormLoginAuthenticator.php
@@ -157,6 +157,10 @@ class FormLoginAuthenticator extends AbstractLoginFormAuthenticator
 
         $request->getSession()->set(Security::LAST_USERNAME, $credentials['username']);
 
+        if (!\is_string($credentials['password']) && (!\is_object($credentials['password']) || !method_exists($credentials['password'], '__toString'))) {
+            throw new BadRequestHttpException(sprintf('The key "%s" must be a string, "%s" given.', $this->options['password_parameter'], \gettype($credentials['password'])));
+        }
+
         return $credentials;
     }
 
--- a/src/Symfony/Component/Security/Http/Tests/Authenticator/FormLoginAuthenticatorTest.php
+++ b/src/Symfony/Component/Security/Http/Tests/Authenticator/FormLoginAuthenticatorTest.php
@@ -23,6 +23,7 @@ use Symfony\Component\Security\Http\Authentication\AuthenticationSuccessHandlerI
 use Symfony\Component\Security\Http\Authenticator\FormLoginAuthenticator;
 use Symfony\Component\Security\Http\Authenticator\Passport\Badge\CsrfTokenBadge;
 use Symfony\Component\Security\Http\Authenticator\Passport\Badge\PasswordUpgradeBadge;
+use Symfony\Component\Security\Http\Authenticator\Passport\Credentials\PasswordCredentials;
 use Symfony\Component\Security\Http\HttpUtils;
 use Symfony\Component\Security\Http\Tests\Authenticator\Fixtures\PasswordUpgraderProvider;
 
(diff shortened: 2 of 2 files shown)


src/Symfony/Component/Security/Http/Authenticator/FormLoginAuthenticator.php
```diff
@@ -157,6 +157,10 @@ private function getCredentials(Request $request): array
 
         $request->getSession()->set(Security::LAST_USERNAME, $credentials['username']);
 
+        if (!\is_string($credentials['password'])) {
```
dmaicher: for the username we allow "Stringable" as well. Is this relevant for the password? I guess not, or?

src/Symfony/Component/Security/Http/Tests/Authenticator/FormLoginAuthenticatorTest.php
```diff
@@ -126,6 +126,36 @@ public function testHandleNonStringUsernameWithToString($postOnly)
...
+        $request->setSession($this->createSession());
+
+        $this->setUpAuthenticator(['post_only' => $postOnly]);
+        $this->authenticator->authenticate($request);
+    }
+
+    /**
+     * @dataProvider postOnlyDataProvider
+     */
+    public function testHandleNonStringPasswordWithInt($postOnly)
```
mtarld: I think only one test is needed :slightly_smiling_face:

src/Symfony/Component/Security/Http/Authenticator/FormLoginAuthenticator.php
mtarld: IMHO, it can be relevant to allow "Stringable" passwords, because it can ease the use of value objects.

src/Symfony/Component/Security/Http/Tests/Authenticator/FormLoginAuthenticatorTest.php
```diff
@@ -126,6 +126,36 @@ public function testHandleNonStringUsernameWithToString($postOnly)
         $this->authenticator->authenticate($request);
     }
 
+    /**
+     * @dataProvider postOnlyDataProvider
+     */
+    public function testHandleNonStringPasswordWithArray($postOnly)
```
derrabus:
```suggestion
    public function testHandleNonStringPasswordWithArray(bool $postOnly)
```

src/Symfony/Component/Security/Http/Authenticator/FormLoginAuthenticator.php
dmaicher: But is that a real use-case? I mean do people have value objects for their passwords in the request bag somehow? :thinking:

src/Symfony/Component/Security/Http/Tests/Authenticator/FormLoginAuthenticatorTest.php
dmaicher: yeah I followed the existing test cases for the username :man_shrugging: also happy to just keep the one test for the array case

src/Symfony/Component/Security/Http/Authenticator/FormLoginAuthenticator.php
derrabus: I've seen codebases where value objects were used to "protect" sensitive data (e.g. they throw if you try to serialize them). I don't know if anyone uses this kind of objects in this context. But if someone does we'd break their app with an LTS bugfix release which would be unfortunate. Let's have the authenticator accept anything that can be cast to a string.

mtarld (approved):

nicolas-grekas:
Thank you @dmaicher.

PR #51450, opened 2023-08-22
[Mailer] [Smtp] Add DSN param `peer_fingerprint` for fingerprint verification

xdavidwu:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | symfony/symfony-docs#18778

This allows a more secure TLS connection for SMTP on even self-signed certificates, by verifying certificate fingerprint.

### DIFF
--- a/src/Symfony/Component/Mailer/CHANGELOG.md
+++ b/src/Symfony/Component/Mailer/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+ * Add DSN parameter `peer_fingerprint` to verify TLS certificate fingerprint
+
 6.3
 ---
 
--- a/src/Symfony/Component/Mailer/Tests/Transport/Smtp/EsmtpTransportFactoryTest.php
+++ b/src/Symfony/Component/Mailer/Tests/Transport/Smtp/EsmtpTransportFactoryTest.php
@@ -110,6 +110,23 @@ class EsmtpTransportFactoryTest extends TransportFactoryTestCase
         ];
 
         $transport = new EsmtpTransport('example.com', 465, true, null, $logger);
+        /** @var SocketStream $stream */
+        $stream = $transport->getStream();
+        $streamOptions = $stream->getStreamOptions();
+        $streamOptions['ssl']['peer_fingerprint'] = '6A1CF3B08D175A284C30BC10DE19162307C7286E';
+        $stream->setStreamOptions($streamOptions);
+
+        yield [
+            new Dsn('smtps', 'example.com', '', '', 465, ['peer_fingerprint' => '6A1CF3B08D175A284C30BC10DE19162307C7286E']),
+            $transport,
+        ];
+
+        yield [
+            Dsn::fromString('smtps://:@example.com?peer_fingerprint=6A1CF3B08D175A284C30BC10DE19162307C7286E'),
+            $transport,
+        ];
+
+        $transport = new EsmtpTransport('example.com', 465, true, null, $logger);
         $transport->setLocalDomain('example.com');
 
         yield [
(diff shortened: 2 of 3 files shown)


src/Symfony/Component/Mailer/Tests/Transport/Smtp/EsmtpTransportFactoryTest.php
```diff
@@ -109,6 +109,23 @@ public static function createProvider(): iterable
             $transport,
         ];
 
+        $transport = new EsmtpTransport('example.com', 465, true, null, $logger);
+        /** @var SocketStream $stream */
+        $stream = $transport->getStream();
+        $streamOptions = $stream->getStreamOptions();
+        $streamOptions['ssl']['peer_fingerprint'] = '6A1CF3B08D175A284C30BC10DE19162307C7286E';
```
mtarld, line 116: Does it make sense to name it `verify_peer_fingerprint` like `verify_peer_name` is named?

src/Symfony/Component/Mailer/Transport/Smtp/EsmtpTransportFactory.php
```diff
@@ -29,17 +29,21 @@ public function create(Dsn $dsn): TransportInterface
...
-            /** @var SocketStream $stream */
-            $stream = $transport->getStream();
-            $streamOptions = $stream->getStreamOptions();
+        /** @var SocketStream $stream */
+        $stream = $transport->getStream();
+        $streamOptions = $stream->getStreamOptions();
 
+        if ('' !== $dsn->getOption('verify_peer') && !filter_var($dsn->getOption('verify_peer', true), \FILTER_VALIDATE_BOOL)) {
             $streamOptions['ssl']['verify_peer'] = false;
             $streamOptions['ssl']['verify_peer_name'] = false;
```
mtarld, line 38: Maybe we should set the `peer_fingerprint` to `false` here as well?

src/Symfony/Component/Mailer/Tests/Transport/Smtp/EsmtpTransportFactoryTest.php
xdavidwu: I think aligning with PHP options name would give a better idea what it actually does. Also `verify_*` sounds like a boolean, but we need to specify the fingerprint.

src/Symfony/Component/Mailer/Transport/Smtp/EsmtpTransportFactory.php
xdavidwu: `verify_peer` verifies certificate content, while `peer_fingerprint` works independently (with or without `verify_peer`) and verifies the certificate by hash without looking into it.

My motivation of this patch was actually to use `peer_fingerprint` with `verify_peer=false`, to ensure server authenticity even when using self-signed certs.

mtarld (approved):

src/Symfony/Component/Mailer/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+ * Add DSN parameter `peer_fingerprint` to verify TLS certficate fingerprint
```
nicolas-grekas:
```suggestion
 * Add DSN parameter `peer_fingerprint` to verify TLS certificate fingerprint
```

src/Symfony/Component/Mailer/Transport/Smtp/EsmtpTransportFactory.php
```diff
@@ -29,17 +29,21 @@ public function create(Dsn $dsn): TransportInterface
...
+        }
 
-            $stream->setStreamOptions($streamOptions);
+        if (null !== ($peerFingerprint = $dsn->getOption('peer_fingerprint'))) {
```
nicolas-grekas:
```suggestion
        if (null !== $peerFingerprint = $dsn->getOption('peer_fingerprint')) {
```

nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @xdavidwu.

PR #51425, opened 2023-08-18
[FrameworkBundle][Validator] Deprecate annotation occurrences

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Tickets       | Part of https://github.com/symfony/symfony/issues/51381
| License       | MIT
| Doc PR        | -

 * Deprecate `framework.validation.enable_annotations` in favor of `framework.validation.enable_attributes`
 * Deprecate `framework.serializer.enable_annotations` in favor of use `framework.serializer.enable_attributes`
 * Deprecate `ValidatorBuilder::enableAnnotationMapping()` in favor of `ValidatorBuilder::enableAttributeMapping()`
 * Deprecate `ValidatorBuilder::disableAnnotationMapping()` in favor of `ValidatorBuilder::disableAttributeMapping()`
 * Deprecate `AnnotationLoader` in favor of `AttributeLoader`

PR #51454, opened 2023-08-22
[Validator] Un-deprecate passing an annotation reader to AnnotationLoader

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Follow-up to #51425
| License       | MIT
| Doc PR        | N/A

#51425 deprecated the `AnnotationLoader` class and its replacement already does not accept an annotation reader anymore. Thus, the deprecation of the constructor parameter is redundant, which is why I propose to take it back.

cc @alexandre-daubois

### DIFF
--- a/src/Symfony/Component/Validator/Tests/Mapping/Loader/AttributeLoaderTest.php
+++ b/src/Symfony/Component/Validator/Tests/Mapping/Loader/AttributeLoaderTest.php
@@ -30,9 +30,10 @@ use Symfony\Component\Validator\Constraints\Type;
 use Symfony\Component\Validator\Constraints\Valid;
 use Symfony\Component\Validator\Mapping\ClassMetadata;
 use Symfony\Component\Validator\Mapping\Loader\AnnotationLoader;
+use Symfony\Component\Validator\Mapping\Loader\AttributeLoader;
 use Symfony\Component\Validator\Tests\Fixtures\ConstraintA;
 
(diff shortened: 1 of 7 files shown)


nicolas-grekas (approved):

src/Symfony/Component/Validator/Tests/Mapping/Loader/AttributeLoaderTest.php
```diff
@@ -212,7 +213,7 @@ public function testLoadGroupSequenceProviderAnnotation()
 
     protected function createAnnotationLoader(): AnnotationLoader
     {
-        return new AnnotationLoader();
+        return new AttributeLoader();
```
stof, line 216: This does not match the return type

derrabus: It does, `AttributeLoader` extends `AnnotationLoader`. This class (`AttributeLoaderTest`) has two child classes which override this method. I have to keep the return type wide, otherwise I would break those child classes.

alexandre-daubois:
Just for the record, I'm working on deprecating the serializer's AnnotationLoader too (and replacing it by an AttributeLoader). 🙂 I'll include this same undeprecation in the coming PR

PR #43028, opened 2021-09-14
[Console] Relax return type on HelperInterface::getName()

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 6.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | Part of #43021
| License       | MIT
| Doc PR        | N/A

Removing this one return type would allow Doctrine ORM 2.9 and DBAL 2.13 to work with Symfony 6.

PR #51458, opened 2023-08-22
[Console] Revert adding return type on Command::execute()

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

I tried running a Symfony v7 "webapp-pack" and this is all that is needed to not force doctrine to release new major versions to support it.

### DIFF
--- a/.github/expected-missing-return-types.diff
+++ b/.github/expected-missing-return-types.diff
@@ -65,6 +65,13 @@ diff --git a/src/Symfony/Component/Console/Command/Command.php b/src/Symfony/Com
 +    protected function configure(): void
      {
      }
+@@ -181,5 +181,5 @@ class Command
+      * @see setCode()
+      */
+-    protected function execute(InputInterface $input, OutputInterface $output)
++    protected function execute(InputInterface $input, OutputInterface $output): int
+     {
+         throw new LogicException('You must override the execute() method in the concrete command class.');
 @@ -195,5 +195,5 @@ class Command
       * @return void
       */
--- a/src/Symfony/Component/Console/Command/Command.php
+++ b/src/Symfony/Component/Console/Command/Command.php
@@ -180,7 +180,7 @@ class Command
      *
      * @see setCode()
      */
-    protected function execute(InputInterface $input, OutputInterface $output): int
+    protected function execute(InputInterface $input, OutputInterface $output)
     {
         throw new LogicException('You must override the execute() method in the concrete command class.');
     }


src/Symfony/Component/Config/Definition/ConfigurationInterface.php
```diff
@@ -22,6 +22,8 @@ interface ConfigurationInterface
 {
     /**
      * Generates the configuration tree builder.
+     *
+     * @return TreeBuilder
      */
-    public function getConfigTreeBuilder(): TreeBuilder;
+    public function getConfigTreeBuilder();
```
stof: this one has already been delayed from 6.0 to 7.0, throwing a deprecation warning since 5.3.

Same for the Command execute method. If 2.5 years is not enough to get them to release a version with a return type, what will happen in 8.0 in 2 years ? Will we delay it again ?

stof:
Btw, the dev version of TwigExtraBundle has the native return type.

src/Symfony/Component/Config/Definition/ConfigurationInterface.php
nicolas-grekas: Indeed. We need to be sure maintainers have received the message. /cc @Seldaek for monolog-bundle and @derrabus I guess for Doctrine?

Good for twig-extra-bundle (but it's still is missing a return type for TwigExtraExtension::load)

stof: Well, for the void type in load, I'm fine with delaying it to give more time if needed, as that's much newer (we started reporting a deprecation for it in 6.2 only)

src/Symfony/Bridge/Doctrine/DependencyInjection/AbstractDoctrineExtension.php
```diff
@@ -313,9 +313,11 @@ private function detectMappingType(string $directory, ContainerBuilder $containe
     /**
      * Loads a configured object manager metadata, query or result cache driver.
      *
+     * @return void
+     *
      * @throws \InvalidArgumentException in case of unknown driver type
      */
-    protected function loadObjectManagerCacheDriver(array $objectManager, ContainerBuilder $container, string $cacheName): void
+    protected function loadObjectManagerCacheDriver(array $objectManager, ContainerBuilder $container, string $cacheName)
```
stof: this one could be skipped from this PR. There is an open PR in DoctrineBundle removing the method override entirely (which can be done in a minor version) because the implementation in DoctrineBundle matches the implementation in the bridge in all supported Symfony versions (as it does not support older versions anymore where the bridge was different)

nicolas-grekas: thanks for the hint (here is the PR in question: https://github.com/doctrine/DoctrineBundle/pull/1692)

src/Symfony/Component/Config/Definition/ConfigurationInterface.php
nicolas-grekas: monolog-bunde updated https://github.com/symfony/monolog-bundle/pull/461

nicolas-grekas:
PR updated to target only `Command::execute()`

stof:
@nicolas-grekas `Command::execute` is also a method for which we already delayed the update. So do we have a plan to avoid delaying it forever ?

nicolas-grekas:
Forever I don't know, but we must ensure we keep working with Doctrine, and their next major is not close to any adoption. DBAL also would need a new major. We don't have a choice for now.

derrabus:
Do you have a list of Doctrine commands that are affected by this? I don't want to be the one blocking that return type forever. The ORM should be fine and the DBAL has two commands that need to be patched. Anything else?

stof:
@derrabus looks like most ORM commands in 2.16.x are affected. They don't have native return types.

derrabus:
Right, my bad. I was looking at the wrong branch. 🙈

derrabus:
See doctrine/orm#10919.

`Command::execute()` wasn't the only method that was in my way: The other one was `HelperInterface::getName()`. With that PR applied (and after patching the DBAL the same way), the ORM's test suite runs green with Symfony 7.

wouterj:
`HelperInterface::getName()` is also one that we delayed explicitly for Doctrine in 6.0 already: #43028

So if the ORM PR gets merged, I think we can close this one without merging. I fully agree with @stof that we should be very reluctant to delay return types for yet another major version, even if this means pushing maintainers for important ecosystem projects like Doctrine or Monolog imho.

derrabus:
PRs have been merged to DBAL 3.7 and ORM 2.17.

nicolas-grekas:
Closing then, thanks for the update on Doctrine's side.

PR #51468, opened 2023-08-23
[Messenger] Fix forced bus name gone after an error in delayed message handling

valtzu:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

When consuming messages from an external source which do not have metadata/headers like bus name, you have to force the bus name with `--bus` in the `messenger:consume` command. However, if in the message handler you dispatch another message with `DispatchAfterCurrentBusStamp`, which then fails, end result is that the original message from external source is retried, but that forced bus name (`BusNameStamp`) is lost – so it will be retried on default bus instead. This isn't intended behavior, is it?

---

Not sure if changing `DelayedMessageHandlingException::__construct` signature is considered a BC break, it is not marked as internal so I am a bit worried :thinking:

### DIFF
--- a/src/Symfony/Component/Messenger/Exception/DelayedMessageHandlingException.php
+++ b/src/Symfony/Component/Messenger/Exception/DelayedMessageHandlingException.php
@@ -11,6 +11,8 @@
 
 namespace Symfony\Component\Messenger\Exception;
 
+use Symfony\Component\Messenger\Envelope;
+
 /**
  * When handling queued messages from {@link DispatchAfterCurrentBusMiddleware},
  * some handlers caused an exception. This exception contains all those handler exceptions.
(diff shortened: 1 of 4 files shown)


valtzu:
Added a test case to demonstrate the issue.

---

I wonder if `HandledStamp` should still be dropped :thinking:

nicolas-grekas:
Thank you @valtzu.

jakubtobiasz:
@nicolas-grekas shouldn't it be this way?
```
    public function __construct(array $exceptions, ?Envelope $envelope = null)
    {
		if (null === $envelope) {
            trigger_deprecation(/* ... */);
        }

		// ...
    }
```

Currently it seems like a BC for me. I know it's a bug, but it can be fixed in a backward compatible way.

Issue #51381, opened 2023-08-14
[RFC] Consistently use the word "attribute" in all symbols


derrabus:
### Description

Doctrine Annotations have served us well, but PHP attributes are the future.

Our strategy for Symfony 6 was to support Doctrine Annotations and PHP attributes as two ways of annotating pieces of configuration to code. In Symfony 7, attributes will be the only supported way. However, we still use the word "annotation" in the codebase in an abstract sense (PHP attributes are a form of annotation). This might be confusing, so we might consider using the work "attribute" everywhere consistently.

I've opened this issue to collect all occurrences of the term and to discuss the possible impact on downstream projects.

- [x] The `serializer` integration of FrameworkBundle exposes a configuration option `enable_annotations`. Rename to `enable_attributes`.
- [x] The `validation` integration of FrameworkBundle exposes a configuration option `enable_annotations`. Rename to `enable_attributes`.
- [x] The Routing and Serializer components organize their attribute classes in a namespace called `Annotation`. Rename that namespace to `Attribute`.
- [x] When loading routes from attributes, the loader type is `annotation` although `attribute` is already an alias for that. Make `attribute` the only possible type.
- [x] Serializer uses a class named `AnnotationLoader` to load metadata from attributes. Rename to `AttributeLoader`.
- [x] Validation uses a class named `AnnotationLoader` to load metadata from attributes. Rename to `AttributeLoader`.
- [x] Routing contains classes `AnnotationClassLoader` and `AnnotationDirectoryLoader` for loading routes from attributes. Rename to `AttributeClassLoader` and `AttributeDirectoryLoader`.

If you find more TODOs, please drop a comment.

nicolas-grekas:
I'm fine with all items except the second one:

>  The Routing and Serializer components organize their attribute classes in a namespace called Annotation. Rename that namespace to Attribute?

I think this can lead to costly migrations from 6 to 7. The benefit might not be worth the troubles.
If we really want to go this way, we could add the Attribute namespace and keep the Annotation one as an alias. But I wouldn't deprecate in the same version that introduces the alternative.

For validation, I'm also wondering if the move shouldn't do one more step and clean the way we create those objects. Right now, constructors are a bit strange. Maybe relying only on named args is possible?

derrabus:
> > The Routing and Serializer components organize their attribute classes in a namespace called Annotation. Rename that namespace to Attribute?
> 
> I think this can lead to costly migrations from 6 to 7.

[...]

PR #51470, opened 2023-08-24
[FrameworkBundle][Serializer] Deprecate annotations

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Tickets       | -
| License       | MIT
| Doc PR        | -

Part of https://github.com/symfony/symfony/issues/51381


src/Symfony/Component/Serializer/Mapping/Loader/AnnotationLoader.php
```diff
@@ -243,7 +252,7 @@ private function isKnownAttribute(string $attributeName): bool
     /**
      * @return object[]
      */
-    private function getClassAnnotations(\ReflectionClass $reflector): array
+    private function getClassAttributes(\ReflectionClass $reflector): array
```
derrabus: Please revert. This method reads Doctrine annotations only. Thus, its name is accurate. On 7.0, the method has been removed already.

src/Symfony/Component/Serializer/Mapping/Loader/AnnotationLoader.php
```diff
@@ -258,7 +267,7 @@ private function getClassAnnotations(\ReflectionClass $reflector): array
     /**
      * @return object[]
      */
-    private function getMethodAnnotations(\ReflectionMethod $reflector): array
+    private function getMethodAttributes(\ReflectionMethod $reflector): array
```
derrabus: Same

src/Symfony/Component/Serializer/Mapping/Loader/AnnotationLoader.php
```diff
@@ -273,7 +282,7 @@ private function getMethodAnnotations(\ReflectionMethod $reflector): array
     /**
      * @return object[]
      */
-    private function getPropertyAnnotations(\ReflectionProperty $reflector): array
+    private function getPropertyAttributes(\ReflectionProperty $reflector): array
```
derrabus: Same

src/Symfony/Component/Serializer/Mapping/Loader/AnnotationLoader.php
```diff
@@ -176,7 +175,7 @@ public function loadClassMetadata(ClassMetadataInterface $classMetadata): bool
         return $loaded;
     }
 
-    public function loadAnnotations(\ReflectionMethod|\ReflectionClass|\ReflectionProperty $reflector): iterable
```
derrabus, line 190: I wonder why this method is public anyway. All calls come from within this class. Let's just deprecate it without replacement. The new `loadAttributes()` method can be private.

derrabus (changes requested):

src/Symfony/Component/Serializer/Mapping/Loader/AnnotationLoader.php
```diff
@@ -203,16 +202,26 @@ public function loadAnnotations(\ReflectionMethod|\ReflectionClass|\ReflectionPr
...
+            yield from $this->getPropertyAttributes($reflector);
         }
     }
 
+    /**
+     * @deprecated since Symfony 6.4, use "loadAttributes()" instead
+     */
+    public function loadAnnotations(\ReflectionMethod|\ReflectionClass|\ReflectionProperty $reflector): iterable
+    {
+        trigger_deprecation('symfony/serializer', '6.4', 'Method "%s()" is deprecated, use "loadAttributes()" instead.', __METHOD__);
```
mtarld: Isn't it triggering the deprecation twice? One with the phpdoc and one with the `trigger_deprecation`

src/Symfony/Component/Serializer/Mapping/Loader/AnnotationLoader.php
alexandre-daubois: I think I've confused it with the Validator's AnnotationLoader, which reads attributes. I reverted

src/Symfony/Component/Serializer/Mapping/Loader/AnnotationLoader.php
alexandre-daubois: Changed visibility to private and removed the lines from the Changelog about it

src/Symfony/Component/Serializer/Mapping/Loader/AnnotationLoader.php
alexandre-daubois: I don't think so :thinking: If I remember correctly, the PHPDoc one is more for the IDE and DX than anything else

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -1935,12 +1935,7 @@ private function registerSerializerConfiguration(array $config, ContainerBuilder
                 $container->removeDefinition('serializer.mapping.cache_class_metadata_factory');
             }
 
-            $annotationLoader = new Definition(
-                AnnotationLoader::class,
-                [new Reference('annotation_reader', ContainerInterface::NULL_ON_INVALID_REFERENCE)]
-            );
-
-            $serializerLoaders[] = $annotationLoader;
+            $serializerLoaders[] = new Definition(AttributeLoader::class);
```
derrabus: If you don't pass the annotation reader, annotations will stop working immediately. This is not what we want for 6.4, is it?

derrabus (changes requested):

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
alexandre-daubois: That's right, I updated this part

alexandre-daubois:
Rebased and conflicts fixed

nicolas-grekas:
Thank you @alexandre-daubois.

PR #51476, opened 2023-08-24
[Serializer] Allow Context to target classes

### LINKED ISSUE #49450: [Serializer] Allow `Context` attribute to target class
### Description

As a developer I'd like to use `Context` attribute on a class.

In the example bellow serializer will try to set the `readonly` `$type` property (it is already set via the constructor from the concrete class.

The solution I'd like is to exclude `$type` from denormalization. `Context` attribute on a class would allow this intent to be written in clear form.

Currently I have to annotate with `Context` each property of the class in question (this is error prone).
I've tried few structural changes on the DTO classes but all of them feel like monkey-patching.

### Example

```php
<?php

namespace App\Dto;

use Symfony\Component\Serializer\Annotation\Context;
use Symfony\Component\Serializer\Annotation\DiscriminatorMap;
use Symfony\Component\Serializer\Normalizer\AbstractNormalizer;

#[DiscriminatorMap(
    typeProperty: 'type',
    mapping: [
        PositionPointDto::POSITION_TYPE => PositionPointDto::class,
    ],
)]
#[Context( /* <-- this cannot compile -- */
    denormalizationContext: [
        AbstractNormalizer::IGNORED_ATTRIBUTES => [
            'type',
        ],
    ],
)]
abstract readonly class PositionDto
{
    public const POSITION_TYPE = 'abstract_position';

    protected function __construct(
        public string $type = self::POSITION_TYPE,
    ) {
    }
}
```

```php
<?php

namespace App\Dto;

final readonly class PositionPointDto extends PositionDto
{
    public const POSITION_TYPE = 'point';

    public function __construct(
        public int $x,
        public int $y,
    ) {
        parent::__construct(
            self::POSITION_TYPE,
        );
    }
}
```

mtarld:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #49450
| License       | MIT
| Doc PR        | TODO

Allow to target class with the `Context` attribute. The related context can be overridden by each property.

### DIFF
--- a/src/Symfony/Component/Serializer/Mapping/Loader/AnnotationLoader.php
+++ b/src/Symfony/Component/Serializer/Mapping/Loader/AnnotationLoader.php
@@ -57,6 +57,7 @@ class AnnotationLoader implements LoaderInterface
         $className = $reflectionClass->name;
         $loaded = false;
         $classGroups = [];
+        $classContextAnnotation = null;
 
         $attributesMetadata = $classMetadata->getAttributesMetadata();
 
@@ -71,6 +72,12 @@ class AnnotationLoader implements LoaderInterface
 
             if ($annotation instanceof Groups) {
                 $classGroups = $annotation->getGroups();
+
+                continue;
+            }
+
+            if ($annotation instanceof Context) {
+                $classContextAnnotation = $annotation;
             }
         }
 
@@ -81,6 +88,10 @@ class AnnotationLoader implements LoaderInterface
             }
 
             if ($property->getDeclaringClass()->name === $className) {
+                if ($classContextAnnotation) {
+                    $this->setAttributeContextsForGroups($classContextAnnotation, $attributesMetadata[$property->name]);
+                }
+
                 foreach ($classGroups as $group) {
                     $attributesMetadata[$property->name]->addGroup($group);
                 }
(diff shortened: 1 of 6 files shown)


chalasr (approved):

src/Symfony/Component/Serializer/Mapping/Loader/AnnotationLoader.php
```diff
@@ -71,6 +72,14 @@ public function loadClassMetadata(ClassMetadataInterface $classMetadata): bool
...
             if ($annotation instanceof Groups) {
                 $classGroups = $annotation->getGroups();
+
+                continue;
+            }
+
+            if ($annotation instanceof Context) {
+                $classContextAnnotation = $annotation;
+
+                continue;
```
dunglas:
```suggestion
```

Continue is useless here.

dunglas (approved):

nicolas-grekas (approved):
I fixed a few things, see https://github.com/symfony/symfony/compare/dfcec29218f5328284a2caa4460bf37e5277eb21..caaf0a6b71c3bcc8650e11a23bde6df4a6e4fb90

nicolas-grekas:
Thank you @mtarld.

PR #51478, opened 2023-08-24
[Process] Fix bug where $this->callback is never null, resulting in bad argument

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | Not needed

Hi!

Caused by #51068 - by the time `$this->getDescriptors()` is called, `$this->callback` is ALWAYS a callable thanks to `$this->callback = $this->buildCallback($callback)`.

Previously, there was a separate `hasCallback` property that handled this logic. I chose to NOT re-add that because it was only needed in `getDescriptors()` and that is only called from this one place: easier to pass `$hasCallback` as an argument.

Caught by the symfony/ux test suite ❤️ 

Cheers!

### DIFF
--- a/src/Symfony/Component/Process/Process.php
+++ b/src/Symfony/Component/Process/Process.php
@@ -304,7 +304,7 @@ class Process implements \IteratorAggregate
         $this->resetProcessData();
         $this->starttime = $this->lastOutputTime = microtime(true);
         $this->callback = $this->buildCallback($callback);
-        $descriptors = $this->getDescriptors();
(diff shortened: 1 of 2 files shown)


stof:
is there a way to write a test catching this, so that we don't have similar regressions ?

weaverryan:
Test added 👍

fabpot (approved):

fabpot:
Thank you @weaverryan.

PR #51483, opened 2023-08-25
[FrameworkBundle][Routing] Deprecate annotations

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Tickets       | Part of https://github.com/symfony/symfony/issues/51381
| License       | MIT
| Doc PR        | -

 * Deprecate the `routing.loader.annotation` service, use the `routing.loader.attribute` service instead
 * Deprecate the `routing.loader.annotation.directory` service, use the `routing.loader.attribute.directory` service instead
 * Deprecate the `routing.loader.annotation.file` service, use the `routing.loader.attribute.file` service instead
 * Deprecate `AnnotationClassLoader`, use `AttributeClassLoader` instead
 * Deprecate `AnnotationDirectoryLoader`, use `AttributeDirectoryLoader` instead
 * Deprecate `AnnotationFileLoader`, use `AttributeFileLoader` instead
 * Deprecate `AnnotatedRouteControllerLoader`, use `AttributeRouteControllerLoader`


src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -1222,13 +1221,6 @@ private function registerRouterConfiguration(array $config, ContainerBuilder $co
         if (!class_exists(Psr4DirectoryLoader::class)) {
             $container->removeDefinition('routing.loader.psr4');
         }
-
-        if ($this->isInitializedConfigEnabled('annotations') && (new \ReflectionClass(AnnotationClassLoader::class))->hasProperty('reader')) {
-            $container->getDefinition('routing.loader.annotation')->setArguments([
-                new Reference('annotation_reader'),
-                '%kernel.environment%',
-            ]);
-        }
```
derrabus: Are you sure we can just delete this?

alexandre-daubois: I think it's a little too much on second though. I've updated to use the `routing.loader.attribute` service and keeping the logic.

alexandre-daubois:
Rebased and conflicts fixed

src/Symfony/Bundle/FrameworkBundle/Routing/AttributedRouteControllerLoader.php
```diff
@@ -0,0 +1,23 @@
...
+namespace Symfony\Bundle\FrameworkBundle\Routing;
+
+/**
+ * AttributedRouteControllerLoader is an implementation of AttributeClassLoader
+ * that sets the '_controller' default based on the class and method names.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ * @author Alexandre Daubois <alex.daubois@gmail.com>
+ */
+class AttributedRouteControllerLoader extends AnnotatedRouteControllerLoader
```
chalasr:
```suggestion
class AttributeRouteControllerLoader extends AnnotatedRouteControllerLoader
```
as `attributed` has a different meaning

chalasr (changes requested):

src/Symfony/Bundle/FrameworkBundle/Routing/AttributedRouteControllerLoader.php
alexandre-daubois: Updated, thanks!

src/Symfony/Bundle/FrameworkBundle/Resources/config/routing.php
```diff
@@ -92,26 +92,35 @@
...
-        ->set('routing.loader.annotation', AnnotatedRouteControllerLoader::class)
+        ->set('routing.loader.attribute', AttributeRouteControllerLoader::class)
             ->args([
                 '%kernel.environment%',
             ])
             ->tag('routing.loader', ['priority' => -10])
 
-        ->set('routing.loader.annotation.directory', AnnotationDirectoryLoader::class)
+        ->alias('routing.loader.annotation', 'routing.loader.attribute')
+            ->deprecate('symfony/router', '6.4', 'The "%alias_id%" service is deprecated, use the "routing.loader.attribute" service instead.')
```
fancyweb:
```suggestion
            ->deprecate('symfony/routing', '6.4', 'The "%alias_id%" service is deprecated, use the "routing.loader.attribute" service instead.')
```

fancyweb: Same everywhere

src/Symfony/Bundle/FrameworkBundle/Routing/AttributeRouteControllerLoader.php
```diff
@@ -0,0 +1,23 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\FrameworkBundle\Routing;
+
+/**
+ * AttributedRouteControllerLoader is an implementation of AttributeClassLoader
```
fancyweb:
```suggestion
 * AttributeRouteControllerLoader is an implementation of AttributeClassLoader
```

src/Symfony/Component/Routing/Loader/AnnotationClassLoader.php
```diff
@@ -47,6 +47,8 @@
  *         }
  *     }
  *
+ * @deprecated since Symfony 6.4, use  {@see AttributeClassLoader} instead
```
fancyweb:
```suggestion
 * @deprecated since Symfony 6.4, use {@see AttributeClassLoader} instead
```

src/Symfony/Component/Routing/Loader/AnnotationDirectoryLoader.php
```diff
@@ -18,9 +18,11 @@
  * AnnotationDirectoryLoader loads routing information from annotations set
  * on PHP classes and methods.
  *
+ * @deprecated since Symfony 6.4, use  {@see AttributeDirectoryLoader} instead
```
fancyweb:
```suggestion
 * @deprecated since Symfony 6.4, use {@see AttributeDirectoryLoader} instead
```

src/Symfony/Component/Routing/Loader/AnnotationDirectoryLoader.php
```diff
@@ -18,9 +18,11 @@
...
+ *
```
fancyweb: If this class is deprecated, then there should be a trigger_deprecation here at the top of it.

fancyweb:
Instead of the current strategy, what about copy-pasting the existing Annotation classes code in the new Attribute classes, and then make existing Annotation classes extends those new Attribute classes.

alexandre-daubois:
@fancyweb I think that the current way of doing is better for upmerges. It will be easier for potential fixes if the annotation classes remain the same. Otherwise, all fixes in `Annotation*Loader` will have to be manually reflected in `Attribute*Loader` when merging to upper branches

src/Symfony/Component/Routing/Loader/AnnotationDirectoryLoader.php
alexandre-daubois: I agree but this triggers deprecations in a lot of places, like FWB or Security bundle functional tests. There are place like `src/Symfony/Component/DependencyInjection/ContainerAwareInterface.php` where we did not put the `trigger_deprecation` call. Maybe we should do the same here? :thinking: Because even if it's deprecated, we still need to extends these classes with the current strategy

src/Symfony/Component/Routing/Loader/AttributeFileLoader.php
```diff
@@ -0,0 +1,23 @@
...
+namespace Symfony\Component\Routing\Loader;
+
+/**
+ * AttributeFileLoader loads routing information from attributes set
+ * on a PHP class and its methods.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ * @author Alexandre Daubois <alex.daubois@gmail.com>
+ */
+class AttributeFileLoader extends AnnotationFileLoader
```
nicolas-grekas: I think the other way around would be better if possible: old extends new. The reason is that this will make the code in 6.4 as close as possible to the one we anticipate in 7.0. Please consider this swap for all classes.

src/Symfony/Component/Routing/Loader/AnnotationDirectoryLoader.php
alexandre-daubois: I managed to add it :slightly_smiling_face:

src/Symfony/Component/Routing/Loader/AttributeFileLoader.php
alexandre-daubois: Alright, I did the swap for all three classes, added `trigger_deprecation` on top of them and updated `.github/expected-missing-return-types.diff`

nicolas-grekas (approved):
The class_alias trick is the best!

nicolas-grekas:
Thank you @alexandre-daubois.

alexandre-daubois:
I'll work on the cleanup :slightly_smiling_face:

PR #48156, opened 2022-11-08
[Mime] When serializing File parts convert to string to allow proper unserialization

ovrflo:
| Q             | A
| ------------- | ---
| Branch?       | 6.2 
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #47991
| License       | MIT

I took a stab at #47991. I found an existing serialization test and added the missing case, for regression purposes.

The bug was a `\BadMethodCallException` thrown in `\Symfony\Component\Mime\Part\DataPart::__wakeup` when it would encounter a `File` instance instead of a `string`.

My fix comes with the implication that, when serializing a `TextPart` with a `File`, it will read and serialize the content of the file, discarding the `File` instance in the process. The assumption being that the machine producing a serialized message might not be the machine consuming that message.

PR #47462, opened 2022-09-02
[Mime] Simplify adding Parts to an Email

fabpot:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Tickets       | n/a 
| License       | MIT
| Doc PR        | 

While fixing some MIME bugs, I realized we have a substantial cyclomatic complexity due to the attach/attachFromPath/embed/embedFromPath/attachPart methods on the Email class.

This PR simplifies all of that and introduces a way to have a file for TextPart as well (via the new `file://` notation) and it keeps the lazy-loading feature which was why those methods were introduced in the first place.

From now, I've kept all the methods, but I'm wondering if we should deprecate all of them and only keep `attachPart()` (which I would like to rename `addPart()`).

PR #51489, opened 2023-08-25
[Mime] Fix email (de)serialization issues

### LINKED ISSUE #47991: [Mime][Mailer][Profiler] 6.2-dev Cannot unserialize Symfony\Component\Mime\Part\DataPart

6.2-dev

### Description

I am sending a message with Mailer. And when the profile data is being collected I get an error. Stack strace to relevant parents.

I get an error, full stack trace:
```
Scenario: I can change my password                             # features/user/change_password_form.feature:12
      And I should get a "change_password_notification" email sent # features/user/change_password_form.feature:27
        BadMethodCallException: Cannot unserialize Symfony\Component\Mime\Part\DataPart in vendor/symfony/mime/Part/DataPart.php:162
        Stack trace:
        #0 [internal function]: Symfony\Component\Mime\Part\DataPart->__wakeup()
```

### How to reproduce

Send an email message with a file attached and profiler enabled.

### Possible Solution

The DataPart parent is...
```
array:6 [
  "body" => Symfony\Component\Mime\Part\BodyFile {#11584
    -path: "xxx/_CWA/api-components-bundle/src/Resources/views/assets/images/email_logo.png"
  }
  "charset" => null
  "subtype" => "png"
  "disposition" => "inline"
  "name" => "@SilverbackApiComponents/assets/images/email_logo.png"
  "encoding" => "base64"
]
```
As body is not a string, it causes this error.

For this perhaps... Symfony\Component\Mime\Part\BodyFile::__toString could be implemented
https://github.com/symfony/mime/blob/6.2/Part/BodyFile.php

and try to cast as a string in Symfony\Component\Mime\Part\DataPart::__wakeup
https://github.com/symfony/mime/blob/6.2/Part/DataPart.php#L161

### Additional Context

_No response_

X-Coder264:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #47991
| License       | MIT
| Doc PR        | -

https://github.com/symfony/symfony/pull/48156 fixed https://github.com/symfony/symfony/issues/47991 while introducing a big breaking change (the `File` lazy load feature is broken and that was the whole point of that class when it was introduced in https://github.com/symfony/symfony/pull/47462 as that feature existed even prior to that PR) on a minor Symfony version (updating from 6.1 to 6.2 broke our application). More context can be found here: https://github.com/symfony/symfony/pull/48156#issuecomment-1693489346

This PR aims to revert back the `attachFromPath` behavior to what it was before https://github.com/symfony/symfony/pull/48156 while still fixing the deserialization issue reported in #47991

The first commit fixes the serialization logic to work the same way it had worked on both 5.4 and 6.1 (which means we are reverting https://github.com/symfony/symfony/pull/48156), while the second commit fixes the deserialization issue reported in #47991.

I've also added tests to prevent serialization/deserialization regressions in the future.

### DIFF
--- a/src/Symfony/Component/Mime/Part/DataPart.php
+++ b/src/Symfony/Component/Mime/Part/DataPart.php
@@ -156,7 +156,7 @@ class DataPart extends TextPart
             throw new \BadMethodCallException('Cannot unserialize '.__CLASS__);
         }
         foreach (['body', 'charset', 'subtype', 'disposition', 'name', 'encoding'] as $name) {
-            if (null !== $this->_parent[$name] && !\is_string($this->_parent[$name])) {
+            if (null !== $this->_parent[$name] && !\is_string($this->_parent[$name]) && !$this->_parent[$name] instanceof File) {
                 throw new \BadMethodCallException('Cannot unserialize '.__CLASS__);
             }
             $r = new \ReflectionProperty(TextPart::class, $name);
--- a/src/Symfony/Component/Mime/Part/TextPart.php
+++ b/src/Symfony/Component/Mime/Part/TextPart.php
@@ -226,7 +226,7 @@ class TextPart extends AbstractPart
     public function __sleep(): array
     {
         // convert resources to strings for serialization
-        if (null !== $this->seekable || $this->body instanceof File) {
+        if (null !== $this->seekable) {
             $this->body = $this->getBody();
             $this->seekable = null;
         }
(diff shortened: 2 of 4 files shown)


src/Symfony/Component/Mime/Part/DataPart.php
```diff
@@ -156,6 +156,9 @@ public function __wakeup()
             throw new \BadMethodCallException('Cannot unserialize '.__CLASS__);
         }
         foreach (['body', 'charset', 'subtype', 'disposition', 'name', 'encoding'] as $name) {
+            if ('body' === $name && isset($this->_parent[$name]) && $this->_parent[$name] instanceof File) {
+                $this->_parent[$name] = file_get_contents($this->_parent[$name]->getPath());
```
X-Coder264, line 160: This is the same logic as how this would usually happen in `\Symfony\Component\Mime\Part\TextPart::getBody` if the email had not been serialized.

fabpot (approved):

fabpot:
Thank you @X-Coder264.

PR #51493, opened 2023-08-26
Remove `GuardEvent::getContext()` method and add `HasContextTrait` trait

hhamon:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | Related to #51484
| License       | MIT
| Doc PR        | ~

As discussed with @lyrixx, the `GuardEvent::getContext` method was confusing as the context given as the 3rd argument to the WorflowInterface::appy() method is never passed along to the guard events. According to @lyrixx, the guard listeners must take any decisions based on the subject itself and not on the given contextual data (which are supposed to remain metadata). As a consequence, calling the `GuardEvent::getContext()` method always returned an empty context array so far.

To prevent this confusion any longer, the method has been deprecated in Symfony 6.4 (see #51484) and removed in this PR. The `GuardEvent` class hierarchy no longer has the `getContext` method while other event classes keep having it.

### DIFF
--- a/src/Symfony/Component/Workflow/CHANGELOG.md
+++ b/src/Symfony/Component/Workflow/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * Require explicit argument when calling `Definition::setInitialPlaces()`
+ * `GuardEvent::getContext()` method has been removed. Method was not supposed to be called within guard event listeners as it always returned an empty array anyway.
 
 6.4
 ---
--- a/src/Symfony/Component/Workflow/Event/HasContextTrait.php
+++ b/src/Symfony/Component/Workflow/Event/HasContextTrait.php
@@ -1,0 +1,29 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Workflow\Event;
+
+/**
+ * @author Fabien Potencier <fabien@symfony.com>
+ * @author Grégoire Pineau <lyrixx@lyrixx.info>
+ * @author Hugo Hamon <hugohamon@neuf.fr>
+ *
+ * @internal
+ */
+trait HasContextTrait
+{
+    private array $context = [];
+
+    public function getContext(): array
+    {
+        return $this->context;
+    }
+}
--- a/UPGRADE-7.0.md
+++ b/UPGRADE-7.0.md
@@ -490,6 +490,7 @@ Workflow
 --------
 
  * Require explicit argument when calling `Definition::setInitialPlaces()`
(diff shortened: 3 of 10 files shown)


src/Symfony/Component/Workflow/Event/HasContextTrait.php
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * @author Fabien Potencier <fabien@symfony.com>
+ * @author Grégoire Pineau <lyrixx@lyrixx.info>
+ * @author Hugo Hamon <hugohamon@neuf.fr>
+ *
+ * @internal
+ */
+trait HasContextTrait
+{
+    private array $context = [];
```
hhamon, line 23: I made the property `private` instead of `protected` as all concrete classes using this trait are already all `final`. Although it may be a potential BC break, it should not as event classes are final and internal. Application code should are not supposed to extend the base abstract `Event` class to produce new subtypes of workflow events.

src/Symfony/Component/Workflow/Event/HasContextTrait.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+namespace Symfony\Component\Workflow\Event;
+
+/**
+ * @author Fabien Potencier <fabien@symfony.com>
+ * @author Grégoire Pineau <lyrixx@lyrixx.info>
+ * @author Hugo Hamon <hugohamon@neuf.fr>
+ *
+ * @internal
```
hhamon, line 19: Trait is marked `internal` as it's not supposed to be used outside of the Workflow component itself.

src/Symfony/Component/Workflow/Event/HasContextTrait.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+trait HasContextTrait
```
hhamon, line 21: I'm not sure about the naming of this trait although it's purely made for an internal use. I was also thinking of naming it `HoldContextTrait`. Any other better suggestions are welcome and appreciated.

src/Symfony/Component/Workflow/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ CHANGELOG
 ---
 
  * Require explicit argument when calling `Definition::setInitialPlaces()`
+ * `GuardEvent::getContext()` method has been removed. Method was not supposed to be called within guard event listeners as it always returned an empty array anyway.
+ * Add internal `HasContextTrait` trait used by all workflow event classes except by `GuardEvent` class
```
lyrixx: I don't think it should be documented since everything is internal here

hhamon: I was unsure if it was worth it to add it to the changelog while it's purely internal. I've removed the line.

lyrixx (approved):
Thanks

fabpot:
Thank you @hhamon.

PR #35109, opened 2019-12-26
[FrameworkBundle] cache ClassMetadataFactory in debug

bastnic:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | no (performance)
| New feature?  | performance...
| Deprecations? | no
| Tickets       | #35096
| License       | MIT
| Doc PR        | 

Cf #35096, validator and serialization yaml/xml files are parsed at each requests. But they are already tracked by a `DirectoryResource` in 
https://github.com/symfony/symfony/blob/bd9dc7c573f537c5a751fc3543774cab9ad4770b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L1246-L1261

So it only needs a real cache purge on validator / serializer sytem cache, and I use the [CachePoolClearerCacheWarmer](https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/CacheWarmer/CachePoolClearerCacheWarmer.php) from APIP that already does similar things.

Handling that sort of cache in Symfony fixes a lot of others performance issue.

image

PR #51497, opened 2023-08-27
[FrameworkBundle] no serializer mapping cache in debug mode without enable_annotations

soyuka:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | #51497
| License       | MIT
| Doc PR        | not needed

There's no reason we should disable the cache only without `enable_annotations`, when working only with attributes, in debug mode the cache is enabled which is why we often need to clear cache when changing a serialized object to get the changes.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
@@ -1790,13 +1790,14 @@ class FrameworkExtension extends Extension
             $container->removeDefinition('serializer.normalizer.mime_message');
         }
 
+        if ($container->getParameter('kernel.debug')) {
+            $container->removeDefinition('serializer.mapping.cache_class_metadata_factory');
+        }
+
         $serializerLoaders = [];
         if (isset($config['enable_annotations']) && $config['enable_annotations']) {
             if (\PHP_VERSION_ID < 80000 && !$this->annotationsConfigEnabled) {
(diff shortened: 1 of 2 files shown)


dunglas (approved):

fabpot:
Please, keep the full header in the description.

soyuka:
done I guess you spoke about the commit description

derrabus (approved):

nicolas-grekas:
Thank you @soyuka.

bastnic:
The reason is that for yaml/xml, it's ban by CachePoolClearerCacheWarmer (from apip...) from this PR https://github.com/symfony/symfony/pull/35109/files.

With that PR, it means that all yaml/xml are parsed at each request without any change :(.

xabbuh:
@bastnic comments on merged PRs are likely to get lost. Please open a new issue if you think that you discovered an issue with this implementation and provide a small example application that allows to reproduce it.

nicolas-grekas:
@bastnic you were right, and this went unanswered for too long. Thanks for your patience.

Applications that map with files did lose their metadata cache in dev, and the `CachePoolClearerCacheWarmer` from #35109 was not enough to keep that cache fresh. Three things were in the way: the mapping files were not tracked as container resources, so editing one never rebuilt the container; the metadata was stored in `serialization.php`, written by an optional cache warmer that does not run when the container is rebuilt while handling a request; and the pool clearer itself was only registered together with the profiler.

So the staleness reported here was real, it just had another cause. #66137 fixes the three and brings the cache back in debug mode when the metadata comes from mapping files. On an application with 83 mapping files, the first serialization of a request goes from about 140 ms to about 1 ms.

PR #51502, opened 2023-08-27
[HttpKernel] Add temporary URI signed

### LINKED ISSUE #51501: [HttpFoundation] Temporary URI signed
### Description

I would like to add the possibility to create a temporary URI signed.

To achieve this, we can add an expiration parameter to the `UriSigner`'s `sign `method.
By default at null, if a value is passed, a query parameter with the timestamp of the expiration date + time will be added before the computation of the URI hash.
Thus when checking the URI, if this query parameter is present, the timestamp will be used by the `check `method to determine if the URI has expired.

I propose a signature like this :

```php
public function sign(string $uri, \DateTimeInterface|\DateInterval|int $expiration = null): string
```

- If it's a` \DatetimeInterface`, it's converted to a timestamp.
- If it's a `\DateInterval`, it's added to "now" then converted to a timestamp.
- If it's an `int`, it's expected to be a timestamp in seconds.

If null, nothing is added.

### Example

```php

use Symfony\Component\HttpKernel\UriSigner;

class DemoController extends AbstractController
{

    #[Route('/', name: 'index')]
    public function index(UriSigner $signer): Response
    {
        $signed = $signer->sign('demo?foo=bar', new \DateTime('2050-01-01')); // Valid for years..
        dump($signer->check($signed)); // true

        $signed = $signer->sign('demo?foo=bar', new \DateInterval('PT10S')); // Valid for 10 seconds from now
        dump($signer->check($signed)); // true
        sleep(30);
        dump($signer->check($signed)); // false
        
        $signed = $signer->sign('demo?foo=bar', 4070908800); // 2099-01-01
        dump($signer->check($signed)); // true
    }
    
}
```

BaptisteContreras:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #51501
| License       | MIT
| Doc PR        | TODO

This feature add the possibilty to create a temporary URI signed as mentionned in #51501.

Few things about this implementation :

- I added an optional parameter to the constructor of the `UriSigner`. I'm not sure if it can be considered as a **BC** ?
- The query parameter for expiration (by default "_expiration") is considered as a **reserved parameter** for the signer and if you try to sign an URI with this parameter, the method will throw 
a LogicalException.  In fact, the `check `method will test the presence of the expiration query parameter to determine if it must (or not) verify the expiration.
A problem can arise in this case : Lets say you want to sign this URI : "/demo?_expiration=foo" and you don't want a temporary URL. The URI is signed without problem but when you try to call `check`, the method will notice the
presence of "_expiration" and try to determine if it has expired, considering the value "foo" as a timestamp.

### DIFF
--- a/src/Symfony/Component/HttpFoundation/CHANGELOG.md
+++ b/src/Symfony/Component/HttpFoundation/CHANGELOG.md
@@ -4,6 +4,9 @@ CHANGELOG
 7.1
 ---
 
+ * Add optional `$expirationParameter` argument to `UriSigner::__construct()`
+ * Add optional `$expiration` argument to `UriSigner::sign()`
+ * Rename `$parameter` argument of `UriSigner::__construct()` to `$hashParameter`
  * Add `UploadedFile::getClientOriginalPath()`
 
 7.0
--- a/src/Symfony/Component/HttpFoundation/UriSigner.php
+++ b/src/Symfony/Component/HttpFoundation/UriSigner.php
@@ -11,25 +11,30 @@
 
 namespace Symfony\Component\HttpFoundation;
 
+use Symfony\Component\HttpFoundation\Exception\LogicException;
+
 /**
  * @author Fabien Potencier <fabien@symfony.com>
  */
 class UriSigner
 {
     private string $secret;
-    private string $parameter;
+    private string $hashParameter;
+    private string $expirationParameter;
 
     /**
-     * @param string $parameter Query string parameter to use
+     * @param string $hashParameter       Query string parameter to use
+     * @param string $expirationParameter Query string parameter to use for expiration
      */
-    public function __construct(#[\SensitiveParameter] string $secret, string $parameter = '_hash')
+    public function __construct(#[\SensitiveParameter] string $secret, string $hashParameter = '_hash', string $expirationParameter = '_expiration')
     {
         if (!$secret) {
             throw new \InvalidArgumentException('A non-empty secret is required.');
         }
 
         $this->secret = $secret;
-        $this->parameter = $parameter;
+        $this->hashParameter = $hashParameter;
+        $this->expirationParameter = $expirationParameter;
     }
 
     /**
@@ -37,8 +42,16 @@ class UriSigner
      *
      * The given URI is signed by adding the query string parameter
      * which value depends on the URI and the secret.
+     *
+     * @param \DateTimeInterface|\DateInterval|int|null $expiration The expiration for the given URI.
+     *                                                              If $expiration is a \DateTimeInterface, it's expected to be the exact date + time.
+     *                                                              If $expiration is a \DateInterval, the interval is added to "now" to get the date + time.
+     *                                                              If $expiration is an int, it's expected to be a timestamp in seconds of the exact date + time.
+     *                                                              If $expiration is null, no expiration.
+     *
+     * The expiration is added as a query string parameter.
      */
-    public function sign(string $uri): string
+    public function sign(string $uri, \DateTimeInterface|\DateInterval|int|null $expiration = null): string
     {
         $url = parse_url($uri);
         $params = [];
@@ -47,14 +60,27 @@ class UriSigner
             parse_str($url['query'], $params);
         }
 
+        if (isset($params[$this->hashParameter])) {
+            throw new LogicException(sprintf('URI query parameter conflict: parameter name "%s" is reserved.', $this->hashParameter));
+        }
+
+        if (isset($params[$this->expirationParameter])) {
+            throw new LogicException(sprintf('URI query parameter conflict: parameter name "%s" is reserved.', $this->expirationParameter));
+        }
+
+        if (null !== $expiration) {
+            $params[$this->expirationParameter] = $this->getExpirationTime($expiration);
+        }
+
         $uri = $this->buildUrl($url, $params);
-        $params[$this->parameter] = $this->computeHash($uri);
+        $params[$this->hashParameter] = $this->computeHash($uri);
 
         return $this->buildUrl($url, $params);
     }
 
     /**
      * Checks that a URI contains the correct hash.
+     * Also checks if the URI has not expired (If you used expiration during signing).
      */
     public function check(string $uri): bool
     {
@@ -65,14 +91,22 @@ class UriSigner
             parse_str($url['query'], $params);
         }
 
-        if (empty($params[$this->parameter])) {
+        if (empty($params[$this->hashParameter])) {
             return false;
         }
 
-        $hash = $params[$this->parameter];
-        unset($params[$this->parameter]);
+        $hash = $params[$this->hashParameter];
+        unset($params[$this->hashParameter]);
 
-        return hash_equals($this->computeHash($this->buildUrl($url, $params)), $hash);
+        if (!hash_equals($this->computeHash($this->buildUrl($url, $params)), $hash)) {
+            return false;
+        }
+
+        if ($expiration = $params[$this->expirationParameter] ?? false) {
+            return time() < $expiration;
+        }
+
+        return true;
     }
 
     public function checkRequest(Request $request): bool
@@ -104,5 +138,18 @@ class UriSigner
         $fragment = isset($url['fragment']) ? '#'.$url['fragment'] : '';
 
         return $scheme.$user.$pass.$host.$port.$path.$query.$fragment;
+    }
+
+    private function getExpirationTime(\DateTimeInterface|\DateInterval|int $expiration): string
+    {
+        if ($expiration instanceof \DateTimeInterface) {
+            return $expiration->format('U');
+        }
+
+        if ($expiration instanceof \DateInterval) {
+            return \DateTimeImmutable::createFromFormat('U', time())->add($expiration)->format('U');
+        }
+
+        return (string) $expiration;
     }
 }
--- a/src/Symfony/Component/HttpFoundation/Exception/ExceptionInterface.php
+++ b/src/Symfony/Component/HttpFoundation/Exception/ExceptionInterface.php
@@ -1,0 +1,25 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\HttpFoundation\Exception;
+
+interface ExceptionInterface extends \Throwable
+{
+}
--- a/src/Symfony/Component/HttpFoundation/Exception/LogicException.php
+++ b/src/Symfony/Component/HttpFoundation/Exception/LogicException.php
@@ -1,0 +1,28 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\HttpFoundation\Exception;
+
+/**
+ * Base LogicException for Http Foundation component.
+ */
+class LogicException extends \LogicException implements ExceptionInterface
+{
+}
--- a/src/Symfony/Component/HttpFoundation/Tests/UriSignerTest.php
+++ b/src/Symfony/Component/HttpFoundation/Tests/UriSignerTest.php
@@ -12,6 +12,7 @@
 namespace Symfony\Component\HttpFoundation\Tests;
 
 use PHPUnit\Framework\TestCase;
+use Symfony\Component\HttpFoundation\Exception\LogicException;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\UriSigner;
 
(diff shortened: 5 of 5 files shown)


src/Symfony/Component/HttpKernel/UriSigner.php
```diff
@@ -106,4 +134,22 @@ private function buildUrl(array $url, array $params = []): string
...
+        if ($expiration instanceof \DateInterval) {
+            return (new \DateTimeImmutable())->add($expiration)->format('U');
+        }
+
+        return (string) $expiration;
+    }
+
+    private function isExpired(int $expiration): bool
+    {
+        return ((int) microtime(true)) >= $expiration;
```
stof: use `time()` rather than `microtime()` if you don't care about sub-second precision. It is better than doing a precision-losing cast IMO.

src/Symfony/Component/HttpKernel/Tests/UriSignerTest.php
```diff
@@ -81,6 +114,61 @@ public function testSignerWorksWithFragments()
...
+        );
+
+        $this->assertTrue($signer->check($signer->sign('http://example.com/foo?bar=foo&foo=bar#foobar', new \DateTime('2099-01-01 00:00:00'))));
+    }
+
+    public function testSignWithUriExpiration()
+    {
+        $signer = new UriSigner('foobar');
+
+        $this->assertSame($signer->sign('http://example.com/foo?foo=bar&bar=foo', new \DateTime('2099-01-01 00:00:00')), $signer->sign('http://example.com/foo?bar=foo&foo=bar', 4070908800));
```
stof: please use `new DateTimeImmutable` unless the code is specifically testing support for the mutable class (which is not the case here).

We try to avoid the usage of the mutable value object as much as we can in Symfony in favor of the immutable one.

stof (changes requested):

src/Symfony/Component/HttpKernel/UriSigner.php
BaptisteContreras: I totally agree !

BaptisteContreras:
PR updated :)

src/Symfony/Component/HttpKernel/CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ CHANGELOG
  * Add argument `$validationFailedStatusCode` to `#[MapQueryString]` and `#[MapRequestPayload]`
  * Add argument `$debug` to `Logger`
  * Add class `DebugLoggerConfigurator`
+ * Add optional `$expirationParameter` argument to `UriSigner::__construct()`
+ * Add optional `$expiration` argument to `UriSigner::sign()`
```
OskarStark, line 15:
```suggestion
 * Add optional `$expiration` and `$expirationParameter` argument to `UriSigner::__construct()`
```

src/Symfony/Component/HttpKernel/UriSigner.php
```diff
@@ -48,6 +59,14 @@ public function sign(string $uri): string
             parse_str($url['query'], $params);
         }
 
+        if (isset($params[$this->expirationParameter])) {
+            throw new \LogicException(sprintf('URI query parameter conflict. Parameter name "%s" is reserved.', $this->expirationParameter));
```
OskarStark: You should use the LogicExceptiion from HttpKernel component

src/Symfony/Component/HttpKernel/UriSigner.php
```diff
@@ -106,4 +134,22 @@ private function buildUrl(array $url, array $params = []): string
 
         return $scheme.$user.$pass.$host.$port.$path.$query.$fragment;
     }
+
+    private function getExpirationTime(\DateTimeInterface|\DateInterval|int $expiration): string
```
OskarStark: We should leverage `DatePoint` here

https://github.com/symfony/symfony/pull/51415

OskarStark:
Can you please rebase? Thanks

src/Symfony/Component/HttpKernel/UriSigner.php
BaptisteContreras: You want me to just replace `\DateTimeInterface` by `DatePoint` ?

OskarStark: Best way would be to add it to the union type, no?

BaptisteContreras: `DatePoint` extends `\DateTimeImmutable` so it is also a `\DateTimeInterface`. Should we really add it to the union type ?
IMO it is not necessary.

src/Symfony/Component/HttpKernel/UriSigner.php
BaptisteContreras: The class moved from **HttpKernel** to **HttpFoundation** and there is no `LogicException `in this component. Can we keep `\LogicException` ?

src/Symfony/Component/HttpKernel/UriSigner.php
stof: We should keep allowing any `\DateTimeInterface` here. There is no reason to restrict it to `DatePoint` (especially not in a component that has no hard dependency on symfony/clock)

src/Symfony/Component/HttpFoundation/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ CHANGELOG
  * Add `UriSigner` from the HttpKernel component
  * Add `partitioned` flag to `Cookie` (CHIPS Cookie)
  * Add argument `bool $flush = true` to `Response::send()`
+ * Add optional `$expirationParameter` argument to `UriSigner::__construct()`
+ * Add optional `$expiration` argument to `UriSigner::sign()`
```
nicolas-grekas: should be moved to a new 7.1 section now

src/Symfony/Component/HttpFoundation/UriSigner.php
```diff
@@ -71,7 +91,15 @@ public function check(string $uri): bool
         $hash = $params[$this->parameter];
         unset($params[$this->parameter]);
 
-        return hash_equals($this->computeHash($this->buildUrl($url, $params)), $hash);
+        if (!hash_equals($this->computeHash($this->buildUrl($url, $params)), $hash)) {
+            return false;
+        }
+
+        if ($expiration = $params[$this->expirationParameter] ?? false) {
+            return !$this->isExpired((int) $expiration);
```
nicolas-grekas: let's inline the logic, no need to wrap a comparison in a method

src/Symfony/Component/HttpFoundation/UriSigner.php
```diff
@@ -104,6 +132,24 @@ private function buildUrl(array $url, array $params = []): string
...
     }
+
+    private function getExpirationTime(\DateTimeInterface|\DateInterval|int $expiration): string
+    {
+        if ($expiration instanceof \DateTimeInterface) {
+            return $expiration->format('U');
+        }
+
+        if ($expiration instanceof \DateInterval) {
+            return (new \DateTimeImmutable())->add($expiration)->format('U');
```
nicolas-grekas:
```suggestion
            return (\DateTimeImmutable::createFromFormat('U', time()))->add($expiration)->format('U');
```

src/Symfony/Component/HttpFoundation/UriSigner.php
```diff
@@ -46,6 +57,14 @@ public function sign(string $uri): string
             parse_str($url['query'], $params);
         }
 
+        if (isset($params[$this->expirationParameter])) {
+            throw new \LogicException(sprintf('URI query parameter conflict. Parameter name "%s" is reserved.', $this->expirationParameter));
```
nicolas-grekas: we don't have this check for `$this->parameter`, but we need to be consistent: we check either none or both, but not a mix

nicolas-grekas (commented):
LGTM with some comments thanks

src/Symfony/Component/HttpFoundation/UriSigner.php
```diff
@@ -104,6 +132,24 @@ private function buildUrl(array $url, array $params = []): string
...
+        if ($expiration instanceof \DateInterval) {
+            return (new \DateTimeImmutable())->add($expiration)->format('U');
+        }
+
+        return (string) $expiration;
+    }
+
+    private function isExpired(int $expiration): bool
+    {
+        return time() >= $expiration;
```
norkunas: Maybe using Clock component optionally would be better with fallback, to allow using mock clock in tests for devs?

BaptisteContreras: It will create a dependency between HttpFoundation and the Clock component. How would you do it ?

src/Symfony/Component/HttpFoundation/UriSigner.php
BaptisteContreras: Done 👍 . I'm just curious, why should we use the static constructor method here ? To be more explicit and therefore 
 have more lisibility I guess ?

src/Symfony/Component/HttpFoundation/UriSigner.php
BaptisteContreras: I added the check for the "_hash" parameter (with two new tests to cover that).

src/Symfony/Component/HttpFoundation/UriSigner.php
norkunas: Require-dev? :)

BaptisteContreras: Mmmh I see, and in the `check` method I should get the current time like that ? 

```php
if (class_exists(\Symfony\Component\Clock\Clock::class)) {
            $time =  \Symfony\Component\Clock\Clock::get()->now()->format('U');
        } else {
            $time = time();
        }
```

norkunas: Probably. Or via DI, not sure what maintainers would prefer

nicolas-grekas: I wouldn't use the Clock component in this PR. Let's use time() for now and discuss for Clock applied to HttpFoundation separately

src/Symfony/Component/HttpFoundation/UriSigner.php
nicolas-grekas: to make time() mockable

src/Symfony/Component/HttpFoundation/UriSigner.php
```diff
@@ -47,6 +58,14 @@ public function sign(string $uri): string
             parse_str($url['query'], $params);
         }
 
+        if (isset($params[$this->parameter]) || isset($params[$this->expirationParameter])) {
+            throw new \LogicException(sprintf('URI query parameter conflict. Parameter name "%s" and "%s" are reserved.', $this->parameter, $this->expirationParameter));
```
nicolas-grekas:
```suggestion
            throw new \LogicException(sprintf('URI query parameter conflict: parameter name "%s" and "%s" are reserved.', $this->parameter, $this->expirationParameter));
```

src/Symfony/Component/HttpFoundation/UriSigner.php
```diff
@@ -72,7 +92,15 @@ public function check(string $uri): bool
         $hash = $params[$this->parameter];
         unset($params[$this->parameter]);
 
-        return hash_equals($this->computeHash($this->buildUrl($url, $params)), $hash);
+        if (!hash_equals($this->computeHash($this->buildUrl($url, $params)), $hash)) {
+            return false;
+        }
+
+        if ($expiration = $params[$this->expirationParameter] ?? false) {
+            return !(time() >= (int) $expiration);
```
nicolas-grekas: the cast to int is dangerous on 32-bits system, better let php handle this for us:
```suggestion
            return time() < $expiration;
```

src/Symfony/Component/HttpFoundation/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1
+---
+ * Add optional `$expirationParameter` argument to `UriSigner::__construct()`
```
nicolas-grekas, line 7: missing blank line before

BaptisteContreras:
@nicolas-grekas if this PR is OK for you, can we have this in 7.1 ?

src/Symfony/Component/HttpFoundation/UriSigner.php
```diff
@@ -18,27 +18,38 @@ class UriSigner
 {
     private string $secret;
     private string $parameter;
```
fabpot: Let's rename if to `hashParameter`

src/Symfony/Component/HttpFoundation/UriSigner.php
```diff
@@ -47,6 +58,14 @@ public function sign(string $uri): string
             parse_str($url['query'], $params);
         }
 
+        if (isset($params[$this->parameter]) || isset($params[$this->expirationParameter])) {
+            throw new \LogicException(sprintf('URI query parameter conflict: parameter name "%s" and "%s" are reserved.', $this->parameter, $this->expirationParameter));
```
fabpot: Let's give the one that conflicts in the error message.

src/Symfony/Component/HttpKernel/UriSigner.php
fabpot: Let's create `LogicException` in the component than.

fabpot (changes requested):

fabpot:
@BaptisteContreras Are you still interested in finishing this PR?

BaptisteContreras:
@fabpot , yes ! I'm on it :)

src/Symfony/Component/HttpFoundation/Exception/ExceptionInterface.php.orig
```diff
@@ -0,0 +1,25 @@
+<?php
```
fabpot: This file must be removed

src/Symfony/Component/HttpFoundation/Exception/ExceptionInterface.php.rej
```diff
@@ -0,0 +1,35 @@
+--- src/Symfony/Component/HttpFoundation/Exception/ExceptionInterface.php	2024-03-17 17:28:21.134766504 +0000
```
fabpot: This file must be removed

src/Symfony/Component/HttpFoundation/Exception/LogicException.php.orig
```diff
@@ -0,0 +1,28 @@
+<?php
```
fabpot: This file must be removed

src/Symfony/Component/HttpFoundation/Exception/LogicException.php.rej
```diff
@@ -0,0 +1,41 @@
+--- src/Symfony/Component/HttpFoundation/Exception/LogicException.php	2024-03-17 17:28:21.346773890 +0000
```
fabpot: This file must be removed

fabpot (approved):

fabpot:
Thank you @BaptisteContreras.

src/Symfony/Component/HttpFoundation/UriSigner.php
```diff
@@ -11,34 +11,47 @@
...
+     * @param \DateTimeInterface|\DateInterval|int|null $expiration The expiration for the given URI.
+     *                                                              If $expiration is a \DateTimeInterface, it's expected to be the exact date + time.
+     *                                                              If $expiration is a \DateInterval, the interval is added to "now" to get the date + time.
+     *                                                              If $expiration is an int, it's expected to be a timestamp in seconds of the exact date + time.
+     *                                                              If $expiration is null, no expiration.
+     *
+     * The expiration is added as a query string parameter.
      */
-    public function sign(string $uri): string
+    public function sign(string $uri, \DateTimeInterface|\DateInterval|int|null $expiration = null): string
```
nicolas-grekas, line 54: I missed that this is a BC break. Instead, we should comment the argument on the signature and read it with func_get_arg (see on 6.4 for similar new arguments).

BaptisteContreras: Is it a BC even if the default value does not change the old behavior of the method ?
Do you want another PR linked to a new issue to fix this ?

nicolas-grekas: It's a BC break for classes that extend this class. Yes, it'd be great to fix it, PR welcome :)

PR #51507, opened 2023-08-28
[Validator] Remove `final` modifier on `PasswordStrength` and `PasswordStrengthValidator`

### LINKED ISSUE #51439: make PasswordStrength extends able
### Description

I try to support it in Drupal contrib module "Field validation", Drupal core's has own plugin system of symfony constraint, to make symfony constraint works at Drupal, we need extend the constraint and add plugin system info.

I got fatal error:
Drupal\field_validation\Plugin\Validation\Constraint\PasswordStrengthConstraint cannot extend final class Symfony\Component\Validator\Constraints\PasswordStrength in E:\xampp\htdocs\drupal10\modules\field_validation\src\Plugin\Validation\Constraint\PasswordStrengthConstraint.php

Please remove "final", make it extends able.

### Example

_No response_

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #51439
| License       | MIT
| Doc PR        | -

More than allowing the author to do what is specified in the ticket, I see several benefits in making this change:

- I can see no validator declared `final` other than `WhenValidator` (this may be another subject but not the main one here). Such change would bring consistency to the code.
- Same for the constraint itself: none is declared final other than `PasswordStrength` at this time
- If we'd like to extend the scale of password strength but still use `PasswordStrength` (but also use its public constants), we're currently blocked as the constructor restricts the `minScore` to be between 1 and 4. These values are hard-coded and not-configurable.
- Removing `final` from the validator class `PasswordStrengthValidator` would greatly help to implement any `PasswordStrength` child validator. Indeed, we could extend this validator by passing a closure to evaluate the password strength with the extended strength scale.

Note that I kept `estimateStrength()` private instead of mutating it to protected, as I can't see any advantage of redefining it when a closure is allowed to be passed in the validator constructor.

### DIFF
--- a/src/Symfony/Component/Validator/CHANGELOG.md
+++ b/src/Symfony/Component/Validator/CHANGELOG.md
@@ -14,6 +14,7 @@ CHANGELOG
  * Deprecate `ValidatorBuilder::enableAnnotationMapping()`, use `ValidatorBuilder::enableAttributeMapping()` instead
  * Deprecate `ValidatorBuilder::disableAnnotationMapping()`, use `ValidatorBuilder::disableAttributeMapping()` instead
  * Deprecate `AnnotationLoader`, use `AttributeLoader` instead
+ * Remove `final` modifier on `PasswordStrength` and `PasswordStrengthValidator`
 
 6.3
 ---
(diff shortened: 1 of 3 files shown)


stof:
The closure returning a score is expected to return one of the `PasswordStrength::STRENGTH_*` constants. So there is no reason to have any other values for the min score.

To me, if you need a different thing, it looks like you should have a custom constraint independent from `PasswordStrength` (maybe your own custom validator could reuse a PasswordStrength as part of some delegated validation).

As is, this PR gets a -1 from me as I don't see a use case requiring inheritance.

> * Such change would bring consistency to the code.

The ones being final right now are actually the newer ones, where we started to be stricter (removing final is easy once we have a compelling use case to support inheritance, while adding it requires a major version). We haven't worked on making older constraints final because the benefits of it are not big enough to justify the migration cost for now (but most of our existing validators don't have useful inheritance-based extension points either as they keep implementation details private anyway).

henrikac:
@fabpot since this PR was closed should the issue #51439 be closed as well?

derrabus:
Done.

PR #48121, opened 2022-11-05
[Messenger] Do not return fallback senders when other senders were already found

wouterj:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | n/a

See https://github.com/wouterj/sf-reproducer/tree/messenger-star-routing for a reproducer of the current situation.

Currently, if you use the special `*` in the Messenger's routing configuration, this sender is always returned for all messages (also for messages explicitly configured to use another sender). According to the [documentation](https://symfony.com/doc/current/messenger.html#routing-messages-to-a-transport), the star should act as a fallback instead:

> You may use `'*'` as the message class. This will act as a default routing rule for any message not matched under `routing`. This is useful to ensure no message is handled synchronously by default.

PR #51508, opened 2023-08-28
[Messenger] Fix routing to multiple fallback transports

valtzu:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

The change made in https://github.com/symfony/symfony/pull/48121 means that it's not possible to route into multiple transport as fallback like

```yaml
framework:
  messenger:
    routing:
      My\Message\*: [async, audit]
```

F.e. the message `My\Message\ToBeSentToTwoSenders` would only reach `async`.

The comment in the code mentions 

> if other senders already matched **with previous types**, skip the senders bound to the fallback

While in reality how it works is

> if other senders already matched, skip the senders bound to the fallback

After the changes in this MR it works as mentioned in the original comment, meaning it does not skip for the same type, but routes to all transports for that type.

---

This is not a BC break because anyone who has more than one transport in a wildcard rule has never received anything matching the wildcard on other than the 1st transport – so only if you had a misconfiguration in the first place you're affected.

### DIFF
--- a/src/Symfony/Component/Messenger/Tests/Transport/Sender/SendersLocatorTest.php
+++ b/src/Symfony/Component/Messenger/Tests/Transport/Sender/SendersLocatorTest.php
@@ -56,17 +56,19 @@ class SendersLocatorTest extends TestCase
     {
         $firstSender = $this->createMock(SenderInterface::class);
         $secondSender = $this->createMock(SenderInterface::class);
+        $thirdSender = $this->createMock(SenderInterface::class);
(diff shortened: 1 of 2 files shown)


valtzu:
@wouterj Could you take a look at this when you have time, since you made the original PR?

nicolas-grekas:
Thank you @valtzu.

PR #51519, opened 2023-08-30
[Mailer] Add logger setter on mailer transport

parallels999:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #... 
| License       | MIT

This is usefull for testing, we need to change NullLogger of a existing instance

### DIFF
--- a/src/Symfony/Component/Mailer/Transport/AbstractTransport.php
+++ b/src/Symfony/Component/Mailer/Transport/AbstractTransport.php
@@ -115,6 +115,13 @@ abstract class AbstractTransport implements TransportInterface
         return array_map(fn (Address $a) => $a->toString(), $addresses);
     }
 
+    public function setLogger(LoggerInterface $logger): static
+    {
+        $this->logger = $logger;
+
+        return $this;
+    }
+
     protected function getLogger(): LoggerInterface
     {
         return $this->logger;


angeljqv (approved):

angeljqv:
For 5.4?

stloyd:
It's not a bugfix but a new feature so 6.4 is the correct choice.

angeljqv:
@stloyd  So must be public instead?

derrabus:
Can you elaborate why this change is useful? The logger can be set via the constructor. Why would you want to change it at runtime?

parallels999:
I'm using laravel, it is using NullLogger(_constructor_), i'm trying to make a full debug(_on demand with [php-debugbar](https://github.com/maximebf/php-debugbar)_) including mailer, but there is no way to add a `LoggerInterface` on the fly

stloyd:
There is no point in adding this when it can be replaced via DI.

I can not see how this setter would be useful more than a dedicated declaration for the test environment.

stloyd:
@parallels999 If Laravel is forcing null in the constructor, then the change should be done there to allow passing something different (I'm not aware how Laravel works/uses that code)

parallels999:
https://github.com/laravel/framework/blob/909ea24597b86cfc7a49ce41074280d0347bf6a3/src/Illuminate/Mail/MailManager.php#L174-L196
>```php
>$factory = new EsmtpTransportFactory;
>$transport = $factory->create(new Dsn(
>```

stloyd:
So as you can see there is no way to set a logger in Laravel while Symfony gives you this option.

IMHO the change should be done to Laravel, not to Symfony.

parallels999:
https://github.com/laravel/framework/blob/909ea24597b86cfc7a49ce41074280d0347bf6a3/src/Illuminate/Mail/Mailer.php#L582-L590

I need it from a third party package, from outside, after laravel booting, no laravel config access, laravel give me
`getSymfonyTransport()`, so i am trying `getSymfonyTransport()->setLogger('MyLoggerCollector')`

Even if Laravel changed this in the configuration, i would not be able to change the logger on demand.

derrabus:
> IMHO the change should be done to Laravel, not to Symfony.

I tend to agree here. Please let's try to change Laravel's integration for Symfony Mailer first. Laravel is open source just as Symfony is, so let's apply this change where it fits best.

> Even if Laravel changed this in the configuration, i would not be able to change the logger on demand.

I have already asked you why you would want to do that and you did not answer that question.

I'm a bit rusty on Laravel these days, but it still uses Monolog for logging under the hood, doesn't it? So, pass a Monolog logger to the transport. When you want to tap into the log messages, push a handler to that logger and pop it when you're done. You don't need to replace the logger ever.

parallels999:
> I have already asked you why you would want to do that and you did not answer that question.

I am trying [php-debugbar](https://github.com/maximebf/php-debugbar), so it changes some loggers on demand, i enable/disable php-debugbar on the fly, so i need change mail logger sometimes, not from config

[php-debugbar(DataCollector/MessagesCollector.php)](https://github.com/maximebf/php-debugbar/blob/17ec0485bd84d84a982dff5dcd8d0a224664f027/src/DebugBar/DataCollector/MessagesCollector.php#L20)
`Psr\Log\AbstractLogger`
```php
$messagesCollector = new \DebugBar\DataCollector\MessagesCollector('mails');
getSymfonyTransport()->setLogger($messagesCollector)
```

derrabus:
Would this work?

```php
use DebugBar\DataCollector\MessagesCollector;
use Monolog\Handler\PsrHandler;
use Monolog\Logger;

// In your bootstrap: Create a logger and your service.

$logger = new Logger();
$service = new SomeClassThatConsumesALogger($logger);

// Later: Use Monolog to tap into your logs whenever you like.

$messageCollector = new MessagesCollector();
$logger->pushHandler(
    new PsrHandler($messageCollector)
);

try {
    $service->doSomethingThatEmitsLogMessages();
} finally {
    $logger->popHandler();
}
```

I'm not familiar with that library, but it advertises integration with Monolog. So maybe there's a more elegant way than what I just did.

parallels999:
> Would this work?

Is that assuming laravel adds logger on transport creation? _because I'm sure they don't accept that change_
Also it is for a third party package, so `bootstrap`/`configs` are not available, _maybe from ServiceProvider?_
I assume that `SomeClassThatConsumesALogger/doSomethingThatEmitsLogMessages` must be Symfony/Mailer, So I still don't understand how to set the logger in the transport class on that example, It will help me to get the messages from my own services, for which I can do it without Monolog and without so much problem

parallels999:
Other components allow this, for example:
https://github.com/symfony/symfony/blob/2bac801f8a4cba02e8ecc840fa7bf512034f860a/src/Symfony/Component/Templating/Loader/Loader.php#L37
https://github.com/symfony/symfony/blob/2bac801f8a4cba02e8ecc840fa7bf512034f860a/src/Symfony/Component/HttpClient/CurlHttpClient.php#L81
https://github.com/symfony/symfony/blob/2bac801f8a4cba02e8ecc840fa7bf512034f860a/src/Symfony/Component/HttpClient/NoPrivateNetworkHttpClient.php#L80
https://github.com/symfony/symfony/blob/2bac801f8a4cba02e8ecc840fa7bf512034f860a/src/Symfony/Component/HttpClient/TraceableHttpClient.php#L92

derrabus:
> > Would this work?
> 
> Is that assuming laravel adds logger on transport creation?

Not particularly: this is a direct response to your statement that you need to be able to change the logger "on demand". The point I want to make with that piece of pseudo-code is that you should not need to. There are ways to redirect logs without cluttering `setLogger()` methods on each an every class.

Even if you don't use Monolog, you can inject this logger everywhere you want to make the logger replaceable at runtime:

```php
use Psr\Log\AbstractLogger;
use Psr\Log\LoggerAwareInterface;
use Psr\Log\LoggerAwareTrait;

class ProxyLogger extends AbstractLogger implements LoggerAwareInterface
{
    use LoggerAwareTrait;

    public function log(mixed $level, \Stringable|string $message, array $context = []): void
    {
        $this->logger?->log($level, $message, $context);
    }
}
```

> _because I'm sure they don't accept that change_

Have you tried?

They abstract away the constructor call from you. Would it be such a ridiculous request to leverage a constructor parameter? Also, if Laravel is using Monolog for logging, they should be able to configure a dedicated channel for the mailer, inject a logger for that channel into the transports and let you tap into that channel.

All I ask from you is ask Laravel to make that change before we talk about working around them not making that change.

> Other components allow this, for example:

I am aware, but this is a slippery slope: I really don't want to clutter `setLogger()` methods on every class that could possible log something. If we add a `setLogger()` to a class, I want to make sure we've explored all other options.

parallels999:
> Have you tried?

Here https://github.com/laravel/framework/issues/48317#issuecomment-1709988640, the same answer always, for almost everyone, same on PRs, nothing is reviewed there that does not come from known contributors

>you can inject this logger everywhere you want to make the logger replaceable at runtime

Yes, i know, same as monolog

>Also it is for a third party package, so bootstrap/configs are not available

As third party package, The idea is not to change the Laravel loggers, just capture them for a data collector, only in case of activating debugging, this third-party package should not require changing the operation/configs of the application, just debug it sometimes

ro0NL:
if anything it could be a wither method

angeljqv:
I don't think it affects anything having it, a matter of perception.

systemsolutionweb (approved):

PaolaRuby:
LGTM, but must return void
https://github.com/symfony/symfony/blob/7c833eea4a13fd8dffca99d5a8c9f4b8664154df/src/Symfony/Component/Templating/Loader/Loader.php#L35-L37

src/Symfony/Component/Mailer/Transport/AbstractTransport.php
```diff
@@ -38,7 +38,7 @@ abstract class AbstractTransport implements TransportInterface
     public function __construct(EventDispatcherInterface $dispatcher = null, LoggerInterface $logger = null)
     {
         $this->dispatcher = $dispatcher;
-        $this->logger = $logger ?? new NullLogger();
+        $this->setLogger($logger ?? new NullLogger());
```
stof: no need to use the setter here

parallels999: I just added it, psalm throws error without that, and it seems this PR is going to be closed

stof: what do you mean by "it throws error" ? The existing code `$this->logger = $logger ?? new NullLogger();` should be totally fine.

parallels999: Yes, it is fine, error is "`setLogger` method is never used"

parallels999: Yes, it is fine, error is "setLogger method is never used"
Would it be merge if I returned it to how it was initially?

stof: Don't care about this one. We should change our psalm config to disable those dead code detection rules.

The method being added is expected to be part of the public API, so we don't care that it is never used in Symfony itself.

parallels999: OK, thanks for reviewing

src/Symfony/Component/Mailer/Transport/AbstractTransport.php
```diff
@@ -115,6 +115,11 @@ protected function stringifyAddresses(array $addresses): array
         return array_map(fn (Address $a) => $a->toString(), $addresses);
     }
 
+    public function setLogger(LoggerInterface $logger): void
```
fabpot: It should return `$this` like done for other setters in this class.

fabpot (changes requested):
I'm still not convinced about the use case.

stloyd:
> If Laravel is forcing null in the constructor, then the change should be done there to allow passing something different (I'm not aware how Laravel works/uses that code)

@fabpot I still think that it's not a proper way to fix the "case" here, cause it adds "hack" instead of fixing the "problem" correctly: add a way on Laravel to set logger instead of forcing default `null` value. Especially since [the Laravel team is open to reviewing such change](https://github.com/laravel/framework/issues/48317#issuecomment-1710180956).

derrabus:
> > Have you tried?
> 
> Here [laravel/framework#48317 (comment)](https://github.com/laravel/framework/issues/48317#issuecomment-1709988640), the same answer always, for almost everyone,

Well, your initial question did sound like a support ticket, so that canned response was somewhat justified.

Please try to fix the mailer integration in Laravel. If they abstract the constructor call away from you, there should be a way to either provide a logger that is to be injected to give you a way to tap into a preconfigured Monolog channel for mailer transports.

> same on PRs, nothing is reviewed there that does not come from known contributors

For the sake of the argument, let's assume what you're saying is right: Should we accept PRs that work around other projects not accepting contributions?

fabpot:
I’m going to close here as there is no point in arguing more. Everything has been said already.

parallels999:
> Please try to fix the mailer integration in Laravel. If they abstract the constructor call away from you

Needed for "plug an play"
>**As third party package, The idea is not to change the Laravel logger**

fabpot:
I suppose that Laravel has a way to inject a different logger than the "global" one to a service.

parallels999:
**UPDATE:**

- https://github.com/laravel/framework/pull/48581#issuecomment-1742261235 PR on laravel added, same answer as always

PR #51525, opened 2023-08-31
[Messenger][Scheduler] Add AsCronTask & AsPeriodicTask attributes

### LINKED ISSUE #51432: [RFC][Scheduler] Simplify schedule usage
### Description

Currently, it takes a lot of effort to create a scheduled task.
- define schedule provider
- create a message class
- create a messenger handler. 

These steps look too redundant, if you need to create a couple of simple tasks such as cleaning archive dir, check users, etc.

### Proposal

To simplify using of the scheduled tasks, proposed add `AsPeriodicTask` and `AsCronTask` attributes. They can be added as class or method attributes to any symfony service or console command.

Additionally, it is need to allow the merging of scheduled tasks obtained from schedule providers with the same name.

```php
#[AsPeriodicTask('1 day')]

#[AsCronTask('35,45,55 */3 * * *', jitter: 60, arguments: ['level' => '500mb'])]

#[AsCronTask('@random 3600', schedule: 'default')]
```

### Example

```php
<?php declare(strict_types=1);

namespace App\Service;

#[AsCronTask('*/5 * * * *')]
class SyncAppWorker
{
    public function __invoke(array $arguments = []): void
    {
        // code
    }
}
```

valtzu:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #51432 
| License       | MIT
| Doc PR        | symfony/symfony-docs#... _// todo_

Simplify scheduler usage by allowing to declare an attribute `AsCronTask` / `AsPeriodicTask` on any registered & autoconfigured service.

Example usage:

```php
#[AsPeriodicTask(frequency: 5, jitter: 1, arguments: ['Hello from periodic trigger'])]
class BusinessLogic
{
    public function __invoke(string $message): void
    {
        echo "$message\n";
    }
}
```

```php
#[AsCronTask('* * * * *', arguments: 'hello -v')]
#[AsCommand('app:do-stuff')]
class DoStuffCommand extends Command
{
    // ...
}
```

```yaml
services:
  some_other_service:
    class: # ...
    tags:
      - name: scheduler.task
        trigger: cron
        expression: '0 9-17 * * *'
        method: 'someMethod'
        transports: [async]
```

`bin/console debug:schedule` output:

```bash
Scheduler
=========

default
-------

 ------------------------------------------- ---------------------------------------- --------------------------------- 
  Message                                     Trigger                                  Next Run                         
 ------------------------------------------- ---------------------------------------- --------------------------------- 
  @App\BusinessLogic                          every 5 seconds with 0-1 second jitter   Sat, 02 Sep 2023 10:55:36 +0000  
  app:do-stuff hello -v                       * * * * *                                Sat, 02 Sep 2023 13:56:00 +0300  
  @some_other_service::someMethod via async   0 9-17 * * *                             Sat, 02 Sep 2023 14:00:00 +0300  
 ------------------------------------------- ---------------------------------------- --------------------------------- 

```

And then run `bin/console messenger:consume scheduler_default` to run the scheduler, like the usual.

---

**To-do (help needed):**
1. tests
2. docs
3. validate the approach (creating `RecurringMessage`s using DI)

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/Functional/app/Scheduler/config.yml
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/Functional/app/Scheduler/config.yml
@@ -10,6 +10,9 @@ services:
     Symfony\Bundle\FrameworkBundle\Tests\Fixtures\Messenger\DummySchedule:
         autoconfigure: true
 
+    Symfony\Bundle\FrameworkBundle\Tests\Fixtures\Messenger\DummyTask:
+        autoconfigure: true
+
     clock:
         synthetic: true
 
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/UnusedTagsPass.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/UnusedTagsPass.php
@@ -84,6 +84,7 @@ class UnusedTagsPass implements CompilerPassInterface
         'routing.loader',
         'routing.route_loader',
         'scheduler.schedule_provider',
+        'scheduler.task',
         'security.authenticator.login_linker',
         'security.expression_language_provider',
         'security.remember_me_handler',
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
@@ -145,6 +145,8 @@ use Symfony\Component\RateLimiter\Storage\CacheStorage;
 use Symfony\Component\RemoteEvent\Attribute\AsRemoteEventConsumer;
 use Symfony\Component\RemoteEvent\RemoteEvent;
 use Symfony\Component\Routing\Loader\AnnotationClassLoader;
+use Symfony\Component\Scheduler\Attribute\AsCronTask;
+use Symfony\Component\Scheduler\Attribute\AsPeriodicTask;
 use Symfony\Component\Scheduler\Attribute\AsSchedule;
 use Symfony\Component\Scheduler\Messenger\SchedulerTransportFactory;
 use Symfony\Component\Security\Core\AuthenticationEvents;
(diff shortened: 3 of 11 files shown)


valtzu:
What could also be a cool feature here is add `transports` property to both attributes/tags, then if it is set, wrap `ServiceCallMessage` in [`RedispatchMessage`](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Messenger/Message/RedispatchMessage.php).

kbond:
This looks nice! I'll give a proper review and help finish up for 6.4 when I have some time.

What are you thinking for commands? I was thinking:

```php
#[AsCronTask(expression: '* * * * *', arguments: 'arg --option')]
class MyCommand extends Command
{
}

**Comment by valtzu** [comment]
> What are you thinking for commands?

I was hoping we don't need such, as business logic usually is wrapped in a separate class and commands just act as an entry point, just like controllers. But I think this is only my reality.

About the attribute syntax for commands, `ArrayInput` syntax was my first thought, but I see `RunCommandMessage` uses string input so probably your suggestion makes the most sense if the approach would be to produce those instead of `ServiceCallMessage`. Also `StringInput` syntax is less verbose and since the whole point here is to make thing easier,  that's the way to go then.

**Comment by kbond** [comment]
Looking good! Glad you got commands working - using `RunCommandMessage` is exactly what I was thinking.

**Comment by valtzu** [comment]
~~I didn't realize it's possible to use `RedispatchMessage` without explicitly defining the transports (and rely on `routing` configuration), we need to add support for that also. Probably rename `transports` to `redispatch` and allow boolean.~~

Hmm, ok it does not seem possible after all, if transportNames are not defined, then the message is not dispatched to any transport – unlike the Messenger docs are stating.

**Review by fabpot — APPROVED** [review]

**Comment by fabpot** [comment]
Thank you @valtzu.

File: src/Symfony/Bundle/FrameworkBundle/Tests/Functional/app/Scheduler/config.yml [hunk 1]
```diff
@@ -10,6 +10,9 @@ services:
     Symfony\Bundle\FrameworkBundle\Tests\Fixtures\Messenger\DummySchedule:
         autoconfigure: true
 
+    Symfony\Bundle\FrameworkBundle\Tests\Fixtures\Messenger\DummyTask:
+        autoconfigure: true
```
fabpot, line 14: There are no tests associated with this, can you add some @valtzu?
I've moved these to a new schedule to avoid breaking the existing tests (https://github.com/symfony/symfony/pull/51802).

valtzu: @fabpot I'm not too familiar with the framework tests yet, and I'm also a bit busy now. I would need a week or two, if that's ok then I can try.

fabpot: Sure, there is no rush. Ping me if you need help.

faizanakram99:
~~@valtzu @kbond iiuc, this doesn't support stateful schedules (equivalent to `$schedule->stateful($cache)`), right?~~

nvm, it will only add tasks to given schedule (and schedule can be optionally be stateful)

PR #51528, opened 2023-08-31
[PropertyInfo] Reset inner array cache to reduce memory usage

rmikalkenas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

For projects, where long running processes are used, `PropertyInfoCacheExtractor` slowly leaking memory due to inner array cache. 
Implemented `ResetInterface` to clear array cache when services resetter is called

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/property_info.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/property_info.php
@@ -38,6 +38,7 @@ return static function (ContainerConfigurator $container) {
         ->set('property_info.cache', PropertyInfoCacheExtractor::class)
             ->decorate('property_info')
             ->args([service('property_info.cache.inner'), service('cache.property_info')])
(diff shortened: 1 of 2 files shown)


rmikalkenas:
Could someone help with why tests fail with missing method exception even though I added it? :thinking:

dunglas:
Same comment as on API Platform:

> I wonder if it's a good idea to do that. This cache is especially useful when using a long-running server such as RoadRunner or FrankenPHP in worker mode: on e populated, it will be reused for subsequent HTTP requests.

> I'm curious of why the memory leaks. Once all metadata are in cache, no new memory should be allocated.

> Maybe could we provide a method to manually clear the cache for advanced use cases (dynamic resource types for instance), but IMHO we shouldn't clear it by default.

rmikalkenas:
Closing as resolved differently

PR #51533, opened 2023-09-01
[ErrorHandler] [POC] Blame error

maidmaid:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

As a developer or devops, when an error is thrown on my development machine or in my production monitoring tool, one of the very first thing to do is to open the files listed in the stack trace error to get more context : who has written this line of code ? How recent is this code ? What is the reason of theses changes ?

Knowing immediately who is author, along with the date and reason of its changes could bring all the required context to understand faster the situation. Moreover, if the author is "blamed" is the stack trace error, it let the door open to automate a Slack ping to the blamed author. Some nice DX imo.

Below, a printscreen of the blamed stack trace in the exception page:

image

Below, the enhanced data added in the stack trace that could be used in the production logs, then to ping the author thanks to a monitoring/alerting tool like Datadog:

image

It's a poc. If I get some interest, I can keep going the work.

### DIFF
--- a/src/Symfony/Component/ErrorHandler/ErrorEnhancer/BlamedErrorEnhancer.php
+++ b/src/Symfony/Component/ErrorHandler/ErrorEnhancer/BlamedErrorEnhancer.php
@@ -1,0 +1,101 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\ErrorHandler\ErrorEnhancer;
+
+use Symfony\Component\Process\Exception\ExceptionInterface;
+use Symfony\Component\Process\Process;
+
+/**
+ * @author Dany Maillard <danymaillard93b@gmail.com>
+ */
+class BlamedErrorEnhancer implements ErrorEnhancerInterface
+{
+    private ?bool $isEnabled = null;
+
+    public function enhance(\Throwable $error): ?\Throwable
+    {
+        if (!$this->isEnabled()) {
+            return null;
+        }
+
+        $blamedTrace = array_map($this->blameStep(...), $error->getTrace());
+
+        (new \ReflectionProperty(\Error::class, 'trace'))->setValue($error, $blamedTrace);
+
+        return $error;
+    }
+
+    private function blameStep(array $step): array
+    {
+        ['file' => $file, 'line' => $line] = $step;
+
+        return $file && $line ? $step + $this->blame($file, $line) : $step;
+    }
+
+    private function blame(string $file, int $line): array
+    {
+        $process = Process::fromShellCommandline(sprintf('git blame %s -L %s,+1 -p', $file, $line));
+        $process->run();
+
+        if (!$process->isSuccessful()) {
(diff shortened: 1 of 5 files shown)


derrabus:
This is certainly an interesting experiment. But something about this idea of pinging developers because their name happens to appear somewhere in the stack trace of an exception feels very wrong to me. I don't think that this feature belongs into the Symfony core.

src/Symfony/Component/ErrorHandler/ErrorEnhancer/BlamedErrorEnhancer.php
```diff
@@ -0,0 +1,101 @@
...
+        if (null !== $this->isEnabled) {
+            return $this->isEnabled;
+        }
+
+        if (!class_exists(Process::class)) {
+            return $this->isEnabled = false;
+        }
+
+        try {
+            Process::fromShellCommandline('type git')->mustRun();
```
stof, line 94: `type` is not cross-platform

stof:
I agree with @derrabus here. Having this in the core looks wrong to me (especially if you also try to run the blaming on vendors, as you do right now).

maidmaid:
> pinging developers because their name happens to appear somewhere in the stack trace of an exception feels very wrong to me

In my job, we have already a similar workflow but in a manual way: we get alerts in Slack from Datadog listing the production, then a random developer investigates on the bug and usually pings the author(s) of the changes to fix it together. Generally, the best candidat to fix it is very likely the most recent author of the stack trace of the maintained codebase. If the alert message was directly able to ping author(s), it could be interesting in our situation. It's a little bit like when [Github pings the code owner](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) when we touch its files in a PR. But it's not really the point here, it was just for the example.

> especially if you also try to run the blaming on vendors

Of course, it's something improvable. Blaming only in `src` should be much reliable. The enhancer should know where the real sources of the project are, to avoid blaming irrelevant paths.

src/Symfony/Component/ErrorHandler/ErrorEnhancer/BlamedErrorEnhancer.php
maidmaid: What about `git blame -h` ?

derrabus:
> then a random developer investigates on the bug and usually pings the author(s) of the changes to fix it together.

And that workflow sounds perfectly fine to me.

> Generally, the best candidat to fix it is very likely the most recent author of the stack trace of the maintained codebase.

Let's assume I get a stack trace that is 20 calls deep. Which of those 20 lines is the correct one to blame for this ping? The line that raised the exception? The line that calls the method that raised the exception? Maybe the line that caused the exception isn't even in the stack trace.

What if the exception has been upcast twice while bubbling up? Do I investigate the stack trace of the exception that reached the error handler or do I traverse to the innermost exception?

Your "random" human developer is certainly able answer these questions on a case-by-case basis. But automatically picking the "right" line out of the stack trace is a hard problem. And I doubt that we find a generic solution to that.

fabpot:
Let's close as I don't think it belongs to Symfony core. I hope you understand.

PR #51538, opened 2023-09-02
[HttpFoundation] Support root-level Generator in StreamedJsonResponse

Jeroeny:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

Currently the `StreamedJsonResponse` only supports streaming nested Generators within an array data structure. 
However if a response is a list of items (for example database entities) on the root level, this isn't usable. 
I think both usecases can be supported with the change in this PR.

The root level generator doesn't account for additional nested generators yet. I could add that by doing `is_array($item)` and the call the recursive placeholder logic. 

Link to first PR that introduced StreamedJsonResponse: https://github.com/symfony/symfony/pull/47709

~~Also something I noticed is I only got intermediate output, when adding a `flush()` call after each item has been echo'd (with a `sleep(1)` after each item to see it output the parts individually).~~ Edit: I see the class' PhpDoc describes this and it's probably expected to be done in userland implementations.

### DIFF
--- a/src/Symfony/Component/HttpFoundation/CHANGELOG.md
+++ b/src/Symfony/Component/HttpFoundation/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * Make `HeaderBag::getDate()`, `Response::getDate()`, `getExpires()` and `getLastModified()` return a `DateTimeImmutable`
+ * Support root-level `Generator` in `StreamedJsonResponse`
 
 6.3
 ---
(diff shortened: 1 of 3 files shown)


derrabus:
Thank you. Please be aware that we never merge new features without any tests.

Jeroeny:
@derrabus Thanks for the hint, added tests.

nicolas-grekas (commented):
Thanks for the PR. Please add a line about this to the changelog of the component.

Jeroeny:
Added the changelog entry

nicolas-grekas (approved):

fabpot:
Thank you @Jeroeny.

PR #51542, opened 2023-09-03
[Scheduler] Trigger unique messages at runtime

Jeroeny:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

The messages in a schedule (`RecurringMessage`) are created at compile time, per Trigger. This means a trigger will always give the same message. This limits the usefulness of the component, because messages can often be dynamic. E.g. with a specific payload from the database or with a Uuid per message (often the case with architectures like EventSourcing / cqrs). 
Also there is no way for a trigger to only 'check' what the getNextRunDate is, without always triggering a message.

This PR makes it possible to yield zero or more messages when a trigger is triggered. Due to the trigger being a method, the messages can be created at runtime, with a unique payload (or directly serving messages from a DB table). 
Since the component is still experimental in 6.3, it's probably good to get something like this in before 6.4 is released.


src/Symfony/Component/Scheduler/Trigger/MessageProvider.php
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Scheduler\Trigger;
+
+use Symfony\Component\Scheduler\Generator\MessageContext;
+
+interface MessageProvider
```
fabpot:
```suggestion
interface MessageProviderInterface
```

src/Symfony/Component/Scheduler/Trigger/CallbackMessageProvider.php
```diff
@@ -0,0 +1,37 @@
...
+        $this->callback = $callback(...);
+        $this->description = $description ?? spl_object_hash($this->callback);
+    }
+
+    public function getMessages(MessageContext $context): iterable
+    {
+        return ($this->callback)($context);
+    }
+
+    public function __toString()
```
fabpot:
```suggestion
    public function __toString(): string
```

src/Symfony/Component/Scheduler/RecurringMessage.php
```diff
@@ -42,30 +45,40 @@ private function __construct(
      */
     public static function every(string|int|\DateInterval $frequency, object $message, string|\DateTimeImmutable $from = new \DateTimeImmutable(), string|\DateTimeImmutable $until = new \DateTimeImmutable('3000-01-01')): self
```
fabpot: Would it make sense to type hint `MessageProviderInterface|object` or maybe add a `@param` doc that explains the two possibilities? Or maybe both :)

src/Symfony/Component/Scheduler/Trigger/CallbackMessageProvider.php
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+namespace Symfony\Component\Scheduler\Trigger;
+
+use Symfony\Component\Scheduler\Generator\MessageContext;
+
+final class CallbackMessageProvider implements MessageProvider, \Stringable
+{
+    private \Closure $callback;
+
```
fabpot, line 19: Blank line can be removed

src/Symfony/Component/Scheduler/RecurringMessage.php
```diff
@@ -42,30 +45,40 @@ private function __construct(
...
 
     public static function trigger(TriggerInterface $trigger, object $message): self
     {
-        return new self($trigger, $message);
+        if ($message instanceof MessageProvider) {
+            return new self($trigger, $message);
+        }
+
+        try {
+            $description = $message instanceof \Stringable ? (string) $message : serialize($message);
```
fabpot, line 79:
```suggestion
            $description = $message instanceof \Stringable ? (string) $message : '';
```
and remove the try/catch.

src/Symfony/Component/Scheduler/Trigger/CallbackMessageProvider.php
```diff
@@ -0,0 +1,37 @@
...
+    private ?string $description;
+
+    public function __construct(callable $callback, string $description = null)
+    {
+        $this->callback = $callback(...);
+        $this->description = $description ?? spl_object_hash($this->callback);
```
fabpot:
```suggestion
        $this->description = $description;
```

src/Symfony/Component/Scheduler/Trigger/CallbackMessageProvider.php
```diff
@@ -0,0 +1,37 @@
...
+
+use Symfony\Component\Scheduler\Generator\MessageContext;
+
+final class CallbackMessageProvider implements MessageProvider, \Stringable
+{
+    private \Closure $callback;
+
+    private ?string $description;
+
+    public function __construct(callable $callback, string $description = null)
```
fabpot:
```suggestion
    public function __construct(callable $callback, string $description = '')
```

fabpot (changes requested):

src/Symfony/Component/Scheduler/RecurringMessage.php
Jeroeny: That typehint triggers this error for me:
```
PHP Fatal error:  Type Symfony\Component\Scheduler\Trigger\MessageProviderInterface|object contains both object and a class type, which is redundant in \symfony\src\Symfony\Component\Scheduler\RecurringMessage.php on line 64
```

I did add it to the doc of `RecurringMessage::trigger()`

src/Symfony/Component/Scheduler/Tests/Generator/MessageGeneratorTest.php
```diff
@@ -262,3 +262,14 @@ private static function makeDateTime(string $time): \DateTimeImmutable
         return new \DateTimeImmutable('2020-02-20T'.$time, new \DateTimeZone('UTC'));
     }
 }
+
+class Message {
```
fabpot: `{` should be on its own line.

src/Symfony/Component/Scheduler/RecurringMessage.php
fabpot: ok, let's only add phpdocs then.

src/Symfony/Component/Scheduler/RecurringMessage.php
Jeroeny: I had to create a `Message` class in the tests since the Scheduler required their Id's to be unique. But I've applied the changes.

src/Symfony/Component/Scheduler/Trigger/CallbackMessageProvider.php
```diff
@@ -0,0 +1,36 @@
...
+namespace Symfony\Component\Scheduler\Trigger;
+
+use Symfony\Component\Scheduler\Generator\MessageContext;
+
+final class CallbackMessageProvider implements MessageProviderInterface, \Stringable
+{
+    private \Closure $callback;
+    private string $description;
+
+    public function __construct(callable $callback, string $description = '')
```
fabpot:
```suggestion
    public function __construct(callable $callback, private string $description = '')
```
So we can remove the relate line in the constructor.

src/Symfony/Component/Scheduler/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ CHANGELOG
  * Add `AbstractTriggerDecorator`
  * Make `ScheduledStamp` "send-able"
  * Add `ScheduledStamp` to `RedispatchMessage`
+ * Add `MessageProvider` to trigger unique messages at runtime
```
fabpot:
```suggestion
 * Add `MessageProviderInterface` to trigger unique messages at runtime
```

src/Symfony/Component/Scheduler/RecurringMessage.php
fabpot: Oh, my bad then, I forgot about this. So, let's revert to use `serialize` as before. Sorry for that.

src/Symfony/Component/Scheduler/RecurringMessage.php
Jeroeny: Does doc for just `::trigger()` suffice or would it also be good on `every()` and `cron()` ?

fabpot: I think we should do it eveywhere it's supported to make it clear.

src/Symfony/Component/Scheduler/Command/DebugCommand.php
```diff
@@ -106,20 +107,30 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
         $trigger = $recurringMessage->getTrigger();
 
+        $next = $trigger->getNextRunDate($date)?->format('r') ?? '-';
+        if ('-' === $next && !$all) {
+            return null;
+        }
+
+        $messages = $recurringMessage->getMessages(new MessageContext($name, $recurringMessage->getId(), $trigger, $date));
+        $message = null;
+        if (\is_array($messages) && 1 === \count($messages)) {
```
stof: What if this returns several messages ? It seems to ignore all except the first one, which looks like a confusing behavior for a debug command. It should represent the actual state of what will happen

src/Symfony/Component/Scheduler/Trigger/CallbackMessageProvider.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Symfony\Component\Scheduler\Trigger;
+
+use Symfony\Component\Scheduler\Generator\MessageContext;
+
+final class CallbackMessageProvider implements MessageProviderInterface, \Stringable
+{
+    private \Closure $callback;
+
+    public function __construct(callable $callback, private string $description = '')
```
stof: Please add `@param callable(MessageContext): iterable<object>` as phpdoc, to document the expected signature of the callable. This way, static analysis tools will be able to detect invalid usages.

src/Symfony/Component/Scheduler/RecurringMessage.php
```diff
@@ -37,35 +40,53 @@ private function __construct(
...
+        } catch (\Exception) {
+            $description = $message::class;
+        }
+
+        return new self($trigger, new CallbackMessageProvider(static fn (): array => [$message], $description));
```
stof: instead of creating a closure here, maybe we should have a `FixedMessageProvider` taking an array of objects, or something like that.

src/Symfony/Component/Scheduler/Trigger/CallbackMessageProvider.php
```diff
@@ -0,0 +1,34 @@
...
+    {
+        $this->callback = $callback(...);
+    }
+
+    public function getMessages(MessageContext $context): iterable
+    {
+        return ($this->callback)($context);
+    }
+
+    public function __toString(): string
```
stof: Why do we need the message provider to be stringable ? And if we actually need it, why isn't it enforced in the interface ?

stof: Also, if the goal is to provide a unique id, I would rather not call that `description` (and I would probably rather use a `getId` method)

src/Symfony/Component/Scheduler/Command/DebugCommand.php
Jeroeny: Tbh now that `getMessages()` may dynamically construct messages (e.g. using external sources for data), I think it shouldn't be called in this debug command. Perhaps we can replace it with `$recurringMessage->getProvider()::class` ?

src/Symfony/Component/Scheduler/Trigger/CallbackMessageProvider.php
Jeroeny: It is because of https://github.com/symfony/symfony/blob/efd41f476df0e9ec73804c8b482046cac8021f6f/src/Symfony/Component/Scheduler/RecurringMessage.php#L95

In which I just tried to keep the existing behaviour where possible, even though I don't fully understand the need for a duplication-check like this.

The `getId` method is fine by me.

src/Symfony/Component/Scheduler/Trigger/FixedMessageProvider.php
```diff
@@ -0,0 +1,34 @@
...
+use Symfony\Component\Scheduler\Generator\MessageContext;
+
+final class FixedMessageProvider implements MessageProviderInterface
+{
+    /**
+     * @param array<object> $messages
+     */
+    public function __construct(private array $messages, private string $id = '')
+    {
+    }
```
fabpot:
```suggestion
    public function __construct(
        private array $messages,
        private string $id = ''
    ) {
    }
```

src/Symfony/Component/Scheduler/Trigger/FixedMessageProvider.php
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Scheduler\Trigger;
+
+use Symfony\Component\Scheduler\Generator\MessageContext;
+
+final class FixedMessageProvider implements MessageProviderInterface
```
fabpot:
```suggestion
final class StaticMessageProvider implements MessageProviderInterface
```

fabpot (changes requested):
It looks good to me. I've made a few small suggestions.

src/Symfony/Component/Scheduler/Trigger/StaticMessageProvider.php
```diff
@@ -0,0 +1,36 @@
...
+use Symfony\Component\Scheduler\Generator\MessageContext;
+
+final class StaticMessageProvider implements MessageProviderInterface
+{
+    /**
+     * @param array<object> $messages
+     */
+    public function __construct(
+        private array $messages,
+        private string $id = ''
```
fabpot, line 23:
```suggestion
        private string $id = '',
```

src/Symfony/Component/Scheduler/Trigger/StaticMessageProvider.php
```diff
@@ -0,0 +1,36 @@
...
+    ) {
+    }
```
fabpot:
```suggestion
    public function __construct(
        public readonly array $messages,
        public readonly string $id = '',
    ) {
    }
```

and then, we can remove the 2 getters?

Jeroeny: Those methods are implementations of `MessageProviderInterface` (I realised after applying the suggestion..)

fabpot: Indeed, let's revert the last change and keep private properties. Sorry for the noise.

Jeroeny: No problem, done

fabpot (approved):

fabpot:
Thank you @Jeroeny.

Jeroeny:
Thanks!

PR #51543, opened 2023-09-04
[AssetMapper] Add support for CSS files in the importmap

### LINKED ISSUE #51417: [importmap] Wrong url for js modulepreload with site not at root of domain

6.3.3

### Description

Hi !

I've try to put somes CSS and JS files with the AssetMapper.
I've just installed the package with composer, and files are updated with asset() and importmap(), great.

CSS Files are OK and work well
JS Files can't be used atm, the generated URL is wrong and don't get the root directory of my project.
I've tryied severals options but can't get a correct URL after compilation.
Tryied to change base_url of framework : the only url changed are CSS with asset()

Example : 
```
#App on http//myserver/SymfES/

#CSS File genertaed with asset()
<link rel="stylesheet" href="[/SymfES/assets/styles/app-713d9828d7058036a44786836080554c.css]">

#JS File generated with importmap()
<link rel="modulepreload" href="[/assets/app-bbae9f648af22c8bdc294ce35f4a1e4c.js]">
#Should be : 
<link rel="modulepreload" href="[/SymfES/assets/app-bbae9f648af22c8bdc294ce35f4a1e4c.js]">
```

Actually, my configuration files are from scratch with no modification.

asset_mapper.yaml
```
framework:
    asset_mapper:
        paths:
            - assets/
```

importmap.php
```
return [
    'app' => [
        'path' => 'app.js',
        'preload' => true,
    ],
];
```

Did I miss something ?

Thanks

### How to reproduce

Just install AssetMapper from documentation, with a fresh SF6 install : 
`composer require symfony/asset-mapper symfony/asset symfony/twig-pack`

Apache with Alias configuration : 
```
# /etc/httpd/conf.d/SymfES.conf
Alias /SymfES /var/www/SymfES/public

<Directory "/var/www/SymfES/public/">
  AllowOverride All
  order deny,allow
  Allow from all
</Directory>


Options +FollowSymlinks
```

### Possible Solution

_No response_

### Additional Context

_No response_

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | fixes #51417 (fixed this along the way)
| New feature?  | yes
| Deprecations? | no
| Tickets       | None
| License       | MIT
| Doc PR        | TODO - but important!

Hi!

This is the brainchild of @nicolas-grekas & I (mostly him). This PR adds 2 big user-facing features:

## A) CSS Handling including ability to import CSS

The `importmap.php` now supports a `'type' => 'css'`:

```php
return [
    'app.css' => [
        'path' => 'styles/app.css',
        'type' => 'css',
    ],
```

This, by itself, won't cause the CSS file to be loaded. But it WILL be added to the importmap, though the exact behavior will depend on the entrypoint (see next section).

But, in the simplest case, it will output something like this, which adds the `<link rel="stylesheet">` tag to the page if this file is every imported in JS `import 'app.css'`.

```html
<script type="importmap">{"imports": {
    "app.css": "data:application/javascript,const%20d%3Ddocument%2Cl%3Dd.createElement%28%22link%22%29%3Bl.href%3D%22%2Fassets%2Fstyles%2Fapp-f86cc618674db9b39b179e440063aab4.css%22%2Cl.rel%3D%22stylesheet%22%2C%28d.head%7C%7Cd.getElementsByTagName%28%22head%22%29%5B0%5D%29.appendChild%28l%29"
}}</script>
```

More commonly, in the same way that AssetMapper finds relative JavaScript imports and adds them to the importmap, it also finds relative CSS imports and adds *those* to the importmap. This allows you to: `import './styles/foo.css'` without needing to add `foo.css` to the importmap. It "just works". This would result in `foo.css` being added to the page via JavaScript unless it is in the import chain of an entrypoint (see next section), in which case it would be a true `<link>` tag.

Also, you can now download CSS files via `importmap:require` (or, if a package comes with a CSS file, it will be downloaded and added automatically - i.e. if the package has a `style` key):

```bash
# will download `bootstrap` AND `bootstrap/dist/css/bootstrap.min.css`

php bin/console importmap:require bootstrap
```

## B) Auto-preload based on entrypoints

Like with Webpack, there is now a concept of "entrypoints". The ONE time you call `{{ importmap() }}`  on the page, you will pass the 1 or many "entrypoint" names from `importmap.php` that should be loaded - e.g. `importmap('app')` (the most common) or `importmap(['app', 'checkout'])`. 

Most simply (and this is true already in AssetMapper), this causes a `<script type="module">import 'app';</script>` to be rendered into the page. 

But in this PR, it also has another critical role:

* For each entrypoint, all of the non-lazy JavaScript dependencies are found. So if `app.js` imports `other.js` imports `yet-another.js` imports `some_styles.css`, using non-lazy imports (i.e. `import './other.js` vs the lazy `import('./other.js')`), then `other.js`, `yet-another.js` and `some_styles.css` are all returned.
* For all of these dependencies, they are "preloaded"
  * `other.js` -> preloaded (i.e. `modulepreload` tag rendered)
  * `yet-another.js` -> preloaded (i.e. `modulepreload` tag rendered)
  * `some_styles.css` "preloaded" - i.e. a `<link rel="stylesheet"> is added to the page.

The idea is that, if `app.js` is your entrypoint, then *every* non-lazy import in its import chain will need to be downloaded before the JavaScript starts executing. So all files should be preloaded. Additionally, if we find any CSS that is imported in a non-lazy way, we render those as `link` tags.

The `preload` option in `importmap.php` is GONE. Preloading is controlled entirely through the entrypoint.

This entrypoint logic also affects the ordering of the non-lazy CSS files (i.e. the CSS files that will be rendered as `link` tags). It finds all (in order) non-lazy imported CSS files from the entrypoint and render them as `link` tags in order (like Webpack).

I propose the recipe starting `importmap.php` is updated to be:

```php
return [
    'app' => [
        'path' => 'app.js',
        'entrypoint' => true,
    ],
];
```

And then in `assets/app.js`:

```js
import './styles/app.css';
```

## C) Other Improvements

* Fixed #51417 where deploying to a subdirectory didn't work
* Added a new event in `asset-map:compile` so other processes can hook into this (will make bundles like TailwindBundle) nicer.
* Removed `importmap:export` command: I don't see a need for this
* Added basic, conservative checking for commented-out imports - if lines START with `//` or `/*`

```
// both will correctly be ignored
// import 'foo';
/* import 'foo'; */
```

### TODOs
* [ ] Update the 6.4 docs
* [ ] update the 6.4 recipe


src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -168,15 +176,18 @@ private function buildImportMapJson(): void
         $dumpedModulePreloadPath = $this->assetsPathResolver->getPublicFilesystemPath().'/'.self::IMPORT_MAP_PRELOAD_FILE_NAME;
         if (is_file($dumpedImportMapPath) && is_file($dumpedModulePreloadPath)) {
             $this->json = file_get_contents($dumpedImportMapPath);
-            $this->modulesToPreload = json_decode(file_get_contents($dumpedModulePreloadPath), true, 512, \JSON_THROW_ON_ERROR);
+            $preload = json_decode(file_get_contents($dumpedModulePreloadPath), true, 512, \JSON_THROW_ON_ERROR);
+            $this->linkTags = $preload['linkTags'] ?? [];
+            $this->modulesToPreload = $preload['modules'] ?? [];
```
stof: this should check for dumped files being in the old format to handle the upgrade gracefully.

It can either convert that content to be the `modules` key (assuming we don't miss some info) or it could do as if the files were not there.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapEntry.php
```diff
@@ -27,6 +30,12 @@ public function __construct(
         public readonly ?string $url = null,
         public readonly bool $isDownloaded = false,
         public readonly bool $preload = false,
+        public readonly string $type = self::TYPE_JS,
     ) {
     }
+
+    public static function getValidTypes(): array
+    {
+        return [self::TYPE_JS, self::TYPE_CSS];
```
stof: This actually looks like a good use case for a BackedEnum.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
weaverryan: Yea, good catch. With the current code, if you have the old format, you'll end up with an empty `modulesToPreload` property. So, basically, the old format will be ignored.

We could do something like `$this->modulesToPreload = $preload['modules'] ?? $preload` to catch the old format. But because the `assetmap:compile` command is meant to run on deploy, I don't think we need to. The user would upgrade `symfony/asset-mapper` locally. Then, when they deploy that, they'll re-run the compile command and get the new format immediately.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapEntry.php
weaverryan: It's a nice idea, but there's a hiccup. When we dump the `importmap.php` file, we use `var_export()`. It will dump the full class name inline, which is quite ugly:

```
'file3.css' => [
    'path' => 'file3.css',
    'type' => \Symfony\Component\AssetMapper\ImportMap\ImportMapType::CSS,
    'preload' => true,
],
```

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
stof: this could cause issues if they ran the compile command locally.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapEntry.php
stof: the dumped file could be using the backing value, using `ImportMapType::from()` in the place where you currently use this `getValidTypes` to validate the content of the file being loaded. the dumped file does not contain exported ImportMapEntry objects.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
weaverryan: If you ran the compile command before upgrading, then upgraded, locally, the result is that the dumped file would effectively be ignored. That seems ok to me, right?

stof: don't we have a broken behavior if we load the compiled json without loading the compiled preload info ?

weaverryan: In a minor way only: you'd just be missing the "module preloads". Everything would still work, you'd just have the performance hit of not having the preloads. And this, in practice, would only affect local dev (since the compile command should be run on deploy). That's why I'm thinking it's minor enough (+ the experimental on the component) to not introduce the code. But we can.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapEntry.php
weaverryan: Done!

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -427,28 +458,55 @@ private function convertEntriesToImports(array $entries): array
...
             }
+            $imports[$entryOptions->importName] = $path;
 
             $dependencyImportMapEntries = array_map(function (AssetDependency $dependency) use ($entryOptions) {
                 return new ImportMapEntry(
                     $dependency->asset->publicPathWithoutDigest,
                     $dependency->asset->logicalPath,
+                    // if parent is preload & dependency is not lazy, then preload
                     preload: $entryOptions->preload && !$dependency->isLazy,
+                    type: 'css' === $dependency->asset->publicExtension ? ImportMapType::CSS : ImportMapType::JS,
```
nicolas-grekas: ImportMapType::from?

src/Symfony/Component/AssetMapper/Tests/ImportMap/Resolver/JsDelivrEsmResolverTest.php
```diff
@@ -233,6 +269,84 @@ public static function provideResolvePackagesTests(): iterable
...
+                ],
+            ],
+            'expectedResolvedPackages' => [
+                'bootstrap/dist/modal.js' => [
+                    'url' => 'https://cdn.jsdelivr.net/npm/bootstrap.js@5.2.0/dist/modal.js+esm',
+                ],
+            ],
+        ];
+
+        // TODO: test grabbing a specific file from JS skips style key
```
nicolas-grekas: TODO :)

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -380,13 +404,14 @@ private function writeImportMapConfig(array $entries): void
...
+         * Returns the importmap for this application.
          *
          * - "path" is a path inside the asset mapper system. Use the
          *     "debug:asset-map" command to see the full list of paths.
          *
-         * - "preload" set to true for any modules that are loaded on the initial
-         *     page load to help the browser download them earlier.
+         * - "preload"
+         *     For JavaScript: adds a modulepreload link to help the browser download it earlier.
+         *     For CSS: adds a <link rel="stylesheet" tag for the CSS.
```
nicolas-grekas: `For CSS: adds a <link rel="stylesheet"> tag for the CSS.` (missing`>`)

src/Symfony/Bridge/Twig/Extension/ImportMapExtension.php
```diff
@@ -23,6 +23,7 @@ public function getFunctions(): array
     {
         return [
             new TwigFunction('importmap', [ImportMapRuntime::class, 'importmap'], ['is_safe' => ['html']]),
+            new TwigFunction('importmap_link_tags', [ImportMapRuntime::class, 'linkTags'], ['is_safe' => ['html']]),
```
nicolas-grekas: what about making the `importmap()` helper also output those link tags and get rid of the new one?

nicolas-grekas (commented):
I played with this, it just works :)

When using `--download`, this will break assets referenced by `url()` / `@import` (except `data:` ones of course). To fix this, we'd need to parse those and fetch them, recursively.

Also I'm wondering: currently, the importmap recipe adds a link tag. Could we leverage this new feat to not do this in the recipe anymore? Would be nice :)

weaverryan:
> When using --download, this will break assets referenced by url() / @import (except data: ones of course). To fix this, we'd need to parse those and fetch them, recursively.

I'll check into that - we already do something similar for CSS.

> Also I'm wondering: currently, the importmap recipe adds a link tag. Could we leverage this new feat to not do this in the recipe anymore? Would be nice :)

Right now, I believe that `link` tag will be removed. Instead, we'll either start the user's `importmap.php` with an `app.css` entry. Or (what I'm leaning to now) is exactly like Encore: in `assets/app.js`, you'd `import './styles/app.css'`. That would trigger the link tag.

stof:
> Right now, I believe that `link` tag will be removed. Instead, we'll either start the user's `importmap.php` with an `app.css` entry. Or (what I'm leaning to now) is exactly like Encore: in `assets/app.js`, you'd `import './styles/app.css'`. That would trigger the link tag.

Please don't.
Until now, the asset-mapper component is about building on web standards. This `import './styles/app.css'` in JS file is a totally non-standard feature added by webpack.

weaverryan:
> Until now, the asset-mapper component is about building on web standards. This import './styles/app.css' in JS file is a totally non-standard feature added by webpack.

That's 100% true. But it's also massively user-friendly. To be precise, there are 2 different sides to this:

A) The ability to have `'type' => 'css'` in `importmap.php` - super nice because you can `importmap:require` CSS files instead of needing to manage `link` tags manually or downloading vendor CSS if you want to avoid a CDN.

B) The ability to `import './some.css'`. Totally non-standard, totally optional, but very convenient. This can be used in lazy Stimulus controllers to import CSS only when it's needed.

If/when web standards catch up, we can migrate to whatever that standard is. As a compromise, we could put `app.css` inside of `importmap.php` instead of importing it by default in `app.js`. `importmap.php` is already a combination of "importmap standard" + "Symfony additions". Users would still be able to import CSS files from their JavaScript files. But if we set up their main `app.css` file in `importmap.php`, then we could better encourage import CSS only when needed.

stof:
Having the JS being the trigger to load the CSS is very bad from a performance point of view: the CSS must be render-blocking while the recommended setup is to run the JS code as `defer` (so non-blocking).
Relying on an import in the JS to load the CSS through the importmap hack that makes `import 'app.css'` load a JS script adding a link tag would lead to a huge FOUC on page load because you would render the page with no CSS at all.

weaverryan:
>  Having the JS being the trigger to load the CSS is very bad from a performance point of view: the CSS must be render-blocking while the recommended setup is to run the JS code as defer (so non-blocking).
Relying on an import in the JS to load the CSS through the importmap hack that makes import 'app.css' load a JS script adding a link tag would lead to a huge FOUC on page load because you would render the page with no CSS at all.

You're totally right. Fortunately, that behavior will only be for non-preloaded CSS. The basic idea is:

* A) If a CSS file is `preload => true`, then an actual `<link rel="stylesheet">` is added. An importmap entry is also added (to avoid an error in case you DO import the CSS from JS), but it's a noop.

* B) If a CSS file is `preload => false`, then it uses the importmap trick that you mentioned. This is the case where you have some lazily-loaded JS file that imports the CSS file.

Btw, a CSS file can be `preload => true` either because it is directly in `importmap.php` with `preload: true`, or it is non-lazily imported by an entrypoint on the page (this entrypoint idea is still on my machine locally only, I'm finishing some tests).

weaverryan:
I need to fix a few tests, but PR is ready for review & description has been updated. This is now a robust system for handling CSS and it's fun to use :).

weaverryan:
Test failures are unrelated or expected (due to the high deps grabbing 7.0 code that doesn't contain the code from the PR).

This is ready for review! You can see some example usage at https://github.com/weaverryan/ux/tree/assetmapper-css/ux.symfony.com

weaverryan:
~~FYI - tracking down some last details on this...~~

weaverryan:
Found / fixed a problem with cascading the preload into "remote" importmap entries. This is ready again.

src/Symfony/Component/AssetMapper/Event/BeforeAssetsCompileEvent.php
```diff
@@ -0,0 +1,42 @@
...
+
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Contracts\EventDispatcher\Event;
+
+/**
+ * Dispatched during the asset-map:compile command, before the assets are compiled.
+ *
+ * @author Ryan Weaver <ryan@symfonycasts.com>
+ */
+class BeforeAssetsCompileEvent extends Event
```
nicolas-grekas: Isn't Pre/Post instead of Before/After more common for us?

src/Symfony/Component/AssetMapper/ImportMap/ImportMapEntries.php
```diff
@@ -0,0 +1,66 @@
...
+
+        return $this->entries[$importName];
+    }
+
+    /**
+     * @return \Traversable<ImportMapEntry>
+     */
+    public function getIterator(): \Traversable
+    {
+        return new \ArrayIterator(array_values($this->entries));
```
nicolas-grekas, line 59: what about removing `array_values`?

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
```diff
@@ -21,34 +23,58 @@ class ImportMapRenderer
...
         private readonly ImportMapManager $importMapManager,
+        private readonly ?Packages $assetPackages = null,
         private readonly string $charset = 'UTF-8',
         private readonly string|false $polyfillUrl = ImportMapManager::POLYFILL_URL,
         private readonly array $scriptAttributes = [],
     ) {
     }
 
-    public function render(string $entryPoint = null, array $attributes = []): string
+    public function render(string|array $entryPointNames, array $attributes = []): string
```
nicolas-grekas: like in ImportMapRuntime?
```suggestion
    public function render(string|array $entryPoint, array $attributes = []): string
```

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
```diff
@@ -21,34 +23,58 @@ class ImportMapRenderer
...
+                    $importMap[$importName] = 'data:application/javascript,';
+                } else {
+                    $importMap[$importName] = 'data:application/javascript,'.rawurlencode(sprintf('const d=document,l=d.createElement("link");l.rel="stylesheet",l.href="%s",(d.head||d.getElementsByTagName("head")[0]).appendChild(l)', $path));
+                }
+            } else {
+                $importMap[$importName] = $path;
+                if ($preload) {
+                    $modulePreloads[] = $path;
+                }
+            }
```
nicolas-grekas: pure CS:
```suggestion
            $preload = $data['preload'] ?? false;
            if ('css' !== $data['type']) {
                $importMap[$importName] = $path;
                if ($preload) {
                    $modulePreloads[] = $path;
                }
            } elseif ($preload) {
                $cssLinks[] = $path;
                // importmap entry is a noop
                $importMap[$importName] = 'data:application/javascript,';
            } else {
                $importMap[$importName] = 'data:application/javascript,'.rawurlencode(sprintf('const d=document,l=d.createElement("link");l.rel="stylesheet",l.href="%s",(d.head||d.getElementsByTagName("head")[0]).appendChild(l)', $path));
            }
```

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
```diff
@@ -21,34 +23,58 @@ class ImportMapRenderer
...
         }
 
-        $output = <<<HTML
-            <script type="importmap"{$attributeString}>
-            {$this->importMapManager->getImportMapJson()}
+        $scriptAttributes = $this->createAttributesString($attributes);
+        $importMapJson = json_encode(['imports' => $importMap], \JSON_THROW_ON_ERROR | \JSON_PRETTY_PRINT | \JSON_UNESCAPED_SLASHES | \JSON_HEX_TAG);
+        $output .= <<<HTML
+            <script type="importmap"{$scriptAttributes}>
+            {$importMapJson}
```
nicolas-grekas: let's be consistent (or use brackets everywhere):
```suggestion
            <script type="importmap"$scriptAttributes>
            $importMapJson
```

src/Symfony/Component/AssetMapper/ImportMap/JavaScriptImport.php
```diff
@@ -0,0 +1,36 @@
...
+use Symfony\Component\AssetMapper\MappedAsset;
+
+/**
+ * Represents a module that was imported by a JavaScript file.
+ */
+final class JavaScriptImport
+{
+    /**
+     * @param string $importName The name of the import needed in the importmap, e.g. "/foo.js" or "react".
+     * @param bool   $isLazy     whether this import was lazy or eager
```
nicolas-grekas:
```suggestion
     * @param bool   $isLazy     Whether this import was lazy or eager
```

src/Symfony/Component/AssetMapper/ImportMap/JavaScriptImport.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public function __construct(
+        public string $importName,
```
nicolas-grekas:
```suggestion
        public readonly string $importName,
```

src/Symfony/Component/AssetMapper/ImportMap/JavaScriptImport.php
```diff
@@ -0,0 +1,36 @@
...
+        public readonly bool $isLazy = false,
+        public readonly ?MappedAsset $asset = null,
+    ) {
+    }
+
+    public function isRelativeImport(): bool
+    {
+        return str_starts_with($this->importName, '.') || str_starts_with($this->importName, '/');
```
nicolas-grekas: what about protocol-relative URLs? `//foo/bar`

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -125,19 +134,34 @@ public function resolvePackages(array $packagesToRequire): array
...
+            $dependencies[] = new PackageRequireOptions($packageName, $version, $download);
 
             return sprintf('from"%s"', $packageName);
         }, $content);
 
         // source maps are not also downloaded - so remove the sourceMappingURL
-        return preg_replace('{//# sourceMappingURL=.*$}m', '', $content);
+        $content = preg_replace('{//# sourceMappingURL=.*$}m', '', $content);
+
+        return preg_replace('{/\*# sourceMappingURL=.* \*/}m', '', $content);
```
```suggestion
        return preg_replace('{/\*# sourceMappingURL=[^ ]*+ \*/}', '', $content);
```

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -125,19 +134,34 @@ public function resolvePackages(array $packagesToRequire): array
...
+    }
+
+    private function addCssEntrypoint(string $packageName, string $version, array &$packagesToRequire, bool $download): void
+    {
+        $response = $this->httpClient->request('GET', sprintf(self::URL_PATTERN_ENTRYPOINT, $packageName, $version));
+        $entrypoints = $response->toArray()['entrypoints'] ?? [];
```
nicolas-grekas: could there be a way to queue these requests instead of resolving them synchronously? the other requests are done concurrently, it'd be nice to keep this as it provides the best DX

src/Symfony/Component/AssetMapper/MappedAsset.php
```diff
@@ -76,19 +84,22 @@ public function __construct(
         foreach ($fileDependencies as $fileDependency) {
             $this->addFileDependency($fileDependency);
         }
+        foreach ($javaScriptImports as $javaScriptImport) {
+            $this->addJavaScriptImport($javaScriptImport);
+        }
```
nicolas-grekas: can't we make this shorter and faster?
```suggestion
        $this->fileDependencies = $fileDependencies;
        $this->addJavaScriptImports = $javaScriptImports;
```

src/Symfony/Component/AssetMapper/ImportMap/ImportMapConfigReader.php
```diff
@@ -0,0 +1,110 @@
...
+            $validKeys = ['path', 'url', 'downloaded_to', 'type', 'entrypoint'];
+            if ($invalidKeys = array_diff(array_keys($data), $validKeys)) {
+                throw new \InvalidArgumentException(sprintf('The following keys are not valid for the importmap entry "%s": "%s". Valid keys are: "%s".', $importName, implode('", "', $invalidKeys), implode('", "', $validKeys)));
+            }
+
+            $type = isset($data['type']) ? ImportMapType::tryFrom($data['type']) : ImportMapType::JS;
+            $isEntry = $data['entrypoint'] ?? false;
+
+            if ($isEntry && ImportMapType::JS !== $type) {
+                throw new RuntimeException(sprintf('The "entrypoint" option can only be used with the "js" type. Found in importmap.php for key "%s".', $importName));
```
nicolas-grekas:
```suggestion
                throw new RuntimeException(sprintf('The "entrypoint" option can only be used with the "js" type. Found "%s" in importmap.php for key "%s".', $importName, $type->value));
```

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
```diff
@@ -27,59 +28,65 @@ final class JavaScriptImportPathCompiler implements AssetCompilerInterface
 {
     use AssetCompilerPathResolverTrait;
 
-    // https://regex101.com/r/VFdR4H/1
-    private const IMPORT_PATTERN = '/(?:import\s+(?:(?:\*\s+as\s+\w+|[\w\s{},*]+)\s+from\s+)?|\bimport\()\s*[\'"`](\.\/[^\'"`]+|(\.\.\/)+[^\'"`]+)[\'"`]\s*[;\)]?/m';
+    // https://regex101.com/r/5Q38tj/1
+    private const IMPORT_PATTERN = '/(?:import\s+(?:(?:\*\s+as\s+\w+|[\w\s{},*]+)\s+from\s+)?|\bimport\()\s*[\'"`](\.\/[^\'"`]+|(\.\.\/)?[^\'"`]+)[\'"`]\s*[;\)]?/m';
```
nicolas-grekas: why this change?

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
```diff
@@ -104,4 +111,79 @@ private function handleMissingImport(string $message, \Throwable $e = null): voi
...
+            return $offsetStart < $commentEnd;
+        }
+
+        return false;
+    }
+
+    private function findAssetForBareImport(string $importedModule, AssetMapperInterface $assetMapper): ?MappedAsset
+    {
+        $importMapEntry = $this->importMapManager->findRootImportMapEntry($importedModule);
+        if (!$importMapEntry) {
```
nicolas-grekas:
```suggestion
        if (!$importMapEntry = $this->importMapManager->findRootImportMapEntry($importedModule)) {
```

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
```diff
@@ -104,4 +111,79 @@ private function handleMissingImport(string $message, \Throwable $e = null): voi
...
+            return $dependentAsset;
+        }
+
+        $message = sprintf('Unable to find asset "%s" imported from "%s".', $importedModule, $asset->sourcePath);
+
+        try {
+            if (null !== $assetMapper->getAsset(sprintf('%s.js', $resolvedPath))) {
+                $message .= sprintf(' Try adding ".js" to the end of the import - i.e. "%s.js".', $importedModule);
+            }
+        } catch (CircularAssetsException $e) {
```
nicolas-grekas:
```suggestion
        } catch (CircularAssetsException) {
```

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -141,6 +158,45 @@ public function downloadMissingPackages(): array
...
+    {
+        $entries = $this->importMapConfigReader->getEntries();
+
+        return $entries->has($moduleName) ? $entries->get($moduleName) : null;
+    }
+
+    /**
+     * @internal
+     */
+    public function getRawImportMapData(): array
```
nicolas-grekas: let's rename buildRawImportMapData instead and make it public + `@internal`

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -50,37 +49,54 @@ class ImportMapManager
...
         private readonly string $vendorDir,
         private readonly PackageResolverInterface $resolver,
         HttpClientInterface $httpClient = null,
     ) {
         $this->httpClient = $httpClient ?? HttpClient::create();
     }
 
-    public function getModulesToPreload(): array
+    /**
+     * @internal
```
nicolas-grekas: let's move `@internal` method after non-internal (public) ones

nicolas-grekas (commented):
LGTM, I just found minor things before approving :pray:

src/Symfony/Component/AssetMapper/ImportMap/ImportMapEntries.php
weaverryan: We use `$entry->importName` internally as the array key in `$this->entries` to help quickly return in `->has()`, and `->get()`. But, externally (i.e. in the iterator), I thought it better to have indexed keys, and force people to rely on `$entry->importName`. I had noticed myself sometimes using the array key as the import name, and other times `$entry->importName`, so I made it more strict :)

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
weaverryan: We're looking for `import` statements inside JS files. Before this PR, we did this for only one reason: to find relative imports (`import './foo.js'`) so we could automatically add those to the importmap.

Now, there is one other purpose: to know which files we should preload (i.e. if `app.js` imports `foo.js`, and `app.js` is the entrypoint, then preload `foo.js`). This preloading applies now to ALL imports - relative or "bare" (i.e. `import 'bootstrap'`). This regex change allows us to match both relative `import './foo.js'` and bare `import 'foo'` imports.

nicolas-grekas: this should then be a `*` instead of a `?`?

src/Symfony/Component/AssetMapper/ImportMap/JavaScriptImport.php
weaverryan: This method name was super unclear. What I wanted to say was "does this import represent a module that we should automatically add to the importmap". Changes in: https://github.com/weaverryan/symfony/commit/2a968ef97b6fca3dea28f1f309b289383ccfb2b3

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
weaverryan: Both seem to work, but I see your point. Changed to `*`

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -160,30 +208,48 @@ public static function parsePackageName(string $packageName): ?array
         return $matches;
     }
 
-    private function buildImportMapJson(): void
+    /**
+     * @internal
+     *
+     * @return array<string, array{path: string, type: string}>
+     */
+    public function getRawImportMapData(): array
```
nicolas-grekas, line 197: please move it up before the static method :)

nicolas-grekas (approved):

src/Symfony/Bridge/Twig/Extension/ImportMapRuntime.php
```diff
@@ -22,7 +22,7 @@ public function __construct(private readonly ImportMapRenderer $importMapRendere
     {
     }
 
-    public function importmap(?string $entryPoint = 'app', array $attributes = []): string
+    public function importmap(string|array $entryPoint, array $attributes = []): string
```
nicolas-grekas: this breaks existing apps that use [the default recipe](https://github.com/symfony/recipes/blob/d222415825b0e1605ec6b6271823a5894ed4cbc5/symfony/asset-mapper/6.3/manifest.json#L14)

is this desired?

nicolas-grekas (commented):
please add a line to the changelog (of the twig-bridge also I guess)

src/Symfony/Bridge/Twig/Extension/ImportMapRuntime.php
stof: Removing support for the implement `app` entrypoint should go through a deprecation even if the component is experimental IMO due to this.

stof: and probably also the support of passing `null` then if we have a deprecation layer.

weaverryan: good catch. Check the last commit. I've left `app` as the default (in part for BC). But in the recipe, I WOULD like to start people with `importmap('app')` because it's more explicit.

nicolas-grekas (approved):
(merging to unlock follow up PRs on the topic, let's iterate. Please report back if you have any concerns of course)

nicolas-grekas:
Thank you @weaverryan.

PR #46328, opened 2022-05-11
[Config] Allow scalar configuration in PHP Configuration

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix https://github.com/symfony/monolog-bundle/pull/417#issuecomment-1123294337
| License       | MIT
| Doc PR        | -

Fixes passing scalar values to array nodes that have a `beforeNormalization` hook, eg:

```php
return static function (MonologConfig $config): void {
    $config->handler('console')
        // ...
        ->processPsr3Messages()
            ->enabled(true)
    ;
};
```

can be shortened thanks to a [`beforeNormalization` hook](https://github.com/symfony/monolog-bundle/blob/a41bbcdc1105603b6d73a7d9a43a3788f8e0fb7d/DependencyInjection/Configuration.php#L453):

```php
return static function (MonologConfig $config): void {
    $config->handler('console')
        // ...
        ->processPsr3Messages(true)
    ;
};
```

I've used some of the code from https://github.com/symfony/symfony/pull/44166 by @jderusse. Since his PR is a feature it can't go on 5.4, but it still helped a lot.

PR #51544, opened 2023-09-04
[Config] Allow to always use config classes

alexndlm:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

When the config node doesn't have normalization, using the config object is impossible.

When the config node has some array normalization (like `http_codes` in `\Symfony\Bundle\FrameworkBundle\DependencyInjection\Configuration::addHttpClientRetrySection`), it is not possible to use the config object.

This PR relates to changes introduced in:
- https://github.com/symfony/symfony/pull/46328
- https://github.com/symfony/symfony/pull/44166

Снимок экрана 2023-09-01 в 22 22 28

### DIFF
--- a/src/Symfony/Component/Config/Builder/ConfigBuilderGenerator.php
+++ b/src/Symfony/Component/Config/Builder/ConfigBuilderGenerator.php
@@ -260,7 +260,6 @@ public function NAME(string $VAR, TYPE $VALUE): static
         $comment .= sprintf(' * @psalm-return (TValue is array ? %s : static)'."\n ", $childClass->getFqcn());
         $comment = "/**\n$comment*/\n";
 
-
         if (null === $key = $node->getKeyAttribute()) {
             $body = '
 COMMENTpublic function NAME(mixed $value = []): CLASS|static
@@ -544,10 +543,5 @@ public function NAME(string $key, mixed $value): static
         }
 
         return [] !== $r->getValue($node);
-    }
-
-    private function getType(string $classType, bool $hasNormalizationClosures): string
-    {
-        return $classType.($hasNormalizationClosures ? '|scalar' : '');
     }
 }
--- a/src/Symfony/Component/Config/Builder/ConfigBuilderGenerator.php
+++ b/src/Symfony/Component/Config/Builder/ConfigBuilderGenerator.php
@@ -14,7 +14,6 @@ namespace Symfony\Component\Config\Builder;
 use Symfony\Component\Config\Definition\ArrayNode;
 use Symfony\Component\Config\Definition\BaseNode;
 use Symfony\Component\Config\Definition\BooleanNode;
-use Symfony\Component\Config\Definition\Builder\ExprBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
 use Symfony\Component\Config\Definition\EnumNode;
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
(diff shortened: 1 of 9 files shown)


nicolas-grekas:
Thanks for working on this, it's related to the limitation I found in #51273
Can you please rebase+target 6.4? That's an improvement, not a bugfix to me.

src/Symfony/Component/Config/Definition/Builder/ArrayNodeDefinition.php
```diff
@@ -374,7 +374,7 @@ protected function createNode(): NodeInterface
 
             if ($this->default) {
                 if (!\is_array($this->defaultValue)) {
-                    throw new \InvalidArgumentException(sprintf('%s: the default value of an array node has to be an array.', $node->getPath()));
+                    throw new \InvalidArgumentException(sprintf('"%s": the default value of an array node has to be an array.', $node->getPath()));
```
nicolas-grekas, line 377: should be reverted, that's a false-positive from fabbot I suppose

src/Symfony/Component/Config/Definition/Builder/ExprBuilder.php
```diff
@@ -21,14 +21,7 @@
...
 class ExprBuilder
 {
-    public const TYPE_ANY = 'any';
-    public const TYPE_STRING = 'string';
-    public const TYPE_NULL = 'null';
-    public const TYPE_ARRAY = 'array';
-
     protected $node;
-
-    public $allowedTypes;
```
nicolas-grekas, line 31: removing consts and public properties is a BC break, we need to figure out a way that doesn't break BC

nicolas-grekas (commented):
The changes on the Config components are BC breaks. Can we find a way that doesn't break BC?

src/Symfony/Component/Config/Builder/ConfigBuilderGenerator.php
```diff
@@ -428,8 +377,11 @@ private function getComment(BaseNode $node): string
             if ($node instanceof EnumNode) {
                 $comment .= sprintf(' * @param ParamConfigurator|%s $value', implode('|', array_unique(array_map(fn ($a) => !$a instanceof \UnitEnum ? var_export($a, true) : '\\'.ltrim(var_export($a, true), '\\'), $node->getValues()))))."\n";
             } else {
-                $parameterTypes = $this->getParameterTypes($node);
-                $comment .= ' * @param ParamConfigurator|'.implode('|', $parameterTypes).' $value'."\n";
+                $parameterType = $this->getParameterType($node);
+                if (null === $parameterType || '' === $parameterType) {
```
stof, line 381: why returning an empty string to mean `mixed` instead of returning `mixed` ?

src/Symfony/Component/Config/Tests/Builder/Fixtures/NodeInitialValues/Symfony/Config/NodeInitialValuesConfig.php
```diff
@@ -28,9 +41,22 @@ public function someCleverName(array $value = []): \Symfony\Config\NodeInitialVa
         return $this->someCleverName;
     }
 
-    public function messenger(array $value = []): \Symfony\Config\NodeInitialValues\MessengerConfig
+    /**
+     * @template TValue
+     * @param TValue $value
+     * @return \Symfony\Config\NodeInitialValues\MessengerConfig|$this
+     * @psalm-return (TValue is array ? \Symfony\Config\NodeInitialValues\MessengerConfig : static)
```
stof, line 48: why should passing an array return a MessengerConfig while passing another kind of value would be fluent ? What does this correspond to ?

src/Symfony/Component/Config/Tests/Builder/Fixtures/ScalarNormalizedTypes/Symfony/Config/ScalarNormalizedTypesConfig.php
```diff
@@ -39,7 +39,7 @@ public function simpleArray(ParamConfigurator|string|array $value): static
     /**
      * @return $this
      */
-    public function keyedArray(string $name, ParamConfigurator|string|array $value): static
+    public function keyedArray(string $name, mixed $value): static
```
stof, line 42: this type becoming wider looks like a regression to me.

src/Symfony/Component/Config/Tests/Builder/Fixtures/NodeInitialValues/Symfony/Config/NodeInitialValuesConfig.php
nicolas-grekas: I wondered about the same and I think I've got the answer:
- When you pass a non-array, you likely won't want to configure anything else nested in the tree you're configuring. That's what happens for other config formats, structurally.
- When you pass an array, you can configure nested options, so that then it might make sense to return the nested config-builder.

I'm just wondering about the check: should we return the config-builder when the array is non empty?
Or should we do this?

```suggestion
     * @psalm-return (TValue is non-empty-array ? $this :\Symfony\Config\NodeInitialValues\MessengerConfig)
```

nicolas-grekas:
Let me close as stalled and superseded by #51273
Feel free to open an issue to describe any remauning issue of course.

PR #51545, opened 2023-09-04
[AssetMapper] Make importmap:install command download the same version

jmsche:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

This PR improves the behaviour of the new `importmap:install` command by downloading the same version of the files that were already installed, given we already have the URL.

### DIFF
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
@@ -13,10 +13,11 @@ namespace Symfony\Component\AssetMapper\ImportMap;
 
 use Symfony\Component\AssetMapper\AssetDependency;
 use Symfony\Component\AssetMapper\AssetMapperInterface;
-use Symfony\Component\AssetMapper\Exception\RuntimeException;
 use Symfony\Component\AssetMapper\ImportMap\Resolver\PackageResolverInterface;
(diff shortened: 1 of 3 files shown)


weaverryan (approved):
Yup, for the original PR, I gave @jmsche some bad advice: I told him to NOT simply download the `url` and instead go through the provider system so that, if the downloaded file contains any other imports, we ALSO add those to the importmap.

But that was wrong. For this workflow, the user has already downloaded (via `--download`) any entries that have the `download_to` option. And when they did, any "imports" for other libraries were already added to the importmap. At this point, we simply want to download the `url` and do nothing else.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -55,14 +56,17 @@ class ImportMapManager
     private array $importMapEntries;
     private array $modulesToPreload;
     private string $json;
+    private HttpClientInterface $httpClient;
```
stof: should be readonly for consistency with other properties assigned in the constructor

jmsche:
Failures unrelated

fabpot (approved):

fabpot:
Thank you @jmsche.

PR #51546, opened 2023-09-04
[Messenger] Added generics for message in envelope

RikudouSage:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| License       | MIT

Adds generics to the Envelope class which allows user code to typehint it.

I created this because I have this piece of code which would benefit from it:

```
    /**
     * @template T of object
     * @param class-string<T> $type
     * @return array<Envelope<T>>
     */
    public function getJobsByType(string $type): array
    {
    }
```

### DIFF
--- a/src/Symfony/Component/Messenger/Envelope.php
+++ b/src/Symfony/Component/Messenger/Envelope.php
@@ -17,6 +17,8 @@ use Symfony\Component\Messenger\Stamp\StampInterface;
  * A message wrapped in an envelope with stamps (configurations, markers, ...).
  *
  * @author Maxime Steinhausser <maxime.steinhausser@gmail.com>
+ *
+ * @template TMessage of object
  */
 final class Envelope
 {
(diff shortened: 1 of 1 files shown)


RikudouSage:
I don't quite understand why the Psalm errors happen.

RikudouSage:
Can I get some assistance, please?

derrabus:
> I don't quite understand why the Psalm errors happen.

You used a template type on a static method that you did not declare. Static methods don't inherit template types from their class, you have to use a method-level template type.

src/Symfony/Component/Messenger/Envelope.php
```diff
@@ -17,18 +17,24 @@
...
  */
 final class Envelope
 {
     /**
      * @var array<class-string<StampInterface>, list<StampInterface>>
      */
     private array $stamps = [];
+
+    /**
+     * @var TMessage|Envelope<TMessage>
```
ro0NL, line 31: a union wont really help us with autocompletion isnt it?

src/Symfony/Component/Messenger/Envelope.php
```diff
@@ -42,6 +48,7 @@ public function __construct(object $message, array $stamps = [])
     /**
      * Makes sure the message is in an Envelope and adds the given stamps.
      *
+     * @param TMessage         $message
      * @param StampInterface[] $stamps
      */
```
wouterj, line 53: This is wrong, the method is supposed to handle any `object`, see line 56.

I think we can make it something like this:

```suggestion
     * @template TNewMessage of object
     *
     * @param TNewMessage      $message
     * @param StampInterface[] $stamps
     *
     * @return self<TNewMessage>
     */
```

wouterj (changes requested):

ro0NL:
`MessageBusInterface` should be updated btw

fabpot:
Closing as there is no more activity. Feel free to reopen when you have time.

PR #51548, opened 2023-09-04
[Uid] Make child methods match parent's argument names

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

`AbstractUid` defines methods with arguments named `$uid`, and I believe child methods should follow these names unless there's a reason I am now aware of?


nicolas-grekas:
This was on purpose to me.

PR #51553, opened 2023-09-04
[Scheduler] Allow modifying the schedule at runtime and recalculate heap

### LINKED ISSUE #51206: [Scheduler] "dynamic" schedules
### Description

I've talked to a few users who keep their scheduled tasks in a database and want to use scheduler to run them. This is not easily possible now as the schedules are set when first running `messenger:consume scheduler_default`.

Within the current scheduler architecture, what would be required to achieve this? I'm don't have a great understanding of the deep scheduler internals (the trigger heap system).

### Example

```php
class MySchedule implements ScheduleProviderInterface
{
    {
        $schedule = new Schedule();

        // fetch tasks from db and add to schedule as recurring messages

        return $schedule;
    }
}
```

Jeroeny:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #51206
| License       | MIT

Schedules are created statically by a ScheduleProviderInterface. This is nice if you have a small app with some predetermined scheduled actions. But if you want to schedule business logic dynamically, e.g. have admins Crud a schedule in the DB, then the schedule needs to be updateable at runtime. This also allows for the app to re-add schedules that have returned `null` as nextRunDate in the past (making them unscheduled) and to remove schedules.

Usage example: reading the schedule from a DB repository, adding messages based on an custom event:

```php
class MySchedule implements ScheduleProviderInterface
{
    private ?Schedule $schedule = null;

    public function __construct(private RecurrenceRepository $repository)
    {
    }

    {
        return $this->schedule ??= Schedule::with(...$this->repository->all());
    }

    #[AsEventListener(ScheduledMessageAdded::class)]
    public function onScheduleAdded(ScheduledMessageAdded $event): void
    {
        $this->schedule->add($event->getMessage());
    }
}
```

Or in the context of Messenger, if some periodic batch task has no more work to do, the recurring message can be removed:

```php
class MySchedule implements ScheduleProviderInterface
{
    private ?Schedule $schedule = null;

    {
        return $this->schedule ??= Schedule::with(RecurringMessage::every(60, new BatchGenerateFiles()));
    }
    
    public function clear()
    {
        $this->schedule->clear();
    }
}

class MyHandler 
{
    #[AsMessageHandler]
    public function handle(BatchGenerateFiles $message)
    {
        // <-- handle message code -->

        if ($noMoreFilesToGenerate) {
            $this->mySchedule->clear();
        }
    }
}
```

### DIFF
--- a/src/Symfony/Component/Scheduler/CHANGELOG.md
+++ b/src/Symfony/Component/Scheduler/CHANGELOG.md
@@ -10,6 +10,7 @@ CHANGELOG
  * Add `AbstractTriggerDecorator`
  * Make `ScheduledStamp` "send-able"
  * Add `ScheduledStamp` to `RedispatchMessage`
+ * Allow modifying the Schedule at runtime
 
 6.3
 ---
--- a/src/Symfony/Component/Scheduler/Schedule.php
+++ b/src/Symfony/Component/Scheduler/Schedule.php
@@ -21,19 +21,54 @@ final class Schedule implements ScheduleProviderInterface
     private array $messages = [];
     private ?LockInterface $lock = null;
     private ?CacheInterface $state = null;
+    private bool $shouldRestart = false;
+
+    public static function with(RecurringMessage $message, RecurringMessage ...$messages): static
+    {
+        return static::doAdd(new self(), $message, ...$messages);
+    }
 
     /**
      * @return $this
      */
     public function add(RecurringMessage $message, RecurringMessage ...$messages): static
     {
+        $this->setRestart(true);
+
+        return static::doAdd($this, $message, ...$messages);
+    }
+
+    private static function doAdd(self $schedule, RecurringMessage $message, RecurringMessage ...$messages): static
+    {
         foreach ([$message, ...$messages] as $m) {
-            if (isset($this->messages[$m->getId()])) {
+            if (isset($schedule->messages[$m->getId()])) {
                 throw new LogicException('Duplicated schedule message.');
             }
 
-            $this->messages[$m->getId()] = $m;
+            $schedule->messages[$m->getId()] = $m;
         }
+
+        return $schedule;
+    }
+
+    /**
+     * @return $this
+     */
+    public function remove(RecurringMessage $message): static
+    {
+        unset($this->messages[$message->getId()]);
+        $this->setRestart(true);
+
+        return $this;
+    }
+
+    /**
+     * @return $this
+     */
+    public function clear(): static
+    {
+        $this->messages = [];
+        $this->setRestart(true);
 
         return $this;
     }
@@ -82,5 +117,15 @@ final class Schedule implements ScheduleProviderInterface
     public function getSchedule(): static
     {
         return $this;
+    }
+
+    public function shouldRestart(): bool
+    {
+        return $this->shouldRestart;
+    }
+
+    public function setRestart(bool $shouldRestart): bool
+    {
+        return $this->shouldRestart = $shouldRestart;
     }
 }
--- a/src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
+++ b/src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
@@ -19,7 +19,7 @@ use Symfony\Component\Scheduler\ScheduleProviderInterface;
 
 final class MessageGenerator implements MessageGeneratorInterface
 {
-    private Schedule $schedule;
+    private ?Schedule $schedule = null;
     private TriggerHeap $triggerHeap;
     private ?\DateTimeImmutable $waitUntil;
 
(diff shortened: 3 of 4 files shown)


kbond:
Interesting, is this a companion PR for #51542?

Jeroeny:
@kbond they don't depend on each other. But when both of them are merged, it enables a lot of capabilities for the Scheduler.

src/Symfony/Component/Scheduler/Schedule.php
```diff
@@ -22,19 +22,49 @@ final class Schedule implements ScheduleProviderInterface
     private ?LockInterface $lock = null;
     private ?CacheInterface $state = null;
```
fabpot: Blank line to be removed

src/Symfony/Component/Scheduler/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ CHANGELOG
  * Add `AbstractTriggerDecorator`
  * Make `ScheduledStamp` "send-able"
  * Add `ScheduledStamp` to `RedispatchMessage`
+ * Allow modifying the Schedule at runtime and recalculate heap
```
fabpot:
```suggestion
 * Allow modifying the Schedule at runtime
```

(the heap is an implementation detail)

src/Symfony/Component/Scheduler/Schedule.php
```diff
@@ -83,4 +113,9 @@ public function getSchedule(): static
     {
         return $this;
     }
+
+    public function restart(bool $restart = null): bool
```
fabpot: For more clarity, I would split this method: `setRestart(bool)` and `shouldRestart(): bool`

src/Symfony/Component/Scheduler/Schedule.php
```diff
@@ -22,19 +22,49 @@ final class Schedule implements ScheduleProviderInterface
     private ?LockInterface $lock = null;
     private ?CacheInterface $state = null;
 
+    private bool $restart = false;
```
fabpot:
```suggestion
    private bool $shouldRestart = false;
```

src/Symfony/Component/Scheduler/Schedule.php
```diff
@@ -22,19 +22,49 @@ final class Schedule implements ScheduleProviderInterface
...
             }
 
-            $this->messages[$m->getId()] = $m;
+            $schedule->messages[$m->getId()] = $m;
         }
 
+        return $schedule;
+    }
+
+    public function remove(RecurringMessage $message): static
```
fabpot, line 57:
```suggestion
    /**
     * @return $this
     */
    public function remove(RecurringMessage $message): static
```

fabpot (changes requested):
Can you add some usage example in the PR description?

src/Symfony/Component/Scheduler/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ CHANGELOG
  * Add `AbstractTriggerDecorator`
  * Make `ScheduledStamp` "send-able"
  * Add `ScheduledStamp` to `RedispatchMessage`
+ * Allow modifying the Schedule at runtime
```
fabpot, line 13:
```suggestion
 * Allow modifying Schedule instances at runtime
```

fabpot (approved):

fabpot:
Thank you @Jeroeny.

Sophikitis:
Hi ! 

Do you have any examples of use?
I'm trying to implement this, to dynamically retrieving frequencies and modifying the behavior with a custom event. 
But this->schedule is always null in the listener function. 
When I retrieve the instance from $this->getSchedule() I get an instance but when I try to delete a message or add one, nothing happens.

And as far as I understand, there's no documentation for this feature.
I'll try my luck here. 
Thanks in advance

Hanmac:
@fabpot I'm also looking for an example, currently i uses this Doctrine Query to build the Schedule

```php
    {
        $result = new Schedule();

        $repo = $this->manager->getRepository(Terminal::class);
        $q = $repo->createQueryBuilder('t');

        $q->where('t.status = true AND t.automatic = true');

        foreach ($q->getQuery()->getResult() as $r) {
            /**
             * @var Terminal $r
             */
            $result->add(RecurringMessage::cron(
                $r->getCronExpr(),
                new KassenschlussTask($r->getId()),
                $this->timezoneDetector->getTimezone())
            );
        }
        return $result;
    }
```

because each object might have its own CronExpr, each of them needs its own RecurringMessage

and how do i handle if they get disabled?

Hanmac:
@Sophikitis while maybe not the best way, i found `StopWorkersCommand`

```php
$cacheItem = $this->restartSignalCachePool->getItem(StopWorkerOnRestartSignalListener::RESTART_REQUESTED_TIMESTAMP_KEY);
$cacheItem->set(microtime(true));
$this->restartSignalCachePool->save($cacheItem);
```

i used `[AsEntityListener]` on my DB Entities, and then on `Events::postRemove, Events::postUpdate, Events::postPersist` i stop the messenger worker

gremo:
I'm facing the same issue when fetching messages from the database. I haven't found a reliable way to make it work. We really need some documentation and a simple example to help users!

gremo:
Ping. Is it possible that no one has even the faintest idea of how to update the scheduler without restarting the worker, which in some cases can be completely impractical?

This need to be documentend because it's so confusing.

PR #51559, opened 2023-09-04
[DependencyInjection] `#[Autowire]` attribute should have precedence over bindings

### LINKED ISSUE #49082: TaggedLocator attribute does not work in controller constructor

6.2.* (and maybe other)

### Description

A wrong Container is injected when using `#[TaggedLocator('XXX')]` in an argument of a controller constructor 

### How to reproduce

add this controller

```php
class AController extends AbstractController
{
    public function __construct(
        #[TaggedLocator('foobar')]
        private readonly ContainerInterface $foo,
    ) {
    }

    #[Route('/', name: 'app_a')]
    public function index(
    ): JsonResponse
    {
        dd($this->foo);
        return $this->json([
            'message' => 'Welcome to your new controller!',
            'path' => 'src/Controller/AController.php',
        ]);
    }
}
```

### Possible Solution

We can work-arround this issue with the following constructor:

```diff
     public function __construct(
         #[TaggedLocator('foobar')]
-        private readonly ContainerInterface $foo,
+        private readonly ServiceLocator $foo,
     ) {
     }
```

but ...

### Additional Context

I guess - there is a conflict with the "ServiceSubscriberInterface"

image

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #49082
| License       | MIT
| Doc PR        | -

It seems logical (I think) that the `#[Autowire]` attribute has precedence over any bindings.

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Compiler/ResolveBindingsPass.php
+++ b/src/Symfony/Component/DependencyInjection/Compiler/ResolveBindingsPass.php
@@ -14,6 +14,7 @@ namespace Symfony\Component\DependencyInjection\Compiler;
 use Symfony\Component\DependencyInjection\Argument\BoundArgument;
 use Symfony\Component\DependencyInjection\Argument\ServiceLocatorArgument;
 use Symfony\Component\DependencyInjection\Argument\TaggedIteratorArgument;
+use Symfony\Component\DependencyInjection\Attribute\Autowire;
 use Symfony\Component\DependencyInjection\Attribute\Target;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
@@ -183,6 +184,13 @@ class ResolveBindingsPass extends AbstractRecursivePass
                     continue;
                 }
                 if (\array_key_exists($parameter->name, $arguments) && '' !== $arguments[$parameter->name]) {
+                    continue;
+                }
+                if (
+                    $value->isAutowired()
+                    && !$value->hasTag('container.ignore_attributes')
+                    && $parameter->getAttributes(Autowire::class, \ReflectionAttribute::IS_INSTANCEOF)
+                ) {
                     continue;
                 }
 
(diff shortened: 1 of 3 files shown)


ruudk:
Thanks for fixing this 🙏 . Could you rebase this?

stof:
@nicolas-grekas is it right to have to make ResolveBindingPass aware of the Autowire attribute ? Shouldn't this be handled by the AutowirePass ?

nicolas-grekas:
I agree, the bindings pass has no reasons to know about autowiring stuff.
Can't we fix this by changing the order of the two passes?

HypeMC:
@nicolas-grekas @stof My initial approach was to fix it in `AutowirePass`. I don't remember the exact reason, but I couldn't do it like that so I went with this instead.

I can try to change the order of the passes, but I'm worried that might break other stuff. I'll try it sometime today or tomorrow.

nicolas-grekas:
It's also not obvious to me what we want here.

What this PR achieves is this precedence, isn't it?

1. autowiring by attribute
2. binding by type+name of argument
3. binding by type only
4. autowiring by type+name of argument
5. autowiring by type only

Dunno if we should account for that but bindings can also be defined by explicit configuration and/or by autoconfiguration.

Dunno how we can make AutowirePass achieve this ordering (or another) given that once bindings are resolved, we don't know if some argument's values came through bindings or via another path (explicit/auto config).

Maybe this approach is the most pragmatic.

WDYT?

src/Symfony/Component/DependencyInjection/Compiler/ResolveBindingsPass.php
```diff
@@ -185,6 +186,9 @@ protected function processValue(mixed $value, bool $isRoot = false): mixed
                 if (\array_key_exists($parameter->name, $arguments) && '' !== $arguments[$parameter->name]) {
                     continue;
                 }
+                if ($parameter->getAttributes(Autowire::class, \ReflectionAttribute::IS_INSTANCEOF)) {
```
nicolas-grekas: we should at least check the `'container.ignore_attributes'` tag here
and also if the class is autowired

HypeMC: Done

nicolas-grekas (approved):

nicolas-grekas:
Any other opinion here? @stof?

chalasr (approved):

nicolas-grekas:
Thank you @HypeMC.

PR #51575, opened 2023-09-06
[DependencyInjection] Add date_time and date_time_format env processors

secit-pl:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| BC breaks? | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | [symfony/symfony-docs#18835](https://github.com/symfony/symfony-docs/pull/18835)

Added two env processors:

date_time - cast parameter do \DateTimeImmutable object instance
date_time_format - convert \DateTimeImmutable object to formatted string

**Before PR:**

.env
```
DATE_TIME='2023-09-11 09:00:00'
DATE_TIME_FORMAT='Y.m.d H:i'
```

config/services.yaml
```yaml
    ...

    App\Example\Service:
        arguments:
            $dateTime: '%env(DATE_TIME)%'
            $dateTimeFormat: '%env(DATE_TIME_FORMAT)%'
```

App\Example\Service
```php
namespace App\Example;

class Service
{
    private \DateTimeImmutable $dateTime;
    private string $dirName;

    public function __construct(string $dateTime, string $dateTimeFormat)
    {
        $this->dateTime = new \DateTimeImmutable($dateTime);
        $this->dirName = '/path/to/'.$this->dateTime->format($dateTimeFormat).'/data/';
    }
}

```

**After PR:**

.env
```
DATE_TIME='2023-09-11 09:00:00'
DATE_TIME_FORMAT='Y.m.d H:i'
```

config/services.yaml
```yaml
    ...
    parameters:
        format: '%env(DATE_TIME_FORMAT)%'

    App\Example\Service:
        arguments:
            $dateTime: '%env(date_time:DATE_TIME)%'
            $dirName: '/path/to/%env(date_time_format:format:date_time:DATE_TIME)%/data/'
```

App\Example\Service
```php
namespace App\Example;

class Service
{
    public function __construct(
        private \DateTime $dateTime,
        private string $dirName,
    ) {
    }
}

```


src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
```diff
@@ -57,6 +57,8 @@ public static function getProvidedTypes(): array
             'enum' => \BackedEnum::class,
             'shuffle' => 'array',
             'defined' => 'bool',
+            'date_time' => \DateTime::class,
+            'date_time_immutable' => \DateTimeImmutable::class,
```
GromNaN: `date_time_immutable` should be the default for `date_time`. No need to have 2 distinct processors. Allowing the mutable `DateTime` would lead to unexpected behaviors if it's used in 2 services with one of them altering the value.
```suggestion
            'date_time' => \DateTimeImmutable::class,
```

src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
```diff
@@ -341,6 +343,22 @@ public function getEnv(string $prefix, string $name, \Closure $getEnv): mixed
...
+            try {
+                return new \DateTime($env);
+            } catch (\Exception) {
+                throw new RuntimeException(sprintf('Env var "%s" is not a valid date time string.', $name));
+            }
+        }
+
+        if ('date_time_immutable' === $prefix) {
+            try {
+                return new \DateTimeImmutable($env);
```
GromNaN: You should use the [Clock component](https://symfony.com/doc/current/components/clock.html) when available to support mockable `now`.

GromNaN (changes requested):
Interesting feature. You should use the Clock component when it is available and only work with DateTimeImmutable.

I'm not sure if DateTimeInterface objects are supported in Yaml configuration. I think it's important to make sure that environment variables aren't the only way to define a date object in config.

ro0NL:
we have a similar usecase where we need to get a formatted datetime for `DATE_TIME=now` at compile time

perhaps we could support `date_time:format_param:DATE_TIME` while at it :angel: (`format_param=Ymd` in our case)

src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
secit-pl: I'm not sure I understand you correctly. Clock component (as I see in documentation) can return only ->now(), so I can change the code to use Clock::now() (if available) when the $env === 'now' but in other cases I still need to use \DateTimeImmutable, right?

GromNaN: The clock allows to set the default timezone.
You can find some conversion logic in `DateTimeValueResolver`.

https://github.com/symfony/symfony/blob/2bac801f8a4cba02e8ecc840fa7bf512034f860a/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/DateTimeValueResolver.php#L86

stof:
@ro0NL an env var processor will not resolve `now` at compile time but at runtime.

src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
stof: I agree about that. I would not add support for the mutable value object.

ro0NL:
@stof you're right, i meant to inject Ymd string in config: `some: '/path/%env(date_time:format:DATE_TIME)%/subdir'`

GromNaN:
Note that's it's actually possible to define a service with `DateTimeImmutable` class that get the env var as argument.

```yaml
services:
    example_date:
        class: 'DateTimeImmutable'
        arguments: ['%env(DATE_TIME)%']

    App\Example\Service:
        arguments:
            $dateTime: '@example_date'
```

It's inlined by the compiler:

```php
new \App\Example\Service(new \DateTimeImmutable($container->getEnv('EXAMPLE_DATE'));
```

src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
secit-pl: Done.

src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
secit-pl: Done. I hope I did what you expected.

secit-pl:
I have updated the code and PR name and description to match new implementation.

Unfortunately after deeper testing I found that I need to make changes in other classes so this PR was changed to BC-break.

@ro0NL see PR description and `format_date_time`, I hole this is what you expected to be achieved with this PR :)

ro0NL:
i would use the intermediate parameter approach, thus `date_time_format:param:ENV`

since the format can be a arbitrary string

i would keep the prefix `date_time_` for better grouping

i believe it can be done in a single processor, as per existing `env(default:fallback_param:BAR)`

secit-pl:
You are right in case of processor name, I have changed the name from `format_date_time` to `date_time_format` but I do not agree with you to achieve both functionalities in single processor: converting string to \DateTimeImmutable object and converting date string to another date string but in different format. I think it's better to keep those things separated. Also IMHO the implementation will be more clean because if we connect those two processors in single one it will return the \DateTimeImmutable and string values which may lead to more "if-y" code.

stof:
Parsing a date and formatting a date should indeed be different processors.
Thus, if we try to make it the same based on whether you provide a format or no, the syntax might become ambiguous. A given processor needs to always expect the same number of `:` delimiters for the part of the string it owns, before reaching the next processor.

nicolas-grekas:
Ideally, we should settle on a name in #51415 and support only this name as a prefix. I think we should ban any mutable datetime.

ro0NL:
Date time prefix seems fine as per what clock contract gives us

secit-pl:
I have found one problem with date time format `Y-m-d H:i:s`.

The problem is with `:` character which is used as processor separator:

`%env(date_time_format:Y-m-d H:i:s:date_time:DATE_TIME)%`

I have tried to urlencode the date format in this case:

`%env(date_time_format:Y-m-d%20H%3Ai%3As:date_time:DATE_TIME)%`

but using % in string breaks the parsing and slices the env to `%env(date_time_format:Y-m-d%`.

Next idea is to encode the date format in base64, it should work but I do not think it's user friendly solution. 

Do you have any idea how to escape/encode the `:` and `%` in date format parameter? The problem is not directly in PR code but in the way how parser handles the unexpected `:` and `%` characters which are required in this case.

ro0NL:
hence

> i would use the intermediate parameter approach, thus date_time_format:param:ENV

secit-pl:
All problems solved. I have updated the issue description to match the current implementation. I think the failed tests are not related to PR.

PR is done and it's waiting for review and merge.

fabpot:
You need to rebase as spotted by fabbot.

nicolas-grekas (commented):
On closer look, I'm :-1: on the PR, the reason being all the changes to the return/parameter types.

As @GromNaN pointed out, it's already possible to inject a date-time from an env var. There's also an inline version of it:
```yaml
services:
   foo:
       arguments:
           - !service
                   class: DateTimeImmutable
                   arguments: ['%env(DATE)%']
```

Instead of adding a new processor, we could consider providing a simpler way to do such declarations. If an attribute is enough, it shouldn't be too hard to add support for eg `#[AutowireDateTime(etc)]`.

nicolas-grekas:
Closing as explained in my last comment. Thanks for submitting.

Issue #36668, opened 2020-05-03
[DX][Security] Add authentication information to the profiler


wouterj:
**Description**  

It would be nice to display information about authentication in the profiler:

1. Know which authenticators are checked for support and which is actually used to authenticate (maybe also show if they support lazy-authentication?)
2. Inspect the Security passport + badges generated by this authenticator (maybe erase plain text passwords from the dump?)
3. See which badges are resolved and which aren't

I think (1) and (2) can be done by creating a `TraceableAuthenticator` that decorates all authenticators and (3) should probably be done by an event listener on `CheckPassportEvent` (with a very low priority)

alexandre-daubois:
Hi @wouterj! Started to work on a little something for this issue. For now, I'm able to retrieve and display in profiler if an authenticator is able to support request and is used to authenticate the current request. Here is the result in the profiler for now:

![Capture d’écran du 2021-06-02 16-31-27](https://user-images.githubusercontent.com/2144837/120499023-00d49100-c3c0-11eb-8d85-dc8088b5cc18.png)

I've been struggling a bit with something. As `Authenticator` isn't a service, I wasn't able to create a `TraceableAuthenticator`. Instead, I inspired myself from the `TraceableAccessDecisionManager` and created a `TraceableAuthenticatorManager`.
As `AuthenticatorManager` is declared as an abstract service, I can't decorate it directly. So here's what I tried:

```
->set('debug.security.authenticator.manager', TraceableAuthenticatorManager::class)
            ->decorate('security.authenticator.manager.main')
            ->args([
                service('debug.security.authenticator.manager.inner')
            ])
```

Although it works, that's a problem if the application has more than one firewall, other than the `main` one. Do you have any idea how to deal with this? Thanks!

carsonbot:
Thank you for this suggestion.
There has not been a lot of activity here for a while. Would you still like to see this feature?

Amoifr:
Hey @wouterj! 👋

I took a shot at this one with #63867.

I was thrilled to work on this issue — it gave me a great opportunity to dive deep into the security component internals. Tracing the token refresh flow through `ContextListener`, understanding how `TraceableAuthenticator` unwraps sensitive exceptions, and figuring out how the profiler collects data... quite the learning journey! 🎓

The PR adds:
- A visible "lazy" badge on authenticators
- Exception type + message shown directly on failed authenticators
- Deauthentication details (reason, user, provider) when session token refresh fails

Looking forward to your feedback!

PR #51585, opened 2023-09-06
[Security] Add badge resolution to profiler

Jean-Beru:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | #36668
| License       | MIT
| Doc PR        | 

This PR add badges resolution status in Security profiler as mentioned in #36668 (" See which badges are resolved and which aren't").

### CSRF error

image

### Wrong credentials

image

### Authentication successful

image

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/Resources/views/Collector/security.html.twig
+++ b/src/Symfony/Bundle/SecurityBundle/Resources/views/Collector/security.html.twig
@@ -28,6 +28,25 @@
             border: 0;
             padding: 0 0 8px 0;
         }
+
+        #collector-content .authenticators .badge {
+            color: var(--white);
+            display: inline-block;
+            text-align: center;
+        }
+        #collector-content .authenticators .badge.badge-resolved {
+            background-color: var(--green-500);
+        }
+        #collector-content .authenticators .badge.badge-not_resolved {
+            background-color: var(--yellow-500);
+        }
+
+        #collector-content .authenticators svg[data-icon-name="icon-tabler-check"] {
+            color: var(--green-500);
+        }
+        #collector-content .authenticators svg[data-icon-name="icon-tabler-x"] {
+            color: var(--red-500);
+        }
     </style>
 {% endblock %}
 
@@ -316,13 +335,15 @@
                 <h3 class="tab-title">Authenticators</h3>
                 <div class="tab-content">
                     {% if collector.authenticators|default([]) is not empty %}
-                        <table>
+                        <table class="authenticators">
                             <thead>
                             <tr>
                                 <th>Authenticator</th>
                                 <th>Supports</th>
+                                <th>Authenticated</th>
                                 <th>Duration</th>
                                 <th>Passport</th>
+                                <th>Badges</th>
                             </tr>
                             </thead>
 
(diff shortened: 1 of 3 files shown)


src/Symfony/Bundle/SecurityBundle/Resources/views/Collector/security.html.twig
```diff
@@ -28,6 +28,21 @@
...
+            text-align: center;
+            min-width: 60px;
+        }
+
+        #collector-content .authenticators .badge.badge-resolved {
+            background-color: var(--green-500);
+        }
+
+        #collector-content .authenticators .badge.badge-not_resolved {
+            background-color: var(--red-500);
```
94noni: I like the badge, but red seems "errored", what about using a yellow/orange color instead ?

src/Symfony/Bundle/SecurityBundle/Resources/views/Collector/security.html.twig
```diff
@@ -340,8 +357,18 @@
                                 <tr>
                                     <td class="font-normal">{{ profiler_dump(authenticator.stub) }}</td>
                                     <td class="no-wrap">{{ source('@WebProfiler/Icon/' ~ (authenticator.supports ? 'yes' : 'no') ~ '.svg') }}</td>
+                                    <td class="no-wrap">{{ authenticator.authenticated is not null ? source('@WebProfiler/Icon/' ~ (authenticator.authenticated ? 'yes' : 'no') ~ '.svg') : '' }}</td>
```
94noni, line 364: on other profiler pages, those svg are colored, can we have this also here?

src/Symfony/Bundle/SecurityBundle/Resources/views/Collector/security.html.twig
Jean-Beru: Sure! Screenshots are updated too.

src/Symfony/Bundle/SecurityBundle/Resources/views/Collector/security.html.twig
Jean-Beru: The color is set when the svg is inside a `.metrics` tag. This logic may be expanded to other classes in all profilers.

Since it's not related to this PR, I updated my PR to color these svg in the Security panel only.

fabpot (approved):

fabpot:
Thank you @Jean-Beru.

PhilETaylor:
Composer updated a stable project and now broken :( 

ScreenShot-2023-09-10-15 26 32

PhilETaylor:
ok on debugging I can see that on the login request, when viewed In debugger, I can see the badges - so the point of this PR is working fine. 

But after login when redirected to the user dashboard of the app, and trying to load any page when already logged in, then the passport is null in the TraceableAuthenticator and this means the whole app is down with the error previously provided.

Changing `$this->passport->getBadges()` to `$this->passport?->getBadges() ?? []` suppresses the error


ScreenShot-2023-09-10-15 48 36

chalasr:
Would you mind sending a PR (with test ideally)?

PhilETaylor:
like this https://github.com/symfony/symfony/pull/51612 ?

PR #51586, opened 2023-09-06
[ErrorHandler] Handle PHP 8.3 `highlight_file` function output changes 

PhilETaylor:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | PHP 8.3 Compatibility
| New feature?  | no 
| Deprecations? | no 
| License       | MIT

PHP 8.3 changes the output from the `highlight_file` function as described here: https://php.watch/versions/8.3/highlight_file-highlight_string-html-changes

I accidentally ran into this head first when using PHP 8.3.0-RC1 in development with Symfony 6.4-dev when my error had no code rendered 

Also affected the `{{ '/path/to/file.html.twig'|file_excerpt(140,5) }}` twig code ` file_except` renderer - easier to replicate with that code just throw it into your twig file with a valid `/path/to/file.html.twig` - before this PR there would be no output, after this PR the file except would be shown. 

Same fix applied to both files.

ScreenShot-2023-09-06-21 42 56

After this fix the code rendered returned.

ScreenShot-2023-09-06-21 29 54

### DIFF
--- a/src/Symfony/Bridge/Twig/Extension/CodeExtension.php
+++ b/src/Symfony/Bridge/Twig/Extension/CodeExtension.php
@@ -123,13 +123,25 @@ final class CodeExtension extends AbstractExtension
             // highlight_file could throw warnings
             // see https://bugs.php.net/25725
             $code = @highlight_file($file, true);
-            // remove main code/span tags
-            $code = preg_replace('#^<code.*?>\s*<span.*?>(.*)</span>\s*</code>#s', '\\1', $code);
-            // split multiline spans
-            $code = preg_replace_callback('#<span ([^>]++)>((?:[^<]*+<br \/>)++[^<]*+)</span>#', function ($m) {
-                return "<span $m[1]>".str_replace('<br />', "</span><br /><span $m[1]>", $m[2]).'</span>';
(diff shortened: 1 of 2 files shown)


derrabus:
Thank you. Does this affect 5.4 or 6.3 as well? If that's the case, please retarget your PR.

PhilETaylor:
looks like it needs to be 5.4. https://github.com/symfony/symfony/blob/fcb754a1c08ef0b283f1f0caf0c8666098cb7fb6/src/Symfony/Component/ErrorHandler/ErrorRenderer/HtmlErrorRenderer.php#L271

PhilETaylor:
its late and my eyes sleepy, but I think Ive downgraded it ok, and removed short function syntax in my original to be compatible with PHP 7 also. Should be ok to test now.

javiereguiluz (approved):
It's very important to update Symfony for this BC-break in PHP 8.3. Thanks Phil for taking care of this!

stof (approved):

lyrixx (approved):

nicolas-grekas:
Thank you @PhilETaylor.

PR #51588, opened 2023-09-06
[FrameworkBundle] Always use buildDir as `ConfigBuilderGenerator` outputDir

### LINKED ISSUE #51582: [Config] Dumped config builders in cache dir when defining a custom `kernel.build_dir`

6.4

### Description

I tried to play with [config builders](https://github.com/symfony/symfony/pull/40600).

The initial PR states:
>This PR will build classes and store them in the build_dir. The classes will help you write PHP config. It will basically generate an array for you.

When I use the defaults, the build dir is the same as the cache dir and all builders are compiled in `var/cache/dev/Symfony/` as expected.

When I use a custom build dir, only the FrameworkBundle config builders are compiled in `build/Symfony/`, but for all other bundles they are still compiled in cache dir, including duplicates for the framework bundle ones.

### How to reproduce

Override the `Kernel::getBuildDir()` method and clear the cache in any Symfony application.

### Possible Solution

I did not invest more time to find out where it comes from as I don't know this part of the internals.
If someone has a better understanding of this feature, I'll be glad to have some heads-up, otherwise I'll try to have a look this week-end.

### Additional Context

_No response_

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #51582
| License       | MIT
| Doc PR        | -

The `ConfigBuilderGenerator` is instantiated in three places, two of which use `kernel.build_dir` as the output dir, and one which uses `kernel.cache_dir`:
- https://github.com/symfony/symfony/blob/29f427fcbd78b700a1b253c7927e51017d964f59/src/Symfony/Component/HttpKernel/Kernel.php#L736
- https://github.com/symfony/symfony/blob/29f427fcbd78b700a1b253c7927e51017d964f59/src/Symfony/Component/DependencyInjection/Extension/ExtensionTrait.php#L55-L61
- https://github.com/symfony/symfony/blob/29f427fcbd78b700a1b253c7927e51017d964f59/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ConfigBuilderCacheWarmer.php#L43-L45

This PR fixes this inconsistency so that `kernel.build_dir` is always used.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ConfigBuilderCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ConfigBuilderCacheWarmer.php
@@ -44,7 +44,7 @@ class ConfigBuilderCacheWarmer implements CacheWarmerInterface
      */
     public function warmUp(string $cacheDir)
     {
-        $generator = new ConfigBuilderGenerator($cacheDir);
+        $generator = new ConfigBuilderGenerator($this->kernel->getBuildDir());
 
         foreach ($this->kernel->getBundles() as $bundle) {
             $extension = $bundle->getContainerExtension();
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Thank you @HypeMC.

PR #51592, opened 2023-09-07
[RateLimiter] Fix SlidingWindow calculation

### LINKED ISSUE #40289: [RateLimiter] retry-after of SlidingWindowLimiter is incorrect
**Symfony version(s) affected**: 5.2.x

**Description**

The SlidingWindowLimiter exposes a `retry-after` value [intended for exposing to the user](https://symfony.com/doc/current/rate_limiter.html#exposing-the-rate-limiter-status). 

Instead of the time after which you can retry the current request (for when the request was rejected) the value represents when the current window will end. This is confusing as the limiter can potentially have a previous window set, in which case the retry-after is way sooner than expected.

**How to reproduce**
- Given a previous "full" 60-second 60-token window
- We're 12 seconds into the new window with already 12 tokens accounted for (that's 12+ 60*(0.8) = 60 tokens used)
- Given a 1-token costing consumer
- This would mean the `X-RateLimit-Retry-After`, as seen in the documentation linked above, would give a timestamp 48 seconds in the future, instead of the 1 second the limiter would need for the needed tokens to become available again

Jeroeny:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #40289
| License       | MIT

Reopened https://github.com/symfony/symfony/pull/47557 but as a bugfix PR targeting 5.4 instead. 

This implements `LimiterInterface->reserve()` for `SlidingWindowLimiter`. I'm not sure if the lack of implementation was due to time/scope or if it's actually not possible and my approach is incorrect. But I like to give it a try anyway. Perhaps @wouterj you could elaborate on that? 

The calculation does the following:

1. Calculate tokens to be released within this window. E.g. if 4 were used in the last window, at 50% into the current, 2 are still to be released.
2. Calculate the time-per-token within the remainder of the window. If the requested tokens will be available before the end of the current window, return the time-per-token * needed-tokens. 
3. Otherwise return time-per-token of the regular interval * needed-tokens(-after the current window). 

Some other things I've noticed in the RateLimiter:
- `FixedWindowLimiter` uses a `Window` class, whereas `SlidingWindowLimiter` uses a `SlidingWindow`. Shouldn't the first have a `FixedWindow` class?
- The `Window` class takes its `$windowSize` as argument, `SlidingWindow` does not. Perhaps the latter could also take this argument in this PR, since it's used in the calculation. But I guess those classes have to be backwards compatible.
- `SlidingWindow->getRetryAfter()` is no longer used and could be deprecated in favor of `calculateTimeForTokens`. Making it more like the fixed window.

### DIFF
--- a/src/Symfony/Component/RateLimiter/Policy/SlidingWindow.php
+++ b/src/Symfony/Component/RateLimiter/Policy/SlidingWindow.php
@@ -100,9 +100,26 @@ final class SlidingWindow implements LimiterStateInterface
         return (int) floor($this->hitCountForLastWindow * (1 - $percentOfCurrentTimeFrame) + $this->hitCount);
     }
 
-    public function getRetryAfter(): \DateTimeImmutable
+    public function calculateTimeForTokens(int $maxSize, int $tokens): float
     {
-        return \DateTimeImmutable::createFromFormat('U.u', sprintf('%.6F', $this->windowEndAt));
+        $remaining = $maxSize - $this->getHitCount();
+        if ($remaining >= $tokens) {
+            return 0;
+        }
+
+        $time = microtime(true);
+        $startOfWindow = $this->windowEndAt - $this->intervalInSeconds;
+        $timePassed = $time - $startOfWindow;
+        $windowPassed = min($timePassed / $this->intervalInSeconds, 1);
+        $releasable = max(1, $maxSize - floor($this->hitCountForLastWindow * (1 - $windowPassed))); // 2 * (0.7) =1, 3
+        $remainingWindow = $this->intervalInSeconds - $timePassed;
+        $needed = $tokens - $remaining;
+
+        if ($releasable >= $needed) {
+            return $needed * ($remainingWindow / max(1, $releasable));
+        }
+
+        return ($this->windowEndAt - $time) + ($needed - $releasable) * ($this->intervalInSeconds / $maxSize);
     }
 
     public function __serialize(): array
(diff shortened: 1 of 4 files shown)


Jeroeny:
@wouterj @Nyholm I see you have some history in this component. Maybe one of you would be interested to review this fix?

src/Symfony/Component/RateLimiter/Policy/SlidingWindow.php
```diff
@@ -100,9 +100,26 @@ public function getHitCount(): int
         return (int) floor($this->hitCountForLastWindow * (1 - $percentOfCurrentTimeFrame) + $this->hitCount);
     }
 
-    public function getRetryAfter(): \DateTimeImmutable
```
Nyholm, line 103: You cannot remove public methods. That is a BC break.

Nyholm (changes requested):
Thank you for this PR. 

This is a bit tricky part of the code base. I suggest some extra eyes to review this. 
Can you please add some more test for the new `calculateTimeForToken()`?

I also have to ask you to target 6.4 since this is a new feature and not a bugfix. The feature is "Add SlidingWindowLimiter::reserve()".

src/Symfony/Component/RateLimiter/Policy/SlidingWindow.php
Jeroeny: My bad, thought the `@internal` class allowed this.

Jeroeny:
Reopened as feature PR: https://github.com/symfony/symfony/pull/51676

PR #51593, opened 2023-09-07
[Messenger] Add the `--all` option to the `messenger:failed:remove` command

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | Todo

We have a development server (which we don't have direct access to the database). As this server serves as a test for our devs, error messages can accumulate in our failure transport. We wanted to use the `messenger:failed:remove` command to remove them, but unfortunately, we must provide ids individually. This is problematic as we have several hundreds of failed messages.

This PR adds the `--all` option to the command. This option **must** be used with the `--force` option (juste like `doctrine:schema:update --force` actually) to work. Example output:

```bash
$ bin/console messenger:failed:remove --all --force

...

Failed Message Details
======================
                                                                                                                        
 [WARNING] Message does not appear to have been sent to this transport after failing                                    
                                                                                                                        
 ------------ ------------------------- 
  Class        App\Message\YourMessage  
  Message Id   6                        
 ------------ ------------------------- 

 ! [NOTE] 4 messages were removed.                                                                                      
```

As you can see, you can of course still use the `--show-messages` option jointly.

### DIFF
--- a/src/Symfony/Component/Messenger/CHANGELOG.md
+++ b/src/Symfony/Component/Messenger/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
  * Deprecate `StopWorkerOnSignalsListener` in favor of using the `SignalableCommandInterface`
  * Add `HandlerDescriptor::getOptions`
  * Add support for multiple Redis Sentinel hosts
+ * Add `--all` option to the `messenger:failed:remove` command
 
 6.3
 ---
--- a/src/Symfony/Component/Messenger/Command/FailedMessagesRemoveCommand.php
+++ b/src/Symfony/Component/Messenger/Command/FailedMessagesRemoveCommand.php
@@ -20,7 +20,7 @@ use Symfony\Component\Console\Output\ConsoleOutputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\Messenger\Transport\Receiver\ListableReceiverInterface;
-use Symfony\Component\Messenger\Transport\Receiver\ReceiverInterface;
+use Symfony\Component\Messenger\Transport\Receiver\MessageCountAwareInterface;
 
 /**
  * @author Ryan Weaver <ryan@symfonycasts.com>
@@ -32,7 +32,8 @@ class FailedMessagesRemoveCommand extends AbstractFailedMessagesCommand
     {
         $this
             ->setDefinition([
-                new InputArgument('id', InputArgument::REQUIRED | InputArgument::IS_ARRAY, 'Specific message id(s) to remove'),
+                new InputArgument('id', InputArgument::OPTIONAL | InputArgument::IS_ARRAY, 'Specific message id(s) to remove'),
+                new InputOption('all', null, InputOption::VALUE_NONE, 'Remove all failed messages from the transport'),
                 new InputOption('force', null, InputOption::VALUE_NONE, 'Force the operation without confirmation'),
                 new InputOption('transport', null, InputOption::VALUE_OPTIONAL, 'Use a specific failure transport', self::DEFAULT_TRANSPORT_OPTION),
                 new InputOption('show-messages', null, InputOption::VALUE_NONE, 'Display messages before removing it (if multiple ids are given)'),
@@ -43,6 +44,10 @@ The <info>%command.name%</info> removes given messages that are pending in the f
     <info>php %command.full_name% {id1} [{id2} ...]</info>
 
 The specific ids can be found via the messenger:failed:show command.
+
+You can remove all failed messages from the failure transport by using the "--all" option:
+
+    <info>php %command.full_name% --all</info>
 EOF
             )
         ;
@@ -61,18 +66,32 @@ EOF
 
         $shouldForce = $input->getOption('force');
         $ids = (array) $input->getArgument('id');
-        $shouldDisplayMessages = $input->getOption('show-messages') || 1 === \count($ids);
-        $this->removeMessages($failureTransportName, $ids, $receiver, $io, $shouldForce, $shouldDisplayMessages);
+        $shouldDeleteAllMessages = $input->getOption('all');
 
-        return 0;
-    }
+        $idsCount = \count($ids);
+        if (!$shouldDeleteAllMessages && !$idsCount) {
+            throw new RuntimeException('Please specify at least one message id. If you want to remove all failed messages, use the "--all" option.');
+        } elseif ($shouldDeleteAllMessages && $idsCount) {
+            throw new RuntimeException('You cannot specify message ids when using the "--all" option.');
+        }
 
-    private function removeMessages(string $failureTransportName, array $ids, ReceiverInterface $receiver, SymfonyStyle $io, bool $shouldForce, bool $shouldDisplayMessages): void
-    {
+        $shouldDisplayMessages = $input->getOption('show-messages') || 1 === $idsCount;
+
         if (!$receiver instanceof ListableReceiverInterface) {
             throw new RuntimeException(sprintf('The "%s" receiver does not support removing specific messages.', $failureTransportName));
         }
 
+        if ($shouldDeleteAllMessages) {
+            $this->removeAllMessages($receiver, $io, $shouldForce, $shouldDisplayMessages);
+        } else {
+            $this->removeMessagesById($ids, $receiver, $io, $shouldForce, $shouldDisplayMessages);
+        }
+
+        return 0;
+    }
+
+    private function removeMessagesById(array $ids, ListableReceiverInterface $receiver, SymfonyStyle $io, bool $shouldForce, bool $shouldDisplayMessages): void
+    {
         foreach ($ids as $id) {
             $this->phpSerializer?->acceptPhpIncompleteClass();
             try {
@@ -98,5 +117,32 @@ EOF
                 $io->note(sprintf('Message with id %s not removed.', $id));
             }
         }
+    }
+
+    private function removeAllMessages(ListableReceiverInterface $receiver, SymfonyStyle $io, bool $shouldForce, bool $shouldDisplayMessages): void
+    {
+        if (!$shouldForce) {
+            if ($receiver instanceof MessageCountAwareInterface) {
+                $question = sprintf('Do you want to permanently remove all (%d) messages?', $receiver->getMessageCount());
+            } else {
+                $question = 'Do you want to permanently remove all failed messages?';
+            }
+
+            if (!$io->confirm($question, false)) {
+                return;
+            }
+        }
+
+        $count = 0;
+        foreach ($receiver->all() as $envelope) {
+            if ($shouldDisplayMessages) {
+                $this->displaySingleMessage($envelope, $io);
+            }
+
+            $receiver->reject($envelope);
+            ++$count;
+        }
+
+        $io->note(sprintf('%d messages were removed.', $count));
     }
 }
--- a/src/Symfony/Component/Messenger/Tests/Command/FailedMessagesRemoveCommandTest.php
+++ b/src/Symfony/Component/Messenger/Tests/Command/FailedMessagesRemoveCommandTest.php
@@ -12,9 +12,11 @@
 namespace Symfony\Component\Messenger\Tests\Command;
 
 use PHPUnit\Framework\TestCase;
+use Symfony\Component\Console\Exception\RuntimeException;
 use Symfony\Component\Console\Tester\CommandCompletionTester;
 use Symfony\Component\Console\Tester\CommandTester;
 use Symfony\Component\DependencyInjection\ServiceLocator;
+use Symfony\Component\Messenger\Bridge\Doctrine\Transport\DoctrineReceiver;
 use Symfony\Component\Messenger\Command\FailedMessagesRemoveCommand;
 use Symfony\Component\Messenger\Envelope;
 use Symfony\Component\Messenger\Exception\InvalidArgumentException;
@@ -207,7 +209,7 @@ class FailedMessagesRemoveCommandTest extends TestCase
         $globalFailureReceiverName = 'failure_receiver';
 
         $receiver = $this->createMock(ListableReceiverInterface::class);
-        $receiver->expects($this->once())->method('all')->with(50)->willReturn([
+        $receiver->expects($this->once())->method('all')->willReturn([
             Envelope::wrap(new \stdClass(), [new TransportMessageIdStamp('2ab50dfa1fbf')]),
             Envelope::wrap(new \stdClass(), [new TransportMessageIdStamp('78c2da843723')]),
         ]);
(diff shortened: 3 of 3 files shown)


kbond (approved):

src/Symfony/Component/Messenger/Command/FailedMessagesRemoveCommand.php
```diff
@@ -43,6 +45,10 @@ protected function configure(): void
     <info>php %command.full_name% {id1} [{id2} ...]</info>
 
 The specific ids can be found via the messenger:failed:show command.
+
+It is possible to remove all messages from the failure transport by using the "--all" option.
```
fabpot:
```suggestion
You can remove all messages from the failure transport by using the "--all" option (note that you also need the `--force` option):
```

src/Symfony/Component/Messenger/Command/FailedMessagesRemoveCommand.php
```diff
@@ -60,19 +66,38 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $receiver = $this->getReceiver($failureTransportName);
 
         $shouldForce = $input->getOption('force');
+
```
fabpot: Let's revert

src/Symfony/Component/Messenger/Command/FailedMessagesRemoveCommand.php
```diff
@@ -60,19 +66,38 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
         $ids = (array) $input->getArgument('id');
-        $shouldDisplayMessages = $input->getOption('show-messages') || 1 === \count($ids);
-        $this->removeMessages($failureTransportName, $ids, $receiver, $io, $shouldForce, $shouldDisplayMessages);
+        $shouldDeleteAllMessages = $input->getOption('all');
 
-        return 0;
-    }
+        $idsCount = \count($ids);
+        if (false === $shouldDeleteAllMessages && 0 === $idsCount) {
```
fabpot:
```suggestion
        if (!$shouldDeleteAllMessages && !$idsCount) {
```

src/Symfony/Component/Messenger/Command/FailedMessagesRemoveCommand.php
```diff
@@ -60,19 +66,38 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+            throw new RuntimeException('Message ids must be specified. If you want to remove all messages, use the "--all" option.');
+        } elseif (true === $shouldDeleteAllMessages && 0 !== $idsCount) {
```
fabpot:
```suggestion
        } elseif ($shouldDeleteAllMessages && $idsCount) {
```

fabpot (changes requested):

alexandre-daubois:
All comments addressed, thanks! :slightly_smiling_face:

src/Symfony/Component/Messenger/Command/FailedMessagesRemoveCommand.php
```diff
@@ -61,18 +67,36 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
         $ids = (array) $input->getArgument('id');
-        $shouldDisplayMessages = $input->getOption('show-messages') || 1 === \count($ids);
-        $this->removeMessages($failureTransportName, $ids, $receiver, $io, $shouldForce, $shouldDisplayMessages);
+        $shouldDeleteAllMessages = $input->getOption('all');
 
-        return 0;
-    }
+        $idsCount = \count($ids);
+        if (!$shouldDeleteAllMessages && !$idsCount) {
+            throw new RuntimeException('Message ids must be specified. If you want to remove all messages, use the "--all" option.');
```
nicolas-grekas: we try to not use "must", neither in doc nor in any messages

```suggestion
            throw new RuntimeException('Please specify at least one message id. If you want to remove all messages, use the "--all" option.');
```

src/Symfony/Component/Messenger/Command/FailedMessagesRemoveCommand.php
```diff
@@ -61,18 +67,36 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+        } elseif ($shouldDeleteAllMessages && $idsCount) {
+            throw new RuntimeException('You cannot specify message ids when using the "--all" option.');
+        }
+
+        if ($shouldDeleteAllMessages && !$shouldForce) {
+            throw new RuntimeException('You must use the "--force" option when using "--all" to confirm the removal of all failed messages in the transport.');
```
nicolas-grekas: it's not an error to not use `--force` currently
I don't think it should be one either with `--all`
instead, we should display an explanatory message

src/Symfony/Component/Messenger/Command/FailedMessagesRemoveCommand.php
```diff
@@ -61,18 +67,36 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+        $shouldDisplayMessages = $input->getOption('show-messages') || 1 === $idsCount;
 
-    private function removeMessages(string $failureTransportName, array $ids, ReceiverInterface $receiver, SymfonyStyle $io, bool $shouldForce, bool $shouldDisplayMessages): void
-    {
         if (!$receiver instanceof ListableReceiverInterface) {
             throw new RuntimeException(sprintf('The "%s" receiver does not support removing specific messages.', $failureTransportName));
         }
 
+        if ($shouldDeleteAllMessages) {
+            $this->removeEnvelopes($receiver->all(), $receiver, $io, $shouldDisplayMessages);
```
nicolas-grekas: removeAllMessages?

src/Symfony/Component/Messenger/Tests/Command/FailedMessagesRemoveCommandTest.php
```diff
@@ -253,4 +254,94 @@ public function testCompleteIdWithSpecifiedTransport()
...
+            new Envelope(new \stdClass()),
+        ];
+
+        $receiver->expects($this->once())->method('all')->willReturn($series);
+
+        $serviceLocator = $this->createMock(ServiceLocator::class);
+        $serviceLocator->expects($this->once())->method('has')->with($globalFailureReceiverName)->willReturn(true);
+        $serviceLocator->expects($this->any())->method('get')->with($globalFailureReceiverName)->willReturn($receiver);
+
+        $command = new FailedMessagesRemoveCommand(
```
nicolas-grekas: on one line

src/Symfony/Component/Messenger/Tests/Command/FailedMessagesRemoveCommandTest.php
```diff
@@ -253,4 +254,94 @@ public function testCompleteIdWithSpecifiedTransport()
...
+
+    public function testOptionAllIsNotSetNorIdsThrows()
+    {
+        $globalFailureReceiverName = 'failure_receiver';
+
+        $serviceLocator = $this->createMock(ServiceLocator::class);
+        $serviceLocator->expects($this->once())->method('has')->with($globalFailureReceiverName)->willReturn(true);
+        $serviceLocator->expects($this->any())->method('get')->with($globalFailureReceiverName)->willReturn($this->createMock(ListableReceiverInterface::class));
+
+        $command = new FailedMessagesRemoveCommand(
```
nicolas-grekas: one line

src/Symfony/Component/Messenger/Tests/Command/FailedMessagesRemoveCommandTest.php
```diff
@@ -253,4 +254,94 @@ public function testCompleteIdWithSpecifiedTransport()
...
+
+    public function testOptionAllIsSetWithoutForceThrows()
+    {
+        $globalFailureReceiverName = 'failure_receiver';
+
+        $serviceLocator = $this->createMock(ServiceLocator::class);
+        $serviceLocator->expects($this->once())->method('has')->with($globalFailureReceiverName)->willReturn(true);
+        $serviceLocator->expects($this->any())->method('get')->with($globalFailureReceiverName)->willReturn($this->createMock(ListableReceiverInterface::class));
+
+        $command = new FailedMessagesRemoveCommand(
```
nicolas-grekas: oneliner

src/Symfony/Component/Messenger/Tests/Command/FailedMessagesRemoveCommandTest.php
```diff
@@ -253,4 +254,94 @@ public function testCompleteIdWithSpecifiedTransport()
...
+
+    public function testOptionAllIsSetWithIdsThrows()
+    {
+        $globalFailureReceiverName = 'failure_receiver';
+
+        $serviceLocator = $this->createMock(ServiceLocator::class);
+        $serviceLocator->expects($this->once())->method('has')->with($globalFailureReceiverName)->willReturn(true);
+        $serviceLocator->expects($this->any())->method('get')->with($globalFailureReceiverName)->willReturn($this->createMock(ListableReceiverInterface::class));
+
+        $command = new FailedMessagesRemoveCommand(
```
nicolas-grekas: oneliner

src/Symfony/Component/Messenger/Command/FailedMessagesRemoveCommand.php
```diff
@@ -99,4 +123,22 @@ private function removeMessages(string $failureTransportName, array $ids, Receiv
             }
         }
     }
+
+    /**
+     * @param iterable<array-key, Envelope> $envelopes
+     */
+    private function removeEnvelopes(iterable $envelopes, ReceiverInterface $receiver, SymfonyStyle $io, bool $shouldDisplayMessages): void
```
nicolas-grekas: no need to pass $envelopes when we pass the $receiver

src/Symfony/Component/Messenger/Command/FailedMessagesRemoveCommand.php
alexandre-daubois: Got it :+1:

src/Symfony/Component/Messenger/Command/FailedMessagesRemoveCommand.php
alexandre-daubois: Updated to display an error and return code 1

alexandre-daubois:
Addressed all your comments Nicolas, thank you!

src/Symfony/Component/Messenger/Command/FailedMessagesRemoveCommand.php
```diff
@@ -61,18 +68,38 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
-    }
+        $idsCount = \count($ids);
+        if (!$shouldDeleteAllMessages && !$idsCount) {
+            throw new RuntimeException('Please specify at least one message id. If you want to remove all messages, use the "--all" option.');
+        } elseif ($shouldDeleteAllMessages && $idsCount) {
+            throw new RuntimeException('You cannot specify message ids when using the "--all" option.');
+        }
+
+        if ($shouldDeleteAllMessages && !$shouldForce) {
+            $io->error('Please provide the "--force" option when using "--all" to confirm the removal of all failed messages in the transport.');
```
nicolas-grekas: please check how we do when --force is not passed but and $id is: I don't think we display at the error level nor that we return 1, do we? also, we should display something useful, like the number or messages that would be deleted

src/Symfony/Component/Messenger/Command/FailedMessagesRemoveCommand.php
```diff
@@ -99,4 +126,19 @@ private function removeMessages(string $failureTransportName, array $ids, Receiv
             }
         }
     }
+
+    private function removeAllMessages(ListableReceiverInterface $receiver, SymfonyStyle $io, bool $shouldDisplayMessages): void
```
nicolas-grekas: we don't need a ListableReceiverInterface here, do we? we should accept a simple ReceiverInterface then

alexandre-daubois: We do, to be sure we can access to the `all()` method of `$receiver`

nicolas-grekas: what about using get in a loop? would that work?

src/Symfony/Component/Messenger/Command/FailedMessagesRemoveCommand.php
alexandre-daubois: Alright, I misunderstood the comment above and took example on another command like `TranslationUpdateCommand` that also defines the `--force` option. I updated accordingly. Without `--force`, the number of messages that would be removed is displayed as well as a hint that encourage to use `--force` with `--all` to remove them

src/Symfony/Component/Messenger/Command/FailedMessagesRemoveCommand.php
alexandre-daubois: It doesn't seem to work unfortunately. Using `get()` removes the message from the transport after the call, thus the message isn't present anymore on the next command run

src/Symfony/Component/Messenger/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Deprecate `StopWorkerOnSignalsListener` in favor of using the `SignalableCommandInterface`
  * Add `HandlerDescriptor::getOptions`
+ * Add the `--all` option to the `messenger:failed:remove` command
```
welcoMattic:
```suggestion
 * Add `--all` option to the `messenger:failed:remove` command
```

alexandre-daubois: Updated :slightly_smiling_face:

src/Symfony/Component/Messenger/Command/FailedMessagesRemoveCommand.php
```diff
@@ -43,6 +43,10 @@ protected function configure(): void
     <info>php %command.full_name% {id1} [{id2} ...]</info>
 
 The specific ids can be found via the messenger:failed:show command.
+
+You can remove all messages from the failure transport by using the "--all" option (note that you also need the "--force" option):
+
+    <info>php %command.full_name% --all --force</info>
```
nicolas-grekas: can we do this differently: when `--force` is NOT passed, we ask something like: `Do you want to permanently remove all (%d) messages?`
that would match the behavior when using ids and would allow removing this `--force` option from the help message

alexandre-daubois: I updated the code to match your idea. Currently, it displays the message count only if the receiver implements `MessageCountAwareInterface`. However, we may simplify this part by not displaying the message count when asking for confirmation. Suits me both ways

nicolas-grekas (approved):

nicolas-grekas:
Thank you @alexandre-daubois.

PR #51598, opened 2023-09-07
[Cache] fix using multiple Redis Sentinel hosts when the first one is not resolvable

### LINKED ISSUE #51570: [Cache] Redis Sentinel only attempting connection via first host with PHP Redis extension

\>= 5.4

### Description

#47003 added support for using multiple Redis Sentinel hosts. However, if the first Sentinel host cannot be resolved (e.g. because a DNS name cannot be resolved or because the server is not responding), a `\RedisException` is thrown and the code does not try to connect to the other Sentinel hosts.

In this code block, it tries to connect to Sentinel and resolve the master address:

https://github.com/symfony/symfony/blob/64303aa30a26c7d2b3b5319141222fe6f539915e/src/Symfony/Component/Cache/Traits/RedisTrait.php#L228-L232

The call to `getMasterAddrByName()` is throwing the `RedisException` and will prevent any attempt to connect to other Sentinel hosts.

This exception is thrown if the hostname cannot be resolved:

```
 RedisException {#407 ▼
  #message: "Redis server sentinel1:26379 went away"
  #code: 0
  #file: "/srv/www/vendor/symfony/cache/Traits/RedisTrait.php"
  #line: 231
  -previous: ErrorException {#408 ▼
    #message: "Warning: RedisSentinel::getMasterAddrByName(): php_network_getaddresses: getaddrinfo for sentinel1 failed: Name or service not known"
    #code: 0
    #file: "/srv/www/vendor/symfony/cache/Traits/RedisTrait.php"
    #line: 231
    #severity: E_WARNING
    trace: {▶}
  }
  trace: {▶}
}
```

And this exception if the server port is invalid (no previous exception in this case):

```
 RedisException {#408 ▼
  #message: "Redis server redis:16379 went away"
  #code: 0
  #file: "/srv/www/vendor/symfony/cache/Traits/RedisTrait.php"
  #line: 231
  trace: {▶}
}
```

### How to reproduce

You could use this environment variable / DSN to observe the behavior with invalid host names at first and valid host names afterwards:

```
REDIS_DSN=redis:?host[sentinel1:26379]&host[sentinel2:26379]&host[sentinel3:26379]&redis_sentinel=db
```

If you have a Redis Sentinel instance running, just add the host as second (or later) and observe that nothing is written into your Redis instance.

If you do not have a Redis Sentinel instance running, the same behavior can be verified by adding an `exit` statement after the if-block from the code snippet above. The code will not terminate, since it never reaches that statement.

### Possible Solution

The call to `getMasterAddrByName()` could be wrapped in a simple try...catch to suppress any error.

I am not sure if there are valid errors that should not be caught. In that case, the exception message could be checked for specific errors.

### Additional Context

I haven't investigated how this behaves with Relay.

digilist:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #51570
| License       | MIT
| Doc PR        | 

See ticket #51570 for details on this bugfix.

As mentioned in the ticket, I am not sure if it's wise to catch all exceptions or if it would be better to only check for specific ones. But since I cannot think about any reasons other than an unreachable host to raise any exception, I decided to catch all exceptions for now.

### DIFF
--- a/src/Symfony/Component/Cache/Traits/RedisTrait.php
+++ b/src/Symfony/Component/Cache/Traits/RedisTrait.php
@@ -217,8 +217,11 @@ trait RedisTrait
                     }
                     $sentinel = new \RedisSentinel($host, $port, $params['timeout'], (string) $params['persistent_id'], $params['retry_interval'], $params['read_timeout'], ...$extra);
 
-                    if ($address = $sentinel->getMasterAddrByName($params['redis_sentinel'])) {
-                        [$host, $port] = $address;
+                    try {
+                        if ($address = $sentinel->getMasterAddrByName($params['redis_sentinel'])) {
+                            [$host, $port] = $address;
+                        }
+                    } catch (\RedisException $e) {
                     }
                 } while (++$hostIndex < \count($hosts) && !$address);
 


src/Symfony/Component/Cache/Traits/RedisTrait.php
```diff
@@ -271,8 +274,10 @@ public static function createConnection(string $dsn, array $options = [])
         } elseif (is_a($class, \RedisArray::class, true)) {
             foreach ($hosts as $i => $host) {
                 switch ($host['scheme']) {
-                    case 'tcp': $hosts[$i] = $host['host'].':'.$host['port']; break;
-                    case 'tls': $hosts[$i] = 'tls://'.$host['host'].':'.$host['port']; break;
+                    case 'tcp': $hosts[$i] = $host['host'].':'.$host['port'];
+                        break;
+                    case 'tls': $hosts[$i] = 'tls://'.$host['host'].':'.$host['port'];
+                        break;
```
nicolas-grekas: let's keep all those on the same line and ignore fabbot on this one

digilist: Alright, I removed that commit.

nicolas-grekas:
Thank you @digilist.

PR #51601, opened 2023-09-08
[Mime] Allow to add some headers as a strings

### LINKED ISSUE #51584: symfony/mime does not allow to add some headers as a strings, only as arrays

6.3.3

### Description

Hi, currently `Symfony\Component\Mime\Header\Headers` does not allow to set Mime headers in general way: some headers must be strings, some - arrays, but this fact cannot be guessed from the method signature or documentation. We observed this issue only in runtime - when the `Symfony\Component\Mime\Header\Headers::addMailboxListHeader(): Argument #2 ($addresses) must be of type array, string given` exception reported.

We wrote a common method that sends email with simple signature `public function send(string $email, string $subject, string $content, array $headers = []): void;`.
We assumed that the `$headers` is `array<string, string>` and that array is completely controlled by a client of our library, but at the moment we should have knowledge "which data type do we need to use given header name" and that knowledge is hard-coded inside the private constant `Headers::HEADER_CLASS_MAP`.

I think it's better to hide that knowledge inside the lybrary code and make client's code more simplier and clear.

### How to reproduce

// First, run "composer require symfony/mime"
// Then, execute this file:

```php
<?php
require_once __DIR__.'/vendor/autoload.php';

use Symfony\Component\Mime\Email;

$headers = [
    'x-foo' => 'bar',
    'reply-to' => 'foo@bar.com',
];

$message = (new Email())
    ->from('bar@foo.com')
    ->to('baz@foo.com')
    ->subject('hello')
    ->html('')
;

foreach ($headers as $name => $value) {
    $message->getHeaders()->addTextHeader($name, $value);
    /* when $name == 'reply-to' the Exception "Argument #2 ($addresses) must be of type array, string given" will be thrown */
}
```

### Possible Solution

it's better to check the argument type inside the `addHeader` method when `$method === 'addMailboxListHeader'`, because it's not clear for the client that calls the `addHeader` which exact data type have to be used here.
Suggested variant is:
```php
      elseif ($method === 'addMailboxListHeader' && !\is_array($argument)) {
            $argument = [$argument];
        }
```

### Additional Context

_No response_

Oipnet:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      |  no
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #51584
| License       | MIT

When adding a MailboxListHeader header, the constructor expects an array as parameter.

To simplify the UX, we now allow you to pass a string to the addHeader method in the case of a MailBoxListHeader. The header value is then converted to an array.

### DIFF
--- a/src/Symfony/Component/Mime/Header/Headers.php
+++ b/src/Symfony/Component/Mime/Header/Headers.php
@@ -147,6 +147,8 @@ final class Headers
             $method = 'addTextHeader';
         } elseif ('addIdentificationHeader' === $method) {
             $method = 'addIdHeader';
+        } elseif ('addMailboxListHeader' === $method && !\is_array($argument)) {
+            $argument = [$argument];
         }
 
         return $this->$method($name, $argument, $more);
--- a/src/Symfony/Component/Mime/Tests/Header/HeadersTest.php
+++ b/src/Symfony/Component/Mime/Tests/Header/HeadersTest.php
@@ -69,6 +69,7 @@ class HeadersTest extends TestCase
     {
         $headers = new Headers();
         $headers->addHeader('from', ['from@example.com']);
+        $headers->addHeader('reply-to', 'reply@example.com');
         $headers->addHeader('return-path', 'return@example.com');
         $headers->addHeader('foo', 'bar');
         $headers->addHeader('date', $now = new \DateTimeImmutable());
@@ -76,6 +77,9 @@ class HeadersTest extends TestCase
 
         $this->assertInstanceOf(MailboxListHeader::class, $headers->get('from'));
         $this->assertEquals([new Address('from@example.com')], $headers->get('from')->getBody());
+
+        $this->assertInstanceOf(MailboxListHeader::class, $headers->get('reply-to'));
+        $this->assertEquals([new Address('reply@example.com')], $headers->get('reply-to')->getBody());
 
         $this->assertInstanceOf(PathHeader::class, $headers->get('return-path'));
         $this->assertEquals(new Address('return@example.com'), $headers->get('return-path')->getBody());


derrabus:
Thank you.

We don't allow merge commits. Please perform a rebase and squash your commits.

AurelienPillevesse:
In addition to the !is_array verification, verifying that it's a string should be added no?
What do you think about it ?

Oipnet:
@AurelienPillevesse 

When something other than a string is passed, the message seems rather explicit.

**_Argument #2 ($value) must be of type string, DateTime given_**

What do you think should happen if you don't have a string?

derrabus (approved):

derrabus:
> When something other than a string is passed, the message seems rather explicit.
> 
> **_Argument #2 ($value) must be of type string, DateTime given_**

This is fine, I think.

src/Symfony/Component/Mime/Header/Headers.php
```diff
@@ -147,6 +147,8 @@ public function addHeader(string $name, mixed $argument, array $more = []): stat
             $method = 'addTextHeader';
         } elseif ('addIdentificationHeader' === $method) {
             $method = 'addIdHeader';
+        } elseif ($method === 'addMailboxListHeader' && !\is_array($argument)) {
```
nicolas-grekas:
```suggestion
        } elseif ('addMailboxListHeader' === $method && !\is_array($argument)) {
```

nicolas-grekas (commented):
can you please update the PR title (+ commit title) so that it's turned into a positive sentence about what this does?

maxbeckers (approved):

nicolas-grekas:
Thank you @Oipnet.

PR #47561, opened 2022-09-13
[Cache] Allow colon (:) in cache key

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #45599
| License       | MIT
| Doc PR        |

PR #51603, opened 2023-09-08
[Cache] Allow to use namespace delimiter in cache key

### LINKED ISSUE #45599: [Cache] Allow colon in key key (with redis at least)
### Description

I noticed something strange with symfony cache. It [forbids](https://github.com/symfony/symfony/blob/6.1/src/Symfony/Contracts/Cache/ItemInterface.php#L43) to use a colon `:` in the key name.

But, namespacing keys with redis is (was?) a best practice, and so we lose some power when using [redis insight](https://redis.com/redis-enterprise/redis-insight/) for example.

In insight, I can profile redis, and it shows me what keys I use the most by grouping keys by pattern (`a:b:c:*`)

image

As you can see, the "TOP KEY PATTERNS" is poorly detailed.

1. Redis recommends to namespace key with a colon
2. Redis insight uses this pattern to analyze keys and group results
3. Symfony forbit me to use a colon in the but
4. And moreover, Symfony itself, in the very end does use a colon (between the app prefix seed and my key (`f-yOWMrlsA:*`on the screenshot)

---

So to me, we should simply remove this restriction 

Note: [psr](https://www.php-fig.org/psr/psr-6/) restricts this usage

> Key - [...] The following characters are reserved for future extensions and MUST NOT be supported by implementing libraries: {}()/\@:

But I don't use PSR, And I personally don't care about this PSR since I'm using Symfony contracts. 

### Example

_No response_

vtsykun:
| Q             | A
| ------------- | ---
| Branch?       | 6.4  
| Bug fix?      |  no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #45599
| License       | MIT
| Doc PR        | -

This PR allow colon char `:` in cache key. It may be useful for redis grouping keys by pattern without creating a many pools for each namespace. 

Difference between implementation https://github.com/symfony/symfony/pull/47561
- colon is only allowed for Symfony contracts cache, the PSR-6/16 adapters keeps this validation as before.
- Allow colon in namespace name too

### DIFF
--- a/src/Symfony/Component/Cache/CacheItem.php
+++ b/src/Symfony/Component/Cache/CacheItem.php
@@ -127,7 +127,7 @@ final class CacheItem implements ItemInterface
      *
      * @throws InvalidArgumentException When $key is not valid
      */
-    public static function validateKey($key): string
+    public static function validateKey($key, string $allowChars = null): string
     {
         if (!\is_string($key)) {
             throw new InvalidArgumentException(sprintf('Cache key must be string, "%s" given.', get_debug_type($key)));
@@ -135,8 +135,9 @@ final class CacheItem implements ItemInterface
         if ('' === $key) {
             throw new InvalidArgumentException('Cache key length must be greater than zero.');
         }
-        if (false !== strpbrk($key, self::RESERVED_CHARACTERS)) {
-            throw new InvalidArgumentException(sprintf('Cache key "%s" contains reserved characters "%s".', $key, self::RESERVED_CHARACTERS));
+        $reservedChars = null === $allowChars ? self::RESERVED_CHARACTERS : str_replace(str_split($allowChars), '', self::RESERVED_CHARACTERS);
+        if ('' !== $reservedChars && false !== strpbrk($key, $reservedChars)) {
+            throw new InvalidArgumentException(sprintf('Cache key "%s" contains reserved characters "%s".', $key, $reservedChars));
         }
 
         return $key;
--- a/src/Symfony/Component/Cache/Traits/AbstractAdapterTrait.php
+++ b/src/Symfony/Component/Cache/Traits/AbstractAdapterTrait.php
@@ -346,7 +346,7 @@ trait AbstractAdapterTrait
         if (\is_string($key) && isset($this->ids[$key])) {
             return $this->namespace.$this->namespaceVersion.$this->ids[$key];
         }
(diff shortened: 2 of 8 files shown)


src/Symfony/Component/Cache/CacheItem.php
```diff
@@ -127,16 +127,17 @@ public function getMetadata(): array
...
     {
         if (!\is_string($key)) {
             throw new InvalidArgumentException(sprintf('Cache key must be string, "%s" given.', get_debug_type($key)));
         }
         if ('' === $key) {
             throw new InvalidArgumentException('Cache key length must be greater than zero.');
         }
-        if (false !== strpbrk($key, self::RESERVED_CHARACTERS)) {
-            throw new InvalidArgumentException(sprintf('Cache key "%s" contains reserved characters "%s".', $key, self::RESERVED_CHARACTERS));
+        $reservedChars = null === $allowChars ? self::RESERVED_CHARACTERS : str_replace(str_split($allowChars), '', self::RESERVED_CHARACTERS);
```
nicolas-grekas, line 138: would it be possible to get rid of the calls to str_replace and str_split? this is on the hotpath and we'd better make this as fast as possible

src/Symfony/Component/Cache/Traits/AbstractAdapterTrait.php
```diff
@@ -346,7 +346,7 @@ protected function getId(mixed $key): string
         if (\is_string($key) && isset($this->ids[$key])) {
             return $this->namespace.$this->namespaceVersion.$this->ids[$key];
         }
-        \assert('' !== CacheItem::validateKey($key));
+        \assert('' !== CacheItem::validateKey($key, static::NS_SEPARATOR));
```
nicolas-grekas, line 349: this one is used for PSR-6 also, isn't it?

nicolas-grekas:
@vtsykun up to resume this PR?

dorrogeray:
It would be great to have this feature :+1: I can try to pick up this PR @vtsykun if you are too busy.

nicolas-grekas:
@dorrogeray follow up PR welcome then!

dorrogeray:
@nicolas-grekas I will work on it this week.

nicolas-grekas:
Closing in favor of #59813, thanks for giving this a try!

PR #47148, opened 2022-08-02
[WebProfilerBundle] Profiler redesign

javiereguiluz:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

This PR updates the design of the toolbar + profiler. The goal is to keep everything the same, but improve it to (1) make it look more modern; (2) improve accessibility (e.g. more contrast in some places); (3) make the design more efficient (rearranging some elements).

**How to test it in your projects**

```
# clone this repository branch locally
$ git clone -b update_profiler https://github.com/javiereguiluz/symfony symfony-profiler
$ cd symfony-profiler/

# make your project use the code of this PR
php link /path/to/your/symfony/6.x/project
```

The `link` command changes the Symfony dependencies in the `vendor/` dir of your project to create symlinks pointing to the code of this PR. See https://symfony.com/doc/current/contributing/code/pull_requests.html#use-your-branch-in-an-existing-project

**Some comments to reviewers**:

* Please, ignore the exception panel in the profiler and the exception page for now. I'll tackle those in a separate future PR.
* Try to review more thoroughly the panels which I don't use much: forms, serializer, messenger.
* I'll make a PR to Doctrine bundle repository to update the icons and other things in Doctrine panels.

-----

In this Twitter thread we explained the reasons behind some of these changes: https://twitter.com/symfony_en/status/1553998433760993280

For reference purposes, I'm posting here some selected screenshots too:

## Before / After in light mode

profiler-example-comparison-light-mode

## Before / After in dark mode

profiler-example-comparison-dark-mode

### Header

profiler-before-after-1

profiler-before-after-2

### Sidebar menu

profiler-before-after-3

The icons are from this project: https://github.com/tabler/tabler-icons. There are MIT licensed so I added their license as a file inside the icon directory.

### Design efficiency

profiler-before-after-4

profiler-before-after-5

### File source viewer

profiler-before-after-6

### Toolbar

profiler-before-after-7

PR #51621, opened 2023-09-10
[WebProfilerBundle] Replace last "old" icon + delete ICONS_LICENCE.txt

smnandre:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no 
| License       | MIT

Two (related) things in this PR 
* replace two icons used in the Messenger profiler panel with a Tabler-Icon one (following the conventions used since  the [profiler redesign](https://github.com/symfony/symfony/pull/47148) )
* delete the now outdated `ICONS_LICENCE.txt` file (as there is no concerned icons remaining, and there is now a [LICENCE.txt](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Icon/LICENSE.txt) file in the `Resources/views/Icon/` directory)

I targeted the 6.4 branch as there is no BC in code.. but i don't know how to handle a licence change 🤷🏻‍♂️

---

| Previous icons | After this PR | 
| - | - |
| previous-icons | tabler-icon |


src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/messenger.html.twig
```diff
@@ -131,10 +132,10 @@
                     {% if dispatchCall.exception is defined %}
                         <span class="label status-error">exception</span>
                     {% endif %}
-                    <a class="toggle-button">
-                        <span class="icon icon-close">{{ source('@WebProfiler/images/icon-minus-square.svg') }}</span>
-                        <span class="icon icon-open">{{ source('@WebProfiler/images/icon-plus-square.svg') }}</span>
-                    </a>
+                    <button class="btn btn-link toggle-button" type="button">
+                        <span class="icon icon-close">{{ source('@WebProfiler/Icon/chevron-down.svg') }}</span>
+                        <span class="icon icon-open">{{ source('@WebProfiler/Icon/chevron-down.svg') }}</span>
```
fabpot, line 137: One of them should be chevron-up, right?

javiereguiluz:
Simon, thanks for this improvement. I like it!

About the removal of the previous icon license, I'd say there's no problem.

-----

While reviewing this, I found two other occurrences of `toggle-button` in the `form` panel:

https://github.com/symfony/symfony/blob/d8a4b3c16a08be15aa24e0899bbdcb34c3fef002/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/form.html.twig#L408-L410

https://github.com/symfony/symfony/blob/d8a4b3c16a08be15aa24e0899bbdcb34c3fef002/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/form.html.twig#L500-L502

Would you have some time to try to update those too? If you can't, we can tackle that in another PR. Thanks.

smnandre:
> While reviewing this, I found two other occurrences of `toggle-button` in the `form` panel:

That's not the same icon: it uses a "[Tabler icon but in base64](https://github.com/symfony/symfony/blob/d8a4b3c16a08be15aa24e0899bbdcb34c3fef002/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/form.html.twig#L100)" ..  would you like to change it for a "chevron" ? 

(i rather ask if you did not see)

javiereguiluz:
@smnandre I didn't realize they were different icons. Sorry.

In any case, we should probably be consistent and use a chevron in all examples related to expand/collapse elements. Another thing; the current icon in form seem to NOT be working because I always see the `[+]`, no matter if it's expanded or collapsed. So, maybe leave it for another PR?

smnandre:
> @smnandre I didn't realize they were different icons. Sorry.

Don't be @javiereguiluz , i missed it entirely so .. :) 

Works for me to leave the rest for another PR..  i'm feeling we may raise other questions / needs.

fabpot (approved):

fabpot:
Thank you @smnandre.

PR #51625, opened 2023-09-11
Add AutowireCustom

olvlvl:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | --
| License       | MIT

Hi,

I want to use an attribute to qualify types. Something I would do with generics if we had them.

```php
<?php

final class SampleService
{
    /**
     * @param Model<int, Article> $model
     * @param Query<Article> $query
     */
    public function __construct(
        #[Record(Article::class)] public readonly Model $model,
        #[Record(Article::class)] public readonly Query $query,
    ) {
    }
}
```

Unfortunately, for `Record` to produce the correct service reference it needs to know about the property it's applied to. I created an interface `AutowireCustom`, called by `AutowirePass` to create the reference. Because it provides access to the container and the property, you can do whatever you want.

```php
<?php

#[Attribute(Attribute::TARGET_PARAMETER)]
class Record implements AutowireCustom
{
    public const SERVICE_PREFIX_FOR_MODEL = 'active_record.model.';
    public const SERVICE_PREFIX_FOR_QUERY = 'active_record.query.';

    /**
     * @param class-string<ActiveRecord> $activerecord_class
     */
    public function __construct(
        public readonly string $activerecord_class,
    ) {
    }

    public function buildReference(ContainerBuilder $container, ReflectionParameter $parameter): Reference
    {
        $type = $parameter->getType()
            ?? throw new LogicException("The parameter $parameter->name doesn't have a type");

        if (!$type instanceof \ReflectionNamedType) {
            throw new LogicException("Can only handle a single type");
        }

        $type = $type->getName();

        if (is_a($type, Model::class, true)) {
            $reference = $this->set_model_factory($container);
        } elseif (is_a($type, Query::class, true)) {
            $reference = $this->set_query_factory($container);
        } else {
            throw new LogicException("Unsupported type: $type");
        }

        return new Reference($reference);
    }

    private function set_model_factory(ContainerBuilder $container): Reference
    {
        $id = self::SERVICE_PREFIX_FOR_MODEL . $this->activerecord_class;
        $definition = (new Definition(Model::class))
            ->setFactory([ new Reference(ModelProvider::class), 'model_for_record' ])
            ->setArguments([ $this->activerecord_class ]);

        $container->setDefinition($id, $definition);

        return new Reference($id);
    }

    private function set_query_factory(ContainerBuilder $container): Reference
    {
        $id = self::SERVICE_PREFIX_FOR_QUERY . $this->activerecord_class;
        $definition = (new Definition(Query::class))
            ->setFactory([ $this->set_model_factory($container), 'query' ]);

        $container->setDefinition($id, $definition);

        return new Reference($id);
    }
}
```

Wdyt?


derrabus (changes requested):
Thank you. Please add tests.

src/Symfony/Component/DependencyInjection/Attribute/AutowireCustom.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace Symfony\Component\DependencyInjection\Attribute;
+
+use ReflectionParameter;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+interface AutowireCustom
```
nicolas-grekas: I would name this `AutowireableAttributeInterface`

src/Symfony/Component/DependencyInjection/Attribute/AutowireCustom.php
```diff
@@ -0,0 +1,12 @@
...
+{
+    public function buildReference(ContainerBuilder $container, ReflectionParameter $parameter): Reference;
```
nicolas-grekas: `public function getAutowireAttribute(): Autowire;`?

Can't the target use cases be handled by allowing one to cast one of their attributes into an instance of `Autowire`?
Attributes should be purely descriptive, so that attaching logic to them should be avoided.

nicolas-grekas (commented):
This opens for some coupling between your custom attributes and Symfony DI.
This also hides the fact that some attributes come with DI rules attached.
Currently, did you try using two attributes: one yours, two `Autowire`? Could that work for your use cases? Why not, if not?

olvlvl:
Hi,

@derrabus I'd rather see if this is viable before investing any more time.

> This opens for some coupling between your custom attributes and Symfony DI.
>  This also hides the fact that some attributes come with DI rules attached.

@nicolas-grekas I would argue coupling is similar between `extends Autowire` and `implement AutowireCustom`. One couples the constructor, the other a contract function. I think the constructor is actually worse in that regard because you cannot create the attribute without triggering a whole lot of things.

Regarding visibility, if I take the example from the [blog post](https://symfony.com/blog/new-in-symfony-6-3-dependency-injection-improvements#allow-extending-the-autowire-attribute), nothing tells me `Repository` is related to auto-wiring. It could be anything. Hence, it wouldn't make much difference whether it was extending `Autowire` or implementing `AutowireCustom`.

```php
/**
 * @param ObjectRepository<User> $repository
 */
public function __construct(
    #[Repository(User::class)] private ObjectRepository $repository
) {
}
```

> Currently, did you try using two attributes: one yours, two Autowire? Could that work for your use cases? Why not, if not?

At first, I just had `Record` for `Model`, but when `Query` came along, I thought it would be neat to be able to use it for both, instead of creating `QueryRecord` and `ModelRecord`. I didn't want to continue multiplying the number of attributes when they were referencing the same thing. So, I looked into it and I found it was super simple to implement. And here we are :)

ro0NL:
> #[Repository(User::class)] private ObjectRepository $repository

cant this be achieved at compilation level already with existing `#[Target]`? 

also, are Model/Query really meant to be a service? :thinking:

olvlvl:
@ro0NL Isn't `#[Target]` used for aliases? I found the following example in the docs. That would require creating aliases for all services, and that doesn't really help make things neat. I'd rather go through a factory.

```php
class MastodonClient
{
    public function __construct(
        #[Target('app.uppercase_transformer')]
        private TransformerInterface $transformer
    ){
    }
}
```

https://symfony.com/doc/current/service_container/autowiring.html

> also, are Model/Query really meant to be a service?

Sure, why not? Basically `Model` is for writing, and `Query` is for reading. But let's not get sidetracked there :)

nicolas-grekas:
Works for me. Let me know what you think of my other question.

olvlvl:
Hi @nicolas-grekas,

I pushed another commit with an interface that returns an `Autowire` instance.

The attribute looks like this now:

```php
<?php

// …

#[Attribute(Attribute::TARGET_PARAMETER)]
class Record implements AutowireableAttributeInterface
{
    /**
     * @param class-string<ActiveRecord> $activerecord_class
     */
    public function __construct(
        public readonly string $activerecord_class,
    ) {
    }

    public function getAutowireAttribute(ReflectionParameter $parameter): Autowire
    {
        $type = $parameter->getType()
            ?? throw new LogicException("The parameter $parameter->name doesn't have a type");

        if (!$type instanceof \ReflectionNamedType) {
            throw new LogicException("Can only handle a single type");
        }

        $type = $type->getName();

        if (is_a($type, Model::class, true)) {
            return $this->autowireModel();
        } elseif (is_a($type, Query::class, true)) {
            return $this->autowireQuery();
        }

        throw new LogicException("Unsupported type: $type");
    }

    private function autowireModel(): Autowire
    {
        return new Autowire(
            expression: sprintf(
                "service('%s').model_for_record('%s')",
                self::escape(ModelProvider::class),
                self::escape($this->activerecord_class),
            )
        );
    }

    private function autowireQuery(): Autowire
    {
        return new Autowire(
            expression: sprintf(
                "service('%s').query_for_record('%s')",
                self::escape(QueryProvider::class),
                self::escape($this->activerecord_class),
            )
        );
    }

    private static function escape(string $str): string
    {
        return str_replace('\\', '\\\\', $str);
    }
}
```

It works, but I'm not sure it's better. Also, it requires the use of `symfony/expression-language`.

I think the first proposal gives freedom to the developer to build the reference because of the access to the container. The second one confines it to the features offered by `Autowire`.

One flaw I see in the usage of the proposal (both versions) is that the attribute would only provide flexibility inside the same package, there's no simple way to add more bindings externally.

PS: I won't chain myself to a tree over this PR. If you think there's no potential in it, let's bin it and enjoy the weekend :)

stof:
To me, this goes too far into trying to expand the Autowire attribute. If you have such custom needs, a compiler pass should probably be used to implement the logic.

Here, you could have a compiler pass which would generic some bindings based on the `Record` attribute.

nicolas-grekas:
I think I agree with @stof.
Did you try using `ContainerBuilder::registerAttributeForAutoconfiguration()` BTW?

nicolas-grekas:
Another idea that could be more tailored: instead of returning an instance of `Autowire`, maybe we could cover your use case by returning an instance of Target?

`TargettingAttributeInterface::getTarget(): Target`

Can you please investigate this idea and the previous one to see how they could cover your use case?

nicolas-grekas:
Closing as discussed. I'd still be happy to read your answers to my questions above :)

olvlvl:
Hi,

I didn't know about `ContainerBuilder::registerAttributeForAutoconfiguration()`, but I cannot figure out how to make it work, the callable is never called. I traced `AttributeAutoconfigurationPass`, a parameter reference is created with this line:

```php
$this->{$symbol.'AttributeConfigurators'}[$attributeName] = $callable;
```

But in `processValue`, the condition `!$value instanceof Definition || !$value->isAutoconfigured()` is always true and the code never reaches `if ($this->parameterAttributeConfigurators) {`. 🤷

PR #51626, opened 2023-09-11
[TwigBridge][TwigBundle] Drop support for Twig 2

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

Let's make our lives easier and only support one major version of Twig in the future. Upgrading projects from Twig 2 to 3 shouldn't be much trouble anyway.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/composer.json
+++ b/src/Symfony/Bundle/FrameworkBundle/composer.json
@@ -70,7 +70,7 @@
         "symfony/uid": "^6.4|^7.0",
         "symfony/web-link": "^6.4|^7.0",
         "phpdocumentor/reflection-docblock": "^3.0|^4.0|^5.0",
-        "twig/twig": "^2.10|^3.0"
+        "twig/twig": "^3.0.4"
     },
     "conflict": {
         "doctrine/persistence": "<1.3",
--- a/src/Symfony/Component/VarDumper/composer.json
+++ b/src/Symfony/Component/VarDumper/composer.json
@@ -25,7 +25,7 @@
         "symfony/http-kernel": "^6.4|^7.0",
         "symfony/process": "^6.4|^7.0",
         "symfony/uid": "^6.4|^7.0",
-        "twig/twig": "^2.13|^3.0.4"
+        "twig/twig": "^3.0.4"
     },
     "conflict": {
         "symfony/console": "<6.4"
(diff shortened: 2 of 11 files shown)


fabpot (approved):

GromNaN (approved):

stof (commented):
The changelog file must be updated

src/Symfony/Bundle/FrameworkBundle/composer.json
```diff
@@ -70,7 +70,7 @@
         "symfony/uid": "^6.4|^7.0",
         "symfony/web-link": "^6.4|^7.0",
         "phpdocumentor/reflection-docblock": "^3.0|^4.0|^5.0",
-        "twig/twig": "^2.10|^3.0"
+        "twig/twig": "^3.0.4"
```
GromNaN, line 73: I think we should add a conflict rule, as is already done for symfony components in require-dev.

derrabus: Effectively, the http-kernel package does that for us currently. Shall I duplicate the rule in the bundles' composer.json file?

derrabus:
> The changelog file must be updated

Done for TwigBundle and TwigBridge. Not sure if I need to do that for the other bundles/components as well.

stof:
@derrabus I think it is OK as is.

src/Symfony/Component/VarDumper/composer.json
```diff
@@ -25,7 +25,7 @@
         "symfony/http-kernel": "^6.4|^7.0",
         "symfony/process": "^6.4|^7.0",
         "symfony/uid": "^6.4|^7.0",
-        "twig/twig": "^2.13|^3.0.4"
+        "twig/twig": "^3.0.4"
```
stof, line 28: should this add the conflict rule ?

derrabus: Not sure. It didn't have one before and I don't know why that is.

src/Symfony/Bundle/FrameworkBundle/composer.json
GromNaN: You're right, the constraint is already imposed by http-kernel, which is required by framework-bundle.

src/Symfony/Component/VarDumper/composer.json
GromNaN: The code that depends on twig is very robust.

https://github.com/symfony/symfony/blob/d8a4b3c16a08be15aa24e0899bbdcb34c3fef002/src/Symfony/Component/VarDumper/Caster/ExceptionCaster.php#L238

stof (approved):

nicolas-grekas:
Thank you @derrabus.

Issue #51563, opened 2023-09-05
Symfony 7.0 Type Declarations: Requesting Feedback from the Community


wouterj:
Symfony 7.0 will come with type declarations to almost all methods, parameters and properties. See [this blogpost](https://symfony.com/blog/symfony-7-0-type-declarations) for all information.

We acknowledge that introducing types impacts the whole Symfony open source ecosystem. As adding a type is a BC break, libraries are forced to release a new major version between now and December. Based on the experiences of bundles maintained by the Symfony team, we've postponed some impactful types to Symfony 8.0 already (increasing the timeframe for a new major version with two years).

We are open to, *on a case by case basis*, consider postponing more type declarations. If your library is blocked by one of the types we introduced, we would love to hear from you.

Test compatibility with 6.4/7.0 *as early as possible* by following these steps:

1. Update your Symfony dependencies to 6.4-dev, and temporarily install the `symfony/error-handler` component.
2. Run the [type patch script](https://symfony.com/doc/6.4/setup/upgrade_major.html#upgrading-native-return-types) *twice*:

   ```bash
   # the script relies on Composer's class map, check "exclude-from-classmap" in composer.json
   $ composer dump-autoload -o
   $ SYMFONY_PATCH_TYPE_DECLARATIONS=force=1 ./vendor/bin/patch-type-declarations
   $ SYMFONY_PATCH_TYPE_DECLARATIONS=force=1 ./vendor/bin/patch-type-declarations
   ```
3. The script patches any methods that are private, final, `@internal` or tests (i.e. methods that can safely be updated).

   The second run reports deprecations for methods that are not patched. **Please double check if the unpatched methods should not be final or internal** (i.e. should a user really extend or use this class?).
4. For the left-over deprecations, **check if the type exists in the [7.0 branch](https://github.com/symfony/symfony/tree/7.0)**. The 7.0 branch reflects the actual state, as postponed types will still trigger a deprecation in 6.4. You can suppress these deprections by adding the correct `@return` PHPdoc.

   If this is the case, please leave a comment with all details below.

Thanks for helping us towards type safety!

eurosat7:
Is there a rector for that? Should be doable.

driesvints:
Hey @wouterj. We've started testing Symfony 6.4 on Laravel v10 and Symfony v7.0 on upcoming Laravel v11. Currently seeing CI failures but we'll first weed out the ones we can solve ourselves before reporting back here.

driesvints:
[...]

PR #51630, opened 2023-09-11
Conflict symfony/validator 7.0 in symfony/doctrine-bridge 6.3

alexander-schranz:
| Q             | A
| ------------- | ---
| Branch?       | 5.4 / 6.3 / 6.4 -> Revert in 7.0
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #... 
| License       | MIT

Currently it is possible to have `doctrine-bridge` 6.3 and `validator` 7.0 installed which ends in the following error: 

> Testing PHP Fatal error:  Declaration of Symfony\Bridge\Doctrine\Validator\DoctrineInitializer::initialize(object $object) must be compatible with Symfony\Component\Validator\ObjectInitializerInterface::initialize(object $object): void in /home/runner/work/sulu/sulu/vendor/symfony/doctrine-bridge/Validator/DoctrineInitializer.php on line 34

The conflict can be removed then on the 7.0 branch where the `: void` return type is added.

Tested it also appears on 5.4 branch so conflict there would also be good, rebased so to 5.4.

You maybe ask why `symfony/doctrine-bridge` was in my example not on `7.0`. If a project still is using `doctrine/persistence` `2.0` instead of `3.0` it can not update the bridge to 6.4 or 7.0. So it could be a common error, to avoid this I suggest to add a conflict.

### DIFF
--- a/src/Symfony/Bridge/Doctrine/composer.json
+++ b/src/Symfony/Bridge/Doctrine/composer.json
@@ -63,7 +63,7 @@
         "symfony/proxy-manager-bridge": "<4.4.19",
         "symfony/security-bundle": "<5",
         "symfony/security-core": "<5.3",
-        "symfony/validator": "<5.4.25|>=6,<6.2.12|>=6.3,<6.3.1"
+        "symfony/validator": "<5.4.25|>=6,<6.2.12|>=6.3,<6.3.1,>=7.0"
     },
     "suggest": {
         "symfony/form": "",


src/Symfony/Bridge/Doctrine/composer.json
```diff
@@ -63,7 +63,7 @@
         "symfony/proxy-manager-bridge": "<4.4.19",
         "symfony/security-bundle": "<5",
         "symfony/security-core": "<5.3",
-        "symfony/validator": "<5.4.25|>=6,<6.2.12|>=6.3,<6.3.1"
```
alexander-schranz, line 66: Not sure if it would better to add a conflict to `symfony/validator` `7.0` as adding a conflict after a release in this case (5.4 and 6.3) will composer maybe just install a old version of the `doctrine-bridge` e.g. (5.4.28 or 6.3.4) without that conflict and then still run into the same error 🤔.

But it feels strange to add a conflict to `symfony/validator` as it has not even a optional dependency to the bridge.

derrabus (approved):

src/Symfony/Bridge/Doctrine/composer.json
stof: I think we should add this conflict rule here (where it is legitimate), but we should also add it in validator 7.0 to avoid the issue of installing validator 7.0 and an outdated doctrine-bridge.

nicolas-grekas:
This won't work: previous versions don't have the conflict so that composer might decide to install them instead of the latest 6.3.
Instead, what about removing the return type on 7.0 as advised in #51563?

src/Symfony/Bridge/Doctrine/composer.json
alexander-schranz: That would work I think. @nicolas-grekas what do you think about that?

stof:
@nicolas-grekas just removing the return type would fix compat with 7.0. But there is no conflict rule with 8.0 or 9.0 either so this cannot be the proper solution unless we avoid adding that return type forever.
See my proposal in the line comments for a solution that will work properly.

nicolas-grekas:
We bump deps in major version so that this guards naturally against such situations. Also, while we do seek for cross-majors compat between components, we don't between three majors. I would just add the return type in 8.0 and tell ppl that encounter the situation they have to update (6.3 will be EOLed since a long time) since this highly theoretical to me.

stof:
@nicolas-grekas this is a case of an _optional_ dependency between components, for which our existing usages of conflict rules only restrict one side of the range (equivalent to use an `optionalRequire` on `symfony/validator: >= 6.3.1` right now, so unbounded).

nicolas-grekas:
Sure, but still :) combining validator v8 with the bridge v6.3 should be an unsupported situation IMHO so I wouldn't care.

alexander-schranz:
For future we maybe could make sure to add all `optional` dependencies also to the `conflict` range to define the supported optional dependency versions. I did that for my search abstraction here: https://github.com/schranz-search/schranz-search/blob/48087364eb6b8b6e147bf475ed5bfb83bf77c88b/integrations/symfony/composer.json#L75-L85. But yeah that is just a solution for future incompatibility, for current I would go with @stof recommendation here if all others are fine with it: https://github.com/symfony/symfony/pull/51630#discussion_r1322636602

nicolas-grekas:
> I would go with @stof recommendation

We never do that AFAIK: making a component aware of a bridge (or if we do, that should be exceptionnal). That's broken package principles to me.

stof:
@nicolas-grekas we already made that in the past for conflict rules to handle the fact that we could not retro-actively add an upper bound to an optional dependency that was defined with a conflict rule allowing all future versions.

The way to avoid that (to respect package principles) would be that all our optional dependencies should forbid future major versions of that dependency with a conflict rule until we can ensure compatibility (basically avoiding unbounded optional dependencies)

nicolas-grekas:
That'd be too complex to manage I fear. I'm still up for the simple solution: revert the return type on v7, add it on v8 and don't care about what's going to be theoretical issue in 2/3 years. Alternatively, we could add the conflict in validator v7, but only that then. That's not my preference thought as it "pollutes" the deps of v7, while the point of a major is also to cleanup those things from the past.

src/Symfony/Bridge/Doctrine/composer.json
```diff
@@ -63,7 +63,7 @@
...
+        "symfony/validator": "<5.4.25|>=6,<6.2.12|>=6.3,<6.3.1,>=7.0"
```
derrabus, line 66:
```suggestion
        "symfony/validator": "<5.4.25|>=6,<6.2.12|>=6.3,<6.3.1|>=7.0"
```

derrabus (changes requested):

nicolas-grekas:
I feel like y'all prefer the conflict-based approach, see https://github.com/symfony/symfony/pull/51634

PR #51638, opened 2023-09-12
[FrameworkBundle] [Test] add token attributes in `KernelBrowser::loginUser()`

Valmonzo:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | ~
| License       | MIT
| Doc PR        | TODO

We have an advanced use case where we define custom attributes in tokens from success handlers.
Then, we use them for access controls.
Currently we cannot define those attributes when testing and using the `loginUser()` method, and so we needed to copy paste the whole logic to do it.

What do you think about this feature?

I did not add tests because the method does not have any yet, should we add some?

It's my first contribution, thanks to my brother @HeahDude for his help 🙏🏽

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/KernelBrowser.php
+++ b/src/Symfony/Bundle/FrameworkBundle/KernelBrowser.php
@@ -107,12 +107,15 @@ class KernelBrowser extends HttpKernelBrowser
     }
 
     /**
-     * @param UserInterface $user
+     * @param UserInterface        $user
+     * @param array<string, mixed> $tokenAttributes
      *
      * @return $this
      */
-    public function loginUser(object $user, string $firewallContext = 'main'): static
+    public function loginUser(object $user, string $firewallContext = 'main'/* , array $tokenAttributes = [] */): static
     {
+        $tokenAttributes = 2 < \func_num_args() ? func_get_arg(2) : [];
+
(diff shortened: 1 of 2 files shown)


nicolas-grekas (approved):

kbond (approved):

src/Symfony/Bundle/FrameworkBundle/KernelBrowser.php
```diff
@@ -107,12 +107,15 @@ public function enableReboot()
...
-     * @param UserInterface $user
+     * @param UserInterface        $user
+     * @param array<string, mixed> $tokenAttributes
      *
      * @return $this
      */
-    public function loginUser(object $user, string $firewallContext = 'main'): static
+    public function loginUser(object $user, string $firewallContext = 'main'/* , array $tokenAttributes = [] */): static
     {
+        $tokenAttributes = \func_num_args() > 2 ? func_get_arg(2) : [];
```
chalasr:
```suggestion
        $tokenAttributes = 2 < \func_num_args() ? func_get_arg(2) : [];
```

chalasr (approved):

chalasr:
Thank you @Valmonzo.

PR #51644, opened 2023-09-13
[String] Fix AbstractUnicodeString

### LINKED ISSUE #51643: [String] Transliteration RU

6.3.4

### Description

The current transliteration functionality is producing an incorrect output when transliterating the name "Татьяна". The expected result should be "Tatiana", but instead, the transliterated output is "Tat-ana".

### How to reproduce

```php
new AsciiSlugger())->slug('Татьяна');
// Tat-ana
```

### Possible Solution

_No response_

### Additional Context

_No response_

tpatartmajeur:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #51643 
| License       | MIT

This PR add `ь` in TRANSLIT_FROM and `i` in TRANSLIT_TO respecting the correspondence


nicolas-grekas:
Looking at https://github.com/unicode-org/cldr/blob/main/common/transforms/Cyrillic-Latin.xml
`ь` is transliterated to a single quote `'`, which is then turned into a `-` for slugging.
According to CLDR, `татьяна` is thus turned into `Tatʹâna` (which is then slugged into Tat-ana).
Adding one this letter in the map solves the issue but I don't know if the `i` is the correct transliteration for `ь` in all situations. It's not the case according to the CLDR database...
Do you have more background on this topic?

tpatartmajeur:
I don't have any more knowledge on the subject, I encountered the issue after migrating from Behat/Transliterator to the String component.

```php
use Behat\Transliterator\Transliterator;
Transliterator::transliterate($string)// tatiana

```

stof:
@tpatartmajeur the Behat transliterator is based on data from https://metacpan.org/pod/Text::Unidecode

I would trust more the CLDR data than this quick&dirty Perl module as having the correct transliteration.

nicolas-grekas:
Dunno where the Behat transliteration rule comes from
https://en.wikipedia.org/wiki/BGN/PCGN_romanization_of_Russian gives more rules and ``ь` is described also as a single quote is most cases.
Closing therefor.

nicolas-grekas:
After further research, the rule is like follows. But the CLDR mapping doesn't describe these rules.

> The transliteration of the Cyrillic letter "я" as "ya" is context-dependent.
> 
> * At the beginning of a word or following a vowel, a soft or hard sign (ь, ъ), or another consonant than "й," it is usually transliterated as "ya." For example:
>  "Яблоко" becomes "Yabloko."
>  "Татьяна" becomes "Tatyana."
> 
> * When "я" follows the consonant "й," it is typically transliterated simply as "a," with the "й" becoming "y," to avoid the redundancy of "yya."
>For example, "майя" would be transliterated as "mayya."
> 
> * When "я" appears after a consonant within a word, the "y" component often serves to indicate that the preceding consonant is palatalized (softened), and the "a" represents the vowel sound. In such cases, some transliteration systems may omit the "y," and the palatalization may or may not be explicitly indicated.
>For example, in the name "Петя" ("Petya"), the "y" indicates the softening of the "t" sound, but some systems might transliterate it simply as "Peta."
> 
> So the rule is context-dependent, and the choice of transliteration can depend on the system being used and the need to indicate palatalization or other phonetic features.

tpatartmajeur:
Thank you @nicolas-grekas for these clarifications. So if I understand correctly, the transliteration of Татьяна could (should?) be Tatyana. Do we need to adapt the String component?

nicolas-grekas:
If we can implement those rules, I guess yes.

PR #51649, opened 2023-09-13
[FeatureFlags] Propose a new component

Neirda24:
| Q             | A
| ------------- | ---
| Branch?       | 8.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

co-authored with the amazing help of @Jean-Beru .

# FeatureFlags Component

## Introduction

There is no simple way (yet) to enable part of the source to be executed depending on certain context.
This PR tries to solve this issue by integrating some easy way to check is this or that feature should be enabled.

First check out [Martin Fowler's article](https://martinfowler.com/articles/feature-toggles.html#De-couplingDecisionPointsFromDecisionLogic) about different use cases for feature toggling.
He categorize feature toggling like this :

- Experiment: show a beta version of your website to users who subscribed to.

- Release: deploy a new version of your code but keep the old one to compare them easily and rollback quickly if needed or control when a feature is released.

- Permission: grant access to a feature for paid accounts using the Security component.

- Ops: remove access to a consuming feature if server ressources are low (a k.a. kill switch). During Black Friday for example it is common to deactivate certain features for Ops because the load will be on other pages.

There are already some libraries / bundles out there but they either lack extensibility or are locked in to a SAAS tool (Unleash, Gitlab (which uses Unleash), ...).

- https://github.com/akeneo/pim-community-dev/tree/master/src/Akeneo/Platform/Bundle/FeatureFlagBundle
- https://medium.com/devwarlocks/implementing-feature-toggles-in-symfony-920993afd2b4
- https://smaine-milianni.medium.com/feature-flag-and-strategy-pattern-with-the-symfony-framework-7863ecc9556a

## Proposal
It is a simple Feature class that have a single Strategy instance. But strategy could be anything. At the moment this PR already provides a few. There can (should ?) be many more.
At the moment we simply check in headers, query string, with a start / end date.
There are also three combinators : Unanimous, Affirmative & Priority strategies which are heavily inspired by the AccessDecisionManager.

### As a standalone component
```php
use Symfony\Component\FeatureFlags\Feature;
use Symfony\Component\FeatureFlags\FeatureChecker;
use Symfony\Component\FeatureFlags\FeatureCollection;
use Symfony\Component\FeatureFlags\Strategy\AffirmativeStrategy;
use Symfony\Component\FeatureFlags\Strategy\DateStrategy;
use Symfony\Component\FeatureFlags\Strategy\RequestHeaderStrategy;
use Symfony\Component\FeatureFlags\Strategy\StrategyInterface;
use Symfony\Component\FeatureFlags\StrategyResult;
use Psr\Clock\ClockInterface;

$clock = new class() implements ClockInterface {
    public function now(): DateTimeImmutable
    {
        return new DateTimeImmutable();
    }
};

$collection = FeatureCollection::withFeatures([
    new Feature(
        name: 'christmas-banner',
        description: 'Special banner for christmas',
        default: false,
        strategy: new DateStrategy(
            $clock,
            since: DateTimeImmutable::createFromFormat('Y/m/d', $clock->now()->format('Y') . '/12/20'),
            until: DateTimeImmutable::createFromFormat('Y/m/d', $clock->now()->format('Y') . '/12/25'),
            includeSince: true,
            includeUntil: true,
        ),
    ),
    new Feature(
        name: 'some-beta-feature',
        description: 'Has access to the "some beta feature".',
        default: false,
        strategy: new class() implements StrategyInterface {
            public function compute(): StrategyResult
            {
                // TODO: Fetch logic from database for example.
                return StrategyResult::Grant;
            }
        },
    ),
    new Feature(
        name: 'black-friday-banner',
        description: 'New banner for the incoming black friday with preview.',
        default: false,
        strategy: new AffirmativeStrategy([
            new DateStrategy($clock, new \DateTimeImmutable('14 Nov. 2023'), new \DateTimeImmutable('24 Nov. 2023'), true, false),
            new RequestHeaderStrategy('X-Preview-Black-Friday'),
        ]),
    ),
]);

$checker = new FeatureChecker(
    features: $collection,
    whenNotFound: true, // What happens when the feature does not exist
);

foreach ($collection as $feature) {
    echo sprintf("| %-20s | %10s |\n", $feature->getName(), $checker->isEnabled($feature->getName()) ? 'enabled' : 'disabled');
}
```

### As a bundle

Here are some configuration examples:

**Simple example**
```yaml
# features.yaml
framework:
    feature_flags:
        strategies:
            -
                name: 'my_query.query.feature-strategy'
                type: 'request_query'
                with:
                    name: 'my_query'
        features:
            -
                name: 'my-feature'
                description: 'Desc feature 1'
                default: true
                strategy: 'my_query.query.feature-strategy'

```

**Advanced example**
```yaml
# features.yaml
framework:
    feature_flags:
        strategies:
            -
                name: 'my_query.query.feature-strategy'
                type: 'request_query'
                with:
                    name: 'my_query'
            -
                name: 'some_attribute.request_stack.feature-strategy'
                type: 'request_attribute'
                with:
                    name: 'some_attribute'
            -
                name: 'my-custom.feature-strategy'
                type: 'Some\Service\Id'
            -
                name: 'priority_1.feature-strategy'
                type: 'priority'
                with:
                    strategies: ['my_query.query.feature-strategy', 'some_attribute.request_stack.feature-strategy', 'my-custom.feature-strategy']
        features:
            -
                name: 'my-feature'
                description: 'Desc feature 1'
                default: true
                strategy: 'priority_1.feature-strategy'

```

And here is how to use them:

**Inside any registered service**

```php
use Symfony\Component\FeatureFlags\FeatureCheckerInterface;

class MyService
{
    public function __construct(private FeatureCheckerInterface $featureChecker) {}

    public function doSomething()
    {
        if ($this->featureChecker->isEnabled('my-feature')) {
            // do stuff
        }
    }
}
```

**Using Twig**
```twig
{% if is_feature_enabled('my-feature') %}
    {# do stuff #}
{% endif %}
```

**Using the Router component**
```yaml
my_route:
    # ...
    condition: 'isFeatureEnabled("myFeature")'
```
or using attributes
```php
#[Route(condition: 'isFeatureEnabled("myFeature")')]
```

To trigger those feature, using the Simple Configuration example above, you simply need to add `my_query` parameter and
set it to a truthy value (see [FILTER_VALIDATE_BOOL](https://www.php.net/manual/fr/filter.filters.validate.php#:~:text=FILTER_VALIDATE_BOOLEAN%2C%20FILTER_VALIDATE_BOOL)),
for example: http://my-website?my_query=1.

#### Toolbar (feature "beta")
toolbar_false
toolbar_true

#### Panel
panel_false
panel_true

#### Abstain (fallback to default)
default

#### Unknown (fallback to default)
unknown

#### Nested feature
nested_false
nested_true

#### List all feature flags
List all features with a duplication

#### Details for a given feature
Details of a feature
In case other providers also provides it
Details of a feature in case of duplication

#### Not found
Feature not found src=
With suggestions :
Not found with a suggestion

### Have some kind of distribution system
For example "Enable this feature for 30% of visiting users.".
It requires the component to "calculate" a unique ID per visitor and check if a decision was already made (needs a storage).
* % of visiting users
* % of requests
* Gradual addition of users (10% for a month, 50% for next 6 months, 100% at 1 year)
* ...

### Full integration with any SAAS
Gitlab, like other SAAS, offers a way to enable / disable feature flags. It would be great to have it fully integrated using a simple `composer install`.

Some SAAS tools :
* Gitlab
* Unleash
* LaunchDarkly

### Add more strategies
* PHP-ext availability
* Package version (eg Container::willBeAvailable())
* PHP Version
* User property (using the Security component)
* Firewall in use (using the Security component)
* ROLES / IS_GRANTED (using the Security component)
* Path prefix
* RequestFormat. Ex: json, html, other
* Reverse proxy usage (Cloudflare / Varnish / Traefik / Caddy / Other)

### Ease the configuration / debugging with a string syntax
A way to configure it like so : `(header("plop") == "true" && getenv("MY_ENV") == true) || request_attribute("attr") == "plop"` or dumping a whole Feature configuration as the previous string to easily understand the logic.

### Distinguish compile & runtime available features
* getenv might be resolvable at compile time
  * :warning: might differ with CGI
* Headers will be resolvable at runtime only
* For a feature to be available at compile time all its inner strategies must be as well

### Deeper integrations
* [Twig] Add a twig test `{% if 'my-feature' is enabled %}` (thanks to @ajgarlag )

### Make tests easier
* MockChecker

## Pending questions
* Change `$whenNotFound` to `bool|StrategyInterface $default` to apply a default strategy when feature not found.

## TODO
* [x] Add a debug command
* [ ] Merge in framework bundle
* [ ] Write documentation
* [ ] Add a flex recipe


src/Symfony/Bundle/FeatureToggleBundle/README.md
```diff
@@ -0,0 +1,18 @@
+ToggleFeatureBundle
```
yceruto: it's `FeatureToggleBundle` in code

src/Symfony/Bundle/FeatureToggleBundle/README.md
```diff
@@ -0,0 +1,18 @@
...
+===================
+
+ToggleFeatureBundle provides a tight integration of the Symfony ToggleFeature
```
yceruto: it's `FeatureToggle` component in code

src/Symfony/Bundle/FeatureToggleBundle/Resources/config/feature.php
```diff
@@ -0,0 +1,25 @@
...
+namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+use Symfony\Component\FeatureToggle\FeatureChecker;
+use Symfony\Component\FeatureToggle\FeatureCheckerInterface;
+use Symfony\Component\FeatureToggle\FeatureCollection;
+
+return static function (ContainerConfigurator $container) {
+    $services = $container->services();
+
+    $services->set('toggle_feature.feature_collection', FeatureCollection::class)
```
yceruto: the service prefix should match the component name, i.e. `feature_toggle.feature_collection` same everywhere

yceruto (changes requested):
Just a quick look so far, thanks for this proposal!

src/Symfony/Bundle/FeatureToggleBundle/FeatureToggleBundle.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+namespace Symfony\Bundle\FeatureToggleBundle;
+
+use Symfony\Bundle\FeatureToggleBundle\DependencyInjection\CompilerPass\DebugPass;
+use Symfony\Bundle\FeatureToggleBundle\DependencyInjection\CompilerPass\FeatureCollectionPass;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\HttpKernel\Bundle\Bundle;
+
+final class FeatureToggleBundle extends Bundle
```
stof: having a new bundle does not match the PR description where you show it being configured as part of the FrameworkBundle configuration.

stof (changes requested):

src/Symfony/Component/FeatureToggle/FeatureCollection.php
```diff
@@ -0,0 +1,98 @@
...
+        $this->compile();
+        return array_key_exists($id, $this->features);
+    }
+
+    /**
+     * @throws FeatureNotFoundException If the feature is not registered in this provider.
+     */
+    public function get(string $id): Feature
+    {
+        $this->compile();
```
stof: This implementation compiling all features from various providers on usage makes it impossible to build any optimization where we would build the list in a compiler pass during the cache warmup.
It also forces to instantiate all features to be able to use a single one.

IMO, a better API would be to inject a PSR ContainerInterface in the constructor. This way, the wiring in the framework could inject a ServiceLocator, which would lazy-load the Feature object that we need and no other.

For the iterator, I see 2 options:
- don't make it iterable at all (if we can)
- inject a list of feature ids in that class so that it could iterate over them and lazy-load the Feature objects associated with them.

src/Symfony/Bundle/FeatureToggleBundle/FeatureToggleBundle.php
Neirda24: Indeed. It was something I wanted to discuss in the PR actually. I first worked on this as a separate package a few years ago and I've reworked it lately. I was wondering if it should be merged onto FrameworkBundle or should it be a separate one ? Should I update the description while it is decided ?

src/Symfony/Component/FeatureToggle/FeatureCollection.php
```diff
@@ -0,0 +1,98 @@
...
+        return $this->features[$id] ?? throw new FeatureNotFoundException($id);
+    }
+
+    /**
+     * @return \ArrayIterator<int, Feature>
+     */
+    public function getIterator(): \ArrayIterator
```
stof: I would use `\Traversable` as return type. There is no reason to enforce using an ArrayIterator here (maybe we would want to use a Generator in the future)

src/Symfony/Component/FeatureToggle/README.md
```diff
@@ -0,0 +1,75 @@
+ToggleFeature Component
+=======================
+
+The ToggleFeature component provides TODO
+
+**This Component is experimental**.
```
stof: Experimental components are forbidden in LTS versions, so this PR must target the 7.0 branch instead.

src/Symfony/Bundle/FeatureToggleBundle/Resources/config/feature.php
Neirda24: done

src/Symfony/Bundle/FeatureToggleBundle/README.md
Neirda24: done

src/Symfony/Bundle/FeatureToggleBundle/README.md
Neirda24: done

src/Symfony/Component/FeatureToggle/Strategy/AffirmativeStrategy.php
```diff
@@ -0,0 +1,48 @@
...
+
+            if (StrategyResult::Deny === $innerResult) {
+                $result = StrategyResult::Deny;
+            }
+        }
+
+        return $result;
+    }
+
+    public function getInnerStrategies(): iterable
```
stof: There is no reason to have such public getter

src/Symfony/Component/FeatureToggle/Strategy/OuterStrategiesInterface.php
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\FeatureToggle\Strategy;
+
+interface OuterStrategiesInterface extends StrategyInterface
```
stof: What is the goal of those interfaces exposing inner implementation details ?

src/Symfony/Component/FeatureToggle/Strategy/RequestHeaderStrategy.php
```diff
@@ -0,0 +1,32 @@
...
+final class RequestHeaderStrategy implements StrategyInterface
+{
+    public function __construct(
+        private readonly string $headerName,
+    ) {
+    }
+
+    public function compute(): StrategyResult
+    {
+        if (!array_key_exists('HTTP_'.$this->headerName, $_SERVER)) {
```
stof: This should not be based on `$_SERVER` but integration with the HttpFoundation component instead, which would make it work in all runtimes in which Symfony could be used.

src/Symfony/Component/FeatureToggle/StrategyResult.php
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Symfony\Component\FeatureToggle;
+
+enum StrategyResult
+{
+    case Grant;
+    case Deny;
+    case Abstain;
+
+    public function isEnabled(bool $fallback): bool
```
stof: I don't think this should be a method on the `StrategyResult` as this would encourage strategies to rely on that method while they should not. As this is only needed inside `Feature::isEnabled` AFAICT, I would put the logic there instead.

src/Symfony/Component/FeatureToggle/Strategy/DateStrategy.php
```diff
@@ -0,0 +1,57 @@
...
+            if ($this->includeUntil && $this->until < $now) {
+                return StrategyResult::Deny;
+            }
+
+            if (!$this->includeUntil && $this->until <= $now) {
+                return StrategyResult::Deny;
+            }
+        }
+
+        return StrategyResult::Grant;
```
stof: should this grant or abstain ?

src/Symfony/Component/FeatureToggle/README.md
```diff
@@ -0,0 +1,75 @@
...
+    // Use the new feature
+} else {
+    // Use the legacy code
+}
+```
+
+Available strategies
+--------------------
+
+**AffirmativeStrategy** : Takes a list of `StrategyInterface` and stops at the first `Grant`.
```
stof: this strategy is currently implementing a `OR`. But there is no implementation of a `AND`

src/Symfony/Component/FeatureToggle/Debug/TraceableStrategy.php
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\FeatureToggle\Debug;
+
+use Symfony\Bundle\FeatureToggleBundle\DataCollector\FeatureCheckerDataCollector;
```
stof: the component cannot depend on the bundle.

src/Symfony/Bundle/FeatureToggleBundle/DataCollector/FeatureCheckerDataCollector.php
```diff
@@ -0,0 +1,156 @@
...
+        private readonly FeatureCollection $featureCollection,
+    ) {
+        $this->data = ['features' => [], 'toggles' => []];
+        $this->currentToggle = new \SplStack();
+        $this->currentCompute = new \SplStack();
+    }
+
+    public function collect(Request $request, Response $response, \Throwable $exception = null): void
+    {
+        foreach ($this->featureCollection as $feature) {
```
stof: Should we really iterate over all registered features to collect them ? Those don't represent what happens during that request (and this will force to instantiate them all).

To me, this looks like a use case for a `debug:feature-toggles` command allowing to show the configuration of toggles (or maybe `debug:container --tag="feature_toggle.feature"` would already be enough)

src/Symfony/Bundle/FeatureToggleBundle/DataCollector/FeatureCheckerDataCollector.php
```diff
@@ -0,0 +1,156 @@
...
+        ];
+    }
+
+    public function lateCollect(): void
+    {
+        $this->data = $this->cloneVar($this->data);
+    }
+
+    /**
+     * @return list<FeatureType>|Data
```
stof: using a non-prefixed `@return` using a phpstan type alias is a no-go as other tools would interpret that as returning a list of objects of a non-existent class FeatureType.

and in Symfony, we don't use phpstan type aliases.

stof (changes requested):

stof:
If you want a string syntax, this should rely on the ExpressionLanguage component IMO.

src/Symfony/Bundle/FeatureToggleBundle/FeatureToggleBundle.php
stof: This should go in FrameworkBundle, like all core components except Security ones.

smnandre:
Having recently worked on a similar (but much more limited) subject.. here are some questions :) 

Disclaimner: all are geniune questions and i have no strong position on any :) 

* should Feature be an object ? 
* is_feature_enabled(..) indicate a unique name reference. how would you handle priorities / conflicts ?
* if it's an object... should "isEnabled" be in it ? debug-aside, what is the usage of a non-enabled feature ?
* is the description only a debug utility ? if so, maybe it's not usefull here. if yes how would you handle translations ?
* should Feature "know"/"contains" its strategy ?

src/Symfony/Component/FeatureToggle/Strategy/OuterStrategiesInterface.php
Neirda24: It was based on this https://www.php.net/manual/en/iteratoriterator.getinneriterator.php . Makes it both clear that it "composes" other strategies and easier to debug / trace. How would you suggest to change it ?

src/Symfony/Component/FeatureToggle/Strategy/RequestHeaderStrategy.php
Neirda24: Agreed. But this is for standalone purposes. I was thinking of having both. One that is available in teh based component and the other one for the bundle. WDYT ?

src/Symfony/Component/FeatureToggle/StrategyResult.php
Neirda24: It could be both. To me it makes sense to keep it in StrategyResult as it is the one holding the Grant/Deny decision to convert it to a bool value. Let me know. and Strategies don't rely on this method because Feature::isEnabled already calls this method.

src/Symfony/Component/FeatureToggle/Strategy/DateStrategy.php
Neirda24: I didn't find any valid use case for Abstain in this strategy. You are either valid according to dates or not. There is "I can't answer" Maybe the Deny part could be kind of dynamic to allow either Deny or Abstain ? But I think that would make things much more complicated to understand.

src/Symfony/Component/FeatureToggle/README.md
Neirda24: agreed. Didn't think of a valid use case but this can be easily implemented using a `unanimous` strategy (to keep the same names)

src/Symfony/Component/FeatureToggle/Debug/TraceableStrategy.php
Neirda24: We followed the same rules as other components for the debug part (Serializer for example)

src/Symfony/Component/FeatureToggle/Strategy/OuterStrategiesInterface.php
stof: Drop this entirely.

We don't care that it composes other strategies.

And for tracing, it does not help as the TraceableDecorator still cannot replace the implementation by reading the inner strategies as it does not know *how* they compose.

src/Symfony/Component/FeatureToggle/Debug/TraceableStrategy.php
stof: The TraceableSerializer does not depend on FrameworkBundle. The SerializerDataCollector is part of the component.

src/Symfony/Component/FeatureToggle/Strategy/OuterStrategiesInterface.php
Neirda24: Indeed. I thought we were using it already on Traceable but seems not. If you look at the screenshots (where we show the available features and their strategies) I'd like to improve it a bit further and thought this could come in handy. Will remove for the moment then.

src/Symfony/Component/FeatureToggle/StrategyResult.php
stof: Well, that's precisely my point. The only place where it is **legitimate** to use this method is inside `Feature::isEnabled`, which is also in the component. So there is no reason to move this logic in the enum. It only encourages authors of custom strategies to use it while they should never use it.

src/Symfony/Bundle/FeatureToggleBundle/DataCollector/FeatureCheckerDataCollector.php
Neirda24: a dedicated command could be indeed the solution. To me the debug is not enough as it doesn't show each inner configuration and nested strategies.

src/Symfony/Bundle/FeatureToggleBundle/DataCollector/FeatureCheckerDataCollector.php
Neirda24: Got it. So it should be `@phpstan-return` ? Cannot seem to find the documentation for that one.

Neirda24:
> If you want a string syntax, this should rely on the ExpressionLanguage component IMO.

Maybe. But I'd like to be able to parse the string to guess which to import and (at the moment) didn' find a clean way to do it using ExpressionLanguage. I think this can be future scope.

src/Symfony/Component/FeatureToggle/Strategy/DateStrategy.php
stof: Well, with those strategies that return a result based on the first one that does not abstain, it is important to define clearly those cases to make combinators useful.

However, with the current API of the component where each Feature holds a strategy, the abstain case looks quite useless. You would wire the strategy you need, skipping the irrelevant one entirely (or having a default as part of the config of that strategy alone).

For security voters, the abstain state is very important because the authorization checker does not have different voters for each permission being checked. It has a single list of voters (and a strategy about how their decisions should be combined).

stof:
the suggested FrameworkBundle integration looks hard to use to me. The configuration of strategies looks complex to me, requiring lots of boilerplate.

Neirda24:
> Having recently worked on a similar (but much more limited) subject.. here are some questions :)
> 
> Disclaimner: all are geniune questions and i have no strong position on any :)
> 
> * should Feature be an object ?

An object is always more capable of evolving instead of a raw string IMO. It also allows to encapsulate some logic.

> * is_feature_enabled(..) indicate a unique name reference. how would you handle priorities / conflicts ?

Last in wins if I remember correctly.

> * if it's an object... should "isEnabled" be in it ? debug-aside, what is the usage of a non-enabled feature ?

There are plenty of use cases where Features might not be enabled. Let's say you run an e-commerce platform. You could set things ahead by enabling the BlackFriday feature at the right time. Otherwise it would not be enabled.

> * is the description only a debug utility ? if so, maybe it's not usefull here. if yes how would you handle translations ?

Indeed it is mainly debug / auto-documentation. No translations handled. It does not provide built-in features (yet ?). So no idea ATM. This might be removed indeed.

> * should Feature "know"/"contains" its strategy ?

I tried a totally different approach first but then it seemed to make sense yes. The feature exists only because there is a strategy to support it. And each feature might use a different strategy or the same but with different configuration.

Neirda24:
> the suggested FrameworkBundle integration looks hard to use to me. The configuration of strategies looks complex to me, requiring lots of boilerplate.

We tried other syntax but none was as straigthforward as this one. Do you have a suggestion in mind ?

src/Symfony/Component/FeatureToggle/Strategy/RequestHeaderStrategy.php
stof: Having some strategies that make sense only for standalone usage but not for usage in the full stack framework (while they have similar needs) looks wrong to me. This would actually make it easier to write buggy code.
Even though we make components usable in a standalone way, we should still ensure that we provide the best DX possible when this gets used in the context of the framework.

Note that my previous message talked about runtimes as I was thinking of the runtime component to support more exotic deployments than php-fpm, but there is actually another case that is much more common that does not allow reading headers from super-globals due to using a special "runtime" (quotes are because this is not actually a runtime in the meaning of the runtime component): running functional tests with the WebTestCase.

src/Symfony/Component/FeatureToggle/StrategyResult.php
Neirda24: Yup missed that one. Will move 👍 .

src/Symfony/Component/FeatureToggle/Strategy/DateStrategy.php
Neirda24: We allow custom strategies and nested strategies. Abstain allows to fallback to other strategies easily. We could get rid of it with syntax like ((a OR b) AND c) but the problem is then how do we parse this to automatically create the strategies (with the right configuration)

src/Symfony/Bundle/FeatureToggleBundle/DataCollector/FeatureCheckerDataCollector.php
stof: When using a phpstan type alias, the phpdoc tags should indeed use the `phpstan-` prefix. This way, a tool than does not know about phpstan would not read them at all (a tool reading `phpstan-return` without supporting the `phpstan-type` would be a buggy tool, and it would be their fault).

However, in this case, this is not the solution because of the second paragraph in my previous message.

src/Symfony/Component/FeatureToggle/Strategy/DateStrategy.php
stof: Custom strategies and nested strategies don't need a feature to abstain. The possibility to abstain only makes those cases harder.
And as each feature currently has its own strategy object, I see to reason to register strategies that will abstain. Registering the exact strategy it needs is much easier to understand.

A case like `(a OR b) AND c` would be instantiated as `new AndStrategy([new OrStrategy([new A(), new ()]), new C()])`. And `AndStrategy` and `OrStrategy` are much simpler to implement if they don't need to care about potential abstain results.

With your current combined strategies, it is not even possible to represent that case (AffirmativeStrategy is a OrStrategy as any grant provides a grant, but there is no AndStrategy). And PriorityStrategy is meaningful only because of abstain, while I see no case where I would actually want this behavior for an understandable behavior (especially when many strategies never abstain).

> We allow custom strategies and nested strategies. Abstain allows to fallback to other strategies easily. We could get rid of it with syntax like ((a OR b) AND c) but the problem is then how do we parse this to automatically create the strategies (with the right configuration)

to me, we should not have both an expression syntax with functions corresponding to some strategies and operators while having a single strategies where combining other strategies together can be done in custom ways. those are simply 2 different models which will be hard to reconcile (your PriorityStrategy has no sane representation as an expression for instance)

src/Symfony/Component/FeatureToggle/StrategyResult.php
```diff
@@ -0,0 +1,28 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\FeatureToggle;
+
+enum StrategyResult
+{
+    case Grant;
+    case Deny;
+    case Abstain;
```
andersonamuller: I think this naming makes more sense in a feature toggle context. 

```suggestion
    case Enable;
    case Disable;
    case Fallback;
```

src/Symfony/Component/FeatureToggle/Strategy/DateStrategy.php
```diff
@@ -0,0 +1,57 @@
...
+use Psr\Clock\ClockInterface;
+
+final class DateStrategy implements StrategyInterface
+{
+    public function __construct(
+        private readonly ClockInterface $clock,
+        private readonly \DateTimeImmutable|null $from = null,
+        private readonly \DateTimeImmutable|null $until = null,
+        private readonly bool $includeFrom = true,
+        private readonly bool $includeUntil = true,
```
andersonamuller:
```suggestion
        private readonly \DateTimeImmutable|null $since = null,
        private readonly \DateTimeImmutable|null $until = null,
        private readonly bool $includeSince = true,
        private readonly bool $includeUntil = true,
```

src/Symfony/Component/FeatureToggle/Feature.php
```diff
@@ -0,0 +1,40 @@
...
+
+    public function getDescription(): string
+    {
+        return $this->description;
+    }
+
+    public function isEnabled(): bool
+    {
+        return $this->strategy->compute()->isEnabled($this->default);
+    }
```
andersonamuller: The Feature class looks more like a value object, so how about passing the strategy to the method as a collaborator and remove from constructor.

The `FeatureChecker` implementation can contain a mapping of feature name -> strategy service ID following [this comment](https://github.com/symfony/symfony/pull/51649/files#r1324812652) 

```suggestion
    public function isEnabled(StrategyInterface $strategy): bool
    {
        return $strategy->compute()->isEnabled($this->default);
    }
```

src/Symfony/Component/FeatureToggle/StrategyResult.php
Neirda24: To me Disable does not make much sense in a Startegy context. It does not Disable a feature. Simply stating "I would not enabling it"

src/Symfony/Component/FeatureToggle/Feature.php
Jean-Beru: If `FeatureChecker` contains the mapping, this code could be in 
```php
    public function isEnabled(string $featureName): bool
    {
        // Retrieve Feature and Strategy from mapping

        match ($strategy->compute()) {
            StrategyResult::Grant => true,
            StrategyResult::Deny => false,
            StrategyResult::Abstain => $feature->getDefault(),
        }
    }
```

The Feature object won't have any logic anymore.

src/Symfony/Component/FeatureToggle/StrategyResult.php
Jean-Beru: Also, "Abstain" means "I can't give a result, ask someone else". Renaming it to "Fallback" may be interpreted as "I can't give a result, I give you my default value".

src/Symfony/Component/FeatureToggle/Debug/TraceableStrategy.php
Jean-Beru: Good catch, I'll move it.

src/Symfony/Component/FeatureToggle/StrategyResult.php
andersonamuller: Yeah, could be. Still, to me it reads that fallback is "I can't know for sure, so fallback to the feature default state".

In a strategy depending on a request param or env var for instance. If it's not defined fallback to default, if is defined use the value to enable/disable.

In a date period strategy, it doesn't return a fallback, the date is either inside the period or not

wadjeroudi:
when a feature is unleashed to all users, what would be the process ?

Seb33300:
Will it be easy to support external providers like LaunchDarkly?
or this component is intended to work as standalone only?

Jean-Beru:
> when a feature is unleashed to all users, what would be the process ?

We could associate this feature with a GrantStrategy which always returns "Granted". It could be interesting if you want to revert it later instead of deleting this feature and its checks from your codebase.

Jean-Beru:
> LaunchDarkly

Sure! It was one of our favorite future improvements.
That's why Feature providers were used: to allow external features.

Jean-Beru:
> Will it be easy to support external providers like LaunchDarkly?
> or this component is intended to work as standalone only?

Sure! It was one of our favorite future improvements.
That's why Feature providers were used: to allow external features.

src/Symfony/Component/FeatureToggle/FeatureChecker.php
```diff
@@ -0,0 +1,30 @@
...
+
+final class FeatureChecker implements FeatureCheckerInterface
+{
+    public function __construct(
+        private readonly FeatureCollection $features,
+        private readonly bool $whenNotFound,
+    ) {
+    }
+
+    public function isEnabled(string $featureName): bool
```
94noni: can't we imagine adding an optional context here?
this logic of computation reminds me in some way the `workflow apply` logic

we can also define some internal context keys that override the `bool $whenNotFound` for example

src/Symfony/Component/FeatureToggle/Strategy/RequestHeaderStrategy.php
Neirda24: Ok. I didn't incldued yet because I wasn't sure on which Request I should rely on ? Current Request or Master request ? Both makes sense. I also thought of using both (if it differs).

src/Symfony/Component/FeatureToggle/Strategy/DateStrategy.php
Neirda24: It is not simply cases or OR and AND. More like `a == grant || b == grant` , `a != abstain || b != abstain`, ... Abstain makes teh difference between "I couldn't answer ask someone else or use default" and "I was able to answer and the answer is to not activate it".

`PriorityStrategy` is represented like this : `a != abstain || b != abstain`. That is waht I meant by "parsing" might be hard is we need to check for the operator + the value it is compared to.

`Affirmative` is `a == grant || b == grant`

`Consensus` would be `a != abstain && b != abstain` and so on.

src/Symfony/Component/FeatureToggle/StrategyResult.php
Neirda24: For the date strategy there is indeed no "abstain" possible. Either in or out. Regarding others, it means "I cannot answer for sure neither grant nor deny" Instead of `Abstain` (which made sense to me because of Security voters) it could be renamed something else. `Neither` for example. But `Fallback` gives too much meaning as to what happens as a consequence. To me, it shouldn't have to care about what will happen only what is the Strategy answer.

And because it cannot answer, I thought "Abstain" makes sense.

src/Symfony/Component/FeatureToggle/Feature.php
Neirda24: ATM the Feature object is meant to hold the strategy so you cannot instanciate it without it. If we move to some kind of mapping, then yes the logic would probably be in the FeatureChecker and would require to have some kind of "default strategy".

src/Symfony/Component/FeatureToggle/FeatureChecker.php
Neirda24: We gave it a thought based on Serializer. ATM there is no need for a context. There will be when we introduce the "allow only for 10% of users" because we will need to calculate some IDs per request. Open to discuss )(either now if it must be part of the initial release or later if not)

src/Symfony/Component/FeatureToggle/FeatureChecker.php
```diff
@@ -0,0 +1,30 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\FeatureToggle;
+
+final class FeatureChecker implements FeatureCheckerInterface
+{
+    public function __construct(
+        private readonly FeatureCollection $features,
+        private readonly bool $whenNotFound,
```
fancyweb: Shouldn't the feature you ask for always exist? :thinking: What's the point of calling it with a feature that does not exist?

src/Symfony/Component/FeatureToggle/FeatureCollection.php
```diff
@@ -0,0 +1,98 @@
...
+
+        $this->features = [];
+
+        foreach ($this->featureProviders as $featureProvider) {
+            if (is_callable($featureProvider)) {
+                $featureProvider = $featureProvider();
+            }
+
+            foreach ($featureProvider as $feature) {
+                $this->features[$feature->getName()] = $feature;
```
fancyweb: Shouldn't we throw if 2 features have the same name?

src/Symfony/Component/FeatureToggle/FeatureCollection.php
```diff
@@ -0,0 +1,98 @@
...
+        return array_key_exists($id, $this->features);
+    }
+
+    /**
+     * @throws FeatureNotFoundException If the feature is not registered in this provider.
+     */
+    public function get(string $id): Feature
+    {
+        $this->compile();
+        return $this->features[$id] ?? throw new FeatureNotFoundException($id);
```
fancyweb: Why do we sometimes call it $featureName and sometimes $id?

src/Symfony/Component/FeatureToggle/FeatureChecker.php
Neirda24: You might have missed some checks and you don't want to display an error. You want a global fallback value. Usually it will be `false` but it might be `true` for some reason.

src/Symfony/Component/FeatureToggle/FeatureCollection.php
Neirda24: Maybe. I am not sure. Either we do some kind of last in "wins" or we ensure each is unique. But because we don't pull from teh providers during cache creation, we will figure this out only during runtime. I think it would be too much to throw an error because of this.

src/Symfony/Component/FeatureToggle/FeatureCollection.php
Neirda24: my bad. Will make it consistent with `$featureName`. good catch.

src/Symfony/Component/FeatureToggle/FeatureCollection.php
OskarStark: To provide a good DX I would throw as proposed by @fancyweb

javiereguiluz:
@Neirda24 this proposal is impressive 🙇 and the feature looks useful, so I hope this is merged 🙌 

I have a minor comment about something important: the new component name.

[Martin Fowler's article](https://martinfowler.com/articles/feature-toggles.html) calls this "Feature Toggles (aka Feature Flags)" ... but I've always seen this called "Feature Flags".

I searched in Google and:

* `"feature toggles"` = 88,900 results
* `"feature flags"` = 1,750,000 results

So, I think that, if merged, we should rename this component as `FeatureFlags`. Thanks.

src/Symfony/Bundle/FeatureToggleBundle/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2004-present Fabien Potencier
```
OskarStark:
```suggestion
Copyright (c) 2023-present Fabien Potencier
```

src/Symfony/Component/FeatureToggle/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2004-present Fabien Potencier
```
OskarStark:
```suggestion
Copyright (c) 2023-present Fabien Potencier
```

src/Symfony/Component/FeatureToggle/FeatureCollection.php
Neirda24: > makes it impossible to build any optimization where we would build the list in a compiler pass during the cache warmup.

Actually  this is what is currently dumped in the cache :
```php
$instance->withFeatures((new \Symfony\Component\FeatureToggle\Provider\InMemoryProvider([
    new \Symfony\Component\FeatureToggle\Feature('my-feature', 'Desc feature 1', true, 
        new \Symfony\Component\FeatureToggle\Strategy\PriorityStrategy([
            new \Symfony\Component\FeatureToggle\Strategy\RequestQueryStrategy('my_query'), 
            new \Symfony\Bundle\FeatureToggleBundle\Strategy\RequestStackAttributeStrategy('some_attribute', ($container->services['request_stack'] ??= new \Symfony\Component\HttpFoundation\RequestStack())),
        ])
    )
]))->provide(...));
```

I agree all features (in case of `InMemoryProvider`) are always instanciated but I don't see what else could be optimized during cache warmup.

Also take into account that feature names can come from SAAS as well. So unless we accept doing http calls during cache warmup it won't be possible to build the full list of available features. Also doing this during cache warmup would prevent people from updating the list on those SAAS providers without having to refresh symfony's cache.

> IMO, a better API would be to inject a PSR ContainerInterface in the constructor. This way, the wiring in the framework could inject a ServiceLocator, which would lazy-load the Feature object that we need and no other.

Like I said, this implies to have access to all the features which is currently not possible unless doing http calls or else. Lazy loading the Feature Object why not but it is a very light object. What would need lazy is the strategy classes IMO

src/Symfony/Component/FeatureToggle/README.md
Neirda24: done

src/Symfony/Component/FeatureToggle/Strategy/RequestHeaderStrategy.php
Neirda24: Ok I added new classes to base the strategy on RequestStack. ATM I didn't deleted the ones based on Super globals because I find it creates a heavy dependency when used outside the framework. Would that be okay to keep them or should I delete them anyway ?

src/Symfony/Component/FeatureToggle/Strategy/OuterStrategiesInterface.php
Neirda24: removed

src/Symfony/Component/FeatureToggle/Strategy/AffirmativeStrategy.php
Neirda24: removed

src/Symfony/Component/FeatureToggle/FeatureCollection.php
Neirda24: done

src/Symfony/Component/FeatureToggle/README.md
Neirda24: Added UnanimousStrategy

src/Symfony/Bundle/FeatureToggleBundle/DataCollector/FeatureCheckerDataCollector.php
Neirda24: fixed

src/Symfony/Component/FeatureToggle/Strategy/DateStrategy.php
Neirda24: fixed

src/Symfony/Component/FeatureToggle/FeatureCollection.php
Neirda24: Sorry I checked why in fact it is because here we implement the Psr/ContainerInterface. And to not break API with it (named arguments and such) We kept it `$id`

Neirda24:
> @Neirda24 this proposal is impressible 🙇 and the feature looks useful, so I hope this is merged 🙌
> 
> I have a minor comment about something important: the new component name.
> 
> [Martin Fowler's article](https://martinfowler.com/articles/feature-toggles.html) calls this "Feature Toggles (aka Feature Flags)" ... but I've always seen this called "Feature Flags".
> 
> I searched in Google and:
> 
> * `"feature toggles"` = 88,900 results
> * `"feature flags"` = 1,750,000 results
> 
> So, I think that, if merged, we should rename this component as `FeatureFlags`. Thanks.

Thanks for the encouragements ! That is indeed a good point. Renaming things might be a bit of a tedious work so if everyone settle on "FeatureFlag" or "FeatureFlags" I'll rename it after the review on the other parts ;)

src/Symfony/Component/FeatureToggle/FeatureCollection.php
Neirda24: Bear in mind that feature can be remotely provided. Thus making it impossible to know in advance what they will be without iterating over all providers. Because of this we tend to have to do this during runtime.

I think raising an exception during runtime would be bad DX. But we could make it a parameter `$throwOnError`. In the class it would default to false and in framework to `kernel.debug`.

WDYT ?

src/Symfony/Component/FeatureToggle/FeatureCollection.php
stof: If the feature comes from a SaaS tool, loading the full list might still not be the right API (and which strategy would they use then ?). I'm not convinced the existing API is good for those SaaS providers either.

src/Symfony/Component/FeatureToggle/FeatureCollection.php
stof: The PR description does not show any explanation about how this remote loading is expected to work. As the public API of the component takes a feature name as input, modeling a concept of feature loader taking the feature name as argument might be a much better fit than a concept of provider returning an iterable forcing to load everything all the time.

src/Symfony/Component/FeatureToggle/FeatureChecker.php
diimpp: Might be worth doing a command similar to `translation:extract` to find all feature flag usages in the code, which are not covered by config. Not sure how it will work with external provider though.

src/Symfony/Component/FeatureToggle/FeatureCollection.php
Neirda24: We got inspiration with how they do it in their own package and the HTTP API they exposes. They don't provide several (one to list, one to determine and so on), they pull everything locally. They continue to have some remote call to store per request which user was given which answer. What they do is on first call they put everything in cache. ATM this component does not provide full integration with such external tools. Yes there will be improvement to bring these capabilities "10% of users" for example. ATM this can already be plugged in with those tools by wrapping their actual plugins.

Regarding which strategies they will choose this has to be discussed because this can go several ways. We implement those logic in our own micro components (like notifier) or we simply bridge with their existing packages.

The strategies they offer always relies on a user list or percentage or globally available but is not as flexible as what this component tries to bring to the table (switch based on query, header, request attribute, etc etc).

For now what I can provide as a better implementation is to not compile everything right from the start but iterate over providers and stop as soon as one does provide the feature.

src/Symfony/Component/FeatureToggle/FeatureChecker.php
Neirda24: that's the thing. External provider can be configured on the fly to provide a flag or remove it. Having a not found feature flag is not an invalid state in these cases. But a debug command to list where each feature is being checked yes that would be awesome. Before doing that let's see if the component will be accepted or not as it is or if it need a major rework.

src/Symfony/Component/FeatureToggle/FeatureCollection.php
Neirda24: This would mean only one provider at a time ? Agree I can make things to not load everything all the time. Like stated in a previous comment :
- only iterate until I find a provider that provide the feature
- then instanciate the strategy classes only when calling the isEnabled (this can already be achieved by making the first strategy as lazy)

src/Symfony/Component/FeatureToggle/FeatureCollection.php
Neirda24: I've done an update on the FeatureCollection class. What could be done to improve further would be to add a `has` and a `get` method on providers. Regarding the InMemory it is an easy one and could avoid to instanciate feature if not needed. For external ones it would up to each of them to optimize things. For gitlab for example : 
- fetch from https://docs.gitlab.com/ee/api/feature_flags.html#list-feature-flags-for-a-project 
- store it in cache or keep it in memory (don't know yet which will be better suited)
- and simply extract feature names from it to implement the has / get methods.

WDYT ?

ismail1432:
Thanks for the hard work and for this great idea! 🙌🏿

Having this component makes sense nowadays 👍, I really like the idea of having some default strategy (Header, Date...). 👌 
Just a remark, I don't think that the concept of `StrategyResult` is relevant, I'm curious to see if you have use cases? 
In the end, it makes the usage less straightforward and more complex than what is done in [Laravel](https://laravel.com/docs/10.x/pennant) or in [Ruby on Rails](https://github.com/flippercloud/flipper)

Neirda24:
> Thanks for the hard work and for this great idea! 🙌🏿
> 
> Having this component makes sense nowadays 👍, I really like the idea of having some default strategy (Header, Date...). 👌 Just a remark, I don't think that the concept of `StrategyResult` is relevant, I'm curious to see if you have use cases? In the end, it makes the usage less straightforward and more complex than what is done in [Laravel](https://laravel.com/docs/10.x/pennant) or in [Ruby on Rails](https://github.com/flippercloud/flipper)

Hi. Thanks for your feedback. Really appreciated. The idea is to have a differenciation between "I provided that header with a true value", "I provided that header with a false value" & "I did not provided any header". You respectively get "force true", "force false", "fallback to default"

Neirda24:
@stof : regarding providers and lazyness we are working on a rework with @Jean-Beru . The PR should be updated soon.

src/Symfony/Bundle/FeatureToggleBundle/DataCollector/FeatureCheckerDataCollector.php
Jean-Beru: :+1: 

Features are not listed anymore in profiler. This will be done in a dedicated command.

src/Symfony/Component/FeatureToggle/FeatureCollection.php
Neirda24: Providers were totally reworked. Let me know what you think.

src/Symfony/Component/FeatureToggle/StrategyResult.php
Neirda24: moved

src/Symfony/Component/FeatureToggle/Debug/TraceableStrategy.php
Neirda24: moved

src/Symfony/Component/FeatureToggle/FeatureCollection.php
```diff
@@ -12,43 +12,33 @@
...
-        $this->append($features);
+        $this->providers = $providers;
     }
 
     /**
-     * @param iterable<Feature>|(\Closure(): iterable<Feature>) $features
+     * @param list<Feature> $features
      */
-    private function append(iterable|\Closure $features): void
+    public static function withFeatures(array $features): self
```
andersonamuller:
```suggestion
    public static function withFeatures(Feature ...$features): self
```

wouterj: This makes it much harder to add more parameters in the future in a backwards compatible way. Let's stick with array.

stof: it does not only makes it harder. It makes it impossible (unless we want to put them at the beginning of the signature, with a complex BC layer, but if would be weird if `$features` was not the first argument).

Neirda24:
@javiereguiluz : done it is now renamed to `FeatureFlags`

src/Symfony/Component/FeatureToggle/FeatureCollection.php
andersonamuller: It does not make much sense to add other parameters to a function named `withFeatures`, for BC then it would be better to add another method and deprecate this one instead.

Neirda24: I don't see real benefits here. It makes the call to that method less intuitive, for something static analysis can do. Even if people don't use it, it will break somewhere down the road. No need to enforce anything here just for the sake of it IMO

Neirda24:
A debug command was just added.

fancyweb:
Could features be "simple" classes/services where you just implement your own logic and return a boolean?
```php
#[AsFeature]
final readonly class ChristmasBannerFeature
{
    public function __invoke(): bool
    {
        $now = now();
        $year = now()->format('Y');

        return $now >= \DateTimeImmutable::createFromFormat('Y/m/d', "$year/12/20")
            && $now <= \DateTimeImmutable::createFromFormat('Y/m/d', "$year/12/25");
    }
}

```

```php
#[AsFeature('black-friday-banner')]
final readonly class BlackFridayBannerFeature
{
    public function __construct(private RequestStack $requestStack)
    {
    }

    public function __invoke(): bool
    {
        if ($this->requestStack->getCurrentRequest()?->headers->has('X-Preview-Black-Friday')) {
            return true;
        }

        $now = now();

        return $now >= new \DateTimeImmutable('14 Nov. 2023')
            && $now < new \DateTimeImmutable('24 Nov. 2023');
    }
}
```

I feel like all the "strategy" stuff and configuration might be overkill for the majority of real use cases.

src/Symfony/Component/FeatureFlags/Strategy/RequestQueryStrategy.php
```diff
@@ -0,0 +1,32 @@
...
+final class RequestQueryStrategy implements StrategyInterface
+{
+    public function __construct(
+        private readonly string $queryParameterName,
+    ) {
+    }
+
+    public function compute(): StrategyResult
+    {
+        if (!array_key_exists($this->queryParameterName, $_GET)) {
```
94noni, line 26: I've not read all the PR comments, yet
but is there any reason this $_GET is used and not the http component as wrapper?

Neirda24: This one is part of the component. But in the bundle it is wrapped with the http foundation component. I am not sure to keep the one with $_GET but I find it easier to use outside the framework as a whole

src/Symfony/Component/FeatureFlags/README.md
```diff
@@ -0,0 +1,77 @@
...
+);
+
+if ($featureChecker->isEnabled('new_feature')) {
+    // Use the new feature
+} else {
+    // Use the legacy code
+}
+```
+
+Available strategies
```
94noni, line 48: as per https://github.com/symfony/symfony/pull/51649#issuecomment-1748543083
an abstract strategy or a way to plug a "simple class" could be valuable :)

Neirda24: Implement the StrategyInterface and with autowire you're good to go.

94noni: yes I got it already, it is to avoid having to deal with internal component "strategy result" but more a simple boolean (flag on|off) from the ppl dev code

like for example the security voter does
the code+attribute on the comment mentioned would be very useful if plugged into the strategy logic :)

Neirda24: Ok got it. I misunderstood you're point. It could indeed be something added. I'll have to think about it.

OskarStark:
From my POV I am very 👍🏻 for having such a component in the core.
However this PR is to big for a first round IMHO.

I am for splitting this PR in a first version with a small basic set of features and then iterate PR over PR. Let's see what the other say.

@symfony/mergers 

Thank you so far for your effort ☀️

OskarStark:
I like your proposal @fancyweb 😃

ajgarlag:
@Neirda24 Congrats on your PR.

What do you think about providing a [twig test](https://twig.symfony.com/doc/3.x/advanced.html#tests)? With a test, the twig syntax will become:

```twig
{% if 'my-feature' is enabled %}
    {# do stuff #}
{% endif %}
{% if 'other-feature' is not enabled %}
    {# do stuff #}
{% endif %}
```

Neirda24:
Hi @fancyweb. Thank you for this idea.
It is indeed nice to read and easy to setup.
It also avoids having to declare anything in yaml.

We suggest the following instead :

```php
#[AsStrategyForFeature(name: 'black-friday-banner')]
final readonly class BlackFridayBannerStrategy
{
    public function __construct(private RequestStack $requestStack)
    {
    }

    public function __invoke(): bool
    {
        if ($this->requestStack->getCurrentRequest()?->headers->has('X-Preview-Black-Friday')) {
            return true;
        }

        $now = now();

        return $now >= new \DateTimeImmutable('14 Nov. 2023')
            && $now < new \DateTimeImmutable('24 Nov. 2023');
    }
}
```

We changed `AsFeature` to `AsStrategyForFeature` because this code is in fact a strategy in the way we designed it.
This is not implemented yet in our proposal but could be easily added later on.

It would look like this :

```php
new Feature('black-friday-banner', strategy: new CallbackStrategy((new BlackFridayBannerStrategy())(...)));
```

This allows to both have a simple solution and a very extendable one.

This solution would not be enough for more complex scenarios where we need to calculate based on number of requests or users (percentage based strategy), and would require each of your strategies to be injected with a lot of different services to handle such cases.

We think that would be the best compromise between simplicity and extendibility.

Neirda24:
> @Neirda24 Congrats on your PR.
> 
> What do you think about providing a [twig test](https://twig.symfony.com/doc/3.x/advanced.html#tests)? With a test, the twig syntax will become:
> 
> ```
> {% if 'my-feature' is enabled %}
>     {# do stuff #}
> {% endif %}
> {% if 'other-feature' is not enabled %}
>     {# do stuff #}
> {% endif %}
> ```

@ajgarlag  Nice idea 😉 . Added to description for later integration

composer.json
```diff
@@ -73,6 +73,7 @@
         "symfony/error-handler": "self.version",
         "symfony/event-dispatcher": "self.version",
         "symfony/expression-language": "self.version",
+        "symfony/feature-toggle": "self.version",
```
OskarStark:
```suggestion
        "symfony/feature-flags": "self.version",
```

src/Symfony/Bundle/FeatureFlagsBundle/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2004-present Fabien Potencier
```
OskarStark:
```suggestion
Copyright (c) 2023-present Fabien Potencier
```

src/Symfony/Bundle/FeatureFlagsBundle/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+**This Bundle is experimental**.
+[Experimental features](https://symfony.com/doc/current/contributing/code/experimental.html)
+are not covered by Symfony's
+[Backward Compatibility Promise](https://symfony.com/doc/current/contributing/code/bc.html).
+
+Resources
+---------
+
+* [Contributing](https://symfony.com/doc/current/contributing/index.html)
```
OskarStark:
```suggestion
 * [Contributing](https://symfony.com/doc/current/contributing/index.html)
```

src/Symfony/Bundle/FeatureFlagsBundle/README.md
```diff
@@ -0,0 +1,18 @@
...
+* [Report issues](https://github.com/symfony/symfony/issues) and
+  [send Pull Requests](https://github.com/symfony/symfony/pulls)
+  in the [main Symfony repository](https://github.com/symfony/symfony)
```
OskarStark:
```suggestion
 * [Report issues](https://github.com/symfony/symfony/issues) and
   [send Pull Requests](https://github.com/symfony/symfony/pulls)
   in the [main Symfony repository](https://github.com/symfony/symfony)
```

94noni:
hello, kind of nitpick but reading the PR code I found it weird the plural, why `flags` and not `flag`?
I mean, it is not `workflows` neither `messengers` even if we know there are chances there will be multiple

nicolas-grekas:
Let me sum up what I understand from the domain of feature flags.

A feature:
- has a name
- resolves to a boolean most of the time but could resolve to an int/string/enum sometimes (eg bg-color)
- decides based on a context, which can be a user, a group of users, a request attribute, etc.
- is persistent, so that the same input context leads to the same result until this state is reset.

Then, we want to be able to resolve features by name as if there were callables with no arguments.

In twig, this could mean doing `{% if feature('feat-name') %}` when the result is a bool, or `{% if feature('feat-name', 'the-needed-result') %}` when something else.

`#[Route(condition: "feature('feat-name', 'the-needed-result')"]` makes sense. We could also imagine an `#[Feature('feat-name', 'the-needed-result')]` attribute for controllers if that'd make sense.

Inside PHP code, we'd like to be able to do the same: `$featureChecker->has('feat-name')`.

I think that's all from the end-user POV, is that correct?

To make this work, I would suggest a design like the following:

To start very simply, a feature would be an invokable service with a name, eg.:
```php
#[AsFeature('the-name')] # the name could be optional and default to the FQCN
class MyFeature
{
    function __invoke(MyUser $user) { return $user->canUseMyFeature(); }
}
```

Then, we would build a `FeatureChecker` that gets the collection of all feature services and would be in charge of:
- resolving their arguments (akin controller arguments - we could even provide an adapter for existing ones)
- generating a context string from the resolved arguments
- use a cache-pool to cache already checked features
- call the feature implementation when not cached
- return the resolved value of the feature

In order to give control over the cached entries, features could accept an `ItemInterface` object so that features could define their lifetime or even attach some cache-tags to them.

Then, we would also provide some generic services that ppl could use in their feature implementation e.g. to leverage SaaS providers (since features are services, they could use whatever services they'd need to compute the result.)

We could also allow declaring features using configuration, if we can figure our some generic implementations that need just some parameters to be useful. But I'm not sure about this part. Maybe it's simpler to provide only the "write-a-class" way of doing things so that it's easier to teach/explain. Such generic features could be made trivial to implement by providing all the building blocks in PHP.

This design would not allow managing features with arbitrary names coming from a SaaS provider. But I'm not convinced we need to cover this: if any part of the code knows about a feature named `"foo"`, then we can have a description of it in the app. This means I'm not convinced we need the concept of feature providers that exists in the current state of the PR.

Jean-Beru:
> Let me sum up what I understand from the domain of feature flags.
> 
> A feature:
> 
> * has a name
> * resolves to a boolean most of the time but could resolve to an int/string/enum sometimes (eg bg-color)
> * decides based on a context, which can be a user, a group of users, a request attribute, etc.
> * is persistent, so that the same input context leads to the same result until this state is reset.
> 
> Then, we want to be able to resolve features by name as if there were callables with no arguments.
> 
> In twig, this could mean doing `{% if feature('feat-name') %}` when the result is a bool, or `{% if feature('feat-name', 'the-needed-result') %}` when something else.
> 
> `#[Route(condition: "feature('feat-name', 'the-needed-result')"]` makes sense. We could also imagine an `#[Feature('feat-name', 'the-needed-result')]` attribute for controllers if that'd make sense.
> 
> Inside PHP code, we'd like to be able to do the same: `$featureChecker->has('feat-name')`.
> 
> I think that's all from the end-user POV, is that correct?

I totally agree with this introduction.

> To start very simply, a feature would be an invokable service with a name, eg.:
> 
> ```
> #[AsFeature('the-name')] # the name could be optional and default to the FQCN
> class MyFeature
> {
>     function __invoke(MyUser $user) { return $user->canUseMyFeature(); }
> }
> ```

Sure! We already worked on a such attribute. BTW, it can also be used on methods:
```php
#[AsFeature('the-name', method: 'compute')]
class MyFeature
{
    function compute() { return /* ... */; }
}
// or
class MyFeature
{
    #[AsFeature('the-name')]
    function compute() { return /* ... */; }
}
```

> Then, we would build a `FeatureChecker` that gets the collection of all feature services and would be in charge of:
> 
> * resolving their arguments (akin controller arguments - we could even provide an adapter for existing ones)
> * generating a context string from the resolved arguments
> * use a cache-pool to cache already checked features
> * call the feature implementation when not cached
> * return the resolved value of the feature
> 
> In order to give control over the cached entries, features could accept an `ItemInterface` object so that features could define their lifetime or even attach some cache-tags to them.

I think that resolving argument will make this first round harder to read while we could inject services using the DI. Also, current `ArgumentResolver` depends on a `Request` which is not always true for feature flags.

Also, if the feature's result is cacheable (probably always true, at least to get the same value during the request), what about using a dedicated service to generate the cache key ? The logic behind the cache key may be complex and different for each feature. 

Ex: generating a cache key grouped by the User's company `'my_feature_'.$user->getCompany()->getId()`.
 
> Then, we would also provide some generic services that ppl could use in their feature implementation e.g. to leverage SaaS providers (since features are services, they could use whatever services they'd need to compute the result.)

A solution could be to merge `FeatureProvider` and `FeatureCollection` in a single service which can return a feature from a predefined list, a local database or a SaaS provider from its name. This service could implement a `FeatureCollectionInterface::get(string $featureName): callable` to allow people to implement their own source of features. WDYT ?
 
> We could also allow declaring features using configuration, if we can figure our some generic implementations that need just some parameters to be useful. But I'm not sure about this part. Maybe it's simpler to provide only the "write-a-class" way of doing things so that it's easier to teach/explain. Such generic features could be made trivial to implement by providing all the building blocks in PHP.

"write-a-class" way will be OK to declare predefined features. Current yaml configuration is convenient to apply the same logic but with different configuration to features. A solution could be to allow declaring custom `FeatureCollection`.

Ex:
```php
class CustomFeatureCollection implement FeatureCollectionInterface
{
    public function get(string $featureName): callable
    {
        // check if $featureName is whitelisted

        // return a callable using RequestStack, a SaaS provider, etc.
    }
}
```
 
> This design would not allow managing features with arbitrary names coming from a SaaS provider. But I'm not convinced we need to cover this: if any part of the code knows about a feature named `"foo"`, then we can have a description of it in the app. This means I'm not convinced we need the concept of feature providers that exists in the current state of the PR.

Sometimes, the feature name cannot be known by the application (ex: set by an admin in a Form and stored in DB).

But we could implement the previously mentioned `FeatureCollectionInterface::get(string $featureName): callable` method to return a `callable` which queries the SaaS provider instead of returning a default value or throwing an exception.

src/Symfony/Component/FeatureFlags/FeatureChecker.php
```diff
@@ -0,0 +1,30 @@
...
+
+final class FeatureChecker implements FeatureCheckerInterface
+{
+    public function __construct(
+        private readonly FeatureCollection $features,
+        private readonly bool $whenNotFound,
+    ) {
+    }
+
+    public function isEnabled(string $featureName): bool
```
94noni, line 22: passing by, wanted to ask if it may be relevant to have an optional $context = [] passed here?
so implementation of each feature flag can leverage this and do more computation to decide true|false

94noni:
hello, what is the state of the PR? can it be reviewed as is or need another round of code after recent comments?
thx

Jean-Beru:
Hello, nice to see that some people are still interested in this feature 😊
Reviews show that this component is too complex for a first strike.
We recently worked on a simpler PR and I hope I can submit it next week 😉

Jean-Beru:
@94noni done in https://github.com/symfony/symfony/pull/53213 and https://github.com/Jean-Beru/symfony/pull/2 :santa:

soyuka:
I really like the way this is going, I'd like to recommend openfeature as a first citizen integration framework for feature flags or at least as an inspiration. Thanks for the work!

Jean-Beru:
> I really like the way this is going, I'd like to recommend openfeature as a first citizen integration framework for feature flags or at least as an inspiration. Thanks for the work!

Of course! OpenFeature, just like Unleashed, is a strong candidate for priority integration as an adapter.  Working on both could also help improve the interface.

Neirda24:
a reminder that unleash is what is used on Gitlab and Jira as well.

SofHad:
good!

I propose managing feature toggles via an environment variable, which would be far more practical for sysadmins and operations teams.

```
Example:
# Enable all features except the specified ones
FEATURES_CONFIG=!some-beta-feature,christmas-banner

# Enable only the listed feature(s)
FEATURES_CONFIG=some-beta-feature,...
```

Jean-Beru:
> good!
> 
> I propose managing feature toggles via an environment variable, which would be far more practical for sysadmins and operations teams.
> 
> ```
> Example:
> # Enable all features except the specified ones
> FEATURES_CONFIG=!some-beta-feature,christmas-banner
> 
> # Enable only the listed feature(s)
> FEATURES_CONFIG=some-beta-feature,...
> ```

A specific `EnvVarProvider` could be implemented later in a dedicated PR (to keep this one as simple as possible) to address this need :slightly_smiling_face:  

Edit: I though you was talking about https://github.com/symfony/symfony/pull/53213 but this is also true for this one.

nicolas-grekas:
Closing in favour of #53213, which the authors opened as the simpler first iteration of this proposal and which is still active. This PR also still targets the unmaintained 7.0 branch. Thanks a lot for the original design work, the discussion here fed directly into #53213.

PR #51650, opened 2023-09-13
[AssetMapper] Add audit command

Jean-Beru:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | TODO

After discussing with @WebMamba, I made this PR to introduce the `importmap:audit` command inspired from `yarn audit` and `npm audit`.

It reads the `importmap.php` and extract packages version to query the `https://registry.npmjs.org/-/npm/v1/security/audits` API.

image

image

### DIFF
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
@@ -1,0 +1,119 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\AssetMapper\ImportMap;
+
+use Symfony\Component\AssetMapper\Exception\RuntimeException;
+use Symfony\Component\AssetMapper\ImportMap\Resolver\PackageResolverInterface;
+use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+class ImportMapAuditor
+{
+    private const AUDIT_URL = 'https://api.github.com/advisories';
+
+    private readonly HttpClientInterface $httpClient;
+
+    public function __construct(
+        private readonly ImportMapConfigReader $configReader,
+        private readonly PackageResolverInterface $packageResolver,
+        HttpClientInterface $httpClient = null,
+    ) {
+        $this->httpClient = $httpClient ?? HttpClient::create();
+    }
+
+    /**
+     * @return list<ImportMapPackageAudit>
+     */
+    public function audit(): array
+    {
+        $entries = $this->configReader->getEntries();
+
+        if ([] === $entries) {
+            return [];
+        }
+
+        /** @var array<string, array<string, ImportMapPackageAudit>> $installed */
+        $packageAudits = [];
+
+        /** @var array<string, list<string>> $installed */
+        $installed = [];
+        $affectsQuery = [];
+        foreach ($entries as $entry) {
+            if (null === $entry->url) {
+                continue;
+            }
+            $version = $entry->version ?? $this->packageResolver->getPackageVersion($entry->url);
+
+            $installed[$entry->importName] ??= [];
+            $installed[$entry->importName][] = $version;
+
+            $packageVersion = $entry->importName.($version ? '@'.$version : '');
+            $packageAudits[$packageVersion] ??= new ImportMapPackageAudit($entry->importName, $version);
+            $affectsQuery[] = $packageVersion;
+        }
+
+        // @see https://docs.github.com/en/rest/security-advisories/global-advisories?apiVersion=2022-11-28#list-global-security-advisories
+        $response = $this->httpClient->request('GET', self::AUDIT_URL, [
+            'query' => ['affects' => implode(',', $affectsQuery)],
+        ]);
+
+        if (200 !== $response->getStatusCode()) {
+            throw new RuntimeException(sprintf('Error %d auditing packages. Response: %s', $response->getStatusCode(), $response->getContent(false)));
+        }
+
+        foreach ($response->toArray() as $advisory) {
+            foreach ($advisory['vulnerabilities'] ?? [] as $vulnerability) {
+                if (
+                    null === $vulnerability['package']
+                    || 'npm' !== $vulnerability['package']['ecosystem']
+                    || !array_key_exists($package = $vulnerability['package']['name'], $installed)
+                ) {
+                    continue;
+                }
+                foreach ($installed[$package] as $version) {
+                    if (!$version || !$this->versionMatches($version, $vulnerability['vulnerable_version_range'] ?? '>= *')) {
+                        continue;
+                    }
+                    $packageAudits[$package.($version ? '@'.$version : '')] = $packageAudits[$package.($version ? '@'.$version : '')]->withVulnerability(
+                        new ImportMapPackageAuditVulnerability(
+                            $advisory['ghsa_id'],
+                            $advisory['cve_id'],
+                            $advisory['url'],
+                            $advisory['summary'],
+                            $advisory['severity'],
+                            $vulnerability['vulnerable_version_range'],
+                            $vulnerability['first_patched_version'],
+                        )
+                    );
+                }
+            }
+        }
+
+        return array_values($packageAudits);
+    }
+
+    private function versionMatches(string $version, string $ranges): bool
+    {
+        foreach (explode(',', $ranges) as $rangeString) {
+            $range = explode(' ', trim($rangeString));
+            if (1 === count($range)) {
+                $range = ['=', $range[0]];
+            }
+
+            if (!version_compare($version, $range[1], $range[0])) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+}
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapConfigReader.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapConfigReader.php
@@ -38,7 +38,7 @@ class ImportMapConfigReader
 
         $entries = new ImportMapEntries();
         foreach ($importMapConfig ?? [] as $importName => $data) {
-            $validKeys = ['path', 'url', 'downloaded_to', 'type', 'entrypoint'];
+            $validKeys = ['path', 'url', 'downloaded_to', 'type', 'entrypoint', 'version'];
             if ($invalidKeys = array_diff(array_keys($data), $validKeys)) {
                 throw new \InvalidArgumentException(sprintf('The following keys are not valid for the importmap entry "%s": "%s". Valid keys are: "%s".', $importName, implode('", "', $invalidKeys), implode('", "', $validKeys)));
             }
@@ -57,6 +57,7 @@ class ImportMapConfigReader
                 isDownloaded: isset($data['downloaded_to']),
                 type: $type,
                 isEntrypoint: $isEntry,
+                version: $data['version'] ?? null,
             ));
         }
 
@@ -82,6 +83,9 @@ class ImportMapConfigReader
             }
             if ($entry->isEntrypoint) {
                 $config['entrypoint'] = true;
+            }
+            if ($entry->version) {
+                $config['version'] = $entry->version;
             }
             $importMapConfig[$entry->importName] = $config;
         }
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapPackageAudit.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapPackageAudit.php
@@ -1,0 +1,32 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\AssetMapper\ImportMap;
+
+final class ImportMapPackageAudit
+{
+    public function __construct(
+        public readonly string $package,
+        public readonly ?string $version,
+        /** @var array<ImportMapPackageAuditVulnerability> */
+        public readonly array $vulnerabilities = [],
+    ) {
+    }
+
+    public function withVulnerability(ImportMapPackageAuditVulnerability $vulnerability): self
+    {
+        return new self(
+            $this->package,
+            $this->version,
+            [...$this->vulnerabilities, $vulnerability],
+        );
+    }
+}
--- a/src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
@@ -158,6 +158,15 @@ final class JsDelivrEsmResolver implements PackageResolverInterface
         return array_values($resolvedPackages);
     }
 
+    public function getPackageVersion(string $url): ?string
+    {
+        if (1 === preg_match("#^https://cdn.jsdelivr.net/npm/(?<package>(?:@[a-z0-9-~][a-z0-9-._~]*/)?[a-z0-9-~][a-z0-9-._~]*)@(?<version>[\w\._-]+)(?<subpath>/.*)?$#", $url, $matches)) {
+            return $matches['version'];
+        }
+
+        return null;
+    }
+
     /**
      * Parses the very specific import syntax used by jsDelivr.
      *
--- a/src/Symfony/Component/AssetMapper/ImportMap/Resolver/JspmResolver.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/Resolver/JspmResolver.php
@@ -96,4 +96,13 @@ final class JspmResolver implements PackageResolverInterface
             throw $e;
         }
     }
+
+    public function getPackageVersion(string $url): ?string
+    {
+        if (1 === preg_match("#^https://ga.jspm.io/npm:(?<package>(?:@[a-z0-9-~][a-z0-9-._~]*/)?[a-z0-9-~][a-z0-9-._~]*)@(?<version>[\w\._-]+)(?<subpath>/.*)?$#", $url, $matches)) {
+            return $matches['version'];
+        }
+
+        return null;
+    }
 }
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/asset_mapper.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/asset_mapper.php
@@ -17,6 +17,7 @@ use Symfony\Component\AssetMapper\AssetMapperDevServerSubscriber;
 use Symfony\Component\AssetMapper\AssetMapperInterface;
 use Symfony\Component\AssetMapper\AssetMapperRepository;
 use Symfony\Component\AssetMapper\Command\AssetMapperCompileCommand;
+use Symfony\Component\AssetMapper\Command\ImportMapAuditCommand;
 use Symfony\Component\AssetMapper\Command\DebugAssetMapperCommand;
 use Symfony\Component\AssetMapper\Command\ImportMapInstallCommand;
 use Symfony\Component\AssetMapper\Command\ImportMapRemoveCommand;
@@ -27,6 +28,7 @@ use Symfony\Component\AssetMapper\Compiler\JavaScriptImportPathCompiler;
 use Symfony\Component\AssetMapper\Compiler\SourceMappingUrlsCompiler;
 use Symfony\Component\AssetMapper\Factory\CachedMappedAssetFactory;
 use Symfony\Component\AssetMapper\Factory\MappedAssetFactory;
+use Symfony\Component\AssetMapper\ImportMap\ImportMapAuditor;
 use Symfony\Component\AssetMapper\ImportMap\ImportMapConfigReader;
 use Symfony\Component\AssetMapper\ImportMap\ImportMapManager;
 use Symfony\Component\AssetMapper\ImportMap\ImportMapRenderer;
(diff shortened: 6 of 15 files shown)


src/Symfony/Component/AssetMapper/Command/AuditCommand.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+
+        $audit = $this->importMapAuditor->audit();
+
+        if ([] === $audit) {
+            $io->info('No packages to audit.');
```
stof: instead of always outputting text, I think we should support selecting the output format. The audit command looks like a good candidate for usage in scripted contexts.
For reference, `composer audit` supports a JSON output.

src/Symfony/Component/AssetMapper/Command/AuditCommand.php
```diff
@@ -0,0 +1,78 @@
...
+            $table->addRow([
+                sprintf('<fg=%s>%s</>', self::SEVERITY_COLORS[$advisory['severity']] ?? 'default', ucfirst($advisory['severity'])),
+                $advisory['title'],
+                $advisory['module_name'],
+                $audit['packages'][$advisory['module_name']] ?? 'n/a',
+                $advisory['patched_versions'],
+                $advisory['url'],
+            ]);
+        }
+        $table->render();
```
stof: what if there is no advisories ? Rendering table headers without any rows looks weird to me.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
```diff
@@ -0,0 +1,79 @@
...
+    ) {
+        $this->httpClient = $httpClient ?? HttpClient::create();
+    }
+
+    public function audit(): array
+    {
+        $packages = $this->getPackagesVersion();
+
+        if ([] === $packages) {
+            return [];
```
stof, line 41: this looks wrong to me. We should report an array with the proper array shape.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
```diff
@@ -0,0 +1,79 @@
...
+    private readonly HttpClientInterface $httpClient;
+
+    public function __construct(
+        private readonly string $importMapConfigPath,
+        HttpClientInterface $httpClient = null,
+    ) {
+        $this->httpClient = $httpClient ?? HttpClient::create();
+    }
+
+    public function audit(): array
```
stof, line 36: it would be great to add phpdoc describing the structure of the returned array. This would make it easier to use this class by knowing what its signature actually is.
And it will also make it easier to maintain that class by knowing better which kind of changes would be BC breaks.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
```diff
@@ -0,0 +1,79 @@
...
+        $data = $response->toArray();
+
+        return [
+            'packages' => $packages,
+            'advisories' => $data['advisories'],
+            'summary' => $data['metadata']['vulnerabilities'],
+        ];
+    }
+
+    private function getPackagesVersion(): array
```
stof: This should have a phpdoc describing the return type as `array<string, string>`

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
```diff
@@ -0,0 +1,79 @@
...
+            }
+
+            // try to grab the version from the URL
+            $registryRegex = '(?<registry>\w+)';
+            $packageRegex = '(?<package>(?:@[a-z0-9-~][a-z0-9-._~]*/)?[a-z0-9-~][a-z0-9-._~]*)';
+            $versionRegex = '(?<version>[\w\._-]+)';
+            $subPathRegex = '(?<subpath>/.*)';
+
+            if (
+                1 === preg_match("#^https://ga.jspm.io/{$registryRegex}:{$packageRegex}@{$versionRegex}{$subPathRegex}?$#", $url, $matches)
```
stof: why using a registry regex when the only valid value for something supported by the npmjs endpoint would be `npm` ?

For instance, if you use the `gh` ecosystem for cdn.jsdelivr.net URLs, you won't get an npm package name after it, so this won't be valid.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
```diff
@@ -0,0 +1,79 @@
...
+
+namespace Symfony\Component\AssetMapper\ImportMap;
+
+use Symfony\Component\AssetMapper\Exception\RuntimeException;
+use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+class ImportMapAuditor
+{
+    private const AUDIT_URL = 'https://registry.npmjs.org/-/npm/v1/security/audits';
```
WebMamba: I saw that npm uses the gh API to process their audit. Why not directly work with the gh api?

WebMamba (commented):
Wooo thanks @Jean-Beru 🧡

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
```diff
@@ -0,0 +1,79 @@
...
+            $versionRegex = '(?<version>[\w\._-]+)';
+            $subPathRegex = '(?<subpath>/.*)';
+
+            if (
+                1 === preg_match("#^https://ga.jspm.io/{$registryRegex}:{$packageRegex}@{$versionRegex}{$subPathRegex}?$#", $url, $matches)
+                || 1 === preg_match("#^https://cdn.jsdelivr.net/{$registryRegex}/{$packageRegex}@{$versionRegex}{$subPathRegex}?$#", $url, $matches)
+                || 1 === preg_match("#^https://unpkg.com/{$packageRegex}@{$versionRegex}{$subPathRegex}?$#", $url, $matches)
+            ) {
+                $packages[$matches['package']] = $matches['version'];
+            }
```
weaverryan, line 114: If we can't successfully extract the package or version, we need to report that to the user so they know that there is at least on package that we are not sure about. So, somehow we need to also return problematic importmap entries

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
```diff
@@ -0,0 +1,79 @@
...
+            'packages' => $packages,
+            'advisories' => $data['advisories'],
+            'summary' => $data['metadata']['vulnerabilities'],
+        ];
+    }
+
+    private function getPackagesVersion(): array
+    {
+        $path = $this->importMapConfigPath;
+        $importMapConfig = is_file($path) ? (static fn () => include $path)() : [];
```
weaverryan: I'm leaning towards injecting `ImportMapManager` into this class and making `loadImportMapEntries()` public inside that class. Then we could just call that from here and work with `ImportMapEntry` objects.

weaverryan (commented):
Fantastic! Thanks for creating this!

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
weaverryan: Let's extract this section to the individual [PackageResolverInterface](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/AssetMapper/ImportMap/Resolver/PackageResolverInterface.php) classes - e.g. a new `getPackageInfo(string $url): ?array` which would return... an array with `package` and `version` keys or `null` of a resolver doesn't support the URL (or maybe we make a new small class instead of the array).

We'll need to reuse this logic for an `asset-map:outdated` command. So putting it in the package resolvers would allow that.

stof: Given that the interface is an extension point of the package, I suggest using a value object rather than an array shape.

Jean-Beru: Maybe we can add this check at "resolvePackages" step, add a `?string $version` property to `ResolvedImportMapPackage` and dump it in `importmap.php` file. It will look like:
```php
<?php

return [
    '@hotwired/stimulus' => [
        'url' => 'https://unpkg.com/@hotwired/stimulus@3.2.1/dist/stimulus.js',
        'version' => '3.2.1',
    ],
    'file' => [
        'path' => 'some/file.js',
    ],
];
```

WDYT ?

javiereguiluz:
This looks like a nice proposal to me. Thanks Hubert.

I have a minor comment about the command output. At the bottom we can see this:

```
5 packages audited
* jquery
* bootstrap
* json5
* semver
* minimist

9 vulnerabilities found
* 0 Info
* 0 Low
* 6 Moderate
* 2 High
* 1 Critical
```

Two comments:

* I'd remove the "N packages audited" because this list could be very long in practice and we already have that information in the `Package` column of the table above
* I'd change the "N vulnerabilities found" to display them in a single line and omit the categories that have `0` issues (and sort the rest from highest to lowest)

So, this is how the output would look like:

```
9 vulnerabilities found: 1 Critical / 2 High / 6 Moderate
```

src/Symfony/Component/AssetMapper/Command/AuditCommand.php
Jean-Beru: I added a `format` option which supports "txt" or "json" for now.

src/Symfony/Component/AssetMapper/Command/AuditCommand.php
Jean-Beru: Good catch. Done.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
Jean-Beru: It is now replaced by a `ImportMapPackageAudit` DTO.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
Jean-Beru: This logic has been moved to resolvers. `JsDelivrEsmResolver` already get the version when requiring a package. `JspmResolver` still use a regex (which might be improved I think).

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
Jean-Beru: Good catch, API calls have been updated.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
Jean-Beru: I added an output to show how many packages are ignored due to a `null` version.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
Jean-Beru: You're right. Using `ImportMapManager` will avoid breaks if `importmap.php` format changes.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
Jean-Beru: You can review this in the last commit.

Jean-Beru:
> * I'd remove the "N packages audited" because this list could be very long in practice and we already have that information in the `Package` column of the table above

We can keep it to inform that, there were n packages that could not be audited. 

> * I'd change the "N vulnerabilities found" to display them in a single line and omit the categories that have `0` issues (and sort the rest from highest to lowest)

Good idea. I made changes and updated screenshots in the PR description.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
Jean-Beru: Fixed by returning `list<ImportMapPackageAudit>`.

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JspmResolver.php
```diff
@@ -96,4 +96,13 @@ public function resolvePackages(array $packagesToRequire): array
             throw $e;
         }
     }
+
+    private function extractVersion(string $url): ?string
+    {
+        if (1 === preg_match("#^https://ga.jspm.io/npm:(?<package>(?:@[a-z0-9-~][a-z0-9-._~]*/)?[a-z0-9-~][a-z0-9-._~]*)@(?<version>[\w\._-]+)(?<subpath>/.*)?$#", $url, $matches)) {
```
Jean-Beru, line 102: Not sure that this regex is enough to handle all possible resolved URLs...

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -72,6 +72,7 @@ public function resolvePackages(array $packagesToRequire): array
 
             $version = $response->toArray()['version'];
             $requiredPackages[$i][1] = $this->httpClient->request('GET', sprintf($this->distUrlPattern, $packageName, $version, $filePath));
+            $requiredPackages[$i][4] = $version;
```
Jean-Beru: Test should be added if we agree to fetch version at "require" step.

src/Symfony/Component/AssetMapper/Command/AuditCommand.php
```diff
@@ -0,0 +1,166 @@
...
+            mode: InputOption::VALUE_REQUIRED,
+            description: sprintf('The output format ("%s")', implode(', ', $this->getAvailableFormatOptions())),
+            default: 'txt',
+        );
+    }
+
+    protected function initialize(InputInterface $input, OutputInterface $output): void
+    {
+        $this->io = new SymfonyStyle($input, $output);
+        $this->format = $input->getOption('format');
```
weaverryan: I'm not sure if setting these types of things on properties in commands is a normal practice. Someone else can please tell me if I'm wrong or right about that :)

src/Symfony/Component/AssetMapper/Command/AuditCommand.php
```diff
@@ -0,0 +1,166 @@
...
+        $table->addRows($rows);
+        $table->render();
+        $this->io->newLine();
+
+        $packagesCount = count($audit);
+        $this->io->text(sprintf('%d package%s found: %d audited / %d skipped (no version)',
+            $packagesCount,
+            1 === $packagesCount ? '' : 's',
+            $packagesCount - $packagesWithoutVersionCount,
+            $packagesWithoutVersionCount,
```
weaverryan: I think if we find any of these, the command should return with the failure status code. People will depend on this command to 100% assure them that they have no vulnerabilities

src/Symfony/Component/AssetMapper/Command/AuditCommand.php
```diff
@@ -0,0 +1,166 @@
...
+
+    private function displayTxt(array $audit): int
+    {
+        $rows = [];
+
+        $packagesWithoutVersionCount = 0;
+        $vulnerabilitiesCount = array_map(fn() => 0, self::SEVERITY_COLORS);
+        foreach ($audit as $packageAudit) {
+            if (!$packageAudit->version) {
+                ++$packagesWithoutVersionCount;
```
weaverryan: Instead of the number (or in addition to the number), we should list the packages so the use could go add a `version` key manually to `importmap.php` if they'd like to fix the problem.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -318,7 +319,7 @@ private function cleanupPackageFiles(ImportMapEntry $entry): void
     /**
      * @return array<string, ImportMapEntry>
      */
-    private function loadImportMapEntries(): array
+    public function loadImportMapEntries(): array
```
weaverryan: This is fine - in #51543 I extract this to another service anyway because I also needed to reuse it elsewhere.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapPackageAudit.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Symfony\Component\AssetMapper\ImportMap;
```
weaverryan, line 12: Check fabbot in the CI to get help adding things like the LICENSE headers

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
weaverryan: I like that you added `version` to `importMapEntry`. My thinking is that we should *not* set this when you require a package, because for all of our normal resolvers, the package+version pattern is perfectly predictable. I think we can only realistically have the problem where we can't find the version of someone added a package manually to importmap.php that points somewhere different. If the user has that situation, they can add the `version`. But otherwise, you will never see or be aware of this key.

To support this, we'll probably need to add a method to `PackageResolverInterface` - e.g. `getVersion(string $url): ?string`.

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JspmResolver.php
weaverryan: That's ok - if you have a few test cases to cover the cases we know about, it should be good enough. If we find another case later, we can fix it.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
```diff
@@ -0,0 +1,106 @@
...
+
+        /** @var array<string, list<string>> $installed */
+        $installed = [];
+        $affectsQuery = [];
+        foreach ($entries as $entry) {
+            if (null === $entry->url) {
+                continue;
+            }
+            $installed[$entry->importName] ??= [];
+            $installed[$entry->importName][] = $entry->version;
```
weaverryan: We can save for another PR, but probably we need to add an optional `packageName` to `ImportMapEntry`. 99% of the time the key (`importName` here) IS the package name. But that's not required. In fact, you can run `importmap:require "vue/dist/vue.esm-bundler.js=vue` to do this exact thing :).

The idea would be that `packageName` never appears in `importmap.php` unless you have changed the key, and then you would need it.

weaverryan (commented):
This is really, really nice. I left a few comments - the only major thing is that I'd like to avoid putting a `version` key in `importmap.php` unless it's needed. The other 99% of the time, we should grab the version from the `url`.

Thanks!

weaverryan:
We've merged in some changes to AssetMapper - let me know if you need help finishing this / fixing the conflicts :).

src/Symfony/Component/AssetMapper/ImportMap/ImportMapPackageAudit.php
Jean-Beru: Done.

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JspmResolver.php
Jean-Beru: OK. Tests are ready to accept new URLs in their data providers.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
Jean-Beru: [Update] Resolver now must implement a `getPackageVersion` method with extract version from URL.

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
Jean-Beru: I added a `getPackageVersion` to the `PackageResolverInterface` in order to retrieve the package version. It's called when a version is not explicitly defined in the `importmap.php` file.

src/Symfony/Component/AssetMapper/Command/AuditCommand.php
Jean-Beru: We can move it. IMO it is more readable like this instead of passing `$io` in every method.

src/Symfony/Component/AssetMapper/Command/AuditCommand.php
Jean-Beru: Done.

src/Symfony/Component/AssetMapper/Command/AuditCommand.php
Jean-Beru: Done. I also updated screenshots to show when a version cannot be retrieved.

Jean-Beru:
> We've merged in some changes to AssetMapper - let me know if you need help finishing this / fixing the conflicts :).

@weaverryan Thanks for the proposal. I rebased this PR, it should be OK now.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapConfigReader.php
```diff
@@ -57,6 +57,7 @@ public function getEntries(): ImportMapEntries
                 isDownloaded: isset($data['downloaded_to']),
                 type: $type,
                 isEntrypoint: $isEntry,
+                version: $data['version'] ?? null,
```
weaverryan, line 60: `version` needs to be written also if it's not null in `writeEntries()`.

weaverryan (approved):
This is good to go. It's going to conflict with my #51786, which adds a `version` key to `importmap.php` (a bit similar to what's done in this PR), but if this is merged first, I can patch things up over in my PR after.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapConfigReader.php
Jean-Beru: sure. it's done.

Jean-Beru:
> This is good to go. It's going to conflict with my https://github.com/symfony/symfony/pull/51786, which adds a version key to importmap.php (a bit similar to what's done in this PR), but if this is merged first, I can patch things up over in my PR after.

You can ping me when your PR is merged. I'll update mine.

fabpot (approved):

fabpot:
Thank you @Jean-Beru.

Issue #51384, opened 2023-08-14
[Scheduler] Caching does not work for multiple messages in same schedule


StanJansen:
### Symfony version(s) affected

6.3

### Description

When you have multiple messages in the same schedule, only the timestamp of the latest message will be saved in the cache. When your consumer stops (by `--time-limit` or unexpected) the next run date for the other messages will be calculated again. So when you have recurring messages for 1 minute and 30 minutes, with a restart policy of 60 seconds, the 30 minutes message will never be executed when you do do not pass the `from` argument. 

As a temporary workaround I disabled jitter and set the `from` argument to a date in the past on 00:00:00, which makes the message be calculated based on that (so always on 00:30, 01:00 etc.). However if the consumer is restarting or down due to a crash at that moment, they will not be catched up when booting up the consumer again as the last-ran-date is not cached. This is why I disabled jitter to reduce that risk to the specific time instead of adding the jitter time to that.

### How to reproduce

- Create a stateful schedule with a recurring message scheduled every minute (message A), and one message every 30 minutes (message B). Do not pass a from date.
- Dump the next run dates in `MessageGenerator.php`
- Start the consumer
- Assert that message A is scheduled for +1 minute, assert that message B is scheduled for +30 minutes
- Wait a minute for message A to be executed
- Restart the consumer
- Message A is scheduled for +1 minute (relative to the previous datetime), message B is scheduled for +30 minutes again (so ~1 minute later than the last next run date it dumped)

### Possible Solution

Edit the caching so it works per message, not per schedule.

### Additional Context

I  created a PR that got merged in 6.3.2 that fixed the `from` argument. This fixed the scheduler from executing 30 minute messages every minutely restart for example, but due to this bug that fix now causes those messages to never get executed. So unless you pass a `from` date, those schedules are completely broken at the moment.

valtzu:
I tried stateful schedule for the first time and it seems to have very strange behavior indeed.


```php
        (new Schedule())
            ->stateful($this->cache)
            ->add(RecurringMessage::every(1, new TestMessage('every 1 seconds')))
            ->add(RecurringMessage::every(4, new TestMessage('every 4 seconds')))
```

When continuing on a cached state:

```
[18:44:11.000] every 1 seconds
[18:44:12.000] every 1 seconds
[18:44:13.000] every 1 seconds
[18:44:14.000] every 4 seconds
[18:44:14.000] every 1 seconds
[18:44:15.000] every 1 seconds
[...]

PR #51651, opened 2023-09-13
[Scheduler] Fix stateful scheduler

### LINKED ISSUE #51646: [Scheduler] Intermittent Runs

6.3.4

### Description

When using the scheduler with a single daily recurring message we are often seeing days where the scheduler appears to have not run. 

Here is the schedule config:

```php
return (new Schedule())
    ->add(RecurringMessage::every('1 day', new OurDailyMessage(), from: '10:30'))
    ->stateful($this->cache)
;
```

We have configured caching for the scheduler provider using the database, to allow it to catchup if the task was to crash.  The servers run in UTC.

We have added some middleware to attempt to give some visibility to scheduler activity, which just logs when each message is dispatched on the schedule.

```php
class SchedulerLoggingMiddleware implements MiddlewareInterface
{
    public function __construct(
        private readonly Connection $connection,
    ) {
    }

    public function handle(Envelope $envelope, StackInterface $stack): Envelope
    {
        if (null !== $envelope->last(ScheduledStamp::class)) {
            $this->connection->insert('scheduler_log', [
                'date_created' => new \DateTimeImmutable(),
                'message' => $envelope->getMessage()::class,
            ], [
                Types::DATETIME_IMMUTABLE,
                \PDO::PARAM_STR,
            ]);
        }

        return $stack->next()->handle($envelope, $stack);
    }
}
```

This is then run in production alongside our other Messenger transports (eg. `async another_transport`) with an ECS task.

```bash
bin/console messenger:consume scheduler_default async another_transport etc --limit=1000 --time-limit=3600 
```

As you can see from the logs it appears some days are missing (ie. 10th and 7th in this example).

image

I know this is a tricky one, as I don't have an exact reproduction, or really know where to start debugging this. I'm hoping in raising an issue perhaps there's something obvious which is incorrectly configured, or someone can indicate an area to look at to attempt to debug or reproduce this.

### How to reproduce

As indicated above I do not have a reproduction, running the scheduler locally on shorter timeframes seems to behave as intended, and debugging a daily job is hard so I'm hoping for some pointers before beginning on that.

### Possible Solution

Unsure if this is related: https://github.com/symfony/symfony/issues/51384

### Additional Context

 1. Is running the scheduler in the same command as other messenger transports a valid use? I thought the cache should allow the scheduler to catch up if it happens to be busy at that moment...  but should the scheduler always be run as the sole transport?

valtzu:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #51646, #51384
| License       | MIT

Stateful scheduler seems rather broken at the moment, see https://github.com/symfony/symfony/issues/51384#issuecomment-1703915652.

Let's fix it by storing the original first run start time, that way it's always possible to recalculate the state.

Catching-up works now:
```
[23:14:11.709710] Worker started
[23:14:11.759318] every 2 seconds
[23:14:11.760291] every 2 seconds
[23:14:11.761257] every 2 seconds
[23:14:11.763244] every 2 seconds
[23:14:12.637054] every 2 seconds
[23:14:14.620595] every 2 seconds
[23:14:16.632170] every 2 seconds
```

Whereas before it would only start on from the current item, possibly skipping previous items like stated in #51646. _(is this a bc break?)_

I will be waiting for input from authors of the related issues.

---

One test is failing because because `getNextRunDate` is called with `2020-02-20T01:59:00+02:00` and then next run date is expected at `2020-02-20T02:09:00+02:00` but we get `2020-02-20T02:00:00+02:00` because that's set as `from`. I don't quite get the logic, I would assume that it is expected to be run immediately on `from` :thinking:

### DIFF
--- a/src/Symfony/Component/Scheduler/Generator/CheckpointInterface.php
+++ b/src/Symfony/Component/Scheduler/Generator/CheckpointInterface.php
@@ -15,6 +15,8 @@ interface CheckpointInterface
 {
     public function acquire(\DateTimeImmutable $now): bool;
 
+    public function from(): \DateTimeImmutable;
+
     public function time(): \DateTimeImmutable;
 
     public function index(): int;
--- a/src/Symfony/Component/Scheduler/Trigger/PeriodicalTrigger.php
+++ b/src/Symfony/Component/Scheduler/Trigger/PeriodicalTrigger.php
@@ -13,17 +13,18 @@ namespace Symfony\Component\Scheduler\Trigger;
 
 use Symfony\Component\Scheduler\Exception\InvalidArgumentException;
 
-class PeriodicalTrigger implements TriggerInterface
+class PeriodicalTrigger implements StatefulTriggerInterface
 {
     private float $intervalInSeconds = 0.0;
-    private \DateTimeImmutable $from;
+    private ?\DateTimeImmutable $from;
     private \DateTimeImmutable $until;
     private \DatePeriod $period;
     private string $description;
+    private string|int|float|\DateInterval $interval;
 
     public function __construct(
         string|int|float|\DateInterval $interval,
-        string|\DateTimeImmutable $from = new \DateTimeImmutable(),
+        string|\DateTimeImmutable|null $from = null,
         string|\DateTimeImmutable $until = new \DateTimeImmutable('3000-01-01'),
     ) {
         $this->from = \is_string($from) ? new \DateTimeImmutable($from) : $from;
@@ -70,7 +71,7 @@ class PeriodicalTrigger implements TriggerInterface
                     $this->description = sprintf('every %s seconds', $this->intervalInSeconds);
                 }
             } else {
-                $this->period = new \DatePeriod($this->from, $i, $this->until);
+                $this->interval = $i;
             }
         } catch (\Exception $e) {
             throw new InvalidArgumentException(sprintf('Invalid interval "%s": ', $interval instanceof \DateInterval ? 'instance of \DateInterval' : $interval).$e->getMessage(), 0, $e);
@@ -82,15 +83,22 @@ class PeriodicalTrigger implements TriggerInterface
         return $this->description;
     }
 
+    public function continue(\DateTimeImmutable $startedAt): void
+    {
+        $this->from ??= $startedAt;
+    }
+
     public function getNextRunDate(\DateTimeImmutable $run): ?\DateTimeImmutable
     {
+        $this->from ??= $run;
+
         if ($this->intervalInSeconds) {
             if ($this->until <= $run) {
                 return null;
             }
 
-            $fromDate = min($this->from, $run);
-            $from = $fromDate->format('U.u');
+            $fromDate = $this->from;
+            $from = (float) $fromDate->format('U.u');
             $delta = $run->format('U.u') - $from;
             $recurrencesPassed = floor($delta / $this->intervalInSeconds);
             $nextRunTimestamp = sprintf('%.6F', ($recurrencesPassed + 1) * $this->intervalInSeconds + $from);
@@ -103,6 +111,7 @@ class PeriodicalTrigger implements TriggerInterface
             return $this->until > $nextRun ? $nextRun : null;
         }
 
+        $this->period ??= new \DatePeriod($this->from, $this->interval, $this->until);
         $iterator = $this->period->getIterator();
         while ($run >= $next = $iterator->current()) {
             $iterator->next();
@@ -130,6 +139,6 @@ class PeriodicalTrigger implements TriggerInterface
 
     private function calcInterval(\DateInterval $interval): float
     {
-        return $this->from->setTimestamp(0)->add($interval)->format('U.u');
+        return (float) (new \DateTimeImmutable('@0'))->add($interval)->format('U.u');
     }
 }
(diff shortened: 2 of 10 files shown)


src/Symfony/Component/Scheduler/Trigger/PeriodicalTrigger.php
```diff
@@ -85,19 +86,24 @@ public function __toString(): string
...
 
-            $fromDate = min($this->from, $run);
-            $from = $fromDate->format('U.u');
+            $fromDate = $this->from;
+            $from = (float) $fromDate->format('U.u');
             $delta = $run->format('U.u') - $from;
             $recurrencesPassed = floor($delta / $this->intervalInSeconds);
             $nextRunTimestamp = sprintf('%.6F', ($recurrencesPassed + 1) * $this->intervalInSeconds + $from);
-            $nextRun = \DateTimeImmutable::createFromFormat('U.u', $nextRunTimestamp, $fromDate->getTimezone());
+            $nextRun = \DateTimeImmutable::createFromFormat('U.u', $nextRunTimestamp)->setTimezone($fromDate->getTimezone());
```
fabpot: It looks like this change is not needed, is it?

src/Symfony/Component/Scheduler/Generator/CheckpointInterface.php
```diff
@@ -18,6 +18,8 @@ interface CheckpointInterface
 {
     public function acquire(\DateTimeImmutable $now): bool;
 
+    public function from(): \DateTimeImmutable;
```
fabpot, line 18: Not a BC break as the component is experimental in 6.3, but this should be described in the component CHANGELOG file.

src/Symfony/Component/Scheduler/Trigger/PeriodicalTrigger.php
```diff
@@ -133,6 +140,6 @@ private function canBeConvertedToSeconds(\DateInterval $interval): bool
 
     private function calcInterval(\DateInterval $interval): float
     {
-        return $this->from->setTimestamp(0)->add($interval)->format('U.u');
+        return (new \DateTimeImmutable('@0'))->add($interval)->format('U.u');
```
fabpot: Why do we need this change?
We might add `(float) ` to make Psalm happy :)

fabpot (changes requested):
I was working on the same.
The failing test must be updated as the current one is broken and does not make sense as you already noticed.

src/Symfony/Component/Scheduler/Trigger/PeriodicalTrigger.php
valtzu: From my testing, UTC is always forced in the DateTime instance when it is created from unix timestamp (timezone constructor arg has no effect), so to have the same timezone, it must be set after the instance is created.

But you are right in that this fix is not related to the issue and can/should probably be excluded here.

src/Symfony/Component/Scheduler/Trigger/PeriodicalTrigger.php
valtzu: I originally made this because $this->from was not always available at this point, I'm not sure if that's the case anymore

fabpot: ok, can you submit a PR on this change on 6.3?

fabpot:
This must be merged on 6.4 (because of the BC break, which is allowed in 6.4 for experimental components, but not for a patch release).

valtzu:
> I was working on the same. 

@fabpot Would you like to take over this, or if you had done work on this already, discard this MR as you probably have better knowledge in this :sweat_smile:

src/Symfony/Component/Scheduler/Trigger/PeriodicalTrigger.php
fabpot: Can you submit another PR on 6.3 for this change then?

fabpot:
> > I was working on the same.
> 
> @fabpot Would you like to take over this, or if you had done work on this already, discard this MR as you probably have better knowledge in this 😅

Actually, I like your code better, so if you take into account my few comments, I will be able to merge it fast.

src/Symfony/Component/Scheduler/Generator/CheckpointInterface.php
stof: Well, technically it is a BC break, but an allowed one. the changelog needs to explicitly mark it as a BC break.

valtzu: If we didn't want this BC break, we could store 2 checkpoints, one for the start state (I guess index would always be -1) and other for latest state (=the current). Let me know if you'd prefer that

fabpot: As we will merge this PR in 6.4, adding a new method is acceptable, so no worries.

valtzu:
Sorry I think I misunderstood your comment @fabpot and retargeted this to 6.4 :sweat_smile: 

The triggers have changed quite much between the 6.3&6.4  branches, but I should still target 6.3 and leave conflict resolution to someone else?

Oh my, I think I have notified everyone... sorry guys!

I will finish for today not to cause any more trouble.

fabpot (approved):

fabpot (approved):

fabpot:
Thank you @valtzu.

PR #51653, opened 2023-09-14
[Messenger] Add WrappedExceptionsInterface for nested exceptions

Jeroeny:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| License | MIT

Metric & logging tools often want to measure / log individual exceptions from the messenger. There are currently two exception classes that hold a bunch of collected exceptions from different messages or handlers. It would be nice if there was a single interface to check and call upon when extracting these nested exceptions.

Example usecase: https://github.com/getsentry/sentry-symfony/pull/760/files#diff-da0fb4498178e4866e794b813999618022c327dea59f9277b86a7abf784aeafaR98

### DIFF
--- a/UPGRADE-6.4.md
+++ b/UPGRADE-6.4.md
@@ -113,6 +113,7 @@ Messenger
 ---------
 
  * Deprecate `StopWorkerOnSignalsListener` in favor of using the `SignalableCommandInterface`
+ * Deprecate `HandlerFailedException::getNestedExceptions()`, `HandlerFailedException::getNestedExceptionsOfClass()` and  `DelayedMessageHandlingException::getExceptions()` which are replaced by a new `getWrappedExceptions()` method
 
 MonologBridge
 -------------
--- a/src/Symfony/Component/Messenger/CHANGELOG.md
+++ b/src/Symfony/Component/Messenger/CHANGELOG.md
@@ -9,6 +9,9 @@ CHANGELOG
  * Add support for multiple Redis Sentinel hosts
  * Add `--all` option to the `messenger:failed:remove` command
  * `RejectRedeliveredMessageException` implements `UnrecoverableExceptionInterface` in order to not be retried
+ * Add `WrappedExceptionsInterface` interface for exceptions that hold multiple individual exceptions
+ * Deprecate `HandlerFailedException::getNestedExceptions()`, `HandlerFailedException::getNestedExceptionsOfClass()`
+   and `DelayedMessageHandlingException::getExceptions()` which are replaced by a new `getWrappedExceptions()` method
 
 6.3
 ---
--- a/src/Symfony/Component/Messenger/Exception/DelayedMessageHandlingException.php
+++ b/src/Symfony/Component/Messenger/Exception/DelayedMessageHandlingException.php
@@ -19,8 +19,10 @@ use Symfony\Component\Messenger\Envelope;
  *
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  */
-class DelayedMessageHandlingException extends RuntimeException
+class DelayedMessageHandlingException extends RuntimeException implements WrappedExceptionsInterface
 {
+    use WrappedExceptionsTrait;
+
     private array $exceptions;
     private Envelope $envelope;
 
@@ -41,11 +43,16 @@ class DelayedMessageHandlingException extends RuntimeException
 
         $this->exceptions = $exceptions;
 
-        parent::__construct($message, 0, $exceptions[0]);
+        parent::__construct($message, 0, $exceptions[array_key_first($exceptions)]);
     }
 
+    /**
+     * @deprecated since Symfony 6.4, use {@see self::getWrappedExceptions()} instead
+     */
     public function getExceptions(): array
     {
+        trigger_deprecation('symfony/messenger', '6.4', 'The "%s()" method is deprecated, use "%s::getWrappedExceptions" instead.', __METHOD__, self::class);
+
         return $this->exceptions;
     }
 
--- a/src/Symfony/Component/Messenger/Exception/HandlerFailedException.php
+++ b/src/Symfony/Component/Messenger/Exception/HandlerFailedException.php
@@ -13,8 +13,10 @@ namespace Symfony\Component\Messenger\Exception;
 
 use Symfony\Component\Messenger\Envelope;
 
-class HandlerFailedException extends RuntimeException
+class HandlerFailedException extends RuntimeException implements WrappedExceptionsInterface
 {
+    use WrappedExceptionsTrait;
+
     private array $exceptions;
     private Envelope $envelope;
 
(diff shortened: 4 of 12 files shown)


kbond (approved):

nicolas-grekas (commented):
In order to reduce the cost of upgrading, I'd prefer not deprecating anything and instead add `NestedExceptionsInterface::getExceptions()`

This would mean removing the `$recursive` argument, but I think this would be legit: flattening the exception hierarchy isn't needed for the interface to be sufficient.

WDYT?

src/Symfony/Component/Messenger/Exception/HandlerFailedException.php
```diff
@@ -45,14 +47,6 @@ public function getEnvelope(): Envelope
         return $this->envelope;
     }
 
-    /**
-     * @return \Throwable[]
-     */
-    public function getNestedExceptions(): array
```
nicolas-grekas, line 51: ah, I missed this one, so we have to deprecate one of them :(

src/Symfony/Component/Messenger/Exception/NestedExceptionsTrait.php
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * @author Jeroen <https://github.com/Jeroeny>
+ */
+trait NestedExceptionsTrait
+{
+    /**
+     * @return \Throwable[]
+     */
+    public function getNestedExceptions(bool $recursive = true): array
```
nicolas-grekas: ok, so if we go with a new method, what about this signature? (and deprecate getNestedExceptionOfClass)
`getWrappedExceptions(string $class = null, bool $recursive = false)`

(wrapped or embedded)

Jeroeny: That should be fine too, I've committed the changes. Lmk what you think.

kbond:
On the topic of nested exceptions. I'm looking for feedback on #51331 - getting the context of these exceptions. In the case of `HandlerFailedException`: what handler threw the exception. Is there a way currently?

UPGRADE-6.4.md
```diff
@@ -112,6 +112,7 @@ Messenger
 ---------
 
  * Deprecate `StopWorkerOnSignalsListener` in favor of using the `SignalableCommandInterface`
+ * Deprecate `HandlerFailedException::getNestedExceptionOfClass` which is replaced by `HandlerFailedException::getWrappedExceptions`
```
```suggestion
 * Deprecate `HandlerFailedException::getNestedExceptions()`, `HandlerFailedException::getNestedExceptionsOfClass()` and  `DelayedMessageHandlingException::getExceptions()` which are replaced by a new `getWrappedExceptions()` method
```

src/Symfony/Component/Messenger/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Deprecate `StopWorkerOnSignalsListener` in favor of using the `SignalableCommandInterface`
  * Add `HandlerDescriptor::getOptions`
+ * Add `NestedExceptionsInterface` interface for exceptions that hold multiple individual exceptions
```
nicolas-grekas:
```suggestion
 * Add `WrappedExceptionsInterface` interface for exceptions that hold multiple individual exceptions
```

src/Symfony/Component/Messenger/Exception/DelayedMessageHandlingException.php
```diff
@@ -17,8 +17,10 @@
  *
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  */
-class DelayedMessageHandlingException extends RuntimeException
+class DelayedMessageHandlingException extends RuntimeException implements NestedExceptionsInterface
 {
+    use NestedExceptionsTrait;
```
nicolas-grekas:
```suggestion
    use WrappedExceptionsTrait;
```

src/Symfony/Component/Messenger/Exception/HandlerFailedException.php
```diff
@@ -55,6 +57,8 @@ public function getNestedExceptions(): array
 
     public function getNestedExceptionOfClass(string $exceptionClassName): array
     {
+        trigger_deprecation('symfony/messenger', '6.4', 'The "%s()" method is deprecated, use "%s::getWrappedExceptions" instead.', __METHOD__, self::class);
```
nicolas-grekas, line 67: the runtime deprecation should be doubled by an annotation for all deprecated methods

src/Symfony/Component/Messenger/Exception/NestedExceptionsTrait.php
```diff
@@ -0,0 +1,44 @@
...
+    private function getWrappedExceptionsRecursively(?string $class, iterable $exceptions): \Traversable
+    {
+        foreach ($exceptions as $exception) {
+            if ($exception instanceof NestedExceptionsInterface) {
+                yield from $this->getWrappedExceptionsRecursively($class, $exception->getWrappedExceptions());
+
+                return;
+            }
+
+            yield $exception;
```
nicolas-grekas: better not use a generator to create an array, let's create the array directly

src/Symfony/Component/Messenger/Tests/Exception/HandlerFailedExceptionTest.php
```diff
@@ -66,4 +67,15 @@ public function testThatNestedExceptionClassAreNotFoundIfNotPresent()
         $handlerException = new HandlerFailedException($envelope, [$exception]);
         $this->assertCount(0, $handlerException->getNestedExceptionOfClass(MyOwnException::class));
     }
+
+    public function testThatNestedExceptionsRecursive()
```
nicolas-grekas:
```suggestion
    public function testThatWrappedExceptionsRecursive()
```

nicolas-grekas (commented):
Let's replace the "nested" vocabulary by "wrapped" everywhere.
Can you please also account for #51331 and preserve non-numeric keys (using array_merge(_recursive) should do it).
Last but not least, please also add test covering the $class argument.

src/Symfony/Component/Messenger/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Deprecate `StopWorkerOnSignalsListener` in favor of using the `SignalableCommandInterface`
  * Add `HandlerDescriptor::getOptions`
+ * Add `WrappedExceptionsInterface` interface for exceptions that hold multiple individual exceptions
```
nicolas-grekas, line 15:
```suggestion
 * Deprecate `HandlerFailedException::getNestedExceptions()`, `HandlerFailedException::getNestedExceptionsOfClass()` and  `DelayedMessageHandlingException::getExceptions()` which are replaced by a new `getWrappedExceptions()` method
 
```

src/Symfony/Component/Messenger/Exception/WrappedExceptionsInterface.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+namespace Symfony\Component\Messenger\Exception;
+
+/**
+ * Exception that holds multiple exceptions thrown by one or more handlers and/or messages.
+ *
+ * @author Jeroen <https://github.com/Jeroeny>
+ */
+interface WrappedExceptionsInterface extends \Throwable
```
nicolas-grekas:
```suggestion
interface WrappedExceptionsInterface
```

src/Symfony/Component/Messenger/Exception/WrappedExceptionsTrait.php
```diff
@@ -0,0 +1,52 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Messenger\Exception;
+
+/**
+ * @author Jeroen <https://github.com/Jeroeny>
```
nicolas-grekas, line 15: let's mark the trait as `@internal`

nicolas-grekas (approved):
Can you please rebase? This LGTM after some minor tweaks on my side.

Jeroeny:
@nicolas-grekas Done. I also changed usages of `HandlerFailedException::getNestedExceptions()`  to `getWrappedExceptions()`.

fabpot (approved):

fabpot:
Thank you @Jeroeny.

PR #51665, opened 2023-09-15
[Security] OidcTokenHandler support JWKSet 

### LINKED ISSUE #53491: `JWKSet` support for `OIDCTokenHandler`
### Description

We would like to use the symfony/security-http OIDCTokenHandler but the Identity and Access Management system currently employed exposes a JWK Keyset via it's realm certs url as opposed to a single JWK of which the token hander in question only currently supports.

It would be great if the OIDCTokenHandler could also accept an instance of JWKSet to be passed into it's constuctor which will also incur modifications in how the jms is verified.

### Example
```diff
final class OidcTokenHandler implements AccessTokenHandlerInterface
{
    use OidcTrait;

    public function __construct(
        private Algorithm $signatureAlgorithm,
-       private JWK$jwk,
+       private JWK|JWKSet $jwk,
        private string $audience,
        private array $issuers,
        private string $claim = 'sub',
        private ?LoggerInterface $logger = null,
        private ClockInterface $clock = new Clock()
    ) {
    }
}
```

louismariegaborit:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #53491
| License       | MIT
| Doc PR        | ToDo

This PR can supports now : 
- ~~RSA256 algorithm if `web-token/jwt-signature-algorithm-rsa` package is installed.~~ (dedicated PR https://github.com/symfony/symfony/pull/53682)
- JWKSet from url

The need comes from the validation of an AWS Cognito Token.
Amazon gived a url to get JWKSet and signature is RS256.

P.S.: It's my first feature PR on Symfony. The description may be missing information or the changes may be clumsy. :blush:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -157,8 +157,7 @@
         "twig/inky-extra": "^2.12|^3",
         "twig/markdown-extra": "^2.12|^3",
         "web-token/jwt-checker": "^3.1",
-        "web-token/jwt-signature-algorithm-ecdsa": "^3.1",
-        "web-token/jwt-signature-algorithm-rsa": "^3.1"
+        "web-token/jwt-signature-algorithm-ecdsa": "^3.1"
     },
     "conflict": {
         "ext-psr": "<1.1|>=2",
--- a/composer.json
+++ b/composer.json
@@ -157,7 +157,8 @@
         "twig/inky-extra": "^2.12|^3",
         "twig/markdown-extra": "^2.12|^3",
         "web-token/jwt-checker": "^3.1",
-        "web-token/jwt-signature-algorithm-ecdsa": "^3.1"
+        "web-token/jwt-signature-algorithm-ecdsa": "^3.1",
+        "web-token/jwt-signature-algorithm-rsa": "^3.1"
     },
     "conflict": {
         "ext-psr": "<1.1|>=2",
--- a/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
+++ b/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
@@ -39,7 +39,7 @@ class OidcTokenHandlerFactory implements TokenHandlerFactoryInterface
 
         // @see Symfony\Bundle\SecurityBundle\DependencyInjection\Security\Factory\SignatureAlgorithmFactory
         // for supported algorithms
-        if (\in_array($config['algorithm'], ['ES256', 'ES384', 'ES512'], true)) {
+        if (\in_array($config['algorithm'], ['ES256', 'ES384', 'ES512', 'RS256'], true)) {
             $tokenHandlerDefinition->replaceArgument(0, new Reference('security.access_token_handler.oidc.signature.'.$config['algorithm']));
         } else {
             $tokenHandlerDefinition->replaceArgument(0, (new ChildDefinition('security.access_token_handler.oidc.signature'))
--- a/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
+++ b/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
@@ -39,12 +39,18 @@ class OidcTokenHandlerFactory implements TokenHandlerFactoryInterface
         // @see Symfony\Bundle\SecurityBundle\DependencyInjection\Security\Factory\SignatureAlgorithmFactory
         // for supported algorithms
         if (\in_array($config['algorithm'], ['ES256', 'ES384', 'ES512'], true)) {
-            $tokenHandlerDefinition->replaceArgument(0, new Reference('security.access_token_handler.oidc.signature.'.$config['algorithm']));
+            $algorithmDefinition = new Reference('security.access_token_handler.oidc.signature.'.$config['algorithm']);
         } else {
-            $tokenHandlerDefinition->replaceArgument(0, (new ChildDefinition('security.access_token_handler.oidc.signature'))
+            $algorithmDefinition = (new ChildDefinition('security.access_token_handler.oidc.signature'))
                 ->replaceArgument(0, $config['algorithm'])
-            );
+            ;
         }
+
+        $algorithmManagerDefinition = $container->setDefinition($id.'.algorithm_manager', (new ChildDefinition('security.access_token_handler.oidc.algorithm_manager'))
+            ->replaceArgument(0, [$algorithmDefinition])
+        );
+
+        $tokenHandlerDefinition->replaceArgument(0, $algorithmManagerDefinition);
 
         if (!isset($config['jwks_url']) && !isset($config['key'])) {
             throw new LogicException('You should defined key or jwks_url parameter in configuration.');
--- a/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
+++ b/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
@@ -38,7 +38,7 @@ class OidcTokenHandlerFactory implements TokenHandlerFactoryInterface
 
         // @see Symfony\Bundle\SecurityBundle\DependencyInjection\Security\Factory\SignatureAlgorithmFactory
         // for supported algorithms
-        if (\in_array($config['algorithm'], ['ES256', 'ES384', 'ES512', 'RS256'], true)) {
+        if (\in_array($config['algorithm'], ['ES256', 'ES384', 'ES512'], true)) {
             $tokenHandlerDefinition->replaceArgument(0, new Reference('security.access_token_handler.oidc.signature.'.$config['algorithm']));
         } else {
             $tokenHandlerDefinition->replaceArgument(0, (new ChildDefinition('security.access_token_handler.oidc.signature'))
--- a/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
+++ b/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
@@ -12,6 +12,8 @@
 namespace Symfony\Bundle\SecurityBundle\DependencyInjection\Security\AccessToken;
 
 use Jose\Component\Core\Algorithm;
+use Jose\Component\Core\JWK;
+use Jose\Component\Core\JWKSet;
 use Symfony\Component\Config\Definition\Builder\NodeBuilder;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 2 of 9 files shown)


src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
```diff
@@ -46,9 +47,20 @@ public function create(ContainerBuilder $container, string $id, array|string $co
...
 
-        $tokenHandlerDefinition->replaceArgument(1, (new ChildDefinition('security.access_token_handler.oidc.jwk'))
-            ->replaceArgument(0, $config['key'])
-        );
+        if (!isset($config['jwks_url']) && !isset($config['key'])) {
+            throw new LogicException('You should defined key or jwks_url parameter in configuration.');
+        }
+
+        if (isset($config['jwks_url'])) {
+            $jwksJson = file_get_contents($config['jwks_url']);
```
stof: this should use a proper http client, not `file_get_contents`, to support proper features (like a timeout for instance).

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/SignatureAlgorithmFactory.php
```diff
@@ -29,6 +29,7 @@ public static function create(string $algorithm): AlgorithmInterface
             case 'ES256':
             case 'ES384':
             case 'ES512':
+            case 'RS256':
```
stof: This is wrong. It should go in a different section because the package to install to support it is not `web-token/jwt-signature-algorithm-ecdsa` but `web-token/jwt-signature-algorithm-rsa`

stof: And the exception message at the end of the method should also be updated.

src/Symfony/Component/Security/Http/AccessToken/Oidc/OidcTokenHandler.php
```diff
@@ -39,7 +39,7 @@ final class OidcTokenHandler implements AccessTokenHandlerInterface
 
     public function __construct(
         private Algorithm $signatureAlgorithm,
-        private JWK $jwk,
+        private JWKSet $jwkSet,
```
stof: This is a BC break. The constructor must keep accepting a JWK (turning the argument type into `JWKSet|JWK`).

the property itself can be turned into a JWKSet as it is private.

src/Symfony/Component/Security/Http/Tests/AccessToken/Oidc/OidcTokenHandlerTest.php
```diff
@@ -54,7 +55,7 @@ public function testGetsUserIdentifierFromSignedToken(string $claim, string $exp
 
         $userBadge = (new OidcTokenHandler(
             new ES256(),
-            $this->getJWK(),
+            $this->getJWKSet(),
```
stof: having to updating the arguments passed to a class constructor in its test is a good indicator that your change does not preserve BC.

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/SignatureAlgorithmFactory.php
stof: The new algorithm also needs to be supported in `src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php`

stof (changes requested):
This misses updating the changelog to mention the addition of support for the RS256 algorithm and the `jwks_url` option

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/SignatureAlgorithmFactory.php
louismariegaborit: I did the change.

src/Symfony/Component/Security/Http/AccessToken/Oidc/OidcTokenHandler.php
louismariegaborit: I re added JWK as an authorized type.

src/Symfony/Component/Security/Http/Tests/AccessToken/Oidc/OidcTokenHandlerTest.php
louismariegaborit: I rollacked the change.
I think that I add specific test for JWKSet no ?

louismariegaborit:
> This misses updating the changelog to mention the addition of support for the RS256 algorithm and the `jwks_url` option

Changelog added.
Thanks for your review. I do the last change in few days.

src/Symfony/Component/Security/Http/Tests/AccessToken/Oidc/OidcTokenHandlerTest.php
stof: yes, the case of jwkset should indeed be tested

louismariegaborit: test added

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
louismariegaborit: file_get_contents replaced by http client.
I hope this was what you expected.

src/Symfony/Bundle/SecurityBundle/composer.json
```diff
@@ -40,7 +41,6 @@
         "symfony/expression-language": "^5.4|^6.0|^7.0",
         "symfony/form": "^5.4|^6.0|^7.0",
         "symfony/framework-bundle": "^6.4|^7.0",
-        "symfony/http-client": "^5.4|^6.0|^7.0",
```
stloyd: It still can be an option IMO, you would just need to check for it's availability and throw an exception if not found.

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
louismariegaborit: Would you prefer inject HttpClient in the TokenHandler class like OidcUserInfoTokenHandler ?  
`$jwk` would therefore have either the JWK or HttpClientInterface type. And if it's HttpClientInterface we construct the JWKSet with response of http request.

src/Symfony/Bundle/SecurityBundle/composer.json
louismariegaborit: I replace the require with an exception if HttpClient class was not found.

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
louismariegaborit: @stof Do you have any time for this comment ?

composer.json
```diff
@@ -157,7 +157,8 @@
         "twig/inky-extra": "^2.12|^3",
         "twig/markdown-extra": "^2.12|^3",
         "web-token/jwt-checker": "^3.1",
-        "web-token/jwt-signature-algorithm-ecdsa": "^3.1"
+        "web-token/jwt-signature-algorithm-ecdsa": "^3.1",
+        "web-token/jwt-signature-algorithm-rsa": "^3.1"
```
nicolas-grekas: Is there a reason why adding support for RSA is in this PR? If it's independent, this should be submitted as a separate PR to me.

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
```diff
@@ -37,17 +38,35 @@ public function create(ContainerBuilder $container, string $id, array|string $co
...
-        );
+        if (!isset($config['jwks_url']) && !isset($config['key'])) {
+            throw new LogicException('You should defined key or jwks_url parameter in configuration.');
+        }
+
+        if (isset($config['jwks_url'])) {
+            if (!class_exists(HttpClient::class)) {
+                throw new LogicException(sprintf('You cannot use "%s" as the HttpClient component is not installed. Try running "composer require symfony/http-client".', __CLASS__));
+            }
+            $httpClient = HttpClient::create();
```
nicolas-grekas: This cannot be the way because creating the client inline breaks IoC and also means we're going to hit the oidc server for every single requests.
I suggest removing the "url" part from this PR so that we can consider how this request should be done separately, and keep this PR on adding support for JWKSet only.

nicolas-grekas:
The http-client could be injected as a constructor argument to the factory - that'd solve the IoC issue.
But that wouldn't solve hitting the oidc server for every single requests. We would need a cache for that.
Then, I'm wondering which sort of cache we'd need. Can the JWKSet change from time to time while the app is running or is it fine to wait for a new redeploy to get the fresh keys?
Our `OidcTokenHandler` doesn't support updating JWKs directly. We could improve that by making it accept a `\Closure $jwsVerifier($jws)` that'd be in charge of fetching JWKs from the cache/http-client.
Or we could extend JKWSet to make a variant of it that does it instead. Or maybe @Spomky has a better idea? Could this extended JWKSet be added to the `Jose` namespace for example?

composer.json
louismariegaborit: Yes it makes sense.  
PR for RSA support is opened.
https://github.com/symfony/symfony/pull/53682

Spomky: Hi,

Please not that RSA1_5 algorithms are now deprecated. See [RFC 8725](https://datatracker.ietf.org/doc/html/rfc8725#section-3.2) for more information.

The package `web-token/jwt-signature-algorithm-rsa` contains both RSA1_5 (`RSxxx`) and RSA-OAEP (`PSxxx`) algorithms.
To me, even if the RS256 is very common, it should be avoiding when possible. I don't know what algorithms are used by Amazon Cognito, but I am pretty sure they provide alternatives to this algorithm.

nicolas-grekas: They tell only about RSA:
https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-verifying-a-jwt.html

Spomky:
> Or maybe @Spomky has a better idea? 

Indeed, the JWKSet may change from time to time. It completely depends on the distant service policy and keys may rotate on a periodic manner or be revoked.
Having a HTTP Client with caching features is a solution I used in the past. Another solution I prefer if to share the keysets to the application using a read-only file. The keys are changed by cron task that periodically. In other words: the application is not responsible of managing the keys. This way, there is no client injection or latency waiting for the keysets to be updated.
It requires a bit more work, but does not impact the application performance and reduces its complexity.

> Or we could extend JKWSet to make a variant of it that does it instead. 

In the past, such feature [was implemented](https://github.com/web-token/jwt-framework/blob/v1.3.10/src/Component/KeyManagement/UrlKeySetFactory.php), but removed because of caching/performance issues.

louismariegaborit:
Thanks @nicolas-grekas and @Spomky for review.

You're right. Let's start by supporting JWKSet from a file.
And in review my first draft of this work, I realize that it's not complete. Support JWKSet means that many algorithm signatures could be verify.

The OidcTokenHandler class is final. Can I change the construct signature to replace the type of the first argument with AlgorithmManager or I must keep the type Algorithm and add a deprecate notice to accept only AlgorithManager in 8.0 ?
Which means that we must also update the SecurityBundle to accept an array in the `algorithm` config key.

WDYT ?

stof:
@louismariegaborit you must change the type to a union type of the old and new one, and trigger a deprecation when the old one is passed.

src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/AccessToken/OidcTokenHandlerFactory.php
louismariegaborit: I removed the code for the support of JwkSet in SecurityBundle.  
I think we can do this in another PR ?

louismariegaborit:
> @louismariegaborit you must change the type to a union type of the old and new one, and trigger a deprecation when the old one is passed.

@stof Can I rename the argument ?

louismariegaborit:
As we discuss with @Spomky and @vincentchalamon in the https://github.com/symfony/symfony/pull/53682#issuecomment-1923129583 PR/comment, we propose an update of the oidc_token_handler in the SecurityBundle to authorize JWKSet and multiple algorithms.  

One proposal would be to replace the algorithm and key properties with algorithms and keys to process arrays.

WDYT ? (cc @nicolas-grekas)

louismariegaborit:
@Spomky We are agree that this PR can be closed as the PR #53682 was merged ?

Spomky:
Agreed!

louismariegaborit:
Duplicate #53682

PR #51667, opened 2023-09-15
[HttpClient] typing variables and functions

Tenjiin:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Tickets       | -
| License       | MIT
| Doc PR        | -


nicolas-grekas:
Thanks for the PR. I'm closing because we don't add types to test cases (they're useless) and turning the comment into a real argument is done on 7.0 already - and not on 6.4 because that would break BC.

Issue #51636, opened 2023-09-12
[Console]  strlen(): Argument #1 ($string) must be of type string, array given 


issamkhadiri1989:
### Symfony version(s) affected

6.3.4

### Description

Hello, 

When using a tag with some additional attribute of type array, I got error 

```
[TypeError]
strlen(): Argument #1 ($string) must be of type string, array given
```

### How to reproduce

With the config bellow

```
  App\Mail\SmtpTransport:
    arguments: ['%kernel.project_dir%']
    tags:
      - { name: 'app.mail_transport', alias: 'smtp'}

  App\Mail\SendmailTransport:
    tags:
      - { name: 'app.mail_transport', alias: ['sendmail', 'mail'] }
```

Try to debug the tag using the `php bin/console debug:container --tag=app.mail_transport` 

### Possible Solution

In the file [https://github.com/symfony/framework-bundle/blob/6.3/Console/Descriptor/TextDescriptor.php#L212](https://github.com/symfony/framework-bundle/blob/6.3/Console/Descriptor/TextDescriptor.php#L212)

first need to check whether the `$value` is a type of array.

PR #51669, opened 2023-09-15
[FrameworkBundle] Handle tags array attributes in descriptors

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | https://github.com/symfony/symfony/issues/51636
| License       | MIT
| Doc PR        | -

Targeting 6.3 since array attributes are only possible in tags since 6.2.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/MarkdownDescriptor.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/MarkdownDescriptor.php
@@ -254,7 +254,7 @@ class MarkdownDescriptor extends Descriptor
                 foreach ($tagData as $parameters) {
                     $output .= "\n".'- Tag: `'.$tagName.'`';
                     foreach ($parameters as $name => $value) {
-                        $output .= "\n".'    - '.ucfirst($name).': '.$value;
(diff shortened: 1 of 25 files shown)


src/Symfony/Bundle/FrameworkBundle/Console/Descriptor/MarkdownDescriptor.php
```diff
@@ -254,7 +254,7 @@ protected function describeContainerDefinition(Definition $definition, array $op
                 foreach ($tagData as $parameters) {
                     $output .= "\n".'- Tag: `'.$tagName.'`';
                     foreach ($parameters as $name => $value) {
-                        $output .= "\n".'    - '.ucfirst($name).': '.$value;
+                        $output .= "\n".'    - '.ucfirst($name).': '.(\is_array($value) ? $this->formatParameter($value) : $value);
```
fancyweb, line 257: We can't do a simple explode because we allow array of array of array etc. So I think it's alright to reuse the formatParameter method (see tests for example output).

nicolas-grekas:
Thank you @fancyweb.

PR #51670, opened 2023-09-15
[SecurityBundle] Throw on using both "id" and "migrate_from" option in hasher config

ogizanagi:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Tickets       | N/A... 
| License       | MIT
| Doc PR        | N/A

I stumbled upon this DX issue lately when trying to define **a custom hasher service** while handling password migration with a legacy one using `migrate_from`.

Such a config:

```yml
security:
    # https://symfony.com/doc/current/security.html#registering-the-user-hashing-passwords
    password_hashers:
        legacy: md5
        App\User\Identity:
            id: App\Security\CustomHasher
            migrate_from:
                - legacy
```

will not indicate any issue to the developper, while the `migrate_from` is completely ignored and will not wrap the `CustomHasher` into a `MigratingPasswordHasher` with the `legacy` extra hasher ; which means the password migration will not work as expected.

➜ This PR suggests to throw a configuration exception with a hint regarding this.

Perhaps it would be great to make this possible for Symfony 7.1?


src/Symfony/Bundle/SecurityBundle/DependencyInjection/MainConfiguration.php
```diff
@@ -435,6 +435,10 @@ private function addPasswordHashersSection(ArrayNodeDefinition $rootNode): void
                         ->canBeUnset()
                         ->performNoDeepMerging()
                         ->beforeNormalization()->ifString()->then(fn ($v) => ['algorithm' => $v])->end()
+                        ->validate()
+                            ->ifTrue(fn ($v) => isset($v['migrate_from'], $v['id']) && 0 !== \count($v['migrate_from']))
+                            ->thenInvalid('You cannot use "migrate_from" when using a custom service id.')
```
stof, line 440: As it was silently ignored before, it would be better to trigger a deprecation than throwing an error, to avoid breaking projects in 6.4

chalasr:
What issue could arise if we just make it work as a bugfix instead?

ogizanagi:
> What issue could arise if we just make it work as a bugfix instead?

@chalasr : Do you mean applying this patch to lower branches as well, or making the configuration work with custom hasher ids?

src/Symfony/Bundle/SecurityBundle/DependencyInjection/MainConfiguration.php
ogizanagi: Since it would mean such projects are configured in a way the developers think it'll handle the password migration properly, but is not, making them aware of the issue the hard way during upgrade might be better? 

The fix for removing this exception is straightforward enough, unless this config is hidden somehow in a dependency or whatever (but I would expect the issue would have been already spotted in such a case).

chalasr:
@ogizanagi I mean making it work as it is the expected behavior to me. Even if we just didn’t think about it, if making it work doesn’t have much potential to break userland code then I think we should do it instead of deprecating to make it work later.

ogizanagi:
@chalasr : Alright, done in https://github.com/symfony/symfony/pull/51686

PR #51674, opened 2023-09-16
[Scheduler] Match next run timezone with "from" timezone

valtzu:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

As discussed in https://github.com/symfony/symfony/pull/51651#discussion_r1326994462, when a datetime object is created from unix timestamp, the timezone constructor argument is ignored as demonstrated in https://onlinephp.io/c/b07d1 and also mentioned in [PHP documentation](https://www.php.net/manual/en/datetime.construct.php#refsect1-datetime.construct-parameters):

> The $timezone parameter and the current timezone are ignored when the $datetime parameter either is a UNIX timestamp (e.g. @946684800) or specifies a timezone (e.g. 2010-01-28T15:00:00+02:00).

This change shouldn't break any existing logic, given the places where this time is used already include timezone in the date format string.

### Changes in effect

```diff
  ------------- ------------------------------------ --------------------------- 
   Message       Trigger                              Next Run                   
  ------------- ------------------------------------ --------------------------- 
-  TestMessage   PeriodicalTrigger: every 2 seconds   2023-09-16T15:54:46+00:00  
+  TestMessage   PeriodicalTrigger: every 2 seconds   2023-09-16T18:54:46+03:00  
  ------------- ------------------------------------ --------------------------- 
```

### DIFF
--- a/src/Symfony/Component/Scheduler/Tests/Trigger/PeriodicalTriggerTest.php
+++ b/src/Symfony/Component/Scheduler/Tests/Trigger/PeriodicalTriggerTest.php
@@ -23,9 +23,9 @@ class PeriodicalTriggerTest extends TestCase
      */
     public function testConstructor(PeriodicalTrigger $trigger, bool $optimizable = true)
     {
-        $run = new \DateTimeImmutable('2922-02-22 13:34:00+00:00');
+        $run = new \DateTimeImmutable('2922-02-22 12:34:00+00:00');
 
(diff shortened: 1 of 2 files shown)


fabpot (approved):

fabpot:
Thank you @valtzu.

PR #51675, opened 2023-09-17
[Messenger] Fix cloned TraceableStack not unstacking the stack independently 

### LINKED ISSUE #51564: cloned TraceableStack operates differently than StackMiddleware

6.x

### Description

The `StackMiddleware` has it's own state (`$offset` field), which changes during the operation flow. Therefore, if at some point of time we clone this stack, we get a completely different instance, which may be safely used.

The problem lies in the absence of `TraceableStack::__clone()` method . 
It is not possible to use cloned stack after main stack _has been iterated over_, since internally it references the same object.

See provided test for details.

### How to reproduce

Below two tests are provided. First passess (`StackMiddleware`), second fails (`TraceableStack`).

```php
class_exists(TraceableMiddleware::class);

final class TraceableStackCloneTest extends TestCase
{
    /** @dataProvider stackDataProvider */
    public function testClonedStackFramesFollowOriginalFrames(StackInterface $stack): void
    {
        $clonedStack = clone $stack;

        $nextFrame = $stack->next();
        $clonedStackNextFrame = $clonedStack->next();

        self::assertSame($nextFrame, $clonedStackNextFrame);
    }

    public function stackDataProvider(): array
    {
        $middlewareIterator = [
            $this->createMock(MiddlewareInterface::class),
            $this->createMock(MiddlewareInterface::class),
        ];

        return [
            [new StackMiddleware($middlewareIterator)], // passes
            [new TraceableStack(new StackMiddleware($middlewareIterator), $this->createMock(Stopwatch::class), 'command.bus', 'test')], // fails
        ];
    }
}
```

### Possible Solution

I guess adding `__clone` method to `TraceableStack` would fix the issue:

```php
    public function __clone(): void
    {
        $this->stack = clone $this->stack;
    }
```

### Additional Context

PHP version: 8.2.8
Symfony messenger version: 6.2.7

krciga22:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #51564
| License       | MIT

Fixed a bug with cloned `TraceableStack` not unstacking the stack independently from the original due to the __clone() method not yet being implemented. Clones of `StackMiddleware` currently unstack the stack independently, but not `TraceableStack`.

### DIFF
--- a/src/Symfony/Component/Messenger/Tests/Middleware/TraceableMiddlewareTest.php
+++ b/src/Symfony/Component/Messenger/Tests/Middleware/TraceableMiddlewareTest.php
@@ -16,6 +16,7 @@ use Symfony\Component\Messenger\Middleware\MiddlewareInterface;
 use Symfony\Component\Messenger\Middleware\StackInterface;
 use Symfony\Component\Messenger\Middleware\StackMiddleware;
 use Symfony\Component\Messenger\Middleware\TraceableMiddleware;
+use Symfony\Component\Messenger\Middleware\TraceableStack;
 use Symfony\Component\Messenger\Test\Middleware\MiddlewareTestCase;
 use Symfony\Component\Messenger\Tests\Fixtures\DummyMessage;
 use Symfony\Component\Stopwatch\Stopwatch;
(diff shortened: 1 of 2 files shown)


stof (approved):

maxbeckers (approved):

src/Symfony/Component/Messenger/Tests/Middleware/TraceableMiddlewareTest.php
```diff
@@ -140,4 +141,92 @@ public function handle(Envelope $envelope, StackInterface $stack): Envelope
         $traced->handle($envelope, new StackMiddleware(new \ArrayIterator([null, $middleware])));
         $this->assertSame(1, $middleware->calls);
     }
+
+    public function testClonedTraceableStackUnstacksIndependently()
+    {
+        // import TraceableStack
+        class_exists(TraceableMiddleware::class);
```
nicolas-grekas, line 148: why is this needed?

krciga22: @nicolas-grekas, I added this test method to test if a clone can be unstacked independently. I noticed that the existing method `testHandle()` in the same file does test `TraceableStack` unstacking via the use of `TraceableMiddleware`, but it doesn't test if a cloned version will unstack the stack independently, so I was thinking their could be a new test for that as well?

nicolas-grekas: I meant the class_exists check, why is it needed?

krciga22: Right, without it I get the following error: 
`Error: Class 'Symfony\Component\Messenger\Middleware\TraceableStack' not found`

To fix the error, I tried to import `TraceableStack` at the top of the test file, but I can't get it to work because it's located within `TraceableMiddleware.php` instead of within its own file.

nicolas-grekas:
Thank you @krciga22.

PR #51676, opened 2023-09-18
[RateLimiter] Add SlidingWindowLimiter::reserve()

### LINKED ISSUE #40289: [RateLimiter] retry-after of SlidingWindowLimiter is incorrect

### LINKED ISSUE #46875: [Bug][RateLimiter] `retryAfter` returns bad value

6.1.2

### Description

Based on this part of the documentation : https://symfony.com/doc/current/rate_limiter.html#exposing-the-rate-limiter-status
We can retrieve the statuses of the RateLimiter.

```php
$limiter = $anonymousApiLimiter->create($request->getClientIp());
$limit = $limiter->consume();
$headers = [
    'X-RateLimit-Remaining' => $limit->getRemainingTokens(),
    'X-RateLimit-Retry-After' => $limit->getRetryAfter()->getTimestamp(),
    'X-RateLimit-Limit' => $limit->getLimit(),
];
```

But the retryAfter does not return interesting information in the current case in my opinion.

Take for example this configuration with **fixed_window** rate limiter :

**rate_limiter.yaml**
```yaml
framework:
    rate_limiter:
        custom_thing:
            policy: 'fixed_window'
            limit: 5
            interval: '60 minutes'
```

And we use it : 

```php
    public function myControllerAction(RateLimiterFactory $customThingLimiter): void
    {
        $limiter = $customThingLimiter->create('custom_thing_limit');

        $limit = $limiter->consume();

        $headers = [
            'X-RateLimit-Remaining' => $limit->getRemainingTokens(),
            'X-RateLimit-Retry-After' => $limit->getRetryAfter()->getTimestamp(),
            'X-RateLimit-Limit' => $limit->getLimit(),
        ];

        dump($headers);

      //...
    }
```

Now let's start with this:

* We can make 5 requests by hour

1. The first request of the interval begin at **10:05**
2. The last request of the interval begin at **10:45**

From this moment, the user is supposed to be able to make 5 new requests 1 hour after making the very 1st request of the interval, so : **11:05AM**

During the 5 requests, the value of `$limit->getRetryAfter()` will be the current date, which is OK.

If after my 5 requests, I try a new request (Who therefore should not be accepted), the value of `$limit->getRetryAfter()` should be something like : `2022-07-07 11:05` because I started the first interval request at 10:05.

image

But, his value will be something like : `2022-07-07 11:45`. It just added 1 hour from the last request date, and not from the first interval request date.

Now, it's 10:50, I still shouldn't be able to reapply. If I try one, it won't be consumed. The value of `$limit->getRetryAfter()` should be, a new time `11:05`. But it will be `11:50` (One hour after last request date).

The value of retryAfter is meant to represent the date from which a user is expected to be able to make at least one request again.

But, as we can see, that will not be the case.

### How to reproduce

Follow the reproducer : https://github.com/bastien70/symfony-rate-limiter/tree/main
Instructions are on the README.

Here's what you might get as results from first request : 

1. 
image

2. 
image

3. 
image

4. 
image

5. 
image

At this moment, the **retryAfter** should already be the date of the 1st request, plus 1 hour, so `15:30:51`.

6. 
image

At this moment, it just added 1 hour to the CURRENT date (should be to the FIRST interval request date)

7. 
image

A new time, it just added 1 hour to the CURRENT date

Same thing for next tries.

### Possible Solution

1. Keep in memory the exact date of the 1st request of the interval.

2. When the last request is consumed, then that's when it should add 1 hour to the retryAfter, but still from the 1st request.

3. If the user retries, then the retryAfter date will not change until the interval is over.

### Additional Context

_No response_

Jeroeny:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | yes
| Deprecations? | yes
| Tickets | Fix https://github.com/symfony/symfony/issues/40289, Fixes #46875
| License       | MIT

This implements `LimiterInterface->reserve()` for `SlidingWindowLimiter`. I'm not sure if the lack of implementation was due to time/scope or if it's actually not possible and my approach is incorrect. But I like to give it a try anyway. Perhaps @wouterj you could elaborate on that? 

The calculation does the following:

1. Calculate tokens to be released within this window. E.g. if 4 were used in the last window, at 50% into the current, 2 are still to be released.
2. Calculate the time-per-token within the remainder of the window. If the requested tokens will be available before the end of the current window, return the time-per-token * needed-tokens. 
3. Otherwise return time-per-token of the regular interval * needed-tokens(-after the current window). 

This wasn't a [bugfix](https://github.com/symfony/symfony/pull/51592), so back to feature PR. I couldn't reopen https://github.com/symfony/symfony/pull/47557, So had to create this new PR.

### DIFF
--- a/UPGRADE-6.4.md
+++ b/UPGRADE-6.4.md
@@ -123,6 +123,11 @@ PsrHttpMessageBridge
 
  * Remove `ArgumentValueResolverInterface` from `PsrServerRequestResolver`
 
+RateLimiter
+-----------
+
+ * Deprecate `SlidingWindow::getRetryAfter`, use `SlidingWindow::calculateTimeForTokens` instead
+
 Routing
 -------
 
--- a/src/Symfony/Component/RateLimiter/CHANGELOG.md
+++ b/src/Symfony/Component/RateLimiter/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+ * Add `SlidingWindowLimiter::reserve()`
+
 6.2
 ---
 
--- a/src/Symfony/Component/RateLimiter/Policy/SlidingWindow.php
+++ b/src/Symfony/Component/RateLimiter/Policy/SlidingWindow.php
@@ -84,9 +84,36 @@ final class SlidingWindow implements LimiterStateInterface
         return (int) floor($this->hitCountForLastWindow * (1 - $percentOfCurrentTimeFrame) + $this->hitCount);
     }
 
+    /**
+     * @deprecated since Symfony 6.4, use {@see self::calculateTimeForTokens} instead
+     */
     public function getRetryAfter(): \DateTimeImmutable
     {
-        return \DateTimeImmutable::createFromFormat('U.u', sprintf('%.6F', $this->windowEndAt));
+        trigger_deprecation('symfony/ratelimiter', '6.4', 'The "%s()" method is deprecated, use "%s::calculateTimeForTokens" instead.', __METHOD__, self::class);
+
+        return \DateTimeImmutable::createFromFormat('U.u', sprintf('%.6F', microtime(true) + $this->calculateTimeForTokens(max(1, $this->getHitCount()), 1)));
+    }
+
+    public function calculateTimeForTokens(int $maxSize, int $tokens): float
+    {
+        $remaining = $maxSize - $this->getHitCount();
+        if ($remaining >= $tokens) {
+            return 0;
+        }
+
+        $time = microtime(true);
+        $startOfWindow = $this->windowEndAt - $this->intervalInSeconds;
+        $timePassed = $time - $startOfWindow;
+        $windowPassed = min($timePassed / $this->intervalInSeconds, 1);
+        $releasable = max(1, $maxSize - floor($this->hitCountForLastWindow * (1 - $windowPassed)));
+        $remainingWindow = $this->intervalInSeconds - $timePassed;
+        $needed = $tokens - $remaining;
+
+        if ($releasable >= $needed) {
+            return $needed * ($remainingWindow / max(1, $releasable));
+        }
+
+        return ($this->windowEndAt - $time) + ($needed - $releasable) * ($this->intervalInSeconds / $maxSize);
     }
 
     public function __serialize(): array
--- a/src/Symfony/Component/RateLimiter/Policy/SlidingWindowLimiter.php
+++ b/src/Symfony/Component/RateLimiter/Policy/SlidingWindowLimiter.php
@@ -12,7 +12,7 @@
 namespace Symfony\Component\RateLimiter\Policy;
 
 use Symfony\Component\Lock\LockInterface;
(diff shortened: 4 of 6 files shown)


UPGRADE-6.4.md
```diff
@@ -123,6 +123,11 @@ PsrHttpMessageBridge
 
  * Remove `ArgumentValueResolverInterface` from `PsrServerRequestResolver`
 
+RateLimiter
+-------
+
+* Deprecate `SlidingWindow::getRetryAfter`, use `SlidingWindow::calculateTimeForTokens` instead
```
```suggestion
-----------

 * Deprecate `SlidingWindow::getRetryAfter`, use `SlidingWindow::calculateTimeForTokens` instead
```

src/Symfony/Component/RateLimiter/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+* Add `SlidingWindowLimiter::reserve()`
```
nicolas-grekas:
```suggestion
 * Add `SlidingWindowLimiter::reserve()`
```

src/Symfony/Component/RateLimiter/Policy/SlidingWindow.php
```diff
@@ -84,9 +84,34 @@ public function getHitCount(): int
         return (int) floor($this->hitCountForLastWindow * (1 - $percentOfCurrentTimeFrame) + $this->hitCount);
     }
 
+    /**
+     * @deprecated since Symfony 6.4, use {@see self::calculateTimeForTokens} instead
```
nicolas-grekas, line 88: is the deprecation really required?
if yes, this should be accompanied with a runtime deprecation triggered using symfony/deprecation-contracts

Jeroeny: I think it is a leftover from a previous version of these classes, where maybe the Limiter class used this method. But it doesn't seem to be used now within Symfony. It also has to assume the `$maxSize` and `$tokens` to call `calculateTimeForTokens`. 

If it needs to stay that's fine too. For now I'll add the deprecation trigger.

fabpot (approved):

fabpot:
Thank you @Jeroeny.

PR #51677, opened 2023-09-18
[Annotation] as the function's return type

Tenjiin:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       |  ~
| License       | MIT
| Doc PR        |


stof:
Adding return types is a BC break so it can only happen in a major version. And we already added those in the 7.0 branch.

PR #47003, opened 2022-07-21
[Cache] Ensured that redis adapter can use multiple redis sentinel hosts

warslett:
This PR ensures that RedisAdapter for Symfony Cache component works as expected when a redis_sentinel argument is provided with multiple redis sentinel hosts. Previously if using the redis php extension and multiple hosts were provided RedisTrait::createConnection would ignore the redis_sentinel parameter and instead return a RedisArray object which would not work because it would be an array of sentinel hosts rather than redis hosts.

This PR introduces the correct behaviour where by RedisTrait::createConnection will loop through each of the provided sentinel hosts trying each one to get the address of the mast redis instance. If none of the provided hosts can return the address of a master instance it will throw an exception.

I'm not too sure how to test this behaviour. There doesn't appear to be any tests for redis sentinel behaviour in place currently. I have tested manually both with a multi redis sentinel setup and with a single standard redis instance and the code now functions as expected in both scenarios.

| Q             | A
| ------------- | ---
| Branch?         |  5.4
| Bug fix?      | yes/no
| New feature?  | no - bugfix
| Deprecations? | no - behaviour as documented and as implemented in 4.4
| Tickets       | [#46998](https://github.com/symfony/symfony/issues/46998)
| License       | MIT
| Doc PR        |  none - bugfix

PR #51683, opened 2023-09-18
[Cache] Fix support for Redis Sentinel using php-redis 6.0.0

Qonstrukt:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #51678 
| License       | MIT

Added support for [php-redis 6.0.0](https://github.com/phpredis/phpredis/blob/develop/CHANGELOG.md#600---2023-09-09-github-pecl) to `Symfony/Component/Cache/Traits/RedisTrait`.

Made changes to how `RedisSentinel` is constructed, which now uses [the same constructor](https://github.com/phpredis/phpredis/commit/ebb2386e) as the main `Redis` class. Didn't find any other changes that might be relevant yet, but it's why I mention Redis Sentinel explicitly.

PR #51687, opened 2023-09-18
[Messenger] Add support for multiple Redis Sentinel hosts

digilist:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Similar to #47003 which added support for multiple Redis Sentinel hosts for the Cache component, this PR adds support for multiple Sentinel hosts for the Messenger component.

This PR is inspired by the implementation in the cache component and works very similar. A DSN could look like this: `redis:?host[localhost:26377]&host[localhost:26379]&sentinel_master=db`.

I changed the Sentinel host environment variable for the ingegration to an invalid host at. As a result I noticed that Relay also fails in such case and so I expanded my earlier changes from #51598 to also ignore unreachable hosts with the Relay extension.

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Redis/Tests/Transport/RedisExtIntegrationTest.php
+++ b/src/Symfony/Component/Messenger/Bridge/Redis/Tests/Transport/RedisExtIntegrationTest.php
@@ -220,6 +220,35 @@ class RedisExtIntegrationTest extends TestCase
         $connection->ack($message['id']);
     }
 
+    public function testSentinel()
+    {
+        if (!$hosts = getenv('REDIS_SENTINEL_HOSTS')) {
+            $this->markTestSkipped('REDIS_SENTINEL_HOSTS env var is not defined.');
+        }
+
+        if (!getenv('MESSENGER_REDIS_SENTINEL_MASTER')) {
+            $this->markTestSkipped('MESSENGER_REDIS_SENTINEL_MASTER env var is not defined.');
+        }
+
+        $dsn = 'redis:?host['.str_replace(' ', ']&host[', $hosts).']';
+
+        $connection = Connection::fromDsn($dsn,
+            ['delete_after_ack' => true,
+             'sentinel_master' => getenv('MESSENGER_REDIS_SENTINEL_MASTER') ?: null,
+            ], $this->redis);
+
+        $connection->add('1', []);
+        $this->assertNotEmpty($message = $connection->get());
+        $this->assertSame([
+            'message' => json_encode([
+                'body' => '1',
+                'headers' => [],
+            ]),
+        ], $message['data']);
+        $connection->reject($message['id']);
+        $connection->cleanup();
+    }
+
     public function testLazySentinel()
     {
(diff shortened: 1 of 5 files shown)


src/Symfony/Component/Messenger/Bridge/Redis/Tests/Transport/RedisExtIntegrationTest.php
```diff
@@ -11,6 +11,7 @@
 
 namespace Symfony\Component\Messenger\Bridge\Redis\Tests\Transport;
 
+use PHPUnit\Framework\SkippedTestSuiteError;
```
nicolas-grekas: we prefer using `$this->markTestSkipped()` instead for this

nicolas-grekas (commented):
LGTM thanks, I just have one minor comment

src/Symfony/Component/Messenger/Bridge/Redis/Tests/Transport/RedisExtIntegrationTest.php
digilist: Makes sense, I've adjusted this.

nicolas-grekas:
Thank you @digilist.

nicolas-grekas:
@digilist there are [some failures](https://github.com/symfony/symfony/actions/runs/6339900320/job/17220094382#step:11:47) on our integration tests that might be related to this PR. Would you have some availability to have a look please? :pray:

src/Symfony/Component/Messenger/Bridge/Redis/Tests/Transport/RedisExtIntegrationTest.php
stof: `SkippedTestSuiteError` is an internal API of PHPUnit while `markTestSkipped()` is the public API btw. So `SkippedTestSuiteError` should never be used when writing tests (in any project, not just Symfony)

digilist: Yes, good point. I actually only copy and pasted it from the Redis tests in the Cache component in `src/Symfony/Component/Cache/Tests/Traits/RedisTraitTest.php` :sweat_smile:

digilist:
Yes, sure. It looks like a merge conflict between #51683 and this PR was resolved incorrectly, as there are now two connection attempts to Sentinel:

https://github.com/symfony/symfony/blob/3c4cfbd7f129580d8cbc823b85817f137a178790/src/Symfony/Component/Cache/Traits/RedisTrait.php#L226-L253

I'll provide a PR.

pepeh:
@nicolas-grekas, why is this change added to Symfony 6.4/7.0, but a similar fix for the cache component is released on 5.4? https://github.com/symfony/symfony/pull/47003
Right now, Symfony cache supports new ext-redis, but Symfony messenger does not.

Issue #35316, opened 2020-01-12
[RFC][Validator] Move the filtering/sanitation/normalization out of the validators


sylfabre:
Following up on https://github.com/symfony/symfony/issues/28232

**Addressed issues**
1. Some validators alter the data they are validating before running the validation.
For instance, IbanValidator will remove any space and capitalize the string it is given before running validation.

2. The behavior described in 1 is not consistent across validators.
For instance, IbanValidator accepts lower-cased strings while BicValidator refuses them.

3. The behavior described in 1 breaks the Single Responsibility principle as the role of the validator is to validate, not sanitize nor filter nor normalize the data it is working on.

**Description**
Filtering some common accepted formattings is good UX (like stripping space off an IBAN) as it widens the acceptable data set, so it must be kept. The related use-case are forms written with Form component.

Each filtering should be placed in dedicated classes implementing the DataTransformInterface of the Form component.

Dedicated FormTypes should be created and register these DataTransform implementation (like `\Symfony\Component\Form\Extension\Core\Type\NumberType` does with both `NumberToLocalizedStringTransformer` and `\Symfony\Component\Form\Extension\Core\DataTransformer\StringToFloatTransformer`)

In the next minor version, validators should have an option `strict` set to `false` by default to preserve BC. Setting this option to `true` should skip normalization of the data under validation.

In the next major version:
- Validators should drop normalization and only run strict validation.
- Developers should use the new types in their forms. 

Identified validators with this flaw:
- BicValidator (removes spaces)
- IsbnValidator (removes dashes)
- IbanValidator (removes spaces, capitalize) 
- LocaleValidator (because of the `canonicalize` option)

All validators with a `normalizer` option should be updated too (cf https://github.com/symfony/validator/commit/fdf8dfc7193c60162bc375e455ba7ec830c899ce#diff-5b9cb67322b0684885ea5ae77e6932da) to remove this `normalizer`

UuidValidator already has a `strict` option which may be out of the scope: I'm just wondering what is the point of accepting non-compliant UUIDs.

**Example**  

For instance, we could rely on the following code for IBANs.

```
<?php

class StringToIbanTransformer implements DataTransformerInterface
{
    public function transform($value)
    {
        if (null === $value) {
            return null;
        }

        if (!\is_string($value)) {
            throw new TransformationFailedException('Expected a string.');
        }
    
        return str_replace(' ', '', strtoupper($value));
    }
    
    public function reverseTransform($value)
    {
        if (null === $value) {
[...]

PR #51689, opened 2023-09-18
[Validator] [Language] Add case-insensitive option to Validator\Constraints\Language

### LINKED ISSUE #49615: LanguageValidator case sensitive

6.1.4

### Description

LanguageValidator is case-sensitive, while writing `EN`, `en`, `DE`, `de` are valid.

### How to reproduce

- create field that is validated by `\Symfony\Component\Validator\Constraints\Language`
- validate object/model with this field, try: `EN` validation won't pass, try `en` everything will be okay.

`\Symfony\Component\Intl\Languages::getAlpha3Codes` is returning keys are lower-case.

### Possible Solution

- Add option ignore case and if it's `TRUE`, normalize case (to match \Symfony\Component\Intl\Languages::getAlpha3Codes)

### Additional Context

_No response_

gpetraroli:
Add case-insensitive option to Validator\Constraints\Language

| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #49615
| License       | MIT
| Doc PR        | symfony/symfony-docs/pull/18884

Language validator now accepts case-insensitive language codes if option is set to true (es. it, IT, iT, It are all valid).
Use: pass caseInsensitive => true to Language constraint.
By default caseInsensitive is false.

This fix issue #49615

### DIFF
--- a/src/Symfony/Component/Validator/CHANGELOG.md
+++ b/src/Symfony/Component/Validator/CHANGELOG.md
@@ -15,6 +15,7 @@ CHANGELOG
  * Deprecate `ValidatorBuilder::enableAnnotationMapping()`, use `ValidatorBuilder::enableAttributeMapping()` instead
  * Deprecate `ValidatorBuilder::disableAnnotationMapping()`, use `ValidatorBuilder::disableAttributeMapping()` instead
  * Deprecate `AnnotationLoader`, use `AttributeLoader` instead
+ * Add `caseInsensitive` option for `Validator\Constraints\Language`
 
 6.3
 ---
--- a/src/Symfony/Component/Validator/Constraints/Language.php
+++ b/src/Symfony/Component/Validator/Constraints/Language.php
@@ -37,11 +37,13 @@ class Language extends Constraint
 
     public $message = 'This value is not a valid language.';
     public $alpha3 = false;
+    public $caseInsensitive = false;
 
     public function __construct(
         array $options = null,
         string $message = null,
         bool $alpha3 = null,
+        bool $caseInsensitive = null,
         array $groups = null,
         mixed $payload = null
     ) {
@@ -53,5 +55,6 @@ class Language extends Constraint
 
         $this->message = $message ?? $this->message;
         $this->alpha3 = $alpha3 ?? $this->alpha3;
+        $this->caseInsensitive = $caseInsensitive ?? $this->caseInsensitive;
     }
 }
--- a/src/Symfony/Component/Validator/Constraints/LanguageValidator.php
+++ b/src/Symfony/Component/Validator/Constraints/LanguageValidator.php
@@ -43,6 +43,8 @@ class LanguageValidator extends ConstraintValidator
 
         $value = (string) $value;
 
+        $value = $constraint->caseInsensitive ? mb_strtolower($value, 'UTF-8') : $value;
+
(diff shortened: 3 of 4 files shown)


javiereguiluz:
I'm not sure about this proposal because:

* ISO spec for countries/territories use **uppercase** (https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes)
* ISO spec for languages use **lowercase** (https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)

_"ISO-639 is a standardized nomenclature used to classify languages. Each language is assigned a two-letter (639-1) and three-letter (639-2 and 639-3) **lowercase** abbreviation, amended in later versions of the nomenclature."_

@gpetraroli could you please give some more context about who gives you a language code in uppercase? Some browser? Some other well-known app or service? Thanks!

gpetraroli:
This is a solution for the issue 49615
After further investigation on iso.org codes are also represented in uppercase

https://www.iso.org/iso-3166-country-codes.html
https://www.iso.org/obp/ui/#iso:pub:PUB500001:en
https://www.iso.org/obp/ui/#iso:code:3166:IT

javiereguiluz:
Giuseppe, yes the country/territories code (ISO 3166) are always in uppercase ... but here we're talking about the language codes (ISO 639) which are always in lowercase.

What would help us make a decision here is:

* Is this wrong uppercase language code being generated by some popular and well-known app? (e.g. a browser in the HTTP headers)
* Or is it just that in some app of yours some language codes were wrongly in uppercase for some reason? _(if that's the case, we're not judging you 🙏 , all of our own apps have problems and issues with data)_

gpetraroli:
@javiereguiluz I see your point and I agree. I cannot tell if some api are actually breaking the ISO.
This is a possible solution to issue https://github.com/symfony/symfony/issues/49615 opened by **oleg-andreyev**

derrabus:
ping @oleg-andreyev

oleg-andreyev:
I've been parsing XML / JSON from a supplier and we've noticed that they use language in **uppercase** and I have spend some time in understanding why validation is failing.

I haven't been aware that ISO 3166 and ISO 639 strictly define in what case value should be.

Question: **does it really matter?  if it uses strictly defined list from Intl.**

If it does matter, then I'd propose to improve DX a bit and at least give a hint for developers that ISO 3166 is **uppercase** and ISO 639 is **lowercase**

ro0NL:
>  does it really matter?

IMHO kinda yes :) you start leaking non-iso data in your app/persistence layer.

we had a topic on canonicalization: https://github.com/symfony/symfony/issues/35316

> I'd propose to improve DX

i agree it's the best solution currently: a special error message if we detect uppercase language codes

gpetraroli:
@ro0NL I agree with your solution, I can give it a try

OskarStark:
I also agree on a better error message and keep the code as is

gpetraroli:
Thank you @OskarStark
since we can pass a custom message as option wouldn't be more appropriate to just better specify the accepted standard on the doc instead?

Otherwise what kind of message would you show?

nicolas-grekas:
Thanks for the PR. Looking at the discussion, we'd prefer not doing that, so let's close. I'm not convinced this warrants a dedicated message. Closing therefor.

Issue #51691, opened 2023-09-18
AutowireCallable Doesn't Respect Non-Shared Services


miqrogroove:
### Symfony version(s) affected

6.3.3

### Description

Using the examples from [Generate Closures With Autowiring](https://symfony.com/doc/current/service_container/autowiring.html#generate-closures-with-autowiring), I found that the two different attributes don't work the same way.

AutowireServiceClosure returns a _new instance_ on every call when the encapsulated service is non-shared.

AutowireCallable returns the _same instance_ on every call when the encapsulated service is non-shared.

I couldn't figure out whether this is a documentation problem or an oversight in the design of these new attributes.  The docs for the Callable attribute literally said, "a service accepts a closure with a specific signature", and one of the options is named "service", so to me this means the closure should be aware of how services work.

### How to reproduce

Test case added here:

https://github.com/miqrogroove/symfony-bug-51691/tree/main/trunk

At your command line use:

`bin/console app:test`

Also note "Usages" under:

`bin/console debug:container MyService`

Whereas AutowireServiceClosure will show usage by a specific class like `App\Service\Neato`, the AutowireCallable usage shows up like `.1_Closure~lIx8M9C`

OskarStark:
Friendly ping @nicolas-grekas @stof

derrabus:
> AutowireCallable returns the _same instance_ on every call when the encapsulated service is non-shared.

Can you be a bit more specific here? `AutowireCallable` does not return anything. It's a marker that tells the container which dependency to wire into your service.

What is the Closure that you would expect to be wired and what closure did you actually get?

miqrogroove:
@derrabus  Correct.  Let's say, at least for testing, I'm wiring a non-shared service method that returns itself, i.e. `return $this`.

Now, am I to expect repeated calls to the same AutowireCallable will call the specified method of the _same_ non-shared instance every time, or do I expect that each call will encapsulate a _new_ non-shared instance and call the specified method without reading or overwriting the state of any other object?

And the obvious follow-up:  How am I supposed to know which is what?  The documentation makes it sound like these two attributes are both Closures where one conveniently calls a method and the other doesn't.  But testing shows these two Closures don't do the same thing.

derrabus:
> @derrabus Correct. Let's say, at least for testing, I'm wiring a non-shared service method that returns itself, i.e. `return $self`.

A method cannot return itself. 🤔 

[...]

PR #51695, opened 2023-09-19
[DependencyInjection] Ensure AutowireCallable handles non-shared services

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | #51691
| License       | MIT
| Doc PR        | -

@miqrogroove can you please confirm this fixes the issue for you?

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Argument/LazyClosure.php
+++ b/src/Symfony/Component/DependencyInjection/Argument/LazyClosure.php
@@ -13,7 +13,6 @@ namespace Symfony\Component\DependencyInjection\Argument;
 
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
-use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
 use Symfony\Component\DependencyInjection\Exception\RuntimeException;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\VarExporter\ProxyHelper;
@@ -25,26 +24,9 @@ use Symfony\Component\VarExporter\ProxyHelper;
  */
 class LazyClosure
 {
-    public readonly object $service;
-
     public function __construct(
-        private \Closure $initializer,
+        protected readonly \Closure $service,
     ) {
-        unset($this->service);
-    }
-
-    public function __get(mixed $name): mixed
-    {
-        if ('service' !== $name) {
-            throw new InvalidArgumentException(sprintf('Cannot read property "%s" from a lazy closure.', $name));
-        }
-
-        if (isset($this->initializer)) {
-            $this->service = ($this->initializer)();
-            unset($this->initializer);
-        }
-
-        return $this->service;
     }
 
     public static function getCode(string $initializer, array $callable, Definition $definition, ContainerBuilder $container, ?string $id): string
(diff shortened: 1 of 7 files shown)


src/Symfony/Component/DependencyInjection/ContainerBuilder.php
```diff
@@ -1062,9 +1062,17 @@ private function createService(Definition $definition, array &$inlineServices, b
                 $callable[0] instanceof Reference
                 || $callable[0] instanceof Definition && !isset($inlineServices[spl_object_hash($callable[0])])
             )) {
-                $initializer = function () use ($callable, &$inlineServices) {
-                    return $this->doResolveServices($callable[0], $inlineServices);
-                };
+                if ($callable[0] instanceof Definition && $callable[0]->isShared()) {
```
stof, line 1065: if `$callable[0]` is a reference, shouldn't we also check whether this reference targets a shared service ?

src/Symfony/Component/DependencyInjection/Argument/LazyClosure.php
```diff
@@ -25,26 +24,9 @@
  */
 class LazyClosure
 {
-    public readonly object $service;
```
stof, line 28: isn't it a BC break to change the property from public to protected ?

src/Symfony/Component/DependencyInjection/ContainerBuilder.php
nicolas-grekas: if it's a reference, we already dump code that does the memoization

src/Symfony/Component/DependencyInjection/Argument/LazyClosure.php
nicolas-grekas: `@internal` to the rescue ;)

miqrogroove:
One of us broke something.

```
Cannot autowire service "App\Controller\MyController": argument "$myCallable" of method "__construct()" references class "Closure" but no such service exists.
```

I haven't changed the property syntax so...

```
    public function __construct(
        #[AutowireCallable(service: 'App\NonShared\MyService', method: 'myMethod')]
        private \Closure $myCallable,
    ) {
    }
```

I skipped the Test patches.  I was hoping it would just work 🙂

nicolas-grekas:
Do you have the corresponding `use` statement for the attribute?

miqrogroove:
Adding a `use` statement doesn't change anything for the Fully Qualified Name of `\Closure`.  I tried it just to make sure.  No difference.

miqrogroove:
Sorry I rushed that test 😊 The attribute's `use` was needed.

Now, it seems like I'm still getting the same instance for each call to the closure...

nicolas-grekas:
Can you please share a reproducer, or better: a test case?

miqrogroove:
Yes.  If it takes a while to make a clean test case then I will follow up tonight.

stof:
Personally, I think the behavior described in https://github.com/symfony/symfony/issues/51691#issuecomment-1725873377 by @derrabus would make sense.

So in your test, if you have a single service in which you use AutowireCallable on a single argument, it would still be expected that the same instance is used.

miqrogroove:
Testcase https://github.com/miqrogroove/symfony-bug-51691/tree/main/trunk

`bin/console app:test`

nicolas-grekas:
Thanks. Did you forget to commit some files? There's no bin/console file there :)

miqrogroove:
> So in your test, if you have a single service in which you use AutowireCallable on a single argument, it would still be expected that the same instance is used.

It would be a documentation issue then.  If that's the conclusion then the PR wouldn't be needed.

miqrogroove:
> Thanks. Did you forget to commit some files? There's no bin/console file there :)

Added bin/console.

nicolas-grekas:
Thanks. So we currently generate something like this:
```php
$container->privates['App\\Command\\TestCommand'] = $instance = new \App\Command\TestCommand((new \App\NonShared\MyService())->setter(...))
```

This is the expected behavior to me. The service is non-shared with other services, but it's still the same one that is reused by individual consumer services when used as a callable.

A service-closure is different in the sense that it's a sort of factory for a service, so it's expected that it returns a new instance for non-shared ones.

Closing therefor, thanks for the help understanding this.

PR #51696, opened 2023-09-19
[FrameworkBundle] Add `LoggerAssertionsTrait` which provide shortcuts to assert a log has been written

welcoMattic:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT

Like #50662, this PR brings a new collection to enhance assertions to help users to ensure some logs has been written during a Request processing.

In a controller:
```php
public function index(LoggerInterface $logger)
{
    $logger->debug('test1_' . __CLASS__);
    $logger->debug('test2_' . __CLASS__);
    $logger->debug('test3_' . __CLASS__);

    return new Response();
}
```

In the test:

```php
public function testLoggerAssertion()
{
    $client = $this->createClient();
    $client->request('GET', '/log');

    $this->assertLogExists('test1_Symfony\Bundle\FrameworkBundle\Tests\Functional\Bundle\TestBundle\Controller\LoggerController');
    $this->assertLogMatches("/(test2_).*(LoggerController)/");
    $this->assertLogContains('test3');
}
```

## Todo

- [ ] Write doc

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Test/LoggerAssertionsTrait.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Test/LoggerAssertionsTrait.php
@@ -1,0 +1,48 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\FrameworkBundle\Test;
+
+use Monolog\Handler\TestHandler;
+use Monolog\Logger;
+use Monolog\LogRecord;
+
+trait LoggerAssertionsTrait
+{
+    public static function assertLogExists(string $expectedLog, string $level = Logger::DEBUG): void
+    {
+        /** @var TestHandler $logger */
+        $logger = self::getContainer()->get('monolog.handler.test');
+
+        self::assertTrue($logger->hasRecordThatPasses(
+            function (array|LogRecord $record) use ($expectedLog) {
+                return $record['message'] === $expectedLog;
+            },
+            $level,
+        ));
+    }
+
+    public static function assertLogMatches(string $expectedRegex, string $level = Logger::DEBUG): void
+    {
+        /** @var TestHandler $logger */
+        $logger = self::getContainer()->get('monolog.handler.test');
+
+        self::assertTrue($logger->hasRecordThatMatches($expectedRegex, $level));
+    }
+
+    public static function assertLogContains(string $expectedLog, string $level = Logger::DEBUG): void
+    {
+        /** @var TestHandler $logger */
+        $logger = self::getContainer()->get('monolog.handler.test');
+
+        self::assertTrue($logger->hasRecordThatContains($expectedLog, $level));
+    }
+}
--- a/src/Symfony/Bundle/FrameworkBundle/Test/LoggerAssertionsTrait.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Test/LoggerAssertionsTrait.php
@@ -19,6 +19,8 @@ trait LoggerAssertionsTrait
 {
     public static function assertLogExists(string $expectedLog, string $level = Logger::DEBUG): void
     {
+        self::ensureMonologHandlerIsAvailable();
+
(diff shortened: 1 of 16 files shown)


ismail1432:
Good idea!

src/Symfony/Bundle/FrameworkBundle/Test/LoggerAssertionsTrait.php
```diff
@@ -0,0 +1,48 @@
...
+use Monolog\Handler\TestHandler;
+use Monolog\Logger;
+use Monolog\LogRecord;
+
+trait LoggerAssertionsTrait
+{
+    public static function assertLogExists(string $expectedLog, string $level = Logger::DEBUG): void
+    {
+        /** @var TestHandler $logger */
+        $logger = self::getContainer()->get('monolog.handler.test');
```
ogizanagi, line 25: Who's in charge of declaring this test handler? The developer? 
There is no such service declared out-of-the-box AFAIK.

Should there be an option to register one, or a section added into the monolog recipe about this?

welcoMattic: Good catch, I've to add it in the app where I made those assertions first.

I can check if it's registered, and if not, display an info message to suggest the developer to register the test handler.

src/Symfony/Bundle/FrameworkBundle/Test/LoggerAssertionsTrait.php
```diff
@@ -40,9 +44,21 @@ public static function assertLogMatches(string $expectedRegex, string $level = L
...
         self::assertTrue($logger->hasRecordThatContains($expectedLog, $level));
     }
+
+    /**
+     * @internal
+     */
+    private static function ensureMonologHandlerIsAvailable(): void
+    {
+        if (!self::getContainer()->has('monolog.handler.test')) {
+            self::fail('The "monolog.handler.test" service is not available. Try registering the service "Monolog\Handler\TestHandler" as "monolog.handler.test" in your test configuration.');
```
ogizanagi, line 61: What about mentioning using the Monolog bundle configuration if available?

```yml
when@test:
    monolog:
        handlers:
            test: { type: test }
            main:
                type: fingers_crossed
                # […]
```

This would work, because it'll be registered through the Monolog bundle extension in the handlers chain, but I don't get how just registering a `monolog.handler.test` would ensure this logger handler is used when using the logger in your application.

ogizanagi (commented):
Shouldn't this be part of the Monolog bundle rather than the FrameworkBundle?
It does only work with Monolog

src/Symfony/Bundle/FrameworkBundle/Test/LoggerAssertionsTrait.php
welcoMattic: You are right, I should belongs to the Monolog bridge, as it only works with Monolog. Or I have to reimplement `hasRecord*` methods for generic loggers (which I do not want to do :smile:).

I'm going to move it, thanks for your feedback

src/Symfony/Bundle/FrameworkBundle/Test/LoggerAssertionsTrait.php
```diff
@@ -0,0 +1,64 @@
...
+
+namespace Symfony\Bundle\FrameworkBundle\Test;
+
+use Monolog\Handler\TestHandler;
+use Monolog\Logger;
+use Monolog\LogRecord;
+
+trait LoggerAssertionsTrait
+{
+    public static function assertLogExists(string $expectedLog, string $level = Logger::DEBUG): void
```
stof, line 20: The signature is wrong. Monolog's level constants on the Logger class are integers, not strings.

And they are deprecated in favor of the enum in 3.x.

and in 3.x, the various `hasRecord*` methods of theTestHandler expect the Level enum.

src/Symfony/Bundle/FrameworkBundle/Test/LoggerAssertionsTrait.php
welcoMattic: I don't well know this part, especially how to use this bridge in tests. Would you help me?

Issue #29933, opened 2019-01-18
[PropertyInfo] ReflectionExtractor not supporting snake_case properties


antonmakarenko:
**Symfony version(s) affected**: 3.4.18

**Description** 

When extracting metadata about properties, the reflection extractor fails to recognize setters/getters for class properties that have snake_case.

`\Symfony\Component\PropertyInfo\Extractor\ReflectionExtractor`
`getMutatorMethod()` and `getAccessorMethod()` have the same logic:
```
$ucProperty = ucfirst($property);
```

[According to PSR-1](https://www.php-fig.org/psr/psr-1/#42-properties), there is no restriction about property names - they intentionally omit any provisions whether to use SturdlyCase, camelCase or snake_case.

At the same time, PSR-1 states that methods must be in camelCase. 

**How to reproduce**  
```
<?php
use Symfony\Component\PropertyInfo\Extractor\ReflectionExtractor;

require __DIR__ . '/vendor/autoload.php';

class MyEntity
{
    /** @var \DateTime */
    private $foo_bar;

    public function getFooBar(): DateTime
    {
        return $this->foo_bar;
    }

    public function setFooBar(DateTime $value): void
    {
        $this->foo_bar = $value;
    }
}

$extractor = new ReflectionExtractor();
$extractor->getTypes('MyEntity', 'foo_bar');
// returns null
// attempts setFoo_bar()

```

**Possible Solution**  

For best compatibility with PSR-1, the extractor should attempt methods that do not contain "_".
E.g. in addition to "setFoo_bar", it should also attempt "setFooBar" (or "setfoobar" - in PHP, method names are not case-sensitive)

antonmakarenko:
This topic, in a more broader sense, was discussed in #16889 - but it was closed with a partial resolution only.

carsonbot:
Hey, thanks for your report!
There has not been a lot of activity here for a while. Is this bug still relevant? Have you managed to find a workaround?

carsonbot:
Could I get a reply or should I close this?

antonmakarenko:
I tested it again on a Symfony 4.4.x setup with symfony/property-info v5.1.8 in composer.lock, and the issue is still reproducing.

The extractor attempts mutator methods as follows:
```
setFoo_bar
removeFoo_bar
addFoo_bar
```

carsonbot:
Hey, thanks for your report!
There has not been a lot of activity here for a while. Is this bug still relevant? Have you managed to find a workaround?

[...]

Issue #16889, opened 2015-12-07
[PropertyInfo] Implement Naming Strategy for Properties


zanderbaldwin:
The [`ReflectionExtractor`](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/PropertyInfo/Extractor/ReflectionExtractor.php) sees a public property as separate from its accessor/mutator methods. For example:

```php
class Website
{
    public $domainName;
    public $author;

    public function getDomainName()
    {
        return $this->domainName;
    }

    public function getAuthor()
    {
        return $this->author;
    }
}

var_dump($reflectionExtractor->getProperties());
/*
array (size=4)
  0 => string 'domainName' (length=10)
  1 => string 'author' (length=6)
  2 => string 'DomainName' (length=10)
  3 => string 'Author' (length=6)
*/
```

The easiest fix would be to lowercase the first letter of the property name determined from a method, though this would cause problems with properties that actually do start with an uppercase letter.

Keeping in mind that it's likely that this component will be used in conjunction with the PropertyAccess component, would it be suitable for a naming strategy to be added (the most obvious being the underscore naming strategy used in both PropertyAccess and Doctrine)?

/cc @dunglas

dunglas:
It's definitely a bug. The first letter must be lower cased.

About the naming strategy, why not adding support for snake_case because as you pointed out it exists in PropertyAccess. But IMO it's better to have be coherent between these 2 components: support only camelCase/snake_case for both or a support for custom strategies for both.

zanderbaldwin:
`snake_case` for PropertyInfo it is. It would be nice to implement naming strategies in both components for edge cases eventually, but I don't see it as a priority.

However, a basic property name transformer might be a good idea to maintain consistency across extractors.

iltar:
Currently doctrine does not fully support snake-case properties: https://github.com/doctrine/collections/pull/29

zanderbaldwin:
Good point.

It's probably best that the component does not take on the role of dictating naming conventions, especially since PSRs actively avoid taking a stance on such issues.

How about implementing a closest-match property name transformer to match identifiers to the correct class properties/methods?

| Identifier | Attempt (in order) |
|---|---|
| `first_name` | `$first_name`, `$firstName`, `{prefix}First_name()`, `{prefix}FirstName()` |
| `firstName` | `$firstName`, `{prefix}FirstName()` |
| `FirstName` | `$FirstName`, `$firstName`, `{prefix}FirstName()` |

[...]

PR #51697, opened 2023-09-19
[PropertyInfo] Make isWriteable() more consistent with isReadable() when checking snake_case properties

jbtronics:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Related with issues #29933 and #16889
| License       | MIT
| Doc PR        | 

Currently PropertyInfo is a bit inconsistent in the behavior of isWriteable() and isReadable() when using it with properties in snake_case format, which are only accessible via a getter and setter. To be readable the getter function has to be in camel_case (e.g. `getSnakeCase()`), while the setter function has to remain in the snake case format (e.g. `setSnake_case()`).

In this example class:
```php
class Dummy {
  private string $snake_case;

  public function getSnakeCase(): string
  { }

  public function setSnakeCase(string $snake_case): void
  { }
 
}
````

the property `snake_case` is considered readable like you would expect, but not writeable, even though the property has a useable setter and the value can be actually be modified fine by something like the PropertyAccess component.
To make the property actually considered writeable, the setter would need to be named `setSnake_case`, which is pretty inconsistent with the behavior of isReadable and makes it very hard to use this component on snake_case properties.

This inconsistencies are caused by the fact, that `isReadable` in ReflectionExtractor uses the `getReadInfo()` function which does a camelization of the property name internally, while the `isWriteable()` function uses the `getMutatorMethod()` function which just perform a capitalization of the first letter.

This PR fixes this inconsistencies between isReadable() and isWriteable() by allowing to use a camelCase style setter to be considered writeable, as this is much more common then to use the mix of snake and camelCase currently required.

The getWriteInfo() function is not useable here, because it needs both a add and remove Function on collection setters to give proper infos, while the current `isWriteable()` implementation considers a class with just a add or a remove function as writeable. Therefore the property name just gets camelized before being feed into the `getMutatorMethod()`, which gives the desired result.

To maximize backwards compatibility, if no camelCase style setter is found, it is still checked for a snake_case setter, so that classes having these, are still considered writeable after the change.

The current behavior is causing some inconsistencies in higher level libraries, which rely on this component. In API Platform for example properties in snake_case are considered read only even though they have a (camel case) setter, because of this. See issue: https://github.com/api-platform/core/issues/5641 and https://github.com/api-platform/core/issues/1554

### DIFF
--- a/src/Symfony/Component/PropertyInfo/Tests/Extractor/ReflectionExtractorTest.php
+++ b/src/Symfony/Component/PropertyInfo/Tests/Extractor/ReflectionExtractorTest.php
@@ -26,6 +26,7 @@ use Symfony\Component\PropertyInfo\Tests\Fixtures\Php74Dummy;
 use Symfony\Component\PropertyInfo\Tests\Fixtures\Php7Dummy;
 use Symfony\Component\PropertyInfo\Tests\Fixtures\Php7ParentDummy;
 use Symfony\Component\PropertyInfo\Tests\Fixtures\Php81Dummy;
+use Symfony\Component\PropertyInfo\Tests\Fixtures\SnakeCaseDummy;
 use Symfony\Component\PropertyInfo\Type;
 
 /**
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/Symfony/Component/PropertyInfo/Tests/Extractor/ReflectionExtractorTest.php [hunk 1]
```diff
@@ -409,6 +410,21 @@ public static function getWritableProperties()
...
+        $this->assertTrue($this->extractor->isReadable(SnakeCaseDummy::class, 'snake_property'));
+        $this->assertTrue($this->extractor->isReadable(SnakeCaseDummy::class, 'snake_readonly'));
+
+    }
+
+    public function testIsWriteableSnakeCase()
+    {
+        $this->assertTrue($this->extractor->isWritable(SnakeCaseDummy::class, 'snake_property'));
+        $this->assertFalse($this->extractor->isWritable(SnakeCaseDummy::class, 'snake_readonly'));
+        //Ensure thaat it is still possible to write to the property using the (old) snake name
```
maxbeckers:
```suggestion
        // Ensure that it's still possible to write to the property using the (old) snake name
```

nicolas-grekas (approved):

nicolas-grekas:
Thank you @jbtronics.

PR #51703, opened 2023-09-20
[PhpUnitBridge] Add some more native types

ghost:
| Q             | A
| ------------- | ---
| Branch?       | 7.0 
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | ~ 
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Symfony/Bridge/PhpUnit/ClockMock.php
+++ b/src/Symfony/Bridge/PhpUnit/ClockMock.php
@@ -19,10 +19,7 @@ class ClockMock
 {
     private static $now;
 
-    /**
-     * @return bool|null
-     */
-    public static function withClockMock($enable = null)
+    public static function withClockMock($enable = null): ?bool
     {
         if (null === $enable) {
             return null !== self::$now;
@@ -33,10 +30,7 @@ class ClockMock
         return null;
     }
 
-    /**
-     * @return int
-     */
-    public static function time()
+    public static function time(): int
     {
         if (null === self::$now) {
             return \time();
@@ -45,10 +39,7 @@ class ClockMock
         return (int) self::$now;
     }
 
-    /**
-     * @return int
-     */
-    public static function sleep($s)
+    public static function sleep($s): int
     {
         if (null === self::$now) {
             return \sleep($s);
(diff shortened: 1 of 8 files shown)


src/Symfony/Bridge/PhpUnit/ClockMock.php
```diff
@@ -71,7 +63,7 @@ public static function usleep($us)
         }
     }
 
-    public static function microtime($asFloat = false)
+    public static function microtime($asFloat = false): string
```
stof: this is a wrong return type

src/Symfony/Bridge/PhpUnit/ClockMock.php
```diff
@@ -108,10 +96,8 @@ public static function gmdate($format, $timestamp = null)
         return \gmdate($format, $timestamp);
     }
 
-    /**
-     * @return array|int|float
-     */
-    public static function hrtime($asNumber = false)
+
+    public static function hrtime($asNumber = false): array|int|float
```
stof: The min PHP version of the phpunit-bridge is PHP 7.2 (this bridge has a special status regarding the min PHP version as we use it to test older maintained branches). So the union type cannot become a native return type.

src/Symfony/Bridge/Twig/Mime/NotificationEmail.php
```diff
@@ -82,9 +80,7 @@ public function markAsPublic(): static
         return $this;
     }
 
-    /**
-     * @return $this
```
stof: this must be kept as it is more specific than `static`. This is true for all methods using it.

stof (changes requested):

src/Symfony/Bridge/PhpUnit/ClockMock.php
ghost: I understand, but what type of float or mixed is recommended?

stof: this should use `@return string|float`, as done for the signature of `microtime`. But this must be done in phpdoc due to the union type.

nicolas-grekas (approved):
I've force-pushed some more native types and reverted the union ones since they're not compatible with PHP 7.2

nicolas-grekas:
Thank you @d-eff-it.

PR #51704, opened 2023-09-20
[Routing] Fix routing collection defaults when adding a new route to a collection

bram123:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | ~
| License       | MIT
| Doc PR        | ~

In our current Symfony project we use PHP routing files, and use routing collections that are setup before adding routes.

Before upgrading to Symfony 6.3 this meant we could add an attribute to the collection, add X routes and they would all have this routing attribute (be it stateless / methods / defaults). After upgrading to S6.3 this no longer works.

PR #49517 removed the createRoute method from the CollectionConfigurator, because it was thought to be unused, but this method was used when adding a new Route to a RouteCollection. Without the method, the createRoute method of the LocalizedRouteTrait was used.
This means that when setting up a route collection before adding routes, the new routes no longer get the collection settings.

Added a new unittest to prevent this from happening again. For now only added to the PhpFileLoaderTest, I don't know if this is also possible for XML/YML routing config to also add a test in those LoaderTests.

### DIFF
--- a/src/Symfony/Component/Routing/Loader/Configurator/CollectionConfigurator.php
+++ b/src/Symfony/Component/Routing/Loader/Configurator/CollectionConfigurator.php
@@ -114,4 +114,12 @@ class CollectionConfigurator
 
         return $this;
     }
+
+    /**
+     * This method overrides the one from LocalizedRouteTrait.
+     */
+    private function createRoute(string $path): Route
+    {
+        return (clone $this->route)->setPath($path);
+    }
 }
--- a/src/Symfony/Component/Routing/Tests/Fixtures/collection-defaults.php
+++ b/src/Symfony/Component/Routing/Tests/Fixtures/collection-defaults.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Symfony\Component\Routing\Loader\Configurator;
+
+return function (RoutingConfigurator $routes) {
+    $collection = $routes->collection();
+    $collection
+        ->methods(['GET'])
+        ->defaults(['attribute' => true])
+        ->stateless();
+
+    $collection->add('defaultsA', '/defaultsA')
+        ->locale('en')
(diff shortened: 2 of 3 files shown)


bram123:
Test failures are (hopefully) unrelated; or did I do something wrong?

src/Symfony/Component/Routing/Loader/Configurator/CollectionConfigurator.php
```diff
@@ -114,4 +114,12 @@ final public function host(string|array $host): static
 
         return $this;
     }
+
+    /**
+     * Create a new route with the collection's settings.
```
fancyweb: Let's add a comment saying this method is actually used since it overrides the one from LocalizedRouteTrait

fancyweb (approved):

src/Symfony/Component/Routing/Loader/Configurator/CollectionConfigurator.php
bram123: Added a comment with `@see` pointing to the trait::method

src/Symfony/Component/Routing/Loader/Configurator/CollectionConfigurator.php
```diff
@@ -114,4 +114,14 @@ final public function host(string|array $host): static
 
         return $this;
     }
+
+    /**
+     * Create a new route with the collection's settings, used in LocalizedRouteTrait.
```
fancyweb: I think something like "This method overrides the one from LocalizedRouteTrait" is enough to prevent PRs removing it again :sweat_smile:

nicolas-grekas:
Good catch, thanks @bram123.

PR #51706, opened 2023-09-20
[Translation] [Phrase] Refacto ReadConfig and WriteConfig into arrays

welcoMattic:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix https://github.com/symfony/symfony/pull/49231
| License       | MIT

This PR follows up https://github.com/symfony/symfony/pull/49231 which brings a new Translation Provider for [Phrase](https://phrase.com/).

But there was a complexity layer about read and write configuration (means API calls options) in the original code. 

I've moved `ReadConfig` and `WriteConfig` from their own class into `PhraseProvider` properties. Which is easier to manipulate, to initialize and to test.


src/Symfony/Component/Translation/Bridge/Phrase/PhraseProviderFactory.php
```diff
@@ -60,14 +70,46 @@ public function create(Dsn $dsn): ProviderInterface
...
+
+        return filter_var($options['fallback_locale_enabled'] ?? false, \FILTER_VALIDATE_BOOL);
+    }
+
+    private function readConfigFromDsn(Dsn $dsn): array
+    {
+        $options = $dsn->getOptions()['read'] ?? [];
+
+        // enforce empty translations when fallback locale is enabled
+        if (true === $this->isFallbackLocaleEnabled($dsn)) {
```
nicolas-grekas: we don't compare to booleans unless another type can be returned (not the case here)

nicolas-grekas (approved):

nicolas-grekas:
Thank you @welcoMattic.

PR #51709, opened 2023-09-21
[Routing] Do not decode semicolons in query strings

### LINKED ISSUE #51708: [Routing] Symfony falsely decodes the semicolon character in query strings

6.2.8, 6.3.3

### Description

The `UrlGenerator` builds a URL with encoded query string arguments, but decoded a subset of characters again. While this might be fine in most cases, it causes issues with the semicolon character with traefik (https://github.com/traefik/traefik/issues/9164). When the character stays encoded, everything is fine in our case. Also, the statement "reserved chars that have no special meaning for HTTP URIs" (https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Routing/Generator/UrlGenerator.php#L35) is wrong, as semicolons are sub-delimiters as per [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt):

```
sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
            / "*" / "+" / "," / ";" / "="
```

### How to reproduce

That's a bit tricky. You need a recent ddev version which will bring traefik by default.

I've setup up a dummy project to demonstrate the issue easily: https://github.com/andreasfernandez/symfony-url-semicolon. Remember: ddev is a hard requirement to test this.

The project renders the current query string and a link containing a semicolon in the query string: `/?foo=bar;baz`. When clicking the link, the request is processed by traefik and the query string is transformed to `baz=&foo=bar`.

### Possible Solution

The low-hanging fix is to remove[ `'%3B' => ';',`](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Routing/Generator/UrlGenerator.php#L40).

### Additional Context

_No response_

andreaskienast:
| Q             | A
| ------------- | ---
| Branch?       | 5.4, 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #51708
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Routing/Generator/UrlGenerator.php
+++ b/src/Symfony/Component/Routing/Generator/UrlGenerator.php
@@ -36,7 +36,6 @@ class UrlGenerator implements UrlGeneratorInterface, ConfigurableRequirementsInt
         '%40' => '@',
         '%3A' => ':',
         '%21' => '!',
-        '%3B' => ';',
         '%2C' => ',',
         '%2A' => '*',
     ];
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Closing as explained in the linked issue. You might try decorating the generator to have the special behavior on your side.

PR #51710, opened 2023-09-21
[Validator] Fix `File::$extensions`’ PHPDoc

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

[The documentation’s example](https://symfony.com/doc/current/reference/constraints/File.html#extensions) is incompatible with the current PHPDoc because keys can be integers **and** strings in a single array:

image

It would now accept `string[][]` as well, but a false positive seems better to me than a false negative.

### DIFF
--- a/src/Symfony/Component/Validator/Constraints/File.php
+++ b/src/Symfony/Component/Validator/Constraints/File.php
@@ -73,7 +73,7 @@ class File extends Constraint
     protected $maxSize;
 
     /**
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Thank you @MatTheCat.

PR #51711, opened 2023-09-21
Update InteractiveAuthenticatorInterface description wording to match documentation

bobvandevijver:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | -
| License       | MIT
| Doc PR        | -

The wording in code mentions the authenticator **must** implement the interface, while its only functionality is to dispatch an additional event which is not required for default functionality. The current documentation also uses may to indicate it is optional:

> If your login method is interactive, which means that the user actively logged into your application, you may want your authenticator to implement the [InteractiveAuthenticatorInterface](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Security/Http/Authenticator/InteractiveAuthenticatorInterface.php) so that it dispatches an [InteractiveLoginEvent](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Security/Http/Event/InteractiveLoginEvent.php)

This PR updates the wording in the interface to indicate it is indeed optional. I do not see the need to add this to the changelog.

If targeted 6.4 as I do not see this as a bug fix, but I can easily be retargeted when needed.

### DIFF
--- a/src/Symfony/Component/Security/Http/Authenticator/InteractiveAuthenticatorInterface.php
+++ b/src/Symfony/Component/Security/Http/Authenticator/InteractiveAuthenticatorInterface.php
@@ -12,7 +12,7 @@
 namespace Symfony\Component\Security\Http\Authenticator;
 
 /**
- * This is an extension of the authenticator interface that must
+ * This is an extension of the authenticator interface that may
  * be used by interactive authenticators.
  *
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Thank you @bobvandevijver.

PR #18048, opened 2016-03-07
[HttpKernel] Fix mem usage when stripping the prod container

nicolas-grekas:
| Q | A |
| --- | --- |
| Branch | 2.3 |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| Tests pass? | yes |
| Fixed tickets | #18007 |
| License | MIT |
| Doc PR | - |

I propose to just replace doc comments by regular comments, so that the parser removes them and opcache doesn't have to keep them in memory, which is the target.

PR #51712, opened 2023-09-21
Deprecate `Kernel::stripComments()`

alamirault:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #... 
| License       | MIT

This PR replace `method_exists` by `class_exists` in order to check if component is available. It harmonize codebase

~This PR remove checking if method `Kernel::setIgnore` exists.~

~This method was introduced in symfony 2.3 (https://github.com/symfony/symfony/pull/18048), so now it's always available~

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Dumper/PhpDumper.php
+++ b/src/Symfony/Component/DependencyInjection/Dumper/PhpDumper.php
@@ -44,7 +44,6 @@ use Symfony\Component\DependencyInjection\TypedReference;
 use Symfony\Component\DependencyInjection\Variable;
 use Symfony\Component\ErrorHandler\DebugClassLoader;
 use Symfony\Component\ExpressionLanguage\Expression;
-use Symfony\Component\HttpKernel\Kernel;
 
 /**
  * PhpDumper dumps a service container as a PHP class.
@@ -571,7 +570,7 @@ EOF;
         $proxyClasses = [];
         $alreadyGenerated = [];
         $definitions = $this->container->getDefinitions();
-        $strip = '' === $this->docStar && method_exists(Kernel::class, 'stripComments');
+        $strip = '' === $this->docStar;
         $proxyDumper = $this->getProxyDumper();
         ksort($definitions);
         foreach ($definitions as $id => $definition) {
@@ -620,7 +619,7 @@ EOF;
 
             if ($strip) {
                 $proxyCode = "<?php\n".$proxyCode;
-                $proxyCode = substr(Kernel::stripComments($proxyCode), 5);
+                $proxyCode = substr(self::stripComments($proxyCode), 5);
             }
 
             $proxyClass = explode(' ', $this->inlineRequires ? substr($proxyCode, \strlen($code)) : $proxyCode, 3)[1];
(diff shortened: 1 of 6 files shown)


derrabus (changes requested):
The DI component can be used without the HttpKernel component, so this check is still relevant!

alamirault:
Hum, I understand now !
But shouldn't we use `class_exists` instead ? Like is done in codebase ?

derrabus:
We could do that, yes.

stof:
given that this method is not used anywhere in the HttpKernel component and does not depend on anything else in the component (its only dependency is `token_get_all`), maybe we should copy it in the PhpDumper of the DI component as a private method and deprecate the Kernel method.

src/Symfony/Component/DependencyInjection/Dumper/PhpDumper.php
```diff
@@ -2339,4 +2338,65 @@ private function isProxyCandidate(Definition $definition, ?bool &$asGhostObject,
...
         return $this->getProxyDumper()->isProxyCandidate($definition, $asGhostObject, $id) ? $definition : null;
     }
+
+    /**
+     * Removes comments from a PHP source string.
+     *
+     * We don't use the PHP php_strip_whitespace() function
+     * as we want the content to be readable and well-formatted.
+     */
+    private function stripComments(string $source): string
```
alamirault: I don't know how to test because method is now private. I think we should re-use [same dataprovider](https://github.com/symfony/symfony/pull/51712/files#diff-d85a8adbc4a569ab97ca42339d9bd47d41025f2af53cea0c537d5ed55276625bR271).

Can I use ReflectionClass or what is the best practice in this repository ?

derrabus: Testing via reflection should be fine, I think.

alamirault:
I deprecated method in `Kernel` and created a private in `PhpDumper`

src/Symfony/Component/DependencyInjection/Dumper/PhpDumper.php
stof: I suggest keeping that method static as it does not depend on anything. It will make it easier to test it (not need to create a fake container to create a dumper)

src/Symfony/Component/DependencyInjection/Tests/Dumper/PhpDumperTest.php
```diff
@@ -1976,6 +1976,115 @@ public function testCallableAdapterConsumer()
...
+     */
+    public function testStripComments(string $source, string $expected)
+    {
+        $container = new ContainerBuilder();
+        $container->compile();
+        $dumper = new PhpDumper($container);
+
+        $reflection = new \ReflectionClass($dumper);
+        $method = $reflection->getMethod('stripComments');
+        $method->setAccessible(true);
```
stof: `setAccessible` is useless. It is a no-op in PHP 8.1+

alamirault: I didn't know, great !

src/Symfony/Component/DependencyInjection/Dumper/PhpDumper.php
```diff
@@ -620,7 +619,7 @@ private function generateProxyClasses(): array
 
             if ($strip) {
                 $proxyCode = "<?php\n".$proxyCode;
-                $proxyCode = substr(Kernel::stripComments($proxyCode), 5);
+                $proxyCode = substr($this->stripComments($proxyCode), 5);
```
derrabus:
```suggestion
                $proxyCode = substr(self::stripComments($proxyCode), 5);
```

derrabus (approved):

stof (approved):

fabpot (approved):

fabpot:
Thank you @alamirault.

PR #49986, opened 2023-04-10
[Notifier] Add options to Telegram Bridge

Nsbx:
| Q             | A
| ------------- | ---
| Branch?       | 6.3 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Tickets       | 
| License       | MIT
| Doc PR        | symfony/symfony-docs#... 

Discord, Mastodon and other notifier have possibility to send photo.

Telegram can send photo too but not the bridge.

With this small change, this is now possible.

PR #51717, opened 2023-09-22
[Notifier] [Telegram] Extend options for `location`, `document`, `audio`, `video`, `venue`, `photo`, `animation`, `sticker` & `contact`

igrizzli:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT

Extends https://github.com/symfony/symfony/pull/49986 for sending local files as photo

Add possibility to send point on map in notification.


src/Symfony/Component/Notifier/Bridge/Telegram/README.md
```diff
@@ -76,6 +76,34 @@ $chatMessage->options($telegramOptions);
 $chatter->send($chatMessage);
 ```
 
+Adding Local Photo to a Message
```
OskarStark: AFAICS the only change to the paragraph above is the relative file path, so I would rather extend this paragraph instead of copy pasting the whole one and just change a small line. WDYT?

src/Symfony/Component/Notifier/Bridge/Telegram/TelegramTransport.php
```diff
@@ -128,4 +132,12 @@ private function getAction(array $options): string
             default => 'post',
         };
     }
+
+    private function isPhotoPrivate(string $photo): bool
```
OskarStark: The method name is a bit misleading, what about isRemote or isLocal?

OskarStark (approved):
Overall this PR looks good to me 👍🏻

src/Symfony/Component/Notifier/Bridge/Telegram/TelegramOptions.php
```diff
@@ -112,6 +112,16 @@ public function photo(string $url): static
         return $this;
     }
 
+    /**
+     * @return $this
+     */
+    public function uploadPhoto(string $path): static
```
igrizzli, line 118: I have to introduce new option, because passing local path to `photo` will break undocumented feature - sending photo by file_id.

OskarStark:
Do you know if there is an option to send documents?

igrizzli:
> Do you know if there is an option to send documents?

Yes, i can add this feature on this weekend

OskarStark:
Oh that would be great 😍

igrizzli:
@OskarStark sending photo, location and document are exclusive options, it's can be combined in single message between themselves. Should we validate it, to avoid implicit behavior? If we should - Which level is better to use for validation TelegramOptions or TelegramTransport?

OskarStark:
I really like to validate such things, but we can't validate all the things plus we tightly couple the transport to the API, in this case, if things get changed we need to release new code, that's not ideal.

So let's not validate it, document it properly with a note/warning and let the API throw the error message if someone wants to send all of them together.

igrizzli:
@OskarStark I have added document and all other telegram send types, can you look implementation? And if it ok i will finish it with tests and docs.

Also i have question about tests, since we are mocking real telegram api in tests, would it be ok to use single sample file for all types in tests, to avoid adding large binary to repo?

OskarStark:
> Also i have question about tests, since we are mocking real telegram api in tests, would it be ok to use single sample file for all types in tests, to avoid adding large binary to repo?

Yes

Can you please update PR title, header, CHANGELOG etc. and mention also the `document`? Thanks

src/Symfony/Component/Notifier/Bridge/Telegram/TelegramTransport.php
```diff
@@ -115,6 +126,14 @@ private function getPath(array $options): string
             isset($options['message_id']) => 'editMessageText',
             isset($options['callback_query_id']) => 'answerCallbackQuery',
             isset($options['photo']) => 'sendPhoto',
+            isset($options['location']) => 'sendLocation',
+            isset($options['audio']) => 'sendAudio',
+            isset($options['document']) => 'sendDocument',
```
OskarStark, line 147: If I get this right, there is only one Endpoint for document, but we have `uploadDocument($filepath)` and `document($url)`, can you explain why we need both?

However, what I would check inside the methods without `upload` prefix, is, that the file exists. Please also add unit tests for this.

OskarStark:
I saw you added a lot more, like audio, venue etc. very nice.

Please keep the CHANGELOG short and add a list of newly added capabilities. I will update the PR title

src/Symfony/Component/Notifier/Bridge/Telegram/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+* Add support for using local files in `sendPhoto` API method
+* Add support for `sendLocation` API method
```
OskarStark:
```suggestion
* Extend options for `location`, `document`, `audio`, `video`, `venue`, `photo`, `animation`, `sticker` & `contact`
```

src/Symfony/Component/Notifier/Bridge/Telegram/TelegramOptions.php
```diff
@@ -156,4 +166,156 @@ public function answerCallbackQuery(string $callbackQueryId, bool $showAlert = f
...
+            $this->options['emoji'] = $emoji;
+        }
+
+        return $this;
+    }
+
+    /**
+     * @return $this
+     */
+    public function contact(string $phoneNumber, string $firstName, ?string $lastName = null, ?string $vCard = null): static
```
OskarStark: $vCard is content or a path?

src/Symfony/Component/Notifier/Bridge/Telegram/Tests/TelegramTransportTest.php
```diff
@@ -411,4 +411,259 @@ public function testSendPhotoWithOptions()
...
+                            "width": 90,
+                            "height": 51
+                        },
+                        {
+                            "file_id": "ABCDEF",
+                            "file_unique_id" : "ABCDEF2",
+                            "file_size": 19987,
+                            "width": 320,
+                            "height": 180
+                        }
```
OskarStark: Why adding 2 photos?

src/Symfony/Component/Notifier/Bridge/Telegram/TelegramTransport.php
igrizzli: Telegram supports 3 ways for sending files - http url, file_id and uploading via api request 

docs for photo for example - https://github.com/symfony/symfony/pull/51717/files#diff-d5c6f680237182e620c35f9de66750b1d9363cd6cd69032b68286945990fb4caR84

We can't detect is value file_id or file_path for upload. Before this pull request we had method `photo` which worked with http url and file_id, for avoiding BC we have to add uploadPhoto. For all other types i added same 2 methods to keep methods consistent for all types.

We can try to detect is argument `file_path` by checking `file_exists`, and if not pass value as `file_id`, but in this case if user will try to pass `file_path` to not existing file (because some application issues), he will not catch correct api error.

src/Symfony/Component/Notifier/Bridge/Telegram/Tests/TelegramTransportTest.php
igrizzli: This is wrong, will be fixed.
Copy pasted from previous test, will fix it too.

src/Symfony/Component/Notifier/Bridge/Telegram/TelegramOptions.php
igrizzli: Content - https://en.wikipedia.org/wiki/VCard

igrizzli:
PR in work now, i have not added docs and tests yet, i have some doubt about implementation, after we will fix or approve them i will finish with docs and tests

src/Symfony/Component/Notifier/Bridge/Telegram/TelegramTransport.php
igrizzli: Instead of using extra methods for uploading (`uploadPhoto`, `uploadDocument` and etc.) we can use flag `upload`:

```
  (new TelegramOptions())
    ->document('files/test.xlsx')
    ->upload();
```

In this case if `upload` flag exists then will handle argument as `filePath` if not as `fileId` or `httpUrl`

OskarStark: I understand, from my POV we can go with the 2 methods, it's a clear API. 👍

OskarStark:
Let's wait for @stof or @fabpot to have a look 👍

src/Symfony/Component/Notifier/Bridge/Telegram/TelegramOptions.php
```diff
@@ -156,4 +166,156 @@ public function answerCallbackQuery(string $callbackQueryId, bool $showAlert = f
...
+    }
+
+    /**
+     * @return $this
+     */
+    public function uploadSticker(string $path, string $emoji = null): static
+    {
+        $this->options['upload']['sticker'] = $path;
+        if ($emoji !== null) {
+            $this->options['emoji'] = $emoji;
```
stof: shouldn't we also assign `null` here, to properly account for a case where the caller wants to change a previous config ?

stof:
The fact that many of those methods are incompatible with each other as they trigger the usage of different endpoints (with an order that is not even related to which one was configured first or last, but an ordered list of which ones wins based on the order of cases in your matching) looks like a confusing API. But this was already the case before.

src/Symfony/Component/Notifier/Bridge/Telegram/TelegramOptions.php
stof: OK, based on the fact that calling several of those methods will do wonky stuff anyway, this might be fine...

OskarStark:
> But this was already the case before.

Exactly, shall we throw exceptions if one of those apis/endpoints was already used within the TelegramOptions class? I voted against and let the API bubble up the error

igrizzli:
> > But this was already the case before.
> 
> Exactly, shall we throw exceptions if one of those apis/endpoints was already used within the TelegramOptions class? I voted against and let the API bubble up the error

I have tested it - API will not throw error, they just ignore options which not match endpoint API. So if we will try to send photo and document we will send only photo, because `sendPhoto` endpoint first in match.
```
$this->chatter->send(new ChatMessage('photo', (new TelegramOptions())->uploadPhoto('fixtures/photo.png')->uploadDocument('fixtures/document.xlsx')));
```

May be it will be better if will throw exception if multiple endpoints match options?

OskarStark:
Lets keep it simple for now

src/Symfony/Component/Notifier/Bridge/Telegram/CHANGELOG.md
nicolas-grekas: please add a leading space also

nicolas-grekas:
Don't miss applying this patch from fabbot

igrizzli:
Tests and docs done. 
I have also added validation if options targeted to different endpoints, because adding it in a future will be BC.

I have used in tests single file for all endpoints because in real we just need any binary file for testing, and adding multiple fixture files, especially videos, to repo can significantly increase repo size and provide legal issue. But if you think that is critical i will replace it with matched type files.

OskarStark (approved):

src/Symfony/Component/Notifier/Bridge/Telegram/TelegramOptions.php
```diff
@@ -156,4 +166,164 @@ public function answerCallbackQuery(string $callbackQueryId, bool $showAlert = f
...
+    public function sticker(string $url, string $emoji = null): static
+    {
+        $this->options['sticker'] = $url;
+        if (null !== $emoji) {
+            $this->options['emoji'] = $emoji;
+        } else {
+            if (isset($this->options['emoji'])) {
+                unset($this->options['emoji']);
+            }
+        }
```
nicolas-grekas: what about removing all those if/else from the class (aka just do `$this->options['emoji'] = $emoji;) and add an array_filter in the class that turns the options into a URL?

src/Symfony/Component/Notifier/Bridge/Telegram/CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+ * Add support for `sendLocation`, `sendAudio`, `sendDocument`, `sendVideo`, `sendAnimation`, `sendVenue`, `sendContact` and `sendSticker` API methods
+ * Add support for sending local files
+
+
```
nicolas-grekas: extra line :)

OskarStark:
Good to go @nicolas-grekas ?

fabpot (approved):

fabpot:
Thank you @igrizzli.

PR #37576, opened 2020-07-14
[WebProfilerBundle] modified url generation to use absolute urls

smatyas:
| Q             | A
| ------------- | ---
| Branch?       | 4.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

## TODO
- [x] gather feedback for my changes

## Use case

The web debug toolbar ajax calls are broken in the case when you have to work with a page template that uses HTML `<base>` Tag for whatever reason.

In this case, the ajax calls of the debug toolbar will target that base url, as these urls are rendered as relative urls in the injected toolbar HTML content.

A simple solution for this issue is to generate absolute urls for these calls - this is my first approach in this PR.

A more complex solution would be a configurable behavior with the current method as the default.

I believe though that the simple solution is enough here, as it will result in the same requests for developers using templates without any `<base>` tag, and it will also make the dev toolbar usable for my case (using a `<base>` tag).

(Please let me know if I should make this PR against another branch.)

PR #51722, opened 2023-09-22
[WebProfilerBundle] make webprofiler links absolute path

shyim:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

We use a nodejs proxy to hot-reload js/css changes. The proxy url is different to the symfony url, so we replace some urls. It would make our life easier when all generated links are absolute paths (without the host).

### DIFF
--- a/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/base_js.html.twig
+++ b/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/base_js.html.twig
@@ -537,7 +537,7 @@ if (typeof Sfjs === 'undefined' || typeof Sfjs.loadToolbar === 'undefined') {
 
                 this.load(
                     'sfwdt' + token,
-                    '{{ url("_wdt", { "token": "xxxxxx" })|escape('js') }}'.replace(/xxxxxx/, newToken),
+                    '{{ path("_wdt", { "token": "xxxxxx" })|escape('js') }}'.replace(/xxxxxx/, newToken),
                     function(xhr, el) {
                         /* Do nothing in the edge case where the toolbar has already been replaced with a new one */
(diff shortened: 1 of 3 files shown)


src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/base_js.html.twig
```diff
@@ -537,7 +537,7 @@ if (typeof Sfjs === 'undefined' || typeof Sfjs.loadToolbar === 'undefined') {
 
                 this.load(
                     'sfwdt' + token,
-                    '{{ url("_wdt", { "token": "xxxxxx" })|escape('js') }}'.replace(/xxxxxx/, newToken),
+                    '{{ path("_wdt", { "token": "xxxxxx" })|escape('js') }}'.replace(/xxxxxx/, newToken),
```
nicolas-grekas, line 540: isn't `url()` for absolute URLs and `path()` for relative ones? the PR title says the opposite

ro0NL:
ref #37576

stof:
This breaks cases where your configure ajax requests to replace the toolbar for ajax responses and those responses from your API come from a different domain than the main document, which is exactly the use case for which we switched to absolute URLs.

For your use case of proxying, the right solution IMO is to configure your proxy as a trusted proxy to make Symfony aware of the public-facing domain name to use.

stof:
This PR gets a -1 vote from me.

shyim:
Ahh yea forgot the `<base>`. 

trusted_proxies are annoying for development, especially when you are a framework and not just a single website.

PR #51724, opened 2023-09-22
[Security] Add CSRF migration strategy

### LINKED ISSUE #47276: CSRF Token is not kept when session is renewed with REMEMBERME

6.1.X

### Description

CSRF tokens by default are stored and bound to the session and have no expiration per se.  However when using the REMEMBER-ME feature,  when the session is renewed, the csrf token is not kept,  causing invalid form

### How to reproduce

1. create a symfony project with a login form and remember me activated

```
security:
    enable_authenticator_manager: true
    # https://symfony.com/doc/current/security.html#where-do-users-come-from-user-providers
    providers:
        endusers:
            id: 'App\Provider\EndUserProvider'
    firewalls:
        dev:
            security: false
        endusers:
            host: app.*
            pattern: ^/
            provider: endusers
            lazy: true
            remember_me:
                secret: '%kernel.secret'
                # 2 months in seconds, the goal is to have the user logged-in
                # enough time to start to appreciate the service, so that he will
                # care about doing 'forgot password'
                lifetime: 5356800
                secure: true
                path: /
                always_remember_me: true
                token_provider: 'Symfony\Bridge\Doctrine\Security\RememberMe\DoctrineTokenProvider'
            form_login:
                login_path: end_users_login
                check_path: end_users_login
                enable_csrf: true
                default_target_path: end_users_login
```

2. create an other form that is behind the `endusers` firewall ,  nothing special , let symfony manage it completly

### now as a user 

1.  login
2. open the page with the form
3. do noting during 24 minutes and 1 seconds , or at least session.gc-maxlifetime + 1 seconds ( https://www.php.net/manual/en/session.configuration.php#ini.session.gc-maxlifetime )
4. submit the form
=>  you get a `Invalid CSRF token`

## Expected the behaviour 

As the session is recreated by remember-me ,  the csrf should be kept

### Possible Solution

When creating the remember-me cookie,  the csrf token should be stored in it,  so that when recreating the session, the csrf token can be kept 

### Additional Context

_No response_

Spomky:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Tickets       | Fix #47276
| License       | MIT
| Doc PR        | **TODO**

This PR is a proposal for fixing #47276 where the CSRF tokens are cleared when the session is migrated.
It adds a new option to the `security` section where the developer can choose to clear all CSRF tokens (default and current behavior) or keep them.

```yaml
security:
    csrf_migration_strategy: 'keep'
```

> **Note: As I have not yet created a sandbox to reproduce the issue, please send me your feedback on the proposal.**


nicolas-grekas:
Isn't this opening a CSRF attack vector? This is especially dangerous with remember-me, which can be triggered by an evil site quite easily.

Isn't the real solution to NOT use session tokens for CSRF protection in this case? Eg use double-submit or checking the Origin header?

Spomky:
Hi @nicolas-grekas,

Many thanks for your comment.
I think you are right. I tested with a very low session lifetime and, with the `keep` option, the CSRF still exists after a long period of time. A malicious party could take advantages of it.
As you suggest, in the case of a remember-me cookie, no form data is submit and the CSRF as no added value here.
I close this PR.

PR #51744, opened 2023-09-25
[Security] Add a normalization step for the user-identifier in firewalls

### LINKED ISSUE #49269: Add a normalization step for the user-identifier in firewalls
When user identifiers are submitted, what about normalizing them before hitting the rate-limiter and then the provider?
That'd allow easier comparisons, and maybe some more hardening against spoofing/confusables.

Related to #49268

A simple normalizer could eg lowercase + turn them in NFKC.

Spomky:
| Q             | A
| ------------- | ---
| Branch?       | 7.3
| Bug fix?      | yes
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #49269
| License       | MIT
| Doc PR        | **TODO**

With this PR, the user identifier is lowered and normalized to make sure they are case-insensitive.
When enable, please note that the user provider or the user repository should be adapt:

* user identifier stored in the future shall be lowered and normalized as well
* the queries for existing user should be adapt

This is a very first attempt to resolve issue #49269. Any comments and reactions are welcome.

-----

See also https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html#user-ids

### DIFF
--- a/src/Symfony/Component/Security/Http/Authenticator/Passport/Badge/UserBadge.php
+++ b/src/Symfony/Component/Security/Http/Authenticator/Passport/Badge/UserBadge.php
@@ -33,6 +33,7 @@ class UserBadge implements BadgeInterface
     /** @var callable|null */
     private $userLoader;
     private UserInterface $user;
+    private ?\Closure $identifierNormalizer = null;
 
     /**
      * Initializes the user badge.
@@ -51,6 +52,7 @@ class UserBadge implements BadgeInterface
         private string $userIdentifier,
         ?callable $userLoader = null,
         private ?array $attributes = null,
+        ?\Closure $identifierNormalizer = null,
     ) {
         if ('' === $userIdentifier) {
             trigger_deprecation('symfony/security-http', '7.2', 'Using an empty string as user identifier is deprecated and will throw an exception in Symfony 8.0.');
@@ -60,12 +62,20 @@ class UserBadge implements BadgeInterface
         if (\strlen($userIdentifier) > self::MAX_USERNAME_LENGTH) {
             throw new BadCredentialsException('Username too long.');
         }
+        if ($identifierNormalizer) {
+            $this->identifierNormalizer = static fn () => $identifierNormalizer($userIdentifier);
+        }
 
         $this->userLoader = $userLoader;
     }
 
     public function getUserIdentifier(): string
     {
+        if (isset($this->identifierNormalizer)) {
+            $this->userIdentifier = ($this->identifierNormalizer)();
+            $this->identifierNormalizer = null;
+        }
+
         return $this->userIdentifier;
     }
 
@@ -88,15 +98,15 @@ class UserBadge implements BadgeInterface
         }
 
         if (null === $this->getAttributes()) {
-            $user = ($this->userLoader)($this->userIdentifier);
+            $user = ($this->userLoader)($this->getUserIdentifier());
         } else {
-            $user = ($this->userLoader)($this->userIdentifier, $this->getAttributes());
+            $user = ($this->userLoader)($this->getUserIdentifier(), $this->getAttributes());
         }
 
         // No user has been found via the $this->userLoader callback
         if (null === $user) {
             $exception = new UserNotFoundException();
-            $exception->setUserIdentifier($this->userIdentifier);
+            $exception->setUserIdentifier($this->getUserIdentifier());
 
             throw $exception;
         }
--- a/src/Symfony/Component/Security/Http/CHANGELOG.md
+++ b/src/Symfony/Component/Security/Http/CHANGELOG.md
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Add encryption support to `OidcTokenHandler` (JWE)
  * Replace `$hideAccountStatusExceptions` argument with `$exposeSecurityErrors` in `AuthenticatorManager` constructor
+ * Add argument `$identifierNormalizer` to `UserBadge::__construct()` to allow normalizing the identifier
 
 7.2
 ---
--- a/src/Symfony/Component/Security/Http/Tests/Authenticator/Passport/Badge/UserBadgeTest.php
+++ b/src/Symfony/Component/Security/Http/Tests/Authenticator/Passport/Badge/UserBadgeTest.php
@@ -15,6 +15,10 @@ use PHPUnit\Framework\TestCase;
 use Symfony\Bridge\PhpUnit\ExpectUserDeprecationMessageTrait;
 use Symfony\Component\Security\Core\Exception\UserNotFoundException;
 use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
+use Symfony\Component\String\Slugger\AsciiSlugger;
+use Symfony\Component\String\UnicodeString;
+
+use function Symfony\Component\String\u;
 
 class UserBadgeTest extends TestCase
 {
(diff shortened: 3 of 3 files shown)


src/Symfony/Component/Security/Http/Authenticator/Passport/Badge/UserBadge.php
```diff
@@ -49,19 +53,26 @@ class UserBadge implements BadgeInterface
      * is thrown). If this is not set, the default user provider will be used with
      * $userIdentifier as username.
      */
-    public function __construct(string $userIdentifier, callable $userLoader = null, array $attributes = null)
+    public function __construct(string $userIdentifier, callable $userLoader = null, array $attributes = null, bool $normalizeIdentifier = null)
```
nicolas-grekas: I'd suggest accepting a closure instead, which would be used as the service doing the normalization, allowing to use whatever logic for that + pass it to all places that require the same processing.

Spomky: Indeed. I will update this PR and propose a callable instead.

Spomky: 👌🏼 The PR is now up to date and accepts a callable.

Spomky:
Rebased to target Symfony 7.1

src/Symfony/Component/Security/Http/Authenticator/Passport/Badge/UserBadge.php
```diff
@@ -36,6 +36,9 @@ class UserBadge implements BadgeInterface
     private UserInterface $user;
     private ?array $attributes;
 
+    /** @var callable */
+    private $identifierNormalizer;
```
nicolas-grekas:
```suggestion
    private ?\Closure $identifierNormalizer = null;
```

src/Symfony/Component/Security/Http/Authenticator/Passport/Badge/UserBadge.php
```diff
@@ -49,20 +52,20 @@ class UserBadge implements BadgeInterface
...
      */
-    public function __construct(string $userIdentifier, callable $userLoader = null, array $attributes = null)
+    public function __construct(string $userIdentifier, callable $userLoader = null, array $attributes = null, callable $identifierNormalizer = null)
     {
         if (\strlen($userIdentifier) > self::MAX_USERNAME_LENGTH) {
             throw new BadCredentialsException('Username too long.');
         }
-
+        $this->identifierNormalizer = $identifierNormalizer ?? fn ($userIdentifier) => $userIdentifier;
         $this->userIdentifier = $userIdentifier;
```
nicolas-grekas:
```suggestion
        if ($identifierNormalizer) {
            $this->identifierNormalizer = static fn () => $identifierNormalizer($userIdentifier);
        } else {
            $this->userIdentifier = $userIdentifier;
        }
        ;
```

src/Symfony/Component/Security/Http/Authenticator/Passport/Badge/UserBadge.php
```diff
@@ -49,20 +52,20 @@ class UserBadge implements BadgeInterface
...
         $this->userLoader = $userLoader;
         $this->attributes = $attributes;
     }
 
     public function getUserIdentifier(): string
     {
-        return $this->userIdentifier;
+        return ($this->identifierNormalizer)($this->userIdentifier);
```
nicolas-grekas: trying to provide memoization:
```suggestion
        $userIdentifier = $this->userIdentifier ??= ($this->identifierNormalizer)();
        $this->identifierNormalizer = null;
        
        return $userIdentifier;
```

alternative we allow userIdentifier to be a closure and when that's the case, we call it to resolve it to a string

nicolas-grekas (commented):
needs a changelog entry :)

src/Symfony/Component/Security/Http/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Add argument `$identifierNormalizer` to `UseBadge::__construct()` to allow normalizing the username
```
stloyd:
```suggestion
 * Add argument `$identifierNormalizer` to `UserBadge::__construct()` to allow normalizing the identifier
```

Spomky: Many thanks!

src/Symfony/Component/Security/Http/Authenticator/Passport/Badge/UserBadge.php
Spomky: Many thanks.
PR up to date.

src/Symfony/Component/Security/Http/Authenticator/Passport/Badge/UserBadge.php
```diff
@@ -49,19 +51,29 @@ class UserBadge implements BadgeInterface
...
         if (\strlen($userIdentifier) > self::MAX_USERNAME_LENGTH) {
             throw new BadCredentialsException('Username too long.');
         }
-
-        $this->userIdentifier = $userIdentifier;
+        if ($identifierNormalizer) {
+            $this->identifierNormalizer = static fn () => $identifierNormalizer($userIdentifier);
+        } else {
+            $this->userIdentifier = $userIdentifier;
+        }
```
nicolas-grekas, line 67:
```suggestion
        }

```

src/Symfony/Component/Security/Http/Authenticator/Passport/Badge/UserBadge.php
```diff
@@ -49,19 +51,29 @@ class UserBadge implements BadgeInterface
...
         $this->userLoader = $userLoader;
         $this->attributes = $attributes;
     }
 
     public function getUserIdentifier(): string
     {
+        if (isset($this->userIdentifier)) {
+            return $this->userIdentifier;
+
+        }
```
nicolas-grekas:
```suggestion
        }

```

src/Symfony/Component/Security/Http/Authenticator/Passport/Badge/UserBadge.php
```diff
@@ -36,6 +36,8 @@ class UserBadge implements BadgeInterface
     private UserInterface $user;
     private ?array $attributes;
 
+    private ?\Closure $identifierNormalizer;
```
nicolas-grekas:
```suggestion
    private ?\Closure $identifierNormalizer = null;
```

Spomky:
Comments addressed, PR rebased and squashed.
Ready for the next step

src/Symfony/Component/Security/Http/Tests/Authenticator/Passport/Badge/UserBadgeTest.php
```diff
@@ -23,4 +27,29 @@ public function testUserNotFound()
...
+     * @dataProvider provideUserIdentifierNormalizationData
+     */
+    public function testUserIdentifierNormalization(string $identifier, string $expectedNormalizedIdentifier, callable $normalizer)
+    {
+        $badge = new UserBadge($identifier, fn () => null, identifierNormalizer: $normalizer);
+
+        static::assertSame($expectedNormalizedIdentifier, $badge->getUserIdentifier());
+    }
+
+    public function provideUserIdentifierNormalizationData(): iterable
```
nicolas-grekas:
```suggestion
    public static function provideUserIdentifierNormalizationData(): iterable
```

nicolas-grekas:
>  When enable, please note that the user provider or the user repository should be adapt:

do we have any advice on the topic? Or any things we can anticipate that could be part of the same PR?

Spomky:
> do we have any advice on the topic? Or any things we can anticipate that could be part of the same PR?

I checked how the built-in user providers could handle this, but I'm not sure if any of them can be changed as well.
How the user ID is stored or retrieved is highly dependent on the developer's implementation.
To me, the most important now seems to propagate this into the documentation.

nicolas-grekas:
What's the next step here? How do we make it easy to opt-in for the OWASP-recommended normalization?

Spomky:
> What's the next step here? How do we make it easy to opt-in for the OWASP-recommended normalization?

Hello @nicolas-grekas,

This PR is now up to date, rebased against the 7.3 branch, and squashed. It’s ready for review.

The next step is to finalize the documentation. It should demonstrate how to inject a normalizer to ensure that different variations (e.g., `JOHN.DOE`, `john.doe`, `John.DOe`) are treated as equivalent.

BR.

nicolas-grekas:
Can you give us a hint about how this can be used? : )

Spomky:
This feature will most likely be used by the firewall Authenticator service that creates the UserBadge, but it could also be used by the `TokenHandler::getUserBadge` when handling access tokens.

Below is a simple example of how to implement a `NormalizedUserBadge`.
In this example, it ensures that any variations in the username—such as uppercase, lowercase, or accented characters—are normalized to a consistent format.

```php
namespace App\Security;

use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
use Symfony\Component\String\UnicodeString;
use function Symfony\Component\String\u;

final class NormalizedUserBadge extends UserBadge
{
    public function __construct(string $identifier)
    {
        $callback = static fn (string $identifier) => u($identifier)->normalize(UnicodeString::NFKC)->ascii()->lower()->toString();
        
        parent::__construct($identifier, identifierNormalizer: $callback);
    }
}
```

From the Authenticator:

```php
<?php

namespace App\Security;

final class PasswordAuthenticator extends AbstractLoginFormAuthenticator
{
    // Simplified for brievety
    public function authenticate(Request $request): Passport
    {
        $username = (string) $request->request->get('username', '');
        $password = (string) $request->request->get('password', '');

        $request->getSession()
            ->set(SecurityRequestAttributes::LAST_USERNAME, $username);

        return new Passport(
            new NormalizedUserBadge($username),
            new PasswordCredentials($password),
            [
                //All other useful badges
            ]
        );
    }
```

When a user attempts to log in with “John.Doe” or “JOhn.doe,” both will be treated as “john.doe.”
Keep in mind that any existing usernames in the database should also be normalized accordingly.

Another potential use case involves scenarios where users can log in with various forms of the same identifier: `john.doe@acme.com`, `acme.com\jdoe`, `https://amce.com/+jdoe` or `acct:jdoe@acme.com`.
By using this feature, you can unify all these variants into a single, standardized format (e.g., “domain\nickname”) without having to create custom queries on the repository side.
In this case, the callback could be as follows (not tested):

```php
        $callback = static function (string $identifier) {
            $normalized = u($identifier)->normalize(UnicodeString::NFKC)->ascii()->lower()->toString();

            $username = '';
            $domain = '';

            match (true) {
                str_contains($normalized, '@') => [
                    $localPart, $domain = explode('@', $normalized, 2),
                    $username = str_contains($localPart, '.')
                        ? substr(explode('.', $localPart, 2)[0], 0, 1) . explode('.', $localPart, 2)[1]
                        : $localPart,
                ],
                str_contains($normalized, '\\') => [
                    $domain, $username = explode('\\', $normalized, 2)
                ],
                str_contains($normalized, '://') => [
                    $parts = parse_url($normalized),
                    isset($parts['host'], $parts['path']) && [
                        $domain = $parts['host'],
                        $username = ltrim($parts['path'], '/')
                    ]
                ],
                default => $username = $normalized
            };

            $username = preg_replace('/[^a-z0-9]/', '', $username);

            return $domain . '\\' . $username;
        };
```

welcoMattic (approved):
Thanks for this @Spomky! 

I think the documentation of this feature must provide real life example to help users to use it properly (I think mostly about username policy of Google where `foo.bar@gmail.com`, `foobar@gmail.com`, and `f.o.o.b.a.r@gmail.com` refers to the same user).

For the code, it's a 👍 for me

fabpot (approved):

fabpot:
Thank you @Spomky.

fabpot:
As a next step, I'm wondering if we should somehow define some reusable normalizers that developers can use without reinventing the wheel (like the Google strategy mentioned above).

PR #51746, opened 2023-09-25
[Notifier] Add `debugData` property to `SentMessage` class

### LINKED ISSUE #49793: [Notifier] SentMessage add debug property
### Description

Some Api like the AllMySms return informations like total of sms sent or the cost. [Api](https://doc.allmysms.com/api/fr/#api-SMS-sendsimple)

It should be great if we can get those informations in the Notifier SentMessage class.

In the SentMessage of Mailer there is a debug property. I suggest to add this property in the SentMessage of the Notifier componant and to update bridge transport like AllMySmsTransport to update this property.

Maybe debug property is not the best property.

Thanks.

### Example

```php
namespace Symfony\Component\Notifier\Message;

class SentMessage
{
    private MessageInterface $original;
    private string $transport;
    private ?string $messageId = null;
    ...
    private ?array $debug = null;
    ...
    public function getDebug(): array
    {
        return $this->debug;
    }
    public function setDebug(array $debug): void
    {
        $this->debug = $debug;
    }
    ...
```

```php
namespace Symfony\Component\Notifier\Bridge\AllMySms;

final class AllMySmsTransport extends AbstractTransport
{
    ...
    protected function doSend(MessageInterface $message): SentMessage
    {
    ...
    $success = $response->toArray(false);
    ...
    $sentMessage = new SentMessage($message, (string) $this);
    $sentMessage->setMessageId($success['smsId']);
    
    $sentMessage->setDebug($success);
    ...
    }
}
```

ahmedghanem00:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #49793
| License       | MIT
| Doc PR        |

As stated in the title, adding such a property would allow transports to include additional information in the returned `SentMessage` object (e.g. http response)

### DIFF
--- a/src/Symfony/Component/Notifier/Message/SentMessage.php
+++ b/src/Symfony/Component/Notifier/Message/SentMessage.php
@@ -19,6 +19,7 @@ class SentMessage
     private MessageInterface $original;
     private string $transport;
     private ?string $messageId = null;
+    private ?array $debugData = null;
 
     public function __construct(MessageInterface $original, string $transport)
     {
@@ -44,5 +45,15 @@ class SentMessage
     public function getMessageId(): ?string
     {
         return $this->messageId;
+    }
+
+    public function getDebugData(): ?array
+    {
+        return $this->debugData;
+    }
+
+    public function setDebugData(array $data): void
+    {
+        $this->debugData = $data;
     }
 }
--- a/src/Symfony/Component/Notifier/Bridge/AllMySms/AllMySmsTransport.php
+++ b/src/Symfony/Component/Notifier/Bridge/AllMySms/AllMySmsTransport.php
@@ -88,6 +88,7 @@ final class AllMySmsTransport extends AbstractTransport
 
         $sentMessage = new SentMessage($message, (string) $this);
         $sentMessage->setMessageId($success['smsId']);
+        $sentMessage->setDebugData(['response' => $response]);
 
         return $sentMessage;
     }
--- a/src/Symfony/Component/Notifier/Bridge/AllMySms/CHANGELOG.md
+++ b/src/Symfony/Component/Notifier/Bridge/AllMySms/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+* Include the http response in the returned `SentMessage` object
+
 6.3
 ---
 
--- a/src/Symfony/Component/Notifier/CHANGELOG.md
+++ b/src/Symfony/Component/Notifier/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+* Add `debugData` property to `SentMessage` class
+
 6.3
 ---
 


OskarStark:
Not sure about this, I mean it makes sense to have more debug data available, but not sure about the method name and signature.

WDYT @fabpot ?

smnandre:
The response is already logged in the HttpClient i believe, no ? 

I see how this could be used, but maybe add a toggle/config to enable it only in debug mode ? 

Maybe i'm wrong, but i'm feeling a bit "afraid" of the added "weight" (messenger / serialization  ..)

nicolas-grekas:
Looking at the linked issue, this is not about debugging but rather about providing extra info.
Maybe a generic `getInfo()` could work - and we'd ensure this contains the response header (and the json-decoded body)?

src/Symfony/Component/Notifier/Message/SentMessage.php
```diff
@@ -45,4 +46,14 @@ public function getMessageId(): ?string
     {
         return $this->messageId;
     }
+
+    public function getDebugData(): ?array
+    {
+        return $this->debugData;
+    }
+
+    public function setDebugData(array $data): void
```
nicolas-grekas, line 55: this should instead be a new constructor argument

ahmedghanem00: I chose the setter method initially because some transporters may work differently and not have http-response or extra-data to include?

So, it would make sense to make it an optional same as `setMessageId()`?

Not sure but just clarifying

ahmedghanem00:
> Looking at the linked issue, this is not about debugging but rather about providing extra info. Maybe a generic `getInfo()` could work - and we'd ensure this contains the response header (and the json-decoded body)?

@nicolas-grekas Nice, We can then ensure this via adding the whole response object?

fabpot:
Having a free-form array where each transport can add anything looks like a bad idea.

OskarStark:
@nicolas-grekas what about something like this:

```diff
diff --git a/src/Symfony/Component/Notifier/Message/SentMessage.php b/src/Symfony/Component/Notifier/Message/SentMessage.php
index 6febf9ebc9..bd662f68aa 100644
--- a/src/Symfony/Component/Notifier/Message/SentMessage.php
+++ b/src/Symfony/Component/Notifier/Message/SentMessage.php
@@ -19,6 +19,7 @@ class SentMessage
     private MessageInterface $original;
     private string $transport;
     private ?string $messageId = null;
+    private array $info = [];

     public function __construct(MessageInterface $original, string $transport)
     {
@@ -45,4 +46,23 @@ class SentMessage
     {
         return $this->messageId;
     }
+
+    public function setInfo(int $responseStatusCode, array $responseHeaders = [], array $body = []): void
+    {
+        $this->info = [
+            'response' => [
+                'status_code' => $responseStatusCode,
+                'headers' => $responseHeaders,
+                'body' => $body,
+            ]
+        ];
+    }
+
+    /**
+     * @return array{response: array{status_code: int, headers: array, body: array}
+     */
+    public function getInfo(): array
+    {
+        return $this->info;
+    }
 }
```

smnandre:
> [...]

With some `#[\SensitiveParameter]` ? I'm not still sure when to add this attribute or not

nicolas-grekas:
> Having a free-form array where each transport can add anything looks like a bad idea.

In HttpClient we documented the standard info keys, and decorators can add their own things there alsp (e.g. the number of retries for the RetryableHttpClient). This works well. What's the alternative?

chaours:
How can we move forward on that PR ? I think the @OskarStark solution is better than original. We don't want to put debug info but more info from return response if provided. Do you want me to make another PR with your solution directly ? Or could we update this one ? We've been waiting for this improvement for a long time now to pick the fees natively. Is the header is really necessary inside info (too much as @smnandre said) ?

fabpot:
Closing as this PR won't be fixed as is. If someone wants to continue working on this, please open a new PR, maybe based on @OskarStark idea.

Issue #36842, opened 2020-05-16
[Mailer] sendmail transport not working on Windows


Seb33300:
**Symfony version(s) affected**: 5.0.8

**Description**  
I am upgrading my application to Symfony 5 and moving from SwiftMailer to Symfony Mailer.

I would like to send emails using sendmail:
```
MAILER_DSN=sendmail+smtp://default
```

This is working well on Linux but not on Windows.
I have properly configured my php.ini by setting the `sendmail_path` option but it looks like Symfony does not take it in account.

When sending an email, this error is returned:
> Process could not be started: Le chemin d'accès spécifié est introuvable.
> in C:\dev\vendor\symfony\mailer\Transport\Smtp\Stream\ProcessStream.php (line 44) 

While trying to debug, I found that `$this->command` contains `/usr/sbin/sendmail -bs` which looks to be the default linux path for sendmail.

fabpot:
Were you using sendmail on Windows with Swiftmailer as well? If yes, how did you configure it?

Seb33300:
I think the way I presented this was confusing but it was not working with SwiftMailer too.

nescim:
### Found the problem:

I had a look at `SendmailTransportFactory`, and found these three lines (version 4.4.10, lines 23-25):
```php
        if ('sendmail+smtp' === $dsn->getScheme() || 'sendmail' === $dsn->getScheme()) {
            return new SendmailTransport(null, $this->dispatcher, $this->logger);
        }
```
The issue seems to be that the factory is disregarding the path given in the DSN (never calls `getHost()`), and just passes `null` as the command to `SendmailTransport`'s constructor.
On its end, `SendmailTransport` defines the command property with a default value:
```php
    private $command = '/usr/sbin/sendmail -bs';
```
and does not overwrite it if the passed parameter is `null`.

`php.ini` never enters the picture.

### How to fix it
A one-line fix may be to put `$dns->getHost()` in place of null in the factory? I'll try that later but I'm afraid there would be some additional work needed (do spaces need to be escaped?).
A nice feature could be to use `php.ini` configuration if the host (command) is "default".

Personally, I doubt that having the default command hardcoded in `SendmailTransport` is a good idea but it can be handy and altering that would likely be a BC break.

fabpot:
This has been fixed now in 5.2 where we have a new NativeTransport that reads configuration from php.ini. See #36131

PR #36131, opened 2020-03-18
[Mailer] Add a transport that uses php.ini settings for configuration

l-vo:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | symfony/symfony-docs#14083

This PR aims to allow to use the mailer when sendmail is not on the `/usr/sbin` directory or when the `-bs` option is not supported.

PR #51749, opened 2023-09-26
[Mailer] Support getting sendmail path from php.ini

jtojnar:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | TODO

Not all environments have sendmail installed in `/usr/sbin`.
Let’s ask for the path using `ini_get`.
See also https://github.com/symfony/swiftmailer-bundle/pull/302

### DIFF
--- a/src/Symfony/Component/Mailer/Tests/Transport/SendmailTransportTest.php
+++ b/src/Symfony/Component/Mailer/Tests/Transport/SendmailTransportTest.php
@@ -38,13 +38,24 @@ class SendmailTransportTest extends TestCase
 
     public function testToString()
     {
+        $previousSendmailPath = \ini_set('sendmail_path', '');
         $t = new SendmailTransport();
         $this->assertEquals('smtp://sendmail', (string) $t);
     }
 
     public function testToStringNonSmtp()
     {
+        $previousSendmailPath = \ini_set('sendmail_path', '');
         $t = new SendmailTransport('/usr/sbin/sendmail -t -i');
+        \ini_set('sendmail_path', $previousSendmailPath);
+        $this->assertEquals('sendmail://default', (string) $t);
+    }
+
+    public function testToStringPhpConfig()
+    {
+        $previousSendmailPath = \ini_set('sendmail_path', '/usr/sbin/sendmail -t -i');
+        $t = new SendmailTransport();
+        \ini_set('sendmail_path', $previousSendmailPath);
         $this->assertEquals('sendmail://default', (string) $t);
     }
 
--- a/src/Symfony/Component/Mailer/Tests/Transport/SendmailTransportTest.php
+++ b/src/Symfony/Component/Mailer/Tests/Transport/SendmailTransportTest.php
@@ -42,6 +42,12 @@ class SendmailTransportTest extends TestCase
         $this->assertEquals('smtp://sendmail', (string) $t);
     }
 
+    public function testToStringNonSmtp()
+    {
+        $t = new SendmailTransport('/usr/sbin/sendmail -t -i');
+        $this->assertEquals('sendmail://default', (string) $t);
+    }
+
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/Mailer/Tests/Transport/SendmailTransportTest.php
```diff
@@ -38,10 +38,27 @@ protected function tearDown(): void
 
     public function testToString()
     {
+        $previousSendmailPath = \ini_set('sendmail_path', '');
```
derrabus, line 41: This should happen in `setUp()`.

src/Symfony/Component/Mailer/Tests/Transport/SendmailTransportTest.php
```diff
@@ -38,10 +38,27 @@ protected function tearDown(): void
...
         $t = new SendmailTransport();
         $this->assertEquals('smtp://sendmail', (string) $t);
     }
 
+    public function testToStringNonSmtp()
+    {
+        $previousSendmailPath = \ini_set('sendmail_path', '');
+        $t = new SendmailTransport('/usr/sbin/sendmail -t -i');
+        \ini_set('sendmail_path', $previousSendmailPath);
```
derrabus, line 50: This should happen in `tearDown()`.

derrabus (changes requested):

SzymonKaminski:
Isn't this case of specifying `sendmail_path` covered by `native` transport?

Seb33300:
I think this feature already exists using the `native` DSN protocol: https://symfony.com/doc/current/mailer.html#using-built-in-transports

This was introduced by https://github.com/symfony/symfony/pull/36131 to fix that issue: https://github.com/symfony/symfony/issues/36842

src/Symfony/Component/Mailer/Tests/Transport/SendmailTransportTest.php
nicolas-grekas: or we should use `$this->iniSet()`?

src/Symfony/Component/Mailer/Transport/SendmailTransport.php
```diff
@@ -59,6 +60,8 @@ public function __construct(string $command = null, EventDispatcherInterface $di
             }
 
             $this->command = $command;
+        } else {
+            $this->command = \ini_get('sendmail_path') ?: self::SENDMAIL_FHS_PATH;
```
nicolas-grekas, line 64: This will change the default behavior: before: `-bs`, after `-t -i` on my Ubuntu.
We need another way to opt-in for reading from ini

src/Symfony/Component/Mailer/Transport/SendmailTransport.php
```diff
@@ -83,7 +86,7 @@ public function __toString(): string
             return (string) $this->transport;
         }
 
-        return 'smtp://sendmail';
+        return 'sendmail://default';
```
nicolas-grekas, line 89: is that a bugfix? I think this should go on 5.4

src/Symfony/Component/Mailer/Tests/Transport/SendmailTransportTest.php
derrabus: 👀 sebastianbergmann/phpunit#5214

jtojnar: It is not relevant to all tests, though.

jtojnar:
Actually, the native transport is what I am looking for. Unfortunately it is not available in the Symfony version used by Wallabag (4.4).

PR #51756, opened 2023-09-26
[Messenger] RejectRedeliveredMessageException should not be retried

nikophil:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT
| Doc PR        | I think it is not needed

Hello,

when a `RejectRedeliveredMessageException` is thrown once, it is actually retried because it does not implement `UnrecoverableExceptionInterface` but we know it will always because the stamp `AmqpReceivedStamp` will always be there


nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @nikophil.

PR #50438, opened 2023-05-25
Add is_valid function to Expression constraint

DEVizzent:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 for features 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Tickets       | Fix https://github.com/symfony/symfony/issues/11940
| License       | MIT
| Doc PR        | TODO

Add `is_valid` function to the Expression constraint, api the same as `ValidatorInterface::validate`

PR #51761, opened 2023-09-27
[Validator] Fix registering "is_valid()" for `#[Expression]`

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | #50438
| License       | MIT
| Doc PR        | -

@bendavies can you please confirm this fixes the issue for you?


bendavies:
yes it does.

could use tags like the with [security](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Compiler/AddExpressionLanguageProvidersPass.php) and [routing](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/AddExpressionLanguageProvidersPass.php)

nicolas-grekas:
Let's wait for someone to need that :)
The alternative would be to redefine the `validator.expression_language_provider` service.

PR #51741, opened 2023-09-25
[ObjectMapper] Object to Object mapper component

soyuka:
| Q             | A
| ------------- | ---
| Branch?       | 7.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT
| Doc PR        | TBD this description will be used as docs

## Why ?

In the train back from [API Platform Con](https://api-platform.com/con/2023/) and after watching @weaverryan conference about API Platform's feature that helps [separating Doctrine Entities from POPOs](https://symfonycasts.com/screencast/api-platform-extending/state-options), we had the feeling that such a component is definitely missing from the Symfony ecosystem. Current implementations (automapper-plus, janephp/automapper) are good but feel to complicated when you want something accessible (DX and complexity). Here we're not trying to have exceptional performance, we just want an accessible API with few mandatory features. Basically those are:

- map properties from a class/object to another class/object (`Map(target: A::class)`)
- configure the property name each property is mapped to (`Map(target: 'prop')`)
- change the value during transformation (`Map(transform: 'ucfirst')`)
- control whether the value should be mapped or not (`Map(if: 'boolval')`)
- reuse objects if possible (works better with doctrine UOW)

Other implementation details:
- if a property exists in B, it'll "automatically" be mapped (doesn't handle type errors, should we ?)
- if a property doesn't exists in B it'll "automatically" **not** be mapped

The rest is not "auto".

## Mapper Component

The Mapper component allows you to map an object to another object,
facilitating the mapping using attributes.

### Usage

```php
use Symfony\Component\ObjectMapper\ObjectMapper;
use Symfony\Component\ObjectMapper\Attributes\Map;

// This maps class `A` to class `B`.
#[Map(B::class)]
class A
{
    // This maps A::foo to B::bar.
    #[Map(target: 'bar')]
    public string $foo;

    // This calls ucfirst on A::transform
    #[Map(transform: 'ucfirst')]
    public string $name;

    // This doesn't map A::bar if it's value is falsy.
    #[Map(if: 'boolval')]
    public bool $bar = false;
}

$mapper = new ObjectMapper();
$mapper->map(new A);
```
The `Map` attribute has the following signature: 

```php
#[Attribute(Attribute::TARGET_CLASS | Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
final class Map
{
    /**
     *
     * @param null|string|class-string                                  $source     The property or the class to map to
     * @param null|string|class-string                                  $target     The property or the class to map from
[...]

PR #51769, opened 2023-09-28
[ErrorHandler] no intersection type with callable

soyuka:
| Q             | A
| ------------- | ---
| Branch?       | not sure please let me know which one to target
| Bug fix?      | yes
| License       | MIT

https://wiki.php.net/rfc/pure-intersection-types#supported_types says:

>  Although an intersection with callable can make sense (e.g. string&callable), we think it is unwise and points to a bug.

We can't patch a return type if there's a callable intersecting with other types, which can theoretically be supported. 

Blocks #51741.

### DIFF
--- a/src/Symfony/Component/ErrorHandler/DebugClassLoader.php
+++ b/src/Symfony/Component/ErrorHandler/DebugClassLoader.php
@@ -848,6 +848,10 @@ class DebugClassLoader
         $phpTypes = [];
         $docTypes = [];
 
+        if (\count($typesMap) > 1 && array_key_exists('callable', $typesMap)) {
+            return;
+        }
+
         foreach ($typesMap as $n => $t) {
             if ('null' === $n) {
                 $nullable = true;


src/Symfony/Component/ErrorHandler/DebugClassLoader.php
```diff
@@ -848,6 +848,10 @@ private function setReturnType(string $types, string $class, string $method, str
         $phpTypes = [];
         $docTypes = [];
 
+        if (\count($typesMap) > 1 && array_key_exists('callable', $typesMap)) {
```
stof, line 851: wouldn't this also return for _union_ types here as you don't check `$glue` ?

soyuka: Mhh my bad, I made a mistake where it's not possible do declare a property type using `callable` but it's actually allowed to use `union` between callable and other types. Thanks!

PR #33189, opened 2019-08-15
New welcome page for 4.4 LTS & 5.0

yceruto:
| Q             | A
| ------------- | ---
| Branch?       | 4.4
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

A new LTS version of Symfony is coming and a new major version too. That deserves a little more :heart: than usual.

This picture simulates two components of the life: the Ocean and the sky/Universe, with the aim of inspiring calm at the same time desire to build good & great things:

 welcome-page

The HUE value changes randomly each time you enter the homepage. That's just an idea to make it dynamic. I invite you to try it and send your ideas too. I'd like this experience to be extremely exciting for newcomers.

About all text & messages, this would be a good time to say what people want to hear :) Symfony's amazing \o/. Some communication skills are required here.

I'm not a designer, so I'd want some expert on this topic destroying the proposal making it even better :)

HTML+CSS only: I tried to use some modern features that were compatible with most browsers, but I might be forgetting something.

The icons were taken from https://material.io/resources/icons/ (free), but I don't know how to deal with this license:
> Our icons are free for everyone to use. Please don’t try to sell them.
> Available under [Apache license version 2.0](https://www.apache.org/licenses/LICENSE-2.0.html).

Should we copy it somewhere? should it be visible to the user?

**Before:**
old-welcome-page

A good first impression is really, really important. Is it worth it? Thoughts!

ping to my friend @javiereguiluz this is for sure something you would like to check.

Cheers!

PR #51771, opened 2023-09-28
Update the design of the Symfony Welcome Page

javiereguiluz:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT

This PR refreshes the design of the Symfony Welcome Page for Symfony 7.0. This is the page that you see when you create a new Symfony app and browse it without making any change to it.

The new design keeps most of the features of the previous design, including the random color each time you reload the page.

This is how it looks:

| Smartphone | Tablet | Desktop
| ---------- | ------ | -------
| new-welcome-page-smartphone | new-welcome-page-tablet | new-welcome-page-desktop

This is how it looked **before**:

| Smartphone | Tablet | Desktop
| ---------- | ------ | -------
| before-welcome-page-smartphone | before-welcome-page-tablet | before-welcome-page-desktop


antonkomarev:
It will be great to add screenshots how it looks like right now.

javiereguiluz:
@antonkomarev I've updated the PR to include the before screenshots.

evertharmeling:
Nice! Could we maybe use more specific icons instead of checkmarks for the 3 main bullet points? Something like: 
- A  [tag](https://tabler-icons.io/i/tag) for the version
- A  [directory](https://tabler-icons.io/i/folder) for the path
- A  [terminal](https://tabler-icons.io/i/terminal-2) (could not a find a specific icon for debug :)) for the debug mode

The list contains 'important' information and I feel it's easily to not read them because it's just another list of bullet points.

smnandre:
Really nice and clean.

I'd suggest separating the 3 factual (& positive-ish) elements from the item requiring user action 

-- 

✅  You are using Symfony XXX version
✅  Your application is ready at XXX 
✅  Debug mode is activated

❓ Your homepage is not configured: create your first page.

-- 

I have a feeling that more people will wonder, 'cool, it works, what do I do now?' than 'why is this page displayed?'... but with no certitude.

antonkomarev:
Good point, but I suggest having a warning sign near the debug mode:
⚠️ Debug mode is activated

javiereguiluz:
Thank you all for your nice reviews. I'm going to iterate over some of these ideas and will come back with an update.

derrabus:
Since we usually don't merge features into x.0 releases, what's preventing us from merging this into 6.4?

stof:
I tend to agree with @derrabus here. We communicate regularly than the X.4 releases contains all the new features and Y.0 cleans the deprecated APIs. So this should go in 6.4.

src/Symfony/Component/HttpKernel/Resources/welcome.html.php
```diff
@@ -2,123 +2,253 @@
...
+            <section class="checks">
+                <ul>
+                    <li>
+                        <span><?= renderCheckIconSvg() ?></span>
+                        <span>You are using Symfony <strong><?= $version ?></strong> version</span>
+                    </li>
+
+                    <li>
+                        <span><?= renderCheckIconSvg() ?></span>
+                        <span>Your application is ready at: <code translate="no" class="notranslate project_dir_path"><?= $projectDir ?></code></span>
```
ro0NL, line 106: i'd make this the first list-item, the major version is already on top in heading1

javiereguiluz:
@derrabus @stof this is the classic "engineering vs product" problem.

Engineering team says that new features must go to 6.4 because X.0 branches don't get new features.

But for Product team it's important that this new design is only visible in Symfony 7.0 and up, because this is the new design for Symfony 7, not Symfony 6.4 and 7.

-----

Personally, I'd make an exception here and merge this into 7.0 only. But I'll do what the team decides 🙏  Thanks!

94noni:
the previous PR updating the design was done on [v4.4+v5.0](https://github.com/symfony/symfony/pull/33189)
it would be nice to keep same logic (moreover as .4 are LTS so displayed during long time after initial release)

fabpot:
7.0 sounds fine to me (it's not "a feature" per se but more like a refresh of the design for a major version).

MGDSoft:
What are your thoughts on adding a subtle motion effect to the stars and water?

src/Symfony/Component/HttpKernel/Resources/welcome.html.php
```diff
@@ -2,123 +2,253 @@
...
+        main article { padding: 30px 0 0 10px; }
+        main h2 { align-items: center; display: flex; font-size: 18px; font-weight: 600; margin: 0; }
+        main h2 span { display: inline-flex; }
+        main h2 svg { height: 28px; width: 28px; margin-right: 10px; }
+        main article section + section { margin-top: 45px; }
+        main article section ul { padding-left: 17px; }
+        main article section li + li { margin-top: 15px; }
 
         @media (min-width: 768px) {
             @-webkit-keyframes fade-in { 0% { opacity: 0; } 100% { opacity: 1; } }
```
stof: Wa can drop the prefixed at-rule here. The non-prefixed one is supported by all browsers since years: https://caniuse.com/mdn-css_at-rules_keyframes (webkit and Chrome both shipped the unprefixed name in 2015)

src/Symfony/Component/HttpKernel/Resources/welcome.html.php
```diff
@@ -2,123 +2,253 @@
...
             @keyframes fade-in { 0% { opacity: 0; } 100% { opacity: 1; } }
             .sf-toolbar { opacity: 0; -webkit-animation: fade-in 1s .2s forwards; animation: fade-in 1s .2s forwards; z-index: 99999; }
```
stof: The `-webkit-animation` prefix is also useless, as only browsers from before 2015 need it.

stof (changes requested):

sstok:
> What are your thoughts on adding a subtle motion effect to the stars and water?

Motion effects don't add anything useful here, and might cause trouble on the front of accessibility as movement can cause motion sickness for people with motion sensitivity.

https://www.linkedin.com/advice/0/how-do-you-avoid-animation-overload-motion

MGDSoft:
> > What are your thoughts on adding a subtle motion effect to the stars and water?
> 
> Motion effects don't add anything useful here, and might cause trouble on the front of accessibility as movement can cause motion sickness for people with motion sensitivity.
> 
> https://www.linkedin.com/advice/0/how-do-you-avoid-animation-overload-motion

Well, then, 50% of webpages have motion on their homepage, and they are all incorrect.

https://laravel.com/
https://nextjs.org/
https://www.meteor.com/

I believe adding some effects will generate more excitement for newcomers interested in learning Symfony. For example, when I install Drupal, I encounter an unattractive default page that prompts me to leave quickly.

fabpot:
The current page is animated, right? So, I suppose the new version will be as well.

MGDSoft:
> The current page is animated, right? So, I suppose the new version will be as well.

If I'm correct, @javiereguiluz, I didn't notice any effect on the page; I only saw the background color change (which is cool).

smnandre:
```css
@media (prefers-reduced-motion) {
  /* styles to apply if a user's device settings are set to reduced motion */
}
```

I believe animation is not a problem if we leverage this media query

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

MGDSoft:
It's just an idea but something like this, We can do better, of course, if everyone agrees.

https://github.com/symfony/symfony/assets/3816465/537f9c1b-bc50-473f-998b-c880067fae2d

javiereguiluz:
@MGDSoft can you please share the CSS code that you used to add that subtle animation? Thanks!

MGDSoft:
ofc!

Waves (SVG is the same)
```javascript
function renderWavesSvg()
{
    return <<<SVG
                <svg aria-hidden="true" focusable="false" style="pointer-events: none" class="wave" width="100%" height="50px" preserveAspectRatio="none" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 1920 75"><defs>
                <style>
                .a{fill:none}.b{clip-path:url(#a)}.c,.d {fill: var(--light-color); }.d{opacity:0.5;isolation:isolate;}
                @keyframes waveMotion {
                    0% {
                        transform: scaleX(1) translateX(0);
                    }
                    50% {
                        transform: scaleX(1.5) translateX(-30%);
                    }
                    0% {
                        transform: scaleX(1) translateX(0);
                    }
                }
                .b:nth-child(5) {
                    animation: waveMotion 10s infinite alternate;
                }
                
                .b:nth-child(2) {
                    animation: waveMotion 8s infinite alternate;
                }
                
                .b:nth-child(3) {
                    animation: waveMotion 9s infinite alternate;
                }
                
                .b:nth-child(4) {
                    animation: waveMotion 4s infinite alternate;
                }
                </style><clipPath id="a"><rect class="a" width="1920" height="75"></rect></clipPath></defs><g class="b"><path class="c" d="M1963,327H-105V65A2647.49,2647.49,0,0,1,431,19c217.7,3.5,239.6,30.8,470,36,297.3,6.7,367.5-36.2,642-28a2511.41,2511.41,0,0,1,420,48"></path></g><g class="b"><path class="d" d="M-127,404H1963V44c-140.1-28-343.3-46.7-566,22-75.5,23.3-118.5,45.9-162,64-48.6,20.2-404.7,128-784,0C355.2,97.7,341.6,78.3,235,50,86.6,10.6-41.8,6.9-127,10"></path></g><g class="b"><path class="d" d="M1979,462-155,446V106C251.8,20.2,576.6,15.9,805,30c167.4,10.3,322.3,32.9,680,56,207,13.4,378,20.3,494,24"></path></g><g class="b"><path class="d" d="M1998,484H-243V100c445.8,26.8,794.2-4.1,1035-39,141-20.4,231.1-40.1,378-45,349.6-11.6,636.7,73.8,828,150"></path></g></svg>
            SVG;
}
```

STARS

```css
@keyframes stars{
    0% {
        transform: translateX(0) translateY(0px);
    }
    33% {
        transform: translateX(-20px)  translateY(-20px);
    }
    66% {
        transform: translateX(-40px) translateY(-40px);
    }
    100% {
        transform: translateX(0px)  translateY(0px);
    }
}
header:after {
    animation: stars 15s infinite;
}
```

javiereguiluz:
Thank you all for your reviews. I've made these changes:

* Use different icons in the list of items
* Separate the "What do do next?" item and turned it into "Next Step"
* Added the nice animation proposed by @MGDSoft (thanks!) but I made it slower to make it more subtle
* Added a slight blur to stars to increase readability

This is how it looks now:

| Smartphone | Tablet | Desktop
| ---------- | ------ | -------
| smartphone | tablet | desktop


And this is in movement:

https://github.com/symfony/symfony/assets/73419/7b5d3219-e8dd-4864-ac73-cd99ab18a883

yceruto:
I love these improvements! ❤️

yceruto:
On a wide screen, it looks great, but perhaps on mobile we should add more space between texts as it looks a bit cramped.

yceruto (approved):

src/Symfony/Component/HttpKernel/Resources/welcome.html.php
```diff
@@ -2,123 +2,356 @@
...
+            .info ul { margin: 45px 0; padding-left: 15px; }
+            .info li + li { margin-top: 20px; }
+            .info svg { height: 28px; width: 28px; margin-right: 13px; position: relative; top: -2px; }
+            .info code { margin-top: 10px; }
+            .info .next-step { display: flex; align-items: center; margin: 0 0 45px; }
+            .info .next-step svg { display: inline-flex; margin: 0 10px; top: 0; }
+            main article { display: flex; justify-content: space-between; padding-top: 45px; }
+            main article section + section { margin-top: 0; }
+            main h2 { font-size: 21px; }
+            main h2 svg { height: 36px; width: 36px; margin-righ: 15px; }
```
Aweptimum: little typo here, should be `margin-right`

smnandre:
Love it, thx @javiereguiluz for this!

src/Symfony/Component/HttpKernel/Resources/welcome.html.php
javiereguiluz: Good catch. Fixed. Thanks.

javiereguiluz:
I've tweaked some margins/paddings/spacing as requested. Thanks.

fabpot (approved):

fabpot:
Thank you @javiereguiluz.

PR #51773, opened 2023-09-28
[Mailer] [Mailgun] Fix outlook sender

### LINKED ISSUE #51596: Sending through Mailgun mailer with a subdomain results in "on behalf of" in sender name

6.3.2

### Description

I am using symfony/mailgun-mailer (latest 6.3.2).

I have a subdomain set up: mg.domain.com and I send emails from user@domain.com. This should work when all dkim and spf align. However, the headers are not set correctly.

There is [this issue](https://serverfault.com/questions/1097879/sender-and-from-header-domain-mismatch-causes-malformed-display-in-outlook) with exact description of the problem and this [solution](https://stackoverflow.com/a/45445015/397646)

Can we please add this header as well? 

### How to reproduce

* Create a subdomain in Mailgun like mg,domain.com
* Send an email to a Microsoft email from user@domain.com

### Possible Solution

https://stackoverflow.com/a/45445015/397646

### Additional Context

_No response_

Romanavr:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix https://github.com/symfony/symfony/issues/51596
| License       | MIT

This PR solves the problem related to displaying the sender in the OutLook email client when sending an email from a subdomain.
Check more details by following these links:
https://serverfault.com/questions/1097879/sender-and-from-header-domain-mismatch-causes-malformed-display-in-outlook
https://stackoverflow.com/questions/28401673/removing-on-behalf-of-when-sending-mail-using-mailgun/45445015#45445015

### DIFF
--- a/src/Symfony/Component/Mailer/Bridge/Mailgun/Tests/Transport/MailgunApiTransportTest.php
+++ b/src/Symfony/Component/Mailer/Bridge/Mailgun/Tests/Transport/MailgunApiTransportTest.php
@@ -61,6 +61,8 @@ class MailgunApiTransportTest extends TestCase
         $deliveryTime = (new \DateTimeImmutable('2020-03-20 13:01:00'))->format(\DateTimeInterface::RFC2822);
 
         $email = new Email();
+        $envelope = new Envelope(new Address('alice@system.com'), [new Address('bob@system.com')]);
+        $email->getHeaders()->addTextHeader('h:sender', $envelope->getSender()->toString());
         $email->getHeaders()->addTextHeader('h:X-Mailgun-Variables', $json);
         $email->getHeaders()->addTextHeader('h:foo', 'foo-value');
         $email->getHeaders()->addTextHeader('t:text', 'text-value');
@@ -69,7 +71,6 @@ class MailgunApiTransportTest extends TestCase
         $email->getHeaders()->addTextHeader('template', 'template-value');
         $email->getHeaders()->addTextHeader('recipient-variables', 'recipient-variables-value');
         $email->getHeaders()->addTextHeader('amp-html', 'amp-html-value');
-        $envelope = new Envelope(new Address('alice@system.com'), [new Address('bob@system.com')]);
 
         $transport = new MailgunApiTransport('ACCESS_KEY', 'DOMAIN');
         $method = new \ReflectionMethod(MailgunApiTransport::class, 'getPayload');
@@ -78,6 +79,8 @@ class MailgunApiTransportTest extends TestCase
         $this->assertArrayHasKey('h:X-Mailgun-Variables', $payload);
         $this->assertEquals($json, $payload['h:X-Mailgun-Variables']);
 
+        $this->assertArrayHasKey('h:sender', $payload);
+        $this->assertEquals($envelope->getSender()->toString(), $payload['h:sender']);
         $this->assertArrayHasKey('h:foo', $payload);
         $this->assertEquals('foo-value', $payload['h:foo']);
         $this->assertArrayHasKey('t:text', $payload);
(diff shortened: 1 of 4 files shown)


nicolas-grekas:
Thank you @Romanavr.

svendecabooter:
It seems like this change is causing issues for me, when sending mails to Gmail accounts via Mailgun.
I am getting errors like the following in Mailgun:

> 5.7.1 This message is not RFC 5322 compliant. There are multiple Sender\n5.7.1 headers. To reduce the amount of spam sent to Gmail, this message has\n5.7.1 been blocked.

After some digging, it seems my sent message contains 2 sender headers, one capitalized and one not:

> Sender: mail@example.com
> sender: mail@example.com

I am still trying to figure out where the capitalized "Sender" header gets added in my flow, but haven't pinned that down yet.
However I'd like to point this issue out, for people encountering the same problem.
If I have more information that is relevant, I will update here.

I am not familiar enough with email headers, but if it is considered best practice to capitalize them, then maybe the custom header added via this change should also be capitalized?

Romanavr:
> It seems like this change is causing issues for me, when sending mails to Gmail accounts via Mailgun. I am getting errors like the following in Mailgun:
> 
> > 5.7.1 This message is not RFC 5322 compliant. There are multiple Sender\n5.7.1 headers. To reduce the amount of spam sent to Gmail, this message has\n5.7.1 been blocked.
> 
> After some digging, it seems my sent message contains 2 sender headers, one capitalized and one not:
> 
> > Sender: [mail@example.com](mailto:mail@example.com)
> > sender: [mail@example.com](mailto:mail@example.com)
> 
> I am still trying to figure out where the capitalized "Sender" header gets added in my flow, but haven't pinned that down yet. However I'd like to point this issue out, for people encountering the same problem. If I have more information that is relevant, I will update here.
> 
> I am not familiar enough with email headers, but if it is considered best practice to capitalize them, then maybe the custom header added via this change should also be capitalized?

I just tested - everything is ok for me, what transport type are you using? API?

svendecabooter:
@Romanavr yes API.

Romanavr:
> @Romanavr yes API.

I was able to reproduce. Indeed, good point about capitalize the header.

Romanavr:
@svendecabooter Please try [52017#fix](https://github.com/symfony/symfony/pull/52017) and let me know if it works for you

spajxo:
Hello, 

I have come across a problem caused by this change. 

The header h:Sender is added [here](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Mailer/Bridge/Mailgun/Transport/MailgunApiTransport.php#L88) as UnstructuredHeader.


When it is added into payload [here](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Mailer/Bridge/Mailgun/Transport/MailgunApiTransport.php#L144), sometimes it encodes incorrectly


The issue occurs only in some cases when using Sender Address with name part with diacritics, because of the difference in encoding of UnstructuredHeader and MailboxHeader.

example:
```php
$unstructuredHeader = new UnstructuredHeader('From', '"Klaudie Šulcová" <foo@bar.com>');
$mailboxHeader = new MailboxHeader('From', new Address('foo@bar.com', 'Klaudie Šulcová'));

$unstructuredHeader->getBodyAsString(),  // ""Klaudie =?utf-8?Q?=C5=A0ulcov=C3=A1=22?= <foo@bar.com>"
$mailboxHeader->getBodyAsString()           // "Klaudie =?utf-8?Q?=C5=A0ulcov=C3=A1?= <foo@bar.com>"
```

I discovered this problem because some mailservers complained about the Sender header being invalid.  

A simple solution to fix this problem would be adding the `h:Sender` header as MailboxHeader [here](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Mailer/Bridge/Mailgun/Transport/MailgunApiTransport.php#L88)
```php
$headers->addMailboxHeader('h:Sender', $envelope->getSender());
``` 

I'll send a PR

PR #51777, opened 2023-09-28
[Validator] Add additional versions  (`*_NO_PUBLIC`, `*_ONLY_PRIV` & `*_ONLY_RES`) in IP address constraint

Ninos:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

Possibility to allow only private or only reserved ips.

- Enhancement: Add `*_NO_PUBLIC`, `*_ONLY_PRIV` & `*_ONLY_RES` as possible versions in `Ip` constraint
- Enhancement: Possibility to use all `Ip` versions in `Cidr` constraint

~~PS: In `Ip` constraint I changed the default version from `V4` to `ALL`, think it's time to support also `V6` by default :-) Otherwise we can revert that, but normally should not break much...~~


UPGRADE-6.4.md
```diff
@@ -157,6 +157,7 @@ Templating
 Validator
 ---------
 
+ * Changed default version flag from `V4` to `ALL` (also `V6`) in `Ip` constraint
```
nicolas-grekas: This is a BC break, we can't do this I fear.

src/Symfony/Component/Validator/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ CHANGELOG
 6.4
 ---
 
+ * Add `Mac` constraint
```
nicolas-grekas: MacAddress?

src/Symfony/Component/Validator/Constraints/Ip.php
```diff
@@ -19,10 +19,12 @@
  * Validates that a value is a valid IP address.
  *
  * @Annotation
+ *
```
nicolas-grekas: should be reverted, that's a false positive

src/Symfony/Component/Validator/Constraints/Ip.php
```diff
@@ -80,7 +100,7 @@ class Ip extends Constraint
      */
     protected static $errorNames = self::ERROR_NAMES;
 
-    public $version = self::V4;
+    public $version = self::ALL;
```
nicolas-grekas:
```suggestion
    public $version = self::V4;
```

src/Symfony/Component/Validator/Constraints/Mac.php
```diff
@@ -0,0 +1,61 @@
...
+namespace Symfony\Component\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\Exception\InvalidArgumentException;
+
+/**
+ * Validates that a value is a valid MAC address.
+ *
+ * @Annotation
+ *
```
nicolas-grekas:
```suggestion
```

src/Symfony/Component/Validator/Constraints/Mac.php
```diff
@@ -0,0 +1,61 @@
...
+
+    protected const ERROR_NAMES = [
+        self::INVALID_MAC_ERROR => 'INVALID_MAC_ERROR',
+    ];
+
+    /**
+     * @deprecated since Symfony 6.1, use const ERROR_NAMES instead
+     */
+    protected static $errorNames = self::ERROR_NAMES;
+
```
nicolas-grekas:
```suggestion
```

src/Symfony/Component/Validator/Constraints/Mac.php
```diff
@@ -0,0 +1,61 @@
...
+    public $message = 'This is not a valid MAC address.';
```
nicolas-grekas:
```suggestion
    public string $message = 'This is not a valid MAC address.';
```

src/Symfony/Component/Validator/Constraints/Mac.php
```diff
@@ -0,0 +1,61 @@
...
+ * Validates that a value is a valid MAC address.
+ *
+ * @Annotation
+ *
+ * @Target({"PROPERTY", "METHOD", "ANNOTATION"})
+ *
+ * @author Ninos Ego <me@ninosego.de>
+ */
+#[\Attribute(\Attribute::TARGET_PROPERTY | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
+class Mac extends Constraint
```
nicolas-grekas:
```suggestion
class MacAddress extends Constraint
```

src/Symfony/Component/Validator/Constraints/Mac.php
```diff
@@ -0,0 +1,61 @@
...
+
+    /** @var callable|null */
+    public $normalizer;
+
+    public function __construct(
+        array $options = null,
+        string $message = null,
+        callable $normalizer = null,
+        array $groups = null,
+        mixed $payload = null
```
nicolas-grekas:
```suggestion
        mixed $payload = null,
```

src/Symfony/Component/Validator/Constraints/MacValidator.php
```diff
@@ -0,0 +1,56 @@
...
+ * Validates whether a value is a valid MAC address.
+ *
+ * @author Ninos Ego <me@ninosego.de>
+ */
+class MacValidator extends ConstraintValidator
+{
+    /**
+     * @return void
+     */
+    public function validate(mixed $value, Constraint $constraint)
```
nicolas-grekas:
```suggestion
    public function validate(mixed $value, Constraint $constraint): void
```

nicolas-grekas (commented):
Thanks for the PR. Here are some comments. Changing the defaults look undesired to me since it would change the behavior of existing apps without notice.
Can you also please update the title of your PR so that it can be useful for readers in the auto-generated changelog when doing the release?

xabbuh:
I am not convinced that a MAC address validator is a common enough use case to be included in the Symfony core.

src/Symfony/Component/Validator/CHANGELOG.md
Ninos: Changed back to `Mac`, `Ip` is also not called `IpAddress`, to be a bit consistent. Also there are no different meanings of that word (I think). But we can change that to `MacAddress` or kick it completely out, I'm fine with both that :-)

Ninos:
> Thanks for the PR. Here are some comments. Changing the defaults look undesired to me since it would change the behavior of existing apps without notice. Can you also please update the title of your PR so that it can be useful for readers in the auto-generated changelog when doing the release?

Thx for review, merged your commits, reverted your comments & changed title :-)

stof:
If we want to change the default version, we would have to make it configurable in the validator:

- make the `version` field of the constraint nullable and default to `null`
- configure a default version in the validator (defaulting to `V4` for BC reasons), that is used when the constraint version is `null`
- add a configuration setting in FrameworkBundle to configure the default IP version.

But to me, this deserves a separate PR than this one adding new version constants. It deserves to have a separate discussion instead of being bundled in the same decision to merge or no (note that I would have separated the Mac constraint as well, to allow separating the decisions as well).

Ninos:
Removed mac constraint from MR, see also #51862

src/Symfony/Component/Validator/Constraints/Ip.php
```diff
@@ -46,6 +47,16 @@ class Ip extends Constraint
     public const V6_ONLY_PUBLIC = '6_public';
     public const ALL_ONLY_PUBLIC = 'all_public';
 
+    // adds inverse FILTER_FLAG_NO_PRIV_RANGE
+    public const V4_ONLY_PRIV = '4_priv';
```
OskarStark: PRIV is an abbreviation for "PRIVATE" ? If yes I would suffix the constants with PRIVATE instead of PRIV

Same for the other constants.

Excepts this is commonly used like this, what I don't know

Seb33300: He reused the same syntax as the existing `*_NO_PRIV`

src/Symfony/Component/Validator/Constraints/Ip.php
```diff
@@ -64,6 +75,14 @@ class Ip extends Constraint
...
         self::V6_ONLY_PUBLIC,
         self::ALL_ONLY_PUBLIC,
+
+        self::V4_ONLY_PRIV,
+        self::V6_ONLY_PRIV,
+        self::ALL_ONLY_PRIV,
+
+        self::V4_ONLY_RES,
+        self::V6_ONLY_RES,
+        self::ALL_ONLY_RES,
```
nicolas-grekas: `RES` is super cryptic, let's go with the full `RESERVED` word

Let's duplicate all consts to add these full-word names as aliases (PRIV also)

nicolas-grekas (commented):
LGTM with one comment about naming

Ninos:
Seems my responses are still on pending... I'll reply as seperate comment:

> I used the same suffix, which was used as filter flag (e.g. `FILTER_FLAG_NO_PRIV_RANGE`, see also: https://www.php.net/manual/de/filter.filters.flags.php).
> 
> Also this constraint already contains constants with `_PRIV` & `_RES` suffix, see already existing constant `*_ONLY_PRIV`:
https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Validator/Constraints/Ip.php#L35
> 
> We cannot just rename new constants to `PRIVATE` or `RESERVED`, this will be inconsistent. We could also rename old constrants to new suffix, but then we need to add additional complexity for backward-compatibility...

Seb33300:
> We cannot just rename new constants to `PRIVATE` or `RESERVED`, this will be inconsistent. We could also rename old constants to new suffix, but then we need to add additional complexity for backward-compatibility...

@nicolas-grekas suggested to duplicate all names in his latest comment.

Ninos:
@Seb33300 @nicolas-grekas @OskarStark renamed & deprecated old constants. Hope it's fine now :-)

Ninos:
I've also updated `Cidr` constraint to use possible versions defined in `Ip` constraint. Please check also :-)

src/Symfony/Component/Validator/Constraints/Cidr.php
```diff
@@ -55,31 +76,40 @@ class Cidr extends Constraint
...
+    /** @var callable|null */
+    public $normalizer;
+
     public function __construct(
         array $options = null,
         string $version = null,
         int $netmaskMin = null,
         int $netmaskMax = null,
         string $message = null,
+        callable $normalizer = null,
```
nicolas-grekas: this should be added last to not break BC

src/Symfony/Component/Validator/Constraints/Cidr.php
```diff
@@ -55,31 +76,40 @@ class Cidr extends Constraint
...
         array $groups = null,
         $payload = null
     ) {
         $this->version = $version ?? $options['version'] ?? $this->version;
 
-        if (!\array_key_exists($this->version, self::NET_MAXES)) {
+        if (!\in_array($this->version, array_keys(self::NET_MAXES))) {
```
nicolas-grekas: why changing this?

src/Symfony/Component/Validator/Constraints/Ip.php
```diff
@@ -31,39 +32,101 @@ class Ip extends Constraint
...
+
+    // adds inverse FILTER_FLAG_NO_RES_RANGE
+    public const V4_ONLY_RESERVED = '4_reserved';
+    public const V6_ONLY_RESERVED = '6_reserved';
+    public const ALL_ONLY_RESERVED = 'all_reserved';
+
     public const INVALID_IP_ERROR = 'b1b427ae-9f6f-41b0-aa9b-84511fbb3c5b';
 
+    /**
+     * @deprecated since Symfony 6.4, use const V4_NO_PRIVATE instead
```
nicolas-grekas: I wouldn't deprecate any constants: the cost of migrating to the new ones is too high IMHO compared to the benefit, which is mostly cosmetic
let's just keep aliases

src/Symfony/Component/Validator/Constraints/IpValidator.php
```diff
@@ -41,32 +83,19 @@ public function validate(mixed $value, Constraint $constraint)
             throw new UnexpectedValueException($value, 'string');
         }
 
-        $value = (string) $value;
+        $value = (string)$value;
```
nicolas-grekas: boo :)

src/Symfony/Component/Validator/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ CHANGELOG
 6.4
 ---
 
+ * Add `*_NO_PUBLIC`, `*_ONLY_PRIVATE` and `*_ONLY_RESERVED` versions to `Ip` constraint
+ * Possibility to use all `Ip` constraint versions for `Cidr` constraint
```
nicolas-grekas: please move the new lines to a new 7.1 section

nicolas-grekas:
(please rebase, don't merge - you can also squash as we won't keep the history)

Ninos:
> (please rebase, don't merge - you can also squash as we won't keep the history)

Ok now I have a problem.. :D Give me some min, may I'll create a new branch if possible to retag this MR :/

Ninos:
@nicolas-grekas I've created a new branch based on 7.1, hope this is fine :( Sry for spamming, I'll try to rebase my mac addresses branch instead :)

See also: #52658

PR #51778, opened 2023-09-29
[Console][Form][FrameworkBundle][HttpKernel][PropertyInfo][Validator][WebProfilerBundle] Remove optional before required param

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT

I'm guessing this trick was used to make arguments nullable for PHP prior to 7.1, but since Symfony now requires at least PHP 8.1 it shouldn't be necessary any more. As far as I can tell there's no breaking change: https://3v4l.org/USkgi .

There are still four places that this is used, but I believe those to be intentional test cases:
https://github.com/symfony/symfony/blob/38506585012540306b4a7166a4812d35ad6b8a3c/src/Symfony/Component/DependencyInjection/Tests/Fixtures/includes/autowiring_classes.php#L179-L184
https://github.com/symfony/symfony/blob/38506585012540306b4a7166a4812d35ad6b8a3c/src/Symfony/Component/DependencyInjection/Tests/Fixtures/includes/MultipleArgumentsOptionalScalarNotReallyOptional.php#L8-L13
https://github.com/symfony/symfony/blob/38506585012540306b4a7166a4812d35ad6b8a3c/src/Symfony/Component/VarDumper/Tests/Caster/ReflectionCasterTest.php#L653-L655
https://github.com/symfony/symfony/blob/38506585012540306b4a7166a4812d35ad6b8a3c/src/Symfony/Component/HttpKernel/Tests/DependencyInjection/RegisterControllerArgumentLocatorsPassTest.php#L589-L598

### DIFF
--- a/.github/expected-missing-return-types.diff
+++ b/.github/expected-missing-return-types.diff
@@ -13166,8 +13166,8 @@ diff --git a/src/Symfony/Component/Validator/Context/ExecutionContextInterface.p
 @@ -127,5 +127,5 @@ interface ExecutionContextInterface
       * @return void
       */
--    public function setNode(mixed $value, ?object $object, MetadataInterface $metadata = null, string $propertyPath);
-+    public function setNode(mixed $value, ?object $object, MetadataInterface $metadata = null, string $propertyPath): void;
+-    public function setNode(mixed $value, ?object $object, ?MetadataInterface $metadata, string $propertyPath);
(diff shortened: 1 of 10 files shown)


yceruto (approved):

nicolas-grekas:
Thank you @HypeMC.

PR #51782, opened 2023-09-29
[EventDispatcher] add event listener argument resolver

### LINKED ISSUE #51781: [EventDispatcher] Listener argument resolver
### Description

Hello,

I've noticed that in many listeners, we only use the event to get the subject, to which we apply our processing. Our tests check that the event contains our object, that it's of the right instance, and then we get to the heart of the function's actual behavior.

I think that if we could inject the subject directly, we'd gain a bit:
 - It respects the [law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) because it does not adds an extra layer (I agree it's debatable, the event isn't really an extra layer...)
- it reduces code but also tests
- business functions can become listeners by moving closer to the model

First, I thought of an ArgumentResolverInterface like the HttpKernel component, with the event, event name and dispatcher (the actual arguments) as the default resolver.
What makes me doubt this solution is the openness offered by an ArgumentResolverInterface. 

Ultimately, I just want a shortcut to event properties, not dependency injection.
Which leads me to think that an attribute on listener function parameters would be more appropriate.
An ArgumentResolverAttribute would itself determine what it wants to extract from the event and could be extended.

### Example

```php

class MyListener {
    public function preDummy(#[CustomSubjectResolver] Contract $subject /*, Event $event*/)
    {
        // This code is not necessary anymore 🔥
    	/*
        $subject = $event->getSubject();
        if (!$subject instanceof Contract) {
            throw new \RuntimeException('Not a contract');
        }
    	*/
    
        // Do my stuff
    }
}

```

KerianMM:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #51781 
| License       | MIT

Add listener function argument resolver to fix #51781  

It is a proposition but we can imagine same for workflows event for example

### DIFF
--- a/src/Symfony/Component/EventDispatcher/Attribute/ArgumentResolver/GenericEventSubjectResolver.php
+++ b/src/Symfony/Component/EventDispatcher/Attribute/ArgumentResolver/GenericEventSubjectResolver.php
@@ -1,0 +1,33 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\EventDispatcher\Attribute\ArgumentResolver;
+
+use Symfony\Component\EventDispatcher\Attribute\ArgumentResolverInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
+
+/**
+ * Resolves argument for event listener.
+ *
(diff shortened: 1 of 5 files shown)


nicolas-grekas (commented):
Thanks for the PR. My first impression is that this would impact performance too much on something that is called quite often and is on the hotpath. Also, it's adding complexity we might want to avoid.

stof:
I have the same feeling than @nicolas-grekas here.

fabpot:
Closing then. Thank you for proposing.

PR #51786, opened 2023-09-29
[AssetMapper] Always downloading vendor files

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | yes
| Deprecations? | no
| Tickets       | None
| License       | MIT

Hi!

I discovered a few problems with our CDN set up:

A) The `jsdelivr` CDN uses imports statements with specific versions - e.g. `import{Controller as t}from"/npm/@hotwired/stimulus@3.2.1/+esm";`. This could cause the same dependency to be downloaded (at different verisons) every time.

B) The `jspm` CDN has correct bare imports - e.g. `import '@hotwired/stimulus';`. However, to they also include some "in-package" relative imports - e.g.`import '../other-file.js'`. That is ok on the CDN, but to support local download, it means we ALSO need to download those files. This is possible, but adds quite a bit of complexity and would probably require an `importmap.php.local` file so we could keep track of all of the files needed for some `foo-bar` package.

### Solution

Talking with @nicolas-grekas, we're going to change AssetMapper to "download-only" - like Composer, npm, etc. The `importmap.php` file will change to look like:

```php
return [
    // no change for local files
    'app' => [
        'path' => './assets/app.js',
        'entrypoint' => true,
    ],

    // remote packages have only version
    'bootstrap' => [
        'version' => '5.3.2',
    ],
```

We will exclusively use `jsdelivr` as the "backend", as it avoids the "in-package" relative imports. So, every package == 1 file.

To download the remote packages, run:

```
php bin/console importmap:install
```

This downloads everything into an `assets/vendor/` directory (which will be ignored by git). 

### Other Notable Items

* Added `MappedAsset.isVendor` property, which is a simple check to see if the file lives in `assets/vendor/`. If it does, we still run the compilers, but we can silence invalid import warnings.

* Mapped assets ending in `.php` are excluded from AssetMapper. There could be edge-cases where you don't want this, but it felt like a safer default, and allowed me to easily NOT expose the new `vendor/assets/installed.php` file.

TODO

* [x] Re-add code to change downloaded contents to bare imports
* [x] Download vendor packages after update/require
* [x] In JsCompiler, don't warn on vendor invalid imports
* [x] Remove package resolver system
* [x] Fix tests
* [x]  Audit to see what tests are missing


weaverryan:
This is ready to go! And important for 6.4. Test failures look unrelated.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -939,10 +938,6 @@ private function addAssetMapperSection(ArrayNodeDefinition $rootNode, callable $
                             ->info('The directory to store JavaScript vendors.')
                             ->defaultValue('%kernel.project_dir%/assets/vendor')
                         ->end()
-                        ->scalarNode('provider')
```
stof, line 942: We should make it a deprecated no-op instead. FrameworkBundle is subject to the BC policy.

src/Symfony/Bundle/FrameworkBundle/Resources/config/schema/symfony-1.0.xsd
```diff
@@ -202,7 +202,6 @@
         <xsd:attribute name="importmap-path" type="xsd:string" />
         <xsd:attribute name="importmap-polyfill" type="xsd:string" />
         <xsd:attribute name="vendor-dir" type="xsd:string" />
-        <xsd:attribute name="provider" type="xsd:string" />
```
stof: this must be kept if we keep the config

src/Symfony/Component/AssetMapper/Command/ImportMapInstallCommand.php
```diff
@@ -11,12 +11,14 @@
 
 namespace Symfony\Component\AssetMapper\Command;
 
-use Symfony\Component\AssetMapper\ImportMap\ImportMapManager;
+use Symfony\Component\AssetMapper\Tests\ImportMap\RemotePackageDownloader;
```
stof: A command must not reference a class in the `Tests` namespace. this namespace is not shipped when installed packages from dist

stof (changes requested):

src/Symfony/Component/AssetMapper/Command/ImportMapInstallCommand.php
weaverryan: Good catch - that namespace being in `Tests` was on accident.

weaverryan:
Ready to go - test failures are expected (due to FWBundle not using the code from this PR)

chalasr:
Rebase needed

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -238,8 +193,8 @@ public function getRawImportMapData(): array
      */
     public static function parsePackageName(string $packageName): ?array
     {
-        // https://regex101.com/r/MDz0bN/1
-        $regex = '/(?:(?P<registry>[^:\n]+):)?((?P<package>@?[^=@\n]+))(?:@(?P<version>[^=\s\n]+))?(?:=(?P<alias>[^\s\n]+))?/';
+        // https://regex101.com/r/z1nj7P/1
+        $regex = '/((?P<package>@?[^=@\n]+))(?:@(?P<version>[^=\s\n]+))?(?:=(?P<alias>[^\s\n]+))?/';
```
weaverryan, line 187: The registry is a jspm concept, so I'm removing it.

src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ CHANGELOG
  * Add `array $tokenAttributes = []` optional parameter to `KernelBrowser::loginUser()`
  * Add support for relative URLs in BrowserKit's redirect assertion.
  * Change BrowserKitAssertionsTrait::getClient() to be protected
+ * Deprecate the `framework.asset_mapper.provider` config option.
```
nicolas-grekas: no final dots (same on L29 while at it)

src/Symfony/Component/AssetMapper/AssetMapperRepository.php
```diff
@@ -111,6 +111,11 @@ public function all(): array
                     continue;
                 }
 
+                // avoid potentially exposing PHP files
+                if (str_ends_with($file->getPathname(), '.php')) {
```
nicolas-grekas: getExtension?

src/Symfony/Component/AssetMapper/ImportMap/ImportMapConfigReader.php
```diff
@@ -116,4 +129,20 @@ public function getRootDirectory(): string
     {
         return \dirname($this->importMapConfigPath);
     }
+
+    private function extractVersionFromLegacyUrl(string $url): ?string
+    {
+        // URL pattern https://ga.jspm.io/npm:bootstrap@5.3.2/dist/js/bootstrap.esm.js
+        $lastAt = strrpos($url, '@');
+        if (false === $lastAt) {
```
nicolas-grekas: I'll ever stop suggesting those CS changes ;P
```suggestion
        if (false === $lastAt = strrpos($url, '@')) {
```

src/Symfony/Component/AssetMapper/ImportMap/ImportMapConfigReader.php
```diff
@@ -38,26 +38,42 @@ public function getEntries(): ImportMapEntries
...
         foreach ($importMapConfig ?? [] as $importName => $data) {
-            $validKeys = ['path', 'url', 'downloaded_to', 'type', 'entrypoint', 'version'];
+            $validKeys = ['path', 'version', 'type', 'entrypoint', 'url'];
             if ($invalidKeys = array_diff(array_keys($data), $validKeys)) {
                 throw new \InvalidArgumentException(sprintf('The following keys are not valid for the importmap entry "%s": "%s". Valid keys are: "%s".', $importName, implode('", "', $invalidKeys), implode('", "', $validKeys)));
             }
 
+            // should solve itself when the config is written again
+            if (isset($data['url'])) {
+                trigger_deprecation('symfony/asset-mapper', '6.4', 'The "url" option is deprecated, use "version" instead.');
```
nicolas-grekas, line 48: missing dep in composer.json

src/Symfony/Component/AssetMapper/Command/ImportMapInstallCommand.php
```diff
@@ -36,8 +39,36 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+
+            if ('finished' === $event) {
+                ++$finishedCount;
+                $progressBar->advance();
+            }
+        });
+        $progressBar->finish();
+        $progressBar->clear();
+
+        if (0 === \count($downloadedPackages)) {
```
nicolas-grekas: another CS thing I'm not gonna give up with ;P
```suggestion
        if (!$downloadedPackages) {
```

src/Symfony/Component/AssetMapper/Command/ImportMapInstallCommand.php
```diff
@@ -36,8 +39,36 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+            $io->success('No assets to install.');
+
+            return Command::SUCCESS;
+        }
+
+        $io->success(sprintf(
+            'Downloaded %d asset%s into %s.',
+            \count($downloadedPackages),
+            1 == \count($downloadedPackages) ? '' : 's',
```
nicolas-grekas:
```suggestion
            1 === \count($downloadedPackages) ? '' : 's',
```

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -274,27 +219,18 @@ private function updateImportMapConfig(bool $update, array $packagesToRequire, a
         if ($update) {
             foreach ($currentEntries as $entry) {
                 $importName = $entry->importName;
-                if (null === $entry->url || (0 !== \count($packagesToUpdate) && !\in_array($importName, $packagesToUpdate, true))) {
+                if (!$entry->isRemotePackage() || (0 !== \count($packagesToUpdate) && !\in_array($importName, $packagesToUpdate, true))) {
```
nicolas-grekas:
```suggestion
                if (!$entry->isRemotePackage() || ($packagesToUpdate && !\in_array($importName, $packagesToUpdate, true))) {
```

src/Symfony/Component/AssetMapper/ImportMap/RemotePackageDownloader.php
```diff
@@ -0,0 +1,161 @@
...
+    /**
+     * Downloads all packages.
+     *
+     * @return string[] The downloaded packages
+     */
+    public function downloadPackages(callable $progressCallback = null): array
+    {
+        try {
+            $installed = $this->loadInstalled();
+        } catch (\LogicException) {
```
nicolas-grekas: InvalidArgumentException?

src/Symfony/Component/AssetMapper/ImportMap/RemotePackageDownloader.php
```diff
@@ -0,0 +1,161 @@
...
+                && file_exists($this->vendorDir.'/'.$installed[$entry->importName]['path'])
+            ) {
+                $newInstalled[$entry->importName] = $installed[$entry->importName];
+                continue;
+            }
+
+            $remoteEntriesToDownload[$entry->importName] = $entry;
+        }
+
+        if (0 === \count($remoteEntriesToDownload)) {
```
nicolas-grekas:
```suggestion
        if (!$remoteEntriesToDownload) {
```

src/Symfony/Component/AssetMapper/ImportMap/RemotePackageDownloader.php
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        $contents = $this->packageResolver->downloadPackages($remoteEntriesToDownload, $progressCallback);
+        $downloadedPackages = [];
+        foreach ($contents as $package => $content) {
+            if (!isset($remoteEntriesToDownload[$package])) {
+                throw new \LogicException(sprintf('The package "%s" was not in the requested packages.', $package));
+            }
+
+            $filename = $this->savePackage(
```
nicolas-grekas: call could fit in one line

src/Symfony/Component/AssetMapper/ImportMap/RemotePackageDownloader.php
```diff
@@ -0,0 +1,161 @@
...
+        foreach ($installed as $package => $data) {
+            if (!isset($data['path'])) {
+                throw new \LogicException(sprintf('The package "%s" is missing its path.', $package));
+            }
+
+            if (!isset($data['version'])) {
+                throw new \LogicException(sprintf('The package "%s" is missing its version.', $package));
+            }
+
+            if (!file_exists($this->vendorDir.'/'.$data['path'])) {
```
nicolas-grekas: we try to use is_file instead of file_exists as much as possible (it's faster)
```suggestion
            if (!is_file($this->vendorDir.'/'.$data['path'])) {
```

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -173,15 +204,46 @@ public function getPackageVersion(string $url): ?string
...
+     *
+     * Replaces those with normal import "package/name" statements.
+     */
+    private function makeImportsBare(string $content): string
+    {
+        $content = preg_replace_callback(self::IMPORT_REGEX, function ($matches) {
+            $packageName = $matches[1];
 
             return sprintf('from"%s"', $packageName);
         }, $content);
```
```suggestion
        $content = preg_replace_callback(self::IMPORT_REGEX, fn ($m) => sprintf('from"%s"', $m[1]), $content);
```

nicolas-grekas (commented):
I love it!

weaverryan:
Thanks for the check! Changes made

nicolas-grekas (approved):

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
```diff
@@ -48,10 +48,10 @@ public function audit(): array
         $installed = [];
         $affectsQuery = [];
         foreach ($entries as $entry) {
-            if (null === $entry->url) {
+            if (!$entry->isRemotePackage()) {
                 continue;
             }
-            $version = $entry->version ?? $this->packageResolver->getPackageVersion($entry->url);
+            $version = $entry->version;
```
smnandre, line 52:
```
public function __construct(
        private readonly ImportMapConfigReader $configReader,
        private readonly PackageResolverInterface $packageResolver,
        HttpClientInterface $httpClient = null,
) {
```

The $packageResolver property is now unused

src/Symfony/Component/AssetMapper/Command/ImportMapRequireCommand.php
```diff
@@ -125,19 +123,6 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             $newPackage = $newPackages[0];
             $message = sprintf('Package "%s" added to importmap.php', $newPackage->importName);
 
-            if ($newPackage->isDownloaded && null !== $downloadedAsset = $this->assetMapper->getAsset($newPackage->path)) {
```
smnandre, line 128: AssetMapperInterface $assetMapper property is now unused

src/Symfony/Component/AssetMapper/ImportMap/RemotePackageDownloader.php
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        if (!$remoteEntriesToDownload) {
+            return [];
+        }
+
+        $contents = $this->packageResolver->downloadPackages($remoteEntriesToDownload, $progressCallback);
+        $downloadedPackages = [];
+        foreach ($contents as $package => $content) {
+            if (!isset($remoteEntriesToDownload[$package])) {
```
smnandre: I'd do the opposite. Iterate over $remoteEntriesToDownload to ensure every $entry has been downloaded. 

Or unset $remoteEntriesToDownload each iteration and throw after the loop if it still contains entries ?

src/Symfony/Component/AssetMapper/ImportMap/RemotePackageDownloader.php
```diff
@@ -0,0 +1,157 @@
...
+        if (isset($this->installed)) {
+            return $this->installed;
+        }
+
+        $installedPath = $this->vendorDir.'/installed.php';
+        if (!is_file($installedPath)) {
+            return [];
+        }
+
+        $installed = is_file($installedPath) ? (static fn () => include $installedPath)() : [];
```
smnandre, line 132: `is_file` is already checked in previous condition

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -103,16 +98,15 @@ public function resolvePackages(array $packagesToRequire): array
...
+            $resolvedPackages[$packageName] = new ResolvedImportMapPackage($options, $version, $type);
 
-            if ($options->download) {
-                $content = $this->parseJsDelivrImports($response->getContent(), $packagesToRequire, $options->download);
-            }
+            $packagesToRequire = array_merge($packagesToRequire, $this->fetchPackageRequirementsFromImports($response->getContent()));
 
             $packageName = trim($options->packageName, '/');
-            $resolvedPackages[$packageName] = new ResolvedImportMapPackage($options, $url, $content);
+            $resolvedPackages[$packageName] = new ResolvedImportMapPackage($options, $version, $type);
```
smnandre: Lines 108/109 duplicate lignes 101&106 :)

fabpot:
Thank you @weaverryan.

PR #51800, opened 2023-10-01
[DoctrineBridge] Pass `Request` to `EntityValueResolver`'s expression

### LINKED ISSUE #51694: [DoctrineBridge] Passing the request object to the EntityValueResolver's expression
### Description

I would like to suggest passing the request object to the expression evaluation of the `EntityValueResolver`. This would make it possible to use parameters of the request body which is useful in REST APIs.

In the resolver we only need to change the following line in the `findViaExpression` method:

```patch
-         $variables = array_merge($request->attributes->all(), ['repository' => $repository]);
+         $variables = array_merge($request->attributes->all(), [
+             'repository' => $repository,
+             'request' => $request,
+         ]); 
```

### Example

Now, one could use the expression like this:

```php
public function createAction(
    #[MapEntity(expr: 'repository.find(request.toArray()["userId"])')]
    User $user,
) {}
```

What do you think?

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #51694
| License       | MIT

Passes the `Request` object to `EntityValueResolver`'s expression.

### DIFF
--- a/src/Symfony/Bridge/Doctrine/ArgumentResolver/EntityValueResolver.php
+++ b/src/Symfony/Bridge/Doctrine/ArgumentResolver/EntityValueResolver.php
@@ -199,7 +199,10 @@ final class EntityValueResolver implements ValueResolverInterface
         }
 
         $repository = $manager->getRepository($options->class);
-        $variables = array_merge($request->attributes->all(), ['repository' => $repository]);
+        $variables = array_merge($request->attributes->all(), [
+            'repository' => $repository,
+            'request' => $request,
+        ]);
 
         try {
             return $this->expressionLanguage->evaluate($options->expr, $variables);
--- a/src/Symfony/Bridge/Doctrine/CHANGELOG.md
+++ b/src/Symfony/Bridge/Doctrine/CHANGELOG.md
@@ -8,6 +8,7 @@ CHANGELOG
  * Deprecate not constructing `DoctrineDataCollector` with an instance of `DebugDataHolder`
  * Deprecate `DoctrineDataCollector::addLogger()`, use a `DebugDataHolder` instead
  * Deprecate `ContainerAwareLoader`, use dependency injection in your fixtures instead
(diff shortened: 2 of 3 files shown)


src/Symfony/Bridge/Doctrine/ArgumentResolver/EntityValueResolver.php
```diff
@@ -199,7 +199,10 @@ private function findViaExpression(ObjectManager $manager, Request $request, Map
         }
 
         $repository = $manager->getRepository($options->class);
-        $variables = array_merge($request->attributes->all(), ['repository' => $repository]);
+        $variables = array_merge($request->attributes->all(), [
+            'repository' => $repository,
+            'request' => $request,
```
HypeMC, line 204: This could be a possible BC break if there's a request attribute with the same name that's not Symfony's request object, but I don't think that's likely to happen. Another solution would be:

```php
$variables['request'] ??= $request;
```

OskarStark:
Can you add a testcase or extend an existing one? 

From my POV this makes sense.

Let's see what others think before you check the tests. Thanks

HypeMC:
> Can you add a testcase or extend an existing one?
> 
> From my POV this makes sense.
> 
> Let's see what others think before you check the tests. Thanks

@OskarStark I've already extended an [exiting one](https://github.com/symfony/symfony/pull/51800/files#diff-92ed254b959bf4b4dc3efbd3dc399df97dec5bae56631a07d00aafea62d0ebc5R337), not sure what else I can test here, the request is either present or not.

OskarStark:
Oh sorry, didn't realized it. All good then. 😃

OskarStark (approved):

fabpot (approved):

fabpot:
Thank you @HypeMC.

PR #51801, opened 2023-10-02
[Messenger] Fix WrappedExceptionsTrait

chalasr:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT

Fixes some minor flaws introduced in #51653

### DIFF
--- a/src/Symfony/Component/Messenger/Exception/WrappedExceptionsTrait.php
+++ b/src/Symfony/Component/Messenger/Exception/WrappedExceptionsTrait.php
@@ -18,6 +18,8 @@ namespace Symfony\Component\Messenger\Exception;
  */
 trait WrappedExceptionsTrait
 {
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/Messenger/Exception/WrappedExceptionsTrait.php
```diff
@@ -18,6 +18,8 @@
  */
 trait WrappedExceptionsTrait
 {
+    private array $exceptions;
```
chalasr, line 21: instead of assuming that the property exists on use-ing classes

fabpot (approved):

fabpot:
Thank you @chalasr.

PR #51804, opened 2023-10-02
[Security] Make `impersonation_path()` argument mandatory and add `impersonation_url()`

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT

Follow-up of https://github.com/symfony/symfony/pull/50030

When documenting this function, I found out that the`identifier` argument was optional, which seemed weird to me given the function purpose.

I then had a look at the implementation, and I saw that `ImpersonateUrlGenerator::generateImpersonationPath()` accepts a nullable string. However, the underlying call to `ImpersonateUrlGenerator::buildPath()` doesn't accept a nullable string. I propose to make the `identifier` argument mandatory, which makes more sense here.

I also added the missing Changelog line and `impersonation_url()`

### DIFF
--- a/src/Symfony/Bridge/Twig/CHANGELOG.md
+++ b/src/Symfony/Bridge/Twig/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
  * Allow an array to be passed as the first argument to the `importmap()` Twig function
  * Add `TemplatedEmail::locale()` to set the locale for the email rendering
  * Add `AppVariable::getEnabledLocales()` to retrieve the enabled locales
+ * Add `impersonation_path()` and `impersonation_url()` Twig functions
 
 6.3
 ---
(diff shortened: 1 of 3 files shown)


stof:
Should we also add `impersonate_url()` to be consistent with `impersonate_exit_*` exposing both `path` and `url` functions in Twig ?

alexandre-daubois:
I think that's a good idea, I added the new function

stof (approved):

fabpot (approved):

fabpot:
Thank you @alexandre-daubois.

PhilETaylor:
Thank you for cleaning up my mess! Appreciate it.

Issue #49803, opened 2023-03-25
[Scheduler] events?


kbond:
### Description

I know when using the scheduler with messenger, you can use worker message events but I wonder if we should add scheduler-specific events? The stand-alone `Scheduler` could benefit.

### Example

I think it could be nice DX to define events when defining your schedule:

```php
public function getSchedule(): Schedule
{
    return (new Schedule())
        ->add(RecurringMessage::cron('* * * * *', new DoSomething()))
        ->before(function(PreRunEvent $event) {})
        ->after(function(PostRunEvent $event) {})
        ->onFailure(function(FailureEvent $event) {})

        // maybe some out of the box helpers
        ->emailOnFailure(...) // uses symfony/mailer
        ->notifyOnFailure(...) // uses symfony/notifier
        ->pingBefore('https://some-uptime-service.com/oihjnjf8sd')
        ->pingAfter('https://some-uptime-service.com/jdhdud832hd')
        ->pingOnFailure('https://some-uptime-service.com/8978df7d')
    ;
}
```

carsonbot:
Thank you for this suggestion.
There has not been a lot of activity here for a while. Would you still like to see this feature?

kbond:
Closing as I no longer see a need for this now that the `ScheduledStamp` has more detail.

If anyone has a specific use-case that can't be solved with standard messenger events (ping @alli83), feel free to re-open.

alli83:
> Closing as I no longer see a need for this now that the `ScheduledStamp` has more detail.
> 
> If anyone has a specific use-case that can't be solved with standard messenger events (ping @alli83), feel free to re-open.

Thank you @kbond I'll reopen this issue and I'll ping you as soon as I open my PR :)

PR #51805, opened 2023-10-02
[Scheduler] pre_run and post_run events

alli83:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | https://github.com/symfony/symfony/issues/49803#event-10476636096
| License       | MIT

Based on https://github.com/symfony/symfony/issues/49803 @kbond  and taking into account https://github.com/symfony/symfony/pull/51553

The aim of this PR is to be able to act on the accumulated messages 'if something happens' and to be able to recalculate the heap via events (currently pre_run and post_run). 
The aim is to have access to 
- the  the schedule and therefore add/cancel a certain type of message.
- MessageContexte (e.g. access the id)
- The message itself

This PR would complement @Jeroeny https://github.com/symfony/symfony/pull/51553 PR by enabling action via events.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/scheduler.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/scheduler.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Component\DependencyInjection\Loader\Configurator;
 
+use Symfony\Component\Scheduler\EventListener\DispatchSchedulerEventListener;
 use Symfony\Component\Scheduler\Messenger\SchedulerTransportFactory;
 use Symfony\Component\Scheduler\Messenger\ServiceCallMessageHandler;
 
@@ -27,5 +28,11 @@ return static function (ContainerConfigurator $container) {
                 service('clock'),
             ])
             ->tag('messenger.transport_factory')
+        ->set('scheduler.event_listener', DispatchSchedulerEventListener::class)
+            ->args([
+                tagged_locator('scheduler.schedule_provider', 'name'),
+                service('event_dispatcher'),
+            ])
+            ->tag('kernel.event_subscriber')
     ;
 };
--- a/src/Symfony/Component/Scheduler/Event/PostRunEvent.php
+++ b/src/Symfony/Component/Scheduler/Event/PostRunEvent.php
@@ -1,0 +1,40 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Scheduler\Event;
+
+use Symfony\Component\Scheduler\Generator\MessageContext;
+use Symfony\Component\Scheduler\ScheduleProviderInterface;
+
+class PostRunEvent
+{
+    public function __construct(
+        private readonly ScheduleProviderInterface $schedule,
+        private readonly MessageContext $messageContext,
+        private readonly object $message,
+    ) {
+    }
+
+    public function getMessageContext(): MessageContext
+    {
+        return $this->messageContext;
+    }
+
+    public function getSchedule(): ScheduleProviderInterface
+    {
+        return $this->schedule;
+    }
+
+    public function getMessage(): object
+    {
+        return $this->message;
+    }
+}
--- a/src/Symfony/Component/Scheduler/Event/PreRunEvent.php
+++ b/src/Symfony/Component/Scheduler/Event/PreRunEvent.php
@@ -1,0 +1,51 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Scheduler\Event;
+
+use Symfony\Component\Scheduler\Generator\MessageContext;
+use Symfony\Component\Scheduler\ScheduleProviderInterface;
+
+class PreRunEvent
+{
+    private bool $shouldCancel = false;
+
+    public function __construct(
+        private readonly ScheduleProviderInterface $schedule,
+        private readonly MessageContext $messageContext,
+        private readonly object $message,
+    ) {
+    }
+
+    public function getMessageContext(): MessageContext
+    {
+        return $this->messageContext;
+    }
+
+    public function getSchedule(): ScheduleProviderInterface
+    {
+        return $this->schedule;
+    }
+
+    public function getMessage(): object
+    {
+        return $this->message;
+    }
+
+    public function shouldCancel(bool $shouldCancel = null): bool
+    {
+        if (null !== $shouldCancel) {
+            $this->shouldCancel = $shouldCancel;
+        }
+
+        return $this->shouldCancel;
+    }
+}
--- a/src/Symfony/Component/Scheduler/EventListener/DispatchSchedulerEventListener.php
+++ b/src/Symfony/Component/Scheduler/EventListener/DispatchSchedulerEventListener.php
@@ -1,0 +1,73 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Scheduler\EventListener;
+
+use Psr\Container\ContainerInterface;
+use Psr\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Messenger\Event\WorkerMessageHandledEvent;
+use Symfony\Component\Messenger\Event\WorkerMessageReceivedEvent;
+use Symfony\Component\Scheduler\Event\PostRunEvent;
+use Symfony\Component\Scheduler\Event\PreRunEvent;
+use Symfony\Component\Scheduler\Messenger\ScheduledStamp;
+
+class DispatchSchedulerEventListener implements EventSubscriberInterface
+{
+    public function __construct(
+        private readonly ContainerInterface $scheduleProviderLocator,
+        private readonly EventDispatcherInterface $eventDispatcher,
+    ) {
+    }
+
+    public function onMessageHandled(WorkerMessageHandledEvent $event): void
+    {
+        $envelope = $event->getEnvelope();
+        if (!$scheduledStamp = $envelope->last(ScheduledStamp::class)) {
+            return;
+        }
+
+        if (!$this->scheduleProviderLocator->has($scheduledStamp->messageContext->name)) {
+            return;
+        }
+
+        $this->eventDispatcher->dispatch(new PostRunEvent($this->scheduleProviderLocator->get($scheduledStamp->messageContext->name), $scheduledStamp->messageContext, $envelope->getMessage()));
+    }
+
+    public function onMessageReceived(WorkerMessageReceivedEvent $event): void
+    {
+        $envelope = $event->getEnvelope();
+
+        if (!$scheduledStamp = $envelope->last(ScheduledStamp::class)) {
+            return;
+        }
+
+        if (!$this->scheduleProviderLocator->has($scheduledStamp->messageContext->name)) {
+            return;
+        }
+
+        $preRunEvent = new PreRunEvent($this->scheduleProviderLocator->get($scheduledStamp->messageContext->name), $scheduledStamp->messageContext, $envelope->getMessage());
+
+        $this->eventDispatcher->dispatch($preRunEvent);
+
+        if ($preRunEvent->shouldCancel()) {
+            $event->shouldHandle(false);
+        }
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            WorkerMessageReceivedEvent::class => ['onMessageReceived'],
+            WorkerMessageHandledEvent::class => ['onMessageHandled'],
+        ];
+    }
+}
--- a/src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
+++ b/src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
@@ -93,6 +93,11 @@ final class MessageGenerator implements MessageGeneratorInterface
         $checkpoint->release($now, $this->waitUntil);
     }
 
+    public function getSchedule(): Schedule
+    {
+        return $this->schedule ??= $this->scheduleProvider->getSchedule();
+    }
+
     private function heap(\DateTimeImmutable $time, \DateTimeImmutable $startTime): TriggerHeap
     {
         if (isset($this->triggerHeap) && $this->triggerHeap->time <= $time) {
@@ -101,7 +106,7 @@ final class MessageGenerator implements MessageGeneratorInterface
 
         $heap = new TriggerHeap($time);
 
-        foreach ($this->schedule()->getRecurringMessages() as $index => $recurringMessage) {
+        foreach ($this->getSchedule()->getRecurringMessages() as $index => $recurringMessage) {
             $trigger = $recurringMessage->getTrigger();
 
             if ($trigger instanceof StatefulTriggerInterface) {
@@ -118,13 +123,8 @@ final class MessageGenerator implements MessageGeneratorInterface
         return $this->triggerHeap = $heap;
     }
 
-    private function schedule(): Schedule
-    {
-        return $this->schedule ??= $this->scheduleProvider->getSchedule();
-    }
-
     private function checkpoint(): Checkpoint
     {
-        return $this->checkpoint ??= new Checkpoint('scheduler_checkpoint_'.$this->name, $this->schedule()->getLock(), $this->schedule()->getState());
+        return $this->checkpoint ??= new Checkpoint('scheduler_checkpoint_'.$this->name, $this->getSchedule()->getLock(), $this->getSchedule()->getState());
     }
 }
--- a/src/Symfony/Component/Scheduler/Schedule.php
+++ b/src/Symfony/Component/Scheduler/Schedule.php
@@ -11,21 +11,29 @@
 
 namespace Symfony\Component\Scheduler;
 
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\Lock\LockInterface;
+use Symfony\Component\Scheduler\Event\PostRunEvent;
+use Symfony\Component\Scheduler\Event\PreRunEvent;
 use Symfony\Component\Scheduler\Exception\LogicException;
 use Symfony\Contracts\Cache\CacheInterface;
 
 final class Schedule implements ScheduleProviderInterface
 {
+    public function __construct(
+        private readonly ?EventDispatcherInterface $dispatcher = null,
+    ) {
+    }
+
     /** @var array<string,RecurringMessage> */
     private array $messages = [];
     private ?LockInterface $lock = null;
     private ?CacheInterface $state = null;
     private bool $shouldRestart = false;
 
-    public static function with(RecurringMessage $message, RecurringMessage ...$messages): static
+    public function with(RecurringMessage $message, RecurringMessage ...$messages): static
     {
-        return static::doAdd(new self(), $message, ...$messages);
+        return static::doAdd(new self($this->dispatcher), $message, ...$messages);
     }
 
     /**
@@ -57,6 +65,17 @@ final class Schedule implements ScheduleProviderInterface
     public function remove(RecurringMessage $message): static
     {
         unset($this->messages[$message->getId()]);
+        $this->setRestart(true);
+
+        return $this;
+    }
+
+    /**
+     * @return $this
+     */
+    public function removeById(string $id): static
+    {
+        unset($this->messages[$id]);
         $this->setRestart(true);
 
         return $this;
@@ -116,6 +135,20 @@ final class Schedule implements ScheduleProviderInterface
      */
     public function getSchedule(): static
     {
+        return $this;
+    }
+
+    public function before(callable $listener, int $priority = 0): static
+    {
+        $this->dispatcher->addListener(PreRunEvent::class, $listener, $priority);
+
+        return $this;
+    }
+
+    public function after(callable $listener, int $priority = 0): static
+    {
+        $this->dispatcher->addListener(PostRunEvent::class, $listener, $priority);
+
         return $this;
     }
 
--- a/src/Symfony/Component/Scheduler/Scheduler.php
+++ b/src/Symfony/Component/Scheduler/Scheduler.php
@@ -11,8 +11,11 @@
 
 namespace Symfony\Component\Scheduler;
 
+use Psr\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\Clock\Clock;
 use Symfony\Component\Clock\ClockInterface;
+use Symfony\Component\Scheduler\Event\PostRunEvent;
+use Symfony\Component\Scheduler\Event\PreRunEvent;
 use Symfony\Component\Scheduler\Generator\MessageGenerator;
 
 final class Scheduler
@@ -31,6 +34,7 @@ final class Scheduler
         private readonly array $handlers,
         array $schedules,
         private readonly ClockInterface $clock = new Clock(),
+        private readonly ?EventDispatcherInterface $dispatcher = null,
     ) {
         foreach ($schedules as $schedule) {
             $this->addSchedule($schedule);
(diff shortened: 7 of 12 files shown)


src/Symfony/Component/Scheduler/Event/SchedulerEvents.php
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Scheduler\Event;
+
+final class SchedulerEvents
```
kbond: I'm pretty sure we don't want this class alias system anymore. We just the use event's FQCN.

src/Symfony/Component/Messenger/Worker.php
```diff
@@ -99,6 +102,10 @@ public function run(array $options = []): void
                 if ($queueNames) {
                     $envelopes = $receiver->getFromQueues($queueNames);
                 } else {
+                    if (method_exists($receiver, 'getMessageGenerator')) {
```
kbond: I don't think we'll want the messenger to be aware of the scheduler. Ideally, we want these events dispatched in the scheduler but can't currently think of how.

src/Symfony/Component/Scheduler/Event/PostRunEvent.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+namespace Symfony\Component\Scheduler\Event;
+
+use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Scheduler\Generator\MessageGenerator;
+
+class PostRunEvent
+{
+    public function __construct(private readonly MessageGenerator $messageGenerator, private readonly Envelope $envelope)
```
kbond: Assuming we dispatch these events in the scheduler itself, we cannot use `Envelope` here as it could be any object.

src/Symfony/Component/Scheduler/Schedule.php
```diff
@@ -11,21 +11,27 @@
...
 
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\Lock\LockInterface;
+use Symfony\Component\Scheduler\Event\SchedulerEvents;
 use Symfony\Component\Scheduler\Exception\LogicException;
 use Symfony\Contracts\Cache\CacheInterface;
 
 final class Schedule implements ScheduleProviderInterface
 {
+    public function __construct(private readonly EventDispatcherInterface $dispatcher)
```
kbond: This would need to be optional.

kbond (commented):
Quick initial review

src/Symfony/Component/Messenger/Worker.php
alli83: Indeed, as discussed, I will add  listeners that listens to the WorkerMessageReceivedEvent  and to the WorkerMessageHandledEvent and dispatch the corresponding scheduler event. Thank you for your feedback!

src/Symfony/Component/Scheduler/Event/PostRunEvent.php
alli83: Indeed the standalone usage

src/Symfony/Component/Messenger/EventListener/DispatchSchedulerEventListener.php
```diff
@@ -0,0 +1,86 @@
...
+
+        if (!$this->receiverLocator->has($event->getReceiverName())) {
+            return;
+        }
+
+        /** @var SchedulerTransport $schedulerTransport */
+        $schedulerTransport =  $this->receiverLocator->get($event->getReceiverName());
+        $schedule = $schedulerTransport->getMessageGenerator()->schedule();
+
+        $this->eventDispatcher?->dispatch(new PostRunEvent($schedule, $messageContext, $envelope->getMessage()));
```
alli83: It seems that I would need the same info for pre_run and post_run event. 
I will handle this code duplication.

src/Symfony/Component/Scheduler/Schedule.php
```diff
@@ -11,21 +11,28 @@
...
 use Symfony\Component\Lock\LockInterface;
+use Symfony\Component\Scheduler\Event\PostRunEvent;
+use Symfony\Component\Scheduler\Event\PreRunEvent;
 use Symfony\Component\Scheduler\Exception\LogicException;
 use Symfony\Contracts\Cache\CacheInterface;
 
 final class Schedule implements ScheduleProviderInterface
 {
+    public function __construct(private readonly ?EventDispatcherInterface $dispatcher = null)
+    {
```
fabpot:
```suggestion
    public function __construct(
        private readonly ?EventDispatcherInterface $dispatcher = null,
    ) {
```

src/Symfony/Component/Scheduler/Event/PreRunEvent.php
```diff
@@ -0,0 +1,53 @@
...
+
+use Symfony\Component\Scheduler\Generator\MessageContext;
+use Symfony\Component\Scheduler\ScheduleProviderInterface;
+
+class PreRunEvent
+{
+    private bool $isCancel = false;
+
+    public function __construct(private readonly ScheduleProviderInterface $schedule, private readonly MessageContext $messageContext, private readonly object $message)
+    {
```
fabpot:
```suggestion
    public function __construct(
        private readonly ScheduleProviderInterface $schedule,
        private readonly MessageContext $messageContext,
        private readonly object $message,
    ) {
```

src/Symfony/Component/Scheduler/Event/PostRunEvent.php
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Symfony\Component\Scheduler\Event;
+
+use Symfony\Component\Scheduler\Generator\MessageContext;
+use Symfony\Component\Scheduler\ScheduleProviderInterface;
+
+class PostRunEvent
+{
+    public function __construct(private readonly ScheduleProviderInterface $schedule, private readonly MessageContext $messageContext, private readonly object $message)
+    {
```
fabpot:
```suggestion
    public function __construct(
        private readonly ScheduleProviderInterface $schedule,
        private readonly MessageContext $messageContext,
        private readonly object $message,
    ) {
```

src/Symfony/Component/Messenger/EventListener/DispatchSchedulerEventListener.php
```diff
@@ -0,0 +1,86 @@
...
+use Symfony\Component\Scheduler\Event\PostRunEvent;
+use Symfony\Component\Scheduler\Event\PreRunEvent;
+use Symfony\Component\Scheduler\Messenger\ScheduledStamp;
+use Symfony\Component\Scheduler\Messenger\SchedulerTransport;
+
+class DispatchSchedulerEventListener implements EventSubscriberInterface
+{
+    public function __construct(
+        private readonly ContainerInterface $receiverLocator,
+        private readonly ?EventDispatcherInterface $eventDispatcher = null
```
fabpot:
```suggestion
        private readonly ?EventDispatcherInterface $eventDispatcher = null,
```

src/Symfony/Bundle/FrameworkBundle/Tests/Fixtures/Messenger/DummySchedule.php
```diff
@@ -3,6 +3,7 @@
 namespace Symfony\Bundle\FrameworkBundle\Tests\Fixtures\Messenger;
 
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
+use Symfony\Component\EventDispatcher\EventDispatcher;
```
fabpot: Not used?

src/Symfony/Bundle/FrameworkBundle/Resources/config/messenger.php
```diff
@@ -179,6 +180,13 @@
             ->tag('kernel.event_subscriber')
             ->tag('monolog.logger', ['channel' => 'messenger'])
 
+        ->set('messenger.dispatch.scheduler_event_listener', DispatchSchedulerEventListener::class)
+            ->args([
+                service('messenger.receiver_locator'),
+                service('event_dispatcher'),
+        ])
```
fabpot:
```suggestion
            ])
```

src/Symfony/Component/Scheduler/Event/PreRunEvent.php
```diff
@@ -0,0 +1,53 @@
...
+        return $this->schedule;
+    }
+
+    public function getMessage(): object
+    {
+        return $this->message;
+    }
+
+    public function shouldCancel(bool $shouldCancel = null): bool
```
fabpot, line 43: I'm not sure we have a lot of similar methods in the code base where one method is used for both the setter and the getter.

src/Symfony/Component/Scheduler/Event/PreRunEvent.php
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+namespace Symfony\Component\Scheduler\Event;
+
+use Symfony\Component\Scheduler\Generator\MessageContext;
+use Symfony\Component\Scheduler\ScheduleProviderInterface;
+
+class PreRunEvent
+{
+    private bool $isCancel = false;
```
fabpot:
```suggestion
    private bool $isCancelled = false;
```

fabpot (changes requested):
Just some CS stuff for now.

src/Symfony/Component/Scheduler/Event/PreRunEvent.php
alli83: yes, sorry. Everything is under `hasBeenCancelled()` now

alli83: it's a duplicate of `shouldHandle()` of `WorkerMessageReceivedEvent` needed in the `PreRunEvent`

alli83:
To access the Schedule, it has been decided to inject the transports via the locator service (`messenger.receiver_locator`). While it would have been possible to inject the schedules directly via the tagged locator `scheduler.schedule_provider` doing so would require memoizing the created schedule to avoid calling getSchedule() twice.

src/Symfony/Component/Scheduler/Scheduler.php
```diff
@@ -62,9 +66,17 @@ public function run(array $options = []): void
 
             $ran = false;
             foreach ($this->generators as $generator) {
-                foreach ($generator->getMessages() as $message) {
-                    $this->handlers[$message::class]($message);
+                foreach ($generator->getMessages() as $context => $message) {
+                    $preRunEvent = new PreRunEvent($generator->schedule(), $context, $message);
```
dunglas: Nit: maybe could we avoid creating this event if `$this->dispatcher` is null?

src/Symfony/Component/Messenger/EventListener/DispatchSchedulerEventListener.php
```diff
@@ -0,0 +1,86 @@
...
+use Symfony\Component\Scheduler\Event\PostRunEvent;
+use Symfony\Component\Scheduler\Event\PreRunEvent;
+use Symfony\Component\Scheduler\Messenger\ScheduledStamp;
+use Symfony\Component\Scheduler\Messenger\SchedulerTransport;
+
+class DispatchSchedulerEventListener implements EventSubscriberInterface
+{
+    public function __construct(
+        private readonly ContainerInterface $receiverLocator,
+        private readonly ?EventDispatcherInterface $eventDispatcher = null,
```
dunglas: I'm probably missing something, but what is the purpose of this class if no dispatcher is provided? Shouldn't this argument be mandatory, and this service removed or not registered at all if no dispatcher is available?

src/Symfony/Component/Messenger/EventListener/DispatchSchedulerEventListener.php
```diff
@@ -0,0 +1,86 @@
...
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Messenger\EventListener;
```
fabpot: This class must be moved to the Scheduler component.

src/Symfony/Bundle/FrameworkBundle/Resources/config/messenger.php
```diff
@@ -179,6 +180,13 @@
             ->tag('kernel.event_subscriber')
             ->tag('monolog.logger', ['channel' => 'messenger'])
 
+        ->set('messenger.dispatch.scheduler_event_listener', DispatchSchedulerEventListener::class)
```
fabpot: To be moved to the scheduler configuration.

fabpot (changes requested):

src/Symfony/Bundle/FrameworkBundle/Resources/config/messenger.php
fabpot:
```suggestion
        ->set('scheduler.event_listener', DispatchSchedulerEventListener::class)
```

src/Symfony/Component/Scheduler/Scheduler.php
alli83: I will update it!

src/Symfony/Bundle/FrameworkBundle/Resources/config/messenger.php
alli83: ok i will update it!

src/Symfony/Component/Scheduler/Event/PostRunEvent.php
```diff
@@ -0,0 +1,40 @@
...
+
+use Symfony\Component\Scheduler\Generator\MessageContext;
+use Symfony\Component\Scheduler\ScheduleProviderInterface;
+
+class PostRunEvent
+{
+    public function __construct(
+        private readonly ScheduleProviderInterface $schedule,
+        private readonly MessageContext $messageContext,
+        private readonly object $message
```
fabpot:
```suggestion
        private readonly object $message,
```

src/Symfony/Component/Scheduler/Event/PreRunEvent.php
```diff
@@ -0,0 +1,51 @@
...
+use Symfony\Component\Scheduler\ScheduleProviderInterface;
+
+class PreRunEvent
+{
+    private bool $hasBeenCancelled = false;
+
+    public function __construct(
+        private readonly ScheduleProviderInterface $schedule,
+        private readonly MessageContext $messageContext,
+        private readonly object $message
```
fabpot:
```suggestion
        private readonly object $message,
```

src/Symfony/Component/Scheduler/Event/PreRunEvent.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        return $this->schedule;
+    }
+
+    public function getMessage(): object
+    {
+        return $this->message;
+    }
+
+    public function hasBeenCancelled(bool $hasBeenCancelled = null): bool
```
fabpot: I still think this is confusing. Let's revert to `shouldCancel`.

src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
```diff
@@ -93,6 +93,11 @@ public function getMessages(): \Generator
         $checkpoint->release($now, $this->waitUntil);
     }
 
+    public function schedule(): Schedule
```
fabpot: As a public method, it should be `getSchedule()`.

src/Symfony/Component/Scheduler/EventListener/DispatchSchedulerEventListener.php
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public function onMessageHandled(WorkerMessageHandledEvent $event): void
+    {
+        $envelope = $event->getEnvelope();
+        if (!$scheduledStamp = $envelope->last(ScheduledStamp::class)) {
+            return;
+        }
+
+        $messageContext = $scheduledStamp->messageContext;
```
fabpot: Let's inline, it's used only once.

src/Symfony/Component/Scheduler/EventListener/DispatchSchedulerEventListener.php
```diff
@@ -0,0 +1,86 @@
...
+
+    public function onMessageReceived(WorkerMessageReceivedEvent $event): void
+    {
+        $envelope = $event->getEnvelope();
+
+        if (!$scheduledStamp = $envelope->last(ScheduledStamp::class)) {
+            return;
+        }
+
+        $messageContext = $scheduledStamp->messageContext;
```
fabpot: To be inlined.

src/Symfony/Component/Scheduler/Generator/MessageGeneratorInterface.php
```diff
@@ -11,10 +11,14 @@
...
+use Symfony\Component\Scheduler\Schedule;
+
 interface MessageGeneratorInterface
 {
     /**
      * @return iterable<MessageContext, object>
      */
     public function getMessages(): iterable;
+
+    public function schedule(): Schedule;
```
fabpot:
```suggestion
    public function getSchedule(): Schedule;
```

src/Symfony/Component/Scheduler/Schedule.php
```diff
@@ -62,6 +70,17 @@ public function remove(RecurringMessage $message): static
         return $this;
     }
 
+    /**
+     * @return $this
+     */
+    public function removeById(string $id): static
```
fabpot, line 76: Not used?

src/Symfony/Component/Scheduler/Scheduler.php
```diff
@@ -31,6 +34,7 @@ public function __construct(
         private readonly array $handlers,
         array $schedules,
         private readonly ClockInterface $clock = new Clock(),
+        private readonly ?EventDispatcherInterface $dispatcher = null
```
fabpot:
```suggestion
        private readonly ?EventDispatcherInterface $dispatcher = null,
```

src/Symfony/Component/Scheduler/Scheduler.php
```diff
@@ -62,9 +66,17 @@ public function run(array $options = []): void
...
             $ran = false;
             foreach ($this->generators as $generator) {
-                foreach ($generator->getMessages() as $message) {
-                    $this->handlers[$message::class]($message);
+                foreach ($generator->getMessages() as $context => $message) {
+                    $preRunEvent = new PreRunEvent($generator->schedule(), $context, $message);
+                    $this->dispatcher?->dispatch($preRunEvent);
+
+                    if (!$preRunEvent->hasBeenCancelled()) {
+                        $this->handlers[$message::class]($message);
```
fabpot, line 71: missing `continue` here?

fabpot (changes requested):

src/Symfony/Component/Scheduler/Scheduler.php
alli83: I'll push an update because it can't depend on the event: the event is optional

src/Symfony/Component/Scheduler/Schedule.php
alli83: Not here but I thought it might be useful to be able to remove it directly by id (can be used easily in a listener that listens to the 'pre_run' event e.g)

src/Symfony/Component/Scheduler/Event/PreRunEvent.php
alli83: ok I will revert it

src/Symfony/Component/Scheduler/Generator/MessageGeneratorInterface.php
```diff
@@ -11,10 +11,14 @@
...
+use Symfony\Component\Scheduler\Schedule;
+
 interface MessageGeneratorInterface
 {
     /**
      * @return iterable<MessageContext, object>
      */
     public function getMessages(): iterable;
+
+    public function getSchedule(): Schedule;
```
fabpot: I'm uncomfortable adding this method to the interface as a MessageGenerator implementation might not use a Schedule.
In any case, I don't see a reason to interact with the schedule before or after handling a message; it doesn't seem right (especially as we now - in 6.4 - have other ways) and again a Schedule is not a requirement to send scheduled messages.
I would recommend removing the Schedule from the events.

fabpot (changes requested):

src/Symfony/Component/Scheduler/Generator/MessageGeneratorInterface.php
alli83: This is the case when a message is scheduled to be sent every x hours until an event occurs. 
It cannot be processed in the handler (messenger case), because it cannot be already processed at that time. The time at which its state (to be processed or not) can change lies within this x-hour interval. The aim, for example in pre_run, is to be able to check (before sending it to the handler) whether it is still relevant or whether, on the contrary, it should be removed from the stack of messages to be processed (since, for example, it will not have to be sent again if it has been processed in the meantime). The aim is to have access to the schedule and to be able to use the remove method, for example, which will recalculate the messages to be processed (using the setRestart introduced in 6.4). 

Alternatively, I could inject the ScheduleProviderInterface directly into the event ?

what I don't understand is that on the Scheduler class side, generators will have a Schedule via `ScheduleProviderInterface $scheduleProvider`.

Sorry if I misunderstood your point/comment

src/Symfony/Component/Messenger/Tests/EventListener/DispatchSchedulerEventListenerTest.php
```diff
@@ -0,0 +1,73 @@
...
+use Symfony\Component\Messenger\Event\WorkerMessageReceivedEvent;
+use Symfony\Component\Scheduler\Event\PreRunEvent;
+use Symfony\Component\Scheduler\EventListener\DispatchSchedulerEventListener;
+use Symfony\Component\Scheduler\Generator\MessageContext;
+use Symfony\Component\Scheduler\Messenger\ScheduledStamp;
+use Symfony\Component\Scheduler\RecurringMessage;
+use Symfony\Component\Scheduler\Tests\Messenger\SomeScheduleProvider;
+use Symfony\Component\Scheduler\Trigger\TriggerInterface;
+
+class DispatchSchedulerEventListenerTest extends TestCase
```
fabpot: Must be moved to the Scheduler component test suite.

src/Symfony/Bundle/FrameworkBundle/Resources/config/scheduler.php
```diff
@@ -27,5 +28,11 @@
                 service('clock'),
             ])
             ->tag('messenger.transport_factory')
+        ->set('scheduler.event_listener', DispatchSchedulerEventListener::class)
+            ->args([
+                tagged_locator('scheduler.schedule_provider', 'name'),
+                service('event_dispatcher'),
+        ])
```
fabpot:
```suggestion
            ])
```

src/Symfony/Bundle/FrameworkBundle/Resources/config/scheduler.php
```diff
@@ -27,5 +28,11 @@
                 service('clock'),
             ])
             ->tag('messenger.transport_factory')
+        ->set('scheduler.event_listener', DispatchSchedulerEventListener::class)
```
fabpot, line 31: I suppose we should bump the Scheduler min version to 6.4 in framework bundle?

src/Symfony/Component/Scheduler/Scheduler.php
```diff
@@ -62,8 +66,23 @@ public function run(array $options = []): void
 
             $ran = false;
             foreach ($this->generators as $generator) {
-                foreach ($generator->getMessages() as $message) {
+                foreach ($generator->getMessages() as $context => $message) {
+                    if ($this->dispatcher) {
```
fabpot: What about having testing `!$this->dispatcher` to have less code in the condition?

src/Symfony/Component/Scheduler/Schedule.php
```diff
@@ -11,21 +11,29 @@
...
 use Symfony\Component\Lock\LockInterface;
+use Symfony\Component\Scheduler\Event\PostRunEvent;
+use Symfony\Component\Scheduler\Event\PreRunEvent;
 use Symfony\Component\Scheduler\Exception\LogicException;
 use Symfony\Contracts\Cache\CacheInterface;
 
 final class Schedule implements ScheduleProviderInterface
 {
+    public function __construct(
+        private readonly ?EventDispatcherInterface $dispatcher = null
```
fabpot:
```suggestion
        private readonly ?EventDispatcherInterface $dispatcher = null,
```

fabpot (commented):
LGTM now (just a few minor comments).

dunglas (approved):
(when last comments from Fabien will be addressed)

src/Symfony/Bundle/FrameworkBundle/Resources/config/scheduler.php
fabpot: @alli83 You forgot this one apparently.

fabpot (approved):

fabpot:
Thank you @alli83.

PR #51808, opened 2023-10-02
[Mailer] Use idn encoded address otherwise Brevo throws an error

etlam:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | 
| License       | MIT

We have to use the encoded email address in case of special character domains like 'kältetechnik-xyz.de' otherwise brevo fails with 400 error.

### DIFF
--- a/src/Symfony/Component/Mailer/Bridge/Brevo/Tests/Transport/BrevoApiTransportTest.php
+++ b/src/Symfony/Component/Mailer/Bridge/Brevo/Tests/Transport/BrevoApiTransportTest.php
@@ -34,7 +34,7 @@ class BrevoApiTransportTest extends TestCase
         $this->assertSame($expected, (string) $transport);
     }
 
-    public static function getTransportData()
+    public static function getTransportData(): \Generator
     {
         yield [
             new BrevoApiTransport('ACCESS_KEY'),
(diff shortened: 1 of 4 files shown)


OskarStark:
Can you please add a testcase? Thank you

etlam:
@OskarStark Tests are added and fix is also contributed to the new Brevo transport.  
I can not reproduce this one failing unit test locally, what am I doing wrong and do I have to fix it?  

As far as I understand the test, it should not fail after doing changes to brevo or sendinblue.

OskarStark:
That's caused by another PR I guess.

cc @StaffNowa

StaffNowa:
> That's caused by another PR I guess.
> 
> cc @StaffNowa

@OskarStark I think it is not the same case as nullable transport.

@etlam as I see your PR it is about brevo API, but why do you edit another API? Is it the same company?

OskarStark:
@etlam you can ignore the failing test case.

> @OskarStark I think it is not the same case as nullable transport.

Yes, but AFAIR you had a PR regarding null transport, and this PR is failing, can you please have a look? Thanks

> @etlam as I see your PR it is about brevo API, but why do you edit another API? Is it the same company?

Yes it is!

etlam:
> @etlam as I see your PR it is about brevo API, but why do you edit another API? Is it the same company?

In 2023 Sendinblue changed its name to Brevo.

StaffNowa:
@OskarStark I am running tests now with https://github.com/symfony/symfony/pull/51830, and we will see if it helps or not )

fabpot (approved):

fabpot:
Thank you @etlam.

PR #51811, opened 2023-10-02
Add "dev" keyword to symfony/symfony package

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT

We're periodically wondering about how we should tell people that they should require individual symfony packages instead of the meta repo. Marking the symfony/symfony package as "abandonned" on packagist is very close on a technical level, but semantically it's inaccurate.

I thought we might want to declare symfony/symfony as "for dev". WDYT?

image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "symfony/symfony",
     "type": "library",
     "description": "The Symfony PHP framework",
-    "keywords": ["framework"],
+    "keywords": ["framework", "dev"],
     "homepage": "https://symfony.com",
     "license": "MIT",
     "authors": [
--- a/src/Symfony/Contracts/composer.json
+++ b/src/Symfony/Contracts/composer.json
@@ -2,7 +2,7 @@
     "name": "symfony/contracts",
     "type": "library",
     "description": "A set of abstractions extracted out of the Symfony components",
-    "keywords": ["abstractions", "contracts", "decoupling", "interfaces", "interoperability", "standards"],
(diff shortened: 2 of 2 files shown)


fabpot:
That's brilliant!

stof (approved):

stof:
@nicolas-grekas can you do the same for the polyfill mono-repo then ?

nicolas-grekas:
On it. Meanwhile I updated the PR for the contracts.

ro0NL:
i appreciate the effort to keep symfony/symfony available as a package, but what for exactly?

i assumed git clone+link would be the better practice for dev

wouterj:
In short:

- We can't remove the package, it will break applications on <3.4 beyond fixable 
- We can't keep the package, but no longer propagate new tags to Packagist
- There is a big "marketing value" in making releases on the mono repo
- Marking Symfony as abandoned gives the wrong signal, unless you are well into the community and are aware of the Flex initiative

ro0NL:
i was thinking `"abandoned": "Use individual symfony/* packages instead"`

then yes, signaling something to all existing installations could create some drama and/or unnessecary work. Perhaps it's not worth it.

OskarStark (approved):

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

PR #51812, opened 2023-10-02
[HttpFoundation] Fix type of properties in Request class

### LINKED ISSUE #51792: [HttpFoundation] unset `Request` properties within `initialize()` will cause issue on any extended Request implementing `__isset()` and `__get()`

7.0.0

### Description

Laravel Framework is preparing support for Symfony 7 for the upcoming Laravel 11 (target release Q1 2024) and based on our existing tests the usage such as `protected string $baseUrl` and using `unset($this->baseUrl)` will cause issues for `Illuminate\Http\Request` as this class extends `Symfony\Component\HttpFoundation\Request` and also implements `__isset()` and `__get()`.

### How to reproduce

The issue can be seen via our GitHub Actions: https://github.com/laravel/framework/actions/runs/6361912914/job/17276631512

To make it easier to understand, here is a simplified version of the usage:

```php
class SymfonyRequest 
{
  protected string $baseUrl;

  public function __construct() 
  {
    unset($this->baseUrl);
  }

  public function getPathInfo()
  {
    if (null === ($baseUrl = $this->getBaseUrlReal())) {
      return 'http://localhost';
    }
    
    assert($baseUrl === 'foo', 'BaseURL should be `foo` from `prepareBaseUrl()` but instead is `'.$baseUrl.'`');

    return 'http://laravel.com';
  }

  private function getBaseUrlReal(): string
  {

    return $this->baseUrl ??= $this->prepareBaseUrl();
  }

  protected function prepareBaseUrl(): string
  {
    return 'foo';
  }
}

class LaravelRequest extends SymfonyRequest
{
    public function path() {
      return $this->getPathInfo();
    }

    public function __isset($key)
    {
      return ! is_null($this->__get($key));
    }
    
    public function __get($key)
    {
      if ($key === 'baseUrl') {
        return 'bar';
      }

      return null;
    }
}

$a = new LaravelRequest();

$a->path();
```

CleanShot 2023-10-01 at 11 00 07

### Possible Solution

One possible solution is to opt for the following:

```diff
- protected string $baseUrl;
+protected ?string $baseUrl;
```

and set to null instead of unset.

```diff
-unset($this->baseUrl);
+$this->baseUrl = null;
```

### Additional Context

Without the above change, any extended Request class similar to Laravel would need to look into exposing Symfony Request via `__isset()` and `__get()` which might not be ideal.

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #51792
| License       | MIT

Partially reverts #51121 where I didn't consider the behavior change reported in the linked issue.

### DIFF
--- a/src/Symfony/Component/Config/Definition/Builder/TreeBuilder.php
+++ b/src/Symfony/Component/Config/Definition/Builder/TreeBuilder.php
@@ -20,7 +20,14 @@ use Symfony\Component\Config\Definition\NodeInterface;
  */
 class TreeBuilder implements NodeParentInterface
 {
+    /**
+     * @var NodeInterface|null
+     */
     protected $tree;
+
+    /**
+     * @var NodeDefinition
+     */
     protected $root;
 
     public function __construct(string $name, string $type = 'array', NodeBuilder $builder = null)
@@ -53,7 +60,7 @@ class TreeBuilder implements NodeParentInterface
     public function setPathSeparator(string $separator)
     {
         // unset last built as changing path separator changes all nodes
-        unset($this->tree);
+        $this->tree = null;
 
         $this->root->setPathSeparator($separator);
     }
--- a/src/Symfony/Component/HttpClient/Response/MockResponse.php
+++ b/src/Symfony/Component/HttpClient/Response/MockResponse.php
@@ -28,7 +28,7 @@ class MockResponse implements ResponseInterface, StreamableInterface
     use CommonResponseTrait;
     use TransportResponseTrait;
 
-    private string|iterable $body;
+    private string|iterable|null $body;
     private array $requestOptions = [];
     private string $requestUrl;
     private string $requestMethod;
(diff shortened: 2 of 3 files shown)


stof (approved):

stof:
@nicolas-grekas don't we have the same issue in TreeBuilder (and any other place using protected properties that are unset actually) ?

nicolas-grekas:
> don't we have the same issue in TreeBuilder (and any other place using protected properties that are unset actually) ?

Correct, and fixed. I didn't find any other occurrence but I did change something in HttpClient because it made more sense to make the property nullable.

stof (approved):

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

taylorotwell:
Thanks all 👍 much appreciated.

PR #50957, opened 2023-07-12
[Security] Remove unused code about ROLE_PREVIOUS_ADMIN

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

PR #51813, opened 2023-10-02
[Security] Deprecated ROLE_PREVIOUS_ADMIN

mfadul24:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | yes
| License       | MIT

The SwitchUserListener is still using the `ROLE_PREVIOUS_ADMIN` role. I've renamed it to `IS_IMPERSONATOR` that is the role since SF 6.0.


chalasr:
Thanks for submitting your first contribution to Symfony.
This code was necessary for the remember-me system to work before it was refactored. With the new implementation and the `IS_IMPERSONATOR` attribute, it is not needed anymore. 
Even if it should have been removed in 6.0 given the role has been deprecated in 5.x, I think it is safer to do it for the next major version. And it has been done already in https://github.com/symfony/symfony/pull/50957.
So unfortunately I think we can close this PR. 

/cc @wouterj in case you think otherwise

wouterj:
I agree. Also note that `IS_IMPERSONATOR` is an attribute, not a role, which is why #50957 didn't add it to the roles array but rather removes the deprecated `ROLE_PREVIOUS_ADMIN` role.

Issue #51814, opened 2023-10-02
[HttpFoundation] `Cookie::fromString()` swallows trailing `=` in value.


pschultz:
### Symfony version(s) affected

6.3.5

### Description

Cookie::fromString destroys trailing equal signs in the value. Trailing '=' are very common in base64 encoded values.

From a glance at the code the cause seems to be the call to `$parts = HeaderUtils::split($cookie, ';=');`, where `$cookie` is the string as it appears in an HTTP header. 

### How to reproduce

```php
<?php

require 'vendor/autoload.php';

use Symfony\Component\HttpFoundation\Cookie;

$c = Cookie::fromString('foo=AAA=; Path=/; Expires=Sun, 31 Dec 2023 14:32:53 GMT; Max-Age=7776000; HttpOnly; Secure; SameSite=None');

var_dump($c->getValue()); // string(3) "AAA", but expected string(4) "AAA="
```

### Possible Solution

Split only on ';' to determine the parts. Split each part into key-value pairs on the first '=' only.

OskarStark:
I'm on it 🔥 
I can confirm the bug with `5.4`

PR #51819, opened 2023-10-02
[HttpFoundation] Do not swallow trailing `=` in cookie value

OskarStark:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix https://github.com/symfony/symfony/issues/51814
| License       | MIT

cc @pschultz as you opened the bug report

### DIFF
--- a/src/Symfony/Component/HttpFoundation/HeaderUtils.php
+++ b/src/Symfony/Component/HttpFoundation/HeaderUtils.php
@@ -33,17 +33,21 @@ class HeaderUtils
      *
      * Example:
      *
-     *     HeaderUtils::split("da, en-gb;q=0.8", ",;")
+     *     HeaderUtils::split('da, en-gb;q=0.8', ',;')
      *     // => ['da'], ['en-gb', 'q=0.8']]
      *
      * @param string $separators List of characters to split on, ordered by
-     *                           precedence, e.g. ",", ";=", or ",;="
+     *                           precedence, e.g. ',', ';=', or ',;='
      *
      * @return array Nested array with as many levels as there are characters in
      *               $separators
      */
     public static function split(string $header, string $separators): array
     {
+        if ('' === $separators) {
+            throw new \InvalidArgumentException('At least one separator must be specified.');
+        }
+
         $quotedSeparators = preg_quote($separators, '/');
 
         preg_match_all('
@@ -77,8 +81,8 @@ class HeaderUtils
      *
      * Example:
      *
-     *     HeaderUtils::combine([["foo", "abc"], ["bar"]])
-     *     // => ["foo" => "abc", "bar" => true]
+     *     HeaderUtils::combine([['foo', 'abc'], ['bar']])
+     *     // => ['foo' => 'abc', 'bar' => true]
      */
     public static function combine(array $parts): array
     {
@@ -95,13 +99,13 @@ class HeaderUtils
     /**
      * Joins an associative array into a string for use in an HTTP header.
      *
-     * The key and value of each entry are joined with "=", and all entries
+     * The key and value of each entry are joined with '=', and all entries
      * are joined with the specified separator and an additional space (for
      * readability). Values are quoted if necessary.
      *
      * Example:
      *
-     *     HeaderUtils::toString(["foo" => "abc", "bar" => true, "baz" => "a b c"], ",")
+     *     HeaderUtils::toString(['foo' => 'abc', 'bar' => true, 'baz' => 'a b c'], ',')
      *     // => 'foo=abc, bar, baz="a b c"'
      */
     public static function toString(array $assoc, string $separator): string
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/HttpFoundation/Tests/CookieTest.php
```diff
@@ -324,6 +324,9 @@ public function testFromString()
         $cookie = Cookie::fromString('foo=bar', true);
         $this->assertEquals(Cookie::create('foo', 'bar', 0, '/', null, false, false, false, null), $cookie);
 
+        $cookie = Cookie::fromString('foo=bar=', true);
```
OskarStark, line 327: This test case was failing before

OskarStark:
Or should this be fixed inside HeaderUtils::split() method?

stof:
I guess we have similar bugs in other places using `HeaderUtils::split` with multiple separator (i.e. all of them). That's because HeaderUtils considers that all separators are list separators. But `=` is a key-value separator.

OskarStark:
So do you have a proposal on where to fix this?

stof:
Well, HeaderUtils should be the place to fix it, but not sure we can fix it with the existing signature (unless we hardcode a special behavior for `=`).

nicolas-grekas:
The bug is clearly in HeaderUtils, see
https://github.com/symfony/symfony/blob/adcd3d06d078c86618ec3b27892e474483c63895/src/Symfony/Component/HttpFoundation/Tests/HeaderUtilsTest.php#L48

`foo_cookie=foo==` is correctly parsed while `foo_cookie=foo=` isn't

OskarStark:
Will check on Monday 👍🏻

OskarStark:
Review please

src/Symfony/Component/HttpFoundation/HeaderUtils.php
```diff
@@ -257,10 +257,20 @@ private static function groupParts(array $matches, string $separators, bool $fir
         $i = 0;
         $partMatches = [];
         $previousMatchWasSeparator = false;
-        foreach ($matches as $match) {
+        foreach ($matches as $key => $match) {
             if (!$first && $previousMatchWasSeparator && isset($match['separator']) && $match['separator'] === $separator) {
                 $previousMatchWasSeparator = true;
                 $partMatches[$i][] = $match;
+            } elseif (!$first && !isset($match['separator']) && isset($matches[$key + 1], $matches[$key + 1]['separator']) && $matches[$key+1]['separator'] === $separator && !isset($matches[$key+2])) { // ends with a separator, so it must be part of the value
```
nicolas-grekas:
```suggestion
            } elseif (!$first && !isset($match['separator']) && ($matches[$key + 1]['separator'] ?? null) === $separator && !isset($matches[$key + 2])) { // ends with a separator, so it must be part of the value
```

src/Symfony/Component/HttpFoundation/HeaderUtils.php
```diff
@@ -257,10 +257,20 @@ private static function groupParts(array $matches, string $separators, bool $fir
...
+                $numberOfSeparators = 0;
+                for ($i = $key+1; $i < \count($matches); ++$i) {
```
nicolas-grekas: do we need this loop? didn't we rule out with `!isset($matches[$key+2])`?

nicolas-grekas (commented):
The whole logic is opaque to me :sweat:

src/Symfony/Component/HttpFoundation/HeaderUtils.php
OskarStark: Yes

src/Symfony/Component/HttpFoundation/HeaderUtils.php
OskarStark: Yes we rule out, but we don't know if one or 2 trailing `=`

I agree, its very opaque and I really don't like the whole code :-(

src/Symfony/Component/HttpFoundation/HeaderUtils.php
OskarStark: Works

nicolas-grekas (approved):
I rewrote the implementation, it now makes sense, at least to me :)
AND, it handles more edge cases.

OskarStark:
Thank you Nicolas!! 👏

src/Symfony/Component/HttpFoundation/Tests/HeaderUtilsTest.php
```diff
@@ -46,13 +46,15 @@ public static function provideHeaderToSplit(): array
 
             [[['foo_cookie', 'foo=1&bar=2&baz=3'], ['expires', 'Tue, 22-Sep-2020 06:27:09 GMT'], ['path', '/']], 'foo_cookie=foo=1&bar=2&baz=3; expires=Tue, 22-Sep-2020 06:27:09 GMT; path=/', ';='],
             [[['foo_cookie', 'foo=='], ['expires', 'Tue, 22-Sep-2020 06:27:09 GMT'], ['path', '/']], 'foo_cookie=foo==; expires=Tue, 22-Sep-2020 06:27:09 GMT; path=/', ';='],
```
OskarStark, line 48:
```suggestion
            [[['foo_cookie', 'foo==='], ['expires', 'Tue, 22-Sep-2020 06:27:09 GMT'], ['path', '/']], 'foo_cookie=foo===; expires=Tue, 22-Sep-2020 06:27:09 GMT; path=/', ';='],
            [[['foo_cookie', 'foo=='], ['expires', 'Tue, 22-Sep-2020 06:27:09 GMT'], ['path', '/']], 'foo_cookie=foo==; expires=Tue, 22-Sep-2020 06:27:09 GMT; path=/', ';='],
```

Does this pass as well?

nicolas-grekas: yes

nicolas-grekas:
Thank you @OskarStark.

OskarStark:
Thank you, as you did the reimplementation 👍

Issue #51823, opened 2023-10-03
[Serializer] Deserialization order of priority changed `#[SerializedPath]` vs. property name


jerowork:
### Symfony version(s) affected

6.3.5

### Description

With the update of Serializer to 6.3.5, some deserialization of array to objects does behave differently (changed order of priority of configuration via attribute `#[SerializedPath]` vs. property name, when there is a key on root level with the same name as the private property.

Related to https://github.com/symfony/symfony/pull/49700.

### How to reproduce

Example to explain changed behavior:

```json
{
  "data": {
    "item": {
      "id": "id-1"
    }
  },
  "id": "id-2"
}
```

```php
final class SomeEvent
{
    #[SerializedPath('[data][item][id]')]
    public string $id;
}
```

Before 6.3.5, the value of the id was `id-1`, with the change of https://github.com/symfony/symfony/pull/49700, the value of the id becomes `id-2`.

https://github.com/symfony/symfony/pull/49700 changes `array_merge` with `array + array`. It seems that the problem stated above is related to the fact that array_merge does overwrite keys differently than array + array:

```php
$a = ['key' => 'value-a'];
$b = ['key' => 'value-b'];

var_dump(array_merge($a, $b));
// Results in:
// array(1) {
//   ["key"]=>
//   string(7) "value-b"
// }

var_dump($a + $b);
// Results in:
// array(1) {
//   ["key"]=>
//   string(7) "value-a"
// }
```

### Possible Solution

As `array_merge` does behave slightly differently that array + array, the solution could be to switch array order to:
```diff
- $normalizedData = $normalizedData + $nestedData;
+ $normalizedData = $nestedData + $normalizedData;
```

This would result in the same, while keeping the fix (https://github.com/symfony/symfony/pull/49700) for the numeric key value.

PR: https://github.com/symfony/symfony/pull/51825

PR #49700, opened 2023-03-16
[Serializer] Fix reindex normalizedData array in AbstractObjectNormalizer::denormalize()

elevado:
| Q             | A
| ------------- | ---
| Branch       |  6.2
| Bug fix      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | #49538 
| License       | MIT
| Doc PR        | no

In the method `AbstractObjectNormalizer::denormalize()` the index of the array `$normalizedData` is reindexed after an `array_merge`.

This error occurs when a JSON is deserialised and when the SerializedName is numeric. This results in an incorrect mapping to the properties.

```json
{
   "1":  "John",
   "2": "Doe",
   "10031":  "john.doe@example.com",
}

```
Array before merge:

```php
array (
  1 => 'John',
  2 => 'Doe',
  10031 => 'john.doe@example.com',
)
```

After merge with `array_merge`:

```php
array (
  0 => 'John',
  1 => 'Doe',
  2 => 'john.doe@example.com',
)
```

After merge with `array_replace`:

```php
array (
  0 => 'John',
  1 => 'Doe',
  10031 => 'john.doe@example.com',
)
```

All Serializer unittests run successfully.

PR #51825, opened 2023-10-03
Fix order array sum normalizedData and nestedData

jerowork:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | https://github.com/symfony/symfony/issues/51823
| License       | MIT

### Description

With the update of Serializer to 6.3.5, some deserialization of array to objects does behave differently (changed order of priority of configuration via attribute `#[SerializedPath]` vs. property name, when there is a key on root level with the same name as the private property.

Related to https://github.com/symfony/symfony/pull/49700.

### How to reproduce

Example to explain changed behavior:

```json
{
  "data": {
    "item": {
      "id": "id-1"
    }
  },
  "id": "id-2"
}
```

```php
final class SomeEvent
{
    #[SerializedPath('[data][item][id]')]
    public string $id;
}
```

Before 6.3.5, the value of the id was `id-1`, with the change of https://github.com/symfony/symfony/pull/49700, the value of the id becomes `id-2`.

https://github.com/symfony/symfony/pull/49700 changes `array_merge` with `array + array`. It seems that the problem stated above is related to the fact that array_merge does overwrite keys differently than array + array:

```php
$a = ['key' => 'value-a'];
$b = ['key' => 'value-b'];

var_dump(array_merge($a, $b));
// Results in:
// array(1) {
//   ["key"]=>
//   string(7) "value-b"
// }

var_dump($a + $b);
// Results in:
// array(1) {
//   ["key"]=>
//   string(7) "value-a"
// }
```

### Solution

As `array_merge` does behave slightly differently that array + array, the solution could be to switch array order to:
```diff
- $normalizedData = $normalizedData + $nestedData;
+ $normalizedData = $nestedData + $normalizedData;
```

This would result in the same, while keeping the fix (https://github.com/symfony/symfony/pull/49700) for the numeric key value

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
@@ -339,7 +339,7 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
             $normalizedData = $this->removeNestedValue($serializedPath->getElements(), $normalizedData);
         }
 
-        $normalizedData = $normalizedData + $nestedData;
+        $normalizedData = $nestedData + $normalizedData;
 
         $object = $this->instantiateObject($normalizedData, $mappedClass, $context, new \ReflectionClass($mappedClass), $allowedAttributes, $format);
         $resolvedClass = ($this->objectClassResolver)($object);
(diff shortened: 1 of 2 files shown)


HypeMC:
I've encountered the same issue and can confirm that this fixes it without (at least in my case) introducing any additional problems.

OskarStark (approved):

fabpot (approved):

fabpot:
Thank you @jerowork.

PR #51828, opened 2023-10-03
[AssetMapper] Put importmap in polyfill so it can be hosted locally easily

### LINKED ISSUE #51302: [AssetMapper] Allow importmap_polyfill to be an AssetMapper path
### Description

Hi!

A user requested to be able to avoid using a CDN for the `importmap_polyfill`. This is configurable via `framework.asset_mapper.importmap_polyfill`. However, that path is not processed through AssetMapper, which is a bit awkward: you need to put it directly into your public directory.

Overall, it'd be nice to provide an easier path to using a local importmap polyfill.

Cheers!

### Example

_No response_

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #51302
| License       | MIT

Note: Built on top of #51786

In #51302, it was asked to allow the polyfill to be an asset mapper path. We could do that, but it would still require the user to "manage" this vendor file locally - i.e. commit it into their repository. So this PR goes a bit further and requires your polyfill to be an item in `importmap.php`. That's a bit odd, but because of #51786, it allows the JS package to be downloaded like any other JS package. The item is actually removed from the `importmap` before it's finally dumped. So the small oddity of having this item in `importmap.php` makes a really nice DX.

Cheers!


src/Symfony/Component/AssetMapper/Command/ImportMapInstallCommand.php
```diff
@@ -27,7 +29,8 @@
 final class ImportMapInstallCommand extends Command
 {
     public function __construct(
-        private readonly ImportMapManager $importMapManager,
+        private readonly RemotePackageDownloader $packageDownloader,
+        private readonly string $projectDir,
```
smnandre: Is this necessary "just" to render a success message ?

src/Symfony/Component/AssetMapper/Command/ImportMapRequireCommand.php
```diff
@@ -42,7 +41,7 @@ protected function configure(): void
     {
         $this
             ->addArgument('packages', InputArgument::IS_ARRAY | InputArgument::REQUIRED, 'The packages to add')
-            ->addOption('download', 'd', InputOption::VALUE_NONE, 'Download packages locally')
+            ->addOption('entrypoint', null, InputOption::VALUE_NONE, 'Make the package an entrypoint?')
```
smnandre: package**s** ?

src/Symfony/Component/AssetMapper/Command/ImportMapInstallCommand.php
weaverryan: It is... which is kind of lame, but otherwise, I'll be rendering the whole "Downloaded assets into /foo/bar/baz/my-project/assets/vendor"

src/Symfony/Component/AssetMapper/Tests/ImportMap/RemotePackageDownloaderTest.php
```diff
@@ -0,0 +1,177 @@
...
+
+class RemotePackageDownloaderTest extends TestCase
+{
+    private Filesystem $filesystem;
+    private static string $writableRoot = __DIR__.'/../fixtures/importmaps_for_writing';
+
+    protected function setUp(): void
+    {
+        $this->filesystem = new Filesystem();
+        if (!file_exists(__DIR__.'/../fixtures/importmaps_for_writing')) {
```
smnandre: self::writableRoot ?

weaverryan:
~~I think this is needed, but one negative I thought of. We will likely need to start the `importmap.php` recipe with the polyfill inside of it. And since vendor files are all now always downloaded, the user will need to run `importmap:install` before the page will run.~~

~~We could... just for the polyfill, if not installed, revert to the CDN URL instead? Or just not render it? It seems iffy. However, missing the polyfill locally would not be an issue (we all have modern browsers). And it is HIGHLY unlikely that someone deploys with NO other vendor libraries installed in their app. And so, they would see 404 errors anyway / their `importmap:compile` command would explode due forgetting `importmap:install`.~~

UPDATE

All good now - here is the flow:

A) If the polyfill is set to `es-module-shims` (which is the default in FrameworkBundle) BUT it is not in your importmap, then it outputs the CDN URL. This will be the default behavior.

B) If the user chooses to add `es-module-shims` to their importmap (`importmap:require es-module-shims`) the system will instantly start pointing to that local file.

So the default behavior is basically identical to now, but with the option to add `es-module-shims` to the `importmap.php` to serve a local file. The user can, of course, still entirely disable the importmap and avoid it entirely, or do whatever they want manually.

fabpot (approved):

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
```diff
@@ -27,11 +27,13 @@
  */
 class ImportMapRenderer
 {
+    private const DEFAULT_ES_MODULE_SHIMS_POLYFILL_URL = 'https://ga.jspm.io/npm:es-module-shims@1.8.0/dist/es-module-shims.js';
```
smnandre, line 30: Do we have another option ? Maybe in the recipe ? 
This package seem to release versions quite often.. https://www.npmjs.com/package/es-module-shims?activeTab=versions

weaverryan: The thinking is that:

A) By default, the user does nothing and has no config and we give them this CDN URL. We can bump this URL when needed in core (other than if they have a major new version - but we can cross that bridge then).

B) This string itself doesn't need to be configurable because, if you care enough to fix it at a specific version, you'd add it to your importmap.

So I don't want to make this string configurable or put it in the recipe because then there would be 2 ways to control it - i.e. you'd be able to configure the fallback CDN URL and also take full control over it by putting it in your importmap.

smnandre: That make sense... maybe then just create a "monthly" reminder to check the new versions :)

weaverryan: I think people will make PR's if they care enough. There's not a lot happening with that polyfill at this point, as it's quite feature complete.

fabpot:
Thank you @weaverryan.

PR #51829, opened 2023-10-03
[AssetMapper] Automatically preload CSS files if WebLink available

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | None
| License       | MIT

Hi!

In AssetMapper 6.4, we now know which CSS files will be rendered on the page. So, I think there is now downside to adding a `Link` header in the response to load those CSS files, but someone tell me if I'm missing something.

The only possible situation I can think of where this isn't wanted is if you decided to handle your critical CSS manually (i.e. with normal `<link rel="stylesheet">` tags in `base.html.twig`) and load a few less-critical CSS files through the ImportMap system. In that case, we would preload only these "less-critical" CSS, which could take a slight priority over the others. But I think the benefit this would give to most apps outweighs this. We could always add an opt-out later.

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/CHANGELOG.md
+++ b/src/Symfony/Component/AssetMapper/CHANGELOG.md
@@ -9,6 +9,7 @@ CHANGELOG
  * Add "entrypoints" concept to the importmap
  * Always download packages locally instead of using a CDN
  * Allow relative path strings in the importmap
+ * Automatically set `_links` attribute for preload CSS files for WebLink integration
  * Add `PreAssetsCompileEvent` event when running `asset-map:compile`
  * Add support for importmap paths to use the Asset component (for subdirectories)
  * Removed the `importmap:export` command
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
@@ -11,7 +11,13 @@
 
 namespace Symfony\Component\AssetMapper\ImportMap;
 
+use Psr\Link\EvolvableLinkProviderInterface;
 use Symfony\Component\Asset\Packages;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\WebLink\EventListener\AddLinkHeaderListener;
+use Symfony\Component\WebLink\GenericLinkProvider;
+use Symfony\Component\WebLink\Link;
 
 /**
  * @author Kévin Dunglas <kevin@dunglas.dev>
@@ -27,6 +33,7 @@ class ImportMapRenderer
         private readonly string $charset = 'UTF-8',
         private readonly string|false $polyfillUrl = ImportMapManager::POLYFILL_URL,
         private readonly array $scriptAttributes = [],
+        private readonly ?RequestStack $requestStack = null,
     ) {
     }
 
@@ -66,6 +73,10 @@ class ImportMapRenderer
             $url = $this->escapeAttributeValue($url);
 
             $output .= "\n<link rel=\"stylesheet\" href=\"$url\">";
+        }
+
+        if (class_exists(AddLinkHeaderListener::class) && $request = $this->requestStack?->getCurrentRequest()) {
+            $this->addWebLinkPreloads($request, $cssLinks);
         }
 
         $scriptAttributes = $this->createAttributesString($attributes);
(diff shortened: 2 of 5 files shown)


smnandre:
The only problem would be: Google Pagespeed and other performances tests/tools rate really bad if you preload a CSS you do not use "quickly"

```
Modern browsers are already pretty good at prioritizing resources, that's why it's important to use preload sparingly and only preload the most critical resources.
```

https://web.dev/preload-critical-assets/#how-preloading-works

weaverryan:
> The only problem would be: Google Pagespeed and other performances tests/tools rate really bad if you preload a CSS you do not use "quickly"

But if we KNOW that what we are preloading WILL be `<link rel="stylesheet">` inside the initial HTML, then we would be good, right?

smnandre:
> > The only problem would be: Google Pagespeed and other performances tests/tools rate really bad if you preload a CSS you do not use "quickly"
> 
> But if we KNOW that what we are preloading WILL be `<link rel="stylesheet">` inside the initial HTML, then we would be good, right?

I'll really need to try it "in the wild" (ux.symfony.com ? 👼)  to be 100% sure... but i'd say "yes" :) 

And anyone so obsessed with performances still can call manually the stylesheets so i'm not worried :)

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
```diff
@@ -27,6 +33,7 @@ public function __construct(
         private readonly string $charset = 'UTF-8',
         private readonly string|false $polyfillUrl = ImportMapManager::POLYFILL_URL,
         private readonly array $scriptAttributes = [],
+        private readonly ?RequestStack $requestStack = null,
```
weaverryan, line 36: Injecting the request stack here feels a bit odd, but it's the best way I can think of to handle this.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
```diff
@@ -68,6 +75,10 @@ public function render(string|array $entryPoint, array $attributes = []): string
             $output .= "\n<link rel=\"stylesheet\" href=\"$url\">";
         }
 
+        if (class_exists(AddLinkHeaderListener::class) && $request = $this->requestStack?->getCurrentRequest()) {
```
weaverryan, line 78: Same logic as in `AbstractController`

weaverryan:
high-deps for asset mapper failing because 6.4 hasn't been merged up to 7.0, so FWBundle & AssetMapper tests are using not-yet-up-to-date code.

fabpot (approved):

src/Symfony/Component/AssetMapper/CHANGELOG.md
```diff
@@ -9,13 +9,15 @@ CHANGELOG
...
  * Allow relative path strings in the importmap
+ * Automatically set `_links` attribute for preload CSS files for WebLink integration
  * Add `PreAssetsCompileEvent` event when running `asset-map:compile`
  * Add support for importmap paths to use the Asset component (for subdirectories)
  * Removed the `importmap:export` command
  * Add a `importmap:install` command to download all missing downloaded packages
  * Allow specifying packages to update for the `importmap:update` command
  * Add a `importmap:audit` command to check for security vulnerability advisories in dependencies
  * Add a `importmap:outdated` command to check for outdated packages
+ * Automatically add Link preload headers for CSS files when using the WebLink component
```
fabpot: Looks like you've added the changelog twice.

weaverryan: I just think it's a really great change 🫠

fabpot:
Thank you @weaverryan.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapRenderer.php
```diff
@@ -131,4 +142,25 @@ private function createAttributesString(array $attributes): string
 
         return $attributeString;
     }
+
+    private function addWebLinkPreloads(Request $request, array $cssLinks): void
+    {
+        $cssPreloadLinks = array_map(fn ($url) => new Link('preload', $url), $cssLinks);
```
gharlan, line 148: I'm not sure, but shouldn't the nopush attribute be better set here? So that the css files are not sent again and again, even if they are already in client cache.

From https://developer.chrome.com/blog/early-hints/:
> with HTTP2/Push it was extremely hard to avoid pushing sub-resources that the browser already had. This "over-pushing" effect resulted in a less efficient usage of the network bandwidth, which significantly hindered the performance benefits.

In current Chrome versions it wouldn't make any difference, because Server Push is already disabled there. But afaik other browsers still support Server Push in current versions.

```diff
- $cssPreloadLinks = array_map(fn ($url) => new Link('preload', $url), $cssLinks);
+ $cssPreloadLinks = array_map(fn ($url) => (new Link('preload', $url))->withAttribute('nopush', true), $cssLinks);
```

smnandre:
Not sure who implemented "nopush" but it seems not to be considered by most used browsers

https://wpt.fyi/results/preload?label=master&label=experimental&aligned&q=nopush 

(compared to https://wpt.fyi/results/preload?label=master&label=experimental&aligned&q=preload` )

gharlan:
> Not sure who implemented "nopush" but it seems not to be considered by most used browsers

I think the browsers do not need to implement "nopush". It is the server (apache, nginx, ...) that decides based on the nopush attribute whether to push the resource or just send the preload header.

weaverryan:
See #52143 - thanks for bringing up this possible issue

PR #44774, opened 2021-12-23
[DI] Add `exclude` to `TaggedIterator` and `TaggedLocator`

ruudk:
| Q             | A
| ------------- | ---
| Branch?       | 6.1 
| Bug fix?      |no
| New feature?  | yes
| Deprecations? |no 
| Tickets       | 
| License       | MIT
| Doc PR        | 

With this change, you can exclude one or more services from a tagged iterator
or locator.

It's common when working with Delegator or Chain services that call a list of services that
implement an interface, to also want to implement the interface on Delegator/Chain.

An example of this is in the Symfony HttpKernel component:

https://github.com/symfony/symfony/blob/0d6e859db236e37b8baa2b2bf4c8b7d14d151570/src/Symfony/Component/HttpKernel/CacheClearer/CacheClearerInterface.php#L19-L25
https://github.com/symfony/symfony/blob/0d6e859db236e37b8baa2b2bf4c8b7d14d151570/src/Symfony/Component/HttpKernel/CacheClearer/ChainCacheClearer.php#L21-L31

The `ChainCacheClearer` implements `CacheClearerInterface` and calls a list of `CacheClearerInterface`
clearers.

If that would have been configured with `#[AutoconfigureTag]` and `#[TaggedIterator]`, the
`ChainCacheClearer` would receive itself in the `iterable $clearers`.

With this change, it can exclude itself and rely fully on autowire/configure.

Another example:

```php
#[AutoconfigureTag]
interface ErrorTracker
{
    public function trackError(string $error): void;
}

final class SentryErrorTracker implement ErrorTracker
{
    public function trackError(string $error): void
    {
        echo "Send error to Sentry\n";
    }
}

final class NewRelicErrorTracker implement ErrorTracker
{
    public function trackError(string $error): void
    {
        echo "Send error to NewRelic\n";
    }
}

final class DelegatingErrorTracker implements ErrorTracker
{
    public function __construct(
        #[TaggedIterator(ErrorTracker::class, exclude: self::class)]
        private iterable $trackers
    ) {}

    public function trackError(string $error): void
    {
        foreach($this->trackers as $tracker) {
            $tracker->trackError($error);
        }
    }
}

// Alias ErrorTracker interface to DelegatingErrorTracker service

final class MyController
{
    public function __construct(private ErrorTracker $errorTracker) {}
    public function __invoke()
    {
        $this->errorTracker->trackError('Hello, World!');
    }
}
```

Without this change, the `DelegatingErrorTracker` would receive itself in the tagged
iterator.

PR #51832, opened 2023-10-03
[DependencyInjection] Add `#[AutowireIterator]` attribute and improve `#[AutowireLocator]`

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT

This PR is building on #51392 to add an `#[AutowireIterator]` attribute and improve `#[AutowireLocator]`.

The new `#[AutowireIterator]` attribute can be used to describe what `#[AutowireLocator]` can do, except that we get an iterator instead of a container.

And  `#[AutowireLocator]` can now be used instead of `#[TaggedLocator]`: `#[AutowireLocator('foo')]` and done.

In order to describe that you want a list of services, we cannot use named arguments anymore so we have to pass an array now: `#[AutowireLocator(['foo' => 'F', 'bar' => 'B'])]` should be used instead of `#[AutowireLocator(foo: 'F', bar: 'B')]`.

Last but not least, this adds support for nesting `SubscribedService` objects in the list of described services. This provides feature-parity with what we can do when implementing `ServiceSubscriberInterface`.

I didn't deprecate `TaggedIterator` nor `TaggedLocator`. We could, but maybe we should wait for 7.1?

TODO:
- [x] add tests - thanks @kbond!

~PS: while writing this, I realize that we may merge both tags in one, and let `AutowirePass` decide if it should build a locator or an iterator based on the type of the argument that has the attribute. We'd "just" need to find a name that'd work for that.~

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
+++ b/src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
@@ -1,0 +1,77 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\DependencyInjection\Attribute;
+
+use Symfony\Component\DependencyInjection\Argument\IteratorArgument;
+use Symfony\Component\DependencyInjection\Argument\TaggedIteratorArgument;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
+use Symfony\Component\DependencyInjection\TypedReference;
+use Symfony\Contracts\Service\Attribute\SubscribedService;
+use Symfony\Contracts\Service\ServiceSubscriberInterface;
+
+/**
+ * Autowires an iterator of services based on a tag name or an explicit list of key => service-type pairs.
+ */
+#[\Attribute(\Attribute::TARGET_PARAMETER)]
+class AutowireIterator extends Autowire
+{
+    /**
+     * @see ServiceSubscriberInterface::getSubscribedServices()
+     *
+     * @param string|array<string|SubscribedService> $services A tag name or an explicit list of services
+     * @param string|string[]                        $exclude  A service or a list of services to exclude
+     */
+    public function __construct(
+        string|array $services,
+        string $indexAttribute = null,
+        string $defaultIndexMethod = null,
+        string $defaultPriorityMethod = null,
+        string|array $exclude = [],
+        bool $excludeSelf = true,
+    ) {
+        if (\is_string($services)) {
+            parent::__construct(new TaggedIteratorArgument($services, $indexAttribute, $defaultIndexMethod, false, $defaultPriorityMethod, (array) $exclude, $excludeSelf));
+
+            return;
+        }
+
+        $references = [];
+
+        foreach ($services as $key => $type) {
+            $attributes = [];
+
+            if ($type instanceof SubscribedService) {
+                $key = $type->key ?? $key;
+                $attributes = $type->attributes;
+                $type = ($type->nullable ? '?' : '').($type->type ?? throw new InvalidArgumentException(sprintf('When "%s" is used, a type must be set.', SubscribedService::class)));
+            }
+
+            if (!\is_string($type) || !preg_match('/(?(DEFINE)(?<cn>[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*+))(?(DEFINE)(?<fqcn>(?&cn)(?:\\\\(?&cn))*+))^\??(?&fqcn)(?:(?:\|(?&fqcn))*+|(?:&(?&fqcn))*+)$/', $type)) {
+                throw new InvalidArgumentException(sprintf('"%s" is not a PHP type for key "%s".', \is_string($type) ? $type : get_debug_type($type), $key));
+            }
+            $optionalBehavior = ContainerInterface::EXCEPTION_ON_INVALID_REFERENCE;
+            if ('?' === $type[0]) {
+                $type = substr($type, 1);
+                $optionalBehavior = ContainerInterface::IGNORE_ON_INVALID_REFERENCE;
+            }
+            if (\is_int($name = $key)) {
+                $key = $type;
+                $name = null;
+            }
+
+            $references[$key] = new TypedReference($type, $type, $optionalBehavior, $name, $attributes);
+        }
+
+        parent::__construct(new IteratorArgument($references));
+    }
+}
--- a/src/Symfony/Component/DependencyInjection/Attribute/AutowireLocator.php
+++ b/src/Symfony/Component/DependencyInjection/Attribute/AutowireLocator.php
@@ -11,32 +11,40 @@
 
 namespace Symfony\Component\DependencyInjection\Attribute;
 
+use Symfony\Component\DependencyInjection\Argument\IteratorArgument;
 use Symfony\Component\DependencyInjection\Argument\ServiceLocatorArgument;
-use Symfony\Component\DependencyInjection\ContainerInterface;
-use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\DependencyInjection\Argument\TaggedIteratorArgument;
+use Symfony\Contracts\Service\Attribute\SubscribedService;
+use Symfony\Contracts\Service\ServiceSubscriberInterface;
 
+/**
+ * Autowires a service locator based on a tag name or an explicit list of key => service-type pairs.
+ */
 #[\Attribute(\Attribute::TARGET_PARAMETER)]
 class AutowireLocator extends Autowire
 {
-    public function __construct(string ...$serviceIds)
-    {
-        $values = [];
+    /**
+     * @see ServiceSubscriberInterface::getSubscribedServices()
+     *
+     * @param string|array<string|SubscribedService> $services An explicit list of services or a tag name
+     * @param string|string[]                        $exclude  A service or a list of services to exclude
+     */
+    public function __construct(
+        string|array $services,
+        string $indexAttribute = null,
+        string $defaultIndexMethod = null,
+        string $defaultPriorityMethod = null,
+        string|array $exclude = [],
+        bool $excludeSelf = true,
+    ) {
+        $iterator = (new AutowireIterator($services, $indexAttribute, $defaultIndexMethod, $defaultPriorityMethod, (array) $exclude, $excludeSelf))->value;
 
-        foreach ($serviceIds as $key => $serviceId) {
-            if ($nullable = str_starts_with($serviceId, '?')) {
-                $serviceId = substr($serviceId, 1);
-            }
-
-            if (is_numeric($key)) {
-                $key = $serviceId;
-            }
-
-            $values[$key] = new Reference(
-                $serviceId,
-                $nullable ? ContainerInterface::IGNORE_ON_INVALID_REFERENCE : ContainerInterface::EXCEPTION_ON_INVALID_REFERENCE,
-            );
+        if ($iterator instanceof TaggedIteratorArgument) {
+            $iterator = new TaggedIteratorArgument($iterator->getTag(), $iterator->getIndexAttribute(), $iterator->getDefaultIndexMethod(), true, $iterator->getDefaultPriorityMethod(), $iterator->getExclude(), $iterator->excludeSelf());
+        } elseif ($iterator instanceof IteratorArgument) {
+            $iterator = $iterator->getValues();
         }
 
-        parent::__construct(new ServiceLocatorArgument($values));
+        parent::__construct(new ServiceLocatorArgument($iterator));
     }
 }
--- a/src/Symfony/Component/DependencyInjection/Tests/Compiler/IntegrationTest.php
+++ b/src/Symfony/Component/DependencyInjection/Tests/Compiler/IntegrationTest.php
@@ -15,6 +15,7 @@ use PHPUnit\Framework\TestCase;
 use Psr\Container\ContainerInterface;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Alias;
+use Symfony\Component\DependencyInjection\Argument\RewindableGenerator;
 use Symfony\Component\DependencyInjection\Argument\ServiceLocatorArgument;
 use Symfony\Component\DependencyInjection\Argument\TaggedIteratorArgument;
 use Symfony\Component\DependencyInjection\ChildDefinition;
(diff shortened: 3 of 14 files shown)


src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
```diff
@@ -0,0 +1,71 @@
...
+            return;
+        }
+
+        $references = [];
+
+        foreach ($services as $key => $type) {
+            $attributes = [];
+
+            if ($type instanceof SubscribedService) {
+                $key = $type->key ?? $key;
```
kbond, line 54: I added to `?? $key` here but I'm not sure how this should work? Should the `$services` array key take precedence over the `SubscribedService`'s key?

src/Symfony/Component/DependencyInjection/Tests/Compiler/IntegrationTest.php
```diff
@@ -411,6 +413,34 @@ public function testLocatorConfiguredViaAttribute()
...
+            ->setAutowired(true)
+            ->setPublic(true)
+        ;
+
+        $container->compile();
+
+        /** @var AutowireIteratorConsumer $s */
+        $s = $container->get(AutowireIteratorConsumer::class);
+
+        self::assertIsArray($s->iterator);
```
kbond: Is this correct? I expected this to be a `RewindableGenerator`.

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
nicolas-grekas: Why not, but we have to sync with https://github.com/symfony/symfony/blob/9767270b34e97d1bbf340f9ab3358a0178dccb8a/src/Symfony/Component/DependencyInjection/Compiler/RegisterServiceSubscribersPass.php#L82
No opinion yet, we need to think about why we didn't back then and why we would now

src/Symfony/Component/DependencyInjection/Tests/Compiler/IntegrationTest.php
nicolas-grekas: And let AutowirePass unwrap when only an array is accepted, right?

kbond: It was my understanding that a tagged iterator argument is _always_ injected as `RewindableGenerator` (regardless of the type-hint) so it's lazy. In the docs we show using `iterable` for the type-hint. In other tests it's clearly a `RewindableGenerator`, I'm not sure why in this case it isn't.

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
kbond: Hmm, yeah. I think it's safe to propagate this change to `RegisterServiceSubscriberPass` so _the key in `SubscribedService` takes precedence over the array key_ (the opposite of what I proposed above).

nicolas-grekas: I updated `RegisterServiceSubscriberPass` (not sure about why you mean "the opposite", this look correct as is to me?)

src/Symfony/Component/DependencyInjection/Tests/Compiler/IntegrationTest.php
nicolas-grekas: Updated!

nicolas-grekas:
PR updated and ready. Big Hurray to @kbond for the tests :pray: :hugs:

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
kbond: All good for me _as-is_.

kbond (approved):

KerianMM:
TaggedIterator and TaggedLocator attributes are they still use full?

nicolas-grekas:
@KerianMM that's an open question. I see no hurry in deprecating them, I'd be fine waiting for 7.1.

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
```diff
@@ -0,0 +1,72 @@
...
+     * @see ServiceSubscriberInterface::getSubscribedServices()
+     *
+     * @param array<string|SubscribedService>|string $services An explicit list of services or a tag name
+     */
+    public function __construct(
+        array|string $services,
+        string $indexAttribute = null,
+        string $defaultIndexMethod = null,
+        string $defaultPriorityMethod = null,
+        string|array $exclude = [],
```
weaverryan, line 39: Is this an array of service ids? Also, super minor, we have `array|string` above and `string|array` here.

weaverryan (approved):

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
```diff
@@ -0,0 +1,72 @@
...
+            if ($type instanceof SubscribedService) {
+                $key = $type->key ?? $key;
+                $attributes = $type->attributes;
+                $type = ($type->nullable ? '?' : '').($type->type ?? throw new InvalidArgumentException(sprintf('When "%s" is used, a type must be set.', SubscribedService::class)));
+            }
+
+            if (!\is_string($type) || !preg_match('/(?(DEFINE)(?<cn>[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*+))(?(DEFINE)(?<fqcn>(?&cn)(?:\\\\(?&cn))*+))^\??(?&fqcn)(?:(?:\|(?&fqcn))*+|(?:&(?&fqcn))*+)$/', $type)) {
+                throw new InvalidArgumentException(sprintf('"%s" is not a PHP type for key "%s".', \is_string($type) ? $type : get_debug_type($type), $key));
+            }
+            if ($optionalBehavior = '?' === $type[0]) {
```
GromNaN: I know you like to make assignments in `if`, but in this case it's much more readable if extracted (or use `else`). The `TypedReference` creation can also be simplified.
```suggestion
            $optionalBehavior = ContainerInterface::EXCEPTION_ON_INVALID_REFERENCE;
            if ('?' === $type[0]) {
```

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
```diff
@@ -0,0 +1,72 @@
...
+use Symfony\Component\DependencyInjection\Argument\IteratorArgument;
+use Symfony\Component\DependencyInjection\Argument\TaggedIteratorArgument;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
+use Symfony\Component\DependencyInjection\TypedReference;
+use Symfony\Contracts\Service\Attribute\SubscribedService;
+use Symfony\Contracts\Service\ServiceSubscriberInterface;
+
+#[\Attribute(\Attribute::TARGET_PARAMETER)]
+class AutowireIterator extends Autowire
```
GromNaN, line 26: This can be done in a future PR, I find there's not enough comment in the header of attribute classes to describe what they're for. Yet that's where we look first when an attribute is used.

src/Symfony/Component/DependencyInjection/Attribute/AutowireLocator.php
```diff
@@ -11,32 +11,36 @@
...
-            }
-
-            $values[$key] = new Reference(
-                $serviceId,
-                $nullable ? ContainerInterface::IGNORE_ON_INVALID_REFERENCE : ContainerInterface::EXCEPTION_ON_INVALID_REFERENCE,
-            );
+    /**
+     * @see ServiceSubscriberInterface::getSubscribedServices()
+     *
+     * @param array<string|SubscribedService>|string $services An explicit list of services or a tag name
```
stof: for locators, don't we expect an array with string keys ? If yes, this should be documented in the type.

stof:
Why not improving the existing attributes instead of adding new ones ?

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
stof, line 39: The type of what goes in this array should be described in `@param` to help static analyzers to report bad usages.

nicolas-grekas: Correct, see https://github.com/symfony/symfony/pull/44774/files#diff-e95ab96f0f1aa21244f00f2494014aae9e8d951e09c02e67c94e3a42bdbaba95R58
There's some overlap with `$excludeSelf` since the use case in the PR is to `$exclude = self::class`. We might have some legacy here but :shrug:

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
chalasr: that would break the `?:`  below though. Maybe rename it to `$invalidBehavior` and use a good old `else` instead (I agree this one is hard to read)

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
```diff
@@ -0,0 +1,72 @@
...
+
+            if ($type instanceof SubscribedService) {
+                $key = $type->key ?? $key;
+                $attributes = $type->attributes;
+                $type = ($type->nullable ? '?' : '').($type->type ?? throw new InvalidArgumentException(sprintf('When "%s" is used, a type must be set.', SubscribedService::class)));
+            }
+
+            if (!\is_string($type) || !preg_match('/(?(DEFINE)(?<cn>[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*+))(?(DEFINE)(?<fqcn>(?&cn)(?:\\\\(?&cn))*+))^\??(?&fqcn)(?:(?:\|(?&fqcn))*+|(?:&(?&fqcn))*+)$/', $type)) {
+                throw new InvalidArgumentException(sprintf('"%s" is not a PHP type for key "%s".', \is_string($type) ? $type : get_debug_type($type), $key));
+            }
```
chalasr, line 61: it would be nice to factorize this heavy validation check somehow

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
nicolas-grekas: Wouldn't mean the same :) Also, it's borrowed from RegisterServiceSubscribersPass

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
nicolas-grekas: updated

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
nicolas-grekas: Borrowed from RegisterServiceSubscribersPass also. I would create a new API for that so I'm fine as is personally. It's not going to be a moving target - I don't expect any changes on this.

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
nicolas-grekas: good idea, updated

src/Symfony/Component/DependencyInjection/Attribute/AutowireLocator.php
nicolas-grekas: It's fine to use "no key", then we use the type as the key, FQCN ids FTW

nicolas-grekas:
> Why not improving the existing attributes instead of adding new ones ?

Because I think it's important to push the `Autowire*` convention. It helps discovering/remembering those autowiring-related attributes.

chalasr (approved):

chalasr:
> I didn't deprecate TaggedIterator nor TaggedLocator. We could, but maybe we should wait for 7.1?

👍 to do it on 7.1 and give time to people having them since 5.3. The docs should probably recommend the new ones asap though

OskarStark:
I updated the syntax for `#[AutowireLocator]`:
* https://github.com/symfony/symfony-docs/pull/18993

OskarStark:
> Last but not least, this adds support for nesting SubscribedService objects in the list of described services. This provides feature-parity with what we can do when implementing ServiceSubscriberInterface.

Can you add a code example please? Thanks

nicolas-grekas:
What's in ServiceSubscriberInterface:
https://github.com/symfony/symfony/blob/d27190a4cf8560b77d8aaab65104e51a2dd5c06c/src/Symfony/Contracts/Service/ServiceSubscriberInterface.php#L53-L57

OskarStark:
Updated

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
GromNaN: I understand it was done this way 6 years ago in [`RegisterServiceSubscribersPass`](https://github.com/symfony/symfony/blame/d27190a4cf8560b77d8aaab65104e51a2dd5c06c/src/Symfony/Component/DependencyInjection/Compiler/RegisterServiceSubscribersPass.php#L90-L93). But let me insist, I don't see why the implementation "Wouldn't mean the same".

```php
$optionalBehavior = ContainerInterface::EXCEPTION_ON_INVALID_REFERENCE;
if ('?' === $type[0]) {
    $type = substr($type, 1);
    $optionalBehavior = ContainerInterface::IGNORE_ON_INVALID_REFERENCE;
}
// ...
$references[$key] = new TypedReference($type, $type, $optionalBehavior, $name, $attributes);
```

nicolas-grekas: alright, updated!

nicolas-grekas:
Thank you @kbond.

src/Symfony/Component/HttpKernel/composer.json
```diff
@@ -63,6 +63,7 @@
         "symfony/http-client-contracts": "<2.5",
         "symfony/mailer": "<5.4",
         "symfony/messenger": "<5.4",
+        "symfony/service-contracts": "<3.2",
```
bendavies, line 66: @kbond why was this added? is there a way to remove this conflict?

this makes it not possible for us to upgrade to symfony 6.4 because this forces a requirement for `psr/container ^2.0`, and we are still on v1 because the latest version of `laminas/laminas-servicemanager` requires v1:
https://github.com/laminas/laminas-servicemanager/blob/3.23.x/composer.json#L26

this is sort of the same issue as https://github.com/api-platform/core/pull/5811

kbond: /cc @nicolas-grekas

bendavies: Sorry @kbond, tagged you as it looked like your commit.

nicolas-grekas: This is needed because the test case uses the $type argument of SubscribedService, which exists since 3.2.
PR welcome to change the test case and relax this.

bendavies: thanks

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        $references = [];
+
+        foreach ($services as $key => $type) {
+            $attributes = [];
+
+            if ($type instanceof SubscribedService) {
+                $key = $type->key ?? $key;
+                $attributes = $type->attributes;
```
bendavies, line 55: `symfony/dependency-injection` requires `"symfony/service-contracts": "^2.5|^3.0`, but `attributes`, `nullable` and `type` are not defined on `SubscribedService` in until 3.2.
is this a mistake?

PR #51835, opened 2023-10-04
[DependencyInjection] Remove inaccurate `@throws` on `Container::get()`

### LINKED ISSUE #51834: [DependencyInjection] `ContainerInterface::get` and `Container::get` throws tag are not consistent

5.4.x

### Description

ContainerInterface::get throws tags are
https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/DependencyInjection/ContainerInterface.php
```
* @throws ServiceCircularReferenceException When a circular reference is detected
* @throws ServiceNotFoundException          When the service is not defined
```

Which means I should be able to write code 
```
public function do(ContainerInterface $c) {
     try {
         $c->get(...);
    } catch (ServiceCircularReferenceException|ServiceNotFoundException) {
        return;
    }
}
```
with no exception thrown.

But seems like one implementation Container can throw any exception, cf 
https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/DependencyInjection/Container.php#L204
So it doesn't respect the interface.

### How to reproduce

Irrelevant

### Possible Solution

Either `\Exception` should be removed from `Container` or added to the interface.

But even `\Exception` should be improved:
- An interface `CustomExceptionInterface` or `Throwable` should be preferred
- If it's an exception thrown by Symfony it should be Symfony-scoped like every others `Symfony\Foo\Exception`
- If it's an exception thrown by the User it should be `Throwable`

I think the `\Exception` is coming from here,
https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/DependencyInjection/Container.php#L237

A nice solution could be to:
- Introduce a `ServiceLoadingException implements Symfony\Component\DependencyInjection\Exception\ExceptionInterface`
- re-throw new `ServiceLoadingException('...', $e->getCode(), $e);` instead of `$e` directly.
- Updating the ContainerInterface to add `@throws ServiceLoadingException`

WDYT ? This would be done with https://github.com/symfony/symfony/pull/51835

### Additional Context

This was added here: https://github.com/symfony/symfony/commit/1775da59255010724f2864751c8d1256729727e7#diff-98156f7ccbe7f78db5bc7ff3f81ddb5a76532f652dd7520f38b472219d92f394R264 by @xabbuh

VincentLanglet:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #51834
| License       | MIT

This would solve the inconsistency between ContainerInterface and Container described in https://github.com/symfony/symfony/issues/51834 because currently the Container phpdoc doesn't respect the ContainerInterface one.

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Container.php
+++ b/src/Symfony/Component/DependencyInjection/Container.php
@@ -201,7 +201,6 @@ class Container implements ContainerInterface, ResetInterface
      *
      * @throws ServiceCircularReferenceException When a circular reference is detected
      * @throws ServiceNotFoundException          When the service is not defined
-     * @throws \Exception                        if an exception has been thrown when the service has been resolved
      *
      * @see Reference
      */


derrabus:
Wouldn't we violate PSR-11 if we upcast `ServiceNotFoundException`?

VincentLanglet:
> Wouldn't we violate PSR-11 if we upcast `ServiceNotFoundException`?

Sorry, I'm not sure to understand ; can you develop ?

Currently PSR\Container\ContainerInterface have the following phpdoc:
```
interface ContainerInterface
{
    /**
     * Finds an entry of the container by its identifier and returns it.
     *
     * @param string $id Identifier of the entry to look for.
     *
     * @throws NotFoundExceptionInterface  No entry was found for **this** identifier.
     * @throws ContainerExceptionInterface Error while retrieving the entry.
     *
     * @return mixed Entry.
     */
    public function get(string $id);
}
```
So, to be PSR, Symfony should only throw NotFoundExceptionInterface and ContainerExceptionInterface.

Currently, since Symfony allows to throws `\Exception` in the Container::get method,  in my understanding the Symfony Container doesn't fully respect the PSR interface.

The introduced `ServiceLoadingException` implements `ContainerExceptionInterface` so it will now respect the PSR interface.

src/Symfony/Component/DependencyInjection/Container.php
```diff
@@ -234,7 +231,7 @@ private static function make(self $container, string $id, int $invalidBehavior):
         } catch (\Exception $e) {
             unset($container->services[$id]);
 
-            throw $e;
```
stof: if the exception is already a PSR one, should we rethrow it as is instead of wrapping it ?

VincentLanglet: Indeed it's better.

I have added 
```
if ($e instanceof ExceptionInterface) {
     throw $e;
}
```
To only wrap non SymfonyContainerException, I think it's enough and kinda follow the "Always throws Symfony exception" policy. Or do you think I should extends to all PSR exceptions ?

VincentLanglet:
Is this PR ok @derrabus @stof ? I'm not sure but failures seem unrelated

src/Symfony/Component/DependencyInjection/Container.php
```diff
@@ -199,10 +201,6 @@ public function has(string $id): bool
     /**
      * Gets a service.
      *
-     * @throws ServiceCircularReferenceException When a circular reference is detected
-     * @throws ServiceNotFoundException          When the service is not defined
-     * @throws \Exception                        if an exception has been thrown when the service has been resolved
```
nicolas-grekas, line 204: I would simply remove this line and nothing else.

VincentLanglet: This would kinda hide the problem.

Currently there is a reproducer in the Symfony tests:
https://github.com/symfony/symfony/blob/ce48fbe3de4734101d98a98615c0afbd6e88c7d4/src/Symfony/Component/DependencyInjection/Tests/ContainerTest.php#L334-L348

https://github.com/symfony/symfony/blob/ce48fbe3de4734101d98a98615c0afbd6e88c7d4/src/Symfony/Component/DependencyInjection/Tests/ContainerTest.php#L491-L494

Should it be Symfony responsibility to try/catch and rethrow a compliant exception or should we rely on the developer to throw the right exception ?

The first solution seemed safer

nicolas-grekas: `@throws` is to document exceptions that can be thrown under the responbility of the called method. Re-thrown exceptions don't fit this definition so this `@throws` is just wrong. Wrapping them all won't provide anything more.

VincentLanglet: To me the purpose of the `@throws` tag is to know which Exceptions I should catch when doing
```
$container->get()
```
That's also how static analysis like PHPStan works.

But I'm ok removing the `\Exception` line, it solves my issue.
Pr is update

nicolas-grekas (approved):

nicolas-grekas:
Thank you @VincentLanglet.

PR #51838, opened 2023-10-04
[Cache] Add type to redis scan call to avoid using default "" which r…

mfadul24:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

The `cache:pool:clear` command is not clearing the existing redis keys correctly. The method is being call without a valid `type`, which gets a empty array as response.

The redis `MONITOR` command shows the following type of call:
```"SCAN" "0" "COUNT" "1000" "MATCH" "<prefix>:*" "TYPE" ""```

Correct would be to set the `TYPE` to `string`.
```"SCAN" "0" "COUNT" "1000" "MATCH" "<prefix>:*" "TYPE" "string"```

See Documentation: https://redis.io/commands/type/

This PullRequest also fix the deprecation:
```Deprecated: Redis::scan(): Passing null to parameter #4 ($type) of type ?string is deprecated {"exception":"[object] (ErrorException(code: 0): Deprecated: Redis::scan(): Passing null to parameter #4 ($type) of type ?string is deprecated at /var/www/vendor/symfony/cache/Traits/Redis6Proxy.php:893)"```

### DIFF
--- a/src/Symfony/Component/Cache/Traits/RedisTrait.php
+++ b/src/Symfony/Component/Cache/Traits/RedisTrait.php
@@ -476,7 +476,7 @@ trait RedisTrait
 
             $cursor = null;
             do {
-                $keys = $host instanceof \Predis\ClientInterface ? $host->scan($cursor, 'MATCH', $pattern, 'COUNT', 1000) : $host->scan($cursor, $pattern, 1000);
+                $keys = $host instanceof \Predis\ClientInterface ? $host->scan($cursor, 'MATCH', $pattern, 'COUNT', 1000) : $host->scan($cursor, $pattern, 1000, 'string');
                 if (isset($keys[1]) && \is_array($keys[1])) {
                     $cursor = $keys[0];
                     $keys = $keys[1];
--- a/src/Symfony/Component/Cache/Tests/Traits/RedisTraitTest.php
+++ b/src/Symfony/Component/Cache/Tests/Traits/RedisTraitTest.php
@@ -17,13 +17,6 @@ use Symfony\Component\Cache\Traits\RedisTrait;
 
 class RedisTraitTest extends TestCase
 {
-    public static function setUpBeforeClass(): void
-    {
-        if (!getenv('REDIS_CLUSTER_HOSTS')) {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by groensch24 — APPROVED** [review]

**Comment by OskarStark** [comment]
I removed the deprecations label, as this does not introduce a new deprecation, but fixes one.

Are you sure, this bug does not exist in 5.4?

**Comment by fabpot** [comment]
The code looks the same on 5.4.
Can you add a test as well?

**Comment by mfadul24** [comment]
Hi, I've update the PR to use the 5.4 branch and add a test for this use case. I hope it looks good enough for you. Look forward to my first contribution.

File: src/Symfony/Component/Cache/Traits/RedisTrait.php [hunk 1]
```diff
@@ -328,9 +332,12 @@ public static function createConnection(string $dsn, array $options = [])
                     $redis->setOption(\Redis::OPT_TCP_KEEPALIVE, $params['tcp_keepalive']);
                 }
                 switch ($params['failover']) {
-                    case 'error': $redis->setOption(\RedisCluster::OPT_SLAVE_FAILOVER, \RedisCluster::FAILOVER_ERROR); break;
-                    case 'distribute': $redis->setOption(\RedisCluster::OPT_SLAVE_FAILOVER, \RedisCluster::FAILOVER_DISTRIBUTE); break;
-                    case 'slaves': $redis->setOption(\RedisCluster::OPT_SLAVE_FAILOVER, \RedisCluster::FAILOVER_DISTRIBUTE_SLAVES); break;
+                    case 'error': $redis->setOption(\RedisCluster::OPT_SLAVE_FAILOVER, \RedisCluster::FAILOVER_ERROR);
```
OskarStark: Can you revert these CI fixes?

mfadul24: Done

nicolas-grekas:
Thanks for giving this a try. I'm closing in favor of #52036, which provides a more generic fix to such issues. This should still be fixed in phpredis, but at least we won't fall into the trap.

PR #51845, opened 2023-10-04
[AssetMapper] Add outdated command

maelanleborgne:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT

As suggested by @WebMamba, I added a new command for the AssetMapper component to list outdated 3rd party packages. It is inspired by the `composer outdated` command so I tried to replicate its options as much as I could. 
It reads the importmap.php and extract packages version to query the https://registry.npmjs.org/%package% API and read the latest version from metadata.

image

:warning: The code is base on #51650 branch so it is not ready to be merged yet, but I'd be happy to get some reviews and feedback in the meantime. This is my first PR on symfony, so there will probably be a lot to say !

- [x] gather feedback
- [x] wait for #51650 to be merged and rebase
- [x] write documentation

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/asset_mapper.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/asset_mapper.php
@@ -20,6 +20,7 @@ use Symfony\Component\AssetMapper\Command\AssetMapperCompileCommand;
 use Symfony\Component\AssetMapper\Command\DebugAssetMapperCommand;
 use Symfony\Component\AssetMapper\Command\ImportMapAuditCommand;
 use Symfony\Component\AssetMapper\Command\ImportMapInstallCommand;
+use Symfony\Component\AssetMapper\Command\ImportMapOutdatedCommand;
 use Symfony\Component\AssetMapper\Command\ImportMapRemoveCommand;
 use Symfony\Component\AssetMapper\Command\ImportMapRequireCommand;
 use Symfony\Component\AssetMapper\Command\ImportMapUpdateCommand;
@@ -32,6 +33,7 @@ use Symfony\Component\AssetMapper\ImportMap\ImportMapAuditor;
 use Symfony\Component\AssetMapper\ImportMap\ImportMapConfigReader;
 use Symfony\Component\AssetMapper\ImportMap\ImportMapManager;
 use Symfony\Component\AssetMapper\ImportMap\ImportMapRenderer;
+use Symfony\Component\AssetMapper\ImportMap\ImportMapUpdateChecker;
 use Symfony\Component\AssetMapper\ImportMap\RemotePackageDownloader;
 use Symfony\Component\AssetMapper\ImportMap\Resolver\JsDelivrEsmResolver;
 use Symfony\Component\AssetMapper\MapperAwareAssetPackage;
@@ -179,6 +181,11 @@ return static function (ContainerConfigurator $container) {
             service('asset_mapper.importmap.config_reader'),
             service('http_client'),
         ])
+        ->set('asset_mapper.importmap.update_checker', ImportMapUpdateChecker::class)
+        ->args([
+            service('asset_mapper.importmap.config_reader'),
+            service('http_client'),
+        ])
 
         ->set('asset_mapper.importmap.command.require', ImportMapRequireCommand::class)
             ->args([
@@ -204,6 +211,10 @@ return static function (ContainerConfigurator $container) {
 
         ->set('asset_mapper.importmap.command.audit', ImportMapAuditCommand::class)
             ->args([service('asset_mapper.importmap.auditor')])
+            ->tag('console.command')
+
+        ->set('asset_mapper.importmap.command.outdated', ImportMapOutdatedCommand::class)
+            ->args([service('asset_mapper.importmap.update_checker')])
             ->tag('console.command')
     ;
 };
--- a/src/Symfony/Component/AssetMapper/Command/ImportMapOutdatedCommand.php
+++ b/src/Symfony/Component/AssetMapper/Command/ImportMapOutdatedCommand.php
@@ -1,0 +1,106 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\AssetMapper\Command;
+
+use Symfony\Component\AssetMapper\ImportMap\ImportMapUpdateChecker;
+use Symfony\Component\AssetMapper\ImportMap\PackageUpdateInfo;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+#[AsCommand(name: 'importmap:outdated', description: 'List outdated JavaScript packages and their latest versions')]
+final class ImportMapOutdatedCommand extends Command
+{
+    private const COLOR_MAPPING = [
+        'update-possible' => 'yellow',
+        'semver-safe-update' => 'red',
+    ];
+
+    public function __construct(
+        private readonly ImportMapUpdateChecker $updateChecker,
+    ) {
+        parent::__construct();
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->addArgument(
+                name: 'packages',
+                mode: InputArgument::IS_ARRAY | InputArgument::OPTIONAL,
+                description: 'A list of packages to check',
+            )
+            ->addOption(
+                name: 'format',
+                mode: InputOption::VALUE_REQUIRED,
+                description: sprintf('The output format ("%s")', implode(', ', $this->getAvailableFormatOptions())),
+                default: 'txt',
+            )
+            ->setHelp(<<<'EOT'
+The <info>%command.name%</info> command will list the latest updates available for the 3rd party packages in <comment>importmap.php</comment>.
+Versions showing in <fg=red>red</> are semver compatible versions and you should upgrading.
+Versions showing in <fg=yellow>yellow</> are major updates that include backward compatibility breaks according to semver.
+
+   <info>php %command.full_name%</info>
+
+Or specific packages only:
+
+   <info>php %command.full_name% <packages></info>
+EOT
+            );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+        $packages = $input->getArgument('packages');
+        $packagesUpdateInfos = $this->updateChecker->getAvailableUpdates($packages);
+        $packagesUpdateInfos = array_filter($packagesUpdateInfos, fn ($packageUpdateInfo) => $packageUpdateInfo->hasUpdate());
+        if (0 === \count($packagesUpdateInfos)) {
+            return Command::SUCCESS;
+        }
+
+        $displayData = array_map(fn ($importName, $packageUpdateInfo) => [
+            'name' => $importName,
+            'current' => $packageUpdateInfo->currentVersion,
+            'latest' => $packageUpdateInfo->latestVersion,
+            'latest-status' => PackageUpdateInfo::UPDATE_TYPE_MAJOR === $packageUpdateInfo->updateType ? 'update-possible' : 'semver-safe-update',
+        ], array_keys($packagesUpdateInfos), $packagesUpdateInfos);
+
+        if ('json' === $input->getOption('format')) {
+            $io->writeln(json_encode($displayData, \JSON_PRETTY_PRINT | \JSON_UNESCAPED_SLASHES));
+        } else {
+            $table = $io->createTable();
+            $table->setHeaders(['Package', 'Current', 'Latest']);
+            foreach ($displayData as $datum) {
+                $color = self::COLOR_MAPPING[$datum['latest-status']] ?? 'default';
+                $table->addRow([
+                    sprintf('<fg=%s>%s</>', $color, $datum['name']),
+                    $datum['current'],
+                    sprintf('<fg=%s>%s</>', $color, $datum['latest']),
+                ]);
+            }
+            $table->render();
+        }
+
+        return Command::FAILURE;
+    }
+
+    private function getAvailableFormatOptions(): array
+    {
+        return ['txt', 'json'];
+    }
+}
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapConfigReader.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapConfigReader.php
@@ -68,12 +68,16 @@ class ImportMapConfigReader
                 throw new RuntimeException(sprintf('The importmap entry "%s" cannot have both a "path" and "version" option.', $importName));
             }
 
+            [$packageName, $filePath] = self::splitPackageNameAndFilePath($importName);
+
             $entries->add(new ImportMapEntry(
                 $importName,
                 path: $path,
                 version: $version,
                 type: $type,
                 isEntrypoint: $isEntry,
+                packageName: $packageName,
+                filePath: $filePath,
             ));
         }
 
@@ -143,5 +147,19 @@ class ImportMapConfigReader
         }
 
         return substr($url, $lastAt + 1, $nextSlash - $lastAt - 1);
+    }
+
+    public static function splitPackageNameAndFilePath(string $packageName): array
+    {
+        $filePath = '';
+        $i = strpos($packageName, '/');
+
+        if ($i && (!str_starts_with($packageName, '@') || $i = strpos($packageName, '/', $i + 1))) {
+            // @vendor/package/filepath or package/filepath
+            $filePath = substr($packageName, $i);
+            $packageName = substr($packageName, 0, $i);
+        }
+
+        return [$packageName, $filePath];
     }
 }
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapUpdateChecker.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapUpdateChecker.php
@@ -1,0 +1,90 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\AssetMapper\ImportMap;
+
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+class ImportMapUpdateChecker
+{
+    private const URL_PACKAGE_METADATA = 'https://registry.npmjs.org/%s';
+
+    public function __construct(
+        private readonly ImportMapConfigReader $importMapConfigReader,
+        private readonly HttpClientInterface $httpClient,
+    ) {
+    }
+
+    /**
+     * @param string[] $packages
+     *
+     * @return PackageUpdateInfo[]
+     */
+    public function getAvailableUpdates(array $packages = []): array
+    {
+        $entries = $this->importMapConfigReader->getEntries();
+        $updateInfos = [];
+        $responses = [];
+        foreach ($entries as $entry) {
+            if (null === $entry->packageName || null === $entry->version) {
+                continue;
+            }
+            if (\count($packages) && !\in_array($entry->packageName, $packages, true)) {
+                continue;
+            }
+
+            $responses[$entry->importName] = $this->httpClient->request('GET', sprintf(self::URL_PACKAGE_METADATA, $entry->packageName), ['headers' => ['Accept' => 'application/vnd.npm.install-v1+json']]);
+        }
+
+        foreach ($responses as $importName => $response) {
+            $entry = $entries->get($importName);
+            if (200 !== $response->getStatusCode()) {
+                throw new \RuntimeException(sprintf('Unable to get latest version for package "%s".', $entry->packageName));
+            }
+            $updateInfo = new PackageUpdateInfo($entry->packageName, $entry->version);
+            try {
+                $updateInfo->latestVersion = json_decode($response->getContent(), true)['dist-tags']['latest'];
+                $updateInfo->updateType = $this->getUpdateType($updateInfo->currentVersion, $updateInfo->latestVersion);
+            } catch (\Exception $e) {
+                throw new \RuntimeException(sprintf('Unable to get latest version for package "%s".', $entry->packageName), 0, $e);
+            }
+            $updateInfos[$importName] = $updateInfo;
+        }
+
+        return $updateInfos;
+    }
+
+    private function getVersionPart(string $version, int $part): ?string
+    {
+        return explode('.', $version)[$part] ?? $version;
+    }
+
+    private function getUpdateType(string $currentVersion, string $latestVersion): string
+    {
+        if (version_compare($currentVersion, $latestVersion, '>')) {
+            return PackageUpdateInfo::UPDATE_TYPE_DOWNGRADE;
+        }
+        if (version_compare($currentVersion, $latestVersion, '==')) {
+            return PackageUpdateInfo::UPDATE_TYPE_UP_TO_DATE;
+        }
+        if ($this->getVersionPart($currentVersion, 0) < $this->getVersionPart($latestVersion, 0)) {
+            return PackageUpdateInfo::UPDATE_TYPE_MAJOR;
+        }
+        if ($this->getVersionPart($currentVersion, 1) < $this->getVersionPart($latestVersion, 1)) {
+            return PackageUpdateInfo::UPDATE_TYPE_MINOR;
+        }
+        if ($this->getVersionPart($currentVersion, 2) < $this->getVersionPart($latestVersion, 2)) {
+            return PackageUpdateInfo::UPDATE_TYPE_PATCH;
+        }
+
+        throw new \LogicException(sprintf('Unable to determine update type for "%s" and "%s".', $currentVersion, $latestVersion));
+    }
+}
--- a/src/Symfony/Component/AssetMapper/ImportMap/PackageUpdateInfo.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/PackageUpdateInfo.php
@@ -1,0 +1,34 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\AssetMapper\ImportMap;
+
+class PackageUpdateInfo
+{
+    public const UPDATE_TYPE_DOWNGRADE = 'downgrade';
+    public const UPDATE_TYPE_UP_TO_DATE = 'up-to-date';
+    public const UPDATE_TYPE_MAJOR = 'major';
+    public const UPDATE_TYPE_MINOR = 'minor';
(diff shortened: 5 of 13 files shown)


src/Symfony/Component/AssetMapper/Command/ImportMapOutdatedCommand.php
```diff
@@ -0,0 +1,164 @@
...
+                shortcut: 'm',
+                mode: InputOption::VALUE_NONE,
+                description: 'Only shows packages that have minor SemVer-compatible updates.',
+            )
+            ->addOption(
+                name: 'patch-only',
+                shortcut: 'p',
+                mode: InputOption::VALUE_NONE,
+                description: 'Only shows packages that have patch-level SemVer-compatible updates.',
+            )
```
weaverryan: I would be ok with NO including the 5 options above. You did an amazing job replicating Composer - seriously. But I'm not sure we need ALL the same options - and it might be better to have a simpler version, at least at first until the need proves itself.

src/Symfony/Component/AssetMapper/Command/ImportMapOutdatedCommand.php
```diff
@@ -0,0 +1,164 @@
...
+            ->addOption(
+                name: 'strict',
+                shortcut: 's',
+                mode: InputOption::VALUE_NONE,
+                description: 'Exit with a non-zero status if any outdated packages were found.',
+            )
```
weaverryan: Let's remove this option and make this the default behavior.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapUpdateChecker.php
```diff
@@ -0,0 +1,107 @@
...
+
+    public function __construct(
+        private readonly ImportMapConfigReader $importMapConfigReader,
+        private readonly PackageResolverInterface $resolver,
+        private readonly HttpClientInterface $httpClient,
+    ) {
+    }
+
+    /**
+     * @return array<PackageUpdateInfo>
```
weaverryan:
```suggestion
     * @param string[] $packages
     * @return PackageUpdateInfo[]
```

src/Symfony/Component/AssetMapper/ImportMap/ImportMapUpdateChecker.php
```diff
@@ -0,0 +1,107 @@
...
+            $updateInfos[] = $this->getAvailableUpdateForEntry($entry);
+        }
+
+        return $updateInfos;
+    }
+
+    private function getAvailableUpdateForEntry(ImportMapEntry $entry): PackageUpdateInfo
+    {
+        $version = $this->resolver->getPackageVersion($entry->url) ?? $entry->version;
+        $updatesInfo = new PackageUpdateInfo($entry->importName, $entry->url, $version);
```
weaverryan:
```suggestion
        $updateInfo = new PackageUpdateInfo($entry->importName, $entry->url, $version);
```

src/Symfony/Component/AssetMapper/ImportMap/PackageUpdateInfo.php
```diff
@@ -0,0 +1,51 @@
...
+            $this->majorUpdate = $version;
+        }
+    }
+
+    public function addMinorUpdate(string $version): void
+    {
+        if (null === $this->minorUpdate) {
+            $this->minorUpdate = $version;
+        }
+    }
```
weaverryan: A package will only have a major OR minor OR patch update, but never more than one, right? Perhaps we just have one property - e.g. `latestVersion` with some internal "type" flag to track if it's a major, minor, patch.

weaverryan (commented):
Congrats on your first PR! This is really nice work - I've left some comments below to get things started. FYI, you don't need to (you shouldn't) click to merge my suggestions. Instead, "steal" them and commit them yourself. That will keep your commits clean and all authored by you.

Thanks!

maelanleborgne:
Thank you very much for the review and the feedback @weaverryan :pray: . I've made changes according to your suggestions, let me know if you see anything else I could do.

weaverryan:
@maelanleborgne #51650 is merged, so you can rebase your commits to remove them from this PR. Probably THAT PR was also rebased, so this can get a little tricky - I've been doing it a lot lately 😆

src/Symfony/Component/AssetMapper/ImportMap/ImportMapUpdateChecker.php
```diff
@@ -0,0 +1,101 @@
...
+        }
+        $updateInfo->latestVersion = $this->getLatestPackageVersion($entry->importName);
+        $updateInfo->updateType = $this->getUpdateType($currentVersion, $updateInfo->latestVersion);
+
+        return $updateInfo;
+    }
+
+    private function getLatestPackageVersion(string $packageName): string
+    {
+        $response = $this->httpClient->request('GET', sprintf(self::URL_PACKAGE_METADATA, $packageName, ['headers' => ['Accept' => 'application/vnd.npm.install-v1+json']]));
```
weaverryan: the `sprintf` closing `)` needs to end a bit earlier :)

src/Symfony/Component/AssetMapper/ImportMap/ImportMapUpdateChecker.php
```diff
@@ -0,0 +1,101 @@
...
+        $entries = $this->importMapConfigReader->getEntries();
+        $updateInfos = [];
+        foreach ($entries as $entry) {
+            if (null === $entry->url) {
+                continue;
+            }
+            if (ImportMapType::CSS === $entry->type) {
+                continue;
+            }
+            if (\count($packages) && !\in_array($entry->importName, $packages, true)) {
```
weaverryan: I think there's a problem here - it's a bit annoying, tbh, and I deal with it in other spots as well. `$entry->importName` may be a bare package name or it may be a package + path - e.g. `chart.js/auto` (`chart.js` is the package name and `auto` is a path within the package).

So, we need to do some work to split the package from the path. I do it in my PR - https://github.com/symfony/symfony/pull/51786/files#diff-1a5c8dfbbfec4f3e0bc1fd209545bb56a483fa92e2212f0a3a99c8137468eb35R224 - but perhaps we should put this inside of `ImportMapEntry`? A new `getPackageName()` method and a `getPath()` method?

src/Symfony/Component/AssetMapper/ImportMap/ImportMapUpdateChecker.php
```diff
@@ -0,0 +1,101 @@
...
+    {
+        $entries = $this->importMapConfigReader->getEntries();
+        $updateInfos = [];
+        foreach ($entries as $entry) {
+            if (null === $entry->url) {
+                continue;
+            }
+            if (ImportMapType::CSS === $entry->type) {
+                continue;
+            }
```
weaverryan: I think we can allow CSS entries. You may have added this due to the "path" problem I mention 1 line below. Once we have that sorted, it should be no problem to handle CSS.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapUpdateChecker.php
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    private function getAvailableUpdateForEntry(ImportMapEntry $entry): PackageUpdateInfo
+    {
+        $currentVersion = $this->resolver->getPackageVersion($entry->url) ?? $entry->version;
+        $updateInfo = new PackageUpdateInfo($entry->importName, $entry->url, $currentVersion);
+
+        if (ImportMapType::CSS === $entry->type) {
+            return $updateInfo;
+        }
```
weaverryan: I think we don't need this? CSS can also be something that we show can be updated?

src/Symfony/Component/AssetMapper/ImportMap/ImportMapUpdateChecker.php
```diff
@@ -0,0 +1,101 @@
...
+            return PackageUpdateInfo::UPDATE_TYPE_MAJOR;
+        }
+        if ($this->getVersionPart($currentVersion, 1) < $this->getVersionPart($latestVersion, 1)) {
+            return PackageUpdateInfo::UPDATE_TYPE_MINOR;
+        }
+        if ($this->getVersionPart($currentVersion, 2) < $this->getVersionPart($latestVersion, 2)) {
+            return PackageUpdateInfo::UPDATE_TYPE_PATCH;
+        }
+
+        return 'unknown';
```
weaverryan: Let's throw an exception. If the user gets here, it looks like it would be some bug in our logic, right? I mean: there is no normal way that we can think of right now where this would happen, right?

src/Symfony/Component/AssetMapper/ImportMap/PackageUpdateInfo.php
```diff
@@ -0,0 +1,36 @@
...
+    public const UPDATE_TYPE_UNKNOWN = 'unknown';
+    public const UPDATE_TYPE_DOWNGRADE = 'downgrade';
+    public const UPDATE_TYPE_UP_TO_DATE = 'up-to-date';
+    public const UPDATE_TYPE_MAJOR = 'major';
+    public const UPDATE_TYPE_MINOR = 'minor';
+    public const UPDATE_TYPE_PATCH = 'patch';
+
+    public function __construct(
+        public readonly string $importName,
+        public readonly string $url,
```
weaverryan: I think the `url` property may not be needed.

weaverryan (commented):
Looking good 👍  - nice job rebasing - next round of comments

src/Symfony/Component/AssetMapper/Command/ImportMapOutdatedCommand.php
```diff
@@ -0,0 +1,104 @@
...
+   <info>php %command.full_name% <packages></info>
+EOT
+            );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+        $packages = $input->getArgument('packages');
+        $packagesUpdateInfos = $this->updateChecker->getAvailableUpdates($packages);
```
smnandre, line 70: Should this be used in the update command too ?

src/Symfony/Component/AssetMapper/ImportMap/ImportMapUpdateChecker.php
maelanleborgne: Yes I think having this inside the `ImportMapEntry`  would make sense, thought that would mean being able to parse these infos in `ImportMapConfigReader::getEntries`.
I see that you removed `JspmResolver` in your PR. If we handle a single 'url format',  we can move this parsing method elsewhere, but if we need to have multiple url format depending on the resolver we could instead add a `PackageResolverInterface` dependency to `ImportMapConfigReader`, keeping the logic in the resolver and make the method public so we can call it and use the infos to create the entry.

weaverryan: Good news - I don't think this parsing is dependent on the resolver. The string we're parsing is something like `chart.js/auto`, which is a standard string - it's even what you would import, if you were in Node - e.g. `import 'chart.js/auto'`. 

So, I like your idea of putting this parsing logic into `ImportMapConfigReader::getEntries()`.

src/Symfony/Component/AssetMapper/Command/ImportMapOutdatedCommand.php
weaverryan: Can you explain more? Do you mean: in the `update` command, we would first use this function to determine *which* updates are possible and then loop over and make them? If so, that's a different way than we're doing it now - but I can't think of any big difference in final behavior.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapUpdateChecker.php
maelanleborgne: Yes you're right, the package name format should always stay consistent :smiley:
So just to make sure we are clear : when creating an `ImportMapEntry` we keep the `importName` unchanged (ie: name of the index in `importmap.php`) and we add 2 new properties `packageName` and `filePath` to the `ImportMapEntry` class ?

```
// ImportMapConfigReader.php
public function getEntries(): ImportMapEntries
{
    // ...
    [$packageName, $filePath] =  self::splitPackageNameAndFilePath($importName);

    $entries->add(new ImportMapEntry(
        $importName,
        // ...
        packageName: $packageName,
        filePath: $filePath,
    ));
    // ...
}
```

And that raises another question :
We may have the case where we have a package in different versions depending on the files in the importMap.
For instance
```
...
'bootstrap' => [
	'url' => 'https://cdn.jsdelivr.net/npm/bootstrap@5.3.1/+esm',
],
'bootstrap/dist/css/bootstrap.min.css' => [
	'url' => 'https://cdn.jsdelivr.net/npm/bootstrap@5.3.2/dist/css/bootstrap.min.css',
	'type' => 'css',
],
...
```
I can't really see why a user would need to do this, but it makes me wonder what should be the grouping rule : the packageName (only one line displayed by the command for 'bootstrap') or the importName (two lines, 'bootstrap' and 'bootstrap/dist/css/bootstrap.min.css')

weaverryan: Yes, this is what I’m thinking 👍. Though perhaps we should do this in a separate PR to keep things moving - that change will impact various parts of the code. 

For your last question, in the command, we should render the key used in the importmap (so the importName) so that users know exactly which entry is the problem. Then they could have multiple versions like you mentioned and we could accurately report that one has a problem but another does not.

weaverryan:
One more relevant PR just got merged- can you rebase? Let me know if you have any questions - hopefully that’ll simplify things, as remote packages ALWAYS have a version now.

src/Symfony/Component/AssetMapper/Command/ImportMapOutdatedCommand.php
smnandre: Yes that was the idea, but i'm not talking about "right now" ... but in a near future, as we already said :)

OskarStark:
Is this still WIP?

maelanleborgne:
> Is this still WIP?

Yes I still have some work to do now that #51786 has been merged.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapUpdateChecker.php
maelanleborgne: Alright I've rebased and made the changes as suggested. I can open a separate PR for the 'packageName' and 'FilePath' if you want.

src/Symfony/Bundle/FrameworkBundle/Resources/config/asset_mapper.php
```diff
@@ -179,6 +181,12 @@
             service('asset_mapper.importmap.config_reader'),
             service('http_client'),
         ])
+        ->set('asset_mapper.importmap.update_checker', ImportMapUpdateChecker::class)
+        ->args([
+            service('asset_mapper.importmap.config_reader'),
+            service('asset_mapper.importmap.resolver'),
```
weaverryan: This argument is gone now

src/Symfony/Component/AssetMapper/ImportMap/ImportMapUpdateChecker.php
```diff
@@ -0,0 +1,100 @@
...
+     * @return PackageUpdateInfo[]
+     */
+    public function getAvailableUpdates(array $packages = []): array
+    {
+        $entries = $this->importMapConfigReader->getEntries();
+        $updateInfos = [];
+        foreach ($entries as $entry) {
+            if (null === $entry->packageName) {
+                continue;
+            }
```
weaverryan, line 39: Also add:

```php
if (null === $entry->version) {
    continue;
}
```

src/Symfony/Component/AssetMapper/Command/ImportMapOutdatedCommand.php
```diff
@@ -0,0 +1,104 @@
...
+            $io->writeln(json_encode($displayData, \JSON_PRETTY_PRINT | \JSON_UNESCAPED_SLASHES));
+        } else {
+            foreach ($displayData as $datum) {
+                $io->writeln(sprintf(
+                    '  - <comment>%s</comment> (%s => <fg=%s>%s</>)',
+                    $datum['name'],
+                    $datum['current'],
+                    self::COLOR_MAPPING[$datum['latest-status']],
+                    $datum['latest'],
+                ));
```
weaverryan: This might look a bit better as a table - for example, the yarn outdated command:

Screenshot 2023-10-09 at 6 46 03 AM

(we don't need as many columns - e.g. we don't have a "Wanted" column, but otherwise, it's a nice model)

src/Symfony/Component/AssetMapper/Command/ImportMapOutdatedCommand.php
```diff
@@ -0,0 +1,104 @@
...
+            'latest' => $packageUpdateInfo->latestVersion,
+            'latest-status' => PackageUpdateInfo::UPDATE_TYPE_MAJOR === $packageUpdateInfo->updateType ? 'update-possible' : 'semver-safe-update',
+        ], array_values($packagesUpdateInfos));
+
+        if ('json' === $input->getOption('format')) {
+            $io->writeln(json_encode($displayData, \JSON_PRETTY_PRINT | \JSON_UNESCAPED_SLASHES));
+        } else {
+            foreach ($displayData as $datum) {
+                $io->writeln(sprintf(
+                    '  - <comment>%s</comment> (%s => <fg=%s>%s</>)',
```
weaverryan: Probably the package name color should match the COLOR_MAPPING that is used for this item.

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -159,7 +160,7 @@ public function downloadPackages(array $importMapEntries, callable $progressCall
         $responses = [];
 
         foreach ($importMapEntries as $package => $entry) {
-            [$packageName, $filePath] = self::splitPackageNameAndFilePath($entry->importName);
+            [$packageName, $filePath] = ImportMapConfigReader::splitPackageNameAndFilePath($entry->importName);
```
weaverryan: Since you've already added the two new properties to `ImportMapEntry`, let's use those here instead of splitting again.

```
$packageName = $entry->packageName;
$filePath = $entry->filePath
```

src/Symfony/Component/AssetMapper/ImportMap/ImportMapConfigReader.php
```diff
@@ -144,4 +148,18 @@ private function extractVersionFromLegacyUrl(string $url): ?string
...
+        $i = strpos($packageName, '/');
+
+        if ($i && (!str_starts_with($packageName, '@') || $i = strpos($packageName, '/', $i + 1))) {
+            // @vendor/package/filepath or package/filepath
+            $filePath = substr($packageName, $i);
+            $packageName = substr($packageName, 0, $i);
+        }
+
+        return [$packageName, $filePath];
+    }
```
weaverryan, line 164: We need to update `ImportMapConfigReaderTest` to look for and test for this new functionality

src/Symfony/Component/AssetMapper/ImportMap/ImportMapUpdateChecker.php
weaverryan: You have already (nicely) added the new 2 properties to `ImportMapEntry` 🥳 . So the only remaining things to do to accomplish this are to find all usages of `->importName` that need to be updated. Not all need to be updated - sometimes we DO want to reference the "import name" and other times we want to reference the "package name". I did a `git grep '\->importName'` to find them.

I thought about having you include them in THIS PR, but at least on instance (in `ImportMapAudit`) looks complex. So let's finish this PR with the 2 new properties, get it merged, then go find and update all of the `$entry->importName` items that need it.

weaverryan (commented):
A few minor, but important tweaks. Also, don't forget to check fabbot in the CI after you push for the little phpcs details :).

src/Symfony/Component/AssetMapper/Command/ImportMapRequireCommand.php
```diff
@@ -130,7 +130,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $messages = [$message];
 
         if (1 === \count($newPackages)) {
-            $messages[] = sprintf('Use the new package normally by importing "%s".', $newPackages[0]->importName);
+            $messages[] = sprintf('Use the new package normally by importing "%s".', $newPackages[0]->packageName);
```
weaverryan: This should stay as `importName`. They will need to `import 'foo/some/path'` where `foo/some/path` is the exact "import name" in importmap.php

src/Symfony/Component/AssetMapper/Command/ImportMapRequireCommand.php
```diff
@@ -119,7 +119,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $newPackages = $this->importMapManager->require($packages);
         if (1 === \count($newPackages)) {
             $newPackage = $newPackages[0];
-            $message = sprintf('Package "%s" added to importmap.php', $newPackage->importName);
+            $message = sprintf('Package "%s" added to importmap.php', $newPackage->packageName);
```
weaverryan: I'd keep this as importName also. Saying `Package "foo/path" added` is not quite accurate, but I think it's more meaningful than hiding the path.

src/Symfony/Component/AssetMapper/Tests/ImportMap/ImportMapManagerTest.php
```diff
@@ -1032,7 +1035,7 @@ private function mockDownloader(array $importMapEntries): void
             ->method('getDownloadedPath')
             ->willReturnCallback(function (string $packageName) use ($importMapEntries) {
                 foreach ($importMapEntries as $entry) {
-                    if ($entry->importName === $packageName) {
+                    if ($entry->packageName === $packageName) {
```
weaverryan: The argument to `getDownloadedPath()` is `$importName` (and that is correct: that's the "key" in the importmap that uniquely identifies each entry). So this, I think, should remain `->importName`. And the `string $packageName` above should change to `string $importName` for clarity.

maelanleborgne: Okay thank you for the advices, I wasn't quite sure of where it was relevant to make the change to packageName

src/Symfony/Component/AssetMapper/Command/ImportMapOutdatedCommand.php
```diff
@@ -0,0 +1,106 @@
...
+use Symfony\Component\AssetMapper\ImportMap\PackageUpdateInfo;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+#[AsCommand(name: 'importmap:outdated', description: 'Lists outdated JavaScript packages and their latest versions')]
```
weaverryan:
```suggestion
#[AsCommand(name: 'importmap:outdated', description: 'List outdated JavaScript packages and their latest versions')]
```

src/Symfony/Component/AssetMapper/Command/ImportMapOutdatedCommand.php
```diff
@@ -0,0 +1,106 @@
...
+final class ImportMapOutdatedCommand extends Command
+{
+    private const COLOR_MAPPING = [
+        'update-possible' => 'yellow',
+        'semver-safe-update' => 'red',
+    ];
+
+    public function __construct(
+        protected readonly ImportMapUpdateChecker $updateChecker,
```
weaverryan: Can be `private`, right?

src/Symfony/Component/AssetMapper/Command/ImportMapOutdatedCommand.php
```diff
@@ -0,0 +1,106 @@
...
+        parent::__construct();
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->addArgument(
+                name: 'packages',
+                mode: InputArgument::IS_ARRAY | InputArgument::OPTIONAL,
+                description: 'List packages\' names',
```
weaverryan:
```suggestion
                description: 'A list of packages to check,
```

src/Symfony/Component/AssetMapper/ImportMap/ImportMapUpdateChecker.php
```diff
@@ -0,0 +1,100 @@
...
+        $updateInfos = [];
+        foreach ($entries as $entry) {
+            if (null === $entry->packageName || null === $entry->version) {
+                continue;
+            }
+            if (\count($packages) && !\in_array($entry->packageName, $packages, true)) {
+                continue;
+            }
+
+            $updateInfos[$entry->importName] = $this->getAvailableUpdateForEntry($entry);
```
weaverryan: It makes the code a bit trickier, but we should try to make the requests in parallel - `JsDelivrEsmResolver` has an example of this. The flow would probably be:

A) Loop over `$entries`. Instead of an `$updateInfos` array set to the final project, we would probably have a `$responses` array set to the individual `Response` objects from making the HTTPClient request.

B) Then, do another `foreach` over `$responses` and process each one (get the latest version, determine the update type, etc).

This will allow all of the responses to start at once, and *then* we start processing them.

weaverryan (commented):
Ok, very close now!

weaverryan (approved):
Test failures all look unrelated. 👍

src/Symfony/Component/AssetMapper/Tests/ImportMap/ImportMapUpdateCheckerTest.php
```diff
@@ -0,0 +1,205 @@
...
+                importName: 'json5',
+                version: '1.0.0',
+                packageName: 'json5',
+            ),
+            'bootstrap' => new ImportMapEntry(
+                importName: 'bootstrap',
+                version: '5.3.1',
+                packageName: 'bootstrap',
+            ),
+            // Css will always be considered up to date
```
squrious: Why CSS should be always considered up-to-date? At L95 it's expected to have `updateType: 'patch'`

maelanleborgne: Nice catch, that's a comment that should have been deleted when I made the change to allow css to be managed. Thanks

src/Symfony/Component/AssetMapper/ImportMap/ImportMapUpdateChecker.php
maelanleborgne: Thanks for your time and advices @weaverryan !

fabpot (approved):

fabpot:
Thank you @maelanleborgne.

weaverryan:
Congrats and thanks for the nice work @maelanleborgne! No expectations, but if you'd like to work on a PR to find & change the `->importName` spots that should be updated, I'd appreciated it - one less thing on my list :) 

Edit: I found one issue. There are 3 places in `ImportMapManager` (that I forgot about) where we create new `ImportMapEntry` objects... and some or all of these need to have the `packageName` and `filePath` stuff added to it. Currently, because this is missing, `importmap:require some-package` fails (because eventually an `ImportMapEntry` is passed to `JsDelivrEsmResolver::downloadPackages()`, which is missing `$entry->packageName`, so the URL fails to build correctly there.

maelanleborgne:
> Edit: I found one issue. There are 3 places in `ImportMapManager` (that I forgot about) where we create new `ImportMapEntry` objects... and some or all of these need to have the `packageName` and `filePath` stuff added to it. Currently, because this is missing, `importmap:require some-package` fails (because eventually an `ImportMapEntry` is passed to `JsDelivrEsmResolver::downloadPackages()`, which is missing `$entry->packageName`, so the URL fails to build correctly there.

@weaverryan I've checked and it seems like we need to make a change only on this occurence :
https://github.com/symfony/symfony/blob/fc12885833f68a8d8e68dae89ba59dbbfc6c4918/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php#L299
The other two are local assets that should not cause trouble. Should I open a PR for the fix or is there an already opened one that we could fit this in ?

PR #51847, opened 2023-10-04
[AssetMapper] Allowing for files to be written to some non-local location

### LINKED ISSUE #50221: [AssetMapper] Add a Filesystem abstraction
### Description

[Right now the AssetMapper command directly writes into the public folder. ](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/AssetMapper/Command/AssetMapperCompileCommand.php#L88)

It would be nice if there would be an abstraction to the Filesystem so I could directly use Flysystem and upload it to S3/whatever. A workaround could be to let it copy to an temporary directory and from there to S3 (assuming the public folder is read-only). 

### Example

_No response_

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #50221
| License       | MIT

Hi!

An attempt at making AssetMapper flexible enough to support non-local filesystems. The "hook point" would be that you could replace the `asset_mapper.local_public_assets_filesystem` service with your own that implements `PublicAssetsFilesystemInterface`. I'm not worried about making the hook point *super* user-friendly: I just want the system to support this now, as trying to add this later (when we need to protect BC) will be harder.

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/Command/AssetMapperCompileCommand.php
+++ b/src/Symfony/Component/AssetMapper/Command/AssetMapperCompileCommand.php
@@ -13,16 +13,15 @@ namespace Symfony\Component\AssetMapper\Command;
 
 use Symfony\Component\AssetMapper\AssetMapper;
 use Symfony\Component\AssetMapper\AssetMapperInterface;
+use Symfony\Component\AssetMapper\CompiledAssetMapperConfigReader;
 use Symfony\Component\AssetMapper\Event\PreAssetsCompileEvent;
 use Symfony\Component\AssetMapper\ImportMap\ImportMapGenerator;
-use Symfony\Component\AssetMapper\Path\PublicAssetsPathResolverInterface;
+use Symfony\Component\AssetMapper\Path\PublicAssetsFilesystemInterface;
 use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
-use Symfony\Component\Console\Exception\InvalidArgumentException;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
-use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
 /**
@@ -36,12 +35,11 @@ use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 final class AssetMapperCompileCommand extends Command
 {
     public function __construct(
-        private readonly PublicAssetsPathResolverInterface $publicAssetsPathResolver,
+        private readonly CompiledAssetMapperConfigReader $compiledConfigReader,
         private readonly AssetMapperInterface $assetMapper,
         private readonly ImportMapGenerator $importMapGenerator,
-        private readonly Filesystem $filesystem,
+        private readonly PublicAssetsFilesystemInterface $assetsFilesystem,
         private readonly string $projectDir,
-        private readonly string $publicDirName,
         private readonly bool $isDebug,
         private readonly ?EventDispatcherInterface $eventDispatcher = null,
     ) {
@@ -51,7 +49,6 @@ final class AssetMapperCompileCommand extends Command
     protected function configure(): void
     {
         $this
-            ->addOption('clean', null, null, 'Whether to clean the public directory before compiling assets')
             ->setHelp(<<<'EOT'
 The <info>%command.name%</info> command compiles and dumps all the assets in
 the asset mapper into the final public directory (usually <comment>public/assets</comment>).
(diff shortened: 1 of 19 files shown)


src/Symfony/Component/AssetMapper/CompiledAssetMapperConfigReader.php
```diff
@@ -0,0 +1,50 @@
...
+ */
+class CompiledAssetMapperConfigReader
+{
+    public function __construct(private readonly string $directory)
+    {
+    }
+
+    public function configExists(string $filename): bool
+    {
+        return is_file($this->directory.'/'.$filename);
```
shyim: should we make use of `\Symfony\Component\Filesystem\Path::join` for paths joining?

src/Symfony/Component/AssetMapper/CompiledAssetMapperConfigReader.php
```diff
@@ -0,0 +1,50 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\AssetMapper;
+
+/**
+ * Reads and writes compiled configuration files for asset mapper.
+ */
+class CompiledAssetMapperConfigReader
```
shyim, line 19: we could rely on symfony fs, so we are sure that there will be a exception when the file cannot be found (composer.json anyway requires symfony fs)

shyim:
looks pretty good. I am right now in Berlin for SymfonyLive. I will check next week and build something with flysystem :).

But arent the most things breaking changes? 🤔

chalasr:
> But arent the most things breaking changes? 🤔

Indeed, but the component has been released as [experimental](https://symfony.com/doc/current/contributing/code/experimental.html) to allow for such changes if needed.

weaverryan:
> Indeed, but the component has been released as [experimental](https://symfony.com/doc/current/contributing/code/experimental.html) to allow for such changes if needed.

That's also why I'd like to get this in NOW for 6.4 - the component will be stable in 6.4, so making a change like this will be harder. We're already dipping a little beyond the 6.4 feature-freeze date :)

src/Symfony/Component/AssetMapper/Command/AssetMapperCompileCommand.php
```diff
@@ -130,21 +102,14 @@ private function shortenPath(string $path): string
         return str_replace($this->projectDir.'/', '', $path);
     }
 
-    private function createManifestAndWriteFiles(SymfonyStyle $io, string $publicDir): array
+    private function createManifestAndWriteFiles(SymfonyStyle $io): array
     {
         $allAssets = $this->assetMapper->allAssets();
 
-        $io->comment(sprintf('Compiling assets to <info>%s%s</info>', $publicDir, $this->publicAssetsPathResolver->resolvePublicPath('')));
+        $io->comment(sprintf('Compiling and writing asset files to <info>%s</info>', $this->assetsFilesystem->getDestinationPathString()));
```
smnandre: $this->shorthenPath($this->assetsFilesystem->getDEstinationPathString()) ?

src/Symfony/Component/AssetMapper/Event/PreAssetsCompileEvent.php
```diff
@@ -21,20 +21,13 @@
...
+    public function __construct(OutputInterface $output)
     {
-        $this->outputDir = $outputDir;
         $this->output = $output;
     }
 
-    public function getOutputDir(): string
-    {
-        return $this->outputDir;
-    }
```
weaverryan, line 36: I created this event - and just included `$outputDir` because I had it available. Now that it's not readily available, I don't think it's a big deal to not include it. My use-case for this event is to run some code that pre-compiles some internal files (not into the output directory) so that when the true compilation process happens, those files are ready and available.

weaverryan:
@shyim While I think this PR is valid and should be merged, since you created the original issue, if you were able to validate that this gives you what you need, that would help :)

shyim (approved):
I exchanged the interface with flysystem and it works fine for me. Thanks for your awesome work and sorry for my late reply.

src/Symfony/Component/AssetMapper/CompiledAssetMapperConfigReader.php
weaverryan: What do you mean exactly? Rely on the `Filesystem` class? To do what exactly? Like `Filesystem::dumpFile()`? If so, I'll defer to others. I've used simple `mkdir` and `file_put_contents()` other places - but I don't care either way - whatever is more "normal" in core :)

shyim: the native php functions like `mkdir` don't throw an error if a folder cannot be created. idk what's the default way is. I just saw in the composer.json that `symfony/filesystem` is required, so I thought just replace in this class the php functions with the symfony class as it does the error checking for you :)

weaverryan:
~~Test failures in AssetMapper - but it looks like it's coming from #50734 - the `$this` variable in~~ https://github.com/symfony/symfony/blob/934aea0b32ed6caea3cf5e1e4bb4cbc05417b424/src/Symfony/Component/ErrorHandler/Resources/views/trace.html.php#L14 ~~is an instance of `HtmlErrorRenderer`~~

Rebased and those tests are fixed.

src/Symfony/Component/AssetMapper/Path/LocalPublicAssetsFilesystem.php
```diff
@@ -0,0 +1,35 @@
...
+        $targetPath = $this->publicDir.'/'.ltrim($path, '/');
+
+        if (!is_dir($dir = \dirname($targetPath))) {
+            @mkdir($dir, 0777, true);
+        }
+
+        file_put_contents($targetPath, $contents);
+    }
+
+    public function getDestinationPathString(): string
```
nicolas-grekas: why `String`?

nicolas-grekas (approved):
LGTM, just one minor question :)

src/Symfony/Component/AssetMapper/Path/LocalPublicAssetsFilesystem.php
weaverryan: Good question - I removed it :)

fabpot:
@weaverryan Can you rebase this one?

src/Symfony/Component/AssetMapper/ImportMap/ImportMapGenerator.php
```diff
@@ -202,13 +200,6 @@ private function addImplicitEntries(ImportMapEntry $entry, array $currentImportE
         return $currentImportEntries;
     }
 
-    private function findRootImportMapEntry(string $moduleName): ?ImportMapEntry
-    {
-        $entries = $this->importMapConfigReader->getEntries();
-
-        return $entries->has($moduleName) ? $entries->get($moduleName) : null;
-    }
```
weaverryan, line 210: From an earlier PR - method was no longer used

weaverryan (commented):
Rebased and re-tested on a project locally. Should be good to go!

fabpot:
Thank you @weaverryan.

PR #51848, opened 2023-10-04
[Messenger] Resend failed retries back to failure transport

### LINKED ISSUE #51815: [Messenger] Requeue failed retries
### Description

Given `messenger:failed:retry`, it seems sane if a retry failure is put back on the failure queue, till either manually deleted or the retry succeeds.

I tried using a recursive failure transport:

```yaml
failed:
  dsn: ...
  failure_transport: failed
```

but it didnt make any changes.

### Example

_No response_

ro0NL:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 / 5.4
| Bug fix?      | yes
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #51815 
| License       | MIT

i want to argue this is a bugfix :)

i can confirm with this fix, you can indefinitely retry failures, till they succeed, are deleted manually, or are handled explicitly by other means. What should NOT happen is systems implicitly rejecting/acking messages without any confirmation whatsoever.

cc @fabpot, @weaverryan

### DIFF
--- a/src/Symfony/Component/Messenger/EventListener/SendFailedMessageToFailureTransportListener.php
+++ b/src/Symfony/Component/Messenger/EventListener/SendFailedMessageToFailureTransportListener.php
@@ -52,11 +52,6 @@ class SendFailedMessageToFailureTransportListener implements EventSubscriberInte
 
         $envelope = $event->getEnvelope();
 
-        // avoid re-sending to the failed sender
-        if (null !== $envelope->last(SentToFailureTransportStamp::class)) {
-            return;
-        }
-
         $envelope = $envelope->with(
             new SentToFailureTransportStamp($event->getReceiverName()),
             new DelayStamp(0),


weaverryan:
That code goes all the way back to the original PR for the failure transport system. I can't remember why I had this, however. Since we're dealing with the failure transport, this seems like a safe area, but I'd still probably do it for 6.4.

src/Symfony/Component/Messenger/EventListener/SendFailedMessageToFailureTransportListener.php
```diff
@@ -52,11 +52,6 @@ public function onMessageFailed(WorkerMessageFailedEvent $event)
 
         $envelope = $event->getEnvelope();
 
-        // avoid re-sending to the failed sender
-        if (null !== $envelope->last(SentToFailureTransportStamp::class)) {
-            return;
-        }
```
weaverryan, line 58: If we have the `SentToFailureTransportStamp`, should we avoid adding the stamps below?

ro0NL: i thought it was a nice semantic to keep for future purposes maybe

weaverryan: I can't really think of a downside, other than more and more stamps stacked on each other, but these stamps are tiny...

weaverryan (approved):

src/Symfony/Component/Messenger/EventListener/SendFailedMessageToFailureTransportListener.php
ro0NL: actually it remains used: https://github.com/symfony/symfony/blob/9767270b34e97d1bbf340f9ab3358a0178dccb8a/src/Symfony/Component/Messenger/Command/AbstractFailedMessagesCommand.php#L119

ro0NL: oh sorry, i misunderstood. We don't need to stack them per se, i agree

edit: i think i'd keep them for now to keep the flow transparent

fabpot (approved):

fabpot:
Thank you @ro0NL.

PR #51849, opened 2023-10-04
[AssetMapper] Warn of missing or incompat dependencies

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | None
| License       | MIT

Hi!

The aim of the `importmap.php` system is to be a simple way to manage your JS dependencies. But to make it robust enough for production use, it does need a few things - like the `importmap:audit` command in #51650. This PR adds a check, during `importmap:require` and `importmap:update`, that reports any missing dependencies or dependencies with invalid versions. This is necessary so that, if package `A` requires package `B`, their versions don't "drift" over time without you being aware (e.g. you update package `A` to v3 but keep package `B` at v1, even though v3 of `A` requires v2 of `B`).

Screenshot 2023-10-04 at 2 44 04 PM

Built on top of #51786.

Cheers!

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/asset_mapper.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/asset_mapper.php
@@ -34,6 +34,7 @@ use Symfony\Component\AssetMapper\ImportMap\ImportMapConfigReader;
 use Symfony\Component\AssetMapper\ImportMap\ImportMapManager;
 use Symfony\Component\AssetMapper\ImportMap\ImportMapRenderer;
 use Symfony\Component\AssetMapper\ImportMap\ImportMapUpdateChecker;
+use Symfony\Component\AssetMapper\ImportMap\ImportMapVersionChecker;
 use Symfony\Component\AssetMapper\ImportMap\RemotePackageDownloader;
 use Symfony\Component\AssetMapper\ImportMap\RemotePackageStorage;
 use Symfony\Component\AssetMapper\ImportMap\Resolver\JsDelivrEsmResolver;
@@ -171,6 +172,12 @@ return static function (ContainerConfigurator $container) {
                 service('asset_mapper.importmap.resolver'),
             ])
 
+        ->set('asset_mapper.importmap.version_checker', ImportMapVersionChecker::class)
+            ->args([
+                service('asset_mapper.importmap.config_reader'),
+                service('asset_mapper.importmap.remote_package_downloader'),
+            ])
+
         ->set('asset_mapper.importmap.resolver', JsDelivrEsmResolver::class)
             ->args([service('http_client')])
 
@@ -199,6 +206,7 @@ return static function (ContainerConfigurator $container) {
             ->args([
                 service('asset_mapper.importmap.manager'),
                 param('kernel.project_dir'),
+                service('asset_mapper.importmap.version_checker'),
             ])
             ->tag('console.command')
 
@@ -207,7 +215,10 @@ return static function (ContainerConfigurator $container) {
             ->tag('console.command')
 
         ->set('asset_mapper.importmap.command.update', ImportMapUpdateCommand::class)
-            ->args([service('asset_mapper.importmap.manager')])
+            ->args([
+                service('asset_mapper.importmap.manager'),
+                service('asset_mapper.importmap.version_checker'),
+            ])
             ->tag('console.command')
 
         ->set('asset_mapper.importmap.command.install', ImportMapInstallCommand::class)
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapVersionChecker.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapVersionChecker.php
@@ -1,0 +1,179 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\AssetMapper\ImportMap;
+
+use Composer\Semver\Semver;
+use Symfony\Component\AssetMapper\Exception\RuntimeException;
+use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Contracts\HttpClient\Exception\HttpExceptionInterface;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+class ImportMapVersionChecker
+{
+    private const PACKAGE_METADATA_PATTERN = 'https://registry.npmjs.org/%package%/%version%';
+
+    private HttpClientInterface $httpClient;
+
+    public function __construct(
+        private ImportMapConfigReader $importMapConfigReader,
+        private RemotePackageDownloader $packageDownloader,
+        HttpClientInterface $httpClient = null,
+    ) {
+        $this->httpClient = $httpClient ?? HttpClient::create();
+    }
+
+    /**
+     * @return PackageVersionProblem[]
+     */
+    public function checkVersions(): array
+    {
+        $entries = $this->importMapConfigReader->getEntries();
+
+        $packages = [];
+        foreach ($entries as $entry) {
+            if (!$entry->isRemotePackage()) {
+                continue;
+            }
+
+            $dependencies = $this->packageDownloader->getDependencies($entry->importName);
+            if (!$dependencies) {
+                continue;
+            }
+
+            $packageName = $entry->getPackageName();
+
+            $url = str_replace(
+                ['%package%', '%version%'],
+                [$packageName, $entry->version],
+                self::PACKAGE_METADATA_PATTERN
+            );
+            $packages[$packageName] = [
+                $this->httpClient->request('GET', $url),
+                $dependencies,
+            ];
+        }
+
+        $errors = [];
+        $problems = [];
+        foreach ($packages as $packageName => [$response, $dependencies]) {
+            if (200 !== $response->getStatusCode()) {
+                $errors[] = [$packageName, $response];
+                continue;
+            }
+
+            $data = json_decode($response->getContent(), true);
+            // dependencies seem to be found in both places
+            $packageDependencies = array_merge(
+                $data['dependencies'] ?? [],
+                $data['peerDependencies'] ?? []
+            );
+
+            foreach ($dependencies as $dependencyName) {
+                // dependency is not in the import map
+                if (!$entries->has($dependencyName)) {
+                    $dependencyVersionConstraint = $packageDependencies[$dependencyName] ?? 'unknown';
+                    $problems[] = new PackageVersionProblem($packageName, $dependencyName, $dependencyVersionConstraint, null);
+
+                    continue;
+                }
+
+                $dependencyPackageName = $entries->get($dependencyName)->getPackageName();
+                $dependencyVersionConstraint = $packageDependencies[$dependencyPackageName] ?? null;
+
+                if (null === $dependencyVersionConstraint) {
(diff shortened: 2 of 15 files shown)


ahmedyakoubi:
why not installing the missing dependency automatically instead of doing it manually?

weaverryan:
> why not installing the missing dependency automatically instead of doing it manually?

They *will* be installed automatically when you first require the package - e.g. `importmap:require bootstrap` will also grab `@popper/core`. So, if you ever see the `but it is not in the importmap.php` message from this PR, something has gone wrong or you've removed something by hand :)

src/Symfony/Component/AssetMapper/ImportMap/Resolver/PackageResolverInterface.php
```diff
@@ -35,7 +35,7 @@ public function resolvePackages(array $packagesToRequire): array;
      *
      * @param array<string, ImportMapEntry> $importMapEntries
      *
-     * @return array<string, string>
+     * @return array<string, array{content: string, dependencies: string[]}>
```
weaverryan, line 40: Not ideal... not sure if it merits a new object - as this is entirely internal, unless someone makes their own package resolver.

src/Symfony/Component/AssetMapper/composer.json
```diff
@@ -17,6 +17,7 @@
     ],
     "require": {
         "php": ">=8.1",
+        "composer/semver": "^3.0",
```
weaverryan, line 20: New dependency - composer seems like a quite safe one. And I don't want to make this feature opt-in: all users should see warnings about version incompatibilities.

src/Symfony/Component/AssetMapper/Tests/ImportMap/ImportMapVersionChecker.php
```diff
@@ -0,0 +1,130 @@
...
+
+    private function extractPackageNameFromImport(string $importName): string
+    {
+        // TODO: remove duplication from JsDelivrEsmResolver?
+        $i = strpos($importName, '/');
+        if ($i && (!str_starts_with($importName, '@') || $i = strpos($importName, '/', $i + 1))) {
+            return substr($importName, 0, $i);
+        }
+
+        return $importName;
```
weaverryan: A bit of duplication with `JsDelivrEsmResolver`, but we have a plan to centralize this in a future PR.

src/Symfony/Bundle/FrameworkBundle/Resources/config/asset_mapper.php
```diff
@@ -162,6 +163,12 @@
                 abstract_arg('vendor directory'),
             ])
 
+        ->set('asset_mapper.importmap.version_checker', ImportMapVersionChecker::class)
```
fabpot, line 175: We cannot rely on a class in the `Tests` namespace, this should be moved elsewhere.

fabpot (changes requested):

src/Symfony/Bundle/FrameworkBundle/Resources/config/asset_mapper.php
weaverryan: Thank you - accidental, but I've done that twice lately 🙃

src/Symfony/Component/AssetMapper/composer.json
fabpot: It should be added to the main `composer.json` as well, right? Or maybe we don't have tests that cover that?

weaverryan: Guilty - a lot moving pieces - thanks for catching these things.

src/Symfony/Component/AssetMapper/ImportMap/PackageVersionProblem.php
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Symfony\Component\AssetMapper\ImportMap;
+
+final class PackageVersionProblem
+{
+    public function __construct(
+        public string $packageName,
+        public string $dependencyPackageName,
+        public string $requiredVersionConstraint,
+        public ?string $installedVersion
```
fabpot: Does it make sense to make them `readonly`?

```suggestion
    public function __construct(
        public readonly string $packageName,
        public readonly string $dependencyPackageName,
        public readonly string $requiredVersionConstraint,
        public readonly ?string $installedVersion
```

src/Symfony/Component/AssetMapper/ImportMap/ImportMapVersionChecker.php
```diff
@@ -0,0 +1,131 @@
...
+        return $problems;
+    }
+
+    private function isVersionSatisfied(string $versionConstraint, ?string $version): bool
+    {
+        if (null === $version) {
+            return false;
+        }
+
+        return Semver::satisfies($version, $versionConstraint);
```
fabpot: Could easily be a one-line:

```suggestion
        return $version ? Semver::satisfies($version, $versionConstraint) : false;
```

src/Symfony/Component/AssetMapper/ImportMap/ImportMapVersionChecker.php
```diff
@@ -0,0 +1,131 @@
...
+            $packageDependencies = array_merge(
+                $data['dependencies'] ?? [],
+                $data['peerDependencies'] ?? []
+            );
+
+            foreach ($dependencies as $dependencyName) {
+                $dependencyVersionConstraint = $packageDependencies[$dependencyName] ?? null;
+
+                if (null === $dependencyVersionConstraint) {
+                    // odd that this would happen
```
fabpot: It makes me nervous :joy:

weaverryan: Haha, good point - that comment was a bit ominous. It would be an unexpected but not dangerous situation where we see that a vendor file has `import 'foo'`, but then when we check the metadata for that package, it doesn't list `foo` as a dependency at all. I have no idea if this will ever happen under normal circumstances - I can't think of a way (some misconfigured package ?). 

But, to be safe, I've changed this to also issue a warning (like the other warnings) so the user is aware of it.

src/Symfony/Component/AssetMapper/composer.json
stof, line 20: Be careful about that. The constraints are not 100% compatible between the npm and composer ecosystems.

weaverryan: Hmm, is there an alternative? Checking into this quickly, differences I found:

* npm has a `1.0.0 - 2.0.0` syntax.
* npm seems to have a `5.4.x` syntax which means `5.4.*`.

I think this feature is pretty important, without it, it gets easy for your dependencies to "drift" out of sync (e.g. you install `boostrap` today and we also give you `@popper/core`, then you upgrade bootstrap but forget to upgrade popper and have no warning). So perhaps, IF we run into problems with odd, npm-specific dependency things we can handle those manually? Or maybe we can handle the above 2 items right now manually.

EDIT: I'll add a bit of pre-processing and tests to cover the edge-case differences.

stof: IIRC, the `~` operator does not work the same either

weaverryan: I've added a simple conversion system using all of the differences I could find. There are some edge-case things that can't be converted - e.g. `stable` or `https://example.com/file.js`, but I think that's ok. This is just a system to warn users of potential problems (like the peer dependency warning in npm). And if we're missing some real-world cases, we can find out about them and patch them. The "conversion" code is fairly minimal, as the differences don't appear to be too many. 

Thanks for catching this!

weaverryan:
~~Failing tests coming from #50734~~

weaverryan:
@stof I'd love your thoughts on this PR now. A few notes:

A) I do think this PR is important, so that your dependency versions don't drift away from each other. We're not building a full-blown package manager, but this seems like a pragmatic precaution.

B) I think i've covered the npm-specific version constraint differences nicely

C) In practice, it seems that when some packages are published, their dependencies in `package.json` change to *exact* versions. An example is vue - notice `@vue/runtime-dom` in https://registry.npmjs.org/vue/3.3.0 is listed at exactly 3.3.0. This isn't a problem, it just looks a bit odd and I wanted to mention it (and thought you might have even more insight).

But, I do still think this PR is ready to go

stof:
For C, it is a choice of each package maintainer.
Some JS maintainers prefer to require exact versions of their dependencies. Unlike in composer, this does not create a dependency hell as the npm ecosystem does not require a flat dependency list in the installed dependencies with a single version of each package (Composer requires that for PHP package because PHP defines all classes in the same global scope so you cannot have 2 different version of the same package loaded. But JS works with a concept of modules exporting things, not with a shared naming scope).
There are mostly 2 cases there:
- several related packages from the same maintainer having a shared release cycle (often in a mono-repo): exact version constraint avoids the need to deal with the case of having an updated version of one package while others are not updated. Especially useful when the public API of one package actually relies on the other package
- dependencies on unrelated packages. For those, I don't have an argument in favor of it (as I'm actually in the other side there). The big drawback is that this requires actively monitoring releases of the dependencies to update them even when they release patch versions instead of relying on semver)

Note that the fact that the JS ecosystem does not require flat dependencies means that it might be impossible to resolve dependencies in a flat way (and doing that requires a composer-like solver anyway and would fail in cases of a dependency hell). And this means that your installer should be able to support the case of having several versions of the same package to respect constraints otherwise it might cause issues.

weaverryan:
That's exactly the insight I was hoping for! 🙏

> Note that the fact that the JS ecosystem does not require flat dependencies means that it might be impossible to resolve dependencies in a flat way. And this means that your installer should be able to support the case of having several versions of the same package to respect constraints otherwise it might cause issues

I think this is possible, via the importmap [scopes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap#scoped_module_specifier_maps). But I'd like to see if we actually need this. The consequence is that, right now, the system will work in the simple, "flat dependencies" kind of way. And I think that's fine. This PR would at least help users discover this potential problem more readily, then we can chat in the future about solving it.

src/Symfony/Component/AssetMapper/composer.json
nicolas-grekas, line 20: ~What about making this an optional dep and throw a useful LogicException in the command when used?~

nvm, it's not an optional feat

src/Symfony/Component/AssetMapper/ImportMap/ImportMapVersionChecker.php
```diff
@@ -0,0 +1,177 @@
...
+     * Converts npm-specific version constraints to composer-style.
+     *
+     * @internal
+     */
+    public static function convertNpmConstraint(string $versionConstraint): ?string
+    {
+        // special npm constraint that don't translate to composer
+        if (\in_array($versionConstraint, ['latest', 'next'])
+            || preg_match('/^(git|http|file):/', $versionConstraint)
+            || str_contains($versionConstraint, '/')) {  // GitHub shorthand like user/repo
```
nicolas-grekas:
```suggestion
            || str_contains($versionConstraint, '/')
        ) {  // GitHub shorthand like user/repo
```

nicolas-grekas:
Thank you @weaverryan.

src/Symfony/Component/AssetMapper/composer.json
weaverryan: I thought about that... but we don't use it in a specific command, we run this in the background when using `importmap:require` or `importmap:update`. So in practice, it WILL be needed. And if we just didn't show the warning if this wasn't installed, then nobody would have it installed and it wouldn't do anything. I couldn't find a nice way to avoid it :)

PR #51871, opened 2023-10-06
Fix typo in method resolvePackages

smnandre:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

The wrong var is used in the loop

Reproduce bugs:

```php
php bin/console importmap:require bootstrap
php bin/console importmap:install
php bin/console importmap:update
```

6.4 targetted because the code with bug was just commited a couple of days ago

### DIFF
--- a/src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
@@ -77,7 +77,7 @@ final class JsDelivrEsmResolver implements PackageResolverInterface
             $requiredPackages[$i][4] = $version;
 
             if (!$filePath) {
-                $cssEntrypointResponses[$options->packageName] = $this->httpClient->request('GET', sprintf(self::URL_PATTERN_ENTRYPOINT, $packageName, $version));
+                $cssEntrypointResponses[$packageName] = $this->httpClient->request('GET', sprintf(self::URL_PATTERN_ENTRYPOINT, $packageName, $version));
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -139,7 +139,7 @@ public function resolvePackages(array $packagesToRequire): array
                 continue;
             }
 
-            $packagesToRequire[] = new PackageRequireOptions($packageName.$cssFile, $version, $options->download);
+            $packagesToRequire[] = new PackageRequireOptions($package.$cssFile, $version, $options->download);
```
weaverryan: Line 85 should also be `$packageName` I think (for the array key). Because `$options->packageName` could actually be a package + path - like `chart.js/auto` . It doesn't matter in practice, because we only request the CSS entrypoint when installing a pure package (with no path), but we might as well fix that up too.

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -72,7 +72,7 @@ public function resolvePackages(array $packagesToRequire): array
         $cssEntrypointResponses = [];
         foreach ($requiredPackages as $i => [$options, $response, $packageName, $filePath]) {
             if (200 !== $response->getStatusCode()) {
-                $errors[] = [$options->packageName, $response];
+                $errors[] = [$packageName, $response];
```
weaverryan: I think this should stay the same. It’s only used in the error. And there, we’d want to mention the entire “package name” that the user has in their importmap.

The “package name” that is a package+path vs the “package name” that is actually the package name is a lot the root of the confusion. This is something we’re discussing on another PR to clarify.

weaverryan (approved):

fabpot (approved):

fabpot:
Thank you @smnandre.

PR #51872, opened 2023-10-06
[Mime] Fix memory leak

### LINKED ISSUE #51764: Memory Leak in Symfony Mailer when sending many attachments of >5MB

fabpot:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #51764
| License       | MIT

Not sure why I thought it would have been a good idea back then, but after some investigation, I think it's safe to remove this code.

### DIFF
--- a/src/Symfony/Component/Mime/RawMessage.php
+++ b/src/Symfony/Component/Mime/RawMessage.php
@@ -33,11 +33,8 @@ class RawMessage implements \Serializable
         if (\is_string($this->message)) {
             return $this->message;
         }
-        if ($this->message instanceof \Traversable) {
-            $this->message = iterator_to_array($this->message, false);
-        }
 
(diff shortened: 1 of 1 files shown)


stof:
IIRC, this was done to support the case of passing a Generator to produce the message, as this is an iterator that can be used only once.

stof:
And tests are telling me I'm right as we have a test that covered this case.

fabpot:
Yep, that was the use case I had in mind... but my tests were passing... because I was looking at the wrong directory. Whoops.

PR #51874, opened 2023-10-06
[Mime] Fix memory leak

### LINKED ISSUE #51764: Memory Leak in Symfony Mailer when sending many attachments of >5MB

fabpot:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #51764
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Mime/RawMessage.php
+++ b/src/Symfony/Component/Mime/RawMessage.php
@@ -30,11 +30,13 @@ class RawMessage
         if (\is_string($this->message)) {
             return $this->message;
         }
-        if ($this->message instanceof \Traversable) {
-            $this->message = iterator_to_array($this->message, false);
+
+        $message = '';
+        foreach ($this->message as $chunk) {
+            $message .= $chunk;
         }
 
-        return $this->message = implode('', $this->message);
+        return $this->message = $message;
     }
 
     public function toIterable(): iterable
@@ -45,12 +47,9 @@ class RawMessage
             return;
         }
 
-        $message = '';
         foreach ($this->message as $chunk) {
-            $message .= $chunk;
             yield $chunk;
         }
-        $this->message = $message;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/Mime/RawMessage.php
```diff
@@ -48,12 +46,19 @@ public function toIterable(): iterable
             return;
         }
 
-        $message = '';
-        foreach ($this->message as $chunk) {
-            $message .= $chunk;
-            yield $chunk;
+        if ($this->message instanceof \Generator) {
```
fabpot: If the message is a generator, we can only use it once, so we need to store the result.

fabpot:
Fixing it is not possible in 5.4.
I propose to deprecate the possibility of sending a message with a closed generator in 6.4 and throw in 7.0. That way, we don't need to store the generator result and that avoids the memory leak.

stof:
@fabpot the issue is not just about sending a message with a closed generator. If you try to both access the string content and serialize the object, this will make 2 usages of the iterator, with the RawMessage being the one closing the generator in the first usage.

fabpot:
@stof Sure, but that's not something people will do. What would you suggest?

chalasr:
Isn't patching only `toIterable()` enough to fix the issue? 
```diff
diff --git a/src/Symfony/Component/Mime/RawMessage.php b/src/Symfony/Component/Mime/RawMessage.php
index d2a311daeb..14223d3d7f 100644
--- a/src/Symfony/Component/Mime/RawMessage.php
+++ b/src/Symfony/Component/Mime/RawMessage.php
@@ -48,12 +48,9 @@ class RawMessage implements \Serializable
             return;
         }
 
-        $message = '';
         foreach ($this->message as $chunk) {
-            $message .= $chunk;
             yield $chunk;
         }
-        $this->message = $message;
     }
 
     /**
```
Transforming the generator to an array in `toString()` seems unavoidable, unless we forbid calling toString() more than once which sounds worse.

PR #51880, opened 2023-10-06
[AssetMapper] Fix entrypoint scripts are not preloaded

smnandre:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no 
| License       | MIT

Preload the "rendered" importmap entryoints (ex: "app.js") 
(expected behaviour)

Simple test with "`app.js`" entrypoint and two imported scripts "`foo.js`" and ... "`Crocodile.js`".

**Before** (priority: normal / order: last)

before

**After** (expected behaviour: highest priority)

after

(screenshots made with Firefox on macOS)

### DIFF
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
@@ -87,23 +87,22 @@ class ImportMapManager
     {
         $rawImportMapData = $this->getRawImportMapData();
         $finalImportMapData = [];
-        foreach ($entrypointNames as $entry) {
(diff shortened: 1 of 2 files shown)


weaverryan (approved):
Perfect - I had originally thought the presence if `import 'app'` immediately in the HTML would mean that the preload wasn't needed. However, it appears that the preload even controls the priority. Making the "entry" preloaded is consistent also with an MDN example - https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/modulepreload#examples

(This is also how it works in AssetMapper 6.3!)

fabpot:
Thank you @smnandre.

PR #51884, opened 2023-10-07
[Form] add "model_type" option to MoneyType

### LINKED ISSUE #50707: Add new option to MoneyType

garak:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #50707
| License       | MIT

This is a replacement for abandoned #50720 

Please notice that the previous PR misunderstood the related issue: the intended casting should be applied when the divisor is not 1, otherwise it doesn't make sense.

### DIFF
--- a/src/Symfony/Component/Form/CHANGELOG.md
+++ b/src/Symfony/Component/Form/CHANGELOG.md
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Deprecate not configuring the `default_protocol` option of the `UrlType`, it will default to `null` in 8.0
  * Add a `keep_as_list` option to `CollectionType`
+ * Add a new `model_type` option to `MoneyType`, to be able to cast the transformed value to `integer`
 
 7.0
 ---
--- a/src/Symfony/Component/Form/Extension/Core/DataTransformer/MoneyToLocalizedStringTransformer.php
+++ b/src/Symfony/Component/Form/Extension/Core/DataTransformer/MoneyToLocalizedStringTransformer.php
@@ -22,12 +22,14 @@ use Symfony\Component\Form\Exception\TransformationFailedException;
 class MoneyToLocalizedStringTransformer extends NumberToLocalizedStringTransformer
 {
     private int $divisor;
+    private string $modelType;
 
-    public function __construct(?int $scale = 2, ?bool $grouping = true, ?int $roundingMode = \NumberFormatter::ROUND_HALFUP, ?int $divisor = 1, string $locale = null)
+    public function __construct(?int $scale = 2, ?bool $grouping = true, ?int $roundingMode = \NumberFormatter::ROUND_HALFUP, ?int $divisor = 1, string $locale = null, string $modelType = 'float')
     {
         parent::__construct($scale ?? 2, $grouping ?? true, $roundingMode, $locale);
 
         $this->divisor = $divisor ?? 1;
+        $this->modelType = $modelType;
     }
 
     /**
@@ -62,7 +64,8 @@ class MoneyToLocalizedStringTransformer extends NumberToLocalizedStringTransform
     {
         $value = parent::reverseTransform($value);
         if (null !== $value && 1 !== $this->divisor) {
-            $value = (float) (string) ($value * $this->divisor);
+            $value = (string) ($value * $this->divisor);
+            $value = 'integer' === $this->modelType ? (int) $value : (float) $value;
         }
 
         return $value;
--- a/src/Symfony/Component/Form/Extension/Core/Type/MoneyType.php
+++ b/src/Symfony/Component/Form/Extension/Core/Type/MoneyType.php
@@ -34,7 +34,8 @@ class MoneyType extends AbstractType
                 $options['grouping'],
                 $options['rounding_mode'],
                 $options['divisor'],
-                $options['html5'] ? 'en' : null
+                $options['html5'] ? 'en' : null,
+                $options['model_type'],
             ))
         ;
     }
@@ -59,6 +60,7 @@ class MoneyType extends AbstractType
             'compound' => false,
             'html5' => false,
             'invalid_message' => 'Please enter a valid money amount.',
+            'model_type' => 'float',
         ]);
 
         $resolver->setAllowedValues('rounding_mode', [
@@ -75,9 +77,19 @@ class MoneyType extends AbstractType
 
         $resolver->setAllowedTypes('html5', 'bool');
 
+        $resolver->setAllowedValues('model_type', ['float', 'integer']);
+
         $resolver->setNormalizer('grouping', static function (Options $options, $value) {
             if ($value && $options['html5']) {
                 throw new LogicException('Cannot use the "grouping" option when the "html5" option is enabled.');
+            }
+
+            return $value;
+        });
+
+        $resolver->setNormalizer('model_type', static function (Options $options, $value) {
+            if ('integer' === $value && 1 === $options['divisor']) {
+                throw new LogicException('When the "model_type" option is set to "integer", the "divisor" option should not be set to "1".');
             }
 
             return $value;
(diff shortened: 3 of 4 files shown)


src/Symfony/Component/Form/Extension/Core/DataTransformer/MoneyToLocalizedStringTransformer.php
```diff
@@ -62,7 +64,8 @@ public function reverseTransform(mixed $value): int|float|null
     {
         $value = parent::reverseTransform($value);
         if (null !== $value && 1 !== $this->divisor) {
-            $value = (float) (string) ($value * $this->divisor);
+            $value = (string) ($value * $this->divisor);
+            $value = 'integer' === $this->modelType ? (int) $value : (float) $value;
```
yceruto, line 68: If the `model_type = integer` only works when `divisor` isn't 1, I'd raise a logic exception (for better DX) if the given options don't match this condition? You could add this in an option normalizer, similar to the `grouping` normalizer.

yceruto (changes requested):
Thanks Massimiliano for this improvement! I left a minor request.

src/Symfony/Component/Form/Extension/Core/DataTransformer/MoneyToLocalizedStringTransformer.php
garak: I implemented the suggested change

src/Symfony/Component/Form/Extension/Core/Type/MoneyType.php
```diff
@@ -84,13 +86,23 @@ public function configureOptions(OptionsResolver $resolver)
...
+
         $resolver->setNormalizer('grouping', static function (Options $options, $value) {
             if ($value && $options['html5']) {
                 throw new LogicException('Cannot use the "grouping" option when the "html5" option is enabled.');
             }
 
             return $value;
         });
+
+        $resolver->setNormalizer('divisor', static function (Options $options, $value) {
```
yceruto: A minor nitpick: it would make more sense to target `model_type` because it's the option being validated here, and it aligns better with the exception message.

yceruto (approved):

src/Symfony/Component/Form/Extension/Core/Type/MoneyType.php
garak: Yeah, you're right. Changed.

garak:
@xabbuh could you review, please?

src/Symfony/Component/Form/Extension/Core/Type/MoneyType.php
```diff
@@ -37,7 +37,8 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 $options['grouping'],
                 $options['rounding_mode'],
                 $options['divisor'],
-                $options['html5'] ? 'en' : null
+                $options['html5'] ? 'en' : null,
+                $options['model_type']
```
nicolas-grekas:
```suggestion
                $options['model_type'],
```

src/Symfony/Component/Form/Extension/Core/Type/MoneyType.php
```diff
@@ -84,13 +86,23 @@ public function configureOptions(OptionsResolver $resolver)
...
             if ($value && $options['html5']) {
                 throw new LogicException('Cannot use the "grouping" option when the "html5" option is enabled.');
             }
 
             return $value;
         });
+
+        $resolver->setNormalizer('model_type', static function (Options $options, $value) {
+            if ('integer' === $value && 1 === $options['divisor']) {
+                throw new LogicException('Can use integer "model_type" option only when the "divisor" option is not 1.');
```
nicolas-grekas:
```suggestion
                throw new LogicException('When the "model_type" option is set to "integer", the "divisor" option should not be set to "1".');
```

but why?

src/Symfony/Component/Form/Tests/Extension/Core/Type/MoneyTypeTest.php
```diff
@@ -123,4 +124,28 @@ public function testHtml5EnablesSpecificFormatting()
...
+    {
+        $form = $this->factory->create(static::TESTED_TYPE, null, ['divisor' => 100, 'model_type' => 'integer']);
+        $form->submit('12345.67');
+
+        $this->assertSame(1234567, $form->getData());
+    }
+
+    public function testIntegerModelTypeExpectsDivisorNotEqualToOne()
+    {
+        $this->expectException(LogicException::class);
```
nicolas-grekas, line 146: should be moved before the submit()

nicolas-grekas (commented):
Please add a line in the chagelog of the component in section 7.1

src/Symfony/Component/Form/Extension/Core/Type/MoneyType.php
garak: Because if you set the divisor to 1, you can end up with a float instead of an integer, and lose decimals in the casting

garak:
> Please add a line in the chagelog of the component in section 7.1

I see no changelog for 7.1, should I create a new file?

nicolas-grekas:
I mean this file: https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Form/CHANGELOG.md

src/Symfony/Component/Form/Tests/Extension/Core/Type/MoneyTypeTest.php
garak: As you can see in one of the commits, that doesn't work. The exception is raised before. So maybe I could remove the submit from the tests

src/Symfony/Component/Form/Extension/Core/Type/MoneyType.php
nicolas-grekas: I'm sorry I don't get it. Also the original PR had exactly the opposite : throwing when the divisor is *not* 1.
https://github.com/symfony/symfony/pull/50720/files#diff-eb30dc48f2b749fa5d76ce9598bb51c00fcb97bd7a10337f0630db00f5ba3b9aR100

I'm pretty sure I'm missing something :grimacing:

src/Symfony/Component/Form/Tests/Extension/Core/Type/MoneyTypeTest.php
nicolas-grekas: that'd make sense yes since the line is never executed

src/Symfony/Component/Form/Extension/Core/Type/MoneyType.php
garak: I know, and indeed I added a note about that when I opened this PR.

src/Symfony/Component/Form/CHANGELOG.md
```diff
@@ -1,6 +1,20 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Deprecate not configuring the `default_protocol` option of the `UrlType`, it will default to `null` in 8.0
+ * Add a new `model_type` option to `MoneyType`, to be able to cast the transformed value to `integer`
+
+7.0
```
fabpot, line 11: Looks like there are unrelated entries in both 7.0 and 7.1.

fabpot (changes requested):

fabpot (approved):

fabpot:
Thank you @garak.

PR #51887, opened 2023-10-08
[HttpFoundation] Fix type of properties in Request class

### LINKED ISSUE #51792: [HttpFoundation] unset `Request` properties within `initialize()` will cause issue on any extended Request implementing `__isset()` and `__get()`

crynobone:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #51792 
| License       | MIT

Additional fixes to #51812

`static::$formats` can be `null` to allow lazy-loading:

https://github.com/symfony/symfony/blob/33fedcee2814947c59df2570952f8c800dfdb23e/src/Symfony/Component/HttpFoundation/Request.php#L1252-L1254

### DIFF
--- a/src/Symfony/Component/HttpFoundation/Request.php
+++ b/src/Symfony/Component/HttpFoundation/Request.php
@@ -201,7 +201,7 @@ class Request
     protected $defaultLocale = 'en';
 
     /**
-     * @var array<string, string[]>
+     * @var array<string, string[]>|null
      */
     protected static $formats;
 


nicolas-grekas:
Unlike magic methods, isn't this a situation you can deal with on your side?

crynobone:
CleanShot 2023-10-08 at 09 50 15

The error on Symfony 7 is the same with as before, and also causing Symfony own tests to failed on 7.0 branch:

CleanShot 2023-10-08 at 18 21 09

~~https://github.com/symfony/symfony/actions/runs/6442148533/job/17492677886~~

https://github.com/symfony/symfony/actions/runs/6442148533/job/17492678134

xabbuh (approved):

xabbuh:
see also #51904

fabpot (approved):

fabpot:
Thank you @crynobone.

PR #51888, opened 2023-10-08
[WebProfiler] Profiler improvements / extract Font from stylesheet

smnandre:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| License       | MIT

This PR remove 30% of profiler responses size, removing unused font/data64

---

**AndMore 1** - fixed a couple of CSS bugs
**AndMore 2** - made the logo clickable, allowing anyone to quickly returning to the profiler homepage

---

In the HTML source code of  _every_ profiler page (excluding the toolbar) ....  was inejected around **100KB** of base64 encoded font.

With the raise of ui / system fonts, the adoption of HTTP2, and all the other techniques to handle assets in a simplier way...  i think we have to improve this (minor) situation.

This PR `extract the woff` file in its own file, handled by the Profiler controller, and leverage the "local" stores.

Result: **30%of  immediate reduction**

| Before | After |
| - | - |
|BEFORE  |  AFTER |

Now it's
* less energy  used
* easier to debug/read 
* lighter to log

---


smnandre:
I'm having troubles with tests... i'll fix / ask help just after the implementation feedback .

smnandre:
The Twig changes made those last 24h may explain (at least partially) my troubles with the tests

https://github.com/twigphp/Twig/pull/3882
https://github.com/twigphp/Twig/pull/3887

src/Symfony/Bundle/WebProfilerBundle/Controller/ProfilerController.php
```diff
@@ -323,6 +324,26 @@ public function xdebugAction(): Response
         return new Response($xdebugInfo, 200, ['Content-Type' => 'text/html']);
     }
 
+    /**
+     * Return the JetBrainMono font.
```
javiereguiluz: I'd make this comment more generic: "Downloads the custom web fonts used in the profiler"

src/Symfony/Bundle/WebProfilerBundle/Controller/ProfilerController.php
```diff
@@ -323,6 +324,26 @@ public function xdebugAction(): Response
...
+     *
+     * @throws NotFoundHttpException
+     */
+    public function downloadFontAction(string $font): Response
```
javiereguiluz: I'd rename the argument as `$fontFamilyName` to be more precise.

src/Symfony/Bundle/WebProfilerBundle/Controller/ProfilerController.php
```diff
@@ -323,6 +324,26 @@ public function xdebugAction(): Response
...
+    {
+        $this->denyAccessIfProfilerDisabled();
+        $this->profiler?->disable();
+
+        $filename = basename($font).'.woff2';
```
javiereguiluz: Let's be super cautious here and reduce the attack surface by restricting the font families that we can load.

Something like:

```php
if ('JetBrainsMono' !== $fontFamilyName) {
    // return nothing
}
```

src/Symfony/Bundle/WebProfilerBundle/Resources/fonts/JetBrainsMono.woff2
javiereguiluz, line 1: Please, double check that you took the Woff2 file of the "variable font".

We need the variable font because we display normal and bold monospace contents. The exact font we'd need is this one:

https://github.com/JetBrains/JetBrainsMono/blob/master/fonts/variable/JetBrainsMono%5Bwght%5D.woff2

It's the regular font (not italic) but variable in the `[wght]` axis so we have all weights  available. Thanks!

javiereguiluz (commented):
Thanks for this Simon. I left some comments.

src/Symfony/Bundle/WebProfilerBundle/Controller/ProfilerController.php
smnandre: Not sure if this is open to discussion.. but i find this block more readable with "`font`" (ie: a _file_) than "`fontFamilyName`" (`font_family_name` ?)

```twig
{# simplified #} 
@font-face {
    font-family: 'JetBrains Mono';
    font-style: normal;
    font-weight: 100 900;
    src:
        local('JetBrains Mono'),
        local('JetBrainsMono'),
        url('{{ url('_profiler_download_font', {font: 'JetBrainsMono-Variable'}) }}') format('woff2');
}
```

I'd keep the variable name "$font", but i'd rename '`JetBrainsMono`' into '`JetBrainsMono-Variable`' 

What do you think ?

src/Symfony/Bundle/WebProfilerBundle/Resources/fonts/JetBrainsMono.woff2
smnandre: I used one that i (wrongly?) compressed but you're right. 

Now that it's only downloaded once we clearly don't need the extra compression... 

And using the official one will be easier to maintain. 

👍

src/Symfony/Bundle/WebProfilerBundle/Controller/ProfilerController.php
```diff
@@ -32,7 +33,7 @@
  */
 class ProfilerController
 {
-    private TemplateManager $templateManager;
+    private ?TemplateManager $templateManager;
```
stof: this should be reverted

smnandre:
Test errors are now unrelated

src/Symfony/Bundle/WebProfilerBundle/Controller/ProfilerController.php
javiereguiluz: Sorry for the nitpick, but I think the variable name should be precise about its contents. To me:

* `$font`, the binary data of a single font file (e.g. `0001010100...`)
* `$fontFile` or `$fontFilePath`, the path of some binary font file (e.g. `./fonts/JetBrainsMono-ExtraLightItalic.woff2`
* `$fontName`, the name of a font file (e.g. `JetBrainsMono-MediumItalic.otf`)
* `$fontFamily`, either a ZIP file with all the fonts that belong to the family or the path to that file (e.g. `/uploads/fonts/JetBrainsMono-2.304.zip`)
* `$fontFamilyName`, only the name of the font family (as used e.g. by CSS) (e.g. `JetBrainsMono`)

stof: but was is passed here is the basename of the font file. It is not the font family name.

javiereguiluz: Correct! Sorry. Then, the variable should be `$fontFileName`, right?

We use it here, so it's not a path, URL, or binary file. It's only the `fileName`:

```php
$file = \dirname(__DIR__).'/Resources/fonts/'.$font.'.woff2';
```

smnandre: I agree with almost all your namings, except $fontName.

If $fontFamilyName is 'JetBrainsMono' ...
* $fontName would be 'JetBrainsMono-MediumItalic 
* $fontFileName would be 'JetBrainsMono-MediumItalic.otf'

So in this case we could do

```php
    public function downloadFontAction(string $fontName): Response
```

and 

```twig
url('_profiler_download_font', {fontName: 'JetBrainsMono-Variable'})
```


Would that work for you ?

javiereguiluz: Yes! Thank you ... and sorry for insisting so much on this 🙏

smnandre: Don't be! And thank _you_ for the feedbacks :)

smnandre:
Is there anything else i can do ? Should i rebase the commits ?

javiereguiluz (approved):

src/Symfony/Bundle/WebProfilerBundle/Controller/ProfilerController.php
```diff
@@ -323,6 +324,28 @@ public function xdebugAction(): Response
         return new Response($xdebugInfo, 200, ['Content-Type' => 'text/html']);
     }
 
+    /**
+     * Downloads the custom web fonts used in the profiler.
```
fabpot, line 328:
```suggestion
     * Returns the custom web fonts used in the profiler.
```

src/Symfony/Bundle/WebProfilerBundle/Controller/ProfilerController.php
```diff
@@ -323,6 +324,28 @@ public function xdebugAction(): Response
...
+     *
+     * @throws NotFoundHttpException
+     */
+    public function downloadFontAction(string $fontName): Response
```
fabpot, line 332:
```suggestion
    public function getFontAction(string $fontName): Response
```

I find `download` misleading as this action does not download anything.

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/header.html.twig
```diff
@@ -1,5 +1,5 @@
 <div id="header">
-    <h1>{{ source('@WebProfiler/Icon/symfony.svg') }} Symfony Profiler</h1>
+    <h1><a href="{{ path('_profiler_home') }}">{{ source('@WebProfiler/Icon/symfony.svg') }} Symfony Profiler</a></h1>
```
ro0NL, line 2: :100:

src/Symfony/Bundle/WebProfilerBundle/Controller/ProfilerController.php
94noni: perhaps just "load" is enough

fabpot (approved):

fabpot:
Thank you @smnandre.

PR #51904, opened 2023-10-09
[HttpFoundation] do not access properties before initialization

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | 
| License       | MIT

### DIFF
--- a/src/Symfony/Component/HttpFoundation/Request.php
+++ b/src/Symfony/Component/HttpFoundation/Request.php
@@ -152,9 +152,9 @@ class Request
     protected string $defaultLocale = 'en';
 
     /**
-     * @var array<string, string[]>
+     * @var array<string, string[]>|null
      */
-    protected static array $formats;
+    protected static ?array $formats = null;
 
     protected static ?\Closure $requestFactory = null;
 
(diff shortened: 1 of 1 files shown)


stof:
the right solution is actually to add a `null` default value there (otherwise, we break BC for child classes)

xabbuh:
changed it to initialise the properties with `null` instead

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -152,13 +152,13 @@ class Request
...
-     * @var array<string, string[]>
+     * @var array<string, string[]>|null
      */
-    protected static array $formats;
+    protected static ?array $formats = null;
 
     protected static ?\Closure $requestFactory = null;
 
-    private string $preferredFormat;
+    private ?string $preferredFormat = null;
```
stof: I don't think this one needs to be changed. As it is a private method, we can fully control the lazy initialization and the code handling it was fine (as we use `??=`)

xabbuh: we can keep it as is, but then we need to update `getPreferredFormat()` to not assign `null` to this property (changes pushed)

stof: ah, I missed that we were sometimes assigning `null`

fabpot (approved):

fabpot:
Thank you @xabbuh.

PR #51866, opened 2023-10-06
add all missing properties

y4roc:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

Current behavior: If multiple arguments are missing in the constructor, only the first argument is listed as missing in the array.

Expected behavior: All missing arguments are listed.

PR #49832, opened 2023-03-27
[Serializer] Fix MissingConstructorArgumentsException returning missing argument one by one

ktherage:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

This PR is an improvment of #49013 and fix #42712.

She aims to fix the problem reported in #49013 indicating that the method `getMissingConstructorArguments()` of `\Symfony\Component\Serializer\Exception\MissingConstructorArgumentsException` class introduced in #42712 is returning only one missing arguments when there's more than one missing.

PR #49013, opened 2023-01-17
[Serializer] Replace the MissingConstructorArgumentsException class with MissingConstructorArgumentException

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | -
| License       | MIT
| Doc PR        | -

Followup to #42712.

Currently only the missing arguments are saved in the `MissingConstructorArgumentsException`, so there's no way of knowing which class is missing those arguments when nested objects are being deserialized. This PR adds this information to the exception.

Also, while working on this I've noticed that the `$missingArguments` argument accepts an array of missing arguments, but only one is ever passed. I'm not sure if this was done intentionally or by accident, but I've changed it so that only one is expected. Please let me know if this change is ok.

UPDATE:
I've added a new `MissingConstructorArgumentException` class which accepts only one argument and the class name and deprecated the `MissingConstructorArgumentsException` class. However, since the main goal of this PR is to add the class to the exception, I can revert the renaming part if you think the change is not worth it.

PR #51907, opened 2023-10-09
[Serializer] Fix collecting only first missing constructor argument

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT

Alternative to #51866, sort of followup to #49832 

Currently on 5.4 only the first exception is added to the `not_normalizable_value_exceptions` array when `COLLECT_DENORMALIZATION_ERRORS` is `true` or only the first argument is mentioned in the `MissingConstructorArgumentsException` when it is `false`.
On 6.3 however, the part with the `MissingConstructorArgumentsException` was fix with #49832, but the part with the `not_normalizable_value_exceptions` was overlooked.
IMO this is inconsistent behavior as the two cases are actually the same thing with the only difference being that in one case an exception is thrown while in the other the errors are collected.

I'm not sure if #51866 really qualifies as a bug or is actually more a feature, but the reason #49832 was merged onto 6.3 was because of the changes originally done in #49013, which itself was a feature. 

If #51866 does qualify as a bug then it would make sense to backport #49832 to 5.4 for consistency, which is what this PR does.

The PR contains two commits:
1) backport of #49832 
2) alternative to #51866

If #51866 does not qualify as a bug, the first commit can be drooped and the second one can be rebased with 6.4.

PS I think it's easier to review the changes commit by commit.

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
@@ -348,7 +348,7 @@ abstract class AbstractNormalizer implements NormalizerInterface, DenormalizerIn
             }
 
             $constructorParameters = $constructor->getParameters();
-
+            $missingConstructorArguments = [];
             $params = [];
             foreach ($constructorParameters as $constructorParameter) {
                 $paramName = $constructorParameter->name;
@@ -401,7 +401,8 @@ abstract class AbstractNormalizer implements NormalizerInterface, DenormalizerIn
                     $params[] = null;
                 } else {
                     if (!isset($context['not_normalizable_value_exceptions'])) {
-                        throw new MissingConstructorArgumentsException(sprintf('Cannot create an instance of "%s" from serialized data because its constructor requires parameter "%s" to be present.', $class, $constructorParameter->name), 0, null, [$constructorParameter->name]);
+                        $missingConstructorArguments[] = $constructorParameter->name;
+                        continue;
                     }
 
                     $exception = NotNormalizableValueException::createForUnexpectedDataType(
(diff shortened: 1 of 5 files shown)


src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
```diff
@@ -412,24 +413,22 @@ protected function instantiateObject(array &$data, string $class, array &$contex
...
                 }
             }
 
-            if ($constructor->isConstructor()) {
-                try {
-                    return $reflectionClass->newInstanceArgs($params);
-                } catch (\TypeError $th) {
-                    if (!isset($context['not_normalizable_value_exceptions'])) {
-                        throw $th;
-                    }
```
HypeMC, line 426: This was originally added in #50024, but I think it's not necessary any more because of the new `if` on line 419.

y4roc:
Your solution looks great, but I see a conflict with RequestMapper.

Do you tested your solution with PayloadMapper in 6.3?

HypeMC:
@y4roc Hi, you were right, my changes did cause an issue with the `RequestPayloadValueResolver` because the functionality used there was not covered on 5.4.
I've added a test (`testCollectDenormalizationErrorsWithInvalidConstructorTypes`) so this wouldn't be a problem in the future. I've also tested my changes on 6.3 to check that there are no other failing tests.

src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
```diff
@@ -412,24 +413,26 @@ protected function instantiateObject(array &$data, string $class, array &$contex
                         true
                     );
                     $context['not_normalizable_value_exceptions'][] = $exception;
-
-                    return $reflectionClass->newInstanceWithoutConstructor();
```
HypeMC, line 416: This is now handled in the `try...catch` block.

nicolas-grekas (approved):

nicolas-grekas:
Thank you @HypeMC.

PR #51911, opened 2023-10-09
[Validator] Validate file with empty host

### LINKED ISSUE #51908: [Validator] Url validator don't accept file scheme with empty hostname

6.3.4

### Description

Url validator don't accept file scheme with empty hostname.

### How to reproduce

```php
$this->getValidator()->validate(
   'file:///path',
   new Url(null, null, ['file'], false),
   null
)
```

return `ConstraintViolation` with message `This value is not a valid URL.` but value `file:///path` (empty hostname) is a valid uri (see https://en.wikipedia.org/wiki/File_URI_scheme)

### Possible Solution

_No response_

### Additional Context

_No response_

miranovy:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #51908 
| License       | MIT

Url validator accept file scheme with empty hostname. Value `file:///path` (empty hostname) is now a valid uri :)

### DIFF
--- a/src/Symfony/Component/Validator/Constraints/UrlValidator.php
+++ b/src/Symfony/Component/Validator/Constraints/UrlValidator.php
@@ -71,6 +71,12 @@ class UrlValidator extends ConstraintValidator
             $value = ($constraint->normalizer)($value);
         }
 
+
+        if (in_array('file', $constraint->protocols, true) &&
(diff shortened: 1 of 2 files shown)


OskarStark (commented):
Please add tests, thanks

nicolas-grekas:
I don't think this change is valid. Use another constraint if you want to accept those. I don't think people that use this constraint would expect that a local path is valid.

PR #51916, opened 2023-10-09
[DependencyInjection] Simplify using DI attributes with `ServiceLocator/Iterator`'s

kbond:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | n/a
| License       | MIT

Instead of:
```php
#[AutowireLocator([
    'subscribed' => new SubscribedService(type: 'string', attributes: new Autowire('%some.parameter%')),
])]
```

This PR allows:
```php
#[AutowireLocator([
    'subscribed' => new Autowire('%some.parameter%'),
])]
```


kbond:
The current implementation restricts the usage to just `AutowireLocator|Iterator` and only with the `Autowire` attribute. Should I make it work with `ServiceSubscriberInterface`? Are there any other DI attributes it should work with (I'm thinking no as all the relevent ones extend `Autowire`)?

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
```diff
@@ -50,6 +50,12 @@ public function __construct(
         foreach ($services as $key => $type) {
             $attributes = [];
 
+            if ($type instanceof Autowire) {
+                $references[$key] = new TypedReference($key, 'mixed', attributes: [$type]);
```
kbond: Should I throw an exception is the key is numeric?

```php
#[AutowireIterator([new Autowire('%some.parameter%')])]
```

nicolas-grekas: Why would we?  It's fine to not use keys, isn't it?

kbond: Ah, for the iterator, yes. For the locator though?

nicolas-grekas:
> The current implementation restricts the usage to just AutowireLocator|Iterator and only with the Autowire attribute. Should I make it work with ServiceSubscriberInterface?

Making `ServiceSubscriberInterface` aware of `Autowire*` attributes would mean we'd have to move them to contracts. I don't think we're ready for that. Before, we might need to figure out how SubscribedService and Autowire relate to each other.

But we could make RegisterServiceSubscribersPass aware of `Autowire*` attributes yes.

> Are there any other DI attributes it should work with (I'm thinking no as all the relevent ones extend Autowire)?

Nope as you spotted.

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
nicolas-grekas: Don't we have some logic in place that turns numeric indexes into service-ids for locator? It's not used here?

nicolas-grekas: I'm removing this line, it's not needed

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
```diff
@@ -11,67 +11,25 @@
...
     ) {
-        if (\is_string($services)) {
-            parent::__construct(new TaggedIteratorArgument($services, $indexAttribute, $defaultIndexMethod, false, $defaultPriorityMethod, (array) $exclude, $excludeSelf));
-
-            return;
-        }
-
-        $references = [];
-
-        foreach ($services as $key => $type) {
```
nicolas-grekas, line 50: I've moved this logic to AutowireLocator: while working on this PR, I realized that it doesn't make sense to support `AutowireLocator([FooInterface::class])`. People that really want a static iterable as default can use `Autowire([new Autowire(FooInterface::class)])` instead.

src/Symfony/Component/DependencyInjection/Attribute/AutowireLocator.php
```diff
@@ -37,14 +39,44 @@ public function __construct(
...
         }
 
-        parent::__construct(new ServiceLocatorArgument($iterator));
+        $references = [];
+
+        foreach ($services as $key => $type) {
+            $attributes = [];
+
+            if ($type instanceof Autowire) {
+                $references[$key] = $type;
```
nicolas-grekas, line 54: no need for more here, AutowirePass will turn that into a TypedReference on its own

src/Symfony/Component/DependencyInjection/Compiler/RegisterServiceSubscribersPass.php
```diff
@@ -78,6 +79,11 @@ protected function processValue(mixed $value, bool $isRoot = false): mixed
         foreach ($class::getSubscribedServices() as $key => $type) {
             $attributes = [];
 
+            if (!isset($serviceMap[$key]) && $type instanceof Autowire) {
+                $subscriberMap[$key] = $type;
```
nicolas-grekas, line 83: Adding support for `new Autowire` being returned by `getSubscribedServices`, for consistency, altough it's not in contracts, it might still be convenient.

src/Symfony/Component/HttpKernel/composer.json
```diff
@@ -63,7 +63,6 @@
         "symfony/http-client-contracts": "<2.5",
         "symfony/mailer": "<5.4",
         "symfony/messenger": "<5.4",
-        "symfony/service-contracts": "<3.2",
```
nicolas-grekas, line 66: as requested, since this is an optional feature

nicolas-grekas (commented):
I pushed some changes, let's wait for tests and GTM.

kbond:
Thanks for finalizing!

src/Symfony/Component/HttpKernel/composer.json
bendavies: 👏 thanks!

PR #51926, opened 2023-10-10
[Mime] Forbid messages that are generators to be used more than once

### LINKED ISSUE #51764: Memory Leak in Symfony Mailer when sending many attachments of >5MB

fabpot:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #51764, Fix #51874
| License       | MIT

This is a continuation of #51874, when someone tries to send the same message more than once.
@chalasr @stof

### DIFF
--- a/src/Symfony/Component/Mime/RawMessage.php
+++ b/src/Symfony/Component/Mime/RawMessage.php
@@ -19,6 +19,7 @@ use Symfony\Component\Mime\Exception\LogicException;
 class RawMessage
 {
     private iterable|string|null $message = null;
+    private bool $isGeneratorClosed;
 
     public function __construct(iterable|string $message)
     {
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/Mime/RawMessage.php
```diff
@@ -47,6 +47,22 @@ public function toIterable(): iterable
             return;
         }
 
+        if ($this->message instanceof \Generator) {
+            if (!$this->message->valid()) {
+                trigger_deprecation('symfony/mime', '6.4', 'Sending an email with a closed generator is deprecated and will throw in 7.0.');
```
stof: This deprecation won't be effective: on the first iteration, it is not closed yet (otherwise we already get an exception anyway) and for next iterations, we will have the string assigned on line 61, not the generator.

src/Symfony/Component/Mime/Tests/RawMessageTest.php
```diff
@@ -46,6 +46,21 @@ public function testSerialization(mixed $messageParameter, bool $supportReuse)
...
 
+    /**
+     * @dataProvider provideMessages
+     */
+    public function testToIterable(mixed $messageParameter, bool $supportReuse)
+    {
+        $message = new RawMessage($messageParameter);
+        $this->assertEquals('some string', implode('', iterator_to_array($message->toIterable())));
+
+        // in 7.0, the test with a generator will throw an exception
```
stof: you should be asserting the deprecation for cases not supporting reuse. This would prove you that it currently does not get triggered.

stof (changes requested):

src/Symfony/Component/Mime/RawMessage.php
fabpot: Oops :) Good catch, should be better now.

PR #51947, opened 2023-10-10
[Cache][Doctrine][DoctrineBridge][Lock][Messenger] Compatibility with ORM 3 and DBAL 4

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT

This PR makes the 5.4 branch compatible with ORM 3 and DBAL 4.

Note: On the 5.4 branch, I cannot add ORM 3 to the root composer.json because ORM 3 requires VarExporter 6. So, only high-deps tests it is. 🤷🏻‍♂️

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/ConnectionTest.php
+++ b/src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/ConnectionTest.php
@@ -19,8 +19,10 @@ use Doctrine\DBAL\Exception as DBALException;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Platforms\MariaDBPlatform;
 use Doctrine\DBAL\Platforms\MySQL57Platform;
+use Doctrine\DBAL\Platforms\MySQLPlatform;
 use Doctrine\DBAL\Platforms\OraclePlatform;
 use Doctrine\DBAL\Platforms\SQLServer2012Platform;
+use Doctrine\DBAL\Platforms\SQLServerPlatform;
 use Doctrine\DBAL\Query\QueryBuilder;
 use Doctrine\DBAL\Result;
 use Doctrine\DBAL\Schema\AbstractSchemaManager;
@@ -98,7 +100,7 @@ class ConnectionTest extends TestCase
     {
         $this->expectException(TransportException::class);
         $driverConnection = $this->getDBALConnectionMock();
-        $driverConnection->method('delete')->willThrowException(new DBALException());
+        $driverConnection->method('delete')->willThrowException($this->createStub(DBALException::class));
 
         $connection = new Connection([], $driverConnection);
         $connection->ack('dummy_id');
@@ -108,7 +110,7 @@ class ConnectionTest extends TestCase
     {
         $this->expectException(TransportException::class);
         $driverConnection = $this->getDBALConnectionMock();
-        $driverConnection->method('delete')->willThrowException(new DBALException());
+        $driverConnection->method('delete')->willThrowException($this->createStub(DBALException::class));
 
         $connection = new Connection([], $driverConnection);
         $connection->reject('dummy_id');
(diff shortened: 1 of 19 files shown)


src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/ConnectionTest.php
```diff
@@ -495,13 +506,22 @@ public function provideFindAllSqlGeneratedByPlatform(): iterable
...
+            class_exists(SQLServerPlatform::class) ? new SQLServerPlatform() : new SQLServer2012Platform(),
             'SELECT m.* FROM messenger_messages m WHERE (m.delivered_at is null OR m.delivered_at < ?) AND (m.available_at <= ?) AND (m.queue_name = ?) ORDER BY (SELECT 0) OFFSET 0 ROWS FETCH NEXT 50 ROWS ONLY',
         ];
 
-        yield 'Oracle' => [
-            new OraclePlatform(),
-            'SELECT a.* FROM (SELECT m.id AS "id", m.body AS "body", m.headers AS "headers", m.queue_name AS "queue_name", m.created_at AS "created_at", m.available_at AS "available_at", m.delivered_at AS "delivered_at" FROM messenger_messages m WHERE (m.delivered_at is null OR m.delivered_at < ?) AND (m.available_at <= ?) AND (m.queue_name = ?)) a WHERE ROWNUM <= 50',
-        ];
+        if (!class_exists(MySQL57Platform::class)) {
+            // DBAL >= 4
```
stof, line 514: Can't we detect the Oracle changes based on something related to OraclePlatform rather than checking for the removal of MySQL57Platform ?

src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/ConnectionTest.php
```diff
@@ -391,7 +393,7 @@ public function testGeneratedSql(AbstractPlatform $platform, string $expectedSql
     public static function providePlatformSql(): iterable
     {
         yield 'MySQL' => [
```
stof, line 395: DBAL 4+ separates MariaDBPlatform from MySQLPlatform. So we should add a test for it in the data provider to avoid loosing support for MariaDB.

stof:
I'm wondering whether it make sense to do that work in the 5.4 branch while 6.4 is likely to be the active LTS by the time of the Doctrine major releases.

src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/ConnectionTest.php
derrabus: That test is already there, scroll down a little.

src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/ConnectionTest.php
derrabus: If you have an idea, please let me know. 🙂

derrabus:
> I'm wondering whether it make sense to do that work in the 5.4 branch while 6.4 is likely to be the active LTS by the time of the Doctrine major releases.

We will maintain ORM 2 for quite a while, so only making Symfony 6 compatible with ORM 3 would be fine. A big problem is though, that no Symfony 5 package (except VarExporter) is actively conflicting with ORM 3. It's a dev dependency everywhere. So basically, people will be able to install ORM 3 with Symfony 5 and flood our tracker with bug reports. 😢

OskarStark:
We could explicitly conflict with ORM 3 to avoid such situation

derrabus:
> We could explicitly conflict with ORM 3 to avoid such situation

We could've done that two years ago, yes. If we do that now, Composer will just pin an old Doctrine Bridge and move on.

nicolas-grekas:
Thank you @derrabus.

Issue #51920, opened 2023-10-09
Improve PHPDoc descriptions of Attribute classes and parameters


GromNaN:
### Description

Attributes are used in application codebase to configure specific behaviors. It's sometimes difficult, especially for a novice, to understand the purpose of an attribute or one of its parameters. To improve the developer experience, I suggest adding a description for each attribute class and parameter. Even if it seems obvious, it's a good idea to be explicit.

If you want to work on a class, or a set of classes, please comment this issue before to avoid duplicate work by someone else.

List of all attributes classes to track the progress (checked = doc is complete):

### Can be ignored -> deprecated
- [x] [`Symfony\Component\DependencyInjection\Attribute\MapDecorated`](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/DependencyInjection/Attribute/MapDecorated.php)

### Can be ignored -> already good 👍 
- [x] [`Symfony\Component\DependencyInjection\Attribute\Exclude`](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/DependencyInjection/Attribute/Exclude.php)

### Todo

- [ ] [`Symfony\Component\Validator\Constraints\GroupSequenceProvider`](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Validator/Constraints/GroupSequenceProvider.php)

### In progress

- [ ] [`Symfony\Component\Validator\Constraints\GroupSequence`](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Validator/Constraints/GroupSequence.php) #53428
- [x] [`Symfony\Bundle\FrameworkBundle\Routing\Attribute/AsRoutingConditionService`](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Bundle/FrameworkBundle/Routing/Attribute/AsRoutingConditionService.php) #51996
- [ ] [`Symfony\Component\RemoteEvent\Attribute\AsRemoteEventConsumer`](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/RemoteEvent/Attribute/AsRemoteEventConsumer.php) #51996
- [ ] [`Symfony\Component\Routing\Annotation\Route`](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Routing/Annotation/Route.php) #51996
[...]

PR #51968, opened 2023-10-10
[HttpFoundation]  Improve PHPDoc of Cache attribute

GromNaN:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | Part of #51920
| License       | MIT

@javiereguiluz was right to point out that the PHPDoc wasn't perfect.

### DIFF
--- a/src/Symfony/Component/HttpKernel/Attribute/Cache.php
+++ b/src/Symfony/Component/HttpKernel/Attribute/Cache.php
@@ -13,6 +13,10 @@ namespace Symfony\Component\HttpKernel\Attribute;
 
 /**
  * Describes the default HTTP cache headers on controllers.
+ * Headers defined in the Cache attribute are ignored if they are already set
+ * by the controller.
+ *
+ * @see https://symfony.com/doc/current/http_cache.html#making-your-responses-http-cacheable
  *
  * @author Fabien Potencier <fabien@symfony.com>
  */
@@ -38,27 +42,46 @@ final class Cache
         public int|string|null $smaxage = null,
 
         /**
-         * Whether the response is public or not.
+         * If true, the contents will be stored in a public cache and served to all
+         * the next requests.
          */
         public ?bool $public = null,
 
         /**
-         * Whether or not the response must be revalidated.
+         * If true, the response is not served stale by a cache in any circumstance
+         * without first revalidating with the origin.
          */
         public bool $mustRevalidate = false,
 
         /**
-         * Additional "Vary:"-headers.
+         * Set "Vary" header.
+         *
+         * Example:
+         * ['Accept-Encoding', 'User-Agent']
+         *
+         * @see https://symfony.com/doc/current/http_cache/cache_vary.html
+         *
+         * @var string[]
          */
         public array $vary = [],
 
         /**
          * An expression to compute the Last-Modified HTTP header.
+         *
+         * The expression is evaluated by the ExpressionLanguage component, it
+         * receives all the request attributes and the resolved controller arguments.
+         *
+         * The result of the expression must be a DateTimeInterface.
          */
         public ?string $lastModified = null,
 
         /**
          * An expression to compute the ETag HTTP header.
+         *
+         * The expression is evaluated by the ExpressionLanguage component, it
+         * receives all the request attributes and the resolved controller arguments.
+         *
+         * The result must be a string that will be hashed.
          */
         public ?string $etag = null,
 


src/Symfony/Component/HttpKernel/Attribute/Cache.php
```diff
@@ -38,27 +38,41 @@ public function __construct(
...
          */
         public ?string $lastModified = null,
 
         /**
          * An expression to compute the ETag HTTP header.
+         *
+         * The expression is evaluated by the ExpressionLanguage component, it
+         * receives all the request attributes and the resolved controller arguments.
+         *
+         * The result must be a string that will hashed.
```
OskarStark:
```suggestion
         * The result must be a string that will be hashed.
```
Will it result in a hash or does it need to be a hash already?

GromNaN: The listener hashes the result of the expression.
https://github.com/symfony/symfony/blob/d6c879747d19fabbf220ea58b915c3eadf121259/src/Symfony/Component/HttpKernel/EventListener/CacheAttributeListener.php#L72-L75

OskarStark: Thanks 👍

src/Symfony/Component/HttpKernel/Attribute/Cache.php
```diff
@@ -38,27 +38,41 @@ public function __construct(
         public int|string|null $smaxage = null,
 
         /**
-         * Whether the response is public or not.
+         * If true, the contents will be stored in a public cache and serve to all
```
javiereguiluz: `serve` -> `served`

src/Symfony/Component/HttpKernel/Attribute/Cache.php
```diff
@@ -38,27 +38,41 @@ public function __construct(
...
 
         /**
-         * Whether or not the response must be revalidated.
+         * If true, the response is not served stale by a cache in any circumstance
+         * without first revalidating with the origin.
          */
         public bool $mustRevalidate = false,
 
         /**
          * Additional "Vary:"-headers.
```
javiereguiluz: This might be a bit confusing. "additional" headers ... additional to which ones? Which `Vary` headers do Symfony apply by default?

Also ... when using arrays sometimes is difficult to know exactly the required format:

```php
#[Cache(vary: ['Accept-Language', 'Accept-Encoding'])]

#[Cache(vary: ['Vary' => 'Accept-Language', 'Vary' => 'Accept-Encoding'])]

#[Cache(vary: ['Vary: Accept-Language', 'Vary: Accept-Encoding'])]

#[Cache(vary: ['Accept-Language' => '*', 'Accept-Encoding' => '*'])]
```

So maybe we could add a very short example.

GromNaN: Example added.

src/Symfony/Component/HttpKernel/Attribute/Cache.php
```diff
@@ -50,7 +54,12 @@ public function __construct(
         public bool $mustRevalidate = false,
 
         /**
-         * Additional "Vary:"-headers.
+         * Set "Vary" header.
+         *
+         * Example:
+         * #[Cache(vary: ['Accept-Encoding', 'User-Agent'])]
```
OskarStark:
```suggestion
         * ['Accept-Encoding', 'User-Agent']
```
I propose to just use the value here

javiereguiluz: Yes, I think that's easier to understand here.

src/Symfony/Component/HttpKernel/Attribute/Cache.php
```diff
@@ -13,6 +13,10 @@
 
 /**
  * Describes the default HTTP cache headers on controllers.
+ * Headers defined in with the Cache attribute are ignored if they are already set
```
javiereguiluz: `defined in with the Cache attribute` -> `defined in the Cache attribute`

javiereguiluz (approved):
Jérôme, thanks for this.

OskarStark (approved):

src/Symfony/Component/HttpKernel/Attribute/Cache.php
GromNaN:
```suggestion
 * Headers defined in the Cache attribute are ignored if they are already set
```

GromNaN:
Thanks for your help @OskarStark @javiereguiluz. I realize @alexandre-daubois is much more productive than I am 🥇

fabpot (approved):

fabpot:
Thank you @GromNaN.

PR #51969, opened 2023-10-10
[FrameworkBundle] Fix calling `Kernel::warmUp()` when running `cache:warmup`

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT

Maps to https://github.com/symfony/symfony/blob/e86d26abd07187d5ec5b2d780cfd1d872d57fe2a/src/Symfony/Component/HttpKernel/Kernel.php#L581-L587

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Command/CacheWarmupCommand.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Command/CacheWarmupCommand.php
@@ -18,6 +18,7 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\DependencyInjection\Dumper\Preloader;
 use Symfony\Component\HttpKernel\CacheWarmer\CacheWarmerAggregate;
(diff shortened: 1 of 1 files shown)


src/Symfony/Bundle/FrameworkBundle/Command/CacheWarmupCommand.php
```diff
@@ -74,6 +75,10 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             $this->cacheWarmer->enableOptionalWarmers();
         }
 
+        if ($kernel instanceof WarmableInterface) {
+            $kernel->warmUp($kernel->getContainer()->getParameter('kernel.cache_dir'));
```
HeahDude: It seems you can create the `$cacheDir` variable before this new `if` instead of line 82.

nicolas-grekas: Right, updated!

PR #51971, opened 2023-10-10
[DependencyInjection][HttpKernel] Add PHPDoc to attribute classes and properties

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

Friendly ping @GromNaN, is that what you had in mind? :slightly_smiling_face:

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Attribute/AsAlias.php
+++ b/src/Symfony/Component/DependencyInjection/Attribute/AsAlias.php
@@ -19,6 +19,10 @@ namespace Symfony\Component\DependencyInjection\Attribute;
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
 final class AsAlias
 {
+    /**
+     * @param string|null $id     The id of the alias
+     * @param bool        $public Whether to declare the alias public
+     */
     public function __construct(
         public ?string $id = null,
         public bool $public = false,
--- a/src/Symfony/Component/DependencyInjection/Attribute/AsDecorator.php
+++ b/src/Symfony/Component/DependencyInjection/Attribute/AsDecorator.php
@@ -13,9 +13,17 @@ namespace Symfony\Component\DependencyInjection\Attribute;
 
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
+/**
+ * Declares a decorating service.
+ */
 #[\Attribute(\Attribute::TARGET_CLASS)]
 class AsDecorator
 {
+    /**
+     * @param string $decorates The service id to decorate
+     * @param int    $priority  The priority of this decoration when multiple decorators are declared for the same service
+     * @param int    $onInvalid The behavior to adopt when the decoration is invalid; must be one of the {@see ContainerInterface} constants
+     */
     public function __construct(
         public string $decorates,
         public int $priority = 0,
--- a/src/Symfony/Component/DependencyInjection/Attribute/AsTaggedItem.php
+++ b/src/Symfony/Component/DependencyInjection/Attribute/AsTaggedItem.php
@@ -19,6 +19,10 @@ namespace Symfony\Component\DependencyInjection\Attribute;
 #[\Attribute(\Attribute::TARGET_CLASS)]
 class AsTaggedItem
 {
+    /**
+     * @param string|null $index    The property or method to use to index the item in the locator
+     * @param int|null    $priority The priority of the item; the higher the number, the earlier the tagged service will be located in the locator
+     */
     public function __construct(
         public ?string $index = null,
         public ?int $priority = null,
--- a/src/Symfony/Component/DependencyInjection/Attribute/Autoconfigure.php
+++ b/src/Symfony/Component/DependencyInjection/Attribute/Autoconfigure.php
@@ -19,6 +19,18 @@ namespace Symfony\Component\DependencyInjection\Attribute;
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
 class Autoconfigure
 {
+    /**
+     * @param array<array-key, array<array-key, mixed>>|null $tags         The tags to add to the service
+     * @param array<string, array<array-key, mixed>>|null    $calls        The calls to be made when instantiating the service
+     * @param array<string, mixed>|null                      $bind         The bindings to declare for the service
+     * @param bool|string|null                               $lazy         Whether the service is lazy-loaded
+     * @param bool|null                                      $public       Whether to declare the service as public
+     * @param bool|null                                      $shared       Whether to declare the service as shared
+     * @param bool|null                                      $autowire     Whether to declare the service as autowired
+     * @param array<string, mixed>|null                      $properties   The properties to define when creating the service
+     * @param array<class-string, string>|string|null        $configurator A PHP function, reference or an array containing a class/Reference and a method to call after the service is fully initialized
+     * @param string|null                                    $constructor  The public static method to use to instantiate the service
+     */
     public function __construct(
         public ?array $tags = null,
         public ?array $calls = null,
--- a/src/Symfony/Component/DependencyInjection/Attribute/AutoconfigureTag.php
+++ b/src/Symfony/Component/DependencyInjection/Attribute/AutoconfigureTag.php
@@ -19,6 +19,10 @@ namespace Symfony\Component\DependencyInjection\Attribute;
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
 class AutoconfigureTag extends Autoconfigure
 {
+    /**
+     * @param string|null             $name       The tag name to add
+     * @param array<array-key, mixed> $attributes The tag attributes to attach to the tag
+     */
     public function __construct(string $name = null, array $attributes = [])
     {
         parent::__construct(
--- a/src/Symfony/Component/DependencyInjection/Attribute/AutowireDecorated.php
+++ b/src/Symfony/Component/DependencyInjection/Attribute/AutowireDecorated.php
@@ -11,6 +11,9 @@
 
 namespace Symfony\Component\DependencyInjection\Attribute;
 
+/**
+ * Autowires the inner object of decorating services.
+ */
 #[\Attribute(\Attribute::TARGET_PARAMETER)]
 class AutowireDecorated
 {
--- a/src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
+++ b/src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
@@ -20,7 +20,14 @@ use Symfony\Component\DependencyInjection\Argument\TaggedIteratorArgument;
 class AutowireIterator extends Autowire
 {
     /**
-     * @param string|string[] $exclude A service or a list of services to exclude
+     * @see ServiceSubscriberInterface::getSubscribedServices()
+     *
+     * @param string               $tag                   A tag name to search for to populate the iterator
+     * @param string|null          $indexAttribute        The name of the attribute that defines the key referencing each service in the tagged collection
+     * @param string|null          $defaultIndexMethod    The static method that should be called to get each service's key when their tag doesn't define the previous attribute
+     * @param string|null          $defaultPriorityMethod The static method that should be called to get each service's priority when their tag doesn't define the "priority" attribute
+     * @param string|array<string> $exclude               A service id or a list of service ids to exclude
+     * @param bool                 $excludeSelf           Whether to automatically exclude the referencing service from the iterator
      */
     public function __construct(
         string $tag,
--- a/src/Symfony/Component/DependencyInjection/Attribute/AutowireLocator.php
+++ b/src/Symfony/Component/DependencyInjection/Attribute/AutowireLocator.php
@@ -28,8 +28,12 @@ class AutowireLocator extends Autowire
     /**
      * @see ServiceSubscriberInterface::getSubscribedServices()
      *
-     * @param string|array<string|SubscribedService> $services An explicit list of services or a tag name
-     * @param string|string[]                        $exclude  A service or a list of services to exclude
+     * @param string|array<string|SubscribedService> $services              A tag name or an explicit list of service ids
+     * @param string|null                            $indexAttribute        The name of the attribute that defines the key referencing each service in the locator
+     * @param string|null                            $defaultIndexMethod    The static method that should be called to get each service's key when their tag doesn't define the previous attribute
+     * @param string|null                            $defaultPriorityMethod The static method that should be called to get each service's priority when their tag doesn't define the "priority" attribute
+     * @param string|array                           $exclude               A service id or a list of service ids to exclude
+     * @param bool                                   $excludeSelf           Whether to automatically exclude the referencing service from the locator
      */
     public function __construct(
         string|array $services,
--- a/src/Symfony/Component/DependencyInjection/Attribute/AutowireServiceClosure.php
+++ b/src/Symfony/Component/DependencyInjection/Attribute/AutowireServiceClosure.php
@@ -20,6 +20,9 @@ use Symfony\Component\DependencyInjection\Reference;
 #[\Attribute(\Attribute::TARGET_PARAMETER)]
 class AutowireServiceClosure extends Autowire
 {
+    /**
+     * @param string $service The service id to wrap in the closure
+     */
     public function __construct(string $service)
     {
         parent::__construct(new ServiceClosureArgument(new Reference($service)));
--- a/src/Symfony/Component/DependencyInjection/Attribute/TaggedIterator.php
+++ b/src/Symfony/Component/DependencyInjection/Attribute/TaggedIterator.php
@@ -11,9 +11,20 @@
 
 namespace Symfony\Component\DependencyInjection\Attribute;
 
+/**
+ * Autowires an iterator of services based on a tag name.
+ */
 #[\Attribute(\Attribute::TARGET_PARAMETER)]
 class TaggedIterator extends AutowireIterator
 {
+    /**
+     * @param string          $tag                   The tag to look for to populate the iterator
+     * @param string|null     $indexAttribute        The name of the attribute that defines the key referencing each service in the tagged collection
+     * @param string|null     $defaultIndexMethod    The static method that should be called to get each service's key when their tag doesn't define the previous attribute
+     * @param string|null     $defaultPriorityMethod The static method that should be called to get each service's priority when their tag doesn't define the "priority" attribute
+     * @param string|string[] $exclude               A service id or a list of service ids to exclude
+     * @param bool            $excludeSelf           Whether to automatically exclude the referencing service from the iterator
+     */
     public function __construct(
         public string $tag,
         public ?string $indexAttribute = null,
--- a/src/Symfony/Component/DependencyInjection/Attribute/TaggedLocator.php
+++ b/src/Symfony/Component/DependencyInjection/Attribute/TaggedLocator.php
@@ -11,9 +11,20 @@
 
 namespace Symfony\Component\DependencyInjection\Attribute;
 
+/**
+ * Autowires a locator of services based on a tag name.
+ */
 #[\Attribute(\Attribute::TARGET_PARAMETER)]
 class TaggedLocator extends AutowireLocator
 {
+    /**
+     * @param string          $tag                   The tag to look for to populate the locator
+     * @param string|null     $indexAttribute        The name of the attribute that defines the key referencing each service in the tagged collection
+     * @param string|null     $defaultIndexMethod    The static method that should be called to get each service's key when their tag doesn't define the previous attribute
+     * @param string|null     $defaultPriorityMethod The static method that should be called to get each service's priority when their tag doesn't define the "priority" attribute
+     * @param string|string[] $exclude               A service id or a list of service ids to exclude
+     * @param bool            $excludeSelf           Whether to automatically exclude the referencing service from the locator
+     */
     public function __construct(
         public string $tag,
         public ?string $indexAttribute = null,
--- a/src/Symfony/Component/DependencyInjection/Attribute/Target.php
+++ b/src/Symfony/Component/DependencyInjection/Attribute/Target.php
@@ -22,9 +22,11 @@ use Symfony\Component\DependencyInjection\Exception\LogicException;
 #[\Attribute(\Attribute::TARGET_PARAMETER)]
 final class Target
 {
-    public function __construct(
-        public ?string $name = null,
-    ) {
+    /**
+     * @param string|null $name The name of the target autowiring alias
+     */
+    public function __construct(public ?string $name = null)
+    {
     }
 
     public function getParsedName(): string
--- a/src/Symfony/Component/DependencyInjection/Attribute/When.php
+++ b/src/Symfony/Component/DependencyInjection/Attribute/When.php
@@ -19,8 +19,10 @@ namespace Symfony\Component\DependencyInjection\Attribute;
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::TARGET_FUNCTION | \Attribute::IS_REPEATABLE)]
 class When
 {
-    public function __construct(
-        public string $env,
-    ) {
+    /**
+     * @param string $env The environment under which the class will be registered as a service (i.e. "dev", "test", "prod")
+     */
+    public function __construct(public string $env)
+    {
     }
 }
--- a/src/Symfony/Component/HttpKernel/Attribute/MapDateTime.php
+++ b/src/Symfony/Component/HttpKernel/Attribute/MapDateTime.php
@@ -12,6 +12,7 @@
 namespace Symfony\Component\HttpKernel\Attribute;
 
 use Symfony\Component\HttpKernel\Controller\ArgumentResolver\DateTimeValueResolver;
+use Symfony\Component\HttpKernel\Controller\ValueResolverInterface;
 
 /**
  * Controller parameter tag to configure DateTime arguments.
@@ -19,6 +20,11 @@ use Symfony\Component\HttpKernel\Controller\ArgumentResolver\DateTimeValueResolv
 #[\Attribute(\Attribute::TARGET_PARAMETER)]
 class MapDateTime extends ValueResolver
 {
+    /**
+     * @param string|null                                 $format   The DateTime format to use, @see https://php.net/datetime.format
+     * @param bool                                        $disabled Whether this value resolver is disabled; this allows to enable a value resolver globally while disabling it in specific cases
+     * @param class-string<ValueResolverInterface>|string $resolver The name of the resolver to use
+     */
     public function __construct(
         public readonly ?string $format = null,
         bool $disabled = false,
--- a/src/Symfony/Component/HttpKernel/Attribute/MapQueryParameter.php
+++ b/src/Symfony/Component/HttpKernel/Attribute/MapQueryParameter.php
@@ -12,6 +12,7 @@
 namespace Symfony\Component\HttpKernel\Attribute;
 
 use Symfony\Component\HttpKernel\Controller\ArgumentResolver\QueryParameterValueResolver;
+use Symfony\Component\HttpKernel\Controller\ValueResolverInterface;
 
 /**
  * Can be used to pass a query parameter to a controller argument.
@@ -24,7 +25,11 @@ final class MapQueryParameter extends ValueResolver
     /**
      * @see https://php.net/filter.filters.validate for filter, flags and options
      *
-     * @param string|null $name The name of the query parameter. If null, the name of the argument in the controller will be used.
+     * @param string|null                                         $name     The name of the query parameter; if null, the name of the argument in the controller will be used
+     * @param (FILTER_VALIDATE_*)|(FILTER_SANITIZE_*)|null        $filter   The filter to pass to "filter_var()"
+     * @param int-mask-of<(FILTER_FLAG_*)|FILTER_NULL_ON_FAILURE> $flags    The flags to pass to "filter_var()"
+     * @param array                                               $options  The options to pass to "filter_var()"
+     * @param class-string<ValueResolverInterface>|string         $resolver The name of the resolver to use
      */
     public function __construct(
         public ?string $name = null,
--- a/src/Symfony/Component/HttpKernel/Attribute/ValueResolver.php
+++ b/src/Symfony/Component/HttpKernel/Attribute/ValueResolver.php
@@ -13,11 +13,15 @@ namespace Symfony\Component\HttpKernel\Attribute;
 
 use Symfony\Component\HttpKernel\Controller\ValueResolverInterface;
 
+/**
+ * Defines which value resolver should be used for a given parameter.
+ */
 #[\Attribute(\Attribute::TARGET_PARAMETER | \Attribute::IS_REPEATABLE)]
 class ValueResolver
 {
     /**
-     * @param class-string<ValueResolverInterface>|string $resolver
+     * @param class-string<ValueResolverInterface>|string $resolver The class name of the resolver to use
+     * @param bool                                        $disabled Whether this value resolver is disabled; this allows to enable a value resolver globally while disabling it in specific cases
      */
     public function __construct(
         public string $resolver,
--- a/src/Symfony/Component/HttpKernel/Attribute/WithHttpStatus.php
+++ b/src/Symfony/Component/HttpKernel/Attribute/WithHttpStatus.php
@@ -12,13 +12,16 @@
 namespace Symfony\Component\HttpKernel\Attribute;
 
 /**
+ * Defines the HTTP status code applied to an exception.
+ *
  * @author Dejan Angelov <angelovdejan@protonmail.com>
  */
 #[\Attribute(\Attribute::TARGET_CLASS)]
 class WithHttpStatus
 {
     /**
-     * @param array<string, string> $headers
+     * @param int                   $statusCode The HTTP status code to use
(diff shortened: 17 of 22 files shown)


src/Symfony/Component/DependencyInjection/Attribute/AsAlias.php
```diff
@@ -19,6 +19,10 @@
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
 final class AsAlias
 {
+    /**
+     * @param string|null   $id     The id of the service to alias
```
GromNaN: You can move the comment before each parameter for the promoted properties. Like in Cache class. So that you have more space for description and example if necessary.

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
```diff
@@ -28,8 +28,12 @@ class AutowireIterator extends Autowire
     /**
      * @see ServiceSubscriberInterface::getSubscribedServices()
      *
-     * @param string|array<string|SubscribedService> $services A tag name or an explicit list of services
-     * @param string|string[]                        $exclude  A service or a list of services to exclude
+     * @param string|array<string|SubscribedService>    $services               A tag name or an explicit list of services
```
GromNaN: To be consistant, use "service ids"

src/Symfony/Component/DependencyInjection/Attribute/When.php
```diff
@@ -19,6 +19,9 @@
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::TARGET_FUNCTION | \Attribute::IS_REPEATABLE)]
 class When
 {
+    /**
+     * @param string $env The environment under which the class will be registered as a service
```
GromNaN: We can provide an example: dev, test, prod.

GromNaN (commented):
Great work.

src/Symfony/Component/DependencyInjection/Attribute/AsAlias.php
alexandre-daubois: Addressed this comment as well as the other ones :slightly_smiling_face:

src/Symfony/Component/DependencyInjection/Attribute/When.php
alexandre-daubois: Great idea!

src/Symfony/Component/DependencyInjection/Attribute/AsAlias.php
```diff
@@ -20,7 +20,13 @@
 final class AsAlias
 {
     public function __construct(
+        /**
+         * The id of the service to alias.
+         */
         public ?string $id = null,
+        /**
```
OskarStark:
```suggestion

        /**
```

src/Symfony/Component/DependencyInjection/Attribute/AutoconfigureTag.php
```diff
@@ -19,7 +19,16 @@
...
+        /*
+         * The tag name to add.
+         */
+        string $name = null,
+
+        /*
+         * The tag attributes to attach to the tag.
+         */
+        array $attributes = [])
     {
```
OskarStark:
```suggestion
        array $attributes = []
    ) {
```

src/Symfony/Component/DependencyInjection/Attribute/AutowireServiceClosure.php
```diff
@@ -20,8 +20,12 @@
 #[\Attribute(\Attribute::TARGET_PARAMETER)]
 class AutowireServiceClosure extends Autowire
 {
-    public function __construct(string $service)
-    {
+    public function __construct(
+        /*
```
OskarStark: Should we use one or two `*`? Should be the same everywhere

alexandre-daubois: Fabbot would like only one it seems (on my other PR at least), so I'd say one. But two are required when using syntax like `{@see}` :thinking:

OskarStark:
@GromNaN is there any reason, why we don't use proper `@param` annotations?

src/Symfony/Component/DependencyInjection/Attribute/AutowireServiceClosure.php
alexandre-daubois: It looks like fabbot requires `/**` when CPP is used and `/*` otherwise

GromNaN: Indeed, commenting just before the parameters only works for the CPP 😞. It allows to write documentation once and access it on both the argument and the property. It also avoids having to repeat the name and type of the argument.

src/Symfony/Component/DependencyInjection/Attribute/TaggedLocator.php
```diff
@@ -11,16 +11,45 @@
...
+ */
 #[\Attribute(\Attribute::TARGET_PARAMETER)]
 class TaggedLocator extends AutowireLocator
 {
     public function __construct(
+        /**
+         * The tag to look for to populate the locator.
+         */
         public string $tag,
-        public ?string $indexAttribute = null,
```
GromNaN, line 19: You removed the CPP. Was it intentional?

src/Symfony/Component/DependencyInjection/Attribute/When.php
```diff
@@ -20,6 +20,9 @@
 class When
 {
     public function __construct(
+        /**
+         * The environment under which the class will be registered as a service (i.e. "dev", "staging", "prod").
```
GromNaN: I'm not sure "staging" is frequently used. It's not in the documentation. "test" is documented and have special behavior in DotEnv.

```suggestion
         * The environment under which the class will be registered as a service (i.e. "dev", "test", "prod").
```

src/Symfony/Component/HttpKernel/Attribute/MapDateTime.php
```diff
@@ -20,8 +20,19 @@
 class MapDateTime extends ValueResolver
 {
     public function __construct(
+        /**
+         * The DateTime format to use.
```
GromNaN:
```
@see https://www.php.net/manual/en/datetime.format.php
```

src/Symfony/Component/HttpKernel/Attribute/MapDateTime.php
```diff
@@ -20,8 +20,19 @@
...
+         */
         public readonly ?string $format = null,
+
+        /*
+         * Whether this value resolver is disabled.
```
GromNaN: This comment is not very descriptive. It could start with `If true, ...`.
Why would someone disable this resolver rather than remove the attribute?

src/Symfony/Component/DependencyInjection/Attribute/TaggedLocator.php
alexandre-daubois: Definitely not, thank you for pointing this out!

src/Symfony/Component/DependencyInjection/Attribute/AsDecorator.php
```diff
@@ -13,12 +13,27 @@
...
+         */
         public string $decorates,
+
+        /**
+         * The priority of this decoration when multiple decorators are declared for the same service.
+         */
         public int $priority = 0,
+
+        /**
+         * The behavior to adopt when the decoration is invalid.Must be one of
```
GromNaN:
```suggestion
         * The behavior to adopt when the decoration is invalid. Must be one of
```

src/Symfony/Component/DependencyInjection/Attribute/AsDecorator.php
```diff
@@ -13,12 +13,27 @@
...
+         * the {@see ContainerInterface} constants.
+         */
```
GromNaN: We can help IDE autocompletion and static analysis with this annotation
```suggestion
         * @var ContainerInterface::*
         */
```

src/Symfony/Component/DependencyInjection/Attribute/TaggedIterator.php
```diff
@@ -11,15 +11,44 @@
...
 #[\Attribute(\Attribute::TARGET_PARAMETER)]
 class TaggedIterator extends AutowireIterator
 {
     public function __construct(
+        /**
+         * The tag to look for to populate the iterator.
+         */
         public string $tag,
+
+        /*
```
GromNaN: There is CPP in this constructor.
```suggestion
        /**
```

src/Symfony/Component/DependencyInjection/Attribute/AutowireServiceClosure.php
stof: this only works for CPP indeed, because only CPP defines a property there. And the phpdoc spec supports phpdoc on properties or on methods, but not inline on arguments.

src/Symfony/Component/DependencyInjection/Attribute/AutoconfigureTag.php
```diff
@@ -19,8 +19,17 @@
...
-    public function __construct(string $name = null, array $attributes = [])
-    {
+    public function __construct(
+        /*
+         * The tag name to add.
+         */
+        string $name = null,
+
+        /*
+         * The tag attributes to attach to the tag.
```
stof: This should use `@param` on the method instead as those are not promoted properties. Otherwise, we won't have the benefits of exposing the description in IDEs supporting to show the phpdoc.

src/Symfony/Component/DependencyInjection/Attribute/AsDecorator.php
alexandre-daubois: Thanks, added here and in `WithLogLevel`!

src/Symfony/Component/DependencyInjection/Attribute/AutoconfigureTag.php
alexandre-daubois: Added, thanks! :+1:

src/Symfony/Component/DependencyInjection/Attribute/AutowireServiceClosure.php
alexandre-daubois: Alright, thank you! I updated the whole PR by using `/**` on CPP and `@param` in other case. You can see example of this in the different value resolver like `MapRequestPayload`.

src/Symfony/Component/DependencyInjection/Attribute/TaggedLocator.php
```diff
@@ -11,15 +11,44 @@
...
         public ?string $defaultIndexMethod = null,
+
+        /**
+         * The static method that should be called to get each service's
+         * priority when their tag doesn't define the "priority" attribute.
+         */
         public ?string $defaultPriorityMethod = null,
+
+        /**
+         * A service id or a list of services ids to exclude.
```
OskarStark:
```suggestion
         * A service id or a list of service ids to exclude.
```

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
```diff
@@ -28,8 +28,12 @@ class AutowireIterator extends Autowire
     /**
      * @see ServiceSubscriberInterface::getSubscribedServices()
      *
-     * @param string|array<string|SubscribedService> $services A tag name or an explicit list of services
-     * @param string|string[]                        $exclude  A service or a list of services to exclude
+     * @param string|array<string|SubscribedService> $services              A tag name or an explicit list of services ids
```
OskarStark:
```suggestion
     * @param string|array<string|SubscribedService> $services              A tag name or an explicit list of service ids
```

src/Symfony/Component/DependencyInjection/Attribute/AutowireIterator.php
```diff
@@ -28,8 +28,12 @@ class AutowireIterator extends Autowire
...
+     * @param string|null                            $indexAttribute        The name of the attribute that defines the key referencing each service in the tagged collection
+     * @param string|null                            $defaultIndexMethod    The static method that should be called to get each service's key when their tag doesn't define the previous attribute
+     * @param string|null                            $defaultPriorityMethod The static method that should be called to get each service's priority when their tag doesn't define the "priority" attribute
+     * @param string|array                           $exclude               A service id or a list of services ids to exclude
```
OskarStark:
```suggestion
     * @param string|array                           $exclude               A service id or a list of service ids to exclude
```

src/Symfony/Component/DependencyInjection/Attribute/AutowireLocator.php
```diff
@@ -26,8 +26,12 @@ class AutowireLocator extends Autowire
     /**
      * @see ServiceSubscriberInterface::getSubscribedServices()
      *
-     * @param string|array<string|SubscribedService> $services An explicit list of services or a tag name
-     * @param string|string[]                        $exclude  A service or a list of services to exclude
+     * @param string|array<string|SubscribedService> $services              A tag name or an explicit list of services ids
```
OskarStark:
```suggestion
     * @param string|array<string|SubscribedService> $services              A tag name or an explicit list of service ids
```

src/Symfony/Component/DependencyInjection/Attribute/AutowireLocator.php
```diff
@@ -26,8 +26,12 @@ class AutowireLocator extends Autowire
...
+     * @param string|null                            $indexAttribute        The name of the attribute that defines the key referencing each service in the locator
+     * @param string|null                            $defaultIndexMethod    The static method that should be called to get each service's key when their tag doesn't define the previous attribute
+     * @param string|null                            $defaultPriorityMethod The static method that should be called to get each service's priority when their tag doesn't define the "priority" attribute
+     * @param string|array                           $exclude               A service id or a list of services ids to exclude
```
OskarStark:
```suggestion
     * @param string|array                           $exclude               A service id or a list of service ids to exclude
```

src/Symfony/Component/DependencyInjection/Attribute/TaggedIterator.php
```diff
@@ -11,15 +11,44 @@
...
         public ?string $defaultIndexMethod = null,
+
+        /**
+         * The static method that should be called to get each service's
+         * priority when their tag doesn't define the "priority" attribute.
+         */
         public ?string $defaultPriorityMethod = null,
+
+        /**
+         * A service id or a list of services ids to exclude.
```
OskarStark:
```suggestion
         * A service id or a list of service ids to exclude.
```

src/Symfony/Component/DependencyInjection/Attribute/TaggedLocator.php
alexandre-daubois: 6 times in a row :face_with_spiral_eyes: Thank for pointing this, I updated! :+1:

src/Symfony/Component/HttpKernel/Attribute/MapDateTime.php
alexandre-daubois: Updated to `Whether this value resolver is disabled. This allows to enable a value resolver globally while disabling it in specific cases.`

src/Symfony/Component/DependencyInjection/Attribute/AsAlias.php
```diff
@@ -20,7 +20,14 @@
 final class AsAlias
 {
     public function __construct(
+        /**
+         * The id of the service to alias.
```
stof: This is actually the id of the alias service.
The current description makes me think that this is the id that the alias will target but this is not the case (the target of the alias will be the service defined for the class on which the attribute is used)

src/Symfony/Component/DependencyInjection/Attribute/Autoconfigure.php
```diff
@@ -20,15 +20,55 @@
...
+         */
         public ?array $tags = null,
+
+        /**
+         * The calls to be made when instantiating the service.
+         */
         public ?array $calls = null,
+
+        /**
+         * The bindings to declare for the service.
```
stof: It would be great to provide the more precise type in phpdoc for those array properties to help SA tools and IDEs to report invalid usages.

src/Symfony/Component/DependencyInjection/Attribute/AutoconfigureTag.php
```diff
@@ -19,6 +19,10 @@
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
 class AutoconfigureTag extends Autoconfigure
 {
+    /**
+     * @param string|null $name       The tag name to add
+     * @param array       $attributes The tag attributes to attach to the tag
```
stof:
```suggestion
     * @param array<string, mixed> $attributes The tag attributes to attach to the tag
```

src/Symfony/Component/DependencyInjection/Attribute/AutowireServiceClosure.php
```diff
@@ -20,8 +20,12 @@
...
+         * The service id to wrap in the closure.
```
stof: This one should be migrated to use `@param`

src/Symfony/Component/HttpKernel/Attribute/MapQueryParameter.php
```diff
@@ -24,12 +24,27 @@ final class MapQueryParameter extends ValueResolver
...
+     * @param class-string $resolver The class name of the resolver to use
      */
     public function __construct(
+        /**
+         * The name of the query parameter. If null, the name of the argument in the controller will be used.
+         */
         public ?string $name = null,
+
+        /**
+         * The filter to pass to "filter_var()".
```
stof: you could add `@var FILTER_VALIDATE_*|FILTER_SANITIZE_*|null` to help static analysis tools

src/Symfony/Component/HttpKernel/Attribute/MapQueryParameter.php
```diff
@@ -24,12 +24,27 @@ final class MapQueryParameter extends ValueResolver
...
+         */
         public ?int $filter = null,
+
+        /**
+         * The flags to pass to "filter_var()".
```
stof: `@var int-mask-of<FILTER_FLAG_*|FILTER_NULL_ON_FAILURE>` to tell SA tools that this is a bitmask of the flag constants.

src/Symfony/Component/HttpKernel/Attribute/MapQueryParameter.php
alexandre-daubois: That's better! However, PHPStorm highlights the first part with the following message: `Undefined class 'FILTER_VALIDATE_' `. It feels a bit weird to have this highlight, and I wasn't able to find other examples of this other than in Parameter bag with the following snippet:

```php
    /**
     * Filter key.
     *
     * @param int                                     $filter  FILTER_* constant
     * @param int|array{flags?: int, options?: array} $options Flags from FILTER_* constants
     *
     * @see https://php.net/filter-var
     */
```

Maybe we should add something similar here

src/Symfony/Component/HttpKernel/Attribute/MapQueryParameter.php
alexandre-daubois: Same as above : 
image

I think we should use the notation used in ParameterBag

src/Symfony/Component/DependencyInjection/Attribute/Autoconfigure.php
alexandre-daubois: Updated all arrays of this PR with more precise type :+1:

src/Symfony/Component/HttpKernel/Attribute/MapQueryParameter.php
alexandre-daubois: Also thanks for the tip of `int-mask-of`, I wasn't aware of this notation :slightly_smiling_face:

src/Symfony/Component/DependencyInjection/Attribute/Autoconfigure.php
```diff
@@ -20,15 +20,65 @@
...
+         * The calls to be made when instantiating the service.
+         *
+         * @var array<string, array<array-key, mixed>>
+         */
         public ?array $calls = null,
+
+        /**
+         * The bindings to declare for the service.
+         *
+         * @var array<string, mixed>
```
GromNaN:
```suggestion
         * @var array<string, mixed>|null
```

src/Symfony/Component/DependencyInjection/Attribute/Autoconfigure.php
```diff
@@ -20,15 +20,65 @@
...
+
+        /**
+         * Whether to declare the service as autowired.
+         */
         public ?bool $autowire = null,
+
+        /**
+         * The properties to define when creating the service.
+         *
+         * @var array<string, mixed>
```
GromNaN:
```suggestion
         * @var array<string, mixed>|null
```

src/Symfony/Component/HttpKernel/Attribute/MapQueryParameter.php
GromNaN: Neither [PHPStan](https://phpstan.org/r/0f0e2aea-cf8b-4f88-b380-477e41f9b3c4) nor [Psalm](https://psalm.dev/r/33a8e3a779) recognizes the `@var FILTER_VALIDATE_*|FILTER_SANITIZE_*|null`. Go for your suggestion.

You can rise an issue on Psalm/PHPStan/PHPStorm to support global const pattern.

src/Symfony/Component/HttpKernel/Attribute/MapQueryParameter.php
GromNaN: Same issue with global const.

GromNaN (commented):
LGTM after a few corrections to the types.

src/Symfony/Component/HttpKernel/Attribute/MapQueryParameter.php
stof: it actually requires braces to work inside union types: https://phpstan.org/r/99fe52c1-ffca-4485-9386-a9a9a8f36211

alexandre-daubois: Works like a charm, thanks!

src/Symfony/Component/HttpKernel/Attribute/MapQueryParameter.php
alexandre-daubois: Updated with Christophe's hint :slightly_smiling_face:

src/Symfony/Component/DependencyInjection/Attribute/Autoconfigure.php
alexandre-daubois: Updated

src/Symfony/Component/DependencyInjection/Attribute/Autoconfigure.php
alexandre-daubois: Oups sorry, forgot about it! I addressed your comment :slightly_smiling_face:

src/Symfony/Component/DependencyInjection/Attribute/AsAlias.php
```diff
@@ -20,7 +20,14 @@
 final class AsAlias
 {
     public function __construct(
```
OskarStark, line 26: `@param` should be used

alexandre-daubois: Updated the whole PR 👍

src/Symfony/Component/DependencyInjection/Attribute/AsDecorator.php
```diff
@@ -13,9 +13,17 @@
...
+/**
+ * An attribute to declare a decorating service.
+ */
 #[\Attribute(\Attribute::TARGET_CLASS)]
 class AsDecorator
 {
+    /**
+     * @param string $decorates The service id to decorate
+     * @param int    $priority  The priority of this decoration when multiple decorators are declared for the same service
+     * @param int    $onInvalid The behavior to adopt when the decoration is invalid. Must be one of the {@see ContainerInterface} constants
```
nicolas-grekas:
```suggestion
     * @param int    $onInvalid The behavior to adopt when the decoration is invalid; must be one of the {@see ContainerInterface} constants
```

src/Symfony/Component/DependencyInjection/Attribute/AsTaggedItem.php
```diff
@@ -19,6 +19,10 @@
 #[\Attribute(\Attribute::TARGET_CLASS)]
 class AsTaggedItem
 {
+    /**
+     * @param string|null $index    The property or method to use to index the item in the locator
+     * @param int|null    $priority The priority of the item. The higher the number, the earlier the tagged service will be located in the locator
```
nicolas-grekas:
```suggestion
     * @param int|null    $priority The priority of the item; the higher the number, the earlier the tagged service will be located in the locator
```

src/Symfony/Component/DependencyInjection/Attribute/AutowireDecorated.php
```diff
@@ -11,6 +11,9 @@
 
 namespace Symfony\Component\DependencyInjection\Attribute;
 
+/**
+ * An attribute to autowire the inner object of decorating services.
```
nicolas-grekas:
```suggestion
 * Autowires the inner object of decorating services.
```

src/Symfony/Component/DependencyInjection/Attribute/AsDecorator.php
```diff
@@ -13,9 +13,17 @@
 
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
+/**
+ * An attribute to declare a decorating service.
```
nicolas-grekas:
```suggestion
 * Declares a decorating service.
```

src/Symfony/Component/DependencyInjection/Attribute/Target.php
```diff
@@ -22,9 +22,11 @@
 #[\Attribute(\Attribute::TARGET_PARAMETER)]
 final class Target
 {
-    public function __construct(
-        public ?string $name = null,
-    ) {
+    /**
+     * @param string|null $name The service alias to autowire
```
nicolas-grekas:
```suggestion
     * @param string|null $name The name of the target autowiring alias
```

src/Symfony/Component/HttpKernel/Attribute/MapDateTime.php
```diff
@@ -19,6 +19,11 @@
 #[\Attribute(\Attribute::TARGET_PARAMETER)]
 class MapDateTime extends ValueResolver
 {
+    /**
+     * @param string|null  $format   The DateTime format to use, @see https://www.php.net/manual/en/datetime.format.php
```
nicolas-grekas:
```suggestion
     * @param string|null  $format   The DateTime format to use, @see https://php.net/datetime.format```

File: src/Symfony/Component/HttpKernel/Attribute/MapDateTime.php [hunk 36]
```diff
@@ -19,6 +19,11 @@
...
+     * @param bool         $disabled Whether this value resolver is disabled. This allows to enable a value resolver globally while disabling it in specific cases.
```
nicolas-grekas:
```suggestion
     * @param bool         $disabled Whether this value resolver is disabled; this allows to enable a value resolver globally while disabling it in specific cases
```

src/Symfony/Component/HttpKernel/Attribute/MapDateTime.php
```diff
@@ -19,6 +19,11 @@
...
+     * @param class-string $resolver The class name of the resolver to use
```
nicolas-grekas: this doesn't have to be a FQCN, does it?

src/Symfony/Component/HttpKernel/Attribute/MapQueryParameter.php
```diff
@@ -24,7 +24,11 @@ final class MapQueryParameter extends ValueResolver
     /**
      * @see https://php.net/filter.filters.validate for filter, flags and options
      *
-     * @param string|null $name The name of the query parameter. If null, the name of the argument in the controller will be used.
+     * @param string|null $name The name of the query parameter. If null, the name of the argument in the controller will be used
+     * @param (FILTER_VALIDATE_*)|(FILTER_SANITIZE_*)|null $filter   The filter to pass to "filter_var()"
+     * @param int-mask-of<(FILTER_FLAG_*)|FILTER_NULL_ON_FAILURE> $flags    The flags to pass to "filter_var()"
```
nicolas-grekas, line 29: alignment is odd

src/Symfony/Component/HttpKernel/Attribute/MapQueryParameter.php
```diff
@@ -24,7 +24,11 @@ final class MapQueryParameter extends ValueResolver
     /**
      * @see https://php.net/filter.filters.validate for filter, flags and options
      *
-     * @param string|null $name The name of the query parameter. If null, the name of the argument in the controller will be used.
+     * @param string|null $name The name of the query parameter. If null, the name of the argument in the controller will be used
```
nicolas-grekas:
```suggestion
     * @param string|null $name The name of the query parameter; if null, the name of the argument in the controller will be used
```

src/Symfony/Component/HttpKernel/Attribute/ValueResolver.php
```diff
@@ -13,11 +13,15 @@
...
+/**
+ * An attribute to tell which value resolver should be used for a given parameter.
+ */
 #[\Attribute(\Attribute::TARGET_PARAMETER | \Attribute::IS_REPEATABLE)]
 class ValueResolver
 {
     /**
-     * @param class-string<ValueResolverInterface>|string $resolver
+     * @param class-string<ValueResolverInterface>|string $resolver The class name of the resolver to use
+     * @param bool                                        $disabled Whether this value resolver is disabled. This allows to enable a value resolver globally while disabling it in specific cases.
```
nicolas-grekas:
```suggestion
     * @param bool                                        $disabled Whether this value resolver is disabled; this allows to enable a value resolver globally while disabling it in specific cases
```

src/Symfony/Component/HttpKernel/Attribute/WithHttpStatus.php
```diff
@@ -12,16 +12,28 @@
 namespace Symfony\Component\HttpKernel\Attribute;
 
 /**
+ * An attribute to define the HTTP status code applied to an exception.
```
nicolas-grekas:
```suggestion
 * Defines the HTTP status code applied to an exception.
```

src/Symfony/Component/HttpKernel/Attribute/WithLogLevel.php
```diff
@@ -14,13 +14,15 @@
 use Psr\Log\LogLevel;
 
 /**
+ * An attribute to define the log level applied to an exception.
```
nicolas-grekas:
```suggestion
 * Defines the log level applied to an exception.
```

src/Symfony/Component/HttpKernel/Attribute/MapQueryParameter.php
```diff
@@ -24,7 +25,11 @@ final class MapQueryParameter extends ValueResolver
     /**
      * @see https://php.net/filter.filters.validate for filter, flags and options
      *
-     * @param string|null $name The name of the query parameter. If null, the name of the argument in the controller will be used.
+     * @param string|null                                         $name The name of the query parameter; if null, the name of the argument in the controller will be used
```
nicolas-grekas, line 28: :wave:

nicolas-grekas:
```suggestion
     * @param string|null                                         $name     The name of the query parameter; if null, the name of the argument in the controller will be used
```

src/Symfony/Component/HttpKernel/Attribute/ValueResolver.php
```diff
@@ -13,11 +13,15 @@
 
 use Symfony\Component\HttpKernel\Controller\ValueResolverInterface;
 
+/**
+ * An attribute to tell which value resolver should be used for a given parameter.
```
nicolas-grekas: Defines the value resolver to use for a given parameter.

src/Symfony/Component/DependencyInjection/Attribute/Autoconfigure.php
```diff
@@ -19,6 +19,18 @@
...
+     * @param array<string, array<array-key, mixed>>|null    $calls        The calls to be made when instantiating the service
+     * @param array<string, mixed>|null                      $bind         The bindings to declare for the service
+     * @param bool|string|null                               $lazy         Whether the service is lazy-loaded
+     * @param bool|null                                      $public       Whether to declare the service as public
+     * @param bool|null                                      $shared       Whether to declare the service as shared
+     * @param bool|null                                      $autowire     Whether to declare the service as autowired
+     * @param array<string, mixed>|null                      $properties   The properties to define when creating the service
+     * @param array<class-string, string>|string|null        $configurator A PHP function, reference or an array containing a class/Reference and a method to call after the service is fully initialized
+     * @param string|null                                    $constructor  The public static inner method to use to instantiate the service
```
nicolas-grekas:
```suggestion
     * @param string|null                                    $constructor  The public static method to use to instantiate the service
```

alexandre-daubois:
Fabbot's having (and giving) hard times with some of the advanced types 😅 Addressed remaining comments, thanks

nicolas-grekas:
Thank you @alexandre-daubois.

PR #51974, opened 2023-10-10
[Console][EventDispatcher][Security][Serializer][Workflow] Add PHPDoc to attribute classes and properties

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

One more round.

:information_source: A first review of this kind is being done [here](https://github.com/symfony/symfony/pull/51971), I'll adjust this PR accordingly once done :slightly_smiling_face:

### DIFF
--- a/src/Symfony/Component/Console/Attribute/AsCommand.php
+++ b/src/Symfony/Component/Console/Attribute/AsCommand.php
@@ -17,6 +17,12 @@ namespace Symfony\Component\Console\Attribute;
 #[\Attribute(\Attribute::TARGET_CLASS)]
 class AsCommand
 {
+    /**
+     * @param string      $name        The name of the command, used when calling it (i.e. "cache:clear")
+     * @param string|null $description The description of the command, displayed with the help page
+     * @param string[]    $aliases     The list of aliases of the command. The command will be executed when using one of them (i.e. "cache:clean")
+     * @param bool        $hidden      If true, the command won't be shown when listing all the available commands, but it can still be run as any other command
+     */
     public function __construct(
         public string $name,
         public ?string $description = null,
--- a/src/Symfony/Component/EventDispatcher/Attribute/AsEventListener.php
+++ b/src/Symfony/Component/EventDispatcher/Attribute/AsEventListener.php
@@ -19,6 +19,12 @@ namespace Symfony\Component\EventDispatcher\Attribute;
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
 class AsEventListener
 {
+    /**
+     * @param string|null $event      The event name to listen to
+     * @param string|null $method     The method to run when the listened event is triggered
+     * @param int         $priority   The priority of this listener if several are declared for the same event
+     * @param string|null $dispatcher The service id of the event dispatcher to listen to
+     */
     public function __construct(
         public ?string $event = null,
         public ?string $method = null,
--- a/src/Symfony/Component/Security/Http/Attribute/CurrentUser.php
+++ b/src/Symfony/Component/Security/Http/Attribute/CurrentUser.php
@@ -20,6 +20,10 @@ use Symfony\Component\Security\Http\Controller\UserValueResolver;
 #[\Attribute(\Attribute::TARGET_PARAMETER)]
 class CurrentUser extends ValueResolver
 {
+    /**
+     * @param bool   $disabled Whether this value resolver is disabled, which allows to enable a value resolver globally while disabling it in specific cases
+     * @param string $resolver The class name of the resolver to use
+     */
     public function __construct(bool $disabled = false, string $resolver = UserValueResolver::class)
     {
         parent::__construct($resolver, $disabled);
--- a/src/Symfony/Component/Security/Http/Attribute/IsGranted.php
+++ b/src/Symfony/Component/Security/Http/Attribute/IsGranted.php
@@ -14,38 +14,27 @@ namespace Symfony\Component\Security\Http\Attribute;
 use Symfony\Component\ExpressionLanguage\Expression;
 
 /**
+ * Checks if user has permission to access to some resource using security roles and voters.
+ *
+ * @see https://symfony.com/doc/current/security.html#roles
+ *
  * @author Ryan Weaver <ryan@knpuniversity.com>
  */
 #[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::TARGET_FUNCTION)]
 final class IsGranted
 {
+    /**
+     * @param string|Expression            $attribute     The attribute that will be checked against a given authentication token and optional subject
+     * @param array|string|Expression|null $subject       An optional subject - e.g. the current object being voted on
+     * @param string|null                  $message       A custom message when access is not granted
+     * @param int|null                     $statusCode    If set, will throw HttpKernel's HttpException with the given $statusCode; if null, Security\Core's AccessDeniedException will be used
+     * @param int|null                     $exceptionCode If set, will add the exception code to thrown exception
+     */
     public function __construct(
-        /**
-         * Sets the first argument that will be passed to isGranted().
-         */
         public string|Expression $attribute,
-
-        /**
-         * Sets the second argument passed to isGranted().
-         *
-         * @var array<string|Expression>|string|Expression|null
-         */
         public array|string|Expression|null $subject = null,
-
-        /**
-         * The message of the exception - has a nice default if not set.
-         */
         public ?string $message = null,
-
-        /**
-         * If set, will throw HttpKernel's HttpException with the given $statusCode.
-         * If null, Security\Core's AccessDeniedException will be used.
-         */
         public ?int $statusCode = null,
-
-        /**
-         * If set, will add the exception code to thrown exception.
-         */
         public ?int $exceptionCode = null,
     ) {
     }
--- a/src/Symfony/Component/Serializer/Attribute/Context.php
+++ b/src/Symfony/Component/Serializer/Attribute/Context.php
@@ -22,7 +22,10 @@ class Context
     private array $groups;
 
     /**
-     * @param string|string[] $groups
+     * @param array<string, mixed> $context                The common context to use when serializing or deserializing
+     * @param array<string, mixed> $normalizationContext   The context to use when serializing
+     * @param array<string, mixed> $denormalizationContext The context to use when deserializing
+     * @param string|string[]      $groups                 The groups to use when serializing or deserializing
      *
      * @throws InvalidArgumentException
      */
--- a/src/Symfony/Component/Workflow/Attribute/AsAnnounceListener.php
+++ b/src/Symfony/Component/Workflow/Attribute/AsAnnounceListener.php
@@ -14,6 +14,8 @@ namespace Symfony\Component\Workflow\Attribute;
 use Symfony\Component\EventDispatcher\Attribute\AsEventListener;
 
 /**
+ * Defines a listener for the "announce" event of a workflow.
+ *
  * @author Grégoire Pineau <lyrixx@lyrixx.info>
  */
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
@@ -21,6 +23,13 @@ final class AsAnnounceListener extends AsEventListener
 {
     use BuildEventNameTrait;
 
+    /**
+     * @param string|null $workflow   The id of the workflow to listen to
+     * @param string|null $transition The transition name to which the listener listens to
+     * @param string|null $method     The method to run when the listened event is triggered
+     * @param int         $priority   The priority of this listener if several are declared for the same transition
+     * @param string|null $dispatcher The service id of the event dispatcher to listen to
+     */
     public function __construct(
         string $workflow = null,
         string $transition = null,
--- a/src/Symfony/Component/Serializer/Attribute/DiscriminatorMap.php
+++ b/src/Symfony/Component/Serializer/Attribute/DiscriminatorMap.php
@@ -19,6 +19,12 @@ use Symfony\Component\Serializer\Exception\InvalidArgumentException;
 #[\Attribute(\Attribute::TARGET_CLASS)]
 class DiscriminatorMap
 {
+    /**
+     * @param string                      $typeProperty The property holding the type discriminator
+     * @param array<string, class-string> $mapping      The mapping between types and classes (i.e. ['admin_user' => AdminUser::class])
+     *
+     * @throws InvalidArgumentException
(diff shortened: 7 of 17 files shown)


src/Symfony/Component/EventDispatcher/Attribute/AsEventListener.php
```diff
@@ -20,9 +20,24 @@
...
-        public ?string $method = null,
-        public int $priority = 0,
+
+        /*
+         * The method that acts as the listener.
+         */
+        string $method = null,
+
+        /*
+         * The priority of this listener if several are declared for the same transition.
```
GromNaN: We can explain how priority works and how to debug.
```suggestion
         * The priority of this listener if several are declared for the same transition.
         * Listeners with the highest priority are called first.
         * Use the "debug:event-dispatcher" command to show priorities of other listeners.
```

src/Symfony/Component/Security/Http/Attribute/CurrentUser.php
```diff
@@ -20,8 +20,17 @@
 #[\Attribute(\Attribute::TARGET_PARAMETER)]
 class CurrentUser extends ValueResolver
 {
-    public function __construct(bool $disabled = false, string $resolver = UserValueResolver::class)
-    {
+    public function __construct(
+        /*
+         * Whether to disable thi value resolver.
```
GromNaN: What does it mean to disable a value resolver?

```suggestion
         * Whether to disable this value resolver.
```

src/Symfony/Component/Console/Attribute/AsCommand.php
```diff
@@ -18,9 +18,25 @@
...
+
+        /**
+         * The description of the command, displayed with the help page.
+         */
         public ?string $description = null,
+
+        /*
+         * The list of aliases of the command. The command will be executed when using
+         * one of them (i.e. "cache:clean").
+         */
```
GromNaN:
```suggestion
         *
         * @var string[]
         */
```

src/Symfony/Component/Workflow/Attribute/AsAnnounceListener.php
```diff
@@ -22,10 +24,29 @@ final class AsAnnounceListener extends AsEventListener
     use BuildEventNameTrait;
 
     public function __construct(
+        /*
+         * The id of the workflow to listen to.
```
GromNaN: This need to move back to `@param` 😕

alexandre-daubois: I'll do a new pass on this PR when the other PR is all good so I know what to change all at once. Thank you for this early review Jerôme and spotting these typos! 😃

src/Symfony/Component/Console/Attribute/AsCommand.php
```diff
@@ -18,9 +18,25 @@
...
         array $aliases = [],
+
+        /*
+         * Whether to hide the command from the available commands list.
```
javiereguiluz: I'd mention that the command can still be run as normal. E.g.

```
If true, the command won't be shown when listing all the available commands, but it can still be run as any other command.
```

src/Symfony/Component/EventDispatcher/Attribute/AsEventListener.php
```diff
@@ -20,9 +20,24 @@
...
     public function __construct(
+        /**
+         * The event name to listen to.
+         */
         public ?string $event = null,
-        public ?string $method = null,
-        public int $priority = 0,
+
+        /*
+         * The method that acts as the listener.
```
javiereguiluz: The method run when the listened event is triggered.

src/Symfony/Component/Security/Core/Validator/Constraints/UserPassword.php
```diff
@@ -14,7 +14,10 @@
 use Symfony\Component\Validator\Constraint;
 
 /**
+ * An attribute to check that the target isn't equal to the current user password.
```
javiereguiluz: I wonder if this description could be outdated "soon" if we add more logic to this attribute.

Maybe we could use a more generic description like: _"Performs some validations on user passwords."_

But ... given that this is a constraint called "user password" ... maybe that's the typical "no-description description". You are describing what you can immediately infer from the name of this thing. So, maybe it's better to not add a description of the attribute?

src/Symfony/Component/Security/Core/Validator/Constraints/UserPassword.php
```diff
@@ -29,8 +32,32 @@ class UserPassword extends Constraint
     public $message = 'This value should be the user\'s current password.';
     public $service = 'security.validator.user_password';
 
-    public function __construct(array $options = null, string $message = null, string $service = null, array $groups = null, mixed $payload = null)
-    {
+    public function __construct(
+        /*
+         * Options to pass the validation constraint.
```
javiereguiluz: This reads a bit confusing ... maybe: _"Options passed to the validation constraint"_

src/Symfony/Component/Security/Core/Validator/Constraints/UserPassword.php
```diff
@@ -29,8 +32,32 @@ class UserPassword extends Constraint
...
+         */
+        string $message = null,
+
+        /*
+         * The service id to use to check the user current password.
+         */
+        string $service = null,
+
+        /*
+         * The groups on which this validation constraint applies.
```
javiereguiluz:
```
The validation groups on which this constraint applies.

@see https://symfony.com/doc/current/validation/groups.html
```

src/Symfony/Component/Security/Core/Validator/Constraints/UserPassword.php
```diff
@@ -29,8 +32,32 @@ class UserPassword extends Constraint
...
+         */
+        array $groups = null,
+
+        /*
+         * Arbitrary data to be attached to the constraint.
```
javiereguiluz: I don't fully understand what `attached` means here. Maybe we can use a simpler verb like: "included", "passed", etc. Also, maybe we can explain very concisely the purpose of this?

E.g.

```
Arbitrary data passed to the constraint. This is useful for XXX or YYYY.
```

src/Symfony/Component/Security/Http/Attribute/IsGranted.php
```diff
@@ -14,6 +14,8 @@
 use Symfony\Component\ExpressionLanguage\Expression;
 
 /**
+ * An attribute to check if a user has one or more specific roles.
```
javiereguiluz: The _"An attribute..."_ intro might be redundant.

A suggestion:

```
Checks if users have permission to access to some resources using security roles.

@see https://symfony.com/doc/current/security.html#roles
```

javiereguiluz (commented):
Alex, thanks a lot for taking care of improving attributes doc.

We want to make these descriptions both complete and concise. They need to be very short, but they also need to explain all the basics needed to fully understand them _(what happens if I enable this? what's the format required to pass data here? etc.)_

So, thanks in advance for your patience because we'll add many comments 🙏

alexandre-daubois:
@GromNaN I revamped the PR with the feedback that was given in the other one and by taking into account your review on this one :slightly_smiling_face: 

@javiereguiluz Thanks, I agree with every point! I updated the PR accordingly. I understand how challenging this kind of PR it is, we don't have that much "space" to explain things like in the doc :smile:

src/Symfony/Component/Console/Attribute/AsCommand.php
```diff
@@ -17,8 +17,19 @@
 #[\Attribute(\Attribute::TARGET_CLASS)]
 class AsCommand
 {
+    /**
+     * @param array $aliases The list of aliases of the command. The command will be executed when using one of them (i.e. "cache:clean")
```
stof: `string[]`

src/Symfony/Component/Security/Core/Validator/Constraints/UserPassword.php
```diff
@@ -29,6 +30,13 @@ class UserPassword extends Constraint
     public $message = 'This value should be the user\'s current password.';
     public $service = 'security.validator.user_password';
 
+    /**
+     * @param array|null  $options Options passed to the validation constraint
+     * @param string|null $message The message to display if this constraint raises a violation
+     * @param string|null $service The service id to use to check the user current password
+     * @param array|null  $groups  The groups on which this validation constraint applies (@see https://symfony.com/doc/current/validation/groups.html)
```
stof: groups are `string[]|null` (everywhere)

src/Symfony/Component/Security/Core/Validator/Constraints/UserPassword.php
```diff
@@ -29,6 +30,13 @@ class UserPassword extends Constraint
...
+     * @param mixed|null  $payload Arbitrary data passed to the constraint to make any additional checks in the constraint validator
```
stof: `mixed|null` is just `mixed`

src/Symfony/Component/Serializer/Annotation/DiscriminatorMap.php
```diff
@@ -26,7 +31,14 @@
...
     public function __construct(
+        /**
+         * The property holding the type discriminator.
+         */
         private readonly string $typeProperty,
+
+        /**
+         * The mapping between types and classes (i.e. ['admin_user' => AdminUser::class]).
+         */
         private readonly array $mapping,
```
stof: this should add `@var array<string, class-string>`

alexandre-daubois:
Thanks, I addressed your comments :slightly_smiling_face:

OskarStark:
Rebase please

alexandre-daubois:
Rebased and updated with `@param` 🙂

src/Symfony/Component/Security/Http/Attribute/IsGranted.php
```diff
@@ -14,6 +14,10 @@
 use Symfony\Component\ExpressionLanguage\Expression;
 
 /**
+ * Checks if user has permission to access to some resource using security roles.
```
derrabus: This attribute is actually not limited to roles. You can check any permission that you've registered voters for.

src/Symfony/Component/Serializer/Attribute/Context.php
```diff
@@ -22,7 +22,10 @@ class Context
     private array $groups;
 
     /**
-     * @param string|string[] $groups
+     * @param array           $context                The common context to use when serializing or deserializing
+     * @param array           $normalizationContext   The context to use when serializing
+     * @param array           $denormalizationContext The context to use when deserializing
```
derrabus: `array` of what?

src/Symfony/Component/Security/Http/Attribute/IsGranted.php
alexandre-daubois: I updated the description to mention voters 🙂

src/Symfony/Component/Serializer/Attribute/Context.php
alexandre-daubois: Types added 👍

derrabus (approved):

src/Symfony/Component/Security/Http/Attribute/IsGranted.php
```diff
@@ -14,38 +14,27 @@
...
+ * @see https://symfony.com/doc/current/security.html#roles
+ *
  * @author Ryan Weaver <ryan@knpuniversity.com>
  */
 #[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::TARGET_FUNCTION)]
 final class IsGranted
 {
+    /**
+     * @param string|Expression            $attribute     Sets the first argument that will be passed to isGranted()
+     * @param array|string|Expression|null $subject       Sets the second argument passed to isGranted()
```
nicolas-grekas: Maybe we can make these two lines more self explanatory?

src/Symfony/Component/Security/Http/Attribute/IsGranted.php
```diff
@@ -14,38 +14,27 @@
...
+     * @param string|null                  $message       The message of the exception - has a nice default if not set
```
nicolas-grekas:
```suggestion
     * @param string|null                  $message       A custom message when access is not granted
```

src/Symfony/Component/Workflow/Attribute/AsAnnounceListener.php
```diff
@@ -14,13 +14,22 @@
 use Symfony\Component\EventDispatcher\Attribute\AsEventListener;
 
 /**
+ * An attribute to define a listener for the "announce" event of a workflow.
```
nicolas-grekas: not need to state the obvious:
```suggestion
 * Defines a listener for the "announce" event of a workflow.
```

nicolas-grekas (commented):
As a general note, it'd be nice to add more titles to classes. Sometimes, this can even replace an `@param` FTW

src/Symfony/Component/Workflow/Attribute/AsAnnounceListener.php
alexandre-daubois: Updated and added to missing ones

src/Symfony/Component/Security/Http/Attribute/IsGranted.php
alexandre-daubois: Tried something else that may be more explicit 👍

src/Symfony/Component/Security/Http/Attribute/IsGranted.php
```diff
@@ -14,38 +14,27 @@
...
  * @author Ryan Weaver <ryan@knpuniversity.com>
  */
 #[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::TARGET_FUNCTION)]
 final class IsGranted
 {
+    /**
+     * @param string|Expression            $attribute     The attribute that will be checked against a given authentication token and optional subject
+     * @param array|string|Expression|null $subject       An optional subject - e.g. the current object being voted on
+     * @param string|null                  $message       A custom message when access is not granted
+     * @param int|null                     $statusCode    If set, will throw HttpKernel's HttpException with the given $statusCode. If null, Security\Core's AccessDeniedException will be used
```
nicolas-grekas:
```suggestion
     * @param int|null                     $statusCode    If set, will throw HttpKernel's HttpException with the given $statusCode; if null, Security\Core's AccessDeniedException will be used
```

nicolas-grekas (approved):

src/Symfony/Component/Security/Http/Attribute/IsGranted.php
alexandre-daubois: Updated 👍

nicolas-grekas:
Thank you @alexandre-daubois.

PR #51976, opened 2023-10-10
[Workflow] Revert deprecation about Registry

### LINKED ISSUE #51400: [Workflow] Undeprecate `Registry::get`
### Description

Hi,

Our system relies heavily on getting a workflow from a object, aka `Registry::get()`

It was deprecated in #46000

But this seems to defeat dealing with many workflows that have a different support strategy, aka `framework.workflows.N.supports`, aka `WorkflowSupportStrategyInterface`

At this point im wondering what the alternative way is, or why `supports` configuration still exists?

### Example

_No response_

### LINKED ISSUE #51662: [Workflow] Deprecated Registry how to deal with
### Description

Hi,
Our application heavily use `Workflow` and have many validators, custom Twig function, and services that use an Entity or collection of Entity as entry, and check it against its workflow, to do it we use the `Registry` and it `get` method that return us the good workflow.

Since the `Registry` injection is deprecated, we knw that from 7.0 all the application should break, we can use something like:
`#[TaggedLocator(tag: 'workflow')]` or `#[TaggedIterator(tag: 'workflow')]` to inject all Workflow, but it seems there is not way to use retrieve the `supports` attribute to do it (find the good Workflow).

That said, even if this were the case, it's a shame to perform this operation on many validators and many services when an existing class already does it.

### Example

_No response_

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #51662, fix #49199, fix #51400
| License       | MIT

### DIFF
--- a/UPGRADE-6.2.md
+++ b/UPGRADE-6.2.md
@@ -132,10 +132,6 @@ VarDumper
 Workflow
 --------
 
- * The `Registry` is marked as internal and should not be used directly. use a tagged locator instead
-    ```
-    tagged_locator('workflow', 'name')
-    ```
  * The first argument of `WorkflowDumpCommand` should be a `ServiceLocator` of
    all workflows indexed by names
  * Deprecate calling `Definition::setInitialPlaces()` without arguments
--- a/src/Symfony/Component/Workflow/Registry.php
+++ b/src/Symfony/Component/Workflow/Registry.php
@@ -17,8 +17,6 @@ use Symfony\Component\Workflow\SupportStrategy\WorkflowSupportStrategyInterface;
 /**
  * @author Fabien Potencier <fabien@symfony.com>
  * @author Grégoire Pineau <lyrixx@lyrixx.info>
- *
- * @internal since Symfony 6.2. Inject the workflow where you need it.
  */
 class Registry
 {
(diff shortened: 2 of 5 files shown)


src/Symfony/Component/Workflow/Registry.php
```diff
@@ -17,8 +17,6 @@
 /**
  * @author Fabien Potencier <fabien@symfony.com>
  * @author Grégoire Pineau <lyrixx@lyrixx.info>
- *
- * @internal since Symfony 6.2. Inject the workflow where you need it.
```
ro0NL, line 21: perhaps move it to `addWorkflow` instead

stof: addWorkflow is not internal. It is the only way to make an instance of the registry useful.

ro0NL: excusi, there's no ctor -.-

UPGRADE-6.2.md
```diff
@@ -132,10 +132,6 @@ VarDumper
 Workflow
 --------
 
- * The `Registry` is marked as internal and should not be used directly. use a tagged locator instead
-    ```
-    tagged_locator('workflow', 'name')
-    ```
```
OskarStark, line 138: Shouldn't we keep this? back in time in 6.2 it still is deprecated

fabpot (approved):

fabpot:
Thank you @lyrixx.

PR #51983, opened 2023-10-11
[AssetMapper] [Bugfix] Added packageName and filePath to ImportMapEntry when requiring package

maelanleborgne:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | 
| License       | MIT

In #51845, 2 new properties have been added to `ImportMapEntry` that are not initialized in `ImportMapManager` when calling `importmap:require a_package`.
This PR fixes this.

### DIFF
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
@@ -295,13 +295,15 @@ class ImportMapManager
         $resolvedPackages = $this->resolver->resolvePackages($packagesToRequire);
         foreach ($resolvedPackages as $resolvedPackage) {
             $importName = $resolvedPackage->requireOptions->importName ?: $resolvedPackage->requireOptions->packageName;
-
+            [$packageName, $filePath] = ImportMapConfigReader::splitPackageNameAndFilePath($importName);
             $newEntry = new ImportMapEntry(
                 $importName,
                 path: $resolvedPackage->requireOptions->path,
                 version: $resolvedPackage->version,
                 type: $resolvedPackage->type,
                 isEntrypoint: $resolvedPackage->requireOptions->entrypoint,
+                packageName: $packageName,
+                filePath: $filePath,
             );
             $importMapEntries->add($newEntry);
             $addedEntries[] = $newEntry;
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -295,13 +295,15 @@ private function requirePackages(array $packagesToRequire, ImportMapEntries $imp
         $resolvedPackages = $this->resolver->resolvePackages($packagesToRequire);
         foreach ($resolvedPackages as $resolvedPackage) {
             $importName = $resolvedPackage->requireOptions->importName ?: $resolvedPackage->requireOptions->packageName;
-
+            [$packageName, $filePath] = ImportMapConfigReader::splitPackageNameAndFilePath($importName);
```
weaverryan, line 298:
```suggestion
            [$packageName, $filePath] = ImportMapConfigReader::splitPackageNameAndFilePath($resolvedPackage->requireOptions->packageName;);
```

This is a bit odd - it reflects why the two new properties were needed. When you use `importmap:require`, you can require a package but give it a different import name - e.g. `vue/dist/vue.esm-bundler.js=vue`. So, the `->packageName` here represents the "package + path" from that - `vue/dist/vue.esm-bundle.js`.

The fact that the import name (i.e. key) could possibly not equal the package+path for an entry in `importmap.php` is a bit of a problem now that we just store the `version` there instead of the full CDN path. Probably, when writing the `importmap.php`, if `importName != packageName / filePath`, then we should output a new `package` key that contains `packageName + filePath` (I wish I could think of a better name for this packageName + filePath combination :) ). 

Anyway, that can be a different PR - you can work on it if you'd like, but it's also on my list - I'm mostly thinking out loud.

weaverryan (changes requested):
Could you also update `JsDelivrEsmResolver::downloadPackages()` to throw an exception before the HTTP request if `$entry->packageName` or `$entry->version` is null?

maelanleborgne:
> Could you also update `JsDelivrEsmResolver::downloadPackages()` to throw an exception before the HTTP request if `$entry->packageName` or `$entry->version` is null?  

@weaverryan I'm having a hard time writing an exception message to describe the problem because I don't see how we could reach the `downloadPackages` method with an entry that has a null packageName of filePath.
The `splitPackageNameAndFilePath` method defaults packageName to `importName` and filePath to `''` when creating the `ImportMapEntries` in `ImportMapConfigReader`.

weaverryan:
Yes, the problem should actually NEVER happen. It's more of coding defensively... in case we make some mistake in the future. So, the error message doesn't need to be super specific. For example:

```php
if (null === $entry->packageName) {
    throw new \InvalidArgumentException(sprintf('Missing packageName for importmap entry "%s"., $entry->importName));
}
```

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
weaverryan: FYI - I'm working on the PR to make the `packageName + filePath` stuff a bit clearer, as it affects many places.

weaverryan:
I'm going to close this - but thanks for getting it going @maelanleborgne. Based on our conversation here, I could see how important it was to go and fix my ugly "packageName + filePath" everywhere. I've done that now on #52024

Thanks!

PR #51986, opened 2023-10-11
[Security] Do not match request twice in `HttpUtils`

Toflar:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | 
| License       | MIT

Found a pretty heavy performance issue in [Contao](https://contao.org) (thank you, Blackfire for sponsoring my Open Source license ❤️). It's probably not very apparent if you only use Symfony core components (due to fast regex matching) but as soon as you use dynamic router matching using e.g. [Symfony CMF Routing](https://github.com/symfony-cmf/Routing) with database hits like we do, you may encounter this issue.

The problem is that currently, `HttpUtils::checkRequestPath()` always matches a given `Request`, even if that has already been matched before. This is the **default** case in any Symfony application because

* In the `kernel.request` event stack, the `RouterListener` is called first. Route matching happens here.
* Later, the `FirewallListener` is called. In its stack, it calls `LogoutListener::supports()` which **always** calls `requiresLogout()` and thus triggers an additional match via `HttpUtils::checkRequestPath()` for every single request (https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Security/Http/Firewall/LogoutListener.php#L142).

So if route matching has already happened before, we should not match again.

### DIFF
--- a/src/Symfony/Component/Security/Http/HttpUtils.php
+++ b/src/Symfony/Component/Security/Http/HttpUtils.php
@@ -118,6 +118,11 @@ class HttpUtils
     public function checkRequestPath(Request $request, string $path)
     {
         if ('/' !== $path[0]) {
+            // Shortcut if request has already been matched before
+            if ($request->attributes->has('_route')) {
+                return $path === $request->attributes->get('_route');
+            }
+
             try {
                 // matching a request is more powerful than matching a URL path + context, so try that first
                 if ($this->urlMatcher instanceof RequestMatcherInterface) {
--- a/src/Symfony/Component/Security/Http/Tests/HttpUtilsTest.php
+++ b/src/Symfony/Component/Security/Http/Tests/HttpUtilsTest.php
@@ -318,6 +318,22 @@ class HttpUtilsTest extends TestCase
         $utils->checkRequestPath($this->getRequest(), 'foobar');
     }
 
+    public function testCheckRequestPathWithRequestAlreadyMatchedBefore()
+    {
+        $urlMatcher = $this->createMock(RequestMatcherInterface::class);
+        $urlMatcher
+            ->expects($this->never())
+            ->method('matchRequest')
+        ;
+
+        $request = $this->getRequest();
+        $request->attributes->set('_route', 'route_name');
+
+        $utils = new HttpUtils(null, $urlMatcher);
+        $this->assertTrue($utils->checkRequestPath($request, 'route_name'));
+        $this->assertFalse($utils->checkRequestPath($request, 'foobar'));
+    }
+
     public function testCheckPathWithoutRouteParam()
     {
         $urlMatcher = $this->createMock(UrlMatcherInterface::class);


aschempp (approved):

src/Symfony/Component/Security/Http/HttpUtils.php
```diff
@@ -118,6 +118,11 @@ public function createRequest(Request $request, string $path)
     public function checkRequestPath(Request $request, string $path)
     {
         if ('/' !== $path[0]) {
+            // Shortcut if request has already been matched before
+            if ($request->attributes->has('_route')) {
+                return $path === $request->attributes->get('_route');
+            }
```
wouterj, line 124: You might have already considered this, but what about instead caching this in the URL matcher? This means other calls to `matchRequest()` also benefit from this optimization. My initial feeling is that this logic is an "implementation detail" of the URL matcher, rather than the Security component.

Toflar: I have and it would be doable of course. Having it here just solves that one problem for potentially a lot of other users, not just for us :)

fabpot: I agree with Wouter, the `_route` attribute is an implementation detail we should not rely on here.

Toflar: What exactly is the implementation detail? The fact that we store `_route` as a request attribute? Or the fact that it is named `_route`? I mean, I get your point but I'd argue that the entire `HttpUtils::checkRequestPath()` only makes sense to be used in the Symfony fullstack framework kind of context because it already relies on the fact that the parameter is named `_route`. I mean, look at the entire method now, it's meant to check if the url matcher returns a matching `_route` parameter.

stof: Well, the UrlMatcher is guaranteed to return `_route` on matches. However, the URL matcher is not the one storing anything in request attributes (that's done by the RoutingListener of HttpKernel).

So to me, doing that optimization inside the routing component looks wrong to me. The component that guarantes that `$request->attributes->get('_route')` contains the matching route after matching is done is the HttpKernel one.

However, I would not treat that route attribute as an internal implementation detail of HttpKernel. It has been documented since years and lots of packages are relying on it.

stof: I vote for keeping the current implementation of this PR.

nicolas-grekas: We already rely on that parameter [10 lines below](https://github.com/symfony/symfony/pull/51986/files#diff-09ea53eab5f9bcb734da5c64182bfa774b1e3e63a209f2c1e444dd83ce7fd70aR134) so this makes sense as is to me also.

nicolas-grekas (approved):

chalasr (approved):

fabpot (approved):

fabpot:
Thank you @Toflar.

PR #51992, opened 2023-10-11
[Serializer] Fix using `DateIntervalNormalizer` with union types

Jeroeny:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

The union logic of `AbstractObjectNormalizer` tries to denormalize each union type and catches `NotNormalizableValueException` (among some other exceptions). If a non-catching exception is thrown, denormalization fails on that first type, while a later type might have succeeded. 

If I try to denormalize a `DateTimeInterface` value into a `DateInterval|DateTimeInterface`  type, it will fail because the `DateIntervalNormalizer` throws `UnexpectedValueException` instead of `NotNormalizableValueException`.   
Denormalizing a `DateInterval` into `DateTimeInterface|DateInterval` does work, because `DateTimeNormalizer` throws `NotNormalizableValueException`. I also checked some other Object-specific normalizers like `Uid`, `Problem`, `DateTimeZone`, `DataUri`, `BackedEnum`, they are using `NotNormalizableValueException`. 

See reproducer: https://github.com/Jeroeny/reproduce/tree/union/src

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/DateIntervalNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/DateIntervalNormalizer.php
@@ -12,7 +12,7 @@
 namespace Symfony\Component\Serializer\Normalizer;
 
 use Symfony\Component\Serializer\Exception\InvalidArgumentException;
-use Symfony\Component\Serializer\Exception\UnexpectedValueException;
+use Symfony\Component\Serializer\Exception\NotNormalizableValueException;
 
 /**
  * Normalizes an instance of {@see \DateInterval} to an interval string.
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Could it be possible to add a test case that covers the situation you describe, ie that uses a union type?

src/Symfony/Component/Serializer/Normalizer/DateIntervalNormalizer.php
```diff
@@ -70,17 +71,16 @@ public function hasCacheableSupportsMethod(): bool
...
      *
-     * @throws InvalidArgumentException
-     * @throws UnexpectedValueException
+     * @throws NotNormalizableValueException
      */
     public function denormalize($data, string $type, string $format = null, array $context = [])
     {
         if (!\is_string($data)) {
-            throw new InvalidArgumentException(sprintf('Data expected to be a string, "%s" given.', get_debug_type($data)));
+            throw NotNormalizableValueException::createForUnexpectedDataType('Data expected to be a string.', $data, [Type::BUILTIN_TYPE_STRING], $context['deserialization_path'] ?? null, true);
```
stof: This should use `string` instead of `Type::BUILTIN_TYPE_STRING` for the expected type as the PropertyInfo component might not be installed (and yes, we have the same bug in other normalizers)

Jeroeny:
@nicolas-grekas I think this tests the union type with a failure first: https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Serializer/Tests/SerializerTest.php#L740

But it won't test _every_ normalizer for throwing the correct exception. 

And secondly the `DateIntervalNormalizerTest` now does assert the exception is thrown for this normalizer.

HypeMC:
@nicolas-grekas Sure, I'll take a look sometime today.

src/Symfony/Component/Serializer/Normalizer/DateIntervalNormalizer.php
```diff
@@ -114,8 +113,10 @@ public function denormalize($data, string $type, string $format = null, array $c
             }
 
             return new \DateInterval($data);
+        } catch (NotNormalizableValueException $e) {
+            throw $e;
```
HypeMC: The `NotNormalizableValueException` is never thrown inside the try block, so this is not needed.

HypeMC (commented):
Looks good, tried it out and it works as expected.

src/Symfony/Component/Serializer/Normalizer/DateIntervalNormalizer.php
Jeroeny: Of course, removed it.

HypeMC (approved):

mtarld (approved):

nicolas-grekas (approved):

fabpot:
Thank you @Jeroeny.

PR #49911, opened 2023-04-03
[HttpClient] Support file uploads by nesting resource streams in `body` option

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

This PR makes it easy to upload files using `multipart/form-data`.

Nesting streams in option "body" is all one needs to do:
```php
$h = fopen('/path/to/the/file' 'r');
$client->request('POST', 'https://...', ['body' => ['the_file' => $h]]);
```

By default, the code will populate the `filename` using the base-name of the opened file.
It's possible to override this by calling `stream_context_set_option($h, 'http', 'filename', 'the-name.txt')`.

When forwarding an HTTP request coming either from the native `http` stream wrapper or from Symfony's `StreamableInterface`, the code will forward the content-type. It's also possible to set the content-type using `stream_context_set_option($h, 'http', 'content_type', 'my/content-type')`.

When possible, the code will generate a `Content-Length` header. This enables the destination server to reject the request early if it's too big and it allows tracking the progress of uploads on the client-side.

PR #51994, opened 2023-10-11
[Notifier] [Telegram] Fix version and exception signature

OskarStark:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | --
| License       | MIT

The PR
* https://github.com/symfony/symfony/pull/51717

introduced a new exception in the Notifier component. I tested this locally and faced the following error:
```
Class "Symfony\Component\Notifier\Exception\MultipleExclusiveOptionsUsedException" not found
```

The version bump of the notifier was missing.
I also added tests for the exception class.

I am wondering about the parameter signature, and I am not sure if the second parameter should be nullable:
https://github.com/symfony/symfony/blob/c868be44ab61ac02e978dd18438915bcaed1aa8d/src/Symfony/Component/Notifier/Exception/MultipleExclusiveOptionsUsedException.php#L17-L31

@igrizzli is there anything I am missing or can e adjust the signature:
```diff
- public function __construct(array $usedExclusiveOptions, array $exclusiveOptions = null, \Throwable $previous = null)
+ public function __construct(array $usedExclusiveOptions, array $exclusiveOptions, \Throwable $previous = null)
``

### DIFF
--- a/src/Symfony/Component/Notifier/Bridge/Telegram/Tests/TelegramTransportTest.php
+++ b/src/Symfony/Component/Notifier/Bridge/Telegram/Tests/TelegramTransportTest.php
@@ -1242,6 +1242,6 @@ JSON;
         $transport = self::createTransport($client, 'testChannel');
 
         $this->expectException(MultipleExclusiveOptionsUsedException::class);
-        $sentMessage = $transport->send(new ChatMessage('', $messageOptions));
+        $transport->send(new ChatMessage('', $messageOptions));
     }
 }
--- a/src/Symfony/Component/Notifier/Bridge/Telegram/composer.json
+++ b/src/Symfony/Component/Notifier/Bridge/Telegram/composer.json
@@ -17,8 +17,9 @@
     ],
     "require": {
         "php": ">=8.1",
-        "symfony/http-client": "^5.4|^6.0|^7.0",
-        "symfony/notifier": "^6.2.7|^7.0"
+        "symfony/http-client": "^6.3|^7.0",
+        "symfony/mime": "^5.4|^6.3|^7.0",
+        "symfony/notifier": "^6.4|^7.0"
     },
     "autoload": {
         "psr-4": { "Symfony\\Component\\Notifier\\Bridge\\Telegram\\": "" },
--- a/src/Symfony/Component/Notifier/Exception/MultipleExclusiveOptionsUsedException.php
+++ b/src/Symfony/Component/Notifier/Exception/MultipleExclusiveOptionsUsedException.php
@@ -17,15 +17,16 @@ namespace Symfony\Component\Notifier\Exception;
 class MultipleExclusiveOptionsUsedException extends InvalidArgumentException
 {
     /**
-     * @param string[]      $usedExclusiveOptions
-     * @param string[]|null $exclusiveOptions
+     * @param string[] $usedExclusiveOptions
+     * @param string[] $exclusiveOptions
      */
-    public function __construct(array $usedExclusiveOptions, array $exclusiveOptions = null, \Throwable $previous = null)
+    public function __construct(array $usedExclusiveOptions, array $exclusiveOptions, \Throwable $previous = null)
     {
-        $message = sprintf('Multiple exclusive options have been used "%s".', implode('", "', $usedExclusiveOptions));
-        if (null !== $exclusiveOptions) {
-            $message .= sprintf(' Only one of %s can be used.', implode('", "', $exclusiveOptions));
-        }
+        $message = sprintf(
+            'Multiple exclusive options have been used "%s". Only one of "%s" can be used.',
+            implode('", "', $usedExclusiveOptions),
+            implode('", "', $exclusiveOptions)
+        );
 
         parent::__construct($message, 0, $previous);
     }
--- a/src/Symfony/Component/Notifier/Tests/Exception/MultipleExclusiveOptionsUsedExceptionTest.php
+++ b/src/Symfony/Component/Notifier/Tests/Exception/MultipleExclusiveOptionsUsedExceptionTest.php
@@ -1,0 +1,28 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Notifier\Tests\Exception;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Notifier\Exception\MultipleExclusiveOptionsUsedException;
+
+class MultipleExclusiveOptionsUsedExceptionTest extends TestCase
+{
+    public function testMessage()
+    {
+        $exception = new MultipleExclusiveOptionsUsedException(['foo', 'bar'], ['foo', 'bar', 'baz']);
+
+        $this->assertSame(
+            'Multiple exclusive options have been used "foo", "bar". Only one of "foo", "bar", "baz" can be used.',
+            $exception->getMessage()
+        );
+    }
+}
--- a/src/Symfony/Component/Notifier/Bridge/Telegram/TelegramTransport.php
+++ b/src/Symfony/Component/Notifier/Bridge/Telegram/TelegramTransport.php
@@ -36,7 +36,7 @@ final class TelegramTransport extends AbstractTransport
     private string $token;
     private ?string $chatChannel;
 
-    public const EXCLUSIVE_OPTIONS = [
+    private const EXCLUSIVE_OPTIONS = [
         'message_id',
         'callback_query_id',
         'photo',

### CODE REVIEW

File: src/Symfony/Component/Notifier/Exception/MultipleExclusiveOptionsUsedException.php [hunk 1]
```diff
@@ -16,15 +16,11 @@
  */
 class MultipleExclusiveOptionsUsedException extends InvalidArgumentException
 {
-    /**
-     * @param string[]      $usedExclusiveOptions
-     * @param string[]|null $exclusiveOptions
-     */
-    public function __construct(array $usedExclusiveOptions, array $exclusiveOptions = null, \Throwable $previous = null)
+    public function __construct(array $usedOptions, array $availableOptions = null, \Throwable $previous = null)
```
igrizzli: I have concern about second argument - it's not list of available options, it's list of options which should be exclusive (not used together).

igrizzli:
```suggestion
    public function __construct(array $usedOptions, array $exclusiveOptions, \Throwable $previous = null)
```

src/Symfony/Component/Notifier/Exception/MultipleExclusiveOptionsUsedException.php
```diff
@@ -16,15 +16,11 @@
...
     {
-        $message = sprintf('Multiple exclusive options have been used "%s".', implode('", "', $usedExclusiveOptions));
-        if (null !== $exclusiveOptions) {
-            $message .= sprintf(' Only one of %s can be used.', implode('", "', $exclusiveOptions));
+        $message = sprintf('Multiple exclusive options have been used "%s".', implode('", "', $usedOptions));
+        if (null !== $availableOptions) {
```
igrizzli:
```suggestion
        if (count($exclusiveOptions) > 0) {
```
If we will make second argument not nullable, we should check that array not empty here, or remove this check at all and generate full message on previous line.

igrizzli (changes requested):

igrizzli:
Good catch, thank you.

src/Symfony/Component/Notifier/Tests/Exception/MultipleExclusiveOptionsUsedExceptionTest.php
```diff
@@ -0,0 +1,47 @@
...
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Bridge\PhpUnit\ClassExistsMock;
+use Symfony\Component\Notifier\Bridge;
+use Symfony\Component\Notifier\Exception\MultipleExclusiveOptionsUsedException;
+use Symfony\Component\Notifier\Exception\UnsupportedSchemeException;
+use Symfony\Component\Notifier\Transport\Dsn;
+
+/**
+ * @runTestsInSeparateProcesses
```
OskarStark:
```suggestion
```

src/Symfony/Component/Notifier/Exception/MultipleExclusiveOptionsUsedException.php
OskarStark: Done

src/Symfony/Component/Notifier/Exception/MultipleExclusiveOptionsUsedException.php
OskarStark: I removed the check

OskarStark:
Ready to merge from my side

src/Symfony/Component/Notifier/Tests/Exception/MultipleExclusiveOptionsUsedExceptionTest.php
```diff
@@ -0,0 +1,31 @@
...
+use Symfony\Component\Notifier\Exception\MultipleExclusiveOptionsUsedException;
+
+/**
+ * @see MultipleExclusiveOptionsUsedException
+ */
+final class MultipleExclusiveOptionsUsedExceptionTest extends TestCase
+{
+    public function testMessage()
+    {
+        $exception = new MultipleExclusiveOptionsUsedException(['foo', 'bar'], ['baz', 'qux']);
```
OskarStark:
```suggestion
        $exception = new MultipleExclusiveOptionsUsedException(['foo', 'bar'], ['foo', 'bar', 'baz']);
```

src/Symfony/Component/Notifier/Tests/Exception/MultipleExclusiveOptionsUsedExceptionTest.php
```diff
@@ -0,0 +1,31 @@
...
+ * @see MultipleExclusiveOptionsUsedException
+ */
+final class MultipleExclusiveOptionsUsedExceptionTest extends TestCase
+{
+    public function testMessage()
+    {
+        $exception = new MultipleExclusiveOptionsUsedException(['foo', 'bar'], ['foo', 'bar', 'baz']);
+
+        $this->assertSame(
+            'Multiple exclusive options have been used "foo", "bar". Only one of "baz", "qux" can be used.',
```
OskarStark:
```suggestion
            'Multiple exclusive options have been used "foo", "bar". Only one of "foo", "bar", "baz" can be used.',
```

igrizzli (approved):

src/Symfony/Component/Notifier/Bridge/Telegram/composer.json
```diff
@@ -18,7 +18,7 @@
     "require": {
         "php": ">=8.1",
         "symfony/http-client": "^5.4|^6.0|^7.0",
-        "symfony/notifier": "^6.2.7|^7.0"
+        "symfony/notifier": "^6.4|^7.0"
```
stof, line 22: This would have been caught by the prefer-lowest job if the tests of this bridges were including a test covering the validation of exclusive options. This is the test that should be added IMO, more than a test of the exception constructor in the component.

src/Symfony/Component/Notifier/Exception/MultipleExclusiveOptionsUsedException.php
```diff
@@ -17,15 +17,16 @@
 class MultipleExclusiveOptionsUsedException extends InvalidArgumentException
 {
     /**
-     * @param string[]      $usedExclusiveOptions
-     * @param string[]|null $exclusiveOptions
+     * @param string[] $usedOptions
+     * @param string[] $exclusiveOptions
      */
-    public function __construct(array $usedExclusiveOptions, array $exclusiveOptions = null, \Throwable $previous = null)
+    public function __construct(array $usedOptions, array $exclusiveOptions, \Throwable $previous = null)
```
stof: I don't think we should rename the argument. The code might have used other options that were not part of the exclusive options and those should not be passed to this argument.

src/Symfony/Component/Notifier/Bridge/Telegram/composer.json
OskarStark: You are right, will do it

src/Symfony/Component/Notifier/Exception/MultipleExclusiveOptionsUsedException.php
```diff
@@ -17,15 +17,16 @@
...
     {
-        $message = sprintf('Multiple exclusive options have been used "%s".', implode('", "', $usedExclusiveOptions));
-        if (null !== $exclusiveOptions) {
-            $message .= sprintf(' Only one of %s can be used.', implode('", "', $exclusiveOptions));
-        }
+        $message = sprintf(
+            'Multiple exclusive options have been used "%s". Only one of %s can be used.',
```
stof:
```suggestion
            'Multiple exclusive options have been used "%s". Only one of "%s" can be used.',
```

stof (changes requested):

stof:
The telegram tests are failing on CI with an error about an undefined key. Is it coming from that PR as well ?

src/Symfony/Component/Notifier/Bridge/Telegram/Tests/TelegramTransportTest.php
```diff
@@ -1242,6 +1242,6 @@ public function testUsingMultipleExclusiveOptionsWillProvideExceptions(TelegramO
         $transport = self::createTransport($client, 'testChannel');
 
         $this->expectException(MultipleExclusiveOptionsUsedException::class);
```
OskarStark, line 1244: We have a test here, so it should have failed 🤔

OskarStark:
> The telegram tests are failing on CI with an error about an undefined key. Is it coming from that PR as well ?

I will check later

OskarStark:
I cannot reproduce the missing key, I tried to composer update --prefer-lowest, but tests are still passing 🤔 

Besides that, I think we should not write json and decode it, but write an array and encode it in the assertion. But that should be a floor up PR once green

src/Symfony/Component/Notifier/Bridge/Telegram/Tests/TelegramTransportTest.php
OskarStark: And I can see it failing in `8.2, high deps`:
```
1) Symfony\Component\Notifier\Bridge\Telegram\Tests\TelegramTransportTest::testUsingMultipleExclusiveOptionsWillProvideExceptions with data set "edit" (Symfony\Component\Notifier\Bridge\Telegram\TelegramOptions Object (...))
PHPUnit\Framework\Exception: Class "Symfony\Component\Notifier\Exception\MultipleExclusiveOptionsUsedException" does not exist
```

But this shouldn't fail anymore, as I raised symfony/notifier to ^6.4 🤔

OskarStark:
@igrizzli any idea?

src/Symfony/Component/Notifier/Exception/MultipleExclusiveOptionsUsedException.php
OskarStark: reverted

OskarStark:
As far as @xabbuh and I can find out so far, dropping `symfony/http-client:^5.4` helps, so now we have the same errors in high and low deps test.

For the errors itself, maybe 
* https://github.com/symfony/symfony/pull/49911

is related....

src/Symfony/Component/Notifier/Tests/Exception/MultipleExclusiveOptionsUsedExceptionTest.php
```diff
@@ -0,0 +1,31 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Notifier\Tests\Exception;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Notifier\Exception\MultipleExclusiveOptionsUsedException;
+
+/**
+ * @see MultipleExclusiveOptionsUsedException
```
nicolas-grekas: should be removed

src/Symfony/Component/Notifier/Bridge/Telegram/composer.json
```diff
@@ -17,8 +17,8 @@
     ],
     "require": {
         "php": ">=8.1",
-        "symfony/http-client": "^5.4|^6.0|^7.0",
-        "symfony/notifier": "^6.2.7|^7.0"
+        "symfony/http-client": "^6.3|^7.0",
```
nicolas-grekas, line 20: why this bump?

nicolas-grekas (approved):
LGTM thanks, just minor details

src/Symfony/Component/Notifier/Bridge/Telegram/composer.json
xabbuh: we need #49911

OskarStark:
Thanks @xabbuh, looks like `symfony/mime` was missing

src/Symfony/Component/Notifier/Tests/Exception/MultipleExclusiveOptionsUsedExceptionTest.php
OskarStark: done

src/Symfony/Component/Notifier/Bridge/Telegram/composer.json
OskarStark: and for this we need `symfony/mime` also

src/Symfony/Component/Notifier/Tests/Exception/MultipleExclusiveOptionsUsedExceptionTest.php
```diff
@@ -0,0 +1,28 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Notifier\Tests\Exception;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Notifier\Exception\MultipleExclusiveOptionsUsedException;
+
+final class MultipleExclusiveOptionsUsedExceptionTest extends TestCase
```
nicolas-grekas: I'm removing `final` : a test case is not an API, we don't need extra visual debt (same as `void`) ;)

nicolas-grekas:
Thank you @OskarStark.

src/Symfony/Component/Notifier/Bridge/Telegram/Tests/TelegramTransportTest.php
stof: for high-deps, this will install the 7.0 component. Is the other PR merged up already ?

the high-deps job has this weird situation for the time were we have 2 active development branches

PR #44405, opened 2021-12-01
[Routing] Allow using services in the route condition

renanbr:
| Q             | A
| ------------- | ---
| Branch?       | 6.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

```php
#[AsRoutingConditionService(alias: 'foo')]
class SomeFooService
{
    public function bar(): bool { ... }
}
```

```php
class DefaultController
{
    #[Route('/page', condition: "service('foo').bar()")]
    public function page(): Response { ... }
} 
```

PR #51996, opened 2023-10-11
[FrameworkBundle][RemoteEvent][Routing][Scheduler] Add PHPDoc to attributes properties

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| License       | MIT

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Routing/Attribute/AsRoutingConditionService.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Routing/Attribute/AsRoutingConditionService.php
@@ -41,6 +41,10 @@ use Symfony\Component\DependencyInjection\Attribute\AutoconfigureTag;
 #[\Attribute(\Attribute::TARGET_CLASS)]
 class AsRoutingConditionService extends AutoconfigureTag
 {
+    /**
+     * @param string|null $alias    The alias of the service to use it in routing condition expressions
+     * @param int         $priority Defines a priority that allows the routing condition service to override a service with the same alias
+     */
     public function __construct(
         string $alias = null,
         int $priority = 0,
--- a/src/Symfony/Component/RemoteEvent/Attribute/AsRemoteEventConsumer.php
+++ b/src/Symfony/Component/RemoteEvent/Attribute/AsRemoteEventConsumer.php
@@ -17,6 +17,9 @@ namespace Symfony\Component\RemoteEvent\Attribute;
 #[\Attribute(\Attribute::TARGET_CLASS)]
 class AsRemoteEventConsumer
 {
+    /**
+     * @param string $name The name of the remote event consumer, used to identify it when defining remote events
+     */
     public function __construct(
         public string $name,
     ) {
--- a/src/Symfony/Component/Routing/Attribute/Route.php
+++ b/src/Symfony/Component/Routing/Attribute/Route.php
@@ -26,9 +26,21 @@ class Route
     private array $schemes;
 
     /**
-     * @param array<string|\Stringable> $requirements
-     * @param string[]|string           $methods
-     * @param string[]|string           $schemes
+     * @param string|array<string,string>|null $path         The route path (i.e. "/user/login")
+     * @param string|null                      $name         The route name (i.e. "app_user_login")
+     * @param array<string|\Stringable>        $requirements Requirements for the route attributes, @see https://symfony.com/doc/current/routing.html#parameters-validation
+     * @param array<string, mixed>             $options      Options for the route (i.e. ['prefix' => '/api'])
+     * @param array<string, mixed>             $defaults     Default values for the route attributes and query parameters
+     * @param string|null                      $host         The host for which this route should be active (i.e. "localhost")
+     * @param string|string[]                  $methods      The list of HTTP methods allowed by this route
+     * @param string|string[]                  $schemes      The list of schemes allowed by this route (i.e. "https")
+     * @param string|null                      $condition    An expression that must evaluate to true for the route to be matched, @see https://symfony.com/doc/current/routing.html#matching-expressions
+     * @param int|null                         $priority     The priority of the route if multiple ones are defined for the same path
+     * @param string|null                      $locale       The locale accepted by the route
+     * @param string|null                      $format       The format returned by the route (i.e. "json", "xml")
+     * @param bool|null                        $utf8         Whether the route accepts UTF-8 in its parameters
+     * @param bool|null                        $stateless    Whether the route is defined as stateless or stateful, @see https://symfony.com/doc/current/routing.html#stateless-routes
+     * @param string|null                      $env          The env in which the route is defined (i.e. "dev", "test", "prod")
      */
     public function __construct(
         string|array $path = null,
--- a/src/Symfony/Component/Scheduler/Attribute/AsPeriodicTask.php
+++ b/src/Symfony/Component/Scheduler/Attribute/AsPeriodicTask.php
@@ -19,6 +19,19 @@ namespace Symfony\Component\Scheduler\Attribute;
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
 class AsPeriodicTask
 {
+    /**
+     * @param string|int                          $frequency  A string (i.e. "every hour") or an integer (the number of seconds) representing the frequency of the task
+     * @param string|null                         $from       A string representing the start time of the periodic task (i.e. "08:00:00")
+     * @param string|null                         $until      A string representing the end time of the periodic task (i.e. "20:00:00")
+     * @param int|null                            $jitter     The cron jitter, in seconds; for example, if set to 60, the cron
+     *                                                        will randomly wait for a number of seconds between 0 and 60 before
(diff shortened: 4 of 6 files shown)


src/Symfony/Component/Routing/Annotation/Route.php
```diff
@@ -30,25 +32,67 @@ class Route
     private array $schemes;
 
     /**
-     * @param array<string|\Stringable> $requirements
-     * @param string[]|string           $methods
-     * @param string[]|string           $schemes
+     * @param string|array|null $path      The route path (i.e. "/user/login")
```
stof: `string|array<string, string>|null`

src/Symfony/Component/Scheduler/Attribute/AsPeriodicTask.php
```diff
@@ -20,13 +20,48 @@
...
         public readonly string $schedule = 'default',
+
+        /**
+         * The method to run as the task when the attribute target is a class.
+         */
         public readonly ?string $method = null,
+
+        /**
+         * One or many transports through which the message scheduling
+         * the task will go.
```
stof: `@var string[]|string|null`

alexandre-daubois:
I addressed all your comments, thank you for the review

src/Symfony/Component/Routing/Annotation/Route.php
```diff
@@ -15,7 +15,9 @@
  * Annotation class for @Route().
  *
  * @Annotation
+ *
  * @NamedArgumentConstructor
+ *
```
GromNaN: It seems that we don't want empty lines here: https://github.com/symfony/symfony/pull/52389

src/Symfony/Component/Routing/Annotation/Route.php
```diff
@@ -30,25 +32,71 @@ class Route
...
+         *
+         * @var array<string, mixed>
+         */
         private array $options = [],
+
+        /**
+         * The default values for the route attributes and
+         * query parameters.
+         *
+         * @var array<array-key, mixed>
```
GromNaN: Do we really accept integers as valid attributes?
```suggestion
         * @var array<string, mixed>
```

If we want to catch if someone forget the name of the attribute and write this incorrect form:
```
#Route('/route/{key}', name: 'app_route', defaults: ['default'])
```

src/Symfony/Component/Routing/Annotation/Route.php
```diff
@@ -30,25 +32,71 @@ class Route
...
+        /**
+         * A host for which the route will be matched.
+         */
         private ?string $host = null,
         array|string $methods = [],
         array|string $schemes = [],
+
+        /**
+         * An expression that must evaluate to true for the route
+         * to be matched (@see https://symfony.com/doc/current/routing.html#matching-expressions).
```
GromNaN: To avoid inline `@see` if possible.
```suggestion
         * An expression that must evaluate to true for the route to be matched.
         * 
         * @see https://symfony.com/doc/current/routing.html#matching-expressions
```

GromNaN (commented):
Could you rebase, the file `Route` class have been moved by https://github.com/symfony/symfony/pull/52294

alexandre-daubois:
Thanks for the review @GromNaN, I addressed your comments :+1:

src/Symfony/Component/Routing/Attribute/Route.php
```diff
@@ -30,25 +30,74 @@ class Route
...
+        /**
+         * Other route configuration like its prefix (i.e. ['prefix' => '/api']).
+         *
+         * @var array<string, mixed>
+         */
         private array $options = [],
+
+        /**
+         * The default values for the route attributes and
+         * query parameters.
```
GromNaN: Line break not necessary. It's 76 characters long.
```suggestion
         * The default values for the route attributes and query parameters.
```

GromNaN (approved):
LGTM

src/Symfony/Component/Routing/Annotation/Route.php
GromNaN: Oh, fabbot wants to add this newline. That's annoying.

alexandre-daubois: If I remember correctly, we ignored those precise fabbot fixes in the past 🙂

src/Symfony/Component/RemoteEvent/Attribute/AsRemoteEventConsumer.php
```diff
@@ -18,6 +18,9 @@
 class AsRemoteEventConsumer
 {
     public function __construct(
+        /**
+         * The name of the remote event consumer, used to identify it when defining remote events.
```
nicolas-grekas: all these inline comments should be turned into `@param` annotations

alexandre-daubois: Forgot about this PR, will update accordingly with what's been done lately. Thanks!

alexandre-daubois:
I updated with `@param` annotation and also avoid "sentences punctuation" by using `;` where possible

src/Symfony/Bundle/FrameworkBundle/Routing/Attribute/AsRoutingConditionService.php
```diff
@@ -41,6 +41,10 @@
 #[\Attribute(\Attribute::TARGET_CLASS)]
 class AsRoutingConditionService extends AutoconfigureTag
 {
+    /**
+     * @param string|null $alias    The alias of the service to use it in routing condition expressions
+     * @param int         $priority The priority of the service when gathered in a service iterator or locator
```
nicolas-grekas: the "when" part is obscure to me, it tells about internal things, isn't it? what's the use case for end users?

src/Symfony/Component/Routing/Attribute/Route.php
```diff
@@ -26,9 +26,21 @@ class Route
...
+     * @param array<string|\Stringable>        $requirements Requirements for the route attributes, @see https://symfony.com/doc/current/routing.html#parameters-validation
+     * @param array<string, mixed>             $options      Options for the route (i.e. ['prefix' => '/api'])
+     * @param array<string, mixed>             $defaults     Default values for the route attributes and query parameters
+     * @param string|null                      $host         The host for which this route should be active (i.e. "localhost")
+     * @param string|string[]                  $methods      The list of HTTP methods allowed by this route
+     * @param string|string[]                  $schemes      The list of schemes allowed by this route (i.e. "https")
+     * @param string|null                      $condition    An expression that must evaluate to true for the route to be matched, @see https://symfony.com/doc/current/routing.html#matching-expressions
+     * @param int|null                         $priority     The evaluation priority of the route if multiple are defined with the same path
```
nicolas-grekas:
```suggestion
     * @param int|null                         $priority     The priority of the route if multiple ones are defined for the same path
```

src/Symfony/Bundle/FrameworkBundle/Routing/Attribute/AsRoutingConditionService.php
alexandre-daubois: Hi @renanbr, it seems you created this attribute. Could you please help me clearly understand what's the use of the `priority` parameter? I'm having a look at the code, here exactly:

https://github.com/symfony/symfony/pull/44405/files#diff-dad788051d1d21f8285f23cd25cbec3e9068393ae468fef0f359acd2112bea15R32

Because we're using `service(...)` to gather _one_ routing condition service at a time, I'm not sure to understand to goal of setting a priority on this for the end user?

Thanks!

renanbr: 👋🏼 the goal is to allow overriding a service with the same alias

alexandre-daubois: Crystal clear. Thank you very much :pray:

nicolas-grekas (approved):

src/Symfony/Component/Scheduler/Attribute/AsPeriodicTask.php
```diff
@@ -19,6 +19,19 @@
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
 class AsPeriodicTask
 {
+    /**
+     * @param string|int                          $frequency  A string or an integer representing the frequency of the task (i.e. "every hour")
```
fabpot:
```suggestion
     * @param string|int                          $frequency  A string (i.e. "every hour"), or an integer (the number of seconds) representing the frequency of the task
```

fabpot (changes requested):

src/Symfony/Component/Scheduler/Attribute/AsPeriodicTask.php
alexandre-daubois: Way better, thanks :+1:

kbond (approved):

fabpot (approved):

fabpot:
Thank you @alexandre-daubois.

PR #52002, opened 2023-10-12
[HttpFoundation] Cookies Having Independent Partitioned State (CHIPS)

fabricecw:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT

Due to [Chrome's roadmap](https://developer.chrome.com/docs/privacy-sandbox/third-party-cookie-phase-out/) (and all other major browsers) to phase out third-party cookies starting from midway through 2024, "partitioned" cookies were introduced. If a cookie is flagged as `partitioned`, its cross-site boundry is tied to the top-level site.

Considerations: According to current security design, browser will only accept partitioned cookies with the `secure` flag and `SameSite` attribute `none` (otherwise it isn't a third-party cookie...). I classified this as an implementation topic and therefore omitted this validation in the Cookie class itself.

Further information:
- [Chrome for Developers](https://developer.chrome.com/docs/privacy-sandbox/chips/)
- [Mozilla Dev](https://developer.mozilla.org/en-US/docs/Web/Privacy/Partitioned_cookies)
- [CHIPS](https://github.com/privacycg/CHIPS)

### DIFF
--- a/src/Symfony/Component/HttpFoundation/CHANGELOG.md
+++ b/src/Symfony/Component/HttpFoundation/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
  * Make `HeaderBag::getDate()`, `Response::getDate()`, `getExpires()` and `getLastModified()` return a `DateTimeImmutable`
  * Support root-level `Generator` in `StreamedJsonResponse`
  * Add `UriSigner` from the HttpKernel component
+ * Add `partitioned` flag to `Cookie` (CHIPS Cookie)
 
 6.3
 ---
--- a/src/Symfony/Component/HttpFoundation/Cookie.php
+++ b/src/Symfony/Component/HttpFoundation/Cookie.php
@@ -32,6 +32,7 @@ class Cookie
 
     private bool $raw;
     private ?string $sameSite = null;
+    private bool $partitioned = false;
     private bool $secureDefault = false;
 
     private const RESERVED_CHARS_LIST = "=,; \t\r\n\v\f";
@@ -51,6 +52,7 @@ class Cookie
             'httponly' => false,
             'raw' => !$decode,
             'samesite' => null,
+            'partitioned' => false,
         ];
 
         $parts = HeaderUtils::split($cookie, ';=');
@@ -66,17 +68,20 @@ class Cookie
             $data['expires'] = time() + (int) $data['max-age'];
         }
 
-        return new static($name, $value, $data['expires'], $data['path'], $data['domain'], $data['secure'], $data['httponly'], $data['raw'], $data['samesite']);
+        return new static($name, $value, $data['expires'], $data['path'], $data['domain'], $data['secure'], $data['httponly'], $data['raw'], $data['samesite'], $data['partitioned']);
     }
 
     /**
      * @see self::__construct
      *
      * @param self::SAMESITE_*|''|null $sameSite
+     * @param bool                     $partitioned
      */
-    public static function create(string $name, string $value = null, int|string|\DateTimeInterface $expire = 0, ?string $path = '/', string $domain = null, bool $secure = null, bool $httpOnly = true, bool $raw = false, ?string $sameSite = self::SAMESITE_LAX): self
+    public static function create(string $name, string $value = null, int|string|\DateTimeInterface $expire = 0, ?string $path = '/', string $domain = null, bool $secure = null, bool $httpOnly = true, bool $raw = false, ?string $sameSite = self::SAMESITE_LAX /* , bool $partitioned = false */): self
     {
-        return new self($name, $value, $expire, $path, $domain, $secure, $httpOnly, $raw, $sameSite);
+        $partitioned = 9 < \func_num_args() ? func_get_arg(9) : false;
+
+        return new self($name, $value, $expire, $path, $domain, $secure, $httpOnly, $raw, $sameSite, $partitioned);
     }
 
     /**
@@ -92,7 +97,7 @@ class Cookie
      *
      * @throws \InvalidArgumentException
      */
-    public function __construct(string $name, string $value = null, int|string|\DateTimeInterface $expire = 0, ?string $path = '/', string $domain = null, bool $secure = null, bool $httpOnly = true, bool $raw = false, ?string $sameSite = self::SAMESITE_LAX)
+    public function __construct(string $name, string $value = null, int|string|\DateTimeInterface $expire = 0, ?string $path = '/', string $domain = null, bool $secure = null, bool $httpOnly = true, bool $raw = false, ?string $sameSite = self::SAMESITE_LAX, bool $partitioned = false)
     {
         // from PHP source code
         if ($raw && false !== strpbrk($name, self::RESERVED_CHARS_LIST)) {
@@ -112,6 +117,7 @@ class Cookie
         $this->httpOnly = $httpOnly;
         $this->raw = $raw;
         $this->sameSite = $this->withSameSite($sameSite)->sameSite;
+        $this->partitioned = $partitioned;
     }
 
     /**
@@ -238,6 +244,17 @@ class Cookie
     }
 
     /**
+     * Creates a cookie copy that is tied to the top-level site in cross-site context.
+     */
+    public function withPartitioned(bool $partitioned = true): static
+    {
+        $cookie = clone $this;
+        $cookie->partitioned = $partitioned;
+
+        return $cookie;
+    }
+
+    /**
      * Returns the cookie as a string.
      */
     public function __toString(): string
@@ -268,16 +285,20 @@ class Cookie
             $str .= '; domain='.$this->getDomain();
         }
 
-        if (true === $this->isSecure()) {
+        if ($this->isSecure()) {
             $str .= '; secure';
         }
 
-        if (true === $this->isHttpOnly()) {
+        if ($this->isHttpOnly()) {
             $str .= '; httponly';
         }
 
         if (null !== $this->getSameSite()) {
             $str .= '; samesite='.$this->getSameSite();
+        }
+
+        if ($this->isPartitioned()) {
+            $str .= '; partitioned';
         }
 
         return $str;
@@ -363,6 +384,14 @@ class Cookie
     public function isRaw(): bool
     {
         return $this->raw;
+    }
+
+    /**
+     * Checks whether the cookie should be tied to the top-level site in cross-site context.
+     */
+    public function isPartitioned(): bool
+    {
+        return $this->partitioned;
     }
 
     /**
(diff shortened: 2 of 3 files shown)


src/Symfony/Component/HttpFoundation/Cookie.php
```diff
@@ -22,13 +22,14 @@ class Cookie
...
-    protected $path;
-    protected $secure;
-    protected $httpOnly;
+    protected string $name;
+    protected ?string $value;
+    protected ?string $domain;
+    protected int $expire;
+    protected string $path;
+    protected ?bool $secure;
+    protected bool $httpOnly;
```
OskarStark: this must be reverted IMHO, as it's a BC break, isn't it? 🤔

src/Symfony/Component/HttpFoundation/Tests/CookieTest.php
```diff
@@ -187,6 +187,17 @@ public function testIsHttpOnly()
         $this->assertTrue($cookie->isHttpOnly(), '->isHttpOnly() returns whether the cookie is only transmitted over HTTP');
     }
 
+    public function testIsPartitioned()
+    {
+        $cookie = Cookie::create('foo', 'bar', 0, '/', '.myfoodomain.com', true, true,false, 'Lax', true);
+
+        $this->assertTrue($cookie->isPartitioned(),);
```
OskarStark:
```suggestion
        $this->assertTrue($cookie->isPartitioned());
```

src/Symfony/Component/HttpFoundation/Tests/CookieTest.php
```diff
@@ -187,6 +187,17 @@ public function testIsHttpOnly()
...
+
+        $cookie = Cookie::create('foo')->withPartitioned(true);
+
+        $this->assertTrue($cookie->isPartitioned(),);
```
OskarStark:
```suggestion
        $this->assertTrue($cookie->isPartitioned());
```

src/Symfony/Component/HttpFoundation/Tests/CookieTest.php
```diff
@@ -262,6 +273,20 @@ public function testToString()
...
+
+        $cookie = Cookie::create('foo')
+            ->withExpires(1)
+            ->withPath('/admin/')
+            ->withDomain('.myfoodomain.com')
+            ->withSecure(true)
+            ->withHttpOnly(true)
+            ->withSameSite('none')
+            ->withPartitioned(true);
+        $this->assertEquals($expected, (string) $cookie, '->__toString() returns string representation of a cleared cookie if value is NULL');
```
OskarStark, line 301:
```suggestion
        $this->assertSame($expected, (string) $cookie, '->__toString() returns string representation of a cleared cookie if value is NULL');
```
? could be changed for the others as well I guess

OskarStark:
Please rebase your branch. We cannot merge PRs containing merge commits. Thanks

src/Symfony/Component/HttpFoundation/Cookie.php
fabricecw: The constructor's signature and setter methods are type-hinted, so there wouldn't be any BC. But an extended class using these parameters with different types would lead to a BC. I considered this as an "unlikely BC" and a "misuse" of class attributes. But anyway, it could be a BC.

Let me know what's your guideline here and I will fix it.

OskarStark: Before you could overwrite the variable with a non matching type, which is not possible anymore, so from a technical POV this is a BC break for me.

I agree that this is unlikely, but lets ask @nicolas-grekas.

src/Symfony/Component/HttpFoundation/Tests/CookieTest.php
fabricecw: removed typo

src/Symfony/Component/HttpFoundation/Tests/CookieTest.php
fabricecw: Changed on all CookieTest. There are two tests `testConstructorWithDateTime` and `testConstructorWithDateTimeImmutable` with an ambigious assertion (integer value with string from DateTime). I will correct this in a dedicated commit.

fabricecw:
> Please rebase your branch. We cannot merge PRs containing merge commits. Thanks

Thanks @OskarStark, that was unintentional. Cleaned up.

Nyholm:
Lovely. Thank you for this PR. What a great first contribution!

I agree with all your changes, but I am a bit hesitant on adding types to protected properties (which is not mentioned in our [BC policy](https://symfony.com/doc/current/contributing/code/bc.html)). If you want this PR to get a quick merge, I suggest you remove all unrelated changes. Example: Changing the existing properties and doing `assertSame()` on existing tests. 

Doing this will make sure the discussions are only focus on the `partitioned` cookie feature. 

Also, can you please have a look on the fabbot.io CI job?

fabricecw:
I agree. Apart from a very low likelyhood of a BC, typed attributes can wait until a major release. I further reverted the `assertSame` change as suggested by @OskarStark even I like the stricter assertions.

All `HttpFoundation` tests are passing.

src/Symfony/Component/HttpFoundation/Tests/CookieTest.php
```diff
@@ -321,6 +346,9 @@ public function testFromString()
 
         $cookie = Cookie::fromString('foo_cookie=foo==; expires=Tue, 22 Sep 2020 06:27:09 GMT; path=/');
         $this->assertEquals(Cookie::create('foo_cookie', 'foo==', strtotime('Tue, 22 Sep 2020 06:27:09 GMT'), '/', null, false, false, true, null), $cookie);
```
OskarStark, line 361: We should add `false` for the `$partitioned` argument in the current tests

fabricecw: If we do so, I suggest an additional "minimal construction" test which compares a minimum object instantiation against a fully defined object (with the current default parameter values). This would cover modifications of the constructor's default parameter. Otherwise, all tests would still pass if you change a constructor's parameter default value since we overrule it in the tests. What do you think?

OskarStark: Why not 👍

fabricecw: added minimal constructor and `create` test to cover the "default cookie state".

src/Symfony/Component/HttpFoundation/Cookie.php
```diff
@@ -29,6 +29,7 @@ class Cookie
     protected $path;
     protected $secure;
     protected $httpOnly;
+    protected $partitioned;
```
stof: The new property should be private (as we have no BC reasons to keep it protected because of the older versions having it protected). For instance, `$sameSite` was also added as private.

and the new property should be typed.

src/Symfony/Component/HttpFoundation/Cookie.php
```diff
@@ -66,17 +68,17 @@ public static function fromString(string $cookie, bool $decode = false): static
...
+        return new static($name, $value, $data['expires'], $data['path'], $data['domain'], $data['secure'], $data['httponly'], $data['raw'], $data['samesite'], $data['partitioned']);
     }
 
     /**
      * @see self::__construct
      *
      * @param self::SAMESITE_*|''|null $sameSite
      */
-    public static function create(string $name, string $value = null, int|string|\DateTimeInterface $expire = 0, ?string $path = '/', string $domain = null, bool $secure = null, bool $httpOnly = true, bool $raw = false, ?string $sameSite = self::SAMESITE_LAX): self
+    public static function create(string $name, string $value = null, int|string|\DateTimeInterface $expire = 0, ?string $path = '/', string $domain = null, bool $secure = null, bool $httpOnly = true, bool $raw = false, ?string $sameSite = self::SAMESITE_LAX, bool $partitioned = false): self
```
stof: Adding a new argument in a non-final method is a BC break (even when static as a child class could redefine the static method and PHP would enforce the signature compatibility).
This must be added as a forward-compability argument instead.

stof (changes requested):

src/Symfony/Component/HttpFoundation/Cookie.php
fabricecw: That's right, completely missed that. It's now implemented as a forward-compatibility argument. I also changed the tests and suggest to keep the assertions by using object instantiation. 

What do you think about another PR to extend the `create` method with the `$partitioned` argument in the next major release?

fabricecw:
Status: Needs Review

nicolas-grekas (approved):
LGTM thanks. I think this feature is important to have for 6.4!

I pushed some minor changes, see https://github.com/symfony/symfony/compare/9e0b944c1f6a0ceeb005eba148c7e5aba19c257f..9d1d4415914b7d4412c93d41ebae0f2db3c1b221

fabricecw:
Thanks @nicolas-grekas! That forward compatibility modification is great.

As you said, it makes sense (if possible) to ship it with 6.4 so we don't fall behind the current pace of major browsers.

src/Symfony/Component/HttpFoundation/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ CHANGELOG
  * Make `HeaderBag::getDate()`, `Response::getDate()`, `getExpires()` and `getLastModified()` return a `DateTimeImmutable`
  * Support root-level `Generator` in `StreamedJsonResponse`
  * Add `UriSigner` from the HttpKernel component
+ * Add `partitioned` flag to `Cookie` (CHIP Cookie)
```
stof: If you want to use the acronym, it must be `CHIPS`. The S is for "Storage", not for a plural.

nicolas-grekas: I fixed it

fabricecw: You were quicker than me. :-) Let me force-push a more descriptive docblock for the new methods since the current one is a bit ambigious.

fabricecw: Modified docblock, see https://github.com/symfony/symfony/compare/3b54b4cbdf70a9735c6c875fa0319838670f595e..26df07bce02e50fbd261212814dd3e4dd74bab5c

stof (approved):

welcoMattic (approved):

fabpot (approved):

fabpot:
Thank you @fabricecw.

PR #52005, opened 2023-10-12
[Translation] Prevent creating empty keys when key ends with a period

### LINKED ISSUE #51536: [Translation] extract-option 'as-tree' creates empty property keys in Yaml

5.2.0 and higher

### Description

When generating yaml translation files and using the `--as-tree=3` option on the `translation:extract` command, the Translation component generates a tree of translatable strings. Therefore, it [explodes the translatable string by the `.`](https://github.com/symfony/symfony/blob/73a6b4b8ad315d294fe18c03012019694863daa9/src/Symfony/Component/Translation/Util/ArrayConverter.php#L39), which can result in odd behavior when the user is not using translation keys, but full sentences. For example, a lot of validation error messages end in with a dot, resulting in yaml like this:

```yaml
'This value should be false':
    '': 'This value should be false.'
'This value should be true':
    '': 'This value should be true.'
'This value should be of type {{ type }}':
    '': 'This value should be of type {{ type }}.'
```

While it's impossible to prevent this behavior when dots are in the middle of a strings, I think dots at the start and end of a translation key are supposed to be trimmed.

### How to reproduce

`symfony create --webapp` and then `bin/console translation:extract --force --format=yaml --as-tree=3 en`

### Possible Solution

_No response_

### Additional Context

_No response_

javleds:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #51536 
| License       | MIT

Small fix that prevents creating empty keys on translations when the translation key ends with a period, example:
```php
'This is a validation.'
```

Before the fix, the yaml output was:
```yaml
'This is a validation':
         '': 'This is a validation.'
```

After the fix we got:
```yaml
'This is a validation.': 'This is a validation.'
```

This can be replicated by:
1. Create a new symfony project using the `--webapp` flag.
2. Execute the command to extract translations as yaml: `php bin/console translation:extract --force --format=yaml --as-tree=3 --prefix='' --clean en`

Results:
Before the fix:
image

After the fix:
image

### DIFF
--- a/src/Symfony/Component/Translation/Tests/Util/ArrayConverterTest.php
+++ b/src/Symfony/Component/Translation/Tests/Util/ArrayConverterTest.php
@@ -69,6 +69,34 @@ class ArrayConverterTest extends TestCase
                     ],
                 ],
             ],
+            [
+                // input
+                [
+                    'foo.' => 'foo.',
+                    '.bar' => '.bar',
+                    'abc.abc' => 'value',
+                    'bcd.bcd.' => 'value',
+                    '.cde.cde.' => 'value',
+                    '.def.def' => 'value',
+                ],
+                // expected output
+                [
+                    'foo.' => 'foo.',
+                    '.bar' => '.bar',
+                    'abc' => [
+                        'abc' => 'value',
+                    ],
+                    'bcd' => [
+                        'bcd.' => 'value',
+                    ],
+                    '.cde' => [
+                        'cde.' => 'value',
+                    ],
+                    '.def' => [
+                        'def' => 'value',
+                    ],
+                ],
+            ],
         ];
     }
 }
--- a/src/Symfony/Component/Translation/Util/ArrayConverter.php
+++ b/src/Symfony/Component/Translation/Util/ArrayConverter.php
@@ -38,7 +38,7 @@ class ArrayConverter
         $tree = [];
 
         foreach ($messages as $id => $value) {
-            $referenceToElement = &self::getElementByPath($tree, explode('.', $id));
+            $referenceToElement = &self::getElementByPath($tree, self::getKeyParts($id));
 
             $referenceToElement = $value;
 
@@ -65,6 +65,7 @@ class ArrayConverter
                 $elem = &$elem[implode('.', \array_slice($parts, $i))];
                 break;
             }
+
             $parentOfElem = &$elem;
             $elem = &$elem[$part];
         }
@@ -95,5 +96,46 @@ class ArrayConverter
                 self::cancelExpand($tree, $prefix.$id, $value);
             }
         }
+    }
+
+    private static function getKeyParts(string $key)
+    {
+        $parts = explode('.', $key);
+        $partsCount = \count($parts);
+
+        $result = [];
+        $buffer = '';
+
+        foreach ($parts as $index => $part) {
+            if (0 === $index && '' === $part) {
+                $buffer = '.';
+
+                continue;
+            }
+
+            if ($index === $partsCount - 1 && '' === $part) {
+                $buffer .= '.';
+                $result[] = $buffer;
+
+                continue;
+            }
+
+            if (isset($parts[$index + 1]) && '' === $parts[$index + 1]) {
+                $buffer .= $part;
+
+                continue;
+            }
+
+            if ($buffer) {
+                $result[] = $buffer.$part;
+                $buffer = '';
+
+                continue;
+            }
+
+            $result[] = $part;
+        }
+
+        return $result;
     }
 }


src/Symfony/Component/Translation/Tests/Util/ArrayConverterTest.php
```diff
@@ -69,6 +69,20 @@ public static function messagesData()
...
+            [
+                // input
+                [
+                    'foo.' => 'foo.',
+                    'bar.' => 'bar.',
+                    'foo bar.' => 'foo bar.',
+                ],
+                // expected output
+                [
+                    'foo' => 'foo.',
```
squrious: As mentioned [here](https://github.com/symfony/symfony/issues/51536#issuecomment-1703371496), the periods should not be trimmed, as it modifies the keys and leads to inconsistent output files with the extracted translations. 

We could also test that periods at the beginning of the key don't create empty keys in the tree :smiley:

javleds: Nice catch.

In order to cover both scenarios a new `getKeyParts()` function was added, this ensures the message keys contain the expected strings, and, additional test cases were added with the suggestions done.

Thanks in advance.

src/Symfony/Component/Translation/Util/ArrayConverter.php
```diff
@@ -96,4 +97,45 @@ private static function cancelExpand(array &$tree, string $prefix, array $node)
             }
         }
     }
+
+    private static function getKeyParts(string $key)
```
welcoMattic, line 101: I think we can simplify this method with something like:

```
function getKeyParts(string $key) {
    $parts = explode('.', $key);
    $nb = count($parts);

    if (str_starts_with($key, '.')) {
        $parts[0] = '.';
    }
    if (str_ends_with($key, '.')) {
        $parts[$nb-1] = '.';
    }
    return $parts;
}
```

javleds: Thanks @welcoMattic. I love short code...

Nevertheless the shared code has 2 small disadvantages:

1. It is using functions available from php 8.0^ (str_starts_with and, str_ends_with), this code is intended to be part of the version 5.4 which still supports older versions of php.
2. Even if this covers the main goal (which is prevent empty keys with dots at the begining and, in the end) this is not covering some edge cases:

```
'abc 1.abc 2' => 'value',
'bcd 1.bcd 2.' => 'value',
'.cde 1.cde 2.' => 'value',
'.def 1.def 2' => 'value',
```

Please see [the test](https://github.com/symfony/symfony/pull/52005/files#diff-972cebacde3182f368ea028839c181153a8ba996565dc88c4854e65b8a6bd343R90) of this section for more info.

If I am not addressing something, please let me know, I will be glad to adjust the PR if needed.

94noni: for php version, you can leverage [symfony/polyfill-php](https://packagist.org/search/?query=symfony%2Fpolyfill-php) (find the one v that introduced the relevant function)

squrious: The polyfill that introduces `str_starts_with` and `str_ends_with` is [symfony/polyfill-php80](https://packagist.org/packages/symfony/polyfill-php80), and it is [already required](https://github.com/javleds/symfony/blob/ticket_51536_fix_empty_yam_tranlation_keys/src/Symfony/Component/Translation/composer.json#L22) in the Translation component so using those functions is ok imo

94noni: ho indeed cool if already included

src/Symfony/Component/Translation/Tests/Util/ArrayConverterTest.php
```diff
@@ -69,6 +69,38 @@ public static function messagesData()
                     ],
                 ],
             ],
+            [
+                // input
+                [
+                    'foo.' => 'foo.',
+                    'bar.' => 'bar.',
```
welcoMattic:
```suggestion
                    '.bar' => '.bar',
```
As you already test trailing period in 1 word-key on the previous line

src/Symfony/Component/Translation/Util/ArrayConverter.php
welcoMattic: @javleds It seems the edge cases you mention are covered by simplier code here:
https://3v4l.org/2ic9k
```php
function getKeyParts(string $key) {
    $parts = explode('.', $key);
    $nb = count($parts);

    if (str_starts_with($key, '.')) {
        $parts[0] = '.';
    }
    if (str_ends_with($key, '.')) {
        $parts[$nb-1] = '.';
    }
    return $parts;
}


var_dump(getKeyParts('abc 1.abc 2'));
var_dump(getKeyParts('bcd 1.bcd 2.'));
var_dump(getKeyParts('.cde 1.cde 2.'));
var_dump(getKeyParts('.def 1.def 2'));
```

welcoMattic: @javleds It seems the edge cases you mention are covered by simplier code here:
https://3v4l.org/2ic9k
```php
function getKeyParts(string $key) {
    $parts = explode('.', $key);
    $nb = count($parts);

    if (str_starts_with($key, '.')) {
        $parts[0] = '.';
    }
    if (str_ends_with($key, '.')) {
        $parts[$nb-1] = '.';
    }
    return $parts;
}


var_dump(getKeyParts('abc 1.abc 2'));
var_dump(getKeyParts('bcd 1.bcd 2.'));
var_dump(getKeyParts('.cde 1.cde 2.'));
var_dump(getKeyParts('.def 1.def 2'));
```

javleds: > Thanks for the advise of php polyfill, I will keep in mind in the case we need to perform additional updates. 🙏 

Getting back to the code discussion:

The real problem here is to avoid trimming the periods at the beginning/end of the parts. Since the `getElementByPath` function is itended to process each part as a valid string. Having empty positions or a position wihtn a single dot won't give the expected result.

For example:
- `['foo', 'bar']` will give the expected result.
- `['foo', '', 'bar']` won't give the expected result due the empty position.
- `['foo', 'bar', '.']` won't give the expected result due the period at position 2.
- `['.foo', 'bar.']` will give the expected result.

The simpler solution is to keep the period concatenated to the key part, and prevent the period occupate a single possition in the array. Otherwise, we will have to add extra processing inside the `getElementByPath` function.

Having this in mind, the code proposed above has small variations on the expected result.

Let me give the examples:

| input     | expected         | output with current fix | output with simpler code |
|-----------|------------------|-------------------------|--------------------------|
| abc.abc   | ['abc', 'abc']   | ['abc', 'abc']          | ['abc', 'abc']           |
| bcd.bcd.  | ['bcd', 'bcd.']  | ['bcd', 'bcd.']         | ['bcd', 'bcd', '.']      |
| .cde.cde. | ['.cde', 'cde.'] | ['.cde', 'cde.']        | ['.', 'cde', 'cde', '.'] |
| .def.def  | ['.def', 'def']  | ['.def', 'def']         | ['.','def', 'def']       |


If the proposed solution is hard to read, maybe, we can create named variables for every `if` statement.

src/Symfony/Component/Translation/Tests/Util/ArrayConverterTest.php
javleds: Right.

Both cases were intended to test no affectation with or without blank spaces, but, since the spaces are no affecting the output we will simplify reading by removing those (we can remove 2 entries lines 77 and, 78).

Next update will contain this cahnges.

stof (approved):

nicolas-grekas:
Thank you @javleds.

Issue #51393, opened 2023-08-15
[RFC][Validator] Remove constraint construction from associative arrays


derrabus:
### Description

From @nicolas-grekas' comment: https://github.com/symfony/symfony/issues/51381#issuecomment-1677066808

> For validation, I'm also wondering if the move shouldn't do one more step and clean the way we create those objects. Right now, constructors are a bit strange. Maybe relying only on named args is possible?

Currently, we have two ways of constructing our constraints: By passing all properties via named arguments or by passing one big associative array. The former is needed so we have a nice completion when using the constraint as attribute while the latter is still supported for historical reasons: Named arguments did not exist in PHP 5 and the array parameter was the ways Doctrine Annotations used to work back in the days.

```php
new Assert\Choice(
    choices: ['fiction', 'non-fiction'],
    message: 'Choose a valid genre.',
);
```

```php
new Assert\Choice([
    'choices' => ['fiction', 'non-fiction'],
    'message' => 'Choose a valid genre.',
]);
```

The example has been taken from https://symfony.com/doc/current/validation.html#constraint-configuration.

The code we maintain in each constraint constructor to allow both ways is pretty complex and it would make the maintenance of those classes a lot easier if we switched to named arguments completely. However, the migration path for downstream projects might be a bit bumpy.

maxbeckers:
I like the idea to reduce the complexity in the code 👍 
But of course the migration could be a big issue for some applications (and can not be done by just search & replace in the IDE). 🤔

stof:
IIRC, the Yaml and XML loaders still rely on passing all options as array in the constructor. So we would need to have an opt-in for the new behavior for each constraints.

Note that this is similar to what doctrine/annotations did in the past with their `@NamedArgumentConstructor` annotation allowing to opt-in an annotation into using named arguments in the constructor instead of the old way.

norkunas:
> IIRC, the Yaml and XML loaders still rely on passing all options as array in the constructor. So we would need to have an opt-in for the new behavior for each constraints.

There is already a way https://github.com/symfony/symfony/pull/45072 :)

stof:
hmm, apparently, I missed the introduction of this `HasNamedArguments` attribute allowing to opt-in into using constructor named arguments for the Yaml and XML loaders.
So what we would need is deprecating defining constraints that don't use it and get used in YAML or XML

derrabus:
[...]

PR #52012, opened 2023-10-12
[Validator] Add PHPDoc to validator constraints

squrious:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Part of #51920 
| License       | MIT

Add PHPDoc to Validator constraints attributes.

### DIFF
--- a/src/Symfony/Component/Validator/Attribute/HasNamedArguments.php
+++ b/src/Symfony/Component/Validator/Attribute/HasNamedArguments.php
@@ -11,6 +11,11 @@
 
 namespace Symfony\Component\Validator\Attribute;
 
+/**
+ * Hints the loader that some constraint options are required.
+ *
+ * @see https://symfony.com/doc/current/validation/custom_constraint.html
+ */
 #[\Attribute(\Attribute::TARGET_METHOD)]
 final class HasNamedArguments
 {
--- a/src/Symfony/Component/Validator/Constraints/All.php
+++ b/src/Symfony/Component/Validator/Constraints/All.php
@@ -14,6 +14,9 @@ namespace Symfony\Component\Validator\Constraints;
 use Symfony\Component\Validator\Constraint;
 
 /**
+ * When applied to an array (or Traversable object), this constraint allows you to apply
+ * a collection of constraints to each element of the array.
+ *
  * @author Bernhard Schussek <bschussek@gmail.com>
  */
 #[\Attribute(\Attribute::TARGET_PROPERTY | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
@@ -21,6 +24,10 @@ class All extends Composite
 {
     public array|Constraint $constraints = [];
 
+    /**
+     * @param array<Constraint>|array<string,mixed>|null $constraints
+     * @param string[]|null                              $groups
+     */
     public function __construct(mixed $constraints = null, array $groups = null, mixed $payload = null)
     {
         parent::__construct($constraints ?? [], $groups, $payload);
--- a/src/Symfony/Component/Validator/Constraints/Bic.php
+++ b/src/Symfony/Component/Validator/Constraints/Bic.php
@@ -18,6 +18,10 @@ use Symfony\Component\Validator\Exception\ConstraintDefinitionException;
 use Symfony\Component\Validator\Exception\LogicException;
 
 /**
+ * Ensures that the value is valid against the BIC format.
+ *
+ * @see https://en.wikipedia.org/wiki/ISO_9362
+ *
  * @author Michael Hirschler <michael.vhirsch@gmail.com>
  */
 #[\Attribute(\Attribute::TARGET_PROPERTY | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
@@ -43,6 +47,12 @@ class Bic extends Constraint
     public ?string $iban = null;
     public ?string $ibanPropertyPath = null;
 
+    /**
+     * @param array<string,mixed>|null $options
+     * @param string|null              $iban             An IBAN value to validate that its country code is the same as the BIC's one
+     * @param string|null              $ibanPropertyPath Property path to the IBAN value when validating objects
+     * @param string[]|null            $groups
+     */
     public function __construct(array $options = null, string $message = null, string $iban = null, string $ibanPropertyPath = null, string $ibanMessage = null, array $groups = null, mixed $payload = null)
     {
         if (!class_exists(Countries::class)) {
--- a/src/Symfony/Component/Validator/Constraints/Blank.php
+++ b/src/Symfony/Component/Validator/Constraints/Blank.php
@@ -14,6 +14,8 @@ namespace Symfony\Component\Validator\Constraints;
 use Symfony\Component\Validator\Constraint;
 
 /**
+ * Validates that a value is blank, i.e. an empty string or null.
+ *
  * @author Bernhard Schussek <bschussek@gmail.com>
  */
 #[\Attribute(\Attribute::TARGET_PROPERTY | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
@@ -27,6 +29,10 @@ class Blank extends Constraint
 
     public string $message = 'This value should be blank.';
 
+    /**
+     * @param array<string,mixed>|null $options
+     * @param string[]|null            $groups
+     */
     public function __construct(array $options = null, string $message = null, array $groups = null, mixed $payload = null)
     {
         parent::__construct($options ?? [], $groups, $payload);
--- a/src/Symfony/Component/Validator/Constraints/Callback.php
+++ b/src/Symfony/Component/Validator/Constraints/Callback.php
@@ -14,6 +14,8 @@ namespace Symfony\Component\Validator\Constraints;
 use Symfony\Component\Validator\Constraint;
 
 /**
+ * Defines custom validation rules through arbitrary callback methods.
+ *
  * @author Bernhard Schussek <bschussek@gmail.com>
  */
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_PROPERTY | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
@@ -24,6 +26,10 @@ class Callback extends Constraint
      */
     public $callback;
 
+    /**
+     * @param string|string[]|callable|array<string,mixed>|null $callback The callback definition
+     * @param string[]|null                                     $groups
+     */
     public function __construct(array|string|callable $callback = null, array $groups = null, mixed $payload = null, array $options = [])
     {
         // Invocation through attributes with an array parameter only
--- a/src/Symfony/Component/Validator/Constraints/Cascade.php
+++ b/src/Symfony/Component/Validator/Constraints/Cascade.php
@@ -15,6 +15,8 @@ use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\Exception\ConstraintDefinitionException;
 
 /**
+ * Validates a whole class, including nested objects in properties.
+ *
  * @author Jules Pietri <jules@heahprod.com>
  */
 #[\Attribute(\Attribute::TARGET_CLASS)]
@@ -22,6 +24,10 @@ class Cascade extends Constraint
 {
     public array $exclude = [];
 
+    /**
+     * @param string[]|string|array<string,mixed>|null $exclude Properties excluded from validation
+     * @param array<string,mixed>|null                 $options
+     */
     public function __construct(array|string $exclude = null, array $options = null)
     {
         if (\is_array($exclude) && !array_is_list($exclude)) {
--- a/src/Symfony/Component/Validator/Constraints/Cidr.php
+++ b/src/Symfony/Component/Validator/Constraints/Cidr.php
@@ -17,6 +17,8 @@ use Symfony\Component\Validator\Exception\ConstraintDefinitionException;
 /**
  * Validates that a value is a valid CIDR notation.
  *
+ * @see https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing
+ *
  * @author Sorin Pop <popsorin15@gmail.com>
  * @author Calin Bolea <calin.bolea@gmail.com>
  */
@@ -43,6 +45,13 @@ class Cidr extends Constraint
     public int $netmaskMin = 0;
     public int $netmaskMax;
 
+    /**
+     * @param array<string,mixed>|null $options
+     * @param string|null              $version    The CIDR version to validate (4, 6 or all, defaults to all)
+     * @param int|null                 $netmaskMin The lowest valid for a valid netmask (defaults to 0)
+     * @param int|null                 $netmaskMax The biggest valid for a valid netmask (defaults to 32 for IPv4, 128 for IPv6)
+     * @param string[]|null            $groups
+     */
     public function __construct(
         array $options = null,
         string $version = null,
--- a/src/Symfony/Component/Validator/Constraints/CssColor.php
+++ b/src/Symfony/Component/Validator/Constraints/CssColor.php
@@ -15,6 +15,8 @@ use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\Exception\InvalidArgumentException;
 
 /**
+ * Validates that a value is a valid CSS color.
+ *
  * @author Mathieu Santostefano <msantostefano@protonmail.com>
  */
 #[\Attribute(\Attribute::TARGET_PROPERTY | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
@@ -60,7 +62,9 @@ class CssColor extends Constraint
     public array|string $formats;
 
     /**
-     * @param array|string $formats The types of CSS colors allowed (e.g. hexadecimal only, RGB and HSL only, etc.).
+     * @param string[]|string|array<string,mixed> $formats The types of CSS colors allowed ({@see https://symfony.com/doc/current/reference/constraints/CssColor.html#formats})
+     * @param string[]|null                       $groups
+     * @param array<string,mixed>|null            $options
      */
     public function __construct(array|string $formats = [], string $message = null, array $groups = null, $payload = null, array $options = null)
     {
--- a/src/Symfony/Component/Validator/Constraints/Currency.php
+++ b/src/Symfony/Component/Validator/Constraints/Currency.php
@@ -16,6 +16,10 @@ use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\Exception\LogicException;
 
 /**
+ * Validates that a value is a valid 3-letter ISO 4217 currency name.
+ *
+ * @see https://en.wikipedia.org/wiki/ISO_4217
+ *
  * @author Miha Vrhovnik <miha.vrhovnik@pagein.si>
  * @author Bernhard Schussek <bschussek@gmail.com>
  */
@@ -30,6 +34,10 @@ class Currency extends Constraint
 
     public string $message = 'This value is not a valid currency.';
 
+    /**
+     * @param array<string,mixed>|null $options
+     * @param string[]|null            $groups
+     */
     public function __construct(array $options = null, string $message = null, array $groups = null, mixed $payload = null)
     {
         if (!class_exists(Currencies::class)) {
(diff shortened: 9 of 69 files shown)


squrious:
I move this PR to WIP as I'm currently updating all type hints that could be more specific

src/Symfony/Component/Validator/Attribute/HasNamedArguments.php
```diff
@@ -11,6 +11,9 @@
 
 namespace Symfony\Component\Validator\Attribute;
 
+/**
+ * Hint the loader that some constraint options are required.
```
GromNaN: We can add a link to https://symfony.com/doc/current/validation/custom_constraint.html

```suggestion
 * Hint the loader that some constraint options are required.
 *
 * @see https://symfony.com/doc/current/validation/custom_constraint.html
```

src/Symfony/Component/Validator/Constraints/Cascade.php
```diff
@@ -25,10 +28,14 @@ class Cascade extends Constraint
...
 
+    /**
+     * @param string[]|string|array<string,mixed>|null $exclude Properties excluded from validation
+     * @param array<string,mixed>|null                 $options The options as an associative array
+     */
     public function __construct(array|string $exclude = null, array $options = null)
     {
         if (\is_array($exclude) && !array_is_list($exclude)) {
-            $options = array_merge($exclude, $options);
+            $options = array_merge($exclude, $options ?? []);
```
GromNaN: Good catch. I'm surprised that this did not trigger any warning in the [tests](https://github.com/symfony/symfony/blob/dc23c8ed3d595e62d79947cbcb7c4238169167d0/src/Symfony/Component/Validator/Tests/Constraints/CascadeTest.php).

src/Symfony/Component/Validator/Constraints/Currency.php
```diff
@@ -16,7 +16,12 @@
 use Symfony\Component\Validator\Exception\LogicException;
 
 /**
+ * Validates that a value s a valid 3-letter ISO 4217 currency name.
```
GromNaN:
```suggestion
 * Validates that a value is a valid 3-letter ISO 4217 currency name.
```

src/Symfony/Component/Validator/Constraints/IsNull.php
```diff
@@ -35,6 +37,12 @@ class IsNull extends Constraint
 
     public $message = 'This value should be null.';
 
+    /**
+     * @param array<string,mixed>|null $options The options as an associative array
```
GromNaN: Why aren't the arguments aligned here when they were before?

GromNaN (commented):
Still reviewing.

src/Symfony/Component/Validator/Constraints/IsNull.php
squrious: Hmm you mean when each part of the param doc starts at the same column?

squrious: Got it, looks like I messed up with this in some classes. I'll fix them along with your other suggestions!

src/Symfony/Component/Validator/Attribute/HasNamedArguments.php
squrious: Link added!

src/Symfony/Component/Validator/Constraints/Cascade.php
OskarStark: I created a bugfix against `6.3`

* https://github.com/symfony/symfony/pull/52194

squrious: Should I remove the fix from this PR?

squrious:
@GromNaN I fixed typos and params alignment :)

src/Symfony/Component/Validator/Constraints/Cascade.php
OskarStark: Yes you can rebase your PR

squrious: Rebase done!

src/Symfony/Component/Validator/Constraints/PasswordStrength.php
```diff
@@ -40,6 +41,13 @@ final class PasswordStrength extends Constraint
 
     public int $minScore;
 
+    /**
+     * @param array<string,mixed>|null $options  The options as an associative array
+     * @param self::STRENGTH_*|null    $minScore The minimum required strength of the password
```
GromNaN:
```suggestion
     * @param self::STRENGTH_*|null    $minScore The minimum required strength of the password (STRENGTH_MEDIUM by default)
```

src/Symfony/Component/Validator/Constraints/Timezone.php
```diff
@@ -46,6 +48,15 @@ class Timezone extends Constraint
      */
     protected static $errorNames = self::ERROR_NAMES;
 
+    /**
+     * @param int|array<string,mixed>|null $zone           Restrict valid timezones to this geographical zone(s) (see {@see \DateTimeZone}), or this constraint's options as an associative array
+     * @param string|null                  $message        The message if the validation fails
+     * @param string|null                  $countryCode    Restrict the valid timezones to this country if the <pre>zone</pre> option is <pre>\DateTimeZone::PER_COUNTRY</pre>
```
GromNaN: I don't think we want to put HTML in docblocks.

```suggestion
     * @param string|null                  $countryCode    Restrict the valid timezones to this country if the zone option is {@see \DateTimeZone::PER_COUNTRY}
```

src/Symfony/Component/Validator/Constraints/File.php
```diff
@@ -15,7 +15,14 @@
 use Symfony\Component\Validator\Exception\ConstraintDefinitionException;
 
 /**
+ * Validates that a value is a valid "file".
+ *
+ * A file can be one of the following:
+ *   - A string (or object with a __toString() method) path to an existing file;
+ *   - A valid {@link \Symfony\Component\HttpFoundation\File\File} object (including objects of {@link \Symfony\Component\HttpFoundation\File\UploadedFile} class).
```
GromNaN: You used inline `{@see ...}` and `{@link ...}` in the PR. Both are used in Symfony codebase. I don't know what's the appropriate syntax.

GromNaN (commented):
Thank you for your hard work. I've made a few comments, nothing certain, I let others give their opinion.

src/Symfony/Component/Validator/Constraints/Timezone.php
squrious: Ok, and what about code samples? For example in [File](https://github.com/squrious/symfony/blob/05047c5c1a0d7104a334fe6645ace3c363c64af8/src/Symfony/Component/Validator/Constraints/File.php#L104) I used a `pre` tag to provide an array example. Without it it's quite unreadable. Is it better to not provide the example at all?

src/Symfony/Component/Validator/Constraints/File.php
squrious: I thought `{@link ...}` was used to reference elements, and `{@see ...}` to reference links. But the official [phpDocumentor doc](https://docs.phpdoc.org/2.9/references/phpdoc/tags/link.html) says the opposite. I'll replace with `{@see ...}` to be consistent.

src/Symfony/Component/Validator/Constraints/PasswordStrength.php
squrious: Shoud I document all the defaults in all constraints too?

src/Symfony/Component/Validator/Constraints/Timezone.php
stof: I would not put code snippets in the phpdoc. If the explanation involves code snippets, it deserves being in the symfony documentation IMO (which then allows to have the code snippet for all the mapping formats)

src/Symfony/Component/Validator/Constraints/File.php
stof: According to the phpdocumentor 3 docs, `@see` supports both while `@link` supports only URIs.

src/Symfony/Component/Validator/Constraints/PasswordStrength.php
GromNaN: I would say, only when the value is not already the default value of the arg (mainly not when it's not `null`).

squrious: Ok I'll do another pass!

src/Symfony/Component/Validator/Constraints/CssColor.php
```diff
@@ -71,7 +71,7 @@ class CssColor extends Constraint
     public $formats;
 
     /**
-     * @param string[]|string|array<string,mixed> $formats The types of CSS colors allowed (e.g. hexadecimal only, RGB, HSL only, etc.), or this constraint's options as an associative array
+     * @param string[]|string|array<string,mixed> $formats The types of CSS colors allowed ({@see https://symfony.com/doc/6.4/reference/constraints/CssColor.html#formats}), or this constraint's options as an associative array
```
stof: please use `/current` rather than `/6.4` so that we don't have to update the link in each branch of the code repository (or keep linking to a version after it goes EOL).

src/Symfony/Component/Validator/Constraints/Email.php
```diff
@@ -62,7 +62,7 @@ class Email extends Constraint
     /**
      * @param array<string,mixed>|null     $options    The options as an associative array
      * @param string|null                  $message    The message if the validation fails
-     * @param self::VALIDATION_MODE_*|null $mode       The pattern used to validate the email address (e.g. strict, loose, html5, etc.)
+     * @param self::VALIDATION_MODE_*|null $mode       The pattern used to validate the email address (e.g. strict, loose, html5, defaults to framework.validation.email_validation_mode configuration value)
```
stof: this should rather say that `null` will use the default configured for the EmailValidator. Referring to the FrameworkBundle configuration setting in the component phpdoc looks wrong to me.

squrious:
I applied your recommendations in a new commit so we can see the changes:
- Document defaults behaviors when it was possible
- Replace `@link` with `@see`
- Remove all `<pre>` tags
- Fix some mistakes
- Use more `{@see}` to reference structural elements like class constants

GromNaN (approved):

src/Symfony/Component/Validator/Constraints/Timezone.php
squrious: I put a link to the documentation

src/Symfony/Component/Validator/Constraints/Email.php
squrious: Ok! I should also update the `NoSuspiciousCharacters` that references the `framework.enabled_locales` config

src/Symfony/Component/Validator/Constraints/CssColor.php
squrious: Fixed!

src/Symfony/Component/Validator/Constraints/Email.php
squrious: Fixed in both files

OskarStark:
Can you please rebase?

src/Symfony/Component/Validator/Constraints/All.php
```diff
@@ -11,8 +11,14 @@
...
 namespace Symfony\Component\Validator\Constraints;
 
+use Symfony\Component\Validator\Constraint;
+
 /**
+ * When applied to an array (or Traversable object), this constraint allows you to apply
+ * a collection of constraints to each element of the array.
+ *
  * @Annotation
+ *
```
nicolas-grekas: I'm sorry but please revert all those added blank lines, they're false positives.

src/Symfony/Component/Validator/Attribute/HasNamedArguments.php
```diff
@@ -11,6 +11,11 @@
...
```
nicolas-grekas:
```suggestion
 * Hints the loader that some constraint options are required.
```

src/Symfony/Component/Validator/Constraints/Bic.php
```diff
@@ -18,7 +18,12 @@
 use Symfony\Component\Validator\Exception\LogicException;
 
 /**
+ * Ensure that the value is valid against the BIC format.
```
nicolas-grekas:
```suggestion
 * Ensures that the value is valid against the BIC format.
```

src/Symfony/Component/Validator/Constraints/Blank.php
```diff
@@ -14,7 +14,10 @@
 use Symfony\Component\Validator\Constraint;
 
 /**
+ * Validate that a value is blank, i.e. an empty string or null.
```
nicolas-grekas:
```suggestion
 * Validates that a value is blank, i.e. an empty string or null.
```

src/Symfony/Component/Validator/Constraints/Callback.php
```diff
@@ -14,7 +14,10 @@
 use Symfony\Component\Validator\Constraint;
 
 /**
+ * Define custom validation rules through arbitrary callback methods.
```
nicolas-grekas:
```suggestion
 * Defines custom validation rules through arbitrary callback methods.
```

nicolas-grekas (commented):
Thanks for working on this. I have mixed feelings here and I'm already sorry about that.
That's a lot of added lines to review and maintain. Are we sure all this is not just boilerplate? Eg is the repetition  of the description for groups, payload, options, error message, etc. that useful? I'm not the best person to answer this question because I'm too much into this already but I'm still wondering.
Any other opinion?

src/Symfony/Component/Validator/Constraints/All.php
squrious: I removed annotations when rebasing on 7.1

squrious:
Rebased on 7.1!

@nicolas-grekas I also applied your suggestions :)

> Thanks for working on this. I have mixed feelings here and I'm already sorry about that. That's a lot of added lines to review and maintain. Are we sure all this is not just boilerplate? Eg is the repetition of the description for groups, payload, options, error message, etc. that useful? I'm not the best person to answer this question because I'm too much into this already but I'm still wondering. Any other opinion?

I agree the repetitions are very redundant and add a lot to maintain. Maybe it's not so useful for groups, payload and error messages as their behavior is likely to be known when working with the validator. But for the options it's a bit different, as the first constructor argument may be used as the default option depending on the constraint. Documenting this aspect may worth it I guess.

nicolas-grekas:
In #51393, we're wondering about a way to deprecate passing options as an array and require using named arguments instead, so maybe we shouldn't document the array-style?

I'm all for removing redundant descriptions (groups, payload, etc), can you please update accordingly?

squrious:
> In #51393, we're wondering about a way to deprecate passing options as an array and require using named arguments instead, so maybe we shouldn't document the array-style?
 
If it's about to be deprecated, yup let's get rid of those too.

> I'm all for removing redundant descriptions (groups, payload, etc), can you please update accordingly?

Yes! Just a little question: should I remove the messages documentation if the behavior is a little bit specific to the constraint? Eg in `Choice`, we have `$multipleMessage`, `$minMessage`, etc. Should they be removed too?

nicolas-grekas:
We should remove low-value descriptions (things that should be obvious once you get the basics of how things work)

src/Symfony/Component/Validator/Constraints/Cidr.php
```diff
@@ -43,6 +45,14 @@ class Cidr extends Constraint
     public int $netmaskMin = 0;
     public int $netmaskMax;
 
+    /**
+     * @param array<string,mixed>|null $options
+     * @param string|null              $version    The CIDR version to validate (4, 6 or all, defaults to all)
+     * @param int|null                 $netmaskMin The lowest valid for a valid netmask (defaults to 0)
+     * @param int|null                 $netmaskMax The biggest valid for a valid netmask (defaults to 32 for IPv4, 128 for IPv6)
+     * @param string[]|null            $groups
+     * @param mixed|null               $payload
```
nicolas-grekas: we're fine with partial docblocks:
```suggestion

File: src/Symfony/Component/Validator/Constraints/CssColor.php [hunk 16]
```diff
@@ -60,7 +62,10 @@ class CssColor extends Constraint
     public array|string $formats;
 
     /**
-     * @param array|string $formats The types of CSS colors allowed (e.g. hexadecimal only, RGB and HSL only, etc.).
+     * @param string[]|string|array<string,mixed> $formats The types of CSS colors allowed ({@see https://symfony.com/doc/current/reference/constraints/CssColor.html#formats})
+     * @param string[]|null                       $groups
+     * @param mixed|null                          $payload
```
nicolas-grekas: same or all of them:
```suggestion

File: src/Symfony/Component/Validator/Constraints/Email.php [hunk 17]
```diff
@@ -43,6 +45,11 @@ class Email extends Constraint
     /** @var callable|null */
     public $normalizer;
 
+    /**
+     * @param array<string,mixed>|null     $options
+     * @param self::VALIDATION_MODE_*|null $mode    The pattern used to validate the email address (e.g. strict, loose, html5). Pass null to use the default mode configured for the EmailValidator. (defaults to null)
```
nicolas-grekas:
```suggestion
     * @param self::VALIDATION_MODE_*|null $mode    The pattern used to validate the email address; pass null to use the default mode configured for the EmailValidator
```

src/Symfony/Component/Validator/Constraints/Image.php
```diff
@@ -85,6 +87,56 @@ class Image extends File
...
+     * @param string|null              $maxHeightMessage
+     * @param string|null              $minHeightMessage
+     * @param string|null              $minPixelsMessage
+     * @param string|null              $maxPixelsMessage
+     * @param string|null              $maxRatioMessage
+     * @param string|null              $minRatioMessage
+     * @param string|null              $allowSquareMessage
+     * @param string|null              $allowLandscapeMessage
+     * @param string|null              $allowPortraitMessage
+     * @param string|null              $corruptedMessage
```
nicolas-grekas:
```suggestion

```

src/Symfony/Component/Validator/Constraints/Image.php
```diff
@@ -85,6 +87,56 @@ class Image extends File
...
+     * @param string|null              $uploadIniSizeErrorMessage   Message if the file size exceeds the max size configured in php.ini
+     * @param string|null              $uploadFormSizeErrorMessage  Message if the file size exceeds the max size configured in the HTML input field
+     * @param string|null              $uploadPartialErrorMessage   Message if the file is only partially uploaded
+     * @param string|null              $uploadNoFileErrorMessage
+     * @param string|null              $uploadNoTmpDirErrorMessage  Message if there is no upload_tmp_dir in php.ini
+     * @param string|null              $uploadCantWriteErrorMessage Message if the uploaded file can not be stored in the temporary directory
+     * @param string|null              $uploadExtensionErrorMessage
+     * @param string|null              $uploadErrorMessage          Message if an unknown error occurred on upload
+     * @param string[]|null            $groups
+     * @param mixed|null               $payload
```
nicolas-grekas:
```suggestion
```

src/Symfony/Component/Validator/Constraints/Image.php
```diff
@@ -85,6 +87,56 @@ class Image extends File
...
+    /**
+     * @param array<string,mixed>|null $options
+     * @param int|string|null          $maxSize                     The max size of the underlying file
+     * @param bool|null                $binaryFormat                Pass true to use binary-prefixed units (KiB, MiB, etc.) or false to use SI-prefixed units (kB, MB) in displayed messages. Pass null to guess the format from the maxSize option. (defaults to null)
+     * @param string[]|null            $mimeTypes                   Acceptable media types
+     * @param int|null                 $filenameMaxLength           Maximum length of the file name
+     * @param string|null              $notFoundMessage
+     * @param string|null              $notReadableMessage
+     * @param string|null              $maxSizeMessage
+     * @param string|null              $mimeTypesMessage
```
nicolas-grekas:
```suggestion

```

nicolas-grekas (commented):
much better :)
let's also remove docblocks that just duplicate the signature - we're fine with partial `@param` entries

nicolas-grekas:
Thank you @squrious.

PR #52015, opened 2023-10-12
DX: PHP CS Fixer - drop explicit no_superfluous_phpdoc_tags config, as it's part of ruleset already

keradus:
https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/src/RuleSet/Sets/SymfonySet.php#L111


derrabus (approved):

nicolas-grekas:
Thank you @keradus.

PR #52016, opened 2023-10-12
DX: PHP CS Fixer - drop explicit nullable_type_declaration_for_default_null_value config, as it's part of ruleset anyway

keradus:
https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/src/RuleSet/Sets/SymfonySet.php#L137


derrabus (approved):

nicolas-grekas:
Thank you @keradus.

PR #52017, opened 2023-10-12
[Mailer] Capitalize sender header for Mailgun

Romanavr:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix https://github.com/symfony/symfony/pull/51773#issuecomment-1759392593
| License       | MIT

This fix resolves the duplicate header issue.
Without a fix, using API & HTTP transport type, emails could leave with double headers, causing problems for some email providers, for example, the problem was reproduced with Gmail:
`5.7.1 This message is not RFC 5322 compliant. There are multiple Sender 5.7.1 headers. To reduce the amount of spam sent to Gmail, this message has 5.7.1 been blocked.`

Moreover, I found that changes to HTTP transport type were not necessary since this problem could be solved by simply adding a sender, so I reverted everything for HTTP transport type.


svendecabooter:
This fixes the issue I reported for me.

nicolas-grekas:
Thank you @Romanavr.

PR #52018, opened 2023-10-12
[Serializer] Add support for union collection value types in `ArrayDenormalizer`

Jeroeny:
| Q             | A
| ------------- | ---
| Branch?       | 7.2
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

Currently only the following array types combination is supported: `array<MyObject>|array<Another>`. This results in arrays with only 1 type of objects.

This PR aims to add support for arrays where values can be either of those, like union types: `array<MyObject|Another>`.

I've currently tested is successfully against the following types:
- `DateTimeInterface[]`
- `array<array<DateTimeInterface|DateInterval|string>>`
- `array<DateTimeInterface>|array<DateInterval>`
- `array<DateTimeInterface|DateInterval>`
- `array<DateTimeInterface|null>`

See also https://github.com/Jeroeny/reproduce/blob/mixedarray/src/Test.php

### DIFF
--- a/src/Symfony/Component/Serializer/CHANGELOG.md
+++ b/src/Symfony/Component/Serializer/CHANGELOG.md
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+7.2
+---
+* Add support for union collection value types in `ArrayDenormalizer`
+
 7.1
 ---
 
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
@@ -444,11 +444,9 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
                 return null;
             }
 
-            $collectionValueType = $type->isCollection() ? $type->getCollectionValueTypes()[0] ?? null : null;
-
             // Fix a collection that contains the only one element
             // This is special to xml format only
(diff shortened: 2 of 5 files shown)


nicolas-grekas:
Anyone with more background on Serializer would be willing to review this PR? (note that it should target 7.1 for now, maybe 7.2 if we can't make it since we're in feature freeze now).
/cc @mtarld maybe? Anyone else?

mtarld:
As https://github.com/symfony/symfony/pull/53160 has been merged, the `validateAndDenormalize` method is now using `Type` of TypeInfo component, which can make this patch easier.
So, this PR has to be rebased and updated accordingly.

Jeroeny:
Rebased onto 7.1 and description targeting 7.2.

The new `validateAndDenormalizeLegacy` vs validateAndDenormalize` are a bit complex, so not sure yet how to apply the changes to the method using the new types.

nicolas-grekas:
Thanks for this and for the earlier rebase. mtarld asked in April 2024 for the implementation to be redone on top of the TypeInfo-based `validateAndDenormalize()` from #53160, you asked how to handle the legacy and new paths, and that question never got an answer, which is on us. The legacy path has since been removed entirely, so the change no longer applies. Closing for now; if you want to take another pass at union collection value types against the current dev branch, open a fresh PR and I will make sure it gets a review.

PR #52021, opened 2023-10-12
[Form] Fix merging params & files when "multiple" is enabled

priyadi:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | 
| License       | MIT

If a form field is configured with the "multiple" option, the request handler merges param values with uploaded files. However, it does not merge correctly if "multiple" is enabled.

Example:
```html
<input type="file" name="param[]" />
<input type="hidden" name="param[]" value="data1" />
<input type="hidden" name="param[]" value="data2" />
```

With the example, the file field will incorrectly replace the first hidden field, when they should be merged. This commit fixes the problem.


nicolas-grekas (approved):
array_merge_recursive would be a fit *if* it didn't cast objects into arrays.
FYI I moved the util method to FormUtil + tweaked it a bit.

nicolas-grekas:
Thank you @priyadi.

PR #51997, opened 2023-10-11
[Doctrine][DoctrineBridge] Compatibility with ORM 3 and DBAL 4

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | Follows #51947
| License       | MIT

This PR makes the 6.3 branch compatible with Doctrine DBAL 4 and ORM 3. It contains all changes from #51947 which does the same on the 5.4 branch.

Please note that I intentionally did not allow DBAL 4 in the dependencies of the monorepo's composer.json. If I did that, we would lose coverage for DBAL 3 completely which would be very unfortunate.

Right now, we should get the following test matrix:

* same deps: DBAL 3
* low deps: DBAL 2
* high deps DBAL 4

PR #52022, opened 2023-10-12
[Doctrine][DoctrineBridge][HttpFoundation][VarDumper] Run tests with ORM 3 and DBAL 4

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Follow-up to #51997
| License       | MIT

Everything that got lost in the merge-up + a fix for the VarDumper testsuite. 😇

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -132,7 +132,7 @@
         "doctrine/collections": "^1.0|^2.0",
         "doctrine/data-fixtures": "^1.1",
         "doctrine/dbal": "^2.13.1|^3.0",
-        "doctrine/orm": "^2.12",
+        "doctrine/orm": "^2.12|^3",
         "dragonmantank/cron-expression": "^3.1",
         "egulias/email-validator": "^2.1.10|^3.1|^4",
         "guzzlehttp/promises": "^1.4",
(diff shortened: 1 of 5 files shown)


derrabus:
I've opened doctrine/orm#11001 to address the deprecations reported by the CI.

nicolas-grekas:
Thank you @derrabus.

composer.json
```diff
@@ -132,7 +132,7 @@
         "doctrine/collections": "^1.0|^2.0",
         "doctrine/data-fixtures": "^1.1",
         "doctrine/dbal": "^2.13.1|^3.0",
```
nicolas-grekas, line 134: should we add dbal v4?

nicolas-grekas:
merged up to 7.0
please let me know if we should add dbal v4 to the root composer.json

derrabus:
Not on 6.x because we would lose coverage for DBAL 3. From #51997:

> Right now, we should get the following test matrix:
>
> * same deps: DBAL 3
> * low deps: DBAL 2
> * high deps DBAL 4

PR #52024, opened 2023-10-12
[AssetMapper] Add a "package specifier" to importmap in case import name != package+path

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | yes
| Deprecations? | no
| Tickets       | None
| License       | MIT

Hi!

Sorry for the large PR - this tightens up the component after the many recent enhancements.

`tl;dr` Makes "vendor" importmap entries work identically to "local" entries when it comes to preloading (a bug). Also tightens up `ImportMapEntry`, which is a tricky class because if an entry is "remote", it has 2 extra, required, properties.

Full changes:

* Make ImportMapEntry stricter so that, when remote, version & packageModuleSpecifier are always set.
* Set the "path" property always for ImportMapEntry, making them no different than local files through most of the system.
* Related, fix a bug where if a vendor .js file imported another module, that other module was not previously preloaded when the vendor module was preloaded.
* Fix a bug with downloaded package paths if you include a specific package - chart.js - and also a path - chart.js/auto. Previously, this would try to create both a file and directory called chart.js,
* Fix bug where imports weren't matching if some spaces were missing

Cheers!


src/Symfony/Component/AssetMapper/Command/ImportMapRequireCommand.php
```diff
@@ -87,15 +87,6 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
             $path = $input->getOption('path');
-            if (!is_file($path)) {
-                $path = $this->projectDir.'/'.$path;
-
-                if (!is_file($path)) {
-                    $io->error(sprintf('The path "%s" does not exist.', $input->getOption('path')));
-
-                    return Command::FAILURE;
-                }
-            }
```
weaverryan, line 98: Check is done in `ImportMapManager` with a nice error

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
```diff
@@ -28,8 +28,8 @@ final class JavaScriptImportPathCompiler implements AssetCompilerInterface
 {
     use AssetCompilerPathResolverTrait;
 
-    // https://regex101.com/r/5Q38tj/1
-    private const IMPORT_PATTERN = '/(?:import\s+(?:(?:\*\s+as\s+\w+|[\w\s{},*]+)\s+from\s+)?|\bimport\()\s*[\'"`](\.\/[^\'"`]+|(\.\.\/)*[^\'"`]+)[\'"`]\s*[;\)]?/m';
+    // https://regex101.com/r/fquriB/1
+    private const IMPORT_PATTERN = '/(?:import\s*(?:(?:\*\s*as\s+\w+|[\w\s{},*]+)\s*from\s*)?|\bimport\()\s*[\'"`](\.\/[^\'"`]+|(\.\.\/)*[^\'"`]+)[\'"`]\s*[;\)]?/m';
```
weaverryan, line 32: Covered with a test

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
```diff
@@ -145,12 +145,11 @@ private function findAssetForBareImport(string $importedModule, AssetMapperInter
...
 
-        // remote entries have no MappedAsset
-        if ($importMapEntry->isRemotePackage()) {
-            return null;
+        if ($asset = $assetMapper->getAsset($importMapEntry->path)) {
+            return $asset;
         }
 
-        return $assetMapper->getAsset($importMapEntry->path);
+        return $assetMapper->getAssetFromSourcePath($importMapEntry->path);
```
weaverryan, line 152: e.g. `bootstrap` imports `@popperjs/core`. If `bootstrap` is ultimately preloaded, this is what allows us to also preload `@popperjs/core`.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapAuditor.php
```diff
@@ -51,14 +47,19 @@ public function audit(): array
...
 
-            $installed[$entry->importName] ??= [];
-            $installed[$entry->importName][] = $version;
+            $packageName = $entry->getPackageName();
+            $installed[$packageName] ??= [];
+            $installed[$packageName][] = $version;
 
-            $packageVersion = $entry->importName.($version ? '@'.$version : '');
-            $packageAudits[$packageVersion] ??= new ImportMapPackageAudit($entry->importName, $version);
+            $packageVersion = $packageName.'@'.$version;
```
weaverryan, line 54: Bug fix - e.g. if you have `chart.js/auto`, `$packageName` will now be just `chart.js`, which is what we need for the API call. Covered with a test.

Also, `$version` is now guaranteed not null here.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapConfigReader.php
```diff
@@ -38,7 +40,7 @@ public function getEntries(): ImportMapEntries
 
         $entries = new ImportMapEntries();
         foreach ($importMapConfig ?? [] as $importName => $data) {
-            $validKeys = ['path', 'version', 'type', 'entrypoint', 'url'];
+            $validKeys = ['path', 'version', 'type', 'entrypoint', 'url', 'package_specifier'];
```
weaverryan, line 43: Supports the rare-use case that you require a package (e.g. `bootstrap`) but want to have its import called something else - `shoestrap` - so you can `import 'shoestrap'`. The `importmap:require` command supports this directly & it's documented.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapEntry.php
```diff
@@ -18,22 +18,75 @@
...
+
+    public static function createLocal(string $importName, ImportMapType $importMapType, string $path, bool $isEntrypoint): self
+    {
+        return new self($importName, $importMapType, $path, $isEntrypoint, null, null);
+    }
+
+    public static function createRemote(string $importName, ImportMapType $importMapType, string $path, string $version, string $packageModuleSpecifier, bool $isEntrypoint): self
+    {
+        return new self($importName, $importMapType, $path, $isEntrypoint, $version, $packageModuleSpecifier);
+    }
```
weaverryan, line 48: private constructor + static factories. Done this way so that IF you are remote, we can guarantee that both `$version` and `$packageModuleSpecifier` ARE included.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -154,19 +154,9 @@ public function getRawImportMapData(): array
...
-            } else {
-                $sourcePath = $this->packageDownloader->getDownloadedPath($entry->importName);
-                $asset = $this->assetMapper->getAssetFromSourcePath($sourcePath);
-
-                if (!$asset) {
-                    throw new \InvalidArgumentException(sprintf('The "%s" vendor asset is missing. Run "php bin/console importmap:install".', $entry->importName));
-                }
+            $asset = $this->findAsset($entry->path);
+            if (!$asset) {
+                throw $this->createMissingImportMapAssetException($entry);
```
weaverryan, line 160: Example of why it's nice now that `ImportMapEntry.path` is populated even for vendor/remote assets :)

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -345,14 +321,9 @@ private function addImplicitEntries(ImportMapEntry $entry, array $currentImportE
             return $currentImportEntries;
         }
 
-        // remote packages aren't in the asset mapper & so don't have dependencies
-        if ($entry->isRemotePackage()) {
-            return $currentImportEntries;
-        }
```
weaverryan, line 351: An edge case (that does not currently apply) where vendor assets have relative imports, so we need to add those to the importmap.

More broadly, just another example of not treating the remote packages any differently.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapConfigReader.php
```diff
@@ -49,35 +51,43 @@ public function getEntries(): ImportMapEntries
...
+
+            if (isset($data['path'])) {
+                if (isset($data['version'])) {
+                    throw new RuntimeException(sprintf('The importmap entry "%s" cannot have both a "path" and "version" option.', $importName));
+                }
+                if (isset($data['package_specifier'])) {
+                    throw new RuntimeException(sprintf('The importmap entry "%s" cannot have both a "path" and "package_specifier" option.', $importName));
+                }
+
+                $entries->add(ImportMapEntry::createLocal(
```
nicolas-grekas: oneliners FTW ;)

src/Symfony/Component/AssetMapper/ImportMap/ImportMapConfigReader.php
```diff
@@ -49,35 +51,43 @@ public function getEntries(): ImportMapEntries
...
             }
-            if (null !== $version && null !== $path) {
-                throw new RuntimeException(sprintf('The importmap entry "%s" cannot have both a "path" and "version" option.', $importName));
-            }
-
-            [$packageName, $filePath] = self::splitPackageNameAndFilePath($importName);
 
-            $entries->add(new ImportMapEntry(
+            $packageModuleSpecifier = $data['package_specifier'] ?? $importName;
+            $entries->add($this->createRemoteEntry(
```
nicolas-grekas: oneliner love

src/Symfony/Component/AssetMapper/ImportMap/ImportMapEntry.php
```diff
@@ -18,22 +18,75 @@
...
+    public readonly bool $isEntrypoint;
+
+    // remote-only
+    public readonly string $version;
+    /**
+     * The "package-name/path" of the remote package.
+     */
+    public readonly string $packageModuleSpecifier;
+
+    private function __construct(
```
nicolas-grekas, line 21: one line also

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -457,4 +429,13 @@ private function findAsset(string $path): ?MappedAsset
 
         return $this->assetMapper->getAssetFromSourcePath($path);
     }
+
+    private function createMissingImportMapAssetException(ImportMapEntry $entry): \InvalidArgumentException
```
nicolas-grekas, line 434: should return a LogicException I suppose, that's how we report such errors usually IIRC - let's also add the exception class to the component's namespace with its interface?

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -457,4 +429,13 @@ private function findAsset(string $path): ?MappedAsset
...
+    {
+        if ($entry->isRemotePackage()) {
+            throw new \InvalidArgumentException(sprintf('The "%s" vendor asset is missing. Run "php bin/console importmap:install".', $entry->importName));
```
nicolas-grekas: Trying to decouple from `php bin/console`, since it could very well be `symfony console` instead:
```suggestion
            throw new \LogicException(sprintf('The "%s" vendor asset is missing. Try running the "importmap:install" command.', $entry->importName));
```

src/Symfony/Component/AssetMapper/ImportMap/ImportMapUpdateChecker.php
```diff
@@ -34,27 +34,30 @@ public function getAvailableUpdates(array $packages = []): array
         $updateInfos = [];
         $responses = [];
         foreach ($entries as $entry) {
-            if (null === $entry->packageName || null === $entry->version) {
+            if (!$entry->isRemotePackage()) {
                 continue;
             }
-            if (\count($packages) && !\in_array($entry->packageName, $packages, true)) {
+            if (\count($packages)
```
nicolas-grekas:
```suggestion
            if ($packages
```

src/Symfony/Component/AssetMapper/ImportMap/RemotePackageStorage.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public function isDownloaded(ImportMapEntry $entry): bool
+    {
+        return is_file($this->getDownloadPath($entry->packageModuleSpecifier, $entry->type));
+    }
+
+    public function save(ImportMapEntry $entry, string $contents): void
+    {
+        $vendorPath = $this->getDownloadPath(
```
nicolas-grekas: oneline

src/Symfony/Component/AssetMapper/Tests/Command/DebugAssetsMapperCommandTest.php
```diff
@@ -9,7 +9,7 @@
  * file that was distributed with this source code.
  */
 
-namespace Command;
+namespace Symfony\Component\AssetMapper\Tests\Command;
```
nicolas-grekas, line 12: :see_no_evil:

nicolas-grekas (approved):
Trusting you here :)

src/Symfony/Component/AssetMapper/ImportMap/ImportMapEntry.php
```diff
@@ -18,22 +18,75 @@
...
+    public readonly string $importName;
+    public readonly ImportMapType $type;
+    /**
+     * A logical path, relative path or absolute path to the file.
+     */
+    public readonly string $path;
+    public readonly bool $isEntrypoint;
+
+    // remote-only
+    public readonly string $version;
```
stof: Should this be a public property in an uninitialized state for remote ones ? This looks like a weird API (and no SA tool out there will tell you to handle that case).

I'd rather keep them as nullable properties.

stof: Note that if you want it to be even nicer for SA tools, you could add the following phpdoc on the `isRemotePackage` method (once the property is nullable rather than left uninitialized):

```php
/**
 * @psalm-assert-if-true !null $this->version
 * @psalm-assert-if-true !null $this->packageModuleSpecifier
 */
```

src/Symfony/Component/AssetMapper/ImportMap/RemotePackageStorage.php
```diff
@@ -0,0 +1,67 @@
...
+        private readonly string $vendorDir,
+    ) {
+    }
+
+    public function getStorageDir(): string
+    {
+        return $this->vendorDir;
+    }
+
+    public function isDownloaded(ImportMapEntry $entry): bool
```
stof, line 28: This should add proper error handling for the case of passing a non-remote entry to this method (as this is not an internal API where we can rely on the caller doing the right thing).
right now, we would get an error about `$entry->packageModuleSpecifier` being uninitialized.

src/Symfony/Component/AssetMapper/ImportMap/RemotePackageStorage.php
```diff
@@ -0,0 +1,67 @@
...
+    public function getDownloadPath(string $packageModuleSpecifier, ImportMapType $importMapType): string
+    {
+        [$packageName, $packagePathString] = ImportMapEntry::splitPackageNameAndFilePath($packageModuleSpecifier);
+        $filename = $packageName;
+        if ($packagePathString) {
+            $filename .= '/'.ltrim($packagePathString, '/');
+        } else {
+            // if we're requiring a bare package, we put it into the directory
+            // (in case we also import other files from the package) and arbitrarily
+            // name it the same as the package name + ".main"
```
stof: shouldn't we name it index rather than main ? that's the common name for the file being imported when loading a directory.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapEntry.php
weaverryan: I did that weird api on purpose… but it is weird. Previously, we made these values null for local packages. But that made code around the component more delicate: if we made a mistake and tried to reference version for a local package, instead of an error, other more silent problems would happen (since we would get null when and use it in a context expecting a string). 

The other 2 API ideas I had were: (1) create a RemoteImportmapEntry for remote and then type-hint that or check instanceof around the code and (2) convert ImportMapEntry to getters with private properties so we can throw a clear error when accessing version for a local package.

WDYT?

src/Symfony/Component/AssetMapper/ImportMap/RemotePackageStorage.php
weaverryan: “main” is the key in package.json? But yea, index might be a better choice?

src/Symfony/Component/AssetMapper/ImportMap/ImportMapEntry.php
stof: Having getters with a clear error is much better than having a TypeError due to uninitialized properties.

However, this would still not allow SA tools to help you write correct code (as code triggering those errors would still not be correct).

Personally, I would either go with the solution with `@psalm-assert-if-true` on `isRemotePackage` (allowing Psalm to know that checking for remote packages ensures we get non-null values) or a separate class having those 2 additional properties (which has the benefit of being usable in types when a function supports only remote entries)

fabpot:
Thank you @weaverryan.

PR #52033, opened 2023-10-13
[Mime] Throw InvalidArgumentException on invalid form field type inside array

l-naumann:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | n/a
| License       | MIT

Example code:
```php
$f = new FormDataPart([
            'foo' => [
                'bar' => 'baz',
                'qux' => [
                    'quux' => 1,
                ],
            ],
        ]);
$f->getParts();
```

Currently, when you use a disallowed type inside an array, a TypeError is thrown:
` Symfony\Component\Mime\Part\Multipart\FormDataPart::configurePart(): Argument #2 ($part) must be of type Symfony\Component\Mime\Part\TextPart, int given`

This change adds a clearly understandable error message:
`The value of the form field "foo[qux][quux]" can only be a string, an array, or an instance of TextPart ("int" given).`

### DIFF
--- a/src/Symfony/Component/Mime/Part/Multipart/FormDataPart.php
+++ b/src/Symfony/Component/Mime/Part/Multipart/FormDataPart.php
@@ -32,13 +32,8 @@ final class FormDataPart extends AbstractMultipartPart
     {
         parent::__construct();
 
-        foreach ($fields as $name => $value) {
-            if (!\is_string($value) && !\is_array($value) && !$value instanceof TextPart) {
-                throw new InvalidArgumentException(sprintf('The value of the form field "%s" can only be a string, an array, or an instance of TextPart ("%s" given).', $name, get_debug_type($value)));
-            }
+        $this->fields = $fields;
 
-            $this->fields[$name] = $value;
-        }
         // HTTP does not support \r\n in header values
         $this->getHeaders()->setMaxLineLength(\PHP_INT_MAX);
     }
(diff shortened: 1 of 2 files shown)


fabpot (approved):

src/Symfony/Component/Mime/Part/Multipart/FormDataPart.php
```diff
@@ -75,6 +75,10 @@ private function prepareFields(array $fields): array
                 return;
             }
 
+            if (!\is_string($item) && !$item instanceof TextPart) {
```
fabpot, line 73: Maybe we can move this check to the constructor directly? That would be more consistent as we already check some of that there.

l-naumann: I also considered this, but in my opinion it has two disadvantages:
1. We have to traverse through the array recursively. In `prepareFields` this is already the case, but in the constructor we only traverse at top-level.
2. We don't have the field name in form of "foo[qux][quux]" in the constructor. Either we cannot include it in the error message, or we have to generate it redundantly.

Also we already have a check (regarding the form field values with integer keys) inside `prepareFields`, so I don't feel like adding another check there affects the consistency much.

How about we remove the check in the constructor altogether? All the checks could be done inside `prepareFields` and the constructor would look like this:
```
 public function __construct(array $fields = [])
    {
        parent::__construct();

        $this->fields = $fields;

        // HTTP does not support \r\n in header values
        $this->getHeaders()->setMaxLineLength(\PHP_INT_MAX);
    }
```

fabpot: Having all the checks in prepareFields works for me.

nicolas-grekas:
Thank you @l-naumann.

PR #52039, opened 2023-10-13
[Scheduler] Continue with stored `Checkpoint::$time` on lock

Jeroeny:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

When there are multiple message-consumer processes running, the Lock mechanism can be used to ensure a single process takes responsibility of the scheduling. 
However when the responsible process releases the lock and a new process acquires it, the state of the checkpoint is reset on purpose.  

I'm not sure what the reason behind this is. Maybe I'm missing something and this PR isn't valid. But it seems strange that a Checkpoint stores it's progress which is then lost when that process stops.

Perhaps the 'reset' was more intended like a `reload', since the cached value may have changed by another process? Then this PR would fix it to be an actual reload. 

Also I think the change in logic requires this to go to `6.4` branch?

### DIFF
--- a/src/Symfony/Component/Scheduler/Generator/Checkpoint.php
+++ b/src/Symfony/Component/Scheduler/Generator/Checkpoint.php
@@ -31,20 +31,18 @@ final class Checkpoint implements CheckpointInterface
     public function acquire(\DateTimeImmutable $now): bool
     {
         if ($this->lock && !$this->lock->acquire()) {
-            // Reset local state if a Lock is acquired by another Worker.
+            // Reset local state if a Lock is acquired by another Worker and state is not shared through cache.
             $this->reset = true;
 
             return false;
         }
 
-        if ($this->reset) {
-            $this->reset = false;
-            $this->save($now, -1);
-        }
-
         if ($this->cache) {
             [$this->time, $this->index, $this->from] = $this->cache->get($this->name, fn () => [$now, -1, $now]) + [2 => $now];
             $this->save($this->time, $this->index);
+        } elseif ($this->reset) {
+            $this->reset = false;
+            $this->save($now, -1);
         }
 
         $this->time ??= $now;
--- a/src/Symfony/Component/Scheduler/Tests/Generator/CheckpointTest.php
+++ b/src/Symfony/Component/Scheduler/Tests/Generator/CheckpointTest.php
@@ -190,6 +190,30 @@ class CheckpointTest extends TestCase
         $this->assertFalse($concurrentLock->isAcquired());
     }
 
+    public function testWithLockResetStateAfterLockedAcquiringCache()
+    {
+        $concurrentLock = new Lock(new Key('locked'), $store = new InMemoryStore(), autoRelease: false);
+        $concurrentLock->acquire();
+        $this->assertTrue($concurrentLock->isAcquired());
+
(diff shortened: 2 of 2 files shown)


src/Symfony/Component/Scheduler/Generator/Checkpoint.php
```diff
@@ -31,17 +30,9 @@ public function __construct(
...
         }
 
-        if ($this->reset) {
-            $this->reset = false;
-            $this->save($now, -1);
-        }
-
         if ($this->cache) {
             [$this->time, $this->index, $this->from] = $this->cache->get($this->name, fn () => [$now, -1, $now]) + [2 => $now];
             $this->save($this->time, $this->index);
```
Jeroeny, line 42: Or perhaps:

```suggestion
      if ($this->reload) {
             [$this->time, $this->index, $this->from] = $this->cache->get($this->name, fn () => [$now, -1, $now]) + [2 => $now];
        }

        if ($this->cache) {
            $this->save($this->time, $this->index);
```

With property set to true as default:
```php
 private bool $reload = false;
```

And set it to true where currently `reset` was set to true.

nicolas-grekas:
I'm trying to understand what you mean here, I'm not deep enough into the topic :grimacing: 

Does your description also apply when there is no cache defined?

How about this patch instead?

```patch
--- a/src/Symfony/Component/Scheduler/Generator/Checkpoint.php
+++ b/src/Symfony/Component/Scheduler/Generator/Checkpoint.php
@@ -37,14 +37,12 @@ final class Checkpoint implements CheckpointInterface
             return false;
         }
 
-        if ($this->reset) {
-            $this->reset = false;
-            $this->save($now, -1);
-        }
-
         if ($this->cache) {
             [$this->time, $this->index, $this->from] = $this->cache->get($this->name, fn () => [$now, -1, $now]) + [2 => $now];
             $this->save($this->time, $this->index);
+        } elseif ($this->reset) {
+            $this->reset = false;
+            $this->save($now, -1);
         }
```

src/Symfony/Component/Scheduler/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ CHANGELOG
  * Add `ScheduledStamp` to `RedispatchMessage`
  * Allow modifying Schedule instances at runtime
  * Add `MessageProviderInterface` to trigger unique messages at runtime
+ * Continue from stored `Checkpoint` time when a lock is acquired
```
nicolas-grekas: I think we should consider this as a bugfix, so no need for that line

Jeroeny:
> Does you description also apply when there is no cache defined?

That's an interesting case. I think that may indeed be the thought behind this reset. 
Because if you have two processes running without shared cache, apparently you do not want them to continue from exactly the last saved time. So maybe to prevent re-processing a certain time frame in a second worker when it has acquired a lock after another worker, it gets reset. 

But, if they do have a shared cache, I think worker two can continue where worker one left off. As I'd say that is the purpose of the cache. 
 
> How about this patch instead?

With the above reasoning, that patch does indeed work. I've added a test to specifically test a second checkpoint acquiring the lock after the first and test that it continues from the saved time.

nicolas-grekas:
Thank you @Jeroeny.

Issue #51867, opened 2023-10-06
Symfony\Component\Cache\Adapter\ArrayAdapter::freeze(): Return value must be of type array|string|int|float|bool|null


motoronik:
### Symfony version(s) affected

6.3.5

### Description

After updating to the latest version of Symfony the application was broken. I received a large number of the same errors

```
[TypeError] Symfony\Component\Cache\Adapter\ArrayAdapter::freeze(): Return value must be of type array|string|int|float|bool|null, MyClassName returned
#1  /var/www/apps/hybrid/vendor/symfony/cache/Adapter/ArrayAdapter.php:328
#2  /var/www/apps/hybrid/vendor/symfony/cache/Adapter/ArrayAdapter.php:177
```

The problem is that not so long ago a typification of the return value of the freeze method appeared

https://github.com/symfony/symfony/blob/d27190a4cf8560b77d8aaab65104e51a2dd5c06c/src/Symfony/Component/Cache/Adapter/ArrayAdapter.php#L301

But it is possible to return a value that does not correspond to these types - an object.

https://github.com/symfony/symfony/blob/d27190a4cf8560b77d8aaab65104e51a2dd5c06c/src/Symfony/Component/Cache/Adapter/ArrayAdapter.php#L323-L325

If the serialized value does not meet this condition, then an object will be returned that does not match the method's typing.
And since the return type of the method does not support the object, the code will fail. 

What happens if enum is serialized

```
php > enum TestEnum: string { case Test1 = 'test1'; }
php > class TestClass { public string $property1 = '';}
php >
php > $obj = new TestClass();
php > $enum = TestEnum::Test1;
php >
php > var_dump(serialize($obj));
string(43) "O:9:"TestClass":1:{s:9:"property1";s:0:"";}"
php >
php > var_dump(serialize($enum));
string(22) "E:14:"TestEnum:Test1";"
php >
```
This means that the enum will not match the condition and the enum object itself will be returned and not its serialized representation

### How to reproduce

```php
$obj = new TestClass();
$enum = TestEnum::Test1;

$item1 = new CacheItem();
$item1->expiresAt(new \DateTime('now +1 day'));
$item1->set($obj);

$item2 = new CacheItem();
$item2->expiresAt(new \DateTime('now +1 day'));
$item2->set($enum);

$adapter->save($item1);
$adapter->save($item2);
```

```
[...]

PR #52040, opened 2023-10-13
[Cache] Fix ArrayAdapter::freeze() return type

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | https://github.com/symfony/symfony/issues/51867
| License       | MIT

Enums are not serialized and this is good: they don't need to be frozen since they are immutable.
The fix is to widen the return type.


nicolas-grekas:
Thank you @fancyweb.

PR #52044, opened 2023-10-13
[DoctrineBridge] Spread some PHP 8 love

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT

This PR modernizes the codebase of the DoctrineBridge with property types, CPP and `readonly` where we can do that without breaking backwards compatibility.

### DIFF
--- a/src/Symfony/Bridge/Doctrine/DependencyInjection/CompilerPass/RegisterMappingsPass.php
+++ b/src/Symfony/Bridge/Doctrine/DependencyInjection/CompilerPass/RegisterMappingsPass.php
@@ -74,25 +74,6 @@ abstract class RegisterMappingsPass implements CompilerPassInterface
     protected $enabledParameter;
 
     /**
-     * Naming pattern for the configuration service id, for example
-     * 'doctrine.orm.%s_configuration'.
-     */
-    private string $configurationPattern;
-
-    /**
-     * Method name to call on the configuration service. This depends on the
-     * Doctrine implementation. For example addEntityNamespace.
-     */
-    private string $registerAliasMethodName;
-
-    /**
-     * Map of alias to namespace.
-     *
-     * @var string[]
-     */
-    private array $aliasMap;
-
-    /**
      * The $managerParameters is an ordered list of container parameters that could provide the
      * name of the manager to register these namespaces and alias on. The first non-empty name
      * is used, the others skipped.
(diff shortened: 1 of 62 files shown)


src/Symfony/Bridge/Doctrine/DependencyInjection/CompilerPass/RegisterMappingsPass.php
```diff
@@ -108,24 +89,32 @@ abstract class RegisterMappingsPass implements CompilerPassInterface
...
+        private readonly string $registerAliasMethodName = '',
+        private readonly array $aliasMap = [],
+    ) {
         $this->driver = $driver;
         $this->namespaces = $namespaces;
         $this->managerParameters = $managerParameters;
         $this->driverPattern = $driverPattern;
         $this->enabledParameter = $enabledParameter;
+
         if (\count($aliasMap) && (!$configurationPattern || !$registerAliasMethodName)) {
```
nicolas-grekas: while at it
```suggestion
        if ($aliasMap && (!$configurationPattern || !$registerAliasMethodName)) {
```

src/Symfony/Bridge/Doctrine/Form/ChoiceList/DoctrineChoiceLoader.php
```diff
@@ -36,18 +34,17 @@ class DoctrineChoiceLoader extends AbstractChoiceLoader
      *
      * @param string $class The class name of the loaded objects
      */
-    public function __construct(ObjectManager $manager, string $class, IdReader $idReader = null, EntityLoaderInterface $objectLoader = null)
-    {
-        $classMetadata = $manager->getClassMetadata($class);
-
+    public function __construct(
+        private readonly ObjectManager $manager,
+        string $class,
```
nicolas-grekas, line 39: we can still declare the property here if we want

src/Symfony/Bridge/Doctrine/Form/ChoiceList/DoctrineChoiceLoader.php
```diff
@@ -36,18 +34,17 @@ class DoctrineChoiceLoader extends AbstractChoiceLoader
      *
      * @param string $class The class name of the loaded objects
```
nicolas-grekas, line 35:
```suggestion
     * @param class-string $class The class name of the loaded objects
```

nicolas-grekas (commented):
LGTM, I just have minor comments.

src/Symfony/Bridge/Doctrine/Form/ChoiceList/DoctrineChoiceLoader.php
derrabus: We use the the EM to resolve the actual class from this string which could be the ORM's legacy short notation. We could deprecate not passing an actual FQCN though, but I'd like to do that in a separate PR.

src/Symfony/Bridge/Doctrine/Form/ChoiceList/DoctrineChoiceLoader.php
derrabus: Yes, but not as `readonly` then because we change its value within the constructor.

src/Symfony/Bridge/Doctrine/Form/ChoiceList/DoctrineChoiceLoader.php
stof: Well, if you pass a legacy short notation, Doctrine ORM will trigger a deprecation warning in `$manager->getClassMetadata($class)` in the constructor. So we are already covered here. No need for a duplicate deprecation warning.
The code will be simplified once we drop support for ORM 2.x (which will probably be in Symfony 8.0)

derrabus: We could simplify it for 7.0 already by not resolving the class name through the EM anymore.

nicolas-grekas:
Thank you @derrabus.

Issue #51912, opened 2023-10-09
[HttpKernel] ErrorListener::onKernelException should ignore exception thrown after `kernel.terminate`


VincentLanglet:
### Symfony version(s) affected

all

### Description

Related to https://github.com/symfony/symfony/issues/45205 issue, which was ignored...

When an exception is thrown in a `kernel.terminate` event,
- `kernel.exception` is thrown
- the Symfony ErrorListener::onKernelException handler the exception
- `$response = $event->getKernel()->handle($request, HttpKernelInterface::SUB_REQUEST, false);` is called
- this trigger the `ErrorController` and then the `TwigErrorRenderer`

1) The response rendered seems useless since the response was already sent before the terminate event. (cf https://symfony.com/doc/current/components/http_kernel.html#8-the-kernel-terminate-event) 
2) It can produce extra-useless exception, in my case I have a custom view `TwigBundle/Exception/error.html.twig`
which use the session and/or flashbag and I'm getting some `Failed to start the session because headers have already been sent`.

Introducing a `kernel.terminate_exception` and using it instead of `kernel.exception` could even appear as a BC break for `kernel.exception` listeners or a bugfix because according to the schema in the doc https://symfony.com/doc/current/components/http_kernel.html#handling-exceptions-the-kernel-exception-event, we can see the order `Exception -> Response -> Terminate`.

### How to reproduce

- Throw an exception in a `kernel.terminate` listener -> You'll get extra code executed for no reason
- Add `app.session.flashbag.get('success')` call in a custom `TwigBundle/Exception/error.html.twig` template. -> You'll get failed to start the session error in your logs.

### Possible Solution

Either throw a different event for exception thrown after the terminate exception

Either, at least provide a `ExceptionEvent::isTerminated` or `TerminableInterface::isTerminated`, which would allow to write
```
if ($thing->isTerminated()) {
     return;
}
```
at the beginning of some ExceptionListener like the ErrorListener

fancyweb:
The current logic might actually be "OK" enough.
When `fastcgi_finish_request()` is not called, then the extra response and exception are actually output and are useful.
Also, I'm pretty sure some people rely on the current logic without knowing it :sweat_smile:

VincentLanglet:
> The current logic might actually be "OK" enough. When `fastcgi_finish_request()` is not called, then the extra response and exception are actually output and are useful. 
> Also, I'm pretty sure some people rely on the current logic without knowing it 😅

[...]

Issue #45205, opened 2022-01-27
[HttpKernel] Exception after kernel.terminate runs kernel.exception, kernel.response and kernel.finish_request listeners again


fancyweb:
### Symfony version(s) affected

All

### Description

If a `kernel.terminate` listener throws an exception, `HttpKernel` ends up dispatching `kernel.exception` again.

Then, if a `kernel.exception` listener sets a response, `kernel.response` is redispatched a second time while the response has already been sent (a lot of "useless" code is executed).

Finally, `kernel.finish_request` is always redispatched a second time.

### How to reproduce

Throw an exception in a `kernel.terminate` listener.

### Possible Solution

1. I think dispatching `kernel.exception` again can make sense. But we should stop there.

2. We can also stop before dispatching it.

Also if this is really a wrong behavior, the subject is probably sensible, I don't think we can treat it as real a bug fix, we would need to deprecate it first.

fancyweb:
Also if we go with option 1 it would be great to know whether we are in a pre or post terminate situation in `ExceptionEvent`. We could also have `kernel.terminate_exception` instead :man_shrugging:

carsonbot:
Hey, thanks for your report!
There has not been a lot of activity here for a while. Is this bug still relevant? Have you managed to find a workaround?

carsonbot:
Friendly reminder that this issue exists. If I don't hear anything I'll close this.

carsonbot:
Hey,

I didn't hear anything so I'm going to close it. Feel free to comment if this is still relevant, I can always reopen!

VincentLanglet:
@carsonbot, this issue is still relevant.

Hi @fancyweb, I recently encounter an issue which I think seems related to your point.

In my TerminateListener an exception is thrown
- with Symfony ErrorHandler/DebugHandlersListener a `KernelEvents::EXCEPTION` is thrown
- this trigger the Symfony ErrorListener
- this trigger the ErrorController
- then the TwigErrorRenderer
- and it try to render my custom TwigBundle/Exception/error.html.twig
- which logs an exception because I use app.session.flashbag when the header are already sent

But like you said the extra response is useless.

I think indeed a `kernel.terminate_exception` should be sent instead of a classic `kernel.exception` or at least a way to differentiate those situation would help listeners sont handle differently the situation, because the Symfony ErrorController shouldn't be called for exception thrown in terminate events therefore
```
ErrorListener::onKernelException
```
should have a early return if the exception was thrown in a terminate event.

PR #52047, opened 2023-10-13
[HttpFoundation][Runtime] Add $flush parameter to Response::send()

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT

Inspired by https://github.com/symfony/symfony/issues/51912 and https://github.com/symfony/symfony/issues/45205 subjects

Adds optional argument `$flush` to `Response::send()`. If `$flush === false`, output buffers are not flushed/`*_finish_request()` and alike functions are not called.

We leverage that in the Runtime component by not flushing the output buffers when debug mode is on in order to see exceptions thrown in listeners listening on the `TerminateEvent` and also to feel a more "real" processing time of things happening on terminate.

### DIFF
--- a/src/Symfony/Component/HttpFoundation/Response.php
+++ b/src/Symfony/Component/HttpFoundation/Response.php
@@ -415,12 +415,19 @@ class Response
     /**
      * Sends HTTP headers and content.
      *
+     * @param bool $flush Whether output buffers should be flushed
+     *
      * @return $this
      */
-    public function send(): static
+    public function send(/* bool $flush = true */): static
     {
         $this->sendHeaders();
         $this->sendContent();
+
+        $flush = 1 <= \func_num_args() ? func_get_arg(0) : true;
+        if (!$flush) {
+            return $this;
+        }
 
         if (\function_exists('fastcgi_finish_request')) {
             fastcgi_finish_request();
--- a/src/Symfony/Component/Runtime/Runner/Symfony/HttpKernelRunner.php
+++ b/src/Symfony/Component/Runtime/Runner/Symfony/HttpKernelRunner.php
@@ -24,13 +24,14 @@ class HttpKernelRunner implements RunnerInterface
     public function __construct(
         private readonly HttpKernelInterface $kernel,
         private readonly Request $request,
+        private readonly bool $debug = false,
     ) {
     }
 
     public function run(): int
     {
         $response = $this->kernel->handle($this->request);
-        $response->send();
+        $response->send(!$this->debug);
 
         if ($this->kernel instanceof TerminableInterface) {
             $this->kernel->terminate($this->request, $response);
--- a/src/Symfony/Component/HttpFoundation/CHANGELOG.md
+++ b/src/Symfony/Component/HttpFoundation/CHANGELOG.md
@@ -8,6 +8,7 @@ CHANGELOG
  * Support root-level `Generator` in `StreamedJsonResponse`
  * Add `UriSigner` from the HttpKernel component
  * Add `partitioned` flag to `Cookie` (CHIPS Cookie)
+ * Add argument `bool $flush = true` to `Response::send()`
 
 6.3
 ---
--- a/src/Symfony/Component/Runtime/CHANGELOG.md
+++ b/src/Symfony/Component/Runtime/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+ * Add argument `bool $debug = false` to `HttpKernelRunner::__construct()`
+
 5.4
 ---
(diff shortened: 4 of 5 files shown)


src/Symfony/Component/Runtime/Runner/Symfony/HttpKernelRunner.php
```diff
@@ -24,13 +25,24 @@ class HttpKernelRunner implements RunnerInterface
...
         private readonly Request $request,
+        private readonly bool $debug = false,
     ) {
     }
 
     public function run(): int
     {
         $response = $this->kernel->handle($this->request);
-        $response->send();
+        $response->sendHeaders();
```
fancyweb: Copy-pasting things here is easier and looks cleaner than somehow adding a "debug" logic in the Response class.

nicolas-grekas (dismissed):
Difficult to test, anyway, LGTM!

src/Symfony/Component/Runtime/Runner/Symfony/HttpKernelRunner.php
```diff
@@ -24,13 +25,24 @@ class HttpKernelRunner implements RunnerInterface
...
         private readonly HttpKernelInterface $kernel,
         private readonly Request $request,
+        private readonly bool $debug = false,
     ) {
     }
 
     public function run(): int
     {
         $response = $this->kernel->handle($this->request);
-        $response->send();
```
stof, line 33: What if a custom response class customizes its send method ? The method is not final.

fancyweb: Yes, that's a problem for sure. Should we make it final then? Or we could try to add an optional $finishRequest arg to it?

nicolas-grekas: Good idea the extra argument!

src/Symfony/Component/HttpFoundation/Response.php
```diff
@@ -415,16 +415,18 @@ public function sendContent(): static
...
      * @return $this
      */
-    public function send(): static
+    public function send(/* bool $finishRequest = true */): static
     {
         $this->sendHeaders();
         $this->sendContent();
 
-        if (\function_exists('fastcgi_finish_request')) {
+        if (\function_exists('fastcgi_finish_request') && (1 <= \func_num_args() ? func_get_arg(0) : true)) {
```
nicolas-grekas: The `$finishRequest` variable should be created on a separate line. This way, the diff with 7.0 will be minimal.

fancyweb: I did that to not compute it if the functions don't exist. Overkill?

stof: yes, this is overkill and makes maintenance harder

src/Symfony/Component/HttpFoundation/Response.php
```diff
@@ -415,16 +415,19 @@ public function sendContent(): static
     /**
      * Sends HTTP headers and content.
      *
+     * @param bool $finishRequest Whether fastcgi_finish_request() and alike functions should be called or not
```
nicolas-grekas:
```suggestion
     * @param bool $finishRequest Whether fastcgi_finish_request() and alike functions should be called
```

nicolas-grekas (commented):
BTW @dunglas : does frankenphp provide an equivalent to `fastcgi_finish_request()`?

dunglas:
Yes, it implements the function: https://github.com/dunglas/frankenphp/pull/69

src/Symfony/Component/HttpFoundation/Response.php
```diff
@@ -415,16 +415,19 @@ public function sendContent(): static
...
      */
-    public function send(): static
+    public function send(/* bool $finishRequest = true */): static
     {
         $this->sendHeaders();
         $this->sendContent();
 
-        if (\function_exists('fastcgi_finish_request')) {
+        $finishRequest = 1 <= \func_num_args() ? func_get_arg(0) : true;
+        if (\function_exists('fastcgi_finish_request') && $finishRequest) {
```
nicolas-grekas: I think it would be even more flexible to also skip calling closeOutputBuffers when $finishRequest is false. This way, we can make this method side-effect-free. It'd also allow decoupling from PHP_SAPI, which will benefit #52079

nicolas-grekas (approved):

nicolas-grekas:
Thank you @fancyweb.

PR #52054, opened 2023-10-14
[Console] `InputArgument` and `InputOption` code cleanup

jnoordsij:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | n/a
| License       | MIT

This updates the `InputOption` and `InputArgument` classes to be slightly more consistent. Moreover it adds some additional dockblocks that were not present yet.

### DIFF
--- a/src/Symfony/Component/Console/Input/InputOption.php
+++ b/src/Symfony/Component/Console/Input/InputOption.php
@@ -46,18 +46,18 @@ class InputOption
     public const VALUE_IS_ARRAY = 8;
 
     /**
-     * The option may have either positive or negative value (e.g. --ansi or --no-ansi).
+     * The option allows passing a negated variant (e.g. --ansi or --no-ansi).
      */
     public const VALUE_NEGATABLE = 16;
 
     private string $name;
-    private string|array|null $shortcut;
+    private ?string $shortcut;
     private int $mode;
     private string|int|bool|array|float|null $default;
 
     /**
      * @param string|array|null                                                             $shortcut        The shortcuts, can be null, a string of shortcuts delimited by | or an array of shortcuts
-     * @param int|null                                                                      $mode            The option mode: One of the VALUE_* constants
+     * @param int-mask-of<InputOption::*>|null                                              $mode            The option mode: One of the VALUE_* constants
      * @param string|bool|int|float|array|null                                              $default         The default value (must be null for self::VALUE_NONE)
      * @param array|\Closure(CompletionInput,CompletionSuggestions):list<string|Suggestion> $suggestedValues The values used for input completion
      *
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/Console/Input/InputOption.php
```diff
@@ -51,9 +51,9 @@ class InputOption
     public const VALUE_NEGATABLE = 16;
 
     private string $name;
-    private string|array|null $shortcut;
+    private ?string $shortcut;
```
GromNaN, line 54: Good catch.

src/Symfony/Component/Console/Input/InputOption.php
```diff
@@ -173,13 +173,22 @@ public function isArray(): bool
         return self::VALUE_IS_ARRAY === (self::VALUE_IS_ARRAY & $this->mode);
     }
 
+    /**
+     * Returns true if the option may have either a positive or negative value.
```
GromNaN: Reading "negative value" I understand that I can pass a negative int (i.e. something like `-5`, that is not supported by the input parser).
That's an other feature described in this blog post: https://symfony.com/blog/new-in-symfony-5-3-negatable-command-options

src/Symfony/Component/Console/Input/InputOption.php
```diff
@@ -173,13 +173,22 @@ public function isArray(): bool
...
     {
         return self::VALUE_NEGATABLE === (self::VALUE_NEGATABLE & $this->mode);
     }
 
     /**
+     * Sets the default value.
+     *
      * @return void
+     *
+     * @throws LogicException When incorrect default value is given
```
GromNaN: We have a lot of logical errors in the codebase, I don't think it's appropriate to add this annotation here. `@throw` is more for runtime exceptions, when the exception may occur at random.

GromNaN (commented):
- We can use [`int-mask-of<InputArgument::*>`](https://phpstan.org/writing-php-code/phpdoc-types#integer-masks) for the constructor `$mode` param.
- The comment for `InputArgument::complete` should be improved to explain that it's called when the command resolves input completion for this argument, same for `InputOption`.

src/Symfony/Component/Console/Input/InputOption.php
jnoordsij: Good point. I added it here to match `InputOption`, but I've now gone with removing it on both places.

src/Symfony/Component/Console/Input/InputOption.php
jnoordsij: I copied it directly from the `VALUE_NEGATABLE` docblock. I've now rephrased both to be more specific and clear in my opinion; please let me know if you have other ideas about the phrasing.

GromNaN (approved):

OskarStark:
rebased, squashed, cs fixed ✅

OskarStark (approved):

fabpot (approved):

fabpot:
Thank you @jnoordsij.

PR #52055, opened 2023-10-14
[Console] Deprecate combining incompatible mode flags in `InputArgument` and `InputOption`

jnoordsij:
| Q             | A
| ------------- | ---
| Branch?       | 8.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Tickets       | n/a
| License       | MIT

Currently it is allowed to to combine various combinations of input value requirement specifiers, which are logically exclusive behavior-wise. To prevent mistakes or possibly ambiguous behavior, this PR adds additional validation to ensure there's always exactly one specifier part of the mode after finishing the constructor.

This is done by adding a deprecation message for these cases, which can be replaced with an `InvalidArgumentException` in a future major version. The tests added are currently checking for the deprecation message and could be updated to check for this exception instead.

PR #52058, opened 2023-10-14
[Console] Allow setting a boolean default value on `InputOption::VALUE_NEGATABLE` options

jnoordsij:
| Q             | A
| ------------- | ---
| Branch?       | 8.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       |  -
| License       | MIT

(Encountered on test run for #52055)

Currently, the following call/construction of an `InputOption` is valid:
```php
new InputOption('foo', null, InputOption::VALUE_NEGATABLE, '', false)
```
while the following call is not
```php
new InputOption('foo', null, InputOption::VALUE_NONE | InputOption::VALUE_NEGATABLE, '', false)
```

To me, this seems inconsistent, as I would expect both calls to behave in exactly the same manner. This PR ensures both calls are equally valid.

Note: a follow-up might be ensuring the passed default value is boolean when using `InputOption::VALUE_NEGATABLE`; I can add that to #52055 once this is merged.

### DIFF
--- a/src/Symfony/Component/Console/Input/InputOption.php
+++ b/src/Symfony/Component/Console/Input/InputOption.php
@@ -190,8 +190,12 @@ class InputOption
      */
     public function setDefault(mixed $default): void
     {
-        if (self::VALUE_NONE === (self::VALUE_NONE & $this->mode) && null !== $default) {
+        if (self::VALUE_NONE === (self::VALUE_NONE & $this->mode) && !$this->isNegatable() && null !== $default) {
             throw new LogicException('Cannot set a default value when using InputOption::VALUE_NONE mode.');
+        }
+
+        if ($this->isNegatable() && null !== $default && !\is_bool($default)) {
+            throw new LogicException('A default value for a negatable option must be a boolean or null.');
         }
 
         if ($this->isArray()) {
--- a/src/Symfony/Component/Console/CHANGELOG.md
+++ b/src/Symfony/Component/Console/CHANGELOG.md
@@ -16,6 +16,7 @@ CHANGELOG
  * Add `TesterTrait::assertCommandIsInvalid()` to test command
  * Add a result-based testing API with `CommandTester::run()`, `ExecutionResult`, and `ConsoleAssertionsTrait` to assert output and error streams together
  * Add optional `$format` argument to `SymfonyStyle::createProgressBar()`, `SymfonyStyle::progressStart()`, and `SymfonyStyle::progressIterate()` to allow passing a custom `ProgressBar` format string
+ * Allow setting a boolean default value on `InputOption::VALUE_NEGATABLE` options
 
 8.0
 ---
--- a/src/Symfony/Component/Console/Tests/Input/ArgvInputTest.php
+++ b/src/Symfony/Component/Console/Tests/Input/ArgvInputTest.php
@@ -228,6 +228,12 @@ class ArgvInputTest extends TestCase
                 ['foo' => false],
                 '->parse() parses long options without a value',
             ],
+            [
+                ['cli.php'],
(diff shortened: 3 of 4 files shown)


src/Symfony/Component/Console/Input/InputOption.php
```diff
@@ -177,7 +177,7 @@ public function isNegatable(): bool
      */
     public function setDefault($default = null)
     {
-        if (self::VALUE_NONE === (self::VALUE_NONE & $this->mode) && null !== $default) {
+        if (self::VALUE_NONE === (self::VALUE_NONE & $this->mode) && !$this->isNegatable() && null !== $default) {
```
GromNaN, line 193: If `VALUE_NEGATABLE` means `VALUE_NONE`, we can force `$this->mode |= self::VALUE_NODE` when `VALUE_NEGATABLE` is detected. That would also throw an error when `VALUE_NEGATABLE | VALUE_OPTIONAL` is given (with your proposition in #52055).

jnoordsij: The first bit is already covered in #52055: if `VALUE_REQUIRED` or `VALUE_OPTIONAL` is not provided, then `VALUE_NONE` is added by default:
https://github.com/symfony/symfony/blob/251b2424bc234ac0676db53f398194da1c822fbb/src/Symfony/Component/Console/Input/InputOption.php#L95-L96
So #52055 makes `VALUE_NEGATABLE | VALUE_NONE` and (just) `VALUE_NEGATABLE` behave in exactly the same manner, which was the reason for me finding this bug.

The error on `VALUE_NEGATABLE | VALUE_OPTIONAL` (or `VALUE_REQUIRED`) already exists:
https://github.com/symfony/symfony/blob/ba2f4a5d9cbab20945c60d57e570fe3004ca1c51/src/Symfony/Component/Console/Input/InputOption.php#L103-L105

nicolas-grekas:
I'm not sure I understand the purpose. Why don't you use `new InputOption('foo', null, InputOption::VALUE_NONE | InputOption::VALUE_NEGATABLE)` if you don't want a value?

jnoordsij:
> I'm not sure I understand the purpose. Why don't you use `new InputOption('foo', null, InputOption::VALUE_NONE | InputOption::VALUE_NEGATABLE)` if you don't want a value?

I think passing an option for the negatable case does make sense, since it basically ensures the option resolves as boolean instead of nullable boolean, which could be useful or at least clearer in some applications. 

I do admit it looks slightly off when combined with an explicit `VALUE_NONE`, but the main issue I had with this was the inconsistency of how the options were handled, which prompted me to open #52055. To illustrate, the following docblock actually claims not setting `VALUE_REQUIRED` or `VALUE_OPTIONAL` is considered to be equal to actually setting `VALUE_NONE`:
https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Console/Input/InputOption.php#L130-L138

So given this, I would expect `VALUE_NONE` and `VALUE_NONE | VALUE_NEGATABLE` to behave exactly the same in all cases, as `VALUE_NEGATABLE` in my eyes is just a slightly more specific variant of `VALUE_NONE`, i.e. an option that does not allow a manual value, but does allow passing the negated version.

jnoordsij:
Retargeted to `6.4` and rebased

stof:
we should probably validate also that the default value of a negatable option is a boolean.

However, this looks more like a new feature than a bugfix to me, so it should probably go in 7.4.

nicolas-grekas (approved):
(rebased on 8.1 as a new feature)

nicolas-grekas:
Thank you @jnoordsij.

PR #52063, opened 2023-10-14
[Mime] Update mime types 

smnandre:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

The recent update done in august (#51191) used an outdated version of the jshttp/mime-db database ([See 1.49.0 .. 1.52.0](https://github.com/jshttp/mime-db/compare/v1.49.0...v1.52.0#diff-bd4a9ff1dadaff3890766954fd9c12347e8cea7cd9bcc189f04c59d3049158a9))

The original reason I started this PR was to add the `application/yaml` type
* https://www.ietf.org/archive/id/draft-ietf-httpapi-yaml-mediatypes-10.html
* https://www.iana.org/assignments/media-types/media-types.xhtml

I fixed update_mime_types.php too, because the script failed to update the date in the `MAP` docblock.


smnandre:
(test failures unrelated)

src/Symfony/Component/Mime/Tests/RawMessageTest.php
```diff
@@ -52,7 +52,11 @@ public static function provideMessages(): array
             'string' => ['some string', true],
             'traversable' => [new \ArrayObject(['some', ' ', 'string']), true],
             'array' => [['some', ' ', 'string'], true],
-            'generator' => [(function () { yield 'some'; yield ' '; yield 'string'; })(), false],
+            'generator' => [(function () {
```
OskarStark: Should be reverted

OskarStark (approved):

src/Symfony/Component/Mime/Tests/RawMessageTest.php
smnandre: Done

src/Symfony/Component/Mime/Resources/bin/update_mime_types.php
```diff
@@ -14,7 +14,7 @@
 }
 
 // load new map
-$data = json_decode(file_get_contents('https://cdn.jsdelivr.net/gh/jshttp/mime-db@v1.49.0/db.json'), true);
+$data = json_decode(file_get_contents('https://cdn.jsdelivr.net/gh/jshttp/mime-db@v1.52.0/db.json'), true);
```
stof: is there a way to make it use the latest version each time the script is run (as done for the freedesktop data below where we use the master branch) ?

nicolas-grekas: `https://cdn.jsdelivr.net/gh/jshttp/mime-db/db.json`?

src/Symfony/Component/Mime/MimeTypes.php
```diff
@@ -3145,7 +3209,6 @@ public function guessMimeType(string $path): ?string
         'snd' => ['audio/basic'],
         'snf' => ['application/x-font-snf'],
         'so' => ['application/x-sharedlib'],
-        'so.[0-9]*' => ['application/x-sharedlib'],
```
nicolas-grekas, line 3148: Can you check why this was removed?

stof: The update script was updated to only put actual extensions in that list and not patterns (as our API does not deal with patterns). My guess is that the file was not properly regenerated when we did that change last time (otherwise it would have been removed at that time)

nicolas-grekas:
Thank you @smnandre.

PR #52066, opened 2023-10-15
Proofread UPGRADE guide

wouterj:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT

* Clarify some entries by mentioning the replacement or showing before/example examples
* Consistently use "Rename ... to ..." when the behavior didn't change and "Remove ..., use ... instead" when the behavior *did* change
* Add table of contents, to make it easier to navigate and group bundles and components (the alphabetical order of the sections messes this up, even mentioning e.g. the SecurityBundle after the Security component).
* Mentioning the native PHP types in general.

### DIFF
--- a/UPGRADE-7.0.md
+++ b/UPGRADE-7.0.md
@@ -1,15 +1,58 @@
 UPGRADE FROM 6.4 to 7.0
 =======================
 
-Symfony 6.4 and Symfony 7.0 will be released simultaneously at the end of November 2023. According to the Symfony
-release process, both versions will have the same features, but Symfony 7.0 won't include any deprecated features.
+Symfony 6.4 and Symfony 7.0 are released simultaneously at the end of November 2023. According to the Symfony
+release process, both versions have the same features, but Symfony 7.0 doesn't include any deprecated features.
 To upgrade, make sure to resolve all deprecation notices.
-Read more about this in the [Symfony documentation](https://symfony.com/doc/current/setup/upgrade_major.html).
+Read more about this in the [Symfony documentation](https://symfony.com/doc/7.0/setup/upgrade_major.html).
+
+Symfony 7.0 introduced many native return and property types. Read [the announcement blogpost](https://symfony.com/blog/symfony-7-0-type-declarations)
+on how to quickly make your code compatible.
+
+Table of Contents
+-----------------
+
+Bundles
+ * [FrameworkBundle](#FrameworkBundle)
+ * [SecurityBundle](#SecurityBundle)
+ * [TwigBundle](#TwigBundle)
+
+Bridges
+ * [DoctrineBridge](#DoctrineBridge)
+ * [MonologBridge](#MonologBridge)
+ * [ProxyManagerBridge](#ProxyManagerBridge)
+
+Components
+ * [Cache](#Cache)
+ * [Config](#Config)
+ * [Console](#Console)
+ * [DependencyInjection](#DependencyInjection)
+ * [DomCrawler](#DomCrawler)
+ * [ExpressionLanguage](#ExpressionLanguage)
+ * [Filesystem](#Filesystem)
+ * [Form](#Form)
+ * [HttpFoundation](#HttpFoundation)
+ * [HttpClient](#HttpClient)
+ * [HttpKernel](#HttpKernel)
+ * [Lock](#Lock)
+ * [Mailer](#Mailer)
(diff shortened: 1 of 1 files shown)


UPGRADE-7.0.md
```diff
@@ -199,20 +240,25 @@ FrameworkBundle
...
+ * Change default of some config options:
+
+   | option                                       | default Symfony <7.0     | default in Symfony 7.0+                                                   |
+   | -------------------------------------------- | ------------------------ | ------------------------------------------------------------------------- |
+   | `framework.http_method_override`             | `true`                   | `false`                                                                   |
+   | `framework.handle_all_throwables`            | `false`                  | `true`                                                                    |
+   | `framework.php_errors.log`                   | `kernel.debug`           | `true`                                                                    |
+   | `framework.session.cookie_secure`            | `false`                  | `auto`                                                                    |
+   | `framework.session.cookie_samesite`          | `lax`                    | `lax`                                                                     |
```
wouterj: I'm a bit unsure about this deprecation. As you can see, I believe we did not change the default value between 6.4 and 7.0. Either I'm missing something, or we incorrectly force everyone to set this config option on 6.4?

UPGRADE-7.0.md
```diff
@@ -4,12 +4,55 @@ UPGRADE FROM 6.4 to 7.0
...
+ * [Validator](#Validator)
+ * [VarDumper](#VarDumper)
+ * [Workflow](#Workflow)
+ * [Yaml](#Yaml)
 
 Cache
 -----
 
- * Add parameter `$isSameDatabase` to `DoctrineDbalAdapter::configureSchema()`
+ * Add argument `\Closure $isSameDatabase` to `DoctrineDbalAdapter::configureSchema()`
```
derrabus: Parameter is the correct term.

A method signature contains _parameters_, you call that method with _arguments_.

UPGRADE-7.0.md
```diff
@@ -98,8 +141,8 @@ DependencyInjection
    ```
 
    To fetch services lazily, you can use a [service subscriber](https://symfony.com/doc/6.4/service_container/service_subscribers_locators.html#defining-a-service-subscriber).
- * Add argument `$id` and `$asGhostObject` to `DumperInterface::isProxyCandidate()` and `getProxyCode()`
- * Add argument `$source` to `FileLoader::registerClasses()`
+ * Add argument `string $id = null` and `bool &$asGhostObject = null` to `LazyProxy\PhpDumper\DumperInterface::isProxyCandidate()` and `getProxyCode()`
+ * Add argument `string $source = null` to `FileLoader::registerClasses()`
```
chalasr:
```suggestion
 * Add parameter `string $source = null` to `FileLoader::registerClasses()`
```
While at it 😇

UPGRADE-7.0.md
```diff
@@ -260,14 +306,66 @@ Mailer
...
+       // ...
+   }
+   ```
+
+   *After*
+   ```php
+   use Symfony\Component\Messenger\Attribute\AsMessageHandler;
+
+   #[AsMessageHandler]
+   class SmsNotificationHandler implements MessageHandlerInterface
```
BafS:
```suggestion
   class SmsNotificationHandler
```

Isn't it like that if we remove the interface?

UPGRADE-7.0.md
```diff
@@ -260,14 +306,66 @@ Mailer
...
+   {
+       public function __invoke(SmsNotification $message): void
+       {
+           // ...
+       }
+   }
+
+   class UploadedImageHandler implements MessageSubscriberInterface
```
BafS:
```suggestion
   class UploadedImageHandler
```

same question here

UPGRADE-7.0.md
chalasr:
```suggestion
   | `framework.session.cookie_samesite`          | `null`                    | `lax`                                                                     |
```
We did change it from `null` to `lax` in https://github.com/symfony/symfony/commit/4fe6f5b2bc41ad515998a4d977b3158ddbd6fe98#diff-041b166c698c254a3f61cfc6390785d4f5097794b77a4b2f3f888996729cbce4L669

chalasr (approved):

fabpot:
Thank you @wouterj.

PR #52075, opened 2023-10-16
[Messenger] Fix DoctrineOpenTransactionLoggerMiddleware

ro0NL:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #... 
| License       | MIT

When a message is routed via `sync://` we hit the middleware stack twice (meh, but OK).

A side effect is we always get an error log due open transaction, even though it was closed properly. But it's an inner transaction.

This solves it.


stof (approved):

ro0NL:
@stof took a different approach, now only 1 log is produced

nicolas-grekas:
Thank you @ro0NL.

PR #52079, opened 2023-10-16
[HttpKernel] Add parameters `kernel.runtime_mode` and `kernel.runtime_mode.*`, all set from env var `APP_RUNTIME_MODE`

### LINKED ISSUE #51340: Function to check if the kernel is loaded via the console or a web request.

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #51340
| License       | MIT
| Doc PR        | TODO

Alternative to #51408. I think this approach is simpler and more powerful.

Here, we ensure that the kernel always provides a new `kernel.runtime_mode` parameter. This parameter is an array derived by default from the `APP_RUNTIME_MODE` env var, using the `query_string` processor.

This also creates 3 new parameters that should be the most common: `kernel.runtime_mode.web`, `kernel.runtime_mode.cli`, and `kernel.runtime_mode.worker`.

A long-running server would typically set `APP_RUNTIME_MODE` to `web=1&worker=1` or `web=0&worker=1` when appropriate (eg https://github.com/php-runtime/frankenphp-symfony/ should do so when `FRANKENPHP_WORKER` is set.)

I screened the codebase and updated them all except cache pools (where the SAPI is used to enable/disable locking) and error renderers (where the SAPI is used to turn html-rendering on/off.) These require more work that could be done later on. There are a few other remaining usages of `PHP_SAPI` but these look not appropriate for the new flag.


src/Symfony/Component/DependencyInjection/Dumper/PhpDumper.php
```diff
@@ -388,6 +388,7 @@ class %s extends {$options['class']}
     'container.build_hash' => '$hash',
     'container.build_id' => '$id',
     'container.build_time' => $time,
+    'container.runtime_mode' => \PHP_SAPI === 'cli' || \PHP_SAPI === 'phpdbg' ? 'cli' : 'web',
```
dunglas: Could we only check if the SAPI name starts with `cli` but allow suffixes? This will help for alternative CLI implementations. Shouldn't also check for the `embed` SAPI (it behaves more or less like CLI).

src/Symfony/Component/HttpKernel/DataCollector/DumpDataCollector.php
```diff
@@ -42,15 +42,17 @@ class DumpDataCollector extends DataCollector implements DataDumperInterface
...
-    public function __construct(Stopwatch $stopwatch = null, string|FileLinkFormatter $fileLinkFormat = null, string $charset = null, RequestStack $requestStack = null, DataDumperInterface|Connection $dumper = null)
+    public function __construct(Stopwatch $stopwatch = null, string|FileLinkFormatter $fileLinkFormat = null, string $charset = null, RequestStack $requestStack = null, DataDumperInterface|Connection $dumper = null, bool $webMode = null)
     {
         $fileLinkFormat = $fileLinkFormat ?: \ini_get('xdebug.file_link_format') ?: get_cfg_var('xdebug.file_link_format');
         $this->stopwatch = $stopwatch;
         $this->fileLinkFormat = $fileLinkFormat instanceof FileLinkFormatter && false === $fileLinkFormat->format('', 0) ? false : $fileLinkFormat;
         $this->charset = $charset ?: \ini_get('php.output_encoding') ?: \ini_get('default_charset') ?: 'UTF-8';
         $this->requestStack = $requestStack;
         $this->dumper = $dumper;
+        $this->webMode = $webMode ?? !\in_array(\PHP_SAPI, ['cli', 'phpdbg'], true);
```
dunglas: This weird to have `web` in the name while it can be CLI? Maybe `phpMode` or `runtimeType`?

nicolas-grekas: This is a boolean: true in web mode, false in cli mode.

src/Symfony/Component/DependencyInjection/Dumper/PhpDumper.php
nicolas-grekas: Updated to add "embed" almost everywhere.
Bad idea for the "cli" prefix: "cli-server" starts with "cli" but serves HTML (php -S)

nicolas-grekas:
I just renamed `kernel.is_web_runtime_mode` to `kernel.web_mode`.

stof:
shouldn't the error rendering depend on web_mode ? For a long-running server, we would want to provide the web error rendering.

src/Symfony/Component/ErrorHandler/ErrorHandler.php
```diff
@@ -640,7 +640,7 @@ public static function handleFatalError(array $error = null): void
      */
     private function renderException(\Throwable $exception): void
     {
-        $renderer = \in_array(\PHP_SAPI, ['cli', 'phpdbg'], true) ? new CliErrorRenderer() : new HtmlErrorRenderer($this->debug);
+        $renderer = \in_array(\PHP_SAPI, ['cli', 'phpdbg', 'embed'], true) ? new CliErrorRenderer() : new HtmlErrorRenderer($this->debug);
```
stof, line 643: this change looks weird to me. In the case of frankenphp, it behaves like a web server, not as a CLI (or would this report as a different value than `embed` ?)

nicolas-grekas:
> shouldn't the error rendering depend on web_mode

yes, but it's too much work for me for now, could be handled later on (same for cache locking, see PR description)

src/Symfony/Component/ErrorHandler/ErrorHandler.php
dunglas: FrankenPHP doesn't use the `embed` SAPI in web mode, and in CLI mode... it uses the embed SAPI but changes its name to  `cli`, which is a hack... but needed for interoperability with existing code like this :-D

`embed` is documented and usually used as a very simple SAPI to execute PHP scripts in C/C++ programs. AFAIK it is never used in web context. IMHO defaulting to `cli` for it is safer (and can be override). Anyway, it's an edge case, I'm fine with not supporting it at all too.

nicolas-grekas: We already have a few spots where we consider embed as cli. Note that I didn't add embed everywhere cli is listed, especially when opcache is considered. The reason is that I don't have enough background on the topic to decide if we should change.

dunglas: opcache doesn't work with embed: https://github.com/php/php-src/blob/221b4fe246e62c5d59f617ee4cbe6fd4c614fb5a/ext/opcache/ZendAccelerator.c#L2800-L2814

nicolas-grekas:
PR updated. No special service anymore, only parameters. Here is the new description:

Here, we ensure that the kernel always provides a new `kernel.runtime_mode` parameter. This parameter is an array derived by default from the `APP_RUNTIME_MODE` env var, using the `query_string` processor.

This also creates 3 new parameters that should be the most common: `kernel.runtime_mode.web`, `kernel.runtime_mode.cli`, and `kernel.runtime_mode.worker`.

A long-running server would typically set `APP_RUNTIME_MODE` to `web=1&worker=1` or `web=0&worker=1` when appropriate (eg https://github.com/php-runtime/frankenphp-symfony/ should do so when `FRANKENPHP_WORKER` is set.)

src/Symfony/Component/ErrorHandler/ErrorHandler.php
nicolas-grekas: > opcache doesn't work with embed

thanks for the hint, I updated a few more lines to account for that

src/Symfony/Component/Cache/Adapter/AbstractAdapter.php
```diff
@@ -97,7 +97,7 @@ public static function createSystemCache(string $namespace, int $defaultLifetime
             return $opcache;
         }
 
-        if (\in_array(\PHP_SAPI, ['cli', 'phpdbg'], true) && !filter_var(\ini_get('apc.enable_cli'), \FILTER_VALIDATE_BOOL)) {
+        if ('cli' === \PHP_SAPI && !filter_var(\ini_get('apc.enable_cli'), \FILTER_VALIDATE_BOOL)) {
```
nicolas-grekas, line 100: aligning with other checks we have in place for APCu (not need to enable it for phpdbg/embed)

src/Symfony/Component/DependencyInjection/Dumper/PhpDumper.php
```diff
@@ -1591,7 +1592,7 @@ private function addDefaultParametersMethod(): string
             $export = $this->exportParameters([$value], '', 12, $hasEnum);
             $export = explode('0 => ', substr(rtrim($export, " ]\n"), 2, -1), 2);
 
-            if ($hasEnum || preg_match("/\\\$container->(?:getEnv\('(?:[-.\w\\\\]*+:)*+\w++'\)|targetDir\.'')/", $export[1])) {
+            if ($hasEnum || preg_match("/\\\$container->(?:getEnv\('(?:[-.\w\\\\]*+:)*+\w*+'\)|targetDir\.'')/", $export[1])) {
```
nicolas-grekas, line 1595: this change allows to derivate dynamic parameters from other ones, eg `%env(key:foo:default:bar:)%` extracts key `foo` from parameter `bar` (which is supposed to be an array here)

src/Symfony/Component/Runtime/Runner/Symfony/HttpKernelRunner.php
```diff
@@ -31,7 +33,21 @@ public function __construct(
...
+
+        if (Kernel::VERSION_ID >= 60400) {
+            $response->send(false);
+
+            if (\function_exists('fastcgi_finish_request') && !$this->debug) {
+                fastcgi_finish_request();
+            } elseif (\function_exists('litespeed_finish_request') && !$this->debug) {
+                litespeed_finish_request();
+            } else {
+                Response::closeOutputBuffers(0, true);
```
nicolas-grekas, line 45: Unlike the logic in `$response->send()`, we call this unconditionally. The reason is that here, in the runtime, we know we want to flush whatever the SAPI.

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

Issue #37985, opened 2020-08-29
[messenger] Failed message never delivered to failed queue after retry


jmhoueg:
**Symfony version(s) affected**: 5.1.3

**Description**  
Using messenger with RabbitMQ, after a message fails and is retried 2 times, the message should be sent to the failed queue but it doesn't happen for me. This is my messenger.yaml:
```yaml
framework:
    messenger:
        # Uncomment this (and the failed transport below) to send failed messages to this transport for later handling.
        failure_transport: amqp_create_order_error
        transports:
            # https://symfony.com/doc/current/messenger.html#transport-configuration
            amqp_create_order:
                dsn: '%env(MESSENGER_TRANSPORT_DSN)%'
                options:
                    exchange:
                        name: 'ms.customer-order'
                        type: topic
                    queues:
                        ms.customer-order;onestock_create_order:
                            binding_keys:
                                - 'create_order'
                retry_strategy:
                    max_retries: 2
            amqp_create_order_retry_30000:
                dsn: '%env(MESSENGER_TRANSPORT_DSN)%'
                options:
                    exchange:
                        name: 'ms.customer-order'
                        type: topic
                    queues:
                        ms.customer-order;create_order_retry_30000:
                            binding_keys:
                                - 'create_order_retry_30000'
            amqp_create_order_error:
                dsn: '%env(MESSENGER_TRANSPORT_DSN)%'
                options:
                    exchange:
                        name: 'ms.customer-order'
                        type: topic
                    queues:
                        ms.customer-order;create_order_error:
                            binding_keys:
                                - 'create_order_error'
            # failed: 'doctrine://default?queue_name=failed'
            # sync: 'sync://'
        routing:
            # Route your messages to the transports
            'App\Message\CreateOneStockOrderMessage': amqp_create_order
```

I read https://github.com/symfony/symfony/issues/31455 and according to what was said it should work without trouble. What can i do please?

**How to reproduce**  
Use my messenger.yaml, create a handler which fails and see what happens

carsonbot:
Hey, thanks for your report!
There has not been a lot of activity here for a while. Is this bug still relevant? Have you managed to find a workaround?

carsonbot:
Hello? This issue is about to be closed if nobody replies.

carsonbot:
Hey,

I didn't hear anything so I'm going to close it. Feel free to comment if this is still relevant, I can always reopen!

yarovskiy:
Hello! I have the same trouble(
[...]

PR #52083, opened 2023-10-16
[Messenger][Amqp] Don't use retry routing key when sending to failure transport

fabcdl:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | #37985
| License       | MIT

When a message with a `RedeliveryStamp` is sent to a failure transport via AMQP, the `AmqpSender` uses the original queue name as the routing key (intended for retry routing). This causes the message to be published with the wrong routing key, so with a `direct` exchange it never reaches the failure queue.

The fix: skip the retry routing key override when the envelope carries a `SentToFailureTransportStamp`. This lets the failure transport's own `default_publish_routing_key` be used naturally, without any new API surface.

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Amqp/Tests/Transport/AmqpSenderTest.php
+++ b/src/Symfony/Component/Messenger/Bridge/Amqp/Tests/Transport/AmqpSenderTest.php
@@ -13,11 +13,14 @@ namespace Symfony\Component\Messenger\Bridge\Amqp\Tests\Transport;
 
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Messenger\Bridge\Amqp\Tests\Fixtures\DummyMessage;
+use Symfony\Component\Messenger\Bridge\Amqp\Transport\AmqpReceivedStamp;
 use Symfony\Component\Messenger\Bridge\Amqp\Transport\AmqpSender;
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/Messenger/Bridge/Amqp/Transport/Connection.php
```diff
@@ -295,7 +295,7 @@ public function publish(string $body, array $headers = [], int $delayInMs = 0, A
         $this->publishOnExchange(
             $this->exchange(),
             $body,
-            $this->getRoutingKeyForMessage($amqpStamp),
+            $forceDefaultRoutingKey ? $this->getDefaultPublishRoutingKey() : $this->getRoutingKeyForMessage($amqpStamp),
```
nicolas-grekas: is there a way without adding the new argument, by using/pushing some state in AmqpStamp?

fabcdl: Thanks for your feedback. I followed your recommendations and used AmqpStamp to transport the information.

nicolas-grekas (approved):
I updated the implementation to make it less invasive

nicolas-grekas:
Thank you @fabcdl.

PR #52087, opened 2023-10-16
[Scheduler] Add `FailureEvent`

alli83:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT

Following PR #51805, It would be interesting to add a `failureEvent` allowing you, for instance, to remove the recurring message, depending on the error caught.

### DIFF
--- a/src/Symfony/Component/Scheduler/Scheduler.php
+++ b/src/Symfony/Component/Scheduler/Scheduler.php
@@ -14,6 +14,7 @@ namespace Symfony\Component\Scheduler;
 use Psr\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\Clock\Clock;
 use Symfony\Component\Clock\ClockInterface;
+use Symfony\Component\Scheduler\Event\FailureEvent;
 use Symfony\Component\Scheduler\Event\PostRunEvent;
 use Symfony\Component\Scheduler\Event\PreRunEvent;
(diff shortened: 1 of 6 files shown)


src/Symfony/Component/Scheduler/Scheduler.php
```diff
@@ -81,10 +82,14 @@ public function run(array $options = []): void
...
-                    $this->handlers[$message::class]($message);
-                    $ran = true;
+                    try {
+                        $this->handlers[$message::class]($message);
+                        $ran = true;
 
-                    $this->dispatcher->dispatch(new PostRunEvent($generator->getSchedule(), $context, $message));
+                        $this->dispatcher->dispatch(new PostRunEvent($generator->getSchedule(), $context, $message));
+                    } catch (\Throwable $error) {
+                        $this->dispatcher->dispatch(new FailureEvent($generator->getSchedule(), $context, $message, $error));
```
fabpot: We need to rethrow the exception, right?
And maybe the FailureEvent can have a way to ignore the error?

alli83: yes indeed. Maybe I could add a `shouldIgnore` like the one we already have for the `WorkerMessageReceivedEvent`

fabpot (approved):

fabpot:
Thank you @alli83.

PR #52090, opened 2023-10-17
[Dotenv] Update README with variable override note

### LINKED ISSUE #52073: env variable is being overriden even when override is disabled

6.3.x

### Description

Faced issue that env. variables are being overridden by sequential files even if `$overrideExistingVars=false`

### How to reproduce

```
<?php

use Symfony\Component\Dotenv\Dotenv;

require_once './vendor/autoload.php';

$projectDir = __DIR__;

$files = array_reverse(array_filter([
    $projectDir.'/.env',
    $projectDir.'/.env.local'
], 'file_exists'));

dump($files); // order of loading .env.local -> .env
(new Dotenv())->load(...$files); // using doLoad with  $overrideExistingVars = false
dd($_SERVER['EMAIL_SENDER_MAIL'], $_ENV['EMAIL_SENDER_MAIL']);
// ER: `foobar` because .env.local has higher priority
// AR: <empty>
```

```
# .env
EMAIL_SENDER_MAIL=
```

```
# .env.local
EMAIL_SENDER_MAIL=foobar
```

### Possible Solution

Related task https://github.com/symfony/symfony/commit/08291b1665a96c77a6516ac52c4a100135e7061f

if stmt should be like:

```php
            if (
                !$overrideExistingVars // checking if opt-in into override
                && (isset($loadedVars[$name]) || isset($_ENV[$name])) // checking that variable exists
            ) {
                continue;
            }
```

### Additional Context

_No response_

oleg-andreyev:
In the Dotenv README.md, a note has been added to clarify how the `$overrideExistingVars` flag behaves. This flag works only with existing environment variables (defined before application), and not with `.env` file variables. This change aims to prevent potential misunderstandings or misuse of the flag in the future.

| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #52073
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Dotenv/README.md
+++ b/src/Symfony/Component/Dotenv/README.md
@@ -27,6 +27,10 @@ $dotenv->overload(__DIR__.'/.env');
 $dotenv->loadEnv(__DIR__.'/.env');
 ```
 
+Note
+---
+Keep in mind that `$overrideExistingVars` work only with **existing** environment variables (defined before application) and not with `.env` file variables.
+
 Resources
 ---------
 


src/Symfony/Component/Dotenv/README.md
```diff
@@ -27,6 +27,10 @@ $dotenv->overload(__DIR__.'/.env');
 $dotenv->loadEnv(__DIR__.'/.env');
 ```
 
+Note
+---
+Keep in mind that `$overrideExistingVars` work only with **existing** environment variables (defined before application) and not with `.env` file variables.
```
OskarStark, line 32:
```suggestion
----

Keep in mind that `$overrideExistingVars` work only with **existing**
environment variables (defined before application) and not with `.env`
file variables.
```

oleg-andreyev: @OskarStark don't see the difference except multiline, right?

Nyholm: Yes, he just added some more line breaks so it comply with the coding standard. 

(empty line after heading + short lines)

Nyholm:
Readmes generally dont include information like this. Readmes has a short description and then usage examples. 

What do you think about adding some lines in the method doc instead? And maybe one more usage example?

stof:
this should indeed go in the phpdoc and maybe in the symfony documentation rather than in the readme.

PR #52095, opened 2023-10-17
[Notifier][Sendinblue] Handle error responses without a message key

stof:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no
| Tickets       | 
| License       | MIT

During their outage yesterday where the endpoint returned errors with status code 500, there was no `message` key in the json response.
This avoids a notice when the key does not exist. Instead, it uses the full response content, as done when the decoding fails.

### DIFF
--- a/src/Symfony/Component/Mailer/Bridge/Sendinblue/Transport/SendinblueApiTransport.php
+++ b/src/Symfony/Component/Mailer/Bridge/Sendinblue/Transport/SendinblueApiTransport.php
@@ -65,7 +65,7 @@ final class SendinblueApiTransport extends AbstractApiTransport
         }
 
         if (201 !== $statusCode) {
-            throw new HttpTransportException('Unable to send an email: '.$result['message'].sprintf(' (code %d).', $statusCode), $response);
+            throw new HttpTransportException('Unable to send an email: '.($result['message'] ?? $response->getContent(false)).sprintf(' (code %d).', $statusCode), $response);
         }
 
         $sentMessage->setMessageId($result['messageId']);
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/Mailer/Bridge/Sendinblue/Transport/SendinblueApiTransport.php
```diff
@@ -65,7 +65,7 @@ protected function doSendApi(SentMessage $sentMessage, Email $email, Envelope $e
         }
 
         if (201 !== $statusCode) {
-            throw new HttpTransportException('Unable to send an email: '.$result['message'].sprintf(' (code %d).', $statusCode), $response);
+            throw new HttpTransportException('Unable to send an email: '.($error['message'] ?? $response->getContent(false)).sprintf(' (code %d).', $statusCode), $response);
```
OskarStark, line 68: Variable `$error` seems not to be defined

OskarStark:
```suggestion
            throw new HttpTransportException('Unable to send an email: '.($result['message'] ?? $response->getContent(false)).sprintf(' (code %d).', $statusCode), $response);
```

OskarStark:
As Brevo is the 1:1 replacement for Sendinblue after their rebranding, can you please also update Mailer/Brevo and Notifier/Brevo bridges? Thanks

stof:
@OskarStark done in https://github.com/symfony/symfony/pull/52096 as the Brevo transports only exist in the 6.4 branch, not in 5.4

OskarStark (approved):

nicolas-grekas:
Thank you @stof.

PR #52096, opened 2023-10-17
[Notifier] [Brevo] Handle error responses without a message key

stof:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       |
| License       | MIT

During their outage yesterday where the endpoint returned errors with status code 500, there was no message key in the json response.
This avoids a notice when the key does not exist. Instead, it uses the full response content, as done when the decoding fails.

This is the same change than https://github.com/symfony/symfony/pull/52095 but for the new Brevo bridge


OskarStark (approved):

nicolas-grekas:
Thank you @stof.

PR #52099, opened 2023-10-17
[Security] Allow wildcards as placeholders in the role hierarchy

squrious:
| Q             | A
| ------------- | ---
| Branch?       | 8.2
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | 
| License       | MIT
| Doc PR        | symfony/symfony-docs#19079

A key of `role_hierarchy` can now carry a `*` wildcard, so a whole family of roles inherits from the same parents without listing them one by one.

```yaml
role_hierarchy:
    ROLE_USER_*:    ROLE_USER

    ROLE_SHOP_*:    ROLE_USER_SHOP
    ROLE_USER_SHOP: ROLE_SHOP_VIEW

    ROLE_BLOG_*:    ROLE_USER_BLOG
    ROLE_USER_BLOG: ROLE_BLOG_VIEW
```

`ROLE_SHOP_BUY` then reaches `ROLE_USER_SHOP`, `ROLE_SHOP_VIEW` and `ROLE_USER`, without any of them being declared for that role.

A wildcard only counts when it sits between underscores or ends the role, so `ROLE_FOO*`, `ROLE_*FOO`, `ROLE_FOO_*BAR` and `ROLE_FOO*_BAR` are read as ordinary role names and existing hierarchies are untouched. Placeholders resolve recursively, each one at most once, so two placeholders pointing at each other terminate.

A placeholder is a pattern, not a role: it is never returned by `getReachableRoleNames()` nor by `getParentRoleNames()`.

Applications that declare no placeholder keep the previous code path, so they pay nothing for the feature.

Documentation should cover: where a wildcard is recognised and where it is not, that placeholders resolve recursively, and that a placeholder is never itself a role.

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/Debug/DebugRoleHierarchy.php
+++ b/src/Symfony/Bundle/SecurityBundle/Debug/DebugRoleHierarchy.php
@@ -1,0 +1,105 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\SecurityBundle\Debug;
+
+use Symfony\Component\Security\Core\Role\RoleHierarchy;
+
+/**
+ * Extended Role Hierarchy to access inner configuration data.
+ *
+ * @author Nicolas Rigaud <squrious@protonmail.com>
+ *
+ * @internal
+ */
+final class DebugRoleHierarchy extends RoleHierarchy
+{
+    private readonly array $debugHierarchy;
+
+    public function __construct(array $hierarchy)
+    {
+        $this->debugHierarchy = $hierarchy;
+
+        parent::__construct($hierarchy);
+    }
+
+    /**
+     * Get the hierarchy tree.
+     *
+     * Example output:
+     *
+     *     [
+     *       'ROLE_A' => [
+     *          'ROLE_B' => [],
+     *          'ROLE_C' => [
+     *             'ROLE_D' => []
+     *          ]
+     *       ],
+     *       'ROLE_C' => [
+     *          'ROLE_D' => []
+     *       ]
+     *     ]
+     *
+     * @param string[] $roles Optionally restrict the tree to these roles
+     *
+     * @return array<string,array<string,array>>
+     */
+    public function getHierarchy(array $roles = []): array
+    {
+        $hierarchy = [];
+
+        foreach ($roles ?: array_keys($this->debugHierarchy) as $role) {
+            $hierarchy += $this->buildHierarchy([$role]);
+        }
+
+        return $hierarchy;
+    }
+
+    /**
+     * Get the computed role map.
+     *
+     * @return array<string,string[]>
+     */
+    public function getMap(): array
+    {
+        return $this->map;
+    }
+
+    /**
+     * Return whether a given role is processed as a placeholder.
+     */
+    public function isPlaceholder(string $role): bool
+    {
+        return \in_array($role, array_keys($this->rolePlaceholdersPatterns));
+    }
+
+    private function buildHierarchy(array $roles, array &$visited = []): array
+    {
+        $tree = [];
+        foreach ($roles as $role) {
+            $visited[] = $role;
+
+            $tree[$role] = [];
+
+            // Get placeholders matches
+            $placeholders = array_diff($this->getMatchingPlaceholders([$role]), $visited) ?? [];
+            array_push($visited, ...$placeholders);
+            $tree[$role] += $this->buildHierarchy($placeholders, $visited);
+
+            // Get regular inherited roles
+            $inherited = array_diff($this->debugHierarchy[$role] ?? [], $visited);
+            array_push($visited, ...$inherited);
+            $tree[$role] += $this->buildHierarchy($inherited, $visited);
+        }
+
+        return $tree;
+    }
+}
--- a/src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
+++ b/src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
@@ -18,6 +18,13 @@ namespace Symfony\Component\Security\Core\Role;
  */
 class RoleHierarchy implements RoleHierarchyInterface
 {
+    /**
+     * Map role placeholders with their regex pattern.
+     *
+     * @var array<string,string>
+     */
+    private array $rolePlaceholdersPatterns;
+
     /** @var array<string, list<string>> */
     protected array $map;
 
@@ -35,24 +42,14 @@ class RoleHierarchy implements RoleHierarchyInterface
 
     public function getReachableRoleNames(array $roles): array
     {
-        $reachableRoles = $roles;
-
-        foreach ($roles as $role) {
-            if (!isset($this->map[$role])) {
-                continue;
-            }
-
-            foreach ($this->map[$role] as $r) {
-                $reachableRoles[] = $r;
-            }
-        }
-
-        return array_values(array_unique($reachableRoles));
+        return \array_values(\array_unique($this->resolveReachableRoleNames($roles)));
     }
 
     protected function buildRoleMap(): void
     {
         $this->map = [];
+        $this->rolePlaceholdersPatterns = [];
+
         foreach ($this->hierarchy as $main => $roles) {
             $this->map[$main] = $roles;
             $visited = [];
@@ -74,6 +71,49 @@ class RoleHierarchy implements RoleHierarchyInterface
             }
 
             $this->map[$main] = array_unique($this->map[$main]);
+
+            if (str_contains($main, '*')) {
+                $this->rolePlaceholdersPatterns[$main] = sprintf('/%s/', strtr($main, ['*' => '[^\*]+']));
+            }
         }
+    }
+
+    private function resolveReachableRoleNames(array $roles, array &$visitedPlaceholders = []): array
+    {
+        $reachableRoles = $roles;
+
+        foreach ($roles as $role) {
+            if (!isset($this->map[$role])) {
+                continue;
+            }
+
+            foreach ($this->map[$role] as $r) {
+                $reachableRoles[] = $r;
+            }
+        }
+
+        $placeholderRoles = array_diff($this->getMatchingPlaceholders($reachableRoles), $visitedPlaceholders);
+        if (!empty($placeholderRoles)) {
+            array_push($visitedPlaceholders, ...$placeholderRoles);
+            $resolvedPlaceholderRoles = $this->resolveReachableRoleNames($placeholderRoles, $visitedPlaceholders);
+            foreach (array_diff($resolvedPlaceholderRoles, $placeholderRoles) as $r) {
+                $reachableRoles[] = $r;
+            }
+        }
+
+        return $reachableRoles;
+    }
+
+    private function getMatchingPlaceholders(array $roles): array
+    {
+        $resolved = [];
+
+        foreach ($this->rolePlaceholdersPatterns as $placeholder => $pattern) {
+            if (!\in_array($placeholder, $resolved) && \count(preg_grep($pattern, $roles) ?? null)) {
+                $resolved[] = $placeholder;
+            }
+        }
+
+        return $resolved;
     }
 }
--- a/src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
+++ b/src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
@@ -131,6 +131,6 @@ class RoleHierarchy implements RoleHierarchyInterface
         /** @var int $count */
         $placeholderPattern = preg_replace(pattern: '/(?<=_)\\\\\*(?=_|$)/', replacement: '[^\*]*', subject: preg_quote($role), count: $count);
 
-        return ($count > 0) ? sprintf('/%s/', $placeholderPattern) : false;
+        return ($count > 0) ? sprintf('/^%s$/', $placeholderPattern) : false;
     }
 }
--- a/src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
+++ b/src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
@@ -23,7 +23,7 @@ class RoleHierarchy implements RoleHierarchyInterface
      *
      * @var array<string,string>
      */
-    private array $rolePlaceholdersPatterns;
+    protected array $rolePlaceholdersPatterns;
 
     /** @var array<string, list<string>> */
     protected array $map;
@@ -72,7 +72,7 @@ class RoleHierarchy implements RoleHierarchyInterface
 
             $this->map[$main] = array_unique($this->map[$main]);
 
-            if (str_contains($main, '*') && false !== ($pattern = $this->getPlaceholderPattern($main))) {
+            if (str_contains($main, '*') && null !== ($pattern = $this->getPlaceholderPattern($main))) {
                 $this->rolePlaceholdersPatterns[$main] = $pattern;
             }
         }
@@ -104,12 +104,12 @@ class RoleHierarchy implements RoleHierarchyInterface
         return $reachableRoles;
     }
 
-    private function getMatchingPlaceholders(array $roles): array
+    protected function getMatchingPlaceholders(array $roles): array
     {
         $resolved = [];
 
         foreach ($this->rolePlaceholdersPatterns as $placeholder => $pattern) {
-            if (!\in_array($placeholder, $resolved) && \count(preg_grep($pattern, $roles) ?? null)) {
+            if (!\in_array($placeholder, $resolved) && \count(preg_grep($pattern, $roles) ?: [])) {
                 $resolved[] = $placeholder;
             }
         }
@@ -119,18 +119,18 @@ class RoleHierarchy implements RoleHierarchyInterface
 
     /**
      * Build the regex pattern for the given role:
-     *   - Replace valid wildcards with a non-wildcard matching pattern and
+     *   - Replace valid wildcards with a non-wildcard matching pattern.
      *   - Escape reserved regex characters.
      *
      * A valid wildcard is a * prefixed with _ and immediately followed by _ or EOL.
      *
-     * @return string|false The regex pattern, or false if there is no valid wildcard in the role
+     * @return string|null The regex pattern, or null if there is no valid wildcard in the role
      */
-    private function getPlaceholderPattern(string $role): string|false
+    private function getPlaceholderPattern(string $role): ?string
     {
         /** @var int $count */
-        $placeholderPattern = preg_replace(pattern: '/(?<=_)\\\\\*(?=_|$)/', replacement: '[^\*]*', subject: preg_quote($role), count: $count);
+        $placeholderPattern = preg_replace(pattern: '/(?<=_)\\\\\*(?=_|$)/', replacement: '.*', subject: preg_quote($role), count: $count);
 
-        return ($count > 0) ? sprintf('/^%s$/', $placeholderPattern) : false;
+        return ($count > 0) ? sprintf('/^%s$/', $placeholderPattern) : null;
     }
 }
--- a/src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
+++ b/src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
@@ -42,7 +42,7 @@ class RoleHierarchy implements RoleHierarchyInterface
 
     public function getReachableRoleNames(array $roles): array
     {
-        return \array_values(\array_unique($this->resolveReachableRoleNames($roles)));
+        return array_values(array_unique($this->resolveReachableRoleNames($roles)));
     }
 
     protected function buildRoleMap(): void
--- a/src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
+++ b/src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
@@ -72,8 +72,8 @@ class RoleHierarchy implements RoleHierarchyInterface
 
             $this->map[$main] = array_unique($this->map[$main]);
 
-            if (str_contains($main, '*')) {
-                $this->rolePlaceholdersPatterns[$main] = sprintf('/%s/', strtr($main, ['*' => '[^\*]+']));
+            if (str_contains($main, '*') && false !== ($pattern = $this->getPlaceholderPattern($main))) {
+                $this->rolePlaceholdersPatterns[$main] = $pattern;
             }
         }
     }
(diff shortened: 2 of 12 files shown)


OskarStark:
I really like this feature, but it adds a lot of complexity and I am not sure it's worth.
If this feature would get accepted, we would need a DebugRolesCommand IMHO 🤷‍♂️

Neirda24:
Nice ! I would suggest to not match on `*` only. But either prefix or suffixed with `_`. So `*` act as `[a-zA-Z]+`

src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
```diff
@@ -76,6 +74,49 @@ protected function buildRoleMap()
             }
 
             $this->map[$main] = array_unique($this->map[$main]);
+
+            if (str_contains($main, '*')) {
+                $this->rolePlaceholdersPatterns[$main] = sprintf('/%s/', strtr($main, ['*' => '[^\*]+']));
```
stof: if you want to generate a regexp, you need to escape any other char that could have a special meaning.

src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
```diff
@@ -76,6 +74,49 @@ protected function buildRoleMap()
...
+        $placeholderRoles = array_diff($this->getMatchingPlaceholders($reachableRoles), $visitedPlaceholders);
+        if (!empty($placeholderRoles)) {
+            array_push($visitedPlaceholders, ...$placeholderRoles);
+            $resolvedPlaceholderRoles = $this->resolveReachableRoleNames($placeholderRoles, $visitedPlaceholders);
+            foreach (array_diff($resolvedPlaceholderRoles, $placeholderRoles) as $r) {
+                $reachableRoles[] = $r;
+            }
+        }
+
+        return array_values(array_unique($reachableRoles));
```
stof: Do we really need to do the uniqueness filtering in the method that gets called recursively ? couldn't we do it only once at the end of the recursion ?

squrious:
> I really like this feature, but it adds a lot of complexity and I am not sure it's worth. If this feature would get accepted, we would need a DebugRolesCommand IMHO 🤷‍♂️

I agree that adds complexity in the hierarchy config, but since it's optional we keep simple what should remain simple, no? The drawback I see is the config could be less meaningful as all roles involved in the hierarchy are not necessary listed, but here again, we let the choice to the developer ^^

100% OK for a debug roles command, maybe something like this?
```
$ bin/console debug:security:role-hierarchy

Role hierarchy configuration: 

ROLE_SHOP_* 
  => ROLE_USER_SHOP 
    => ROLE_USER_*
      => ROLE_USER
ROLE_BLOG_*
  => ROLE_USER_BLOG 
    => ROLE_USER_*
      => ROLE_USER
ROLE_USER_*
  => ROLE_USER

$ bin/console debug:security:role-hierarchy ROLE_SHOP_VIEW ROLE_BLOG_POST

Effective roles for ROLE_SHOP_VIEW, ROLE_BLOG_POST:
 - ROLE_SHOP_VIEW
 - ROLE_BLOG_POST
 - ROLE_USER_SHOP
 - ROLE_BLOG_USER
 - ROLE_USER

$ bin/console debug:security:role-hierarchy ROLE_SHOP_VIEW --tree
   
Effective roles for ROLE_SHOP_VIEW:
 - ROLE_SHOP_VIEW
 - ROLE_USER_SHOP
   - ROLE_USER
```

src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
squrious: Thank you I totally missed that!

src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
squrious: No you're right, I'll fix that in the next commit, thanks

squrious:
> Nice ! I would suggest to not match on `*` only. But either prefix or suffixed with `_`. So `*` act as `[a-zA-Z]`

@Neirda24 Thanks! Which use cases would you like to avoid with this restriction? `ROLE_*_CREATE` should not match roles like `ROLE_BLOG_POST_CREATE`? Or you think about the placeholder syntax itself, so `ROLE*` is not considered as a valid placeholder?

Neirda24:
> > Nice ! I would suggest to not match on `*` only. But either prefix or suffixed with `_`. So `*` act as `[a-zA-Z]`
> 
> @Neirda24 Thanks! Which use cases would you like to avoid with this restriction? `ROLE_*_CREATE` should not match roles like `ROLE_BLOG_POST_CREATE`? Or you think about the placeholder syntax itself, so `ROLE*` is not considered as a valid placeholder?

The latter. About your first suggestion you are right it should be matcheable. Maybe with `ROLE_**_CREATE` to allow any number of words in between ?

squrious:
> > > Nice ! I would suggest to not match on `*` only. But either prefix or suffixed with `_`. So `*` act as `[a-zA-Z]`
> > 
> > 
> > @Neirda24 Thanks! Which use cases would you like to avoid with this restriction? `ROLE_*_CREATE` should not match roles like `ROLE_BLOG_POST_CREATE`? Or you think about the placeholder syntax itself, so `ROLE*` is not considered as a valid placeholder?
> 
> The latter. About your first suggestion you are right it should be matcheable. Maybe with `ROLE_**_CREATE` to allow any number of words in between ?

Got it, I'll add the restriction for placeholder syntax.

For `**`, I'm not sure, I mean it will force the use of `_` as a word separator, and thus introduce the concept of words in the role syntax, but only when using placeholders. If find it a little bit confusing, while `*` could just mean _match anything here_.

src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
squrious: I force pushed this one to let the next commit about regex cleaner

src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
squrious: Fixed with `preg_quote` [here](https://github.com/symfony/symfony/pull/52099/files#diff-cffbd9a47ea9c09114ad41f4138ca3d8d56c0d8a7c6588c428b8d195dc5e3b29R135)

squrious:
@Neirda24 I updated the placeholder pattern matching in [this commit](https://github.com/symfony/symfony/pull/52099/commits/0240d67cb0a028f8041b375c41202aeaa44066f4), looks fine to you?

smnandre:
I'm not sure your example really show the (real) advantage of this feature.

Current configuration:

```
role_hierarchy:
  ROLE_USER_SHOP:         ROLE_USER
  ROLE_USER_BLOG:         ROLE_USER
  
  ROLE_SHOP_VIEW:         ROLE_USER_SHOP
  ROLE_SHOP_BUY:          ROLE_USER_SHOP
  ROLE_SHOP_ADD_TO_CART:  ROLE_USER_SHOP
  
  ROLE_BLOG_VIEW:         ROLE_USER_BLOG 
  ROLE_BLOG_POST:         ROLE_USER_BLOG
  ROLE_BLOG_REPORT:       ROLE_USER_BLOG
  ROLE_BLOG_COMMENT:      ROLE_USER_BLOG
```

You kinda uses your "roles" as "permission", so you could have written it like that no ?

```
role_hierarchy:
  ROLE_USER:                 [ROLE_USER_BLOG, ROLE_USER_SHOP]
  ROLE_USER_BLOG:     [ROLE_BLOG_VIEW, ROLE_BLOG_POST, ROLE_BLOG_REPORT, ROLE_BLOG_COMMENT]
  ROLE_USER_SHOP:     [ROLE_SHOP_VIEW, ROLE_SHOP_BUY, ROLE_SHOP_ADD_TO_CART]
```

OskarStark:
> 100% OK for a debug roles command, maybe something like this?

Yes 👍

squrious:
> You kinda uses your "roles" as "permission", so you could have written it like that no ?
> 
> ```
> role_hierarchy:
>   ROLE_USER:                 [ROLE_USER_BLOG, ROLE_USER_SHOP]
>   ROLE_USER_BLOG:     [ROLE_BLOG_VIEW, ROLE_BLOG_POST, ROLE_BLOG_REPORT, ROLE_BLOG_COMMENT]
>   ROLE_USER_SHOP:     [ROLE_SHOP_VIEW, ROLE_SHOP_BUY, ROLE_SHOP_ADD_TO_CART]
> ```

This configuration is not really equivalent. By giving the role `ROLE_USER` to someone you also give all inherited roles. In my example, you can give the single role `ROLE_BLOG_VIEW` to a user, and thus `ROLE_USER_BLOG` and `ROLE_USER` are inherited, but not the others (`ROLE_BLOG_POST`, `ROLE_SHOP_VIEW`, `ROLE_USER_SHOP`, etc.).

smnandre:
That is why i used the term "permission", that for me is quite different from "role"  (a role beeing itself an abstraction for a set of permissions) [^1]

In your example, today you'd have to give this user "ROLE_BLOG_VIEW" and let's say "ROLE_SHOP_VIEW", and then you inherit ROLE_BLOG_USER and ROLE_SHOP_USER, then ROLE_USER. 

How does the new placeholder system help/ease things ? 

Again, i **do** think it's a great idea, but i'd like to see a before/after example to illustrate.

In the one you gave, i don't see someone with ROLE_BLOG_REPORT but not ROLE_BLOG_VIEW.  

[^1]: even if Symfony sometimes uses both, like ROLE_ADMIN & ROLE_ALLOWED_TO_SWITCH.

src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
```diff
@@ -76,6 +74,66 @@ protected function buildRoleMap()
...
+     * A valid wildcard is a * prefixed with _ and immediately followed by _ or EOL.
+     *
+     * @return string|false The regex pattern, or false if there is no valid wildcard in the role
+     */
+    private function getPlaceholderPattern(string $role): string|false
+    {
+        /** @var int $count */
+        $placeholderPattern = preg_replace(pattern: '/(?<=_)\\\\\*(?=_|$)/', replacement: '[^\*]*', subject: preg_quote($role), count: $count);
+
+        return ($count > 0) ? sprintf('/%s/', $placeholderPattern) : false;
```
smnandre: ⚠️  Security risk there

```php
        return ($count > 0) ? sprintf('/%s/', $placeholderPattern) : false;
```

If you use only `/%s/`, it's as if you wrote "`*ROLE_*`" .. i think `/^%s/` would be better no ?

I suggest adding this test: 

````
$role = new RoleHierarchy(['ROLE_FOO_*' => ['ROLE_FOO_BAR']]));
$this->assertEquals(['ROLE_ROLE_FOO_BAR'], $role->getReachableRoleNames(['ROLE_ROLE_FOO_BAR'])); 
````

Comment (squrious) [reply src/Symfony/Component/Security/Core/Role/RoleHierarchy.php]: Good catch, thanks! I'd also add a trailing `$` to strictly match placeholders like `ROLE_FOO_*_BAR`

**Comment by squrious** [comment]
> That is why i used the term "permission", that for me is quite different from "role" (a role beeing itself an abstraction for a set of permissions) [1](#user-content-fn-1-01e7e834a34b75a3b9952d53661fe3f8)

I think it depends on the usage.

Without a role hierarchy, what's behind a role depends on the code that is actually protected with it. If the role is used only to protect a single action (ex: create a specific kind of entity in database), then it acts like a permission. If it's used to protect multiple actions in various places, then it "grants" multiple things and acts like a set of permissions.

With a role hierarchy, yes, roles used in keys are a set of permissions, as they inherit other ones. But I would think that roles used only in the values could then be either permissions or sets of permissions, with the same reasoning as above.

I see roles as a special type of argument for the authorization layer, that triggers the role voter and look for matching elements in the user's token. 

But that's my understanding, I may be wrong :grin:.

> In your example, today you'd have to give this user "ROLE_BLOG_VIEW" and let's say "ROLE_SHOP_VIEW", and then you inherit ROLE_BLOG_USER and ROLE_SHOP_USER, then ROLE_USER.
>
> How does the new placeholder system help/ease things ?

I think the main advantage here is to _map_ all potential roles that matches the pattern to a well-defined role name, that could be used to protect a domain section of the application easily. In the future, it will be easy to add/rename roles that will be covered with the pattern, without rewriting the hierarchy.

> Again, i **do** think it's a great idea, but i'd like to see a before/after example to illustrate.
>
> In the one you gave, i don't see someone with ROLE_BLOG_REPORT but not ROLE_BLOG_VIEW.

I agree that my first example is not sufficient to demonstrate the potential usages of the feature, and doesn't reflect a real configuration. Thanks for pointing this out!

Rewriting this example like this could feel more real?

Current configuration:
```yaml
role_hierarchy:
  ROLE_USER_SHOP:         [ROLE_USER, ROLE_SHOP_VIEW]
  ROLE_USER_BLOG:         [ROLE_USER, ROLE_BLOG_VIEW]
  
  ROLE_SHOP_BUY:          ROLE_USER_SHOP
  ROLE_SHOP_ADD_TO_CART:  ROLE_USER_SHOP
  
  ROLE_BLOG_POST:         ROLE_USER_BLOG
  ROLE_BLOG_REPORT:       ROLE_USER_BLOG
  ROLE_BLOG_COMMENT:      ROLE_USER_BLOG
```

With placeholder:
```yaml
role_hierarchy:
  ROLE_USER_*:    ROLE_USER
  
  ROLE_SHOP_*:    ROLE_USER_SHOP
  ROLE_USER_SHOP: ROLE_SHOP_VIEW
  
  ROLE_BLOG_*:    ROLE_USER_BLOG
  ROLE_USER_BLOG: ROLE_BLOG_VIEW
```


But now I find this use case too simple. It was intended to show the placeholder syntax, but it doesn't fit a real situation.

I made another one that, I hope, better illustrates the advantages:

Let's have an application with:
- A blog, with readers, authors, and moderators. Both authors and moderators are readers. 
- A shop, with basic users who can browse, buyers, sellers, and moderators. Sellers and moderators are not necessarily buyers. 
- A moderator zone, where moderators can see all pending requests or whatever, protected with access control (`/moderation` requires role `ROLE_MODERATOR`).

Current configuration:
```yaml
role_hierarchy:
  # Blog readers can view posts and add comments
  ROLE_BLOG_READER: 
    - ROLE_USER
    - ROLE_BLOG_VIEW
    - ROLE_BLOG_COMMENT
  # Blog moderators are blog readers who can moderate comments and posts
  ROLE_BLOG_MODERATOR:
    - ROLE_MODERATOR
    - ROLE_BLOG_READER
    - ROLE_BLOG_MODERATE_COMMENT
    - ROLE_BLOG_MODERATE_POST
  # Blog authors are blog readers who can create posts
  ROLE_BLOG_AUTHOR: 
    - ROLE_BLOG_READER
    - ROLE_BLOG_CREATE_POST

  # Shop users can view items
  ROLE_SHOP_USER:
    - ROLE_USER
    - ROLE_SHOP_VIEW
  # Shop buyers are shop users who can view items, buy them, and post reviews
  ROLE_SHOP_BUYER: 
    - ROLE_SHOP_USER
    - ROLE_SHOP_BUY_ITEM
    - ROLE_SHOP_POST_REVIEW
  # Shop sellers are shop users who can add items 
  ROLE_SHOP_SELLER:
    - ROLE_SHOP_USER
    - ROLE_SHOP_CREATE_ITEM
  # Shop moderators are shop users who and moderate items and reviews
  ROLE_SHOP_MODERATOR:
    - ROLE_SHOP_USER
    - ROLE_MODERATOR
    - ROLE_SHOP_MODERATE_ITEM
    - ROLE_SHOP_MODERATE_REVIEW
```

With placeholders:
```yaml
role_hierarchy:
  # All users with roles have ROLE_USER
  ROLE_*: ROLE_USER
  # All moderators have ROLE_MODERATOR
  ROLE_*_MODERATOR: ROLE_MODERATOR

  # Having a blog role allows to view the blog  
  ROLE_BLOG_*: ROLE_BLOG_VIEW
  # Blog readers can comment posts  
  ROLE_BLOG_READER:
    - ROLE_BLOG_COMMENT
  # Blog moderators can also moderate comments and posts
  ROLE_BLOG_MODERATOR:
    - ROLE_BLOG_READER
    - ROLE_BLOG_MODERATE_COMMENT
    - ROLE_BLOG_MODERATE_POST
  # Blog authors can also create posts 
  ROLE_BLOG_AUTHOR:
    - ROLE_BLOG_READER
    - ROLE_BLOG_CREATE_POST

  # Having a shop role allows to view the shop
  ROLE_SHOP_*: ROLE_SHOP_VIEW
  # Shop buyers can buy and review items
  ROLE_SHOP_BUYER:
    - ROLE_SHOP_BUY_ITEM
    - ROLE_SHOP_POST_REVIEW
  # Shop sellers can add items
  ROLE_SHOP_SELLER: ROLE_SHOP_CREATE_ITEM
  # Shop moderators can moderate items and reviews
  ROLE_SHOP_MODERATOR:
    - ROLE_SHOP_MODERATE_ITEM
    - ROLE_SHOP_MODERATE_REVIEW
```

The advantages I see here:
- All roles grant `ROLE_USER`.
- The role `ROLE_MODERATOR` is easy to configure globally.
- It's easy to configure that _anyone who can do something on the shop can view the shop_, even if the role is not explicitly declared in the hierarchy configuration (for example a user who can only post reviews on items but not buy anything).
- It's scalable: we can add more domain sections to the application, following the naming convention for moderators will handle their access to the moderator zone automatically.

The drawbacks: 
- It requires some naming conventions. But here the example is simple, I also consider more complex use cases for application having dozens of domain-specific roles, and I assume having a good naming convention is a must-have in this case.

Wdyt about this one? If it's better I can update the PR description with it.

src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
squrious: Should be fixed with [495b2ce](https://github.com/symfony/symfony/pull/52099/commits/495b2ce9178436300adf49b0a9b39e1a34ebadb5)

smnandre:
> Wdyt about this one? If it's better I can update the PR description with it.

Absolutely perfect for me, thanks a lot 😄   (and you have now your documentation PR 100% ready 👏 )

squrious:
Hi there!

I worked on the `debug:roles` command. Here are some usages: 

`bin/console debug:roles`:

image

`bin/console debug:roles --tree`:

image


`bin/console debug:roles ROLE_BLOG_MODERATOR`:

image

`bin/console debug:roles ROLE_BLOG_MODERATOR --tree`:

image

However I'm not sure about the strategy to debug the `RoleHierarchy`. I added a wrapper so it can access needed variables through reflection, but in find it weak at the moment:
- It doesn't work with custom implementations of `RoleHierarchyInterface`
- I needed to access the method to match placeholders through reflection too, and at this point I wonder if a dedicated service for the placeholder matching logic would be better

Does it make sense to dynamically configure the command options depending on the current `RoleHierarchyInterface` implementation? The tree view is only applicable for the built in impl, but it could be nice to still resolve reachable roles for custom impl.

Jean-Beru:
Nice PR @squrious !

Wrapping `RoleHierarchyInterface` to access protected/private properties/methods could introduce bugs when using a custom implementation (as you said) or decorating it. 

I think we have to found an other way or make this command simpler.

src/Symfony/Bundle/SecurityBundle/Debug/DebugRoleHierarchy.php
```diff
@@ -0,0 +1,138 @@
...
+    public function __construct(private readonly RoleHierarchyInterface $decorated)
+    {
+        if (
+            property_exists($this->decorated, 'map')
+            && property_exists($this->decorated, 'hierarchy')
+            && property_exists($this->decorated, 'rolePlaceholdersPatterns')
+            && method_exists($this->decorated, 'getMatchingPlaceholders')
+        ) {
+            $this->map = (new \ReflectionProperty($this->decorated, 'map'))->getValue($this->decorated);
+            $this->hierarchy = (new \ReflectionProperty($this->decorated, 'hierarchy'))->getValue($this->decorated);
```
stof: Using Reflection to access private properties is a bad idea. Using reflection to access private properties in a different package is a nightmare, as there is strictly no BC guarantee for those (the fact that the property exist does not even mean that it has the type that you expect here, as it might be a custom implementation as well)

src/Symfony/Component/Security/Core/Role/RoleHierarchy.php
```diff
@@ -76,6 +74,66 @@ protected function buildRoleMap()
...
+    /**
+     * Build the regex pattern for the given role:
+     *   - Replace valid wildcards with a non-wildcard matching pattern.
+     *   - Escape reserved regex characters.
+     *
+     * A valid wildcard is a * prefixed with _ and immediately followed by _ or EOL.
+     *
+     * @return string|false The regex pattern, or false if there is no valid wildcard in the role
+     */
+    private function getPlaceholderPattern(string $role): string|false
```
stof: The proper representation of the absence of value is `null`, not `false`

src/Symfony/Bundle/SecurityBundle/Debug/DebugRoleHierarchy.php
squrious: That's what I thought... Still trying to find a better solution

squrious:
> I think we have to found an other way or make this command simpler.

I reworked the solution to avoid the use of reflection. Now a dedicated `debug.security.role_hierarchy` is registered and injected in the command. If the role hierarchy uses a custom implementation, it is injected "as it is" and the command is only able to display reachable role names. I also added some tests to enforce this behavior. 

I also added a changelog entry to the Security Bundle. But the tests fail because the command is in a different package, so maybe it would be better to add the command in another PR when/if this one gets merged?

src/Symfony/Bundle/SecurityBundle/Debug/DebugRoleHierarchy.php
squrious: I removed reflection stuff in favor of inheritance and a dedicated service for the debug purposes.

nicolas-grekas:
Thanks for your patience @squrious, this sat far too long for work that was essentially done in 2023.

I split it in two, as the discussion here was mostly about the command rather than the feature:

* this PR keeps the placeholder support alone;
* the `debug:roles` command moved to #65345, which builds on this one and should be merged after it.

The hierarchy had been rewritten on 8.2 in the meantime, `$map` is now keyed by role name and `getParentRoleNames()` was added, so I ported the feature onto the current shape rather than merging. Three things changed along the way:

**A role holding a regex delimiter broke the pattern.** `getPlaceholderPattern()` called `preg_quote($role)` without the delimiter while building a `/…/` pattern:

```
ROLE_A/B_*  ->  PHP warning: preg_grep(): Unknown modifier 'B'
                and the placeholder silently never matched
```

It is `preg_quote($role, '/')` now, with a test.

**Applications without placeholders were paying for the feature.** `getReachableRoleNames()` runs on every `RoleHierarchyVoter` vote and the voter does not cache, so the extra work landed on every authorization check. Measured over 20k calls:

```
8.2                              0.34 us/call
before, no placeholder declared  0.52 us/call
after,  no placeholder declared  0.34 us/call
```

The resolution is skipped entirely when the hierarchy declares no placeholder.

**Placeholders leaked into `getParentRoleNames()`.** That method is new since this PR was written. It walks the map, so it returned `ROLE_FOO_*` as a parent of `ROLE_USER`, which is not a role anyone can hold. Placeholder keys are excluded there now, with a test.

Security/Core is green at 407 tests, and reverting `RoleHierarchy.php` fails 7 of them.

nicolas-grekas:
Follow-up: `$rolePlaceholdersPatterns` and `getMatchingPlaceholders()` are private here now. They only need to be protected for `DebugRoleHierarchy`, so #65345 widens them itself. Nothing is exposed unless the command is taken, and private to protected is a safe widening within the same release.

nicolas-grekas:
Thank you @squrious.

PR #52100, opened 2023-10-17
[AssetMapper] Disable profiler when the "dev server" respond

smnandre:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

In debug mode, the "dev server" send assets as if they were compiled in the public directory.

To improve the DX and not "pollute" too much the profiler (see capture below), this PR does two things : 
* disable the profiler for those requests 
* stop the request & response event propagation, to speed things a bit

Capture d’écran 2023-10-17 à 10 57 44

### DIFF
--- a/src/Symfony/Component/AssetMapper/AssetMapperDevServerSubscriber.php
+++ b/src/Symfony/Component/AssetMapper/AssetMapperDevServerSubscriber.php
@@ -15,8 +15,10 @@ use Psr\Cache\CacheItemPoolInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Event\RequestEvent;
+use Symfony\Component\HttpKernel\Event\ResponseEvent;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\HttpKernel\Profiler\Profiler;
 
 /**
  * Functions like a controller that returns assets from the asset mapper.
@@ -104,6 +106,7 @@ final class AssetMapperDevServerSubscriber implements EventSubscriberInterface
         string $publicPrefix = '/assets/',
         array $extensionsMap = [],
         private readonly ?CacheItemPoolInterface $cacheMapCache = null,
+        private readonly ?Profiler $profiler = null,
     ) {
         $this->publicPrefix = rtrim($publicPrefix, '/').'/';
         $this->extensionsMap = array_merge(self::EXTENSIONS_MAP, $extensionsMap);
@@ -126,6 +129,8 @@ final class AssetMapperDevServerSubscriber implements EventSubscriberInterface
             throw new NotFoundHttpException(sprintf('Asset with public path "%s" not found.', $pathInfo));
         }
 
+        $this->profiler?->disable();
+
         $mediaType = $this->getMediaType($asset->publicPath);
         $response = (new Response(
             $asset->content,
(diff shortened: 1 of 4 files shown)


src/Symfony/Component/AssetMapper/AssetMapperDevServerSubscriber.php
```diff
@@ -134,17 +139,28 @@ public function onKernelRequest(RequestEvent $event): void
             ->setPublic()
             ->setMaxAge(604800)
             ->setImmutable()
-            ->setEtag($asset->digest)
-        ;
+            ->setEtag($asset->digest);
+
+        $response->headers->set('X-AssetMapper-Server', true);
```
nicolas-grekas: just bikeshedding :)
```suggestion
        $response->headers->set('X-DevMappedAsset', true);
```

nicolas-grekas (approved):
There's one thing I'm not comfortable with in this class, it's `string $publicPrefix = '/assets/'`.
This works only when the app is hosted at the root of the domain, and we know that's not always the case, and that we can deal with this in other components.
That's unrelated to this PR but still :)

smnandre:
> There's one thing I'm not comfortable with in this class, it's `string $publicPrefix = '/assets/'`.

~I agree, but how could we mimic an external CDN on localhost ? :|~ (missread, sorry)

How would you handle that ?

src/Symfony/Component/AssetMapper/AssetMapperDevServerSubscriber.php
smnandre: ✅  🚲

nicolas-grekas:
The routing component is the one that can deal with this. IIRC, it uses `RequestContext` to do so.
Instead of hardcoding the prefix, maybe we could let the routing component match the request and act on that?
That'd require ppl to define a route in their apps (thanks to a recipe that should be easy). Then, this listener would act when the request has an attribute  (`_controller`?) that matches some convention to trigger it.
WDYT? /cc @weaverryan

smnandre:
> Instead of hardcoding the prefix

It's just a default value... the instance from the container uses the "asset_mapper.public_prefix"

src/Symfony/Component/AssetMapper/AssetMapperDevServerSubscriber.php
```diff
@@ -134,17 +139,28 @@ public function onKernelRequest(RequestEvent $event): void
             ->setPublic()
             ->setMaxAge(604800)
             ->setImmutable()
-            ->setEtag($asset->digest)
-        ;
+            ->setEtag($asset->digest);
```
fabpot: Should be reverted (not related)

src/Symfony/Component/AssetMapper/AssetMapperDevServerSubscriber.php
```diff
@@ -134,17 +139,28 @@ public function onKernelRequest(RequestEvent $event): void
...
+
+        $response->headers->set('X-DevMappedAsset', true);
```
fabpot: To be consistent/more readable, a dash should probably separate words. What about something simpler as well. Maybe `X-Assets-Dev`?

fabpot (approved):
We might consider changing the Response header name? @weaverryan?

src/Symfony/Component/AssetMapper/AssetMapperDevServerSubscriber.php
smnandre: 👍

weaverryan:
> The routing component is the one that can deal with this. IIRC, it uses RequestContext to do so.
Instead of hardcoding the prefix, maybe we could let the routing component match the request and act on that?

I'm not sure there's a problem - the logic for checking to see if this controller should be used is:

```php
$pathInfo = $event->getRequest()->getPathInfo();
if (!str_starts_with($pathInfo, $this->publicPrefix)) {
    return;
}
```

So the `pathInfo` removes the subdirectory for the comparison. I just tested with a subdirectory, and the paths all seem fine - but let me know if you still see some risk.

weaverryan (approved):

nicolas-grekas:
Thanks for checking :rocket:

nicolas-grekas:
Thank you @smnandre.

PR #52102, opened 2023-10-17
DX: PHP CS Fixer cleanup for 5.4 line

keradus:
Hey, I was wondering recently to cleanup `@Symfony` ruleset.
Yet, it is tricky to adjust ruleset to codebase, as... I took a look for current config file on 5.3 line, and codebase is not following it already, actually.

Would it make sense to clean it up? It would help to decide and potentially apply new rules that pop up in Fixer.
Or maybe you want this to potentially be cleaned up only on higher version of Symfony?

Tell me your preference, I can do as you prefer.

For now, I applied the following config only, which is subset of `@Symfony` ruleset.

```
        'phpdoc_trim' => true,
        'no_superfluous_phpdoc_tags' => ['allow_mixed' => true, 'remove_inheritdoc' => true],
        'no_empty_phpdoc' => true,
        'no_extra_blank_lines' => true,
        'phpdoc_trim' => true,
        'phpdoc_trim_consecutive_blank_line_separation' => true,
        'no_trailing_whitespace' => true,
        'class_attributes_separation' => ['elements' => ['method' => 'one']],
```

```
php php-cs-fixer.phar list-files --config=.php-cs-fixer.dist.php | xargs -n 50 -P 16 php php-cs-fixer.phar fix --config=.php-cs-fixer.dist.php --path-mode intersection -v
```


nicolas-grekas:
We changed a bit our practice and we now do such massive cleanups on the latest dev-branch. I think those rules have already been applied there.

src/Symfony/Bridge/Doctrine/DependencyInjection/AbstractDoctrineExtension.php
```diff
@@ -135,8 +135,6 @@ protected function setMappingDriverConfig(array $mappingConfig, string $mappingN
      *
      * Returns false when autodetection failed, an array of the completed information otherwise.
      *
-     * @param string|null $bundleDir The bundle directory path
```
nicolas-grekas, line 138: FYI, these are false-positive for us: we use `@param` to tell about parameters we'll add in the next major (see also the special comment in the signature)

keradus:
> We changed a bit our practice and we now do such massive cleanups on the latest dev-branch.

Perfect, that simplifies a lot, also on my side!

> I think those rules have already been applied there.

Some of them were, but I couldn't determine if all, due to mentioned issue.
But, not important anymore ;)

src/Symfony/Bridge/Doctrine/DependencyInjection/AbstractDoctrineExtension.php
keradus: I wonder if we can do anything about this.
https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/7376

PR #52117, opened 2023-10-17
DX: PHP CS Fixer - update excluded paths and apply some minor CS

keradus:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT

### DIFF
--- a/src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler.php
+++ b/src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler.php
@@ -96,7 +96,7 @@ class DeprecationErrorHandler
     {
         $deprecations = [];
         $previousErrorHandler = set_error_handler(function ($type, $msg, $file, $line, $context = []) use (&$deprecations, &$previousErrorHandler) {
-            if (\E_USER_DEPRECATED !== $type && \E_DEPRECATED !== $type && (\E_WARNING !== $type || false === strpos($msg, '" targeting switch is equivalent to "break'))) {
+            if (\E_USER_DEPRECATED !== $type && \E_DEPRECATED !== $type && (\E_WARNING !== $type || !str_contains($msg, '" targeting switch is equivalent to "break'))) {
                 if ($previousErrorHandler) {
                     return $previousErrorHandler($type, $msg, $file, $line, $context);
                 }
@@ -128,7 +128,7 @@ class DeprecationErrorHandler
      */
     public function handleError($type, $msg, $file, $line, $context = [])
     {
-        if ((\E_USER_DEPRECATED !== $type && \E_DEPRECATED !== $type && (\E_WARNING !== $type || false === strpos($msg, '" targeting switch is equivalent to "break'))) || !$this->getConfiguration()->isEnabled()) {
+        if ((\E_USER_DEPRECATED !== $type && \E_DEPRECATED !== $type && (\E_WARNING !== $type || !str_contains($msg, '" targeting switch is equivalent to "break'))) || !$this->getConfiguration()->isEnabled()) {
             return \call_user_func(self::getPhpUnitErrorHandler(), $type, $msg, $file, $line, $context);
         }
 
--- a/src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler/Deprecation.php
+++ b/src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler/Deprecation.php
@@ -104,7 +104,7 @@ class Deprecation
                     $this->getOriginalFilesStack();
                     array_splice($this->originalFilesStack, 0, $j, [$this->triggeringFile]);
 
-                    if (preg_match('/(?|"([^"]++)" that is deprecated|should implement method "(?:static )?([^:]++))/', $message, $m) || (false === strpos($message, '()" will return') && false === strpos($message, 'native return type declaration') && preg_match('/^(?:The|Method) "([^":]++)/', $message, $m))) {
+                    if (preg_match('/(?|"([^"]++)" that is deprecated|should implement method "(?:static )?([^:]++))/', $message, $m) || (!str_contains($message, '()" will return') && !str_contains($message, 'native return type declaration') && preg_match('/^(?:The|Method) "([^":]++)/', $message, $m))) {
                         $this->triggeringFile = (new \ReflectionClass($m[1]))->getFileName();
                         array_unshift($this->originalFilesStack, $this->triggeringFile);
                     }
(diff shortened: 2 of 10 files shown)


src/Symfony/Bridge/PhpUnit/CoverageListener.php
```diff
@@ -29,7 +29,7 @@ class CoverageListener implements TestListener
     public function __construct(callable $sutFqcnResolver = null, bool $warningOnSutNotFound = false)
     {
         $this->sutFqcnResolver = $sutFqcnResolver ?? static function (Test $test): ?string {
-            $class = \get_class($test);
+            $class = $test::class;
```
xabbuh: these changes in the PHPUnit bridge need to be reverted as it still supports PHP 7

xabbuh (changes requested):

src/Symfony/Bridge/PhpUnit/CoverageListener.php
stof: Does this work on PHP 7.1 ? The PHPUnit Bridge is special in term on min supported PHP version.

src/Symfony/Component/Validator/Mapping/GenericMetadata.php
```diff
@@ -38,8 +38,6 @@ class GenericMetadata implements MetadataInterface
     public array $constraints = [];
 
     /**
-     * @var array
```
stof, line 41: I suggest updating it to `array<string, list<Constraint>>` to have the precise type instead of removing the type.

src/Symfony/Bridge/PhpUnit/CoverageListener.php
keradus: thanks for noticing, let's move discussion to https://github.com/symfony/symfony/pull/52117/commits/2b03e8322fa247da90ee6af144af9993dd75e773#r1363446539

keradus: ack, reverted.

then, I strongly encourage to disable the rule, as one cannot apply current Fixer config in such case and trust to use the results. 
https://github.com/symfony/symfony/pull/52117/commits/aadeffb2b11ca03b8868ae983b8531c8c6a29996

stof: What we would need would be to be able to disable some rules only for some paths. But php-cs-fixer does not support this.

keradus: Very open to see such PR.
Yet, for now, if John Doe comes, see the Fixer config, apply it and see lot of changes needed, and those changes breaking CI, it's very confusing.

src/Symfony/Component/Validator/Mapping/GenericMetadata.php
keradus: sure. automate dropped the pointless type hint. if you have better one, happy to add it.

also looking at other annotations around those properties, they use `Constraint[]`, let's be consistent on it - i followed what's already present
de34f51629eb7372a43217aa85510051fb7bcf17

src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler/Deprecation.php
```diff
@@ -211,7 +211,7 @@ public function originatingClass()
 
         $class = $this->originClass;
 
-        return false !== strpos($class, "@anonymous\0") ? (get_parent_class($class) ?: key(class_implements($class)) ?: 'class').'@anonymous' : $class;
+        return   str_contains($class, "@anonymous\0") ? (get_parent_class($class) ?: key(class_implements($class)) ?: 'class').'@anonymous' : $class;
```
keradus, line 214:
```suggestion
        return str_contains($class, "@anonymous\0") ? (get_parent_class($class) ?: key(class_implements($class)) ?: 'class').'@anonymous' : $class;
```

src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler/Deprecation.php
```diff
@@ -262,10 +262,10 @@ public function isMuted()
             return false;
         }
         if (isset($this->trace[1]['class'])) {
-            return 0 === strpos($this->trace[1]['class'], 'PHPUnit\\');
+            return   str_starts_with($this->trace[1]['class'], 'PHPUnit\\');
         }
 
-        return false !== strpos($this->triggeringFile, \DIRECTORY_SEPARATOR.'vendor'.\DIRECTORY_SEPARATOR.'phpunit'.\DIRECTORY_SEPARATOR);
+        return   str_contains($this->triggeringFile, \DIRECTORY_SEPARATOR.'vendor'.\DIRECTORY_SEPARATOR.'phpunit'.\DIRECTORY_SEPARATOR);
```
keradus, line 268:
```suggestion
        return str_contains($this->triggeringFile, \DIRECTORY_SEPARATOR.'vendor'.\DIRECTORY_SEPARATOR.'phpunit'.\DIRECTORY_SEPARATOR);
```

src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler/Deprecation.php
```diff
@@ -262,10 +262,10 @@ public function isMuted()
             return false;
         }
         if (isset($this->trace[1]['class'])) {
-            return 0 === strpos($this->trace[1]['class'], 'PHPUnit\\');
+            return   str_starts_with($this->trace[1]['class'], 'PHPUnit\\');
```
keradus, line 265:
```suggestion
            return str_starts_with($this->trace[1]['class'], 'PHPUnit\\');
```

fabpot (approved):

fabpot:
Thank you @keradus.

src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler.php
```diff
@@ -96,7 +96,7 @@ public static function collectDeprecations($outputFile)
     {
         $deprecations = [];
         $previousErrorHandler = set_error_handler(function ($type, $msg, $file, $line, $context = []) use (&$deprecations, &$previousErrorHandler) {
-            if (\E_USER_DEPRECATED !== $type && \E_DEPRECATED !== $type && (\E_WARNING !== $type || false === strpos($msg, '" targeting switch is equivalent to "break'))) {
+            if (\E_USER_DEPRECATED !== $type && \E_DEPRECATED !== $type && (\E_WARNING !== $type || !str_contains($msg, '" targeting switch is equivalent to "break'))) {
```
nicolas-grekas, line 99: I'm reverting all those, the phpunit-bridge doesn't support php and we don't want it to use polyfills

keradus: @nicolas-grekas , i'm curious then, as bridge is already using 8.1 polyfill. Why not using 8.0 polyfill ?

nicolas-grekas: After checking, this is needed for testing the enum_exists mock. Also, this is require-dev!

Issue #52077, opened 2023-10-16
Behavior changed with messenger:consume and signals


hlecorche:
### Symfony version(s) affected

6.3.5 and 6.3-dev

### Description

Since version 6.3.5, there is a compatibility break with the `messenger:consume` command and signals.

* After : If  `SIGINT` or `SIGTERM` signals were sent, the worker stopped AFTER processing the current message
* Now (6.3.5 and 6.3-dev) : If  `SIGINT` or `SIGTERM` signals are sent, the worker stops immediately even if a message is being processed (hard stop)

Here are some issues with this change :
* If Supervisor is used, the `supervisorctl stop xxx` command can stop the worker before the processing of the current message is completed (hard stop).
* I'm using the `Symfony\Component\Messenger\Event\WorkerMessageFailedEvent` event to catch an error. If an error is encountered, the listener stops the task group with Supervisor (with its API and  `SIGTERM`) and sends an email. Now when the event calls the Supervisor API to stop, all actions performed after its call are no longer performed:
  * The email is no longer sent
  * Retry strategy does not work 
  * Message processing was stopped abruptly

Link with #50787 ?

### How to reproduce

Repository : 
* Works with < 6.3.5 : https://github.com/hlecorche/symfony-messenger-signal-issue
* Doesn't work with 6.3.5 and 6.3-dev : https://github.com/hlecorche/symfony-messenger-signal-issue/tree/6.3-dev

Commands :
* Execute the command `php bin/console doctrine:database:create && php bin/console doctrine:schema:update --force`
* Execute the command `php bin/console  app:create-message && php bin/console messenger:consume async -vvv`
* Send SIGTERM signal

After (< 6.3.5) :
```
php bin/console  app:create-message && php bin/console messenger:consume async -vvv

                                                                                                                        
 [OK] Consuming messages from transport "async".                                                                        
                                                                                                                        

 // The worker will automatically exit once it has received a stop signal via the messenger:stop-workers command.       

 // Quit the worker with CONTROL-C.                                                                                     

[info] Received message App\Messenger\Message\MessageSuccess
[info] Received signal 15.
[info] Stopping worker.
^ "The worker will stop correctly"
[info] Message App\Messenger\Message\MessageSuccess handled by App\Messenger\MessageHandler\MessageSuccessHandler::__invoke
[info] App\Messenger\Message\MessageSuccess was handled successfully (acknowledging to transport).
```

Now (6.3.5 and 6.3-dev) :
```

[...]

PR #52080, opened 2023-10-16
[Messenger] Fix graceful exit

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #52077
| License       | MIT

My previous PR #50787 accidentally broke the behavior of the `messenger:consume` command. It no longer waits for the handler to finish, instead it exists immediately.

PR #52123, opened 2023-10-18
Fix missing dispatch `WorkerStoppedEvent` after `SIGINT` or `SIGTERM`

Chris53897:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes?
| New feature?  | no 
| Deprecations? | no
| License       | MIT

Related to https://github.com/symfony/symfony/issues/52077

In my point of view this Event should be dispatched here.
But maybe there are reasons for not doing this in the past.

I am not sure if bugfix or new feature.

### DIFF
--- a/src/Symfony/Component/Messenger/Command/ConsumeMessagesCommand.php
+++ b/src/Symfony/Component/Messenger/Command/ConsumeMessagesCommand.php
@@ -28,6 +28,7 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\ChoiceQuestion;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\Messenger\Event\WorkerStoppedEvent;
 use Symfony\Component\Messenger\EventListener\ResetServicesListener;
 use Symfony\Component\Messenger\EventListener\StopWorkerOnFailureLimitListener;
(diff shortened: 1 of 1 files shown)


src/Symfony/Component/Messenger/Command/ConsumeMessagesCommand.php
```diff
@@ -271,6 +272,8 @@ public function handleSignal(int $signal, int|false $previousExitCode = 0): int|
 
         $this->worker->stop();
 
+        $this->eventDispatcher?->dispatch(new WorkerStoppedEvent($this->worker));
```
xabbuh, line 275: This doesn't look necessary to me. The event should already be dispatched at the end of the `Worker::run()` method.

Chris53897: `Worker::run()` is not called in the case of SIGINT or SIGTERM

xabbuh: Are you sure that's still the case after the fix from #52080?

Chris53897: i will check that

OskarStark:
Any news @Chris53897 ?

Chris53897:
Sorry for late response.
Look like the the other PR fixes this.
Last time i checked only 6.3.4 (as i thought it will have the same behaviour. like described in the issue). 6.3.4 does not throw the event.

I will close this PR.

PR #52124, opened 2023-10-18
[HttpKernel] add missing class alias to satisfy autoloading

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | 
| License       | MIT

### DIFF
--- a/src/Symfony/Component/HttpKernel/Debug/FileLinkFormatter.php
+++ b/src/Symfony/Component/HttpKernel/Debug/FileLinkFormatter.php
@@ -17,6 +17,10 @@ trigger_deprecation('symfony/http-kernel', '6.4', 'The "%s" class is deprecated,
 
 class_exists(ErrorHandlerFileLinkFormatter::class);
 
+if (!class_exists(FileLinkFormatter::class, false)) {
+    class_alias(ErrorHandlerFileLinkFormatter::class, FileLinkFormatter::class);
+}
+
 if (false) {
     /**
      * @deprecated since Symfony 6.4, use FileLinkFormatter from the ErrorHandler component instead


src/Symfony/Component/HttpKernel/Debug/FileLinkFormatter.php
```diff
@@ -16,6 +16,7 @@
 trigger_deprecation('symfony/http-kernel', '6.4', 'The "%s" class is deprecated, use "%s" instead.', FileLinkFormatter::class, ErrorHandlerFileLinkFormatter::class);
 
 class_exists(ErrorHandlerFileLinkFormatter::class);
+class_alias(ErrorHandlerFileLinkFormatter::class, FileLinkFormatter::class);
```
nicolas-grekas: the alias already exists in `ErrorHandlerFileLinkFormatter` so this should not be needed
are you checking cross version failures ? I removed the alias from 7.0, maybe that's the issue?

xabbuh: tests are failing on 6.3 with 6.4 of the HttpKernel component installed: https://github.com/symfony/symfony/actions/runs/6550029688/job/17788242726#step:8:10353

xabbuh: okay, but the ErrorHandler component is installed in version 7

nicolas-grekas: okay! then yes, we can add this alias! maybe wrap it in a class_exists(, false) check to not collide with the alias defined in the new class on 6.4?

xabbuh: good point, I wasn't aware of the other class alias, `class_exists()` check added

PR #52128, opened 2023-10-18
[HttpKernel] Introduce `ExceptionEvent::isKernelTerminating()` to skip error rendering when kernel is terminating

### LINKED ISSUE #51912: [HttpKernel] ErrorListener::onKernelException should ignore exception thrown after `kernel.terminate`

VincentLanglet:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #51912
| License       | MIT

### DIFF
--- a/UPGRADE-7.1.md
+++ b/UPGRADE-7.1.md
@@ -1,0 +1,7 @@
+UPGRADE FROM 7.0 to 7.1
+=======================
+
+HttpKernel
+----------
+
+ * `ExceptionEvent` now takes an optional `$isKernelTerminating` parameter
--- a/src/Symfony/Component/HttpKernel/EventListener/ErrorListener.php
+++ b/src/Symfony/Component/HttpKernel/EventListener/ErrorListener.php
@@ -102,6 +102,10 @@ class ErrorListener implements EventSubscriberInterface
             return;
         }
 
+        if (!$this->debug && $event->isKernelTerminating()) {
+            return;
+        }
+
         $throwable = $event->getThrowable();
 
         if ($exceptionHandler = set_exception_handler(var_dump(...))) {
--- a/src/Symfony/Component/HttpKernel/HttpKernel.php
+++ b/src/Symfony/Component/HttpKernel/HttpKernel.php
@@ -56,6 +56,7 @@ class HttpKernel implements HttpKernelInterface, TerminableInterface
     protected $requestStack;
     private ArgumentResolverInterface $argumentResolver;
     private bool $handleAllThrowables;
+    private bool $terminating = false;
 
     public function __construct(EventDispatcherInterface $dispatcher, ControllerResolverInterface $resolver, RequestStack $requestStack = null, ArgumentResolverInterface $argumentResolver = null, bool $handleAllThrowables = false)
     {
@@ -112,7 +113,12 @@ class HttpKernel implements HttpKernelInterface, TerminableInterface
      */
     public function terminate(Request $request, Response $response)
     {
-        $this->dispatcher->dispatch(new TerminateEvent($this, $request, $response), KernelEvents::TERMINATE);
+        try {
+            $this->terminating = true;
+            $this->dispatcher->dispatch(new TerminateEvent($this, $request, $response), KernelEvents::TERMINATE);
+        } finally {
+            $this->terminating = false;
+        }
     }
 
     /**
@@ -235,7 +241,7 @@ class HttpKernel implements HttpKernelInterface, TerminableInterface
      */
     private function handleThrowable(\Throwable $e, Request $request, int $type): Response
     {
-        $event = new ExceptionEvent($this, $request, $type, $e);
+        $event = new ExceptionEvent($this, $request, $type, $e, isKernelTerminating: $this->terminating);
         $this->dispatcher->dispatch($event, KernelEvents::EXCEPTION);
 
         // a listener might have replaced the exception
--- a/src/Symfony/Component/HttpKernel/Tests/EventListener/ErrorListenerTest.php
+++ b/src/Symfony/Component/HttpKernel/Tests/EventListener/ErrorListenerTest.php
@@ -244,6 +244,19 @@ class ErrorListenerTest extends TestCase
         $this->assertFalse($response->headers->has('content-security-policy'), 'CSP header has been removed');
     }
 
+    public function testTerminating()
+    {
+        $listener = new ErrorListener('foo', $this->createMock(LoggerInterface::class));
+
+        $kernel = $this->createMock(HttpKernelInterface::class);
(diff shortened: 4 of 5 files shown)


nicolas-grekas (commented):
I likely misunderstood your proposal but I was expecting the new method on class `ExceptionEvent`.

VincentLanglet:
> I likely misunderstood your proposal but I was expecting the new method on class `ExceptionEvent`.

Indeed I thought about it first, then I saw the kernel was accessible in the ExceptionEvent, this would have avoid an extra param. Also, it would have avoid having to think about passing both the kernel and the terminated state.

I updated the PR to follow this strategy if it's preferred @nicolas-grekas

src/Symfony/Component/HttpKernel/EventListener/ErrorListener.php
```diff
@@ -102,6 +102,13 @@ public function onKernelException(ExceptionEvent $event)
             return;
         }
 
+        if (
+            !$this->debug
+            && $event->isKernelTerminated()
+        ) {
```
nicolas-grekas:
```suggestion
        if (!$this->debug && $event->isKernelTerminated()) {
```

src/Symfony/Component/HttpKernel/HttpKernel.php
```diff
@@ -112,6 +113,7 @@ public function handle(Request $request, int $type = HttpKernelInterface::MAIN_R
      */
     public function terminate(Request $request, Response $response)
     {
+        $this->terminated = true;
```
nicolas-grekas: This adds state to the kernel - please check when this state can be reset in case the kernel is reused after calling this method. Or maybe a try/finally + rename this `terminating`?

nicolas-grekas:
Don't miss adding a line to the changelog of the component.

VincentLanglet:
> Don't miss adding a line to the changelog of the component.

Should it be in the 6.4 changelog or should I create a 7.1 one ?

nicolas-grekas:
7.1 indeed now

src/Symfony/Component/HttpKernel/Tests/EventListener/ErrorListenerTest.php
```diff
@@ -244,6 +244,19 @@ public function testCSPHeaderIsRemoved()
         $this->assertFalse($response->headers->has('content-security-policy'), 'CSP header has been removed');
     }
 
+    public function testTerminated(): void
```
nicolas-grekas:
```suggestion
    public function testTerminating(): void
```

src/Symfony/Component/HttpKernel/Tests/EventListener/ErrorListenerTest.php
```diff
@@ -244,6 +244,19 @@ public function testCSPHeaderIsRemoved()
...
+    {
+        $listener = new ErrorListener('foo', $this->createMock(LoggerInterface::class));
+
+        $kernel = $this->createMock(HttpKernelInterface::class);
+        $kernel->expects($this->never())->method('handle');
+
+        $request = Request::create('/');
+
+        $event = new ExceptionEvent($kernel, $request, HttpKernelInterface::MAIN_REQUEST, new \Exception('foo'), true);
```
nicolas-grekas, line 256:
```suggestion
        $event = new ExceptionEvent($kernel, $request, HttpKernelInterface::MAIN_REQUEST, new \Exception('foo'), isKernelTerminating: true);
```

UPGRADE-7.1.md
```diff
@@ -0,0 +1,7 @@
+UPGRADE FROM 7.0 to 7.1
+=======================
+
+HttpKernel
+-----
+
+* `ExceptionEvent` now takes an optional `$isKernelTerminating` parameter
```
nicolas-grekas:
```suggestion
 * `ExceptionEvent` now takes an optional `$isKernelTerminating` parameter
```

UPGRADE-7.1.md
```diff
@@ -0,0 +1,7 @@
+UPGRADE FROM 7.0 to 7.1
+=======================
+
+HttpKernel
+-----
```
nicolas-grekas:
```suggestion
----------
```

nicolas-grekas (approved):

UPGRADE-7.1.md
VincentLanglet: Done

fabpot (approved):

fabpot:
Thank you @VincentLanglet.

UPGRADE-7.1.md
```diff
@@ -0,0 +1,7 @@
+UPGRADE FROM 7.0 to 7.1
+=======================
+
+HttpKernel
+----------
+
+ * `ExceptionEvent` now takes an optional `$isKernelTerminating` parameter
```
OskarStark, line 7: Shouldn't we add this to the CHANGELOG instead to the upgrade guide, as this is an optional one?

PR #51791, opened 2023-09-30
[Messenger] Check if PCNTL is installed

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #51790
| License       | MIT

Fixes #51790

PR #52129, opened 2023-10-18
[Messenger] Add ext-pcntl to composer.json

beesofts:
Since ConsumeMessagesCommand.php uses SIGTERM and SIGINT, ext-pcntl must be available on system

### DIFF
--- a/src/Symfony/Component/Messenger/composer.json
+++ b/src/Symfony/Component/Messenger/composer.json
@@ -17,6 +17,7 @@
     ],
     "require": {
         "php": ">=8.1",
+        "ext-pcntl": "*",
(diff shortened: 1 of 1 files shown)


fabpot:
It's optional.
Maybe #51791 is what you're looking for.

derrabus (changes requested):
This change will basically forbid installing messenger on Windows. 👎🏻

SerheyDolgushev:
@beesofts can you please check https://github.com/symfony/symfony/issues/52586 and confirm it describes your case?

beesofts:
in my point of view, https://github.com/symfony/symfony/pull/51791/files did the job

PR #52131, opened 2023-10-18
[HttpKernel] Fix `RequestPayloadValueResolver` handling error with no ExpectedTypes

Jeroeny:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

The `NotNormalizableValueException` errors in `PartialDenormalizationException->getErrors()` have a `->getExpectedTypes()` method that returns `array|null`.  The `RequestPayloadValueResolver` assumes this is always an array.

When the returned value is `null`, an Exception is thrown with `implode(): Argument #1 ($pieces) must be of type array, string given`. 

Also the exception message in case of an empty array was: `This value should be of type .`.

#SymfonyHackday

### DIFF
--- a/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php
+++ b/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php
@@ -108,11 +108,15 @@ class RequestPayloadValueResolver implements ValueResolverInterface, EventSubscr
                 } catch (PartialDenormalizationException $e) {
                     $trans = $this->translator ? $this->translator->trans(...) : fn ($m, $p) => strtr($m, $p);
                     foreach ($e->getErrors() as $error) {
-                        $parameters = ['{{ type }}' => implode('|', $error->getExpectedTypes())];
+                        $parameters = [];
+                        $template = 'This value was of an unexpected type.';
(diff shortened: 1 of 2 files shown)


maxhelias (approved):

chalasr (approved):

chalasr:
Could you please rebase?

nicolas-grekas:
Thank you @Jeroeny.

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php
```diff
@@ -108,11 +108,15 @@ public function onKernelControllerArguments(ControllerArgumentsEvent $event): vo
                 } catch (PartialDenormalizationException $e) {
                     $trans = $this->translator ? $this->translator->trans(...) : fn ($m, $p) => strtr($m, $p);
                     foreach ($e->getErrors() as $error) {
-                        $parameters = ['{{ type }}' => implode('|', $error->getExpectedTypes())];
+                        $parameters = [];
+                        $template = 'This value was of an unexpected type.';
```
nicolas-grekas, line 112: Note that this message is not in translation maps. Dunno if it matters. Does it?

PR #52134, opened 2023-10-18
[HttpKernel] Add option to map empty data with `MapQueryString` and `MapRequestPayload`

Jeroeny:
| Q             | A
| ------------- | ---
| Branch?       | 8.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

When `#[MapQueryString]` or `#[MapRequestPayload]` is used on a nullable/default-valued parameter and the query string or request body is empty, the resolver short-circuits and returns `null` without ever calling the serializer. This prevents custom denormalizers from constructing the object.

This PR adds `bool $mapWhenEmpty = false` to both attributes. When `true`, the resolver passes `[]` to `denormalize()` even when no data is present, giving custom denormalizers a chance to populate the DTO.

```php
public function __construct(
    #[MapRequestPayload(mapWhenEmpty: true)] SearchFilters $filters,
) {}
```

**Use case:** a DTO where some fields come from the request and others are injected by a custom denormalizer (e.g. from the security context or session):

```php
class SearchFilters {
    public function __construct(
        public ?string $keyword = null,  // from query string
        public int $userId = 0,          // set by a custom denormalizer
    ) {}
}
```

Without `mapWhenEmpty`, an empty query string yields `null`, the denormalizer never runs. With `mapWhenEmpty: true`, denormalization proceeds and the custom denormalizer can populate `$userId`.

### DIFF
--- a/src/Symfony/Component/HttpKernel/Attribute/MapQueryString.php
+++ b/src/Symfony/Component/HttpKernel/Attribute/MapQueryString.php
@@ -41,6 +41,7 @@ class MapQueryString extends ValueResolver
         string $resolver = RequestPayloadValueResolver::class,
         public readonly int $validationFailedStatusCode = Response::HTTP_NOT_FOUND,
         public readonly ?string $key = null,
+        public bool $mapWhenEmpty = false,
     ) {
         parent::__construct($resolver);
     }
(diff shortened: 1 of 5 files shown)


Victor-Truhanovich (approved):
It is excellent, necessary and useful pull request

Victor-Truhanovich:
@nicolas-grekas May I know when this PR will be accepted and merged?

nicolas-grekas (approved):
I rebased on 8.1 and renamed to mapWhenEmpty + updated PR description

Jeroeny:
Thanks :+1:

nicolas-grekas:
Thank you @Jeroeny.

PR #52137, opened 2023-10-18
[FrameworkBundle][Routing] Remove remaining deprecations

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT

Follows https://github.com/symfony/symfony/pull/51483, I took `Annotation*Loader` from 7.0 to update `Attribute*Loader` here with latest changes.


nicolas-grekas:
Thank you @alexandre-daubois.

src/Symfony/Bundle/FrameworkBundle/Routing/AttributeRouteControllerLoader.php
```diff
@@ -25,10 +25,8 @@ class AttributeRouteControllerLoader extends AttributeClassLoader
 {
     /**
      * Configures the _controller default parameter of a given Route instance.
-     *
-     * @return void
      */
-    protected function configureRoute(Route $route, \ReflectionClass $class, \ReflectionMethod $method, object $annot)
+    protected function configureRoute(Route $route, \ReflectionClass $class, \ReflectionMethod $method, object $annot): void
```
jderusse, line 29: This makes SymfonyBundle 6.4 incompatible with SymfonyRouter 7.0

> PHP Fatal error:  Declaration of Symfony\Bundle\FrameworkBundle\Routing\AttributeRouteControllerLoader::configureRoute(Symfony\Component\Routing\Route $route, ReflectionClass $class, ReflectionMethod $method, object $annot) must be compatible with Symfony\Component\Routing\Loader\AttributeClassLoader::configureRoute(Symfony\Component\Routing\Route $route, ReflectionClass $class, ReflectionMethod $method, object $annot): void in /home/runner/work/aws/aws/vendor/symfony/framework-bundle/Routing/AttributeRouteControllerLoader.php on line 31

alexandre-daubois: What's the best way to mitigate this? Return types were added nearly everywhere but all constraints have been kept to `^6.4|^7.0`.

Should this return type be removed until the next occasion we can add it?

jderusse: I see 2 options:

- make Bundle require `^7.0`
- adds `: void` in `Bundle:7.0` and in `Route:8.0`

I think the first is better. 
ping @nicolas-grekas

nicolas-grekas: I'm going to remove the added return type in 7.0. That's what we did with the previous name, but I forgot this part when renaming.

PR #47416, opened 2022-08-28
[Console][FrameworkBundle][HttpKernel][WebProfilerBundle] Enable profiling commands

HeahDude:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #45241
| License       | MIT
| Doc PR        | ~

TLDR;

I've shown a POC of this feature at the Symfony Live Paris last April to some of the core team members (ping @nicolas-grekas, @stof, @lyrixx, @chalasr, @GromNaN).
I propose here a new work from scratch addressing the comments I already got and based on Javier's profiler redesign (#47148).
Reviews should better be done by commits.

Summary
---------
This PR aims to leverage the profiler and its collectors by using a `VirtualRequestStack` to aggregate data on virtual requests.
Such requests are obfuscated by default to avoid side effects.
It can feel like a hack... or a pragmatic way to get, without much complexity, tons of useful feedback on what's going on during console execution, from basic info about the command, time/memory metrics, to every existing features already available in HTTP context: events, message dispatching, http requests, emails, serialization, validation, cache, database queries... and so on, all that just out of the box!

Previous work
--------------

There were some work to extract the Profiler logic in a dedicated component, that proved to require a lot of complexity and BC breaks in the API:
* #10374
* #14809 (see https://github.com/symfony/symfony/pull/14809#issuecomment-313340589)
* #20502

Screenshots
------------
For now I've focused only on the functional parts.

<details><summary>Search view</summary>
Screenshot 2022-08-28 at 11 29 25 PM
</details>
<details><summary>Command panel</summary>
Screenshot 2022-08-28 at 11 30 54 PM
Screenshot 2022-08-28 at 11 31 08 PM
Screenshot 2022-08-28 at 11 31 21 PM
Screenshot 2022-08-28 at 11 31 34 PM

If the command is signal-able the following panel will be available:
Screenshot 2022-08-28 at 11 31 46 PM

If sub commands are run using `$this->getApplication()->run()` sub profiles will be shown as for requests:
Screenshot 2022-08-28 at 11 31 56 PM
</details>

The server tab is the same as in the request panel.

<details><summary>Performance panel</summary>
Screenshot 2022-08-28 at 11 32 23 PM
Screenshot 2022-08-28 at 11 32 32 PM
</details>
<details>
<summary>Failing command</summary>
The exception panel is shown by default as for requests:
[...]

PR #52138, opened 2023-10-18
[WebProfilerBundle] UI tweaks for the command profiler

javiereguiluz:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT

This PR proposes some minor UI tweaks for the amazing feature contributed by @HeahDude in #47416.

## First change

I propose to not display the `HTTP / Commands` toggle in the header of all pages. I don't think this should be an option with quick permanent access from all profiler pages.

My proposal is to move the toggle to the search sidebar:

search-before-after

And here in action:
http-commands-toggle

## Second change

In my opinion, the current header of command profiles looks too similar to HTTP profiles:

### Before / Light

command-profile-before-light

### Before / Dark

command-profile-before-dark

-----

I propose to differentiate them a bit more and use the well-known "fake terminals" used on Symfony website and docs:

### After / Light

command-profile-after-light

### After / Dark

command-profile-after-dark

-----

The "fake terminals" look different depending on your OS. See this image (from top to bottom: macOS, Windows, Linux)

terminal-os

### DIFF
--- a/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/_command_summary.html.twig
+++ b/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/_command_summary.html.twig
@@ -2,26 +2,32 @@
 {% set interrupted = command_collector is same as false ? null : command_collector.interruptedBySignal %}
 {% set css_class = status_code == 113 or interrupted is not null ? 'status-warning' : status_code > 0 ? 'status-error' : 'status-success' %}
 
-<div class="status {{ css_class }}">
+<div class="terminal status {{ css_class }}">
     <div class="container">
-        {% if status_code > 0 %}
-            <p class="status-error-details">
-                <span class="icon">{{ source('@WebProfiler/Icon/alert-circle.svg') }}</span>
-                <span class="status-response-status-code">Error ({{ status_code }})</span>
-            </p>
-        {% endif %}
-
         <h2>
             <span class="status-request-method">
                 {{ profile.method|upper }}
             </span>
 
-            {{ profile.url|length < 160 ? profile.url : profile.url[:160] ~ '…' }}
+            <span class="status-command">
+                {{ profile.url }}
+            </span>
         </h2>
 
         <dl class="metadata">
-            {% if status_code == 0 or interrupted %}
-                <span class="status-response-status-code">{{ interrupted is null ? 'Success' : 'Interrupted by signal: ' ~ interrupted }}</span>
+            {% if interrupted %}
+                <span class="status-response-status-code">Interrupted</span>
+                <dt>Signal</dt>
+                <dd class="status-response-status-text">{{ interrupted }}</dd>
+
+                <dt>Exit code</dt>
+                <dd class="status-response-status-text">{{ status_code }}</dd>
+            {% elseif status_code == 0 %}
+                <span class="status-response-status-code">Success</span>
+            {% elseif status_code > 0 %}
+                <span class="status-response-status-code">Error</span>
+                <dt>Exit code</dt>
+                <dd class="status-response-status-text"><span class="status-response-status-code">{{ status_code }}</span></dd>
             {% endif %}
 
             {% if request_collector.requestserver.has('SYMFONY_CLI_BINARY_NAME') %}
--- a/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/header.html.twig
+++ b/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/header.html.twig
@@ -1,21 +1,6 @@
 <div id="header">
     <h1><a href="{{ path('_profiler_home') }}">{{ source('@WebProfiler/Icon/symfony.svg') }} Symfony Profiler</a></h1>
 
-    <div class="sf-tabs" data-processed="true">
-        <div class="tab-navigation">
-            <span class="tab-control{{ 'request' == profile_type ? ' active' }}">
-                <a href="{{ path('_profiler_search_results', {token: 'empty', limit: 10}) }}">
-                    HTTP Requests
-                </a>
-            </span>
-            <span class="tab-control{{ 'command' == profile_type ? ' active' }}">
-                <a href="{{ path('_profiler_search_results', {token: 'empty', limit: 10, type: 'command'}) }}">
-                    Console Commands
-                </a>
-            </span>
-        </div>
-    </div>
-
     <div class="search">
         <form method="get" action="https://symfony.com/search" target="_blank">
             <div class="form-row">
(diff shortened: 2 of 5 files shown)


src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/profiler.css.twig
```diff
@@ -1258,6 +1271,102 @@ tr.status-warning td {
     margin: 0 1.5em 0 0;
 }
 
+#summary .terminal {
```
stof, line 1274: Should we really restrict the `.terminal` styles to children of `#summary` ? Why not making the component reusable in other parts of the profiler by removing the ancestor restriction ?

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/profiler.css.twig
```diff
@@ -1258,6 +1271,102 @@ tr.status-warning td {
...
+    box-shadow: 0 0 0 2px rgba(255, 255, 255, 0.3), 20px 0 0 2px rgba(255, 255, 255, 0.3), 40px 0 0 2px rgba(255, 255, 255, 0.3);
+    content: '';
+    display: block;
+    height: 7px;
+    left: 8px;
+    position: absolute;
+    top: -18px;
+    width: 7px;
+}
+.windows #summary .terminal:before {
```
stof: this should use `::before` to match the modern CSS spec and not the legacy alias (our usage of custom properties makes the profiler UI incompatible with browsers that are so old that they don't even support `::before`).

`::before` is a pseudo-element not a pseudo-class. And as of CSS 3, pseudo-elements use `::` as syntax (`before` and `after` were the only pseudo-elements defined in CSS 2 and so the only ones being supported through a legacy alias using the pseudo-class syntax)

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/profiler.css.twig
```diff
@@ -1258,6 +1271,102 @@ tr.status-warning td {
...
+    color: var(--black);
+    margin-right: 1.5em;
+    padding: 1px 5px;
+    text-transform: uppercase;
+}
+#summary .terminal.status-error,
+#summary .terminal.status-error .status-response-status-text {
+    color: var(--terminal-error-color);
+}
+.macos #summary .terminal:before {
```
stof: As the display depend on the OS, could you provide screenshots of the 3 different renderings ?

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/_command_summary.html.twig
```diff
@@ -20,8 +13,17 @@
         </h2>
 
         <dl class="metadata">
-            {% if status_code == 0 or interrupted %}
-                <span class="status-response-status-code">{{ interrupted is null ? 'Success' : 'Interrupted by signal: ' ~ interrupted }}</span>
+            {% if status_code == 0 %}
+                {% if interrupted %}
```
HeahDude: Command can be interrupted with an exit code != 0.

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/header.html.twig
```diff
@@ -1,21 +1,6 @@
 <div id="header">
     <h1><a href="{{ path('_profiler_home') }}">{{ source('@WebProfiler/Icon/symfony.svg') }} Symfony Profiler</a></h1>
 
-    <div class="sf-tabs" data-processed="true">
```
HeahDude, line 4: We may add more types in the future, hopefully consumed messages in 7.1.
There may be more space here to do so than in the search panel.

HeahDude (commented):
Thanks for this PR :)

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/profiler.css.twig
javiereguiluz: I tried doing that ... but terminal styles are not applied because we use the `#summary` selector in other styles and that gives them priority.

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/profiler.css.twig
javiereguiluz: Updated. Thanks for the explanation.

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/profiler.css.twig
javiereguiluz: Good idea. I've updated the description to add that screenshot.

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/_command_summary.html.twig
javiereguiluz: I've updated the logic. I hope I got it right this time. Thanks.

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/header.html.twig
javiereguiluz: I see. We could quickly move the toggle to show it above search results:

profile-type-toggle

Do you prefer to do that now or in the near future when we add more profile types?

OskarStark: I would keep it in the search bar for now. If we add Messenger, we can still think about a better solution

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/_command_summary.html.twig
HeahDude: It may be worth adding the exit code when interrupted. WDYT?

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/header.html.twig
HeahDude: I'd rather put it in a place that won't change to keep the UI consistent in future versions.

OskarStark: Agreed, but it should not be located in the content area. We can put it where proposed by Javier and find another kind of display if needed

javiereguiluz: I've moved the toggle to the top of the search results page:

search-by-profile-type

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/_command_summary.html.twig
javiereguiluz: Yes, let's do it. This is how it looks:

interrupted-exit-code

HeahDude (approved):
This is great thanks! 😻

fabpot (approved):

fabpot:
Thank you @javiereguiluz.

PR #52142, opened 2023-10-18
[7.0] Cleanup legacy code paths

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT


src/Symfony/Component/Routing/Loader/AttributeFileLoader.php
```diff
@@ -30,7 +30,7 @@ class AttributeFileLoader extends FileLoader
     public function __construct(FileLocatorInterface $locator, AttributeClassLoader $loader)
```
nicolas-grekas: Do we really need this class in 7.0 BTW?

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
```diff
@@ -91,8 +90,6 @@
 
 abstract class FrameworkExtensionTestCase extends TestCase
 {
-    use ExpectDeprecationTrait;
```
OskarStark, line 94: This would need to be readied once we deprecate sth. in Symfony 7, right?

src/Symfony/Component/HttpKernel/UriSigner.php
```diff
@@ -1,27 +0,0 @@
...
-
-namespace Symfony\Component\HttpKernel;
-
-use Symfony\Component\HttpFoundation\UriSigner as HttpFoundationUriSigner;
-
-trigger_deprecation('symfony/http-kernel', '6.4', 'The "%s" class is deprecated, use "%s" instead.', UriSigner::class, HttpFoundationUriSigner::class);
-
-class_exists(HttpFoundationUriSigner::class);
-
-if (false) {
```
OskarStark, line 20: Shouldn't e keep this class and remove HttpFOundationUriSigner instead?

src/Symfony/Component/Serializer/Annotation/Context.php
```diff
@@ -33,14 +33,14 @@ public function __construct(
         string|array $groups = [],
```
OskarStark: The namespace is still `...Component\Serializer\Annotation`. Worth deprecating and move to `Attribute`?

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
nicolas-grekas: indeed

src/Symfony/Component/Serializer/Annotation/Context.php
nicolas-grekas: I don't think so.

PR #52150, opened 2023-10-18
[Validator] Added missing translations for Romanian language for Validator component #51901

### LINKED ISSUE #51901: Missing translations for Romanian (ro)
Hello,

There are some missing translation for Romanian (ro). These should be added in branch 5.4. 
See https://github.com/symfony/symfony/issues/38710 for details and [this page](https://symfony-translations.nyholm.tech/#pr) for an example.

These are the files that should be updated: 
- [src/Symfony/Component/Validator/Resources/translations/validators.ro.xlf](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Validator/Resources/translations/validators.ro.xlf)

> [!NOTE]
> If you want to work on this issue, add a comment to assign it to yourself and let others know that this is already taken.

vetali:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #51901 
| License       | MIT

This PR adds missing translations for Romanian language for Validator component

### DIFF
--- a/src/Symfony/Component/Validator/Resources/translations/validators.ro.xlf
+++ b/src/Symfony/Component/Validator/Resources/translations/validators.ro.xlf
@@ -398,6 +398,30 @@
                 <source>The value of the netmask should be between {{ min }} and {{ max }}.</source>
                 <target>Valoarea netmask-ului trebuie sa fie intre {{ min }} si {{ max }}.</target>
             </trans-unit>
+            <trans-unit id="104">
+                <source>The filename is too long. It should have {{ filename_max_length }} character or less.|The filename is too long. It should have {{ filename_max_length }} characters or less.</source>
+                <target>Denumirea fișierului este prea lungă. Ea trebuie să conțină {{ filename_max_length }} caractere sau mai puține.|Denumirea fișierului este prea lungă. Ea trebuie să conțină {{ filename_max_length }} caractere sau mai puține.</target>
+            </trans-unit>
+            <trans-unit id="105">
+                <source>The password strength is too low. Please use a stronger password.</source>
+                <target>Complexitatea parolei este prea mică. Vă rugăm să folosiți o parolă mai puternică.</target>
+            </trans-unit>
+            <trans-unit id="106">
+                <source>This value contains characters that are not allowed by the current restriction-level.</source>
+                <target>Această valoare conține caractere care nu sunt premise de nivelul de restricționare curent.</target>
+            </trans-unit>
+            <trans-unit id="107">
+                <source>Using invisible characters is not allowed.</source>
+                <target>Folosirea caracterelor invizibile nu este permisă.</target>
+            </trans-unit>
+            <trans-unit id="108">
+                <source>Mixing numbers from different scripts is not allowed.</source>
+                <target>Combinarea numerelor din diferite script-uri nu este permisă.</target>
+            </trans-unit>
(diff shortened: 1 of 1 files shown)


src/Symfony/Component/Validator/Resources/translations/validators.ro.xlf
```diff
@@ -398,6 +398,30 @@
...
+                <source>The filename is too long. It should have {{ filename_max_length }} character or less.|The filename is too long. It should have {{ filename_max_length }} characters or less.</source>
+                <target>Denumirea fișierului este prea lungă. Ea trebuie să conțină {{ filename_max_length }} caractere sau mai puține.|Denumirea fișierului este prea lungă. Ea trebuie să conțină {{ filename_max_length }} caractere sau mai puține.</target>
+            </trans-unit>
+            <trans-unit id="105">
+                <source>The password strength is too low. Please use a stronger password.</source>
+                <target>Complexitatea parolei este prea mică. Vă rugăm să folosiți o parolă mai puternică.</target>
+            </trans-unit>
+            <trans-unit id="106">
+                <source>This value contains characters that are not allowed by the current restriction-level.</source>
+                <target>Această valoare conține caractere care nu sunt premise de nivelul de restricționare curent.</target>
```
nicolas-grekas, line 411: `nivelul curent de restricție`?

src/Symfony/Component/Validator/Resources/translations/validators.ro.xlf
```diff
@@ -398,6 +398,30 @@
...
+            </trans-unit>
+            <trans-unit id="107">
+                <source>Using invisible characters is not allowed.</source>
+                <target>Folosirea caracterelor invizibile nu este permisă.</target>
+            </trans-unit>
+            <trans-unit id="108">
+                <source>Mixing numbers from different scripts is not allowed.</source>
+                <target>Combinarea numerelor din diferite stript-uri nu este permisă.</target>
```
nicolas-grekas: typo: `script-uri`

src/Symfony/Component/Validator/Resources/translations/validators.ro.xlf
vetali: Thank you for the feedback @nicolas-grekas. I still believe `restricționare` makes more sense in the context, since `restricție` means more like a rule and `restricționare` means the process of enforcing that rule. Either place of the word `curent` sounds correct to me, so I wasn't sure which one to choose. Can we reach a middle ground by having `nivelul curent de restricționare` ?

nicolas-grekas: you're the native speaker :)

nicolas-grekas:
Thank you @vetali.

src/Symfony/Component/Validator/Resources/translations/validators.ro.xlf
vetali: This one was the trickiest to translate for me. I checked other translation files and I saw many use the `script` word as the root. The problem is if we use the word `scripturi` and it is read with an accent on `u` instead of `i`, the meaning will become `religious writing`, which I tried to avoid. Maybe it makes more sense to translate the word as it was done for Russian? So it will become `scenarii` ?

PR #52155, opened 2023-10-18
[WebProfilerBundle] Rewrite to avoid using request word when nothing found

alamirault:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #... 
| License       | MIT

When profiling commands, we don't have request.

image

Others collectors templates don't use "Request" when there is no item


HeahDude (approved):
Good catch, thanks!

fabpot (approved):

fabpot:
Good catch, thanks @alamirault.

PR #52156, opened 2023-10-18
[AssetMapper] Optimize memory usage (load only "compilable" assets)

smnandre:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| License       | MIT

A bit late for 6.4... but this PR will have a sensible impact on performances.

Currently if the assets directory contains 500 images, they are all loaded in memory during the compile command. 
Their contents are also serialized/unserialized and stored in cache.

This PR allow to load/store in memory/cache only the assets supported by the compilers (JS/CSS files today)

### DIFF
--- a/src/Symfony/Component/AssetMapper/Factory/MappedAssetFactory.php
+++ b/src/Symfony/Component/AssetMapper/Factory/MappedAssetFactory.php
@@ -37,13 +37,12 @@ class MappedAssetFactory implements MappedAssetFactoryInterface
 
     public function createMappedAsset(string $logicalPath, string $sourcePath): ?MappedAsset
     {
-        if (\in_array($logicalPath, $this->assetsBeingCreated, true)) {
+        if (isset($this->assetsBeingCreated[$logicalPath])) {
             throw new CircularAssetsException(sprintf('Circular reference detected while creating asset for "%s": "%s".', $logicalPath, implode(' -> ', $this->assetsBeingCreated).' -> '.$logicalPath));
         }
+        $this->assetsBeingCreated[$logicalPath] = $logicalPath;
 
         if (!isset($this->assetsCache[$logicalPath])) {
-            $this->assetsBeingCreated[] = $logicalPath;
-
             $isVendor = $this->isVendor($sourcePath);
             $asset = new MappedAsset($logicalPath, $sourcePath, $this->assetsPathResolver->resolvePublicPath($logicalPath), isVendor: $isVendor);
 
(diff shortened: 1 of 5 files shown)


src/Symfony/Component/AssetMapper/Factory/MappedAssetFactory.php
```diff
@@ -83,28 +82,38 @@ private function getDigest(MappedAsset $asset): array
             return [$matches[1], true];
         }
 
+        // Use the compiled content if any
+        if (null !== $content = $this->compileContent($asset)) {
+            return [
+                hash('xxh128', $content),
+                false,
+            ];
```
nicolas-grekas:
```suggestion
            return [hash('xxh128', $content), false];
```

nicolas-grekas (approved):

src/Symfony/Component/AssetMapper/MappedAsset.php
```diff
@@ -126,4 +124,9 @@ public function addJavaScriptImport(JavaScriptImport $import): void
     {
         $this->javaScriptImports[] = $import;
     }
+
+    public function getMappedContent(): ?string
```
stof: No need to create a getter. This is a public readonly property.

src/Symfony/Component/AssetMapper/MappedAsset.php
```diff
@@ -24,7 +24,7 @@ final class MappedAsset
     public readonly string $publicPath;
     public readonly string $publicPathWithoutDigest;
     public readonly string $publicExtension;
-    public readonly string $content;
+    public readonly ?string $content;
```
weaverryan, line 31: It might deserve some PHPdoc above:

```php
/**
 * The final content of this asset, or null if the source file's content should be used.
 */
```

weaverryan (approved):
This is fantastic - I can't believe how small the diff was to fix this!

src/Symfony/Component/AssetMapper/MappedAsset.php
smnandre: Sorry / thanks  i originally kept content separated from mappedContent. Fixed

smnandre:
Should i do something special for the Psalm alert ? The code is safe and "hash_file" will return a string, but Psalm has no way to know the file existence has just been tested.... and i don't want to check it twice in the code.

nicolas-grekas:
Nope, you can ignore psalm, we didn't surrender to robots (yet) ;)

smnandre:
So all is ok for me :)

fabpot (approved):

fabpot:
Thank you @smnandre.

PR #52162, opened 2023-10-19
add ext-pcntl as dependency

stefanniedermaier:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | -
| License       | MIT

If PCNTL is not available on the system an exception will be htrown in ConsumeMessagesCommand.php, because it requires SIGTERM and SIGINT
Adding the dependency makes it clearer and more understandable which requirements are needed.

### DIFF
--- a/src/Symfony/Component/Messenger/composer.json
+++ b/src/Symfony/Component/Messenger/composer.json
@@ -17,6 +17,7 @@
     ],
     "require": {
         "php": ">=8.1",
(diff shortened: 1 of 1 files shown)


wouterj:
This requirement is a mistake, which has been fixed and will be included in the next release. See also #51791

PR #52166, opened 2023-10-19
[HtmlSanitizer] Add support for sanitizing unlimited length of HTML document

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       |
| License       | MIT

### DIFF
--- a/src/Symfony/Component/HtmlSanitizer/CHANGELOG.md
+++ b/src/Symfony/Component/HtmlSanitizer/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+ * Add support for sanitizing unlimited length of HTML document
+
 6.1
 ---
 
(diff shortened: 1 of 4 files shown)


nicolas-grekas (approved):

nicolas-grekas:
What about `withMaxInputLength(PHP_INT_MAX)`? Isn't this risky practice?

lyrixx:
> What about `withMaxInputLength(PHP_INT_MAX)`? Isn't this risky practice?

I did that in my app (still on 6.3 😅) But If feels more natural to disable the feature completely if you don't want it.

tgalopin (approved):
Given that there is a default at 20 000, I think anyone choosing to change it probably had a a look at why it exists and therefore understands at least a bit the consequences. I'm fine disabling it completely IMO

stof:
@tgalopin there is no phpdoc at all in `HtmlSanitizerConfig` that would explain the reason to have a limit though.

nicolas-grekas:
Thank you @lyrixx.

PR #52172, opened 2023-10-19
[Serializer] Fix denormalizing empty string into `object|null` parameter

Jeroeny:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

> In XML and CSV all basic datatypes are represented as strings

The `AbstractObjectNormalizer` handles this for bool, int and float types. But if a parameter is typed `Object|null`, a serialization and then deserialization will fail. 

This will throw: 

```php
final class Xml
{
    public function __construct(public Uuid|null $element = null)
    {
    }
}

$test = new Xml(null);
$serialized = $this->serializer->serialize($test, XmlEncoder::FORMAT);
$deserialized = $this->serializer->deserialize($serialized, Xml::class, XmlEncoder::FORMAT);
```

```
  [Symfony\Component\Serializer\Exception\NotNormalizableValueException]       
  The data is not a valid "Symfony\Component\Uid\Uuid" string representation.  
```

Reproducer: https://github.com/Jeroeny/reproduce/blob/xmlnull/src/Test.php

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
@@ -467,8 +467,10 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
     {
         $expectedTypes = [];
         $isUnionType = \count($types) > 1;
+        $e = null;
         $extraAttributesException = null;
         $missingConstructorArgumentException = null;
+        $isNullable = false;
         foreach ($types as $type) {
             if (null === $data && $type->isNullable()) {
                 return null;
@@ -491,18 +493,22 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
                 // In XML and CSV all basic datatypes are represented as strings, it is e.g. not possible to determine,
                 // if a value is meant to be a string, float, int or a boolean value from the serialized representation.
                 // That's why we have to transform the values, if one of these non-string basic datatypes is expected.
+                $builtinType = $type->getBuiltinType();
                 if (\is_string($data) && (XmlEncoder::FORMAT === $format || CsvEncoder::FORMAT === $format)) {
                     if ('' === $data) {
-                        if (Type::BUILTIN_TYPE_ARRAY === $builtinType = $type->getBuiltinType()) {
+                        if (Type::BUILTIN_TYPE_ARRAY === $builtinType) {
                             return [];
                         }
 
-                        if ($type->isNullable() && \in_array($builtinType, [Type::BUILTIN_TYPE_BOOL, Type::BUILTIN_TYPE_INT, Type::BUILTIN_TYPE_FLOAT], true)) {
-                            return null;
+                        if (Type::BUILTIN_TYPE_STRING === $builtinType) {
+                            return '';
                         }
+
+                        // Don't return null yet because Object-types that come first may accept empty-string too
+                        $isNullable = $isNullable ?: $type->isNullable();
                     }
 
-                    switch ($builtinType ?? $type->getBuiltinType()) {
+                    switch ($builtinType) {
                         case Type::BUILTIN_TYPE_BOOL:
                             // according to https://www.w3.org/TR/xmlschema-2/#boolean, valid representations are "false", "true", "0" and "1"
                             if ('false' === $data || '0' === $data) {
@@ -603,11 +609,11 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
                     return $data;
                 }
             } catch (NotNormalizableValueException $e) {
-                if (!$isUnionType) {
+                if (!$isUnionType && !$isNullable) {
                     throw $e;
                 }
             } catch (ExtraAttributesException $e) {
-                if (!$isUnionType) {
+                if (!$isUnionType && !$isNullable) {
                     throw $e;
                 }
 
@@ -615,7 +621,7 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
                     $extraAttributesException = $e;
                 }
             } catch (MissingConstructorArgumentsException $e) {
-                if (!$isUnionType) {
+                if (!$isUnionType && !$isNullable) {
                     throw $e;
                 }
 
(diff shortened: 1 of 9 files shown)


OskarStark:
Looks like a bugfix to me, in this case 🎯 `6.3`, but let's wait for advice of @nicolas-grekas

nicolas-grekas:
This would need tests for sure. But I'm no expert in the Serializer component so I'd appreciate if someone more familiar with it could review.

Jeroeny:
I've added tests. I also discovered that `$builtinType` isn't reset in the for-loop in some cases. Which means a `Object|bool` type wouldn't hit this switch-case: https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php#L471 (The added tests also cover this).

Jeroeny:
Another interesting failure scenario:

```php
// this fails
$this->serializer->denormalize(['element' => '0'], Example::class, context: [AbstractObjectNormalizer::DISABLE_TYPE_ENFORCEMENT => true]);

class Example {
    public function __construct(public Args|int $element) {}
}

class Args {
    public function __construct(public Uuid $nested, private string $test)
}

//  [Symfony\Component\Serializer\Exception\MissingConstructorArgumentsException]
//  Cannot create an instance of "App\Args" from serialized data because its constructor requires the following parameters to be present : "$nested", "$test".

// whereas this returns an instance succesfully
$this->serializer->denormalize(['element' => '0'], Test::class, context: [AbstractObjectNormalizer::DISABLE_TYPE_ENFORCEMENT => true]);

class Test {
    public function __construct(public Uuid|int $element)
}
```

Because the first denormalization throws `MissingConstructorArgumentsException`, which goes before the return on `DISABLE_TYPE_ENFORCEMENT`.
The second throws `NotNormalizableValueException`, but is caught and will return `0` because of `DISABLE_TYPE_ENFORCEMENT`.

I'm not sure what the reasoning was here or if this is actually not intended.

It could be fixed by adding `$hasTypeMismatch = $builtinType !== Type::BUILTIN_TYPE_OBJECT` here:
https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php#L565. To establish that it did not throw for at least one of the `$types`.  
And the later using that to return before `MissingConstructorArgumentsException`. (Or always return before that exception when DISABLE_TYPE_ENFORCEMENT is true?)

src/Symfony/Component/Serializer/Tests/SerializerTest.php
```diff
@@ -842,6 +847,44 @@ public function testTrueBuiltInTypes()
         $this->assertEquals(new TrueBuiltInDummy(), $actual);
     }
 
+    public function testDeserializeUntypedFormat()
```
mtarld, line 826: Is it really needed to test the whole deserialization instead of only the denormalization?

If yes, the test might be located in the `AbstractObjectNormalizerTest`

src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
```diff
@@ -456,9 +456,10 @@ private function validateAndDenormalize(array $types, string $currentClass, stri
                 // In XML and CSV all basic datatypes are represented as strings, it is e.g. not possible to determine,
                 // if a value is meant to be a string, float, int or a boolean value from the serialized representation.
                 // That's why we have to transform the values, if one of these non-string basic datatypes is expected.
+                $builtinType = $type->getBuiltinType();
                 if (\is_string($data) && (XmlEncoder::FORMAT === $format || CsvEncoder::FORMAT === $format)) {
                     if ('' === $data) {
-                        if (Type::BUILTIN_TYPE_ARRAY === $builtinType = $type->getBuiltinType()) {
+                        if (Type::BUILTIN_TYPE_ARRAY === $builtinType) {
                             return [];
                         }
```
mtarld, line 502: What about only replacing
```php
if ($type->isNullable() && \in_array($builtinType, [Type::BUILTIN_TYPE_BOOL, Type::BUILTIN_TYPE_INT, Type::BUILTIN_TYPE_FLOAT], true)) {
    return null;
}
```
by
```php
if ($type->isNullable() && Type::BUILTIN_TYPE_STRING === $builtinType) {
    return '';
}

if ($type->isNullable()) {
    return null;
}
```
instead?

This will simplify a lot the patch.

mtarld (changes requested):

nicolas-grekas:
Let's resume this one? Is this a bugfix as it's documented in the PR description? If yes can you please target the lowest affected branch?

Jeroeny:
Applied feedback and targeting `6.3`

nicolas-grekas:
Isn't 5.4 affected also?

Jeroeny:
Eh yes, I was looking into the wrong part when checking that branch. Now based on 5.4.

nicolas-grekas:
Parse error on PHP 7.2 ;)

src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
Jeroeny: That's not possible, because it needs to `return null` only when other options fail (like an object that accepts empty string as denormalization data). I've added `testDenormalizeUntypedStringObject` to test against this.

Jeroeny:
Serializer tests are ok now

src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
```diff
@@ -623,6 +632,10 @@ private function validateAndDenormalize(array $types, string $currentClass, stri
             throw $missingConstructorArgumentException;
         }
 
+        if (!$isUnionType && isset($e)) {
```
nicolas-grekas: better set $e = null before than using isset

Jeroeny: Agreed, fixed

nicolas-grekas:
(rebase needed ;) )

nicolas-grekas:
Thank you @Jeroeny.

PR #52179, opened 2023-10-19
[CI] Add step to verify `symfony/deprecation-contracts` requirements

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT

Screenshot from 2023-10-19 19-23-01

### DIFF
--- a/.github/workflows/package-tests.yml
+++ b/.github/workflows/package-tests.yml
@@ -101,3 +101,43 @@ jobs:
           done
 
           exit $ok
+      - name: Verify symfony/deprecation-contracts requirements
+        run: |
+          set +e
+
+          ok=0
+          json='${{ steps.find-packages.outputs.packages }}'
+          for package in $(echo "${json}" | jq -r '.[] | @base64'); do
+              _jq() {
+                echo ${package} | base64 --decode | jq -r ${1}
+              }
+
+              NAME=$(_jq '.name')
+              if [[ $NAME = 'symfony/deprecation-contracts' || $NAME = 'symfony/contracts' ]]; then
+                continue
+              fi
+
+              echo "::group::$NAME"
+              DIR=$(_jq '.directory')
+              localExit=0
+              grep -rq 'trigger_deprecation(' --include=*.php --exclude-dir=Tests/ --exclude-dir=Bridge/ $DIR
+              triggersDeprecation=$?
+              if [[ $triggersDeprecation -eq 0 && $(_jq '.requires_deprecation_contracts') == false ]]; then
+                errorMessage="::error::$NAME does not require symfony/deprecation-contracts but triggers at least one deprecation"
+                localExit=1
+              elif [[ $triggersDeprecation -eq 1 && $(_jq '.requires_deprecation_contracts') == true ]]; then
+                errorMessage="::error::$NAME requires symfony/deprecation-contracts but does not trigger any deprecation"
+                localExit=1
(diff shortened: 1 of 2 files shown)


.github/workflows/package-tests.yml
```diff
@@ -100,4 +100,41 @@ jobs:
...
+          ok=0
+
+          json='${{ steps.find-packages.outputs.packages }}'
+          for package in $(echo "${json}" | jq -r '.[] | @base64'); do
+              _jq() {
+                echo ${package} | base64 --decode | jq -r ${1}
+              }
+
+              NAME=$(_jq '.name')
+              if [[ $NAME = 'symfony/deprecation-contracts' || $NAME = 'symfony/contracts' ]]; then
```
fancyweb, line 116: Those two don't require it but they always have it

.github/workflows/package-tests.yml
```diff
@@ -100,4 +100,41 @@ jobs:
...
+                continue
+              fi
+
+              echo "::group::$NAME"
+              DIR=$(_jq '.directory')
+              localExit=0
+              grep -rq 'trigger_deprecation(' --include=*.php --exclude-dir=Tests/ --exclude-dir=Bridge/ $DIR
```
fancyweb, line 123: We exclude tests and component bridges that are in the root component code but that are released as separate packages

.github/workflows/package-tests.yml
```diff
@@ -100,4 +100,41 @@ jobs:
...
+              triggersDeprecation=$?
+              if [[ $triggersDeprecation -eq 0 && $(_jq '.requires_deprecation_contracts') == false ]]; then
```
fancyweb, line 125: if there is at least one trigger_deprecation( match and the dep is not required -> fail

.github/workflows/package-tests.yml
```diff
@@ -100,4 +100,41 @@ jobs:
...
+                localExit=1
+              elif [[ $triggersDeprecation -eq 1 && $(_jq '.requires_deprecation_contracts') == true ]]; then
```
fancyweb, line 128: if there no trigger_deprecation( match and the dep is required -> fail

fancyweb:
There will probably be false positives one day but it currently works for every maintained branch.

.github/workflows/package-tests.yml
```diff
@@ -100,4 +100,41 @@ jobs:
...
+                localExit=1
+              elif [[ $triggersDeprecation -ne 0 && $triggersDeprecation -ne 1 ]]; then
+                echo "::error::grep failed"
+                exit 2
+              fi
+
+              ok=$(( $localExit || $ok ))
+              echo ::endgroup::
+              if [ $localExit -ne 0 ]; then
+                echo "::error::$NAME failed"
```
stof: I would rather echo separate messages in the 2 failure cases, which allow to be more precise about why they fail (missing requirement vs useless requirement)

OskarStark:
Maybe a dumb question, but couldn't we use composer-unused for this task? 

cc @icanhazstring

stof:
@OskarStark not for the part about missing the necessary requirement (which is the most important part of this PR)

stof (approved):

OskarStark:
Ah yes, you are right the other way around would be [ComposerRequireChecker ](https://github.com/maglnet/ComposerRequireChecker)

PR #52198, opened 2023-10-20
[String] New locale aware casing methods

### LINKED ISSUE #52161: [String] New `LocaleString` class
### Description

Working with Greek texts we got the report that when turning the text into uppercase, we should strip the accents from the input.
Note that this is not simply turning the string into ascii.

Example:
In Greek:
`άδικος, κείμενο, ίριδα → ΑΔΙΚΟΣ, ΚΕΙΜΕΝΟ, ΙΡΙΔΑ`

In Turkish:
`ISTANBUL` -> `İSTANBUL`

It doesn't seem like we can do that with the current String component, is it maybe an idea to add a new LocaleString class to this component?

Source:
https://icu.unicode.org/design/case/greek-upper
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleUpperCase

### Very quick PoC

```php
$string       = "άδικος, κείμενο, ίριδα";
$localeString = new class ($string, "el") extends UnicodeString {
    public function __construct(string $string = '', private readonly ?string $languageCode = null)
    {
        parent::__construct($string);
    }

    public function upper(): static
    {
        $str         = clone $this;
        $str->string = \Transliterator::create($this->languageCode . "-Upper")->transliterate($str->string);
 
        return $str;
    }
};

echo $localeString->upper(); // ΑΔΙΚΟΣ, ΚΕΙΜΕΝΟ, ΙΡΙΔΑ
```

to be updated to implement more methods, and to reuse Transliterator instances.
And I needed to make the languageCode constructor var nullable because of the parent class, default this to the php default?

bram123:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #52161
| License       | MIT

Adds new localeUpper/localeLower/localeTitle methods to the AbstractUnicodeString class
- To change the string case according to locale-specific case mappings

Code examples:
```php
$string = new UnicodeString('άδικος');
echo $string->upper(), PHP_EOL; // ΆΔΙΚΟΣ
echo $string->localeUpper('el'), PHP_EOL; // ΑΔΙΚΟΣ

$string = new UnicodeString('ijssel');
echo $string->title(), PHP_EOL; // Ijssel
echo $string->localeTitle('nl'), PHP_EOL; // IJssel

$string = new UnicodeString('İSTANBUL');
echo $string->lower(), PHP_EOL; // i̇stanbul (LATIN SMALL LETTER I COMBINING DOT ABOVE)
echo $string->localeLower('tr'), PHP_EOL; // istanbul (LATIN SMALL LETTER I)
```

### DIFF
--- a/src/Symfony/Component/String/AbstractUnicodeString.php
+++ b/src/Symfony/Component/String/AbstractUnicodeString.php
@@ -220,6 +220,21 @@ abstract class AbstractUnicodeString extends AbstractString
         return $str;
     }
 
+    /**
+     * @param string $locale In the format language_region (e.g. tr_TR)
+     */
+    public function localeLower(string $locale): static
+    {
+        if (null !== $transliterator = $this->getLocaleTransliterator($locale, 'Lower')) {
+            $str = clone $this;
+            $str->string = $transliterator->transliterate($str->string);
+
+            return $str;
+        }
+
+        return $this->lower();
+    }
+
     public function match(string $regexp, int $flags = 0, int $offset = 0): array
     {
         $match = ((\PREG_PATTERN_ORDER | \PREG_SET_ORDER) & $flags) ? 'preg_match_all' : 'preg_match';
@@ -363,6 +378,21 @@ abstract class AbstractUnicodeString extends AbstractString
         return $str;
     }
 
+    /**
+     * @param string $locale In the format language_region (e.g. tr_TR)
+     */
+    public function localeTitle(string $locale): static
+    {
+        if (null !== $transliterator = $this->getLocaleTransliterator($locale, 'Title')) {
+            $str = clone $this;
+            $str->string = $transliterator->transliterate($str->string);
+
+            return $str;
+        }
+
+        return $this->title();
+    }
+
     public function trim(string $chars = " \t\n\r\0\x0B\x0C\u{A0}\u{FEFF}"): static
     {
         if (" \t\n\r\0\x0B\x0C\u{A0}\u{FEFF}" !== $chars && !preg_match('//u', $chars)) {
@@ -448,6 +478,21 @@ abstract class AbstractUnicodeString extends AbstractString
         $str->string = mb_strtoupper($str->string, 'UTF-8');
 
         return $str;
+    }
+
+    /**
+     * @param string $locale In the format language_region (e.g. tr_TR)
+     */
+    public function localeUpper(string $locale): static
+    {
+        if (null !== $transliterator = $this->getLocaleTransliterator($locale, 'Upper')) {
+            $str = clone $this;
+            $str->string = $transliterator->transliterate($str->string);
+
+            return $str;
+        }
+
+        return $this->upper();
     }
 
     public function width(bool $ignoreAnsiDecoration = true): int
@@ -586,5 +631,34 @@ abstract class AbstractUnicodeString extends AbstractString
         }
 
         return $width;
+    }
+
+    private function getLocaleTransliterator(string $locale, string $id): ?\Transliterator
+    {
+        $rule = $locale.'-'.$id;
+        if (\array_key_exists($rule, self::$transliterators)) {
+            return self::$transliterators[$rule];
+        }
+
+        if (null !== $transliterator = self::$transliterators[$rule] = \Transliterator::create($rule)) {
+            return $transliterator;
+        }
+
+        // Try to find a parent locale (nl_BE -> nl)
+        if (false === $i = strpos($locale, '_')) {
+            return null;
+        }
+
+        $parentRule = substr_replace($locale, '-'.$id, $i);
+
+        // Parent locale was already cached, return and store as current locale
+        if (\array_key_exists($parentRule, self::$transliterators)) {
+            return self::$transliterators[$rule] = self::$transliterators[$parentRule];
+        }
+
+        // Create transliterator based on parent locale and cache the result on both initial and parent locale values
+        $transliterator = \Transliterator::create($parentRule);
+
+        return self::$transliterators[$rule] = self::$transliterators[$parentRule] = $transliterator;
     }
 }
--- a/src/Symfony/Component/String/CHANGELOG.md
+++ b/src/Symfony/Component/String/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Add `localeLower()`, `localeUpper()`, `localeTitle()` methods to `AbstractUnicodeString`
+
 6.2
 ---
 
--- a/src/Symfony/Component/String/Tests/AbstractUnicodeTestCase.php
+++ b/src/Symfony/Component/String/Tests/AbstractUnicodeTestCase.php
@@ -50,6 +50,48 @@ END'],
         $this->assertSame('Dieser Wert sollte grOEsser oder gleich', (string) $s->ascii([$rule]));
     }
 
+    /**
+     * @dataProvider provideLocaleLower
+     *
+     * @requires extension intl
+     */
+    public function testLocaleLower(string $locale, string $expected, string $origin)
+    {
+        $instance = static::createFromString($origin)->localeLower($locale);
+
+        $this->assertNotSame(static::createFromString($origin), $instance);
+        $this->assertEquals(static::createFromString($expected), $instance);
+        $this->assertSame($expected, (string) $instance);
+    }
+
+    /**
+     * @dataProvider provideLocaleUpper
+     *
+     * @requires extension intl
+     */
+    public function testLocaleUpper(string $locale, string $expected, string $origin)
+    {
+        $instance = static::createFromString($origin)->localeUpper($locale);
+
+        $this->assertNotSame(static::createFromString($origin), $instance);
+        $this->assertEquals(static::createFromString($expected), $instance);
+        $this->assertSame($expected, (string) $instance);
+    }
+
+    /**
+     * @dataProvider provideLocaleTitle
+     *
+     * @requires extension intl
+     */
+    public function testLocaleTitle(string $locale, string $expected, string $origin)
+    {
+        $instance = static::createFromString($origin)->localeTitle($locale);
+
+        $this->assertNotSame(static::createFromString($origin), $instance);
+        $this->assertEquals(static::createFromString($expected), $instance);
+        $this->assertSame($expected, (string) $instance);
+    }
+
     public function provideCreateFromCodePoint(): array
     {
         return [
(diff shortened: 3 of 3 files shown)


src/Symfony/Component/String/AbstractUnicodeString.php
```diff
@@ -587,4 +623,21 @@ private function wcswidth(string $string): int
 
         return $width;
     }
+
+    private function getLocaleTransliterator(string $locale, string $id): ?\Transliterator
+    {
+        $rule = $locale.'-'.$id;
+        if (null !== $transliterator = self::$transliterators[$rule] ??= \Transliterator::create($rule)) {
```
nicolas-grekas: we need to cache when create() returns `null`
this means we should use `\array_key_exists()` instead (same at the end of the method)

bram123: You're right, fixes with latest commit:
```php
$string->localeTitle("nl");
$string->localeTitle("nl_BE");
$string->localeTitle("nl_BE");
```
NL: creates and caches NL transliterator
NL_BE: Unable to create NL_BE, uses NL from cache, caches as NL_BE
NL_BE: Uses NL_BE from cache

src/Symfony/Component/String/AbstractUnicodeString.php
```diff
@@ -587,4 +623,34 @@ private function wcswidth(string $string): int
...
     }
+
+    private function getLocaleTransliterator(string $locale, string $id): ?\Transliterator
+    {
+        $rule = $locale.'-'.$id;
+        if (\array_key_exists($rule, self::$transliterators)) {
+            return self::$transliterators[$rule];
+        }
+
+        if (null !== $transliterator = self::$transliterators[$rule] ??= \Transliterator::create($rule)) {
```
nicolas-grekas:
```suggestion
        if (null !== $transliterator = self::$transliterators[$rule] = \Transliterator::create($rule)) {
```

src/Symfony/Component/String/AbstractUnicodeString.php
```diff
@@ -587,4 +623,34 @@ private function wcswidth(string $string): int
...
+            return $transliterator;
+        }
+
+        // Try to find a parent locale (nl_BE -> nl)
+        if (false === $str = strrchr($locale, '_')) {
```
nicolas-grekas:
```suggestion
        if (false === $i = strpos($locale, '_')) {
```

and adapt the rest

nicolas-grekas (commented):
Can you please also add a line to the changelog of the component?

src/Symfony/Component/String/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+  * Add `localeLower`, `localeUpper`, `localeTitle` methods to `AbstractUnicodeString`
```
nicolas-grekas:
```suggestion
 * Add `localeLower()`, `localeUpper()`, `localeTitle()` methods to `AbstractUnicodeString`
```

src/Symfony/Component/String/AbstractUnicodeString.php
```diff
@@ -587,4 +623,34 @@ private function wcswidth(string $string): int
...
+            return $transliterator;
+        }
+
+        // Try to find a parent locale (nl_BE -> nl)
+        if (false === $i = strpos($locale, '_')) {
+            return null;
+        }
+
+        $locale = substr($locale, 0, $i);
+        $parentRule = $locale.'-'.$id;
```
nicolas-grekas:
```suggestion
        $parentRule = substr_replace($locale, '-'.$id, $i);
```

nicolas-grekas (approved):
Two minor nitpicking and GTM, thanks.

src/Symfony/Component/String/AbstractUnicodeString.php
```diff
@@ -450,6 +474,18 @@ public function upper(): static
         return $str;
     }
 
+    public function localeUpper(string $locale): static
+    {
+        if (null !== $transliterator = $this->getLocaleTransliterator($locale, 'Upper')) {
```
smnandre, line 488: If i'm not mistaken, there is only a few locales concerned... maybe we can add a small test for the hot path ?

src/Symfony/Component/String/AbstractUnicodeString.php
```diff
@@ -587,4 +623,33 @@ private function wcswidth(string $string): int
...
     }
+
+    private function getLocaleTransliterator(string $locale, string $id): ?\Transliterator
+    {
+        $rule = $locale.'-'.$id;
+        if (\array_key_exists($rule, self::$transliterators)) {
+            return self::$transliterators[$rule];
+        }
+
+        if (null !== $transliterator = self::$transliterators[$rule] = \Transliterator::create($rule)) {
```
smnandre, line 643: Maybe check getLastErrorMessage ?

smnandre:
Re-copying this there in case we need it in a following PR or to write doc : 

--> official [ICU list of special casings](https://github.com/unicode-org/icu/blob/main/icu4c/source/data/unidata/SpecialCasing.txt)

src/Symfony/Component/String/AbstractUnicodeString.php
```diff
@@ -220,6 +220,18 @@ public function lower(): static
...
+    public function localeLower(string $locale): static
+    {
+        if (null !== $transliterator = $this->getLocaleTransliterator($locale, 'Lower')) {
+            $str = clone $this;
+            $str->string = $transliterator->transliterate($str->string);
+
+            return $str;
+        }
+
+        return $this->lower();
```
derrabus, line 235: I wonder if this silent fallback is the right thing to do. If someone calls `localeLower()`, they explicitly opted into using a locale-aware transliterator. If that transliterator is not available (e.g. because of a typo in the passed locale), how would they be made aware of that?

nicolas-grekas: Not all locales have special rules, the fallback is the correct behavior for them.

derrabus: I see. So we cannot tell apart a mistyped locate from a valid one without a special transliterator?

bram123: This is the same behaviour as JavaScript's String.toLocaleUpperCase().

const city = 'istanbul';
console.log(city.toLocaleUpperCase('foobar')); // "ISTANBUL"

smnandre: The transliterator use IDs composed with source and filters

> The source specifier describes the characters or strings that the transform will modify. The target specifier describes the result of the modification. If the source is not given, then the source is “Any”, the set of all characters. Source and Target specifiers can be [Script IDs](http://www.unicode.org/cldr/utility/properties.jsp#Script) (long like “Latin” or short like “Latn”), [Unicode language Identifiers](http://www.unicode.org/reports/tr35/#Unicode_Language_and_Locale_Identifiers) (like fr, en_US, or zh_Hant), or special tags (like Any or Hex)

https://unicode-org.github.io/icu/userguide/transforms/general/#basic-ids 

We could validate the identifier/locale passed in parameter (with [locale_parse](https://www.php.net/manual/fr/locale.parselocale.php) ?) but indeed, nothing guarantee the matching transliterator does exist.

derrabus: All right, makes sense. Let's keep it this way. Thank you for the explanation!

src/Symfony/Component/String/AbstractUnicodeString.php
bram123: Do you mean: `error_get_last()`, `transliterator_get_error_message` (to get errors from Transliterator instance) or `intl_get_error_message` (to get errors for static calls to intl classes) ?
Not seeing these used by AsciiSlugger or the other Transliterator usages in UnicodeString btw.

smnandre: It was on the same topic than derrabus raised about errors and typos, and it seems we agreed to silent them. So let's forget this :)

src/Symfony/Component/String/AbstractUnicodeString.php
bram123: Yes, at the moment it only affects "TR", "AZ", "EL" and "NL" and these are all used in the unittest's dataprovider + I added "EN"/"EN_US" to show the difference. These tests have a 100% code coverage for the new methods.
Are there more testcases you want me to add?

smnandre: Oh i am sorry i must have badly express myself. My goal never was that you feel i "wanted" something. That was really just a question / suggestion.

What i meant was: knowing those special casing only concern a few langages, is the cost of trying to instanciate the transliterator something we should avoid ? (this is a geniune question, nothing more :) ) 

If so, i wondered if we would want to bypass it testing the given locale against a hard-coded set.

Something like: 
```php
    private function getLocaleTransliterator(string $locale, string $id): ?\Transliterator
    {

        // if locale not in special_locale then return null 

        $rule = $locale.'-'.$id;
        if (\array_key_exists($rule, self::$transliterators)) {
            return self::$transliterators[$rule];
        }
```

bram123: Ah I understand, I thought you meant a PHPUnit test for the hot path.
I'd vote against having a hardcoded set, sure now it's only TR/AZ/EL/NL, but we don't know if that stays the same in the future. And first checking Transliterator::listIDs, would break fetching NL-Title when passing nl_NL as locale.

btw, Transliterator::listIDs contains "el-Title", not "el_GR-Title", but when passing "el_GR-Title" it does return an instance.. However this doesn't work for "nl_NL-Title", that has to be "nl-Title".

With the static cache the impact is not that big aswell, and if your application is only in "EN" for example, there is no need to call this method at all.

Do you have an idea on how to pick this up?

smnandre: > Transliterator::listIDs contains "el-Title", not "el_GR-Title", but when passing "el_GR-Title" it does return an instance.. However this doesn't work for "nl_NL-Title", that has to be "nl-Title".

Well that could be something to specify in the documentation :) 

> sure now it's only TR/AZ/EL/NL, but we don't know if that stays the same in the future

That's why i was thinking at first to parse the offocial ICU docs (not on real time of course, once in a while to update the script)... but that is a story for another time, i do not want to slow you down (even if it seems so)

;)

bram123: With the current implementation of looking at input and also parent locale, it always works.
I'll add a comment to the new localeX methods, to describe the expected input `language_region`.

Parsing ICU and using that dataset: I'd have no idea how to begin that, glad it can be outside of this PR ;)

> i do not want to slow you down (even if it seems so)

No problem, it's all to make the implementation better (Y).

nicolas-grekas: There's one reason that could make @smnandre's proposal worth it: instantiation time. IIRC, transliterators can be slow to create. Can you have a look at this from this angle after doing some benchs please?

bram123: Quick benchmark
```php
$string = static::createFromString('istanbul');
$iterations = 1_000_000;

$startTime = microtime(true);
$i = $iterations;
while (--$i) {
    $string->upper();
}
$endTime = microtime(true);
echo "default: " . $endTime - $startTime, PHP_EOL;

$startTime = microtime(true);
$i = $iterations;
while (--$i) {
    $string->localeUpper('en_US');
}
$endTime = microtime(true);
echo "locale en_US: " . $endTime - $startTime, PHP_EOL;

$startTime = microtime(true);
$i = $iterations;
while (--$i) {
    $string->localeUpper('tr_TR');
}
$endTime = microtime(true);
echo "locale tr_TR: " . $endTime - $startTime, PHP_EOL;

$startTime = microtime(true);
$i = $iterations;
while (--$i) {
    $string->localeUpper($i);
}
$endTime = microtime(true);
echo "locale all unique: " . $endTime - $startTime, PHP_EOL;
```

Results in:
default: 1.517627954483
locale en_US: 2.7830729484558
locale tr_TR: 3.431088924408
locale all unique: 4.8467578887939

The results do fluctuate a bit btw, I've seen between 2.4 and 3.0 for "en_US".
Also, placing a `return null` right above the first Transliterator::create() still gives the same times for "en_US".

smnandre: This is something we can optimize in a following step.. but indeed there is a big memory difference :| 

I simplified the two options 

# Option One

```php
<?php

$word = 'étienne';
$locale = 'fr';

if (preg_match('/^(EL|AB|CD|EF|GH|IJ)/i', $locale)) {
    $transliterator = \Transliterator::create($locale.'-UPPER');
    echo 'A' . ($word = $transliterator->transliterate($word));
    return $word;
}

echo 'B' . ($word = mb_strtoupper($word));
return $word;
```

https://3v4l.org/FArNT/perf


### Option Two

```php
<?php

$word = 'étienne';
$locale = 'fr';

if (null !== $transliterator = Transliterator::create($locale.'-UPPER')) {
    echo 'A' . ($word = $transliterator->transliterate($word));
    return $word;
}

echo 'B' . ($word = mb_strtoupper($word));
return $word;
```

https://3v4l.org/NsKEe/perf

## Results

### One - "fr"

Capture d’écran 2023-10-23 à 14 44 07 

### One - "el" 

Capture d’écran 2023-10-23 à 14 44 19 

### Two - "fr" 

Capture d’écran 2023-10-23 à 14 44 27

### Two - "el" 

Capture d’écran 2023-10-23 à 14 44 42


I'm sincerly surprised, but there is a **massive difference** for the "FR" path (any langage with no transliterator available)


I do not have the time right now, but we should check with the transliterator_ids method

smnandre: But we could implement the optimizations in a following step (and maybe there are some to find in the current component) ... what do you think @nicolas-grekas  ?

nicolas-grekas: Sure! Note that this should be merged in 7.1 since 6.4/7.0 are closed now.

bram123: Just missed the window? That's too bad.
I don't see a 7.1 branch to base this PR on, what should I do?

nicolas-grekas: Just be patient - we'll resume merges for 7.1 at the end of November.

bram123: Alright, sorry. Thanks for the help.

yceruto (approved):

src/Symfony/Component/String/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
```
OskarStark:
```suggestion
7.1
```

OskarStark (approved):

fabpot (approved):

fabpot:
Thank you @bram123.

Issue #52210, opened 2023-10-20
Help Wanted: fix failure on WebProfilerBundle 5.4 loaded with higest deps


nicolas-grekas:
On 5.4, we have a failure to fix:
https://github.com/symfony/symfony/actions/runs/6590102626/job/17912208613#step:7:393

I can reproduce it by going into the `src/Symfony/Bundle/WebProfilerBundle` folder on branch 5.4, then running composer up and
`SYMFONY_DEPRECATIONS_HELPER=weak ../../../../phpunit --filter testPanelActionWithLatestToken$`

Help welcome to figure out the issue and fix it!

smnandre:
```php
1) Symfony\Bundle\WebProfilerBundle\Tests\Controller\ProfilerControllerTest::testPanelActionWithLatestToken
UnexpectedValueException: The profiler template "@WebProfiler/Collector/command.html.twig" for data collector "command" does not exist.
```

Two choices:
* either the command profiler template is missing [in this folder](https://github.com/symfony/symfony/tree/5.4/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector)
* either the definition of the collector must be deleted

```
  ".data_collector.command" => array:2 [
    0 => "command"
    1 => "@WebProfiler/Collector/command.html.twig"
  ]
```

nicolas-grekas:
The change should be done in 6.4 ideally /cc @HeahDude

HeahDude:
It seems `WebProfilerBundle` 5.4 cannot be compatible with `FrameworkBundle` 6.4, the latter being responsible for configuring collectors.

I see tow options:
- add a compiler pass in `WebProfilerBundle` 5.4 to remove new data collectors introduced by 6.x (I guess we will have the same error with the Workflow data collector if we fix this only for the command data collector).
- add a conflict rule in `FrameworkBundle` 6.4 to not allow `WebProfilerBundle` < 6.4.

I would go for the second solution.

nicolas-grekas:
Patching 5.4 to account for 6.4 would be wrong, so solution 2 would be the way. Ideally, I'd prefer FWB to not know about WPB. Is there no solution that doesn't involve adding any conflict line?

HeahDude:
See #52218.

HeahDude:
The conflict line was already there because I guess the problem was always there, adding collectors in FWB while adding templates in WebProfilerBundle.

HeahDude:
So we can close here I guess.

PR #52218, opened 2023-10-21
[FrameworkBundle] Add conflict with `WebProfilerBundle` < 6.4

HeahDude:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | #52210
| License       | MIT

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/composer.json
+++ b/src/Symfony/Bundle/FrameworkBundle/composer.json
@@ -100,7 +100,7 @@
         "symfony/twig-bridge": "<5.4",
         "symfony/twig-bundle": "<5.4",
         "symfony/validator": "<6.4",
-        "symfony/web-profiler-bundle": "<5.4",
+        "symfony/web-profiler-bundle": "<6.4",
         "symfony/workflow": "<6.4"
     },
     "autoload": {


nicolas-grekas:
Works for me, but can you think of any other way that wouldn't involve a conflict?

HeahDude:
The only other way I can think of would be to disallow FWB 6.4 by adding the conflict in WPB 5.4.

smnandre:
If the problem happen only with 2 datacollectors, would it be possible to add a specific conflict ?

FMB 6.4 has a conflict with worflow < 6.4, but allow console 5.4.. so i guess there won't be a problem with the new worflow profiler

What if we add a conflict on console < 6.4 on FMB ?

nicolas-grekas (approved):
On closer look, this is fine to me :rocket:

nicolas-grekas:
Thank you @HeahDude.

chalasr:
@smnandre That wouldn’t have been enough to fix the issue at hand as the missing template is both located in and rendered by WPB. See the failing build which does run with Console v6.4-dev

smnandre:
@nicolas-grekas @chalasr 

 I "fixed" it in another way..... by declaring the collector separately

--> Changes: https://github.com/symfony/symfony/compare/6.4...smnandre:symfony:fix/console-profiler 
--> Runs: https://github.com/smnandre/symfony/actions/runs/6599172839/job/17927922180 

But..... is it only because this one was merged ? (even if i restored the conflict as it was before ?)   😨

nicolas-grekas:
Send the PR! :)

PR #52222, opened 2023-10-21
[MonologBridge] Fix support for monolog 3.0

louismariegaborit:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no 
| Issues        | 
| License       | MIT

`handle` function in HandlerInterface supports only LogRecord argument in monolog 3.0.

### DIFF
--- a/src/Symfony/Bridge/Monolog/Command/ServerLogCommand.php
+++ b/src/Symfony/Bridge/Monolog/Command/ServerLogCommand.php
@@ -13,7 +13,9 @@ namespace Symfony\Bridge\Monolog\Command;
 
 use Monolog\Formatter\FormatterInterface;
 use Monolog\Handler\HandlerInterface;
+use Monolog\Level;
 use Monolog\Logger;
+use Monolog\LogRecord;
 use Symfony\Bridge\Monolog\Formatter\ConsoleFormatter;
 use Symfony\Bridge\Monolog\Handler\ConsoleHandler;
 use Symfony\Component\Console\Attribute\AsCommand;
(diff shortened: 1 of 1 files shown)


xabbuh:
Can you add a test to prevent future regressions?

louismariegaborit:
> Can you add a test to prevent future regressions?

I think it's difficult in this PR to add test because mock stream functions will cause a refactoring of the command.
I might try doing this in an other PR. What do you think ?

louismariegaborit:
I don't know how to fix psalm error because in monolog 2.x, LogRecord is an interface and psalm runs with this version.
Anyone have an idea?

src/Symfony/Bridge/Monolog/Command/ServerLogCommand.php
```diff
@@ -156,6 +158,16 @@ private function displayLog(OutputInterface $output, int $clientId, array $recor
         $logBlock = sprintf('<bg=%s> </>', self::BG_COLOR[$clientId % 8]);
         $output->write($logBlock);
 
+        if (Logger::API >= 3) {
+            $record = new LogRecord(
+                level: Level::fromValue($record['level']),
+                channel: $record['channel'],
+                context: $record['context']->getContext(),
+                datetime: $record['datetime'],
+                message: $record['message'],
```
nicolas-grekas: let's not use named arguments if possible

stof:
Don't care about Psalm in this case. This is an expected error due to supporting several versions of Monolog. Our psalm setup only reports *new* issues for each PR, so this won't both future PRs.

src/Symfony/Bridge/Monolog/Command/ServerLogCommand.php
louismariegaborit: Ok I did the change. But why don't use named arguments ?
Named arguments are already used for the LogRecord. https://github.com/symfony/symfony/blob/547485e99c2ce377211679674f18c90ccc7fb753/src/Symfony/Bridge/Monolog/Tests/RecordFactory.php#L25

derrabus (approved):

derrabus:
Good catch, thanks @louismariegaborit.

PR #52232, opened 2023-10-22
[Messenger] declare constructor argument as optional for backwards compatibility

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix https://github.com/symfony/symfony/pull/51468#issuecomment-1746200733
| License       | MIT


ste93cry:
Shouldn't a deprecation be triggered so that accepting a `null` value is not allowed anymore in the next major? This would make the behavior consistent with similar exceptions, e.g. `HandlerFailedException`

xabbuh:
We cannot add new deprecations in already released minor versions.

nicolas-grekas:
Thank you @xabbuh.

PR #52234, opened 2023-10-22
 add return type hints to EntityFactory

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

see https://github.com/EasyCorp/EasyAdminBundle/actions/runs/6603120076/job/17935909075

### DIFF
--- a/.github/expected-missing-return-types.diff
+++ b/.github/expected-missing-return-types.diff
@@ -151,30 +151,6 @@ diff --git a/src/Symfony/Bridge/Doctrine/DependencyInjection/CompilerPass/Regist
 +    public function process(ContainerBuilder $container): void
      {
          if (!$this->enabled($container)) {
-diff --git a/src/Symfony/Bridge/Doctrine/DependencyInjection/Security/UserProvider/EntityFactory.php b/src/Symfony/Bridge/Doctrine/DependencyInjection/Security/UserProvider/EntityFactory.php
---- a/src/Symfony/Bridge/Doctrine/DependencyInjection/Security/UserProvider/EntityFactory.php
-+++ b/src/Symfony/Bridge/Doctrine/DependencyInjection/Security/UserProvider/EntityFactory.php
-@@ -34,5 +34,5 @@ class EntityFactory implements UserProviderFactoryInterface
-      * @return void
-      */
--    public function create(ContainerBuilder $container, string $id, array $config)
-+    public function create(ContainerBuilder $container, string $id, array $config): void
-     {
-         $container
-@@ -47,5 +47,5 @@ class EntityFactory implements UserProviderFactoryInterface
-      * @return string
-      */
--    public function getKey()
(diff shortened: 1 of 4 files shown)


wouterj:
An alternative would be to add the native return type and document the BC break. We've done that for multiple other method implementations as well: #50890 Given `EntityFactory` is a DI factory class, the likelihood of anyone being hit by this BC break seems rather slim.

nicolas-grekas:
I like @wouterj's proposal!

derrabus:
Yes, let's do what @wouterj proposes. If someone complains about the BC break, we could still make the return type conditional using a trait, but I doubt that someone extends this class.

While we're at it, should we make `EntityFactory` final in 7.0? There's really no point in extending it: It only exposes the public methods from `UserProviderFactoryInterface`. If someone wants to extend the functionality of that class, a decorator would be the better way to do that.

xabbuh:
I have added the return types and also marked the class as final.

nicolas-grekas (approved):

nicolas-grekas (approved):

derrabus (approved):

smnandre:
Is this the same problem there ? 

> PHP Fatal error:  Declaration of Symfony\WebpackEncoreBundle\CacheWarmer\EntrypointCacheWarmer::doWarmUp($cacheDir, Symfony\Component\Cache\Adapter\ArrayAdapter $arrayAdapter): bool must be compatible with Symfony\Bundle\FrameworkBundle\CacheWarmer\AbstractPhpFileCacheWarmer::doWarmUp(string $cacheDir, Symfony\Component\Cache\Adapter\ArrayAdapter $arrayAdapter, ?string $buildDir = null): bool in /home/runner/work/ux/ux/src/Turbo/vendor/symfony/webpack-encore-bundle/src/CacheWarmer/EntrypointCacheWarmer.php on line 27

https://github.com/symfony/ux/issues/1223

If so, could you explain what are the steps to fix ? (i must admit i'm not yet fluent in "multi-php multi-version multi-bundle" troubles 😅

wouterj:
@smnandre that's a different issue: the first parameter `$cacheDir` must get the `string` type in the WebpackEncoreBundle. Adding types to parameters can be done safely (it's not a BC break), as a child can have a less specific parameter type than its parent).

smnandre:
Thank you @wouterj !

fabpot (approved):

fabpot:
Thank you @xabbuh.

derrabus:
> the first parameter `$cacheDir` must get the `string` type in the WebpackEncoreBundle

Actually, the issue is the new and optional third parameter. But, as @wouterj said, this is a different and unrelated problem. If you don't know how to solve it, please open a new issue or a dicussion.

PR #52235, opened 2023-10-22
[FrameworkBundle] Remove mailer and notification traits from KernelTestCase

BafS:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | not exactly (BC)
| Issues        | -
| License       | MIT

I propose to remove `MailerAssertionsTrait` and `NotificationAssertionsTrait` from the KernelTestCase;

 - It bloats the class for no reason
 - Mailer and notification components may not be used at all in an app, it's confusing to have asserts available that will not work
 - It's not possible to remove a trait from a class, yet it's easy to add it in a subclass if needed
 - It's not exactly composition over inheritance but the idea is similar (a better way would be to use `MailerAssert::*` to use composition instead of `self::*` with traits but that's another story)

Because of that we have our own implementation of KernelTestCase (just to remove those traits) which is a pity.

It's a BC BREAK and I'm not sure if I should target v6.4 or not. Let me know and if it's accepted I can update the correct CHANGELOG too.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php
@@ -24,9 +24,6 @@ use Symfony\Contracts\Service\ResetInterface;
  */
 abstract class KernelTestCase extends TestCase
 {
-    use MailerAssertionsTrait;
-    use NotificationAssertionsTrait;
-
     protected static ?string $class = null;
     protected static ?KernelInterface $kernel = null;
     protected static bool $booted = false;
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
We don't do hard BC breaks so this would need a deprecation layer first.
But I'm not convinced this is worth the trouble. A base test case is not something I would try to make that lean.

nicolas-grekas:
Closing as explained.

BafS:
Thank you, I still think we should be careful to not have god classes but PHPUnit itself is not great because it extends the `Assert` class. Hopefully it will change one day and we would favor composition (`Assert::*` instead of `self::*`).

Issue #52206, opened 2023-10-20
Help Wanted: Fix failures on appveyor


nicolas-grekas:
Branch 7.0 has failures that look legit on appveyor:

https://ci.appveyor.com/project/fabpot/symfony/builds/48330699?fullLog=true

Note that apppveyor tests Windows, but not only. There, we also test running the test suite without any extensions (e.g. to check that we added all `@requires` annotations in test, and we test on x86 (32-bit arch).

maxbeckers:
For the tests failed in the log with the date format:

>1) Symfony\Component\Form\Tests\Extension\Core\Type\BirthdayTypeTest::testSubmitFromSingleTextDateTime
null does not match expected type "object".
C:\projects\symfony\src\Symfony\Component\Form\Tests\Extension\Core\Type\DateTypeTest.php:108
>2) Symfony\Component\Form\Tests\Extension\Core\Type\BirthdayTypeTest::testSubmitFromSingleTextDateTimeImmutable
Failed asserting that null is an instance of class "DateTimeImmutable".
C:\projects\symfony\src\Symfony\Component\Form\Tests\Extension\Core\Type\DateTypeTest.php:130
>3) Symfony\Component\Form\Tests\Extension\Core\Type\BirthdayTypeTest::testSubmitFromSingleTextString
Failed asserting that null matches expected '2010-06-02'.
C:\projects\symfony\src\Symfony\Component\Form\Tests\Extension\Core\Type\DateTypeTest.php:153
>4) Symfony\Component\Form\Tests\Extension\Core\Type\BirthdayTypeTest::testSubmitFromSingleTextTimestamp
Failed asserting that null matches expected '1275436800'.
C:\projects\symfony\src\Symfony\Component\Form\Tests\Extension\Core\Type\DateTypeTest.php:177
>5) Symfony\Component\Form\Tests\Extension\Core\Type\BirthdayTypeTest::testSubmitFromSingleTextRaw
null does not match expected type "array".
C:\projects\symfony\src\Symfony\Component\Form\Tests\Extension\Core\Type\DateTypeTest.php:205
>6) Symfony\Component\Form\Tests\Extension\Core\Type\DateTypeTest::testSubmitFromSingleTextDateTime
null does not match expected type "object".
C:\projects\symfony\src\Symfony\Component\Form\Tests\Extension\Core\Type\DateTypeTest.php:108
[...]

PR #52245, opened 2023-10-23
[Form] skip tests that do not work with ICU 71.1

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Form/Tests/Extension/Core/Type/DateTypeTest.php
+++ b/src/Symfony/Component/Form/Tests/Extension/Core/Type/DateTypeTest.php
@@ -14,6 +14,7 @@ namespace Symfony\Component\Form\Tests\Extension\Core\Type;
 use Symfony\Component\Form\ChoiceList\View\ChoiceView;
 use Symfony\Component\Form\FormError;
 use Symfony\Component\Form\FormInterface;
+use Symfony\Component\Intl\Intl;
 use Symfony\Component\Intl\Util\IntlTestHelper;
 use Symfony\Component\OptionsResolver\Exception\InvalidOptionsException;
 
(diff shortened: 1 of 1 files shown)


maxbeckers:
@xabbuh does this help you https://github.com/symfony/symfony/issues/52206#issuecomment-1776590581 ?

xabbuh:
@maxbeckers Kind of, I suspect the ICU version used to be the cause here. At least that the only relevant difference that I can spot compared to 6.4 (see #52251).

Which version do you have locally? And can you also try different versions?

maxbeckers:
@xabbuh you're right.

When i run it with ``70.1``, it works. When I use ``71.1``, it fails.

xabbuh:
73.2 works for me too.

maxbeckers:
I currently only have the two versions I tried for testing.

src/Symfony/Component/Form/Tests/Extension/Core/Type/DateTypeTest.php
```diff
@@ -105,6 +107,17 @@ public function testSubmitFromSingleTextDateTime()
...
+        if (!$form->isValid()) {
+            var_dump(Intl::getIcuVersion());
+            /** @var FormError $error */
+            foreach ($form->getErrors(true) as $error) {
+                var_dump($error->getMessage());
+                var_dump($error->getCause()->getMessage());
+                var_dump($error->getCause()->getInvalidMessage());
+                var_dump($error->getCause()->getInvalidMessageParameters());
+            }
+        }
```
maxbeckers: The output of this looks like that if it helps:
```
string(4) "71.1"
string(26) "Please enter a valid date."
string(84) "Unable to reverse value for property path "date": Date parsing failed: U_PARSE_ERROR"
NULL
array(0) {
}
```

PR #52253, opened 2023-10-23
[Messenger] Add call to `gc_collect_cycles()` after each message is handled

jwage:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| License       | MIT

I have been working on improving memory usage in my messenger worker queues and I had been chasing what I thought was a memory leak, but I haven't been able to find it. In my troubleshooting, I pushed a change to production which simply logs the result of `gc_status()` after each message is handled, and I noticed that it was taking a very long time for the automatic garbage collection to run. Additionally, each time it would finally run, the PHP algorithm internally would extend the threshold, making it take even longer to run the next time. So the number of objects in memory waiting to be collected would keep growing and the memory per process would grow. In my application, it is important for the memory for each worker process to stay relatively flat so that I can know how many worker processes I can fit on each worker server.

I added a call to `gc_collect_cycles()` in my application after each message is handled, and it seems to have improved things for me. I suppose it is still possible I have a memory leak, but I haven't been able to find it.

Here is a screenshot of memory usage from one of my worker servers after I deployed that change.

Screenshot 2023-10-23 at 12 32 46 PM

In my application, it takes < 1ms to call `gc_collect_cycles()` after each message is handled and that is satisfactory for me in order to keep memory usage lower.

If this is accepted, should we make it optionally enabled?

### DIFF
--- a/src/Symfony/Component/Messenger/Worker.php
+++ b/src/Symfony/Component/Messenger/Worker.php
@@ -118,6 +118,8 @@ class Worker
                 // this should prevent multiple lower priority receivers from
                 // blocking too long before the higher priority are checked
                 if ($envelopeHandled) {
+                    gc_collect_cycles();
+
                     break;
                 }
             }
--- a/src/Symfony/Component/Messenger/Tests/WorkerTest.php
+++ b/src/Symfony/Component/Messenger/Tests/WorkerTest.php
@@ -578,6 +578,25 @@ class WorkerTest extends TestCase
 
         $this->assertSame($expectedMessages, $handler->processedMessages);
     }
+
+    public function testGcCollectCyclesIsCalledOnMessageHandle()
+    {
+        $apiMessage = new DummyMessage('API');
+
+        $receiver = new DummyReceiver([[new Envelope($apiMessage)]]);
+
+        $bus = $this->createMock(MessageBusInterface::class);
+
+        $dispatcher = new EventDispatcher();
+        $dispatcher->addSubscriber(new StopWorkerOnMessageLimitListener(1));
+
+        $worker = new Worker(['transport' => $receiver], $bus, $dispatcher);
+        $worker->run();
+
+        $gcStatus = gc_status();
+
+        $this->assertGreaterThan(0, $gcStatus['runs']);
+    }
 }
 
 class DummyReceiver implements ReceiverInterface


src/Symfony/Component/Messenger/Worker.php
```diff
@@ -109,6 +109,8 @@ public function run(array $options = []): void
                     $this->handleMessage($envelope, $transportName);
                     $this->eventDispatcher?->dispatch(new WorkerRunningEvent($this, false));
 
+                    gc_collect_cycles();
```
nicolas-grekas: Did you try move the call after this foreach? does that produce the same benefit for your workload?

jwage: In my app, I used the `Symfony\Component\Messenger\Event\WorkerMessageHandledEvent` event to run `gc_collect_cycles()` and log the result of `gc_status()` so it is after each message is handled. I don't have a way to test it outside of the foreach in production with a real workload without modifying Symfony source code (I think).

I suppose it could be similar if I put it outside the foreach, but under heavy workload, we could get many `$envelopes` back from the receiver? Not sure how many could be returned from the receiver. In that case, many messages could be consumed without calling `gc_collect_cycles()` and the memory could go higher than wanted.

I was chatting with @weaverryan about this and the question came up if this should be configurable to enable/disable and even configure where it gets called, after each message inside the foreach or outside. Not sure if that is necessary or not.

nicolas-grekas: The issue I see with putting this here is for batch processing: `gc_collect_cycles()` is not a cheap call. Inserting it between each messages of a batch might kill the perf for batches.
Putting it just after might be a better compromise.

goetas:
We have almost the same middleware on all our projects, it brings the same memory improvement. My guess is that the number of roots the php garbage collector waits is too high for most of the situations

lyrixx:
Can you share the impact on CPU? 
Usually RAM is much cheaper than CPU

src/Symfony/Component/Messenger/Worker.php
jwage: @nicolas-grekas that makes sense. Calling it after each batch makes more sense. Is that the only scenario where the receiver returns multiple envelopes at a time?

nicolas-grekas: I don't remember precisely :) But calling this a few times per X messages should be good enough. If experience proves we need to be more aggressive, then we might adjust.

jwage:
> Can you share the impact on CPU? 
> 
> Usually RAM is much cheaper than CPU

We haven't seen any noticeable difference in the CPU load graphs but our worker servers aren't under heavy load. It takes around 1ms each time we call gc_collect_cycles().

image

src/Symfony/Component/Messenger/Worker.php
jwage: 👍 I will adjust the location of the call.

jwage:
> We have almost the same middleware on all our projects, it brings the same memory improvement. My guess is that the number of roots the php garbage collector waits is too high for most of the situations

Yes, that was my conclusion as well. And it is my understanding that PHP will adjust that threshold higher or lower each time gc runs depending on how much was cleaned up. If not much was cleaned up, then it will wait longer. For some reason in my case, the threshold gets extended higher and higher and it takes longer and longer for gc to run on its own.

src/Symfony/Component/Messenger/Worker.php
94noni: i dont know the faisability and if it is relevant
but cant a stamp be used here, if the message is GCStamped then launch GC?

nicolas-grekas (approved):

lyrixx:
> And it is my understanding that PHP will adjust that threshold higher or lower each time gc runs depending on how much was cleaned up

PHP run GC when the number of object is > 10K (it's configurable, but only during compile time)

jwage:
> > And it is my understanding that PHP will adjust that threshold higher or lower each time gc runs depending on how much was cleaned up
> 
> PHP run GC when the number of object is > 10K (it's configurable, but only during compile time)

:+1: and the threshold was getting extended to 20k, then 30k, etc. each time the automatic gc would run.

stof:
We discussed that on Slack and I investigated the way the GC works in PHP.

Manual runs of the GC (triggered by the `gc_collect_cycles()` function don't adjust the threshold at all.
Automatic runs are adjusting the threshold based on the number of collected cycles: if less than 100, the threshold gets increased by 10000 (it starts at 10001) otherwise it gets decreased by 10000 if higher than the initial one.

But this means that if an automatic GC run cannot collect objects (because you have lots of alive objects at that time), it will increase the threshold and that threshold won't ever decrease again if those objects are cleaned by the manual run.

PhilETaylor:
> I added a call to gc_collect_cycles() in my application after each message is handled,

I have applied this PR (along with an echo statement to tell me when that line is run) to my dev stack, and started my 5 workers (We have over 1000 workers in production), 

Without running any messages, just sitting there with a worker started, the gc_collect_cycles is running over and over and over again - continuously every few seconds (evidenced by logging out to the terminal) 

This is not "after each message is handled" like you stated, its near continuous... and obviously if there are more than worker on the same server, that "continuous" is x the n number of workers also. 

If you needed it  "after each message is handled" surely then it should be listened with `WorkerMessageHandledEvent`? 

Or am I missing something?

jwage:
@PhilETaylor initially, it was inside the `foreach ($envelopes as $envelope) {` loop, but in the conversation here https://github.com/symfony/symfony/pull/52253#discussion_r1369096871, we discussed moving it outside of the foreach and that was done in this commit 06b071344384a82d2a1a9f5f1d9bcfa8dbe35e6a. We are missing some additional logic around the `gc_collect_cycles()` call now to only run it when the receivers return envelopes. I will work on another commit to address this.

Edit: Try this commit 0084c455fa642d57e584a6de3d0095ecfee7a1a0

PhilETaylor:
@jwage Thanks - the 0084c455fa642d57e584a6de3d0095ecfee7a1a0 fixes the "continuous" calling. I'm deploying this to "production" now, if you dont hear from me, it didn't break haha.

jwage:
> @jwage Thanks - the [0084c45](https://github.com/symfony/symfony/commit/0084c455fa642d57e584a6de3d0095ecfee7a1a0) fixes the "continuous" calling. I'm deploying this to "production" now, if you dont hear from me, it didn't break haha.

@PhilETaylor 😮 you live on the bleeding edge :) :)

PhilETaylor:
@jwage nothing like running a SaaS with 1000s of paying users on decent code :) Symfony makes me powerful 
Current status:
ScreenShot-2023-10-24-23 30 23

edit: Nothing broke :-) but no huge decrease in memory for me either... dip is the deployment


ScreenShot-2023-10-24-23 32 56

jwage:
@PhilETaylor it is my understanding that if your code doesn't have any cyclic references, then the `gc_collect_cycles()` won't have to do anything because the objects will be destroyed immediately when the refcount === 0. So if your code doesn't have any cyclic references, then you won't see any benefit. Did you see any difference in your cpu/load graphs?

PhilETaylor:
The CPU and Load graphs have their normal spike during deployment but other than that the loads before and after are pretty consistent with each other with no major decrease and no major increase

jwage:
> The CPU and Load graphs have their normal spike during deployment but other than that the loads before and after are pretty consistent with each other with no major decrease and no major increase

Thanks. I am not expecting a drop, but there was some concern that it could increase cpu usage since we're calling gc more often than we were before. Thanks for testing!

nicolas-grekas:
Thank you @jwage.

PR #52213, opened 2023-10-21
[AssetMapper] Normalizing path in test to fix / \ comparison in Windows

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | Fixes the failing AppVeyor test
| License       | MIT

https://ci.appveyor.com/project/fabpot/symfony/builds/48330693#L1059

PR #52258, opened 2023-10-24
[AssetMapper][Validator] Fix tests on Windows

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Part of https://github.com/symfony/symfony/issues/52206
| License       | MIT

### DIFF
--- a/src/Symfony/Component/AssetMapper/Tests/CompiledAssetMapperConfigReaderTest.php
+++ b/src/Symfony/Component/AssetMapper/Tests/CompiledAssetMapperConfigReaderTest.php
@@ -14,6 +14,7 @@ namespace Symfony\Component\AssetMapper\Tests;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\AssetMapper\CompiledAssetMapperConfigReader;
 use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Filesystem\Path;
 
 class CompiledAssetMapperConfigReaderTest extends TestCase
 {
@@ -54,8 +55,9 @@ class CompiledAssetMapperConfigReaderTest extends TestCase
     public function testSaveConfig()
     {
         $reader = new CompiledAssetMapperConfigReader($this->writableRoot);
-        $this->assertEquals($this->writableRoot.\DIRECTORY_SEPARATOR.'foo.json', realpath($reader->saveConfig('foo.json', ['foo' => 'bar'])));
-        $this->assertEquals(['foo' => 'bar'], json_decode(file_get_contents($this->writableRoot.'/foo.json'), true));
+
+        $this->assertSame(Path::normalize($this->writableRoot.\DIRECTORY_SEPARATOR.'foo.json'), Path::normalize(realpath($reader->saveConfig('foo.json', ['foo' => 'bar']))));
+        $this->assertSame(['foo' => 'bar'], json_decode(file_get_contents($this->writableRoot.'/foo.json'), true));
     }
 
     public function testRemoveConfig()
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/Validator/Tests/Constraints/FileTest.php
```diff
@@ -26,7 +26,7 @@ public function testMaxSize($maxSize, $bytes, $binaryFormat)
     {
         $file = new File(['maxSize' => $maxSize]);
 
-        $this->assertSame($bytes, $file->maxSize);
+        $this->assertEquals($bytes, $file->maxSize);
```
alexandre-daubois: This test fails on Appveyor when dealing with the "4G" and "4g" datasets. I suspect PHP to cast 4 billion to a string when running on a 32bits architecture

stof: Can we make it 3G instead, to fit on 32bits ?

alexandre-daubois: That wouldn't fit, but good idea! But I set it to 2G to fit signed integers.

Another solution would be to use `assertEquals` and `assertSame` depending on the architecture but I don't really like it

src/Symfony/Component/AssetMapper/Tests/CompiledAssetMapperConfigReaderTest.php
```diff
@@ -54,8 +55,9 @@ public function testLoadConfig()
     public function testSaveConfig()
     {
         $reader = new CompiledAssetMapperConfigReader($this->writableRoot);
-        $this->assertEquals($this->writableRoot.\DIRECTORY_SEPARATOR.'foo.json', realpath($reader->saveConfig('foo.json', ['foo' => 'bar'])));
-        $this->assertEquals(['foo' => 'bar'], json_decode(file_get_contents($this->writableRoot.'/foo.json'), true));
+
+        $this->assertSame(Path::normalize($this->writableRoot.\DIRECTORY_SEPARATOR.'foo.json'), Path::normalize(realpath($reader->saveConfig('foo.json', ['foo' => 'bar']))));
```
stof, line 59: Why normalizing paths here ? Comparing 2 real paths should not require any normalization.

alexandre-daubois: `CompiledAssetMapperConfigReader` does some internal operations on the directory string (in `saveConfig()`) which leads to a path with mixed `/` and `\`.

xabbuh: Is this still an issue after #52213?

src/Symfony/Component/Validator/Tests/Constraints/FileTest.php
```diff
@@ -101,8 +101,8 @@ public static function provideValidSizes()
             ['3M', 3000000, false],
             ['1gi', 1073741824, true],
             ['1GI', 1073741824, true],
-            ['4g', 4000000000, false],
-            ['4G', 4000000000, false],
+            ['2g', 2000000000, false],
+            ['2G', 2000000000, false],
```
xabbuh, line 105: Should we make this change on 5.4 to avoid conflicts when merging up?

alexandre-daubois: I agree, I'm opening the PR 👍

src/Symfony/Component/AssetMapper/Tests/CompiledAssetMapperConfigReaderTest.php
stof: But the assertion is not on the return value of `saveConfig`. It is on the result of `realpath` for that return value.

alexandre-daubois: It doesn't @xabbuh, I had a look at an outdated Appveyor log. Closing this PR, thanks for the review!

alexandre-daubois:
Closing, see https://github.com/symfony/symfony/pull/52258#discussion_r1369975865

PR #52259, opened 2023-10-24
[Validator] Fix `FileTest` to fit in 32bits architecture

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Following https://github.com/symfony/symfony/pull/52258#discussion_r1369819785

### DIFF
--- a/src/Symfony/Component/Validator/Tests/Constraints/FileTest.php
+++ b/src/Symfony/Component/Validator/Tests/Constraints/FileTest.php
@@ -31,6 +31,38 @@ class FileTest extends TestCase
         $this->assertTrue($file->__isset('maxSize'));
     }
 
+    /**
+     * @dataProvider provideValidSizesForIntegerOverflow
+     */
+    public function testMaxSizeWithIntegerOverflow($maxSize, $bytes, $binaryFormat)
+    {
+        if (4 !== \PHP_INT_SIZE) {
+            $this->markTestSkipped('This test can only run on 32-bits systems.');
+        }
+
+        $file = new File(['maxSize' => $maxSize]);
+
+        $this->assertSame($bytes, $file->maxSize);
+        $this->assertSame($binaryFormat, $file->binaryFormat);
+        $this->assertTrue($file->__isset('maxSize'));
+    }
+
+    /**
+     * @dataProvider provideValidSizesFor64bitsIntegers
+     */
+    public function testMaxSizeWith64bitsIntegers($maxSize, $bytes, $binaryFormat)
+    {
+        if (8 !== \PHP_INT_SIZE) {
+            $this->markTestSkipped('This test can only run on 64-bits systems.');
+        }
+
+        $file = new File(['maxSize' => $maxSize]);
+
+        $this->assertSame($bytes, $file->maxSize);
+        $this->assertSame($binaryFormat, $file->binaryFormat);
+        $this->assertTrue($file->__isset('maxSize'));
+    }
+
     public function testMagicIsset()
     {
         $file = new File(['maxSize' => 1]);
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
But we do want to test the behavior on x86 when the limit is above PHP_INT_MAX. That's removing it, isn't it?

stof:
@nicolas-grekas We would need a separate test for that, which would test that behavior. But the behavior on 32bits cannot be that we return an integer bigger than PHP_INT_MAX

nicolas-grekas:
Let's add this test case here then?

alexandre-daubois:
Good idea, I added tests for the 32 bits architecture

src/Symfony/Component/Validator/Tests/Constraints/FileTest.php
```diff
@@ -31,6 +31,22 @@ public function testMaxSize($maxSize, $bytes, $binaryFormat)
         $this->assertTrue($file->__isset('maxSize'));
     }
 
+    /**
+     * @dataProvider provideValidSizesForIntegerOverflow
+     */
+    public function testMaxSizeWithIntegerOverflow($maxSize, $bytes, $binaryFormat)
+    {
+        if (32 !== 8 * \PHP_INT_SIZE) {
```
xabbuh:
```suggestion
        if (4 !== \PHP_INT_SIZE) {
```

(we do this in other places too)

src/Symfony/Component/Validator/Tests/Constraints/FileTest.php
```diff
@@ -52,6 +68,22 @@ public function testMaxSizeCanBeSetAfterInitialization($maxSize, $bytes, $binary
         $this->assertSame($binaryFormat, $file->binaryFormat);
     }
 
+    /**
+     * @dataProvider provideValidSizesForIntegerOverflow
+     */
+    public function testMaxSizeCanBeSetAfterInitializationWithIntegerOverflow($maxSize, $bytes, $binaryFormat)
+    {
+        if (32 !== 8 * \PHP_INT_SIZE) {
```
xabbuh:
```suggestion
        if (4 !== \PHP_INT_SIZE) {
```

src/Symfony/Component/Validator/Tests/Constraints/FileTest.php
```diff
@@ -101,11 +133,17 @@ public static function provideValidSizes()
...
-            ['4G', 4000000000, false],
+            ['2g', 2000000000, false],
+            ['2G', 2000000000, false],
         ];
     }
 
+    public static function provideValidSizesForIntegerOverflow(): \Generator
+    {
+        yield ['4g', '4000000000', false];
+        yield ['4G', '4000000000', false];
```
xabbuh, line 176: I think we should also have a test for 64 bit systems exclusively that ensure the value is an integer.

xabbuh (changes requested):

src/Symfony/Component/Validator/Tests/Constraints/FileTest.php
xabbuh: Or maybe add a flag to the existing data provider which indicates if the provided fixture exceeds `PHP_INT_MAX` and cast to string on 32-bit systems?

xabbuh: see #52263 for what I have in mind

alexandre-daubois: I think your way is cleaner and better!

alexandre-daubois:
Closing in favor of https://github.com/symfony/symfony/pull/52263

PR #52265, opened 2023-10-24
[HttpClient] Configure `MockClient` if mock_response_factory has been set on a scoped client

### LINKED ISSUE #52257: [FrameworkBundle] [HttpClient] Make responsefactory configurable per client
### Description

This issue refs #51913. I would like to take a stab at making it possible to define the responsefactory on a per service level in the http_client.yaml file:

```config/packages/test/http_client.yaml:
framework:
  http_client:
    scoped_clients:
      apiServiceClient:
        mock_response_factory: "acme_mock_factory"

```

Two questions: 
1. I think I need some pointers to which files I should edit to parse the config and setup a mockhttpclient for the service.
2. Is the configuration ok? I.e. mock_response_factory ?

### Example

_No response_

tarjei:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #52257 
| License       | MIT

This makes it possible to use different mock response factories on different services and also to easily test services that uses http client endpoints.

It also generates an extra test service http_client.mock_client.$name that can be used to configure the MockHttpClient instance if you want to inject a different responsefactory.

Example usage:
```yaml
#config/packages/test/http_client.yaml
framework:
  http_client:
    scoped_clients:
      apiServiceClient:
        base_uri: ...
        mock_response_factory: "acme_mock_factory"
     otherClient:
        base_uri: ...
        mock_response_factory: true
     notMockedClient:
        base_uri: ...

```

The first client will use the acme_mock_factory for mock_factory while `otherClient` will be a MockHttpClient but without a set responsefactory (it can be configured in the test).


tarjei:
@nicolas-grekas  I'll fix the last tests if you like the direction of the PR :)

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2543,18 +2543,30 @@ private function registerHttpClientConfiguration(array $config, ContainerBuilder
...
+
+            $httpClientTransport = new Reference('http_client.transport');
+
+            if ($responseFactoryId = $scopeConfig['mock_response_factory'] ?? null) {
+
+
+                $container->register('http_client.mock_client.' . $name, MockHttpClient::class)
+                    ->setDecoratedService($httpClientTransport, null, -10)  // lower priority than TraceableHttpClient (5)
+                    ->setArguments(
+                        $responseFactoryId === true ? [] : [new Reference($responseFactoryId)]);
```
nicolas-grekas: this looks good to me but I don't get the `true` case, what is it for? we don't have it for the general mock factory

tarjei: It is if I want to have a mock client here but I do not want to configure the responsefactory as a service but just do it in the test. I considered having a separate attribute for it, i.e. `mock: true` and then having the mock_response_factory be an optional attribute if needed. 

Maybe being able to just say "mock this" and then add responses in the client is the main use case - i.e. that mock should be a separate parameter to make it more clear.

nicolas-grekas: Got it. Let's do the same for the global factory?

tarjei: Good idea. So add a mock: bool to both the global factory and the specific factory?

tarjei: @nicolas-grekas : ping :) I just wanted to get a reaction to the last question before working on the PR. Also: Is it possible to get this into the 6.x series?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2468,18 +2468,27 @@ private function registerHttpClientConfiguration(array $config, ContainerBuilder
             $retryOptions = $scopeConfig['retry_failed'] ?? ['enabled' => false];
             unset($scopeConfig['retry_failed']);
 
+            $httpClientTransport = new Reference('http_client.transport');
+
+            if ($responseFactoryId = $scopeConfig['mock_response_factory'] ?? null) {
+                $container->register('http_client.mock_client.'.$name, MockHttpClient::class)
+                    ->setDecoratedService($httpClientTransport, null, -10)  // lower priority than TraceableHttpClient (5)
```
Neirda24: it is also decorated with priority `10` for the global `mock_response_factory`. Maybe worth having a lower priority here so it superseeds the global one ?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2468,18 +2468,27 @@ private function registerHttpClientConfiguration(array $config, ContainerBuilder
...
+            }
+
             if (null === $scope) {
                 $baseUri = $scopeConfig['base_uri'];
                 unset($scopeConfig['base_uri']);
 
                 $container->register($name, ScopingHttpClient::class)
                     ->setFactory([ScopingHttpClient::class, 'forBaseUri'])
-                    ->setArguments([new Reference('http_client.transport'), $baseUri, $scopeConfig])
+                    ->setArguments([$httpClientTransport, $baseUri, $scopeConfig])
```
Neirda24: not sure this change is needed. Sure you save a few objects and memory but only during build time and this is like 3 objects.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2521,7 +2530,7 @@ private function registerHttpClientConfiguration(array $config, ContainerBuilder
         if ($responseFactoryId = $config['mock_response_factory'] ?? null) {
             $container->register('http_client.mock_client', MockHttpClient::class)
                 ->setDecoratedService('http_client.transport', null, -10)  // lower priority than TraceableHttpClient (5)
-                ->setArguments([new Reference($responseFactoryId)]);
+                ->setArguments(true === $responseFactoryId ? [] : [new Reference($responseFactoryId)]);
```
Neirda24: What is the goal here ? Tell me if I'm wrong but it looks like you are trying to set it to `[]` only if at least one dedicated mock was set. But then what if you have a global one and one specific for only one of the clients ? 

Because that variable is set only during foreach.

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/php/http_client_scoped_mock_response_factory_service.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+$container->loadFromExtension('framework', [
```
Neirda24, line 3: missing several test case IMO. Like what if you have a global one in addition to both a non mocked and a mocked one ?

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/yml/http_client_mock_response_factory.yml
```diff
@@ -6,4 +6,4 @@ framework:
         log: true
     http_client:
         default_options: ~
-        mock_response_factory: my_response_factory
+        mock_response_factory: true
```
Neirda24: I don't understand properly the `true`. Do you have a pending documentation PR that could help me understand ?

Neirda24 (commented):
Not core team so those remarks are my own. Only here to help ;)

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
tarjei: "Because that variable " <-- do you mean $responseFactoryId which is set in the if ?

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/yml/http_client_mock_response_factory.yml
tarjei: From a comment above:


It is if I want to have a mock client here but I do not want to configure the responsefactory as a service but just do it in the test. I considered having a separate attribute for it, i.e. mock: true and then having the mock_response_factory be an optional attribute if needed.

Maybe being able to just say "mock this" and then add responses in the client is the main use case - i.e. that mock should be a separate parameter to make it more clear.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
Neirda24: yup It is created and overidden in foreach at each iteration. So the only value it could have would be `the last scoped client says its true`

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/yml/http_client_mock_response_factory.yml
Neirda24: Yes I think a dedicated parameter is better because I do not find intuitive to set this to true to delegate at runtime during test to be able to provide the factory in question. Maybe some other approach would be a better fit ? Don't have an idea at the moment though.

tarjei: @Neirda24 maybe the problem is the parameter name in general?

If we renamed the parameter name to `mock_responses` then it could work the same way going forward. The problem then is backward compatibility which is harder to solve.

IMHO The right way would be be to change the parameter name, BUT due to backwards compability (as I understand Symfony's policy) this would require some extra gluecode to keep both options. One of the reasons I'm inclined to keep this current PRs solution is that it requires less code and tests than two options. 

I'm more than willing to do a documentation PR for it though :)

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
tarjei: I'll look into this :)

tarjei:
Ok, i've now updated this PR to separate out mock_client from mock_response_factory and also add tests and handling of overriding the options in scoped clients. 

Example config:

```
    http_client:
        default_options: ~
        mock_client: true
        scoped_clients:
            notMocked:
                base_uri : https://symfony.com
                mock_client : false
            mocked:
                base_uri: https://symfony.com
            mocked_with_factory:
                base_uri: https://symfony.com
                mock_response_factory: 'my_response_factory'

```

Note: Today mock_response_factory is directly under the http_client key. Maybe it both should be moved under the `default_options`  key to correspond better with the other options of the http client?

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/php/http_client_scoped_mock_response_factory_service.php
tarjei: Check now :)

tarjei:
@Neirda24 do the changes satisfy your concerns? :)

nicolas-grekas (approved):
I rebased on 8.1, lots of things changed meanwhile so the implementation is quite different, but that's details.
I removed `mock_client` and added back `true` as a special value for `mock_response_factory`, but now with a note about its meaning.

nicolas-grekas:
Thank you @tarjei.

PR #52272, opened 2023-10-24
[VarDump] Fix order of dumped properties - parent goes first

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? |
| Issues        |
| License       | MIT

### DIFF
--- a/src/Symfony/Component/VarDumper/Tests/Caster/CasterTest.php
+++ b/src/Symfony/Component/VarDumper/Tests/Caster/CasterTest.php
@@ -185,4 +185,29 @@ EOTXT
             }
         });
     }
+
+    public function testClassHierarchy()
+    {
+        $this->assertDumpMatchesFormat(<<<'DUMP'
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/VarDumper/Tests/Caster/CasterTest.php
```diff
@@ -185,4 +185,31 @@ public function __debugInfo(): array
             }
         });
     }
+
+    public function testClassHierarchy()
+    {
+        $c = eval(<<<'PHP'
```
stof: I don't think we need `eval` for that test. Create actual fixtures instead.

lyrixx: Fixed.

nicolas-grekas:
Thank you @lyrixx.

PR #52287, opened 2023-10-25
[FrameworkBundle] Fix deprecation layer for "enable_annotations" in validation and serializer configuration

lyrixx:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        |
| License       | MIT

## With Symfony 6.4 (without this patch) and the following configuration

```yaml
framework:
    serializer:
        enable_annotations: true
```

```
>…gregoire/dev/labs/symfony/symfony-6.4(main *) bin/console 

In BaseNode.php line 435:
                                                                                                                                                                                            
  Invalid configuration for path "framework.serializer": The "enable_annotations" and "enable_attributes" options at path "framework.serializer" must not be both set. Only the "enable_at  
  tributes" option must be used.                                                                                                                                                            
                                                                                                                                                                                            

In Configuration.php line 1163:
                                                                                                                                                                   
  The "enable_annotations" and "enable_attributes" options at path "framework.serializer" must not be both set. Only the "enable_attributes" option must be used.                                                                                                                                                                     
```

## With this PR

tests:

```yaml
framework:
    serializer:
        enable_annotations: true
```

Compilation is OK
Note : this is the previous configuration for some app

---

```yaml
framework:
    serializer:
        enable_annotations: true
        enable_attributes: true # same results with false, or null
```

Compilation is KO
Error :

> Invalid configuration for path "framework.serializer": The "enable_annotations" and "enable_attributes" options at path "framework.serializer" must not be both set. Only the "enable_attributes" option must be used.

---

```yaml
framework:
    serializer:
        enable_attributes: true # same results with false, or null
```

Compilation is OK

---

```yaml
framework:
    serializer: ~
```

Compilation is OK

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
@@ -1026,27 +1026,30 @@ class Configuration implements ConfigurationInterface
                         trigger_deprecation('symfony/framework-bundle', '6.4', 'Not setting the "framework.validation.email_validation_mode" config option is deprecated. It will default to "html5" in 7.0.');
                     }
 
-                    if (isset($v['enable_annotations'])) {
-                        trigger_deprecation('symfony/framework-bundle', '6.4', 'Option "enable_annotations" at "framework.validation" is deprecated. Use the "enable_attributes" option instead.');
-
-                        if (!isset($v['enable_attributes'])) {
-                            $v['enable_attributes'] = $v['enable_annotations'];
-                        } else {
-                            throw new LogicException('The "enable_annotations" and "enable_attributes" options at path "framework.validation" must not be both set. Only the "enable_attributes" option must be used.');
-                        }
-                    }
-
                     return $v;
                 })
             ->end()
             ->children()
                 ->arrayNode('validation')
+                    ->beforeNormalization()
+                        ->ifTrue(fn ($v) => isset($v['enable_annotations']))
+                        ->then(function ($v) {
+                            trigger_deprecation('symfony/framework-bundle', '6.4', 'Option "enable_annotations" at "framework.validation" is deprecated. Use the "enable_attributes" option instead.');
+
+                            if (isset($v['enable_attributes'])) {
+                                throw new LogicException('The "enable_annotations" and "enable_attributes" options at path "framework.validation" must not be both set. Only the "enable_attributes" option must be used.');
+                            }
+                            $v['enable_attributes'] = $v['enable_annotations'];
+
+                            return $v;
+                        })
+                    ->end()
                     ->info('validation configuration')
                     ->{$enableIfStandalone('symfony/validator', Validation::class)}()
                     ->children()
                         ->scalarNode('cache')->end()
                         ->booleanNode('enable_annotations')->end()
-                        ->booleanNode('enable_attributes')->{!class_exists(FullStack::class) ? 'defaultTrue' : 'defaultFalse'}()->end()
+                        ->booleanNode('enable_attributes')->{class_exists(FullStack::class) ? 'defaultFalse' : 'defaultTrue'}()->end()
                         ->arrayNode('static_method')
                             ->defaultValue(['loadValidatorMetadata'])
                             ->prototype('scalar')->end()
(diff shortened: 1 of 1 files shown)


src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1029,11 +1029,10 @@ private function addValidationSection(ArrayNodeDefinition $rootNode, callable $e
                     if (isset($v['enable_annotations'])) {
                         trigger_deprecation('symfony/framework-bundle', '6.4', 'Option "enable_annotations" at "framework.validation" is deprecated. Use the "enable_attributes" option instead.');
 
-                        if (!isset($v['enable_attributes'])) {
-                            $v['enable_attributes'] = $v['enable_annotations'];
-                        } else {
+                        if (isset($v['enable_attributes'])) {
```
stof: even this could be a BC break if it is done in `->validate()` as this would break if one file uses the new setting and another uses the old one.
Rewriting from the old one to the new one should be done in `beforeNormalization`

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1170,7 +1168,7 @@ private function addSerializerSection(ArrayNodeDefinition $rootNode, callable $e
                     ->{$enableIfStandalone('symfony/serializer', Serializer::class)}()
                     ->children()
                         ->booleanNode('enable_annotations')->end()
-                        ->booleanNode('enable_attributes')->{!class_exists(FullStack::class) ? 'defaultTrue' : 'defaultFalse'}()->end()
+                        ->booleanNode('enable_attributes')->{class_exists(FullStack::class) ? 'defaultFalse' : 'defaultNull'}()->end()
```
stof: why defaulting to null ?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1029,11 +1029,10 @@ private function addValidationSection(ArrayNodeDefinition $rootNode, callable $e
                     if (isset($v['enable_annotations'])) {
                         trigger_deprecation('symfony/framework-bundle', '6.4', 'Option "enable_annotations" at "framework.validation" is deprecated. Use the "enable_attributes" option instead.');
 
-                        if (!isset($v['enable_attributes'])) {
-                            $v['enable_attributes'] = $v['enable_annotations'];
-                        } else {
+                        if (isset($v['enable_attributes'])) {
                             throw new LogicException('The "enable_annotations" and "enable_attributes" options at path "framework.validation" must not be both set. Only the "enable_attributes" option must be used.');
```
stof: should we even forbid having both ? Does `enable_attributes` also enable the support for annotations ?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
stof: Attributes should be enabled by default, and this is not the case anymore.

lyrixx: To handle the BC layer. If the default to true, then it'll always throw an error with the following config:
```
    serializer:
        enable_annotations: true
```

How should I handle the logic then?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
lyrixx: IMHO having both should be forbidden. `enable_attributes` is new, so if one use it, they must remove the value of `enable_annotations`

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
stof: See my other comment about the fact that the BC layer should be before normalization (which is also before applying the defaults)

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
stof: having both _in the same file_ might be forbidden. but here, you check that after the merge of files, not when normalizing each file.

lyrixx:
@stof I pushed a new version. I now use `beforeNormalization` and I was able to revert the change about `defaultNull()`

derrabus (approved):

nicolas-grekas:
Thank you @lyrixx.

PR #52290, opened 2023-10-25
[DebugBundle] ignore a not-existing virtual request stack

### LINKED ISSUE #52280: [FrameworkBundle][DebugBundle][Console] Non-existant service .virtual_request_stack

6.4.x-dev

### Description

On a freshly installed project, commands are not working when APP_DEBUG=false

> In CheckExceptionOnInvalidReferenceBehaviorPass.php line 88:
                                                                                                          
  The service "data_collector.dump" has a dependency on a non-existent service ".virtual_request_stack".  

### How to reproduce

```code
 symfony new --webapp --version=next  debug-console \
    && cd debug-console \
    && APP_DEBUG=false php bin/console about
```

### Details / explanation

In the DebugBundle / [services.php](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Bundle/DebugBundle/Resources/config/services.php)

```php
        ->set('data_collector.dump', DumpDataCollector::class)
            ->public()
            ->args([
                service('debug.stopwatch')->ignoreOnInvalid(),
                service('debug.file_link_formatter')->ignoreOnInvalid(),
                param('kernel.charset'),
                service('.virtual_request_stack'),
                null, // var_dumper.cli_dumper or var_dumper.server_connection when debug.dump_destination is set
            ])
            ->tag('data_collector', [
                'id' => 'dump',
                'template' => '@Debug/Profiler/dump.html.twig',
                'priority' => 240,
            ])
```

In the FrameworkBundle / debug.php

```php
->set('.virtual_request_stack', VirtualRequestStack::class)
    ->args([service('request_stack')])
    ->public()
```

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52280
| License       | MIT

### DIFF
--- a/src/Symfony/Bundle/DebugBundle/DependencyInjection/Compiler/DumpDataCollectorPass.php
+++ b/src/Symfony/Bundle/DebugBundle/DependencyInjection/Compiler/DumpDataCollectorPass.php
@@ -14,6 +14,7 @@ namespace Symfony\Bundle\DebugBundle\DependencyInjection\Compiler;
 use Symfony\Bundle\WebProfilerBundle\EventListener\WebDebugToolbarListener;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
 
 /**
  * Registers the file link format for the {@link \Symfony\Component\HttpKernel\DataCollector\DumpDataCollector}.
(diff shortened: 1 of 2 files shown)


src/Symfony/Bundle/DebugBundle/Resources/config/services.php
```diff
@@ -50,7 +50,7 @@
                 service('debug.stopwatch')->ignoreOnInvalid(),
                 service('debug.file_link_formatter')->ignoreOnInvalid(),
                 param('kernel.charset'),
-                service('.virtual_request_stack'),
+                service('.virtual_request_stack')->ignoreOnInvalid(),
```
nicolas-grekas: shouldn't we instead pass the request stack when the virtual one doesn't exist? that'd match the previous behavior

the solution might be to alias the virtual stack to the real one in FWB when not in debug, WDYT?

xabbuh: I did that in a compiler pass now. Adding an alias is probably not a solution as long as we do not want to list older versions of the FrameworkBundle as a conflict.

nicolas-grekas:
Thank you @xabbuh.

PR #52296, opened 2023-10-25
[HttpClient] Allow define connect timeout with Curl Client

ppelisset:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

The purpose of this patch is to allow the use of the `CURLOPT_CONNECTTIMEOUT` and `CURLOPT_CONNECTTIMEOUT_MS` constants in the extra parameters of HttpClient when the Curl HTTP client is used.

In the case of using HttpClient for long HTTP calls, where the response, including headers and the HTTP response code, is only returned at the end, it is important to have an earlier timeout for establishing the TCP connection with the target HTTP server (to avoid waiting for the entire total timeout in case of network issues, for example).

Unfortunately, the two options currently provided by HttpClient do not offer this possibility. `max_duration` is a timeout for the complete execution of the request, and `timeout` is a timeout for the absence of a response from the HTTP server (which is the case here with a long process where the HTTP status code indicates success or failure).

Initially, I wanted to propose adding a `connect_timeout` option within the HTTP client. However, after several attempts, it seems impossible with the NativeClient, as the `fopen` function only returns a stream when the headers are available and not when the TCP connection is established. I explored various approaches, including changing the stream context timeout during the connection notification (`STREAM_NOTIFY_CONNECT`), but it doesn't seem possible.

Therefore, I see no option other than making these two parameters accessible to enable this usage with the Curl client.


nicolas-grekas:
That's already possible, you should use the `$timeout` argument of the `stream()` method.

ppelisset:
Thank you for the help; I managed to achieve the expected behavior by combining the use of streams with CurlResponse::getInfo('connect_time_us') to obtain information about the TCP connection status during the stream timeout. This only works with the Curl client, but it suits my needs.

The code is not perfect, but since others before me were looking for a solution, I'd like to share mine here before closing the PR :
```
function checkResponseConnectTimeout(ResponseInterface $response, float $connectTimeout): void
{
    try {
        foreach ($this->httpClient->stream($response, $connectTimeout) as $chunk) {
            if ($chunk->isLast()) {
                return;
            }
        }
    } catch (TimeoutException) {
        if ($response->getInfo('connect_time_us') === 0) {
            $response->cancel();
            throw new \RuntimeException(sprintf("Timeout during connect to %s", $response->getInfo('url')));
        }
    }
}
```

PR #52307, opened 2023-10-26
[Scheduler] Save checkpoint in a finally block

### LINKED ISSUE #52288: [Scheduler] Stateful checkpoint is not saved if the worker has stopped

6.3 6.4

### Description

The message generator appears to store stateful data by loading and saving the checkpoint with the cache.

But if the worker is configured to stop after X message(s), **once the last message is handled, the (messenger) worker will break its loop and stop**. In this case, the execution stop and **the stateful data of the message  (time of last execution for example) will not be saved**. 

So when the consumer restarts it will re-handle the message because the last execution was not stored.
This bug appears to break the stateful behavior :thinking:   

`MessageGenerator.php`
```php
 public function getMessages(): \Generator
    {
        $checkpoint = $this->checkpoint();
        // ... 
        while (!$heap->isEmpty() && $heap->top()[0] <= $now) {
            // ...
            if ($yield) {
                $context = new MessageContext($this->name, $id, $trigger, $time, $nextTime);
                foreach ($recurringMessage->getMessages($context) as $message) {
                    yield $context => $message; // Message is handled here
                }

                $checkpoint->save($time, $index); // Checkpoint is saved here, /!\ not reached if worker has stopped ! 
            }
        }
        // ...
    }
```

### How to reproduce

Create a simple stateful schedule provider like below : 

```php

#[AsSchedule]
readonly class ScheduleProvider implements ScheduleProviderInterface
{
    public function __construct(
        private CacheInterface $cache,
        private LockFactory $lockFactory
    ) {
    }

    {
        $schedule = (new Schedule())
            ->stateful($this->cache)
            ->lock($this->lockFactory->createLock('surface_scheduler'));

        $schedule->add(RecurringMessage::every('10 seconds', new TestMessage(data: 'Cron', duration: 1)));

        return $schedule;
    }
}
```

Run the following command : 
```bash
console messenger:consume scheduler_default --limit 1
```

The worker will wait 10 seconds, handle the message and stop.

Just after that, rerun the command, and you will see that the message will immediately be re-executed. 

### Exemple
 ```bash
$ console messenger:consume scheduler_default --limit 1
                                                                                                           
 [OK] Consuming messages from transport "scheduler_default".                                                            
                                                                                                                        
 // The worker will automatically exit once it has processed 1 messages or received a stop signal via the               
 // messenger:stop-workers command.                                                                                                                    

Message received: Cron at 2023-10-25 11:06:00 // First execution -> OK


$ console messenger:consume scheduler_default --limit 1
                                                                                                    
 [OK] Consuming messages from transport "scheduler_default".                                                            
                                                                                                                        
 // The worker will automatically exit once it has processed 1 messages or received a stop signal via the               
 // messenger:stop-workers command.                                                                                     

Message received: Cron at 2023-10-25 11:06:03 // Second execution -> KO, did not wait 10 seconds since last time


$ console messenger:consume scheduler_default --limit 1
                                                                                                                        
 [OK] Consuming messages from transport "scheduler_default".                                                            
                                                                                                                        
 // The worker will automatically exit once it has processed 1 messages or received a stop signal via the               
 // messenger:stop-workers command.                                              

Message received: Cron at 2023-10-25 11:06:06 // Third execution -> KO, idem
```

### Possible Solution
The dumbest solution would be to move the `break` in the Messenger `Worker.php` like below.
But I'm not sure if it would have any side effects.

```diff
foreach ($envelopes as $envelope) {
+   if ($this->shouldStop) {
+     break 2;
+   }

    $envelopeHandled = true;

    $this->rateLimit($transportName);
    $this->handleMessage($envelope, $transportName);
    $this->eventDispatcher?->dispatch(new WorkerRunningEvent($this, false));

-   if ($this->shouldStop) {
-     break 2;
-   }
}
```

### Additional Context

_No response_

FrancoisPog:
| Q             | A
| ------------- | ---
| Branch?       | 6.3  
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Issues        | Fix #52288 
| License       | MIT

As described in the linked issue : 
> The message generator appears to store stateful data by loading and saving the checkpoint with the cache.
> 
> But if the worker is configured to stop after X message(s), once the last message is handled, the (messenger) worker will break its loop and stop. In this case, the execution stop and the stateful data of the message (time of last execution for example) will not be saved.
> 
> So when the consumer restarts it will re-handle the message because the last execution was not stored.
> 

So the solution here is to put the code that save the checkpoint in a `finally` block. 
Thanks [maelanleborgne](https://github.com/maelanleborgne).

### DIFF
--- a/src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
+++ b/src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
@@ -72,8 +72,11 @@ final class MessageGenerator implements MessageGeneratorInterface
             }
 
             if ($yield) {
-                yield (new MessageContext($this->name, $id, $trigger, $time, $nextTime)) => $message;
-                $checkpoint->save($time, $index);
+                try {
+                    yield (new MessageContext($this->name, $id, $trigger, $time, $nextTime)) => $message;
+                } finally {
+                    $checkpoint->save($time, $index);
+                }
             }
         }
 
--- a/src/Symfony/Component/Scheduler/Tests/Generator/MessageGeneratorTest.php
+++ b/src/Symfony/Component/Scheduler/Tests/Generator/MessageGeneratorTest.php
@@ -14,6 +14,8 @@ namespace Symfony\Component\Scheduler\Tests\Generator;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
 use Symfony\Component\Clock\ClockInterface;
+use Symfony\Component\Clock\MockClock;
+use Symfony\Component\Scheduler\Generator\Checkpoint;
 use Symfony\Component\Scheduler\Generator\MessageContext;
 use Symfony\Component\Scheduler\Generator\MessageGenerator;
 use Symfony\Component\Scheduler\RecurringMessage;
@@ -126,6 +128,32 @@ class MessageGeneratorTest extends TestCase
         $this->assertSame($message->getTrigger(), $context->trigger);
         $this->assertEquals(self::makeDateTime('22:14:00'), $context->triggeredAt);
         $this->assertEquals(self::makeDateTime('22:16:00'), $context->nextTriggerAt);
+    }
+
+    public function testCheckpointSavedInBrokenLoop()
+    {
+        $clock = new MockClock(self::makeDateTime('22:12:00'));
+
+        $message = $this->createMessage((object) ['id' => 'message'], '22:13:00', '22:14:00', '22:16:00');
+        $schedule = (new Schedule())->add($message);
+
+        $cache = new ArrayAdapter();
+        $schedule->stateful($cache);
+        $checkpoint = new Checkpoint('dummy', cache: $cache);
+
+        $scheduler = new MessageGenerator($schedule, 'dummy', clock: $clock, checkpoint: $checkpoint);
+
+        // Warmup. The first run is always returns nothing.
+        $this->assertSame([], iterator_to_array($scheduler->getMessages(), false));
+
+        $clock->sleep(60 + 10); // 22:13:10
+
+        foreach ($scheduler->getMessages() as $message) {
+            // Message is handled but loop is broken just after
+            break;
+        }
+
+        $this->assertEquals(self::makeDateTime('22:13:00'), $checkpoint->time());
     }
 
     public static function messagesProvider(): \Generator


src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
```diff
@@ -72,8 +72,11 @@ public function getMessages(): \Generator
             }
 
             if ($yield) {
-                yield (new MessageContext($this->name, $id, $trigger, $time, $nextTime)) => $message;
-                $checkpoint->save($time, $index);
+                try{
```
fabpot:
```suggestion
                try {
```

src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
```diff
@@ -72,8 +72,11 @@ public function getMessages(): \Generator
...
+                    yield (new MessageContext($this->name, $id, $trigger, $time, $nextTime)) => $message;
+                }finally {
```
fabpot:
```suggestion
                } finally {
```

fabpot (changes requested):

src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
```diff
@@ -72,8 +72,11 @@ public function getMessages(): \Generator
             }
 
             if ($yield) {
-                yield (new MessageContext($this->name, $id, $trigger, $time, $nextTime)) => $message;
-                $checkpoint->save($time, $index);
+                try {
+                    yield (new MessageContext($this->name, $id, $trigger, $time, $nextTime)) => $message;
+                } finally {
+                    $checkpoint->save($time, $index);
```
fabpot, line 78: Moving this line before the `yield` would do the same, right?

stof:
Please add a test preventing regressions.

src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
maelanleborgne: No, that causes the scheduler to create messages without interruption (tested)

FrancoisPog: > No, that causes the scheduler to create messages without interruption (tested)

@maelanleborgne 
I didn't catch what wouldn't work, I added a test and it makes no difference.
Feel free to refine it to highlight the difference.

src/Symfony/Component/Scheduler/Tests/Generator/MessageGeneratorTest.php
```diff
@@ -128,6 +129,35 @@ public function testYieldedContext()
         $this->assertEquals(self::makeDateTime('22:16:00'), $context->nextTriggerAt);
     }
 
+    public function testCheckpointSavedInBrokenLoop()
+    {
+        $now = self::makeDateTime('22:12:00');
+
+        $clock = $this->createMock(ClockInterface::class);
```
OskarStark: You can use ClockMock here

src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
maelanleborgne: > > No, that causes the scheduler to create messages without interruption (tested)
> 
> @maelanleborgne I didn't catch what wouldn't work, I added a test and it makes no difference. Feel free to refine it to highlight the difference.

Sorry I misread the comment. I tested on the [6.4 version of the component,](https://github.com/symfony/symfony/commit/06b16bd921cee96ff4c2738d647cf05d59aad0ad#diff-3431200d32b70ea151463d9c189f082bca9767fa7fbd128ed6d9646f05506df4R83) which uses à `foreach ($recurringMessage...)`, and there we need to wrap the `foreach` in a `try` .
That will be good to know when merging up 6.3 into 6.4

src/Symfony/Component/Scheduler/Tests/Generator/MessageGeneratorTest.php
FrancoisPog: I did that to keep consistency with other tests, but yeah it's would be better

src/Symfony/Component/Scheduler/Generator/MessageGenerator.php
FrancoisPog: yes, I thought about it, because the related issue was on 6.4. 
So for 6.3, shouldn't we keep the `try/finally` ?

maelanleborgne: Not sure what's the recommended way of doing it.
As suggested by @fabpot, the try/finally isn't necessary in 6.3, so we might as well no overcomplicate the code and just move the `$chekpoint->save()` above the `yield`.
Then you could just keep an eye on the 6.4 and 7 branches to see how it reflects once merged up. And open a new PR if needed :smiley:

FrancoisPog: Yes, that makes sense.
I couldn't make a choice on my own, both solutions seem equivalent, so maybe the core team could help us ? 😃

fabpot: Let's keep the `try/finally` block as it will be more consistent with the code in 6.4/7.0.

fabpot (approved):

fabpot:
Thank you @FrancoisPog.

Jeroeny:
You might be interested to know that a checkpoint resets anyway after another process takes over: https://github.com/symfony/symfony/pull/52039

PR #52313, opened 2023-10-26
[FrameworkBundle][Scheduler] Add test for autoconfigured schedule

valtzu:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| License       | MIT

Add functional tests for autoconfigured schedules / tasks, as discussed in https://github.com/symfony/symfony/pull/51525#discussion_r1342467103.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/composer.json
+++ b/src/Symfony/Bundle/FrameworkBundle/composer.json
@@ -35,6 +35,7 @@
     "require-dev": {
         "doctrine/annotations": "^1.13.1|^2",
         "doctrine/persistence": "^1.3|^2|^3",
+        "dragonmantank/cron-expression": "^3.1",
         "seld/jsonlint": "^1.10",
         "symfony/asset": "^5.4|^6.0|^7.0",
         "symfony/asset-mapper": "^6.4|^7.0",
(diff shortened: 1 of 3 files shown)


valtzu:
Added autoconfigured scheduler tests as discussed in https://github.com/symfony/symfony/pull/51525#discussion_r1342467103, but those are failing in the pipeline due to missing cron dev-dependency. Should I just remove cron attributes from DummyTask or how to proceed here @fabpot ?

Locally with `dragonmantank/cron-expression` installed the tests pass just fine.

fabpot:
> Added autoconfigured scheduler tests as discussed in [#51525 (comment)](https://github.com/symfony/symfony/pull/51525#discussion_r1342467103), but those are failing in the pipeline due to missing cron dev-dependency. Should I just remove cron attributes from DummyTask or how to proceed here @fabpot ?
> 
> Locally with `dragonmantank/cron-expression` installed the tests pass just fine.

You can add `dragonmantank/cron-expression` as a `dev` dependency for the bundle and in the main `composer.json` file.

src/Symfony/Bundle/FrameworkBundle/composer.json
```diff
@@ -71,6 +71,7 @@
         "symfony/uid": "^5.4|^6.0|^7.0",
         "symfony/web-link": "^5.4|^6.0|^7.0",
         "phpdocumentor/reflection-docblock": "^3.0|^4.0|^5.0",
+        "dragonmantank/cron-expression": "^3.1",
```
fabpot: You need to move it to keep the list ordered.
Also, you must add it to the main `composer.json` file as well.

fabpot (changes requested):

src/Symfony/Bundle/FrameworkBundle/composer.json
valtzu: I suspected that but then saw `phpdocumentor` after `symfony` :thinking: 

Fixed the order, and it is already added to main composer.json previously

OskarStark (approved):
Tests are 💚 so I think this can be merged @nicolas-grekas @fabpot

fabpot:
Thank you @valtzu.

PR #52323, opened 2023-10-27
[AssetMapper] Allowing circular references in JavaScriptImportPathCompiler

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | None
| License       | MIT

Hi!

In AssetMapper, we throw a `CircularAssetsException` if you ask for the `app.js` asset, but it's still being created. This circular situation only happens via asset compilers: if, while compiling `app.js`, we ask for `other.js`... and then while compiling that, we ask for `app.js`.

However, JS modules *do* allow for circular references - e.g. `app.js` imports `other.js` which imports `app.js`. Before this PR, this situation would trigger the `CircularAssetsException`. After, it is handled correctly. `JavaScriptImportPathCompiler` calls `$assetMapper->getAsset()` only to populate the `MappedAsset.javaScriptImports` property. So, allowing for the half-created `MappedAsset` is safe here: we do not use any of the not-yet-populated properties.

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
+++ b/src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
@@ -148,11 +148,15 @@ final class JavaScriptImportPathCompiler implements AssetCompilerInterface
             return null;
         }
 
-        if ($asset = $assetMapper->getAsset($importMapEntry->path)) {
-            return $asset;
-        }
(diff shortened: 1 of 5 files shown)


fabpot (approved):

fabpot:
Thank you @weaverryan.

PR #52329, opened 2023-10-27
[HttpClient] Psr18Client: parse HTTP Reason Phrase for Response

### LINKED ISSUE #51527: [HttpClient] `Psr18Client` can't get custom HTTP status messages

5,4, 6.4

### Description

for example if some API i want to connect uses HTTP Status Codes like this:

```
HTTP/1.1 404 countryCodeNotFound
```

then when using Symfony HttpClient as its own Client, then the Exception sets the Message correct

but the Psr18Client doesn't parse the `response_headers` like the Exceptions do
https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/HttpClient/Exception/HttpExceptionTrait.php#L34

the Psr18Client sets the status code without trying to set the message too:
https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/HttpClient/Psr18Client.php#L108

### How to reproduce

have an API with custom status messages:

```
HTTP/1.1 404 countryCodeNotFound
```

### Possible Solution

Similar code that is done in the Exception Trait should also be done in the Psr18Client

### Additional Context

_No response_

Hanmac:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no 
| Issues        | Fix #51527
| License       | MIT

### DIFF
--- a/src/Symfony/Component/HttpClient/Internal/HttplugWaitLoop.php
+++ b/src/Symfony/Component/HttpClient/Internal/HttplugWaitLoop.php
@@ -116,7 +116,12 @@ final class HttplugWaitLoop
 
     public function createPsr7Response(ResponseInterface $response, bool $buffer = false): Psr7ResponseInterface
     {
-        $psrResponse = $this->responseFactory->createResponse($response->getStatusCode());
+        $responseParameters = [$response->getStatusCode()];
+        if ($phrase = $this->getReasonPhrase($response)) {
+            $responseParameters[] = $phrase;
+        }
+
+        $psrResponse = $this->responseFactory->createResponse(...$responseParameters);
 
         foreach ($response->getHeaders(false) as $name => $values) {
             foreach ($values as $value) {
@@ -141,5 +146,18 @@ final class HttplugWaitLoop
         }
 
         return $psrResponse->withBody($body);
+    }
+
+    private function getReasonPhrase(ResponseInterface $response): ?string
+    {
+        $reason = null;
+
+        foreach ($response->getInfo('response_headers') as $h) {
+            if (11 <= \strlen($h) && '/' === $h[4] && preg_match('#^HTTP/\d+(?:\.\d+)? (?:\d\d\d) (.+)#', $h, $m)) {
+                $reason = $m[1];
+            }
+        }
+
+        return $reason;
     }
 }
--- a/src/Symfony/Component/HttpClient/Psr18Client.php
+++ b/src/Symfony/Component/HttpClient/Psr18Client.php
@@ -31,6 +31,7 @@ use Symfony\Component\HttpClient\Response\StreamableInterface;
 use Symfony\Component\HttpClient\Response\StreamWrapper;
 use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
+use Symfony\Contracts\HttpClient\ResponseInterface as HttpClientResponseInterface;
 use Symfony\Contracts\Service\ResetInterface;
 
 if (!interface_exists(ClientInterface::class)) {
@@ -105,7 +106,9 @@ final class Psr18Client implements ClientInterface, RequestFactoryInterface, Str
 
             $response = $this->client->request($request->getMethod(), (string) $request->getUri(), $options);
 
-            $psrResponse = $this->responseFactory->createResponse($response->getStatusCode());
+            $psrResponse = $this->responseFactory->createResponse(
+                $response->getStatusCode(), $this->getReasonPhrase($response)
+            );
 
             foreach ($response->getHeaders(false) as $name => $values) {
                 foreach ($values as $value) {
@@ -194,6 +197,19 @@ final class Psr18Client implements ClientInterface, RequestFactoryInterface, Str
         if ($this->client instanceof ResetInterface) {
             $this->client->reset();
         }
+    }
+
+    protected function getReasonPhrase(HttpClientResponseInterface $response): ?string
+    {
+        foreach (array_reverse($response->getInfo('response_headers')) as $h) {
+            if (str_starts_with($h, 'HTTP/')) {
+
+                if (preg_match('#^HTTP/\d+(?:\.\d+)? \d\d\d (.+)$#', $h, $m)) {
+                    return $m[1];
+                }
+            }
+        }
+        return null;
     }
 }
 
--- a/src/Symfony/Component/HttpClient/Psr18Client.php
+++ b/src/Symfony/Component/HttpClient/Psr18Client.php
@@ -201,15 +201,15 @@ final class Psr18Client implements ClientInterface, RequestFactoryInterface, Str
 
     protected function getReasonPhrase(HttpClientResponseInterface $response): ?string
     {
-        foreach (array_reverse($response->getInfo('response_headers')) as $h) {
-            if (str_starts_with($h, 'HTTP/')) {
+        $reason = null;
 
-                if (preg_match('#^HTTP/\d+(?:\.\d+)? \d\d\d (.+)$#', $h, $m)) {
-                    return $m[1];
-                }
+        foreach ($response->getInfo('response_headers') as $h) {
+            if (11 <= \strlen($h) && '/' === $h[4] && preg_match('#^HTTP/\d+(?:\.\d+)? (?:\d\d\d) (.+)#', $h, $m)) {
+                $reason = $m[1];
             }
         }
-        return null;
+
+        return $reason;
     }
 }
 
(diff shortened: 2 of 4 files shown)


stof:
How does this behave when the actual request is done with HTTP/2, which does not transmit a reason phrase ?

Hanmac:
> How does this behave when the actual request is done with HTTP/2, which does not transmit a reason phrase ?

the Regex shouldn't find any, and then return nothing, and then the factory should fallback to the old one

src/Symfony/Component/HttpClient/Psr18Client.php
```diff
@@ -195,6 +198,19 @@ public function reset(): void
...
+    {
+        foreach (array_reverse($response->getInfo('response_headers')) as $h) {
+            if (str_starts_with($h, 'HTTP/')) {
+
+                if (preg_match('#^HTTP/\d+(?:\.\d+)? \d\d\d (.+)$#', $h, $m)) {
+                    return $m[1];
+                }
+            }
+        }
+        return null;
```
nicolas-grekas: Let's borrow from `TransportResponseTrait`:

```suggestion
        $reason = null;

        foreach ($response->getInfo('response_headers') as $h) {
            if (11 <= \strlen($h) && '/' === $h[4] && preg_match('#^HTTP/\d+(?:\.\d+)? (?:\d\d\d) (.+)#', $h, $m)) {
                $reason = $m[1];
            }
        }

        return $reason;
```

src/Symfony/Component/HttpClient/Tests/Psr18ClientTest.php
```diff
@@ -101,4 +101,20 @@ public function testInvalidHeaderResponse()
         $resultResponse = $client->sendRequest($request);
         $this->assertCount(1, $resultResponse->getHeaders());
     }
+
+    public function testResponseReasonPhrase()
+    {
+        #'HTTP/1.1 103 Early Hints'
```
nicolas-grekas: can be removed

nicolas-grekas (commented):
Thanks for the PR, here are some comments.

I think `HttplugClient` might need the same, can you have a look?

Hanmac:
> It looks like you unchecked the "Allow edits from maintainer" box. That is fine, but please note that if you have multiple commits, you'll need to squash your commits into one before this can be merged. Or, you can check the "Allow edits from maintainers" box and the maintainer can squash for you.
> 
> Cheers!
> 
> Carsonbot

Where can I check this right now?

I didn't saw it when making this MR

stof:
I'm not sure this checkbox exists for PRs sent from a fork owned by an organization instead of a user.

Hanmac:
i probably need to do some trickery with `responseFactory->createResponse` and the amount of parameters.

while the Response class allows the Reason Phase to be null (and falls back to the hard codes ones, MAYBE)
the factory doesn't allow null as parameter, and passing empty string is not the same as passing 1 parameter.

so i probably need to do something like:
```php

$responseParameters = [$response->getStatusCode()];
if ($phrase = $this->getReasonPhrase($response)) {
  $responseParameters []= $phrase;
}

$psrResponse = $this->responseFactory->createResponse(...$responseParameters);

```

i need to try this after weekend

src/Symfony/Component/HttpClient/Psr18Client.php
```diff
@@ -195,6 +198,19 @@ public function reset(): void
             $this->client->reset();
         }
     }
+
+    protected function getReasonPhrase(HttpClientResponseInterface $response): ?string
```
derrabus:
```suggestion
    private function getReasonPhrase(HttpClientResponseInterface $response): ?string
```

This is not an extension point.

derrabus (approved):

derrabus (changes requested):
This test failure appears to be related.

```php
1) Symfony\Component\HttpClient\Tests\Psr18ClientTest::testInvalidHeaderResponse
TypeError: Http\Discovery\Psr17Factory::createResponse(): Argument #2 ($reasonPhrase) must be of type string, null given, called in /home/runner/work/symfony/symfony/src/Symfony/Component/HttpClient/Psr18Client.php on line 109

/home/runner/work/symfony/symfony/vendor/php-http/discovery/src/Psr17Factory.php:84
/home/runner/work/symfony/symfony/src/Symfony/Component/HttpClient/Psr18Client.php:109
/home/runner/work/symfony/symfony/src/Symfony/Component/HttpClient/Tests/Psr18ClientTest.php:101
```

Hanmac:
> This test failure appears to be related.
> 
> ```
> 1) Symfony\Component\HttpClient\Tests\Psr18ClientTest::testInvalidHeaderResponse
> TypeError: Http\Discovery\Psr17Factory::createResponse(): Argument #2 ($reasonPhrase) must be of type string, null given, called in /home/runner/work/symfony/symfony/src/Symfony/Component/HttpClient/Psr18Client.php on line 109
> 
> /home/runner/work/symfony/symfony/vendor/php-http/discovery/src/Psr17Factory.php:84
> /home/runner/work/symfony/symfony/src/Symfony/Component/HttpClient/Psr18Client.php:109
> /home/runner/work/symfony/symfony/src/Symfony/Component/HttpClient/Tests/Psr18ClientTest.php:101
> ```

That's why I said I need to at this maybe after weekend

Hanmac:
> This test failure appears to be related.
> 
> ```
> 1) Symfony\Component\HttpClient\Tests\Psr18ClientTest::testInvalidHeaderResponse
> TypeError: Http\Discovery\Psr17Factory::createResponse(): Argument #2 ($reasonPhrase) must be of type string, null given, called in /home/runner/work/symfony/symfony/src/Symfony/Component/HttpClient/Psr18Client.php on line 109
> 
> /home/runner/work/symfony/symfony/vendor/php-http/discovery/src/Psr17Factory.php:84
> /home/runner/work/symfony/symfony/src/Symfony/Component/HttpClient/Psr18Client.php:109
> /home/runner/work/symfony/symfony/src/Symfony/Component/HttpClient/Tests/Psr18ClientTest.php:101
> ```

took my laptop to work on this thing ... see my idea above with the variable parameters

Hanmac:
> This test failure appears to be related.
> 
> ```
> 1) Symfony\Component\HttpClient\Tests\Psr18ClientTest::testInvalidHeaderResponse
> TypeError: Http\Discovery\Psr17Factory::createResponse(): Argument #2 ($reasonPhrase) must be of type string, null given, called in /home/runner/work/symfony/symfony/src/Symfony/Component/HttpClient/Psr18Client.php on line 109
> 
> /home/runner/work/symfony/symfony/vendor/php-http/discovery/src/Psr17Factory.php:84
> /home/runner/work/symfony/symfony/src/Symfony/Component/HttpClient/Psr18Client.php:109
> /home/runner/work/symfony/symfony/src/Symfony/Component/HttpClient/Tests/Psr18ClientTest.php:101
> ```

took my laptop to work on this thing ... see my idea above with the variable parameters

src/Symfony/Component/HttpClient/Internal/HttplugWaitLoop.php
```diff
@@ -142,4 +147,17 @@ public function createPsr7Response(ResponseInterface $response, bool $buffer = f
 
         return $psrResponse->withBody($body);
     }
+
+    private function getReasonPhrase(ResponseInterface $response): ?string
```
nicolas-grekas, line 151: because the class is internal, I propose to make this method public static and reuse call it from Psr18Client

I also propose to return the empty string when no reason phrase is found, and simplify the logic on the caller sides: no need for the complexity about arguments anymore

Hanmac: The problem is that the psr17factory reacts different depending on the number of parameters

Hanmac:
The last failures seems to be unrelated?

src/Symfony/Component/HttpClient/Internal/HttplugWaitLoop.php
Hanmac: There it checks for the amount of parameters:

https://github.com/Nyholm/psr7/blob/3cb4d163b58589e47b35103e8e5e6a6a475b47be/src/Factory/Psr17Factory.php#L25-L28

nicolas-grekas: oops, thanks for the link

nicolas-grekas:
Thank you @Hanmac.

nicolas-grekas:
Merged as 115a5a1a097a1355875e7a2f8e175e257fc6ad89 in 5.4
For your next PR, please allow edits from maintainers.

GrahamCampbell:
> For your next PR, please allow edits from maintainers.

Edits were not possible because it was forked to an organization account.

PR #52330, opened 2023-10-27
[AssetMapper] Fixing memory bug where we stored way more file content than needed

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes (on Symfonycasts, without this, `asset-map:compile` takes >500mb of memory)
| New feature?  | no
| Deprecations? | none
| Issues        | None
| License       | MIT

Hi!

This drastically improves the memory footprint when ALL assets are built, which happens in `asset-map:compile`, `debug:asset-map`, etc. Blackfire is a huge fan! https://blackfire.io/profiles/compare/4eb36732-8805-4c49-b636-d4bf8f9e2b27/graph

The MAIN optimization by far is to set `MappedAsset.content` to null if the final, compiled content matches the source file's content. This is the case for the vast-majority of mapped assets, and so the result is MUCH less content stored in memory for no reason.

Thanks to @smnandre and his earlier PR, which allows for the `null` content on mapped assets.

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/Command/AssetMapperCompileCommand.php
+++ b/src/Symfony/Component/AssetMapper/Command/AssetMapperCompileCommand.php
@@ -104,11 +104,9 @@ EOT
 
     private function createManifestAndWriteFiles(SymfonyStyle $io): array
     {
-        $allAssets = $this->assetMapper->allAssets();
-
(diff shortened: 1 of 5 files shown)


stof (approved):

nicolas-grekas:
Thank you @weaverryan.

PR #52331, opened 2023-10-27
[AssetMapper] Fix file deleting errors & remove nullable MappedAsset on JS import

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | None
| License       | MIT

Hi!

This PR accomplishes 2 things:

A) If `app.js` imported `foo.js`, and you deleted `foo.js`, you got a big, unobvious error from Symfony. This was because the `MappedAsset` behind `app.js` remained cached, including the `JavaScriptImport` for `foo.js`. So then, the system was surprised when `app.js` had this `JavaScriptImport`... but the underlying file didn't exist. The fix was to add a `FileExistenceResource` for `foo.js`. We don't need to update the `app.js` cache if `foo.js` changes, but we DO need to update it if `foo.js` no longer exists (so that we can create a new `MappedAsset` without the import).

B) Upon looking at this, previously, `JavaScriptImport.asset` was nullable. That doesn't make sense, now that all vendor files are downloaded locally and exist in the AssetMapper. The `MappedAsset.javascriptImports` are used to implicitly add importmap entries for relative assets (which already required a `MappedAsset`) and to preload other dependencies. Previously, we added a `JavaScriptImport` with a null asset for things like absolute imports (`import 'https://example.com/foo.js') or bare imports that we couldn't find. But we can't preload things like this anyway - or we shouldn't bother to in the case of an absolute assets.

For (B), trying to tighten things up and do LESS, if we don't need it.

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
+++ b/src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
@@ -61,15 +61,19 @@ final class JavaScriptImportPathCompiler implements AssetCompilerInterface
                 $dependentAsset = $this->findAssetForRelativeImport($importedModule, $asset, $assetMapper);
             }
 
+            if (!$dependentAsset) {
+                return $fullImportString;
+            }
+
             // List as a JavaScript import.
             // This will cause the asset to be included in the importmap (for relative imports)
             // and will be used to generate the preloads in the importmap.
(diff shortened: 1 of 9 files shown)


stof (approved):

fabpot:
Thank you @weaverryan.

Issue #52326, opened 2023-10-27
symfony/yaml PHP 8.1 Compatibility


timnolte:
### Symfony version(s) affected

6.3.3

### Description

The `trim()` method will no longer accept a null value.

https://github.com/symfony/yaml/blob/6.3/Inline.php#L62

Deprecation warnings are thrown when this method is used.

### How to reproduce

```php
Call `Inline::parse( null );
```

### Possible Solution

Need to validate that `$value` is not `null` before attempting to `trim()`.

nicolas-grekas:
This class is internal.
Did you find a way to do this call while using only non-internal API?

stof:
@nicolas-grekas we should still check why our argument has a default value being `null` though...

timnolte:
@nicolas-grekas I think we probably need to dig in deeper to how we are using this method tbh. So I'll do some digging into this as well.

ghost:
> This class is internal. Did you find a way to do this call while using only non-internal API?

I think the author reports deprecation warnings he wants to get rid of.

stof:
@javaDeveloperKid but then, we still need to know which code path triggers that deprecation, to find where we end up passing null to this method.

ghost:
@stof we can't rule out that someone violates `@internal` annotation

timnolte:
@javaDeveloperKid & @stof if you look at the function the default value if none is supplied is `null`. https://github.com/symfony/yaml/blob/6.3/Inline.php#L58

```php
public static function parse(string $value = null, int $flags = 0, array &$references = []): mixed
```

timnolte:
FYI, I patched this code using a `null` check with an early return before the `trim()` even happens.

GromNaN:
The class is marked as `@internal`. This method must not be called outside of the Yaml component. Everywhere we call it with a string as first argument. There is no supported case of passing `null` as first argument.
If your code uses this function, you must fix it.
I submitted a PR for 6.4 to disallow `null` value https://github.com/symfony/symfony/pull/52334

timnolte:
@GromNaN FYI, we are not calling that method directly actually. We are calling `Yaml::dump()` & `Yaml::parse()`. https://github.com/symfony/yaml/blob/6.3/Yaml.php#L91
https://github.com/symfony/yaml/blob/6.3/Yaml.php#L73

GromNaN:
Could you provide an example? So that we can fix the `Yaml` class if necessary.

timnolte:
[...]

PR #52332, opened 2023-10-27
[Yaml] Fix deprecated passing null to trim()

ghost:
| Q             | A
| ------------- | ---
| Branch?       |5.4 
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? |no 
| Issues        | Closes https://github.com/symfony/symfony/issues/52326 
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Yaml/Inline.php
+++ b/src/Symfony/Component/Yaml/Inline.php
@@ -60,6 +60,10 @@ class Inline
      */
     public static function parse(string $value = null, int $flags = 0, array &$references = [])
     {
+        if (null === $value) {
+            return '';
+        }
+
         self::initialize($flags);
 
         $value = trim($value);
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/Yaml/Inline.php
```diff
@@ -59,7 +59,7 @@ public static function parse(string $value = null, int $flags = 0, array &$refer
     {
         self::initialize($flags);
 
-        $value = trim($value);
+        $value = null === $value ? '' : trim($value);
```
GromNaN: The early return could be put first.

```suggestion
        if (null === $value) {
            return '';
        }

        self::initialize($flags);

        $value = trim($value);
```

We may change the return value to `null` is a future version.

GromNaN (commented):
This PR must target the oldest maintained branch where the deprecation exists: `5.4`.

You should add a test case in `InlineTest::getTestsForParse()`.

derrabus (approved):

xabbuh (approved):

GromNaN (approved):

fabpot (approved):

fabpot:
Thank you @javaDeveloperKid.

PR #52336, opened 2023-10-27
[HttpFoundation][Lock] Makes MongoDB adapters usable with `ext-mongodb` directly

GromNaN:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

[`mongodb/mongodb`](https://packagist.org/packages/mongodb/mongodb) is complex to handle for libraries with optional support of MongoDB, as it requires [`ext-mongodb`](https://www.php.net/manual/en/set.mongodb.php). In order to reduce complexity for maintainers, I reimplemented the session and lock adapters to use only the C driver classes.

Some features of `MongoDB\Client` are missing (server selection, session, transaction). But they are not necessary to store Sessions and Lock.

Changes:
- Lock & Session accept a `MongoDB\Driver\Manager`
- The lock uses exclusively UTC date. Before, there was a mix of `time()` and `UTCDatetime` objects.
- Session tests require a mongo server.
- `mongodb/mongodb` not needed in the CI

And of course also allows developers to use this adapters without installing `mongodb/mongodb` if they want, with the same features as before.

### DIFF
--- a/src/Symfony/Component/HttpFoundation/CHANGELOG.md
+++ b/src/Symfony/Component/HttpFoundation/CHANGELOG.md
@@ -9,6 +9,7 @@ CHANGELOG
  * Add `UriSigner` from the HttpKernel component
  * Add `partitioned` flag to `Cookie` (CHIPS Cookie)
  * Add argument `bool $flush = true` to `Response::send()`
+* Make `MongoDbSessionHandler` instantiable with the mongodb extension directly
 
 6.3
 ---
--- a/src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
+++ b/src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
@@ -14,20 +14,22 @@ namespace Symfony\Component\HttpFoundation\Session\Storage\Handler;
 use MongoDB\BSON\Binary;
 use MongoDB\BSON\UTCDateTime;
 use MongoDB\Client;
-use MongoDB\Collection;
+use MongoDB\Driver\BulkWrite;
+use MongoDB\Driver\Manager;
+use MongoDB\Driver\Query;
 
 /**
- * Session handler using the mongodb/mongodb package and MongoDB driver extension.
+ * Session handler using the MongoDB driver extension.
  *
  * @author Markus Bachmann <markus.bachmann@bachi.biz>
+ * @author Jérôme Tamarelle <jerome@tamarelle.net>
  *
- * @see https://packagist.org/packages/mongodb/mongodb
  * @see https://php.net/mongodb
  */
 class MongoDbSessionHandler extends AbstractSessionHandler
 {
-    private Client $mongo;
-    private Collection $collection;
+    private Manager $manager;
+    private string $namespace;
     private array $options;
     private int|\Closure|null $ttl;
 
@@ -62,13 +64,18 @@ class MongoDbSessionHandler extends AbstractSessionHandler
      *
      * @throws \InvalidArgumentException When "database" or "collection" not provided
      */
-    public function __construct(Client $mongo, array $options)
+    public function __construct(Client|Manager $mongo, array $options)
     {
         if (!isset($options['database']) || !isset($options['collection'])) {
             throw new \InvalidArgumentException('You must provide the "database" and "collection" option for MongoDBSessionHandler.');
         }
 
-        $this->mongo = $mongo;
+        if ($mongo instanceof Client) {
+            $mongo = $mongo->getManager();
+        }
+
+        $this->manager = $mongo;
+        $this->namespace = $options['database'].'.'.$options['collection'];
 
         $this->options = array_merge([
             'id_field' => '_id',
@@ -86,36 +93,47 @@ class MongoDbSessionHandler extends AbstractSessionHandler
 
     protected function doDestroy(#[\SensitiveParameter] string $sessionId): bool
     {
-        $this->getCollection()->deleteOne([
-            $this->options['id_field'] => $sessionId,
-        ]);
+        $write = new BulkWrite();
+        $write->delete(
+            [$this->options['id_field'] => $sessionId],
+            ['limit' => 1]
+        );
+
+        $this->manager->executeBulkWrite($this->namespace, $write);
 
         return true;
     }
 
     public function gc(int $maxlifetime): int|false
     {
-        return $this->getCollection()->deleteMany([
-            $this->options['expiry_field'] => ['$lt' => new UTCDateTime()],
-        ])->getDeletedCount();
+        $write = new BulkWrite();
+        $write->delete(
+            [$this->options['expiry_field'] => ['$lt' => $this->getUTCDateTime()]],
+        );
+        $result = $this->manager->executeBulkWrite($this->namespace, $write);
+
+        return $result->getDeletedCount() ?? false;
     }
 
     protected function doWrite(#[\SensitiveParameter] string $sessionId, string $data): bool
     {
         $ttl = ($this->ttl instanceof \Closure ? ($this->ttl)() : $this->ttl) ?? \ini_get('session.gc_maxlifetime');
-        $expiry = new UTCDateTime((time() + (int) $ttl) * 1000);
+        $expiry = $this->getUTCDateTime($ttl);
 
         $fields = [
-            $this->options['time_field'] => new UTCDateTime(),
+            $this->options['time_field'] => $this->getUTCDateTime(),
             $this->options['expiry_field'] => $expiry,
-            $this->options['data_field'] => new Binary($data, Binary::TYPE_OLD_BINARY),
+            $this->options['data_field'] => new Binary($data, Binary::TYPE_GENERIC),
         ];
 
-        $this->getCollection()->updateOne(
+        $write = new BulkWrite();
+        $write->update(
             [$this->options['id_field'] => $sessionId],
             ['$set' => $fields],
             ['upsert' => true]
         );
+
+        $this->manager->executeBulkWrite($this->namespace, $write);
 
         return true;
     }
@@ -123,40 +141,46 @@ class MongoDbSessionHandler extends AbstractSessionHandler
     public function updateTimestamp(#[\SensitiveParameter] string $sessionId, string $data): bool
     {
         $ttl = ($this->ttl instanceof \Closure ? ($this->ttl)() : $this->ttl) ?? \ini_get('session.gc_maxlifetime');
-        $expiry = new UTCDateTime((time() + (int) $ttl) * 1000);
+        $expiry = $this->getUTCDateTime($ttl);
 
-        $this->getCollection()->updateOne(
+        $write = new BulkWrite();
+        $write->update(
             [$this->options['id_field'] => $sessionId],
             ['$set' => [
-                $this->options['time_field'] => new UTCDateTime(),
+                $this->options['time_field'] => $this->getUTCDateTime(),
                 $this->options['expiry_field'] => $expiry,
-            ]]
+            ]],
+            ['multi' => false],
         );
+
+        $this->manager->executeBulkWrite($this->namespace, $write);
 
         return true;
     }
 
     protected function doRead(#[\SensitiveParameter] string $sessionId): string
     {
-        $dbData = $this->getCollection()->findOne([
+        $cursor = $this->manager->executeQuery($this->namespace, new Query([
             $this->options['id_field'] => $sessionId,
-            $this->options['expiry_field'] => ['$gte' => new UTCDateTime()],
-        ]);
+            $this->options['expiry_field'] => ['$gte' => $this->getUTCDateTime()],
+        ], [
+            'projection' => [
+                '_id' => false,
+                $this->options['data_field'] => true,
+            ],
+            'limit' => 1,
+        ]));
 
-        if (null === $dbData) {
-            return '';
+        foreach ($cursor as $document) {
+            return (string) $document->{$this->options['data_field']} ?? '';
         }
 
-        return $dbData[$this->options['data_field']]->getData();
+        // Not found
+        return '';
     }
 
-    private function getCollection(): Collection
+    private function getUTCDateTime(int $additionalSeconds = 0): UTCDateTime
     {
-        return $this->collection ??= $this->mongo->selectCollection($this->options['database'], $this->options['collection']);
-    }
-
-    protected function getMongo(): Client
-    {
-        return $this->mongo;
+        return new UTCDateTime((time() + $additionalSeconds) * 1000);
     }
 }
--- a/src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
+++ b/src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
@@ -11,56 +11,98 @@
 
 namespace Symfony\Component\HttpFoundation\Tests\Session\Storage\Handler;
 
+use MongoDB\BSON\Binary;
+use MongoDB\BSON\UTCDateTime;
 use MongoDB\Client;
-use PHPUnit\Framework\MockObject\MockObject;
-use PHPUnit\Framework\SkippedTestSuiteError;
+use MongoDB\Driver\BulkWrite;
+use MongoDB\Driver\Command;
+use MongoDB\Driver\Exception\CommandException;
+use MongoDB\Driver\Exception\ConnectionException;
+use MongoDB\Driver\Manager;
+use MongoDB\Driver\Query;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\HttpFoundation\Session\Storage\Handler\MongoDbSessionHandler;
+
+require_once __DIR__.'/stubs/mongodb.php';
 
 /**
  * @author Markus Bachmann <markus.bachmann@bachi.biz>
  *
+ * @group integration
  * @group time-sensitive
  *
  * @requires extension mongodb
  */
 class MongoDbSessionHandlerTest extends TestCase
 {
-    public array $options;
-    private MockObject&Client $mongo;
-    private MongoDbSessionHandler $storage;
+    private const DABASE_NAME = 'sf-test';
+    private const COLLECTION_NAME = 'session-test';
 
-    public static function setUpBeforeClass(): void
-    {
-        if (!class_exists(Client::class)) {
-            throw new SkippedTestSuiteError('The mongodb/mongodb package is required.');
-        }
-    }
+    public array $options;
+    private Manager $manager;
+    private MongoDbSessionHandler $storage;
 
     protected function setUp(): void
     {
         parent::setUp();
 
-        $this->mongo = $this->getMockBuilder(Client::class)
-            ->disableOriginalConstructor()
-            ->getMock();
+        $this->manager = new Manager('mongodb://'.getenv('MONGODB_HOST'));
+
+        try {
+            $this->manager->executeCommand(self::DABASE_NAME, new Command(['ping' => 1]));
+        } catch (ConnectionException $e) {
+            $this->markTestSkipped(sprintf('MongoDB Server "%s" not running: %s', getenv('MONGODB_HOST'), $e->getMessage()));
+        }
 
         $this->options = [
             'id_field' => '_id',
             'data_field' => 'data',
             'time_field' => 'time',
             'expiry_field' => 'expires_at',
-            'database' => 'sf-test',
-            'collection' => 'session-test',
+            'database' => self::DABASE_NAME,
+            'collection' => self::COLLECTION_NAME,
         ];
 
-        $this->storage = new MongoDbSessionHandler($this->mongo, $this->options);
+        $this->storage = new MongoDbSessionHandler($this->manager, $this->options);
     }
 
-    public function testConstructorShouldThrowExceptionForMissingOptions()
+    public function testCreateFromClient()
+    {
+        $client = $this->createMock(Client::class);
+        $client->expects($this->once())
+            ->method('getManager')
+            ->willReturn($this->manager);
+
+        $this->storage = new MongoDbSessionHandler($client, $this->options);
+        $this->storage->write('foo', 'bar');
+
+        $this->assertCount(1, $this->getSessions());
+    }
+
+    protected function tearDown(): void
+    {
+        try {
+            $this->manager->executeCommand(self::DABASE_NAME, new Command(['drop' => self::COLLECTION_NAME]));
+        } catch (CommandException $e) {
+            // The server may return a NamespaceNotFound error if the collection does not exist
+            if (26 !== $e->getCode()) {
+                throw $e;
+            }
+        }
+    }
+
+    /** @dataProvider provideInvalidOptions */
+    public function testConstructorShouldThrowExceptionForMissingOptions(array $options)
     {
         $this->expectException(\InvalidArgumentException::class);
-        new MongoDbSessionHandler($this->mongo, []);
+        new MongoDbSessionHandler($this->manager, $options);
+    }
+
+    public function provideInvalidOptions()
+    {
+        yield 'empty' => [[]];
+        yield 'collection missing' => [['database' => 'foo']];
+        yield 'database missing' => [['collection' => 'foo']];
     }
 
     public function testOpenMethodAlwaysReturnTrue()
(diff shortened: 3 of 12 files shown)


src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
```diff
@@ -86,77 +94,92 @@ public function close(): bool
...
+            ['limit' => 1]
+        );
+
+        $this->manager->executeBulkWrite($this->namespace, $write);
 
         return true;
     }
 
-    public function gc(int $maxlifetime): int|false
+    public function gc(int $maxlifetime): int
```
OskarStark: We used getDeletedCount() before and had int|false as return type.

Now narrow it down to int. 

Was it wrong before?

GromNaN: Indeed, I didn't realized that changing this return type would be breaking change. Reverted.

nicolas-grekas:
Some test failures are related in the integration job apparently.

GromNaN:
> Some test failures are related in the integration job apparently.

Thanks. I fixed the tests, the CI uses an old version of the ext-mongodb. But that's fine, we don't set version constraint.

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
```diff
@@ -54,163 +59,141 @@ protected function setUp(): void
...
+        $this->storage = new MongoDbSessionHandler($this->manager, $this->options);
     }
 
-    public function testConstructorShouldThrowExceptionForMissingOptions()
+    public function testCreateFromClient(): void
     {
-        $this->expectException(\InvalidArgumentException::class);
-        new MongoDbSessionHandler($this->mongo, []);
+        if (!class_exists(Client::class)) {
+            $this->markTestSkipped('The "mongodb/mongodb" library is required.');
```
GromNaN: This test still requires the `mongodb/mongodb` library. It will not run in the CI. In order to mock the Client class, what do you think of creating stubs?

src/Symfony/Component/HttpFoundation/composer.json
```diff
@@ -25,6 +25,7 @@
         "doctrine/dbal": "^2.13.1|^3|^4",
         "predis/predis": "^1.1|^2.0",
         "symfony/cache": "^6.3|^7.0",
+        "symfony/clock": "^6.3|^7.0",
```
GromNaN: To be reverted if we don't use the clock.

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
```diff
@@ -86,77 +94,92 @@ public function close(): bool
...
-        return $dbData[$this->options['data_field']]->getData();
+        return $data->get($this->options['data_field']);
     }
 
-    private function getCollection(): Collection
+    private function getUTCDateTime(int $additionalSeconds = 0): UTCDateTime
     {
-        return $this->collection ??= $this->mongo->selectCollection($this->options['database'], $this->options['collection']);
-    }
+        $timestamp = (new \DateTimeImmutable('now', new \DateTimeZone('UTC')));
```
GromNaN: This is time sensitive. What do you think of using the clock here, if available. That would help testing expiration.

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
```diff
@@ -86,77 +94,92 @@ public function close(): bool
...
+        $write = new BulkWrite();
+        $write->update(
             [$this->options['id_field'] => $sessionId],
             ['$set' => [
-                $this->options['time_field'] => new UTCDateTime(),
+                $this->options['time_field'] => $this->getUTCDateTime(),
                 $this->options['expiry_field'] => $expiry,
-            ]]
+            ]],
+            ['limit' => 1],
```
alcaeus: `limit` is not an available options for `BulkWrite::update`, but only for `BulkWrite::delete`. For updates, the `multi` option is used to decide whether one or all matching documents are updated. As this option defaults to `false`, you can safely remove the `limit` option here:

```suggestion
```

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
```diff
@@ -86,77 +94,92 @@ public function close(): bool
...
     protected function doRead(#[\SensitiveParameter] string $sessionId): string
     {
-        $dbData = $this->getCollection()->findOne([
+        $dbData = $this->manager->executeQuery($this->namespace, new Query([
             $this->options['id_field'] => $sessionId,
-            $this->options['expiry_field'] => ['$gte' => new UTCDateTime()],
-        ]);
+            $this->options['expiry_field'] => ['$gte' => $this->getUTCDateTime()],
+        ]));
+        $dbData->setTypeMap(['root' => 'bson']);
```
alcaeus: I'll note that `MongoDB\BSON\Document` and the `bson` type map type have been introduced in version 1.16 of the driver. I didn't see a requirement or conflict in this component's composer.json, but wanted to point this out as this change might silently break things if people are running an older extension version.

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
```diff
@@ -86,77 +94,92 @@ public function close(): bool
...
 
     protected function doRead(#[\SensitiveParameter] string $sessionId): string
     {
-        $dbData = $this->getCollection()->findOne([
+        $dbData = $this->manager->executeQuery($this->namespace, new Query([
             $this->options['id_field'] => $sessionId,
-            $this->options['expiry_field'] => ['$gte' => new UTCDateTime()],
-        ]);
+            $this->options['expiry_field'] => ['$gte' => $this->getUTCDateTime()],
+        ]));
```
alcaeus: Since we're only working with the data field, it might be sensible to also add a projection to the query:

```suggestion
            $this->options['expiry_field'] => ['$gte' => $this->getUTCDateTime()],
        ], [
            'projection' => [$this->options['data_field'] => true]
        ]));
```

While the metadata is not that big, reducing the result document to just binary data would almost completely eliminate the overhead of using the default driver type map, allowing you to skip working with raw BSON and thus changing the driver version requirement.

(Apologies for the suboptimal formatting, but GitHub wouldn't let me add a suggestion for the whole block due to the "deleted" line).

src/Symfony/Component/Lock/Store/MongoDbStore.php
```diff
@@ -88,7 +92,7 @@ class MongoDbStore implements PersistingStoreInterface
      * readPreference is primary for all queries.
      * @see https://docs.mongodb.com/manual/applications/replication/
      */
-    public function __construct(Collection|Client|string $mongo, array $options = [], float $initialTtl = 300.0)
+    public function __construct(Collection|Client|Manager|string $mongo, array $options = [], float $initialTtl = 300.0)
```
GromNaN: I haven't added `MongoDB\Database` as intermediary accepted type. There is no real use-case but that would be easy to implement.
The main motivation would be to prevent unexpected error if someone pass an instance of `MongoDB\Database` (Stringable), it would be be casted to `string` and used as DSN.

src/Symfony/Component/Lock/Store/MongoDbStore.php
```diff
@@ -142,6 +149,10 @@ public function __construct(Collection|Client|string $mongo, array $options = []
      */
     private function skimUri(string $uri): string
```
alcaeus, line 154: Leaving this here: note that as per the [Connection Strings documentation](https://www.mongodb.com/docs/manual/reference/connection-string/), the path in the connection string refers to the default authentication database to be used if the `authSource` does not define one. While most people consider this to be the default database, I wanted to point out the difference. Not necessarily an issue of this PR as it's already the current behaviour, so feel free to resolve this comment after reading if you think this isn't worth changing.

src/Symfony/Component/Lock/Store/MongoDbStore.php
```diff
@@ -257,23 +273,35 @@ public function putOffExpiration(Key $key, float $ttl)
...
     {
-        $this->getCollection()->deleteOne([ // filter
-            '_id' => (string) $key,
-            'token' => $this->getUniqueToken($key),
-        ]);
+        $write = new BulkWrite();
+        $write->delete(
+            [ // filter
+                '_id' => (string) $key,
+                'token' => $this->getUniqueToken($key),
```
alcaeus, line 284: Out of curiosity, with the ID already uniquely identifying a record, what's the reason for adding a condition on `token` as well?

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
GromNaN: What should I use to ensure compatibility with older versions? `array`?

OskarStark: As this is for 6.4, add a requirement for 1.16?

src/Symfony/Component/Lock/Store/MongoDbStore.php
GromNaN: The uniqueness of `_id` ensure only 1 process can acquire this lock. The random token is used to identify which process owns the lock. It's necessary here to ensure we don't release a lock owned by an other process.

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
GromNaN: As [mentioned](https://github.com/symfony/symfony/pull/52336/files#r1375872731) by Andreas, the BSON optimisation is not necessary if we fetch a single field. Updated to use the default `stdClass`.

src/Symfony/Component/Lock/Store/MongoDbStore.php
GromNaN: Sure, that would be better if the option `database` had precedence over the database in the path of the URI.

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
GromNaN: 💡 Very smart.

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
GromNaN: Too bad I didn't get an error message to say so.
Updated to pass `multi: false` explicitly.

src/Symfony/Component/Lock/Store/MongoDbStore.php
alcaeus: Ah, I see. Thanks for explaining!

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
nicolas-grekas: I wouldn't use the clock component to keep deps simple.

GromNaN: Ok 👍🏻

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
GromNaN: Added minimal stubs.

src/Symfony/Component/Lock/Store/MongoDbStore.php
GromNaN: Added support for `MongoDB\Database`

nicolas-grekas (approved):
LGTM for 6.4, I like that this makes the CI simpler !

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
```diff
@@ -11,39 +11,46 @@
...
 
+require_once __DIR__.'/stubs/mongodb.php';
+
 /**
  * @author Markus Bachmann <markus.bachmann@bachi.biz>
  *
  * @group time-sensitive
+ * @group integration
  *
  * @requires extension mongodb
```
stof, line 34: is the extension still required to run tests or are the stubs enough ?

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
```diff
@@ -11,39 +11,46 @@
...
 use PHPUnit\Framework\TestCase;
+use Symfony\Component\Clock\MockClock;
 use Symfony\Component\HttpFoundation\Session\Storage\Handler\MongoDbSessionHandler;
 
+require_once __DIR__.'/stubs/mongodb.php';
+
 /**
  * @author Markus Bachmann <markus.bachmann@bachi.biz>
  *
  * @group time-sensitive
```
stof: as you use `new DatetimeImmutable('now')` in the code rather than `time()`, I don't think you are compatible with the ClockMock of the phpunit bridge, which would make this group useless AFAICT.

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
```diff
@@ -11,39 +11,46 @@
...
 use MongoDB\Client;
-use PHPUnit\Framework\MockObject\MockObject;
-use PHPUnit\Framework\SkippedTestSuiteError;
+use MongoDB\Driver\BulkWrite;
+use MongoDB\Driver\Command;
+use MongoDB\Driver\Exception\ConnectionException;
+use MongoDB\Driver\Manager;
+use MongoDB\Driver\Query;
 use PHPUnit\Framework\TestCase;
+use Symfony\Component\Clock\MockClock;
```
stof: The clock component needs to be added as a dev dependency.

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
```diff
@@ -11,39 +11,46 @@
...
+ * @group integration
  *
  * @requires extension mongodb
  */
 class MongoDbSessionHandlerTest extends TestCase
 {
     public array $options;
-    private MockObject&Client $mongo;
+    private Manager $manager;
+    private MockClock $clock;
```
stof: is this property actually used ?

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
GromNaN: Good catch. The ruins of an abandoned beginning of implementation.

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
GromNaN: Removed.

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
GromNaN: Yes, the extension is required, not the library. 
Stubs are for some library classes.

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
GromNaN: Removed.

GromNaN (commented):
Thanks for the review @stof. I removed clock related code.

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
```diff
@@ -86,77 +93,97 @@ public function close(): bool
...
-        return $this->getCollection()->deleteMany([
-            $this->options['expiry_field'] => ['$lt' => new UTCDateTime()],
-        ])->getDeletedCount();
+        $write = new BulkWrite();
+        $write->delete(
+            [$this->options['expiry_field'] => ['$lt' => $this->getUTCDateTime()]],
+        );
+        $result = $this->manager->executeBulkWrite($this->namespace, $write);
+
+        return $result->getDeletedCount() ?? false;
```
jmikola, line 115: Note: `null` is only [returned for unacknowledged writes](https://www.php.net/manual/en/mongodb-driver-writeresult.getdeletedcount.php), so the `false` case should not happen in practice here. Although I suppose it's possible if the user passed in such a Manager with `w=0` in its connection string.

And even beyond that, the static analysis tooling would probably require handling the `null` case anyway.

But beyond all of that, [SessionHandlerInterface](https://www.php.net/manual/en/sessionhandlerinterface.gc.php) docs note that `false` indicates _failure_. An unacknowledged write isn't a failure, so I reckon it'd only make sense to return `false` if you were maybe catching and suppressing an exception. Otherwise, defaulting to `0` seems sensible.

In the original code, `DeleteResult::getDeletedCount()` would have [raised a BadMethodCallException](https://github.com/mongodb/mongo-php-library/blob/1.16.1/src/DeleteResult.php#L49), which doesn't seem like a good user experience either.

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
```diff
@@ -86,77 +93,97 @@ public function close(): bool
...
-
-        if (null === $dbData) {
+            $this->options['expiry_field'] => ['$gte' => $this->getUTCDateTime()],
+        ], [
+            'projection' => [
+                '_id' => false,
+                $this->options['data_field'] => true,
+            ],
+        ]));
+        $dbData->setTypeMap(['root' => 'object']);
```
jmikola: Is this necessary? A root document is already returned as `stdClass` by default.

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
```diff
@@ -86,77 +93,97 @@ public function close(): bool
...
+        $dbData->rewind();
+        $data = $dbData->current();
+
+        if (!$data) {
             return '';
         }
 
-        return $dbData[$this->options['data_field']]->getData();
+        return (string) $data->{$this->options['data_field']};
```
jmikola: Is it preferable to explicitly call `rewind()` and `current()`?

This could be rewritten as:

```php
foreach ($dbData as $data) {
    return (string) $data->{$this->options['data_field']};
}

return '';
```

Or even:

```php
return (string) $dbData->toArray()[0]->{$this->options['data_field']} ?? '';
```

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
```diff
@@ -86,77 +93,97 @@ public function close(): bool
...
-        $dbData = $this->getCollection()->findOne([
+        $dbData = $this->manager->executeQuery($this->namespace, new Query([
             $this->options['id_field'] => $sessionId,
-            $this->options['expiry_field'] => ['$gte' => new UTCDateTime()],
-        ]);
-
-        if (null === $dbData) {
+            $this->options['expiry_field'] => ['$gte' => $this->getUTCDateTime()],
+        ], [
+            'projection' => [
```
jmikola, line 167: Does anything enforce that `$this->options['id_field']` is actually `_id` or even a field with a unique index?

I assume this will only ever return one or no documents, but I noticed you use `limit: 1` for the delete operation in `doDestroy()`, but omit the limit here. I don't disagree with an explicit limit on the delete operation, since the default is to remove all matching documents.

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
```diff
@@ -86,77 +93,97 @@ public function close(): bool
...
-        return $dbData[$this->options['data_field']]->getData();
+        return (string) $data->{$this->options['data_field']};
     }
 
-    private function getCollection(): Collection
+    private function getUTCDateTime(int $additionalSeconds = 0): UTCDateTime
     {
-        return $this->collection ??= $this->mongo->selectCollection($this->options['database'], $this->options['collection']);
-    }
+        $timestamp = new \DateTimeImmutable('now', new \DateTimeZone('UTC'));
```
jmikola: I don't think the overhead of constructing these objects is significant, but is this preferable to using `time()` directly if you're just going to end up fetching that timestamp below?

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
```diff
@@ -11,39 +11,43 @@
...
     {
         parent::setUp();
 
-        $this->mongo = $this->getMockBuilder(Client::class)
-            ->disableOriginalConstructor()
-            ->getMock();
+        $this->manager = new Manager('mongodb://'.getenv('MONGODB_HOST'));
+
+        try {
+            $this->manager->executeCommand('sf-test', new Command(['ping' => 1]));
```
jmikola: You can consider defining a constant for the database and/or collection name to avoid duplication of string literals in this file.

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
```diff
@@ -54,13 +58,39 @@ protected function setUp(): void
...
+        $this->storage = new MongoDbSessionHandler($client, $this->options);
+        $this->storage->write('foo', 'bar');
+
+        $this->assertCount(1, $this->getSessions());
     }
 
-    public function testConstructorShouldThrowExceptionForMissingOptions()
+    protected function tearDown(): void
+    {
+        $this->manager->executeCommand('sf-test', new Command(['drop' => 'session-test']));
```
jmikola: Note that the `drop` command may fail if the collection doesn't exist. Since this is a general `tearDown()` method, that could be problematic if some test was added to this file and didn't actually write to the collection to implicitly create it.

Consider suppressing the relevant exception code as is done in PHPLIB's [DropCollection](https://github.com/mongodb/mongo-php-library/blob/master/src/Operation/DropCollection.php#L108).

Looking at `testConstructorShouldThrowExceptionForMissingOptions()` below, I'm surprised that this doesn't already fail since that test never gets far enough to write to the collection.

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
```diff
@@ -86,77 +93,97 @@ public function close(): bool
...
     {
         $ttl = ($this->ttl instanceof \Closure ? ($this->ttl)() : $this->ttl) ?? \ini_get('session.gc_maxlifetime');
-        $expiry = new UTCDateTime((time() + (int) $ttl) * 1000);
+        $expiry = $this->getUTCDateTime($ttl);
 
         $fields = [
-            $this->options['time_field'] => new UTCDateTime(),
+            $this->options['time_field'] => $this->getUTCDateTime(),
             $this->options['expiry_field'] => $expiry,
             $this->options['data_field'] => new Binary($data, Binary::TYPE_OLD_BINARY),
```
jmikola: You should change this to [`TYPE_GENERIC`](https://www.php.net/manual/en/class.mongodb-bson-binary.php#mongodb-bson-binary.constants.type-generic) while you're modifying this file, as `TYPE_OLD_BINARY` is deprecated. There should be no BC concern here, since you're always casting the Binary to a string when reading the session data later.

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
```diff
@@ -75,142 +105,93 @@ public function testCloseMethodAlwaysReturnTrue()
...
     {
-        $collection = $this->getMockBuilder(\MongoDB\Collection::class)
-            ->disableOriginalConstructor()
-            ->getMock();
+        $timestamp = (new \DateTimeImmutable())->getTimestamp() + $timeDiff;
+
+        $write = new BulkWrite();
+        $write->insert([
+            '_id' => $sessionId,
+            'data' => new Binary($data, Binary::TYPE_OLD_BINARY),
```
jmikola: Per my earlier comment, I think it'd be preferable to use [`TYPE_GENERIC`](https://www.php.net/manual/en/class.mongodb-bson-binary.php#mongodb-bson-binary.constants.type-generic) here.

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
```diff
@@ -75,142 +105,93 @@ public function testCloseMethodAlwaysReturnTrue()
...
+        return $this->manager->executeQuery('sf-test.session-test', new Query([]))->toArray();
     }
 
-    private function createMongoCollectionMock(): \MongoDB\Collection
+    private function insertSession(string $sessionId, string $data, int $timeDiff): void
     {
-        $collection = $this->getMockBuilder(\MongoDB\Collection::class)
-            ->disableOriginalConstructor()
-            ->getMock();
+        $timestamp = (new \DateTimeImmutable())->getTimestamp() + $timeDiff;
```
jmikola: In the session handler, you explicitly passed `new \DateTimeZone('UTC')` as the second parameter. Here, you're going to create a timestamp using the local time zone per [`DateTimeImmutable::__construct()`](https://www.php.net/manual/en/datetimeimmutable.construct.php). Is that problematic?

Would `time()` just be simpler here, as I suggested in a previous comment?

src/Symfony/Component/HttpFoundation/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ CHANGELOG
  * Add `UriSigner` from the HttpKernel component
  * Add `partitioned` flag to `Cookie` (CHIPS Cookie)
  * Add argument `bool $flush = true` to `Response::send()`
+* `MongoDbSessionHandler` can be used with the mongodb extension only
```
jmikola: This reads as if PHPLIB is no longer supported at all, which is not the case since you extract a Manager from a Client. Consider:

```suggestion
* `MongoDbSessionHandler` can be used with the mongodb extension directly
```

Alternatively: "can now be used"

src/Symfony/Component/Lock/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+6.4
+---
+
+ * `MongoDbStore` can be used with the mongodb extension only
```
jmikola: Per my earlier suggestion in the main changelog:

```suggestion
 * `MongoDbStore` can be used with the mongodb extension directly
```

Optionally: "can now be used"

src/Symfony/Component/Lock/Store/MongoDbStore.php
```diff
@@ -108,21 +113,27 @@ public function __construct(Collection|Client|string $mongo, array $options = []
...
 
-        if (!($mongo instanceof Collection)) {
-            if (null === $this->options['database']) {
-                throw new InvalidArgumentException(sprintf('"%s()" requires the "database" in the URI path or option.', __METHOD__));
-            }
-            if (null === $this->options['collection']) {
-                throw new InvalidArgumentException(sprintf('"%s()" requires the "collection" in the URI querystring or option.', __METHOD__));
-            }
+        if (null === $this->options['database']) {
+            throw new InvalidArgumentException(sprintf('"%s()" requires the "database" in the URI path or option.', __METHOD__));
```
jmikola, line 131: What is "in the URI path" referring to? We should not be using the [auth database](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#auth-database-optional) component in the URI for anything beyond a default auth database in the driver, which itself is deprecated in favor of the `authSource` URI option.

Is there any code in Symfony that actually parses the database name from the URI? If so, I'm missing it here.

Edit: found `skimUri()` below. See my later comment about collection parsing.

src/Symfony/Component/Lock/Store/MongoDbStore.php
```diff
@@ -108,21 +113,27 @@ public function __construct(Collection|Client|string $mongo, array $options = []
...
         }
+        if (null === $this->options['collection']) {
+            throw new InvalidArgumentException(sprintf('"%s()" requires the "collection" in the URI querystring or option.', __METHOD__));
```
jmikola, line 134: Likewise, I don't see anything in the code here that would parse a collection name from the URI. There isn't a documented connection string option for that, so I'm not sure what this would even be referring to (hopefully not passing a namespace in the [auth database](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#auth-database-optional) component).

Edit: just found the `skimUri()` function below, which parses the auth DB component and reads a `collection` query string parameter. I'd propose adding some comment in this file about why this isn't ideal so the explanations aren't only found in old PRs (this isn't the first time I've brought this up).

src/Symfony/Component/Lock/Store/MongoDbStore.php
jmikola: Note that [authSource](https://www.mongodb.com/docs/manual/reference/connection-string/#mongodb-urioption-urioption.authSource) should never be assumed to be a database name. For some auth mechanisms it might be set to `$external` (see: [authSource docs](https://www.mongodb.com/docs/manual/reference/connection-string/#mongodb-urioption-urioption.authSource)).

src/Symfony/Component/Lock/Store/MongoDbStore.php
```diff
@@ -257,23 +277,35 @@ public function putOffExpiration(Key $key, float $ttl)
      */
     public function delete(Key $key)
     {
-        $this->getCollection()->deleteOne([ // filter
-            '_id' => (string) $key,
-            'token' => $this->getUniqueToken($key),
-        ]);
+        $write = new BulkWrite();
+        $write->delete(
+            [ // filter
```
jmikola:
```suggestion
            [
```

This seemed self-explanatory, and you didn't have any such comments in the session handler class.

src/Symfony/Component/Lock/Store/MongoDbStore.php
```diff
@@ -257,23 +277,35 @@ public function putOffExpiration(Key $key, float $ttl)
...
+                '_id' => (string) $key,
+                'token' => $this->getUniqueToken($key),
+                'expires_at' => [
+                    '$gt' => $this->createMongoDateTime(microtime(true)),
+                ],
             ],
-        ], [
-            'readPreference' => new ReadPreference(\defined(ReadPreference::PRIMARY) ? ReadPreference::PRIMARY : ReadPreference::RP_PRIMARY),
-        ]);
+            [ // options
```
jmikola: Consider removing these `// options` and `// filter` comments everywhere for consistency with the session handler class.

src/Symfony/Component/Lock/Store/MongoDbStore.php
```diff
@@ -257,23 +277,35 @@ public function putOffExpiration(Key $key, float $ttl)
...
+                'limit' => 1,
+                'projection' => ['_id' => 1],
+            ]
+        ));
+
+        return [] !== $cursor->toArray();
```
jmikola, line 308: No objections here, but comparing [`iterator_count()`](https://www.php.net/iterator_count) to zero is another option. Although that's probably a bit more overhead as it's going to end up doing exactly what `toArray()` does internally anyway.

src/Symfony/Component/Lock/Store/MongoDbStore.php
```diff
@@ -326,20 +361,9 @@ private function isDuplicateKeyException(WriteException $e): bool
...
-
-        $this->client ??= new Client($this->uri, $this->options['uriOptions'], $this->options['driverOptions']);
-
-        $this->collection = $this->client->selectCollection(
-            $this->options['database'],
-            $this->options['collection']
-        );
-
-        return $this->collection;
+        return $this->manager ??= new Manager($this->uri, $this->options['uriOptions'], $this->options['driverOptions']);
```
jmikola, line 366: Offhand, is there anything in these Symfony adapters that sets driver information like we do for Doctrine ODM and Laravel? If so, should we consider doing so in a subsequent PR? /cc @alcaeus

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
GromNaN: Yes, that's what I've found too. But phpstan warns that `getDeletedCount` returns `int|null`, which doesn't match with the return type of the function `int|false`. So I did this to satisfy the static analysis.

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
GromNaN: Removed this 2nd parameter to rely on the default `Binary::TYPE_GENERIC`

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
GromNaN: I guess this options were copied from `PdoSessionHandler` when this provider was created. This provider has a `configureSchema` or `createTable` methods to init the table. We can add a similar method for MongoDB. But since the collection already have an unique index on `_id`, I think it's fine to let people that would like to change the id field, initialize the collection themselve.

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
GromNaN: This was to be explicit, in case we change the default value. But I removed this line.

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
GromNaN: Good point, `time()` is not timezone-sensitive.

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
GromNaN: I reverted the `@group time-sensitive` and usage of `time()`.

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
GromNaN: Added constants `DATABASE_NAME` and `COLLECTION_NAME`.

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
GromNaN: Just tested, it looks like this operation is idempotent and never throw an exception when the database doesn't exist.
Same behavior with mongosh in compass.

src/Symfony/Component/Lock/CHANGELOG.md
GromNaN: "can now" is implicit, since this is a **change**log.

Updated both changelogs with "Make `<class>` instantiable with the mongodb extension directly"

src/Symfony/Component/Lock/Store/MongoDbStore.php
GromNaN: Removed this comments that were inherited from previous implementation.

src/Symfony/Component/Lock/Store/MongoDbStore.php
GromNaN: The current syntax is the way we check for empty arrays in Symfony codebase.

src/Symfony/Component/Lock/Store/MongoDbStore.php
GromNaN: The connection will mostly come from Doctrine, but I can add user-agent for standalone usage.

jmikola: I think it makes sense if you're constructing the Manager yourself, but no need to hold up this PR to add that if you want to address it down the line.

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
jmikola: The no-op behavior was only recently implemented for collections in server 7.0 ([SERVER-43894](https://jira.mongodb.org/browse/SERVER-43894)). For databases, it was implemented in server 5.0 ([SERVER-55145](https://jira.mongodb.org/browse/SERVER-55145)).

GromNaN: Added the try...catch for compatibility with mongodb 6 and lower.

src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MongoDbSessionHandler.php
GromNaN: The 2nd argument was required until the driver version 1.15
https://github.com/mongodb/mongo-php-driver/commit/9df0785fea9c8cea3a2c2ba976088c011e3deae0

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
```diff
@@ -11,56 +11,98 @@
...
+        $this->assertCount(1, $this->getSessions());
     }
 
-    public function testConstructorShouldThrowExceptionForMissingOptions()
+    protected function tearDown(): void
+    {
+        try {
+            $this->manager->executeCommand(self::DABASE_NAME, new Command(['drop' => self::COLLECTION_NAME]));
+        } catch (CommandException $e) {
+            // The server may return an error if the collection does not exist
```
jmikola:
```suggestion
            // The server may return a NamespaceNotFound error if the collection does not exist
```

If we're not creating a constant for `26` below, I think it's a good idea to note the error name here for reference.

jmikola (approved):

src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/MongoDbSessionHandlerTest.php
GromNaN: Fixed.

fabpot:
Thank you @GromNaN.

PR #52342, opened 2023-10-28
[Console] Strip tags before escaping text in `OutputFormatter`

### LINKED ISSUE #50007: SymfonyQuestionHelper undesirably escapes formatted options when rendering default choice value

6.2.7

#### Abstract

When using the choice helper, I sometimes find it helpful to include colour formatting in an option, for example to distinguish a "magic" choice, like seen in the attached screenshot. However, when the default option is formatted, the source formatting string is undesirably rendered, as the [`SymfonyQuestionHelper`](https://github.com/symfony/symfony/blob/9de1fd76b5509d31e3b99b112c7ad21a7eb32171/src/Symfony/Component/Console/Helper/SymfonyQuestionHelper.php#L62) method escapes it.

#### Actual

image

#### Expected

image

### How to reproduce

Create a choice question where the default option uses console output formatting, like here, (using Laravel Artisan as a wrapper)

```php
$this->choice('Select a tag', [
    '<comment>Add new tag</comment>',
    'hello',
    'world',
], default: 0);
```

#### This outputs the following:
```
 Select a tag [<comment>Add new tag</comment>]:
  [0] Add new tag
  [1] hello
  [2] world
 >
```

#### Whereas the following would be expected

```
 Select a tag [Add new tag]:
  [0] Add new tag
  [1] hello
  [2] world
 >
```

### Possible Solution

I have narrowed this down to the following line, which runs the escape:

https://github.com/symfony/symfony/blob/9de1fd76b5509d31e3b99b112c7ad21a7eb32171/src/Symfony/Component/Console/Helper/SymfonyQuestionHelper.php#L62

I think the best solution for my edge case would be if the escaping was/could be disabled to allow the colour to pass through, as seen in the screenshot below, but that would likley add a considerable amount of complexity, furthermore, I assume there is a reason the escapement was added in the first place.

So instead, I would go with the absolute easiest solution, stripping the tags before escaping:

```php
$text = sprintf(' <info>%s</info> [<comment>%s</comment>]:', $text, OutputFormatter::escape(strip_tags($choices[$default] ?? $default)));
```

I would be more than happy to make a pull request for this.

### Additional Context

I am running Symfony Console through the Laravel Artisan based [Laravel Zero](https://github.com/laravel-zero/laravel-zero), version `v10.0.0`. Using Windows 10 Terminal (PowerShell) but I do not think that matters, as I don't see any logic in the source close to the method call which would be affected by the runner system.

emmadesilva:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #50007
| License       | MIT

This fixes https://github.com/symfony/symfony/issues/50007 by adding a `strip_tags` call before running the `OutputFormatter::escape()` logic.

### DIFF
--- a/src/Symfony/Component/Console/Formatter/OutputFormatter.php
+++ b/src/Symfony/Component/Console/Formatter/OutputFormatter.php
@@ -40,7 +40,7 @@ class OutputFormatter implements WrappableOutputFormatterInterface
      */
     public static function escape(string $text): string
     {
-        $text = preg_replace('/([^\\\\]|^)([<>])/', '$1\\\\$2', $text);
+        $text = preg_replace('/([^\\\\]|^)([<>])/', '$1\\\\$2', strip_tags($text));
 
         return self::escapeTrailingBackslash($text);
     }
--- a/src/Symfony/Component/Console/Tests/Formatter/OutputFormatterTest.php
+++ b/src/Symfony/Component/Console/Tests/Formatter/OutputFormatterTest.php
@@ -115,6 +115,16 @@ class OutputFormatterTest extends TestCase
         );
     }
 
+    public function testTagEscaping()
+    {
+        $formatter = new OutputFormatter(true);
+
+        $this->assertEquals(
+            'foo',
+            $formatter->escape('<info>foo</info>')
+        );
+    }
+
     public function testDeepNestedStyles()
     {
         $formatter = new OutputFormatter(true);
--- a/src/Symfony/Component/Console/Tests/Helper/FormatterHelperTest.php
+++ b/src/Symfony/Component/Console/Tests/Helper/FormatterHelperTest.php
@@ -83,9 +83,9 @@ class FormatterHelperTest extends TestCase
         $formatter = new FormatterHelper();
 
         $this->assertEquals(
-            '<error>                              </error>'."\n".
-            '<error>  \<info\>some info\</info\>  </error>'."\n".
-            '<error>                              </error>',
+            '<error>             </error>'."\n".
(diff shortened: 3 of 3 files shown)


emmadesilva:
Sorry, it seems like I targeted the wrong branch. First time PRing Symfony. Changing target messed stuff up, how should we handle this best?

OskarStark:
All good, we will change the branch while merging, no worries

emmadesilva:
> All good, we will change the branch while merging, no worries

Okay great! Let me take a look at the failing tests and I'll report back!

emmadesilva:
Okay, so it seems that this causes these two tests to fail:

https://github.com/symfony/symfony/blob/11c60aa432da965d45fe4fcc4659f2fd80db1053/src/Symfony/Bridge/Twig/Tests/Command/DebugCommandTest.php#L266-L273

https://github.com/symfony/symfony/blob/11c60aa432da965d45fe4fcc4659f2fd80db1053/src/Symfony/Bridge/Twig/Tests/Command/DebugCommandTest.php#L275-L283

I can update these tests, but am wondering if this is an acceptable side-effect. @OskarStark What do you think?

OskarStark:
Can you please add a testcase to avoid further regressions? Thanks

emmadesilva:
> Can you please add a testcase to avoid further regressions? Thanks

I think the sideeffects of this are large enough to warrant a different approach. I'm going to try fixing this closer in the call stack to the original issue.

PR #52349, opened 2023-10-28
[AssetMapper] Fix in-file imports to resolve via filesystem

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | None
| License       | MIT

Hi!

Another fix from further testing in the complex world!

`tl;dr` Relative imports in CSS & JS files are now resolved via the filesystem path, not the AssetMapper "logical path".

**Full Explanation**

In AssetMapper, we configure different paths/directories, which exposes them to the public directory. If you need to fetch the info about an asset, you can use its "logical path".

However, when you're inside of a JavaScript or CSS file, these "asset mapper" paths or "logical asset paths" are correctly irrelevant. Instead, you should code as if all of your files are already public and won't be moved around:

```js
// /path/to/project/assets/app.js

// (1️⃣) 👍 /path/to/project/assets/files/app.js
import './files/other.js'; 
// (2️⃣) 👍 /path/to/project/vendor/foo/bar/assets/baz.js
import '../vendor/foo/bar/assets/baz.js

// (3️⃣) 👎
// this is a valid asset mapper "logical path", but this is nonsense and doesn't work
// well, you could add this to the importmap. But "asset mapper" paths still have no relevancy
import 'files/other.js';
```

This PR fixes case 2️⃣ which previously did not work. Before, we used `../vendor/foo/bar/assets/baz.js` and the "asset mapper logical path" of the importing file (`app.js`) and tried to construct a final, logical path from that. But because of the `../vendor`, that went above the `app.js` root, and we panicked. Even if we didn't panic, logical paths should not be relevant here. Instead, it's simpler: the relative import in `app.js` uses the filesystem path. Of course, if the final file - `/vendor/foo/bar/assets/baz.js` is not, itself, in *some* AssetMapper path, an clear exception is thrown (if the file is not exposed publicly, we can't resolve a proper path to it).

Cheers!


src/Symfony/Component/AssetMapper/Compiler/CssAssetUrlCompiler.php
```diff
@@ -38,23 +37,29 @@ public function compile(string $content, MappedAsset $asset, AssetMapperInterfac
     {
         return preg_replace_callback(self::ASSET_URL_PATTERN, function ($matches) use ($asset, $assetMapper) {
             try {
-                $resolvedPath = $this->resolvePath(\dirname($asset->logicalPath), $matches[1]);
+                $resolvedSourcePath = Path::join(\dirname($asset->sourcePath), $matches[1]);
```
weaverryan, line 40: A bit unrelated: let's let `Path` do the work

src/Symfony/Component/AssetMapper/Compiler/JavaScriptImportPathCompiler.php
```diff
@@ -165,26 +164,32 @@ private function findAssetForRelativeImport(string $importedModule, MappedAsset
...
+        $dependentAsset = $assetMapper->getAssetFromSourcePath($resolvedSourcePath);
 
         if ($dependentAsset) {
             return $dependentAsset;
         }
 
+        // avoid warning about vendor imports - these are often comments
+        if ($asset->isVendor) {
+            return null;
+        }
```
weaverryan, line 176: This just represents a reorg so we can return earlier

src/Symfony/Component/AssetMapper/Compiler/CssAssetUrlCompiler.php
```diff
@@ -38,23 +37,29 @@ public function compile(string $content, MappedAsset $asset, AssetMapperInterfac
     {
         return preg_replace_callback(self::ASSET_URL_PATTERN, function ($matches) use ($asset, $assetMapper) {
             try {
```
weaverryan, line 39: It looks like we can ignore this. In practice, our `preg_replace_callback()` will return a string. But if there is some change we want to make to satisfy psalm, we can.

fabpot (approved):

fabpot:
Thank you @weaverryan.

PR #52368, opened 2023-10-30
[AssetMapper] Fixing bug where JSCompiler used non-absolute importmap entry path

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | None
| License       | MIT

Introduced over the weekend in #52349.

The `path` key inside `importmap.php` can be a full filesystem path or it can start with `./` to indicate a relative path to the root directory. That has been the case for weeks. In #52349, when start using the `path` of an importmap entry to find the `MappedAsset`. When doing that, I didn't consider the case where `path` is a relative path.

This PR fixes that and moves the code handling this into `ImportMapConfigReader` so that we're not repeating it in several places.

Cheers!


nicolas-grekas:
Thank you @weaverryan.

PR #52369, opened 2023-10-30
[DependencyInjection] Add `urlencode` function to `EnvVarProcessor`

crtl:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

This PR adds `urlencode` to the default `EnvVarProcessor` to allow encoding environment variables in config files.
The comes handy when environment variables are provided by the OS and not through .env because those values are often not  compatible with DSN syntax for doctrine or mailer components.

In my case our app was deployed on an AWS stack and environment variables for SMPT and database where automatically provided in the deployment process and replaced in the respective .env files.
Because those values where not encoded they would cause issues with DSN format requireing me to commit them in my `.env.prod` which circumvents the purpose of the whole system.

Usage:
```yaml
# config/packages/mailer.yaml
framework:
    mailer:
        dsn: 'smtp://%env(urlencode:SMTP_USER)%:%env(urlencode:SMTP_PASSWORD)%@%env(urlencode:SMTP_HOST)%?encryption=%env(SMTP_ENCRYPTION)%&port=%env(SMTP_PORT)%&auth_mode=login'  
```

### DIFF
--- a/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
+++ b/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
@@ -57,6 +57,7 @@ class EnvVarProcessor implements EnvVarProcessorInterface
             'enum' => \BackedEnum::class,
             'shuffle' => 'array',
             'defined' => 'bool',
+            'urlencode' => 'string',
         ];
     }
 
@@ -342,6 +343,10 @@ class EnvVarProcessor implements EnvVarProcessorInterface
 
         if ('trim' === $prefix) {
             return trim($env);
+        }
+
+        if ('urlencode' === $prefix) {
+            return rawurlencode($env);
         }
 
         throw new RuntimeException(sprintf('Unsupported env var prefix "%s" for env name "%s".', $prefix, $name));
--- a/src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php
+++ b/src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php
@@ -948,6 +948,17 @@ CSV;
         $this->assertSame($expected, (new EnvVarProcessor(new Container()))->getEnv('defined', 'NO_SOMETHING', $callback));
     }
 
+    public function testGetEnvUrlencode()
+    {
+        $processor = new EnvVarProcessor(new Container());
+
+        $result = $processor->getEnv('urlencode', 'URLENCODETEST', function () {
+            return 'foo: Data123!@-_ + bar: Not the same content as Data123!@-_ +';
+        });
+
+        $this->assertSame('foo%3A%20Data123%21%40-_%20%2B%20bar%3A%20Not%20the%20same%20content%20as%20Data123%21%40-_%20%2B', $result);
+    }
+
     public static function provideGetEnvDefined(): iterable
     {
         yield 'Defined' => [true, fn () => 'foo'];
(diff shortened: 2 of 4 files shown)


src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
```diff
@@ -344,6 +345,10 @@ public function getEnv(string $prefix, string $name, \Closure $getEnv): mixed
             return trim($env);
         }
 
+        if ('urlencode' === $prefix) {
+            return urlencode($env);
```
GromNaN: `rawurlencode` should be used instead. It encodes spaces as `%20` while `rawurlencode` encode them as `+`.

src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php
```diff
@@ -956,4 +967,5 @@ public static function provideGetEnvDefined(): iterable
         yield 'Null' => [false, fn () => null];
         yield 'Env var not defined' => [false, fn () => throw new EnvNotFoundException()];
     }
+
```
GromNaN: Please apply the patch provided by fabbot in the CI.

GromNaN (commented):
Great idea. I've already had a project that crashed because a `&` slipped into an automatically generated password.

src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
crtl: Ive replaced it.

src/Symfony/Component/DependencyInjection/Tests/EnvVarProcessorTest.php
crtl: If you were talking about the empty line I have  removed it.

GromNaN: > [Review](https://fabbot.io/report/symfony/symfony/52369/b792342c6f24f3d04f8f860a237aa61d1ef30d05#patchcs) the proposed patch then [download](https://fabbot.io/patch/symfony/symfony/52369/b792342c6f24f3d04f8f860a237aa61d1ef30d05/cs.diff) it to apply it manually or execute the following command from the repository root directory:

```console
curl https://fabbot.io/patch/symfony/symfony/52369/b792342c6f24f3d04f8f860a237aa61d1ef30d05/cs.diff | patch -p0
```

src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
```diff
@@ -78,7 +79,7 @@ public function getEnv(string $prefix, string $name, \Closure $getEnv): mixed
             }
 
             if (!isset($array[$key]) && !\array_key_exists($key, $array)) {
-                throw new EnvNotFoundException(sprintf('Key "%s" not found in %s (resolved from "%s").', $key, json_encode($array), $next));
+                throw new EnvNotFoundException(sprintf('Key "%s" not found in "%s" (resolved from "%s").', $key, json_encode($array), $next));
```
GromNaN: This is an incorrect fix from fabbot.io. To be reverted, sorry.

GromNaN (approved):
SGTM.
We are feature freeze in preparation of 6.4 and 7.0. This will be for version 7.1.

OskarStark (approved):

fabpot (approved):

fabpot:
Thank you @crtl.

PR #52374, opened 2023-10-31
Fix tests

OskarStark:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | --
| License       | MIT

It looks like `tearDown()` methods sets properties null, which aren't nullable.

PR #52379, opened 2023-10-31
[Tests] Use `unset()` instead of nullable properties

OskarStark:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | --
| License       | MIT

Follows
* #52374


xabbuh:
Why do we have the `tearDown()` method at all. I don't see the real value in unsetting the properties (or setting them to `null`).

OskarStark:
I was thinking the same as we always have the `setUp` method run

sstok:
Properly because garbage collection cannot detect that these references are no longer used? Or does PHPUnit handle this transparently?

stof (approved):

xabbuh (approved):

nicolas-grekas:
Changes reverted in 714ca9b24c

PR #52380, opened 2023-10-31
[Tests] Add `array` return type for provider methods

OskarStark:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | --
| License       | MIT


derrabus (approved):
LGTM. Not sure if we want to fix stuff like that on old branches though.

nicolas-grekas:
We merged a few similar changes this morning but I agree: 5.4 is for bug fixes only, and this is not a bug fix.
I'm still going to merge because this is minor but let's follow this principle more strictly.

nicolas-grekas:
Thank you @OskarStark.

PR #52381, opened 2023-10-31
[WebProfilerBundle] do not install FrameworkBundle `6.4.0-BETA1`

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

### DIFF
--- a/src/Symfony/Bundle/WebProfilerBundle/composer.json
+++ b/src/Symfony/Bundle/WebProfilerBundle/composer.json
@@ -18,7 +18,7 @@
     "require": {
         "php": ">=7.2.5",
         "symfony/config": "^4.4|^5.0|^6.0",
-        "symfony/framework-bundle": "^5.3|^6.0",
+        "symfony/framework-bundle": "^5.3|^6.0,<6.4",
         "symfony/http-kernel": "^5.3|^6.0",
         "symfony/polyfill-php80": "^1.16",
         "symfony/routing": "^4.4|^5.0|^6.0",


OskarStark (approved):

derrabus:
Shall we have FrameworkBundle 6.4 conflict with WebProfilerBundle 5.4 instead? Otherwise, Composer might decide to pin WebProfilerBundle 5.4 to an old patch release.

xabbuh:
> Shall we have FrameworkBundle 6.4 conflict with WebProfilerBundle 5.4 instead? Otherwise, Composer might decide to pin WebProfilerBundle 5.4 to an old patch release.

We have in BETA2, but we didn't have one in BETA1. Maybe we should simply conflict with BETA1?

nicolas-grekas:
Thank you @xabbuh.

src/Symfony/Bundle/WebProfilerBundle/composer.json
```diff
@@ -18,7 +18,7 @@
     "require": {
         "php": ">=7.2.5",
         "symfony/config": "^4.4|^5.0|^6.0",
-        "symfony/framework-bundle": "^5.3|^6.0",
+        "symfony/framework-bundle": "^5.3|^6.0,<6.4",
```
nicolas-grekas, line 21: I updated the way the constraint is expressed, but this means the same

PR #52384, opened 2023-10-31
[Console] Add `Command::addUsages()` method

### LINKED ISSUE #48974: [Console] Add multiple usages at once
### Description

There is a `Command::addUsage()` there should also be a `Command::addUsages()`
(if we have a list of usages from a strings db it would be easier to add these)

(perhaps a ::setUsage too?)
``` php
    public function setUsages(array $usages): static
    {
        $this->usages = $usages;

        return $this;
    }
```
(or empty the array and foreach over $usages and call addUsage)

### Example

_No response_

feamsr00:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Issues        | Fix #48974
| License       | MIT

Users can now add multiple command usages in one call.

``` php
protected function configure(): void
{
  $this->setName('MyCommand')
    ->setHelp(GetMyComandHelpText())
    ->addUsages(GetMyComandUsageStrings())
  ;
  //...
}
```


src/Symfony/Component/Console/Command/Command.php
```diff
@@ -684,6 +684,20 @@ public function addUsage(string $usage): static
         return $this;
     }
 
+    /**
+     * Adds a set of command usage examples.
+     *
+     * @param array[string] $usages
```
94noni:
```suggestion
     * @param string[] $usages
```

usually in the sf codebase, phpdoc is like so

94noni (approved):

src/Symfony/Component/Console/Command/Command.php
feamsr00: thanks!

OskarStark:
As we are already in feature freeze, this is for 7.1

maxbeckers (approved):

OskarStark:
Can you please rebase on top of `7.1` and move the changeling to the 7.1 section? Thanks

fabpot:
I don't think we need this additional method. It clutters the class and using `addUsage()` several times is good enough.

Issue #31348, opened 2019-05-01
[Intl] Parent/Child aliases are ignored


ro0NL:
**Symfony version(s) affected**: 3.4

After looking at #28070, shouldnt the fallback lookup mapping apply in the Intl component as well? Currrently the resource bundles only deal with the alias mapping :thinking: so this seems inconsistent.

Vice versa, the Translator component doesnt handle the alias locales.

No intend to solve it right now, but created an issue for reference.

PR #52388, opened 2023-10-31
Update es.php

Gonzalo1987:
| Q             | A
| ------------- | ---
| Branch?    |   5.4 and forward 
| Bug fix?  |    yes
| New feature?  | no 
| Deprecations?  | no 
| Issues        | 
| License       | MIT

Minor fix on the Ivory Coast translation to 'es'

### DIFF
--- a/src/Symfony/Component/Intl/Resources/data/regions/es.php
+++ b/src/Symfony/Component/Intl/Resources/data/regions/es.php
@@ -45,7 +45,7 @@ return [
         'CF' => 'República Centroafricana',
         'CG' => 'Congo',
         'CH' => 'Suiza',
-        'CI' => 'Côte d’Ivoire',
+        'CI' => 'Costa de Marfil',
         'CK' => 'Islas Cook',
         'CL' => 'Chile',
         'CM' => 'Camerún',


ro0NL:
it's upstream data ;) it should be commented on-top in the artifact files.

for reference, it's related to `es` vs `es_419` locales. Im not sure the fix is correct.

https://github.com/unicode-org/icu/blob/a7c7d8f214926ef23d2b54500ab77d95e5ea0068/icu4c/source/data/region/es.txt#L81

https://github.com/unicode-org/icu/blob/a7c7d8f214926ef23d2b54500ab77d95e5ea0068/icu4c/source/data/region/es_419.txt#L23

ro0NL:
the parent locales are also dealt with:

https://github.com/symfony/symfony/blob/5bdf2f0aca06f1e91b633d75e66e06b31bb4205d/src/Symfony/Component/Translation/Resources/data/parents.json#L98

but perhaps the issue is related to #31348

stof:
those files must not be edited manually. They are generated from the upstream data and any manual change would be lost on the next run of the command.

gonzalo1987:
Thank you very much! I'll try to PR in the upstream so.

PR #52397, opened 2023-10-31
[Validator] Add the `callbackArgs` option to the `Choice` constraint

BooleanType:
| Q             | A
| ------------- | ---
| Branch?       | 7.1 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT
| Doc PR | TODO

In my real project at work I need to validate `autocomplete` field, which is retrieved from route parameter. `autocomplete` may contain these values: "email", "email_2", "email_3", "email_4", ..., "email_10" (looks weird but these values are coming from another project which I can't influence).

Controller action:

```php
#[Route('/{autocomplete}', name: 'autocomplete')]
public function autocomplete(
     $autocomplete,
     SerializerInterface $serializer,
     ValidatorInterface $validator
): Response {
     // Fulfill DTO with route parameter.
     $dto = $serializer->deserialize(
           $serializer->serialize(['autocomplete' => $autocomplete]),
           AutocompleteDto::class,
           'json'
     );

     $violations = $validator->validate($dto);
 
     // ...
}
```

`AutocompleteDto` validates route parameter. I use `Choice` constraint with `callback` option:
```php
class AutocompleteDto
{
    #[Assert\Choice(
        callback: [ParticipantAutocomplete::class, 'getAvailableEmailAutocompletes']
    )]
    private $autocomplete;
    
    // ...
}
```

And below is `ParticipantAutocomplete` with `getAvailableEmailAutocompletes()` callback. This callback just returns "email", "email_2", "email_3", "email_4", ..., "email_10" series.
```php
class ParticipantAutocomplete
{
    // This number may change someday due to client request.
    const AUTOCOMPLETES_COUNT = 10;
    
    public static function getAvailableEmailAutocompletes(): array
    {
        $autocompletes = ['email'];

        for ($i = 2; $i <= static::AUTOCOMPLETES_COUNT; $i++) {
            $autocompletes[] = 'email_' . $i;
        }

        return $autocompletes;
    }
}
```

Everything works OK, but it turned out that there could be _several_ such series. They are based on `firstname` ("firstname", "firstname_2", ..., "firstname_10"), `lastname` ("lastname", "lastname_2", ...), `country`, `city`, etc.

Producing additional methods (like `getAvailableCountryAutocompletes()`, `getAvailableFirstnameAutocompletes()`, etc.) depending on autocompletes is not an option for me. So I thought it would be nice to allow arguments to be passed to the callback. My current commit contains the corresponding changes.

With allowed callback parameters I could do the following.

`Choice` definition in `AutocompleteDto`:

```php
#[Assert\Choice(
      callback: [ParticipantAutocomplete::class, 'getAvailableAutocompletes'],
      callbackArgs: ['email']
)]
```

`ParticipantAutocomplete::getAvailableAutocompletes()` definition:
```php
public static function getAvailableAutocompletes(string $autocompleteFieldName): array
{
      $autocompletes = [$autocompleteFieldName];

      for ($i = 2; $i <= static::AUTOCOMPLETES_COUNT; $i++) {
          $autocompletes[] = $autocompleteFieldName . '_' . $i;
      }

      return $autocompletes;
}
```

With `callbackArgs`, when I need new action, where another (not `email`) autocomplete should be used, I don't have to create new method in `ParticipantAutocomplete`, I just change value in `callbackArgs` option.


OskarStark:
Because of the feature freeze, this must target 7.1

BooleanType:
> Because of the feature freeze, this must target 7.1

I've changed base branch to 7.0.

Regarding automated tests failure, it looks unrelated to my changes.

OskarStark:
Please rebase on 7.1 to get rid of the commits, thanks

lyrixx:
Hello. Thanks for your contribution

However I'm not convinced by this pr. You can easily do that in your application. For example you can create as many methods as you need. And theses methods will call another internal method With a parameter.

xabbuh:
I am not convinced either. It doesn’t look like each of the provided examples couldn’t be already achieved with a few lines of code. The use case doesn’t look generic enough to justify the added feature.

derrabus:
I agree. Let's close then.

PR #52398, opened 2023-10-31
DX: re-apply PHP CS Fixer, partially

keradus:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Issues        | Fix CS 
| License       | MIT

One further step to clean the CS status-quo, aiming for all-green in long-term.
~90 files remaining, yet they may require some further update on tool itself

best to review commit by commit

### DIFF
--- a/src/Symfony/Bridge/ProxyManager/Tests/LazyProxy/PhpDumper/ProxyDumperTest.php
+++ b/src/Symfony/Bridge/ProxyManager/Tests/LazyProxy/PhpDumper/ProxyDumperTest.php
@@ -93,7 +93,7 @@ class ProxyDumperTest extends TestCase
     {
         return [
             [
-                (new Definition(__CLASS__)),
+                new Definition(__CLASS__),
                 'privates',
             ],
             [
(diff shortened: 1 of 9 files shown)


src/Symfony/Component/Uid/BinaryUtil.php
```diff
@@ -128,7 +128,7 @@ public static function hexToDateTime(string $time): \DateTimeImmutable
 
             if (self::TIME_OFFSET_BIN <= $time) {
                 $time = self::add($time, self::TIME_OFFSET_COM2);
-                $time[0] = $time[0] & "\x7F";
+                $time[0] &= "\x7F";
```
nicolas-grekas: AFAIK, this is not the same for string indexes. Would need double checking if this was fixed in recent versions of PHP.

keradus: What is the string index here? (not the `[0]`, right?)

It is causing CI to fail, let's revert for now indeed

nicolas-grekas: yes it is, same for the other similar changes in the PR

keradus: > this is not the same for string indexes.

can you elaborate on that `string index` a bit? I'm not aware what is the underlying issue here.

> same for the other similar changes in the PR

after revert ot `&=` , CI is green (ok, except of fabbot for file license-headers)

nicolas-grekas: Sure: https://3v4l.org/GI1NW

nicolas-grekas:
Thank you @keradus.

src/Symfony/Component/Uid/BinaryUtil.php
keradus: Rule marked as risky, at least: https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/7418
perhaps, with time we can have better detection if we touch index or not

Issue #43162, opened 2021-09-24
[Process] Allow running multiple commands at once


m-vo:
**Description**
When running a lot of small shell commands after each other, the overhead of creating the underlying processes becomes noticeable. It would be handy if the process component would support chaining commands.

Working directly on a Linux/Windows shell you would typically do this:
```shell
# Linux
$ foo; bar; baz
$ foo && bar && baz

# Windows
$ foo & bar & baz
$ foo && bar && baz
```
(Yes there are other chaining operators as well like `||`.)

This is currently not possible with the process component.

**API Example**
I like that `Symfony\Component\Process\Process` became somewhat immutable now (the command line is baked in after the constructor ran), so the API could maybe look like this:

```php
$process = new Process(['foo']);

// foo; bar
$process = $process->withChainedCommand(['bar']); // returns a new process instance

// foo && bar
$process = $process->withAndChainedCommand(['bar']); // returns a new process instance
```

What do you think? Is this a worthwhile addition?

m-vo:
In non-windows environments, we're currently replacing the current shell (that PHP starts automatically) with the command to run by prepending it with `exec `. This makes sure signaling to the process is possible.

For chained commands this cannot work because there is no shell to start the subsequent commands (`exec foo && bar`). We could however prepend a pipe in that case so that subprocesses are started (`exec | foo && bar`), though I'm not 100% sure this will work in all cases. 

Ideas and insights are welcome. :relaxed: /cc @nicolas-grekas

derrabus:
Sounds like a good idea to me.

nicolas-grekas:
I'm not sure I would recommend adding `withChainedCommand` on the `Process` class.
But a shell command line builder would allow doing that I suppose.

m-vo:
> But a shell command line builder would allow doing that I suppose.

You mean a builder that creates shell system specific shell commands that are then treated as "raw input" like with `Process::fromShellCommandline()`? I think we would loose things like signaling then as well. 

We could also introduce a new `Operator` class that you could use like this:
```php
$process = new Process([
    'foo', Operator::chain(), 'bar', Operator::and(), 'baz'
]);
```

So the array of commands would be `array<string|Process>` and the operators are substituted with unescaped literals when starting the process (here: `foo; bar && baz` or `foo & bar && baz`).

nicolas-grekas:
[...]

PR #52399, opened 2023-10-31
Remove unused code from Process

ausi:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

While trying to work on https://github.com/symfony/symfony/issues/43162#issuecomment-1787670224 I noticed that the private member `useFileHandles` is not used for anything anymore.
It is always set in the constructor to `'\\' === \DIRECTORY_SEPARATOR` so it basically has a constant value and its only usage is in an `elseif` after a check to the very same constant value: `if ('\\' === \DIRECTORY_SEPARATOR)`.
Therefore this code does nothing and can be removed.

### DIFF
--- a/src/Symfony/Component/Process/Process.php
+++ b/src/Symfony/Component/Process/Process.php
@@ -77,7 +77,6 @@ class Process implements \IteratorAggregate
     private bool $pty;
     private array $options = ['suppress_errors' => true, 'bypass_shell' => true];
 
-    private bool $useFileHandles;
     private WindowsPipes|UnixPipes $processPipes;
 
     private ?int $latestSignal = null;
@@ -163,7 +162,6 @@ class Process implements \IteratorAggregate
 
         $this->setInput($input);
         $this->setTimeout($timeout);
-        $this->useFileHandles = '\\' === \DIRECTORY_SEPARATOR;
         $this->pty = false;
     }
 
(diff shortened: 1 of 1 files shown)


GromNaN:
I don't know this particular part of the code but usually this kind of comparison `DIRECTORY_SEPARATOR` is used to detect windows systems. As the value is different on Windows and Unix systems.

GromNaN (approved):
Ok, this condition is actually checked in the previous `if`.

chapterjason (approved):

src/Symfony/Component/Process/Process.php
```diff
@@ -322,7 +320,7 @@ public function start(callable $callback = null, array $env = [])
 
         if ('\\' === \DIRECTORY_SEPARATOR) {
             $commandline = $this->prepareWindowsCommandLine($commandline, $env);
-        } elseif (!$this->useFileHandles && $this->isSigchildEnabled()) {
+        } elseif ($this->isSigchildEnabled()) {
```
nicolas-grekas, line 326:
```suggestion
        } elseif ('\\' !== \DIRECTORY_SEPARATOR && $this->isSigchildEnabled()) {
```

chapterjason: The `elseif` will only be considered if `if ('\\' === \DIRECTORY_SEPARATOR)` (2 lines before) doesn't hit. Basically it will be checked twice but vice versa, as suggested change. 🤔

xabbuh (approved):

nicolas-grekas:
Thank you @ausi.

PR #52401, opened 2023-11-01
[Serializer] Fix issue related to handling namespace attributes in the `XmlEncoder` class

DevNwamini:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | --
| License       | MIT

The XmlEncoder in Symfony does not parse the namespace for a node without children, resulting in invalid XML when encoding. The suggested solution is to modify the XmlEncoder::parseXmlAttributes method to fetch the namespace from the node if it has one.

### DIFF
--- a/src/Symfony/Component/Serializer/Encoder/XmlEncoder.php
+++ b/src/Symfony/Component/Serializer/Encoder/XmlEncoder.php
@@ -261,33 +261,31 @@ class XmlEncoder implements EncoderInterface, DecoderInterface, NormalizationAwa
     /**
      * Parse the input DOMNode attributes into an array.
      */
-    private function parseXmlAttributes(\DOMNode $node, array $context = []): array
-    {
-        if (!$node->hasAttributes()) {
-            return [];
-        }
-
-        $data = [];
-        $typeCastAttributes = (bool) ($context[self::TYPE_CAST_ATTRIBUTES] ?? $this->defaultContext[self::TYPE_CAST_ATTRIBUTES]);
-
-        foreach ($node->attributes as $attr) {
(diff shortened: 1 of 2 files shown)


xabbuh:
Can you clarify what "the issue" issue is that you are trying to fix here please? Please also add a test that ensures we don't break the fixed behaviour in the future.

Youngemmy5956:
> Can you clarify what "the issue" issue is that you are trying to fix here please? Please also add a test that ensures we don't break the fixed behaviour in the future.

DevNwamini:
> Can you clarify what "the issue" issue is that you are trying to fix here please? Please also add a test that ensures we don't break the fixed behaviour in the future.

The XmlEncoder in Symfony does not parse the namespace for a node without children, resulting in invalid XML when encoding. The suggested solution is to modify the XmlEncoder::parseXmlAttributes method to fetch the namespace from the node if it has one.

OskarStark:
I cherry picked the test from
* https://github.com/symfony/symfony/pull/52414

And added the explanation to the PR header

OskarStark:
Your test case is failing, can you please check? Thanks

```
There was 1 failure:

1) Symfony\Component\Serializer\Tests\Encoder\XmlEncoderTest::testParseXmlAttributesWithNamespace
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
-    '@xmlns:ns' => 'http://example.com/'
     '@ns:attribute' => 'value'
+    '#' => ''
 )

/home/runner/work/symfony/symfony/src/Symfony/Component/Serializer/Tests/Encoder/XmlEncoderTest.php:412
```

mtarld:
Hey @Youngemmy5956, I didn't notice your PR and came up with the following: https://github.com/symfony/symfony/pull/52589.

Does it solve your issue?

fabpot:
Closing in favor of #52589

PR #52409, opened 2023-11-01
[Process] Pass the commandline as array to `proc_open()`

### LINKED ISSUE #43162: [Process] Allow running multiple commands at once

ausi:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix https://github.com/symfony/symfony/issues/43162#issuecomment-931145776
| License       | MIT

Since PHP 7.4, `proc_open()` accepts an array which makes it twice as fast (4x as fast with PHP 8.3), see https://github.com/symfony/symfony/issues/43162#issuecomment-1787670224

This pull request enables calling `proc_open()` with an array if:

1. `$this->commandline` is an array
2. and we are not on Windows
3. and PHP was compiled without `--enable-sigchild`

### DIFF
--- a/src/Symfony/Component/HttpClient/Tests/HttpClientTestCase.php
+++ b/src/Symfony/Component/HttpClient/Tests/HttpClientTestCase.php
@@ -327,7 +327,12 @@ abstract class HttpClientTestCase extends BaseHttpClientTestCase
             'KEY_FILE' => __DIR__.'/Fixtures/tls/server.key',
             'CERT_FILE' => __DIR__.'/Fixtures/tls/server.crt',
         ]);
-        $process->start();
+
+        try {
+            $process->start();
+        } catch (ProcessFailedException $e) {
+            self::markTestSkipped('vulcain failed: '.$e->getMessage());
+        }
 
         register_shutdown_function($process->stop(...));
         sleep('\\' === \DIRECTORY_SEPARATOR ? 10 : 1);
--- a/src/Symfony/Component/Process/Process.php
+++ b/src/Symfony/Component/Process/Process.php
@@ -15,6 +15,7 @@ use Symfony\Component\Process\Exception\InvalidArgumentException;
 use Symfony\Component\Process\Exception\LogicException;
 use Symfony\Component\Process\Exception\ProcessFailedException;
 use Symfony\Component\Process\Exception\ProcessSignaledException;
+use Symfony\Component\Process\Exception\ProcessStartFailedException;
 use Symfony\Component\Process\Exception\ProcessTimedOutException;
 use Symfony\Component\Process\Exception\RuntimeException;
 use Symfony\Component\Process\Pipes\UnixPipes;
@@ -233,11 +234,11 @@ class Process implements \IteratorAggregate
      *
      * @return int The exit status code
      *
-     * @throws RuntimeException         When process can't be launched
-     * @throws RuntimeException         When process is already running
-     * @throws ProcessTimedOutException When process timed out
-     * @throws ProcessSignaledException When process stopped after receiving signal
-     * @throws LogicException           In case a callback is provided and output has been disabled
+     * @throws ProcessStartFailedException When process can't be launched
+     * @throws RuntimeException            When process is already running
+     * @throws ProcessTimedOutException    When process timed out
+     * @throws ProcessSignaledException    When process stopped after receiving signal
+     * @throws LogicException              In case a callback is provided and output has been disabled
      *
      * @final
      */
@@ -284,9 +285,9 @@ class Process implements \IteratorAggregate
      * @param callable|null $callback A PHP callback to run whenever there is some
      *                                output available on STDOUT or STDERR
      *
-     * @throws RuntimeException When process can't be launched
-     * @throws RuntimeException When process is already running
-     * @throws LogicException   In case a callback is provided and output has been disabled
+     * @throws ProcessStartFailedException When process can't be launched
+     * @throws RuntimeException            When process is already running
+     * @throws LogicException              In case a callback is provided and output has been disabled
      */
     public function start(callable $callback = null, array $env = []): void
     {
@@ -306,12 +307,7 @@ class Process implements \IteratorAggregate
         $env += '\\' === \DIRECTORY_SEPARATOR ? array_diff_ukey($this->getDefaultEnv(), $env, 'strcasecmp') : $this->getDefaultEnv();
 
         if (\is_array($commandline = $this->commandline)) {
-            $commandline = implode(' ', array_map($this->escapeArgument(...), $commandline));
-
-            if ('\\' !== \DIRECTORY_SEPARATOR) {
-                // exec is mandatory to deal with sending a signal to the process
-                $commandline = 'exec '.$commandline;
-            }
+            $commandline = array_values(array_map(strval(...), $commandline));
         } else {
             $commandline = $this->replacePlaceholders($commandline, $env);
         }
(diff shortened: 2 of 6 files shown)


src/Symfony/Component/Process/Process.php
```diff
@@ -327,6 +322,11 @@ public function start(callable $callback = null, array $env = [])
             // last exit code is output on the fourth pipe and caught to work around --enable-sigchild
             $descriptors[3] = ['pipe', 'w'];
 
+            if (\is_array($commandline)) {
+                // exec is mandatory to deal with sending a signal to the process
+                $commandline = 'exec '.$this->buildShellCommandline($commandline);
+            }
```
ausi, line 324: I’m not sure if `exec ` is needed here, on my Platform (macOS) it does not make a difference.

ausi:
The failing unit test in PHP 8.3 seems to be a bug in PHP: https://github.com/php/php-src/issues/12589

OskarStark:
It is a new feature and must therefore target 7.1 because of feature freeze period

Should we wait wait for a PHP bugfix before merge?

ausi:
> It is a new feature and must therefore target 7.1 because of feature freeze period

Changed the base to 7.0 (there is no 7.1 branch yet)

> Should we wait wait for a PHP bugfix before merge?

I think so, yes.

ausi:
> > Should we wait wait for a PHP bugfix before merge?
> 
> I think so, yes.

The PHP bug got closed, the new behavior is intended. I updated the pull request to handle that case.

src/Symfony/Component/HttpClient/Tests/HttpClientTestCase.php
```diff
@@ -327,7 +328,12 @@ private static function startVulcain(HttpClientInterface $client)
             'KEY_FILE' => __DIR__.'/Fixtures/tls/server.key',
             'CERT_FILE' => __DIR__.'/Fixtures/tls/server.crt',
         ]);
-        $process->start();
+
+        try {
+            $process->start();
+        } catch (RuntimeException $e) {
+            self::markTestSkipped('vulcain failed: '.$e->getMessage());
+        }
```
derrabus, line 335: Unrelated change?

src/Symfony/Component/Process/Process.php
```diff
@@ -340,10 +340,17 @@ public function start(callable $callback = null, array $env = []): void
...
         }
 
+        $lastError = 'unknown reason';
+        set_error_handler(function ($type, $msg) use (&$lastError) {
+            $lastError = $msg;
+
+            return true;
+        });
         $process = @proc_open($commandline, $descriptors, $this->processPipes->pipes, $this->cwd, $envPairs, $this->options);
+        restore_error_handler();
```
derrabus:
```suggestion
        try {
            $process = @proc_open($commandline, $descriptors, $this->processPipes->pipes, $this->cwd, $envPairs, $this->options);
        } finally {
            restore_error_handler();
        }
```

src/Symfony/Component/HttpClient/Tests/HttpClientTestCase.php
ausi: No, by using `proc_open()` with an array, starting the process can fail directly since PHP 8.3 if the binary does not exist. This change is needed to restore the desired behavior of that test.

src/Symfony/Component/Process/Tests/ProcessTest.php
```diff
@@ -66,6 +66,24 @@ public function testInvalidCwd()
         $cmd->run();
     }
 
+    public function testInvalidCommand()
+    {
+        $cmds = [
+            new Process(['invalid']),
+            Process::fromShellCommandline('invalid'),
+        ];
```
OskarStark: What about using a data provider?

stof: This would indeed be better with a data provider or 2 tests rather than a loop inside the test emulating having multiple tests

ausi: Done in d603793013e2294d8f97c378b0256e86c8dfedb9.

ausi:
~~The PHP bug https://github.com/php/php-src/issues/12589 got reopened, so we should wait until that was resolved there I think.~~ 
UPDATE 2023-11-13: was closed again

ausi:
@carsonbot the 7.1 branch does not exist yet.

src/Symfony/Component/Process/Process.php
```diff
@@ -338,10 +338,20 @@ public function start(callable $callback = null, array $env = []): void
...
+        });
+        try {
+            $process = @proc_open($commandline, $descriptors, $this->processPipes->pipes, $this->cwd, $envPairs, $this->options);
+        } finally {
+            restore_error_handler();
+        }
 
         if (!\is_resource($process)) {
-            throw new RuntimeException('Unable to launch a new process.');
+            throw new RuntimeException('Unable to launch a new process: '.$lastError);
```
ausi: Should we throw a `ProcessFailedException` here? That would improve backwards compatibility I think.

Or should we maybe create a new exception like `ProcessStartFailedException`?

src/Symfony/Component/Process/Process.php
nicolas-grekas: I think so, let's keep it.

src/Symfony/Component/Process/Process.php
nicolas-grekas: ProcessFailedException would be fine to me

nicolas-grekas (commented):
Thanks for working on this!
Can't we use an array on Windows also? Sorry I didn't read everything :grimacing: 
While at it, WDYT of the warnings on appveyor :)
```
'where' is not recognized as an internal or external command,
operable program or batch file.
'where' is not recognized as an internal or external command,
operable program or batch file.
ERROR: Invalid argument or option - '/usr/local/php/bin/php-invalid'.
Type "WHERE /?" for usage help.
ERROR: Invalid argument or option - '/usr/local/php/bin/php-invalid'.
Type "WHERE /?" for usage help.
```

src/Symfony/Component/Process/Process.php
ausi: I created a new `ProcessStartFailedException` that extends the `ProcessFailedException` in 833bf1d2dbd88559d84d8528aa825c45df1bb86b.
This is a clean API I think and makes the changed behavior in PHP 8.3 a little bit more BC.

nicolas-grekas:
Thank you @ausi.

ausi:
> Can't we use an array on Windows also?

I don’t think so as we use some heavy `cmd.exe` magic on Windows. Also, there would be no real difference on Windows I think as internally PHP just converts the array back into a string and sets `bypass_shell = 1` which we also set.

Issue #51556, opened 2023-09-04
[Messenger] messenger:consume offers sync transports to consume from and then immediately throws an exception


dkarlovi:
### Symfony version(s) affected

6.3.3

### Description

When running `messenger:consume` without a transport, it asks to select which transport to consume from.

```
$ bin/console messenger:consume

                                                                                                                        
 Which transports/receivers do you want to consume?                                                                     
                                                                                                                        

Choose which receivers you want to consume messages from in order of priority.
Hint: to consume from multiple, use a list of their names, e.g. async, default

 Select receivers to consume: [async]:
  [0] async
  [1] default
```

But, if I select `default` (which is sync), it immediately throws

```
In SyncTransport.php line 37:
                                                                 
  You cannot receive messages from the Messenger SyncTransport.  
                                                                 
```

making this choice superfluous because the only possible selection is `async`.

### How to reproduce

See above.

### Possible Solution

Filter out sync transports and, if only one item remains, consume that immediately.

### Additional Context

I guess this could be considered an UX bug.

carsonbot:
Hey, thanks for your report!
There has not been a lot of activity here for a while. Is this bug still relevant? Have you managed to find a workaround?

dkarlovi:
Yup.

carsonbot:
Hey, thanks for your report!
There has not been a lot of activity here for a while. Is this bug still relevant? Have you managed to find a workaround?

carsonbot:
Just a quick reminder to make a comment on this. If I don't hear anything I'll close this.

dkarlovi:
AFAIK this wasn't fixed in other related PRs but was mentioned as a nice to have.

wazum:
I tried to solve this, please take a look at the pull request.

PR #52411, opened 2023-11-01
[Messenger] Add `--all` option to `messenger:consume`

### LINKED ISSUE #52364: [Messenger] Provide a way to consume all transports without specifying/knowing their names
### Description

When using https://github.com/php-enqueue/enqueue, a command:
```
enqueue:consume
```
is provided, in which there is no need to specify the consumers.

But when I run 
```
bin/console messenger:consume
```
I'm getting prompt about which transport I want to consume.

I saw in the code there was
```
$defaultReceiverName = 1 === \count($this->receiverNames) ? current($this->receiverNames) : null;
$defaultReceivers = $defaultReceiverName ? [$defaultReceiverName] : [];
```

What about something like --all in order to consume everything ? Would it be a good idea ?
Or maybe a commande should be added to list all the transports and then the output could be used for the consume command (or even to setup different supervisor with a foreach loop on the output ?)

I'd like to avoid updating my conf/commands
```
bin/console messenger:consume foo1,foo2,...
```
everytime.

### Example

_Originally posted by @VincentLanglet in https://github.com/symfony/symfony/discussions/51989_

ghost:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Issues        | Closes #52364 
| License       | MIT

When implementing this feature the problem with sync transports came out so the `if` statement for this was needed. I can see someone reported this 2 months ago in https://github.com/symfony/symfony/issues/51556. I think this be can fixed properly in a dedicated PR because it requires to dig into MessengerPass I guess.

### DIFF
--- a/src/Symfony/Component/Messenger/Command/ConsumeMessagesCommand.php
+++ b/src/Symfony/Component/Messenger/Command/ConsumeMessagesCommand.php
@@ -34,6 +34,7 @@ use Symfony\Component\Messenger\EventListener\StopWorkerOnMemoryLimitListener;
 use Symfony\Component\Messenger\EventListener\StopWorkerOnMessageLimitListener;
 use Symfony\Component\Messenger\EventListener\StopWorkerOnTimeLimitListener;
 use Symfony\Component\Messenger\RoutableMessageBus;
+use Symfony\Component\Messenger\Transport\Sync\SyncTransport;
 use Symfony\Component\Messenger\Worker;
 
 /**
@@ -83,6 +84,7 @@ class ConsumeMessagesCommand extends Command implements SignalableCommandInterfa
                 new InputOption('bus', 'b', InputOption::VALUE_REQUIRED, 'Name of the bus to which received messages should be dispatched (if not passed, bus is determined automatically)'),
                 new InputOption('queues', null, InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Limit receivers to only consume from the specified queues'),
                 new InputOption('no-reset', null, InputOption::VALUE_NONE, 'Do not reset container services after each message'),
+                new InputOption('all', null, InputOption::VALUE_NONE, 'Consume messages from all receivers'),
             ])
             ->setHelp(<<<'EOF'
 The <info>%command.name%</info> command consumes messages and dispatches them to the message bus.
@@ -123,6 +125,10 @@ Use the --queues option to limit a receiver to only certain queues (only support
 Use the --no-reset option to prevent services resetting after each message (may lead to leaking services' state between messages):
 
     <info>php %command.full_name% <receiver-name> --no-reset</info>
+
+Use the --all option to consume from all receivers:
+
+    <info>php %command.full_name% --all</info>
 EOF
             )
         ;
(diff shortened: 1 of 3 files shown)


OskarStark (approved):
Looks good to me

src/Symfony/Component/Messenger/Command/ConsumeMessagesCommand.php
```diff
@@ -135,6 +141,10 @@ protected function interact(InputInterface $input, OutputInterface $output)
     {
         $io = new SymfonyStyle($input, $output instanceof ConsoleOutputInterface ? $output->getErrorOutput() : $output);
 
+        if ($input->getOption('all')) {
+            return;
+        }
+
         if ($this->receiverNames && !$input->getArgument('receivers')) {
             $io->block('Which transports/receivers do you want to consume?', null, 'fg=white;bg=blue', ' ', true);
```
OskarStark, line 147: we can add a hint here (line 153) to also use the `all` option

OskarStark:
This must target 7.1 due to feature freeze period

src/Symfony/Component/Messenger/Command/ConsumeMessagesCommand.php
ghost: Hi. I started adding this but then I realized that the hint in line 153 relates to how to interact with the question (see the `e.g. receiver1, receiver2` part which works only when answering question, not when we pass multiple receivers as command arguments [commas are not allowed here]).

fabpot (approved):

fabpot:
Thank you @javaDeveloperKid.

Wirone:
Can you consider this as a bug fix and backport it to 6.4 😂 (jk)? This exactly something we need in our app, where we migrate maaaany workers using multiple transports (because our AMQP instance uses multiple exchanges). It would be much better to use `--all` instead of specifying all the transports (receivers) in the CLI command.

alexisLefebvre:
> Can you consider this as a bug fix and backport it to 6.4 😂 (jk)? This exactly something we need in our app, where we migrate maaaany workers using multiple transports (because our AMQP instance uses multiple exchanges). It would be much better to use `--all` instead of specifying all the transports (receivers) in the CLI command.

What if you use 7.1 in your app?

```shell
composer require symfony/messenger:^7.1
```

Update: I just realized that there is a `7.1` branch but it isn't released yet. It will be [released in May](https://symfony.com/releases/7.1), then upgrading should work and allow you to use the `--all` option.

In the meantime, using a commit or the Composer flag `@dev` for this package may work.

Wirone:
@alexislefebvre Thanks for the feedback, but actually I knew it before ☺️. I've added `@TODO` for migrating to 7.1 when it's available and then we'll use `--all`. Unfortunately `@dev` is not suitable for our system, which is too big and too important.

ghost:
@Wirone Another approach would be to copy past this class from 7.1 branch to your respository and register this in your composer,json. Composer will pick up the first class found in this namespace which will be yours. 
When you upgrade to 7.1+ one day, you will just remove this class.

P.S. Remember to check that changes between your 6.4.x locked in your composer.lock and 7.1 branch are only the ones from this PR. If not then decide if it's safe to use it.

Wirone:
FYI: I was able to introduce `--all` in Symfony 6.4 pretty easily - [here's how](https://blog.codito.dev/2024/04/messenger-consume-all-transports-on-symfony-6.4/) 😎.

PR #52418, opened 2023-11-02
[HttpClient] add native argument for the test server working directory

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

### DIFF
--- a/src/Symfony/Contracts/HttpClient/Test/TestHttpServer.php
+++ b/src/Symfony/Contracts/HttpClient/Test/TestHttpServer.php
@@ -18,12 +18,9 @@ class TestHttpServer
 {
     private static array $process = [];
 
-    /**
-     * @param string|null $workingDirectory
-     */
-    public static function start(int $port = 8057/* , string $workingDirectory = null */): Process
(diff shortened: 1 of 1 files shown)


GromNaN (approved):

src/Symfony/Contracts/HttpClient/Test/TestHttpServer.php
```diff
@@ -18,13 +18,8 @@ class TestHttpServer
 {
     private static array $process = [];
 
-    /**
-     * @param string|null $workingDirectory
-     */
-    public static function start(int $port = 8057/* , string $workingDirectory = null */): Process
+    public static function start(int $port = 8057, string $workingDirectory = __DIR__.'/Fixtures/web'): Process
```
nicolas-grekas: I think we should still accept null and have it be the default

xabbuh: I have pushed an update. Is that what you had in mind?

nicolas-grekas:
Thank you @xabbuh.

PR #52425, opened 2023-11-02
[Messenger] Dispatch events before & after each handler

BenMorel:
| Q             | A
| ------------- | ---
| Branch?       | 8.2
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

`HandleMessageMiddleware` now dispatches three events around each handler call:

- `HandlerStartingEvent`, right before the handler is invoked;
- `HandlerSuccessEvent`, once the handler succeeded;
- `HandlerFailureEvent`, once it failed, carrying the exception.

The worker events (`WorkerMessageHandledEvent`, `WorkerMessageFailedEvent`) are per message, only fire for messages consumed by a worker, and report several handlers as one aggregated outcome. These events are per handler, fire for sync and async dispatch alike, and carry the `HandlerDescriptor`, so auditing each handler run no longer requires diffing `HandledStamp`s and unwrapping `HandlerFailedException`.

For a **batch handler**, the outcome does not exist when the message is enqueued: it resolves through the `Acknowledger`, at flush time at the earliest. `HandlerStartingEvent` is dispatched when the handler is invoked, and the outcome event is dispatched by the acknowledger callback once the result or error is known, whether the handler acknowledges synchronously or at a later flush. Every handler run gets exactly one outcome event.

The events are dispatched on the `event_dispatcher` service wired by FrameworkBundle. Standalone users pass a `Psr\EventDispatcher\EventDispatcherInterface` as the fourth constructor argument of `HandleMessageMiddleware`; with none, no event is dispatched and nothing changes.

Documentation should cover: the three events and their properties, the per-handler versus per-message distinction with the worker events, that they fire for sync dispatch too, and the batch-handler timing (outcome at acknowledgment, not at enqueue).

### DIFF
--- a/src/Symfony/Component/Messenger/Event/AbstractHandlerEvent.php
+++ b/src/Symfony/Component/Messenger/Event/AbstractHandlerEvent.php
@@ -1,0 +1,24 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Messenger\Event;
+
+use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\Handler\HandlerDescriptor;
+
+abstract class AbstractHandlerEvent
+{
+    public function __construct(
+        public readonly Envelope $envelope,
+        public readonly HandlerDescriptor $handlerDescriptor,
+    ) {
+    }
+}
--- a/src/Symfony/Component/Messenger/Event/HandlerStartingEvent.php
+++ b/src/Symfony/Component/Messenger/Event/HandlerStartingEvent.php
@@ -1,0 +1,19 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Messenger\Event;
+
+/**
+ * Event dispatched before a handler is called.
+ */
+final class HandlerStartingEvent extends AbstractHandlerEvent
+{
+}
--- a/src/Symfony/Component/Messenger/Event/HandlerStartingEvent.php
+++ b/src/Symfony/Component/Messenger/Event/HandlerStartingEvent.php
@@ -1,0 +1,27 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Messenger\Event;
+
+use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\Handler\HandlerDescriptor;
+
+/**
+ * Event dispatched before a handler is called.
+ */
+final class HandlerStartingEvent
+{
+    public function __construct(
+        public readonly Envelope $envelope,
+        public readonly HandlerDescriptor $handlerDescriptor,
+    ) {
+    }
+}
--- a/src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
+++ b/src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
@@ -11,8 +11,12 @@
 
 namespace Symfony\Component\Messenger\Middleware;
 
+use Psr\EventDispatcher\EventDispatcherInterface;
 use Psr\Log\LoggerAwareTrait;
 use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\Event\HandlerFailureEvent;
+use Symfony\Component\Messenger\Event\HandlerStartingEvent;
+use Symfony\Component\Messenger\Event\HandlerSuccessEvent;
 use Symfony\Component\Messenger\Exception\HandlerFailedException;
 use Symfony\Component\Messenger\Exception\LogicException;
 use Symfony\Component\Messenger\Exception\NoHandlerForMessageException;
@@ -35,6 +39,7 @@ class HandleMessageMiddleware implements MiddlewareInterface
     public function __construct(
         private HandlersLocatorInterface $handlersLocator,
         private bool $allowNoHandlers = false,
+        private ?EventDispatcherInterface $eventDispatcher = null,
     ) {
     }
 
@@ -57,6 +62,10 @@ class HandleMessageMiddleware implements MiddlewareInterface
                 $alreadyHandled = true;
                 continue;
             }
+
+            $this->eventDispatcher?->dispatch(new HandlerStartingEvent($envelope, $handlerDescriptor));
+
+            $e = null;
 
             try {
                 $handler = $handlerDescriptor->getHandler();
@@ -96,6 +105,14 @@ class HandleMessageMiddleware implements MiddlewareInterface
                 $this->logger?->info('Message {class} handled by {handler}', $context + ['handler' => $handledStamp->getHandlerName()]);
             } catch (\Throwable $e) {
                 $exceptions[$handlerDescriptor->getName()] = $e;
+            }
+
+            if (null !== $this->eventDispatcher) {
+                $event = (null !== $e)
+                    ? new HandlerFailureEvent($envelope, $handlerDescriptor, $e)
+                    : new HandlerSuccessEvent($envelope, $handlerDescriptor);
+
+                $this->eventDispatcher->dispatch($event);
             }
         }
 
--- a/src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
+++ b/src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
@@ -12,8 +12,12 @@
 namespace Symfony\Component\Messenger\Middleware;
 
 use Psr\Clock\ClockInterface;
+use Psr\EventDispatcher\EventDispatcherInterface;
 use Psr\Log\LoggerAwareTrait;
 use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\Event\HandlerFailureEvent;
+use Symfony\Component\Messenger\Event\HandlerStartingEvent;
+use Symfony\Component\Messenger\Event\HandlerSuccessEvent;
 use Symfony\Component\Messenger\Exception\HandlerFailedException;
 use Symfony\Component\Messenger\Exception\LogicException;
 use Symfony\Component\Messenger\Exception\NoHandlerForMessageException;
@@ -37,6 +41,7 @@ class HandleMessageMiddleware implements MiddlewareInterface
         private HandlersLocatorInterface $handlersLocator,
         private bool $allowNoHandlers = false,
         private ?ClockInterface $clock = null,
+        private ?EventDispatcherInterface $dispatcher = null,
     ) {
     }
 
@@ -60,16 +65,25 @@ class HandleMessageMiddleware implements MiddlewareInterface
                 continue;
             }
 
+            $this->dispatcher?->dispatch(new HandlerStartingEvent($envelope, $handlerDescriptor));
+
+            $e = null;
+            $ack = null;
+            $ackPending = false;
+
             try {
                 $handler = $handlerDescriptor->getHandler();
                 $batchHandler = $handlerDescriptor->getBatchHandler();
 
                 if ($batchHandler && $ackStamp = $envelope->last(AckStamp::class)) {
-                    $ack = new Acknowledger(get_debug_type($batchHandler), static function (?\Throwable $e = null, $result = null) use ($envelope, $ackStamp, $handlerDescriptor) {
+                    $dispatcher = $this->dispatcher;
+                    $ack = new Acknowledger(get_debug_type($batchHandler), static function (?\Throwable $e = null, $result = null) use ($envelope, $ackStamp, $handlerDescriptor, $dispatcher) {
                         if (null !== $e) {
+                            $dispatcher?->dispatch(new HandlerFailureEvent($envelope, $handlerDescriptor, $e));
                             $e = new HandlerFailedException($envelope, [$handlerDescriptor->getName() => $e]);
                         } else {
                             $envelope = $envelope->with(HandledStamp::fromDescriptor($handlerDescriptor, $result));
+                            $dispatcher?->dispatch(new HandlerSuccessEvent($envelope, $handlerDescriptor));
                         }
 
                         $ackStamp->ack($envelope, $e);
@@ -82,6 +96,7 @@ class HandleMessageMiddleware implements MiddlewareInterface
                     }
 
                     if (!$ack->isAcknowledged()) {
+                        $ackPending = true;
                         $envelope = $envelope->with(new NoAutoAckStamp($handlerDescriptor));
                     } elseif ($ack->getError()) {
                         throw $ack->getError();
@@ -97,6 +112,16 @@ class HandleMessageMiddleware implements MiddlewareInterface
                 $this->logger?->info('Message {class} handled by {handler}', $context + ['handler' => $handledStamp->getHandlerName()]);
             } catch (\Throwable $e) {
                 $exceptions[$handlerDescriptor->getName()] = $e;
+            }
+
+            // a batch handler resolves its outcome through the acknowledger, at flush time
+            // at the earliest, and the event is then dispatched by the callback above
+            if (null !== $this->dispatcher && !$ack?->isAcknowledged()) {
+                if (null !== $e) {
+                    $this->dispatcher->dispatch(new HandlerFailureEvent($envelope, $handlerDescriptor, $e));
+                } elseif (!$ackPending) {
+                    $this->dispatcher->dispatch(new HandlerSuccessEvent($envelope, $handlerDescriptor));
+                }
             }
         }
 
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/messenger.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/messenger.php
@@ -82,6 +82,8 @@ return static function (ContainerConfigurator $container) {
             ->abstract()
             ->args([
                 abstract_arg('bus handler resolver'),
+                false,
+                service('event_dispatcher'),
             ])
             ->tag('monolog.logger', ['channel' => 'messenger'])
             ->call('setLogger', [service('logger')->ignoreOnInvalid()])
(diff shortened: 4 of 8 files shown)


Pi-Bouf (changes requested):
YES !

Pi-Bouf (approved):

OskarStark:
This must target 7.1, due to feature freeze period.

src/Symfony/Component/Messenger/Event/AbstractHandlerEvent.php
```diff
@@ -0,0 +1,36 @@
...
+abstract class AbstractHandlerEvent
+{
+    private Envelope $envelope;
+    private string $handlerName;
+
+    public function __construct(Envelope $envelope, string $handlerName)
+    {
+        $this->envelope = $envelope;
+        $this->handlerName = $handlerName;
+    }
```
OskarStark: Please use constructor property promotion

src/Symfony/Component/Messenger/Event/HandlerStartingEvent.php
```diff
@@ -0,0 +1,19 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Messenger\Event;
+
+/**
+ * Event dispatched before a handler is called.
+ */
+final class HandlerStartingEvent extends AbstractHandlerEvent
```
OskarStark: The name of the class does not make it clear, that it is dispatched before.

Maybe BeforeHandlerStartEvent?

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
```diff
@@ -35,6 +39,7 @@ class HandleMessageMiddleware implements MiddlewareInterface
     public function __construct(
         private HandlersLocatorInterface $handlersLocator,
         private bool $allowNoHandlers = false,
+        private ?EventDispatcherInterface $eventDispatcher = null,
```
OskarStark:
```suggestion
        private ?EventDispatcherInterface $dispatcher = null,
```

I like it more, but lets see what the others say

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
```diff
@@ -97,6 +106,14 @@ public function handle(Envelope $envelope, StackInterface $stack): Envelope
             } catch (\Throwable $e) {
                 $exceptions[$handlerDescriptor->getName()] = $e;
```
OskarStark, line 114:
```suggestion
                $exceptions[$handlerDescriptor->getName()] = $e;
                
                $this->eventDispatcher?->dispatch(new HandlerFailureEvent($envelope, $handlerDescriptor->getName(), $e));
```

Same for the success case and remove the `$e` variable?

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
```diff
@@ -58,6 +63,10 @@ public function handle(Envelope $envelope, StackInterface $stack): Envelope
                 continue;
             }
 
+            $this->eventDispatcher?->dispatch(new HandlerStartingEvent($envelope, $handlerDescriptor->getName()));
```
ruudk: It would be great if the `$handlerDescriptor` is fully passed to this event, and that a subscriber can mutate it.

This would solve my use case that I tried to implement in:
- https://github.com/symfony/symfony/pull/50980

It would allow event subscribers to:
- read everything from the HandlerDescriptor, so also the options 
- set a new HandlerDescriptor that has a custom callable that does specific logic.

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
ruudk, line 114: It would be great if a HandlerFailureEvent subscriber could silence an exception.

Right now, it's only possible to log them. But by always passing the exception to the HandlerFailureEvent and dispatching that, a subscriber could say "ignore this".

A use case could be to annotate `#[IgnoreException(SomeTimeoutException::class)]` on the handler. In case of an exception, a HandlerFailureEvent subscriber would see that the handler wants to ignore this exception, and it will silence it. 

From Messenger perspective all will be fine again. 

Related to
* https://github.com/symfony/symfony/pull/50980

BenMorel:
> This must target 7.1, due to feature freeze period.

Thank you, where is this documented please? I can't find any information on the [Release Process](https://symfony.com/doc/current/contributing/community/releases.html) page nor on the [Releases](https://symfony.com/releases) page. Googling only gives me blog entries for older Symfony releases.

src/Symfony/Component/Messenger/Event/AbstractHandlerEvent.php
BenMorel: Yes! A couple questions:

- as we target PHP 8.1, do we want to keep `private` + getters, or can we use `public readonly` instead? (and if targeting Symfony 7 and PHP 8.2, straight `readonly class` with public promoted properties)
- does Symfony allow a mix of promoted and non-promoted, like this:
  ```php
  final class HandlerFailureEvent extends AbstractHandlerEvent
  {
      public function __construct(
          Envelope $envelope,
          string $handlerName,
          public \Throwable $exception,
      ) {
          parent::__construct($envelope, $handlerName);
      }
  }
  ```

OskarStark:
It should be mentioned [here](https://symfony.com/doc/current/contributing/code/pull_requests.html#choose-the-right-branch) IMHO, which is not the case.
The feature freeze period is just a small timeframe, IIRC it started around mid October and will end end November when Symfony 6.4 and 7 gets released.

src/Symfony/Component/Messenger/Event/HandlerStartingEvent.php
BenMorel: I actually thought about this, but wanted to keep things organized by prefix, like `Worker*` events. I don't mind though, what about just `BeforeHandlerEvent`?

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
BenMorel: I don't mind dispatching the failure event in the `catch` block, however I'm a bit reluctant about dispatching the success event in the `try` block, as a failure in any listener/subscriber of this event would mark the *handler* as failed.

 That's why I kept things outside the `try/catch`.

OskarStark:
I added some info
* https://github.com/symfony/symfony-docs/pull/19110

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
BenMorel: I'm all for adding the whole `HandlerDescriptor` in the events, you're right that it opens up a whole lot of possibilities.

However, I have no idea of the potential consequences of live-replacing a handler, so I'll leave it to others to decide if making the `HandlerDescriptor` replaceable is the way to go!

BenMorel: I'm all for adding the whole `HandlerDescriptor` in the events, you're right that it opens up a whole lot of possibilities.

However, I have no idea of the potential consequences of live-replacing a handler, so I'll leave it to others to decide if making the `HandlerDescriptor` replaceable is the way to go!

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
BenMorel: We could add a `HandlerFailureEvent::ignoreException(bool $ignore)` method indeed. I don't mind adding this feature, and again I'll leave it to others to decide if adding it is a good idea.

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
ruudk: @ro0NL @OskarStark What do you think?

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
ruudk: @ro0NL @OskarStark What do you think?

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
OskarStark: I have to admit that I also don't know what consequences this can have

BenMorel: @ruudk Do you actually have a need to replace the `HandlerDescriptor` at the moment, or can we leave it for a later PR to maybe make it replaceable in the future?

ruudk: Having it would be great, but maybe I can use the event system already to do what I want. I agree that we can later always decide to open this up.

Could we at least add the full HandlerDescriptor instead of only the name? That would make it a bit better.

BenMorel: > Could we at least add the full HandlerDescriptor instead of only the name? That would make it a bit better.

Yes, definitely. I'll change this.

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
BenMorel: Actually `SendMessageMiddleware` [uses `$eventDispatcher`](https://github.com/symfony/symfony/blob/7.0/src/Symfony/Component/Messenger/Middleware/SendMessageMiddleware.php#L33), so I guess it's better to keep it consistent here?

src/Symfony/Component/Messenger/Event/AbstractHandlerEvent.php
BenMorel: I've settled on `public readonly` as I could find other occurrences in the `7.0` branch. I also used a mix of promoted and non-promoted in `HandlerFailureEvent`, as in the code sample above.

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
BenMorel: @ruudk events now have the full `HandlerDescriptor`!

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
ro0NL: [eventDispatcher](https://github.com/search?q=repo%3Asymfony%2Fsymfony+%22EventDispatcherInterface+%24eventDispatcher%22+language%3APHP&type=code) vs [dispatcher](https://github.com/search?q=repo%3Asymfony%2Fsymfony+%22EventDispatcherInterface+%24dispatcher%22+language%3APHP&type=code&l=PHP)

but we dont mind :D

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
ro0NL: im not really convinced about `#[IgnoreException(`, i'd prefer try/catching it at the handler level which is 100% explicit

src/Symfony/Component/Messenger/Event/AbstractHandlerEvent.php
```diff
@@ -0,0 +1,24 @@
...
+namespace Symfony\Component\Messenger\Event;
+
+use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\Handler\HandlerDescriptor;
+
+abstract class AbstractHandlerEvent
+{
+    public function __construct(
+        public readonly Envelope $envelope,
+        public readonly HandlerDescriptor $handlerDescriptor,
```
ro0NL, line 21: i'd inline these where needed, and avoid inheritance (or make it internal)

src/Symfony/Component/Messenger/Event/HandlerStartingEvent.php
ro0NL: > The name of the class does not make it clear, that it is dispatched before.

i dont see how it could mean something else

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
ruudk: Also fine, but then we should let a subscriber mutate the handler descriptor. See my previous comment.

src/Symfony/Component/Messenger/Event/AbstractHandlerEvent.php
BenMorel: Do you mean no abstract? What about `AbstractWorkerMessageEvent`?

BenMorel:
I've just rebased this PR against the new 7.1 branch. Is there something I can do to get it merged?

As it stands, there are a few points in the comments that would need to be ruled on. Ideally I'd like to keep this PR simple for now, but keep things extendable if someone else wants to build extra features on top of it.

Jeroeny:
What if the callHandler method was moved to an interface (HandlerExecutor) ? Then there can also be a Traceable version of it to measure and debug handling per handler.

ruudk:
> What if the callHandler method was moved to an interface (HandlerExecutor) ? Then there can also be a Traceable version of it to measure and debug handling per handler.

I like this a lot 🙌  

It makes a lot of sense to be able to probe execution of the handler(s). This makes #50980 and #50998 absolute. As with this interface, I can now create my own HandlerExecutor middleware that to do transaction wrapping or exception ignoring.

jerowork:
> What if the callHandler method was moved to an interface (HandlerExecutor) ? Then there can also be a Traceable version of it to measure and debug handling per handler.

Exactly what I was thinking (its in the name I guess 😉 ). 
Making `callHandler` protected would solve it, but extracting the calling handler to a separate class makes it more flexible and open for extension for everybody who wants to change the calling handler for whatever reason (for me I would like to have a upgraded way of injecting other (optional) parameters in a handler)

nicolas-grekas:
Thanks for your patience @BenMorel, and for the clear write-up of the auditing gymnastics this removes. The feature is wanted, I rebased on 8.2 and pushed with one behavioral fix.

**Batch handlers got a wrong success event.** As submitted, the outcome event was dispatched right after `callHandler()`. For a batch handler that call only *enqueues* the message, so `HandlerSuccessEvent` fired before the handler had run, and the real outcome at flush time produced no event at all. For an auditing feature, that is the case that cannot be wrong. The outcome events now come from the acknowledger callback, which is where the result or error actually lands, whether the handler acknowledges synchronously or at a later flush. `HandlerStartingEvent` stays at invoke time, which is honest for batch handlers too since `__invoke()` does run to enqueue. Every handler run gets exactly one outcome event; three tests cover queued-then-flushed success, a nacked batch, and the no-double-dispatch guarantee.

**The test needed rebuilding anyway**: it used `withConsecutive()`, which no longer exists in the PHPUnit in use, so it errored rather than ran. It now records events with a small PSR-14 dispatcher fixture and asserts the sequence.

On @Jeroeny's `HandlerExecutor` suggestion, which several people liked: a traceable executor wrapping `callHandler()` would hit exactly the bug fixed above, because `callHandler()` on a batch handler returns at enqueue, and the execution happens in `flush()` and resolves through the `Acknowledger`, neither of which passes through `callHandler()`. So an executor interface is not a substitute for these events, and per-handler timing for the profiler can still be built on top of them later.

Also rebased around the new `ClockInterface` constructor argument (the dispatcher is the fourth parameter now), moved the changelog entry to 8.2, and dropped the empty `HandlerStartingEvent`/`HandlerSuccessEvent` bodies into place unchanged. Messenger is green apart from the pre-existing `WorkerTest` parallel failures my local checkout has without `amphp/parallel`; the middleware test file runs 19 green, and reverting the middleware while keeping the tests fails the three event tests.

nicolas-grekas:
Dropped `AbstractHandlerEvent`: Symfony's dispatcher matches listeners by concrete class, so an abstract base is not a subscription point, and sharing two promoted properties did not justify a fourth API class. The three events are now self-contained finals. 7 files, +264/-1, still one commit.

src/Symfony/Component/Messenger/Event/AbstractHandlerEvent.php
nicolas-grekas: Done on the current head: the three events use promoted public readonly properties.

src/Symfony/Component/Messenger/Event/HandlerStartingEvent.php
nicolas-grekas: Keeping HandlerStartingEvent: the Handler* prefix keeps the family grouped, the same way the Worker* events are.

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
nicolas-grekas: The current head keeps the outcome dispatch outside the try/catch, for the reason Benjamin gave: a throwing listener must not mark the handler as failed. For batch handlers the outcome is dispatched by the acknowledger callback once it is known.

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
nicolas-grekas: The events carry the full HandlerDescriptor on the current head. Making it replaceable is left out on purpose: the events are read-only notifications, and mutability can be added later without BC concerns.

src/Symfony/Component/Messenger/Middleware/HandleMessageMiddleware.php
nicolas-grekas: Left out on purpose: the events are read-only notifications for now. An opt-in to silence the exception can be added later without breaking BC, once there is a concrete case for it.

src/Symfony/Component/Messenger/Event/AbstractHandlerEvent.php
nicolas-grekas: Settled the other way on the current head: the abstract base is gone. The dispatcher matches listeners by concrete class, so a base class is not something listeners can subscribe to, and sharing two properties did not justify a fourth API class.

nicolas-grekas (approved):

nicolas-grekas:
Pushed one more fix on top of my rebase, to the batch guard I introduced there.

`$outcomeDispatched` was captured by reference in each acknowledger closure, and PHP closures share the enclosing function's variable slot, not the loop iteration's. So when a batch handler's acknowledger resolved during a later handler's run in the same loop, which is exactly what a flush does, the flag was already true and that later handler's own outcome event was suppressed.

Replaced the flag with `$ack?->isAcknowledged()` on the per-iteration acknowledger, which cannot cross iterations. Regression test added: a batch handler plus a second handler that forces a flush; it fails with 3 events instead of 4 before the fix. Messenger middleware suite green, CS clean.

nicolas-grekas:
Thank you @BenMorel.

nicolas-grekas:
One test to add on top of this, now sent as #65450: a message handled by two handlers where the first is a batch handler that leaves its acknowledger pending. That is the one interaction the suite does not cover, and it is where an outcome event can silently go missing: the pending batch must not suppress the second handler's `HandlerSuccessEvent`, and both batch outcomes must still arrive when the batch is flushed.

The current code passes it as is, so it only locks the behavior in. I checked it fails if the pending state leaks across handlers.

BenMorel:
Thank you for bringing this to life, @nicolas-grekas! :yellow_heart:

PR #52426, opened 2023-11-02
[HttpKernel] the debug log processor must be a callable

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

A Monolog processor must be a callable, but a `DebugLoggerInterface` implementation is not necessarily a callable.

### DIFF
--- a/src/Symfony/Component/HttpKernel/Log/DebugLoggerConfigurator.php
+++ b/src/Symfony/Component/HttpKernel/Log/DebugLoggerConfigurator.php
@@ -18,12 +18,12 @@ use Monolog\Logger;
  */
 class DebugLoggerConfigurator
 {
-    private ?DebugLoggerInterface $processor = null;
+    private ?\Closure $processor = null;
 
-    public function __construct(DebugLoggerInterface $processor, bool $enable = null)
+    public function __construct(callable $processor, bool $enable = null)
     {
         if ($enable ?? !\in_array(\PHP_SAPI, ['cli', 'phpdbg', 'embed'], true)) {
-            $this->processor = $processor;
+            $this->processor = $processor(...);
         }
     }
 


src/Symfony/Component/HttpKernel/Log/DebugLoggerConfigurator.php
```diff
@@ -18,20 +18,19 @@
...
         }
     }
 
-    public function pushDebugLogger(Logger $logger): void
+    public function registerDebugLogger(Logger $logger): void
     {
-        if ($this->processor) {
-            $logger->pushProcessor($this->processor);
-        }
+        self::$registeredLoggers[spl_object_id($logger)] = $this->debugLogger;
```
nicolas-grekas: should we use a WeakMap?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/AddDebugLogProcessorPass.php
```diff
@@ -33,7 +33,7 @@ public function process(ContainerBuilder $container)
         }
 
         $container->getDefinition('monolog.logger_prototype')
-            ->setConfigurator([new Reference('debug.debug_logger_configurator'), 'pushDebugLogger']);
+            ->setConfigurator([new Reference('debug.debug_logger_configurator'), 'registerForDebugLogger']);
```
stof: This is wrong here. You are registering the Monolog\Logger itself for debug logger, not the debug logger implementation, breaking the intent of this compiler pass (which is about registering the DebugProcessor of the bridge into the logger instances).

src/Symfony/Component/HttpKernel/Log/DebugLoggerConfigurator.php
```diff
@@ -18,20 +18,21 @@
...
-            $this->processor = $processor;
+            $this->debugLogger = $debugLogger;
         }
     }
 
-    public function pushDebugLogger(Logger $logger): void
+    public function registerForDebugLogger(Logger $logger): void
     {
-        if ($this->processor) {
-            $logger->pushProcessor($this->processor);
```
stof: If you are not pushing the processor in the monolog logger, it won't be receiving logs anymore.

stof (changes requested):

xabbuh:
re-reading the existing code I have updated this PR to only enforce the processor to be a callable and reverted the other changes

stof (approved):

src/Symfony/Component/HttpKernel/Log/DebugLoggerConfigurator.php
```diff
@@ -18,12 +18,12 @@
  */
 class DebugLoggerConfigurator
 {
-    private ?DebugLoggerInterface $processor = null;
+    private ?\Closure $processor = null;
 
-    public function __construct(DebugLoggerInterface $processor, bool $enable = null)
+    public function __construct(callable $processor, bool $enable = null)
```
nicolas-grekas, line 23: should be `ProcessorInterface|callable` for monolog 3, isn't it?

xabbuh: We could change it that way, but it's not needed IMO as the `ProcessorInterface` does nothing more than enforcing the `__invoke()` method to be present: https://github.com/Seldaek/monolog/blob/main/src/Monolog/Processor/ProcessorInterface.php#L26

nicolas-grekas: oh perfect then

nicolas-grekas:
Thank you @xabbuh.

PR #52428, opened 2023-11-02
[HttpKernel] Preventing error 500 when function putenv is disabled

ShaiMagal:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

**Description:**
Some webhostings are blocking function "**putenv**". So, it will show error 500 Error (UndefinedFunction) with text:
Attempted to call function "putenv" from namespace "Symfony\Component\HttpKernel

This PR is checking callability of putenv first, so preventing error 500.

**How to test?**
  -> php.ini - disabled_functions=putenv
  -> enable debug
  - >Then you will see error 500
  -> After this PR/commit, it's without problem.

For example it's very common problem for PrestaShop users. (PrestaShop is using Symfony)

### DIFF
--- a/src/Symfony/Component/HttpKernel/Kernel.php
+++ b/src/Symfony/Component/HttpKernel/Kernel.php
@@ -778,7 +778,9 @@ abstract class Kernel implements KernelInterface, RebootableInterface, Terminabl
             $this->startTime = microtime(true);
         }
         if ($this->debug && !isset($_ENV['SHELL_VERBOSITY']) && !isset($_SERVER['SHELL_VERBOSITY'])) {
-            putenv('SHELL_VERBOSITY=3');
+            if (\function_exists('putenv')) {
+                putenv('SHELL_VERBOSITY=3');
+            }
             $_ENV['SHELL_VERBOSITY'] = 3;
             $_SERVER['SHELL_VERBOSITY'] = 3;
         }


src/Symfony/Component/HttpKernel/Kernel.php
```diff
@@ -778,7 +778,9 @@ private function preBoot(): ContainerInterface
             $this->startTime = microtime(true);
         }
         if ($this->debug && !isset($_ENV['SHELL_VERBOSITY']) && !isset($_SERVER['SHELL_VERBOSITY'])) {
-            putenv('SHELL_VERBOSITY=3');
+            if (is_callable('putenv')) {
```
stof: You should use `\function_exists('putenv')` instead, as this will allow OPCache to optimise this to a `if (true)` (and then removing the `if` entirely) for the common case where this function is not disabled (this optimisation of OPCache works for functions defined in PHP core or its extensions, but this is where the function will be defined in this case for most users)

stof (changes requested):

ShaiMagal:
@stof Updated :)

src/Symfony/Component/HttpKernel/Kernel.php
```diff
@@ -778,7 +778,9 @@ private function preBoot(): ContainerInterface
             $this->startTime = microtime(true);
         }
         if ($this->debug && !isset($_ENV['SHELL_VERBOSITY']) && !isset($_SERVER['SHELL_VERBOSITY'])) {
-            putenv('SHELL_VERBOSITY=3');
+            if (function_exists('putenv')) {
```
stof: For the optimization to apply, this must be a qualified call, not a call going through the global fallback. So you need the `\` that was in my previous review (btw, fabbot also asks to add it)

nicolas-grekas (approved):

OskarStark (approved):

ShaiMagal:
Guys, will this be merged with branch 6.x as well in future or I should make another PR specially for 6.x? (this PR I targeted to 5.4 but I don't know if this will be at 6.x in future... so I am asking, how to do this correctly.

Thank you.

stof:
@ShaiMagal this will be merged up by the core team once this PR is merged (the merge up happens usually once a week for dev version, and it will happen for sure before tagging a release)

fabpot:
Should we do the same in `src/Symfony/Component/BrowserKit/AbstractBrowser.php` and `src/Symfony/Component/Runtime/SymfonyRuntime.php`?

nicolas-grekas:
I had a look and I don't think we need to: AbstractBrowser + the spot in Runtime are for command line usages, and to me those are not affected by (cheap) hosters.

xabbuh (approved):

derrabus (approved):

fabpot (approved):

fabpot:
Thank you @ShaiMagal.

PR #52429, opened 2023-11-02
[HttpClient] Replace `escapeshellarg` to prevent overpassing `ARG_MAX`

### LINKED ISSUE #49693: [HttpClient] Hardcoded ARG_MAX in data collector

6.2.6

### Description

The HTTP Client Data Collector still throws fatal errors if the payload for `escapeshellarg` is too large. This has already been tried to be fixed with the following pull request: https://github.com/symfony/symfony/pull/46700

In our Docker environment, the error still happens. I ran the following command to see what our limit is:

```bash
getconf ARG_MAX
131072
```

We use `alpine` docker image and there are different limits. You can find this out quite easily:

```bash
# alpine
docker run -it alpine getconf ARG_MAX 
131072

# ubuntu
docker run -it ubuntu getconf ARG_MAX
2097152

# debian
docker run -it debian getconf ARG_MAX
2097152

# macos
getconf ARG_MAX
1048576
```

I did a little research, I think in php it's not that easy to find out the limit without executing a shell command.
And also `getconf ARG_MAX` is unfortunately not the maximum that is available in the user space (php?). There is an article about it here: https://www.in-ulm.de/~mascheck/various/argmax/

By the way, just because it was possible to create the "copy-curl" string doesn't mean that you can actually execute it somewhere, since it depends on ARG_MAX there too. :sweat_smile:

### How to reproduce

Start symfony in an `alpine` docker container and create a big http client payload (between 131072 and 256000 bytes) with enabled profiler.

### Possible Solution

* Keep lowering the limit and we hope it's enough.
* Try to get the correct ARG_MAX value at runtime from the underlying system.

### Additional Context

_No response_

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #49693
| License       | MIT

I'm not 100% sure if it is a bugfix or a feature.

I used Nicolas' suggestion in the issue to sanitize the input and used `--data-raw` to avoid any automatic formatting.

Removing the use of `escapeshellarg()` also allows to remove `HttpClientDataCollectorTest::testItDoesNotGeneratesCurlCommandsForNotEncodableBody()`. Indeed, the body can now be encoded and will result on the following cURL command:

```
curl \\n
      --compressed \\n
      --request POST \\n
      --url 'http://localhost:8057/json' \\n
      --header 'Accept: */*' \\n
      --header 'Content-Length: 1' \\n
      --header 'Content-Type: application/x-www-form-urlencoded' \\n
      --header 'Accept-Encoding: gzip' \\n
      --header 'User-Agent: Symfony HttpClient (Native)' \\n
      --data-raw '\x00'

```

### DIFF
--- a/src/Symfony/Component/HttpClient/DataCollector/HttpClientDataCollector.php
+++ b/src/Symfony/Component/HttpClient/DataCollector/HttpClientDataCollector.php
@@ -17,6 +17,7 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\DataCollector\DataCollector;
 use Symfony\Component\HttpKernel\DataCollector\LateDataCollectorInterface;
+use Symfony\Component\Process\Process;
 use Symfony\Component\VarDumper\Caster\ImgStub;
 
 /**
@@ -193,27 +194,14 @@ final class HttpClientDataCollector extends DataCollector implements LateDataCol
         $dataArg = [];
 
         if ($json = $trace['options']['json'] ?? null) {
-            if (!$this->argMaxLengthIsSafe($payload = self::jsonEncode($json))) {
-                return null;
-            }
-            $dataArg[] = '--data '.escapeshellarg($payload);
+            $dataArg[] = '--data-raw '.$this->escapePayload(self::jsonEncode($json));
         } elseif ($body = $trace['options']['body'] ?? null) {
             if (\is_string($body)) {
-                if (!$this->argMaxLengthIsSafe($body)) {
-                    return null;
-                }
-                try {
-                    $dataArg[] = '--data '.escapeshellarg($body);
-                } catch (\ValueError) {
-                    return null;
-                }
+                $dataArg[] = '--data-raw '.$this->escapePayload($body);
             } elseif (\is_array($body)) {
                 $body = explode('&', self::normalizeBody($body));
                 foreach ($body as $value) {
-                    if (!$this->argMaxLengthIsSafe($payload = urldecode($value))) {
-                        return null;
-                    }
-                    $dataArg[] = '--data '.escapeshellarg($payload);
+                    $dataArg[] = '--data-raw '.$this->escapePayload(urldecode($value));
                 }
             } else {
                 return null;
@@ -250,13 +238,18 @@ final class HttpClientDataCollector extends DataCollector implements LateDataCol
         return implode(" \\\n  ", $command);
     }
 
-    /**
-     * Let's be defensive : we authorize only size of 8kio on Windows for escapeshellarg() argument to avoid a fatal error.
-     *
-     * @see https://github.com/php/php-src/blob/9458f5f2c8a8e3d6c65cc181747a5a75654b7c6e/ext/standard/exec.c#L397
-     */
-    private function argMaxLengthIsSafe(string $payload): bool
+    private function escapePayload(string $payload): string
     {
-        return \strlen($payload) < ('\\' === \DIRECTORY_SEPARATOR ? 8100 : 256000);
+        static $useProcess;
+
+        if ($useProcess ??= class_exists(Process::class)) {
+            return (new Process([$payload]))->getCommandLine();
+        }
+
+        if ('\\' === \DIRECTORY_SEPARATOR) {
+            return '"'.str_replace('"', '""', $payload).'"';
+        }
+
+        return "'".str_replace("'", "'\\''", $payload)."'";
     }
 }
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/HttpClient/DataCollector/HttpClientDataCollector.php
```diff
@@ -250,13 +237,8 @@ private function getCurlCommand(array $trace): ?string
...
-    /**
-     * Let's be defensive : we authorize only size of 8kio on Windows for escapeshellarg() argument to avoid a fatal error.
-     *
-     * @see https://github.com/php/php-src/blob/9458f5f2c8a8e3d6c65cc181747a5a75654b7c6e/ext/standard/exec.c#L397
-     */
-    private function argMaxLengthIsSafe(string $payload): bool
+    private function sanitizePayload(string $payload): string
     {
-        return \strlen($payload) < ('\\' === \DIRECTORY_SEPARATOR ? 8100 : 256000);
+        return "'".str_replace("'", "'\\''", $payload)."'";
```
stof, line 253: This is broken on Windows, where single quotes are not valid shell quoting.

stof (changes requested):

src/Symfony/Component/HttpClient/DataCollector/HttpClientDataCollector.php
alexandre-daubois: Nicolas suggested to do this, even for Windows: https://github.com/symfony/symfony/issues/49693#issuecomment-1754694141. However, I can update this bit and find another way to do it if necessary

Edit: Having a look to update it for Windows

alexandre-daubois: I updated the method to work on Windows using double quotes when necessary

alexandre-daubois:
@nicolas-grekas I really like your idea of leveraging Process. I updated the code to use it first if it exists.

src/Symfony/Component/HttpClient/DataCollector/HttpClientDataCollector.php
```diff
@@ -250,13 +238,16 @@ private function getCurlCommand(array $trace): ?string
...
-    /**
-     * Let's be defensive : we authorize only size of 8kio on Windows for escapeshellarg() argument to avoid a fatal error.
-     *
-     * @see https://github.com/php/php-src/blob/9458f5f2c8a8e3d6c65cc181747a5a75654b7c6e/ext/standard/exec.c#L397
-     */
-    private function argMaxLengthIsSafe(string $payload): bool
+    private function sanitizeArg(string $payload): string
     {
-        return \strlen($payload) < ('\\' === \DIRECTORY_SEPARATOR ? 8100 : 256000);
+        if (class_exists(Process::class)) {
```
nicolas-grekas:
```suggestion
        static $useProcess;

        if ($useProcess ??= class_exists(Process::class)) {
```

src/Symfony/Component/HttpClient/DataCollector/HttpClientDataCollector.php
```diff
@@ -193,27 +194,14 @@ private function getCurlCommand(array $trace): ?string
         $dataArg = [];
 
         if ($json = $trace['options']['json'] ?? null) {
-            if (!$this->argMaxLengthIsSafe($payload = self::jsonEncode($json))) {
-                return null;
-            }
-            $dataArg[] = '--data '.escapeshellarg($payload);
+            $dataArg[] = '--data-raw '.$this->sanitizeArg(self::jsonEncode($json));
```
nicolas-grekas: we don't sanitize, we escape

src/Symfony/Component/HttpClient/DataCollector/HttpClientDataCollector.php
```diff
@@ -250,13 +238,16 @@ private function getCurlCommand(array $trace): ?string
...
-     */
-    private function argMaxLengthIsSafe(string $payload): bool
+    private function sanitizeArg(string $payload): string
     {
-        return \strlen($payload) < ('\\' === \DIRECTORY_SEPARATOR ? 8100 : 256000);
+        if (class_exists(Process::class)) {
+            return (new Process([$payload]))->getCommandLine();
+        }
+
+        if (\DIRECTORY_SEPARATOR === '\\') {
```
nicolas-grekas:
```suggestion
        if ('\\' === \DIRECTORY_SEPARATOR) {
```

nicolas-grekas (approved):
LGTM after minor changes

src/Symfony/Component/HttpClient/DataCollector/HttpClientDataCollector.php
alexandre-daubois: I addressed your comments, thanks :slightly_smiling_face:

fabpot (approved):

fabpot:
Thank you @alexandre-daubois.

PR #52434, opened 2023-11-03
[Console][FrameworkBundle] Fix missing `profile` option for console commands

### LINKED ISSUE #52433: [Console][FrameworkBundle] `--profile` option might not be set on every command

6.4

### Description

With the new command profiling feature, a event listener was added that grabs the `profile` option from the command inputs:
https://github.com/symfony/symfony/blob/6.4/src/Symfony/Bundle/FrameworkBundle/EventListener/ConsoleProfilerListener.php#L62

Normaly that option is added to every command over the Application: https://github.com/symfony/symfony/blob/6.4/src/Symfony/Bundle/FrameworkBundle/Console/Application.php#L47

However not every command is executed over the Application class, e.g. during unit tests with the `CommandTester` or by just dispatching the `ConsoleCommandEvent` directly.

In those cases the following error is thrown:
```
Symfony\Component\Console\Exception\InvalidArgumentException: The "profile" option does not exist.

/.../symfony/console/Input/Input.php:127
/.../symfony/framework-bundle/EventListener/ConsoleProfilerListener.php:62
/.../symfony/event-dispatcher/Debug/WrappedListener.php:116
/.../symfony/event-dispatcher/EventDispatcher.php:206
/.../symfony/event-dispatcher/EventDispatcher.php:56
```

### How to reproduce

Use the command tester to test a manually created command:

```php
$command = new MyCommandToTest();
$commandTester = new CommandTester($command);

$commandTester->run($input);
```

Note: this differs from how the command tester is documented: https://symfony.com/doc/current/console.html#testing-commands
As there the command is fetched from the application, however, to properly unit test commands without having the whole application booted up, I think the above code is valid for testing commands and should be supported.

### Possible Solution

Check if the option exists before accessing it -> PR incoming

### Additional Context

_No response_

keulinho:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no 
| Issues        | Fix https://github.com/symfony/symfony/issues/52433
| License       | MIT

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/EventListener/ConsoleProfilerListener.php
+++ b/src/Symfony/Bundle/FrameworkBundle/EventListener/ConsoleProfilerListener.php
@@ -59,7 +59,8 @@ final class ConsoleProfilerListener implements EventSubscriberInterface
 
     public function initialize(ConsoleCommandEvent $event): void
     {
-        if (!$event->getInput()->getOption('profile')) {
+        $input = $event->getInput();
+        if (!$input->hasOption('profile') || !$input->getOption('profile')) {
             $this->profiler->disable();
 
             return;


keulinho:
Unit test failure seems unrelated to this change :thinking:

keulinho:
Also the fail in the appveyor CI seems unrelated as it is some error in the redis integration test for the messenger component :thinking:

OskarStark:
friendly ping @HeahDude as you contributed this feature, thanks

GromNaN:
I could not reproduce the issue https://github.com/symfony/symfony/issues/52433.

This listener is part of the `FrameworkBundle`, and so it should be used with `Symfony\Bundle\FrameworkBundle\Console
\Application`. Could you provide more details on your application or test setup?

keulinho:
Ok took a look again and seemed i messed something up with the command tester, that seems to work as it does not dispatch the event. 

However manually triggering events by dispatching the `ConsoleCommandEvent` does not work anymore.
E.g.
```php
$commandEvent = new ConsoleCommandEvent($command, $input, $output);
$kernel->getContainer()->get('event_dispatcher')->dispatch($commandEvent, ConsoleEvents::COMMAND);
```

We used something like that in an internal command tester, that's why i first assumed the issue is in the symfony command tester as well.

However, I still think the fix makes sense, as you should not rely in the listener that the event was correctly configured in the application class. IMHO the benefit of using events is to decouple different parts of the system from each other, but this assumption (that there is always an `profile` option) in the listener basically couples the listener to the application class-

chalasr:
I've no strong opinion here. I think I can live with that coupling, especially given it's part of a bundle

HeahDude (approved):
Given https://github.com/symfony/symfony/issues/52433#issuecomment-1794803064, I think this fix looks sensible. Thanks!

chalasr (approved):

chalasr:
Thank you @keulinho.

PR #52438, opened 2023-11-03
[HttpKernel] Fix uninitialized property in Bundle class

### LINKED ISSUE #52436: [HttpKernel] Uninitialized property in the abstract Bundle class

7.0

### Description

In a Symfony bundle (https://github.com/symfonycorp/connect) we have this:

```php
namespace SymfonyCorp\Connect;

use Symfony\Component\HttpKernel\Bundle\Bundle;

class SymfonyConnectBundle extends Bundle
{
    public function getContainerExtension(): ?ExtensionInterface
    {
        if (null === $this->extension) {
            $this->extension = new SymfonyConnectExtension();
        }

        return $this->extension;
    }

    // ...
}
```

It works fine in all Symfony versions up to and including 6.4.

But, in 7.0 (beta2) we're seeing this error:

```
Typed property Symfony\Component\HttpKernel\Bundle\Bundle::$extension
must not be accessed before initialization.
```

The error happens because this line in 6.4:

https://github.com/symfony/symfony/blob/8a69f6766815bd457ce12ab6a04293207efcec52/src/Symfony/Component/HttpKernel/Bundle/Bundle.php#L28

Turned into this line in 7.0:

https://github.com/symfony/symfony/blob/fa590af43d297995b24511d04f805550ffcadc4a/src/Symfony/Component/HttpKernel/Bundle/Bundle.php#L28

If I make this change in Symfony code, everything works again:

```diff
-protected ExtensionInterface|false $extension;
+protected ExtensionInterface|false|null $extension = null;
```

So, what should we do here? Thanks!

### How to reproduce

Use this bundle https://github.com/symfonycorp/connect in a Symfony 7 application.

### Possible Solution

_No response_

### Additional Context

_No response_

javiereguiluz:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52436
| License       | MIT

### DIFF
--- a/src/Symfony/Component/HttpKernel/Bundle/Bundle.php
+++ b/src/Symfony/Component/HttpKernel/Bundle/Bundle.php
@@ -25,7 +25,7 @@ use Symfony\Component\DependencyInjection\Extension\ExtensionInterface;
 abstract class Bundle implements BundleInterface
 {
     protected string $name;
-    protected ExtensionInterface|false $extension;
+    protected ExtensionInterface|false|null $extension = null;
     protected string $path;
     protected ?ContainerInterface $container;
 


OskarStark (approved):

nicolas-grekas:
Thank you @javiereguiluz.

PR #52442, opened 2023-11-03
Disable the "Copy as cURL" button when the debug info are disabled

### LINKED ISSUE #52441: [HttpClient] Issues with debug panel

6.3.latest

### Description

I used the profiler, and noticed many things that seems weird / wrong

### The copy as curl button

image

When I click on the button I got:

```bash
curl \
  --compressed \
  --header 'Due to a bug in curl 7.64.0, the debug log is disabled; use another version to work around the issue'```
```

1. the URL is missing
2. the method is missing - (occurs even if method != GET)
3. the headers are missing
4. the current header is wrong

### Duration

image

I'm not sure which total I must read - It could be nice to display it in the profiler directly, like with doctrine query

stof:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52441
| License       | MIT

Some versions of curl have a bug that prevents collecting the debug info. This is reported by writing an explanation message in the debug buffer.
When this is detecting, the "Copy as cURL" button is now skipped (like for other unsupported cases) instead of copying a broken command.

### DIFF
--- a/src/Symfony/Component/HttpClient/DataCollector/HttpClientDataCollector.php
+++ b/src/Symfony/Component/HttpClient/DataCollector/HttpClientDataCollector.php
@@ -230,6 +230,11 @@ final class HttpClientDataCollector extends DataCollector implements LateDataCol
                 break;
             }
 
+            if (str_starts_with('Due to a bug in curl ', $line)) {
+                // When the curl client disables debug info due to a curl bug, we cannot build the command.
+                return null;
+            }
+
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):

fabpot (approved):

fabpot:
Thank you @stof.

PR #52443, opened 2023-11-03
[Yaml] Fix uid binary parsing

mRoca:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

Currently, on SF 5.4 & 6.4, a YAML file containing `{ uid: !!binary Ju0Yh+uqSXOagJZFTlUt8g== }` cannot be parsed because of a `preg_match` error: `Malformed UTF-8 characters, possibly incorrectly encoded`.

This is because `preg_match` should not be run over a binary string, but this specific binary string matches the following tests : `\is_string($value) && '' !== $value && '&' === $value[0]` (the string starts with `0x26` , which is the UTF-8 `&` char).

### DIFF
--- a/src/Symfony/Component/Yaml/Inline.php
+++ b/src/Symfony/Component/Yaml/Inline.php
@@ -531,7 +531,7 @@ class Inline
                         if ('<<' === $key) {
                             $output += $value;
                         } elseif ($allowOverwrite || !isset($output[$key])) {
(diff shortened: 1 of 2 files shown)


xabbuh (approved):
LGTM

derrabus (approved):

fabpot (approved):

fabpot:
Thank you @mRoca.

PR #52449, opened 2023-11-04
[TwigBridge] Mark CodeExtension as @internal

fabpot:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Issues        | n/a
| License       | MIT

This is really just internal code used by our own tools. Let's make it clear that nobody should use it in production.

### DIFF
--- a/src/Symfony/Bridge/Twig/Extension/CodeExtension.php
+++ b/src/Symfony/Bridge/Twig/Extension/CodeExtension.php
@@ -18,7 +18,12 @@ use Twig\TwigFilter;
 /**
  * Twig extension relate to PHP code and used by the profiler and the default exception templates.
  *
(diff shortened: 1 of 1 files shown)


nicolas-grekas (approved):

GromNaN:
Does marking this class `@internal` mean that we can no longer use the filters it provides for external packages?

For instance `file_link` is used in DoctrineBundle to render the backtrace in the profiler.
https://github.com/doctrine/DoctrineBundle/blob/42f5834825eaa967af833722c336d5dd05cbd4f3/Resources/views/Collector/db.html.twig#L283C1-L283C129

PR #52454, opened 2023-11-04
[Translation] add bundles trans paths into commands (extract/pull/push)

Jibbarth:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | N/A
| License       | MIT

Hello ! 

While playing with `translation:extract` command to trying to gather all sylius translation in one file in order to send them to a translation provider, I found that although translation file are correctly loaded from a bundle while browsing the application, it was mandatory to launch the `translation:extract` for EACH bundle, and in a Sylius App, there is a lot :sweat_smile: 

This fix pass all $transPaths discovered during compilation into that command (and also push/pull commands). 

In my mind, it's a bugfix, that's why I targetted 5.4, but could be considered as new feature, I'm not sure.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
@@ -1072,7 +1072,19 @@ abstract class FrameworkExtensionTestCase extends TestCase
 
     public function testTranslator()
     {
-        $container = $this->createContainerFromFile('full');
+        require_once __DIR__.'/Fixtures/TestBundle/TestBundle.php';
+
+        $container = $this->createContainerFromFile('full', [
+            'kernel.bundles' => [
+                'FrameworkBundle' => 'Symfony\\Bundle\\FrameworkBundle\\FrameworkBundle',
+                'TestBundle' => 'Symfony\\Bundle\\FrameworkBundle\\Tests\\TestBundle',
+            ],
+            'kernel.bundles_metadata' => [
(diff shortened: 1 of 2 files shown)


src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
```diff
@@ -1115,6 +1127,11 @@ public function testTranslator()
             '->registerTranslatorConfiguration() finds translation resources with dots in domain'
         );
         $this->assertContains(strtr(__DIR__.'/translations/security.en.yaml', '/', \DIRECTORY_SEPARATOR), $files);
+        $this->assertContains(
+            strtr(__DIR__.'/Fixtures/TestBundle/Resources/translations/messages.en.yaml', '/', \DIRECTORY_SEPARATOR),
+            $files,
+            '->registerTranslatorConfiguration() finds translation resources with in bundles'
```
OskarStark:
```suggestion
            '->registerTranslatorConfiguration() finds translation resources within bundles'
```

welcoMattic:
Thanks @Jibbarth for your contribution. I think we should consider this as a feature, and it should be enabled under a configuration option. I understand your use case, but I'm not sure everyone needs to extract all messages from all bundles in their app. Could you add an option in translation configuration?

Jibbarth:
@welcoMattic thanks for the review 🤗 okay I'll give it a try. 

Any idea of configuration name that could be added ?

Jibbarth:
Implemented in #52456 as a feature

PR #52457, opened 2023-11-05
[Cache][HttpFoundation][Lock] Fix empty username/password for PDO PostgreSQL

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Currently [pdo_pgsql has a bug](https://github.com/php/php-src/issues/12423) where the username & password arguments have precedence over the DSN, even thought [according to the docs](https://www.php.net/manual/en/ref.pdo-pgsql.connection.php) it should be the other way. This was recently fixed on PHP's side, but it [won't be available till 8.4](https://github.com/php/php-src/pull/12424).

Since the bug is not present when the values passed are `null`, which are the default argument values anyway, this PR changes the default values of the properties to match.

### DIFF
--- a/src/Symfony/Component/Cache/Adapter/PdoAdapter.php
+++ b/src/Symfony/Component/Cache/Adapter/PdoAdapter.php
@@ -34,8 +34,8 @@ class PdoAdapter extends AbstractAdapter implements PruneableInterface
     private $dataCol = 'item_data';
     private $lifetimeCol = 'item_lifetime';
     private $timeCol = 'item_time';
-    private $username = '';
-    private $password = '';
+    private $username = null;
+    private $password = null;
     private $connectionOptions = [];
     private $namespace;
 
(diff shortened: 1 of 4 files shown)


HypeMC:
Don't know if I should ignore fabbot. Even though the default property values are useless on 5.4, they are required on 6.x because of the type hints.

src/Symfony/Component/Cache/Adapter/PdoAdapter.php
```diff
@@ -34,8 +34,8 @@ class PdoAdapter extends AbstractAdapter implements PruneableInterface
     private $dataCol = 'item_data';
     private $lifetimeCol = 'item_lifetime';
     private $timeCol = 'item_time';
-    private $username = '';
-    private $password = '';
+    private $username = null;
+    private $password = null;
```
HypeMC, line 38: Requires #52457 for the test to pass.

nicolas-grekas:
Thank you @HypeMC.

PR #43332, opened 2021-10-06
[Lock] Split PdoStore into DoctrineDbalStore

GromNaN:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Tickets       | Fix #42962 
| License       | MIT
| Doc PR        | WIP

Main changes:
- New: Created classes `DoctrineDbalStore` and `DoctrineDbalPostgreSqlStore`.
- Deprecated: Using `Doctrine\DBAL\Connection` or `<driver>://` DSN in `PdoStore` or `PostgreSqlStore` is deprecated in 5.4 and will be removed in 6.0. Use `DoctrineDbalStore` or `DoctrineDbalPostgreSqlStore` instead.
- New: Mysqli driver is supported by `DoctrineDbalStore` ([thanks to numerical params](https://github.com/symfony/symfony/pull/43332#discussion_r723548944))
- BC: When using a malformed DBAL DSN, the error will be thrown when the `DoctrineDbalStore` or `DoctrineDbalPostgreSqlStore` is instanced, instead of the first usage of the connection.

PR #52459, opened 2023-11-05
[Cache][HttpFoundation][Lock] Fix PDO store not creating table + add tests

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Unlike e.g. SQLite, PostgreSQL doesn't throw an exception on `$conn->prepare()` if the table is missing, it instead throws it on `$stmt->execute()`, so the table never gets created.

The table used to get created, but the behavior was broken with #43332.

### DIFF
--- a/src/Symfony/Component/Cache/Adapter/DoctrineDbalAdapter.php
+++ b/src/Symfony/Component/Cache/Adapter/DoctrineDbalAdapter.php
@@ -420,7 +420,8 @@ class DoctrineDbalAdapter extends AbstractAdapter implements PruneableInterface
             return $this->serverVersion;
         }
 
-        $conn = $this->conn->getWrappedConnection();
+        // The condition should be removed once support for DBAL <3.3 is dropped
+        $conn = method_exists($this->conn, 'getNativeConnection') ? $this->conn->getNativeConnection() : $this->conn->getWrappedConnection();
         if ($conn instanceof ServerInfoAwareConnection) {
             return $this->serverVersion = $conn->getServerVersion();
         }
--- a/src/Symfony/Component/Cache/Tests/Adapter/DoctrineDbalAdapterTest.php
+++ b/src/Symfony/Component/Cache/Tests/Adapter/DoctrineDbalAdapterTest.php
@@ -18,12 +18,13 @@ use Doctrine\DBAL\Driver\Middleware;
 use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\Schema\DefaultSchemaManagerFactory;
 use Doctrine\DBAL\Schema\Schema;
-use PHPUnit\Framework\SkippedTestSuiteError;
 use Psr\Cache\CacheItemPoolInterface;
 use Symfony\Component\Cache\Adapter\DoctrineDbalAdapter;
 use Symfony\Component\Cache\Tests\Fixtures\DriverWrapper;
 
 /**
+ * @requires extension pdo_sqlite
+ *
  * @group time-sensitive
  */
 class DoctrineDbalAdapterTest extends AdapterTestCase
@@ -32,10 +33,6 @@ class DoctrineDbalAdapterTest extends AdapterTestCase
 
     public static function setUpBeforeClass(): void
     {
-        if (!\extension_loaded('pdo_sqlite')) {
-            throw new SkippedTestSuiteError('Extension pdo_sqlite required.');
-        }
-
         self::$dbFile = tempnam(sys_get_temp_dir(), 'sf_sqlite_cache');
     }
 
@@ -107,13 +104,12 @@ class DoctrineDbalAdapterTest extends AdapterTestCase
     }
 
     /**
-     * @dataProvider provideDsn
+     * @dataProvider provideDsnWithSQLite
      */
-    public function testDsn(string $dsn, string $file = null)
+    public function testDsnWithSQLite(string $dsn, string $file = null)
     {
         try {
             $pool = new DoctrineDbalAdapter($dsn);
-            $pool->createTable();
 
             $item = $pool->getItem('key');
             $item->set('value');
@@ -125,12 +121,35 @@ class DoctrineDbalAdapterTest extends AdapterTestCase
         }
     }
 
-    public static function provideDsn()
+    public static function provideDsnWithSQLite()
     {
         $dbFile = tempnam(sys_get_temp_dir(), 'sf_sqlite_cache');
-        yield ['sqlite://localhost/'.$dbFile.'1', $dbFile.'1'];
-        yield ['sqlite3:///'.$dbFile.'3', $dbFile.'3'];
-        yield ['sqlite://localhost/:memory:'];
+        yield 'SQLite file' => ['sqlite://localhost/'.$dbFile.'1', $dbFile.'1'];
+        yield 'SQLite3 file' => ['sqlite3:///'.$dbFile.'3', $dbFile.'3'];
+        yield 'SQLite in memory' => ['sqlite://localhost/:memory:'];
+    }
+
+    /**
+     * @requires extension pdo_pgsql
+     *
+     * @group integration
+     */
+    public function testDsnWithPostgreSQL()
+    {
+        if (!$host = getenv('POSTGRES_HOST')) {
+            $this->markTestSkipped('Missing POSTGRES_HOST env variable');
+        }
+
+        try {
+            $pool = new DoctrineDbalAdapter('pgsql://postgres:password@'.$host);
+
+            $item = $pool->getItem('key');
+            $item->set('value');
+            $this->assertTrue($pool->save($item));
+        } finally {
+            $pdo = new \PDO('pgsql:host='.$host.';user=postgres;password=password');
+            $pdo->exec('DROP TABLE IF EXISTS cache_items');
+        }
     }
 
     protected function isPruned(DoctrineDbalAdapter $cache, string $name): bool
--- a/src/Symfony/Component/Lock/Store/PdoStore.php
+++ b/src/Symfony/Component/Lock/Store/PdoStore.php
@@ -115,7 +115,7 @@ class PdoStore implements PersistingStoreInterface
         try {
             $stmt = $conn->prepare($sql);
         } catch (\PDOException $e) {
-            if (!$conn->inTransaction() || \in_array($this->driver, ['pgsql', 'sqlite', 'sqlsrv'], true)) {
+            if ($this->isTableMissing($e) && (!$conn->inTransaction() || \in_array($this->driver, ['pgsql', 'sqlite', 'sqlsrv'], true))) {
                 $this->createTable();
             }
             $stmt = $conn->prepare($sql);
@@ -127,8 +127,18 @@ class PdoStore implements PersistingStoreInterface
         try {
             $stmt->execute();
         } catch (\PDOException $e) {
-            // the lock is already acquired. It could be us. Let's try to put off.
-            $this->putOffExpiration($key, $this->initialTtl);
+            if ($this->isTableMissing($e) && (!$conn->inTransaction() || \in_array($this->driver, ['pgsql', 'sqlite', 'sqlsrv'], true))) {
+                $this->createTable();
+
+                try {
+                    $stmt->execute();
+                } catch (\PDOException $e) {
+                    $this->putOffExpiration($key, $this->initialTtl);
+                }
+            } else {
+                // the lock is already acquired. It could be us. Let's try to put off.
+                $this->putOffExpiration($key, $this->initialTtl);
+            }
         }
 
         $this->randomlyPrune();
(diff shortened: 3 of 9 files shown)


src/Symfony/Component/Lock/Store/PdoStore.php
```diff
@@ -316,4 +327,21 @@ private function getCurrentTimestampStatement(): string
...
+    {
+        $driver = $this->getDriver();
+        $code = $exception->getCode();
+
+        switch (true) {
+            case 'pgsql' === $driver && '42P01' === $code:
+            case 'sqlite' === $driver && str_contains($exception->getMessage(), 'no such table:'):
+            case 'oci' === $driver && 942 === $code:
+            case 'sqlsrv' === $driver && 208 === $code:
+            case 'mysql' === $driver && 1146 === $code:
```
HypeMC, line 340: Taken from Doctrine.

derrabus:
Can we add a test?

HypeMC:
@derrabus Didn't notice postgres was already part of the integration suit, I'll add a test.

src/Symfony/Component/Lock/Store/PdoStore.php
```diff
@@ -40,8 +40,8 @@ class PdoStore implements PersistingStoreInterface
     private $conn;
     private $dsn;
     private $driver;
-    private $username = '';
-    private $password = '';
+    private $username = null;
+    private $password = null;
```
HypeMC: Requires #52457 for the test to pass.

src/Symfony/Component/Lock/Tests/Store/PdoStoreTest.php
```diff
@@ -99,7 +99,11 @@ public function testDsn(string $dsn, string $file = null)
...
     {
         $dbFile = tempnam(sys_get_temp_dir(), 'sf_sqlite_cache');
-        yield ['sqlite:'.$dbFile.'2', $dbFile.'2'];
-        yield ['sqlite::memory:'];
+        yield 'SQLite file' => ['sqlite:'.$dbFile.'2', $dbFile.'2'];
+        yield 'SQLite in memory' => ['sqlite::memory:'];
+
+        if ($host = getenv('POSTGRES_HOST')) {
+            yield 'PostgreSQL' => ['pgsql:host='.$host.';user=postgres;password=password'];
+        }
```
HypeMC: @derrabus This should be ok, without my changes this test fails:

```
1) Symfony\Component\Lock\Tests\Store\PdoStoreTest::testDsn with data set "PostgreSQL" ('pgsql:host=localhost;user=pos...d=pass')
PDOException: SQLSTATE[42P01]: Undefined table: 7 ERROR:  relation "lock_keys" does not exist
LINE 1: UPDATE lock_keys SET key_expiration = CAST(EXTRACT(epoch FRO...
```

src/Symfony/Component/Lock/Store/PdoStore.php
```diff
@@ -127,8 +127,19 @@ public function save(Key $key)
         try {
             $stmt->execute();
         } catch (\PDOException $e) {
-            // the lock is already acquired. It could be us. Let's try to put off.
-            $this->putOffExpiration($key, $this->initialTtl);
+            if ($this->isTableMissing($e) && (!$conn->inTransaction() || \in_array($this->driver, ['pgsql', 'sqlite', 'sqlsrv'], true))) {
```
HypeMC, line 130: I just noticed that the `PdoAdapter` has [the same logic](https://github.com/symfony/symfony/blob/5725df0b0dae86a5e7cbc30493f4c900412b7882/src/Symfony/Component/Cache/Adapter/PdoAdapter.php#L321-L328) but without the `$this->isTableMissing()` part so I'm wondering, should I add it there as well, or remove it form here to keep it consistent?

src/Symfony/Component/Lock/Store/PdoStore.php
nicolas-grekas: rebase unlocked

HypeMC: Done

src/Symfony/Component/Lock/Store/PdoStore.php
nicolas-grekas: yes indeed, also in DoctrineDbalAdapter, right?

stof: The DBAL adapter catches TableNotFoundException so it is already fine (and the DBAL adapter uses the DBAL API that does the preparation and the execution internally so the try/catch wraps both).

nicolas-grekas: the try/catch is around the call to prepare, not to execute, that's why I'm wondering

HypeMC: I've added the logic to `PdoAdapter`, the `DoctrineDbalAdapter` already works as expected. I've added some tests to the adapters/stores to make super everything works as expected.

nicolas-grekas:
There are some deprecations, are they related?

HypeMC:
> There are some deprecations, are they related?

@nicolas-grekas Should be good now.

nicolas-grekas:
Aren't the remaining deprecations related?

https://github.com/symfony/symfony/actions/runs/6788097634/job/18452337658?pr=52459#step:11:1001

HypeMC:
> Aren't the remaining deprecations related?
> 
> https://github.com/symfony/symfony/actions/runs/6788097634/job/18452337658?pr=52459#step:11:1001

@nicolas-grekas Yes, you're right, took me a while to figure this one out. I initially thought that it wasn't related to my changes cause I didn't make any in `DoctrinePostgreSqlIntegrationTest`. However, after a little debugging it turns out that the deprecations were being triggered because I never cleaned up the database (dropped the tables) after the tests were done, which then triggered some additional code paths in `DoctrinePostgreSqlIntegrationTest` that use deprecated Doctrine methods. Sorry for the confusion.

src/Symfony/Component/Cache/Tests/Adapter/DoctrineDbalAdapterTest.php
```diff
@@ -107,13 +104,15 @@ public function testConfigureSchemaTableExists()
     }
 
     /**
+     * @group legacy
```
nicolas-grekas: Legacy test cases are meant to be removed in the next major, I guess that's not the case for this one.
Can we remove this annotation?

src/Symfony/Component/Cache/Adapter/DoctrineDbalAdapter.php
```diff
@@ -420,7 +420,7 @@ private function getServerVersion(): string
             return $this->serverVersion;
         }
 
-        $conn = $this->conn->getWrappedConnection();
+        $conn = method_exists($this->conn, 'getNativeConnection') ? $this->conn->getNativeConnection() : $this->conn->getWrappedConnection();
```
GromNaN, line 424: You can add a comment that it was added in doctine/dbal 3.3 https://github.com/doctrine/dbal/pull/5037 so it's easier to remove when support for older version will be dropped.

src/Symfony/Component/Cache/Tests/Adapter/DoctrineDbalAdapterTest.php
```diff
@@ -122,15 +120,22 @@ public function testDsn(string $dsn, string $file = null)
...
     {
         $dbFile = tempnam(sys_get_temp_dir(), 'sf_sqlite_cache');
-        yield ['sqlite://localhost/'.$dbFile.'1', $dbFile.'1'];
-        yield ['sqlite3:///'.$dbFile.'3', $dbFile.'3'];
-        yield ['sqlite://localhost/:memory:'];
+        yield 'SQLite file' => ['sqlite://localhost/'.$dbFile.'1', 'sqlite:'.$dbFile.'1', $dbFile.'1'];
+        yield 'SQLite3 file' => ['sqlite3:///'.$dbFile.'3', 'sqlite:'.$dbFile.'3', $dbFile.'3'];
+        yield 'SQLite in memory' => ['sqlite://localhost/:memory:', 'sqlite::memory:'];
+
+        if ($host = getenv('POSTGRES_HOST')) {
```
GromNaN: It would be more explicit to have a separate test for Postgres, which could be marked as skipped when `POSTGRES_HOST` is missing.

src/Symfony/Component/Cache/Tests/Adapter/DoctrineDbalAdapterTest.php
HypeMC: Done, https://github.com/symfony/symfony/compare/ccdf227a72a678c87f382bfc0f20ba021ab9f192..710c1c933ff64a22dc11ea408336b24f5880b615

I had to fix the Doctrine deprecation or else the CI would fail.

nicolas-grekas: perfect, thanks

src/Symfony/Component/Cache/Tests/Adapter/DoctrineDbalAdapterTest.php
```diff
@@ -107,30 +104,55 @@ public function testConfigureSchemaTableExists()
...
         try {
             $pool = new DoctrineDbalAdapter($dsn);
-            $pool->createTable();
 
             $item = $pool->getItem('key');
             $item->set('value');
             $this->assertTrue($pool->save($item));
         } finally {
+            $pdo = new \PDO($pdoDsn);
+            $pdo->exec('DROP TABLE IF EXISTS cache_items');
```
GromNaN: Was there a problem with deleting the DB file just after?

src/Symfony/Component/Cache/Adapter/DoctrineDbalAdapter.php
HypeMC: Done, https://github.com/symfony/symfony/compare/710c1c933ff64a22dc11ea408336b24f5880b615..b6e48698c382744f28713eaf05dc81690b786365

src/Symfony/Component/Cache/Tests/Adapter/DoctrineDbalAdapterTest.php
HypeMC: Done, https://github.com/symfony/symfony/compare/710c1c933ff64a22dc11ea408336b24f5880b615..b6e48698c382744f28713eaf05dc81690b786365

src/Symfony/Component/Cache/Tests/Adapter/DoctrineDbalAdapterTest.php
HypeMC: No, it just makes more sense to do the drop before deleting the file, otherwise the db is already gone when the sql is executed (if it's an SQLite file).

GromNaN: There's no point in deleting the table if the entire database file is deleted subsequently.

HypeMC: Sorry, completely misunderstood the question, you're right, it's not needed any more, see https://github.com/symfony/symfony/compare/b6e48698c382744f28713eaf05dc81690b786365..6a4254d90a3f8aab8728a9ac3ad08776b7df56d6

HypeMC: And https://github.com/symfony/symfony/compare/6a4254d90a3f8aab8728a9ac3ad08776b7df56d6..b75880ec4d38d951fe48e1612ca77bf976077d2e

GromNaN (approved):

src/Symfony/Component/Lock/Tests/Store/PdoStoreTest.php
```diff
@@ -96,10 +94,36 @@ public function testDsn(string $dsn, string $file = null)
...
+    }
+
+    /**
+     * @requires extension pdo_pgsql
+     *
+     * @group integration
+     */
+    public function testDsnWithPostgreSQL()
+    {
+        if (!($host = getenv('POSTGRES_HOST'))) {
```
nicolas-grekas: For all similar lines:
```suggestion
        if (!$host = getenv('POSTGRES_HOST')) {
```

nicolas-grekas (approved):

src/Symfony/Component/Lock/Tests/Store/PdoStoreTest.php
HypeMC: Done, https://github.com/symfony/symfony/compare/b75880ec4d38d951fe48e1612ca77bf976077d2e..cb5d832e4e9a58da300ce0a737613950114bfa34

nicolas-grekas:
Thank you @HypeMC.

PR #52469, opened 2023-11-06
Check whether secrets are empty and mark them all as sensitive

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Some security-hardening.

I made this throw a deprecation (as was already the case in some places) but I'd be fine making them throw an exception right away since an empty secret is a bold mistake. WDYT? (Webhook was experimental in 6.3 so it's fine throwing there I think.)

### DIFF
--- a/src/Symfony/Component/HttpFoundation/UriSigner.php
+++ b/src/Symfony/Component/HttpFoundation/UriSigner.php
@@ -12,8 +12,6 @@
 namespace Symfony\Component\HttpFoundation;
 
 /**
- * Signs URIs.
- *
  * @author Fabien Potencier <fabien@symfony.com>
  */
 class UriSigner
@@ -22,11 +20,14 @@ class UriSigner
     private string $parameter;
 
     /**
-     * @param string $secret    A secret
      * @param string $parameter Query string parameter to use
      */
     public function __construct(#[\SensitiveParameter] string $secret, string $parameter = '_hash')
     {
+        if (!$secret) {
+            throw new \InvalidArgumentException('A non-empty secret is required.');
+        }
+
         $this->secret = $secret;
         $this->parameter = $parameter;
     }
--- a/src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
+++ b/src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
@@ -14,6 +14,7 @@ namespace Symfony\Component\Security\Http\RateLimiter;
 use Symfony\Component\HttpFoundation\RateLimiter\AbstractRequestRateLimiter;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\RateLimiter\RateLimiterFactory;
+use Symfony\Component\Security\Core\Exception\InvalidArgumentException;
 use Symfony\Component\Security\Http\SecurityRequestAttributes;
 
 /**
@@ -35,10 +36,10 @@ final class DefaultLoginRateLimiter extends AbstractRequestRateLimiter
      */
     public function __construct(RateLimiterFactory $globalFactory, RateLimiterFactory $localFactory, #[\SensitiveParameter] string $secret = '')
     {
-        if ('' === $secret) {
-            trigger_deprecation('symfony/security-http', '6.4', 'Calling "%s()" with an empty secret is deprecated. A non-empty secret will be mandatory in version 7.0.', __METHOD__);
-            // throw new \Symfony\Component\Security\Core\Exception\InvalidArgumentException('A non-empty secret is required.');
+        if (!$secret) {
+            throw new InvalidArgumentException('A non-empty secret is required.');
         }
+
         $this->globalFactory = $globalFactory;
         $this->localFactory = $localFactory;
         $this->secret = $secret;
--- a/src/Symfony/Component/Mailer/Bridge/Brevo/Webhook/BrevoRequestParser.php
+++ b/src/Symfony/Component/Mailer/Bridge/Brevo/Webhook/BrevoRequestParser.php
@@ -41,7 +41,7 @@ final class BrevoRequestParser extends AbstractRequestParser
         ]);
     }
 
-    protected function doParse(Request $request, string $secret): ?AbstractMailerEvent
+    protected function doParse(Request $request, #[\SensitiveParameter] string $secret): ?AbstractMailerEvent
     {
         $content = $request->toArray();
         if (
--- a/src/Symfony/Component/Mailer/Bridge/Mailgun/Webhook/MailgunRequestParser.php
+++ b/src/Symfony/Component/Mailer/Bridge/Mailgun/Webhook/MailgunRequestParser.php
@@ -17,6 +17,7 @@ use Symfony\Component\HttpFoundation\RequestMatcher\IsJsonRequestMatcher;
 use Symfony\Component\HttpFoundation\RequestMatcher\MethodRequestMatcher;
 use Symfony\Component\HttpFoundation\RequestMatcherInterface;
 use Symfony\Component\Mailer\Bridge\Mailgun\RemoteEvent\MailgunPayloadConverter;
+use Symfony\Component\Mailer\Exception\InvalidArgumentException;
 use Symfony\Component\RemoteEvent\Event\Mailer\AbstractMailerEvent;
 use Symfony\Component\RemoteEvent\Exception\ParseException;
 use Symfony\Component\Webhook\Client\AbstractRequestParser;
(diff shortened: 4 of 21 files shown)


OskarStark (approved):

src/Symfony/Component/HttpFoundation/UriSigner.php
```diff
@@ -22,11 +20,15 @@ class UriSigner
...
 
     /**
-     * @param string $secret    A secret
      * @param string $parameter Query string parameter to use
      */
     public function __construct(#[\SensitiveParameter] string $secret, string $parameter = '_hash')
     {
+        if (!$secret) {
+            trigger_deprecation('symfony/http-foundation', '6.4', 'Calling "%s()" with an empty secret is deprecated. A non-empty secret will be mandatory in version 7.0.', __METHOD__);
+            // throw new \InvalidArgumentException('A non-empty secret is required.');
```
dkarlovi: IMO, when check() runs and UriSigner sees the secret is empty, it should start throwing immediately even in 6.4

If you're asking me to validate an URL and the secret is empty, that's the worst possible case scenario for the app (because of the circumstances) and you want to patch it ASAP. I understand it's a BC break, but since it allows arbitrary RCE as it stands, it's a huge hole and you actually want a BC here to notice your bad config, everyone who'll be bitten by it should actually thank you instead of being upset. :+1:

dbu (approved):
thanks! was not aware the secret is used in so many places.

imho it would make sense to deprecate the empty secret in all currently supported symfony versions, not only 6.4.

about the throwing exceptions immediately i am less sure. this problem has been around for quite a long time, and is only a problem when mistakenly not setting the secret (likely some deployment / environment configuration mistake). i would not throw the exception in 6.4 so people can notice the issue in their deprecation logs if they care. and when upgrading to 7.0 we start throwing.

fabpot:
As having an empty secret can only be a mistake, I would throw an exception in 6.4.

fabpot:
Would it make sense (in addition) to throw an exception at build time when the secret is empty (fail early)?

nicolas-grekas:
> Would it make sense (in addition) to throw an exception at build time when the secret is empty (fail early)?

at build time we reference an env var so we cannot check this reliably I think.

nicolas-grekas:
> As having an empty secret can only be a mistake, I would throw an exception in 6.4.

Updated!

src/Symfony/Component/HttpFoundation/UriSigner.php
```diff
@@ -22,11 +20,14 @@ class UriSigner
     private string $parameter;
 
     /**
-     * @param string $secret    A secret
      * @param string $parameter Query string parameter to use
      */
     public function __construct(#[\SensitiveParameter] string $secret, string $parameter = '_hash')
     {
+        if (!$secret) {
```
fancyweb, line 27: Do we consider `'0'` as an empty secret?

nicolas-grekas: yes, it's equally weak

javiereguiluz:
I like this, but I wonder if the error message is enough to fully understand the problem and how to solve it.

Imagine reading this with not much additional context --> `A non-empty secret is required.`

Would something like the following be acceptable?

```
The "secret" value cannot be empty. Configure it in the "framework.secret" option.
```

stof (approved):

nicolas-grekas:
> The "secret" value cannot be empty. Configure it in the "framework.secret" option.

I'm not following this proposal because the exceptions should be agnostic of FWB.
We can improve the message if ppl see it too often, but to me this shouldn't happen in practice. It's a safeguard.

src/Symfony/Component/Security/Http/RateLimiter/DefaultLoginRateLimiter.php
```diff
@@ -35,10 +36,10 @@ final class DefaultLoginRateLimiter extends AbstractRequestRateLimiter
      */
     public function __construct(RateLimiterFactory $globalFactory, RateLimiterFactory $localFactory, #[\SensitiveParameter] string $secret = '')
     {
-        if ('' === $secret) {
-            trigger_deprecation('symfony/security-http', '6.4', 'Calling "%s()" with an empty secret is deprecated. A non-empty secret will be mandatory in version 7.0.', __METHOD__);
-            // throw new \Symfony\Component\Security\Core\Exception\InvalidArgumentException('A non-empty secret is required.');
+        if (!$secret) {
```
RobertMe, line 39: This is a breaking change. Both the argument, as the deprecation, have been added in  #51434 (@Spomky), which is also 6.4+. Converting the deprecation to an exception means that on  upgrade from 6.3 and lower users of this class will immediately run into this exception. (Because it is a new argument it obviously isn't provided yet).

It isn't a big deal to fix, but just wanted to give a heads up about this. Wheteher you decide to revert to a deprecation on 6.4 and keep throwing on 7.0 is up to you. As said, it's not a big issue to fix as an end user, but it's a BC break without previous warning nevertheless.

nicolas-grekas: We decided it's OK to add this exception when security is at risk.
That being said, I'm not sure why we use the secret here to hash the logged info.

Spomky: > 'm not sure why we use the secret here to hash the logged info.

This is the [pepper](https://en.wikipedia.org/wiki/Pepper_(cryptography)) for preventing [rainbow tables](https://en.wikipedia.org/wiki/Rainbow_table) to be used on IP addresses or username/IPs combinations
Edit: article (in French) on bad design where simple hashes are used issued few days after the PR => https://theconversation.com/les-dangers-de-la-cryptographie-non-maitrisee-212881

dkarlovi: As noted in the MR, if you notice this breaking change happening, you had your app **severely** misconfigured and should send thanks to the core team for pointing it out. 😃

RobertMe: > We decided it's OK to add this exception when security is at risk.

I do get that argument for the existing `$secret` arguments, this as it indeed is a security risk to not provide it (/use an empty string). But in this case it isn't an existing argument for which the user (knowingly) provided an unsecure value. It's a new parameter, so to not break BC it does require a default value. But then this default value is considered invalid and still throws an exception.

So if it is decided that the exception must stay, I now think it makes sense to just drop the default value. Then users are at least made aware they're not providing a (now required, for security) parameter. Instead of the currently "vague" exception that they're providing an invalid value (to then discover they don't provide any value at all, the default is "broken", and they must provide a value themselfs).

nicolas-grekas: Can you please send a PR on branch 6.4?

RobertMe: So in short / to put it differently:
On version 6.3 one must provide 2 parameters, `$globalFactory` and `$localFactory`. But the exact same code on 6.4 leads to an exception, that a non-empty secret must be provided. But I'm not providing a secret at all. So in an ordinary case this would be an "illegal" change as it breaks BC. For the sake of security I do get the "we want to require a secret". But IMO then the error should be about the missing parameter (so just the PHP error), and not a custom exception. I.e.: dropping the default value so PHP gives an error for the invalid method call / missing parameter, and keeping the check it's not empty as well. As IMO the standard PHP error is a better DX because it's a common error, the custom error requires more time to figure out what exactly is going wrong and why (to still come to the conclusion the argument is now required).

And in all(/most?) other cases changed by this PR the `$secret` parameter already existed. So IMO indeed can be considered a bug / security fix where values which have always been invalid now actually are detected and result in an error. But that's not the case for this specific scenario (as it's a new parameter which could not have been provided beforehand).

RobertMe: (Sorry, I apparently missed some comments in between)

> Can you please send a PR on branch 6.4?

I'll try to do so, but it will most likely be after the weekend. (I read the blogpost about #51434 on the go, and it reminded me about the fix I had to make yesterday).
And just to be sure: I presume a PR to drop the default value?

> As noted in the MR, if you notice this breaking change happening, you had your app severely misconfigured and should send thanks to the core team for pointing it out. 😃

That's not the case here. Because it is a _new_ parameter. And I can't severly misconfigure anything when there was nothing to configure it in the first place. The "configuration" is added in 6.4 and the default is "severely misconfigured".

PR #52471, opened 2023-11-06
[HttpKernel] Add `ControllerResolver::allowControllers()` to define which callables are legit controllers when the `_check_controller_is_allowed` request attribute is set

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

Right now, when one doesn't configure properly their APP_SECRET, this can too easily lead to an RCE.

This PR proposes to harden security by rejecting any not-allowed controllers when the `_check_controller_is_allowed` request attribute is set. We leverage this in FragmentListener to close the RCE gap.

In order to allow a controller, one should call `ControllerResolver::allowControllers()` during instantiation to tell which types or attributes should be accepted. #[AsController] is always allowed, and FrameworkBundle also allows instances of `AbstractController`.

Third-party bundles that provide controllers meant to be used as fragments should ensure their controllers are allowed by adding the method call to the `controller_resolver` service definition.

I propose this as a late 6.4 feature so that we can provide this hardening right away in 7.0. In 6.4, this would be only a deprecation.

### DIFF
--- a/src/Symfony/Component/HttpKernel/Controller/ControllerResolver.php
+++ b/src/Symfony/Component/HttpKernel/Controller/ControllerResolver.php
@@ -12,7 +12,9 @@
 namespace Symfony\Component\HttpKernel\Controller;
 
 use Psr\Log\LoggerInterface;
+use Symfony\Component\HttpFoundation\Exception\BadRequestException;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Attribute\AsController;
 
 /**
  * This implementation uses the '_controller' request attribute to determine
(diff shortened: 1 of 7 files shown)


src/Symfony/Component/HttpKernel/Controller/ControllerResolver.php
```diff
@@ -199,4 +221,59 @@ private function getClassMethodsWithoutMagicMethods($classOrObject): array
...
+            if (isset($this->allowedControllerAttributes[$attribute->getName()])) {
+                return $controller;
+            }
+        }
+
+        if (str_contains($name, '@anonymous')) {
+            $name = preg_replace_callback('/[a-zA-Z_\x7f-\xff][\\\\a-zA-Z0-9_\x7f-\xff]*+@anonymous\x00.*?\.php(?:0x?|:[0-9]++\$)[0-9a-fA-F]++/', fn ($m) => class_exists($m[0], false) ? (get_parent_class($m[0]) ?: key(class_implements($m[0])) ?: 'class').'@anonymous' : $m[0], $name);
+        }
+
+        if (-1 === $request->attributes->get('_check_controller_is_allowed')) {
```
stof, line 271: I'm wondering whether we should add the special value `-1` here to trigger the deprecation (while setting `true` directly already triggers the exception) instead of actually doing what the deprecation says by throwing the exception only in 7.0 (and not having this special `-1` value)

nicolas-grekas: I think we need a way for ppl to opt into the new behavior provided by the attribute right in 6.4. But for FragmentListener, we need a way to trigger a deprecation. That's why I made it this way.

stof (approved):

chalasr (approved):

marein:
@nicolas-grekas Great feature :). I've a question regarding it's implementation and I thought it's a good place to ask here first. The current functionality automatically allows controllers with the `#[AsController]` argument or those that inherit from `AbstractController`. However, it doesn't consider controllers tagged with `controller.service_arguments`, which would require special handling. I'm considering submitting a PR that automatically allows all controllers tagged with `controller.service_arguments` too, because I think that it shouldn't make a difference whether one uses `#[AsController]` or `controller.service_arguments`. However, I want to ensure I'm not overlooking any considerations you might have had before proceeding. Looking forward to your feedback :).

nicolas-grekas:
I don't know why it's not supported already but I'd happily review a PR that does so.

PR #45729, opened 2022-03-13
[HttpClient] Add resolve to copy as Curl

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Adds the resolve option to the generated Curl command.

PR #52472, opened 2023-11-06
[HttpClient][WebProfilerBundle] Do not generate cURL command when files are uploaded

### LINKED ISSUE #51366: [HttpClient] Uploading files causes the data collector to crash

≥ 6.3.0

### Description

When investigating https://github.com/symfony/symfony/discussions/51360 I found out the `HttpClientTrait::normalizeBody` checks if upload streams have ended when they are nested in `body` (cf. #49911).

In that case, the `HttpClientDataCollector` will trigger an exception because it will call `normalizeBody` again when all streams have been read (to generate the cURL command).

### How to reproduce

```php
$stream = fopen('filepath', 'r');

$httpClient->request('POST', '/endpoint', [
    'body' => ['stream' => $stream],
]);
```
will trigger a `TransportException` if the profiler is enabled.

### Possible Solution

The `HttpClientDataCollector` does not generate a cURL command when the body is a stream, so I guess it shouldn’t either in this case.

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #51366
| License       | MIT

I also removed `@requires extension openssl` annotations since that does not seem to be the case since #45729.

Failures in AppVeyor occur because double quotes are missing around `--data-raw` values. Possibly related to #52429.

### DIFF
--- a/src/Symfony/Component/HttpClient/DataCollector/HttpClientDataCollector.php
+++ b/src/Symfony/Component/HttpClient/DataCollector/HttpClientDataCollector.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Component\HttpClient\DataCollector;
 
+use Symfony\Component\HttpClient\Exception\TransportException;
 use Symfony\Component\HttpClient\HttpClientTrait;
 use Symfony\Component\HttpClient\TraceableHttpClient;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Up to fix the quote issue in the same PR for tests?

MatTheCat:
@nicolas-grekas I'm not sure what should be done regarding these double quotes since it seems to be Windows-specific 🤔

nicolas-grekas:
I think we should just adjust the expected output depending on Windows or not. Failing tests should pass.

nicolas-grekas:
Or maybe @alexandre-daubois can provide the fix for the tests?

nicolas-grekas:
Thank you @MatTheCat.

MatTheCat:
I'll look into it tomorrow if @alexandre-daubois hasn't in the meantime.

alexandre-daubois:
Thank you for proposing your help Mathieu! Trying to understand what I missed in my PR 😄

MatTheCat:
From what I saw (did not investigate further), `HttpClientDataCollector::escapePayload` returns different strings whether on Unix or Windows, but `HttpClientDataCollector::provideCurlRequests` expects the Unix return. That would explain why passing tests fail on AppVeyor.

PR #52473, opened 2023-11-06
[HttpKernel] Fix quotes expectations in tests

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Following https://github.com/symfony/symfony/pull/52472#issuecomment-1795734816

On Windows, quotes are only added when necessary. This means that for a value like `foobar`, no quotes are added. I updated the test so the first dataset shows the escaping whereas the second updated dataset shows that it is expected that no quotes are added around data.


nicolas-grekas:
Thank you @alexandre-daubois.

PR #52476, opened 2023-11-06
[Messenger] fix compatibility with Doctrine DBAL 4

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
+++ b/src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
@@ -177,7 +177,24 @@ class Connection implements ResetInterface
 
             // Append pessimistic write lock to FROM clause if db platform supports it
             $sql = $query->getSQL();
-            if (preg_match('/FROM (.+) WHERE/', (string) $sql, $matches)) {
+
+            // Wrap the rownum query in a sub-query to allow writelocks without ORA-02014 error
+            if ($this->driverConnection->getDatabasePlatform() instanceof OraclePlatform) {
+                $query = $this->createQueryBuilder('w')
+                    ->where('w.id IN ('.str_replace('SELECT a.* FROM', 'SELECT a.id FROM', $sql).')');
+
+                if (method_exists(QueryBuilder::class, 'forUpdate')) {
+                    $query->forUpdate();
+                }
+
+                $sql = $query->getSQL();
+            } elseif (method_exists(QueryBuilder::class, 'forUpdate')) {
+                $query->forUpdate();
+                try {
+                    $sql = $query->getSQL();
+                } catch (DBALException $e) {
+                }
+            } elseif (preg_match('/FROM (.+) WHERE/', (string) $sql, $matches)) {
                 $fromClause = $matches[1];
                 $sql = str_replace(
                     sprintf('FROM %s WHERE', $fromClause),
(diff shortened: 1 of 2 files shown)


xabbuh:
see doctrine/dbal#6191

derrabus:
The idea behind that feature was that Symfony could leverage the query builder for building the "skip locked rows" query.

xabbuh:
Thanks for your feedback. I guess it's better now.

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
```diff
@@ -177,7 +178,16 @@ public function get(): ?array
...
+
+            if (method_exists(QueryBuilder::class, 'forUpdate')) {
+                $query->forUpdate();
+                try {
+                    $sql = $query->getSQL();
+                } catch (Exception $e) {
+                }
+            }
+
+            if (!method_exists(QueryBuilder::class, 'forUpdate') && preg_match('/FROM (.+) WHERE/', (string) $sql, $matches)) {
```
nicolas-grekas:
```suggestion
            } elseif (preg_match('/FROM (.+) WHERE/', (string) $sql, $matches)) {
```

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
```diff
@@ -194,8 +204,12 @@ public function get(): ?array
             }
 
             // use SELECT ... FOR UPDATE to lock table
+            if (!method_exists(QueryBuilder::class, 'forUpdate')) {
```
nicolas-grekas, line 207: We cannot lock anymore in DBAL4?

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
```diff
@@ -177,7 +178,16 @@ public function get(): ?array
 
             // Append pessimistic write lock to FROM clause if db platform supports it
             $sql = $query->getSQL();
-            if (preg_match('/FROM (.+) WHERE/', (string) $sql, $matches)) {
+
+            if (method_exists(QueryBuilder::class, 'forUpdate')) {
+                $query->forUpdate();
+                try {
+                    $sql = $query->getSQL();
+                } catch (Exception $e) {
```
nicolas-grekas: can this be a bit more specific? I don't like swallowing any exceptions :grimacing:

nicolas-grekas: !
```suggestion
                } catch (DBALException $e) {
```

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
xabbuh: as far as I understand the related changes in Doctrine DBAL this now happens with the `forUpdate()` call above

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
xabbuh: https://github.com/doctrine/dbal/pull/6191#discussion_r1379382594

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
stof: indeed

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
```diff
@@ -177,7 +177,14 @@ public function get(): ?array
 
             // Append pessimistic write lock to FROM clause if db platform supports it
             $sql = $query->getSQL();
-            if (preg_match('/FROM (.+) WHERE/', (string) $sql, $matches)) {
+
+            if (method_exists(QueryBuilder::class, 'forUpdate')) {
+                $query->forUpdate();
```
stof, line 192: For the case of the oracle platform, this is now applying the lock on the inner query and not the outer one (see the specific code below). Is this expected ?

nicolas-grekas (approved):

src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php
xabbuh: it's probably better not to change the outcome here, updated

stof (approved):

nicolas-grekas:
Thank you @xabbuh.

PR #52491, opened 2023-11-07
[String] Method toByteString conversion using iconv is unreachable

### LINKED ISSUE #52489: [String] Method toByteString conversion using iconv is unreachable

5.4.30

### Description

Method `\Symfony\Component\String\AbstractString::toByteString` throws `ValueError` exception when unsupported encoding for `mb_convert_encoding` is passed instead of using fallback `iconv`.

### How to reproduce

`u('žsžsý')->toByteString('windows-1250');`

Throws `ValueError` with message `mb_convert_encoding(): Argument #2 ($to_encoding) must be a valid encoding, "windows-1250" given`.

Instead of returning `ByteString` with output of `iconv`

### Possible Solution

_No response_

### Additional Context

`iconv('UTF-8', 'windows-1250', 'žsžsý')` returns expected value.

Using PHP 8.1

It is possible that nobody noticed that PHP 8.0 added ValueError exception https://www.php.net/manual/en/function.mb-convert-encoding so the code is not compatible with php >= 8.0

Vincentv92:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Issues        | Fix #52489 
| License       | MIT

Originating from original issue: 
>It is possible that nobody noticed that PHP 8.0 added ValueError exception https://www.php.net/manual/en/function.mb-convert-encoding so the code is not compatible with php >= 8.0

That seems indeed to be the case

### DIFF
--- a/src/Symfony/Component/String/AbstractString.php
+++ b/src/Symfony/Component/String/AbstractString.php
@@ -577,8 +577,11 @@ abstract class AbstractString implements \Stringable, \JsonSerializable
         try {
             try {
                 $b->string = mb_convert_encoding($this->string, $toEncoding, 'UTF-8');
-            } catch (InvalidArgumentException $e) {
+            } catch (InvalidArgumentException|\ValueError $e) {
                 if (!\function_exists('iconv')) {
+                    if ($e instanceof \ValueError) {
+                        throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
+                    }
                     throw $e;
                 }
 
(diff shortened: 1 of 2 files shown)


derrabus (commented):
LGTM. Can we add a test to prevent another regression in the future?

xabbuh (approved):

src/Symfony/Component/String/AbstractString.php
```diff
@@ -577,7 +578,7 @@ public function toByteString(string $toEncoding = null): ByteString
         try {
             try {
                 $b->string = mb_convert_encoding($this->string, $toEncoding, 'UTF-8');
-            } catch (InvalidArgumentException $e) {
+            } catch (InvalidArgumentException|ValueError $e) {
                 if (!\function_exists('iconv')) {
                     throw $e;
```
nicolas-grekas, line 585: Here, we should wrap the ValueError in an InvalidArgumentException

nicolas-grekas:
Thank you @Vincentv92.

PR #42112, opened 2021-07-14
[HttpFoundation] fix FileBag under PHP 8.1

alexpott:
| Q             | A
| ------------- | ---
| Branch?       | 4.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no 
| Tickets       | Fix #... 
| License       | MIT
| Doc PR        | symfony/symfony-docs#... 

See https://php.watch/versions/8.1/$_FILES-full-path - in \Symfony\Component\HttpFoundation\FileBag::convertFileInformation() we check against a list of hardcoded keys. This logic breaks in PHP 8.1 because of the new key.

PR #52493, opened 2023-11-08
[HttpFoundation] Add `UploadedFile::getClientOriginalPath()` to support directory uploads

danielburger1337:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      |  no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

This removes the "hotfix" https://github.com/symfony/symfony/pull/42112/ and now supports "full_path" PHP file uploads.

Since PHP 8.1 is now the lowest allowed version of PHP, this shouldnt cause any problems. 

---

Although I'm quite certain that I understand the messy FileBag code, I'm very concerned for the potential impact this change might have because of the widespread use of http-foundation.

To the best of my knowledge, this won't break Symfony BC (yet) because this is all done by an `optional` argument in the constructor of UploadedFile (this should eventually be replaced by a mandatory argument though).

Also, I'm not quite happy with the tests. Since this is only a "passthrough" of the PHP value, this _shouldnt_ be a huge problem. 

Someone should definitly go through this change very carefully. The tests pass and I also threw some "real world" application tests at it and everything seems to work just fine. But because of the nature of HttpFoundation and its intangled mess, I can't guarantee that I accounted for every possible or even simple edge case.

---

Though it would be nice to still see this in 7.0, I know that it will probably be pushed back to 7.1.

### DIFF
--- a/src/Symfony/Component/HttpFoundation/CHANGELOG.md
+++ b/src/Symfony/Component/HttpFoundation/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Add `UploadedFile::getClientOriginalPath()`
+
 7.0
 ---
 
--- a/src/Symfony/Component/HttpFoundation/File/UploadedFile.php
+++ b/src/Symfony/Component/HttpFoundation/File/UploadedFile.php
@@ -35,6 +35,7 @@ class UploadedFile extends File
     private string $originalName;
     private string $mimeType;
     private int $error;
+    private string $originalPath;
 
     /**
      * Accepts the information of the uploaded file as provided by the PHP global $_FILES.
@@ -63,6 +64,7 @@ class UploadedFile extends File
     public function __construct(string $path, string $originalName, string $mimeType = null, int $error = null, bool $test = false)
     {
         $this->originalName = $this->getName($originalName);
+        $this->originalPath = strtr($originalName, '\\', '/');
         $this->mimeType = $mimeType ?: 'application/octet-stream';
         $this->error = $error ?: \UPLOAD_ERR_OK;
         $this->test = $test;
@@ -90,6 +92,21 @@ class UploadedFile extends File
     public function getClientOriginalExtension(): string
     {
         return pathinfo($this->originalName, \PATHINFO_EXTENSION);
+    }
+
+    /**
+     * Returns the original file full path.
+     *
+     * It is extracted from the request from which the file has been uploaded.
+     * This should not be considered as a safe value to use for a file name/path on your servers.
+     *
+     * If this file was uploaded with the "webkitdirectory" upload directive, this will contain
+     * the path of the file relative to the uploaded root directory. Otherwise this will be identical
+     * to getClientOriginalName().
+     */
+    public function getClientOriginalPath(): string
+    {
+        return $this->originalPath;
     }
 
     /**
--- a/src/Symfony/Component/HttpFoundation/FileBag.php
+++ b/src/Symfony/Component/HttpFoundation/FileBag.php
@@ -21,7 +21,7 @@ use Symfony\Component\HttpFoundation\File\UploadedFile;
  */
 class FileBag extends ParameterBag
 {
-    private const FILE_KEYS = ['error', 'name', 'size', 'tmp_name', 'type'];
+    private const FILE_KEYS = ['error', 'full_path', 'name', 'size', 'tmp_name', 'type'];
 
     /**
      * @param array|UploadedFile[] $parameters An array of HTTP files
@@ -65,18 +65,18 @@ class FileBag extends ParameterBag
         }
 
         $file = $this->fixPhpFilesArray($file);
-        $keys = array_keys($file);
+        $keys = array_keys($file + ['full_path' => null]);
         sort($keys);
 
-        if (self::FILE_KEYS == $keys) {
-            if (\UPLOAD_ERR_NO_FILE == $file['error']) {
+        if (self::FILE_KEYS === $keys) {
+            if (\UPLOAD_ERR_NO_FILE === $file['error']) {
                 $file = null;
             } else {
-                $file = new UploadedFile($file['tmp_name'], $file['name'], $file['type'], $file['error'], false);
+                $file = new UploadedFile($file['tmp_name'], $file['full_path'] ?? $file['name'], $file['type'], $file['error'], false);
             }
         } else {
             $file = array_map(fn ($v) => $v instanceof UploadedFile || \is_array($v) ? $this->convertFileInformation($v) : $v, $file);
-            if (array_keys($keys) === $keys) {
+            if (array_is_list($file)) {
                 $file = array_filter($file);
             }
         }
@@ -98,12 +98,10 @@ class FileBag extends ParameterBag
      */
     protected function fixPhpFilesArray(array $data): array
     {
-        // Remove extra key added by PHP 8.1.
-        unset($data['full_path']);
-        $keys = array_keys($data);
+        $keys = array_keys($data + ['full_path' => null]);
         sort($keys);
 
-        if (self::FILE_KEYS != $keys || !isset($data['name']) || !\is_array($data['name'])) {
+        if (self::FILE_KEYS !== $keys || !isset($data['name']) || !\is_array($data['name'])) {
             return $data;
         }
 
@@ -119,7 +117,9 @@ class FileBag extends ParameterBag
                 'type' => $data['type'][$key],
                 'tmp_name' => $data['tmp_name'][$key],
                 'size' => $data['size'][$key],
-            ]);
+            ] + (isset($data['full_path'][$key]) ? [
+                'full_path' => $data['full_path'][$key],
+            ] : []));
         }
 
         return $files;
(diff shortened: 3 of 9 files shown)


src/Symfony/Component/HttpFoundation/Tests/File/Fixtures/webkitdirectory/nested/test.txt
```diff
@@ -0,0 +1 @@
+nested webkitdirectory text
```
OskarStark, line 1: unused file?

nicolas-grekas:
That's a new feature for sure so for 7.1
What about also adding support for the `webkitdirectory` attribute in the Form component?

danielburger1337:
> What about also adding support for the `webkitdirectory` attribute in the Form component?

I haven't contributed to the form component yet, so I will check it out. Should be fairly simple though.

danielburger1337:
I just realized that merging this should break many legacy test cases where a file upload is manually submitted. 
See [FormTypeTest](https://github.com/symfony/symfony/blob/7.0/src/Symfony/Component/Form/Tests/Extension/Core/Type/FileTypeTest.php) as an example. These submit the mock file upload without the "full_path" option. 

Even though it is more work than simply making this option optional in FileBag, these tests SOULD be fixed because they do not adhere to the default PHP behaviour and in my opinion the FileBag should not account for that. It would also require to rework the entire logic of the FileBag.

alexisLefebvre:
> That's a new feature for sure so for 7.1 What about also adding support for the `webkitdirectory` attribute in the Form component?

See this answer from stof: https://github.com/symfony/symfony/pull/52528#issuecomment-1805529461

src/Symfony/Component/HttpFoundation/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ CHANGELOG
  * Throw an `InvalidArgumentException` when calling `Request::create()` with a malformed URI
  * Require explicit argument when calling `JsonResponse::setCallback()`, `Response::setExpires/setLastModified/setEtag()`, `MockArraySessionStorage/NativeSessionStorage::setMetadataBag()`, `NativeSessionStorage::setSaveHandler()`
  * Add argument `$statusCode` to `Response::sendHeaders()` and `StreamedResponse::sendHeaders()`
+ * Support PHP 8.1 "full_path" in `FileBag`
```
nicolas-grekas:
```suggestion
 * Add `UploadedFile::getClientOriginalPath()`
```

src/Symfony/Component/HttpFoundation/File/UploadedFile.php
```diff
@@ -50,23 +51,29 @@ class UploadedFile extends File
...
+     * @param int|null    $error            The error constant of the upload (one of PHP's UPLOAD_ERR_XXX constants); null defaults to UPLOAD_ERR_OK
+     * @param bool        $test             Whether the test mode is active
+     *                                      Local files are used in test mode hence the code should not enforce HTTP uploads
+     * @param string      $originalFullPath The original full path of the uploaded file
      *
      * @throws FileException         If file_uploads is disabled
      * @throws FileNotFoundException If the file does not exist
      */
-    public function __construct(string $path, string $originalName, string $mimeType = null, int $error = null, bool $test = false)
+    public function __construct(string $path, string $originalName, string $mimeType = null, int $error = null, bool $test = false, string $originalFullPath = null)
```
nicolas-grekas: I don't think we need a new constructor argument. Instead, we should accept a path in $originalName and populate `$originalPath` with it when `$originalName` contains a `/` (after normalizing `\`).
The code already sanitizes `$orginalName` so this is already supported.

src/Symfony/Component/HttpFoundation/File/UploadedFile.php
```diff
@@ -92,6 +99,21 @@ public function getClientOriginalExtension(): string
...
+     * Returns the original file full path.
+     *
+     * It is extracted from the request from which the file has been uploaded.
+     * This should not be considered as a safe value to use for a file name/path on your servers.
+     *
+     * If this file was uploaded with the "webkitdirectory" upload directive, this will contain
+     * the path of the file relative to the uploaded root directory. Otherwise this will be identical
+     * to getClientOriginalName().
+     */
+    public function getClientOriginalFullPath(): string
```
nicolas-grekas:
```suggestion
    public function getClientOriginalPath(): string
```

src/Symfony/Component/HttpFoundation/FileBag.php
```diff
@@ -72,7 +72,7 @@ protected function convertFileInformation(array|UploadedFile $file): array|Uploa
             if (\UPLOAD_ERR_NO_FILE == $file['error']) {
                 $file = null;
             } else {
-                $file = new UploadedFile($file['tmp_name'], $file['name'], $file['type'], $file['error'], false);
+                $file = new UploadedFile($file['tmp_name'], $file['name'], $file['type'], $file['error'], false, $file['full_path']);
```
nicolas-grekas:
```suggestion
                $file = new UploadedFile($file['tmp_name'], $file['full_path'], $file['type'], $file['error'], false);
```

nicolas-grekas (commented):
Thanks for working on this. I played a bit with the code+feature, here are my comments.

src/Symfony/Component/HttpFoundation/File/UploadedFile.php
danielburger1337: Yeah, this is a great idea

src/Symfony/Component/HttpFoundation/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+7.1
+---
+ * Add `UploadedFile::getClientOriginalPath()`
```
nicolas-grekas, line 7: a blank line should be added before

src/Symfony/Component/HttpFoundation/File/UploadedFile.php
```diff
@@ -63,6 +64,7 @@ class UploadedFile extends File
     public function __construct(string $path, string $originalName, string $mimeType = null, int $error = null, bool $test = false)
     {
         $this->originalName = $this->getName($originalName);
+        $this->originalPath = $originalName;
```
nicolas-grekas: just for normalizing a bit:
```suggestion
        $this->originalPath = strtr($originalName, '\\', '/');
```

danielburger1337: Is that Windows related? Otherwise I don't get why.

nicolas-grekas: could be browsers on windows sending this yes

nicolas-grekas:
I think the failure on `PsrHttpFactoryTest::testUploadErrNoFile` is related, maybe the ones of Form also.
Can you please have a look?

danielburger1337:
Yeah, this failure is related. Like I mentioned [here](https://github.com/symfony/symfony/pull/52493#issuecomment-1804971961), pretty much every test that submits a file upload manually will fail now and will have to be migrated because they don't include the `full_path` key. 

This is standard PHP behaviour since 8.1 and thats also the minimum required PHP version now. So I don't really see the "point" in supporting these legacy cases and they should be updated to include `full_path`. 

If it is the teams desire to keep this legacy behaviour, the logic of FileBag has to be heavily modified because it requires that every possible "file key" in $_FILES is included.

IMO fixing all of these test cases should be a separate PR which I'm willing to create after this is merged.

stof:
I think we could support it more gracefully by using `$file['full_path'] ?? $file['name']` so that passing a file without a `full_path` key still works like before.

However, it might be great to update the tests emulating file uploads to include this `full_path` key.

danielburger1337:
I had a quick look at where files are submitted manually and what tests will have to be fixed:

- PsrHttpMessage (Bridge)
- Form

It looks like these aren't that many and we can literally just copy the value `name` and set it to `full_path`.
I changed the tests I found locally and the tests now pass. **Should I add them to this PR?**

BrowserKit/DomCrawler also submits files but their logic seem to be independant.

---

I also discovered `NativeRequestHandler` in the Form component includes basically just a copy of FileBag and probably should be updated to support directory uploads as well. Add to this PR or new?

stof:
I think updating NativeRequestHandler in this PR makes sense as well.

danielburger1337:
> I think we could support it more gracefully by using `$file['full_path'] ?? $file['name']` so that passing a file without a `full_path` key still works like before.

Adding something like the following in FileBag should in theory work: 

```php
protected function convertFileInformation(array|UploadedFile $file): array|UploadedFile|null
{
    if ($file instanceof UploadedFile) {
        return $file;
    }

    if (!\array_key_exists('full_path', $file) && \array_key_exists('name', $file)) {
        $file['full_path'] = $file['name'];
    }
        
    // ...
}
```

This should then both work with "fixed file arrays and "normal" ones.

nicolas-grekas (approved):
I pushed an alternative implementation that should fix BC issues.

fabpot (approved):

fabpot:
Thank you @danielburger1337.

PR #52503, opened 2023-11-08
[DoctrineBridge][Form] Introducing new `LazyChoiceLoader` class and `choice_lazy` option for `ChoiceType`

yceruto:
| Q             | A
| ------------- | ---
| Branch?       | 7.2
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | #57724
| License       | MIT

It's quite usual to work with forms that process large datasets. In Symfony Form + Doctrine ORM, if you define an `EntityType`, it typically loads all choices/entities fully into memory, and this can lead to serious performance problems if your entity table contain several hundred or thousands of records.

The new `LazyChoiceLoader` class addresses this performance issue by implementing an on-demand choice loading strategy. This class is integrated with any `ChoiceType` subtype by using a new boolean option named `choice_lazy`, which activates the feature.

Basic usage in a Symfony form looks like this:
```php
$formBuilder->add('user', EntityType::class, [
    'class' => User::class, // a ton of users...
    'choice_lazy' => true,
]);
```

**How does it work?**

The loader operates by keeping the choice list empty until values are needed (avoiding unnecessary database queries). When form values are provided or submitted, it retrieves and caches only the necessary choices.

As you can see in the code, all this happens behind the `LazyChoiceLoader` class, which delegates the loading of choices to a wrapped `ChoiceLoaderInterface` adapter (in this case, the `DoctrineChoiceLoader`).

**Frontend Considerations**

Certainly, you may need a JavaScript component for dynamically loading `<select>` options, aka autocomplete plugins. You'll need to develop the endpoint/controller to fetch this data on your own, ensuring it corresponds to the form field data source. This aspect is not included in this project.

As a point of reference, the [Autocomplete UX Component](https://symfony.com/bundles/ux-autocomplete/current/index.html) now uses this choice loading strategy, simplifying its autocomplete form type to a single field:

**A Handy Use Case without Javascript?**

The `disabled` option renders an `EntityType` form field read-only, and when combined with the `choice_lazy` option, it prevents the loading of unnecessary entities in your choice list (only the pre-selected entities will be loaded), thereby enhancing performance.

---

Hope this helps to create simpler autocomplete components for Symfony forms.

Cheers!

### DIFF
--- a/src/Symfony/Component/Form/Tests/Extension/Core/Type/ChoiceTypeTest.php
+++ b/src/Symfony/Component/Form/Tests/Extension/Core/Type/ChoiceTypeTest.php
@@ -14,6 +14,7 @@ namespace Symfony\Component\Form\Tests\Extension\Core\Type;
 use Symfony\Component\Form\ChoiceList\Loader\CallbackChoiceLoader;
 use Symfony\Component\Form\ChoiceList\View\ChoiceGroupView;
 use Symfony\Component\Form\ChoiceList\View\ChoiceView;
+use Symfony\Component\Form\Exception\LogicException;
 use Symfony\Component\Form\Exception\TransformationFailedException;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\FormInterface;
@@ -2276,5 +2277,112 @@ class ChoiceTypeTest extends BaseTypeTestCase
         $this->assertSame('10', $view['choice_two']->vars['choices'][0]->value);
         $this->assertSame('20', $view['choice_two']->vars['choices'][1]->value);
         $this->assertSame('30', $view['choice_two']->vars['choices'][2]->value);
+    }
+
+    public function testChoiceLazyThrowsWhenChoiceLoaderIsNotSet()
+    {
+        $this->expectException(LogicException::class);
+        $this->expectExceptionMessage('The "choice_lazy" option can only be used if the "choice_loader" option is set.');
+
+        $this->factory->create(static::TESTED_TYPE, null, [
+            'choice_lazy' => true,
+        ]);
+    }
+
+    public function testChoiceLazyLoadsAndRendersNothingWhenNoDataSet()
+    {
+        $form = $this->factory->create(static::TESTED_TYPE, null, [
+            'choice_loader' => new CallbackChoiceLoader(fn () => ['a' => 'A', 'b' => 'B']),
+            'choice_lazy' => true,
+        ]);
+
+        $this->assertNull($form->getData());
+
+        $view = $form->createView();
+        $this->assertArrayHasKey('choices', $view->vars);
+        $this->assertSame([], $view->vars['choices']);
+    }
+
+    public function testChoiceLazyLoadsAndRendersOnlyDataSetViaDefault()
+    {
+        $form = $this->factory->create(static::TESTED_TYPE, 'A', [
+            'choice_loader' => new CallbackChoiceLoader(fn () => ['a' => 'A', 'b' => 'B']),
+            'choice_lazy' => true,
+        ]);
+
+        $this->assertSame('A', $form->getData());
+
+        $view = $form->createView();
+        $this->assertArrayHasKey('choices', $view->vars);
+        $this->assertCount(1, $view->vars['choices']);
+        $this->assertSame('A', $view->vars['choices'][0]->value);
+    }
+
+    public function testChoiceLazyLoadsAndRendersOnlyDataSetViaSubmit()
+    {
+        $form = $this->factory->create(static::TESTED_TYPE, null, [
+            'choice_loader' => new CallbackChoiceLoader(fn () => ['a' => 'A', 'b' => 'B']),
+            'choice_lazy' => true,
+        ]);
+
+        $form->submit('B');
+        $this->assertSame('B', $form->getData());
+
+        $view = $form->createView();
+        $this->assertArrayHasKey('choices', $view->vars);
+        $this->assertCount(1, $view->vars['choices']);
+        $this->assertSame('B', $view->vars['choices'][0]->value);
+    }
+
+    public function testChoiceLazyErrorWhenInvalidSubmitData()
+    {
+        $form = $this->factory->create(static::TESTED_TYPE, null, [
+            'choice_loader' => new CallbackChoiceLoader(fn () => ['a' => 'A', 'b' => 'B']),
+            'choice_lazy' => true,
+        ]);
+
+        $form->submit('invalid');
+        $this->assertNull($form->getData());
+
+        $view = $form->createView();
+        $this->assertArrayHasKey('choices', $view->vars);
+        $this->assertCount(0, $view->vars['choices']);
+        $this->assertCount(1, $form->getErrors());
+        $this->assertSame('ERROR: The selected choice is invalid.', trim((string) $form->getErrors()));
+    }
+
+    public function testChoiceLazyMultipleWithDefaultData()
+    {
+        $form = $this->factory->create(static::TESTED_TYPE, ['A', 'B'], [
+            'choice_loader' => new CallbackChoiceLoader(fn () => ['a' => 'A', 'b' => 'B', 'c' => 'C']),
+            'choice_lazy' => true,
+            'multiple' => true,
+        ]);
+
+        $this->assertSame(['A', 'B'], $form->getData());
+
+        $view = $form->createView();
+        $this->assertArrayHasKey('choices', $view->vars);
+        $this->assertCount(2, $view->vars['choices']);
+        $this->assertSame('A', $view->vars['choices'][0]->value);
+        $this->assertSame('B', $view->vars['choices'][1]->value);
+    }
+
+    public function testChoiceLazyMultipleWithSubmittedData()
+    {
+        $form = $this->factory->create(static::TESTED_TYPE, null, [
+            'choice_loader' => new CallbackChoiceLoader(fn () => ['a' => 'A', 'b' => 'B', 'c' => 'C']),
+            'choice_lazy' => true,
+            'multiple' => true,
+        ]);
+
+        $form->submit(['B', 'C']);
+        $this->assertSame(['B', 'C'], $form->getData());
+
+        $view = $form->createView();
+        $this->assertArrayHasKey('choices', $view->vars);
+        $this->assertCount(2, $view->vars['choices']);
+        $this->assertSame('B', $view->vars['choices'][0]->value);
+        $this->assertSame('C', $view->vars['choices'][1]->value);
     }
 }
--- a/src/Symfony/Bridge/Doctrine/Tests/Form/Type/EntityTypeTest.php
+++ b/src/Symfony/Bridge/Doctrine/Tests/Form/Type/EntityTypeTest.php
@@ -30,6 +30,7 @@ use Symfony\Bridge\Doctrine\Tests\Fixtures\SingleIntIdNoToStringEntity;
 use Symfony\Bridge\Doctrine\Tests\Fixtures\SingleStringCastableIdEntity;
 use Symfony\Bridge\Doctrine\Tests\Fixtures\SingleStringIdEntity;
 use Symfony\Component\Form\ChoiceList\LazyChoiceList;
+use Symfony\Component\Form\ChoiceList\Loader\LazyChoiceLoader;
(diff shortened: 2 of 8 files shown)


yceruto:
Over there, there should be other special use cases, such as using API payloads against forms where the choice lists are not really necessary, so people work around with custom form types and special `EntityToIdDataTransformer`. Now they can use the `EntityType` directly with the `choice_lazy` option activated.

Seb33300:
> Over there, there should be other special use cases, such as using API payloads against forms where the choice lists are not really necessary, so people work around with custom form types and special `EntityToIdDataTransformer`. Now they can use the `EntityType` directly with the `extra_lazy` option activated.

+1

I had a similar case few years ago with an API relying on a Symfony form with an EntityType loading thousands of choices.
I ended by adding the EntityType field to the form with an event on submit and filtering the SQL query to select only the items submitted by the user.

Pixelshaped:
Thank you I was already using this code from UX Autocomplete and I think it's the right move to make it part of Doctrine Bridge as this is a very common use case, even if part of the solution is front end (one would still need to implement a route to get a paginated list of entities and pass it to Select2 or whatever solution you're using).

I had duplicated the ExtraLazyChoiceLoader in my project to not require UX Autocomplete (as I'm using Select2 already everywhere), and now I can use this and stay updated.

Nice job.

stof:
I'm not sure this `extra_lazy` option in the existing type extending ChoiceType is the right option. Such autocomplete type requires a totally different rendering than a choice type (and this autocomplete rendering might make sense with data sources that are not from doctrine).

yceruto:
@stof certainly there's additional work required to create a fully-autocomplete type, which is not the goal of this PR btw. However, it does streamline implementations that previously needed custom/complex solutions. The `choice_lazy` option facilitates this for `DoctrineType` although its use is not restricted to that purpose alone.

I believe the `LazyChoiceLoader` is the main feature here. It's compatible with any `ChoiceType` and you can wrap any data source (Local Storage, External APIs, ...) allowing it to load those data efficiently by following this "on-demand" approach.

norkunas:
Something is not right with this solution. Coming from ux's autocomplete. Currently we are facing a problem that doctrine choice loader does an extra query without any criteria and tries to load 200k rows from our db before the autocomplete queries with criteria in this order:
```
SELECT ... FROM companies c0_
SELECT COUNT(*) AS dctrn_count FROM (SELECT DISTINCT id_32 FROM (SELECT ... FROM companies c0_ WHERE c0_.supplier = ? ORDER BY c0_.created_at DESC) dctrn_result) dctrn_table
SELECT DISTINCT id_32 FROM (SELECT DISTINCT id_32, created_at_30 FROM (SELECT ... FROM companies c0_ WHERE c0_.supplier = ?) dctrn_result_inner ORDER BY created_at_30 DESC) dctrn_result LIMIT 25	
SELECT ... FROM companies c0_ WHERE c0_.supplier = ? AND c0_.id IN (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?) ORDER BY c0_.created_at DESC
```

Backtrace for the first query:
```
1	Doctrine\DBAL\Connection->executeQuery (line 33)
2	Doctrine\ORM\Query\Exec\SingleSelectExecutor->execute (line 327)
3	Doctrine\ORM\Query->_doExecute (line 1212)
4	Doctrine\ORM\AbstractQuery->executeIgnoreQueryCache (line 1166)
5	Doctrine\ORM\AbstractQuery->execute (line 44)
6	Symfony\Bridge\Doctrine\Form\ChoiceList\ORMQueryBuilderLoader->getEntities (line 56)
7	Symfony\Bridge\Doctrine\Form\ChoiceList\DoctrineChoiceLoader->loadChoices (line 29)
8	Symfony\Component\Form\ChoiceList\Loader\AbstractChoiceLoader->loadChoiceList (line 59)
9	Symfony\Component\Form\ChoiceList\Loader\AbstractChoiceLoader->doLoadChoicesForValues (line 108)
10	Symfony\Bridge\Doctrine\Form\ChoiceList\DoctrineChoiceLoader->doLoadChoicesForValues (line 38)
11	Symfony\Component\Form\ChoiceList\Loader\AbstractChoiceLoader->loadChoicesForValues (line 36)
12	Symfony\Component\Form\ChoiceList\Factory\Cache\ChoiceLoader->loadChoicesForValues (line 37)
13	Symfony\UX\Autocomplete\Form\ChoiceList\Loader\ExtraLazyChoiceLoader->loadChoicesForValues (line 46)
14	Symfony\UX\Autocomplete\Form\ChoiceList\Loader\ExtraLazyChoiceLoader->loadValuesForChoices (line 81)
15	Symfony\Component\Form\ChoiceList\LazyChoiceList->getValuesForChoices (line 34)
16	Symfony\Component\Form\Extension\Core\DataTransformer\ChoiceToValueTransformer->transform (line 992)
17	Symfony\Component\Form\Form->normToView (line 301)
18	Symfony\Component\Form\Form->setData (line 412)
19	Symfony\Component\Form\Form->initialize (line 176)
20	Symfony\Component\Form\FormBuilder->getForm (line 28)
21	Symfony\Component\Form\FormFactory->create (line 142)
22	Symfony\UX\Autocomplete\Form\WrappedEntityTypeAutocompleter->getForm (line 107)
23	Symfony\UX\Autocomplete\Form\WrappedEntityTypeAutocompleter->isGranted (line 49)
24	Symfony\UX\Autocomplete\AutocompleteResultsExecutor->fetchResults (line 44)
25	Symfony\UX\Autocomplete\Controller\EntityAutocompleteController->__invoke (line 181)
26	Symfony\Component\HttpKernel\HttpKernel->handleRaw (line 76)
27	Symfony\Component\HttpKernel\HttpKernel->handle (line 197)
28	Symfony\Component\HttpKernel\Kernel->handle (line 35)
29	Symfony\Component\Runtime\Runner\Symfony\HttpKernelRunner->run (line 29)
```

Edit: okay, found the cause - autocomplete allows to configure `filter_query` which allows to query choices based on the user entered value, but then the query_builder is not configured and DoctrineChoiceLoader loads everything, so fix should be in symfony/ux

src/Symfony/Component/Form/ChoiceList/Loader/ExtraLazyChoiceLoader.php
```diff
@@ -0,0 +1,54 @@
...
+
+use Symfony\Component\Form\ChoiceList\ArrayChoiceList;
+use Symfony\Component\Form\ChoiceList\ChoiceListInterface;
+
+/**
+ * A choice loader that loads its choices and values lazily, only when necessary.
+ *
+ * @author Yonel Ceruto <yonelceruto@gmail.com>
+ */
+class ExtraLazyChoiceLoader implements ChoiceLoaderInterface
```
ro0NL: Why is it called "extra" lazy?

Can't we have laziness by default also?

yceruto: It's already lazy because there is a wrapped loader performing the main function, and the "extra" refers to the `loadChoiceList()` strategy, which caches the choices in memory but changes them on demand. I thought of this name by looking at the `extra_lazy` strategy of Doctrine associations.

Naming things is always so hard. 😅

ro0NL: i understand "extra" conveys the technical aspect, but it doesnt convey to users what it's about.

shouldnt we simply call it `autocompletes=true`

ro0NL: otherwise, i'd go with just `lazy`, since that option currently does not exists

yceruto: `lazy` works for me

yceruto:
>I'm not sure this extra_lazy option in the existing type extending ChoiceType is the right option. Such autocomplete type requires a totally different rendering than a choice type (and this autocomplete rendering might make sense with data sources that are not from doctrine).

@stof how about adding the `lazy` option in `ChoiceType` instead, and requiring the `choice_loader` to be set when `lazy` is set to `true`? So, not only the `DoctrineType` could benefit from it, but any subtype of `ChoiceType` as well

src/Symfony/Component/Form/ChoiceList/Loader/ExtraLazyChoiceLoader.php
yceruto: changed to `choice_lazy` to be more specific

yceruto:
I renamed the option to `choice_lazy` and moved it to `ChoiceType` as a generic option. Now it's part of the family of `choice_*` options.

src/Symfony/Component/Form/Tests/Extension/Core/Type/ChoiceTypeTest.php
```diff
@@ -2276,4 +2277,59 @@ public function testWithSameLoaderAndDifferentChoiceValueCallbacks()
...
+    {
+        $this->expectException(LogicException::class);
+        $this->expectExceptionMessage('The "choice_lazy" option can only be used if the "choice_loader" option is set.');
+
+        $this->factory->create(static::TESTED_TYPE, null, [
+            'choice_lazy' => true,
+        ]);
+    }
+
+    public function testChoiceLazyLoadsAndRendersNothingWhenNotDataSet()
```
xabbuh:
```suggestion
    public function testChoiceLazyLoadsAndRendersNothingWhenNoDataSet()
```

src/Symfony/Component/Form/Tests/Extension/Core/Type/ChoiceTypeTest.php
```diff
@@ -2276,4 +2277,59 @@ public function testWithSameLoaderAndDifferentChoiceValueCallbacks()
...
+
+        $this->assertSame('A', $form->getData());
+
+        $view = $form->createView();
+        $this->assertArrayHasKey('choices', $view->vars);
+        $this->assertCount(1, $view->vars['choices']);
+        $this->assertSame('A', $view->vars['choices'][0]->value);
+    }
+
+    public function testChoiceLazyLoadsAndRendersOnlyDataSetViaSubmit()
```
xabbuh, line 2321: I think we should also add a test where we submit a value that is not any of the valid choices.

yceruto: added!

yceruto:
Thanks @xabbuh for your review! I also added two test cases more about `multiple` option combined with `choice_lazy`.

xabbuh (approved):

yceruto:
Just rebased to fix conflicts.

nicolas-grekas (approved):
Maybe stupid comment: can't we always behave lazily when choice_loader is set and get rid of the option altogether?

stof:
@nicolas-grekas this lazy mode assumes that you have a form rendering that does not need the list of available choices (and so *not* rendering it as a `<select>` like the default rendering of a ChoiceType).

This is the reason why I suggested that it may fit better as a *separate* form type instead (which would have a separate rendering).

Making all choice loaders wrapped in this LazyChoiceLoader in ChoiceType itself would break it.

yceruto:
>@nicolas-grekas this lazy mode assumes that you have a form rendering that does not need the list of available choices (and so not rendering it as a `<select>` like the default rendering of a ChoiceType).

I can add that if there is data bound to that field (either default or submitted), it will render the choices as usual, rendering only the selected ones in this case.

>This is the reason why I suggested that it may fit better as a separate form type instead (which would have a separate rendering).

I might be missing your point here. Still wondering how it could be different from the current one and why... you can currently have different renderings for the same form type using `block_prefix` option.

>Making all choice loaders wrapped in this LazyChoiceLoader in ChoiceType itself would break it.

Sorry, I'm confused by this comment. Is it related to Nicolas' comment? Otherwise, could you please elaborate 🙏

yceruto:
> Maybe stupid comment: can't we always behave lazily when choice_loader is set and get rid of the option altogether?

You might still want to use a `choice_loader` even without lazy loading.

yceruto:
This is ready on my side. @stof any blocker on your side?

chalasr (approved):

nicolas-grekas:
Thank you @yceruto.

PR #52508, opened 2023-11-09
[AssetMapper] Fix jsdelivr import parsing with no imported value

### LINKED ISSUE #52467: [AssetMapper] jQuery plugin fails locally but not remotely

Assetmapper 6.4

### Description

I'm trying to load a jquery plugin on datatables.  I'm able to get it to work if the libraries are loaded remotely, but not when assetmapper copies them to the local directory.  In throws a 404 related to jquery

image

### How to reproduce

```bash
symfony new bug --webapp --version=next && cd bug
composer req symfony/asset-mapper
bin/console importmap:require datatables.net-bs5  datatables.net-select-bs5
bin/console make:controller AppController
sed -i "s|Route('/app'|Route('/'|" src/Controller/AppController.php

cat > assets/app.js <<'END'
import jquery from 'jquery'
import DataTables from 'datatables.net-bs5'
import 'datatables.net-select-bs5';
END
symfony server:start -d 
symfony open:local
```
I originally thought it was an issue with the datatables library, so I created an example script to show the problem, using the esm urls instead of the local files that assetmapper create, but I couldn't recreate the bug.

```html
    <script type="importmap">
      {
    "imports": {
        "datatables.net-bs5": "https://cdn.jsdelivr.net/npm/datatables.net-bs5@1.13.7/+esm",
        "datatables.net-select-bs5": "https://cdn.jsdelivr.net/npm/datatables.net-select-bs5@1.7.0/+esm"

    }
}
</script>
    <script type="module">
      

      import DataTable from 'datatables.net-bs5'
        import 'datatables.net-select-bs5'
        console.log('datatables loaded.');
        let t = new DataTable('#example');
        /* console.log(t.id); */

    </script>
```

See https://jsfiddle.net/tacman1123/b2f3hj08/44/

I suspect this has something to do with the 'global' jquery, webpack had a autoProvideJquery() option to help with this.  I'll also ask the library author, but since I can't provide a jsfiddle example of it not working, I'm wondering if perhaps it's related to AssetMapper.

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52467
| License       | MIT

Another new import syntax found by @tacman! I think this one is kind of a bug in the library - a module environment should not `import 'foo'` (i.e. import a module and not actually use any value), but our job is just to find these, not judge them ;).

FYI - I have an issue on jsdelivr proposing that they create an API endpoint to expose this info, so we don't need to parse it. They agree and already were thinking about this - https://github.com/jsdelivr/jsdelivr/issues/18538 - so this may be something helpful in the future. 

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
@@ -26,7 +26,7 @@ final class JsDelivrEsmResolver implements PackageResolverInterface
     public const URL_PATTERN_DIST = self::URL_PATTERN_DIST_CSS.'/+esm';
     public const URL_PATTERN_ENTRYPOINT = 'https://data.jsdelivr.com/v1/packages/npm/%s@%s/entrypoints';
 
-    public const IMPORT_REGEX = '{from"/npm/((?:@[^/]+/)?[^@]+?)(?:@([^/]+))?((?:/[^/]+)*?)/\+esm"}';
+    public const IMPORT_REGEX = '#(?:import\s*(?:(?:\{[^}]*\}|\w+|\*\s*as\s+\w+)\s*\bfrom\s*)?|export\s*(?:\{[^}]*\}|\*)\s*from\s*)("/npm/((?:@[^/]+/)?[^@]+?)(?:@([^/]+))?((?:/[^/]+)*?)/\+esm")#';
 
     private HttpClientInterface $httpClient;
 
@@ -221,9 +221,9 @@ final class JsDelivrEsmResolver implements PackageResolverInterface
         // imports from jsdelivr follow a predictable format
         preg_match_all(self::IMPORT_REGEX, $content, $matches);
         $dependencies = [];
-        foreach ($matches[1] as $index => $packageName) {
-            $version = $matches[2][$index] ?: null;
-            $packageName .= $matches[3][$index]; // add the path if any
+        foreach ($matches[2] as $index => $packageName) {
+            $version = $matches[3][$index] ?: null;
+            $packageName .= $matches[4][$index]; // add the path if any
 
             $dependencies[] = new PackageRequireOptions($packageName, $version);
         }
@@ -239,10 +239,11 @@ final class JsDelivrEsmResolver implements PackageResolverInterface
     private function makeImportsBare(string $content, array &$dependencies): string
     {
         $content = preg_replace_callback(self::IMPORT_REGEX, function ($matches) use (&$dependencies) {
-            $packageName = $matches[1].$matches[3]; // add the path if any
+            $packageName = $matches[2].$matches[4]; // add the path if any
             $dependencies[] = $packageName;
 
-            return sprintf('from"%s"', $packageName);
+            // replace the "/npm/package@version/+esm" with "package@version"
+            return str_replace($matches[1], sprintf('"%s"', $packageName), $matches[0]);
         }, $content);
 
         // source maps are not also downloaded - so remove the sourceMappingURL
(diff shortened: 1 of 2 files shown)


stof:
Actually, if the import has side effects, this is a valid usage.

tacman:
I _think_ in this case the side effect is to add the plugin to the global jquery object.

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -26,7 +26,7 @@ final class JsDelivrEsmResolver implements PackageResolverInterface
     public const URL_PATTERN_DIST = self::URL_PATTERN_DIST_CSS.'/+esm';
     public const URL_PATTERN_ENTRYPOINT = 'https://data.jsdelivr.com/v1/packages/npm/%s@%s/entrypoints';
 
-    public const IMPORT_REGEX = '{from"/npm/((?:@[^/]+/)?[^@]+?)(?:@([^/]+))?((?:/[^/]+)*?)/\+esm"}';
+    public const IMPORT_REGEX = '#(?:import|export)\s*(?:\{[^}]*\}|\w+)?\s*\**[from]*("/npm/((?:@[^/]+/)?[^@]+?)(?:@([^/]+))?((?:/[^/]+)*?)/\+esm")#';
```
stof: `[from]*` is wrong as this would match `f`, `morf` or `ffffffffffffffffffffffffffffffffffffffffffffff`. Having an optional `from` keyword is done with `(?:from)?` (or `(from)?` but this captures it).

weaverryan:
> Actually, if the import has side effects, this is a valid usage.

You're right. I guess i'm pleading into the void: "Can we please stop causing side effects with modules!!!" :)

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
stof: This regex looks also from as it would match without the `from` keyword when specifying the binding for the import as well, which is wrong.

I think the regex should look like that:

```
(?:import\s*(?:(?:\{[^}]*\}|\w+|\*\s*as\s+\w+)\s*\bfrom\s*)?|export\s*(?:\{[^}]*\}|\*)\s*from\s*)("/npm/((?:@[^/]+/)?[^@]+?)(?:@([^/]+))?((?:/[^/]+)*?)/\+esm")
```

tacman:
I'm no expert, but I also think the idea of explicitly adding the plugin to jquery is better than automatically adding it, even though it's an additional line of code.  Because modules don't allow for conditional imports, I don't see how it's possible to selectively add theme-specific plugins.

That is, if there's are select-bs4, select-bs5 and select-tailwind plugins, and merely importing them adds the plugin to jquery, there's no way to say "no, just bs5".

Better, I think would be 

```javascript
import selectBs5 from 'select-bs5';
import selectBs4 from 'select-bs4';
let plugin =  (theme=='bs4') ? selectBs4 : selectBs5;
$.addPlugin(plugin);
```

Allan is working on version 2 of datatables, as importmaps and javascript modules become more prominent perhaps he'll allow something like that.

stof:
@tacman that's something to discuss elsewhere. AssetMapper has to handle JS files as they exist.

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
weaverryan: I don't think we have the burden of catching all the edge cases, as we are guaranteed some regularity from jsdelivr. My regex incorrectly matches the invalid `import { something } "/npm/@some/package/+esm"` but your regex is smart enough to NOT match this. But jsdelivr won't serve invalid things like this to us.

Still, if we happen to make the regex smarter, I don't see any downside. Thanks for the suggestion - i've just updated the regex (all tests still green)

weaverryan: resolved thanks to your suggestion below

tacman:
Yes, it's working now, the datatables modules are working.  And fontawesome is working when I add those lines, I'm not sure if there's supposed to be a way to simply import.

Issue #44455, opened 2021-12-03
[PropertyInfo] Incorrect handling of pseudo-types


EmilMassey:
### Symfony version(s) affected

4.x, 5.x, 6.0

### Description

Since static-analysis tools like PHPStan or Psalm became more popular, it is not uncommon to see usage of pseudo-type like:
```php
/** @var non-empty-string */
public $foo;
```
Some pseudo-types are currently supported (see #43916 introducing `list` type) but others are not. We can add support for popular custom types used by PHPStan or phpDocumentor (see #44451) but there will always be less common pseudo-types which we are not aware of. 

Handling of these cases is inconsistent between extractors and `PhpDocExtractor` has a bug where the first character of the pseudo-type is trimmed. That one bug is quite easy to fix, but I believe handling of custom, unknown types is overall broken and we should decide how we want to handle them.

### How to reproduce

Tested on `v6.0.0` tag:

```php
class TestClass {
    /** @var non-empty-string */
    public $foo;

    /** @var some-other-custom-type */
    public $bar;
}

$extractor = new \Symfony\Component\PropertyInfo\Extractor\PhpDocExtractor();
$fooType = $extractor->getTypes(TestClass::class, 'foo'); // $builtinType: object, $class: on-empty-string (first character trimmed)
$barType = $extractor->getTypes(TestClass::class, 'bar'); // NULL

$extractor = new \Symfony\Component\PropertyInfo\Extractor\PhpStanExtractor();
$fooType = $extractor->getTypes(TestClass::class, 'foo'); // $builtinType: object, $class: \non-empty-string (backslash prepended)
$barType = $extractor->getTypes(TestClass::class, 'bar'); // $builtinType: object, $class: \some-other-custom-type
```

## Expected result
* pseudo-types shouldn't be automatically treated as objects
* if the type cannot be correctly mapped to built-in type, `getType()` should return `null` or throw an exception
* **or** the developer should be informed it is a pseudo-type and decide what to do with doctype
* the doctype shouldn't be modified and returned as-is (i.e. `some-other-custom-type` and not `\some-other-custom-type`)
* it would be perfect if the results were consistent between both extractors

### Additional Context

[...]

PR #52510, opened 2023-11-09
[TypeInfo] Introduce component

mtarld:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        | TODO

Introducing the brand new `TypeInfo` component

This work has been done with @Korbeil 

### Scope of the current `Symfony\Component\PropertyInfo\Type` class
Nowadays, when we need to work with types within Symfony, we have to use the `Type` class of the `PropertyInfo` component.

But what if we want to represent the type of a function's return type?

We still can use that `Type` class, but it won't make much sense as the `Symfony\Component\PropertyInfo\Type` is closely related to a property (as the namespace suggests).

Plus, when we need to extract types, we must use the `PropertyTypeExtractorInterface` service:

```php
readonly class Person
{
  public function __construct(
    public string $firstName,
  ) {
  }
}

// will return an array with a single Type object representing a string
$types = $this->propertyTypeExtractor->getTypes(Person::class, 'firstName');
```

Therefore, type retrieval in Symfony is limited to properties only.

### `Symfony\Component\PropertyInfo\Type`'s conceptual limitations

On top of that, there is a clear limitation of the current `Type` class where unions, intersections or
even generics can't be properly described.

The actual workaround is that the `PropertyTypeExtractorInterface` is returning an array of `Type`, which can be interpreted as a union type.

## The `TypeInfo` component

All these reasons bring us to create the `TypeInfo` component.

The aim here is to address these issues and:

- Have a powerful `Type` definition that can handle union, intersections, and generics (and could be even more extended)
- Being able to get types from anything, such as properties, method arguments, return types, and raw strings (and can also be extended).

### `Type` classes

To ensure a powerful `Type` definition, we defined multiple classes:
Type classes

The base `Type` class is an abstract one, so you'll always need to use one of the classes that extend it. 
Other types of classes are kinda self-explanatory.

### Type resolving

In the `TypeInfo` component, we added a `TypeResolverInterface`, and several implementations which allow developers to get a `Type` from many things:
- `ReflectionParameterTypeResolver` to resolve a function/method parameter type thanks to reflection
- `ReflectionPropertyTypeResolver` to resolve a property type thanks to reflection
- `ReflectionReturnTypeResolver` to resolve a function/method return type thanks to reflection
- `ReflectionTypeResolver` to resolve a `ReflectionNamedType`
- `StringTypeResolver` to resolve a type string representation. This can greatly work in combination with PHP documentation.
  > That resolver will only be available if the `phpstan/phpdoc-parser` package is installed.
- `ChainTypeResolver` to iterate over resolvers and to return a type as soon as a nested resolver succeeds in resolving.

### Type Creation

We also improved a lot the DX for the type creation with factories:
```php
<?php

use Symfony\Component\TypeInfo\Type;

Type::int();
Type::nullable(Type::string());
Type::generic(Type::object(Collection::class), Type::int());
Type::list(Type::bool());
Type::intersection(Type::object(\Stringable::class), Type::object(\Iterator::class));

// Many others are available and can be 
// found in Symfony\Component\TypeInfo\TypeFactoryTrait
```

### Upgrade path

This PR only introduces the `TypeInfo` component, but another one (which is already ready) will deprecate the `Symfony\Component\PropertyInfo\Type` in favor of `Symfony\Component\TypeInfo\Type`.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
@@ -44,6 +44,7 @@ use Symfony\Component\Semaphore\Semaphore;
 use Symfony\Component\Serializer\Encoder\JsonDecode;
 use Symfony\Component\Serializer\Serializer;
 use Symfony\Component\Translation\Translator;
+use Symfony\Component\TypeInfo\Type;
 use Symfony\Component\Uid\Factory\UuidFactory;
 use Symfony\Component\Validator\Validation;
 use Symfony\Component\Webhook\Controller\WebhookController;
@@ -164,6 +165,7 @@ class Configuration implements ConfigurationInterface
         $this->addAnnotationsSection($rootNode);
         $this->addSerializerSection($rootNode, $enableIfStandalone);
         $this->addPropertyAccessSection($rootNode, $willBeAvailable);
+        $this->addTypeInfoSection($rootNode, $enableIfStandalone);
         $this->addPropertyInfoSection($rootNode, $enableIfStandalone);
         $this->addCacheSection($rootNode, $willBeAvailable);
         $this->addPhpErrorsSection($rootNode);
@@ -1157,6 +1159,18 @@ class Configuration implements ConfigurationInterface
                 ->arrayNode('property_info')
                     ->info('Property info configuration')
                     ->{$enableIfStandalone('symfony/property-info', PropertyInfoExtractorInterface::class)}()
+                ->end()
+            ->end()
+        ;
+    }
+
+    private function addTypeInfoSection(ArrayNodeDefinition $rootNode, callable $enableIfStandalone): void
+    {
+        $rootNode
+            ->children()
+                ->arrayNode('type_info')
+                    ->info('Type info configuration')
+                    ->{$enableIfStandalone('symfony/type-info', Type::class)}()
                 ->end()
             ->end()
         ;
--- a/src/Symfony/Component/TypeInfo/CHANGELOG.md
+++ b/src/Symfony/Component/TypeInfo/CHANGELOG.md
@@ -1,0 +1,7 @@
+CHANGELOG
+=========
+
+7.1
+---
+
+ * Add the component
--- a/src/Symfony/Component/TypeInfo/README.md
+++ b/src/Symfony/Component/TypeInfo/README.md
@@ -1,0 +1,42 @@
+TypeInfo Component
+==================
+
+The TypeInfo component extracts PHP types information.
+
+Getting Started
+---------------
+
+```bash
+composer require symfony/type-info
+composer require phpstan/phpdoc-parser # to support raw string resolving
+```
+
+```php
+<?php
+
+use Symfony\Component\TypeInfo\Type;
+use Symfony\Component\TypeInfo\TypeResolver\TypeResolver;
+
+// Instantiate a new resolver
+$typeResolver = TypeResolver::create();
+
+// Then resolve types for any subject
+$typeResolver->resolve(new \ReflectionProperty(Dummy::class, 'id')); // returns an "int" Type instance
+$typeResolver->resolve('bool'); // returns a "bool" Type instance
+
+// Types can be instantiated thanks to static factories
+$type = Type::list(Type::nullable(Type::bool()));
+
+// Type instances have several helper methods
+$type->getBaseType() // returns an "array" Type instance
+$type->getCollectionKeyType(); // returns an "int" Type instance
+$type->getCollectionValueType()->isNullable(); // returns true
+```
+
+Resources
+---------
+
+ * [Contributing](https://symfony.com/doc/current/contributing/index.html)
+ * [Report issues](https://github.com/symfony/symfony/issues) and
+   [send Pull Requests](https://github.com/symfony/symfony/pulls)
+   in the [main Symfony repository](https://github.com/symfony/symfony)
--- a/src/Symfony/Component/TypeInfo/Tests/TypeResolver/StringTypeResolverTest.php
+++ b/src/Symfony/Component/TypeInfo/Tests/TypeResolver/StringTypeResolverTest.php
@@ -1,0 +1,190 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\TypeInfo\Tests\TypeResolver;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\TypeInfo\Exception\InvalidArgumentException;
+use Symfony\Component\TypeInfo\Exception\UnsupportedException;
+use Symfony\Component\TypeInfo\Tests\Fixtures\AbstractDummy;
+use Symfony\Component\TypeInfo\Tests\Fixtures\Dummy;
+use Symfony\Component\TypeInfo\Tests\Fixtures\DummyWithTemplates;
+use Symfony\Component\TypeInfo\Type;
+use Symfony\Component\TypeInfo\TypeContext\TypeContext;
+use Symfony\Component\TypeInfo\TypeContext\TypeContextFactory;
+use Symfony\Component\TypeInfo\TypeResolver\StringTypeResolver;
+
+class StringTypeResolverTest extends TestCase
+{
+    private StringTypeResolver $resolver;
+
+    protected function setUp(): void
+    {
+        $this->resolver = new StringTypeResolver();
+    }
+
+    /**
+     * @dataProvider resolveDataProvider
+     */
+    public function testResolve(Type $expectedType, string $string, TypeContext $typeContext = null)
+    {
+        $this->assertEquals($expectedType, $this->resolver->resolve($string, $typeContext));
+    }
+
+    /**
+     * @return iterable<array{0: Type, 1: string, 2?: TypeContext}>
+     */
+    public function resolveDataProvider(): iterable
+    {
+        $typeContextFactory = new TypeContextFactory(new StringTypeResolver());
+
+        // callable
+        yield [Type::callable(), 'callable(string, int): mixed'];
+
+        // array
+        yield [Type::array(Type::bool()), 'bool[]'];
+
+        // array shape
+        yield [Type::array(), 'array{0: true, 1: false}'];
+
+        // object shape
+        yield [Type::object(), 'object{foo: true, bar: false}'];
+
+        // this
+        yield [Type::object(Dummy::class), '$this', $typeContextFactory->createFromClassName(Dummy::class, AbstractDummy::class)];
+
+        // const
+        yield [Type::array(), 'array[1, 2, 3]'];
+        yield [Type::false(), 'false'];
+        yield [Type::float(), '1.23'];
+        yield [Type::int(), '1'];
+        yield [Type::null(), 'null'];
+        yield [Type::string(), '"string"'];
+        yield [Type::true(), 'true'];
+
+        // identifiers
+        yield [Type::bool(), 'bool'];
+        yield [Type::bool(), 'boolean'];
+        yield [Type::true(), 'true'];
+        yield [Type::false(), 'false'];
+        yield [Type::int(), 'int'];
+        yield [Type::int(), 'integer'];
+        yield [Type::int(), 'positive-int'];
+        yield [Type::int(), 'negative-int'];
+        yield [Type::int(), 'non-positive-int'];
+        yield [Type::int(), 'non-negative-int'];
+        yield [Type::int(), 'non-zero-int'];
+        yield [Type::float(), 'float'];
+        yield [Type::float(), 'double'];
+        yield [Type::string(), 'string'];
+        yield [Type::string(), 'class-string'];
+        yield [Type::string(), 'trait-string'];
+        yield [Type::string(), 'interface-string'];
+        yield [Type::string(), 'callable-string'];
+        yield [Type::string(), 'numeric-string'];
+        yield [Type::string(), 'lowercase-string'];
+        yield [Type::string(), 'non-empty-lowercase-string'];
+        yield [Type::string(), 'non-empty-string'];
+        yield [Type::string(), 'non-falsy-string'];
+        yield [Type::string(), 'truthy-string'];
+        yield [Type::string(), 'literal-string'];
+        yield [Type::string(), 'html-escaped-string'];
+        yield [Type::resource(), 'resource'];
+        yield [Type::object(), 'object'];
+        yield [Type::callable(), 'callable'];
+        yield [Type::array(), 'array'];
+        yield [Type::array(), 'non-empty-array'];
+        yield [Type::list(), 'list'];
+        yield [Type::list(), 'non-empty-list'];
+        yield [Type::iterable(), 'iterable'];
+        yield [Type::mixed(), 'mixed'];
+        yield [Type::null(), 'null'];
+        yield [Type::void(), 'void'];
+        yield [Type::never(), 'never'];
+        yield [Type::never(), 'never-return'];
+        yield [Type::never(), 'never-returns'];
+        yield [Type::never(), 'no-return'];
+        yield [Type::union(Type::int(), Type::string()), 'array-key'];
+        yield [Type::union(Type::int(), Type::float(), Type::string(), Type::bool()), 'scalar'];
+        yield [Type::union(Type::int(), Type::float()), 'number'];
+        yield [Type::union(Type::int(), Type::float(), Type::string()), 'numeric'];
+        yield [Type::object(AbstractDummy::class), 'self', $typeContextFactory->createFromClassName(Dummy::class, AbstractDummy::class)];
+        yield [Type::object(Dummy::class), 'static', $typeContextFactory->createFromClassName(Dummy::class, AbstractDummy::class)];
+        yield [Type::object(AbstractDummy::class), 'parent', $typeContextFactory->createFromClassName(Dummy::class)];
+        yield [Type::object(Dummy::class), 'Dummy', $typeContextFactory->createFromClassName(Dummy::class)];
+        yield [Type::template('T', Type::union(Type::int(), Type::string())), 'T', $typeContextFactory->createFromClassName(DummyWithTemplates::class)];
+        yield [Type::template('V'), 'V', $typeContextFactory->createFromReflection(new \ReflectionMethod(DummyWithTemplates::class, 'getPrice'))];
+
+        // nullable
+        yield [Type::nullable(Type::int()), '?int'];
+
+        // generic
+        yield [Type::generic(Type::object(), Type::string(), Type::bool()), 'object<string, bool>'];
+        yield [Type::generic(Type::object(), Type::generic(Type::string(), Type::bool())), 'object<string<bool>>'];
+        yield [Type::int(), 'int<0, 100>'];
+
+        // union
+        yield [Type::union(Type::int(), Type::string()), 'int|string'];
+
+        // intersection
+        yield [Type::intersection(Type::int(), Type::string()), 'int&string'];
+
+        // DNF
+        yield [Type::union(Type::int(), Type::intersection(Type::string(), Type::bool())), 'int|(string&bool)'];
+
+        // collection objects
+        yield [Type::collection(Type::object(\Traversable::class)), \Traversable::class];
+        yield [Type::collection(Type::object(\Traversable::class), Type::string()), \Traversable::class.'<string>'];
+        yield [Type::collection(Type::object(\Traversable::class), Type::bool(), Type::string()), \Traversable::class.'<string, bool>'];
+        yield [Type::collection(Type::object(\Iterator::class)), \Iterator::class];
+        yield [Type::collection(Type::object(\Iterator::class), Type::string()), \Iterator::class.'<string>'];
+        yield [Type::collection(Type::object(\Iterator::class), Type::bool(), Type::string()), \Iterator::class.'<string, bool>'];
+        yield [Type::collection(Type::object(\IteratorAggregate::class)), \IteratorAggregate::class];
+        yield [Type::collection(Type::object(\IteratorAggregate::class), Type::string()), \IteratorAggregate::class.'<string>'];
+        yield [Type::collection(Type::object(\IteratorAggregate::class), Type::bool(), Type::string()), \IteratorAggregate::class.'<string, bool>'];
+    }
+
+    public function testCannotResolveNonStringType()
+    {
+        $this->expectException(UnsupportedException::class);
+        $this->resolver->resolve(123);
+    }
+
+    public function testCannotResolveThisWithoutTypeContext()
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->resolver->resolve('$this');
+    }
+
+    public function testCannotResolveSelfWithoutTypeContext()
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->resolver->resolve('self');
+    }
+
+    public function testCannotResolveStaticWithoutTypeContext()
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->resolver->resolve('static');
+    }
+
+    public function testCannotResolveParentWithoutTypeContext()
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->resolver->resolve('parent');
+    }
+
+    public function testCannotUnknownIdentifier()
+    {
+        $this->expectException(UnsupportedException::class);
+        $this->resolver->resolve('unknown');
+    }
+}
--- a/src/Symfony/Component/TypeInfo/Tests/TypeTest.php
+++ b/src/Symfony/Component/TypeInfo/Tests/TypeTest.php
@@ -1,0 +1,74 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\TypeInfo\Tests;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\TypeInfo\Exception\LogicException;
+use Symfony\Component\TypeInfo\Type;
+use Symfony\Component\TypeInfo\TypeIdentifier;
+
+class TypeTest extends TestCase
+{
+    public function testIs()
+    {
+        $isInt = fn (Type $t) => TypeIdentifier::INT === $t->getBaseType()->getTypeIdentifier();
+
+        $this->assertTrue(Type::int()->is($isInt));
+        $this->assertTrue(Type::union(Type::string(), Type::int())->is($isInt));
+        $this->assertTrue(Type::generic(Type::int(), Type::string())->is($isInt));
+
+        $this->assertFalse(Type::string()->is($isInt));
+        $this->assertFalse(Type::union(Type::string(), Type::float())->is($isInt));
+        $this->assertFalse(Type::generic(Type::string(), Type::int())->is($isInt));
+    }
+
+    public function testIsA()
+    {
+        $this->assertTrue(Type::int()->isA(TypeIdentifier::INT));
+        $this->assertTrue(Type::union(Type::string(), Type::int())->isA(TypeIdentifier::INT));
+        $this->assertTrue(Type::generic(Type::int(), Type::string())->isA(TypeIdentifier::INT));
+
+        $this->assertFalse(Type::string()->isA(TypeIdentifier::INT));
+        $this->assertFalse(Type::union(Type::string(), Type::float())->isA(TypeIdentifier::INT));
+        $this->assertFalse(Type::generic(Type::string(), Type::int())->isA(TypeIdentifier::INT));
+    }
+
+    public function testIsNullable()
+    {
+        $this->assertTrue(Type::null()->isNullable());
+        $this->assertTrue(Type::mixed()->isNullable());
+        $this->assertTrue(Type::nullable(Type::int())->isNullable());
+        $this->assertTrue(Type::union(Type::int(), Type::null())->isNullable());
+        $this->assertTrue(Type::union(Type::int(), Type::mixed())->isNullable());
+        $this->assertTrue(Type::generic(Type::null(), Type::string())->isNullable());
+
+        $this->assertFalse(Type::int()->isNullable());
+        $this->assertFalse(Type::union(Type::int(), Type::string())->isNullable());
+        $this->assertFalse(Type::generic(Type::int(), Type::nullable(Type::string()))->isNullable());
+        $this->assertFalse(Type::generic(Type::int(), Type::mixed())->isNullable());
+    }
+
+    public function testGetBaseType()
+    {
+        $this->assertEquals(Type::string(), Type::string()->getBaseType());
+        $this->assertEquals(Type::object(self::class), Type::object(self::class)->getBaseType());
+        $this->assertEquals(Type::object(), Type::generic(Type::object(), Type::int())->getBaseType());
+        $this->assertEquals(Type::builtin(TypeIdentifier::ARRAY), Type::list()->getBaseType());
+        $this->assertEquals(Type::int(), Type::collection(Type::generic(Type::int(), Type::string()))->getBaseType());
+    }
+
+    public function testCannotGetBaseTypeOnCompoundType()
+    {
+        $this->expectException(LogicException::class);
+        Type::union(Type::int(), Type::string())->getBaseType();
+    }
+}
--- a/src/Symfony/Component/TypeInfo/Type.php
+++ b/src/Symfony/Component/TypeInfo/Type.php
@@ -1,0 +1,89 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\TypeInfo;
+
+use Symfony\Component\TypeInfo\Exception\LogicException;
+use Symfony\Component\TypeInfo\Type\BuiltinType;
+use Symfony\Component\TypeInfo\Type\CollectionType;
+use Symfony\Component\TypeInfo\Type\GenericType;
+use Symfony\Component\TypeInfo\Type\IntersectionType;
+use Symfony\Component\TypeInfo\Type\ObjectType;
+use Symfony\Component\TypeInfo\Type\UnionType;
+
+/**
+ * @author Mathias Arlaud <mathias.arlaud@gmail.com>
+ * @author Baptiste Leduc <baptiste.leduc@gmail.com>
+ */
+abstract class Type implements \Stringable
+{
+    use TypeFactoryTrait;
+
+    public function getBaseType(): BuiltinType|ObjectType
+    {
+        if ($this instanceof UnionType || $this instanceof IntersectionType) {
+            throw new LogicException(sprintf('Cannot get base type on "%s" compound type.', (string) $this));
+        }
+
+        $baseType = $this;
+
+        if ($baseType instanceof CollectionType) {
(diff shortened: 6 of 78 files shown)


src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1148,6 +1150,18 @@ private function addPropertyInfoSection(ArrayNodeDefinition $rootNode, callable
         ;
     }
 
+    private function addTypeInfoSection(ArrayNodeDefinition $rootNode, callable $enableIfStandalone): void
+    {
+        $rootNode
+            ->children()
+                ->arrayNode('type_info')
+                    ->info('Type info configuration')
+                    ->{$enableIfStandalone('symfony/type-info', Type::class)}()
```
GromNaN, line 1173: This should be in composer `require-dev`.

src/Symfony/Component/TypeInfo/BuiltinType.php
```diff
@@ -0,0 +1,41 @@
...
+    case NULL = 'null';
+    case OBJECT = 'object';
+    case RESOURCE = 'resource';
+    case STRING = 'string';
+    case TRUE = 'true';
+
+    /**
+     * @return list<string>
+     */
+    public static function values(): array
```
GromNaN: This method doesn't seem to be used.

src/Symfony/Component/TypeInfo/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+CHANGELOG
+=========
+
+7.1
+---
+
+ * Added component
```
OskarStark:
```suggestion
 * Add the component
```

WebMamba (commented):
I like this! I know that some part of the LiveComponent can be refactor to use this component. We faced some limitation with the PropertyInfo on UnionType for exemple, and it look like this new Component can solve those limitations 😁

dunglas:
As the original author of PropertyInfo, I agree that we need a new, more powerful, TypeInfo component. PHP evolved a lot on this topic since the original release of PropertyInfo, and PropertyInfo is now too limited.

One of my regrets with `PropertyInfo\Type` is that it didn't try to be in sync with the features introduced in PHP directly.
In my opinion, TypeInfo should be a kind of "polyfill" for new and upcoming features introduced to PHP own's [`ReflectionType`](https://www.php.net/manual/en/class.reflectiontype.php) class and should inherit from it.

When/if new features (e.g.: generics) are added to PHP directly, then TypeInfo should deprecate its own equivalent feature and use the builtin one. WDYT?

stof:
Extending the PHP classes would mean that we cannot really deprecate things in our API as it would be PHP itself doing changes as well. And it might also create confusion if we create our own subclasses of `ReflectionType` as most code expecting `ReflectionType`  will support only the classes defined in PHP.
Thus, we cannot "polyfill" upcoming features as we have no idea how that feature will define its PHP API in the future.

src/Symfony/Component/TypeInfo/Type/EnumType.php
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Symfony\Component\TypeInfo\Type;
+
+use Symfony\Component\TypeInfo\Exception\InvalidArgumentException;
+
+/**
+ * @author Mathias Arlaud <mathias.arlaud@gmail.com>
+ * @author Baptiste Leduc <baptiste.leduc@gmail.com>
+ */
+class EnumType extends ObjectType
```
helyakin, line 22: Question: Maybe I'm mistaken but have you forgotten the final keyword here ?
```suggestion
final class EnumType extends ObjectType
```

src/Symfony/Component/TypeInfo/Type/ObjectType.php
```diff
@@ -0,0 +1,51 @@
...
+namespace Symfony\Component\TypeInfo\Type;
+
+use Symfony\Component\TypeInfo\BuiltinType;
+use Symfony\Component\TypeInfo\Type;
+
+/**
+ * @author Mathias Arlaud <mathias.arlaud@gmail.com>
+ * @author Baptiste Leduc <baptiste.leduc@gmail.com>
+ */
+class ObjectType extends Type
```
helyakin, line 23: final ?

src/Symfony/Component/TypeInfo/Type/EnumType.php
Korbeil: Since `BackedEnumType` extends `EnumType` we cannot make it final :wink: 

Quick example: https://3v4l.org/lH4KG

src/Symfony/Component/TypeInfo/Type/ObjectType.php
Korbeil: Same as above, we are extending `ObjectType` in `EnumType` class

dunglas:
@stof we can implement the methods with the names we want in our subclass and deprecate them in favor of the official ones when they will be added to PHP.

mtarld:
@dunglas, if I understood well, you'd love to have a type class hierarchy like the following:

```mermaid
classDiagram
ReflectionNamedType <|-- Type
ReflectionUnionType <|-- UnionType
ReflectionIntersectionType <|-- IntersectionType
Type <|-- BuiltinType
Type <|-- ObjectType
Type <|-- TemplateType
Type <|-- CollectionType
Type <|-- GenericType
Type <|-- UnionType
Type <|-- IntersectionType
ObjectType <|-- EnumType
EnumType <|-- BackedEnumType
```

By doing that way, `UnionType` and `IntersectionType` inherit from the `Type` class and, as well, from the `ReflectionUnionType` and the `ReflectionIntersectionType`. And of course, it's not possible in PHP to inherit from two classes at the same type.

Plus, the `UnionType` and `IntersectionType` will be considered as a `ReflectionNamedType` which can be weird.

So I see two solutions here:
1. Do not inherit from `ReflectionUnionType` neither `ReflectionIntersectionType`, and make the `Type` inherit from the `ReflectionType`:
```mermaid
classDiagram
ReflectionType <|-- Type
Type <|-- BuiltinType
Type <|-- ObjectType
Type <|-- TemplateType
Type <|-- CollectionType
Type <|-- GenericType
Type <|-- UnionType
Type <|-- IntersectionType
ObjectType <|-- EnumType
EnumType <|-- BackedEnumType
```
But we won't be that precise, and we'll miss out if something is added into `ReflectionNamedType`, `ReflectionUnionType`, or `ReflectionIntersectionType`.

2. Turn the `Type` class into an interface and make types inherit explicitly from the proper `ReflectionType` (`ReflectionNamedType`, `ReflectionUnionType`, and `ReflectionIntersectionType`)
```mermaid
classDiagram
Type <|-- BuiltinType
ReflectionNamedType <|-- BuiltinType
Type <|-- ObjectType
ReflectionNamedType <|-- ObjectType
Type <|-- TemplateType
ReflectionNamedType <|-- TemplateType
Type <|-- CollectionType
ReflectionNamedType <|-- CollectionType
Type <|-- GenericType
ReflectionNamedType <|-- GenericType
Type <|-- UnionType
ReflectionUnionType <|-- UnionType
Type <|-- IntersectionType
ReflectionIntersectionType <|-- IntersectionType
ObjectType <|-- EnumType
EnumType <|-- BackedEnumType
<<interface>> Type
```
But we have no way to enforce the `Type` interface to be implemented by a `ReflectionType`

WDYT?

Korbeil:
About using Reflection and make the types classes extend it I'm kinda against it since it will only make the component harder to maintain with no real DX improvement. All the stuff we added are there to make types easier to manage and current Reflection isn't that easy to use.

src/Symfony/Component/TypeInfo/Type.php
```diff
@@ -0,0 +1,83 @@
...
+
+        return $callable($this);
+    }
+
+    public function isBuiltinType(BuiltinTypeEnum $builtinType): bool
+    {
+        return $this->is(static function (self $t) use ($builtinType): bool {
+            try {
+                $b = $t->getBaseType();
+            } catch (\LogicException) {
```
ihmels: The `Symfony\Component\TypeInfo\Exception\LogicException` must be caught here:

```suggestion
            } catch (LogicException) {
```

dunglas:
@Korbeil Reflection is a low level API used only for advanced use cases. Type introspection as well. App developers (unlike framework and library authors) should not have to use them directly. The DX matters less than performance and maintainability for these use cases. And best perf and maintainability is always achieved by relying only on the standard library without needing extra code. Currently, the PHP stdlib doesn't expose all the features we need, but the history of PropertyInfo proved that features are being added, and it's an open goal for the PHP project to support generics and the like. In the future, this component will hopefully not be useful anymore and will be deprecated (same for PropertyInfo), which is good. IMHO TypeInfo and PropertyInfo should try to reduce the amount of work that maintainers will have to do to switch to the PHP stdlib when generics and similar features will be supported.

Korbeil:
@dunglas I don't think there is much work needed out of resolvers (which was the same in PropertyInfo extractor) for the maintainers. And if PHP starts to have a better Reflection for types will be welcome but we don't have that yet. Also about performance, extending Reflection would mean having to call Reflection everytime we use a Type class, this is clearly not something that should be done, we want this component to be the most performant possible.

src/Symfony/Component/TypeInfo/BuiltinType.php
```diff
@@ -0,0 +1,33 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\TypeInfo;
+
+/**
+ * @author Mathias Arlaud <mathias.arlaud@gmail.com>
+ * @author Baptiste Leduc <baptiste.leduc@gmail.com>
+ */
+enum BuiltinType: string
```
rosier:
```suggestion
enum BuiltinTypeEnum: string
```

src/Symfony/Component/TypeInfo/Type.php
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        if ($this instanceof IntersectionType) {
+            return $this->everyTypeIs($callable);
+        }
+
+        return $callable($this);
+    }
+
+    public function isBuiltinType(BuiltinTypeEnum $builtinType): bool
```
mathroc: I'm not sure `is()` is the right name, I think `accepts()` or `includes()` might be for what the function does

and `acceptsBuiltinType()` or `includesBuiltinType()`.

`isNullable()` just below is fine however

mathroc:
Hi! nice work there :)

2 questions:

* Do you think it could be useful to use generics on the classes defined here?
* is it scope to include even more stuff found in Psalm/PHPStan? such as `int<0,max>`, `value-of<MyEnum>`, etc. ?

src/Symfony/Component/TypeInfo/BuiltinType.php
Korbeil: I see no point justifying this, I would prefer the enum to stay like it is.

src/Symfony/Component/TypeInfo/Type/GenericType.php
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Symfony\Component\TypeInfo\Type;
+
+use Symfony\Component\TypeInfo\Type;
+
+/**
+ * @author Mathias Arlaud <mathias.arlaud@gmail.com>
+ * @author Baptiste Leduc <baptiste.leduc@gmail.com>
+ */
+final class GenericType extends Type
```
chalasr, line 27: It's not obvious what it is about without reading the full diff. At least this one ~~and CollectionType~~ deserve some documentation I think.

src/Symfony/Component/TypeInfo/BuiltinType.php
chalasr: I see one: DX as there are 2 classes named `BuiltinType` currently which is painful when looking for one via IDE' search feature. Also you're aliasing this to `BuiltinTypeEnum` in `BuiltinType` (the other, here we are :)) which makes that part of the patch harder to follow.

src/Symfony/Component/TypeInfo/Type/BuiltinType.php
```diff
@@ -0,0 +1,39 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\TypeInfo\Type;
+
+use Symfony\Component\TypeInfo\BuiltinType as BuiltinTypeEnum;
+use Symfony\Component\TypeInfo\Type;
+
+/**
+ * Represents a PHP type.
```
chalasr: this states the obvious, I'd remove it

src/Symfony/Component/TypeInfo/Type/EnumType.php
```diff
@@ -0,0 +1,35 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\TypeInfo\Type;
+
+use Symfony\Component\TypeInfo\Exception\InvalidArgumentException;
+
+/**
+ * Represents a PHP Enum.
```
chalasr: this is also stating the obvious

src/Symfony/Component/TypeInfo/Type/IntersectionType.php
```diff
@@ -0,0 +1,43 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\TypeInfo\Type;
+
+use Symfony\Component\TypeInfo\Type;
+
+/**
+ * Represent an intersection type. It contains a list of types.
```
chalasr: seems useless as well

src/Symfony/Component/TypeInfo/Type/ObjectType.php
```diff
@@ -0,0 +1,53 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\TypeInfo\Type;
+
+use Symfony\Component\TypeInfo\BuiltinType;
+use Symfony\Component\TypeInfo\Type;
+
+/**
+ * Represent an object type.
```
chalasr: useless for sure

src/Symfony/Component/TypeInfo/Type/UnionType.php
```diff
@@ -0,0 +1,51 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\TypeInfo\Type;
+
+use Symfony\Component\TypeInfo\BuiltinType as BuiltinTypeEnum;
+use Symfony\Component\TypeInfo\Type;
+
+/**
+ * Represent an union type. It contains a list of types.
```
chalasr: same

chalasr (commented):
Some cosmetic comments

src/Symfony/Component/TypeInfo/BuiltinType.php
mtarld: I can imagine that users will most of the time do the following for creating builtin types:
```php
use Symfony\Component\TypeInfo\Type;
use Symfony\Component\TypeInfo\BuiltinType;

Type::builtin(BuiltinType::INT);
// Type::int() will of course be the best way
```

instead of:
```php
use Symfony\Component\TypeInfo\Type\BuiltinType;
use Symfony\Component\TypeInfo\BuiltinType as BuiltinTypeEnum;

new BuiltinType(BuiltinTypeEnum::INT);
```

Therefore, if we add the `Enum` suffix, i'll be a little more verbose:
```php
use Symfony\Component\TypeInfo\Type;
use Symfony\Component\TypeInfo\BuiltinTypeEnum;

Type::builtin(BuiltinTypeEnum::INT);
```
But this is really negligible.

My main argument to keep it without the suffix is that we don't have any `Enum` suffix in the codebase actually, but I'm not against adding it.

What's your POV?

rosier: Thx for the use case examples. I now think it's oke to keep it without the suffix.

Looking at https://www.php.net/manual/en/language.types.type-system.php#language.types.type-system.atomic
The enum includes more then the built-in types maybe another option is to name it `BaseType`

```php
Type::builtin(BaseType::INT);
```

mtarld: Fair point, we do have either built-in, value, and aliases.

I'm not really for `BaseXXX` because it makes me think about inheritance.

PHPStan seems to document it as [basic types](https://phpstan.org/writing-php-code/phpdoc-types#basic-types), and to name it `IdentifierType` in the codebase.

`TypeIdentifier` is a good candidate for me. Indeed, as it doesn't end with "Type", it cannot be confused with other types.

mtarld: `TypeName` could be a great candidate too

src/Symfony/Component/TypeInfo/TypeResolver/TypeResolverInterface.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+interface TypeResolverInterface
+{
+    /**
+     * Try to resolve a {@see Type} on a $subject.
+     * If the resolver cannot resolve the type, it will throw a {@see UnsupportedException}.
+     *
+     * @throws UnsupportedException
+     */
+    public function resolve(mixed $subject, TypeContext $typeContext = null): Type;
```
chalasr, line 32: Some hint about what's expected to be passed as `$subject` would be useful. 
Maybe typehint on `\Reflector` also? (just wondering)

Korbeil: I had set this as `mixed` first to avoid having BC when handling new stuff, but it should be safe to put it as `string|\Reflector` only. @mtarld, your opinion ?

src/Symfony/Component/TypeInfo/README.md
```diff
@@ -0,0 +1,14 @@
+TypeInfo Component
+==================
+
+The TypeInfo component extracts information about PHP types using
+metadata of popular sources.
```
chalasr: We're moving docs for standalone usage from symfony-docs to each component's README, it'd be nice to have some snippet here too. Check [password-hasher](https://github.com/symfony/password-hasher?tab=readme-ov-file#getting-started) or [error-handler](https://github.com/symfony/error-handler?tab=readme-ov-file#getting-started) for inspiration

src/Symfony/Component/TypeInfo/TypeResolver/TypeResolverInterface.php
mtarld: In my opinion, subject could be literally anything, that's why the mixed makes sense to me.

Maybe we can add some documentation to make it more understandable?

chalasr: Yes, that's the part of my proposal I'm sure about :) Also I'd suggest to narrow `$subect`'s type in each implementation via `@param` phpdoc since they're are all tight to a single type. Perhaps mention that null is not allowed in the interface also (if correct ofc)

src/Symfony/Component/TypeInfo/composer.json
```diff
@@ -0,0 +1,42 @@
...
+            "homepage": "https://symfony.com/contributors"
+        }
+    ],
+    "require": {
+        "php": ">=8.2"
+    },
+    "require-dev": {
+        "symfony/dependency-injection": "^7.1",
+        "phpdocumentor/type-resolver": "^1.7",
+        "phpstan/phpdoc-parser": "^1.0"
```
ruudk: Should these be added to `suggests`? Or am I missing something.

src/Symfony/Component/TypeInfo/TypeResolver/StringTypeResolver.php
```diff
@@ -0,0 +1,251 @@
...
+
+    public function __construct()
+    {
+        $this->lexer = new Lexer();
+        $this->parser = new TypeParser(new ConstExprParser());
+    }
+
+    public function resolve(mixed $subject, TypeContext $typeContext = null): Type
+    {
+        if (!class_exists(TypeParser::class)) {
```
chalasr: this check cannot be reached as it would break earlier at construct l66

src/Symfony/Component/TypeInfo/TypeResolver/ChainTypeResolver.php
```diff
@@ -0,0 +1,43 @@
...
+ * @author Mathias Arlaud <mathias.arlaud@gmail.com>
+ * @author Baptiste Leduc <baptiste.leduc@gmail.com>
+ */
+final readonly class ChainTypeResolver implements TypeResolverInterface
+{
+    /**
+     * @param iterable<TypeResolverInterface> $typeResolvers
+     */
+    public function __construct(
+        private iterable $typeResolvers,
```
chalasr: I'd like to challenge the priority-based resolver list here as I'm wondering if it's really appropriate.
Couldn't this parameter be given a map of resolvers keyed by "resolver type" instead? e.g.
```php
ServiceLocator([
    \ReflectionParameter::class => fn () => $this->container->get('type_info.resolver.reflection_parameter'),
    \ReflectionProperty::class => fn () => $this->container->get('type_info.resolver.reflection_property'),
    'string' => fn () => ....,
    // ...
])
```
And this "entrypoint" service be responsible for finding the right resolver based on the `$subject instanceof X`  dance or `Container::has(\is_string($subject) ? 'string' : $subject::class)`?

chalasr (changes requested):

src/Symfony/Component/TypeInfo/composer.json
chalasr: We don't add composer' `suggests` anymore as it appears to not be that helpful, mostly noise that very few people read

src/Symfony/Component/TypeInfo/TypeResolver/ChainTypeResolver.php
mtarld: I really like the idea! It'll be a bit less extensible, but I don't think that point is the important one. By doing that, it'll be way more predictable.

I'd create an "entrypoint" service so that we'll be able to add custom logic at some point if it's necessary.

src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ CHANGELOG
  * Move the Router `cache_dir` to `kernel.build_dir`
  * Deprecate the `router.cache_dir` config option
  * Add `rate_limiter` tags to rate limiter services
+ * Register TypeInfo services
```
fabpot: I don't think we've ever documented the same for other components, can probably be removed.

src/Symfony/Component/TypeInfo/README.md
```diff
@@ -0,0 +1,42 @@
+TypeInfo Component
+==================
+
+The TypeInfo component extracts information about PHP types.
```
fabpot:
```suggestion
The TypeInfo component extracts PHP types information.
```

src/Symfony/Component/TypeInfo/Type/EnumType.php
fabpot: But you can add `@final`, right?

Korbeil: Since the class is already extended, won't it make the final keyword kinda non-sense here ?

chalasr: Even if both classes are in the same vendor,`DebugClassLoader` would trigger a deprecation notice so we cannot make it `@final`

root-aza:
Inventing new bicycles :)))
https://github.com/typhoon-php/type
https://github.com/typhoon-php/reflection

Many more types and support for statistical analysis types.


@vudaltsov cc

chalasr:
Cool, thanks for sharing. 
Good luck with the PSR also! I don't know if it's something to standardize but if it happens, keep in mind it's about framework interop so don't miss taking Symfony's bicycle into account :)

chalasr (approved):
Nice work.

src/Symfony/Component/TypeInfo/TypeResolver/StringTypeResolver.php
```diff
@@ -0,0 +1,248 @@
...
+            new \ReflectionClass($classNameOrTemplate);
+            self::$classExistCache[$classNameOrTemplate] = true;
+
+            return Type::object($classNameOrTemplate);
+        } catch (\Throwable) {
+        }
+
+        self::$classExistCache[$classNameOrTemplate] = false;
+
+        return Type::template($classNameOrTemplate);
```
EmilMassey: For me personally, this is a bit unexpected to fallback to `TemplateType` for every identifier that isn't a known class and is not a pseudo-type already supported by the component.

Let's assume some static analysis / documentation tool introduces a new pseudo-type, e.g. `odd-number` and someone uses it in their code. Until support for this type is added to the component, the type for the node described with this new pseudo-type is resolved to `TemplateType` which is not true. Similar case is when someone makes a typo when typing a class name, e.g. `@param stdClas $foo` - it will be reported as `TemplateType`.

I don't know if there's a way to detect if the identifier is potentially a pseudo-class or not, but maybe we shouldn't assume it's a template if we don't know what it really is.

See related issue #44455

Korbeil: We should make a map with all pseudo-types and map them to a native type.
I would say this could be a nice first PR once this is merged since current component has the same features we currently have in Symfony.

mtarld: Actually, I think doing the following:

- First gather as much templates as possible with the TypeContextFactory (class templates are actually gathered but I'll gather method templates as well).
- Then, in the StringTypeResolver, if we find a template in the TypeContext, we return a TemplateType, and if not, we throw an exception.

WDYT?

EmilMassey: @Korbeil 
It won't solve the issue I described because it's impossible to know **all** pseudo-types that exist. The ones that are already resolved correctly are used by the most popular tools like phpstan, but there may be some less popular libraries that we don't know of. Nevertheless having such a map would be nice from code readability perspective probably but it's not crucial.

@mtarld 
That sounds much better in my opinion. I think throwing an exception is a safest result in this case.

mtarld: Here you go: https://github.com/symfony/symfony/pull/52510/files#diff-82935a7583645f83e164c76862126f68307e568b869f677980270a317e37f120R247-R251 :slightly_smiling_face: 

Thanks for that nice catch!

kbond (commented):
This is really nice!

Any thoughts on adding the following to the api?

```php
/** @var Type $type1 **/
/** @var Type $type2 **/

$type1->supports($type2); // bool

// and

$type1->accepts('some value'); // bool
```

_Or maybe I'm missing the way to achieve currently._

src/Symfony/Component/TypeInfo/Type/UnionType.php
```diff
@@ -0,0 +1,49 @@
...
+final class UnionType extends Type
+{
+    /**
+     * @use CompositeTypeTrait<T>
+     */
+    use CompositeTypeTrait;
+
+    public function asNonNullable(): Type
+    {
+        $types = array_values(array_filter($this->getTypes(), fn (Type $t): bool => !$t->isA(TypeIdentifier::NULL)));
```
EmilMassey: Here's some kind of inconsistency given than `Type::isNullable()` is currently true for `mixed`, and `asNonNullable()` filters out only `null` type and keeps `mixed`.

This results in behavior that I'm not sure is expected:
```php
$type = new UnionType(Type::int(), Type::mixed());
$type->asNonNullable()->isNullable(); // true which is kinda contradictory 
```

Maybe `Type::isNullable()` shouldn't return true for `mixed` in the first place? But if not true, then what? This is probably one of the reasons why phpstan uses [trinary logic](https://phpstan.org/developing-extensions/trinary-logic) in their type system.

src/Symfony/Component/TypeInfo/TypeResolver/StringTypeResolver.php
```diff
@@ -0,0 +1,253 @@
...
+                'iterable' => Type::iterable(),
+                'mixed' => Type::mixed(),
+                'null' => Type::null(),
+                'array-key' => Type::union(Type::int(), Type::string()),
+                'scalar' => Type::union(Type::int(), Type::float(), Type::string(), Type::bool()),
+                'number' => Type::union(Type::int(), Type::float()),
+                'numeric' => Type::union(Type::int(), Type::float(), Type::string()),
+                'self' => Type::object($typeContext->getDeclaringClass()),
+                'static' => Type::object($typeContext->getCalledClass()),
+                'parent' => Type::object($typeContext->getParentClass()),
```
EmilMassey: There's no check if `$typeContext` is not null, so this would be an error:

```php
$resolver->resolve('self'); // Error : Call to a member function getDeclaringClass() on null
$resolver->resolve('static'); // Error : Call to a member function getCalledClass() on null
$resolver->resolve('parent'); // Error : Call to a member function getParentClass() on null
```

src/Symfony/Component/TypeInfo/Tests/TypeTest.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        $this->assertTrue(Type::null()->isNullable());
+        $this->assertTrue(Type::mixed()->isNullable());
+        $this->assertTrue(Type::nullable(Type::int())->isNullable());
+        $this->assertTrue(Type::union(Type::int(), Type::null())->isNullable());
+        $this->assertTrue(Type::union(Type::int(), Type::mixed())->isNullable());
+        $this->assertTrue(Type::generic(Type::mixed(), Type::string())->isNullable());
+
+        $this->assertFalse(Type::int()->isNullable());
+        $this->assertFalse(Type::int()->isNullable());
```
EmilMassey: This line is duplicated

src/Symfony/Component/TypeInfo/Tests/TypeResolver/StringTypeResolverTest.php
```diff
@@ -0,0 +1,172 @@
...
+     */
+    public function resolveDataProvider(): iterable
+    {
+        $typeContextFactory = new TypeContextFactory(new StringTypeResolver());
+
+        // callable
+        yield [Type::callable(), 'callable(string, int): mixed'];
+
+        // array
+        yield [Type::list(Type::bool()), 'bool[]'];
```
EmilMassey: I think `bool[]` should be resolved to array, because it's not guaranteed that it's a list. 
https://phpstan.org/r/3f13252d-da15-4d6c-b140-c2a611b65cf3

Or maybe this should even be an iterable: https://phpstan.org/r/b22e9edc-2487-4448-a297-3aedda9c2625

src/Symfony/Component/TypeInfo/Type/UnionType.php
mtarld: Indeed, `mixed` is a special type... As stated in the documentation, it says that "It is equivalent to the union type object|resource|array|string|float|int|bool|null".

Therefore, instead of having a `mixed` built-in type, the static factory could convert it to the proper union type, so that there is no ambiguity anymore. WDYT?

src/Symfony/Component/TypeInfo/Tests/TypeResolver/StringTypeResolverTest.php
mtarld: Totally agreed, I resolved it as an array then.

src/Symfony/Component/TypeInfo/Type/UnionType.php
mtarld: By doing that, it gives the following:
```php
$type = Type::union(Type::int(), Type::mixed());
(string) $type; // array|bool|float|int|null|object|resource|string
$type->isNullable(); // true

$type = $type->asNonNullable();
(string) $type; // array|bool|float|int|object|resource|string
$type->isNullable(); // false
```

Korbeil:
@kbond thanks for your feedback !

Actually `accepts` and `supports` are really good ideas but we think it should be done in a future PR.
For example for `supports`, if `$type1` and `$type2` are two objects and the first inherits the second.
Then `supports` should return `true` but we don't have the required metadata to do so at the moment.
I would prefer to store anything required during runtime into the `Type` object (or children classes) so it requires some changes (nothing that will break BC so should be all fine :smile:)

src/Symfony/Component/TypeInfo/Type/UnionType.php
EmilMassey: It's an interesting idea, but I don't know if it will work well as a project-wide solution. I can't say how the component is going to be used by developers, but perhaps there are use cases where we would want to know exactly whether the type is written as `mixed` or as union. Without the `mixed` built-in type, we will lose this information if I understand your idea correctly. 

But as for that narrow scope of the `asNonNullable` method, it seems to me that this conversion and then stripping `null` would make sense. Again, I can't tell right now how this method is going to be used so I'm curious what others think about this, perhaps their perspective is different.

mtarld: I can see one unexpected behavior with that approach: `(string) $type` will give `array|bool|float|int|null|object|resource|string` instead of the expected `mixed`.

Therefore, I introduced again the `mixed` built-in type and convert it to its related union type only when really needed, which is when `asNonNullable` is involved.

chalasr:
👍 To polish and make the component more feature-rich later, we even have time to do so before it gets released. The diff is already quite big for what's really needed in other ongoing work.

EmilMassey (approved):
Excellent work, guys! This component is going to be far more pleasant to work with than PropertyInfo.

mtarld:
Many thanks for your accurate reviews @EmilMassey!

src/Symfony/Component/TypeInfo/Type/UnionType.php
mathroc: don't know if it's easier but you can also have it expanded to `array|bool|float|int|object|resource|string` inside, and transforming that to `mixed` when printing its name if it's an union matching that, or `mixed|A|B|C` if there's more

to me it feels like it would make the internal logic easier to deal with in the future if it doesn't have to care about that special case

kbond:
> Actually accepts and supports are really good ideas but we think it should be done in a future PR.

Absolutely, I just wanted to verify the idea!

src/Symfony/Component/TypeInfo/Type.php
```diff
@@ -0,0 +1,93 @@
...
+    {
+        if ($this instanceof UnionType) {
+            return $this->atLeastOneTypeIs($callable);
+        }
+
+        if ($this instanceof IntersectionType) {
+            return $this->everyTypeIs($callable);
+        }
+
+        return $callable($this);
```
welcoMattic:
```suggestion
        return match(true) {
            $this instanceof UnionType => $this->atLeastOneTypeIs($callable),
            $this instanceof IntersectionType => $this->everyTypeIs($callable),
            default => $callable($this),
        }
```

src/Symfony/Component/TypeInfo/TypeResolver/TypeResolver.php
```diff
@@ -0,0 +1,100 @@
...
+    public function resolve(mixed $subject, TypeContext $typeContext = null): Type
+    {
+        $subjectType = match (\is_object($subject)) {
+            true => match (true) {
+                is_subclass_of($subject::class, \ReflectionType::class) => \ReflectionType::class,
+                is_subclass_of($subject::class, \ReflectionFunctionAbstract::class) => \ReflectionFunctionAbstract::class,
+                default => $subject::class,
+            },
+            false => get_debug_type($subject),
+        };
```
welcoMattic, line 46:
```suggestion
    $subjectType = \is_object($subject) ?
        match (true) {
            is_subclass_of($subject::class, \ReflectionType::class) => \ReflectionType::class,
            is_subclass_of($subject::class, \ReflectionFunctionAbstract::class) => \ReflectionFunctionAbstract::class,
            default => $subject::class,
        } : 
        get_debug_type($subject)
    ;
```

welcoMattic (approved):
Minor comments, great work from both of you and reviewers :clap:

src/Symfony/Component/TypeInfo/TypeResolver/TypeResolver.php
Korbeil: I think the current version is more readable and I see no big improvement with your version of this snippet so I think I'll keep it as it is.

fabpot (approved):

fabpot:
Thank you @mtarld.

PR #52511, opened 2023-11-09
[ErrorHandler] Do not use hyphens in exception message

jkrzefski:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

This is just a minor cosmetic change in the error handler. Currently, error messages are shown with hyphenation if they exceed the available with. But in technical texts like exception messages, a hyphen could give the message a slightly different meaning. To reduce the risk for confusion, I removed the hyphenation.

### DIFF
--- a/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/profiler.css.twig
+++ b/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/profiler.css.twig
@@ -641,6 +641,7 @@ table tbody td.num-col {
     -webkit-hyphens: auto;
     -moz-hyphens: auto;
     hyphens: auto;
+    hyphenate-character: '';
 }
 .text-small {
     font-size: 12px !important;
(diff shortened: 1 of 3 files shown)


fabpot:
@javiereguiluz Do you like this change?

javiereguiluz:
@jkrzefski thanks for this proposal.

The problem with exception messages is that they usually include super-long "words" (e.g. FQCN) that don't fit the available space. If we don't force breaking them, the result is that the message shows big blank spaces.

**Example of current exception design:**

image

**Same exception after this PR:**

image

-----

In any case, I agree with you that the extra `-` character could be confusing for some people. I looked for if the hyphens could be customized ... and I learned about [hyphenate-character CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphenate-character).

So, instead of removing this CSS class, could you please change it to add the following?

```css
.break-long-words {
    /* ... */
    hyphenate-character: '';
}
```

This way, it'd look like this:

image

Thanks!

javiereguiluz:
I've added a commit here to make this PR mergeable. Symfony 7.1 enters "feature freeze" period next Monday and I'd like to have this included in it. Thanks!

yceruto (approved):

fabpot (approved):

fabpot:
Thank you @jkrzefski.

PR #52521, opened 2023-11-09
[AssetMapper] Improving exception if a vendor asset's path is not mapped

### LINKED ISSUE #52513: [AssetMapper] Adding the `assets/vendor` path by default to `framework.asset_mapper.paths`
### Description

I was upgrading my application to use `asset-mapper` (coming from `webpack-encore`) and inspired by the very useful [AssetMapper Livestream](https://www.youtube.com/watch?v=pyj1hvhcxGc) of Ryan, thanks for that!

I came across the following; since my application is sort of multi-application as it consists of a website, dashboard, (easy-)admin etc etc. I have the following `assets` structure in my project.

```
assets/
  admin/
  dashboard
  website/
  ...
```

In `6.3` the `AssetMapper` relied on the `url` of the package (within `importmap.php`) and that way loaded the packages externally. That all worked, but again inspired by the livestream to get all the new features, I upgraded to `6.4` and the way packages are loaded externally (by default) changed to loading the packages locally (in `assets/vendor` by default). 
I updated my `importmap.php` to drop the `url` usage (since it's been made deprecated directly in `6.4`) and switched to the `version` keys. And ran the new `importmap:install` command, which downloads the packages locally, cool!

I then encountered the following error:
Screenshot 2023-11-09 at 14 46 15

Since the packages were already installed (locally, in `assets/vendor` like it supposed to) I was pulling my hairs out for some time to find out what I did wrong... 😫

It finally came to mind that `AssetMapper` couldn't find the `assets/vendor` directory, and thus the packages, because of my config, not having the whole `assets` root:

```yaml
framework:
    asset_mapper:
        paths:
            assets/dashboard: 'dashboard'
            assets/easy_admin: 'easy_admin'
            assets/website: 'website'
```

As I didn't want to make my whole `assets` directory being available publicly, the application didn't automatically load the `assets/vendor` directory. Adding `assets/vendor: 'vendor'` to the paths solved my problem 😅

```yaml
framework:
    asset_mapper:
        paths:
            assets/dashboard: 'dashboard'
            assets/easy_admin: 'easy_admin'
            assets/vendor: 'vendor'         # <===
            assets/website: 'website'
```

So my suggestion would be to:
- Add the `assets/vendor` (or more precisely the `framework.asset_mapper.vendor_dir` config) directory to the `framework.asset_mapper.paths` array _**by default**_. Which seems logical as it would always need to be publicly available.
- Have a better exception message for this specific case, which would ease solving the issue.

### Example

_No response_

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #52513 
| License       | MIT

Screenshot 2023-11-09 at 1 33 49 PM

It's a bit long - but this should be a RARE error, when users are messing around with asset mapper paths. So giving them more info is better.

/cc @evertharmeling :)

### DIFF
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapGenerator.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapGenerator.php
@@ -232,7 +232,11 @@ class ImportMapGenerator
     private function createMissingImportMapAssetException(ImportMapEntry $entry): \InvalidArgumentException
     {
         if ($entry->isRemotePackage()) {
-            throw new LogicException(sprintf('The "%s" vendor asset is missing. Try running the "importmap:install" command.', $entry->importName));
+            if (!is_file($entry->path)) {
+                throw new LogicException(sprintf('The "%s" vendor asset is missing. Try running the "importmap:install" command.', $entry->importName));
+            }
+
+            throw new LogicException(sprintf('The "%s" vendor file exists locally (%s), but cannot be found in any asset map paths. Be sure the assets vendor directory is an asset mapper path.', $entry->importName, $entry->path));
         }
 
         throw new LogicException(sprintf('The asset "%s" cannot be found in any asset map paths.', $entry->path));


evertharmeling:
That's great! Thanks for the fast response!

And for the record, I wasn't 'messing around' 😅

weaverryan:
> And for the record, I wasn't 'messing around' 😅

Ha! Fair. You were doing the "completely normal use-cases... except that Ryan never considered them before" :). For most projects, I'm assuming users will never care or touch the asset mapper paths, but we'll see.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapGenerator.php
```diff
@@ -232,7 +232,11 @@ private function findEagerImports(MappedAsset $asset): array
     private function createMissingImportMapAssetException(ImportMapEntry $entry): \InvalidArgumentException
     {
         if ($entry->isRemotePackage()) {
-            throw new LogicException(sprintf('The "%s" vendor asset is missing. Try running the "importmap:install" command.', $entry->importName));
+            if (!file_exists($entry->path)) {
```
nicolas-grekas: is_file FTW

PR #52523, opened 2023-11-09
[AssetMapper] avoid caching MappedAsset inside JavaScript Import

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | None
| License       | MIT

Hi!

Subtle bug. In dev, we cache the `MappedAsset` objects. Before this PR, `MappedAsset->javascriptImports[0]->asset` was also a `MappedAsset` object.

Suppose `app.js` imports `foo.js` which imports `bar.js`. Then, the user changes `foo.js` to also import `baz.js`.

The `foo.js` MappedAsset will now have TWO `JavaScriptImport` objects (for `bar.js` and `baz.js`). However, when we calculate the preloads, this happens:

1) We load the cached `MappedAsset` for `app.js` because this file has not changed
2) It correctly has 1 `JavaScriptImport` for `foo.js`
3) But the `JavaScriptImport->asset` property holds an OUT-OF-DATE version of the `MappedAsset` for `foo.js`... because it was cached inside the `MappedAsset` structure for `app.js`. This out-of-date `MappedAsset` for `foo.js` only contains the ONE JavaScript import.

The result is that `baz.js` will be missing from the preloads.

This PR breaks the structure by only storing the asset's logical path (and sourcePath - needed by the cache system). Then, at runtime, we load the fresh `MappedAsset` from the logical path.

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapGenerator.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapGenerator.php
@@ -179,10 +179,15 @@ class ImportMapGenerator
 
             // check if this import requires an automatic importmap entry
             if ($javaScriptImport->addImplicitlyToImportMap) {
+                if (!$importedAsset = $this->assetMapper->getAsset($javaScriptImport->assetLogicalPath)) {
+                    // should not happen at this point, unless something added a bogus JavaScriptImport to this asset
+                    throw new LogicException(sprintf('Cannot find imported JavaScript asset "%s" in asset mapper.', $javaScriptImport->assetLogicalPath));
+                }
+
                 $nextEntry = ImportMapEntry::createLocal(
                     $importName,
-                    ImportMapType::tryFrom($javaScriptImport->asset->publicExtension) ?: ImportMapType::JS,
-                    $javaScriptImport->asset->logicalPath,
+                    ImportMapType::tryFrom($importedAsset->publicExtension) ?: ImportMapType::JS,
+                    $importedAsset->logicalPath,
                     false,
                 );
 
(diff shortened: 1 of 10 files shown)


src/Symfony/Component/AssetMapper/ImportMap/ImportMapGenerator.php
```diff
@@ -179,10 +179,16 @@ private function addImplicitEntries(ImportMapEntry $entry, array $currentImportE
 
             // check if this import requires an automatic importmap entry
             if ($javaScriptImport->addImplicitlyToImportMap) {
+                $importedAsset = $this->assetMapper->getAsset($javaScriptImport->assetLogicalPath);
+                if (!$importedAsset) {
```
nicolas-grekas:
```suggestion
                if (!$importedAsset = $this->assetMapper->getAsset($javaScriptImport->assetLogicalPath)) {
```

src/Symfony/Component/AssetMapper/ImportMap/ImportMapGenerator.php
```diff
@@ -223,7 +229,12 @@ private function findEagerImports(MappedAsset $asset): array
             $dependencies[] = $javaScriptImport->importName;
 
             // Follow its imports!
-            $dependencies = array_merge($dependencies, $this->findEagerImports($javaScriptImport->asset));
+            $nextAsset = $this->assetMapper->getAsset($javaScriptImport->assetLogicalPath);
+            if (!$nextAsset) {
```
nicolas-grekas: my usual (but consistent!) random comments :P
```suggestion
            if (!$nextAsset = $this->assetMapper->getAsset($javaScriptImport->assetLogicalPath)) {
```

src/Symfony/Component/AssetMapper/ImportMap/JavaScriptImport.php
```diff
@@ -11,21 +11,20 @@
...
  * Represents a module that was imported by a JavaScript file.
  */
 final class JavaScriptImport
 {
     /**
-     * @param string      $importName               The name of the import needed in the importmap, e.g. "/foo.js" or "react"
-     * @param MappedAsset $asset                    The asset that was imported
-     * @param bool        $addImplicitlyToImportMap Whether this import should be added to the importmap automatically
+     * @param string $importName               The name of the import needed in the importmap, e.g. "/foo.js" or "react"
+     * @param string $assetLogicalPath         logical path to the mapped ass that was imported
```
nicolas-grekas:
```suggestion
     * @param string $assetLogicalPath         Logical path to the mapped ass that was imported
```

PR #52526, opened 2023-11-09
Add some more non-countable English nouns

paullallier:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | N/A
| License       | MIT

@acirulis spotted that equipment was missing here: https://github.com/api-platform/core/issues/5957

I've added that and a few more non-countable nouns that I think are reasonably likely to be used by someone.  Arguably, time and work could also be here since they are normally non-countable, but there cases where times and works would be correct too.

And I corrected vocal to vowel, which is the correct English term

I'm not sure if this needs extra tests - happy to add if needed.

### DIFF
--- a/src/Symfony/Component/String/Inflector/EnglishInflector.php
+++ b/src/Symfony/Component/String/Inflector/EnglishInflector.php
@@ -21,7 +21,7 @@ final class EnglishInflector implements InflectorInterface
     private const PLURAL_MAP = [
         // First entry: plural suffix, reversed
         // Second entry: length of plural suffix
-        // Third entry: Whether the suffix may succeed a vocal
+        // Third entry: Whether the suffix may succeed a vowel
         // Fourth entry: Whether the suffix may succeed a consonant
         // Fifth entry: singular suffix, normal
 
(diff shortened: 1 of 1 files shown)


paullallier:
There's a coding standards failure in code I didn't really change - at least not in a way that changes the style.  I'm happy to add the changes anyway, if appropriate?

(The coding standard patch seems to use the str_contains function, which I think was new in PHP 8, so I assume we don't really want that in Symfony 5.4?)

paullallier:
Looking at this again, history -> histories is probably valid (though there are situations where history is also a valid plural).  

Health is trickier.  There are a few places that suggest things like this:
Screenshot 2023-11-10 at 21 21 12
But, as a native (British) English speaker I can't think of a case where I'd use healths rather than health for the plural.   doctrine/inflector 2.0.8 does also offer healths though...

So possibly we just conclude that English is an inconsistent mess.

stof:
regarding `health`, maybe we should offer both plural candidates

paullallier:
> regarding `health`, maybe we should offer both plural candidates

Perhaps yes, but I think it's mostly going to sounds strange with the s.  In fact, I think there's an argument to be made that nouns ending "LTH" are almost always uncountable.  Though I haven't even heard of some of these...
https://word.tips/words-end-in/lth/

nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @paullallier.

acirulis:
Oh, now I got plural "settingss" out of "settings" - should I open new issue?

xabbuh:
@acirulis please do

acirulis:
https://github.com/symfony/symfony/issues/52800

PR #52528, opened 2023-11-10
[Form] Add `webkitdirectory` to `FileType`

danielburger1337:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

See https://github.com/symfony/symfony/pull/52493

### DIFF
--- a/src/Symfony/Component/Form/CHANGELOG.md
+++ b/src/Symfony/Component/Form/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Add `webkitdirectory` to `FileType`
+
 7.0
 ---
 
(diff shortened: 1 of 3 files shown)


danielburger1337:
> You are targeting branch "7.0" but it seems your PR description refers to branch "7.1". 

Branch 7.1 doesnt exit yet. Just ignore it for now?

stof:
This can already be configured through the `attr` option, so I don't think we should add this option. We generally don't add options that just get copied into attr (for instance, we removed the `readonly` option a long time ago due to that).

So -1 from me for now, unless convinced otherwise.

danielburger1337:
Fair point by stof. I don't care either way.

nicolas-grekas:
I was wondering if we could provide a nicer way to support upload of directories. But I never tried so I have no clue :grimacing:

xabbuh:
I agree with @stof. 👎 from me

nicolas-grekas:
Thanks for submitting.

PR #52531, opened 2023-11-10
[WebProfilerBundle] Mark CodeExtension as non-internal

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

From https://github.com/symfony/symfony/pull/52449#issuecomment-1804257834 by @GromNaN 

Using those twig helpers in a profiler page is 100% legit. I think that the notice in the class' docbloc + the fact it moved to the WebProfilerBundle namespace allows us to mark the class back as non-internal.

### DIFF
--- a/src/Symfony/Bundle/WebProfilerBundle/Profiler/CodeExtension.php
+++ b/src/Symfony/Bundle/WebProfilerBundle/Profiler/CodeExtension.php
@@ -22,8 +22,6 @@ use Twig\TwigFilter;
  * that is never executed in a production environment.
  *
  * @author Fabien Potencier <fabien@symfony.com>
- *
- * @internal
  */
 final class CodeExtension extends AbstractExtension
 {


GromNaN:
It's also used in `symfony/demo`. https://github.com/symfony/demo/blob/206f0c9aab07cdc9f7437640b12947d35e9d5e72/templates/debug/source_code.html.twig#L33-L34
It will break if the demo is run with `env=prod`.

nicolas-grekas:
Demo needs to adjust, we don't want to support such use cases.

wouterj:
If we don't want users to use this functions/filters, should we remove them from the [twig reference](https://symfony.com/doc/current/reference/twig_reference.html)?

GromNaN:
> If we don't want users to use this functions/filters, should we remove them from the [twig reference](https://symfony.com/doc/current/reference/twig_reference.html)?

Or add a warning that those are only available for web profiler.

Is it a possible and good idea to register the extension conditionally in the profiler controller?

stof:
We cannot register extensions conditionally. Twig does allow registering new extensions after initialization is done, so this would be very fragile (and would break if your kernel handles multiple requests without a full shutdown for instance). It would also break the twig linting as it would complain about those functions not being available.

For the documentation, it would make sense to document that those functions are defined by WebProfilerBundle and not by TwigBundle. But I would not remove them entirely. Bundles providing custom profiler panels can legitimately use them.

GromNaN (approved):

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

PR #52294, opened 2023-10-25
[Routing][Serializer] Add annotation -> attribute aliases

wouterj:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #51381 
| License       | MIT

PR #52544, opened 2023-11-10
[Routing] Extend old Annotations from new Attributes

curry684:
If the base class is missing PHPStan fails to recognize the aliasing correctly.

| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | sort of
| New feature?  | no
| Deprecations? | no
| License       | MIT

Due to a shortcoming in PHPStan it does not pick up the previous aliasing correctly (ref. https://phpstan.org/r/2c7a9166-e8e1-42e5-8448-54c7d814e7d9).

This will cause projects to start failing their static analysis when they upgrade to 6.4. Explicitly extending the new attribute in the Attribute namespace has no runtime impact and shuts up PHPStan.

### DIFF
--- a/src/Symfony/Component/Serializer/Attribute/Ignore.php
+++ b/src/Symfony/Component/Serializer/Attribute/Ignore.php
@@ -18,9 +18,11 @@ namespace Symfony\Component\Serializer\Attribute;
  * @Target({"PROPERTY", "METHOD"})
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
+ *
+ * @final since Symfony 6.4
  */
 #[\Attribute(\Attribute::TARGET_METHOD | \Attribute::TARGET_PROPERTY)]
-final class Ignore
+class Ignore
 {
 }
 
--- a/src/Symfony/Component/Serializer/Attribute/SerializedName.php
+++ b/src/Symfony/Component/Serializer/Attribute/SerializedName.php
@@ -21,9 +21,11 @@ use Symfony\Component\Serializer\Exception\InvalidArgumentException;
  * @Target({"PROPERTY", "METHOD"})
  *
  * @author Fabien Bourigault <bourigaultfabien@gmail.com>
+ *
+ * @final since Symfony 6.4
  */
 #[\Attribute(\Attribute::TARGET_METHOD | \Attribute::TARGET_PROPERTY)]
-final class SerializedName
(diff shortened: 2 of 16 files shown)


nicolas-grekas:
We should do the same for all classes patched in #52294 
Can you please update them all?

curry684:
Yes, but not this weekend   :)

curry684:
Turns out I had some time to spare, so done. I had to remove some `final` keywords but imho they were pretty useless anyway in these attributes.

src/Symfony/Component/Serializer/Attribute/SerializedName.php
```diff
@@ -21,9 +21,10 @@
  * @Target({"PROPERTY", "METHOD"})
  *
  * @author Fabien Bourigault <bourigaultfabien@gmail.com>
+ * @todo
```
derrabus: ?

curry684: Oops, started to type a todo to restore the `final` before I concluded there was no reason to. Fixed  😆

nicolas-grekas (approved):

src/Symfony/Component/Serializer/Attribute/Ignore.php
```diff
@@ -20,7 +20,7 @@
  * @author Kévin Dunglas <dunglas@gmail.com>
  */
 #[\Attribute(\Attribute::TARGET_METHOD | \Attribute::TARGET_PROPERTY)]
-final class Ignore
```
stof, line 23: Please add the `@final` annotation so that we don't support BC for child classes (as they might not work when reading attributes anyway)

curry684: Fair enough, should be noted that many other attributes (ie. in DependencyInjection) are also not marked final. Out of scope for this PR to look into that.

derrabus (approved):

fabpot (approved):

fabpot:
Thank you @curry684.

src/Symfony/Component/Serializer/Attribute/Ignore.php
stof: For some of them, it is intended as we explicit added support for extending them.

PR #52555, opened 2023-11-12
[Intl][PhpUnitBridge] Fix invalid composer.json files & add composer.json validation integration tests

### LINKED ISSUE #52529: Fix Failing `composer validate` Command for `symfony/phpunit-bridge` & `symfony/intl`

6.2.*

### Description

Composer offers the [composer validate](https://getcomposer.org/doc/03-cli.md#validate) subcommand to ensure that a `composer.json` file is valid. I noticed when running the following command in a project that was using `symfony/phpunit-bridge` as a dev dependency that it issued the following warning:

```
composer validate --strict --no-ansi --no-interaction --with-dependencies

symfony/phpunit-bridge is valid, but with a few warnings
# General warnings
- Key php is a duplicate in /Users/ryanprussin/myproject/vendor/symfony/phpunit-bridge/composer.json at line 20
```
The file itself appears to repeat the php key in order to try to leave a comment, but duplicate keys are not deemed fully valid even though they are loosely allowed:
image

Looking at the `symfony/phpunit-bridge` repository itself, I see that it is a subtree split of this core Symfony project. Digging deeper, I cloned this core Symfony project and used the following command to find all `composer.json` files in it:
```
find ~/symfony -regex ".*composer.json"
```

I then modified it to pipe each `composer.json` file to the composer validate command in order to execute the validate command for each file and see if any others fail or have warnings:
```
find ~/symfony -regex ".*composer.json" | xargs -n1 composer validate --strict --no-ansi --no-interaction --with-dependencies
```

Of the 168 `composer.json` files in the project, only the following 2 composer files appear to have errors or warnings that need to be fixed:

```
src/Symfony/Bridge/PhpUnit/composer.json is valid, but with a few warnings
See https://getcomposer.org/doc/04-schema.md for details on the schema
# General warnings
- Key php is a duplicate in /Users/ryanprussin/symfony/symfony/src/Symfony/Bridge/PhpUnit/composer.json at line 20
```
```
src/Symfony/Component/Intl/Resources/emoji/composer.json is valid for simple usage with Composer but has
strict errors that make it unable to be published as a package
See https://getcomposer.org/doc/04-schema.md for details on the schema
# Publish errors
- name : The property name is required
- description : The property description is required
# General warnings
- No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
- require.unicode-org/cldr : unbound version constraints (*) should be avoided
```

### How to reproduce

1) Clone this project to your local device
2) Ensure composer is installed on your device
3) Execute the following command (assuming `~/symfony` is your project root):
```
find ~/symfony -regex ".*composer.json" | xargs -n1 composer validate --strict --no-ansi --no-interaction --with-dependencies
```
4) Notice that two of the files have an error

### Possible Solution
1) Fix Errors in `src/Symfony/Bridge/PhpUnit/composer.json`.
2) Fix Errors in `src/Symfony/Component/Intl/Resources/emoji/composer.json`.
3) Since this project is split up into a bunch of different subtree split repositories that depend on it, it might be nice to automate a complete composer validate check for all `composer.json` files in this repository via GitHub Action.

Edit: I have created a PR that does all of the above. See #52555.

### Additional Context

_No response_

RyanPrussin:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52529 
| License       | MIT / Unicode-DFS-2016 (see clarifying [comment](https://github.com/symfony/symfony/pull/52555#discussion_r1390331865))

As described in issue #52529, this fix addresses the following things:

1) Fix Errors in `src/Symfony/Bridge/PhpUnit/composer.json` (introduced before Symfony 5.4).
2) Fix Errors in `src/Symfony/Component/Intl/Resources/emoji/composer.json` (introduced in Symfony v6.2).
3) Since this project is split up into a bunch of different subtree split repositories that depend on it, I have added integration tests to run a complete composer validate check for all `composer.json` files in the `src/` directory of this repository as well as the main `composer.json` file via 2 new GitHub Actions. This will ensure that when developers add new composer files in the future we will have a full guarantee of validity for all Symfony subtrees splits.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -151,6 +151,12 @@ jobs:
           entrypoint: /bin/bash
           args: -c "(/opt/bitnami/openldap/bin/ldapwhoami -H ldap://ldap:3389 -D cn=admin,dc=symfony,dc=com -w symfony||sleep 5) && /opt/bitnami/openldap/bin/ldapadd -H ldap://ldap:3389 -D cn=admin,dc=symfony,dc=com -w symfony -f src/Symfony/Component/Ldap/Tests/Fixtures/data/fixtures.ldif && /opt/bitnami/openldap/bin/ldapdelete -H ldap://ldap:3389 -D cn=admin,dc=symfony,dc=com -w symfony cn=a,ou=users,dc=symfony,dc=com"
 
+      - name: Composer Validate Main composer.json
+        run: composer validate --strict --no-ansi --no-interaction --with-dependencies
+
+      - name: Composer Validate composer.json Files in src/ Directory
+        run: find src -regex ".*composer.json" | xargs -n1 composer validate --strict --no-ansi --no-interaction --with-dependencies
+
       - name: Install dependencies
         run: |
           COMPOSER_HOME="$(composer config home)"
--- a/src/Symfony/Bridge/PhpUnit/composer.json
+++ b/src/Symfony/Bridge/PhpUnit/composer.json
@@ -16,8 +16,6 @@
         }
     ],
     "require": {
-        "php": ">=7.1.3 EVEN ON LATEST SYMFONY VERSIONS TO ALLOW USING",
-        "php": "THIS BRIDGE WHEN TESTING LOWEST SYMFONY VERSIONS.",
         "php": ">=7.1.3"
     },
     "require-dev": {
--- a/src/Symfony/Component/Intl/Resources/emoji/composer.json
+++ b/src/Symfony/Component/Intl/Resources/emoji/composer.json
@@ -1,4 +1,7 @@
 {
+    "name": "symfony/intl-emoji",
+    "description": "Import Unicode Common Locale Data Repository and dependencies for the Emoji Transliterator Builder",
+    "license": "Unicode-DFS-2016",
     "repositories": [
         {
             "type": "package",
@@ -18,6 +21,6 @@
         "symfony/filesystem": "^6",
         "symfony/finder": "^6",
         "symfony/var-exporter": "^6",
-        "unicode-org/cldr": "*"
+        "unicode-org/cldr": "^2022"
     }
 }


src/Symfony/Component/Intl/Resources/emoji/composer.json
```diff
@@ -1,4 +1,7 @@
 {
+    "name": "symfony/intl-emoji",
+    "description": "Import Unicode Common Locale Data Repository and dependencies for the Emoji Transliterator Builder",
+    "license": "Unicode-DFS-2016",
```
RyanPrussin, line 4: License is required for `composer validate --strict` enforcement, but it was hard to figure out how to find the [composer-accepted](https://getcomposer.org/doc/04-schema.md#license) [SPDX license](https://spdx.org/licenses/) type that relates to the [UNICODE LICENSE V3](https://github.com/unicode-org/cldr/blob/main/LICENSE) from the `unicode-org/cldr` project.

[This page](https://spdx.org/licenses/Unicode-DFS-2016.html) for the `Unicode-DFS-2016` license eventually confirmed for me that https://www.unicode.org/license.txt (UNICODE LICENSE V3) is listed as another page for the same license.

src/Symfony/Bridge/PhpUnit/composer.json
```diff
@@ -16,8 +16,6 @@
         }
     ],
     "require": {
-        "php": ">=7.1.3 EVEN ON LATEST SYMFONY VERSIONS TO ALLOW USING",
-        "php": "THIS BRIDGE WHEN TESTING LOWEST SYMFONY VERSIONS.",
```
RyanPrussin, line 20: I believe this was left as a comment reminder for convenience, but JSON files do not allow comments so I wasn't sure how to keep it while also keeping the file valid. Open to suggestions (perhaps adding the comment in another key?), but ultimately I figured that it should be relatively self-evident when working on `[PhpUnitBridge]` that it needs to maintain support for the lowest Symfony versions, but perhaps this is a false assumption on my part.

src/Symfony/Component/Intl/Resources/emoji/composer.json
RyanPrussin: Also, to be clear, I see in the Symfony contribution guide for [proposing a change](https://symfony.com/doc/current/contributing/code/pull_requests.html#make-a-pull-request) that a license not marked as MIT will be automatically rejected.

It is worth noting that I am not adding any new non-MIT dependencies, I am simply flagging the actual license of `unicode-org/cldr`, which we are already using in the codebase.

.github/workflows/integration-tests.yml
```diff
@@ -151,6 +151,12 @@ jobs:
           entrypoint: /bin/bash
           args: -c "(/opt/bitnami/openldap/bin/ldapwhoami -H ldap://ldap:3389 -D cn=admin,dc=symfony,dc=com -w symfony||sleep 5) && /opt/bitnami/openldap/bin/ldapadd -H ldap://ldap:3389 -D cn=admin,dc=symfony,dc=com -w symfony -f src/Symfony/Component/Ldap/Tests/Fixtures/data/fixtures.ldif && /opt/bitnami/openldap/bin/ldapdelete -H ldap://ldap:3389 -D cn=admin,dc=symfony,dc=com -w symfony cn=a,ou=users,dc=symfony,dc=com"
 
+      - name: Composer Validate Main composer.json
+        run: composer validate --strict --no-ansi --no-interaction --with-dependencies
+
+      - name: Composer Validate composer.json Files in src/ Directory
+        run: find src -regex ".*composer.json" | xargs -n1 composer validate --strict --no-ansi --no-interaction --with-dependencies
```
RyanPrussin, line 158: This will validate all 163 `composer.json` files in the `src/` directory under strict mode (which ensures that warnings will return a non-zero exit code as well, just like errors already do by default). Furthermore, if any new `composer.json` files are added in that directory in the future, it will immediately catch them and ensure they are up to project standards from the very beginning.

In case a reader is wondering why warnings matter, aside from ensuring that we keep the code up to the tightest standards, there are also external projects that depend on this level of strictness. For example, the [Nix package manager](https://github.com/NixOS/nixpkgs/blob/8213089f8dc69fe36b1d4cb5a4d8eeb3618ec357/pkgs/build-support/php/hooks/composer-install-hook.sh#L110), which is used to package up software for general use, development environments, and deployments.

nicolas-grekas:
Thanks for your proposal. I'm sorry but I'm not convinced we need automation for these.
The patched composer.json files are good as is - they're on purpose to me.

src/Symfony/Component/Intl/Resources/emoji/composer.json
nicolas-grekas: This directory is under MIT - but actually we don't care because this is composer.json is just a helper to run a dev command

src/Symfony/Bridge/PhpUnit/composer.json
nicolas-grekas: this is left on purpose - it's allowed and it works. it doesn't validate, but this is not an issue in practice so we prefer keeping the notice, which prevents unwanted changes

nicolas-grekas:
Closing as explained, thanks for proposing.

PR #52558, opened 2023-11-12
UriResolver support path with columns

vdauchy:
| Q             | A
| ------------- | ---
| Branch?       | 6.3 
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix UriResolver bad handling of column in path
| License       | MIT

Resolving links on pages using weird pagination like: ```https://localhost/domain/search/page:5``` fails due to `:` making 

```
var_dump(parse_url('/page:1', \PHP_URL_SCHEME));
```

Return `false` (and not null as expected in the code).

This simply ensure the absolute URL is returned only if the SCHEME is found (ie a string is returned by `parse_url`).


vdauchy:
Re-openned on branch 6.3: https://github.com/symfony/symfony/pull/52559

PR #52559, opened 2023-11-12
[DomCrawler] UriResolver support path with columns

vdauchy:
| Q             | A
| ------------- | ---
| Branch?       | 6.3 
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix UriResolver bad handling of column in path
| License       | MIT

Resolving links on pages using weird pagination like: ```https://localhost/domain/search/page:5``` fails due to `:` making 

```
var_dump(parse_url('/page:1', \PHP_URL_SCHEME));
```

Return `false` (and not null as expected in the code).

This simply ensure the absolute URL is returned only if the SCHEME is found (ie a string is returned by `parse_url`).

### DIFF
--- a/CHANGELOG-5.4.md
+++ b/CHANGELOG-5.4.md
@@ -7,6 +7,23 @@ in 5.4 minor versions.
 To get the diff for a specific change, go to https://github.com/symfony/symfony/commit/XXX where XXX is the change hash
 To get the diff between two versions, go to https://github.com/symfony/symfony/compare/v5.4.0...v5.4.1
 
+* 5.4.31 (2023-11-10)
+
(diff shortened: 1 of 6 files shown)


derrabus:
Is 5.4 not affected by this bug?

vdauchy:
@derrabus From https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/DomCrawler/UriResolver.php I think the but does exist in 5.4.

Shall I re-do the a PR targeting 5.4 branch then the patch will be automatically applied upstream ?

derrabus:
Yes, please target 5.4 then. No need to "re-do" the PR, just rebase this one and change the target.

vdauchy:
Moved to https://github.com/symfony/symfony/pull/52579

PR #52566, opened 2023-11-13
[Workflow] Add `AsMarkingStore` attribute

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

Originally, this idea came up after I implemented a new marking store as stated [here](https://github.com/symfony/symfony/issues/44211#issuecomment-1075401547).

In a project we're working on, we're using the Workflow component. We created a marking store to support enums for our markings. We'd like to make it a bit generic (understand: support enum and just give to the marking store the property that hold the marking on the subject). Here is the simplified marking store:

```php
final readonly class DeploymentStatusMarkingStore implements MarkingStoreInterface
{
    public function __construct(private string $property)
    {
    }

    public function getMarking(object $subject): Marking
    {
        return new Marking(/* ... */);
    }

    public function setMarking(object $subject, Marking $marking, array $context = []): void
    {
        // ...

        try {
            $marking = [$enumClass, 'from']($marking);
        } catch (\Error) {
            throw new InvalidArgumentException(sprintf('The constant "%s" does not exist.', $marking));
        }

        $subject->{$method}($marking);
    }
}
```

The `$property` attribute holds the fieldname of the marking in the subject. However, in the current state, we must define manually the service to provide `$property` to the constructor:

```php
return static function (FrameworkConfig $frameworkConfig, ContainerConfigurator $configurator): void {
    $configurator->services()
        ->set('workflow.marking_store.status_enum', DeploymentStatusMarkingStore::class)
        ->args(['status']); // This is the boilerplate that I'd like to remove

    // ...

    $deploymentStatus->markingStore()
        ->service('workflow.marking_store.status_enum');

    // ...
};
```

Yes! We can now use the marking store.

But I'd like to remove this boilerplate by introducing `AsMarkingStore`. Thanks to this attribute, we can now define our marking store like this:

```php
#[AsMarkingStore('status')]
final readonly class DeploymentStatusMarkingStore extends AbstractMarkingStore
{
    public function getMarking(object $subject): Marking
    {
        // ...
    }

    public function setMarking(object $subject, Marking $marking, array $context = []): void
    {
        // ...
    }
}
```

The configuration is also simplified:

```php
return static function (FrameworkConfig $frameworkConfig, ContainerConfigurator $configurator): void {
    // ...

    // No more marking store definition to do by hand!
    $deploymentStatus->markingStore()
        ->service(DeploymentStatusMarkingStore::class);

    // ...
};
```

This attribute also allows to configure the name of the property that hold the marking field name in the store:

```php
#[AsMarkingStore(markingName: 'status', property: 'markingField')]
final readonly class SingleStateEnumMarkingStore implements MarkingStoreInterface
{
    public function __construct(private string $markingField)
    {
    }

    public function getMarking(object $subject): Marking
    {
        // ...
    }

    public function setMarking(object $subject, Marking $marking, array $context = []): void
    {
        // ...
    }
}
```

Thanks to this, we could even create in our app an `AbstractEnumMarkingStore`, making some store inherit this class and just define the `AsMarkingStore` attribute with the good field name.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/FrameworkBundle.php
+++ b/src/Symfony/Bundle/FrameworkBundle/FrameworkBundle.php
@@ -71,6 +71,7 @@ use Symfony\Component\VarExporter\Internal\Hydrator;
 use Symfony\Component\VarExporter\Internal\Registry;
 use Symfony\Component\Workflow\DependencyInjection\WorkflowDebugPass;
 use Symfony\Component\Workflow\DependencyInjection\WorkflowGuardListenerPass;
+use Symfony\Component\Workflow\DependencyInjection\WorkflowMarkingStorePass;
 
 // Help opcache.preload discover always-needed symbols
 class_exists(ApcuAdapter::class);
@@ -157,6 +158,7 @@ class FrameworkBundle extends Bundle
         $container->addCompilerPass(new CachePoolPrunerPass(), PassConfig::TYPE_AFTER_REMOVING);
         $this->addCompilerPassIfExists($container, FormPass::class);
         $this->addCompilerPassIfExists($container, WorkflowGuardListenerPass::class);
+        $this->addCompilerPassIfExists($container, WorkflowMarkingStorePass::class);
         $container->addCompilerPass(new ResettableServicePass(), PassConfig::TYPE_BEFORE_OPTIMIZATION, -32);
         $container->addCompilerPass(new RegisterLocaleAwareServicesPass());
         $container->addCompilerPass(new TestServiceContainerWeakRefPass(), PassConfig::TYPE_BEFORE_REMOVING, -32);
--- a/src/Symfony/Component/Workflow/Attribute/AsMarkingStore.php
+++ b/src/Symfony/Component/Workflow/Attribute/AsMarkingStore.php
@@ -1,0 +1,31 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Workflow\Attribute;
+
+/**
+ * @author Alexandre Daubois <alex.daubois@gmail.com>
+ */
+#[\Attribute(\Attribute::TARGET_CLASS)]
+final class AsMarkingStore
+{
+    public function __construct(
+        /**
+         * The name of the property where the marking will be stored in the subject.
+         */
+        public string $markingName,
+        /**
+         * The name of the property where the marking name will be stored.
+         */
+        public string $property = 'property',
+    ) {
+    }
+}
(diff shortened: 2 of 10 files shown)


lyrixx:
I'm not really convinced by this PR. but I may have missed something.

You have this, and you want to remove it:
```php
    $configurator->services()
        ->set('workflow.marking_store.status_enum', DeploymentStatusMarkingStore::class)
        ->args(['status']); // This is the boilerplate that I'd like to remove
```

OK, why not!

Instead you want to do this : 
```php
#[AsMarkingStore(markingName: 'status', property: 'markingField')]
final readonly class SingleStateEnumMarkingStore implements MarkingStoreInterface
{
    public function __construct(private string $markingField)
    {
    }
```

I got it right?

Instead, why don't you do:


```diff
-#[AsMarkingStore(markingName: 'status', property: 'markingField')]
 final readonly class SingleStateEnumMarkingStore implements MarkingStoreInterface
 {
-    public function __construct(private string $markingField)
+    public function __construct(private string $markingField = 'status')
     {
     }
```

alexandre-daubois:
This would allow to do something like this:

```php
abstract class AbstractEnumMarkingStore extends AbstractMarkingStore
{
    public function getMarking(object $subject): Marking
    {
        // ...
    }

    public function setMarking(object $subject, Marking $marking, array $context = []): void
    {
        // ...
    }
}

#[AsMarkingStore('deploymentStatus')]
class DeploymentStatusEnumStore extends AbstractEnumMarkingStore
{
}

#[AsMarkingStore('postStatus')]
class BlogPostEnumStore extends AbstractEnumMarkingStore
{
}
```

Instead of declaring the constructor everytime to pass the field name like this:

```php
class DeploymentStatusEnumStore extends AbstractEnumMarkingStore
{
    public function __construct()
    {
        parent::__construct('deploymentStatus');
    }
}
```

It makes it even more concise (and also I like this attribute-style declaration, but that's a more subjective point of view :smile: )

Also, the PR adds a `workflow.marking_store` tag to the class (which could be generalized to `MarkingStoreInterface` by the way). This would allow to inject for instance a tagged locator with all those stores and create a more "global" marking store choosing the right store depending on some logic.

lyrixx:
I don't like, but I'm not against it. 

I found it weird that it configure a constructor this way. And it does not scale that good with many different service, you have to create a new class for each. IMHO, this is better to create only one class with N services definition. Or create N classes with a constructor override, like you explain in your last example.

Anyway, This is not related to the workflow, it could be generalized to any kind of service, couldn't it?

joelwurtz:
I'm also not a fan of this, adding an attribute to avoid calling a parameter in the parent constructor seems superfluous comparing to the maintenance cost.

We can compare this when declaring a specific repository for Doctrine (calling the parent constructor with the class name could be replaced by an attribute) and i was never bother by this way of doing it.

alexandre-daubois:
With hindsight, I think you're both right. I may try a different approach, but not this one. Thank you both for your feedback :+1:

PR #52567, opened 2023-11-13
[AssetMapper] Fixing js sourceMappingURL extraction when sourceMappingURL used in code

### LINKED ISSUE #52547: Requiring es-module-shims locally in AssetMapper causes console error

6.3, 6.4-dev, 6.4.0-beta3

### Description

Using the default remotely-hosted shims works as expected. If requiring the polyfills as a local vendor asset, an error occurs:

**Javascript console**
```
es-module-shims.index-50eab39155ca286f16894eb2ba1c2740.js:7 Uncaught SyntaxError: Invalid or unexpected token (at es-module-shims.index-50eab39155ca286f16894eb2ba1c2740.js:7:12535)
```

The generated file (`assets/vendor/es-module-shims/es-module-shims.index.js`) appears to be mangled.

For projects not needing the shim support, explicitly disabling it removes the error. Including it as a CDN asset, however, means the user will need to write a server CSP allowing the remote host, which may be undesirable. In Apache2:

**security.conf**
```
Header always set Content-Security-Policy "script-src blob: https://ga.jspm.io;"
```

### How to reproduce

In a Symfony ^6.3 project using [AssetMapper](https://symfony.com/doc/6.4/frontend/asset_mapper.html#serving-assets-in-dev-vs-prod), require ES Module Shims  as a local dependency:

```
./bin/console importmap:require es-module-shims

[OK] Package "es-module-shims" added to importmap.php.                               

      Use the new package normally by importing "es-module-shims".
```

Load the project page in a browser and review the inspector console for errors:

es-module-shims-console-error

### Possible Solution

A workaround if IE11 support isn't needed is to just disable the polyfill:

**config/packages/asset_mapper.yaml**
```yaml
framework:
    asset_mapper:
        paths:
            - assets/
        importmap_polyfill: false
```

### Additional Context

_No response_

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52547
| License       | MIT

We extract & remove `sourceMappingURL` because we don't (at least currently) also download the source map file. This fixes the case where the `sourceMappingURL` string itself is used in the code.

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
@@ -248,7 +248,10 @@ final class JsDelivrEsmResolver implements PackageResolverInterface
         }, $content);
 
         // source maps are not also downloaded - so remove the sourceMappingURL
-        $content = preg_replace('{//# sourceMappingURL=.*$}m', '', $content);
+        // remove the final one only (in case sourceMappingURL is used in the code)
+        if (false !== $lastPos = strrpos($content, '//# sourceMappingURL=')) {
+            $content = substr($content, 0, $lastPos).preg_replace('{//# sourceMappingURL=.*$}m', '', substr($content, $lastPos));
+        }
 
         return preg_replace('{/\*# sourceMappingURL=[^ ]*+ \*/}', '', $content);
     }
--- a/src/Symfony/Component/AssetMapper/Tests/ImportMap/Resolver/JsDelivrEsmResolverTest.php
+++ b/src/Symfony/Component/AssetMapper/Tests/ImportMap/Resolver/JsDelivrEsmResolverTest.php
@@ -430,7 +430,30 @@ class JsDelivrEsmResolverTest extends TestCase
             ],
         ];
 
-        yield 'css file removes importmap' => [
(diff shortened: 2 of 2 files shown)


src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -248,7 +248,11 @@ private function makeImportsBare(string $content, array &$dependencies): string
         }, $content);
 
         // source maps are not also downloaded - so remove the sourceMappingURL
-        $content = preg_replace('{//# sourceMappingURL=.*$}m', '', $content);
+        $lastPos = strrpos($content, '//# sourceMappingURL');
+        // remove the final one only (in case sourceMappingURL is used in the code)
+        if (false !== $lastPos) {
+            $content = substr($content, 0, $lastPos).preg_replace('{//# sourceMappingURL=.*$}m', '', substr($content, $lastPos));
+        }
```
```suggestion
        // remove the final one only (in case sourceMappingURL is used in the code)
        if (false !== $lastPos = strrpos($content, '//# sourceMappingURL=')) {
            $content = substr($content, 0, $lastPos).preg_replace('{//# sourceMappingURL=.*$}m', '', substr($content, $lastPos));
        }
```

nicolas-grekas:
Thank you @weaverryan.

PR #52568, opened 2023-11-13
[VarExporter] Deprecate per-property lazy-initializers

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Issues        | -
| License       | MIT

This is a very late deprecation for a complex feature that we'd better drop ASAP IMHO.
[Property hooks](https://wiki.php.net/rfc/property-hooks) should be a better solution, even thought it's still WIP.

### DIFF
--- a/UPGRADE-6.4.md
+++ b/UPGRADE-6.4.md
@@ -38,6 +38,7 @@ Components
 * [Serializer](#Serializer)
 * [Templating](#Templating)
 * [Validator](#Validator)
+* [VarExporter](#VarExporter)
 * [Workflow](#Workflow)
 
 BrowserKit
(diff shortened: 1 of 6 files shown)


stof:
Do you know any place where this feature has been used ?

nicolas-grekas:
None. That's also why I'm proposing this :)

derrabus:
> None. That's also why I'm proposing this :)

😬 

https://github.com/doctrine/DoctrineBundle/blob/ca64ca70247d7b1b56a57c3b147b77253eaea2f5/Repository/LazyServiceEntityRepository.php#L52-L56

stof:
Looks like this is actually used...

nicolas-grekas:
Here the PR to stop relying on this feature: https://github.com/doctrine/DoctrineBundle/pull/1727

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

PR #52569, opened 2023-11-13
[VarExporter] Drop support for partially initialized lazy object

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Sidekick of #52568

### DIFF
--- a/UPGRADE-7.0.md
+++ b/UPGRADE-7.0.md
@@ -46,6 +46,7 @@ Components
  * [Translation](#Translation)
  * [Validator](#Validator)
  * [VarDumper](#VarDumper)
+ * [VarExporter](#VarExporter)
  * [Workflow](#Workflow)
  * [Yaml](#Yaml)
 
@@ -607,6 +608,11 @@ VarDumper
 
  * Add parameter `string $label = null` to `VarDumper::dump()`
  * Require explicit argument when calling `VarDumper::setHandler()`
+
+VarExporter
+-----------
+
+ * Remove support for per-property lazy-initializers
 
 Workflow
 --------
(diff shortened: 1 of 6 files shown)


derrabus:
⚠️ This PR will break DoctrineBundle which has already been made compatible with Symfony 7.

nicolas-grekas:
Ah, I forgot about this one. As I wrote on another PR, ORMv3 makes `LazyServiceEntityRepository` useless, we should deprecate and drop it.

derrabus:
> ORMv3 makes `LazyServiceEntityRepository` useless

Yes, but we still need a solution for ORM 2 which will stay around for a while.

> we should deprecate and drop it.

`LazyServiceEntityRepository` is marked as internal, so we can drop that class without deprecation. If we find a way to keep the laziness of `ServiceEntityRepository` for ORM 2, I'm fine with the late removal of this feature. I won't have time to work on that though. 😕

back-2-95:
So how to overcome this? My SF7 project I now need to keep `symfony/var-exporter` at `7.0.0-BETA1`.
Or is there some release coming which would fix this?

stof:
There is an in-progress PR for DoctrineBundle.

stof:
Actually, the PR is even merged. So if you use the dev version of DoctrineBundle, it should be OK.

back-2-95:
2.11-dev or 2.12-dev ?

nicolas-grekas:
Here is the PR: https://github.com/doctrine/DoctrineBundle/pull/1727

stof:
2.11-dev for now. The change has not yet reached 2.12-dev as the merge-up between branches has not yet been done after the merge of that PR.

back-2-95:
Ok, I required 2.11-dev for Doctrine Bundle and it worked. Now all my Symfony packages are RC1. Thanks for quick responses.

derrabus:
I've released DoctrineBundle 2.11.1 that includes the fix.

Issue #50759, opened 2023-06-23
#[MapRequestPayload] Returns full ViolationsList if object cannot be created due to a single missing value


Havrin:
### Symfony version(s) affected

6.3.0

### Description

If you use #[MapRequestPayload] and the payload is missing a single (or multiple) fields, the ViolationList created by the resolver does not contain the single missing field, but **all** asserts on every field. 

### How to reproduce

Controller:
```php
...
public function testPayloadAction(
        #[MapRequestPayload] User $user
    ): JsonResponse {
        return new JsonResponse(data: $user->getEmail());
    }
...
```

DTO:
```php
final readonly class User
{
    public function __construct(
        #[Assert\NotNull,
            Assert\NotBlank,
            Assert\Email]
        private string $email,
        #[Assert\NotNull,
            Assert\NotBlank]
        private string $password,
    ) {
    }

    public function getEmail(): string
    {
        return $this->email;
    }

    public function getPassword(): string
    {
        return $this->password;
    }
}
```
Payload to route:
```json
{
  "email": "test@mail.com"
}
```
Excpected: I would expect the error message that [`PartialDenormalizationException`](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Serializer/Exception/PartialDenormalizationException.php) holds in [`RequestPayloadValueResolver.php`](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php#L108) to be returned. In this case: 
> Failed to create object because the class misses the "password" property.

Actual:
```
:
    This value should be of type unknown.
Object(...\DTO\User).email:
    This value should not be null. (code ad32d13f-c3d4-423b-909a-857b961eb720)
Object(...\DTO\User).email:
    This value should not be blank. (code c1051bb4-d103-4f74-8988-acbcafc7fdc3)
Object(...\DTO\User).password:
    This value should not be null. (code ad32d13f-c3d4-423b-909a-857b961eb720)
Object(...\DTO\User).password:
    This value should not be blank. (code c1051bb4-d103-4f74-8988-acbcafc7fdc3)

```

### Possible Solution

Do not run the validation on an empty object in [`RequestPayloadValueResolver.php`](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php#L123C17-L123C17)

[...]

PR #52578, opened 2023-11-14
[Serializer] Fix denormalize constructor arguments

### LINKED ISSUE #52422: [Serializer] Constructor arguments are mapped wrong if request is missing an argument

6.3.6

### Description

This issue rose while using the `MapRequestPayload`  Attribute on a request.

Consider the following DTO and Controller:

```php
final class ChangePassword
{
    #[SerializedName(serializedName: 'old_password')]
    private ?string $oldPassword = null;

    public function __construct(
        #[Assert\NotNull,
            Assert\NotBlank,
            Assert\Email]
        private string $email,
        #[Assert\NotNull,
            Assert\NotBlank]
        private string $password,
        ?string $oldPassword = null
    ) {
        $this->oldPassword = $oldPassword;
    }

    public function getEmail(): string
    {
        return $this->email;
    }

    public function getPassword(): string
    {
        return $this->password;
    }

    public function getOldPassword(): ?string
    {
        return $this->oldPassword;
    }
}
```
  
```php
   #[Route(path: 'api/change-password', name: 'change-password', methods: ['POST'])]
    public function changePasswordAction(#[MapRequestPayload] ChangePassword $changePassword): JsonResponse
    {
        return new JsonResponse($changePassword->getEmail());
    }
```

When calling this route with
```json
{
  "password": "abc",
  "old_password": "def"
}
```

the created object looks like this:
```
object(App\DTO\ChangePassword)#2315 (3) {
["oldPassword":"App\DTO\ChangePassword":private]=>
NULL
["email":"App\DTO\ChangePassword":private]=>
string(3) "abc"
["password":"App\DTO\ChangePassword":private]=>
string(3) "def"
}
```

the values are not correctly mapped

### How to reproduce

A simple way to reproduce it is described above.

It's also possible to use my github project: https://github.com/Havrin/symfony-payload-bug

start it and throw the following request against `https://localhost/api/change-password`
```
{
  "password": "abc",
  "old_password": "def"
}
```

### Possible Solution

While debugging I noticed, that https://github.com/symfony/serializer/blob/1197823bacefc3397fafb8f10709c1574b2f69ce/Normalizer/AbstractNormalizer.php#L376 is creating an array of values with just an index-number as key.
After creating the object in https://github.com/symfony/serializer/blob/1197823bacefc3397fafb8f10709c1574b2f69ce/Normalizer/AbstractNormalizer.php#L420 the values are wrongly mapped. Seems like the code is not correctly using the name as key but rather the index-number.

### Additional Context

The wrong mapping started occurring after this fix: https://github.com/symfony/symfony/issues/51907

### LINKED ISSUE #52499: [Serializer] Only unset data in AbstractNormalizer::instantiateObject if constructor invocation succeeded

6.3.x,5.4.x

### Description

The `AbstractNormalizer::instantiateObject` unsets the corresponding `&$data` value for a constructor argument if it believes it will be set by constructor invocation so that no `set` is called later on. However, if constructor invocation fails as it is missing arguments, the value won't be set again at a later stage, even if the object could be instantiated partially when setting the property later on.

### How to reproduce

If we have an example object

```php
class ExampleObject
{
    public function __construct(public readonly string $propertyMissing, public readonly string $propertyGiven)
    {
    }
}
```

I would expect the following test to pass:

```php
try {
    $exampleObject = $serializer->denormalize(['propertyGiven' => 'foo'],
        ExampleObject::class,
        null,
        [DenormalizerInterface::COLLECT_DENORMALIZATION_ERRORS => true]);
} catch (PartialDenormalizationException $e) {
    $exampleObject = $e->getData();
}
self::assertInstanceOf(ExampleObject::class, $exampleObject);
self::assertEquals('foo', $exampleObject->propertyGiven);
```

however it says: `$propertyGiven must not be accessed before initialization` as there was no attempt to set it later on after the constructor instantiation failed.

### Possible Solution

Remember the properties that have to be unset if the constructor instantiation succeeds and only unset them if there are no missing parameters and the instantiation actually worked.

### Additional Context

I realized this as the behaviour changed with commit [c54cfbb6f3129c41ea5b26b4e3115919132ab726](https://github.com/symfony/serializer/commit/c54cfbb6f3129c41ea5b26b4e3115919132ab726) as it now collects all missing arguments. Before it returned early and the `$data` values for later constructor arguments were not unset. So this is kind of a regression, although before it just worked "accidentially" in my use case as the first argument was already missing for my data.

mtarld:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52499, Fix #52422
| License       | MIT

Since this PR: https://github.com/symfony/symfony/pull/51907, objects with partial constructor parameters were wrongly instantiated.

This PR fixes that issue by delegating the properties values assignment, by unsetting normalized data only when the constructor has been called properly.

This might correct #50759 as well.

### DIFF
--- a/src/Symfony/Component/Serializer/Serializer.php
+++ b/src/Symfony/Component/Serializer/Serializer.php
@@ -228,8 +228,20 @@ class Serializer implements SerializerInterface, ContextAwareNormalizerInterface
             $context['not_normalizable_value_exceptions'] = [];
             $errors = &$context['not_normalizable_value_exceptions'];
             $denormalized = $normalizer->denormalize($data, $type, $format, $context);
+
             if ($errors) {
-                throw new PartialDenormalizationException($denormalized, $errors);
+                // merge errors so that one path has only one error
+                $uniqueErrors = [];
+                foreach ($errors as $error) {
+                    if (null === $error->getPath()) {
+                        $uniqueErrors[] = $error;
+                        continue;
+                    }
+
+                    $uniqueErrors[$error->getPath()] = $uniqueErrors[$error->getPath()] ?? $error;
+                }
+
+                throw new PartialDenormalizationException($denormalized, array_values($uniqueErrors));
             }
 
             return $denormalized;
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
@@ -350,6 +350,7 @@ abstract class AbstractNormalizer implements NormalizerInterface, DenormalizerIn
             $constructorParameters = $constructor->getParameters();
             $missingConstructorArguments = [];
             $params = [];
+            $unsetKeys = [];
             foreach ($constructorParameters as $constructorParameter) {
                 $paramName = $constructorParameter->name;
                 $key = $this->nameConverter ? $this->nameConverter->normalize($paramName, $class, $format, $context) : $paramName;
@@ -368,18 +369,17 @@ abstract class AbstractNormalizer implements NormalizerInterface, DenormalizerIn
                         }
 
                         $params = array_merge($params, $variadicParameters);
-                        unset($data[$key]);
+                        $unsetKeys[] = $key;
                     }
                 } elseif ($allowed && !$ignored && (isset($data[$key]) || \array_key_exists($key, $data))) {
                     $parameterData = $data[$key];
                     if (null === $parameterData && $constructorParameter->allowsNull()) {
                         $params[] = null;
-                        // Don't run set for a parameter passed to the constructor
-                        unset($data[$key]);
+                        $unsetKeys[] = $key;
+
                         continue;
                     }
 
-                    // Don't run set for a parameter passed to the constructor
                     try {
                         $params[] = $this->denormalizeParameter($reflectionClass, $constructorParameter, $paramName, $parameterData, $context, $format);
                     } catch (NotNormalizableValueException $exception) {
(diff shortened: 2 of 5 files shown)


mtarld:
fabbot errors aren't related.

src/Symfony/Component/Serializer/Serializer.php
```diff
@@ -228,8 +228,24 @@ public function denormalize($data, string $type, string $format = null, array $c
             $context['not_normalizable_value_exceptions'] = [];
             $errors = &$context['not_normalizable_value_exceptions'];
             $denormalized = $normalizer->denormalize($data, $type, $format, $context);
+
             if ($errors) {
-                throw new PartialDenormalizationException($denormalized, $errors);
+                // merge errors so that one path has only one error
+                $uniqueErrors = [];
+                foreach ($errors as $i => $error) {
+                    if (isset($uniqueErrors[$error->getPath()])) {
```
nicolas-grekas: shouldn't this be done only when `null !== $error->getPath()`?

src/Symfony/Component/Serializer/Serializer.php
```diff
@@ -228,8 +228,24 @@ public function denormalize($data, string $type, string $format = null, array $c
...
+                        continue;
+                    }
+
+                    if (null !== $error->getPath()) {
+                        $uniqueErrors[$error->getPath()] = $error;
+                        continue;
+                    }
+
+                    $uniqueErrors[$i] = $error;
```
nicolas-grekas: `[]` or `[$i]`?

nicolas-grekas (approved):
LGTM after minor changes

nicolas-grekas:
Thank you @mtarld.

nicolas-grekas:
@mtarld could you please have a look at the current failures on branch 6.3 for Serializer? I need help to resolve them after the merge up :pray:

mtarld:
Sorry @nicolas-grekas I'm in a different timezone, so I hadn't time to take of it.

But the fix provided by @xabbuh in https://github.com/symfony/symfony/pull/52665 is good (it's the one I already had on my local machine)

PR #52584, opened 2023-11-14
[WebProfilerBundle] Fix intercept external redirects

HeahDude:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | ~
| License       | MIT

When intercepting a redirect to an external host the current output gives:

> This request redirects to **http://current-host.orghttp://target-host.org/**

With this PR, it fixes it too:

> This request redirects to **http://target-host.org/**

We could eventually get rid of the first part of the condition, WDYT?

### DIFF
--- a/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/toolbar_redirect.html.twig
+++ b/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/toolbar_redirect.html.twig
@@ -39,10 +39,10 @@
             <div class="block-exception">
                 <h1>Redirection Intercepted</h1>
 
-                {% set absolute_url = host in location ? location : host ~ location %}
+                {% set absolute_url = absolute_url(location) %}
                 <p>This request redirects to <strong>{{ absolute_url }}</strong></p>
 
-                <p><a class="btn" href="{{ location }}">Follow redirect</a></p>
+                <p><a class="btn" href="{{ absolute_url }}">Follow redirect</a></p>
 
                 <p class="sf-redirection-help">
                     The redirect was intercepted by the Symfony Web Debug toolbar to help debugging.


src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/toolbar_redirect.html.twig
```diff
@@ -39,7 +39,7 @@
             <div class="block-exception">
                 <h1>Redirection Intercepted</h1>
 
-                {% set absolute_url = host in location ? location : host ~ location %}
+                {% set absolute_url = host in location or not (location starts with '/') ? location : host ~ location %}
```
nicolas-grekas: `//protocol-relative ` URLs are a thing :) Also, `.` is a hint of a relative URL.

smnandre:
Maybe we could do this computation in the Controller directly ? 

Currently the code look like this (just formatted for clarity)


[WebProfilerBundle/EventListener/WebDebugToolbarListener.php](https://github.com/symfony/symfony/blob/19f52407d9e151e660dc15dc1fb2f2a0ca6af2de/src/Symfony/Bundle/WebProfilerBundle/EventListener/WebDebugToolbarListener.php#L116)

```php
$content = $this->twig->render('@WebProfiler/Profiler/toolbar_redirect.html.twig', [
    'location' => $response->headers->get('Location'),
    'host' => $request->getSchemeAndHttpHost()
]);
$response->setContent($content);
$response->setStatusCode(200);
$response->headers->remove('Location');
```

Something like this could do the job WDYT ? 

```php
$content = $this->twig->render('@WebProfiler/Profiler/toolbar_redirect.html.twig', [
    'location' => $location = $response->headers->get('Location'),
    'absolute_location' => (new UrlHelper($requestStack))->getAbsoluteUrl($location),
]);
$response->setContent($content);
$response->setStatusCode(200);
$response->headers->remove('Location');
```
And that'd allow to keep in one place all those 'Location' manipulation / transformations. 

Because, even with the best intentions, i'm not sure the template code would be ever "obvious" (e g : `host` in the template is in fact "scheme + host" but not `app.request.host` )

WDYT ?

nicolas-grekas:
I agree with @smnandre's proposal, the template can't contain this logic.

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/toolbar_redirect.html.twig
smnandre:
```suggestion
                {% set absolute_url = absolute_url(location) %}
```

Seems to do perfectly the job :) 

Passing UrlHelper into the controller would require a lot of dependency injection, i'm not sure it's worth it as it only would be used to format a link...

HeahDude: Indeed, thanks.

nicolas-grekas (approved):

chalasr (approved):

fabpot (approved):

fabpot:
Thank you @HeahDude.

Issue #48888, opened 2023-01-05
[Mailer] Support OAuth2


nguyenk:
### Description

I'm submitting this issue related to https://github.com/symfony/symfony/issues/35528 (closed)

Starting from the 1st January 2023, Microsoft permanently disabled basic auth for sending emails, and it seems app password will not help anymore. Microsoft's recomendations are : 
 * Modify app to use modern auth, hence implement Oauth 2 support
 * Migrate app to use Graph API and modern auth

Knowing that, may be it make more sense now to consider supporting oauth token - handling  - renewing inside a dedicated mailer component ?

carsonbot:
Thank you for this suggestion.
There has not been a lot of activity here for a while. Would you still like to see this feature?

Purek:
> 
Hi,Do you have a support plan?

pandancode:
*bump* would love to have this as google has also deactivated app passwords for custom domains it seems :cry:

xabbuh:
Someone (could be you) would have to work on a pull request.

Gorkyman:
Any progress on this one maybe?

nguyenk:
Hello @Gorkyman @Purek.

Actually, I ended up switching to the Microsoft Graph API for sending emails. 

Basically I had to :
- Implement a `MicrosoftOAuthMailTransport` (`TransportInterface` that has the 'send' logic)
- Register it using a `MicrosoftOAuthMailTransportFactory` (`AbstractTransportFactory`)
- at the end, I can have a quite standard MAILER_DSN: `microsoft://client_api:client_secret@login.microsoftonline.com?tenant=tenant_id&graph=https://graph.microsoft.com`

If you think it's worse a package, please tell my, I might find some time to do it. Else ping me if you want me to share the code in a gist.

Cheers!

limenet:
Hi @nguyenk, thanks for offering to share your code! Could you please create a Gist with it? That would be very helpful :)

nguyenk:
Hello @limenet here you go ! https://gist.github.com/nguyenk/15dd504dbcf7322807cb8f93371937d7

Any issue please ask

@xabbuh If you think it's worth a PR, I'll be happy to contribute. I think the more logical would be creating a new 3rd party transport : `composer require symfony/microsoft-graph-mailer`

xabbuh:
@nguyenk please do so

carsonbot:
Thank you for this suggestion.
[...]

PR #52585, opened 2023-11-14
[Mailer] Improve OAuth setup

dbu:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

Setting up the Symfony Mailer with XOAuth based authentication is possible but quite complicated. I hacked something together in https://gist.github.com/dbu/3094d7569aebfc94788b164bd7e59acc but with a few changes, I would only need to implement the token provider and configure it.

I guess this is too late for 7.0, so it should go to 7.1. I will adjust the target branch once 7.1 is created.


src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2556,6 +2556,11 @@ private function registerMailerConfiguration(array $config, ContainerBuilder $co
         } else {
             $mailer->replaceArgument(1, $messageBus ? new Reference($messageBus) : new Reference('messenger.default_bus', ContainerInterface::NULL_ON_INVALID_REFERENCE));
         }
+        $authenticators = [];
+        foreach ($config['smtp']['authenticators'] ?? [] as $authenticator) {
+            $authenticators[] = new Reference($authenticator);
+        }
+        $container->getDefinition('mailer.transport_factory.smtp')->setArgument(3, $authenticators);
```
dbu, line 2563: according to `mailer_transport.php` this service is the EsmtpTransportFactory, but the CI tells that we somehow end up setting an argument to "Symfony\Component\Mailer\Mailer" - any idea what i am messing up here?

stof: the weird error message could be because of inlining the transport factory services. It might be inside one of the inlined services inside the mailer service.

src/Symfony/Bundle/FrameworkBundle/Resources/config/schema/symfony-1.0.xsd
```diff
@@ -758,6 +759,12 @@
         </xsd:sequence>
     </xsd:complexType>
 
+    <xsd:complexType name="smtp">
```
stof: I suggest naming this type `mailer_smtp` to avoid potential conflicts with other types in that schema.

dbu: i was wondering - but saw that even more generic things like "header" are not prefixed. but i agree, adding the prefix.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
dbu: is that a bug of the DI container, or should i be doing something differently?

stof: I would not call it a bug. The hard thing is that inline definitions don't have an id. So that's hard to know how to mention them in error messages.

dbu: if it is not a bug, what do i need differently to set the parameter on the correct place?

nicolas-grekas:
Did you consider adding the oauth token providen to core? Could this be done instead of the current extra interface? Is there an approach that wouldn't require adding this extra interface - or at least making `XOAuth2Authenticator` unaware of it? I don't really like the instanceof check there :grimacing:

dbu:
> Did you consider adding the oauth token providen to core? Could this be done instead of the current extra interface? Is there an approach that wouldn't require adding this extra interface - or at least making XOAuth2Authenticator unaware of it?

oauth seems to be more of a concept than an actual standard. every provider is doing it differently. if symfony had a fully fledged oauth component with support for all the major players, we could maybe make this use that. but to me an interface is really the way to go here.

oauth means sending a post with some json body to some url. but which field names to use, whats required or optional and what values to use e.g. for the scope is totally provider dependent.

for office365/exchange, the values happen to be

```
    private const OAUTH_URL = 'https://login.microsoftonline.com/{tenant}/oauth2/v2.0/token';
    private const SCOPE = 'https://outlook.office365.com/.default';
    private const GRANT_TYPE = 'client_credentials';

        $data = [
            'client_id' => $this->clientId,
            'client_secret' => $this->clientSecret,
            'scope' => self::SCOPE,
            'grant_type' => self::GRANT_TYPE,
        ];

    // post that as x-www-form-urlencoded to
    UriTemplate::expand(self::OAUTH_URL, [
            'tenant' => $this->tenant,
    ]));
```

> I don't really like the instanceof check there 😬

i think the null check is required for BC.

nicolas-grekas:
Sorry I wrote "instanceof" but that's inaccurate. I meant this line:
https://github.com/symfony/symfony/pull/52585/files#diff-3ac0365eb0e52c96210c4fc1848acd4be3eb71ceb86aed561b76b8d1d3626b6dR42

src/Symfony/Component/Mailer/Transport/Smtp/Auth/XOAuth2Authenticator.php
```diff
@@ -32,6 +39,7 @@ public function getAuthKeyword(): string
      */
     public function authenticate(EsmtpTransport $client): void
     {
-        $client->executeCommand('AUTH XOAUTH2 '.base64_encode('user='.$client->getUsername()."\1auth=Bearer ".$client->getPassword()."\1\1")."\r\n", [235]);
+        $token = $this->tokenProvider ? $this->tokenProvider->getToken() : $client->getPassword();
```
nicolas-grekas:
```suggestion
        $token = $this->tokenProvider?->getToken() : $client->getPassword();
```

But I'm wondering about this line as now we give two ways to pass that secret - what if both are set for example? I feel like it'd be better to provide only one way to pass that secret - the existing one if possible :) Does that make sense? (it doesn't have to :) )

stof: the issue with the existing one is that oauth tokens tend to be short-lived. In your worker process, you cannot just set the token as the password in the Transport as this will not properly refresh it when it expires.

dbu: changing the client to fetch the password dynamically would be a large refactor, and would mix responsibilities for the auth method.

the thing is that an oauth token can not be statically configured. and in a long running process it might expire before the process ends, so each time we need it we need to be ready to re-fetch it.

we could inject some TokenProviderInterface into the client and use that for getPassword(), but i don't think that would be better.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -2091,6 +2091,15 @@ private function addMailerSection(ArrayNodeDefinition $rootNode, callable $enabl
                                 ->end()
                             ->end()
                         ->end()
+                        ->arrayNode('smtp')
+                            ->fixXmlConfig('authenticator')
```
nicolas-grekas, line 2095: instead of adding a new config entry, I'd suggest passing a tagged_locator in the service definition
we should define the default authenticators as services with low priority so that then adding a new authenticator is just a matter of creating a new class that implements AuthenticatorInterface, thanks to registerForAutoconfiguration.

src/Symfony/Component/Mailer/Transport/Smtp/Auth/AuthTokenProviderInterface.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Symfony\Component\Mailer\Transport\Smtp\Auth;
+
+/**
+ * The auth token provider knows how to create a valid token for the XOAuth2Authenticator.
+ *
+ * Usually with OAuth, you will need to do some web request to fetch the token.
+ * You also want to cache the token for as long as it is valid.
+ */
+interface AuthTokenProviderInterface
```
nicolas-grekas, line 20: TokenProviderInterface?

src/Symfony/Component/Mailer/Transport/Smtp/EsmtpTransportFactory.php
```diff
@@ -11,23 +11,41 @@
...
 {
+    /**
+     * @var AuthenticatorInterface[]
+     */
+    private ?array $authenticators;
+
+    /**
+     * @param AuthenticatorInterface[]|null $authenticators
+     */
+    public function __construct(EventDispatcherInterface $dispatcher = null, HttpClientInterface $client = null, LoggerInterface $logger = null, array $authenticators = null)
```
nicolas-grekas, line 36: I'm not sure we need the new argument: calling setAuthenticators on the resulting instance should be enough

nicolas-grekas (commented):
Got it for the geToken??getPassword logic, makes sense to me now.
Here are some suggestions to help move forward.

nicolas-grekas:
Note that it'd be cool to contribute the office365 bridge while at it :)

nicolas-grekas:
Oh and maybe also an AbstractOauthTokenProvider?

dbu:
thanks for the feedback. yeah that sounds good. i will try to find time, might be after christmas only :see_no_evil: 

the office365 provider would go into a new Office365 namespace at https://github.com/symfony/symfony/tree/7.1/src/Symfony/Component/Mailer/Bridge and need a composer.json for the subtree split, right?

i will see if an abstract base class makes sense that could be shared in the messenger package.

nicolas-grekas:
> the office365 provider would go into a new Office365 namespace

yes!

pounard:
Hello, @dbu @nicolas-grekas, I stumbled here while doing some research, because I am since a few days working a new component proposal, that does exactly the job of the introduced here `AuthTokenProvider`, and more. Would you be interested in participating instead of adding this new interface in this PR ?

dbu:
i do not have much time at the moment, so more than glad if you put some brain energy into proposing a nice architecture for the auth token system. i can try to give my input on the proposal when you have a draft ready, and will be happy to rebase this PR when/if your proposal gets merged.

pounard:
@dbu thanks, I will in few days I hope, open an issue for it.

pounard:
@dbu if you have a few minutes to spare, I'd very much like to have your sentiment about https://github.com/symfony/symfony/pull/54013, thanks in advance !

pounard:
> Did you consider adding the oauth token providen to core?

@nicolas-grekas I did ! Please see my PR https://github.com/symfony/symfony/pull/54013 - your insights would be invaluable !

solverat:
@dbu: Is there any update on this? Microsoft will disable Basic Auth in 6 months and will only support OAuth as the new standard. I believe this PR could therefore be a valuable addition to the Symfony ecosystem. Is there any way this can be supported?

dbu:
i was hoping that #54013 would provide a clean solution for most of this. the problem with this PR we are discussing here is that it introduces a lot of general concepts but only for the specific scope of microsoft things.

if somebody wants to pick this up and try to get it mergeable, i am happy to provide feedback, clarify things, check how our application does things. i don't really have the time to finish it myself however.

pounard:
> i was hoping that #54013 would provide a clean solution for most of this. the problem with this PR we are discussing here is that it introduces a lot of general concepts but only for the specific scope of microsoft things.

I'm still alive! I had a rough year. Now I'm getting back in business, I'll try to find some time at work to rebase the https://github.com/symfony/symfony/pull/54013 PR, please note it should also fix #48888 if that happens.

edwasue-cmyk (approved):

nicolas-grekas:
Closing this one, and thanks for the two years of patience on it.

The reason is the one you named yourself in September: this PR introduces general concepts for a specific scope, and #54013 is trying to solve the general problem. That decision should come before any more work on the plumbing, and it is not a decision this PR can make.

For the record, since the PR page currently suggests otherwise. The `Status: Reviewed` label comes from a single approval with an empty body from an account with no association to the project, and carsonbot flipped the label three seconds later. Every other review here is a `COMMENTED` one from stof, you or me, and the last substantive round is from November 2023. The label is not a maintainer judgement.

The head also does not parse:

```php
$token = $this->tokenProvider?->getToken() : $client->getPassword();
```

```
PHP Parse error: syntax error, unexpected token ":"
```

That is my fault: it came from a suggestion I wrote in review 1399204662 where a `?->` replaced the `?` of the ternary and left the `:` orphaned, and it went in through the "Commit suggestion" button. CI has never run on that commit, `check-runs` returns zero, so nothing ever caught it. The fix is `?? $client->getPassword()`, for whoever picks this up.

One substantive finding worth carrying forward, because it would bite any version of this. `registerMailerConfiguration()` sets argument 3 unconditionally, so with no `smtp` config it becomes `[]`, and `EsmtpTransport` only falls back to its four default authenticators when that argument is `null`:

```
base:  CRAM-MD5,PLAIN,LOGIN,XOAUTH2
patch: <EMPTY>
```

Every mailer-enabled application with an authenticated SMTP DSN would fail on its first send. The `tagged_locator` approach I suggested in 1411214064, with the defaults registered as low-priority services, avoids this by construction, which is the main reason I would still go that way.

If #54013 lands and a thin XOAUTH2 token provider is still wanted on top of it, please open a fresh PR and I will review it properly.

PR #52589, opened 2023-11-15
[Serializer] Fix XML attributes not added on empty node

### LINKED ISSUE #52385: [Serializer] XmlEncoder does not parse namespace for node without children

6.3

### Description

When using `XmlEncoder::decode` on a node that does not have children and the node has a namespace defined then the namespace is not in the array even if it is needed for attributes on that node.
This results in an invalid xml when encoding it again to XML.

In the following example on how to reproduce there is a namespaced attribute `xsi:noNamespaceSchemaLocation` but after reading the xml and writing it again this is now invalid xml and throws a warning when trying to read it:
`PHP Warning:  XMLReader::read(): test.xml:2: namespace error : Namespace prefix xsi for noNamespaceSchemaLocation on property is not defined`

### How to reproduce

```php
$encoder = new XmlEncoder();

$data = $encoder->decode(
    <<<'XML'
<?xml version="1.0" encoding="UTF-8"?>
<property xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://domain.com/path/to/schema.xsd" modification_timestamp="2009-11-10T10:52:16.0Z"></property>
XML,
    'array'
);

var_export($data);
// array (
//  '@xsi:noNamespaceSchemaLocation' => 'http://domain.com/path/to/schema.xsd',
//  '@modification_timestamp' => '2009-11-10T10:52:16.0Z',
//  '#' => '',
//)

echo $encode->encode($data, 'xml');
// <response xsi:noNamespaceSchemaLocation="http://domain.com/path/to/schema.xsd" modification_timestamp="2009-11-10T10:52:16.0Z"></response>
```

### Possible Solution

The method `XmlEncoder::parseXmlAttributes` could check if an attribute has a namespace and at least fetch this namespace from the `$node` like it is done in `XmlEncoder::decode` in case the root node has child nodes:
```
foreach ($xpath->query('namespace::*', $dom->documentElement) as $nsNode) {
    $data['@'.$nsNode->nodeName] = $nsNode->nodeValue;
}
```

### Additional Context

_No response_

mtarld:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52385
| License       | MIT

Add XML namespace attributes on node whether they have chlidren or not.

### DIFF
--- a/src/Symfony/Component/Serializer/Tests/Encoder/XmlEncoderTest.php
+++ b/src/Symfony/Component/Serializer/Tests/Encoder/XmlEncoderTest.php
@@ -450,6 +450,17 @@ XML;
         $array = $this->getNamespacedArray();
 
         $this->assertEquals($array, $this->encoder->decode($source, 'xml'));
+
+        $source = '<?xml version="1.0"?>'."\n".
+            '<response xmlns="http://www.w3.org/2005/Atom" xmlns:app="http://www.w3.org/2007/app" app:foo="bar">'.
+            '</response>'."\n";
+
+        $this->assertEquals([
+            '@xmlns' => 'http://www.w3.org/2005/Atom',
+            '@xmlns:app' => 'http://www.w3.org/2007/app',
+            '@app:foo' => 'bar',
+            '#' => '',
+        ], $this->encoder->decode($source, 'xml'));
     }
 
     public function testDecodeScalarWithAttribute()
--- a/src/Symfony/Component/Serializer/Encoder/XmlEncoder.php
+++ b/src/Symfony/Component/Serializer/Encoder/XmlEncoder.php
@@ -140,26 +140,22 @@ class XmlEncoder implements EncoderInterface, DecoderInterface, NormalizationAwa
         // todo: throw an exception if the root node name is not correctly configured (bc)
 
         if ($rootNode->hasChildNodes()) {
-            $xpath = new \DOMXPath($dom);
-            $data = [];
-            foreach ($xpath->query('namespace::*', $dom->documentElement) as $nsNode) {
-                $data['@'.$nsNode->nodeName] = $nsNode->nodeValue;
+            $data = $this->parseXml($rootNode, $context);
+            if (\is_array($data)) {
+                $data = $this->addXmlNamespaces($data, $rootNode, $dom);
             }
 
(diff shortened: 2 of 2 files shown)


mtarld:
fabbot errors are not related to that PR.

src/Symfony/Component/Serializer/Tests/Encoder/XmlEncoderTest.php
```diff
@@ -797,6 +802,23 @@ protected function getNamespacedArray()
...
     }
 
+    protected function getEmptyNamespacedXmlSource()
+    {
+        return '<?xml version="1.0"?>'."\n".
+            '<response xmlns="http://www.w3.org/2005/Atom" xmlns:app="http://www.w3.org/2007/app" app:foo="bar">'.
+            '</response>'."\n";
+    }
+
+    protected function getEmptyNamespacedArray()
```
nicolas-grekas: let's remove those two methods and inline them?

nicolas-grekas (approved):

nicolas-grekas:
@mtarld can you please have a look at #52401 and help us decide which approach we should merge? It looks like the test cases are the same, right?

mtarld:
I think that there is an issue with https://github.com/symfony/symfony/pull/52401. 

Indeed, the following XML:
```xml
<?xml version="1.0"?>
<response xmlns="http://www.w3.org/2005/Atom" xmlns:app="http://www.w3.org/2007/app" app:foo="bar">
</response>
```
will be decoded as:
```php
[
    '@app:foo' => 'bar',
    '#' => '',
]
```
whereas it should in my opinion be decoded as:
```php
[
    '@xmlns' => 'http://www.w3.org/2005/Atom',
    '@xmlns:app' => 'http://www.w3.org/2007/app',
    '@app:foo' => 'bar',
    '#' => '',
]
```
otherwise we lose the XML namespaces information.

But this is valid only if we do want to care about XML namespaces, of course.

nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @mtarld.

PR #40964, opened 2021-04-27
[HttpFoundation] Fixes for PHP 8.1 deprecations

jrmajor:
| Q             | A
| ------------- | ---
| Branch?       | 4.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

Fixes passing `null` to non-nullable arguments.

PR #52593, opened 2023-11-15
[HttpFoundation] Fix/parameterbag get method

wandesnet:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no
| Issues        | 
| License       | MIT

This pull request replaces the **array_key_exists** check with the **null coalescing operator**. The change aims to improve the robustness and consistency of the code.

Problem:

Using **array_key_exists** in this specific context results in a problem identified in the **basename()** function. When encountering a null value for parameter ($path), the function call becomes obsolete, generating the following warning message:

`basename(): Passing null to parameter ($path) of type string is deprecated in Request.php`

This is because **array_key_exists** returns true when an array's key matches a null value, which leads to deprecation of the **basename()** function.

Solution:

Replacing the null navigation **operator (??)** provides a safer approach by properly handling null values. This solves the problem with the **basename()** function and contributes to improving the overall stability of the code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ composer.phar
 package.tar
 /packages.json
 /.phpunit
+.idea/
--- a/src/Symfony/Component/HttpFoundation/ParameterBag.php
+++ b/src/Symfony/Component/HttpFoundation/ParameterBag.php
@@ -89,7 +89,7 @@ class ParameterBag implements \IteratorAggregate, \Countable
      */
     public function get(string $key, $default = null)
     {
-        return \array_key_exists($key, $this->parameters) ? $this->parameters[$key] : $default;
+        return $this->parameters[$key] ?? $default;
     }
 
     /**


xabbuh:
This change breaks backwards compatibility (see the failing tests).

wandesnet:
> This change breaks backwards compatibility (see the failing tests).

I understand, I noticed that it breaks in all versions, an alternative would be to force the retention type to (string) of the ->get() method when calling it, that is, in basename() in the place where this bug is occurring.
Example:

`basename((string) $this->server->get('SCRIPT_NAME', ''))` file Request.php on line 1802

.gitignore
```diff
@@ -8,3 +8,4 @@ composer.phar
 package.tar
 /packages.json
 /.phpunit
+.idea/
```
ghost, line 11: Files that are not releated to the project should not be commited here but ignored in `.git/info/exclude` .

ghost (changes requested):
I'm sorry but I don't understand your issue. 

Do you get a deprecation notice from `basename` function because `ParameterBag:get()` returns null? 
Then this is on your side to code against such a case (e.g. add if statement)

wandesnet:
> basename(): Passing null to parameter ($path) of type string is deprecated in Request.php

This is not a critical bug, but a warning message is appearing in my application log, `basename(): Passing null to parameter ($path) of type string is deprecated in Request.php.`

This happens because the **basename()** function expects to receive a string as a parameter, but is receiving null.

.gitignore
wandesnet: > Files that are not releated to the project should not be commited here but ignored in `.git/info/exclude` .

I didn't know, thanks for letting me know.

xabbuh:
I would first investigate why that value is `null`. Can you create a small example application that allows to reproduce?

stof:
If your parameter bag contains an explicit `null` value, I find it totally expected that you get a `null` value when reading it. And our tests even ensure that this behavior happens so changing that would be a BC break.

The fact that your own logic does not properly handle `null` values that it receives is not a bug in ParameterBag.

ghost:
~The only line where `Request` class on the current 5.4 branch calls `basename()` with a parameter that can possibly be null is:~
https://github.com/symfony/symfony/blob/4be970654b4a049374318c06f83cf4a80f9a3a80/src/Symfony/Component/HttpFoundation/Request.php#L1932

~However in 6.4 it's fine:~
https://github.com/symfony/symfony/blame/de3d115496c0ab76d9849896e9d4c8a15b91eb1f/src/Symfony/Component/HttpFoundation/Request.php#L1800

~but this was fixed in v4.4.42.~

~Weird...~

ghost:
Got it:

on 5.4 branch https://github.com/symfony/symfony/blame/4be970654b4a049374318c06f83cf4a80f9a3a80/src/Symfony/Component/HttpFoundation/Request.php#L1932

on 6.4 the same.
I think this line was unintentionally lost in https://github.com/symfony/symfony/pull/40964

wandesnet:
To provide context, the warning message in question relates to the following situation in the Request.php file, specifically on line 1802:

`if (basename($this->server->get('SCRIPT_NAME', '')) === $filename)` 

The basename() function expects to receive a string, correct?

When calling `$this->server->get('SCRIPT_NAME', '')`, if the **SCRIPT_NAME** key does not exist, it returns an empty string, correct?

However, a problem arises if the **SCRIPT_NAME** key exists and the value is null. The **array_key_exists** function returns true, indicating that the key exists, even if its value is null. As a result, the error warning occurs while executing the **basename()** function.

To resolve this issue, I tried to fix the implementation by replacing **array_key_exists** with the null navigation **operator (??)**. With this change, whenever the key contains a null value, the second parameter passed to the get() function will be returned, providing a safer solution and avoiding the aforementioned error warning.

ghost:
It's weird that SCRIPT_NAME is set and is null. How to you run your PHP server?

nicolas-grekas:
:-1: with the proposed change, the code is on purpose - parameters can be null and shouldn't fallback in that case

> It's weird that SCRIPT_NAME is set and is null. How to you run your PHP server?

that's the important question for the next step

xabbuh:
I am going to close here for now due to the lack of feedback. Please let us know when you have more information and we can consider to reopen.

PR #47259, opened 2022-08-12
[ProgressBar] Empty iterable throws Exception on "maximum number of steps is not set"

freimair:
| Q             | A
| ------------- | ---
| Branch?       | 6.2 
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #47244 
| License       | MIT
| Doc PR        |  

**Bug Description**
When using the progress bar with an empty input and time estimates, an error has been thrown which stopped the execution:

```
            $bar = new ProgressBar($output = $this->getOutputStream());
            $bar->setFormat("%elapsed%/%estimated%");

            $input = [];
            foreach ($bar->iterate($input) as $item)
                var_dump($item);
```

I dug around in the code and found a semantically incorrect check for `getMaxSteps()` that evaluated a maxSteps of 0 (because of an empty input) to be interpreted like maxSteps has not been set at all und thus, triggering

```
                if (!$bar->getMaxSteps())) {
                    throw new LogicException('Unable to display the .* time if the maximum number of steps is not set.');
                }
```

**Fix Proposal**

I propose to change the check to be semantically correct as in "0 is a valid value of maxSteps, null is not".

More details and a bit of discussion in #47244.

PR #52605, opened 2023-11-15
[Console] Support `ProgressBar::iterate()` on empty array

### LINKED ISSUE #47244: [ProgressBar] Empty iterable throws Exception on "maximum number of steps is not set"

latest

### Description

I stumbled upon an empty array that has been used with a progress bar. And my execution stopped. Turns out the ProgressBar checks if its max steps are set when "remaining" and "estimated" values are used in formatting. Nothing wrong so far.

However, in case there is an empty array supplied to the progress bar, code internals interpret a max value of 0 as the max value not being set:
```
                if (!$bar->getMaxSteps()) {
                    throw new LogicException('Unable to display the ... if the maximum number of steps is not set.');
                }
```
(https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Console/Helper/ProgressBar.php#L522 and https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Console/Helper/ProgressBar.php#L529)

That to me seems not entirely accurate. Because in this very case, the max value IS set and it IS 0. Yet, it is treated like it is NOT set.

Expected:
I would expect the progress bar not to fail if an empty Iterable is presented to it.

### How to reproduce

```
$progressBar = new ProgressBar(new ConsoleOutput(OutputInterface::VERBOSITY_DEBUG);
$progressBar->setFormat('debug');

// this very array sources somewhere in a bigger project and it just happens to be empty some times
$incoming = array();

foreach($progressBar->iterate($incoming) as $current) {
   var_dump($current);
}
```

### Possible Solution

Changing the line mentioned above:
```
                if (!$bar->getMaxSteps()) {
                    throw new LogicException('Unable to display the ... if the maximum number of steps is not set.');
                }
```
(https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Console/Helper/ProgressBar.php#L522 and https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Console/Helper/ProgressBar.php#L529)

to

```
                if (!isset($bar->max)) {
                    throw new LogicException('Unable to display the ... if the maximum number of steps is not set.');
                }
```

seem to be semantically more correct. And it makes the ProgressBar behave like I expected it to do. If you guys are ok with this approach, I can provider a PR.

### Additional Context

_No response_

Thanks!

GromNaN:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #47244
| License       | MIT

Alternative to #47259

Use `$max = null` to indicate that the value is unknown. This allows `0` to be displayed by directly setting the progress bar to 100%.

Zero is only supported for `iterate()`. When passed to the constructor or `setMaxSteps`, it means "unknown max".

### DIFF
--- a/src/Symfony/Component/Console/Helper/ProgressBar.php
+++ b/src/Symfony/Component/Console/Helper/ProgressBar.php
@@ -195,7 +195,7 @@ final class ProgressBar
 
     public function getMaxSteps(): int
     {
-        return $this->max;
+        return $this->max ?? 0;
     }
 
     public function getProgress(): int
@@ -215,7 +215,7 @@ final class ProgressBar
 
     public function getBarOffset(): float
     {
-        return floor($this->max ? $this->percent * $this->barWidth : (null === $this->redrawFreq ? (int) (min(5, $this->barWidth / 15) * $this->writeCount) : $this->step) % $this->barWidth);
+        return floor(null !== $this->max ? $this->percent * $this->barWidth : (null === $this->redrawFreq ? (int) (min(5, $this->barWidth / 15) * $this->writeCount) : $this->step) % $this->barWidth);
     }
 
     public function getEstimated(): float
@@ -253,7 +253,7 @@ final class ProgressBar
 
     public function getBarCharacter(): string
     {
-        return $this->barChar ?? ($this->max ? '=' : $this->emptyBarChar);
+        return $this->barChar ?? (null !== $this->max ? '=' : $this->emptyBarChar);
     }
 
     public function setEmptyBarCharacter(string $char): void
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/Console/Helper/ProgressBar.php
```diff
@@ -195,7 +195,7 @@ public function getStartTime(): int
 
     public function getMaxSteps(): int
     {
-        return $this->max;
+        return $this->max ?? 0;
```
GromNaN, line 198: `$this->max` can be `null`, but keep returning `0` for backward compatibility.

GromNaN (commented):
We could deprecate passing `0` for unknown max, but that's not necessary to fix the reported issue.

I create a branch with the deprecation, that we can discuss for 7.1: https://github.com/symfony/symfony/compare/6.4...GromNaN:symfony:issue-47259-deprecate

maxbeckers (approved):

fabpot (approved):

fabpot:
Merging in 7.1 as this is a behavior change.

fabpot:
Thank you @GromNaN.

PR #52612, opened 2023-11-16
[Validator] Luhn for numbers

MrYamous:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | minor text changes
| New feature?  | no
| Deprecations? | no
| Issues        | Not related to existing issue
| License       | MIT

Credit card numbers are a common use of the Luhn formula, but it can also be used for other things: IMEI number, EU patent application number, various numbers in some countries...

The implementation of this validator seems to have been made for use with card numbers (#4734) but I think the default error message and comments in the code should be independent of that as this validator is about Luhn and not only card numbers.

PR for doc will also be needed if accepted, i will work on it. 

What do you think about it ?

### DIFF
--- a/src/Symfony/Component/Validator/Constraints/Luhn.php
+++ b/src/Symfony/Component/Validator/Constraints/Luhn.php
@@ -34,7 +34,7 @@ class Luhn extends Constraint
         self::CHECKSUM_FAILED_ERROR => 'CHECKSUM_FAILED_ERROR',
     ];
 
-    public $message = 'Invalid card number.';
+    public $message = 'This value is not a valid number for Lohn formula.';
 
     public function __construct(
         array $options = null,
--- a/src/Symfony/Component/Validator/Constraints/LuhnValidator.php
+++ b/src/Symfony/Component/Validator/Constraints/LuhnValidator.php
@@ -19,7 +19,7 @@ use Symfony\Component\Validator\Exception\UnexpectedValueException;
 /**
  * Validates a PAN using the LUHN Algorithm.
  *
- * For a list of example card numbers that are used to test this
+ * For a list of example numbers that are used to test this
  * class, please see the LuhnValidatorTest class.
  *
  * @see    http://en.wikipedia.org/wiki/Luhn_algorithm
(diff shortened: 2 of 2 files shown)


nicolas-grekas (commented):
This change would break existing translations. I don't think we should do it.
If the vocabulary doesn't fit some use case, translations should be used instead IMHO.

src/Symfony/Component/Validator/Constraints/Luhn.php
```diff
@@ -34,7 +34,7 @@ class Luhn extends Constraint
         self::CHECKSUM_FAILED_ERROR => 'CHECKSUM_FAILED_ERROR',
     ];
 
-    public $message = 'Invalid card number.';
+    public $message = 'This value is not a valid number for Lohn formula.';
```
alexisLefebvre, line 37:
```suggestion
    public $message = 'This value is not a valid number for Luhn formula.';
```

stof:
I would say that if you use it for a different purpose, you should use a custom message. But I don't see any case where I would like to report a message saying `This value is not a valid number for Luhn formula.` to users (they probably don't know what the Luhn formula is).

MrYamous:
> I would say that if you use it for a different purpose, you should use a custom message. But I don't see any case where I would like to report a message saying `This value is not a valid number for Luhn formula.` to users (they probably don't know what the Luhn formula is).

That's a good point, user probably dont know that, a more generic message could be better.

But i think this constraint is for Luhn formula, not card number, imo the default message should not be designed for an usage.
If we use this constraint for an IMEI number, we probably don't want "this card number is invalid" as message

stof:
@MrYamous changing the default message will change existing projects, which is a bad idea (and even worse when doing that in a patch release). This would require going through an opt-in period if we were doing that. But I'm not sure this case would justify a deprecation + BC break (that would be effective in 2 years as it is far too late to have the new default in 7.0). And that would first require finding the replacement message anyway in the default constraint.

Thus, there is already a way for projects to provide a good error message when they use the Luhn constraint for something else than a card number. So projects that care about that probably already have their own more detailed error message (specific to their use case, so for instance saying `This is not a valid IMEI number`).

MrYamous:
I obviously didn't realize this change would BC break. I'm fine with leaving the message as it is

Issue #52607, opened 2023-11-16
[DependencyInjection] Allow multiple configurators


maelanleborgne:
### Description

Would you be open to add a `configurators` or `configurator_stack` option to the service definition to allow multiple configurators for a service ? Maybe using a priority property to resolve processing order ? 

(I stumbled upon this while making a #[WithConfigurator] attribute and wondering how to resolve conflicts)

### Example

```yaml
services:
  App\MyService:
    configurator_stack:
      - ['@App\MyConfigurator1']
      - ['@App\MyConfigurator2', 'methodName']
      - ['@App\MyConfigurator3', 'methodName', 10]
```

GromNaN:
I had the same need recently: I need to use a configurator to add monitoring features to a service, only when the profiler is enabled, but I want the developers to be able to use the configurator for this service if they need.

So I'm all for it.

```php
class Symfony\Components\DependencyInjection\Definition {
    // Add a configurator to the list
    public function addConfigurator(string|array|Reference $configurator): void;

    // Set the full list of configurators
    public function setConfigurators(array $configurators): void;

    // Get the full list of configurators
    public function getConfigurators(): array
}
```

Thinking about the backward compatibility, the easiest would be to the single configurator and the new list of configurators separated in the `Definition` class. And deprecate the existing `setDefinition` method.

Unsetting the configurator (`setConfigurator(null)`) would be done by setting an empty list of configurators.

GromNaN:
Do you have a use-case for the priority?

maelanleborgne:
Not really, but it could be as simple as :
One of the configurators (1) initializes a logger and another configurator (2) manipulate the service in ways that produce logs, so we may want the configurator 1 to be called before configurator 2.

nicolas-grekas:
How would AsConfigurator work?

GromNaN:
> How would AsConfigurator work?

I don't think this attribute already exist. Based on other DI attributes, it would be like this:

```php
#[AsConfigurator(service = MyService::class, method = '__invoke', priority = 0)]
class MyServiceConfigurator
{
    public function __invoke(MyService $service): void
    {
        $service->doSomething();
    }
}
```

PR #52615, opened 2023-11-16
[DependencyInjection] Allow multiple configurators

maelanleborgne:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Issues        | #52607 
| License       | MIT

This feature allows the use of multiple configurators for a single service.  
A new service definition option is introduced : `configurators`.  
The previous `configurator`, as well as  the`Defintion::getConfigurator` and `Defintion::setConfigurator` that now trigger deprecations and make use of the new `Defintion::getConfigurators`,  `Defintion::setConfigurators` and `Defintion::setConfigurators`.  

```yaml
services:
  App\MyService:
    configurators:
      - ['@App\MyConfigurator1']
      - ['@App\MyConfigurator2', 'methodName']
      - ['@App\MyConfigurator3', 'methodName']
```

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Compiler/AnalyzeServiceReferencesPass.php
+++ b/src/Symfony/Component/DependencyInjection/Compiler/AnalyzeServiceReferencesPass.php
@@ -185,7 +185,7 @@ class AnalyzeServiceReferencesPass extends AbstractRecursivePass
         if (!$this->onlyConstructorArguments) {
             $this->processValue($properties);
             $this->processValue($setters);
-            array_map([$this, 'processValue'], $value->getConfigurators());
+            array_map($this->processValue(...), $value->getConfigurators());
         }
         $this->lazy = $lazy;
 
--- a/src/Symfony/Component/DependencyInjection/Compiler/AnalyzeServiceReferencesPass.php
+++ b/src/Symfony/Component/DependencyInjection/Compiler/AnalyzeServiceReferencesPass.php
@@ -185,7 +185,7 @@ class AnalyzeServiceReferencesPass extends AbstractRecursivePass
         if (!$this->onlyConstructorArguments) {
             $this->processValue($properties);
             $this->processValue($setters);
-            $this->processValue($value->getConfigurator());
+            array_map([$this, 'processValue'], $value->getConfigurators());
         }
         $this->lazy = $lazy;
 
--- a/src/Symfony/Component/DependencyInjection/ContainerBuilder.php
+++ b/src/Symfony/Component/DependencyInjection/ContainerBuilder.php
@@ -1156,7 +1156,18 @@ class ContainerBuilder extends Container implements TaggedContainerInterface
                 }
 
                 if (!\is_callable($callable)) {
-                    throw new InvalidArgumentException(sprintf('The configure callable for class "%s" is not a callable.', get_debug_type($service)));
+                    if (\is_object($callable)) {
+                        $stringConfigurator = \get_class($callable);
+                    } elseif (\is_array($callable)) {
+                        if (\is_object($callable[0]))
+                        {
+                            $callable[0] = \get_class($callable[0]);
+                        }
+                        $stringConfigurator = implode('::', array_splice($callable, 0, 2));
+                    } else {
+                        $stringConfigurator = $callable;
+                    }
+                    throw new InvalidArgumentException(sprintf('The configure callable "%s" for class "%s" is not a callable.', $stringConfigurator, get_debug_type($service)));
                 }
 
                 $callable($service);
--- a/src/Symfony/Component/DependencyInjection/ContainerBuilder.php
+++ b/src/Symfony/Component/DependencyInjection/ContainerBuilder.php
@@ -1143,22 +1143,24 @@ class ContainerBuilder extends Container implements TaggedContainerInterface
             }
         }
 
-        if ($callable = $definition->getConfigurator()) {
-            if (\is_array($callable)) {
-                $callable[0] = $parameterBag->resolveValue($callable[0]);
+        foreach ($definition->getConfigurators() as $configurator) {
+            if ($callable = $configurator) {
+                if (\is_array($callable)) {
+                    $callable[0] = $parameterBag->resolveValue($callable[0]);
 
-                if ($callable[0] instanceof Reference) {
-                    $callable[0] = $this->doGet((string) $callable[0], $callable[0]->getInvalidBehavior(), $inlineServices);
-                } elseif ($callable[0] instanceof Definition) {
-                    $callable[0] = $this->createService($callable[0], $inlineServices);
+                    if ($callable[0] instanceof Reference) {
+                        $callable[0] = $this->doGet((string)$callable[0], $callable[0]->getInvalidBehavior(), $inlineServices);
+                    } elseif ($callable[0] instanceof Definition) {
+                        $callable[0] = $this->createService($callable[0], $inlineServices);
+                    }
                 }
-            }
 
-            if (!\is_callable($callable)) {
-                throw new InvalidArgumentException(sprintf('The configure callable for class "%s" is not a callable.', get_debug_type($service)));
-            }
+                if (!\is_callable($callable)) {
+                    throw new InvalidArgumentException(sprintf('The configure callable for class "%s" is not a callable.', get_debug_type($service)));
+                }
 
-            $callable($service);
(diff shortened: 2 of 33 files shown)


src/Symfony/Component/DependencyInjection/ContainerBuilder.php
```diff
@@ -1143,22 +1143,24 @@ private function createService(Definition $definition, array &$inlineServices, b
...
+                        $callable[0] = $this->createService($callable[0], $inlineServices);
+                    }
                 }
-            }
 
-            if (!\is_callable($callable)) {
-                throw new InvalidArgumentException(sprintf('The configure callable for class "%s" is not a callable.', get_debug_type($service)));
-            }
+                if (!\is_callable($callable)) {
+                    throw new InvalidArgumentException(sprintf('The configure callable for class "%s" is not a callable.', get_debug_type($service)));
```
GromNaN: As there may be several configurators, this message should be modified to indicate which one is invalid, by formatting the `$configurator` variable.

src/Symfony/Component/DependencyInjection/Definition.php
```diff
@@ -694,28 +695,73 @@ public function getDeprecation(string $id): array
...
+            $this->configurators[] = $configurator;
+        }
+
+        return $this;
+    }
+
+    /**
+     * Sets the configurator list to call after the service is fully initialized.
+     *
+     * @param array $configurators A PHP function, reference or an array containing a class/Reference and a method to call
```
GromNaN:
```suggestion
     * @param list<callable|Reference|array{0:class-string|Reference,1:string}> $configurators
```

src/Symfony/Component/DependencyInjection/Definition.php
```diff
@@ -694,28 +695,73 @@ public function getDeprecation(string $id): array
...
     {
-        $this->changes['configurator'] = true;
+        trigger_deprecation('symfony/dependency-injection', '7.1', 'Method "%s()" is deprecated, use "%s()" or "%s()" instead.', __METHOD__, 'setConfigurators', 'addConfigurator');
+        return $this->setConfigurators([$configurator]);
+    }
+
+    /**
+     * Add a configurator to the configurator list.
+     *
+     * @param string|array|Reference|null $configurator A PHP function, reference or an array containing a class/Reference and a method to call
```
GromNaN: `null` isn't valid. It was used to unset the configurator.

You can describe the array shape:
```suggestion
     * @param callable|Reference|array{0:class-string|Reference,1:string} $configurators A PHP function, reference or an array containing a class/Reference and a method to call
```

src/Symfony/Component/DependencyInjection/Definition.php
```diff
@@ -694,28 +695,73 @@ public function getDeprecation(string $id): array
...
     {
-        return $this->configurator;
+        trigger_deprecation('symfony/dependency-injection', '7.1', 'Method "%s()" is deprecated, use "%s()" instead.', __METHOD__, 'getConfigurators');
+        return $this->configurators[0] ?? null;
+    }
+
+    /**
+     * Gets the configurators list after the service is fully initialized.
+     *
+     * @return array<array|callable>
```
GromNaN:
```suggestion
     * @return list<callable|array{0:class-string|Reference,1:string}>
```

src/Symfony/Component/DependencyInjection/Loader/YamlFileLoader.php
```diff
@@ -542,9 +545,19 @@ private function parseDefinition(string $id, array|string|null $service, string
         }
 
         if (isset($service['configurator'])) {
-            $definition->setConfigurator($this->parseCallable($service['configurator'], 'configurator', $id, $file));
+            $definition->addConfigurator($this->parseCallable($service['configurator'], 'configurator', $id, $file));
```
GromNaN, line 549: Have you added a depreciation message when this attribute is used? I think we should: having both `configurator` and `configurators` is confusing.

Not needed for other configuration formats as the name don't have to be pluralized in XML or PHP.

GromNaN:
The parameter `$configurator` in `Autoconfigure` attribute should be deprecated and replaced by `$configurators`. https://github.com/symfony/symfony/blob/4e944bed01e6627a4d10cf071b118fb7c7cef6de/src/Symfony/Component/DependencyInjection/Attribute/Autoconfigure.php#L31

src/Symfony/Component/DependencyInjection/Compiler/AnalyzeServiceReferencesPass.php
```diff
@@ -185,7 +185,7 @@ protected function processValue(mixed $value, bool $isRoot = false): mixed
         if (!$this->onlyConstructorArguments) {
             $this->processValue($properties);
             $this->processValue($setters);
-            $this->processValue($value->getConfigurator());
+            array_map([$this, 'processValue'], $value->getConfigurators());
```
GromNaN:
```suggestion
            array_map($this->processValue(...), $value->getConfigurators());
```

src/Symfony/Component/DependencyInjection/Definition.php
```diff
@@ -694,28 +695,73 @@ public function getDeprecation(string $id): array
      * @param string|array|Reference|null $configurator A PHP function, reference or an array containing a class/Reference and a method to call
      *
      * @return $this
+     * @deprecated since Symfony 7.1 and will be removed in 8.0, use addConfigurator() or setConfigurators() instead
      */
     public function setConfigurator(string|array|Reference|null $configurator): static
     {
-        $this->changes['configurator'] = true;
+        trigger_deprecation('symfony/dependency-injection', '7.1', 'Method "%s()" is deprecated, use "%s()" or "%s()" instead.', __METHOD__, 'setConfigurators', 'addConfigurator');
+        return $this->setConfigurators([$configurator]);
```
GromNaN: If `$configurator` is null, set an empty array.

nicolas-grekas:
Personally, I am not at all convinced we need this. This would create a way to alter how a service is defined in a completely unexpected location, as in: I change something on one side and it impacts another side without any obvious link (from the pov of the affected service). In my opinion, a configurator should be configured where the service itself is defined.

nicolas-grekas:
Just to clarify, I mean I'm not convinced we need an AsConfigurator attribute. Since adding that attribute is the motivation behind this PR, it means I'm not convinced we need this PR at all.

stof:
I also don't think we should have a `AsConfigurator` where one class decides to become a configurator for other services. The configurator should be defined by the service, not the other way around.

If you want a declarative way to define the configurator of a service, I would rather have a `#[WithConfigurator]` attribute applied on the service.

maelanleborgne:
> Personally, I am not at all convinced we need this. This would create a way to alter how a service is defined in a completely unexpected location, as in: I change something on one side and it impacts another side without any obvious link (from the pov of the affected service). In my opinion, a configurator should be configured where the service itself is defined.

My bad I made a mistake writing the original issue, the attribute was #[WithConfigurator]. But anyway, that will be a concern for another PR :)

This PR is only about adding a `configurators` option to the service definition to set multiple configurators from the config file, so it's still defined at the same location as the service. 

IMO there is a legitimate use case : if you have multiple configurators and you want to apply different apply different "compositions" depending on the service (exemple: c1,c2 on service1, c1,c2,c3 on service2, ...) you would have to create a boilerplate class for every different composition of configurators. With this new option you would just arrange your configurators in the service definition.

What do you think ?

nicolas-grekas:
> if you have multiple configurators and you want to apply different apply different "compositions" depending on the service (exemple: c1,c2 on service1, c1,c2,c3 on service2, ...) you would have to create a boilerplate class for every different composition of configurators

I think this is the way to go. Configurators are not common enough to warrant adding complexity (priorities are complex for end users).

nicolas-grekas:
We also don't need `#[WithConfigurator]`:  ` #[Autoconfigure(configurator: foo]`  does the job to me.

stof:
I also think that multiple configurators (which require managing priorities for the declarative way) are overkill. Having a single PHP callable responsible for altering the configuration is much simpler to me (especially given that most services don't need configurators). And if you have very complex needs, you could still create a configurator service in which you inject other services that get used to delegate the configuration...

nicolas-grekas:
Let me close then.  Thanks for proposing!

PR #52619, opened 2023-11-16
[Notifier][PropertyInfo][VarDumper] Remove unused code

alamirault:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Issues        | Fix #... 
| License       | MIT

Remove unused code

### DIFF
--- a/src/Symfony/Component/PropertyInfo/Extractor/ReflectionExtractor.php
+++ b/src/Symfony/Component/PropertyInfo/Extractor/ReflectionExtractor.php
@@ -688,7 +688,7 @@ class ReflectionExtractor implements PropertyListExtractorInterface, PropertyTyp
      */
     private function findAdderAndRemover(\ReflectionClass $reflClass, array $singulars): array
     {
-        if (!\is_array($this->arrayMutatorPrefixes) && 2 !== \count($this->arrayMutatorPrefixes)) {
+        if (2 !== \count($this->arrayMutatorPrefixes)) {
             return [null, null, []];
         }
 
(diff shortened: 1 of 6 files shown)


src/Symfony/Component/PropertyInfo/Extractor/ReflectionExtractor.php
```diff
@@ -688,10 +688,6 @@ private function getPropertyName(string $methodName, array $reflectionProperties
      */
     private function findAdderAndRemover(\ReflectionClass $reflClass, array $singulars): array
     {
-        if (!\is_array($this->arrayMutatorPrefixes) && 2 !== \count($this->arrayMutatorPrefixes)) {
```
nicolas-grekas, line 691: why don't we need the `2 !== \count($this->arrayMutatorPrefixes)` check anymore?

stof: being a public static method, I would keep the safeguard here as there is nothing avoiding mutating it, even though it is internal.

OskarStark (approved):
Please check fabbot, thanks

LGTM for the Notifier part

src/Symfony/Component/PropertyInfo/Extractor/ReflectionExtractor.php
nicolas-grekas: IIUC, this means this change should be reverted, isn't it @alamirault?

alamirault: I reverted `2 !== \count($this->arrayMutatorPrefixes)` part

fabpot (approved):

fabpot:
Thank you @alamirault.

PR #52626, opened 2023-11-17
[Serializer] Fix denormalizing date intervals having both weeks and days

### LINKED ISSUE #47629: DateIntervalNormalizer does not support weeks and days together

5.4.12,6.1.4

### Description

Since PHP 8.0 [DateInterval::construct](https://www.php.net/manual/en/dateinterval.construct.php) supports periods containing both D and W period designators

```php
echo (new \DateInterval('P3W2D'))->format('%d');
// outputs '2' prior PHP 8.0
// outputs '23' in PHP >= 8.0
```

The problem is the built-in [DateIntervalNormalizer](https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/Serializer/Normalizer/DateIntervalNormalizer.php) does not support it, so it throws exception if both weeks and date exist in source period string as current regular expression in method isISO8601 does not allow it 

```php
        if (!$this->isISO8601($data)) {
            throw new UnexpectedValueException('Expected a valid ISO 8601 interval string.');
        }
``` 

### How to reproduce

```php
// PHP >= 8.0
$normalizer = new \Symfony\Component\Serializer\Normalizer\DateIntervalNormalizer();
$object = $normalizer->denormalize('P2W6D', \DateInterval::class, null, [
    'dateinterval_format' => '%rP%yY%mM%wW%dDT%hH%iM%sS',
]);
// expected: $object instanceof \DateInterval
// actual: \Symfony\Component\Serializer\Exception\UnexpectedValueException thrown
```

### Possible Solution

I believe a possible solution would be to update regular expression used in isISO8601 method for PHP versions >= 8.0, like this:

```php
    private function isISO8601(string $string): bool
    {
        if (PHP_VERSION_ID >= 80000) {
            return preg_match('/^[\-+]?P(?=\w*(?:\d|%\w))(?:\d+Y|%[yY]Y)?(?:\d+M|%[mM]M)?(?:\d+W|%[wW]W)?(?:\d+D|%[dD]D)?(?:T(?:\d+H|[hH]H)?(?:\d+M|[iI]M)?(?:\d+S|[sS]S)?)?$/', $string);
        }

        return preg_match('/^[\-+]?P(?=\w*(?:\d|%\w))(?:\d+Y|%[yY]Y)?(?:\d+M|%[mM]M)?(?:(?:\d+D|%[dD]D)|(?:\d+W|%[wW]W))?(?:T(?:\d+H|[hH]H)?(?:\d+M|[iI]M)?(?:\d+S|[sS]S)?)?$/', $string);
    }
```

### Additional Context

_No response_

oneNevan:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #47629
| License       | MIT

Added support for date periods having both W and D period designators combined together (which is valid duration for ISO 8601-2). Starting from PHP 8 this is supported by \DateTimeInterval, see https://bugs.php.net/bug.php?id=61366 for reference

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/DateIntervalNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/DateIntervalNormalizer.php
@@ -128,6 +128,10 @@ class DateIntervalNormalizer implements NormalizerInterface, DenormalizerInterfa
 
     private function isISO8601(string $string): bool
     {
+        if (\PHP_VERSION_ID >= 80000) {
+            return preg_match('/^[\-+]?P(?=\w*(?:\d|%\w))(?:\d+Y|%[yY]Y)?(?:\d+M|%[mM]M)?(?:\d+W|%[wW]W)?(?:\d+D|%[dD]D)?(?:T(?:\d+H|[hH]H)?(?:\d+M|[iI]M)?(?:\d+S|[sS]S)?)?$/', $string);
+        }
+
         return preg_match('/^[\-+]?P(?=\w*(?:\d|%\w))(?:\d+Y|%[yY]Y)?(?:\d+M|%[mM]M)?(?:(?:\d+D|%[dD]D)|(?:\d+W|%[wW]W))?(?:T(?:\d+H|[hH]H)?(?:\d+M|[iI]M)?(?:\d+S|[sS]S)?)?$/', $string);
     }
 }
--- a/src/Symfony/Component/Serializer/Tests/Normalizer/DateIntervalNormalizerTest.php
+++ b/src/Symfony/Component/Serializer/Tests/Normalizer/DateIntervalNormalizerTest.php
@@ -119,6 +119,21 @@ class DateIntervalNormalizerTest extends TestCase
         $this->assertDateIntervalEquals($this->getInterval('P0Y0M0DT12H34M0S'), $normalizer->denormalize('PT12H34M', \DateInterval::class));
     }
 
+    /**
+     * Since PHP 8.0 DateInterval::construct supports periods containing both D and W period designators.
+     *
+     * @requires PHP 8
+     */
+    public function testDenormalizeIntervalWithBothWeeksAndDays()
+    {
+        $input = 'P1W1D';
+        $interval = $this->normalizer->denormalize($input, \DateInterval::class, null, [
+            DateIntervalNormalizer::FORMAT_KEY => '%rP%yY%mM%wW%dDT%hH%iM%sS',
+        ]);
+        $this->assertDateIntervalEquals($this->getInterval($input), $interval);
+        $this->assertSame(8, $interval->d);
+    }
+
     public function testDenormalizeExpectsString()
     {
         $this->expectException(NotNormalizableValueException::class);


src/Symfony/Component/Serializer/Normalizer/DateIntervalNormalizer.php
```diff
@@ -34,8 +34,6 @@ public function __construct(array $defaultContext = [])
     }
 
     /**
-     * {@inheritdoc}
-     *
```
mtarld: These comment should not be removed in that PR

src/Symfony/Component/Serializer/Tests/Normalizer/DateIntervalNormalizerTest.php
```diff
@@ -119,6 +119,22 @@ public function testDenormalizeIntervalsWithOmittedPartsBeingZero()
         $this->assertDateIntervalEquals($this->getInterval('P0Y0M0DT12H34M0S'), $normalizer->denormalize('PT12H34M', \DateInterval::class));
     }
 
+    /**
+     * Since PHP 8.0 DateInterval::construct supports periods containing both D and W period designators.
+     *
+     * @requires PHP 8
+     *
+     * @testWith ["%rP%yY%mM%wW%dDT%hH%iM%sS","P3W2D", 23]
```
mtarld: As you have only one dataset, I think `@testWith` is not needed

mtarld (approved):
So, to make it work, we need to update the format.

Maybe it could be great to create as well a PR targeting 7.1 to update the format const value, WDYT?

src/Symfony/Component/Serializer/Normalizer/DateIntervalNormalizer.php
```diff
@@ -118,16 +108,17 @@ public function denormalize($data, string $type, string $format = null, array $c
         }
     }
 
-    /**
-     * {@inheritdoc}
```
nicolas-grekas: let's revert all those changes, they're false positive from the cs fixer on branch 5.4

nicolas-grekas (approved):
LGTM once comments are fixed.

src/Symfony/Component/Serializer/Normalizer/DateIntervalNormalizer.php
oneNevan: fixed

src/Symfony/Component/Serializer/Tests/Normalizer/DateIntervalNormalizerTest.php
oneNevan: fixed

src/Symfony/Component/Serializer/Normalizer/DateIntervalNormalizer.php
oneNevan: fixed

nicolas-grekas:
Thank you @oneNevan.

oneNevan:
> So, to make it work, we need to update the format.
> 
> Maybe it could be great to create as well a PR targeting 7.1 to update the format const value, WDYT?

@mtarld Regarding this, I agree it does make sense, but there is one thing to consider - adding `%wW` to FORMAT_KEY constant by default would break normalization as [DateInterval::format does not support weeks](https://3v4l.org/AiKWU), php docs does not provide it either https://www.php.net/manual/en/dateinterval.format.php - no weeks available :)
Which means we can only use W when denormalizing values from period strings into DateInterval instance.

So, it looks like if we want to support weeks out of the box, we would have to split FORMAT_KEY constant, into two constants: one for normalizer (without `%wW`), and another one for denormalizer (with `%wW`).

At this point I wanted to fix the regular expression at least, so that devs could use custom format if they deal with periods having weeks (as I do), but I'm not sure if it worth changing the current default format to support it out of the box, do you think it makes sense?

mtarld:
Indeed, it sounds like an edge case finally. I think that we might keep it like that, thank you for that feedback :slightly_smiling_face:

PR #52632, opened 2023-11-17
[PropertyInfo] Make `PhpDocExtractor::getDocBlock()` public

Nyholm:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | N/a
| License       | MIT

Im parsing properties and I would love to find out what is in the doc block for a property. 
Help me figure out why it would be a bad idea to make this method public.


stof:
As this method does returns a shape with 3 info and not just the docblock, I would not make it public. I don't think we want to provide BC for the current API.

Nyholm:
Good point. 
I suggest to create a new public method then with a simpler API. What do you think about that?

ro0NL:
> Help me figure out why it would be a bad idea to make this method public.

I think because it's not a component responsibility, as per current contracts.

Nyholm:
Hm.. Im not sure I agree with that. You could expect the "property info" component to be able to extract the Doc Block of a property. 

Or maybe I misunderstand you. Are you suggesting I should introduce a proper interface for this method?

ro0NL:
Shouldn't you be using https://github.com/phpDocumentor/ReflectionDocBlock instead?

ro0NL:
> You could expect the "property info" component to be able to extract the Doc Block of a property

Yes, but there's no real contract for it currently

Nyholm:
> Shouldn't you be using https://github.com/phpDocumentor/ReflectionDocBlock instead?

I looked into that. But I do like the feature from PropertyInfo component that would find where the docblock is. Ie, if the property is not documented it goes to the mutator. 

I'll make sure to prepare a proper contract and prepare this PR.

Nyholm:
I've updated the PR with a proper interface and a good return value.

src/Symfony/Component/PropertyInfo/PropertyDocBlockExtractorInterface.php
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Extract a property's doc block.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+interface PropertyDocBlockExtractorInterface
+{
+    /**
+     * Get doc block for a property.
```
dunglas: I would detail the fallback logic.

Nyholm: Thank you. 

Sure let's put that in the contract too.

dunglas (approved):

mtarld (approved):

src/Symfony/Component/PropertyInfo/Extractor/PhpDocExtractor.php
```diff
@@ -217,9 +225,11 @@ private function filterDocBlockParams(DocBlock $docBlock, string $allowedParam):
     }
 
     /**
+     * Get DocBlock for a property.
```
fabpot:
```suggestion
     * Gets a property DocBlock.
```

src/Symfony/Component/PropertyInfo/PropertyDocBlockExtractorInterface.php
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * A property's doc block may be located on a constructor promoted argument, on
+ * the property or on a mutator for that property.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+interface PropertyDocBlockExtractorInterface
+{
+    /**
+     * Get the first available doc block for a property. It finds the doc block
```
fabpot:
```suggestion
     * Gets the first available doc block for a property. It finds the doc block
```

fabpot (approved):

Nyholm:
Thank you for the reviews. I've updated according to the suggestions.

src/Symfony/Component/PropertyInfo/Extractor/PhpDocExtractor.php
```diff
@@ -217,9 +225,11 @@ private function filterDocBlockParams(DocBlock $docBlock, string $allowedParam):
     }
 
     /**
+     * Gets a property DocBlock.
```
OskarStark: It's a little bit confusing, that this and not the `get*` method _gets_ the docblock

You know what I mean? Maybe remove the description entirely?

src/Symfony/Component/PropertyInfo/PropertyDocBlockExtractorInterface.php
```diff
@@ -0,0 +1,36 @@
...
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+interface PropertyDocBlockExtractorInterface
+{
+    /**
+     * Gets the first available doc block for a property. It finds the doc block
+     * by the following priority:
+     * - constructor promoted argument
+     * - the class property
+     * - a mutator method for that property
```
OskarStark:
```suggestion
     *  - constructor promoted argument
     *  - the class property
     *  - a mutator method for that property
```

src/Symfony/Component/PropertyInfo/Tests/Extractor/PhpDocExtractorTest.php
```diff
@@ -43,6 +44,19 @@ public function testExtract($property, ?array $type, $shortDescription, $longDes
         $this->assertSame($longDescription, $this->extractor->getLongDescription('Symfony\Component\PropertyInfo\Tests\Fixtures\Dummy', $property));
     }
 
+    public function testGetDocBlock()
+    {
+        $docBlock = $this->extractor->getDocBlock('Symfony\Component\PropertyInfo\Tests\Fixtures\Dummy', 'g');
```
OskarStark:
```suggestion
        $docBlock = $this->extractor->getDocBlock(Dummy::class, 'g');
```
Same for the other ones in this file

OskarStark (approved):
Some comments

src/Symfony/Component/PropertyInfo/Tests/Extractor/PhpDocExtractorTest.php
Nyholm: Thank you!

src/Symfony/Component/PropertyInfo/Extractor/PhpDocExtractor.php
Nyholm: Removed

fabpot (approved):

fabpot:
Thank you @Nyholm.

PR #52636, opened 2023-11-17
[DependencyInjection] Prepend extension config with `ContainerConfigurator`

yceruto:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

I found this by fixing an issue in a bundle that was trying to prepend extension configs using `$container->extension('namespace', [...])` in `AbstractBundle::prependExtension()`, which indeed was appending that config instead of prepending it. Most importantly, the append strategy requires the extension namespace to be loaded beforehand, which is not required when prepend is used.

This title DX improvement helps to avoid the confusion between `ContainerConfigurator $container` and `ContainerBuilder $builder` to prepend extension config by allowing `ContainerConfigurator` to do the same now.

Example:
```php
class AcmeFooBundle extends AbstractBundle
{
    {
        // Before (only way)
        $builder->prependExtensionConfig('namespace', ['foo' => 'bar']);

        // After (also this way) passing "true" as the 3rd parameter
        $container->extension('namespace', ['foo' => 'bar'], true);
    }
}
```
Instead of adding a new `$prepend` argument to the existing method, I could create a new method, e.g., `ContainerConfigurator::prependExtension()`. What do you prefer?

This also helps when you want to prepend several or large configs in your bundle or extension class. Actually, using just `$container->import('...')` doesn't work because internally it will always append the configs, unless you do the following hidden trick below.

```php
// acme-bundle/config/packages/configs.php

return static function (ContainerBuilder $container) {
    $container->prependExtensionConfig('namespace', ['large' => 'config', ...]);
};
```
If you type `ContainerBuilder` instead of `ContainerConfigurator` in the external PHP config file, the builder instance will be passed instead, allowing you to use the `prependExtensionConfig()` method.

But with this proposal, it's simpler as you can keep using `ContainerConfigurator` to prepend extension configs without doing any tactic.

### DIFF
--- a/src/Symfony/Component/DependencyInjection/CHANGELOG.md
+++ b/src/Symfony/Component/DependencyInjection/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Add argument `$prepend` to `ContainerConfigurator::extension()` to prepend the configuration instead of appending it
+
 7.0
 ---
 
--- a/src/Symfony/Component/DependencyInjection/Loader/Configurator/ContainerConfigurator.php
+++ b/src/Symfony/Component/DependencyInjection/Loader/Configurator/ContainerConfigurator.php
@@ -48,8 +48,14 @@ class ContainerConfigurator extends AbstractConfigurator
         $this->env = $env;
     }
 
-    final public function extension(string $namespace, array $config): void
+    final public function extension(string $namespace, array $config, bool $prepend = false): void
     {
+        if ($prepend) {
+            $this->container->prependExtensionConfig($namespace, static::processValue($config));
+
+            return;
+        }
+
         if (!$this->container->hasExtension($namespace)) {
             $extensions = array_filter(array_map(fn (ExtensionInterface $ext) => $ext->getAlias(), $this->container->getExtensions()));
             throw new InvalidArgumentException(sprintf('There is no extension able to load the configuration for "%s" (in "%s"). Looked for namespace "%s", found "%s".', $namespace, $this->file, $namespace, $extensions ? implode('", "', $extensions) : 'none'));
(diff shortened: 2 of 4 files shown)


yceruto:
Ideally, I would like to import configs from the `AbstractBundle::prependExtension()` method this way:
```php
$container->import('../config/packages/framework.yaml', prepend: true); // or using a new method
```
But the impact is currently too high, as this prepend value should pass across multiple interfaces and processes until reaching the final `ContainerBuilder::loadFromExtension()` invocation.

yceruto:
I was also thinking about a new marker in the config file, similar to `when@env`, but using it for the importing config strategy, e.g.:
```yaml
import@prepend:
    framework:
        messenger: ...
```
And then prepend that config instead of appending. Even with less impact on the code, I don't like it too much as the config itself shouldn't be aware of the importing strategy.

Similarly, in the `imports` section:
```yaml
imports:
    - { resource: 'config.yaml', prepend: true }
```
But still, I'm not sure as you will need to create two config files to achieve that and again the prepend value needs to by pass some loading interfaces where this prepend value doesn't mean anything.

nicolas-grekas:
> the append strategy requires the extension namespace to be loaded beforehand, which is not required when prepend is used.

Can you please refresh my mind about this: why don't we check the namespace when prepending?

yceruto:
> Can you please refresh my mind about this: why don't we check the namespace when prepending?

It's on purpose because the extension might not even exist. This feature was initially created to provide preset settings in bundles, so if the extension doesn't exist in your app, that prepended config is ignored.

yceruto:
Also because the extension order according to the `bundles.php`.

nicolas-grekas (approved):
Got it, makes sense to me also.

nicolas-grekas:
Thank you @yceruto.

PR #52639, opened 2023-11-17
[Messenger] Add `SKIP LOCKED` to the query that retrieves messages

hgraca:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Issues        |  -
| License       | MIT

The current `SELECT ... FOR UPDATE` retrieves rows, using the index to find and lock the retrieved rows.
This means that when we have more than one consumer, while one consumer is locking those rows, the whole index is locked thus other queries (consumers) will be put on hold.
While with a small table this might not be noticeable, as the table grows the meddling with the index becomes slower and the other consumers have to wait more time, eventually making the MQ inoperable.

The `SKIP LOCKED` addition will allow other consumers to query the table and get messages immediately, ignoring the rows that other consumers are locking.

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Doctrine/CHANGELOG.md
+++ b/src/Symfony/Component/Messenger/Bridge/Doctrine/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Use `SKIP LOCKED` in the doctrine transport for MySQL, PostgreSQL and MSSQL
+
 5.1.0
 -----
 
--- a/src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/ConnectionTest.php
+++ b/src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/ConnectionTest.php
@@ -14,10 +14,15 @@ namespace Symfony\Component\Messenger\Bridge\Doctrine\Tests\Transport;
 use Doctrine\DBAL\Connection as DBALConnection;
 use Doctrine\DBAL\Exception as DBALException;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Platforms\MariaDb1060Platform;
 use Doctrine\DBAL\Platforms\MariaDBPlatform;
 use Doctrine\DBAL\Platforms\MySQL57Platform;
+use Doctrine\DBAL\Platforms\MySQL80Platform;
 use Doctrine\DBAL\Platforms\MySQLPlatform;
 use Doctrine\DBAL\Platforms\OraclePlatform;
+use Doctrine\DBAL\Platforms\PostgreSQL100Platform;
+use Doctrine\DBAL\Platforms\PostgreSQL94Platform;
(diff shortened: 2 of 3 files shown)


src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/ConnectionTest.php
```diff
@@ -391,28 +396,94 @@ class_exists(MySQLPlatform::class) ? new MySQLPlatform() : new MySQL57Platform()
...
+                'SELECT m.* FROM messenger_messages m WITH (UPDLOCK, ROWLOCK, READPAST) WHERE (m.queue_name = ?) AND (m.delivered_at is null OR m.delivered_at < ?) AND (m.available_at <= ?) ORDER BY available_at ASC OFFSET 0 ROWS FETCH NEXT 1 ROWS ONLY  ',
+            ];
+        }
+
+        if (!method_exists(QueryBuilder::class, 'forUpdate')) {
+            yield 'Oracle & DBAL<3.8' => [
                 new OraclePlatform(),
                 'SELECT w.id AS "id", w.body AS "body", w.headers AS "headers", w.queue_name AS "queue_name", w.created_at AS "created_at", w.available_at AS "available_at", w.delivered_at AS "delivered_at" FROM messenger_messages w WHERE w.id IN (SELECT a.id FROM (SELECT m.id FROM messenger_messages m WHERE (m.queue_name = ?) AND (m.delivered_at is null OR m.delivered_at < ?) AND (m.available_at <= ?) ORDER BY available_at ASC) a WHERE ROWNUM <= 1) FOR UPDATE',
             ];
+        } elseif (class_exists(MySQL57Platform::class)) {
```
OskarStark, line 477: Oracle?

hgraca: No, this was already there, it's a way to know if we are using DBAL 4 or not.
Since MySQL57Platform is deprecated, it will not be included in DBAL4 anymore.
I guess we could/should put this in a method with a meaningful name...
I can do that if/when I have time, otherwise, I think it's fine to leave it as is since it was already like this.

src/Symfony/Component/Messenger/Bridge/Doctrine/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1.0
+-----
+
+* Use `SKIP LOCKED` in the doctrine transport for MySQL, PostgreSQL and MSSQL
```
nicolas-grekas:
```suggestion
7.1
---

 * Use `SKIP LOCKED` in the doctrine transport for MySQL, PostgreSQL and MSSQL
```

nicolas-grekas (approved):

nicolas-grekas:
Thank you @hgraca.

hgraca:
You are most welcome @nicolas-grekas! Thank you for Symfony! :)

Btw, do you think you can help me get [this other PR](https://github.com/symfony/symfony/pull/51034) merged as well?

PR #52654, opened 2023-11-20
Rate Limiter requires Symfony Lock

DavidGarciaCat:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | N/A
| License       | MIT

When we install and configure Symfony's Rate Limiter component, we get an error reporting this component depends on Symfony's Lock component (understandable to prevent further access/attempts). Hence, it should be included as one of the required dependencies.

I assume previous versions of this component are also suffering from the same problem, but I just found it by testing the RC1 for the Symfony 7 branch

I understand this can be flagged as a bug because we should not expect to get an error after installing it

### DIFF
--- a/src/Symfony/Component/RateLimiter/composer.json
+++ b/src/Symfony/Component/RateLimiter/composer.json
@@ -17,6 +17,7 @@
     ],
     "require": {
         "php": ">=8.2",
+        "symfony/lock": "^6.4|^7.0",
         "symfony/options-resolver": "^6.4|^7.0"
     },
     "require-dev": {


DavidGarciaCat:
It seems the GitHub actions are failing due to an error with the serializer component, but this PR doesn't do anything with it 🤔 

```
Error: KO src/Symfony/Component/Serializer
```

Any suggestion?

wouterj:
Hi!
This dependency is optional. You can use the RateLimiter component without the Lock component (see e.g. the [minimal example in the README](https://github.com/symfony/symfony/tree/7.1/src/Symfony/Component/RateLimiter#getting-started)).

DavidGarciaCat:
Based on the [website's documentation](https://symfony.com/doc/current/rate_limiter.html), we can use the dependency injection to pass the factory and inject it automatically (based on the camelCase notation).

I assume removing the lock component is possible using the ```lock_factory: null``` setting under the [Using Locks to Prevent Race Conditions](https://symfony.com/doc/current/rate_limiter.html#using-locks-to-prevent-race-conditions) section? If so, the other question is how we can pass an in-memory (or another) storage in a case like this (especially useful when working behind a load balancer).

PR #52676, opened 2023-11-21
[Serializer] add normalizer for non-backed enumerations

klkvsk:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

This PR adds a de/normalizer for enum values, that do not have a backing type.

Before, a default object normalization approach is used on those values, resulting in a `{name: "CASE_NAME"}` serialized value, that can not be denormalized.

With this PR, `MyEnum::CASE_NAME` is normalized as `"CASE_NAME"`. For denormalization, `defined(...)` and `constant(...)` are used to check and instantiate the enum value.

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/NonBackedEnumNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/NonBackedEnumNormalizer.php
@@ -54,18 +54,18 @@ final class NonBackedEnumNormalizer implements NormalizerInterface, Denormalizer
         }
 
         if (!\is_string($data) && !($context[BackedEnumNormalizer::ALLOW_INVALID_VALUES] ?? false)) {
-            throw NotNormalizableValueException::createForUnexpectedDataType("The data is not a string, you should pass a string that can be parsed as an enumeration case of type $type.", $data, [Type::BUILTIN_TYPE_STRING], $context['deserialization_path'] ?? null, true);
+            throw NotNormalizableValueException::createForUnexpectedDataType(sprintf("The data is not a string, you should pass a string that can be parsed as an enumeration case of type %s.", $type), $data, [Type::BUILTIN_TYPE_STRING], $context['deserialization_path'] ?? null, true);
         }
 
         try {
             $constantName = "$type::$data";
             if (!\defined($constantName)) {
-                throw new \ValueError("$data is not a valid case for enum $type.");
+                throw new \ValueError(sprintf("%s is not a valid case for enum %s.", $data, $type));
             }
 
             $value = \constant($constantName);
             if (!\is_object($value) || $value::class !== $type) {
-                throw new \ValueError("$constantName is not a valid enum case.");
+                throw new \ValueError(sprintf("%s is not a valid enum case.", $constantName));
             }
 
             return $value;
@@ -74,9 +74,9 @@ final class NonBackedEnumNormalizer implements NormalizerInterface, Denormalizer
                 return null;
             }
             if (isset($context['has_constructor'])) {
-                throw new InvalidArgumentException("The data must belong to a non-backed enumeration of type $type.");
+                throw new InvalidArgumentException(sprintf("The data must belong to a non-backed enumeration of type %s.", $type));
             }
-            throw NotNormalizableValueException::createForUnexpectedDataType("The data must belong to a non-backed enumeration of type $type.", $data, [$type], $context['deserialization_path'] ?? null, true, 0, $e);
+            throw NotNormalizableValueException::createForUnexpectedDataType(sprintf("The data must belong to a non-backed enumeration of type %s.", $type), $data, [$type], $context['deserialization_path'] ?? null, true, 0, $e);
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/Serializer/Normalizer/NonBackedEnumNormalizer.php
```diff
@@ -0,0 +1,88 @@
...
+     * @throws NotNormalizableValueException
+     */
+    public function denormalize(mixed $data, string $type, string $format = null, array $context = []): mixed
+    {
+        if (!is_subclass_of($type, \UnitEnum::class) || is_subclass_of($type, \BackedEnum::class)) {
+            throw new InvalidArgumentException('The data must belong to a non-backed enumeration.');
+        }
+
+        if (!\is_string($data) && !($context[BackedEnumNormalizer::ALLOW_INVALID_VALUES] ?? false)) {
+            throw NotNormalizableValueException::createForUnexpectedDataType("The data is not a string, you should pass a string that can be parsed as an enumeration case of type $type.", $data, [Type::BUILTIN_TYPE_STRING], $context['deserialization_path'] ?? null, true);
```
yceruto: According to our Coding Standard: Exception message strings must be concatenated using `sprintf()` function. Same for the other cases above.

src/Symfony/Component/Serializer/Normalizer/NonBackedEnumNormalizer.php
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        try {
+            $constantName = "$type::$data";
+            if (!\defined($constantName)) {
+                throw new \ValueError("$data is not a valid case for enum $type.");
+            }
+
+            $value = \constant($constantName);
+            if (!\is_object($value) || $value::class !== $type) {
```
yceruto, line 67: Could you elaborate a bit more on under what scenario this condition can reach `true` at this point?

yceruto (changes requested):
I've some doubts about whether it's better to create just one Normalizer that supports Enums (any kind) with these two strategies for `UnitEnum` and `BackedEnum`. It seems mergeable to me.

klkvsk:
>just one Normalizer that supports Enums (any kind) with these two strategies for `UnitEnum` and `BackedEnum`

I agree. I did not add new logic to BackedEnumNormalizer since it does not feel clean to me to still have it named "Backed". And decided not to rename it too, so it wouldn't create a BC break.

Maybe should I create EnumNormalizer, and make BackedEnumNormalizer extend it with logging deprecations?
What is your suggested way of handling such merge correctly?

src/Symfony/Component/Serializer/Normalizer/NonBackedEnumNormalizer.php
klkvsk: Using `constant(..)` is a way to address enum cases by their fqcn+name, but enums can have real constants too:

```php
enum MyEnum: string {
    const A = "a";
    case B = "b";
}

var_dump(constant('MyEnum::A')); // string(1) "a"
var_dump(constant('MyEnum::B')); // enum(MyEnum::B)
```

This is why we should check again that data was an enum reference.

yceruto:
>Maybe should I create EnumNormalizer, and make BackedEnumNormalizer extend it with logging deprecations?
What is your suggested way of handling such merge correctly?

Maybe an abstract EnumNormalizer with common code (e.g., `support*` methods and some abstract methods) and then BackedEnumNormalizer and UnitEnumNormalizer as subclasses. It's probable that we won't need to deprecate anything.

stof:
Non-backed enums are precisely meant to not have a backing scalar value. If you need to serialize your enum to a scalar, that's the best reason to use a backed enum.
And if for some reason you cannot change the enum, you can still write a custom normalizer supporting *that* enum in your project with a custom serialization.

I'm -1 for that. Symfony should not have a core normalizer that explicitly goes against the design of PHP enums.

derrabus (changes requested):
This is a proposal that we've gotten in various flavors and so far we've rejected them mostly. If an enum is not backed it means that its serialized representation is undefined. The solution is to back it.

This PR implements a piece of generic guesswork that I believe does not represent good practices.

yceruto:
>Non-backed enums are precisely meant to not have a backing scalar value. If you need to serialize your enum to a scalar, that's the best reason to use a backed enum.

That makes sense, so I'm closing this one as it won't be implemented. Anyway thanks @klkvsk, for your first contribution, and sorry that it won't be accepted. Best of luck with the next one 😅.

klkvsk:
Well, okay... 
Doesn't it make sense then to throw an exception when a non-backed enum is going through serializer?
My point of making PR was to address the ambiguity of getting an array of `{name: "case"}` for non-backed and a scalar `"case"` for backed as a result of normalization of almost uniform objects.

stof:
We should probably make ObjectNormalizer reject `UnitEnum` indeed, to have a better DX.

PR #52677, opened 2023-11-21
[Translation] [Lokalise] Fix language format on Lokalise Provider

welcoMattic:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Issues        | Fix #... 
| License       | MIT

Lokalise has changed their API a little bit, and ISO languages are formatted with a `-`, which is not wrong. But we use it as locale to load the translation with XliffLoader (and other loaders), and it leads to wrong filename like `messages+intl-icu.en-US.xlf`.

This PR fixes the issue.

### DIFF
--- a/src/Symfony/Component/Translation/Bridge/Lokalise/LokaliseProvider.php
+++ b/src/Symfony/Component/Translation/Bridge/Lokalise/LokaliseProvider.php
@@ -147,7 +147,6 @@ final class LokaliseProvider implements ProviderInterface
             'json' => [
                 'format' => 'symfony_xliff',
                 'original_filenames' => true,
-                'directory_prefix' => '%LANG_ISO%',
                 'filter_langs' => array_values($locales),
                 'filter_filenames' => array_map([$this, 'getLokaliseFilenameFromDomain'], $domains),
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/Translation/Bridge/Lokalise/LokaliseProvider.php
```diff
@@ -167,7 +166,10 @@ private function exportFiles(array $locales, array $domains): array
             throw new ProviderException(sprintf('Unable to export translations from Lokalise: "%s".', $response->getContent(false)), $response);
         }
 
-        return $responseContent['files'];
+        // Lokalise returns languages with "-" separator, we need to reformat them to "_" separator.
+        $reformattedLanguages = array_map(static fn ($language) => strtr($language, '-', '_'), array_keys($responseContent['files']));
```
nicolas-grekas: php7.2 for 5.4 ;)

nicolas-grekas:
Thank you @welcoMattic.

PR #52679, opened 2023-11-22
[Process] `ExecutableFinder::addSuffix()` has no effect

TravisCarden:
...except (probably) on Windows.

| Q             | A
| ------------- | ---
| Branch?       | 7.2
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

`ExecutableFinder::addSuffix()` currently has no effect on non-Windows systems because `ExecutableFinder::find()` altogether ignores added suffixes on them. This PR adds tests that prove that it's broken and then fixes it. It clarifies a related docblock along the way.

**Note**: I tried to follow the pattern in similar tests of creating fixture files in the temp directory, but `ExecutableFinder::find()` returned a different temp path than `sys_get_temp_dir()`, rendering path comparison impossible. So I added a few fixture files directly to the repo.

### DIFF
--- a/src/Symfony/Component/Process/ExecutableFinder.php
+++ b/src/Symfony/Component/Process/ExecutableFinder.php
@@ -19,7 +19,15 @@ namespace Symfony\Component\Process;
  */
 class ExecutableFinder
 {
-    private array $suffixes = ['.exe', '.bat', '.cmd', '.com'];
+    private array $suffixes = [];
+
+    public function __construct()
+    {
+        // Set common extensions on Windows.
+        if ('\\' === \DIRECTORY_SEPARATOR) {
+            $this->suffixes = ['.exe', '.bat', '.cmd', '.com'];
+        }
+    }
 
     /**
      * Replaces default suffixes of executable.
@@ -30,7 +38,10 @@ class ExecutableFinder
     }
 
     /**
-     * Adds new possible suffix to check for executable.
+     * Adds new possible suffix to check for executable, including the dot (.).
+     *
+     *     $finder = new ExecutableFinder();
+     *     $finder->addSuffix('.foo');
      */
     public function addSuffix(string $suffix): void
     {
@@ -52,10 +63,10 @@ class ExecutableFinder
         );
 
         $suffixes = [''];
-        if ('\\' === \DIRECTORY_SEPARATOR) {
-            $pathExt = getenv('PATHEXT');
-            $suffixes = array_merge($pathExt ? explode(\PATH_SEPARATOR, $pathExt) : $this->suffixes, $suffixes);
+        if ('\\' === \DIRECTORY_SEPARATOR && $pathExt = getenv('PATHEXT')) {
+            $suffixes = array_merge(explode(\PATH_SEPARATOR, $pathExt), $suffixes);
         }
+        $suffixes = array_merge($suffixes, $this->suffixes);
         foreach ($suffixes as $suffix) {
             foreach ($dirs as $dir) {
                 if (@is_file($file = $dir.\DIRECTORY_SEPARATOR.$name.$suffix) && ('\\' === \DIRECTORY_SEPARATOR || @is_executable($file))) {
--- a/src/Symfony/Component/Process/Tests/ExecutableFinderTest.php
+++ b/src/Symfony/Component/Process/Tests/ExecutableFinderTest.php
@@ -85,6 +85,31 @@ class ExecutableFinderTest extends TestCase
         $this->assertSamePath(\PHP_BINARY, $result);
     }
 
+    public function testFindWithoutSuffix()
+    {
+        $fixturesDir = __DIR__.\DIRECTORY_SEPARATOR.'Fixtures';
+        $name = 'executable_without_suffix';
+
+        $finder = new ExecutableFinder();
+        $result = $finder->find($name, null, [$fixturesDir]);
+
+        $this->assertSamePath($fixturesDir.\DIRECTORY_SEPARATOR.$name, $result);
+    }
+
+    public function testFindWithAddedSuffixes()
+    {
+        $fixturesDir = __DIR__.\DIRECTORY_SEPARATOR.'Fixtures';
+        $name = 'executable_with_added_suffix';
+        $suffix = '.foo';
+
+        $finder = new ExecutableFinder();
+        $finder->addSuffix($suffix);
+
+        $result = $finder->find($name, null, [$fixturesDir]);
+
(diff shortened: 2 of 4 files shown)


nicolas-grekas (approved):

nicolas-grekas:
Can you please check if this bugfix shouldn't be applied to either 5.4 or 6.3 and rebase/retarget if yes?

xabbuh (approved):

TravisCarden:
@nicolas-grekas Do you mean you want me to see if the bug already exists in 5.4 or 6.3 and rebase if so? It _does_ already exist in both. I've rebased onto 6.3. Is that what you want? Or do you want me to go _all the way_ back to 5.4? (I thought I understood the bugfix policy on this, but maybe I don't.)

stof:
This option also has no effect on Windows when the `PATH_EXT` environment variable is provided.

As this option has been a no-op for other OS since the creation of the process component, I don't think we should change that as a bugfix.

stof:
Btw, what is your use case for using this on non-Windows systems ? AFAICT, only Windows automatically adds an implicit extension to the executable name when running it.

TravisCarden:
> what is your use case for using this on non-Windows systems ?

I need to look for executables that may end in `.phar`--Composer, in this case.

> AFAICT, only Windows automatically adds an implicit extension to the executable name when running it.

That is, in fact, the current behavior--and that's the problem. The in-code documentation gives no indication that the behavior is conditioned on OS, leaving no reason for a user to suspect otherwise:

https://github.com/symfony/symfony/blob/d625fa0c5cf344198c05ada94a52db62c6a866bd/src/Symfony/Component/Process/ExecutableFinder.php#L34-L39

In other words, unless users are expected to search the code before using _any and every_ function in the codebase to see whether it works to their OS, this doesn't behave as advertised. That seems like a bug to me.

stof:
@TravisCarden if you need to find an executable named `composer.phar`, look for that. Because running `composer` in a shell won't automatically find `composer.phar` either anyway.

stof:
Given that suffixes are totally ignored on non-Windows systems and are ignored on Windows systems that have the `PATH_EXT` environment variable (which is taken as the source of truth instead), I would rather deprecate those methods in PHP 7.1 than suddenly turning them into doing something.
the current behavior has been implemented in April 2011.

TravisCarden:
> if you need to find an executable named `composer.phar`, look for that.

I can't, because I'm writing a Packagist package, so I don't know that detail about my end users. Isn't that the point of suffixes?

> Because running composer in a shell won't automatically find `composer.phar` either anyway.

Actually, it will, if I understand you. After applying this PR, and adding `composer2.phar` to avoid conflict with my global Composer install, it works for me from start to finish:

```php
$finder = new ExecutableFinder();
$finder->addSuffix('.phar');
$path = $finder->find('composer2');
var_dump($path);

// string(28) "/usr/local/bin/composer2.phar"

$process = new Process([$path, '--version']);
$process->run();
var_dump($process->getOutput());

// string(43) "Composer version 2.6.5 2023-10-06 10:11:52"
```

> I would rather deprecate those methods in PHP 7.1 than suddenly turning them into doing something.

Obviously, that's up to you. As I see it, that would make sense if you were talking about adding new behavior that was never claimed or advertised in the first place. But as I argued above, this is how it claimed to work in the first place, so this would just be making good on a promise it has always made but never kept.

nicolas-grekas:
@stof do you think we shouldn't merge this PR? To me it does no harm to have this capability so I'm fine merging it.

nicolas-grekas:
@TravisCarden can you please rebase for 7.1? I think it'd be safer and more consensual to consider this as an improvement.

stof:
This is clearly not a bugfix to me.

And if we introduce such new feature, we should then make it work on all systems. This PR still ignores the custom suffixes on Windows when `PATH_EXT` is defined on the system.

TravisCarden:
@nicolas-grekas I'll gladly make any changes requested, as soon as there's consensus on your end on what they should be. I don't want to spend time on work that has no chance of being committed.

TravisCarden:
Rebased, @nicolas-grekas.

I have my opinion as to whether this is a bug or not, but I don't really care about the semantics if this could be committed to _some version_. If we can't commit it as-is, can someone tell me what it would take?

src/Symfony/Component/Process/ExecutableFinder.php
```diff
@@ -55,6 +58,8 @@ public function find(string $name, ?string $default = null, array $extraDirs = [
         if ('\\' === \DIRECTORY_SEPARATOR) {
             $pathExt = getenv('PATHEXT');
             $suffixes = array_merge($pathExt ? explode(\PATH_SEPARATOR, $pathExt) : $this->suffixes, $suffixes);
+        } else {
+            $suffixes = array_merge($suffixes, $this->suffixes);
         }
```
nicolas-grekas: As @stof said, we need to account for the listed suffixes on Windows too when the `PATHEXT` env var is set. Then we have to decide on the best priority. I suggest adding the listed suffixes last:

```suggestion
        if ('\\' === \DIRECTORY_SEPARATOR && $pathExt = getenv('PATHEXT')) {
            $suffixes = array_merge(explode(\PATH_SEPARATOR, $pathExt), $suffixes);
        }
        $suffixes = array_merge($suffixes, $this->suffixes);
```

nicolas-grekas: Ah, and one more thing: we should add a constructor that sets the default to the empty array when not on windows (or populates the array on windows)

TravisCarden: Like so, @nicolas-grekas? 1651d3f7b1590d03e0e1921369c91d3a8048cecc

TravisCarden:
It doesn't look like the CI failures are from my changes. Requesting re-review.

fabpot (approved):

fabpot (changes requested):
Looking at the CI failure, it does seem that tests are broken by this change:

```
1) Symfony\Component\Process\Tests\ExecutableFinderTest::testFindWithDefaultSuffix
Failed asserting that null matches expected '/home/runner/work/symfony/symfony/src/Symfony/Component/Process/Tests/Fixtures/executable_with_a_default_suffix.bat'.
```

TravisCarden:
Oh, of course. Because that test asserts on behavior that is now Windows-only. I've just removed it. I'll go ahead and rebase, while I'm at it.

fabpot (approved):

fabpot:
Thank you @TravisCarden.

Issue #49710, opened 2023-03-16
Serializer throw exception when Ignore attribute is used


ludekbenedik:
### Symfony version(s) affected

5.4.21

### Description

Hi,
the serializer throws an exception when the `Ignore` attribute is used and an entity contains a private getter/isser.

### How to reproduce

This is OK:

```php
class EntityWithSerializerIgnore
{
    public function getFoo(): int { return 1; }
    public function getBar(): int { return 2; }
    private function isSomething(): bool { return true; }
}

$entity = new EntityWithSerializerIgnore();
$serializer->serialize($entity, 'json'); // {"foo":1,"bar":2}
```

This throws an exception:

```php
class EntityWithSerializerIgnore
{
    public function getFoo(): int { return 1; }
    #[Ignore]
    public function getBar(): int { return 2; }
    private function isSomething(): bool { return true; }
}

$entity = new EntityWithSerializerIgnore();
$serializer->serialize($entity, 'json'); // throws: Can't get a way to read the property "something" in class "EntityWithSerializerIgnore".
```

### Possible Solution

The `AbstractNormalizer::getAllowedAttributes` method returns all attributes (public, private) loaded by the `AnnotationLoader` when the `Ignore` attribute is used. But the `ObjectNormalizer` load only public attributes (methods).

The `AbstractObjectNormalizer::getAttributes` method cannot returns all allowed attributes, but have to combine these attributes with extracted attributes.

xabbuh:
Can you create a small example application that allows to reproduce your issue?

ludbene:
src/Controller/SerializerBugExampleController.php

```php
<?php declare(strict_types = 1);

namespace App\Controller;

use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
use Symfony\Component\HttpFoundation\JsonResponse;
use Symfony\Component\Routing\Annotation\Route;
use Symfony\Component\Serializer\Annotation\Ignore;
use Symfony\Component\Serializer\SerializerInterface;

class SerializerBugExampleController extends AbstractController
{
    #[Route('/serializer-bug-example', name: 'serializer_bug_example')]
    public function __invoke(SerializerInterface $serializer): JsonResponse
    {
        return new JsonResponse($serializer->serialize(new SerializerBugExampleEntity(), 'json'), 200, [], true);
    }
}

class SerializerBugExampleEntity
{
    public function getFoo(): int { return 1; }
    #[Ignore]
    public function getBar(): int { return 2; }
[...]

PR #52680, opened 2023-11-22
[Serializer] Fix access to private properties/getters when using the `@Ignore` annotation

### LINKED ISSUE #52673: [Serializer] All fields are not included by default, but upon adding any attribute, all fields are then required

6.3.8

### Description

According to https://symfony.com/doc/current/components/serializer.html#ignoring-attributes all properties are included by default, which makes sense. However, this doesn't seem to be the case at least for traits and abstract classes.

When we don't add the `#[Ignore]` attribute in the `TestTrait` in the reproducer, everything works fine. However, adding that attribute causes all of the other properties to now be seen by the serializer and then fail because there is no public access to them via getters or via public properties.

### How to reproduce

https://github.com/nesl247/symfony-serialization-ignore-bug

1) git clone https://github.com/nesl247/symfony-serialization-ignore-bug.git
2) composer install
3) bin/console test:command

### Possible Solution

_No response_

### Additional Context

```[critical] Error thrown while running command "test:command". Message: "Can't get a way to read the property "testProperty2" in class "App\TestMessage"."

In PropertyAccessor.php line 456:
                                                                                    
  Can't get a way to read the property "testProperty2" in class "App\TestMessage".  
                                                                                    
```

mtarld:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52673 #49710
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Serializer/Tests/Normalizer/AbstractObjectNormalizerTest.php
+++ b/src/Symfony/Component/Serializer/Tests/Normalizer/AbstractObjectNormalizerTest.php
@@ -15,6 +15,7 @@ use Doctrine\Common\Annotations\AnnotationReader;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\PropertyInfo\Extractor\PhpDocExtractor;
 use Symfony\Component\PropertyInfo\Type;
+use Symfony\Component\Serializer\Annotation\Ignore;
 use Symfony\Component\Serializer\Exception\ExtraAttributesException;
 use Symfony\Component\Serializer\Exception\InvalidArgumentException;
 use Symfony\Component\Serializer\Exception\LogicException;
@@ -444,6 +445,14 @@ class AbstractObjectNormalizerTest extends TestCase
         $normalizedData = $normalizer->normalize(new EmptyDummy(), 'any', ['preserve_empty_objects' => true]);
         $this->assertEquals(new \ArrayObject(), $normalizedData);
     }
+
+    public function testNormalizeWithIgnoreAnnotationAndPrivateProperties()
+    {
+        $classMetadataFactory = new ClassMetadataFactory(new AnnotationLoader(new AnnotationReader()));
+        $serializer = new Serializer([new ObjectNormalizer($classMetadataFactory)]);
+
+        $this->assertSame(['foo' => 'foo'], $serializer->normalize(new ObjectDummyWithIgnoreAnnotationAndPrivateProperty()));
+    }
 }
 
 class AbstractObjectNormalizerDummy extends AbstractObjectNormalizer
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/Serializer/Tests/Normalizer/AbstractObjectNormalizerTest.php
```diff
@@ -484,6 +493,16 @@ class EmptyDummy
 {
 }
 
+class ObjectDummyWithIgnoreAttributeAndPrivateProperty
```
OskarStark: Annotation instead of Attribute?

src/Symfony/Component/Serializer/Tests/Normalizer/AbstractObjectNormalizerTest.php
```diff
@@ -444,6 +445,14 @@ public function testNormalizeEmptyObject()
         $normalizedData = $normalizer->normalize(new EmptyDummy(), 'any', ['preserve_empty_objects' => true]);
         $this->assertEquals(new \ArrayObject(), $normalizedData);
     }
+
+    public function testNormalizeWithIgnoreAttributeAndPrivateProperties()
```
OskarStark:
```suggestion
    public function testNormalizeWithIgnoreAnnotationAndPrivateProperties()
```
Sorry, I missed it on my first review

mtarld: Aha, it's ok, I missed that as well!

nesl247:
@mtarld am I misunderstanding the documentation though? It says

> All attributes are included by default when serializing objects. There are two options to ignore some of those attributes.

Which to mean means even private and protected. While I think the fix here makes sense, it doesn't align with the documentation. Should the documentation also be updated to say only publicly accessible attributes are included? Or is that what `attribute` means and I am missing the definition?

stof:
Private or protected properties are included if they have getters to access them publicly.

nesl247:
> Private or protected properties are included if they have getters to access them publicly.

That's what I thought. Then the documentation should be updated to specify that instead of all attributes correct?

mtarld:
Indeed, we could go with something like: "All **accessible** attributes are included by default when serializing objects.".

WDYT? Up for a PR?

nesl247:
> Indeed, we could go with something like: "All **accessible** attributes are included by default when serializing objects.".
> 
> WDYT? Up for a PR?

Yes. I’m out of the office until Tuesday, but when I get back I’m happy to submit a PR.

nicolas-grekas (approved):

nicolas-grekas:
> PHP Parse error:  syntax error, unexpected 'string' (T_STRING), expecting function (T_FUNCTION) or const (T_CONST) in /home/runner/work/symfony/symfony/src/Symfony/Component/Serializer/Tests/Normalizer/AbstractObjectNormalizerTest.php on line 498

mtarld:
Failures should be fixed in https://github.com/symfony/symfony/pull/52713

nicolas-grekas:
Thank you @mtarld.

PR #36695, opened 2020-05-05
[Serializer] PropertyNormalizer doesn't support DiscriminatorMap

agven:
| Q             | A
| ------------- | ---
| Branch?       | 4.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #36585 
| License       | MIT
| Doc PR        |

PR #52681, opened 2023-11-22
[Serializer] Fix support for DiscriminatorMap in PropertyNormalizer

### LINKED ISSUE #36585: [Serializer] PropertyNormalizer doesn't support DiscriminatorMap
**Symfony version(s) affected**: 4.4.7 (probably lower and higher also)

**Description**  

Could someone explain me why PropertyNormalizer doesn't support DiscriminatorMap ?

Everything works fine if I use ObjectNormalizer but if I choose PropertyNormalizer instead of ObjectNormalizer and do serialization or desialization as a result this functionality doesn't work. However both normalizers have one common parent class this is AbstractObjectNormalizer and if you look at the constructor declaration you will see a dependency on ClassDiscriminatorResolverInterface but PropertyNormalizer doesn't support a "discriminator class mapping". 

As seems to me something is wrong with Barbora-Liskov substitution principle. Am I right, or were there reasons for that ?

**How to reproduce**  

```php
// ...
use Symfony\Component\Serializer\Annotation\DiscriminatorMap;
use Symfony\Component\Serializer\Mapping\ClassDiscriminatorFromClassMetadata;
use Symfony\Component\Serializer\Mapping\Factory\ClassMetadataFactory;
use Symfony\Component\Serializer\Mapping\Loader\AnnotationLoader;
use Symfony\Component\Serializer\Normalizer\PropertyNormalizer;

/**
 * @DiscriminatorMap(typeProperty="type", mapping={
 *    "github"="GitHubCodeRepository",
 *    "bitbucket"="BitBucketCodeRepository"
 * })
 */
interface CodeRepository {}

class GitHubCodeRepository implements CodeRepository 
{
    private $url;
}

class BitBucketCodeRepository implements CodeRepository 
{
    private $url;
}

$classMetadataFactory = new ClassMetadataFactory(new AnnotationLoader(new AnnotationReader()));
$discriminator = new ClassDiscriminatorFromClassMetadata($classMetadataFactory);

$serializer = new Serializer(
    [new PropertyNormalizer($classMetadataFactory, null, null, $discriminator)],
    ['json' => new JsonEncoder()]
);

$serialized = $serializer->serialize(new GitHubCodeRepository(), 'json');
// actual: {"type":null, "url":"null"}
// expected: {"type":"github"}

$repository = $serializer->deserialize($serialized, CodeRepository::class, 'json');
// actual:
//     exception: \Symfony\Component\Serializer\Exception\NotNormalizableValueException
//     message: Could not denormalize object of type "CodeRepository", no supporting normalizer found.
```

mtarld:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #36585
| License       | MIT

Follow up of #36695

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
@@ -179,8 +179,15 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
 
             $attributeContext = $this->getAttributeNormalizationContext($object, $attribute, $context);
 
+            $discriminatorProperty = null;
+            if (null !== $this->classDiscriminatorResolver && null !== $mapping = $this->classDiscriminatorResolver->getMappingForMappedObject($object)) {
+                $discriminatorProperty = $mapping->getTypeProperty();
+            }
+
             try {
-                $attributeValue = $this->getAttributeValue($object, $attribute, $format, $attributeContext);
+                $attributeValue = $attribute === $discriminatorProperty
+                    ? $this->classDiscriminatorResolver->getTypeForMappedObject($object)
+                    : $this->getAttributeValue($object, $attribute, $format, $attributeContext);
             } catch (UninitializedPropertyException $e) {
                 if ($context[self::SKIP_UNINITIALIZED_VALUES] ?? $this->defaultContext[self::SKIP_UNINITIALIZED_VALUES] ?? true) {
                     continue;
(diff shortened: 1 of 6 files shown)


src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
```diff
@@ -787,4 +794,19 @@ private function isUninitializedValueError(\Error $e): bool
             && str_starts_with($e->getMessage(), 'Typed property')
             && str_ends_with($e->getMessage(), 'must not be accessed before initialization');
     }
+
+    protected function getDiscriminatorProperty(object $object): ?string
```
nicolas-grekas: private? why protected?

src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
```diff
@@ -37,6 +37,9 @@
  */
 abstract class AbstractObjectNormalizer extends AbstractNormalizer
 {
+    /** @var array<string, string|null> */
+    private array $discriminatorCache = [];
```
nicolas-grekas: there's already a cache in ClassDiscriminatorFromClassMetadata to maybe we don't need this one?
maybe this could mean we can remove the added protected method and inline the logic instead?

mtarld: Indeed, the less protected method we have, the better. It kind of duplicate the code a bit but IMHO it acceptable.

nicolas-grekas:
Thank you @mtarld.

PR #46680, opened 2022-06-15
[Serializer] Provide context information from attribute for promoted properties

DanielBadura:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | -

I refactored some classes to using property promotion wiht `#[Context]` attributes and my tests started to fail since the context was missing then. So i tried to retrieve this context at the `AbstractNormalizer` and added a testcase for that.

I'm not sure if this is the right place for getting this metadata information and if i have overseen something.

PR #52683, opened 2023-11-22
[Serializer] Fix constructor deserialization path

### LINKED ISSUE #44925: [Serializer] deserialization_path missing in custom DenormalizerInterface upon using object's constructor

5.4 and above

### Description

In denormalizing to object by constructor, symfony serializer is not sending `deserialization_path` in $context.
In AbstractObjectNormalizer there is foreach which goes through all attributes and creating context `(symfony/serializer/Normalizer/AbstractObjectNormalizer.php:374)`
```
foreach ($normalizedData as $attribute => $value) {
$attributeContext = $this->getAttributeDenormalizationContext($resolvedClass, $attribute, $context);
```
```
private function getAttributeDenormalizationContext(string $class, string $attribute, array $context): array
{
    // (...)
     $context['deserialization_path'] = ($context['deserialization_path'] ?? false) ? $context['deserialization_path'].'.'.$attribute : $attribute;
    // (...)
}
  ```
  
And it works fine when we want to denormalize by properties etc.
Problem occurs if object has only constructor available, then Serializer fires AbstractNormalizer.
In AbstractNormalizer there is foreach which goes through construct arguments 
`(symfony/serializer/Normalizer/AbstractNormalizer.php:362)`
```
foreach ($constructorParameters as $constructorParameter) {
    $paramName = $constructorParameter->name;
    $key = $this->nameConverter ? $this->nameConverter->normalize($paramName, $class, $format, $context) : $paramName;
```
`key` would be equivalent to `attribute` from AbstractObjectNormalizer, hence the question why not add `key` to context as `deserialization_path` as construct arguments needs to be same as class attributes.
When you want to use [Collecting Type Errors While Denormalizing](https://symfony.com/doc/5.4/components/serializer.html#collecting-type-errors-while-denormalizing) path is empty, so you cannot return info which property had incorrect type.

### How to reproduce

To reproduce you will need serializer, annotations and property access
```
composer require symfony/serializer
composer require doctrine/annotations
composer require symfony/property-access
```

example.php:
```php
<?php
// composer require symfony/serializer
// composer require doctrine/annotations
// composer require symfony/property-access

use Symfony\Component\Serializer\Exception\NotNormalizableValueException;
use Symfony\Component\Serializer\Mapping\Factory\ClassMetadataFactory;
use Symfony\Component\Serializer\Mapping\Loader\AnnotationLoader;
use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
use Symfony\Component\Serializer\Normalizer\ObjectNormalizer;

require 'vendor/autoload.php';

class ExampleObject
{
    private int $number;

    public function __construct(int $number)
    {
        $this->number = $number;
    }
}

class ExampleDTO
{
    private ExampleObject $exampleObject;

    public function __construct(ExampleObject $exampleObject)
    {
        $this->exampleObject = $exampleObject;
    }
}

$classMetadataFactory = new ClassMetadataFactory(new AnnotationLoader(new AnnotationReader()));
$serializer = new Serializer([new ObjectNormalizer($classMetadataFactory)],['json' => new JsonEncoder()]);

try {
    $data = $serializer->deserialize(
        '{"exampleObject":"test"}',
        ExampleDTO::class,
        'json',
        [
            DenormalizerInterface::COLLECT_DENORMALIZATION_ERRORS => true
        ]
    );
    print_r($data);
} catch (\Symfony\Component\Serializer\Exception\PartialDenormalizationException $e) {
    // as seen at https://symfony.com/doc/5.4/components/serializer.html#collecting-type-errors-while-denormalizing
    /** @var NotNormalizableValueException */
    foreach ($e->getErrors() as $exception) {
        echo sprintf('The type must be one of "%s" ("%s" given). in: "%s"',
                implode(', ', $exception->getExpectedTypes()),
                $exception->getCurrentType(),
                $exception->getPath()
            ) . PHP_EOL;
        // Outputs: The type must be one of "unknown" ("array" given). in: "" <- no path and expected type is unknown
    }
}
```
Above example will output `The type must be one of "unknown" ("array" given). in: "" <- no path and expected type is unknown`

### Possible Solution

For the path in `\Symfony\Component\Serializer\Normalizer\AbstractNormalizer::instantiateObject` add at the beginning of constructor arguments foreach or same context build as in AbstractObjectNormalizer:

```php
$context['deserialization_path'] = $key;
```

As for the "unknown" I have not looked into it.

### Additional Context

I use DTOs in my project which have custom denormalizers which shows same issue.
Also I have noticed issue when collecting errors that despite mismatch of constructor arguments Serializer Component will try to create object resulting in `\Symfony\Component\Serializer\Exception\ExceptionInterface` when using custom Denormalizer which throws `NotNormalizableValueException` in `denormalize` method and I think it is a bit odd.

mtarld:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #44925
| License       | MIT

This fix doesn't have to be upmerged because it already has been covered by https://github.com/symfony/symfony/pull/46680.

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
@@ -489,6 +489,8 @@ abstract class AbstractNormalizer implements NormalizerInterface, DenormalizerIn
      */
     protected function createChildContext(array $parentContext, string $attribute, ?string $format): array
     {
+        $parentContext['deserialization_path'] = ($parentContext['deserialization_path'] ?? false) ? $parentContext['deserialization_path'].'.'.$attribute : $attribute;
+
         if (isset($parentContext[self::ATTRIBUTES][$attribute])) {
             $parentContext[self::ATTRIBUTES] = $parentContext[self::ATTRIBUTES][$attribute];
         } else {


nicolas-grekas:
Thank you @mtarld.

PR #52685, opened 2023-11-22
[Cache] remove Doctrine DBAL < 3.3 related code

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

following the merge up of #52459

### DIFF
--- a/src/Symfony/Component/Cache/Adapter/DoctrineDbalAdapter.php
+++ b/src/Symfony/Component/Cache/Adapter/DoctrineDbalAdapter.php
@@ -375,8 +375,7 @@ class DoctrineDbalAdapter extends AbstractAdapter implements PruneableInterface
             return $this->serverVersion;
         }
 
-        // The condition should be removed once support for DBAL <3.3 is dropped
-        $conn = method_exists($this->conn, 'getNativeConnection') ? $this->conn->getNativeConnection() : $this->conn->getWrappedConnection();
+        $conn = $this->conn->getNativeConnection();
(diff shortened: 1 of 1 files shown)


src/Symfony/Component/Cache/Adapter/DoctrineDbalAdapter.php
```diff
@@ -375,8 +375,7 @@ private function getServerVersion(): string
             return $this->serverVersion;
         }
 
-        // The condition should be removed once support for DBAL <3.3 is dropped
-        $conn = method_exists($this->conn, 'getNativeConnection') ? $this->conn->getNativeConnection() : $this->conn->getWrappedConnection();
+        $conn = $this->conn->getNativeConnection();
         if ($conn instanceof ServerInfoAwareConnection) {
```
derrabus, line 379: Hmm, this does not really make sense. If we call `getNativeConnection()`, we get the native handle of the PHP extension used, e.g. a `mysqli` instance or an `sqlsrv` resource. Those objects/resources won't ever implement `ServerInfoAwareConnection`. Something's off here.

xabbuh: I guess you are right. Looks like we first need to fix it in a way similar to #52715.

derrabus:
My alternative proposal: #52720.

xabbuh:
closing in favour of #52720

PR #52688, opened 2023-11-22
[Cache] Add url decoding of password in `RedisTrait` DSN

### LINKED ISSUE #52420: [Cache] The `RedisTrait` DSN parsing doesn't `urldecode` the params

5.4

### Description

The [RedisTrait doesn't urldecode](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Cache/Traits/RedisTrait.php#L101-L112) the username/password/host when creating the connection from the dsn. In most other components, params are decoded (in the [Mailer](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Mailer/Transport/Dsn.php#L52-L53) for example). That makes it hard to use special chars.
It would be easy enough to fix this but I'm affraid this would impact users that have character sequences that match url-encoded signs (for instance a raw password `password%3D` would be converted to `password=`).

Is this something we could fit in 7.0 ?

### How to reproduce

Use redis with a password containing à `@` char. 
Using no encoding the parsing will be wrong 
`redis://user:p@ssword@localhost`
If encoded the auth will fail
`redis://user:p%40assword@localhost`

### Possible Solution

_No response_

### Additional Context

_No response_

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #52420
| License       | MIT

This brings support for `@` char in passwords

### DIFF
--- a/src/Symfony/Component/Cache/Tests/Traits/RedisTraitTest.php
+++ b/src/Symfony/Component/Cache/Tests/Traits/RedisTraitTest.php
@@ -15,15 +15,11 @@ use PHPUnit\Framework\SkippedTestSuiteError;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Cache\Traits\RedisTrait;
 
+/**
+ * @requires extension redis
+ */
 class RedisTraitTest extends TestCase
 {
-    public static function setUpBeforeClass(): void
-    {
-        if (!getenv('REDIS_CLUSTER_HOSTS')) {
-            throw new SkippedTestSuiteError('REDIS_CLUSTER_HOSTS env var is not defined.');
-        }
-    }
-
     /**
      * @dataProvider provideCreateConnection
      */
@@ -40,6 +36,19 @@ class RedisTraitTest extends TestCase
         $connection = $mock::createConnection($dsn);
 
         self::assertInstanceOf($expectedClass, $connection);
+    }
+
+    public function testUrlDecodeParameters()
+    {
+        if (!getenv('REDIS_AUTHENTICATED_HOST')) {
+            self::markTestSkipped('REDIS_AUTHENTICATED_HOST env var is not defined.');
+        }
+
+        $mock = self::getObjectForTrait(RedisTrait::class);
+        $connection = $mock::createConnection('redis://:p%40ssword@'.getenv('REDIS_AUTHENTICATED_HOST'));
+
+        self::assertInstanceOf(\Redis::class, $connection);
+        self::assertSame('p@ssword', $connection->getAuth());
     }
 
     public static function provideCreateConnection(): array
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
Why not as a bug fix? This looks like one to me. For 5.4 then.

src/Symfony/Component/Cache/Tests/Traits/RedisTraitTest.php
```diff
@@ -14,15 +14,11 @@
...
 
+/**
+ * @requires extension redis
+ */
 class RedisTraitTest extends TestCase
 {
-    public static function setUpBeforeClass(): void
-    {
-        if (!getenv('REDIS_CLUSTER_HOSTS')) {
-            self::markTestSkipped('REDIS_CLUSTER_HOSTS env var is not defined.');
```
nicolas-grekas: but we still need the env var to run all the other test cases, don't we?

src/Symfony/Component/Cache/Traits/RedisTrait.php
```diff
@@ -369,9 +369,9 @@ public static function createConnection(#[\SensitiveParameter] string $dsn, arra
                 if (\is_array($auth)) {
                     // ACL
                     $params['parameters']['username'] = $auth[0];
-                    $params['parameters']['password'] = $auth[1];
+                    $params['parameters']['password'] = urldecode($auth[1]);
                 } else {
-                    $params['parameters']['password'] = $auth;
+                    $params['parameters']['password'] = urldecode($auth);
```
nicolas-grekas: everywhere this?
```suggestion
                    $params['parameters']['password'] = rawurldecode($auth);
```

src/Symfony/Component/Cache/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Add url decoding of password in `RedisTrait` DSN
```
nicolas-grekas: not needed for bug fixes

alexandre-daubois: Oh alright, I thought it was considered as a feature. I'll update this as well as the base branch

src/Symfony/Component/Cache/Tests/Traits/RedisTraitTest.php
alexandre-daubois: Yes, but this env var presence is already checked in `testCreateConnection` so it can be removed from here

src/Symfony/Component/Cache/Traits/RedisTrait.php
alexandre-daubois: Good idea!

alexandre-daubois:
I addressed your comments and rebased on 5.4. The fix is different from other branches, I'll provide a fix for those if needed :slightly_smiling_face:

nicolas-grekas:
Thank you @alexandre-daubois.

nicolas-grekas:
> I addressed your comments and rebased on 5.4. The fix is different from other branches, I'll provide a fix for those if needed 🙂

No need, that was easy to fix as part of the merge conflict.

Note that I didn't change CouchbaseCollectionAdapter nor CouchbaseBucketAdapter because while they have the same issue, they're more severely broken and would deserve a dedicated PR, if anyone is up to have a look ;)

PR #52689, opened 2023-11-22
[DependencyInjection] Fix tests on configuration prepend

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

### DIFF
--- a/src/Symfony/Component/HttpKernel/Tests/Fixtures/AcmeFooBundle/AcmeFooBundle.php
+++ b/src/Symfony/Component/HttpKernel/Tests/Fixtures/AcmeFooBundle/AcmeFooBundle.php
@@ -31,7 +31,7 @@ class AcmeFooBundle extends AbstractBundle
 
     public function prependExtension(ContainerConfigurator $container, ContainerBuilder $builder): void
     {
-        $container->extension('loaded', ['bar' => 'baz'], true);
+        $builder->prependExtensionConfig('loaded', ['bar' => 'baz']);
     }
 
     public function loadExtension(array $config, ContainerConfigurator $container, ContainerBuilder $builder): void
--- a/src/Symfony/Component/HttpKernel/Tests/DependencyInjection/MergeExtensionConfigurationPassTest.php
+++ b/src/Symfony/Component/HttpKernel/Tests/DependencyInjection/MergeExtensionConfigurationPassTest.php
@@ -48,7 +48,7 @@ class MergeExtensionConfigurationPassTest extends TestCase
         $configPass = new MergeExtensionConfigurationPass(['loaded', 'acme_foo']);
         $configPass->process($container);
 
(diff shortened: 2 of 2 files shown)


yceruto (approved):

src/Symfony/Component/HttpKernel/composer.json
```diff
@@ -29,7 +29,7 @@
         "symfony/config": "^6.4|^7.0",
         "symfony/console": "^6.4|^7.0",
         "symfony/css-selector": "^6.4|^7.0",
-        "symfony/dependency-injection": "^6.4|^7.0",
+        "symfony/dependency-injection": "^7.1",
```
nicolas-grekas: is there a way to prevent this? I'd like to preserve compat with 6.4 for as long as possible

alexandre-daubois: I pushed an alternate solution. We can actually remove this bit that depends on the new `prepend` argument as well as the test assertion that goes with it. I think it is safe to do so because the `prepend` argument is tested here:

https://github.com/symfony/symfony/blob/638f7d0840e838ef5183eeab622892f56c3cd8b2/src/Symfony/Component/DependencyInjection/Tests/Extension/AbstractExtensionTest.php#L57-L79

What do you think?

nicolas-grekas: indeed - the issue was having the tests for the new feature in another component

src/Symfony/Component/HttpKernel/Tests/Fixtures/AcmeFooBundle/AcmeFooBundle.php
```diff
@@ -31,7 +31,7 @@ public function configure(DefinitionConfigurator $definition): void
 
     public function prependExtension(ContainerConfigurator $container, ContainerBuilder $builder): void
     {
-        $container->extension('loaded', ['bar' => 'baz'], true);
+        $container->extension('loaded', ['bar' => 'baz']);
```
yceruto: I read the previous discussion, but if we don't pass the prepend `true` we are not prepending that config, so I suggest use `$builder->prepend*` instead, or just remove the `prependExtension()` altogether if it's no longer tested or used.

alexandre-daubois: I think this is a better solution indeed. We're testing the `MergeExtensionConfigurationPass` so it makes sense to me to have as much assertions as possible on these cases, including on `prependExtension()`.

:+1: on this :slightly_smiling_face:

yceruto (approved):

nicolas-grekas:
Thank you @alexandre-daubois.

PR #52690, opened 2023-11-22
[Cache] Fix return type of `Redis6Proxy::mget()`

### LINKED ISSUE #52668: [Cache] \Redis::mget() can return bool, Redis6Proxy::mget() doesn't account for that

6.3.5

### Description

`Redis::mget` can return false when no keys are provided. https://github.com/phpredis/phpredis/issues/1810

This is annoying because the signature says `mget(): \Redis|array`, but it's actually `mget(): \Redis|array|false`.

Redis6Proxy::mget has the correct signature, but it doesn't handle the possible false.

### How to reproduce

Set up PHPRedis v6 and do `Redis6Proxy::mget([])`

### Possible Solution

Return [] if false.

### Additional Context

IMO this is a bug in the module, but it's a bug since 2020.

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52668
| License       | MIT

This class is internal, the return type can be safely updated. For the record, the `false` return type is already added on `Redis6Proxy::hmget()`.

### DIFF
--- a/src/Symfony/Component/Cache/Traits/Redis6Proxy.php
+++ b/src/Symfony/Component/Cache/Traits/Redis6Proxy.php
@@ -651,7 +651,7 @@ class Redis6Proxy extends \Redis implements ResetInterface, LazyObjectInterface
         return ($this->lazyObjectState->realInstance ??= ($this->lazyObjectState->initializer)())->ltrim(...\func_get_args());
     }
 
-    public function mget($keys): \Redis|array
+    public function mget($keys): \Redis|array|false
     {
         return ($this->lazyObjectState->realInstance ??= ($this->lazyObjectState->initializer)())->mget(...\func_get_args());
     }


nicolas-grekas:
We cannot widen the return type before the extension does it. The CI failures are correct.

alexandre-daubois:
Alright, I was looking at an outdated class stub of the extension indeed. Let's close for now

dkarlovi:
The in between fix might be to test for false and return []

alexandre-daubois:
I don't think so, because the extension cannot, currently, return false. That's the problem. So we can't test that either 😕

dkarlovi:
The extension can return false, that's the whole bug here. 🤔

alexandre-daubois:
It doesn't seem: https://github.com/phpredis/phpredis/blob/ff305349dba87ab857a8f28acbc3b22af5a271cc/redis.stub.php#L2178

As long as this class isn't updated, we cannot widen the return type as stated by Nicolas. Maybe I'm looking at the wrong stub however?

dkarlovi:
No, that's correct, the stub is wrong, it's mentioned in the bug. :) It returns false on error even though it's not in the stub. Check the bug and the PHPRedis bug linked within. :smile:

alexandre-daubois:
My bad, got it now :+1: Thanks for taking this time. I cannot reopen this PR, so I created this new one: https://github.com/symfony/symfony/pull/52700

PR #52695, opened 2023-11-23
[Mime] Allow array on `setParameter` `$value` parameter

### LINKED ISSUE #52696: Bug: Unable to send array on setParameter method

6.3.*

### Description
When attempting to pass an array of objects an error is thrown that it should contain a string and not an array. Using json_encode() to encapsulate the array resolves the problem. However, Brevo expects the value to be an array, not a string.

### How to reproduce

1. Create a new email, set the sender and send email.

```php
$message = (new Email())
   ->From(...)
   // ...more code here
  
   ->getHeaders()
      ->addTextHeader('templateId', $template)
      ->addParameterizedHeader('params', 'params', $params);

$this->mailer->send($message);
```

2. Passing an array instead of string or null value.
```
$param_one = [];

$params = [
            ...more params here,
            'param_one' => $param_one,
        ];
``` 

### Expected Behavior
Allow an array to be passed in the parameter other than string and null value.

### Actual Behavior
Error is thrown when we passed an array on the parameter instead of string and null value.

### Possible Solution
Removing the type declaration for the $value to avoid the error being thrown. 
Reference solution: https://github.com/symfony/symfony/pull/52695

botjaeger:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Issues        | Fix #52696
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Mime/Header/ParameterizedHeader.php
+++ b/src/Symfony/Component/Mime/Header/ParameterizedHeader.php
@@ -41,7 +41,10 @@ final class ParameterizedHeader extends UnstructuredHeader
         }
     }
 
-    public function setParameter(string $parameter, ?string $value): void
+    /**
+     * params null|string|array $value
+     */
+    public function setParameter(string $parameter, $value): void
     {
         $this->setParameters(array_merge($this->getParameters(), [$parameter => $value]));
     }


src/Symfony/Component/Mime/Header/ParameterizedHeader.php
```diff
@@ -41,7 +41,10 @@ public function __construct(string $name, string $value, array $parameters = [])
         }
     }
 
-    public function setParameter(string $parameter, ?string $value): void
+    /**
+     * params null|string|array $value
```
ghost, line 45: the doc block is redundant, you can use native union type hint

ghost (changes requested):

nicolas-grekas:
I'm sorry but this doesn't make sense to me. An array cannot be sent as is on the network. At some point, something needs to serialize it. The mime component doesn't know how to serialize this array. If you need to send many values for some header, you need to generate the serialized string on your side before passing it to setParameter.
IIUC.

xabbuh:
I am closing as explained before. Thank you for understanding.

Issue #49924, opened 2023-04-04
[Serializer] TypeResolver update breaks when non-collection generics are used


schodemeiss:
### Symfony version(s) affected

6.2.8

### Description

Since removing the TypeResolver 1.7 restriction [in this commit](https://github.com/symfony/symfony/commit/77e9020576b2708ba9c5543196ba7f41f9d2e501), I'm getting serialisation errors when classes have docblocks that are not collections.

For example, if I use Psalms "value-of" docblock, the Serializer thinks this is an array of Type, but in reality it's just a string.

### How to reproduce

```php
final class StatusClass
{
    final public const STATUS_VALUES = ['accept', 'reject'];

    /**
     * @param value-of<self::STATUS_VALUES> $status
     */
    public function __construct(
        public readonly string $status,
    ) {
    }
}
```

Attempt to serialise data that looks a little like this and you'll receive an error that $status is an "array".

### Possible Solution

1. Replace in user-land code @param with @psalm-param. Psalm still understands the type, but the serializer doesn't.

2. Re-add the composer restriction until this issue in TypeResolver is resolved: https://github.com/phpDocumentor/TypeResolver/issues/180

carsonbot:
Hey, thanks for your report!
There has not been a lot of activity here for a while. Is this bug still relevant? Have you managed to find a workaround?

schodemeiss:
> Hey, thanks for your report! There has not been a lot of activity here for a while. Is this bug still relevant? Have you managed to find a workaround?

Yes, still happening today on any version of the Serializer.

mtarld:
Hey!
I think this is the same issue as https://github.com/symfony/symfony/issues/41996.

The PR https://github.com/symfony/symfony/pull/52699 may fix it.

PR #52699, opened 2023-11-23
[Serializer] [PropertyAccessor] Ignore non-collection interface generics

### LINKED ISSUE #41996: [Serializer] Allow using generic annotations and not parsing them as an array
**Symfony version(s) affected**: 5.3.3

**Description**
Denormalizer breaks with an error `Data expected to be an array, string given.` when it handles an object, which contains  a generic-like annotation, but is not an array or iterable.

**How to reproduce**
```php
interface Foo {}

/** @template TFoo of Foo */
interface Bar {}

final class FooBar {
    /** @var Bar<Foo> */
    public Bar $bar; // This prop will fail
    // ...
}
```

**Possible Solution**
I digged it in PhpDocTypeHelper, which is part of phpDocumentor, so I don't know how to solve it. :worried:

mtarld:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #41996 #49924
| License       | MIT

`PhpDocExtractor` and `PhpStanDocExtractor` are following an old version of PSR-5 with used to define collections as the following:
```
generic = collection-type "<" [type-expression "," *SP] type-expression ">"
```
But, it does conflict with non-collection generics.

This issue will automatically be solved if the `TypeInfo` is merged in Symfony. But for older versions (<7.1 at least), it needs a fix.

~I was not able to find a proper bug fix without introducing a BC break, so this PR proposes to opt-on the "non-collection generics" parsing, such as `stcClass<int>` for example.~

~To opt-out from parsing these generics, it'll be required to set `ignore_docblock_generics` in the context. And this key/value will become obsolete as soon as the `TypeInfo` is introduced.~

~I'm not sure how to proceed though, should we only merge it in 5.4, and 6.3 supposing that the `TypeInfo` might be merged in 7.x? Should we document it only in these branches?~

EDIT: I finally ignored PHPDoc generics that aren't well known collection generic types so that the process will fall back to other resolvers (such as reflection resolver for example)


nicolas-grekas:
What about considering this as an improvement and merging into 6.4? Then we can do the BC break in 7.0.
I'd be fine saying those annotations aren't supported in 5.4 personnally.

mtarld:
It also means that it won't be supported in 6.3 and 7.0 (supposing that the `TypeInfo` will come in 7.1). But I'm fine with that as well, as 6.4 matters the most, being an LTS.

maxhelias (approved):

src/Symfony/Component/PropertyInfo/Util/PhpDocTypeHelper.php
```diff
@@ -115,10 +115,17 @@ private function createType(DocType $type, bool $nullable, ?string $docType = nu
 
             [$phpType, $class] = $this->getPhpTypeAndClass((string) $fqsen);
 
+            $collection = \in_array($class, [\Traversable::class, \Iterator::class, \IteratorAggregate::class, \ArrayAccess::class, \Generator::class], true);
```
stof: shouldn't we at least support the Collection and ReadableCollection interfaces of doctrine/collections there, to limit the impact of the BC break ?

mtarld: I think that by doing the following: `$collection = \is_a($class, \Traversable::class, true) || \is_a($class, \ArrayAccess::class, true);`, most of the collections will be handled properly.

mtarld: Similar is being done in https://github.com/symfony/symfony/pull/54661

fabpot (approved):

fabpot:
Thank you @mtarld.

xabbuh:
@mtarld I have merged these changes up to the `7.1` branch which now fails with one error. I tried to find out what we need to change but wasn't able to resolve it. In case you want to take a look at it. :)

mtarld:
@xabbuh, indeed new tests must be updated like the current ones. I created PR fixing it: https://github.com/symfony/symfony/pull/57530.
Thanks for reaching me :slightly_smiling_face:

soyuka:
HI, since that we're observing a `collection: false` instead of `true` for properties typed as: 

```
\Doctrine\Common\Collections\ArrayCollection<Foo>
```

`is_a(ArrayCollection::class, \ArrayAccess::class, true)` is `true` not sure why it fails, any idea @mtarld ?

andyexeter:
We're seeing the same issue as @soyuka. Looks like it will be fixed once #57617 is merged

Issue #52391, opened 2023-10-31
[Cache] `igbinary_serialize` is not compatible with `serialize` anymore


martijnc:
### Symfony version(s) affected

5.x, 6.x

### Description

As of PHP 7.4.6 (https://github.com/doctrine/common/issues/886#issuecomment-618263326), PHP's `serialize` function ignores uninitialized properties and no longer triggers an error. The same change was not (yet) made in igbinary's `igbinary_serialize`, there is a [ticket](https://github.com/igbinary/igbinary/issues/273) to change this though.

Symfony Cache `DefaultMarshaller` still treats `igbinary_serialize` as a drop-in replacement for `serialize` but this is no longer true. Trying to serialize entities with igbinary installed now results in the same error we used to get with `serialize`.

```
Typed property Proxies_CG_\App\Entity\EntityName::$ must not be accessed before initialization (in __sleep)
```

### How to reproduce

Serialize an object with an uninitialised property through Symfony cache with the default marshaller config.

### Possible Solution

It is possible to disable igbinary through the `cache.default_marshaller` service but Symfony may want to consider disabling it by default as it no longer is a drop-in replacement.

nicolas-grekas:
Isn't the solution to provide a fix to https://github.com/igbinary/igbinary/issues/273?
A friendly ping there should help.
Meanwhile, we could skip using igbinary by default indeed when saving objects.
Can you please send a PR doing so?

PegaseMarket:
Hello,
I have the same problem, but my php version is 7.4.33, igbinary is 3.2.7 and Symfony 5.4.38.
I don't understand why I always get this same error

PR #52700, opened 2023-11-23
[Cache] Fix return type of `Redis6Proxy::mget()`

### LINKED ISSUE #52668: [Cache] \Redis::mget() can return bool, Redis6Proxy::mget() doesn't account for that

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52668
| License       | MIT

After @dkarlovi's [comment](https://github.com/symfony/symfony/pull/52690#issuecomment-1823544628), here is another try that will not break returns types. Thank you for taking the time to explain where the exact problem is, Dalibor :+1: 

_Todo: having a look at failing tests_

### DIFF
--- a/src/Symfony/Component/Cache/Traits/Redis6Proxy.php
+++ b/src/Symfony/Component/Cache/Traits/Redis6Proxy.php
@@ -653,7 +653,14 @@ class Redis6Proxy extends \Redis implements ResetInterface, LazyObjectInterface
 
     public function mget($keys): \Redis|array
     {
-        return ($this->lazyObjectState->realInstance ??= ($this->lazyObjectState->initializer)())->mget(...\func_get_args());
+        /**
+         * Handle the special case where `mget()` returns false, waiting for the stub to be fixed.
+         *
+         * @see https://github.com/phpredis/phpredis/issues/1810
+         */
+        $v = ($this->lazyObjectState->realInstance ??= ($this->lazyObjectState->initializer)())->mget(...\func_get_args());
+
+        return $v === false ? [] : $v;
     }
 
     public function migrate($host, $port, $key, $dstdb, $timeout, $copy = false, $replace = false, #[\SensitiveParameter] $credentials = null): \Redis|bool


nicolas-grekas:
Did yo confirm the extension can return false?
Its signature is not allowing false so it shouldn't be possible.
To me this should be fixed upstream. Not here.
If upstream is not production ready, don't use it.

alexandre-daubois:
Yes, I can confirm the extension can return `false`:

image

_Tested on PHP 8.2, Redis Extension 6.0.2_

Shouldn't we provide a fix of this kind to comply to what the extension actually does? Or are you :-1: on this one because of the potential maintenance cost maybe?

(this case makes me think of [this comment](https://github.com/symfony/symfony/issues/52391#issuecomment-1795714371))

nicolas-grekas:
I'd like to see the pressure on this issue go from this repo to the source one yes. Sending fixes, etc ;)

dkarlovi:
> Its signature is not allowing false so it shouldn't be possible.

PHP extensions don't need to follow their own signatures, confirmed in Slack with @alcaeus

alcaeus:
> > Its signature is not allowing false so it shouldn't be possible.
> 
> PHP extensions don't need to follow their own signatures, confirmed in Slack with @alcaeus

I'll add a clarification to that: the type of a value returned in an extension method is only checked when PHP has been compiled with debug symbols. When not running in debug mode, extension developers are trusted to "not do stupid things" and return only values that conform to the method signature.

I absolutely agree that this is an upstream issue that needs fixing.

alexandre-daubois:
I am currently trying to work on a fix on the extension. Will link this in the issue, closing for now :+1: Thank you for all your answers!

alexandre-daubois:
For the record: https://github.com/phpredis/phpredis/pull/2422

PR #52702, opened 2023-11-23
[AssetMapper] Fix eager imports are not deduplicated

smnandre:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes ?
| New feature?  | no
| Deprecations? | no
| Issues        | Fix 
| License       | MIT

While recursively collecting eager imports, there may be duplicates. 

```
app.js
L  A_1.js
    L B.js 
L  A_2.js
    L B.js     
```

Removing them early avoid a lot of useless computation afterwards (and eventual side effects)

Exemple with the generated app.entrypoint.json of the Symfony UX website
- before: 78 entries
- after: 53 entries

### DIFF
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapGenerator.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapGenerator.php
@@ -217,25 +217,36 @@ class ImportMapGenerator
         return $this->assetMapper->getAssetFromSourcePath($this->importMapConfigReader->convertPathToFilesystemPath($path));
     }
 
+    /**
+     * Finds recursively all the non-lazy modules imported by an asset.
+     *
+     * @return array<string> The array of deduplicated import names
+     */
     private function findEagerImports(MappedAsset $asset): array
     {
         $dependencies = [];
-        foreach ($asset->getJavaScriptImports() as $javaScriptImport) {
-            if ($javaScriptImport->isLazy) {
-                continue;
-            }
+        $queue = [$asset];
 
-            $dependencies[] = $javaScriptImport->importName;
+        while ($asset = array_shift($queue)) {
+            foreach ($asset->getJavaScriptImports() as $javaScriptImport) {
+                if ($javaScriptImport->isLazy) {
+                    continue;
+                }
+                if (isset($dependencies[$javaScriptImport->importName])) {
+                    continue;
+                }
+                $dependencies[$javaScriptImport->importName] = true;
 
-            // Follow its imports!
-            if (!$nextAsset = $this->assetMapper->getAsset($javaScriptImport->assetLogicalPath)) {
-                // should not happen at this point, unless something added a bogus JavaScriptImport to this asset
-                throw new LogicException(sprintf('Cannot find imported JavaScript asset "%s" in asset mapper.', $javaScriptImport->assetLogicalPath));
+                // Follow its imports!
+                if (!$javaScriptAsset = $this->assetMapper->getAsset($javaScriptImport->assetLogicalPath)) {
+                    // should not happen at this point, unless something added a bogus JavaScriptImport to this asset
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/AssetMapper/ImportMap/ImportMapGenerator.php
```diff
@@ -217,6 +217,13 @@ private function findAsset(string $path): ?MappedAsset
         return $this->assetMapper->getAssetFromSourcePath($this->importMapConfigReader->convertPathToFilesystemPath($path));
     }
 
+    /**
+     * Finds recursively all the non-lazy modules imported by an asset.
+     *
+     * @internal
```
OskarStark: As the method is private, we don't need to make it internal

smnandre: Thanks, removed !

src/Symfony/Component/AssetMapper/ImportMap/ImportMapGenerator.php
```diff
@@ -225,17 +230,20 @@ private function findEagerImports(MappedAsset $asset): array
...
             }
-            $dependencies = array_merge($dependencies, $this->findEagerImports($nextAsset));
+
+            if ($nextImports = $this->findEagerImports($nextAsset)) {
+                $dependencies += array_combine($nextImports, $nextImports);
+            }
         }
 
-        return $dependencies;
+        return array_values($dependencies);
```
weaverryan:
```suggestion
        return array_unique($dependencies);
```

And removing the other changes?

smnandre: You're right... i often try to find the "smartest" way of doing this kind of things.... but it often end with a harder code lisibility and... no gain at all 

(🦦 i'm sure i won 0.001s in the sorting of my 200 items)

Hard habits :) Fixing asap

src/Symfony/Component/AssetMapper/ImportMap/ImportMapGenerator.php
```diff
@@ -232,10 +237,11 @@ private function findEagerImports(MappedAsset $asset): array
                 // should not happen at this point, unless something added a bogus JavaScriptImport to this asset
                 throw new LogicException(sprintf('Cannot find imported JavaScript asset "%s" in asset mapper.', $javaScriptImport->assetLogicalPath));
             }
+
             $dependencies = array_merge($dependencies, $this->findEagerImports($nextAsset));
```
norkunas: I wouldn't use array_merge in the loop https://github.com/kalessil/phpinspectionsea/blob/master/docs/performance.md#slow-array-function-used-in-loop

weaverryan: If this is something we try to avoid in other parts of Symfony, then let's do it (but I'm not sure it is). In this case, this code is called in dev but not in prod (the result is cached in prod). So it's not SUCH an important spot for performance.

weaverryan (approved):

src/Symfony/Component/AssetMapper/ImportMap/ImportMapGenerator.php
norkunas: Sometimes with each new SF version dev env feels slower and slower..

smnandre: @norkunas do you suggest a foreach.. or to just use the array_unique at the end ?

smnandre: Rewrote the traversal .. low memory + better handling of duplicates

fabpot (approved):

fabpot:
Thank you @smnandre.

PR #52704, opened 2023-11-23
[FrameworkBundle] Update docblock AbstractController

fmata:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | n/a
| License       | MIT

Small doc update to mention invokable controller usage.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php
@@ -145,7 +145,7 @@ abstract class AbstractController implements ServiceSubscriberInterface
     /**
      * Forwards the request to another controller.
      *
-     * @param string $controller The controller name (a string like Bundle\BlogBundle\Controller\PostController::indexAction)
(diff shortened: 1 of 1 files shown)


src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php
```diff
@@ -145,7 +145,7 @@ protected function generateUrl(string $route, array $parameters = [], int $refer
     /**
      * Forwards the request to another controller.
      *
-     * @param string $controller The controller name (a string like Bundle\BlogBundle\Controller\PostController::indexAction)
+     * @param string $controller The controller name (a string like Bundle\BlogBundle\Controller\PostController::indexAction or  Bundle\BlogBundle\Controller\PostController::class if it is invokable)
```
94noni: would it be better to add it in [doc page](https://symfony.com/doc/current/controller/forwarding.html) instead?

fmata: IMHO it's a good idea but in addition to the comment

nicolas-grekas:
Thank you @fmata.

PR #52707, opened 2023-11-23
[HttpKernel] Fix logging deprecations to the "php" channel when channel "deprecation" is not defined

### LINKED ISSUE #52672: Deprecated warnings sent to another channel after minor upgrade (v5.4.29 -> v5.4.31)

v5.4.31

### Description

I upgrade symfony from v5.4.29 to v5.4.31 and got huge amount of deprecated warning in my monolog logger in php channel.

My research showed me, that previously all deprecated errors were sent to security channel, and after upgrade it started to send to php channel.

Can you please explain, why symfony team decide to change default channel for deprecate error (from security to php) and how is it possible to return it back to security?

Example of deprecate errors:

`User Deprecated: Since symfony/security-guard 5.3: The "Symfony\Component\Security\Guard\GuardAuthenticatorHandler" class is deprecated, use the new authenticator system instead.`

`User Deprecated: Since symfony/security-core 5.3: The "Symfony\Component\Security\Core\Authentication\Provider\AuthenticationProviderInterface" interface is deprecated, use the new authenticator system instead.`

__

Thank you for your time.

p.s. i checked comparison [https://github.com/symfony/symfony/compare/v5.4.29...v5.4.31](https://github.com/symfony/symfony/compare/v5.4.29...v5.4.31) and cant see anything related to change of the channel

### How to reproduce

Use version of symfony/symfony - v5.4.31
Use some deprecated component from Security Core.

### Possible Solution

_No response_

### Additional Context

_No response_

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52672
| License       | MIT

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/debug_prod.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/debug_prod.php
@@ -26,7 +26,7 @@ return static function (ContainerConfigurator $container) {
                 param('debug.error_handler.throw_at'),
                 param('kernel.debug'),
                 param('kernel.debug'),
-                service('logger')->nullOnInvalid(),
+                null, // Deprecation logger if different from the one above
             ])
             ->tag('kernel.event_subscriber')
             ->tag('monolog.logger', ['channel' => 'php'])


stof:
This will not log deprecations at all when not using MonologBundle and also won't log them at all when using MonologBundle without explicitly creating a `deprecation` channel.

This re-introduces the same issue than what was fixed previously. The "bug" you tried to fix is that someone was relying on the previous regression as a feature rather than a bug.

So -1 for me.

nicolas-grekas:
To me this will log deprecations on the php channel, as wanted

nicolas-grekas:
I think you missed that deprecations will go on the logger defined as $logger argument when $deprecationLogger is null.

stof:
If this is indeed the case, why did we have to replace the usage of `service('monolog.logger.deprecation')->nullOnInvalid()` by a dynamic configuration in the compiler pass in the previous PR ?

nicolas-grekas:
We did that to wire "logger" by default so that when monolog is not there, we still have a logger for the php notices/deprecations.

stof:
but you now wire `null` by default, not `logger`.

nicolas-grekas:
Sure, that's the code in DebugHandlersListener that does the trick:

https://github.com/symfony/symfony/blob/11990909dfaade1b86fcb5d22a71ed963b9be4cb/src/Symfony/Component/HttpKernel/EventListener/DebugHandlersListener.php#L169-L177

PR #52712, opened 2023-11-24
[AssetMapper] Exclude dot files

### LINKED ISSUE #52697: [AssetMapper] Dotfiles exposed due to cache prefix

6.4,7.0,7.1

### Description

Any hidden / dotfile in the assets directory is deployed and publically accessible.

Assets files are versionned via a prefix inserted before the file extension.

The servers standard protection against dotfile access cannot work, as the mapped files are not "dotfiles"  anymore.

```
your-project/
├─ assets/
│  ├─ .DS_Store 
│  ├─ .cert
│  └─ ... 
```

```
your-project/
├─ public/
│  ├─ assets/
│  │   ├─ -XXXXX.DS_Store
│  │  └─ -YYYYY.cert
```

### How to reproduce

```
symfony new asset-mapper-dot-files --webapp --version="next" \
	&& cd asset-mapper-dot-files \
	&& composer req symfony/asset-mapper \
	&& touch assets/.FOO \
	&& APP_ENV=prod php bin/console asset-map:compile 
	
ls -la public/assets	
```

#### Exclude patterns

This behaviour can be prevented with a specific config directive 

```conf
# config/asset_mapper.yaml
framework:
    asset_mapper:
        paths:
            - assets/

        excluded_patterns:
            - '*/.*'
```            

As of today this is neither documented or "default behaviour".

That could be a first move, but this is something risky, as the pattern is not self-explanatory 

### Fix asset collection

Another way would be to modify the assets collection (in the Repository) to ignore dotfiles

..

What do you think ?

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes - could possibly be considered a security fix 
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #52697
| License       | MIT

See #52697. The biggest question is: should we do this? Is it enough to say "Hey! When you map an assets directory, EVERYTHING is published publicly?". Or should we be on the safe side and exclude dot files by default.

Cheers!

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
@@ -899,6 +899,11 @@ class Configuration implements ConfigurationInterface
                             ->prototype('scalar')->end()
                             ->example(['*/assets/build/*', '*/*_.scss'])
                         ->end()
+                        // boolean called  defaulting to true
+                        ->booleanNode('exclude_dotfiles')
+                            ->info('If true, any files starting with "." will be excluded from the asset mapper')
+                            ->defaultTrue()
+                        ->end()
                         ->booleanNode('server')
                             ->info('If true, a "dev server" will return the assets from the public directory (true in "debug" mode only by default)')
                             ->defaultValue($this->debug)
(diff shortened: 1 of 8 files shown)


smnandre:
🥇 🧞‍♂️ 🙇‍♂️ 

You're amazing, thank you!

94noni (approved):
🚀

smnandre:
For me it should 100% be the default behaviour, as the files do change their "nature" (losing the "dotfile" type on the way).

If it's not enabled per default, i'd like a commented (+explanation) setting in the configuration / recipe :) 

Thanks again for the quick reaction @weaverryan

fabpot (approved):

fabpot:
Thank you @weaverryan.

PR #52719, opened 2023-11-24
[Mime] Add `TemplatedEmail::$locale` to the serialized props

### LINKED ISSUE #52718: [Mime]TemplatedEmail::$locale is not serialized

6.4

### Description

In https://github.com/symfony/symfony/pull/51690 the locale-property was added to TemplatedEmail, but this property is not serialized.

If Mails are sent asynchronously via the Messenger-Component, the locale-value is not available, once the TemplatedEmail is rendered. 

### How to reproduce

Create and sent a TemplatedEmail with a locale and send it via a message queue. The locale-property is not serialized to the message-queue

### Possible Solution

_No response_

### Additional Context

_No response_

mkrauser:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | yes
| New feature?  |no
| Deprecations? | no 
| Issues        | Fix #52718
| License       | MIT

To use the locale-property when working with Symfony Messenger asynchronously it has to be serialized.

### DIFF
--- a/src/Symfony/Bridge/Twig/Mime/TemplatedEmail.php
+++ b/src/Symfony/Bridge/Twig/Mime/TemplatedEmail.php
@@ -100,7 +100,7 @@ class TemplatedEmail extends Email
      */
     public function __serialize(): array
     {
-        return [$this->htmlTemplate, $this->textTemplate, $this->context, parent::__serialize()];
+        return [$this->htmlTemplate, $this->textTemplate, $this->context, parent::__serialize(),  $this->locale];
     }
 
     /**
@@ -109,6 +109,7 @@ class TemplatedEmail extends Email
     public function __unserialize(array $data): void
     {
         [$this->htmlTemplate, $this->textTemplate, $this->context, $parentData] = $data;
+        $this->locale = $data[4] ?? null;
 
         parent::__unserialize($parentData);
     }
--- a/src/Symfony/Bridge/Twig/Tests/Mime/TemplatedEmailTest.php
+++ b/src/Symfony/Bridge/Twig/Tests/Mime/TemplatedEmailTest.php
@@ -43,12 +43,14 @@ class TemplatedEmailTest extends TestCase
             ->textTemplate('text.txt.twig')
             ->htmlTemplate('text.html.twig')
             ->context($context = ['a' => 'b'])
+            ->locale($locale = 'fr_FR')
         ;
 
         $email = unserialize(serialize($email));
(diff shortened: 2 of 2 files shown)


src/Symfony/Bridge/Twig/Mime/TemplatedEmail.php
```diff
@@ -100,15 +100,15 @@ public function markAsRendered(): void
      */
     public function __serialize(): array
     {
-        return [$this->htmlTemplate, $this->textTemplate, $this->context, parent::__serialize()];
+        return [$this->htmlTemplate, $this->textTemplate, $this->context, $this->locale, parent::__serialize()];
```
nicolas-grekas:
```suggestion
        return [$this->htmlTemplate, $this->textTemplate, $this->context, parent::__serialize(), $this->locale];
```

src/Symfony/Bridge/Twig/Mime/TemplatedEmail.php
```diff
@@ -100,15 +100,15 @@ public function markAsRendered(): void
...
     }
 
     /**
      * @internal
      */
     public function __unserialize(array $data): void
     {
-        [$this->htmlTemplate, $this->textTemplate, $this->context, $parentData] = $data;
+        [$this->htmlTemplate, $this->textTemplate, $this->context, $this->locale, $parentData] = $data;
```
nicolas-grekas:
```suggestion
        [$this->htmlTemplate, $this->textTemplate, $this->context, $this->locale, $parentData] = $data;
        $this->locale = $data[5] ?? null;
`````

**Review by nicolas-grekas — COMMENTED** [review]
Here are some suggestions that should help provide some compat with previously serialized payloads

File: src/Symfony/Bridge/Twig/Mime/TemplatedEmail.php (hunk 2 above)
Comment (mkrauser) [reply src/Symfony/Bridge/Twig/Mime/TemplatedEmail.php]: Thank you, but the 5 should be a 4, right?
And is the assignment in the line before still necessary then?

```suggestion
        [$this->htmlTemplate, $this->textTemplate, $this->context, $parentData] = $data;
        $this->locale = $data[4] ?? null;
```

nicolas-grekas: correct it should be 4

OskarStark (approved):

fabpot (approved):

fabpot:
Thank you @mkrauser.

PR #52720, opened 2023-11-24
[Cache] Remove database server version detection

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | N/A
| License       | MIT

Our DoctrineDbalAdapter is currently sniffing for the database server version to implement workarounds for very old database servers:

* SQL Server before 2008: DBAL 3 does not support older SQL server releases than this one.
* Postgres before 9.5: Postgres 9.5 has been released in Jan 2016, it predecessor 9.4 is EOL since Feb 2020.

I think it's fair to not support those old releases anymore at this point.

### DIFF
--- a/UPGRADE-7.0.md
+++ b/UPGRADE-7.0.md
@@ -54,6 +54,7 @@ Cache
 -----
 
  * Add parameter `\Closure $isSameDatabase` to `DoctrineDbalAdapter::configureSchema()`
+ * Drop support for Postgres < 9.5 and SQL Server < 2008 in `DoctrineDbalAdapter`
 
 Config
 ------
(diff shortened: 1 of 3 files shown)


GromNaN:
What will be the behavior if someone uses an old version of Postgres?

This should be mentioned in the changelog and the minimum database server versions needs to be documented.

derrabus:
> What will be the behavior if someone uses an old version of Postgres?

The adapter will use an SQL syntax the server won't understand. That's a hard exception.

> This should be mentioned in the changelog and the minimum database server versions needs to be documented.

I'll update the change log and open a PR for the docs.

nicolas-grekas (approved):

derrabus:
I've opened symfony/symfony-docs#19187 for the docs.

xabbuh (approved):

GromNaN (approved):

PR #52722, opened 2023-11-24
[Validator] Fix type error for non-array items when `Unique::fields` is set

aprat84:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

The `UniqueValidator` crashes if:
- `fields` argument is set
- `value` is an array but some of the elements are not

Example value:
```php
$value = [
    [
        "field": 1
    ],
    "abc",
];
```

Example constraint config:
```php
new Unique(fields: ['field']);
```

Error thrown:
```text
Symfony\Component\Validator\Constraints\UniqueValidator::reduceElementKeys(): Argument #2 ($element) must be of type array, string given, called in /var/www/html/vendor/symfony/validator/Constraints/UniqueValidator.php on line 46
```

Solution:

Looking at the actual behavior, if one of its items, being an array, has none of the `fields`, is ignored.

So, continuing on this premise, if one of its items, is not an array, it should be ignored too.

That's what this PR is about (hope so)!

### DIFF
--- a/src/Symfony/Component/Validator/Constraints/UniqueValidator.php
+++ b/src/Symfony/Component/Validator/Constraints/UniqueValidator.php
@@ -45,7 +45,7 @@ class UniqueValidator extends ConstraintValidator
         foreach ($value as $element) {
             $element = $normalizer($element);
 
-            if ($fields && !$element = $this->reduceElementKeys($fields, $element)) {
+            if ($fields && !(\is_array($element) && $element = $this->reduceElementKeys($fields, $element))) {
                 continue;
             }
 
--- a/src/Symfony/Component/Validator/Tests/Constraints/UniqueValidatorTest.php
+++ b/src/Symfony/Component/Validator/Tests/Constraints/UniqueValidatorTest.php
@@ -244,6 +244,30 @@ class UniqueValidatorTest extends ConstraintValidatorTestCase
     }
 
     /**
+     * @dataProvider getValidCollectionValues
+     */
+    public function testValidCollectionValues(array $value, array $fields)
+    {
+        $this->validator->validate($value, new Unique(fields: $fields));
+
+        $this->assertNoViolation();
+    }
+
+    public static function getValidCollectionValues(): array
+    {
+        return [
+            'unique empty item' => [
+                [['field' => 1], ['field' => 2], []],
+                ['field'],
+            ],
+            'unique non-array item' => [
+                [['field' => 1], ['field' => 2], '', 1, null],
+                ['field'],
+            ],
+        ];
+    }
+
+    /**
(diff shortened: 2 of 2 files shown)


src/Symfony/Component/Validator/Tests/Constraints/UniqueValidatorTest.php
```diff
@@ -242,6 +242,30 @@ public static function getInvalidFieldNames(): array
...
+     * @dataProvider getValidCollectionValues
+     */
+    public function testValidCollectionValues(array $value, array $fields)
+    {
+        $this->validator->validate($value, new Unique(fields: $fields));
+
+        $this->assertNoViolation();
+    }
+
+    public function getValidCollectionValues(): array
```
nicolas-grekas:
```suggestion
    public static function getValidCollectionValues(): array
```

nicolas-grekas (commented):
Can you please rebase to get rid of the first commit?

src/Symfony/Component/Validator/Constraints/UniqueValidator.php
```diff
@@ -43,7 +43,7 @@ public function validate(mixed $value, Constraint $constraint)
         $collectionElements = [];
         $normalizer = $this->getNormalizer($constraint);
         foreach ($value as $element) {
-            if ($fields && !$element = $this->reduceElementKeys($fields, $element)) {
+            if ($fields && !(\is_array($element) && $element = $this->reduceElementKeys($fields, $element))) {
```
xabbuh, line 48: I wonder if we shouldn't rather throw an `UnexpectedValueException` instead.

aprat84: Good point!

Checking types is not this constraint validator responsibility, as it doesn't need a certain type to do its validation.

When `fields` is passed, and an item doesn't have them, even if it's an array, it is ignored. Why not do the same when the item is not an array?

Every item in the following collection is unique, event when passing `fields: 'a'` to the constraint:
```php
[
  ['a' => 1, 'b' => 1],
  ['a' => 2, 'b' => 2],
  ['b' => 3],
  null,
]
```

Btw, I'll be happy with any solution, as long as it doesn't crash.

---

On a wider thought, I think the `UnexpectedValueException` should not exist, as we already have the `Type` constraint.
A value that can not be validated by a constraint should be ignored (ie: `EmailValidator` should skip anything that's not a string)
But this is another topic, and probably I lack the knowledge to suggest so 😅

nicolas-grekas:
Thank you @aprat84.

PR #52725, opened 2023-11-24
[AssetMapper] Fix: also download files referenced by url() in CSS

### LINKED ISSUE #52620: [AssetMapper] bootstrap icons works with direct link not with assetmapper require

6.4

### Description

```html
<link rel="stylesheet"
          href="https://cdn.jsdelivr.net/npm/bootstrap-icons@1.11.1/font/bootstrap-icons.min.css">
```

works as expected, but 

```bash
bin/console importmap:require bootstrap-icons/font/bootstrap-icons.min.css
```

doesn't.

See https://bootstrap-icons-bug.survos.com/ for a working and non-working version.

### How to reproduce

```bash
symfony new bootstrap-icons-bug --webapp --version=next --php=8.2 && cd bootstrap-icons-bug
composer config minimum-stability beta
composer config extra.symfony.allow-contrib true
composer req symfony/asset-mapper:^6.4 
bin/console importmap:require bootstrap-icons/font/bootstrap-icons.min.css
echo "import 'bootstrap-icons/font/bootstrap-icons.min.css'" >> assets/app.js


bin/console make:controller App -i
sed -i "s|/app|/|" src/Controller/AppController.php 

cat > templates/app.html.twig <<'END'
{% extends 'base.html.twig' %}
{% set useAssetMap = app.request.get('useAssetMap', false) %}
{% block javascripts %}
    {{ useAssetMap ? parent() }}
{% endblock %}

{% block stylesheets %}
    {% if not useAssetMap %} <link rel="stylesheet"
          href="https://cdn.jsdelivr.net/npm/bootstrap-icons@1.11.1/font/bootstrap-icons.min.css">
    {% endif %}
{% endblock %}

{% block body %}
    {% for useAssetMap in [true, false] %}
    <p>
        <a href="{{ path('app_app', {useAssetMap:useAssetMap}) }}">{{ useAssetMap ? 'with downloaded assetmap file' : 'with link tag to stylesheet on jsdelivr' }} </a>
        </p>
    {% endfor %}
    
    icon: <span class="bi bi-github"></span>
{% endblock %}

END

symfony server:start -d
symfony open:local
```

### Possible Solution

There's something amiss with parsing out the fonts, the javascript console gives some warnings/errors.

### Additional Context

_No response_

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52620
| License       | MIT

Hi!

@tacman found another situation where our "asset downloader" wasn't complete.

If we're downloading a CSS file, it may reference other files via `url()`. In #52620, these were font files, but they could be images or even other CSS Files. Currently, we do NOT download these, so the local references fail. This fixes that. I tried to keep the PR as small as possible, given the late stage of 6.4. But this *is* a bug fix: the downloader currently doesn't work for CSS files with `url()` inside.

Tested locally on ux.symfony.com with the bootstrap-icons package.

Cheers!

### DIFF
--- a/src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
@@ -11,10 +11,12 @@
 
 namespace Symfony\Component\AssetMapper\ImportMap\Resolver;
 
+use Symfony\Component\AssetMapper\Compiler\CssAssetUrlCompiler;
 use Symfony\Component\AssetMapper\Exception\RuntimeException;
 use Symfony\Component\AssetMapper\ImportMap\ImportMapEntry;
 use Symfony\Component\AssetMapper\ImportMap\ImportMapType;
 use Symfony\Component\AssetMapper\ImportMap\PackageRequireOptions;
+use Symfony\Component\Filesystem\Path;
 use Symfony\Component\HttpClient\HttpClient;
 use Symfony\Contracts\HttpClient\Exception\HttpExceptionInterface;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
@@ -157,12 +159,11 @@ final class JsDelivrEsmResolver implements PackageResolverInterface
     /**
      * @param ImportMapEntry[] $importMapEntries
      *
-     * @return array<string, array{content: string, dependencies: string[]}>
+     * @return array<string, array{content: string, dependencies: string[], extraFiles: array<string, string>}>
      */
     public function downloadPackages(array $importMapEntries, callable $progressCallback = null): array
     {
         $responses = [];
-
         foreach ($importMapEntries as $package => $entry) {
             if (!$entry->isRemotePackage()) {
                 throw new \InvalidArgumentException(sprintf('The entry "%s" is not a remote package.', $entry->importName));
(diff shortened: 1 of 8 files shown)


src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -205,6 +218,47 @@ public function downloadPackages(array $importMapEntries, callable $progressCall
...
+                    foreach ($extraFiles as $newExtraFile) {
+                        $extraFileResponses[$package][] = [$this->httpClient->request('GET', sprintf(self::URL_PATTERN_DIST_CSS, $packageName, $version, $newExtraFile)), $newExtraFile, $packageName, $version];
+                    }
+                }
+            }
+        }
+
+        if ($extraFileResponses) {
+            goto download_extra_files;
+        }
```
weaverryan, line 251: This got a bit complex because it needs to be recursive (`first.css` has a `url()` pointing to `second.css`, so then we need to check `second.css` also for `url()`). However, the recursion is tested, though I don't know of any examples "in the wild" to try this on.

I imagine vendor packages importing one css file from another would be quite rare because CSS files are always packaged into one to reduce requests (even with HTTP2 this is still important for CSS files, as they are page-render-blocking and you can't discover that you need to download `second.css` until you download `first.css`).

fabpot:
Thank you @weaverryan.

PR #51049, opened 2023-07-20
[Form] Relax assertions for ICU 72.1

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

> In many formatting patterns, ASCII spaces are replaced with Unicode spaces
> (e.g., a "thin space").

see https://icu.unicode.org/download/72

PR #52740, opened 2023-11-26
[Validator] Made tests forward-compatible with ICU 72.1

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | N/A
| License       | MIT

Starting with ICU 72.1 (CLDR 42), the INTL extension renders a so-called narrow non-breaking space into various localized date and time formats. Our CI does not catch that change yet because it runs an older version of the ICU library.

I'm running PHP on a Mac with Homebrew as package manager which already ships a newer ICU (73.2 currently). This causes a lot of validator tests to fail because they expect a space in time expressions like `12:00 AM` where a NNBSP character is rendered instead.

This change is an attempt to give me a green test suite while maintaining the ability to run against older ICU versions, such as the one our CI uses.


xabbuh (approved):

xabbuh:
We could update some tests in the Form component in a similar manner (relates to #51049).

PR #52748, opened 2023-11-26
[TwigBundle] Enable `#[AsTwigFilter]`, `#[AsTwigFunction]` and `#[AsTwigTest]` attributes to configure runtime extensions

### LINKED ISSUE #50016: [RFC][TwigBundle] Register custom functions & filters using an attribute
### Description

Currently, in order to create a custom twig function/filter, we need create a class that extends the AbstractExtension, that looks like this:

```php
class TwigExtension extends AbstractExtension
{
    public function getFilters(): array
    {
        return [
            new TwigFilter('my_custom_filter', 'myCustomFilter'),
        ];
    }

    public function getFunctions(): array
    {
        return [
            new TwigFunction('my_custom_function', 'myCustomFunction'),
        ];
    }

    public function myCustomFilter(): string
    {
        // code here
    }

    public function myCustomFunction(): string
    {
        // code here
    }
}
```

### Example

I've always thought that the DX regarding twig extension could be better when working with Symfony. Today, I imagined the following:
```php
class TwigExtension extends AbstractExtension
{
    #[TwigFilter('my_custom_filter')]
    public function myCustomFilter(): string
    {
        // code here
    }

    #[TwigFunction('my_custom_function')]
    public function myCustomFunction(): string
    {
        // code here
    }
}
```

If we don't want to modify or conflict with the existing `TwigFunction` / `TwigFilter` classes, we could name them like this and store them in the twig bridge I think?
- `#[AsTwigFilter]`
- `#[AsTwigFunction]`

**What do you think?**

~~(If the feedbacks are positive, I'll open a PR.)~~ Hum, this seems more difficult than I initially thought.

GromNaN:
| Q             | A
| ------------- | ---
| Branch?       | 7.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #50016
| License       | MIT

Integration for new PHP attributes introduced by https://github.com/twigphp/Twig/pull/3916.

### DIFF
--- a/src/Symfony/Bundle/TwigBundle/CHANGELOG.md
+++ b/src/Symfony/Bundle/TwigBundle/CHANGELOG.md
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+7.3
+---
+
+ * Enable `#[AsTwigFilter]`, `#[AsTwigFunction]` and `#[AsTwigTest]` attributes
+   to configure extensions on runtime classes
+
 7.1
 ---
 
--- a/src/Symfony/Bundle/TwigBundle/DependencyInjection/TwigExtension.php
+++ b/src/Symfony/Bundle/TwigBundle/DependencyInjection/TwigExtension.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Bundle\TwigBundle\DependencyInjection;
 
+use Symfony\Bundle\TwigBundle\DependencyInjection\Compiler\AttributeExtensionPass;
 use Symfony\Component\AssetMapper\AssetMapper;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\Config\Resource\FileExistenceResource;
@@ -25,6 +26,9 @@ use Symfony\Component\Mailer\Mailer;
 use Symfony\Component\Translation\LocaleSwitcher;
 use Symfony\Component\Translation\Translator;
 use Symfony\Contracts\Service\ResetInterface;
+use Twig\Attribute\AsTwigFilter;
+use Twig\Attribute\AsTwigFunction;
+use Twig\Attribute\AsTwigTest;
 use Twig\Environment;
 use Twig\Extension\ExtensionInterface;
 use Twig\Extension\RuntimeExtensionInterface;
@@ -178,6 +182,10 @@ class TwigExtension extends Extension
         $container->registerForAutoconfiguration(ExtensionInterface::class)->addTag('twig.extension');
         $container->registerForAutoconfiguration(LoaderInterface::class)->addTag('twig.loader');
         $container->registerForAutoconfiguration(RuntimeExtensionInterface::class)->addTag('twig.runtime');
+
+        $container->registerAttributeForAutoconfiguration(AsTwigFilter::class, AttributeExtensionPass::autoconfigureFromAttribute(...));
+        $container->registerAttributeForAutoconfiguration(AsTwigFunction::class, AttributeExtensionPass::autoconfigureFromAttribute(...));
+        $container->registerAttributeForAutoconfiguration(AsTwigTest::class, AttributeExtensionPass::autoconfigureFromAttribute(...));
 
         if (false === $config['cache']) {
             $container->removeDefinition('twig.template_cache_warmer');
--- a/src/Symfony/Bundle/TwigBundle/DependencyInjection/Compiler/AttributeExtensionPass.php
+++ b/src/Symfony/Bundle/TwigBundle/DependencyInjection/Compiler/AttributeExtensionPass.php
@@ -1,0 +1,52 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\TwigBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\ChildDefinition;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Twig\Attribute\AsTwigFilter;
+use Twig\Attribute\AsTwigFunction;
+use Twig\Attribute\AsTwigTest;
+use Twig\Extension\AttributeExtension;
+
+/**
+ * Register an instance of AttributeExtension for each service using the
+ * PHP attributes to declare Twig callables.
+ *
+ * @author Jérôme Tamarelle <jerome@tamarelle.net>
+ *
+ * @internal
+ */
+final class AttributeExtensionPass implements CompilerPassInterface
+{
+    private const TAG = 'twig.attribute_extension';
+
+    public static function autoconfigureFromAttribute(ChildDefinition $definition, AsTwigFilter|AsTwigFunction|AsTwigTest $attribute, \ReflectionMethod $reflector): void
+    {
+        $definition->addTag(self::TAG);
+
+        // The service must be tagged as a runtime to call non-static methods
+        if (!$reflector->isStatic()) {
+            $definition->addTag('twig.runtime');
+        }
+    }
+
+    public function process(ContainerBuilder $container): void
+    {
+        foreach ($container->findTaggedServiceIds(self::TAG, true) as $id => $tags) {
+            $container->register('.twig.extension.'.$id, AttributeExtension::class)
+                ->setArguments([$container->getDefinition($id)->getClass()])
+                ->addTag('twig.extension');
+        }
+    }
+}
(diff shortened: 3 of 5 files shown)


src/Symfony/Bundle/TwigBundle/DependencyInjection/TwigExtension.php
```diff
@@ -179,6 +180,11 @@ public function load(array $configs, ContainerBuilder $container): void
         if (false === $config['cache']) {
             $container->removeDefinition('twig.template_cache_warmer');
         }
+
+        // Attributes declaration requires Twig 3.9+
+        if (!class_exists(AttributeExtension::class)) {
```
GromNaN: This may require a `ClassExistenceResource`.

src/Symfony/Bundle/TwigBundle/DependencyInjection/Compiler/RuntimeLoaderPass.php
```diff
@@ -35,5 +35,10 @@ public function process(ContainerBuilder $container): void
         }
 
         $definition->replaceArgument(0, ServiceLocatorTagPass::register($container, $mapping));
+
+        // Scan all Twig runtimes for attributes
+        if ($container->hasDefinition('twig.extension.attributes')) {
+            $container->getDefinition('twig.extension.attributes')->replaceArgument(0, array_keys($mapping));
```
nicolas-grekas: `$mapping` contains too many classes isn't it? We don't need to look for attributes on classes that have a corresponding extension. Can we find a way to make the distinction? Otherwise, we're breaking the declarative native of attributes I think.

GromNaN: By using a distinct tag for classes with the attribute `#[AsTwigExtension]` we can be more selective.

nicolas-grekas: :rocket:

src/Symfony/Bundle/TwigBundle/DependencyInjection/Compiler/RuntimeLoaderPass.php
```diff
@@ -33,7 +33,17 @@ public function process(ContainerBuilder $container): void
             $def = $container->getDefinition($id);
             $mapping[$def->getClass()] = new Reference($id);
         }
+        $attributesMapping = [];
+        foreach ($container->findTaggedServiceIds('twig.attribute_extension', true) as $id => $attributes) {
+            $def = $container->getDefinition($id);
+            $attributesMapping[$def->getClass()] = new Reference($id);
```
nicolas-grekas: shouldn't we skip this registration if `$mapping[$def->getClass()]` is set?

GromNaN: I don't get your suggestion.

Classes with `RuntimeExtensionInterface` need to be injected into the runtime loader. Classes with `AsTwigExtension` need to be injected into the runtime loader, but also into the AttributeExtension. That's why there is 2 lists.

nicolas-grekas: but if a class is registered both using the regular way and using the attribute, we should skip registering it with the attribute

GromNaN: Ah, correct. I don't expect anyone to use both the attribute and the interface on the same class, but this is not incompatible.

The + operator will remove duplicates.

nicolas-grekas: I would still skip here because of L46

GromNaN: I've made a small modification to have the list of classes distinct from the class => service mapping. It ends to the same thing.

src/Symfony/Bundle/TwigBundle/DependencyInjection/Compiler/RuntimeLoaderPass.php
```diff
@@ -33,7 +33,18 @@ public function process(ContainerBuilder $container): void
             $def = $container->getDefinition($id);
             $mapping[$def->getClass()] = new Reference($id);
         }
+        $classesWithAttribute = [];
+        foreach ($container->findTaggedServiceIds('twig.attribute_extension', true) as $id => $attributes) {
+            $def = $container->getDefinition($id);
+            $mapping[$def->getClass()] = new Reference($id);
```
nicolas-grekas: This gives precedence to attributes. This is less flexible because one cannot remove an attribute from a class. ;)

GromNaN: `twig.attribute_extension` implies `twig.runtime` and cannot work without it. The conflict would be if there is 2 services with a different id, but the same class name.  One is a `twig.runtime`, the other a `twig.attribute_extension`.

Do you think the priority should be reversed like this?
```suggestion
            $mapping[$def->getClass()] ??= new Reference($id);
```

src/Symfony/Bundle/TwigBundle/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Enable `#[AsTwigFilter]`, `#[AsTwigFunction]` and `#[AsTwigTest]` method attributes
+   to configure extensions on classes with marked with `#[AsTwigExtension]`
```
GromNaN: To be changed, we can track attributes on methods directly.

src/Symfony/Bundle/TwigBundle/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+7.1
```
OskarStark, line 10: 7.3

src/Symfony/Bundle/TwigBundle/DependencyInjection/TwigExtension.php
```diff
@@ -15,16 +15,19 @@
...
 use Symfony\Component\Console\Application;
+use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader\PhpFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\Form\AbstractRendererEngine;
 use Symfony\Component\Form\Form;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 use Symfony\Component\Mailer\Mailer;
+use Symfony\Component\RemoteEvent\Attribute\AsRemoteEventConsumer;
```
OskarStark: ?

kbond:
> We use the existing empty interface Twig\Extension\RuntimeExtensionInterface to identify services that are registered as Twig...

This shouldn't be required?

GromNaN:
I need to update the PR to register attributes on methods directly for autoconfiguration. The `RuntimeExtensionInterface` interface and the `#[AsTwigExtension]` attribute I added at one point, are not needed.

GromNaN:
PR updated. I added tests for the autoconfiguration. The service is registered as Twig runtime only when there are non-static methods.

src/Symfony/Bundle/TwigBundle/DependencyInjection/TwigExtension.php
```diff
@@ -179,6 +183,10 @@ public function load(array $configs, ContainerBuilder $container): void
         $container->registerForAutoconfiguration(LoaderInterface::class)->addTag('twig.loader');
         $container->registerForAutoconfiguration(RuntimeExtensionInterface::class)->addTag('twig.runtime');
 
+        $container->registerAttributeForAutoconfiguration(AsTwigFilter::class, AttributeExtensionPass::autoconfigureFromAttribute(...));
+        $container->registerAttributeForAutoconfiguration(AsTwigFunction::class, AttributeExtensionPass::autoconfigureFromAttribute(...));
+        $container->registerAttributeForAutoconfiguration(AsTwigTest::class, AttributeExtensionPass::autoconfigureFromAttribute(...));
+
```
GromNaN, line 189: If the attribute class is not declared (Older Twig version), this will be skipped. https://github.com/symfony/symfony/blob/6a029ec1025824be960454ea130feabf57e3f15a/src/Symfony/Component/DependencyInjection/Compiler/AttributeAutoconfigurationPass.php#L58-L62

fabpot:
Twig counter-part has been merged now.

fabpot (approved):

fabpot:
Thank you @GromNaN.

PR #52749, opened 2023-11-26
[Serializer] Add discriminator map to debug commmand output

jschaedl:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | little addition
| Deprecations? | no 
| Issues        | - 
| License       | MIT

The discriminator mapping was missing in the `debug:serializer` output. 

In case a `#[DiscriminatorMap]` is configured the output will now look like this: 

https://github.com/symfony/symfony/pull/52749/files#diff-d5830f2c54bb89f9819ca3c7c357a2f8df9319cf4d82f71b0704e693a2bfe9ceR106-R109

### DIFF
--- a/src/Symfony/Component/Serializer/Command/DebugCommand.php
+++ b/src/Symfony/Component/Serializer/Command/DebugCommand.php
@@ -97,6 +97,9 @@ class DebugCommand extends Command
     {
         $data = [];
 
+        $mapping = $classMetadata->getClassDiscriminatorMapping();
+        $typeProperty = $mapping?->getTypeProperty();
+
         foreach ($classMetadata->getAttributesMetadata() as $attributeMetadata) {
             $data[$attributeMetadata->getName()] = [
                 'groups' => $attributeMetadata->getGroups(),
@@ -107,6 +110,10 @@ class DebugCommand extends Command
                 'normalizationContexts' => $attributeMetadata->getNormalizationContexts(),
                 'denormalizationContexts' => $attributeMetadata->getDenormalizationContexts(),
             ];
+
+            if ($mapping && $typeProperty === $attributeMetadata->getName()) {
+                $data[$attributeMetadata->getName()]['discriminatorMap'] = $mapping->getTypesMapping();
+            }
         }
 
         return $data;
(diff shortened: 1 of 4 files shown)


src/Symfony/Component/Serializer/Command/DebugCommand.php
```diff
@@ -107,6 +107,11 @@ private function getAttributesData(ClassMetadataInterface $classMetadata): array
                 'normalizationContexts' => $attributeMetadata->getNormalizationContexts(),
                 'denormalizationContexts' => $attributeMetadata->getDenormalizationContexts(),
             ];
+
+            $mapping = $classMetadata->getClassDiscriminatorMapping();
```
mtarld: This (and the type property) could be fetched outside the loop

src/Symfony/Component/Serializer/Tests/Command/DebugCommandTest.php
```diff
@@ -79,6 +80,41 @@ public function testOutputWithClassArgument()
         );
     }
 
+    public function testOutputWithDiscriminatorMapClass()
+    {
+        $command = new DebugCommand(new ClassMetadataFactory(new AttributeLoader()));
+
+        $tester = new CommandTester($command);
+        $tester->execute(['class' => DummyClass::class], ['decorated' => false]);
```
mtarld: You may can reuse `DummyMessageInterface` instead? If you prefer to create a class, I'd named it with a more meaningful name, such as `DummyWithDiscriminatorMap`

mtarld (changes requested):
Looks almost good to me!

src/Symfony/Component/Serializer/Command/DebugCommand.php
jschaedl: Thanks. Good point. Changed it.

src/Symfony/Component/Serializer/Tests/Command/DebugCommandTest.php
jschaedl: using the Interface doesn't work, but i changed the name to be more meaningful.

mtarld: Alright! Just by curiosity, why it didn't work?

mtarld (approved):

fabpot (approved):

fabpot:
Thank you @jschaedl.

PR #52750, opened 2023-11-26
Allow custom query string parameters in the login link

janklan:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | No
| New feature?  | Yes, signature change recorded in `src/Symfony/Component/Security/Http/CHANGELOG.md`
| Deprecations? | No
| Issues        | None
| License       | MIT

This change allows the user to generate a login link with additional attributes, not just `user`, `expiry` and `hash`. The key reason is to allow passing things to the success handler through the URL of the login link.

This effectively adds support for the `_target_path` mechanism, which was the missing capability triggering my work on this PR: I use the magic link as the only authentication method, and I want to redirect the user to the URL requested before being redirected to the login page. This works well for password-driven form-based authentication scenarios, but in case of a magic link, neither the HTTP referrer nor a hidden input field is usable.

Another option would be to store the URL in session, but that assumes a stateless firewall and the willingness to start a session just to save this value. Also, the login link could be valid and opened after the session expires, or from a different browser or even different machine. So session doesn't really work. In my opinion, including `&_target_path=/welcome` is much simpler way to do it. 

If this gets merged, I'm happy to update docs as well.

### DIFF
--- a/src/Symfony/Component/Security/Http/LoginLink/LoginLinkHandler.php
+++ b/src/Symfony/Component/Security/Http/LoginLink/LoginLinkHandler.php
@@ -110,10 +110,10 @@ final class LoginLinkHandler implements LoginLinkHandlerInterface
         $hashParametersList = $request->query->get('_hash_parameters');
         if (!empty($hashParametersList)) {
             $hashParameters = [
-                '_hash_parameters' => $hashParametersList
+                '_hash_parameters' => $hashParametersList,
             ];
 
-            foreach(explode(',', $hashParametersList) as $hashParameterName) {
+            foreach (explode(',', $hashParametersList) as $hashParameterName) {
                 if (!$request->query->has($hashParameterName)) {
                     throw new InvalidLoginLinkException(sprintf('Missing "%s" parameter.', $hashParameterName));
                 }
--- a/src/Symfony/Component/Security/Http/LoginLink/LoginLinkHandler.php
+++ b/src/Symfony/Component/Security/Http/LoginLink/LoginLinkHandler.php
@@ -52,13 +52,12 @@ final class LoginLinkHandler implements LoginLinkHandlerInterface
         $expires = time() + ($lifetime ?: $this->options['lifetime']);
         $expiresAt = new \DateTimeImmutable('@'.$expires);
 
-        $hash = $this->signatureHasher->computeSignatureHash($user, $expires, $parameters);
+        ksort($parameters);
 
-        $parameters += [
-            'user' => $user->getUserIdentifier(),
-            'expires' => $expires,
-            'hash' => $hash,
-        ];
+        unset($parameters['_hash_parameters']);
+        if (!empty($parameters)) {
+            $parameters['_hash_parameters'] = implode(',', array_keys($parameters));
+        }
 
         if ($request) {
             $currentRequestContext = $this->urlGenerator->getContext();
@@ -72,7 +71,12 @@ final class LoginLinkHandler implements LoginLinkHandlerInterface
         try {
             $url = $this->urlGenerator->generate(
                 $this->options['route_name'],
-                $parameters,
+                [
+                    ...$parameters,
+                    '_user' => $user->getUserIdentifier(),
+                    '_expires' => $expires,
+                    '_hash' => $this->signatureHasher->computeSignatureHash($user, $expires, $parameters),
+                ],
                 UrlGeneratorInterface::ABSOLUTE_URL
             );
         } finally {
@@ -86,17 +90,27 @@ final class LoginLinkHandler implements LoginLinkHandlerInterface
 
     public function consumeLoginLink(Request $request): UserInterface
     {
-        $userIdentifier = $request->get('user');
+        $userIdentifier = $request->get('_user');
 
-        if (!$hash = $request->get('hash')) {
-            throw new InvalidLoginLinkException('Missing "hash" parameter.');
+        if (!$hash = $request->get('_hash')) {
+            throw new InvalidLoginLinkException('Missing "_hash" parameter.');
         }
-        if (!$expires = $request->get('expires')) {
-            throw new InvalidLoginLinkException('Missing "expires" parameter.');
+        if (!$expires = $request->get('_expires')) {
+            throw new InvalidLoginLinkException('Missing "_expires" parameter.');
         }
 
-        $parameters = $request->query->all();
-        unset($parameters['hash'], $parameters['expires'], $parameters['user']);
+        $parameters = [];
+
+        if (!empty($hashParameters = $request->query->get('_hash_parameters'))) {
+            $parameters['_hash_parameters'] = $hashParameters;
+            foreach(explode(',', $hashParameters) as $hashParameter) {
+                if (!$request->query->has($hashParameter)) {
+                    throw new InvalidLoginLinkException(sprintf('Missing "%s" parameter.', $hashParameter));
+                }
+
+                $parameters[$hashParameter] = $request->query->get($hashParameter);
+            }
+        }
 
         try {
             $this->signatureHasher->acceptSignatureHash($userIdentifier, $expires, $hash, $parameters);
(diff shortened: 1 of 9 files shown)


janklan:
I don't think the failing tests are caused by my changes, are they?

weaverryan:
> I don't think the failing tests are caused by my changes, are they?

Probably not :). Things get in flux a bit sometimes near release time.

src/Symfony/Component/Security/Http/LoginLink/LoginLinkHandler.php
```diff
@@ -44,15 +44,20 @@ public function __construct(UrlGeneratorInterface $urlGenerator, UserProviderInt
...
+     */
+    public function createLoginLink(UserInterface $user, Request $request = null, int $lifetime = null, array $parameters = []): LoginLinkDetails
     {
         $expires = time() + ($lifetime ?: $this->options['lifetime']);
         $expiresAt = new \DateTimeImmutable('@'.$expires);
 
-        $parameters = [
+        $hash = $this->signatureHasher->computeSignatureHash($user, $expires, $parameters);
+
+        $parameters += [
```
smnandre: ⚠️  Warning! 

By using` += ` there, the priority is given to `$parameters`  (the first array) 

The potential shared keys won't be erased... so $user, $hash or $expires could be predefined by values in $parameters and not overwritten.

```php

function  createLoginLink(array $parameters) {
	$hash = '$this->signatureHasher->computeSignatureHash($user, $expires, $parameters)';
	$parameters += [
	    'user' => '$user->getUserIdentifier()',
	    'expires' => '$expires',
	    'hash' => $hash,
	];
	var_dump($parameters);
}

createLoginLink(['user' => 'not the one', 'hash' => 'a pre computed hash']);
```

```php
array(3) {
  ["user"]=>
  string(11) "not the one"
  ["hash"]=>
  string(19) "a pre computed hash"
  ["expires"]=>
  string(8) "$expires"
}
```

src/Symfony/Component/Security/Http/LoginLink/LoginLinkHandler.php
```diff
@@ -90,12 +95,15 @@ public function consumeLoginLink(Request $request): UserInterface
             throw new InvalidLoginLinkException('Missing "expires" parameter.');
         }
 
+        $parameters = $request->query->all();
```
smnandre: This is impossible without breaking many code live, because
* webmails can append query parameters
* ...entreprise proxies/network too


That could break authentification for every link out there that has been generated

smnandre:
> In my opinion, including `&_target_path=/welcome` is much simpler way to do it.

Does it not work currently ?  `your-generated-link&_target_path=/welcome` ? 


And more on the reason of the feature, if you can compute the parameters [_path=/welcome] when you generate the link... can you do it in the SuccessHandler once the user is authenticated ?

src/Symfony/Component/Security/Http/LoginLink/LoginLinkHandler.php
janklan: Good point, I'll swap for `array_merge`.

src/Symfony/Component/Security/Http/LoginLink/LoginLinkHandler.php
janklan: I've never seen any _new_ attributes added to the query string, but I assume it could happen. 

I'll update the link so that it contains the list of keys used to generate the signature, and will only extract those keys from that parameter bag, ignoring all others.

janklan:
> Does it not work currently? 

You can get the link and append whatever you want, but then the _whatever_ is not part of the signature, which I thought isn't a good idea.

The risk of abusing is low, but if a malicious MITM actor changes the link to redirect an admin to a "delete everything" URL that's not protected by CSRF or other means, then that's gonna end up in tears. I know this scenario is somewhat sci-fi, but it illustrates why I think all of the URL parameters that are somehow handled by Symfony (which specifically `_target_path` is) should be signed and trustworthy.

> can you do it in the SuccessHandler once the user is authenticated

The login link handler already verifies the signature [generated from user name, roles and expiry](https://github.com/symfony/symfony/pull/52750/files#diff-f1d2fe225cb0988ec2b1989df508c394b42616d7657c3993b230d608e130be85L128). If you wanted to allow additional verification in a SuccessHandler, you'd have to include two signatures - one for the three built-in values, another for the second check in SuccessHandler. 

If the generated link can sign all intended values, why not use it? As long as any extra parameters are[ ignored for the purpose of the signature verification](https://github.com/symfony/symfony/pull/52750#discussion_r1412694681) (good point, btw, thanks!), I think it's safe to ignore any extra attributes because the application probably doesn't rely on them anyway.

src/Symfony/Component/Security/Http/LoginLink/LoginLinkHandler.php
smnandre: https://mailchimp.com/resources/utm-medium-email/

https://emaillabs.io/en/cloud-smtp-integration-with-google-analytics/

https://www.campaignmonitor.com/blog/email-marketing/utm-codes-in-email/

;)

janklan: You'd send an e-mail with pre-validated authentication credential through a newsletter delivery service? Even more so, you'd ask that service to manipulate the links for you to add tracking of said emails via Google Analytics? 

I wouldn't.

Wherever I came by any kind of tracking, it was done by the service replacing all links in the email to make them point at a redirecting proxy/tracker, which then forwarded the request to the original URL.

In any case, I don't disagree. Accepting all query parameters isn't a good idea, period.

janklan:
I addressed @smnandre's feedback.

@weaverryan the underscores are a brave proposal, feel free to reject it. The idea was to use the prefix on any value that's generated by the magic link mechanism to (1) make them stand out for the user looking at the link, (2) reduce a chance the generated name will collide with custom parameter name, and (3) because Symfony is using the underscore as a prefix on internal values elsewhere (`_locale`, `_route`, ...)

smnandre:
Sadly you cannot change the previous query parameters, and that is what the failing test proves :/ 

```
/home/runner/work/symfony/symfony/src/Symfony/Bundle/SecurityBundle/Tests/Functional/LoginLinkAuthenticationTest.php:42
```

As some people will have generated links using "user", "hash" or "expires" in the query string, if you change for "_user", "_hash" or "_expires" all the previously links become invalid :/

janklan:
>  all the previously links become invalid 

Yeah I know. That's why I called it a brave change :)

I assumed it's not a problem as the links can simply be re-generated.

The main case this could be a headache would be a long-lived link that's hardcoded somewhere and used repeatedly. I don't think there is a way to tell if anyone even does that (hey, anyone reading this, do you do that?), but I'm convinced it should not be done in the first place. That's like leaving a back door. 

@weaverryan WDYT? If we had to keep old links valid, I could add a compatibility layer to accept the values from attributes without the underscore. All new links would be generated with the underscore, all old links would work.

janklan:
Come to think of it, there isn't a reason for _not_ adding that compatibility layer and throw a deprecation notice.

Anyone against changing the parameter names to add the underscore, if we manage to keep the old links working until 8.0 release?

pbowyer:
What's current thinking on this PR? I came to open an issue for adding this feature and found it, so I'm in favour.

My use-case is as @janklan said: to add extra parameters to redirect to different locations afterwards, and it's nice if they're part of the hash.

I'm also wondering if adding custom query string parameters is the first step to generalising the login link feature and removing the need for a repository behind https://github.com/SymfonyCasts/reset-password-bundle because what it does can be embedded into the URL. @weaverryan what do you think?

janklan:
I've been bogged by other work but plan to push this to be ready for review in the next few weeks.

When I left it there was one failing test, but everything was working otherwise well. 

I think the ability to add things to the signed URL doesn't remove any function, bud adds flexibility, so "why not"?

janklan:
Superseded by https://github.com/symfony/symfony/pull/64271

PR #52758, opened 2023-11-27
[Cache] Fix Couchbase password parsing

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

This is the followup for https://github.com/symfony/symfony/pull/52688#issuecomment-1825676774.

### DIFF
--- a/src/Symfony/Component/Cache/Adapter/CouchbaseCollectionAdapter.php
+++ b/src/Symfony/Component/Cache/Adapter/CouchbaseCollectionAdapter.php
@@ -59,10 +59,7 @@ class CouchbaseCollectionAdapter extends AbstractAdapter
 
         set_error_handler(static fn ($type, $msg, $file, $line) => throw new \ErrorException($msg, 0, $type, $file, $line));
 
-        $dsnPattern = '/^(?<protocol>couchbase(?:s)?)\:\/\/(?:(?<username>[^\:]+)\:(?<password>[^\@]{6,})@)?'
-            .'(?<host>[^\:]+(?:\:\d+)?)(?:\/(?<bucketName>[^\/\?]+))(?:(?:\/(?<scopeName>[^\/]+))'
-            .'(?:\/(?<collectionName>[^\/\?]+)))?(?:\/)?(?:\?(?<options>.*))?$/i';
-
+        $pathPattern = '/^(?:\/(?<bucketName>[^\/\?]+))(?:(?:\/(?<scopeName>[^\/]+))(?:\/(?<collectionName>[^\/\?]+)))?(?:\/)?$/';
         $newServers = [];
         $protocol = 'couchbase';
         try {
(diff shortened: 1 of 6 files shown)


src/Symfony/Component/Cache/Adapter/CouchbaseCollectionAdapter.php
```diff
@@ -85,7 +85,7 @@ public static function createConnection($dsn, array $options = [])
                 preg_match($dsnPattern, $server, $matches);
 
                 $username = $matches['username'] ?: $username;
-                $password = $matches['password'] ?: $password;
+                $password = rawurldecode($matches['password'] ?: $password);
```
nicolas-grekas: The issue in this class (and the other) is not only about the password - it's about all URL components.
The fix should involve using parse_url (as done in all other DSN parsers) to let PHP decode everything as needed.

alexandre-daubois: I updated the DSN parsing with `parse_url` 👍  I'll try to have a look at other adapters in a separate PR 👍

alexandre-daubois:
What about deprecating the Couchbase Bucket adapter in another PR? It is only compatible with Couchbase 2, which conflicts with this adapter compatible with Couchbase 3

alexandre-daubois:
After discussing with @nicolas-grekas, the following changes have been made:

- Replaced `$parsed` by `$params` to match other DSN parsing
- Target 7.1 instead of 5.4

Note: `rawurldecode()` is still needed even after `parse_url`, which doesn't decode HTML chars. I'll send another PR to deprecate the `CouchbaseBucketAdapter`, only compatible with Couchbase 2 (which conflicts with this adapter requiring Couchbase 3 and making the other one not possible to test on the CI).

alexandre-daubois:
Rebased and conflicts fixed. All green 🙂

src/Symfony/Component/Cache/Adapter/CouchbaseCollectionAdapter.php
```diff
@@ -74,31 +71,32 @@ public static function createConnection(#[\SensitiveParameter] array|string $dsn
                     throw new InvalidArgumentException('Invalid Couchbase DSN: it does not start with "couchbase:".');
                 }
 
-                preg_match($dsnPattern, $server, $matches);
+                $params = parse_url($server);
 
-                $username = $matches['username'] ?: $username;
-                $password = $matches['password'] ?: $password;
-                $protocol = $matches['protocol'] ?: $protocol;
+                $username = $params['user'] ?? $username;
```
nicolas-grekas: same for the password:
```suggestion
                $username = isset($params['user']) ? rawurldecode($params['user']) : $username;
```

nicolas-grekas (commented):
what a mess parse_url...
so we should now add rawurldecode on all user/pass in the codebase that use parse_url, isn't it?

alexandre-daubois:
It seems 😕 https://3v4l.org/eshPt

I'll have a look at it. Is this something that also should target 7.1? I'd say yes.

nicolas-grekas:
Thank you @alexandre-daubois.

Issue #52753, opened 2023-11-27
[VarExporter][FrameworkBundle][7.0] Error when using a  lazy `AbstractController`


keulinho:
### Symfony version(s) affected

7.0

### Description

For 7.0 the protected `$container` property was typed as  `ContainerInterface`, this leads to an error when using it together with the `LazyGhost` as it tries to access the property before it is initialized.

### How to reproduce

Extend the AbstractController, configure it in the container as lazy and with the `setContainer` method

### Possible Solution

Catch the exception, and set null as fallback:
Before:
```php
        $previous = $this->container ?? null;
```
After:
```php
        try {
            $previous = $this->container;
        } catch (\Error) {
            $previous = null;
        }
```

I tried to use isset instead of null coalesce, but that leads to memory issues, so i guess catching the exception would be best

### Additional Context

See stack trace:
```
Error: Cannot access uninitialized non-nullable property Symfony\Bundle\FrameworkBundle\Controller\AbstractController::$container by reference
/builds/shopware/6/product/platform/vendor/symfony/var-exporter/LazyGhostTrait.php:133
/builds/shopware/6/product/platform/vendor/symfony/framework-bundle/Controller/AbstractController.php:63
/builds/shopware/6/product/platform/var/cache/test_h8458ab63ca9647a85b0d42ef842db976/ContainerMg28vrx/Shopware_Core_KernelTestDebugContainer.php:2851
/builds/shopware/6/product/platform/var/cache/test_h8458ab63ca9647a85b0d42ef842db976/ContainerMg28vrx/Shopware_Core_KernelTestDebugContainer.php:2846
/builds/shopware/6/product/platform/vendor/symfony/var-exporter/Internal/LazyObjectState.php:61
/builds/shopware/6/product/platform/vendor/symfony/var-exporter/LazyGhostTrait.php:206
/builds/shopware/6/product/platform/vendor/symfony/framework-bundle/Controller/AbstractController.php:63
/builds/shopware/6/product/platform/vendor/symfony/framework-bundle/Controller/ControllerResolver.php:28
/builds/shopware/6/product/platform/vendor/symfony/http-kernel/Controller/ControllerResolver.php:66
```

nicolas-grekas:
Can you please provide a small reproducer?
This might be caused by https://github.com/php/php-src/issues/12695
Is your controller lazy? That shouldn't be needed I think (but that doesn't invalidate the issue)

nicolas-grekas:
[...]

PR #52762, opened 2023-11-27
[VarExporter] Work around php/php-src#12695 for lazy objects, fixing nullsafe-related behavior

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | #52753
| License       | MIT

This PR fixes another issue discovered with @arnaud-lb while working on lazy-objects.
The root of the issue is described in php/php-src#12695 and this PR implements a fix based on a suggestion by @iluuu1994 

Only ghost objects are affected; proxy objects are OK.

### DIFF
--- a/src/Symfony/Component/VarExporter/Tests/LazyGhostTraitTest.php
+++ b/src/Symfony/Component/VarExporter/Tests/LazyGhostTraitTest.php
@@ -65,8 +65,10 @@ class LazyGhostTraitTest extends TestCase
 
         $this->assertSame(["\0".TestClass::class."\0lazyObjectState"], array_keys((array) $instance));
         unset($instance->public);
-        $this->assertFalse(isset($instance->public));
         $this->assertSame(4, $instance->publicReadonly);
+        $this->expectException(\BadMethodCallException::class);
+        $this->expectExceptionMessage('__isset(public)');
+        isset($instance->public);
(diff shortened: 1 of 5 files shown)


src/Symfony/Component/VarExporter/Tests/LazyGhostTraitTest.php
```diff
@@ -65,8 +65,10 @@ public function testUnsetPublic()
 
         $this->assertSame(["\0".TestClass::class."\0lazyObjectState"], array_keys((array) $instance));
         unset($instance->public);
-        $this->assertFalse(isset($instance->public));
         $this->assertSame(4, $instance->publicReadonly);
+        $this->expectException(\BadMethodCallException::class);
+        $this->expectExceptionMessage('__isset(public)');
+        isset($instance->public);
```
nicolas-grekas, line 71: This test case was asserting a behavior that relied on a work around with buggy side effects.
It's now asserting the correct behavior.
Fixing that incorrect behavior is how I spotted the root issue: other existing test cases that use the nullsafe operator were failing.
This means the work around for https://github.com/php/php-src/issues/12695 implemented here is already covered by tests.

PR #52763, opened 2023-11-27
docs: how to get the Slack message and channel IDs when the Messenger component is installed

### LINKED ISSUE #52701: Cannot get message ID from sent Slack message if Messenger component is enabled

6.3.5

### Description

I would like to get the message ID and channel ID used in a sent Slack message. However, when sending a Slack message with `symfony/messenger` installed then the value of `$sentMessage` is `null`:

```php
$sentMessage = $chatter->send(new ChatMessage('Original message'));
```

This is not mentioned in the docs at https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Notifier/Bridge/Slack/README.md#updating-a-slack-message

### How to reproduce

Install `symfony/messenger` and `symfony/notifier`. Set up Slack integration as per the [docs](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Notifier/Bridge/Slack/README.md).

Send a Slack message as per [the docs](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Notifier/Bridge/Slack/README.md#updating-a-slack-message) and inspect `$sentMessage`. It is `null` but it should have a value.

### Possible Solution

It might possible to override the constructor arguments for `Symfony\Component\Notifier\Chatter` which are:

```php
public function __construct(TransportInterface $transport, MessageBusInterface $bus = null, EventDispatcherInterface $dispatcher = null)
```

However, none of the following attempting in `services.yaml` worked`:

```yaml
Symfony\Component\Notifier\Chatter:
    public: true
    autowire: false
    autoconfigure: false
    arguments:
        - '@chatter.transports'
        - null
        - null
    # arguments:
    #     $transport: chatter.transports
    #     $bus: null
    #     $dispatcher: null
```

This is most likely because `Chatter` isn't defined as a service in the container:

```
❯ console debug:container | grep Transport | grep chatter
  chatter.transport_factory    Symfony\Component\Notifier\Transport                                                                                  
  chatter.transports                Symfony\Component\Notifier\Transport\Transports  
```

Is there another way to override the arguments used in the `Chatter` constructor?

### Additional Context

When `Chatter` is created, an instance of `MessageBusInterface` is injected, and when the chat message is sent, the message is dispatched on the bus and `null` is returned.

From https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Notifier/Chatter.php#L49-L57:

```php
public function send(MessageInterface $message): ?SentMessage
{
    if (null === $this->bus) {
        return $this->transport->send($message);
    }

    $this->dispatcher?->dispatch(new MessageEvent($message, true));

    $this->bus->dispatch($message);

    return null;
}
```

Using a message handler on the event doesn't give me enough context, as:

1. I need to save the message to the database on an entity, and the message doesn't include the context of which entity and id to use.
2. The message ID and channel ID are not in the message.

BurningDog:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52701
| License       | MIT

Document how to get the Slack message and channel IDs when the Messenger component is installed.

### DIFF
--- a/src/Symfony/Component/Notifier/Bridge/Slack/README.md
+++ b/src/Symfony/Component/Notifier/Bridge/Slack/README.md
@@ -29,7 +29,7 @@ SLACK_DSN=slack://xoxb-......@default?channel=fabien
 Adding Interactions to a Message
 --------------------------------
 
-With a Slack message, you can use the `SlackOptions` class to add some 
+With a Slack message, you can use the `SlackOptions` class to add some
 interactive options called [Block elements](https://api.slack.com/reference/block-kit/block-elements).
 
 ```php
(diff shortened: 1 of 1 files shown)


OskarStark:
Thanks for the PR, however I think this is sth related to `SentMessage` in general and should therefore be part of the docs, as it is valid for all chatter transports. WDYT?

nicolas-grekas:
WDYT @BurningDog? Can you please send the PR to https://github.com/symfony/symfony-docs/?

src/Symfony/Component/Notifier/Bridge/Slack/README.md
```diff
@@ -226,6 +226,55 @@ $options = new UpdateMessageSlackOptions($channelId, $messageId);
...
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            SentMessageEvent::class => 'onSentMessageEvent',
+        ];
+    }
+}
+```
+
+If you need the ID of an entity, put it into your `ChatMessage` subject (which doesn't show in Slack if the message is sent with `SlackOptions`):
```
OskarStark, line 262: I am not sure about the following part, I would not recommend to rely on a string containing data 🤔

src/Symfony/Component/Notifier/Bridge/Slack/README.md
```diff
@@ -226,6 +226,55 @@ $options = new UpdateMessageSlackOptions($channelId, $messageId);
...
+To get the message ID and channel ID from the sent message, create an event subscriber as follows:
+
+```php
+namespace App\EventSubscriber;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Notifier\Bridge\Slack\SlackSentMessage;
+use Symfony\Component\Notifier\Event\SentMessageEvent;
+
+class SlackSentMessageSubscriber implements EventSubscriberInterface
```
OskarStark, line 240: Can we use the Attribute style here?

fabpot:
Closing as there is no more activity.

PR #52764, opened 2023-11-27
[Serializer] Use `DenormalizerAwareInterface` instead of `SerializerAwareInterface`

jschaedl:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

The `UnwrappingDenormalizer` does not depend on the `SerializerInterface`. Using the `DenormalizerInterfacce` is enough, because only `denormalize()` is called. Also, the `SerializerInterface` does not define method `denormalize()` and it currently only works, because at runtime the `serializer` service is injected, which implements the `DenormalizerInterface`.

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/UnwrappingDenormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/UnwrappingDenormalizer.php
@@ -20,9 +20,9 @@ use Symfony\Component\Serializer\SerializerAwareTrait;
 /**
  * @author Eduard Bulava <bulavaeduard@gmail.com>
  */
-final class UnwrappingDenormalizer implements DenormalizerInterface, SerializerAwareInterface
+final class UnwrappingDenormalizer implements DenormalizerInterface, DenormalizerAwareInterface
 {
-    use SerializerAwareTrait;
+    use DenormalizerAwareTrait;
 
     public const UNWRAP_PATH = 'unwrap_path';
 
@@ -51,11 +51,7 @@ final class UnwrappingDenormalizer implements DenormalizerInterface, SerializerA
             $data = $this->propertyAccessor->getValue($data, $propertyPath);
         }
 
-        if (!$this->serializer instanceof DenormalizerInterface) {
-            throw new LogicException('Cannot unwrap path because the injected serializer is not a denormalizer.');
-        }
-
-        return $this->serializer->denormalize($data, $type, $format, $context);
+        return $this->denormalizer->denormalize($data, $type, $format, $context);
     }
 
     public function supportsDenormalization(mixed $data, string $type, string $format = null, array $context = []): bool
--- a/src/Symfony/Component/Serializer/Normalizer/UnwrappingDenormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/UnwrappingDenormalizer.php
@@ -14,8 +14,7 @@ namespace Symfony\Component\Serializer\Normalizer;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
 use Symfony\Component\Serializer\Exception\LogicException;
-use Symfony\Component\Serializer\SerializerAwareInterface;
-use Symfony\Component\Serializer\SerializerAwareTrait;
+use Symfony\Component\Serializer\SerializerInterface;
 
 /**
  * @author Eduard Bulava <bulavaeduard@gmail.com>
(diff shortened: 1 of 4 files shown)


derrabus:
This is a breaking change that we need a deprecation layer for. By removing `SerializerAwareInterface`/`SerializerAwareTrait` from this class, we're effectively removing a public method from that class.

jschaedl:
Hi @derrabus,

> This is a breaking change that we need a deprecation layer for. By removing SerializerAwareInterface/SerializerAwareTrait from this class, we're effectively removing a public method from that class.

You are right, I missed this. I am going to add a deprecation layer.

src/Symfony/Component/Serializer/Normalizer/UnwrappingDenormalizer.php
```diff
@@ -51,15 +50,22 @@ public function denormalize(mixed $data, string $type, string $format = null, ar
...
-        return $this->serializer->denormalize($data, $type, $format, $context);
+        return $this->denormalizer->denormalize($data, $type, $format, $context);
     }
 
     public function supportsDenormalization(mixed $data, string $type, string $format = null, array $context = []): bool
     {
         return \array_key_exists(self::UNWRAP_PATH, $context) && !isset($context['unwrapped']);
     }
+
+    public function setSerializer(SerializerInterface $serializer): void
```
derrabus, line 64: `/** @deprecated */`?

src/Symfony/Component/Serializer/Normalizer/UnwrappingDenormalizer.php
```diff
@@ -14,15 +14,14 @@
...
 use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
 use Symfony\Component\Serializer\Exception\LogicException;
-use Symfony\Component\Serializer\SerializerAwareInterface;
-use Symfony\Component\Serializer\SerializerAwareTrait;
+use Symfony\Component\Serializer\SerializerInterface;
 
 /**
  * @author Eduard Bulava <bulavaeduard@gmail.com>
  */
-final class UnwrappingDenormalizer implements DenormalizerInterface, SerializerAwareInterface
```
derrabus, line 23: I think we should keep `SerializerAwareInterface` as long as we implement `setSerializer()`.

src/Symfony/Component/Serializer/composer.json
```diff
@@ -17,7 +17,8 @@
     ],
     "require": {
         "php": ">=8.2",
-        "symfony/polyfill-ctype": "~1.8"
+        "symfony/polyfill-ctype": "~1.8",
+        "symfony/deprecation-contracts": "^3.4@dev"
```
derrabus: That's a bit too strict. Also, the `@dev` is unnecessary.

```suggestion
        "symfony/deprecation-contracts": "^2.5|^3"
```

derrabus (changes requested):

src/Symfony/Component/Serializer/Normalizer/UnwrappingDenormalizer.php
stof: If we implement SerializerAwareInterface, the Serializer class will be calling the deprecated method.

jschaedl: Yes, that's the reason why I removed `SerializerAwareInterface`. The `Serializer` class will call `setSerializer()` here: https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Serializer/Serializer.php#L85 

But without it, its a bc-break for code relying on `UnwrappingDenormalizer` implementing `SerializerAwareInterface`. I am not sure how to solve this problem.

src/Symfony/Component/Serializer/composer.json
jschaedl: adjusted

src/Symfony/Component/Serializer/Normalizer/UnwrappingDenormalizer.php
jschaedl: deprecated annotation added

dmitryuk:
@jschaedl may be close PR since it too outdated?

nicolas-grekas:
Thanks for spotting this, the observation is correct: `UnwrappingDenormalizer` only needs a denormalizer. The blocker was never the idea but the BC layer, since removing `SerializerAwareInterface` drops a public method while keeping it makes `Serializer` call the deprecated `setSerializer()`, as stof noted, and that was never resolved. Closing for now. This kind of change is much easier to land at the start of a major cycle, so it would be worth bringing back then.

PR #52769, opened 2023-11-28
[WebProfilerBundle] Fix Symfony icon when toolbar is collapsed

javiereguiluz:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

There's a minor but annoying issue with the toolbar when it's collapsed. The "sf" letters in the Symfony logo are transparent instead of white. If the page background is dark, you can't see the clickable icon:

I tried to fix this issue reusing the current SVG file. I couldn't. That icon is used in many parts (in the config profiler panel; in the main profiler layout as the Symfony logo; etc.) When I did some change in the existing SVG, I broke some of the other use cases.

This PR adds a new SVG icon that it's used only when the toolbar is collapsed. The colors are invariant (always white letters over a black circle) but the logo includes a thin border to make it visible in pages with dark backgrounds:


stof (approved):

xabbuh:
Are the changes for the previous icon intended?

javiereguiluz:
Yes, I had to make that change to make everything work ... but yes, it's strange. Let me double check it's correct.

smnandre:
Could we assume that dark background = dark mode 
(in most of the cases) ? 

We maybe could use the Profiler theme option ?

wouterj:
Imho, the stroke is too thin. 1 unit on a viewbox of 115 units almost makes it look like a rendering bug (also due to browsers being somewhat bad at anti aliasing SVG strokes it seems). What about increasing it to 5?

image
from left: current 1 width stroke; proposed 5 width stroke; 1px border with 50% border radius on the SVG

stof:
@smnandre We cannot assume anything as this gets displayed on top of the content of the page, not of something controlled by Symfony.

Nyholm:
> ... but yes, it's strange. Let me double check it's correct.

(friendly ping) 

@javiereguiluz Are you happy with the PR in its current state?

javiereguiluz:
Closing because the bug was not in Symfony 5.4 but in higher versions. I'm replacing this PR with this other PR: #53225. Thanks.

PR #52780, opened 2023-11-29
[DependencyInjection] don't check parameter values if they are not set

### LINKED ISSUE #52768: lint:container command fails on named argument service

5.4.31

### Description

I receive an error like:
```
[ERROR] Invalid definition for service "service.one": argument 1 of
"ServiceOne::__construct()" accepts                    
         "Psr\Log\LoggerInterface", "null" passed.           
```

When executing
```
bin/console lint:container
```

In debugging I noticed it comes from `CheckTypeDeclarationsPass` where `::checkTypeDeclarations` expects a numbered list of arguments. I wonder why that just now fails as we have some more of these already, but anyway.

I checked the class in Symfony 6 and 7 and noticed @nicolas-grekas fixed that in PR https://github.com/symfony/symfony/pull/49944 which does sound like something different in general. Picking just the change of the for loop would fix the command for me.

### How to reproduce

Create a service with named arguments (not sure if this itself is really enough)

### Possible Solution

_No response_

### Additional Context

_No response_

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52768
| License       | MIT


nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @xabbuh.

PR #52786, opened 2023-11-29
[Serializer] Revert allowed attributes fix

mtarld:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

Revert https://github.com/symfony/symfony/pull/52767 and https://github.com/symfony/symfony/pull/52680, as it does not cover some specific use cases, and can't be fixed in time.

I'll attempt to fix it in a better way, taking more time.

PR #52767, opened 2023-11-28
[Serializer] Fix normalization relying on allowed attributes only

mtarld:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52744
| License       | MIT

PR #52765, opened 2023-11-27
[Translation] Remove `@internal` from abstract testcases

OskarStark:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | --
| License       | MIT

The test cases are in the `Test` namespace and therefore under BC promise.
All third-party providers should benefit from it.
We do the same for Mailer and Notifier.

cc @m2mtech as this PR is inspired by your comment in the copied class 😃 in 

https://github.com/m2mtech/weblate-translation-provider/blob/fa38b3c50b5db832e993a0035f516ee0885482b1/tests/ProviderFactoryTestCase.php#L1-L5

@nicolas-grekas or do you want me to target `5.4`?

PR #52727, opened 2023-11-24
[String] Fix Inflector for 'icon'

podhy:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Hello,
this PR fixes problem when inflecting `icon` word into `icons`. Currently it returns wrong result `ica`

Please make someone review because I'm not 100% sure if this is the right way how to fix it.

Thank you.

Btw: older versions are also affected and it should be possible to merge it there

PR #52791, opened 2023-11-29
Release v7.0.0

fabpot:
**Changelog** (https://github.com/symfony/symfony/compare/v7.0.0-RC2...v7.0.0)

 * bug #52786 [Serializer] Revert allowed attributes fix (@mtarld)
 * bug #52765 [Translation] Remove ``@internal`` from abstract testcases (@OskarStark)
 * bug #52780 [DependencyInjection] don't check parameter values if they are not set (@xabbuh)
 * bug #52762 [VarExporter] Work around php/php-src#12695 for lazy objects, fixing nullsafe-related behavior (@nicolas-grekas)
 * bug #52759 [VarExporter] Fix serializing objects that implement __sleep() and that are made lazy (@nicolas-grekas)
 * bug #52767 [Serializer] Fix normalization relying on allowed attributes only (@mtarld)
 * bug #52727 [String] Fix Inflector for 'icon' (@podhy)

### DIFF
--- a/CHANGELOG-7.0.md
+++ b/CHANGELOG-7.0.md
@@ -7,6 +7,16 @@ in 7.0 minor versions.
 To get the diff for a specific change, go to https://github.com/symfony/symfony/commit/XXX where XXX is the change hash
 To get the diff between two versions, go to https://github.com/symfony/symfony/compare/v7.0.0...v7.0.1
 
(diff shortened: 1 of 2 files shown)


GromNaN:
If you want a big picture of what is in this major release, follow this link: https://symfony.com/7

image

PR #52792, opened 2023-11-29
[Form] Fix DateTypeTest

raffaelecarelle:
The date submitted cause an unexcpected Intl date parse error exception.

| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        |  None
| License       | MIT

I've fixed a test related to DateType that failed because submit an invalid format expected throwing a Intl parse exception.

### DIFF
--- a/src/Symfony/Component/Form/Tests/Extension/Core/Type/DateTypeTest.php
+++ b/src/Symfony/Component/Form/Tests/Extension/Core/Type/DateTypeTest.php
@@ -169,7 +169,7 @@ class DateTypeTest extends BaseTypeTestCase
             'input' => 'timestamp',
         ]);
 
-        $form->submit('2.6.2010');
+        $form->submit('02.06.2010');
 
         $dateTime = new \DateTime('2010-06-02 UTC');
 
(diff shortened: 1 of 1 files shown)


yceruto (approved):

xabbuh:
I am not sure about this change. Doesn’t it indicate that existing applications will break when certain ICU versions are installed?

raffaelecarelle:
Hello @xabbuh, I've tested with this ICU versions:

ICU version => 71.1
ICU Data version => 71.1
ICU Unicode version => 14.0

derrabus:
The test in question does not fail on my machine (ICU 73.2), neither with nor without this change.

xabbuh:
I think that explains things. If I remember correctly the behaviour changed with ICU 71. So changing the test would only mean that the CI would pass with ICU 71.1, but that would hide the underlying issue as your users would suddenly not be able anymore to fill the form as they used to before.

xabbuh:
or maybe it's just 71.1 that is buggy (see #52245)

xabbuh:
I updated #52245 instead to skip the failing tests with ICU 71.1, thanks for pushing this forward

PR #52798, opened 2023-11-29
Fix legacy class palceholder definitions for static analysis

ausi:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

In #50392 the `UriSigner` class was moved from HttpKernel to HttpFoundation. The empty definition `class UriSigner {}` causes many issues with static analysis tools like PHPStan.

As this code is wrapped in an `if (false)` changing it to `class UriSigner extends HttpFoundationUriSigner` should not cause any behavior change while fixing most static analysis issues.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Routing/AnnotatedRouteControllerLoader.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Routing/AnnotatedRouteControllerLoader.php
@@ -19,7 +19,7 @@ if (false) {
     /**
      * @deprecated since Symfony 6.4, to be removed in 7.0, use {@link AttributeRouteControllerLoader} instead
      */
-    class AnnotatedRouteControllerLoader
+    class AnnotatedRouteControllerLoader extends AttributeRouteControllerLoader
     {
     }
 }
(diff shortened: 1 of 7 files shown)


GromNaN (commented):
Some other moved classes are not extending the new class:

- `Symfony\Bundle\FrameworkBundle\Routing\AnnotatedRouteControllerLoader`
- `Symfony\Component\HttpKernel\Debug\FileLinkFormatter`
- `Symfony\Component\Routing\Loader\AnnotationClassLoader`
- `Symfony\Component\Routing\Loader\AnnotationDirectoryLoader`
- `Symfony\Component\Routing\Loader\AnnotationFileLoader`
- `Symfony\Component\Serializer\Mapping\Loader\AnnotationLoader`

It may be necessary to do the same change on them.

ausi:
> It may be necessary to do the same change on them.

Will do.

GromNaN (approved):
Thanks. This is consistent with how we deprecated other moved classes.

nicolas-grekas:
Thank you @ausi.

PR #52803, opened 2023-11-29
[String] do not add an extra `s` at the end if last character is `s`

### LINKED ISSUE #52800: Inflector turns settings -> settingss (settings is already plural)

 6.3.9

### Description

It is a continuation of https://github.com/api-platform/core/issues/5957 and https://github.com/symfony/symfony/pull/52526

I discovered that plural of "settings" is now "settingss" - double s.

### How to reproduce

```php
$inflector = new EnglishInflector();
$inflector->pluralize('equipment')); // will return correctly: equipment
$inflector->pluralize('setting')); // will return correctly: settings
$inflector->pluralize('settings')); // will return INCORRECTLY: settingss - it is already plural
```

### Possible Solution

_No response_

### Additional Context

_No response_

k0d3r1s:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52800 
| License       | MIT

As already written in #52800, settings, for example, already end with `s`, so assumption to add extra `s` at the end is incorrect

### DIFF
--- a/src/Symfony/Component/Validator/Resources/translations/validators.pl.xlf
+++ b/src/Symfony/Component/Validator/Resources/translations/validators.pl.xlf
@@ -434,6 +434,10 @@
                 <source>The detected character encoding is invalid ({{ detected }}). Allowed encodings are {{ encodings }}.</source>
                 <target>Wykryte kodowanie znaków ({{ detected }}) jest nieprawidłowe. Dozwolone kodowania to {{ encodings }}.</target>
             </trans-unit>
+            <trans-unit id="112">
+                <source>This is not a valid MAC address.</source>
+                <target>Podana wartość nie jest poprawnym adresem MAC.</target>
+            </trans-unit>
         </body>
     </file>
 </xliff>


OskarStark:
Can you please add a test case to avoid further regressions? Thank you

acirulis:
How about words, such as :
Corps -> corps
Diagnosis -> diagnoses
Ethos -> ethe or ethea or ethoses
Gas -> gases

Current output of Inflector is as follows:
```
Corpss
Diagnoses # only one correct
Ethoss
Gass
```

k0d3r1s:
@OskarStark added tests

\+ extra words from comments

derrabus:
The inflector is meant to pluralize singular terms. That's its purpose. If the term in question is already a plural, you should not use it. And neither is this class meant to detect or normalize plural forms.

acirulis:
@derrabus maybe so but this issue started when Api Platform switched from doctrine/inflector (which does check if plural form is already used) to symfony/string (not sure the reasoning). Please see https://github.com/api-platform/core/issues/5957

fiburonsk:
I came across `alias` it should be `aliases` but the EnglishInflector generates `aliass` instead, similar to gas and gases above.

PR #52804, opened 2023-11-29
[Serializer] Fix support of plain object types denormalization

andersonamuller:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

When denormalizing properties with a plain `object` type without a class the `ObjectNormalizer` would fail with a `TypeError`.

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
@@ -584,7 +584,7 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
 
                 $expectedTypes[Type::BUILTIN_TYPE_OBJECT === $builtinType && $class ? $class : $builtinType] = true;
 
-                if (Type::BUILTIN_TYPE_OBJECT === $builtinType) {
+                if (Type::BUILTIN_TYPE_OBJECT === $builtinType && null !== $class) {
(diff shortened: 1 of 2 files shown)


OskarStark:
Can you please add test case to avoid further regressions? Thank you!

andersonamuller:
> Can you please add test case to avoid further regressions? Thank you!

I added a test case

OskarStark:
friendly ping @mtarld as you are deeply involved in the serializer component, thanks!

mtarld (commented):
I'd do instead:
```php
if (Type::BUILTIN_TYPE_OBJECT === $builtinType && null !== $class) {
```
By doing that, it'll fallback to:
```php
if (('is_'.$builtinType)($data)) {
```
And IMHO, it might be better to handle a raw object as a builtin type. WDYT?

andersonamuller:
> I'd do instead:
> 
> ```
> if (Type::BUILTIN_TYPE_OBJECT === $builtinType && null !== $class) {
> ```
> 
> By doing that, it'll fallback to:
> 
> ```
> if (('is_'.$builtinType)($data)) {
> ```
> 
> And IMHO, it might be better to handle a raw object as a builtin type. WDYT?

Makes sense, I have pushed the change.

mtarld (approved):

nicolas-grekas:
Thank you @andersonamuller.

Issue #52801, opened 2023-11-29
Route definitions incompatible with `symfony/routing: 6.4.0`


fritzmg:
### Symfony version(s) affected

6.4.0

### Description

Currently I usually define my routes this way:

```php
#Route(path: '/foobar', name: self::class)
class FoobarController
{
}
```

This way I can use the FQCN to generate the controller's URL:

```php
$this->urlGenerator->generate(FoobarController::class)
```

This definition does not work in Symfony `6.4.0` anymore:

```
 [Symfony\Component\Config\Exception\LoaderLoadException]                                                                                                    
  Route alias "App\Controller\FoobarController" can not reference itself in src/Controller (which is being imported from "config/routes.yaml"). Make sure there is a loader supporting the "annotation" type.                                             
                                                                                                                                                              

Exception trace:
  at vendor\symfony\config\Loader\FileLoader.php:182
 Symfony\Component\Config\Loader\FileLoader->doImport() at vendor\symfony\config\Loader\FileLoader.php:98
 Symfony\Component\Config\Loader\FileLoader->import() at vendor\symfony\routing\Loader\YamlFileLoader.php:212
 Symfony\Component\Routing\Loader\YamlFileLoader->parseImport() at vendor\symfony\routing\Loader\YamlFileLoader.php:99
```

Changing the definition to

```php
#Route(path: '/foobar')
class FoobarController
{
}
```

or another name for the route fixes the error, but I want to use the FQCN as the route name.

### How to reproduce

Create a controller like this:

```php
// src/Controller/FoobarController.php
namespace App\Controller;

use Symfony\Component\HttpFoundation\Response;
use Symfony\Component\Routing\Annotation\Route;

#[Route('/foobar', self::class)]
class FoobarController
{
    public function __invoke(): Response
    {
        return new Response('Hello World!');
    }
}
```

Then execute a `cache:warmup`.

### Possible Solution

_No response_

### Additional Context

_No response_

HypeMC:
The issue was introduced in https://github.com/symfony/symfony/pull/50084. The simplest solution would be to remove `self::class` since this is now done implicitly:

```diff
-#Route(path: '/foobar', name: self::class)
+#Route(path: '/foobar')
class FoobarController
{
}
```

fritzmg:
> The simplest solution would be to remove self::class since this is now done implicitly:

[...]

PR #52805, opened 2023-11-29
[Routing] Fix conflicting FQCN aliases with route name

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | https://github.com/symfony/symfony/issues/52801
| License       | MIT

When the route name === the FQCN alias we want to add, `addAlias()` throws. In this case, we can safely ignore the exception and move on.

### DIFF
--- a/src/Symfony/Component/Routing/Loader/AttributeClassLoader.php
+++ b/src/Symfony/Component/Routing/Loader/AttributeClassLoader.php
@@ -144,7 +144,9 @@ abstract class AttributeClassLoader implements LoaderInterface
 
                 if (1 === $collection->count() - \count($routeNamesBefore)) {
                     $newRouteName = current(array_diff(array_keys($collection->all()), $routeNamesBefore));
-                    $collection->addAlias(sprintf('%s::%s', $class->name, $method->name), $newRouteName);
+                    if ($newRouteName !== $aliasName = sprintf('%s::%s', $class->name, $method->name)) {
+                        $collection->addAlias($aliasName, $newRouteName);
+                    }
                 }
             }
             if (0 === $collection->count() && $class->hasMethod('__invoke')) {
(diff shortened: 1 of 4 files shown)


src/Symfony/Component/Routing/Loader/AttributeClassLoader.php
```diff
@@ -144,7 +145,10 @@ public function load(mixed $class, string $type = null): RouteCollection
 
                 if (1 === $collection->count() - \count($routeNamesBefore)) {
                     $newRouteName = current(array_diff(array_keys($collection->all()), $routeNamesBefore));
-                    $collection->addAlias(sprintf('%s::%s', $class->name, $method->name), $newRouteName);
+                    try {
+                        $collection->addAlias(sprintf('%s::%s', $class->name, $method->name), $newRouteName);
```
fancyweb: We could also assign everything to variables, compare and call `addAlias()` only when the name & alias are different.

src/Symfony/Component/Routing/Tests/Fixtures/AttributeFixtures/InvokableFQCNAliasConflictController.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Symfony\Component\Routing\Tests\Fixtures\AttributeFixtures;
+
+use Symfony\Component\Routing\Attribute\Route;
+
+#[Route(path: '/foobarccc', name: self::class)]
```
fancyweb, line 7: I added only this test fixture case but the 3 FQCN aliases are susceptible to fail.

src/Symfony/Component/Routing/Loader/AttributeClassLoader.php
nicolas-grekas: or we skip throwing in this case in addAlias?
I'm not sure this exception makes sense to me

fancyweb: Yes, I agree but we'd change the behavior. Surely someone, somewhere, relies on this exception :sweat_smile:

nicolas-grekas: mmm :) 
then I'd prefer checking the value and not the exception
one day we might throw for another reason

fancyweb: Works for me!

src/Symfony/Component/Routing/Loader/AttributeClassLoader.php
```diff
@@ -144,7 +145,9 @@ public function load(mixed $class, string $type = null): RouteCollection
 
                 if (1 === $collection->count() - \count($routeNamesBefore)) {
                     $newRouteName = current(array_diff(array_keys($collection->all()), $routeNamesBefore));
-                    $collection->addAlias(sprintf('%s::%s', $class->name, $method->name), $newRouteName);
+                    if ($newRouteName !== ($aliasName = sprintf('%s::%s', $class->name, $method->name))) {
```
nicolas-grekas: no need for the brackets after !== :)

stof (approved):

nicolas-grekas:
Thank you @fancyweb.

Issue #52802, opened 2023-11-29
[Routing] Prefix feature for i18n routes is broken


plandolt:
### Symfony version(s) affected

6.4.0

### Description

When trying to translate the url using the [router prefix feature](https://symfony.com/doc/current/routing.html#localized-routes-i18n) the following error occurs:
`Target route "app_your_route_name" for alias "App\Controller\YourController::index" does not exist.`

### How to reproduce

- Create a fresh installation of 6.4.0 (or upgrade a fresh installation from 6.3 to 6.4). 
- Create a dummy controller (e.g. with the maker bundle)
- Change `routes.yaml` (https://symfony.com/doc/current/routing.html#localized-routes-i18n):

```diff
controllers:
    resource:
        path: ../src/Controller/
        namespace: App\Controller
    type: attribute
+   prefix:
+       de: '/de'
+       fr: '/fr'
+       it: '/it'
```

- run `bin/console cache:clear`

HypeMC:
The issue was introduced in https://github.com/symfony/symfony/pull/50084.

HypeMC:
I think the problem is this line: https://github.com/symfony/symfony/blob/914496290fc8777edbb8e58db772b106c2d99c87/src/Symfony/Component/Routing/Loader/AttributeClassLoader.php#L145-L147

It seems to be missing the `$fqcnAlias && ` condition.

plandolt:
> I think the problem is this line:
> 
> https://github.com/symfony/symfony/blob/914496290fc8777edbb8e58db772b106c2d99c87/src/Symfony/Component/Routing/Loader/AttributeClassLoader.php#L145-L147
> 
> It seems to be missing the `$fqcnAlias && ` condition.

This seems to solve the problem for Controller Actions that do not use `__invoke` as its method name. With __invoke the same error still occurs:
`Target route "app_dummy" for alias "App\Controller\DummyController::__invoke" does not exist.`

HypeMC:
Yeah, it looks like the original [PR for aliases](https://github.com/symfony/symfony/pull/38464) never covered aliasing routes with prefixes.

fancyweb:
Thank you for your report, I'll have a look.

fancyweb:
I think it comes from an already existing bug on 5.4 :sweat_smile: Removing a route doesn't remove aliases pointing to it as expected.

PR #52806, opened 2023-11-29
[Routing] Fix removing aliases pointing to removed route in `RouteCollection::remove()`

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | 
| Issues        | https://github.com/symfony/symfony/issues/52802
| License       | MIT

When a route is removed from the collection, we want to remove all aliases pointing to it. The implementation is flawed because the `$aliases` array is indexed by the alias name, and not by the route since several aliases can point to the same route.

It'll fix the related issue because added FQCN aliases will be correctly removed when  `PrefixTrait::addPrefix()` removes the route before reading it for each locale.

### DIFF
--- a/src/Symfony/Component/Routing/RouteCollection.php
+++ b/src/Symfony/Component/Routing/RouteCollection.php
@@ -157,8 +157,15 @@ class RouteCollection implements \IteratorAggregate, \Countable
      */
     public function remove($name)
     {
-        foreach ((array) $name as $n) {
-            unset($this->routes[$n], $this->priorities[$n], $this->aliases[$n]);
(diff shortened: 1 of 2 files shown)


yceruto (approved):

nicolas-grekas:
Thank you @fancyweb.

stof:
AFAICT, the goal was not to remove the aliases pointing to a route but really to remove the alias when passing the alias name.

stof:
This PR makes it impossible to remove an alias.

PR #52811, opened 2023-11-30
[Messenger] Avoid reconnecting an already-connected Redis instance.

### LINKED ISSUE #45057: [Messenger (Redis transport)] Connection should use custom Redis instance
### Description

If I provide custom Redis instance to `Symfony\Component\Messenger\Bridge\Redis\Transport\Connection` and it's already connected, Redis goes "Server went away". Thats because(probably) its already connected.

Problem is in `symfony/redis-messenger/Transport/Connection.php:76`. There should be another elseif with `elseif ($redis && $redis->isConnected()) { $this->connection = $redis; }`

### Example

```
class MessengerTransportFactory extends RedisTransportFactory
{
    public function createTransport(string $dsn, array $options, SerializerInterface $serializer): TransportInterface
    {
         return new RedisTransport(
            new Connection([], [], [], $CUSTOM_REDIS_INSTANCE),
            $serializer
        );
    }
}
```

BusterNeece:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | yes?
| New feature?  | yes?
| Deprecations? | no
| Issues        | Fix #45057
| License       | MIT

With the Messenger component's internal Redis `Connection` class, if you pass it a `\Redis` or `\Relay` instance that's already connected, it will still try to reconnect that instance using the host/port/etc. that's passed to the connection object's "options" array in the constructor.

Not only does this cause problems when the Redis instance is already connected, but it also prevents you from, say, using socket connections instead (as the only connection method supported by the class itself is IP-based connections) and just passing in a ready-to-go Redis instance.

While I personally use this class heavily in my own applications, I'm not aware of every possible implication of making a change like this, i.e. if there are other parts of the application that are counting on this working as-is (even though I can't imagine that unnecessarily reconnecting is a desired feature in this case). While it's intended to fix a bug in my specific use-case, I'd consider it also arguably "new" functionality that the class can now handle pre-connected instances being passed to it.

Basically, I'm new to contributing to Symfony so I'm a little shaky on my legs here, but the reported issue above mentioned needing a PR, and we're running into the issue ourselves, so here's the PR!

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Redis/Transport/Connection.php
+++ b/src/Symfony/Component/Messenger/Bridge/Redis/Transport/Connection.php
@@ -166,9 +166,11 @@ class Connection
      */
     private static function initializeRedis(\Redis|Relay $redis, string $host, int $port, string|array|null $auth, array $params): \Redis|Relay
     {
-        $connect = isset($params['persistent_id']) ? 'pconnect' : 'connect';
-        $redis->{$connect}($host, $port, $params['timeout'], $params['persistent_id'], $params['retry_interval'], $params['read_timeout'], ...(\defined('Redis::SCAN_PREFIX') || \extension_loaded('relay')) ? [['stream' => $params['ssl'] ?? null]] : []);
-
+        if (!$redis->isConnected()) {
+            $connect = isset($params['persistent_id']) ? 'pconnect' : 'connect';
+            $redis->{$connect}($host, $port, $params['timeout'], $params['persistent_id'], $params['retry_interval'], $params['read_timeout'], ...(\defined('Redis::SCAN_PREFIX') || \extension_loaded('relay')) ? [['stream' => $params['ssl'] ?? null]] : []);
+        }
+        
         $redis->setOption($redis instanceof \Redis ? \Redis::OPT_SERIALIZER : Relay::OPT_SERIALIZER, $params['serializer']);
 
         if (null !== $auth && !$redis->auth($auth)) {


src/Symfony/Component/Messenger/Bridge/Redis/Transport/Connection.php
```diff
@@ -166,9 +166,11 @@ public function __construct(array $options, \Redis|Relay|\RedisCluster $redis =
      */
     private static function initializeRedis(\Redis|Relay $redis, string $host, int $port, string|array|null $auth, array $params): \Redis|Relay
     {
-        $connect = isset($params['persistent_id']) ? 'pconnect' : 'connect';
-        $redis->{$connect}($host, $port, $params['timeout'], $params['persistent_id'], $params['retry_interval'], $params['read_timeout'], ...(\defined('Redis::SCAN_PREFIX') || \extension_loaded('relay')) ? [['stream' => $params['ssl'] ?? null]] : []);
-
+        if (!$redis->isConnected()) {
```
nicolas-grekas, line 169: What about auth, serializer option and dbindex? should we consider them already initialized?

```suggestion
        if ($redis->isConnected()) {
            return $redis;
        }
```

nicolas-grekas (commented):
I'd be fine considering this as a bugfix, can you please target 5.4?

BusterNeece:
@nicolas-grekas Thank you for the review! I've submitted an updated version against the 5.4 branch in PR #52896.

PR #52814, opened 2023-11-30
[Workflow] Add `getEnabledTransition()` to TraceableWorkflow

### LINKED ISSUE #52812: Twig function workflow_transition is broken when using TraceableWorkflow

6.4.0

### Description

In dev environment, when using the `workflow_transition()` (not to be confused with `workflow_transactions()`) in a twig template to get a specific workflow transition, the following error occurs:
```
An exception has been thrown during the rendering of a template ("Call to undefined method Symfony\Component\Workflow\Debug\TraceableWorkflow::getEnabledTransition()").
```

This is due to the `TraceableWorkflow`, introduced in Symfony 6.4.0 not having the `getEnabledTransition` method implemented, which _is_ implemented in the 'normal' `Workflow` class. This method is called from the `symfony/twig-bridge/Extension/WorkflowExtension.php` file.

The `WorkflowInterface`, which is implemented by both the `Workflow` and `TraceableWorkflow` classes does not contain the `getEnabledTransaction` method as well.

Note that this isn't an issue in the prod environment, as the `TraceableWorkflow` isn't used there.

### How to reproduce

Have a Symfony application with the workflow component set up, running in dev environment.
Then, call the `workflow_transition()` function in a twig template and render the page.

### Possible Solution

The best solution would be to add the `getEnabledTransition` method to the `WorkflowInterface` and implement it in `TraceableWorkflow`.

However, since `workflow_transitions()` (plural) does work, a temporary workaround can be instead of:
```twig
{% set myTransition = workflow_transition(myWorkflow, 'myTransition') %}
```

Doing it like this:
```twig
{% set myTransition = null %}
{% for transition in workflow_transitions(myWorkflow) %}
    {% if transition.name == 'myTransition' %}
        {% set myTransition = transition %}
    {% endif %}
{% endfor %}
```
It's not an elegant solution, but it works.

### Additional Context

_No response_

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52812
| License       | MIT

I added an annotation to `WorkflowInterface` to not break BC. I think it could deserve an additional `not implementing this method is deprecated since Symfony 7.1` while upmerging (I guess ?).

Added the tests while at it.

### DIFF
--- a/src/Symfony/Component/Workflow/Debug/TraceableWorkflow.php
+++ b/src/Symfony/Component/Workflow/Debug/TraceableWorkflow.php
@@ -16,6 +16,7 @@ use Symfony\Component\Workflow\Definition;
 use Symfony\Component\Workflow\Marking;
 use Symfony\Component\Workflow\MarkingStore\MarkingStoreInterface;
 use Symfony\Component\Workflow\Metadata\MetadataStoreInterface;
+use Symfony\Component\Workflow\Transition;
 use Symfony\Component\Workflow\TransitionBlockerList;
 use Symfony\Component\Workflow\WorkflowInterface;
 
@@ -53,6 +54,11 @@ class TraceableWorkflow implements WorkflowInterface
     }
 
     public function getEnabledTransitions(object $subject): array
+    {
+        return $this->callInner(__FUNCTION__, \func_get_args());
+    }
+
+    public function getEnabledTransition(object $subject, string $name): ?Transition
     {
         return $this->callInner(__FUNCTION__, \func_get_args());
     }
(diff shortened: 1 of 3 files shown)


lyrixx (approved):
Good catch, thanks 👍🏼

src/Symfony/Component/Workflow/WorkflowInterface.php
```diff
@@ -19,6 +19,8 @@
  * Describes a workflow instance.
  *
  * @author Amrouche Hamza <hamza.simperfit@gmail.com>
+ *
+ * @method Transition|null getEnabledTransition(object $subject, string $name)
```
xabbuh: This should be reverted. We can do this change in 7.1.

alexandre-daubois: Alright, I'll add it if it isn't done during upmerge :)

xabbuh (approved):

nicolas-grekas:
Thank you @alexandre-daubois.

PR #52817, opened 2023-11-30
[Serializer] Do not instantiate object if it is not instantiable

maxbaldanza:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

If you pass an object that can't be instantiable such as enum to deserialize then you get the following
error `Error: Cannot instantiate enum` as the object is tried to be created without checking if it's instantiable

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
@@ -458,6 +458,15 @@ abstract class AbstractNormalizer implements NormalizerInterface, DenormalizerIn
 
         unset($context['has_constructor']);
 
+        if (!$reflectionClass->isInstantiable()) {
+            throw NotNormalizableValueException::createForUnexpectedDataType(
+                sprintf('Failed to create object because the class "%s" is not instantiable.', $class),
+                $data,
+                ['unknown'],
+                $context['deserialization_path'] ?? null,
+            );
+        }
+
         return new $class();
     }
 
--- a/src/Symfony/Component/Serializer/Tests/Normalizer/AbstractNormalizerTest.php
+++ b/src/Symfony/Component/Serializer/Tests/Normalizer/AbstractNormalizerTest.php
@@ -15,6 +15,7 @@ use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\PropertyInfo\Extractor\PhpDocExtractor;
 use Symfony\Component\Serializer\Encoder\JsonEncoder;
+use Symfony\Component\Serializer\Exception\NotNormalizableValueException;
 use Symfony\Component\Serializer\Mapping\AttributeMetadata;
 use Symfony\Component\Serializer\Mapping\ClassMetadata;
 use Symfony\Component\Serializer\Mapping\Factory\ClassMetadataFactory;
@@ -32,6 +33,7 @@ use Symfony\Component\Serializer\Tests\Fixtures\NullableConstructorArgumentDummy
 use Symfony\Component\Serializer\Tests\Fixtures\NullableOptionalConstructorArgumentDummy;
 use Symfony\Component\Serializer\Tests\Fixtures\StaticConstructorDummy;
 use Symfony\Component\Serializer\Tests\Fixtures\StaticConstructorNormalizer;
+use Symfony\Component\Serializer\Tests\Fixtures\UnitEnumDummy;
 use Symfony\Component\Serializer\Tests\Fixtures\VariadicConstructorTypedArgsDummy;
 
 /**
@@ -278,5 +280,17 @@ class AbstractNormalizerTest extends TestCase
         $normalizer = new PropertyNormalizer($this->classMetadata);
 
         $this->assertSame([], $normalizer->normalize($dummy));
+    }
+
+    /**
+     * @requires PHP 8.1
+     */
+    public function testDenormalizeWhenObjectNotInstantiable()
+    {
+        $this->expectException(NotNormalizableValueException::class);
+
+        $normalizer = new ObjectNormalizer();
+
+        $normalizer->denormalize('{}', UnitEnumDummy::class);
     }
 }


maxbaldanza:
I'm not sure if the psalm error is related to something I've changed here as it fails when generating the baseline. 
If it's something I've changed then let me know and I can take a look 😄

src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
```diff
@@ -458,6 +458,13 @@ protected function instantiateObject(array &$data, string $class, array &$contex
 
         unset($context['has_constructor']);
 
+        if ($reflectionClass->isInstantiable() === false) {
+            throw NotNormalizableValueException::createForUnexpectedDataType(
```
maxbaldanza, line 462: Wasn't sure if a new exception would be better or a different one may be better so let me know your thoughts

src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
```diff
@@ -458,6 +458,13 @@ protected function instantiateObject(array &$data, string $class, array &$contex
 
         unset($context['has_constructor']);
 
+        if ($reflectionClass->isInstantiable() === false) {
```
nicolas-grekas:
```suggestion
        if (!$reflectionClass->isInstantiable()) {
```

src/Symfony/Component/Serializer/Tests/SerializerTest.php
```diff
@@ -1411,6 +1412,28 @@ public function testNoCollectDenormalizationErrorsWithWrongEnumOnConstructor()
...
+            ],
+            ['json' => new JsonEncoder()]
+        );
+
+        try {
+            $serializer->deserialize('{}', UnitEnumDummy::class, 'json', [
+                DenormalizerInterface::COLLECT_DENORMALIZATION_ERRORS => true,
+            ]);
+        } catch (\Throwable $th) {
+            $this->assertInstanceOf(NotNormalizableValueException::class, $th);
```
nicolas-grekas:
```suggestion
        } catch (\Throwable $e) {
            $this->assertInstanceOf(NotNormalizableValueException::class, $e);
```

but this should use expectException instead

src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
maxbaldanza: Updated 👍

src/Symfony/Component/Serializer/Tests/SerializerTest.php
maxbaldanza: Agreed it makes more sense to use expectException so updated the test 👍

src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
```diff
@@ -458,6 +458,14 @@ protected function instantiateObject(array &$data, string $class, array &$contex
 
         unset($context['has_constructor']);
 
+        if (!$reflectionClass->isInstantiable()) {
```
mtarld, line 461: What about putting it at the top of the method to fail as fast as possible?

src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
```diff
@@ -458,6 +458,14 @@ protected function instantiateObject(array &$data, string $class, array &$contex
...
+            throw NotNormalizableValueException::createForUnexpectedDataType(
+                sprintf('Failed to create object because the class "%s" is not instantiable.', $class),
+                $data,
+                []
```
mtarld:
```suggestion
                ['unknown'],
                $context['deserialization_path'] ?? null,
                false,
                $e->getCode(),
                $e,
```

src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
mtarld: This exception looks good to me

src/Symfony/Component/Serializer/Tests/SerializerTest.php
```diff
@@ -716,6 +717,23 @@ public function testDeserializeInconsistentScalarType()
...
+    /**
+     * @requires PHP 8.1
+     */
+    public function testDeserializeWhenObjectNotInstantiable()
+    {
+        $this->expectException(NotNormalizableValueException::class);
+
+        $serializer = new Serializer(
+            [
+                new BackedEnumNormalizer(),
```
mtarld: I don't think that it's needed

src/Symfony/Component/Serializer/Tests/SerializerTest.php
```diff
@@ -716,6 +717,23 @@ public function testDeserializeInconsistentScalarType()
         $serializer->deserialize('"42"', 'int', 'json');
     }
 
+    /**
+     * @requires PHP 8.1
+     */
+    public function testDeserializeWhenObjectNotInstantiable()
```
mtarld: What about moving that test to the `AbstractNormalizerTest` as it's only related to the denormalization?

src/Symfony/Component/Serializer/Tests/SerializerTest.php
```diff
@@ -716,6 +717,23 @@ public function testDeserializeInconsistentScalarType()
...
+        $this->expectException(NotNormalizableValueException::class);
+
+        $serializer = new Serializer(
+            [
+                new BackedEnumNormalizer(),
+                new ObjectNormalizer(),
+            ],
+            ['json' => new JsonEncoder()]
+        );
+        $serializer->deserialize('{}', UnitEnumDummy::class, 'json');
```
mtarld:
```suggestion
        $serializer->denormalize('{}', UnitEnumDummy::class);
```

mtarld (changes requested):

src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
maxbaldanza: Thanks 👍

src/Symfony/Component/Serializer/Tests/SerializerTest.php
maxbaldanza: I've moved it 👍

src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
maxbaldanza: Can't add it to the top because in case of static classes where the constructor is private for instance the `StaticConstructorDummy`.
Would have to also check not static so not sure it makes sense but happy to do so if you think it's worth it

src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
maxbaldanza: I've added `['unknown'],` and `$context['deserialization_path'] ?? null,`
There is no exception though so can't add the other two

src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
mtarld: Indeed, nice catch!

mtarld (approved):

maxbaldanza:
@nicolas-grekas / @mtarld anything left for me to do on this one?

src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
```diff
@@ -458,6 +458,15 @@ protected function instantiateObject(array &$data, string $class, array &$contex
 
         unset($context['has_constructor']);
 
+        if (!$reflectionClass->isInstantiable()) {
+            throw NotNormalizableValueException::createForUnexpectedDataType(
+                sprintf('Failed to create object because the class "%s" is not instantiable.', $class),
+                $data,
+                ['unknown'],
+                $context['deserialization_path'] ?? null
```
fabpot:
```suggestion
                $context['deserialization_path'] ?? null,
```

fabpot (approved):

fabpot:
Thank you @maxbaldanza.

PR #52820, opened 2023-11-30
[DependencyInjection] Add `#[AutowireInline]` attribute to allow service definition at the class level

### LINKED ISSUE #52819: [DependencyInjection] Add new attributes to allow inlined services while using autowiring
### Description

Currently when using autowiring and declaring service configuration on class level it's (as far as I know) not possible to define a inlined service or define a factory that may create an instance of the service with custom parameters.

So let's take a look at this:
```php
class SomeSourceAwareLogger
{
    public function __construct(
        private readonly LoggerInterface $logger,
        private readonly string $someSource,
    ) {
    }
}

class SomeClass
{
    public function __construct(
        private readonly SomeSourceAwareLogger $logger,
    ) {
    }
}
```

This would currently lead to an error when trying to build an instance of `SomeClass` since `SomeSourceAwareLogger::$someSource` is a string parameter that the autowiring can not resolve.

So for this to work one would either have to
a) add an `#[Autowire]` attribute to `SomeSourceAwareLogger::$someSource` and would have to create multiple classes for every source
b) configure one or multiple instances of `SomeSourceAwareLogger` in the config files with different sources and then configure `SomeClass::logger` as well

This would then look like this:
```php
class SomeSourceAwareLogger1
{
    public function __construct(
        private readonly LoggerInterface $logger,
        #[Autowire(param: 'someSource1')]
        private readonly string $someSource,
    ) {
    }
}

class SomeSourceAwareLogger2
{
    public function __construct(
        private readonly LoggerInterface $logger,
        #[Autowire(param: 'someSource2')]
        private readonly string $someSource,
    ) {
    }
}

class SomeClass
{
    public function __construct(
        private readonly SomeSourceAwareLogger1 $logger1,
        private readonly SomeSourceAwareLogger2 $logger2,
    ) {
    }
}

// OR
class SomeSourceAwareLogger
{
    public function __construct(
        private readonly LoggerInterface $logger,
        private readonly string $someSource,
    ) {
    }
}

class SomeClass
{
    public function __construct(
        #[Autowire(service: 'someSourceAwareLogger1')]
        private readonly SomeSourceAwareLogger $logger1,
        #[Autowire(service: 'someSourceAwareLogger2')]
        private readonly SomeSourceAwareLogger $logger2,
    ) {
    }
}

// In services.yaml or wherever:
// services:
//    someSourceAwareLogger1:
//        class: SomeSourceAwareLogger
//        arguments:
//            - '@LoggerInterface'
//            - 'someSource1'
//
//    someSourceAwareLogger2:
//        class: SomeSourceAwareLogger
//        arguments:
//            - '@LoggerInterface'
//            - 'someSource2'
```

What I'd like to introduce is having it just at a single point by saying something like:
```php
class SomeSourceAwareLogger
{
    public function __construct(
        private readonly LoggerInterface $logger,
        private readonly string $someSource,
    ) {
    }
}

// And then just either:

class SomeClass
{
    public function __construct(
        #[InlineService(class: SomeSourceAwareLogger::class, params: [new Reference(LoggerInterface::class), self::class])]
        private readonly SomeSourceAwareLogger $logger,
    ) {
    }
}

// OR

class SomeClass
{
    public function __construct(
        #[InlineFactory(service: SomeSourceAwareLoggerFactory::class, params: [self::class])]
        private readonly SomeSourceAwareLogger $logger,
    ) {
    }
}
```

### Example

_No response_

DaDeather:
| Q             | A
| ------------- | ---
| Branch       | 7.1
| Bug fix      | no
| New feature  | yes
| Deprecations | no
| Issues        | Fix #52819 
| License       | MIT

For the idea behind this feature see the issue that contains examples #52819 

### Example usage:
```php
class SomeSourceAwareLogger
{
    public function __construct(
        private readonly LoggerInterface $logger,
        private readonly string $someSource,
    ) {
    }
}

class SomeSourceAwareLoggerFactory
{
    public function __construct(
        private readonly LoggerInterface $logger,
    ) {
    }

    public function create(string $someSource): SomeSourceAwareLogger
    {
        return new SomeSourceAwareLogger($this->logger, $someSource);
    }

    public static function staticCreate(LoggerInterface $logger, string $someSource): SomeSourceAwareLogger
    {
        return new SomeSourceAwareLogger($logger, $someSource);
    }
}

// -----------

class SomeClass1
{
    public function __construct(
        #[AutowireInline(class: SomeSourceAwareLogger::class, args: [new Reference(LoggerInterface::class), 'bar'])]
        public SomeSourceAwareLogger $someSourceAwareLogger,
    ) {
    }
}

// AND/OR

class SomeClass2
{
    public function __construct(
        #[AutowireInline(
            class: SomeSourceAwareLogger::class,
            factory: [SomeSourceAwareLoggerFactory::class, 'staticCreate'],
            args: [new Reference(LoggerInterface::class), 'someParam'],
        )]
        public SomeSourceAwareLogger $someSourceAwareLogger,
    ) {
    }
}

// AND/OR

class SomeClass3
{
    public function __construct(
        #[AutowireInline(
            class: SomeSourceAwareLogger::class,
            factory: [new Reference(SomeSourceAwareLoggerFactory::class), 'create'],
            args: ['someParam'],
        )]
        public SomeSourceAwareLogger $someSourceAwareLogger,
    ) {
    }
}
```


nicolas-grekas:
Thanks for proposing and for submitting!

I'm wondering if we couldn't get rid of the interface and rename the proposed argument to e.g. `AutowireInline`?
To remove the interface, we could maybe make AutowireCallable extend AutowireInline.
Also, instead of a dedicated attribute for services built by factory, can't we add a `factory` argument to the AutowireInline attribute?

DaDeather:
> Thanks for proposing and for submitting!
> 
> I'm wondering if we couldn't get rid of the interface and rename the proposed argument to e.g. `AutowireInline`? To remove the interface, we could maybe make AutowireCallable extend AutowireInline. Also, instead of a dedicated attribute for services built by factory, can't we add a `factory` argument to the AutowireInline attribute?

Sure I was unsure about the naming anyway though :see_no_evil:.

About the interface:
Wouldn't it rather make sense to keep it to allow extending / allowing custom Attributes that may create definitions that could be used for the Autowiring to be handled?

I'd rather see it as a benefit there allowing one to implement the interface and go for it instead of extending the base classes like the `AutowireInline` (after renaming). Unless this may be undesired anyway :shrug:. Just being curious here.

Combining the factory thing into the `AutowireInline` seems fine as well since both are inlined :+1:.

I made the adjustments according to your comment unless the one with the Interface. Just let me know if I still should remove it / rename it or whatever :+1:.

And thanks for your fast feedback on this!

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
```diff
@@ -0,0 +1,72 @@
...
+#[\Attribute(\Attribute::TARGET_PARAMETER)]
+class AutowireInline extends Autowire implements BuildDefinitionInterface
+{
+    /**
+     * @param class-string|null $class
+     * @param array             $params Static parameters to pass to instance creation
+     * @param bool|class-string $lazy   Whether to use lazy-loading for this argument
+     */
+    public function __construct(
+        private readonly ?string $class = null,
```
nicolas-grekas: In `CheckDefinitionValidityPass`  we ask ppl to give a class even when a factory is used so we should make this required

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
```diff
@@ -0,0 +1,72 @@
...
+        private readonly null|string|array $factory = null,
+        private readonly array $params = [],
+        bool|string $lazy = false,
```
nicolas-grekas: we also need a "$calls" argument for method calls I think

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
```diff
@@ -0,0 +1,72 @@
...
+    ) {
+        if (!(null !== $class xor null !== $factory)) {
+            throw new LogicException('#[AutowireInline] attribute must declare exactly one of $class or $factory.');
+        }
+
+        if (
+            \is_array($factory)
+            && false === \is_string($factory[0])
```
nicolas-grekas:
```suggestion
            && !\is_string($factory[0])
            && null !== $factory[0]
```

it should be possible to put `null` here, this allows using the class of the definition instead

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
```diff
@@ -0,0 +1,72 @@
...
+            && !$factory[0] instanceof Reference
+        ) {
+            throw new LogicException('#[AutowireInline] attribute defines a $factory which does not contain a callable.');
```
nicolas-grekas: let's use `static::class` in error messages so that we can warn about the real attribute (child classes)

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
```diff
@@ -0,0 +1,72 @@
...
+        $definition = new Definition($this->class);
+        $configurator = (new InlineServiceConfigurator($definition))
+            ->args($this->params)
+            ->lazy($this->lazy);
+
+        if (null !== $this->factory) {
+            $configurator->factory($this->normalizeFactory($this->factory));
+        }
+
+        return AbstractConfigurator::processValue($configurator);
```
nicolas-grekas: We should use any class from the Configurator namespace. That's undesired indirection here.

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    private function normalizeFactory(null|string|array $factory): null|string|array
+    {
+        return \is_array($factory) ? $factory + [1 => '__invoke'] : $factory;
```
nicolas-grekas: isn't this already handled somewhere in a compiler pass?

src/Symfony/Component/DependencyInjection/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ CHANGELOG
  * Deprecate `ContainerAwareInterface` and `ContainerAwareTrait`, use dependency injection instead
  * Add `defined` env var processor that returns `true` for defined and neither null nor empty env vars
  * Add `#[AutowireLocator]` and `#[AutowireIterator]` attributes
+ * Add `#[AutowireInline]` attribute
```
nicolas-grekas: should go in a new 7.1 section

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
```diff
@@ -384,7 +384,7 @@ private function autowireMethod(\ReflectionFunctionAbstract $reflectionMethod, a
                     $type = ProxyHelper::exportType($parameter);
                     $type = $type ? sprintf('is type-hinted "%s"', preg_replace('/(^|[(|&])\\\\|^\?\\\\?/', '\1', $type)) : 'has no type-hint';
 
-                    throw new AutowiringFailedException($this->currentId, sprintf('Cannot autowire service "%s": argument "$%s" of method "%s()" %s, you should configure its value explicitly.', $this->currentId, $parameter->name, $class !== $this->currentId ? $class.'::'.$method : $method, $type));
+                    throw new AutowiringFailedException($this->currentId, sprintf('Cannot autowire service "%s": argument "$%s" of method "%s()" "%s", you should configure its value explicitly.', $this->currentId, $parameter->name, $class !== $this->currentId ? $class.'::'.$method : $method, $type));
```
nicolas-grekas: should be reverted, that's a false positive

nicolas-grekas (commented):
I'm not convinced by the interface, I'd prefer removing it.
I'm generally not convinced by interfaces for value objects and attributes are value objects.
Extending the class makes sense to me in this case.

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
DaDeather: I hope I got it right with what you meant there :see_no_evil:

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
DaDeather: I've removed this part and return the definition directly. I wasn't sure if I got you right here.

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
DaDeather: I'm not familiar with the process there. The only thing I can tell you is that if I'd remove this the tests in the `PhpDumperTest.php` would fail :shrug:

src/Symfony/Component/DependencyInjection/CHANGELOG.md
DaDeather: I've moved this to that section.

Should / must I rebase this feature onto 7.1?

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
DaDeather: I've removed my changes in regards of this.

DaDeather:
> I'm not convinced by the interface, I'd prefer removing it. I'm generally not convinced by interfaces for value objects and attributes are value objects. Extending the class makes sense to me in this case.

Alright :+1: I have removed the interface and based the `AutowireCallable` on the new `AutowireInline` as you proposed before.

src/Symfony/Component/DependencyInjection/CHANGELOG.md
nicolas-grekas: yes please, new features go on 7.1 now

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * @param class-string                                $class
+     * @param array                                       $params Static parameters to pass to instance creation
+     * @param list<array{0: string, 1?: array, 2?: bool}> $calls  A list of methods to be called after service initialization
+     * @param bool|class-string                           $lazy   Whether to use lazy-loading for this argument
+     */
+    public function __construct(
+        private readonly string $class,
+        private readonly null|string|array $factory = null,
+        private readonly array $params = [],
```
nicolas-grekas: args instead I suppose?

stof:
Please update the PR description to show usages of the feature. This makes it a lot easier for the documentation team (and also for reviewers)

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
DaDeather: Changed to `$args` :+1:

src/Symfony/Component/DependencyInjection/CHANGELOG.md
DaDeather: I've rebased this onto 7.1 thanks for letting me know :-)

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
```diff
@@ -0,0 +1,88 @@
...
+            }
+        }
+
+        parent::__construct($this->normalizeFactory($factory) ?? $class, lazy: $lazy);
+    }
+
+    public function buildDefinition(mixed $value, ?string $type, \ReflectionParameter $parameter): Definition
+    {
+        $definition = new Definition($this->class);
+        $configurator = (new InlineServiceConfigurator($definition))
```
nicolas-grekas: we don't want to use the configurator, let's play with $definition directly instead in this method

DaDeather: I've removed the configurator from the `AutowireInline` attribute.

DaDeather:
> Please update the PR description to show usages of the feature. This makes it a lot easier for the documentation team (and also for reviewers)

Done. Hope that fits your needs :smiley:

DaDeather:
@nicolas-grekas is there anything left to do here for me :-)?

Besides that I wasn't sure about the process:
Do I resolve the threads or do you once you check if it's done?

DaDeather:
> @nicolas-grekas is there anything left to do here for me :-)?
> 
> Besides that I wasn't sure about the process: Do I resolve the threads or do you once you check if it's done?

bump :shrug:

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
```diff
@@ -315,8 +315,9 @@ private function autowireMethod(\ReflectionFunctionAbstract $reflectionMethod, a
                         continue 2;
                     }
 
-                    if ($attribute instanceof AutowireCallable) {
+                    if ($attribute instanceof AutowireInline) {
                         $value = $attribute->buildDefinition($value, $type, $parameter);
+                        $value = $this->doProcessValue($value->setAutowired(true));
```
nicolas-grekas: This line allows autowiring services defined using `AutowireInline`

nicolas-grekas (commented):
I rebased and pushed a commit to go one step forward.
I didn't care about tests for now at least but I played with a small webapp skeleton instead.
This does NOT work because the added line in AutowirePass would require making the autowiring logic reentrant, which it's not at the moment: it's not possible to autowire a definition while already autowiring a definition because some related state is stored in properties. This needs to be improved before we can move forward here.

Status: needs work

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
```diff
@@ -315,8 +315,9 @@ private function autowireMethod(\ReflectionFunctionAbstract $reflectionMethod, a
                         continue 2;
                     }
 
-                    if ($attribute instanceof AutowireCallable) {
+                    if ($attribute instanceof AutowireInline) {
                         $value = $attribute->buildDefinition($value, $type, $parameter);
+                        //$value = $this->doProcessValue($value);
```
nicolas-grekas: Can be postponed for another PR (making AutowirePass reentrant)

src/Symfony/Component/DependencyInjection/Tests/Attribute/AutowireInlineTest.php
```diff
@@ -0,0 +1,190 @@
...
+ */
+
+namespace Symfony\Component\DependencyInjection\Tests\Attribute;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\DependencyInjection\Attribute\AutowireInline;
+use Symfony\Component\DependencyInjection\Exception\LogicException;
+use Symfony\Component\DependencyInjection\Reference;
+
+class AutowireInlineTest //extends TestCase
```
nicolas-grekas: @DaDeather I disabled your test case because it tested behaviors that I removed. Validation is now delegated to existing code infrastructure in YamlFileLoader (similarly to what we do in `RegisterAutoconfigureAttributesPass`)
Would you be able to update it to make it fit the new implem?

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
```diff
@@ -0,0 +1,53 @@
...
+ * @author Ismail Özgün Turan <oezguen.turan@dadadev.com>
+ */
+#[\Attribute(\Attribute::TARGET_PARAMETER)]
+class AutowireInline extends Autowire
+{
+    public function __construct(string|array $class, array $arguments = [], array $calls = [], array $properties = [], string $parent = null, bool|string $lazy = false)
+    {
+        parent::__construct([
+            \is_array($class) ? 'factory' : 'class' => $class,
+            'arguments' => $arguments,
```
nicolas-grekas, line 40: Drawback here: named arguments won't work the same as in other places (no "$" prefix here). We might want to run `ResolveNamedArgumentsPass` on the resulting definition. Can be added later on.

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
```diff
@@ -0,0 +1,53 @@
...
+            'calls' => $calls,
+            'properties' => $properties,
+            'parent' => $parent,
```
nicolas-grekas, line 43: Could be nice to support but won't work as is because `ResolveChildDefinitionPass` already ran when this is discovered.

nicolas-grekas (commented):
All my remaining concerns are related to the compilation process: discovering new service definitions in AutowirePass means we won't execute previously ran compiler passes on them. Not sure what to do with that.

Naming also: AutowireInline? AutowireService? AutowireDefinition? AutowireInstance? better?

nicolas-grekas:
> All my remaining concerns are related to the compilation process: discovering new service definitions in AutowirePass means we won't execute previously ran compiler passes on them.

We could register those definitions found in attributes in loaders: when a service is declared as autowired, we scan its class attributes in the loader for extra services, using some naming convention that can easily be matched by AutowirePass at its turn.

Thinking loud :)

src/Symfony/Component/DependencyInjection/Tests/Attribute/AutowireInlineTest.php
DaDeather: I adapted the changes you made in the tests here:
https://github.com/symfony/symfony/pull/52820/commits/8215026333c5071e9be4efc4792e49e90cef1eb7

DaDeather:
> > All my remaining concerns are related to the compilation process: discovering new service definitions in AutowirePass means we won't execute previously ran compiler passes on them.
> 
> We could register those definitions found in attributes in loaders: when a service is declared as autowired, we scan its class attributes in the loader for extra services, using some naming convention that can easily be matched by AutowirePass at its turn.
> 
> Thinking loud :)

I don't have that much knowledge in the compilation process of the dependency injection yet but if you'd point me in the direction I can do what's necessary to get this feature done / complete though :+1:.

Looking at your previous comments it seems to me that we may have to make some adjustments to the way the `AutowireInline` attribute is being processed.

So could you probably summarize for me if and what there is to do here so I can take over?

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
nicolas-grekas: I rebased the PR on top of #54047, which fixed this concern.

src/Symfony/Component/DependencyInjection/Tests/Attribute/AutowireInlineTest.php
```diff
@@ -0,0 +1,200 @@
...
+            [['someMethod', ['someArgument'], true]],
+            [['someMethod', ['someArgument'], true]],
+        ];
+        yield 'method with arguments with return clone false' => [
+            [['someMethod', ['someArgument'], false]],
+            [['someMethod', ['someArgument']]],
+        ];
+    }
+
+    private function createReflectionParameterMock()
```
nicolas-grekas: we'd better not use a mock here - mocks are risky so we try to avoid them when possible

src/Symfony/Component/DependencyInjection/Tests/Attribute/AutowireInlineTest.php
```diff
@@ -0,0 +1,200 @@
...
+    public static function provideFactories(): iterable
+    {
+        yield 'string callable' => [[null, 'someFunction'], [null, 'someFunction']];
+
+        yield 'class only' => [['someClass'], ['someClass', '__invoke']];
+        yield 'reference only' => [[new Reference('someClass')], [new Reference('someClass'), '__invoke']];
+
+        yield 'class with method' => [['someClass', 'someStaticMethod'], ['someClass', 'someStaticMethod']];
+        yield 'reference with method' => [[new Reference('someClass'), 'someMethod'], [new Reference('someClass'), 'someMethod']];
+    }
```
nicolas-grekas, line 144:
```suggestion
        yield '@reference with method' => [['@someClass', 'someMethod'], [new Reference('someClass'), 'someMethod']];
    }
```

src/Symfony/Component/DependencyInjection/Tests/Attribute/AutowireInlineTest.php
```diff
@@ -0,0 +1,200 @@
...
+    /**
+     * @dataProvider provideFactories
+     */
+    public function testFactoryAndParamsLazy(string|array $factory, string|array $expectedResult)
+    {
+        $attribute = new AutowireInline($factory, ['someParam'], lazy: true);
+
+        $buildDefinition = $attribute->buildDefinition($attribute->value, null, $this->createReflectionParameterMock());
+
+        self::assertSame(null, $buildDefinition->getClass());
```
nicolas-grekas: everywhere:
```suggestion
        self::assertNull($buildDefinition->getClass());
```

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
```diff
@@ -0,0 +1,53 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\DependencyInjection\Attribute;
+
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Loader\YamlFileLoader;
+
+/**
+ * Attribute to tell a parameter to inline a service with custom parameters.
```
nicolas-grekas:
```suggestion
 * Allows inline service definition at the class level.
```

src/Symfony/Component/DependencyInjection/Tests/Attribute/AutowireInlineTest.php
```diff
@@ -0,0 +1,200 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\DependencyInjection\Tests\Attribute;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\DependencyInjection\Attribute\AutowireInline;
+use Symfony\Component\DependencyInjection\Exception\LogicException;
+use Symfony\Component\DependencyInjection\Reference;
+use function PHPUnit\Framework\assertSame;
```
nicolas-grekas: should be removed

DaDeather: [Done here](https://github.com/symfony/symfony/compare/4b0e7643ef8444649c452c490fc6335e390d324f..5955b18f44df3374fd1eb423e025f0a0f04d956c)

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
DaDeather: [Done here](https://github.com/symfony/symfony/compare/4b0e7643ef8444649c452c490fc6335e390d324f..5955b18f44df3374fd1eb423e025f0a0f04d956c)

src/Symfony/Component/DependencyInjection/Tests/Attribute/AutowireInlineTest.php
DaDeather: [Done here](https://github.com/symfony/symfony/compare/4b0e7643ef8444649c452c490fc6335e390d324f..5955b18f44df3374fd1eb423e025f0a0f04d956c)

src/Symfony/Component/DependencyInjection/Tests/Attribute/AutowireInlineTest.php
DaDeather: [Done here](https://github.com/symfony/symfony/compare/4b0e7643ef8444649c452c490fc6335e390d324f..5955b18f44df3374fd1eb423e025f0a0f04d956c)

src/Symfony/Component/DependencyInjection/Tests/Attribute/AutowireInlineTest.php
DaDeather: [Added here](https://github.com/symfony/symfony/compare/5955b18f44df3374fd1eb423e025f0a0f04d956c..6a7e3fcd73b2ea87aac29d89823e212f97287899)

nicolas-grekas:
About my remaining concerns (`ResolveNamedArgumentsPass`, `ResolveChildDefinitionPass` and other pass not running), I'm wondering if we shouldn't add a new `ResolveAutowireInlineAttributesPass` that would scan autowired services for this attribute and would register definitions for each of them. Then, based on some naming convention (likely built on `Container::hash($definition)`), `AutowirePass` would just reference these definitions instead of calling `AutowireInline::buildDefinition()`.

WDYT? Up to give it a try?

DaDeather:
> About my remaining concerns (`ResolveNamedArgumentsPass`, `ResolveChildDefinitionPass` and other pass not running), I'm wondering if we shouldn't add a new `ResolveAutowireInlineAttributesPass` that would scan autowired services for this attribute and would register definitions for each of them. Then, based on some naming convention (likely built on `Container::hash($definition)`), `AutowirePass` would just reference these definitions instead of calling `AutowireInline::buildDefinition()`.
> 
> WDYT? Up to give it a try?

Sure!
Could you maybe give a few hints how / where to implement this correctly?
I'm struggling a bit with the input style and the expected outcome here and therefore maybe didn't understand what's really expected here.

Would you expect a definition be like this:
```php
class AutowireInlineAttributesBar {
    public function __construct(Foo $foo, string $someString)
    {
    }
}

class AutowireInlineAttributes
{
    public function __construct(
        #[AutowireInline(AutowireInlineAttributesBar::class, [
            '$foo' => Foo::class,
            '$someString' => 'testString',
        ])]
        public AutowireInlineAttributesBar $inlined,
    ) {
    }
}
```

Where the class `Foo` then would be a autowired as the provided service name and the `someString` injected as a string or what would you rather expect here to happen?

Besides that if I got you right I would have to create a class called `ResolveAutowireInlineAttributesPass` which then checks for a passed `Definition` to be `$definition->isAutowired() === true` and check for it to have any `AutowireInline` attributes to be processed (calling `->buildDefinition(...)` and registering those via `$this->container->setDefinition(ContainerBuilder::hash($autowireInlineAttributeInstance), $definition);`.

Is this assumption correct?

nicolas-grekas:
You've got it right I think :)

DaDeather:
> You've got it right I think :)

I've prepared something that doesn't quite work :shrug: as mentioned before maybe I'm missing something here. Could you maybe take a look so that we may correct the issues there or my misunderstanding :see_no_evil:?

I would be grateful :+1: https://github.com/symfony/symfony/pull/52820/commits/dd16390a5290a9a08b2c997fa9d959296bb6f0e0

@nicolas-grekas

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
```diff
@@ -320,9 +320,9 @@ private function autowireMethod(\ReflectionFunctionAbstract $reflectionMethod, a
                         continue 2;
                     }
 
-                    if ($attribute instanceof AutowireCallable) {
-                        $value = $attribute->buildDefinition($value, $type, $parameter);
-                        $value = $this->doProcessValue($value);
+                    if ($attribute instanceof AutowireInline) {
+                        $value = $this->container->getDefinition(ContainerBuilder::hash($attribute));
+                        $value = $this->processValue($value);
```
nicolas-grekas:
```suggestion
                        $value = $attribute->buildDefinition($value, $type, $parameter);
                        $value = new Reference('.autowire_inline.'.ContainerBuilder::hash($value));
```

src/Symfony/Component/DependencyInjection/Compiler/PassConfig.php
```diff
@@ -66,6 +66,7 @@ public function __construct()
             new ServiceLocatorTagPass(),
             new DecoratorServicePass(),
             new CheckDefinitionValidityPass(),
+            new ResolveAutowireInlineAttributesPass(),
```
nicolas-grekas: Any reason why you put that one here? I would register it just before ResolveChildDefinitionsPass, so that resulting definitions can have a parent.

src/Symfony/Component/DependencyInjection/Tests/Fixtures/includes/autowiring_classes_80.php
```diff
@@ -140,3 +141,32 @@ public function __construct(#[Lazy, Autowire(lazy: true)] A $a)
     {
     }
 }
+
+class AutowireInlineAttributesBar {
```
nicolas-grekas:
```suggestion
class AutowireInlineAttributesBar
{
```

src/Symfony/Component/DependencyInjection/Tests/Compiler/ResolveAutowireInlineAttributesPassTest.php
```diff
@@ -0,0 +1,49 @@
...
+
+        $container->register('autowire_inline1', AutowireInlineAttributes1::class)
+            ->setAutowired(true);
+
+        $container->register('autowire_inline2', AutowireInlineAttributes2::class)
+            ->setAutowired(true);
+
+        (new ResolveNamedArgumentsPass())->process($container);
+        (new ResolveClassPass())->process($container);
+        (new ResolveChildDefinitionsPass())->process($container);
```
nicolas-grekas: the new pass should be move before all these, isn't it?

src/Symfony/Component/DependencyInjection/Compiler/ResolveAutowireInlineAttributesPass.php
```diff
@@ -0,0 +1,67 @@
...
+            return $value;
+        }
+
+        if (!$reflectionClass = $this->container->getReflectionClass($value->getClass(), false)) {
+            $this->container->log($this, sprintf('Skipping service "%s": Class or interface "%s" cannot be loaded.', $this->currentId, $value->getClass()));
+
+            return $value;
+        }
+
+        $constructorReflectionMethod = $reflectionClass->getConstructor();
```
nicolas-grekas: this should call `$this->getConstructor($value, false)` instead, and we should also parse `$value->getMethodCalls()`
Basically we should visit the same parameters as we do in AutowirePass.

src/Symfony/Component/DependencyInjection/Compiler/ResolveAutowireInlineAttributesPass.php
```diff
@@ -0,0 +1,67 @@
...
+        if ($constructorReflectionMethod === null) {
+            return $value;
+        }
+
+        $reflectionParameters = $constructorReflectionMethod->getParameters();
+        foreach ($reflectionParameters as $reflectionParameter) {
+            $autowireInlineAttributes = $reflectionParameter->getAttributes(AutowireInline::class);
+            if ($autowireInlineAttributes === []) {
```
nicolas-grekas:
```suggestion
            if (!$autowireInlineAttributes = $reflectionParameter->getAttributes(AutowireInline::class, \ReflectionAttribute::IS_INSTANCEOF)) {
```

src/Symfony/Component/DependencyInjection/Compiler/ResolveAutowireInlineAttributesPass.php
```diff
@@ -0,0 +1,67 @@
...
+            }
+
+            foreach ($autowireInlineAttributes as $autowireInlineAttribute) {
+                /** @var AutowireInline $autowireInlineAttributeInstance */
+                $autowireInlineAttributeInstance = $autowireInlineAttribute->newInstance();
+
+                $type = ProxyHelper::exportType($reflectionParameter, true);
+                $definition = $autowireInlineAttributeInstance->buildDefinition($autowireInlineAttributeInstance->value, $type, $reflectionParameter);
+
+                $this->container->setDefinition(ContainerBuilder::hash($autowireInlineAttributeInstance), $definition);
```
nicolas-grekas:
```suggestion
                $this->container->setDefinition('.autowire_inline.'.ContainerBuilder::hash($definition), $definition);
```

nicolas-grekas (commented):
Thanks for the update, the path looks good to me.
Here are some comments to go to the next step.

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
```diff
@@ -0,0 +1,53 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\DependencyInjection\Attribute;
+
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Loader\YamlFileLoader;
+
+/**
+ * Allows inline service definition at the class level.
```
stof: this description seems confusing as the attribute is not applied at class level.

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
stof: instead of building a definition again here, shouldn't `ResolveAutowireInlineAttributesPass` inject a reference to the definition it created in the arguments (or maybe a binding depending on what works best) ?

src/Symfony/Component/DependencyInjection/Compiler/ResolveAutowireInlineAttributesPass.php
stof: This `if` can be removed entirely. Looping over an empty array is not an error.

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
nicolas-grekas: yes, that's what my code suggests, but first we need to build the definition so that we can get its hash

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @author Ismail Özgün Turan <oezguen.turan@dadadev.com>
+ */
+#[\Attribute(\Attribute::TARGET_PARAMETER)]
+class AutowireInline extends Autowire
```
chalasr, line 30: `AutowireInline` looks pretty cryptic, any newcomer wouldn't be able to get what the attribute is for by just looking at its name nor its description as it currently is. I think we either need to find a super self-explanatory name (I've no clue) or extend the description so that it tells what's the purpose of the attribute and when it should be used (inline service definition only means something to advanced Symfony's DIC hackers)

nicolas-grekas: Naming things... :)
"inline" refers to "inline_service()" in the PHP-DSL

AutowireInlineService? but verbose
AutowireNew? AutowireObject? AutowireInstance? or keep AutowireInline?

of course, a top notch description is also desirable

src/Symfony/Component/DependencyInjection/Compiler/PassConfig.php
DaDeather: No there was no specific reason to it.
I just thought putting it next to `AutowirePass` may have made sense since the order here doesn't matter anyway since it's `krsort` [here](https://github.com/symfony/symfony/blob/dd16390a5290a9a08b2c997fa9d959296bb6f0e0/src/Symfony/Component/DependencyInjection/Compiler/PassConfig.php#L177-L180) anyway :shrug:.

I've moved it now before `ResolveChildDefinitionsPass`.

src/Symfony/Component/DependencyInjection/Tests/Compiler/ResolveAutowireInlineAttributesPassTest.php
DaDeather: I thought since they are executed in a sorted way it may make sense to have them processed in the same order as the code is being executed in reality is this assumption wrong?

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
DaDeather: I added a better description [here](https://github.com/symfony/symfony/compare/dd16390a5290a9a08b2c997fa9d959296bb6f0e0..c94322321acafdbaf32044e58b213223645497ce) feel free to request further suggestments to it :+1:.

src/Symfony/Component/DependencyInjection/Compiler/ResolveAutowireInlineAttributesPass.php
DaDeather: Done [here](https://github.com/symfony/symfony/compare/dd16390a5290a9a08b2c997fa9d959296bb6f0e0..c94322321acafdbaf32044e58b213223645497ce)

src/Symfony/Component/DependencyInjection/Attribute/AutowireInline.php
DaDeather: Done [here](https://github.com/symfony/symfony/compare/c94322321acafdbaf32044e58b213223645497ce..0ad67afe94ac26fd51bb7d4a67792721c50ed9de)

src/Symfony/Component/DependencyInjection/Compiler/ResolveAutowireInlineAttributesPass.php
DaDeather: Done [here](https://github.com/symfony/symfony/compare/dd16390a5290a9a08b2c997fa9d959296bb6f0e0..c94322321acafdbaf32044e58b213223645497ce)

src/Symfony/Component/DependencyInjection/Compiler/ResolveAutowireInlineAttributesPass.php
DaDeather: Done [here](https://github.com/symfony/symfony/compare/dd16390a5290a9a08b2c997fa9d959296bb6f0e0..c94322321acafdbaf32044e58b213223645497ce)

src/Symfony/Component/DependencyInjection/Compiler/ResolveAutowireInlineAttributesPass.php
DaDeather: Done [here](https://github.com/symfony/symfony/compare/dd16390a5290a9a08b2c997fa9d959296bb6f0e0..c94322321acafdbaf32044e58b213223645497ce) but still not sure if this was now done correctly by me. Do I have to process the parameters there as well similar to what `AutowirePass` does internally with the `methodCalls` and so on?

src/Symfony/Component/DependencyInjection/Tests/Compiler/ResolveAutowireInlineAttributesPassTest.php
DaDeather: Besides that I'm still running into test issues there though :disappointed: not sure what I'm missing in general.

src/Symfony/Component/DependencyInjection/Tests/Fixtures/includes/autowiring_classes_80.php
DaDeather: Done [here](https://github.com/symfony/symfony/compare/dd16390a5290a9a08b2c997fa9d959296bb6f0e0..c94322321acafdbaf32044e58b213223645497ce)

src/Symfony/Component/DependencyInjection/Compiler/PassConfig.php
DaDeather: Done [here](https://github.com/symfony/symfony/compare/dd16390a5290a9a08b2c997fa9d959296bb6f0e0..c94322321acafdbaf32044e58b213223645497ce)

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
DaDeather: Done [here](https://github.com/symfony/symfony/compare/dd16390a5290a9a08b2c997fa9d959296bb6f0e0..c94322321acafdbaf32044e58b213223645497ce)

nicolas-grekas (approved):
PR is ready.
I added a second commit to make the implementation a bit more generic.
Reviews welcome :pray:

src/Symfony/Component/DependencyInjection/ContainerBuilder.php
```diff
@@ -494,7 +494,9 @@ public function removeDefinition(string $id): void
     {
         if (isset($this->definitions[$id])) {
             unset($this->definitions[$id]);
-            $this->removedIds[$id] = true;
+            if ('.' !== ($id[0] ?? '-')) {
```
nicolas-grekas, line 497: I added this rule so that we don't use internal service ids in error messages. This matters to this PR because the added code creates new removed-ids, which means fixtures have to be updated with (what I consider as) noise.

fabpot:
Thank you @DaDeather.

PR #52824, opened 2023-11-30
[FrenchInflector] Add New Special Cases

meliben:
Add new special cases of words to the FrenchInflector

| Q             | A
| ------------- | ---
| Branch?       | 7.1 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | none
| License       | MIT

This PR adds to the FrenchInflector component a list of new special cases of french words : 

- words finishing with "al" in their singular form and finishing with "aux" in their plural form, so the singularize() method can take them into account
- words which should not be inflected


src/Symfony/Component/String/Inflector/FrenchInflector.php
```diff
@@ -80,7 +80,7 @@ final class FrenchInflector implements InflectorInterface
 
         // Words finishing with "al" are pluralized with a "aux" expected
         // Les mots finissant en "al" se terminent en "aux" sauf
-        ['/(amir|anim|arsen|boc|can|capit|capor|chev|crist|génér|hopit|hôpit|idé|journ|littor|loc|m|mét|minér|princip|radic|termin)aux$/i', '\1al'],
+        ['/(amir|anim|arsen|boc|can|capit|capor|chev|crist|génér|hopit|hôpit|idé|journ|littor|loc|m|mét|minér|princip|radic|termin|sign|médiév|flor|origin|géni)aux$/i', '\1al'],
```
smnandre, line 83: I think we should maintain the alphabetic order there

src/Symfony/Component/String/Inflector/FrenchInflector.php
```diff
@@ -108,7 +108,7 @@ final class FrenchInflector implements InflectorInterface
...
+    private const UNINFLECTED = '/^(abcès|accès|abus|albatros|anchois|anglais|autobus|bois|brebis|carquois|cas|chas|colis|concours|corps|cours|cyprès|décès|devis|discours|dos|embarras|engrais|entrelacs|excès|fils|fois|gâchis|gars|glas|héros|intrus|jars|jus|kermès|lacis|legs|lilas|marais|mars|matelas|mépris|mets|mois|mors|obus|os|palais|paradis|parcours|pardessus|pays|plusieurs|poids|pois|pouls|printemps|processus|progrès|puits|pus|rabais|radis|recors|recours|refus|relais|remords|remous|rictus|rhinocéros|repas|rubis|sans|sas|secours|sens|souris|succès|talus|tapis|tas|taudis|temps|tiers|univers|velours|verglas|vernis|virus|albanais|angolais|bangladais|bostwanais|burundais|camerounais|congolais|finlandais|français|gabonais|guyanais|irlandais|islandais|japonais|libanais|malais|maltais|martiniquais|népalais|néo-zélandais|ougandais|pakistanais|néerlandais|polonais|portugais|réunionnais|rwandais|sénégalais|sierra-léonais|soudanais|sri lankais|taïwanais|thaïlandais|togolais|québécois|chinois|suèdois|hongrois|danois|luxembourgeois|béninois|seychellois)$/i';
```
smnandre, line 111: (Typographic note: if we list only nouns, those one should be capitalized.)

My problem if we add the geographic nouns that come from geographics places (and derived from the related adjectives) is ... where do we stop ? 

Why some country and not others ? Why not regions, states, big cities ?

smnandre: https://fr.wiktionary.org/wiki/Cat%C3%A9gorie:Mots_en_fran%C3%A7ais_suffix%C3%A9s_avec_-ais

meliben: Hi @smnandre 

My proposal only concerned adjectives, but what I understand is that nouns will also match because of the case insensitive option of the regex, right ?

This proposal was only regarding countries to keep the FrenchInflector capabilities to a more global geographic level. I've selected all the countries I found for which the adjective does not vary from singular to plural in French, but I may have missed some. Do you have some in mind ?

Concerning nouns or adjectives that come from smaller geographic levels, I preferred avoiding it to keep the regex performant, but we could indeed add them. But then I have the same interrogation than you: where do we stop ?

smnandre: But this class does not handle adjectives (and that's a good thing because the concept of singular/plural for adjectives is really locale dependant)

```
/**
 * French inflector.
 *
 * This class does only inflect nouns; not adjectives nor composed words like "soixante-dix".
 */
final class FrenchInflector implements InflectorInterface
{
```

And i realize your pluralize exceptions are only adjectives, right ? 


--- 


Concerning the coutry lists, in a quick look i got : 

**Not a country:** 
- guyanais
- quebecois
- réunionnais
- ... and some other but i'm not in a mood for such debate haha

Missing:
- Azerbaïdjanais
- Bhoutanais
- Dominiquais
- Écossais
- Kazakhstanais
- Nord-Irlandais
- ... and some there too are up to a debate i will not have

fabpot:
Closing as we're only dealing with nouns, not adjectives.

PR #52831, opened 2023-11-30
[Routing] Revert " Fix removing aliases pointing to removed route in RouteCollection::remove()"

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | https://github.com/symfony/symfony/pull/52806#issuecomment-1832397866
| License       | MIT

We need to find another solution that doesn't change the behavior on 5.4 :confused:

### DIFF
--- a/src/Symfony/Component/Routing/RouteCollection.php
+++ b/src/Symfony/Component/Routing/RouteCollection.php
@@ -157,15 +157,8 @@ class RouteCollection implements \IteratorAggregate, \Countable
      */
     public function remove($name)
     {
-        $names = (array) $name;
-        foreach ($names as $n) {
-            unset($this->routes[$n], $this->priorities[$n]);
-        }
-
-        foreach ($this->aliases as $k => $alias) {
-            if (\in_array($alias->getId(), $names, true)) {
-                unset($this->aliases[$k]);
-            }
+        foreach ((array) $name as $n) {
+            unset($this->routes[$n], $this->priorities[$n], $this->aliases[$n]);
         }
     }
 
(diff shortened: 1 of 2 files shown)


xabbuh (approved):

src/Symfony/Component/Routing/RouteCollection.php
```diff
@@ -157,15 +157,8 @@ public function get(string $name)
...
-        foreach ($names as $n) {
-            unset($this->routes[$n], $this->priorities[$n]);
-        }
-
-        foreach ($this->aliases as $k => $alias) {
-            if (\in_array($alias->getId(), $names, true)) {
-                unset($this->aliases[$k]);
-            }
+        foreach ((array) $name as $n) {
+            unset($this->routes[$n], $this->priorities[$n], $this->aliases[$n]);
```
HypeMC, line 161: Maybe I'm missing something here, but why not have this method remove both an alias, if it's an alias, or a route with all its aliases, if it's a route?

nicolas-grekas: I also wondered the same. That'd make sense to me.

mkrauser:
Another thing about this solution: 
The Aliases will be removed when the Route itself is removed within the PrefixTrait or HostTrait. But the alias is never re-added. 

So when someone uses FQCN for their routing and they add Hosts or Prefixes, that will not work anymore. 
Or am I missing something?

nicolas-grekas:
Closing in favor of #52845

PR #52833, opened 2023-11-30
[Security][SecurityBundle] Add `LogoutUrlGeneratorInterface` and autowiring alias

GromNaN:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | https://github.com/symfony/demo/pull/1461#discussion_r1410550115
| License       | MIT

The `LogoutUrlGenerator` is useful to be used in a controller as seen in the demo: https://github.com/symfony/demo/blob/72040e73b3deed4281706bb5dda1ff1321f00396/src/Controller/UserController.php#L65
I'm not sure about our strategy for autowiring bundle-provided services. In this case, this service can't be redefined with a different class as it doesn't implement any interface.

### DIFF
--- a/src/Symfony/Bridge/Twig/Extension/LogoutUrlExtension.php
+++ b/src/Symfony/Bridge/Twig/Extension/LogoutUrlExtension.php
@@ -12,6 +12,7 @@
 namespace Symfony\Bridge\Twig\Extension;
 
 use Symfony\Component\Security\Http\Logout\LogoutUrlGenerator;
+use Symfony\Component\Security\Http\Logout\LogoutUrlGeneratorInterface;
 use Twig\Extension\AbstractExtension;
 use Twig\TwigFunction;
 
@@ -22,9 +23,9 @@ use Twig\TwigFunction;
  */
 final class LogoutUrlExtension extends AbstractExtension
 {
-    private LogoutUrlGenerator $generator;
+    private LogoutUrlGenerator|LogoutUrlGeneratorInterface $generator;
 
-    public function __construct(LogoutUrlGenerator $generator)
+    public function __construct(LogoutUrlGenerator|LogoutUrlGeneratorInterface $generator)
     {
         $this->generator = $generator;
     }
--- a/src/Symfony/Bundle/SecurityBundle/DataCollector/SecurityDataCollector.php
+++ b/src/Symfony/Bundle/SecurityBundle/DataCollector/SecurityDataCollector.php
@@ -26,6 +26,7 @@ use Symfony\Component\Security\Core\Role\RoleHierarchyInterface;
 use Symfony\Component\Security\Http\Firewall\SwitchUserListener;
 use Symfony\Component\Security\Http\FirewallMapInterface;
 use Symfony\Component\Security\Http\Logout\LogoutUrlGenerator;
+use Symfony\Component\Security\Http\Logout\LogoutUrlGeneratorInterface;
 use Symfony\Component\VarDumper\Caster\ClassStub;
 use Symfony\Component\VarDumper\Cloner\Data;
 
@@ -38,13 +39,13 @@ class SecurityDataCollector extends DataCollector implements LateDataCollectorIn
 {
     private ?TokenStorageInterface $tokenStorage;
     private ?RoleHierarchyInterface $roleHierarchy;
-    private ?LogoutUrlGenerator $logoutUrlGenerator;
+    private LogoutUrlGenerator|LogoutUrlGeneratorInterface|null $logoutUrlGenerator;
     private ?AccessDecisionManagerInterface $accessDecisionManager;
     private ?FirewallMapInterface $firewallMap;
     private ?TraceableFirewallListener $firewall;
     private bool $hasVarDumper;
 
-    public function __construct(TokenStorageInterface $tokenStorage = null, RoleHierarchyInterface $roleHierarchy = null, LogoutUrlGenerator $logoutUrlGenerator = null, AccessDecisionManagerInterface $accessDecisionManager = null, FirewallMapInterface $firewallMap = null, TraceableFirewallListener $firewall = null)
+    public function __construct(TokenStorageInterface $tokenStorage = null, RoleHierarchyInterface $roleHierarchy = null, LogoutUrlGenerator|LogoutUrlGeneratorInterface $logoutUrlGenerator = null, AccessDecisionManagerInterface $accessDecisionManager = null, FirewallMapInterface $firewallMap = null, TraceableFirewallListener $firewall = null)
     {
         $this->tokenStorage = $tokenStorage;
         $this->roleHierarchy = $roleHierarchy;
--- a/src/Symfony/Component/Security/Http/Logout/LogoutUrlGeneratorInterface.php
+++ b/src/Symfony/Component/Security/Http/Logout/LogoutUrlGeneratorInterface.php
@@ -1,0 +1,32 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Security\Http\Logout;
+
+/**
+ * Generate the absolute logout path and URL for a firewall.
+ */
+interface LogoutUrlGeneratorInterface
+{
+    /**
+     * Generates the absolute logout path for the firewall.
+     *
+     * @param string|null $key The firewall name (null to generate a path for the current firewall)
+     */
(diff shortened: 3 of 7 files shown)


yceruto (approved):

nicolas-grekas:
It might be worth adding an interface for that because `LogoutUrlGenerator` has many methods that we shouldn't be used as part of a regular public API (setters, etc).

src/Symfony/Bridge/Twig/Extension/LogoutUrlExtension.php
```diff
@@ -22,9 +22,9 @@
  */
 final class LogoutUrlExtension extends AbstractExtension
 {
-    private LogoutUrlGenerator $generator;
+    private LogoutUrlGeneratorInterface $generator;
 
-    public function __construct(LogoutUrlGenerator $generator)
+    public function __construct(LogoutUrlGeneratorInterface $generator)
```
nicolas-grekas: To keep compat with 6.4?
```suggestion
    public function __construct(LogoutUrlGeneratorInterface|LogoutUrlGenerator $generator)
```

GromNaN: Good idea.

src/Symfony/Component/Security/Http/Logout/LogoutUrlGeneratorInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Security\Http\Logout;
+
+interface LogoutUrlGeneratorInterface
```
nicolas-grekas, line 17: let's add a descriptive docblock so that debug:autowiring can display something useful about this interface

src/Symfony/Bundle/SecurityBundle/Resources/config/templating_twig.php
```diff
@@ -13,12 +13,13 @@
...
 use Symfony\Bridge\Twig\Extension\LogoutUrlExtension;
 use Symfony\Bridge\Twig\Extension\SecurityExtension;
+use Symfony\Component\Security\Http\Logout\LogoutUrlGeneratorInterface;
 
 return static function (ContainerConfigurator $container) {
     $container->services()
         ->set('twig.extension.logout_url', LogoutUrlExtension::class)
             ->args([
-                service('security.logout_url_generator'),
+                service(LogoutUrlGeneratorInterface::class),
```
nicolas-grekas: why change? we don't reference autowiring aliases usually

src/Symfony/Bundle/SecurityBundle/Resources/config/collectors.php
```diff
@@ -12,14 +12,15 @@
...
+use Symfony\Component\Security\Http\Logout\LogoutUrlGeneratorInterface;
 
 return static function (ContainerConfigurator $container) {
     $container->services()
         ->set('data_collector.security', SecurityDataCollector::class)
             ->args([
                 service('security.untracked_token_storage'),
                 service('security.role_hierarchy'),
-                service('security.logout_url_generator'),
+                service(LogoutUrlGeneratorInterface::class),
```
nicolas-grekas: should be reverted

src/Symfony/Bundle/SecurityBundle/DataCollector/SecurityDataCollector.php
```diff
@@ -38,13 +38,13 @@ class SecurityDataCollector extends DataCollector implements LateDataCollectorIn
...
     private ?RoleHierarchyInterface $roleHierarchy;
-    private ?LogoutUrlGenerator $logoutUrlGenerator;
+    private ?LogoutUrlGeneratorInterface $logoutUrlGenerator;
     private ?AccessDecisionManagerInterface $accessDecisionManager;
     private ?FirewallMapInterface $firewallMap;
     private ?TraceableFirewallListener $firewall;
     private bool $hasVarDumper;
 
-    public function __construct(TokenStorageInterface $tokenStorage = null, RoleHierarchyInterface $roleHierarchy = null, LogoutUrlGenerator $logoutUrlGenerator = null, AccessDecisionManagerInterface $accessDecisionManager = null, FirewallMapInterface $firewallMap = null, TraceableFirewallListener $firewall = null)
+    public function __construct(TokenStorageInterface $tokenStorage = null, RoleHierarchyInterface $roleHierarchy = null, LogoutUrlGeneratorInterface $logoutUrlGenerator = null, AccessDecisionManagerInterface $accessDecisionManager = null, FirewallMapInterface $firewallMap = null, TraceableFirewallListener $firewall = null)
```
nicolas-grekas: the union will help here also

src/Symfony/Bundle/SecurityBundle/composer.json
```diff
@@ -28,7 +28,7 @@
         "symfony/password-hasher": "^6.4|^7.0",
         "symfony/security-core": "^6.4|^7.0",
         "symfony/security-csrf": "^6.4|^7.0",
-        "symfony/security-http": "^6.4|^7.0",
+        "symfony/security-http": "^7.1",
```
nicolas-grekas: do we really need this? I'm not sure

src/Symfony/Bundle/SecurityBundle/Resources/config/templating_twig.php
GromNaN: It allows to change the implementation of the interface without replacing the service. `security.logout_url_generator` must be an instance of the class `LogoutUrlGenerator` for the `FirewallListener`. Same for the data collector service.

I don't have an actual use-case in mind.

nicolas-grekas: this should be reverted, we don't do this usually

src/Symfony/Bundle/SecurityBundle/composer.json
GromNaN: If this version constraint is not updated to 7.1, we have to detect the version in order to create the service alias or not.

GromNaN: Maybe I'm wrong, since the interface don't have to exist for it's name to be used as a service name.

nicolas-grekas: why? the alias is just a string, we don't need the interface to be here to have it work

GromNaN: I've reverted the constraint changes.

src/Symfony/Bundle/SecurityBundle/Resources/config/templating_twig.php
GromNaN: Reverted

src/Symfony/Component/Security/Http/Logout/LogoutUrlGeneratorInterface.php
GromNaN: Done

src/Symfony/Bundle/SecurityBundle/Resources/config/collectors.php
```diff
@@ -12,6 +12,7 @@
 namespace Symfony\Component\DependencyInjection\Loader\Configurator;
 
 use Symfony\Bundle\SecurityBundle\DataCollector\SecurityDataCollector;
+use Symfony\Component\Security\Http\Logout\LogoutUrlGeneratorInterface;
```
nicolas-grekas: can be removed

src/Symfony/Bundle/SecurityBundle/Resources/config/templating_twig.php
```diff
@@ -13,6 +13,7 @@
 
 use Symfony\Bridge\Twig\Extension\LogoutUrlExtension;
 use Symfony\Bridge\Twig\Extension\SecurityExtension;
+use Symfony\Component\Security\Http\Logout\LogoutUrlGeneratorInterface;
```
nicolas-grekas: drop :)

nicolas-grekas (approved):

chalasr:
Should we push for `Security::logout()` instead and use it in the demo? And then do we really need to make that generator a first-class service?

GromNaN:
> Should we push for `Security::logout()` instead and use it in the demo? And then do we really need to make that generator a first-class service?

You're right, that would be much better. Logout URLs are meant to generate links (in views), and shouldn't be used as redirects. https://github.com/symfony/demo/pull/1467

GromNaN:
Closing since the security service should be used.

PR #49070, opened 2023-01-23
[RateLimiter] fix incorrect retryAfter of FixedWindow

RobertMe:
| Q             | A
| ------------- | ---
| Branch?       |5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

A FixedWindow always ends `intervalInSeconds` after the start time (`timer`). So when calculating the time to consume some tokens the tokens will always be available at `timer + intervalInSeconds`. But currently this is reported incorrectly as `calculateTimeForTokens()` calculates the time based on the desired amount of tokens (and cycles) while it doesn't take into account `maxSize` amount of tokens become available at the windows end.

Furthermore calculating the amount of needed cycles is incorrect. This as all tokens become available at once (at the windows end) and you can't consume more tokens than `maxSize` (which is validated at the start of `FixedWindowLimiter::reserve`, in case of `tokens > limit` it throws).

**Note:** I don't think that changing the signature of `calculateTimeForTokens` is a deprecation. This as the `Window` class is marked as `@internal`. So it should only be used by the `RateLimiter` component.

PR #46110, opened 2022-04-19
[RateLimiter][Security] Improve performance of login/request rate limiter

Seldaek:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Tickets       | Fix #40371 
| License       | MIT
| Doc PR        | symfony/symfony-docs#... 

The login rate limiter currently works this way:

- consume 1 token on auth start (this READ + WRITES to cache backend / rate limiter storage), if this fails we fail the request as it exceeds the rate limit
- if auth succeeded, clear the rate limit (this WRITES again)]
- if auth failed, nothing further happens as we already consumed

So the happy path which is the most common (login succeeds) looks something like that on redis, as there are two rate limiter (global per IP and local per username+IP combo):

```
step1

1650362918.653657 [3 127.0.0.1:41584] "MGET" "lIoZODSNyI:921c2aa99364ab8bbb0e4525c4297aa71c61d2c6"
1650362918.654027 [3 127.0.0.1:41584] "MGET" "lIoZODSNyI:921c2aa99364ab8bbb0e4525c4297aa71c61d2c6"
1650362918.654294 [3 127.0.0.1:41584] "SETEX" "lIoZODSNyI:921c2aa99364ab8bbb0e4525c4297aa71c61d2c6" "119" "\x00\x00\x00\x02\x172Symfony\\Component\\RateLimiter\\Policy\\SlidingWindow\x14\x01\x11\"\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00<login_per_ip-127.0.0.1\x0cA\xd8\x97\xa2\x98\xa9\xd8\xcf"
1650362918.654651 [3 127.0.0.1:41584] "MGET" "lIoZODSNyI:d97b05c8178d81ea0e2009445cb6465d5fcffa39"
1650362918.654877 [3 127.0.0.1:41584] "MGET" "lIoZODSNyI:d97b05c8178d81ea0e2009445cb6465d5fcffa39"
[...]

PR #52835, opened 2023-11-30
[RateLimiter] Fix `RateLimit->getRetryAfter()` return value when consuming `0` or last tokens.

ERuban:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| License       | MIT

Have got some BC after updating the project to Symfony 6.4 (after that PR https://github.com/symfony/symfony/pull/51676).

Sometimes I need to get `RateLimit` object without consuming just before consuming a try, in example:
```
$rateLimit = $limiter->consume(0);
if ($rateLimit->getRemainingTokens() === 0) {
   throw new SomeException($rateLimit->getRetryAfter());
}
...
$limiter->consume(1)
...
```
and found that in that case `$rateLimit->getRetryAfter()` always returns `now`.

So this PR fixes it.

**UPDATE**

Have found that `RateLimit->getRetryAfter()` returns `now` when the last token was consumed for `sliding_window` and `fixed_window`. So fixed it too.

### DIFF
--- a/src/Symfony/Component/RateLimiter/Tests/Policy/SlidingWindowLimiterTest.php
+++ b/src/Symfony/Component/RateLimiter/Tests/Policy/SlidingWindowLimiterTest.php
@@ -114,6 +114,9 @@ class SlidingWindowLimiterTest extends TestCase
 
         // 2 over the limit, causing the WaitDuration to become 2/10th of the 12s interval
         $this->assertEqualsWithDelta(12 / 5, $limiter->reserve(4)->getWaitDuration(), 1);
+
+        $limiter->reset();
+        $this->assertEquals(0, $limiter->reserve(10)->getWaitDuration());
     }
 
     private function createLimiter(): SlidingWindowLimiter
--- a/src/Symfony/Component/RateLimiter/Tests/Policy/SlidingWindowLimiterTest.php
+++ b/src/Symfony/Component/RateLimiter/Tests/Policy/SlidingWindowLimiterTest.php
@@ -54,6 +54,21 @@ class SlidingWindowLimiterTest extends TestCase
         $this->assertSame(10, $rateLimit->getLimit());
     }
 
+    public function testConsumeLastToken()
+    {
+        $limiter = $this->createLimiter();
+        $limiter->reset();
+        $limiter->consume(9);
+
+        $rateLimit = $limiter->consume(1);
+        $this->assertSame(0, $rateLimit->getRemainingTokens());
+        $this->assertTrue($rateLimit->isAccepted());
+        $this->assertEquals(
+            \DateTimeImmutable::createFromFormat('U', (string) floor(microtime(true) + 12 / 10)),
+            $rateLimit->getRetryAfter()
+        );
+    }
+
     public function testConsumeZeroTokens()
     {
         $limiter = $this->createLimiter();
@@ -70,7 +85,7 @@ class SlidingWindowLimiterTest extends TestCase
         $limiter->consume(10);
 
         $rateLimit = $limiter->consume(0);
-        $this->assertFalse($rateLimit->isAccepted());
+        $this->assertTrue($rateLimit->isAccepted());
         $this->assertEquals(
             \DateTimeImmutable::createFromFormat('U', (string) floor(microtime(true) + 12 / 10)),
             $rateLimit->getRetryAfter()
--- a/src/Symfony/Component/RateLimiter/Tests/Policy/SlidingWindowLimiterTest.php
+++ b/src/Symfony/Component/RateLimiter/Tests/Policy/SlidingWindowLimiterTest.php
@@ -59,7 +59,7 @@ class SlidingWindowLimiterTest extends TestCase
         $rateLimit = $limiter->consume(0);
         $this->assertTrue($rateLimit->isAccepted());
         $this->assertEquals(
-            \DateTimeImmutable::createFromFormat('U', (string)floor(microtime(true))),
+            \DateTimeImmutable::createFromFormat('U', (string) floor(microtime(true))),
             $rateLimit->getRetryAfter()
         );
 
@@ -69,7 +69,7 @@ class SlidingWindowLimiterTest extends TestCase
         $rateLimit = $limiter->consume(0);
         $this->assertFalse($rateLimit->isAccepted());
         $this->assertEquals(
-            \DateTimeImmutable::createFromFormat('U', (string)floor(microtime(true) + 12 / 10)),
+            \DateTimeImmutable::createFromFormat('U', (string) floor(microtime(true) + 12 / 10)),
             $rateLimit->getRetryAfter()
         );
     }
--- a/src/Symfony/Component/RateLimiter/Tests/Policy/SlidingWindowLimiterTest.php
+++ b/src/Symfony/Component/RateLimiter/Tests/Policy/SlidingWindowLimiterTest.php
@@ -52,6 +52,26 @@ class SlidingWindowLimiterTest extends TestCase
         $rateLimit = $limiter->consume(10);
         $this->assertTrue($rateLimit->isAccepted());
         $this->assertSame(10, $rateLimit->getLimit());
+
+        // Test without consuming a try
+        $limiter->reset();
+
+        $rateLimit = $limiter->consume(0);
+        $this->assertTrue($rateLimit->isAccepted());
+        $this->assertEquals(
+            \DateTimeImmutable::createFromFormat('U', (string)floor(microtime(true))),
+            $rateLimit->getRetryAfter()
+        );
+
+        $limiter->reset();
+        $limiter->consume(10);
+
+        $rateLimit = $limiter->consume(0);
+        $this->assertFalse($rateLimit->isAccepted());
+        $this->assertEquals(
+            \DateTimeImmutable::createFromFormat('U', (string)floor(microtime(true) + 12 / 10)),
+            $rateLimit->getRetryAfter()
+        );
     }
 
     public function testWaitIntervalOnConsumeOverLimit()
--- a/src/Symfony/Component/RateLimiter/Tests/Policy/SlidingWindowLimiterTest.php
+++ b/src/Symfony/Component/RateLimiter/Tests/Policy/SlidingWindowLimiterTest.php
@@ -52,8 +52,11 @@ class SlidingWindowLimiterTest extends TestCase
         $rateLimit = $limiter->consume(10);
         $this->assertTrue($rateLimit->isAccepted());
         $this->assertSame(10, $rateLimit->getLimit());
+    }
 
-        // Test without consuming a try
+    public function testConsumeZeroTokens()
+    {
+        $limiter = $this->createLimiter();
         $limiter->reset();
 
         $rateLimit = $limiter->consume(0);
--- a/src/Symfony/Component/RateLimiter/Policy/FixedWindowLimiter.php
+++ b/src/Symfony/Component/RateLimiter/Policy/FixedWindowLimiter.php
@@ -59,21 +59,33 @@ final class FixedWindowLimiter implements LimiterInterface
             $now = microtime(true);
             $availableTokens = $window->getAvailableTokens($now);
 
-            if ($availableTokens >= max(1, $tokens)) {
+            if (0 !== $tokens && $availableTokens > $tokens) {
                 $window->add($tokens, $now);
 
                 $reservation = new Reservation($now, new RateLimit($window->getAvailableTokens($now), \DateTimeImmutable::createFromFormat('U', floor($now)), true, $this->limit));
             } else {
+                if ($availableTokens === $tokens) {
+                    $window->add($tokens, $now);
+                }
+
                 $waitDuration = $window->calculateTimeForTokens(max(1, $tokens), $now);
 
-                if (null !== $maxTime && $waitDuration > $maxTime) {
+                if ($availableTokens !== $tokens && 0 !== $tokens && null !== $maxTime && $waitDuration > $maxTime) {
                     // process needs to wait longer than set interval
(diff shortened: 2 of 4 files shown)


derrabus:
Thank you for your contribution. Please not that we don't merge any changes without a test that covers it.

ERuban:
Hey @derrabus,
I'm not sure how and where to test that case. Could you help me a bit?

derrabus:
> Hey @derrabus, I'm not sure how and where to test that case. Could you help me a bit?

Well, you're changing `SlidingWindowLimiter`, so `SlidingWindowLimiterTest` sounds like a good place to add a test, doesn't it.

ERuban:
@derrabus, done

src/Symfony/Component/RateLimiter/Tests/Policy/SlidingWindowLimiterTest.php
```diff
@@ -52,6 +52,26 @@ public function testConsume()
...
+
+        $limiter->reset();
+        $limiter->consume(10);
+
+        $rateLimit = $limiter->consume(0);
+        $this->assertFalse($rateLimit->isAccepted());
+        $this->assertEquals(
+            \DateTimeImmutable::createFromFormat('U', (string) floor(microtime(true) + 12 / 10)),
+            $rateLimit->getRetryAfter()
+        );
```
derrabus: This should be a separate test.

ERuban: done

derrabus:
@wouterj Can you have a look at this? Consuming a rate limiter without actually consuming anything feels like a hack. Is this a scenario we support? I understand that it simply has somehow worked previously.

ERuban:
> @wouterj Can you have a look at this? Consuming a rate limiter without actually consuming anything feels like a hack. Is this a scenario we support? I understand that it simply has somehow worked previously.

Yeah, otherwise we need some another way to get current RateLimit.
Just imagine you need to check available tokens somewhere in another class.

ERuban:
And, yes, using it before I got the wrong time to retry, it was the time when all tokens will be available, not the first one. With that fix I got the time for the next available token now.

RobertMe:
@derrabus 
> @wouterj Can you have a look at this? Consuming a rate limiter without actually consuming anything feels like a hack. Is this a scenario we support? I understand that it simply has somehow worked previously.

This is used in the framework itself too:
https://github.com/symfony/symfony/blob/8cd61c542a7e242d2f1963449b993cfb75c63bf9/src/Symfony/Component/HttpFoundation/RateLimiter/AbstractRequestRateLimiter.php#L32-L46
This allows checking whether a token can be consumed. In this case it's used in the `LoginThrottlingListener` which checks if a token is available in the `CheckPassportEvent` listener, if there are none available (as indicated by `peek` / `consume(0)`) the login will be denied. If it's allowed (still consumable) it will only be actually consumed when the login fails. That's at least one use case where "peeking" / "consume-ing without actually consuming" is used.

ERuban:
> @derrabus
> 
> > @wouterj Can you have a look at this? Consuming a rate limiter without actually consuming anything feels like a hack. Is this a scenario we support? I understand that it simply has somehow worked previously.
> 
> This is used in the framework itself too: https://github.com/symfony/symfony/blob/8cd61c542a7e242d2f1963449b993cfb75c63bf9/src/Symfony/Component/HttpFoundation/RateLimiter/AbstractRequestRateLimiter.php#L34C1-L46C53 This allows checking whether a token can be consumed. In this case it's used in the `LoginThrottlingListener` which checks if a token is available in the `CheckPassportEvent` listener, if there are none available (as indicated by `peek` / `consume(0)`) the login will be denied. If it's allowed (still consumable) it will only be actually consumed when the login fails. That's at least one use case where "peeking" / "consume-ing without actually consuming" is used.

Nice example, thank you.

So then, what about to implement smth like `LimiterInterface::peek()` to get the `RateLimit` object partly using the logic from `LimiterInterface::reserve()` (where we instantiating `RateLimit`). We should handle `$tokens === 0` in `LimeterInterface::consume()` and `LimeterInterface::reserve()` then.

I can try to do it then or update that PR.  
(Anyway we have a BC break here, bcs seems that the old `RateLimit->getRetryAfter()` returns the time when all tokens will be available (if there is no available tokens), and the new one will return the time when the new one will be available)

what do you think?

RobertMe:
> So then, what about to implement smth like `LimiterInterface::peek()` to get the `RateLimit` object partly using the logic from `LimiterInterface::reserve()`

That would be a BC break (you can't change the interface). And doing it in a backwards compatible manner would probably mean it becomes a feature, not a bugfix, so would end up in 7.1 earliest.
And personally (but I'm not a Symfony dev, just a users) I don't really like the idea as the implementation of such a method would/could be the same for all implementations (i.e.: just `consume(0)`).

ERuban:
> > So then, what about to implement smth like `LimiterInterface::peek()` to get the `RateLimit` object partly using the logic from `LimiterInterface::reserve()`
> 
> That would be a BC break (you can't change the interface). And doing it in a backwards compatible manner would probably mean it becomes a feature, not a bugfix, so would end up in 7.1 earliest. And personally (but I'm not a Symfony dev, just a users) I don't really like the idea as the implementation of such a method would/could be the same for all implementations (i.e.: just `consume(0)`).

We're anyway have a BC break here (for `sliding_window`):
> Anyway we have a BC break here, bcs seems that the old RateLimit->getRetryAfter() returns the time when all tokens will be available (if there is no available tokens), and the new one will return the time when the new one will be available



Hm, is it a BC break if we will add some new into the interface w/o changing the existing methods? Not sure.

ERuban:
@derrabus any updates? can we add @wouterj as a reviewer here?

derrabus:
I only did an initial triage on this PR. I've mentioned @wouterj because I believe he can conduct a better review on this topic than I could. However, it's SymfonyCon this week, so he's probably busy.

ERuban:
Have found that `RateLimit->getRetryAfter()` returns `now` when the last token was consumed for `sliding_window` and `fixed_window`.

Have updated the PR.

wouterj:
First a big major disclaimer, as I think it's very important:

The pattern of "peeking" is not something we want to encourage (which is also why it's not part of the RateLimiter component). A lot of the limiter is about dealing with concurrency. For this to work, consume and checking if you can consume must be part of the same action/method call.

Otherwise, you run the risk of one process "stealing" the one free token between the if statement and the `consume(1)` call. This is even more important, as you don't verify that `consume(1)` did indeed result in success.

The code should be restructured to do it like this, which combines consuming and checking if it's accepted in one call (and thus one lock, which means we're protected against concurrent processes):

```php
$rateLimit = $limiter->consume(1);
if (!$rateLimit->isAccepted()) {
   throw new SomeException($rateLimit->getRetryAfter());
}

// ...
```

Login throttling doesn't do this, but instead "peeks" at the limit state for a couple reasons. At first, it's a bit of a weird type of rate limiting as it is limiting wrong logins, but shouldn't limit good logins (rather than normal rate limited processes, where you want to limit all processes). Secondly, it's a careful balancing act between performance and security. Authentication is right on the hot path of an application (also for good/valid calls), so we must keep performance impact low. See https://github.com/symfony/symfony/pull/46110 for more background.

---

Now to this issue: We are indeed doing this in the login throttling, and even rely on the `getRetryAfter()` value in here: https://github.com/symfony/symfony/blob/c4e97ebac3df1094c3a50538f122d64852beb871/src/Symfony/Component/Security/Http/EventListener/LoginThrottlingListener.php#L50-L56

So I do think this is something we must fix. Even more, `sliding_window` was actually the only limiter that returned a correct retry after value. `fixed_window` and `token_bucket` are both broken in all versions, and `sliding_window` since 6.4. This probably went unnoticed, as the default rate throttling policy is sliding_window.

The course of action, I think, is:

* Fix fixed_window and token_bucket in 5.4
* Fix sliding_window in 6.4 (given it worked <6.4)

Are you willing to make a PR for 5.4 to fix the 2 policies there? (and then remove the fixed window fix from this PR)

RobertMe:
@wouterj could you elaborate on what's wrong with `fixed_window`? Because at he beginning of the year I changed it (#49070) because the `retryAfter` didn't make sense IMO. As it always returned the full time window, even if some time already had passed. I.e.: when applying it to login with a window of 5 minutes you would constantly get a retryAfter of 5 minutes, even if you already were blocked 4 minutes ago and you actually only had to wait 1 minute anymore. (I.e.: another login attempt after 1 minute would pass at that point).
So at least if the `retryAfter` of `fixed_window` is broken because of #49070 then the actual implementation of when a new token gets available is broken too. As obviously the `retryAfter` value should match up with when you actually can retry. So calling `consume` earlier should end up with the same "end time" (so "now + retryAfter" must always be the same) and calling `consume` right after the `retryAfter` has passed should result in a token being available. Which thus wasn't the case without #49070, as before "now + retryAfter" would constantly move, while the actual moment you could call `consume` resulting in an available token didn't move.

ERuban:
@wouterj 
> Are you willing to make a PR for 5.4 to fix the 2 policies there? (and then remove the fixed window fix from this PR)

Let me try 👌🏽

ERuban:
@RobertMe no, your fix is ok, I think @wouterj means the case when the last token was consumed I've fixed in that PR. In that case the `retryAfter` value should be `now + full time`

ERuban:
@wouterj done.
- PR with fix for `fixed_window` and `token_bucket` https://github.com/symfony/symfony/pull/53064
- Updated that PR removing `fixed_window` fix

Kocal:
Hi everyone, I've also faced the same issue where `$limit->getRetryAfter()->getTimestamp()` can returns the same value than `time()`:

When putting a `dd($this->limiter, $limit, $limit->getRetryAfter()->getTimestamp(), time());` at `vendor/symfony/security-http/EventListener/LoginThrottlingListener.php:55`: 
image

We can see `$limit->getRetryAfter()->getTimestamp()` and `time()` returns the same value.

When applying this PR locally, I have now more real values than before: 
image

Thanks for the PR :)

ERuban:
@wouterj 👋

wouterj:
Sorry for causing confusion here, I might have tested a bit wrongly.

In any case, I've created a PR for 6.3 to make the behavior across all policies consistent and make sure it's covered by a test: #53259 When that is merged, we can rebase this PR and fix the remaining issues caused by the reservation feature in 6.4.
Still, note that it is **discouraged to use this peeking behavior** unless you know what you're doing (e.g. you know the race conditions) and double check consuming a token afterwards is still accepted (so **do not copy the example in the PR description**).

_As a little side note, please do not ping maintainers if nothing changed. We're all doing this completely voluntary, so there is no guarantee whatsoever that we're responding within a certain time frame._

nicolas-grekas:
I merged https://github.com/symfony/symfony/pull/53259 into 6.3 but *NOT* into 6.4 because I don't know how to resolve conflicts.
@wouterj @ERuban help wanted to resolve this topic on 6.4 :pray:

wouterj:
Hi!

I've rebased the branch of this PR and reintroduced the testcase added by my PR in 6.3. I've also reorganized your code a bit, to reduce the number of if-else statements (I always struggle reading rate limiter code, so I like to keep things as simple as possible even if this comes with a cost of some code duplication).

However, GitHub for some reason doesn't allow me to push to this PR (even though the UI states maintainers should be allowed). So I've started a new PR: #53282
I always find it a bit of a bummer to have to close a PR, but be aware that I kept your authorship over the commit. This means you still get the credits you deserve for fixing this bug. Thanks for finding, reproducing and fixing it!

ERuban:
@wouterj huh, just one more approved but closed PR, ok. Last one was here https://github.com/symfony/symfony-docs/pull/19205
wtf with my PRs?

wouterj:
In the sidebar of this PR, is this box checked or not? If it's not checked, can you maybe check it? I can then try to push to your fork and reopen this one.

image

ERuban:
@wouterj ok, done

wouterj:
I'm sorry, it still doesn't seem to work:

```
> git push git@github.com:ERuban/symfony.git
ERROR: Permission to ERuban/symfony.git denied to wouterj.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

I haven't had this before, so not sure how to help you resolve this in the future. In any case, keeping your commit is the most important bit of information :)

ERuban:
ok, nevermind. just let it work at the end)

ERuban:
also, `git push git@github.com:ERuban/symfony.git` looks a bit strange for me. 
Maybe I'm wrong, but why you push it into my git, why not to push just to that branch, in that repo?

ERuban:
It was my last PR to Symfony. (It is a bit weird to be approved but not merged more than one time)

Thanks to all, nice work.

nicolas-grekas:
@ERuban this is how github works, we cannot work around.
You might be happy to know that what matters most from a historical perspective is the git history, and @wouterj made it so you're still the author of the corresponding commit in #53282

ERuban:
@nicolas-grekas https://github.com/symfony/symfony/pull/53282#issuecomment-1906016905
meh... I'm not sure that I want to do this work again.

Also there a second PR that waiting for you @wouterj a very long time... https://github.com/symfony/symfony/pull/53064

wouterj:
> Also there a second PR that waiting for you @wouterj a very long time... https://github.com/symfony/symfony/pull/53064

Sorry, I can't recall having any agreement with the Symfony organization or community about my response time. Been enjoying new years, some busy work weeks and a winter holiday, knowing that the BC break in 6.4 is fixed (which this PR was about).
Maybe I'll look at your other PR (which is about the bug you mention in your comment on #53282) some time before next release, but I must be honest that messages like this don't really motivate me to do so.

ERuban:
@wouterj sorry for that, but it was your initiative here at the end of the comment https://github.com/symfony/symfony/pull/52835#issuecomment-1847832956

PR #52843, opened 2023-12-01
[DependencyInjection] Prepending extension configs with file loaders

### LINKED ISSUE #52789: Allow prepending extension config in DI Loader (XML/YAML)
### Description

Given: A Bundle extension with an implemented `PrependExtensionInterface`

```php
public function prepend(ContainerBuilder $container): void
{
    $loader = new YamlFileLoader($container, new FileLocator(__DIR__ . '/../../config'));
 
    if ($specialConditionTriggers) {
        $loader->load('my_config_and_services.yaml');
    }
}
```

The Problem: The Loader (in this example, I'm using the YAML file loader) calls `loadFromExtension`, and so the configuration part gets appended:

https://github.com/symfony/symfony/blob/d04ef91ba2ab25b5b4decb3bc6a75a13c45d935f/src/Symfony/Component/DependencyInjection/Loader/YamlFileLoader.php#L952-L954

Adding a new argument to the loader methods `load()` is probably not the best solution regarding BC, but maybe we could add a second load method? For example:

```php
public function prependLoad(mixed $resource, string $type = null) {}
```

For now, I need to split up those configuration / service files, then I also have to parse the configuration manually to finally use `prependExtensionConfig` in my extension. This feels a bit odd because there is a service that could actually do it for me. :)

### Example

_No response_

yceruto:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #52789
| License       | MIT

https://github.com/symfony/symfony/pull/52636 continuation.

This is another try to simplify even more the prepending extension config strategy for Bundles and Extensions.

Feature: "Import and prepend an extension configuration from an external YAML, XML or PHP file"

> Useful when you need to pre-configure some functionalities in your app, such as mapping some DBAL types provided by your bundle, or simply pre-configuring some default values automatically when your bundle is installed, without losing the ability to override them if desired in userland.

```php
class AcmeFooBundle extends AbstractBundle
{
    {
        // Before
        $config = Yaml::parse(file_get_contents(__DIR__.'/../config/doctrine.yaml'));
        $builder->prependExtensionConfig('doctrine', $config['doctrine']);

        // After
        $container->import('../config/doctrine.yaml');
    }
}
```
The "Before" section is limited to importing/parsing this kind of configuration by hand; it doesn't support features like `when@env`, recursive importing, or defining parameters/services in the same file. Further, you can't simply use `ContainerConfigurator::import()` or any `*FileLoader::load()` here, as they will append the extension config instead of prepending it, defeating the prepend method purpose and breaking your app's config strategy. Thus, you are forced to use `$builder->prependExtensionConfig()` as the only way. 

> This is because if you append any extension config using `$container->import()`, then you won't be able to change that config in your app as it's merged with priority. If anyone is doing that currently, it shouldn't be intentional.

Therefore, the "After" proposal changes that behavior for `$container->import()`. *Now, all extension configurations encountered in your external file will be prepended instead. BUT, this will only happen here, at this moment, during the `prependExtension()` method.*

Of course, this little behavior change is a "BC break". However, it is a behavior that makes more sense than the previous one, enabling the usage of `ContainerConfigurator::import()` here, which was previously ineffective. 

This capability is also available for `YamlFileLoader`, `XmlFileLoader` and `PhpFileLoader` through a new boolean constructor argument:
```php
class AcmeFooBundle extends AbstractBundle
{
    {
        // Before
        // - It can't be used for prepending config purposes

        // After
        $loader = new YamlFileLoader($builder, new FileLocator(__DIR__.'/../config/'), prepend: true);
        $loader->load('doctrine.yaml'); // now it prepends extension configs as default behavior
        // or
        $loader->import('prepend/*.yaml');
    }
}
```
These changes only affect the loading strategy for extension configs; everything else keeps working as before.

Cheers!

### DIFF
--- a/src/Symfony/Component/DependencyInjection/CHANGELOG.md
+++ b/src/Symfony/Component/DependencyInjection/CHANGELOG.md
@@ -10,6 +10,8 @@ CHANGELOG
  * Add `#[Lazy]` attribute as shortcut for `#[Autowire(lazy: [bool|string])]` and `#[Autoconfigure(lazy: [bool|string])]`
  * Add `#[AutowireMethodOf]` attribute to autowire a method of a service as a callable
  * Make `ContainerBuilder::registerAttributeForAutoconfiguration()` propagate to attribute classes that extend the registered class
+ * Add argument `$prepend` to `FileLoader::construct()` to prepend loaded configuration instead of appending it
+ * [BC BREAK] When used in the `prependExtension()` methods, the `ContainerConfigurator::import()` method now prepends the configuration instead of appending it
 
 7.0
 ---
--- a/src/Symfony/Component/DependencyInjection/Loader/FileLoader.php
+++ b/src/Symfony/Component/DependencyInjection/Loader/FileLoader.php
@@ -45,10 +45,17 @@ abstract class FileLoader extends BaseFileLoader
     /** @var array<string, Alias> */
     protected array $aliases = [];
     protected bool $autoRegisterAliasesForSinglyImplementedInterfaces = true;
+    protected bool $prepend = false;
+    protected array $extensionConfigs = [];
+    protected int $importing = 0;
 
-    public function __construct(ContainerBuilder $container, FileLocatorInterface $locator, ?string $env = null)
+    /**
+     * @param bool $prepend Whether to prepend extension config instead of appending them
+     */
+    public function __construct(ContainerBuilder $container, FileLocatorInterface $locator, ?string $env = null, bool $prepend = false)
     {
         $this->container = $container;
+        $this->prepend = $prepend;
 
         parent::__construct($locator, $env);
     }
@@ -66,6 +73,7 @@ abstract class FileLoader extends BaseFileLoader
             throw new \TypeError(sprintf('Invalid argument $ignoreErrors provided to "%s::import()": boolean or "not_found" expected, "%s" given.', static::class, get_debug_type($ignoreErrors)));
         }
 
+        ++$this->importing;
         try {
             return parent::import(...$args);
         } catch (LoaderLoadException $e) {
@@ -82,6 +90,8 @@ abstract class FileLoader extends BaseFileLoader
             if (__FILE__ !== $frame['file']) {
                 throw $e;
             }
+        } finally {
+            --$this->importing;
         }
 
         return null;
@@ -215,6 +225,41 @@ abstract class FileLoader extends BaseFileLoader
         }
 
         $this->interfaces = $this->singlyImplemented = $this->aliases = [];
+    }
+
+    final protected function loadExtensionConfig(string $namespace, array $config): void
+    {
+        if (!$this->prepend) {
+            $this->container->loadFromExtension($namespace, $config);
+
+            return;
+        }
+
+        if ($this->importing) {
+            if (!isset($this->extensionConfigs[$namespace])) {
+                $this->extensionConfigs[$namespace] = [];
+            }
+            array_unshift($this->extensionConfigs[$namespace], $config);
+
+            return;
+        }
+
+        $this->container->prependExtensionConfig($namespace, $config);
+    }
+
+    final protected function loadExtensionConfigs(): void
+    {
+        if ($this->importing || !$this->extensionConfigs) {
+            return;
+        }
+
+        foreach ($this->extensionConfigs as $namespace => $configs) {
+            foreach ($configs as $config) {
+                $this->container->prependExtensionConfig($namespace, $config);
+            }
+        }
+
+        $this->extensionConfigs = [];
     }
 
     /**
--- a/src/Symfony/Component/DependencyInjection/Tests/Extension/AbstractExtensionTest.php
+++ b/src/Symfony/Component/DependencyInjection/Tests/Extension/AbstractExtensionTest.php
@@ -54,28 +54,55 @@ class AbstractExtensionTest extends TestCase
         self::assertSame($expected, $this->processConfiguration($extension));
     }
 
-    public function testPrependAppendExtensionConfig()
+    public function testPrependExtensionConfig()
     {
         $extension = new class() extends AbstractExtension {
+            public function configure(DefinitionConfigurator $definition): void
+            {
+                $definition->rootNode()
+                    ->children()
+                        ->scalarNode('foo')->end()
+                    ->end();
+            }
+
             public function prependExtension(ContainerConfigurator $container, ContainerBuilder $builder): void
             {
-                // append config
-                $container->extension('third', ['foo' => 'append']);
+                // prepend config from plain array
+                $container->extension('third', ['foo' => 'pong'], true);
 
-                // prepend config
-                $container->extension('third', ['foo' => 'prepend'], true);
+                // prepend config from external file
+                $container->import('../Fixtures/config/packages/ping.yaml');
+            }
+
+            public function loadExtension(array $config, ContainerConfigurator $container, ContainerBuilder $builder): void
+            {
+                $container->parameters()->set('foo_param', $config['foo']);
+            }
+
+            public function getAlias(): string
+            {
+                return 'third';
             }
         };
 
         $container = $this->processPrependExtension($extension);
 
         $expected = [
-            ['foo' => 'prepend'],
+            ['foo' => 'a'],
+            ['foo' => 'c1'],
+            ['foo' => 'c2'],
+            ['foo' => 'b'],
+            ['foo' => 'ping'],
+            ['foo' => 'zaa'],
+            ['foo' => 'pong'],
             ['foo' => 'bar'],
-            ['foo' => 'append'],
         ];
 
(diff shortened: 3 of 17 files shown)


src/Symfony/Component/DependencyInjection/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ CHANGELOG
 ---
 
  * Add argument `$prepend` to `ContainerConfigurator::extension()` to prepend the configuration instead of appending it
+ * Add argument `$prepend` to `FileLoader::construct()` to prepend loaded configuration instead of appending it
+ * [BC BREAK] The `ContainerConfigurator::import()` method now prepend extension configs instead of appending it, but only when used in the `AbstractBundle::prependExtension()` and `AbstractExtension::prependExtension` methods
```
nicolas-grekas:
```suggestion
 * [BC BREAK] When used in the `prependExtension()` methods, the `ContainerConfigurator::import()` method now prepends the configuration instead of appending it
 ```

**Review by nicolas-grekas — COMMENTED** [review]
I'm just wondering about recursive imports: are they also reversed?

**Comment by yceruto** [comment]
> I'm just wondering about recursive imports: are they also reversed?

The behavior is set up for the whole process, so yes, all extension configs imported recursively will be prepended too. However, unless there are split configs of the same extension distributed into several files matching the same config key, it doesn't matter.

**Comment by yceruto** [comment]
> The behavior is set up for the whole process, so yes, all extension configs imported recursively will be prepended too.

Which is expected if you're importing from the prepend method.

**Comment by nicolas-grekas** [comment]
Maybe you already answered but just to be sure: if I have a yaml file that imports two other files, I'd expect them to be loaded in top-down order, is that going to be the case?

**Comment by yceruto** [comment]
> Maybe you already answered but just to be sure: if I have a yaml file that imports two other files, I'd expect them to be loaded in top-down order, is that going to be the case?

I will confirm with a test case to ensure we know what the behavior will be.

File: src/Symfony/Component/DependencyInjection/Tests/Fixtures/config/packages/ping.yaml [hunk 2]
```diff
@@ -0,0 +1,6 @@
+imports:
+    - { resource: './third_a.yaml' }
+    - { resource: './third_b.yaml' }
+
+third:
+    foo: ping
```
yceruto, line 6: They will be loaded in top-down order using the prepend method, meaning that the first one imported will have priority over the rest. This matters when there are collisions between them (like the test case shows).

We can think about it as `array_merge()` works (per node level).

nicolas-grekas: Is that the same order than regular imports?
I think it would make sense.

yceruto: Nope, it's not the same order as the regular imports because appending configs will output in the opposite order than prepend (before merging process). But it makes sense to me as it will be the expected behavior according to the `prependExtension()` method. Here, you want to prepend [everything](https://github.com/symfony/symfony/pull/52843?notification_referrer_id=NT_kwDOAB7yprI4NjA5ODY0NzA2OjIwMjgxOTg&notifications_query=is%3Aunread#discussion_r1414106629) that you import, and that will include caring about sorting the recursive imports when necessary.

Note that, until now, using imports in the prepend method was not possible or defective (due to the reason given in the PR's description), so this is a new scenario, convention and a behavior change for the `ContainerConfigurator::import()` method (when used in the `prependExtension()` only) that must be well documented.

yceruto: Note that this behavior change only affects the importing of extension configs during `prependExtension()` method. Importing parameters, services, and everything else will follow the regular order.

nicolas-grekas: Is there a way to keep the regular order within the same yaml file even for recursive imports?
I feel like that might make more sense (the fact this is prepended is external to the file so within the file I would expect to not know about this ordering concern - or at least to easily forget about it - aka FTW pending :) )

src/Symfony/Component/DependencyInjection/Tests/Extension/AbstractExtensionTest.php
```diff
@@ -54,28 +54,52 @@ public function configure(DefinitionConfigurator $definition): void
...
 
         $expected = [
-            ['foo' => 'prepend'],
+            ['foo' => 'ping'],
+            ['foo' => 'b'],
+            ['foo' => 'a'],
+            ['foo' => 'pong'],
             ['foo' => 'bar'],
-            ['foo' => 'append'],
         ];
```
yceruto: @nicolas-grekas do you meant this instead, right?
```
['foo' => 'a'],     // third.yaml (recursive)
['foo' => 'b'],     // third.yaml (recursive)
['foo' => 'ping'],  // third.yaml
['foo' => 'pong'],  // directly in prepend extension
['foo' => 'bar'],   // app config
```

src/Symfony/Component/DependencyInjection/Tests/Fixtures/config/packages/ping.yaml
yceruto: I made the changes to keep the regular order within the same file, while the resulting extension configs are prepended at the end.

nicolas-grekas (approved):
LGTM (small rebase needed)

yceruto:
Thanks Nicolas! rebased done.

yceruto:
Friendly ping @symfony/mergers, there is a BC behavior break on the table here; it would be great to have a double check in case we miss something.

src/Symfony/Component/DependencyInjection/Loader/FileLoader.php
```diff
@@ -217,6 +227,41 @@ public function registerAliasesForSinglyImplementedInterfaces(): void
...
+            }
+            array_unshift($this->extensionConfigs[$namespace], $config);
+
+            return;
+        }
+
+        $this->container->prependExtensionConfig($namespace, $config);
+    }
+
+    protected function loadExtensionConfigs(): void
```
nicolas-grekas: do we have to make those 2 methods protected?

nicolas-grekas: I see that child classes call them. What about making them final?

yceruto: Done Nicolas! 🙏

yceruto:
Hey @nicolas-grekas, I rebased, rechecked, and updated another part where `when@env` wasn't working as expected. I added another test case to cover it. So, after a while, this looks ready to me. Wdyt?

src/Symfony/Component/DependencyInjection/Loader/FileLoader.php
```diff
@@ -217,6 +227,41 @@ public function registerAliasesForSinglyImplementedInterfaces(): void
...
+
+            return;
+        }
+
+        $this->container->prependExtensionConfig($namespace, $config);
+    }
+
+    final protected function loadExtensionConfigs(): void
+    {
+        if ($this->importing || [] === $this->extensionConfigs) {
```
nicolas-grekas:
```suggestion
        if ($this->importing || !$this->extensionConfigs) {
```

src/Symfony/Component/DependencyInjection/Loader/FileLoader.php
```diff
@@ -217,6 +227,41 @@ public function registerAliasesForSinglyImplementedInterfaces(): void
...
+            $this->container->loadFromExtension($namespace, $config);
+
+            return;
+        }
+
+        if ($this->importing) {
+            if (!isset($this->extensionConfigs[$namespace])) {
+                $this->extensionConfigs[$namespace] = [];
+            }
+            array_unshift($this->extensionConfigs[$namespace], $config);
```
nicolas-grekas, line 242:
```suggestion
            array_unshift($this->extensionConfigs[$namespace] ??= [], $config);
```

nicolas-grekas (approved):
Just some minor CS things and :rocket:

src/Symfony/Component/DependencyInjection/Loader/FileLoader.php
yceruto: Updated

src/Symfony/Component/DependencyInjection/Loader/FileLoader.php
yceruto: This one cannot be applied because https://3v4l.org/f6t7FP

nicolas-grekas:
Anyone @symfony/mergers ? (after a small rebase)

nicolas-grekas:
Thank you @yceruto.

PR #52861, opened 2023-12-01
[Form] Adding `@var` PHPDoc to silence psalm

### LINKED ISSUE #52829: [Form] `ChoiceType` is still calling the removed `PostSubmitEvent::setData()`

7.0.0

### Description

@HeahDude in https://github.com/symfony/symfony/pull/51043 you removed `FormEvent::setData()`.

But `ChoiceType::buildForm()` still contains this line at https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Form/Extension/Core/Type/ChoiceType.php#L162
```php
$event->setData($knownValues);
```

But this will now lead to a `BadMethodCallException` in [`PostSetDataEvent`](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Form/Event/PostSetDataEvent.php#L25) and [`PostSubmitEvent`](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Form/Event/PostSubmitEvent.php#L25).

Did you overlook this? Or is this line never executed in fact (and thus could be deleted)?

### How to reproduce

See above.

### Possible Solution

_No response_

### Additional Context

_No response_

ThomasLandauer:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes (kind of)
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52829
| License       | MIT

The goal is to add some PHPDoc to silence the psalm warnings mentioned in the linked issue, without changing the actual type hint (would be a BC break).

I found three ways to achieve this:
* Adding `@var` to the variable *in the next line*. That's what I'm doing in this PR. The idea is taken from https://stackoverflow.com/a/46842721/1668200 . It looks weird, but it's only a small change...
* Adding `@param` above the anonymous function. Would require to move it to a new line and subsequently indent the following ~50 lines:
    ```php
    $builder->addEventListener(FormEvents::PRE_SUBMIT,
        /** @param PreSubmitEvent $event */
        static function (FormEvent $event) use ($choiceList, $options, &$unknownValues) {
            // ...
        }
    );
    ```
* Refactoring by extracting the anonymous callback function to a dedicated method.

Waiting for feedback here, then I'll do the same in [`TimeType`](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Form/Extension/Core/Type/TimeType.php#L64) and [`FileType`](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Form/Extension/Core/Type/FileType.php#L50)

### DIFF
--- a/src/Symfony/Component/Form/Extension/Core/Type/ChoiceType.php
+++ b/src/Symfony/Component/Form/Extension/Core/Type/ChoiceType.php
@@ -30,6 +30,7 @@ use Symfony\Component\Form\ChoiceList\Loader\ChoiceLoaderInterface;
 use Symfony\Component\Form\ChoiceList\View\ChoiceGroupView;
 use Symfony\Component\Form\ChoiceList\View\ChoiceListView;
 use Symfony\Component\Form\ChoiceList\View\ChoiceView;
+use Symfony\Component\Form\Event\PreSubmitEvent;
 use Symfony\Component\Form\Exception\TransformationFailedException;
 use Symfony\Component\Form\Extension\Core\DataMapper\CheckboxListMapper;
 use Symfony\Component\Form\Extension\Core\DataMapper\RadioListMapper;
@@ -101,6 +102,7 @@ class ChoiceType extends AbstractType
             // Make sure that scalar, submitted values are converted to arrays
             // which can be submitted to the checkboxes/radio buttons
             $builder->addEventListener(FormEvents::PRE_SUBMIT, static function (FormEvent $event) use ($choiceList, $options, &$unknownValues) {
+                /** @var PreSubmitEvent $event */
                 $form = $event->getForm();
                 $data = $event->getData();
 


src/Symfony/Component/Form/Extension/Core/Type/ChoiceType.php
```diff
@@ -101,6 +101,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             // Make sure that scalar, submitted values are converted to arrays
             // which can be submitted to the checkboxes/radio buttons
             $builder->addEventListener(FormEvents::PRE_SUBMIT, static function (FormEvent $event) use ($choiceList, $options, &$unknownValues) {
+                /** @var PreSubmitEvent $event */
```
yceruto, line 105: Missing class import

stof:
As Symfony ensures that the `FormEvents::PRE_SUBMIT` event is triggered using a PreSubmitEvent object (and as this is inside the Form component, we don't have to care about older versions where this PreSubmitEvent was not there and FormEvent was instantiated), I would suggest changing the typehint instead.

ThomasLandauer:
That's what I suggested too, but others said it's not worth the BC break, see https://github.com/symfony/symfony/issues/52829#issuecomment-1834625465

src/Symfony/Component/Form/Extension/Core/Type/ChoiceType.php
ThomasLandauer: Sorry, added now.

yceruto:
@stof technically you can dispatch any FormEvent subtype (different from PreSubmitEvent) through the form event dispatcher accessible through the `$form->getConfig()->getEventDispatcher()` instance for a `ChoiceType` field. Why someone would want to do that, I don't know, but it's not only the Form class that is able to do it.

I'm not against changing it directly without deprecation, though, as it's really a special use case.

nicolas-grekas:
Thank you @ThomasLandauer.

PR #52865, opened 2023-12-02
remove $ so gitclip works

tacman:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | no
| New feature?  no 
| License       | MIT

Using the copy button puts the $ into the clipboard, so it work or to simply paste it on the command line.

### DIFF
--- a/src/Symfony/Component/Translation/README.md
+++ b/src/Symfony/Component/Translation/README.md
@@ -6,8 +6,8 @@ The Translation component provides tools to internationalize your application.
 Getting Started
 ---------------
 
-```
-$ composer require symfony/translation
+```bash
+composer require symfony/translation
 ```
 
 ```php


derrabus:
We have a lot more README files with that issue. If we change it here, we should change it everywhere.

tacman:
Agreed!  If it gets accepted I can do the rest as well.  If it's rejected
(I think it has been in the past), I won't bother.

But it's always bugged me.  clip, switch to terminal, paste, ctrl-a,
forward-arrow twice, backspace twice, enter.

derrabus:
> If it's rejected (I think it has been in the past), I won't bother.

If you have a link to a previous discussion, this might help getting a decision here.

tacman:
Not sure -- I've submitted this issue to many repositories, but even if it
was rejected, it shouldn't be.  Nobody wants the $ in the clipboard when
they're pasting.

And the $ isn't even the prompt for windows, so it's not like it's so
important that we need to have it even though it makes it look like it
doesn't work.

smnandre:
It was discussed recently there https://symfony.com/blog/new-in-symfony-6-4-assetmapper-improvements 

Maybe it is possible to filter this char for Windows users just before it’s copied into the clipboard ?

OskarStark (approved):

chalasr:
👍 for every README file across the board, once for all.

smnandre:
13 in this repo it seems

```
```\s?\n\s?\$ 
```

```
src/Symfony/Component/WebLink/README.md
src/Symfony/Component/Translation/README.md
src/Symfony/Component/Templating/README.md
src/Symfony/Component/Stopwatch/README.md
src/Symfony/Component/Security/Http/README.md
src/Symfony/Component/Security/Core/README.md
src/Symfony/Component/Routing/README.md
src/Symfony/Component/RateLimiter/README.md
src/Symfony/Component/PasswordHasher/README.md
src/Symfony/Component/Mailer/README.md
src/Symfony/Component/ErrorHandler/README.md
src/Symfony/Component/Dotenv/README.md
src/Symfony/Component/Clock/README.md
```

fabpot (approved):

fabpot:
Thank you @tacman.

fabpot:
Follow-up in ecc628aa71

PR #52867, opened 2023-12-02
[Validator] Only trigger deprecation when Validator annotations are used

### LINKED ISSUE #52828: [Validator] Unable to solve "uses Doctrine Annotations to configure validation constraints, which is deprecated. Use PHP attributes instead." deprecation

6.4.0

### Description

We have entities where all the validation constraints have been converted to attributes, however symfony validator still complains about using validation constraints through annotations. This is because logic for detecting these is oversimplified and it throws deprecation whenever it detects _any_ attributes

https://github.com/symfony/symfony/blob/e3e3d5fb4ddcad3887a0d01c878ab9ff88cd06ff/src/Symfony/Component/Validator/Mapping/Loader/AnnotationLoader.php#L120-L122

Situation is made worse by being unable to disable annotations while having attributes on. Following config
```yml
framework:
    validation:
        enable_annotations: false
        enable_attributes: true
```

produces compilation error `The "enable_annotations" and "enable_attributes" options at path "framework.validation" must not be both set. Only the "enable_attributes" option must be used.`

### How to reproduce

Run validation in SF fullstack on following object
```php
/** @\ApiPlatform\Core\Annotation\ApiResource()) */
#[\Doctrine\ORM\MappingORM\Entity]
class ThirdParty {}
```

### Possible Solution

Allow turning off `enable_annotations` while having `enable_attributes` on, or improve deprecation detection so it complains only if it finds annotation _Constraint_

### Additional Context

_No response_

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52828
| License       | MIT

The deprecations are now triggered only if one of the annotations is from the validator component.

### DIFF
--- a/src/Symfony/Component/Validator/Mapping/Loader/AnnotationLoader.php
+++ b/src/Symfony/Component/Validator/Mapping/Loader/AnnotationLoader.php
@@ -118,13 +118,13 @@ class AnnotationLoader implements LoaderInterface
         $annotations = [];
 
         if ($reflection instanceof \ReflectionClass && $annotations = $this->reader->getClassAnnotations($reflection)) {
-            trigger_deprecation('symfony/validator', '6.4', 'Class "%s" uses Doctrine Annotations to configure validation constraints, which is deprecated. Use PHP attributes instead.', $reflection->getName());
+            $this->triggerDeprecationIfAnnotationIsUsed($annotations, sprintf('Class "%s"', $reflection->getName()));
         }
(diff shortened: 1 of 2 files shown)


xabbuh (approved):

derrabus (approved):

OskarStark (approved):

derrabus:
Thank you @HypeMC.

PR #52874, opened 2023-12-03
[Scheduler] Separate id and description in message providers

### LINKED ISSUE #52853: [Scheduler] Cannot have multiple #[AsPeriodicTask] on the same method with the same frequency

6.4

### How to reproduce
Wire a class containing the following into a framework app with scheduler. 

```php	
#[AsPeriodicTask('PT30M', arguments: ['arg_1'])]
#[AsPeriodicTask('PT30M', arguments: ['arg_2']')]
public function method(string $arg): void {}
```

### Description
Fails to instantiate a `Schedule` with `Duplicated schedule message.`
Visibly, they are different tasks, but `ServiceCallMessage::__toString()` does not consider arguments.

### Possible Solution
Serialize args, potentially crc32 them too, into `ServiceCallMessage::__toString()`

valtzu:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52853
| License       | MIT

Separate id and description in message providers to keep `debug:schedule` output clean while allowing arbitrary distinct id.


valtzu:
Not sure what to do with the tests since `FrameworkBundle` tests now depend on the unreleased/unmerged Scheduler version (this PR).

src/Symfony/Component/Scheduler/Messenger/ServiceCallMessage.php
```diff
@@ -42,6 +42,17 @@ public function getArguments(): array
 
     public function __toString(): string
     {
-        return "@$this->serviceId".('__invoke' !== $this->method ? "::$this->method" : '');
+        return "@$this->serviceId".('__invoke' !== $this->method ? "::$this->method" : '').$this->getSerializedArguments();
```
fabpot: The value returned by `__toString()` should probably be human-readable, which is not the case anymore.

valtzu: Because of arguments can be long?

Or just the method name is misleading – I'm not using php serialization like the name may imply.

```
$ bin/console debug:schedule
 ----------------- ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ --------------------------------- 
  Trigger           Provider                                                                                                                                                                                   Next Run                         
 ----------------- ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ --------------------------------- 
  every 2 seconds   Symfony\Component\Scheduler\Messenger\ServiceCallMessage(@App\MessageHandler\TriggeringMessageHandler::handleTriggeringMessage(1))                                                         Mon, 04 Dec 2023 10:59:45 +0200  
  every 2 seconds   Symfony\Component\Scheduler\Messenger\ServiceCallMessage(@App\MessageHandler\TriggeringMessageHandler::handleTriggeringMessage(2))                                                         Mon, 04 Dec 2023 10:59:45 +0200  
  every 2 seconds   Symfony\Component\Scheduler\Messenger\ServiceCallMessage(@App\MessageHandler\TriggeringMessageHandler::handleTriggeringMessage('asdasd asd asda sdasd asd asdasd asdasdasdasd asd asd'))   Mon, 04 Dec 2023 10:59:45 +0200  
 ----------------- ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ --------------------------------- 
```

I think that's still pretty much human readable, if the length is the issue then we'd need to truncate + add some hash for longer arguments. :thinking: 

But as mentioned in https://github.com/symfony/symfony/issues/52853#issuecomment-1837235609, I'd rather not use this `__toString` at all to define the uniqueness; instead it could be up to provider to define the id for the message. For example, implementation providing messages from DB would use ids from database, and static provider could just use array keys.

---

EDIT: I guess you mean if I use schedule provider instead of autoconfigured schedule, then it's possible to pass objects as arguments and that would indeed make it too complex show

fabpot: I was more thinking about what you mentioned in your "EDIT" paragraph.
I agree with you that using __toString() for uniqueness is really bad.

valtzu: I tried playing around it but I don't think this `getId`/`__toString` thing is properly fixable without a BC break. 

Should we just include hash of serialized args in `__toString` as a bugfix for 6.4 and try to start separating id & human readable description in 7.x? :thinking:

valtzu: :point_up: Did just that, now it's like

```
 ----------------- -------------------------------------------------------------------------------------------------------------------------------------------- --------------------------------- 
  Trigger           Provider                                                                                                                                     Next Run                         
 ----------------- -------------------------------------------------------------------------------------------------------------------------------------------- --------------------------------- 
  every 2 seconds   Symfony\Component\Scheduler\Messenger\ServiceCallMessage(@App\MessageHandler\TriggeringMessageHandler::handleTriggeringMessage)              Wed, 06 Dec 2023 12:41:06 +0200  
  every 2 seconds   Symfony\Component\Scheduler\Messenger\ServiceCallMessage(@App\MessageHandler\TriggeringMessageHandler::handleTriggeringMessage(#cc40c189))   Wed, 06 Dec 2023 12:41:06 +0200  
  every 2 seconds   Symfony\Component\Scheduler\Messenger\ServiceCallMessage(@App\MessageHandler\TriggeringMessageHandler::handleTriggeringMessage(#266e01fa))   Wed, 06 Dec 2023 12:41:06 +0200  
```

Can we go with this for 6.4 to fix the issue (which is quite big one in my opinion)?

alex-dev:
While you're addressing this, could you check other `__toString` scheduled messages? `RedispatchMessage` is also affected.

valtzu:
> While you're addressing this, could you check other `__toString` scheduled messages? `RedispatchMessage` is also affected.

@alex-dev but `RedispatchMessage` uses the wrapped message's `__toString()` and it does include the transport names in __toString so I don't think it's affected itself directly?

---

Or yeah if you don't implement `__toString` in your message, then it is affected indeed. But since right now `__toString` is how you define uniqueness, I would consider that a feature and not a bug.

alex-dev:
You don't define uniqueness only by `__toString`. If a message doesn't implement a `__toString`, `Schedule` use `serialize`. So, if a message wrapped by a `RedispatchMessage` is not `Stringable`, they should be serialized to determine uniqueness. `RedispatchMessage` should be transparent for identifying uniqueness.

valtzu:
Ok then, I separated id & description and used hashed serialized message for uniqueness with `StaticMessageProvider`.

I would've rather added explicit `getDescription()` on `MessageProviderInterface` but with optional `__toString` we should be able to avoid dealing with BC.

What do you think?

We could also consider adding `id` parameter to `#[AsPeriodicTask]` & `#[AsCronTask]` attributes in case someone would rather have static ids which don't depend on message content.

alex-dev:
Seems good!

src/Symfony/Component/Scheduler/RecurringMessage.php
```diff
@@ -75,14 +75,15 @@ public static function trigger(TriggerInterface $trigger, object $message): self
...
+        if ($message instanceof \Stringable) {
+            try {
+                $description .= "($message)";
+            } catch (\Exception) {
+            }
         }
-        $description = sprintf('%s(%s)', $message::class, $description);
 
-        return new self($trigger, new StaticMessageProvider([$message], $description));
+        return new self($trigger, new StaticMessageProvider([$message], hash('crc32c', serialize($message)), $description));
```
nicolas-grekas:
```suggestion
        return new self($trigger, new StaticMessageProvider([$message], strtr(substr(base64_encode(hash('xxh128', serialize($message), true)), 0, 7), '/+', '._'), -7), $description));
```

src/Symfony/Component/Scheduler/RecurringMessage.php
```diff
@@ -75,14 +75,15 @@ public static function trigger(TriggerInterface $trigger, object $message): self
...
         }
 
-        $description = '';
-        try {
-            $description = $message instanceof \Stringable ? (string) $message : serialize($message);
-        } catch (\Exception) {
+        $description = $message::class;
+        if ($message instanceof \Stringable) {
+            try {
+                $description .= "($message)";
```
nicolas-grekas:
```suggestion
                $description .= " ($message)";
```

nicolas-grekas (approved):
LGTM with a minor comment

nicolas-grekas:
Thank you @valtzu.

src/Symfony/Component/Scheduler/RecurringMessage.php
```diff
@@ -75,14 +75,15 @@ public static function trigger(TriggerInterface $trigger, object $message): self
...
+        if ($message instanceof \Stringable) {
+            try {
+                $description .= " ($message)";
+            } catch (\Exception) {
+            }
         }
-        $description = sprintf('%s(%s)', $message::class, $description);
 
-        return new self($trigger, new StaticMessageProvider([$message], $description));
+        return new self($trigger, new StaticMessageProvider([$message], strtr(substr(base64_encode(hash('xxh128', serialize($message), true)), 0, 7), '/+', '._'), -7), $description));
```
xabbuh, line 86: FYI, there was a syntax error on this line. I fixed that in a45f6cbeb6e05feb24b8d67e6f361da418511d3b. Please double-check that I didn't mess anything with the logic implemented here.

Issue #52872, opened 2023-12-03
[Doctrine Messenger] Igbinary de/serialization support and optimization for modern databases


FluffyDiscord:
### Description

The default php `serialize` work fine for most situations, but is incredibly slow (1-5 times slower) and it's outputs are a lot longer (30-120% more).

The current approach to de/serialization is to `base64_encode` the message and insert it to `string` column, then `base64_decode` back when retrieving, further slowing down the whole process. We could completely skip these steps - if the used database supports binary columns, which most current databases do (Sqlite is an exception, fallback would be triggered), we could simply store this as binary instead.

The only "drawbacks" from these approaches would be that messages will no longer be "kind of readable" in database.

I have added myself very naive implemetation [in my fork](https://github.com/symfony/doctrine-messenger/compare/6.3...FluffyDiscord:doctrine-messenger:6.3) which works wonders, but is not safe (removed all error checks) and sets the serializer globally, which breaks queues that do not support binary content (did not test, since I do not use any except DB, just speculating).

derrabus:
Sounds reasonable, please send a PR.

FluffyDiscord:
Hi, [here](#52877) with questions at the end.

carsonbot:
Thank you for this suggestion.
There has not been a lot of activity here for a while. Would you still like to see this feature?

carsonbot:
Could I get an answer? If I do not hear anything I will assume this issue is resolved or abandoned. Please get back to me <3

FluffyDiscord:
yes I would, the PR was silently rejected

nicolas-grekas:
This looks like a "won't fix" to me: the current behavior is safe and as such it's the one we should ship in Symfony.
You can always provide your own transport for more tailored needs.

PR #52877, opened 2023-12-03
[Messenger Doctrine] Support for `igbinary` serializer and `binary` message `body` column

FluffyDiscord:
| Q             | A
| ------------- | ---
| Branch       | 7.1, but [here you say that new features should be targetting 6.4](https://symfony.com/doc/current/contributing/code/pull_requests.html#choose-the-right-branch), so I am not sure 
| Bug fix      | no
| New feature  | yes
| Deprecations | no
| Issues        | #52872
| License       | MIT

### Description

This PR introduces alternative Messenger Transport, the [igbinary](https://github.com/igbinary/igbinary) serializer `Symfony\Component\Messenger\Transport\Serialization\IgbinarySerializer` and deepens the support by also modifying `Symfony\Component\Messenger\Bridge\Doctrine\Transport\Connection` to support `binary` column type whenever this serializer is enabled.

The `igbinary` serializer is by default 1-5 times faster and it's output is around 30-120% shorter than PHP's build-in `serialize`. Added serializer was also stripped of the `base64` pass both on `encode` and `decode` to further speed up the conversion. To support `binary` message `body` we also need to modify the Doctrine transport connection, so it works with `binary` column type instead of `text`.
 
Every DBAL supported database except Sqlite supports `binary` columns and installing `igbinary` extension is easy as installing alteady built `php-igbinary` package from repos and in other cases `pecl install igbinary` can be used.

With this PR tests might be broken as they now require `ext-igbinary` to be installed - maybe this can be skipped if it's not installed?

### Questions

1. I was not sure how to pass the information to `Symfony\Component\Messenger\Bridge\Doctrine\Transport\Connection` so it knows if it should use `binary` or `text` column type for `body` without making BC. Please let me know if the current approach is fine or not. 
2. How do I fix the `psalm` error about non existent `Doctrine\DBAL\Platforms\SqlitePlatform`? Do I need to use `class_implements($platform)` with `in_array()` or is there a better solution? I need to check whenever the used platform is Sqlite to throw exception.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/console.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/console.php
@@ -198,6 +198,7 @@ return static function (ContainerConfigurator $container) {
                 service('logger')->nullOnInvalid(),
                 service('messenger.transport.native_php_serializer')->nullOnInvalid(),
                 null,
+                service('messenger.transport.igbinary_serializer')->nullOnInvalid(),
             ])
             ->tag('console.command')
             ->tag('monolog.logger', ['channel' => 'messenger'])
@@ -207,6 +208,7 @@ return static function (ContainerConfigurator $container) {
                 abstract_arg('Default failure receiver name'),
                 abstract_arg('Receivers'),
                 service('messenger.transport.native_php_serializer')->nullOnInvalid(),
+                service('messenger.transport.igbinary_serializer')->nullOnInvalid(),
             ])
             ->tag('console.command')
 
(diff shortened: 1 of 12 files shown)


nicolas-grekas:
There's one catch with igbinary: it's not really well maintained (or better: I'd say it misses contributors)
See #52391

I'm not sure enabling more igbinary use cases is a good idea until issues like https://github.com/igbinary/igbinary/issues/273 are resolved.

FluffyDiscord:
The main usage for this is to serialize Messenger message, which should not really be an issue in this case, you always initialize properties and you don't have any business logic in them, they are just a "fancy" data payloads. The issue you are referring to is a different situation.

PR #52881, opened 2023-12-04
[DoctrineBridge] Global query time always at 0.00 ms on profiler

### LINKED ISSUE #52880: [Bug][DoctrineBridge] Global query time always at 0.00 ms on profiler

7.0

### Description

Since I updated a project from 6.3 to 7.0, the global query time for Doctrine queries is always at 0.00 ms on the profiler.

For the same http request :
Screenshot from 2023-12-04 05-33-24
Screenshot from 2023-12-04 05-33-39

### How to reproduce

Do multiple doctrine calls on a same http request

### Possible Solution

After a few tests, it seems that the problem comes from the switch to native return types on `Symfony\Bridge\Doctrine\DataCollector\DoctrineDataCollector` method.

For context :

```
public function getTime(): int
{
    $time = 0;
    foreach ($this->data['queries'] as $queries) {
        foreach ($queries as $query) {
            $time += $query['executionMS'];
        }
    }

    return $time;
}
```

Inside the `getTime` method, when I check what's inside `$time` before the return, I have a float smaller than 0 (something like 0.0015671253204346 for example)
So when returning, this value is casted to int and become 0.

After changing the return type of the `getTime` method to float, the calculation worked again.

Working on a PR.

### Additional Context

_No response_

maximethiry:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52880
| License       | MIT

(original explanation made from 7.0 branch, now targeting 6.4 cf comments)

The related issue, when i explained the possible solution : https://github.com/symfony/symfony/issues/52880

On `Symfony\Bridge\Doctrine\DataCollector\DoctrineDataCollector`, updated the native return type from `int` to `float` for the method `getTime`.

Context : 
```
    public function getTime(): int
    {
        $time = 0;
        foreach ($this->data['queries'] as $queries) {
            foreach ($queries as $query) {
                $time += $query['executionMS'];
            }
        }

        return $time;
    }
```

Inside this method, the value of `$time` was always a float smaller than 0. The transition to native return type (which was correct regarding to old phpdoc type) made the return cast int to the floating value, hence returning always 0.

I updated the return type to float and updated the test to check for the calculus on floating values.
I had to switch to usleep instead of sleep, since the value might not be an int.

### DIFF
--- a/src/Symfony/Bridge/Doctrine/DataCollector/DoctrineDataCollector.php
+++ b/src/Symfony/Bridge/Doctrine/DataCollector/DoctrineDataCollector.php
@@ -146,7 +146,7 @@ class DoctrineDataCollector extends DataCollector
     }
 
     /**
-     * @return int
+     * @return float
      */
     public function getTime()
     {
--- a/src/Symfony/Bridge/Doctrine/Tests/DataCollector/DoctrineDataCollectorTest.php
+++ b/src/Symfony/Bridge/Doctrine/Tests/DataCollector/DoctrineDataCollectorTest.php
@@ -183,7 +183,7 @@ class DoctrineDataCollectorTest extends TestCase
             $debugDataHolder->addQuery('default', $query);
 
             if (isset($queryData['executionMS'])) {
-                sleep($queryData['executionMS']);
+                usleep($queryData['executionMS'] * 1000000);
             }
             $query->stop();
         }
--- a/src/Symfony/Bridge/Doctrine/Tests/DataCollector/DoctrineDataCollectorTestTrait.php
+++ b/src/Symfony/Bridge/Doctrine/Tests/DataCollector/DoctrineDataCollectorTestTrait.php
@@ -73,6 +73,36 @@ trait DoctrineDataCollectorTestTrait
         $this->assertEquals(3, $c->getTime());
     }
 
+    public function testCollectTimeWithFloatExecutionMS()
+    {
+        $queries = [
+            ['sql' => 'SELECT * FROM table1', 'params' => [], 'types' => [], 'executionMS' => 0.23],
+        ];
+        $c = $this->createCollector($queries);
+        $c->collect(new Request(), new Response());
+        $c = unserialize(serialize($c));
+        $this->assertEqualsWithDelta(0.23, $c->getTime(), .01);
+
+        $queries = [
+            ['sql' => 'SELECT * FROM table1', 'params' => [], 'types' => [], 'executionMS' => 1.02],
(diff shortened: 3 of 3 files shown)


src/Symfony/Bridge/Doctrine/Tests/DataCollector/DoctrineDataCollectorTest.php
```diff
@@ -88,6 +88,16 @@ public function testCollectTime()
...
+
+        $queries = [
+            ['sql' => 'SELECT * FROM table1', 'params' => [], 'types' => [], 'executionMS' => 0.15],
+            ['sql' => 'SELECT * FROM table2', 'params' => [], 'types' => [], 'executionMS' => 0.32],
+            ['sql' => 'SELECT * FROM table2', 'params' => [], 'types' => [], 'executionMS' => 0.07],
+        ];
+        $c = $this->createCollector($queries);
+        $c->collect(new Request(), new Response());
+        $c = unserialize(serialize($c));
+        $this->assertEquals(0.54, round($c->getTime(), 2));
```
derrabus:
```suggestion
        $this->assertEqualsWithDelta(0.54, $c->getTime(), .01);
```

derrabus: This should be a separate test.

derrabus (changes requested):

src/Symfony/Bridge/Doctrine/DataCollector/DoctrineDataCollector.php
```diff
@@ -85,7 +85,7 @@ public function getQueries(): array
         return $this->data['queries'];
     }
 
-    public function getTime(): int
+    public function getTime(): float
```
derrabus: We should change the annotated type on 6.4 if it's incorrect.

nicolas-grekas:
Can you please rebase the PR on 6.4 to change the annotation there instead?
We will handle the case for 7.0 while merging that PR on 6.4 into 7.0.

maximethiry:
Thanks for the feedback, will do the rebase :+1: 
Should I keep the test for floats values, after moving it in a more appropriate function as requested by @derrabus ?

derrabus:
Yes, please keep the test!

derrabus (approved):

derrabus:
Thank you @maximethiry.

derrabus:
> We will handle the case for 7.0 while merging that PR on 6.4 into 7.0.

Done.

PR #52885, opened 2023-12-04
[Serializer] fix nullable int cannot be serialized

nikophil:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

Hello,

previous to [this PR](https://github.com/symfony/symfony/pull/52172) such XML could be deserialized correctly, setting null in the value:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<DummyNullableInt>
	<value/>
</DummyNullableInt>
```
```php
class DummyNullableInt
{
    public function __construct(
        public int|null $value = null
    )
    {}
}
```

but now it creates the following error:
```
Uninitialized string offset 0

/home/nikophil/works/symfony/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php:495
/home/nikophil/works/symfony/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php:630
/home/nikophil/works/symfony/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php:377
/home/nikophil/works/symfony/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php:246
/home/nikophil/works/symfony/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php:346
```

I looked for any issue or PR mentioning this problem, but couldn't find it. So here is a fix 

ping @Jeroeny

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
@@ -524,7 +524,7 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
                             }
                             break;
                         case Type::BUILTIN_TYPE_INT:
-                            if (ctype_digit($data) || '-' === $data[0] && ctype_digit(substr($data, 1))) {
+                            if (ctype_digit($data) || isset($data[0]) && '-' === $data[0] && ctype_digit(substr($data, 1))) {
                                 $data = (int) $data;
                             } else {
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
```diff
@@ -292,8 +292,6 @@ abstract protected function extractAttributes(object $object, string $format = n
 
     /**
      * Gets the attribute value.
-     *
-     * @return mixed
```
nicolas-grekas: please revert those, they're false positives

nicolas-grekas (commented):
that's for 5.4 I suppose?
can you please rebase + retarget?

nikophil:
@nicolas-grekas done
 
the problem only occurs with nullable types properties, hence the `@requires PHP 8` tag

nicolas-grekas:
Thank you @nikophil.

PR #52886, opened 2023-12-04
[HttpKernel] Catch `TypeError` if the wrong type is used in `BackedEnumValueResolver`

### LINKED ISSUE #52053: BackedEnumValueResolver fails to resolve \IntBackedEnum when passing string

6.3.5

### Description

If I pass a string as the value for an `\IntBackedEnum` controller argument with `MapQueryParameter` attribute and `BackedEnumValueResolver` resolver, symfony generates a `\TypeError` exception.

### How to reproduce

Controller:
```php
#[Route('/category/{category}', name: 'list_category', methods: ['GET'])]
public function listAction(
    #[MapQueryParameter(resolver: BackedEnumValueResolver::class)] CategoryEnum $category = null,
): Response {
}
```

CategoryEnum:

```php
enum ArtistRank: int
{
    case FOO = 1;
    case BAR = 2;
}
```

### Possible Solution

Create separate resolvers for \IntBackedEnum and \StringBackedEnum

OR

Catch `\TypeError` along with the `\ValueError` in `BackedEnumValueResolver` and throw `NotFoundHttpException`

OR

Use `tryFrom` instead of `try` in `BackedEnumValueResolver` (probably a bad idea, because it will return null

### Additional Context

I can avoid this by using requirements in the router description like this:

```php
requirements: ['category' => new EnumRequirement(CategoryEnum::class)],
```

But this makes the route attribute unnecessary huge and less readable

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52053
| License       | MIT

### DIFF
--- a/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/BackedEnumValueResolver.php
+++ b/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/BackedEnumValueResolver.php
@@ -86,7 +86,7 @@ class BackedEnumValueResolver implements ArgumentValueResolverInterface, ValueRe
 
         try {
             return [$enumType::from($value)];
-        } catch (\ValueError $e) {
+        } catch (\ValueError|\TypeError $e) {
             throw new NotFoundHttpException(sprintf('Could not resolve the "%s $%s" controller argument: ', $argument->getType(), $argument->getName()).$e->getMessage(), $e);
         }
     }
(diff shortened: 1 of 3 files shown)


derrabus (approved):

nicolas-grekas:
Thank you @alexandre-daubois.

PR #52887, opened 2023-12-04
[WebProfilerBundle] Don't expect collector.compilerLogs to be defined

arjenm:
collector.compilerLogs is not always available and since reduce expects an array, it will cause a php-error.

| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

In my environment the debug toolbar's logs-panel would occasionally give a fatal error at the changed line:
```
The "reduce" filter only works with arrays or "Traversable", got "NULL" as first argument.
```

So apparently compilerLogs is null. Since there is a 'is not empty' test a bit later on, that is clearly expected behavior. So this is is probably broken since the for-loop that preceded this reduce-change was replaced a few months back.

There are probably more elegant fixes possible, like doing that empty test much earlier. But this is a much less invasive fix :)

### DIFF
--- a/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/logger.html.twig
+++ b/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/logger.html.twig
@@ -409,7 +409,7 @@
         <script>Sfjs.initializeLogsTable();</script>
     {% endif %}
 
-    {% set compilerLogTotal = collector.compilerLogs|reduce((total, logs) => total + logs|length, 0) %}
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Can you figure our when this can be null? Looking at the code, I don't see how this can happen.
Maybe you have a reproducer?

arjenm:
Good point, I also see no path where it can return null... Unless someone has their own data collector and uses Symfony's logger.html.twig-template...

Which we did and we clearly missed a public method-override.

And apparently we didn't feel the need to check that panel until the new Symfony 6.4 came out :)

PR #52891, opened 2023-12-04
[HttpKernel] Fix request attribute value ignored with pinned resolvers

### LINKED ISSUE #52875: #[MapEntity] and render controller from twig

6.4

### Description

When updating project and replacing `ParamConverter` to `MapEntity`, when render controller from a template, it fails.
Exception example:

An exception has been thrown during the rendering of a template ("Controller "App\Controller\ListController" requires that you provide a value for the "$tag" argument. Either the argument is nullable and no null value has been provided, no default value has been provided or there is a non-optional argument after this one.").

When used `ParamConverter` two options worked: calling via http and via render(controller()), now only http.

### How to reproduce

Below there will be a simplified example, the controller is available by route and we want to call it in the template with a pre-selected entity.

Controller:
```php
<?php

namespace App\Controller;

use App\Entity\Tag;
use Symfony\Bridge\Doctrine\Attribute\MapEntity;

final class ListController
{
    #[Route('/list/{slug}', name: 'app_list', methods: ['GET'])]
    public function __invoke(Request $request, #[MapEntity] Tag $tag): Response
    {
        return new Response($tag->getId());
    }
}
```

Template:
```twig
{{ render(controller('App\\Controller\\ListController', {
    'tag': tag,
})) }}
```

### Possible Solution

_No response_

### Additional Context

_No response_

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52875
| License       | MIT

This is similar to #50458. When pinned resolvers are used, values already resolved in the request attributes bag are ignored.

### DIFF
--- a/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver.php
+++ b/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver.php
@@ -73,6 +73,7 @@ final class ArgumentResolver implements ArgumentResolverInterface
 
                     $argumentValueResolvers = [
                         $this->namedResolvers->get($resolverName),
+                        new RequestAttributeValueResolver(),
                         new DefaultValueResolver(),
                     ];
                 }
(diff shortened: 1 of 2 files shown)


xabbuh (approved):

fabpot (approved):

fabpot:
Thank you @HypeMC.

PR #52894, opened 2023-12-05
[Mime] Fix address from string pattern regex

ghost:
| Q             | A
| ------------- | ---
| Branch?       |7.1
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | n/a
| License       | MIT

Fixes the `FROM_STRING_PATTERN` inside Address.php.

As it is right now strings with the character inside the display name "<" will not parse correctly.
For example the string `Andre Hoong <3 <hoongandre@gmail.com>` will have the following matches returned using the current pattern:
```
displayName => "Andre Hoong"
addrSpec => "3 <hoongandre@gmail.com"

```
The new string pattern '(?.)?<(?.)>[^>]*' allows the displayName to include the "<" character and correctly parses the addrSpec. It does this by taking the last occurence of the "<" to singal the start of the addrSpec

### DIFF
--- a/src/Symfony/Component/Mime/Address.php
+++ b/src/Symfony/Component/Mime/Address.php
@@ -31,7 +31,7 @@ final class Address
      * config, and allows to have more readable config.
      * This does not try to cover all edge cases for address.
      */
-    private const FROM_STRING_PATTERN = '~(?<displayName>[^<]*)<(?<addrSpec>.*)>[^>]*~';
+    private const FROM_STRING_PATTERN = '~(?<displayName>.*)?<(?<addrSpec>.*)>[^>]*~';
 
     private static EmailValidator $validator;
     private static IdnAddressEncoder $encoder;
--- a/src/Symfony/Component/Mime/Tests/AddressTest.php
+++ b/src/Symfony/Component/Mime/Tests/AddressTest.php
@@ -148,7 +148,7 @@ class AddressTest extends TestCase
                 'Jane Doe <3 <example@example.com>',
                 'Jane Doe <3',
                 'example@example.com'
-            ]
+            ],
         ];
     }
 
--- a/src/Symfony/Component/Mime/Tests/AddressTest.php
+++ b/src/Symfony/Component/Mime/Tests/AddressTest.php
@@ -144,6 +144,11 @@ class AddressTest extends TestCase
                 'Jane Doe',
                 'example@example.com',
             ],
+            [
+                'Jane Doe <3 <example@example.com>',
+                'Jane Doe <3',
+                'example@example.com'
+            ]
         ];
     }
 
--- a/src/Symfony/Component/Mime/Tests/AddressTest.php
+++ b/src/Symfony/Component/Mime/Tests/AddressTest.php
@@ -147,7 +147,7 @@ class AddressTest extends TestCase
             [
                 'Jane Doe <3 <example@example.com>',
                 'Jane Doe <3',
-                'example@example.com'
+                'example@example.com',
             ],
         ];
     }


GromNaN (changes requested):
Hello @andrehoong-pixieset, thanks for your contribution.

I see tests failing for other cases: 
```
1) Symfony\Component\Mime\Tests\AddressTest::testCreateWithString with data set #6 ('Jane Doe <"ex<ample"@example.com>', 'Jane Doe', '"ex<ample"@example.com')
Symfony\Component\Mime\Exception\RfcComplianceException: Email "ample"@example.com" does not comply with addr-spec of RFC 2822.

2) Symfony\Component\Mime\Tests\AddressTest::testCreateWithString with data set #7 ('Jane Doe <"ex<amp>le"@example.com>', 'Jane Doe', '"ex<amp>le"@example.com')
Symfony\Component\Mime\Exception\RfcComplianceException: Email "amp>le"@example.com" does not comply with addr-spec of RFC 2822.

3) Symfony\Component\Mime\Tests\AddressTest::testCreateWithString with data set #8 ('Jane Doe > <"ex<am  p>le"@example.com>', 'Jane Doe >', '"ex<am  p>le"@example.com')
Symfony\Component\Mime\Exception\RfcComplianceException: Email "am  p>le"@example.com" does not comply with addr-spec of RFC 2822.
```

How do you know if the address is `3 <example@example.com` or `example@example.com`. This notation if ambiguous.

It might be necessary to exclude spaces from the `addrSpec` part. They are not allowed in the EmailValidator (let see what other think): https://github.com/symfony/symfony/blob/dc330b087e496d2851cf9461fe27b16e69a66c89/src/Symfony/Component/Validator/Constraints/EmailValidator.php#L29-L30

src/Symfony/Component/Mime/Address.php
```diff
@@ -31,7 +31,7 @@ final class Address
      * config, and allows to have more readable config.
      * This does not try to cover all edge cases for address.
      */
-    private const FROM_STRING_PATTERN = '~(?<displayName>[^<]*)<(?<addrSpec>.*)>[^>]*~';
+    private const FROM_STRING_PATTERN = '~(?<displayName>.*)?<(?<addrSpec>.*)>[^>]*~';
```
nicolas-grekas, line 34: what about this instead? would it allow fixing the failing test cases?
note that the comment above needs an update also
```suggestion
    private const FROM_STRING_PATTERN = '~(?<displayName>.*?)<(?<addrSpec>[^<>]*)>~';
```

jprivet-dev: Hi everyone! 

This version does not solve the problem. This regex `~(?<displayName>.*?)<(?<addrSpec>[^<>]*)>~` crashes 3 initial tests with the following values:

- `Jane Doe <"ex<ample"@example.com>`
- `Jane Doe <"ex<amp>le"@example.com>`
- `Jane Doe > <"ex<am  p>le"@example.com>`

jprivet-dev:
Hi !

Here are some of the cases I've just tested:

| Case | Test | String | Expected Name | Expected Addr |
|-|-|-|-|-|
| 1 | ✅ | `Jane Doe ><example@example.com>` | `Jane Doe >` | `example@example.com` |
| 2 | ✅ | `Jane Doe > <example@example.com>` | `Jane Doe >` | `example@example.com` |
| 3 | ✅ | `Jane Doe > string <example@example.com>` | `Jane Doe > string` | `example@example.com` |
| 4 | ✅ | `Jane Doe >><example@example.com>` | `Jane Doe >>` | `example@example.com` |
| 5 | ✅ | `Jane Doe >> <example@example.com>` | `Jane Doe >>` | `example@example.com` |
| 6 | ✅ | `Jane Doe >> string <example@example.com>` | `Jane Doe >> string` | `example@example.com` |
| 7 | ❌ | `Jane Doe <<example@example.com>` | `Jane Doe <` | `example@example.com` |
| 8 | ❌ | `Jane Doe < <example@example.com>` | `Jane Doe <` | `example@example.com` |
| 9 | ❌ | `Jane Doe < string <example@example.com>` | `Jane Doe < string` | `example@example.com` |
| 10 | ❌ | `Jane Doe <<<example@example.com>` | `Jane Doe <<` | `example@example.com` |
| 11 | ❌ | `Jane Doe << <example@example.com>` | `Jane Doe <<` | `example@example.com` |
| 12 | ❌ | `Jane Doe << string <example@example.com>` | `Jane Doe << string` | `example@example.com` |
| 13 | ❌ | `Jane Doe <><example@example.com>` | `Jane Doe <>` | `example@example.com` |
| 14 | ❌ | `Jane Doe <example@example.com>>` | `Jane Doe` | `example@example.com` |
| 15 | ❌ | `Jane Doe <<example@example.com>>` | `Jane Doe <` | `example@example.com` |
| 16 | ✅ | `Jane Doe ¯\_(ツ)_/¯ <example@example.com>` | `Jane Doe ¯\_(ツ)_/¯` | `example@example.com` |
| 17 | ✅ | `Jane Doe 🚀 <example@example.com>` | `Jane Doe 🚀` | `example@example.com` |

Cases 7 to 15 do not pass the tests with the initial `FROM_STRING_PATTERN`:

https://github.com/symfony/symfony/blob/ad68f738d00388d2a8c691b92d43704fa4fcec91/src/Symfony/Component/Mime/Address.php#L34

Question: which cases (from 7 to 15) do we consider to be invalid and which do we have to deal with?

fabpot:
The regex will never be perfect. So, two possibilities here: we stop using a regex and implement a small parser or we fix "most" problems but not all. In any case, in most cases, the constructor should be used.

PR #52900, opened 2023-12-05
[Serializer] Add ChainNormalizer and ChainDenormalizer

Nyholm:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Issues        | 
| License       | MIT

This PR is less of a feature and more refactoring.
Our Serializer currently implements 5 interfaces. You may serialize and normalize (and reverse). There are also a lot of logic to figure out context, error messages and what normalizer to use. 

This PR suggest to separate responsibilities a bit by splitting the normalizing and denormalizing into `ChainNormalizer` and `ChainDenormalizer`. 

I found this work when I was creating custom denormalizers. A low level class like `MyBlogDenormalizer` should not have a dependency on the high level `Serializer`. It would make more sense to use a `ChainDenormalizer`. (Yes, there is still an ugly circular reference). 

I've marked `SerializerAwareInterface` as deprecated for this reason. Currently, `SerializerAwareInterface` only works if you combine it with `NormalizerInterface`, `DenormalizerInterface` or `EncoderInterface`. I deprecate it because one should use `NormalizerAwareInterface` instead.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/composer.json
+++ b/src/Symfony/Bundle/FrameworkBundle/composer.json
@@ -93,7 +93,7 @@
         "symfony/property-info": "<6.4",
         "symfony/property-access": "<6.4",
         "symfony/scheduler": "<6.4.4|>=7.0.0,<7.0.4",
-        "symfony/serializer": "<6.4",
+        "symfony/serializer": "<7.2",
         "symfony/security-csrf": "<6.4",
         "symfony/security-core": "<6.4",
         "symfony/stopwatch": "<6.4",
--- a/src/Symfony/Bundle/FrameworkBundle/composer.json
+++ b/src/Symfony/Bundle/FrameworkBundle/composer.json
@@ -59,7 +59,7 @@
         "symfony/scheduler": "^6.4.4|^7.0.4",
         "symfony/security-bundle": "^6.4|^7.0",
         "symfony/semaphore": "^6.4|^7.0",
-        "symfony/serializer": "^6.4|^7.0",
+        "symfony/serializer": "^7.2",
         "symfony/stopwatch": "^6.4|^7.0",
         "symfony/string": "^6.4|^7.0",
         "symfony/translation": "^6.4|^7.0",
--- a/src/Symfony/Component/ErrorHandler/composer.json
+++ b/src/Symfony/Component/ErrorHandler/composer.json
@@ -22,7 +22,7 @@
     },
     "require-dev": {
         "symfony/http-kernel": "^6.4|^7.0",
-        "symfony/serializer": "^6.4|^7.0",
+        "symfony/serializer": "^7.1",
         "symfony/deprecation-contracts": "^2.5|^3"
     },
     "conflict": {
--- a/src/Symfony/Component/ErrorHandler/composer.json
+++ b/src/Symfony/Component/ErrorHandler/composer.json
@@ -22,7 +22,7 @@
     },
     "require-dev": {
         "symfony/http-kernel": "^6.4|^7.0",
-        "symfony/serializer": "^7.1",
+        "symfony/serializer": "^6.4|^7.0",
         "symfony/deprecation-contracts": "^2.5|^3"
     },
     "conflict": {
--- a/src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/RequestPayloadValueResolverTest.php
+++ b/src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/RequestPayloadValueResolverTest.php
@@ -27,6 +27,8 @@ use Symfony\Component\Serializer\Encoder\XmlEncoder;
 use Symfony\Component\Serializer\Exception\NotNormalizableValueException;
 use Symfony\Component\Serializer\Exception\PartialDenormalizationException;
 use Symfony\Component\Serializer\Normalizer\ArrayDenormalizer;
+use Symfony\Component\Serializer\Normalizer\ChainDenormalizer;
+use Symfony\Component\Serializer\Normalizer\ChainNormalizer;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 use Symfony\Component\Serializer\Normalizer\ObjectNormalizer;
 use Symfony\Component\Serializer\Serializer;
@@ -204,7 +206,10 @@ class RequestPayloadValueResolverTest extends TestCase
         }
     }
 
-    public function testWithoutValidatorAndCouldNotDenormalize()
+    /**
+     * @group legacy
+     */
+    public function testWithoutValidatorAndCouldNotDenormalizeWithLegacySerializer()
     {
         $content = '{"price": 50, "title": ["not a string"]}';
         $serializer = new Serializer([new ObjectNormalizer()], ['json' => new JsonEncoder()]);
@@ -228,10 +233,34 @@ class RequestPayloadValueResolverTest extends TestCase
         }
     }
 
+    public function testWithoutValidatorAndCouldNotDenormalize()
+    {
+        $content = '{"price": 50, "title": ["not a string"]}';
+        $serializer = new Serializer([], ['json' => new JsonEncoder()], new ObjectNormalizer(), new ObjectNormalizer());
+
+        $resolver = new RequestPayloadValueResolver($serializer);
+
+        $argument = new ArgumentMetadata('invalid', RequestPayload::class, false, false, null, false, [
+            MapRequestPayload::class => new MapRequestPayload(),
+        ]);
+        $request = Request::create('/', 'POST', server: ['CONTENT_TYPE' => 'application/json'], content: $content);
+
+        $kernel = $this->createMock(HttpKernelInterface::class);
+        $arguments = $resolver->resolve($request, $argument);
+        $event = new ControllerArgumentsEvent($kernel, function () {}, $arguments, $request, HttpKernelInterface::MAIN_REQUEST);
+
+        try {
+            $resolver->onKernelControllerArguments($event);
+            $this->fail(sprintf('Expected "%s" to be thrown.', HttpException::class));
+        } catch (HttpException $e) {
+            $this->assertInstanceOf(PartialDenormalizationException::class, $e->getPrevious());
+        }
+    }
+
     public function testValidationNotPassed()
     {
         $content = '{"price": 50, "title": ["not a string"]}';
-        $serializer = new Serializer([new ObjectNormalizer()], ['json' => new JsonEncoder()]);
+        $serializer = new Serializer([], ['json' => new JsonEncoder()], new ObjectNormalizer(), new ObjectNormalizer());
 
         $validator = $this->createMock(ValidatorInterface::class);
         $validator->expects($this->never())
@@ -262,7 +291,7 @@ class RequestPayloadValueResolverTest extends TestCase
     public function testValidationNotPerformedWhenPartialDenormalizationReturnsViolation()
     {
         $content = '{"password": "abc"}';
-        $serializer = new Serializer([new ObjectNormalizer()], ['json' => new JsonEncoder()]);
+        $serializer = new Serializer([], ['json' => new JsonEncoder()], new ObjectNormalizer(), new ObjectNormalizer());
 
         $validator = $this->createMock(ValidatorInterface::class);
         $validator->expects($this->never())
@@ -316,7 +345,7 @@ class RequestPayloadValueResolverTest extends TestCase
     {
         $content = '{"price": 50}';
         $payload = new RequestPayload(50);
-        $serializer = new Serializer([new ObjectNormalizer()], ['json' => new JsonEncoder()]);
+        $serializer = new Serializer([], ['json' => new JsonEncoder()], new ObjectNormalizer(), new ObjectNormalizer());
 
         $validator = $this->createMock(ValidatorInterface::class);
         $validator->expects($this->once())
@@ -372,7 +401,7 @@ class RequestPayloadValueResolverTest extends TestCase
         $payload = new RequestPayload(50);
         $query = ['price' => '50'];
 
-        $serializer = new Serializer([new ObjectNormalizer()], ['json' => new JsonEncoder()]);
+        $serializer = new Serializer([], ['json' => new JsonEncoder()], new ObjectNormalizer(), new ObjectNormalizer());
 
         $validator = $this->createMock(ValidatorInterface::class);
         $validator->expects($this->once())
@@ -400,7 +429,7 @@ class RequestPayloadValueResolverTest extends TestCase
         $input = ['price' => '50'];
         $payload = new RequestPayload(50);
 
-        $serializer = new Serializer([new ObjectNormalizer()], ['json' => new JsonEncoder()]);
+        $serializer = new Serializer([], ['json' => new JsonEncoder()], new ObjectNormalizer(), new ObjectNormalizer());
 
         $validator = $this->createMock(ValidatorInterface::class);
         $validator->expects($this->once())
@@ -434,7 +463,7 @@ class RequestPayloadValueResolverTest extends TestCase
             new RequestPayload(23),
         ];
 
-        $serializer = new Serializer([new ArrayDenormalizer(), new ObjectNormalizer()], ['json' => new JsonEncoder()]);
+        $serializer = new Serializer([], ['json' => new JsonEncoder()], null, new ChainDenormalizer([new ArrayDenormalizer(), new ObjectNormalizer()]));
 
         $validator = $this->createMock(ValidatorInterface::class);
         $validator->expects($this->once())
@@ -513,7 +542,7 @@ class RequestPayloadValueResolverTest extends TestCase
     public function testAcceptFormatPassed(mixed $acceptFormat, string $contentType, string $content)
     {
         $encoders = ['json' => new JsonEncoder(), 'xml' => new XmlEncoder()];
-        $serializer = new Serializer([new ObjectNormalizer()], $encoders);
+        $serializer = new Serializer([], $encoders, new ObjectNormalizer(), new ObjectNormalizer());
         $validator = (new ValidatorBuilder())->getValidator();
         $resolver = new RequestPayloadValueResolver($serializer, $validator);
 
@@ -576,7 +605,7 @@ class RequestPayloadValueResolverTest extends TestCase
      */
     public function testAcceptFormatNotPassed(mixed $acceptFormat, string $contentType, string $content, string $expectedExceptionMessage)
     {
-        $serializer = new Serializer([new ObjectNormalizer()]);
+        $serializer = new Serializer([], [], new ObjectNormalizer(), new ObjectNormalizer());
         $validator = (new ValidatorBuilder())->getValidator();
         $resolver = new RequestPayloadValueResolver($serializer, $validator);
 
@@ -647,7 +676,7 @@ class RequestPayloadValueResolverTest extends TestCase
         $payload = new RequestPayload(50);
         $payload->title = 'A long title, so the validation passes';
 
-        $serializer = new Serializer([new ObjectNormalizer()]);
+        $serializer = new Serializer([], [], new ObjectNormalizer(), new ObjectNormalizer());
         $validator = (new ValidatorBuilder())->enableAttributeMapping()->getValidator();
         $resolver = new RequestPayloadValueResolver($serializer, $validator);
 
@@ -673,7 +702,7 @@ class RequestPayloadValueResolverTest extends TestCase
     {
         $input = ['price' => '50', 'title' => 'Too short'];
 
-        $serializer = new Serializer([new ObjectNormalizer()]);
+        $serializer = new Serializer([], [], new ObjectNormalizer(), new ObjectNormalizer());
         $validator = (new ValidatorBuilder())->enableAttributeMapping()->getValidator();
         $resolver = new RequestPayloadValueResolver($serializer, $validator);
 
@@ -732,7 +761,7 @@ class RequestPayloadValueResolverTest extends TestCase
 
     public function testQueryValidationErrorCustomStatusCode()
     {
-        $serializer = new Serializer([new ObjectNormalizer()], []);
+        $serializer = new Serializer([], [], new ObjectNormalizer(), new ObjectNormalizer());
 
         $validator = $this->createMock(ValidatorInterface::class);
 
@@ -765,7 +794,7 @@ class RequestPayloadValueResolverTest extends TestCase
     public function testRequestPayloadValidationErrorCustomStatusCode()
     {
         $content = '{"price": 50, "title": ["not a string"]}';
-        $serializer = new Serializer([new ObjectNormalizer()], ['json' => new JsonEncoder()]);
+        $serializer = new Serializer([], ['json' => new JsonEncoder()], new ObjectNormalizer(), new ObjectNormalizer());
 
         $validator = $this->createMock(ValidatorInterface::class);
         $validator->expects($this->never())
@@ -798,7 +827,7 @@ class RequestPayloadValueResolverTest extends TestCase
      */
     public function testBoolArgumentInQueryString(mixed $expectedValue, ?string $parameterValue)
     {
-        $serializer = new Serializer([new ObjectNormalizer()]);
+        $serializer = new Serializer([], [], new ObjectNormalizer(), new ObjectNormalizer());
         $validator = $this->createMock(ValidatorInterface::class);
         $resolver = new RequestPayloadValueResolver($serializer, $validator);
 
@@ -821,7 +850,7 @@ class RequestPayloadValueResolverTest extends TestCase
      */
     public function testBoolArgumentInBody(mixed $expectedValue, ?string $parameterValue)
     {
-        $serializer = new Serializer([new ObjectNormalizer()]);
+        $serializer = new Serializer([], [], new ObjectNormalizer(), new ObjectNormalizer());
         $validator = $this->createMock(ValidatorInterface::class);
         $resolver = new RequestPayloadValueResolver($serializer, $validator);
 
@@ -857,7 +886,7 @@ class RequestPayloadValueResolverTest extends TestCase
      */
     public function testBoolArgumentInJsonBody()
     {
-        $serializer = new Serializer([new ObjectNormalizer()]);
+        $serializer = new Serializer([], [], new ObjectNormalizer(), new ObjectNormalizer());
         $validator = $this->createMock(ValidatorInterface::class);
         $resolver = new RequestPayloadValueResolver($serializer, $validator);
 
--- a/src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/RequestPayloadValueResolverTest.php
+++ b/src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/RequestPayloadValueResolverTest.php
@@ -28,7 +28,6 @@ use Symfony\Component\Serializer\Exception\NotNormalizableValueException;
 use Symfony\Component\Serializer\Exception\PartialDenormalizationException;
 use Symfony\Component\Serializer\Normalizer\ArrayDenormalizer;
 use Symfony\Component\Serializer\Normalizer\ChainDenormalizer;
-use Symfony\Component\Serializer\Normalizer\ChainNormalizer;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 use Symfony\Component\Serializer\Normalizer\ObjectNormalizer;
 use Symfony\Component\Serializer\Serializer;
--- a/src/Symfony/Component/Serializer/CHANGELOG.md
+++ b/src/Symfony/Component/Serializer/CHANGELOG.md
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+7.2
+---
+
+ * Add `ChainNormalizer` and `ChainDenormalizer`
+ * Add two more constructor arguments for `Serializer` to accept a `NormalizerInterface` and `DenormalizerInterface`
+
 7.1
 ---
 
--- a/src/Symfony/Component/Serializer/Debug/TraceableEncoder.php
+++ b/src/Symfony/Component/Serializer/Debug/TraceableEncoder.php
@@ -15,6 +15,8 @@ use Symfony\Component\Serializer\DataCollector\SerializerDataCollector;
 use Symfony\Component\Serializer\Encoder\DecoderInterface;
 use Symfony\Component\Serializer\Encoder\EncoderInterface;
 use Symfony\Component\Serializer\Encoder\NormalizationAwareInterface;
+use Symfony\Component\Serializer\Normalizer\DenormalizerAwareInterface;
+use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerAwareInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 use Symfony\Component\Serializer\SerializerAwareInterface;
@@ -27,7 +29,7 @@ use Symfony\Component\Serializer\SerializerInterface;
  *
  * @internal
  */
-class TraceableEncoder implements EncoderInterface, DecoderInterface, SerializerAwareInterface, NormalizerAwareInterface
+class TraceableEncoder implements EncoderInterface, DecoderInterface, SerializerAwareInterface, NormalizerAwareInterface, DenormalizerAwareInterface
 {
     public function __construct(
         private EncoderInterface|DecoderInterface $encoder,
@@ -103,6 +105,16 @@ class TraceableEncoder implements EncoderInterface, DecoderInterface, Serializer
         }
 
         $this->encoder->setNormalizer($normalizer);
+    }
+
+
+    public function setDenormalizer(DenormalizerInterface $denormalizer): void
+    {
+        if (!$this->encoder instanceof DenormalizerInterface) {
+            return;
+        }
+
+        $this->encoder->setDenormalizer($denormalizer);
     }
 
     public function needsNormalization(): bool
--- a/src/Symfony/Component/Serializer/Debug/TraceableEncoder.php
+++ b/src/Symfony/Component/Serializer/Debug/TraceableEncoder.php
@@ -15,6 +15,8 @@ use Symfony\Component\Serializer\DataCollector\SerializerDataCollector;
 use Symfony\Component\Serializer\Encoder\DecoderInterface;
 use Symfony\Component\Serializer\Encoder\EncoderInterface;
 use Symfony\Component\Serializer\Encoder\NormalizationAwareInterface;
+use Symfony\Component\Serializer\Normalizer\NormalizerAwareInterface;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 use Symfony\Component\Serializer\SerializerAwareInterface;
 use Symfony\Component\Serializer\SerializerInterface;
 
@@ -25,7 +27,7 @@ use Symfony\Component\Serializer\SerializerInterface;
  *
  * @internal
  */
-class TraceableEncoder implements EncoderInterface, DecoderInterface, SerializerAwareInterface
+class TraceableEncoder implements EncoderInterface, DecoderInterface, SerializerAwareInterface, NormalizerAwareInterface
 {
     public function __construct(
         private EncoderInterface|DecoderInterface $encoder,
@@ -92,6 +94,15 @@ class TraceableEncoder implements EncoderInterface, DecoderInterface, Serializer
         }
 
         $this->encoder->setSerializer($serializer);
+    }
+
+    public function setNormalizer(NormalizerInterface $normalizer): void
+    {
+        if (!$this->encoder instanceof NormalizerAwareInterface) {
+            return;
+        }
+
+        $this->encoder->setNormalizer($normalizer);
     }
 
     public function needsNormalization(): bool
--- a/src/Symfony/Component/Serializer/DependencyInjection/SerializerPass.php
+++ b/src/Symfony/Component/Serializer/DependencyInjection/SerializerPass.php
@@ -40,17 +40,16 @@ class SerializerPass implements CompilerPassInterface
             return;
         }
 
-        if (!$normalizers = $this->findAndSortTaggedServices('serializer.normalizer', $container)) {
-            throw new RuntimeException('You must tag at least one service as "serializer.normalizer" to use the "serializer" service.');
-        }
-
         if (!$encoders = $this->findAndSortTaggedServices('serializer.encoder', $container)) {
             throw new RuntimeException('You must tag at least one service as "serializer.encoder" to use the "serializer" service.');
         }
 
+        $normalizers = $this->findAndSortTaggedServices('serializer.normalizer', $container);
+        $denormalizers = $this->findAndSortTaggedServices('serializer.denormalizer', $container);
+
         if ($container->hasParameter('serializer.default_context')) {
             $defaultContext = $container->getParameter('serializer.default_context');
-            foreach (array_merge($normalizers, $encoders) as $service) {
+            foreach (array_merge($normalizers, $denormalizers, $encoders) as $service) {
                 $definition = $container->getDefinition($service);
                 $definition->setBindings(['array $defaultContext' => new BoundArgument($defaultContext, false)] + $definition->getBindings());
             }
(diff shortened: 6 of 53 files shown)


Nyholm:
Hm.. Im considering renaming this to `NormalizerAggregate` instead. Similar to `CacheWarmerAggregate`.

EDIT: 
Changed my mind. We do have `ChainEncoder` and `ChainDecoder` at the moment. Using "Chain" is good here.

src/Symfony/Component/Serializer/Debug/TraceableEncoder.php
```diff
@@ -25,7 +27,7 @@
  *
  * @internal
  */
-class TraceableEncoder implements EncoderInterface, DecoderInterface, SerializerAwareInterface
+class TraceableEncoder implements EncoderInterface, DecoderInterface, SerializerAwareInterface, NormalizerAwareInterface
```
mtarld: We can remove the `setSerializer` method

src/Symfony/Component/Serializer/DependencyInjection/SerializerPass.php
```diff
@@ -67,8 +70,37 @@ public function process(ContainerBuilder $container): void
...
+            if (isset($interfaces[DenormalizerInterface::class])) {
+                $denormalizers[] = $reference ?? $definition;
+            }
+        }
+
         $serializerDefinition = $container->getDefinition('serializer');
-        $serializerDefinition->replaceArgument(0, $normalizers);
         $serializerDefinition->replaceArgument(1, $encoders);
+
+        if ($container->hasDefinition('serializer.normalizer') && $container->hasDefinition('serializer.normalizer')) {
```
mtarld:
```suggestion
        if ($container->hasDefinition('serializer.normalizer') && $container->hasDefinition('serializer.denormalizer')) {
```

Plus, I think it'll always have these definition (so the `else` wouldn't be triggered at all)

src/Symfony/Component/Serializer/Encoder/XmlEncoder.php
```diff
@@ -23,8 +25,9 @@
  * @author Kévin Dunglas <dunglas@gmail.com>
  * @author Dany Maillard <danymaillard93b@gmail.com>
  */
-class XmlEncoder implements EncoderInterface, DecoderInterface, NormalizationAwareInterface, SerializerAwareInterface
+class XmlEncoder implements EncoderInterface, DecoderInterface, NormalizationAwareInterface, SerializerAwareInterface, NormalizerAwareInterface
 {
+    use NormalizerAwareTrait;
     use SerializerAwareTrait;
```
mtarld, line 31: We might deprecate the `XmlEncoder::setSerializer` method

src/Symfony/Component/Serializer/Serializer.php
```diff
@@ -51,46 +51,42 @@ class Serializer implements SerializerInterface, NormalizerInterface, Denormaliz
...
-
     /**
      * @param array<NormalizerInterface|DenormalizerInterface> $normalizers
      * @param array<EncoderInterface|DecoderInterface>         $encoders
      */
     public function __construct(
-        private array $normalizers = [],
+        array $normalizers = [],
         array $encoders = [],
+        ChainNormalizerInterface $normalizerAggregator = null,
```
mtarld: I understand the fact of adding optional arguments for the upgrade path, but what's the next step? How to get rid of the two first arguments?

src/Symfony/Component/Serializer/Normalizer/ChainDenormalizerInterface.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    /**
+     * Add a Denormalizer last in the line.
+     */
+    public function addDenormalizer(DenormalizerInterface $denormalizer): void;
+
+    /**
+     * Add a Denormalizer first in line.
+     */
+    public function prependDenormalizer(DenormalizerInterface $denormalizer): void;
```
mtarld: Same here, why is this useful?

src/Symfony/Component/Serializer/Normalizer/ChainNormalizerInterface.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    /**
+     * Add a Normalizer last in the line.
+     */
+    public function addNormalizer(NormalizerInterface $normalizer): void;
+
+    /**
+     * Add a Normalizer first in line.
+     */
+    public function prependNormalizer(NormalizerInterface $normalizer): void;
```
mtarld: I don't get the purpose of that method

src/Symfony/Component/Serializer/Serializer.php
mtarld: Does it make sense to specify `NormalizerInterface` instead?

IMHO, at some point, the very best signature of that constructor should be:
```php
public function __construct(
  NormalizerInterface $normalizer,
  DenormalizerInterface $denormalizer,
  EncoderInterface $encoder,
  DecodeInterface $decoder,
) {}
```

Indeed, I do think that the fact that normalizers or encoders are "chaining" should not be known by the serializer.

By doing that, we could get rid of the `ChainNormalizerInterface` and `ChainDenormalizerInterface`.
And we can harmonize the normalization and the encoding.

src/Symfony/Component/Serializer/Tests/SerializerTest.php
```diff
@@ -76,6 +78,9 @@
 
 class SerializerTest extends TestCase
 {
+    /**
+     * @legacy
```
mtarld: Is it really legacy?

src/Symfony/Component/Serializer/SerializerAwareTrait.php
```diff
@@ -12,6 +12,8 @@
 namespace Symfony\Component\Serializer;
 
 /**
+ * @deprecated Use NormalizerAwareTrait instead
```
mtarld: I won't suggest any replacement as the normalizer doesn't fulfill all the serializer features

src/Symfony/Component/Serializer/SerializerAwareInterface.php
```diff
@@ -12,6 +12,8 @@
 namespace Symfony\Component\Serializer;
 
 /**
+ * @deprecated Use NormalizerAwareInterface or DenormalizerAwareInterface instead
```
mtarld: Same here, I won't suggest anything

src/Symfony/Component/Serializer/DependencyInjection/SerializerPass.php
```diff
@@ -67,8 +70,37 @@ public function process(ContainerBuilder $container): void
             }
         }
 
+        $denormalizers = [];
+        $trueNormalizers = [];
+        /** @var Reference|Definition $definition */
```
mtarld: According to `PriorityTaggedServiceTrait`, `$normalizers` is a list of `Reference`

src/Symfony/Component/Serializer/Normalizer/ChainDenormalizer.php
```diff
@@ -0,0 +1,242 @@
...
+{
+    private const SCALAR_TYPES = [
+        'int' => true,
+        'bool' => true,
+        'float' => true,
+        'string' => true,
+    ];
+
+    /**
+     * @deprecated since Symfony 7.1
```
mtarld: I don't think it's needed as the property is private

src/Symfony/Component/Serializer/Normalizer/ChainDenormalizer.php
```diff
@@ -0,0 +1,242 @@
...
+use Symfony\Component\Serializer\SerializerAwareInterface;
+use Symfony\Component\Serializer\SerializerInterface;
+
+/**
+ * This holds a collection of denormalizers. It tries to be smart how it selects
+ * the correct Denormalizer.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class ChainDenormalizer implements ChainDenormalizerInterface, SerializerAwareInterface
```
mtarld: Is it really needed to use that `SerializerAwareInterface` deprecated interface?

src/Symfony/Component/Serializer/Normalizer/ChainDenormalizer.php
```diff
@@ -0,0 +1,242 @@
...
+            foreach ($this->denormalizers as $denormalizer) {
+                $this->supportedCache + $denormalizer->getSupportedTypes($format);
+            }
+        }
+
+        return $this->supportedCache[$format];
+    }
+
+    /**
+     * This method exists only for BC reasons. Will be removed in 8.0.
```
mtarld, line 189: Should it be tagged as internal?

src/Symfony/Component/Serializer/Serializer.php
```diff
@@ -51,46 +51,42 @@ class Serializer implements SerializerInterface, NormalizerInterface, Denormaliz
...
+        ChainDenormalizerInterface $denormalizerAggregator = null,
```
mtarld:
```suggestion
        ChainDenormalizerInterface $denormalizerAggregator = new ChainDenormalizer(),
```

src/Symfony/Component/Serializer/Serializer.php
```diff
@@ -102,8 +98,12 @@ public function __construct(
         $realEncoders = [];
         foreach ($encoders as $encoder) {
             if ($encoder instanceof SerializerAwareInterface) {
+                trigger_deprecation('symfony/serializer', '7.1', 'Interface %s is deprecated,  use %s instead.', SerializerAwareInterface::class, NormalizerAwareInterface::class);
```
mtarld: What about moving that `trigger_deprecation` to the `SerializerAwareInterface` instead?

mtarld (commented):
I really think that's a good idea to have a better separation between normalization and serialization, thanks for pushing it forward!

src/Symfony/Component/Serializer/Normalizer/ChainDenormalizer.php
chalasr: at least this allows to not forget about removing the property when we'll start working on 8.0

src/Symfony/Component/Serializer/Normalizer/ChainDenormalizer.php
chalasr: also `@deprecated` to quickly spot it when it's time to

src/Symfony/Component/Serializer/Serializer.php
chalasr: I tend to agree that the `chain` aspect should remain an implementation detail

src/Symfony/Component/Serializer/Serializer.php
chalasr: That would first require to completely stop using it in core

src/Symfony/Component/Serializer/Debug/TraceableEncoder.php
Nyholm: I need to keep it around to not break existing code. Even though it is internal, but if I remove `setSerializer` you will miss some data in your WebProfiler page.

src/Symfony/Component/Serializer/DependencyInjection/SerializerPass.php
Nyholm: Thank you. 

If you are using Serializer 7.1 and FrameworkBundle 7.0 the else will be triggered.

src/Symfony/Component/Serializer/Encoder/XmlEncoder.php
Nyholm: Yes. I have deprecated the trait, but I have not yet found out how to trigger_error.

src/Symfony/Component/Serializer/Serializer.php
Nyholm: I would like all users to pass an empty array as first argument. Then at Symfony 8.0 we will just drop the first argument.

src/Symfony/Component/Serializer/Normalizer/ChainNormalizerInterface.php
Nyholm: They were for altering the service at runtime "in a smart way". They are never used and I should remove them.

Thank you

src/Symfony/Component/Serializer/Serializer.php
Nyholm: Okey. Cool. I'll try to see if it make sense.

src/Symfony/Component/Serializer/Tests/SerializerTest.php
Nyholm: I have to mark the test as legacy because it will generate a deprecation warning. 

Ie it runs `new Serializer([new \stdClass()]);` which is deprecated.

src/Symfony/Component/Serializer/Normalizer/ChainDenormalizer.php
Nyholm: Yes... sadly it is.

src/Symfony/Component/Serializer/Serializer.php
Nyholm: Yeah.. and we cannot really do that in a good way while keep things working. 

I could stop using the trait but add `setSerializer()` on the classes that uses the trait. But.. That does not help with the deprecation of the interface. 

--------

What I ended up doing: I am checking if the encoder also implements NormalizerInterface. If so, we dont trigger deprecation.

src/Symfony/Component/Serializer/Serializer.php
Nyholm: It did!

src/Symfony/Component/Serializer/SerializerAwareTrait.php
Nyholm: True... But all the usages of `SerializerAwareTrait` in core has been used for normalizing. I'll remove it.

src/Symfony/Component/Serializer/Serializer.php
Nyholm: That is a good idea. But I cannot do this now when we removed `ChainDenormalizerInterface` interface. (I need to throw exception if this is not null and you also pass an array of $normalizers.

src/Symfony/Component/Serializer/DependencyInjection/SerializerPass.php
Nyholm: Yes. But we also add more items to this array. See around line 60. 

```php
foreach ($normalizers as $i => $normalizer) {
   $normalizers[$i] = $container->register('.debug.serializer.normalizer.'.$normalizer, TraceableNormalizer::class)
        ->setArguments([$normalizer, new Reference('serializer.data_collector')]);
}
```

Nyholm:
Wohoo. All tests are passing.

While working on this PR I found an issues like #52992. Ie some code assume a certain behaviour without any contract. 
One example that we have in 7.0 right now is that you can pass `$context[DenormalizerInterface::COLLECT_DENORMALIZATION_ERRORS]` to `Serializer::denormalize()` and never get a `PartialDenormalizationException` because you have used the wrong normalizers. 

This is something we should fix and this PR is a tiny step towards that direction.

src/Symfony/Component/Serializer/Serializer.php
mtarld: Does the first argument drop will be possible from a BC point of view?

Nyholm: No. That is why we need to do it in SF 8.0.

Nyholm:
PR is rebased and all tests are green again.

Nyholm:
PR is rebased =)

Nyholm:
@mtarld and @chalasr Can you give another review on this please?

src/Symfony/Bundle/FrameworkBundle/composer.json
```diff
@@ -59,7 +59,7 @@
         "symfony/scheduler": "^6.4.4|^7.0.4",
         "symfony/security-bundle": "^6.4|^7.0",
         "symfony/semaphore": "^6.4|^7.0",
-        "symfony/serializer": "^6.4|^7.0",
+        "symfony/serializer": "^7.1",
```
mtarld: Is that bump really needed? AFAIK, we try to avoid as much as possible to bump dependencies (see https://github.com/symfony/symfony/pull/53160#discussion_r1451510668)

(same for other `composer.json` files)

src/Symfony/Component/Serializer/Serializer.php
```diff
@@ -51,59 +50,70 @@ class Serializer implements SerializerInterface, NormalizerInterface, Denormaliz
...
     /**
      * @param array<NormalizerInterface|DenormalizerInterface> $normalizers
      * @param array<EncoderInterface|DecoderInterface>         $encoders
      */
     public function __construct(
-        private array $normalizers = [],
+        array $normalizers = [],
         array $encoders = [],
+        NormalizerInterface $normalizer = null,
+        DenormalizerInterface $denormalizer = null,
```
mtarld: Instantiating new classes (and doing method calls) can be costly.
I'm just wondering if it'll be an issue in terms of performances (but I think it will be ok as it isn't called that much)

src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/RequestPayloadValueResolverTest.php
```diff
@@ -695,7 +724,7 @@ public function testQueryValidationErrorCustomStatusCode()
     public function testRequestPayloadValidationErrorCustomStatusCode()
     {
         $content = '{"price": 50, "title": ["not a string"]}';
-        $serializer = new Serializer([new ObjectNormalizer()], ['json' => new JsonEncoder()]);
+        $serializer = new Serializer([], ['json' => new JsonEncoder()], new ChainNormalizer([new ObjectNormalizer()]), new ChainDenormalizer([new ObjectNormalizer()]));
```
mtarld:
```suggestion
        $serializer = new Serializer([], ['json' => new JsonEncoder()], new ObjectNormalizer(), new ObjectNormalizer());
```

src/Symfony/Component/Serializer/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ CHANGELOG
 7.1
 ---
 
+ * Add `ChainNormalizer` and `ChainDenormalizer`
+ * Add two more constructor arguments for `Serializer` to accept a `NormalizerInterface` and `DenormalizerInterface`
```
mtarld, line 8: The deprecation has to be added (and `UPGRADE-7.1.md` to be updated as well)

src/Symfony/Component/Serializer/Normalizer/ChainDenormalizer.php
```diff
@@ -0,0 +1,207 @@
...
+use Symfony\Component\Serializer\SerializerAwareInterface;
+use Symfony\Component\Serializer\SerializerInterface;
+
+/**
+ * This holds a collection of denormalizers. It tries to be smart how it selects
+ * the correct Denormalizer.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class ChainDenormalizer implements DenormalizerInterface, SerializerAwareInterface
```
mtarld: I think that it can be final (Same for `ChainNormalizer`)

src/Symfony/Component/Serializer/Serializer.php
mtarld:
```suggestion
        ?NormalizerInterface $normalizer = null,
        ?DenormalizerInterface $denormalizer = null,
```
To fit with the actual CS

src/Symfony/Component/Serializer/SerializerAwareInterface.php
```diff
@@ -12,6 +12,8 @@
 namespace Symfony\Component\Serializer;
 
 /**
+ * @deprecated
```
mtarld:
```suggestion
 * @deprecated since Symfony 7.1, use {@see NormalizerAwareInterface} instead
```

src/Symfony/Component/Serializer/SerializerAwareTrait.php
```diff
@@ -12,6 +12,8 @@
 namespace Symfony\Component\Serializer;
 
 /**
+ * @deprecated
```
mtarld:
```suggestion
 * @deprecated since Symfony 7.1
```

src/Symfony/Component/Serializer/Tests/Encoder/XmlEncoderTest.php
```diff
@@ -35,8 +36,8 @@ class XmlEncoderTest extends TestCase
     protected function setUp(): void
     {
         $this->encoder = new XmlEncoder();
-        $serializer = new Serializer([new CustomNormalizer()], ['xml' => new XmlEncoder()]);
-        $this->encoder->setSerializer($serializer);
+        $normalizer = new ChainNormalizer([new CustomNormalizer()]);
+        $serializer = new Serializer([], ['xml' => $this->encoder], $normalizer);
```
mtarld, line 40:
```suggestion
        $serializer = new Serializer([], ['xml' => $this->encoder], new CustomNormalizer());
```
Same for other occurrences

src/Symfony/Component/Serializer/Tests/Normalizer/AbstractNormalizerTest.php
```diff
@@ -191,8 +193,7 @@ public function testObjectWithVariadicConstructorTypedArguments(AbstractNormaliz
         $d2->qux = 'QUZ';
         $obj = new VariadicConstructorTypedArgsDummy($d1, $d2);
 
-        $serializer = new Serializer([$normalizer], [new JsonEncoder()]);
-        $normalizer->setSerializer($serializer);
+        $serializer = new Serializer([], [new JsonEncoder()], new ChainNormalizer([$normalizer]), new ChainDenormalizer([$normalizer]));
```
mtarld, line 196:
```suggestion
        $serializer = new Serializer([], [new JsonEncoder()], $normalizer, $normalizer);
```
Same for other occurrences

src/Symfony/Component/Serializer/Tests/SerializerTest.php
```diff
@@ -1352,15 +1347,68 @@ public function testCollectDenormalizationErrorsWithConstructor(?ClassMetadataFa
         $this->assertSame($expected, $exceptionsAsArray);
     }
 
+    /**
+     * @group legacy
+     * If we use the Symfony 7.0 way to create a Serializer, will we still get a PartialDenormalizationException?
```
mtarld:
```suggestion
     * @group legacy
     *
     * If we use the Symfony 7.0 way to create a Serializer, will we still get a PartialDenormalizationException?
```

src/Symfony/Component/Serializer/Tests/SerializerTest.php
```diff
@@ -80,6 +82,9 @@
 
 class SerializerTest extends TestCase
 {
+    /**
+     * @group legacy
+     */
```
mtarld, line 87: I think that we can add the non-legacy version as well

mtarld (changes requested):

dunglas:
It's a good step forward. I'm +1 for the idea.

src/Symfony/Component/Serializer/Serializer.php
Nyholm: I dont think this will be an issue at all. It is called once per request cycle. Just like every other service.

src/Symfony/Component/ErrorHandler/composer.json
```diff
@@ -22,7 +22,7 @@
     },
     "require-dev": {
         "symfony/http-kernel": "^6.4|^7.0",
-        "symfony/serializer": "^6.4|^7.0",
+        "symfony/serializer": "^7.1",
```
yceruto: `^7.2`

src/Symfony/Component/Serializer/Normalizer/ChainDenormalizer.php
```diff
@@ -0,0 +1,207 @@
...
+{
+    private const SCALAR_TYPES = [
+        'int' => true,
+        'bool' => true,
+        'float' => true,
+        'string' => true,
+    ];
+
+    /**
+     * @deprecated since Symfony 7.1
```
yceruto: now `7.2`

src/Symfony/Component/Serializer/Normalizer/ChainDenormalizer.php
```diff
@@ -0,0 +1,207 @@
...
+    /**
+     * @throws NotNormalizableValueException
+     */
+    public function denormalize(mixed $data, string $type, ?string $format = null, array $context = []): mixed
+    {
+        if (isset($context[DenormalizerInterface::COLLECT_DENORMALIZATION_ERRORS], $context['not_normalizable_value_exceptions'])) {
+            throw new LogicException('Passing a value for "not_normalizable_value_exceptions" context key is not allowed.');
+        }
+
+        $denormalizer = $this->getDenormalizer($data, $type, $format, $context);
```
yceruto, line 88: There’s nothing to do here, I suppose, but it seems odd that we don’t first check if it’s a scalar value/type and skip this process earlier, knowing that denormalizers are intended to work with classes/objects only, or at least that’s what the `getDenormalizer` implementation suggests.

src/Symfony/Component/Serializer/Normalizer/ChainDenormalizer.php
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        return $denormalizer->denormalize($data, $type, $format, $context);
+    }
+
+    /**
+     * Returns a matching denormalizer.
+     *
+     * @param mixed       $data    Data to restore
+     * @param string      $class   The expected class to instantiate
```
yceruto: Related to my previous comment, this is not true if a scalar type is expected here.

src/Symfony/Component/Serializer/Serializer.php
```diff
@@ -51,59 +50,70 @@ class Serializer implements SerializerInterface, NormalizerInterface, Denormaliz
...
+        array $normalizers = [],
         array $encoders = [],
+        ?NormalizerInterface $normalizer = null,
+        ?DenormalizerInterface $denormalizer = null,
     ) {
-        foreach ($normalizers as $normalizer) {
-            if ($normalizer instanceof SerializerAwareInterface) {
-                $normalizer->setSerializer($this);
+        if ([] !== $normalizers && (null !== $normalizer || null !== $denormalizer)) {
+            throw new InvalidArgumentException('You cannot use an array of $normalizers with a $normalizer/$denormalizer.');
```
yceruto: I’d add a clear suggestion for the correct usage at the end:
```
You cannot use an array of $normalizers with a $normalizer/$denormalizer. Please use the $normalizer/$denormalizer arguments only instead.
```

src/Symfony/Component/Serializer/Serializer.php
```diff
@@ -51,59 +50,70 @@ class Serializer implements SerializerInterface, NormalizerInterface, Denormaliz
...
+        }
+
+        $this->normalizer = $normalizer ?? new ChainNormalizer();
+        $this->denormalizer = $denormalizer ?? new ChainDenormalizer();
+
+        if ([] !== $normalizers) {
+            trigger_deprecation('symfony/serializer', '7.1', 'Passing normalizers as first argument to "%s" is deprecated, use a chain normalizer/denormalizer instead.', __METHOD__);
```
yceruto: `7.2`

src/Symfony/Component/Serializer/Serializer.php
```diff
@@ -51,59 +50,70 @@ class Serializer implements SerializerInterface, NormalizerInterface, Denormaliz
...
+        }
+
+        foreach ($normalizers as $item) {
+            if ($item instanceof SerializerAwareInterface) {
+                if (!$item instanceof NormalizerAwareInterface) {
+                    // We assume they have started to migrate
+                    trigger_deprecation('symfony/serializer', '7.1', 'Interface %s is deprecated, maybe you are interested to use %s instead.', SerializerAwareInterface::class, NormalizerAwareInterface::class);
```
yceruto: `7.2`

yceruto: To me, it's clear that they should use it, so I'd go directly with `Interface %s is deprecated, use %s instead.`

src/Symfony/Component/Serializer/Serializer.php
```diff
@@ -51,59 +50,70 @@ class Serializer implements SerializerInterface, NormalizerInterface, Denormaliz
...
             }
         }
 
         $decoders = [];
         $realEncoders = [];
         foreach ($encoders as $encoder) {
             if ($encoder instanceof SerializerAwareInterface) {
+                if (!$encoder instanceof NormalizerAwareInterface) {
+                    // We assume they have started to migrate
+                    trigger_deprecation('symfony/serializer', '7.1', 'Interface %s is deprecated, maybe you are interested to use %s instead.', SerializerAwareInterface::class, NormalizerAwareInterface::class);
```
yceruto: 7.2

yceruto (commented):
Nice refactoring! I left minor comments only.

src/Symfony/Component/Serializer/Normalizer/ChainDenormalizer.php
Nyholm: Yes. You are very correct. This logic is just moved from the `Serializer` class. 

I have an idea for the next PR, This should be moved to a new `ScalarDenormalizer`. If we that in before 7.2 then there be no issues to keep BC. 

I'll do it in next PR to keep the "number of new things" to a minimum.

src/Symfony/Bundle/FrameworkBundle/composer.json
stof: FrameworkBundle often bumps the min version instead of trying to support multiple versions of the service definitions, for simplicity.
This is different for other package, which can be reused outside the fullstack framework and for which supporting compatibility with 2 different major versions of Symfony reduces the risk of dependency hell in those projects (the lower level the package is, the more useful this is).

Nyholm: Thank you. 

I've bumped version on FrameworkBundle and HttpKernel. I have reverted the other packages and make sure they (still) support both 6.4 and 7.2.

src/Symfony/Component/Serializer/Serializer.php
```diff
@@ -51,59 +50,70 @@ class Serializer implements SerializerInterface, NormalizerInterface, Denormaliz
...
-     */
-    private array $normalizerCache = [];
-
     /**
      * @param array<NormalizerInterface|DenormalizerInterface> $normalizers
      * @param array<EncoderInterface|DecoderInterface>         $encoders
      */
     public function __construct(
-        private array $normalizers = [],
+        array $normalizers = [],
```
stof, line 67: defining the new signature of the constructor as being `new Serializer([], $encoders, $chainNormalizer, $chainDenormalizer)` looks weird to me. It don't think the array as first argument should be kept in the new signature

src/Symfony/Component/Serializer/Serializer.php
```diff
@@ -51,59 +50,70 @@ class Serializer implements SerializerInterface, NormalizerInterface, Denormaliz
...
         $realEncoders = [];
         foreach ($encoders as $encoder) {
             if ($encoder instanceof SerializerAwareInterface) {
+                if (!$encoder instanceof NormalizerAwareInterface) {
+                    // We assume they have started to migrate
+                    trigger_deprecation('symfony/serializer', '7.2', 'Interface %s is deprecated, use %s instead.', SerializerAwareInterface::class, NormalizerAwareInterface::class);
+                }
                 $encoder->setSerializer($this);
             }
+            if ($encoder instanceof NormalizerAwareInterface) {
```
stof, line 116: what about DenormalizerAwareInterface in case they need the Denormalizer instead (which is much more likely for decoders for instance)

src/Symfony/Component/Serializer/Serializer.php
```diff
@@ -51,59 +50,70 @@ class Serializer implements SerializerInterface, NormalizerInterface, Denormaliz
...
+                throw new InvalidArgumentException(sprintf('The class "%s" neither implements "%s" nor "%s".', get_debug_type($item), NormalizerInterface::class, DenormalizerInterface::class));
             }
         }
 
         $decoders = [];
         $realEncoders = [];
         foreach ($encoders as $encoder) {
             if ($encoder instanceof SerializerAwareInterface) {
+                if (!$encoder instanceof NormalizerAwareInterface) {
+                    // We assume they have started to migrate
```
stof: they could migrate to DenormalizerAwareInterface as well

src/Symfony/Component/Serializer/SerializerAwareInterface.php
```diff
@@ -12,6 +12,8 @@
 namespace Symfony\Component\Serializer;
 
 /**
+ * @deprecated since Symfony 7.2, use {@see NormalizerAwareInterface} instead
```
stof:
```suggestion
 * @deprecated since Symfony 7.2, use {@see NormalizerAwareInterface} or {@see DenormalizerAwareInterface} instead
```

src/Symfony/Component/Serializer/Serializer.php
stof: alternatively, we could keep the current constructor signature, making the ChainNormalizer and ChainDenormalizer internal implementation details, as done for the ChainEncoder and ChainDecoder (for which we classify them as being encoders, decoders or both inside the constructor). As most implementations are implementing both interfaces in the same class, it allows passing a single list in instantiation, making the configuration easier

Btw, this would mean that the chain classes would not need to implement the `*aware` interfaces (as the injection would be handled by the Serializer on each registered class)

src/Symfony/Component/Serializer/DependencyInjection/SerializerPass.php
```diff
@@ -67,8 +70,37 @@ public function process(ContainerBuilder $container): void
...
+        foreach ($normalizers as $definition) {
+            $reference = null;
+            if ($definition instanceof Reference) {
+                $reference = $definition;
+                $definition = $container->getDefinition($definition);
+            }
+
+            $class = $definition->getClass();
+            if (null === $class) {
```
stof: this logic looks wrong to me. As this compiler pass is registered as a before-optimization pass, the definition class might not be known yet as a class name:

- when using ChildDefinition, it can be `null` (if the class will be inherited from the parent), but this does _not_ mean that skipping the registration of the service is the valid processing (btw, this is a BC break in case someone uses a ChildDefinition to configure a normalizer today)
- if a service definition uses a parameter in the class name (which is supported since 2.0), the parameter is not yet resolved at that point, requiring to resolve it explicitly

src/Symfony/Component/Serializer/DependencyInjection/SerializerPass.php
```diff
@@ -67,8 +70,37 @@ public function process(ContainerBuilder $container): void
...
+                continue;
+            }
+            $interfaces = class_implements($class);
+            if (isset($interfaces[NormalizerInterface::class])) {
+                $trueNormalizers[] = $reference ?? $definition;
+            }
+            if (isset($interfaces[DenormalizerInterface::class])) {
+                $denormalizers[] = $reference ?? $definition;
```
stof: Duplicating inline service definitions in separate arguments might lead to instantiating a _separate_ instance of the service in each of them (btw, this might already be an issue with the current compiler pass if we put a reference elsewhere, as the definition is both added in the array and registered under its own id).
Instead, the logic creating the traceable normalizers above and replacing values in `$normalizers` should be registering the service and assigning a Reference in the array.

src/Symfony/Component/Serializer/Debug/TraceableEncoder.php
stof: this should also implement DenormalizerAwareInterface

src/Symfony/Component/Serializer/Encoder/XmlEncoder.php
stof: Override the `setSerializer` method of the trait to trigger the deprecation (and maybe remove the SerializerAwareInterface without removing the method from the class, so that the Serializer class no longer calls the method)

src/Symfony/Component/Serializer/SerializerAwareInterface.php
stof: I suggest also flagging the `setSerializer` method as `@deprecated`, as static analysis tend to inherit the deprecated status from interface methods to implementation, so that they would report `(new XmlEncoder())->setSerializer($s);` as a call to a deprecated function when they see it as a call to `XmlEncoder::setSerializer`

src/Symfony/Component/Serializer/Normalizer/ChainNormalizer.php
```diff
@@ -0,0 +1,213 @@
...
+    {
+        foreach ($normalizers as $normalizer) {
+            $this->addNormalizer($normalizer);
+        }
+    }
+
+    /**
+     * Add Normalizer last in the line.
+     */
+    public function addNormalizer(NormalizerInterface $normalizer): void
```
stof: do we need such public API ? I would make the class immutable (the Serializer class was immutable as well).

src/Symfony/Component/Serializer/Normalizer/ChainNormalizer.php
```diff
@@ -0,0 +1,213 @@
...
+    {
+        if ($normalizer instanceof NormalizerAwareInterface) {
+            $normalizer->setNormalizer($this);
```
stof, line 64: if we make the ChainNormalizer available publicly, this is flawed as it breaks when nesting a ChainNormalizer inside another ChainNormalizer (the normalizers inside the inner ChainNormalizer won't get access to the main normalizer supporting everything)

src/Symfony/Component/Serializer/Normalizer/ChainNormalizer.php
```diff
@@ -0,0 +1,213 @@
...
+        } else {
+            $type = 'native-'.\gettype($data);
+            $genericType = '*';
+        }
+
+        if (!isset($this->normalizerCache[$format][$type])) {
+            $this->normalizerCache[$format][$type] = [];
+
+            foreach ($this->normalizers as $k => $normalizer) {
+                if (!$normalizer instanceof NormalizerInterface) {
```
stof: this does not make sense to me. Once we separate the lists, we should not have anything else than a NormalizerInterface here

src/Symfony/Component/Serializer/Normalizer/ChainNormalizer.php
```diff
@@ -0,0 +1,213 @@
...
+
+    public function getSupportedTypes(?string $format): array
+    {
+        if (null === $format) {
+            $format = '__null__';
+        }
+
+        if (!isset($this->supportedCache[$format])) {
+            foreach ($this->normalizers as $normalizer) {
+                $this->supportedCache + $normalizer->getSupportedTypes($format);
```
stof: this logic is flawed:

- it does not write in `$this->supportedCache` at all, so if it is not set, you will get a notice below because it still won't be set
- it does not report that this normalizer knows how to normalize any non-object value without the help of any registered normalizer
- it does not properly deal with cases where multiple normalizers report support for the same type with different cacheability status (for which even the *order* of answers can be important, see the complexity in `getSupportedTypes`)

IMO, the only sane implementation of that method (especially considering that the only case where it will be used is when nesting a ChainNormalizer inside another so we won't need it for 99.999% of use cases) is `return ['*' => false];`

stof: Note that the same is true for the Denormalizer

stof (changes requested):

src/Symfony/Component/Serializer/Normalizer/ChainDenormalizer.php
stof: @yceruto you can have custom denormalizers for scalars if you want.

yceruto: Yeah, I was guessing that between the lines, but we should use better variable names to make it clearer for maintainers

src/Symfony/Component/Serializer/Serializer.php
Nyholm: Yeah. I dont like the current signature too. 

I want to have a signature like: 

```php
new Serializer($encoders, $normalizer, $denormalizer);
// or
new Serializer($normalizer, $denormalizer, $encoders);
```

Should I immediately go to that signature or should we change it in SF8?

---------

On your separate topic of making the `Chain*` classes an implementation detail.. That was an earlier version of the PR, but we decided against that to make the `Serializer` more simple. See https://github.com/symfony/symfony/pull/52900#discussion_r1417771466

Treating them like `$encoders` is a valid idea. However, you have 1-3 encoders MAX. But you may have 100+ normalizers/denormalizers. So I think it makes more sense to keep them separate and give the developer more control.

src/Symfony/Component/Serializer/Normalizer/ChainNormalizer.php
Nyholm: Oh, shit, you are correct. We cannot have such API. 

The serialzier cannot assume we have a `ChainNormalizer`. 

I'll update. I missed to change this when the PR changed.

src/Symfony/Component/Serializer/Normalizer/ChainNormalizer.php
Nyholm: Would a solution be to have `ChainNormalizer` also implement the `NormalizerAwareInterface`?

src/Symfony/Component/Serializer/Encoder/XmlEncoder.php
Nyholm: That is a good idea. But I cannot remove the `SerializerAwareInterface`. A class that extends `XmlEncoder` will assume that `XmlEncoder::$serializer` has a value. 

If I keep the `SerializerAwareInterface` and add a `trigger_deprecation()`, then the user has no way of removing that warning. 

So I dont think I can add any `trigger_deprecation()` in `setSerializer()`.

src/Symfony/Component/Serializer/Normalizer/ChainNormalizer.php
Nyholm: Thank you. I've updated this and added a test to prove correctness.

src/Symfony/Component/Serializer/Serializer.php
stof: > Should I immediately go to that signature or should we change it in SF8?

You **must** support the new signature immediately. Otherwise doing the change in 8.0 will be forbidden as it would be a BC break without migration path.

> On your separate topic of making the Chain* classes an implementation detail.. That was an earlier version of the PR, but we decided against that to make the Serializer more simple. See https://github.com/symfony/symfony/pull/52900#discussion_r1417771466

That comment was actually suggesting to _also_ change the way encoders are injected (we should avoid making 2 successive signature changes on the constructor, as that would be a pain to have 2 deprecated signatures to support in the BC layer)

> Treating them like $encoders is a valid idea. However, you have 1-3 encoders MAX. But you may have 100+ normalizers/denormalizers. So I think it makes more sense to keep them separate and give the developer more control.

but then, you make it more likely to have nested ChainNormalizer instance, making it necessary to be 100% sure about the reliability of `getSupportedTypes for instance (my suggested `return ['*' => false]` will be reliable, even though it might not be the most optimized one).

nicolas-grekas:
Closing as stale: the last push here is from June 2024 and stof's review points from the same week (ship the final constructor signature in one go, fix getSupportedTypes() for nested chains) were never addressed. Meanwhile Serializer.php has moved on and the 44-file diff no longer applies. The idea itself still has support, so please reopen or send a fresh PR if you want to pick it up again.

PR #52909, opened 2023-12-06
[HtmlSanitizer] Add functions to handle operations on multiple attributes and/or elements

EdouardCourty:
Add functions to handle operations on multiple attributes or elements at the same time

| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | None
| License       | MIT

This branch brings supports for array parameters to handle mass allow / block / drop of elements and/or attributes.

Here is a code screenshot that shows how these news methods work:

```php
$config = new HtmlSanitizerConfig();

$attributes = ['height', 'width', 'src'];
$elements = ['img',' iframe'];

/* Current way */
        
foreach ($elements as $element) {
    $config = $config->allowElement($element, $attributes);
}

/* New way */

$config = $config->allowElements($elements, $attributes);
```

### DIFF
--- a/src/Symfony/Component/HtmlSanitizer/HtmlSanitizerConfig.php
+++ b/src/Symfony/Component/HtmlSanitizer/HtmlSanitizerConfig.php
@@ -260,16 +260,7 @@ class HtmlSanitizerConfig
     public function allowElement(string $element, array|string $allowedAttributes = []): static
     {
         $clone = clone $this;
-
-        // Unblock the element is necessary
-        unset($clone->blockedElements[$element]);
-
-        $clone->allowedElements[$element] = [];
-
-        $attrs = ('*' === $allowedAttributes) ? array_keys(W3CReference::ATTRIBUTES) : (array) $allowedAttributes;
-        foreach ($attrs as $allowedAttr) {
-            $clone->allowedElements[$element][$allowedAttr] = true;
-        }
+        $this->handleAllowElement($clone, $element, $allowedAttributes);
 
         return $clone;
     }
@@ -279,11 +270,11 @@ class HtmlSanitizerConfig
      *
      * Allowed elements are elements the sanitizer should retain from the input.
      *
-     * A list of allowed attributes for this element can be passed as a second argument.
+     * A list of allowed attributes for these elements can be passed as a second argument.
      * Passing "*" will allow all standard attributes on this element. By default, no
      * attributes are allowed on the element.
      *
-     * @param list<string>        $elements
+     * @param string[]            $elements
      * @param list<string>|string $allowedAttributes
      */
     public function allowElements(array $elements, array|string $allowedAttributes = []): static
@@ -291,7 +282,7 @@ class HtmlSanitizerConfig
         $clone = clone $this;
 
         foreach ($elements as $element) {
-            $clone = $clone->allowElement($element, $allowedAttributes);
+            $this->handleAllowElement($clone, $element, $allowedAttributes);
         }
 
         return $clone;
@@ -306,11 +297,7 @@ class HtmlSanitizerConfig
     public function blockElement(string $element): static
     {
         $clone = clone $this;
-
-        // Disallow the element is necessary
-        unset($clone->allowedElements[$element]);
-
-        $clone->blockedElements[$element] = true;
+        $this->handleBlockElement($clone, $element);
 
         return $clone;
     }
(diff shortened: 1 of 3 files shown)


EdouardCourty:
By the way, why is the `HtmlSanitizerConfig` class using `clone` statements in all its methods?
image

stof:
@EdouardCourty being the Config object is immutable (which means that even if you keep a reference to the Config object, you cannot mutate the configuration of a Sanitizer that has been instantiated with it)

src/Symfony/Component/HtmlSanitizer/HtmlSanitizerConfig.php
```diff
@@ -274,6 +274,29 @@ public function allowElement(string $element, array|string $allowedAttributes =
         return $clone;
     }
 
+    /**
+     * Configures the given elements as allowed.
+     *
+     * Allowed elements are elements the sanitizer should retain from the input.
+     *
+     * A list of allowed attributes for this element can be passed as a second argument.
```
stof: `this element` does not make sense in this sentence, as you pass multiple ones.

src/Symfony/Component/HtmlSanitizer/HtmlSanitizerConfig.php
```diff
@@ -292,6 +315,23 @@ public function blockElement(string $element): static
...
+     *
+     * Blocked elements are elements the sanitizer should remove from the input, but retain
+     * their children.
+     */
+    public function blockElements(array $elements): static
+    {
+        $clone = clone $this;
+
+        foreach ($elements as $element) {
+            $clone = $clone->blockElement($element);
```
stof: Instead of using the public API (which will clone the config many times), you should modify the state of the clone directly, which would allow cloning only once. Otherwise, there is no real benefit for those methods

src/Symfony/Component/HtmlSanitizer/HtmlSanitizerConfig.php
```diff
@@ -367,6 +454,32 @@ public function dropAttribute(string $attribute, array|string $droppedElements):
...
+     * Dropped attributes are attributes the sanitizer should remove from the input.
+     *
+     * A list of elements on which to drop these attributes can be passed as a second argument.
+     * Passing "*" will drop this attribute from all currently allowed elements.
+     *
+     * Note: when using an empty configuration, all unknown attributes are dropped
+     * automatically. This method let you drop attributes that were allowed earlier
+     * in the configuration.
+     *
+     * @param list<string>        $attributes
```
stof: should be `string[]`. As the method iterate over the attributes without caring about keys at all, there is no reason to force the caller to pass a list.

EdouardCourty: Hi, I agree!
The rest of the methods are documented this way, which is why I thought I'd stay in the same style.

EdouardCourty:
> @EdouardCourty being the Config object is immutable (which means that even if you keep a reference to the Config object, you cannot mutate the configuration of a Sanitizer that has been instantiated with it)

Thanks for your reply, I'll fix the PR asap

src/Symfony/Component/HtmlSanitizer/HtmlSanitizerConfig.php
EdouardCourty: Hi, I wrapped the logic inside private functions that I can reuse in the loops and in the base functions

nicolas-grekas:
Thanks for the work here @EdouardCourty, and for your patience with how long it took to get back to you. Welcome to Symfony, and sorry that your first PR is one we are not taking.

The five plural methods wrap a loop that already reads fine with the existing API, which your own description shows:

```php
foreach ($elements as $element) {
    $config = $config->allowElement($element, $attributes);
}
```

That is ordinary PHP with nothing to discover and no trap in it, so the saving does not carry its weight: public API stays for a whole major version, and on a component whose job is to make a sanitizer hard to misconfigure, a wider surface is a cost of its own.

The sixth method is a separate problem. `handleAllowAttribute(self $clone, array|string $allowedElements, string $attribute): void` is public and writes into another instance's internals:

```php
$clone->allowedElements[$element][$attribute] = true;
```

`HtmlSanitizerConfig` is immutable, which is exactly what @stof explained in this thread, and a public method that mutates a config passed to it puts a hole in that. It reads as an internal helper that became public by accident.

EdouardCourty:
Hi, you're very clear!

Thanks for the update on this PR after all this time, see you around 😉

PR #52910, opened 2023-12-06
[DependencyInjection] Do not autowire private and protected properties

dontub:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

The [https://symfony.com/doc/5.4/service_container/autowiring.html#autowiring-other-methods-e-g-setters-and-public-typed-properties](documentation) says that only public properties are autowired. Nevertheless not only public, but also protected and private properties are handled by `\Symfony\Component\DependencyInjection\Compiler\AutowireRequiredPropertiesPass`. With this change only public properties are handled.

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Compiler/AutowireRequiredPropertiesPass.php
+++ b/src/Symfony/Component/DependencyInjection/Compiler/AutowireRequiredPropertiesPass.php
@@ -42,7 +42,7 @@ class AutowireRequiredPropertiesPass extends AbstractRecursivePass
         }
 
         $properties = $value->getProperties();
-        foreach ($reflectionClass->getProperties() as $reflectionProperty) {
+        foreach ($reflectionClass->getProperties(\ReflectionProperty::IS_PUBLIC) as $reflectionProperty) {
             if (!($type = $reflectionProperty->getType()) instanceof \ReflectionNamedType) {
                 continue;
             }
--- a/src/Symfony/Component/DependencyInjection/Compiler/AutowireRequiredPropertiesPass.php
+++ b/src/Symfony/Component/DependencyInjection/Compiler/AutowireRequiredPropertiesPass.php
@@ -24,9 +24,6 @@ use Symfony\Contracts\Service\Attribute\Required;
  */
 class AutowireRequiredPropertiesPass extends AbstractRecursivePass
 {
-    /**
-     * {@inheritdoc}
-     */
     protected function processValue($value, bool $isRoot = false)
     {
         if (\PHP_VERSION_ID < 70400) {
--- a/src/Symfony/Component/DependencyInjection/Tests/Compiler/AutowireRequiredPropertiesPassTest.php
+++ b/src/Symfony/Component/DependencyInjection/Tests/Compiler/AutowireRequiredPropertiesPassTest.php
@@ -43,6 +43,9 @@ class AutowireRequiredPropertiesPassTest extends TestCase
 
         $this->assertArrayHasKey('plop', $properties);
         $this->assertEquals(Bar::class, (string) $properties['plop']);
+
+        $this->assertArrayNotHasKey('plopProtected', $properties);
+        $this->assertArrayNotHasKey('plopPrivate', $properties);
     }
 
     /**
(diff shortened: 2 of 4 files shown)


nicolas-grekas:
How does the code behave currently? I suppose it throws an error at instantiation time isn't it?
It might be better to throw an exception when that annotation is found on non-public properties. WDYT?

RobertMe:
Nice catch. But shouldn't the behavior be consistent for properties and methods? As it seems like method injection also is possible for private and protected methods:
https://github.com/symfony/symfony/blob/c71348af6002339097a5b8c22916240dc184f3b0/src/Symfony/Component/DependencyInjection/Compiler/AutowireRequiredMethodsPass.php#L44
Applying the fix for methods as well, while it's not documented (as not being possible) doesn't seem like a BC break to me. This as it would already lead to a runtime error when the service is created (I think).

> How does the code behave currently? I suppose it throws an error at instantiation time isn't it?

Theoretically I do think so indeed. As it's dumped/generated as plain PHP code. If it was generated using reflection to inject into protected/private methods and properties it could work. Not sure whether that would be desired though (I think Java can do injection into private properties. But as noone seemed to have complained about this not working in Symfony I think it's okay to not support it).

> It might be better to throw an exception when that annotation is found on non-public properties. WDYT?

IMO that would be a must (failing at container compilation). Silently ignoring this would mean it's only detected when actually using a code path using the property (resulting in an uninitialized / `null` property error). While it could easily be detected at kernel compilation time.

dontub:
I'm working on a CiviCRM extension where on special classes `@required` on protected properties is a validation hint. When I try to use autowiring for those classes it fails here: https://github.com/symfony/symfony/blob/8d06a8ac0db4ce6c3d6dbe0d0ec7240b894d454d/src/Symfony/Component/DependencyInjection/Compiler/CheckExceptionOnInvalidReferenceBehaviorPass.php#L86

For variables type hinted as scalar or array it must fail there. I haven't tested what happens if the variable is type hinted with a class that can actually be resolved to a service. Though if Symfony doesn't use reflection to set properties it must fail somewhere else.

For now I decided to use a compiler pass that calls `$definition->setProperties([])` for those classes. So throwing an exception if `@required` is found on a non public property wouldn't be a solution for me :-)

nicolas-grekas:
My understanding: using `@requried` on non-public protected properties or methods is *not* allowed, because the engine forbids so at runtime. I'd be fine throwing an exception at compile time but I'd consider it as an improvement, so for 7.1.

dontub:
Then I'd suggest to make handling the attributes deactivatable per service so that it is possible to autowire a service with constructor injection and ignore any `required` attribute.

RobertMe:
> Then I'd suggest to make handling the attributes deactivatable per service so that it is possible to autowire a service with constructor injection and ignore any `required` attribute.

I think you mean `@required` annotation? As the attribute is based on an actual class and must be used with a FQCN so can't overlap with the product (CRM) you are using.
Another seemingly possible option might be to upgrade to Symfony 7? As that doesn't support the `@required` annotation anymore.

In any case: I don't think that the issue for your specific use case will be resolved in an older version. Either it's a feature and it will be fixed in 7.1, or it's a "minor" BC break and will be fixed in 7.1, or you'd have to upgrade to 7.0 to have Symfony not read the annotation anyway.

(For clarity: aggreed that Symfony could/should throw when either a protected/private property or method is marked as "required" (using annotation / attribute) during container compilation. Not sure though whether this is already done for manually configured property / method injection? As IMO it should be applied to that as well. If the service defintion (in YAML for example) states `setFoo()` must be called and said method doesn't exist of is not accessible the compilation should fail as well. So I guess that would require a separate pass which is executed after the autowiring is applied and thus automatically works for all scenarios, autowiring by marking as required, manual configuration, etc).

nicolas-grekas:
> I'm working on a CiviCRM extension where on special classes @required on protected properties is a validation hint.
> When I try to use autowiring for those classes it fails

OK, then I think this is a "won't fix" on my side: we're not going to change the current semantics of the annotation to fit an external use case. You *can* opt-out from `@required` and use another way to interpret it but then you're going to have to opt-out from autowiring altogether as you figured out.

My recommendation: don't use `@required` for another purpose than it's original semantics. Use another annotation in your extension instead.

Closing therefor.

PR #52913, opened 2023-12-06
[Routing] Fixed priority getting lost when setting localized prefix

### LINKED ISSUE #52912: [Routing] Route priority is lost when setting localized route prefix

6.4.1

### Description

Route for controller B has priority set:
```php
    #[Route(path:  '/b', name: 'b_method', priority: 9999)]
    public function bMethodOne(): Response
```

When the prefix in `routes.yaml` is commented out, the routing priority works as expected. The `bin/console debug:router` command outputs the following:

```bash
❯ bin/console debug:router
 ---------------- -------- -------- ------ -------------------------- 
  Name             Method   Scheme   Host   Path                      
 ---------------- -------- -------- ------ -------------------------- 
  b_method         ANY      ANY      ANY    /b                        
  _preview_error   ANY      ANY      ANY    /_error/{code}.{_format}  
  a_method         ANY      ANY      ANY    /a                        
 ---------------- -------- -------- ------ -------------------------- 
```

However, when the prefix is set, the routing priority seems to be lost. The `bin/console debug:router` command then shows a different order:

```yaml
controllers:
    resource: '../src/**/*Controller.php'
    type: attribute
    prefix:
        cs: ~
        en: '/en'
 ```

```bash
❯ bin/console debug:router
 ---------------- -------- -------- ------ -------------------------- 
  Name             Method   Scheme   Host   Path                      
 ---------------- -------- -------- ------ -------------------------- 
  _preview_error   ANY      ANY      ANY    /_error/{code}.{_format}  
  a_method.cs      ANY      ANY      ANY    /a                        
  a_method.en      ANY      ANY      ANY    /en/a                     
  b_method.cs      ANY      ANY      ANY    /b                        
  b_method.en      ANY      ANY      ANY    /en/b                     
 ---------------- -------- -------- ------ -------------------------- 
 ```

### How to reproduce

https://github.com/SystematicCZ/Symfony-Routing-bug

### Possible Solution

I will try to submit PR soon

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by SystematicCZ**
| Q             | A
| ------------- | ---
| Branch?       |  5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52912
| License       | MIT

When a route prefix is an array, the original route is removed from the collection and then re-added with a new name and path. During this process, the route's priority is lost as it's not preserved during the removal and re-addition.

Current state of `PrefixTrait.php`
```php
 $routes->remove($name);
 ...
  $routes->add($name.'.'.$locale, $localizedRoute);
 ```

To resolve this issue, I implemented a change where the priority of the route is stored before removal. This stored priority is then reassigned to the route when it's added back to the collection.

### DIFF
--- a/src/Symfony/Component/Routing/Tests/Fixtures/AttributeFixtures/RouteWithPriorityController.php
+++ b/src/Symfony/Component/Routing/Tests/Fixtures/AttributeFixtures/RouteWithPriorityController.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Symfony\Component\Routing\Tests\Fixtures\AttributeFixtures;
+
+use Symfony\Component\Routing\Annotation\Route;
+
+class RouteWithPriorityController
+{
+    /**
+     * @Route("/important", name="important", priority=2)
+     */
+    public function important()
+    {
+    }
+
+    /**
+     * @Route("/also-important", name="also_important", priority=1, defaults={"_locale": "cs"})
+     */
+    public function alsoImportant()
+    {
+    }
+}
--- a/src/Symfony/Component/Routing/Tests/Loader/YamlFileLoaderTest.php
+++ b/src/Symfony/Component/Routing/Tests/Loader/YamlFileLoaderTest.php
@@ -11,9 +11,12 @@
 
 namespace Symfony\Component\Routing\Tests\Loader;
 
+use Doctrine\Common\Annotations\AnnotationReader;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Config\FileLocator;
+use Symfony\Component\Config\Loader\LoaderResolver;
 use Symfony\Component\Config\Resource\FileResource;
+use Symfony\Component\Routing\Loader\AnnotationClassLoader;
 use Symfony\Component\Routing\Loader\YamlFileLoader;
 use Symfony\Component\Routing\Route;
 use Symfony\Component\Routing\RouteCollection;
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Review by nicolas-grekas — COMMENTED** [review]
I'm annoyed by the need for adding a method but I don't see any other non-hacky way.
Since this issue exists in 5.4, can you please rebase/retarget the PR for 5.4?

**Comment by SystematicCZ** [comment]
Thx. I will rebase tommorow.

File: src/Symfony/Component/Routing/Tests/Fixtures/AttributeFixtures/RouteWithPriorityController.php [hunk 1]
```diff
@@ -0,0 +1,20 @@
...
+namespace Symfony\Component\Routing\Tests\Fixtures\AttributeFixtures;
+
+use Symfony\Component\Routing\Attribute\Route;
+
+class RouteWithPriorityController
+{
+    #[Route(path: '/important', name: 'important', priority: 2)]
+    public function important()
+    {
+
```
nicolas-grekas, line 15:
```suggestion
```

src/Symfony/Component/Routing/Tests/Fixtures/AttributeFixtures/RouteWithPriorityController.php
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    #[Route(path: '/also-important', name: 'also_important', defaults: ['_locale' => 'cs'], priority: 1)]
+    public function alsoImportant()
+    {
+
```
nicolas-grekas:
```suggestion
```

src/Symfony/Component/Routing/Tests/Loader/YamlFileLoaderTest.php
```diff
@@ -458,4 +460,23 @@ public function testImportingAliases()
 
         $this->assertEquals($expectedRoutes('yaml'), $routes);
     }
+
+    public function testPriorityWithPrefix()
+    {
+        new LoaderResolver([
+            $loader = new YamlFileLoader(new FileLocator(\dirname(__DIR__).'/Fixtures/localized')),
+            new class() extends AttributeClassLoader {
```
nicolas-grekas: This class doesn't exist in 5.4.
Maybe we can define the routes using plain yaml or xml instead so that we don't have to deal with this concern (migrating from annotations to attributes when merging up ;) )

SystematicCZ: Ok, thats a good idea. I will fix

SystematicCZ: According to docs priority can be set only in annotations or attributes. So I don't see a clear way how to test this easily on 5.4 and 6 without changing the code between those

```In YAML and XML you can move the route definitions up or down in the configuration file to control their priority. In routes defined as PHP attributes this is much harder to do, so you can set the optional priority parameter in those routes to control their priority```

nicolas-grekas:
Thank you @SystematicCZ.

PR #52915, opened 2023-12-06
[Messenger] Fix missing `@throws` phpdoc on MiddlewareInterface, MessageBusInterface and SenderInterface

VincentLanglet:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes (kinda, it's missing phpdoc)
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #... 
| License       | MIT

`@throws` phpdoc help the developper to know those methods can throw an exception and to have a contract about which exception can be thrown.

I recently got an exception in production because of the call `$messageBus->dispatch()` without knowing it could throw an exception. Seems like it's because it iterable on every Middleware and a lot of middleware are throwing exception.

As an example:
- https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Messenger/Middleware/RejectRedeliveredMessageMiddleware.php#L38
- Almost every sender are throwing `TransportException` in case of errors, which happen in the SendMessageMiddleware.

### DIFF
--- a/src/Symfony/Component/Messenger/MessageBusInterface.php
+++ b/src/Symfony/Component/Messenger/MessageBusInterface.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Component\Messenger;
 
+use Symfony\Component\Messenger\Exception\ExceptionInterface;
 use Symfony\Component\Messenger\Stamp\StampInterface;
 
 /**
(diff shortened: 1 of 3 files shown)


chalasr (approved):

fabpot (approved):
For 7.1.

fabpot:
Thank you @VincentLanglet.

PR #52916, opened 2023-12-06
[Mailer] Dispatch event for Postmark's "inactive recipient" API error

### LINKED ISSUE #50165: [Mailer] Sending to a blocked email address through Postmark

vicdelfant:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      |no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #50165
| License       | MIT

Given the use of Postmark and a recipient that previously generated a bounce, attempting to send another email to that email address results in an HTTP 422 response, along with the error code [`406 - Inactive Recipient`](https://postmarkapp.com/developer/api/overview#error-codes). In the real world, this situation can arise easily in cases of a typo or an email address that had temporary issues.

Because `PostmarkApiTransport` requires an HTTP 200 and throws a `HttpTransportException` for any other HTTP code, something that's of minor interest to the application itself (i.e. a _possibly_ inactive e-mail address) now causes an exception. Depending on the userland logic, this can halt a process that sends out survey reminders, cause the Messenger component to queue the message for retrying etc.

To handle this more elegantly, I'm proposing the following changes:
* Add a `PostmarkDeliveryEventFactory` for casting any (future) delivery events to an instance of `PostmarkDeliveryEvent`. Currently, only support the 406 'inactive recipient' error is included;
* Adjust the `PostmarkApiTransport` so it checks for supported delivery events on an HTTP code other than 200, and if so, dispatches the event accordingly.

We cannot port this logic to the Postmark SMTP transport because, according to Postmark's own documentation, these error response codes are only provided by their API endpoints.

### DIFF
--- a/src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkDeliveryEvent.php
+++ b/src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkDeliveryEvent.php
@@ -1,0 +1,64 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Mailer\Bridge\Postmark\Event;
+
+use Symfony\Component\Mime\Header\Headers;
+
+class PostmarkDeliveryEvent
+{
+    public const CODE_INACTIVE_RECIPIENT = 406;
+
+    private int $errorCode;
+
+    private Headers $headers;
+
+    private ?string $message;
+
+    public function __construct(string $message, int $errorCode)
+    {
+        $this->message = $message;
+        $this->errorCode = $errorCode;
+
+        $this->headers = new Headers();
+    }
+
+    public function getErrorCode(): int
+    {
+        return $this->errorCode;
+    }
+
+    public function getHeaders(): Headers
+    {
+        return $this->headers;
+    }
+
+    public function getMessage(): ?string
+    {
+        return $this->message;
+    }
+
+    public function getMessageId(): ?string
+    {
+        if (!$this->headers->has('Message-ID')) {
+            return null;
+        }
+
+        return $this->headers->get('Message-ID')->getBodyAsString();
+    }
+
+    public function setHeaders(Headers $headers): self
+    {
+        $this->headers = $headers;
+
+        return $this;
+    }
+}
--- a/src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkDeliveryEventFactory.php
+++ b/src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkDeliveryEventFactory.php
@@ -1,0 +1,34 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Mailer\Bridge\Postmark\Event;
+
+use Symfony\Component\Mime\Email;
+
+class PostmarkDeliveryEventFactory
+{
+    public function create(int $errorCode, string $message, Email $email): PostmarkDeliveryEvent
+    {
+        if (!$this->supports($errorCode)) {
+            throw new \InvalidArgumentException(sprintf('Error code "%s" is not supported.', $errorCode));
+        }
+
+        return (new PostmarkDeliveryEvent($message, $errorCode))
+            ->setHeaders($email->getHeaders());
+    }
+
+    public function supports(int $errorCode): bool
+    {
+        return \in_array($errorCode, [
+            PostmarkDeliveryEvent::CODE_INACTIVE_RECIPIENT,
+        ]);
+    }
+}
--- a/src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkEvents.php
+++ b/src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkEvents.php
@@ -1,0 +1,17 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Mailer\Bridge\Postmark\Event;
+
+class PostmarkEvents
+{
+    public const DELIVERY = 'postmark.delivery';
+}
--- a/src/Symfony/Component/Mailer/Bridge/Postmark/Transport/PostmarkApiTransport.php
+++ b/src/Symfony/Component/Mailer/Bridge/Postmark/Transport/PostmarkApiTransport.php
@@ -13,6 +13,8 @@ namespace Symfony\Component\Mailer\Bridge\Postmark\Transport;
 
 use Psr\EventDispatcher\EventDispatcherInterface;
 use Psr\Log\LoggerInterface;
+use Symfony\Component\Mailer\Bridge\Postmark\Event\PostmarkDeliveryEventFactory;
+use Symfony\Component\Mailer\Bridge\Postmark\Event\PostmarkEvents;
 use Symfony\Component\Mailer\Envelope;
 use Symfony\Component\Mailer\Exception\HttpTransportException;
 use Symfony\Component\Mailer\Exception\TransportException;
@@ -33,6 +35,8 @@ class PostmarkApiTransport extends AbstractApiTransport
 {
     private const HOST = 'api.postmarkapp.com';
 
+    private ?EventDispatcherInterface $dispatcher;
+
     private string $key;
 
     private ?string $messageStream = null;
@@ -40,6 +44,7 @@ class PostmarkApiTransport extends AbstractApiTransport
     public function __construct(string $key, HttpClientInterface $client = null, EventDispatcherInterface $dispatcher = null, LoggerInterface $logger = null)
     {
         $this->key = $key;
+        $this->dispatcher = $dispatcher;
 
         parent::__construct($client, $dispatcher, $logger);
     }
@@ -69,6 +74,18 @@ class PostmarkApiTransport extends AbstractApiTransport
         }
 
         if (200 !== $statusCode) {
+            $eventFactory = new PostmarkDeliveryEventFactory();
+
+            // Some delivery issues can be handled silently - route those through EventDispatcher
+            if (null !== $this->dispatcher && $eventFactory->supports($result['ErrorCode'])) {
+                $this->dispatcher->dispatch(
+                    $eventFactory->create($result['ErrorCode'], $result['Message'], $email),
+                    PostmarkEvents::DELIVERY,
+                );
+
+                return $response;
+            }
+
             throw new HttpTransportException('Unable to send an email: '.$result['Message'].sprintf(' (code %d).', $result['ErrorCode']), $response);
         }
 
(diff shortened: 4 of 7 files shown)


src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkDeliveryEventFactory.php
```diff
@@ -0,0 +1,34 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Mailer\Bridge\Postmark\Event;
+
+use Symfony\Component\Mime\Email;
+
+class PostmarkDeliveryEventFactory
+{
+    public function create($errorCode, string $message, Email $email): PostmarkDeliveryEvent
```
OskarStark:
```suggestion
    public function create(int $errorCode, string $message, Email $email): PostmarkDeliveryEvent
```

vicdelfant: Good point, I missed that one - thanks!

src/Symfony/Component/Mailer/Bridge/Postmark/Transport/PostmarkApiTransport.php
```diff
@@ -69,6 +74,18 @@ protected function doSendApi(SentMessage $sentMessage, Email $email, Envelope $e
         }
 
         if (200 !== $statusCode) {
+            $eventFactory = new PostmarkDeliveryEventFactory();
+
+            // Some delivery issues can be handled silently - route those through EventDispatcher
+            if (null !== $this->dispatcher && $eventFactory->supports($result['ErrorCode'])) {
+                $this->dispatcher->dispatch(
+                    $eventFactory->create($result['ErrorCode'], $result['Message'], $email),
+                    PostmarkEvents::DELIVERY
```
fabpot:
```suggestion
                    PostmarkEvents::DELIVERY,
```

fabpot (approved):

fabpot:
Thank you @vicdelfant.

vicdelfant:
Thanks for merging @fabpot! On to the next one :)

atymic:
Wooo! I assume this will be in the 7.1 release, will check if laravel needs a PR to handle.

src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkDeliveryEvent.php
```diff
@@ -0,0 +1,64 @@
...
+    public function getMessageId(): ?string
+    {
+        if (!$this->headers->has('Message-ID')) {
+            return null;
+        }
+
+        return $this->headers->get('Message-ID')->getBodyAsString();
+    }
+
+    public function setHeaders(Headers $headers): self
```
stof, line 58: what is the reason for making this mutable instead of passing it in the constructor ?

src/Symfony/Component/Mailer/Bridge/Postmark/Transport/PostmarkApiTransport.php
```diff
@@ -69,6 +74,18 @@ protected function doSendApi(SentMessage $sentMessage, Email $email, Envelope $e
         }
 
         if (200 !== $statusCode) {
+            $eventFactory = new PostmarkDeliveryEventFactory();
```
stof, line 77: should we instantiate a new one each time or treat it as a dependency of the class stored in a property ?

src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkEvents.php
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Mailer\Bridge\Postmark\Event;
+
+class PostmarkEvents
+{
+    public const DELIVERY = 'postmark.delivery';
```
stof, line 16: I don't think we need this class with event names. For new code, we should rely on the event class name instead (by passing only 1 argument to `dispatch()`)

src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkDeliveryEvent.php
```diff
@@ -0,0 +1,64 @@
...
+
+class PostmarkDeliveryEvent
+{
+    public const CODE_INACTIVE_RECIPIENT = 406;
+
+    private int $errorCode;
+
+    private Headers $headers;
+
+    private ?string $message;
```
stof, line 24: why is this nullable when the constructor does not allow passing null ?

src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkDeliveryEvent.php
```diff
@@ -0,0 +1,64 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Mailer\Bridge\Postmark\Event;
+
+use Symfony\Component\Mime\Header\Headers;
+
+class PostmarkDeliveryEvent
+{
+    public const CODE_INACTIVE_RECIPIENT = 406;
```
stof, line 18: what is the reason to expose this constant in the public API of the bridge ?

src/Symfony/Component/Mailer/Bridge/Postmark/Transport/PostmarkApiTransport.php
```diff
@@ -69,6 +74,18 @@ protected function doSendApi(SentMessage $sentMessage, Email $email, Envelope $e
...
+
+            // Some delivery issues can be handled silently - route those through EventDispatcher
+            if (null !== $this->dispatcher && $eventFactory->supports($result['ErrorCode'])) {
+                $this->dispatcher->dispatch(
+                    $eventFactory->create($result['ErrorCode'], $result['Message'], $email),
+                    PostmarkEvents::DELIVERY,
+                );
+
+                return $response;
```
stof, line 86: shouldn't we still throw an exception by default to report that sending the email failed ? Right now, if you have no listener doing something special, the failure will be silently ignored.
To me, skipping the exception should require that a listener marks the failure as handled in some way (by calling a method on the event object)

src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkDeliveryEvent.php
vicdelfant: No specific reason, except for keeping the event with the error code and the error code itself close together. Alternatively, the mapping/constant could be moved to `PostmarkDeliveryEventFactory`.

src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkDeliveryEvent.php
vicdelfant: Good eye, this could indeed be a non-nullable property.

src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkDeliveryEvent.php
vicdelfant: No particular reason — it's a personal preference to keep the constructor as compact as possible, with the absolute minimal required data to set up the object in a valid, initialized state.

src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkEvents.php
vicdelfant: Fair point, should this be adjusted while the PR has already been merged @fabpot?

src/Symfony/Component/Mailer/Bridge/Postmark/Transport/PostmarkApiTransport.php
vicdelfant: Certainly, this could be a dependency that we instantiate in the constructor.

src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkEvents.php
stof: This has not been released yet, so we can still adjust it.

src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkDeliveryEvent.php
stof: This setter means that listeners can replace headers in the event (which will at least impact next listeners). Anything that should not be mutable from listeners should not have setters.

src/Symfony/Component/Mailer/Bridge/Postmark/Transport/PostmarkApiTransport.php
vicdelfant: In my opinion, this would defeat the purpose of the PR. If we would take a traditional SMTP server as a reference, any e-mail that is valid gets sent - simple as that. This also means that you can send 100 e-mails to an e-mail address that bounced on the first try, because the SMTP server will simply accept the e-mails and try to deliver them. There would be no way for the SMTP server or the application to know that delivering the e-mail, in fact, failed (or would fail down the line). Ergo, no exceptions.

Postmark's API, on the other hand, has stricter policies and is kind enough to throw an 'inactive recipient' API error to let you know that the e-mail address bounced before, or is otherwise suppressed. This is where the problem lies: with the pre-PR implementation, the application all of a sudden has to deal with exceptions while trying to deliver an otherwise valid e-mail. Also because the exception means that the recipient _could_ be inactive, it's in fact not 100% sure that it still is.

Sure we could wrap every sending operation in a try/catch, but in my opinion we should keep the logic of the mailer as close to its core as possible: using a traditional SMTP server that just queues mails and tries to deliver them. Anything extra is, well, extra.

Then again this is my opinion, but if we would be throwing an exception I'd rather revert the PR so keep things as simple as possible, and rely on the webhooks to detect post-delivery issues.

src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkDeliveryEvent.php
stof: As the factory is the only usage of the constant (outside tests), I would suggest moving it there as a private constant (and making tests use the integer directly). Once we introduce a public API, it becomes covered by the BC promise of Symfony. So we should not introduce them without reason (it only make maintenance harder)

src/Symfony/Component/Mailer/Bridge/Postmark/Transport/PostmarkApiTransport.php
stof: but this won't be sent to webhooks either (as the API will reject the sending)

vicdelfant: If the intent is to have an exception then reverting this PR would do just that, you'd get the exception back. The bounce will actually trigger a webhook, but only the first time: https://postmarkapp.com/developer/webhooks/bounce-webhook.

src/Symfony/Component/Mailer/Bridge/Postmark/Event/PostmarkDeliveryEvent.php
xabbuh: already fixed in #53242

PR #52917, opened 2023-12-06
[Serializer] Fix unexpected allowed attributes

### LINKED ISSUE #52673: [Serializer] All fields are not included by default, but upon adding any attribute, all fields are then required

mtarld:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52673 #49710
| License       | MIT

A more accurate approach than #52680

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/GetSetMethodNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/GetSetMethodNormalizer.php
@@ -36,6 +36,7 @@ use Symfony\Component\Serializer\Annotation\Ignore;
  */
 class GetSetMethodNormalizer extends AbstractObjectNormalizer
 {
+    private static $reflectionCache = [];
     private static $setterAccessibleCache = [];
 
     /**
@@ -43,7 +44,7 @@ class GetSetMethodNormalizer extends AbstractObjectNormalizer
      */
     public function supportsNormalization($data, ?string $format = null)
     {
-        return parent::supportsNormalization($data, $format) && $this->supports(\get_class($data));
+        return parent::supportsNormalization($data, $format) && $this->supports(\get_class($data), true);
     }
 
     /**
@@ -51,7 +52,7 @@ class GetSetMethodNormalizer extends AbstractObjectNormalizer
      */
     public function supportsDenormalization($data, string $type, ?string $format = null)
     {
-        return parent::supportsDenormalization($data, $type, $format) && $this->supports($type);
+        return parent::supportsDenormalization($data, $type, $format) && $this->supports($type, false);
     }
 
     /**
@@ -63,18 +64,22 @@ class GetSetMethodNormalizer extends AbstractObjectNormalizer
     }
 
     /**
-     * Checks if the given class has any getter method.
+     * Checks if the given class has any getter or setter method.
      */
-    private function supports(string $class): bool
+    private function supports(string $class, bool $readAttributes): bool
     {
         if (null !== $this->classDiscriminatorResolver && $this->classDiscriminatorResolver->getMappingForClass($class)) {
             return true;
         }
 
-        $class = new \ReflectionClass($class);
-        $methods = $class->getMethods(\ReflectionMethod::IS_PUBLIC);
-        foreach ($methods as $method) {
-            if ($this->isGetMethod($method)) {
+        if (!isset(self::$reflectionCache[$class])) {
+            self::$reflectionCache[$class] = new \ReflectionClass($class);
+        }
+
+        $reflection = self::$reflectionCache[$class];
+
+        foreach ($reflection->getMethods(\ReflectionMethod::IS_PUBLIC) as $reflectionMethod) {
+            if ($readAttributes ? $this->isGetMethod($reflectionMethod) : $this->isSetMethod($reflectionMethod)) {
                 return true;
             }
         }
@@ -93,6 +98,17 @@ class GetSetMethodNormalizer extends AbstractObjectNormalizer
             && ((2 < ($methodLength = \strlen($method->name)) && str_starts_with($method->name, 'is'))
                 || (3 < $methodLength && (str_starts_with($method->name, 'has') || str_starts_with($method->name, 'get')))
             );
+    }
+
+    /**
+     * Checks if a method's name matches /^set.+$/ and can be called non-statically with one parameter.
+     */
+    private function isSetMethod(\ReflectionMethod $method): bool
+    {
+        return !$method->isStatic()
+            && (\PHP_VERSION_ID < 80000 || !$method->getAttributes(Ignore::class))
+            && 1 === $method->getNumberOfRequiredParameters()
+            && str_starts_with($method->name, 'set');
     }
 
     /**
@@ -124,19 +140,17 @@ class GetSetMethodNormalizer extends AbstractObjectNormalizer
      */
     protected function getAttributeValue(object $object, string $attribute, ?string $format = null, array $context = [])
     {
-        $ucfirsted = ucfirst($attribute);
-
-        $getter = 'get'.$ucfirsted;
+        $getter = 'get'.$attribute;
         if (method_exists($object, $getter) && \is_callable([$object, $getter])) {
             return $object->$getter();
         }
 
-        $isser = 'is'.$ucfirsted;
+        $isser = 'is'.$attribute;
         if (method_exists($object, $isser) && \is_callable([$object, $isser])) {
             return $object->$isser();
         }
 
-        $haser = 'has'.$ucfirsted;
+        $haser = 'has'.$attribute;
         if (method_exists($object, $haser) && \is_callable([$object, $haser])) {
             return $object->$haser();
         }
@@ -149,7 +163,7 @@ class GetSetMethodNormalizer extends AbstractObjectNormalizer
      */
     protected function setAttributeValue(object $object, string $attribute, $value, ?string $format = null, array $context = [])
     {
-        $setter = 'set'.ucfirst($attribute);
+        $setter = 'set'.$attribute;
         $key = \get_class($object).':'.$setter;
 
         if (!isset(self::$setterAccessibleCache[$key])) {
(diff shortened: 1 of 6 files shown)


mtarld:
fabbot errors aren't related to that PR.

src/Symfony/Component/Serializer/Normalizer/GetSetMethodNormalizer.php
```diff
@@ -63,21 +64,30 @@ public function hasCacheableSupportsMethod(): bool
...
         if (null !== $this->classDiscriminatorResolver && $this->classDiscriminatorResolver->getMappingForClass($class)) {
             return true;
         }
 
-        $class = new \ReflectionClass($class);
-        $methods = $class->getMethods(\ReflectionMethod::IS_PUBLIC);
-        foreach ($methods as $method) {
-            if ($this->isGetMethod($method)) {
-                return true;
+        $reflection = new \ReflectionClass($class);
```
chalasr: any specific reason not to cache it?

src/Symfony/Component/Serializer/Normalizer/GetSetMethodNormalizer.php
```diff
@@ -63,21 +64,30 @@ public function hasCacheableSupportsMethod(): bool
...
+
+        // fallback to "read" to ensure backward compatibility
```
chalasr: Talking about BC while there is no deprecated code path feels weird. Do you plan to deprecate not setting that option explicitly in another PR on the dev branch? If not, I'd say better remove the comment or explain the reasoning otherwise if that fallback is going to be the default forever.

src/Symfony/Component/Serializer/Normalizer/GetSetMethodNormalizer.php
```diff
@@ -160,4 +181,53 @@ protected function setAttributeValue(object $object, string $attribute, $value,
...
+
+                if ($constructor && $constructor->isPublic()) {
+                    foreach ($constructor->getParameters() as $parameter) {
+                        if ($parameter->getName() === $attribute) {
+                            return true;
+                        }
+                    }
+                }
+            } else {
+                // ensure backward compatibility when no attribute access has been defined
```
chalasr: same

src/Symfony/Component/Serializer/Normalizer/GetSetMethodNormalizer.php
```diff
@@ -36,22 +36,23 @@
...
 class GetSetMethodNormalizer extends AbstractObjectNormalizer
 {
+    private static $reflectionCache = [];
     private static $setterAccessibleCache = [];
 
     /**
      * {@inheritdoc}
      */
-    public function supportsNormalization($data, string $format = null)
+    public function supportsNormalization($data, string $format = null, array $context = [])
```
chalasr: we took care of introducing the very same parameter smoothly in upper branches (see https://github.com/symfony/serializer/blob/6.3/Normalizer/GetSetMethodNormalizer.php) so I'm not sure we can do it this way

chalasr (changes requested):
As discussed at SymfonyCon, the information (access type) would better transit through a method parameter than a context option since the need is specific to the abstract method at hand

mtarld:
Once this PR will be upmerged to 7.1, I'll create a new one to trigger the proper deprecations.

chalasr (approved):

nicolas-grekas (commented):
The implementation fits the definition of a new feature to me: adding an argument on a constructor, and adding an argument on public/protected methods.
This should come with an `@param` annotation for methods, and this will yield a new deprecation.
Maybe another approach is possible? I don't know at all :)

chalasr:
That read/write mode do need to be passed throughout method calls to fix the bug at hand, so the plan here is to make it a parameter that's hidden at first on 5.4, then make it virtual on 7.1 and finally make it real on 8.0.
The only alternative I can think of is to put the info in `$context` instead. 

@mtarld would that work conceptually and technically?

mtarld:
It appears that it works indeed, and it is much better thanks!

I'm wondering if we should make that read/write information explicit in 7.1, maybe a context value like here is enough (but it is less secure).

nesl247:
Is there a chance this merges soon? We're actually stuck on an older version of api platform that accidentally stopped this bug from happening but the newer version removed that code.

If not we may move forward with applying the changes via a patch in our project. Just trying to figure out the right path for us.

Thanks!

src/Symfony/Component/Serializer/Normalizer/GetSetMethodNormalizer.php
```diff
@@ -63,22 +64,28 @@ public function hasCacheableSupportsMethod(): bool
...
 
+        $reflection = self::$reflectionCache[$class];
+
+        do {
+            foreach ($reflection->getMethods(\ReflectionMethod::IS_PUBLIC) as $reflectionMethod) {
+                if ($readAttributes && $this->isGetMethod($reflectionMethod) || !$readAttributes && $this->isSetMethod($reflectionMethod)) {
+                    return true;
+                }
+            }
+        } while ($reflection = $reflection->getParentClass());
```
nicolas-grekas: this loop is not needed if we scan only public methods

src/Symfony/Component/Serializer/Normalizer/GetSetMethodNormalizer.php
```diff
@@ -63,22 +64,28 @@ public function hasCacheableSupportsMethod(): bool
...
-            }
+        if (!isset(self::$reflectionCache[$class])) {
+            self::$reflectionCache[$class] = new \ReflectionClass($class);
         }
 
+        $reflection = self::$reflectionCache[$class];
+
+        do {
+            foreach ($reflection->getMethods(\ReflectionMethod::IS_PUBLIC) as $reflectionMethod) {
+                if ($readAttributes && $this->isGetMethod($reflectionMethod) || !$readAttributes && $this->isSetMethod($reflectionMethod)) {
```
nicolas-grekas:
```suggestion
                if ($readAttributes ? $this->isGetMethod($reflectionMethod) : $this->isSetMethod($reflectionMethod)) {
```

src/Symfony/Component/Serializer/Normalizer/GetSetMethodNormalizer.php
```diff
@@ -160,4 +178,49 @@ protected function setAttributeValue(object $object, string $attribute, $value,
...
+            if ($context['_read_attributes'] ?? true) {
+                foreach (['get', 'is', 'has'] as $getterPrefix) {
+                    $getter = $getterPrefix.ucfirst($attribute);
+                    $reflectionMethod = $reflection->hasMethod($getter) ? $reflection->getMethod($getter) : null;
+                    if ($reflectionMethod && $this->isGetMethod($reflectionMethod)) {
+                        return true;
+                    }
+                }
+            } else {
+                $setter = 'set'.ucfirst($attribute);
```
nicolas-grekas:
```suggestion
                $setter = 'set'.$attribute;
```

src/Symfony/Component/Serializer/Normalizer/GetSetMethodNormalizer.php
```diff
@@ -160,4 +178,49 @@ protected function setAttributeValue(object $object, string $attribute, $value,
...
+                $reflectionMethod = $reflection->hasMethod($setter) ? $reflection->getMethod($setter) : null;
+                if ($reflectionMethod && $this->isSetMethod($reflectionMethod)) {
```
nicolas-grekas:
```suggestion
                if ($reflection->hasMethod($setter) && $this->isSetMethod($reflection->getMethod($setter))) {
```

src/Symfony/Component/Serializer/Normalizer/GetSetMethodNormalizer.php
```diff
@@ -160,4 +178,49 @@ protected function setAttributeValue(object $object, string $attribute, $value,
...
+
+                if ($constructor && $constructor->isPublic()) {
+                    foreach ($constructor->getParameters() as $parameter) {
+                        if ($parameter->getName() === $attribute) {
+                            return true;
+                        }
+                    }
+                }
+            }
+        } while ($reflection = $reflection->getParentClass());
```
nicolas-grekas: do we deal only with public methods? if yes, we don't need this loop (please confirm that aspect is tested)

src/Symfony/Component/Serializer/Normalizer/ObjectNormalizer.php
```diff
@@ -174,4 +180,19 @@ protected function getAllowedAttributes($classOrObject, array $context, bool $at
...
+            return false;
+        }
+        $class = \is_object($classOrObject) ? \get_class($classOrObject) : $classOrObject;
+
+        if ($context['_read_attributes'] ?? true) {
+            return $this->propertyInfoExtractor->isReadable($class, $attribute);
+        }
+
+        return $this->propertyInfoExtractor->isWritable($class, $attribute)
+            || null !== ($writeInfo = $this->propertyInfoExtractor->getWriteInfo($class, $attribute)) && PropertyWriteInfo::TYPE_NONE !== $writeInfo->getType();
```
nicolas-grekas:
```suggestion
            || ($writeInfo = $this->propertyInfoExtractor->getWriteInfo($class, $attribute)) && PropertyWriteInfo::TYPE_NONE !== $writeInfo->getType();
```

src/Symfony/Component/Serializer/Normalizer/GetSetMethodNormalizer.php
mtarld: Yes it is actually tested. `GetSetDummyChild` stands for that purpose

maxhelias (approved):

nicolas-grekas (approved):

fabpot:
Thank you @mtarld.

stloyd:
Those changes generate a big performance drop in one of the projects I work with where we extensively use attributes on DTOs.

Comparing Serializer in version `7.0.6` vs `7.0.7`:
Zrzut ekranu 2024-05-6 o 16 27 33
Zrzut ekranu 2024-05-6 o 16 27 17

nicolas-grekas:
Please open a dedicated issue if you want any follow up.

PR #52922, opened 2023-12-07
[DependencyInjection] Add Lazy attribute for classes and arguments

Tiriel:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | #
| License       | MIT

This commit adds a Lazy attribute shortcut for service definitions. The goal here is simply to improve the DX with a single attribute for lazy loaded services, while adding the strict minimum amount of new code.
This attribute can be used as a replacement for the `Autowire(lazy: bool|string)` attribute on any autowired argument or for the `Autoconfigure(lazy: bool|string)` attribute on an class definition. Both use case support an optional parameter to specify with class/interface should be used for the proxy.

Usage on a class:
```php
use Symfony\Component\DependencyInjection\Attribute\Lazy;

#[Lazy]
class LazyLoaded
{
    // ...
}
```

On a method argument:
```php
use Symfony\Component\DependencyInjection\Attribute\Lazy;

class SuperAwesomeService
{
    public function __construct(#[Lazy] SomeOtherService $service)
    {
        // ...
    }
}

```

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Attribute/Lazy.php
+++ b/src/Symfony/Component/DependencyInjection/Attribute/Lazy.php
@@ -1,0 +1,21 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\DependencyInjection\Attribute;
+
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_PARAMETER)]
+class Lazy
+{
+    public function __construct(
+        public bool|string|null $lazy = true,
+    ) {
+    }
+}
--- a/src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
+++ b/src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
@@ -15,6 +15,7 @@ use Symfony\Component\Config\Resource\ClassExistenceResource;
 use Symfony\Component\DependencyInjection\Attribute\Autowire;
 use Symfony\Component\DependencyInjection\Attribute\AutowireCallable;
 use Symfony\Component\DependencyInjection\Attribute\AutowireDecorated;
+use Symfony\Component\DependencyInjection\Attribute\Lazy;
 use Symfony\Component\DependencyInjection\Attribute\Target;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\ContainerInterface;
@@ -299,7 +300,13 @@ class AutowirePass extends AbstractRecursivePass
             };
 
             if ($checkAttributes) {
-                foreach ($parameter->getAttributes(Autowire::class, \ReflectionAttribute::IS_INSTANCEOF) as $attribute) {
+                $attributes = array_merge($parameter->getAttributes(Autowire::class, \ReflectionAttribute::IS_INSTANCEOF), $parameter->getAttributes(Lazy::class, \ReflectionAttribute::IS_INSTANCEOF));
+
+                if (1 < \count($attributes)) {
+                    throw new AutowiringFailedException($this->currentId, 'Using both attributes #[Lazy] and #[Autowire] on an argument is not allowed; use the "lazy" parameter of #[Autowire] instead.');
+                }
+
+                foreach ($attributes as $attribute) {
                     $attribute = $attribute->newInstance();
                     $invalidBehavior = $parameter->allowsNull() ? ContainerInterface::NULL_ON_INVALID_REFERENCE : ContainerBuilder::EXCEPTION_ON_INVALID_REFERENCE;
 
--- a/src/Symfony/Component/DependencyInjection/Compiler/RegisterAutoconfigureAttributesPass.php
+++ b/src/Symfony/Component/DependencyInjection/Compiler/RegisterAutoconfigureAttributesPass.php
@@ -12,8 +12,10 @@
 namespace Symfony\Component\DependencyInjection\Compiler;
 
 use Symfony\Component\DependencyInjection\Attribute\Autoconfigure;
+use Symfony\Component\DependencyInjection\Attribute\Lazy;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Exception\AutoconfigureFailedException;
 use Symfony\Component\DependencyInjection\Loader\YamlFileLoader;
 
 /**
@@ -42,7 +44,16 @@ final class RegisterAutoconfigureAttributesPass implements CompilerPassInterface
 
     public function processClass(ContainerBuilder $container, \ReflectionClass $class): void
     {
-        foreach ($class->getAttributes(Autoconfigure::class, \ReflectionAttribute::IS_INSTANCEOF) as $attribute) {
+        $autoconfigure = $class->getAttributes(Autoconfigure::class, \ReflectionAttribute::IS_INSTANCEOF);
+        $lazy = $class->getAttributes(Lazy::class, \ReflectionAttribute::IS_INSTANCEOF);
+
+        if ($autoconfigure && $lazy) {
+            throw new AutoconfigureFailedException($class->name, 'Using both attributes #[Lazy] and #[Autoconfigure] on an argument is not allowed; use the "lazy" parameter of #[Autoconfigure] instead.');
+        }
+
+        $attributes = array_merge($autoconfigure, $lazy);
+
+        foreach ($attributes as $attribute) {
             self::registerForAutoconfiguration($container, $class, $attribute);
         }
     }
--- a/src/Symfony/Component/DependencyInjection/Tests/Fixtures/LazyAutoconfigured.php
+++ b/src/Symfony/Component/DependencyInjection/Tests/Fixtures/LazyAutoconfigured.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Symfony\Component\DependencyInjection\Tests\Fixtures;
+
+use Symfony\Component\DependencyInjection\Attribute\Autoconfigure;
+use Symfony\Component\DependencyInjection\Attribute\Lazy;
+
+#[Lazy, Autoconfigure(lazy: true)]
+class LazyAutoconfigured
+{
+}
--- a/src/Symfony/Component/DependencyInjection/Tests/Fixtures/includes/autowiring_classes_80.php
+++ b/src/Symfony/Component/DependencyInjection/Tests/Fixtures/includes/autowiring_classes_80.php
@@ -5,6 +5,7 @@ namespace Symfony\Component\DependencyInjection\Tests\Compiler;
 use Symfony\Component\DependencyInjection\Attribute\AsDecorator;
 use Symfony\Component\DependencyInjection\Attribute\Autowire;
 use Symfony\Component\DependencyInjection\Attribute\AutowireDecorated;
+use Symfony\Component\DependencyInjection\Attribute\Lazy;
 use Symfony\Component\DependencyInjection\Attribute\TaggedIterator;
 use Symfony\Component\DependencyInjection\Attribute\TaggedLocator;
 use Symfony\Component\DependencyInjection\ContainerInterface;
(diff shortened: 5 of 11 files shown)


nicolas-grekas (commented):
This made me wonder whether this blurs any lines but I think I'm good with the proposal.
The implementation should be updated to forbid using both Lazy and Autowire since on the attribute is not repeatable.

Tiriel:
> The implementation should be updated to forbid using both Lazy and Autowire since on the attribute is not repeatable.

Going to work on this asap, thanks!

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
```diff
@@ -299,7 +300,15 @@ private function autowireMethod(\ReflectionFunctionAbstract $reflectionMethod, a
...
             if ($checkAttributes) {
-                foreach ($parameter->getAttributes(Autowire::class, \ReflectionAttribute::IS_INSTANCEOF) as $attribute) {
+                $autowire = $parameter->getAttributes(Autowire::class, \ReflectionAttribute::IS_INSTANCEOF);
+                $lazy = $parameter->getAttributes(Lazy::class, \ReflectionAttribute::IS_INSTANCEOF);
+
+                if (\count($autowire) > 0 && \count($lazy) > 0) {
+                    throw new AutowiringFailedException($this->currentId, "'Lazy' and 'Autowire' attributes cannot be used on the same argument.");
+                }
+
+                $attributes = $autowire + $lazy;
```
nicolas-grekas:
```suggestion
                $attributes = array_merge($parameter->getAttributes(Autowire::class, \ReflectionAttribute::IS_INSTANCEOF), $parameter->getAttributes(Lazy::class, \ReflectionAttribute::IS_INSTANCEOF));

                if (1 < \count($attributes)) {
                    throw new AutowiringFailedException($this->currentId, 'Using both attributes #[Lazy] and #[Autowire] on an argument is not allowed; use the "lazy" parameter of #[Autowire] instead.");
                }
```

src/Symfony/Component/DependencyInjection/Compiler/RegisterAutoconfigureAttributesPass.php
```diff
@@ -42,7 +43,8 @@ public function accept(Definition $definition): bool
 
     public function processClass(ContainerBuilder $container, \ReflectionClass $class): void
     {
-        foreach ($class->getAttributes(Autoconfigure::class, \ReflectionAttribute::IS_INSTANCEOF) as $attribute) {
+        $attributes = $class->getAttributes(Autoconfigure::class, \ReflectionAttribute::IS_INSTANCEOF) + $class->getAttributes(Lazy::class, \ReflectionAttribute::IS_INSTANCEOF);
```
nicolas-grekas: we should use array_merge instead of the `+` operator so that we don't loose attributes
can you please add a test case to cover this also?

src/Symfony/Component/DependencyInjection/Tests/Fixtures/includes/autowiring_classes_80.php
```diff
@@ -125,3 +126,17 @@ public function __construct(
...
+class LazyServiceAttributeAutowiring
+{
+    public function __construct(#[Lazy] A $a)
+    {
+    }
+}
+
+class LazyAutowireServiceAttributesAutowiring
+{
+    public function __construct(#[Lazy]#[Autowire(lazy: true)] A $a)
```
nicolas-grekas:
```suggestion
    public function __construct(#[Lazy, Autowire(lazy: true)] A $a)
```

src/Symfony/Component/DependencyInjection/Compiler/RegisterAutoconfigureAttributesPass.php
Tiriel: Will do!
Does this case also need a check to not use `#[Lazy]` when `#[Autoconfigure(lazy: true)]` is on, or should we just ignore the `#[Lazy]` in this case?

nicolas-grekas: I wouldn't do anything in case both attributes are there since we also allow two Autoconfigure attributes and we already have a rule to take many into account.

Tiriel: Added a test to check adding both still works.
Just to be sure, what other cases did you want to check?

src/Symfony/Component/DependencyInjection/Compiler/RegisterAutoconfigureAttributesPass.php
```diff
@@ -42,7 +43,8 @@ public function accept(Definition $definition): bool
 
     public function processClass(ContainerBuilder $container, \ReflectionClass $class): void
     {
-        foreach ($class->getAttributes(Autoconfigure::class, \ReflectionAttribute::IS_INSTANCEOF) as $attribute) {
+        $attributes = array_merge($class->getAttributes(Autoconfigure::class, \ReflectionAttribute::IS_INSTANCEOF), $class->getAttributes(Lazy::class, \ReflectionAttribute::IS_INSTANCEOF));
```
nicolas-grekas: OK, second though: let's make them conflict :)
I was reviewing and considering that we now ignore the order of declaration of attributes when both are set.
We could preserve the order but that'd make the code more complex.
Let's add the same conflict rule that we have for Autowire and keep things simple.

src/Symfony/Component/DependencyInjection/Compiler/RegisterAutoconfigureAttributesPass.php
```diff
@@ -42,7 +44,12 @@ public function accept(Definition $definition): bool
...
+
+        if (1 < \count($attributes)) {
```
nicolas-grekas: the implem cannot be the same because it is allowed to have many autoconfigure attributes ;)

Tiriel: Yeah figured the same while re reading, sorry for the "too soon". working on it

src/Symfony/Component/DependencyInjection/Compiler/RegisterAutoconfigureAttributesPass.php
```diff
@@ -42,7 +44,16 @@ public function accept(Definition $definition): bool
 
     public function processClass(ContainerBuilder $container, \ReflectionClass $class): void
     {
-        foreach ($class->getAttributes(Autoconfigure::class, \ReflectionAttribute::IS_INSTANCEOF) as $attribute) {
+        $autoconfigure = $class->getAttributes(Autoconfigure::class, \ReflectionAttribute::IS_INSTANCEOF);
+        $lazy = $class->getAttributes(Lazy::class, \ReflectionAttribute::IS_INSTANCEOF);
+
+        if (0 < \count($autoconfigure) && 0 < \count($lazy)) {
```
nicolas-grekas:
```suggestion
        if ($autoconfigure && $lazy) {
```

nicolas-grekas (commented):
LGTM but CS, see also fabbot

src/Symfony/Component/DependencyInjection/Attribute/Lazy.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
```
stloyd: Duplicated header:
```suggestion
/*
 * This file is part of the Symfony package.
 *
 * (c) Fabien Potencier <fabien@symfony.com>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */
```

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
```diff
@@ -299,7 +300,13 @@ private function autowireMethod(\ReflectionFunctionAbstract $reflectionMethod, a
             };
 
             if ($checkAttributes) {
-                foreach ($parameter->getAttributes(Autowire::class, \ReflectionAttribute::IS_INSTANCEOF) as $attribute) {
+                $attributes = array_merge($parameter->getAttributes(Autowire::class, \ReflectionAttribute::IS_INSTANCEOF), $parameter->getAttributes(Lazy::class, \ReflectionAttribute::IS_INSTANCEOF));
+
+                if (1 < \count($attributes)) {
```
stloyd, line 305:
```suggestion
                if (!$attributes) {
```

src/Symfony/Component/DependencyInjection/Tests/Fixtures/LazyAutoconfigured.php
```diff
@@ -0,0 +1,12 @@
...
+
+namespace Symfony\Component\DependencyInjection\Tests\Fixtures;
+
+use Symfony\Component\DependencyInjection\Attribute\Autoconfigure;
+use Symfony\Component\DependencyInjection\Attribute\Lazy;
+
+#[Lazy, Autoconfigure(lazy: true)]
+class LazyAutoconfigured
+{
+
```
nicolas-grekas: you should apply the patch provided by fabbot

Tiriel: I did! This didn't register in it

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
Tiriel: Can't, we don't want to check if it's empty, we want to check we have max 1 attribute, `Lazy` or `Autowire` but not both.

src/Symfony/Component/DependencyInjection/Tests/Fixtures/LazyAutoconfigured.php
nicolas-grekas: ah yes, fixtures are excluded
no empty lines as body please

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
```diff
@@ -299,7 +300,13 @@ private function autowireMethod(\ReflectionFunctionAbstract $reflectionMethod, a
             };
 
             if ($checkAttributes) {
-                foreach ($parameter->getAttributes(Autowire::class, \ReflectionAttribute::IS_INSTANCEOF) as $attribute) {
+                $attributes = array_merge($parameter->getAttributes(Autowire::class, \ReflectionAttribute::IS_INSTANCEOF), $parameter->getAttributes(Lazy::class, \ReflectionAttribute::IS_INSTANCEOF));
+
+                if (1 > \count($attributes)) {
```
nicolas-grekas:
```suggestion
                if (1 < \count($attributes)) {
```

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
```diff
@@ -374,7 +381,7 @@ private function autowireMethod(\ReflectionFunctionAbstract $reflectionMethod, a
                     $type = ProxyHelper::exportType($parameter);
                     $type = $type ? sprintf('is type-hinted "%s"', preg_replace('/(^|[(|&])\\\\|^\?\\\\?/', '\1', $type)) : 'has no type-hint';
 
-                    throw new AutowiringFailedException($this->currentId, sprintf('Cannot autowire service "%s": argument "$%s" of method "%s()" %s, you should configure its value explicitly.', $this->currentId, $parameter->name, $class !== $this->currentId ? $class.'::'.$method : $method, $type));
+                    throw new AutowiringFailedException($this->currentId, sprintf('Cannot autowire service "%s": argument "$%s" of method "%s()" "%s", you should configure its value explicitly.', $this->currentId, $parameter->name, $class !== $this->currentId ? $class.'::'.$method : $method, $type));
```
nicolas-grekas: sorry this one is a false positive

src/Symfony/Component/DependencyInjection/Tests/Fixtures/includes/autowiring_classes_80.php
nicolas-grekas: unresolving as the suggestion has not been applied and I see no reason why

src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php
Tiriel: Dammit sorry, forgot to remove that, thanks for the catch

src/Symfony/Component/DependencyInjection/Tests/Fixtures/includes/autowiring_classes_80.php
Tiriel: Sorry, mistook it for the other one. Fixed!

nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @Tiriel.

PR #52924, opened 2023-12-07
[FrameworkBundle] add a private_ranges shortcut for trusted_proxies

### LINKED ISSUE #51826: [Framework] trusted_proxies: private_ranges
### Description

Currently, if your app is behind a reverse proxy, you need to enable trusted proxies to be sure Symfony will understand it's using HTTPS.

Typically it looks like this:

```yaml
framework:
    # localhost, Docker, Kubernetes
    trusted_proxies: 127.0.0.1, 172.16.0.0/12, 10.0.0.0/8
```

[Caddy solves the same issue by having a dedicated alias called `private_ranges`](https://caddyserver.com/docs/caddyfile/options#trusted-proxies):

> As a shortcut, private_ranges can be used to match all private IPv4 and IPv6 ranges. It's the same as specifying all of these ranges: 192.168.0.0/16 172.16.0.0/12 10.0.0.0/8 127.0.0.1/8 fd00::/8 ::1

Proposal is to introduce the same alias for these specific values to allow easier configuration in this circumstance.

### Example

```yaml
framework:
    trusted_proxies: private_ranges
```

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #51826
| License       | MIT

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
@@ -52,6 +52,7 @@ use Symfony\Component\HtmlSanitizer\HtmlSanitizerInterface;
 use Symfony\Component\HttpClient\MockHttpClient;
 use Symfony\Component\HttpClient\RetryableHttpClient;
 use Symfony\Component\HttpClient\ScopingHttpClient;
+use Symfony\Component\HttpFoundation\IpUtils;
 use Symfony\Component\HttpKernel\DependencyInjection\LoggerPass;
 use Symfony\Component\HttpKernel\Fragment\FragmentUriGeneratorInterface;
 use Symfony\Component\Messenger\Bridge\AmazonSqs\Transport\AmazonSqsTransportFactory;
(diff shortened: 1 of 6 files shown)


danielburger1337:
There are some scenarios where additional private IPs are used for reverse proxies. For example I know of multiple setups where the forwarded proxy is 127.0.0.1 (mainly Nginx -> Apache on the same machine without containerization).

There is also the case of azure load balancer. I don't know if they still actively deploy this or not but at least they used to use a little known feature called IPv6 "IPv4 mapped addresses" `::ffff:0:0/96`. 

Wouldn't it make sense to add this to the shortcut as well? And if yes, why not just use the `IpUtils::PRIVATE_SUBNETS` constant? symfony/framework-bundle has a dependency on http-foundation anyways.

xabbuh:
> Wouldn't it make sense to add this to the shortcut as well? And if yes, why not just use the `IpUtils::PRIVATE_SUBNETS` constant? symfony/framework-bundle has a dependency on http-foundation anyways.

sounds reasonable to me, I have changed the code accordingly

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
```diff
@@ -2264,6 +2265,13 @@ public function testNotifierWithSpecificMessageBus()
         $this->assertEquals(new Reference('app.another_bus'), $container->getDefinition('notifier.channel.sms')->getArgument(1));
     }
 
+    public function testTrustedProxiesWithPrivateRanges()
+    {
+        $container = $this->createContainerFromFile('trusted_proxies_private_ranges');
+
+        $this->assertEquals(IpUtils::PRIVATE_SUBNETS, array_map('trim', explode(',', $container->getParameter('kernel.trusted_proxies'))));
```
nicolas-grekas: assertSame?

nicolas-grekas (approved):
LGTM, rebase needed

nicolas-grekas:
Thank you @xabbuh.

PR #52928, opened 2023-12-07
[Dotenv] Allow environment variables starting with an underscore

### LINKED ISSUE #52630: Unable to create environement variables that starts with an underscore ( `_` )

6.3.7 ( Dotenv component )

### Description

I've been trying to setup a symfony project and it requires me to have a `.env` file that contains variable that starts with an underscore. For example:

```
_APP_ID=123
```

However, when I'm trying to parse this file, to use the variable in my code, I'm having the follow error message.

> Uncaught Symfony\Component\Dotenv\Exception\FormatException: Invalid character in variable name in "/path/to/my/project/.env"

This is unexpected since it looks like [POSIX variable should be able to start with an underscore](https://stackoverflow.com/a/2821201/5784924).

### How to reproduce

1) Create a new symfony project
2) In this project, update the `.env` file to add the following variable
```
_TEST_VARIABLE=result
```
3) Create a route that will retrieve the variable's value using the `getenv` function
4) Start the project
5) In your browser, call the previously created route
6) In your `serve` console, notice the error.

### Possible Solution

I've done some digging around and it looks like all we need is to allow the underscore character, in the first characters set of the `VARNAME_REGEX` of the `Dotenv` class. 

I will create a PR that fixes the problem.

### Additional Context

Here is a screenshot of the console with the error
image

And here is the complete `.env` file. There is no other `.env` file overriding it.

```
# In all environments, the following files are loaded if they exist,
# the latter taking precedence over the former:
#
#  * .env                contains default values for the environment variables needed by the app
#  * .env.local          uncommitted file with local overrides
#  * .env.$APP_ENV       committed environment-specific defaults
#  * .env.$APP_ENV.local uncommitted environment-specific overrides
#
# Real environment variables win over .env files.
#
# DO NOT DEFINE PRODUCTION SECRETS IN THIS FILE NOR IN ANY OTHER COMMITTED FILES.
# https://symfony.com/doc/current/configuration/secrets.html
#
# Run "composer dump-env prod" to compile .env files for production use (requires symfony/flex >=1.2).
# https://symfony.com/doc/current/best_practices.html#use-environment-variables-for-infrastructure-configuration

###> symfony/framework-bundle ###
APP_ENV=dev
APP_SECRET=secret
###< symfony/framework-bundle ###

# The variable that causes problem
_TEST_ENV_1=default

```

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52630
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Dotenv/Dotenv.php
+++ b/src/Symfony/Component/Dotenv/Dotenv.php
@@ -25,7 +25,7 @@ use Symfony\Component\Process\Process;
  */
 final class Dotenv
 {
-    public const VARNAME_REGEX = '(?i:[A-Z][A-Z0-9_]*+)';
+    public const VARNAME_REGEX = '(?i:_?[A-Z][A-Z0-9_]*+)';
     public const STATE_VARNAME = 0;
     public const STATE_VALUE = 1;
 
@@ -351,8 +351,8 @@ final class Dotenv
                 ++$this->cursor;
                 $value = str_replace(['\\"', '\r', '\n'], ['"', "\r", "\n"], $value);
                 $resolvedValue = $value;
-                $resolvedValue = $this->resolveVariables($resolvedValue, $loadedVars);
                 $resolvedValue = $this->resolveCommands($resolvedValue, $loadedVars);
+                $resolvedValue = $this->resolveVariables($resolvedValue, $loadedVars);
                 $resolvedValue = str_replace('\\\\', '\\', $resolvedValue);
                 $v .= $resolvedValue;
             } else {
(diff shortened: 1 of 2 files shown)


OskarStark:
This sounds like a new feature to me

src/Symfony/Component/Dotenv/Dotenv.php
```diff
@@ -25,7 +25,7 @@
  */
 final class Dotenv
 {
-    public const VARNAME_REGEX = '(?i:[A-Z][A-Z0-9_]*+)';
+    public const VARNAME_REGEX = '(?i:[A-Z_][A-Z0-9_]*+)';
```
fabpot: This new regexp makes `_` a valid env var, which is wrong. Can you fix the regex and add a test to make sure that `_` is not allowed?

fabpot (changes requested):

src/Symfony/Component/Dotenv/Dotenv.php
OskarStark: This should do the trick, except the mentioned double quotes issue:
https://regex101.com/r/BK7AN6/1

xabbuh: this should be fixed now

derrabus (approved):
LGTM

Following @fabpot's remark, shall we add a test to make sure `_` is not a valid env var?

OskarStark:
There is a test already

fabpot (approved):

fabpot:
Thank you @xabbuh.

PR #52930, opened 2023-12-07
[Messenger] Fix Redis messenger scheme comparison

### LINKED ISSUE #52899: Redis Sentinel messenger implementation not allowing multiple sentinel server / wrong documentation

6.3.10

### Description

The redis sentinel DSN is documented as
```
MESSENGER_DSN=redis:?host[host1:5000]&host[host2:5000]&host[host3:5000]&sentinel_master=test&dbindex=0
```
While this works for the cache DSN, it doesn't for the messenger. After a lot of debugging, I came up with this working syntax:
```
MESSENGER_DSN=redis://host1:5000?sentinel_master=test&dbindex=1
```

Unfortunately, this doesn't allow to set multiple sentinel hosts in case one is down.

### How to reproduce

Set 
```
MESSENGER_DSN=redis:?host[host1:5000]&host[host2:5000]&host[host3:5000]&sentinel_master=test&dbindex=0
```
and try to start `console messenger:consume` (or try to queue something via messenger).

### Possible Solution

Use a similar redis sentinel implementation as in cache would be peferably the best outcome, but a fixed documentation would improve the situation as well.

### Additional Context

_No response_

freswa:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Issues        | Fix #52899 
| License       | MIT

Commit 3380518b88f8303ad9428f26c103e6b3b64a9fc5 introduced a new Redis Sentinel DSN for the redis messenger transport which uses a scheme syntax like `redis:?host[rs:1234]&host[rs2:1234]`.
Though, the coresponding factory only supports schemes which start with `redis://` or `rediss://` which renders the redis sentinel features for the messenger unusable. This commit fixes the supported schemes by removing the `//` portion of them.

fixes #52899

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Redis/Tests/Transport/RedisTransportFactoryTest.php
+++ b/src/Symfony/Component/Messenger/Bridge/Redis/Tests/Transport/RedisTransportFactoryTest.php
@@ -28,22 +28,48 @@ class RedisTransportFactoryTest extends TestCase
 
         $this->assertTrue($factory->supports('redis://localhost', []));
         $this->assertTrue($factory->supports('rediss://localhost', []));
+        $this->assertTrue($factory->supports('redis:?host[host1:5000]&host[host2:5000]&host[host3:5000]&sentinel_master=test&dbindex=0', []));
         $this->assertFalse($factory->supports('sqs://localhost', []));
         $this->assertFalse($factory->supports('invalid-dsn', []));
     }
 
     /**
      * @group integration
+     *
+     * @dataProvider createTransportProvider
      */
-    public function testCreateTransport()
+    public function testCreateTransport(string $dsn, array $options = [])
     {
         $this->skipIfRedisUnavailable();
 
         $factory = new RedisTransportFactory();
         $serializer = $this->createMock(SerializerInterface::class);
-        $expectedTransport = new RedisTransport(Connection::fromDsn('redis://'.getenv('REDIS_HOST'), ['stream' => 'bar', 'delete_after_ack' => true]), $serializer);
 
-        $this->assertEquals($expectedTransport, $factory->createTransport('redis://'.getenv('REDIS_HOST'), ['stream' => 'bar', 'delete_after_ack' => true], $serializer));
+        $this->assertEquals(
+            new RedisTransport(Connection::fromDsn($dsn, $options), $serializer),
+            $factory->createTransport($dsn, $options, $serializer)
+        );
+    }
+
+    /**
+     * @return iterable<array{0: string, 1: array}>
+     */
+    public static function createTransportProvider(): iterable
+    {
+        yield 'scheme "redis" without options' => [
+            'redis://'.getenv('REDIS_HOST'),
+            [],
+        ];
+
+        yield 'scheme "redis" with options' => [
+            'redis://'.getenv('REDIS_HOST'),
+            ['stream' => 'bar', 'delete_after_ack' => true],
+        ];
+
+        yield 'redis_sentinel' => [
+            'redis:?host['.str_replace(' ', ']&host[', getenv('REDIS_SENTINEL_HOSTS')).']',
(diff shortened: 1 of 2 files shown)


xabbuh:
Can you please add a test (for `supports()` as well as for `createTransport()`)?.

OskarStark:
I pushed some tests

src/Symfony/Component/Messenger/Bridge/Redis/Tests/Transport/RedisTransportFactoryTest.php
```diff
@@ -28,22 +28,59 @@ public function testSupportsOnlyRedisTransports()
...
+            ['stream' => 'bar', 'delete_after_ack' => true],
+        ];
+
+        yield 'scheme "rediss" with options' => [
+            'rediss://'.getenv('REDIS_HOST'),
+            ['stream' => 'bar', 'delete_after_ack' => true],
+        ];
+
+        yield 'redis_sentinel' => [
+            'redis:?host[host1:5000]&host[host2:5000]&host[host3:5000]&sentinel_master=test&dbindex=0',
```
OskarStark: as we used `getenv('REDIS_HOST')` before I think we need a real connection and I am not sure if we can connect using this sentinel stuff 🤔

freswa: Sentinel has its own protocol, which is sort of a telnet connection (as well as the redis core protocol). But here we just connect to ask for the current master in the cluster which, since sentinel can handle multiple clusters, is set in the `sentinel_master` parameter. Sentinel returns an IP and we connect to that IP with the usual Redis protocol.
So if you want to connect for a `PING`, you can do so to a sentinel as well. Almost all other commands differ and would require to setup a sentinel instance.

OskarStark: Can you please adjust my code to make it work? I have never worked with sentinel

freswa: The sentinel code is fixed, but the `rediss` schemes still fail. I barely found any ssl tests. Any idea @OskarStark? We may have to add another process (with ssl enabled) to the pipeline...?

OskarStark: @alexandre-daubois do you know anything related?

alexandre-daubois: We don't have Redis Secure (accepting the `rediss` scheme) setup in the integration workflow. I'll try to have a look at it, it would be interesting to add it! But that would definitely be for a separate PR. I think we can pass on this case :man_shrugging:

alexandre-daubois: We don't have Redis Secure instance (accepting the `rediss` scheme) setup in the integration workflow. I'll try to have a look at it, it would be interesting to add it! But that would definitely be for a separate PR. I think we can pass on this case :man_shrugging:

OskarStark: ok lets remove it then, or skip it for now with a meaningful comment

freswa: Removed.

nicolas-grekas:
Please squash your PR and/or allow edits from maintainers.

nicolas-grekas (approved):

freswa:
Squashed.

nicolas-grekas:
Thank you @freswa.

PR #52937, opened 2023-12-07
[WebProfilerBundle] Fix "Copy as cURL"

### LINKED ISSUE #52920: [Profiler][HttpClient] „Copy as cURL” button missing in Symfony 6.4.0 — 6.4.1

6.4.0, 6.4.1

### Description

When using the `HttpClientInterface` in Symfony <=6.3.8, you had a nice little „Copy as cURL” button within the Profiler's „HTTP Client” section. This button is missing since Symfony 6.4.0 and still missing in 6.4.1.

This feature has been introduced in 6.1.0: https://symfony.com/blog/new-in-symfony-6-1-profiler-improvements-part-2#copying-http-client-requests-as-curl-commands

### How to reproduce

- Use the `HttpClientInterface` to make any request
- Open the web profiler, go the „HTTP Client” section
- See the missing button

### Possible Solution

I had a quick glimpse at the changed Twig templates and the `http_client.html.twig` template still contains the necessary button: https://github.com/symfony/web-profiler-bundle/blob/14752d3fb77c3c69b6cee7c03c06e2d6494a196b/Resources/views/Collector/http_client.html.twig#L132

The `curlCommand` trace is also present: https://github.com/symfony/http-client/blob/db714986d3b84330bb6196fdb201c9f79b3a8853/DataCollector/HttpClientDataCollector.php#L171

I can also confirm that I'm using a `Symfony\Component\HttpClient\TraceableHttpClient` to make my HTTP requests and that the `HttpClientDataCollector`'s logic is being called that generates the `curlCommand` information for the traces.

## Working in Symfony 6.3 (all dependencies at their latest ~6.3 version)

CleanShot 2023-12-06 at 21 54 10@2x

## Missing in Symfony 6.4.0 / 6.4.1

CleanShot 2023-12-06 at 21 55 11@2x

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52920
| License       | MIT

The bug was introduced in #50790.

#SymfonyHackday

### DIFF
--- a/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/base_js.html.twig
+++ b/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/base_js.html.twig
@@ -13,6 +13,7 @@
         constructor() {
             this.#createTabs();
             this.#createToggles();
+            this.#createCopyToClipboard();
             this.#convertDateTimesToUserTimezone();
         }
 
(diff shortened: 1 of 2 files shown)


src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/base_js.html.twig
```diff
@@ -161,18 +162,39 @@
...
+                copyToClipboardElement.addEventListener('click', (e) => {
+                    /* Prevents from disallowing clicks on "copy to clipboard" elements inside toggles */
+                    e.stopPropagation();
+
+                    navigator.clipboard.writeText(copyToClipboardElement.getAttribute('data-clipboard-text'));
+
+                    let oldContent = copyToClipboardElement.textContent;
+
+                    copyToClipboardElement.textContent = `✅ Copied!`;
+                    copyToClipboardElement.classList.add('status-success');
```
keichinger: Excellent improvement over the previous logic. One tiny issue: If you're using the dark theme, the text color will be black, which makes it hard/impossible to read. The default state itself is fine, just the `status-success` state needs a `color`-update.

Default state:

CleanShot 2023-12-08 at 22 20 01@2x

With focus state:

CleanShot 2023-12-08 at 22 19 37@2x

Without focus state:

CleanShot 2023-12-08 at 22 19 39@2x

keichinger (approved):
This indeed fixes the issue I'm having. From my perspective, this one is good to go as soon as the text color in the copied state has been addressed.

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/base_js.html.twig
HypeMC: Thanks for the notice, unfortunately, I'm not sure about the expected design here. Maybe @javiereguiluz can give us a clue.

javiereguiluz: I'd say we don't need to change the color when copying things.

Just changing the button text to "Copied" is enough. That's e.g. what GitHub does, so everybody is used to that behavior.

Thanks.

welcoMattic (approved):

xabbuh (approved):

OskarStark (approved):

fabpot:
Thank you @HypeMC.

PR #27831, opened 2018-07-03
 Check for Hyper terminal on all operating systems. 

azjezz:
| Q             | A
| ------------- | ---
| Branch?       | 2.8
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no   
| Deprecations? | no 
| Tests pass?   | yes   
| License       | MIT

return true if the terminal is Hyper.is on all operating systems not limited to windows.

PR #27794, opened 2018-07-01
Add color support for Hyper terminal .

azjezz:
| Q             | A
| ------------- | ---
| Branch?       | master 
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no    
| Deprecations? | no
| Tests pass?   | yes    
| License       | MIT

Added color support for [Hyper terminal](https://hyper.is), also shade characters support for progress bar. 

image

PR #52940, opened 2023-12-08
[Console] Fix color support check on non-Windows platforms

### LINKED ISSUE #45917: CLI colors are broken in Docker

5.4.5

### Description

see https://github.com/composer/composer/issues/10691#issuecomment-1086616111 /w repro DockerFile

### Possible Solution

colors supported correctly

theofidry:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #45917
| License       | MIT

Currently checking the color support based on `ANSICON`, `ConEmuANSI=ON` or `TERM=xTerm` is done only for Widows. I could not find any reason as to why and it does not make much sense as it is. Especially if we consider that `TERM=xTerm` is a term check and we do another one (not Widows specific) which is `TERM_PROGRAM=Hyper`.

This potentially fixes #45917.

This also looks more in line with the intent (based on the title) of #27831 and #27794.

### DIFF
--- a/src/Symfony/Component/Console/Output/StreamOutput.php
+++ b/src/Symfony/Component/Console/Output/StreamOutput.php
@@ -95,15 +95,17 @@ class StreamOutput extends Output
             return false;
         }
 
-        if (\DIRECTORY_SEPARATOR === '\\') {
-            return (\function_exists('sapi_windows_vt100_support')
-                && @sapi_windows_vt100_support($this->stream))
-                || false !== getenv('ANSICON')
-                || 'ON' === getenv('ConEmuANSI')
-                || str_starts_with((string) getenv('TERM'), 'xterm');
+        if (\DIRECTORY_SEPARATOR === '\\'
+            && \function_exists('sapi_windows_vt100_support')
+            && @sapi_windows_vt100_support($this->stream)
+        ) {
+            return true;
         }
 
         return 'Hyper' === getenv('TERM_PROGRAM')
+            || false !== getenv('ANSICON')
+            || 'ON' === getenv('ConEmuANSI')
+            || str_starts_with((string) getenv('TERM'), 'xterm')
             || stream_isatty($this->stream);
     }
 }


src/Symfony/Component/Console/Output/StreamOutput.php
```diff
@@ -97,18 +97,17 @@ protected function hasColorSupport(): bool
...
-                || false !== getenv('ANSICON')
-                || 'ON' === getenv('ConEmuANSI')
-                || 'xterm' === getenv('TERM');
-        }
-
-        return stream_isatty($this->stream);
+        return 'Hyper' === getenv('TERM_PROGRAM')
+            || false !== getenv('ANSICON')
+            || 'ON' === getenv('ConEmuANSI')
+            || 'xterm' === getenv('TERM')
```
theofidry: as noted with @tucksaun, this also looks incorrect as you could have `'xterm-256'` instead of `'xterm'`

OskarStark:
The referenced issue mentioned 5.4, should this PR target 5.4 too?

src/Symfony/Component/Console/Output/StreamOutput.php
tucksaun: actually `xterm` AND `xterm-256`

OskarStark:
```suggestion
            || str_starts_with(getenv('TERM'), 'xterm')
```

theofidry: Extracted this fix in #52941.

theofidry:
>The referenced issue mentioned 5.4, should this PR target 5.4 too?

Ah indeed, my bad I forgot 5.4 is still supported.

theofidry:
Waiting on #52941 and #52942.

theofidry:
Yo @carsonbot not cool

theofidry:
thanks @xabbuh ❤️

fabpot (approved):

fabpot:
Thank you @theofidry.

mallardduck:
I may have encountered a bug in my application that seems to be related to this PR.

The "short version" of the issue is that I have CLI commands who's output get piped to other commands. And this has stopped working when updating to [v6.4.2](https://github.com/symfony/console/releases/tag/v6.4.2).

From what I can tell debugging the commands output with `cat -v` - I see additional color codes on the strings after this PR that break subsequent commands.

In both the before and after, the text will print in green (info) colors. However with 6.4.1 the text pipes to subsequent commands fine. But in 6.4.2 the text has color codes embedded to it which breaks subsequent commands.

```
^[[32mTHIS IS THE TEXT^[[39m
```

In my environment I'm on a Mac, using iTerm.
```
echo $TERM
xterm-256color
echo $TERM_PROGRAM
iTerm.app
```

Ultimately I'm wondering if this is expected in my environment after this change, or if this is a new bug/regression? If this is expected, how can I continue to use the `info()` output text and maintain ability to pipe text?

xabbuh:
@mallardduck Can you check if the behaviour you experience vanishes with #53576?

mallardduck:
@xabbuh - thanks for that, applying that patch manually to the vendor files fixes it.

tamr:
Just updated the composer and noticed the missing colors. I'm using Kitty(cyd01) with termtype set to tmux-256color.
Shouldn't there be another check?
str_ends_with(getenv('TERM'), '256color ')

theofidry:
@tamr this would require more information. `tmux-256color` is captured by [this regex](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Console/Output/StreamOutput.php#L120C29-L120C114) so there is likely something else going on (provided the version you are using has this patch).

If you do please open an issue with the required information to fix it 🙏

PR #52941, opened 2023-12-08
[Console] Fix xterm detection

theofidry:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | None reported
| License       | MIT

Currently `StreamOutput::hasColorSupport()` returns `true` if `TERM=xterm`. However xterm may have different values such as `xterm-256color`.

Technically I think it would be more correct to check `xterm` or starts with `xterm-`, but I am not convinced there is a need to do so and one may prefer the (tiny) performance advantage? This can easily be changed though.

### DIFF
--- a/src/Symfony/Component/Console/Output/StreamOutput.php
+++ b/src/Symfony/Component/Console/Output/StreamOutput.php
@@ -64,9 +64,6 @@ class StreamOutput extends Output
         return $this->stream;
     }
 
-    /**
-     * {@inheritdoc}
-     */
     protected function doWrite(string $message, bool $newline)
     {
         if ($newline) {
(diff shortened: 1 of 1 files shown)


maxbeckers (approved):

OskarStark (approved):

OskarStark:
5.4?

theofidry:
@OskarStark did I misunderstand? I assumed you meant since 5.4 is maintained fixes should go there rather than 6.4.

fabpot (approved):

fabpot:
Thank you @theofidry.

PR #52942, opened 2023-12-08
Refactor hyper check location

theofidry:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | None
| License       | MIT

Extracted from #52940 to reduce the noise of that PR. There is no logic change, only a potentially a performance (I assume non-existent given it's a if check of a constant). If it is a concern however, this could be checked _before_ the windows specific checks.

### DIFF
--- a/src/Symfony/Component/Console/Output/StreamOutput.php
+++ b/src/Symfony/Component/Console/Output/StreamOutput.php
@@ -98,10 +98,6 @@ class StreamOutput extends Output
             return false;
         }
 
-        if ('Hyper' === getenv('TERM_PROGRAM')) {
-            return true;
-        }
-
         if (\DIRECTORY_SEPARATOR === '\\') {
             return (\function_exists('sapi_windows_vt100_support')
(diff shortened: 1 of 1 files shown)


src/Symfony/Component/Console/Output/StreamOutput.php
```diff
@@ -110,6 +106,7 @@ protected function hasColorSupport()
                 || 'xterm' === getenv('TERM');
         }
 
-        return stream_isatty($this->stream);
+        return 'Hyper' === getenv('TERM_PROGRAM'
```
maxbeckers:
```suggestion
        return 'Hyper' === getenv('TERM_PROGRAM')
```

theofidry: Thanks for the catch

maxbeckers (approved):

fabpot (approved):

fabpot:
Thank you @theofidry.

PR #52944, opened 2023-12-08
[Mailer] Be able to create a stream without TLS

aslangery:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | --
| License       | MIT

It was not possible to create a stream without TLS


OskarStark:
Can you please add a test case? Thanks

OskarStark:
Why not using `àWhy don't you using `smtps` as scheme instead? Can you explain?

aslangery:
> Why not using `àWhy don't you using `smtps` as scheme instead? Can you explain?

Our client has an incorrectly configured mail server. On swiftmailer this did not cause problems, but when switching to symfonymailer the connection stopped working

OskarStark:
I am not sure we want to work around wrong configured mail servers 🤔 

cc @fabpot

aslangery:
> I am not sure we want to work around wrong configured mail servers 🤔
> 
> cc @fabpot

This isn't really a mistake. On servers you can configure an arbitrary port for smtp. Even 465, although it is the default for smtps

nicolas-grekas:
This might break existing client that forgot to add the final "s" to their scheme so I'm not sure I would do it this way.
Instead, we could add a `tls` option - maybe by tweaking #53621
Closing for this reason, let's continue on the linked PR.

PR #52946, opened 2023-12-08
[HttpClient] Add HttpOptions->addHeader as a shortcut to add an header in an existing options object

Dean151:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | 
| License       | MIT

Currently, the HttpOptions object only provide a ->setHeaders option that override every existing headers. When we are provided an HttpOptions to customize a request, adding a request header requires to get the whole headers array, update it, and then set it back in the HttpOptions object. This DX improvement brings an addHeader option, allowing to add an header to the set without having to manage the existing ones.

When the header option is not yet provided, the ??= operator makes sure the headers is initialized with an empty array (PHP >= 7.4) When a header is already provided, it is overwritten.

### DIFF
--- a/src/Symfony/Component/HttpClient/HttpOptions.php
+++ b/src/Symfony/Component/HttpClient/HttpOptions.php
@@ -66,6 +66,17 @@ class HttpOptions
     /**
      * @return $this
      */
+    public function addHeader(string $key, string $value): static
+    {
+        $this->options['headers'] ??= [];
+        $this->options['headers'][$key] = $value;
+
+        return $this;
+    }
+
+    /**
+     * @return $this
+     */
     public function setHeaders(iterable $headers): static
     {
         $this->options['headers'] = $headers;
--- a/src/Symfony/Component/HttpClient/Tests/HttpOptionsTest.php
+++ b/src/Symfony/Component/HttpClient/Tests/HttpOptionsTest.php
@@ -39,4 +39,15 @@ class HttpOptionsTest extends TestCase
     {
         $this->assertSame('foobar', (new HttpOptions())->setAuthBearer('foobar')->toArray()['auth_bearer']);
     }
+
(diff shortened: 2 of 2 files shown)


src/Symfony/Component/HttpClient/HttpOptions.php
```diff
@@ -63,6 +63,17 @@ public function setQuery(array $query): static
         return $this;
     }
 
+    /**
+     * @return $this
+     */
```
welcoMattic, line 68: ~~You can remove this PHP doc as we have native return type here~~

welcoMattic (approved):
Thanks and congrats for your first contribution 👍

src/Symfony/Component/HttpClient/HttpOptions.php
```diff
@@ -63,9 +63,6 @@ public function setQuery(array $query): static
         return $this;
     }
 
-    /**
-     * @return $this
```
nicolas-grekas, line 67: this was fine, @welcoMattic had too much :beer: ;)

src/Symfony/Component/HttpClient/HttpOptions.php
welcoMattic: Rollback, sorry. In fact `$this` is more precise than `static`

Dean151: Reverted ;)

src/Symfony/Component/HttpClient/HttpOptions.php
Dean151: Reverted ;)

fabpot (approved):

fabpot:
Thank you @Dean151.

PR #52948, opened 2023-12-08
Use faster hashing algorithms when possible

javiereguiluz:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

`xxh128` is 60 times faster than `sha256` (source: https://php.watch/versions/8.1/xxHash) but it's not cryptographically secure.

In previous versions we already used `xxh128` in some places. In this PR I propose to use it in all places where we don't need a cryptographically secure hash.

Comments:

* ~I also remove some `base64` encoding of binary hashes, etc. It looks convoluted and to me it looks unnecessary. But I can revert these changes if needed.~ **REVERTED**
* There are some places where we can probably use `xxh128` but I'm not sure, so I didn't. These:
  * https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/HttpKernel/EventListener/CacheAttributeListener.php#L71
  * https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/HttpKernel/HttpCache/Store.php#L421-L434
  * https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/HttpKernel/HttpCache/Store.php#L421-L434
* And there's this use case, where a function explicitly tells that we're generating a sha256 hash, so I'm not sure if changing it is a BC break: https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Form/ChoiceList/Factory/CachingFactoryDecorator.php#L42-L65

#SymfonyHackday

### DIFF
--- a/src/Symfony/Bridge/Twig/NodeVisitor/TranslationDefaultDomainNodeVisitor.php
+++ b/src/Symfony/Bridge/Twig/NodeVisitor/TranslationDefaultDomainNodeVisitor.php
@@ -114,6 +114,6 @@ final class TranslationDefaultDomainNodeVisitor extends AbstractNodeVisitor
 
     private function getVarName(): string
     {
-        return sprintf('__internal_%s', hash('sha256', uniqid(mt_rand(), true), false));
+        return sprintf('__internal_%s', hash('xxh128', uniqid(mt_rand(), true)));
     }
 }
--- a/src/Symfony/Bundle/FrameworkBundle/EventListener/ConsoleProfilerListener.php
+++ b/src/Symfony/Bundle/FrameworkBundle/EventListener/ConsoleProfilerListener.php
@@ -72,7 +72,7 @@ final class ConsoleProfilerListener implements EventSubscriberInterface
             return;
         }
 
-        $request->attributes->set('_stopwatch_token', substr(hash('sha256', uniqid(mt_rand(), true)), 0, 6));
+        $request->attributes->set('_stopwatch_token', substr(hash('xxh128', uniqid(mt_rand(), true)), 0, 6));
         $this->stopwatch->openSection();
     }
 
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/FrameworkExtensionTestCase.php
@@ -1693,7 +1693,7 @@ abstract class FrameworkExtensionTestCase extends TestCase
                     ->replaceArgument(0, $expectedSeed)
                     ->replaceArgument(1, 12),
                 (new ChildDefinition('cache.adapter.redis'))
-                    ->replaceArgument(0, new Reference('.cache_connection.kYdiLgf'))
+                    ->replaceArgument(0, new Reference('.cache_connection.U5HliuY'))
                     ->replaceArgument(1, $expectedSeed)
                     ->replaceArgument(2, 12),
             ],
--- a/src/Symfony/Bundle/SecurityBundle/Tests/DependencyInjection/CompleteConfigurationTestCase.php
+++ b/src/Symfony/Bundle/SecurityBundle/Tests/DependencyInjection/CompleteConfigurationTestCase.php
@@ -137,7 +137,7 @@ abstract class CompleteConfigurationTestCase extends TestCase
             [
                 'simple',
                 'security.user_checker',
-                '.security.request_matcher.h5ibf38',
+                '.security.request_matcher.rud_2nr',
                 false,
                 false,
                 '',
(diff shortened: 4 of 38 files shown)


src/Symfony/Component/Lock/Tests/Store/FlockStoreTest.php
```diff
@@ -69,7 +69,7 @@ public function testSaveSanitizeName()
         $file = sprintf(
             '%s/sf.-php-echo-hello-word-.%s.lock',
             sys_get_temp_dir(),
-            strtr(substr(base64_encode(hash('sha256', $key, true)), 0, 7), '/', '_')
+            substr(hash('xxh128', $key), 0, 7)
```
jderusse: base64 was used to increase possible values and avoid collision (64^7 = 4,398046511×10¹² values instead of 16^7 = 268435456)

jderusse: note: this does not matter on test, but my comment is still true for src/Symfony/Component/Lock/Store/FlockStore.php

javiereguiluz: That's a good argument ... we need to decide if we want to keep base64 here for those reasons ... and if true, in which other parts we also want to keep base64 for the same reasons.

Thanks.

javiereguiluz: I've reverted all changes related to base64 encoding. Thanks!

src/Symfony/Component/Lock/Store/DatabaseTableTrait.php
```diff
@@ -50,7 +50,7 @@ private function init(array $options, float $gcProbability, int $initialTtl): vo
      */
     private function getHashedKey(Key $key): string
     {
-        return hash('sha256', (string) $key);
+        return hash('xxh128', (string) $key);
```
nicolas-grekas: I'd keep this one as is because the key might contain input data

src/Symfony/Component/Lock/Store/FlockStore.php
```diff
@@ -86,7 +86,7 @@ private function lock(Key $key, bool $read, bool $blocking): void
             $fileName = sprintf('%s/sf.%s.%s.lock',
                 $this->lockPath,
                 substr(preg_replace('/[^a-z0-9\._-]+/i', '-', $key), 0, 50),
-                strtr(substr(base64_encode(hash('sha256', $key, true)), 0, 7), '/', '_')
+                strtr(substr(base64_encode(hash('xxh128', $key, true)), 0, 7), '/', '_')
```
nicolas-grekas: same

src/Symfony/Component/Lock/Store/DatabaseTableTrait.php
javiereguiluz: OK. I reverted this and the other occurrence. Thanks.

fabpot (approved):

fabpot:
@javiereguiluz Can you have a look at the tests as they seem to be broken by the changes made here.

src/Symfony/Bridge/Twig/NodeVisitor/TranslationDefaultDomainNodeVisitor.php
```diff
@@ -114,6 +114,6 @@ private function isNamedArguments(Node $arguments): bool
 
     private function getVarName(): string
     {
-        return sprintf('__internal_%s', hash('sha256', uniqid(mt_rand(), true), false));
+        return sprintf('__internal_%s', hash('xxh128', uniqid(mt_rand(), true)));
```
GromNaN, line 117: I don't think we need a random var name here. A global sequence would be enough to ensure there is no conflict.
Similar to what I did in https://github.com/twigphp/Twig/pull/3601/files#diff-47bc8653b21716576958e25b6d7356ecb0f0070f17554d12d2bee985ac211b26R59

mahono:
Why not xxh3? (just asking for curiosity)

GromNaN (approved):

GromNaN:
> Why not xxh3? (just asking for curiosity)

In fact, where the hash is substr, `xxh3` is sufficient and [slightly faster](https://xxhash.com/).

javiereguiluz:
I think it's better to use `xxh128` instead of `xxh3` because:

* We already use `xxh128` in other parts of Symfony
* `xxh3` is faster, but only 3% faster
* We're moving from 256 bits (sha256) to 128 bits (xxh128) ... using xxh3 would mean going to 64 bits ... maybe it's too low compared to the original

fabpot:
Thank you @javiereguiluz.

PR #52949, opened 2023-12-08
[Messenger] Add `UnwrapHandlerExceptionMiddleware`

### LINKED ISSUE #52300: [Messenger] Make `HandleTrait` unwrap `HandlerFailedException`
### Description

The `HandleTrait` allows to use Messenger as a function call by enforcing a single handler being called and returning its result (cf. https://symfony.com/doc/current/messenger.html#getting-results-when-working-with-command-query-buses).

Still, if this handler throws an exception it will be wrapped in a `HandlerFailedException`. The latter is useful when multiple handlers are called, but it seems to me it just adds noise when using the `HandleTrait`.

Would it make sense to make `HandleTrait::handle` catch `HandlerFailedException`s and throw the wrapped exception instead?

### Example

_No response_

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #52300
| License       | MIT

In competition with #52952

When using [the `HandleTrait`](https://symfony.com/doc/current/messenger.html#getting-results-when-working-with-command-query-buses), the `HandlerFailedException` just adds noise as the wrapped exception is the one we would expect to be thrown.

This PR adds an `UnwrapHandlerExceptionMiddleware` (idea stolen from [Sulu’s `UnpackExceptionMiddleware`](https://github.com/sulu/messenger#unpackexceptionmiddleware)): added to a bus before the `HandleMessageMiddleware`, it will throw the single exception wrapped in a `HandlerFailedException`.

### DIFF
--- a/src/Symfony/Component/Messenger/Middleware/UnwrapHandlerExceptionMiddleware.php
+++ b/src/Symfony/Component/Messenger/Middleware/UnwrapHandlerExceptionMiddleware.php
@@ -1,0 +1,34 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Messenger\Middleware;
+
+use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\Exception\HandlerFailedException;
+use Symfony\Component\Messenger\Exception\LogicException;
+
+class UnwrapHandlerExceptionMiddleware implements MiddlewareInterface
+{
+    public function handle(Envelope $envelope, StackInterface $stack): Envelope
+    {
+        try {
+            return $stack->next()->handle($envelope, $stack);
+        } catch (HandlerFailedException $exception) {
+            $wrappedExceptions = $exception->getWrappedExceptions();
+
+            if (1 !== \count($wrappedExceptions)) {
+                throw new LogicException(sprintf('"%s" can only unwrap a single exception, but got %d.', __CLASS__, \count($wrappedExceptions)));
+            }
+
+            throw reset($wrappedExceptions);
+        }
+    }
+}
--- a/src/Symfony/Component/Messenger/Middleware/UnwrapHandlerExceptionMiddleware.php
+++ b/src/Symfony/Component/Messenger/Middleware/UnwrapHandlerExceptionMiddleware.php
@@ -25,7 +25,7 @@ class UnwrapHandlerExceptionMiddleware implements MiddlewareInterface
             $wrappedExceptions = $exception->getWrappedExceptions();
 
             if (1 !== \count($wrappedExceptions)) {
-                throw new LogicException(sprintf('"%s" can only unwrap a single exception, but got %d.', __CLASS__, \count($wrappedExceptions)));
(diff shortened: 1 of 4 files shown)


ro0NL:
> When using [the HandleTrait](https://symfony.com/doc/current/messenger.html#getting-results-when-working-with-command-query-buses), the HandlerFailedException just adds noise

why not unwrap from within HandleTrait then?

edit: oh BC -.-, perhaps introduce a new property flag instead :thinking:

MatTheCat:
> perhaps introduce a new property flag instead

Will give it a try thanks! The property would only serve as a BC layer so I’m not sure if https://symfony.com/doc/current/contributing/code/bc.html#adding-an-argument-to-a-public-method would apply?

ro0NL:
adding new private property to traits is allowed :)

it not sure why it would 'serve as a BC layer', it could be just a new feature


if we think it's the better default, then perhaps deprecating `HandleTrait` in favor of a new `QueryBusTrait` is the better way. It would solve ambigious naming, and we could mark the message bus property readonly right away.

I dont like the middleware approach, since it might force users to split their bus into multiple buses. IMHO it breaks the abstraction of `$theBus->dispatch($aMessage)`, and comes with double middleware bookkeeping.

MatTheCat:
> adding new private property to traits is allowed :)

Oh indeed I thought you were talking about a new argument for `HandleTrait::handle`.

> it not sure why it would 'serve as a BC layer', it could be just a new feature

AFAIU we first need a way to make the new behavior opt-in (to keep BC), but then remove the old one. The opt-in mechanism would then only serve as a BC layer.

> perhaps deprecating `HandleTrait` in favor of a new `QueryBusTrait` is the better way.

Don’t know about that, but `QueryBus` seems too specific: you could use the `HandleTrait` to implement a command bus e.g.
Something like `SingleHandlingTrait` maybe..?

src/Symfony/Component/Messenger/Middleware/UnwrapHandlerExceptionMiddleware.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    public function handle(Envelope $envelope, StackInterface $stack): Envelope
+    {
+        try {
+            return $stack->next()->handle($envelope, $stack);
+        } catch (HandlerFailedException $exception) {
+            $wrappedExceptions = $exception->getWrappedExceptions();
+
+            if (1 !== \count($wrappedExceptions)) {
+                throw new LogicException(sprintf('"%s" can only unwrap a single exception, but got %d.', __CLASS__, \count($wrappedExceptions)));
```
OskarStark:
```suggestion
                throw new LogicException(sprintf('%s can only unwrap a single exception, but got %d.', __CLASS__, \count($wrappedExceptions)));
```

I think it is more readable

MatTheCat: fabbot doesn't like it 😅

OskarStark: I think we can ignore it, @nicolas-grekas?

chalasr: indeed

MatTheCat:
@OskarStark @chalasr FYI I also opened https://github.com/symfony/symfony/pull/52952 which is an alternative proposed by @ro0NL I find better. Do you have a preference?

MatTheCat:
Closing, since there is no traction and a solution is trivial to implement in userland.

PR #30707, opened 2019-03-26
[Messenger][DX] Allow stamps to be passed directly to MessageBusInterface::dispatch()

weaverryan:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none
| License       | MIT
| Doc PR        | TODO

Me again o/!

This proposal is *purely* for DX. With `DelayStamp`, the proposal of QueueNameStamp and future things like `AmqpRoutingKeyStamp`, stamps are becoming more common for end users to use. This changes how it looks to use them:

```php
// before
$bus->dispatch(new Envelope(new SendSmsNotification('Hi!'), new DelayStamp(10), new QueueNameStamp('low')));

// after
$bus->dispatch(new SendSmsNotification('Hi!'), [new DelayStamp(10), new QueueNameStamp('low')]);
```

It's definitely a BC break, which is allowed because the component is experimental, though it should be minimized. This BC break shouldn't be felt by most end users, as creating your own bus is an advanced use-case. Even if you decorated it, you'll get an obvious error.

PR #29167, opened 2018-11-10
[Messenger] Add a trait for synchronous query & command buses

ogizanagi:
| Q             | A
| ------------- | ---
| Branch?       | 4.2 
| Bug fix?      | no
| New feature?  | yes 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed tickets | N/A   
| License       | MIT
| Doc PR        | symfony/symfony-docs/issues/10662

PR #52952, opened 2023-12-08
[Messenger] Deprecate `HandleTrait` in favor of a new `SingleHandlingTrait`

### LINKED ISSUE #52300: [Messenger] Make `HandleTrait` unwrap `HandlerFailedException`

MatTheCat:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Issues        | Fix #52300
| License       | MIT

In competition with #52949

When using [the `HandleTrait`](https://symfony.com/doc/current/messenger.html#getting-results-when-working-with-command-query-buses), the `HandlerFailedException` just adds noise as the wrapped exception is the one we would expect to be thrown.

This PR aims to replace the `HandleTrait` by a new `SingleHandlingTrait`, the latter throwing any single exception wrapped in a `HandlerFailedException`. In addition, it checks for a single handler even if one failed.

### DIFF
--- a/src/Symfony/Component/Messenger/SingleHandlingTrait.php
+++ b/src/Symfony/Component/Messenger/SingleHandlingTrait.php
@@ -1,0 +1,69 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Messenger;
+
+use Symfony\Component\Messenger\Exception\HandlerFailedException;
+use Symfony\Component\Messenger\Exception\LogicException;
+use Symfony\Component\Messenger\Stamp\HandledStamp;
+use Symfony\Component\Messenger\Stamp\StampInterface;
+
+trait SingleHandlingTrait
+{
+    private readonly MessageBusInterface $messageBus;
+
+    /**
+     * Dispatches the given message, expecting to be handled by a single handler
+     * and returns the result from the handler returned value.
+     * This behavior is useful for both synchronous command & query buses,
+     * the last one usually returning the handler result.
+     *
+     * @param object|Envelope $message The message or the message pre-wrapped in an envelope
+     * @param StampInterface[] $stamps
+     */
+    private function handle(object $message, array $stamps = []): mixed
+    {
+        if (!isset($this->messageBus)) {
+            throw new LogicException(sprintf('You must provide a "%s" instance in the "%s::$messageBus" property, but that property has not been initialized yet.', MessageBusInterface::class, static::class));
+        }
+
+        $exceptions = [];
+
+        try {
+            $envelope = $this->messageBus->dispatch($message, $stamps);
+        } catch (HandlerFailedException $exception) {
+            $envelope = $exception->getEnvelope();
+            $exceptions = $exception->getWrappedExceptions();
+        }
+
+        /** @var HandledStamp[] $handledStamps */
+        $handledStamps = $envelope->all(HandledStamp::class);
+
+        $handlers = array_merge(
+            array_map(static fn (HandledStamp $stamp) => $stamp->getHandlerName(), $handledStamps),
+            array_keys($exceptions),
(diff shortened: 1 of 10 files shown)


src/Symfony/Component/Messenger/SingleHandlingTrait.php
```diff
@@ -0,0 +1,67 @@
...
+
+namespace Symfony\Component\Messenger;
+
+use Symfony\Component\Messenger\Exception\HandlerFailedException;
+use Symfony\Component\Messenger\Exception\LogicException;
+use Symfony\Component\Messenger\Stamp\HandledStamp;
+
+trait SingleHandlingTrait
+{
+    private MessageBusInterface $messageBus;
```
ro0NL: \+ readonly?

src/Symfony/Component/Messenger/SingleHandlingTrait.php
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Dispatches the given message, expecting to be handled by a single handler
+     * and returns the result from the handler returned value.
+     * This behavior is useful for both synchronous command & query buses,
+     * the last one usually returning the handler result.
+     *
+     * @param object|Envelope $message The message or the message pre-wrapped in an envelope
```
ro0NL, line 29: perhaps allow the 2nd dispatch arg here as well https://github.com/symfony/symfony/blob/a3686fbfbbbfec317213f831eabeb5a450720545/src/Symfony/Component/Messenger/MessageBusInterface.php#L26

src/Symfony/Component/Messenger/SingleHandlingTrait.php
```diff
@@ -0,0 +1,67 @@
...
+use Symfony\Component\Messenger\Stamp\HandledStamp;
+
+trait SingleHandlingTrait
+{
+    private MessageBusInterface $messageBus;
+
+    /**
+     * Dispatches the given message, expecting to be handled by a single handler
+     * and returns the result from the handler returned value.
+     * This behavior is useful for both synchronous command & query buses,
```
ro0NL, line 26: fact is we're querying for a value, hence i prefer naming `QueryBusTrait::query`, but not a blocker :)

MatTheCat: I don’t want this trait’s name to imply a single use-case as there can be many (a command bus returning IDs e.g.).

EDIT: heh, just found https://github.com/symfony/symfony/pull/29167#issuecomment-439336457 :grin:

src/Symfony/Component/Messenger/SingleHandlingTrait.php
MatTheCat: Indeed #29167 got released before #30707 but I wonder if that’s the reason `HandleTrait::handle` does not offer a `$stamps` parameter :thinking: 

Will add it for now as it does not seem harmful.

src/Symfony/Component/Messenger/SingleHandlingTrait.php
MatTheCat: Well, why not!

chalasr:
The value brought by this change doesn't seem worth the cost of deprecating HandleTrait entirely to me. Also the name `SingleHandlingTrait` doesn't convey the fact it's about synchronous handling. 
All in all, I feel like this is not the right approach to solve the issue at hand - IMHO either we address it in `HandleTrait` with BC preserved or we need to look at the middleware alternative instead

MatTheCat:
Changing for a more explicit name and handling exceptions correctly felt enough for me, but will close then.

PR #52954, opened 2023-12-08
[Validator] Add `list` and `associative_array` types to `Type` constraint

hermann8u:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

Hello!

This PR aims to add support for `list` type in the `Type`constraint.

This is done by using the *new* PHP 8.1 function `array_is_list`, after checking that the value is an array with `is_array` function.

Here is an example of use:
```php
#[Assert\Type('list')]
private $value;
```

I consider doing an other PR to add `associative_array` type aswell. Or if you think it's ok, I can also add it in an other commit on this one.

--- UPDATE

After fabpot comment, I added an other commit to support `associative_array` type aswell :
```php
#[Assert\Type('associative_array')]
private $value;
```
An `associative_array` is an array that is not a list. Moreover, this means that an empty array will not be considered as an `associative_array`.

Best regards


OskarStark (approved):

src/Symfony/Component/Validator/Tests/Constraints/TypeValidatorTest.php
```diff
@@ -166,6 +168,8 @@ public static function getInvalidValues()
             [$file, 'float', 'resource'],
             [$file, 'string', 'resource'],
             [$file, 'object', 'resource'],
+            [['1' => 1], 'list', 'array'],
```
derrabus: This array expression is misleading because it will actually create an array with an integer `1` as key.

```suggestion
            [[1 => 1], 'list', 'array'],
```

In addition to this, I believe we should also add an obvious associative array as invalid value.

fabpot:
Let's add `associative_array` support in the same PR as this is closely related.

welcoMattic (approved):

OskarStark:
Please update the PR title + description to reflect the additional option, thanks

fabpot (approved):

fabpot:
Thank you @hermann8u.

ro0NL:
in php any array is associative ;)

ro0NL:
@hermann8u i fail to see what usecase would allow for [1 => some], but not [0 => some]

hermann8u:
@ro0NL I don't know, maybe when 1 is an id for example? I'm currently working on an other PR that add a constraint to validate array keys that could be used in association with this one.

hermann8u:
@ro0NL Ok I got your point finally. What you mean is that list are also associative_array?

PR #52959, opened 2023-12-09
[String] Skip a test when an issue is detected in PCRE2

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Part of #52206
| License       | MIT

I propose to ignore this test when [this issue of PCRE2](https://github.com/PCRE2Project/pcre2/issues/361) is detected until it's resolved and the polyfill updated.

### DIFF
--- a/src/Symfony/Component/String/Tests/AbstractUnicodeTestCase.php
+++ b/src/Symfony/Component/String/Tests/AbstractUnicodeTestCase.php
@@ -94,14 +94,21 @@ END'],
 
     public static function provideCodePointsAt(): array
     {
-        return [
+        $data = [
             [[], '', 0],
             [[], 'a', 1],
             [[0x53], 'Späßchen', 0],
             [[0xE4], 'Späßchen', 2],
             [[0xDF], 'Späßchen', -5],
-            [[0x260E], '☢☎❄', 1],
         ];
+
+        // Skip this set if we encounter an issue in PCRE2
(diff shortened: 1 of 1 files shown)


src/Symfony/Component/String/Tests/AbstractUnicodeTestCase.php
```diff
@@ -86,6 +86,10 @@ public function testCodePointsAt(array $expected, string $string, int $offset, i
             $this->markTestSkipped('Skipping due to issue ICU-21661.');
         }
 
+        if (3 !== grapheme_strlen('☢☎❄') && '☢☎❄' === $string) {
+            $this->markTestSkipped('Skipping due to issue in PCRE2 (https://github.com/PCRE2Project/pcre2/issues/361).');
```
xabbuh: What about doing this in the data provider instead (i.e. not yield the fixture when the length is incorrectly detected)?

alexandre-daubois: Why not, should this be done also for the check above this one?

alexandre-daubois: Done 🙂

OskarStark (approved):

fabpot (approved):

fabpot:
Thank you @alexandre-daubois.

PR #52961, opened 2023-12-09
[Security][SecurityBundle] Add `#[IsCsrfTokenValid]` attribute

### LINKED ISSUE #52947: What about an `IsCsrfTokenValid` attribute?
### Description

POST only actions like delete actions are likely to use such kind of code:
```php
if (!$this->isCsrfTokenValid('delete_example', $request->request->getString('_token'))) {
    throw new BadRequestHttpException('This token is invalid');
}
```
What about introducing a new `IsCsrfTokenValid` attribute that takes the id and optionally the token key to avoid repeating the `if` check.

If accepted I can try a PR during the Hackday 🙂

### Example

```php
#[IsCsrfTokenValid('delete_example')]
public function deleteAction(): Response
{
    /// ...
```

yguedidi:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #52947
| License       | MIT

#SymfonyHackday

### DIFF
--- a/src/Symfony/Component/Security/Http/Attribute/IsCsrfTokenValid.php
+++ b/src/Symfony/Component/Security/Http/Attribute/IsCsrfTokenValid.php
@@ -1,0 +1,29 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Security\Http\Attribute;
+
+#[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::TARGET_FUNCTION)]
+final class IsCsrfTokenValid
+{
+    public function __construct(
+        /**
+         * Sets the id used when generating the token.
+         */
+        public string $id,
+
+        /**
+         * Sets the key of the request that contains the actual token value that should be validated.
+         */
+        public ?string $tokenKey = '_token',
+    ) {
+    }
+}
--- a/src/Symfony/Component/Security/Http/CHANGELOG.md
+++ b/src/Symfony/Component/Security/Http/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Add `#[IsCsrfTokenValid]` attribute
+
 7.0
 ---
 
--- a/src/Symfony/Component/Security/Http/EventListener/IsCsrfTokenValidAttributeListener.php
+++ b/src/Symfony/Component/Security/Http/EventListener/IsCsrfTokenValidAttributeListener.php
@@ -1,0 +1,52 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Security\Http\EventListener;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\ControllerArgumentsEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\Security\Core\Exception\InvalidCsrfTokenException;
+use Symfony\Component\Security\Csrf\CsrfToken;
+use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
+use Symfony\Component\Security\Http\Attribute\IsCsrfTokenValid;
+
+/**
+ * Handles the IsCsrfTokenValid attribute on controllers.
+ */
+final class IsCsrfTokenValidAttributeListener implements EventSubscriberInterface
+{
+    public function __construct(
+        private readonly CsrfTokenManagerInterface $csrfTokenManager,
+    ) {
+    }
+
+    public function onKernelControllerArguments(ControllerArgumentsEvent $event): void
+    {
+        /** @var IsCsrfTokenValid[] $attributes */
+        if (!\is_array($attributes = $event->getAttributes()[IsCsrfTokenValid::class] ?? null)) {
+            return;
+        }
+
+        $request = $event->getRequest();
+
+        foreach ($attributes as $attribute) {
+            if (!$this->csrfTokenManager->isTokenValid(new CsrfToken($attribute->id, $request->request->getString($attribute->tokenKey)))) {
+                throw new InvalidCsrfTokenException('Invalid CSRF token.');
+            }
+        }
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [KernelEvents::CONTROLLER_ARGUMENTS => ['onKernelControllerArguments', 25]];
+    }
+}
--- a/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Compiler/RegisterCsrfFeaturesPass.php
+++ b/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Compiler/RegisterCsrfFeaturesPass.php
@@ -17,6 +17,7 @@ use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\Security\Csrf\TokenStorage\ClearableTokenStorageInterface;
 use Symfony\Component\Security\Http\EventListener\CsrfProtectionListener;
 use Symfony\Component\Security\Http\EventListener\CsrfTokenClearingLogoutListener;
+use Symfony\Component\Security\Http\EventListener\IsCsrfTokenValidAttributeListener;
 
 /**
  * @author Christian Flothmann <christian.flothmann@sensiolabs.de>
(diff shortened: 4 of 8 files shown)


src/Symfony/Component/Security/Http/EventListener/IsCsrfTokenValidAttributeListener.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        /** @var IsCsrfTokenValid[] $attributes */
+        if (!\is_array($attributes = $event->getAttributes()[IsCsrfTokenValid::class] ?? null)) {
+            return;
+        }
+
+        $request = $event->getRequest();
+
+        foreach ($attributes as $attribute) {
+            if (!$this->csrfTokenManager->isTokenValid(new CsrfToken($attribute->id, $request->request->getString($attribute->tokenKey)))) {
```
yguedidi, line 42: should support for query parameters be added?

src/Symfony/Component/Security/Http/EventListener/IsCsrfTokenValidAttributeListener.php
```diff
@@ -0,0 +1,52 @@
...
+                throw new InvalidCsrfTokenException('Invalid CSRF token.');
+            }
+        }
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [KernelEvents::CONTROLLER_ARGUMENTS => ['onKernelControllerArguments', 20]];
```
yguedidi: should the priority be 21 to be executed before `IsGranted`?

src/Symfony/Component/Security/Http/EventListener/IsCsrfTokenValidAttributeListener.php
wouterj: Hmm, is there a use-case for CSRF protection on GET requests? I think we should advocate using POST/PATCH/PUT whenever your action needs CSRF protection.

src/Symfony/Component/Security/Http/EventListener/IsCsrfTokenValidAttributeListener.php
wouterj: I'm not sure. Given the current situation always runs CSRF protection after authorization, I think it's OK to keep it after, unless there is a good (security) reason to move it before.

src/Symfony/Component/Security/Http/EventListener/IsCsrfTokenValidAttributeListener.php
yguedidi: I agree!

src/Symfony/Component/Security/Http/EventListener/IsCsrfTokenValidAttributeListener.php
yguedidi: My questioning here was more about performance optimization than security.
As an invalid CSRF will make the request to fail anyway, maybe not executing the potentially heavy computations involved by an `IsGranted` (e.g. database calls) could be a win?

yguedidi: Go for 25 as seen face to face :)

chalasr (approved):

src/Symfony/Component/Security/Http/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+* Add `#[IsCsrfTokenValid]` attribute
```
OskarStark:
```suggestion
 * Add `#[IsCsrfTokenValid]` attribute
```

src/Symfony/Component/Security/Http/EventListener/IsCsrfTokenValidAttributeListener.php
```diff
@@ -0,0 +1,52 @@
...
+use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\Security\Core\Exception\InvalidCsrfTokenException;
+use Symfony\Component\Security\Csrf\CsrfToken;
+use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
+use Symfony\Component\Security\Http\Attribute\IsCsrfTokenValid;
+
+/**
+ * Handles the IsCsrfTokenValid attribute on controllers.
+ */
+class IsCsrfTokenValidAttributeListener implements EventSubscriberInterface
```
OskarStark: Do we want to make it final?

src/Symfony/Component/Security/Http/Attribute/IsCsrfTokenValid.php
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Security\Http\Attribute;
+
+#[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::TARGET_FUNCTION)]
```
OskarStark, line 14: I am not sure what's the difference between METHOD and FUNCTION, but it looks like we only have tests for methods, do we? 🤔

yguedidi: @OskarStark I inspired from the IsGranted attribute that apparently supports FUNCTION in addition to METHOD. And FUNCTION wasn't tested for IsGranted case.
Should I add test for it? or better remove the support of FUNCTION for now?

OskarStark: If we have a use case add a test case, we should ofc provide a test case for the `IsGranted` in an extra PR too.

src/Symfony/Component/Security/Http/EventListener/IsCsrfTokenValidAttributeListener.php
chalasr: I think so 👍 The container wouldn't be able to properly handle any potential extra capacity added to a child attribute, hence extending autoconfigured attributes isn't supported basically.

dsdeboer (approved):
Nicely done 👏

src/Symfony/Component/Security/Http/Attribute/IsCsrfTokenValid.php
chalasr: I do agree there should be a test ideally, at least to avoid regressions especially. But, I'd be fine moving forward without one: here the use case for `TARGET_FUNCTION`  is that controllers can be any callable. Since controllers as callables are test-covered and there is no target-specific behavior with this attribute. We have plently of attributes in the same case including validator constraints.

OskarStark: I am fine then 👍

wouterj: I agree, imho we don't necessarily need to test this. This is by-design for all controller argument listeners and the controller argument event has tests to check if we support callables.

chalasr:
I know the name of the proposed attribute is inline with the `#[IsGranted]` one and I'm fundamentally fine with it, yet I'm wondering if some more imperative name wouldn't read better  e.g. `#[ValidateCsrfToken]`, `#[ValidCsrfToken]`  or even `#[RequireValidCsrfToken`. Thoughts?

wouterj (approved):

OskarStark:
> I know the name of the proposed attribute is inline with the #[IsGranted] one and I'm fundamentally fine with it, yet I'm wondering if some more imperative name wouldn't read better e.g. #[ValidateCsrfToken], #[ValidCsrfToken] or even #[RequireValidCsrfToken. Thoughts?

`#[HasValidCsrfToken]` ?

src/Symfony/Component/Security/Http/Attribute/IsCsrfTokenValid.php
yguedidi: then this is ready for final review 🙂

yguedidi:
@chalasr it's more inline with the `AbstractController` helper function `->isCsrfTokenValid()`, like the `#[IsGranted]` is with `->isGranted()`

wouterj:
> @chalasr it's more inline with the AbstractController helper function ->isCsrfTokenValid(), like the #[IsGranted] is with ->isGranted()

Sounds valid to me :+1: (I honestly didn't know we had this helper method in the abstract controller)

fabpot:
@yguedidi Can you have a look at the tests, they seem broken by this PR.

OskarStark:
Let's go with the name then 👍🏻

Thanks for the explanation ℹ️

yguedidi:
> Can you have a look at the tests, they seem broken by this PR.

@fabpot Indeed low deps unit tests are broken by this PR!
I checked and I'm wondering what is the right fix.
Issue is that SecurityBundle will install security/http 6.4, so without the new listener.
Should I just upgrade the requirement to 7.1 or conditionally add the service if the class exists? or something else?

fabpot:
> > Can you have a look at the tests, they seem broken by this PR.
> 
> @fabpot Indeed low deps unit tests are broken by this PR! I checked and I'm wondering what is the right fix. Issue is that SecurityBundle will install security/http 6.4, so without the new listener. Should I just upgrade the requirement to 7.1 or conditionally add the service if the class exists? or something else?

You need to bump the minimum version in SecurityBundle.

yguedidi:
@fabpot SecurityBundle tests now pass! other failures seems unrelated

fabpot (approved):

fabpot:
Thank you @yguedidi.

PR #52967, opened 2023-12-09
[Form] fix DateTypeTest.php

toxicity1985:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? |no 
| License       | MIT

image

Hello,

When I launch my test on my laptop, i was having some issue with the DateTypeTest.
Here is my fixes.

Best regards.

### DIFF
--- a/src/Symfony/Component/Form/Tests/Extension/Core/Type/DateTypeTest.php
+++ b/src/Symfony/Component/Form/Tests/Extension/Core/Type/DateTypeTest.php
@@ -106,7 +106,7 @@ class DateTypeTest extends BaseTypeTestCase
             'input' => 'datetime',
         ]);
 
-        $form->submit('2.6.2010');
+        $form->submit('02.06.2010');
 
         $this->assertEquals(new \DateTime('2010-06-02 UTC'), $form->getData());
         $this->assertEquals('02.06.2010', $form->getViewData());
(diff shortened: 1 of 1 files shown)


yceruto:
Duplicated of https://github.com/symfony/symfony/pull/52792? Shouldn't it have already been fixed by https://github.com/symfony/symfony/pull/52245? /cc @xabbuh

xabbuh:
Can you please check which ICU version you have?

toxicity1985:
I have this version:

uconv v2.1  ICU 72.1

xabbuh:
Thanks for making us aware of this @toxicity1985. I suggest we skip these five tests with ICU 72.1 too (see #53335).

smnandre:
It seems there is a bug until the most recent ICU version (74), so it may concern all ICU versions < 74.2
(almost the same we had with emojis)

ICU 74.2 release notes:
 > Important: DateFormat.getInstanceForSkeleton() and the DateTimePatternGenerator sometimes used the wrong patterns because they failed to use/inherit certain data. ([ICU-22575](https://unicode-org.atlassian.net/browse/ICU-22575) — CLDR 44 had removed some redundant data that ICU relied on.) 
--> ICU 74 : https://icu.unicode.org/download/74

Bug issue:  https://bugzilla.mozilla.org/show_bug.cgi?id=1864612#c8

As it may concern a lot of users, should we (can we) create a temporary patch ? 

cc @xabbuh

xabbuh:
For me, tests pass locally with ICU 73.1. So not sure if that's the same issue.

smnandre:
@xabbuh ok! I’ll investigate tomorrow and see if there is something here that impact us

PR #52970, opened 2023-12-09
[HttpClient] Add `JsonMockResponse::fromFile()` and `MockResponse::fromFile()` shortcuts

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

I have noticed I use a lot of file fixtures for tests since they are more readable as stand-alone files.

I think adding small shortcut methods would help the DX.


OskarStark:
I am not really sold on the new `$json` parameter 🤔 Could't we just check if the data `is_string()` and remove the additional parameter?

fancyweb:
We can't do the `is_string` trick because we already allow passing a string.

src/Symfony/Component/HttpClient/Response/JsonMockResponse.php
```diff
@@ -16,18 +16,35 @@
 class JsonMockResponse extends MockResponse
 {
     /**
-     * @param mixed $body Any value that `json_encode()` can serialize
+     * @param mixed $data Any value that `json_encode()` can serialize
+     * @param bool  $json Whether the data is already a JSON string
      */
-    public function __construct(mixed $body = [], array $info = [])
+    public function __construct(mixed $data = [], array $info = [], bool $json = false)
```
OskarStark:
```suggestion
    public function __construct(mixed $data = [], array $info = [], bool $isJson = false)
```
to be more speakable

src/Symfony/Component/HttpClient/Tests/Response/MockResponseTest.php
```diff
@@ -15,11 +15,9 @@
 use Symfony\Component\HttpClient\Exception\InvalidArgumentException;
 use Symfony\Component\HttpClient\Exception\JsonException;
 use Symfony\Component\HttpClient\Exception\TransportException;
+use Symfony\Component\HttpClient\MockHttpClient;
 use Symfony\Component\HttpClient\Response\MockResponse;
 
-/**
- * Test methods from Symfony\Component\HttpClient\Response\*ResponseTrait.
```
fancyweb, line 21: That's already not the case btw.

src/Symfony/Component/HttpClient/Response/JsonMockResponse.php
fancyweb: Looking at this again, I think it's acceptable to json_decode and then json_encode for simplicity instead of having to add a new arg etc. The perf gain is minimal I guess?

welcoMattic (approved):

src/Symfony/Component/HttpClient/Response/JsonMockResponse.php
OskarStark: Thanks

OskarStark (approved):

fabpot (approved):

fabpot:
Thank you @fancyweb.

PR #52971, opened 2023-12-09
[Messenger] Make `#[AsMessageHandler]` final

### LINKED ISSUE #52898: [Messenger] Support extending `AsMessageHandler` attribute
### Description

The `AsMessageHandler` attribute has some options which would be pretty common / supplied by numerous locations using the attribute. For example the `$bus` argument, if you define multiple buses you would most likely want to define this argument for all (or all except one, the default, bus type). Being able to extend the `AsMessageHandler` would make this a lot simpler.

Currently one can actually extend the class. But this doesn't work as the `AttributeAutoconfigurationPass` doesn't handle inheritance. Something which it IMO also shouldn't do by default. But in some cases there are attribute where it really makes sense to allow inheritance.

In case this wouldn't be accepted, it is my opinion that the attribute class, i.e. `AsMessageHandler` should be `final`, as extending it actually doesn't work / extending it only has the meaning of inheriting the properties, but not the behavior (autoconfiguration) applied to the attribute.

### Example

```php

#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
class AsCommandHandler extends AsMessageHandler
{
    public function __construct(
        ?string $fromTransport = null,
        ?string $handles = null,
        ?string $method = null,
        int $priority = 0,
    ) {
        parent::__construct('command.bus', $fromTransport, $handles, $method, $priority);
    }
}
```
in combination with
```php
#[AsCommandHandler]
class FooHandler
{
  ...
}

#[AsCommandHandler]
class BarHandler
{
  ...
}
```

Instead of the current:
```php
#[AsMessageHandler('command.bus')]
class FooHandler
{
  ...
}

#[AsMessageHandler('command.bus')]
class BarHandler
{
  ...
}
```

(obviously having multiple `As...Handler` attributes which use different `$bus` arguments)

Valmonzo:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Issues        | Fix #52898 
| License       | MIT

#SymfonyHackDay

### DIFF
--- a/src/Symfony/Component/Messenger/CHANGELOG.md
+++ b/src/Symfony/Component/Messenger/CHANGELOG.md
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Add option `redis_sentinel` as an alias for `sentinel_master`
  * Add `--all` option to the `messenger:consume` command
+ * Make `#[AsMessageHandler]` final
 
 7.0
 ---
--- a/UPGRADE-7.1.md
+++ b/UPGRADE-7.1.md
@@ -1,6 +1,11 @@
 UPGRADE FROM 7.0 to 7.1
 =======================
 
+Messenger
+---------
+
+ * Make `#[AsMessageHandler]` final
+
 Workflow
 --------
 
--- a/src/Symfony/Component/Messenger/Attribute/AsMessageHandler.php
+++ b/src/Symfony/Component/Messenger/Attribute/AsMessageHandler.php
@@ -14,6 +14,8 @@ namespace Symfony\Component\Messenger\Attribute;
 /**
  * Service tag to autoconfigure message handlers.
  *
+ * @final since Symfony 7.1
+ *
  * @author Alireza Mirsepassi <alirezamirsepassi@gmail.com>
  */
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]


src/Symfony/Component/Messenger/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Add option `redis_sentinel` as an alias for `sentinel_master`
  * Add `--all` option to the `messenger:consume` command
+ * Make `AsMessageHandler` final
```
chalasr:
```suggestion
 * Make `#[AsMessageHandler]` final
```

chalasr (approved):

OskarStark (approved):

derrabus (approved):
Let's also update UPGRADE-7.1.md?

RobertMe:
> Issues: 	Fix #52898

I don't think that's really the case. This "fix" changes something which didn't work into something which still doesn't work, but just for a different reason. So while it does relate to my feature request, it doesn't fix it. And I'm not even sure that the acceptation of this change would also answer my feature request. It does close of one possibility to support / implement the feature request. But it doesn't void the feature request to be able to "shortcut" `#[AsMessageHandler]` so not to have to repeat (for example) the `bus` for every usage of the attribute. So personally I would prefer a separate discussion on whether such a feature would be possible, with or without `AsMessageHandler` being `final`. While such discussion would be ended "automatically" when the fixes / closes "tag" is applied.

Furthermore, for the reasons given by @chalasr in #52898 it is my believe that _all_ attribute classes should be made `final`. While when only checking the `DependencyInjection\Attribute` namespace there are more attribute classes _not_ `final` than ones which actually _are_ final. And as a matter of fact, there are also some attributes which are "open" on purpose as other attributes extend from them (for example `Autowire` which is extended by `AutowireCallable`, `AutowireIterator`, `AutowireLocator` and `AutowireServiceClosure`). So while I understand the reasoning to make attribute classes final (as extending them and adding extra properties / functionality doesn't magically work), it would be my believe that _all_ attribute classes should be final.

So to conclude it is my believe that Symfony should have some guideline / "code style" on whether (all) attribute classes must always be final, or whether they may or should be open. And in that discussion the point made by @chalasr obviously should be taken into consideration (but also that fact that said argument is already "violated" by for example all the mentioned `Autowire*` tags where it previously was considered fine to extend the attribute class).

nicolas-grekas:
Thank you @Valmonzo.

PR #52972, opened 2023-12-09
[HttpClient] Add `NoRetryStrategy` to help testing errors going through `RetryableHttpClient`

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

A very small test-related DX feature again.

When I test how my code reacts to a 500 error (`new MockResponse('', ['http_code' => 500])`) and it goes through `RetryableHttpClient`, I don't want to retry anything at all (or I'd have to provide several responses to the `MockHttpClient`), so I always create a `NoRetryStrategy`. I'd like this code to be part of Symfony directly so it can be reused by anyone more easily :smile:

### DIFF
--- a/src/Symfony/Component/HttpClient/Test/Retry/NoRetryStrategy.php
+++ b/src/Symfony/Component/HttpClient/Test/Retry/NoRetryStrategy.php
@@ -1,0 +1,29 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\HttpClient\Test\Retry;
+
+use Symfony\Component\HttpClient\Response\AsyncContext;
+use Symfony\Component\HttpClient\Retry\RetryStrategyInterface;
+use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
+
+final class NoRetryStrategy implements RetryStrategyInterface
+{
+    public function shouldRetry(AsyncContext $context, ?string $responseContent, ?TransportExceptionInterface $exception): ?bool
+    {
+        return false;
+    }
+
+    public function getDelay(AsyncContext $context, ?string $responseContent, ?TransportExceptionInterface $exception): int
+    {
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/HttpClient/Test/Retry/NoRetryStrategy.php
```diff
@@ -0,0 +1,29 @@
...
+final class NoRetryStrategy implements RetryStrategyInterface
+{
+    public function shouldRetry(AsyncContext $context, ?string $responseContent, ?TransportExceptionInterface $exception): ?bool
+    {
+        return false;
+    }
+
+    public function getDelay(AsyncContext $context, ?string $responseContent, ?TransportExceptionInterface $exception): int
+    {
+        throw new \LogicException();
```
stloyd, line 27: Maybe it would be worth it to pass also context and exception to this one?

fancyweb: This will never be called since `shouldRetry()` returns false.

stloyd: Then why not return 0 and add a comment it will not happen?

danielburger1337:
Since Symfony 6.4 you can set the "max_retries" option. The value "0" will disable retrying.

```php
$httpClient->request('GET', '/some-path', [
    'max_retries' => 0
]);
```

Wouldn't this solve your problem as well?

Nyholm:
Thank you. 

Could you update the PR description with a config example how you would use this strategy in testing only? (That would help when writing docs)

@danielburger1337
Since you only want to use this strategy when testing, you cannot really use `'max_retries' => 0`. Imagine the `$httpClient->request('GET', '/some-path')` is inside your API client.

fancyweb:
I'm closing this one because I can leverage the maxRetries arg, I hadn't thought about using it with a 0 value :sweat_smile:

danielburger1337:
@Nyholm I could've been a bit more clearer with my description: 

You can also pass "0" to the constructor `RetryableHttpClient::$maxRetries`. 
Then you don't have to pass `"max_retries" => 0` to every request.

PR #52973, opened 2023-12-09
[EventDispatcher] Enable the possibility of deprecating events

Jeroeny:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

Currently there is no way to deprecate events in the framework used by the EventDispatcher. It's probably not even desired to enable this possibility, but I wanted to have a go at it anyway. 

The changes in `RegisterListenersPass` might be reverted if throwing the deprecation in `addListener`. Though if users are DI-configuring there own `EventDispatcher`, they might not have set the `$deprecated` argument and won't get the deprecation notices. 

#SymfonyHackday

### DIFF
--- a/src/Symfony/Component/EventDispatcher/Tests/DependencyInjection/RegisterListenersPassTest.php
+++ b/src/Symfony/Component/EventDispatcher/Tests/DependencyInjection/RegisterListenersPassTest.php
@@ -12,6 +12,7 @@
 namespace Symfony\Component\EventDispatcher\Tests\DependencyInjection;
 
 use PHPUnit\Framework\TestCase;
+use Symfony\Bridge\PhpUnit\ExpectDeprecationTrait;
 use Symfony\Component\DependencyInjection\Argument\ServiceClosureArgument;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\Compiler\AttributeAutoconfigurationPass;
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/EventDispatcher/Tests/DependencyInjection/RegisterListenersPassTest.php
```diff
@@ -503,6 +507,27 @@ public function testOmitEventNameOnSubscriber()
         ];
         $this->assertEquals($expectedCalls, $definition->getMethodCalls());
     }
+
+    /**
+     * @group legacy
+     */
+    public function testEventDeprecations()
+    {
+        $this->expectDeprecation('Since symfony/event-dispatcher 7.1: Event is deprecated');
```
OskarStark: Spult be moved before the compile() method, as this triggers the deprecation, does it?

Jeroeny: Indeed, done

nicolas-grekas:
Closing for inactivity. Nobody from the core team picked this up in over two years, and the mechanism it extends has moved on: RegisterListenersPass::setHotPathEvents() and setNoPreloadEvents() were deprecated in 8.1 in favour of AddEventAliasesPass, so addDeprecatedEvent() would land on the wrong side of that split, and widening the EventDispatcher constructor needs its own discussion. Please reopen if you want to revive the idea against the current branch.

PR #52974, opened 2023-12-09
[Cache] Deprecate `CouchbaseBucketAdapter`, use `CouchbaseCollectionAdapter`

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Issues        | -
| License       | MIT

`CouchebaseBucketAdapter` requires Couches 2.x, while the `CouchbaseCollectionAdapter` requires 3.x. This conflicts in the CI where both versions can't be installed. Moreover, the last release of Couchbase extension 2.x was 3 years ago (https://pecl.php.net/package/couchbase). I think it's safe to deprecate the former one and only keep `CouchbaseCollectionAdapter`. Couchbase 4.x support can be added eventually in another PR if someone wants to have a look 🙂

### DIFF
--- a/src/Symfony/Component/Cache/Tests/Adapter/CouchbaseBucketAdapterTest.php
+++ b/src/Symfony/Component/Cache/Tests/Adapter/CouchbaseBucketAdapterTest.php
@@ -12,6 +12,7 @@
 namespace Symfony\Component\Cache\Tests\Adapter;
 
 use Psr\Cache\CacheItemPoolInterface;
+use Symfony\Bridge\PhpUnit\ExpectDeprecationTrait;
 use Symfony\Component\Cache\Adapter\AbstractAdapter;
 use Symfony\Component\Cache\Adapter\CouchbaseBucketAdapter;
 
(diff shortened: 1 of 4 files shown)


src/Symfony/Component/Cache/Tests/Adapter/CouchbaseBucketAdapterTest.php
```diff
@@ -50,8 +50,8 @@ public function createCachePool($defaultLifetime = 0): CacheItemPoolInterface
                 .':'.getenv('COUCHBASE_PASS')
                 .'@'.getenv('COUCHBASE_HOST')
                 .'/cache')
-            : self::$client;
+            : $this->client;
 
-        return new CouchbaseBucketAdapter($client, str_replace('\\', '.', __CLASS__), $defaultLifetime);
+        return new \Symfony\Component\Cache\Adapter\CouchbaseBucketAdapter($client, str_replace('\\', '.', __CLASS__), $defaultLifetime);
```
alexandre-daubois: Changing from an import to FQCN allows to add `trigger_deprecation()` and don't have the self-deprecation. Indeed, tests from this class can be skipped which leads to not checking the raised deprecation. Now, the deprecation is checked whenever the tests are run.

OskarStark: The tests of the deprecated adapter should get `@group legacy`

alexandre-daubois: Oh of course, totally forgot about it 🤦 I removed the FQCN and revert to the import. Thank you 🙏

nicolas-grekas (approved):

OskarStark (approved):

fabpot (approved):

fabpot:
Thank you @alexandre-daubois.

PR #52977, opened 2023-12-09
[AssetMapper] require scss files

### LINKED ISSUE #52957: [AssetMapper] Support requiring Sass .scss files in the same way as CSS
### Description

Pretty simple request here, as the tite says: AssetMapper could support requiring Sass .scss files in the same way as CSS.  That would allow me to do something like:

```
php bin/console importmap:require bootswatch/dist/zephyr/_variables.scss
```

This currently fails as there is no ESM js package for this.  However the CSS from the same package works:

```
php bin/console importmap:require bootswatch/dist/zephyr/bootstrap.css
```

### Example

_As above_

tip2tail:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #52957 
| License       | MIT

Modification of the AssetMapper `importmap:require` command to support importing Sass `.scss` files in the same way as CSS.

#SymfonyHackday

### DIFF
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
@@ -200,12 +200,11 @@ class ImportMapManager
 
     private static function getImportMapTypeFromFilename(string $path): ImportMapType
     {
-        if (str_ends_with($path, '.css')) {
-            return ImportMapType::CSS;
-        } else if (str_ends_with($path, '.scss')) {
-            return ImportMapType::SCSS;
-        }
-        return ImportMapType::JS;
+        return match (true) {
+            str_ends_with($path, '.css') => ImportMapType::CSS,
+            str_ends_with($path, '.scss') => ImportMapType::SCSS,
+            default => ImportMapType::JS,
+        };
     }
 
     /**
--- a/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
@@ -200,7 +200,12 @@ class ImportMapManager
 
     private static function getImportMapTypeFromFilename(string $path): ImportMapType
     {
-        return str_ends_with($path, '.css') ? ImportMapType::CSS : ImportMapType::JS;
+        if (str_ends_with($path, '.css')) {
+            return ImportMapType::CSS;
+        } else if (str_ends_with($path, '.scss')) {
+            return ImportMapType::SCSS;
+        }
+        return ImportMapType::JS;
     }
 
     /**
--- a/src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
@@ -78,7 +78,10 @@ final class JsDelivrEsmResolver implements PackageResolverInterface
                 throw new RuntimeException(sprintf('Unable to find the latest version for package "%s" - try specifying the version manually.', $packageName));
             }
 
-            $pattern = str_ends_with($filePath, '.css') || str_ends_with($filePath, '.scss') ? self::URL_PATTERN_DIST_STYLE : self::URL_PATTERN_DIST;
+            $pattern = match (true) {
+                str_ends_with($filePath, '.css'), str_ends_with($filePath, '.scss') => self::URL_PATTERN_DIST_STYLE,
+                default => self::URL_PATTERN_DIST,
+            };
             $requiredPackages[$i][1] = $this->httpClient->request('GET', sprintf($pattern, $packageName, $version, $filePath));
             $requiredPackages[$i][4] = $version;
 
--- a/src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
+++ b/src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
@@ -175,11 +175,10 @@ final class JsDelivrEsmResolver implements PackageResolverInterface
                 throw new \InvalidArgumentException(sprintf('The entry "%s" is not a remote package.', $entry->importName));
             }
 
-            if (ImportMapType::CSS === $entry->type || ImportMapType::SCSS === $entry->type) {
-                $pattern = self::URL_PATTERN_DIST_STYLE;
-            } else {
-                $pattern = self::URL_PATTERN_DIST;
-            }
+            $pattern = match ($entry->type) {
+                ImportMapType::CSS, ImportMapType::SCSS => self::URL_PATTERN_DIST_STYLE,
+                default => self::URL_PATTERN_DIST,
+            };
             $url = sprintf($pattern, $entry->getPackageName(), $entry->version, $entry->getPackagePathString());
 
             $responses[$package] = [$this->httpClient->request('GET', $url), $entry];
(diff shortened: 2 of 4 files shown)


weaverryan:
Hi there!

I was just bumping up against this a little bit with the symfony-demo. I have some questions/comments

A) Being able to download the scss file is super nice... so that you can have it locally on your machine and can then import it from your scss file. But we do not want it inside the actual `importmap` that appears on the screen. This is... kind of the first case where we're using `importmap:require` as a "vendor downloader" independent of "the thing that makes the importmap". I'm not sure how I feel about that... but I see WHY this would be desired.

B) Related to above, IF we decided to go this direction... in theory, you might want this as a way to download image or font files from 3rd party packages (not: fonts & images are supported indirectly today: if you `importmap:require some/file.css` and it references via `url()` fonts or images in the package, we download all of that).

C) Finally, what if I download a `.scss` file an it imports other local `.scss` files from the repo? For example https://cdn.jsdelivr.net/npm/bootstrap@5.3.2/scss/bootstrap.scss

And extra thought:

D) It makes me wonder: do we want a "mode" where you can say that you want to download an entire package vs just a single file? I'm worried about the scope of the `importmap:require` feature - and not making it too large. Also, if we did allow an entire package to be downloaded, every file in that package would be publicly accessible. So if you download all of bootstrap, your site is technically serving up every file in that package publicly. That's not a security problem... but feels a bit odd.

src/Symfony/Component/AssetMapper/ImportMap/ImportMapManager.php
```diff
@@ -200,7 +200,12 @@ private function cleanupPackageFiles(ImportMapEntry $entry): void
 
     private static function getImportMapTypeFromFilename(string $path): ImportMapType
     {
-        return str_ends_with($path, '.css') ? ImportMapType::CSS : ImportMapType::JS;
+        if (str_ends_with($path, '.css')) {
+            return ImportMapType::CSS;
+        } else if (str_ends_with($path, '.scss')) {
+            return ImportMapType::SCSS;
+        }
+        return ImportMapType::JS;
```
dsdeboer:
```suggestion
        return match (true) {
            str_ends_with($path, '.css') => ImportMapType::CSS,
            str_ends_with($path, '.scss') => ImportMapType::SCSS,
            default => ImportMapType::JS,
        };
```

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -78,7 +78,7 @@ public function resolvePackages(array $packagesToRequire): array
                 throw new RuntimeException(sprintf('Unable to find the latest version for package "%s" - try specifying the version manually.', $packageName));
             }
 
-            $pattern = str_ends_with($filePath, '.css') ? self::URL_PATTERN_DIST_CSS : self::URL_PATTERN_DIST;
+            $pattern = str_ends_with($filePath, '.css') || str_ends_with($filePath, '.scss') ? self::URL_PATTERN_DIST_STYLE : self::URL_PATTERN_DIST;
```
dsdeboer:
```suggestion
            $pattern = match (true) {
                str_ends_with($filePath, '.css'), str_ends_with($filePath, '.scss') => self::URL_PATTERN_DIST_STYLE,
                default => self::URL_PATTERN_DIST,
            };
```

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -106,7 +106,13 @@ public function resolvePackages(array $packagesToRequire): array
...
 
             $contentType = $response->getHeaders()['content-type'][0] ?? '';
-            $type = str_starts_with($contentType, 'text/css') ? ImportMapType::CSS : ImportMapType::JS;
+            if (str_starts_with($contentType, 'text/css')) {
+                $type = ImportMapType::CSS;
+            } else if (str_starts_with($contentType, 'text/x-scss')) {
+                $type = ImportMapType::SCSS;
+            } else {
+                $type = ImportMapType::JS;
+            }
```
dsdeboer:
```suggestion
            $type = match (true) {
                str_starts_with($contentType, 'text/css') => ImportMapType::CSS,
                str_starts_with($contentType, 'text/x-scss') => ImportMapType::SCSS,
                default => ImportMapType::JS,
            };
```

src/Symfony/Component/AssetMapper/ImportMap/Resolver/JsDelivrEsmResolver.php
```diff
@@ -169,7 +175,11 @@ public function downloadPackages(array $importMapEntries, callable $progressCall
                 throw new \InvalidArgumentException(sprintf('The entry "%s" is not a remote package.', $entry->importName));
             }
 
-            $pattern = ImportMapType::CSS === $entry->type ? self::URL_PATTERN_DIST_CSS : self::URL_PATTERN_DIST;
+            if (ImportMapType::CSS === $entry->type || ImportMapType::SCSS === $entry->type) {
+                $pattern = self::URL_PATTERN_DIST_STYLE;
+            } else {
+                $pattern = self::URL_PATTERN_DIST;
+            }
```
dsdeboer:
```suggestion
            $pattern = match ($entry->type) {
                ImportMapType::CSS, ImportMapType::SCSS => self::URL_PATTERN_DIST_STYLE,
                default => self::URL_PATTERN_DIST,
            };
```

dsdeboer (commented):
Hi @tip2tail, I left some suggestions as improvements as I can't figure out how to push them as a PR 😊. I hope these help, as the tests seem to work on my machine after the changes. 

Also, rebasing from the target branch could help otherwise

tip2tail:
Thanks for the suggestions @dsdeboer .  I will refactor as suggested - I always forget about `match` 🙄.

Hi @weaverryan - I have been having the same thoughts.  It's a bit of a step change from how we as developers would have been used to with `node` and installing the required packages - so I think this will take some consideration and adjustment.  That said I love this concept and I am moving forward with converting my project to **AssetMapper**.  I have however hit some of the issues you mention, namely that some of my Sass resources require additional components.  In my instance, Bootswatch is actually provided as a Composer package as well - solves it for _me_ for _this package only_, but I think we can agree the problem persists.

I am not sure what the best approach here is... 🫤

weaverryan:
If the biggest problem is Sass, I'm not eager to make any changes to how `importmap:require` is handled (e.g. downloading the entire package or having a way to hide something from the importmap & so just use `importmap:require` as a way to fetch packages). Sass's usefulness, in theory, may be waning due to CSS nesting and CSS properties. And if Bootstrap & bootswatch both come via Packagist (we should document bootswatch being on packagist on the symfonycasts bundle), then perhaps that, sort of, "polyfills" the situation: the most popular libraries that a user might want Sass for, *do* have a solution for fetching them.

Basically, I'm more eager for us to find other solutions to Sass (or potentially other use-cases for other files in a package) than to change AssetMapper. That doesn't mean we can't make some changes, but the need should be proven first.

WebMamba (commented):
I agree with Ryan here. If we do it for sass, what about typescript, less, images, or anything other specific file you want ?

My tough about it is it should not be a the AssetMapper responsability to deal with other file type than js or css. 
Maybe we can let bundles add extra allowing type to download so if you want to download sass the responsabilies goes to the Sassbundle.
But I not really sure about it.

weaverryan:
> Maybe we can let bundles add extra allowing type to download so if you want to download sass the responsabilies goes to the Sassbundle.

This is interesting and makes me think of another idea. I "think" that Sass is a bit unique in needing extra source files. Perhaps, if we care enough, SassBundle itself could have a small system for this - e.g. in config, you say you need the `bootstrap` source files. Then, on `sass:build`, it downloads the `bootstrap` package (e.g. hidden into `var/` like it does with the `sass` binary itself or potentially in `assets/vendor/`) if it's not already present.

PR #52981, opened 2023-12-09
[FrameworkBundle] Make `ValidatorCacheWarmer` and `SerializeCacheWarmer` use `kernel.build_dir` instead of `kernel.cache_dir`

okhoshi:
| Q             | A
| ------------- | ---
| Branch?       | 7.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | none
| License       | MIT

Follow up to #50391, set up the `ValidatorCacheWarmer` and `SerializerCacheWarmer` to use the `kernel.build_dir` instead of `kernel.cache_dir`. The value conveyed by their constructor parameter has been changed to prevent developers to configure the cache file outside of the build_dir.

`AbstractPhpFileCacheWarmer` has been reworked a bit too, to reduce the risk of misconfiguration.

#SymfonyHackday

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ValidatorCacheWarmer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ValidatorCacheWarmer.php
@@ -41,6 +41,10 @@ class ValidatorCacheWarmer extends AbstractPhpFileCacheWarmer
 
     protected function doWarmUp(string $cacheDir, ArrayAdapter $arrayAdapter, ?string $buildDir = null): bool
     {
+        if (!$buildDir) {
+            return false;
+        }
+
         $loaders = $this->validatorBuilder->getLoaders();
         $metadataFactory = new LazyLoadingMetadataFactory(new LoaderChain($loaders), $arrayAdapter);
 
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/serializer.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/serializer.php
@@ -56,7 +56,7 @@ use Symfony\Component\Serializer\SerializerInterface;
 
 return static function (ContainerConfigurator $container) {
     $container->parameters()
-        ->set('serializer.mapping.cache.file', '%kernel.cache_dir%/serialization.php')
+        ->set('serializer.mapping.cache.file', '%kernel.build_dir%/serialization.php')
     ;
 
     $container->services()
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/validator.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/validator.php
@@ -28,7 +28,7 @@ use Symfony\Component\Validator\ValidatorBuilder;
 
 return static function (ContainerConfigurator $container) {
     $container->parameters()
-        ->set('validator.mapping.cache.file', param('kernel.cache_dir').'/validation.php');
+        ->set('validator.mapping.cache.file', '%kernel.build_dir%/validation.php');
 
     $validatorsDir = \dirname((new \ReflectionClass(EmailValidator::class))->getFileName());
 
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/CacheWarmer/SerializerCacheWarmerTest.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/CacheWarmer/SerializerCacheWarmerTest.php
@@ -30,9 +30,50 @@ class SerializerCacheWarmerTest extends TestCase
         @unlink($file);
 
         $warmer = new SerializerCacheWarmer($loaders, $file);
-        $warmer->warmUp(\dirname($file));
+        $warmer->warmUp(\dirname($file), \dirname($file));
 
         $this->assertFileExists($file);
+
+        $arrayPool = new PhpArrayAdapter($file, new NullAdapter());
+
+        $this->assertTrue($arrayPool->getItem('Symfony_Bundle_FrameworkBundle_Tests_Fixtures_Serialization_Person')->isHit());
+        $this->assertTrue($arrayPool->getItem('Symfony_Bundle_FrameworkBundle_Tests_Fixtures_Serialization_Author')->isHit());
+    }
+
+    /**
+     * @dataProvider loaderProvider
+     */
+    public function testWarmUpAbsoluteFilePath(array $loaders)
+    {
+        $file = sys_get_temp_dir().'/0/cache-serializer.php';
+        @unlink($file);
+
+        $cacheDir = sys_get_temp_dir().'/1';
+
+        $warmer = new SerializerCacheWarmer($loaders, $file);
+        $warmer->warmUp($cacheDir, $cacheDir);
+
+        $this->assertFileExists($file);
+        $this->assertFileDoesNotExist($cacheDir.'/cache-serializer.php');
+
+        $arrayPool = new PhpArrayAdapter($file, new NullAdapter());
+
+        $this->assertTrue($arrayPool->getItem('Symfony_Bundle_FrameworkBundle_Tests_Fixtures_Serialization_Person')->isHit());
+        $this->assertTrue($arrayPool->getItem('Symfony_Bundle_FrameworkBundle_Tests_Fixtures_Serialization_Author')->isHit());
+    }
+
+    /**
+     * @dataProvider loaderProvider
+     */
+    public function testWarmUpWithoutBuildDir(array $loaders)
+    {
+        $file = sys_get_temp_dir().'/cache-serializer.php';
+        @unlink($file);
+
+        $warmer = new SerializerCacheWarmer($loaders, $file);
+        $warmer->warmUp(\dirname($file));
+
+        $this->assertFileDoesNotExist($file);
 
         $arrayPool = new PhpArrayAdapter($file, new NullAdapter());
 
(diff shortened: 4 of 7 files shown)


nicolas-grekas (commented):
I wonder if it wouldn't be nicer to adapt the behavior based on whether the file is relative or absolute?
This way we don't need any deprecations nor any new methods:
- when the path is absolute, we use that as we do now
- when the path is relative, we put the file in the build dir

And of course we make the path relative by default.

okhoshi:
> I wonder if it wouldn't be nicer to adapt the behavior based on whether the file is relative or absolute? 

I also considered that option, and haven't realized `symfony/filesystem` was already required in the Bundle. Will change to that 👍 

> This way we don't need any deprecations nor any new methods:
> 
> * when the path is absolute, we use that as we do now
> * when the path is relative, we put the file in the build dir

However, the deprecations and the new method are still necessary to let the extending class make the choice where to store the file. If we go with the current code, and the logic that when the path is relative we store in the build dir, it would force every CacheWarmer extending the `AbstractPhpFileCacheWarmer` to use the build dir.
Or did I miss something ? :)

src/Symfony/Bundle/FrameworkBundle/CacheWarmer/AbstractPhpFileCacheWarmer.php
```diff
@@ -20,11 +20,13 @@
...
-     * @param string $phpArrayFile The PHP file where metadata are cached
+     * @param string|null $phpArrayFile The PHP file where metadata are cached
      */
-    public function __construct(
-        private string $phpArrayFile,
-    ) {
+    public function __construct(private ?string $phpArrayFile = null)
+    {
+        if (\func_num_args() > 0) {
+            trigger_deprecation('symfony/framework-bundle', '7.1', 'Setting "string $phpArrayFile" argument is deprecated. The argument will be removed in 8.0.');
```
nicolas-grekas:
```suggestion
            trigger_deprecation('symfony/framework-bundle', '7.1', 'The $phpArrayFile of [...]() is deprecated and will be removed in 8.0.');
```

src/Symfony/Bundle/FrameworkBundle/CacheWarmer/AbstractPhpFileCacheWarmer.php
```diff
@@ -50,7 +52,12 @@ public function warmUp(string $cacheDir, ?string $buildDir = null): array
         // so here we un-serialize the values first
         $values = array_map(fn ($val) => null !== $val ? unserialize($val) : null, $arrayAdapter->getValues());
 
-        return $this->warmUpPhpArrayAdapter(new PhpArrayAdapter($this->phpArrayFile, new NullAdapter()), $values);
+        $filepath = $this->getPhpArrayFile($cacheDir, $buildDir);
+        if (!$filepath) {
```
nicolas-grekas:
```suggestion
        if (!$filepath = $this->getPhpArrayFile($cacheDir, $buildDir)) {
```

src/Symfony/Bundle/FrameworkBundle/Resources/config/validator.php
```diff
@@ -63,7 +63,7 @@
         ->set('validator.mapping.cache.adapter', PhpArrayAdapter::class)
             ->factory([PhpArrayAdapter::class, 'create'])
             ->args([
-                param('validator.mapping.cache.file'),
+                param('kernel.build_dir').'/'.param('validator.mapping.cache.file'),
```
nicolas-grekas: this changes the meaning of the parameter, potential BC break ahead

src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ValidatorCacheWarmer.php
```diff
@@ -70,6 +75,19 @@ protected function warmUpPhpArrayAdapter(PhpArrayAdapter $phpArrayAdapter, array
         return parent::warmUpPhpArrayAdapter($phpArrayAdapter, $values);
     }
 
+    protected function getPhpArrayFile(string $cacheDir, ?string $buildDir = null): ?string
+    {
+        if (!$buildDir) {
+            return null;
+        }
+
+        if (Path::isRelative($this->phpArrayFile)) {
```
nicolas-grekas: I'm not convinced we need this logic, it adds complexity IMHO

nicolas-grekas (commented):
I'm not convinced about the refactoring, it brings a potential BC break.
Why not just change the value in the DI?
We can't prevent people from doing mistakes anyway :shrug:

nicolas-grekas:
I'm having a hard time reasoning about the way things work during warmup...

Right now, we made warmers be called with $buildDir set to null during post-compile warmups.
Would it make sense to change this to be:
- set cache_dir to build_dir during compile-time warmups
- keep cache_dir and build_dir to their defined values during post-compile warmups?

cache warmers would then be able to compare $cacheDir and $buildDir to decide if there is a readonly cache they shouldn't touch.

Wouldn't that make all your related PRs simpler? (The most important change being to set $cacheDir to the build-dir during compile-time warmups)

okhoshi:
@nicolas-grekas I took a step back regarding this whole build_dir/cache_dir situation, and you are right that we can't prevent people to do mistakes anyway 👍 I could indeed drastically simplify this PR once I accepted that fact 🙂. 

However, passing the buildDir as null when we run the warmers in post-compile warmup is equivalent to your suggestion, so I suggest to keep that as it is.

src/Symfony/Bundle/FrameworkBundle/CacheWarmer/ValidatorCacheWarmer.php
okhoshi: Removed

src/Symfony/Bundle/FrameworkBundle/Resources/config/validator.php
okhoshi: Reverted to the previous meaning

nicolas-grekas:
Can you please rebase on top of 7.2?

src/Symfony/Bundle/FrameworkBundle/Tests/CacheWarmer/ValidatorCacheWarmerTest.php
```diff
@@ -122,8 +177,12 @@ public function testClassAutoloadExceptionWithUnrelatedException()
             }
         }, true, true);
 
-        $warmer->warmUp('foo');
+        try {
+            $warmer->warmUp(\dirname($file), \dirname($file));
+        } finally {
+            $this->assertFileDoesNotExist($file);
```
nicolas-grekas, line 183: why in finally also here?

src/Symfony/Bundle/FrameworkBundle/Tests/CacheWarmer/SerializerCacheWarmerTest.php
```diff
@@ -112,8 +160,12 @@ public function testClassAutoloadExceptionWithUnrelatedException()
             }
         }, true, true);
 
-        $warmer->warmUp('foo');
+        try {
+            $warmer->warmUp(\dirname($file), \dirname($file));
+        } finally {
+            $this->assertFileDoesNotExist($file);
```
nicolas-grekas, line 166: why put this in the finally block?

nicolas-grekas (approved):

src/Symfony/Bundle/FrameworkBundle/Tests/CacheWarmer/SerializerCacheWarmerTest.php
okhoshi: Well, it's been a while since I wrote that, but IIRC, it's because the test is purposefully making `warmUp` throws an exception, so there's no way to ensure no file gets written in that case without the finally block.

nicolas-grekas: then we should catch the expected type of exception instead and put that there

src/Symfony/Bundle/FrameworkBundle/Tests/CacheWarmer/SerializerCacheWarmerTest.php
```diff
@@ -112,8 +160,12 @@ public function testClassAutoloadExceptionWithUnrelatedException()
             }
         }, true, true);
 
-        $warmer->warmUp('foo');
+        try {
+            $warmer->warmUp(\dirname($file), \dirname($file));
+        } finally {
```
okhoshi:
```suggestion
        } catch (\DomainException) {
```

@nicolas-grekas Is it what you had in mind ?

nicolas-grekas: almost: we should rethrow the exception
and keep the other call in finally
while at it, we should move the expectException calls just before the call to wamrup (outside of the try/catch)

okhoshi: Will apply the changes later today 👍

okhoshi:
@nicolas-grekas Anything missing on this PR to get it merged ? :)

okhoshi:
@nicolas-grekas @OskarStark Could we still get this merged in 7.3 ?

src/Symfony/Bundle/FrameworkBundle/Resources/config/serializer.php
```diff
@@ -164,7 +164,10 @@
 
         ->set('serializer.mapping.cache.symfony', CacheItemPoolInterface::class)
             ->factory([PhpArrayAdapter::class, 'create'])
-            ->args([param('serializer.mapping.cache.file'), service('cache.serializer')])
+            ->args([
```
fabpot: Can be reverted

fabpot (approved):

fabpot:
Thank you @Okhoshi.

PR #52986, opened 2023-12-10
[HttpFoundation] Similar locale selection

Spomky:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Issues        | none
| License       | MIT

Allow the nearest locale to be selected instead the default one.
Ping @welcoMattic 

I noted a non-optimized locale selection. Let say I have the following controller and root route:

```php
#[Route('/', name: 'app_root', methods: [Request::METHOD_GET])]
public function indexNoLocale(Request $request): Response
{
    $locale = $request->getPreferredLanguage($this->supportedLocales) ?? $this->defaultLocale;

    return $this->redirectToRoute('app_homepage', [
        '_locale' => $locale,
    ], Response::HTTP_SEE_OTHER);
}
```

And the following parameters:
* `$this->supportedLocales` = `['fr_FR', 'en_US']`
* `$this->defaultLocale` = `'en_US'`

When a user arrives on this page with the header `accept-language: ja-JP,fr_CA;q=0.7,fr;q=0.5`, the default locale `en_US` is returned.
In this situation, I would expect to use `fr_FR` as the browser indicates French is a possible choice (`fr_CA` and `fr`).

With this change, and if no exact match is found, a similar language is selected. In this example, `fr_FR` is acceptable and then returned.

### DIFF
--- a/src/Symfony/Component/HttpFoundation/Request.php
+++ b/src/Symfony/Component/HttpFoundation/Request.php
@@ -1532,24 +1532,25 @@ class Request
             return $preferredLanguages[0] ?? null;
         }
 
+        $locales = array_map($this->formatLocale(...), $locales ?? []);
         if (!$preferredLanguages) {
             return $locales[0];
         }
 
-        $extendedPreferredLanguages = [];
-        foreach ($preferredLanguages as $language) {
-            $extendedPreferredLanguages[] = $language;
-            if (false !== $position = strpos($language, '_')) {
-                $superLanguage = substr($language, 0, $position);
-                if (!\in_array($superLanguage, $preferredLanguages, true)) {
-                    $extendedPreferredLanguages[] = $superLanguage;
+        if ($matches = array_intersect($preferredLanguages, $locales)) {
+            return current($matches);
+        }
+
+        $combinations = array_merge(...array_map($this->getLanguageCombinations(...), $preferredLanguages));
+        foreach ($combinations as $combination) {
+            foreach ($locales as $locale) {
+                if (str_starts_with($locale, $combination)) {
+                    return $locale;
                 }
             }
         }
 
-        $preferredLanguages = array_values(array_intersect($extendedPreferredLanguages, $locales));
-
-        return $preferredLanguages[0] ?? $locales[0];
+        return $locales[0];
     }
 
     /**
@@ -1567,30 +1568,89 @@ class Request
         $this->languages = [];
         foreach ($languages as $acceptHeaderItem) {
             $lang = $acceptHeaderItem->getValue();
-            if (str_contains($lang, '-')) {
-                $codes = explode('-', $lang);
-                if ('i' === $codes[0]) {
-                    // Language not listed in ISO 639 that are not variants
-                    // of any listed language, which can be registered with the
-                    // i-prefix, such as i-cherokee
-                    if (\count($codes) > 1) {
-                        $lang = $codes[1];
-                    }
-                } else {
-                    for ($i = 0, $max = \count($codes); $i < $max; ++$i) {
-                        if (0 === $i) {
-                            $lang = strtolower($codes[0]);
-                        } else {
-                            $lang .= '_'.strtoupper($codes[$i]);
-                        }
-                    }
-                }
-            }
-
-            $this->languages[] = $lang;
+            $this->languages[] = $this->formatLocale($lang);
         }
+        $this->languages = array_unique($this->languages);
 
         return $this->languages;
+    }
+
+    /**
+     * Strips the locale to only keep the canonicalized language value.
+     *
+     * Depending on the $locale value, this method can return values like :
+     * - language_Script_REGION: "fr_Latn_FR", "zh_Hans_TW"
+     * - language_Script: "fr_Latn", "zh_Hans"
+     * - language_REGION: "fr_FR", "zh_TW"
+     * - language: "fr", "zh"
+     *
+     * Invalid locale values are returned as is.
+     *
+     * @see https://wikipedia.org/wiki/IETF_language_tag
+     * @see https://datatracker.ietf.org/doc/html/rfc5646
+     */
+    private static function formatLocale(string $locale): string
+    {
+        [$language, $script, $region] = self::getLanguageComponents($locale);
+
+        return implode('_', array_filter([$language, $script, $region]));
+    }
+
+    /**
+     * Returns an array of all possible combinations of the language components.
+     *
+     * For instance, if the locale is "fr_Latn_FR", this method will return:
+     * - "fr_Latn_FR"
+     * - "fr_Latn"
+     * - "fr_FR"
+     * - "fr"
+     *
+     * @return string[]
+     */
+    private static function getLanguageCombinations(string $locale): array
+    {
+        [$language, $script, $region] = self::getLanguageComponents($locale);
+
+        return array_unique([
+            implode('_', array_filter([$language, $script, $region])),
+            implode('_', array_filter([$language, $script])),
+            implode('_', array_filter([$language, $region])),
+            $language,
+        ]);
+    }
+
+    /**
+     * Returns an array with the language components of the locale.
+     *
+     * For example:
+     * - If the locale is "fr_Latn_FR", this method will return "fr", "Latn", "FR"
+     * - If the locale is "fr_FR", this method will return "fr", null, "FR"
+     * - If the locale is "zh_Hans", this method will return "zh", "Hans", null
+     *
+     * @see https://wikipedia.org/wiki/IETF_language_tag
+     * @see https://datatracker.ietf.org/doc/html/rfc5646
+     *
+     * @return array{string, string|null, string|null}
+     */
+    private static function getLanguageComponents(string $locale): array
+    {
+        $locale = str_replace('_', '-', strtolower($locale));
+        $pattern = '/^([a-zA-Z]{2,3}|i-[a-zA-Z]{5,})(?:-([a-zA-Z]{4}))?(?:-([a-zA-Z]{2}))?(?:-(.+))?$/';
+        if (!preg_match($pattern, $locale, $matches)) {
+            return [$locale, null, null];
+        }
+        if (str_starts_with($matches[1], 'i-')) {
+            // Language not listed in ISO 639 that are not variants
+            // of any listed language, which can be registered with the
+            // i-prefix, such as i-cherokee
+            $matches[1] = substr($matches[1], 2);
+        }
+
+        return [
+            $matches[1],
+            isset($matches[2]) ? ucfirst(strtolower($matches[2])) : null,
+            isset($matches[3]) ? strtoupper($matches[3]) : null,
+        ];
     }
 
     /**
--- a/src/Symfony/Component/HttpFoundation/Tests/RequestTest.php
+++ b/src/Symfony/Component/HttpFoundation/Tests/RequestTest.php
@@ -1502,27 +1502,43 @@ class RequestTest extends TestCase
     {
         $request = new Request();
         $this->assertNull($request->getPreferredLanguage());
-        $this->assertNull($request->getPreferredLanguage([]));
-        $this->assertEquals('fr', $request->getPreferredLanguage(['fr']));
-        $this->assertEquals('fr', $request->getPreferredLanguage(['fr', 'en']));
-        $this->assertEquals('en', $request->getPreferredLanguage(['en', 'fr']));
-        $this->assertEquals('fr-ch', $request->getPreferredLanguage(['fr-ch', 'fr-fr']));
+    }
 
+    /**
+     * @dataProvider providePreferredLanguage
+     */
+    public function testPreferredLanguageWithLocales(?string $expectedLocale, ?string $acceptLanguage, array $locales)
+    {
         $request = new Request();
-        $request->headers->set('Accept-language', 'zh, en-us; q=0.8, en; q=0.6');
-        $this->assertEquals('en', $request->getPreferredLanguage(['en', 'en-us']));
+        if ($acceptLanguage) {
+            $request->headers->set('Accept-language', $acceptLanguage);
+        }
+        $this->assertSame($expectedLocale, $request->getPreferredLanguage($locales));
+    }
 
-        $request = new Request();
-        $request->headers->set('Accept-language', 'zh, en-us; q=0.8, en; q=0.6');
-        $this->assertEquals('en', $request->getPreferredLanguage(['fr', 'en']));
-
-        $request = new Request();
-        $request->headers->set('Accept-language', 'zh, en-us; q=0.8');
-        $this->assertEquals('en', $request->getPreferredLanguage(['fr', 'en']));
-
-        $request = new Request();
-        $request->headers->set('Accept-language', 'zh, en-us; q=0.8, fr-fr; q=0.6, fr; q=0.5');
-        $this->assertEquals('en', $request->getPreferredLanguage(['fr', 'en']));
+    public static function providePreferredLanguage(): iterable
+    {
+        yield '"es_PA" is selected as no supported locale is set' => ['es_PA', 'es-pa, en-us; q=0.8, en; q=0.6', []];
+        yield 'No supported locales' => [null, null, []];
+        yield '"fr" selected as first choice when no header is present' => ['fr', null, ['fr', 'en']];
+        yield '"en" selected as first choice when no header is present' => ['en', null, ['en', 'fr']];
+        yield '"fr_CH" selected as first choice when no header is present' => ['fr_CH', null, ['fr-ch', 'fr-fr']];
+        yield '"en_US" is selected as an exact match is found (1)' => ['en_US', 'zh, en-us; q=0.8, en; q=0.6', ['en', 'en-us']];
+        yield '"en_US" is selected as an exact match is found (2)' => ['en_US', 'ja-JP,fr_CA;q=0.7,fr;q=0.5,en_US;q=0.3', ['en_US', 'fr_FR']];
+        yield '"en" is selected as an exact match is found' => ['en', 'zh, en-us; q=0.8, en; q=0.6', ['fr', 'en']];
+        yield '"fr" is selected as an exact match is found' => ['fr', 'zh, en-us; q=0.8, fr-fr; q=0.6, fr; q=0.5', ['fr', 'en']];
+        yield '"en" is selected as "en-us" is a similar dialect' => ['en', 'zh, en-us; q=0.8', ['fr', 'en']];
+        yield '"fr_FR" is selected as "fr_CA" is a similar dialect (1)' => ['fr_FR', 'ja-JP,fr_CA;q=0.7,fr;q=0.5', ['en_US', 'fr_FR']];
+        yield '"fr_FR" is selected as "fr_CA" is a similar dialect (2)' => ['fr_FR', 'ja-JP,fr_CA;q=0.7', ['en_US', 'fr_FR']];
+        yield '"fr_FR" is selected as "fr" is a similar dialect' => ['fr_FR', 'ja-JP,fr;q=0.5', ['en_US', 'fr_FR']];
+        yield '"fr_FR" is selected as "fr_CA" is a similar dialect and has a greater "q" compared to "en_US" (2)' => ['fr_FR', 'ja-JP,fr_CA;q=0.7,ru-ru;q=0.3', ['en_US', 'fr_FR']];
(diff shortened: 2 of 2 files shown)


Nyholm (commented):
Thank you. It looks cool.

Can you also test these: 

```
ja-JP,fr;q=0.5
ja-JP,fr_CA;q=0.7
```

Both of them should go to `fr_FR`, right? 

----

How about these?

```
ja-JP,fr_CA;q=0.7,fr;q=0.5,en_US;q=0.3
ja-JP,fr;q=0.5,en_US;q=0.3
ja-JP,fr_CA;q=0.7,en_US;q=0.3
ja-JP,fr_CA;q=0.7,en;q=0.5
```

welcoMattic:
Maybe for another PR, but shouldn't we add support for `q=` parameter? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language#directives

Spomky:
Many thanks for your comments. Indeed there is something wrong with some cases [such as this one](https://github.com/symfony/symfony/blob/48c8c981851580b877c3b82c7980fc55350a886d/src/Symfony/Component/HttpFoundation/Tests/RequestTest.php#L1529) where `en` is selected when `en-us` is preferred just because of the order of the array values.
I will add use cases and make sure `q` is understood.

Spomky:
Hi,

I updated the PR and changed the comparison logic:
* If an exact match exists, it is returned
* Else if a similar locale exists, the corresponding language is returned
* If nothing is acceptable, the first supported locale is returned

Please note that the `q` parameter is already used. The [test with a reversed order priority](https://github.com/symfony/symfony/pull/52986/files#diff-a6e313421b40e37ad1c1a23ad66131c48790cb435dc1d040ac33a4800e243780R1553) returns the expected language.

In addition, the returned locale value is always `xx_XX` to make it consistant with the value used internally. 
Also, it accepts locales such as `hi_Latn_IN` to be compared.

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1609,6 +1609,7 @@ public function getPreferredFormat(?string $default = 'html'): ?string
     public function getPreferredLanguage(array $locales = null): ?string
     {
         $preferredLanguages = $this->getLanguages();
+        $locales = array_map([$this, 'formatLocale'], $locales ?? []);
```
yceruto: Since PHP 8.1, the project preference is to use first-class callable syntax instead of the array syntax.

```suggestion
        $locales = array_map($this->formatLocale(...), $locales ?? []);
```

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1649,32 +1662,39 @@ public function getLanguages(): array
...
 
             $this->languages[] = $lang;
         }
 
         return $this->languages;
     }
 
+    private static function formatLocale(string $locale): string
+    {
+        if (str_contains($locale, '-')) {
```
yceruto: Now that you have moved this code to a new method, let's refactor it a bit by returning early if `!str_contains($locale, '-')`, thus removing one nested level.

Spomky: You are right, I will update this PR and remove a level of nesting.

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1649,32 +1662,40 @@ public function getLanguages(): array
...
+        if (!str_contains($locale, '-')) {
+            return $locale;
+        }
+        $codes = explode('-', $locale);
+        if ('i' === $codes[0]) {
+            // Language not listed in ISO 639 that are not variants
+            // of any listed language, which can be registered with the
+            // i-prefix, such as i-cherokee
+            if (\count($codes) > 1) {
+                $locale = $codes[1];
```
stloyd:
```suggestion
                return $codes[1];
```

Spomky: Hi @stloyd,

Many thanks for the suggestion.
I will not commit it as I am working on this part that needs other modifications.
I noticed issues when variant are present in the locale e.g. zh-hans-tw (simplified Chinese from Taiwan) or when components/extensions follow the region (`pt-BR-u-ca-gregory-nu-latn` for Brazilian Portuguese with use of the Gregorian calendar and Latin numbering).
This function should return `zh_Hans_TW` or `pt_BR`, but it does not and an exact match may be missed because of this.
For `zh_Hans_TW`, it could fallback to `zh_TW` and `zh` but this is not the case at the moment.

@welcoMattic I think we could add a class to handle these locale parts in the Translation component. This will help the Request object handle the languages and regions parts correctly to find the most appropriate locale. WDYT?
See https://en.wikipedia.org/wiki/IETF_language_tag or https://andreipfeiffer.dev/slides/i18n-and-l10n.pdf

Something like as follows

```php
final class LanguageTag
{
    public string $language;
    public ?string $script = null;
    public ?string $region = null;
    public array $variants = [];
    public array $extensions = [];

    public function __construct(string $languageTag)
    {
        $components = explode('-', strtolower($languageTag));
        $this->language = array_shift($components);

        foreach ($components as $component) {
            if (strlen($component) === 4) {
                $this->script = ucfirst($component);
            } elseif (strlen($component) === 2) {
                $this->region = mb_strtoupper($component);
            }
            // Variants and extensions to handle
        }
    }
}
```

welcoMattic: I'm wondering the right place for this class, Translation or Intl? @Nyholm @OskarStark any thoughts?

Spomky: Note that for this PR we only need to capture the main parts.

Spomky: Hi,

I completely reviewed the method and it now able to differentiate locales with language/script/region.
Also, I added some tests to make sure the selection is more accurate.

OskarStark: > I'm wondering the right place for this class, Translation or Intl? @Nyholm @OskarStark any thoughts?

My feeling is about intl, but not sure

cc @stof

Spomky: I will start working on this soon. I'll create it in the Intl component and see if it needs to be moved elsewhere as it progresses.

smnandre: That is what intl provide (the PHP extension) with the [Locale](https://www.php.net/manual/en/class.locale.php) classes.

symfony/intl is kinda polyfill for those without intl extension. You may want to look at the int-contracts, but i really do not think symfony/intl should be the place for that.

As nowadays it contains massive amount of data... and finally not much class / helpers.

smnandre:
In your exemple, what would happen if the default locale were defined as « fr, en » ?

src/Symfony/Component/HttpFoundation/Request.php
Spomky: OK thanks. I will reconsider this.
For this PR, there is no need to depend on other components.

Spomky:
> In your exemple, what would happen if the default locale were defined as « fr, en » ?

You mean if the supported locales are `['fr', 'en']` instead of `['fr-FR', 'en-US']`? In this case it will match as `accept-language: ja-JP,fr_CA;q=0.7,fr;q=0.5` contains the exact match `fr`.

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1649,32 +1654,85 @@ public function getLanguages(): array
...
+     */
+    private static function formatLocale(string $locale): string
+    {
+        [$language, $script, $region] = self::getLanguageComponents($locale);
+
+        return implode('_', array_filter([$language, $script, $region]));
+    }
+
+    /**
+     * This method returns an array of all possible combinations of the language components.
```
```suggestion
     * Returns an array of all possible combinations of the language components.
     *
```

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1649,32 +1654,85 @@ public function getLanguages(): array
...
-            $this->languages[] = $lang;
+            $this->languages[] = $this->formatLocale($lang);
         }
+        $this->languages = array_unique($this->languages);
 
         return $this->languages;
     }
 
+    /**
+     * This method strips the locale to only keep the canonicalized language value.
```
nicolas-grekas: blank line after (to separate the title from the description)

```suggestion
     * Strips the locale to only keep the canonicalized language value.
     *
```

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1649,32 +1654,85 @@ public function getLanguages(): array
...
+     * @see https://wikipedia.org/wiki/IETF_language_tag
+     * @see https://datatracker.ietf.org/doc/html/rfc5646
+     *
+     * @return array{language: string, script: string|null, region: string|null}
+     */
+    private static function getLanguageComponents(string $locale): array
+    {
+        $locale = str_replace('_', '-', strtolower($locale));
+        $pattern = '/^([a-zA-Z]{2,3}|i-[a-zA-Z]{5,})(?:-([a-zA-Z]{4}))?(?:-([a-zA-Z]{2}))?(?:-(.+))?$/';
+        if (1 !== preg_match($pattern, $locale, $matches)) {
```
```suggestion
        if (!preg_match($pattern, $locale, $matches)) {
```

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1618,20 +1619,24 @@ public function getPreferredLanguage(array $locales = null): ?string
             return $locales[0];
         }
 
-        $extendedPreferredLanguages = [];
+        $matches = array_values(array_intersect($preferredLanguages, $locales));
+        if ($matches) {
```
```suggestion
        if ($matches = array_values(array_intersect($preferredLanguages, $locales))) {
```

nicolas-grekas (commented):
Here's some nitpicking. What's the state of this PR? Any votes from previous reviewers?

Spomky:
Many thanks for the last suggestions.
The PR is now rebased, squashed and ready for other reviewers comments

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1618,20 +1619,23 @@ public function getPreferredLanguage(?array $locales = null): ?string
             return $locales[0];
         }
 
-        $extendedPreferredLanguages = [];
+        if ($matches = array_values(array_intersect($preferredLanguages, $locales))) {
```
nicolas-grekas: we'd better remove `array_values` and `return reset($matches);` on the next line

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1618,20 +1619,23 @@ public function getPreferredLanguage(?array $locales = null): ?string
...
+            return $matches[0];
+        }
+        $combinations = [];
```
nicolas-grekas: `$combinations = array_merge(...array_map($this->getLanguageCombinations(...), $peferredLanguages));`

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1649,32 +1653,87 @@ public function getLanguages(): array
...
+            implode('_', array_filter([$language, $script, $region])),
+            implode('_', array_filter([$language, $script])),
+            implode('_', array_filter([$language, $region])),
+            $language,
+        ]);
+    }
+
+    /**
+     * This method returns an array with the language components of the locale.
+     * For example:
```
nicolas-grekas, line 1625: missing blank line before this line

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1649,32 +1653,87 @@ public function getLanguages(): array
         $this->languages = [];
         foreach ($languages as $acceptHeaderItem) {
             $lang = $acceptHeaderItem->getValue();
-            if (str_contains($lang, '-')) {
-                $codes = explode('-', $lang);
-                if ('i' === $codes[0]) {
-                    // Language not listed in ISO 639 that are not variants
-                    // of any listed language, which can be registered with the
-                    // i-prefix, such as i-cherokee
```
nicolas-grekas, line 1575: can we keep this comment somewhere?

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1649,32 +1653,87 @@ public function getLanguages(): array
...
+     * This method returns an array with the language components of the locale.
+     * For example:
+     * - If the locale is "fr_Latn_FR", this method will return "fr", "Latn", "FR"
+     * - If the locale is "fr_FR", this method will return "fr", null, "FR"
+     * - If the locale is "zh_Hans", this method will return "zh", "Hans", null.
+     *
+     * @see https://wikipedia.org/wiki/IETF_language_tag
+     * @see https://datatracker.ietf.org/doc/html/rfc5646
+     *
+     * @return array{language: string, script: string|null, region: string|null}
```
```suggestion
     * @return array{string, string|null, string|null}
```

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1618,20 +1619,23 @@ public function getPreferredLanguage(?array $locales = null): ?string
...
         foreach ($preferredLanguages as $language) {
-            $extendedPreferredLanguages[] = $language;
-            if (false !== $position = strpos($language, '_')) {
-                $superLanguage = substr($language, 0, $position);
-                if (!\in_array($superLanguage, $preferredLanguages)) {
-                    $extendedPreferredLanguages[] = $superLanguage;
+            $combinations = array_merge($combinations, $this->getLanguageCombinations($language));
+        }
+
+        foreach (array_unique($combinations) as $combination) {
```
nicolas-grekas: let's skip array_unique

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1649,32 +1653,87 @@ public function getLanguages(): array
...
+    }
+
+    /**
+     * Returns an array of all possible combinations of the language components.
+     *
+     * For example, if the locale is "fr_Latn_FR", this method will return :
+     * - "fr_Latn_FR"
+     * - "fr_Latn"
+     * - "fr_FR"
+     * - "fr".
```
```suggestion
     * - "fr"
```

nicolas-grekas (commented):
That's an improvement, for 7.1 ;)

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1609,6 +1609,7 @@ public function getPreferredFormat(?string $default = 'html'): ?string
     public function getPreferredLanguage(?array $locales = null): ?string
     {
         $preferredLanguages = $this->getLanguages();
+        $locales = array_map($this->formatLocale(...), $locales ?? []);
```
smnandre: Should be moved line 1621 maybe ? (micro-picking :) )

src/Symfony/Component/HttpFoundation/Request.php
Spomky: Yes. Done.

src/Symfony/Component/HttpFoundation/Request.php
Spomky: The shorter, the better. Done

src/Symfony/Component/HttpFoundation/Request.php
Spomky: Blak line added.

src/Symfony/Component/HttpFoundation/Request.php
Spomky: I restored this comment on line 1643 to 1645

src/Symfony/Component/HttpFoundation/Request.php
Spomky: Done

src/Symfony/Component/HttpFoundation/Request.php
Spomky: `array_unique` is now gone

src/Symfony/Component/HttpFoundation/Request.php
Spomky: DOne

src/Symfony/Component/HttpFoundation/Request.php
Spomky: Indeed. It moved this line

dunglas:
It looks like this isn't a good practice according to the spec: https://www.rfc-editor.org/rfc/rfc9110#field.accept-language

If `fr_BE` is requested, `fr_FR` shouldn't be sent. Instead, the browser should include both `fr_BE` and `fr` in `Accept-Language`.

dunglas (approved):

stof:
PHP already has a locale matching algorithm in ext-intl: https://www.php.net/manual/fr/locale.lookup.php (or maybe some of the other methods there)

Would this algorithm help for this PR ? If yes, the proper way to solve it might be to implement them in the polyfill in https://github.com/symfony/polyfill/blob/1.x/src/Intl/Icu/Locale.php (where it currently throw an exception saying it is not implemented) and to use it.

nicolas-grekas (approved):
LGTM thanks

> PHP already has a locale matching algorithm in ext-intl [...] the proper way to solve it might be to implement them in the polyfill

Good idea! I suggest merging this PR, and switch to the polyfill once we have it. Nothing is public here so the switch should be easy.

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1567,32 +1568,91 @@ public function getLanguages(): array
...
     }
 
+    /**
+     * Strips the locale to only keep the canonicalized language value.
+     *
+     * Depending on the $locale value, this method can return values like :
+     * - language_Script_REGION: "fr_Latn_FR", "zh_Hans_TW"
+     * - language_Script: "fr_Latn", "zh_Hans"
+     * - language_REGION: "fr_FR", "zh_TW"
+     * - language: "fr", "zh".
```
OskarStark:
```suggestion
     * - language: "fr", "zh"
```

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1567,32 +1568,91 @@ public function getLanguages(): array
...
+        ]);
+    }
+
+    /**
+     * This method returns an array with the language components of the locale.
+     *
+     * For example:
+     * - If the locale is "fr_Latn_FR", this method will return "fr", "Latn", "FR"
+     * - If the locale is "fr_FR", this method will return "fr", null, "FR"
+     * - If the locale is "zh_Hans", this method will return "zh", "Hans", null.
```
OskarStark:
```suggestion
     * - If the locale is "zh_Hans", this method will return "zh", "Hans", null
```

src/Symfony/Component/HttpFoundation/Tests/RequestTest.php
```diff
@@ -2199,7 +2213,7 @@ public function testFactory()
 
     public function testFactoryCallable()
     {
-        $requestFactory = new class {
+        $requestFactory = new class() {
```
OskarStark, line 2216: not sure about our code style here, with `()` or without

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1567,32 +1568,91 @@ public function getLanguages(): array
...
+    {
+        [$language, $script, $region] = self::getLanguageComponents($locale);
+
+        return implode('_', array_filter([$language, $script, $region]));
+    }
+
+    /**
+     * Returns an array of all possible combinations of the language components.
+     *
+     * For example, if the locale is "fr_Latn_FR", this method will return :
```
smnandre:
```suggestion
     * For instance, if the locale is "fr_Latn_FR", this method will return:
```

src/Symfony/Component/HttpFoundation/Tests/RequestTest.php
Spomky: fabbot patched it and added `()`

src/Symfony/Component/HttpFoundation/Request.php
```diff
@@ -1567,32 +1568,91 @@ public function getLanguages(): array
...
+        return array_unique([
+            implode('_', array_filter([$language, $script, $region])),
+            implode('_', array_filter([$language, $script])),
+            implode('_', array_filter([$language, $region])),
+            $language,
+        ]);
+    }
+
+    /**
+     * This method returns an array with the language components of the locale.
```
fabpot:
```suggestion
     * Returns an array with the language components of the locale.
```

fabpot (approved):

fabpot:
Thank you @Spomky.

Issue #52987, opened 2023-12-10
[TwigBridge] function `twig_test_empty` is deprecated in Twig 3.9


GromNaN:
### Symfony version(s) affected

5.4.31, 6.4.0, 7.0.0, 7.1

### Description

The `twig_*` global functions are deprecated in Twig 3.9 https://github.com/twigphp/Twig/pull/3883, and removed in Twig 4.0 https://github.com/twigphp/Twig/pull/3937

The function `twig_test_empty` is used in `Symfony\Bridge\Twig\Node\SearchAndRenderBlockNode`:

https://github.com/symfony/symfony/blob/228a275ddcc1a4301fc8865945c0686b56658135/src/Symfony/Bridge/Twig/Node/SearchAndRenderBlockNode.php#L53

https://github.com/symfony/symfony/blob/228a275ddcc1a4301fc8865945c0686b56658135/src/Symfony/Bridge/Twig/Node/SearchAndRenderBlockNode.php#L100

### How to reproduce

```
composer create-project symfony/demo
composer require "twig/twig: 3.x-dev"
```

Go to `https://localhost:8000/fr/admin/post/1/edit`

Deprecation message:

```
Since twig/twig 3.9.0: Using the internal "twig_test_empty" function is deprecated.
```

### Possible Solution

Duplicate the function as a static method in `SearchAndRenderBlockNode`.

xabbuh:
I'm on it

TMWagner:
@GromNaN 
OK - so it is deprecated. Is there a replacement?

GromNaN:
There is no replacement. This function is an implementation detail of the [`empty` test](https://twig.symfony.com/doc/3.x/tests/empty.html), that's why the new `CodeExtension::testEmpty()` is flagged as `@internal` and should not be used. It can be modified or removed in any minor version.

You should write your own function.

PR #52989, opened 2023-12-11
allow Twig 4

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | 
| License       | MIT

### DIFF
--- a/src/Symfony/Bridge/Twig/composer.json
+++ b/src/Symfony/Bridge/Twig/composer.json
@@ -18,7 +18,7 @@
     "require": {
         "php": ">=8.2",
         "symfony/translation-contracts": "^2.5|^3",
-        "twig/twig": "^3.0.4"
+        "twig/twig": "^3.0.4|^4.0"
     },
     "require-dev": {
         "egulias/email-validator": "^2.1.10|^3|^4",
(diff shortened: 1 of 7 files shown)


GromNaN (approved):
It should work once https://github.com/symfony/symfony/issues/52987 is fixed.

derrabus:
Why don't we target 6.4?

xabbuh:
> Why don't we target 6.4?

I considered this change as a new feature. If we chose 6.4 as the target, then why not 5.4?

fabpot:
Twig 4.x will probably never be released as is; this is "just" a branch that allows me to work on the next version that should be part of Symfony (7.1 hopefully). So, allowing it only on 7.1 for now works for me until we have a solid migration plan.

derrabus:
> I considered this change as a new feature.

We usually don't consider supporting new versions of 3rd party libraries as a new feature, but what @fabpot says makes sense to me. If Twig 4.x is a transitional branch that will eventually become symfony/twig 7.1, we should limit this change to the 7.1.x branch.

chalasr (approved):

fabpot (approved):

fabpot:
Thank you @xabbuh.

PR #52990, opened 2023-12-11
[TwigBridge] don't use deprecated and internal Twig functions

### LINKED ISSUE #52987: [TwigBridge] function `twig_test_empty` is deprecated in Twig 3.9

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52987
| License       | MIT

`twig_test_empty()` and `twig_escape_filter()` are deprecated since Twig 3.9

### DIFF
--- a/src/Symfony/Bridge/Twig/Node/SearchAndRenderBlockNode.php
+++ b/src/Symfony/Bridge/Twig/Node/SearchAndRenderBlockNode.php
@@ -12,6 +12,7 @@
 namespace Symfony\Bridge\Twig\Node;
 
 use Twig\Compiler;
+use Twig\Extension\CoreExtension;
 use Twig\Node\Expression\ArrayExpression;
 use Twig\Node\Expression\ConstantExpression;
 use Twig\Node\Expression\FunctionExpression;
@@ -50,7 +51,7 @@ final class SearchAndRenderBlockNode extends FunctionExpression
                         $labelIsExpression = false;
 
                         // Only insert the label into the array if it is not empty
-                        if (!twig_test_empty($label->getAttribute('value'))) {
+                        if (null !== $label->getAttribute('value') && false !== $label->getAttribute('value') && '' !== (string) $label->getAttribute('value')) {
                             $originalVariables = $variables;
                             $variables = new ArrayExpression([], $lineno);
                             $labelKey = new ConstantExpression('label', $lineno);
@@ -97,7 +98,12 @@ final class SearchAndRenderBlockNode extends FunctionExpression
 
                         // Check at runtime whether the label is empty.
                         // If not, add it to the array at runtime.
-                        $compiler->raw('(twig_test_empty($_label_ = ');
+                        if (method_exists(CoreExtension::class, 'testEmpty')) {
+                            $compiler->raw('(CoreExtension::testEmpty($_label_ = ');
+                        } else {
+                            $compiler->raw('(twig_test_empty($_label_ = ');
+                        }
+
                         $compiler->subcompile($label);
                         $compiler->raw(') ? [] : ["label" => $_label_])');
                     }
--- a/src/Symfony/Bundle/WebProfilerBundle/Twig/WebProfilerExtension.php
+++ b/src/Symfony/Bundle/WebProfilerBundle/Twig/WebProfilerExtension.php
@@ -14,6 +14,7 @@ namespace Symfony\Bundle\WebProfilerBundle\Twig;
 use Symfony\Component\VarDumper\Cloner\Data;
 use Symfony\Component\VarDumper\Dumper\HtmlDumper;
 use Twig\Environment;
+use Twig\Extension\EscaperExtension;
 use Twig\Extension\ProfilerExtension;
 use Twig\Profiler\Profile;
 use Twig\TwigFunction;
@@ -60,9 +61,6 @@ class WebProfilerExtension extends ProfilerExtension
         }
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function getFunctions(): array
     {
         return [
@@ -87,12 +85,12 @@ class WebProfilerExtension extends ProfilerExtension
 
     public function dumpLog(Environment $env, string $message, Data $context = null)
     {
-        $message = twig_escape_filter($env, $message);
+        $message = self::escape($env, $message);
         $message = preg_replace('/&quot;(.*?)&quot;/', '&quot;<b>$1</b>&quot;', $message);
 
         $replacements = [];
         foreach ($context ?? [] as $k => $v) {
-            $k = '{'.twig_escape_filter($env, $k).'}';
+            $k = '{'.self::escape($env, $k).'}';
             if (str_contains($message, $k)) {
                 $replacements[$k] = $v;
             }
@@ -109,11 +107,18 @@ class WebProfilerExtension extends ProfilerExtension
         return '<span class="dump-inline">'.strtr($message, $replacements).'</span>';
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function getName()
     {
         return 'profiler';
+    }
+
+    private static function escape(Environment $env, string $s): string
+    {
+        if (method_exists(EscaperExtension::class, 'escape')) {
+            return EscaperExtension::escape($env, $s);
+        }
+
+        // to be removed when support for Twig 3 is dropped
+        return twig_escape_filter($env, $s);
     }
 }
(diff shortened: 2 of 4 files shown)


src/Symfony/Bridge/Twig/Node/SearchAndRenderBlockNode.php
```diff
@@ -97,7 +98,7 @@ public function compile(Compiler $compiler): void
 
                         // Check at runtime whether the label is empty.
                         // If not, add it to the array at runtime.
-                        $compiler->raw('(twig_test_empty($_label_ = ');
+                        $compiler->raw('(FormExtension::testEmpty($_label_ = ');
```
GromNaN: I guess the FQCN should be used?

src/Symfony/Bridge/Twig/Extension/FormExtension.php
```diff
@@ -158,6 +158,26 @@ public function getFieldChoices(FormView $view): iterable
         yield from $this->createFieldChoicesList($view->vars['choices'], $view->vars['choice_translation_domain']);
     }
 
+    /**
+     * @internal
```
GromNaN: Maybe we can put a comment that this function has been copied from `twig_test_empty`, to help track changes.

OskarStark:
The same seems to be needed for `twig_escape_filter `

>  11x: Since twig/twig 3.9.0: Using the internal "twig_escape_filter" function is deprecated.
>    11x in WebProfilerExtensionTest::testDumpHeaderIsDisplayed from Symfony\Bundle\WebProfilerBundle\Tests\Twig

src/Symfony/Bridge/Twig/Extension/FormExtension.php
derrabus: Or maybe we use `CoreExtension::testEmpty()` if it exists?

GromNaN: [`CoreExtension::testEmpty()` is internal](https://github.com/twigphp/Twig/blob/3.x/src/Extension/CoreExtension.php#L1322), its signature could change, or it could be moved or renamed.

derrabus: Given that we intend to migrate Twig into a Symfony component, I think it's fine to access `@internal` functions.

stof: Even in Symfony components, it is not fine to use `@internal` APIs of a *different* component, as components relies on semver ranges in their inter-component requirements.

src/Symfony/Bridge/Twig/Extension/FormExtension.php
```diff
@@ -158,6 +158,26 @@ public function getFieldChoices(FormView $view): iterable
...
+     */
+    public static function testEmpty($value)
```
stof:
```suggestion
    public static function testEmpty($value): bool
```

src/Symfony/Bridge/Twig/Node/SearchAndRenderBlockNode.php
```diff
@@ -107,4 +113,24 @@ public function compile(Compiler $compiler): void
 
         $compiler->raw(')');
     }
+
+    /**
+     * @see \Twig\Extension\CoreExtension::testEmpty()
+     */
+    private static function isEmpty($value): bool
```
GromNaN: If we rely on the internal twig function for the compiled template, we can do the same for the compilation step and avoid duplicating the code.

But I'm still not sure this is a sane solution to rely on the internal function.

src/Symfony/Bridge/Twig/Node/SearchAndRenderBlockNode.php
```diff
@@ -97,7 +98,12 @@ public function compile(Compiler $compiler): void
 
                         // Check at runtime whether the label is empty.
                         // If not, add it to the array at runtime.
-                        $compiler->raw('(twig_test_empty($_label_ = ');
+                        if (method_exists(CoreExtension::class, 'testEmpty')) {
+                            $compiler->raw('(Twig\Extension\CoreExtension::testEmpty($_label_ = ');
```
fabpot: No need to use the fully qualified class name here: https://github.com/twigphp/Twig/blob/4.x/src/Node/ModuleNode.php#L146

src/Symfony/Bridge/Twig/Node/SearchAndRenderBlockNode.php
fabpot: I agree, let's use the Twig method instead.

src/Symfony/Bundle/WebProfilerBundle/Twig/WebProfilerExtension.php
```diff
@@ -116,4 +116,49 @@ public function getName()
     {
         return 'profiler';
     }
+
+    /**
+     * @see \Twig\Extension\EscaperExtension::escape()
+     */
+    private static function escape(string $s, string $charset): string
```
fabpot: I would also use the Twig internal method instead.

fabpot (changes requested):

src/Symfony/Bridge/Twig/Node/SearchAndRenderBlockNode.php
GromNaN: Maybe these methods shouldn't be `@internal` then?

src/Symfony/Bundle/WebProfilerBundle/Twig/WebProfilerExtension.php
xabbuh: done

src/Symfony/Bridge/Twig/Node/SearchAndRenderBlockNode.php
```diff
@@ -49,8 +50,14 @@ public function compile(Compiler $compiler): void
                         // of variables at compile time.
                         $labelIsExpression = false;
 
+                        if (method_exists(CoreExtension::class, 'testEmpty')) {
```
GromNaN:
```suggestion
                        // twig_test_empty is deprecated in Twig 3.9
                        if (method_exists(CoreExtension::class, 'testEmpty')) {
```
Adding such comment helps when we remove compatibility with older versions years after.

xabbuh: add comments (slightly reworded)

src/Symfony/Bridge/Twig/Node/SearchAndRenderBlockNode.php
```diff
@@ -49,8 +50,15 @@ public function compile(Compiler $compiler): void
...
+                            $isEmpty = CoreExtension::testEmpty($label->getAttribute('value'));
+                        } else {
+                            // to be removed when support for Twig 3 is dropped
+                            $isEmpty = twig_test_empty($label->getAttribute('value'));
+                        }
+
                         // Only insert the label into the array if it is not empty
-                        if (!twig_test_empty($label->getAttribute('value'))) {
+                        if (!$isEmpty) {
```
stof: Can we use `$label !== null && $label !== ''` as condition instead ? We don't need to handle cases about checking emptiness for other kind of values here.

xabbuh: done

fabpot (approved):

fabpot:
Thank you @xabbuh.

PR #52998, opened 2023-12-11
[Notifier] [Bridges] Provide EventDispatcher and HttpClient to the transport

### LINKED ISSUE #52997: [Notifier] Some Transport aren't using event dispatcher

5.4, 6.4, 7.x

### Description

Some Notifier transport factories are redefining the class constructor, but aren't passing the HttpClient nor the EventDispatcher.  
As a result, those transports are not providing Notifier event hooks.  
I think they should in order to act the same as other.
- MercureTransportFactory
- FakeSmsTransportFactory
- FakeChatTransportFactory

### How to reproduce

Use FakeSms bridge, and add an event listener on MessageEvent::class.

### Possible Solution

PR added

rdavaillaud:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52997 
| License       | MIT

Some Notifier transport factories are redefining the class constructor, but aren't passing the HttpClient nor the EventDispatcher.
As a result, those transports are not providing Notifier event hooks.

This PR adds those dependencies.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
@@ -2594,7 +2594,9 @@ class FrameworkExtension extends Extension
 
         if (ContainerBuilder::willBeAvailable('symfony/mercure-notifier', MercureTransportFactory::class, $parentPackages, true) && ContainerBuilder::willBeAvailable('symfony/mercure-bundle', MercureBundle::class, $parentPackages, true) && \in_array(MercureBundle::class, $container->getParameter('kernel.bundles'), true)) {
             $container->getDefinition($classToServices[MercureTransportFactory::class])
-                ->replaceArgument('$registry', new Reference(HubRegistry::class));
+                ->replaceArgument('$registry', new Reference(HubRegistry::class))
+                ->replaceArgument('$client', new Reference('http_client'))
+                ->replaceArgument('$dispatcher', new Reference('event_dispatcher'));
         } elseif (ContainerBuilder::willBeAvailable('symfony/mercure-notifier', MercureTransportFactory::class, $parentPackages, true)) {
             $container->removeDefinition($classToServices[MercureTransportFactory::class]);
         }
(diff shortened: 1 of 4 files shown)


OskarStark:
Hi, thanks for the PR, as this is a new feature, please target 7.1

rdavaillaud:
> Hi, thanks for the PR, as this is a new feature, please target 7.1

Is it really a new feature or an oversight? 
The [Notifier component documentation](https://symfony.com/doc/current/notifier.html#using-events) tells us that the transport use Event to interact in the lifecycle of the event

> The [Transport](https://github.com/symfony/symfony/blob/7.0/src/Symfony/Component/Notifier/Transport.php) class of the Notifier component allows you to optionally hook into the lifecycle via events.

Currently, the hooks aren't working with the FakeSms/FakeChat and Mercure as they should.

OskarStark:
Not sure, from a technical POV it is a new feature, because they were never meant to dispatch those events, but I can understand your point.

cc @nicolas-grekas

rdavaillaud:
I will give some argument in favour of a bugfix :grin: 

Those transports (at least the Fake* ones) are intended to be used in dev/test, but we cannot test the events behaviours, and it is not documented that they don't support those events.

Technically, the transports class **does** have everything to support the dispatch of those events, they correctly extends the AbstractTransport class and call the extended class constructor passing the two HttpClient and EventDispatcher argument.

The factories, in the other hand, don't handle the extended AbstractTransportFactory class constructor arguments. As all the other factories don't override the constructor, there is no precedent.  
As those 3 specific ones need to override the constructor, I see no point not to pass them.

@fabpot , as you are the designer of the component and in order to determine if this is bug since 5.3 or a new feature for 7.x., do you have an opinion on that : does every TransportFactory need to pass EventDispatcher and HttpClient ? 

I will target whichever you finally decide, tell me.

nicolas-grekas (commented):
Works as a bugfix to me. It's surprising that we'd rely on the default http client on purpose.
Can you also remove the use of named arguments? We don't use them in core except here and it looks like an oversight.
Can you please check on 6.4/7.1 if there are more bridges that would need a similar change and submit a PR if yes?

fabpot (approved):

fabpot:
Thank you @rdavaillaud.

rdavaillaud:
> Can you also remove the use of named arguments? We don't use them in core except here and it looks like an oversight.

I will try to see what I can do with this, I need to dig in and understand why. I would open a PR then.

> Can you please check on 6.4/7.1 if there are more bridges that would need a similar change and submit a PR if yes?

From what I can see, those three bridges are the only ones.

OskarStark:
Behause we missed to add a conflict in FrameworkBundle 6.4.2 to require fake-chat and fake-sms at least in 6.4.2 ends up in the following error:
```
Invalid service "notifier.transport_factory.fake-sms": method "Symfony\Component\Notifier\Bridge\FakeSms\FakeSmsTransportFactory::__construct()" has no argument named "$client". Check your service definition.
```

The solution is to upgrade the notifier bridges to `6.4.2` first

cc @xabbuh

xabbuh:
The same issue should be observable too when updating FrameworkBundle from 5.4.33 to 5.4.34, 6.3.10 to 6.3.11 or from 7.0.1 to 7.0.2 without also upgrading these bridges.

rdavaillaud:
Damn, I am not used to those dependencies conflicts, sorry. What are the rules I missed?
Can I do something to mitigate the issue ?

I suppose removing the use of named arguments as requested by @nicolas-grekas is the way to go, but I didn't worked on this until now. And this will maybe need some other things to do in composer that I don't know of.

@OskarStark, as you are the first contributor on these bridges, can you explain to me why was it necessary at the first place to put those lines in Framework Bundle? This may put me in the fast lane to understand.

nicolas-grekas:
Removing named arguments is the way to go, no need for anything else (no conflicts).

xabbuh:
see #53341

PR #53006, opened 2023-12-12
[ErrorHandler] Don't format binary strings

### LINKED ISSUE #53005: [ErrorHandler] Formatting binary strings causes memory exhausion

5.4

### Description

If the trace content contains binary strings, e.g. images, `var_dump` leads to memory exhaustion (`Allowed memory size of X bytes exhausted`).

### How to reproduce

Cause an error with a code path that passes a longer binary string to a method.

### Possible Solution

Check for non-unicode characters in a string and skip adding it to the formatted output.

### Additional Context

_No response_

aleho:
Fixes #53005

| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix  #53005
| License       | MIT

This is an attempt to fix memory exhaustion when a binary string is part of a stack trace.


derrabus:
Please note that we usually don't merge changes without a test that covers them.

aleho:
I can try to add a test case for that.

Would loading a local, binary file into a string in a test-case be OK?

(Asserting memory exhaustion in `var_dump` to me seems a bit hard to test ;))

aleho:
I just had a look at the tests and figured it might indeed be hard to test for `var_export` not being called with a binary string. I hope just a hard-coded check for the replacement string is OK.

src/Symfony/Bridge/Twig/Extension/CodeExtension.php
```diff
@@ -98,7 +98,11 @@ public function formatArgs(array $args): string
             } elseif ('resource' === $item[0]) {
                 $formattedValue = '<em>resource</em>';
             } else {
-                $formattedValue = str_replace("\n", '', htmlspecialchars(var_export($item[1], true), \ENT_COMPAT | \ENT_SUBSTITUTE, $this->charset));
+                if (preg_match('//u', $item[1])) {
```
nicolas-grekas: when ppl don't use the UTF-8 charset, this might lead to false positives
what about checking using `preg_match('/[^\x00\x07-\x0d\x1B\x20-\xff]/', $value)` instead?

also, please unindent the if else by merging it with the surrouding one

aleho: > also, please unindent the if else by merging it with the surrouding one

The parent `if` checks the value of `$item[0]` exclusively. The merged `else-if` would then check `$item[1]`. I explicitly wanted to reduce cognitive load here (having X-branches of an `if` checking the same variable only to have the last one do something different is not very readable in my opinion). If you insist I'll still merge the checks though.

derrabus: > If you insist I'll still merge the checks though.

We do. It's our code style.

src/Symfony/Component/ErrorHandler/Tests/ErrorRenderer/HtmlErrorRendererTest.php
```diff
@@ -54,4 +54,57 @@ public static function getRenderData(): iterable
...
+        }
+
+        $exception = null;
+        try {
+            $binaryData = file_get_contents(__DIR__.'/../Fixtures/pixel.png');
+            $this->throwRuntimeException($binaryData);
+        } catch (\RuntimeException $exception) {
+        } finally {
+            unset($binaryData);
+            $this->assertNotNull($exception, 'Could not catch RuntimeException to check for binary string replacement');
```
derrabus: Does this have to be inside a `finally` block?

nicolas-grekas (approved):
I applied my patch from https://github.com/symfony/symfony/issues/53005#issuecomment-1872161157

Good catch on the ini setting.

nicolas-grekas:
Thank you @aleho.

aleho:
> I applied my patch from [#53005 (comment)](https://github.com/symfony/symfony/issues/53005#issuecomment-1872161157)

FYI, originally I wanted to throw the exception to simulate the conditions where I ran into the problem. Finding out the ini setting is what made the difference I guess your changes make sense.

PR #53056, opened 2023-12-13
[Serializer] Add `DateTimeNormalizer::CAST_KEY` context option

### LINKED ISSUE #53055: [Serializer] DateTimeNormalizer U or U.u format as int/float
### Description

Currently DateTimeNormalizer always uses `datetime->format($format)` so it's logical when expecting human readable dates to get as strings, but one could expect that when providing "U" format to get an integer or "U.u" to get float.

In https://github.com/meilisearch/meilisearch-symfony/issues/304 we have a need for this, because meilisearch api expects dates to be indexed as integers.

So I think this could be a good addition, but of course for BC reasons we'd need to add additional context key to enable this.

### Example

_No response_

norkunas:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #53055
| License       | MIT

This adds `DateTimeNormalizer::CAST_KEY` context option allowing to cast the formatted result to `int` or `float`.

### DIFF
--- a/src/Symfony/Component/Serializer/Context/Normalizer/DateTimeNormalizerContextBuilder.php
+++ b/src/Symfony/Component/Serializer/Context/Normalizer/DateTimeNormalizerContextBuilder.php
@@ -61,4 +61,12 @@ final class DateTimeNormalizerContextBuilder implements ContextBuilderInterface
 
         return $this->with(DateTimeNormalizer::TIMEZONE_KEY, $timezone);
     }
+
+    /**
+     * @param 'int'|'float'|null $cast
+     */
+    public function withCast(?string $cast): static
+    {
+        return $this->with(DateTimeNormalizer::CAST_KEY, $cast);
+    }
 }
--- a/src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
@@ -25,10 +25,12 @@ final class DateTimeNormalizer implements NormalizerInterface, DenormalizerInter
 {
     public const FORMAT_KEY = 'datetime_format';
     public const TIMEZONE_KEY = 'datetime_timezone';
+    public const CAST_KEY = 'datetime_cast';
 
     private array $defaultContext = [
         self::FORMAT_KEY => \DateTimeInterface::RFC3339,
         self::TIMEZONE_KEY => null,
+        self::CAST_KEY => null,
     ];
 
     private const SUPPORTED_TYPES = [
@@ -59,7 +61,7 @@ final class DateTimeNormalizer implements NormalizerInterface, DenormalizerInter
     /**
      * @throws InvalidArgumentException
      */
-    public function normalize(mixed $object, ?string $format = null, array $context = []): string
+    public function normalize(mixed $object, ?string $format = null, array $context = []): int|float|string
     {
         if (!$object instanceof \DateTimeInterface) {
             throw new InvalidArgumentException('The object must implement the "\DateTimeInterface".');
@@ -73,7 +75,11 @@ final class DateTimeNormalizer implements NormalizerInterface, DenormalizerInter
             $object = $object->setTimezone($timezone);
         }
 
-        return $object->format($dateTimeFormat);
+        return match ($context[self::CAST_KEY] ?? $this->defaultContext[self::CAST_KEY] ?? false) {
+            'int' => (int) $object->format($dateTimeFormat),
+            'float' => (float) $object->format($dateTimeFormat),
+            default => $object->format($dateTimeFormat),
+        };
     }
 
     public function supportsNormalization(mixed $data, ?string $format = null, array $context = []): bool
--- a/src/Symfony/Component/Serializer/CHANGELOG.md
+++ b/src/Symfony/Component/Serializer/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Add `DateTimeNormalizer::CAST_KEY` context option
+
 7.0
 ---
 
--- a/src/Symfony/Component/Serializer/Tests/Context/Normalizer/DateTimeNormalizerContextBuilderTest.php
+++ b/src/Symfony/Component/Serializer/Tests/Context/Normalizer/DateTimeNormalizerContextBuilderTest.php
@@ -38,6 +38,7 @@ class DateTimeNormalizerContextBuilderTest extends TestCase
         $context = $this->contextBuilder
             ->withFormat($values[DateTimeNormalizer::FORMAT_KEY])
             ->withTimezone($values[DateTimeNormalizer::TIMEZONE_KEY])
+            ->withCast($values[DateTimeNormalizer::CAST_KEY])
             ->toArray();
 
         $this->assertEquals($values, $context);
@@ -51,11 +52,13 @@ class DateTimeNormalizerContextBuilderTest extends TestCase
         yield 'With values' => [[
             DateTimeNormalizer::FORMAT_KEY => 'format',
             DateTimeNormalizer::TIMEZONE_KEY => new \DateTimeZone('GMT'),
+            DateTimeNormalizer::CAST_KEY => 'int',
         ]];
 
         yield 'With null values' => [[
             DateTimeNormalizer::FORMAT_KEY => null,
             DateTimeNormalizer::TIMEZONE_KEY => null,
+            DateTimeNormalizer::CAST_KEY => null,
         ]];
     }
 
--- a/src/Symfony/Component/Serializer/Tests/Normalizer/DateTimeNormalizerTest.php
+++ b/src/Symfony/Component/Serializer/Tests/Normalizer/DateTimeNormalizerTest.php
@@ -154,6 +154,72 @@ class DateTimeNormalizerTest extends TestCase
         ];
     }
 
+    /**
+     * @dataProvider provideNormalizeUsingCastCases
+     */
+    public function testNormalizeUsingCastPassedInConstructor(\DateTimeInterface $value, string $format, ?string $cast, string|int|float $expectedResult)
+    {
+        $normalizer = new DateTimeNormalizer([DateTimeNormalizer::CAST_KEY => $cast]);
+
+        $this->assertSame($normalizer->normalize($value, null, [DateTimeNormalizer::FORMAT_KEY => $format]), $expectedResult);
+    }
+
+    /**
+     * @dataProvider provideNormalizeUsingCastCases
+     */
+    public function testNormalizeUsingCastPassedInContext(\DateTimeInterface $value, string $format, ?string $cast, string|int|float $expectedResult)
+    {
+        $this->assertSame($this->normalizer->normalize($value, null, [DateTimeNormalizer::FORMAT_KEY => $format, DateTimeNormalizer::CAST_KEY => $cast]), $expectedResult);
+    }
+
+    /**
+     * @return iterable<array{0: \DateTimeImmutable, 1: non-empty-string, 2: 'int'|'float'|null, 3: 'int'|'float'|'string'}>
+     */
+    public static function provideNormalizeUsingCastCases(): iterable
+    {
+        yield [
+            \DateTimeImmutable::createFromFormat('U', '1703071202'),
+            'Y',
+            null,
+            '2023',
+        ];
+
+        yield [
+            \DateTimeImmutable::createFromFormat('U', '1703071202'),
+            'Y',
+            'int',
+            2023,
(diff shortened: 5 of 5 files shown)


src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
```diff
@@ -73,7 +75,16 @@ public function normalize(mixed $object, string $format = null, array $context =
             $object = $object->setTimezone($timezone);
         }
 
-        return $object->format($dateTimeFormat);
+        return match ($context[self::TIMESTAMP_CAST_KEY] ?? $this->defaultContext[self::TIMESTAMP_CAST_KEY]) {
```
mtarld: It could be great to have this by default in 8.0, so what do you think about:
- not setting the default context value
- trigger a deprecation if the context value is not set, saying that it'll default to true in the next major version

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
```diff
@@ -73,7 +75,16 @@ public function normalize(mixed $object, string $format = null, array $context =
...
+            true => match ($dateTimeFormat) {
+                'U' => $object->getTimestamp(),
+                'U.u' => (float) $object->format('U.u'),
+                'Uv' => (int) $object->format('Uv'),
+                'Uu' => (int) $object->format('Uu'),
```
mtarld: Should these formats be handled in the denormalization as well?

src/Symfony/Component/Serializer/Context/Normalizer/DateTimeNormalizerContextBuilder.php
```diff
@@ -61,4 +61,9 @@ public function withTimezone(\DateTimeZone|string|null $timezone): static
 
         return $this->with(DateTimeNormalizer::TIMEZONE_KEY, $timezone);
     }
+
+    public function withTimestampCast(bool $cast): static
+    {
+        return $this->with(DateTimeNormalizer::TIMESTAMP_CAST_KEY, $cast);
```
mtarld: Tests must be updated as well

src/Symfony/Component/Serializer/Context/Normalizer/DateTimeNormalizerContextBuilder.php
```diff
@@ -61,4 +61,9 @@ public function withTimezone(\DateTimeZone|string|null $timezone): static
 
         return $this->with(DateTimeNormalizer::TIMEZONE_KEY, $timezone);
     }
+
+    public function withTimestampCast(bool $cast): static
```
mtarld:
```suggestion
    public function withTimestampCast(bool $cast = true): static
```

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
mtarld: Instead of nesting matches, I'd do:
```php
if ($context[self::TIMESTAMP_CAST_KEY] ?? $this->defaultContext[self::TIMESTAMP_CAST_KEY]) {
    return match ($dateTimeFormat) {
        'U' => $object->getTimestamp(),
        'U.u' => (float) $object->format('U.u'),
        'Uv' => (int) $object->format('Uv'),
        'Uu' => (int) $object->format('Uu'),
       default => $object->format($dateTimeFormat),
    }
}

return $object->format($dateTimeFormat);
```

mtarld (changes requested):

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
norkunas: 👍 will do :)

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
norkunas: In denormalization `U.u` or `U.v` is OK, but not `Uu` or `Uv`

norkunas:
@mtarld I've updated, is this what you had in mind with deprecation?

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
```diff
@@ -40,6 +42,10 @@ final class DateTimeNormalizer implements NormalizerInterface, DenormalizerInter
     public function __construct(array $defaultContext = [])
     {
         $this->setDefaultContext($defaultContext);
+
+        if (!isset($this->defaultContext[self::TIMESTAMP_CAST_KEY])) {
```
mtarld: I won't have put that in the constructor but rather in the `normalize` method, to avoid the trigger occurring too often.

Plus, we will have to create a PR to update https://github.com/symfony/recipes so that the new projects won't have any deprecations.

mtarld (changes requested):

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
norkunas: Ah,okay

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
valtzu: I think we should also support `Uu` and `Uv` in denormalization. I know `DateTime::createFromFormat` does not support it, but it does not mean Symfony can't support it.

This is not really related to the casting only, but to denormalizing those formats in general. I'm happy to submit a PR to add those formats after this one is merged if we consider those being out-of-scope here.

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
norkunas: updated and created recipe in https://github.com/symfony/recipes/pull/1276

norkunas:
Should I skip those test cases on 32bit systems when value is > PHP_INT_MAX?

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
norkunas: I could add them, but then we need to handle the errors our-self

norkunas: But given the reasons in https://github.com/php/php-src/issues/12981 I'd say we should not try to support these formats then.

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
```diff
@@ -59,12 +61,16 @@ public function getSupportedTypes(?string $format): array
...
-    public function normalize(mixed $object, string $format = null, array $context = []): string
+    public function normalize(mixed $object, string $format = null, array $context = []): int|float|string
     {
         if (!$object instanceof \DateTimeInterface) {
             throw new InvalidArgumentException('The object must implement the "\DateTimeInterface".');
         }
 
+        if (!isset($this->defaultContext[self::TIMESTAMP_CAST_KEY])) {
+            trigger_deprecation('symfony/serializer', '7.1', 'Not setting the "%s" in defaultContext for "%s" is deprecated. It will default to "true" in 8.0.', self::TIMESTAMP_CAST_KEY, self::class);
+        }
```
derrabus: I don't think we should deprecate not setting the new key. We're not changing the default behavior anyway, are we?

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
valtzu: PHP's internal implementation does greedy match from left to right, but I don't think we should be limited by that. Milliseconds as a timestamp are super common (f.e. in Javascript, Kafka timestamps etc.) so I think it'd be great not to require custom logic to support denormalizing them.


> but then we need to handle the errors our-self

We could just do `DateTime::createFromFormat('U.v', sprintf('%.3f', $timestamp/1000))` when `Uv` format is given – not sure if we even need any extra error handling for that if it's already checked `$timestamp` is an integer.

---

Anyway, I may post a separate PR on that so we can get more opinions and not block this MR which addresses different issue.

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
```diff
@@ -73,6 +79,16 @@ public function normalize(mixed $object, string $format = null, array $context =
             $object = $object->setTimezone($timezone);
         }
 
+        if ($context[self::TIMESTAMP_CAST_KEY] ?? $this->defaultContext[self::TIMESTAMP_CAST_KEY] ?? false) {
+            return match ($dateTimeFormat) {
+                'U' => $object->getTimestamp(),
+                'U.u' => (float) $object->format('U.u'),
+                'Uv' => (int) $object->format('Uv'),
+                'Uu' => (int) $object->format('Uu'),
```
derrabus: I find those values super obscure, I must admit. Knowing where they're coming from, they might even create the impression that we accept any format modifier here which is clearly not the case.

Maybe an enum would be the better choice?

derrabus:
> Should I skip those test cases on 32bit systems when value is > PHP_INT_MAX?

Either that or skip the int cast on those systems.

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
norkunas: This was requested by @mtarld, so not sure then :)

mtarld: Indeed, as I've written in this comment https://github.com/symfony/symfony/pull/53056#discussion_r1426664940, I think it could be great to cast timestamps by default in 8.0. So, we need to set the default context value to `true` in 8.0 (and therefore it changes the 7.x default behavior)

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
mtarld: Agreed, or maybe we can leverage private constants here?

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
norkunas: I've probably targeted wrong version for recipe?

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
norkunas: I don't see any benefits, but well you decide and I'll update

derrabus: On second thought, forget my enum suggestion. I think, my main problem is the guesswork that is encoded in this match block. We basically add an option to cast the result, but we don't say to what we want to cast, so our code has to guess.

What if my app decides to opt in to casting and uses `'Y'` as formatter. From a user perspective, wouldn't I expect an integer cast here as well? I think, we will never be able to completely enumerate all the cases in which a cast makes sense.

So, maybe instead of making the cast option a boolean, we rather accept a type identifier instead?

```php
$this->normalizer->normalize($date, 'json', [
    DateTimeNormalizer::CAST_KEY => 'int',
    DateTimeNormalizer::FORMAT_KEY => 'Y',
]);
```

This would eliminate the need for this guesswork, but it would also mean that the type cast will always remain opt-in.

norkunas: Given that the new option suggests about **timestamp casting only**, I'm not sure why you'd decide to try in your app same thing for the `Y` format. But okay, I see your point, that this would broaden the scope of this feature and allow people for greater usages, so works for me.

norkunas:
Anybody to review?

mtarld (commented):
Code looks good to me. But I'm wondering that maybe we can add a check to make sure that the cast value is consistent with the format? So that it won't be possible to cast `d/m/Y` to an `int` for example. WDYT?

norkunas:
> Code looks good to me. But I'm wondering that maybe we can add a check to make sure that the cast value is consistent with the format? So that it won't be possible to cast `d/m/Y` to an `int` for example. WDYT?

But then we would go back to format guessing, which derrabus wanted to avoid :)

mtarld:
Oh yeah, I missed that part... Thanks, LGTM then!

mtarld (approved):

fabpot (approved):

fabpot:
Thank you @norkunas.

PR #53057, opened 2023-12-13
[HttpKernel] Move `@internal` from `AbstractSessionListener` class to its methods and properties

Florian-Merle:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        |
| License       | MIT

The `AbstractSessionListener` being marked as internal, its public constant `NO_AUTO_CACHE_CONTROL_HEADER` should not be used while the documentation [states it can](https://symfony.com/doc/current/http_cache.html#http-caching-and-user-sessions).

In fact, static analysis tools like psalm says there is an error with code using this constant.
```
ERROR: InternalClass - xxx.php:32:33 - Symfony\Component\HttpKernel\EventListener\AbstractSessionListener is internal to Symfony but called from xxx (see https://psalm.dev/174)
        $response->headers->set(AbstractSessionListener::NO_AUTO_CACHE_CONTROL_HEADER, 'true');
```

~~Another solution is to make every method of the `AbstractSessionListener`  internal but this means the class could be extended.~~
~~Also, maybe the class `AbstractSessionListener` should not be internal, but I don't think so.~~
~~This is why I introduced a new interface that is not internal and allows to not introduce BC.~~

~~The documentation will need to be updated if this pull request is merged, I'd be happy to do it later.~~

As discussed, I made public/protected properties and methods internal and removed the original internal mark on the class.
This solves the issue and allows us to use the `AbstractSessionListener::NO_AUTO_CACHE_CONTROL_HEADER` const just like the documentation says we can.

### DIFF
--- a/src/Symfony/Component/HttpKernel/EventListener/AbstractSessionListener.php
+++ b/src/Symfony/Component/HttpKernel/EventListener/AbstractSessionListener.php
@@ -36,13 +36,15 @@ use Symfony\Contracts\Service\ResetInterface;
  *
  * @author Johannes M. Schmitt <schmittjoh@gmail.com>
  * @author Tobias Schultze <http://tobion.de>
- *
- * @internal
  */
 abstract class AbstractSessionListener implements EventSubscriberInterface, ResetInterface
 {
     public const NO_AUTO_CACHE_CONTROL_HEADER = 'Symfony-Session-NoAutoCacheControl';
 
+
+    /**
+     * @internal
+     */
     protected $container;
     private $sessionUsageStack = [];
     private $debug;
@@ -52,6 +54,9 @@ abstract class AbstractSessionListener implements EventSubscriberInterface, Rese
      */
     private $sessionOptions;
 
+    /**
+     * @internal
+     */
     public function __construct(ContainerInterface $container = null, bool $debug = false, array $sessionOptions = [])
     {
         $this->container = $container;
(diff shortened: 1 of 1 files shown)


chalasr:
I suggest to ~~make `AbstractSessionListener` `@final` instead of `@internal`~~ remove the `@internal` flag from the abstract class **and** mark all of its methods as `@internal` instead, so we don't have to introduce an interface for a single constant and this PR can target 5.4 as a bugfix.

stof:
an abstract class being defined as `@final` does not make any sense.

chalasr:
Right, stoffed again 😅 
Proposal updated to just mark its methods as internal.

src/Symfony/Component/HttpKernel/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ CHANGELOG
  * Deprecate the `fileLinkFormat` parameter of `DebugHandlersListener`
  * Add support for configuring log level, and status code by exception class
  * Allow ignoring "kernel.reset" methods that don't exist with "on_invalid" attribute
+ * Un-mark `AbstractSessionListener` as internal and mark all of its public/protected method/properties as internal
```
chalasr: Should be reverted as changelogs are auto-generated for bugfix releases

chalasr (approved):
with minor comment

fabpot (approved):

fabpot:
Thank you @Florian-Merle.

PR #52962, opened 2023-12-09
[FrameworkBundle] Move Router cache directory to `kernel.build_dir`

okhoshi:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | none
| License       | MIT

Follow up to #50391, set up Router cache directory to `kernel.build_dir` instead of `kernel.cache_dir` by default, and only warm the cache on read-only resources phase.

#SymfonyHackday

PR #53059, opened 2023-12-13
[FrameworkBundle] Add `router.ignore_cache` config option

okhoshi:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | none
| License       | MIT

In #52962, the `router.cache_dir` is marked as deprecated, but no replacement was given for the specific case when one wants to completely disable the router cache (use to be done by setting `router.cache_dir` to `~` or `null`).

This MR introduces `router.ignore_cache` option to restore that possibility, while still deprecating the choice of cache location.


OskarStark:
@nicolas-grekas I accidentaly merged https://github.com/symfony/symfony-docs/pull/19399

Shall I revert or is PR ready to merge from your side?

Thanks

cc @xabbuh

src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ CHANGELOG
  * Add `secrets:reveal` command
  * Add `rate_limiter` option to `http_client.default_options` and `http_client.scoped_clients`
  * Attach the workflow's configuration to the `workflow` tag
+ * Add `router.ignore_cache` config option
```
nicolas-grekas, line 16: no_cache?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -619,8 +619,9 @@ private function addRouterSection(ArrayNodeDefinition $rootNode): void
                         ->scalarNode('type')->end()
                         ->scalarNode('cache_dir')
                             ->defaultValue('%kernel.build_dir%')
-                            ->setDeprecated('symfony/framework-bundle', '7.1', 'Setting the "%path%.%node%" configuration option is deprecated. It will be removed in version 8.0.')
+                            ->setDeprecated('symfony/framework-bundle', '7.1', 'Setting the "%path%.%node%" configuration option is deprecated. It will be removed in version 8.0. Please use the "ignore_cache" option if you want to ignore the cache in the router.')
```
nicolas-grekas, line 622:
```suggestion
                            ->setDeprecated('symfony/framework-bundle', '7.1', 'Setting the "%path%.%node%" configuration option is deprecated. It will be removed in version 8.0. Use the "no_cache" option if you want to disable the router cache.')
```

nicolas-grekas (commented):
Why would anyone want to disable the router cache?

okhoshi:
> Why would anyone want to disable the router cache?

I don't know, but it was a supported scenario until cache option was deprecated. If you think it is not useful, I can also deprecate that feature as well.

nicolas-grekas:
It is already deprecated, isn't it? I'm calling for use cases that need to disable the cache. We can merge this PR in a later version if people report that they cannot work around the deprecation while they need to disable the cache.

okhoshi:
Then I'll remove the part regarding this option in the doc.

Should I close this PR now ? I can still reopen it or redo the code changes if needed in the future.

nicolas-grekas:
Yep, let's close and wait for feedbacks.

mpdude:
👋🏼 It's me, I want to disable the cache and I added the initial support for this configuration setting (explicitly mentioning the `~`/`null` use case).

The reason is that I have test suites where different sets of routes need to be loaded. IIRC, the router keeps the routes in a `private static` property, and so tests fail once one set of routes has been loaded. Disabling the cache dir was one way to prevent this caching.

What can I do to keep this possibility open?

PR #53060, opened 2023-12-13
[Uid] Add `UuidV1::toV6()`, `UuidV1::toV7()` and `UuidV6::toV7()`

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | false
| Issues        | -
| License       | MIT

The goal is to help working with new UUIDv6 & v7 versions when you receive legacy UUIDv1.

UUIDv6 and v1 are 100% compatible since v6 is just a "reordering" of v1.

UUIDv7 and v1 are not strictly compatible for several reasons: the timestamp precision is not the same. Also, there is no clock seq or variant or node in v7 but full randomness instead. And maybe others?
However, I believe we can still technically do the conversion (timestamp -> timestamp, clock seq + variant + node -> randomness) with at least one concession: the sub-milliseconds v1 timestamp precision is lost since v7 doesn't support it. 

My proposed v1 to v7 implementation has 2 issues:
* sub-microsecond entropy is entirely lost
* monotonicity is lost?

I'm opening this PR so that @nicolas-grekas gets triggered by the inefficiency of my code and finds a better solution :grin:
I'm kidding, I actually know he already has an idea because he explained it to me during the Brussels HackDay but I couldn't understand it well enough to be able to implement it :sweat_smile:

### DIFF
--- a/src/Symfony/Component/Uid/Tests/UuidTest.php
+++ b/src/Symfony/Component/Uid/Tests/UuidTest.php
@@ -428,32 +428,47 @@ class UuidTest extends TestCase
         $this->assertInstanceOf(Uuid::class, Uuid::fromString('111111111u9QRyVM94rdmZ'));
     }
 
-    public function testV6FromV1()
+    public function testV1ToV6()
     {
         $uuidV1 = new UuidV1('8189d3de-9670-11ee-b9d1-0242ac120002');
-        $uuidV6 = UuidV6::fromV1($uuidV1);
+        $uuidV6 = $uuidV1->toV6();
 
         $this->assertEquals($uuidV1->getDateTime(), $uuidV6->getDateTime());
         $this->assertSame($uuidV1->getNode(), $uuidV6->getNode());
-        $this->assertEquals($uuidV6, UuidV6::fromV1($uuidV1));
+        $this->assertEquals($uuidV6, $uuidV1->toV6());
     }
 
-    public function testV7FromV1BeforeUnixEpochThrows()
+    public function testV1ToV7BeforeUnixEpochThrows()
     {
         $this->expectException(\InvalidArgumentException::class);
-        $this->expectExceptionMessage('UUIDv1 with a timestamp before Unix epoch cannot be converted to UUIDv7');
+        $this->expectExceptionMessage('Cannot convert UUID to v7: its timestamp is before the Unix epoch.');
 
-        UuidV7::fromV1(new UuidV1('9aba8000-ff00-11b0-b3db-3b3fc83afdfc')); // Timestamp is 1969-01-01 00:00:00.0000000
+        (new UuidV1('9aba8000-ff00-11b0-b3db-3b3fc83afdfc'))->toV7(); // Timestamp is 1969-01-01 00:00:00.0000000
     }
 
-    public function testV7FromV1()
+    public function testV1ToV7()
     {
         $uuidV1 = new UuidV1('eb248d80-ea4f-11ec-9d2a-839425e6fb88');
         $sameUuidV1100NanosecondsLater = new UuidV1('eb248d81-ea4f-11ec-9d2a-839425e6fb88');
-        $uuidV7 = UuidV7::fromV1($uuidV1);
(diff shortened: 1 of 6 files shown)


src/Symfony/Component/Uid/Tests/UuidTest.php
```diff
@@ -427,4 +427,33 @@ public function testFromStringBase58Padding()
...
+
+    public function testV7FromV1()
+    {
+        $uuidV1 = new UuidV1('eb248d80-ea4f-11ec-9d2a-839425e6fb88');
+        $sameUuidV1100NanosecondsLater = new UuidV1('eb248d81-ea4f-11ec-9d2a-839425e6fb88');
+        $uuidV7 = UuidV7::fromV1($uuidV1);
+
+        $this->assertEquals($uuidV1->getDateTime(), $uuidV7->getDateTime());
+        $this->assertEquals($uuidV7, UuidV7::fromV1($uuidV1));
+        $this->assertNotEquals($uuidV7, UuidV7::fromV1($sameUuidV1100NanosecondsLater));
```
fancyweb: This test fails because both v1 give the same v7 (sub-microsecond entropy is lost).

stof: sub-microsecond entropy should also become part of the randomness part IMO.

nicolas-grekas (commented):
Thanks for pushing this :)

I added a second commit where I renamed `from*()` methods to `to*()`
and I made the resulting UUIDv7 monotonic within the same 100-ns (modulo the clock-seq)

src/Symfony/Component/Uid/UuidV6.php
```diff
@@ -43,6 +43,33 @@ public function getNode(): string
...
+        $time = BinaryUtil::hexToNumericString('0'.substr($uuid, 0, 8).substr($uuid, 9, 4).substr($uuid, 15, 3));
+        if ('-' === $time[0]) {
+            throw new \InvalidArgumentException('Cannot convert UUID to v7: its timestamp is before the Unix epoch.');
+        }
+
+        $ms = \strlen($time) > 4 ? substr($time, 0, -4) : '0';
+        $time = dechex(10000 * hexdec(substr($uuid, 20, 3)) + substr($time, -4));
+
+        if (\strlen($time) > 6) {
+            $uuid[29] = $uuid[29] ^ chr($time[0]);
```
fancyweb: There's an "overflow" here, for example, if `$uuid[29] === 'f'` and `$time[0] === '1'`, xoring them gives `g` which is invalid. I added a failing test for it :smiley:

fancyweb (commented):
Thank you very much for looking at this Nicolas :goat:

It looks like we are close to the perfect solution thanks to you. However, after reviewing it and testing it I've found that the xor doesn't always produce valid UUIDs (see failing test).

Also, what do you think of going directly from v1 to v7 to avoid the unneeded v6 step? I'm actually not even sure v6 -> v7 is needed at all since they are part of the same RFC iteration: I don't see any case where a system would be able to generate v6 but not v7 directly? :thinking:

src/Symfony/Component/Uid/UuidV6.php
nicolas-grekas: indeed, fixed!

nicolas-grekas:
> Also, what do you think of going directly from v1 to v7 to avoid the unneeded v6 step? I'm actually not even sure v6 -> v7 is needed at all since they are part of the same RFC iteration: I don't see any case where a system would be able to generate v6 but not v7 directly? 🤔

It doesn't cost much to provide v6 to v7 so why skip it?

fancyweb:
> I don't cost much to provide v6 to v7 so why skip it?

Because nobody will use it :sweat_smile:

I reviewed it again and it's OK for me. I can't approve it since it's my own PR. I force-pushed again to add more assertions.

chalasr (approved):

nicolas-grekas:
Thank you @fancyweb.

PR #52560, opened 2023-11-12
[Mailer] Update default Mailjet port

Katario:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

Similar to [#49805](https://github.com/symfony/symfony/pull/49805), update the default to 587, which is recommended by Mailjet.
Changelog is also updated accordingly.

References:

> Nowadays, port 587 is used for secure submission of email for delivery. Most of the client software are configured to use this port to send your messages.
[Mailjet Blog Post about this](https://www.mailjet.com/blog/email-best-practices/which-smtp-port-mailjet/).

PR #53061, opened 2023-12-13
[Mailer] [Mailjet] Disable tls for mailjet as it should use STARTTLS

RFreij:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Issues        |  
| License       | MIT

#52560 updates the default mailjet port and references the same change for mailgun.

Similar to #51042 and with the same error message, the mailjet connection doesn't work now. 

The PR above already has a reference, but for completion sake: Ref https://github.com/symfony/symfony/issues/34846#issuecomment-580190302 for more explanation


nicolas-grekas:
Thank you @RFreij.

PR #53073, opened 2023-12-14
Set `strict` parameter of `in_array` to true where possible

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

This pull request introduces a comprehensive update where the `in_array` function is now consistently invoked with the `strict` parameter set to `true`. This change is implemented across various components of the framework to enhance type safety and reliability.

To me, this update is a step towards more explicit and error-resistant code.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Command/TranslationDebugCommand.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Command/TranslationDebugCommand.php
@@ -223,8 +223,8 @@ EOF
                     }
                 }
 
-                if (!\in_array(self::MESSAGE_UNUSED, $states) && $input->getOption('only-unused')
-                    || !\in_array(self::MESSAGE_MISSING, $states) && $input->getOption('only-missing')
+                if (!\in_array(self::MESSAGE_UNUSED, $states, true) && $input->getOption('only-unused')
+                    || !\in_array(self::MESSAGE_MISSING, $states, true) && $input->getOption('only-missing')
                 ) {
                     continue;
                 }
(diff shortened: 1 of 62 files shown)


src/Symfony/Bundle/FrameworkBundle/Command/TranslationDebugCommand.php
```diff
@@ -223,9 +223,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                     }
                 }
 
-                if (!\in_array(self::MESSAGE_UNUSED, $states) && $input->getOption('only-unused')
-                    || !\in_array(self::MESSAGE_MISSING, $states) && $input->getOption('only-missing')
-                ) {
+                if (!\in_array(self::MESSAGE_UNUSED, $states, true) && ($input->getOption('only-unused') || $input->getOption('only-missing'))) {
```
GromNaN: You changed the condition, was it on-purpose?

Adding the strict param would be only:
```suggestion
                if (!\in_array(self::MESSAGE_UNUSED, $states, true) && $input->getOption('only-unused')
                    || !\in_array(self::MESSAGE_MISSING, $states, true) && $input->getOption('only-missing')
                ) {
```

alexandre-daubois: Ah yes indeed, I didn't see that two different consts were used and thought it could be simplified :facepalm: Thanks!

GromNaN (approved):

nicolas-grekas:
Thank you @alexandre-daubois.

PR #53079, opened 2023-12-14
[DoctrineBridge] Add check for lazy object interface

maxbaldanza:
In Symfony 6.4 lazy loading of ghost proxies is used out of the box.

This means if using 6.4 components with the 5.4 version of the doctrine
bridge you get the following error when trying to reset the entity manager:

```
Resetting a non-lazy manager service is not supported.
Declare the "doctrine.orm.default_entity_manager" service as lazy.
```

The entity manager is set as lazy
already in our case but it extends `\Symfony\Component\VarExporter\LazyObjectInterface` instead of the `LazyLoadingInterface` that's expected

| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no
| Issues        | 
| License       | MIT

### DIFF
--- a/src/Symfony/Bridge/Doctrine/ManagerRegistry.php
+++ b/src/Symfony/Bridge/Doctrine/ManagerRegistry.php
@@ -16,6 +16,7 @@ use ProxyManager\Proxy\GhostObjectInterface;
 use ProxyManager\Proxy\LazyLoadingInterface;
 use Symfony\Bridge\ProxyManager\LazyProxy\Instantiator\RuntimeInstantiator;
 use Symfony\Component\DependencyInjection\Container;
+use Symfony\Component\VarExporter\LazyObjectInterface;
 
 /**
  * References Doctrine connections and entity/document managers.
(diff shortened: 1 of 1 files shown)


src/Symfony/Bridge/Doctrine/ManagerRegistry.php
```diff
@@ -51,6 +52,13 @@ protected function resetService($name)
         }
         $manager = $this->container->get($name);
 
+        if ($manager instanceof LazyObjectInterface) {
```
maxbaldanza, line 55: This interface doesn't exist in the 5.4 branch so not sure on the best way to write a test for it so open to ideas 😄

nicolas-grekas (approved):
Works for me. This is special case because we made flex allow recent VarExporter to be combined with 5.4 apps to make it easier to upgrade to latest versions of Doctrine.

maxbaldanza:
> Works for me. This is special case because we made flex allow recent VarExporter to be combined with 5.4 apps to make it easier to upgrade to latest versions of Doctrine.

Thanks @nicolas-grekas 🙌

stof (approved):

fabpot (approved):

fabpot:
Thank you @maxbaldanza.

PR #53080, opened 2023-12-14
[Clock] Return Symfony ClockInterface in ClockSensitiveTrait

ruudk:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

@nicolas-grekas I don't understand why the ClockSensitiveTrait::mockTime returns a PSR ClockInterface, instead of a Symfony ClockInterface. Is this done on purpose? Or was this a mistake?

### DIFF
--- a/src/Symfony/Component/Clock/Test/ClockSensitiveTrait.php
+++ b/src/Symfony/Component/Clock/Test/ClockSensitiveTrait.php
@@ -11,7 +11,7 @@
 
 namespace Symfony\Component\Clock\Test;
 
-use Psr\Clock\ClockInterface;
+use Symfony\Component\Clock\ClockInterface;
 use Symfony\Component\Clock\Clock;
 use Symfony\Component\Clock\MockClock;
 


alexandre-daubois:
In the same idea, I'm not sure to clearly understand the convention in the code base on which class to use when one is available in Contracts, PSR and in the component itself (e.g. EventDispatcher)? :thinking:

nicolas-grekas:
> why the ClockSensitiveTrait::mockTime returns a PSR ClockInterface, instead of a Symfony ClockInterface

No specific reason I can think of...


> In the same idea, I'm not sure to clearly understand the convention in the code base on which class to use when one is available in Contracts, PSR and in the component itself (e.g. EventDispatcher)? 🤔

See SOLID principles: use the most abstract interface when it fits your need. If you need a more capable interface, use the more capable one.

stof:
Anytime a component only needs the API of the PSR interface, we use that as parameter type as it makes the component usable with any PSR-20 implementation and not just `symfony/clock`

alexandre-daubois:
Got it, thank you both :+1:

nicolas-grekas:
@ruudk please advise what you'd like to do with this PR. "Draft status" is blurry for OSS maintainers.

ruudk:
I made it Draft because I wasn't sure. So I made it ready for review now.

I think that the return type should be the thing that is going to be returned. That allows the developer, when using this in the test, to call `withTimeZone` on the returned Symfony Clock. Obviously, that already works, except that PHPStorm and PHPStan don't understand it.

nicolas-grekas (approved):

xabbuh (approved):
for 7.1 (this doesn't look like a bugfix to me)

fabpot (approved):

fabpot:
Thank you @ruudk.

PR #53081, opened 2023-12-14
[Serializer] Keep stack trace for enum value denormalizer error

kylekatarnls:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

Simple `previous: $exception` added not to loose track of the error origin.

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/BackedEnumNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/BackedEnumNormalizer.php
@@ -65,7 +65,7 @@ final class BackedEnumNormalizer implements NormalizerInterface, DenormalizerInt
             return $type::from($data);
         } catch (\ValueError $e) {
             if (isset($context['has_constructor'])) {
-                throw new InvalidArgumentException('The data must belong to a backed enumeration of type '.$type);
+                throw new InvalidArgumentException('The data must belong to a backed enumeration of type '.$type, 0, $e);
             }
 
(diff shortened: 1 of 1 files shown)


mtarld (commented):
Could you target 5.4 instead?

mtarld (approved):

src/Symfony/Component/Serializer/Normalizer/BackedEnumNormalizer.php
```diff
@@ -65,7 +65,7 @@ public function denormalize($data, string $type, string $format = null, array $c
             return $type::from($data);
         } catch (\ValueError $e) {
             if (isset($context['has_constructor'])) {
-                throw new InvalidArgumentException('The data must belong to a backed enumeration of type '.$type);
+                throw new InvalidArgumentException('The data must belong to a backed enumeration of type '.$type, previous: $e);
```
nicolas-grekas: 5.4 is compatible with PHP 7.2 so we cannot use named arguments

kylekatarnls:
```suggestion
                throw new InvalidArgumentException('The data must belong to a backed enumeration of type '.$type, 0, $e);
```

nicolas-grekas:
Thank you @kylekatarnls.

PR #53088, opened 2023-12-15
[Messenger] Fix using negative delay

### LINKED ISSUE #50833: [Messenger] Negative delay with Doctrine bridge can't be used

5.4.25

### Description

When we send envelop with negative delay with messenger configured with Redis bridge, messages are immediately processed. 

But it's not the case with Doctrine bridge because here, available_at column of the database queue table is wrongly set.

### How to reproduce

First we calc a $delayStamp that is representing the delay between the expected execution time (x day in the past) and the current time when we create the envelop (now). 

Because of the expected execution time is in the past, we want it to be executed with no delay.

```
$negativeDelayStamp = (($expectedExecuteAt->getTimestamp() - (new DateTime('now'))->getTimestamp()) * 1000);
```

Here we dispatch a simple envelop with a negative DelayStamp

```
$this->messageBus->dispatch(
            new Envelope(new MyMessage(), [
                new DelayStamp($negativeDelayStamp),
            ]));
```

Next in Symfony\Component\Messenger\Bridge\Doctrine\Transport\Connection $availableAt is calculated using the delay sent :

```
    public function send(string $body, array $headers, int $delay = 0): string
    {
        $now = new \DateTimeImmutable();
        $availableAt = $now->modify(sprintf('+%d seconds', $delay / 1000));

        $queryBuilder = $this->driverConnection->createQueryBuilder()
            ->insert($this->configuration['table_name'])
            ->values([
                'body' => '?',
                'headers' => '?',
                'queue_name' => '?',
                'created_at' => '?',
                'available_at' => '?',
            ]);
            
            ...
```

Because of the `+` in `sprintf('+%d seconds', $delay / 1000)` , the negative delay will be interpreted positively (ex: -1000 will be interpreted as +1000).

So instead of set availableAt at the expected execution time, x second in the past, it will set availableAt at the x second in the future.

### Possible Solution

I see multiple solution here :

- remove the `+` in `$availableAt = $now->modify(sprintf('+%d seconds', $delay / 1000));`

- throw an exception if negative delay is not a wanted feature

- do nothing and let the devs verify the delay value before they sent the envelop

### Additional Context

_No response_

J-roen:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #50833
| License       | MIT

Fixes using negative delays when sending messages.

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/DoctrineIntegrationTest.php
+++ b/src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/DoctrineIntegrationTest.php
@@ -67,10 +67,10 @@ class DoctrineIntegrationTest extends TestCase
         // DBAL 2 compatibility
         $result = method_exists($qb, 'executeQuery') ? $qb->executeQuery() : $qb->execute();
 
-        $available_at = new \DateTimeImmutable($result->fetchOne());
+        $availableAt = new \DateTimeImmutable($result->fetchOne());
 
         $now = new \DateTimeImmutable('now + 60 seconds');
-        $this->assertGreaterThan($now, $available_at);
+        $this->assertGreaterThan($now, $availableAt);
     }
 
     public function testSendWithNegativeDelay()
--- a/src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/DoctrineIntegrationTest.php
+++ b/src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/DoctrineIntegrationTest.php
@@ -73,6 +73,25 @@ class DoctrineIntegrationTest extends TestCase
         $this->assertGreaterThan($now, $available_at);
     }
 
+    public function testSendWithNegativeDelay()
+    {
+        $this->connection->send('{"message": "Hi, I am not actually delayed"}', ['type' => DummyMessage::class], -600000);
+
+        $qb = $this->driverConnection->createQueryBuilder()
+            ->select('m.available_at')
+            ->from('messenger_messages', 'm')
+            ->where('m.body = :body')
+            ->setParameter('body', '{"message": "Hi, I am not actually delayed"}');
+
+        // DBAL 2 compatibility
+        $result = method_exists($qb, 'executeQuery') ? $qb->executeQuery() : $qb->execute();
+
+        $available_at = new \DateTimeImmutable($result->fetchOne());
+
+        $now = new \DateTimeImmutable('now - 60 seconds');
+        $this->assertLessThan($now, $available_at);
+    }
+
     public function testItRetrieveTheFirstAvailableMessage()
     {
         $this->connection->setup();
--- a/src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/DoctrineIntegrationTest.php
+++ b/src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/DoctrineIntegrationTest.php
@@ -89,7 +89,7 @@ class DoctrineIntegrationTest extends TestCase
         $availableAt = new \DateTimeImmutable($result->fetchOne());
 
         $now = new \DateTimeImmutable('now - 60 seconds');
-        $this->assertLessThan($now, $available_at);
+        $this->assertLessThan($now, $availableAt);
     }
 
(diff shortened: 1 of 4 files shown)


src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/DoctrineIntegrationTest.php
```diff
@@ -73,6 +73,25 @@ public function testSendWithDelay()
...
+        $qb = $this->driverConnection->createQueryBuilder()
+            ->select('m.available_at')
+            ->from('messenger_messages', 'm')
+            ->where('m.body = :body')
+            ->setParameter('body', '{"message": "Hi, I am not actually delayed"}');
+
+        // DBAL 2 compatibility
+        $result = method_exists($qb, 'executeQuery') ? $qb->executeQuery() : $qb->execute();
+
+        $available_at = new \DateTimeImmutable($result->fetchOne());
```
OskarStark:
```suggestion
        $availableAt = new \DateTimeImmutable($result->fetchOne());
```

J-roen: Thanks for your feedback. I copied this from an existing test. Is it desired to update existing code in this same PR?

OskarStark: Yes please update it right away 👍 Thanks

J-roen: Done!

xabbuh:
Actually, I do not really understand the use case for using a negative delay. Is that really something that needs to be supported?

OskarStark:
As far as I understand, the problem is that it does not work right now, and this can happen due to calculation, so it should "just" send the message right?

valtzu:
@xabbuh If you have a long queue, you can cut in front of the queue with this negative delay because [messages are sorted by `available_at`](https://github.com/symfony/symfony/blob/6e419d84e8b4d65fd6a0ac416c18b8d2c894d495/src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/Connection.php#L173).

So it's about priority within a single queue basically.

J-roen:
Actually, it was already possible to set negative delays, but the behaviour was unexpected (in my opinion). Another solution could be to set the time to "now" in case of a negative delay. But then use cases like described by @valtzu wouldn't be supported.

smnandre:
I'm not certain on this because:
- we should never expect things based on bridge implementation details
- i'm not certain it really "works" with this PR (everytime for everyone i mean).. I think about postgres notify behaviour, hard indexes on dbs... So this would need a loot of tests .. it this worth the effort ?

As i see it, you removed a valid safeguard in the code to allow a "non-written" behaviour..  and i'm pretty sure this will not lead to good things :) 

And finally... there is a simpler /  documented way to handle queue priority: [queue priority](https://symfony.com/doc/current/messenger.html#prioritized-transports).  On a fonctional level, does this not fill your needs ?

J-roen:
I agree that you can discuss about the expected behaviour, but I think that what you at least wouldn't expect, is that a negative delay is converted to a positive delay. How it's supposed to work apart from that, could maybe remain undefined.

smnandre:
> I agree that you can discuss about the expected behaviour, but I think that what you at least wouldn't expect, is that a negative delay is converted to a positive delay. How it's supposed to work apart from that, could maybe remain undefined.

I totally agree!

nicolas-grekas (approved):

fabpot (approved):

fabpot:
@J-roen Can you create a PR for Symfony 5.4, which is also affected?

J-roen:
> @J-roen Can you create a PR for Symfony 5.4, which is also affected?

Yes, I can do that later this week. And what about 7.x?

Nyholm (approved):
Thank you

>> @J-roen Can you create a PR for Symfony 5.4, which is also affected?
>
> Yes, I can do that later this week. And what about 7.x?

It is no need to add this to 7.x. When this is merged in 6.4 it will be merge "up" to 7.0 and 7.1 by a maintainer.

PR #53096, opened 2023-12-16
[Intl] [Emoji] Move emoji data in a new component

smnandre:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Issues        | 
| License       | MIT

This PR move all the emoji data & code from the Intl component into its own new Emoji component.

Objectives/reasons:
* reduce the size of a standard "--webapp" install
* allow usage of intl (required if your app uses those validators/form types: BIC, country, currency, language, locale, timezone) without downloading all the emoji data
* allow usage of Emoji without downloading all the Intl data 

Thanks to all the reviewers for the feedbacks, opinions, advices ❤️ 

---

--- Original (obselete) post below --- 

This PR move all the emoji data & code from the Intl component into its own new IntlEmoji component.

... and hopefully open a debate aboute the future of the Intl component, its role, and the way we handle "data" in the framework and the repositories 

> [!IMPORTANT]
> 🎙️ DISCLAIMER: This PR contains both metrics and opinions. The metrics were collected this morning in a neutral and transparent manner, ensuring they can be [reproduced by anyone](https://gist.github.com/smnandre/38b081253fc8813a5e638661650b01f6). However, the opinions I present here are just that – opinions, and should not be interpreted as objective truths or claims of fact.

Update: details/summary added to improve the readability

<details>
<summary>

## What is symfony/intl ?

</summary>

Repository: https://github.com/symfony/intl
Documentation: https://symfony.com/doc/current/components/intl.html
Unicode: https://home.unicode.org/technical-quick-start-guide/

### Responsabilities

Currently, it seems to me this component:
* provides a polifill-ish layer for the intl PHP extension
* provides exhaustive formatting + locale data for countries, currencies, date, ... 
* provides methods to check ISO codes or identifiers
* provides a dictionnary emoji -> description for every possible combinaison

### Opiniated remarks

Two comments highlight the "blurred lines" I believe this component navigates:

#### 1) Access or data ?

> This component provides **access** to the localization **data** of the ICU library.

Maybe my english is in fault, but it seems to me it does not provides access to the data... it provides the data.

#### 2) Unicode = CLDR + ICU + UTC

> This component provides access to the localization data of the **ICU library**.

CLDR (where the emoji data comes from) is not in the ICU library. I'm quibbling over details here, i know. But i think that illustrates the volatile   "scope" and "responsabilities" of this component.

So we come close to the problem...
</details>

<details>
<summary>

## Symfony/Intl is massive

</summary>

The data included in the Intl component is massive (especially the emoji descriptions), and will grow more every semester.

I looked at the following cases
* source code
    * symfony/symfony  : https://github.com/symfony/symfony/archive/refs/heads/7.0.zip
    * symfony/intl  : https://github.com/symfony/intl/archive/refs/heads/7.0.zip
* installed 
    * standalone:  new folder + composer require symfony/intl 
    * webapp: new folder + symfony create --webapp

Versions: 
* 6.0 6.1 6.2 6.3 6.4 7.0
* Emoji data were added in 6.2
  

### Some metrics...

|                 | Size (zip) | Size (unzipped) | Files | PHP Files |  
| - | - | - | - | - |
| symfony/symfony |      12.91 |           58.6  |  6006 |      4729 |
| symfony/intl    |       7.5  |           43.2  |  1517 |      1487 |
| %               |      58.1% |           73.8% | 25.3% |     31.5% |

So symfony/intl accounts for 30% of the files in the monorepo ... and nearly **75% of its total disk size**.

### ...over time

**Size (in MB) of the sources** 

|        | 6.0 | 6.1 | 6.2 | 6.3 | 6.4 | 7.0 |
|--------|-----|-----|-----|-----|-----|-----|
| intl   |15.1 |15.1 |41.8 |41.9 |43.2 |43.2 |
| symfony|28.8 |29.1 |56.5 |58.1 |59.8 |58.6 |

It was already big in previous versions, but since the emoji data integration, it's off charts. 

The symfony/intl alone is twice as big as: all the other components, all the bridges and all the bundles. Combined.

And it's not over. At all.

### Why it'll grow more

The ICU components used in the component are well-defined and constrained by 'real-world' factors, so we can expect minor changes regarding countries, formatting data, etc. It's unlikely, for instance, that 200 new countries will suddenly emerge in 2024.

However, emojis may present a major challenge in the near future. New ones are added with every CLDR release. Except for a significant drop (like the upcoming 2000 hieroglyphs), this should be a gradual increase.

What bothers me more is the 'combinatorial nature' of these descriptions. We generate a line of text for every combination. And that's why this component is so large. But it's just the beginning of what could be exponential growth.

As of today, the 'hand emoji' has variations for skin color (I'm not certain, but let's say there are 6 possible colors), and emojis with multiple people often vary by gender ('boy and two girls').

In the upcoming release, a new variable is the concept of 'left-handed' versus 'right-handed'. So, we'll create a new line for every existing emoji with a visible hand. But we'll need way more than just a new line, because of every emoji where two hands are visible. I don't remember if it's already implemented, but there was discussion about including the same thing for the age of a person, or some hairstyles.

So, the symfony/intl component could very soon be 50GB, and a short while later 10^80TB. But there's no way it reduces in size... or even slows its growth.
</details>

<details>
<summary>

## And.... where is the problem ?

</summary>

I see negative effect on three very different layers.

### Developper Experience

Whether these values are low or not in absolute terms (and I have no doubt that everyone will have their own opinion on this)... the reality is that users are downloading a component that is twice as heavy as all the others combined... and this inevitably affects installation times, bandwidth, update times, static analysis, IDE indexing, etc. A prime example is Docker on macOS, which was a real pain until recently with Orbstack, and the performance nightmare was directly related to the number of files mounted in a volume.

### Contributor experience

I've lost count of how many times I've seen a contributor propose a feature only to be told: it's userland. (Full disclosure, I understand and share this point of view). But it can be frustrating to see closed doors for a few classes, while at the same time Symfony contains hundreds of lines like 'young woman with dark hair and kid'

### Real world consequences: Ecological & financial costs

I have no desire to open a debate (on either of those topics). But again, these small things have real-world consequences. We are talking about Symfony, so the impact is enormous, even on small matters

### What is the real impact ?

Downloads data, as provided by packagist (collected today)

| Package | install total | install (last 30 days) |
|---------|---------------|------------------------|
| symfony/intl | 111,000,000 | 2,800,000 |
| [symfony/http-foundation](https://packagist.org/packages/symfony/http-foundation) | 528,000,000 | 11,250,000 |
| symfony/console | 678,000,000 | 12,900,000 |
| symfony/console | 678,000,000 | 12,900,000 |
| symfony/translation | 508,000,000 | 10,400,000 |

Let's agree on: "it's not without an impact".

</details>

<details>
<summary>

## Why is it used ?

</summary>

### For its quality

Please don't misinterpret my message. I'm not criticizing the value of the component or questioning its qualities. Besides, my opinion wouldn't have any value for that matter anyway. And i'm absolutely convinced a lot of people decide to install this component knowing what they do.

### For another reason

But there are also people who install... Symfony. The recommended installation procedure, as outlined in the documentation on the website, is to install the web application skeleton, which requires symfony/intl.

To revisit the argument from earlier, I'm really not sure if anyone realizes after installation why its vendors directory is 80MB and what it's used for (young woman with...).

I'm unsure why symfony/intl is included by default in a new project, while other components are not. As a developer, I would appreciate the ability to install a small, lightweight application or to have more packages for the same amount of overhead :)

### For another reason (bis)

"There is a third reason, and once again, I'm not fully understanding the situation (and may not have all the backstory required for it).

The Country validator requires symfony/intl to validate a given string as a valid ISO alpha country code. To do this, it tries to retrieve the list of country names (indexed by code) from the locale data. Consequently, it's not possible to use BIC, Country, Currency, and probably others without symfony/intl.

So, if a developer installs symfony/validator and then wants to validate a BIC, they cannot do so without downloading 80MB of locale-specific data.

Wouldn't it be simpler to have a couple of ISO classes/methods in the Validator component? Or perhaps create a small component just for that purpose? Because having to parse giant files just to check if "FR" is a valid ISO country code seems quite inefficient to me.

</details>

## Well: Suggestions

So, personal conclusion and some suggestions..

### TODO list

Sooner
* move "emoji" out of Symfony\Intl
* remove "intl" from symfony/webapp
* fix the validator requirements

Later
* move "intl" out of the monorepo
* create a distinct class/component to handle iso lists

### Discussions

- [x] Handle the BC layer (require symfony/intl-emoji in symfony/intl ?)
- [x] Find a way to handle the exponential growth of those files

--

Open to any feedback :)


OskarStark:
Related
* https://github.com/symfony/symfony/issues/49943

src/Symfony/Component/Intl/CHANGELOG.md
```diff
@@ -1,10 +1,16 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+* Transfert all emoji code & data to a new dedicated component: `Symfony\Component\IntlEmoji`
+* Deprecate `EmojiTransliterator` in favor of `Symfony\Component\IntlEmoji\EmojiTransliterator`
```
OskarStark:
```suggestion
 * Transferred all emoji code & data to a new dedicated component: `Symfony\Component\IntlEmoji`
 * Deprecate `EmojiTransliterator` in favor of `Symfony\Component\IntlEmoji\EmojiTransliterator`
```

OskarStark: Why `Symfony\Component\IntlEmoji` and not `Symfony\Component\Emoji` ?

src/Symfony/Component/IntlEmoji/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+CHANGELOG
+=========
+
+7.1
+---
+
+* Create component (import all emoji code & data from `Symfony\Component\Intl`
```
OskarStark:
```suggestion
 * Add component (import all emoji code & data from `Symfony\Component\Intl`
```

src/Symfony/Component/IntlEmoji/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2022-present Fabien Potencier
```
OskarStark:
```suggestion
Copyright (c) 2023-present Fabien Potencier
```

the new component is, if merged introduced in 2023 or 2024

src/Symfony/Component/IntlEmoji/README.md
```diff
@@ -0,0 +1,13 @@
+Intl-Emoji Component
+=============
+
+The Intl-Emoji component provides access to the localization data of the ICU library.
```
OskarStark: The sentence is wrong

src/Symfony/Component/IntlEmoji/phpunit.xml.dist
```diff
@@ -0,0 +1,37 @@
...
+
+    <testsuites>
+        <testsuite name="Symfony Intl Emoji Component Test Suite">
+            <directory>./Tests/</directory>
+        </testsuite>
+    </testsuites>
+
+    <groups>
+        <exclude>
+            <group>intl-data</group>
```
OskarStark: Is the group really needed? it is all about intl data then, isn't it?

src/Symfony/Component/IntlEmoji/README.md
smnandre: Updated

src/Symfony/Component/IntlEmoji/LICENSE
smnandre: Ok i did not know, as the code is "moved". Fixed.

src/Symfony/Component/IntlEmoji/phpunit.xml.dist
smnandre: Wrong copy/past, sorry.

src/Symfony/Component/Intl/CHANGELOG.md
smnandre: My (humble personal) opinion on this would be:
* Symfony\Component\Emoji could be something to manipulate... emojis (so unicode caracters, encoding, ranges)  but it could then probably be into the String component
* Symfony\Component\Intl-Emoji is the "localisation-layer" to get every emojis short name, description and variations in any language (something that could be accessible via the current provided files, but probably sooner than later via OS api, php extensions, ...)

In a ideal world, String component could handle emoji transliteration not knowing where the data is from.. and IntlEmoji installed only by who needs exhaustive data and/or in a "polyfill-ish" way.

smnandre:
> Related
> 
>     * [Splitt intl packages in smaller packages #49943](https://github.com/symfony/symfony/issues/49943)

I have no memory of this PR, but there is a thumbs-up from me, so... 😮‍💨

smnandre:
Fabbot fails but i cannot fix it (as it wants to edit emoji files to add licence, fix typos, etc.)

smnandre:
In the end, i renamed IntlEmoji just Emoji, as nothing from Emoji uses/interact with the Intl component. 

Moreover, Emoji requires native php Intl extension.

nicolas-grekas:
My understanding of this topic:

- the intl data is NOT in the base skeleton. It is in webapp-pack.
- even if the data is not compressed in the source, it's always compressed on the network, so that uncompressed size doesn't matter when considering transfer time/overhead
- a fresh *compressed* webapp is ~56MB
- when we remove all intl data, the compressed size becomes ~48MB (-13%)
- when we remove only the emoji data, the compressed size becomes ~52MB (-7%)

About symfony/intl, I think it does what is should: make the intl data easily available.
About splitting emoji data, it's worth 7% of the webapp skeleton.
I've no strong objection to make the split (but no strong motivation either, so thanks for pushing the topic :) )

.github/workflows/emoji-data-tests.yml
```diff
@@ -0,0 +1,68 @@
+name: Emoji data
```
nicolas-grekas: can we do without a dedicated action?
I prefer centralizing all components in one to ease maintenance

composer.json
```diff
@@ -82,6 +82,7 @@
         "symfony/http-foundation": "self.version",
         "symfony/http-kernel": "self.version",
         "symfony/intl": "self.version",
+        "symfony/intl-emoji": "self.version",
```
nicolas-grekas: alpha order and this:
```suggestion
        "symfony/emoji": "self.version",
```

src/Symfony/Component/Emoji/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+CHANGELOG
+=========
+
+7.1
+---
+
+* Add component (import all emoji code & data from `Symfony\Component\Intl`)
```
nicolas-grekas:
```suggestion
 * Add the component
```

src/Symfony/Component/Emoji/EmojiTransliterator.php
```diff
@@ -0,0 +1,139 @@
...
+ */
+
+namespace Symfony\Component\Emoji;
+
+use Symfony\Component\Emoji\Util\GzipStreamWrapper;
+
+/**
+ * @internal
+ */
+trait EmojiTransliteratorTrait
```
nicolas-grekas: do we still need the trait? let's merge it within the class if not

src/Symfony/Component/Emoji/README.md
```diff
@@ -0,0 +1,21 @@
+Emoji Component
+=============
```
nicolas-grekas:
```suggestion
===============
```

src/Symfony/Component/Emoji/README.md
```diff
@@ -0,0 +1,21 @@
...
+The Emoji component provides access to emoji characters and sequences from
+the [Unicode CLDR](https://cldr.unicode.org/index).
+
+The total size of the included data is about 30MB. You can diminish the size of
+the data by compressing those files using the following command:
+
+    php vendor/symfony/emoji/Resources/bin/compress
+
+The ext-Intl PHP extension is required.
+
```
nicolas-grekas:
```suggestion
If you have the zlib extension enabled, you can compress the data by running:

    php vendor/symfony/emoji/Resources/bin/compress
```

Let's add the same sentence in the intl component also

src/Symfony/Component/Emoji/Resources/bin/build.php
```diff
@@ -25,7 +25,7 @@
 
 final class Builder
 {
-    private const TARGET_DIR = __DIR__.'/../data/transliterator/emoji/';
+    private const TARGET_DIR = __DIR__.'/../data/emoji/';
```
nicolas-grekas:
```suggestion
    private const TARGET_DIR = __DIR__.'/../data/';
```

src/Symfony/Component/Intl/CHANGELOG.md
```diff
@@ -1,10 +1,16 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Transferred all emoji code & data to a new dedicated component: `Symfony\Component\Emoji`
```
nicolas-grekas:
```suggestion
 * Transferred all emoji code & data to a new `symfony/emoji` component
```

src/Symfony/Component/Intl/composer.json
```diff
@@ -24,11 +24,11 @@
         }
     ],
     "require": {
-        "php": ">=8.2"
+        "php": ">=8.2",
+        "symfony/emoji": "^7.0"
```
nicolas-grekas: this should be removed - we don't want to make ext-intl a mandatory dep of symfony/intl
we can ease discoverability by throwing a logic exception when the legacy EmojiTransliterator is missing the new component

nicolas-grekas (commented):
I think this approach works, thanks for proposing.
Here are some comments.

smnandre:
> when we remove all intl data, the compressed size becomes ~48MB (-13%)

There is something wrong with your computation... an **_uncompressed_** webapp without symfony/intl is only 43MB

| webapp | webapp - Intl | 
| - | - | 
| Capture d’écran 2023-12-19 à 12 46 32 | Capture d’écran 2023-12-19 à 12 47 36 |


---


Thank you for the feedbacks, i'll fix/adapt things accordingly this afternoon  😄

src/Symfony/Component/Intl/composer.json
smnandre: Still navigating the BC world in my head (slow learner 😅): if we don't guarantee that symfony/emoji is installed, a project currently using the EmojiTransliterator will break after a composer update, no ?

nicolas-grekas:
```sh
symfony new test-webapp --webapp
du -sh test-webapp
#displays 154MB (uncompressed - 141MB just for vendor)
rm test-webapp/vendor/symfony/intl/ -rf
du -sh test-webapp
#displays 109MB (still uncompressed - 96MB just for vendor)
```

src/Symfony/Component/Intl/composer.json
nicolas-grekas: There are two aspects to BC here:

1. this is the BC promise, and you'll see no mention of deps there: https://symfony.com/doc/current/contributing/code/bc.html
2. requiring an extension is another BC break from another PoV - could be a worse one since it requires a change to the infrastructure

We already did similar changes in the past. Yes, we prefer doing things as smooth as possible, but adding a required extension doesn't fit.

stof:
> Currently, it seems to me this component:
> 
>     * provides a polifill-ish layer for the intl PHP extension

this is not true. The polyfills are in `symfony/polyfill-intl-icu`

src/Symfony/Component/Intl/composer.json
smnandre: Yep ok i understand (not all yet, but more than 12mn ago 😃 )

smnandre:
> ```shell
> du -sh test-webapp
> ```

That is interesting. Really. Because i have clearly not the same numbers. And i wonder why.

```
du -sh test-webapp
105M	test-webapp
rm test-webapp/vendor/symfony/intl/ -rf
rm: test-webapp/vendor/symfony/intl/: is a directory
rm: -rf: No such file or directory
rm -rf test-webapp/vendor/symfony/intl/
du -sh test-webapp
 60M	test-webapp
```

(yes i let you enjoy my `-rf` fail)

smnandre:
I'm not sure how i can fix the remaining test... if someone can help me / point me in the good direction that'd be nice :)

javiereguiluz:
@smnandre just a random comment related to this.

In Symfony we have a config option called [enabled_locales](https://symfony.com/doc/current/reference/configuration/framework.html#enabled-locales). We could delete _(automatically or letting developers doing it explicitly via a command provided by us)_ all the data of all languages not included in that config option. We'd delete 95% of Intl data (emojis transliterators, names of coutnries/languages/currencies in other languages, etc.) without affecting to the application.

smnandre:
> @smnandre just a random comment related to this.
> 
> In Symfony we have a config option called [enabled_locales](https://symfony.com/doc/current/reference/configuration/framework.html#enabled-locales). We could delete _(automatically or letting developers doing it explicitly via a command provided by us)_ all the data of all languages not included in that config option. We'd delete 95% of Intl data without affecting to the application.

I suppose that could solve the "final deployable size" question... even more if combined with the "compress script" 

But... that would still download every emojis variants x every languages everytime you create a webapp, run composer install, or need the CountryValidator. 

I won't push too hard on this, even less if i'm far from beeing in the majority, thinking that this data should not be in the main repo. I guess my pov on this is more "philosophical" or "opiniated" than i imagined... :) 

As i said while investigating [this issue](https://github.com/symfony/symfony/issues/53116), i have some alternative ideas to reduce/optimise those files, it can be a step in the good* direction ! (*according to me :) )

nicolas-grekas:
Let's continue with this approach. 7% compressed diff also means less time to uncompress the app so that can lead to a slightly better DX.
rebase needed though :)

smnandre:
> Let's continue with this approach. 7% compressed diff also means less time to uncompress the app so that can lead to a slightly better DX. rebase needed though :)

Ok will do tomorrow. I can reduce 25% of uncompressed data files... by removing all spaces (only 3/4% when compressed... as expected)

Do we need to php-cs-fix those files ?

nicolas-grekas:
>  I can reduce 25% of uncompressed data files... by removing all spaces (only 3/4% when compressed... as expected)

not worth it to me

> Do we need to php-cs-fix those files ?

nope, fabbot can be ignored on those

smnandre:
Ok ! 

I'll fix what i can/figure out.. and probably will call for a bit of help :)

smnandre:
Failing checks:
* fabbot: want to add LICENCE to emoji resource files 
* Unit Tests (8.2 high/low): related to Form / DateType
* Unit Tests (8.3): related to Form / DateType

So remains Psalm: 
* the Transliterator::id" **is** a readonly string inPHP 8.2+ so it should not raise alarm (?)
* the 2 other errors concern build.php (that requires a composer install)

.github/workflows/intl-data-tests.yml
```diff
@@ -84,11 +90,49 @@ jobs:
...
+            echo "::endgroup::"
+
+            echo "::group::install phpunit"
+            ./phpunit install
+            echo "::endgroup::"
+
+      -   name: Run tests
+          run: ./phpunit src/Symfony/Component/Emoji
+
+      -   name: Test with compressed data
```
nicolas-grekas: let's move these as steps in the previous job, no need to repeat installing php et al.

src/Symfony/Component/Intl/CHANGELOG.md
```diff
@@ -1,10 +1,16 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Transferred all emoji code & data to a new `symfony/emoji` component
```
nicolas-grekas:
```suggestion
 * Move all emoji code & data to a new `symfony/emoji` component
```

src/Symfony/Component/Intl/composer.json
```diff
@@ -24,11 +24,11 @@
         }
     ],
     "require": {
-        "php": ">=8.2"
+        "php": ">=8.2",
+         "symfony/deprecation-contracts": "^2.5|^3"
```
nicolas-grekas:
```suggestion
        "symfony/deprecation-contracts": "^2.5|^3"
```

src/Symfony/Component/Intl/Transliterator/EmojiTransliterator.php
```diff
@@ -11,133 +11,21 @@
...
-
-        public static function listIDs(): array
-        {
-            static $ids = [];
+trigger_deprecation('symfony/intl', '7.1', 'The "%s" class is deprecated, use "%s" instead.', EmojiTransliterator::class, EmojiEmojiTransliterator::class);
 
-            if ($ids) {
-                return $ids;
-            }
+class_exists(EmojiEmojiTransliterator::class);
```
nicolas-grekas:
```suggestion
if (!class_exists(EmojiEmojiTransliterator::class)) { // throw LogicException
```

nicolas-grekas (commented):
We should be able to solve the deps=high failure by stating that intl conflicts with string < 7.1

src/Symfony/Component/Intl/Transliterator/EmojiTransliterator.php
```diff
@@ -11,133 +11,23 @@
...
-    use EmojiTransliteratorTrait;
-
-    private const QUICK_CHECK = "\xA9\xAE\xE2\xE3\xF0";
-    private const REVERSEABLE_IDS = [
-        'emoji-github' => 'github-emoji',
-        'emoji-slack' => 'slack-emoji',
-        'github-emoji' => 'emoji-github',
-        'slack-emoji' => 'emoji-slack',
-    ];
+if (!class_exists(EmojiTransliterator::class, false)) {
```
nicolas-grekas: no need for this check

nicolas-grekas (approved):
LGTM, just one minor thing :rocket:

nicolas-grekas:
> Merge branch 'symfony:7.1' into feat/component-intl-emoji

:boom: 
We don't merge PRs with merge commits :)

smnandre:
If possible, could you configure fabbot to ignore `src/Symfony/Component/Emoji/Resources/data/`

(cf https://fabbot.io/report/symfony/symfony/53096/25c5b345a64dbce84dd180c8e8c779af5e2d8fe0 )

smnandre:
(squashed / rebased on 7.1)

smnandre:
Is there something else i can do here ? :)

stof:
To make the description fairer, installing `symfony/validator` does not force installing `symfony/intl` (we don't have a composer requirement on it. We have a runtime check when using the intl-based constraint)

smnandre:
@stof I updated the description 

> allow usage of intl (required if your app uses those validators/form types: BIC, country, currency, language, locale, timezone) without downloading all the emoji data

Feel free to rephrase or edit directly if you have another idea in mind!

fabpot (approved):

fabpot:
Thank you @smnandre.

src/Symfony/Component/Intl/composer.json
```diff
@@ -24,13 +24,16 @@
...
+        "php": ">=8.2",
+        "symfony/deprecation-contracts": "^2.5|^3"
     },
     "require-dev": {
         "symfony/filesystem": "^6.4|^7.0",
-        "symfony/finder": "^6.4|^7.0",
         "symfony/var-exporter": "^6.4|^7.0"
     },
+    "conflict": {
+        "symfony/string": "<7.1"
```
xabbuh, line 35: Why do we need the conflict?

smnandre: Better link to this message than trying to rephrase it (i'd fail)
https://github.com/symfony/symfony/pull/53096#pullrequestreview-1800071133

src/Symfony/Component/String/composer.json
```diff
@@ -24,8 +24,9 @@
     },
     "require-dev": {
         "symfony/error-handler": "^6.4|^7.0",
-        "symfony/intl": "^6.4|^7.0",
+        "symfony/emoji": "^7.0",
```
xabbuh, line 27: `^7.1` (fixed in f0ddd8da8930eedc03085e2c59b8edb9a2b6b889)

alexander-schranz:
Here is a small bc break if we do not require symfony/emoji component inside `symfony/string` component it breaks AsciiSlugger withEmoji support when update from Symfony 7.0 -> 7.1 and application throws then an exception which did work on Symfony 7.0.

PR #53098, opened 2023-12-16
[Validator] Add `ThisableMessage`

emile-prevot-42:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

Hello !
I really like the way you can place assertions on Symfony :
```php
class Person
{
    protected string $firstName;

    #[Assert\EqualTo(
       value: 20,
       message: 'This value should be equal to {{ compared_value }}.'
    )]
    protected int $age;
}
```

But if our need becomes a little more complex, as when we want to validate several Person at once :

```php
class GroupOfPersons
{
    #[Assert\Valid]
    protected array $persons;
}
```

If we validate this object with some valid Persons and others not, we'll have a large list of exceptions containing "This value should be equal to 20" with the value path. That's enough to meet many needs, and twig with the form component displays them correctly.

But my need was simply to display "This value should be equal to 20 for the person whose first name is Émile." and I couldn't find any elegant way of doing it.

So I've coded this Assertion which makes it possible to do :
```php
class Person
{
    protected string $firstName;

    
    #[Assert\ThisableMessage([
        new Assert\EqualTo(
           value: 20,
           message: 'This value should be equal to {{ compared_value }} for the person whose first name is {{ this.firstName }}.'
       )
    ])]
    protected int $age;
}
```

If you find this feature interesting, I'd be happy to write the documentation and the changelog, but I might need a little help to improve the test and source code. And maybe "ThisableMessage" is a name that can evolve.

I use it to validate CSVs containing hundreds of rows over dozens of columns, and adding this assertion had no noticeable effect on performance.

Thank you,

### DIFF
--- a/src/Symfony/Component/Validator/Constraints/ThisableMessage.php
+++ b/src/Symfony/Component/Validator/Constraints/ThisableMessage.php
@@ -26,8 +26,10 @@ class ThisableMessage extends Composite
 
     public $addRootParameters = [];
 
-    public function __construct(mixed $constraints = null, array $groups = null, mixed $payload = null)
+    public function __construct(mixed $constraints = null, array $addThisParameters = [],  array $addRootParameters = [], array $groups = null, mixed $payload = null)
     {
+        $this->addThisParameters = $addThisParameters;
+        $this->addRootParameters = $addRootParameters;
         parent::__construct($constraints ?? [], $groups, $payload);
     }
 
--- a/src/Symfony/Component/Validator/Constraints/ThisableMessage.php
+++ b/src/Symfony/Component/Validator/Constraints/ThisableMessage.php
@@ -1,0 +1,53 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Validator\Constraints;
+
+/**
+ * @Annotation
+ * @Target({"CLASS", "PROPERTY", "METHOD", "ANNOTATION"})
+ *
+ * @author Émile PRÉVOT <emile@level21.io>
+ */
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_PROPERTY | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
+class ThisableMessage extends Composite
+{
+    public $constraints = [];
+
+    public $addThisParameters = [];
+
+    public $addRootParameters = [];
+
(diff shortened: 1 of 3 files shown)


sstok:
I believe you can already do something a bit more complex using the ExpressionValidator (which is much more advanced than only `$this`). https://symfony.com/doc/current/reference/constraints/Expression.html or https://symfony.com/doc/current/reference/constraints/Callback.html if you need a truly custom message.

emile-prevot-42:
> I believe you can already do something a bit more complex using the ExpressionValidator (which is much more advanced than only `$this`). https://symfony.com/doc/current/reference/constraints/Expression.html or https://symfony.com/doc/current/reference/constraints/Callback.html if you need a truly custom message.

ExpressionValidator does not allow you to customize the error message with {{ this. }}
And CallbackValidator can do the job of any assertion ¯ \ _ (ツ) _ / ¯

94noni:
what about [this one](https://symfony.com/blog/new-in-symfony-6-2-conditional-constraints)?

emile-prevot-42:
> what about [this one](https://symfony.com/blog/new-in-symfony-6-2-conditional-constraints)?

Here too, the message remains unchanged.
`this` is only in `expression` and not `message`.

ro0NL:
i guess it would be nice if this just worked, without the need for extra `ThisableMessage`

the feature would be binding `this`, or crash if used but not bound

emile-prevot-42:
> i guess it would be nice if this just worked, without the need for extra `ThisableMessage`
> 
> the feature would be binding `this`, or crash if used but not bound

It would be nice... but I think it would be a big refactoring job for the team to rethink how violations and their parameters are generated

Because at the moment, all variables implemented as {{ value }} are defined in their validator (IsNullValidator for example).

emile-prevot-42:
> i guess it would be nice if this just worked, without the need for extra `ThisableMessage`
> 
> the feature would be binding `this`, or crash if used but not bound

I'm trying to find a place inside the framework where to insert the {{ this }} behavior without using an annotation.

Without annotation, there's the problem of using translation keys.

With the current assertion, the operation will look like this :

```php
class Person
{
    protected string $firstName;

    #[Assert\ThisableMessage(
        addThisParameters: ['firstName'],
        constraints: [
            new Assert\EqualTo(
               value: 20,
               message: 'person.error.age'
            )
        ]
    )]
    protected int $age;
}

nicolas-grekas:
Closing as stale. This has been open since December 2023 without a core review, the name and the wrapper-constraint approach were both questioned in the discussion, and the preferred direction raised there was to make 'this' resolvable in messages natively rather than through a dedicated constraint. The validator also still uses the pre-7.0 ConstraintValidator signature, so it would need reworking before it could run. Thanks for the proposal, and sorry for the long silence. If you want to pursue the native-binding direction, open an issue first and we can discuss the design.

PR #53107, opened 2023-12-17
[HttpKernel] Don't validate partially denormalized object

### LINKED ISSUE #53075: #[MapRequestPayload] Returns full ViolationsList if object cannot be created due to a single missing value (re-open)

6.4

### Description

Hey! I re create https://github.com/symfony/symfony/issues/50759 since a comment did not re-open it. Unfortunately, this is not fixed in the latest 6.4 Update. 

> If you use #[MapRequestPayload] and the payload is missing a single (or multiple) fields, the ViolationList created by the resolver does not contain the single missing field, but all asserts on every field.

### How to reproduce

Updated the Project https://github.com/Havrin/symfony-payload-bug to 6.4 and tested again against `http://localhost/api/payload` with payload:
```json
{
  "password": "abc"
}
```

Result (first object is a `var_dump` inisde the exception subscriber):
```
object(App\DTO\User)#191 (0) {
    [
        "email":"App\DTO\User":private
    ]=>
  uninitialized(string)
  [
        "password":"App\DTO\User":private
    ]=>
  uninitialized(string)
}
{
    "type": "https://symfony.com/errors/validation",
    "title": "Validation Failed",
    "status": 422,
    "detail": "This value should be of type unknown.\nemail: This value should not be null.\nemail: This value should not be blank.\npassword: This value should not be null.\npassword: This value should not be blank.",
    "violations": [
        {
            "propertyPath": "",
            "title": "This value should be of type unknown.",
            "template": "This value should be of type {{ type }}.",
            "parameters": {
                "{{ type }}": "unknown",
                "hint": "Failed to create object because the class misses the \"email\" property."
            }
        },
        {
            "propertyPath": "email",
            "title": "This value should not be null.",
            "template": "This value should not be null.",
            "parameters": {
                "{{ value }}": "null"
            },
            "type": "urn:uuid:ad32d13f-c3d4-423b-909a-857b961eb720"
        },
        {
            "propertyPath": "email",
            "title": "This value should not be blank.",
            "template": "This value should not be blank.",
            "parameters": {
                "{{ value }}": "null"
            },
            "type": "urn:uuid:c1051bb4-d103-4f74-8988-acbcafc7fdc3"
        },
        {
            "propertyPath": "password",
            "title": "This value should not be null.",
            "template": "This value should not be null.",
            "parameters": {
                "{{ value }}": "null"
            },
            "type": "urn:uuid:ad32d13f-c3d4-423b-909a-857b961eb720"
        },
        {
            "propertyPath": "password",
            "title": "This value should not be blank.",
            "template": "This value should not be blank.",
            "parameters": {
                "{{ value }}": "null"
            },
            "type": "urn:uuid:c1051bb4-d103-4f74-8988-acbcafc7fdc3"
        }
    ],
    "class": "Symfony\\Component\\HttpKernel\\Exception\\HttpException",
...
```
The first Violation of Type `unknown` holds the correct message: `"Failed to create object because the class misses the \"email\" property."` which should be returned instead of every assert/validation.

### Possible Solution

_No response_

### Additional Context

Apologize if re creating a new ticket was not the right move. Feel free to close this and re open the other one.

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #53075
| License       | MIT

When one or more required constructor arguments are missing, the object is instantiated with `newInstanceWithoutConstructor()`:
https://github.com/symfony/symfony/blob/0af28a02dbfabef351de8a739dee65830ed0271a/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php#L415-L421

Since this means that all properties are empty, validating the entire object is unnecessary. This PR suggests avoiding the call to the validator when violations have already been identified through the `PartialDenormalizationException`.

### DIFF
--- a/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php
+++ b/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php
@@ -119,7 +119,7 @@ class RequestPayloadValueResolver implements ValueResolverInterface, EventSubscr
                     $payload = $e->getData();
                 }
 
-                if (null !== $payload) {
+                if (null !== $payload && !\count($violations)) {
                     $violations->addAll($this->validator->validate($payload, null, $argument->validationGroups ?? null));
                 }
 
--- a/src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/RequestPayloadValueResolverTest.php
+++ b/src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/RequestPayloadValueResolverTest.php
@@ -27,7 +27,6 @@ use Symfony\Component\Serializer\Exception\PartialDenormalizationException;
 use Symfony\Component\Serializer\Normalizer\ObjectNormalizer;
 use Symfony\Component\Serializer\Serializer;
 use Symfony\Component\Validator\Constraints as Assert;
-use Symfony\Component\Validator\ConstraintViolation;
 use Symfony\Component\Validator\ConstraintViolationList;
 use Symfony\Component\Validator\Exception\ValidationFailedException;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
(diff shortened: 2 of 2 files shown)


mtarld (approved):

src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/RequestPayloadValueResolver.php
```diff
@@ -119,7 +119,7 @@ public function onKernelControllerArguments(ControllerArgumentsEvent $event): vo
                     $payload = $e->getData();
                 }
 
-                if (null !== $payload) {
+                if (null !== $payload && !\count($violations)) {
```
nicolas-grekas, line 122: shouldn't we move the next "if" before this one instead?

nicolas-grekas: nvm, we need to throw in both cases

nicolas-grekas:
Thank you @HypeMC.

PR #45306, opened 2022-02-04
[PropertyAccessor] Add missing TypeError catch

b1rdex:
| Q             | A
| ------------- | ---
| Branch?       | 4.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #45199 
| License       | MIT

This bugfix wraps `TypeError: Cannot assign TYPE to property *::$* of type TYPE` with `InvalidArgumentException`.

PR #39399, opened 2020-12-09
[Serializer] Allow to provide (de)normalization context in mapping

ogizanagi:
| Q             | A
| ------------- | ---
| Branch?       | 5.x
| Bug fix?      | no
| New feature?  | yes 
| Deprecations? | no 
| Tickets       | Fix #39039 
| License       | MIT
| Doc PR        | TODO 

As explained in the linked feature request, this brings the ability to configure context on a per-property basis, using Serializer mapping.

Considering:

```php
use Symfony\Component\Serializer\Annotation as Serializer;
use Symfony\Component\Serializer\Normalizer\DateTimeNormalizer;

class Foo
{
    /**
     * @Serializer\Context({ DateTimeNormalizer::FORMAT_KEY = 'Y-m-d' })
     */
    public \DateTime $date;

    public \DateTime $anotherDate;
}
```

`$date` will be formatted with a specific format, while `$anotherDate` will use the default configured one (or the one provided in the context while calling `->serialize()` / `->normalize()`).

It can also differentiate normalization and denormalization contexts:

```php
use Symfony\Component\Serializer\Annotation as Serializer;
use Symfony\Component\Serializer\Normalizer\DateTimeNormalizer;

class Foo
{
    /**
     * @Serializer\Context(
     *   normalizationContext = { DateTimeNormalizer::FORMAT_KEY = 'Y-m-d' },
     *   denormalizationContext = { DateTimeNormalizer::FORMAT_KEY = \DateTime::COOKIE },
     * )
     */
    public \DateTime $date;
}
```

As well as act differently depending on groups:

```php
use Symfony\Component\Serializer\Annotation as Serializer;
use Symfony\Component\Serializer\Normalizer\DateTimeNormalizer;

class Foo
{
    /**
     * @Serializer\Groups({ "extended" })
     * @Serializer\Context({ DateTimeNormalizer::FORMAT_KEY = \DateTime::RFC3339 })
     * @Serializer\Context(
     *   context = { DateTimeNormalizer::FORMAT_KEY = \DateTime::RFC3339_EXTENDED },
     *   groups = {"extended"},
     * )
     */
    public \DateTime $date;
}
```

The annotation can be repeated as much as you want to handle the different cases.
Context without groups is always applied first, then context for groups are merged in the provided order.
Context provided when calling `->serialize()` / `->normalize()` acts as the defaults for the properties without context provided in the metadata.

XML mapping (see tests) is a lot verbose due to the required structure to handle groups.

[...]

PR #53108, opened 2023-12-17
[Serializer] Fix using deserialization path 5.4

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

I've noticed this while working on #53107.

Sometimes, the `path` provided to the `NotNormalizableValueException` is incorrect.

### DIFF
--- a/src/Symfony/Component/Serializer/composer.json
+++ b/src/Symfony/Component/Serializer/composer.json
@@ -33,7 +33,7 @@
         "symfony/http-foundation": "^4.4|^5.0|^6.0",
         "symfony/http-kernel": "^4.4|^5.0|^6.0",
         "symfony/mime": "^4.4|^5.0|^6.0",
-        "symfony/property-access": "^5.4|^6.0",
+        "symfony/property-access": "^5.4.26|^6.3",
         "symfony/property-info": "^5.4.24|^6.2.11",
         "symfony/uid": "^5.3|^6.0",
         "symfony/validator": "^4.4|^5.0|^6.0",
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
@@ -414,7 +414,7 @@ abstract class AbstractNormalizer implements NormalizerInterface, DenormalizerIn
                         sprintf('Failed to create object because the class misses the "%s" property.', $constructorParameter->name),
                         $data,
                         ['unknown'],
-                        $objectDeserializationPath,
+                        $context['deserialization_path'],
                         true
                     );
                     $context['not_normalizable_value_exceptions'][] = $exception;
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
@@ -437,7 +437,7 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
                     sprintf('Failed to denormalize attribute "%s" value for class "%s": '.$e->getMessage(), $attribute, $type),
                     $data,
                     ['unknown'],
-                    $context['deserialization_path'] ?? null,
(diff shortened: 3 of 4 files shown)


nicolas-grekas:
Any opinion on this @dunglas @lyrixx @mtarld or anyone else?

src/Symfony/Component/Serializer/composer.json
```diff
@@ -33,7 +33,7 @@
         "symfony/http-foundation": "^4.4|^5.0|^6.0",
         "symfony/http-kernel": "^4.4|^5.0|^6.0",
         "symfony/mime": "^4.4|^5.0|^6.0",
-        "symfony/property-access": "^5.4|^6.0",
```
mtarld, line 36: Is that bump needed?

mtarld (commented):
Actually, before that PR, paths related to constructor weren't taken into account. And it does look good to me to be more precise in that case

src/Symfony/Component/Serializer/composer.json
HypeMC: Yes, otherwise lowest deps fail.

src/Symfony/Component/Serializer/composer.json
```diff
@@ -33,7 +33,7 @@
...
+        "symfony/property-access": "^5.4.4|^6.0.4",
```
chalasr: Can you elaborate on why these versions precisely? The next bugfix release for property-access (supposedly embedding this patch) will be v5.4.27 so I suppose this targets a specific change?

HypeMC: The reason is #45306. Without it the new test fails:

```
1) Symfony\Component\Serializer\Tests\SerializerTest::testCollectDenormalizationErrorsWithoutTypeExtractor
Failed asserting that TypeError Object (...) is an instance of class "Symfony\Component\Serializer\Exception\PartialDenormalizationException".
```

HypeMC: I just noticed that 5.4.4 and 6.0.4 don't exist, so 5.4.5 and 6.0.5 are actually installed instead.

nicolas-grekas:
> Actually, before that PR, paths related to constructor weren't taken into account. And it does look good to me to be more precise in that case

Should we interpret this as an improvement then? For 7.1?

HypeMC:
> > Actually, before that PR, paths related to constructor weren't taken into account. And it does look good to me to be more precise in that case
> 
> Should we interpret this as an improvement then? For 7.1?

@nicolas-grekas To me, this seems like a bug. As far as I can tell, the `$attributeContext` variable was introduced in #39399 for version 5.3, whereas support for collecting type errors was introduced in version 5.4 with #42502. It appears that the wrong variable might have been used by accident :shrug:

HypeMC:
There's also the fact that the messages actually have the correct paths in them:

```diff
            [
...
-                'path' => null,
+                'path' => 'string',
                 'message' => 'Failed to create object because the class misses the "string" property.',
            ],
```

mtarld:
I agree with @HypeMC, it can be considered as a bugfix in my opinion

src/Symfony/Component/Serializer/composer.json
```diff
@@ -33,7 +33,7 @@
         "symfony/http-foundation": "^4.4|^5.0|^6.0",
         "symfony/http-kernel": "^4.4|^5.0|^6.0",
         "symfony/mime": "^4.4|^5.0|^6.0",
-        "symfony/property-access": "^5.4|^6.0",
+        "symfony/property-access": "^5.4.5|^6.0.5",
```
chalasr:
```suggestion
        "symfony/property-access": "^5.4.26|^6.3",
```

HypeMC: Done

mtarld (approved):

chalasr (approved):

nicolas-grekas:
Thank you @HypeMC.

PR #53121, opened 2023-12-18
[Clock] Add static `DatePoint::create` constructor

ruudk:
| Q             | A
| ------------- | ---
| Branch?       | 7.1 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Issues        | 
| License       | MIT

This makes it easier to do things like `DatePoint::create('now')->format('Y-m-d')` instead of having to type `(new DatePoint('now'))->format('Y-m-d')`.

### DIFF
--- a/src/Symfony/Component/Clock/CHANGELOG.md
+++ b/src/Symfony/Component/Clock/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Add static `create` method to `DatePoint` to allow chaining of methods
+
 6.4
 ---
 
--- a/src/Symfony/Component/Clock/DatePoint.php
+++ b/src/Symfony/Component/Clock/DatePoint.php
@@ -49,6 +49,14 @@ final class DatePoint extends \DateTimeImmutable
     }
 
     /**
+     * @throws \DateMalformedStringException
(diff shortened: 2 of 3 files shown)


GromNaN:
This class was introduced mainly to be used as default value in method declaration.
You should use the global `now()` function or `Clock::get()->now()` as factory.

ruudk:
From the docs or PR it reads that this is a drop-in replacement for DateTime/DateTimeImmutable. So why do I need a clock? I think it's the beauty that I don't have to think about the Clock, and just get a DateTimeImmutable.

nicolas-grekas:
Not an objection to this PR, but did you consider using `now()` instead?
`now()->format('Y-m-d')`

derrabus:
In the old days, those `::create()` methods were a workaround because PHP did not allow to dereference the result of a constructor call (`(new SomeClass())->someMethod()`). This changed in some PHP 5.x release which is why we've removed some of the methods already. I would strongly advice against adding new onces. This pattern has been legacy for decades now.

GromNaN:
Closing for the reason exposed by @derrabus. Thank you for your suggestion. I hope you found the answers helpful.

ruudk:
Thanks for the comments. I will try to use the `now` function.

PR #53123, opened 2023-12-18
[VarDumper] Added default message for dd function

shamimulalam:
| Q             | A
| ------------- | ---
| Branch?       | 7.1 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Issues        | 
| License       | MIT

This will facilitate writing empty dd() statements during unorthodox debugging in the code. It will be helpful for both novice and experienced developers.

### DIFF
--- a/src/Symfony/Component/VarDumper/Resources/functions/dump.php
+++ b/src/Symfony/Component/VarDumper/Resources/functions/dump.php
@@ -49,6 +49,12 @@ if (!function_exists('dd')) {
             header('HTTP/1.1 500 Internal Server Error');
         }
 
+        if (!$vars) {
+            VarDumper::dump(new ScalarStub('🐛'));
+
+            exit(1);
+        }
+
         if (array_key_exists(0, $vars) && 1 === count($vars)) {
             VarDumper::dump($vars[0]);
         } else {


src/Symfony/Component/VarDumper/Resources/functions/dump.php
```diff
@@ -49,6 +49,12 @@ function dd(mixed ...$vars): never
             header('HTTP/1.1 500 Internal Server Error');
         }
 
+        if (empty($vars)) {
+            VarDumper::dump('Dumped!');
```
alexandre-daubois: Should we use 🐛 like in the `dump()` function?

shamimulalam: It's not bug! So i think, we shouldn't use `🐛` this. We can use different emoji. What about 🐞 this? Or can you suggest something else?

nicolas-grekas: we should be consistent with what we do in dump:
```php
VarDumper::dump(new ScalarStub('🐛'))
```

src/Symfony/Component/VarDumper/Resources/functions/dump.php
```diff
@@ -49,6 +49,12 @@ function dd(mixed ...$vars): never
             header('HTTP/1.1 500 Internal Server Error');
         }
 
+        if (empty($vars)) {
```
nicolas-grekas:
```suggestion
        if (!$vars) {
```

nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @shamimulalam.

PR #53126, opened 2023-12-18
[Console] Add env var to disable direct output

### LINKED ISSUE #52777: Allow disabling exception rendering
### Description

When using Datadog and Docker, it is recommended to only ever print JSON (for Datadog) to the output (because the output of the main process is the output of the whole container). Datadog documents how to properly configure a Symfony application to fulfill that requirement: https://docs.datadoghq.com/logs/log_collection/php/?tab=phpmonolog

They don't mention it, but you should probably also use `--quiet` when running a command in production, so that only the json logs show up in the output.

There's one thing that you can't however prevent, and that is the exception rendering done by https://github.com/symfony/symfony/blob/c8b962a0bf4ec24ef405222d66f3ae8e9eea94f6/src/Symfony/Component/Console/Application.php#L820-L830

Because of it, what you will see on Datadog is:
- a properly formatted, CRITICAL datadog line thanks to the monolog configuration
- one INFO line for each line in the unformatted output. 

### Example

It might be nice to have something like 
```yaml
framework:
    render_errors: '%kernel.debug%'
```
to configure this.

wouterj:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #52777 
| License       | MIT

This is useful when running in a containerized application and you've configured Monolog to write to stdout (probably using a JSON format).

Alternatively, we can also introduce a `LoggerOutput` that writes each line to an INFO log message. This can result in double messages, but has the benefit that commands don't have to be written so that useful information is logged. What do you think of this?

### DIFF
--- a/src/Symfony/Component/Console/Application.php
+++ b/src/Symfony/Component/Console/Application.php
@@ -48,6 +48,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\ConsoleOutput;
 use Symfony\Component\Console\Output\ConsoleOutputInterface;
+use Symfony\Component\Console\Output\NullOutput;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\SignalRegistry\SignalRegistry;
 use Symfony\Component\Console\Style\SymfonyStyle;
@@ -145,7 +146,7 @@ class Application implements ResetInterface
         }
 
         $input ??= new ArgvInput();
-        $output ??= new ConsoleOutput();
+        $output ??= $this->isDirectOutputDisabled() ? new NullOutput() : new ConsoleOutput();
 
         $renderException = function (\Throwable $e) use ($output) {
             if ($output instanceof ConsoleOutputInterface) {
@@ -934,7 +935,7 @@ class Application implements ResetInterface
                 break;
         }
 
-        if (true === $input->hasParameterOption(['--quiet', '-q'], true)) {
+        if (true === $input->hasParameterOption(['--quiet', '-q'], true) || $this->isDirectOutputDisabled()) {
             $output->setVerbosity(OutputInterface::VERBOSITY_QUIET);
             $shellVerbosity = -1;
         } else {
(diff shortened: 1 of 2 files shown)


ro0NL:
what's the difference with SHELL_VERBOSITY=-1? https://symfony.com/doc/current/console/verbosity.html

ro0NL:
Also, AFAIK the trick in docker envs et al is to use different process: https://github.com/relaxart/monolog-docker-handler/blob/master/src/DockerMonologHandler.php

wouterj:
> what's the difference with SHELL_VERBOSITY=-1?

As indicated in the issue, we have some rendering that still outputs with quiet/verbosity=-1 (e.g. errors).

src/Symfony/Component/Console/Application.php
```diff
@@ -934,7 +935,7 @@ protected function configureIO(InputInterface $input, OutputInterface $output):
                 break;
         }
 
-        if (true === $input->hasParameterOption(['--quiet', '-q'], true)) {
+        if (true === $input->hasParameterOption(['--quiet', '-q'], true) || $this->isDirectOutputDisabled()) {
```
ro0NL, line 938: the same code path can be triggered using SHELL_VERBOSITY=-1 isnt it?

https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Console/Application.php#L919-L921

wouterj: Yes, this was more as a shortcut so `SYMFONY_CONSOLE_OUTPUT=0` also implies verbosity -1

I would also be fine with not rendering throwables/errors when `SHELL_VERBOSITY=-1` in the `Application` class if that's preferred and not considered a BC break.

ro0NL: i'd like to avoid SYMFONY_CONSOLE_OUTPUT vs SHELL_VERBOSITY yes :)

ro0NL: i understand the BC concern, perhaps we need to reconsider what --quiet really means :sweat_smile: 

SHELL_VERBOSITY=-2 maybe?

chalasr: I agree that expanding `SHELL_VERBOSITY` modes would be better than a new env var.

> SHELL_VERBOSITY=-2 maybe?

👍 and `--silent` maybe

ro0NL: i'd consider making -q really quiet

i see no reason to preserve just error output with -q (assuming error logging is setup)

relying on humans to look at output, is error prone anyway

wouterj: I've submitted the alternative in a new PR (to not mix up the discussions): #53632

PR #53128, opened 2023-12-18
Change ProxyCacheWarmer::warmUp signature

llupa:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | none exists
| License       | MIT

I ended up with a composer.lock that has `"symfony/doctrine-bridge:v6.4.0"` and `"symfony/http-kernel:v7.0.1"`, this combination results in a fatal syntax error: 

```php
Compile Error:
 Declaration of Symfony\Bridge\Doctrine\CacheWarmer\ProxyCacheWarmer::warmUp(string $cacheDir): array 
 must be compatible with 
 Symfony\Component\HttpKernel\CacheWarmer\WarmableInterface::warmUp(string $cacheDir, ?string $buildDir = null): array
```

I opened this PR as a starting point to "fix" this, I already wondered if the conflict should also be put over `doctrine-bridge`. Happy to do changes as needed.

### DIFF
--- a/UPGRADE-6.4.md
+++ b/UPGRADE-6.4.md
@@ -96,6 +96,7 @@ DependencyInjection
 DoctrineBridge
 --------------
 
+ * [BC Break] Add argument `$buildDir` to `ProxyCacheWarmer::warmUp()` 
  * [BC Break] Add return type-hints to `EntityFactory`
  * Deprecate `DbalLogger`, use a middleware instead
  * Deprecate not constructing `DoctrineDataCollector` with an instance of `DebugDataHolder`
--- a/src/Symfony/Bridge/Doctrine/CHANGELOG.md
+++ b/src/Symfony/Bridge/Doctrine/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 6.4
 ---
 
+ * [BC BREAK] Add argument `$buildDir` to `ProxyCacheWarmer::warmUp()` 
  * [BC BREAK] Add return type-hints to `EntityFactory`
  * Deprecate `DbalLogger`, use a middleware instead
  * Deprecate not constructing `DoctrineDataCollector` with an instance of `DebugDataHolder`
--- a/src/Symfony/Bridge/Doctrine/CacheWarmer/ProxyCacheWarmer.php
+++ b/src/Symfony/Bridge/Doctrine/CacheWarmer/ProxyCacheWarmer.php
@@ -37,10 +37,7 @@ class ProxyCacheWarmer implements CacheWarmerInterface
         return false;
     }
 
-    /**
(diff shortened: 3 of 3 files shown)


nicolas-grekas:
This won't fix the issue since the faulty combination is still allowed, so that composer can still select it, isn't it?

What about making the bridge compatible with 7.0 instead? This requires a BC break, but it shouldn't affect anybody IMHO: that class/method isn't extended by anyone :crossed_fingers: 
WDYT @symfony/mergers ?

llupa:
I thought about updating `Symfony\Bridge\Doctrine\CacheWarmer\ProxyCacheWarmer::warmUp` too and this would be best from my POV, but wouldn't that mean you will have anyway a conflict between doctrine-bridge 6.4.0 and htt-kernel 7.x.x ? The fix would go to doctrine-bridge 6.4.1 as I understand it.

nicolas-grekas:
> have anyway a conflict between doctrine-bridge 6.4.0 and htt-kernel 7.x.x ?

no need to since .0 will be replaced by .1

xabbuh:
> What about making the bridge compatible with 7.0 instead? This requires a BC break, but it shouldn't affect anybody IMHO: that class/method isn't extended by anyone

works for me, we already did that in the Serializer and Translation components in #50391

llupa:
I am not sure if I completely broke the process, but, I tried my best with what Nicolas and Christian said 😅

GromNaN:
> it shouldn't affect anybody IMHO: that class/method isn't extended by anyone 🤞 WDYT 

Agreed, we should make this classes final in 7.1

UPGRADE-6.4.md
```diff
@@ -96,6 +96,7 @@ DependencyInjection
 DoctrineBridge
 --------------
 
+ * [BC BREAK] Change `ProxyCacheWarmer::warmUp` signature
```
```suggestion
 * [BC Break] Add argument `$buildDir` to `ProxyCacheWarmer::warmUp()` 
```

src/Symfony/Bridge/Doctrine/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ CHANGELOG
 6.4
 ---
 
+ * [BC BREAK] Change `ProxyCacheWarmer::warmUp` signature
```
nicolas-grekas:
```suggestion
 * [BC BREAK] Add argument `$buildDir` to `ProxyCacheWarmer::warmUp()` 
```

nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @llupa.

PR #53133, opened 2023-12-19
[Validator] Fix using known option names as field names

### LINKED ISSUE #52993: Constraints Collection fields with key fields

6.4

### Description

With this assert
```php
#[Assert\Collection(
    fields: [
        'fields' => [new Assert\Type(type: 'array')],
        'properties' => [new Assert\Type(type: 'array')],
        'catalog' => [new Assert\Type(type: 'array')],
    ],
)]
private array $configuration;
```

i got 

```
The options "properties", "catalog" do not exist in constraint "Symfony\Component\Validator\Constraints\Collection".
Symfony\Component\Validator\Exception\InvalidOptionsException: The options "properties", "catalog" do not exist in constraint "Symfony\Component\Validator\Constraints\Collection". in /srv/exmo/book24/vendor/symfony/validator/Constraint.php:182
```

but if i rename 'fields' in 'fileds2', the problem will disappear.

### How to reproduce
```php
<?php
require_once __DIR__.'/vendor/autoload.php';
use Symfony\Component\Validator\Constraints;
use Symfony\Component\Validator\Constraints as Assert;


$constraint = new Constraints\Collection(
    fields: [
        'fields' => [new Assert\Type(type: 'array')],
        'properties' => [new Assert\Type(type: 'array')],
        'catalog' => [new Assert\Type(type: 'array')],
    ]
);
$validator = new Assert\CollectionValidator();

var_dump(
    $validator->validate(
        [
            'fields' => [],
            'properties' => [],
            'catalog' => [],
        ],
        $constraint
    )
);
```
### Possible Solution

_No response_

### Additional Context

_No response_

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52993
| License       | MIT

Instead of assuming `$fields` is not the fields array when any of the keys is a known option, let's assume it is the fields array if all keys are known options. Of course, this approach won't work if someone names all their fields after known options, but I don't believe that will actually happen.

### DIFF
--- a/src/Symfony/Component/Validator/Constraints/Collection.php
+++ b/src/Symfony/Component/Validator/Constraints/Collection.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Component\Validator\Constraints;
 
+use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\Exception\ConstraintDefinitionException;
 
 /**
@@ -41,9 +42,10 @@ class Collection extends Composite
      */
     public function __construct($fields = null, array $groups = null, $payload = null, bool $allowExtraFields = null, bool $allowMissingFields = null, string $extraFieldsMessage = null, string $missingFieldsMessage = null)
     {
-        // no known options set? $fields is the fields array
         if (\is_array($fields)
-            && !array_intersect(array_keys($fields), ['groups', 'fields', 'allowExtraFields', 'allowMissingFields', 'extraFieldsMessage', 'missingFieldsMessage'])) {
+            && (($firstField = reset($fields)) instanceof Constraint
+                || ($firstField[0] ?? null) instanceof Constraint
+            )) {
             $fields = ['fields' => $fields];
         }
 
--- a/src/Symfony/Component/Validator/Tests/Constraints/CollectionTest.php
+++ b/src/Symfony/Component/Validator/Tests/Constraints/CollectionTest.php
@@ -18,6 +18,7 @@ use Symfony\Component\Validator\Constraints\Optional;
 use Symfony\Component\Validator\Constraints\Required;
 use Symfony\Component\Validator\Constraints\Valid;
 use Symfony\Component\Validator\Exception\ConstraintDefinitionException;
+use Symfony\Component\Validator\Exception\InvalidOptionsException;
 
 /**
  * @author Bernhard Schussek <bschussek@gmail.com>
@@ -34,7 +35,7 @@ class CollectionTest extends TestCase
 
     public function testRejectNonConstraints()
     {
-        $this->expectException(ConstraintDefinitionException::class);
+        $this->expectException(InvalidOptionsException::class);
         new Collection([
             'foo' => 'bar',
         ]);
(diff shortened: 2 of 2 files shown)


src/Symfony/Component/Validator/Constraints/Collection.php
```diff
@@ -41,9 +41,12 @@ class Collection extends Composite
      */
     public function __construct($fields = null, array $groups = null, $payload = null, bool $allowExtraFields = null, bool $allowMissingFields = null, string $extraFieldsMessage = null, string $missingFieldsMessage = null)
     {
-        // no known options set? $fields is the fields array
-        if (\is_array($fields)
-            && !array_intersect(array_keys($fields), ['groups', 'fields', 'allowExtraFields', 'allowMissingFields', 'extraFieldsMessage', 'missingFieldsMessage'])) {
```
HypeMC, line 46: I'm pretty sure `payload` was missing in the array of known options. Switching to a dynamic approach using reflection should solve this and any similar issues that may occur in the future.

src/Symfony/Component/Validator/Constraints/Collection.php
```diff
@@ -41,9 +41,12 @@ class Collection extends Composite
...
+        $knowOptions = array_map(function (\ReflectionParameter $parameter) {
+            return $parameter->name;
+        }, (new \ReflectionMethod(__METHOD__))->getParameters());
```
nicolas-grekas:
```suggestion
        static $knowOptions;
        $knownOptions ??= array_column((new \ReflectionMethod(__METHOD__))->getParameters(), 'name');
```

src/Symfony/Component/Validator/Constraints/Collection.php
xabbuh: What about picking the first item of `$fields` instead and check if it is a `Constraint` instance (or an array of)?

HypeMC: Won't work cause of https://github.com/symfony/symfony/blob/ddf3bcbbb2ec1b4110b9998b2414dee2091def36/src/Symfony/Component/Validator/Constraints/Collection.php#L69-L74

src/Symfony/Component/Validator/Constraints/Collection.php
HypeMC: Done, I had to adjust the code for PHP 7.2. I'm wondering, is the static variable necessary since this all happens in the constructor?

src/Symfony/Component/Validator/Constraints/Collection.php
xabbuh: I am not sure I understand. Checking for instances of `Constraint` would still yield `true`, wouldn't it?

HypeMC: Yes, I don't recall why I initially discarded the suggestion, but it works. I did have to modify one test because a different exception is thrown now. I'm not sure if that qualifies as a BC break.

xabbuh (approved):

fabpot (approved):

fabpot:
Thank you @HypeMC.

tscni:
Fyi, this breaks code where `$fields` is an empty array

xabbuh:
@tscni Can you please try #53383?

dresslerdemos:
I tried this (https://github.com/symfony/symfony/pull/53383) fix and it still breaks for an empty list of constraints, which I confirmed was working in `v5.4.29`:

```
    public function testEmptyConstraintList()
    {
        // Symfony\Component\Validator\Exception\InvalidOptionsException : The options "foo" do not exist in constraint "Symfony\Component\Validator\Constraints\Collection".
        new Collection(
            [
                'foo' => []
            ],
            null,
            null,
            true,
            false
        );

        self::assertTrue(true);
    }
```

For my use case I can mitigate this by using `fields`, but I wanted to give a heads up regarding backward compatibility.

```
    public function testEmptyConstraintList2()
    {
        new Collection(
            ['fields' => [
                'foo' => []
            ]],
            null,
            null,
            true,
            false
        );

        self::assertTrue(true);
    }
```

nerones:
just had the same problem as @dresslerdemos

xabbuh:
I have a fix pending locally. I’ll keep you posted.

xabbuh:
@dresslerdemos @nerones Can you check if #53443 fixes it for you?

PR #53140, opened 2023-12-19
[Serializer] Skip uninitialized properties with deep_object_to_populate

### LINKED ISSUE #52883: [Serializer] Typed properties Error when denormalizing with deep_object_to_populate true

6.3.9

### Description

Imagine two classes, both with typed properties and without default values.
First class has an array property of second classes.

Denormalizing into an object with object_to_populate, you could have a case when an array item is being instantiated by the serializer, because it could not be found in the object_to_populate.

In this case with deep_object_to_populate true it will try to read from that freshly instantiated object and fail because typed property is not initialized.

The same case is fixed with normalization: #40578 #41615

### How to reproduce

```php
class DataClass
{
    public string $unInitialized;
}

$object = new DataClass();

$normalizer = new ObjectNormalizer();
$normalizer->denormalize(
    ['unInitialized' => 'something'],
    DataClass::class,
    context: [
        'object_to_populate' => $object,
        'deep_object_to_populate' => true,
    ]
);
```

### Possible Solution

Make option `skip_uninitialized_values` work when denormalizing.

### Additional Context

Inspired from comment https://github.com/symfony/symfony/issues/40578#issuecomment-1177533726

mtarld:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #52883
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Serializer/Tests/Normalizer/Features/SkipUninitializedValuesTestTrait.php
+++ b/src/Symfony/Component/Serializer/Tests/Normalizer/Features/SkipUninitializedValuesTestTrait.php
@@ -12,14 +12,14 @@
 namespace Symfony\Component\Serializer\Tests\Normalizer\Features;
 
 use Symfony\Component\PropertyAccess\Exception\UninitializedPropertyException;
-use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+use Symfony\Component\Serializer\Normalizer\AbstractObjectNormalizer;
 
 /**
  * Test AbstractObjectNormalizer::SKIP_UNINITIALIZED_VALUES.
  */
 trait SkipUninitializedValuesTestTrait
 {
-    abstract protected function getNormalizerForSkipUninitializedValues(): NormalizerInterface;
+    abstract protected function getNormalizerForSkipUninitializedValues(): AbstractObjectNormalizer;
 
     /**
      * @requires PHP 7.4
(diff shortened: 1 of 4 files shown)


dunglas (approved):

src/Symfony/Component/Serializer/Tests/Normalizer/PropertyNormalizerTest.php
```diff
@@ -455,7 +455,7 @@ protected function getNormalizerForCacheableObjectAttributesTest(): AbstractObje
         return new PropertyNormalizer();
     }
 
-    protected function getNormalizerForSkipUninitializedValues(): NormalizerInterface
```
nicolas-grekas, line 458: we don't need the corresponding `use` anymore (spotted by fabbot)

src/Symfony/Component/Serializer/Tests/Normalizer/Features/SkipUninitializedValuesTestTrait.php
```diff
@@ -33,6 +33,15 @@ public function testSkipUninitializedValues(array $context)
         $normalizer = $this->getNormalizerForSkipUninitializedValues();
         $result = $normalizer->normalize($object, null, $context);
         $this->assertSame(['initialized' => 'value'], $result);
+
+        $normalizer->denormalize(
+            ['unInitialized' => 'value'],
+            TypedPropertiesObjectWithGetters::class,
+            null,
+            ['object_to_populate' => $objectToPopulate = new TypedPropertiesObjectWithGetters(), 'deep_object_to_populate' => true] + $context,
```
nicolas-grekas: bad trailing comma for php 7.2 (see CI also)

nicolas-grekas:
Thank you @mtarld.

PR #53146, opened 2023-12-19
[Dependency Injection] Adding a 'date' prefix in EnvVarProcessor

ktherage:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        |
| License       | MIT

Adding a 'date' prefix in EnvVarProcessor that will return a DateTimeImmutable from a valid date format.

Example:

```shell
# .env
LIMIT_DATE='2024-01-01 14:30:00'
```

```yaml
# service.yaml
parameters:
    limit_date: "%env(date:LIMIT_DATE)%"
```

```php
$date = $container->getParameter('limit_date');

// date will be (timezone will be server dependent on server's configuration) : 
// object(DateTimeImmutable)#1 (3) {
//     ["date"]=> string(26) "2024-01-01 14:30:00.000000"
//     ["timezone_type"]=> int(3)
//     ["timezone"]=> string(16) "Europe/Amsterdam"
// }
```


src/Symfony/Component/DependencyInjection/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ CHANGELOG
 7.1
 ---
 
+ * Add method `date` prefix to `EnvVarProcessor` that allows get a DateTimeImmutable from an env var
```
alexandre-daubois:
```suggestion
 * Add `date` env var processor
```

To keep consistency with the other entries adding a processor? :slightly_smiling_face:

ktherage: nice catch :wink: thanks.

nicolas-grekas:
This was proposed before as https://github.com/symfony/symfony/pull/51575, but closed after https://github.com/symfony/symfony/pull/51575#pullrequestreview-1685475826

Is your approach any different?

ktherage:
> This was proposed before as #51575, but closed after [#51575 (review)](https://github.com/symfony/symfony/pull/51575#pullrequestreview-1685475826)
> 
> Is your approach any different?

Not that much in fact, I was not supporting the date format part of that PR but allowing it simply by taking the env var value and passing it to the DateTimeImmutable constructor.

I guess that I shall close that one too so ?

> Instead of adding a new processor, we could consider providing a simpler way to do such declarations. If an attribute is enough, it shouldn't be too hard to add support for eg #[AutowireDateTime(etc)].

Following your review, do you want me to work on that instead ?

ro0NL:
i dont see how "AutowireDateTime" is simpler then datetime processor :sweat_smile:

nicolas-grekas:
Feel free to propose once #52820 is merged of course.

PR #53147, opened 2023-12-19
[VarDumper] Fix CSS alignment in HtmlDumper

### LINKED ISSUE #53103: [Profiler] Some class names are not aligned correctly

6.4.1

### Description

I noticed that some names of classes inside the "Messages" profiler are misaligned, see the attached screenshot

Schermata del 2023-12-17 11-44-05

You can notice that the grey part of the FQCN is too high, compared to the green part.
I noticed this value in the CSS:

```css
pre.sf-dump .sf-dump-ellipsis {
  vertical-align: top;
}
```

if I manually change it to `vertical-align: middle;`, the line seems correctly aligned. I didn't check if this change could affect other parts of the profiler, though.
This is the result after applying the change:

Schermata del 2023-12-17 11-52-11

### How to reproduce

Dispatch a message with the Messenger and open the profiler

### Possible Solution

As mentioned above, changing the vertical alignment could be a possible solution

### Additional Context

_No response_

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #53103
| License       | MIT

Bottom alignment seems the right one here:

image

### DIFF
--- a/src/Symfony/Component/VarDumper/Dumper/HtmlDumper.php
+++ b/src/Symfony/Component/VarDumper/Dumper/HtmlDumper.php
@@ -663,7 +663,7 @@ pre.sf-dump:after {
    clear: both;
 }
 pre.sf-dump span {
-    display: inline;
+    display: inline-flex;
 }
 pre.sf-dump a {
     text-decoration: none;


Nyholm:
Thank you. 

CSS is tricky and I am no expert. But why does `bottom` work and not `top`? How come `middle` doesn't do the trick? 

What I really wonder is: Is this according to the official rules and have this been tested in more than one browser?

stof:
Intuitively, I would say that we want to align the baseline of the text here, so using `baseline`

alexandre-daubois:
It depends on the "bounding box" of the element. The FQCN is actually two parts here: the grey and the green one.

The green part of the FQCN is already centered in its section. If the bounding-box is a bit taller than the text, then the "vertical-align: top" of the grey part pushes it on top of the bounding box. Not sure I'm super clear here :sweat_smile: 

@stof I'm having a look at this solution

alexandre-daubois:
Unfortunately, baseline doesn't work well:

image

MatTheCat:
It’s because `sf-dump-ellipsis` falls in the following case (from https://www.w3.org/TR/css-align-3/#baseline-export):

> for legacy reasons a [block-level](https://www.w3.org/TR/css-display-3/#block-level) or [inline-level](https://www.w3.org/TR/css-display-3/#inline-level) [scroll container](https://www.w3.org/TR/css-overflow-3/#scroll-container) always has a [last baseline set](https://www.w3.org/TR/css-align-3/#last-baseline-set), whose baselines all correspond to its [block-end](https://www.w3.org/TR/css-writing-modes-4/#block-end) [margin edge](https://www.w3.org/TR/css-box-4/#margin-edge).

`vertical-align: bottom` will align this baseline with the line-box bottom, thus making it appear like baselines are aligned.

Alternatively, `vertical-align: top` could be replaced by `baseline-source: first` (or `last`, as there will always be one line).

javiereguiluz:
I've played a bit with this and to me, this was solved by doing this change instead:

```diff
 pre.sf-dump span {
-    display: inline;
+    display: inline-flex;
 }
```

Can you please try if this also solves the design issues for you? Thanks!

alexandre-daubois:
I can confirm this works well. Thanks Javier, PR updated!

javiereguiluz (approved):

Nyholm:
Thank you @alexandre-daubois.

PR #42088, opened 2021-07-13
[Contracts] add return types and bump to v3

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 6.0
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | #42058
| License       | MIT
| Doc PR        | -

PR #53150, opened 2023-12-19
[Contracts] Allow psr/container 1.1 again

derrabus:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | N/A
| License       | MIT

The only difference between psr/container 1.1 and 2.0 is one additional return type on the `ContainerInterface::has()` method, see https://github.com/php-fig/container/compare/1.1.2...2.0.2

This means that if a package is compatible with 2.0 already (== all implementations already have that return type), it remains compatible with 1.1. However, in #42088 we changed the composer.json of our service-contracts package and replaced `^1.1` with `^2.0`, dropping support for 1.1 although there was no need to do so.

I'm migrating a larger application to Symfony 7 at the moment which also consumes various Laminas packages. Due to Laminas still being stuck on psr/container 1.1, I cannot upgrade to service-contracts 3 which in turn blocks my Symfony 7 upgrade.

It would smoothen the upgrade path a lot for me, if we released a service-contracts 3.x that allows the installation of psr/container 1.1 again.

### DIFF
--- a/src/Symfony/Contracts/Service/composer.json
+++ b/src/Symfony/Contracts/Service/composer.json
@@ -17,7 +17,7 @@
     ],
     "require": {
         "php": ">=8.1",
-        "psr/container": "^2.0"
+        "psr/container": "^1.1|^2.0"
     },
     "conflict": {
         "ext-psr": "<1.1|>=2"


bendavies:
related:
https://github.com/symfony/symfony/pull/51832#discussion_r1354705072
which was followed by 
https://github.com/symfony/symfony/pull/51916/files?diff=unified&w=0#r1356756693

xabbuh:
For 6.3?

nicolas-grekas (approved):
Works for me. 3.3 or 3.4 won't make a difference so I'm fine targeting 6.4 (=3.4)

xabbuh (approved):
got it now, the change happens in the contracts which have a different version schema than the rest of the components

derrabus:
Thank you very much for the quick approval. 🚀

derrabus:
> Works for me. 3.3 or 3.4 won't make a difference so I'm fine targeting 6.4 (=3.4)

The change does not appear in the split repository: https://github.com/symfony/service-contracts/branches/all

Did we target the wrong branch or is there something wrong with the splitter?

nicolas-grekas:
I think we might need to merge up

xabbuh:
> I think we might need to merge up

did that, but that doesn't seem to be enough (or the sync takes longer than I expected)

derrabus:
The change _does_ appear on the contracts repository however: https://github.com/symfony/contracts/commit/bd6978fb5cb7e858e26bdaf9c7a261e7087570f8. But apparently, it does not travel further down to the service-contracts package from there. 😕

derrabus:
Problem solved: https://github.com/symfony/service-contracts/commit/11bbf19a0fb7b36345861e85c5768844c552906e 🥳

PR #53151, opened 2023-12-20
mark classes implementing the `WarmableInterface` as `final`

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

### DIFF
--- a/UPGRADE-7.1.md
+++ b/UPGRADE-7.1.md
@@ -6,10 +6,30 @@ Cache
 
  * Deprecate `CouchbaseBucketAdapter`, use `CouchbaseCollectionAdapter` instead
 
+FrameworkBundle
+---------------
+
+ * Mark classes `ConfigBuilderCacheWarmer`, `Router`, `SerializerCacheWarmer`, `TranslationsCacheWarmer`, `Translator` and `ValidatorCacheWarmer` as `final`
+
 Messenger
 ---------
 
  * Make `#[AsMessageHandler]` final
+
+SecurityBundle
+--------------
+
+ * Mark class `ExpressionCacheWarmer` as `final`
+
+Translation
+-----------
+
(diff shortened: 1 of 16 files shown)


src/Symfony/Bundle/FrameworkBundle/CacheWarmer/AbstractPhpFileCacheWarmer.php
```diff
@@ -17,6 +17,9 @@
 use Symfony\Component\Config\Resource\ClassExistenceResource;
 use Symfony\Component\HttpKernel\CacheWarmer\CacheWarmerInterface;
 
+/**
+ * @internal since Symfony 7.1
```
nicolas-grekas: I'm not sure about this one, the class can be useful for third party bundles. We already made a method final/internal so we already closed the class a bit.

xabbuh: Should we mark the `warmUp()` method as final instead?

nicolas-grekas: I'm not sure it would provide much benefits doing so.

nicolas-grekas: up to revert this line @xabbuh?

xabbuh: done

nicolas-grekas:
Thank you @xabbuh.

PR #53154, opened 2023-12-20
[Validator] Add the `Charset` constraint

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

Our use case: we receive some file contents in our DTOs that we only want to process if their encoding matches UTF-8 and reject the whole thing at validation otherwise.

### DIFF
--- a/src/Symfony/Component/Validator/Constraints/Charset.php
+++ b/src/Symfony/Component/Validator/Constraints/Charset.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\Exception\ConstraintDefinitionException;
+
+/**
+ * @author Alexandre Daubois <alex.daubois@gmail.com>
+ */
+#[\Attribute(\Attribute::TARGET_PROPERTY | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
+final class Charset extends Constraint
+{
+    public const BAD_ENCODING_ERROR = '94c5e58b-f892-4e25-8fd6-9d89c80bfe81';
+
+    protected const ERROR_NAMES = [
+        self::BAD_ENCODING_ERROR => 'BAD_ENCODING_ERROR',
+    ];
+
+    public array|string $encodings = [];
+    public string $message = 'The detected encoding "{{ detected }}" does not match one of the accepted encoding: "{{ encodings }}".';
+
+    public function __construct(array|string $encodings = null, string $message = null, array $groups = null, mixed $payload = null, array $options = null)
+    {
+        parent::__construct($options, $groups, $payload);
+
+        $this->message = $message ?? $this->message;
+        $this->encodings = (array) ($encodings ?? $this->encodings);
+
+        if ([] === $this->encodings) {
+            throw new ConstraintDefinitionException(sprintf('The "%s" constraint requires at least one encoding.', static::class));
+        }
+    }
+}
--- a/src/Symfony/Component/Validator/Constraints/CharsetValidator.php
+++ b/src/Symfony/Component/Validator/Constraints/CharsetValidator.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+use Symfony\Component\Validator\Exception\UnexpectedTypeException;
+use Symfony\Component\Validator\Exception\UnexpectedValueException;
+
+/**
+ * @author Alexandre Daubois <alex.daubois@gmail.com>
+ */
+final class CharsetValidator extends ConstraintValidator
+{
+    public function validate(mixed $value, Constraint $constraint): void
+    {
+        if (!$constraint instanceof Charset) {
+            throw new UnexpectedTypeException($constraint, Charset::class);
+        }
+
+        if (null === $value) {
+            return;
+        }
+
+        if (!\is_string($value)) {
+            throw new UnexpectedValueException($value, 'string');
+        }
+
+        if (!\in_array($detected = mb_detect_encoding($value, $constraint->encodings, true), $constraint->encodings, true)) {
+            $this->context->buildViolation($constraint->message)
+                ->setParameter('{{ detected }}', $detected)
+                ->setParameter('{{ encodings }}', implode('", "', $constraint->encodings))
+                ->setCode(Charset::BAD_ENCODING_ERROR)
+                ->addViolation();
+        }
+    }
+}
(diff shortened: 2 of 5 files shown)


94noni (approved):
I second this, exactly the day our production crashed on some file upload ^^
(excel file wrongly saved with bad encoding)

src/Symfony/Component/Validator/Constraints/Encoding.php
```diff
@@ -0,0 +1,43 @@
...
+    public const BAD_ENCODING_ERROR = '94c5e58b-f892-4e25-8fd6-9d89c80bfe81';
+
+    protected const ERROR_NAMES = [
+        self::BAD_ENCODING_ERROR => 'BAD_ENCODING_ERROR',
+    ];
+
+    public array|string $encodings = [];
+    public string $message = 'The detected encoding "{{ detected }}" does not match one of the following encoding: {{ encodings }}.';
+
+    public function __construct(array $options = null, string $message = null, array|string $encodings = null, array $groups = null, mixed $payload = null)
```
xabbuh: Not sure if we discussed that before, but do we still need to have the. `$options` parameter when we no longer support annotations?

alexandre-daubois: I'm not aware of this, but if it is a thing (which I would totally agree with), I'd be glad to remove this artifact from the past :smile:

src/Symfony/Component/Validator/Constraints/Encoding.php
```diff
@@ -0,0 +1,43 @@
...
+namespace Symfony\Component\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\Exception\ConstraintDefinitionException;
+
+/**
+ * @author Alexandre Daubois <alex.daubois@gmail.com>
+ */
+#[\Attribute(\Attribute::TARGET_PROPERTY | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
+class Encoding extends Constraint
```
ro0NL: `CharacterEncoding`?

> mb_detect_encoding — Detect character encoding

src/Symfony/Component/Validator/Constraints/Encoding.php
alexandre-daubois: Alright I had a look at CssColor and it seems you're right. I updated accordingly :slightly_smiling_face:

src/Symfony/Component/Validator/Constraints/Encoding.php
alexandre-daubois: I was hesitating. I think you're right, updated :slightly_smiling_face:

src/Symfony/Component/Validator/Constraints/Encoding.php
xabbuh: maybe we should wait nonetheless (I found #51393 in the meantime)

alexandre-daubois: Even if some constraints are already built without `$options`? (e.g. CssColor)

nicolas-grekas:
Just bikeshadding about the name: what about Charset?

alexandre-daubois:
I like Charset very much! I find it clearer. I updated the PR. I kept the `encodings` constraint option name to match the name of the `mb_detect_encoding()` function argument.

nicolas-grekas:
My thoughts: On all my apps, I'd want to reject any non-UTF-8 content. It might already be the case when the DB refuses to persist invalid UTF-8. So this contraints looks both needed and not needed since I think we might want to add this check earlier when processing the input.
Does that make sense?

@alexandre-daubois can you share your use case more precisely? When can one submit non-UTF-8 in your case?

src/Symfony/Component/Validator/Constraints/Charset.php
```diff
@@ -0,0 +1,43 @@
...
+class Charset extends Constraint
+{
+    public const BAD_ENCODING_ERROR = '94c5e58b-f892-4e25-8fd6-9d89c80bfe81';
+
+    protected const ERROR_NAMES = [
+        self::BAD_ENCODING_ERROR => 'BAD_ENCODING_ERROR',
+    ];
+
+    public array|string $encodings = [];
+    public string $message = 'The detected encoding "{{ detected }}" does not match one of the following encoding: {{ encodings }}.';
```
nicolas-grekas:
```suggestion
    public string $message = 'The detected encoding "{{ detected }}" does not match one of the accepted encoding: "{{ encodings }}".';
```

src/Symfony/Component/Validator/Constraints/CharsetValidator.php
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        if (!\is_string($value)) {
+            throw new UnexpectedValueException($value, 'string');
+        }
+
+        if (!\in_array($detected = mb_detect_encoding($value, $constraint->encodings, true), $constraint->encodings, true)) {
+            $this->context->buildViolation($constraint->message)
+                ->setParameter('{{ detected }}', $detected)
+                ->setParameter('{{ encodings }}', implode(', ', $constraint->encodings))
```
nicolas-grekas:
```suggestion
                ->setParameter('{{ encodings }}', implode('", "', $constraint->encodings))
```

src/Symfony/Component/Validator/Tests/Constraints/CharacterEncodingValidatorTest.php
```diff
@@ -0,0 +1,86 @@
...
+        yield ['ώ', ['UTF-16']];
+    }
+
+    public static function provideInvalidValues(): \Generator
+    {
+        yield ['my non-Äscîi string', ['ASCII']];
+        yield ['😊', ['7bit']];
+    }
+
+    public static function provideInvalidTypes(): \Generator
```
nicolas-grekas: the return type is noise in providers...

nicolas-grekas (approved):
> > we receive some file contents in our DTOs

hum, actually you already answered that sorry :)

here are some minor comments, LGTM

alexandre-daubois:
Thank you for the review, comments are addressed :+1:

dunglas (approved):
Maybe could you make the new classes `final`?

alexandre-daubois:
`final` suits me fine, added :+1:

nicolas-grekas:
Thank you @alexandre-daubois.

smnandre:
Hi @alexandre-daubois 

There is a small problem in the implementation, i poke you there but if you prefer i can create a full issue 

As you use the "valid" encodings in your call to mb_detect_encodings, if the value is _not_ encoded in one of those values, the function will always return false. 

Meaning you'll never be able to display anything in the {{detected}} value of the message.

```
if (!\in_array($detected = mb_detect_encoding($value, $constraint->encodings, true), (array) $constraint->encodings, true)) {
    $this->context->buildViolation($constraint->message)
        ->setParameter('{{ detected }}', $detected)
        ->setParameter('{{ encodings }}', implode(', ', $constraint->encodings))
        ->setCode(Charset::BAD_ENCODING_ERROR)
        ->addViolation();
  }
```        

With  '😊' and ['ASCII'], the generated message will be 

> The detected character encoding is invalid (). Allowed encodings are ASCII.


Oh and just another (minor) thing.. the mb_string extension seems to not be required by Validator, but the mb_string polyfill does only check ASCII, UTF-8 and latin-ish encodings. Maybe something to add in doc, or to check at runtime ?

alexandre-daubois:
Hi! Thank you for letting me know Simon, I'll provide a fix for this :slightly_smiling_face:

zerkms:
Why the constraint is named `Charset` when it validates the encoding? Why it's not `Encoding`?

It even uses the `mb_detect_encoding` internally, which is named `..._encoding` not `..._charset` :-D

OskarStark:
Can you please open a new issue to discuss this before the 7.2 release? Thanks

derrabus:
@OskarStark This feature has been shipped with 7.1 already.

OskarStark:
Indeed, time flies 😄

PR #53157, opened 2023-12-20
[Mailer] Throw `TransportException` when unable to read from socket

xdanik:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? |no
| Issues        | None
| License       | MIT

We are seeing error `fgets(): SSL: Connection reset by peer` multiple times a day from connection to Office 365 SMTP server (smtp.office365.com:587).
It's certainly related to some kind of timeout as we are sending emails from long running queue dispatcher and error shows up only occasionally and never with the first message. We are not seeing this issue with any other SMTP server, but we have not tested much past smtp.mandrillapp.com and local MailHog.

We have tried adjusting the `$pingThreshold` and `$restartThreshold` options, but without much success (well `$restartThreshold = 1` resolves the issue, but it also forces the transport to close connection after each message).

Stack trace:
```
#0 /var/www/vendor/symfony/mailer/Transport/Smtp/Stream/AbstractStream.php(77): fgets(Resource(stream))
#1 /var/www/vendor/symfony/mailer/Transport/Smtp/SmtpTransport.php(315): Symfony\Component\Mailer\Transport\Smtp\Stream\AbstractStream->readLine()
#2 /var/www/vendor/symfony/mailer/Transport/Smtp/SmtpTransport.php(181): Symfony\Component\Mailer\Transport\Smtp\SmtpTransport->getFullResponse()
#3 /var/www/vendor/symfony/mailer/Transport/Smtp/SmtpTransport.php(140): Symfony\Component\Mailer\Transport\Smtp\SmtpTransport->executeCommand("RSET
", Array(1))
#4 /var/www/vendor/symfony/mailer/Mailer.php(45): Symfony\Component\Mailer\Transport\Smtp\SmtpTransport->send(Object(Symfony\Component\Mime\Email), Null)
#5 (our queue dispatcher): Symfony\Component\Mailer\Mailer->send(Object(Symfony\Component\Mime\Email))
```

App is running on PHP 8.0.28 on Debian Linux x64, Mailer v5.4.22.

I would gladly written some tests for this, but I don't know how to simulate calls to low-level stream functions like fgets.

### DIFF
--- a/src/Symfony/Component/Mailer/Transport/Smtp/Stream/AbstractStream.php
+++ b/src/Symfony/Component/Mailer/Transport/Smtp/Stream/AbstractStream.php
@@ -74,7 +74,7 @@ abstract class AbstractStream
             return '';
         }
 
-        $line = fgets($this->out);
+        $line = @fgets($this->out);
         if ('' === $line || false === $line) {
             $metas = stream_get_meta_data($this->out);
             if ($metas['timed_out']) {
@@ -82,6 +82,9 @@ abstract class AbstractStream
             }
             if ($metas['eof']) {
                 throw new TransportException(sprintf('Connection to "%s" has been closed unexpectedly.', $this->getReadConnectionDescription()));
+            }
+            if (false === $line) {
+                throw new TransportException(sprintf('Unable to read from connection to "%s": ', $this->getReadConnectionDescription()).error_get_last()['message']);
             }
         }
 


src/Symfony/Component/Mailer/Transport/Smtp/Stream/AbstractStream.php
```diff
@@ -74,7 +74,7 @@ public function readLine(): string
             return '';
         }
 
-        $line = fgets($this->out);
+        $line = @fgets($this->out);
```
Nyholm, line 77: Do you really need to silence this?

Nyholm (commented):
This is a great PR. Thank you. 

I think we should throw an exception here. 
According to the manual for `fgets`: 

> If there is no more data to read in the file pointer, then false is returned. If an error occurs, false is returned.

We check for `eof()` earlier in this method. So if `fgets` returns false here it must be an error and we should convert it to an exception.

src/Symfony/Component/Mailer/Transport/Smtp/Stream/AbstractStream.php
xdanik: This is the line that throws the `fgets(): SSL: Connection reset by peer` error. It's not an exception - It's error catched by set_error_handler with severity of Error.

I guess that it could be catched by try catch (Error) but that would require rewrite of the current throws for timeout and EOF. I wanted to make as few changes as possible.

Should I rewrite it to use try-catch instead?

xdanik: I believe this is very similar case to silenced `fwrite()` just a few lines above https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Mailer/Transport/Smtp/Stream/AbstractStream.php#L44

xdanik: I was mistaken - The severity is not Error but just a E_WARNING. It cannot be catched by try catch (Error).
Seems like silencing is the only option then.

Nyholm: Hm.. My initial thought was to keep the error/warning. Because if we silence it, the user will never see the text `SSL: Connection reset by peer`. 

> Seems like silencing is the only option then.

Yepp. Im afraid you are correct. Thank you for checking.

xdanik: Agree that the message could be useful for debugging.
We could call `error_clear_last()` before calling `@fgets()` and in case of an error obtain the message from `error_get_last()`.

Not sure what is the performance impact of that though.

fabpot: Performance is not critical here IMHO.

nicolas-grekas:
Thank you @xdanik.

PR #53160, opened 2023-12-20
[PropertyInfo] Deprecate PropertyInfo Type

mtarld:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Issues        |
| License       | MIT

This PR is a follow-up of https://github.com/symfony/symfony/pull/52510.

As the TypeInfo's `Type` aims to represent types in the Symfony ecosystem, the PropertyInfo's `Type` needs to be deprecated in favor of the first one.

### DIFF
--- a/UPGRADE-7.1.md
+++ b/UPGRADE-7.1.md
@@ -11,6 +11,11 @@ Cache
 
  * Deprecate `CouchbaseBucketAdapter`, use `CouchbaseCollectionAdapter` instead
 
+DoctrineBridge
+--------------
+
+ * The `DoctrineExtractor::getTypes()` method is deprecated, use `DoctrineExtractor::getType()` instead
+
 ExpressionLanguage
 ------------------
 
@@ -21,6 +26,11 @@ FrameworkBundle
 ---------------
 
  * Mark classes `ConfigBuilderCacheWarmer`, `Router`, `SerializerCacheWarmer`, `TranslationsCacheWarmer`, `Translator` and `ValidatorCacheWarmer` as `final`
+
+PropertyInfo
+------------
+
+ * The `PropertyTypeExtractorInterface::getTypes()` method is deprecated, use `PropertyTypeExtractorInterface::getType()` instead
 
 SecurityBundle
 --------------
--- a/src/Symfony/Bridge/Doctrine/CHANGELOG.md
+++ b/src/Symfony/Bridge/Doctrine/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Deprecate the `DoctrineExtractor::getTypes()` method, use `DoctrineExtractor::getType()` instead
+
 7.0
 ---
 
@@ -17,7 +22,7 @@ CHANGELOG
 6.4
 ---
 
- * [BC BREAK] Add argument `$buildDir` to `ProxyCacheWarmer::warmUp()` 
+ * [BC BREAK] Add argument `$buildDir` to `ProxyCacheWarmer::warmUp()`
  * [BC BREAK] Add return type-hints to `EntityFactory`
  * Deprecate `DbalLogger`, use a middleware instead
  * Deprecate not constructing `DoctrineDataCollector` with an instance of `DebugDataHolder`
--- a/src/Symfony/Bridge/Doctrine/PropertyInfo/DoctrineExtractor.php
+++ b/src/Symfony/Bridge/Doctrine/PropertyInfo/DoctrineExtractor.php
@@ -24,7 +24,9 @@ use Doctrine\Persistence\Mapping\MappingException;
 use Symfony\Component\PropertyInfo\PropertyAccessExtractorInterface;
 use Symfony\Component\PropertyInfo\PropertyListExtractorInterface;
 use Symfony\Component\PropertyInfo\PropertyTypeExtractorInterface;
-use Symfony\Component\PropertyInfo\Type;
+use Symfony\Component\PropertyInfo\Type as LegacyType;
+use Symfony\Component\TypeInfo\Type;
+use Symfony\Component\TypeInfo\TypeIdentifier;
 
 /**
  * Extracts data using Doctrine ORM and ODM metadata.
@@ -55,7 +57,7 @@ class DoctrineExtractor implements PropertyListExtractorInterface, PropertyTypeE
         return $properties;
     }
 
-    public function getTypes(string $class, string $property, array $context = []): ?array
+    public function getType(string $class, string $property, array $context = []): ?Type
     {
         if (null === $metadata = $this->getMetadata($class)) {
             return null;
@@ -67,16 +69,15 @@ class DoctrineExtractor implements PropertyListExtractorInterface, PropertyTypeE
             if ($metadata->isSingleValuedAssociation($property)) {
                 if ($metadata instanceof ClassMetadata) {
                     $associationMapping = $metadata->getAssociationMapping($property);
-
                     $nullable = $this->isAssociationNullable($associationMapping);
                 } else {
                     $nullable = false;
                 }
 
-                return [new Type(Type::BUILTIN_TYPE_OBJECT, $nullable, $class)];
+                return $nullable ? Type::nullable(Type::object($class)) : Type::object($class);
             }
 
-            $collectionKeyType = Type::BUILTIN_TYPE_INT;
+            $collectionKeyType = TypeIdentifier::INT;
 
             if ($metadata instanceof ClassMetadata) {
                 $associationMapping = $metadata->getAssociationMapping($property);
@@ -104,61 +105,164 @@ class DoctrineExtractor implements PropertyListExtractorInterface, PropertyTypeE
                         }
                     }
 
-                    if (!$collectionKeyType = $this->getPhpType($typeOfField)) {
+                    if (!$collectionKeyType = $this->getTypeIdentifier($typeOfField)) {
                         return null;
                     }
                 }
             }
 
-            return [new Type(
-                Type::BUILTIN_TYPE_OBJECT,
+            return Type::collection(Type::object(Collection::class), Type::object($class), Type::builtin($collectionKeyType));
+        }
+
+        if ($metadata instanceof ClassMetadata && isset($metadata->embeddedClasses[$property])) {
+            return Type::object(self::getMappingValue($metadata->embeddedClasses[$property], 'class'));
+        }
+
+        if (!$metadata->hasField($property)) {
+            return null;
+        }
+
+        $typeOfField = $metadata->getTypeOfField($property);
+
+        if (!$typeIdentifier = $this->getTypeIdentifier($typeOfField)) {
+            return null;
+        }
+
+        $nullable = $metadata instanceof ClassMetadata && $metadata->isNullable($property);
+        $enumType = null;
+
+        if (null !== $enumClass = self::getMappingValue($metadata->getFieldMapping($property), 'enumType') ?? null) {
+            $enumType = $nullable ? Type::nullable(Type::enum($enumClass)) : Type::enum($enumClass);
+        }
+
+        $builtinType = $nullable ? Type::nullable(Type::builtin($typeIdentifier)) : Type::builtin($typeIdentifier);
+
+        return match ($typeIdentifier) {
+            TypeIdentifier::OBJECT => match ($typeOfField) {
+                Types::DATE_MUTABLE, Types::DATETIME_MUTABLE, Types::DATETIMETZ_MUTABLE, 'vardatetime', Types::TIME_MUTABLE => $nullable ? Type::nullable(Type::object(\DateTime::class)) : Type::object(\DateTime::class),
+                Types::DATE_IMMUTABLE, Types::DATETIME_IMMUTABLE, Types::DATETIMETZ_IMMUTABLE, Types::TIME_IMMUTABLE => $nullable ? Type::nullable(Type::object(\DateTimeImmutable::class)) : Type::object(\DateTimeImmutable::class),
+                Types::DATEINTERVAL => $nullable ? Type::nullable(Type::object(\DateInterval::class)) : Type::object(\DateInterval::class),
+                default => $builtinType,
+            },
+            TypeIdentifier::ARRAY => match ($typeOfField) {
+                'array', 'json_array' => $enumType ? null : ($nullable ? Type::nullable(Type::array()) : Type::array()),
+                Types::SIMPLE_ARRAY => $nullable ? Type::nullable(Type::list($enumType ?? Type::string())) : Type::list($enumType ?? Type::string()),
+                default => $builtinType,
+            },
+            TypeIdentifier::INT, TypeIdentifier::STRING => $enumType ? $enumType : $builtinType,
+            default => $builtinType,
+        };
+    }
+
+    /**
+     * @deprecated since Symfony 7.1, use "getType" instead
+     */
+    public function getTypes(string $class, string $property, array $context = []): ?array
+    {
+        trigger_deprecation('symfony/property-info', '7.1', 'The "%s()" method is deprecated, use "%s::getType()" instead.', __METHOD__, self::class);
+
+        if (null === $metadata = $this->getMetadata($class)) {
+            return null;
+        }
+
+        if ($metadata->hasAssociation($property)) {
+            $class = $metadata->getAssociationTargetClass($property);
+
+            if ($metadata->isSingleValuedAssociation($property)) {
+                if ($metadata instanceof ClassMetadata) {
+                    $associationMapping = $metadata->getAssociationMapping($property);
+
+                    $nullable = $this->isAssociationNullable($associationMapping);
+                } else {
+                    $nullable = false;
+                }
+
+                return [new LegacyType(LegacyType::BUILTIN_TYPE_OBJECT, $nullable, $class)];
+            }
+
+            $collectionKeyType = LegacyType::BUILTIN_TYPE_INT;
+
+            if ($metadata instanceof ClassMetadata) {
+                $associationMapping = $metadata->getAssociationMapping($property);
+
+                if (self::getMappingValue($associationMapping, 'indexBy')) {
+                    $subMetadata = $this->entityManager->getClassMetadata(self::getMappingValue($associationMapping, 'targetEntity'));
+
+                    // Check if indexBy value is a property
+                    $fieldName = self::getMappingValue($associationMapping, 'indexBy');
+                    if (null === ($typeOfField = $subMetadata->getTypeOfField($fieldName))) {
+                        $fieldName = $subMetadata->getFieldForColumn(self::getMappingValue($associationMapping, 'indexBy'));
+                        // Not a property, maybe a column name?
+                        if (null === ($typeOfField = $subMetadata->getTypeOfField($fieldName))) {
+                            // Maybe the column name is the association join column?
+                            $associationMapping = $subMetadata->getAssociationMapping($fieldName);
+
+                            $indexProperty = $subMetadata->getSingleAssociationReferencedJoinColumnName($fieldName);
+                            $subMetadata = $this->entityManager->getClassMetadata(self::getMappingValue($associationMapping, 'targetEntity'));
+
+                            // Not a property, maybe a column name?
+                            if (null === ($typeOfField = $subMetadata->getTypeOfField($indexProperty))) {
+                                $fieldName = $subMetadata->getFieldForColumn($indexProperty);
+                                $typeOfField = $subMetadata->getTypeOfField($fieldName);
+                            }
+                        }
+                    }
+
+                    if (!$collectionKeyType = $this->getTypeIdentifierLegacy($typeOfField)) {
+                        return null;
+                    }
+                }
+            }
+
+            return [new LegacyType(
+                LegacyType::BUILTIN_TYPE_OBJECT,
                 false,
                 Collection::class,
                 true,
-                new Type($collectionKeyType),
-                new Type(Type::BUILTIN_TYPE_OBJECT, false, $class)
+                new LegacyType($collectionKeyType),
+                new LegacyType(LegacyType::BUILTIN_TYPE_OBJECT, false, $class)
             )];
         }
 
         if ($metadata instanceof ClassMetadata && isset($metadata->embeddedClasses[$property])) {
-            return [new Type(Type::BUILTIN_TYPE_OBJECT, false, self::getMappingValue($metadata->embeddedClasses[$property], 'class'))];
+            return [new LegacyType(LegacyType::BUILTIN_TYPE_OBJECT, false, self::getMappingValue($metadata->embeddedClasses[$property], 'class'))];
         }
 
         if ($metadata->hasField($property)) {
             $typeOfField = $metadata->getTypeOfField($property);
 
-            if (!$builtinType = $this->getPhpType($typeOfField)) {
+            if (!$builtinType = $this->getTypeIdentifierLegacy($typeOfField)) {
                 return null;
             }
 
             $nullable = $metadata instanceof ClassMetadata && $metadata->isNullable($property);
             $enumType = null;
             if (null !== $enumClass = self::getMappingValue($metadata->getFieldMapping($property), 'enumType') ?? null) {
-                $enumType = new Type(Type::BUILTIN_TYPE_OBJECT, $nullable, $enumClass);
+                $enumType = new LegacyType(LegacyType::BUILTIN_TYPE_OBJECT, $nullable, $enumClass);
             }
 
             switch ($builtinType) {
-                case Type::BUILTIN_TYPE_OBJECT:
+                case LegacyType::BUILTIN_TYPE_OBJECT:
                     switch ($typeOfField) {
                         case Types::DATE_MUTABLE:
                         case Types::DATETIME_MUTABLE:
                         case Types::DATETIMETZ_MUTABLE:
                         case 'vardatetime':
                         case Types::TIME_MUTABLE:
-                            return [new Type(Type::BUILTIN_TYPE_OBJECT, $nullable, 'DateTime')];
+                            return [new LegacyType(LegacyType::BUILTIN_TYPE_OBJECT, $nullable, 'DateTime')];
 
                         case Types::DATE_IMMUTABLE:
                         case Types::DATETIME_IMMUTABLE:
                         case Types::DATETIMETZ_IMMUTABLE:
                         case Types::TIME_IMMUTABLE:
-                            return [new Type(Type::BUILTIN_TYPE_OBJECT, $nullable, 'DateTimeImmutable')];
+                            return [new LegacyType(LegacyType::BUILTIN_TYPE_OBJECT, $nullable, 'DateTimeImmutable')];
 
                         case Types::DATEINTERVAL:
-                            return [new Type(Type::BUILTIN_TYPE_OBJECT, $nullable, 'DateInterval')];
+                            return [new LegacyType(LegacyType::BUILTIN_TYPE_OBJECT, $nullable, 'DateInterval')];
                     }
 
                     break;
-                case Type::BUILTIN_TYPE_ARRAY:
+                case LegacyType::BUILTIN_TYPE_ARRAY:
                     switch ($typeOfField) {
                         case 'array':      // DBAL < 4
                         case 'json_array': // DBAL < 3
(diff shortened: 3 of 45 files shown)


src/Symfony/Bridge/Doctrine/PropertyInfo/DoctrineExtractor.php
```diff
@@ -164,26 +168,39 @@ public function getTypes(string $class, string $property, array $context = []):
...
+
+            return $nullable ? Type::nullable($t) : $t;
         }
 
         return null;
     }
 
+    public function getTypes(string $class, string $property, array $context = []): ?array
+    {
+        trigger_deprecation('symfony/doctrine-bridge', '7.1', 'The "%s()" method is deprecated. Use "%s::getType()" instead.', __METHOD__, self::class);
```
chalasr:
```suggestion
        trigger_deprecation('symfony/doctrine-bridge', '7.1', 'The "%s()" method is deprecated, use "%s::getType()" instead.', __METHOD__, self::class);
```

src/Symfony/Component/PropertyInfo/Extractor/ConstructorExtractor.php
```diff
@@ -28,15 +30,22 @@ public function __construct(
...
                 return $value;
             }
         }
 
         return null;
     }
+
+    public function getTypes(string $class, string $property, array $context = []): ?array
+    {
+        trigger_deprecation('symfony/property-info', '7.1', 'The "%s()" method is deprecated. Use "%s::getType()" instead.', __METHOD__, self::class);
```
chalasr:
```suggestion
        trigger_deprecation('symfony/property-info', '7.1', 'The "%s()" method is deprecated, use "%s::getType()" instead.', __METHOD__, self::class);
```
Same for all notices

src/Symfony/Component/PropertyInfo/PropertyTypeExtractorInterface.php
```diff
@@ -21,7 +24,17 @@ interface PropertyTypeExtractorInterface
...
      */
     public function getTypes(string $class, string $property, array $context = []): ?array;
+
+    /**
+     * Gets type of a property.
+     *
+     * @param class-string         $class
+     * @param array<string, mixed> $context
+     */
+    public function getType(string $class, string $property, array $context = []): ?Type;
```
chalasr: This is a BC break as hinted by the failing CI build, the method needs to be virtual for now
```php
* @method ?Type getType(string $class, string $property, array $context = []) 
```
(I intentionally removed the description as it's stating the obvious)
We'll really add it when removing `getTypes()` in 8.0.

chalasr (changes requested):

chalasr (approved):
Looking good 👍

src/Symfony/Bundle/FrameworkBundle/composer.json
```diff
@@ -64,10 +64,11 @@
         "symfony/string": "^6.4|^7.0",
         "symfony/translation": "^6.4|^7.0",
         "symfony/twig-bundle": "^6.4|^7.0",
+        "symfony/type-info": "^7.1",
         "symfony/validator": "^6.4|^7.0",
         "symfony/workflow": "^6.4|^7.0",
         "symfony/yaml": "^6.4|^7.0",
-        "symfony/property-info": "^6.4|^7.0",
+        "symfony/property-info": "^7.1",
```
chalasr: Can we manage to keep supporting 6.4 somehow? Here and in doctrine-bridge

fabpot:
@mtarld Can you rebase this one now that the dependent one has been merged? Thank you.

chalasr:
Rebase needed again

src/Symfony/Bridge/Doctrine/PropertyInfo/DoctrineExtractor.php
```diff
@@ -110,55 +113,195 @@ public function getTypes(string $class, string $property, array $context = []):
...
+
+            $t = Type::builtin($builtinType);
+
+            return $nullable ? Type::nullable($t) : $t;
+        }
+
+        return null;
+    }
+
+    public function getTypes(string $class, string $property, array $context = []): ?array
```
chalasr: missing `@deprecated` tag

src/Symfony/Component/PropertyInfo/Extractor/ConstructorExtractor.php
```diff
@@ -28,15 +30,22 @@ public function __construct(
...
+            $value = $extractor->getTypeFromConstructor($class, $property);
             if (null !== $value) {
                 return $value;
             }
         }
 
         return null;
     }
+
+    public function getTypes(string $class, string $property, array $context = []): ?array
```
chalasr, line 46: `@deprecated`

src/Symfony/Component/PropertyInfo/Extractor/PhpDocExtractor.php
```diff
@@ -126,46 +133,59 @@ public function getTypes(string $class, string $property, array $context = []):
...
-            return $types;
+            return $type;
         }
 
-        return [new Type(Type::BUILTIN_TYPE_ARRAY, false, null, true, new Type(Type::BUILTIN_TYPE_INT), $types[0])];
+        return Type::list($type);
     }
 
-    public function getTypesFromConstructor(string $class, string $property): ?array
+    public function getTypes(string $class, string $property, array $context = []): ?array
```
chalasr: `@deprecated` - every `getTypes()` should have it

src/Symfony/Component/PropertyInfo/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Introduce `PropertyDocBlockExtractorInterface` to extract a property's doc block
  * Restrict access to `PhpStanExtractor` based on visibility
+ * Deprecate the `Type` class, use `Symfony\Component\TypeInfo\Type` class of `symfony/type-info` component instead
```
chalasr, line 9: we need a note about the deprecated `PropertyTypeExtractorInterface::getTypes`

chalasr (changes requested):

src/Symfony/Component/Validator/composer.json
```diff
@@ -49,6 +50,7 @@
         "symfony/intl": "<6.4",
         "symfony/property-info": "<6.4",
         "symfony/translation": "<6.4.3|>=7.0,<7.0.3",
+        "symfony/type-info": "<7.1",
```
nicolas-grekas: not needed

src/Symfony/Component/TypeInfo/TypeFactoryTrait.php
```diff
@@ -160,31 +160,31 @@ public static function collection(BuiltinType|ObjectType|GenericType $type, Type
...
+    public static function iterable(?Type $value = null, ?Type $key = null, bool $asList = false): CollectionType
     {
         return self::collection(self::builtin(TypeIdentifier::ITERABLE), $value, $key, $asList);
     }
 
     /**
      * @return CollectionType<BuiltinType<TypeIdentifier::ARRAY>>
      */
-    public static function list(Type $value = null): CollectionType
+    public static function list(?Type $value = null): CollectionType
```
nicolas-grekas: can you please submit all CS changes separately to keep the diff clean on this one?

src/Symfony/Component/TypeInfo/Type/UnionType.php
```diff
@@ -57,4 +57,32 @@ public function __toString(): string
 
         return $string;
     }
+
+    /**
+     * BC Layer for Symfony\Component\PropertyInfo\Type.
+     *
+     * @internal
```
nicolas-grekas: can the comment hint future maintainers about when this can be removed? (same in all similar situations)

src/Symfony/Component/TypeInfo/TypeResolver/StringTypeResolver.php
```diff
@@ -172,20 +174,26 @@ private function getTypeFromNode(TypeNode $node, ?TypeContext $typeContext): Typ
...
-            $type = $this->getTypeFromNode($node->type, $typeContext);
+            $type = $this->getTypeFromNode($node->type, $typeContext, $backwardCompatible);
 
             // handle integer ranges as simple integers
             if ($type->isA(TypeIdentifier::INT)) {
                 return $type;
             }
 
-            $variableTypes = array_map(fn (TypeNode $t): Type => $this->getTypeFromNode($t, $typeContext), $node->genericTypes);
+            // BC layer to convert class-string generics as simple strings
```
nicolas-grekas: when can this be removed also?

src/Symfony/Component/TypeInfo/Type/CollectionType.php
```diff
@@ -34,6 +34,11 @@ public function __construct(
         private readonly BuiltinType|ObjectType|GenericType $type,
         private readonly bool $isList = false,
     ) {
+        // BC layer to allow invalid list keys
+        if (\func_get_args()[2] ?? false) {
```
nicolas-grekas: since it's a constructor, we can use a real argument here, it'd be more explicit

src/Symfony/Component/TypeInfo/BackwardCompatibilityHelper.php
```diff
@@ -0,0 +1,294 @@
...
+            if ($type instanceof UnionType) {
+                $types = [$types, ...$type->getTypes()];
+
+                continue;
+            }
+
+            $types[] = $type;
+        }
+
+        if ($nullable && [] === $types) {
```
nicolas-grekas:
```suggestion
        if ($nullable && !$types) {
```

src/Symfony/Component/Serializer/Exception/NotNormalizableValueException.php
```diff
@@ -22,17 +24,17 @@ class NotNormalizableValueException extends UnexpectedValueException
...
+     *                                             safely to your user. In other words, avoid catching other exceptions and
+     *                                             passing their message directly to this class.
      */
     public static function createForUnexpectedDataType(string $message, mixed $data, array $expectedTypes, ?string $path = null, bool $useMessageForUser = false, int $code = 0, ?\Throwable $previous = null): self
     {
         $self = new self($message, $code, $previous);
 
         $self->currentType = get_debug_type($data);
-        $self->expectedTypes = $expectedTypes;
+        $self->expectedTypes = array_map(fn (string|Type $t): string => $t instanceof Type ? (string) $t : $t, $expectedTypes);
```
nicolas-grekas:
```suggestion
        $self->expectedTypes = array_map('strval', $expectedTypes);
```

src/Symfony/Component/PropertyInfo/Type.php
```diff
@@ -160,6 +167,35 @@ public function getCollectionKeyTypes(): array
...
+        return [];
+    }
+
+    private function validateCollectionArgument(array|self|null $collectionArgument, int $argumentIndex, string $argumentName): ?array
+    {
+        if (null === $collectionArgument) {
+            return null;
+        }
+
+        if (\is_array($collectionArgument)) {
```
nicolas-grekas:
```suggestion
        if (!\is_array($collectionArgument)) { // return early
```

src/Symfony/Component/PropertyInfo/PropertyInfoExtractor.php
```diff
@@ -51,6 +53,14 @@ public function getLongDescription(string $class, string $property, array $conte
         return $this->extract($this->descriptionExtractors, 'getLongDescription', [$class, $property, $context]);
     }
 
+    public function getType(string $class, string $property, array $context = []): ?Type
+    {
+        return $this->extract($this->typeExtractors, 'getType', [$class, $property, $context]);
+    }
+
+    /**
+     * @deprecated since Symfony 7.1, use "getType" instead.
```
nicolas-grekas: shouldn't the method trigger a runtime deprecation also? (same in all deprecated methods)

src/Symfony/Component/PropertyInfo/Extractor/PhpDocExtractor.php
```diff
@@ -126,46 +133,62 @@ public function getTypes(string $class, string $property, array $context = []):
...
+                } catch (LogicException) {
+                    // if there is no parent for the current class, we keep the "parent" raw string
                 }
             }
+
+            $types[] = $type->isNullable() ? Type::nullable(Type::object($normalizedClassName)) : Type::object($normalizedClassName);
         }
 
-        if (!isset($types[0])) {
+        if (null === ($type = $types[0] ?? null)) {
```
nicolas-grekas:
```suggestion
        if (null === $type = $types[0] ?? null) {
```

src/Symfony/Component/PropertyInfo/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ CHANGELOG
...
+ * Deprecate the `PropertyTypeExtractorInterface::getTypes` method, use `PropertyTypeExtractorInterface::getType` instead
```
nicolas-grekas:
```suggestion
 * Deprecate the `PropertyTypeExtractorInterface::getTypes()` method, use `PropertyTypeExtractorInterface::getType`() instead
```

src/Symfony/Bridge/Doctrine/Validator/DoctrineLoader.php
```diff
@@ -66,7 +65,7 @@ public function loadClassMetadata(ClassMetadata $metadata): bool
          */
         $existingUniqueFields = $this->getExistingUniqueFields($metadata);
 
-        // Type and nullable aren't handled here, use the PropertyInfo Loader instead.
+        // type and nullable aren't handled here, use the PropertyInfo Loader instead.
```
nicolas-grekas: not needed

src/Symfony/Bridge/Doctrine/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Deprecate the `DoctrineExtractor::getTypes` method, use `DoctrineExtractor::getType` instead
```
nicolas-grekas:
```suggestion
 * Deprecate the `DoctrineExtractor::getTypes()` method, use `DoctrineExtractor::getType()` instead
```

UPGRADE-7.1.md
```diff
@@ -16,6 +21,11 @@ Messenger
 
  * Make `#[AsMessageHandler]` final
 
+PropertyInfo
+------------
+
+ * The `PropertyTypeExtractorInterface::getTypes` method is deprecated, use `PropertyTypeExtractorInterface::getType` instead
```
nicolas-grekas:
```suggestion
 * The `PropertyTypeExtractorInterface::getTypes()` method is deprecated, use `PropertyTypeExtractorInterface::getType()` instead
```

UPGRADE-7.1.md
```diff
@@ -6,6 +6,11 @@ Cache
 
  * Deprecate `CouchbaseBucketAdapter`, use `CouchbaseCollectionAdapter` instead
 
+DoctrineBridge
+--------------
+
+ * The `DoctrineExtractor::getTypes` method is deprecated, use `DoctrineExtractor::getType` instead
```
nicolas-grekas:
```suggestion
 * The `DoctrineExtractor::getTypes()` method is deprecated, use `DoctrineExtractor::getType()` instead
```

src/Symfony/Component/PropertyInfo/PropertyTypeExtractorInterface.php
```diff
@@ -11,17 +11,24 @@
...
+ *
+ * @method Type|null getType(string $class, string $property, array $context = [])
  */
 interface PropertyTypeExtractorInterface
 {
     /**
      * Gets types of a property.
      *
-     * @return Type[]|null
+     * @deprecated since Symfony 7.1, use "getType" instead.
```
nicolas-grekas: what about deprecating the interface instead and providing a new one?
I'm wondering if this wouldn't help with the BC layer

nicolas-grekas (commented):
Thanks for the PR, here are some random comments.
As a general note, please provide insight in the comments about the moment when the BC-related code could be removed.

src/Symfony/Component/TypeInfo/Type/CollectionType.php
mtarld: By doing that, the following, the list keys validation will be explicit like so:
```php
public function __construct(
    private readonly BuiltinType|ObjectType|GenericType $type,
    private readonly bool $isList = false,

    // BC layer to allow invalid list keys
    // Can be removed as soon as PropertyTypeExtractorInterface::getTypes() is removed (8.0).
    private readonly bool $validateListKeys = true,
) {
}
```
Then, how to remove it in 8.0 without breaking backward compatibility?

src/Symfony/Component/PropertyInfo/PropertyTypeExtractorInterface.php
mtarld: I'm not sure that it'll help here.
Instead of doing `method_exists` stuff, we'll do `instanceof` stuff, but the rest will remain the same.
Plus, creating a new interface will require finding a new name (and the current one fits perfectly).

But maybe I'm missing something?

src/Symfony/Component/PropertyInfo/Extractor/PhpStanExtractor.php
```diff
@@ -63,103 +68,111 @@ public function __construct(?array $mutatorPrefixes = null, ?array $accessorPref
...
+        $typeContext = $this->typeContextFactory->createFromClassName($class);
 
-        if (!isset($types[0])) {
-            return null;
-        }
+        return $this->stringTypeResolver->resolve((string) $tagDocNode->type, $typeContext, $backwardCompatible);
+    }
+
+    /**
+     * @deprecated since Symfony 7.1, use "getTypeFromConstructor" instead.
```
chalasr:
```suggestion
     * @deprecated since Symfony 7.1, use "getTypeFromConstructor" instead
```

(nitpick) we don't treat phpdoc annotations' descriptions as sentences as per https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/c1a9344ce4aea47b7926f1128027ffda9de029e8/doc/rules/phpdoc/phpdoc_annotation_without_dot.rst#L2

src/Symfony/Component/PropertyInfo/PropertyTypeExtractorInterface.php
chalasr: looking at it, I don't think an interface would make it simpler either

src/Symfony/Component/TypeInfo/Type/CollectionType.php
chalasr: If allowing invalid keys is an implementation detail of the deprecated PropertyInfo's `Type` that we would like to ultimately get rid of in TypeInfo, we probably want to explicitly deprecate this behavior on its own **after** that the PropertyInfo API is removed anyway. This would allow for smooth upgrades as opposed to silently removing that behavior along with the deprecated PropertyInfo's `Type`.
i.e. make it a real parameter in TypeInfo now as suggested, then deprecate it in 8.1, and finally remove it in 9.0.

chalasr: Or maybe we could introduce the parameter and deprecate it right now with an explicit notice?

mtarld: I'd rather go with the latter one indeed, as `CollectionType` has been released yet.

src/Symfony/Component/PropertyInfo/Type.php
```diff
@@ -11,27 +11,38 @@
...
 namespace Symfony\Component\PropertyInfo;
 
+use Symfony\Component\TypeInfo\BackwardCompatibilityHelper;
+use Symfony\Component\TypeInfo\Type as TypeInfoType;
+use Symfony\Component\TypeInfo\Type\CollectionType;
+use Symfony\Component\TypeInfo\Type\GenericType;
+use Symfony\Component\TypeInfo\Type\ObjectType;
+use Symfony\Component\TypeInfo\TypeIdentifier;
+
+trigger_deprecation('symfony/property-info', '7.1', 'The "%s" class is deprecated. Use "%s" of "symfony/type-info" component instead.', Type::class, TypeInfoType::class);
```
nicolas-grekas:
```suggestion
trigger_deprecation('symfony/property-info', '7.1', 'The "%s" class is deprecated. Use "%s" from the "symfony/type-info" component instead.', Type::class, TypeInfoType::class);
```

src/Symfony/Component/PropertyInfo/Type.php
```diff
@@ -11,27 +11,38 @@
...
+
 /**
  * Type value object (immutable).
  *
  * @author Kévin Dunglas <dunglas@gmail.com>
  *
+ * @deprecated since Symfony 7.1, use "Symfony\Component\TypeInfo\Type" of "symfony/type-info" component instead
```
nicolas-grekas:
```suggestion
 * @deprecated since Symfony 7.1, use "Symfony\Component\TypeInfo\Type" from the "symfony/type-info" component instead
```

src/Symfony/Component/PropertyInfo/Type.php
```diff
@@ -11,27 +11,38 @@
...
+    public const BUILTIN_TYPE_STRING = TypeIdentifier::STRING->value;
+    public const BUILTIN_TYPE_BOOL = TypeIdentifier::BOOL->value;
+    public const BUILTIN_TYPE_RESOURCE = TypeIdentifier::RESOURCE->value;
+    public const BUILTIN_TYPE_OBJECT = TypeIdentifier::OBJECT->value;
+    public const BUILTIN_TYPE_ARRAY = TypeIdentifier::ARRAY->value;
+    public const BUILTIN_TYPE_NULL = TypeIdentifier::NULL->value;
+    public const BUILTIN_TYPE_FALSE = TypeIdentifier::FALSE->value;
+    public const BUILTIN_TYPE_TRUE = TypeIdentifier::TRUE->value;
+    public const BUILTIN_TYPE_CALLABLE = TypeIdentifier::CALLABLE->value;
+    public const BUILTIN_TYPE_ITERABLE = TypeIdentifier::ITERABLE->value;
```
nicolas-grekas: not sure this change is needed, I'd prefer reverting it so make it clear that we're keeping the same values (do we?)

src/Symfony/Component/Validator/Mapping/Loader/PropertyInfoLoader.php
```diff
@@ -159,12 +180,16 @@ private function handleAllConstraint(string $property, ?All $allConstraint, Prop
...
             $constraints[] = new NotNull();
         }
 
-        if (!$containsTypeConstraint) {
-            $constraints[] = $this->getTypeConstraint($propertyInfoType->getBuiltinType(), $propertyInfoType);
+        if (!$containsTypeConstraint && null !== $typeConstraint = $this->getTypeConstraint($typeInfoType)) {
+            $constraints[] = $typeConstraint;
+        }
+
+        if ([] === $constraints) {
```
nicolas-grekas:
```suggestion
        if (!$constraints) {
```

src/Symfony/Component/Serializer/Normalizer/DateTimeNormalizer.php
```diff
@@ -11,9 +11,9 @@
 
 namespace Symfony\Component\Serializer\Normalizer;
 
-use Symfony\Component\PropertyInfo\Type;
 use Symfony\Component\Serializer\Exception\InvalidArgumentException;
 use Symfony\Component\Serializer\Exception\NotNormalizableValueException;
+use Symfony\Component\TypeInfo\Type;
```
nicolas-grekas: This change means that if one loads serializer v7.1 with property-info v6.4, things will explode.
I'd rather make property-info required for serializer (same for validator)
Then we could move BackwardCompatibilityHelper to property-info, where it belongs IMHO (and not `@internal` - internal means "don't use cross-packages, and that applies cross-Symfony packages")

nicolas-grekas (commented):
BC layers are hard. I'm not 100% confident we're there yet.
For PropertyTypeExtractorInterface, I don't see the BC layer: the current plan is to force implementing a deprecated method.
If naming is the issue, what about TypedPropertyExtractorInterface?

src/Symfony/Component/PropertyInfo/Extractor/PhpStanExtractor.php
```diff
@@ -63,103 +68,111 @@ public function __construct(?array $mutatorPrefixes = null, ?array $accessorPref
...
+        trigger_deprecation('symfony/property-info', '7.1', 'The "%s()" method is deprecated, use "%s::getType()" instead.', __METHOD__, self::class);
+
+        return BackwardCompatibilityHelper::convertTypeToLegacyTypes($this->getType($class, $property, $context, true));
+    }
+
+    public function getTypeFromConstructor(string $class, string $property): ?Type
     {
+        // BC layer to allow unconsistent legacy type extraction
+        // Can be removed as soon as PhpStanExtractor::getTypesFromConstructor() is removed (8.0).
+        $backwardCompatible = \func_get_args()[2] ?? false;
```
nicolas-grekas: I feel like all these $backwardCompatible variables might make cross-majors compatibility (7.4x8.0) difficult. Can't we keep and run the legacy code in getTypesFromConstructor instead, and use the same approach everywhere?

src/Symfony/Component/TypeInfo/Type.php
```diff
@@ -86,4 +110,28 @@ private function testIdentifier(callable $test): bool
...
     }
+
+    /**
+     * BC layer for Symfony\Component\PropertyInfo\Type.
+     *
+     * Can be removed as soon as Symfony\Component\PropertyInfo\Type is removed (8.0).
+     *
+     * @internal
+     */
+    public function setCollection(bool $collection): void
```
chalasr: `@internal` means code can only be used at *this* component level, so this method must not be called from PropertyInfo. 
Ideally TypeInfo should not be aware of PropertyInfo at all, not even temporarily. Can we somehow manage to keep PropertyInfo's specific code inside PropertyInfo?

src/Symfony/Component/Serializer/Normalizer/UidNormalizer.php
```diff
@@ -71,7 +71,7 @@ public function denormalize(mixed $data, string $type, ?string $format = null, a
         try {
             return $type::fromString($data);
         } catch (\InvalidArgumentException|\TypeError) {
-            throw NotNormalizableValueException::createForUnexpectedDataType(sprintf('The data is not a valid "%s" string representation.', $type), $data, [Type::BUILTIN_TYPE_STRING], $context['deserialization_path'] ?? null, true);
+            throw NotNormalizableValueException::createForUnexpectedDataType(sprintf('The data is not a valid "%s" string representation.', $type), $data, [Type::string()], $context['deserialization_path'] ?? null, true);
```
stof: This should not use `Type::string()` here. `NotNormalizableValueException::createForUnexpectedDataType` expects a PHP type name and we should use the string `'string'` directly instead of adding a mandatory dependency for the error handling path (btw, we have a similar issue about using `Symfony\Component\PropertyInfo\Type` right now, which is not marked as a mandatory dependency either).

src/Symfony/Component/PropertyInfo/Tests/TypeTest.php
```diff
@@ -85,4 +89,34 @@ public function testInvalidCollectionValueArgument()
...
+
+    /**
+     * @dataProvider createProperInternalTypeDataProvider
+     *
+     * @param list<TypeInfoType> $types
+     * @param list<Type>         $legacyTypes
+     */
+    public function testCreateProperInternalTypes(array $types, array $legacyTypes)
+    {
+        $this->assertEquals($types, array_map(fn (Type $t): TypeInfoType => $t->internalType, $legacyTypes));
```
nicolas-grekas:
```suggestion
        $this->assertEquals($types, array_column($legacyTypes, 'internalType'));
```

dunglas:
@mtarld as discussed at SymfonyLive, could you also try to update API Platform to use this patch? As we do extensive use of PropertyInfo (the component has initially be created as part of API Platform), this will validate that all use cases are covered and that we have a good migration path. Thanks!

src/Symfony/Component/PropertyInfo/Tests/TypeTest.php
chalasr: not resolved it seems

mtarld: Indeed, I missed that part, thanks for poiting out :slightly_smiling_face:

mtarld:
@dunglas, I tried to use the type from TypeInfo component within API Platform, and so far I don't see any blocker. Indeed, the same BC/deprecation tricks in this PR can be done as well in API Platform.

But, it is plenty of work, I'll try to send a draft PR soon.

chalasr (approved):
Much better. 
I think this should make it in 7.1 so that we have at least one usage of the new type-info component internally, and we can iterate.

src/Symfony/Component/PropertyInfo/Extractor/ConstructorExtractor.php
```diff
@@ -28,8 +29,24 @@ public function __construct(
     ) {
     }
 
+    public function getType(string $class, string $property, array $context = []): ?Type
+    {
+        foreach ($this->extractors as $extractor) {
+            if (null !== $value = $value = $extractor->getTypeFromConstructor($class, $property)) {
```
nicolas-grekas:
```suggestion
            if (null !== $value = $extractor->getTypeFromConstructor($class, $property)) {
```

nicolas-grekas (approved):
Let's see how this goes :)

fabpot:
Thank you so much @mtarld, great work here

PR #53163, opened 2023-12-20
[Contracts][DependencyInjection] Add `ServiceCollectionInterface`

kbond:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | n/a
| License       | MIT

I commonly require a service locator which is also an iterator. I noticed during @weaverryan's most recent livestream he could have benefited from this as well. There are currently three ways to achieve:

1. Decorate a `ServiceLocator` and use `ServiceLocator::getProvidedServices()` in the decorator's `getIterator()`
2. Decorate a service iterable and add your own `ContainerInterface::get()` method which iterates over all items until it finds a match (this is what Ryan did in his livestream)
3. Inject a tagged iterator in addition to your locator.

This PR proposes a new `ServiceCollectionInterface` which is a countable, iterable, `ContainerInterface`.

I'm implemented on the current `ServiceLocator` so any place this is used in your code, you can now type-hint `ServiceCollectionInterface`:

```php
public function someAction(
    #[TaggedLocator('some.tag')]
    ServiceCollectionInterface $container,
) {
    foreach ($container as $id => $item) {
        // do something with each service/id
    }

    $container->count(); // # of services
    
    $container->get('some.id');
}
```

### DIFF
--- a/src/Symfony/Component/DependencyInjection/ServiceLocator.php
+++ b/src/Symfony/Component/DependencyInjection/ServiceLocator.php
@@ -16,8 +16,8 @@ use Psr\Container\NotFoundExceptionInterface;
 use Symfony\Component\DependencyInjection\Exception\RuntimeException;
 use Symfony\Component\DependencyInjection\Exception\ServiceCircularReferenceException;
 use Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException;
+use Symfony\Contracts\Service\ServiceCollectionInterface;
 use Symfony\Contracts\Service\ServiceLocatorTrait;
-use Symfony\Contracts\Service\ServiceProviderInterface;
 use Symfony\Contracts\Service\ServiceSubscriberInterface;
 
 /**
@@ -26,9 +26,9 @@ use Symfony\Contracts\Service\ServiceSubscriberInterface;
  *
  * @template-covariant T of mixed
  *
- * @implements ServiceProviderInterface<T>
+ * @implements ServiceCollectionInterface<T>
  */
-class ServiceLocator implements ServiceProviderInterface, \Countable
+class ServiceLocator implements ServiceCollectionInterface
 {
     use ServiceLocatorTrait {
         get as private doGet;
@@ -80,6 +80,13 @@ class ServiceLocator implements ServiceProviderInterface, \Countable
     public function count(): int
     {
         return \count($this->getProvidedServices());
+    }
+
+    public function getIterator(): \Traversable
+    {
+        foreach ($this->getProvidedServices() as $id => $config) {
+            yield $id => $this->get($id);
+        }
     }
 
     private function createNotFoundException(string $id): NotFoundExceptionInterface
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,7 @@
         "psr/http-message": "^1.0|^2.0",
         "psr/link": "^1.1|^2.0",
         "psr/log": "^1|^2|^3",
-        "symfony/contracts": "^2.5|^3.0",
+        "symfony/contracts": "^3.5",
         "symfony/polyfill-ctype": "~1.8",
         "symfony/polyfill-intl-grapheme": "~1.0",
         "symfony/polyfill-intl-icu": "~1.0",
@@ -206,7 +206,7 @@
             "url": "src/Symfony/Contracts",
             "options": {
                 "versions": {
-                    "symfony/contracts": "3.4.x-dev"
+                    "symfony/contracts": "3.5.x-dev"
                 }
             }
         },
--- a/src/Symfony/Component/DependencyInjection/CHANGELOG.md
+++ b/src/Symfony/Component/DependencyInjection/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * Add argument `$prepend` to `ContainerConfigurator::extension()` to prepend the configuration instead of appending it
+ * Have `ServiceLocator` implement `ServiceCollectionInterface`
 
 7.0
 ---
--- a/src/Symfony/Component/DependencyInjection/Tests/ServiceLocatorTest.php
+++ b/src/Symfony/Component/DependencyInjection/Tests/ServiceLocatorTest.php
@@ -101,6 +101,17 @@ class ServiceLocatorTest extends ServiceLocatorTestCase
             'baz' => '?string',
         ]);
     }
+
+    public function testIsCountableAndIterable()
+    {
+        $locator = $this->getServiceLocator([
+            'foo' => fn () => 'bar',
+            'bar' => fn () => 'baz',
+        ]);
+
+        $this->assertCount(2, $locator);
+        $this->assertSame(['foo' => 'bar', 'bar' => 'baz'], iterator_to_array($locator));
+    }
 }
 
 class SomeServiceSubscriber implements ServiceSubscriberInterface
--- a/src/Symfony/Component/DependencyInjection/composer.json
+++ b/src/Symfony/Component/DependencyInjection/composer.json
@@ -19,7 +19,7 @@
         "php": ">=8.2",
         "psr/container": "^1.1|^2.0",
         "symfony/deprecation-contracts": "^2.5|^3",
-        "symfony/service-contracts": "^3.3",
+        "symfony/service-contracts": "^3.5",
         "symfony/var-exporter": "^6.4|^7.0"
     },
     "require-dev": {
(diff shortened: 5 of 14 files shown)


src/Symfony/Component/DependencyInjection/ServiceCollectionInterface.php
```diff
@@ -0,0 +1,26 @@
...
+namespace Symfony\Component\DependencyInjection;
+
+use Psr\Container\ContainerInterface as PsrContainerInterface;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ *
+ * @template-covariant T of mixed
+ *
+ * @extends PsrContainerInterface<T>
```
stof: This is wrong to me given that PsrContainerInterface is not defined as a generic interface.

src/Symfony/Component/DependencyInjection/ServiceCollectionInterface.php
```diff
@@ -0,0 +1,25 @@
...
+use Psr\Container\ContainerInterface as PsrContainerInterface;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ *
+ * @template-covariant T of mixed
+ *
+ * @extends \IteratorAggregate<string, T>
+ */
+interface ServiceCollectionInterface extends PsrContainerInterface, \IteratorAggregate, \Countable
```
ro0NL: extra interface needed?

why not make just ServiceLocator a container+iterator?

src/Symfony/Component/DependencyInjection/ServiceLocator.php
```diff
@@ -27,8 +27,9 @@
  * @template-covariant T of mixed
  *
  * @implements ServiceProviderInterface<T>
+ * @implements ServiceCollectionInterface<T>
  */
-class ServiceLocator implements ServiceProviderInterface, \Countable
+class ServiceLocator implements ServiceProviderInterface, ServiceCollectionInterface
```
GromNaN: As said by @ro0NL, we already have interfaces to cover all the features.
```suggestion
class ServiceLocator implements ServiceProviderInterface, \Countable, \ IteratorAggregate
```

GromNaN (commented):
> I commonly require a service locator which is also an iterator.

Could you provide examples where you need the service locator to be iterable. Maybe a link to the video with a point in time where "he could have benefited from this".

kbond:
> Could you provide examples where you need the service locator to be iterable.

Here's a good example: https://github.com/symfony/ux/pull/1347/files#diff-6487ef7c39e106956904930b49958df26d90de6e025cd0c69a1a9a84e31ed064 (the entire class could be replaced with a `ServiceLocator`). He needs to iterate over a _collection of component services_ for the _index_ page, then, on each _component_ page, he needs to _get_ a specific one.

src/Symfony/Component/DependencyInjection/ServiceCollectionInterface.php
kbond: > extra interface needed?

🤔 Maybe not. I was thinking it is better to type-hint/inject an interface (ie `ServiceProviderInterface`) rather than an implementation (ie `ServiceLocator`)

I know you can typehint `ServiceProviderInterface&\Countable&\IteratorAggregate` but that's a bummer.

> why not make ServiceLocator a container+iterator?

I have no problem with this if that's the consensus.

ro0NL: no need to interface value objects IMHO

kbond:
Ok, I've removed the interface and just made `ServiceLocator` iterable.

yceruto (approved):

derrabus:
> There are currently two ways to achieve:

3. Inject a tagged iterator in addition to your locator.

src/Symfony/Component/DependencyInjection/ServiceLocator.php
```diff
@@ -82,6 +83,13 @@ public function count(): int
         return \count($this->getProvidedServices());
     }
 
+    public function getIterator(): \Traversable
+    {
+        foreach (array_keys($this->getProvidedServices()) as $id) {
+            yield $id => $this->get($id);
+        }
```
derrabus: This basically means that by iterating over the locator you initialize all of its services. To me, this kinda defeats the purpose of a service locator: I could've just injected a tagged iterator and called `iterator_to_array()` on it.

_If_ we really want to make the locator traversable (not sure if we should yet), maybe we'd rather yield a service closure instead?

nicolas-grekas: but because this is a generator, this still does it kinda lazily so it's OK?

src/Symfony/Component/DependencyInjection/ServiceLocator.php
```diff
@@ -82,6 +83,13 @@ public function count(): int
         return \count($this->getProvidedServices());
     }
 
+    public function getIterator(): \Traversable
+    {
+        foreach (array_keys($this->getProvidedServices()) as $id) {
```
nicolas-grekas: array_keys is overhead

nicolas-grekas (commented):
The added method is just a helper: userland can do this iteration with any `ServiceProviderInterface`.

IoC principle would encourage using the interface.

Still worth it?

nicolas-grekas:
> [symfony/ux#1347 (files)](https://github.com/symfony/ux/pull/1347/files#diff-6487ef7c39e106956904930b49958df26d90de6e025cd0c69a1a9a84e31ed064) (the entire class could be replaced with a ServiceLocator)

that's quite common needless overhead - we already simplified many components by using a container
this PR is not needed to kill that class :)

src/Symfony/Component/DependencyInjection/ServiceLocator.php
derrabus: I don't know, this still looks like a pitfall to me. I wouldn't do it.

kbond:
> that's quite common needless overhead - we already simplified many components by using a container
this PR is not needed to kill that class :)

I guess you mean, inject a tagged iterator into the index action and the tagged locator into the show method? Indeed that would work.

> The added method is just a helper: userland can do this iteration with any ServiceProviderInterface.

This is true, and in fact what I've been doing, but in userland, `ServiceProviderInterface::getProvidedServices()` is weird. It's keyed by the service id which, imo, is 95% of the reason to call by end users (the array values aren't important at this stage). It's a bit of a papercut imo as I don't think it's entirely clear this is possible (iterating over all the services).

What about a `ServiceProviderInterface::getProvidedServiceIds(): string[]` method?

> Still worth it?

I do still feel this improves the DX as the need for a service locator/iterator is a common pattern I'm seeing (but let's close if there isn't agreement).

src/Symfony/Component/DependencyInjection/ServiceLocator.php
nicolas-grekas: @kbond WDYT of this concern? I don't feel the need for this extra safety myself but maybe your experience could help best decide?

nicolas-grekas:
To make this full "IoC", one can type hint `ServiceProviderInteface&Traversable`, isn't it? Then LGTM.

kbond:
As discussed with Nicolas on Slack, I've switched back to a `ServiceCollectionInterface` (in `symfony/service-contracts`).

src/Symfony/Component/DependencyInjection/ServiceLocator.php
kbond: > I don't know, this still looks like a pitfall to me. I wouldn't do it.

Does the new interface alleviate the concern at all?

nicolas-grekas (approved):
Works for me, thanks.
Please update the root composer.json for 3.5 also.

kbond:
> Please update the root composer.json for 3.5 also.

Done, but please confirm it's updated correctly.

nicolas-grekas (approved):

chalasr (approved):

src/Symfony/Component/DependencyInjection/ServiceLocator.php
```diff
@@ -26,9 +26,10 @@
  *
  * @template-covariant T of mixed
  *
- * @implements ServiceProviderInterface<T>
+ * @implements ServiceCollectionInterface<T>
+ * @implements \IteratorAggregate<string, T>
```
xabbuh: Can be removed as we inherit this information from the `ServiceCollectionInterface`, right?

chalasr:
Thank you @kbond.

PR #53172, opened 2023-12-21
[SecurityBundle] Prevent to login/logout without a request context

### LINKED ISSUE #53170: [SecurityBundle] Using `Security::login()` in a CLI context throws a `TypeError`

+6.2, 7.0

### Description

Using `Security::login()` in a context without request *(e.g.: CLI)*, throws a type error :
```
TypeError: Security::getFirewallConfig():
Argument #1 ($request) must be of type Symfony\Component\HttpFoundation\Request, null given,
called in Security.php on line 115
```

### How to reproduce

1. Create a PHPUnit test extending the KernelTestCase.
    ```php
    <?php
    // tests/SecurityTest.php
    
    namespace Tests;
    
    use App\Entity\User;
    use Doctrine\ORM\EntityManagerInterface;
    use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
    use Symfony\Bundle\SecurityBundle\Security;
    
    class SecurityTest extends KernelTestCase
    {
        public function test(): void
        {
            /** @var Security $security */
            $security = self::getContainer()->get(Security::class);
            /** @var EntityManagerInterface $security */
            $em = self::getContainer()->get(EntityManagerInterface::class);
            // fetch your user that implements UserInterface
            $user = $em->getRepository(User::class)->find(1);
            $security->login($user);
        }
    }
    ```
2. Run that test with :
    ```sh
    vendor/bin/phpunit tests/SecurityTest.php
    ```
3. We get this fatal error :
    ```
    There was 1 error:
    
    1) Tests\SecurityTest::test
    TypeError: Symfony\Bundle\SecurityBundle\Security::getFirewallConfig():
    Argument #1 ($request) must be of type Symfony\Component\HttpFoundation\Request, null given,
    called in vendor/symfony/security-bundle/Security.php on line 115
    ```

### Possible Solution

Should we prevent to use `Security::login()` without a request ?
```php
// Security.php

public function login(/*...*/): ?Response
{
    $request = $this->container->get('request_stack')->getCurrentRequest();
    if (!$request) {
        throw new LogicException('Unable to login without a request.');
    }
     // ...
}
```

symfonyaml:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #53170
| License       | MIT

Using `Security::login()` in a context without request throws a type error
See all details in the issue https://github.com/symfony/symfony/issues/53170

In this PR, we prevent to use `Security::login()` and `Security::logout()` without a request context, to avoid a fatal error.

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/Security.php
+++ b/src/Symfony/Bundle/SecurityBundle/Security.php
@@ -60,6 +60,10 @@ class Security extends LegacySecurity
     public function login(UserInterface $user, string $authenticatorName = null, string $firewallName = null): void
     {
         $request = $this->container->get('request_stack')->getCurrentRequest();
+        if (null === $request) {
+            throw new LogicException('Unable to login without a request context.');
+        }
+
         $firewallName ??= $this->getFirewallConfig($request)?->getName();
 
         if (!$firewallName) {
@@ -83,14 +87,17 @@ class Security extends LegacySecurity
      */
     public function logout(bool $validateCsrfToken = true): ?Response
     {
+        $request = $this->container->get('request_stack')->getMainRequest();
+        if (null === $request) {
+            throw new LogicException('Unable to logout without a request context.');
+        }
+
         /** @var TokenStorageInterface $tokenStorage */
         $tokenStorage = $this->container->get('security.token_storage');
 
         if (!($token = $tokenStorage->getToken()) || !$token->getUser()) {
             throw new LogicException('Unable to logout as there is no logged-in user.');
         }
-
-        $request = $this->container->get('request_stack')->getMainRequest();
 
         if (!$firewallConfig = $this->container->get('security.firewall.map')->getFirewallConfig($request)) {
             throw new LogicException('Unable to logout as the request is not behind a firewall.');
--- a/src/Symfony/Bundle/SecurityBundle/Tests/SecurityTest.php
+++ b/src/Symfony/Bundle/SecurityBundle/Tests/SecurityTest.php
@@ -201,6 +201,28 @@ class SecurityTest extends TestCase
         $security->login($user);
     }
 
+    public function testLoginWithoutRequestContext()
+    {
+        $requestStack = new RequestStack();
+        $user = $this->createMock(UserInterface::class);
+
+        $container = $this->createMock(ContainerInterface::class);
+        $container
+            ->expects($this->atLeastOnce())
+            ->method('get')
+            ->willReturnMap([
+                ['request_stack', $requestStack],
+            ])
+        ;
+
+        $security = new Security($container, ['main' => null]);
+
+        $this->expectException(\LogicException::class);
+        $this->expectExceptionMessage('Unable to login without a request context.');
+
+        $security->login($user);
+    }
+
     public function testLogout()
     {
         $request = new Request();
@@ -405,6 +427,27 @@ class SecurityTest extends TestCase
 
         $this->assertInstanceOf(Response::class, $response);
         $this->assertEquals('a custom response', $response->getContent());
+    }
+
+    public function testLogoutWithoutRequestContext()
+    {
+        $requestStack = new RequestStack();
+
+        $container = $this->createMock(ContainerInterface::class);
+        $container
+            ->expects($this->atLeastOnce())
+            ->method('get')
+            ->willReturnMap([
+                ['request_stack', $requestStack],
+            ])
+        ;
+
+        $security = new Security($container, ['main' => null]);
+
+        $this->expectException(\LogicException::class);
+        $this->expectExceptionMessage('Unable to logout without a request context.');
+
+        $security->logout();
     }
 
     private function createContainer(string $serviceId, object $serviceObject): ContainerInterface


src/Symfony/Bundle/SecurityBundle/Security.php
```diff
@@ -59,7 +59,9 @@ public function getFirewallConfig(Request $request): ?FirewallConfig
      */
     public function login(UserInterface $user, string $authenticatorName = null, string $firewallName = null): void
     {
-        $request = $this->container->get('request_stack')->getCurrentRequest();
+        if (!$request = $this->container->get('request_stack')->getCurrentRequest()) {
```
Nyholm: Let's be explicit that we expect null here. We should also do it in two lines so it is easier to read.

```suggestion
        $request = $this->container->get('request_stack')->getCurrentRequest();
        if (null === $request) {
```

Nyholm (commented):
Thank you. 

I appreciate this change. Can you fix the CS so it is more similar to the rest of the code? I made a comment on the `login()` but I want you to do the same thing on the `logout()`. 

Could I also ask you to write a test for this?

src/Symfony/Bundle/SecurityBundle/Security.php
symfonyaml: Hi @Nyholm , I appreciate your help.
 - I fixed the CS as you suggested.
 - I added a test for exceptions in `login()` and `logout()`

I realized that [**login** uses **getCurrentRequest**](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Bundle/SecurityBundle/Security.php#L86), and [**logout** uses **getMainRequest**](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Bundle/SecurityBundle/Security.php#L118). Do you think this inconstancy is on purpose ?

Nyholm: Lovely!

> I realized that [login uses getCurrentRequest](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Bundle/SecurityBundle/Security.php#L86), and [logout uses getMainRequest](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Bundle/SecurityBundle/Security.php#L118). Do you think this inconstancy is on purpose ?

Yes, that is on purpose. I guess "logout" should logout from "all". But "login" needs the context from the current request and not the main request.

symfonyaml: ok, makes sense now :)

Nyholm (approved):

src/Symfony/Bundle/SecurityBundle/Tests/SecurityTest.php
```diff
@@ -201,6 +201,30 @@ public function testLoginWithoutAuthenticatorThrows()
         $security->login($user);
     }
 
+    public function testLoginWithoutRequestContext()
+    {
+        $requestStack = $this->createMock(RequestStack::class);
```
xabbuh:
```suggestion
        $requestStack = new RequestStack();
```

(no need for the mock here)

src/Symfony/Bundle/SecurityBundle/Tests/SecurityTest.php
```diff
@@ -201,6 +201,30 @@ public function testLoginWithoutAuthenticatorThrows()
...
+        $user = $this->createMock(UserInterface::class);
+
+        $container = $this->createMock(ContainerInterface::class);
```
xabbuh, line 209: I would also not mock the `ContainerInterface` but use the `ContainerBuilder` class instead

symfonyaml: @xabbuh done ! thanks

src/Symfony/Bundle/SecurityBundle/Tests/SecurityTest.php
symfonyaml: @xabbuh done ! thanks

src/Symfony/Bundle/SecurityBundle/Tests/SecurityTest.php
symfonyaml: All the tests in this class mock the `ContainerInterface`, I am confused why we would not do it for this case ?

Nyholm (approved):
Im happy with this PR. 

The two new tests is written in the same style like other tests in this class. 

I aslo think it is a "bugfix" and not a feature because it solves a hard `TypeError`. 

@xabbuh Let us know if we should update the PR and use a real container instead of the mocked one.

xabbuh (approved):
Fine for me then. We can change tests in a separate PR.

chalasr (approved):

marac19901990:
I ended here somehow. And since here I wanted to be useful so i ran the code and the tests, I hope an extra pair of hands doesn't hurt. It all works and the code is clean. I really prefer the explicit $request === null comparison. Sorry for intruding but since I'm here wanted to support the solution.

Nyholm:
Thank you. Congratulations to your first Symfony contribution!! And also thank you for all the reviews

symfonyaml:
It's an honour, thank you very much !

lon9man:
Dear maintainers, 
I have the case and question.

case:
thinking to migrate from laminas to symfony (due to their deprecation roadmap), so investigating functionality currently.
we have background workers, each worker should process business logic using some concrete virtually logged-in user.

question:
what is the suggested method to login user inside cli-command and then to use it everywhere as it works on web-mode?

thanks!

OskarStark:
Can you please open a new issue ? Thanks

PR #53176, opened 2023-12-21
Fix and add test

OskarStark:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | --
| License       | MIT

The exception is not raised on `validate` but on instantiating the constraint.
I also added a test for an unknown format in an array.


derrabus (approved):

Nyholm:
Thank you Oskar.

PR #53177, opened 2023-12-21
[VarExporter] preventing return by reference error in `LazyGhostTrait`

draevCS:
this can throws
"Only variable references should be returned by reference"  
when trying to load "metadataFactory"

### DIFF
--- a/src/Symfony/Component/VarExporter/LazyGhostTrait.php
+++ b/src/Symfony/Component/VarExporter/LazyGhostTrait.php
@@ -148,7 +148,8 @@ trait LazyGhostTrait
             }
             $accessor = Registry::$classAccessors[$scope] ??= Registry::getClassAccessors($scope);
 
-            return $accessor['get']($this, $name, null !== $readonlyScope);
+            $var = $accessor['get']($this, $name, null !== $readonlyScope);
+            return $var;
         } catch (\Error $e) {
             if (\Error::class !== $e::class || !str_starts_with($e->getMessage(), 'Cannot access uninitialized non-nullable property')) {
                 throw $e;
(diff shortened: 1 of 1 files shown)


derrabus:
Thank you. Since this is a bugfix, please fix it on the oldest maintained branch that is affected by the bug. Also, if you could provide a test that fails without your modifications, we can prevent future regressions.

src/Symfony/Component/VarExporter/LazyGhostTrait.php
```diff
@@ -148,7 +148,8 @@ public function &__get($name): mixed
             }
             $accessor = Registry::$classAccessors[$scope] ??= Registry::getClassAccessors($scope);
 
-            return $accessor['get']($this, $name, null !== $readonlyScope);
+            $var = $accessor['get']($this, $name, null !== $readonlyScope);
+            return $var;
```
derrabus, line 152: Can we leave a clarifying comment for the next person who attempts to inline that variable?

nicolas-grekas:
Can you please provide a reproducer for the issue?
To me this should not be needed because the callback is already taking care of this concern:
https://github.com/symfony/symfony/blob/7f3d8c6b613f8afb1e6c2ca1c060491df200bb1f/src/Symfony/Component/VarExporter/Internal/LazyObjectRegistry.php#L93-L100

I'm wondering why this isn't enough.

nicolas-grekas:
Also this break references when one is expected to be returned so we need more info to understand the root cause and find the proper solution.

draevCS:
Sorry, this reference thing is a bit over my head, i was just trying to update a demo project to the latest version, and tracked down a the problem to this line of code. It will be hard for me to reproduce it in a simple example.

I tried to look up your unit tests for LazyGhostTrait but did not find an actual test,
where the code tries to Get and an Object property (sub object ideally, not just an int) and validate that it is indeed a reference.

nicolas-grekas:
I'm closing because this cannot be the fix.
Please submit an issue with as many info as possible to help reproduce the issue.
Then we'll be able to dissect the problem and figure out the solution.

PR #53178, opened 2023-12-21
[Translation][Crowdin] Use project language mapping

andrii-bodnar:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Sometimes the locale codes may be different from the [codes](https://developer.crowdin.com/language-codes/) used in the Crowdin API. This will cause API errors because the wrong code is passed in the upload and download translation requests. There was already a fix (#50040), but it only works partially.

The suggested approach uses the [language mapping](https://support.crowdin.com/project-settings/#languages) configured in the Crowdin project. If the locale code and the crowdin code are different, the user needs to configure the language mapping in the Crowdin project.

Documentation - https://github.com/symfony/symfony-docs/pull/19325

### DIFF
--- a/src/Symfony/Component/Translation/Bridge/Crowdin/CrowdinProvider.php
+++ b/src/Symfony/Component/Translation/Bridge/Crowdin/CrowdinProvider.php
@@ -56,10 +56,13 @@ final class CrowdinProvider implements ProviderInterface
     public function write(TranslatorBagInterface $translatorBag): void
     {
         $fileList = $this->getFileList();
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Thank you @andrii-bodnar.

andrii-bodnar:
@nicolas-grekas done

andrii-bodnar:
@pculka FYI. It may be necessary to set the language mapping in your project after the update.

PR #53181, opened 2023-12-22
[Clock] Fix calling `mockTime()` in `setUpBeforeClass()`

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Currently, calling `mockTime()` in `setUpBeforeClass()` results in the static `$originalClock` variable always being an instance of `MockClock`.

Since `setUpBeforeClass()` is called **before** `saveClockBeforeTest()`, the native clock is never saved. However, by calling `mockTime()` in `setUpBeforeClass()`, the `Clock::set()` method is called, and the clock is set to an instance of `MockClock`. Then, when `saveClockBeforeTest()` is called, it sets the static `$originalClock` variable to that `MockClock` instance.

The problem becomes more noticeable on Symfony 6.4 where `MockClock` uses `DatePoint`, which itself uses `Clock::get()`:
https://github.com/symfony/symfony/blob/d67685897f2246fd22bf8b2ab684e38c14666079/src/Symfony/Component/Clock/DatePoint.php#L24-L27

### DIFF
--- a/src/Symfony/Component/Clock/Test/ClockSensitiveTrait.php
+++ b/src/Symfony/Component/Clock/Test/ClockSensitiveTrait.php
@@ -42,13 +42,19 @@ trait ClockSensitiveTrait
     }
 
     /**
+     * @beforeClass
+     *
      * @before
      *
      * @internal
      */
-    protected static function saveClockBeforeTest(bool $save = true): ClockInterface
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Thank you @HypeMC.

PR #52402, opened 2023-11-01
[Tests] Streamline

OskarStark:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | --
| License       | MIT

Follows
* #52157
* #52365

PR #53183, opened 2023-12-22
[Messenger] PhpSerializer: TypeError should throw `MessageDecodingFailedException`

B-Galati:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #... 
| License       | MIT

At the moment, some unserialization issues don't throw a `MessageDecodingFailedException` which prevent transport to trigger their decoding failure logic.

There are many possibilities to fix the issue. I've implemented a very conservative one. I believe we could simplify the code by a lot if we rely on the global error handler in user land.

### DIFF
--- a/src/Symfony/Component/Messenger/Tests/Fixtures/DummyMessageTyped.php
+++ b/src/Symfony/Component/Messenger/Tests/Fixtures/DummyMessageTyped.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Symfony\Component\Messenger\Tests\Fixtures;
+
+class DummyMessageTyped implements DummyMessageInterface
+{
+    private string $message;
+
+    public function __construct(string $message)
+    {
+        $this->message = $message;
+    }
+
+    public function getMessage(): string
+    {
+        return $this->message;
+    }
+}
--- a/src/Symfony/Component/Messenger/Tests/Transport/Serialization/PhpSerializerTest.php
+++ b/src/Symfony/Component/Messenger/Tests/Transport/Serialization/PhpSerializerTest.php
@@ -16,6 +16,7 @@ use Symfony\Component\Messenger\Envelope;
 use Symfony\Component\Messenger\Exception\MessageDecodingFailedException;
 use Symfony\Component\Messenger\Stamp\NonSendableStampInterface;
 use Symfony\Component\Messenger\Tests\Fixtures\DummyMessage;
+use Symfony\Component\Messenger\Tests\Fixtures\DummyMessageTyped;
 use Symfony\Component\Messenger\Transport\Serialization\PhpSerializer;
 
 class PhpSerializerTest extends TestCase
@@ -33,20 +34,20 @@ class PhpSerializerTest extends TestCase
 
     public function testDecodingFailsWithMissingBodyKey()
     {
+        $serializer = new PhpSerializer();
+
         $this->expectException(MessageDecodingFailedException::class);
         $this->expectExceptionMessage('Encoded envelope should have at least a "body", or maybe you should implement your own serializer');
-
-        $serializer = new PhpSerializer();
 
         $serializer->decode([]);
     }
 
     public function testDecodingFailsWithBadFormat()
     {
+        $serializer = new PhpSerializer();
+
         $this->expectException(MessageDecodingFailedException::class);
         $this->expectExceptionMessageMatches('/Could not decode/');
-
-        $serializer = new PhpSerializer();
 
         $serializer->decode([
             'body' => '{"message": "bar"}',
@@ -55,10 +56,10 @@ class PhpSerializerTest extends TestCase
 
     public function testDecodingFailsWithBadBase64Body()
     {
+        $serializer = new PhpSerializer();
+
         $this->expectException(MessageDecodingFailedException::class);
         $this->expectExceptionMessageMatches('/Could not decode/');
-
-        $serializer = new PhpSerializer();
 
         $serializer->decode([
             'body' => 'x',
@@ -67,10 +68,10 @@ class PhpSerializerTest extends TestCase
 
     public function testDecodingFailsWithBadClass()
     {
+        $serializer = new PhpSerializer();
+
         $this->expectException(MessageDecodingFailedException::class);
         $this->expectExceptionMessageMatches('/class "ReceivedSt0mp" not found/');
-
-        $serializer = new PhpSerializer();
 
         $serializer->decode([
             'body' => 'O:13:"ReceivedSt0mp":0:{}',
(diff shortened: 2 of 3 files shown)


B-Galati:
Should it target version `5.4` actually?

src/Symfony/Component/Messenger/Tests/Transport/Serialization/PhpSerializerTest.php
```diff
@@ -100,6 +100,19 @@ public function testNonUtf8IsBase64Encoded()
...
+    {
+        $this->expectException(MessageDecodingFailedException::class);
+        $this->expectExceptionMessageMatches('/Could not decode/');
+
+        $serializer = $this->createPhpSerializer();
+        $encodedEnvelope = $serializer->encode(new Envelope(new DummyMessage('true')));
+        // Simulate a change of property type in the code base
+        $encodedEnvelope['body'] = str_replace('s:4:\"true\"', 'b:1', $encodedEnvelope['body']);
+
+        $serializer->decode($encodedEnvelope);
```
OskarStark:
```suggestion
        $serializer = $this->createPhpSerializer();
        $encodedEnvelope = $serializer->encode(new Envelope(new DummyMessage('true')));
        // Simulate a change of property type in the code base
        $encodedEnvelope['body'] = str_replace('s:4:\"true\"', 'b:1', $encodedEnvelope['body']);

        $this->expectException(MessageDecodingFailedException::class);
        $this->expectExceptionMessageMatches('/Could not decode/');

        $serializer->decode($encodedEnvelope);
```

src/Symfony/Component/Messenger/Tests/Transport/Serialization/PhpSerializerWithClassNotFoundSupportTest.php
```diff
@@ -22,6 +22,7 @@ class PhpSerializerWithClassNotFoundSupportTest extends PhpSerializerTest
     public function testDecodingFailsWithBadClass()
     {
         $this->expectException(MessageDecodingFailedException::class);
+        $this->expectExceptionMessageMatches('/Could not decode/');
```
OskarStark: Please move the two expect*() calls before the `decode()` class

OskarStark:
> Should it target version 5.4 actually?

Not sure, for me this is a feature and should target 7.1

@nicolas-grekas @fabpot please decide

src/Symfony/Component/Messenger/Tests/Transport/Serialization/PhpSerializerWithClassNotFoundSupportTest.php
B-Galati: Would you like me to fix all the test class at the same time?

OskarStark: That would be awesome 👍
But lets wait for a target branch decision, because if `7.1` is the target, I already have an open PR which does exactly that:
* #52402 

If the decision is a lower branch, yes please

nicolas-grekas:
Changing the type of thrown exceptions is not a bug fix so that's for 7.1 to me.

B-Galati:
@nicolas-grekas I think it should have been handled since the beginning because it's an unserialization issue that currently ends with generic kind of exception instead of `MessageDecodingFailedException`.

I think I do not understand what's the BC break or added functionality here. For me that's a better error handling logic.

Anyway I will let you know when it's ready for 7.1 👍

If you can explain me further why it's not bugfix I would be glad 😊

B-Galati:
Any chance to get some new feedbacks based on my last comment before rebasing?

src/Symfony/Component/Messenger/Transport/Serialization/PhpSerializer.php
```diff
@@ -93,6 +93,12 @@ private function safelyUnserialize(string $contents): Envelope
         try {
             /** @var Envelope */
             $envelope = unserialize($contents);
+        } catch (\Throwable $e) {
+            if ($e instanceof MessageDecodingFailedException) {
+                throw $e;
+            }
+
+            throw new MessageDecodingFailedException(sprintf('Could not decode message using PHP serialization: "%s".', $contents), previous: $e);
```
nicolas-grekas: I wouldn't put the contents in the error message, it can contain sensitive data.

```suggestion
            throw new MessageDecodingFailedException('Could not unserialize Envelope: '.$e->getMessage(), $contents), previous: $e);
```

nicolas-grekas (commented):
works for me on 5.4

src/Symfony/Component/Messenger/Transport/Serialization/PhpSerializer.php
B-Galati: Agree :+1: 

FYI the content is attached to the `$signalException`.

B-Galati: Maybe We should remove it from it as well?

src/Symfony/Component/Messenger/Tests/Fixtures/DummyMessageTyped.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Symfony\Component\Messenger\Tests\Fixtures;
+
+class DummyMessageTyped implements DummyMessageInterface
```
B-Galati, line 5: I had to add a new object with typed property to assess the behavior of the test.

Can be removed with Symfony 6.0+

src/Symfony/Component/Messenger/Tests/Transport/Serialization/PhpSerializerTest.php
```diff
@@ -99,6 +100,22 @@ public function testNonUtf8IsBase64Encoded()
         $this->assertTrue((bool) preg_match('//u', $encoded['body']), 'Encodes non-UTF8 payloads');
         $this->assertEquals($envelope, $serializer->decode($encoded));
     }
+
+    /**
+     * @requires PHP >= 7.4
```
B-Galati: Requirement can be removed with Symfony 6.0+

B-Galati:
@nicolas-grekas @OskarStark thanks for the review, It's ready for 5.4 :+1: 

Let me know for any other updates that should be performed.

nicolas-grekas:
Thank you @B-Galati.

src/Symfony/Component/Messenger/Transport/Serialization/PhpSerializer.php
nicolas-grekas: I removed it while merging.

B-Galati:
Thank @nicolas-grekas!

BTW, I think we could prevent message losses upon `MessageDecodingFailedException` by requeuing the message in the configured `failed_queue`. We could even configure a dedicated queue.

That could be implemented with either a new `MessageDecodingFailedEvent` or by transports directly.

What do you think? It would fix https://github.com/symfony/symfony/issues/44117

I implemented such a behavior for my company thanks to a custom transport but I would prefer to have the behavior built in Symfony.

src/Symfony/Component/Messenger/Transport/Serialization/PhpSerializer.php
```diff
@@ -62,24 +56,30 @@ private function safelyUnserialize(string $contents)
             throw new MessageDecodingFailedException('Could not decode an empty message using PHP serialization.');
         }
 
-        $signalingException = new MessageDecodingFailedException(sprintf('Could not decode message using PHP serialization: %s.', $contents));
         $prevUnserializeHandler = ini_set('unserialize_callback_func', self::class.'::handleUnserializeCallback');
-        $prevErrorHandler = set_error_handler(function ($type, $msg, $file, $line, $context = []) use (&$prevErrorHandler, $signalingException) {
+        $prevErrorHandler = set_error_handler(function ($type, $msg, $file, $line, $context = []) use (&$prevErrorHandler) {
             if (__FILE__ === $file) {
-                throw $signalingException;
+                throw new \ErrorException($msg, 0, $type, $file, $line);
```
B-Galati, line 62: Thanks that's going to be more debug-able like this :+1:

nicolas-grekas:
I don't know, please follow up on an open issue or a PR :)

PR #53184, opened 2023-12-22
[DoctrineBridge] Adapt deprecation message to include ODM bundle attribute name

GromNaN:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | Refs https://github.com/doctrine/DoctrineMongoDBBundle/issues/816
| License       | MIT

In `DoctrineMongoDBBundle`, the attribute to register event listeners is called `AsDocumentListener`, not `AsDoctrineListener` like in `DoctrineBundle`. (see [exploration details](https://github.com/doctrine/DoctrineMongoDBBundle/pull/817#issuecomment-1864872863))

In order to not confuse developers updating their application with the wrong attribute name, this PR adapts the message to the context.

- in `RegisterEventListenersAndSubscribersPass`, it's simple to check [the tagPrefix provided by the bundle](https://github.com/doctrine/DoctrineMongoDBBundle/blob/7fa2155aed7254f17b14182e8568284f45e33034/DoctrineMongoDBBundle.php#L41).
- in `ContainerAwareEventManager`, I havent found anything to provide context. So I removed the attribute name from the message.

### DIFF
--- a/src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
+++ b/src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
@@ -207,8 +207,8 @@ class ContainerAwareEventManager extends EventManager
             if (\is_string($listener)) {
                 $listener = $this->container->get($listener);
             }
-            // throw new \InvalidArgumentException(sprintf('Using Doctrine subscriber "%s" is not allowed, declare it as a listener instead.', \is_object($listener) ? $listener::class : $listener));
-            trigger_deprecation('symfony/doctrine-bridge', '6.3', 'Registering "%s" as a Doctrine subscriber is deprecated. Register it as a listener instead, using e.g. the #[AsDoctrineListener] attribute.', \is_object($listener) ? get_debug_type($listener) : $listener);
+            // throw new \InvalidArgumentException(sprintf('Using Doctrine subscriber "%s" is not allowed. Register it as a listener instead, using e.g. the #[AsDoctrineListener] or #[AsDocumentListener] attribute.', \is_object($listener) ? $listener::class : $listener));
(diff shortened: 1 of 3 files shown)


src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
```diff
@@ -208,7 +208,7 @@ private function initializeSubscribers(): void
                 $listener = $this->container->get($listener);
             }
             // throw new \InvalidArgumentException(sprintf('Using Doctrine subscriber "%s" is not allowed, declare it as a listener instead.', \is_object($listener) ? $listener::class : $listener));
```
OskarStark:
```suggestion
            // throw new \InvalidArgumentException(sprintf('Using Doctrine subscriber "%s" is not allowed, declare it as a listener instead, using e.g. the attribute provided by the bundle.', \is_object($listener) ? $listener::class : $listener));
```
This is also helpful in the exception message IMHO

OskarStark:
I think most of the users are using the ORM, so having the name in the deprecation makes sense and is very helpful, I wouldn't remove it for a more generic message, which is correct, but ends up in many research for the users.

So if we are not able to get the right context, what about something like this:
```diff
- Since symfony/doctrine-bridge 6.3: Registering "Symfony\Bridge\Doctrine\Tests\MySubscriber" as a Doctrine subscriber is deprecated. Register it as a listener instead, using e.g. the #[AsDoctrineListener] attribute.
+ Since symfony/doctrine-bridge 6.3: Registering "Symfony\Bridge\Doctrine\Tests\MySubscriber" as a Doctrine subscriber is deprecated. Register it as a listener instead, using e.g. the #[AsDoctrineListener] or #[AsDocumentListener] attribute.
```

src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php
GromNaN: Updated, this will make a conflict during upmerge.

nicolas-grekas:
Thank you @GromNaN.

PR #53187, opened 2023-12-22
[Messenger] Fix using negative delay

### LINKED ISSUE #50833: [Messenger] Negative delay with Doctrine bridge can't be used

J-roen:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #50833
| License       | MIT

Fixes using negative delays when sending messages. Cherry-picked from #53088.

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/DoctrineIntegrationTest.php
+++ b/src/Symfony/Component/Messenger/Bridge/Doctrine/Tests/Transport/DoctrineIntegrationTest.php
@@ -73,11 +73,33 @@ class DoctrineIntegrationTest extends TestCase
             $stmt = $stmt->execute();
         }
 
-        $available_at = new \DateTime($stmt instanceof Result || $stmt instanceof DriverResult ? $stmt->fetchOne() : $stmt->fetchColumn());
+        $availableAt = new \DateTime($stmt instanceof Result || $stmt instanceof DriverResult ? $stmt->fetchOne() : $stmt->fetchColumn());
 
         $now = new \DateTime();
         $now->modify('+60 seconds');
-        $this->assertGreaterThan($now, $available_at);
+        $this->assertGreaterThan($now, $availableAt);
+    }
+
+    public function testSendWithNegativeDelay()
+    {
+        $this->connection->send('{"message": "Hi I am not actually delayed"}', ['type' => DummyMessage::class], -600000);
+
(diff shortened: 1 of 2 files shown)


Nyholm (approved):

symfonyaml (approved):

xabbuh (approved):

fabpot (approved):

fabpot:
Thank you @J-roen.

PR #53191, opened 2023-12-22
[HttpKernel] Allow `#[WithHttpStatus]` and `#[WithLogLevel]` to take effect on interfaces

priyadi:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | 
| License       | MIT

This change lets adding `#[WithHttpStatus]` and `#[WithLogLevel]` to an interface, and exceptions implementing such interface will behave as if `#[WithHttpStatus]` and `#[WithLogLevel]` is attached to itself.

`#[WithHttpStatus]` and `#[WithLogLevel]` were created to have configuration like `config/packages/exceptions.yaml` but using attributes: https://symfony.com/blog/new-in-symfony-6-3-http-exception-attributes

However, unlike `exceptions.yaml`, `#[WithHttpStatus]` and `#[WithLogLevel]` did not previously work with interfaces. Thus, the two ways of configuring have slightly different semantics. This PR fixes the issue.

### DIFF
--- a/src/Symfony/Component/HttpKernel/EventListener/ErrorListener.php
+++ b/src/Symfony/Component/HttpKernel/EventListener/ErrorListener.php
@@ -70,19 +70,9 @@ class ErrorListener implements EventSubscriberInterface
         }
 
         // There's no specific status code defined in the configuration for this exception
-        if (!$throwable instanceof HttpExceptionInterface) {
-            $class = new \ReflectionClass($throwable);
-
-            do {
-                if ($attributes = $class->getAttributes(WithHttpStatus::class, \ReflectionAttribute::IS_INSTANCEOF)) {
-                    /** @var WithHttpStatus $instance */
-                    $instance = $attributes[0]->newInstance();
-
-                    $throwable = HttpException::fromStatusCode($instance->statusCode, $throwable->getMessage(), $throwable, $instance->headers);
-                    $event->setThrowable($throwable);
-                    break;
-                }
-            } while ($class = $class->getParentClass());
+        if (!$throwable instanceof HttpExceptionInterface && $withHttpStatus = $this->getInheritedAttribute($throwable::class, WithHttpStatus::class)) {
+            $throwable = HttpException::fromStatusCode($withHttpStatus->statusCode, $throwable->getMessage(), $throwable, $withHttpStatus->headers);
+            $event->setThrowable($throwable);
         }
 
         $e = FlattenException::createFromThrowable($throwable);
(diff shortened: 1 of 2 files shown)


smnandre:
How would that work with the exception mapping already handled ? Would there not be conflicts / priority troubles there ?

priyadi:
> How would that work with the exception mapping already handled ? Would there not be conflicts / priority troubles there ?

Parent classes are prioritized first, starting from the immediate parent first. Then, interfaces are processed, according to the order of `class_implements()`. There should be no BC break.

smnandre:
If i was not precise enough, i was making allusion to the framework exception map that already exists and... map classes or interfaces to HTTP status / Log level : https://symfony.com/doc/current/reference/configuration/framework.html#exceptions

I just find a bit odd that those two implementations both coexists in the [ErrorListener](https://github.com/symfony/symfony/blob/7.0/src/Symfony/Component/HttpKernel/EventListener/ErrorListener.php) (one before and one after)

```php
foreach ($this->exceptionsMapping as $class => $config) {
    if (!$throwable instanceof $class || !$config['status_code']) {
        continue;
    }
    if (!$throwable instanceof HttpExceptionInterface || $throwable->getStatusCode() !== $config['status_code']) {
        $headers = $throwable instanceof HttpExceptionInterface ? $throwable->getHeaders() : [];
        $throwable = new HttpException($config['status_code'], $throwable->getMessage(), $throwable, $headers);
        $event->setThrowable($throwable);
    }
    break;
}
```

priyadi:
> If i was not precise enough, i was making allusion to the framework exception map that already exists and... map classes or interfaces to HTTP status / Log level : https://symfony.com/doc/current/reference/configuration/framework.html#exceptions

Yes, doing it from the configuration is the traditional way to do it before 6.3, and it is still working as intended.

https://symfony.com/blog/new-in-symfony-6-3-http-exception-attributes

nicolas-grekas:
Looking at the implementation, I think this might be too unusual - ie unexpected. The relationship is unnatural to me.

Why don't you leverage `HttpExceptionInterface` instead?

priyadi:
> Looking at the implementation, I think this might be too unusual - ie unexpected. The relationship is unnatural to me.
> 
> Why don't you leverage `HttpExceptionInterface` instead?

That would require my domain model to depend on HttpKernel.

Without this change, I could still do it using `config/packages/exceptions.yaml` which works with interfaces. But I don't see why it should work using `config/packages/exceptions.yaml` but should not work with attributes.

priyadi:
Added more background information in the PR description.

nicolas-grekas (approved):
Got it, LGTM!

dunglas (approved):

nicolas-grekas:
Thank you @priyadi.

PR #53195, opened 2023-12-23
[HttpKernel] Fix default locale is ignored when `set_locale_from_accept_language` is used

### LINKED ISSUE #53086: Default locale is ignored  when `set_locale_from_accept_language` is used

6.4.0

### Description

Default locale setting is ignored when "set_locale_from_accept_language" is used.

### How to reproduce

I have the following configuration:

```yaml
framework:
    default_locale: en
    set_locale_from_accept_language: true
    set_content_language_from_locale: false
    enabled_locales: ['pl', 'en']
```

Then, by issuing `curl https://localhost` which has following code:

```php
class IndexNoLocaleController extends AbstractController
{
    #[Route('/')]
    public function __invoke(Request $request): Response
    {
        return $this->redirectToRoute('app_default', ['_locale' => $request->getLocale()]);
    }
}
```

I am redirected to `/pl`.

I expect to be redirected to `/en` instead.

Note: changing order of `enabled_locales: ['en', 'pl']` will result in redirection to `/en` as the first value from the array is used as "default", yet still default from the "default_locale" is ignored.

### Possible Solution

The `\Symfony\Component\HttpFoundation\Request::getPreferredLanguage` accepts ['pl', 'en'] as input, as there are no languages set by curl it just falls into:

```php
if (!$preferredLanguages) {
    return $locales[0];
}
```

The call originates from `LocaleListener.php:72`:

```php
private function setLocale(Request $request): void
{
    if ($locale = $request->attributes->get('_locale')) {
        $request->setLocale($locale);
    } elseif ($this->useAcceptLanguageHeader) {
        if ($preferredLanguage = $request->getPreferredLanguage($this->enabledLocales)) {
            $request->setLocale($preferredLanguage);
        }
        $request->attributes->set('_vary_by_language', true);
    }
}
```

In that listener, there is `$defaultLocale` property available, as well as in the request, therefore maybe the solution is to use the default locale rather than picking "first entry" from enabled locales.

### Additional Context

This is an edge case, but can affect mostly Google Bots [as bots do not set accept language headers](https://developers.google.com/search/docs/specialty/international/locale-adaptive-pages?hl=en).

jkobus:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #53086
| License       | MIT

Fix as discussed in #53086

### DIFF
--- a/src/Symfony/Component/HttpKernel/EventListener/LocaleListener.php
+++ b/src/Symfony/Component/HttpKernel/EventListener/LocaleListener.php
@@ -68,8 +68,10 @@ class LocaleListener implements EventSubscriberInterface
     {
         if ($locale = $request->attributes->get('_locale')) {
             $request->setLocale($locale);
-        } elseif ($this->useAcceptLanguageHeader && $this->enabledLocales && ($preferredLanguage = $request->getPreferredLanguage($this->enabledLocales))) {
-            $request->setLocale($preferredLanguage);
+        } elseif ($this->useAcceptLanguageHeader && $this->enabledLocales) {
+            if ($request->getLanguages() && $preferredLanguage = $request->getPreferredLanguage($this->enabledLocales)) {
+                $request->setLocale($preferredLanguage);
+            }
             $request->attributes->set('_vary_by_language', true);
         }
     }
--- a/src/Symfony/Component/HttpKernel/Tests/EventListener/LocaleListenerTest.php
+++ b/src/Symfony/Component/HttpKernel/Tests/EventListener/LocaleListenerTest.php
@@ -130,6 +130,28 @@ class LocaleListenerTest extends TestCase
         $this->assertEquals('fr', $request->getLocale());
     }
 
+    public function testRequestDefaultLocaleIfNoAcceptLanguageHeaderIsPresent()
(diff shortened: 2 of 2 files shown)


smnandre:
I'm having just a doubt about the second line of the "if" 

```
  $request->attributes->set('_vary_by_language', true);
```

It is later used by the ResponseListener to send a Vary header.. so i guess we want to send this header regardless of whether the client send its preferences or not (or that coud generate big cache problems)

(sorry, did not think about that when I wrote the example)

chalasr:
Indeed. Nice catch @smnandre

jkobus:
Thanks, I added a test case for that @smnandre

chalasr (approved):

fabpot (approved):

fabpot:
Isn't something we need to merge in 5.4?

chalasr:
@fabpot yes, good catch.

nicolas-grekas:
Thank you @jkobus.

jkobus:
Thanks, happy new year for you all :confetti_ball:

PR #53113, opened 2023-12-17
[Validator] update Italian translation

garak:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | none
| License       | MIT

Add new Italian translation for trans-unit 110

PR #53196, opened 2023-12-23
[Validator] Add missing italian translation

marac19901990:
[…y/issues/53117](https://github.com/symfony/symfony/issues/53117)

| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | Fix https://github.com/symfony/symfony/issues/53117
| License       | MIT

Added missing italian translation as per issue linked above

### DIFF
--- a/src/Symfony/Component/Validator/Resources/translations/validators.it.xlf
+++ b/src/Symfony/Component/Validator/Resources/translations/validators.it.xlf
@@ -426,6 +426,10 @@
                 <source>Using hidden overlay characters is not allowed.</source>
                 <target>Non è consentito utilizzare caratteri sovrapposti nascosti.</target>
             </trans-unit>
+            <trans-unit id="110">
+                <source>The extension of the file is invalid ({{ extension }}). Allowed extensions are {{ extensions }}.</source>
+                <target>L'estensione del file non è valida ({{ extension }}). Le estensioni consentite sono {{ extensions }}.</target>
+            </trans-unit>
         </body>
     </file>
 </xliff>


src/Symfony/Component/Validator/Resources/translations/validators.it.xlf
```diff
@@ -426,6 +426,10 @@
                 <source>Using hidden overlay characters is not allowed.</source>
                 <target>Non è consentito utilizzare caratteri sovrapposti nascosti.</target>
             </trans-unit>
+            <trans-unit id="110">
+                <source>The extension of the file is invalid ({{ extension }}). Allowed extensions are {{ extensions }}.</source>
+                <target>L'estensione del file non è valida ({{ extension }}). L'estensioni consentite sono {{ extensions }}.</target>
```
Nyholm: What is the difference between this and

```
Estensioni consentite sono {{ extensions }}.
```

Nyholm (commented):
Thank you. 

My Italian is very poor, so I used Google Translate. Can you help with just one comment?

yceruto:
Perhaps @garak could help us to validate this one 🙏

src/Symfony/Component/Validator/Resources/translations/validators.it.xlf
marac19901990: It is the same, L'estensioni i would say is the school way, and Estensioni without the prefix is more like a shorter version, street like... Maybe not a great comparison but something like what's up and wassup. I exagerated a bit with the analogy just to prove the point.

marac19901990: And beside that since I'm bilingual, mother from italy, father from Serbia. There is the thing that when i say it out loud it just sounds better to me, like more serious. Anyway both are correct.

Nyholm: I see. So "L'estensioni" is more formal. 

Thank you for the explanation.

marac19901990: Another example would be: Estensioni or L'estensioni consentite sono i would translate that: L'Estensioni valide sono, but didn't change it since all the others were in the same manner. consentite means allowed, valide means valid, even though it is more logical to use consentite, it just sounds right to say valide instead. Maybe i got too much in depth, sorry. :D

Nyholm (approved):

src/Symfony/Component/Validator/Resources/translations/validators.it.xlf
garak: You can't write "l'estensioni" in Italian. The only possible correct form is "le estensioni". The elision of the article is only used for the singular

marac19901990: Exactly, you said exatly like I meant, I just didn't remember the word formal... :) It is a native language to me. It is properly translated, I didn't use any translators, no need to worry :) My mother is the court translator for italian, I can check with her also, but I know it is right. :D

garak:
This was already implemented in PR #53113

Nyholm:
Ah. I see #53113 was targeting 7.1. That was a misstake. We should target lowest branch. 
To add to the confusion is that 5.4 branch does not use this translations string...

Im sorry for the confusion. I still think we should merge this to be consistent with how we usually do with translations. Unless @fabpot disagrees. 

@garak I see that your translation is slightly different. Could you give your 50 cents on this suggestion?

garak:
@Nyholm the only difference I can spot is "L'estensioni" vs "Le estensioni". As noted in my previous comment, the only correct form in Italian is the latter, since we don't use the elision for plural articles.

marac19901990:
> @Nyholm the only difference I can spot is "L'estensioni" vs "Le estensioni". As noted in my previous comment, the only correct form in Italian is the latter, since we don't use the elision for plural articles.

You are right, checked with my mother also :D It seems I haven't used Italian enough recently :D

marac19901990:
@garak I've changed it and fixed the mistake. Thanks for correcting me. If you sometime have the will to speak a little Italian with me, since I have few ocasions, I would be very thankful...

garak:
> @garak I've changed it and fixed the mistake. Thanks for correcting me. If you sometime have the will to speak a little Italian with me, since I have few ocasions, I would be very thankful...

Sure! You can find me on the Symfony Devs slack channel

fabpot (approved):

fabpot:
Thank you @marac19901990.

PR #53201, opened 2023-12-23
[Security] Reduce log level in case of UserNotFoundException in ContextListener

### LINKED ISSUE #50402: ContextListener is logging warning everytime a user is not found

5.4.0

### Description

Maybe I don't understand the Symfony strategy about log level.
To me a warning is something that is not critical but still need to be checked and fixed.
Like php warnings.

In the case of the ContextListener, a lot of false-positive warning are thrown by this logger:
https://github.com/symfony/symfony/blob/5bf5c4d5b77adc00fefda26dc54eea7f0cede9fd/src/Symfony/Component/Security/Http/Firewall/ContextListener.php#L237

### How to reproduce

- Connect yourself with an account.
- Ask someone else to deleted you account on an admin panel.
- Refresh the page, since your account doesn't exist anymore, the Provider doesn't find your user, throw a NotFoundException
- You're logged out

I consider this as a normal behavior, so no warning shouldn't be logged.
It's expected to not found a user after this one being deleted.

That seems to be one of the purpose of the refreshUser method to returns null:
- If the user changed
- If the user doesn't exist

### Possible Solution

Shouldn't be a debug log or at most an info log ?

VincentLanglet:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Issues        | Fix #50402
| License       | MIT

I didn't get a clear answer for https://github.com/symfony/symfony/issues/50402 so I'm opening a PR about the suggested change to get opinion on it.

There was a similar discussion, some message which could help:
- @fabpot https://github.com/symfony/symfony/pull/1073#issuecomment-1234384
- @stof https://github.com/symfony/symfony/pull/1073#issuecomment-1270189

Here, the `UserNotFoundException` should be considered as a "normal" behavior to me. You can trigger it this way
- Connect yourself with an account.
- Ask someone else to deleted you account on an admin panel.
- Refresh the page, since your account doesn't exist anymore, the Provider doesn't find your user, throw a NotFoundException
- You're logged out

I consider this as a normal behavior, so no warning shouldn't be logged and an info should be preferred.
It's expected to not found a user after this one being deleted.

That seems to be one of the purpose of the refreshUser method to returns null:
- If the user changed
- If the user doesn't exist

WDYT ?

### DIFF
--- a/src/Symfony/Component/Security/Http/Firewall/ContextListener.php
+++ b/src/Symfony/Component/Security/Http/Firewall/ContextListener.php
@@ -234,7 +234,7 @@ class ContextListener extends AbstractListener
             } catch (UnsupportedUserException) {
                 // let's try the next user provider
             } catch (UserNotFoundException $e) {
-                $this->logger?->warning('Username could not be found in the selected user provider.', ['username' => $e->getUserIdentifier(), 'provider' => $provider::class]);
+                $this->logger?->info('Username could not be found in the selected user provider.', ['username' => $e->getUserIdentifier(), 'provider' => $provider::class]);
 
                 $userNotFoundByProvider = true;
             }


xabbuh (approved):
sounds reasonable to me

fabpot (approved):

fabpot:
Thank you @VincentLanglet.

PR #53203, opened 2023-12-24
[Intl] [Emoji] Fix emoji files (remove wrong characters / add missing data)

### LINKED ISSUE #53116: [Intl][Emoji] Arrows instead of emoji (CLDR hierarchical ↑↑↑)

6.4,7.0,7.1

### Description

Hierarchical  metadata (this character:` ↑↑↑`) are copied into the resource files, leading to arrows beeing inserted in the texts.

I took a list of "top 100 emojis" to see how what proportion was concerned and went through all the xx_YY files.

| Locale | Emojis | Fleches | % error
| - | - | - | - | 
| en_CA | 0 | 100 | 🔴 🔴 🔴 🔴 🔴 🔴 🔴 🔴 🔴 🔴  |
| fr_CA | 26 | 74 | 🟢 🟢 🟢 🔴 🔴 🔴 🔴 🔴 🔴 🔴  |
| es_MX | 34 | 66 | 🟢 🟢 🟢 🔴 🔴 🔴 🔴 🔴 🔴 🔴  | 
| en_IN | 0 | 100 | 🔴 🔴 🔴 🔴 🔴 🔴 🔴 🔴 🔴 🔴  |

### How to reproduce

```php

$message = 'Hello 🐨 ❤️ 🤣';

echo (EmojiTransliterator::create('fr'))->transliterate($message).PHP_EOL;
echo (EmojiTransliterator::create('fr_CA'))->transliterate($message).PHP_EOL;

// --> Hello koala cœur rouge️ se rouler par terre de rire
// --> Hello ↑↑↑ ↑↑↑️ rire à se rouler par terre
```

Obviously it impacts the Slugger with Emoji 

```php
$message = '🐨❤️';

echo (new AsciiSlugger('fr'))->withEmoji()->slug($message).PHP_EOL;
echo (new AsciiSlugger('fr_CA'))->withEmoji()->slug($message).PHP_EOL;

// --> koala-coeur-rouge
// --> 
```

### Possible Solution

Something to adapt in the files generation.

### Additional Context

_No response_

smnandre:
| Q             | A
| ------------- | ---
| Branch?       | 6.4 
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #53116
| License       | MIT

Fix two things
* unwanted characters (`↑↑↑`) instead of expected translations
* merging between child and parent locales

Also adapted the code to maintain a shared / reproductible order in the generated files.


smnandre:
How can i tell fabbot to _not_ add Licence or correct typo in the generated files ?

src/Symfony/Component/Intl/Resources/emoji/build.php
```diff
@@ -68,9 +68,16 @@ public static function buildRules(array $emojisCodePoints): Generator
         $ignored = [];
         $mapsByLocale = [];
 
+        // Build a map of all emojis, ordered by code points length
+        $orderedEmojis = [...$emojisCodePoints];
+        uksort($orderedEmojis, fn ($a, $b) => mb_strlen($b) <=> mb_strlen($a));
```
nicolas-grekas:
```suggestion
        uksort($orderedEmojis, fn ($a, $b) => mb_strlen($b, 'utf-8') <=> mb_strlen($a, 'utf-8'));
```

nicolas-grekas (commented):
Is the reordering necessary? Can't we get rid of it? I'd prefer keeping the original order if possible.

smnandre:
> Is the reordering necessary? Can't we get rid of it? I'd prefer keeping the original order if possible.

By "original" you mean the one implemented previously ? 

The problem i see: it was not the same between locales (as soon as parent/child where involved), and a new child addind a translatation for a given emoji would change the message.. and its position in the array.

nicolas-grekas:
I mean the order implemented previously yes, which is the order found in the source IIRC.

smnandre:
The previous files were not ordered like the source ones, they were ordered by mb_strlen then source order. 

```
$codePointsCount = mb_strlen($emoji);
$mapsByLocale[$locale][$codePointsCount][$emoji] = $name;
```

And that order was kept only when no parent/child locale was involved

smnandre:
I can mimic this behaviour to build the "model map" if you want

nicolas-grekas:
I like reducing diffs so if you can then yes please :)

smnandre:
> I like reducing diffs so if you can then yes please :)

Much better now indeed ... only the "child locales" have been updated 

As they were the only one concerned by the two problems (arrows instead of messages, and missing parent messages) this is much more logical. 

Thanks ;)

smnandre:
Still don't know how to handle fabbot false positives ❓

nicolas-grekas:
> Still don't know how to handle fabbot false positives ❓

ignoring them once they are checked :brain:

nicolas-grekas:
Thank you @smnandre.

PR #53211, opened 2023-12-26
[Notifier] Rework the Smsbox bridge

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Rework the whole component to match the rest of the codebase.

### DIFF
--- a/src/Symfony/Component/Notifier/Bridge/Smsbox/SmsboxOptions.php
+++ b/src/Symfony/Component/Notifier/Bridge/Smsbox/SmsboxOptions.php
@@ -11,49 +11,25 @@
 
 namespace Symfony\Component\Notifier\Bridge\Smsbox;
 
+use Symfony\Component\Intl\Countries;
+use Symfony\Component\Notifier\Bridge\Smsbox\Enum\Charset;
+use Symfony\Component\Notifier\Bridge\Smsbox\Enum\Day;
+use Symfony\Component\Notifier\Bridge\Smsbox\Enum\Encoding;
+use Symfony\Component\Notifier\Bridge\Smsbox\Enum\Mode;
+use Symfony\Component\Notifier\Bridge\Smsbox\Enum\Strategy;
+use Symfony\Component\Notifier\Bridge\Smsbox\Enum\Udh;
 use Symfony\Component\Notifier\Exception\InvalidArgumentException;
 use Symfony\Component\Notifier\Message\MessageOptionsInterface;
 
 /**
  * @author Alan Zarli <azarli@smsbox.fr>
  * @author Farid Touil <ftouil@smsbox.fr>
+ * @author Alexandre Daubois <alex.daubois@gmail.com>
  */
 final class SmsboxOptions implements MessageOptionsInterface
 {
-    public const MESSAGE_MODE_STANDARD = 'Standard';
-    public const MESSAGE_MODE_EXPERT = 'Expert';
-    public const MESSAGE_MODE_RESPONSE = 'Reponse';
-
-    public const MESSAGE_STRATEGY_PRIVATE = 1;
-    public const MESSAGE_STRATEGY_NOTIFICATION = 2;
-    public const MESSAGE_STRATEGY_NOT_MARKETING_GROUP = 3;
-    public const MESSAGE_STRATEGY_MARKETING = 4;
-
-    public const MESSAGE_CODING_DEFAULT = 'default';
-    public const MESSAGE_CODING_UNICODE = 'unicode';
-    public const MESSAGE_CODING_AUTO = 'auto';
-
-    public const MESSAGE_CHARSET_ISO_1 = 'iso-8859-1';
-    public const MESSAGE_CHARSET_ISO_15 = 'iso-8859-15';
-    public const MESSAGE_CHARSET_UTF8 = 'utf-8';
-
-    public const MESSAGE_DAYS_MONDAY = 1;
-    public const MESSAGE_DAYS_TUESDAY = 2;
-    public const MESSAGE_DAYS_WEDNESDAY = 3;
-    public const MESSAGE_DAYS_THURSDAY = 4;
-    public const MESSAGE_DAYS_FRIDAY = 5;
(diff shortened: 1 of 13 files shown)


src/Symfony/Component/Notifier/Bridge/Smsbox/SmsboxOptions.php
```diff
@@ -250,7 +250,6 @@ public static function validateDate(string $date): string
 
     public static function validateDateTime(\DateTime $dateTime)
```
nicolas-grekas: we should never use DateTime anymore so we should change all occurences to DateTimeImmutable

src/Symfony/Component/Notifier/Bridge/Smsbox/SmsboxOptions.php
```diff
@@ -250,7 +250,6 @@ public static function validateDate(string $date): string
...
     {
-        \Locale::setDefault('fr');
         $now = new \DateTime();
         $tz = new \DateTimeZone('Europe/Paris');
```
nicolas-grekas: this is super strange validation logic

src/Symfony/Component/Notifier/Bridge/Smsbox/SmsboxOptions.php
nicolas-grekas: also, all public static method should be private to me

nicolas-grekas (commented):
looks like there are non-strict comparisons in the bridge
overall, it looks like the component needs a review to make it up to par with the rest of the codebase, could you have a look?

src/Symfony/Component/Notifier/Bridge/Smsbox/SmsboxOptions.php
```diff
@@ -11,375 +11,203 @@
...
-        $this->options = [];
     }
 
     public function getRecipientId(): null
     {
         return null;
     }
 
-    public function mode(string $mode)
+    public function mode(SmsboxMessageMode $mode): self
```
nicolas-grekas: `@return $this` also where appropriate

src/Symfony/Component/Notifier/Bridge/Smsbox/SmsboxOptions.php
```diff
@@ -11,375 +11,203 @@
...
+    public function dateTime(\DateTimeImmutable $dateTime): self
     {
-        $this->options['charset'] = self::validateCharset($charset);
+        if (isset($this->options['date']) || isset($this->options['heure'])) {
+            throw new InvalidArgumentException(sprintf('Either %1$s::dateTime() or %1$s::date() and %1$s::hour() must be called, but not both.', self::class));
+        }
 
-        return $this;
-    }
+        if ($dateTime < now()) {
```
nicolas-grekas: let's not add a dep on Clock just for validation

src/Symfony/Component/Notifier/Bridge/Smsbox/composer.json
```diff
@@ -25,7 +25,9 @@
     ],
     "require": {
         "php": ">=8.2",
+        "symfony/clock": "^6.4|^7.0",
         "symfony/http-client": "^6.4|^7.0",
+        "symfony/intl": "^6.4|^7.0",
```
nicolas-grekas: why do we need this one?

alexandre-daubois: To use the `Countries` class in SmsboxOptions, this ease the check of the country code validity

src/Symfony/Component/Notifier/Bridge/Smsbox/SmsboxOptions.php
```diff
@@ -11,375 +11,203 @@
...
+    public function hour(string $hour): self
     {
-        $this->options['variable'] = $variable;
+        if (isset($this->options['dateTime'])) {
+            throw new InvalidArgumentException(sprintf('Either %1$s::dateTime() or %1$s::date() and %1$s::hour() must be called, but not both.', self::class));
+        }
 
-        return $this;
-    }
+        if (false === \DateTimeImmutable::createFromFormat('H:i', $hour)) {
```
nicolas-grekas: same everywhere similar:
```suggestion
        if (!\DateTimeImmutable::createFromFormat('H:i', $hour)) {
```

src/Symfony/Component/Notifier/Bridge/Smsbox/SmsboxOptions.php
```diff
@@ -11,375 +11,203 @@
...
+            throw new InvalidArgumentException('Hour must be in HH:MM format.');
```
nicolas-grekas: `\DateMalformedStringException` with the 8.3 polyfill?

src/Symfony/Component/Notifier/Bridge/Smsbox/composer.json
nicolas-grekas: let's make it optional and skip the validation is the class isn't found?

alexandre-daubois:
Comments are addressed. I'll regenerate the expected return types file in a moment 🙂

nicolas-grekas (approved):

alexandre-daubois:
Changed `self` return types to `static` in `SmsboxOptions` which resolved the return type issues. This should be ready now

OskarStark:
Great work, I would prefer to remove the SmsboxMessage prefix from the enums, we have the namespace.

Agree @nicolas-grekas ?

nicolas-grekas:
`Enum` subnamespace?

alexandre-daubois:
I was skeptical to just have a `Day` or `Charset` enum for instance at the root of the component, but with an Enum subnamespace, I agree with the idea. I'm doing the change

alexandre-daubois:
Here you go, enums have been moved to their namespace

fabpot:
Thank you @alexandre-daubois.

PR #53209, opened 2023-12-25
[HttpKernel] Add support for custom HTTP status code for the `#[MapQueryParameter]` attribute

ovidiuenache:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | #52782
| License       | MIT

Issue #52782 was created where the need to have a custom HTTP status code for the MapQueryParameter attribute was stated. As part of the discussion, @nicolas-grekas suggested an approach similar to the implementation of the MapQueryString attribute where such a status code can be provided in the attribute's constructor.

I believe having the possibility to control the status code returned for an invalid and non-nullable parameter with no default value would be a great addition so I took the liberty to create this PR.

This change is backward compatible.

PR #53212, opened 2023-12-26
[HttpKernel] Add `HttpException::fromStatusCode()`

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

A small helper idea that popped up when reviewing #53209

### DIFF
--- a/src/Symfony/Component/HttpKernel/Exception/HttpException.php
+++ b/src/Symfony/Component/HttpKernel/Exception/HttpException.php
@@ -29,6 +29,27 @@ class HttpException extends \RuntimeException implements HttpExceptionInterface
         parent::__construct($message, $code, $previous);
     }
 
+    public static function fromStatusCode(int $statusCode, string $message = '', \Throwable $previous = null, array $headers = [], int $code = 0): self
+    {
+        return match ($statusCode) {
+            400 => new BadRequestHttpException($message, $previous, $code, $headers),
+            403 => new AccessDeniedHttpException($message, $previous, $code, $headers),
+            404 => new NotFoundHttpException($message, $previous, $code, $headers),
+            406 => new NotAcceptableHttpException($message, $previous, $code, $headers),
+            409 => new ConflictHttpException($message, $previous, $code, $headers),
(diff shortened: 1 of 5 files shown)


src/Symfony/Component/HttpKernel/Exception/HttpException.php
```diff
@@ -29,6 +29,27 @@ public function __construct(int $statusCode, string $message = '', \Throwable $p
         parent::__construct($message, $code, $previous);
     }
 
+    public static function fromStatusCode(int $statusCode, string $message = '', \Throwable $previous = null, array $headers = [], int $code = 0): self
+    {
+        return match ($statusCode) {
+            400 => new BadRequestHttpException($message, $previous, $code, $headers),
```
alexandre-daubois, line 35: Shouldn't the method signature follow an arguments order similar to exception?

nicolas-grekas: it follows the constructor instead, which provides local consistency (and the code is the less useful argument so it's better last to me)

fabpot (approved):

alexandre-daubois (approved):

derrabus:
A test? 🥺

nicolas-grekas:
This is already tested by the added code paths to me. More tests would look like testing a getter to me :)

OskarStark (approved):

simondaigre:
Great idea. It will help to ignore some HTTP exceptions on Sentry. 
Could we also plug this with `#[WithHttpStatus]` attribute ?

dunglas (approved):
I would also add a test, with a data provider, it will catch typos etc (returning the an exception not corresponding to the passed parameter).

nicolas-grekas:
> I would also add a test, with a data provider

alright, updated

>  Could we also plug this with #[WithHttpStatus] attribute ?

Already handled, that's what the changes on ErrorListener are about.

PR #53230, opened 2023-12-27
[Notifier] [Bridges] Use CPP

OskarStark:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Notifier/Bridge/AllMySms/AllMySmsTransport.php
+++ b/src/Symfony/Component/Notifier/Bridge/AllMySms/AllMySmsTransport.php
@@ -28,16 +28,13 @@ final class AllMySmsTransport extends AbstractTransport
 {
     protected const HOST = 'api.allmysms.com';
 
-    private string $login;
-    private string $apiKey;
-    private ?string $from;
-
-    public function __construct(string $login, #[\SensitiveParameter] string $apiKey, string $from = null, HttpClientInterface $client = null, EventDispatcherInterface $dispatcher = null)
-    {
-        $this->login = $login;
-        $this->apiKey = $apiKey;
-        $this->from = $from;
-
+    public function __construct(
+        private string $login,
+        #[\SensitiveParameter] private string $apiKey,
+        private ?string $from = null,
+        HttpClientInterface $client = null,
+        EventDispatcherInterface $dispatcher = null,
+    ) {
         parent::__construct($client, $dispatcher);
     }
 
--- a/src/Symfony/Component/Notifier/Bridge/Bandwidth/BandwidthTransport.php
+++ b/src/Symfony/Component/Notifier/Bridge/Bandwidth/BandwidthTransport.php
@@ -37,7 +37,7 @@ final class BandwidthTransport extends AbstractTransport
         private readonly string $applicationId,
         private readonly ?string $priority,
         HttpClientInterface $client = null,
-        EventDispatcherInterface $dispatcher = null
+        EventDispatcherInterface $dispatcher = null,
     ) {
         parent::__construct($client, $dispatcher);
     }
--- a/src/Symfony/Component/Notifier/Bridge/RocketChat/RocketChatTransport.php
+++ b/src/Symfony/Component/Notifier/Bridge/RocketChat/RocketChatTransport.php
@@ -28,15 +28,13 @@ final class RocketChatTransport extends AbstractTransport
 {
     protected const HOST = 'rocketchat.com';
 
-    private string $accessToken;
-    private ?string $chatChannel;
-
-    public function __construct(#[\SensitiveParameter] string $accessToken, string $chatChannel = null, HttpClientInterface $client = null, EventDispatcherInterface $dispatcher = null)
+    public function __construct(
+        #[\SensitiveParameter] private string $accessToken,
+        private ?string $chatChannel = null,
+        HttpClientInterface $client = null,
+        EventDispatcherInterface $dispatcher = null,
+)
     {
-        $this->accessToken = $accessToken;
-        $this->chatChannel = $chatChannel;
-        $this->client = $client;
-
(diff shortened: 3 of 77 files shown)


src/Symfony/Component/Notifier/Bridge/RocketChat/RocketChatTransport.php
```diff
@@ -28,15 +28,13 @@ final class RocketChatTransport extends AbstractTransport
...
-    private string $accessToken;
-    private ?string $chatChannel;
-
-    public function __construct(#[\SensitiveParameter] string $accessToken, string $chatChannel = null, HttpClientInterface $client = null, EventDispatcherInterface $dispatcher = null)
+    public function __construct(
+        #[\SensitiveParameter]
+        private string $accessToken,
+        private ?string $chatChannel = null,
+        HttpClientInterface $client = null,
+        EventDispatcherInterface $dispatcher = null)
```
nicolas-grekas: for all of them:
```suggestion
        EventDispatcherInterface $dispatcher = null,
    )
```

OskarStark:
fabbot false positive

wouterj (approved):

derrabus (approved):
Rebase needed.

OskarStark:
Rebase done ✅

src/Symfony/Component/Notifier/Bridge/Bandwidth/BandwidthTransport.php
```diff
@@ -31,7 +31,8 @@ final class BandwidthTransport extends AbstractTransport
 
     public function __construct(
         private readonly string $username,
-        #[\SensitiveParameter] private readonly string $password,
+        #[\SensitiveParameter]
+        private readonly string $password,
```
nicolas-grekas: any opinion on this change @symfony/mergers?
which one is the most readable to you?
on my side, I don't mind long lines.

Nyholm: Im happy with the two lines. I think it is more readable. Especially if you have very long attributes.

wouterj: I like the one line more

chalasr: +1 for long one-liners. 2 lines break my brain when reading the property list

Nyholm: Oh.. I dont want to be responsible for breaking Robin's brain.

OskarStark: So shall I revert all bridges to one line?

In all docs (cc @javiereguiluz), and especially if we use attributes with parameters, we tend to use two lines for readability, because
```php
#[MapRequestPaylod(foo: bar, baz: baz)] private readonly Class $foo
```

is quite noisy.

Please advice and I will take the wanted direction across all bridges

nicolas-grekas: I really depends on the length of the attribute part.
When a line-split feels needed, I think we should require adding blank lines before and after the corresponding argument declaration.
In this case, the oneliner is better to me, so yes, please update all lines found in this PR and other similar PRs, if any :pray:

OskarStark: Will do

wouterj: In the docs, we follow Symfony coding standards except from the line length. This is in part also because code blocks in the documentation are at most 726px wide, we don't want to overwhelm readers with too much characters and code examples are often very short.

In Symfony code, we often favor longer lines. This is in part because editors are often fullscreen on a wide screen and having more "real" lines vertically allows you to see code more in context (at least, that's how I experience it). This is also why we e.g. always one-line exceptions.

OskarStark: Done

src/Symfony/Component/Notifier/Bridge/AllMySms/AllMySmsTransport.php
```diff
@@ -28,16 +28,13 @@ final class AllMySmsTransport extends AbstractTransport
...
-        $this->login = $login;
-        $this->apiKey = $apiKey;
-        $this->from = $from;
-
+    public function __construct(
+        private string $login,
+        #[\SensitiveParameter] private string $apiKey,
+        private ?string $from = null,
+        HttpClientInterface $client = null,
+        EventDispatcherInterface $dispatcher = null
```
nicolas-grekas: please check missing trailing commas in the PR

nicolas-grekas (commented):
rebase needed :)

src/Symfony/Component/Notifier/Bridge/AllMySms/AllMySmsTransport.php
OskarStark: Isn't there a CS-Fixer rule to ensure this? 🤔 

cc @keradus

OskarStark: Done

nicolas-grekas:
Thank you @OskarStark.

src/Symfony/Component/Notifier/Bridge/AllMySms/AllMySmsTransport.php
keradus: not sure if you were aware what you asked for... but here we go #53352

PR #53232, opened 2023-12-27
[Notifier] [Smsc] Require login and password

OskarStark:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | ---
| License       | MIT

Checking the tests, and the DSN, there is no need to make it nullable

From a technical POV this is a BC break, which could not really happen in real life 🤷‍♂️


fabpot (approved):

fabpot:
Thank you @OskarStark.

PR #53233, opened 2023-12-27
DX: re-apply CS

keradus:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Issues        | Fix CS
| License       | MIT


src/Symfony/Component/Cache/Adapter/PhpFilesAdapter.php
```diff
@@ -39,7 +39,7 @@ class PhpFilesAdapter extends AbstractAdapter implements PruneableInterface
 
     /**
      * @param $appendOnly Set to `true` to gain extra performance when the items stored in this pool never expire.
-     *                    Doing so is encouraged because it fits perfectly OPcache's memory model.
+     *                     Doing so is encouraged because it fits perfectly OPcache's memory model.
```
keradus, line 42:
```suggestion
     *                    Doing so is encouraged because it fits perfectly OPcache's memory model.
```

src/Symfony/Component/Console/Command/Command.php
```diff
@@ -420,9 +420,9 @@ public function addArgument(string $name, int $mode = null, string $description
     /**
      * Adds an option.
```
keradus: overall, we got curious if phpdoc_align make sense on Fixer side, especially because of such cases like here. WDYT?
The version proposed by PR is one in-line with current CS rules.

nicolas-grekas: I agree it might be better to not align those

keradus: I would love to have conclusion that is "always align" or "never align".

If we say always, let's pay this little price of having this single case out of 100k phpdocs ugly. Maintaining exception and not letting tool to do it's job for single case is painful and preventing auto-apply of tool results.

src/Symfony/Component/ErrorHandler/DebugClassLoader.php
```diff
@@ -184,7 +184,7 @@ public function getClassLoader(): callable
     public static function enable(): void
     {
         // Ensures we don't hit https://bugs.php.net/42098
-        class_exists(\Symfony\Component\ErrorHandler\ErrorHandler::class);
+        class_exists(ErrorHandler::class);
         class_exists(\Psr\Log\LogLevel::class);
```
OskarStark, line 188: Should also be imported, right?

keradus: In this PR, I applied the [fully_qualified_strict_types](https://cs.symfony.com/doc/rules/import/fully_qualified_strict_types.html) rule, which (with current config) simplifies usage of symbols.

If you want to import symbols, we can enable `import_symbols`.
Please decide if you would like to go this path as separated PR: https://github.com/symfony/symfony/pull/53244

src/Symfony/Component/Console/Command/Command.php
nicolas-grekas: I don't think the current changes on this are improving the source. Consistency is nice, but this goes against readability here to me.
I understand you'd like a rule, so maybe we could just decide which length is too much to align? or say that complex types shouldn't trigger alignment?
we can revert these from the PR and discuss that separately if you don't want to other changes to wait for this to be resolved

keradus: for 6k files in this repo, we have the rule not applied for __TWO__ cases only. instead of making some special handling of this and implementing rule for exception, I would ask to get this applied. Let's be pragmatic here and avoid increasing rule complexity for such rare case.

src/Symfony/Component/Finder/Tests/Iterator/IteratorTestCase.php
```diff
@@ -41,10 +41,10 @@ protected function assertOrderedIterator($expected, \Traversable $iterator)
      *  Same as assertOrderedIterator, but checks the order of groups of
      *  array elements.
```
keradus:
```suggestion
     * Same as assertOrderedIterator, but checks the order of groups of
     * array elements.
```

src/Symfony/Component/Console/Command/Command.php
nicolas-grekas: we accept exceptions so it's fine have two cases that don't follow the rule
please revert the related changes in this PR and :rocket:

keradus: It makes it tricky to apply the CS (and on long run, have CS always green) - because manually reverting the exceptions each time :(

reverting as requested

keradus:
(rebased due to fabbot.io request)

nicolas-grekas:
Thank you @keradus.

src/Symfony/Component/Security/Http/AccessToken/Oidc/OidcTrait.php
```diff
@@ -24,7 +24,7 @@ trait OidcTrait
 {
     private function createUser(array $claims): OidcUser
     {
-        if (!\function_exists(\Symfony\Component\String\u::class)) {
+        if (!\function_exists(u::class)) {
```
xabbuh, line 27: we need the namespace to check for the existence of the function (fixed in 862c6fbf803bd34d653699b15baadcb216708c3b)

keradus: good catch, let me convert it into bug report:
https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/7652

keradus: I find this very curious inconsistency on PHP engine side.

https://3v4l.org/qtBRn

nicolas-grekas: deserves a bug report to php-src IMHO

keradus: I see reason why PHP allows for that. For `::class`, one can do whatever, like `fooisufdhsilufgvhlaiuwrgfliudfzbhsdfbsdfb::class`, and it's not evaluated if that foo..sth exists at all, therefore also not if it's a class or any other symbol like function.
It would need the compiler to already have execution context, or at least import all files... to analyse. IMHO doubtful to get in, but if you think otherwise and have better expertise on it, please consider to drive this report.

PR #53234, opened 2023-12-27
[Notifier] Use CPP

OskarStark:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | --
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Notifier/Channel/BrowserChannel.php
+++ b/src/Symfony/Component/Notifier/Channel/BrowserChannel.php
@@ -22,14 +22,10 @@ use Symfony\Component\Notifier\Recipient\RecipientInterface;
  */
 final class BrowserChannel implements ChannelInterface
 {
-    private RequestStack $stack;
-
-    private FlashMessageImportanceMapperInterface $mapper;
-
-    public function __construct(RequestStack $stack, FlashMessageImportanceMapperInterface $mapper = new DefaultFlashMessageImportanceMapper())
-    {
-        $this->stack = $stack;
-        $this->mapper = $mapper;
+    public function __construct(
+        private RequestStack $stack,
+        private FlashMessageImportanceMapperInterface $mapper = new DefaultFlashMessageImportanceMapper(),
+    ) {
     }
 
     public function notify(Notification $notification, RecipientInterface $recipient, string $transportName = null): void
--- a/src/Symfony/Component/Notifier/Channel/ChannelPolicy.php
+++ b/src/Symfony/Component/Notifier/Channel/ChannelPolicy.php
@@ -18,11 +18,9 @@ use Symfony\Component\Notifier\Exception\InvalidArgumentException;
  */
 final class ChannelPolicy implements ChannelPolicyInterface
 {
-    private array $policy;
-
-    public function __construct(array $policy)
-    {
-        $this->policy = $policy;
+    public function __construct(
+        private array $policy,
+    ) {
     }
 
     public function getChannels(string $importance): array
--- a/src/Symfony/Component/Notifier/Event/FailedMessageEvent.php
+++ b/src/Symfony/Component/Notifier/Event/FailedMessageEvent.php
@@ -19,13 +19,10 @@ use Symfony\Contracts\EventDispatcher\Event;
  */
 final class FailedMessageEvent extends Event
 {
-    private MessageInterface $message;
-    private \Throwable $error;
-
-    public function __construct(MessageInterface $message, \Throwable $error)
-    {
-        $this->message = $message;
-        $this->error = $error;
+    public function __construct(
+        private MessageInterface $message,
+        private \Throwable $error,
+    ) {
     }
 
     public function getMessage(): MessageInterface
--- a/src/Symfony/Component/Notifier/EventListener/SendFailedMessageToNotifierListener.php
+++ b/src/Symfony/Component/Notifier/EventListener/SendFailedMessageToNotifierListener.php
@@ -24,11 +24,9 @@ use Symfony\Component\Notifier\Notifier;
  */
 class SendFailedMessageToNotifierListener implements EventSubscriberInterface
 {
-    private Notifier $notifier;
-
-    public function __construct(Notifier $notifier)
-    {
-        $this->notifier = $notifier;
+    public function __construct(
+        private Notifier $notifier,
+    ) {
     }
 
     public function onMessageFailed(WorkerMessageFailedEvent $event): void
--- a/src/Symfony/Component/Notifier/Exception/IncompleteDsnException.php
+++ b/src/Symfony/Component/Notifier/Exception/IncompleteDsnException.php
@@ -16,11 +16,11 @@ namespace Symfony\Component\Notifier\Exception;
  */
 class IncompleteDsnException extends InvalidArgumentException
 {
-    private ?string $dsn;
-
-    public function __construct(string $message, string $dsn = null, \Throwable $previous = null)
-    {
-        $this->dsn = $dsn;
+    public function __construct(
+        string $message,
+        private ?string $dsn = null,
+        \Throwable $previous = null,
+    ) {
         if ($dsn) {
             $message = sprintf('Invalid "%s" notifier DSN: %s', $dsn, $message);
         }
--- a/src/Symfony/Component/Notifier/Exception/TransportException.php
+++ b/src/Symfony/Component/Notifier/Exception/TransportException.php
@@ -18,12 +18,14 @@ use Symfony\Contracts\HttpClient\ResponseInterface;
  */
 class TransportException extends RuntimeException implements TransportExceptionInterface
 {
-    private ResponseInterface $response;
     private string $debug = '';
 
-    public function __construct(string $message, ResponseInterface $response, int $code = 0, \Throwable $previous = null)
-    {
-        $this->response = $response;
+    public function __construct(
+        string $message,
+        private ResponseInterface $response,
+        int $code = 0,
+        \Throwable $previous = null,
+    ) {
         $this->debug .= $response->getInfo('debug') ?? '';
 
         parent::__construct($message, $code, $previous);
--- a/src/Symfony/Component/Notifier/Message/ChatMessage.php
+++ b/src/Symfony/Component/Notifier/Message/ChatMessage.php
@@ -19,14 +19,12 @@ use Symfony\Component\Notifier\Notification\Notification;
 class ChatMessage implements MessageInterface, FromNotificationInterface
 {
     private ?string $transport = null;
-    private string $subject;
-    private ?MessageOptionsInterface $options;
     private ?Notification $notification = null;
 
-    public function __construct(string $subject, MessageOptionsInterface $options = null)
-    {
-        $this->subject = $subject;
-        $this->options = $options;
+    public function __construct(
+        private string $subject,
+        private ?MessageOptionsInterface $options = null,
+    ) {
     }
 
     public static function fromNotification(Notification $notification): self
--- a/src/Symfony/Component/Notifier/Message/EmailMessage.php
+++ b/src/Symfony/Component/Notifier/Message/EmailMessage.php
@@ -25,14 +25,12 @@ use Symfony\Component\Notifier\Recipient\EmailRecipientInterface;
  */
 class EmailMessage implements MessageInterface, FromNotificationInterface
 {
-    private RawMessage $message;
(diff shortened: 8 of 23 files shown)


src/Symfony/Component/Notifier/Channel/BrowserChannel.php
```diff
@@ -22,14 +22,10 @@
...
-
-    private FlashMessageImportanceMapperInterface $mapper;
-
-    public function __construct(RequestStack $stack, FlashMessageImportanceMapperInterface $mapper = new DefaultFlashMessageImportanceMapper())
-    {
-        $this->stack = $stack;
-        $this->mapper = $mapper;
+    public function __construct(
+        private RequestStack $stack,
+        private FlashMessageImportanceMapperInterface $mapper = new DefaultFlashMessageImportanceMapper()
```
OskarStark:
```suggestion
        private FlashMessageImportanceMapperInterface $mapper = new DefaultFlashMessageImportanceMapper(),
```

src/Symfony/Component/Notifier/Event/FailedMessageEvent.php
```diff
@@ -19,13 +19,10 @@
...
-    private MessageInterface $message;
-    private \Throwable $error;
-
-    public function __construct(MessageInterface $message, \Throwable $error)
-    {
-        $this->message = $message;
-        $this->error = $error;
+    public function __construct(
+        private MessageInterface $message,
+        private \Throwable $error
```
OskarStark:
```suggestion
        private \Throwable $error,
```

src/Symfony/Component/Notifier/EventListener/SendFailedMessageToNotifierListener.php
```diff
@@ -24,11 +24,9 @@
...
 class SendFailedMessageToNotifierListener implements EventSubscriberInterface
 {
-    private Notifier $notifier;
-
-    public function __construct(Notifier $notifier)
-    {
-        $this->notifier = $notifier;
+    public function __construct(
+        private Notifier $notifier
+    ) {
```
OskarStark:
```suggestion
    public function __construct(private Notifier $notifier)
    {
```

src/Symfony/Component/Notifier/Exception/IncompleteDsnException.php
```diff
@@ -16,11 +16,11 @@
  */
 class IncompleteDsnException extends InvalidArgumentException
 {
-    private ?string $dsn;
-
-    public function __construct(string $message, string $dsn = null, \Throwable $previous = null)
+    public function __construct(
+        string $message,
+        private ?string $dsn = null,
+        \Throwable $previous = null)
```
OskarStark:
```suggestion
        \Throwable $previous = null,
        )
```

src/Symfony/Component/Notifier/Exception/TransportException.php
```diff
@@ -18,12 +18,14 @@
...
     private string $debug = '';
 
-    public function __construct(string $message, ResponseInterface $response, int $code = 0, \Throwable $previous = null)
-    {
-        $this->response = $response;
+    public function __construct(
+        string $message,
+        private ResponseInterface $response,
+        int $code = 0,
+        \Throwable $previous = null
```
OskarStark:
```suggestion
        \Throwable $previous = null,
```

src/Symfony/Component/Notifier/Message/ChatMessage.php
```diff
@@ -19,14 +19,12 @@
...
-    private ?MessageOptionsInterface $options;
     private ?Notification $notification = null;
 
-    public function __construct(string $subject, MessageOptionsInterface $options = null)
-    {
-        $this->subject = $subject;
-        $this->options = $options;
+    public function __construct(
+        private string $subject,
+        private ?MessageOptionsInterface $options = null
```
OskarStark:
```suggestion
        private ?MessageOptionsInterface $options = null,
```

src/Symfony/Component/Notifier/Message/EmailMessage.php
```diff
@@ -25,14 +25,12 @@
...
-    private ?Envelope $envelope;
     private ?Notification $notification = null;
 
-    public function __construct(RawMessage $message, Envelope $envelope = null)
-    {
-        $this->message = $message;
-        $this->envelope = $envelope;
+    public function __construct(
+        private RawMessage $message,
+        private ?Envelope $envelope = null
```
OskarStark:
```suggestion
        private ?Envelope $envelope = null,
```

src/Symfony/Component/Notifier/Message/PushMessage.php
```diff
@@ -19,16 +19,13 @@
...
 
-    public function __construct(string $subject, string $content, MessageOptionsInterface $options = null)
-    {
-        $this->subject = $subject;
-        $this->content = $content;
-        $this->options = $options;
+    public function __construct(
+        private string $subject,
+        private string $content,
+        private ?MessageOptionsInterface $options = null
```
OskarStark:
```suggestion
        private ?MessageOptionsInterface $options = null,
```

src/Symfony/Component/Notifier/Message/SentMessage.php
```diff
@@ -16,14 +16,12 @@
...
-    private string $transport;
     private ?string $messageId = null;
 
-    public function __construct(MessageInterface $original, string $transport)
-    {
-        $this->original = $original;
-        $this->transport = $transport;
+    public function __construct(
+        private MessageInterface $original,
+        private string $transport
```
OskarStark:
```suggestion
        private string $transport,
```

src/Symfony/Component/Notifier/Message/SmsMessage.php
```diff
@@ -21,22 +21,17 @@
...
-    private ?MessageOptionsInterface $options;
     private ?Notification $notification = null;
 
-    public function __construct(string $phone, string $subject, string $from = '', MessageOptionsInterface $options = null)
-    {
+    public function __construct(
+        private string $phone,
+        private string $subject,
+        private string $from = '',
+        private ?MessageOptionsInterface $options = null
```
OskarStark:
```suggestion
        private ?MessageOptionsInterface $options = null,
```

src/Symfony/Component/Notifier/Messenger/MessageHandler.php
```diff
@@ -20,11 +20,9 @@
...
 final class MessageHandler
 {
-    private TransportInterface $transport;
-
-    public function __construct(TransportInterface $transport)
-    {
-        $this->transport = $transport;
+    public function __construct(
+        private TransportInterface $transport
+    ) {
```
OskarStark:
```suggestion
    public function __construct(private TransportInterface $transport)
    {
```

derrabus (approved):

src/Symfony/Component/Notifier/Notification/Notification.php
```diff
@@ -36,21 +36,19 @@ class Notification
     public const IMPORTANCE_MEDIUM = 'medium';
     public const IMPORTANCE_LOW = 'low';
 
-    private array $channels = [];
-    private string $subject = '';
```
wouterj: Must be reverted to non-CPP. These properties lose their default values with CPP.

wouterj (changes requested):

src/Symfony/Component/Notifier/Notification/Notification.php
```diff
@@ -36,21 +36,19 @@ class Notification
...
-    {
-        $this->subject = $subject;
-        $this->channels = $channels;
+    public function __construct(
+        private string $subject = '',
+        /**
+         * @var list<string>
+         */
+        private array $channels = [],
+    ) {
```
derrabus:
```suggestion
    /**
     * @param list<string> $channels
     */
    public function __construct(
        private string $subject = '',
        private array $channels = [],
    ) {
```

src/Symfony/Component/Notifier/Notification/Notification.php
derrabus: This might only be a problem when either overriding this class without calling the parent constructor or creating an instance of this class without calling the constructor (e.g. through reflection). Are those scenarios we want to support?

wouterj: We don't provide BC for reflecting non-public API as far as I know. However, the constructor overriding without parent constructor is something we should protect against when the class is not final.

I've mentioned this a previous time in a "move to CPP" PR and Nicolas confirmed that we already had some bug reports about this in other classes in Symfony. So I think it's better to stay on the save side and reject CPP for properties with default values on non-final classes (and we can't make DTO's final as that limits extensibility to zero).

See https://github.com/symfony/symfony/pull/45721#discussion_r1265977774

derrabus: Not calling this constructor looks like a mistake to me, but you're right, it might've happened and we should not break code that does so.

Still, I'd like to see a path forward for such classes. Maybe that change would be acceptable for Symfony 8, although I don't really see how we could build a deprecation layer for that.

src/Symfony/Component/Notifier/Notification/Notification.php
OskarStark: Same discussion in https://github.com/symfony/symfony/pull/53241#discussion_r1437880895

src/Symfony/Component/Notifier/Notification/Notification.php
OskarStark: So in short, revert all CPP on non final classes, having a default value?

wouterj: 100% correct :)

And in both of you PRs, this is the only one I spotted (the other PR has one occurence of properties with default values, but it's on a final class)

src/Symfony/Component/Notifier/Message/SmsMessage.php
```diff
@@ -21,22 +21,17 @@
...
-    private string $from;
-    private ?MessageOptionsInterface $options;
     private ?Notification $notification = null;
 
-    public function __construct(string $phone, string $subject, string $from = '', MessageOptionsInterface $options = null)
-    {
+    public function __construct(
+        private string $phone,
+        private string $subject,
+        private string $from = '',
```
OskarStark, line 29: what about this one @wouterj , this should reverted AFAIU

but the following `MessageOptionsInterface` -> `null`, not?

wouterj: This one is correct. For both parameters in this method, the default value was on the parameter, not the property. This works exactly the same with and without CPP (we're only losing the property default value).

OskarStark: Ah got it, thanks

src/Symfony/Component/Notifier/Notification/Notification.php
OskarStark:
```diff
- So in short, revert all CPP on non final classes, having a default value
+ So in short, revert all CPP on non final classes, having a default value on the property
```

chalasr: > Not calling this constructor looks like a mistake to me, but you're right, it might've happened and we should not break code that does so.

I'm pretty sure we already rejected regression report(s) due to missing parent constructor calls in the past (mostly argued by @stof IIRC) so I'm mitigated, but I tend to agree we should not do it consciously for the sake of adding CPP.

OskarStark: Done

OskarStark:
Rebased, comments addressed, fabbot unrelated

src/Symfony/Component/Notifier/Channel/ChannelPolicy.php
```diff
@@ -18,11 +18,8 @@
  */
 final class ChannelPolicy implements ChannelPolicyInterface
 {
-    private array $policy;
-
-    public function __construct(array $policy)
+    public function __construct(private array $policy)
```
nicolas-grekas: as noted in another similar PR, let's put CPP on their own lines
please review all similar PRs please

OskarStark: Done

nicolas-grekas:
Thank you @OskarStark.

PR #53235, opened 2023-12-27
[Validator] Fix `Charset` validator test data

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

`ώ` is not always detected as an UTF-16 valid string by mbstring. This cause the test to fail [in some condition](https://ci.appveyor.com/project/fabpot/symfony/builds/48837004).

The purpose of this test is to try different encodings accepted by mbstring, thus I suggest to change the last data set to make it more stable. The UTF-16 thing seems related to some mbstring platform-dependant implementation so let's "move away" from this.


xabbuh:
Tests are still failing on AppVeyor. Shouldn't they become green with these changes?

alexandre-daubois:
Seems I chose the exact wrong encoding that always return false on some implementation :facepalm: I updated to use a more "classical" one. All green :+1:

xabbuh (approved):

javiereguiluz (approved):

xabbuh:
Thank you Alexandre.

PR #53241, opened 2023-12-27
[Mailer] [Bridges] Use CPP and add more `#[\SensitiveParameter]` attributes

OskarStark:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | --
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Mailer/Bridge/Azure/Transport/AzureApiTransport.php
+++ b/src/Symfony/Component/Mailer/Bridge/Azure/Transport/AzureApiTransport.php
@@ -29,36 +29,24 @@ final class AzureApiTransport extends AbstractApiTransport
     private const HOST = '%s.communication.azure.com';
 
     /**
-     * User Access Key from Azure Communication Service (Primary or Secondary key).
+     * @param string $key             User Access Key from Azure Communication Service (Primary or Secondary key)
+     * @param string $resourceName    The endpoint API URL to which to POST emails to Azure https://{acsResourceName}.communication.azure.com/
+     * @param bool   $disableTracking Indicates whether user engagement tracking should be disabled
+     * @param string $apiVersion      The version of API to invoke
      */
-    private string $key;
-
-    /**
-     * The endpoint API URL to which to POST emails to Azure
-     * https://{acsResourceName}.communication.azure.com/.
-     */
-    private string $resourceName;
-
-    /**
-     * The version of API to invoke.
-     */
-    private string $apiVersion;
-
-    /**
-     * Indicates whether user engagement tracking should be disabled.
-     */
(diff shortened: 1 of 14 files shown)


src/Symfony/Component/Mailer/Bridge/Azure/Transport/AzureApiTransport.php
```diff
@@ -28,37 +28,32 @@ final class AzureApiTransport extends AbstractApiTransport
...
+         */
+        private bool $disableTracking = false,
+        /**
+         * The version of API to invoke.
+         */
+        private string $apiVersion = '2023-03-31',
+        HttpClientInterface $client = null,
+        EventDispatcherInterface $dispatcher = null,
+        LoggerInterface $logger = null,
+    ) {
```
derrabus:
```suggestion
    /**
     * @param string $key              User Access Key from Azure Communication Service (Primary or Secondary key).
     * @param string $resourceName    The endpoint API URL to which to POST emails to Azure (https://{acsResourceName}.communication.azure.com/).
     * @param bool   $disableTracking Indicates whether user engagement tracking should be disabled.
     * @param string $apiVersion      The version of API to invoke.
     */
    public function __construct(
        private string $key,
        private string $resourceName,
        private bool $disableTracking = false,
        private string $apiVersion = '2023-03-31',
        HttpClientInterface $client = null,
        EventDispatcherInterface $dispatcher = null,
        LoggerInterface $logger = null,
    ) {
```

OskarStark: I think, as these are properties now, I cannot use `@param` annotation. Because of this, we don't use them in 
* https://github.com/symfony/symfony/pull/51971/files

cc @GromNaN @alexandre-daubois

See https://github.com/symfony/symfony/pull/51971#discussion_r1352771614 for more context

nicolas-grekas: @derrabus' version look corrects to me

alexandre-daubois: Indeed, Alexander's suggestion works (in PhpStorm at least). Well, I think I have a PR to update 😄

nicolas-grekas: Don't you want to update as suggested @OskarStark? Why?

OskarStark:
Rebased and comments addressed ✅

nicolas-grekas:
Thank you @OskarStark.

PR #53244, opened 2023-12-27
CS: fully_qualified_strict_types - import_symbols

keradus:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Issues        | Fix CS 
| License       | MIT

suggested in https://github.com/symfony/symfony/pull/53233/files#r1437244439

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Redis/Transport/Connection.php
+++ b/src/Symfony/Component/Messenger/Bridge/Redis/Transport/Connection.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Component\Messenger\Bridge\Redis\Transport;
 
+use Relay\Exception as RelayException;
 use Relay\Relay;
 use Relay\Sentinel;
 use Symfony\Component\Messenger\Exception\InvalidArgumentException;
@@ -132,7 +133,7 @@ class Connection
                             if ($address = $sentinel->getMasterAddrByName($sentinelMaster)) {
                                 [$host, $port] = $address;
                             }
-                        } catch (\RedisException|\Relay\Exception $redisException) {
+                        } catch (\RedisException|RelayException $redisException) {
                         }
                     } while (++$hostIndex < \count($hosts) && !$address);
 
@@ -337,7 +338,7 @@ class Connection
             // This could soon be optimized with https://github.com/antirez/redis/issues/5212 or
             // https://github.com/antirez/redis/issues/6256
             $pendingMessages = $this->getRedis()->xpending($this->stream, $this->group, '-', '+', 1);
-        } catch (\RedisException|\Relay\Exception $e) {
+        } catch (\RedisException|RelayException $e) {
             throw new TransportException($e->getMessage(), 0, $e);
         }
 
@@ -366,7 +367,7 @@ class Connection
                 );
 
                 $this->couldHavePendingMessages = true;
-            } catch (\RedisException|\Relay\Exception $e) {
+            } catch (\RedisException|RelayException $e) {
                 throw new TransportException($e->getMessage(), 0, $e);
             }
         }
@@ -425,7 +426,7 @@ class Connection
                 1,
                 1
             );
-        } catch (\RedisException|\Relay\Exception $e) {
+        } catch (\RedisException|RelayException $e) {
             throw new TransportException($e->getMessage(), 0, $e);
         }
 
@@ -463,7 +464,7 @@ class Connection
             if ($this->deleteAfterAck) {
                 $acknowledged = $redis->xdel($this->stream, [$id]);
             }
-        } catch (\RedisException|\Relay\Exception $e) {
+        } catch (\RedisException|RelayException $e) {
             throw new TransportException($e->getMessage(), 0, $e);
         }
 
@@ -484,7 +485,7 @@ class Connection
             if ($this->deleteAfterReject) {
                 $deleted = $redis->xdel($this->stream, [$id]) && $deleted;
             }
-        } catch (\RedisException|\Relay\Exception $e) {
+        } catch (\RedisException|RelayException $e) {
             throw new TransportException($e->getMessage(), 0, $e);
         }
 
@@ -547,7 +548,7 @@ class Connection
 
                 $id = $added;
             }
-        } catch (\RedisException|\Relay\Exception $e) {
+        } catch (\RedisException|RelayException $e) {
             if ($error = $redis->getLastError() ?: null) {
                 $redis->clearLastError();
             }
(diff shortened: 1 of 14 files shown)


xabbuh (changes requested):
failures look related to me

keradus:
this PR serves a purpose to demonstrate "how the codebase would look alike if we would incorporate another Fixer rule", as one of members suggested to get it enabled.

Before investing any time into failure of this PR, I would like to get confirmation we want to go this path and enable the rule. Diff is massive and I do not want to invest time into investigation, if final conclusion would be that "we do not want this rule to be enabled".

What's your take on the rule itself?

xabbuh:
importing the classes looks indeed like what we want to have in the codebase

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -12,6 +12,7 @@
 namespace Symfony\Bundle\FrameworkBundle\DependencyInjection;
 
 use Composer\InstalledVersions;
+use GuzzleHttp\UriTemplate\UriTemplate;
```
xabbuh:
```suggestion
use GuzzleHttp\UriTemplate\UriTemplate as GuzzleUriTemplate;
```

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -23,12 +24,14 @@
 use Psr\Container\ContainerInterface as PsrContainerInterface;
 use Psr\Http\Client\ClientInterface;
 use Psr\Log\LoggerAwareInterface;
+use Rize\UriTemplate;
```
xabbuh:
```suggestion
use Rize\UriTemplate as RizeUriTemplate;
```

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2432,9 +2438,9 @@ private function registerHttpClientConfiguration(array $config, ContainerBuilder
             $this->registerRetryableHttpClient($retryOptions, 'http_client', $container);
         }
 
-        if (ContainerBuilder::willBeAvailable('guzzlehttp/uri-template', \GuzzleHttp\UriTemplate\UriTemplate::class, [])) {
+        if (ContainerBuilder::willBeAvailable('guzzlehttp/uri-template', UriTemplate::class, [])) {
```
xabbuh:
```suggestion
        if (ContainerBuilder::willBeAvailable('guzzlehttp/uri-template', GuzzleUriTemplate::class, [])) {
```

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2432,9 +2438,9 @@ private function registerHttpClientConfiguration(array $config, ContainerBuilder
...
             $container->setAlias('http_client.uri_template_expander', 'http_client.uri_template_expander.guzzle');
-        } elseif (ContainerBuilder::willBeAvailable('rize/uri-template', \Rize\UriTemplate::class, [])) {
+        } elseif (ContainerBuilder::willBeAvailable('rize/uri-template', UriTemplate::class, [])) {
```
xabbuh:
```suggestion
        } elseif (ContainerBuilder::willBeAvailable('rize/uri-template', RizeUriTemplate::class, [])) {
```

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
derrabus: I would consider this to be a bug in the fixer: The fixer should detect that a class with the same name has already been imported from a different namespace and either do nothing or generate an alias for the import. Currently, the fixer would generate obviously broken code.

keradus:
I take the eagerness to use the rule.

I will covert PR to draft as it is not in merge-ready state.

Wirone:
Related to PHP-CS-Fixer/PHP-CS-Fixer#7629 - this PR probably could prepare a Symfony codebase for that.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
Wirone: @derrabus yes, I agree, obviously it's a bug in the importing logic. I overlooked it in my PR 🙁.

Wirone: Actually there was a test case for names collision, but it involved existing import + new symbol, while the problem here was related to names collision between newly found symbols, that I did not take into consideration. I've prepared fix for that in PHP-CS-Fixer/PHP-CS-Fixer#7635 🙂.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
keradus: (outdated)

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
keradus: (outdated)

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
keradus: (outdated)

src/Symfony/Component/Messenger/Bridge/Redis/Transport/Connection.php
```diff
@@ -132,7 +133,7 @@ public function __construct(array $options, \Redis|Relay|\RedisCluster|null $red
                             if ($address = $sentinel->getMasterAddrByName($sentinelMaster)) {
                                 [$host, $port] = $address;
                             }
-                        } catch (\RedisException|\Relay\Exception $redisException) {
+                        } catch (\RedisException|RelayException $redisException) {
```
keradus, line 136: looks like Psalm cannot handle this, should I bother by that and stop the PR or continue with it?
cc @xabbuh 

also, reported here: https://github.com/vimeo/psalm/issues/10840

nicolas-grekas:
Please let me close this one.
We're fine with doing this case by case.
Usually importing is the best way, but in some cases, inlining the FQCN provides more clarity to the codebase.

PR #53248, opened 2023-12-27
[Notifier/OneSignal] Fix external player IDs

### LINKED ISSUE #50779: Support for external user ids in OneSignal Notifier
### Description

Currently the OneSignal notifier only works with `player_ids` but it would be very convenient to be able to use the `external_user_ids` as well.
https://documentation.onesignal.com/reference/create-notification

External user ids are a kind of unified ID that's set by the developer and then used to send a notification to all devices by that user at the same time.

### Example

_No response_

KDederichs:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | yes (kinda, if it never worked is it a deprecation?)
| Issues        | Fix #50779
| License       | MIT

After digging into it I found that the original implementation seems to plan for external players but the implementation seems like it never would have worked, since the parameter was called `include_external_user_ids` and not `external_id`.

Looking at their docs, you can also see they recently changed their API parameters, so I took the chance to just go ahead and update to the new parameters.

Since the new one is a bit more complicated to set external IDs I ended up marking the old method as deprecated and adding a new option.

I think this should be better since external ids and subscription ids should be mutually exclusive.

See: https://documentation.onesignal.com/reference/create-notification

I didn't have time to test it yet so it's a draft for now, so please someone feel free to check it out, otherwise I'll try to real world test it in the coming days.

### DIFF
--- a/src/Symfony/Component/Notifier/Bridge/OneSignal/OneSignalTransport.php
+++ b/src/Symfony/Component/Notifier/Bridge/OneSignal/OneSignalTransport.php
@@ -81,7 +81,17 @@ final class OneSignalTransport extends AbstractTransport
 
         $options = $opts ? $opts->toArray() : [];
         $options['app_id'] = $this->appId;
-        $options['include_player_ids'] = [$recipientId];
+        if (isset($options['external_id'])) {
+            $options['include_aliases'] = [
+                'external_id' => [
+                    $options['external_id']
+                ]
+            ];
+            $options['target_channel'] = 'push';
+            unset($options['external_id']);
+        } else {
+            $options['include_subscription_ids'] = [$recipientId]; // include_player_ids has been deprecated
+        }
 
         if (!isset($options['headings'])) {
             $options['headings'] = ['en' => $message->getSubject()];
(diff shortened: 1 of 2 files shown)


norkunas:
OneSignal deprecated the old way, but it's not removed yet? It was implemented before it based on their older api

fabpot:
That should be done in 7.1 (on lower branches, we cannot deprecate anything nor add new features).

KDederichs:
> OneSignal deprecated the old way, but it's not removed yet? It was implemented before it based on their older api

yeah, but even the old way the parameter was called `include_external_user_ids` (acroding to the docs) while the notifier just set `external_id`, or am I missing something obvious here?

OskarStark:
We need to split this PR, 5.4 should fix the used key, while 7.1 should introduce the new isExternal method and deprecate the old method.

norkunas:
> > OneSignal deprecated the old way, but it's not removed yet? It was implemented before it based on their older api
> 
> yeah, but even the old way the parameter was called `include_external_user_ids` (acroding to the docs) while the notifier just set `external_id`, or am I missing something obvious here?

I see now, so external ids were not implemented initially,only old player ids, trully not sure why that method is there

Edit: I found this in v9 docs:

KDederichs:
Those docs are a strange mess... at the top it says this:
Screenshot 2023-12-28 at 18 10 46

So... both work?

Edit: Oh seems like it's something different, seems like the `external_id` one is to prevent double sending?

norkunas:
I think that this option is not for an user id,but instead custom id for deduplication,so it should not be removed 🤔

yeah hard to understans their docs.. would be easier if they would have multiple endpoints for notifications based on needed filters..

KDederichs:
So, since it seems to be a new feature after all I'd assume the right course of action would be to close this and re-open a PR against 7.1 right?

norkunas:
Totally

KDederichs:
Ok, closing this then and preparing one for 7.1

PR #53250, opened 2023-12-28
[Serializer] Improve nested payload validation for `#[MapRequestPayload]` using a new serialization context

mmouih:
| Q             | A
| ------------- | ---
| Branch?       | 8.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

Hello everyone,

When validating a **nested object** using MapRequestPayload, and given payload schema doesn't respect the DTO provided, Symfony serializer throws `PartialDenormalizationException`.

Unfortunately the `MapRequestPayloadResolver` doesn't have enough information about the Dto structure to be able to provide a more detailed error, Instead symfony throws the errors "**This value should be of type unknown.**"

image

This **unknown** type is originated from the `AbstractNormalizer` in the serializer component, which injects this value to **NotNormalizableValueException** class.
I propose to add new Serialization context `AbstractNormalizer::USE_CLASS_AS_DEFAULT_EXPECTED_TYPE`, 
which allows us to retrieve the class as the expected type for Our DTO if needed.

Example when using the context AbstractNormalizer::USE_CLASS_AS_DEFAULT_EXPECTED_TYPE:
image

I honestly do not know the reason behind using unkown instead of the FQCN. I find adding a new context more safe, and certainly **Backward compatible** to support both usages.

Usage example in a controller:

[https://github.com/mmouih/apps-demo/blob/19f9edfa4361567a4b592a32041ead327fb3b898/src/Controller/HomeController.php#L22,L28](url)

I included tests for the new feature:

[https://github.com/mmouih/symfony/blob/05ef92028cb75c556ebbf2857617bdf40b8f1643/src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/RequestPayloadValueResolverTest.php#L288C21-L288C94](url)

Best regards.

### DIFF
--- a/src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/RequestPayloadValueResolverTest.php
+++ b/src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/RequestPayloadValueResolverTest.php
@@ -21,7 +21,6 @@ use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
 use Symfony\Component\HttpKernel\Event\ControllerArgumentsEvent;
 use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
-use Symfony\Component\PropertyAccess\Exception\InvalidTypeException;
 use Symfony\Component\Serializer\Encoder\JsonEncoder;
 use Symfony\Component\Serializer\Encoder\XmlEncoder;
 use Symfony\Component\Serializer\Exception\PartialDenormalizationException;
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/HttpKernel/Tests/Controller/ArgumentResolver/RequestPayloadValueResolverTest.php
```diff
@@ -285,6 +284,50 @@ public function testValidationNotPerformedWhenPartialDenormalizationReturnsViola
...
+        $kernel = $this->createMock(HttpKernelInterface::class);
+
+        // Test using use_class_as_default_expected_type context
+        $argument = new ArgumentMetadata('invalid-nested-payload', Employee::class, false, false, null, false, [
+            MapRequestPayload::class => new MapRequestPayload(serializationContext: ['use_class_as_default_expected_type' => true]),
+        ]);
+        $arguments = $resolver->resolve($request, $argument);
+        $event = new ControllerArgumentsEvent($kernel, function () {}, $arguments, $request, HttpKernelInterface::MAIN_REQUEST);
+
+        try {
```
mtarld, line 337: You can leverage `expectException` and `expectExceptionMessage` instead

src/Symfony/Component/Serializer/Normalizer/AbstractNormalizer.php
```diff
@@ -118,6 +118,11 @@ abstract class AbstractNormalizer implements NormalizerInterface, DenormalizerIn
      */
     public const REQUIRE_ALL_PROPERTIES = 'require_all_properties';
 
+    /**
+     * Use class name as default expected type when throwing NotNormalizableValueException instead of unknown.
+     */
+    public const USE_CLASS_AS_DEFAULT_EXPECTED_TYPE = 'use_class_as_default_expected_type';
```
mtarld, line 124: IMHO, having a more precise exception message can be kind of considered as a fix (see https://github.com/symfony/symfony/pull/53108), so I think we can forget that context attribute

nicolas-grekas:
Fixed by #61746

PR #53264, opened 2023-12-28
[DoctrineBridge] Move `symfony/uid` to `require` section 

barlito:
| Q             | A
| ------------- | ---
| Branch?       | 7.1 
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Issues        | X 
| License       | MIT

I'm wondering why the symfony/uid component is required as a dev dependency as it is used in `Symfony\Bridge\Doctrine\IdGenerator\UuidGenerator` for example.

Maybe this PR can be merge in earlier x.4 branches


derrabus:
The bridge can be used without the UID component. It's an optional dependency.

PR #48203, opened 2022-11-12
[Notifier] Add Pusher Notifier Bridge

bitgandtter:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

Add [Pusher](https://pusher.com/) symfony notifier bridge

PR #53267, opened 2023-12-28
[Notifier] Add Pusher.com bridge

OskarStark:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | --
| License       | MIT

Replaces #48203 because `Allow edits from Maintainer` box is unchecked

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -147,6 +147,7 @@
         "predis/predis": "^1.1|^2.0",
         "psr/http-client": "^1.0",
         "psr/simple-cache": "^1.0|^2.0|^3.0",
+        "pusher/pusher-php-server": "^7.0",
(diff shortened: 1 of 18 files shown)


OskarStark:
I have mixed feelings about this bridge, code looks OK now after some fixes, but injecting the build client feels weird. In other bridges we do not depend on third party libraries.
I didn't try this bridge, and don't have the time to implement the plain API.

If someone want to take over and need this bridge, feels free to use this PR as a starting point.

Closing here

fabpot:
I haven't read the code but we don't want bridges that "just" wrap other clients.

PR #53279, opened 2023-12-29
[Uid] Use CPP

OskarStark:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | --
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Uid/Command/GenerateUlidCommand.php
+++ b/src/Symfony/Component/Uid/Command/GenerateUlidCommand.php
@@ -25,12 +25,9 @@ use Symfony\Component\Uid\Factory\UlidFactory;
 #[AsCommand(name: 'ulid:generate', description: 'Generate a ULID')]
 class GenerateUlidCommand extends Command
 {
-    private UlidFactory $factory;
-
-    public function __construct(UlidFactory $factory = null)
-    {
-        $this->factory = $factory ?? new UlidFactory();
-
+    public function __construct(
+        private UlidFactory $factory = new UlidFactory(),
+    ) {
         parent::__construct();
     }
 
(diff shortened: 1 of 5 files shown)


src/Symfony/Component/Uid/Command/GenerateUlidCommand.php
```diff
@@ -25,12 +25,8 @@
 #[AsCommand(name: 'ulid:generate', description: 'Generate a ULID')]
 class GenerateUlidCommand extends Command
 {
-    private UlidFactory $factory;
-
-    public function __construct(UlidFactory $factory = null)
+    public function __construct(private UlidFactory $factory = new UlidFactory())
```
nicolas-grekas: let's always put CPP properties on their own line, that'll reduce also the diff when adding new arguments in the future

OskarStark: I like it 👍🏻

OskarStark: Done

src/Symfony/Component/Uid/Command/GenerateUlidCommand.php
```diff
@@ -25,12 +25,9 @@
...
 class GenerateUlidCommand extends Command
 {
-    private UlidFactory $factory;
-
-    public function __construct(UlidFactory $factory = null)
-    {
-        $this->factory = $factory ?? new UlidFactory();
-
+    public function __construct(
+        private UlidFactory $factory = new UlidFactory(),
```
fancyweb, line 29:
```suggestion
        private readonly UlidFactory $factory = new UlidFactory(),
```
Can we also add readonly where we can?

OskarStark: I think @nicolas-grekas is 👎 on this

nicolas-grekas: It'd be useless or just visual debt in practice to me yes.

fancyweb (approved):

nicolas-grekas:
Thank you @OskarStark.

PR #53282, opened 2023-12-29
[RateLimiter] Fix RateLimit->getRetryAfter() return value when consuming 0 or last tokens

wouterj:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Replaces #52835 Original description:

> Have got some BC after updating the project to Symfony 6.4 (after that PR https://github.com/symfony/symfony/pull/51676).
> 
> Sometimes I need to get `RateLimit` object without consuming just before consuming a try, in example:
> ```php
> $rateLimit = $limiter->consume(0);
> if ($rateLimit->getRemainingTokens() === 0) {
>    throw new SomeException($rateLimit->getRetryAfter());
> }
> ...
> $limiter->consume(1)
> ...
> ```
> and found that in that case `$rateLimit->getRetryAfter()` always returns `now`.
> 
> So this PR fixes it.

### DIFF
--- a/src/Symfony/Component/RateLimiter/Policy/SlidingWindowLimiter.php
+++ b/src/Symfony/Component/RateLimiter/Policy/SlidingWindowLimiter.php
@@ -65,12 +65,18 @@ final class SlidingWindowLimiter implements LimiterInterface
             $now = microtime(true);
             $hitCount = $window->getHitCount();
             $availableTokens = $this->getAvailableTokens($hitCount);
+            if (0 === $tokens) {
+                $resetDuration = $window->calculateTimeForTokens($this->limit, $window->getHitCount());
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Thank you @wouterj.

nicolas-grekas:
And thank you @ERuban

ERuban:
Huh, have found that this PR didn't fix some cases that was fixed by my ones.

1. if you will try to check `getRetryAfter` right after using the last token(s) - then it will always return `now()` but it should be a time when first attempt will be available

2. When there is no available tokens and 0 is consumed then `getRetryAfter` will show the time when all tokens will be available...

/cc @wouterj

PR #53303, opened 2023-12-30
[Cache] harden cache expiration test

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

In the worst case when the process sleeps for exactly one second the current timestamp would be exactly the same as the expiration time. Increasing the sleep time ensures that the tested item is always expired when executing the test.


nicolas-grekas:
Thank you @xabbuh.

PR #53320, opened 2023-12-30
[Cache][DependencyInjection][Lock][Mailer][Messenger][Notifier][Translation] Url decode username and passwords from `parse_url()` results

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Following https://github.com/symfony/symfony/pull/52758#pullrequestreview-1799091301

I guess this doesn't an entry in the changelog, does it? I replaced `urldecode` by `rawurldecode` and `$parsedDsn` by `$params`, as latest updates to the code used the latter.

### DIFF
--- a/src/Symfony/Component/Cache/Adapter/MemcachedAdapter.php
+++ b/src/Symfony/Component/Cache/Adapter/MemcachedAdapter.php
@@ -114,6 +114,8 @@ class MemcachedAdapter extends AbstractAdapter
                 $params = preg_replace_callback('#^memcached:(//)?(?:([^@]*+)@)?#', function ($m) use (&$username, &$password) {
                     if (!empty($m[2])) {
                         [$username, $password] = explode(':', $m[2], 2) + [1 => null];
+                        $username = rawurldecode($username);
+                        $password = null !== $password ? rawurldecode($password) : null;
                     }
 
                     return 'file:'.($m[1] ?? '');
--- a/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
+++ b/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php
@@ -253,15 +253,15 @@ class EnvVarProcessor implements EnvVarProcessorInterface
         }
 
         if ('url' === $prefix) {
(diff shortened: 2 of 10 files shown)


src/Symfony/Component/Cache/Adapter/MemcachedAdapter.php
```diff
@@ -111,6 +111,9 @@ public static function createConnection(#[\SensitiveParameter] array|string $ser
                 $params = preg_replace_callback('#^memcached:(//)?(?:([^@]*+)@)?#', function ($m) use (&$username, &$password) {
                     if (!empty($m[2])) {
                         [$username, $password] = explode(':', $m[2], 2) + [1 => null];
+
+                        $username = rawurldecode($username);
+                        $password = rawurldecode($password);
```
nicolas-grekas: password can be null so that decoding can fail currently

nicolas-grekas (commented):
Thanks for having a look.
Let's go with a bugfix on this one to me.

OskarStark:
And again a DSN component would help us across the whole codebase 

I know this was discussed quite often in the past. Anything changed your mind @nicolas-grekas ?

cc @Nyholm

nicolas-grekas:
All this makes me think we need a working url parser... (a dsn component is another story.)
(and I don't know where I'd put this parser...)

OskarStark:
To be reusable in a dedicated component ofc, but as this is util and no "domain", I don't know either 🤷‍♂️ but it sounds good 👍

alexandre-daubois:
Maybe this UrlParser could belong to HttpFoundation?

If this is something that could happen, I'll gladly implement this and leverage it in the numerous DSN parsing processes the codebase counts.

alexandre-daubois:
Rebased to 5.4 to make it a bug fix. I'll have a look at what can be done with some UrlParser/DsnParser to have a global logic over every DSN parsing

nicolas-grekas:
Thank you @alexandre-daubois.

PR #53328, opened 2023-12-31
[Messenger] Add jitter parameter to MultiplierRetryStrategy

rmikalkenas:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

I'm seeing an interesting case, when multiple queue consumers causes failures in a downstream systems a.k.a. [thundering herd](https://en.wikipedia.org/wiki/Thundering_herd_problem) effect.
Many jobs fail, they all get enqueued to try again in a static interval and bring down the downstream system yet again. This repeats until the retry limit is exhausted.
Introduced delay randomness prevents thundering herd effect. The randomness can be controlled via `$jitter` parameter.

~~**Open question:** I added `$jitter` parameter with a default value of `0.1` (as in [http client's generic retry](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/HttpClient/Retry/GenericRetryStrategy.php)), but maybe it should be set as `0` to keep code in a BC fashion?~~

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
@@ -1592,6 +1592,7 @@ class Configuration implements ConfigurationInterface
                                             ->integerNode('delay')->defaultValue(1000)->min(0)->info('Time in ms to delay (or the initial value when multiplier is used)')->end()
                                             ->floatNode('multiplier')->defaultValue(2)->min(1)->info('If greater than 1, delay will grow exponentially for each retry: this delay = (delay * (multiple ^ retries))')->end()
                                             ->integerNode('max_delay')->defaultValue(0)->min(0)->info('Max time in ms that a retry should ever be delayed (0 = infinite)')->end()
+                                            ->floatNode('jitter')->defaultValue(0.1)->min(0)->max(1)->info('Randomness to apply to the delay (between 0 and 1)')->end()
                                         ->end()
                                     ->end()
                                     ->scalarNode('rate_limiter')
--- a/src/Symfony/Component/Messenger/CHANGELOG.md
+++ b/src/Symfony/Component/Messenger/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
  * Add option `redis_sentinel` as an alias for `sentinel_master`
  * Add `--all` option to the `messenger:consume` command
  * Make `#[AsMessageHandler]` final
+ * Add parameter `$jitter` to `MultiplierRetryStrategy` in order to randomize delay and prevent the thundering herd effect
 
 7.0
 ---
--- a/src/Symfony/Component/Messenger/Retry/MultiplierRetryStrategy.php
+++ b/src/Symfony/Component/Messenger/Retry/MultiplierRetryStrategy.php
@@ -36,14 +36,16 @@ class MultiplierRetryStrategy implements RetryStrategyInterface
     private int $delayMilliseconds;
     private float $multiplier;
     private int $maxDelayMilliseconds;
+    private float $jitter;
 
     /**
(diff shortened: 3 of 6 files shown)


src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1586,6 +1586,7 @@ function ($a) {
                                             ->integerNode('delay')->defaultValue(1000)->min(0)->info('Time in ms to delay (or the initial value when multiplier is used)')->end()
                                             ->floatNode('multiplier')->defaultValue(2)->min(1)->info('If greater than 1, delay will grow exponentially for each retry: this delay = (delay * (multiple ^ retries))')->end()
                                             ->integerNode('max_delay')->defaultValue(0)->min(0)->info('Max time in ms that a retry should ever be delayed (0 = infinite)')->end()
+                                            ->floatNode('jitter')->defaultValue(0.1)->min(0)->max(1)->info('Randomness in percent (between 0 and 1) to apply to the delay')->end()
```
```suggestion
                                            ->floatNode('jitter')->defaultValue(0.1)->min(0)->max(1)->info('Randomness to apply to the delay (between 0 and 1)')->end()
```

src/Symfony/Component/Messenger/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ CHANGELOG
  * Add option `redis_sentinel` as an alias for `sentinel_master`
  * Add `--all` option to the `messenger:consume` command
  * Make `#[AsMessageHandler]` final
+ * Add parameter `$jitter` to `MultiplierRetryStrategy` in order to randomize delay and prevent the thundering herd effect.
```
nicolas-grekas:
```suggestion
 * Add parameter `$jitter` to `MultiplierRetryStrategy` in order to randomize delay and prevent the thundering herd effect
```

src/Symfony/Component/Messenger/Retry/MultiplierRetryStrategy.php
```diff
@@ -36,14 +36,16 @@ class MultiplierRetryStrategy implements RetryStrategyInterface
...
     private float $multiplier;
     private int $maxDelayMilliseconds;
+    private float $jitter;
 
     /**
      * @param int   $maxRetries           The maximum number of times to retry
      * @param int   $delayMilliseconds    Amount of time to delay (or the initial value when multiplier is used)
      * @param float $multiplier           Multiplier to apply to the delay each time a retry occurs
      * @param int   $maxDelayMilliseconds Maximum delay to allow (0 means no maximum)
+     * @param float $jitter               Probability of randomness int delay (0 = none, 1 = 100% random)
```
nicolas-grekas:
```suggestion
     * @param float $jitter               Randomness to apply to the delay (between 0 and 1)
```

nicolas-grekas (commented):
I'm fine with a default non-null jitter.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
rmikalkenas: updated

src/Symfony/Component/Messenger/Retry/MultiplierRetryStrategy.php
rmikalkenas: updated

src/Symfony/Component/Messenger/CHANGELOG.md
rmikalkenas: updated

nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @rmikalkenas.

PR #53329, opened 2023-12-31
[Cache] Fix Couchbase expiration test

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

We changed for a datetime instance because older versions of Couchbase supported an expiration time in seconds only if it was < 30 days. But this limit was raised to 50 years, so I suggest we go back to basics and only provide an integer.

### DIFF
--- a/src/Symfony/Component/Cache/Adapter/CouchbaseCollectionAdapter.php
+++ b/src/Symfony/Component/Cache/Adapter/CouchbaseCollectionAdapter.php
@@ -183,7 +183,7 @@ class CouchbaseCollectionAdapter extends AbstractAdapter
         }
 
         $upsertOptions = new UpsertOptions();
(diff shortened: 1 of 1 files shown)


xabbuh (approved):
for 7.1

alexandre-daubois:
At first, this fix targeted 6.4 because it made the CI fail in this version. So I guess this should still target 6.3 to have consistency between minor version, what do you think?

nicolas-grekas (approved):
For 6.3, yes, fixing an unreleased change.

xabbuh:
Ah, I missed that the change wasn’t released yet.

xabbuh:
Thank you @alexandre-daubois.

PR #53332, opened 2023-12-31
[HttpClient] add DecoratorTrait for response

cyve:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | 
| License       | MIT

When we decorate the HttpClient, we often need to create a dedicated class to wrap the response (ex `TraceableHttpClient` => `TraceableResponse`)

Following the idea introduced in https://github.com/symfony/symfony/pull/41161, I propose to add a trait to simplify the response decoration.

Example:
```php
class MyCustomHttpClient implements HttpClientInterface
{
    use DecoratorTrait;

    public function request(string $method, string $url, array $options = []): ResponseInterface
    {
        return new MyCustomResponse($this->client->request($method, $url, $options));
    }
}

class MyCustomResponse implements ResponseInterface
{
    use Response\DecoratorTrait;
    // overwrite only necessary methods
}
```

### DIFF
--- a/src/Symfony/Component/HttpClient/CHANGELOG.md
+++ b/src/Symfony/Component/HttpClient/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
  * Add `HttpOptions::setHeader()` to add or replace a single header
  * Allow mocking `start_time` info in `MockResponse`
  * Add `MockResponse::fromFile()` and `JsonMockResponse::fromFile()` methods to help using fixtures files
+ * Add `DecoratorTrait` for responses
 
 7.0
 ---
--- a/src/Symfony/Component/HttpClient/Response/DecoratorTrait.php
+++ b/src/Symfony/Component/HttpClient/Response/DecoratorTrait.php
@@ -1,0 +1,98 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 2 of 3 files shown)


nicolas-grekas:
Hello, thanks for proposing. Can you explain your use case?
The AsyncDecoratorTrait + AsyncResponse combo is a way better alternative to me since it also works with `$client->stream()`, while simple response decorators don't.

cyve:
Hi @nicolas-grekas, thanks for your feedback. I took the opportunity to take a deeper look at the AsyncResponse concept. It looks really nice. I was able to do almost everything I wanted to do with response wrappers.

My last use case is a caching client decorator. Same as the internal `CachingHttpClient`, but with a custom cache solution instead of HttpCache. We need to add extra headers to the cached response (age, ID, etc.). So we created a wrapper for the response and decorate the `getHeaders()` method.
With the AsyncResponse, I was able to modify the content of the original response (by yielding extra chunks in the passthru), but not the other information (status code, headers and debug infos). Do you know how I can do that ?

I saw that the `AsyncContext` has a `replaceResponse()` method. But it seems a bit overkill to recreate an entire response. And the method seem to be designed to return the result of a second HTTP request, not overwrite the first one. Maybe adding `replaceResponseHeaders()` and `replaceResponseInfo()` methods to `AsyncContext` will do the trick ? What do you think ?

I don't know if it worth it, but I added a static `stream()` method to the `DecoratorTrait` to make it work with `$client->stream()`.

PR #53336, opened 2024-01-01
[Notifier] [Firebase] Add 'HTTP v1' api endpoint

### LINKED ISSUE #51147: Firebase notifier does not work anymore since messaging protocol updated by google

6

### Description

Google disable /send api see https://firebase.google.com/docs/cloud-messaging/http-server-ref?hl=en

Caution: Sending messages (including upstream messages) with the FCM XMPP and HTTP legacy APIs was deprecated on **June 20, 2023**, and will be removed in June 2024. If you are using the legacy send APIs, we strongly recommend that you [migrate to the HTTP v1 API](https://firebase.google.com/docs/cloud-messaging/migrate-v1) or consider using the [Admin SDK](https://firebase.google.com/docs/cloud-messaging/server#choosing-a-server-option) to build send requests.

Apps using the FCM legacy HTTP API should consider migrating to the HTTP v1 API. See the [migration guide](https://firebase.google.com/docs/cloud-messaging/migrate-v1) for more information about the v1 API's advantages and for migration instruction.

So we cant send any message to old api 

### How to reproduce

When try to send a message to old api method you will get 

````
In FirebaseTransport.php line 92:
                                                             
  [Symfony\Component\Notifier\Exception\TransportException]  
  Unable to post the Firebase message: <HTML>                
  <HEAD>                                                     
  <TITLE>INVALID_KEY</TITLE>                                 
  </HEAD>                                                    
  <BODY BGCOLOR="#FFFFFF" TEXT="#000000">                    
  <H1>INVALID_KEY</H1>                                       
  <H2>Error 401</H2>                                         
  </BODY>                                                    
  </HTML>    
````

Due to     `protected const HOST = 'fcm.googleapis.com/fcm/send';`

### Possible Solution

Need to implement migrated method : https://firebase.google.com/docs/cloud-messaging/migrate-v1?hl=en

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by cesurapp**
| Q             | A
| ------------- | ---
| Branch?       | 7.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #51147
| License       | MIT

### CODE REVIEW

**Review by nicolas-grekas — COMMENTED** [review]
Thanks for working on this!
I'm wondering about two things:
- would it be possible to keep the same scheme for the new API? Maybe the shape of the user/password can help? (e.g. no password means new API?)
- more critically, I'm wondering about the need to deploy credentials.json to production, while the best practice is to use JWT tokens and to let another process refresh those tokens. Maybe we just need a way to also accept a JWT directly?

File: src/Symfony/Component/Notifier/Bridge/Firebase/FirebaseJwtTransport.php [hunk 1]
```diff
@@ -0,0 +1,135 @@
...
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
+use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+/**
+ * @author Cesur APAYDIN <https://github.com/cesurapp>
+ */
+final class FirebaseJwtTransport extends AbstractTransport
+{
+    protected const HOST = "fcm.googleapis.com/v1/projects/project_id/messages:send";
```
OskarStark:
```suggestion
    protected const HOST = 'fcm.googleapis.com/v1/projects/project_id/messages:send';
```

src/Symfony/Component/Notifier/Bridge/Firebase/FirebaseJwtTransport.php
```diff
@@ -0,0 +1,135 @@
...
+
+        $header = $this->urlSafeEncode(['alg' => 'RS256', 'typ' => 'JWT']);
+        $payload = $this->urlSafeEncode($payload);
+        openssl_sign($header . '.' . $payload, $signature, openssl_pkey_get_private($this->credentials['private_key']), OPENSSL_ALGO_SHA256);
+        $signature = $this->urlSafeEncode($signature);
+
+        return $header . '.' . $payload . '.' . $signature;
+    }
+
+    protected function urlSafeEncode($data): string
```
OskarStark:
```suggestion
    protected function urlSafeEncode(string|array $data): string
```

OskarStark:
> would it be possible to keep the same scheme for the new API? Maybe the shape of the user/password can help? (e.g. no password means new API?)

Yes we should do it somehow, having a second transport looks wrong to me

cesurapp:
> Thanks for working on this! I'm wondering about two things:
> 
> * would it be possible to keep the same scheme for the new API? Maybe the shape of the user/password can help? (e.g. no password means new API?)
> * more critically, I'm wondering about the need to deploy credentials.json to production, while the best practice is to use JWT tokens and to let another process refresh those tokens. Maybe we just need a way to also accept a JWT directly?

It seems like the only way to login to firebase is to use jwt. There is no explanation in the documentation other than JWT.

There are 4 parameters required and for compatibility we can do it like this:

    firebase-jwt://firebase-adminsdk@stag.iam.gserviceaccount.com?project_id=<PROJECT_ID>&private_key_id=<PRIVATE_KEY_ID>&private_key=<PRIVATE_KEY>

Additionally, the token created with jwt can be used for 1 hour. We can ease the encryption burden by caching and using openssl.

DaedalusDev:
> > Thanks for working on this! I'm wondering about two things:
> > 
> > * would it be possible to keep the same scheme for the new API? Maybe the shape of the user/password can help? (e.g. no password means new API?)
> > * more critically, I'm wondering about the need to deploy credentials.json to production, while the best practice is to use JWT tokens and to let another process refresh those tokens. Maybe we just need a way to also accept a JWT directly?
> 
> It seems like the only way to login to firebase is to use jwt. There is no explanation in the documentation other than JWT.
> 
> There are 4 parameters required and for compatibility we can do it like this:
> 
> ```
> firebase-jwt://firebase-adminsdk@stag.iam.gserviceaccount.com?project_id=<PROJECT_ID>&private_key_id=<PRIVATE_KEY_ID>&private_key=<PRIVATE_KEY>
> ```
> 
> Additionally, the token created with jwt can be used for 1 hour. We can ease the encryption burden by caching and using openssl.

I'm facing the same issue with an hand made integration.
I'm agree with @cesurapp, according to migration guide (https://firebase.google.com/docs/cloud-messaging/migrate-v1)
> The HTTP v1 API uses short-lived access tokens according to the OAuth2 security model.

So Bearer with infinite TTL is no longer supported by the replacement API.

aschempp:
@cesurapp thanks for the work! I duplicated this into my Symfony ([Contao](https://contao.org) on Sf 5.4) and it basically worked perfectly. However, I noticed there's quite a change in the Firebase API features. 

Looking at https://firebase.google.com/docs/cloud-messaging/migrate-v1?hl=en#update-the-payload-of-send-requests, the `to: /topics/news` has been changed to `topic: news`. This PR however rewrites `to:` to `token:`, and it **requires** `to:` to be set. The current transport makes it impossible to send to topics.

Another thing is that the API now separates configuration for Android and Apple in the same request. The current options classes do not allow these additional configurations - and they incorrectly separate Apple and Android.

I see an issue with semantic versioning here. The options classes cannt be changed until Symfony 8, but that all makes no sense in 3 months once the API is shut down. So maybe `FirebaseJwt` should be a separate implementation with separate transport and options classes?

OskarStark:
> So maybe FirebaseJwt should be a separate implementation with separate transport and options classes?

I think it makes sense as either one or the other will not work anymore, right?

cesurapp:
> > So maybe FirebaseJwt should be a separate implementation with separate transport and options classes?
> 
> I think it makes sense as either one or the other will not work anymore, right?

https://firebase.google.com/docs/cloud-messaging/migrate-v1?hl=en
> Apps using the deprecated FCM legacy APIs for HTTP and XMPP should migrate to the HTTP v1 API at the earliest opportunity. Sending messages (including upstream messages) with those APIs was deprecated on June 20, 2023, and will be removed in June 2024.

The old API will not be available after about 1 month.

aschempp:
Exactly my point. But as far as I understand Symfony's BC promise, we cannot simply remove existing classes in a minor release, even if these classes loos their functionality. But maybe @nicolas-grekas can confirm this.

fabpot:
As 7.2 is going to be released after June 2024, I recommend throwing an exception when people try to use the old way (so that we can remove the code right away).

aschempp:
@cesurapp are you planning to finish this PR for Symfony 7.2? Anything I can assist?

> Looking at https://firebase.google.com/docs/cloud-messaging/migrate-v1?hl=en#update-the-payload-of-send-requests, the `to: /topics/news` has been changed to `topic: news`. This PR however rewrites `to:` to `token:`, and it **requires** `to:` to be set. The current transport makes it impossible to send to topics.

I would suggest to parse `to:` here, and if it is `/topics/...` set it to the correct JSON property.

> Another thing is that the API now separates configuration for Android and Apple in the same request. The current options classes do not allow these additional configurations - and they incorrectly separate Apple and Android.

My solution here would be to not make the `FirebaseOptions` class abstract anymore and add necessary setters there, similar to other Notifier components. We could then throw exceptions (or deprecations if a migration is possible) if people still use the old options classes.

cesurapp:
Legacy code removed and dsn updated.

`firebase://<CLIENT_EMAIL>?project_id=<PROJECT_ID>&private_key_id=<PRIVATE_KEY_ID>&private_key=<PRIVATE_KEY>
`

src/Symfony/Component/Notifier/Bridge/Firebase/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+7.1
```
xabbuh:
```suggestion
7.2
```

src/Symfony/Component/Notifier/Bridge/Firebase/FirebaseOptions.php
```diff
@@ -29,25 +29,28 @@ abstract class FirebaseOptions implements MessageOptionsInterface
 
     private array $data;
 
-    public function __construct(string $to, array $options, array $data = [])
+    private bool $useTopic;
+
+    public function __construct(string $tokenOrTopic, array $options, array $data = [], bool $useTopic = false)
```
aschempp, line 34: would it make sense to parse `$tokenOrTopic` for a string like `/topic/...` and add a deprecation message (but handle the flag correctly)?

src/Symfony/Component/Notifier/Bridge/Firebase/FirebaseTransport.php
```diff
@@ -24,16 +24,20 @@
...
 {
-    protected const HOST = 'fcm.googleapis.com/fcm/send';
+    protected const HOST = 'fcm.googleapis.com/v1/projects/project_id/messages:send';
+
+    private array $credentials;
+
+    public function __construct(#[\SensitiveParameter] array $credentials, HttpClientInterface $client = null, EventDispatcherInterface $dispatcher = null)
+    {
+        $this->credentials = $credentials;
+        $this->client = $client;
```
aschempp, line 38: `$this->client` is `protected` in the parent class `AbstractTransport`, so no need to add it here.

src/Symfony/Component/Notifier/Bridge/Firebase/FirebaseTransport.php
```diff
@@ -24,16 +24,20 @@
 
 /**
  * @author Jeroen Spee <https://github.com/Jeroeny>
+ * @author Cesur APAYDIN <https://github.com/cesurapp>
  */
 final class FirebaseTransport extends AbstractTransport
 {
-    protected const HOST = 'fcm.googleapis.com/fcm/send';
+    protected const HOST = 'fcm.googleapis.com/v1/projects/project_id/messages:send';
```
aschempp, line 31: Looking at all the other Notifier transports, there seems to be almost none that use a full URL in the `HOST` constant – they mostly only have the host in there and build the necessary project URL in the `doSend` method. Not sure if that's really relevant, but it would prevent `$host` from always being set and the `HOST` constant from being an actually invalid URL.

aschempp (commented):
Looks good! Any plans on how to detail with the legacy Android and iOS notification classes?

src/Symfony/Component/Notifier/Bridge/Firebase/FirebaseOptions.php
cesurapp: If set to "$useTopic = true" no deprecation message will be needed. The default topic is used in the previous version.

adrolter:
Is this close to being merged? Google is actively sunsetting the legacy API and about half of my Firebase notifications are now failing to send as a result.

```
Unable to post the Firebase message: {"error":"Deprecated endpoint, see https:\/\/firebase.google.com\/docs\/cloud-messaging\/migrate-v1"}
```

Edit: Just saw this is set to land in 7.2? Is there a temporary solution in the interim to restore functionality?

aschempp:
> Just saw this is set to land in 7.2? Is there a temporary solution in the interim to restore functionality?

The temporary solution for me was to add the services from this PR into the app itself.

src/Symfony/Component/Notifier/Bridge/Firebase/FirebaseTransport.php
```diff
@@ -53,21 +57,20 @@ protected function doSend(MessageInterface $message): SentMessage
...
-        $endpoint = \sprintf('https://%s', $this->getEndpoint());
-        $options = $message->getOptions()?->toArray() ?? [];
-        $options['to'] = $message->getRecipientId();
+        $endpoint = sprintf('https://%s', $this->getEndpoint());
 
-        if (!$options['to']) {
-            throw new InvalidArgumentException(\sprintf('The "%s" transport required the "to" option to be set.', __CLASS__));
+        // Generate Options
+        $options = $message->getOptions()?->toArray() ?? [];
+        if (!$options['token'] && !$options['topic']) {
```
misaert, line 64: Is it not `if (!isset($options['token']) && !isset($options['topic'])) {`?

It generates this warning:
```
Warning: Undefined array key “token”
```

misaert:
For us, this PR doesn't work this way. We have to get the access token to call the Firebase API in a second time.

This is our worked branch: https://github.com/symfony/firebase-notifier/compare/6.4...geonativefr:firebase-notifier:geonative

cesurapp:
> For us, this PR doesn't work this way. We have to get the access token to call the Firebase API in a second time.
> 
> This is our worked branch: [symfony/firebase-notifier@6.4...geonativefr:firebase-notifier:geonative](https://github.com/symfony/firebase-notifier/compare/6.4...geonativefr:firebase-notifier:geonative)

You shouldn't use 3rd party components in Notifier packages. "google/apiclient": "^2.0@dev"

misaert:
> > For us, this PR doesn't work this way. We have to get the access token to call the Firebase API in a second time.
> > This is our worked branch: [symfony/firebase-notifier@6.4...geonativefr:firebase-notifier:geonative](https://github.com/symfony/firebase-notifier/compare/6.4...geonativefr:firebase-notifier:geonative)
> 
> You shouldn't use 3rd party components in Notifier packages. "google/apiclient": "^2.0@dev"

It is just for us to fix the push notifications.

The API client is useful to get the access token.

orangevinz:
@cesurapp I'm running a 5.4 Symfony version and I'm trying to use your patch that way but it fails
Any clue ?

```json
//...
"require": {
        "php": ">=8.1.12",
        // ...
        "symfony/firebase-notifier": "dev-feature_7.1",
        "symfony/notifier": "5.4.*",
    },
"repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/cesurapp/symfony"
        }
    ]
//...
```

cesurapp:
> @cesurapp I'm running a 5.4 Symfony version and I'm trying to use your patch that way but it fails Any clue ?
> 
> ```json
> //...
> "require": {
>         "php": ">=8.1.12",
>         // ...
>         "symfony/firebase-notifier": "dev-feature_7.1",
>         "symfony/notifier": "5.4.*",
>     },
> "repositories": [
>         {
>             "type": "vcs",
>             "url": "https://github.com/cesurapp/symfony"
>         }
>     ]
> //...
> ```

This package required symfony 6.4+ , I have not tested it on Symfony 5x version.

rsereir:
Hello all !

This MR is ready ?  I am trying the current version but it looks not working correctly ( i didn't know how to have a username / password combination on my Firebase dashboard ). It looks be fixed by the new way to call firebase

fabpot:
> Hello all !
> 
> This MR is ready ? I am trying the current version but it looks not working correctly ( i didn't know how to have a username / password combination on my Firebase dashboard ). It looks be fixed by the new way to call firebase

AFAICS, there are still some comments that are not addressed yet.

SherinBloemendaal:
Would be great to have this finished before 7.3. Let me know if I can help in anyway!

OskarStark:
@cesurapp are you interested in finishing this PR?

cesurapp:
> @cesurapp are you interested in finishing this PR?

Sorry I don't have time for that right now.

OskarStark:
> Sorry I don't have time for that right now.

No worries, thanks for the quick feedback.

@SherinBloemendaal do you want to take over?

aschempp:
I'm also happy to contribute my fork that is based on this and running for months now in my Symfony 5 setup. But I didn't wanna take away @cesurapp's work.

SherinBloemendaal:
I do have time to work on it but if @aschempp already has a working version, we can use that of course :)

aschempp:
I looked at my stuff, unfortunately I copied files from this PR to my client's App and modified to make it work (so not actually a fork). Appart from probably not being really usable, I think we need to solve the problem of having an existing API (in Symfony) that no longer works in Firebase. This PR originally introduced a BC layer for the old configuration, but that BC layer makes little sence since the old API does not work anymore anyway.

I wonder if we should throw away all of the current stuff, which will result in an error for existing users. Not sure we can do that in a minor release, but it did cause errors anyway as far as I know, so it can't really be used by anyone?

SherinBloemendaal:
I understand that removing the code might seem logical. But even though Firebase's legacy API is returning 404s which always results in a `TransportException` being thrown, we still need to maintain backward compatibility at the Symfony level in my opinion and according to Symfony's BC policy.

What we could do:
- 7.3: Introduce new Firebase V1 API support and deprecate the legacy API
- 8.0: Drop support for the legacy API

aschempp:
sounds good to me. Just be aware that the separation into `AndroidNotification` and `IOSNotification` etc. do not make sense with the new API, so we need a BC way to keep those classes as well as some new options object to handle the new API. Feel free to take on this if you have some spare time 😊

vojtechsmejkal:
Hi, I needed to use FCM in one of our projects, so I created a working version based on what @cesurapp prepared in this PR.

I was wondering if anyone is currently working on this feature. I don't want to take over someone else's PR, but if no one is working on it, I'm happy to take it on. :slightly_smiling_face:

fabpot:
> Hi, I needed to use FCM in one of our projects, so I created a working version based on what @cesurapp prepared in this PR.
> 
> I was wondering if anyone is currently working on this feature. I don't want to take over someone else's PR, but if no one is working on it, I'm happy to take it on. 🙂

Apparently, @cesurapp does not have time anymore (see https://github.com/symfony/symfony/pull/53336#issuecomment-2768733018). So, feel free to take over.

vojtechsmejkal:
Thank you, I will take over and post an update at the end of this week.

vojtechsmejkal:
A quick update as promised:

I have created a new PR https://github.com/symfony/symfony/pull/60205 for adding support for Firebase v1 API and I think it would be best to close this PR and continue the discussion in the new one.

PR #53341, opened 2024-01-02
[FrameworkBundle] append instead of replacing potentially non-existent named-arguments

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix https://github.com/symfony/symfony/pull/52998#issuecomment-1873746495
| License       | MIT

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
@@ -2594,27 +2594,27 @@ class FrameworkExtension extends Extension
 
         if (ContainerBuilder::willBeAvailable('symfony/mercure-notifier', MercureTransportFactory::class, $parentPackages, true) && ContainerBuilder::willBeAvailable('symfony/mercure-bundle', MercureBundle::class, $parentPackages, true) && \in_array(MercureBundle::class, $container->getParameter('kernel.bundles'), true)) {
             $container->getDefinition($classToServices[MercureTransportFactory::class])
-                ->replaceArgument('$registry', new Reference(HubRegistry::class))
-                ->replaceArgument('$client', new Reference('http_client', ContainerBuilder::NULL_ON_INVALID_REFERENCE))
-                ->replaceArgument('$dispatcher', new Reference('event_dispatcher', ContainerBuilder::NULL_ON_INVALID_REFERENCE));
+                ->replaceArgument(0, new Reference(HubRegistry::class))
+                ->replaceArgument(1, new Reference('event_dispatcher', ContainerBuilder::NULL_ON_INVALID_REFERENCE))
+                ->addArgument(new Reference('http_client', ContainerBuilder::NULL_ON_INVALID_REFERENCE));
         } elseif (ContainerBuilder::willBeAvailable('symfony/mercure-notifier', MercureTransportFactory::class, $parentPackages, true)) {
             $container->removeDefinition($classToServices[MercureTransportFactory::class]);
         }
 
         if (ContainerBuilder::willBeAvailable('symfony/fake-chat-notifier', FakeChatTransportFactory::class, ['symfony/framework-bundle', 'symfony/notifier', 'symfony/mailer'], true)) {
             $container->getDefinition($classToServices[FakeChatTransportFactory::class])
-                ->replaceArgument('$mailer', new Reference('mailer'))
-                ->replaceArgument('$logger', new Reference('logger'))
-                ->replaceArgument('$client', new Reference('http_client', ContainerBuilder::NULL_ON_INVALID_REFERENCE))
-                ->replaceArgument('$dispatcher', new Reference('event_dispatcher', ContainerBuilder::NULL_ON_INVALID_REFERENCE));
+                ->replaceArgument(0, new Reference('mailer'))
+                ->replaceArgument(1, new Reference('logger'))
+                ->addArgument(new Reference('event_dispatcher', ContainerBuilder::NULL_ON_INVALID_REFERENCE))
+                ->addArgument(new Reference('http_client', ContainerBuilder::NULL_ON_INVALID_REFERENCE));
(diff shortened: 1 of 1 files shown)


OskarStark (approved):

nicolas-grekas (commented):
Can't we stop using named args at all?
So those args are already wired in the default config files?

xabbuh:
> Can't we stop using named args at all?
> So those args are already wired in the default config files?

That's what we do now. Or do I misunderstand what you are asking for?

rdavaillaud:
The problem is that the bridges are wired via a factory that is declared via an abstract factory.

see https://github.com/symfony/symfony/blob/317c67bed8f4ede15828438b46d4e96313b2442e/src/Symfony/Bundle/FrameworkBundle/Resources/config/notifier_transports.php#L151

As far as I have tested, if we remove the HttpClient and Dispatcher from the definition, they will not be injected in the constructor.

xabbuh:
I mean we revert the framework extension changes made in #52998 which were not necessary because of the `notifier.transport_factory.abstract` definition: https://github.com/symfony/symfony/blob/5.4/src/Symfony/Bundle/FrameworkBundle/Resources/config/notifier_transports.php#L90-L92

rdavaillaud:
If we revert the changes, the DI will generate this code (at least in 6.3):

```php
class getNotifier_TransportFactory_FakesmsService extends App_KernelDevDebugContainer
{
    /**
     * Gets the private 'notifier.transport_factory.fake-sms' shared service.
     *
     * @return \Symfony\Component\Notifier\Bridge\FakeSms\FakeSmsTransportFactory
     */
    public static function do($container, $lazyLoad = true)
    {
        include_once \dirname(__DIR__, 4).'/vendor/symfony/notifier/Transport/TransportFactoryInterface.php';
        include_once \dirname(__DIR__, 4).'/vendor/symfony/notifier/Transport/AbstractTransportFactory.php';
        include_once \dirname(__DIR__, 4).'/vendor/symfony/fake-sms-notifier/FakeSmsTransportFactory.php';

        return new \Symfony\Component\Notifier\Bridge\FakeSms\FakeSmsTransportFactory(($container->privates['mailer.mailer'] ?? self::getMailer_MailerService($container)), ($container->privates['monolog.logger'] ?? self::getMonolog_LoggerService($container)));
    }
}
```

The HttpClient and EventDispatcher are replaced, not injected.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2594,27 +2594,21 @@ private function registerNotifierConfiguration(array $config, ContainerBuilder $
 
         if (ContainerBuilder::willBeAvailable('symfony/mercure-notifier', MercureTransportFactory::class, $parentPackages, true) && ContainerBuilder::willBeAvailable('symfony/mercure-bundle', MercureBundle::class, $parentPackages, true) && \in_array(MercureBundle::class, $container->getParameter('kernel.bundles'), true)) {
             $container->getDefinition($classToServices[MercureTransportFactory::class])
-                ->replaceArgument('$registry', new Reference(HubRegistry::class))
-                ->replaceArgument('$client', new Reference('http_client', ContainerBuilder::NULL_ON_INVALID_REFERENCE))
-                ->replaceArgument('$dispatcher', new Reference('event_dispatcher', ContainerBuilder::NULL_ON_INVALID_REFERENCE));
+                ->replaceArgument('$registry', new Reference(HubRegistry::class));
```
nicolas-grekas: can't we use index_foo instead of named args? that's what I meant :)

xabbuh: I didn't touch this line as it already has been like this before #52998

nicolas-grekas: I would still remove it, core shouldn't use reflection to do its wiring. We don't do this anywhere else but in one case where it's on purpose. Here, it's just by laziness :)

rdavaillaud:
If we revert the FrameworkBundle, we must also change the signatures of the bridges Fake* constructor to put Mailer and Logger at the end.

rdavaillaud:
@xabbuh your last change seems to work with the old bridge, and the modified one. :+1:

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2594,27 +2594,27 @@ private function registerNotifierConfiguration(array $config, ContainerBuilder $
 
         if (ContainerBuilder::willBeAvailable('symfony/mercure-notifier', MercureTransportFactory::class, $parentPackages, true) && ContainerBuilder::willBeAvailable('symfony/mercure-bundle', MercureBundle::class, $parentPackages, true) && \in_array(MercureBundle::class, $container->getParameter('kernel.bundles'), true)) {
             $container->getDefinition($classToServices[MercureTransportFactory::class])
-                ->replaceArgument('$registry', new Reference(HubRegistry::class))
-                ->replaceArgument('$client', new Reference('http_client', ContainerBuilder::NULL_ON_INVALID_REFERENCE))
-                ->replaceArgument('$dispatcher', new Reference('event_dispatcher', ContainerBuilder::NULL_ON_INVALID_REFERENCE));
+                ->replaceArgument(0, new Reference(HubRegistry::class))
+                ->addArgument(new Reference('event_dispatcher', ContainerBuilder::NULL_ON_INVALID_REFERENCE))
```
nicolas-grekas: replaceArgument here also to not pass the client as 2nd arg as defined by the abstract transport?

nicolas-grekas:
Thank you @xabbuh.

xabbuh:
@rdavaillaud Thank you for testing it!

PR #36999, opened 2020-05-28
New DSN component

Nyholm:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | See [documentation.md](https://github.com/Nyholm/symfony/blob/dsn/src/Symfony/Component/Dsn/documentation.md)

A few years ago I created [nyholm/dsn](https://packagist.org/packages/nyholm/dsn). This PR is building on to the work started in that library. 

I've seen discussions in [about a new DSN component](https://github.com/symfony/symfony/issues/32546) started by @Simperfit. He points out that 7 components is using some form of DSN. I've also seen @jderusse PR in https://github.com/symfony/symfony/pull/24267 where he creates a specific DSN factory class for each component. 

What I found interesting is @csarrazi's comment (https://github.com/symfony/symfony/issues/32546#issuecomment-511900064) saying "there is not clear spec for DSNs". I have also not found a definition myself so I thought I would start by defining that a DSN is. I looked on all uses of DSN in symfony/symfony. I later broaden my search to DSN in PHP and continued to other languages. I came up with the following definition based on [RFC 3986](https://tools.ietf.org/html/rfc3986):

```
dsn_string:
  { function | dsn }

function:
  function_name[:](dsn_string[,dsn_string])[?query]

function_name:
  REGEX: [a-zA-Z0-9\+-]+

dsn:
  { scheme:[//]authority[path][?query] | scheme:[//][userinfo]path[?query] | host:port[path][?query] }

// ...

```

(See full definition in the [docs](https://github.com/Nyholm/symfony/blob/dsn/src/Symfony/Component/Dsn/documentation.md#definition))

Please correct me if I'm wrong and there is a DSN RFC somewhere. 

### Implementation

As you may see from the implementation or from how Symfony is using DSNs, the new component is basically just a bit more advanced `parse_url()` with a few benefits: 

- Comes with a definition
- Better exceptions
- Automatically url-decodes username and passwords
- Value objects instead of arrays for the url parts
- Handles all kinds of functions
- Supports DSNs like: memcached://user:pass@/var/run/memcached.sock

### Documentation

I've started a page that could be converted to an PR to symfony/symfony-docs. I'll refer to that page as detailed PR description. See [documentation.md](https://github.com/Nyholm/symfony/blob/dsn/src/Symfony/Component/Dsn/documentation.md).

### Usage

[...]

Issue #32546, opened 2019-07-15
[DSN] Add a new DSN component


Simperfit:
**Description**  

Following this comment from @fabpot: https://github.com/symfony/symfony/pull/32392#issuecomment-509149726 and searching for information in the symfony repository 
I've seen an attempt from @jderusse https://github.com/symfony/symfony/pull/24267. I still think the proposal was a good idea and we need a DSN component, we could even imagine a DSN contract, let me explain why:
I've looked into all of the symfony components to look for DSN usage, it seems that we use lot's of DSN and even in client project we use more and more  DSN.

Lot's of symfony-linked librairies (doctrine, flysystem) use a DSN too, so instead of rechecking and reparsing the DSN all the time, we could have a simple `dsn-parser` that would do the job and return a DSN parsed as a PHP array/object.

If we talk about usage in symfony:

- In the `Cache`  💾 component where are parsing two time the `redis` DSN:
In the redis trait: https://github.com/symfony/symfony/blob/9ab4f14a6ed98d5e70e1d01abaefd8c31b815e1c/src/Symfony/Component/Cache/Traits/RedisTrait.php#L84-L90
And in the adapter:
https://github.com/symfony/symfony/blob/9ab4f14a6ed98d5e70e1d01abaefd8c31b815e1c/src/Symfony/Component/Cache/Adapter/AbstractAdapter.php#L136-L141
We are also parsing a memcache DSN in the adapter.
Both of the redis check could be merged into one and an object could be passed between the two. The exception logic would be merge an a new exception would be thrown.

- In the `Messenger`📮  component we are using DSN in connections classes and in the supports method, maybe a trait comming from the dns parser would help parsing this directly. I give you the exact lines of the code that each connection classes is doing but we know that's Doctrine AMQP, and redis. As an exemple we will only take doctrine.
https://github.com/symfony/symfony/blob/9ab4f14a6ed98d5e70e1d01abaefd8c31b815e1c/src/Symfony/Component/Messenger/Transport/Doctrine/Connection.php#L68-L70

- In `HttpKernel` 📟 we parse a dsn too, that's the file DSN, it could be reworked to avoid parsing it in here and just getting the value without using substr:
https://github.com/symfony/symfony/blob/9ab4f14a6ed98d5e70e1d01abaefd8c31b815e1c/src/Symfony/Component/HttpKernel/Profiler/FileProfilerStorage.php#L37-L39

[...]

PR #53346, opened 2024-01-02
[HttpFoundation] Add `Url` and `UrlParser`

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

This PR https://github.com/symfony/symfony/pull/53320 raised a problem in the codebase. Each place in the code where a DSN is parsed has its own implemented logic.

This led to various consistency issues: variable names differing from one place to another but having the same meaning, different behavior (some places decode HTML characters in URLs while others don't), work done "twice"... And of course, if a bug is detected somewhere, you have to check that it's not present everywhere (as must have been done with the PR mentioned above).

@OskarStark and @nicolas-grekas raised that there could be some place for something that unifies the work of DSN parsing which is of course always the same: parse the URL, decode the auth part with special chars, check a scheme is present, check the host is optionally present also.

Here is a _first_ take on this. I created the logic in HttpFoundation, because why not, but I don't think this is the best place. However, we don't have any idea on where to put this as you can see in the comments of the linked PR 😄 I'm even sure that this is not the right place because, to me, add HttpFoundation as a dependency for this DSN parser feels weird.

A new component might be the solution, but it seems it was already discussed in the past and it was rejected (note that I couldn't find info about this, so if you were part of the discussion, feel free to link some issue or PR, thanks!).

This PR will give you an idea of what could be possible. I already leveraged this new mechanism in DI, Notifier, Mailer and Translation so you can see this in action. Of course, if this get accepted, the next step will be to leverage this in the Cache component mainly.

Things to discuss:
- What are your thoughts on the API?
- Where should we put this (new component, in an existing component, something else)?
- Which additional feature should be bundled in Parser or Parameters?

### DIFF
--- a/src/Symfony/Component/HttpFoundation/Exception/Parser/InvalidUrlException.php
+++ b/src/Symfony/Component/HttpFoundation/Exception/Parser/InvalidUrlException.php
@@ -1,0 +1,20 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\HttpFoundation\Exception\Parser;
+
+class InvalidUrlException extends \InvalidArgumentException
+{
+    public function __construct()
+    {
+        parent::__construct('The URL is invalid.');
+    }
+}
--- a/src/Symfony/Component/HttpFoundation/UrlParser/Url.php
+++ b/src/Symfony/Component/HttpFoundation/UrlParser/Url.php
@@ -1,0 +1,88 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\HttpFoundation\UrlParser;
+
+/**
+ * @author Alexandre Daubois <alex.daubois@gmail.com>
+ */
+final class Url implements \Stringable
+{
+    public function __construct(
+        public string $scheme,
+        public ?string $user = null,
+        public ?string $password = null,
+        public ?string $host = null,
+        public ?int $port = null,
+        public ?string $path = null,
+        public ?string $query = null,
+        public ?string $fragment = null
+    ) {
+    }
+
+    public function isAuthenticated(): bool
+    {
+        return null !== $this->user || null !== $this->password;
+    }
+
+    public function isScheme(string $scheme): bool
+    {
+        return $this->scheme === $scheme;
+    }
+
+    public function __toString(): string
+    {
+        $dsn = $this->scheme.'://';
+
+        if (null !== $this->user) {
+            $dsn .= rawurlencode($this->user);
+        }
+
+        if (null !== $this->password) {
+            $dsn .= ':'.rawurlencode($this->password);
+        }
+
+        if (null !== $this->user || null !== $this->password) {
+            $dsn .= '@';
+        }
+
+        $dsn .= $this->host;
+
+        if (null !== $this->port) {
+            $dsn .= ':'.$this->port;
+        }
+
+        if (null !== $this->path) {
+            $dsn .= $this->path;
+        }
+
+        if (null !== $this->query) {
+            $dsn .= '?'.$this->query;
+        }
+
+        if (null !== $this->fragment) {
+            $dsn .= '#'.$this->fragment;
+        }
+
+        return $dsn;
+    }
+
+    public function parsedQuery(): array
+    {
+        if (null === $this->query) {
+            return [];
+        }
+
+        parse_str($this->query, $query);
+
+        return $query;
+    }
+}
--- a/src/Symfony/Component/Mailer/Tests/Transport/DsnTest.php
+++ b/src/Symfony/Component/Mailer/Tests/Transport/DsnTest.php
@@ -92,17 +92,17 @@ class DsnTest extends TestCase
     {
         yield [
             'some://',
-            'The mailer DSN is invalid.',
+            'The URL is invalid.',
         ];
 
         yield [
             '//sendmail',
-            'The mailer DSN must contain a scheme.',
+            'The URL must contain a scheme.',
         ];
 
         yield [
             'file:///some/path',
-            'The mailer DSN must contain a host (use "default" by default).',
+            'The URL must contain a host (use "default" by default).',
         ];
     }
 }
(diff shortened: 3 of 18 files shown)


src/Symfony/Component/HttpFoundation/Dsn/Parameters.php
```diff
@@ -0,0 +1,80 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\HttpFoundation\Dsn;
+
+/**
+ * @author Alexandre Daubois <alex.daubois@gmail.com>
+ */
+final class Parameters implements \Stringable
```
OskarStark: I would still name this `àI would name this class `Dsn`

src/Symfony/Component/Mailer/Tests/Transport/DsnTest.php
```diff
@@ -92,17 +92,17 @@ public static function invalidDsnProvider(): iterable
     {
         yield [
             'some://',
-            'The mailer DSN is invalid.',
+            'The "some://" DSN is invalid.',
```
OskarStark: Exposing the DSN here means it could leak sensitive data to a log file

xabbuh:
see #32546 for a related discussion

Nyholm:
Here is a link to a PR I made to try introduce a DSN component. 
https://github.com/symfony/symfony/pull/36999

Symfony is all about following official standards. There is no standard for DSN (believe it or not). That is why it might be hard to push for a dedicated DSN component.

nicolas-grekas:
:100: That's why I wrote we might need a URL parser, not a DSN parser ;)

alexandre-daubois:
Alright I see the distinction. A bit of renaming is needed here. An URL parser could belong to HttpFoundation, as it already has some "UrlHelper".

OskarStark:
> Symfony is all about following official standards. There is no standard for DSN (believe it or not). That is why it might be hard to push for a dedicated DSN component.

I understand, but we created our own standard of DSN in the Symfony world, which is spread across different components

nicolas-grekas:
> we created our own standard of DSN in the Symfony world

That's not the case, each DSN in Symfony is potentially adhoc for the components that need one. There are already differences that are hard to abstract.

Nyholm:
> > Symfony is all about following official standards. There is no standard for DSN (believe it or not). That is why it might be hard to push for a dedicated DSN component.
>
> I understand, but we created our own standard of DSN in the Symfony world, which is spread across different components

Yes, that work has been **implementation** driven. I looked at how Symfony is using DSN and I created a **specification** for it. But making it a dedicated component says more "This is how Symfony think DSN should work". If we are just implementing DSN in various other components means we can eat the cake and have it too (ie don't agree on a standard but use DSN as we like). 

Im obviously not against a DSN component or let a official DSN parser be part of HTTPFoundation, a new URI component. Im just trying to give some background or a TLDR for https://github.com/symfony/symfony/pull/36999 and https://github.com/symfony/symfony/issues/32546.

alexandre-daubois:
>  There are already differences that are hard to abstract.

Would you have an example? From what I saw during the `rawurldecode` URL, "everything" seems pretty aligned. Isn't the differences something that could be solved with some good polymorphism?

> Im just trying to give some background or a TLDR

Thank you for that! 🙏

Nyholm:
> Would you have an example?

It is mostly the "function" part. Some components supports DSN like: 

```
failover(dummy://a dummy://a)
failover(dummy://a,dummy://a)
failover:(dummy://a,dummy://a)
roundrobin(dummy://a failover(dummy://b dummy://a) dummy://b)
```

alexandre-daubois:
Just a quick thought but shouldn't we just use Tobias' package in the framework?

src/Symfony/Component/HttpFoundation/Exception/Parser/InvalidUrlException.php
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\HttpFoundation\Exception\Parser;
+
+class InvalidUrlException extends \InvalidArgumentException
+{
+    public function __construct(string $dsn)
```
smnandre: `string $url`  ?

src/Symfony/Component/HttpFoundation/UrlParser/Url.php
```diff
@@ -0,0 +1,80 @@
...
+    {
+        return $this->scheme === $scheme;
+    }
+
+    public function __toString(): string
+    {
+        $dsn = $this->scheme.'://';
+
+        if ($this->isAuthenticated()) {
+            $dsn .= rawurlencode($this->user).':'.rawurlencode($this->pass).'@';
```
smnandre: Will throw if one of `this->user` or `this->pass` is null, as [rawurlencode](https://www.php.net/manual/fr/function.rawurlencode.php)() requires a string

src/Symfony/Component/HttpFoundation/UrlParser/Url.php
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * @author Alexandre Daubois <alex.daubois@gmail.com>
+ */
+final class Url implements \Stringable
+{
+    public function __construct(
+        public string $scheme,
+        public ?string $user = null,
+        public ?string $pass = null,
```
smnandre: Could we use `password` here ?

src/Symfony/Component/HttpFoundation/Exception/Parser/InvalidUrlException.php
OskarStark: parameter can be removed IMHO

src/Symfony/Component/HttpFoundation/UrlParser/Url.php
alexandre-daubois: Good catch 👍

src/Symfony/Component/HttpFoundation/UrlParser/Url.php
alexandre-daubois: We could, but this would make impossible to do the following:

```php
$url = new Url(...parse_url($url));
```

This is why I kept `pass` here

nicolas-grekas: the goal of the new code should be to completely replace using parse_url, so this argument is void sorry. (parse_url is broken)

alexandre-daubois: Oh yes, of course 🤦 Updated accordingly

welcoMattic:
Is `HttpFoundation` the best place for such a parser? It will be used on Notifier, Translation, etc, in not necessarily HTTP request related context. I'm just back from holidays, maybe the topic has already been discussed elsewhere, but why not creating a small component named `UrlParser`?

alexandre-daubois:
Closed in favor of https://github.com/symfony/polyfill/pull/498

PR #53347, opened 2024-01-02
[Asset][BrowserKit][Cache][Console][CssSelector] Use CPP

OskarStark:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | --
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Asset/Package.php
+++ b/src/Symfony/Component/Asset/Package.php
@@ -23,12 +23,12 @@ use Symfony\Component\Asset\VersionStrategy\VersionStrategyInterface;
  */
 class Package implements PackageInterface
 {
-    private VersionStrategyInterface $versionStrategy;
     private ContextInterface $context;
 
-    public function __construct(VersionStrategyInterface $versionStrategy, ContextInterface $context = null)
-    {
-        $this->versionStrategy = $versionStrategy;
+    public function __construct(
+        private VersionStrategyInterface $versionStrategy,
+        ContextInterface $context = null,
+    ) {
         $this->context = $context ?? new NullContext();
     }
 
--- a/src/Symfony/Component/Asset/VersionStrategy/StaticVersionStrategy.php
+++ b/src/Symfony/Component/Asset/VersionStrategy/StaticVersionStrategy.php
@@ -18,16 +18,16 @@ namespace Symfony\Component\Asset\VersionStrategy;
  */
 class StaticVersionStrategy implements VersionStrategyInterface
 {
-    private string $version;
     private string $format;
 
     /**
      * @param string $version Version number
      * @param string $format  Url format
      */
-    public function __construct(string $version, string $format = null)
-    {
-        $this->version = $version;
+    public function __construct(
+        private string $version,
+        string $format = null,
+    ) {
         $this->format = $format ?: '%s?%s';
     }
 
--- a/src/Symfony/Component/BrowserKit/Request.php
+++ b/src/Symfony/Component/BrowserKit/Request.php
@@ -16,37 +16,29 @@ namespace Symfony\Component\BrowserKit;
  */
 class Request
 {
-    protected string $uri;
-    protected string $method;
-    protected array $parameters;
-    protected array $files;
-    protected array $cookies;
-    protected array $server;
-    protected ?string $content;
-
     /**
-     * @param string $uri        The request URI
-     * @param string $method     The HTTP method request
-     * @param array  $parameters The request parameters
-     * @param array  $files      An array of uploaded files
-     * @param array  $cookies    An array of cookies
-     * @param array  $server     An array of server parameters
-     * @param string $content    The raw body data
+     * @param string      $uri        The request URI
+     * @param string      $method     The HTTP method request
+     * @param array       $parameters The request parameters
+     * @param array       $files      An array of uploaded files
+     * @param array       $cookies    An array of cookies
+     * @param array       $server     An array of server parameters
+     * @param string|null $content    The raw body data
      */
-    public function __construct(string $uri, string $method, array $parameters = [], array $files = [], array $cookies = [], array $server = [], string $content = null)
-    {
-        $this->uri = $uri;
-        $this->method = $method;
-
+    public function __construct(
+        protected string $uri,
+        protected string $method,
+        protected array $parameters = [],
+        protected array $files = [],
+        protected array $cookies = [],
+        protected array $server = [],
+        protected ?string $content = null,
+    ) {
         array_walk_recursive($parameters, static function (&$value) {
             $value = (string) $value;
         });
 
         $this->parameters = $parameters;
-        $this->files = $files;
-        $this->cookies = $cookies;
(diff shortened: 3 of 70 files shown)


nicolas-grekas:
It'd be great if you could group your next PRs on the topic in a bigger one :pray:

OskarStark:
Sure, let this one open, I can work on this PR

OskarStark:
Can be merged please

src/Symfony/Component/Asset/Package.php
```diff
@@ -23,12 +23,12 @@
...
     private ContextInterface $context;
 
-    public function __construct(VersionStrategyInterface $versionStrategy, ContextInterface $context = null)
-    {
-        $this->versionStrategy = $versionStrategy;
+    public function __construct(
+        private VersionStrategyInterface $versionStrategy,
+        ContextInterface $context = null,
+    ) {
         $this->context = $context ?? new NullContext();
```
wouterj, line 32:
```suggestion
        ContextInterface $context = new NullContext(),
    ) {
```

src/Symfony/Component/Asset/VersionStrategy/StaticVersionStrategy.php
```diff
@@ -18,16 +18,16 @@
...
      * @param string $format  Url format
      */
-    public function __construct(string $version, string $format = null)
-    {
-        $this->version = $version;
+    public function __construct(
+        private string $version,
+        string $format = null,
+    ) {
         $this->format = $format ?: '%s?%s';
```
wouterj, line 31:
```suggestion
        private string $format = '%s?%s',
    ) {
```

src/Symfony/Component/BrowserKit/Request.php
```diff
@@ -16,37 +16,31 @@
...
+     * @param array       $files      An array of uploaded files
+     * @param array       $cookies    An array of cookies
+     * @param array       $server     An array of server parameters
+     * @param string|null $content    The raw body data
      */
-    public function __construct(string $uri, string $method, array $parameters = [], array $files = [], array $cookies = [], array $server = [], string $content = null)
+    public function __construct(
+        private string $uri,
+        private string $method,
+        array $parameters = [],
```
nicolas-grekas: I missed checking that on previous PRs on the topic, but here we can declare this argument using CPP.
Can you check of that's the case is other locations where we don't use CPP but we could, even if we assign the property in the body?
(The most common case where we cannot is when the argument accepts null but the property doesn't.)

src/Symfony/Component/Cache/Marshaller/DefaultMarshaller.php
```diff
@@ -21,17 +21,17 @@
 class DefaultMarshaller implements MarshallerInterface
 {
     private bool $useIgbinarySerialize = true;
-    private bool $throwOnSerializationFailure = false;
 
-    public function __construct(bool $useIgbinarySerialize = null, bool $throwOnSerializationFailure = false)
-    {
+    public function __construct(
+        bool $useIgbinarySerialize = null,
+        private bool $throwOnSerializationFailure = false,
```
wouterj: Should be reverted, property has a default value

src/Symfony/Component/Asset/Package.php
OskarStark: This is a BC break, as you cannot use `null` anymore

https://3v4l.org/mm35O

src/Symfony/Component/BrowserKit/Request.php
OskarStark: Ok for now I used it like "when we do anything with the given argument, don't Use CPP. Only Use CPP if we set the exact same value to the property"

so here you want to Use CPP for `$arguments`, even we are doing the `array_walk_recursive` right?

src/Symfony/Component/Asset/VersionStrategy/StaticVersionStrategy.php
OskarStark: Same as above

src/Symfony/Component/Cache/Marshaller/DefaultMarshaller.php
OskarStark: Ah you mean it has a default value on the property, right?

src/Symfony/Component/BrowserKit/Request.php
```diff
@@ -16,37 +16,31 @@
...
+     * @param array       $parameters The request parameters
+     * @param array       $files      An array of uploaded files
+     * @param array       $cookies    An array of cookies
+     * @param array       $server     An array of server parameters
+     * @param string|null $content    The raw body data
      */
-    public function __construct(string $uri, string $method, array $parameters = [], array $files = [], array $cookies = [], array $server = [], string $content = null)
+    public function __construct(
+        private string $uri,
+        private string $method,
```
OskarStark:
```suggestion
        protected string $uri,
        protected string $method,
```

src/Symfony/Component/BrowserKit/Request.php
```diff
@@ -16,37 +16,31 @@
...
+        array $parameters = [],
+        private array $files = [],
+        private array $cookies = [],
+        private array $server = [],
+        private ?string $content = null)
```
OskarStark:
```suggestion
        protected array $files = [],
        protected array $cookies = [],
        protected array $server = [],
        protected ?string $content = null,
        )
```

src/Symfony/Component/BrowserKit/Request.php
OskarStark: Done ✅

src/Symfony/Component/Asset/Package.php
wouterj: Ah, I see. Good thinking!

src/Symfony/Component/Cache/Marshaller/DefaultMarshaller.php
wouterj: Yes :)

src/Symfony/Component/Console/Formatter/OutputFormatterStyleStack.php
```diff
@@ -24,11 +24,9 @@ class OutputFormatterStyleStack implements ResetInterface
      */
     private array $styles = [];
 
-    private OutputFormatterStyleInterface $emptyStyle;
-
-    public function __construct(OutputFormatterStyleInterface $emptyStyle = null)
-    {
-        $this->emptyStyle = $emptyStyle ?? new OutputFormatterStyle();
+    public function __construct(
+        private OutputFormatterStyleInterface $emptyStyle = new OutputFormatterStyle(),
```
wouterj: We should revert this one then :)

OskarStark: Indeed, thanks

nicolas-grekas:
Thank you @OskarStark.

PR #53348, opened 2024-01-02
[HttpFoundation] Fix Request::getContent() phpdoc return value

tuomasva:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Issues        | no issues
| License       | MIT

According to php documentation all of the file-like operations in this method can return false as well.
Not sure if file_get_contents() on php://input can actually never fail, but it would be safe to assume it can.

This should not be any kind of BC but it can lead to projects needing for example a phpstan baseline updates when taken into use or can cause fails on CI pipelines etc.

### DIFF
--- a/src/Symfony/Component/HttpFoundation/Request.php
+++ b/src/Symfony/Component/HttpFoundation/Request.php
@@ -1531,9 +1531,11 @@ class Request
     /**
      * Returns the request body content.
      *
-     * @param bool $asResource If true, a resource will be returned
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Thanks for the PR but this looks incorrect to me. getContent doesn't return false on error. If there is a scenario where false would be tentatively returned, this is what needs a fix. But changing the public API for internal implementation reasons is doing things the reverse way.

PR #53353, opened 2024-01-02
[Form] Fix partial data submitting with collections and files (WIP)

### LINKED ISSUE #53359: [Form] Mismatched index during merge of params/files after submit form with files/collections/checkbox
## Symfony version(s) affected

5.4

## Description

Since PR #52255 (or #52347), merge of files/params after submitting a form seems to not result in a correct form data when using collection of files AND fields that are not POST for each element.

## How to reproduce

I created a PR with a unit test to try to explain at best how we can reproduce this edge case. See #53353 

We could certainty found a more concise example, but here is a form configuration where you could reproduce the bug:

1. Use a formType as bellow _(field names are added to help)_ : 
* CollectionType _(collection)_
  * CheckboxType _(checkbox)_
  * CollectionType _(files)_
    * FileType _(file)_

2. Have a root collection with 3 elements
3. Make sure to have a file uploaded in each FileType
4. Have all checkbox **unchecked**
5. Check the checkbox of item n°2
6. Submit

We got the error "This form should not contain extra fields.".

PS: I didnt get time to create a small project to reproduce issue, sorry.

## Possible Solution / Root cause

I dig in into this error, and here is what I found.

Merge of files data and params data are done by static function [FormUtil::mergeParamsAndFiles](https://github.com/symfony/symfony/blob/bf62cbe5db52d4832355af6f14ffd9e2e1131c9d/src/Symfony/Component/Form/Util/FormUtil.php#L51).

### Params

When we submit our form with only item n°2 checked, params will only contains thoses values: 

```php
[
  "collection" => [
    1 => [
      "checkbox" => "1"
    ]
  ]
]
```

**Very important note here** (and maybe the unique cause ?), fields from collections items number 1 and 3 are not submitted here. It's explained as **checkbox are never submitted in form when they are unchecked**.

Extracted from [W3C](https://www.w3.org/TR/2012/WD-html5-20120329/form-submission.html#constructing-form-data-set) : 

> If any of the following conditions are met, then skip these substeps for this element:
>     [...]
>     The field element is an [input](https://www.w3.org/TR/2012/WD-html5-20120329/the-input-element.html#the-input-element) element whose [type](https://www.w3.org/TR/2012/WD-html5-20120329/the-input-element.html#attr-input-type) attribute is in the [Checkbox](https://www.w3.org/TR/2012/WD-html5-20120329/states-of-the-type-attribute.html#checkbox-state-type-checkbox) state and whose [checkedness](https://www.w3.org/TR/2012/WD-html5-20120329/attributes-common-to-form-controls.html#concept-fe-checked) is false.

### Files

Then we have our files, where each item collection have at least one, so our data when submitting ($_FILES) will look like this: 

```php
[
  "collection" => [
    0 => [
      "files" => [
      	0 => FileObject
      ]
    ],
    1 => [
      "files" => [
      	0 => FileObject
      ]
    ],
    2 => [
      "files" => [
      	0 => FileObject
      ]
    ]
  ]
]
```

Nothing anormal here.

### Merge

Here it comes...

During the merge with the new code of FormUtil function, we will got a result like this: 

```php
[
  "collection" => [
    1 => [
      "checkbox" => "1",
      "files" => [
      	0 => FileObject
      ]
    ],
    2 => [
      "files" => [
      	0 => FileObject
      ]
    ],
    3 => [
      "files" => [
      	0 => FileObject
      ]
    ],
  ]
]
```
😮 

As you can see, index for root collection have changed. We lost index "0" and we have a new index "3".
Obviously, we will be in trouble here as Symfony will lost data from collection item with index 0 and will try to add a new item.

If we didn't authorize it, we will then get the error about extra fields.

**Expected result** was:
```php
[
  "collection" => [
    0 => [
      "files" => [
      	0 => FileObject
      ]
    ],
    1 => [
      "checkbox" => "1",
      "files" => [
      	0 => FileObject
      ]
    ],
    2 => [
      "files" => [
      	0 => FileObject
      ]
    ],
  ]
]
```

## Workaround ?

For my use case, it has been simple to get a workaround by just adding a hidden field in the root collection. This way, my params array is always complete with all index and merge is done successfully.

## Fix ?

I have tried a quick fix by changing `mergeParamsAndFiles` methods like this:

```diff
   public static function mergeParamsAndFiles(array $params, array $files): array
    {
        $isFilesList = array_is_list($files);

        foreach ($params as $key => $value) {
            if (\is_array($value) && \is_array($files[$key] ?? null)) {
                $params[$key] = self::mergeParamsAndFiles($value, $files[$key]);
                unset($files[$key]);
            }
        }

        if (!$isFilesList) {
            return array_replace($params, $files);
        }

-        foreach ($files as $value) {
-            $params[] = $value;
+        foreach ($files as $key => $value) {
+            $params[$key] = $value;
        }

        return $params;
    }
```

It work fine in my use case but unfortunately, it cause regressions in the test function [testMergeParamsAndFilesMultiple](https://github.com/symfony/symfony/blob/bf62cbe5db52d4832355af6f14ffd9e2e1131c9d/src/Symfony/Component/Form/Tests/AbstractRequestHandlerTestCase.php#L263).

## What to do next ?

So here is what I see for now: 
1. We could revert previous PR #52255, but it seems backtracking to me.
2. We could try to find a better way to fix this but it could be hard as it's difficult to differenciate integer index of existing items and integer index of new added items ?
3. Do nothing as it is easy to use a workaround and seems to be an edge case ?
4. Anything else we could do ?

## Who can fix it ?

With some help I could produce a fix as I already created a unit test to reproduce the issue in this PR #53353.

Is this test seems a good start point for you ?
Is this issue enough impacting to trying to create a fix ?
Do you see some more use case where this issue could occur ?

Appreciate your feedbacks 🙂

Clement-B:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  |no
| Deprecations? | no
| Issues        | Fix #53359
| License       | MIT

See ticket #53359 (WIP)

### DIFF
--- a/src/Symfony/Component/Form/Tests/AbstractRequestHandlerTestCase.php
+++ b/src/Symfony/Component/Form/Tests/AbstractRequestHandlerTestCase.php
@@ -355,6 +355,53 @@ abstract class AbstractRequestHandlerTestCase extends TestCase
         $this->assertNotNull($itemsForm->get('0')->get('file'));
     }
 
+    public function testMergePartialDataFromCollection()
+    {
+        $form = $this->createForm('root', 'POST', true);
+        $form->add('items', CollectionType::class, [
+            'entry_type' => ItemFileType::class,
+            'allow_add' => true,
+        ]);
+
+        $file = $this->getUploadedFile();
+        $file2 = $this->getUploadedFile();
+
+        $this->setRequestData('POST', [
+            'root' => [
+                'items' => [
+                    1 => [
+                        'item' => 'test',
+                    ],
+                ],
+            ],
+        ], [
+            'root' => [
+                'items' => [
+                    0 => [
+                        'file' => $file,
+                    ],
+                    1 => [
+                        'file' => $file2,
+                    ],
+                ],
+            ],
+        ]);
+
+        $this->requestHandler->handleRequest($form, $this->request);
+
+        $itemsForm = $form->get('items');
+        $data = $itemsForm->getData();
+        $this->assertTrue($form->isSubmitted());
+        $this->assertTrue($form->isValid());
+
+        $this->assertCount(2, $data);
+        $this->assertArrayHasKey(0, $data);
+        $this->assertArrayHasKey(1, $data);
+
+        $this->assertEquals('test', $itemsForm->get('1')->get('item')->getData());
+        $this->assertNotNull($itemsForm->get('0')->get('file'));
+    }
+
     /**
      * @dataProvider methodExceptGetProvider
      */


nicolas-grekas:
@priyadi is this something you'd like to check? That'd be super helpful :) :pray:

xabbuh:
closing in favor of #54324

Issue #53356, opened 2024-01-02
[Translation] Translation Debug command doesn't work correctly with domain as constant


VincentLanglet:
### Symfony version(s) affected

all

### Description

Currently both
```php
use App\Const\Domain;
$this->translator->trans('foo', Domain::CUSTOM);
```
in php and
```twig
{{ 'foo'|trans({}, constant('App\\Constant\\Domain::CUSTOM')) }};
```
in twig is misunderstood by translation extractor.

The first one thinks the domain is `'messages'`, the second one `'_undefined'`.

It would be nice to correctly support the domain.
If it's really not possible, it seems better to me to have `'_undefined'` in both case.

### How to reproduce

Running
```bash
bin/console debug:translation
```
with the previous code.

### Possible Solution

For the AbstractVisitor, I tried doing something with
```php
if ($node instanceof Node\Expr\ClassConstFetch) {
            try {
                $reflection = new \ReflectionClass($node->class->toString());
                $constant = $reflection->getReflectionConstant($node->name->toString());
                if (false !== $constant && \is_string($constant->getValue())) {
                    return $constant->getValue();
                }
            } catch (\ReflectionException) {}
        }
```
but this works only with full FQCN, and not when it's imported/aliased at the start of the file...

For twig, I tried https://github.com/symfony/symfony/pull/53357

smnandre:
> ```
> {{ 'foo'|trans({}, constant('App\Constant\Domain::CUSTOM')) }};
> ```

Probably a typo but just to be sure, you used double backslashes instead of simple ones ' in your real code ? 

```
{{ 'foo'|trans({}, constant('App\\Constant\\Domain::CUSTOM')) }};
```

VincentLanglet:
> > ```
> > {{ 'foo'|trans({}, constant('App\Constant\Domain::CUSTOM')) }};
> > ```
> 
> Probably a typo but just to be sure, you used double backslashes instead of simple ones ' in your real code ?
> 
> ```
> {{ 'foo'|trans({}, constant('App\\Constant\\Domain::CUSTOM')) }};
> ```

Yes indeed it was a typo, I edited the issue.

VincentLanglet:
@welcoMattic you introduced the PhpAstExtractor in https://github.com/symfony/symfony/pull/46161
maybe you'll know how to support the syntax:
```
use App\Const\Domain;
$this->translator->trans('foo', Domain::CUSTOM);
```
?

welcoMattic:
[...]

PR #53357, opened 2024-01-02
[Translation] Fix `TranslationNodeVisitor` with constant domain

VincentLanglet:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Issues        | Fix #... 
| License       | MIT

Related to https://github.com/symfony/symfony/issues/53356 but fixing only the twig part.

### DIFF
--- a/src/Symfony/Bridge/Twig/NodeVisitor/TranslationNodeVisitor.php
+++ b/src/Symfony/Bridge/Twig/NodeVisitor/TranslationNodeVisitor.php
@@ -158,6 +158,22 @@ final class TranslationNodeVisitor extends AbstractNodeVisitor
             return $node->getAttribute('value');
         }
 
+        if (
+            $node instanceof FunctionExpression
+            && 'constant' === $node->getAttribute('name')
+        ) {
+            $nodeArguments = $node->getNode('arguments');
+            if ($nodeArguments->getIterator()->current() instanceof ConstantExpression) {
+                $constantName = $nodeArguments->getIterator()->current()->getAttribute('value');
+                if (\defined($constantName)) {
+                    $value = \constant($constantName);
+                    if (\is_string($value)) {
+                        return $value;
+                    }
+                }
+            }
+        }
+
         return self::UNDEFINED_DOMAIN;
     }
 
--- a/src/Symfony/Bridge/Twig/Tests/Translation/TwigExtractorTest.php
+++ b/src/Symfony/Bridge/Twig/Tests/Translation/TwigExtractorTest.php
@@ -22,6 +22,8 @@ use Twig\Loader\LoaderInterface;
 
 class TwigExtractorTest extends TestCase
 {
+    public const CUSTOM_DOMAIN = 'domain';
+
     /**
      * @dataProvider getExtractData
      */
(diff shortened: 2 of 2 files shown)


VincentLanglet:
Not sure if I should fix the fail check for code I didn't touch ?

src/Symfony/Bridge/Twig/NodeVisitor/TranslationNodeVisitor.php
```diff
@@ -158,6 +158,19 @@ private function getReadDomainFromNode(Node $node): ?string
             return $node->getAttribute('value');
         }
 
+        if (
+            $node instanceof FunctionExpression
```
xabbuh, line 162: Based on the updated fixture I would have expected a `FilterExpression` here. But that makes me think that we probably need to provide the fix for the `trans` tag, `t()` function as well as the `trans` filter.

VincentLanglet: I added test and they are now supported https://github.com/symfony/symfony/pull/53357/commits/52b111aab733eccda7799f0449d247be9fc2a148

VincentLanglet:
@xabbuh Should I fix the CI ? The error (even from fabbot.io) seems unrelated to my changes, and I'm affraid about conflict during futur merge if I start fixing them...

xabbuh:
@VincentLanglet You can safely ignore things suggested by fabbot that haven't been touched by your changes.

VincentLanglet:
> @VincentLanglet You can safely ignore things suggested by fabbot that haven't been touched by your changes.

Thanks ; I think it's ready then :)

src/Symfony/Bridge/Twig/NodeVisitor/TranslationNodeVisitor.php
```diff
@@ -158,6 +158,19 @@ private function getReadDomainFromNode(Node $node): ?string
...
+            && 'constant' === $node->getAttribute('name')
+        ) {
+            $nodeArguments = $node->getNode('arguments');
+            if ($nodeArguments->getIterator()->current() instanceof ConstantExpression) {
+                $value = \constant($nodeArguments->getIterator()->current()->getAttribute('value'));
```
xabbuh: We should probably check that the constant is actually defined to prevent uncaught errors.

xabbuh (changes requested):

src/Symfony/Bridge/Twig/NodeVisitor/TranslationNodeVisitor.php
VincentLanglet: I added `if (\defined($constantName)) {` ; do you think it's enough ?

xabbuh: Will calling `defined()` trigger the autoloader in case a class constant is referenced?

VincentLanglet: I dunno how the whole defined/constant works.

All I can say is that I tried this code in my project and all the (falsy positive) missing translation are now found.

xabbuh: I tried it, looks like it's taking into account.

xabbuh (approved):

nicolas-grekas:
Thank you @VincentLanglet.

PR #53360, opened 2024-01-02
[Validator] Add the `Base64` constraint

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

Base64 encoding is commonly employed when handling file uploads through DTO. This constraint would really save some time to devs by ensuring that the Base64-encoded data representing the uploaded file content is correctly formatted and valid with a single attribute.

Because of the main use I see is file uploads in this format, I also added an option, `allowDataUri`, which allows the input to be prefixed by something like `data:image/png;base64,`.

### DIFF
--- a/src/Symfony/Component/Validator/Constraints/Base64.php
+++ b/src/Symfony/Component/Validator/Constraints/Base64.php
@@ -1,0 +1,40 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * @author Alexandre Daubois <alex.daubois@gmail.com>
+ */
+#[\Attribute(\Attribute::TARGET_PROPERTY | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
+final class Base64 extends Constraint
+{
+    public const INVALID_STRING_ERROR = '5699d0ba-3cea-4676-9aab-d3303ebd6934';
+    public const MISSING_DATA_URI_ERROR = '9380f49b-582a-49ce-a0b9-e68ff009e80f';
+
+    protected const ERROR_NAMES = [
+        self::INVALID_STRING_ERROR => 'INVALID_STRING_ERROR',
+        self::MISSING_DATA_URI_ERROR => 'MISSING_DATA_URI_ERROR',
+    ];
+
+    public string $messageInvalidString = 'The given string is not a valid Base64 encoded string.';
+    public string $messageMissingDataUri = 'The given string is missing a data URI.';
+
+    public function __construct(public bool $requiresDataUri = false, public bool $urlEncoded = false, string $messageInvalidString = null, string $messageMissingDataUri = null, array $groups = null, mixed $payload = null, array $options = null)
+    {
+        parent::__construct($options, $groups, $payload);
+
+        $this->messageInvalidString = $messageInvalidString ?? $this->messageInvalidString;
+        $this->messageMissingDataUri = $messageMissingDataUri ?? $this->messageMissingDataUri;
+    }
+}
--- a/src/Symfony/Component/Validator/Constraints/Base64Validator.php
+++ b/src/Symfony/Component/Validator/Constraints/Base64Validator.php
@@ -1,0 +1,63 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+use Symfony\Component\Validator\Exception\UnexpectedTypeException;
+use Symfony\Component\Validator\Exception\UnexpectedValueException;
+
+/**
+ * @author Alexandre Daubois <alex.daubois@gmail.com>
+ */
+final class Base64Validator extends ConstraintValidator
+{
+    public function validate(mixed $value, Constraint $constraint): void
+    {
+        if (!$constraint instanceof Base64) {
+            throw new UnexpectedTypeException($constraint, Base64::class);
+        }
+
+        if (null === $value) {
+            return;
+        }
+
+        if (!\is_string($value) && !$value instanceof \Stringable) {
+            throw new UnexpectedValueException($value, 'string');
+        }
+
+        if ($constraint->urlEncoded) {
+            $value = rawurldecode($value);
+        }
+
+        if ($constraint->requiresDataUri) {
+            preg_match('/^data:.+;base64,/', $value, $matches);
+
+            if (0 === \count($matches)) {
+                $this->context->buildViolation($constraint->messageMissingDataUri)
+                    ->setCode(Base64::MISSING_DATA_URI_ERROR)
+                    ->addViolation();
+
+                return;
+            }
+
+            $value = str_replace($matches[0], '', $value);
+        }
+
+        preg_match('/^[a-zA-Z0-9\/\r\n+]*(==)?$/', $value, $matches);
+        if (0 === \count($matches)) {
+            $this->context->buildViolation($constraint->messageInvalidString)
+                ->setCode(Base64::INVALID_STRING_ERROR)
+                ->addViolation();
+        }
+    }
+}
--- a/src/Symfony/Component/Validator/CHANGELOG.md
+++ b/src/Symfony/Component/Validator/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
  * Add `MacAddress` constraint
  * Add `list` and `associative_array` types to `Type` constraint
  * Add the `Charset` constraint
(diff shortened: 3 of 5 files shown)


src/Symfony/Component/Validator/Constraints/Base64Validator.php
```diff
@@ -0,0 +1,48 @@
...
+
+        if (!\is_string($value) && !$value instanceof \Stringable) {
+            throw new UnexpectedValueException($value, 'string');
+        }
+
+        if ($constraint->allowDataUri) {
+            $value = preg_replace('/^data:.+;base64,/', '', $value);
+        }
+
+        if (false === base64_decode($value, true)) {
```
alexandre-daubois: Ensure `false` is strictly returned because an empty string is a valid base64 string

src/Symfony/Component/Validator/Constraints/Base64.php
```diff
@@ -0,0 +1,38 @@
...
+    public const INVALID_STRING_ERROR = '5699d0ba-3cea-4676-9aab-d3303ebd6934';
+
+    protected const ERROR_NAMES = [
+        self::INVALID_STRING_ERROR => 'INVALID_STRING_ERROR',
+    ];
+
+    public bool $allowDataUri = false;
+    public string $message = 'The given string is not a valid Base64 encoded string.';
+
+    public function __construct(array|string $allowDataUri = null, string $message = null, array $groups = null, mixed $payload = null, array $options = null)
```
xabbuh:
```suggestion
    public function __construct(bool $allowDataUri = false, string $message = null, array $groups = null, mixed $payload = null, array $options = null)
```

Doesn't that work?

alexandre-daubois: Of course, my bad 🤦 Thanks for catching this

src/Symfony/Component/Validator/Constraints/Base64Validator.php
```diff
@@ -0,0 +1,48 @@
...
+        if (null === $value) {
+            return;
+        }
+
+        if (!\is_string($value) && !$value instanceof \Stringable) {
+            throw new UnexpectedValueException($value, 'string');
+        }
+
+        if ($constraint->allowDataUri) {
+            $value = preg_replace('/^data:.+;base64,/', '', $value);
```
xabbuh:
```suggestion
            $value = preg_replace('/^data:.++;base64,/', '', $value);
```

Maybe?

alexandre-daubois: I admit I don't know this quantifier and had hard times understand how this works :smile: Unfortunately, it makes  the tests fail with `data:image/png;base64,dGVzdA==` :confused:

src/Symfony/Component/Validator/Constraints/Base64Validator.php
nicolas-grekas: This way of checking consumes memory for no good reasons because it generates the decoded payload and trashes is right away.
A regexp would be better.
Also, there are two styles of base64 encoding (url-safe and not-url-safe), so we need a way to tell which one we want.

src/Symfony/Component/Validator/Constraints/Base64.php
```diff
@@ -0,0 +1,38 @@
...
+
+    public bool $allowDataUri = false;
+    public string $message = 'The given string is not a valid Base64 encoded string.';
+
+    public function __construct(bool $allowDataUri = false, string $message = null, array $groups = null, mixed $payload = null, array $options = null)
+    {
+        parent::__construct($options, $groups, $payload);
+
+        $this->message = $message ?? $this->message;
+        $this->allowDataUri = $allowDataUri;
```
nicolas-grekas: I'm not sure I understand how this is supposed to be useful.
Do you really have a use case where data URLs are accepted as input?
The main issue I see is the fact that the URL prefix is optional when enabling this: either you accept without the prefix, or you accept with the prefix, but both? really?

alexandre-daubois: About the use case, yes: we have use cases where a user uploads, let's say, a PNG. The input data looks like `data:image/png;base64,iVBORw0KGgo...`. However, we also accept, in other DTOs, "simple" base64 without data uri. I think giving this option helps the constraint being flexible for different cases devs will encounter.

That said, I agree that accepting both is useless. It is either one case or another. I'll update the code to have an option like `requiresDataUri` which makes more sense.

src/Symfony/Component/Validator/Constraints/Base64Validator.php
alexandre-daubois: I updated the check to use a regex to check base64 validity.

Also, the `urlEncoded` option has been added. Is that what you had in mind?

src/Symfony/Component/Validator/Constraints/Base64.php
```diff
@@ -0,0 +1,40 @@
...
+
+    protected const ERROR_NAMES = [
+        self::INVALID_STRING_ERROR => 'INVALID_STRING_ERROR',
+        self::MISSING_DATA_URI_ERROR => 'MISSING_DATA_URI_ERROR',
+    ];
+
+    public string $messageInvalidString = 'The given string is not a valid Base64 encoded string.';
+    public string $messageMissingDataUri = 'The given string is missing a data URI.';
+
+    public function __construct(public bool $requiresDataUri = false, public bool $urlEncoded = false, string $messageInvalidString = null, string $messageMissingDataUri = null, array $groups = null, mixed $payload = null, array $options = null)
```
nicolas-grekas, line 33: CPP mean parameters should be on their own lines

Also:
```suggestion
    public function __construct(public bool $dataUri = false, public bool $urlSafe = false, string $message = null, string $messageMissingDataUri = null, array $groups = null, mixed $payload = null, array $options = null)
```

src/Symfony/Component/Validator/Constraints/Base64.php
```diff
@@ -0,0 +1,40 @@
...
+    public const INVALID_STRING_ERROR = '5699d0ba-3cea-4676-9aab-d3303ebd6934';
+    public const MISSING_DATA_URI_ERROR = '9380f49b-582a-49ce-a0b9-e68ff009e80f';
+
+    protected const ERROR_NAMES = [
+        self::INVALID_STRING_ERROR => 'INVALID_STRING_ERROR',
+        self::MISSING_DATA_URI_ERROR => 'MISSING_DATA_URI_ERROR',
+    ];
+
+    public string $messageInvalidString = 'The given string is not a valid Base64 encoded string.';
+    public string $messageMissingDataUri = 'The given string is missing a data URI.';
```
nicolas-grekas, line 31:
```suggestion
    public string $messageInvalidString = 'This value is not a valid Base64 encoded string.';
    public string $messageMissingDataUri = 'This value is missing a data URI prefix.';
```

src/Symfony/Component/Validator/Constraints/Base64Validator.php
```diff
@@ -0,0 +1,63 @@
...
+        if (null === $value) {
+            return;
+        }
+
+        if (!\is_string($value) && !$value instanceof \Stringable) {
+            throw new UnexpectedValueException($value, 'string');
+        }
+
+        if ($constraint->urlEncoded) {
+            $value = rawurldecode($value);
```
nicolas-grekas, line 39: Base64 URL safe encoding is not that. It's about using `-_` instead of `+/`.

src/Symfony/Component/Validator/Constraints/Base64Validator.php
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        if ($constraint->requiresDataUri) {
+            preg_match('/^data:.+;base64,/', $value, $matches);
+
+            if (0 === \count($matches)) {
```
nicolas-grekas, line 45: why not use `if (!preg_match)`?

src/Symfony/Component/Validator/Constraints/Base64Validator.php
```diff
@@ -0,0 +1,63 @@
...
+                    ->addViolation();
+
+                return;
+            }
+
+            $value = str_replace($matches[0], '', $value);
+        }
+
+        preg_match('/^[a-zA-Z0-9\/\r\n+]*(==)?$/', $value, $matches);
+        if (0 === \count($matches)) {
```
nicolas-grekas, line 57: same question

src/Symfony/Component/Validator/Constraints/Base64Validator.php
```diff
@@ -0,0 +1,63 @@
...
+        if (!\is_string($value) && !$value instanceof \Stringable) {
+            throw new UnexpectedValueException($value, 'string');
+        }
+
+        if ($constraint->urlEncoded) {
+            $value = rawurldecode($value);
+        }
+
+        if ($constraint->requiresDataUri) {
+            preg_match('/^data:.+;base64,/', $value, $matches);
```
nicolas-grekas, line 43: Since this is a validator, it'd be great to also validate what is just after `data:`. I don't know the rules, but google or gpt will.

src/Symfony/Component/Validator/Constraints/Base64.php
nicolas-grekas: BTW, the message/naming about data URI is misleading: the prefix could be there but be invalid.

alexandre-daubois:
After discussing with Nicolas, we agreed this constraint doesn't make really sense. Thus, this can still be done by using a regex.

For anyone wondering how to achieve this, use the `Regex` constraint with the following (or similar): `/^[a-zA-Z0-9\/\r\n+]*(==)?$/`.

PR #53362, opened 2024-01-02
[PropertyInfo] Restrict access to `PhpStanExtractor` based on visibility

nikophil:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

Hi,

this small PR gives a bit more control on what `PhpStanExtractor` can see or not, based on the visibility of the property or its accessor.

It's pretty unusual, but sometimes, the getter does not return the same type than the property itself, and before this PR, there was actually no way to extract the type returned by the getter. `ReflectionExtractor` has this granularity, but it cannot read complex PhpDoc types.

### DIFF
--- a/src/Symfony/Component/PropertyInfo/Extractor/PhpStanExtractor.php
+++ b/src/Symfony/Component/PropertyInfo/Extractor/PhpStanExtractor.php
@@ -53,7 +53,7 @@ final class PhpStanExtractor implements PropertyTypeExtractorInterface, Construc
      * @param list<string>|null $accessorPrefixes
      * @param list<string>|null $arrayMutatorPrefixes
      */
-    public function __construct(?array $mutatorPrefixes = null, ?array $accessorPrefixes = null, ?array $arrayMutatorPrefixes = null)
+    public function __construct(?array $mutatorPrefixes = null, ?array $accessorPrefixes = null, ?array $arrayMutatorPrefixes = null, private bool $allowPrivateAccess = true)
     {
         if (!class_exists(ContextFactory::class)) {
             throw new \LogicException(sprintf('Unable to use the "%s" class as the "phpdocumentor/type-resolver" package is not installed. Try running composer require "phpdocumentor/type-resolver".', __CLASS__));
@@ -232,6 +232,10 @@ final class PhpStanExtractor implements PropertyTypeExtractorInterface, Construc
             return null;
         }
 
+        if (!$this->canAccessMemberBasedOnItsVisibility($reflectionProperty)) {
+            return null;
+        }
+
         // Type can be inside property docblock as `@var`
         $rawDocNode = $reflectionProperty->getDocComment();
         $phpDocNode = $rawDocNode ? $this->getPhpDocNode($rawDocNode) : null;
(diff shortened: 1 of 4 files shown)


src/Symfony/Component/PropertyInfo/Extractor/PhpStanExtractor.php
```diff
@@ -47,13 +47,14 @@ final class PhpStanExtractor implements PropertyTypeExtractorInterface, Construc
...
     private array $arrayMutatorPrefixes;
+    private bool $allowPrivateAccess;
 
     /**
      * @param list<string>|null $mutatorPrefixes
      * @param list<string>|null $accessorPrefixes
      * @param list<string>|null $arrayMutatorPrefixes
      */
-    public function __construct(array $mutatorPrefixes = null, array $accessorPrefixes = null, array $arrayMutatorPrefixes = null)
+    public function __construct(array $mutatorPrefixes = null, array $accessorPrefixes = null, array $arrayMutatorPrefixes = null, bool $allowPrivateAccess = true)
```
OskarStark: you can Use CPP here please

nikophil: done. Only for this new parameter, since the other ones need some kind of initialization

Korbeil (approved):
It looks good to me :+1:

src/Symfony/Component/PropertyInfo/Extractor/PhpStanExtractor.php
```diff
@@ -305,4 +312,9 @@ private function getPhpDocNode(string $rawDocNode): PhpDocNode
 
         return $phpDocNode;
     }
+
+    public function canAccessMemberBasedOnItsVisibility(\ReflectionProperty|\ReflectionMethod $member): bool
+    {
+        return $member->isPublic() || $this->allowPrivateAccess;
```
dunglas:
```suggestion
        return $this->allowPrivateAccess || $member->isPublic();
```


Same, prevents a useless function call if `allowPrivateAccess` is true.

src/Symfony/Component/PropertyInfo/Extractor/PhpStanExtractor.php
```diff
@@ -274,8 +278,11 @@ private function getDocBlockFromMethod(string $class, string $ucFirstProperty, i
                 }
 
                 if (
-                    (self::ACCESSOR === $type && 0 === $reflectionMethod->getNumberOfRequiredParameters())
-                    || (self::MUTATOR === $type && $reflectionMethod->getNumberOfParameters() >= 1)
+                    $this->canAccessMemberBasedOnItsVisibility($reflectionMethod)
+                    && (
+                        (self::ACCESSOR === $type && 0 === $reflectionMethod->getNumberOfRequiredParameters())
+                        || (self::MUTATOR === $type && $reflectionMethod->getNumberOfParameters() >= 1)
+                    )
```
dunglas:
```suggestion
                    (
                        (self::ACCESSOR === $type && 0 === $reflectionMethod->getNumberOfRequiredParameters())
                        || (self::MUTATOR === $type && $reflectionMethod->getNumberOfParameters() >= 1)
                    )
                    && $this->canAccessMemberBasedOnItsVisibility($reflectionMethod)
```

To prevent a useless function call in the happy path.

dunglas (commented):
Overall, LGTM!

src/Symfony/Component/PropertyInfo/Tests/Fixtures/DummyPropertyAndGetterWithDifferentTypes.php
```diff
@@ -0,0 +1,25 @@
...
+
+final readonly class DummyPropertyAndGetterWithDifferentTypes
+{
+    public function __construct(
+        /**
+         * @var string
+         */
+        private string $foo
+    )
+    {
```
OskarStark:
```suggestion
    ) {
```

fabpot (approved):

fabpot:
Thank you @nikophil.

PR #46218, opened 2022-04-29
Fix docblock types of Command::addOption()

mostafa-re:
| Q             | A
| ------------- | ---
| Branch?       | 6.0
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #46140
| License       | MIT
| Doc PR        | -

PR #46217, opened 2022-04-29
Fix docblock of two Command methods

mostafa-re:
| Q             | A
| ------------- | ---
| Branch?       | 6.1
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Tickets       | Fix #46140
| License       | MIT
| Doc PR        | -

Issue #46140, opened 2022-04-21
[Console] Docblocks in 6.1 are wrongly formatted


Zarthus:
### Symfony version(s) affected

6.1, 6.0

### Description

https://github.com/symfony/console/blob/6.1/Command/Command.php#L434-L479
https://github.com/symfony/console/blob/6.0/Command/Command.php#L431-L450

In PhpStorm the following docblock is picked up as "The" class rather than, for instance in $mode, a `nullable int`

### How to reproduce

1. in PHPStorm with symfony/console >= 6.1
2. Try the following code snippet
```php
class SfCommand extends \Symfony\Component\Console\Command\Command
{
    protected function configure()
    {
        $this->addOption('foo', 'bar', 'baz', 'desc', 'default',);
    }
}
```

https://github.com/symfony/console/blob/6.1/Command/Command.php#L434-L479

### Possible Solution

The way it looks in 5.4 is fine:

https://github.com/symfony/console/blob/5.4/Command/Command.php#L446-L465

### Additional Context

![image](https://user-images.githubusercontent.com/3660408/164502773-68edc760-e39c-458b-a127-8d9ebd3ce220.png)

chalasr:
Closing for the reason given in https://github.com/symfony/symfony/pull/46218#issuecomment-1113670440 (tldr: not a Symfony bug but a PHPStorm one). Thanks

PR #45324, opened 2022-02-05
[Cache] Mention parameter type in phpdoc explicitly

andrew-demb:
| Q             | A
| ------------- | ---
| Branch?       | 4.4
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Tickets       | 
| License       | MIT
| Doc PR        |

PR #53363, opened 2024-01-02
[Console] CS fix

OskarStark:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | --
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Console/Helper/ProgressIndicator.php
+++ b/src/Symfony/Component/Console/Helper/ProgressIndicator.php
@@ -129,8 +129,6 @@ class ProgressIndicator
 
     /**
      * Finish the indicator with message.
-     *
-     * @param $message
      */
     public function finish(string $message)
     {


keradus (approved):

src/Symfony/Component/Console/Completion/CompletionInput.php
```diff
@@ -53,7 +53,7 @@ public static function fromString(string $inputStr, int $currentIndex): self
      * Create an input based on an COMP_WORDS token list.
      *
      * @param string[] $tokens       the set of split tokens (e.g. COMP_WORDS or argv)
-     * @param          $currentIndex the index of the cursor (e.g. COMP_CWORD)
+     * @param int      $currentIndex the index of the cursor (e.g. COMP_CWORD)
```
wouterj: This is intentional. The type is optional and already given in de PHP code. We've done this in multiple locations (and I would even be in favor of removing all duplicate types from the PHPdoc)

wouterj (changes requested):

src/Symfony/Component/Console/Completion/CompletionInput.php
OskarStark: PHPStorm was complaining about it:
CleanShot 2024-01-03 at 14 43 40@2x

OskarStark: I think we should be content, in 5.4 only two more needs to be fixed, which I've done in this PR.
I also created a PR for 6.3
* #53391 

If both are accepted and (up)merged, I will check 6.4, 7.0 and 7.1

wouterj: To me, that's a bug in PHPstorm.

chalasr: @wouterj Do you have a link to a discussion about this? 
I consider such doc param to be incomplete and less readable: when reading a param's description, not having the type right before it requires to find and look at the corresponding parameter inside the signature to get the full picture

src/Symfony/Component/Cache/Adapter/PhpFilesAdapter.php
```diff
@@ -92,9 +92,6 @@ public function prune()
         return $pruned;
     }
 
-    /**
-     * {@inheritdoc}
-     */
```
OskarStark: Those "inheritdoc" got removed by applying the fabbot patch, shall I revert?

src/Symfony/Component/Console/Completion/CompletionInput.php
```diff
@@ -34,7 +34,7 @@ final class CompletionInput extends ArgvInput
     private $tokens;
     private $currentIndex;
     private $completionType;
-    private $completionName = null;
+    private $completionName;
```
OskarStark:
```suggestion
    private $completionName = null;
```

src/Symfony/Component/Console/Helper/ProgressIndicator.php
```diff
@@ -129,8 +129,6 @@ public function advance()
 
     /**
      * Finish the indicator with message.
-     *
-     * @param $message
```
OskarStark, line 133: Useless

src/Symfony/Component/Console/Completion/CompletionInput.php
wouterj: Sure, it has always been part of our coding standard:

> - Add PHPDoc blocks for classes, methods, and functions only when they add relevant information that does not duplicate the name, native type declaration or context (e.g. instanceof checks);

Discussed in the PR making the change: https://github.com/symfony/symfony/pull/41912#pullrequestreview-696453567

See also https://github.com/symfony/symfony/issues/46140 https://github.com/symfony/symfony/pull/46218 https://github.com/symfony/symfony/pull/46217 https://github.com/symfony/symfony/pull/44935 https://github.com/symfony/symfony/pull/45324 (a couple of these are rejected by you actually Robin 😉)

src/Symfony/Component/Cache/Adapter/PhpFilesAdapter.php
keradus: any value to keep it? i would suggest to remove all.
and if anything, I would suggest to switch to `#[Override]` (probably as alignment for whole project, not this single place only)

src/Symfony/Component/Console/Completion/CompletionInput.php
chalasr: Fair enough :)

src/Symfony/Component/Console/Helper/ProgressIndicator.php
nicolas-grekas: I think that's the only change we might wanna keep. See @wouterj's comment.
Also 5.4 is for bug fixes only, and this is not one.

OskarStark: Done

nicolas-grekas:
Thank you @OskarStark.

Issue #53364, opened 2024-01-02
 Parent call is missing in deprecated FormEvent::setData() for events that do not allow it 


anyt:
### Symfony version(s) affected

6.4.0

### Description

Hi everyone,
As I've spent hours to get into the fact that [the method](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Form/Event/PostSubmitEvent.php#L27) now does literally nothing, while the business logic expects it to update data as it was in 5.4, I think it worth sharing.

Most likely it should trigger a deprecation and call the parent method. Otherwise it's a logical backward compatibility break that is hard to find because it only triggers a deprecation and does nothing.

From my perspective the deprecation is about using an obsolete way to do the thing, while here the method just does nothing and I have no choice except using the new approach to make it work back.

### How to reproduce

Subscribe to FormEvents::POST_SUBMIT and call $event->setData($data); 

### Possible Solution

add `parent::setData($data);` after deprecation trigger to 6.4 version.

PR #53381, opened 2024-01-03
[Form] Fix assigning data in `PostSetDataEvent` and `PostSubmitEvent`

fancyweb:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | https://github.com/symfony/symfony/issues/53364
| License       | MIT

Ref https://github.com/symfony/symfony/pull/51043 where we changed the behavior on 6.4 since we don't assign the data property anymore.

cc @HeahDude

### DIFF
--- a/src/Symfony/Component/Form/Event/PostSetDataEvent.php
+++ b/src/Symfony/Component/Form/Event/PostSetDataEvent.php
@@ -11,7 +11,6 @@
 
 namespace Symfony\Component\Form\Event;
 
-use Symfony\Component\Form\Exception\BadMethodCallException;
 use Symfony\Component\Form\FormEvent;
 
 /**
@@ -29,5 +28,6 @@ final class PostSetDataEvent extends FormEvent
     {
         trigger_deprecation('symfony/form', '6.4', 'Calling "%s()" will throw an exception as of 7.0, listen to "form.pre_set_data" instead.', __METHOD__);
         // throw new BadMethodCallException('Form data cannot be changed during "form.post_set_data", you should use "form.pre_set_data" instead.');
+        parent::setData($data);
     }
 }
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/Form/Event/PostSetDataEvent.php
```diff
@@ -11,7 +11,6 @@
 
 namespace Symfony\Component\Form\Event;
 
-use Symfony\Component\Form\Exception\BadMethodCallException;
```
xabbuh, line 14: should IMO be added back

xabbuh (approved):

src/Symfony/Component/Form/Event/PostSetDataEvent.php
fancyweb: It's not present in PostSubmitEvent.

OskarStark: > It's not present in PostSubmitEvent.

No, but this is about `PostSetDataEvent` and it is in the comment for the next major:

```php
// throw new BadMethodCallException('Form data cannot be changed during "form.post_set_data", you should use "form.pre_set_data" instead.');
```

fancyweb: What I mean is that there's the same code in PostSubmitEvent without the import :sweat_smile:

OskarStark: Oh I would add the import then 👍

nicolas-grekas: I'd follow what fabbot wants on this one, because this will come up again otherwise.
But we should then be careful when merging up to not remove the `use` in 7.0

xabbuh:
Thank you @fancyweb.

PR #53388, opened 2024-01-03
[Mailer] [Azure] Fix resource name validation

### LINKED ISSUE #53385: Incorrect validation for Azure Email Communication Resource Name

7.1

### Description

Generally Azure Email Communication Resource Name might contain the dots.

There is a validation in **AzureApiTransport** class constructor that throws exception when Azure Email Communication Resource contains dot in it.

```php
if (str_contains($resourceName, '.') || str_ends_with($resourceName, '.')) {
    throw new \Exception('Resource name cannot contain or end with a dot.');
  }

```

### How to reproduce

Try to use a Azure Email Communication Resource Name which contains dot.

### Possible Solution

Remove the validation which forbids the use of Azure Email Communication Resource Name with dot in it.

### Additional Context

_No response_

t9toqwerty:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #53385
| License       | MIT

Remove validation where Azure Email Communication Resource with dot in name is not being allowed to use.

### DIFF
--- a/src/Symfony/Component/Mailer/Bridge/Azure/Tests/Transport/AzureApiTransportTest.php
+++ b/src/Symfony/Component/Mailer/Bridge/Azure/Tests/Transport/AzureApiTransportTest.php
@@ -121,7 +121,7 @@ class AzureApiTransportTest extends TestCase
     public function testItDoesNotAllowToAddResourceNameWithDot()
     {
         $this->expectException(\Exception::class);
-        $this->expectExceptionMessage('Resource name cannot contain or end with a dot');
+        $this->expectExceptionMessage('Resource name must not end with a dot "."');
 
         new AzureApiTransport('KEY', 'ACS_RESOURCE_NAME.');
     }
--- a/src/Symfony/Component/Mailer/Bridge/Azure/Transport/AzureApiTransport.php
+++ b/src/Symfony/Component/Mailer/Bridge/Azure/Transport/AzureApiTransport.php
@@ -43,8 +43,8 @@ final class AzureApiTransport extends AbstractApiTransport
         EventDispatcherInterface $dispatcher = null,
         LoggerInterface $logger = null,
     ) {
-        if (str_contains($resourceName, '.') || str_ends_with($resourceName, '.')) {
-            throw new \Exception('Resource name cannot contain or end with a dot.');
+        if (str_ends_with($resourceName, '.')) {
+            throw new \Exception('Resource name must not end with a dot ".".');
         }
 
         parent::__construct($client, $dispatcher, $logger);


src/Symfony/Component/Mailer/Bridge/Azure/Tests/Transport/AzureApiTransportTest.php
```diff
@@ -121,7 +121,7 @@ public function testTagAndMetadataHeaders()
     public function testItDoesNotAllowToAddResourceNameWithDot()
     {
         $this->expectException(\Exception::class);
-        $this->expectExceptionMessage('Resource name cannot contain or end with a dot');
+        $this->expectExceptionMessage('Resource name cannot end with a dot.');
```
OskarStark: Lets reword it to:
```suggestion
        $this->expectExceptionMessage('Resource name must not end with a dot "."');
```

OskarStark:
Thank you for your contribution, can you maybe show your anonymized resource name or link to the corresponding Azure documentation? Thanks

cc @hafael as you contributed the bridge

t9toqwerty:
Hi @OskarStark ,

**I have seen this for the new Azure Communication Resources naming where they are appending the Data Location in resource endpoint url.** 

For Example I have created two Azure Communication Resources (One in US and other In India) and seen the pattern repeated. Check the endpoint in the screenshots:

**US:**
Screenshot 2024-01-04 at 7 50 47 AM

**India:**
Screenshot 2024-01-04 at 7 51 55 AM


The same format for endpoint can be found in the last image attached in their [documentation](https://learn.microsoft.com/en-us/azure/communication-services/quickstarts/create-communication-resource?tabs=windows&pivots=platform-azp#access-your-connection-strings-and-service-endpoints).




IMHO, The way we are formatting the host/endpoint, 
```
private const HOST = '%s.communication.azure.com';
```

We should allow the resource with dots in name so the we can get the correct host/endpoint in these scenarios .

OskarStark:
Thanks for the explanation.

From what I can see in your screenshots, we must change the following too, or does the bride works for you, when removing the "contain dot" part?

```diff
- private const HOST = '%s.communication.azure.com';
+ private const HOST = '%s.communications.azure.com';
```

t9toqwerty:
**I have just tested again and It is working fine for me. I am receiving email without any issue so I think we don't need to change the current host url structure.**

As the same url is reflecting in my dashboard too that I have attached the first two screenshot of, I assume the documentation from I attached the third image is old.

**I think we can go with current url structure i.e. . And the same is being used in their [Java SDK](https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/communication/azure-communication-email/README.md) too.**
```
private const HOST = '%s.communication.azure.com';
```

Just to be on safe side I have tried the other url (`private const HOST = '%s.communications.azure.com';`) too but got `TransportException: Could not resolve host` Exception.

Screenshot 2024-01-04 at 12 50 01 PM

webflo:
👍  This works as expected. communication.azure.com is correct.

OskarStark (approved):

OskarStark:
Thanks for fixing this bug @t9toqwerty.

Issue #51604, opened 2023-09-09
[Messenger] Generated redis transport `consumer` name


danielburger1337:
### Description

As you may now, the redis transport requires that each worker needs to have a unique connection consumer name. The suggested and very easy solution is to use the `HOSTNAME` env variable. This works perfectly in production where we use containerization via Docker, but there is a small issue when developing locally and not inside a devcontainer:

When developing locally in my WSL instance, for some reason the `HOSTNAME` env variable is empty when I try to use it via the `'%env(HOSTNAME)%'` notation in the Yaml configuration, even though the env variable does exist. 

A tiny quality of life improvement would be to allow setting the consumer name to a generated random value. 
AFAIK the consumer names dont have to be stable, so this would simplify configuration quite a bit when having multiple workers work on a single queue or multiple queues.

### Example

```yaml
# config/packages/messenger.yaml
framework:
    messenger:
        transports:
            # https://symfony.com/doc/current/messenger.html#transport-configuration
            async:
                dsn: '%env(MESSENGER_TRANSPORT_DSN)%'
                options:
                    consumer: _RANDOM
```

and then

```php
// Symfony\Component\Messenger\Bridge\Redis\Transport\Connection

public function __construct(array $options, \Redis|Relay|\RedisCluster $redis = null)
{
    // ...
    $this->consumer = $options['consumer'] === '_RANDOM' ? bin2hex(random_bytes(8)) : $options['consumer'];
    // ...
}
```

nicolas-grekas:
Closing along with #53528.

The empty `HOSTNAME` you saw locally is not a Symfony issue: bash sets `HOSTNAME` as a shell variable and does not export it, so it is absent from the process environment and `getenv()` cannot see it. Docker and Kubernetes set it in the environment, which is why the same config works in production. Reproduced here on WSL, `env | grep HOSTNAME` is empty while `echo $HOSTNAME` prints.

For local development, set the name in `.env.local`, or give it a fallback with the `default` processor:

```yaml
consumer: '%env(default:messenger_consumer_default:HOSTNAME)%'
```

On generated names: they would work, but stable ones are better. A new random name on every restart leaves the previous consumer in the Redis group holding its pending entries until `redeliver_timeout` claims them, and the consumer list keeps growing. That is why the docs suggest a `StatefulSet` for stable names on Kubernetes.

Issue #49026, opened 2023-01-18
[Messenger] Batch handlers are not flushed on worker stop


AdamKatzDev:
### Symfony version(s) affected

5.4, 6.3

### Description

The problem is that worker can't flush batch handlers without feeding a message into them.
Not much to write here without diving deep into details, see the sections below.
This bug can cause multiple hard to track issues in other parts of the system, that depends on the way the queue is processed.

### How to reproduce

The bug can be reproduced with any batch handler that does not flush itself without `shouldFlush` being `true`.

### Possible Solution

**TLDR:** to fix write a listener that flushes batch handlers and calls `Worker::ack` method on every `WorkerRunningEvent`(not required, see the explanation below)/`WorkerStoppedEvent` event.
Use a hack like https://ocramius.github.io/blog/accessing-private-php-class-members-without-reflection/ to access `Worker::ack`.

Currently the `Worker` class waits for new messages to flush handlers. In general it makes sense, there is no reason to flush batches if there are no new messages since the simplest `shouldFlush` method that checks current batch size won't change.
But the class also doesn't flush the handlers on stop. And that is an issue.

But there is another problem with the current approach.
For example if we have a handler with the following `shouldFlush` method:
```
private function shouldFlush(): bool
{
    return $this->batchSize <= \count($this->jobs) || random_int(0, 1) /* some external condition, e.g. timer */;
}
```
The handler implements `shouldFlush` method that depends on an external condition that could trigger flush. This might be useful for jobs that are processed in real time. In the current worker implementation the batch might be hanging in memory for a long time if the rate at which the processed messages are generated is low.

To fix the bug and also to enable the previously described use case a listener can be written that flushes batch handlers and calls `Worker::ack` method on every `WorkerRunningEvent`/`WorkerStoppedEvent` event. Alas, this is a hack and not a proper solution that might be integrated into the Messenger component.

AdamKatzDev:
I've provided a workaround for this issue [here](https://github.com/symfony/symfony/discussions/46869#discussioncomment-6164399). But as I mentioned this is not something that could be contributed as a PR.

carsonbot:
Hey, thanks for your report!
There has not been a lot of activity here for a while. Is this bug still relevant? Have you managed to find a workaround?

carsonbot:
Friendly reminder that this issue exists. If I don't hear anything I'll close this.

carsonbot:
Hey,

[...]

PR #53394, opened 2024-01-03
[Messenger] fixes for Redis transport

### LINKED ISSUE #44400: [Messenger] Could not acknowledge redis message with BatchHandlerInterface

5.4.0, 6.0.0

### How to reproduce

I have MessengerHandler which makes http requests (_**with only one parallel consumer**_):

messenger.yaml:
```
transports:
  test:
    dsn: '%env(MESSENGER_TRANSPORT_DSN)%/test'
    options:
      delete_after_ack: true
      consumer: '%env(MESSENGER_CONSUMER_NAME)%'
    retry_strategy:
      max_retries: 0

routing:
     'TestMessage': test
```

TestMessageHandler.php:
```
<?php
class TestMessageHandler implements MessageHandlerInterface, BatchHandlerInterface
{
    use BatchHandlerTrait;

    public function __construct(
        private HttpClientInterface $client
    ) { }

    public function __invoke(TestMessage $message, Acknowledger &$ack = null)
    {        
        return $this->handle($message, $ack);
    }
    
    private function shouldFlush(): bool
    {
        return 5 <= \count($this->jobs);
    }

    private function process(array $jobs): void
    {
        $responses = [];
        
        foreach ($jobs as [$job, $ack]) {          
            try {
                [$headers, $content] = prepareRequest();
                
                $responses[] = $this->client->request('POST', $job->getEndpoint(), [
                    'headers' => $headers,
                    'body' => $content
                ]);

                $ack->ack($job);
            } catch (\Exception $e) {
                $ack->nack($e);
            }
        }
        
        if(0 === count($responses)) {
          return;
        }

        foreach ($this->client->stream($responses) as $response => $chunk) {
            if ($chunk->isFirst()) {
                var_dump($response->getStatusCode());
            } else if ($chunk->isLast()) {
            }
        }
    }
}
```

cli:

```MESSENGER_CONSUMER_NAME=test php bin/console messenger:consume test```

Code works okay, but when i hit ctrl+c before script finished and start it again, it throws me an error:

```
In Connection.php line 441:

  Could not acknowledge redis message "1638374074049-0".
```

I noticed that after consumer crashes or i hit ctrl+c, same message comes to __invoke a few times. When removing BatchHandlerInterface it starts to work as excepected without any issue.

### Possible Solution 1 (a bad one)

1) Add uuid to every Message
2) Destroy Acknowledger and create replacement for it (HandleMessageMiddleware.php:88)

Test.php:
```
class Test
{
  public function isAcknowledged()
  {
    return false;
  }
}
```

TestMessageHandler.php:
```
public function __invoke(SendPushMessage $message, Acknowledger &$ack = null)
{
    $uid = $message->getUuid();
    
    if(isset($this->jobs[$uid])) {
      $this->flush(true);
      
      try {
        $ack = null;
        unset($ack);
      } catch(\Exception) { }
      
      $ack = new Test();

      return 0;
    }
    
    return $this->handle($message, $ack);
}

private function handle(object $message, ?Acknowledger $ack)
{
    $uid = $message->getUuid();

    if (null === $ack) {
        $ack = new Acknowledger(get_debug_type($this));
        $this->jobs[$uid] = [$message, $ack];
        $this->flush(true);

        return $ack->getResult();
    }

    $this->jobs[$uid] = [$message, $ack];
    if (!$this->shouldFlush()) {
        return \count($this->jobs);
    }

    $this->flush(true);

    return 0;
}
```

### Possible Solution 2

Comment __destruct in Symfony\Component\Messenger\Handler\Acknowledger.php

Handler.php:

```
public function __invoke(SendPushMessage $message, Acknowledger $ack = null)
{
    $uid = $message->getUuid();
    
    if(isset($this->jobs[$uid])) {
      $this->flush(true);
    
      return 0;
    }
    
    return $this->handle($message, $ack);
}

private function handle(object $message, ?Acknowledger $ack)
{
    $uid = $message->getUuid();

    if (null === $ack) {
        $ack = new Acknowledger(get_debug_type($this));
        $this->jobs[$uid] = [$message, $ack];
        $this->flush(true);

        return $ack->getResult();
    }

    $this->jobs[$uid] = [$message, $ack];
    if (!$this->shouldFlush()) {
        return \count($this->jobs);
    }

    $this->flush(true);

    return 0;
}
```

### Additional Context

_No response_

### LINKED ISSUE #49023: [Messenger][Redis] Transport can claim only single message per claim_interval

5.4, 6.3

### Description

https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Messenger/Bridge/Redis/Transport/Connection.php

After a message is claimed `nextClaim` field is reseted:
https://github.com/symfony/symfony/blob/1f7bc1098f0b03181095d35b11fa48523864e6f2/src/Symfony/Component/Messenger/Bridge/Redis/Transport/Connection.php#L285-L303
which does not allow to claim next messages:
https://github.com/symfony/symfony/blob/1f7bc1098f0b03181095d35b11fa48523864e6f2/src/Symfony/Component/Messenger/Bridge/Redis/Transport/Connection.php#L337-L339

This logic might be intended and works in many cases but it doesn't work well for batch workers with large batch size or large `claim_interval` setting. It can potentially cause messages to pile up in worst cases or to be processed far too late.

### How to reproduce

Possible to reproduce this by SIGKILL'ing a batch worker with messages stored. I hope that my explanation is good enough and a reproducer is not needed.

### Possible Solution

Should be possible to fix this by moving `nextClaim` update to a different place. The field should not be updated if there is a pending message with the same consumer (already does) or a message from another consumer was claimed (this bug).

### Additional Context

_No response_

AdamKatzDev:
| Q             | A
| ------------- | ---
| Branch?       | 7.1 
| Bug fix?      | yes
| New feature?  | yes
| Deprecations? | no 
| Issues        | Fix #49023, mostly fixes #44400. 
| License       | MIT

New minor feature: separate `pending_interval` for pending messages.

The PR overhauls Redis receiver. Fetch pending messages step is split into getting claimed messages and pending messages. 
Claim step uses XAUTOCLAIM to claim messages instead of using XPENDING and XCLAIM combo. That simplifies code a lot. This step can be rewritten without XAUTOCLAIM to support older Redis versions.
Fetching pending messages is implemented using `WeakMap`, when `WeakMap` is empty, i.e. there are no messages in the worker currently, we just iterate all consumer pending messages using XREADGROUP starting from 0. 

That approach resolves https://github.com/symfony/symfony/issues/49023 and https://github.com/symfony/symfony/issues/44400 part that is specifically caused by using batch handlers.

This PR does not fix other important issues:
- https://github.com/symfony/symfony/issues/51604 - causes `Could not acknowledge redis message` errors when multiple workers with the same consumer are running
- https://github.com/symfony/symfony/issues/49026 - prevents batch processing on worker stop, also does not allow to use batch handlers in real-time applications (details can be found in the issue description).

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Redis/Transport/Connection.php
+++ b/src/Symfony/Component/Messenger/Bridge/Redis/Transport/Connection.php
@@ -64,10 +64,13 @@ class Connection
     private int $maxEntries;
     private int $redeliverTimeout;
     private float $nextClaim = 0.0;
+    private float $nextPending = 0.0;
     private float $claimInterval;
+    private float $pendingInterval;
     private bool $deleteAfterAck;
     private bool $deleteAfterReject;
-    private bool $couldHavePendingMessages = true;
+    private ?string $lastPendingMessageId = null;
+    private ?string $lastClaimMessageId   = null;
 
     public function __construct(array $options, \Redis|Relay|\RedisCluster $redis = null)
     {
@@ -165,6 +168,7 @@ class Connection
         $this->deleteAfterReject = $options['delete_after_reject'];
         $this->redeliverTimeout = $options['redeliver_timeout'] * 1000;
         $this->claimInterval = $options['claim_interval'] / 1000;
+        $this->pendingInterval   = $options['pending_interval'] / 1000;
     }
 
     /**
@@ -331,54 +335,8 @@ class Connection
         return $params;
     }
 
-    private function claimOldPendingMessages(): void
+    private function handleDelayedMessages(): void
     {
-        try {
-            // This could soon be optimized with https://github.com/antirez/redis/issues/5212 or
-            // https://github.com/antirez/redis/issues/6256
-            $pendingMessages = $this->getRedis()->xpending($this->stream, $this->group, '-', '+', 1);
-        } catch (\RedisException|\Relay\Exception $e) {
-            throw new TransportException($e->getMessage(), 0, $e);
-        }
-
-        $claimableIds = [];
-        foreach ($pendingMessages as $pendingMessage) {
-            if ($pendingMessage[1] === $this->consumer) {
-                $this->couldHavePendingMessages = true;
-
-                return;
-            }
-
-            if ($pendingMessage[2] >= $this->redeliverTimeout) {
-                $claimableIds[] = $pendingMessage[0];
-            }
-        }
-
-        if (\count($claimableIds) > 0) {
-            try {
-                $this->getRedis()->xclaim(
-                    $this->stream,
-                    $this->group,
-                    $this->consumer,
-                    $this->redeliverTimeout,
-                    $claimableIds,
-                    ['JUSTID']
-                );
-
-                $this->couldHavePendingMessages = true;
-            } catch (\RedisException|\Relay\Exception $e) {
-                throw new TransportException($e->getMessage(), 0, $e);
-            }
-        }
-
-        $this->nextClaim = microtime(true) + $this->claimInterval;
-    }
-
-    public function get(): ?array
-    {
-        if ($this->autoSetup) {
-            $this->setup();
-        }
         $now = microtime();
         $now = substr($now, 11).substr($now, 2, 3);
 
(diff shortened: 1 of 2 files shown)


AdamKatzDev:
I'll keep this as a draft, I am not planning to finish this, sadly.

chalasr:
No worries, the patch is now available to anyone wiling to take over even if the PR is closed. Thanks for that @AdamKatzDev

PR #53395, opened 2024-01-03
CS: trailing commas for multiline-method params

keradus:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Issues        | CS 
| License       | MIT

separated & extracted from #53352 

To what I understood, @fabpot , @derrabus , @xabbuh and @OskarStark 
while @nicolas-grekas is against.
[but folks, correct that!]

This PR demonstrates how it can look like.


smnandre:
I'll rephrase what i meant (and probably badly expressed): 

Imho the question is not _only_ a matter of "multi-line",  it should depends on the _possibility_ to add another argument.

<details>
<summary> Examples </summary>

Two examples taken from this PR

A) Better

In this case i understand (createArgument can receive a thrid argument)

```diff
$argument = $this->createArgument(
    'stdClass',
-    new MapEntity(mapping: ['foo' => 'Foo'], exclude: ['bar'])
+    new MapEntity(mapping: ['foo' => 'Foo'], exclude: ['bar']),
);
```

B) Doubts 

Here i _personally_ do not see any gain. In the contrary i think this will create confusion about how many arguments can be passed.

```diff
$this->associationIdReader = new self($om, $om->getClassMetadata(
-    $classMetadata->getAssociationTargetClass($this->idField)
+    $classMetadata->getAssociationTargetClass($this->idField),
));
```

(just wanted to be clear about my doubts, won't bother you anymore 😅 )

</details>

keradus:
Just to make it explicit: the rule does not investigate declared function/method signature (and even not having access to it).

The idea behind the trailing comma doesn't connect to "how many elements there can be, ultimately", but to code formatting, easier adding new lines or reordering them (useful for named parameters or assoc arrays), also resulting in smaller diffs when touching the code, etc.

In initial PR some ppl liked it, some not, here we have separated PR for it as requested - for repo owners to decide which way to go, no sad feelings!

derrabus (approved):

mvorisek:
I am not fan of this change. Maybe I am not used to it.

I think the trailing comma should be used only in "declaration side" (array element, method params, ...) only, "method arguments" represents "calling side" instead, which is defined by the "declaration/target side" and thus, as visually harder to read, I do not think it is a good idea.

Are there any popular coding standards, even for different programming languages like Java, C, JS/TS, C#..., or big projects that use trailing comma for arguments?

compare:
```
$this->assertSame(
    [$this->obj2],
    $value
);
```
vs
```
$this->assertSame(
    [$this->obj2],
    $value,
);
```

derrabus:
> Maybe I am not used to it.

I think, that's mainly it.

> Are there any popular coding standards, even for different programming languages like Java, C, JS/TS, C#..., or big projects that use trailing comma for arguments?

Prettier for JS/TS for example.

Trailing commas for multi-line (!) listings are good. They allow for more consistent formatting and avoid changes in adjacent lines (e.g. when adding or removing an argument). We should enforce them where ever we can.

OskarStark (approved):

nicolas-grekas (commented):
I'm :-1: on this one.

The huge diff shows why: unlike the rule on parameters, this is never something we wanted to enforce in any reviews.

On parameters, the comma has the benefit of reducing diffs when adding a new parameter on a constructor typically. We know by experience that this happens quite often.

But we also know by experience that it's very rare to add arguments to an existing function call. BC rules make it not needed.

We could argue about reordering or adding *named* argument, but we don't use named arguments in core so this argument doesn't apply either.

I can get used to it but that doesn't make an argument in favor of the change.

nicolas-grekas:
Another argument against this PR: it's very uncommon coding style. This will add friction for contributors.

nicolas-grekas:
I'm closing here because we won't reach an agreement and the topic doesn't deserve that much attention.
Thanks for submitting.

PR #53399, opened 2024-01-04
[Mailer][Notifier] add more `#[\SensitiveParameter]` attributes

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT


xabbuh:
We are not consistent how we deal with username (public keys). Sometimes we mark them as sensitive parameters, sometimes we don't. That's why I didn't do anything about them.

OskarStark (approved):

nicolas-grekas (approved):
usernames don't need to be marked sensitive when considering security so we can skip marking them with the attribute to me

PR #53404, opened 2024-01-04
[Form][Security][Validator] Normalize translation files

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

I used the below script to spot mistakes and normalize translation files.

I propose to accept English where translations are missing because this makes it easier to automate and also to spot what's missing (opening only one file is needed.)

Carsonbot would need to be updated to account for this when checking for missing translations /cc @Nyholm 

<details>

```php
<?php

use Symfony\Component\Finder\Finder;
use Symfony\Component\Translation\Loader\XliffFileLoader;
use Symfony\Component\Translation\MessageCatalogue;

require __DIR__.'/vendor/autoload.php';

function dumpXliff1(string $defaultLocale, MessageCatalogue $messages, string $domain)
{
    $dom = new \DOMDocument('1.0', 'utf-8');
    $dom->formatOutput = true;

    $xliff = $dom->appendChild($dom->createElement('xliff'));
    $xliff->setAttribute('version', '1.2');
    $xliff->setAttribute('xmlns', 'urn:oasis:names:tc:xliff:document:1.2');

    $xliffFile = $xliff->appendChild($dom->createElement('file'));
    $xliffFile->setAttribute('source-language', str_replace('_', '-', $defaultLocale));
    $xliffFile->setAttribute('target-language', 'no' === $messages->getLocale() ? 'nb' : str_replace('_', '-', $messages->getLocale()));
    $xliffFile->setAttribute('datatype', 'plaintext');
    $xliffFile->setAttribute('original', 'file.ext');

    $xliffBody = $xliffFile->appendChild($dom->createElement('body'));
    foreach ($messages->all($domain) as $source => $target) {
        $translation = $dom->createElement('trans-unit');
        $metadata = $messages->getMetadata($source, $domain);

        $translation->setAttribute('id', $metadata['id']);

        $s = $translation->appendChild($dom->createElement('source'));
        $s->appendChild($dom->createTextNode($source));

        $text = 1 === preg_match('/[&<>]/', $target) ? $dom->createCDATASection($target) : $dom->createTextNode($target);

        $targetElement = $dom->createElement('target');

        if ('en' !== $messages->getLocale() && $target === $source && 'Error' !== $source) {
            $targetElement->setAttribute('state', 'needs-translation');
        }
        if (isset($metadata['target-attributes'])) {
            foreach ($metadata['target-attributes'] as $key => $value) {
                $targetElement->setAttribute($key, $value);
            }
        }

        $t = $translation->appendChild($targetElement);
        $t->appendChild($text);

        $xliffBody->appendChild($translation);
    }

    return preg_replace('/^ +/m', '$0$0', $dom->saveXML());
}


foreach (['Security/Core' => 'security', 'Form' => 'validators', 'Validator' => 'validators'] as $component => $domain) {
    $dir = __DIR__.'/src/Symfony/Component/'.$component.'/Resources/translations';

    $enCatalogue = (new XliffFileLoader())->load($dir.'/'.$domain.'.en.xlf', 'en', $domain);

    file_put_contents($dir.'/'.$domain.'.en.xlf', dumpXliff1('en', $enCatalogue, $domain));
    $finder = new Finder();

    foreach ($finder->files()->in($dir)->name('*.xlf') as $file) {
        $locale = substr($file->getBasename(), 1 + strlen($domain), -4);

        $catalogue = (new XliffFileLoader())->load($file, $locale, $domain);
        $localeCatalogue = new MessageCatalogue($locale);

        foreach ($enCatalogue->all($domain) as $id => $translation) {
            $metadata = [];
            if ($catalogue->defines($id, $domain)) {
                $translation = $catalogue->get($id, $domain);
                $metadata = $catalogue->getMetadata($id, $domain);
            }
            $metadata['id'] = $enCatalogue->getMetadata($id, $domain)['id'];
            $localeCatalogue->set($id, $translation, $domain);
            $localeCatalogue->setMetadata($id, $metadata, $domain);
        }

        file_put_contents($file, dumpXliff1('en', $localeCatalogue, $domain));
    }
}


```

</details>

### DIFF
--- a/src/Symfony/Component/Form/Resources/translations/validators.az.xlf
+++ b/src/Symfony/Component/Form/Resources/translations/validators.az.xlf
@@ -1,6 +1,6 @@
-<?xml version="1.0"?>
-<xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
-    <file source-language="en" datatype="plaintext" original="file.ext">
+<?xml version="1.0" encoding="utf-8"?>
+<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" version="1.2">
+    <file source-language="en" target-language="az" datatype="plaintext" original="file.ext">
         <body>
             <trans-unit id="28">
                 <source>This form should not contain extra fields.</source>
--- a/src/Symfony/Component/Security/Core/Resources/translations/security.en.xlf
+++ b/src/Symfony/Component/Security/Core/Resources/translations/security.en.xlf
@@ -1,6 +1,6 @@
-<?xml version="1.0"?>
-<xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
-    <file source-language="en" datatype="plaintext" original="file.ext">
+<?xml version="1.0" encoding="utf-8"?>
+<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" version="1.2">
+    <file source-language="en" target-language="en" datatype="plaintext" original="file.ext">
         <body>
             <trans-unit id="1">
                 <source>An authentication exception occurred.</source>
@@ -73,10 +73,6 @@
             <trans-unit id="19">
                 <source>Too many failed login attempts, please try again in %minutes% minute.</source>
                 <target>Too many failed login attempts, please try again in %minutes% minute.</target>
-            </trans-unit>
-            <trans-unit id="20">
-                <source>Too many failed login attempts, please try again in %minutes% minutes.</source>
-                <target>Too many failed login attempts, please try again in %minutes% minutes.</target>
             </trans-unit>
         </body>
     </file>
--- a/src/Symfony/Component/Security/Core/Resources/translations/security.th.xlf
+++ b/src/Symfony/Component/Security/Core/Resources/translations/security.th.xlf
@@ -1,6 +1,6 @@
-<?xml version="1.0"?>
-<xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
-    <file source-language="en" datatype="plaintext" original="file.ext">
+<?xml version="1.0" encoding="utf-8"?>
+<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" version="1.2">
+    <file source-language="en" target-language="th" datatype="plaintext" original="file.ext">
         <body>
             <trans-unit id="1">
                 <source>An authentication exception occurred.</source>
@@ -72,10 +72,6 @@
             </trans-unit>
             <trans-unit id="19">
                 <source>Too many failed login attempts, please try again in %minutes% minute.</source>
-                <target>มีความพยายามเข้าสู่ระบบล้มเหลวมากเกินไป โปรดลองอีกครั้งใน %minutes% นาที</target>
-            </trans-unit>
-            <trans-unit id="20">
-                <source>Too many failed login attempts, please try again in %minutes% minutes.</source>
                 <target>มีความพยายามเข้าสู่ระบบล้มเหลวมากเกินไป โปรดลองอีกครั้งใน %minutes% นาที</target>
             </trans-unit>
         </body>
--- a/src/Symfony/Component/Translation/Loader/XliffFileLoader.php
+++ b/src/Symfony/Component/Translation/Loader/XliffFileLoader.php
@@ -111,6 +111,10 @@ class XliffFileLoader implements LoaderInterface
                     continue;
                 }
 
+                if (isset($translation->target) && 'needs-translation' === (string) $translation->target->attributes()['state']) {
+                    continue;
+                }
+
                 $source = isset($attributes['resname']) && $attributes['resname'] ? $attributes['resname'] : $translation->source;
                 // If the xlf file has another encoding specified, try to convert it because
                 // simple_xml will always return utf-8 encoded values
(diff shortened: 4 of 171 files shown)


Nyholm:
> I propose to accept English where translations are missing because this makes it easier to automate and also to spot what's missing. 

But this will make it hard to know if a translation is missing or not. Do we really want to force users to do this? 

```php
$missing = $original !== $translation
```

src/Symfony/Component/Form/Resources/translations/validators.az.xlf
```diff
@@ -1,6 +1,6 @@
-<?xml version="1.0"?>
-<xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
-    <file source-language="en" datatype="plaintext" original="file.ext">
+<?xml version="1.0" encoding="utf-8"?>
+<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" version="1.2">
+    <file source-language="en" target-language="az" datatype="plaintext" original="validators.en.xlf">
```
Nyholm: Great!

nicolas-grekas:
I added `state="needs-translation"` to all `<target>` that aren't translated, as defined by the xliff spec for this purpose.

src/Symfony/Component/Security/Core/Resources/translations/security.th.xlf
```diff
@@ -74,10 +74,6 @@
                 <source>Too many failed login attempts, please try again in %minutes% minute.</source>
                 <target>มีความพยายามเข้าสู่ระบบล้มเหลวมากเกินไป โปรดลองอีกครั้งใน %minutes% นาที</target>
             </trans-unit>
-            <trans-unit id="20">
-                <source>Too many failed login attempts, please try again in %minutes% minutes.</source>
-                <target>มีความพยายามเข้าสู่ระบบล้มเหลวมากเกินไป โปรดลองอีกครั้งใน %minutes% นาที</target>
-            </trans-unit>
```
Nyholm, line 80: Removing this looks like a mistake. Can you verify?

nicolas-grekas: trans-unit 19 and 20 are exactly the same so I removed 20

src/Symfony/Component/Security/Core/Resources/translations/security.en.xlf
```diff
@@ -74,10 +74,6 @@
                 <source>Too many failed login attempts, please try again in %minutes% minute.</source>
                 <target>Too many failed login attempts, please try again in %minutes% minute.</target>
             </trans-unit>
-            <trans-unit id="20">
-                <source>Too many failed login attempts, please try again in %minutes% minutes.</source>
-                <target>Too many failed login attempts, please try again in %minutes% minutes.</target>
-            </trans-unit>
```
Nyholm, line 80: You snuck in this change too, right? 

It is removed because you could not find any use for it anymore.

Nyholm (approved):
Thank you

(Just update according to the test. Ie, the test is correct. )

nicolas-grekas:
PR ready, tests fixed!

fabpot:
I don't think the current xliff reader takes `needs-translation` into account. Something to add in this PR as well, right?

thunderer:
@nicolas-grekas I'm wondering, this will cause conflicts in all existing translation PRs, would it be possible to merge them before this one? Afterwards it seems like you could rerun the script to update the files.

Nyholm:
> I don't think the current xliff reader takes needs-translation into account.

Oh.. Does this mean my fallback locale is never used? 
Ie: 

```
framework:
    default_locale: 'fr'
    translator:
        fallbacks: ['sv']
```

If `validators.fr.xlf` is missing a translation, will it pick it up from validators.sv.xlf? Or, will the XliffReader think the translation is there?

src/Symfony/Component/Security/Core/Resources/translations/security.th.xlf
thunderer: Isn't 19 for singular variant `minute`?

nicolas-grekas:
> xliff reader takes needs-translation into account

Indeed. Updated!

src/Symfony/Component/Translation/Loader/XliffFileLoader.php
```diff
@@ -111,6 +111,10 @@ private function extractXliff1(\DOMDocument $dom, MessageCatalogue $catalogue, s
                     continue;
                 }
 
+                if (isset($translation->target) && 'needs-translation' === (string) $translation->target->attributes()['state']) {
```
stof, line 114: This would deserve a changelog entry IMO, while sneaking it in the normalization PR makes it invisible.

stof:
@nicolas-grekas can you update carsonbot to make it aware of those `needs-translation` attributes to consider them missing ?

kevinpapst:
I will open a bug report, because ignoring translations that have a `needs-translation` state should not be treated like missing translations.

Edit: https://github.com/symfony/symfony/issues/54541

nicolas-grekas:
I tackled this aspect in https://github.com/symfony/symfony/pull/54536

nicolas-grekas:
Ah no sorry, not the exact same topic :grimacing:

src/Symfony/Component/Translation/Loader/XliffFileLoader.php
kevinpapst: See https://github.com/symfony/symfony/issues/54541

PR #53414, opened 2024-01-04
[Serializer] `GetSetMethodNormalizer`: fix BC break with `#[Ignore]` attribute 

nikophil:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

This PR fixes a subtle BC break introduced by https://github.com/symfony/symfony/pull/52294/files: `GetSetMethodNormalizer` no more ignores methods marked as `#[Ignore]` with the old class `Symfony\Component\Serializer\Annotation\Ignore`

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/GetSetMethodNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/GetSetMethodNormalizer.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Component\Serializer\Normalizer;
 
+use Symfony\Component\Serializer\Annotation\Ignore as LegacyIgnore;
 use Symfony\Component\Serializer\Attribute\Ignore;
 
 /**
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/Serializer/Normalizer/GetSetMethodNormalizer.php
```diff
@@ -97,7 +97,7 @@ private function supports(string $class): bool
     private function isGetMethod(\ReflectionMethod $method): bool
     {
         return !$method->isStatic()
-            && !$method->getAttributes(Ignore::class)
+            && !$method->getAttributes(Ignore::class, \ReflectionAttribute::IS_INSTANCEOF)
```
nicolas-grekas: I think it'd be better to check the LegacyIgnore attribute directly instead of opening the attribute to inheritance (which opens to a potentially infinite number of child attributes)

nikophil: fixed

nicolas-grekas:
Thank you @nikophil.

PR #53417, opened 2024-01-04
[Translation] Add missing translations using ChatGPT

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

This is a follow up of #53404

The process around the maintenance of our translation files is too heavy currently. I feel like spending hours on the topic, and I guess it's the same for contributors.

Instead, I propose to account for the GPT tools that are really good at translating our short messages.

In order to still allow and call for human review,  I propose to use the `state="needs-review-translation"` on `<target>` tags as defined by XLIFF.

With some help from carsonbot, we'll be able to call for reviews per language. My proposal would be to make it able to send PRs that remove the `state` attribute but call for reviews by a native speaker before we could merge. We could even make the reviewer the author of the git commit if we want to keep some sort of history on the topic.

~@thunderer I'd prefer being explicit about using GPT. I'm going to replace your PRs by this one sorry.~

### DIFF
--- a/src/Symfony/Component/Form/Resources/translations/validators.cy.xlf
+++ b/src/Symfony/Component/Form/Resources/translations/validators.cy.xlf
@@ -4,135 +4,135 @@
         <body>
             <trans-unit id="28">
                 <source>This form should not contain extra fields.</source>
-                <target state="needs-translation">This form should not contain extra fields.</target>
+                <target state="needs-review-translation">Ni ddylai'r ffurflen gynnwys meysydd ychwanegol.</target>
             </trans-unit>
             <trans-unit id="29">
                 <source>The uploaded file was too large. Please try to upload a smaller file.</source>
-                <target state="needs-translation">The uploaded file was too large. Please try to upload a smaller file.</target>
+                <target state="needs-review-translation">Roedd y ffeil a uwchlwythwyd yn rhy fawr. Ceisiwch uwchlwytho ffeil llai.</target>
             </trans-unit>
             <trans-unit id="30">
                 <source>The CSRF token is invalid. Please try to resubmit the form.</source>
-                <target state="needs-translation">The CSRF token is invalid. Please try to resubmit the form.</target>
+                <target state="needs-review-translation">Mae'r tocyn CSRF yn annilys. Ceisiwch ailgyflwyno'r ffurflen.</target>
             </trans-unit>
             <trans-unit id="99">
                 <source>This value is not a valid HTML5 color.</source>
-                <target state="needs-translation">This value is not a valid HTML5 color.</target>
+                <target state="needs-review-translation">Nid yw'r gwerth hwn yn lliw HTML5 dilys.</target>
             </trans-unit>
             <trans-unit id="100">
                 <source>Please enter a valid birthdate.</source>
-                <target state="needs-translation">Please enter a valid birthdate.</target>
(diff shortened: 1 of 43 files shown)


thunderer:
@nicolas-grekas I specifically mentioned professional translator, because I actually work with one. I can provide a proof of that if necessary. When I had a native speaker available, I mentioned it explicitly.

nicolas-grekas:
OK, great. I'm surprised you're able to deliver so many languages at this rate.
On my side, the constant stream of PRs on the topic is a burden I'd like to cut, so there is my proposal.
I'd be happy to have reviews on GPT-generated translations instead!

thunderer:
Hi Nicolas, I was always explicit in my PRs about where the translations come from. I saw that maybe I'll be able to help quicker so I contacted a company and ordered those translations - all of them were made by native speakers or sworn translators, I have asked for that. Since they were ready today, I submitted all PRs. I didn't assign myself until they were ready, to not block anyone who would like to contribute. Would it be possible for you to merge my PRs once I rebase them?

nicolas-grekas:
Sure, I can do that. If you have more patches on the topic, it'd be nice if you could group them in one PR also. :pray:

thunderer:
@nicolas-grekas excellent, I'll rebase them shortly.

thunderer:
@nicolas-grekas all PRs rebased, checked diffs and they update only the `<target>` line removing the `needs-translation` attribute. I will group the changes next time for all translations I have available at given time, I'm sorry for causing additional work. I thought that one-PR-per-issue is expected since @carsonbot raised separate issues.

nicolas-grekas:
PR rebased and ready.

OskarStark:
@nicolas-grekas what do you think of having a list of native speakers, ready to review such messages, which could then get a ping after ChatGPT proposed a translation?

nicolas-grekas:
@OskarStark we would need a way to manage that list. And since things should remain open, people should be able to subscribe and unsubscribe at will.

One idea in this direction could be to have one special issue managed by carsbonbot that is opened/closed when needed, and that people could register to to be notified of when a new translation is needed. Or maybe one issue per language (but I'm not sure this is a good idea as that'd also mean spamming notifications - one single issue wouldn't). @Nyholm WDYT also?

nicolas-grekas:
> having a list of native speakers

~Oh, another idea would be to add willing translators to .github/CODEOWNER~

thunderer:
@nicolas-grekas I volunteer as tribute for Polish. :smile:

OskarStark:
> Oh, another idea would be to add willing translators to .github/CODEOWNER

That wouldn't work, I already had that idea for bridges in Notifier, Mailer etc. but this file only has some effect for members of the repo/orga 😢

yceruto (approved):
👍 Spanish translations at least :)

rosier:
> In order to still allow and call for human review, I propose to use the `state="needs-review-translation"` on `<target>` tags as defined by XLIFF.

WDYT of using `state="needs-review"` since the translation is already done by GPT and it just needs to be reviewed?

nicolas-grekas:
> WDYT of using state="needs-review"

It's not in the xliff spec: https://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#state

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

PR #53418, opened 2024-01-04
[FrameworkBundle][Notifier] Fix service registration (MessageBird + TurboSms)

smnandre:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #
| License       | MIT

Inside the [registerNotifierConfiguration](https://github.com/symfony/symfony/blob/9069d14de61d6d54ba62a6b611430dd5f9cadc78/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L2486) method, the [$classToServices](https://github.com/symfony/symfony/blob/9069d14de61d6d54ba62a6b611430dd5f9cadc78/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L2537) map (TransportFactoryClass => serviceId) used the Transport class (instead of the TransportFactory) for MessageBird and TurboSms.

The services are registered in [Ressources/config/notifier_transports.php](https://github.com/symfony/symfony/blob/9069d14de61d6d54ba62a6b611430dd5f9cadc78/src/Symfony/Bundle/FrameworkBundle/Resources/config/notifier_transports.php#L261) (with the correct Factory classes).

This map is then used in this loop:

```php
foreach ($classToServices as $class => $service) {
  $package = substr($service, \strlen('notifier.transport_factory.'));

  if (!ContainerBuilder::willBeAvailable(sprintf('symfony/%s-notifier', $package), $class, $parentPackages, true)) {
      $container->removeDefinition($service);
      $container->removeAlias(str_replace('-', '', $service)); // @deprecated to be removed in 6.0
  }
}
``` 

So i'm not sure it's considered a bug as IRL checking the Factory class or the Transport class did not make any difference as i read it....  but it's worth a fix for the future :) 

So i target 5.4 with no certitude :)

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
@@ -138,7 +138,7 @@ use Symfony\Component\Notifier\Bridge\LinkedIn\LinkedInTransportFactory;
 use Symfony\Component\Notifier\Bridge\Mailjet\MailjetTransportFactory as MailjetNotifierTransportFactory;
 use Symfony\Component\Notifier\Bridge\Mattermost\MattermostTransportFactory;
 use Symfony\Component\Notifier\Bridge\Mercure\MercureTransportFactory;
-use Symfony\Component\Notifier\Bridge\MessageBird\MessageBirdTransport;
+use Symfony\Component\Notifier\Bridge\MessageBird\MessageBirdTransportFactory;
 use Symfony\Component\Notifier\Bridge\MessageMedia\MessageMediaTransportFactory;
 use Symfony\Component\Notifier\Bridge\MicrosoftTeams\MicrosoftTeamsTransportFactory;
 use Symfony\Component\Notifier\Bridge\Mobyt\MobytTransportFactory;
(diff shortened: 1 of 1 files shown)


smnandre:
(fabbot unrelated)

OskarStark:
Good catch, to avoid such mistakes we could check inside the loop if class instanceof `AbstractFactory`, but that would be on runtime, and I don't think its worth

WDYT @nicolas-grekas ?

smnandre:
Maybe naive but is something like that possible ? 

```diff
-foreach ($classToServices as $class => $service) {
+foreach ($classToServices as $service) {
    $package = substr($service, \strlen('notifier.transport_factory.'));

-    if (!ContainerBuilder::willBeAvailable(sprintf('symfony/%s-notifier', $package), $class, $parentPackages)) {
+    if (!ContainerBuilder::willBeAvailable(sprintf('symfony/%s-notifier', $package), $container->getDefinition($service)->getClass(), $parentPackages)) {
        $container->removeDefinition($service);
    }
  }
```        

And then keep only the ids (that may be collected thanks to the tag too i guess?)

nicolas-grekas:
> Maybe naive but is something like that possible ?

Possible, can you give it a try on 7.1?

> (that may be collected thanks to the tag too i guess?)

That'd require a compiler pass. Which is maybe not a bad idea since it would allow registering external notifiers/etc.
To be confirmed there's a use case for that.

nicolas-grekas:
Thank you @smnandre.

smnandre:
> Possible, can you give it a try on 7.1?

Yep, i open a PR if that seems ok ?

PR #53421, opened 2024-01-04
Add .github/sync-translations.php

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

That's the script I used in #53404

With current branch numbers as example: when adding a new message to translate on 7.1, we would ask the author to add it to the "en.xlf" file on 5.4, to run this script, and only then to send the PR on 5.4
This would ensure new messages are duplicated in all languages.

We could also run chatgpt on that PR to pre-fill all translations, building on #53417

### DIFF
--- a/.github/sync-translations.php
+++ b/.github/sync-translations.php
@@ -1,0 +1,88 @@
+<?php
+
+// This script should be run after adding a new message to translate.
+// It will ensure that all messages in "*.en.xlf" files are propagated to all languages.
+// The resulting diff should then be submitted as a PR on the lowest maintained branch,
+// possibly after using GPT to translate all the targets it contains
+// (state="needs-review-translation" should then be used on corresponding target tags.)
+
+use Symfony\Component\Finder\Finder;
+use Symfony\Component\Translation\Loader\XliffFileLoader;
+use Symfony\Component\Translation\MessageCatalogue;
+
+require __DIR__.'/../vendor/autoload.php';
+
+function dumpXliff1(string $defaultLocale, MessageCatalogue $messages, string $domain)
+{
+    $dom = new \DOMDocument('1.0', 'utf-8');
+    $dom->formatOutput = true;
+
+    $xliff = $dom->appendChild($dom->createElement('xliff'));
+    $xliff->setAttribute('version', '1.2');
+    $xliff->setAttribute('xmlns', 'urn:oasis:names:tc:xliff:document:1.2');
+
+    $xliffFile = $xliff->appendChild($dom->createElement('file'));
+    $xliffFile->setAttribute('source-language', str_replace('_', '-', $defaultLocale));
+    $xliffFile->setAttribute('target-language', 'no' === $messages->getLocale() ? 'nb' : str_replace('_', '-', $messages->getLocale()));
+    $xliffFile->setAttribute('datatype', 'plaintext');
+    $xliffFile->setAttribute('original', 'file.ext');
+
+    $xliffBody = $xliffFile->appendChild($dom->createElement('body'));
+    foreach ($messages->all($domain) as $source => $target) {
+        $translation = $dom->createElement('trans-unit');
+        $metadata = $messages->getMetadata($source, $domain);
+
+        $translation->setAttribute('id', $metadata['id']);
+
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
With current branch number as example:
when adding a new message to translate on 7.1, we would ask the author to add it to the "en.xlf" file on 5.4, *to run this script*, and only then to send the PR on 5.4
This would ensure new messages are duplicated in all languages.

stof:
The file name looks weird as this script is not really _checking_ translations but adding missing untranslated messages.

.github/check-translations.php
```diff
@@ -0,0 +1,86 @@
+<?php
```
keradus:
```suggestion
<?php

declare(strict_types=1);
```

nicolas-grekas: we don't use strict types as you might know.

nicolas-grekas:
Script renamed to sync-translations.php
I also added a comment to explain how it should be used when adding a new message to translate.

.github/sync-translations.php
```diff
@@ -0,0 +1,90 @@
...
+
+foreach (['Security/Core' => 'security', 'Form' => 'validators', 'Validator' => 'validators'] as $component => $domain) {
+    $dir = __DIR__.'/../src/Symfony/Component/'.$component.'/Resources/translations';
+
+    $enCatalogue = (new XliffFileLoader())->load($dir.'/'.$domain.'.en.xlf', 'en', $domain);
+
+    file_put_contents($dir.'/'.$domain.'.en.xlf', dumpXliff1('en', $enCatalogue, $domain));
+    $finder = new Finder();
+
+    foreach ($finder->files()->in($dir)->name('*.xlf') as $file) {
```
xabbuh, line 69: What about skipping the `en` files here?

nicolas-grekas: I removed the dedicated `file_put_contents` instead (keeping "en" in the loops is the way to fix its cs now)

.github/check-translations.php
keradus: I know for whole project it's challenging, but for standalone script, I would recommend it ;)

.github/sync-translations.php
```diff
@@ -0,0 +1,88 @@
...
+    $dom = new \DOMDocument('1.0', 'utf-8');
+    $dom->formatOutput = true;
+
+    $xliff = $dom->appendChild($dom->createElement('xliff'));
+    $xliff->setAttribute('version', '1.2');
+    $xliff->setAttribute('xmlns', 'urn:oasis:names:tc:xliff:document:1.2');
+
+    $xliffFile = $xliff->appendChild($dom->createElement('file'));
+    $xliffFile->setAttribute('source-language', str_replace('_', '-', $defaultLocale));
+    $xliffFile->setAttribute('target-language', 'no' === $messages->getLocale() ? 'nb' : str_replace('_', '-', $messages->getLocale()));
```
stof, line 26: why this `no` vs `nb` thing ? Why would we have a `no` locale in Symfony if that's not the actual code of the locale ?

.github/check-translations.php
stof: strict mode is not safer than the coercive mode, especially in modern PHP version where unsafe coercion rules have been removed or deprecated (the actual safe version will be PHP 9 when the deprecated ones of PHP 8.x are turned into errors).
Strict mode tend to force you to use explicit casting instead. And explicit casting is **very** unsafe as it will happily perform operations that lose information totally silently. `(int) "I'm not a number"` will give you `0` without any warning, while an implicit coercion by a non-strict-mode parameter will reject it)

keradus: ha, i was actually preventing any casting in my codebase, but I guess I do not handle that much user input.

.github/sync-translations.php
nicolas-grekas: we have a test case that ensures that "no" is an alias of "nb", which it is semantically.

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

PR #53425, opened 2024-01-04
[Translation] Allow default parameters

### LINKED ISSUE #48182: [Translation] Default parameters
### Description

Hello

First of all, thank you for this amazing framework 🙏 

When using translation strings with params in large applications, we keep injecting the same params:
- a user's name, first, or last name
- a brand name
- an application name
- a customer's name
- ...

And it's pretty difficult to enforce a strict naming convention.
For instance, we can have these variations about a user's name:
- `user_name`
- `username`
- `userName`
- `name`

It would be nice to have a way to define a set of common params that don't need to be injected every time the `trans()` method is invoked within PHP code or a Twig template.

Thanks!

### Example

<?php

$translator->setDefaultParams([
    'first_name' => 'Sylvain',
    'company' => 'AssoConnect',
]);

// hello: 'Hello {first_name}, you work at {company}
$translor->trans('hello'); // Hello Sylvain, you work at AssoConnect

Jean-Beru:
| Q             | A
| ------------- | ---
| Branch?       | 7.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no 
| Issues        | Fix #48182
| License       | MIT

This PRs adds a `GlobalsTranslator` which decorates the `Translator` when global parameters are given.

Usage:
```yaml
# config/framework.yaml
framework:
    # ...
    translator:
        # ...
        globals:
            '{version}': '1.2.3'
            '%%application_name%%': 'My app' # "%" must be escaped if it is not a DI parameter
```
 
```yaml
# messages.en.yaml
app:
  version: 'Application version: {version}'
  name: 'Application name: %application_name%'
```

```twig
# twig template
{{ 'app.version'|trans }} # Displays "Application version: 1.2.3"
{{ 'app.version'|trans({ '{version}': '2.3.4' }) }} # Displays "Application version: 2.3.4"
{{ 'app.name'|trans }} # Displays "Application name: My app"
{{ 'app.name'|trans({ '{application_name}': 'My new app' }) }} # Displays "Application name: My new app"
```

Profiler view:
image


Jean-Beru:
Friendly ping @94noni :wink:

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -1580,6 +1581,18 @@ private function registerTranslatorConfiguration(array $config, ContainerBuilder
             }
         }
 
+        if (!empty($config['globals'])) {
+            $def = $container
+                ->register('translator.globals', GlobalsTranslator::class)
```
94noni: what about using the default `translator` service? 
I mean for twig, it is directly inside the `twig` service

94noni (commented):
nice :)
just a small comment regarding if we can add this feature on the default translator or not
obviously in our app we defined a decorator, but here it is core/new feature so I dont see why a new class is needed (or did I miss something?)

(and your example in PR desc part "# twig template" seems wrong for default using global)

### wrong

```twig
{{ 'app.version'|trans }} # Displays "1.2.3"
{{ 'app.version'|trans({ '{version}': '2.3.4' }) }} # Displays "2.3.4"
```

### correct

```twig
{{ 'app.version'|trans }} # Displays "Application version: 1.2.3"
{{ 'app.version'|trans({ '{version}': '2.3.4' }) }} # Displays "Application version: 2.3.4"
```

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
Jean-Beru: It was my first idea. What I had in mind:
* Create a new interface (in Contract ?)
* Tag services with this interface
* Call `addGlobal` with parameters given in configuration
It seems a little bit more complex.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -983,6 +984,13 @@ private function addTranslatorSection(ArrayNodeDefinition $rootNode, callable $e
                             ->end()
                             ->defaultValue([])
                         ->end()
+                        ->arrayNode('globals')
+                            ->info('Global parameters.')
+                            ->example(['pi' => 3.14])
```
94noni: I would use the '`{app_version}`' example as it also showcase the delimiters

94noni (approved):
lgtm, lets wait other reviews :)

src/Symfony/Component/Translation/GlobalsTranslator.php
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Translation;
+
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+final class GlobalsTranslator implements TranslatorInterface
```
nicolas-grekas: I would call that "Defaults", not globals (the scope is local in the implementation)

Jean-Beru: I renamed it to `DefaultParameters` to distinguish from `translator.default` service.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
Jean-Beru: Good idea, it's done.

maxbeckers (approved):

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -929,6 +929,7 @@ private function addTranslatorSection(ArrayNodeDefinition $rootNode, callable $e
                     ->fixXmlConfig('fallback')
                     ->fixXmlConfig('path')
                     ->fixXmlConfig('provider')
+                    ->fixXmlConfig('default_parameter')
```
nicolas-grekas: or just "parameter"?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -1580,6 +1581,18 @@ private function registerTranslatorConfiguration(array $config, ContainerBuilder
             }
         }
 
+        if (!empty($config['default_parameters'])) {
```
nicolas-grekas: we try to never use empty in the codebase

src/Symfony/Component/Translation/DefaultParametersTranslator.php
```diff
@@ -0,0 +1,41 @@
...
+
+        return $this->translator->trans($id, $parameters, $domain, $locale);
+    }
+
+    public function getLocale(): string
+    {
+        return $this->translator->getLocale();
+    }
+
+    public function addDefaultParameter(string $name, mixed $value): void
```
nicolas-grekas: it'd be better to replace this method by a constructor argument. would that work for you?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
welcoMattic: Or "global_parameters"?

nicolas-grekas: or just "global" like in twig? :D

src/Symfony/Component/Translation/Tests/DefaultParametersTranslatorTest.php
```diff
@@ -0,0 +1,41 @@
...
+        $translator->addLoader('array', new ArrayLoader());
+        $translator->addResource('array', ['test' => 'Welcome %name%!'], 'en');
+
+        $defaultParametersTranslator = new DefaultParametersTranslator($translator);
+        $defaultParametersTranslator->addDefaultParameter('%name%', 'Global name');
+
+        $this->assertSame('Welcome Global name!', $defaultParametersTranslator->trans('test'));
+        $this->assertSame('Welcome Name!', $defaultParametersTranslator->trans('test', ['%name%' => 'Name']));
+    }
+
```
welcoMattic: Could you add more tests to cover both Message format and Intl format please?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
Jean-Beru: OK. Changed for a `[] === ` comparison.

src/Symfony/Component/Translation/DefaultParametersTranslator.php
Jean-Beru: No problem, it's done.

src/Symfony/Component/Translation/Tests/DefaultParametersTranslatorTest.php
Jean-Beru: I added some tests with a different locale and with the ICU format. Is it OK for you ?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
Jean-Beru: So let's go for the following naming ?

`globals` in configuration
```yaml
framework:
    translator:
        globals:
            '{version}': '1.2.3'
            '{project_dir}': '%kernel.project_dir%'
```

And `default_parameters` for the service
```php
$container->register('translator.default_parameters', DefaultParametersTranslator::class)
```

Jean-Beru:
Could it be considered as a BC break since we can use a public method from the `Translator` service (from implemented interfaces or not) but not present in this decorator?

luxferre:
Can a global parameter value reference translation key?

Jean-Beru:
> Can a global parameter value reference translation key?

For now, in this PR, a global parameter can be used with a fixed value, a parameter or an environment variable.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
nicolas-grekas: yes please :)

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -985,6 +986,14 @@ private function addTranslatorSection(ArrayNodeDefinition $rootNode, callable $e
                             ->end()
                             ->defaultValue([])
                         ->end()
+                        ->arrayNode('globals')
+                            ->info('Global parameters.')
+                            ->example(['app_version' => 3.14])
+                            ->defaultValue([])
```
stof: this is useless as it is already the default behavior for prototypes array nodes.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -1567,6 +1568,16 @@ private function registerTranslatorConfiguration(array $config, ContainerBuilder
             }
         }
 
+        if ([] !== $config['globals']) {
+            $container
+                ->register('translator.default_parameters', DefaultParametersTranslator::class)
+                ->setDecoratedService('translator', null, -1) // Lower priority than "translator.data_collector"
```
stof: do we need a specific priority compared to the PseudoLocalizationTranslator ?

src/Symfony/Bundle/FrameworkBundle/Resources/config/schema/symfony-1.0.xsd
```diff
@@ -231,13 +231,15 @@
...
             <xsd:element name="provider" type="translation_provider" minOccurs="0" maxOccurs="unbounded" />
+            <xsd:element name="global" type="translation_global" minOccurs="0" maxOccurs="unbounded" />
         </xsd:sequence>
         <xsd:attribute name="enabled" type="xsd:boolean" />
         <xsd:attribute name="fallback" type="xsd:string" />
         <xsd:attribute name="logging" type="xsd:boolean" />
         <xsd:attribute name="formatter" type="xsd:string" />
         <xsd:attribute name="cache-dir" type="xsd:string" />
         <xsd:attribute name="default-path" type="xsd:string" />
+        <xsd:attribute name="global" type="xsd:string" />
```
stof: this attribute does not make sense.

stof (changes requested):

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
Jean-Beru: Done :slightly_smiling_face:

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
Jean-Beru: This priority is needed to avoid the following error: `The Translator "Symfony\Component\Translation\DefaultParametersTranslator" must implement TranslatorInterface, TranslatorBagInterface and LocaleAwareInterface.`.

On reflection, since the `DefaultParametersTranslator` only implements `TranslatorInterface`, using it could limit usages of the initial service (ex: I want to use the `getCatalogues` method from TranslatorBagInterface`).

@stof Do you think that it will be preferable to add this behavior to the `Translator` like mentioned in https://github.com/symfony/symfony/pull/53425#discussion_r1442025289 ?

src/Symfony/Bundle/FrameworkBundle/Resources/config/schema/symfony-1.0.xsd
Jean-Beru: Indeed!

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
Jean-Beru: Good catch

94noni (approved):
Nice :)

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
welcoMattic: IIUC the priority meaning, I guess we should put `DefaultParametersTranslator` before `PseudoLocalizationTranslator`, in order to leverage the pseudo-localization for replaced values of defaults parameters. WDYT?

src/Symfony/Component/Translation/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * Mark class `DataCollectorTranslator` as `final`
+ * Add `DefaultParametersTranslator` to allow defining default parameters
```
welcoMattic:
```suggestion
 * Add `DefaultParametersTranslator` to allow defining default translation parameters
```

src/Symfony/Component/Translation/Tests/DefaultParametersTranslatorTest.php
```diff
@@ -0,0 +1,64 @@
...
+        $this->assertSame('Bienvenue Global name!', $globalsTranslator->trans('welcome', [], null, 'fr'));
+        $this->assertSame('Welcome John!', $globalsTranslator->trans('welcome', ['{name}' => 'John']));
+        $this->assertSame('Bienvenue Jean!', $globalsTranslator->trans('welcome', ['{name}' => 'Jean'], null, 'fr'));
+    }
+
+    public function testTransICU()
+    {
+        if (!class_exists(\MessageFormatter::class)) {
+            $this->markTestSkipped(sprintf('Skipping test as the required "%s" class does not exist. Consider installing the "intl" PHP extension or the "symfony/polyfill-intl-messageformatter" package.', \MessageFormatter::class));
+        }
```
welcoMattic:
```suggestion
    /**
     * @requires extension intl
     */
    public function testTransICU()
    {
```

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
Jean-Beru: > IIUC the priority meaning, I guess we should put `DefaultParametersTranslator` before `PseudoLocalizationTranslator`, in order to leverage the pseudo-localization for replaced values of defaults parameters. WDYT?

Not really because `PseudoLocalizationTranslator` calls the decorated stack before updating the translated text. Also it does not rely on parameters, which are the only thing changed by `DefaultParametersTranslator`.

See https://github.com/symfony/symfony/blob/9519b54417c09c49496a4a6be238e63be9a73465/src/Symfony/Component/Translation/PseudoLocalizationTranslator.php#L91

src/Symfony/Component/Translation/Tests/DefaultParametersTranslatorTest.php
Jean-Beru: Is this class necessarily loaded? If so, should we will have to update https://github.com/symfony/symfony/blob/4ad506b0855f45061c2d4b2600846a8aaf338f98/src/Symfony/Component/Translation/Tests/TranslatorCacheTest.php#L62 and https://github.com/symfony/symfony/blob/33d71aa7b7c368e9752198413c9fbd0ca8d82617/src/Symfony/Component/Translation/Tests/TranslatorTest.php#L406 too.

welcoMattic: Yes, this class lives in the intl extension. So, marking the test requiring the extension avoid to check on the class existence on all tests that require intl extension. Am I wrong @xabbuh?

Jean-Beru: Indeed. I updated the tests this way.

nicolas-grekas (commented):
I'm wondering if we really need a decorator: adding this concern on the existing Translator looks trivial to me.

94noni:
[I totally agree](https://github.com/symfony/symfony/pull/53425#discussion_r1442025289)
would love to have it on main service :)

Jean-Beru:
@nicolas-grekas @94noni Indeed, much simpler like that :+1: 

I added the global parameters list to the profiler to make debugging easier. A screenshot is available in the description of this PR.

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/translation.html.twig
```diff
@@ -156,6 +156,27 @@
...
+            <tr>
+                <th>Name</th>
+                <th>Value</th>
+            </tr>
+            </thead>
+            <tbody>
+            {% for name, value in collector.globalParameters %}
+                <tr>
+                    <td class="font-normal text-small text-bold nowrap">{{ name }}</td>
+                    <td class="font-normal text-small nowrap">{{ value }}</td>
```
94noni: as globalParameters is an array of "string, mixed"
how will `{{ value }}` be rendered here?
perhaps use the `profiler_dump()` function?

also while talking about mixed, perhaps add an example / test / doc of for example a global array params ? (if handled as is)

```yaml
globals:
    hosts:
        'FR': 'https://foo.bar.fr'
        'ES': 'https://foo.bar.es'
```

// totaly arbitrary example

Jean-Beru: For now configuration only allows scalar values. To make it simpler we could restrict to `int|float|string| for now.

Your example seems to be related to [this comment](https://github.com/symfony/symfony/pull/53425#issuecomment-1974769159). Since a parameter can be a `TranslatableMessage`, it could be interesting to allow an other way to configure global parameters:
```yaml
# Use a value
globals:
    host: 
        value: 'https://unique.site.com' # use this value

# Use a value (short alias)
globals:
    host: 'https://unique.site.com' # use this value

# Use a message key
globals:
    host:
        message: 'host' # use TranslatableMessage('host') with the same parameters and domain

# Use a message key from an other domain
globals:
    host:
        message: 'host' # use TranslatableMessage('host') with the same parameters but "common" domain
        domain: 'common'
```

WDYT ?

94noni: Yes can makes sens

Jean-Beru:
I had some difficulties with infinite loops but `TranslatableMessage` can be used as globals now. :tada:

94noni (approved):

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1020,6 +1021,33 @@ private function addTranslatorSection(ArrayNodeDefinition $rootNode, callable $e
...
+                                    ->arrayNode('parameters')
+                                        ->normalizeKeys(false)
+                                        ->useAttributeAsKey('name')
+                                        ->scalarPrototype()->end()
+                                    ->end()
+                                    ->stringNode('domain')->end()
+                                ->end()
+                                ->beforeNormalization()
+                                    ->ifTrue(fn ($v) => !\is_array($v))
+                                    ->then(fn ($v): array => ['value' => $v])
```
nicolas-grekas: the closures could all be made static also
```suggestion
                                    ->then(fn ($v) => ['value' => $v])
```

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -1594,6 +1594,16 @@ private function registerTranslatorConfiguration(array $config, ContainerBuilder
             $translator->replaceArgument(4, $options);
         }
 
+        if ([] !== $config['globals']) {
```
nicolas-grekas:
```suggestion
        if ($config['globals']) {
```

nicolas-grekas: but actually there is no need for this "if" nesting at all

src/Symfony/Component/Translation/Translator.php
```diff
@@ -174,6 +210,13 @@ public function trans(?string $id, array $parameters = [], ?string $domain = nul
             }
         }
 
+        if ([] !== $this->globalTranslatableParameters && !isset($this->globalParameters[$locale])) {
```
nicolas-grekas: we don't use empty-values comparisons in Symfony (same everywhere and for your next PRs ;) )
```suggestion
        if ($this->globalTranslatableParameters && !isset($this->globalParameters[$locale])) {
```

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/php/translator_globals.php
```diff
@@ -0,0 +1,15 @@
...
+$container->loadFromExtension('framework', [
+    'annotations' => false,
+    'http_method_override' => false,
+    'handle_all_throwables' => true,
+    'php_errors' => ['log' => true],
+    'translator' => [
+        'globals' => [
+            '%%app_name%%' => 'My application',
+            '{app_version}' => '1.2.3',
+            'url' => ['message' => 'url', 'parameters' => ['scheme' => 'https://'], 'domain' => 'global'],
```
nicolas-grekas: even in tests, the key should use delimiters to be clear about the fact it's going to be needed in real apps also

src/Symfony/Component/Translation/Translator.php
Jean-Beru: Noted, thanks :slightly_smiling_face:

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/translation.html.twig
```diff
@@ -156,6 +156,31 @@
...
+                <th>Locale</th>
+                <th>Message ID</th>
+                <th>Value</th>
+            </tr>
+            </thead>
+            <tbody>
+            {% for locale, values in collector.globalParameters %}
+                {% for id, value in values %}
+                    <tr>
+                        <td class="font-normal text-small nowrap">{{ locale != '*' ? "#{locale}" }}</td>
```
welcoMattic:
```suggestion
                        <td class="font-normal text-small nowrap">{{ locale != '*' ? "{locale}" }}</td>
```

To stay consistent with existing table in translation panel
image

src/Symfony/Component/Translation/Translator.php
```diff
@@ -155,6 +165,32 @@ public function getFallbackLocales(): array
         return $this->fallbackLocales;
     }
 
+    /**
+     * All a global parameter.
+     */
```
welcoMattic: Not necessary, I think the method name is self-describing

src/Symfony/Component/Translation/Translator.php
```diff
@@ -155,6 +165,32 @@ public function getFallbackLocales(): array
...
+    public function addGlobalParameter(string $id, mixed $value): void
+    {
+        $this->globalParameters['*'][$id] = $value;
+    }
+
+    /**
+     * All a global translatable parameter.
+     */
```
welcoMattic: Not necessary, I think the method name is self-describing

src/Symfony/Component/Translation/Translator.php
```diff
@@ -155,6 +165,32 @@ public function getFallbackLocales(): array
...
+    public function addGlobalTranslatableParameter(string $id, string $message, array $parameters = [], ?string $domain = null): void
+    {
+        $this->globalTranslatableParameters[$id] = new TranslatableMessage($message, $parameters, $domain);
+    }
+
+    /**
+     * Gets the global parameters.
```
welcoMattic: Not necessary, I think the method name is self-describing

welcoMattic (commented):
Minor comments, otherwise it's ✅ for me!
@Jean-Beru could you squash your commits please?

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/translation.html.twig
Jean-Beru: I'm not sure to understand. If we remove `#`, the string interpolation won't be done and it will lead to a column with `{locale}` values.

Jean-Beru:
> Minor comments, otherwise it's ✅ for me! @Jean-Beru could you squash your commits please?

Thanks @welcoMattic !

PR is fixed, squashed and rebased.

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/translation.html.twig
stof: why using a string containing *only an interpolation of a variable ? Use the variable directly instead.
In PHP, are you using `"$myVar"` instead of `$myVar` when you want to pass a string coming from the `$myVar` variable to another part of the code ?

src/Symfony/Bundle/FrameworkBundle/Resources/config/schema/symfony-1.0.xsd
```diff
@@ -285,6 +286,24 @@
...
+        <xsd:sequence>
+            <xsd:element name="parameter" type="translation_global_parameter" minOccurs="0" maxOccurs="unbounded" />
+        </xsd:sequence>
+        <xsd:attribute name="name" type="xsd:string" use="required" />
+        <xsd:attribute name="value" type="xsd:string" />
+        <xsd:attribute name="message" type="xsd:string" />
+        <xsd:attribute name="domain" type="xsd:string" />
+    </xsd:complexType>
+
+    <xsd:complexType name="translation_global_parameter" mixed="true">
```
stof: why `mixed="true` ?

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1020,6 +1021,33 @@ private function addTranslatorSection(ArrayNodeDefinition $rootNode, callable $e
...
                         ->end()
+                        ->arrayNode('globals')
+                            ->info('Global parameters.')
+                            ->example(['app_version' => 3.14])
+                            ->normalizeKeys(false)
+                            ->useAttributeAsKey('name')
+                            ->arrayPrototype()
+                                ->fixXmlConfig('parameter')
+                                ->children()
+                                    ->variableNode('value')->end()
```
stof, line 1033: This should be a scalar node, not a variable node. There is no reason to support non-scalar values as those would fail at runtime (an array cannot be casted to string to be inserted in the translation)

src/Symfony/Component/Translation/Translator.php
```diff
@@ -155,6 +165,24 @@ public function getFallbackLocales(): array
         return $this->fallbackLocales;
     }
 
+    public function addGlobalParameter(string $id, mixed $value): void
```
stof: this is not consistent with the property type that says `string|int|float`

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/translation.html.twig
Jean-Beru: I wanted to go too fast :facepalm:

It's fixed.

src/Symfony/Component/Translation/Translator.php
```diff
@@ -174,6 +202,13 @@ public function trans(?string $id, array $parameters = [], ?string $domain = nul
             }
         }
 
+        if ($this->globalTranslatableParameters && !isset($this->globalParameters[$locale])) {
+            $this->globalParameters[$locale] = []; // Avoid infinite loops
```
stof: the storage of resolved global parameters per locale should be separate from the storage of the user-provided configuration IMO, instead of using a `*` key.

src/Symfony/Component/Translation/Translator.php
```diff
@@ -174,6 +202,13 @@ public function trans(?string $id, array $parameters = [], ?string $domain = nul
...
+            $this->globalParameters[$locale] = array_map(fn ($parameter) => $parameter->trans($this, $locale), $this->globalTranslatableParameters);
```
stof: Given that `$parameters` below has support for TranslatableInterface, do we actually need to handle those global parameters early ?

src/Symfony/Component/Translation/Tests/TranslatorTest.php
```diff
@@ -602,6 +600,56 @@ public function testMissingLoaderForResourceError()
...
+        $translator->addGlobalTranslatableParameter('{link}', 'url', [], 'globals');
+
+        $this->assertSame('Welcome on example.com/admin!', $translator->trans('welcome'));
+        $this->assertSame('Bienvenue sur example.fr/admin!', $translator->trans('welcome', [], null, 'fr'));
+        $this->assertSame('Welcome on other.com!', $translator->trans('welcome', ['{link}' => 'other.com']));
+        $this->assertSame('Bienvenue sur autre.fr!', $translator->trans('welcome', ['{link}' => 'autre.fr'], null, 'fr'));
+    }
+
+    /**
+     * @requires extension intl
```
stof, line 639: Does this really requires the extension ? We support the ICU format without the extension thanks to our polyfill package.

src/Symfony/Component/Translation/Tests/TranslatorTest.php
```diff
@@ -399,14 +399,12 @@ public function testTrans($expected, $id, $translation, $parameters, $locale, $d
     }
 
     /**
+     * @requires extension intl
```
stof: This is not the equivalent of the existing check

src/Symfony/Component/Translation/Tests/TranslatorTest.php
Jean-Beru: If we can use the polyfill, we may remove this requirement from all tests, right?

https://github.com/symfony/symfony/blob/6f4f04b98f6134996a57d21d1851d558a34dc45c/src/Symfony/Component/Translation/Tests/TranslatorTest.php#L561

src/Symfony/Component/Translation/Translator.php
Jean-Beru: We need to do this in order to avoid infinite loops since all parameters are translated next using the same `trans` method.

src/Symfony/Component/Translation/Translator.php
Jean-Beru: Done using a `globalTranslatedParameters`.

nicolas-grekas (commented):
<details>
<summary>Here is my review as a patch.</summary>

```diff
diff --git a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
index 44e9816e18..eba4ee4534 100644
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
@@ -180,6 +180,7 @@ use Symfony\Component\Translation\Command\XliffLintCommand as BaseXliffLintComma
 use Symfony\Component\Translation\Extractor\PhpAstExtractor;
 use Symfony\Component\Translation\LocaleSwitcher;
 use Symfony\Component\Translation\PseudoLocalizationTranslator;
+use Symfony\Component\Translation\TranslatableMessage;
 use Symfony\Component\Translation\Translator;
 use Symfony\Component\TypeInfo\Type;
 use Symfony\Component\TypeInfo\TypeResolver\PhpDocAwareReflectionTypeResolver;
@@ -1612,11 +1613,7 @@ class FrameworkExtension extends Extension
         }
 
         foreach ($config['globals'] as $name => $global) {
-            if (isset($global['value'])) {
-                $translator->addMethodCall('addGlobalParameter', [$name, $global['value']]);
-            } else {
-                $translator->addMethodCall('addGlobalTranslatableParameter', [$name, $global['message'], $global['parameters'] ?? [], $global['domain'] ?? null]);
-            }
+            $translator->addMethodCall('addGlobalParameter', [$name, $global['value'] ?? new Definition(TranslatableMessage::class, [$global['message'], $global['parameters'] ?? [], $global['domain'] ?? null])]);
         }
 
         if ($config['pseudo_localization']['enabled']) {
diff --git a/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/translation.html.twig b/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/translation.html.twig
index 5babc0d893..18319e4993 100644
--- a/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/translation.html.twig
+++ b/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/translation.html.twig
@@ -172,18 +172,9 @@
                 <tr>
                     <td class="font-normal text-small nowrap"></td>
                     <td class="font-normal text-small nowrap">{{ id }}</td>
-                    <td class="font-normal text-small nowrap">{{ value }}</td>
+                    <td class="font-normal text-small nowrap">{{ profiler_dump(value) }}</td>
                 </tr>
             {% endfor %}
-            {% for locale, values in collector.globalTranslatedParameters %}
-                {% for id, value in values %}
-                    <tr>
-                        <td class="font-normal text-small nowrap">{{ locale }}</td>
-                        <td class="font-normal text-small nowrap">{{ id }}</td>
-                        <td class="font-normal text-small nowrap">{{ value }}</td>
-                    </tr>
-                {% endfor %}
-            {% endfor %}
             </tbody>
         </table>
     {% endif %}
diff --git a/src/Symfony/Component/Translation/CHANGELOG.md b/src/Symfony/Component/Translation/CHANGELOG.md
index 64502a89fd..365c5cf1cd 100644
--- a/src/Symfony/Component/Translation/CHANGELOG.md
+++ b/src/Symfony/Component/Translation/CHANGELOG.md
@@ -4,7 +4,7 @@ CHANGELOG
 7.3
 ---
 
- * Add `addGlobalParameter` and `addGlobalTranslatableParameter` to allow defining global translation parameters
+ * Add `Translator::addGlobalParameter()` to allow defining global translation parameters
 
 7.2
 ---
diff --git a/src/Symfony/Component/Translation/DataCollector/TranslationDataCollector.php b/src/Symfony/Component/Translation/DataCollector/TranslationDataCollector.php
index bb7dc6b399..e2e597a705 100644
--- a/src/Symfony/Component/Translation/DataCollector/TranslationDataCollector.php
+++ b/src/Symfony/Component/Translation/DataCollector/TranslationDataCollector.php
@@ -45,7 +45,6 @@ class TranslationDataCollector extends DataCollector implements LateDataCollecto
         $this->data['locale'] = $this->translator->getLocale();
         $this->data['fallback_locales'] = $this->translator->getFallbackLocales();
         $this->data['global_parameters'] = $this->translator->getGlobalParameters();
-        $this->data['global_translated_parameters'] = $this->translator->getGlobalTranslatedParameters();
     }
 
     public function reset(): void
@@ -94,14 +93,6 @@ class TranslationDataCollector extends DataCollector implements LateDataCollecto
         return $this->data['global_parameters'] ?? [];
     }
 
-    /**
-     * @internal
-     */
-    public function getGlobalTranslatedParameters(): Data|array
-    {
-        return $this->data['global_translated_parameters'] ?? [];
-    }
-
     public function getName(): string
     {
         return 'translation';
diff --git a/src/Symfony/Component/Translation/Translator.php b/src/Symfony/Component/Translation/Translator.php
index 6994cb9b6f..d57ec9e2c9 100644
--- a/src/Symfony/Component/Translation/Translator.php
+++ b/src/Symfony/Component/Translation/Translator.php
@@ -61,15 +61,10 @@ class Translator implements TranslatorInterface, TranslatorBagInterface, LocaleA
     private bool $hasIntlFormatter;
 
     /**
-     * @var array<string, string|int|float>
+     * @var array<string, string|int|float|TranslatableInterface>
      */
     private array $globalParameters = [];
 
-    /**
-     * @var array<string, TranslatableInterface>
-     */
-    private array $globalTranslatableParameters = [];
-
     /**
      * @var array<string, string|int|float>
      */
@@ -170,16 +165,11 @@ class Translator implements TranslatorInterface, TranslatorBagInterface, LocaleA
         return $this->fallbackLocales;
     }
 
-    public function addGlobalParameter(string $id, string|int|float $value): void
+    public function addGlobalParameter(string $id, string|int|float|TranslatableInterface $value): void
     {
         $this->globalParameters[$id] = $value;
     }
 
-    public function addGlobalTranslatableParameter(string $id, string $message, array $parameters = [], ?string $domain = null): void
-    {
-        $this->globalTranslatableParameters[$id] = new TranslatableMessage($message, $parameters, $domain);
-    }
-
     /**
      * @internal
      */
@@ -188,14 +178,6 @@ class Translator implements TranslatorInterface, TranslatorBagInterface, LocaleA
         return $this->globalParameters;
     }
 
-    /**
-     * @internal
-     */
-    public function getGlobalTranslatedParameters(): array
-    {
-        return $this->globalTranslatedParameters;
-    }
-
     public function trans(?string $id, array $parameters = [], ?string $domain = null, ?string $locale = null): string
     {
         if (null === $id || '' === $id) {
@@ -215,17 +197,24 @@ class Translator implements TranslatorInterface, TranslatorBagInterface, LocaleA
             }
         }
 
-        if ($this->globalTranslatableParameters && !isset($this->globalTranslatedParameters[$locale])) {
-            $this->globalTranslatedParameters[$locale] = []; // Avoid infinite loops
-            $this->globalTranslatedParameters[$locale] = array_map(fn ($parameter) => $parameter->trans($this, $locale), $this->globalTranslatableParameters);
+        foreach ($parameters as $key => $value) {
+            if ($value instanceof TranslatableInterface) {
+                $parameters[$key] = $value->trans($this, $locale);
+            }
         }
-        if ($this->globalParameters) {
-            $parameters += $this->globalParameters;
+
+        if (null === $globalParameters =& $this->globalTranslatedParameters[$locale]) {
+            $globalParameters = $this->globalParameters;
+            foreach ($globalParameters as $key => $value) {
+                if ($value instanceof TranslatableInterface) {
+                    $globalParameters[$key] = $value->trans($this, $locale);
+                }
+            }
         }
-        if (isset($this->globalTranslatedParameters[$locale])) {
-            $parameters += $this->globalTranslatedParameters[$locale];
+
+        if ($globalParameters) {
+            $parameters += $globalParameters;
         }
-        $parameters = array_map(fn ($parameter) => $parameter instanceof TranslatableInterface ? $parameter->trans($this, $locale) : $parameter, $parameters);
 
         $len = \strlen(MessageCatalogue::INTL_DOMAIN_SUFFIX);
         if ($this->hasIntlFormatter
```

</details>

The TL;DR is that I don't think that we need nor should have two methods for this.
Also, the PR is missing some cross-components checks I think (fwb / profiler are using feats that might not be available depending on the version of translator that's installed.)

Jean-Beru:
> Here is my review as a patch.
> The TL;DR is that I don't think that we need nor should have two methods for this. Also, the PR is missing some cross-components checks I think (fwb / profiler are using feats that might not be available depending on the version of translator that's installed.)

Thanks for the review :heart: 

> the PR is missing some cross-components checks

I updated the profiler part view but what about the bundle? Should we use `InstalledVersions::satisfies` to check component version, use a dedicated interface or just check the method existence?

stof:
The cleanest solution for FrameworkBundle is to bump the min supported version of `symfony/translation` (note that the `conflict` rule is the most important one here, as `require-dev` only impacts our CI)

Jean-Beru:
> The cleanest solution for FrameworkBundle is to bump the min supported version of `symfony/translation` (note that the `conflict` rule is the most important one here, as `require-dev` only impacts our CI)

Thanks for the tip!

src/Symfony/Component/Translation/Tests/TranslatorCacheTest.php
```diff
@@ -55,14 +55,12 @@ protected function deleteTmpDir()
     }
 
     /**
+     * @requires extension intl
```
xabbuh: Is there a reason for this change? If I don't miss anything the test was runnable without the extension if the `symfony/polyfill-intl-messageformatter` was installed. This doesn't seem to be the case anymore.

src/Symfony/Component/Translation/Translator.php
```diff
@@ -174,7 +197,24 @@ public function trans(?string $id, array $parameters = [], ?string $domain = nul
             }
         }
 
-        $parameters = array_map(fn ($parameter) => $parameter instanceof TranslatableInterface ? $parameter->trans($this, $locale) : $parameter, $parameters);
+        foreach ($parameters as $key => $value) {
```
xabbuh, line 198: Do we need this loop?

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/translation.html.twig
```diff
@@ -156,6 +156,27 @@
         {% endblock messages %}
     {% endif %}
 
+    {% if collector.globalParameters is defined and collector.globalParameters is not empty %}
```
xabbuh: can be shortened to

```suggestion
    {% if collector.globalParameters|default([]) %}
```

src/Symfony/Component/Translation/Translator.php
nicolas-grekas: it's more efficient code since we *maybe* mutate an existing array instead of always copying it

src/Symfony/Component/Translation/Translator.php
```diff
@@ -174,7 +197,24 @@ public function trans(?string $id, array $parameters = [], ?string $domain = nul
...
+            if ($value instanceof TranslatableInterface) {
+                $parameters[$key] = $value->trans($this, $locale);
+            }
+        }
+
+        if (null === $globalParameters =& $this->globalTranslatedParameters[$locale]) {
```
xabbuh, line 204: What is the purpose of the `$globalTranslatedParameters` property? Is it there to cache already translated values? If so, isn't this broken for the case where the is a sequence of `trans()`, `addGlobalParameter()`, `trans()` calls?

src/Symfony/Component/Translation/DataCollectorTranslator.php
```diff
@@ -82,6 +82,15 @@ public function getFallbackLocales(): array
         return [];
     }
 
+    public function getGlobalParameters(): array
+    {
+        if ($this->translator instanceof Translator || method_exists($this->translator, 'getGlobalParameters')) {
```
nicolas-grekas, line 87: is the method_exists check needed?

src/Symfony/Component/Translation/Translator.php
nicolas-grekas: Which case do you have in mind? IIUC, the cached data doesn't depend on the input so when would this break?

src/Symfony/Component/Translation/Translator.php
```diff
@@ -155,6 +165,19 @@ public function getFallbackLocales(): array
         return $this->fallbackLocales;
     }
 
+    public function addGlobalParameter(string $id, string|int|float|TranslatableInterface $value): void
+    {
+        $this->globalParameters[$id] = $value;
+    }
+
+    /**
+     * @internal
```
nicolas-grekas: let's remove this annotation? I don't see it helping us (maintainers) much, while it might be useful for userland, next to addGlobalParameter

src/Symfony/Component/Translation/Translator.php
xabbuh: After the first `trans` call `$globalParameters` isn't `null` anymore. So if you add new globals later on they wouldn't be translated on the next `trans()` call, right?

nicolas-grekas: ah so we should empty this->globalTranslatedParameters when we add a global indeed!
in practice this shouldn't happen but better be safe

xabbuh: And we should probably also implement the `ResetInterface()`?

src/Symfony/Component/Translation/DataCollectorTranslator.php
xabbuh: One could ship their own `TranslatorInterface` implementation having global parameters support too. But maybe having a dedicated interface might be a better solution.

nicolas-grekas: this is `instanceof Translator` so the method is going to be always here to me, no?

src/Symfony/Component/Translation/Translator.php
nicolas-grekas: Not sure about ResetInterface: are there any other factors to consider that'd need a reset?

src/Symfony/Component/Translation/DataCollectorTranslator.php
xabbuh: but `$this->translator` doesn't necessarily have to be an instance of `Translator` 🤔

nicolas-grekas: Oh, NVM, I read the `||` as `&&` :facepalm:

src/Symfony/Component/Translation/Translator.php
Jean-Beru: Sure ! It's done.

Is there any reason to keep it for the `getFallbackLocales` method ?

src/Symfony/Bundle/WebProfilerBundle/Resources/views/Collector/translation.html.twig
Jean-Beru: Good catch.

src/Symfony/Component/Translation/Tests/TranslatorCacheTest.php
Jean-Beru: Honestly, I don't remember why I changed it. I revert it.

src/Symfony/Component/Translation/Translator.php
Jean-Beru: > If so, isn't this broken for the case where the is a sequence of trans(), addGlobalParameter(), trans() calls?

Indeed ! `$globalTranslatedParameters` is now emptied when using `addGlobalParameter`.

fabpot (approved):

fabpot:
Thank you @Jean-Beru.

PR #53432, opened 2024-01-05
[HttpFoundation] Request without content-type or content-length header should result in null values, not empty strings

priyadi:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

An HTTP request without content-type and content length will result in the variables `$_SERVER['CONTENT_TYPE']` and `$_SERVER['CONTENT_LENGTH']` having an empty string in PHP.

For example, the request:

> POST / HTTP/1.1
> Host: app.dev.localhost:8000
> User-Agent: curl/8.4.0
> Accept: */*

(from curl invocation: `curl -X 'POST' 'http://app.dev.localhost:8000/' -H 'Content-Type:' -H 'Content-length:' -d '' -v> /dev/null`)

It will result in the variables `$_SERVER['CONTENT_TYPE']` and `$_SERVER['CONTENT_LENGTH']` containing an empty string.

In turn, `$request->headers->get('Content-type')` and `$request->headers->get('Content-length')` also result in an empty string.

This PR changes so that `$request->headers->get('Content-type')` and `$request->headers->get('Content-length')` will reflect the original state of the request header, that they return null (due to the fact the headers did not exist in the request), not an empty string (erroneously implying the client sent an empty content-type and content-length headers).

### DIFF
--- a/src/Symfony/Component/HttpFoundation/Tests/ServerBagTest.php
+++ b/src/Symfony/Component/HttpFoundation/Tests/ServerBagTest.php
@@ -177,4 +177,20 @@ class ServerBagTest extends TestCase
             'PHP_AUTH_PW' => '',
         ], $bag->getHeaders());
     }
+
+    /**
+     * An HTTP request without content-type and content-length will result in
+     * the variables $_SERVER['CONTENT_TYPE'] and $_SERVER['CONTENT_LENGTH']
+     * containing an empty string in PHP.
+     */
+    public function testRequestWithoutContentTypeAndContentLength()
+    {
+        $bag = new ServerBag([
+            'CONTENT_TYPE' => '',
+            'CONTENT_LENGTH' => '',
+            'HTTP_USER_AGENT' => 'foo',
+        ]);
+
+        $this->assertSame(['USER_AGENT' => 'foo'], $bag->getHeaders());
+    }
 }
--- a/src/Symfony/Component/HttpFoundation/ServerBag.php
+++ b/src/Symfony/Component/HttpFoundation/ServerBag.php
@@ -31,7 +31,7 @@ class ServerBag extends ParameterBag
         foreach ($this->parameters as $key => $value) {
             if (str_starts_with($key, 'HTTP_')) {
                 $headers[substr($key, 5)] = $value;
-            } elseif (\in_array($key, ['CONTENT_TYPE', 'CONTENT_LENGTH', 'CONTENT_MD5'], true)) {
+            } elseif (\in_array($key, ['CONTENT_TYPE', 'CONTENT_LENGTH', 'CONTENT_MD5'], true) && '' !== $value) {
                 $headers[$key] = $value;
             }
         }


src/Symfony/Component/HttpFoundation/Tests/ServerBagTest.php
```diff
@@ -177,4 +177,20 @@ public function testItDoesNotOverwriteTheAuthorizationHeaderIfItIsAlreadySet()
...
+     */
+    public function testRequestWithoutContentTypeAndContentLength()
+    {
+        $bag = new ServerBag([
+            'CONTENT_TYPE' => '',
+            'CONTENT_LENGTH' => '',
+            'HTTP_USER_AGENT' => 'foo',
+        ]);
+
+        $this->assertEquals(['USER_AGENT' => 'foo'], $bag->getHeaders());
```
OskarStark:
```suggestion
        $this->assertSame(['USER_AGENT' => 'foo'], $bag->getHeaders());
```

src/Symfony/Component/HttpFoundation/Tests/ServerBagTest.php
```diff
@@ -177,4 +177,20 @@ public function testItDoesNotOverwriteTheAuthorizationHeaderIfItIsAlreadySet()
...
+    /**
+     * An HTTP request without content-type and content-length will result in
+     * the variables $_SERVER['CONTENT_TYPE'] and $_SERVER['CONTENT_LENGTH']
+     * containing an empty string in PHP.
+     */
+    public function testRequestWithoutContentTypeAndContentLength()
+    {
+        $bag = new ServerBag([
+            'CONTENT_TYPE' => '',
+            'CONTENT_LENGTH' => '',
```
OskarStark, line 190: Content md5 should be added to the testcase

priyadi: If the request does not contain 'Content-MD5', PHP does not set $_SERVER['CONTENT_MD5'] to an empty string, unlike 'Content-Type' and 'Content-Length'. So, I don't think it is necessary.

OskarStark: Good to know, thank you

OskarStark:
It sounds useful to me, but I am not sure what the spec is here (cc @stof).
Additionally it feels more like a new feature, because applications could rely on the current behavior 🤔

94noni:
Perhaps add a deprecation layer for previous version and change for next?
Even if it seems not a « normal behavior »

priyadi:
@OskarStark @94noni 

I dug deeper and found that it only happens with `php-fpm` SAPI. Under `apache` and `cli-server`, it exhibits the correct behavior: both CONTENT_TYPE and CONTENT_LENGTH are unset. I did not test the other SAPIs.

The result: https://github.com/priyadi/php-empty-content-type-length-test/blob/main/output.txt

I believe the correct way forward is to treat this as a bug in `php-fpm`. But we'll still need the check in this PR, otherwise we need a specific PHP version requirement, which is a far larger burden for a really minor bug.

We can assume an empty CONTENT_TYPE and CONTENT_LENGTH means the headers are not in the request, as the current state of this PR. Or, we can add a check for `php-fpm` SAPI, and assume the other untested SAPIs are behaving correctly.

I leave the decision to the core devs.

OskarStark:
Ping @symfony/mergers

dunglas:
According to the CGI spec, not set and empty are equivalent and allowed:

>   This specification does not distinguish between zero-length (NULL)
>   values and missing values.  For example, a script cannot distinguish
>   between the two requests http://host/script and http://host/script?
>   as in both cases the QUERY_STRING meta-variable would be NULL.
>
>      meta-variable-value = "" | 1*<TEXT, CHAR or tokens of value>
>
>   An optional meta-variable may be omitted (left unset) if its value is
>   NULL.  Meta-variable values MUST be considered case-sensitive except
>   as noted otherwise.  The representation of the characters in the
>   meta-variables is system-defined; the server MUST convert values to
>   that representation.

https://www.rfc-editor.org/rfc/rfc3875#section-4.1

As PHP inherits from this spec, we should take empty values in consideration.

dunglas (approved):

dunglas:
To me, it's a bug fix as we aren't currently spec-compliant.

priyadi:
@dunglas Do you think we need to extend the same treatment to all headers? i.e. we unset the header if the value is an empty string.

dunglas:
We could apply something like `array_filter()` on the list, yes.

nicolas-grekas:
Thank you @priyadi.

nicolas-grekas:
>  i.e. we unset the header if the value is an empty string.

before doing that, we should verify that there is a SAPI that can report empty strings for headers.

PR #53433, opened 2024-01-05
[DependencyInjection] Use Notifier transport_factory services (without fqcn)

smnandre:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #
| License       | MIT

Follows https://github.com/symfony/symfony/pull/53418 (cc @nicolas-grekas )


OskarStark:
Discoverability for new bridges is much better with the FQCN

smnandre:
It follows the previous PR/discussion, where we realized some of those FQCN were missconfigured since ... a bit of time :) 

Maybe instead we could only use the FQCN then instead of the service id ? 

The main goal was to avoid any situation were we use a map `classA => service` in a file and `classB => service` in another, without ensuring `classA = classB`.

nicolas-grekas:
> Discoverability for new bridges is much better with the FQCN

I think that's a nice argument. Mistakes can happen so while this prevents some kind of mistakes, in might open others.
Closing therefor, thanks for proposing.

PR #53443, opened 2024-01-06
[Validator] deal with fields for which no constraints have been configured

### LINKED ISSUE #53455: [Validator] Breaking changes in Collection

5.4

### Description

Before this https://github.com/symfony/validator/commit/b12535757ba38676de6294e787533bbf087a396b
```
$violations = $this->validator->validate(
            $input,
            new Assert\Collection([
                'channelName' => null,
                'reportType' => null,
                'period' => new Assert\Choice(
                    choices: (new CreateReportScheduleSpecification())->getPeriodAllowableValues(),
                ),
            ]),
        );
```

After this https://github.com/symfony/validator/commit/b12535757ba38676de6294e787533bbf087a396b
```
$violations = $this->validator->validate(
            $input,
            new Assert\Collection([
                'fields' => [
                    'channelName' => null,
                    'reportType' => null,
                    'period' => new Assert\Choice(
                        choices: (new CreateReportScheduleSpecification())->getPeriodAllowableValues(),
                    ),
                ],
            ]),
        );
```

@nicolas-grekas My application stopped working as before after this commit.

### How to reproduce

```
$input = [
            'channelName' => $input->getArgument('channelName'),
            'reportType' => $input->getArgument('reportType'),
            'period' => $input->getArgument('period'),
        ];

$violations = $this->validator->validate(
            $input,
            new Assert\Collection([
                'channelName' => null,
                'reportType' => null,
                'period' => new Assert\Choice(
                    choices: (new CreateReportScheduleSpecification())->getPeriodAllowableValues(),
                ),
            ]),
        );
```

### Possible Solution

_No response_

### Additional Context

_No response_

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix https://github.com/symfony/symfony/pull/53133#issuecomment-1878479736, Fix #53455
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Validator/Constraints/Collection.php
+++ b/src/Symfony/Component/Validator/Constraints/Collection.php
@@ -42,7 +42,7 @@ class Collection extends Composite
      */
     public function __construct($fields = null, array $groups = null, $payload = null, bool $allowExtraFields = null, bool $allowMissingFields = null, string $extraFieldsMessage = null, string $missingFieldsMessage = null)
     {
-        if (\is_array($fields) && ([] === $fields || ($firstField = reset($fields)) instanceof Constraint || ($firstField[0] ?? null) instanceof Constraint)) {
+        if (self::isFieldsOption($fields)) {
             $fields = ['fields' => $fields];
         }
 
@@ -86,5 +86,32 @@ class Collection extends Composite
     protected function getCompositeOption()
     {
         return 'fields';
+    }
+
+    private static function isFieldsOption($options): bool
+    {
+        if (!\is_array($options)) {
+            return false;
+        }
+
+        if ([] === $options) {
+            return true;
+        }
+
+        foreach ($options as $optionOrField) {
+            if ($optionOrField instanceof Constraint) {
+                return true;
+            }
+
+            if (!\is_array($optionOrField)) {
+                return false;
+            }
+
+            if ([] !== $optionOrField && !($optionOrField[0] ?? null) instanceof Constraint) {
+                return false;
+            }
+        }
+
+        return true;
     }
 }
--- a/src/Symfony/Component/Validator/Tests/Constraints/CollectionTest.php
+++ b/src/Symfony/Component/Validator/Tests/Constraints/CollectionTest.php
@@ -170,6 +170,41 @@ class CollectionTest extends TestCase
             'extraFieldsMessage' => 'foo bar baz',
         ]);
 
+        $this->assertSame([], $constraint->fields);
+        $this->assertTrue($constraint->allowExtraFields);
+        $this->assertSame('foo bar baz', $constraint->extraFieldsMessage);
+    }
+
+    public function testEmptyConstraintListFor()
+    {
+        $constraint = new Collection([
+                'foo' => [],
+            ],
+            null,
+            null,
+            true,
+            null,
+            'foo bar baz'
+        );
+
+        $this->assertArrayHasKey('foo', $constraint->fields);
+        $this->assertInstanceOf(Required::class, $constraint->fields['foo']);
+        $this->assertTrue($constraint->allowExtraFields);
+        $this->assertSame('foo bar baz', $constraint->extraFieldsMessage);
+    }
+
+    public function testEmptyConstraintListForFieldInOptions()
+    {
+        $constraint = new Collection([
+            'fields' => [
+                'foo' => [],
+            ],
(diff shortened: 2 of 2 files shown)


src/Symfony/Component/Validator/Constraints/Collection.php
```diff
@@ -87,4 +87,31 @@ protected function getCompositeOption()
...
+        foreach ($options as $optionOrField) {
+            if ($optionOrField instanceof Constraint) {
+                return true;
+            }
+
+            if (!\is_array($optionOrField)) {
+                return false;
+            }
+
+            if ([] !== $optionOrField && !($optionOrField[0] ?? null) instanceof Constraint) {
```
tscni, line 110: In theory this would still break constructions like these
```php
$c = new Collection(
    [
        'foo' => [  // implicit "new Required(...)"
            [
                new Type('string'),
            ],
        ],
    ],
);
$c = new Collection(
    [
        'foo' => [  // implicit "new Required(...)"
            'groups' => ['bar'], // no constraint in first value
            'constraints' => [
                new Type('string'),
            ],
        ],
    ],
);
```

xabbuh: I am not sure I understand what you mean. Both examples where already invalid before, right?

tscni: I meant that they originally worked before https://github.com/symfony/symfony/pull/53133

xabbuh: Are you sure? How were both forms supposed to end up?

tscni: Yes, both work on e.g. 6.4.0. That 0 index access seemed suspicous to me, so I tried to break it a bit.
(Note that I don't care about this working, but thought it worthwhile to point out)

The examples above would turn into
```php
new Collection(
    fields: [
        'foo' => new Required([
            new Type('string'),
        ]),
    ],
);
// more explicit
new Collection(
    fields: [
        'foo' => new Required(
            options:  [
                'constraints' => [
                    new Type('string')
                ],
            ],
            groups: ['bar'],
        ),
    ],
);
```

xabbuh:
Status: Needs work

HypeMC:
@xabbuh Do you need any help finishing this one? I feel responsible cause it was my PR that caused the issues :smile:

xabbuh:
@HypeMC Please do if you have the time for it. :)

src/Symfony/Component/Validator/Constraints/Collection.php
HypeMC: After doing some testing, I don't think it makes sense to support this case, here's why:

```php
$c1 = new Collection(
    [
        'foo' => [
            [
                new Type('string'),
            ],
        ],
    ],
);
$c2 = new Collection(
    [
        'foo' => [
            new Type('string'),
        ],
    ],
);
$c3 = new Collection(
    [
        'foo' => new Type('string'),
    ],
);
```

All three cases result in:

```php
array (
  'foo' => 
  \Symfony\Component\Validator\Constraints\Required::__set_state(array(
     'payload' => NULL,
     'groups' => 
    array (
      0 => 'Default',
    ),
     'constraints' => 
    array (
      0 => 
      \Symfony\Component\Validator\Constraints\Type::__set_state(array(
         'payload' => NULL,
         'groups' => 
        array (
          0 => 'Default',
        ),
         'message' => 'This value should be of type {{ type }}.',
         'type' => 'string',
      )),
    ),
  )),
)
```

The reason why the first example works is because of the following code:

https://github.com/symfony/symfony/blob/ff5dc26d4b4be260fbf74a9a960afa37dd1c18e4/src/Symfony/Component/Validator/Constraints/Collection.php#L69-L77

So when it gets to line 76, it's either `new Required([new Type('string')])` or `new Required(new Type('string'))`, and since `Required` casts non arrays to arrays, the result is the same:

https://github.com/symfony/symfony/blob/ff5dc26d4b4be260fbf74a9a960afa37dd1c18e4/src/Symfony/Component/Validator/Constraints/Composite.php#L64-L66

So my conclusion is that this used to work by accident, and was really not intended to work.

HypeMC:
@xabbuh See #53755

xabbuh:
closing in favor of #53755

PR #53448, opened 2024-01-07
[Cache] Add support for using DSN with `PDOAdapter`

### LINKED ISSUE #52460: [Cache] Using DSN with PDOAdapter

`>=5.4`

### Description

According to [the docs](https://symfony.com/doc/5.4/cache.html) it should be possible to use DSN to configure the PDOAdapter, how an exception is thrown instead:

```
  Unsupported DSN: it does not start with "redis[s]:", "memcached:" nor "couchbase:".
```

This happens because the `AbstractAdapter::createConnection()` supports only certain adapters:

https://github.com/symfony/symfony/blob/58353d12dabe6c5aba30866f8e693d1ce0949126/src/Symfony/Component/Cache/Adapter/AbstractAdapter.php#L127-L144

I'm not sure if this is a bug in the documentation (and possible new feature) or a bug in the code.

### How to reproduce

```yaml
framework:
  cache:
    pools:
      my.cache:
        adapter: cache.adapter.pdo
        provider: "pgsql:host=localhost;port=5432;dbname=postgres;user=postgres;password=pass"
```

```php
#[AsCommand('app:test')]
class TestCommand extends Command
{
    public function __construct(
        #[Autowire(service: 'my.cache')] private CacheInterface $cache,
    ) {
        parent::__construct();
    }

    {
        return Command::SUCCESS;
    }
}
```

### Possible Solution

Add the necessary logic to `AbstractAdapter::createConnection()`.

### Additional Context

_No response_

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #52460
| License       | MIT

Adds support for using DSN with the `PDOAdapter`:

```yaml
framework:
  cache:
    pools:
      my.cache:
        adapter: cache.adapter.pdo
        provider: "pgsql:host=localhost;port=5432;dbname=postgres;user=postgres;password=pass"
```

### DIFF
--- a/src/Symfony/Component/Cache/Adapter/AbstractAdapter.php
+++ b/src/Symfony/Component/Cache/Adapter/AbstractAdapter.php
@@ -124,8 +124,11 @@ abstract class AbstractAdapter implements AdapterInterface, CacheInterface, Logg
 
             return CouchbaseCollectionAdapter::createConnection($dsn, $options);
         }
+        if (preg_match('/^(mysql|oci|pgsql|sqlsrv|sqlite):/', $dsn)) {
+            return PdoAdapter::createConnection($dsn, $options);
+        }
 
-        throw new InvalidArgumentException('Unsupported DSN: it does not start with "redis[s]:", "memcached:" nor "couchbase:".');
+        throw new InvalidArgumentException('Unsupported DSN: it does not start with "redis[s]:", "memcached:", "couchbase:", "mysql:", "oci:", "pgsql:", "sqlsrv:" nor "sqlite:".');
     }
 
     public function commit(): bool
--- a/src/Symfony/Component/Cache/Adapter/PdoAdapter.php
+++ b/src/Symfony/Component/Cache/Adapter/PdoAdapter.php
@@ -89,6 +89,18 @@ class PdoAdapter extends AbstractAdapter implements PruneableInterface
         parent::__construct($namespace, $defaultLifetime);
     }
 
+    public static function createConnection(#[\SensitiveParameter] string $dsn, array $options = []): \PDO|string
+    {
+        if ($options['lazy'] ?? true) {
+            return $dsn;
+        }
+
+        $pdo = new \PDO($dsn);
+        $pdo->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_EXCEPTION);
+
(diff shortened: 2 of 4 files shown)


src/Symfony/Component/Cache/Adapter/PdoAdapter.php
```diff
@@ -89,6 +89,18 @@ public function __construct(#[\SensitiveParameter] \PDO|string $connOrDsn, strin
         parent::__construct($namespace, $defaultLifetime);
     }
 
+    public static function createConnection(#[\SensitiveParameter] string $dsn, array $options = []): \PDO|string
+    {
+        if ($options['lazy'] ?? false) {
+            return $dsn;
```
HypeMC, line 95: I think it makes the most sense to just return the DSN when `lazy` is `true` since the adapter itself supports this and creates the connection lazily.

src/Symfony/Component/Cache/Adapter/AbstractAdapter.php
```diff
@@ -124,8 +124,17 @@ public static function createConnection(#[\SensitiveParameter] string $dsn, arra
 
             return CouchbaseCollectionAdapter::createConnection($dsn, $options);
         }
+        if (
+            str_starts_with($dsn, 'mysql:')
```
stloyd: Wouldn't preg_match be faster than 5 calls?

HypeMC: Good idea, done

src/Symfony/Component/Cache/Adapter/PdoAdapter.php
nicolas-grekas: this should be the default to me

stloyd (approved):

src/Symfony/Component/Cache/Adapter/PdoAdapter.php
HypeMC: Done

nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @HypeMC.

PR #53450, opened 2024-01-07
[FrameworkBundle] deprecate not passing a build dir when warming up the router cache

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Issues        | 
| License       | MIT

If I don't miss anything, #52962 went a bit too far introducing a BC break of the current behaviour if no build dir is passed to the `warmUp()` method.

### DIFF
--- a/UPGRADE-7.1.md
+++ b/UPGRADE-7.1.md
@@ -9,6 +9,7 @@ Cache
 FrameworkBundle
 ---------------
 
+ * Deprecate not passing a build dir to `RouterCacheWarmer::warmUp()` and `Router::warmUp()`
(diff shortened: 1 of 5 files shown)


nicolas-grekas:
Did you encounter any real world issue on the topic?
Because passing null is an expected case: when cache:warmup is called, there is no buildDir to pass because it's too late to write there.

xabbuh:
After discussing with Nicolas I close here for now. All these classes are wired by the FrameworkBundle so the changes will probably have no real affect to users of the classes.

PR #53452, opened 2024-01-07
pin nikic/php-parser to version 4

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -134,6 +134,7 @@
         "masterminds/html5": "^2.6",
         "monolog/monolog": "^1.25.1|^2",
         "nyholm/psr7": "^1.0",
+        "nikic/php-parser": "^4.13",
         "pda/pheanstalk": "^4.0",
         "php-http/httplug": "^1.0|^2.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -134,6 +134,7 @@
         "masterminds/html5": "^2.6",
         "monolog/monolog": "^1.25.1|^2",
         "nyholm/psr7": "^1.0",
+        "nikic/php-parser": "^4.13",
```
xabbuh, line 137: before, the classes were only present as PHPUnit installed them

nicolas-grekas:
Do we need this PR? Where do we reference php-parser in 5.4?

xabbuh:
Nope, we indeed don't need this PR. I mistakenly assumed that the `PhpAstExtractor` was already present in 5.4 when preparing the changes for #53453. So this PR should have targeted 6.3 but is obsolet now that #53453 is based upon 6.3.

PR #53453, opened 2024-01-07
[Translation] fix support for nikic/php-parser 5.0

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -140,6 +140,7 @@
         "league/uri": "^6.5|^7.0",
         "masterminds/html5": "^2.7.2",
         "monolog/monolog": "^1.25.1|^2",
+        "nikic/php-parser": "^4.16|^5.0",
         "nyholm/psr7": "^1.0",
         "pda/pheanstalk": "^4.0",
         "php-http/discovery": "^1.15",
--- a/src/Symfony/Component/Translation/Extractor/PhpAstExtractor.php
+++ b/src/Symfony/Component/Translation/Extractor/PhpAstExtractor.php
@@ -39,7 +39,11 @@ final class PhpAstExtractor extends AbstractFileExtractor implements ExtractorIn
             throw new \LogicException(sprintf('You cannot use "%s" as the "nikic/php-parser" package is not installed. Try running "composer require nikic/php-parser".', static::class));
         }
 
-        $this->parser = (new ParserFactory())->create(ParserFactory::PREFER_PHP7);
+        if (method_exists(ParserFactory::class, 'createForHostVersion')) {
+            $this->parser = (new ParserFactory())->createForHostVersion();
+        } else {
+            $this->parser = (new ParserFactory())->create(ParserFactory::PREFER_PHP7);
+        }
     }
 
     public function extract(iterable|string $resource, MessageCatalogue $catalogue): void
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -128,6 +128,10 @@ jobs:
           echo SYMFONY_REQUIRE=">=$([ '${{ matrix.mode }}' = low-deps ] && echo 4.4 || echo $SYMFONY_VERSION)" >> $GITHUB_ENV
           [[ "${{ matrix.mode }}" = *-deps ]] && mv composer.json.phpunit composer.json || true
 
+          if [[ "${{ matrix.mode }}" = low-deps ]]; then
+            echo SYMFONY_PHPUNIT_REQUIRE="nikic/php-parser:^4.16" >> $GITHUB_ENV
+          fi
+
       - name: Install dependencies
         run: |
           echo "::group::composer update"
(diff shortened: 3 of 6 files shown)


nicolas-grekas:
What about merging this on 5.4? This would reduce the pressure on deps constraints.

xabbuh:
Not sure, this is not a bugfix to me.

nicolas-grekas:
We also merge deps-related changed to 5.4: support for latest Doctrine, latest PHP, etc. This is similar to me. This makes Symfony a low contributor to deps-hell (aka makes upgrading easier)

GromNaN:
If it's compatible with 5.4, I don't see any reason to not allowing this new version.

xabbuh:
fair enough, rebased on `6.3` as the `PhpAstExtractor` class was introduced in 6.2

chalasr (approved):

src/Symfony/Component/Translation/Extractor/PhpAstExtractor.php
```diff
@@ -39,7 +40,11 @@ public function __construct(
             throw new \LogicException(sprintf('You cannot use "%s" as the "nikic/php-parser" package is not installed. Try running "composer require nikic/php-parser".', static::class));
         }
 
-        $this->parser = (new ParserFactory())->create(ParserFactory::PREFER_PHP7);
+        if (method_exists(ParserFactory::class, 'createForVersion')) {
+            $this->parser = (new ParserFactory())->createForVersion(PhpVersion::fromString('8.2'));
```
nicolas-grekas: I'm wondering how we will maintain this version number.
Wild idea: add `// @requires PHP 8.2` so that grepping for test cases can yield this when preparing 8.0?

stof: why hardcoding PHP 8.2 here ? Shouldn't we parse for the current PHP version instead ?

xabbuh: Not sure either, but one possible solution would be to use `createForHostVersion()` instead which will simply use the PHP version the process is executed with. Probably that's also how people most often use this extractor anyway.

xabbuh: replaced with `createForHostVersion()` as this is also basically what @stof asks for in https://github.com/symfony/symfony/pull/53453#discussion_r1444332797

xabbuh: replaced this to make use of `createForHostVersion()` instead

src/Symfony/Component/Translation/Extractor/PhpAstExtractor.php
```diff
@@ -39,7 +39,11 @@ public function __construct(
             throw new \LogicException(sprintf('You cannot use "%s" as the "nikic/php-parser" package is not installed. Try running "composer require nikic/php-parser".', static::class));
         }
 
-        $this->parser = (new ParserFactory())->create(ParserFactory::PREFER_PHP7);
+        if (method_exists(ParserFactory::class, 'createForVersion')) {
```
stof: I suggest checking for the `createForHostVersion` method as that's the one being used.

wouterj:
Note that currently, we explicitly say that this is *not* allowed as a bug fix in the [maintenance docs](https://symfony.com/doc/current/contributing/code/maintenance.html):

> [...] the following changes are never accepted in a patch version:
> 
> - **Support for newer major versions of Composer dependencies**: Taking into account support for newer versions of an existing dependency is not acceptable.

nicolas-grekas:
@wouterj we have some contradiction on the topic, because we also have this "seamless deps" rule that makes us fix compat with Doctrine / PHP / Windows / infra-level stuff on lower branches, the reason being that wider deps ranges makes it easier to upgrade for end users.
Here, we fall in this categry to me, *especially* because we don't set the version boundaries. But also because php-parser is a widely used deps and making Symfony restrict which versions are allowed is contributing to deps hell.

wouterj:
Yes, "Newer versions of PHP" and "Newer versions of popular OSes" are mentioned in the "allowed on a case by case basis" section :)

Btw, not saying I'm against this change, especially as we don't set version ranges explicitly for this optional dep. But I think we should then update our policy on this (and find the reason why we set the current policy like this, is there something non-obvious that bit us in the past and we shouldn't forget now?)

composer.json
```diff
@@ -140,6 +140,7 @@
         "league/uri": "^6.5|^7.0",
         "masterminds/html5": "^2.7.2",
         "monolog/monolog": "^1.25.1|^2",
+        "nikic/php-parser": "^4.16|^5.0",
```
alexandre-daubois, line 143: Shouldn't this be `^4.13` like in the composer.json of the component?

xabbuh: Thanks for the heads up. The dependency should have been bumped in the component's `composer.json` file too thus allowing for the `getParts()` method to be used.

xabbuh:
tests pass, ready to be reviewed

nicolas-grekas:
Thank you @xabbuh.

Issue #44943, opened 2022-01-07
[Messenger] ErrorDetailsStamp +RedeliveryStamp can lead to an oversize message for SQS


magnetik:
### Symfony version(s) affected

5.4

### Description

SQS has a limit of 262,144 bytes (256 KB) message size. 

When there is an error during message handling, the `ErrorDetailsStamp` adds a stacktrace that can lead to a message bigger than this limit. 

### How to reproduce

* Dispatch messages with SQS with a redelivery policy that makes it retry a few times
* Throw an exception with a super big stack trace
* See it fails

### Possible Solution

Not sure, maybe an option to remove stack trace from `ErrorDetailsStamp`

Jean85:
:+1: on this one. I had my whole worker setup crash in a loop due to this. 

This is super critical, I had to:
 * scale the workers x5 to mitigate the issue
 * troubleshoot blindly to find the issue that was getting the message fail
 * deploy an hotfix

The failure is completely out of my code, all inside the framework, so I had no way to understand easily which message it was, even with Sentry:
```
AsyncAws\Core\Exception\Http\ClientException: HTTP 400 returned for "https://sqs.eu-central-1.amazonaws.com/".

Code:    InvalidParameterValue
Message: One or more parameters are invalid. Reason: Message must be shorter than 262144 bytes.
Type:    Sender
Detail:  

#22 /vendor/async-aws/core/src/Response.php(406): AsyncAws\Core\Response::AsyncAws\Core\{closure}
#21 /vendor/async-aws/core/src/Response.php(423): AsyncAws\Core\Response::getResolveStatus
#20 /vendor/async-aws/core/src/Response.php(160): AsyncAws\Core\Response::resolve
#19 /vendor/async-aws/core/src/Response.php(105): AsyncAws\Core\Response::__destruct
#18 /vendor/symfony/amazon-sqs-messenger/Transport/Connection.php(363): Symfony\Component\Messenger\Bridge\AmazonSqs\Transport\Connection::send
#17 /vendor/symfony/amazon-sqs-messenger/Transport/AmazonSqsSender.php(65): Symfony\Component\Messenger\Bridge\AmazonSqs\Transport\AmazonSqsSender::send
#16 /vendor/symfony/amazon-sqs-messenger/Transport/AmazonSqsTransport.php(81): Symfony\Component\Messenger\Bridge\AmazonSqs\Transport\AmazonSqsTransport::send
[...]

PR #53454, opened 2024-01-07
[Messenger][FrameworkBundle] Add option to exclude stack trace from `ErrorDetailsStamp`

### LINKED ISSUE #45944: [Messenger] Growing size of ErrorDetailsStamp breaks AMQP for RecoverableMessageHandlingException

6.0.3

### Description

After the 4th or 5th retry the `ErrorDetailsStamp` header of a message that was sent back to the queue using the `RecoverableMessageHandlingException` grows to **60k** and RabbitMQ rejects the message with the error 

`Library error: table too large for buffer`

I did find a related and merged issue (https://github.com/symfony/symfony/pull/34082) but this does not seem to have solved it, there are recent comments from people who ran into the same issue.

I am able to solve this in using a custom serializer that removes the `ErrorDetailsStamp` but I think that this is not expected behaviour.

This also has implications for the other transports like Doctrine where very large entries for simple messages are being inserted into the databae.

### How to reproduce

Just create a TestMessage and watch the `ErrorDetailsStamp` header grow out of proportion:

```
final class TestMessageHandler .....

public function __invoke(TestMessage $message): int
{
  throw new RecoverableMessageHandlingException('Recovered at '.time());
}
```

### Possible Solution

Create a custom serializer and remove the `ErrorDetailsStamp` in the encode method:

```
class WorkaroundSerializer ...

public function encode(Envelope $envelope): array
{
.....
  $filteredStamps = [];
  $envelopeStamps = $envelope->all();

  foreach ($envelopeStamps as $stamps) {
    foreach ($stamps as $stamp) {
      if (!$stamp instanceof ErrorDetailsStamp) {
        $filteredStamps[] = $stamp;
      }
     }
   }

  return ... 
}
```

I guess another solution is to not even add the `ErrorDetailsStamp` to a user initiated recoverable message as it is not really an error but intended behaviour? Or make headers configurable, eg `traceCount: 1` to only add the latest trace or none with 0?

Does it make logical sense that this is an Exception and not a stamp? With it being an Exception although not being an error it also has to be removed from monitoring tools (Sentry etc.) to not be a false negative.

### Additional Context

_No response_

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 7.3
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #45944, #44943
| License       | MIT

Some transports, in my case Beanstalk, don't like the large payloads that can result from the entire stack trace being in the `ErrorDetailsStamp`.

The idea of this PR is to allow excluding the stack trace through an option. The user could then optionally add a custom listener to log the stack trace to alternative storage.


src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1575,6 +1575,9 @@ function ($a) {
                                         ->defaultNull()
                                         ->info('Transport name to send failed messages to (after all retries have failed).')
                                     ->end()
+                                    ->booleanNode('include_stack_trace_in_error')
```
HypeMC, line 1674: I'm not too crazy about the name of the option, maybe someone has a better suggestion.

src/Symfony/Component/Messenger/EventListener/AddErrorDetailsStampListener.php
```diff
@@ -17,9 +17,16 @@
 
 final class AddErrorDetailsStampListener implements EventSubscriberInterface
 {
+    /**
+     * @param array<string, bool> $includeStackTrace
+     */
+    public function __construct(private readonly array $includeStackTrace = [])
```
OskarStark:
```suggestion
    public function __construct(
private readonly array $includeStackTrace = [])
```
On new line please

HypeMC: Done

odombrovskyi-dev:
Hi, thanks for preparing the fix for this issue. I have a few concerns regarding the solution though. If I get it right, with this option the message will lose pretty important info about an error.

From my understanding, the issue is not only in the big stack trace but also in the fact that the exception contains a previous exception and this previous exception has another previous exception. Eventually, after flattening the exception we could end up in having the stack trace processed from multiple exceptions, and for me this is what we should try to fix. Instead of flattening all previous exceptions, we should flatten only a current exception. This will allow us to keep the stack trace of the "latest" exception that will be useful to identify the error.

HypeMC:
@odombrovskyi-dev Hi, as mentioned in the description, the idea here is to allow the user to implement their own listener and log the exception however they see fit. For example, they can log the entire error into Logstash and add a custom stamp that serves as an identifier connecting the message to the log. I guess I could remove the stack trace and just leave the exception, but that might prove useless in some cases, as a certain exception might get thrown from different places.

mikepage:
We have introduced a decorated transport for amqp to prevent "Invalid AMQP data"
https://github.com/recranet/messenger-amqp-decorated/blob/master/src/MessengerAmqpDecoratedTransport.php

mikepage:
I would be cleaner to just omit the entire ErrorDetailsStamp, not just the stack trace. When GuzzleException is added to ErrorDetails with large body/content it should also be omitted when size is gt 1024 bytes.

nicolas-grekas:
Please have a look at #63288 instead (I don't like adding options :) )

PR #53464, opened 2024-01-09
[Messenger] Improve Redis integration tests

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

This kind of failure can happen when cleaning Redis in the `finally` clause:

image

Unlinking will trigger the cleaning in another process, thus it won't make the test fail if a problem is triggered. I think this is desired so we don't have false negatives.

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Redis/Tests/Transport/RedisExtIntegrationTest.php
+++ b/src/Symfony/Component/Messenger/Bridge/Redis/Tests/Transport/RedisExtIntegrationTest.php
@@ -257,7 +257,7 @@ class RedisExtIntegrationTest extends TestCase
             ], $message['data']);
             $connection->reject($message['id']);
         } finally {
-            $redis->del('messenger-lazy');
+            $redis->unlink('messenger-lazy');
         }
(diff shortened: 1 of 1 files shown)


src/Symfony/Component/Messenger/Bridge/Redis/Tests/Transport/RedisExtIntegrationTest.php
```diff
@@ -295,14 +295,13 @@ public function testJsonError()
...
         } finally {
-            $redis->del('messenger-json-error');
+            $redis->unlink('messenger-json-error');
         }
     }
 
     public function testGetNonBlocking()
     {
         $redis = new \Redis();
-
```
alexandre-daubois, line 305: Code-style to be consistent with the rest

xabbuh (approved):
LGTM

nicolas-grekas:
Thank you @alexandre-daubois.

PR #53466, opened 2024-01-09
Add `SecretsRevealCommand`

danielburger1337:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | n/a
| License       | MIT

Add a new command to reveal a specific secrets value. The output is able to be piped into other commands. 

Docs PR will be submitted if this is a desired feature.

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
+++ b/src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
@@ -9,6 +9,7 @@ CHANGELOG
  * Move the Router `cache_dir` to `kernel.build_dir`
  * Deprecate the `router.cache_dir` config option
  * Add `rate_limiter` tags to rate limiter services
+ * Add `secrets:reveal` command
 
 7.0
 ---
--- a/src/Symfony/Bundle/FrameworkBundle/Command/SecretsRevealCommand.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Command/SecretsRevealCommand.php
@@ -1,0 +1,72 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\FrameworkBundle\Command;
+
+use Symfony\Bundle\FrameworkBundle\Secrets\AbstractVault;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\ConsoleOutputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+/**
+ * @internal
+ */
+#[AsCommand(name: 'secrets:reveal', description: 'Reveal the value of a secret')]
+final class SecretsRevealCommand extends Command
+{
+    public function __construct(
+        private readonly AbstractVault $vault,
+        private readonly ?AbstractVault $localVault = null,
+    ) {
+        parent::__construct();
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->addArgument('name', InputArgument::REQUIRED, 'The name of the secret to reveal', null, fn () => array_keys($this->vault->list()))
+            ->setHelp(<<<'EOF'
+The <info>%command.name%</info> command reveals a stored secret.
+
+    <info>%command.full_name%</info>
+EOF
+            )
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output instanceof ConsoleOutputInterface ? $output->getErrorOutput() : $output);
+
+        $secrets = $this->vault->list(true);
+        $localSecrets = $this->localVault?->list(true);
+
+        $name = (string) $input->getArgument('name');
+
+        if (null !== $localSecrets && \array_key_exists($name, $localSecrets)) {
+            $io->writeln($localSecrets[$name]);
+        } else {
+            if (!\array_key_exists($name, $secrets)) {
+                $io->error(sprintf('The secret "%s" does not exist.', $name));
+
+                return self::INVALID;
+            }
+
+            $io->writeln($secrets[$name]);
+        }
+
+        return self::SUCCESS;
+    }
+}
--- a/src/Symfony/Bundle/FrameworkBundle/Tests/Command/SecretsRevealCommandTest.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Tests/Command/SecretsRevealCommandTest.php
@@ -1,0 +1,86 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\FrameworkBundle\Tests\Command;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Bundle\FrameworkBundle\Command\SecretsRevealCommand;
+use Symfony\Bundle\FrameworkBundle\Secrets\AbstractVault;
+use Symfony\Bundle\FrameworkBundle\Secrets\DotenvVault;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Tester\CommandTester;
+
+class SecretsRevealCommandTest extends TestCase
+{
+    public function testExecute()
+    {
+        $vault = $this->createMock(AbstractVault::class);
+        $vault->method('list')->willReturn(['secretKey' => 'secretValue']);
+
+        $command = new SecretsRevealCommand($vault);
+
+        $tester = new CommandTester($command);
+        $this->assertSame(Command::SUCCESS, $tester->execute(['name' => 'secretKey']));
+
+        $this->assertEquals('secretValue', trim($tester->getDisplay(true)));
+    }
+
+    public function testInvalidName()
+    {
+        $vault = $this->createMock(AbstractVault::class);
+        $vault->method('list')->willReturn(['secretKey' => 'secretValue']);
+
+        $command = new SecretsRevealCommand($vault);
+
+        $tester = new CommandTester($command);
+        $this->assertSame(Command::INVALID, $tester->execute(['name' => 'undefinedKey']));
+
+        $this->assertStringContainsString('The secret "undefinedKey" does not exist.', trim($tester->getDisplay(true)));
+    }
+
+    /**
+     * @backupGlobals enabled
+     */
+    public function testLocalVaultOverride()
+    {
+        $vault = $this->createMock(AbstractVault::class);
+        $vault->method('list')->willReturn(['secretKey' => 'secretValue']);
+
+        $_ENV = ['secretKey' => 'newSecretValue'];
+        $localVault = new DotenvVault('/not/a/path');
+
+        $command = new SecretsRevealCommand($vault, $localVault);
+
+        $tester = new CommandTester($command);
+        $this->assertSame(Command::SUCCESS, $tester->execute(['name' => 'secretKey']));
+
+        $this->assertEquals('newSecretValue', trim($tester->getDisplay(true)));
+    }
+
+    /**
+     * @backupGlobals enabled
+     */
+    public function testOnlyLocalVaultContainsName()
+    {
+        $vault = $this->createMock(AbstractVault::class);
+        $vault->method('list')->willReturn(['otherKey' => 'secretValue']);
+
+        $_ENV = ['secretKey' => 'secretValue'];
+        $localVault = new DotenvVault('/not/a/path');
+
+        $command = new SecretsRevealCommand($vault, $localVault);
+
+        $tester = new CommandTester($command);
+        $this->assertSame(Command::SUCCESS, $tester->execute(['name' => 'secretKey']));
+
+        $this->assertEquals('secretValue', trim($tester->getDisplay(true)));
+    }
+}
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
@@ -1773,6 +1773,7 @@ class FrameworkExtension extends Extension
         if (!$this->readConfigEnabled('secrets', $container, $config)) {
             $container->removeDefinition('console.command.secrets_set');
             $container->removeDefinition('console.command.secrets_list');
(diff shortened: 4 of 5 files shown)


src/Symfony/Bundle/FrameworkBundle/Command/SecretsListCommand.php
```diff
@@ -70,6 +74,12 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         }
 
         $secrets = $this->vault->list($reveal);
+
+        $name = $input->getArgument('name');
+        if (null !== $name) {
```
nicolas-grekas:
```suggestion
        if (null !== $name = $input->getArgument('name')) {
```

nicolas-grekas (approved):

xabbuh (approved):

Nyholm (approved):
This is a great PR. Thank you.

src/Symfony/Bundle/FrameworkBundle/Command/SecretsListCommand.php
```diff
@@ -70,6 +74,11 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         }
 
         $secrets = $this->vault->list($reveal);
+
+        if (null !== $name = $input->getArgument('name')) {
+            $secrets = \array_key_exists($name, $secrets) ? [$name => $secrets[$name]] : [];
```
94noni: perhaps unwanted, but what about leveraging `levenshtein` logic here?

imagine `secrets:list APP_SCRET` => may output `APP_SECRET` ?

Nyholm: Hm. We could maybe do that to implement "did you mean ...?"

But let's do that in a different PR.

stof:
should the command fail when passing a name and there is no secret with that name ?

src/Symfony/Bundle/FrameworkBundle/Command/SecretsListCommand.php
```diff
@@ -45,12 +46,15 @@ public function __construct(AbstractVault $vault, AbstractVault $localVault = nu
     protected function configure(): void
     {
         $this
+            ->addArgument('name', InputArgument::OPTIONAL, 'The name of the secret to return')
```
GromNaN: You can add value completion with the list of known list of names.

danielburger1337: I don't quite get what you mean. Do you mean:

```php
protected function configure(): void
{
    $this
        ->addArgument('name', InputArgument::OPTIONAL, 'The name of the secret to return', null, \array_keys($this->vault->list()))

**Comment by danielburger1337** [comment]
> should the command fail when passing a name and there is no secret with that name ?

Thats one way to do it. I just reused the current behavior when the secret store is empty.

If we go your way, we should refactor a bit so that the comments about how to reference/reveal a secret are not displayed when an invalid name is passed.

```php
    {
        $io = new SymfonyStyle($input, $output instanceof ConsoleOutputInterface ? $output->getErrorOutput() : $output);

        $reveal = $input->getOption('reveal');

        $secrets = $this->vault->list($reveal);

        if (null !== $name = $input->getArgument('name')) {
            if (!\array_key_exists($name, $secrets)) {
                $io->error(\sprintf('The secret "%s" does not exist.', $name));

                return self::INVALID;
            }

            $secrets = [$name => $secrets[$name]];
        }

        $localSecrets = $this->localVault?->list($reveal);

        $io->comment('Use <info>"%env(<name>)%"</info> to reference a secret in a config file.');

        if (!$reveal) {
            $io->comment(\sprintf('To reveal the secrets run <info>php %s %s --reveal</info>', $_SERVER['PHP_SELF'], $this->getName()));
        }
```

src/Symfony/Bundle/FrameworkBundle/Command/SecretsListCommand.php
94noni: related to this for example https://github.com/symfony/symfony/blob/4328f5b16852b8daefc957d370c08ee0bfbf3dbc/src/Symfony/Bundle/FrameworkBundle/Command/SecretsRemoveCommand.php#L84

https://symfony.com/blog/new-in-symfony-5-4-console-autocompletion

danielburger1337: Isn't that in the end identical to what I suggested? By default the `complete` method uses the last argument passed to addArgument/addOption and adds them as suggested values. We don't need any special logic here for what secret names can be suggested or not.

GromNaN: Yes @danielburger1337 your proposition is better as it's the new recommended way of doing it.

```php
    $this
        ->addArgument('name', InputArgument::OPTIONAL, 'The name of the secret to return', null, fn() => \array_keys($this->vault->list()))
```

A closure should be used for lazy loading.

chalasr:
> If we go your way, we should refactor a bit so that the comments about how to reference/reveal a secret are not displayed when an invalid name is passed.

Sounds good to me

fabpot (commented):
Would allowing more than one name to get a subset of the list make sense? It would be more consistent with the command name.

danielburger1337:
> Would allowing more than one name to get a subset of the list make sense? It would be more consistent with the command name.

I agree that this would fit better with the command name, but how would you handle passing a valid AND an invalid secret name? Only return an error? Return the valid secret and an error message? Would this result in exit code 0, 1 or 2?

chalasr:
>  It would be more consistent with the command name.

Good point. Given the number of questions that supporting multiple names raise, what about adding a `secrets:show` command instead? 
Otherwise, I would just make the command fail entirely when an invalid name is passed.

fabpot:
> > It would be more consistent with the command name.
> 
> Good point. Given the number of questions that supporting multiple names raise, what about adding a `secrets:show` command instead? Otherwise, I would just make the command fail entirely when an invalid name is passed.

That's maybe a better option indeed.

nicolas-grekas:
> secrets:show

Or `secrets:reveal`  to use the existing wording?
but I agree with the proposal. That could be even more useful since this wouldn't require decorating the output with fancy layout. Instead, the result could be directly piped to another command when needed.

chalasr:
👍 for `reveal`

fabpot:
@danielburger1337 Are you still interested in moving this PR forward?

danielburger1337:
> @danielburger1337 Are you still interested in moving this PR forward?

Yes I am. Just to make sure, you guys want to add the new SecretsRevealCommand. This would entail removing the added "name" argument from the SecretsListCommand again and then the "-r" option will be deprecated?

nicolas-grekas:
I wouldn't deprecated -r, but yes for the rest

danielburger1337:
Updated title and description to match new proposed idea.

src/Symfony/Bundle/FrameworkBundle/Command/SecretsRevealCommand.php
```diff
@@ -0,0 +1,72 @@
...
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\ConsoleOutputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+/**
+ * @internal
+ */
+#[AsCommand(name: 'secrets:reveal', description: 'Reveal a secrets value')]
```
94noni:
```suggestion
#[AsCommand(name: 'secrets:reveal', description: 'Reveal the value of a secret')]
```

src/Symfony/Bundle/FrameworkBundle/Command/SecretsRevealCommand.php
```diff
@@ -0,0 +1,72 @@
...
+    ) {
+        parent::__construct();
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->addArgument('name', InputArgument::REQUIRED, 'The name of the secret to reveal', null, fn () => array_keys($this->vault->list()))
+            ->setHelp(<<<'EOF'
+The <info>%command.name%</info> command reveals all stored secrets.
```
94noni:
```suggestion
The <info>%command.name%</info> command reveals a stored secret.
```

94noni (approved):
nice one, small comments regarding desc

src/Symfony/Bundle/FrameworkBundle/Command/SecretsRevealCommand.php
danielburger1337: This is a great example why you shouldn't copy/paste your old code 😅

Fixed

src/Symfony/Bundle/FrameworkBundle/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ CHANGELOG
  * Move the Router `cache_dir` to `kernel.build_dir`
  * Deprecate the `router.cache_dir` config option
  * Add `rate_limiter` tags to rate limiter services
+ * Add `SecretsRevealCommand`
```
fabpot:
```suggestion
 * Add `secrets:reveal` command
```

src/Symfony/Bundle/FrameworkBundle/Command/SecretsRevealCommand.php
```diff
@@ -0,0 +1,72 @@
...
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\ConsoleOutputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+/**
+ * @internal
+ */
+#[AsCommand(name: 'secrets:reveal', description: 'Reveal the value of a secrets')]
```
fabpot:
```suggestion
#[AsCommand(name: 'secrets:reveal', description: 'Reveal the value of a secret')]
```

fabpot (approved):

src/Symfony/Bundle/FrameworkBundle/Tests/Command/SecretsRevealCommandTest.php
```diff
@@ -0,0 +1,86 @@
...
+use PHPUnit\Framework\TestCase;
+use Symfony\Bundle\FrameworkBundle\Command\SecretsRevealCommand;
+use Symfony\Bundle\FrameworkBundle\Secrets\AbstractVault;
+use Symfony\Bundle\FrameworkBundle\Secrets\DotenvVault;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Tester\CommandTester;
+
+class SecretsRevealCommandTest extends TestCase
+{
+    public function testExecute(): void
```
fabpot:
```suggestion
    public function testExecute()
```

src/Symfony/Bundle/FrameworkBundle/Tests/Command/SecretsRevealCommandTest.php
```diff
@@ -0,0 +1,86 @@
...
+
+        $command = new SecretsRevealCommand($vault);
+
+        $tester = new CommandTester($command);
+        $this->assertSame(Command::SUCCESS, $tester->execute(['name' => 'secretKey']));
+
+        $this->assertEquals('secretValue', trim($tester->getDisplay(true)));
+    }
+
+    public function testInvalidName(): void
```
fabpot:
```suggestion
    public function testInvalidName()
```

src/Symfony/Bundle/FrameworkBundle/Tests/Command/SecretsRevealCommandTest.php
```diff
@@ -0,0 +1,86 @@
...
+        $tester = new CommandTester($command);
+        $this->assertSame(Command::INVALID, $tester->execute(['name' => 'undefinedKey']));
+
+        $this->assertStringContainsString('The secret "undefinedKey" does not exist.', trim($tester->getDisplay(true)));
+    }
+
+    /**
+     * @backupGlobals enabled
+     */
+    public function testLocalVaultOverride(): void
```
fabpot:
```suggestion
    public function testLocalVaultOverride()
```

src/Symfony/Bundle/FrameworkBundle/Tests/Command/SecretsRevealCommandTest.php
```diff
@@ -0,0 +1,86 @@
...
+        $tester = new CommandTester($command);
+        $this->assertSame(Command::SUCCESS, $tester->execute(['name' => 'secretKey']));
+
+        $this->assertEquals('newSecretValue', trim($tester->getDisplay(true)));
+    }
+
+    /**
+     * @backupGlobals enabled
+     */
+    public function testOnlyLocalVaultContainsName(): void
```
fabpot:
```suggestion
    public function testOnlyLocalVaultContainsName()
```

fabpot (changes requested):
Can you rebase to get rid of the merge commit?

fabpot:
Thank you @danielburger1337.

danielburger1337:
@fabpot All done. Sorry for the extra work on your part.

Is there any particular reason why you guys don't add the "void" return type to test cases?

src/Symfony/Bundle/FrameworkBundle/Command/SecretsRevealCommand.php
```diff
@@ -0,0 +1,72 @@
...
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\ConsoleOutputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+/**
+ * @internal
```
OskarStark, line 24: Is this needed when the class is final?

danielburger1337: I have no idea. I just copied it from SecretsListCommand (as this command is based on it). 

I don't think it makes much sense on either.

chalasr: @OskarStark That gives us more flexibility 👍 as there is no need to use that class from userland, only the command itself (via CLI) matters

OskarStark: Sorry I don't get what you mean.

I can do absolutely the same without the internal tag, the class is final and I can only execute the command in userland

xabbuh: I guess what @chalasr means is that we could also rename this class or move it to another namespace without having to worry about consumers of it. Not sure if we really need that flexibility though tbh.

OskarStark: Ah ok got it

chalasr: > Not sure if we really need that flexibility though tbh.

👍 Thinking about it, I agree we should not fall in being too much defensive in general.

PR #53472, opened 2024-01-09
[HttpClient] Add JSON flags support

den-kuz:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

This PR adds support for passing flags to json_encode function to control encoding json data when sending http requests via http client

### DIFF
--- a/src/Symfony/Component/HttpClient/CHANGELOG.md
+++ b/src/Symfony/Component/HttpClient/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
  * Add `HttpOptions::setHeader()` to add or replace a single header
  * Allow mocking `start_time` info in `MockResponse`
  * Add `MockResponse::fromFile()` and `JsonMockResponse::fromFile()` methods to help using fixtures files
+ * Add `json_flags` to request options to control encoding json data
 
 7.0
 ---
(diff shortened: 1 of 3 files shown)


stof:
what is the use case for that ?

den-kuz:
> what is the use case for that ?

Current private static jsonEncode method hardcodes the json encoding standard with flags and there is no way to control it at this moment:

```php
$flags ??= \JSON_HEX_TAG | \JSON_HEX_APOS | \JSON_HEX_AMP | \JSON_HEX_QUOT | \JSON_PRESERVE_ZERO_FRACTION;

try {
    $value = json_encode($value, $flags | \JSON_THROW_ON_ERROR, $maxDepth);
} catch (\JsonException $e) {
    throw new InvalidArgumentException('Invalid value for "json" option: '.$e->getMessage());
}

return $value;
```

in our project there is a need to send json to an external service with data containing the characters ">" and "<", which are not encoded as we expect

stof:
If the external services uses a JSON parser that respects the JSON specification, those flags should not be an issue as they produce standard-compliant JSON.
And if you need to interact with an API that actually expects a subset of JSON instead of JSON itself, you still have the ability to encore the data yourselves by using the `body` option instead of the `json` one (and then setting yourselves the `Content-Type` expected by the server).

I'm not convinced this use case justifies the addition of a new option here, which we then need to maintain **and document**.

nicolas-grekas:
I agree with @stof 
Closing therefor, thanks for proposing.

PR #53480, opened 2024-01-10
Allow monolog 3

DavidPrevot:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | N/A
| License       | MIT

This is a follow up from #52222.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -139,7 +139,7 @@
         "league/html-to-markdown": "^5.0",
         "league/uri": "^6.5|^7.0",
         "masterminds/html5": "^2.7.2",
-        "monolog/monolog": "^1.25.1|^2",
+        "monolog/monolog": "^1.25.1|^2|^3",
         "nikic/php-parser": "^4.18|^5.0",
         "nyholm/psr7": "^1.0",
         "pda/pheanstalk": "^4.0",


xabbuh:
Are you sure about this? The PHP code in #52222 was adjusted for 6.3 while your PR targets 5.4.

DavidPrevot:
> Are you sure about this? The PHP code in #52222 was adjusted for 6.3 while your PR targets 5.4.

You’re right, thanks a lot, PR updated.

derrabus:
Can you elaborate which problem this change is going to solve? If we merge this PR, we will lose test coverage for Monolog 2.

DavidPrevot:
> Can you elaborate which problem this change is going to solve?

This change should allow one to install `symfony/symfony` with version 3 of `monolog/monolog`. It aims to update the `symfony/symfony` “metapackage”: since it replaces (among others) `symfony/monolog-bridge` that requires `monolog/monolog` in version `^1.25.1|^2|^3`, it shouldn’t restrict the version required to `^1.25.1|^2` (no other package replaced actually require any version of `monolog/monolog`).

> If we merge this PR, we will lose test coverage for Monolog 2.

That would be unfortunate, and not at all the expected outcome, but that seems to outline an issue in the test coverage workflow.

derrabus (changes requested):
> This change should allow one to install `symfony/symfony` with version 3 of `monolog/monolog`.

No. First of all, don't install `symfony/symfony`. Ever. This practice has been discouraged for half a decade now.

Secondly, we're talking about a dependency in `require-dev`. Changing the version range here has absolutely zero impact on projects that use `symfony/symfony` as a dependency.

> That would be unfortunate

Indeed.

> but that seems to outline an issue in the test coverage workflow.

Maybe. If you have a better idea how to solve that, I'm all ears. Note that in Symfony 7, the issue is solved because Monolog 3 is the only supported version there.

xabbuh:
I am closing for the reasons given by @derrabus. Thank you for understanding.

PR #53481, opened 2024-01-10
[Process] Fix executable finder when the command starts with a dash

### LINKED ISSUE #53479: [Process] ExecutableFinder::find() method will execute `exec()` by passed the name parameter

v6.4.2

### How to reproduce

In my Laravel project composer.json file, script defined a few script like that 

```json
"post-update-cmd": [
            "@php -r \"file_put_contents('storage/local/last-composer-lock-hash', md5_file('composer.lock'));\""
        ],
```

when I run `composer run script post-update-cmd` the composer package will remove the prefix `@php ` and extract `-r` to determine if the file is executable, and then the `exec('command -v')` will execute `-r` that `exec('command -v -r')` and output a message that "sh: line 0: command: -r: invalid option
command: usage: command [-pVv] command [arg ...]
"
 

### Possible Solution

I want to pull a PR to update the `exec()` function, What do you think about that?

### Additional Context

_No response_

kayw-geek:
| Q             | A
| ------------- | ---
| Branch?       |  5.4
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Issues        | Fix #53479 
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Process/PhpExecutableFinder.php
+++ b/src/Symfony/Component/Process/PhpExecutableFinder.php
@@ -35,7 +35,7 @@ class PhpExecutableFinder
     {
         if ($php = getenv('PHP_BINARY')) {
             if (!is_executable($php)) {
-                $command = '\\' === \DIRECTORY_SEPARATOR ? 'where' : 'command -v';
+                $command = '\\' === \DIRECTORY_SEPARATOR ? 'where' : 'command -v --';
                 if ($php = strtok(exec($command.' '.escapeshellarg($php)), \PHP_EOL)) {
                     if (!is_executable($php)) {
                         return false;


src/Symfony/Component/Process/ExecutableFinder.php
```diff
@@ -69,7 +69,7 @@ public function find(string $name, string $default = null, array $extraDirs = []
         }
 
         $command = '\\' === \DIRECTORY_SEPARATOR ? 'where' : 'command -v';
-        if (\function_exists('exec') && ($executablePath = strtok(@exec($command.' '.escapeshellarg($name)), \PHP_EOL)) && @is_executable($executablePath)) {
+        if (\function_exists('exec') && !str_starts_with($name,'-') &&($executablePath = strtok(@exec($command.' '.escapeshellarg($name)), \PHP_EOL)) && @is_executable($executablePath)) {
```
xabbuh: this looks wrong to me, if `$name` is not an executable it shouldn't be passed to `find()` in the first place

xabbuh (changes requested):

src/Symfony/Component/Process/ExecutableFinder.php
kayw-geek: In order to avoid users who do not understand the specific implementation of the following error, I think we should avoid similar problems, although the current method's responsibility is to find the executable file, but in the current implementation of the use of exec to execute command -v command to find, I think that to avoid the exec output error also belongs to the function needs to be implemented responsibilities
```bash 
sh: line 0: command: -r: invalid option
command: usage: command [-pVv] command [arg ...]
```

xabbuh: I am sorry but I do not understand what you mean.

kayw-geek: I believe that since the current method uses the `exec()` function, then the `find()` method should have the responsibility to prevent exceptions from occurring with `exec()`.

src/Symfony/Component/Process/ExecutableFinder.php
```diff
@@ -69,7 +69,7 @@ public function find(string $name, string $default = null, array $extraDirs = []
         }
 
         $command = '\\' === \DIRECTORY_SEPARATOR ? 'where' : 'command -v';
```
nicolas-grekas: instead of the proposed change, I'd suggest this:
```suggestion
        $command = '\\' === \DIRECTORY_SEPARATOR ? 'where' : 'command -v --';
```

kayw-geek: @nicolas-grekas missed a blank space?

nicolas-grekas: The space is added on the line below

nicolas-grekas:
Thank you @kayw-geek.

PR #53485, opened 2024-01-10
[AssetMapper] Replace ImportMapConfigReader by interface

damienfern:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        |
| License       | MIT

The direct instantiation of the `ImportMapConfigReader` class has been replaced with the use of the `ImportMapConfigReaderInterface`. Now, by relying on the interface, users can easily configure their own ImportMapConfigReader if needed like this :

```yaml
# config/packages/asset_mapper.yaml
framework:
    asset_mapper:
        config_reader_id: App\CustomImportMapConfigReader
```

TODO:

- [x] Add new config in doc here : https://github.com/symfony/symfony-docs/blob/7.1/frontend/asset_mapper.rst


damienfern:
Thank you @alexandre-daubois for your help :muscle:

src/Symfony/Component/AssetMapper/ImportMap/ImportMapConfigReaderInterface.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace Symfony\Component\AssetMapper\ImportMap;
+
+use Symfony\Component\Filesystem\Path;
+
+interface ImportMapConfigReaderInterface
+{
+    /**
+     * @return ImportMapEntries<ImportMapEntry>
```
alexandre-daubois: `ImportMapEntries` already defines `@template-implements \IteratorAggregate<ImportMapEntry>`, which is why you get that Psalm error. You may just remove these `@return` and static analysis tools should still work :+1:

alexandre-daubois:
Status: Needs review

smnandre:
I see no drawback to use an interface (and let people decorate / replace the instance).

But i do feel it's a bit overkill to add a configuration key + documentation for this.

stof:
What is the use case for this extension point ?

damienfern:
> What is the use case for this extension point ?

My main purpose is to generate an importmap file using another file (which describe all files required and some loading logic). 

Another purpose I can see is people creating their own `importmap.php` with further information.

smnandre:
The `importmap.php` file is almost more a composer.lock than a composer.json file as I see it.

So i'm not sure to follow the "creating their own importmap.php"  : why would they ? 
Do you have any example ? 

Does it still work with all commands, debug, etc ?

damienfern:
Okay, I will clarify my purpose:

I'm trying to use the Asset Mapper with a Vue project. My Vue project is using Vite, a JS Bundler which export a `manifest.json` file. This file can be used to generate the `ImportMapEntries` built by the class `ImportMapConfigReader`. As discussed with @weaverryan, supporting Vite file directly is nothing we should do in core but shouldn't be forbidden. Replacing the use of the class by an interface should allow devs to replace `ImportMapConfigReader` so by their own class or by a third package.

stof:
If you are using Vite to build your files, why trying to use the AssetMapper component ? Make Vite output files inside the `public` folder and then use the Asset component directly to reference those file (as done when using webpack for instance).

OskarStark:
@damienfern after the last comment, do you still think this PR is needed?

Thanks

damienfern:
> @damienfern after the last comment, do you still think this PR is needed?
> 
> Thanks

Hi @OskarStark, sorry for the long time without reply. @smnandre and @stof have come up with very good points and my PR can lead to complicate things in Asset Mapper without big benefits for now. 

So no longer needed

Thanks

PR #53489, opened 2024-01-10
[HttpClient] Add support for `flatten_query_string`

alexandre-daubois:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix https://github.com/orgs/symfony/projects/1#card-30505735
| License       | MIT

Allows to pass a query string in a standard way, where `foo[0]=bar&foo[1]=baz` becomes `foo=bar&foo=baz`.

If someone is asking, the way the query string is formatted by PHP is not standard. This may lead to confusion and errors for non-PHP target that doesn't understand this syntax.

### DIFF
--- a/src/Symfony/Component/HttpClient/CHANGELOG.md
+++ b/src/Symfony/Component/HttpClient/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
  * Add `HttpOptions::setHeader()` to add or replace a single header
  * Allow mocking `start_time` info in `MockResponse`
  * Add `MockResponse::fromFile()` and `JsonMockResponse::fromFile()` methods to help using fixtures files
+ * Add support for `flatten_query_string` in HttpClient options
 
 7.0
 ---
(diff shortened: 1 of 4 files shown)


alamirault (approved):

nicolas-grekas:
Since it's possible to build a query string externally, I'd rather not add an option that can lead to too much mangling (removing any numbers in square brackets looks a bit arbitrary).
There is no standard on the topic. One could very well say that because PHP leads the web, it's the defacto standard...

nicolas-grekas:
What we need to solve is the representation of arbitrary query strings as arrays that we can serialize to query strings.
We do have something similar for headers.
An idea:

```php
$query = [
	['foo', 'bar'],
	['foo', 'baz'],
];
```

Because the root array passes array_is_list, we could decide to serialize it using something else than http_build_query.
We'd need to define something natural that's safe from mistakes as much as possible. E.g I would forbid nested arrays in this representation. For thoughts.

But I'm :-1: on the current approach, closing therefor. Thanks for proposing.

alexandre-daubois:
Thank you for the hints, I'll try to revise my approach

PR #53500, opened 2024-01-11
do not mock `RequestStack` in tests

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

### DIFF
--- a/src/Symfony/Component/HttpKernel/Tests/EventListener/LocaleListenerTest.php
+++ b/src/Symfony/Component/HttpKernel/Tests/EventListener/LocaleListenerTest.php
@@ -11,7 +11,6 @@
 
 namespace Symfony\Component\HttpKernel\Tests\EventListener;
 
-use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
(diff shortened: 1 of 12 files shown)


src/Symfony/Component/HttpKernel/Tests/EventListener/LocaleListenerTest.php
```diff
@@ -26,11 +25,11 @@
 
 class LocaleListenerTest extends TestCase
 {
-    private MockObject&RequestStack $requestStack;
+    private RequestStack $requestStack;
 
     protected function setUp(): void
     {
-        $this->requestStack = $this->createMock(RequestStack::class);
+        $this->requestStack = new RequestStack();
```
OskarStark: Why this setup, I would use new RequestStack() directly for discoverability, no?

OskarStark:
As this hardens the codebase, I am for target branch 5.4

nicolas-grekas (approved):
Thanks!
This is not a bugfix so 7.1 is correct to me. Also the code is different in 5.4 so moving the patch to it would double the effort.

nicolas-grekas:
Thank you @xabbuh.

PR #53505, opened 2024-01-11
[Validator] added reviewed Latvian translations

k0d3r1s:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Validator/Resources/translations/validators.lv.xlf
+++ b/src/Symfony/Component/Validator/Resources/translations/validators.lv.xlf
@@ -136,7 +136,7 @@
             </trans-unit>
             <trans-unit id="37">
                 <source>This is not a valid IP address.</source>
-                <target state="needs-review-translation">Šī vērtība nav derīga IP adrese.</target>
+                <target>Šī nav derīga IP adrese.</target>
             </trans-unit>
             <trans-unit id="38">
                 <source>This value is not a valid language.</source>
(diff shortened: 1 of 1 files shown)


src/Symfony/Component/Validator/Resources/translations/validators.lv.xlf
```diff
@@ -136,7 +136,7 @@
             </trans-unit>
             <trans-unit id="37">
                 <source>This is not a valid IP address.</source>
-                <target state="needs-review-translation">Šī vērtība nav derīga IP adrese.</target>
+                <target>Šī nav derīga IP adrese.</target>
```
nicolas-grekas, line 139: you're removing `vērtība` but we want the translation to be as if the english version where "This *value* is not a valid IP address." (same for other updates). Should we keep the word?

nicolas-grekas: Or use `Šai vērtībai`  as in other existing messages that start with "This value is not"?

k0d3r1s: but source part now does not contain word "value". "vērtība" literally means "value"

nicolas-grekas: yes, that's exactly what I mean.
See #53507

Issue #52587, opened 2023-11-14
[HttpClient] RetryableHttpClient messes timeout exception


rmikalkenas:
### Symfony version(s) affected

6.3.8

### Description

Looks like Symfony's RetryableHttpClient messes up exceptions in case of a timeout.
As showed at the reproducer if request is sent using "GET" method, client throws TimeoutException - which is correct.
But as soon as you change "GET" with "POST", exception class changes to TransportException, even though both exceptions messages are same.
If such behaviour is expected - then how to correctly identify a timeout without weirdly parsing and matching exception message?

Reproduced with php 8.2 and symfony 6.3.8

cc @jderusse maybe you will have some insights, since retryable client is written by you

### How to reproduce

```php
$client = new RetryableHttpClient(HttpClient::create());

try {
    dump('GET');
    $client->request('GET', 'https://httpbin.org/delay/5', ['timeout' => 1.0]);
} catch (\Symfony\Contracts\HttpClient\Exception\ExceptionInterface $e) {
    dump($e::class . '::' . $e->getMessage());
}

// GET
// Symfony\Component\HttpClient\Exception\TimeoutException::Idle timeout reached for "https://httpbin.org/delay/5".

try {
    dump('POST');
    $client->request('POST', 'https://httpbin.org/delay/5', ['timeout' => 1.0]);
} catch (\Symfony\Contracts\HttpClient\Exception\ExceptionInterface $e) {
    dump($e::class . '::' . $e->getMessage());
}

// POST
// Symfony\Component\HttpClient\Exception\TransportException::Idle timeout reached for "https://httpbin.org/delay/5".
```

nicolas-grekas:
I'm reopening, but in the light of #53889, I must say that this might be the expected behavior: timeouts thrown while requesting the status code or headers are not regular timeouts. They're unrecoverable transport errors. Semantically, timeout exceptions are meant to signal a recoverable transport error (one might just need to be more patient to get some activity back and thus recover from the timeout).

About GET vs POST behavior, this might also be the desired behavior: when a timeout is thrown, the remote server might still have processed the request, so that retrying might not be safe.

Of course, if the timeout happens before we sent the headers on our side, then we can safely retry, but we don't have this level of accuracy in the client. Maybe we could detect the situation by leveraging some "info" returned by curl and leverage this in RetryableHttpClient.

If that'd help, AsyncResponse could add the TimeoutException as the "previous" exception of TransportException.

For thoughts.

PR #53506, opened 2024-01-11
[HttpClient] Fix error chunk creation in passthru

rmikalkenas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | #52587 
| License       | MIT

Timeout chunk should not be recreated as TransportException, because ErrorChunk losts information that it's timeout when passing thru.
Change solves issue mentioned at #52587

### DIFF
--- a/src/Symfony/Component/HttpClient/Response/AsyncResponse.php
+++ b/src/Symfony/Component/HttpClient/Response/AsyncResponse.php
@@ -65,7 +65,7 @@ final class AsyncResponse implements ResponseInterface, StreamableInterface
             while (true) {
                 foreach (self::stream([$response], $timeout) as $chunk) {
                     if ($chunk->isTimeout() && $response->passthru) {
-                        foreach (self::passthru($response->client, $response, new ErrorChunk($response->offset, new TransportException($chunk->getError()))) as $chunk) {
+                        foreach (self::passthru($response->client, $response, new ErrorChunk($response->offset, $chunk->getError())) as $chunk) {
                             if ($chunk->isFirst()) {
                                 return false;
                             }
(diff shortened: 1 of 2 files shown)


nicolas-grekas (commented):
It makes sense to remove this wrapper I agree.
Can this be tested somehow?
By curiosity, did you spot why GET vs POST behave differently?

rmikalkenas:
@nicolas-grekas 

> Can this be tested somehow?

It should be possible with MockHttpClient, but looks like timeout mock responses are broken. Returning MockResponse with empty string generator responds with 200 status code and empty content. ErrorChunk seems to be lost somewhere and instead FirstChunk gets created. Later on due to error chunk destructor exception is being thrown.
I will try to dig deeper what's wrong there and fix it -> https://github.com/symfony/symfony/pull/53625

> By curiosity, did you spot why GET vs POST behave differently?

POST method errors is not intended to be retried by the strategy.
So under the hood this is what happens:
- POST request fails and is being passed through the callback (same line I fixed in PR)
- Then the retryable client simply wipes the passthru (`$context->passthru();`) and yields the remaining chunk
- But the error chunk with timeout still is TransportException

With GET method is almost exactly the same, but on the last retry it explicitly wipes the passthru, therefore last retry does not have passthru callback and returns correct TimeoutException

nicolas-grekas:
Got it thanks.
For the tests, MockResponse doesn't support simulating a timeout during the request for now (see my comment in https://github.com/symfony/symfony/pull/53625).
But we do have a real http server available in the test suite, with an endpoint that behaves as we want (`/timeout-header`) so let's use it here?

rmikalkenas:
@nicolas-grekas updated. Failing test doesn't seem to be related with my changes

nicolas-grekas:
Thank you @rmikalkenas.

PR #53420, opened 2024-01-04
[Validator] Consistently use "This value is not" instead of "This is not" in error messages

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

That's longer but more understandable. That's also how most messages are constructed already.

Related to #53417

PR #53507, opened 2024-01-11
[Validator] Make it explicit when English translation differs from its resource name

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Inspired by #53505
Follows #53420

### DIFF
--- a/src/Symfony/Component/Validator/Resources/translations/validators.af.xlf
+++ b/src/Symfony/Component/Validator/Resources/translations/validators.af.xlf
@@ -134,8 +134,8 @@
                 <source>This file is not a valid image.</source>
                 <target>Hierdie lêer is nie 'n geldige beeld nie.</target>
             </trans-unit>
-            <trans-unit id="37">
-                <source>This is not a valid IP address.</source>
+            <trans-unit id="37" resname="This is not a valid IP address.">
+                <source>This value is not a valid IP address.</source>
                 <target state="needs-review-translation">Hierdie waarde is nie 'n geldige IP-adres nie.</target>
             </trans-unit>
             <trans-unit id="38">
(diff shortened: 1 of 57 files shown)


k0d3r1s:
i kinda don't get this one. isn't target supposed to be somewhat exact translation of source?  
why only english can differ in this regard? doesn't this make it somewhat harder to maintain? 
also, regarding #53505, for example the one with ip address translation - source is "This value is not valid ...", rename part is "This is not valid ...". What does latvian part be? With or without "value"? 
Technically with as target is directly translating source? Or latvian part may differ from source, as "value" part is as unnecessary part of that sentence as it is in english?

nicolas-grekas:
@k0d3r1s you might want to check #53420, that should answer your concern.

k0d3r1s:
@nicolas-grekas ok, read through it. now i just feel that this needs to be documented somewhere (probably in docs). as now that message in code differs from translation key, it is harder to find why.  
but again then, why is it not consistent? unit 37 (ip address one) has this new rename but unit 38, for example, does not

nicolas-grekas:
Unit 38 already starts with "This value is not" so I just aligned unit 37. This can be documented for sure, that's what I'm trying to do here, just with a technical PoV. Docs improvement would be nice indeed! Help wanted :)
Why isn't this consistent at the source level? History :man_shrugging: And it's hard to remember when reviewing new messages (but we should still be more careful.)

src/Symfony/Component/Validator/Resources/translations/validators.af.xlf
```diff
@@ -134,8 +134,8 @@
                 <source>This file is not a valid image.</source>
                 <target>Hierdie lêer is nie 'n geldige beeld nie.</target>
             </trans-unit>
-            <trans-unit id="37">
-                <source>This is not a valid IP address.</source>
+            <trans-unit id="37" rename="This is not a valid IP address.">
```
stof:
```suggestion
            <trans-unit id="37" resname="This is not a valid IP address.">
```

stof: the same mistake is done in many other places

stof (changes requested):

stof:
Shouldn't this be implemented by `.github/sync-translations.php` to ensure this is kept uptodate as well ?

The current typos (which are not done for all keys) tell me that this was done manually.

nicolas-grekas:
> Shouldn't this be implemented by .github/sync-translations.php to ensure this is kept uptodate as well ?

yes! script updated.

Status: needs review

PR #53508, opened 2024-01-11
[Console][Messenger] Asynchronously notify transports which messages are still being processed

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 7.2
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

Certain transports, such as Beanstalkd, have a timeout that determines how long a certain message is considered as "in progress". If the message is not acknowledged or rejected within that time period, it is returned to the "ready" queue.

This could cause potential issues when it takes a handler longer to finish processing the message. In those instances, the message being processed could be returned to the ready queue and taken by another instance of the receiver. This would result in the message being processed multiple times, or even worse, in an infinite loop.

Usually, transports that have a timeout also have the option to be notified that the message is still being processed, so the timeout can be postponed.

The idea of this PR is to utilize the `SignalableCommandInterface` and `pcntl_alarm()` to notify transports which messages are still being processed. Since the `SignalRegistry` has [asynchronous signals enabled](https://github.com/symfony/symfony/blob/34915f6e16f04537eb18d9d2c303ec375e63cc4b/src/Symfony/Component/Console/SignalRegistry/SignalRegistry.php#L21) by default, the whole process would happen asynchronously. I've named this feature "keepalive" for lack of a better name.

Currently, I've added this option only to the Beanstalkd transport since that's the one I'm familiar with and use, but as far as I was able to gather, at least one other transport supports this feature. Amazon SQS has a visibility timeout which can be [increased](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html#changing-message-visibility-timeout). This visibility timeout seems to be the same thing as Beanstalkd's TTR. (Disclaimer: I've never used Amazon SQS, so if I got something wrong, please let me know.)

I've split the PR into two commits so it would (hopefully) be easier to review:

1. I've extracted the first commit into a separate PR: #53533
2. The second commit adds the keepalive feature to the Messenger component.

### DIFF
--- a/src/Symfony/Component/Messenger/CHANGELOG.md
+++ b/src/Symfony/Component/Messenger/CHANGELOG.md
@@ -10,6 +10,7 @@ CHANGELOG
  * Add `--format` option to the `messenger:stats` command
  * Add `getRetryDelay()` method to `RecoverableExceptionInterface`
  * Add `skip` option to `messenger:failed:retry` command when run interactively to skip message and requeue it
+ * Add the ability to asynchronously notify transports about which messages are still being processed by the worker, using `pcntl_alarm()`
 
 7.1
 ---
--- a/src/Symfony/Component/Messenger/Worker.php
+++ b/src/Symfony/Component/Messenger/Worker.php
@@ -31,6 +31,7 @@ use Symfony\Component\Messenger\Stamp\FlushBatchHandlersStamp;
 use Symfony\Component\Messenger\Stamp\NoAutoAckStamp;
 use Symfony\Component\Messenger\Stamp\ReceivedStamp;
 use Symfony\Component\Messenger\Stamp\TransportMessageIdStamp;
+use Symfony\Component\Messenger\Transport\Receiver\KeepaliveReceiverInterface;
 use Symfony\Component\Messenger\Transport\Receiver\QueueReceiverInterface;
 use Symfony\Component\Messenger\Transport\Receiver\ReceiverInterface;
 use Symfony\Component\RateLimiter\LimiterInterface;
@@ -47,6 +48,10 @@ class Worker
     private WorkerMetadata $metadata;
     private array $acks = [];
     private \SplObjectStorage $unacks;
+    /**
+     * @var \SplObjectStorage<object, array{0: string, 1: Envelope}>
+     */
+    private \SplObjectStorage $keepalives;
 
     /**
      * @param ReceiverInterface[] $receivers Where the key is the transport name
@@ -63,6 +68,7 @@ class Worker
             'transportNames' => array_keys($receivers),
         ]);
         $this->unacks = new \SplObjectStorage();
+        $this->keepalives = new \SplObjectStorage();
     }
 
     /**
@@ -104,6 +110,10 @@ class Worker
 
                 foreach ($envelopes as $envelope) {
                     $envelopeHandled = true;
+
+                    if ($receiver instanceof KeepaliveReceiverInterface) {
+                        $this->keepalives[$envelope->getMessage()] = [$transportName, $envelope];
+                    }
 
                     $this->rateLimit($transportName);
                     $this->handleMessage($envelope, $transportName);
@@ -186,6 +196,7 @@ class Worker
                 if ($rejectFirst = $e instanceof RejectRedeliveredMessageException) {
                     // redelivered messages are rejected first so that continuous failures in an event listener or while
                     // publishing for retry does not cause infinite redelivery loops
+                    unset($this->keepalives[$envelope->getMessage()]);
                     $receiver->reject($envelope);
                 }
 
@@ -199,6 +210,7 @@ class Worker
                 $envelope = $failedEvent->getEnvelope();
 
                 if (!$rejectFirst) {
+                    unset($this->keepalives[$envelope->getMessage()]);
                     $receiver->reject($envelope);
                 }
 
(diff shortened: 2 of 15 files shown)


valtzu:
I find this interesting since I've faced the issue you described with SQS.

So to add this for SQS, we'd implement `KeepaliveReceiverInterface` on `AmazonSqsTransport` and call `ChangeMessageVisibility` with `visibility_timeout` in `AmazonSqsTransport::keepalive()`. Since this keep-alive mechanism is only used when `--keepalive` is provided to `messenger:consume`, it would not be a BC break even though `visibility_timeout` would have a slightly different meaning then. It would also mean in the configuration `visibility_timeout` should always be bigger than `--keepalive` – probably something to address in the documentation.

valtzu:
Based on my previous comment, I have SQS implementation here https://github.com/valtzu/symfony/commit/59a77be3ffbe0c2744d1362a2a9187cb3ecd277b

HypeMC:
@valtzu That's great, thank you for giving this feature a try and implementing it for SQS. Hopefully, it will prove useful if this PR gets merged.

ro0NL:
This looks good. We will also need it for NATS transport.

Hope it makes 7.1 :crossed_fingers:

src/Symfony/Component/Messenger/Worker.php
```diff
@@ -278,6 +291,20 @@ public function stop(): void
...
+    public function keepalive(): void
+    {
+        foreach ($this->keepalives as $message) {
+            [$transportName, $envelope] = $this->keepalives[$message];
+
+            if (!$this->receivers[$transportName] instanceof KeepaliveReceiverInterface) {
+                throw new RuntimeException(sprintf('Receiver for "%s" does not implement "%s".', $transportName, KeepaliveReceiverInterface::class));
+            }
+
+            $this->logger?->info('Sending keepalive request.', ['transport' => $transportName]);
```
ro0NL: please include the message ID :') ref #53808

HypeMC: Good idea, done

ro0NL (approved):

ro0NL:
So not 7.1? I noticed it skipped feature freeze :weary:

src/Symfony/Component/Messenger/Worker.php
ro0NL: it's not an idea

ro0NL:
Hope it makes 7.2 :crossed_fingers:

src/Symfony/Component/Console/Command/AlarmableCommandInterface.php
```diff
@@ -0,0 +1,34 @@
...
+ * Interface for commands that listen to SIGALRM signals.
+ */
+interface AlarmableCommandInterface
+{
+    /**
+     * The method will be called before the command is run and subsequently on each SIGALRM signal.
+     *
+     * @return int The alarm interval in seconds
+     */
+    public function getAlarmInterval(InputInterface $input): int;
```
nicolas-grekas: instead of adding this interface, what about adding Application::setAlarmInterval? A command would then have to call the method in execute() and that'd make things simpler, don't you think?

HypeMC: @nicolas-grekas Interesting idea, I'm guessing that the `handleAlarm` would then be replaced with the standard `SignalableCommandInterface`?

nicolas-grekas: absolutely

HypeMC: @nicolas-grekas Please take a look at the [new implementation](https://github.com/symfony/symfony/pull/53508/commits/5c32d9c9e507580698a5a029a43eac743d6c37b6).

src/Symfony/Component/Console/Application.php
```diff
@@ -975,34 +990,47 @@ protected function doRunCommand(Command $command, InputInterface $input, OutputI
 
         $commandSignals = $command instanceof SignalableCommandInterface ? $command->getSubscribedSignals() : [];
         if ($commandSignals || $this->dispatcher && $this->signalsToDispatchEvent) {
-            if (!$this->signalRegistry) {
-                throw new RuntimeException('Unable to subscribe to signal events. Make sure that the "pcntl" extension is installed and that "pcntl_*" functions are not disabled by your php.ini\'s "disable_functions" directive.');
-            }
+            $signalRegistry = $this->getSignalRegistry();
```
HypeMC: `getSignalRegistry` throws a nearly identical exception, so it made sense to me to have it centralized.

src/Symfony/Component/Messenger/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ CHANGELOG
  * Add `#[AsMessage]` attribute with `$transport` parameter for message routing
  * Add `--format` option to the `messenger:stats` command
  * Add `getRetryDelay()` method to `RecoverableExceptionInterface`
+ * Add the ability to asynchronously notify transports about which messages are still being processed by the handler, using `pcntl_alarm()`
```
chalasr:
```suggestion
 * Add the ability to asynchronously notify transports about which messages are still being processed by the worker, using `pcntl_alarm()`
```

HypeMC: @chalasr Done

fabpot:
@HypeMC Can you rebase now that #53533 has been merged? Thank you.

HypeMC:
@fabpot Done

fabpot (approved):

chalasr (approved):

chalasr:
Thank you @HypeMC.

PR #53511, opened 2024-01-11
[Routing] Merge Route conditions.

### LINKED ISSUE #53171: [Routing] Do not let `RouteCollection` override `Route` conditions
### Description

Hi,

**TLDR**: General Condition defined in YAML/XML collections overrides specific condition defined in attribute.

I have a trouble with route conditions on a project I'm working on.
Routes are loading via a mix of YAML and Attributes :
```yaml
api:
    resource: ../../src/Controller/
    type: annotation
    condition: "%general_condition%"
```
```php
#[Route(path: "my/route", condition: "specific condition")]
final class MyController {
   ...
}
```
My trouble is that the specific condition, defined by attribute, is overrided by the general condition, defined via YAML.
Looking into the component, I found out that when the YamlFileLoader sets the condition on the RouteCollection, the RouteCollection overrides any previous condition.
I find it disturbing that a general condition overrides a specific condition.
```php
public function setCondition(?string $condition)
{
    foreach ($this->routes as $route) {
        $route->setCondition($condition);
    }
}
```

**Propositions**
1) Make specific overrides general, but I fear it is a BC since previously enforced general condition would not be any more.
```php
public function setCondition(?string $condition)
{
    foreach ($this->routes as $route) {
        if ('' === $route->getCondition()) {
            $route->setCondition($condition);
        }
    }
}
```

2) Combine conditions. I don't think of it as a BC since one cannot expect its attribute condition to **not** work.
```php
public function setCondition(?string $condition)
{
    foreach ($this->routes as $route) {
        if ('' === $route->getCondition()) {
            $route->setCondition($condition);
        } else {
           $route->setCondition('(' . $route->getCondition . ') and (' . $condition . ')');
        }
    }
}
```

mbuliard:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #53171
| License       | MIT

When defining condition for a Route and for the RouteCollection it belongs, the two conditions are now merged instead of letting the RouteCollection overrides the Route condition.

### DIFF
--- a/src/Symfony/Component/Routing/RouteCollection.php
+++ b/src/Symfony/Component/Routing/RouteCollection.php
@@ -256,7 +256,11 @@ class RouteCollection implements \IteratorAggregate, \Countable
     public function setCondition(?string $condition): void
     {
         foreach ($this->routes as $route) {
-            $route->setCondition($condition);
+            if ('' === $route->getCondition()) {
+                $route->setCondition($condition);
+            } else {
+                $route->setCondition('('.$route->getCondition().') and ('.$condition.')');
+            }
         }
     }
 
--- a/src/Symfony/Component/Routing/Tests/Fixtures/validpattern.xml
+++ b/src/Symfony/Component/Routing/Tests/Fixtures/validpattern.xml
@@ -14,5 +14,7 @@
         <condition>context.getMethod() == "GET"</condition>
     </route>
 
-    <route id="blog_show_inherited" path="/blog/{slug}" />
+    <route id="blog_show_inherited" path="/blog/{slug}">
+        <condition>context.getMethod() == "GET"</condition>
+    </route>
 </routes>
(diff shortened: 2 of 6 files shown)


WebMamba (commented):
I am not sure merging the two condition are the right thing we want here. I think the Route condition should overwrite the route collection condition

mbuliard:
> I am not sure merging the two condition are the right thing we want here. I think the Route condition should overwrite the route collection condition

Yes, I also think it would be the best. I proposed this in the related issue. But I fear it will be a BC could not be merge before 8.0

stof:
Even in 8.0, it would still be impossible to merge it if it just changes the order in which the values are applied. We need a migration path where we can opt in the new behavior and trigger a deprecation when it is not done, otherwise the BC break won't be accepted either.

stof:
and the change done in this PR is actually also a BC break.

fabpot:
Closing as there's no more activities.

Issue #45976, opened 2022-04-08
[RFC] Change the way deprecations are reported?


javiereguiluz:
### Description

If you work as a developer of Symfony applications, try to answer this question honestly:

**When do you work on resolving direct/indirect deprecations in your project?**

(1) Almost never. Just when preparing to upgrade the Symfony version.
(2) Randomly, from time to time.
(3) Most of the time. Always.

My own experience is (1) and my guess is that (1) is also the answer for most developers.

-----

When running tests with `./vendor/bin/simple-phpunit` I see things like this _(some parts use an unintelligible font for privacy reasons)_:

![tests-deprecation-before](https://user-images.githubusercontent.com/73419/162440980-450ec05f-bff1-4be8-b35e-dc09f885d3ed.jpg)

I love deprecations and the system created by @nicolas-grekas for Symfony is stunning. All PHP projects should use something like this. It's super useful when fixing deprecations.

**The problem** is that the output of tests is optimized for (3) _(people fixing or caring about deprecations all the time)_, whereas most people are probably in (1) _(I only care about deprecations very occasionally, when upgrading Symfony)_. That's why I think it's annoying to display this long list of deprecations that move the PHPUnit output too far.

-----

A **posible solution** would be to have an output like this one:

<img width="1369" alt="tests-deprecation-after" src="https://user-images.githubusercontent.com/73419/162439840-d86a2242-01d8-44e6-a6dd-2bacf62e1f4f.png">

Tests would always remind you about pending deprecations to fix, but wouldn't list them whenever you run tests.

This proof-of-concept prototype mentions that you should run `simple-phpunit -v` to see deprecations. But we could solve this in other ways such as creating a dedicated command to display deprecations.

So, what do you think? Thanks!

kbond:
Strongly agree! There is a mechanism to quiet non-self/direct deprecations but there is [a bug](https://github.com/symfony/symfony/pull/43371) I haven't had much luck fixing.

Edit: for clarification, I'm on the (3) side for self/direct deprecations (ones I can actually do something about) and (1/2) for other deprecations.

Lustmored:
Well I'm on the (3) side, but I'm surely in minority as large part of my job is refactor/maintenance/improving code of large projects that date to many years ago. Therefore I love as many deprecation notices as possible at any step.

But at the same time it would make completely no difference to me if adding a single parameter (like `-v`) was required to keep seeing those. So to benefit most I believe hiding them by default is still the best idea.

[...]

PR #53513, opened 2024-01-11
[PhpUnitBridge] Hide deprecations by default when running tests

javiereguiluz:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | #45976
| License       | MIT

As explained in #45976, some people don't like that deprecations are always shown when running tests. The list can be very long and it hides the PHPUnit output. So, this PR proposes to hide the deprecations by default but display them if the `simple-phpunit` is run with the `-v` option.

This is an example of how it looks now:

test-deprecations-before

And this is how it'd look after this PR:

test-deprecations-after

~As mentioned in the code, there's a missing piece. I don't know how to detect if the command was run with the `-v` option. I need your help. Thanks.~ This if fixed now.

### DIFF
--- a/src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler.php
+++ b/src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler.php
@@ -298,6 +298,10 @@ class DeprecationErrorHandler
             return $b->count() - $a->count();
         };
 
+        $outputIsCli = !$configuration->shouldWriteToLogFile();
+        // TODO: fix this; we need to detect here if the `simple-phpunit` command was run with `-v` option
+        $commandIsVerbose = false;
+        $deprecationsAreHidden = $outputIsCli && !$commandIsVerbose;
         if ($configuration->shouldWriteToLogFile()) {
             if (false === $handle = @fopen($file = $configuration->getLogFile(), 'a')) {
                 throw new \InvalidArgumentException(sprintf('The configured log file "%s" is not writeable.', $file));
@@ -315,7 +319,7 @@ class DeprecationErrorHandler
                 );
                 if ($configuration->shouldWriteToLogFile()) {
                     fwrite($handle, "\n$deprecationGroupMessage\n");
-                } else {
+                } elseif (!$deprecationsAreHidden) {
                     fwrite($handle, "\n".self::colorize($deprecationGroupMessage, 'legacy' !== $group && 'indirect' !== $group)."\n");
                 }
 
@@ -327,7 +331,9 @@ class DeprecationErrorHandler
                 uasort($notices, $cmp);
 
                 foreach ($notices as $msg => $notice) {
-                    fwrite($handle, sprintf("\n  %sx: %s\n", $notice->count(), $msg));
+                    if (!$deprecationsAreHidden) {
+                        fwrite($handle, sprintf("\n  %sx: %s\n", $notice->count(), $msg));
+                    }
 
                     $countsByCaller = $notice->getCountsByCaller();
                     arsort($countsByCaller);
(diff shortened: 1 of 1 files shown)


src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler.php
```diff
@@ -298,6 +298,10 @@ private function displayDeprecations(array $groups, Configuration $configuration
             return $b->count() - $a->count();
         };
 
+        $outputIsCli = !$configuration->shouldWriteToLogFile();
+        // TODO: fix this; we need to detect here if the `simple-phpunit` command was run with `-v` option
```
wouterj: We shouldn't build this for the simple-phpunit script specifically. We must do this following PHPunit's own executable (especially as we're recommending not using simple-phpunit for applications)

craigh:
In principle, I fully agree with this PR. I don't like seeing all the deprecations _by default_ either.

I would suggest also that direct vs. indirect deprecation display could be controlled with the level of verbosity. E.g. `-v` shows the direct deprecations and `-vv` (or `-vvv`) displays both direct and indirect deprecations.

src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler.php
```diff
@@ -298,6 +298,10 @@ private function displayDeprecations(array $groups, Configuration $configuration
...
+        $commandIsVerbose = false;
```
wouterj:
```suggestion
        $commandIsVerbose = [] !== array_intersect(['v', 'verbose'], array_keys(getopt('v', ['verbose'])));
```

(PHPunit natively supports the `-v|--verbose` options, but they are saved in a `Configuration` class that is not easily available and not covered by the BC promise)

wouterj:
Also shared on Slack: I don't think we need 2 ways to do the same thing and have 2 different styles.

The best thing imho we can do if we want to go this way is to add `<env name="SYMFONY_DEPRECATIONS_HELPER" value="verbose=0" />` to the PHPunit config file in the recipes. I would personally say that `quiet[]=indirect&quiet[]=other` is always a more sensible default and would vote for using it in the recipe (still showing direct/self deprecations).

javiereguiluz:
@wouterj my own workflow with deprecations is like this:

* I don't want deprecations to "bother" me everytime I run tests
* But, when some Symfony upgrade is near, I want to see those deprecations

So, that's why I propose: _"don't display them by default; but make it super easy to display them"_.

I'm afraid that dealing with `quiet[]=indirect&quiet[]=other` creates a big friction for most users ... it's such a confusing and convoluted syntax. Dealing with env vars in the phpunit.xml file is not that bad, but it's not ideal either (maybe the env var is not set up and you need to look up the long-named env var in the docs; you need to edit a file everytime you want to show or hide them, etc.)

src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler.php
javiereguiluz: Thanks Wouter 🙌

wouterj:
Whatever default we would like, the main issue I have with this PR is that it duplicates something we already provide. The handler already has code checking for verbosity and outputting less or more details based on it (in fact, the code of this PR is right after this if statement): https://github.com/symfony/symfony/blob/34915f6e16f04537eb18d9d2c303ec375e63cc4b/src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler.php#L318-L325

That's a no-go in my opinion. However, there are 2 things that you're doing that we might want to tweak in the current system:
- Make the summary output nicer, like you proposed (currently we output one summary message per group, see `$deprecationGroupMessage`)
- Interact with PHPunit's `-v` and `--verbose` by setting a different default value for the `SYMFONY_DEPRECATIONS_HELPER` env var (change the default to whatever we decide and make the current output the `-v` situation).

But I would like to see this as improvements of the existing feature, instead of creating an adjecent feature.

_Btw, I agree that the URL-based configuration of the env var is not easy to read. If you have good ideas for it, I would be happy to know them :)_

ro0NL:
our approach (using native phpunit) is to only show deprecations in CI (ideally in a collapsable group, but it was [rejected](https://github.com/sebastianbergmann/phpunit/issues/5435))

it requires zero config, and we just watch the deprecation list shrink/grow over time

javiereguiluz:
Let's close this as "won't fix" because the related issue (#45976) was closed.

PR #53516, opened 2024-01-11
[Console] Allow '0' as a $shortcut in InputOption.php

### LINKED ISSUE #53514: [Console] InputOption shortcut "-0" is not allowed

2.1-7.1

### Description

In the constructor of `Symfony\Component\Console\Input\InputOption` there is [some code](https://github.com/symfony/symfony/blob/efb9fa0b666d90f252fc7210c9d3c1392553a34d/src/Symfony/Component/Console/Input/InputOption.php#L90) to check for empty values of `$shortcut`:
#
```php
        if (empty($shortcut)) {
            $shortcut = null;
        }

        if (null !== $shortcut) {
            if (\is_array($shortcut)) {
                $shortcut = implode('|', $shortcut);
            }
            $shortcuts = preg_split('{(\|)-?}', ltrim($shortcut, '-'));
            $shortcuts = array_filter($shortcuts);
            $shortcut = implode('|', $shortcuts);

            if (empty($shortcut)) {
                throw new InvalidArgumentException('An option shortcut cannot be empty.');
            }
        }
```

Presumably the goal is to disallow empty strings as shortcut values.

But the use of `array_filter` with the default/empty callback here prevents using a shortcut like `-0`, because `'0'` evaluates to `false`.  (While uncommon, `-0` might be used in a command-line application to enable null-terminated output - see for example [xargs](https://www.gnu.org/software/findutils/manual/html_node/find_html/xargs-options.html) or [locate](https://www.gnu.org/software/findutils/manual/html_node/find_html/Invoking-locate.html))

### How to reproduce

Configure the following input option on any console command:

```php
new InputOption('null-terminated', '0', description: 'Print null-terminated output')
```

Observe the exception when executing the command:

```
 An option shortcut cannot be empty.
```

### Possible Solution

#53516

### Additional Context

_No response_

lawsonjl-ornl:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #53514
| License       | MIT

Fixes some comparisons to no longer erroneously disallow `'0'`/`'-0'` as a shortcut value.

### DIFF
--- a/src/Symfony/Component/Console/Input/InputOption.php
+++ b/src/Symfony/Component/Console/Input/InputOption.php
@@ -69,7 +69,7 @@ class InputOption
             throw new InvalidArgumentException('An option name cannot be empty.');
         }
 
-        if (empty($shortcut)) {
+        if ('' === $shortcut || [] === $shortcut) {
             $shortcut = null;
         }
 
@@ -78,10 +78,10 @@ class InputOption
                 $shortcut = implode('|', $shortcut);
             }
             $shortcuts = preg_split('{(\|)-?}', ltrim($shortcut, '-'));
-            $shortcuts = array_filter($shortcuts);
+            $shortcuts = array_filter($shortcuts, 'strlen');
             $shortcut = implode('|', $shortcuts);
 
-            if (empty($shortcut)) {
+            if ('' === $shortcut) {
                 throw new InvalidArgumentException('An option shortcut cannot be empty.');
             }
         }
--- a/src/Symfony/Component/Console/Tests/Input/InputOptionTest.php
+++ b/src/Symfony/Component/Console/Tests/Input/InputOptionTest.php
@@ -55,6 +55,20 @@ class InputOptionTest extends TestCase
         $this->assertEquals('f|ff|fff', $option->getShortcut(), '__construct() removes the leading - of the shortcuts');
         $option = new InputOption('foo');
         $this->assertNull($option->getShortcut(), '__construct() makes the shortcut null by default');
+        $option = new InputOption('foo', '');
+        $this->assertNull($option->getShortcut(), '__construct() makes the shortcut null when given an empty string');
+        $option = new InputOption('foo', []);
+        $this->assertNull($option->getShortcut(), '__construct() makes the shortcut null when given an empty array');
+        $option = new InputOption('foo', ['f', '', 'fff']);
+        $this->assertEquals('f|fff', $option->getShortcut(), '__construct() removes empty shortcuts');
+        $option = new InputOption('foo', 'f||fff');
+        $this->assertEquals('f|fff', $option->getShortcut(), '__construct() removes empty shortcuts');
+        $option = new InputOption('foo', '0');
+        $this->assertEquals('0', $option->getShortcut(), '-0 is an acceptable shortcut value');
+        $option = new InputOption('foo', ['0', 'z']);
+        $this->assertEquals('0|z', $option->getShortcut(), '-0 is an acceptable shortcut value when embedded in an array');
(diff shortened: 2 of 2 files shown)


lawsonjl-ornl:
I added some test cases to the existing `InputOptionTest::testShortcut` method. Three of those cases test that `-0` is accepted now. The rest of them test some closely related behavior around the handling of `[]` and `''`, and the elision of empty values from string-lists/arrays (e.g. `'f||fff'` or `['f', '', 'fff']`) - these tests were included to ensure I didn't break any existing behavior with my changes.

noplanman (approved):

src/Symfony/Component/Console/Input/InputOption.php
```diff
@@ -78,10 +78,10 @@ public function __construct(string $name, $shortcut = null, int $mode = null, st
                 $shortcut = implode('|', $shortcut);
             }
             $shortcuts = preg_split('{(\|)-?}', ltrim($shortcut, '-'));
-            $shortcuts = array_filter($shortcuts);
+            $shortcuts = array_filter($shortcuts, 'strlen');
             $shortcut = implode('|', $shortcuts);
 
-            if (empty($shortcut)) {
+            if ($shortcut === '') {
```
fabpot:
```suggestion
            if ('' !== $shortcut) {
```

fabpot (changes requested):

src/Symfony/Component/Console/Input/InputOption.php
lawsonjl-ornl: This will change the code's behavior and throw the `InvalidArgumentException` when `$shortcut` is nonempty. 

Perhaps you meant to suggest:

```php
if ('' === $shortcut) {
```

To "flip" the expression into a "yoda condition"?

fabpot: Oops, sorry, my bad, I meant: use Yoda condition

lawsonjl-ornl: I have changed it in bf8cbbd.

fabpot (approved):

fabpot:
Thank you @lawsonjl-ornl.

src/Symfony/Component/Console/Input/InputOption.php
```diff
@@ -69,7 +69,7 @@ public function __construct(string $name, $shortcut = null, int $mode = null, st
             throw new InvalidArgumentException('An option name cannot be empty.');
         }
 
-        if (empty($shortcut)) {
+        if ('' === $shortcut || [] === $shortcut) {
```
vivdesign, line 72: $shortcut === false is not handled

nicolas-grekas: can you please send a PR on branch 5.4 to fix this?

lawsonjl-ornl: The documented acceptable types for `$shortcut` are `string`, `array`, and `null`, and the previously-existing test cases only covered those types, so in that sense passing `false` here is an incorrect use of the API in the first place. 

But passing `false` happened to accidentally work before the change, so I guess it is a breaking change (to undocumented behavior).

A fix that restores the handling of `false` should probably also update the types in the docstring and add a test case, so that the next person who has reason to modify this code knows that behavior should be preserved.

sagar-patona:
this change  causing below issue 
`        if ('' === $shortcut || [] === $shortcut) {
`
  An option shortcut cannot be empty.    "nwidart/laravel-modules":8.2
image

PR #53518, opened 2024-01-11
[Console] Add a stream helper

louismariegaborit:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | 
| License       | MIT

As I explain in [another PR](https://github.com/symfony/symfony/pull/52222#issuecomment-1773910201), testing commands like `server:dump` or `server:log` is very difficult due to the infinite loop that waits for cancellation by the user.

I suggest to create a new helper in console component to use same code between above two commands and test them to avoid regression in these commands.

What do you think ?

Regarding the use of the new StreamHelper, the Dumper component seems to no longer need the DumpServer class because we use now only the getHost function. To replace DumpServer in the command, I would like to use the ParameterBag but I can use it in the binary var-dump-server ? Or I should add a string parameter that I set by CompilerPass for the command ?

### DIFF
--- a/src/Symfony/Component/Console/Helper/StreamHelper.php
+++ b/src/Symfony/Component/Console/Helper/StreamHelper.php
@@ -1,0 +1,104 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Console\Helper;
+
+use Symfony\Component\Console\Exception\RuntimeException;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\StreamableInputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+/**
+ * This class provides helpers to interact with the socket stream server.
+ *
+ * @author Louis-Marie Gaborit <lm.gabo@gmail.com>
+ */
+class StreamHelper extends Helper
+{
+    /**
+     * @var resource|null
+     */
+    private $inputStream;
+
+    /**
+     * @var resource|null
+     */
+    private $socket;
+
+    private static bool $stty = true;
+    private static bool $stdinIsInteractive;
+
+    private function start(string $host): void
+    {
+        if (!$this->socket = stream_socket_server($host, $errno, $errstr)) {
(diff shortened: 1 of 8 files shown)


src/Symfony/Component/VarDumper/Tests/Command/ServerDumpCommandTest.php
```diff
@@ -28,6 +32,24 @@ public function testComplete(array $input, array $expectedSuggestions)
...
+    {
+        $command = $this->createCommand();
+        $commandTester = new CommandTester($command);
+
+        $data = new Data([['my dump']]);
+        $input = base64_encode(serialize([$data, ['timestamp' => time(), 'source' => ['name' => 'sourceName', 'line' => 222, 'file' => 'myFile']]]))."\n";
+
+        $commandTester->setInputs([$input]);
+
+        $commandTester->execute(['--format' => 'html']);
```
louismariegaborit, line 45: For this test, I can only use html because value of the dump doesn't be available in the command result. I seems that it be prompt in php stream not in the command.  
@ogizanagi Could you help me to add cli format in this test ?

src/Symfony/Component/Console/Helper/StreamHelper.php
```diff
@@ -0,0 +1,104 @@
...
+namespace Symfony\Component\Console\Helper;
+
+use Symfony\Component\Console\Exception\RuntimeException;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\StreamableInputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+/**
+ * The StreamHelper class provides helpers to interact with the socket stream server.
```
OskarStark:
```suggestion
 * This class provides helpers to interact with the socket stream server.
```

nicolas-grekas:
Thanks for the PR.
I'm wondering if we shouldn't do it the other way around: make ServerLogCommand use DumpServer?
It would provide two benefits:
- the proposed helper is not really a "stream" helper, but one that helps some specific kind of line-based server - not generic enough maybe - at least not for the name "stream"
- less dependency constraints (no need for bumps)

It'd be fine making VarDumper a required dependency of the command in 7.1 (throwing a LogicException when DumpServer is not found in the constructor of ServerLogCommand)

WDYT?

nicolas-grekas:
I'm closing for now, waiting for a new PR in case you're up to giving my above proposal a try :pray:

PR #53519, opened 2024-01-11
chore: CS - apply `phpdoc_types_order.sort_algorithm=alpha`

keradus:
| Q             | A
| ------------- | ---
| Branch?       | 7.1 
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Issues        | Fix CS 
| License       | MIT

I suggest to unify types in phpdoc, for combined types (eg `foo|bar`).

So far:
- put `null` as last
- everything else as is (random order?)

Idea:
- put `null` as last
- everything else sorted alphabetically

### DIFF
--- a/src/Symfony/Component/Config/Definition/Builder/NormalizationBuilder.php
+++ b/src/Symfony/Component/Config/Definition/Builder/NormalizationBuilder.php
@@ -45,7 +45,7 @@ class NormalizationBuilder
     /**
      * Registers a closure to run before the normalization or an expression builder to build it if null is provided.
      *
-     * @return ExprBuilder|$this
+     * @return $this|ExprBuilder
      */
     public function before(\Closure $closure = null): ExprBuilder|static
     {
--- a/src/Symfony/Component/Process/Process.php
+++ b/src/Symfony/Component/Process/Process.php
@@ -55,7 +55,7 @@ class Process implements \IteratorAggregate
     private array|string $commandline;
     private ?string $cwd;
     private array $env = [];
-    /** @var resource|string|\Iterator|null */
+    /** @var \Iterator|resource|string|null */
     private $input;
     private ?float $starttime = null;
     private ?float $lastOutputTime = null;
@@ -137,7 +137,7 @@ class Process implements \IteratorAggregate
      * @param string|null    $cwd     The working directory or null to use the working dir of the current PHP process
      * @param array|null     $env     The environment variables or null to use the same environment as the current PHP process
      * @param mixed          $input   The input as stream resource, scalar or \Traversable, or null for no input
-     * @param int|float|null $timeout The timeout in seconds or null to disable
+     * @param float|int|null $timeout The timeout in seconds or null to disable
      *
      * @throws LogicException When proc_open is not installed
      */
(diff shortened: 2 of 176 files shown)


xabbuh:
I do not really care about the order. That said, if I would start a new codebase I would probably sort types alphabetically, but also put classes before scalar types and `false` and `null` at the end. But I do not have any hard feelings about how we handle this in the Symfony codebase.

keradus:
I believe it's better to have some sorting than none (kind of `none = sort:random`).
right now, we have only alpha sorting possibility (with null at end).

for any more sophisticated sorting, one would need to implement ;)

src/Symfony/Component/Serializer/Normalizer/NormalizerInterface.php
```diff
@@ -59,7 +59,7 @@ public function supportsNormalization(mixed $data, string $format = null, array
      * Use type "object" to match any classes or interfaces,
      * and type "*" to match any types.
      *
-     * @return array<class-string|'*'|'object'|string, bool|null>
+     * @return array<'*'|'object'|class-string|string, bool|null>
```
smnandre, line 62: I'd let "phpstan-like types" first

src/Symfony/Component/Config/Definition/Builder/NormalizationBuilder.php
```diff
@@ -45,7 +45,7 @@ public function remap(string $key, string $plural = null): static
     /**
      * Registers a closure to run before the normalization or an expression builder to build it if null is provided.
      *
-     * @return ExprBuilder|$this
+     * @return $this|ExprBuilder
```
smnandre, line 48: I'd force "$this" / self / etc at right :|

src/Symfony/Component/Process/Process.php
```diff
@@ -1123,7 +1123,7 @@ public function getInput()
      *
      * This content will be passed to the underlying process standard input.
      *
-     * @param string|resource|\Traversable|self|null $input The content
+     * @param resource|self|string|\Traversable|null $input The content
```
smnandre, line 1126: Same, could "self"  have a special rule ... ? :)

src/Symfony/Component/Serializer/Normalizer/NormalizerInterface.php
keradus: Currently, Fixer does not support such logic.
Feel free to introduce it via [raising PR](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/compare)

src/Symfony/Component/Config/Definition/Builder/NormalizationBuilder.php
keradus: Currently, Fixer does not support such logic.
Feel free to introduce it via [raising PR](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/compare)

src/Symfony/Component/Process/Process.php
keradus: Currently, Fixer does not support such logic.
Feel free to introduce it via [raising PR](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/compare)

nicolas-grekas:
Reading the comments here, I think we might not want to do this change.
I don't think the current order is random - instead, I'd say each method author wrote it in the way that made sense to them.
And I see that alpha ordering doesn't make sense to many. I don't know if there is some non-alpha order that we could all agree on. Eg native types, then FQCN, then literal types, then null - but what about complex types, etc. I can't decide myself.

Better close and do nothing IMHO as I also see a lot of potential for bikeshadding here ;)

chalasr:
I agree with Nicolas. Thanks for proposing.

PR #53520, opened 2024-01-12
[Messenger] [AMQP] Throw exception on `nack` callback

### LINKED ISSUE #53229: [Messenger] Negative publisher confirm swallowed

<=7

### Description

For reliable message publishing, RabbitMQ offers the possibility to enable [publisher confirms](https://www.rabbitmq.com/confirms.html#publisher-confirms) on a channel. If enabled, the success of processing a message will be confirmed by the broker. If the broker responds with `ack`, all is fine. In case of receiving `nack` however, the client is expected to re-publish the message.

Unfortunately, a `nack` seems to be swallowed by Messenger:
In [this line](https://github.com/symfony/amqp-messenger/blob/2ee5d2574b3f745078eb3836ec60d1530aac184b/Transport/Connection.php#L348), Messenger waits for the confirm. The [waitForConfirm() method](https://github.com/php-amqp/php-amqp/blob/4a4194300e5c1519b4a23dd33d3b37b9d0d9e7b8/amqp_channel.c#L1312) then invokes the callback for `ack` or `nack`. Both callbacks have been provided [here](https://github.com/symfony/amqp-messenger/blob/2ee5d2574b3f745078eb3836ec60d1530aac184b/Transport/Connection.php#L494). 

As both callbacks just return "false", no further handling of a `nack`-case is done. There's no way for code which sends a message to know if this has been successful or not.

The expectation would be that at least an exception is thrown in that case. An even better implementation would perform some retries before eventually throwing an exception.

### How to reproduce

When a queue is limited and rejects new messages, the publisher is informed by a `nack` response. This behavior could easily be provoked by creating an overflow scenario.

**messenger.yaml**
``` yaml
...
    myTransport:
        dsn: amqp://...
        options:
          confirm_timeout: 3 # enable publisher confirms
```

**Assigning a policy in RabbitMQ:**
Either assign a policy like explained [here](https://www.rabbitmq.com/maxlength.html). Or use the admin frontend, click "Admin/Policies/Add policy" and enter the values 
- Name: SomeName
- Definitions:
  - max-lenght=2
  - overflow=reject-publish 

**Start a SymfonyMessanger consumer**
``` bash
console messenger:consume myTransport
```

**Sending some messages**
``` php
class MyMessage
{
    public function __construct(readonly string $text){}
}

#[AsMessageHandler]
class MyMessageHandler
{
    public function __construct(private readonly LoggerInterface $logger)
    {
    }

    public function __invoke(MyMessage $message): void
    {
        $this->logger->info("Message received: " . $message->text);
    }
}

for ($i = 0; $i < 100; $i++) {
    $this->messageBus->dispatch(new MyMessage("Message Nr. $i"));
}
```

**Result**
-> Only some message are received. The publisher doesn't get any info that it should retry publishing.

### Possible Solution

_No response_

### Additional Context

_No response_

kvrushifa:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #53229 
| License       | MIT

If the channel is in confirm mode, it is currently not possible to determine if a message has been nack'ed. By throwing an exception within the confirm callback, it is at least possible to react to it.

Return false is not needed to end the wait loop, [since the amqp ext checks for exceptions](https://github.com/php-amqp/php-amqp/blob/e58ac221e317c840ee06f7731e0dc76c7d6a431a/amqp_methods_handling.c#L249).

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Amqp/Transport/Connection.php
+++ b/src/Symfony/Component/Messenger/Bridge/Amqp/Transport/Connection.php
@@ -494,7 +494,7 @@ class Connection
                 $this->amqpChannel->confirmSelect();
                 $this->amqpChannel->setConfirmCallback(
                     static fn (): bool => false,
-                    static fn (): bool => throw new RuntimeException('Confirm calls back with nack.'),
+                    static fn () => throw new RuntimeException('Confirm calls back with nack.'),
                 );
             }
 
--- a/src/Symfony/Component/Messenger/Bridge/Amqp/Transport/Connection.php
+++ b/src/Symfony/Component/Messenger/Bridge/Amqp/Transport/Connection.php
@@ -13,7 +13,7 @@ namespace Symfony\Component\Messenger\Bridge\Amqp\Transport;
 
 use Symfony\Component\Messenger\Exception\InvalidArgumentException;
 use Symfony\Component\Messenger\Exception\LogicException;
-use Symfony\Component\Messenger\Exception\RuntimeException;
+use Symfony\Component\Messenger\Exception\TransportException;
 
 /**
  * An AMQP connection.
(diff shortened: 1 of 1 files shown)


kvrushifa:
⚠️ Please first merge #53525 for SF version 5.4

nicolas-grekas (approved):

nicolas-grekas:
Merged as #53525, this patch will come up to 6.3 as a merge up.

PR #53521, opened 2024-01-12
[VarDumper] Fixes `Typed property Symfony\Component\VarDumper\Dumper\CliDumper::$colors must not be accessed before initialization`

crynobone:
| Q             | A
| ------------- | ---
| Branch?       | 7.0
| Bug fix?      | yes
| New feature?  | no 
| Deprecations? | no 
| Issues        | N/A
| License       | MIT

Fixes the following error:

CleanShot 2024-01-12 at 16 27 25

Other methods in the class start with `$this->colors ??= $this->supportsColors();`. I believe this should have the same code.

### DIFF
--- a/src/Symfony/Component/VarDumper/Dumper/CliDumper.php
+++ b/src/Symfony/Component/VarDumper/Dumper/CliDumper.php
@@ -550,7 +550,7 @@ class CliDumper extends AbstractDumper
 
     protected function dumpLine(int $depth, bool $endOfValue = false): void
     {
-        if ($this->colors) {
+        if ($this->colors ??= $this->supportsColors()) {
             $this->line = sprintf("\033[%sm%s\033[m", $this->styles['default'], $this->line);
         }
         parent::dumpLine($depth);


xabbuh (approved):

src/Symfony/Component/VarDumper/Dumper/CliDumper.php
```diff
@@ -550,6 +550,8 @@ protected function supportsColors(): bool
 
     protected function dumpLine(int $depth, bool $endOfValue = false): void
     {
+        $this->colors ??= $this->supportsColors();
+
         if ($this->colors) {
```
nicolas-grekas:
```suggestion
        if ($this->colors ??= $this->supportsColors()) {
```

nicolas-grekas (commented):
if we're missing an initialization here, it means the patch should likely go on 5.4

crynobone:
> if we're missing an initialization here, it means the patch should likely go on 5.4

Not exactly, since 7.0 changed from `protected $colors;` to `protected bool $colors;`.

fabpot (approved):

fabpot:
Thank you @crynobone.

PR #53523, opened 2024-01-12
[Serializer] Rewrite `AbstractObjectNormalizer::createChildContext()` to use the provided `cache_key` from original context

amne:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | N/A
| License       | MIT

If the normalization starts with an initial 'cache_key' it is currently discarded when creating child contexts. This change fixes that. The main reason behind it is that if the client code sends a unique identifier (ApiPlatform injects the IRI) and we have a use case that iterates a big resultset we end up with a big private cache that quickly runs out of allowed memory. The solution should be to send a 'cache_key' which ignores the entire cache key calculation that hashes the context.

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
@@ -287,6 +287,8 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
 
     /**
      * Gets the attribute value.
+     *
+     * @return mixed
      */
     abstract protected function getAttributeValue(object $object, string $attribute, string $format = null, array $context = []);
 
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
@@ -153,7 +153,9 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
      */
     public function normalize(mixed $object, string $format = null, array $context = [])
     {
-        $context['cache_key'] = $this->getCacheKey($format, $context);
+        if (!isset($context['cache_key'])) {
+            $context['cache_key'] = $this->getCacheKey($format, $context);
+        }
 
         $this->validateCallbackContext($context);
 
@@ -300,7 +302,9 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
 
     public function denormalize(mixed $data, string $type, string $format = null, array $context = [])
     {
-        $context['cache_key'] = $this->getCacheKey($format, $context);
+        if (!isset($context['cache_key'])) {
+            $context['cache_key'] = $this->getCacheKey($format, $context);
+        }
 
         $this->validateCallbackContext($context);
 
@@ -743,10 +747,6 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
      */
     private function getCacheKey(?string $format, array $context): bool|string
     {
-        if (isset($context['cache_key'])) {
-            return $context['cache_key'];
-        }
-
         foreach ($context[self::EXCLUDE_FROM_CACHE_KEY] ?? $this->defaultContext[self::EXCLUDE_FROM_CACHE_KEY] as $key) {
             unset($context[$key]);
         }
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
@@ -153,9 +153,7 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
      */
     public function normalize(mixed $object, string $format = null, array $context = [])
     {
-        if (!isset($context['cache_key'])) {
-            $context['cache_key'] = $this->getCacheKey($format, $context);
-        }
+        $context['cache_key'] = $this->getCacheKey($format, $context);
 
         $this->validateCallbackContext($context);
 
@@ -287,8 +285,6 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
 
     /**
      * Gets the attribute value.
-     *
-     * @return mixed
      */
     abstract protected function getAttributeValue(object $object, string $attribute, string $format = null, array $context = []);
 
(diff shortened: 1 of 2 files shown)


src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
```diff
@@ -287,8 +285,6 @@ abstract protected function extractAttributes(object $object, string $format = n
 
     /**
      * Gets the attribute value.
-     *
-     * @return mixed
```
mtarld: These coding style changes must not be committed as they don't belong to your bugfix.

src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
```diff
@@ -153,9 +153,7 @@ public function supportsNormalization(mixed $data, string $format = null /* , ar
      */
     public function normalize(mixed $object, string $format = null, array $context = [])
     {
-        if (!isset($context['cache_key'])) {
```
mtarld: I'd keep this if to avoid calling a method each time. Indeed, calling a method could be costly and we are in the hot path here.

src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
```diff
@@ -302,14 +298,9 @@ public function supportsDenormalization(mixed $data, string $type, string $forma
         return class_exists($type) || (interface_exists($type, false) && null !== $this->classDiscriminatorResolver?->getMappingForClass($type));
     }
 
-    /**
-     * @return mixed
-     */
     public function denormalize(mixed $data, string $type, string $format = null, array $context = [])
     {
-        if (!isset($context['cache_key'])) {
```
mtarld, line 310: Same here

src/Symfony/Component/Serializer/Tests/Normalizer/AbstractObjectNormalizerTest.php
```diff
@@ -788,6 +788,43 @@ public function supportsNormalization(mixed $data, string $format = null, array
...
+                $childContext = parent::createChildContext($parentContext, $attribute, $format);
+                AbstractObjectNormalizerTest::assertSame('hardcoded-'.$attribute, $childContext['cache_key']);
+
+                return $childContext;
+            }
+        };
+
+        $serializer = new Serializer([$normalizer]);
+        $normalizedObject = $serializer->normalize($foobar, null, ['cache_key' => 'hardcoded']);
+        $this->assertSame($data, $normalizedObject);
```
mtarld: To make sure that an assertion has been made, I'd use a property and make an assertion directly in the test:
```suggestion
        $normalizer = new class() extends AbstractObjectNormalizerDummy {
            public ?string $childContextCacheKey = null;
            
            protected function extractAttributes(object $object, string $format = null, array $context = []): array
            {
                return array_keys((array) $object);
            }

            protected function getAttributeValue(object $object, string $attribute, string $format = null, array $context = []): mixed
            {
                return $object->{$attribute};
            }

            protected function createChildContext(array $parentContext, string $attribute, ?string $format): array
            {
                $childContext = parent::createChildContext($parentContext, $attribute, $format);
                $this->childContextCacheKey = $childContext['cache_key'];

                return $childContext;
            }
        };
        
        $serializer = new Serializer([$normalizer]);
        $normalizedObject = $serializer->normalize($foobar, null, ['cache_key' => 'hardcoded']);

        $this->assertSame($data, $normalizedObject);
        $this->assertSame('hardcoded-'.$attribute, $normalizer->childContextCacheKey);
```

src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
```diff
@@ -746,6 +743,10 @@ protected function createChildContext(array $parentContext, string $attribute, ?
      */
     private function getCacheKey(?string $format, array $context): bool|string
     {
+        if (isset($context['cache_key'])) {
```
mtarld: Should it go after `$context[self::EXCLUDE_FROM_CACHE_KEY]` foreach loop?

mtarld (changes requested):
As it's a bug, you must target the 5.4 branch instead.

src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
amne: If I revert the removal of the inline ifs then this one goes away so it doesn't matter.

amne: But to answer, I don't it should go after because the method would not be called if there is an existing $context['cache_key']

src/Symfony/Component/Serializer/Tests/Normalizer/AbstractObjectNormalizerTest.php
```diff
@@ -788,6 +788,46 @@ public function supportsNormalization(mixed $data, string $format = null, array
...
+                $childContext = parent::createChildContext($parentContext, $attribute, $format);
+                $this->childContextCacheKey = $childContext['cache_key'];
+
+                return $childContext;
+            }
+        };
+
+        $serializer = new Serializer([$normalizer]);
+        $normalizedObject = $serializer->normalize($foobar, null, ['cache_key' => 'hardcoded']);
+        // $this->assertSame($data, $normalizedObject);
```
mtarld, line 827: Unexpected comment I guess

src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
```diff
@@ -302,9 +302,6 @@ public function supportsDenormalization(mixed $data, string $type, string $forma
         return class_exists($type) || (interface_exists($type, false) && null !== $this->classDiscriminatorResolver?->getMappingForClass($type));
     }
 
-    /**
```
mtarld, line 305: This change should be reverted

mtarld (changes requested):
Can you rebase your PR on the 5.4 branch please? :slightly_smiling_face:

mtarld:
This can be closed in favor of https://github.com/symfony/symfony/pull/53530, thanks!

PR #53524, opened 2024-01-12
[Messenger] [AmazonSqs] Allow `async-aws/sqs` version 2

smoench:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Issues        |  -
| License       | MIT

With this PR async-aws/sqs version 2 would be allowed to be installed. [With version 2 they are using the AWS JSON-1.0 protocol instead of the XML one](https://github.com/async-aws/sqs/blob/master/CHANGELOG.md#200). They declared this as a BC-Break as they switced the protocol from query to json, but no public method signatures have been changed.

TODO
- [x] Provide JSON stubs for tests

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -95,9 +95,9 @@ jobs:
           - 8094:8094
           - 11210:11210
       sqs:
-        image: asyncaws/testing-sqs
+        image: localstack/localstack:3.0.2
         ports:
-          - 9494:9494
+          - 4566:4566
       zookeeper:
         image: wurstmeister/zookeeper:3.4.6
       kafka:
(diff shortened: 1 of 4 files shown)


xabbuh (changes requested):
test failures are related

smoench:
Tests are fixed. appveyor failues are unrelated to this PR.

derrabus (approved):

xabbuh (approved):
For 5.4?

smoench:
Base branch changed to 5.4

xabbuh (approved):

stof (changes requested):
The dev requirement in the root composer.json should also be updated.

And once this is done, you will need to fix the integration-tests workflow which will break because the `asyncaws/testing-sqs` docker image only provides a fake implementation of the old protocol. You should migrate to `localstack` for the integration tests (as was done in async-aws itself)

smoench:
@stof I updated the missing parts :-)

nicolas-grekas (approved):

nicolas-grekas:
Thank you @smoench.

PR #53528, opened 2024-01-12
[Messenger][Redis] Add option to auto generate consumer

### LINKED ISSUE #51604: [Messenger] Generated redis transport `consumer` name

danielburger1337:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #51604 
| License       | MIT

I opened the above mentioned issue about 4 months ago. 
In #53394 @AdamKatzDev , who seems to know a lot more about Redis than I do, also thinks that this is a good idea. 

It looks like nobody has had a reason why this should not be a feature, so here it is.

Docs PR will be added when this looks good to merge.

### DIFF
--- a/src/Symfony/Component/Messenger/Bridge/Redis/CHANGELOG.md
+++ b/src/Symfony/Component/Messenger/Bridge/Redis/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Add option to auto generate unique consumer name.
+
 6.3
 ---
 
--- a/src/Symfony/Component/Messenger/Bridge/Redis/Tests/Transport/ConnectionTest.php
+++ b/src/Symfony/Component/Messenger/Bridge/Redis/Tests/Transport/ConnectionTest.php
@@ -107,6 +107,14 @@ class ConnectionTest extends TestCase
         );
     }
 
+    public function testFromDsnGeneratedConsumer()
+    {
+        $this->assertNotEquals(
+            Connection::fromDsn('redis://localhost/queue/group1/_GENERATED', [], $this->createMock(\Redis::class)),
+            Connection::fromDsn('redis://localhost/queue/group1/_GENERATED', [], $this->createMock(\Redis::class))
+        );
+    }
+
     public function testRedisClusterInstanceIsSupported()
     {
         $redis = $this->createMock(\RedisCluster::class);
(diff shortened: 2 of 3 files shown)


OskarStark:
Just asking, but wouldn't it be safer to add an option to the DSN `?generate_consumer_true` ? Or is this not possible?

danielburger1337:
> Just asking, but wouldn't it be safer to add an option to the DSN `?generate_consumer_true` ? Or is this not possible?

This is technically possible. But since the consumer name is currently passed as path to the DSN, I think your approach wouldn't be intuitive because then two options/settings related to the consumer name would exist.

nicolas-grekas:
Closing, thanks for the proposal and sorry for the long silence.

Giving each worker its own consumer name is already covered: set the `consumer` option from a per-instance environment variable, which the Messenger docs show for Supervisor and systemd, and `HOSTNAME` works for one worker per container.

A generated name would also be worse than that. It changes on every restart, so the previous consumer stays in the Redis group holding its pending entries until `redeliver_timeout` claims them, and the consumer list grows over time. Stable names avoid this, which is why the docs point at a `StatefulSet` on Kubernetes.

@OskarStark's point about the magic `_GENERATED` value stands too: it would silently change behaviour for anyone naming a consumer that way.

PR #53530, opened 2024-01-12
[Serializer] Rewrite `AbstractObjectNormalizer::createChildContext()` to use the provided `cache_key` from original context when creating child contexts

amne:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | N/A
| License       | MIT

If the normalization starts with an initial 'cache_key' it is currently discarded when creating child contexts. This change fixes that.
The main reason behind it is that if the client code sends a unique identifier (ApiPlatform injects the IRI) and we have a use case
that iterates a big resultset we end up with a big private cache that quickly runs out of allowed memory. The solution should be to
send a 'cache_key' which skips the entire cache key calculation that hashes the context.

### DIFF
--- a/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
+++ b/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
@@ -781,7 +781,11 @@ abstract class AbstractObjectNormalizer extends AbstractNormalizer
     protected function createChildContext(array $parentContext, string $attribute, ?string $format): array
     {
         $context = parent::createChildContext($parentContext, $attribute, $format);
-        $context['cache_key'] = $this->getCacheKey($format, $context);
+        if ($context['cache_key'] ?? false) {
+            $context['cache_key'] .= '-'.$attribute;
+        } else {
+            $context['cache_key'] = $this->getCacheKey($format, $context);
+        }
 
         return $context;
     }
--- a/src/Symfony/Component/Serializer/Tests/Normalizer/AbstractObjectNormalizerTest.php
+++ b/src/Symfony/Component/Serializer/Tests/Normalizer/AbstractObjectNormalizerTest.php
@@ -548,6 +548,131 @@ class AbstractObjectNormalizerTest extends TestCase
         $this->assertEquals(new DummyWithStringObject(new DummyString()), $actual);
         $this->assertEquals('', $actual->value->value);
     }
+
+    public function testProvidingContextCacheKeyGeneratesSameChildContextCacheKey()
+    {
+        $foobar = new Dummy();
+        $foobar->foo = new EmptyDummy();
+        $foobar->bar = 'bar';
+        $foobar->baz = 'baz';
+        $data = [
+            'foo' => [],
+            'bar' => 'bar',
+            'baz' => 'baz',
+        ];
+
+        $normalizer = new class() extends AbstractObjectNormalizerDummy {
+            public $childContextCacheKey;
+
+            protected function extractAttributes(object $object, string $format = null, array $context = []): array
+            {
+                return array_keys((array) $object);
+            }
+
+            protected function getAttributeValue(object $object, string $attribute, string $format = null, array $context = [])
+            {
+                return $object->{$attribute};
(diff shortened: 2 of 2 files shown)


mtarld:
Follow-up of https://github.com/symfony/symfony/pull/53523

src/Symfony/Component/Serializer/Tests/Normalizer/AbstractObjectNormalizerTest.php
```diff
@@ -548,6 +548,90 @@ public function testDenormalizeUntypedStringObject()
...
+            {
+                $childContext = parent::createChildContext($parentContext, $attribute, $format);
+                $this->childContextCacheKey = $childContext['cache_key'];
+
+                return $childContext;
+            }
+        };
+
+        $serializer = new Serializer([$normalizer]);
+        $serializer->normalize($foobar, null, ['cache_key' => 'hardcoded', 'iri' => '/dummy/1']);
```
mtarld:
```suggestion
        $normalizer->normalize($foobar, null, ['cache_key' => 'hardcoded', 'iri' => '/dummy/1']);
```

src/Symfony/Component/Serializer/Tests/Normalizer/AbstractObjectNormalizerTest.php
```diff
@@ -548,6 +548,90 @@ public function testDenormalizeUntypedStringObject()
...
+            {
+                $childContext = parent::createChildContext($parentContext, $attribute, $format);
+                $this->childContextCacheKey = $childContext['cache_key'];
+
+                return $childContext;
+            }
+        };
+
+        $serializer = new Serializer([$normalizer]);
+        $serializer->normalize($foobar, null, ['cache_key' => false]);
```
mtarld, line 672:
```suggestion
        $normalizer->normalize($foobar, null, ['cache_key' => false]);
```

src/Symfony/Component/Serializer/Tests/Normalizer/AbstractObjectNormalizerTest.php
```diff
@@ -548,6 +548,90 @@ public function testDenormalizeUntypedStringObject()
         $this->assertEquals(new DummyWithStringObject(new DummyString()), $actual);
         $this->assertEquals('', $actual->value->value);
     }
+
+    public function testChildContextKeepsOriginalContextCacheKey()
```
mtarld: Maybe you can test  the `xxx-attribute` cache key creation as well?

src/Symfony/Component/Serializer/Tests/Normalizer/AbstractObjectNormalizerTest.php
amne: Doesn't work because apparently normalizer needs a serializer injected:

```bash
 Symfony\Component\Serializer\Exception\LogicException: Cannot normalize attribute "foo" because the injected serializer is not a normalizer.
```

mtarld (approved):

src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php
```diff
@@ -781,7 +781,13 @@ private function isMaxDepthReached(array $attributesMetadata, string $class, str
     protected function createChildContext(array $parentContext, string $attribute, ?string $format): array
     {
         $context = parent::createChildContext($parentContext, $attribute, $format);
-        $context['cache_key'] = $this->getCacheKey($format, $context);
+        if (isset($context['cache_key'])) {
+            if (false !== $context['cache_key']) {
+                $context['cache_key'] .= '-'.$attribute;
+            }
```
nicolas-grekas:
```suggestion
        if ($context['cache_key'] ?? false) {
            $context['cache_key'] .= '-'.$attribute;
```

nicolas-grekas (approved):
LGTM with minor CS comments

nicolas-grekas:
Thank you @amne.

PR #53531, opened 2024-01-12
[Messenger] Use `now` function to allow mock of time

andersonamuller:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | N/A
| License       | MIT

When doing functional testing there is not really a easy way to mock the time for the `InMemoryTransport`. It's created by the `InMemoryTransportFactory` and you can't set which clock to use.


ro0NL (approved):

ro0NL:
however i think we should inject global clock into InMemoryTransportFactory isnt it?

andersonamuller:
> however i think we should inject global clock into InMemoryTransportFactory isnt it?

Yes, that is the best solution indeed

derrabus (changes requested):
I don't believe that this is the right way. If the transport already receives a clock, we should rather leverage the injected clock for mocking than introduce a _second_ way of accessing mocked date values.

nicolas-grekas:
>It's created by the InMemoryTransportFactory and you can't set which clock to use.

I'd rather fix this aspect indeed: making the InMemoryTransportFactory accept and forward a clock.

PR welcome.

PR #53533, opened 2024-01-13
[Console] Add ability to schedule alarm signals and a `ConsoleAlarmEvent`

### LINKED ISSUE #47920: [Console] add console.alarm event
### Description

Currently, the Doctrine connection can get pinged on each new message if the middleware is enabled. But, the message handlers can be very heavy and timeouts short (for example, default `wait_timeout` on Azure MySQL is 180s).

If you have a very heavy handler, the handler execution process might be longer than the idle timeout and, once the handler completes, `ack()` can throw an `MySQL has gone away` error even though the connection was pinged successfully when the message was received.

The suggestion is to basically incorporate this https://twitter.com/lyrixx/status/1575127719624544258 into DoctrinePingConnectionMiddleware (or in a more appropriate place).

/cc @lyrixx 

### Example

1. new message
2. ping connection, valid
3. run handler, it does a bunch of stuff (say, 3rd party APIs, etc)
4. 5min later...
5. `ack()` fails with `MySQL has gone away`, `SSL: Connection reset by peer` or similar error

This change would ensure the connection is "exercised" during the handler execution window too.

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 7.2
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #47920
| License       | MIT

Part of #53508

This PR introduces the ability to schedule alarm signals and a `console.alarm` event. A command using this feature will automatically dispatch an alarm at the specified interval:

```php
#[AsCommand('app:alarm')]
class AlarmCommand extends Command implements SignalableCommandInterface
{
    private bool $run = true;
    private int $count = 0;
    private OutputInterface $output;

    protected function initialize(InputInterface $input, OutputInterface $output): void
    {
        $this->getApplication()->setAlarmInterval(10);
    }

    {
        $this->output = $output;

        while ($this->run) {
            $this->count++;
            $output->writeln('execute '.$this->count);
            sleep(1);
        }

        $output->writeln('Done');

        return Command::SUCCESS;
    }

    public function getSubscribedSignals(): array
    {
        return [\SIGINT, \SIGALRM];
    }

    public function handleSignal(int $signal, false|int $previousExitCode = 0): int|false
    {
        if (\SIGINT === $signal) {
            $this->run = false;
            $this->output->writeln('Bye');
        } elseif (\SIGALRM === $signal) {
            $this->count = 0;
            $this->output->writeln('handleAlarm');
        }

        return false;
    }
}
```

The `console.alarm` event is dispatched on every `SIGALRM` signal:

```php
#[AsEventListener(event: ConsoleAlarmEvent::class)]
final class ConsoleAlarmListener
{
    public function __invoke(ConsoleAlarmEvent $event): void
    {
        $event->getOutput()->writeln('ConsoleAlarmListener');
    }
}
```

### DIFF
--- a/src/Symfony/Component/Console/Application.php
+++ b/src/Symfony/Component/Console/Application.php
@@ -22,6 +22,7 @@ use Symfony\Component\Console\CommandLoader\CommandLoaderInterface;
 use Symfony\Component\Console\Completion\CompletionInput;
 use Symfony\Component\Console\Completion\CompletionSuggestions;
 use Symfony\Component\Console\Completion\Suggestion;
+use Symfony\Component\Console\Event\ConsoleAlarmEvent;
 use Symfony\Component\Console\Event\ConsoleCommandEvent;
 use Symfony\Component\Console\Event\ConsoleErrorEvent;
 use Symfony\Component\Console\Event\ConsoleSignalEvent;
@@ -88,6 +89,7 @@ class Application implements ResetInterface
     private bool $initialized = false;
     private ?SignalRegistry $signalRegistry = null;
     private array $signalsToDispatchEvent = [];
+    private ?int $alarmInterval = null;
 
     public function __construct(
         private string $name = 'UNKNOWN',
@@ -97,7 +99,7 @@ class Application implements ResetInterface
         $this->defaultCommand = 'list';
         if (\defined('SIGINT') && SignalRegistry::isSupported()) {
             $this->signalRegistry = new SignalRegistry();
-            $this->signalsToDispatchEvent = [\SIGINT, \SIGQUIT, \SIGTERM, \SIGUSR1, \SIGUSR2];
+            $this->signalsToDispatchEvent = [\SIGINT, \SIGQUIT, \SIGTERM, \SIGUSR1, \SIGUSR2, \SIGALRM];
         }
     }
 
@@ -126,6 +128,22 @@ class Application implements ResetInterface
     public function setSignalsToDispatchEvent(int ...$signalsToDispatchEvent): void
     {
         $this->signalsToDispatchEvent = $signalsToDispatchEvent;
+    }
+
+    /**
+     * Sets the interval to schedule a SIGALRM signal in seconds.
+     */
+    public function setAlarmInterval(?int $seconds): void
+    {
+        $this->alarmInterval = $seconds;
+        $this->scheduleAlarm();
+    }
+
+    private function scheduleAlarm(): void
+    {
+        if (null !== $this->alarmInterval) {
+            $this->getSignalRegistry()->scheduleAlarm($this->alarmInterval);
+        }
     }
 
     /**
@@ -981,34 +999,47 @@ class Application implements ResetInterface
 
         $commandSignals = $command instanceof SignalableCommandInterface ? $command->getSubscribedSignals() : [];
         if ($commandSignals || $this->dispatcher && $this->signalsToDispatchEvent) {
-            if (!$this->signalRegistry) {
-                throw new RuntimeException('Unable to subscribe to signal events. Make sure that the "pcntl" extension is installed and that "pcntl_*" functions are not disabled by your php.ini\'s "disable_functions" directive.');
-            }
+            $signalRegistry = $this->getSignalRegistry();
 
             if (Terminal::hasSttyAvailable()) {
                 $sttyMode = shell_exec('stty -g');
 
                 foreach ([\SIGINT, \SIGQUIT, \SIGTERM] as $signal) {
-                    $this->signalRegistry->register($signal, static fn () => shell_exec('stty '.$sttyMode));
+                    $signalRegistry->register($signal, static fn () => shell_exec('stty '.$sttyMode));
                 }
             }
 
             if ($this->dispatcher) {
                 // We register application signals, so that we can dispatch the event
                 foreach ($this->signalsToDispatchEvent as $signal) {
-                    $event = new ConsoleSignalEvent($command, $input, $output, $signal);
+                    $signalEvent = new ConsoleSignalEvent($command, $input, $output, $signal);
+                    $alarmEvent = \SIGALRM === $signal ? new ConsoleAlarmEvent($command, $input, $output) : null;
 
-                    $this->signalRegistry->register($signal, function ($signal) use ($event, $command, $commandSignals) {
-                        $this->dispatcher->dispatch($event, ConsoleEvents::SIGNAL);
-                        $exitCode = $event->getExitCode();
+                    $signalRegistry->register($signal, function ($signal) use ($signalEvent, $alarmEvent, $command, $commandSignals, $input, $output) {
+                        $this->dispatcher->dispatch($signalEvent, ConsoleEvents::SIGNAL);
+                        $exitCode = $signalEvent->getExitCode();
+
+                        if (null !== $alarmEvent) {
+                            if (false !== $exitCode) {
+                                $alarmEvent->setExitCode($exitCode);
+                            } else {
+                                $alarmEvent->abortExit();
+                            }
+                            $this->dispatcher->dispatch($alarmEvent);
+                            $exitCode = $alarmEvent->getExitCode();
+                        }
 
                         // If the command is signalable, we call the handleSignal() method
                         if (\in_array($signal, $commandSignals, true)) {
                             $exitCode = $command->handleSignal($signal, $exitCode);
                         }
 
+                        if (\SIGALRM === $signal) {
+                            $this->scheduleAlarm();
+                        }
+
                         if (false !== $exitCode) {
-                            $event = new ConsoleTerminateEvent($command, $event->getInput(), $event->getOutput(), $exitCode, $signal);
+                            $event = new ConsoleTerminateEvent($command, $input, $output, $exitCode, $signal);
                             $this->dispatcher->dispatch($event, ConsoleEvents::TERMINATE);
 
                             exit($event->getExitCode());
@@ -1021,7 +1052,11 @@ class Application implements ResetInterface
             }
 
             foreach ($commandSignals as $signal) {
-                $this->signalRegistry->register($signal, function (int $signal) use ($command): void {
+                $signalRegistry->register($signal, function (int $signal) use ($command): void {
+                    if (\SIGALRM === $signal) {
+                        $this->scheduleAlarm();
+                    }
+
                     if (false !== $exitCode = $command->handleSignal($signal)) {
                         exit($exitCode);
                     }
--- a/src/Symfony/Component/Console/Tests/phpt/alarm/command_exit.phpt
+++ b/src/Symfony/Component/Console/Tests/phpt/alarm/command_exit.phpt
@@ -21,26 +21,34 @@ require $vendor.'/vendor/autoload.php';
 
 class MyCommand extends Command implements SignalableCommandInterface
 {
+    protected function initialize(InputInterface $input, OutputInterface $output): void
+    {
+        $this->getApplication()->setAlarmInterval(1);
+    }
+
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        posix_kill(posix_getpid(), \SIGINT);
+        sleep(5);
 
         $output->writeln('should not be displayed');
 
         return 0;
     }
 
-
     public function getSubscribedSignals(): array
     {
-        return [\SIGINT];
+        return [\SIGALRM];
     }
 
     public function handleSignal(int $signal, int|false $previousExitCode = 0): int|false
     {
-        echo "Received signal!";
+        if (\SIGALRM === $signal) {
+            echo "Received alarm!";
 
-        return 0;
+            return 0;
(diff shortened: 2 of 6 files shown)


src/Symfony/Component/Console/Command/SignalableCommandInterface.php
```diff
@@ -20,8 +23,11 @@ interface SignalableCommandInterface
 {
     /**
      * Returns the list of signals to subscribe.
+     *
+     * @param InputInterface  $input
+     * @param OutputInterface $output
      */
-    public function getSubscribedSignals(): array;
+    public function getSubscribedSignals(/* InputInterface $input, OutputInterface $output */): array;
```
stof: you cannot add mandatory arguments in a method signature through such BC layer, because the child class is not allowed to add them as mandatory arguments.

Btw, changing the subscribed signals based on the *output* looks weird to me. What is the use case for it ? Is this worth a BC break ?

src/Symfony/Component/Console/Command/TraceableCommand.php
```diff
@@ -86,9 +88,14 @@ public function __call(string $name, array $arguments): mixed
         return $this->command->{$name}(...$arguments);
     }
 
-    public function getSubscribedSignals(): array
+    public function getSubscribedSignals(/* InputInterface $input, OutputInterface $output */): array
```
stof: A final class should add the arguments directly instead of such forward compatibility layer, given that it does not have child classes that prevent adding them.

src/Symfony/Component/Console/Application.php
```diff
@@ -975,7 +977,18 @@ protected function doRunCommand(Command $command, InputInterface $input, OutputI
...
+        // bind before getSubscribedSignals() and the console.command event, so the methods and listeners have access to input options/arguments
+        try {
+            $command->mergeApplicationDefinition();
+            $input->bind($command->getDefinition());
+        } catch (ExceptionInterface) {
+            // ignore invalid options/arguments for now, to allow the event listeners to customize the InputDefinition
+        }
+
+        $commandSignals = $command instanceof SignalableCommandInterface ? $command->getSubscribedSignals($input, $output) : [];
+        if ($command instanceof AlarmableCommandInterface && \defined('SIGALRM') && SignalRegistry::isSupported() && !\in_array(\SIGALRM, $commandSignals, true)) {
```
stof: Making the TraceableCommand implement AlarmableCommandInterface means that in dev, **all** commands will register this `SIGALRM` (even when they are not alarmable). Is this expected ?

src/Symfony/Component/Console/Application.php
```diff
@@ -992,19 +1005,34 @@ protected function doRunCommand(Command $command, InputInterface $input, OutputI
...
+                            $exitCode = $alarmEvent->getExitCode();
+                        }
 
+                        if (\SIGALRM === $signal && $command instanceof AlarmableCommandInterface) {
+                            $exitCode = $command->handleAlarm($exitCode);
+                            $this->signalRegistry->scheduleAlarm($command->getAlarmTime($input, $output));
+                        }
                         // If the command is signalable, we call the handleSignal() method
-                        if (\in_array($signal, $commandSignals, true)) {
+                        elseif (\in_array($signal, $commandSignals, true)) {
```
stof: this will never be called in dev when a SignalableCommand asks for `SIGALRM` because of the TraceableCommand wrapping it and implementing AlarmableCommandInterface. This means this PR is currently a BC break.

src/Symfony/Component/Console/Application.php
```diff
@@ -1017,26 +1045,29 @@ protected function doRunCommand(Command $command, InputInterface $input, OutputI
             }
 
             foreach ($commandSignals as $signal) {
-                $this->signalRegistry->register($signal, function (int $signal) use ($command): void {
-                    if (false !== $exitCode = $command->handleSignal($signal)) {
+                $this->signalRegistry->register($signal, function (int $signal) use ($command, $input, $output): void {
+                    if (\SIGALRM === $signal && $command instanceof AlarmableCommandInterface) {
```
stof: Same issue here regarding TraceableCommand.

src/Symfony/Component/Console/Command/AlarmableCommandInterface.php
```diff
@@ -0,0 +1,35 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Console\Command;
+
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Interface for command that dispatches SIGALRM signal.
```
stof: Is the command actually dispatching SIGALRM signals ? The rest of the code gives me the impression that it *listens* to those signals.

stof (changes requested):

src/Symfony/Component/Console/Command/SignalableCommandInterface.php
HypeMC: The more I thought about it, the less sense it made to me as well. I initially had a use case for #53508, but I've since found a better approach, and now I can't really think of a good use case. Therefore, I've removed the entire commit. If an actual use case comes up in the future, it can be added then. I've also removed the `$output` argument from `getAlarmTime()` as it was also useless.

src/Symfony/Component/Console/Command/TraceableCommand.php
HypeMC: This is no longer an issue since I've removed that commit.

src/Symfony/Component/Console/Application.php
HypeMC: No, good catch, I've totally missed that one. I've added a `TraceableAlarmableCommand`, so this should no longer be an issue. I'm not sure if this is the best approach, but I couldn't think of any other.

src/Symfony/Component/Console/Command/AlarmableCommandInterface.php
HypeMC: Fixed

src/Symfony/Component/Console/Command/AlarmableCommandInterface.php
```diff
@@ -0,0 +1,34 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\Console\Command;
+
+use Symfony\Component\Console\Input\InputInterface;
+
+/**
+ * Interface for command that listens to SIGALRM signals.
```
chalasr:
```suggestion
 * Interface for commands that listen to SIGALRM signals.
```

src/Symfony/Component/Console/Command/AlarmableCommandInterface.php
```diff
@@ -0,0 +1,34 @@
...
+ */
+interface AlarmableCommandInterface
+{
+    /**
+     * The method will be called before the command is run and subsequently on each SIGALRM signal.
+     *
+     * @return int The alarm time in seconds
+     */
+    public function getAlarmTime(InputInterface $input): int;
```
chalasr: I'm wondering if `time` is the right term, isn't this more an interval?

src/Symfony/Component/Console/Tests/phpt/alarm/command_exit.phpt
```diff
@@ -0,0 +1,55 @@
+--TEST--
+Test command that exist
```
chalasr:
```suggestion
Test command that exits
```

chalasr (changes requested):

src/Symfony/Component/Console/Command/AlarmableCommandInterface.php
HypeMC: Done

src/Symfony/Component/Console/Tests/phpt/alarm/command_exit.phpt
HypeMC: Done

src/Symfony/Component/Console/Command/AlarmableCommandInterface.php
HypeMC: God point, I've changed it to `getAlarmInterval()`

valtzu:
Btw, `SIGALRM` has the same effect as any other signal – it interrupts `sleep` & `usleep`. According to docs, one should check based on return code if the `sleep` was interrupted but I doubt many do that – and `usleep`  returns `void` so for that it's impossible.

So if a command logic f.e. includes `sleep(30);` or `$clock->sleep(30);`, it will only sleep until next alarm, potentially causing unexpected behavior.

---

Another (old) thing I found that using `SIGALRM` to provide timer functionality [may result in deadlock](https://nativeguru.wordpress.com/2015/02/19/why-you-should-avoid-using-sigalrm-for-timer/) – not sure if it's still relevant (or if it ever was, for php).

ro0NL:
from the original issue, it was taken into account:

```
    $left = sleep(10);
    while ($left > 0) {
        $left = sleep($left);
    }
```

i think this is a reasonable workaround (but needs documentation), even if it's not supported on windows (that's a PHP API issue/bug), and should not block this PR

for Linux, we could even fix/abstract it in `$clock->sleep()` (as per above example), either optionally or out-of-the-box

ro0NL (approved):
cc @chalasr 

we're running a userland implementation in prod, it works fine

src/Symfony/Component/Console/Application.php
```diff
@@ -975,34 +990,47 @@ protected function doRunCommand(Command $command, InputInterface $input, OutputI
 
         $commandSignals = $command instanceof SignalableCommandInterface ? $command->getSubscribedSignals() : [];
         if ($commandSignals || $this->dispatcher && $this->signalsToDispatchEvent) {
-            if (!$this->signalRegistry) {
-                throw new RuntimeException('Unable to subscribe to signal events. Make sure that the "pcntl" extension is installed and that "pcntl_*" functions are not disabled by your php.ini\'s "disable_functions" directive.');
-            }
+            $signalRegistry = $this->getSignalRegistry();
```
HypeMC, line 1002: `getSignalRegistry` throws a nearly identical exception, so it made sense to me to have it centralized.

chalasr (approved):
The implementation is now much simpler, looks good to me 👍

src/Symfony/Component/Console/ConsoleEvents.php
```diff
@@ -40,6 +41,14 @@ final class ConsoleEvents
      */
     public const SIGNAL = 'console.signal';
 
+    /**
+     * The ALARM event allows you to perform some actions
+     * after the command received a SIGALRM signal.
+     *
+     * @Event("Symfony\Component\Console\Event\ConsoleAlarmEvent")
+     */
+    public const ALARM = 'console.alarm';
```
stof: I'm wondering whether we should add an alias for the ConsoleAlarmEvent as `console.alarm`. Those aliases are needed for events that existed before our migration to PSR-14, but new events should not need them

src/Symfony/Component/Console/Application.php
```diff
@@ -128,6 +130,19 @@ public function setSignalsToDispatchEvent(int ...$signalsToDispatchEvent): void
         $this->signalsToDispatchEvent = $signalsToDispatchEvent;
     }
 
+    public function setAlarmInterval(?int $interval): void
```
stof: this needs some phpdoc explaining what is the unit of that internal.
Alternatively, rename the parameter to `$seconds` (same name than in the SignalRegistry)

HypeMC: Done, https://github.com/symfony/symfony/compare/791685d099b00f050785f3010ab289cf06567122..33fb7b7a503113689e9ce3474e21b9e120865d39

src/Symfony/Component/Console/ConsoleEvents.php
HypeMC: I'm fine either way, but as far as I can tell, currently all built-in events have an alias, maybe keep it for consistency?

stof: that's not true. Messenger events don't have aliases for instance.

Existing console events have aliases because they are old events (i.e. they already existed in Symfony 4)

HypeMC: True, I missed those. In any case, if needed, I'll remove the alias.

chalasr: Yes please remove it, those aliases are mostly legacy

HypeMC: Done https://github.com/symfony/symfony/compare/33fb7b7a503113689e9ce3474e21b9e120865d39..5541f0095dfeeedab95cfba8b585d50cd90400bb

fabpot (approved):

fabpot:
Thank you @HypeMC.

dkarlovi:
Thanks for working on this @HypeMC!

PR #53534, opened 2024-01-13
Trigger WebhookSentEvent after acquiring a Webhook response

### LINKED ISSUE #50932: Webhook sent events
### Description

As a webhook producer I need to track the responses of each webhook.

Maybe I want to provide a UI for a user to debug calls.
Maybe I want to disable the webhook if it fails too often.
...

For this, I need to be informed of the result of a webhook, I need something with :
- the webhook id
- the event
- the response

### Example

I suggest we add an event, (here `WebhookSentEvent`) that will be dispatched in the transport (because it is the only place we are having access to the response).

```diff
<?php

namespace Symfony\Component\Webhook\Server;

+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
use Symfony\Component\HttpClient\HttpOptions;
use Symfony\Component\RemoteEvent\RemoteEvent;
use Symfony\Component\Webhook\Subscriber;
use Symfony\Contracts\HttpClient\HttpClientInterface;

class Transport implements TransportInterface
{
    public function __construct(
        private readonly HttpClientInterface $client,
        private readonly RequestConfiguratorInterface $headers,
        private readonly RequestConfiguratorInterface $body,
        private readonly RequestConfiguratorInterface $signer,
+        private readonly EventDispatcherInterface $eventDispatcher,
    ) {
    }

    public function send(Subscriber $subscriber, RemoteEvent $event): void
    {
        $options = new HttpOptions();

        $this->headers->configure($event, $subscriber->getSecret(), $options);
        $this->body->configure($event, $subscriber->getSecret(), $options);
        $this->signer->configure($event, $subscriber->getSecret(), $options);

-        $this->client->request('POST', $subscriber->getUrl(), $options->toArray());
+        $response = $this->client->request('POST', $subscriber->getUrl(), $options->toArray());
+
+        $this->eventDispatcher->dispatch(new WebhookSentEvent($subscriber, $event, $response));
    }
}
```

yann-eugone:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #50932
| License       | MIT

As a webhook producer I need to track the responses of each webhook.

Either you want to
- provide a UI for a user to debug calls
- disable the webhook if it fails too often
- ...

You will need the acquire the response received when the webhook is called.
But for the moment that `ResponseInterface` object was lost in the `Transport`.
This PR adds a soft dependency to `EventDispatcherInterface`, and trigger a `WebhookSentEvent`, wrapping all available information at that time.

### DIFF
--- a/src/Symfony/Component/Webhook/Server/Transport.php
+++ b/src/Symfony/Component/Webhook/Server/Transport.php
@@ -11,8 +11,10 @@
 
 namespace Symfony\Component\Webhook\Server;
 
+use Psr\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\HttpClient\HttpOptions;
 use Symfony\Component\RemoteEvent\RemoteEvent;
+use Symfony\Component\Webhook\Event\WebhookSentEvent;
 use Symfony\Component\Webhook\Subscriber;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 
@@ -26,6 +28,7 @@ class Transport implements TransportInterface
         private readonly RequestConfiguratorInterface $headers,
         private readonly RequestConfiguratorInterface $body,
         private readonly RequestConfiguratorInterface $signer,
+        private readonly ?EventDispatcherInterface $eventDispatcher = null,
     ) {
     }
 
@@ -37,6 +40,8 @@ class Transport implements TransportInterface
         $this->body->configure($event, $subscriber->getSecret(), $options);
         $this->signer->configure($event, $subscriber->getSecret(), $options);
 
-        $this->client->request('POST', $subscriber->getUrl(), $options->toArray());
+        $response = $this->client->request('POST', $subscriber->getUrl(), $options->toArray());
+
+        $this->eventDispatcher?->dispatch(new WebhookSentEvent($subscriber, $event, $response));
     }
 }
--- a/src/Symfony/Bundle/FrameworkBundle/Resources/config/webhook.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Resources/config/webhook.php
@@ -27,6 +27,7 @@ return static function (ContainerConfigurator $container) {
                 service('webhook.headers_configurator'),
                 service('webhook.body_configurator.json'),
                 service('webhook.signer'),
+                service('event_dispatcher')->nullOnInvalid(),
             ])
 
         ->set('webhook.headers_configurator', HeadersConfigurator::class)
--- a/src/Symfony/Component/Webhook/CHANGELOG.md
+++ b/src/Symfony/Component/Webhook/CHANGELOG.md
@@ -1,6 +1,13 @@
 CHANGELOG
 =========
 
+7.1
+---
+
+ * Added `Symfony\Component\Webhook\Event\WebhookSentEvent`
+ * Added an optional dependency to `Psr\EventDispatcher\EventDispatcherInterface` in `Symfony\Component\Webhook\Server\Transport`
+ * Dispatch `WebhookSentEvent` after a webhook is called
+
 6.4
 ---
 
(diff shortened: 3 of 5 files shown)


smnandre:
I like this new event :)  But do we need the full ResponseInterface object ? Do we even need the response ? If you want to debug, you can decorate the transport and use traceable request / responses. 

I believe in similar a situation (mailer / notifier transports) the response is not attached to the event, but just the strict minimal information (an id, a status..)

In your current implementation, it's either we use this new event with a full response instance (and all the security / privacy questions related) .. or we don't use the event at all. 

wdyt ?

94noni:
nice to see you opened a PR @yann-eugone :)
i tend to agree with what @smnandre said already, as well as seeing for example the messenger component having such event logic (at start, if failed, etc)
perhaps lets wait what others have to say and also the need to be filled here :)

yann-eugone:
I might not add the whole response, maybe I can extract some information from it, for the cases I documented, I believe that the status will be enough.

For instance, if I'm asked to build a gitlab like UI to monitor the registered webhooks
https://docs.gitlab.com/ee/user/project/integrations/webhooks.html#troubleshooting
it is required that I can access the whole response to save response headers, maybe response content, ...

Why do you think it's bad to add the response in the Event?

src/Symfony/Component/Webhook/Server/Transport.php
```diff
@@ -37,6 +40,8 @@ public function send(Subscriber $subscriber, RemoteEvent $event): void
         $this->body->configure($event, $subscriber->getSecret(), $options);
         $this->signer->configure($event, $subscriber->getSecret(), $options);
 
-        $this->client->request('POST', $subscriber->getUrl(), $options->toArray());
+        $response = $this->client->request('POST', $subscriber->getUrl(), $options->toArray());
+
+        $this->eventDispatcher?->dispatch(new WebhookSentEvent($subscriber, $event, $response));
```
stof, line 45: this way this code is written, the event will be dispatched before knowing whether the webhook is actually sent successfully, given that the no blocking call is done on the response before triggering the event.

yann-eugone: I'm not sure to understand your point
Given the event has the Response object, whenever someone will ask for (for example) the status, the response will be fulfilled, am I right?

smnandre: > Responses are always asynchronous, so that the call to the method returns immediately instead of waiting to receive the response
(https://symfony.com/doc/current/http_client.html#making-requests)

Here, that means that you would dispatch a WebhooSentEvent before even knowing if it was ... sent (response could be a 500)

Even more, as long as they don't access the response in the event, EventListener could run code _before_ the HttpClient raises some error/exceptions.

To avoid this, the only safe solution would be to trigger the event _after_ checking the response headers... but then the performance cost would be (imho) too high for an optional feature.

stof: Well, in the existing code, the Transport still waits until the response headers are available (as it does not keep track of the response, its `__destruct` method is called by the time the `request` method returns it, and the `__destruct` method enforces the error handling behavior). this is not a fire-and-forget call for which we hope it will work.

smnandre: Ok, thank you for this correction !

yann-eugone:
I'm not sure to known what I should do here
Maintainers seem to not want it

Still, the requirement for monitoring webhook remain a thing
Not everyone will want to do it, still, the framework should provide a way to do it
I would like to contribute to this optional feature, and I wonder how you would code it, so my PR has a chance to be merged one day

PR #53540, opened 2024-01-15
chore: use _ syntax for `PHP_VERSION_ID` comparison

keradus:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no 
| Deprecations? | no 
| Issues        | Fix CS 
| License       | MIT

I believe it's easier to read the version, as you visually consider `_` as `.` and have 8.1.0 easily read than 80100

### DIFF
--- a/src/Symfony/Bundle/WebProfilerBundle/Profiler/CodeExtension.php
+++ b/src/Symfony/Bundle/WebProfilerBundle/Profiler/CodeExtension.php
@@ -123,7 +123,7 @@ final class CodeExtension extends AbstractExtension
             // highlight_file could throw warnings
             // see https://bugs.php.net/25725
             $code = @highlight_file($file, true);
-            if (\PHP_VERSION_ID >= 80300) {
+            if (\PHP_VERSION_ID >= 8_03_00) {
                 // remove main pre/code tags
                 $code = preg_replace('#^<pre.*?>\s*<code.*?>(.*)</code>\s*</pre>#s', '\\1', $code);
                 // split multiline code tags
--- a/src/Symfony/Component/Clock/Clock.php
+++ b/src/Symfony/Component/Clock/Clock.php
@@ -71,7 +71,7 @@ final class Clock implements ClockInterface
      */
     public function withTimeZone(\DateTimeZone|string $timezone): static
     {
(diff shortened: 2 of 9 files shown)


derrabus (commented):
I like it, but I'd prefer to have a CS Fixer rule for that.

keradus:
There will be followup PR , but PHP Version ID are special, thus heads up PR for it.

stof:
having a fixer handling the PHP version id automatically still make sense as a request though.

keradus:
It would be quite a guessing if 80000 is expected to be PHP 8.0 or simply 80k value. If you have a good proposal how to solve this, please go ahead and raise the PR ;)
I guess it could be partially solved by looking for surroundings (eg PHP_VERSION_ID > number), it will solve some (but not all) cases.

anyway, should not block this PR.

stof:
AFAICT, checking the surrounding to identify comparisons with `PHP_VERSION_ID` with operators `>`, `>=`, `<` or `<=` should identify all the cases in this PR.

keradus:
cases of those PR - yes. cases of what I saw in userland - we have more cases.
Anyway, solving some is better than not solving any, feel free to raise such PR if you up for it.

Anything further I can support to get this PR merged?
I hope it improves the state of codebase in the end ;)

symfonyaml (approved):

OskarStark:
I don't think this is an improvement as `8_03` means `8.3` and not `8.03`... So 👎 from my side.

derrabus:
> as `8_03` means `8.3` and not `8.03`

I don't understand this comment. 🙈

OskarStark:
What I mean is, that it does not improve readability, `80300` is unreadable and `8_03_00` too 🤷

Plus we would need to make some context stunts to get the info if it is a version and plus most of the PHP community uses `80300`.

nicolas-grekas:
> 80300 is unreadable and 8_03_00 too

I have the same feeling :) so I wouldn't change anything

xabbuh:
I agree with Oskar and Nicolas. Let's keep it as is.

nicolas-grekas:
Let's close, thanks for proposing.

PR #53550, opened 2024-01-16
[FrameworkBundle][HttpClient] Add `ThrottlingHttpClient` to limit requests within a timeframe

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

This PR adds a simple `ThrottlingHttpClient` to help with limiting the number of requests within a certain period.

Simple example, don't send more than 10 requests in 5 second:

```yaml
framework:
    http_client:
        scoped_clients:
            ping.client:
                base_uri: 'http://localhost:8080'
                rate_limiter: http_throttling

    rate_limiter:
        http_throttling:
            policy: 'token_bucket'
            limit: 10
            rate: { interval: '5 seconds', amount: 10 }
```

```php
#[AsCommand('app:ping')]
class PingCommand extends Command
{
    public function __construct(
        #[Target('ping.client')] private HttpClientInterface $httpClient,
    ) {
        parent::__construct();
    }

    {
        $requests = [];
        for ($i = 0; $i < 100; $i++) {
            $requests[] = $this->httpClient->request('GET', '/ping');
        }

        foreach ($requests as $request) {
            $output->writeln($request->getContent());
        }

        return Command::SUCCESS;
    }
}
```

Receiving controller:

```php
#[AsController]
class PingController
{
    #[Route('/ping', name: 'app_ping')]
    public function __invoke(): Response
    {
        return new Response((new \DateTime())->format('Y-m-d\TH:i:s.u'));
    }
}
```

Output:

```
$ bin/console app:ping

2024-01-16T11:52:54.922597
2024-01-16T11:52:54.964851
2024-01-16T11:52:55.009504
2024-01-16T11:52:55.053986
2024-01-16T11:52:55.098267
2024-01-16T11:52:55.139621
2024-01-16T11:52:55.182769
2024-01-16T11:52:55.264433
2024-01-16T11:52:55.304775
2024-01-16T11:52:55.223805

2024-01-16T11:52:59.877858
2024-01-16T11:52:59.914724
2024-01-16T11:52:59.950610
2024-01-16T11:52:59.986818
2024-01-16T11:53:00.025638
2024-01-16T11:53:00.063687
2024-01-16T11:53:00.879850
2024-01-16T11:53:00.919651
2024-01-16T11:53:00.959273
2024-01-16T11:53:00.999381

2024-01-16T11:53:04.890126
2024-01-16T11:53:04.965116
2024-01-16T11:53:04.999916
2024-01-16T11:53:05.037780
2024-01-16T11:53:05.073999
2024-01-16T11:53:05.110241
2024-01-16T11:53:05.917045
2024-01-16T11:53:05.971672
2024-01-16T11:53:06.016025
2024-01-16T11:53:06.059220

2024-01-16T11:53:09.920092
2024-01-16T11:53:09.974757
2024-01-16T11:53:10.010406
2024-01-16T11:53:10.045790
2024-01-16T11:53:10.081749
2024-01-16T11:53:10.118100
2024-01-16T11:53:10.924485
2024-01-16T11:53:10.981094
2024-01-16T11:53:11.023935
2024-01-16T11:53:11.067210

...
```

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
@@ -1715,17 +1715,32 @@ class Configuration implements ConfigurationInterface
                     ->fixXmlConfig('scoped_client')
                     ->beforeNormalization()
                         ->always(function ($config) {
-                            if (empty($config['scoped_clients']) || !\is_array($config['default_options']['retry_failed'] ?? null)) {
+                            if (empty($config['scoped_clients'])) {
+                                return $config;
+                            }
+
+                            $hasDefaultRateLimiter = isset($config['default_options']['rate_limiter']);
+                            $hasDefaultRetryFailed = \is_array($config['default_options']['retry_failed'] ?? null);
+
+                            if (!$hasDefaultRateLimiter && !$hasDefaultRetryFailed) {
                                 return $config;
                             }
 
                             foreach ($config['scoped_clients'] as &$scopedConfig) {
-                                if (!isset($scopedConfig['retry_failed']) || true === $scopedConfig['retry_failed']) {
-                                    $scopedConfig['retry_failed'] = $config['default_options']['retry_failed'];
-                                    continue;
+                                if ($hasDefaultRateLimiter) {
+                                    if (!isset($scopedConfig['rate_limiter']) || true === $scopedConfig['rate_limiter']) {
+                                        $scopedConfig['rate_limiter'] = $config['default_options']['rate_limiter'];
+                                    } elseif (false === $scopedConfig['rate_limiter']) {
+                                        $scopedConfig['rate_limiter'] = null;
+                                    }
                                 }
-                                if (\is_array($scopedConfig['retry_failed'])) {
-                                    $scopedConfig['retry_failed'] += $config['default_options']['retry_failed'];
+
+                                if ($hasDefaultRetryFailed) {
+                                    if (!isset($scopedConfig['retry_failed']) || true === $scopedConfig['retry_failed']) {
+                                        $scopedConfig['retry_failed'] = $config['default_options']['retry_failed'];
+                                    } elseif (\is_array($scopedConfig['retry_failed'])) {
+                                        $scopedConfig['retry_failed'] += $config['default_options']['retry_failed'];
+                                    }
                                 }
                             }
 
@@ -1829,6 +1844,10 @@ class Configuration implements ConfigurationInterface
                                     ->info('Extra options for specific HTTP client')
                                     ->normalizeKeys(false)
                                     ->variablePrototype()->end()
+                                ->end()
+                                ->scalarNode('rate_limiter')
+                                    ->defaultNull()
+                                    ->info('Rate limiter name to use for throttling requests')
                                 ->end()
                                 ->append($this->createHttpClientRetrySection())
                             ->end()
@@ -1977,6 +1996,10 @@ class Configuration implements ConfigurationInterface
                                         ->info('Extra options for specific HTTP client')
                                         ->normalizeKeys(false)
                                         ->variablePrototype()->end()
+                                    ->end()
+                                    ->scalarNode('rate_limiter')
+                                        ->defaultNull()
+                                        ->info('Rate limiter name to use for throttling requests')
                                     ->end()
                                     ->append($this->createHttpClientRetrySection())
                                 ->end()
--- a/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
+++ b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
@@ -85,6 +85,7 @@ use Symfony\Component\HttpClient\MockHttpClient;
 use Symfony\Component\HttpClient\Retry\GenericRetryStrategy;
 use Symfony\Component\HttpClient\RetryableHttpClient;
 use Symfony\Component\HttpClient\ScopingHttpClient;
+use Symfony\Component\HttpClient\ThrottlingHttpClient;
 use Symfony\Component\HttpClient\UriTemplateHttpClient;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Attribute\AsController;
@@ -345,6 +346,7 @@ class FrameworkExtension extends Extension
         }
 
         if ($this->readConfigEnabled('http_client', $container, $config['http_client'])) {
+            $this->readConfigEnabled('rate_limiter', $container, $config['rate_limiter']); // makes sure that isInitializedConfigEnabled() will work
             $this->registerHttpClientConfiguration($config['http_client'], $container, $loader);
         }
 
@@ -2409,6 +2411,8 @@ class FrameworkExtension extends Extension
         $loader->load('http_client.php');
 
         $options = $config['default_options'] ?? [];
+        $rateLimiter = $options['rate_limiter'] ?? null;
+        unset($options['rate_limiter']);
         $retryOptions = $options['retry_failed'] ?? ['enabled' => false];
         unset($options['retry_failed']);
         $defaultUriTemplateVars = $options['vars'] ?? [];
@@ -2428,6 +2432,10 @@ class FrameworkExtension extends Extension
             $container->removeDefinition('httplug.http_client');
             $container->removeAlias(HttpAsyncClient::class);
             $container->removeAlias(HttpClient::class);
+        }
+
+        if (null !== $rateLimiter) {
+            $this->registerThrottlingHttpClient($rateLimiter, 'http_client', $container);
         }
 
         if ($this->readConfigEnabled('http_client.retry_failed', $container, $retryOptions)) {
@@ -2451,6 +2459,8 @@ class FrameworkExtension extends Extension
 
             $scope = $scopeConfig['scope'] ?? null;
             unset($scopeConfig['scope']);
+            $rateLimiter = $scopeConfig['rate_limiter'] ?? null;
+            unset($scopeConfig['rate_limiter']);
             $retryOptions = $scopeConfig['retry_failed'] ?? ['enabled' => false];
             unset($scopeConfig['retry_failed']);
 
@@ -2468,6 +2478,10 @@ class FrameworkExtension extends Extension
                     ->setArguments([new Reference('http_client.transport'), [$scope => $scopeConfig], $scope])
                     ->addTag('http_client.client')
                 ;
+            }
+
+            if (null !== $rateLimiter) {
+                $this->registerThrottlingHttpClient($rateLimiter, $name, $container);
             }
 
             if ($this->readConfigEnabled('http_client.scoped_clients.'.$name.'.retry_failed', $container, $retryOptions)) {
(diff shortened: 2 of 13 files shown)


src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -2506,6 +2519,22 @@ private function registerHttpClientConfiguration(array $config, ContainerBuilder
         }
     }
 
+    private function registerThrottlingHttpClient(string $rateLimiter, string $name, ContainerBuilder $container): void
+    {
+        if (!class_exists(ThrottlingHttpClient::class)) {
+            throw new LogicException('Rate limiter support cannot be enabled as version 7.1+ of the HttpClient component is required.');
+        }
+
+        if (!interface_exists(LimiterInterface::class)) {
```
stof: this should actually check whether the rate_limiter is enabled in the FrameworkBundle config, not whether the component is installed, as installing it is not enough (no need to check again for the component being installed as enabling the rate_limiter without installing it is already checked when handling the rate_limiter config).
See `readConfigEnabled` for the checks about that.

stof (changes requested):

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
HypeMC: Done

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/yml/http_client_rate_limiter.yml
```diff
@@ -0,0 +1,19 @@
...
+    http_method_override: false
+    handle_all_throwables: true
+    php_errors:
+        log: true
+    rate_limiter:
+        foo_limiter:
+            lock_factory: null
+            policy: token_bucket
+            limit: 10
+            rate: { interval: '5 second', amount: 10 }
```
OskarStark:
```suggestion
            rate: { interval: '5 seconds', amount: 10 }
```

src/Symfony/Component/HttpClient/Tests/ThrottlingHttpClientTest.php
```diff
@@ -0,0 +1,55 @@
...
+                new MockResponse('', ['http_code' => 200, 'pause_handler' => $failPauseHandler]),
+                new MockResponse('', ['http_code' => 200, 'pause_handler' => $pauseHandler(5)]),
+                new MockResponse('', ['http_code' => 200, 'pause_handler' => $pauseHandler(5)]),
+                new MockResponse('', ['http_code' => 200, 'pause_handler' => $pauseHandler(10)]),
+            ]),
+            new RateLimiterFactory([
+                'id' => 'token_bucket',
+                'policy' => 'token_bucket',
+                'limit' => 2,
+                'rate' => ['interval' => '5 second', 'amount' => 2],
```
OskarStark:
```suggestion
                'rate' => ['interval' => '5 seconds', 'amount' => 2],
```

src/Symfony/Bundle/FrameworkBundle/Tests/DependencyInjection/Fixtures/yml/http_client_rate_limiter.yml
HypeMC: Done

src/Symfony/Component/HttpClient/Tests/ThrottlingHttpClientTest.php
HypeMC: Done

src/Symfony/Component/HttpClient/ThrottlingHttpClient.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+namespace Symfony\Component\HttpClient;
+
+use Symfony\Component\RateLimiter\LimiterInterface;
+use Symfony\Component\RateLimiter\RateLimiterFactory;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+use Symfony\Contracts\HttpClient\ResponseInterface;
+
+class ThrottlingHttpClient implements HttpClientInterface
```
rmikalkenas, line 19: As with all http client decorators - `Symfony\Contracts\Service\ResetInterface` has to be implemented

nicolas-grekas:
Can you please help me get the steps that achieve the throttling? Which part is sleeping especially? The lock is, right?
Do we need to play with the pause handler then? I think this can be checked by comparing how curl vs native http clients behave with the current code.

HypeMC:
@nicolas-grekas Actually no, `$this->rateLimiter()->reserve()->getWaitDuration()` doesn't sleep, it just returns the expected delay between the requests. The pause handler is what actually sleeps. This behavior is consistent across both the curl and native http clients. Maybe I'm wrong, but if the lock was what slept, then that would mean that the creation of the response objects was delayed, but not the actual requests which are triggered afterwards.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
```diff
@@ -1716,17 +1716,32 @@ private function addHttpClientSection(ArrayNodeDefinition $rootNode, callable $e
                     ->fixXmlConfig('scoped_client')
                     ->beforeNormalization()
                         ->always(function ($config) {
-                            if (empty($config['scoped_clients']) || !\is_array($config['default_options']['retry_failed'] ?? null)) {
+                            if (empty($config['scoped_clients'])) {
+                                return $config;
+                            }
+
+                            $hasDefaultRateLimiter = null !== ($config['default_options']['rate_limiter'] ?? null);
```
```suggestion
                            $hasDefaultRateLimiter = isset($config['default_options']['rate_limiter']);
```

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
```diff
@@ -345,6 +346,7 @@ public function load(array $configs, ContainerBuilder $container): void
         }
 
         if ($this->readConfigEnabled('http_client', $container, $config['http_client'])) {
+            $this->readConfigEnabled('rate_limiter', $container, $config['rate_limiter']); // warmup config enabled
```
nicolas-grekas: what does the comment mean?

src/Symfony/Component/HttpClient/ThrottlingHttpClient.php
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Symfony\Component\HttpClient;
+
+use Symfony\Component\RateLimiter\LimiterInterface;
+use Symfony\Component\RateLimiter\RateLimiterFactory;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+use Symfony\Contracts\HttpClient\ResponseInterface;
+use Symfony\Contracts\Service\ResetInterface;
+
+class ThrottlingHttpClient implements HttpClientInterface, ResetInterface
```
nicolas-grekas, line 22: let's add a small title in a phpdoc ?

src/Symfony/Component/HttpClient/ThrottlingHttpClient.php
```diff
@@ -0,0 +1,56 @@
...
+
+        return $response;
+    }
+
+    private function rateLimiter(): LimiterInterface
+    {
+        return $this->rateLimiter ??= $this->rateLimiterFactory->create();
+    }
+
+    public function reset(): void
```
nicolas-grekas, line 46: Where is this method registered? If it's not, what about removing it? I'm not sure we need it: the factory already has a reset method, that's the one that should be used when needing a reset to me.

nicolas-grekas (commented):
Thanks for the insights, makes sense now.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
HypeMC: It makes sure that `isInitializedConfigEnabled()` will work, see https://github.com/symfony/symfony/pull/53550/files#diff-e4f52d5033296421d02e360c8355aa1931bdfb0c12579f4815ea02b724322c0cR2529

OskarStark: Maybe use this as comment instead?

src/Symfony/Component/HttpClient/ThrottlingHttpClient.php
HypeMC: Did you mean the `LimiterInterface`? It has a `reset()` method, but as far as I can tell, it needs to be called manually.

nicolas-grekas: I mean AbstractRequestRateLimiter::reset, which implements what's needed.

nicolas-grekas: I also mean this might be considered as an external concern from the pov of ThrottlingHttpClient - it doesn't need to care if the rate limiter is resettable or not.

HypeMC: I understand your point about it not being a concern of `ThrottlingHttpClient`, but the `AbstractRequestRateLimiter` seems to be something completely different, it's part of the `HttpFoundation` component and used in `DefaultLoginRateLimiter`.

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php
HypeMC: Done

src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php
HypeMC: Done

src/Symfony/Component/HttpClient/ThrottlingHttpClient.php
HypeMC: Done

src/Symfony/Component/HttpClient/ThrottlingHttpClient.php
nicolas-grekas: Would it make sense to pass a LimiterInterface directly then instead of a factory?

HypeMC: Not sure it would make a difference, the limiter itself is not registered as a service, only the factory is, so we would still need to call `reset()` manually.

nicolas-grekas: bu the implementation would be less hacky: `$this->limiter->reset()` would do it.

HypeMC: Done https://github.com/symfony/symfony/compare/84c892749c3555a2c5f9fc0d18cae19e8d2064c9..1a8b068d6149cc40631dd9136f2a3279bbd127d1

nicolas-grekas (approved):

fabpot:
Thank you @HypeMC.

PR #53556, opened 2024-01-17
[Cache] fix Redis proxies tests

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | phpredis/phpredis#2423, phpredis/phpredis#2434
| License       | MIT


stof:
This change looks wrong to me. It does not assert anymore that the committed proxy is compatible with Redis anymore, which is what the test was about.

xabbuh:
It still does. We only replace two return types where we know that the change we do is compatible. Or do I miss anything?

stof:
this will be compatible in term of signature in the child class. But it won't work because the proxy returns the value coming from the actual class and so it cannot guarantee it will return a narrower type. It will fail with a TypeError at runtime when Redis returns one of those "new" values (they are not really new, as before, Redis was still returning them).

xabbuh:
Ah indeed, I missed that not only the PHP stubs have been updated. Here is another approach: #53578

PR #47642, opened 2022-09-21
[DomCrawler] impossible to test a value out of the options in ChoiceFormField

jorissae:
| Q             | A
| ------------- | ---
| Branch?       | 6.2
| Bug fix?      | No
| New feature?  | No
| Deprecations? | No
| Tickets       | >
| License       | MIT
| Doc PR        | 

I couldn't test a value outside of the ChoiceForm field option nodes.
For example, even if my combobox does not offer the empty field, I want to test my NotBlank validator.

```php
<?php
$form = $crawler->selectButton('ea[newForm][btn]')->form();
$client->submit($form, [
    'Item[vat]' => ''
]);
```

in Symfony\Component\BrowserKit\AbstractBrowser::submit ($client->submit())
````php
public function submit(Form $form, array $values = [], array $serverParameters = []): Crawler
{
        $form->setValues($values);

        return $this->request($form->getMethod(), $form->getUri(), $form->getPhpValues(), $form->getPhpFiles(), $serverParameters);
}
````

`InvalidArgumentException: Input "Item[vat]" cannot take "" as a value (possible values: "0", "700", "1900").`

maybe there is another solution ? (like add arguments bool $throw = true to setValues or another ...)

Thanks

PR #53559, opened 2024-01-17
[DomCrawler] Allow to add choices to single select

norkunas:
| Q             | A
| ------------- | ---
| Branch?       | 8.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | https://github.com/symfony/ux/issues/1334#issuecomment-1854470006 https://github.com/symfony/ux/issues/1334#issuecomment-1855811337 https://github.com/symfony/symfony/pull/47642
| License       | MIT

Not sure why this was designed this way, but when using JS autocomplete's we cannot simulate it's behavior with a select element when a new choice was added to the list

cc @yceruto :pray:

### DIFF
--- a/src/Symfony/Component/DomCrawler/Field/ChoiceFormField.php
+++ b/src/Symfony/Component/DomCrawler/Field/ChoiceFormField.php
@@ -141,20 +141,18 @@ class ChoiceFormField extends FormField
     /**
      * Adds a choice to the current ones.
      *
-     * @throws \LogicException When choice provided is not multiple nor radio
-     *
-     * @internal
+     * @throws \LogicException When choice provided is neither multiple, radio nor select
      */
-    public function addChoice(\DOMElement $node): void
+    final public function addChoice(\DOMElement $node): void
     {
-        if (!$this->multiple && 'radio' !== $this->type) {
-            throw new \LogicException(\sprintf('Unable to add a choice for "%s" as it is not multiple or is not a radio button.', $this->name));
+        if (!$this->multiple && !\in_array($this->type, ['radio', 'select'], true)) {
+            throw new \LogicException(\sprintf('Unable to add a choice for "%s" as it is neither multiple, a radio button nor a select field.', $this->name));
         }
 
         $option = $this->buildOptionValue($node);
         $this->options[] = $option;
 
-        if ($node->hasAttribute('checked')) {
+        if ($node->hasAttribute('select' === $this->type ? 'selected' : 'checked')) {
             $this->value = $option['value'];
         }
     }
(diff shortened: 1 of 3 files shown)


src/Symfony/Component/DomCrawler/Tests/Field/ChoiceFormFieldTest.php
```diff
@@ -110,6 +110,18 @@ public function testSelects()
         } catch (\InvalidArgumentException $e) {
             $this->assertTrue(true, '->setValue() throws an \InvalidArgumentException if the value is an array');
         }
+
+        $option = $this->createNode('option', 'Hello World', ['value' => 'hello_world']);
```
stof, line 120: Please create a separate test instead of adding more things in the same test. This provides a much better experience:

- we have a test name telling us what this is about
- the status of the test is separate, giving us better failures

Existing dom-crawler tests are not following the testing best practices because this is one of the oldest components of Symfony, at a time where the usage of phpunit in Symfony was still quite new.

stof:
As this method is `@internal`, I don't think it is the solution to the use case anyway.

norkunas:
> As this method is `@internal`, I don't think it is the solution to the use case anyway.

What would you suggest then? Can we remove `@internal` to allow people simulate this behavior?

src/Symfony/Component/DomCrawler/Field/ChoiceFormField.php
```diff
@@ -143,14 +143,14 @@ public function setValue(string|array|bool|null $value): void
      */
     public function addChoice(\DOMElement $node): void
     {
-        if (!$this->multiple && 'radio' !== $this->type) {
+        if (!$this->multiple && ('radio' !== $this->type && 'select' !== $this->type)) {
             throw new \LogicException(sprintf('Unable to add a choice for "%s" as it is not multiple or is not a radio button.', $this->name));
```
smnandre: Message (and method docblock) should be updated. 

@stof  Would that be a BC break, as select were refused before ?

src/Symfony/Component/DomCrawler/Field/ChoiceFormField.php
```diff
@@ -143,14 +143,14 @@ public function setValue(string|array|bool|null $value): void
...
         }
 
         $option = $this->buildOptionValue($node);
         $this->options[] = $option;
 
-        if ($node->hasAttribute('checked')) {
+        if ($node->hasAttribute('checked') || $node->hasAttribute('selected')) {
```
smnandre: type === "radio" a&& attribute checked || type === select && attribute selected ?

src/Symfony/Component/DomCrawler/Field/ChoiceFormField.php
norkunas: @smnandre given that's an `@internal` method BC break does not apply here

nicolas-grekas (approved):
I rebased the PR and removed the `@internal` annotation on `ChoiceFormField::addChoice()`. I made the method final instead to ease it maintenance a bit.
I also addressed remaining comments.

yceruto (approved):

nicolas-grekas:
Thank you @norkunas.

PR #53565, opened 2024-01-17
[Mime] Fix undefined array key 0 when empty sender

0x346e3730:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | N/A
| License       | MIT

Today I managed to produce the following error :
image

From the following code :
```php
$email = (new TemplatedEmail())
            ->from()
            ->to($member->getEmail())
            ->subject('Adhésion Sauvade')
            ->htmlTemplate('emails/new-membership/index.html.twig')
            ->context([
                'member' => $member,
            ])
            ->addPart(new DataPart(
                $this->pdfService->generatePdfFromTemplate(
                    'emails/new-membership/attachments/recu.html.twig',
                    ['member' => $member]
                ),
            ))
            ->addPart(new DataPart(
                $this->pdfService->generatePdfFromTemplate(
                    'emails/new-membership/attachments/membership-card.html.twig',
                    ['member' => $member]
                ),
            ))
        ;

        $this->mailer->send($email);
```

This happens because the `from` method from the `Symfony\Component\Mime\Email` class adds the header even if the array of addresses is empty. I identified two ways of fixing this :
- Adding a check in the `Symfony\Component\Mime\Message` class before accessing the key 0 and throw a better exception in case the adresses are empty
- Modifying the `from` method so that it doesn't add the header if the adresses are empty; or throw an exception when trying to do so.

I'm unsure what method is the cleaner, I thought that fixing the Message access of an undefined key was the widest fix (as I guess it could happen from somewhere else than the Mailer component). Also not sure how to test that and wanted to make sure this fix was good before testing it.

### DIFF
--- a/src/Symfony/Component/Mime/Email.php
+++ b/src/Symfony/Component/Mime/Email.php
@@ -121,6 +121,10 @@ class Email extends Message
      */
     public function from(...$addresses)
     {
+        if (!$addresses) {
+            throw new LogicException('"from()" must be called with at least one address.');
+        }
+
         return $this->setListAddressHeaderBody('From', $addresses);
     }
 
--- a/src/Symfony/Component/Mime/Message.php
+++ b/src/Symfony/Component/Mime/Message.php
@@ -140,7 +140,10 @@ class Message extends RawMessage
         if ($this->headers->has('Sender')) {
             $sender = $this->headers->get('Sender')->getAddress();
         } elseif ($this->headers->has('From')) {
-            $sender = $this->headers->get('From')->getAddresses()[0];
+            if (!$froms = $this->headers->get('From')->getAddresses()) {
+                throw new LogicException('A "From" header must have at least one email address.');
+            }
+            $sender = $froms[0];
         } else {
             throw new LogicException('An email must have a "From" or a "Sender" header.');
         }
--- a/src/Symfony/Component/Mime/Tests/MessageTest.php
+++ b/src/Symfony/Component/Mime/Tests/MessageTest.php
@@ -13,6 +13,7 @@ namespace Symfony\Component\Mime\Tests;
 
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Mime\Address;
+use Symfony\Component\Mime\Exception\LogicException;
 use Symfony\Component\Mime\Header\Headers;
 use Symfony\Component\Mime\Header\MailboxListHeader;
 use Symfony\Component\Mime\Header\UnstructuredHeader;
@@ -123,6 +124,14 @@ class MessageTest extends TestCase
         $message->getHeaders()->addMailboxListHeader('From', ['fabien@symfony.com', 'lucas@symfony.com']);
         $message->getHeaders()->addMailboxHeader('Sender', 'thomas@symfony.com');
         $this->assertEquals('thomas@symfony.com', $message->getPreparedHeaders()->get('Sender')->getAddress()->getAddress());
+    }
+
+    public function testGenerateMessageIdThrowsWhenHasFromButNoAddresses()
+    {
+        $message = new Message();
+        $message->getHeaders()->addMailboxListHeader('From', []);
+        $this->expectException(LogicException::class);
+        $message->generateMessageId();
     }
 
     public function testToString()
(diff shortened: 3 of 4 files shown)


xabbuh:
Can you please add a test to make sure that we don’t break this again.

0x346e3730:
Done. It seems to me that the failing tests are related to cryptography errors and not my code

src/Symfony/Component/Mime/Tests/MessageTest.php
```diff
@@ -125,6 +125,14 @@ public function testGetPreparedHeadersHasSenderWhenNeeded()
         $this->assertEquals('thomas@symfony.com', $message->getPreparedHeaders()->get('Sender')->getAddress()->getAddress());
     }
 
+    public function testGenerateMessageIdThrowsWhenHasFromButNoAddresses()
+    {
+        $this->expectException(\LogicException::class);
```
xabbuh: please move this line right before the call of `generateMessageId()`, also we should change the class name to actually check for the `LogicException` that is thrown in the `Message` class (i.e. the one from the Mime component)

src/Symfony/Component/Mime/Message.php
```diff
@@ -140,7 +140,10 @@ public function generateMessageId(): string
         if ($this->headers->has('Sender')) {
             $sender = $this->headers->get('Sender')->getAddress();
         } elseif ($this->headers->has('From')) {
-            $sender = $this->headers->get('From')->getAddresses()[0];
+            if (empty($froms = $this->headers->get('From')->getAddresses())) {
```
xabbuh:
```suggestion
            if ([] === $froms = $this->headers->get('From')->getAddresses()) {
```

src/Symfony/Component/Mime/Message.php
```diff
@@ -140,7 +140,10 @@ public function generateMessageId(): string
...
+                throw new LogicException('Your email has a From header that contains no addresses. Did you forgot to add emails in the "from" method ?');
```
xabbuh:
```suggestion
                throw new LogicException('A "From" header must have at least one email address.');
```

fabpot:
What about forbidding calling `Email::from()` without any address instead?

0x346e3730:
> What about forbidding calling `Email::from()` without any address instead?

This is one of the ways I identified to fix this issue, I have no opinion on which is best for DX.

I went with fixing it in the Message class because it could (I guess ?) be used by other components than Mailer so I would suggest to still fix this here even if we block this behavior in the `from` method

src/Symfony/Component/Mime/Message.php
fabpot:
```suggestion
            if (!$froms = $this->headers->get('From')->getAddresses()) {
```

fabpot (changes requested):

fabpot:
I would do the fix in `Email::from()` anyway as there is no good use case for not passing anything to this method.

src/Symfony/Component/Mime/Message.php
0x346e3730: Used the proposition of Fabien instead, letting you resolve the conversation if it's okay with you

0x346e3730:
I adjusted the code with your reviews and added an exception if calling Email::from() with no addres, however the `from` method just calls `setListAddressHeaderBody` that is also called by `replyTo`, `to`, `cc`, `bcc` and `addListAddressHeaderBody`. Should we implement the throw in Email::setListAddressHeaderBody() instead so that it covers all method that add email addresses to certain headers ?

Plus the "add" version of those methods (`addFrom`, `addReplyTo` ect...) call `addListAddressHeaderBody` that does `$header->addAddresses(Address::createArray($addresses));`. Should we do the same throw in `MailboxListHeader::addAddresses` as well ?

src/Symfony/Component/Mime/Email.php
```diff
@@ -121,6 +121,10 @@ public function addFrom(...$addresses)
      */
     public function from(...$addresses)
     {
+        if (!$addresses) {
+            throw new LogicException(sprintf('"from()" must be called with at least one address.'));
```
xabbuh:
```suggestion
            throw new LogicException('"from()" must be called with at least one address.');
```

xabbuh (approved):

fabpot (approved):

fabpot:
Thank you @0x346e3730.

PR #53569, opened 2024-01-17
[Validator] Add callback option to the ExpressionSyntax constraint

symfonyaml:
| Q             | A
| ------------- | ---
| Branch?       | 7.2
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

### ExpressionSyntax
I introduced the ability to use a callback to specify allowed variables *(basically it will set the `$allowedVariables` value from this callback)*. Exactly like [the Choice callback constraint](https://symfony.com/doc/current/reference/constraints/Choice.html#callback).

This feature will be useful if an expression use dynamic variables (like in my personal project 😋).

### Example
```php
class MyEntity
{
    public static function getVars(): array
    {
        // some cool logic with allowed expression variables ... 
        return ['symfony', $this->getUsername()];
    }

    #[Assert\ExpressionSyntax(allowedVariablesCallback: 'getVars')]
    protected string $expression;
}
```


src/Symfony/Component/Validator/Constraints/ExpressionSyntax.php
```diff
@@ -30,20 +30,29 @@ class ExpressionSyntax extends Constraint
...
      * @param string[]|null            $groups
      */
-    public function __construct(array $options = null, string $message = null, string $service = null, array $allowedVariables = null, array $groups = null, mixed $payload = null)
-    {
+    public function __construct(
+        array $options = null,
+        string $message = null,
+        string $service = null,
+        array $allowedVariables = null,
+        callable|string $allowedVariablesCallback = null,
```
xabbuh: adding a new parameter in between existing parameters is a BC break

src/Symfony/Component/Validator/Constraints/ExpressionSyntaxValidator.php
```diff
@@ -36,6 +37,10 @@ public function validate(mixed $expression, Constraint $constraint): void
             throw new UnexpectedTypeException($constraint, ExpressionSyntax::class);
         }
 
+        if (\is_array($constraint->allowedVariables) && $constraint->allowedVariablesCallback) {
```
xabbuh, line 40: What about allow `allowedVariables` to also be a callable instead?

xabbuh (changes requested):

src/Symfony/Component/Validator/Constraints/ExpressionSyntax.php
symfonyaml: Oh that's right, thanks for the reminder !

src/Symfony/Component/Validator/Constraints/ExpressionSyntaxValidator.php
symfonyaml: I like that, but we'll have a problem between these two allowed variables setups :
```php
// I mention 2 variables called 'className' and 'method'
#[ExpressionSyntax(allowedVariables: ['className', 'method'])]
// I mention the callback className::method()
#[ExpressionSyntax(allowedVariables: ['className', 'method'])] 
```
In this PR I took example on the [ChoiceValidator callback option](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Validator/Constraints/ChoiceValidator.php#L48).

symfonyaml:
Closed. We can use the `ExpressionSyntax::$service` option instead.

Issue #53539, opened 2024-01-15
1 returned with $request->getLocale() with array path in route attribute.


cyraid:
### Symfony version(s) affected

7.0.2

### Description

`$request->getLocale()` will return `1` when there is an array in `#[Route]`:
```php
  #[Route(['/{_locale}/{path}', '/{path}'], requirements: ['path' => '.*'])]
  public function home(Request $request): Response {
    return $this->render("home/{$request->getLocale()}/index.html");
  } // Function //
```

When route is:
```php
  #[Route('/{_locale}/{path}', requirements: ['path' => '.*'])]
```

`$request->getLocale()` will return `en` as expected.

**framework.yaml:**
```yaml
framework:
    default_locale: en
    enabled_locales: ['en', 'fr']
    set_locale_from_accept_language: true
    set_content_language_from_locale: true
```

### How to reproduce

Should be able to reproduce using the information in the description. You don't have to return a page, just print the result of `$request->getLocale()` if you need to.

### Possible Solution

Might be a bug in getLocale()? That or a locale listener if there is one. I suppose I could detect a number and use 'en' as default, but the method does not return a string, it returns a number which is a bug, as `getLocale()` is marked as `string` and it should return the actual locale and not an index?

xabbuh:
This looks expected to me as the routes are expected to be indexed by locale when passing an array of routes (see https://symfony.com/doc/current/routing.html#localized-routes-i18n).

cyraid:
> This looks expected to me as the routes are expected to be indexed by locale when passing an array of routes (see https://symfony.com/doc/current/routing.html#localized-routes-i18n).

Thank you for the quick reply. Shouldn't it be then:
```php
    /**
     * Gets the locale, or an index if multiple paths are included in the Route.
     */
    public function getLocale(): string|int
    {
        return $this->locale ?? $this->defaultLocale;
    }
```

And if an index is expected, is there an array of locales I can access if number? If the url is `/fr/` `getLocale()` returns actual `fr` instead of an index. If it's meant to be an index with multiple paths, should it not be an index if a _locale is provided?

cyraid:
@xabbuh Regardless, you helped me fix the problem, so **thank you**!

I ended up fixing it with:
```php
#[Route(['/{_locale}/{path}', 'en' => '/{path}'], requirements: ['path' => '.*', '_locale' => 'en|fr'], priority: -1000)]
```

Cheers!

[...]

PR #53571, opened 2024-01-18
Localized Passthrough Locale for #[Route()]

cyraid:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

As per reference of #53539, I was looking for a way to solve:
https://stackoverflow.com/questions/41721408/symfony-default-locale-without-specifying-in-url/77819142

Currently the first path (either specifying a path, or a path array) the first path index is 0 in AttributeClassLoader.php. Taking advantage of this, my answer in the Stack Overflow and in #53539 works, but is not advised as this is not documented. My PR adds support for:

```php
    #[Route(path: ['{_locale}' => '/{_locale}', 'nl' => '/nl'], name: 'action')]
```

Which adds the path as 'action' and 'action.nl', means the `{_locale}` can be used to catch all other locales. As @xabbuh and I talked about. Using this:

```php
    #[Route(path: ['/{_locale}', 'nl' => '/nl'], name: 'action')]
```

Is still allowed as the first array index of 0 was always allowed anyway (as the Route would turn the `$path` into `$paths[]`), and should remain to not break backwards compatibility with anyone using it, but if `1 =>` etc (integer array keys) used, it will raise an exception (also tested in the Tests).

If you guys could look this over that would be great. If someone needs any clarification for anyone wanting to document this feature near https://symfony.com/doc/current/routing.html#localized-routes-i18n that also would be great!

### DIFF
--- a/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/FormLoginLdapFactory.php
+++ b/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/FormLoginLdapFactory.php
@@ -32,7 +32,7 @@ class FormLoginLdapFactory extends FormLoginFactory
         $node
             ->children()
                 ->scalarNode('service')->defaultValue('ldap')->end()
-                ->scalarNode('dn_string')->defaultValue('{username}')->end()
+                ->scalarNode('dn_string')->defaultValue('{user_identifier}')->end()
                 ->scalarNode('query_string')->end()
                 ->scalarNode('search_dn')->defaultValue('')->end()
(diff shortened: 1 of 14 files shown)


cyraid:
I'll wait for merge of PR #53580 AttributeClassLoaderTestCase, and then update this PR.

cyraid:
@xabbuh since the tests for this require the patch from 7.0, and I need to base it off 7.1 for the feature. Would we be able to merge the 7.0 fix to 7.1 and then use the PR if everyone is alright with the idea?

xabbuh:
@cyraid I have merged up the `7.0` branch

PR #53572, opened 2024-01-18
[FrameworkBundle] grab a service from the container only if it exists

### LINKED ISSUE #53551: `cache:pool:clear --all` command fails in a prod environment

7.0.2, 6.4.2

### Description

When trying to clear cache pools using the `cache:pool:clear --all` command while in a production environment (i.e. `APP_ENV=prod`), the command fails with the following error on Symfony 6.4:

```
In Container.php line 249:
                                                                                                                                                                                                              
  The "cache.annotations" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection   
  instead.                                                                                                                                                                                                    
                                                                                                                                                                                                              

cache:pool:clear [--all] [--exclude EXCLUDE] [--] [<pools>...]

Failed to execute command bin/console cache:pool:clear --all: exit status 1
```

And on Symfony 7.0:

```
In Container.php line 240:
                                                                                                                                                                                                               
  The "cache.asset_mapper" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection   
  instead.                                                                                                                                                                                                     
                                                                                                                                                                                                               

cache:pool:clear [--all] [--exclude EXCLUDE] [--] [<pools>...]

Failed to execute command bin/console cache:pool:clear --all: exit status 1
```

Both of these happen only in the prod env, the dev env works perfectly fine.

### How to reproduce

1. Create a new webapp installation of the app from `symfony/skeleton`.
2. Change `APP_ENV` to `prod`.
3. Attempt to run the `cache:pool:clear --all` command.

### Possible Solution

_No response_

### Additional Context

_No response_

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 6.3
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #53551
| License       | MIT

### DIFF
--- a/src/Symfony/Bundle/FrameworkBundle/Command/CachePoolClearCommand.php
+++ b/src/Symfony/Bundle/FrameworkBundle/Command/CachePoolClearCommand.php
@@ -70,7 +70,7 @@ EOF
         $clearers = [];
 
         $poolNames = $input->getArgument('pools');
-        if ($input->getOption('all')) {
+        if ($clearAll = $input->getOption('all')) {
             if (!$this->poolNames) {
                 throw new InvalidArgumentException('Could not clear all cache pools, try specifying a specific pool or cache clearer.');
             }
@@ -84,7 +84,7 @@ EOF
         foreach ($poolNames as $id) {
             if ($this->poolClearer->hasPool($id)) {
                 $pools[$id] = $id;
-            } else {
+            } elseif (!$clearAll || $kernel->getContainer()->has($id)) {
                 $pool = $kernel->getContainer()->get($id);
 
                 if ($pool instanceof CacheItemPoolInterface) {


xabbuh:
The root cause for the observed behaviour is that the list of pool names passed to the constructor of `CachePoolClearCommand` command contains also the ids of cache pools that are later cleaned up by the `RemoveUnusedDefinitionsPass`. An alternative solution would be to collect available cache pool names at a later stage during compilation when the cleanup of unused service definitions has already happened.

nicolas-grekas:
Can't we check if the service exists before accessing it? Because this strategy is going to keep around pools that are not used, preventing them to be cleaned up by the compilation process.

xabbuh:
indeed, that's a better idea

OskarStark (approved):

OskarStark:
Why the Validator label?

xabbuh:
probably because I made a typo in the initial title and didn't remove the label when fixing that so Carson re-adds it again

src/Symfony/Bundle/FrameworkBundle/Command/CachePoolClearCommand.php
```diff
@@ -70,7 +70,9 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $clearers = [];
 
         $poolNames = $input->getArgument('pools');
+        $clearAll = false;
         if ($input->getOption('all')) {
+            $clearAll = true;
```
nicolas-grekas:
```suggestion
        if ($clearAll = $input->getOption('all')) {
```

nicolas-grekas (approved):
LGTM with one minor tweak

nicolas-grekas:
Thank you @xabbuh.

PR #53573, opened 2024-01-18
[Mailer] add Azure and Resend support to the UnsupportedSchemeException

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

### DIFF
--- a/src/Symfony/Component/Mailer/Exception/UnsupportedSchemeException.php
+++ b/src/Symfony/Component/Mailer/Exception/UnsupportedSchemeException.php
@@ -20,6 +20,10 @@ use Symfony\Component\Mailer\Transport\Dsn;
 class UnsupportedSchemeException extends LogicException
 {
     private const SCHEME_TO_PACKAGE_MAP = [
+        'azure' => [
+            'class' => Bridge\Azure\Transport\AzureTransportFactory::class,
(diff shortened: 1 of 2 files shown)


fabpot (approved):
@OskarStark I'm wondering if we should create a checklist of things we should not forget when adding a new mailer/notifier (and publish it in the docs). WDYT?

fabpot:
Thank you @xabbuh.

OskarStark:
> @OskarStark I'm wondering if we should create a checklist of things we should not forget when adding a new mailer/notifier (and publish it in the docs). WDYT?

Makes sense, if I find some spare time, I will do it.

PS: Another option would be to have dedicated Carsonbot comment with the checklist when a new bridge is added

fabpot:
> > @OskarStark I'm wondering if we should create a checklist of things we should not forget when adding a new mailer/notifier (and publish it in the docs). WDYT?
> 
> Makes sense, if I find some spare time, I will do it.
> 
> PS: Another option would be to have dedicated Carsonbot comment with the checklist when a new bridge is added

Carsonbot would be ideal but probably more complex to do. Starting with a manual list will be easier and bring a lot of value.

PR #53576, opened 2024-01-18
[Console] Only execute additional checks for color support if the output is a TTY

### LINKED ISSUE #53460: Color support detectiong change in 6.3.11 introduced a regression

6.3.11

### Description

See steps the reproduce.

### How to reproduce

I have a custom KISS e2e test script for testing `composer audit` command outputs.

It looks like the [color support detection change](https://github.com/symfony/console/compare/v6.3.9...v6.3.11) that was introduced in 6.3.11 introduced a regression and the output of `composer audit --format=json` become colorized that makes the JSON invalid.

```php
$audit_output = $argv[1] ?? (stream_get_contents(STDIN) ?: null);
if (null === $audit_output) {
    throw new \LogicException('Missing "composer audit" command output.');
}


fwrite(STDERR, $audit_output);


try {
    $audit_result = json_decode($audit_output, true, flags: JSON_THROW_ON_ERROR);
} catch (JsonException $e) {
    throw new \LogicException(sprintf('Malformed JSON input: "%s". %s', base64_encode(gzdeflate($audit_output, 9)), $e->getMessage()), 0, $e);
}
```
(Source: https://github.com/mxr576/ddqg-composer-audit/blob/1.0.0-rc4/tests/fixtures/e2e/test.php#L18-L29)

```sh
tests/fixtures/e2e/test.php "$(DDQG_COMPOSER_AUDIT_TEST_ENV=1 ./vendor/bin/composer -d tests/fixtures/e2e audit --format=json)"
```

(Source: https://github.com/mxr576/ddqg-composer-audit/blob/1.0.0-rc4/.github/workflows/run-e2e-test.yml#L51)

### Possible Solution

_No response_

### Additional Context

_No response_

theofidry:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #53460
| License       | MIT

As reported in #53460, the modifications done in #52940 are incorrect as it results in detecting that the stream can support colors despite not being a TTY.

Rather than doing a simple revert which would re-introduce the pre-existing issue that #52940 attempted to fix, this PR checks if the output is a TTY based on Composer's code and does this check before anything else.

Indeed a TTY only means that colors _may_ be supported, so the various checks that we were doing do make sense to be done but should be done after this TTY (so `Hyper` is not an exception, it can be a TTY or not).

### DIFF
--- a/src/Symfony/Component/Console/Output/StreamOutput.php
+++ b/src/Symfony/Component/Console/Output/StreamOutput.php
@@ -95,6 +95,10 @@ class StreamOutput extends Output
             return false;
         }
 
+        if (!$this->isTty()) {
+            return false;
+        }
+
         if (\DIRECTORY_SEPARATOR === '\\'
             && \function_exists('sapi_windows_vt100_support')
             && @sapi_windows_vt100_support($this->stream)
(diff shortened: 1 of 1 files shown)


src/Symfony/Component/Console/Output/StreamOutput.php
```diff
@@ -105,7 +109,36 @@ protected function hasColorSupport()
...
+            || str_starts_with((string) getenv('TERM'), 'xterm');
+    }
+
+    /**
+     * Checks if the stream is a TTY, i.e; whether the output stream is connected to a terminal.
+     *
+     * Reference: Composer\Util\Platform::isTty
+     * https://github.com/composer/composer
+     */
+    protected function isTty(): bool
```
theofidry: I extracted into a dedicated method for clarity, but this may be a problem especially for 5.4?

fabpot:
```suggestion
    private function isTty(): bool
```

fabpot (changes requested):

src/Symfony/Component/Console/Output/StreamOutput.php
theofidry: Done

xabbuh (approved):

fabpot (approved):

fabpot:
Thank you @theofidry.

PR #53586, opened 2024-01-19
[BrowserKit] do not try to read from empty filenames

### LINKED ISSUE #49014: dom-crawler + browser-kit empty file upload error "Path cannot be empty"

6.0.11

### Description

Using Symfony's dom-crawler and browser-kit (from Goutte) I load a simple page with a simple form: https://webblocks.nl/_form1.html and submit it without giving a file. I would expect an empty file to be uploaded (like a real browser does), or no file to be uploaded at all, but instead browser-kit crashes because it tries to read the no-file like it's a real file.

I thought it had something to do with the special/nested file input name: `files[b]` instead of `file`, but they both break! Special/nested file input name in this form: https://webblocks.nl/_form2.html both it breaks the same exact way.

### How to reproduce

`composer require fabpot/goutte`

```php
<?php
require 'vendor/autoload.php';
$goutte = new \Goutte\Client();
$crawler = $goutte->request('GET', 'https://webblocks.nl/_form1.html'); // OR:
// $crawler = $goutte->request('GET', 'https://webblocks.nl/_form2.html');
$form = $crawler->selectButton('Submit it')->form();
$crawler = $goutte->submit($form, [
	'values' => [
		'text' => 'Oele',
	],
]);
```

Error:

> ValueError: Path cannot be empty

From `vendor/symfony/mime/Part/DataPart.php:68`:

```php
if (false === $handle = @fopen($path, 'r', false)) {
    throw new InvalidArgumentException(sprintf('Unable to open path "%s".', $path));
```

It doesn't crash on the `InvalidArgumentException`, but the `ValueError` from `fopen` I assume. But that's not the point.

`dom-crawler` gives `browser-kit` an empty file, because it sees a file input, and the file is never filled, because I only submit `values[text]=Oele`

----

I'm using the Goutte client, because it's an easy package, but I assume it works with just `dom-crawler` and `browser-kit`.

### Possible Solution

The best solution would be to do exactly what the browser does: send a non-file. I have no idea how that works. Somewhere along the process browser-kit knows the file is empty:

```
^ array:1 [▼
  "file" => array:5 [▼
    "name" => ""
    "type" => ""
    "tmp_name" => ""
    "error" => 4
    "size" => 0
  ]
]
```

but it still tries to read it etc.

The second best solution would be to just completely ignore the file and not send it at all. dom-crawler's `Form::getFiles()` could filter on emptiness.

### Additional Context

The full browser-kit `Request` just before it all breaks:

```
^ Symfony\Component\BrowserKit\Request {#686 ▼
  #uri: "https://webblocks.nl/_http_server.php"
  #method: "POST"
  #parameters: array:2 [▼
    "op" => "Submit it"
    "values" => array:1 [▼
      "text" => "Oele"
    ]
  ]
  #files: array:1 [▼
    "file" => array:5 [▼
      "name" => ""
      "type" => ""
      "tmp_name" => ""
      "error" => 4
      "size" => 0
    ]
  ]
  #cookies: []
  #server: array:4 [▼
    "HTTP_USER_AGENT" => "Symfony BrowserKit"
    "HTTP_REFERER" => "https://webblocks.nl/_form1.html"
    "HTTP_HOST" => "webblocks.nl"
    "HTTPS" => true
  ]
  #content: null
}
```

I feel like I must be doing something wrong, because this is such an obvious bug, but I'm not doing anything weird, just load a form and submit it without a file. Has nobody ever done that? How can this be a bug?

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #49014
| License       | MIT

### DIFF
--- a/src/Symfony/Component/BrowserKit/HttpBrowser.php
+++ b/src/Symfony/Component/BrowserKit/HttpBrowser.php
@@ -144,7 +144,7 @@ class HttpBrowser extends AbstractBrowser
             if (!isset($file['tmp_name'])) {
                 $uploadedFiles[$name] = $this->getUploadedFiles($file);
             }
-            if (isset($file['tmp_name'])) {
+            if (isset($file['tmp_name']) && '' !== $file['tmp_name']) {
                 $uploadedFiles[$name] = DataPart::fromPath($file['tmp_name'], $file['name']);
             }
         }


rudiedirkx:
This does fix the exception from #49014 BUT it doesn't do what a browser does, and it might break the application behind the form. Instead of sending an empty/non-file, it sends the file field as a normal field with empty string:

```
$_POST:
Array
(
    [op] => Submit it
    [values] => Array
        (
            [text] => Oele
        )
    [file] => 
)
```

(see the `[file]`) What it should do is what a real browser does: send a non-file, or whatever you would call that:

```
$_POST:
Array
(
    [op] => Submit it
    [values] => Array
        (
            [text] => asd
        )
)

$_FILES:
Array
(
    [file] => Array
        (
            [name] => 
            [full_path] => 
            [type] => 
            [tmp_name] => 
            [error] => 4
            [size] => 0
        )
)
```

`file` is in the $_FILES list, with error 4: `UPLOAD_ERR_NO_FILE`: https://www.php.net/manual/en/features.file-upload.errors.php#features.file-upload.errors That's what happens if you submit a file field without file.

IMO deleting the file from the form is a better solution than the empty string, but sending the non-file would be even more realistic, but I don't know how symfony/curl would do that.

nicolas-grekas:
Closing in favor of #59752, thanks for having a look!

PR #53589, opened 2024-01-19
#48167 add fallbackCallbackAccessor to PropertyPathAccessor to suppor…

### LINKED ISSUE #48167: [Form] Overriding only `getter` causes exceptions while submitting

5.2.0 upwards

### Description

Symfony 5.2 introduced the `getter` and `setter` options for form types, but the separation is not complete causing exceptions during form submission.

### How to reproduce

Consider this use case on an object tree inheriting settings:
```php
    private ?int $value;
    
    public function setValue(?int $newValue): void
    {
        $this->value = $newValue;
    }

    public function getConfiguredValue(): ?int
    {
        return $this->value;
    }

    public function getActualValue(): int
    {
        return $this->value ?? $this->parent->getActualValue();
    }
```
In this case, the getter is not called `getValue` on purpose, to avoid confusion with programmers calling the wrong one. Since Symfony 5.2 I can now map this in a form:
```php
    $builder
        ->add('value', IntegerType::class, [
            'getter' => fn($object) => $object->getConfiguredValue(),
        ])
```
This works *fine* to render the form. Submitting it however crashes with a `NoSuchPropertyException`, stating `Can't get a way to read the property "value" in class XXX`. Note that it doesn't know how to *read* the property: this is because there is no `setter` defined, as the default works fine, and as such the `PropertyPathAccessor` is used to set the value, which reads the property to detect changes first, is blissfully unaware of the `getter` option, and therefore tries to read the propery via the accessor.

### Possible Solution

Right now the only working solution without modifying the original object is to add a `setter` as well, which is pure boilerplate as it does nothing special:
```php
          'setter' => function($object, ?int $newValue) {
              $object->setValue($newValue);
          },
```

So solutions we could implement are:
 - Make `getter` and `setter` mandatory in pairs - setting one requires the other. This is currently in practice enforced by the implementation, yet it does not throw an error if you don't. Seems weak as in the given example the setter is completely obsolete.
 - Make the DataAccessor properly aware of the existence of a `getter` when defaulting the set behavior, bypassing default PropertyAccess behavior that tries to read the property for changes.

Obviously the second solution is far more correct and better, it is also more complex and may introduce new edge cases.

### Additional Context

_No response_

bartrail:
…t custom getter methods when submitting a form

| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #48167
| License       | MIT

Added a private `CallbackAccessor` to the `PropertyPathAccessor` to have a 2nd look if there is a custom getter defined in the form config and use this as fallback..

### DIFF
--- a/src/Symfony/Component/Form/Extension/Core/DataAccessor/PropertyPathAccessor.php
+++ b/src/Symfony/Component/Form/Extension/Core/DataAccessor/PropertyPathAccessor.php
@@ -109,8 +109,7 @@ class PropertyPathAccessor implements DataAccessorInterface
 
             if (!$e instanceof UninitializedPropertyException
                 // For versions without UninitializedPropertyException check the exception message
-                && (class_exists(UninitializedPropertyException::class) || false === strpos($e->getMessage(),
-                        'You should initialize it'))
+                && (class_exists(UninitializedPropertyException::class) || false === strpos($e->getMessage(), 'You should initialize it'))
             ) {
                 throw $e;
             }
--- a/src/Symfony/Component/Form/Extension/Core/DataAccessor/PropertyPathAccessor.php
+++ b/src/Symfony/Component/Form/Extension/Core/DataAccessor/PropertyPathAccessor.php
@@ -30,10 +30,12 @@ use Symfony\Component\PropertyAccess\PropertyPathInterface;
 class PropertyPathAccessor implements DataAccessorInterface
 {
     private $propertyAccessor;
+    private $fallbackCallbackAccessorForCustomGetter;
 
     public function __construct(?PropertyAccessorInterface $propertyAccessor = null)
     {
         $this->propertyAccessor = $propertyAccessor ?? PropertyAccess::createPropertyAccessor();
+        $this->fallbackCallbackAccessorForCustomGetter = new CallbackAccessor();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Symfony/Component/Form/Extension/Core/DataAccessor/PropertyPathAccessor.php
```diff
@@ -30,10 +30,12 @@
 class PropertyPathAccessor implements DataAccessorInterface
 {
     private $propertyAccessor;
+    private $fallbackCallbackAccessorForCustomGetter;
 
     public function __construct(?PropertyAccessorInterface $propertyAccessor = null)
     {
         $this->propertyAccessor = $propertyAccessor ?? PropertyAccess::createPropertyAccessor();
+        $this->fallbackCallbackAccessorForCustomGetter = new CallbackAccessor();
```
yceruto, line 38: This could be simpler, but it's limited to falling back on `CallbackAccessor` only, when it could be any `DataAccessorInterface` adapter.

See https://github.com/symfony/symfony/pull/54723 for a more generic solution.

yceruto (changes requested):

xabbuh:
Thank you for starting this Conrad. Closing in favour of #54723.

PR #53592, opened 2024-01-20
[Routing] Localized Passthrough Locale for #[Route()]

cyraid:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| License       | MIT

As per reference of #53539, I was looking for a way to solve:
https://stackoverflow.com/questions/41721408/symfony-default-locale-without-specifying-in-url/77819142

Currently the first path (either specifying a path, or a path array) the first path index is 0 in AttributeClassLoader.php. Taking advantage of this, my answer in the Stack Overflow and in #53539 works, but is not advised as this is not documented. My PR adds support for:

```php
    #[Route(path: ['{_locale}' => '/{_locale}', 'nl' => '/nl'], name: 'action')]
```

Which adds the path as 'action' and 'action.nl', means the `{_locale}` can be used to catch all other locales. As @xabbuh and I talked about. Using this:

```php
    #[Route(path: ['/{_locale}', 'nl' => '/nl'], name: 'action')]
```

Is still allowed as the first array index of 0 was always allowed anyway (as the Route would turn the `$path` into `$paths[]`), and should remain to not break backwards compatibility with anyone using it, but if `1 =>` etc (integer array keys) used, it will raise an exception (also tested in the Tests).

If you guys could look this over that would be great. If someone needs any clarification for anyone wanting to document this feature near https://symfony.com/doc/current/routing.html#localized-routes-i18n that also would be great!

### DIFF
--- a/src/Symfony/Component/Routing/Loader/AttributeClassLoader.php
+++ b/src/Symfony/Component/Routing/Loader/AttributeClassLoader.php
@@ -201,8 +201,7 @@ abstract class AttributeClassLoader implements LoaderInterface
             }
         }
 
-        if (isset($paths[0]))
-        {
+        if (isset($paths[0])) {
             $paths['{_locale}'] = isset($paths['{_locale}']) ? $paths['{_locale}'] : $paths[0];
             unset($paths[0]);
         }
@@ -212,7 +211,7 @@ abstract class AttributeClassLoader implements LoaderInterface
             $this->configureRoute($route, $class, $method, $annot);
             if (\is_int($locale)) {
                 throw new \InvalidArgumentException(sprintf('Indexed locale paths in the Route paths array are not supported (%d:%s).', $locale, $path));
-            } else if ('{_locale}' === $locale) {
+            } elseif ('{_locale}' === $locale) {
                 $collection->add($name, $route, $priority);
             } else {
                 $route->setDefault('_locale', $locale);
--- a/src/Symfony/Component/Routing/Loader/AttributeClassLoader.php
+++ b/src/Symfony/Component/Routing/Loader/AttributeClassLoader.php
@@ -202,7 +202,7 @@ abstract class AttributeClassLoader implements LoaderInterface
         }
 
         if (isset($paths[0])) {
-            $paths['{_locale}'] = isset($paths['{_locale}']) ? $paths['{_locale}'] : $paths[0];
+            $paths['{_locale}'] = $paths['{_locale}'] ?? $paths[0];
             unset($paths[0]);
         }
 
--- a/src/Symfony/Component/Routing/Loader/AttributeClassLoader.php
+++ b/src/Symfony/Component/Routing/Loader/AttributeClassLoader.php
@@ -201,16 +201,24 @@ abstract class AttributeClassLoader implements LoaderInterface
             }
         }
 
+        if (isset($paths[0]))
+        {
+            $paths['{_locale}'] = isset($paths['{_locale}']) ? $paths['{_locale}'] : $paths[0];
+            unset($paths[0]);
+        }
+
         foreach ($paths as $locale => $path) {
             $route = $this->createRoute($path, $defaults, $requirements, $options, $host, $schemes, $methods, $condition);
             $this->configureRoute($route, $class, $method, $annot);
-            if (0 !== $locale) {
+            if (\is_int($locale)) {
+                throw new \InvalidArgumentException(sprintf('Indexed locale paths in the Route paths array are not supported (%d:%s)', $locale, $path));
+            } else if ('{_locale}' === $locale) {
+                $collection->add($name, $route, $priority);
+            } else {
                 $route->setDefault('_locale', $locale);
                 $route->setRequirement('_locale', preg_quote($locale));
                 $route->setDefault('_canonical_route', $name);
                 $collection->add($name.'.'.$locale, $route, $priority);
-            } else {
-                $collection->add($name, $route, $priority);
             }
         }
     }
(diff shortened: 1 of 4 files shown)


src/Symfony/Component/Routing/Loader/AttributeClassLoader.php
```diff
@@ -201,16 +201,23 @@ protected function addRoute(RouteCollection $collection, object $annot, array $g
             }
         }
 
+        if (isset($paths[0])) {
+            $paths['{_locale}'] = $paths['{_locale}'] ?? $paths[0];
+            unset($paths[0]);
```
smnandre, line 206: So if both are set, `paths[0]` is silently removed ?

src/Symfony/Component/Routing/Loader/AttributeClassLoader.php
```diff
@@ -201,16 +201,23 @@ protected function addRoute(RouteCollection $collection, object $annot, array $g
...
             $this->configureRoute($route, $class, $method, $annot);
-            if (0 !== $locale) {
+            if (\is_int($locale)) {
+                throw new \InvalidArgumentException(sprintf('Indexed locale paths in the Route paths array are not supported (%d:%s).', $locale, $path));
+            } elseif ('{_locale}' === $locale) {
+                $collection->add($name, $route, $priority);
+            } else {
                 $route->setDefault('_locale', $locale);
                 $route->setRequirement('_locale', preg_quote($locale));
                 $route->setDefault('_canonical_route', $name);
```
smnandre, line 219: Are you sure this route exists now ?

src/Symfony/Component/Routing/Loader/AttributeClassLoader.php
cyraid: Yes, so that two aren't added. :)
The logic is that it silently removes 0 to make backwards compatibility work. Previously index 0 is accepted and works. Just didn't want two added incase both were there.

src/Symfony/Component/Routing/Loader/AttributeClassLoader.php
cyraid: The tests work. ;)

smnandre:
Currently, it is possible to define a **default path**, and override it **only for some** locales

```php
#[Route(['/football', 'us' => '/soccer'])]
public function football(): Response
{
```

After this PR it will not be possible anymore ?

cyraid:
@smnandre  Sorry if I wasn't clear in my original post, but yes, the array 0 catch will still work. Xabbuh mentioned it may not be stable behaviour and not documented, so I went out to make a solution that works and supports that too. :)

Edit: There is a test in the PR that also makes sure what you posted will still work. ^_^

smnandre:
> @smnandre Sorry if I wasn't clear in my original post, but yes, the array 0 catch will still work. Xabbuh mentioned it may not be stable behaviour and not documented, so I went out to make a solution that works and supports that too. :)
> 
> Edit: There is a test in the PR that also makes sure what you posted will still work. ^_^

Could you show me this test ? I cannot find one that set a default value without settings a locale parameter.

cyraid:
Essentially:
```php
 #[Route(path: ['/{_locale}', 'nl' => '/nl'], name: 'action')]
```

This is the default path (array index 0). I could've put '/something' there and it would essentially be the same. If BOTH `{_locale}` index and index 0 (default path) are included, it will use the `{_locale}`, as making two paths with the same name wouldn't really work.

As per mine and Xabbuh conversation in #53539, the behaviour of the 'default path' doesn't seem to be standardized and is not documented if you take a look at: https://symfony.com/doc/current/routing.html#localized-routes-i18n

By including a `{_locale}`, I'm essentially opening the door for future possibility with it if default path were somehow to be deprecated. I do essentially give the 'default path' answer in: https://stackoverflow.com/questions/41721408/symfony-default-locale-without-specifying-in-url/77819142

nicolas-grekas:
> Currently the first path (either specifying a path, or a path array) the first path index is 0 [...] but is not advised as this is not documented.

If it works already, why not document it (and test it) properly instead of adding anything new?

cyraid:
@nicolas-grekas Oh I agree. Just did what Xabbuh suggested as per #53539 . I'm fine with you guys ignoring this PR and documenting (and adding a test) the default Path, I wouldn't be offended.

cyraid:
@nicolas-grekas Perhaps I could turn this PR into just adding a test for the Default path, if someone would do the honours of adding documentation?

nicolas-grekas:
Let's make another PR targeting 5.4 so that we can keep track of this patch as you did it first?

nicolas-grekas:
I think the current way is on purpose. PR welcome to add a test case, and for the doc also!

cyraid:
@nicolas-grekas The AttributeClassLoader in the tests has changed from 5.4/6.* to 7.0. Would I create a test for 5.4 (and for other times it's changed), or just create some tests for 7.0/7.1+ ?

nicolas-grekas:
For 5.4 that's perfect, then we'll take care of the merge up.

cyraid:
@nicolas-grekas I mean that the file names change, the class names change, and etc. So you couldn't merge up to a file that doesn't exist in 7.0+. :)

Edit:

7.0+: Removal of AbstractAnnotationLoaderTestCase.php+AnnotationClassLoaderTestCase.php+AnnotationClassLoaderWithAnnotationsTest.php+AnnotationClassLoaderWithAttributesTest.php, Added Symfony\Component\Routing\Tests\Loader\AttributeClassLoaderTest.php
6.4: Removal of AbstractAnnotationLoaderTestCase.php
5.4-6.0: Symfony\Component\Routing\Tests\Loader\AbstractAnnotationLoaderTestCase.php+AnnotationClassLoaderTestCase.php+AnnotationClassLoaderWithAnnotationsTest.php+AnnotationClassLoaderWithAttributesTest.php

PR #41353, opened 2021-05-21
[Cache] Removed depreacted code

malteschlueter:
| Q             | A
| ------------- | ---
| Branch?       | 6.0
| Bug fix?      | no
| New feature?  | no
| Deprecations? | yes
| Tickets       | -
| License       | MIT
| Doc PR        | -

This remove deprecated code from Cache.

Is it necessary for the Changelog or Upgrade.md that the classes are not configurable anymore?

PR #53593, opened 2024-01-20
[Cache] Fix possible infinite loop in `CachePoolPass`

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Introduced in #41353.

I'm not sure what the purpose of this code is exactly, since I haven't actually run into a problem, but it's definitely wrong this way.

### DIFF
--- a/src/Symfony/Component/Cache/DependencyInjection/CachePoolPass.php
+++ b/src/Symfony/Component/Cache/DependencyInjection/CachePoolPass.php
@@ -209,10 +209,10 @@ class CachePoolPass implements CompilerPassInterface
         }
 
         $notAliasedCacheClearerId = $this->cacheClearerId;
-        while ($container->hasAlias($this->cacheClearerId)) {
-            $this->cacheClearerId = (string) $container->getAlias($this->cacheClearerId);
+        while ($container->hasAlias($notAliasedCacheClearerId)) {
+            $notAliasedCacheClearerId = (string) $container->getAlias($notAliasedCacheClearerId);
         }
-        if ($container->hasDefinition($this->cacheClearerId)) {
+        if ($container->hasDefinition($notAliasedCacheClearerId)) {
             $clearers[$notAliasedCacheClearerId] = $allPools;
         }
 
--- a/src/Symfony/Component/Cache/Tests/DependencyInjection/CachePoolPassTest.php
+++ b/src/Symfony/Component/Cache/Tests/DependencyInjection/CachePoolPassTest.php
@@ -20,8 +20,10 @@ use Symfony\Component\Cache\Adapter\RedisAdapter;
 use Symfony\Component\Cache\DependencyInjection\CachePoolPass;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\HttpKernel\CacheClearer\Psr6CacheClearer;
 
 class CachePoolPassTest extends TestCase
 {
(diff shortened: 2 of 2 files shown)


xabbuh:
The change itself looks good. But can you please add a test case to prevent future regressions?

HypeMC:
> The change itself looks good. But can you please add a test case to prevent future regressions?

@xabbuh Done, but I had to make additional changes to the code. Either I misunderstood the intention, or this code is also broken on v5.4. If that's the case, please let me know, and I'll open a separate PR for v5.4, which should then be merged first.

src/Symfony/Component/Cache/DependencyInjection/CachePoolPass.php
```diff
@@ -181,11 +181,11 @@ public function process(ContainerBuilder $container)
             $container->removeDefinition('cache.early_expiration_handler');
         }
 
-        $notAliasedCacheClearerId = $aliasedCacheClearerId = 'cache.global_clearer';
-        while ($container->hasAlias('cache.global_clearer')) {
-            $aliasedCacheClearerId = (string) $container->getAlias('cache.global_clearer');
+        $notAliasedCacheClearerId = 'cache.global_clearer';
+        while ($container->hasAlias($notAliasedCacheClearerId)) {
+            $notAliasedCacheClearerId = (string) $container->getAlias($notAliasedCacheClearerId);
```
HypeMC, line 213: The `$clearers` array should only use non-alias names for keys since `$container->getDefinition()` (on line 193) doesn't work with aliases.

nicolas-grekas:
Can you please send the PR you have in mind for 5.4?
Also please target 6.3 for this one as we're going to do one more release of it as patch could be part of it.

HypeMC:
@nicolas-grekas I've retargeted this PR to 5.4. A conflict is inevitable when merging with the upstream, so IMO, it would be easiest to resolve it then. Here's the resolved code:

```diff
diff --git a/src/Symfony/Component/Cache/DependencyInjection/CachePoolPass.php b/src/Symfony/Component/Cache/DependencyInjection/CachePoolPass.php
index 5055ba9918..ab212e8ee2 100644
--- a/src/Symfony/Component/Cache/DependencyInjection/CachePoolPass.php
+++ b/src/Symfony/Component/Cache/DependencyInjection/CachePoolPass.php
@@ -181,11 +181,11 @@ class CachePoolPass implements CompilerPassInterface
             $container->removeDefinition('cache.early_expiration_handler');
         }
 
-        $notAliasedCacheClearerId = $aliasedCacheClearerId = 'cache.global_clearer';
-        while ($container->hasAlias('cache.global_clearer')) {
-            $aliasedCacheClearerId = (string) $container->getAlias('cache.global_clearer');
+        $notAliasedCacheClearerId = 'cache.global_clearer';
+        while ($container->hasAlias($notAliasedCacheClearerId)) {
+            $notAliasedCacheClearerId = (string) $container->getAlias($notAliasedCacheClearerId);
         }
-        if ($container->hasDefinition($aliasedCacheClearerId)) {
+        if ($container->hasDefinition($notAliasedCacheClearerId)) {
             $clearers[$notAliasedCacheClearerId] = $allPools;
         }
```

If needed, I'll create a new PR for 6.3.

src/Symfony/Component/Cache/Tests/DependencyInjection/CachePoolPassTest.php
```diff
@@ -232,4 +233,35 @@ public function testChainAdapterPool()
...
+        $container->setDefinition('app.cache_pool', $cachePool);
+
+        $this->cachePoolPass->process($container);
+
+        $definition = $container->getDefinition('cache.foo_bar_clearer');
+
+        $this->assertTrue($definition->hasTag('cache.pool.clearer'));
+        $this->assertCount(1, $arguments = $definition->getArguments());
+        $this->assertIsArray($firstArgument = reset($arguments));
+        $this->assertArrayHasKey('app.cache_pool', $firstArgument);
```
xabbuh: Can we replace the three lines above with something like `$this->assertEquals(['app.cache_pool' => new Reference('app.cache_pool')], $definition->getArgument(0));`?

HypeMC: Done

nicolas-grekas:
Thank you @HypeMC.

xabbuh (approved):

PR #53594, opened 2024-01-20
[Notifier] Updated the NTFY notifier to run without a user parameter

lostfocus:
| Q             | A
| ------------- | ---
| Branch?       | 6.4
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

The ntfy integration of the notifier component was missing the option to run without set username.

### DIFF
--- a/src/Symfony/Component/Notifier/Bridge/Ntfy/NtfyTransport.php
+++ b/src/Symfony/Component/Notifier/Bridge/Ntfy/NtfyTransport.php
@@ -84,6 +84,8 @@ final class NtfyTransport extends AbstractTransport
 
         if (null !== $this->user && null !== $this->password) {
             $headers['Authorization'] = 'Basic '.rtrim(base64_encode($this->user.':'.$this->password), '=');
+        } elseif (null !== $this->password) {
+            $headers['Authorization'] = 'Bearer '.$this->password;
         }
 
         $response = $this->client->request('POST', ($this->secureHttp ? 'https' : 'http').'://'.$this->getEndpoint(), [
@@ -115,8 +117,8 @@ final class NtfyTransport extends AbstractTransport
 
     public function supports(MessageInterface $message): bool
     {
-        return $message instanceof PushMessage &&
-            (null === $message->getOptions() || $message->getOptions() instanceof NtfyOptions);
+        return $message instanceof PushMessage
+            && (null === $message->getOptions() || $message->getOptions() instanceof NtfyOptions);
     }
 
     public function __toString(): string
--- a/src/Symfony/Component/Notifier/Bridge/Ntfy/NtfyTransportFactory.php
+++ b/src/Symfony/Component/Notifier/Bridge/Ntfy/NtfyTransportFactory.php
@@ -41,8 +41,11 @@ final class NtfyTransportFactory extends AbstractTransportFactory
             $transport->setPort($port);
         }
 
-        if (!empty($user = $dsn->getUser()) && !empty($password = $dsn->getPassword())) {
+        if (!empty($user = $dsn->getUser())) {
             $transport->setUser($user);
+        }
+
+        if (!empty($password = $dsn->getPassword())) {
             $transport->setPassword($password);
         }
 
--- a/src/Symfony/Component/Notifier/Bridge/Ntfy/Tests/NtfyTransportFactoryTest.php
+++ b/src/Symfony/Component/Notifier/Bridge/Ntfy/Tests/NtfyTransportFactoryTest.php
@@ -32,6 +32,10 @@ final class NtfyTransportFactoryTest extends TransportFactoryTestCase
             'ntfy://user:password@default/test',
         ];
         yield [
+            'ntfy://ntfy.sh/test',
+            'ntfy://:password@default/test',
+        ];
+        yield [
             'ntfy://ntfy.sh:8888/test',
             'ntfy://user:password@default:8888/test?secureHttp=off',
         ];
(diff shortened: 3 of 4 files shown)


smnandre:
Hi @lostfocus! 

With your implementation, we store two exclusive ways of authorization, that cannot be used together.

The [NTFY documentation](https://docs.ntfy.sh/publish/#access-tokens) makes an interesting point: the ntfy uses/accepts access token as "user-less" Basic authorization.

> Alternatively, you can use [Basic Auth](https://en.wikipedia.org/wiki/Basic_access_authentication) to send the access token. When sending an empty username, the basic auth password is treated by the ntfy server as an access token. This is primarily useful to make curl calls easier, e.g. curl -u:tk_AgQdq7mVBoFD37zQVN29RhuMzNIz2 ...:

This would be smaller to implement while beeing easier to maintain .. what do you think ? 


Something _similar_ is done in the Telegram Bridge: 

```php
private function getToken(Dsn $dsn): string
{
    if (null === $dsn->getUser() && null === $dsn->getPassword()) {
        throw new IncompleteDsnException('Missing token.', 'telegram://'.$dsn->getHost());
    }
    if (null === $dsn->getPassword()) {
        throw new IncompleteDsnException('Malformed token.', 'telegram://'.$dsn->getHost());
    }
    return sprintf('%s:%s', $dsn->getUser(), $dsn->getPassword());
}
```

src/Symfony/Component/Notifier/Bridge/Ntfy/NtfyTransportFactory.php
```diff
@@ -41,6 +41,10 @@ public function create(Dsn $dsn): TransportInterface
             $transport->setPort($port);
         }
 
+        if (!empty($token = $dsn->getOption('token'))) {
+            $transport->setToken($token);
+        }
+
```
OskarStark, line 47: Shall we verify that username, password AND token is not provided at the same time?

src/Symfony/Component/Notifier/Bridge/Ntfy/NtfyTransport.php
```diff
@@ -41,6 +42,13 @@ public function getTopic(): string
         return $this->topic;
     }
 
+    public function setToken(?string $token): self
```
smnandre:
```
public function(#[\SensitiveParameter] ?string $token): self
```

(Probably useless as you may remove the method :) )

lostfocus: Is `#[\SensitiveParameter]` already in the 6.4 branch? `setPassword` has it in 7.0 but not in 6.4.

lostfocus:
@smnandre yeah, I was looking into that. But if we have something like `NTFY_DSN=ntfy://:[:PASSWORD]@default/TOPIC` we're already losing the password when parsing the string into a `Symfony\Component\Notifier\Transport\Dsn` in `Symfony\Component\Notifier\Transport`. And I didn't want to open the can of worms of fiddling around in that one, too.

src/Symfony/Component/Notifier/Bridge/Ntfy/NtfyTransportFactory.php
lostfocus: That seems like an option and would probably also be an answer to @smnandre 's point in https://github.com/symfony/symfony/pull/53594#issuecomment-1902263539 ("With your implementation, we store two exclusive ways of authorization, that cannot be used together.")

src/Symfony/Component/Notifier/Bridge/Ntfy/NtfyTransport.php
smnandre: Sorry forgot you targetted 6.4. 

But as i'm not sure it will be considered a bug ... you may have to target 7.1 (let's wait to see what admins say)

smnandre:
> @smnandre yeah, I was looking into that. But if we have something like `NTFY_DSN=ntfy://:[:PASSWORD]@default/TOPIC` we're already losing the password when parsing the string into a `Symfony\Component\Notifier\Transport\Dsn` in `Symfony\Component\Notifier\Transport`. And I didn't want to open the can of worms of fiddling around in that one, too.

I'm not sure to understand, were are you losing the password ?

i think the "unified" sugegstion would be easier to develop, passing by user:password to the Basic or Bearer header with (almost) no manipulation.

And it would be easier for users to switch from Token to user/password :)

lostfocus:
> I'm not sure to understand, were are you losing the password ?

Okay, yeah, that was a brainfart. For some reason I thought `parse_url` couldn't handle that. I'll reshuffle things tomorrow. (And maybe (_just maybe_) it could even be considered a bugfix then.)

src/Symfony/Component/Notifier/Bridge/Ntfy/NtfyTransport.php
```diff
@@ -115,8 +117,8 @@ protected function doSend(MessageInterface $message): SentMessage
 
     public function supports(MessageInterface $message): bool
     {
-        return $message instanceof PushMessage &&
-            (null === $message->getOptions() || $message->getOptions() instanceof NtfyOptions);
+        return $message instanceof PushMessage
+            && (null === $message->getOptions() || $message->getOptions() instanceof NtfyOptions);
```
OskarStark, line 121: @keradus is there a PHP-CS-Fixer rule for this?

smnandre: https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/src/Fixer/Operator/OperatorLinebreakFixer.php ?

nicolas-grekas:
Thank you @lostfocus.

PR #53598, opened 2024-01-21
[Cache][Lock][Semaphore] Optimize factories

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Inspired by https://github.com/symfony/symfony/pull/53448#discussion_r1444029069


nicolas-grekas:
I doubt this matters, and I also dunno if it's faster at all. PCRE is way heavier that a simple string check.
Also, that's not an improvement to the readability of the code base to me and it will make reviewing PRs that add new schemes harder.
I'm closing therefor. Thanks for proposing.

HypeMC:
@nicolas-grekas When I ran the benchmark for #53448, it was marginally faster, but I agree that readability takes a hit. I just wanted someone to confirm that it's not worth it, thanks :smile:

PR #53605, opened 2024-01-23
[DependencyInjection] Allow registering attributes on classes that cannot be instantiated

### LINKED ISSUE #48392: [DependencyInjection] Allow registering attributes on classes that cannot be instantiated
### Description

In our application we rely Symfony's `ContainerBuilder::registerAttributeForAutoconfiguration` together with auto wire for a lot of things. I also helped improving it:
- https://github.com/symfony/symfony/pull/42039

But now I have an issue. I have a class that has a `private` constructor. It needs to be constructed by calling a static method. 
Example:
```php
#[SomeAttribute]
final class SomeService
{
    private function __construct() {}
}
```

I have a compiler pass that tries to work this class by looking for `SomeAttribute`:
```php
$container->registerAttributeForAutoconfiguration(
    SomeAttribute::class,
    static function (ChildDefinition $definition, SomeAttribute $attribute, ReflectionClass $reflector) : void {
        // do magic
    },
);
```

I load everything using:
```
$services->load('App\\', __DIR__ . '/');
```

Because of this line the class is never loaded because it's not instantiable.
https://github.com/symfony/symfony/blob/69f46f231b16be1bce1e2ecfa7f9a1fb192cda22/src/Symfony/Component/DependencyInjection/Loader/FileLoader.php#L232-L234

Possible solutions:

1. Always run the registered attributes configurators for classes that cannot be instantiated
2. Allow an additional option on the `PrototypeConfigurator` to indicate you want to always load attributes
3. Change the condition above:
    ```diff
    - if ($r->isInstantiable() || $r->isInterface()) {
    -    $classes[$class] = null;
    - }
    + $classes[$class] = null;
    ```

/cc @nicolas-grekas since we worked in this area together in #42039 do you have ideas? 

### Example

_No response_

ruudk:
| Q             | A
| ------------- | ---
| Branch?       | 7.1
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #48392
| License       | MIT

This is a proof of concept for https://github.com/symfony/symfony/issues/48392

Needs some tests.

### DIFF
--- a/src/Symfony/Component/DependencyInjection/Compiler/AbstractRecursivePass.php
+++ b/src/Symfony/Component/DependencyInjection/Compiler/AbstractRecursivePass.php
@@ -65,8 +65,6 @@ abstract class AbstractRecursivePass implements CompilerPassInterface
 
     /**
      * Processes a value found in a definition tree.
-     *
-     * @return mixed
      */
     protected function processValue(mixed $value, bool $isRoot = false)
     {
@@ -76,9 +74,9 @@ abstract class AbstractRecursivePass implements CompilerPassInterface
                     continue;
                 }
                 if ($isRoot) {
-                    if ($v->hasTag('container.excluded')) {
-                        continue;
-                    }
+                    // if ($v->hasTag('container.excluded')) {
+                    //    continue;
+                    // }
                     $this->currentId = $k;
                 }
                 if ($v !== $processedValue = $this->processValue($v, $isRoot)) {
--- a/src/Symfony/Component/DependencyInjection/Loader/FileLoader.php
+++ b/src/Symfony/Component/DependencyInjection/Loader/FileLoader.php
@@ -294,6 +294,10 @@ abstract class FileLoader extends BaseFileLoader
 
             if ($r->isInstantiable() || $r->isInterface()) {
                 $classes[$class] = null;
+            } elseif (!$r->isInstantiable() && !$this->container->has($class)) {
+                $this->container->register($class, $class)
+                    ->setAutoconfigured(true)
+                    ->addTag('container.excluded');
             }
 
             if ($autoconfigureAttributes && !$r->isInstantiable()) {


nicolas-grekas (commented):
I think I had more something like below in mind.
Of course without any changes to AbstractRecursivePass.
If you want to process `container.excluded` services, you'd first need to extract them from the container and remove the tag on them.


```diff
--- a/src/Symfony/Component/DependencyInjection/Loader/FileLoader.php
+++ b/src/Symfony/Component/DependencyInjection/Loader/FileLoader.php
@@ -162,6 +162,11 @@ abstract class FileLoader extends BaseFileLoader
             if (interface_exists($class, false)) {
                 $this->interfaces[] = $class;
             } else {
+                $r = $this->container->getReflectionClass($class);
+                if (!$r->isInstantiable()) {
+                    $this->addContainerExcludedTag($class, $source);
+                    continue;
+                }
                 $this->setDefinition($class, $definition = $getPrototype());
                 if (null !== $errorMessage) {
                     $definition->addError($errorMessage);
@@ -179,7 +184,6 @@ abstract class FileLoader extends BaseFileLoader
                 if (!$autoconfigureAttributes) {
                     continue;
                 }
-                $r = $this->container->getReflectionClass($class);
                 $defaultAlias = 1 === \count($interfaces) ? $interfaces[0] : null;
                 foreach ($r->getAttributes(AsAlias::class) as $attr) {
```

nicolas-grekas:
(Loosely related: this makes me think we could add a `#[Constructor]` attribute that could be used on a static constructor to let autowiring spot them.)

ruudk:
> If you want to process container.excluded services, you'd first need to extract them from the container and remove the tag on them.

Hmm, but then I have to reflect on all of those excluded services, to find attributes on the class again. Basically, doing everything from scratch that Symfony already handles.

Does that make sense? We can rely on `registerAttributeForAutoconfiguration` for _almost_ everything (classes, methods, properties, etc...), except when the constructor is private. I feel like we should handle this, as I'm sure a lot of people are not aware of this hidden logic (private constructor means no service).

nicolas-grekas:
> then I have to reflect on all of those excluded services, to find attributes on the class again

not really: you can build a side container with only the service that are of interest to you, remove that tag from them, and run the compiler passes you want on it by compiling the container

> without any changes to AbstractRecursivePass

this is the critical part - we cannot change this behavior

but let me know if you have another idea that would fit this constraint

ruudk:
What if we do the following:

* Revert changes to `AbstractRecursivePass`
* Tag service with `container.not_instantiable` when not instantiable
* This allows processing the attributes
* When attributes are processed, have another tag that removes definitions tagged with `container.not_instantiable` from the container, just like `container.excluded`.
* Or, have a pass that applies `container.excluded` tag for definitions with `container.not_instantiable` tag.

pounard:
I like this issue because it's near from what I attempt to do: read attributes at compile time on non-services classes. Yet, near is not enough, because this probably would allow to deal with excluded services, but I'm not sure extending usage of this to, in my use case, messenger messages, would be accepted by core contributors. @nicolas-grekas what is your opinion about this?

nicolas-grekas:
I just reopened the topic and submitted #59712 to fill the gap.
Thanks for the help pushing this forward.

PR #53612, opened 2024-01-23
Apply php-cs-fixer fix --rules nullable_type_declaration_for_default_null_value

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 5.4
| Bug fix?      | no
| New feature?  |no
| Deprecations? |no
| Issues        | -
| License       | MIT

This one is important to make on 5.4 because it will likely fix a deprecation in a future PHP version, see https://externals.io/message/122214

(do not squash while merging)


nicolas-grekas:
/cc @keradus FYI

I think we should also make the null default value explicit for untyped properties. What's the fixer for that? :angel:

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Apply php-cs-fixer fix --rules nullable_type_declaration_for_default_null_value
+f4118e110a46de3ffb799e7d79bf15128d1646ea
```
fancyweb, line 2: :open_mouth:

nicolas-grekas: See https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view ;)

lyrixx:
I really prefer this way! Thanks 👍🏼

OskarStark:
> I really prefer this way! Thanks 👍🏼

Same here, this also makes diffs less noisy when converting to CPP 👏

nicolas-grekas:
Now merged up to 7.1, help wanted to fix tests :)

keradus:
@nicolas-grekas , if you wonder about such fixer, I don't think it exists, feel encouraged to drop the PR for it

```diff
 <?php
 
 class X {
-    public string $name;
+    public ?string $name;
 }
```

nicolas-grekas:
```diff
-    public string $name;
+    public ?string $name;
```

Be careful this wouldn't be equivalent! So no, I won't want this rule anytime :)

